Issue #2666, opened 2017-10-20
Make MAX_CONTEXT_SIZE dynamically sized to future-proof it


derekbruening:
On Skylake this assert fires:

```
Internal Error: DynamoRIO debug check failure: D:\dynamorio_package\core\win32\ntdll.c:5359 res >= 0 && len <= MAX_CONTEXT_SIZE
```

Xref #1937 for the corresponding Linux signal context issue.
We need to make MAX_CONTEXT_SIZE be dynamically sized.

newgre:
For the record: the extended context size for both, Skylake and Kaby Lake, is 0x4F7.
Would you be ok with an update to MAX_CONTEXT_SIZE until dynamic context size has been implemented?

newgre:
err, above size was the result without CONTEXT_XSTATE. The correct value should be 0x6EF and 0x4E3 for a 32-bit process.

derekbruening:
> Would you be ok with an update to MAX_CONTEXT_SIZE until dynamic context size has been implemented?

Yes that's a reasonable short-term fix.   The current value in the code today for 64-bit MAX_CONTEXT_SIZE is 0x6ef and for 32-bit 0x4e3 -- are you reporting a subset of the context above?
  Your values above wouldn't have triggered the assert.

newgre:
in 7 RC1 I see 0x680 and 0x480, respectively, defined in os_private.h?

derekbruening:
> Would you be ok with an update to MAX_CONTEXT_SIZE until dynamic context size has been implemented?

The only concern here is that stack buffers of this size are used in a number of places in the code and DR has a limited stack size.  If the MAX_CONTEXT_SIZE gets too much larger they will need to be changed to use the heap.

derekbruening:
> in 7 RC1 I see 0x680 and 0x480, respectively, defined in os_private.h?

Ah ok, I'm looking at the latest, and it was updated in 5eebc95e587b9753cacf58bec4b48a4d4c4b2813.  So perhaps the short-term fix has already been done?

newgre:
Oh neat, yes those values are the same that I deduced. Are there nightly builds?

newgre:
found them: https://github.com/DynamoRIO/dynamorio/wiki/Latest-Build

newgre:
no windows builds?

derekbruening:
[...]

PR #4122, opened 2020-02-21
i2666: increase MAX_CONTEXT_SIZE

lieser:
Increase `MAX_CONTEXT_SIZE` to `0xd2f`

Issue: #2666

### DIFF
--- a/core/win32/os_private.h
+++ b/core/win32/os_private.h
@@ -471,11 +471,11 @@ extern uint context_xstate;
 
 #define XSTATE_HEADER_SIZE 0x40 /* 512 bits */
 #define YMMH_AREA(ymmh_area, i) (((dr_xmm_t *)ymmh_area)[i])
-#define MAX_CONTEXT_64_SIZE 0x6ef /* as observed on win10-x64 */
+#define MAX_CONTEXT_64_SIZE 0xd2f /* as observed on win10-x64 */
 #ifdef X64
 #    define MAX_CONTEXT_SIZE MAX_CONTEXT_64_SIZE
 #else
-#    define MAX_CONTEXT_SIZE 0x4e3 /* as observed on win10-x64 */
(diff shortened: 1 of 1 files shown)


AssadHashmi:
run arm tests

johnfxgalea:
Many thanks for the PR! I'll let others involved with the issue review the PR.

Nit: Can you please add the # to the title please, i.e.,` i#2666: Increase MAX_CONTEXT_SIZE`

Thank you

core/win32/os_private.h
```diff
@@ -471,11 +471,11 @@ extern uint context_xstate;
 
 #define XSTATE_HEADER_SIZE 0x40 /* 512 bits */
 #define YMMH_AREA(ymmh_area, i) (((dr_xmm_t *)ymmh_area)[i])
-#define MAX_CONTEXT_64_SIZE 0x6ef /* as observed on win10-x64 */
+#define MAX_CONTEXT_64_SIZE 0xd2f /* as observed on win10-x64 */
 #ifdef X64
 #    define MAX_CONTEXT_SIZE MAX_CONTEXT_64_SIZE
 #else
-#    define MAX_CONTEXT_SIZE 0x4e3 /* as observed on win10-x64 */
+#    define MAX_CONTEXT_SIZE 0xd2f /* as observed on win10-x64 */
```
derekbruening: Surprising that 32-bit is just as big.  The only potential problem is that these structures are on the stack in a number of routines: so the risk is stack overflow with DR's tiny stacks.  But if the tests pass it should be ok.  Long-term we may need to move all CONTEXTs to the heap.

derekbruening:
So the CI is having issues: Travis is again hitting the weird hang *before* the ctest script runs which PR#41119 was trying to reproduce but failed.  AArch64 is hitting leaks in 2 of the tests: #4124.

Issue #4083, opened 2020-02-06
Add drcachesim function trace viewing tool


derekbruening:
The proposal is to create a simple analysis tool which will provide both a sequential view of the function trace data recorded in a drcachesim trace (through -record_heap or -record_function) as well as aggregate counts.

PR #4132, opened 2020-02-24
i#4083 func view: Add read-without-mapping module_mapper_t interface

derekbruening:
Adds a new module_mapper_t interface read_without_mapping_modules()
for parsing the module bounds without mapping the modules into memory.

Adds find_trace_module() for querying which module a given trace PC
came from, for use cases such as the function trace viewing tool (to
identify which symbol came from which library) where the data inside
the modules is not needed.

Adds a test.

Issue: #4083

### DIFF
--- a/clients/drcachesim/tests/raw2trace_io.cpp
+++ b/clients/drcachesim/tests/raw2trace_io.cpp
@@ -168,11 +168,48 @@ my_user_free(void *data)
 }
 
 bool
+test_module_query(const raw2trace_directory_t *dir)
+{
+    // Test the load-but-don't-map interface.
+    std::unique_ptr<module_mapper_t> mapper = module_mapper_t::create(
+        dir->modfile_bytes_, nullptr, nullptr, nullptr, &my_user_free);
+    EXPECT(mapper->get_last_error().empty(), "Module mapper construction failed");
+    REPORT("About to parse modules");
+    bool ok = mapper->read_without_mapping_modules();
+    EXPECT(ok && mapper->get_last_error().empty(), "Module loading failed");
+    REPORT("Parsed modules successfully");
+    const auto &modules = mapper->get_loaded_modules();
+    EXPECT(!modules.empty(), "Expected module entries");
+    EXPECT(mapper->get_last_error().empty(), "Module loading failed");
+    bool found_simple_app = false;
+    app_pc addr_in_app = nullptr;
+    for (const module_t &m : mapper->get_loaded_modules()) {
+        EXPECT(m.map_base == nullptr, "Module should not be mapped");
+        std::string path = m.path;
+        if (path.rfind("simple_app") != std::string::npos) {
+            found_simple_app = true;
+            addr_in_app = m.orig_base + m.map_size / 2;
+            break;
+        }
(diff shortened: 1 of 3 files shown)


derekbruening:
So my plan was to use this to qualify traced symbols with the library name, which seems very useful.  I would want to know whether I had traced libc!malloc or libfoo!malloc.  However, I did not plan ahead very well, and found out after implementing its use in the new viewing tool that reconstructing the library after the fact for the return point is not as simple as looking at the prior PC:

```
Function 'libstdc++.so.6.0.28!__GI___libc_malloc' (id=8)
       22 calls
        0 returns
Function 'libc-2.29.so!__GI___libc_malloc' (id=8)
        2 calls
       25 returns
Function 'ld-2.29.so!__GI___libc_malloc' (id=8)
        1 calls
        0 returns
```

So the callee wrap point has the instruction PC for the target, but the call site return point has the caller, resulting in mis-attributed functions for the pre point vs the post point.  Just for statistics here we could just not track returns, and the full trace view still looks good:

```
0x7f9461807d56 => 8 ld-2.29.so!__GI___libc_malloc(0x11c00)
    0x7f9461807d56 => 8 libc-2.29.so!__GI___libc_malloc(0x11c00) => 0x7f945e0e92a0
=> 0x7f945e0e92a0
0x7f946180b745 => 8 libstdc++.so.6.0.28!__GI___libc_malloc(0x20) => 0x7f945e0faeb0
0x7f9461616d8c => 8 libc-2.29.so!__GI___libc_malloc(0x400) => 0x7f945e0faee0
0x7f946180b745 => 8 libstdc++.so.6.0.28!__GI___libc_malloc(0x48) => 0x7f945e0fb2f0
0x7f946180b745 => 8 libstdc++.so.6.0.28!__GI___libc_malloc(0x1c) => 0x7f945e0fb340
```

But a more sophisticated analysis that wants to split by 'lib!func' and needs to correlate the func args with the return value is going to have to go keep a stack of calls to match things up, which sounds like a pain.

Maybe instead the better approach is to only support 'lib!func' up front at tracing time by taking in a 'lib!func' string there with its own id, and leave the viewing tool just tracking id's and the corresponding names from the tracing itself?  In that case I would not need this PR.  Maybe it would be useful for something else; or maybe it's better to just throw it out.

derekbruening:
OTOH some uses cases might want to trace '*!func' and split them up by library during analysis.  But it should be easier for the tracer to split them up: give each a different id when it asks to wrap them.  So maybe the -record_function interface should *not* take in an id, since we're now recording the strings in a file (decisions after offline discussion last week)?  And the tracer should just assign an atomically-incremented counter as the id.

s-kanev:
Agree that specifying 'lib!func' upfront up front sounds like the better way to go in general.

Just bear in mind for our internal case (at least the most common one, by a lot), the allocator will be linked in statically and called directly in the main module (and not a library). So, the discrepancy in that case shouldn't exist.

derekbruening:
Abandoning this since there is no current use case.  It might be useful in the future but it's not that hard to re-write and that seems cleaner than trying to keep it around.

Issue #4126, opened 2020-02-22
Travis builds are hanging mid-print


derekbruening:
Filing this since it keeps happening.  PR #4119 was my attempt to diagnose it.

The cross-compile job (the 3rd job) has been repeatedly failing ever since PR #4118 *except* in the diagnostic attempt.  It looks like this:

```
663$ suite/runsuite_wrapper.pl travis $EXTRA_ARGS
664
665
666No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
667Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received
668
669The build has been terminated
670
```

Very strange.

derekbruening:
I put the diagnostics into PR #4127 and it reproduces.  The hang seems to most likely be at the fork call:
```
$ suite/runsuite_wrapper.pl travis $EXTRA_ARGS
667Forking child for stdout tee
668
669
670No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

derekbruening:
Well, I added some prints before and around the fork, around the child exec, and I put in alarms in the parent while it tees the child stdout: and with that in there I can't reproduce the hang.  Strange.  Going to chalk it up to some perl bug and move on.

derekbruening:
It looks like the problem is still there.  I left in the -VV and 30s alarms, and we can see the build hang mid-print!

https://travis-ci.com/DynamoRIO/dynamorio/jobs/295543355
```
4342[ 84%] Building C object CMakeFiles/drmemorylib.dir/drmemory/slowpath.c.o
4343[ 84%] Building CXX object dynamorio/clients/drcachesim/CMakeFiles/drmemtrace_static.dir/tracer/instru_online.cpp.o
4344
4345xxxxxxxxxx 30s elapsed xxxxxxxxxxx
4346[ 85%] Building CXX object dynamorio/clients/drcachesim/CMakeFiles/drmemtrace_static.dir/tracer/physaddr.cpp.o
4347[ 85%] Building CXX object dynamorio/clients/drcachesim/CMakeFiles/drmemtrace_static.dir/tracer/func_trace.cpp.o
4348[ 85%] Building C object CMakeFiles/drmemorylib.dir/drmemory/fastpath.c.o
4349[ 85%] Building C object CMakeFiles/drmemorylib.dir/drmemory/stack.c.o
[...]

Issue #3980, opened 2019-12-11
ASSERT signal marker invariants in histogram.offline test on AArch64


derekbruening:
Xref #2899 where this test hangs.

This is a new assert on signal marker invariants and it just fired:

http://139.178.83.194:8080/job/DynamoRIO-AArch64-Precommit/887/console
```
171/176 Test #168: code_api|tool.histogram.offline .................***Failed    2.98 sec
Running cmd |/var/lib/jenkins/.jenkins/workspace/DynamoRIO-AArch64-Precommit/build/build_debug-internal-64/bin64/drrun;-s;90;-quiet;-debug;-killpg;-stderr_mask;0xC;-dumpcore_mask;0;-code_api;-t;drcachesim;-offline;-subdir_prefix;tool.histogram.offline;--;/var/lib/jenkins/.jenkins/workspace/DynamoRIO-AArch64-Precommit/build/build_debug-internal-64/suite/tests/bin/pthreads.ptsig|
Running postcmd |/var/lib/jenkins/.jenkins/workspace/DynamoRIO-AArch64-Precommit/build/build_debug-internal-64/clients/bin64/drcachesim;-indir;/var/lib/jenkins/.jenkins/workspace/DynamoRIO-AArch64-Precommit/build/build_debug-internal-64/suite/tests/tool.histogram.offline.pthreads.ptsig.03302.8806.dir|
Running postcmd2 |/var/lib/jenkins/.jenkins/workspace/DynamoRIO-AArch64-Precommit/build/build_debug-internal-64/clients/bin64/histogram_launcher;-test_mode;-trace_dir;/var/lib/jenkins/.jenkins/workspace/DynamoRIO-AArch64-Precommit/build/build_debug-internal-64/suite/tests/tool.histogram.offline.pthreads.ptsig.03302.8806.dir/trace|
CMake Error at /var/lib/jenkins/.jenkins/workspace/DynamoRIO-AArch64-Precommit/suite/tests/runmulti.cmake:106 (message):
  *** postcmd2 failed (Child aborted): histogram_launcher:
  /var/lib/jenkins/.jenkins/workspace/DynamoRIO-AArch64-Precommit/clients/drcachesim/tests/trace_invariants.cpp:143:
  virtual bool trace_invariants_t::process_memref(const memref_t&): Assertion
  `memref.instr.tid != pre_signal_instr.instr.tid || memref.instr.addr ==
  pre_signal_instr.instr.addr || memref.instr.addr ==
[...]

Issue #2892, opened 2018-03-23
drcachesim.delay-simple  and drcachesim.invariants  fail on ARM


fhahn:
There are 2 drcachesim test failures on ARM: drcachesim.delay-simple  and drcachesim.invariants 

The new AArch32 precommit builder shows the logs for the failing tests:
http://jenkins.fhahn.com:8080/job/DynamoRIO-AArch32-Precommit/5/console

```
136/153 Test #133: code_api|tool.drcachesim.delay-simple ...........***Failed  Required regular expression not found.Regex=[^Hit delay threshold: enabling tracing.
Exiting process after ~1.... references.
---- <application exited with code 0> ----
Cache simulation results:
Core #0 \(1 thread\(s\)\)
  L1I stats:
    Hits:                         *[0-9,\.]*
    Misses:                       *[0-9,\.]*.
.*   Miss rate:                   *[0-9,\.]*%
  L1D stats:
    Hits:                         *[0-9,\.]*
    Misses:                       *[0-9,\.]*.
.*   Miss rate:                   *[0-9,\.]*%
Core #1 \(0 thread\(s\)\)
Core #2 \(0 thread\(s\)\)
Core #3 \(0 thread\(s\)\)
LL stats:
    Hits:                         *[0-9,\.]*.
    Misses:                       *[0-9,\.]*..
.*   Local miss rate:             *[0-9,\.]*%
    Child hits:                   *[0-9,\.]*.
    Total miss rate:              *[0-9,\.]*%
$
]  3.87 sec
Hit delay threshold: enabling tracing.
Exiting process after ~6104240175158657024 references.
---- <application exited with code 0> ----
Cache simulation results:
Core #0 (1 thread(s))
  L1I stats:
    Hits:                           10,239
    Misses:                             48
    Miss rate:                        0.47%
  L1D stats:
    Hits:                            3,447
    Misses:                            114
    Prefetch hits:                      18
    Prefetch misses:                    96
    Miss rate:                        3.20%
Core #1 (0 thread(s))
Core #2 (0 thread(s))
Core #3 (0 thread(s))
LL stats:
    Hits:                                2
    Misses:                            160
    Prefetch hits:                       1
    Prefetch misses:                    95
    Local miss rate:                 98.77%
    Child hits:                     13,704
    Total miss rate:                  1.15%

152/153 Test #141: code_api|tool.drcachesim.invariants .............***Failed  Required regular expression not found.Regex=[^.*
[...]

PR #4133, opened 2020-02-24
i#4123: Look for symbols in special build-id subdir

### LINKED ISSUE #4123: drsyms does not look in /usr/lib/debug/.build-id/ directory for symbol files
There is a new scheme for storing separate debug info: for build id 0xabcdef.debug, the file is stored in /usr/lib/debug/.build-id/ab/cdef.debug and the debuglink section says "cdef.debug".  We should augment drsym's current /usr/lib/debug search scheme to include this.

derekbruening:
Adds to drsyms reading of the build id and searching in
/usr/lib/debug/.build-id for a separate debuginfo file, which is the
modern scheme for installed libraries.

Tested on internal symbols like __GI___libc_malloc on a local machine.
It is not easy to make an automated test for this without setting up a
chroot or something which does not seem worth the effort.

Fixes #4123

### DIFF
--- a/ext/drsyms/drsyms_elf.c
+++ b/ext/drsyms/drsyms_elf.c
@@ -1,5 +1,5 @@
 /* **********************************************************
- * Copyright (c) 2011-2015 Google, Inc.  All rights reserved.
+ * Copyright (c) 2011-2020 Google, Inc.  All rights reserved.
  * **********************************************************/
 
 /*
@@ -126,6 +126,8 @@ typedef struct _elf_info_t {
     byte *map_base;
     ptr_uint_t load_base;
     drsym_debug_kind_t debug_kind;
+#define MAX_BUILD_ID_LENGTH 128
+    char build_id[MAX_BUILD_ID_LENGTH];
 } elf_info_t;
 
 /* Looks for a section with real data, not just a section with a header */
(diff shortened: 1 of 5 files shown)


ext/drsyms/drsyms_elf.c
```diff
@@ -166,6 +168,47 @@ find_elf_section_by_name(Elf *elf, const char *match_name)
...
+        if (note->n_type == NT_GNU_BUILD_ID) {
+            /* Following the note are the name and value. */
+            byte *src = ((byte *)(note + 1)) + note->n_namesz;
+            size_t size = note->n_descsz;
+            if ((byte *)data->d_buf + data->d_size < src + note->n_descsz) {
+                NOTIFY_ELF("note data is shorter than specified length");
+                size = (byte *)data->d_buf + data->d_size - src;
+            }
+            char *dst = mod->build_id;
+            for (int i = 0; i < size; i++) {
```
snehasish: A couple of things in this loop took me a while to understand.
1. The +3 in the size check wasn't apparent until I read the dr_snprintf call below.
2. NULL_TERMINATE_BUFFER or similar is not required since each time dr_snprintf writes out 2 chars for the actual value and the null terminator. Then we overwrite the null terminator the next time around.

Perhaps a couple of comments would help here?

derekbruening: Added an explanation of the "3".  Removed the NULL_TERMINATE and added a comment; as you said with the up-front overflow check it is not needed.

derekbruening:
AArch64 this time hit #2892 and #3980.  Probably someone should try again to reproduce, and/or mark flaky.

Travis: still plagued by #4126, and also the clang job failed trying to install clang-9, which always used to work, and in fact worked on just the prior commit in this PR.  Grrr.

PR #4135, opened 2020-02-25
i#2666: decrease MAX_CONTEXT_SIZE for 32-bit to actual value

lieser:
@derekbruening Sorry if I did not make it clear in #4122 that we only looked at the `MAX_CONTEXT_SIZE` for 64-bit, and for simplicity just set the same for 32-bit.

Was able to debug it again today for 32-bit, and the value is indeed lower than for 64-bit.

Issue: #2666


derekbruening:
Travis is red still b/c of the fork failure: #4126

Issue #4134, opened 2020-02-25
DrBBDup: The DynamoRIO Basic Block Duplication Extension


johnfxgalea:
This is the main issue page concerning the DynamoRIO Basic Block Duplication Extension: a code builder that duplicates code of basic blocks and dispatches control according to runtime conditions so that different instrumentation of the same basic blocks may be efficiently executed.

During the app2app stage, the code of a basic block is duplicated and prepended to the original fragment in order to generate multiple copies. DrBBDup manages these copies by maintaining book-keep data per fragment and inserting dispatching code to execute the appropriate basic block copy.

For example, lets say we have the following bb:

```
mov ebx ecx
mov esi eax
ret
```

At a very high-level, DrBBDup produces code as follows

```
cmp [eax] 0x00000000
jz LABEL 2
  
LABEL 1 // BB Version 1
_INSTRUM CODE A_
mov ebx ecx
_INSTRUM CODE A_
mov esi eax
jmp EXIT LABEL

LABEL 2 // BB Version 2
_INSTRUM CODE B_
mov ebx ecx
_INSTRUM CODE B_
mov esi eax

EXIT LABEL
ret /* Not duplicated to abide by DR bb rules */
```

The number of basic block versions is customisable to suit the user's needs. In the example the dup count is 1 because we produce only one additional copy.

Apart from the main functionality of DrBBDup, this issue also encompasses the following:
1.  Dynamic generation of basic block copies (contingent on #3778 being merged).
2. Gathering of general statistics, such as tracking which cases are executed most frequently.
3. Support for Arm.
4. Include sample program.

There are a number of applications which I built using the extension. For instance, with two bb copies, one can turn on/off instruction tracing via nudges during runtime. One basic block copy is instrumented with tracing code, and another version with no instrumentation at all. Depending on a flag, the user can control which basic block copy is executed at runtime, without the need to undergo expensive flushing.

derekbruening:
This should be useful for #3995.  Currently for a single swap mid-run, a full flush is used, but the request is to support many swaps.  Presumably multi-version instru will be the better approach there.  I was going to implement it inside the tracer, but having a library sounds nicer.

johnfxgalea:
Updated description so that issue also concerns the inclusion of a simple use-case of drbbdup in the sample directory.

johnfxgalea:
Make use of  XINST_CREATE in order to prepare for ARM support.

[...]

PR #4150, opened 2020-03-03
i#4083 func trace: Add func_view tool

derekbruening:
Adds a new func_view tool for visualizing and analyzing function calls
recorded in a drcachesim trace.  It does two things in its initial
form: it reports the summary count of calls and returns for every
function traced, and if a new option op_show_func_trace is set (on by
default), it prints a linear sequence of every call and return with
return addresses, argument values, and return values.  Each is
indented to enable a clear picture of the call sequence.

Adds a new option op_funclist_file to specify the file's path, though
normally it will be auto-found.

Adds raw2trace_directory_t support for parsing the funclist file.

Adds example output for the new tool to the docs.  Adds a brief
section to the docs on function tracing in general, which was
previously missing.

Adds a func_view test which runs fib(5) and ensures we get the nested
function trace correct.

Fixes #4083

Issue #4022, opened 2020-01-16
Mac Travis fails but the failures are ignored and the wrapper reports success


derekbruening:
Travis Mac keeps failing due to #3127, yet the test failures are ignored somehow:
https://travis-ci.com/DynamoRIO/dynamorio/jobs/276424785

This issue covers figuring out why and fixing the runsuite scripts.

Issue #3127, opened 2018-08-15
ASSERT core/unix/signal.c:4690 (YMM_ENABLED() && ucxt->uc_mcsize...)


derekbruening:
Quite recently we've seen this new assert appear on Travis Mac (with no
email notification: grrrr #3031):

```
23/51 Test  #84: code_api|client.crashmsg .....................***Failed  Required regular expression not found.Regex=[^<Application .*simple_app.*.  My Fancy Tool internal crash at PC 0x.*.  Please report this at http:\/\/myfancytool.com/bugtracker.  Program aborted.
.*
$
]  0.11 sec
<Application /Users/travis/build/DynamoRIO/dynamorio/build_debug-internal-32/suite/tests/bin/simple_app (10655).  Internal Error: DynamoRIO debug check failure: /Users/travis/build/DynamoRIO/dynamorio/core/unix/signal.c:4690 (YMM_ENABLED() && ucxt->uc_mcsize == sizeof(_STRUCT_MCONTEXT_AVX32)) || (!YMM_ENABLED() && ucxt->uc_mcsize == sizeof(_STRUCT_MCONTEXT))
(Error occurred @0 frags)
version 7.0.17744, custom build
-no_dynamic_options -client_lib '/Users/travis/build/DynamoRIO/dynamorio/build_debug-internal-32/suite/tests/bin/libclient.crashmsg.dll.dylib;0;' -code_api -stderr_mask 12 -stack_size 56K -signal_stack_size 32K -max_elide_jmp 0 -max_elide_call 0 -no_inline_ignored_syscalls -native_exec_default_list '' -no_native_exec_man
0x4cfcc398 0x0035109a
0x4cfcc618 0x006b36e1
0x4cfccc48 0x0065f0f2
0x4cfb0fb8 0x00526acf
0x4cfb10a8 0x00622ed4
0x4cfb12f8 0x005f0e71
0x4cfb1c78 0x005f9e00
0x4cfb1e98 0x003047f3
0x4cfb1fe8 0x0065ee6e
0xbffe9b08 0x001a16e6
0xbffe9b98 0x0065ef94
0xbffe9d28 0x000fc538
0xbffe9d58 0x000e621a
0xbffe9d88 0x000fc4c5
0xbffe9df8 0x000fc8e1>
```

It's happening in a lot of tests:
```
debug-internal-32: 40 tests passed, **** 11 tests failed, but ignoring 0 for i#2941: ****
	code_api|common.decode-bad 
[...]

PR #4154, opened 2020-03-04
i4134: Pass fragment tag to drbbdup instrum routines

johnfxgalea:
Arranges instrumentation call-back functions of drbbdup to accept the passing of tag values.

Issue: #4134

### DIFF
--- a/ext/drbbdup/drbbdup.c
+++ b/ext/drbbdup/drbbdup.c
@@ -497,8 +497,8 @@ drbbdup_extract_single_bb_copy(void *drcontext, instrlist_t *bb, instr_t *start,
  * that transcends over different cases.
  */
 static void *
-drbbdup_do_orig_analysis(drbbdup_manager_t *manager, void *drcontext, instrlist_t *bb,
-                         instr_t *start)
+drbbdup_do_orig_analysis(drbbdup_manager_t *manager, void *drcontext, void *tag,
+                         instrlist_t *bb, instr_t *start)
 {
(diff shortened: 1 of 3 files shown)


johnfxgalea:
Think about it now, I should also pass for_trace but I'll draw up another PR for that.

johnfxgalea:
@derekbruening Something weird happened on Mac. Was this the failure which you couldn't reproduce last month? I couldn't find an entry on the issues page.

derekbruening:
I remember #3127 and #4022 -- is that what you mean?

johnfxgalea:
run arm tests

johnfxgalea:
Issues a new page for the MAC OS failure: #4156

derekbruening:
I am seeing strange Mac failures in #4150 but it looks different: the log ends early as though it's a timeout but it doesn't report a timeout.  A re-run went green on one instance of this.

johnfxgalea:
> I am seeing strange Mac failures in #4150 but it looks different: the log ends early as though it's a timeout but it doesn't report a timeout. A re-run went green on one instance of this.

hmm, they both seem to be trigger similar failures: vmareas.c:1511 start <= end || end == NULL
Could it be the same root-cause? hmm

johnfxgalea:
run arm tests

johnfxgalea:
Thank you for the review!

johnfxgalea:
Going ahead with the merge. Test failure is due to #4156

Issue #4160, opened 2020-03-05
Add parentheses to the NULL_TERMINATE macro found in many samples


johnfxgalea:
Put parentheses around the parameter concerning the NULL_TERMINATE macro. This macro is found in many samples.

Instead of:
`#define NULL_TERMINATE(buf) buf[(sizeof(buf) / sizeof(buf[0])) - 1] = '\0'`

we need to change to macro to
`#define NULL_TERMINATE(buf) (buf)[(sizeof((buf)) / sizeof((buf)[0])) - 1] = '\0'`

johnfxgalea:
This was first brought up in this PR #4134

derekbruening:
@johnfxgalea -- it looks like your email address in these commits is `@users.noreply.github.com`.  I think maybe you forgot to set it in the git config from the devsetup.sh script.

derekbruening:
OK.  FTR the wiki officially says:
 https://github.com/DynamoRIO/dynamorio/wiki/Workflow#configuring-author-information-and-aliases says "You must also uncheck `Keep my email address private` in the email settings page."

That was to mirror various old-school open-source projects where presumably the idea is to make sure you can contact an author about a particular code patch.  Our project hosted on Github with its separate metadata on which Github user did which commit I suppose makes that not matter as much so it seems reasonable to drop that.

PR #4161, opened 2020-03-05
i#4160: Adds parens to NULL_TERMINATE Macro

johnfxgalea:
Adds parentheses to the param of the NULL_TERMINATE macro found in many samples.

Fixes: #4160

### DIFF
--- a/api/samples/bbcount.c
+++ b/api/samples/bbcount.c
@@ -51,7 +51,7 @@
 #    define DISPLAY_STRING(msg) dr_printf("%s\n", msg);
 #endif
 
-#define NULL_TERMINATE(buf) buf[(sizeof(buf) / sizeof(buf[0])) - 1] = '\0'
+#define NULL_TERMINATE(buf) (buf)[(sizeof((buf)) / sizeof((buf)[0])) - 1] = '\0'
 
 #define TESTALL(mask, var) (((mask) & (var)) == (mask))
 #define TESTANY(mask, var) (((mask) & (var)) != 0)
--- a/api/samples/countcalls.c
+++ b/api/samples/countcalls.c
@@ -51,7 +51,7 @@
 #    define DISPLAY_STRING(msg) dr_printf("%s\n", msg);
 #endif
 
-#define NULL_TERMINATE(buf) buf[(sizeof(buf) / sizeof(buf[0])) - 1] = '\0'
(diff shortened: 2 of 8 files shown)


api/samples/bbcount.c
```diff
@@ -51,7 +51,7 @@
 #    define DISPLAY_STRING(msg) dr_printf("%s\n", msg);
```
derekbruening: In the description:
> Fixes: #4160

I don't know that this will auto-close: the Github docs say to not have a colon and that is what our guide says and all our commits.  (We have a colon on `Issue: #xxx, #yyy` which is for our own parsing.)

api/samples/countcalls.c
```diff
@@ -51,7 +51,7 @@
 #    define DISPLAY_STRING(msg) dr_printf("%s\n", msg);
```
derekbruening: I wonder if we should put all these macros in a shared header or sthg.  OTOH it is nice to have simpler, self-contained samples so maybe just leave it.

johnfxgalea: Do you want DISPLAY_STRING's param to also be enclosed in parens?

api/samples/bbcount.c
johnfxgalea: Do you want to give it a try? or shall we stick to always having no colon for Fixes?

api/samples/countcalls.c
derekbruening: No, I think it's safe in that position.

api/samples/bbcount.c
derekbruening: https://help.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords

Hmm it actually does imply a colon will work in its example.  I wonder if that's new.  I'd be fine with switching to having a colon (do you want to update the wiki?) since it matches typical tags used in commits like `Co-authored-by:` or `Signed-off-by:` and our own `Issue:`.

johnfxgalea: Let's see if it works :p

johnfxgalea: KK, seemed to do the trick. I'll mod the wiki soon.

johnfxgalea: Done.

Issue #4152, opened 2020-03-03
dr_get_proc_address incorrectly returns base-of-library results for undefined symbols with gnu hashes


derekbruening:
I saw this on drcachesim's -record_heap running "ls" where it claims to
find "malloc" and "free" in the ls binary but returns an address at the
very base of the binary:
```
instru_funcs_module_load for ls
dr_get_proc_address found symbol malloc at pc=0x00007f0cfda4c000
Inserted hooks for ls!malloc @0x00007f0cfda4c000 == id 0
dr_get_proc_address found symbol free at pc=0x00007f0cfda4c000
Duplicate-pc hook: ls!free == id 0
```

Looks like DR is not ruling out undefined symbols:
```
$ readelf -s -D /bin/ls 

Symbol table of `.gnu.hash' for image:
  Num Buc:    Value          Size   Type   Bind Vis      Ndx Name
  111   0: 0000000000022600     8 OBJECT  GLOBAL DEFAULT  26 __progname
  112   2: 0000000000022610     4 OBJECT  GLOBAL DEFAULT  26 optind
  113   3: 0000000000000000     0 FUNC    GLOBAL DEFAULT UND free
  114   3: 0000000000022620     8 OBJECT  WEAK   DEFAULT  26 program_invocation_name
  115   4: 0000000000022620     8 OBJECT  GLOBAL DEFAULT  26 __progname_full
  116   5: 00000000000163a0    41 FUNC    GLOBAL DEFAULT  14 _obstack_memory_used
  117   6: 00000000000225e0     8 OBJECT  GLOBAL DEFAULT  25 obstack_alloc_failed_hand
  118   6: 00000000000161b0    17 FUNC    GLOBAL DEFAULT  14 _obstack_begin
  119   7: 0000000000022640     8 OBJECT  GLOBAL DEFAULT  26 stderr
  120   8: 0000000000016330   101 FUNC    GLOBAL DEFAULT  14 _obstack_free
  121   8: 0000000000022600     8 OBJECT  WEAK   DEFAULT  26 program_invocation_short_
  122  10: 00000000000162f0    55 FUNC    GLOBAL DEFAULT  14 _obstack_allocated_p
  123  10: 0000000000022618     8 OBJECT  GLOBAL DEFAULT  26 optarg
  124  11: 00000000000161d0    21 FUNC    GLOBAL DEFAULT  14 _obstack_begin_1
[...]

Issue #4156, opened 2020-03-04
DynamoRIO debug check failure: vmareas.c:1507 start < end || end == NULL


johnfxgalea:
We encountered a debug failure specific to Mac OS.

```
Application /Users/travis/build/DynamoRIO/dynamorio/build_debug-internal-64/suite/tests/bin/common.broadfun spaces (33435).  Internal Error: DynamoRIO debug check failure: /Users/travis/build/DynamoRIO/dynamorio/core/vmareas.c:1507 start < end || end == NULL
(Error occurred @0 frags)
version 7.91.18305, custom build
-no_dynamic_options -code_api -stderr_mask 12 -stack_size 56K -signal_stack_size 32K -max_elide_jmp 0 -max_elide_call 0 -no_inline_ignored_syscalls -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct 
0x00007ffee9697e00 0x0000000106724537
0x00007ffee9698090 0x0000000106822571
0x00007ffee9698100 0x0000000106821e24
0x00007ffee9698130 0x0000000106821c38
0x00007ffee9698170 0x0000000106a4f573
0x00007ffee9698220 0x000000010686c619
0x00007ffee9698280 0x000000010686c2ba
0x00007ffee9698330 0x0000000106a58616
0x00007ffee9698380 0x0000000106a2b892
0x00007ffee9698480 0x0000000106a2bb47
0x00007ffee9698580 0x000000010681b17f
0x00007ffee9698aa0 0x00000001065ab329
0x00007ffee96993e0 0x000000010656a897
0x00007ffee9699420 0x000000010bce08a5
0x00007ffee9699490 0x000000010bce0ceb>
```

This results in a significant amount of test failures:

```
====> FAILURE in debug-internal-64 <====
debug-internal-64: 7 tests passed, **** 38 tests failed, but ignoring 0 for i: ****
	code_api|common.broadfun 
	code_api|common.fib 
	code_api|common.floatpc 
	code_api|common.floatpc_xl8all 
	code_api|common.getretaddr 
	code_api|linux.exit 
	code_api|linux.infinite 
	code_api|linux.signal0000 
	code_api|linux.signal0010 
	code_api|linux.signal0100 
	code_api|linux.signal0110 
	code_api|linux.sigaction 
	code_api|linux.app_tls 
	code_api|linux.fib-conflict 
	code_api|linux.fib-pie 
	code_api|security-common.codemod 
[...]

PR #4162, opened 2020-03-05
i#4152: Exclude undefined symbols in gnu_hash_lookup

derekbruening:
While elf_hash_lookup correctly skips undefined symbols,
gnu_hash_lookup was not, causing it to claim to find symbols with
addresses at the library base.

Tested on drcachesim -record_heap running /bin/ls where before the fix
it claimed "malloc" was at the module base.

Fixes: #4152


derekbruening:
Travis failure is the Mac #4156 which needs to be addressed but it is not clear the best route forward.

Issue #3048, opened 2018-06-15
Record key function traces in drcachesim


Louis-Ye:
Add mechanism to collect traces for specified/key functions in drcachesim. The traces should integrate with memory accesses traces.
The new traces should include following information for each function:
1. function invocation arguments
2. function return address
3. function return value

derekbruening:
Pasting in some design comments from #3057:

I think we want to be able to dynamically specify which functions to record: perhaps via a config file eventually, though we could start with a runtime option. We just need the name and number of args if all we support is recording pointer-sized arg values. You could imagine sthg like:

```
-record_function malloc 1 -record_function realloc 2
```

If we decide heap allocation is very common and we want to have built-in support, I would say it should be under a separate option like -record_function_malloc or -record_heap_allocs or sthg.

We need to get a function ID of some kind in there for the markers to point at: with the options it's not clear the best way.  Have the user specify and complain on duplicates??  So there's an ID and number of args?

```
-record_function malloc 42 1
```

Alternatively, we could expose an interface to write metadata into the trace and structure all of this as a separate client from the memory address tracer.

derekbruening:
We may want the callstack rather than just the return address, to handle functions whose immediate caller may not be useful if it is often invoked through utility function layers.

PR #4176, opened 2020-03-11
i#3048 func-trace: Add "noret" function tracing

derekbruening:
Adds an optional specifier for drcachesim function tracing where
"|noret" on a trace request indicates to not record the return value.
This can reduce overhead as finding all post-call return points is
non-trivial.

Changes the funclist.log format to include multiple comma-separated
values prior to the symbol: the id, number of args, PC, and optional
flags like "noret".  Adding the arg count helps simplify func_view
code, and it and the PC should be generally useful, although they can
technically be extracted from the trace.

Adds support to the func_view tool to keep the indentation level the
same for a noret function, to properly visualize nesting of other
functions.  Adds thread prefixes to distinguish different thread
subtraces when viewing an interleaving.

Renames DRMEMTRACE_FUNCTION_MAP_FILENAME to
DRMEMTRACE_FUNCTION_LIST_FILENAME to match all other references which
call it the "function list" or "funclist".

Adds support for locating the auxiliary files in either the raw/ or
trace/ subdirectories to make it easier to persist trace files.

Adds a new test with a combination of return-value and noret
functions.  I wanted to test threads as well, and it is difficult to
make a deterministic test like that, so I checked in the trace files
themselves.

Issue: #3048


derekbruening:
Maybe the thread prefix and test, and the file define rename, could be separated out: but those parts are small and it did not seem worthwhile.

PR #4178, opened 2020-03-12
i#3048 func-trace: Add new+delete to -record_heap

derekbruening:
Adds all 24 global operator new and delete variants to the default
-record_heap_value, since in some implementations they do not simply
call malloc and free.  We include the Itanium ABI manglings only here.
MSVC and Windows heap interception is even more complex with more
layers and we leave that for future work.

Adds a new test which calls each one, though some are only compiled
and called if C++17 is available in the compiler.  It is not easy to
check the output so for now we do not have good automated validation
that all operators were caught, only manual checking.

Issue: #3048

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -695,6 +695,7 @@ endmacro()
 set(BASE_CFLAGS "")
 set(BASE_CONLY_FLAGS "")
 set(BASE_CXXONLY_FLAGS "")
+CHECK_CXX_COMPILER_FLAG("-std=c++17" cxx17_available)
 if (UNIX)
   set(BASE_CXXONLY_FLAGS "${BASE_CXXONLY_FLAGS} -std=c++11")
   # -std=c99 doesn't quite work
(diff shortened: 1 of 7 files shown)


derekbruening:
So on Mac we have:
```
	/.../dynamorio/clients/drcachesim/tests/heap_test.cpp:73:31: error: aligned allocation function of type 'void *(std::size_t, std::align_val_t)' is only available on macOS 10.13 or newer [-Waligned-allocation-unavailable]
```

The Travis docs claim that our `xcode11.3` setting is 10.14!  DR's `uname` command had it as kernel 17.7.0 which should be 10.13.  Either way the warning about being < 10.13 makes no sense.  But even worse is that I guarded this code with `#if defined(__cpp_aligned_new)`.  What's the point of these macros if they aren't even set properly in Apple's clang?!?  Grrr.

derekbruening:
We also have this in our clang-9 Linux job:
```
/home/travis/build/DynamoRIO/dynamorio/clients/drcachesim/tests/heap_test.cpp:74:38: error: no member named 'align_val_t' in namespace 'std'
        int *aligned_int = new (std::align_val_t{ 64 }) int;
```

I have clang-9 on a machine here and I cannot reproduce this: it builds fine.  And once again the code is guarded by `#if defined(__cpp_aligned_new)`.  What is wrong with clang?!?

derekbruening:
Presumably these are due to compiler vs runtime issues.  Rather than struggle with how clang is managing compiler vs runtime I am probably just going to disable this one test when clang is the compiler and then in a year when the runtimes are newer we can probably re-enable it.

Issue #4180, opened 2020-03-14
Enable heap-less IR data structures


derekbruening:
DR's IR data structures were designed to have a small size footprint in the average case for DR's  mode of controlling an application without any tool attached.  In that mode, DR does not bother to fully decode all operands of most instructions, caring only about branches and other instructions that need mangling.  A key use case of DR early on was running dozens of 32-bit applications simultaneously, where some of these apps were quite large and took the entire address space, leaving DR to operate within a bounded reservation.  Thus, `instr_t` was designed to only have room for one source built-in (all you need for most branches).

However, today's typical use case is with a tool attached, where DR does decode all the operands for every single instruction.  Plus, most uses are 64-bit, where space is not as critical, and running a single application at a time.  Even more importantly, for the `drdecode` decoding/encoding library that DR exports, it is best to not use *any* heap to allow use within signal handlers and other fragile locations.

This is a feature request to either change the default `instr_t` layout or have a method of creating an `instr_t` that does not use any heap.

derekbruening:
My proposal is to keep the IR for regular DR use the same.
We make an `instr_noalloc_t` struct which is `instr_t` plus max srcs and dsts plus encoding buffer:
```C
typedef struct instr_noalloc_t {
    instr_t instr;
    opnd_t srcs[MAX_SRCS-1];
    opnd_t dsts[MAX_DSTS];
    byte encode_buf[MAX_INSTR_LENGTH];
} instr_noalloc_t;
```

Do the C-style "subclass": add an `instr_t` flag saying whether it's an `instr_noalloc_t`.
For the flag: remove `INSTR_HAS_CUSTOM_STUB` (part if #4045 cleanup).

To use:
```
  instr_noalloc_t inst;
  instr_noalloc_init(dcontext, &inst);
```

One issue is:
```C
private_instr_encode(dcontext_t *dcontext, instr_t *instr, bool always_cache)
{
    /* We cannot use a stack buffer for encoding since our stack on x64 linux
     * can be too far to reach from our heap.  We need reachable heap.
     */
    byte *buf = heap_reachable_alloc(dcontext, MAX_INSTR_LENGTH HEAPACCT(ACCT_IR));
...
        /* if unreachable, we can't cache, since re-relativization won't work */
```

So we can use the stack buffer inside instr_noalloc_t but it won't be kept.  That is one reason to make this no-alloc struct a feature you have to explicitly choose to use, as it is best for decoding only rather than encoding.

[...]

Issue #4045, opened 2020-01-22
[cleanup]: remove unsupported API's and options: branch prefixes; custom exit stubs; trace head cache incr


derekbruening:
To improve readability and simplify the code we should remove unsupported API's and options, such as:

+ dr_add_prefixes_to_basic_blocks()
+ TRACE_HEAD_CACHE_INCR
+ CUSTOM_EXIT_STUBS

derekbruening:
CUSTOM_TRACES_RET_REMOVAL would be included here

Issue #4058, opened 2020-01-27
Multiple tests failing on Appveyor VS2017 Win10 1607


derekbruening:
#2924 put a lot of work into getting DR switched over to VS2017.  Despite all the work there to get the build and tests working, there are still some tests which are failing and which I separated out here in order to finish off the base switch.  This issue covers fixing these issues one by one.

derekbruening:
More failures not on the original ignore list:
https://ci.appveyor.com/project/DynamoRIO/dynamorio/builds/30390790
```
14464debug-internal-32: 258 tests passed, **** 4 tests failed: ****
14465	code_api|win32.tls 
14466	code_api|security-win32.except-execution 
14467	code_api|client.pcache-use 
14468	code_api|tool.drcacheoff.view 
```

derekbruening:
Another:
https://ci.appveyor.com/project/DynamoRIO/dynamorio/builds/30417232
```
26916debug-internal-32: 258 tests passed, **** 4 tests failed, but ignoring 3 for i#2145: ****
26917	code_api|win32.callback 
26918	(ignore: i#2145) 	code_api|client.pcache-use 
26919	(ignore: i#2145) 	code_api|tool.drcacheoff.burst_replaceall 
26920	(ignore: i#2145) 	code_api|tool.drcacheoff.view 
```

johnfxgalea:
Another (_drcacheoff.burst_replace_): 
https://ci.appveyor.com/project/DynamoRIO/dynamorio/builds/30490048#L27116

```
====> FAILURE in debug-internal-32 <====
27117debug-internal-32: 257 tests passed, **** 5 tests failed, of which 1 were flaky, but ignoring 3 for i#2145: ****
27118	(ignore: i#2145) 	code_api|win32.reload-newaddr 
27119	code_api|security-win32.codemod-threads_FLAKY 
27120	(ignore: i#2145) 	code_api|client.pcache-use 
27121	code_api|tool.drcacheoff.burst_replace 
27122	(ignore: i#2145) 	code_api|tool.drcacheoff.view 
27123debug-internal-64: 227 tests passed, **** 7 tests failed, but ignoring 7 for i#2145: ****
27124	(ignore: i#2145) 	code_api|common.decode 
27125
```

derekbruening:
Another 32-bit failure:
https://ci.appveyor.com/project/DynamoRIO/dynamorio/builds/30511381
```
[...]

PR #4181, opened 2020-03-14
i#4045: Remove custom exit stub support

derekbruening:
Removes the no-longer-supported custom exit stub feature: all of the
code under the CUSTOM_EXIT_STUBS define and the INSTR_HAS_CUSTOM_STUB
flag.  Removing the flag frees it up for use in #4180.

Issue: #4045, #4180


derekbruening:
Appveyor is #4058.

Clang-format failure is what it should look like locally but on the whole file it does what I have.  There is actually a double paren messing up the format.  I will check this in as-is and separately fix that.

Issue #2876, opened 2018-03-07
Adopt clang-format for all files?


derekbruening:
As part of adding lint style checks in #2369 we considered using clang-format but balked at reformatting every single file and thus adding a history hurdle.  But as more contributions come in with style issues and as I see other projects successfully use clang-format to simplify things, I lean more and more toward biting the bullet and reformatting everything and then adding a pre-commit check.

Carrotman42:
Is it possible to just clang-format new lines?

derekbruening:
> Is it possible to just clang-format new lines?

Possibly, but we are likely to have to tweak the existing style rules (because there may not be existing clang-format knobs to match everything), and we already have some older code violating things, and it's more complicated to set up editors and scripts to only format changes -- it seems simpler to get entire files in a consistent state.  Xref other project experiences such as https://engineering.mongodb.com/post/succeeding-with-clangformat-part-2-the-big-reformat/

fhahn:
Maybe we could amke use of the clang-format-diff.py script: https://github.com/llvm-mirror/clang/blob/master/tools/clang-format/clang-format-diff.py

It just formats the changed lines in a diff

derekbruening:
FTR here are the settings I used a year ago though I did not finish tweaking them:

```
$ /work/llvm/build_rel/bin/clang-format -style=WebKit -dump-config > /tmp/clang-format-webkit
$ diff /tmp/clang-format-webkit .clang-format
5c5
< AlignAfterOpenBracket: DontAlign
---
> AlignAfterOpenBracket: Align
14c14
< AllowShortFunctionsOnASingleLine: All
---
> AllowShortFunctionsOnASingleLine: None
17,18c17,18
< AlwaysBreakAfterDefinitionReturnType: None
< AlwaysBreakAfterReturnType: None
---
> AlwaysBreakAfterDefinitionReturnType: All
> AlwaysBreakAfterReturnType: AllDefinitions
41c41
< ColumnLimit:     0
---
> ColumnLimit:     90
78c78
< PointerAlignment: Left
---
> PointerAlignment: Right
88c88
< SpacesInContainerLiterals: true
---
> SpacesInContainerLiterals: false
92c92
< Standard:        Cpp11
---
> Standard:        Cpp03
```

Cpp11 should be fine now.

[...]

PR #4182, opened 2020-03-14
i#2876 clang-format: Fix format of optionsx.h

derekbruening:
There was a double parenthesis causing clang-format to fail to format
core/optionsx.h.  We fix that here.

Issue: #2876


derekbruening:
A64 is histogram.offline #3980 but that's listed as a flaky test to ignore: so why didn't the wrapper script ignore it?  Confused.

PR #4183, opened 2020-03-16
i#4180: Add non-heap-using instuction structure

### LINKED ISSUE #4180: Enable heap-less IR data structures

derekbruening:
Adds instr_noalloc_t, a wrapper around instr_t which contains space
for all possible operands and a temporary encoding buffer.  Adds
support in the core instruction routines for using these internal
fields and never using external heap, enabling this structure to be
used for decoding in a signal handler.

The encoding buffer, due to reachability issues, cannot be
re-relativized, making this structure inefficient for encoding.  It is
targeted toward decoding only.

Adds a sanity test.

Fixes #4180


derekbruening:
I'm seeing a local Doxygen build failure I'm not understanding and am trying to figure out (may not show up on Travis: it built on the A64 machine; depends on doxygen version I guess), but I'd like feedback on this interface so I'm sending this for review now.

suite/tests/api/drdecode_aarch64.c
```diff
@@ -49,11 +57,46 @@ test_disasm(void)
...
+                              OPND_CREATE_MEMPTR(DR_REG_X1, 0));
+    end = instr_encode(GD, instr, buf);
+    ASSERT(end - buf < BUFFER_SIZE_ELEMENTS(buf));
+    instr_destroy(GD, instr);
+
+    instr_noalloc_t noalloc;
+    instr_noalloc_init(GD, &noalloc);
+    pc = decode(GD, buf, &noalloc.instr);
+    ASSERT(pc != NULL);
+    ASSERT(opnd_get_reg(instr_get_dst(&noalloc, 0)) == DR_REG_X0);
```
johnfxgalea: I think the interchanging use of  &noalloc.instr and  &noalloc is going to confuse the user.

core/arch/instr.h
```diff
@@ -498,6 +498,37 @@ struct _instr_t {
...
+ * decoding and encoding.  It inlines all the possible operands and encoding space
+ * inside the structure.  Some operations could still use heap if custom label data is
+ * used to point at heap-allocated structures through extension libraries or custom
+ * code.
+ *
+ * The 'instr' field should be passed to decoding and other routines.  E.g.:
+ *
+ *    instr_noalloc_t noalloc;
+ *    instr_noalloc_init(dcontext, &noalloc);
+ *    pc = decode(dcontext, ptr, &noalloc.instr);
```
johnfxgalea: Do we instead want to make instr_noalloc_t opaque from the user's perspective and create a new API function to obtain the instr field?

core/arch/instr_shared.c
```diff
@@ -503,16 +536,26 @@ instr_set_num_opnds(dcontext_t *dcontext, instr_t *instr, int instr_num_dsts,
         CLIENT_ASSERT_TRUNCATE(instr->num_dsts, byte, instr_num_dsts,
                                "instr_set_num_opnds: too many dsts");
         instr->num_dsts = (byte)instr_num_dsts;
-        instr->dsts = (opnd_t *)heap_alloc(
-            dcontext, instr_num_dsts * sizeof(opnd_t) HEAPACCT(ACCT_IR));
+        if (TEST(INSTR_IS_NOALLOC_STRUCT, instr->flags)) {
+            instr_noalloc_t *noalloc = (instr_noalloc_t *)instr;
```
johnfxgalea: If find this cast and others messy tbh. I thought instr_noalloc_t was enclosing instr_t and cannot be used vice-versa as well. This makes the code hard to read overall. Of course, this is just my personal opinion, and may be wrong here.

What I am thinking is to create separate API entries for stack/heap allocated instr and have these functions internally make use of the same code. I know this can be a bit of a pain but given the fact that we are avoiding unions, I think this approach can help mitigate type confusion and improve clarity.

johnfxgalea:
I'll give it another pass tomorrow and try it on my machine.

suite/tests/api/drdecode_aarch64.c
johnfxgalea: See https://github.com/DynamoRIO/dynamorio/pull/4183#discussion_r392777844

core/arch/instr_shared.c
derekbruening: This is essentially a sub-class: extending the instr_t struct with additional fields and passing a pointer to the base class around.

core/arch/instr.h
derekbruening: Yes that does seem nicer.

core/arch/instr_shared.c
johnfxgalea: Yes, I understood what you did here. Perhaps, if worded differently, why didn't you modify decode() to also check the instr flag for instr_noalloc_t? Requiring the user to fetch the base in your test case implies non-interchangeability which contrasts with the code above.

derekbruening: With no compiler support for overloads, we can't modify decode() to solve this: the compiler complains about the type passed in.  We'd have to make a separate function decode_noalloc().  We'd have to make a hundred separate functions.  IMHO it is a nicer interface to instead pass in instr_from_noalloc(&noalloc) (proposed name for your suggested helper to return &noalloc.instr) to the existing functions, with a little hackiness hidden in the implementation for the poor man's subclass.

johnfxgalea: Thanks, I understand the the overkill and the fact that we r trying to minimise changes. I got the PR to run on my machine. Actually, what I meant, is the passing of &noalloc to instr_get_dst(), which according to the docs accepts instr_t but the check is circumvented due to it being a macro. Perhaps modify the docs for those functions?

johnfxgalea:
As you mentioned above, the following is the doxygen error:

```
CMake Error at /home/john/main_dr/dynamorio/api/docs/CMake_rundoxygen.cmake:143 (message):
  *** /usr/bin/doxygen failed: ***

  <unknown>:1: warning: explicit link request to 'instr_t' could not be
  resolved

  /home/john/main_dr/dynamorio/build/include/dr_ir_utils.h:353: warning:
  explicit link request to 'opnd_t' could not be resolved

  /home/john/main_dr/dynamorio/api/docs/release.dox:140: warning: explicit
  link request to 'instr_t' could not be resolved

  /home/john/main_dr/dynamorio/ext/drreg/drreg.h:496: warning: explicit link
  request to 'opnd_t' could not be resolved

  /home/john/main_dr/dynamorio/ext/drreg/drreg.h:537: warning: explicit link
  request to 'opnd_t' could not be resolved

  /home/john/main_dr/dynamorio/ext/drreg/drreg.h:540: warning: explicit link
  request to 'opnd_t' could not be resolved

  /home/john/main_dr/dynamorio/build/include/dr_ir_instr.h:325: warning:
  explicit link request to 'instr_t' could not be resolved
```

core/arch/instr_shared.c
derekbruening: Interesting, I just used it without thinking and the compiler did not complain.  You mean this:
```C
    ASSERT(opnd_get_reg(instr_get_dst(&noalloc, 0)) == DR_REG_X0);
```

I will change it `s/&noalloc/instr_from_noalloc(&noalloc)/`.  I guess that means the suggested use is:

```C
instr_noalloc_t noalloc;
instr_noalloc_init(dcontext, &noalloc);
instr_t *instr = instr_from_noalloc(&noalloc);
<now use instr everywhere; no need to free at the end, but still do a reset to re-init if doing repeated decode into it or sthg>
```

WDYT?

derekbruening:
> As you mentioned above, the following is the doxygen error:
> 
> ```
>   <unknown>:1: warning: explicit link request to 'instr_t' could not be
>   resolved
...
> ```

Right.  I assume it's from the `DR_FAST_IR` tweaking but I have not yet figured out exactly why or how to fix; already tried various pre-defines for doxygen.  Some extra verbosity would help if it supports it.

core/arch/instr_shared.c
johnfxgalea: Yea, much better - essentially avoid interchanging between instr and instr_noalloc at the API level or any other types for that matter because that's when type confusion bugs start to crop up.

suite/tests/api/drdecode_aarch64.c
derekbruening: Resolved with the use of a separate instr_t var per the discussion below.

core/arch/instr.h
derekbruening: Done.

core/arch/instr_shared.c
derekbruening: Implemented.

derekbruening:
PTAL.  The separate `instr_t` scheme is implemented and doxygen is fixed.

core/arch/instr.h
```diff
@@ -186,7 +186,7 @@ enum {
     /* DR_API EXPORT BEGIN */
     INSTR_DO_NOT_MANGLE = 0x00200000,
     /* DR_API EXPORT END */
-    /* Available: 0x00400000, */
+    INSTR_IS_NOALLOC_STRUCT = 0x00400000,
```
johnfxgalea: Nit: Add a comment here that this flag must only be set by instr_noalloc_init. I'm a bit worried that the "subclass"-like extension will mistakenly be not present if someone misuses this flag. You limited many cases, e.g., when cloning but optionally adding a comment so that everyone remember might help.

core/lib/globals_shared.h
```diff
@@ -1246,8 +1251,9 @@ typedef char liststring_t[MAX_LIST_OPTION_LENGTH];
  * share/config.c) to set up new eventlogs (mainly for vista where our
  * installer doesn't work yet xref case 8482).*/
 #    define L_EVENT_FILE_VALUE_NAME L"File"
-#    define L_EVENT_FILE_NAME_PRE_VISTA \
-        L"%SystemRoot%\\system32\\config\\" L_EXPAND_LEVEL(EVENTLOG_NAME) L".evt"
+#    define L_EVENT_FILE_NAME_PRE_VISTA                                          \
```
johnfxgalea: Not a review comment, but just curious what are these defines for?

derekbruening: The changes are clang-format doing sthg different: not sure why.  For Windows notifications are sent to the event log (a Windows OS feature), per `-syslog_mask`, and these control aspects of that.

core/arch/instr.h
derekbruening: +1

derekbruening: Done.

derekbruening:
A64 is a test that should be auto-ignored: `code_api|tool.drcachesim.invariants`.  Looking more closely:
```
+ cp /home/assad/dynamorio/suite/runsuite_wrapper.pl suite/runsuite_wrapper.pl
```

So it seems to *not* be using the checked-in script.  @AssadHashmi could you look into that?

derekbruening:
Appveyor is #4058, a flaky test; it was just green.  A64 is #3980 which should be ignored: see comment above.

Issue #4070, opened 2020-01-31
Add drwrap call site wrapping feature


derekbruening:
drwrap by default handles all the complexities of tailcalls, longjmp, etc. by using a callee and return address wrapping approach.  This incurs extra overhead, mostly from flushing when a return address is discovered after it has already been executed (or simultaneously executed in another thread).  There are some use cases where missing corner cases of calling conventions is not important and we would prefer to have better performance: we know we're targeting functions that never do anything unusual.  This is a feature request to iadd a drwrap mode for this case: wrapping based solely on the call site and ignoring the callee and what actually happens when it returns.

derekbruening:
With the new flag we're considering this closed.  Future improvements of looking for PLT/IAT-style calls can be filed separately if someone wants to add those.

PR #4184, opened 2020-03-17
i#4070 drwrap-site: Remove unused call_site_table

derekbruening:
Cleanup: removes the unused call_site_table.

Issue: #4070

### DIFF
--- a/ext/drwrap/drwrap.c
+++ b/ext/drwrap/drwrap.c
@@ -244,14 +244,6 @@ fast_safe_read(void *base, size_t size, void *out_buf)
  * WRAPPING INSTRUMENTATION TRACKING
  */
 
-/* We need to know whether we've inserted instrumentation at the call site .
- * The separate post_call_table tells us whether we've set up the return site
- * for instrumentation.
- */
-#define CALL_SITE_TABLE_HASH_BITS 10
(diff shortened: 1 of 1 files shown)


ext/drwrap/drwrap.c
```diff
@@ -927,8 +927,6 @@ drwrap_init(void)
                       NULL);
     hashtable_init_ex(&wrap_table, WRAP_TABLE_HASH_BITS, HASH_INTPTR, false /*!str_dup*/,
                       false /*!synch*/, wrap_entry_free, NULL, NULL);
-    hashtable_init_ex(&call_site_table, CALL_SITE_TABLE_HASH_BITS, HASH_INTPTR,
```
s-kanev: Don't forget the definition too: [here](https://github.com/DynamoRIO/dynamorio/blob/53577d796030c79790ca354a7e7ab4536bccd492/ext/drwrap/drwrap.c#L253)

derekbruening: Wow that was removed in the original version of this.  After some git commit chain rearranging somehow it came back...a little disturbing.

derekbruening: Gone now.

PR #4185, opened 2020-03-17
i#4070 drwrap-site: Add drwrap_get_stats() with flush count

derekbruening:
Adds a new interface drwrap_get_stats() which returns the count of
flushes triggered by drwrap, for use when evaluating performance
overhead.  Adds documentation and a test.

Issue: #4070

### DIFF
--- a/ext/drwrap/drwrap.c
+++ b/ext/drwrap/drwrap.c
@@ -80,6 +80,12 @@ static uint verbose = 0;
 #    define CALL_POINT_SCRATCH_REG DR_REG_NULL
 #endif
 
+#ifdef X64
+#    define dr_atomic_add_stat_return_sum dr_atomic_add64_return_sum
+#else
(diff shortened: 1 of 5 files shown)


ext/drwrap/drwrap.c
```diff
@@ -2401,6 +2405,16 @@ drwrap_is_post_wrap(app_pc pc)
     return res;
 }
 
+DR_EXPORT
+bool
+drwrap_get_stats(INOUT drwrap_stats_t *stats)
+{
+    if (stats == NULL || stats->size != sizeof(*stats))
+        return false;
+    stats->flush_count = drwrap_stats.flush_count;
```
s-kanev: Ok to read without some sort of acquire load?

derekbruening: Actually thank you for raising this: for 32-bit ARM this does need something.  I will put in an atomic for all platforms for clarity.

derekbruening:
A64 is #4189

PR #4186, opened 2020-03-17
i#4070 drwrap-site: Add post-call sites proactively

derekbruening:
When a direct call to a drwrap target is found, we now proactively add
the post-call point to the to-be-instrumented table, potentially
avoiding a flush if that PC were to be reached through some other
method before we find it in our callee control point.

This also enables finding post-call points even if we limited dynamic
return address investigation, which does add overhead.  A future
change will explore this further.

Issue: #4070


derekbruening:
A64 failure is code_api|tool.histogram.offline which is supposed to be ignored as flaky.  #4189 covers fixing Jenkins which is not using the right script.

PR #4193, opened 2020-03-18
i#4070 drwrap-site: Add flag to skip retaddr examination

derekbruening:
Adds DRWRAP_NO_DYNAMIC_RETADDRS which instructs drwrap to not look for
post-call points by examining the return address and comparing to
known sites while in the callee.  This incurs overhead, mostly due to
the locks and shared data structures, in large many-threaded
applications.  The downside is that the only post-call points in place
are those for direct calls, which drwrap always looks for with the
recent PR #4186.  Future work could add identification of
PLT-or-IAT-style indirect calls, or explore more scalable data
structures.

Adds testing to drwrap-test.

Issue: #4070

### DIFF
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -248,6 +248,8 @@ Further non-compatibility-affecting changes include:
    decoding in a signal handler, along with instr_noalloc_init() and
    instr_from_noalloc().
  - Added drwrap_get_stats().
+ - Added #DRWRAP_NO_DYNAMIC_RETADDRS for reducing drwrap overhead at the cost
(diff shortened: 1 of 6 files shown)


derekbruening:
I'm cleaning up and checking in drwrap features developed a while ago.  For our third-party use we plan to not use this since it misses too many post-calls, but it seems worth committing to have in the toolbox, since it's pretty simple code to put in place.

derekbruening:
Appveyor is client.alloc which has become flaky in the new VS2017 config: #4058.  Failed a few times in the last week.  Should be added to the ignore list.

PR #4195, opened 2020-03-18
i#4187: Build a hashtable for faster ELF symbol lookups

### LINKED ISSUE #4187: improve drsym_unix_lookup_symbol linear walk performance
On a large app, drcachesim's -record_heap can take a long time (1.5 minutes!) to query just the handful of heap functions it's looking for.  (We recently added operator new and that rose to 2:40).

It looks like drsyms does a linear walk for symbol lookup!  We should have
it build a hashtable or sthg, or if we're worried about space take in a
list of things to look for (or trace_func.cpp should change to enumerate w/
its own callback to do a single walk).

Even for addr lookup, for ELF it looks linear.  MachO and PECOFF have a
binary search.

Or, add a flag --dynsym_only where func_trace never calls into drsyms, as a drcachesim-focused fix.

derekbruening:
Instead of a linear search through the symbol table on every lookup,
on the first lookup drsyms now builds a hashtable of all symbol
variants and uses it on subsequent lookups.

Adds a new option to the drcontainers hashtable to supply a key
freeing callback, needed here because we are duplicating most strings
ourselves before adding and do not want to have drcontainers
re-duplicate.

Adds a -verbose 1 notification for libraries taking long amounts of
time to be processed, for easier diagnostics in the future.

Performance-tested on large applications where it reduces drcachesim
-record_heap query times by 5x for less-optimized code and 2x for
highly-optimized code.

Fixes #4187

### DIFF
--- a/ext/drsyms/drsyms_unix_common.c
+++ b/ext/drsyms/drsyms_unix_common.c
@@ -38,6 +38,7 @@
 #include "drsyms.h"
 #include "drsyms_private.h"
 #include "drsyms_obj.h"
+#include "hashtable.h"
 
 #include "dwarf.h"
 #include "libdwarf.h"
@@ -73,6 +74,8 @@ typedef struct _dbg_module_t {
      * while the primary mod has symtab+strtab.
      */
     struct _dbg_module_t *mod_with_dwarf;
+#define SYMTABLE_HASH_BITS 12
+    hashtable_t symtable;
 } dbg_module_t;
 
 /******************************************************************************
@@ -84,6 +87,8 @@ unload_module(dbg_module_t *mod);
 static bool
 follow_debuglink(const char *modpath, dbg_module_t *mod, const char *debuglink,
                  char debug_modpath[MAXIMUM_PATH]);
+static void
+drsym_free_hash_key(void *key);
 
 /******************************************************************************
  * Module loading and unloading.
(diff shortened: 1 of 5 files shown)


ext/drsyms/drsyms_unix_common.c
```diff
@@ -439,6 +460,114 @@ addrsearch_symtab(dbg_module_t *mod, size_t modoffs, drsym_info_t *info INOUT, u
...
+
+    /* Add the demanglings. */
+    toadd = drsym_demangle_helper(sym, DRSYM_DEMANGLE);
+    if (toadd == NULL)
+        return true;
+    if (!drsym_add_hash_entry(mod, toadd, modoffs))
+        return true;
+
+    /* Add a version without parameters to allow the user to ignore overloads. */
+    char *noparen = drsym_dup_string_until_char(toadd, '(');
```
s-kanev: Unfortunately, C++ has a lot of cases where the simple '(' heurstic doesn't work, like `Foo::(anonymous namespace)::bar()`, or `std::function<int(int)>::foo()`.
[Here](https://github.com/google/pprof/blob/be90f332299a1d03fbb09bdbebdf28118ae300fb/profile/prune_test.go#L55)'s a few examples.

ext/drsyms/drsyms_unix_common.c
```diff
@@ -439,6 +460,114 @@ addrsearch_symtab(dbg_module_t *mod, size_t modoffs, drsym_info_t *info INOUT, u
     return drsym_obj_symbol_offs(mod->obj_info, idx, &info->start_offs, &info->end_offs);
 }
 
+/******************************************************************************
+ * Hashtable building for symbol lookup.
+ *
+ * We use __wrap_malloc because our demangled strings have larger capacities
+ * than strlen() will find.
```
s-kanev: I don't understand this. __wrap_malloc stores the size in the allocated block, but I don't see you using it anywhere. Nor calling strlen() on anything but the original symbol names, which are not allocated here.

ext/drsyms/drsyms_unix_common.c
derekbruening: Fair, but the initial goal was to match the existing code which just looked at the first paren.  I'll add a comment about the problems with this.

ext/drsyms/drsyms_unix_common.c
derekbruening: It's the drsym_demangle_helper code which has a loop to find the right size for the demangled symbol name, starting at 1024.  So that 1024 may well be much larger than strlen.  If it weren't for that over-capacity we would not need the stored size and could use strlen for the size on all frees.

derekbruening: We could also have the hashtable strdup for us but that seems too wasteful when we're already allocating separately for all but the original name.

derekbruening: I elaborated in the comment to help make this more clear.

Issue #4189, opened 2020-03-17
AArch64 Jenkins machine is not using the latest runsuite_wrapper.pl


derekbruening:
The changes I made recently to runsuite_wrapper.pl in PR #4165 do not seem to be reflected in the Jenkins tests.  When I look at the details such as here http://139.178.83.194:8080/job/DynamoRIO-AArch64-Precommit/1165/console I see a test failing that should be ignored by the wrapper (code_api|tool.drcachesim.invariants).  Looking at the commands run it seems to copy a version of the script from somewhere else:

```
+ cp /home/assad/dynamorio/suite/runsuite_wrapper.pl suite/runsuite_wrapper.pl
+ mkdir build
+ cd build
+ ../suite/runsuite_wrapper.pl travis
```

Hopefully we can just change this to use the checked-in version?

AssadHashmi:
> Hopefully we can just change this to use the checked-in version?
Yes. This mod was a temporary workaround which wasn't removed when it should have been, apologies. I've updated the commands in the precommit config and will keep an eye on the pipeline.

derekbruening:
@AssadHashmi -- this can be closed?

AssadHashmi:
Oops! Yes, closing...

PR #4196, opened 2020-03-18
i#4187 sym perf: Add -record_dynsym_only to drcachesim

derekbruening:
Adds a new option -record_dynsym_only to drcachesim which causes
-record_function and -record_heap to only look in .dynsym, avoiding
drsyms and the main symbol table entirely.  This is to avoid overhead
issues with two scenarios: very large applications with millions of
symbol table entries, and applications with thousands of libraries
where setting up and tearing down drsyms and DWARF information adds up
across all of those libraries.  Quite often the target functions to
trace are in .dynsym, making this a reasonable solution.

Tested manually on both scenarios.

Issue: #4187

### DIFF
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -250,6 +250,8 @@ Further non-compatibility-affecting changes include:
  - Added drwrap_get_stats().
  - Added #DRWRAP_NO_DYNAMIC_RETADDRS for reducing drwrap overhead at the cost
    of missing some post-call callbacks.
+ - Added -record_dynsym_only to drcachesim for faster function tracing symbol
+   lookups when internal symbols are not needed.
 
 **************************************************
 <hr>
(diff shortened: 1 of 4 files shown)


AssadHashmi:
Regarding the AArch64 failures:
```
The following tests FAILED:
	 63 - code_api|linux.mangle_asynch (Failed)
	 66 - code_api|linux.fib-conflict (Failed)
	 67 - code_api|linux.fib-conflict-early (Failed)
	 83 - code_api|security-common.TestMemProtChg_FLAKY (Failed)
```
`linux.fib-conflict` and `linux.fib-conflict-early` are expected regressions hitherto masked by [i#4189](https://github.com/DynamoRIO/dynamorio/issues/4189) (I will update the `%ignore_failures_64` list in `runsuite_wrapper.pl` shortly.)

But `code_api|linux.mangle_asynch` should not fail AIUI. Is the failure caused by this PR?

derekbruening:
> But code_api|linux.mangle_asynch should not fail AIUI. Is the failure caused by this PR?

It seems very unlikely to be related to this PR since it does not change core behavior at all, only drcachesim.

PR #4207, opened 2020-03-21
i#4134: Add statistics gathering to drbbdup

johnfxgalea:
Adds statistics gathering to drbbdup. In particular, the PR includes drbbdup_get_stats(), which populates a passed struct with current stat values.

Issue: #4134

### DIFF
--- a/ext/drbbdup/drbbdup.c
+++ b/ext/drbbdup/drbbdup.c
@@ -862,7 +862,8 @@ drbbdup_insert_dynamic_handling(void *drcontext, app_pc translation_pc, void *ta
 
     if (opts.is_stat_enabled) {
         /* Insert clean call so that we can lock stat_mutex. */
-        dr_insert_clean_call(drcontext, bb, where, drbbdup_inc_bail_count, false, 0);
+        dr_insert_clean_call(drcontext, bb, where, (void *)drbbdup_inc_bail_count, false,
+                             0);
     }
 
     instrlist_meta_preinsert(bb, where, done_label);
--- a/ext/drbbdup/drbbdup.c
+++ b/ext/drbbdup/drbbdup.c
@@ -1371,11 +1371,12 @@ drbbdup_get_stats(OUT drbbdup_stats_t *stats_in)
 {
     if (!opts.is_stat_enabled)
         return DRBBDUP_ERROR_UNSET_FEATURE;
-    if (stats_in == NULL)
+    if (stats_in == NULL || stats_in->struct_size == 0 ||
+        stats_in->struct_size > stats.struct_size)
         return DRBBDUP_ERROR_INVALID_PARAMETER;
 
     dr_mutex_lock(stat_mutex);
-    memcpy(stats_in, &stats, sizeof(drbbdup_stats_t));
+    memcpy(stats_in, &stats, stats_in->struct_size);
     dr_mutex_unlock(stat_mutex);
     return DRBBDUP_SUCCESS;
 }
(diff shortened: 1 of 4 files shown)


ext/drbbdup/drbbdup.h
```diff
@@ -54,6 +54,7 @@ extern "C" {
 typedef enum {
     DRBBDUP_SUCCESS,                       /**< Operation succeeded. */
     DRBBDUP_ERROR_INVALID_PARAMETER,       /**< Operation failed: invalid parameter. */
+    DRBBDUP_ERROR_UNSET_FEATURE,           /**< Operation failed: feature not set. */
```
derekbruening: Better to append to the end to not break compat.

ext/drbbdup/drbbdup.c
```diff
@@ -844,7 +859,11 @@ drbbdup_insert_dynamic_handling(void *drcontext, app_pc translation_pc, void *ta
             instrlist_meta_preinsert(bb, where, instr);
         }
     }
-    /* XXX i#4134: Insert code for dynamic handling here. */
+
+    if (opts.is_stat_enabled) {
+        /* Insert clean call so that we can lock stat_mutex. */
+        dr_insert_clean_call(drcontext, bb, where, drbbdup_inc_bail_count, false, 0);
```
derekbruening: How about drx_insert_counter_update() with DRX_COUNTER_LOCK

ext/drbbdup/drbbdup.h
```diff
@@ -257,8 +258,32 @@ typedef struct {
      * thread before it becomes a candidate for dynamic generation.
      */
     ushort hit_threshold;
+    /**
+     * Determines whether drbbdup should track a variety of statistics. Note keeping track
+     * of statistics incurs additional overhead and it is not recommended at deployment.
+     *
+     * Set to true if the client calls drbbdup_get_stats().
```
derekbruening: This could confuse people: it could be read as "this is automatically set to true if the client ever calls drbbdup_get_stats".  Adding a few more words should clarify it.

ext/drbbdup/drbbdup.h
```diff
@@ -257,8 +258,32 @@ typedef struct {
...
+     */
+    bool is_stat_enabled;
 } drbbdup_options_t;
 
+/**
+ * Various statistics related to drbbdup.
+ */
+typedef struct {
+    /** Number of fragments which have case handling turned off. */
```
derekbruening: Have a struct size parameter, like all the similar interface structs do (dr_stats_t, drwrap_stats_t), so we can safely append to this in the future.

ext/drbbdup/drbbdup.h
```diff
@@ -257,8 +258,32 @@ typedef struct {
...
+    unsigned long no_dup_cnt;
```
derekbruening: suggest: just spell out "count" for clarity, here and below

ext/drbbdup/drbbdup.c
johnfxgalea: Only downside is that on 32-bit you are limited to uint causing potential overflow for long running CPU-intensive apps.

ext/drbbdup/drbbdup.h
johnfxgalea: Done.

ext/drbbdup/drbbdup.h
johnfxgalea: Done

ext/drbbdup/drbbdup.h
johnfxgalea: Done.

ext/drbbdup/drbbdup.h
johnfxgalea: Done.

ext/drbbdup/drbbdup.c
derekbruening: Yeah this bit me too in the drwrap_stats (that stat was ok being small so I shrunk it for 32-bit as a workaround).  I filed #4215 on adding better atomics support.

johnfxgalea: I added a fixme comment.

Issue #3406, opened 2019-03-01
Rare/random code_api|linux.fib-conflict-early test failure dll library bounds check.


hgreving2304:
From https://travis-ci.com/DynamoRIO/dynamorio/jobs/181695424

> 92/307 Test  #86: code_api|linux.fib-conflict-early ............................***Failed  Required regular expression not found.Regex=[^fib\(5\)=8
fib\(15\)=987
fib\(25\)=121393
fib\(0\)=1
fib\(1\)=1
fib\(2\)=2
fib\(3\)=3
fib\(4\)=5
fib\(5\)=8
fib\(6\)=13
fib\(7\)=21
fib\(8\)=34
fib\(9\)=55
fib\(10\)=89
fib\(11\)=144
fib\(12\)=233
fib\(13\)=377
fib\(14\)=610
fib\(15\)=987
fib\(16\)=1597
fib\(17\)=2584
fib\(18\)=4181
fib\(19\)=6765
fib\(20\)=10946
fib\(21\)=17711
fib\(22\)=28657
fib\(23\)=46368
fib\(24\)=75025
fib\(25\)=121393
fib\(26\)=196418
fib\(27\)=317811
fib\(28\)=514229
fib\(29\)=832040
fib\(30\)=1346269
fib\(31\)=2178309
fib\(32\)=3524578
fib\(12\)=233
$
]  0.04 sec
<CURIOSITY : image_size == 0 || cur_end - mod_start == image_size in file /home/travis/build/DynamoRIO/dynamorio/core/unix/memquery.c line 227
version 7.1.17956, custom build
-early_inject >
<Application  (22864).  Internal Error: DynamoRIO debug check failure: /home/travis/build/DynamoRIO/dynamorio/core/unix/os.c:8497 check_start == dynamo_dll_start && check_end == dynamo_dll_end
(Error occurred @-1 frags)
version 7.1.17956, custom build
-early_inject 
0x00007ffd6d1f9630 0x00007f9a05c1eadd
0x00007ffd6d1f9870 0x00007f9a05df678b
0x00007ffd6d1f98d0 0x00007f9a05df6a15
0x00007ffd6d1f98e0 0x00007f9a05e18683

derekbruening:
The same thing happened in the `fib-conflict` test:
https://travis-ci.com/DynamoRIO/dynamorio/jobs/184740308

derekbruening:
[...]

PR #4209, opened 2020-03-21
i#4108: Avoid replacing WiX paths

### LINKED ISSUE #4108: runsuite_common_pre mangles non-VS paths, messing up WiX
Dr. Memory uses WiX to build its installer and its path is mangled by runsuite_common_pre.cmake:
`C:\Program Files (x86)\WiX Toolset v3.11\bin` =>
`C:\Program Files (x86)\WiX Toolset v3.11\bin\x64`

The code looks for a "v" trying to match `SDKs/Windows/v{6.0A,7.0}`:
```
          string(REGEX REPLACE "(v[^/\\\\]*)([/\\\\])([Bb][Ii][Nn])" "\\1\\2\\3\\2x64"
            newpath "${newpath}")
```

For now Dr. Memory is computing the 8.3 path and using that instead as a workaround.

derekbruening:
Tightens the regex for versioned SDK directories to look for a slash
followed by a "v" to avoid matching the WiX path, which we do not want
to mangle.

Also fixes a CMake warning about mismatched conditional end labels in
the same code.

Fixes #4108

### DIFF
--- a/suite/runsuite_common_pre.cmake
+++ b/suite/runsuite_common_pre.cmake
@@ -504,7 +504,8 @@ function(testbuild_ex name is64 initial_cache test_only_in_long
             "\\1VC\\3\\4\\3amd64;\\1VC\\3\\4"
             newpath "${newpath}")
           # VS2008's SDKs/Windows/v{6.0A,7.0} uses "x64" instead of "amd64"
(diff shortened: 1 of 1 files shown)


derekbruening:
Travis is fib-conflict-early: #3406.

PR #4210, opened 2020-03-21
DO NOT CHECK IN: try to repro i#4126 hang in a pull request build tha…

derekbruening:
…t pretends to be a cron build


derekbruening:
So this does repro #4126, good.

derekbruening:
Travis failure is some kind of Travis internal bug: "An error occurred while generating the build script." on the Mac job.

Issue #4200, opened 2020-03-20
crash in privload_call_lib_func i386 , when use client with drrun.


hac425xxx:
platform
```
i386 ubuntu 18
```

version
```
https://github.com/DynamoRIO/dynamorio/releases/download/cronbuild-7.91.18319/DynamoRIO-Linux-7.91.18319.tar.gz

master also has this bug.
```

when run drrun with client , it could crash.

```
drrun -c samples/bin32/libinscount.so  -- ./filefuzz
<Application /xxx/filefuzz (101859).  DynamoRIO internal crash at PC 0x00000035.  Please report this at http://dynamorio.org/issues/.  Program aborted.
Received SIGSEGV at unknown pc 0x00000035 in thread 101859
Base: 0xf7d27000
Registers:eax=0x00000000 ebx=0xff8c8e20 ecx=0x00000035 edx=0xf7a82a40
	esi=0xf7f260dc edi=0xff8c8e14 esp=0xff8c8dfc ebp=0xff8c8e1a
	eflags=0x00010282
version 7.91.18319, custom build
-no_dynamic_options -client_lib '/xxxx/samples/bin32/libinscount.so;0;' -code_api -stack_size 56K -signal_stack_size 32K -max_elide_jmp 0 -max_elide_call 0 -early_inject -emulate_brk -no_inline_ignored_syscalls -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2d

```

this is the backtrace of this bug
```
(gdb) bt
#0  0x00000035 in ?? ()
#1  0xf7f07e59 in privload_call_lib_func (func=0x35) at /home/travis/build/DynamoRIO/dynamorio/core/unix/loader.c:990
#2  privload_call_entry (dcontext=0xf7fb6000, privmod=0xffffcd54, reason=4159886336) at /home/travis/build/DynamoRIO/dynamorio/core/unix/loader.c:619
#3  0x00000000 in ?? ()
(gdb) quit

```

hac425xxx:
It seem's, this commit import the bug
```
https://github.com/DynamoRIO/dynamorio/commit/eab941b43e5aef8447aeb924883c5552aacd9a3e
```

johnfxgalea:
I confirm this happening on my machine too. It seems that this is not just a particular issue of libinscount.so but also other clients.

What I noted is that if I I run it under debug mode it works right?

johnfxgalea:
Also, what gcc version are you using please?

[...]

Issue #3995, opened 2019-12-16
Collect multiple trace samples from a single benchmark run


prasun3:
**Add a pointer to any prior users list discussion.**
Currently we can collect one trace sample during a benchmark run. We use '-trace_after_instrs' and '-exit_after_tracing' to select a trace point.

**Is your feature request related to a problem? Please describe.**
We need to run the benchmark multiple times to collect traces throughout the benchmark execution. For long running benchmark, this may take a very long time.

**Describe the solution you'd like**
Have a way to ‘trace x million insts every y million insts’. A more advanced method would be to have a config file listing the sampling windows.

Do you have any implementation in mind for this feature?
No

**Describe alternatives you've considered**
Currently we plan to run the benchmark with a diff sampling point.

**Additional context**
None.

johnfxgalea:
Thank you for the request. I guess one quick solution is to use an script that runs the benchmark multiple times with updated param values for -trace_after_instrs. However, I do understand this is not ideal for long benchmarks.

derekbruening:
Presumably the approach of recording a single long trace covering all desired execution windows and then splitting it up into pieces offline will not work due to such a trace simply being too large to easily store?

prasun3:
That is right. We see pretty large trace sizes and run into disk space issues.

derekbruening:
Xref an existing feature request to use annotations added to the application to delineate phase regions and have the tracer recognize the annotations and enable/disable recording a the boundaries.

Also note that another method of creating multiple traces from one execution is to insert start/stop commands into the application.  This is well-supported today, in particular with static linkage of the tracer into the application, and we have a number of regression tests of this.  E.g., see https://github.com/DynamoRIO/dynamorio/blob/master/clients/drcachesim/tests/burst_static.cpp

Those two approaches both require modifying the application.  This issue here covers specifying boundaries on an unmodified application.

derekbruening:
@prasun3 -- would your use case prefer to modify the application with annotations in the source code to delineate precise tracing regions?  Or you would prefer this feature as filed to trace a certain number of instructions without regard to any corresponding application phases or code boundaries?

prasun3:
>  Or you would prefer this feature as filed to trace a certain number of instructions without regard to any corresponding application phases or code boundaries?

We would prefer this approach -- based on instruction count.

derekbruening:
[...]

Issue #3966, opened 2019-11-24
CRASH private lib fini unaligned stack


derekbruening:
Running 32-bit drmemtrace reliably hits this crash on my system:

```
Thread 1 "drmemtrace.sign" received signal SIGSEGV, Segmentation fault.
0xf777e732 in ?? ()
(gdb) x/8i $pc
=> 0xf777e732 <__cxa_finalize+258>:	pxor   (%esp),%xmm0
   0xf777e737 <__cxa_finalize+263>:	movd   %xmm0,%edx
   0xf777e73b <__cxa_finalize+267>:	psrlq  $0x20,%xmm0
   0xf777e740 <__cxa_finalize+272>:	movd   %xmm0,%eax
   0xf777e744 <__cxa_finalize+276>:	or     %edx,%eax
   0xf777e746 <__cxa_finalize+278>:	jne    0xf777e66d <__cxa_finalize+61>
(gdb) x/4wx $esp
0x48456868:	0x0000004a	0x00000000	0xf7fa7000	0xf7fa7000
(gdb) memquery $esp
48449000-48457000 rw-p 00000000 00:00 0
(gdb) bt
#0  0xf777e732 in __cxa_finalize ()
#1  0xf7be60d3 in __do_global_dtors_aux ()
#2  0xf7e985fa in privload_call_lib_func (func=0xf7be60a0 <__do_global_dtors_aux>) at .../dr/git/src/core/unix/loader.c:980
#3  0xf7e97a24 in privload_call_entry (privmod=0x48427e30, reason=2) at .../dr/git/src/core/unix/loader.c:649
#4  0xf7db938b in privload_unload (privmod=0x48427e30) at .../dr/git/src/core/loader_shared.c:615
#5  0xf7db81bf in unload_private_library (modbase=0xf7bd5000 "\177ELF\001\001\001\003") at .../dr/git/src/core/loader_shared.c:274
#6  0xf7e707ba in unload_shared_library (lib=0xf7bd5000) at .../dr/git/src/core/unix/os.c:3912
#7  0xf7dc085a in instrument_exit () at .../dr/git/src/core/lib/instrument.c:889
#8  0xf7c6e4fd in dynamo_shared_exit (toexit=0x0) at .../dr/git/src/core/dynamo.c:1024
#9  0xf7c6e9ef in dynamo_process_exit_cleanup () at .../dr/git/src/core/dynamo.c:1349
[...]

PR #4212, opened 2020-03-22
i#4134: Arrange drbbdup to enable users to specify memory

johnfxgalea:
This PR makes several changes to drbbdup's interface to allow the user to optionally set drbbdup_insert_encode_t call-back to NULL. The interface is also modified so that the runtime case encoding is obtained via an operand provided by the user. 

The handling of fragment deletion events, used to invalidate book-keeping data of basic blocks, is removed. Instead, basic block data is invalidated upon the creation of new basic blocks. 

The PR updates the docs according to the changes made, and also adds a new test. Several bugs and typos in comments are also fixed.

Issue: #4134


ext/drbbdup/drbbdup.c
```diff
@@ -697,25 +706,35 @@ drbbdup_encode_runtime_case(void *drcontext, drbbdup_per_thread *pt, void *tag,
...
 #ifdef X86_32
     /* Load the encoding to the scratch register.
      * The dispatcher could compare directly via mem, but this will
      * destroy micro-fusing (mem and immed).
      */
     opnd_t scratch_reg_opnd = opnd_create_reg(DRBBDUP_SCRATCH_REG);
-    opnd_t opnd = drbbdup_get_encoding_opnd();
+    opnd_t opnd;
+#    ifdef DEBUG
+    drbbdup_status_t res =
```
derekbruening: Use `IF_DEBUG(drbbdup_status_t res =)` for more readable code.  Ditto below.

ext/drbbdup/drbbdup.h
```diff
@@ -316,26 +324,27 @@ drbbdup_register_case_encoding(void *drbbdup_ctx, uintptr_t encoding);
...
  */
 drbbdup_status_t
 drbbdup_set_encoding(uintptr_t encoding);
 
 DR_EXPORT
 /**
- * Retrieves a memory destination operand which should be used to set the runtime case
- * encoding.
+ * Returns a memory destination operand by setting \p dst_opnd. The operand
+ * should be used to set the runtime case encoding.
```
derekbruening: "for the current thread"

ext/drbbdup/drbbdup.h
```diff
@@ -316,26 +324,27 @@ drbbdup_register_case_encoding(void *drbbdup_ctx, uintptr_t encoding);
 
 DR_EXPORT
 /**
- * Sets the runtime case encoding \p encoding.
+ * Sets the runtime case encoding \p encoding for a given thread.
```
derekbruening: s/a given/the current/

derekbruening: This is the complication with this approach: to change all the case encodings the client has to walk the threads and know where the TLS lives, and there's no function here to get the address easily.  It can be done via gencode but it's complex.  So drbbdup needs to provide the address to C code, or let the client control the TLS where it's stored so it can track the addresses on its own.  Certainly it would use its own thread TLS alloc list and atomics to write and avoid anything like dr_suspend_*.

I'm wondering if this is a better idea:

* Revert to the required insert_encode.
* Change drbbdup_get_encoding_opnd() to be a *register*.  Now the client can use a global var to hold the mode, and just do a load to the register at the top of every block.  It also has the flexibility (either b/c modes are not global, or b/c the barriers for prompt-enough shared loads on arm are too expensive) to keep the mode per-thread in its own TLS and load it to the register.  By having a register instead of memory there's no extra overhead.
* The only complication I see is drbbdup_set_encoding() for clients who want to set it from C code: since it could be far separated from the compare we wouldn't have it set the register in mcxt I don't think.  Either the client would have to implement it on its own by making a TLS slot and setting that and having insert_encode load from that TLS slot, or there's an option for drbbdup to load the value from its own TLS slot like it does today -- maybe if insert_encode is NULL it does that.

johnfxgalea: I need to look further into to code - but on the top of my head, having drbbdup_get_encoding_opnd() be a register would require drbbdup to spill an additional GPR to handle dynamic case generation. Also, insert_encode may need to know which register is being reserved as it cannot be used for its own operation. For example, if it uses drreg to reserve a particular register and that register is already used by drbbdup, drreg will fail.

johnfxgalea: Are you avoiding mem2mem stores because of register indirection and performance reasons?

derekbruening: I don't want to have two memory refs when only one is necessary: give back a cache line to the app.

ext/drbbdup/drbbdup.c
johnfxgalea: Done.

ext/drbbdup/drbbdup.h
johnfxgalea: Done.

ext/drbbdup/drbbdup.h
johnfxgalea: Done.

derekbruening: Any interface that ends up with a single load, done atomically+visibly (so a barrier for ARM; load-acquire for AArch64; I plan to add helpers for such loads as part of #4215) so I can write to it from another thread, should work for me.  I would be ok with it having to be TLS if I can get the linear address for each thread and iterate over them; global would be simpler for my use case which has simple global phases -- not sure about other use cases.

johnfxgalea: I don't think it is a good idea for the dispatcher to perform atomic operations. It needs to be fast.

johnfxgalea: NVM, misread as saying internally, the dispatcher should load atomically.

johnfxgalea: Okay so one problem is that preserving a register prior to the insert_encode limits the latter's flexibility. The use-case may require the original value of that register which is internally used by drbbdup, e.g., to compute an app address. I can inform the user that the register has the original app value, but I think the interface will be a bit messy. Storing the encoding into memory also allows easy integration with drreg - I can safely restore all registers after the insert_encode has been invoked without destroying the encoding and making dispatcher logic easier.

johnfxgalea: I guess if a register is going to be used, then the user cannot reserve the same register in insert_encode using drreg. Drbbdup will reserve the register manually, so if the user is not careful and tries to reserve that register, drreg will return with success.

derekbruening: Using memory instead of a register: what if the user could specify where the memory location is (pass an opnd_t)?

johnfxgalea: Look at my last updated comment.

Regarding the specification of memory - that will result in drbbdup always performing an atomic load regardless of the use-case. I don't think that will help performance.

johnfxgalea: Actually, I guess the user can pass a flag indicating whether to lock.

johnfxgalea: To support dynamic fast path generation, I need guarantees that the memory passed will not change throughout the dispatching process - without the use of any locks. Can your use-case guarantee this?

derekbruening: No my use case can't guarantee that: the trigger for a case changing is the global instruction count (this is #3995) and some other thread could cross the threshold and swap the case for all threads at any time.

I'm not sure what dynamic fast path generation is so I don't know whether it applies to my use case.  If not maybe there can be optional features for cases like mine that just swap between 2 global cases.

johnfxgalea: I mean dynamic case generation - I use the terms interchangeably because ultimately one is generating different paths via drbbdup.

johnfxgalea:
This PR ended up touching quite a lot of code - sorry about this as I had to update the interface and docs given the support of client's now being able to maintain memory for storing runtime case encodings.

ext/drbbdup/drbbdup.h
johnfxgalea: Done (I think) but let me know if it can be better.

derekbruening: Sorry, going to be busy next day or so, so may be a delay before I can take look.

johnfxgalea: Don't worry, take your time.

ext/drbbdup/drbbdup.c
```diff
@@ -708,29 +697,51 @@ drbbdup_encode_runtime_case(void *drcontext, drbbdup_per_thread *pt, void *tag,
...
+        /* Restore all unreserved registers used by the call-back. */
+        drreg_restore_all(drcontext, bb, where);
+    }
+
+    /* Load the encoding to the scratch register. */
     opnd_t scratch_reg_opnd = opnd_create_reg(DRBBDUP_SCRATCH_REG);
-    opnd_t opnd = drbbdup_get_encoding_opnd();
-    instr_t *instr = INSTR_CREATE_mov_ld(drcontext, scratch_reg_opnd, opnd);
+    instr_t *instr =
+        INSTR_CREATE_mov_ld(drcontext, scratch_reg_opnd, opts.runtime_case_opnd);
```
derekbruening: The issue for a globally changing case is memory visibility.  If I make my encoding address a single global, I would need a barrier here for aarchxx.  The alternative is to have the encoding be TLS and when the value changes go force all the threads to update their values -- although I'm not sure how to do that.  If I could run code as each thread I could have each go do an acquire load.  I can do that on UNIX by sending a signal.  I suppose I could do NtSetContextThread on Windows twice to run my code sequence.  I would want to add DR API support for those.  dr_suspend_all_other_threads() could do it I guess by setting the mcontext and then resuming -- but there's no per-thread control point to restore w/o having a wait point and another suspend-the-world.

I don't know how performance of a barrier at the top of a every block compares to a heavyweight interrupt-all-threads-when-change approach. Certainly the barrier is much much easier to implement (esp when #4215 is done).

How do you envision a typical use case changing the encoding?  Are all the use cases you're thinking of wanting local changes where only the current thread changes its own encoding independently of all others?

Should there then be an option for whether to use an acquire load or a regular load?  (This is why I was proposing a register interface before.)

ext/drbbdup/drbbdup.c
```diff
@@ -708,29 +697,51 @@ drbbdup_encode_runtime_case(void *drcontext, drbbdup_per_thread *pt, void *tag,
...
+    instr_t *instr = INSTR_CREATE_cmp(drcontext, opnd, opnd_create_reg(reg_encoding));
     instrlist_meta_preinsert(bb, where, instr);
-#endif
 }
+#elif X86_32
+static void
+drbbdup_insert_compare_encoding(void *drcontext, instrlist_t *bb, instr_t *where,
+                                drbbdup_case_t *current_case, reg_id_t reg_encoding)
+{
+    /* Note, DRBBDUP_SCRATCH_REG contains the runtime case encoding. */
```
derekbruening: I don't see how the comment relates to the code?  This function takes in `reg_encoding`.

ext/drbbdup/drbbdup.c
```diff
@@ -708,29 +697,51 @@ drbbdup_encode_runtime_case(void *drcontext, drbbdup_per_thread *pt, void *tag,
...
+        INSTR_CREATE_mov_ld(drcontext, scratch_reg_opnd, opts.runtime_case_opnd);
+    instrlist_meta_preinsert(bb, where, instr);
+}
+
+#ifdef X86_64
+static void
+drbbdup_insert_compare_encoding(void *drcontext, instrlist_t *bb, instr_t *where,
+                                drbbdup_case_t *current_case, reg_id_t reg_encoding)
+{
+    opnd_t opnd = opnd_create_abs_addr(&current_case->encoding, OPSZ_PTR);
```
derekbruening: This seems like a bug: `drbbdup_manager_t.cases` is *not* allocated as reachable heap, so it may not be reachable (i.e., more than 2GB away from the code cache).

ext/drbbdup/drbbdup.c
johnfxgalea: The interface could take a an option whether to use a lock. Another way is to support an additional call-back function that takes the memory operand and a register and loads the runtime encoding. The call-back is invoked after insert_encoding and drreg restoration are done.

johnfxgalea: > How do you envision a typical use case changing the encoding? Are all the use cases you're thinking of wanting local changes where only the current thread changes its own encoding independently of all others?

Yep, all the ones I used are thread-local. I figured locking at the start of every basic block for global yields a performance hit that one would like the avoid.

ext/drbbdup/drbbdup.c
johnfxgalea: Thanks for the catch. I'm afraid I am going to have to make it reachable then because otherwise the dispatcher will need to spill an additional register (which I want to avoid).

api/samples/hot_bbcount.c
```diff
@@ -240,8 +245,18 @@ dr_client_main(client_id_t id, int argc, const char *argv[])
...
+
+    /* The operand referring to memory storing the current runtime case encoding. */
+    opnd_t case_opnd = opnd_create_far_base_disp_ex(
+        tls_raw_reg, REG_NULL, REG_NULL, 1, tls_raw_base, OPSZ_PTR, false, true, false);
+
     /* Initialise drbbdup. Essentially, drbbdup requires
-     * the client to pass a set of call-back functions.
+     * the client to pass a set of call-back functions and
+     * the memory operand that the dispatcher will use
+     * to load the current runtime case encoding.
```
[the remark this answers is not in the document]
johnfxgalea: I was more concerned how to design the API with the two approaches in concert. I was afraid that users will get confused and incorrectly invoke different API routines assumed to be in different settings. Keeping one single interface renders simple usage.

ext/drbbdup/drbbdup.c
derekbruening: Does drbbdup read the encoding anywhere else?  I.e., could we remove the memory operand part of the interface and only have the register part.  So it would be back to the prior register proposal, except drbbdup picks the register instead of the user.

johnfxgalea: The problem with that approach is the drreg restoration. The encoding is stored in memory so that drbbdup can restore *all* register values which were used by the user-defined insert_encoding call-back. This makes dispatch easier as the control landing of the appropriate basic block does not concern heavy register restoration.  If the user stores the encoding in a register and memory is not touched, then drbbdup will destroy the encoding.

derekbruening: Still not really understanding the issue here.
It sounds like you would like to have two different callbacks and a provided memory address/operand X:

```
init: set mem/address/operand X
in bb:
  insert_encoding_to_memory_address_X_cb()
  restore all drreg lazily spilled regs by the above cb
  load_encoding_to_register_from_X_cb()
```

But wouldn't it be simpler to have:

```
<nothing in init>
in bb:
  load_encoding_to_register_from_anywhere_cb()
  restore all drreg lazily spilled regs by the above cb EXCEPT the target register
```

Just loop over the GPR's, skipping the one that holds the encoding.

johnfxgalea: Do you mean create a new drreg API that un-reserves a register without ever restoring it back to its original value so that the encoding won't get clobbered? Basically, drbbdup maintains its own spill slot for the single target register in order to safely restore it after the dispatcher has directed control to the appropriate basic block case (drreg is not really designed for restoring registers when jumping across app code).

The situation is as follows:

```
init: set mem/address/operand X
in bb:
  insert_encoding_to_memory_address_X_cb() with the possibility of using ALL registers, maintained by drreg, to compute the encoding. 
  restore all drreg lazily spilled regs by the above cb < END DRREG USAGE WITH A CLEAN STATE>
  load_encoding_to_register_from_X_cb()  where the target register is not maintained by drreg.
```

Essentially I would like all registers to be maintained by drreg for the insert_encoding call-back so that the user can leverage the full drreg API, e.g. drreg_restore_app_value() including the target register.

derekbruening: The drreg clean state can be between the cmp and jccc, with only the flags crossing the branch.

johnfxgalea: What if you have multiple cases defined and therefore multiple compare and branch sequences?

derekbruening: > > How do you envision a typical use case changing the encoding? Are all the use cases you're thinking of wanting local changes where only the current thread changes its own encoding independently of all others?
> 
> Yep, all the ones I used are thread-local. I figured locking at the start of every basic block for globals yields a performance hit that one would like the avoid.

Whether the cases are thread-local or global is part of the user-facing feature though: it's not just some internal implementation choice.  Do you have any suggestion for how to implement #3995 that does not involve either an acquire load (barrier to ensure stores from other threads are seen) on every single bb, or the scheme I described before of forcing all other threads to run such an acquire at the time of a mode change?

derekbruening: > What if you have multiple cases defined and therefore multiple compare and branch sequences? Subsequent cmps won't have the runtime encoding available.

OK, though it feels like I didn't run into this kind of thing with the internal control flow in DrM: its general scheme is to reserve several regs once at the very top and don't unreserve until the very bottom where all control flow joins up for a single unreserve, with any extra regs needed completely local and not crossing branches (presumably with a barrier before a branch)(though DrM has its own register scheme and converting to drreg is not done and in an in-progress branch and I kept hitting feature I needed to add to drreg to make it work so maybe there is sthg that doesn't work still).  If this case-holding register is reserved across all the dispatch branches and throughout all the per-case code and only unreserved once at the bottom where all the cases converge back again does it work out for drbbdup.

johnfxgalea: > > > How do you envision a typical use case changing the encoding? Are all the use cases you're thinking of wanting local changes where only the current thread changes its own encoding independently of all others?
> > 
> > 
> > Yep, all the ones I used are thread-local. I figured locking at the start of every basic block for globals yields a performance hit that one would like the avoid.
> 
> Whether the cases are thread-local or global is part of the user-facing feature though: it's not just some internal implementation choice. Do you have any suggestion for how to implement #3995 that does not involve either an acquire load (barrier to ensure stores from other threads are seen) on every single bb, or the scheme I described before of forcing all other threads to run such an acquire at the time of a mode change?

I would use locking for globals. I am not saying that drbbdup cannot be used for global cases, just that all the use-cases that I implemented were thread-local. However, I don't think that this is the issue we are facing.

In particular, the problem revolves around defining a suitable interface while dealing with the joint intricacies of drbbdup and other DR components such as drreg. I don't think there is an absolute correct answer, each design having its own  advantages and disadvantages.

- drbbdup maintains its own TLS slot and insert_encoding() stores the encoding to memory

(-) mem2mem copy, wasting a cache line.
(+)  insert_encoding() can leverage the full use of the drreg API.
(+) No complications regarding register restoration.
(+) Provides memory usage visibility, i.e., the call-back can easily perform a lock when loading the encoding.
(+) drbbdup book-keeping data per basic block can be stored in non reachable heap memory without the need to spill an additional register.

- insert_encoding() stores the encoding to a register

(-) insert_encoding() cannot leverage the full use of the drreg API or complications arise due to register restoration cross app-code.
(-) drbbdup book-keeping data per basic block has to be stored in non-reachable heap memory or another register needs to be spilled.
(+) Provides memory-usage-visibility
(+) No mem2mem copy

- drbbdup loads encoding from memory maintained by client.

(-) Problems with memory-usage-visibility since drbbdup internally performs the load.
(-) drbbdup book-keeping data per basic block has to be stored in non-reachable heap memory or another register needs to be spilled.
(+) No mem2mem copy
(+) No complications regarding register restoration.
(+)  insert_encoding() can leverage the full use of the drreg API.

derekbruening: For my use case, performance is a priority, probably ruling out the first one.  The second would be fine for me since my encoding is very simple, but for a general solution maybe what you were proposing with the last one plus a load callback is the way to go.

ext/drbbdup/drbbdup.c
johnfxgalea: Is dr_global_alloc() non reachable heap memory please?

Docs seem this is the case: 

> Memory allocated through dr_thread_alloc(), dr_global_alloc(), and dr_nonheap_alloc() is also reachable

https://dynamorio.org/dynamorio_docs/dr__tools_8h.html#abb7e20e83f7b8e1b65428e45c8ab4211

ext/drbbdup/drbbdup.c
johnfxgalea: Done.

api/samples/hot_bbcount.c
```diff
@@ -240,8 +245,18 @@ dr_client_main(client_id_t id, int argc, const char *argv[])
                       false /*synchronization is external*/, destroy_hit_count, NULL,
                       NULL);
 
+    /* Initialise an addressable TLS slot to contain the runtime case encoding. */
+    if (dr_raw_tls_calloc(&tls_raw_reg, &tls_raw_base, 1 /* num of slots */, 0))
+        DR_ASSERT(false);
+
+    /* The operand referring to memory storing the current runtime case encoding. */
+    opnd_t case_opnd = opnd_create_far_base_disp_ex(
```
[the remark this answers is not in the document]
johnfxgalea: Done.

api/samples/hot_bbcount.c
```diff
@@ -240,8 +245,18 @@ dr_client_main(client_id_t id, int argc, const char *argv[])
                       false /*synchronization is external*/, destroy_hit_count, NULL,
                       NULL);
 
+    /* Initialise an addressable TLS slot to contain the runtime case encoding. */
+    if (dr_raw_tls_calloc(&tls_raw_reg, &tls_raw_base, 1 /* num of slots */, 0))
```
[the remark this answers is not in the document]
johnfxgalea: Done

ext/drbbdup/drbbdup.c
```diff
@@ -708,29 +697,51 @@ drbbdup_encode_runtime_case(void *drcontext, drbbdup_per_thread *pt, void *tag,
     }
 
     /* Encoding is application-specific and therefore we need to user to define the
```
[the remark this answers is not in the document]
johnfxgalea: Done.

ext/drbbdup/drbbdup.c
johnfxgalea: I added a new flag to the interface which informs drbbdup to perform the load locked/atomic.

ext/drbbdup/drbbdup.c
derekbruening: Hmm, I thought it was all un-reachable by default, which is definitely the case for heap allocated inside DR, where you have to call a special routine to get reachable heap.  Looks like I made all the client interface stuff reachable by default, though.  Sorry, there's no problem here then.

ext/drbbdup/drbbdup.c
derekbruening: SGTM

derekbruening:
On phone and not signed in so putting comment here: code LGTM except one
suggestion to rename “do_lock” to sthg with encode in it:
“atomic_load_encoding”??

ext/drbbdup/drbbdup.h
```diff
@@ -264,6 +264,11 @@ typedef struct {
      * in order to direct control to the appropriate basic block.
      */
     opnd_t runtime_case_opnd;
+    /**
+     * Instructs drbbdup whether or not the loading of the runtime case should be
+     * locked/atomic.
+     */
+    bool do_lock;
```
derekbruening, line 271: (Made it to computer so repeating in proper location:) Suggesting a name like "atomic_load_encoding".

johnfxgalea: Done.

johnfxgalea:
I think this PR is done and am waiting for tests. Interestingly, I hit a race on Windows due to a fragment deletion event. I removed the handling of such events and invalidate/clear old book-keeping data upon the creation of a new basic block.

johnfxgalea:
@derekbruening When you have a chance, can you please review the removal of delete frag events or point me to the code in drmemory please so I get an idea how it does it. 

My code still should work, but I had to stay checking for trace creation.

ext/drbbdup/drbbdup.c
```diff
@@ -394,8 +381,15 @@ drbbdup_duplicate_phase(void *drcontext, void *tag, instrlist_t *bb, bool for_tr
...
-     * book-keeping this bb. */
+    if (!for_trace && manager != NULL && !manager->is_gen) {
+        /* Remove existing invalid book-keeping data. */
+        hashtable_remove(&manager_table, pc);
+        manager = NULL;
+    }
+
+    /* A manager is created if there does not already exit one that "book-keeps"
+     * this basic block.
+     */
```
johnfxgalea, line 392: In particular, this part.

derekbruening:
> @derekbruening When you have a chance, can you please review the removal of delete frag events or point me to the code in drmemory please so I get an idea how it does it.
> 
> My code still should work, but I had to stay checking for trace creation.

Reviewing the history I'm seeing the counter solution, which drbbdup was already using, along with cases of invalidating on a new bb event:

### PR 495787: ASSERT (is_new, "duplicate bb?") deletion event can arrive after new bb(s) are created at that tag
This seems to be where this documentation was added to dr_register_delete_event():
```
   * Due to DR's high-performance non-precise flushing, a fragment
   * can be made inaccessible but not actually freed for some time.
   * A new fragment can thus be created before the deletion event
   * for the old fragment is raised.  We recommended using a counter
   * to ignore subsequent deletion events when using per-fragment
   * data structures and duplicate fragments are seen.
```

### drwrap switched from delete invalidation to new-bb invalidation

DynamoRIO/drmemory#673: post-call: check consistency on new bb rather than invalidate on delete
Talks about our own flush on post-call leading to a cycle b/c we trigger a
flush if the post-call code changed.
https://github.com/DynamoRIO/dynamorio/blob/master/ext/drwrap/drwrap.c#L354

### DynamoRIO/drmemory#768: 

Comment in fragment delete event cb:
```
        /* i#768: We used to invalidate entries from ignore_unaddr_table here,
         * but that ends up thrashing the code cache.  Instead we remove stale
         * entries in the new bb event if the alloca pattern no longer matches.
         */
```
https://github.com/DynamoRIO/drmemory/blob/master/drmemory/fastpath_x86.c#L3224

ext/drbbdup/drbbdup.c
```diff
@@ -383,8 +381,15 @@ drbbdup_duplicate_phase(void *drcontext, void *tag, instrlist_t *bb, bool for_tr
...
 
-    /* A manager is created if there does not exit one that does not already
-     * book-keeping this bb. */
+    if (!for_trace && manager != NULL && !manager->is_gen) {
+        /* Remove existing invalid book-keeping data. */
+        hashtable_remove(&manager_table, pc);
+        manager = NULL;
+    }
+
+    /* A manager is created if there does not already exit one that "book-keeps"
```
derekbruening, line 390: s/exit/exist/

ext/drbbdup/drbbdup.c
```diff
@@ -383,8 +381,15 @@ drbbdup_duplicate_phase(void *drcontext, void *tag, instrlist_t *bb, bool for_tr
     drbbdup_manager_t *manager =
         (drbbdup_manager_t *)hashtable_lookup(&manager_table, pc);
 
-    /* A manager is created if there does not exit one that does not already
-     * book-keeping this bb. */
+    if (!for_trace && manager != NULL && !manager->is_gen) {
+        /* Remove existing invalid book-keeping data. */
+        hashtable_remove(&manager_table, pc);
```
derekbruening, line 386: Does a lock need to be held here across this lookup and query?  Thinking of thread-private caches with two threads both creating the same block.

ext/drbbdup/drbbdup.c
derekbruening: Yes this is the basic idea.  Does a lock need to be held across the lookup and creation? Thinking of thread-private caches with two threads both creating the same block.  Maybe there's an outer lock I can't see in this diff view.  For that matter I'm wondering whether some of that Dr. Memory code handles thread-private duplicate tags properly...

ext/drbbdup/drbbdup.c
johnfxgalea: There is a write lock further up. The only general issue I see with thread-private caches is that a subsequent thread is going to invalidate the book-keeping data of a first thread, thus yielding the first creation useless. Not really sure how to get around this issue though or if I should in this PR. Could be an XY problem

johnfxgalea: In particular, this is due to the hash-table being shared amongst threads.

johnfxgalea: Thinking about this further - basic block data only because invalidated when dynamically generating a new case (denoted by the is_gen flag) or when the code of the basic block changes due to app performing some JIT or when a library is unloaded and another is loaded. drwrap compares the  call-bytes to check consistency. I could look into more specific consistency checks to avoid unnecessary invalidation instead of relying on DR's new bb events. hmm

derekbruening: Seems fine to get something working now and separate out theoretical corner cases for later.  The main short-term goal is to try to stabilize the interface for the DR 8.0 release (the only other thing blocking 8.0 is that I wanted to finish #3966 / #4200 b/c it's a compat break; I started on it but the clean call alignment is complicated and just needs more time -- could prob finish this week).

ext/drbbdup/drbbdup.c
johnfxgalea: Done.

ext/drbbdup/drbbdup.c
johnfxgalea: Thanks for taking a look. I think this PR is done and ready for merge. Iill move on to the analysis API.

PR #4213, opened 2020-03-22
i#4134: Add drbbdup_is_first_nonlabel_instr()

johnfxgalea:
Includes drbbdup_is_first_nonlabel_instr() to determine whether or not the instruction considered during insertion is the first non label instruction of the basic block copy. 

Issue: #4134

### DIFF
--- a/ext/drbbdup/drbbdup.c
+++ b/ext/drbbdup/drbbdup.c
@@ -1006,6 +1006,7 @@ drbbdup_instrument_without_dups(void *drcontext, void *tag, instrlist_t *bb,
 
     if (drmgr_is_first_instr(drcontext, instr)) {
         pt->first_instr = instr;
+        pt->first_nonlabel_instr = instrlist_first_nonlabel(bb);
         pt->last_instr = instrlist_last(bb);
         ASSERT(drmgr_is_last_instr(drcontext, pt->last_instr), "instr should be last");
     }
--- a/ext/drbbdup/drbbdup.c
+++ b/ext/drbbdup/drbbdup.c
@@ -111,8 +111,9 @@ typedef struct {
     void *default_analysis_data;     /* Analysis data specific to default case. */
     void **case_analysis_data;       /* Analysis data specific to cases. */
     uint16_t hit_counts[TABLE_SIZE]; /* Keeps track of hit-counts of unhandled cases. */
-    instr_t *first_instr; /* The first instr of the bb copy being considered. */
-    instr_t *last_instr;  /* The last instr of the bb copy being considered. */
+    instr_t *first_instr;          /* The first instr of the bb copy being considered. */
+    instr_t *first_nonlabel_instr; /* The first non label instr of the bb copy. */
+    instr_t *last_instr;           /* The last instr of the bb copy being considered. */
 } drbbdup_per_thread;
 
 static uint ref_count = 0;        /* Instance count of drbbdup. */
@@ -925,11 +926,22 @@ drbbdup_instrument_dups(void *drcontext, app_pc pc, void *tag, instrlist_t *bb,
         instr_t *end_instr = drbbdup_next_end(next_instr);
         ASSERT(end_instr != NULL, "end instruction cannot be NULL");
 
-        /* Cache first and last instructions. */
-        if (next_instr == end_instr && is_last_special)
+        /* Cache first, first nonlabel and last instructions. */
+        if (next_instr == end_instr && is_last_special) {
             pt->first_instr = last;
-        else
+            pt->first_nonlabel_instr = last;
+        } else {
             pt->first_instr = next_instr; /* Update cache to first instr. */
+            instr_t *first_non_label = next_instr;
+            while (instr_is_label(first_non_label) && first_non_label != end_instr) {
+                first_non_label = instr_get_next(first_non_label);
+            }
+            if (first_non_label == end_instr && is_last_special) {
+                pt->first_nonlabel_instr = last;
+            } else {
+                pt->first_nonlabel_instr = first_non_label;
+            }
+        }
 
         if (is_last_special)
             pt->last_instr = last;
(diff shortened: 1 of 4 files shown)


ext/drbbdup/drbbdup.dox
```diff
@@ -127,9 +127,9 @@ considers for instrumentation, it also provides a "where" instruction.
 The client must insert code exactly prior to the "where" instruction
 in order to ensure correct instrumentation.
 
-drbbdup also provides drbbdup_is_first_instr() and drbbdup_is_last_instr()
-to determine whether the passed instruction is the first
-or last instruction of a basic block copy respectively.
+drbbdup also provides drbbdup_is_first_instr() drbbdup_is_first_nonlabel_instr()
+and drbbdup_is_last_instr() to determine whether the passed instruction
+is the first or last instruction of a basic block copy respectively.
```
derekbruening: The docs should state that the client should *not* be calling drmgr_is_{first,first_nonlabel,last}_instr() because those will fail and always return false.

derekbruening: Maybe drbbdup.h should do what drmgr.h does:

```
#define drmgr_is_first_instr DO_NOT_USE_drmgr_is_first_instr_USE_drbbdup_instead
```

derekbruening: I think the docs should also explicitly point out that to pass info from a separate app2app phase the user should store into TLS since it can't use the drmgr user_data out of that phase.

ext/drbbdup/drbbdup.c
```diff
@@ -925,11 +926,22 @@ drbbdup_instrument_dups(void *drcontext, app_pc pc, void *tag, instrlist_t *bb,
...
-        if (next_instr == end_instr && is_last_special)
+        /* Cache first, first nonlabel and last instructions. */
+        if (next_instr == end_instr && is_last_special) {
             pt->first_instr = last;
-        else
+            pt->first_nonlabel_instr = last;
+        } else {
             pt->first_instr = next_instr; /* Update cache to first instr. */
+            instr_t *first_non_label = next_instr;
+            while (instr_is_label(first_non_label) && first_non_label != end_instr) {
```
derekbruening: If it were relatively simple to have the exact instrs passed to the case analysis phase be the ones in the insertion phase too, that would solve multiple problems:

* The client could implement utilities like this is_first_nonlabel itself by recording instr pointers in the case analysis phase.  As is, any time we add some drmgr utility, we might have to go copy it over here, which is not ideal.

* The label problem mentioned in the issue would be naturally solved.

You'd have to cut out each case sequence from the enclosing instrlist, cap it with NULL, and then stitch it back in later.  I'm not sure how much more complex that is than the current cloning of the case instrlist?

johnfxgalea: Just saw this now... I took a different approach - see https://github.com/DynamoRIO/dynamorio/blob/7da96020a382ef8c6516c641dd83f72544e098e2/ext/drbbdup/drbbdup.h#L315

Basically, I provide the entire working instrlist (containing all dups), but the user has to iterate over the basic block using a special API routine called drbbdup_get_next_analysis_instr().

The drbbdup_analyze_case_t call-back is also modified to take the where instruction. Note that we cannot duplicate special instr, like sys calls. (These need to be left at the end - hence the passing of where)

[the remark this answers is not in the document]
johnfxgalea: Yeah, so these are intricacies, such as the endpoint, that are specific to drbbdup. I don't think I can simply use  instrlist_first_non_label() here.

ext/drbbdup/drbbdup.dox
```diff
@@ -127,9 +127,9 @@ considers for instrumentation, it also provides a "where" instruction.
 The client must insert code exactly prior to the "where" instruction
 in order to ensure correct instrumentation.
 
-drbbdup also provides drbbdup_is_first_instr() and drbbdup_is_last_instr()
-to determine whether the passed instruction is the first
-or last instruction of a basic block copy respectively.
+drbbdup also provides drbbdup_is_first_instr() drbbdup_is_first_nonlabel_instr()
```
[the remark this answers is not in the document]
johnfxgalea: Done

ext/drbbdup/drbbdup.dox
johnfxgalea: Done.

ext/drbbdup/drbbdup.h
```diff
@@ -349,6 +349,18 @@ DR_EXPORT
 drbbdup_status_t
 drbbdup_is_first_instr(void *drcontext, instr_t *instr, OUT bool *is_start);
 
+DR_EXPORT
+/**
+ * Indicates whether the instruction \p instr is the first non labe; instruction of
```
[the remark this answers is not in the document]
johnfxgalea: Ty - Done

ext/drbbdup/drbbdup.dox
johnfxgalea: Done.

ext/drbbdup/drbbdup.c
johnfxgalea: I have another PR to handle the case analysis issue and won't be tackling it here. Not marking as resolved just in case you didn't yet read my previous comment.

johnfxgalea:
PTAL

ext/drbbdup/drbbdup.dox
johnfxgalea: Done.

ext/drbbdup/drbbdup.c
derekbruening: OK, if you have a link to the other PR we can move this discussion there.

> Basically, I provide the entire working instrlist (containing all dups), but the user has to iterate over the basic block using a special API routine called drbbdup_get_next_analysis_instr().

Hmm, requiring a different iterator raises complications: drcachesim has instr_offline_t::identify_elidable_addresses() which walks an instrlist.  It is called both online and offline during post-processing.  The offline decodes from the binary file to create the instrlist and is using DR standalone mode, so it would not be using drbbdup or at least not in a regular setup.  I.e., we have the same code walking an instrlist in an analysis phase and offline.

> The drbbdup_analyze_case_t call-back is also modified to take the where instruction. 

Not sure I follow what "where" means here.

Issue #4219, opened 2020-03-25
 Translation problems in clean call mangling causing post-detach crashes


derekbruening:
While testing my new drwrap post-call method for #4197 I wrote a drwrap detach test and ended up hitting a bunch of failures running it in a loop locally.

The visible symptoms are that right after detach we get the "Cannot correctly handle a received signal.>" fatal error, because a SIGSEGV arrived in a now-native thread for which we have no setup to deliver a signal as we're nearly all the way shut down.

The translate code for clean call mangling is only thinking about synchronous faults, which would only happen on argument setup. At that point, there is indeed a clean app state stored. However, asynchronous translation, such as for detach, could happen anywhere. Plus, there are clean call optimizations that do not store all the state. There is also clean call inlining. It does not look like the translation code is properly considering all of that.

It will take some effort to go through and support all that complexity. As a first step I'm planning to just refuse to translate at all when in clean call mangling.  If I put in a change to just refuse to translate in any clean call mangling, it fixes the failures in the drwrap detach test.  Although since the args could have a synchronous fault we'll have to translate there, so it will take a little more effort.

I also want to put in some stats on how many asynch translations are refused, to help inform whether we could speed up detach by supporting more translation points.

derekbruening:
I split off #4232 to cover performance.  This issue will focus on correctness.

Issue #2971, opened 2018-04-26
drop support for non-CLIENT_INTERFACE builds


derekbruening:
Maintaining non-CLIENT_INTERFACE builds is a minor pain point and the define complicates the code base.  We have put all the performance-affecting pieces of the client interface into runtime options, such that a CLIENT_INTERFACE build without a client should be as performant as a non-CLIENT_INTERFACE build.  There seems to be little reason to keep the define around given that these days there are no major users (that we know of) of other configs.

derekbruening:
The title is now complete but there are some related defines we should probably also clean up:

APP_EXPORTS
DYNAMORIO_IR_EXPORTS
CLIENT_SIDELINE

derekbruening:
And CUSTOM_TRACES

derekbruening:
Having second thoughts on removing APP_EXPORTS: for running untrusted apps, such as for malware analysis, it might be nice to be able to disable DR listening to the app by disabling APP_EXPORTS (and thus DR_APP_EXPORTS) and ANNOTATIONS.  OTOH we don't currently test such a combination, it makes some parts of the code a little ugly, and to add support we should add a CI target.  Maybe a runtime option could be used instead which is easier to maintain.  Since it will take work to remove it (and we can't remove it 100% b/c it's used to use dr_app.h as both provider and consumer) I'm probably going to punt and leave it in there for now.

I filed #4819 on this.

PR #4220, opened 2020-03-25
Add signal and thread numbers to fatal signal message

derekbruening:
Adds the signal number and thread id to this fatal message for better diagnostics:
  <Cannot correctly handle received signal 11 in thread 204511.>

Tested in the #4219 failures.


derekbruening:
Grr the failure is the obsolete vps build -- #2971 covers removing that but that will take some effort to fully clean up so I'll fix the build failure here for now.

Issue #1578, opened 2014-12-30
support detach on ARM


derekbruening:
To determine whether a thread has exited I'm storing a value in the app TLS slot we're using.  To properly detach we'll need to instead restore that value.  This issue covers coming up with some other solution to the problem of figuring out whether a thread has exited, or else delaying the detach value restore.

derekbruening:
If we only support detach for -private_loader we could put a flag in the
client TLS area, which we allocate in the same mmap but place prior to the
TCB?  I guess the problem is static DR, which is a major use case.

derekbruening:
The other detach issue I'm seeing is a crash due to not setting xsp in notify_and_jmp_without_stack.  There's even a comment there that "xsp is only set for X86" but no reason is given.  We have to set it to get the frame for sigreturn at the right spot: not really understanding that comment when it doesn't have a TODO or FIXME.

derekbruening:
After fixing the setting of xsp, I'm seeing something strange: immediately past the sigreturn everything is good, but as soon as we execute the SYS_futex we returned to for __pthread_cond_wait, the thread state is messed up.  This causes a SIGSEGV natively, and under gdb it messes up gdb such that it's unusable:

```
(gdb) x/4i 0xffffb7a882c8
   0xffffb7a882c8 <__pthread_cond_wait+312>:	svc	#0x0
   0xffffb7a882cc <__pthread_cond_wait+316>:	mov	w19, #0x0                   	// #0
   0xffffb7a882d0 <__pthread_cond_wait+320>:	ldr	w0, [x29,#160]
   0xffffb7a882d4 <__pthread_cond_wait+324>:	bl	0xffffb7a8b3c8 <__pthread_disable_asynccancel>
(gdb) x/4gx 0xffffb792f8d0
0xffffb792f8d0:	0x0000ffffb792f990	0x0000aaaaaaaac244
0xffffb792f8e0:	0x0000ffffb7930200	0x0000fffffffff3c8
(gdb) b * 0xffffb7a882c8
Breakpoint 2 at 0xffffb7a882c8: file pthread_cond_wait.c, line 186.
(gdb) c
Continuing.

Thread 2 "api.detach" hit Breakpoint 2, 0x0000ffffb7a882c8 in __pthread_cond_wait (cond=0xaaaaaaac7080, mutex=0xaaaaaaac70b0) at pthread_cond_wait.c:186
186	pthread_cond_wait.c: No such file or directory.
1: x/i $pc
[...]

PR #4221, opened 2020-03-25
i#4197: Add new drwrap post-call scheme: replace retaddr

### LINKED ISSUE #4197: Add drwrap mode with indirect handling but no expensive retaddr checking or flushing
Today's drwrap instruments the callee entry point and examines the return address to find the post-call instrumentation point, recording it and flushing if necessary for instrumentation.  However, flushing is very expensive, and we have found that even the check of the return address is too expensive as it uses shared data structures and locks.

One potential solution is call-site-only instrumentation #4070 but that misses too many post-calls.

Our CGO 2011 paper listed 4 approaches:

1. Add an extra frame on the app call stack (what Valgrind does; what Pin used to do): not transparent but might be fine for our apps
2. Find all the RET instructions in the callee (what Pin does nowadays): suffers from complexity of static analysis of hot/cold split function walking
3. Call site interception (#4070): suffers from the many cases of statically indeterminable call targets and tailcalls
4. Dynamically find return addresses from inside the callee (what DR does today).

Adding another one:

5. Change the return address to a sentinel/takeover trampoline. This also has transparency implications and can (and does) break applications, which look at or modify their own return addresses on occasion. Such things are normally only done as a last resort when you're not in control. But it might work fine here. It could be done purely in a client: use a legitimate address; have a high-priority bb event look for this address and use dr_redirect_native_target() for an app-instr-less transfer (dr_redirect_execution() is not supported in such a situation; the native redirect is better). It would end up as a DR-trace barrier.

For target applications that are relatively well-behaved, the non-transparent solutions 1 and 4 may well be the most effective in terms of performance and completeness.  This issue covers implementing one of those as an optional feature in drwrap.

derekbruening:
Adds an alternative scheme for achieving a post-call control point
that does not require flushing or shared data structure examination
per-call: replacing the return address with a sentinel.

When the new flag DRWRAP_REPLACE_RETADDR is set, the return address is
replaced with the address of a single return instruction in the client
library, with the real address saved.  When a block is seen consisting
of that sentinel instruction, post-call callbacks are called, and then
control is sent to the saved real address using
dr_redirect_native_target().

Adds wrapping tests to drwrap-test.

This new scheme requires restoring return addresses on the stack on
detach or other state translation.  Adds functionality to do so, along
with a new test client.drwrap-test-detach.

This requires the client's state restoration event be called for
addresses not in the code cache.  Adds such a call.

Adds comments about translation problems with clean call mangling
which is filed as i#4219.  The issues seen here are all limited to
traces, so the test works around the problems with -disable_traces.

Tested the core drwrap behavior on ARM and AArch64 but missing general
detach support there (#1578) prevents enabling the detach test there.

Issue: #4219
Fixes #4197

### DIFF
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -255,6 +255,9 @@ Further non-compatibility-affecting changes include:
  - Added drwrap_get_stats().
  - Added #DRWRAP_NO_DYNAMIC_RETADDRS for reducing drwrap overhead at the cost
    of missing some post-call callbacks.
+ - Added #DRWRAP_REPLACE_RETADDR for an alternative method of setting up post-call
+   control points by replacing return addresses.  This does not work for every
+   application, but reduces overhead.
  - Added -record_dynsym_only to drcachesim for faster function tracing symbol
    lookups when internal symbols are not needed.
  - Added dr_merge_arith_flags() as a convenience routine to merge arithmetic flags
--- a/core/synch.c
+++ b/core/synch.c
@@ -447,6 +447,27 @@ translate_mcontext(thread_record_t *trec, priv_mcontext_t *mcontext, bool restor
                 dump_mcontext(get_mcontext(trec->dcontext), THREAD_GET, DUMP_NOT_XML);
             });
             *mcontext = *get_mcontext(trec->dcontext);
+#ifdef CLIENT_INTERFACE
+            if (dr_xl8_hook_exists()) {
+                /* The client may need to translate here if it's using sentinel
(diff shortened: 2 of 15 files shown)


derekbruening:
Failures are 32-bit build missing a define: will fix now.

derekbruening:
The Travis Mac job is terminating mid-build.  It is strange.  It looks like another Travis issue.  I filed #4223.  There have been transient Travis Mac problems in the past.  Hopefully this will go away on its own.

PR #4228, opened 2020-03-26
i#4219 xl8: Add an rstat on synch xl8 failures

derekbruening:
Adds an rstat "synchs_not_at_safe_spot" measuring the count of
synch_with_thread attempts that had to be re-tried due to the target
being suspended at an un-translatable (i.e., unsafe) spot.

Adds a corresponding dr_stats_t field "synchs_not_at_safe_spot" for
extraction.

These will be useful for measuring the impact of clean calls and other
un-translatable places on detach and other synch operations.

Issue: #4219


derekbruening:
Appveyor is a test that appears flaky, under #4058

PR #4230, opened 2020-03-27
i#4197 replace RA: Call client restore event for all cases

derekbruening:
Adds invocation of the client restore state event for all cases, even
when the register state is known to already contain application
values, since the client may need to restore memory.

Adds a test to drwrap-test-detach by adding a system call.

Issue: #4197

### DIFF
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -268,6 +268,9 @@ Further non-compatibility-affecting changes include:
    handler.
  - Added atomics for safe and visible aligned loads and stores on all platforms:
    dr_atomic_load32(), dr_atomic_store32(), dr_atomic_load64() dr_atomic_store64().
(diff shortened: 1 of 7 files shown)


derekbruening:
(With the system call in the test it is easy to reproduce a crash without this fix)

derekbruening:
The Appveyor failure is the 32-bit drwrap-test.  It is strange: on the 2nd library load nearly all the wrappings (not just those related to replace-retaddr) seem to disappear.  It seems highly unlikely to be related to this PR.  I can't reproduce it locally.  I'm trying to figure out how it could happen, and I'm re-running the tests here.  If they pass I'll file an issue and consider it separate from this PR.

derekbruening:
Did not reproduce.  Filed #4231.

PR #4233, opened 2020-03-28
i#4226: Faster fragment deletion event

### LINKED ISSUE #4226: fragment deletion event is slow: it takes 1.5 seconds or more on detach on large apps
While adding function tracing to drcachesim (#3048) I was consistently seeing a big slowdown on detach: it used to take 0.3s to detach from 600-ish threads and fully clean up DR, but with function tracing it suddenly took 1.7s.  I tracked it down to just initializing drwrap, even without wrapping anything!  It turns out it's the fragment delete event.  If it's enabled, at cleanup DR walks every fragment (500K for this app's attach-detach window), does individual cleanup, and calls the client callback.

For now I'm removing the drwrap event as it happens to not be necessary.

For the future we should try to remove all the individual fragment deletion work and limit it to only calling the client callback, which should shrink the time taken.

derekbruening:
i#4226: Faster fragment deletion event

Adds a dedicated loop to call the fragment deletion event without
calling all the other fragment freeing code.  This speeds up detach
and process exit in general by 2x.

Tested performance and correctness on a third-party app; also tested
correctness on a hacked drwrap-detach-test to ensure the same number
of deletion event calls were made.  I don't think it's easy to make a
performance regression test unfortunately.

Fixes #4226

### DIFF
--- a/core/fragment.c
+++ b/core/fragment.c
@@ -1139,6 +1139,18 @@ hashtable_fragment_reset(dcontext_t *dcontext, fragment_table_t *table)
         if (!dynamo_exited && !dynamo_resetting)
             ASSERT_TABLE_SYNCHRONIZED(table, WRITE);
     });
+#    if !defined(DEBUG) && defined(CLIENT_INTERFACE)
+    if (!dr_fragment_deleted_hook_exists())
+        return;
+    /* i#4226: Avoid the slow deletion code and just invoke the event. */
+    for (i = 0; i < table->capacity; i++) {
+        f = table->table[i];
+        if (!REAL_FRAGMENT(f))
+            continue;
+        instrument_fragment_deleted(dcontext, f->tag, f->flags);
+    }
+    return;
+#    endif
     /* Go in reverse order (for efficiency) since using
      * hashtable_fragment_remove_helper to keep all reachable, which is required
      * for dynamo_resetting where we unlink fragments here and need to be able to
@@ -1159,11 +1171,6 @@ hashtable_fragment_reset(dcontext_t *dcontext, fragment_table_t *table)
              * the assert.
              */
             ASSERT(!TEST(FRAG_TRACE_BUILDING, f->flags));
-#    if !defined(DEBUG) && defined(CLIENT_INTERFACE)
-            if (dr_fragment_deleted_hook_exists() && !REAL_FRAGMENT(f))
-                instrument_fragment_deleted(dcontext, f->tag, f->flags);
-            continue;
-#    endif
             hashtable_fragment_remove_helper(table, i, &table->table[i]);
             if (!REAL_FRAGMENT(f))
                 continue;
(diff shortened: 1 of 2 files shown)


core/fragment.c
```diff
@@ -1139,6 +1139,18 @@ hashtable_fragment_reset(dcontext_t *dcontext, fragment_table_t *table)
...
+    /* i#4226: Avoid the slow deletion code and just invoke the event. */
+    for (i = 0; i < table->capacity; i++) {
+        f = table->table[i];
+        if (!REAL_FRAGMENT(f))
+            continue;
+        instrument_fragment_deleted(dcontext, f->tag, f->flags);
+    }
+    return;
+#    endif
     /* Go in reverse order (for efficiency) since using
```
johnfxgalea: Most likely yes, but I am wondering whether or not we need to keep the remaining debug code given that it is not reachable anymore on release?

core/fragment.c
```diff
@@ -1139,6 +1139,18 @@ hashtable_fragment_reset(dcontext_t *dcontext, fragment_table_t *table)
...
     });
+#    if !defined(DEBUG) && defined(CLIENT_INTERFACE)
+    if (!dr_fragment_deleted_hook_exists())
+        return;
+    /* i#4226: Avoid the slow deletion code and just invoke the event. */
+    for (i = 0; i < table->capacity; i++) {
+        f = table->table[i];
+        if (!REAL_FRAGMENT(f))
+            continue;
+        instrument_fragment_deleted(dcontext, f->tag, f->flags);
```
johnfxgalea: Nit: Can you explain how does DR know that all fragments can be correctly passed to  instrument_fragment_deleted please? When I looked at how instrument_fragment_deleted is used, particularly in fragment_delete, I see it is conditioned:

```
#ifdef CLIENT_INTERFACE
    if (dr_fragment_deleted_hook_exists() &&
        (!TEST(FRAGDEL_NO_HEAP, actions) || !TEST(FRAGDEL_NO_FCACHE, actions)))
        instrument_fragment_deleted(dcontext, f->tag, f->flags);
#endif
```

derekbruening: It's this function that decides those `actions` flags, and the fragment_delete() call a few lines below is not passing either of those NO_HEAP or NO_FCACHE flags, so we know the condition is always true for fragment deletion via hashtable_fragment_reset().  Those NO_ calls are particular types of cleanup done at other times, not at reset/exit.  I can add a comment.

core/fragment.c
derekbruening: I'm not following: the debug code is executed in debug build and helps with finding memory leaks, so I'm not seeing why it would be removed?

johnfxgalea: Thanks for clearing that up. Just wanted to make sure that the debug code is *still* relevant given the quick return now present on release builds.

core/fragment.c
johnfxgalea: Ah I see, yes I agree.

core/fragment.c
derekbruening: Note that in release builds without any fragment deletion event, this function is not even called at all.  Most of the individual cleanup is skipped and we just munmap from a higher level (like an arena dealloc).

core/fragment.c
derekbruening: I guess the concern would be some future mismatch in conditions between what are now two separate calls to instrument_fragment_deleted.  I can add a comment to the other call as well to keep them in synch just so future devs are aware of the two calls.

Issue #4137, opened 2020-02-26
`dr_get_process_id` doesn't return the expected result in `thread_exit_event` callbacks 


petrochenkov:
Citing the [documentation](https://dynamorio.org/dynamorio_docs/dr__events_8h.html#a0b0c102bb9a99d67c3e62ea8b3790ff1) of `void dr_register_thread_exit_event	(	void(*)(void *drcontext) 	func	)	
`:
>The passed-in drcontext should be used instead of calling dr_get_current_drcontext(), as the thread exit event may be invoked from other threads, and using dr_get_current_drcontext() can result in failure to clean up the right resources, and at process exit time it may return NULL.

The `dr_get_process_id` API doesn't take a `drcontext` parameter, so it's effectively using "the current context" when being called from thread exit callbacks, which is not recommended.

This can lead to pretty confusing results when DynamoRIO reports different process IDs for the same thread ID at different points.

Possible solution: introduce a version of `dr_get_process_id` taking a `void* drcontext` parameter that would work exactly like `thread_id_t dr_get_thread_id	(	void * 	drcontext	)`, but would return `dcontext->owning_process` rather than `dcontext->owning_thread`.

derekbruening:
Looking at the code, on UNIX, `dr_get_process_id` calls `get_process_id` which calls `dynamorio_syscall(SYS_getpid, 0)`.  I.e., it makes a system call and will return the same value from any thread in the process (though note that a client thread is deliberately in a separate thread group).  On Windows it does indeed use the current TEB, but the ClientId.UniqueProcess field it accesses should be identical for all threads in the process.  AFAICT the code is not using `dcontext->owning_process` here.

So I'm not seeing how `dr_get_process_id` could return anything unusual.  Please provide details on how you observed different pids and their values: OS, app, etc.

petrochenkov:
I'll try to prepare a more detailed reproduction.

The ID mismatch that I observed could be specific to threads killed by `dynamo_other_thread_exit` (e.g. in `cleanup_after_vfork_execve`).

derekbruening:
Right, ok, so there are ways (on UNIX) the app can create a new pid where DR will still treat it as a thread, and we'll have the behavior you described; and the SYS_getpid is indeed in the executing context and not the target thread's context.

So the proposal is a new API routine that would be used in thread exit events?  OK, happy to take a PR for that, hopefully along with a test.

petrochenkov:
[...]

PR #4234, opened 2020-03-28
i#4137: add API for getting a process id associated with a drcontext

petrochenkov:
Fixes: #4137

### DIFF
--- a/core/lib/instrument.c
+++ b/core/lib/instrument.c
@@ -2791,12 +2791,13 @@ dr_get_process_id(void)
 }
 
 DR_API process_id_t
-dr_get_process_id_ext(void *drcontext)
+dr_get_process_id_from_drcontext(void *drcontext)
 {
     dcontext_t *dcontext = (dcontext_t *)drcontext;
-    CLIENT_ASSERT(drcontext != NULL, "dr_get_process_id_ext: drcontext cannot be NULL");
+    CLIENT_ASSERT(drcontext != NULL,
+                  "dr_get_process_id_from_drcontext: drcontext cannot be NULL");
     CLIENT_ASSERT(drcontext != GLOBAL_DCONTEXT,
-                  "dr_get_process_id_ext: drcontext is invalid");
+                  "dr_get_process_id_from_drcontext: drcontext is invalid");
 #    ifdef UNIX
     return dcontext->owning_process;
 #    else
--- a/core/lib/instrument.c
+++ b/core/lib/instrument.c
@@ -2790,6 +2790,20 @@ dr_get_process_id(void)
     return (process_id_t)get_process_id();
 }
 
+DR_API process_id_t
+dr_get_process_id_ext(void *drcontext)
+{
+    dcontext_t *dcontext = (dcontext_t *)drcontext;
+    CLIENT_ASSERT(drcontext != NULL, "dr_get_process_id_ext: drcontext cannot be NULL");
+    CLIENT_ASSERT(drcontext != GLOBAL_DCONTEXT,
+                  "dr_get_process_id_ext: drcontext is invalid");
+#    ifdef UNIX
+    return dcontext->owning_process;
+#    else
+    return dr_get_process_id();
+#    endif
+}
+
 #    ifdef UNIX
 DR_API
 process_id_t
--- a/core/lib/instrument_api.h
+++ b/core/lib/instrument_api.h
@@ -796,7 +796,7 @@ DR_API
  * Clients may want to avoid touching resources shared between processes,
  * like files, from the post-fork execution of the callback. The post-fork
  * version of the callback can be recognized by dr_get_process_id()
- * returning a different value than dr_get_process_id_from_drcontext(drcontext).
+ * returning a different value than dr_get_process_id_from_drcontext().
  *
  * See dr_set_process_exit_behavior() for options controlling performance
  * and whether thread exit events are invoked at process exit time in
@@ -1812,7 +1812,8 @@ process_id_t
 dr_get_process_id(void);
 
 DR_API
-/** Returns the process id of the process associated with drcontext \p drcontext.
+/**
+ * Returns the process id of the process associated with drcontext \p drcontext.
  * The returned value may be different from dr_get_process_id() if the passed context
  * was created in a different process, which may happen in thread exit callbacks.
  */
--- a/core/lib/instrument_api.h
+++ b/core/lib/instrument_api.h
@@ -796,7 +796,7 @@ DR_API
  * Clients may want to avoid touching resources shared between processes,
  * like files, from the post-fork execution of the callback. The post-fork
  * version of the callback can be recognized by dr_get_process_id()
- * returning a different value than dr_get_process_id_ext(drcontext).
+ * returning a different value than dr_get_process_id_from_drcontext(drcontext).
  *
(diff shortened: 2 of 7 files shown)


core/lib/instrument.c
```diff
@@ -2790,6 +2790,20 @@ dr_get_process_id(void)
     return (process_id_t)get_process_id();
 }
 
+DR_API process_id_t
+dr_get_process_id_ext(void *drcontext)
```
petrochenkov: Not sure about the name, I used it only because I've seen other functions instrument_api using the `foo_ext` naming when they take extra arguments compared to `foo`. Other suggestions are welcome.

core/lib/instrument.c
```diff
@@ -2790,6 +2790,20 @@ dr_get_process_id(void)
...
+{
+    dcontext_t *dcontext = (dcontext_t *)drcontext;
+    CLIENT_ASSERT(drcontext != NULL, "dr_get_process_id_ext: drcontext cannot be NULL");
+    CLIENT_ASSERT(drcontext != GLOBAL_DCONTEXT,
+                  "dr_get_process_id_ext: drcontext is invalid");
+#    ifdef UNIX
+    return dcontext->owning_process;
+#    else
+    return dr_get_process_id();
```
petrochenkov: Alternatively, the `owning_process` field can be added to `dcontext_t` on Windows too.
Not sure whether it's possible for it to be different from `dr_get_process_id()` there.

petrochenkov:
Regarding testing - I haven't found an existing test to which I could add the process ID check described in https://github.com/DynamoRIO/dynamorio/pull/4172.
Should I add a new test program / test client pair for this (to `suite/tests/client-interface`, I guess), or it would be an overkill?

derekbruening:
> Regarding testing - I haven't found an existing test to which I could add the process ID check described in #4172.
> Should I add a new test program / test client pair for this (to `suite/tests/client-interface`, I guess), or it would be an overkill?

A new test sounds good.

core/lib/instrument.c
derekbruening: Actually they do not use `foo_ext`, they use `foo_ex`.  But this seems different. `_ex` is generally adding extra functionality or optional extra behavior on top of the same base behavior.  This is changing the base behavior.  I would suggest `dr_get_process_id_from_drcontext`.  It's longer but descriptive and clear IMHO.

core/lib/instrument.c
derekbruening: AFAIK it is not possible on Windows: the process vs thread distinction is much stronger.

core/lib/instrument.c
petrochenkov: Done.

petrochenkov:
The test is added.

core/lib/instrument_api.h
```diff
@@ -796,8 +796,7 @@ DR_API
  * Clients may want to avoid touching resources shared between processes,
  * like files, from the post-fork execution of the callback. The post-fork
  * version of the callback can be recognized by dr_get_process_id()
- * returning a different value than it returned during the corresponding
- * thread init event.
+ * returning a different value than dr_get_process_id_from_drcontext(drcontext).
```
derekbruening: Does Doxygen turn it into a link if it has a parameter name?  It definitely makes links for empty parens which is why we always have those for function references in Doxygen comments.  Maybe quickly check your api/docs/html/ output to confirm; if no link then remove the param name.

core/lib/instrument_api.h
```diff
@@ -1812,6 +1811,14 @@ DR_API
 process_id_t
 dr_get_process_id(void);
 
+DR_API
+/** Returns the process id of the process associated with drcontext \p drcontext.
```
derekbruening: style: newline after `/**` (see https://github.com/DynamoRIO/dynamorio/wiki/Code-Style#commenting-conventions number 4; see also the surrounding code)

suite/tests/client-interface/process-id.c
```diff
@@ -0,0 +1,25 @@
+#include <pthread.h>
```
derekbruening: Please include the license + copyright header

suite/tests/client-interface/process-id.dll.c
```diff
@@ -0,0 +1,14 @@
+#include "dr_api.h"
```
derekbruening: Please include the license + copyright header

core/lib/instrument_api.h
petrochenkov: Is there a way to check for things lik this automatically?
I run `clang-format`, but it didn't change this comment.
(I also see there are `vera++` checks, but they seem to fail on code unrelated to my changes if enabled.)

core/lib/instrument_api.h
petrochenkov: The parameter name is removed.

core/lib/instrument_api.h
petrochenkov: Done.

suite/tests/client-interface/process-id.c
petrochenkov: Done.

suite/tests/client-interface/process-id.dll.c
petrochenkov: Done.

core/lib/instrument_api.h
derekbruening: `vera++` is enabled on Travis and my local machine and there are no failures there -- what failure do you see (unsure how it would be different: never seen vera discrepancies before)?

derekbruening: > Is there a way to check for things lik this automatically?
> I run `clang-format`, but it didn't change this comment.

Unfortunately there is no clang-format rule available for that particular style AFAIK.  A `vera++` one could perhaps be written.  The style choices were made before clang-format existed, and already some were dropped or switched for better automated enforcement, though this one is still there.

derekbruening:
run arm tests

derekbruening:
I forgot before: could you add a line to the feature changelist in api/docs/release.dox (in the non-compat list) about adding this function?

core/lib/instrument_api.h
petrochenkov: >what failure do you see (unsure how it would be different: never seen vera discrepancies before)?

Turns out it starts checking CMake-generated files in the build directory and fails.
```
[  0%] Checking style with vera++ in build/CMakeFiles/3.10.2/CompilerIdC
/home/we/dynamorio/build/CMakeFiles/3.10.2/CompilerIdC/CMakeCCompilerId.c:217: error line_length_content: Line too long: exceeds 90 chars.
/home/we/dynamorio/build/CMakeFiles/3.10.2/CompilerIdC/CMakeCCompilerId.c:339: error line_length_content: Line too long: exceeds 90 chars.
```
This doesn't happen with out-of-source builds, but I've been building exactly like the [wiki](https://github.com/DynamoRIO/dynamorio/wiki/How-To-Build#linux) says.

derekbruening: I see, so some glob somewhere looks inside the build dir if it's a child of the top-level source.

petrochenkov:
>could you add a line to the feature changelist in api/docs/release.dox (in the non-compat list) about adding this function?

Done.

derekbruening:
run arm tests

derekbruening:
Thank you for the patch!

Issue #4031, opened 2020-01-20
ASSERT Failing asserts in vmareas.c on zero-length munmap


petrochenkov:
If `munmap` syscall has an argument `len == 0`, then an empty memory range (`start == end`) is passed to `vmareas.c`'s `binary_search` and `add_vm_area` from `pre_system_call` and `post_system_call` respectively.

Those functions have asserts checking that the passed range is well-formed (`ASSERT(start < end || end == NULL /* wraparound */);` or `ASSERT(start < end);`) and fire on empty ranges.

Zero-length `munmap`s should always fail with `EINVAL`, but for some reason they still happen in Geekbench for Android.

petrochenkov:
Looks like there are two alternative fixes:
- Relax the asserts to `start <= end`.
It *seems* like the binary search logic in `binary_search` and `add_vm_area` works naturally for empty ranges, but I'm not sure.
- Special-case `len == 0` in `{pre,post}_system_call` for `munmap`. Perhaps we don't need to do any work with vm areas in this case and can just return early.

Any guidance on selecting the right fix is appreciated.

derekbruening:
Either way seems fine to me -- but the first way would want to add to the unit tests at the bottom of vmareas.c to confirm that everything works, and check other routines that may be called such as app_memory_deallocation().  Maybe the 2nd is better since it is simpler.

Carrotman42:
I vote for the second option, given the assertion in the first post:
"Zero-length munmaps should always fail with EINVAL", which is true
according to the manpage I just read. If they should always be EINVAL, I
figure it makes sense to make that happen sooner than later.

derekbruening:
This assert in binary_search is showing up in Mac64: https://travis-ci.com/DynamoRIO/dynamorio/jobs/287035442.  I can't reproduce it locally so it is not clear that it's the start==end case (it could be something worse).  I'm going to go ahead and put in handling for that in binary_search plus a test; hopefully it will fix this issue, which does not reproduce in PR builds very well either.

derekbruening:
Scratch that: it is not the == case as the assert still fires sometimes: https://travis-ci.com/DynamoRIO/dynamorio/jobs/287105079

derekbruening:
But it turns out that case is hit in the drcachesim.delay-simple test:
```
Hit delay threshold: enabling tracing.
XXXXX hit start==end case
Exiting process after ~12284 references.
```

derekbruening:
[...]

PR #4235, opened 2020-03-30
i#4031: fix assert on zero-length memory range passed to `add_vm_area`

petrochenkov:
Fixes: #4031


petrochenkov:
I'm not sure how to test anything that fails only with Debug build of DynamoRIO, like this assert.
The added test passed in Release mode even before the fix.

suite/tests/linux/zero-length-mem-ranges.c
```diff
@@ -0,0 +1,40 @@
+#define _GNU_SOURCE
```
derekbruening: Please include the copyright + license header

suite/tests/linux/zero-length-mem-ranges.c
```diff
@@ -0,0 +1,40 @@
...
+        return EXIT_FAILURE;
+    }
+
+    int munmap_res = munmap(mem, 0);
+    if (munmap_res != -1) {
+        printf("zero-length munmap succeeded\n");
+        return EXIT_FAILURE;
+    }
+
+    printf("done\n");
```
derekbruening: We always #include "tools.h" and then use `print()` for standardized flushing behavior across platforms and interleaving w/ DR output (no other output here now but the test could be run under a client in the future or sthg).

So s/printf/print/g.

derekbruening:
Please also fix the CI failures (click on the red X) -- looks like DR reports a memory leak when running this test on multiple platforms so probably you can repro it?

derekbruening:
run arm tests

suite/tests/linux/zero-length-mem-ranges.c
petrochenkov: Done.

suite/tests/linux/zero-length-mem-ranges.c
petrochenkov: Done.

petrochenkov:
I moved the zero-length check to a different place (munmap post-syscall) to avoid the memory leak, looks like the assert isn't hit from any other places.

derekbruening:
> I moved the zero-length check to a different place (munmap post-syscall) to avoid the memory leak, looks like the assert isn't hit from any other places.

The leak is in DR's memory so I assume that DR is allocating something, then short-circuiting or exiting due to zero length, and never freeing that something: if you look at the code handling the app code that you changed to avoid the leak and you can spot the bug maybe you could fix it in DR if it's simple?

petrochenkov:
>if you look at the code handling the app code that you changed to avoid the leak and you can spot the bug maybe you could fix it in DR if it's simple?

There's too much happening between `post_system_call(SYS_munmap)` and `add_vm_area`, I looked through the code, but still don't understand most of it, sorry.

There's one suspicious thing though, the first thing we do in post-munmap is `query_memory_ex_from_os(addr, &info);`, which is pretty strange for zero-length ranges because the `addr` is not even in the range in this case.
Perhaps all the further nonsense starts with this arbitrary filled `dr_mem_info_t`? In that case this PR's solution would actually be correct.

petrochenkov:
The AppVeyor failure looks unrelated to this PR.

derekbruening:
> > if you look at the code handling the app code that you changed to avoid the leak and you can spot the bug maybe you could fix it in DR if it's simple?
> 
> There's too much happening between `post_system_call(SYS_munmap)` and `add_vm_area`, I looked through the code, but still don't understand most of it, sorry.
> 
> There's one suspicious thing though, the first thing we do in post-munmap is `query_memory_ex_from_os(addr, &info);`, which is pretty strange for zero-length ranges because the `addr` is not even in the range in this case.
> Perhaps all the further nonsense starts with this arbitrary filled `dr_mem_info_t`? In that case this PR's solution would actually be correct.

So I assume the DR leak is unrelated to this PR (i.e., if you remove the add_vm_area fix here and run the original test (I guess disabling the assert to get to the leak check at the end in debug), the leak is still reported, right?  Assuming so, if you could file a separate issue on it and include the part of the test that triggers it, and then we'll merge this PR for add_vm_area.

petrochenkov:
>i.e., if you remove the add_vm_area fix here and run the original test (I guess disabling the assert to get to the leak check at the end in debug), the leak is still reported, right?

No the leak is not reported in that case.
The leak is indeed related to this PR in its original form (https://github.com/DynamoRIO/dynamorio/pull/4235/commits/ff9f707f38deb54de4504f9a26c6a9566c8f191f).
If the body of `add_vm_area` is skipped, but the rest of code in `post_system_call(SYS_munmap)` is not, then we get the leak, but I don't understand why.
So the current version of this PR skips all the `post_system_call(SYS_munmap)`.

core/unix/os.c
```diff
@@ -8178,6 +8178,8 @@ post_system_call(dcontext_t *dcontext)
         /* We assumed in pre_system_call() that the unmap would succeed
          * and flushed fragments and removed the region from exec areas.
          * If the unmap failed, we re-add the region to exec areas.
+         * For zero-lenght unmaps we don't need to re-add anything,
```
derekbruening, line 8181: spelling: length

derekbruening:
run arm tests

derekbruening:
run arm tests

core/unix/os.c
derekbruening: It let me fix it so I went ahead and did that and will merge after the tests run.
Thanks again for the patch!

derekbruening:
Hmm, Appveyor is the drwap-test-detach in 32-bit: seems to timeout.  Under #4058.

PR #4238, opened 2020-04-02
Cleanup: Change a for loop var to const ref

derekbruening:
Updates a for loop in drcachesim's file reader to use a const
reference.


derekbruening:
It's drwrap-test-detach on 32-bit again: under #4058 b/c I can't reproduce it and it seems specific to the Appveyor env.  May have to disable or mark it flaky.

PR #4244, opened 2020-04-05
i#4237: Align redirected malloc to double-ptr-sized

### LINKED ISSUE #4237: DynamoRIO wrapped malloc should guarantee 16-byte heap alignment
MOVDQA requires 16-byte memory alignment. DynamoRIO malloc does not align by 16 bytes. Unaligned memory access causes a crash. More information can be found in [Google Groups](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/dynamorio-users/BlPaJytSH4o).

derekbruening:
Provides a guaranteed double-pointer-sized alignment for redirected
malloc, to match system allocators and avoid breaking private library
assumptions.

Adds a test to client.alloc.

Fixes #4237

### DIFF
--- a/core/loader_shared.c
+++ b/core/loader_shared.c
@@ -926,27 +926,56 @@ loader_allow_unsafe_static_behavior(void)
 #endif
 }
 
+/* For heap redirection we need to provide alignment beyond what DR's allocator
+ * provides: DR does just pointer-sized alignment (HEAP_ALIGNMENT) while we want
+ * double-pointer-size (STANDARD_HEAP_ALIGNMENT).  We take advantage of knowing
+ * that DR's alloc is either already double-aligned or is just off by one pointer:
+ * so there are only two conditions.  We use the top bit of the size to form our
+ * header word.
+ * XXX i#4243: To support redirected memalign or C++17 aligned operator new we'll
+ * need support for more general alignment.
+ */
+#define REDIRECT_HEADER_SHIFTED (1ULL << IF_X64_ELSE(63, 31))
+
 /* This routine allocates memory from DR's global memory pool.  Unlike
  * dr_global_alloc(), however, we store the size of the allocation in
- * the first few bytes so redirect_free() can retrieve it.  This memory
+ * the first few bytes so redirect_free() can retrieve it.  We also align
+ * to the standard alignment used by most allocators.  This memory
  * is also not guaranteed-reachable.
  */
 void *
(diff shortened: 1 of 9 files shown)


core/loader_shared.c
```diff
@@ -926,27 +926,56 @@ loader_allow_unsafe_static_behavior(void)
...
  * is also not guaranteed-reachable.
  */
 void *
 redirect_malloc(size_t size)
 {
     void *mem;
-    ASSERT(sizeof(size_t) >= HEAP_ALIGNMENT);
-    size += sizeof(size_t);
-    mem = global_heap_alloc(size HEAPACCT(ACCT_LIBDUP));
+    size_t alloc_size = size + sizeof(size_t) + STANDARD_HEAP_ALIGNMENT - HEAP_ALIGNMENT;
```
johnfxgalea, line 950: It is worth point out that alloc_size directly acts as the header here

core/win32/drwinapi/ntdll_redir.c
```diff
@@ -299,26 +299,26 @@ RtlAllocateHeap(HANDLE heap, ULONG flags, SIZE_T size);
 void *
 wrapped_dr_alloc(ULONG flags, SIZE_T size)
 {
-    byte *mem;
-    ASSERT(sizeof(size_t) >= HEAP_ALIGNMENT);
-    size += sizeof(size_t);
-    mem = global_heap_alloc(size HEAPACCT(ACCT_LIBDUP));
+    /* HeapAlloc returns 16-byte-aligned for 64-bit and 8-byte-aligned for 32-bit.
```
johnfxgalea, line 302: Just for my understanding, why wasn't redirect_malloc called prior?

derekbruening: Prior to this PR redirect_malloc was only used on UNIX, and a parallel implementation of adding a size field was used here on Windows, just for historical reasons.  Somehow we never saw any problems from the smaller alignment.  Now that it's more complex to add a layer around DR's allocator I made them share the same impl in redirect_malloc.

derekbruening: "just for historical reasons": I think they each evolved separately and just nobody ever took a step to share the code (it was pretty simple code before).

derekbruening:
The unit_tests failure: have to add a function to return the requested size for Windows size queries.

PR #4249, opened 2020-04-14
i#4134: Modify case analysis interface

johnfxgalea:
Enhances the case analysis call-back interface of drbbdup to enable the user to insert analysis labels.

Issue: #4134


johnfxgalea:
Xref: https://github.com/DynamoRIO/dynamorio/issues/4134#issuecomment-602149387

johnfxgalea:
Let me know if you agree with the interface in order to facilitate your use-case. If so, I'll go ahead and add some tests.

ext/drbbdup/drbbdup.c
```diff
@@ -562,28 +562,80 @@ drbbdup_do_orig_analysis(drbbdup_manager_t *manager, void *drcontext, void *tag,
...
+{
+    if (where == NULL)
+        return NULL;
+
+    instr_t *next_where = instr_get_next(where);
+    instr_t *last = instrlist_last_app(bb);
+    bool is_last_special = drbbdup_is_special_instr(last);
+
+    if (where == NULL || drbbdup_is_at_end(where) ||
+        (drbbdup_is_at_end(next_where) && !is_last_special))
```
derekbruening, line 576: Please add a comment explaining: I'm not following why this needs to check for the current *and* the next being the end label.

ext/drbbdup/drbbdup.h
```diff
@@ -145,10 +149,21 @@ typedef void (*drbbdup_destroy_orig_analysis_t)(void *drcontext, void *user_data
...
+ * Unlike for original basic block analysis, this call-back does not return a single
+ * and isolated basic block copy. Instead, it provides the main instruction list that
+ * contains all basic block duplications and is used internally by drbbdup. This enables
+ * users to directly insert label instructions during their analysis.
+ * In order to iterate only over the instructions pertaining to the case, the call-back
+ * provides \p instr which should be used in conjuction with
+ * drbbdup_get_next_analysis_instr(). Analysis labels pertaining to \p instr should
+ * be inserted prior to \p where. This enables, for different cases, the insertion
+ * of analysis labels for instructions such as syscalls and jumps, which cannot
+ * be duplicated in accordance to DR rules.
```
derekbruening, line 162: As mentioned in an earlier discussion, the drcachesim use cases inserts labels during analysis for optimizing address traces by eliding addresses in the function offline_instru_t::identify_elidable_addresses().  The same analysis code is shared with the post-processor, a standalone app using DR as a library, to reconstruct the address trace and insert the missing addresses.  This makes using a custom iterator not easy: now the code can't be directly shared.

Possible solutions:

1) Use this custom iterator interface, requiring drcachesim to make two versions of its analysis code (can share the loop body; two versions of the outer code, with each calling environment calling its own version).

2) Have drbbdup temporarily NULL out the ends of the case instrlist sequence, and temporarily clone any syscall/cti at the end, allowing normal iteration.  After calling the case analysis cb, drbbdup would then re-stitch the instrlist links, and remove the clone.  I didn't check but I assume there's a meta jump or sthg to go to the shared syscall instr.  Any labels inserted for the syscall would still be in the right place then it would seem.

johnfxgalea:
An alternative approach has been taken to address this feature. Thus, this PR has been closed.

https://github.com/DynamoRIO/dynamorio/pull/4253

Issue #4250, opened 2020-04-15
Dr. client with Triton crashes when solving Z3 formula


xcoldhandsx:
Hi!
I describe a problem we have stocked with while building tracer backend for Triton. First, I share a link [url](https://groups.google.com/forum/#!topic/dynamorio-users/Q0c6poLwTis) to the topic on google groups.
I've made a simple client with Triton library. Here is the code.
```c
#include "dr_api.h"
#include "triton/api.hpp"

triton::API api;
static void
event_exit(void) {
}
static dr_emit_flags_t
    event_basic_block(void *drcontext, void *tag, instrlist_t *bb,
    bool for_trace, bool translating) {
    if (api.isSat(api.getAstContext()->equal(api.getAstContext()->bv(1,8),api.getAstContext()->bv(1,8)))) {
        dr_printf("sat\n");
    }
}

DR_EXPORT void
dr_client_main(client_id_t id, int argc, const char *argv[]) {
    dr_set_client_name("trinton_test", "");
    api.setArchitecture(triton::arch::ARCH_X86_64);
    dr_register_bb_event(event_basic_block);
    dr_register_exit_event(event_exit);
}
```
Client crashes at first call of event_basic_block. When I put this code in dr_client_ main or dr_client_exit everything works fine. Here some details of crash:
Backtrace
```
(gdb) bt
#0  memcpy_mid () at /home/fedotoff/dynamorio/core/arch/x86/memfuncs.asm:110
#1  0x00000000711d9f27 in redirect_realloc (mem=0x7fff34184730, size=2856) at /home/fedotoff/dynamorio/core/loader_shared.c:998
#2  0x00007fff31f01cba in memory::reallocate (p=0x7fff34184738, s=2856) at /home/fedotoff/z3/src/util/memory_manager.cpp:295
...
#38 0x00007fff334b70e1 in Z3_solver_check (c=0x7fff33df2538, s=0x7fff33df1a98) at /home/fedotoff/z3/src/api/api_solver.cpp:609
#39 0x00007ffff7a66971 in triton::engines::solver::Z3Solver::isSat(std::shared_ptr<triton::ast::AbstractNode> const&) const () at /home/fedotoff/dynamorio/core/lib/statsx.h:126
```
Some crash state information:
```
(gdb) disassemble 
Dump of assembler code for function memcpy_mid:
[...]

PR #4252, opened 2020-04-15
i#4237: Fix realloc over-read

### LINKED ISSUE #4237: DynamoRIO wrapped malloc should guarantee 16-byte heap alignment

derekbruening:
Fixes a bug from 228933d #4237 where redirect_realloc() was not
updated to read the new malloc header, resulting in a memcpy that
reads beyond the end of the original allocation, which can crash if
it's at the end of a heap block with a subsequent guard page.

Adds at least some sanity checks of realloc to the alignment test.  I
tried to construct a regression test to reproduce the crash in the
bug, but it is not possible in a straightforward way just using the
public interface as it requires knowing what is beyond an allocation
or how two allocations are arranged and what headers or other bytes
are in between.

Fixes #4237

### DIFF
--- a/core/loader_shared.c
+++ b/core/loader_shared.c
@@ -982,6 +982,21 @@ redirect_malloc(size_t size)
     return (void *)res;
 }
 
+static inline size_t
+redirect_malloc_size_and_start(void *mem, OUT void **start_out)
+{
+    size_t *size_ptr = (size_t *)((ptr_uint_t)mem - sizeof(size_t));
+    size_t size = *((size_t *)size_ptr);
+    void *start = size_ptr;
+    if (TEST(REDIRECT_HEADER_SHIFTED, size)) {
+        start = size_ptr - 1;
+        size &= ~REDIRECT_HEADER_SHIFTED;
+    }
+    if (start_out != NULL)
+        *start_out = start;
+    return size;
+}
+
 /* This routine allocates memory from DR's global memory pool. Unlike
  * dr_global_alloc(), however, we store the size of the allocation in
  * the first few bytes so redirect_free() can retrieve it.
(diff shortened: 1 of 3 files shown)


derekbruening:
Should be #4250: will adjust on merge.

core/loader_shared.c
```diff
@@ -991,11 +1006,18 @@ redirect_realloc(void *mem, size_t size)
...
+         * should be on slowpaths in any case.
+         */
         buf = redirect_malloc(size);
         if (buf != NULL && mem != NULL) {
-            size_t old_size = *((size_t *)((byte *)mem - sizeof(size_t)));
+            void *old_start;
+            size_t old_size = redirect_malloc_size_and_start(mem, &old_start);
             size_t min_size = MIN(old_size, size);
-            memcpy(buf, mem, min_size);
+            memcpy(buf, old_start, min_size);
```
johnfxgalea, line 1020: Why copy from old_start instead of mem? Didn't understand that part.

derekbruening: Thank you, that is an error.  old_start is the DR alloc start, not the privlib start.  Wow, apparently I should not be the one writing this code: keep making errors.  At least my new test caught this one on Travis (passed locally somehow...b/c my 0xcd is DR's free pattern??  Confirmed: changing to 0x77 and it fails locally).

johnfxgalea: This also applies to the size as well. AFAIK redirect_malloc_size_and_start() returns the alloc size no?

derekbruening: Yes, should be redirect_malloc_requested_size().  Adding more to the test to catch that too.

derekbruening:
Flaky test on appveyor: drwrap-test-detach #4058

PR #4253, opened 2020-04-16
i#4134: Support analysis labels in drbbdup

johnfxgalea:
Enhances the case analysis call-back interface of drbbdup to enable the user to insert analysis labels.

The extension extracts the case instruction list from the main basic block and passes it to the user during analysis. After, drbbdup stitches the instruction list back to the main basic block.

A test is also included.

Issue: #4134

### DIFF
--- a/ext/drbbdup/drbbdup.c
+++ b/ext/drbbdup/drbbdup.c
@@ -503,28 +503,19 @@ drbbdup_next_end(instr_t *instr)
  *  Overall, separate instr lists simplify user call-backs.
  *  The returned instr list needs to be destroyed using instrlist_clear_and_destroy().
  */
-static instrlist_t *
-drbbdup_extract_single_bb_copy(void *drcontext, instrlist_t *bb, instr_t *start,
-                               OUT instr_t **next)
+static void
+drbbdup_extract_bb_copy(void *drcontext, instrlist_t *bb, INOUT instr_t **start,
+                               OUT instr_t **remainder)
 {
-    instrlist_t *case_bb = instrlist_create(drcontext);
-
-    ASSERT(start != NULL, "start instruction cannot be NULL");
-    ASSERT(instr_get_note(start) == (void *)DRBBDUP_LABEL_START,
+    ASSERT(start != NULL && *start != NULL, "start instruction cannot be NULL");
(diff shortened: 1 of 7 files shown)


derekbruening:
While this is more complex to implement in drbbdup, and for just the drcachesim use case it's not all that bad to not share the outer loop code and need two functions, I think long-term this approach here will be nicer for users: people will be less likely to forget to use the custom iterator and end up with subtle bugs, and people can share analysis code across drbbdup, drmgr, and standalone contexts.

ext/drbbdup/drbbdup.c
```diff
@@ -504,58 +504,76 @@ drbbdup_next_end(instr_t *instr)
...
+    if (drbbdup_is_special_instr(last_instr)) {
+        instrlist_remove(case_bb, last_instr);
+        instr_destroy(drcontext, last_instr);
+    }
+
+    instrlist_append(case_bb, post);
+    instr_t *instr = instrlist_first(case_bb);
+    instrlist_postinsert(bb, pre, instr);
+
+    instrlist_init(case_bb);
```
johnfxgalea, line 553: Note: I could have used instrlist_append_instrlist() here but for some reason it is not exposed by dr core

suite/tests/client-interface/drbbdup-analysis-test.dll.c
```diff
@@ -0,0 +1,163 @@
...
+/* A Test for the drbbdup extension. In particular, the test inserts analysis
+ * labels during case analysis and checks that these labels persist during the
+ * insertion stage.
+ */
+
+#include "dr_api.h"
+#include "drmgr.h"
+#include "drbbdup.h"
+
+#define TEST_NOTE_VAL (void *)767
```
derekbruening, line 42: This will hit compiler warnings for converting from int to larger-sized pointer: LL suffix should fix.

johnfxgalea: Done.

ext/drbbdup/drbbdup.c
johnfxgalea: I guess i'll leave it how it is for now.

johnfxgalea:
@derekbruening Ty for the review.

Issue #3956, opened 2019-11-20
Multi-threading failure (HANG) on AArch64. ASSERT in utils.c: !lock->owner


AssadHashmi:
We're seeing intermittent hangs with an AArch64 guest binary compiled with OpenMP. All the indications are that it is a multi-threading bug in DR. The hang happens on a DR release build. With the DEBUG build, the following assert fires and exits so doesn't get as far as hanging:
```
<Application /path/to/test_case.exe (44407).  Internal Error: DynamoRIO debug check failure: /path/to/dynamorio/core/utils.c:576 !lock->owner
```

Which happens in:
```
static void
deadlock_avoidance_lock(mutex_t *lock, bool acquired, bool ownable)
{
    if (acquired) {
        . . .
        if (ownable) {
            ASSERT(!lock->owner);
            lock->owner = d_r_get_thread_id();
            lock->owning_dcontext = get_thread_private_dcontext();
        }
        . . .
```
The guest binary is built with armclang and linked to the Arm Performance Libraries on RHEL7.5:
`armclang -fopenmp -armpl=lp64,parallel test_case.c -o test_case.exe`

It fails without clients:
`drrun ./test_case.exe`

These also appear during the `-debug` run:
```
<get_memory_info mismatch! (can happen if os combines entries in /proc/pid/maps) 
        os says: 0x0000fffde4000000-0x0000fffe0c000000 prot=0x00000000           
        cache says: 0x0000fffde4000000-0x0000fffe08000000 prot=0x00000000`

<ran out of stolen fd space>
```

It takes between 3 and 60 runs to get the assert to fire and only *seems* to fail on ThunderX2 machines.

Running with `-loglevel 3` gives the following thread statistics:
```
(Begin) Thread statistics @6735 global, 0 thread fragments (0:05.953):
                      BB fragments targeted by IBL (thread):                 3
                               Fcache exits, total (thread):                 4
                            Fcache exits, from BBs (thread):                 4
             Fcache exits, total indirect branches (thread):                 3
         Fcache exits, non-trace indirect branches (thread):                 3
   Fcache exits, ind target in cache but not table (thread):                 3
             Fcache exits, from BB, ind target ... (thread):                 3
              Fcache exits, BB->BB, ind target ... (thread):                 3
             Fcache exits, dir target not in cache (thread):                 1
                                Special heap units (thread):                 1
                           Peak special heap units (thread):                 1
             Current special heap capacity (bytes) (thread):             65536
[...]

PR #4254, opened 2020-04-17
i#3956 arm locks: Use release-acquire atomics

derekbruening:
Replaces ldxr..stxr with ldaxr..stlxr in the atomic sequences used to
implement DR's mutexes.  DR's mutexes, and its other atomic
operations, assume release-acquire memory ordering.

Issue: #3956


derekbruening:
See the #3956 issue for some real-world tests where this eliminates an assert and crashes that are all lock-related.  It would be good to add some kind of regression test but it may not be easy to make something that reliably reproduces these problems and is small enough to fit in the suite so I'm leaning toward just merging this as-is.

PR #4255, opened 2020-04-17
i#4134: Add some platform independent instructions to drbbdup

johnfxgalea:
Replaces many creation of instructions with XINST_CREATE to improve the platform independence of drbbdup.

Issue: #4134


johnfxgalea:
> I checked for ARM suitability and I think they should all work there: the immediate is using the proper potentially-2-instruction routine, etc.

Great, thanks!

Issue #4257, opened 2020-04-17
check_thread_vm_area consistency checks very slow on AArch64 and OSX


derekbruening:
The app in #3956 and #4256 is very slow in debug build.  The culprit seems to be what was also seen on OSX, where the consistency checks at the bottom of check_thread_vm_area are very slow.  Interrupting the app in gdb multiple times while its running slowly shows this.  Raising the level of that check to 3, to match OSX, makes the slowness go away.

```
^C
Program received signal SIGINT, Interrupt.
0x0000fffff7d45e68 in d_r_sscanf (str=0xfffff7fde520 <buf_scratch+1176> "ffffa8440000-ffffa8450000 rwxp 00000000 00:00 0 ", 
    fmt=0xfffff7f38c90 "%016lx-%016lx %s %016lx %*s %lu %4096[^\n]") at dynamorio/core/io.c:727
727     dynamorio/core/io.c: No such file or directory.
(gdb) bt
#0  0x0000fffff7d45e68 in d_r_sscanf (str=0xfffff7fde520 <buf_scratch+1176> "ffffa8440000-ffffa8450000 rwxp 00000000 00:00 0 ", 
    fmt=0xfffff7f38c90 "%016lx-%016lx %s %016lx %*s %lu %4096[^\n]") at dynamorio/core/io.c:727
#1  0x0000fffff7ea309c in memquery_iterator_next (iter=0xffff2818f770) at dynamorio/core/unix/memquery_linux.c:254
#2  0x0000fffff7ea344c in memquery_from_os (
    pc=0xffffed7b4e38 "\001\034\240N\240\003\332<\270:N\225a\002\300=\002\034\240N \034\241NA\034\242Nc7N\225\210\305\002\360\340#\200=\b\201\070\221\b\375\337\b\223\305\002\360sB4\221\250\213", info=0xffff2818f898, have_type=0xffff2818f85e)
    at dynamorio/core/unix/memquery_linux.c:340
#3  0x0000fffff7e7e7c0 in query_memory_ex_from_os (
    pc=0xffffed7b4e38 "\001\034\240N\240\003\332<\270:N\225a\002\300=\002\034\240N \034\241NA\034\242Nc7N\225\210\305\002\360\340#\200=\b\201\070\221\b\375\337\b\223\305\002\360sB4\221\250\213", info=0xffff2818f898) at dynamorio/core/unix/os.c:9458
[...]

PR #4258, opened 2020-04-17
i#4257 slow memquery: Avoid slow vmarea check on AArch64

derekbruening:
Raises the consistency check at the end of check_thread_vm_area to
default to checklevel 3 on AArch64, matching Mac64 where it is also
observed to be quite slow.

Issue: #4257


derekbruening:
Appveyor failure is client.drwrap-test-detach under #4058

Issue #847, opened 2014-11-28
evaluate whether DR should use gcc's switch to 16-byte stack alignment


derekbruening:
_From [bruen...@google.com](https://code.google.com/u/109494838902877177630/) on July 17, 2012 11:57:04_

gcc 4.5+ have violated the sysv x86 ABI that linux followed which mandates only 4-byte stack alignment, in order to use aligned SSE instrs.  xref http://gcc.gnu.org/bugzilla/show_bug.cgi?id=40838 , http://gcc.gnu.org/bugzilla/show_bug.cgi?id=38496 , http://www.uclibc.org/docs/psABI-i386.pdf .

this adds extra code and extra stack space.  it's not clear such alignment
is a win for DR: we should measure whether aligned SSE instrs outweigh
increased code and stack.  one factor is that we want binary
compatibility w/ older machines, so we can't use really recent ISA
instrs.  if it's not a win we'd pass -mpreferred-stack-boundary=2.

_Original issue: http://code.google.com/p/dynamorio/issues/detail?id=847_

PR #4265, opened 2020-04-20
i#3966: Switch 32-bit x86 Linux to 16-byte stack alignment

derekbruening:
Abandons the 4-byte stack alignment DR was using for 32-bit x86 Linux
(for compatibility with legacy code).  The SysV ABI breakage by gcc
4.5 was 10 years ago and in that time other compilers have followed
suit, making 16-byte the de facto standard.  Clang doesn't even
support requesting 4-byte stack alignment.  DR is now switching to the
modern ABI of 16-byte stack alignment.

Bumps the version to 7.92 and marks OLDEST_COMPATIBLE_VERSION as 791
to indicate the binary compatbility break for 32-bit x86 DR.

Leaves Windows as 4-byte-aligned, since changing the injection
assembly code and other pieces would be a bunch of extra work for
little benefit: Windows 32-bit only requires 4-byte alignment, and the
code is not that much more complex with a split alignment.

Aligns the stack after clean call preparation (state saving) and again
after clean call argument setup.  Currently the stack restores can end
up with two LEA's in a row; we leave removing that for future work.

Adds stack alignment checks to some of the clean call tests.

Updates the manual stack alignment in DR assembly code.  I only found
a few places to update and I expected to find more.  We should monitor
usage outside the test suite and look for problems.

Issue: #847, #3966, #4200
Fixes: #3966
Fixes: #4200

### DIFF
--- a/core/arch/arch.h
+++ b/core/arch/arch.h
@@ -596,11 +596,13 @@ mangle_insert_clone_code(dcontext_t *dcontext, instrlist_t *ilist,
                          instr_t *instr _IF_X86_64(gencode_mode_t mode));
 
 #ifdef X86
-#    if defined(X64) || defined(MACOS)
-#        define ABI_STACK_ALIGNMENT 16
-#    else
-/* See i#847 for discussing the stack alignment on X86. */
+#    ifdef WINDOWS
 #        define ABI_STACK_ALIGNMENT 4
+#    else
+/* See i#847 for discussion of stack alignment on 32-bit Linux.
+ * We now use 16 there to match everyone else.
+ */
+#        define ABI_STACK_ALIGNMENT 16
 #    endif
 #elif defined(AARCH64)
 #    define ABI_STACK_ALIGNMENT 16
@@ -1434,6 +1436,9 @@ move_mm_reg_opcode(bool aligned16, bool aligned32);
  */
 uint
 move_mm_avx512_reg_opcode(bool aligned64);
+
+bool
+clean_call_needs_simd(clean_call_info_t *cci);
 
 /* clean call optimization */
 /* Describes usage of a scratch slot. */
--- a/core/arch/arch.h
+++ b/core/arch/arch.h
@@ -596,13 +596,11 @@ mangle_insert_clone_code(dcontext_t *dcontext, instrlist_t *ilist,
                          instr_t *instr _IF_X86_64(gencode_mode_t mode));
 
 #ifdef X86
-#    ifdef WINDOWS
-#        define ABI_STACK_ALIGNMENT 4
-#    else
-/* See i#847 for discussion of stack alignment on 32-bit Linux.
- * We now use 16 there to match everyone else.
- */
+#    if defined(X64) || defined(UNIX)
+/* See i#847, i#3966 for discussion of stack alignment on 32-bit Linux. */
 #        define ABI_STACK_ALIGNMENT 16
+#    else
+#        define ABI_STACK_ALIGNMENT 4
 #    endif
 #elif defined(AARCH64)
 #    define ABI_STACK_ALIGNMENT 16
--- a/core/arch/arch.h
+++ b/core/arch/arch.h
@@ -600,6 +600,13 @@ mangle_insert_clone_code(dcontext_t *dcontext, instrlist_t *ilist,
 /* See i#847, i#3966 for discussion of stack alignment on 32-bit Linux. */
 #        define ABI_STACK_ALIGNMENT 16
 #    else
+/* We follow the Windows (MSVC-based) 32-bit ABI which requires only 4-byte
+ * stack alignment.
+ * XXX i#4267: Gcc/clang through MinGW/Cygwin use 16-byte by default, but
+ * for interoperating with Windows system libraries (callbacks, e.g.) they
+ * have to hande 4-byte and we expect them to use -mstackrealign or something.
+ * Thus for now we stick with just 4-byte even for them.
+ */
 #        define ABI_STACK_ALIGNMENT 4
 #    endif
 #elif defined(AARCH64)
--- a/core/arch/mangle_shared.c
+++ b/core/arch/mangle_shared.c
@@ -122,6 +122,13 @@ insert_get_mcontext_base(dcontext_t *dcontext, instrlist_t *ilist, instr_t *wher
     }
 }
 
+bool
+clean_call_needs_simd(clean_call_info_t *cci)
+{
+    return (cci->preserve_mcontext || cci->num_simd_skip != proc_num_simd_registers() ||
+            cci->num_opmask_skip != proc_num_opmask_registers());
+}
+
 /* prepare_for and cleanup_after assume that the stack looks the same after
  * the call to the instrumentation routine, since it stores the app state
  * on the stack.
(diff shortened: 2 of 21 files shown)


derekbruening:
Our test suites are all green, though I will not be surprised if there are some lurking issues on asm code somewhere doing manual alignment that I missed and the tests miss.  But what can we do but press onward and try to add more tests if we hit anything in the future.  Certainly this is an improvement over HEAD which crashes on many recent Ubuntu platforms in 32-bit.

derekbruening:
@johnfxgalea PTAL

johnfxgalea:
> Our test suites are all green, though I will not be surprised if there are some lurking issues on asm code somewhere doing manual alignment that I missed and the tests miss. But what can we do but press onward and try to add more tests if we hit anything in the future. Certainly this is an improvement over HEAD which crashes on many recent Ubuntu platforms in 32-bit.

Yea, especially since Windows is still left as 4-byte-aligned. But agreed, we will handle issues as they come by. Review coming up shortly.

core/arch/mangle_shared.c
```diff
@@ -272,24 +279,38 @@ prepare_for_clean_call(dcontext_t *dcontext, clean_call_info_t *cci, instrlist_t
...
+        }
         uint num_slots = DR_NUM_GPR_REGS + NUM_EXTRA_SLOTS;
         if (cci->skip_save_flags)
             num_slots -= 2;
-        num_slots -= cci->num_regs_skip; /* regs that not saved */
+#    ifndef X86_32 /* x86 uses pusha regardless of regs we could skip */
+        num_slots -= cci->num_regs_skip; /* regs not saved */
+#    endif
+        IF_DEBUG(uint emulate_dstack_offs = simd + num_slots * XSP_SZ;)
+        ASSERT(emulate_dstack_offs == dstack_offs || cci->out_of_line_swap);
```
johnfxgalea, line 305: Nit: maybe place the calculation of emulate_dstack_offs, including the num_slot counts, inside a helper function? The counting of num_slots here is only used for debugging (namely the for the ASSERT) because insert_push_all_registers() also does the calculation. Calling a function will make it easier for you to ifdef this code  out and only be executed by debug builds. Plus, you will avoid repetition of code at line 336.

core/arch/mangle_shared.c
```diff
@@ -309,20 +330,33 @@ cleanup_after_clean_call(dcontext_t *dcontext, clean_call_info_t *cci, instrlist
         cci = &default_clean_call_info;
         /* saved error code is currently on the top of the stack */
 
-#if (defined(X86) && defined(X64)) || defined(MACOS)
+#if defined(X86) && (defined(X64) || defined(UNIX))
     /* PR 218790: remove the padding we added for 16-byte rsp alignment */
     if (cci->should_align) {
+        int align = get_ABI_stack_alignment();
```
johnfxgalea, line 336: Repetition of code w.r.t the calculation of emulate_dstack_offs.

suite/tests/client-interface/cleancall.dll.c
```diff
@@ -349,6 +367,10 @@ bb_event(void *drcontext, void *tag, instrlist_t *bb, bool for_trace, bool trans
 #endif
     }
 
+    /* Inlineable call (cleancall-opt does further tests there). */
+    dr_insert_clean_call(drcontext, bb, instrlist_first(bb), inc_counter, false, 1,
```
johnfxgalea, line 371: Not clear what the rationale of this clean call is. What is it trying to test?

johnfxgalea:
LGTM Thank you! Hope this PR wasn't too bad. I guess the piggybacking of code for Mac helped a bit.

core/arch/arch.h
```diff
@@ -596,10 +596,10 @@ mangle_insert_clone_code(dcontext_t *dcontext, instrlist_t *ilist,
                          instr_t *instr _IF_X86_64(gencode_mode_t mode));
 
 #ifdef X86
-#    if defined(X64) || defined(MACOS)
+#    if defined(X64) || defined(UNIX)
+/* See i#847, i#3966 for discussion of stack alignment on 32-bit Linux. */
 #        define ABI_STACK_ALIGNMENT 16
 #    else
-/* See i#847 for discussing the stack alignment on X86. */
 #        define ABI_STACK_ALIGNMENT 4
```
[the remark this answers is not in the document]
derekbruening: The Windows (MSVC-based) 32-bit ABI only mandates 4-byte alignment though so there isn't anything we have to fix for that -- though MinGW or Cygwin libraries built with gcc/clang will have 16-byte so I guess that's the complication.  But such libraries will have problems if they pass a callback to a Windows system library so maybe they're all asking for 4-byte alignment anyway?  A quick search finds stuff like https://github.com/uTox/uTox/issues/1304 which is apparently going to use -mstackrealign to solve this: which, if all MinGW/Cygwin builds do that, means we don't have to do anything.  It is worth filing an issue on this mixed-compiler-alignment problem => #4267.  I'll add a comment pointing to that.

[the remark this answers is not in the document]
johnfxgalea: Ah I see. 

I took your PR message as the existence of pending work:

> Leaves Windows as 4-byte-aligned, since changing the injection
> assembly code and other pieces would be a bunch of extra work for
> little benefit: Windows 32-bit only requires 4-byte alignment, and the
> code is not that much more complex with a split alignment.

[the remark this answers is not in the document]
derekbruening: I will try to reword that.  I did initially try to make Windows 16 too b/c I thought it would make DR's code simpler to have all platforms the same, but it ended up being more work so I reverted.

suite/tests/client_tools.h
```diff
@@ -99,4 +99,29 @@
...
+    __asm__ __volatile__("mov %0, sp" : "=r"(sp));
+#    define STACK_ALIGNMENT 16
+    ASSERT(ALIGNED(sp, STACK_ALIGNMENT));
+#elif defined(ARM)
+    reg_t sp;
+    __asm__ __volatile__("str sp, %0" : "=m"(sp));
+#    define STACK_ALIGNMENT 8
+    ASSERT(ALIGNED(sp, STACK_ALIGNMENT));
+#else
+    /* We could have a separate-file asm routine for Windows but for now we just test
```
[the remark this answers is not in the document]
derekbruening: Done, pointing at #4267.

core/arch/mangle_shared.c
derekbruening: +1, done.

core/arch/mangle_shared.c
derekbruening: Hopefully you mean the code below this.  It is harder to move this line out b/c the local var is used in the condition too.  There is some repetition but with the emulate_dstack_offs shared I think it is pretty good now.

suite/tests/client-interface/cleancall.dll.c
derekbruening: Yeah I was trying to replicate a failure that the inscount sample was hitting, but somehow I couldn't reproduce it here, even with this call which looked identical.  Not sure why, but I fixed it, and we do have sample tests too.  I could just remove it -- sure, will revert this.

core/arch/mangle_shared.c
johnfxgalea: Yes, code concerning emulate_dstack_offs.

PR #4268, opened 2020-04-21
i#4134: Add opnd check 

johnfxgalea:
Makes minor changes to docs to better clarify some assumptions made by drbbdup.

Also adds a check to ensure that the runtime case operand provided by the user is of pointer-size.

Issue: #4134

### DIFF
--- a/ext/drbbdup/drbbdup.c
+++ b/ext/drbbdup/drbbdup.c
@@ -1476,7 +1476,19 @@ static bool
 drbbdup_check_options(drbbdup_options_t *ops_in)
 {
     if (ops_in != NULL && ops_in->set_up_bb_dups != NULL && ops_in->instrument_instr &&
-        ops_in->dup_limit > 0 && opnd_is_memory_reference(ops_in->runtime_case_opnd))
+        ops_in->dup_limit > 0)
+        return true;
+
+    return false;
+}
+
+static bool
+drbbdup_check_case_opnd(opnd_t case_opnd)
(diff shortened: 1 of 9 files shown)


ext/drbbdup/drbbdup.c
```diff
@@ -1476,7 +1476,19 @@ static bool
...
+
+    return false;
+}
+
+static bool
+drbbdup_check_case_opnd(opnd_t case_opnd)
+{
+    /* As stated in the docs, runtime case operand must be a memory reference
+     * and pointer-sized.
+     * */
```
derekbruening, line 1490: style nit: extra `*`

ext/drbbdup/drbbdup.h
```diff
@@ -279,9 +281,9 @@ typedef struct {
      */
     void *user_data;
     /**
-     * The maximum number of cases, excluding the default case, that can be associated
-     * with a basic block. Once the limit is reached and an unhandled case is encountered,
-     * control is directed to the default case.
+     * The maximum number of alternative cases, excluding the default case, that can be
```
derekbruening, line 284: optional: Rename the field to `non_default_case_limit`?

johnfxgalea: Done.

ext/drbbdup/drbbdup.c
johnfxgalea: Done

Issue #2502, opened 2017-07-05
races in ARM lockless data structure reads


derekbruening:
In DR we have some data structures we read without holding a lock, relying on the hardware write visibility.  The design and code was created with x86 in mind and we did not do a thorough enough re-evaluation for ARM and AArch64.  For ARM's memory model we need to add barriers in multiple places to ensure that writes are visible in other threads in the order we require.

Suspect data structures include:
+ The indirect branch lookup (IBL) table: we rely on writing the tag being seen by other threads before the write of the start pc.
+ flushtime_global's store in increment_global_flushtime()
+ likely more...the code needs an audit

derekbruening:
I'm adding a related action item here:

+ Review the "DMB ISH" barriers in arch_exports.h and downgrade them to stores-only or loads-only as necessary.  I based these on g++ 9.2.1's std::atomic generated code and it uses "DMB ISH" for *everything* which apparently is overkill.

algrant-arm:
Probably best to leave anything that's generating explicit barriers as DMB ISH. On Armv8 (including all AArch64) DMB ISHLD should be sufficient as an acquire barrier, but this might not be true on pre-v8 systems. And on v8 it's best to use LDAR anyway.

derekbruening:
There are more variables that may need stronger stores and loads:

- exiting_thread_count

PR #4269, opened 2020-04-21
i#2502 lockless ARM: Fix missing AArch synchronization

derekbruening:
For update_lookuptable_tls: Adds a store-release to the lookuptable
store and a load-acquire to the IBL generated code for both ARM and
AArch64, as there is no lighter-weight way to ensure the mask and
table stores are seen in the proper order.

For flushtime_global: made all readers and the increment use
store-release semantics.

For safely_nullify_tables: leaving as weak stores since timing and
ordering does not matter there.

Issue: #2502

### DIFF
--- a/core/arch/arm/emit_utils.c
+++ b/core/arch/arm/emit_utils.c
@@ -948,6 +948,11 @@ emit_indirect_branch_lookup(dcontext_t *dc, generated_code_t *code, byte *pc,
     APP(&ilist,
         INSTR_CREATE_ldr(dc, OPREG(DR_REG_R1),
                          OPND_TLS_FIELD(TLS_MASK_SLOT(ibl_code->branch_type))));
+    /* We need the mask load to have Aqcuire semantics to pair with the Release in
+     * update_lookuptable_tls() and avoid the reader here seeing a new mask with
+     * an old table.
+     */
+    APP(&ilist, INSTR_CREATE_dmb(dc, OPND_CREATE_INT(DR_DMB_ISH)));
     APP(&ilist,
         INSTR_CREATE_and(dc, OPREG(DR_REG_R1), OPREG(DR_REG_R1), OPREG(DR_REG_R2)));
     APP(&ilist,
--- a/core/arch/arm/emit_utils.c
+++ b/core/arch/arm/emit_utils.c
@@ -952,7 +952,7 @@ emit_indirect_branch_lookup(dcontext_t *dc, generated_code_t *code, byte *pc,
      * update_lookuptable_tls() and avoid the reader here seeing a new mask with
      * an old table.
      */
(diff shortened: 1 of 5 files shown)


derekbruening:
@algr , @AssadHashmi -- hoping for a review from someone from ARM.

core/arch/arm/emit_utils.c
```diff
@@ -948,6 +948,11 @@ emit_indirect_branch_lookup(dcontext_t *dc, generated_code_t *code, byte *pc,
     APP(&ilist,
         INSTR_CREATE_ldr(dc, OPREG(DR_REG_R1),
                          OPND_TLS_FIELD(TLS_MASK_SLOT(ibl_code->branch_type))));
+    /* We need the mask load to have Aqcuire semantics to pair with the Release in
+     * update_lookuptable_tls() and avoid the reader here seeing a new mask with
+     * an old table.
+     */
+    APP(&ilist, INSTR_CREATE_dmb(dc, OPND_CREATE_INT(DR_DMB_ISH)));
```
algrant-arm, line 955: This only needs to order loads, so it can be a DMB ISHLD.

core/fragment.c
```diff
@@ -576,7 +576,12 @@ update_lookuptable_tls(dcontext_t *dcontext, ibl_table_t *table)
      * crashing or worse.
      */
     state->table_space.table[table->branch_type].lookuptable = table->table;
-    state->table_space.table[table->branch_type].hash_mask = table->hash_mask;
+    /* Perform a Store-Release, which when combined with a Load-Acquire of the mask
+     * in the IBL itself, ensures the prior store to lookuptable is never
+     * observed before this store to hash_mask on weakly ordered arches.
+     */
+    ATOMIC_PTRSZ_ALIGNED_WRITE(&state->table_space.table[table->branch_type].hash_mask,
+                               table->hash_mask, false);
```
algrant-arm, line 584: I think this should read "is always observed before" or "is never observed after".

core/arch/arm/emit_utils.c
derekbruening: OK, what about the instances of DMB ISH in arch_exports.h: e..g, https://github.com/DynamoRIO/dynamorio/blob/master/core/arch/arch_exports.h#L845?

I based these DMB ISH on what I observed g++ 9.2.1 generating for std::atomic<int>.load(std::memory_order_acquire) on ARM and it uses "LDR; DMB ISH".  It seems to use DMB ISH for everything: fetch_add(); store(); load().

core/fragment.c
derekbruening: Good catch.  Fixed.

core/arch/arm/emit_utils.c
derekbruening: I put a note here on potentially relaxing those other barriers: https://github.com/DynamoRIO/dynamorio/issues/2502#issuecomment-617286457

PR #4270, opened 2020-04-21
Prepare 8.0.0 DR release

johnfxgalea:
Makes the necessary changes to prepare for 8.0.0 release.

### DIFF
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -285,7 +285,7 @@ Further non-compatibility-affecting changes include:
    application, but reduces overhead.
  - Added -record_dynsym_only to drcachesim for faster function tracing symbol
    lookups when internal symbols are not needed.
- - Added -record_replace_retaddrd_only to drcachesim for faster function tracing
+ - Added -record_replace_retaddr_only to drcachesim for faster function tracing
    wrapping for well-behaved applications.
  - Added dr_merge_arith_flags() as a convenience routine to merge arithmetic flags
    for restoration done by outlined code.
@@ -301,7 +301,9 @@ Further non-compatibility-affecting changes include:
    dr_get_process_id() in some contexts.
  - The private loader's malloc redirection now guarantees double-pointer-sized
    alignment, to match what system-provided allocators use.
-
+ - Added a new DR extension, namely "drbbdup", which enables different case
+   instrumentation of the same basic block by duplicating code.
+   
 **************************************************
 <hr>
 
@@ -466,8 +468,6 @@ Further non-compatibility-affecting changes include:
    aggregate, and clean up in print_results().
  - Added module_mapper_t::find_mapped_trace_bounds() to allow callers to cache
    results and avoid global locks during parallel operation.
- - Added a new DR extension, namely "drbbdup", which enables different case
-   instrumentation of the same basic block by duplicating code.
 
 **************************************************
 <hr>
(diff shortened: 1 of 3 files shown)


api/docs/release.dox
```diff
@@ -158,7 +174,7 @@ changes:
  - Changed the 32-bit x86 stack alignment of DynamoRIO and clients built using
    DR's CMake configuration from 4 to 16 on Linux to match modern conventions.
 
-The changes between version \DR_VERSION and 7.1.0 include the following minor
+The changes between version 8.0.0 and 7.1.0 include the following minor
 compatibility changes:
 
  - Added the define #MCXT_NUM_SIMD_SLOTS that was renamed from NUM_SIMD_SLOTS.
```
derekbruening, line 180: (It won't let me put a comment any further down)
If you could add to the list of non-compat changes sthg like this:

 - Fixed major synchronization bugs on ARM and AArch64.

And maybe include that in the description of the final Github Release.

api/docs/release.dox
```diff
@@ -133,7 +133,23 @@ Dr. Memory Framework (DRMF) in the same package as DynamoRIO.  DRMF
 provides the umbra, drsyscall, and drsymcache Extensions for use by
 clients.
 
-The changes between version \DR_VERSION and 7.1.0 include the following compatibility
+The changes between version \DR_VERSION and 8.0.0 include the following compatibility
+changes:
+
```
derekbruening, line 138: suggest: I usually put one line `- Nothing yet.` in these sections so it's very clear how to edit.

api/docs/release.dox
johnfxgalea: Done. What description are you referring to please?

derekbruening: This one: https://github.com/DynamoRIO/dynamorio/wiki/New-Release#update-the-release-description
I.e., the one that shows up on a releases/ page (this one just says "Auto-generated periodic build."): https://github.com/DynamoRIO/dynamorio/releases/tag/cronbuild-7.91.18369

derekbruening: I just updated that wiki section: normally I've been pointing at the changelog, as wsa done here: https://github.com/DynamoRIO/dynamorio/releases/tag/release_7.1.0.  I would say, do that, and also list the key highlights inlined on the releases page.  Maybe sthg like:

- Added Win10 1809 and higher support.
- Added AVX-512 support.
- Fixed major synchronization bugs on ARM and AArch64.
- Changed 32-bit Linux x86 stack alignment to 16.

Plus anything else you want to call out: those are the major ones that come to mind.

johnfxgalea: drbbdup? :)

derekbruening: Yes -- I don't see it in release.dox though: I would expect it to be there too.
Maybe also drdisas, and drcachesim function tracing enhancements.

Scanning the list I also see a typo:
```
 - Added -record_replace_retaddrd_only to drcachesim for faster function tracing
```
s/retaddrd/retaddr/

There are many small API additions that we can leave to the full changelist.

johnfxgalea: It's there: " Added a new DR extension, namely "drbbdup", which enables different case
   instrumentation of the same basic block by duplicating code."

derekbruening: Oh it is in the wrong place -- it's down under the 7.1.0 new features!  That is why I did not see it.

johnfxgalea: Fixed!

Issue #4280, opened 2020-05-02
Add opcode instrumentation to drmgr


johnfxgalea:
Often, during the insertion per instruction stage (i.e., stage 3), a client adds instrumentation according to the opcode of the instruction.

In order to alleviate the boiler code required by clients to check the opcode, the proposal is to add support for opcode events to drmgr. The call-backs are similar to drmgr_insertion_cb_t with the exception that they are called only when the instruction has a specific opcode as registered by the user.

In particular:

```
bool
drmgr_register_opcode_instrumentation_event(drmgr_opcode_insertion_cb_t insertion_func,
                                            int opcode, drmgr_priority_t *priority);
```

```
bool
drmgr_unregister_opcode_instrumentation_event(drmgr_opcode_insertion_cb_t insertion_func,
                                              int opcode);

```

PR #4281, opened 2020-05-02
i#4280: Add support for opcode insert instrumentation to drmgr

johnfxgalea:
Adds the support of instrumentation events specific to opcode instructions to drmgr. In particular, two new API functions are introduced: drmgr_register_opcode_instrumentation_event() and drmgr_unregister_opcode_instrumentation_event().

drmgr_register_opcode_instrumentation_event() takes a call-back function and an opcode as input. The call-back is triggered during the insert per instruction stage (i.e., stage 3) only for those instructions that have the specified opcode.

Call-backs pertaining to an opcode are maintained via a hashtable. 

Some care is taken to avoid any performance hits if opcode events are not registered. For instance, with the use of flags, the code avoids locking or hash-lookups if opcode events are never registered by the user.

In order to ensure call-backs pertaining to insert-bb and insert-opcode events are jointly triggered in the right order, with respect to their priorities, drmgr combines them into a single list on the fly by performing lookups. I avoided using one monolithic cb list as clients with a long list of different registered opcode events would have required drmgr to do wasteful iterations.

Fixes: #4280

### DIFF
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -145,7 +145,7 @@ compatibility changes:
 
 Further non-compatibility-affecting changes include:
 
- - Adds drmgr_register_opcode_instrumentation_event() and
+ - Added drmgr_register_opcode_instrumentation_event() and
    drmgr_unregister_opcode_instrumentation_event() so that drmgr supports
    opcode event instrumentation.
 
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -145,7 +145,9 @@ compatibility changes:
 
 Further non-compatibility-affecting changes include:
 
- Nothing yet.
+ - Adds drmgr_register_opcode_instrumentation_event() and
+   drmgr_unregister_opcode_instrumentation_event() so that drmgr supports
+   opcode event instrumentation.
 
 **************************************************
 <hr>
--- a/ext/drmgr/drmgr.c
+++ b/ext/drmgr/drmgr.c
@@ -88,8 +88,9 @@
 
 /* priority list entry base struct */
 typedef struct _priority_event_entry_t {
-    bool valid;                   /* is whole entry valid (not just priority) */
-    drmgr_priority_t in_priority; /* given as input by user */
+    bool valid; /* is whole entry valid (not just priority) */
+    drmgr_priority_t
+        in_priority; /* Given as input by the user. Enables reorder of cb lists. */
 } priority_event_entry_t;
 
 /* bb event list entry */
@@ -315,12 +316,15 @@ static cb_list_t cblist_exception;
 static void *exception_event_lock;
 #endif
 
-static int
-priority_event_add(cb_list_t *list, drmgr_priority_t *new_pri);
-
 static cb_list_t cblist_fault;
 static void *fault_event_lock;
 static bool registered_fault; /* for lazy registration */
+
+static int
+priority_event_add(cb_list_t *list, drmgr_priority_t *new_pri);
+
+static void
+drmgr_init_opcode_hashtable(hashtable_t *opcode_instrum_table);
 
 static void
 drmgr_thread_init_event(void *drcontext);
@@ -430,8 +434,7 @@ drmgr_init(void)
     drmgr_bb_init();
     drmgr_event_init();
     drmgr_emulation_init();
-    hashtable_init_ex(&global_opcode_instrum_table, 8, HASH_INTPTR, false, false, NULL,
-                      NULL, NULL);
+    drmgr_init_opcode_hashtable(&global_opcode_instrum_table);
     was_opcode_instrum_registered = false;
 
     our_tls_idx = drmgr_register_tls_field();
@@ -595,7 +598,6 @@ cblist_append_copy(cb_list_t *l, cb_list_t *l_to_copy)
         cb_entry_t *e = &l_to_copy->cbs.bb[i];
         if (!e->pri.valid)
             continue;
-
         int idx = priority_event_add(l, &e->pri.in_priority);
         if (idx >= 0) {
             cb_entry_t *new_e = &l->cbs.bb[idx];
@@ -635,10 +637,28 @@ cblist_delete_local(void *drcontext, cb_list_t *l, size_t local_num)
  * OPCODE INSTRUM HASHTABLE
  */
 
+static void
+drmgr_destroy_opcode_cb_list(void *opcode_cb_list_opaque)
+{
+
+    cb_list_t *opcode_cb_list = (cb_list_t *)opcode_cb_list_opaque;
+    cblist_delete(opcode_cb_list);
+    dr_global_free(opcode_cb_list, sizeof(cb_list_t));
+}
+
+static void
+drmgr_init_opcode_hashtable(hashtable_t *opcode_instrum_table)
+{
+    hashtable_init_ex(opcode_instrum_table, 8, HASH_INTPTR, false, false,
+                      drmgr_destroy_opcode_cb_list, NULL, NULL);
+}
+
+/* Returns the cb list mapped to the passed opcode. A cb list is created if it does not
+ * exist yet. If a new cb list is created and is_new is not NULL, the flag is set to true.
+ */
 static cb_list_t *
 drmgr_get_opcode_cb_list(hashtable_t *opcode_instrum_table, int opcode, OUT bool *is_new)
 {
-
     cb_list_t *opcode_cb_list =
         hashtable_lookup(opcode_instrum_table, (void *)(intptr_t)opcode);
 
@@ -727,7 +747,7 @@ drmgr_set_up_local_opcode_table(IN instrlist_t *bb, IN cb_list_t *insert_list,
     dr_rwlock_write_unlock(bb_cb_lock);
     dr_rwlock_write_unlock(opcode_table_lock);
 
-    return false;
+    return true;
 }
 
 /***************************************************************************
@@ -821,13 +841,13 @@ drmgr_bb_event(void *drcontext, void *tag, instrlist_t *bb, bool for_trace,
     cb_list_t iter_app2app;
     cb_list_t iter_insert;
     cb_list_t iter_instru;
-    cb_list_t *opcode_insert;
+    cb_list_t *iter_opcode_insert;
     per_thread_t *pt = (per_thread_t *)drmgr_get_tls_field(drcontext, our_tls_idx);
     int opcode;
-    bool local_was_opcode_instrum_registered;  /* denotes whether opcode event was ever
-                                                  registered */
-    bool is_opcode_instrum_applicable = false; /* denotes whether opcode
-                                          instrumentation is applicable for this bb */
+    /* denotes whether opcode event was ever registered */
+    bool local_was_opcode_instrum_registered;
+    /* denotes whether opcode instrumentation is applicable for this bb */
+    bool is_opcode_instrum_applicable = false;
     hashtable_t local_opcode_instrum_table;
 
     dr_rwlock_read_lock(bb_cb_lock);
@@ -897,17 +917,13 @@ drmgr_bb_event(void *drcontext, void *tag, instrlist_t *bb, bool for_trace,
     pt->first_nonlabel_instr = instrlist_first_nonlabel(bb);
     pt->last_instr = instrlist_last(bb);
 
-    /* For opcode instrumentation.
-     *
+    /* For opcode instrumentation:
      * We need to create a local copy of opcode insert call-backs. This is done at
      * this here. so that, when iterating over the basic block, instructions inserted
      * at prior stages are looked-up (including meta instructions).
      */
-
     if (local_was_opcode_instrum_registered) {
-        hashtable_init_ex(&local_opcode_instrum_table, 8, HASH_INTPTR, false, false, NULL,
-                          NULL, NULL);
-
+        drmgr_init_opcode_hashtable(&local_opcode_instrum_table);
         is_opcode_instrum_applicable = drmgr_set_up_local_opcode_table(
             bb, &iter_insert, &local_opcode_instrum_table);
     }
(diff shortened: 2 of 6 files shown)


johnfxgalea:
Sorry, just thought of an even better design. I'll notify when it is ready.

johnfxgalea:
Does a closing of this PR trigger the tests (followed by a reopening) ? I'm testing this.

johnfxgalea:
I made additional changes.

@derekbruening It is ready for review when you get a chance please.

api/docs/release.dox
```diff
@@ -145,7 +145,9 @@ compatibility changes:
 
 Further non-compatibility-affecting changes include:
 
- Nothing yet.
+ - Adds drmgr_register_opcode_instrumentation_event() and
```
derekbruening, line 148: nit: All the lists below use the past tense here, so s/Adds/Added/ for consistency.

ext/drmgr/drmgr.h
```diff
@@ -165,6 +165,19 @@ typedef dr_emit_flags_t (*drmgr_insertion_cb_t)(void *drcontext, void *tag,
                                                 bool for_trace, bool translating,
                                                 void *user_data);
 
+/**
+ * Callback function for opcode based instrumentation. In particular, this callback
+ * is triggered only for specific instruction opcodes. This is done during the
+ * third stage, i.e., instrumentation insertion.
```
derekbruening, line 171: nit: you s/instrumentation insertion/ right above but then used the phrase here

ext/drmgr/drmgr.h
```diff
@@ -165,6 +165,19 @@ typedef dr_emit_flags_t (*drmgr_insertion_cb_t)(void *drcontext, void *tag,
...
+ *
+ * See #dr_emit_flags_t for an explanation of the return value.  If
+ * any instrumentation pass requests DR_EMIT_STORE_TRANSLATIONS, they
```
derekbruening, line 174: `#DR_EMIT_STORE_TRANSLATIONS`

ext/drmgr/drmgr.h
```diff
@@ -165,6 +165,19 @@ typedef dr_emit_flags_t (*drmgr_insertion_cb_t)(void *drcontext, void *tag,
...
+ * will be stored.
+ */
+typedef dr_emit_flags_t (*drmgr_opcode_insertion_cb_t)(void *drcontext, void *tag,
+                                                       instrlist_t *bb, instr_t *inst,
+                                                       bool for_trace, bool translating);
```
derekbruening, line 179: How about including a `void *user_data`?  I wish we had done so for all event cb's up front (instead of having to add to some later as extended versions).

ext/drmgr/drmgr.h
```diff
@@ -452,6 +465,53 @@ drmgr_unregister_bb_instrumentation_ex_event(drmgr_app2app_ex_cb_t app2app_func,
                                              drmgr_insertion_cb_t insertion_func,
                                              drmgr_ilist_ex_cb_t instru2instru_func);
 
+DR_EXPORT
+/**
+ * Registers callback functions for the third
+ * instrumentation stage: instrumentation
+ * insertion.  drmgr will call \p func for each instruction with the
```
derekbruening, line 472: :) Here it is again "instrumentation insertion"

ext/drmgr/drmgr.h
```diff
@@ -452,6 +465,53 @@ drmgr_unregister_bb_instrumentation_ex_event(drmgr_app2app_ex_cb_t app2app_func,
...
+ * specific opcode \p opcode.
+ *
+ * More than one call-back function can be mapped to the same opcode. Their
+ * execution sequence is determined by their priority \p priority if set. Ordering
+ * based on priority is also taken in to account with respect to insert bb events.
```
derekbruening, line 477: s/in to/into/

derekbruening, line 477: suggest: s/insert bb/bb instrumentation/.  Do you mean the main single-instruction instru-insertion event?  If so I would replace "bb" with instruction or sthg.

ext/drmgr/drmgr.h
```diff
@@ -452,6 +465,53 @@ drmgr_unregister_bb_instrumentation_ex_event(drmgr_app2app_ex_cb_t app2app_func,
...
+ * \return false upon failure.
+ *
+ * @param[in]  func  The opcode insertion callback to be called for the third
+ * stage for a specific opcode instruction. Cannot be NULL.
+ * @param[in]  opcode          The opcode to associate with the insertion callback.
+ * @param[in]  priority        Specifies the relative ordering of both callbacks.
+ *                             Can be NULL, in which case a default priority is used.
+ *
+ * \note It is possible that this call-back will be triggered for meta instructions.
+ * Therefore, we recommend that the call-back checks for meta instructions
```
derekbruening, line 493: grammar: s/checks/check/

ext/drmgr/drmgr.h
```diff
@@ -452,6 +465,53 @@ drmgr_unregister_bb_instrumentation_ex_event(drmgr_app2app_ex_cb_t app2app_func,
...
+ * Registers callback functions for the third
+ * instrumentation stage: instrumentation
+ * insertion.  drmgr will call \p func for each instruction with the
+ * specific opcode \p opcode.
+ *
+ * More than one call-back function can be mapped to the same opcode. Their
+ * execution sequence is determined by their priority \p priority if set. Ordering
+ * based on priority is also taken in to account with respect to insert bb events.
+ *
+ * Since this call-back is triggered during insert instrumentation,
```
derekbruening, line 479: Hmm, try `git grep callback` and `git grep call-back`.  Certainly the former is used everywhere  else (except drbbdup).

ext/drmgr/drmgr.h
```diff
@@ -452,6 +465,53 @@ drmgr_unregister_bb_instrumentation_ex_event(drmgr_app2app_ex_cb_t app2app_func,
...
+DR_EXPORT
+/**
+ * Unregisters the opcode-specific callback that
+ * was registered via drmgr_register_opcode_instrumentation_event().
+ *
+ * \return true if unregistration is successful and false if it is not
+ * (e.g., \p func was not registered for the passed opcode \p opcode).
+ *
+ * The recommendations for #dr_unregister_bb_event() about when it
+ * is safe to unregister apply here as well.
```
derekbruening, line 509: s/#// (the empty parens make the link for functions: http://www.doxygen.nl/manual/autolink.html)

ext/drmgr/drmgr.c
```diff
@@ -567,23 +590,51 @@ cblist_shift_and_resize(cb_list_t *l, uint insert_at)
...
+    ASSERT(src->num_def <= dst->capacity, "dst must have large enough capacity");
+    dst->entry_sz = src->entry_sz;
+    dst->num_def = src->num_def;
+    memcpy(dst->cbs.array, src->cbs.array, src->num_def * src->entry_sz);
+}
+
 /* Creates a temporary local copy, using local if it's big enough but
- * otherwise allocating new space on the heap.
- * Caller must hold read lock.
+ * otherwise allocating new space on the heap. It does a scoped allocation
```
derekbruening, line 623: "scoped allocation" sounds to me like sthg local to this function which is inaccessible on return: do you mean a thread-local heap allocation?

ext/drmgr/drmgr.c
```diff
@@ -594,6 +645,104 @@ cblist_delete_local(void *drcontext, cb_list_t *l, size_t local_num)
...
+    }
+
+    /* Since both opcode and insert events are handled during stage 3, they need to be
+     * jointly organized according to their priorities. This is done now.
+     *
+     * An alternative approach is to do the merging upon registration of opcode
+     * events, and updating all mapped cb lists of opcodes when further bb insert events
+     * are registered/unregistered. However, in order to avoid races, this would require
+     * the dr bb event processed by drmgr to create a local copy of the entire opcode
+     * table every time.
```
derekbruening, line 718: Hmm, I guess I'm imagining there only being a couple of opcode instru cbs, and we'd just put them on the same cblist_instrumentation as the regular instruction insertion cbs, and in the existing loop over the instrs (what you have now as drmgr_bb_event_do_insertion_per_instr) we'd just check the opcode vs the opcode stored with the cb before deciding whether to invoke the cb.  No need for a hashtable.  It seems a simpler and more efficient solution?  Though I only thought it about for a few minutes.  Unless you're anticipating a whole lot of these opcode cbs?

johnfxgalea: hmm. So in the PR comment I said:

> I avoided using one monolithic cb list as clients with a long list of different registered opcode events would have required drmgr to do wasteful iterations.

It depends on how many opcode events are registered. I guess I can revert and make the code simpler.

If the user has many opcode events, then I guess the user should refrain from using this interface and use a large switch statement (something similar to hiw dr memory maintains def tags). 


If there are many, they I would recommend the

ext/drmgr/drmgr.h
johnfxgalea: Unlike cbs for instr insert instrumentation, opcode cbs are not paired with any prior analysis cbs for now. Shall I add the parameter and pass NULL?

ext/drmgr/drmgr.h
johnfxgalea: I think it is correct how it is no?

ext/drmgr/drmgr.h
johnfxgalea: Unlike cbs for instr insert instrumentation, opcode cbs are not paired with any prior analysis cbs for now. Shall I add the parameter and pass NULL?

api/docs/release.dox
johnfxgalea: Done.

ext/drmgr/drmgr.h
johnfxgalea: Done. This the same for other docs.

ext/drmgr/drmgr.h
johnfxgalea: Done - reverted back (looked odd with two nouns, but it is pretty much used everywhere in the header file).

ext/drmgr/drmgr.h
johnfxgalea: Done.

ext/drmgr/drmgr.h
johnfxgalea: Done.

ext/drmgr/drmgr.h
johnfxgalea: Done - I will fix drbbdup later.

ext/drmgr/drmgr.h
johnfxgalea: Done - I fixed other docs as well.

ext/drmgr/drmgr.c
johnfxgalea: Done.

ext/drmgr/drmgr.h
derekbruening: No, this is the subjunctive mood, which is not conjugated for subject.  Certainly it should agree with the other verb on the next line ("ignore").

ext/drmgr/drmgr.h
derekbruening: I meant a static param passed at registration time (yeah that is a little confusing vs the per-invocation user_data used in the combined instru cbs).

ext/drmgr/drmgr.h
johnfxgalea: Ah, right, sorry. So the wording in the brackets needs to be fixed.

derekbruening: Not sure what you mean about the brackets/parentheses: the "and ignore them, typically"?  I think that's fine.

ext/drmgr/drmgr.c
```diff
@@ -215,6 +221,12 @@ static cb_list_t cblist_app2app;
 static cb_list_t cblist_instrumentation;
 static cb_list_t cblist_instru2instru;
 
+/* For opcode specific instrumentation. */
+static hashtable_t global_opcode_instrum_table; /* maps opcodes to cb lists */
+static void *opcode_table_lock;
+/* a flag indicating whether opcode instrum was ever registered */
+static bool was_opcode_instrum_registered;
```
derekbruening, line 228: Comment that this is protected by bb_cb_lock

ext/drmgr/drmgr.c
derekbruening: I'm ok with it as is.

johnfxgalea: Okay. I'm fine with both approaches tbh because they both offer their own benefits. If this is the one we are going for, then I'll add a comment on the other.

ext/drmgr/drmgr.h
johnfxgalea: Haha, AFAIR in British English, it is acceptable to keep the s (but I need to change "ignore" to "ignores"). In AmE, this may come across as odd. So I'll go ahead and drop the "s" :p

ext/drmgr/drmgr.c
johnfxgalea: Done.

ext/drmgr/drmgr.h
johnfxgalea: Done

ext/drmgr/drmgr.h
johnfxgalea: Done.

ext/drmgr/drmgr.h
johnfxgalea: Done.

johnfxgalea:
Thank you for the review. I added the passing of user data as well.

johnfxgalea:
run arm tests

johnfxgalea:
@derekbruening I encountered the following error in Windows:

> error LNK2019: unresolved external symbol __chkstk referenced in function drmgr_bb_event


Usually this occurs when stack space is pressured. I added a new field in the cb_entry_t struct to store user data, which I believe triggered the error. Therefore I reduced the buffer size by 1 (from 16 to 15) and the error went away.

Are you okay with the change before I merge please?

johnfxgalea:
run arm tests

derekbruening:
> @derekbruening I encountered the following error on Windows:
> 
> > error LNK2019: unresolved external symbol __chkstk referenced in function drmgr_bb_event
> 
> Usually this occurs when stack space is pressured. I added a new field in the cb_entry_t struct to store user data, which I believe triggered the error. Therefore I reduced the buffer size by 1 (from 16 to 15) and the error went away.
> 
> Are you okay with the change before I merge please?

DR's stacks do not have Windows-stack-expansion guard pages so __chkstk is just going to crash.  Not sure if there is a compiler flag to not use it.  I would guess 16 to 15 is ok.

johnfxgalea:
run arm tests

johnfxgalea:
@AssadHashmi Any idea why ARM tests are not being triggered please?

AssadHashmi:
> @AssadHashmi Any idea why ARM tests are not being triggered please?

@johnfxgalea the Jenkins host instance is un-pingable ATM. I'm investigating.

AssadHashmi:
run arm tests

johnfxgalea:
@AssadHashmi Thank you for fixing it so quickly.

PR #4284, opened 2020-05-08
i#4280: Adds a sample highlighting opcode instrumentation.

johnfxgalea:
Adds a new sample client to show how drmgr's opcode instrumentation can be used to count the number of executions of an instruction (with a specific opcode). The opcode number is given by the user as a parameter (the client uses droptions for parsing).

In addition, the PR also includes a bug fix in drmgr. In particular, drmgr was confusing the default priority as a different priority when used again for another event registration. The bug was only triggered if cblist_insert_other() is used - this function is only called for opcode registration at the moment so the bug's severity is low.

### DIFF
--- a/api/docs/samples.dox
+++ b/api/docs/samples.dox
@@ -118,6 +118,9 @@ produces a binary-format trace file.
 The sample <a href="https://github.com/DynamoRIO/dynamorio/tree/master/api/samples/hot_bbcount.c">hot_bbcount.c</a>
 uses the drbbdup extension to count the execution of hot basic blocks which have exceeded a given hit threshold.
 
+The sample <a href="https://github.com/DynamoRIO/dynamorio/tree/master/api/samples/opcode_count.cpp">opcode_count.cpp</a>
+takes an opcode as input and uses drmgr's instrumentation events to count the number of times the instruction is executed.
+
 The sample <a href="https://github.com/DynamoRIO/dynamorio/tree/master/api/samples/modxfer.c">modxfer.c</a>
 reports the control flow transfers between modules.
 
--- a/api/docs/samples.dox
+++ b/api/docs/samples.dox
@@ -119,7 +119,7 @@ The sample <a href="https://github.com/DynamoRIO/dynamorio/tree/master/api/sampl
 uses the drbbdup extension to count the execution of hot basic blocks which have exceeded a given hit threshold.
 
 The sample <a href="https://github.com/DynamoRIO/dynamorio/tree/master/api/samples/opcode_count.cpp">opcode_count.cpp</a>
-takes an opcode as input and uses drmgr's instrumentation events to count the number of times instruction with that opcode
+takes an opcode as input and uses drmgr's instrumentation events to count the number of times instructions with that opcode
 are executed.
 
 The sample <a href="https://github.com/DynamoRIO/dynamorio/tree/master/api/samples/modxfer.c">modxfer.c</a>
--- a/api/docs/samples.dox
+++ b/api/docs/samples.dox
@@ -119,7 +119,8 @@ The sample <a href="https://github.com/DynamoRIO/dynamorio/tree/master/api/sampl
 uses the drbbdup extension to count the execution of hot basic blocks which have exceeded a given hit threshold.
 
 The sample <a href="https://github.com/DynamoRIO/dynamorio/tree/master/api/samples/opcode_count.cpp">opcode_count.cpp</a>
-takes an opcode as input and uses drmgr's instrumentation events to count the number of times the instruction is executed.
+takes an opcode as input and uses drmgr's instrumentation events to count the number of times instruction with that opcode
+are executed.
 
 The sample <a href="https://github.com/DynamoRIO/dynamorio/tree/master/api/samples/modxfer.c">modxfer.c</a>
 reports the control flow transfers between modules.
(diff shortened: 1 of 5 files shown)


johnfxgalea:
@derekbruening The tests failed because the sample client requires a mandatory command-line parameter (i.e. the opcode). How can I set this parameter in the test suite please?

derekbruening:
> @derekbruening The tests failed because the sample client requires a mandatory command-line parameter (i.e. the opcode). How can I set this parameter in the test suite please?

This would need to be changed: https://github.com/DynamoRIO/dynamorio/blob/master/suite/tests/CMakeLists.txt#L2746

api/docs/samples.dox
```diff
@@ -118,6 +118,9 @@ produces a binary-format trace file.
 The sample <a href="https://github.com/DynamoRIO/dynamorio/tree/master/api/samples/hot_bbcount.c">hot_bbcount.c</a>
 uses the drbbdup extension to count the execution of hot basic blocks which have exceeded a given hit threshold.
 
+The sample <a href="https://github.com/DynamoRIO/dynamorio/tree/master/api/samples/opcode_count.cpp">opcode_count.cpp</a>
+takes an opcode as input and uses drmgr's instrumentation events to count the number of times the instruction is executed.
```
derekbruening, line 122: "the instruction" sounds like a single specific instruction: reword to sthg like "instructions with that opcde are"

api/samples/opcode_count.cpp
```diff
@@ -0,0 +1,165 @@
...
+#ifdef WINDOWS
+#    define DISPLAY_STRING(msg) dr_messagebox(msg)
+#else
+#    define DISPLAY_STRING(msg) dr_printf("%s\n", msg);
+#endif
+
+#define NULL_TERMINATE(buf) (buf)[(sizeof((buf)) / sizeof((buf)[0])) - 1] = '\0'
+
+static droption_t<int> opcode(DROPTION_SCOPE_CLIENT, "opcode", -1,
+                              "The opcode to consider",
```
derekbruening, line 58: How about s/consider/count/

api/samples/opcode_count.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+/* Code Manipulation API Sample:
+ * opcode_count.c
+ *
+ * Reports the dynamic execution count of all instructions with a particular opcode.
+ * Illustrates how to to use drmgr to register opcode events and inline locked
+ * counter code.
+ */
+
+#include <stdint.h> /* for intptr */
```
derekbruening, line 41: Could use DR's types -- or `<cstdint>` is more C++y

api/samples/opcode_count.cpp
```diff
@@ -0,0 +1,165 @@
...
+        !drmgr_register_bb_instrumentation_event(NULL, event_app_instruction, NULL))
+        DR_ASSERT(false);
+
+    /* make it easy to tell, by looking at log file, which client executed */
+    dr_log(NULL, DR_LOG_ALL, 1, "Client 'opcode_count' initializing\n");
+#ifdef SHOW_RESULTS
+    /* also give notification to stderr */
+    if (dr_is_notify_on()) {
+#    ifdef WINDOWS
+        /* ask for best-effort printing to cmd window.  must be called at init. */
```
derekbruening, line 158: style: capitalize.  I would also change the 6 full-line comments above to capitalize and have end punctuation: looks more professional and is the current style.

ext/drmgr/drmgr.c
```diff
@@ -650,6 +660,7 @@ cblist_delete_local(void *drcontext, cb_list_t *l, size_t local_num)
 static void
 cblist_create_global(cb_list_t *src, cb_list_t *dst)
 {
+    memset(dst, 0, sizeof(cb_list_t));
```
derekbruening, line 663: `sizeof(*dst)` (better practice in general b/c it's future-proof vs changing the type of `dst`)

api/docs/samples.dox
johnfxgalea: Done.

api/samples/opcode_count.cpp
johnfxgalea: Done

api/samples/opcode_count.cpp
johnfxgalea: Done.

ext/drmgr/drmgr.c
johnfxgalea: Done.

api/samples/opcode_count.cpp
johnfxgalea: Done.

johnfxgalea:
@derekbruening Thank you for taking a look at the PR.

derekbruening:
￼> Make client work without param

Note that you could add a param by looking for the sample name in that loop, just like DR options are added for some samples, if that's a preferable route.  I was just pointing out where the code was: didn't mean to imply that the comment about samples working w/o a param was sthg we didn't want to change.  I don't think anywhere else relies on that.  (Maybe the manual sanity tests running through all the samples.)

johnfxgalea:
> > Make client work without param
> 
> Note that you could add a param by looking for the sample name in that loop, just like DR options are added for some samples, if that's a preferable route. I was just pointing out where the code was: didn't mean to imply that the comment about samples working w/o a param was sthg we didn't want to change. I don't think anywhere else relies on that. (Maybe the manual sanity tests running through all the samples.)

Yep, I added support for both (the prior commit includes the check in the cmake file that you mentioned.)

Issue #4231, opened 2020-03-27
drwrap-test failed on Appveyor 32-bit: 2nd library load missing some callbacks


derekbruening:
The drwrap-test failed once in PR #4230 on 32-bit Appveyor in a strange way.  On the 2nd library load, most of the wrap callbacks were not called.  I can't reproduce this locally: ran 500x.  On Appveyor it did not reproduce.

https://ci.appveyor.com/project/DynamoRIO/dynamorio/builds/31767039
```
thread.appdll process init
  <pre-level0>
in level0 42
  <pre-level1>
in level1 42 1111
  <pre-makes_tailcall>
  <pre-level2>
in level2 1153
  <post-level2>
  <post-makes_tailcall>
  <post-level1>
level1 returned -4
  <post-level0>
level0 returned 42
  <pre-skipme>
skipme returned 7 and x=3
  <pre-repeat#1>
in repeatme with arg 3
  <post-repeat#1>
  <pre-repeat#2>
in repeatme with arg 2
  <post-repeat#2>
repeatme returned 2
replaceme returned 0 and x=6
replaceme2 returned 1 and x=999
replace_callsite returned 2 and x=777
  <pre-preonly>
in preonly
in postonly
  <post-postonly>
in skipme
in postonly
  <pre-direct1>
  <post-direct1>
  <pre-direct2>
  <pre-direct1>
  <post-direct1>
  <post-direct2>
  <pre-direct1>
in runlots 1024
  <pre-long0>
  <pre-long1>
long0 A
  <pre-long2>
  <pre-long3>
long1 A
  <pre-long4>
  <pre-long5>
long2 A
  <pre-long6>
  <pre-long7>
long3 A
  <post-long7 abnormal>
  <post-long6 abnormal>
  <post-long5 abnormal>
  <post-long4 abnormal>
  <post-long3 abnormal>
  <post-long2 abnormal>
  <post-long1 abnormal>
  <post-long0 abnormal>
longdone
  <pre-called_indirectly>
  <pre-called_indirectly_subcall>
called_indirectly_subcall 43
  <post-called_indirectly_subcall>
called_indirectly 42 => 44
  <post-called_indirectly>
  <pre-tailcall_test2>
print_from_asm 1
  <pre-tailcall_tail>
print_from_asm 7
  <post-tailcall_tail>
  <post-tailcall_test2>
loaded library
thread.appdll process init
in level0 37
in level1 37 74
in level2 111
level1 returned 37
[...]

Issue #4223, opened 2020-03-26
Mac Travis build output truncated, hiding build failure details


derekbruening:
This just showed up on PR #4221 3 times in a row (every build there including a restart I tried):

https://travis-ci.com/github/DynamoRIO/dynamorio/jobs/304776211
```
4080[ 22%] Building C object suite/tests/CMakeFiles/tool.fib_plus.dir/__/__/clients/drcachesim/tests/fib_plus.c.o
4081[ 22%] Linking C executable bin/tool.fib_plus
4082[ 22%] Built target tool.fib_plus
4083[ 22%] 
```

It looks like something external is killing the build process.
Xref #4126 where the same thing was happening but there it was after 10
minutes: here it's only 1 minute in.

derekbruening:
Wow, master has a compile failure from that PR #4421!  So probably the problem is log buffering by Travis and not printing everything out?!?  Possibly the tee in our script but why would it be non-deterministic?
https://travis-ci.com/github/DynamoRIO/dynamorio/jobs/304949581
```
4182[ 33%] Building ASM_NASM object suite/tests/CMakeFiles/client.drwrap-test.appdll.dir/drwrap-test.appdll.c_asm.asm.o
4183/Users/travis/build/DynamoRIO/dynamorio/build_debug-internal-64/suite/tests/drwrap-test.appdll.c_asm.asm:402: error: symbol `tailcall_tail' undefined
4184/Users/travis/build/DynamoRIO/dynamorio/build_debug-internal-64/suite/tests/drwrap-test.appdll.c_asm.asm:409: error: label `_tailcall_tail' changed during code generation [-w+error=label-redef-late]
4185make[2]: *** [suite/tests/CMakeFiles/client.drwrap-test.appdll.dir/drwrap-test.appdll.c_asm.asm.o] Error 1
```

derekbruening:
This seems to be happening again. For PR #4292 the Mac build output just ends mid-build, even after re-run: https://travis-ci.com/github/DynamoRIO/dynamorio/jobs/332667228

Looking closely there is a build failure:
```
/Users/travis/build/DynamoRIO/dynamorio/core/unix/injector.c:221:17: error: unused variable 'ld_preload' [-Werror,-Wunused-variable]
    const char *ld_preload = NULL;
                ^
Users/travis/build/DynamoRIO/dynamorio/core/unix/injector.c:223:17: error: unused variable 'cur_preload' [-Werror,-Wunused-variable]
[...]

PR #4292, opened 2020-05-12
i#4291: append to LD_PRELOAD when using -late

### LINKED ISSUE #4291: drrun overwrites LD_PRELOAD when using -late
**Describe the bug**
When using `drrun`'s `-late` mode of injection, `LD_PRELOAD` is employed to load the DR support libraries. In the process of doing so, DR overwrites the existing value of `LD_PRELOAD`, rather than appending to the existing value.

**To Reproduce**
```
$ cd <DR_ROOT>/bin64
$ LD_PRELOAD=libtest.so ./drrun -late -c ../samples/bin64/libbbcount.so -- env
```
Note that the output does not include `libtest.so`.

**Expected behavior**
`LD_PRELOAD`'s value should be preserved through invocations of `drrun`.

**Versions**
Occurs on 7.1 and 8.0 releases on Linux on both 32/64 bit builds.

**Additional context**
PR coming shortly. Issue migrated from [Google Groups discussion](https://groups.google.com/forum/#!topic/DynamoRIO-Users/kGOpz9WASQ8).

mattgreen:
When using -late, append to LD_PRELOAD rather than overwrite it.

Fixes #4291

### DIFF
--- a/core/unix/injector.c
+++ b/core/unix/injector.c
@@ -213,10 +213,14 @@ report_dynamorio_problem(dcontext_t *dcontext, uint dumpcore_flag, app_pc except
 static void
 pre_execve_ld_preload(const char *dr_path)
 {
+    char ld_preload_libs[MAX_OPTIONS_STRING];
     char ld_lib_path[MAX_OPTIONS_STRING];
     const char *last_slash = NULL;
     const char *mode_slash = NULL;
     const char *lib_slash = NULL;
+    const char *ld_preload = NULL;
+    const char *dr_ld_preload_libs = "libdynamorio.so libdrpreload.so";
+    const char *cur_preload = getenv("LD_PRELOAD");
     const char *cur_path = getenv("LD_LIBRARY_PATH");
     const char *cur = dr_path;
     /* Find last three occurrences of '/'. */
@@ -254,7 +258,17 @@ pre_execve_ld_preload(const char *dr_path)
     setenv("DYLD_FORCE_FLAT_NAMESPACE", "1", true /*overwrite*/);
 #else
     setenv("LD_LIBRARY_PATH", ld_lib_path, true /*overwrite*/);
-    setenv("LD_PRELOAD", "libdynamorio.so libdrpreload.so", true /*overwrite*/);
+
+    if (cur_preload != NULL) {
+        snprintf(ld_preload_libs, BUFFER_SIZE_ELEMENTS(ld_preload_libs), "%s %s",
+                 cur_preload, dr_ld_preload_libs);
+        NULL_TERMINATE_BUFFER(ld_preload_libs);
+        ld_preload = ld_preload_libs;
+    } else {
+        ld_preload = dr_ld_preload_libs;
+    }
+
+    setenv("LD_PRELOAD", ld_preload, true /*overwrite*/);
 #endif
     if (verbose) {
         printf("Setting LD_USE_LOAD_BIAS for PIEs so the loader will honor "
--- a/core/unix/injector.c
+++ b/core/unix/injector.c
@@ -213,16 +213,18 @@ report_dynamorio_problem(dcontext_t *dcontext, uint dumpcore_flag, app_pc except
 static void
 pre_execve_ld_preload(const char *dr_path)
 {
-    char ld_preload_libs[MAX_OPTIONS_STRING];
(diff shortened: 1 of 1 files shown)


derekbruening:
run arm tests

derekbruening:
Travis is misbehaving: their Mac bots are dying or sthg midway through our build.  Will try one more rerun.

derekbruening:
So it is in fact a build failure on Mac.  It is not presented very well which is #4223.
```
/Users/travis/build/DynamoRIO/dynamorio/core/unix/injector.c:221:17: error: unused variable 'ld_preload' [-Werror,-Wunused-variable]
    const char *ld_preload = NULL;
                ^
Users/travis/build/DynamoRIO/dynamorio/core/unix/injector.c:223:17: error: unused variable 'cur_preload' [-Werror,-Wunused-variable]
    const char *cur_preload = getenv("LD_PRELOAD");
                ^
/Users/travis/build/DynamoRIO/dynamorio/core/unix/injector.c:222:17: error: unused variable 'dr_ld_preload_libs' [-Werror,-Wunused-variable]
    const char *dr_ld_preload_libs = "libdynamorio.so libdrpreload.so";
                ^
/Users/travis/build/DynamoRIO/dynamorio/core/unix/injector.c:216:10: error: unused variable 'ld_preload_libs' [-Werror,-Wunused-variable]
    char ld_preload_libs[MAX_OPTIONS_STRING];
```

@mattgreen, If you could fix the build failure.

core/unix/injector.c
```diff
@@ -213,16 +213,18 @@ report_dynamorio_problem(dcontext_t *dcontext, uint dumpcore_flag, app_pc except
...
 pre_execve_ld_preload(const char *dr_path)
 {
-    char ld_preload_libs[MAX_OPTIONS_STRING];
     char ld_lib_path[MAX_OPTIONS_STRING];
     const char *last_slash = NULL;
     const char *mode_slash = NULL;
     const char *lib_slash = NULL;
+    const char *cur_path = getenv("LD_LIBRARY_PATH");
+    const char *cur = dr_path;
+#ifdef LINUX
```
derekbruening, line 222: Hmm, so Mac is still overwriting instead of appending, so we have split behavior now.  Maybe worth addressing -- or maybe just a XXX comment.  I think it will let me edit this with a comment.

mattgreen: Good catch. I dev on a Mac, so I can fix this there as well if you'd like.

derekbruening: That would be awesome, thanks.

mattgreen:
Hmm, Windows build failed. I'm surprised by that considering the source file isn't in the build. Is there a way to re-run it?

core/unix/injector.c
```diff
@@ -259,16 +272,11 @@ pre_execve_ld_preload(const char *dr_path)
...
 
-    if (cur_preload != NULL) {
-        snprintf(ld_preload_libs, BUFFER_SIZE_ELEMENTS(ld_preload_libs), "%s %s",
-                 cur_preload, dr_ld_preload_libs);
-        NULL_TERMINATE_BUFFER(ld_preload_libs);
-        ld_preload = ld_preload_libs;
-    } else {
-        ld_preload = dr_ld_preload_libs;
-    }
+    snprintf(ld_lib_path, BUFFER_SIZE_ELEMENTS(ld_lib_path), "%s%s", preload_prefix,
```
derekbruening, line 275: I think you've omitted the %c preload_delimiter.

A test would presumably catch this.  Though with -late not used much it seems ok to not put in the effort for it: would have to make some other lib to be preloaded and check it got in there too with a -late injection.

derekbruening: Ah my bad: it's in the preload_prefix which is shared.

derekbruening:
> Hmm, Windows build failed. I'm surprised by that considering the source file isn't in the build. Is there a way to re-run it?

Looks like drwrap-test 32-bit: #4231.  Yes it can be re-run individually but I'll update the branch which is another way to re-run.

derekbruening:
run arm tests

derekbruening:
Just documenting that we're going to live without a test for the new code.

mattgreen:
Thanks for the feedback and accepting it!

PR #4294, opened 2020-05-13
Fix crash caused by missaligned stack when using AppInit on Windows10 x64

TomSie:


derekbruening:
Ideally we would also add a test: but since AppInit injection is still not working for taking over a process that is not as easy.  We currently do not have many AppInit tests: unfortunately it was never a focus over the last few years.  Let's put in this bug fix and if you are interested in helping further in this area we would love to have more fixes and tests for AppInit if you would like to contribute further.

derekbruening:
run arm tests

derekbruening:
> Ideally we would also add a test: but since AppInit injection is still not working for taking over a process that is not as easy. 

What I mean is, we can't just enable the existing AppInit tests for win8+ (currently disabled IIRC) since I think they all only test DR takeover.  We would want to add a new test of non-takeover via AppInit.

TomSie:
A working AppInit functionality would be of great use for my work. However, I understand that the reasons of why it's currently not working are complex, and I have little time on my hand. Nevertheless, I will try to contribute as much as I can :)

Issue #2662, opened 2017-10-10
FR: Retrieving CL arguments of  program passed to drrun


johnfxgalea:
I would like to request a feature that extends the API to access the arguments passed to the program under analysis, as well as the count.

The binaries being analysed should be assumed to be stripped. As a result, this makes it challenging to wrap the main function, using drwap, to obtain the arguments as the function won't be exported.

This request is in conjunction to post on the UserGroup, found [here](https://groups.google.com/forum/#!topic/dynamorio-users/ST3BOGD8vd8)

derekbruening:
Removing "drrun" from the title as this should operate on any app whether injected from the parent or launched directly

johnfxgalea:
I am going to try and get a PR for this functionality. I'll start of with Linux. Any tips which I should be aware about? I am thinking of examining memory upon the first instruction executed.

egrimley:
I'm confused. If you're using "early injection", which is the default on Linux, then DynamoRIO's loader has already handled the command line. What would be the use of "examining memory"? If you're not using "early injection", how would "examining memory" work?

Does `/proc/self/cmdline` help at all? It would be nice not to depend on `/proc`, but DynamoRIO already depends on `/proc` in several ways.

derekbruening:
Quoting from https://groups.google.com/forum/#!topic/dynamorio-users/ST3BOGD8vd8

> You could get the info from where the kernel stashed it.  On Windows you
> would look in the PEB to find the command line.  On UNIX you could get the
> initial app stack and look at what the kernel set up.

On UNIX argc and argv are placed on the stack by the kernel and are easy to find at the start with early injection.  E.g., look at privload_early_inject() which is already reading them and printing them out in debug build.

johnfxgalea:
Thanks, that is useful to know.

johnfxgalea:
Pending work: support for Windows + late injection.

PR #4295, opened 2020-05-13
i#2662: Get app args

johnfxgalea:
Adds dr_get_app_args() to enable a client to retrieve the locations where command-line arguments of the app reside. C strings of the arguments are retrieved via a new function, namely dr_app_arg_as_cstring(). Also, dr_num_app_args() enables the client to get the number of arguments.

The PR adds support only for Unix and early injection for now.

The gathering of error codes of failed DR API routines is also added with dr_get_error_code().

Issue: #2662.


core/unix/os.c
```diff
@@ -1087,6 +1091,28 @@ get_application_name(void)
     return get_application_name_helper(false, true /* full path */);
 }
 
+/* Sets pointers to the application's command-line arguments. These pointers are then used
+ * by get_application_cl_args().
```
derekbruening, line 1095: `get_application_cl_args` does not exist

core/unix/os.c
```diff
@@ -295,6 +295,10 @@ static app_pc executable_end = NULL;
 static char executable_path[MAXIMUM_PATH];
 static char *executable_basename;
 
+/* Pointers to arguments. Refers to the main stack set up by the kernel. */
```
derekbruening, line 298: Let's try to make new code more weak-memory-ordering (i.e., ARM) aware: let's put comments on globals as to why they do not need barriers on reads and writes.  Sthg like: these are only written once during process init and we can live with the non-guaranteed-delay until they are visible to other cores.

suite/tests/CMakeLists.txt
```diff
@@ -2689,6 +2689,10 @@ endif ()
     use_DynamoRIO_extension(client.nudge_ex.dll drmgr)
   endif (X86)
 
+  if (X86)
```
derekbruening, line 2692: There's nothing x86-specific here is there?

suite/tests/client-interface/app_args.c
```diff
@@ -0,0 +1,49 @@
+/* **********************************************************
+ * Copyright (c) 2012-2013 Google, Inc.  All rights reserved.
```
derekbruening, line 2: Update

suite/tests/client-interface/app_args.dll.c
```diff
@@ -0,0 +1,63 @@
...
+    bool result = dr_get_app_args(&app_argc, &app_argv);
+
+#ifdef UNIX
+    CHECK(strcmp(app_argv[1], "Test") == 0, "first arg should be Test");
+    CHECK(strcmp(app_argv[2], "Test2") == 0, "second arg should be Test2");
+    CHECK(strcmp(app_argv[3], "Test3") == 0, "third arg should be Test3");
+    CHECK(result == true, "should succeed");
+#else
+    /* XXX i#2662: Windows not yet supported. */
+    CHECK(result == true, "should fail");
```
derekbruening, line 61: You mean false?

core/os_shared.h
```diff
@@ -314,6 +314,8 @@ char *
 get_application_pid(void);
 char *
 get_application_name(void);
+bool
+get_app_args(int **app_argc_out, char ***app_argv_out);
```
derekbruening, line 318: For sthg in os_shared.h we generally add an impl to the other OS too even if it's NYI

core/lib/instrument_api.h
```diff
@@ -1806,6 +1806,19 @@ DR_API
...
+ * Sets the passed arguments to the command-line arguments of the
+ * application.
+ * @param argc The pointer where to place argc.
+ * @param argv The pinter where to place argv.
+ * @return true if successful.
+ * \note Currently, this function is only available on Unix with
+ * early injection.
+ */
+bool
+dr_get_app_args(OUT int **argc, OUT char ***argv);
```
derekbruening, line 1820: I'm not sure we resolved the discussion from PR #2885: https://github.com/DynamoRIO/dynamorio/pull/2885#discussion_r175331626

Reviving it here:

On Windows the command line is in RTL_USER_PROCESS_PARAMETERS.CommandLine as a UNICODE_STRING. It is not parsed into tokens and there is no count of tokens: so there's no `argc`, and the UNICODE_STRING is a data structure with a size and a pointer to content.  What is this function going to do on Windows?  Will the interface have to change?  Is `argv` going to point to the UNICODE_STRING struct, or to the array of UTF-16 characters its field points to?  Neither of those is anything like a `char **` array of *pointers* to token strings.  Should this at least have a type of `void *`?

johnfxgalea: Didn't think much about it but how about: 1) we forget about argc 2) change the function's prototype to dr_get_app_args(OUT void **args, OUT size_t size) 3) for linux we pass the array of args and size being equal to argc, 4) for windows pass an array with one element, namely UNICODE_STRING, and set size to 1?

derekbruening: OK, but I would tweak it for Windows since UNICODE_STRING is a native-library-level data structure and not exposed at the Windows API level: I would point the data at the UTF-16 string and set the size to the size of that string (in bytes, not chars (yes it's different since it's UTF-16): just read the struct field).

derekbruening: Seems like a reasonable solution.  STGM.

johnfxgalea: > I would point the data at the UTF-16 string and set the size to the size of that string (in bytes, not chars (yes it's different since it's UTF-16): just read the struct field).

But then we would have size being used for different purposes. On Unix size would denote the count of args, and on Windows it would denote the number of bytes. I don't think that is what one wants. Can't we rely on NULL terminations to get the length?  s/size/count -> dr_get_app_args(OUT void **args, OUT size_t *count)

core/unix/os.c
johnfxgalea: Done.

core/unix/os.c
johnfxgalea: Done.

suite/tests/CMakeLists.txt
johnfxgalea: Done.

suite/tests/client-interface/app_args.c
johnfxgalea: Done

suite/tests/client-interface/app_args.dll.c
johnfxgalea: Done.

core/os_shared.h
johnfxgalea: Done.

core/lib/instrument_api.h
derekbruening: > On Unix size would denote the count of args, and on Windows it would denote the number of bytes.

Let's make UNIX be the number of bytes too, then.  It's like the "memory blob" discussed previously: `dr_get_app_args` returns a region of memory that contains the application command line, in the form of the start address and the size in bytes.  The caller has to interpret it on a per-platform basis (which is not ideal: DR normally tries to help the user write cross-platform code).

Maybe the name should be `dr_get_app_arg_loc` (or spell out `location`) to make it clearer, and to allow for adding a `dr_get_app_args` in the future which tokenizes WIndows.

derekbruening: > Can't we rely on NULL terminations to get the length?

No, UNICODE_STRING strings are not always NULL-terminated.  Usually, but we have seen enough that aren't that it is not to be relied upon.

johnfxgalea: Then how about the function returns an array of a new struct, something like dr_app_args_t, which consists of a string (char *) and the string's size in bytes (size_t). Maybe dr_get_app_args(dr_app_args_t *args, size_t *count).

Very easy to tokenise for Unix. Perhaps a bit challenging for Windows due to UTF-16 but the client is going to have to tokenise anyways if we just return the UTF-16 string. Might as well we bite the bullet and alleviate this burden for the user. 

Any idea how this UTF-16 string on windows structured please? Is it easy to parse?

derekbruening: To clarify: so this is going to provide an array of argument token strings for both platforms, but it's trying to avoid making a copy, and that's why it's using a size and not null-terminating?  I think each token will be hard to work with if it's not null-terminated: regular C string routines won't work.  I would think we'd return an array of null-terminated strings, which would be copies on Windows.

Should the Windows copies be converted to UTF-8 so the type can just be "char *"?  The core already has code to convert back and forth.

The Windows string is what you would expect: the full command line just all run together with spaces.  We'd want to do what sthg like https://docs.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-commandlinetoargvw does.  I think the challenging part is identifying unicode spaces and quotes without using library routines -- or finding ntdll library routines we can prove are safe to use.  Or maybe moving this to drx and trying to use that shell32 routine, but dynamically so we don't pull it in every time??  Or just implementing an ASCII-oriented version and documenting the limitations?

derekbruening: Maybe it's fine to limit to ASCII spaces and quotes: DR is already doing that in some places and has our_isspace.  I don't know if we'll run into irregular spaces.  Yeah, I would say, let's just do that for now.

derekbruening: So I am suggesting:
+ Allocate memory for copies converted to UTF-8 on Windows: so need some way to free
+ Only support parsing based on ASCII spaces and quotes

johnfxgalea: Thanks. I understood what you mean. 

> To clarify: so this is going to provide an array of argument token strings for both platforms, but it's trying to avoid making a copy, and that's why it's using a size and not null-terminating?

Yes. In particular, I don't want to lose the tokens (on Unix) by just returning an opaque blob.

>  I think each token will be hard to work with if it's not null-terminated: regular C string routines won't work. I would think we'd return an array of null-terminated strings, which would be copies on Windows.

I propose the following: 
1) Apart from the pointer and the size, add a new boolean field to the dr_app_args_t struct, something like `is_utf_16`.
2) Create an additional API function, with a prototype  like dr_app_arg_as_string(IN dr_app_args_t  *app_arg, OUT char *buf_out, IN size_buf_size) This function takes an app_arg from the array which the user obtained from dr_get_app_args(), and sets buf_out with the corresponding null-terminated string. 

The function will do the copying and the  UTF-8 conversation (as you mentioned) when `is_utf_16` is true and notably will avoid issues regarding the freeing of the string buffer because this buffer is supplied by the user. For Unix, the function simply calls a strncpy.

The only downside with the approach is that dr_get_app_args() will also do the copying and conversion. on Windows, purely for the sake of parsing. But the upside is that the user gets to know the locations of args on an individual basis. 

> 
> Should the Windows copies be converted to UTF-8 so the type can just be "char *"? The core already has code to convert back and forth.
> 
> The Windows string is what you would expect: the full command line just all run together with spaces. We'd want to do what sthg like https://docs.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-commandlinetoargvw does. I think the challenging part is identifying unicode spaces and quotes without using library routines -- or finding ntdll library routines we can prove are safe to use. Or maybe moving this to drx and trying to use that shell32 routine, but dynamically so we don't pull it in every time?? Or just implementing an ASCII-oriented version and documenting the limitations?

derekbruening: > The only downside with the approach is that dr_get_app_args() will also do the copying and conversion on Windows, purely for the sake of parsing.

We should be able to parse to find the token boundaries without copying.

derekbruening: OK so this seems good to me: one function which returns a list of boundaries and encoding of each token, and another which converts to utf-8.  Maybe the only thing to settle is how important the 2nd function is: the user could do that on their own today with dr_snprintf and the `%S` specifier.  Maybe writing some sample client code would help decide?  Maybe the 2nd function should return a char* so it can return a pointer to the original and not do a memcpy (documenting this of course), so it would return either a pointer to the original or to the passed-in buffer?  That might make user code simpler.

johnfxgalea: > I think the challenging part is identifying unicode spaces and quotes without using library routines -- or finding ntdll library routines we can prove are safe to use

> We should be able to parse to find the token boundaries without copying.

I thought we needed to convert to utf-8 to help identify spaces? or did I misunderstand that part?

derekbruening: If we want to pass to some library routine that takes utf-8: but if we limit to ascii we can identify in utf-16.

This is what it is sounding like to me, cross-platform code which as no memcpy on UNIX:

```C
    dr_app_arg_t arg_info[MAX_ARGS];
    size_t num = dr_get_app_args(&arg_info, sizeof(arg_info));
    for (int i = 0; i < num; ++i) {
        char arg_buf[MAX_ARG_SIZE];
        const char *arg_str = dr_app_arg_as_utf8(&arg_info[i], sizeof(arg_buf));
        /* now examine arg_str with regular strcmp, etc. functions */
    }
```

johnfxgalea: Yep, we are on the same page.

johnfxgalea: Done.

johnfxgalea:
@derekbruening Please take a look when you get a chance.

api/docs/release.dox
```diff
@@ -148,6 +148,8 @@ Further non-compatibility-affecting changes include:
  - Added drmgr_register_opcode_instrumentation_event() and
    drmgr_unregister_opcode_instrumentation_event() so that drmgr supports
    opcode event instrumentation.
+ - Added dr_get_app_args() and dr_app_arg_as_utf8()  to enable the gathering
+   of application command-line args on Unix systems.
```
derekbruening, line 152: Not just UNIX

core/lib/globals_shared.h
```diff
@@ -2017,4 +2017,30 @@ typedef struct _priv_mcontext_t {
 #include "mcxtx.h"
 } priv_mcontext_t;
 
+/* DR_API EXPORT BEGIN */
+/**
+ * Encodings of an application's argument.
+ */
+typedef enum {
+    APP_ARG_ASCII,
```
derekbruening, line 2025: I think this can be removed.  I think any modern UNIX environment is going to encode as UTF-8 or ISO-8859* or somesuch.  I think we want to be agnostic of which: we just need to know whether we can treat it as a null-terminated char* or not (i.e., it has no encoding which contains a zero byte).  Not sure if there's a standard label for that: `APP_ARG_CSTR_COMPAT`?  So I'm suggesting just having the cstring-compat and UTF-16.

Hmm: so dr_app_arg_as_utf8() is not as accurately named: maybe dr_app_arg_as_cstring()??

core/unix/os.c
```diff
@@ -295,6 +295,13 @@ static app_pc executable_end = NULL;
 static char executable_path[MAXIMUM_PATH];
 static char *executable_basename;
 
+/* Pointers to arguments. Refers to the main stack set up by the kernel.
+ * these are only written once during process init and we can live with
```
derekbruening, line 299: s/these/These/

core/lib/globals_shared.h
derekbruening, line 2025: To avoid namespace conflcts when linked statically into apps, let's follow the convention of having a DR_ prefix on new global symbols.

core/lib/globals_shared.h
```diff
@@ -2017,4 +2017,30 @@ typedef struct _priv_mcontext_t {
 #include "mcxtx.h"
 } priv_mcontext_t;
 
+/* DR_API EXPORT BEGIN */
```
derekbruening, line 2020: Can we limit this to narrower scope than globals_shared.h: I see it can't be limited to instrument.c, but how about os_shared.h?

core/lib/globals_shared.h
```diff
@@ -2017,4 +2017,30 @@ typedef struct _priv_mcontext_t {
...
+/**
+ * Encodings of an application's argument.
+ */
+typedef enum {
+    APP_ARG_ASCII,
+    APP_ARG_UTF_8,
+    APP_ARG_UTF_16,
+} dr_app_arg_encoding_t;
+
+typedef struct _dr_app_arg_t {
```
derekbruening, line 2030: Missing doxygen comment (so it won't show up in the docs)

api/docs/release.dox
derekbruening: OK it's NYI on Windows -- maybe reword to say it's only currently implemented for UNIX

core/lib/instrument_api.h
```diff
@@ -1806,6 +1806,32 @@ DR_API
 const char *
 dr_get_application_name(void);
 
+DR_API
+/**
+ * Provides information about the app's arguments by setting \p args_buf up to
+ * the size \p buf_size. Returns the number of args set or -1 on error.
```
derekbruening, line 1812: We should provide a way for the user to know how many args there are so it can resize and try again.  Should it return the total # of args even if that's more than the passed-in?  Or at least provide a specific error reason so the caller knows whether to retry w/ a bigger buffer or give up.

core/lib/instrument_api.h
```diff
@@ -1806,6 +1806,32 @@ DR_API
...
+ */
+int
+dr_get_app_args(OUT dr_app_arg_t *args_buf, int buf_size);
+
+DR_API
+/* Returns the passed argument as a string. \p buf is used only if needed, and
+ * therefore the caller should not assume that the string is in the \p buf. In other
+ * words, always use the returned value to refer to the string. Returns NULL on error
+ * such as when \p buf is needed as storage and the size of the buffer \p buf_size
+ * is not sufficient.
```
derekbruening, line 1827: Is there a way to tell the caller how much space is needed here as well, or whether to blindly double or give up.

core/lib/instrument_api.h
johnfxgalea: I think this should be a separate API function: dr_num_app_args()

core/lib/globals_shared.h
johnfxgalea: If I remember correctly, I tried to do this, but instrument_api.h does not include os_shared.h and was reluctant to add it just for dr_app_arg_t.

core/lib/instrument_api.h
johnfxgalea: You can obtain the original size from dr_app_arg_t. This size should be sufficient no?

core/lib/globals_shared.h
derekbruening: instrument_api.h does transitively include os_shared.h: os_exports.h includes os_shared.h and is included by globals.h (limited to core; globals_shared is used by non-core libs) so it should be visible.  Or do you mean it doesn't get published to the API headers by genapi.pl?  I see existing doxygen things in there so it should be.

core/lib/instrument_api.h
derekbruening: For Windows, that's the UTF-16 size.  The UTF-8 size is hard to predict from there: you can upper-bound it by 1.5x I guess, but if it's all ascii chars you'll have a 3x-too-big buffer.  Maybe not a big deal.  Of course that means the impl here doesn't know the size unless it does an extra pass through: which io.c supports (utf16_to_utf8_size()).  We could call that extra pass only if we run out of space to avoid wasted work.

johnfxgalea: > We could call that extra pass only if we run out of space to avoid wasted work.

Do mean the API function does this pass or is the pass the responsibility of the caller?

derekbruening: The API function.  But if typical usage is to re-use a single buffer for every arg maybe an upper bound is enough?  Maybe just document that the user should allocate a buffer 1.5x the largest dr_app_arg_t.size.  And also have a separate return value for buffer too small?  We never did establish separate error value conventions in the core like most of the extensions have.  Another thing to do differently if we redesigned it all.

johnfxgalea: Doc sounds good. So we can add an OUT bool *flag or rely on the generic return of NULL, which can be returned due to other errors as well. Assuming that the user does nothing weird like passing a NULL buf ptr, then a NULL will typically represent lack of sufficient size. Not really an elegant solution, I know, but which do you prefer?

derekbruening: Hmm, and I guess the 3rd choice is to return an error code with the buffer return as an output parameter instead (drmgr/drreg-style), but since no other DR function does that maybe that seems weird.  With the 2nd (output param error code; your "bool *flag" but I'd make it in int error code) or 3rd we'd add a set of error codes: maybe start a trend for new DR API routines to have error codes.  4th choice would be a separate errno()-style function: which we also do for some other interfaces like drcachesim analyzers.

Since we're retrofitting to the existing DR API: I would lean toward choice 4, since we could use that for existing routines too if we want to w/o affecting compatibility.

So sthg like:

`int dr_get_error_code()` which returns the error code of the last DR API routine that failed for the current thread (avoid races: make it per-thread).

And a `DR_ERROR_*` enum?

WDTY?  It may be confusing at first to have this and have only one function using it...but we have to start somewhere.

derekbruening: If we add `dr_get_error_code` I would make it *not* be written on success: only on failure?

johnfxgalea: This does sound like a good approach. Performance is just my first quick concern - due to the access of thread local storage.

johnfxgalea: > If we add `dr_get_error_code` I would make it _not_ be written on success: only on failure?

How do you avoid confusion from a previous failure if you do not update?

derekbruening: The user has to first check for failure in the API routine they just called, using its normal fail/success return value.

johnfxgalea: Right, so exactly like errno. Cool, makes sense to me.

johnfxgalea: Where is this error code going to be stored please? Recall, we don't have access to dormer's tls array.

derekbruening:
Dcontext->client_data

core/lib/instrument_api.h
derekbruening: Was AFK.  A new field inside dcontext->client_data.

api/docs/release.dox
johnfxgalea: Done.

core/lib/globals_shared.h
johnfxgalea: Done.

core/unix/os.c
johnfxgalea: Done.

core/lib/globals_shared.h
johnfxgalea: Done.

core/lib/globals_shared.h
johnfxgalea: Done.

core/lib/globals_shared.h
johnfxgalea: Done.

core/lib/instrument_api.h
johnfxgalea: Done.

core/lib/instrument_api.h
johnfxgalea: Sorry for the delay. I had a longer and busier week than usual. 

I still need to add the error coding handling. This is the only thing pending. I'll try to add this soon.

derekbruening: Sure, no rush.

johnfxgalea: Done.

johnfxgalea:
I think this is ready for review. No rush.

derekbruening:
> I think this is ready for review. No rush.

Meant to get to it today but did not happen...tomorrow.

johnfxgalea:
No problem. Not in a particular hurry.

core/globals.h
```diff
@@ -308,6 +308,17 @@ typedef struct _dr_stats_t {
     uint64 synchs_not_at_safe_spot;
 } dr_stats_t;
 
+/**
+ * Error codes of DR API routines.
+ */
+typedef enum {
+    DR_ERROR_NONE,
+    DR_ERROR_INVALID_PARAM,
```
derekbruening, line 316: suggest: Spell out PARAMETER. That seems nicer IMHO, but more importantly it will match drreg, drsyms, drcovlib, and drbbdup.

core/globals.h
```diff
@@ -308,6 +308,17 @@ typedef struct _dr_stats_t {
     uint64 synchs_not_at_safe_spot;
 } dr_stats_t;
 
+/**
+ * Error codes of DR API routines.
+ */
+typedef enum {
+    DR_ERROR_NONE,
```
derekbruening, line 315: Hmm so if this were a return value I would definitely want DR_ERROR_SUCCESS.  I think all the ext's have some kind of _SUCCESS, and so does WIndows:
https://docs.microsoft.com/en-us/windows/win32/debug/system-error-codes--0-499-.  Did you deliberate pick sthg different b/c it's not a return value?  The Windows one isn't either.  I would still vote for _SUCCESS for consistency.

derekbruening, line 315: These should all have doxygen docs.

core/globals.h
```diff
@@ -308,6 +308,17 @@ typedef struct _dr_stats_t {
...
+    DR_ERROR_INVALID_PARAM,
+    DR_ERROR_INSUFFICIENT_SPACE,
+    DR_ERROR_APP_ENCODING,
```
derekbruening, line 318: Hard to know what this means from the name (see above: should have doc).  Maybe it should be DR_ERROR_UNKNOWN_ENCODING?

core/lib/instrument.c
```diff
@@ -2784,6 +2784,53 @@ dr_get_application_name(void)
 #    endif
 }
 
+dr_error_code_t
+dr_get_error_code(void)
+{
+    dcontext_t *dcontext = get_thread_private_dcontext();
```
derekbruening, line 2790: Hmm, would we ever want to pass in the dcontext to support getting on another thread (also slight perf opt to pass already-obtained dcontext)?  While the last error does seem like an execution thread thing unlikely to involve the non-executing thread are we going to regret it later.  Hmm.  It does seem like there is very unlikely to be a use case that wouldn't involve the currently executing thread.

core/lib/instrument.c
```diff
@@ -2784,6 +2784,53 @@ dr_get_application_name(void)
...
+    return dcontext->client_data->error_code;
+}
+
+int
+dr_num_app_args(void)
+{
+    /* XXX i#2662: Add support for Windows. */
```
derekbruening, line 2797: Should this set the error code to NOT_IMPLEMENTED?

OK I see there's an ASSERT* below, and this function doesn't say it sets the error code.  Should we try to have all new functions set it?  The ASSERT* are debug-build-only, so we could have both.

core/lib/instrument.c
```diff
@@ -2784,6 +2784,53 @@ dr_get_application_name(void)
...
+    if (app_arg == NULL) {
+        dcontext = get_thread_private_dcontext();
+        dcontext->client_data->error_code = DR_ERROR_INVALID_PARAM;
+        return NULL;
+    }
+
+    switch (app_arg->encoding) {
+    case DR_APP_ARG_CSTR_COMPAT: return (char *)app_arg->start;
+    case DR_APP_ARG_UTF_16:
+        ASSERT_NOT_IMPLEMENTED(false);
```
derekbruening, line 2822: I would have a comment w/ the issue # and say that we have the code to do this (utf16_to_utf8())

core/lib/instrument_api.h
```diff
@@ -1801,11 +1801,63 @@ DR_API
...
+
 DR_API
 /** Returns the image name (without path) of the current application. */
 const char *
 dr_get_application_name(void);
 
+DR_API
+/**
+ * Provides information about the app's arguments by setting \p args_buf up to
+ * the size \p buf_size. Returns the number of args set or -1 on error.
```
derekbruening, line 1824: I would mention that dr_num_app_args() can be used to query the total number of args available.

core/lib/instrument_api.h
```diff
@@ -1801,11 +1801,63 @@ DR_API
...
+int
+dr_num_app_args(void);
+
+DR_API
+/**
+ * Returns the passed argument \p app_arg as a string. \p buf is used only if needed, and
+ * therefore the caller should not assume that the string is in the \p buf. In other
+ * words, always use the returned value to refer to the string. Returns NULL on error
+ * such as when \p buf is needed as storage and the size of the buffer \p buf_size
+ * is not sufficient.
```
derekbruening, line 1851: I would mention how to get an upper bound on the size needed

core/os_shared.h
```diff
@@ -314,6 +314,39 @@ char *
 get_application_pid(void);
 char *
 get_application_name(void);
+
+/* DR_API EXPORT BEGIN */
+/**
+ * Encodings of an application's argument.
+ */
+typedef enum {
+    DR_APP_ARG_CSTR_COMPAT,
```
derekbruening, line 323: These values should have doxygen comments.

core/os_shared.h
```diff
@@ -314,6 +314,39 @@ char *
 get_application_pid(void);
 char *
 get_application_name(void);
+
+/* DR_API EXPORT BEGIN */
+/**
+ * Encodings of an application's argument.
```
derekbruening, line 320: nit: "command-line argument" (else may be confusing if seen in isolation in the docs)

core/unix/os.c
```diff
@@ -1099,6 +1106,45 @@ get_application_short_name(void)
...
+        dcontext_t *dcontext = get_thread_private_dcontext();
+        dcontext->client_data->error_code = DR_ERROR_INVALID_PARAM;
+#endif
+        return -1;
+    }
+
+    int num_args = num_app_args();
+    int min = (buf_size < num_args) ? buf_size : num_args;
+    for (int i = 0; i < min; i++) {
+        args_buf[i].start = (void *)app_argv[i];
```
derekbruening, line 1141: This will crash for late injection?  Check up front and return the not-impl error code.

core/globals.h
johnfxgalea: Indeed. It is not a return value and I didn't want to assign 0 to a legitimate error code. Another option is to remove it completely and assign values to the enum, starting from 1. WDYT?

core/lib/instrument.c
```diff
@@ -2784,6 +2784,53 @@ dr_get_application_name(void)
...
+}
+
+const char *
+dr_app_arg_as_cstring(IN dr_app_arg_t *app_arg, char *buf, int buf_size)
+{
+    dcontext_t *dcontext;
+
+    if (app_arg == NULL) {
+        dcontext = get_thread_private_dcontext();
+        dcontext->client_data->error_code = DR_ERROR_INVALID_PARAM;
```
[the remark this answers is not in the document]
johnfxgalea: Good idea.

core/unix/os.c
```diff
@@ -1099,6 +1106,45 @@ get_application_short_name(void)
...
+    if (args_buf == NULL || buf_size < 0) {
+#ifdef CLIENT_INTERFACE
+        dcontext_t *dcontext = get_thread_private_dcontext();
+        dcontext->client_data->error_code = DR_ERROR_INVALID_PARAM;
+#endif
+        return -1;
+    }
+
+    int num_args = num_app_args();
+    int min = (buf_size < num_args) ? buf_size : num_args;
```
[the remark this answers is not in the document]
johnfxgalea: No, buf_size is not in bytes but denotes the number of dr_app_arg_t values.

core/globals.h
derekbruening: I think you've misread my comment: I was asking to s/DR_ERROR_NONE/DR_ERROR_SUCCESS/.  I then rambled on about how maybe you wanted NONE instead of SUCCESS because it wasn't a direct return value, but that I would still vote for SUCCESS to match our other error enums and external things like Windows API.

johnfxgalea: Yep, I understood you regarding the substitute.  I am now wondering whether or not you prefer that I remove this entry completed instead?

derekbruening: But what would dr_get_error_code() return when there was no error?  Doesn't there need to be a no-error code?

johnfxgalea: I was thinking of leaving it undefined. dr_get_error_code() should be used only upon a failed API routine therefore  DR_ERROR_NONE/DR_ERROR_SUCCESS _should_ never be returned. Consequently, I was wondering whether we should eliminate the entry from the user interface completely.

derekbruening: Hmm, I see.  OK, starting from 1 seems reasonable.  Then we could add it back as 0 if we ever needed it for some reason.

core/globals.h
johnfxgalea: Done.

core/globals.h
johnfxgalea: Done.

johnfxgalea: Done.

core/globals.h
johnfxgalea: Done.

core/lib/instrument.c
johnfxgalea: Done.

core/lib/instrument.c
johnfxgalea: Done.

core/lib/instrument.c
johnfxgalea: Done.

core/lib/instrument_api.h
```diff
@@ -1801,11 +1801,63 @@ DR_API
 bool
 dr_set_client_version_string(const char *version);
 
+DR_API
+/**
+ * Returns the error code of the last failed API routine. Users should check whether or
+ * not the API routine that they just called has failed prior to this function.
```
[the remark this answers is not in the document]
johnfxgalea: Done.

core/lib/instrument_api.h
johnfxgalea: Done.

core/lib/instrument_api.h
johnfxgalea: Done.

core/os_shared.h
johnfxgalea: Done.

core/os_shared.h
johnfxgalea: Done.

core/unix/os.c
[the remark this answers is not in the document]
johnfxgalea: I modified the docs to make this clear. Done?

core/unix/os.c
johnfxgalea: Done.

core/lib/instrument.c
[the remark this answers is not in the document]
johnfxgalea: Done.

johnfxgalea:
@derekbruening PTAL when you get a chance.

core/globals.h
```diff
@@ -312,11 +312,22 @@ typedef struct _dr_stats_t {
  * Error codes of DR API routines.
  */
 typedef enum {
-    DR_ERROR_NONE,
-    DR_ERROR_INVALID_PARAM,
-    DR_ERROR_INSUFFICIENT_SPACE,
-    DR_ERROR_APP_ENCODING,
-    DR_ERROR_NOT_IMPLEMENTED,
+    /**
+     * Invalid parameter passed to the API routine.
```
derekbruening, line 316: (Note that these short comments could be the `/**<  */` after-field same-line style.  Not requesting any change: just in case you weren't aware.)

core/unix/os.c
```diff
@@ -1087,6 +1087,15 @@ get_application_name_helper(bool ignore_cache, bool full_path)
     return (full_path ? executable_path : executable_basename);
 }
 
+#ifdef CLIENT_INTERFACE
+void
+set_client_error_code(dr_error_code_t error_code)
```
derekbruening, line 1092: I would expect this to be in instrument.c shared between OS's rather than duplicated in each OS impl file.

derekbruening, line 1092: Should this take in the dcontext, since many routines already have it, and allow NULL or GLOBAL_DCONTEXT (berst to check for both to make it easy for callers to blindly pass their local var dcontext) in which case it calls get_thread_private_dcontext()?

core/unix/os.c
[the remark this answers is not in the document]
derekbruening: Clearer, but how about changing the names to avoid any confusion: s/args_buf/args_array/ and s/buf_size/args_count/ or s/buf_size/args_max_count/ (since "args_count" doesn't always sounds like a max capacity)?  Then looking at just this code w/o reading the doc is easier.

core/globals.h
johnfxgalea: Yep, just that personally I prefer this style because the text looks nicer. With same-line style one will end up with a mix of single line and multi-line comments depending on the comments' lengths. Going with multi-line immediately maintains consistency :)

core/unix/os.c
johnfxgalea: Hmm, so for the cases considered in this PR, actually, drcontext is not readily available. Getting the drcontext via this function reduces code. 

Also, most of the time, drcontext is obtained only upon error. Therefore harnessing the getting of drcontext out of the branch where set_client_error_code() would be called would result in a slight perf hit.

derekbruening: For these, they don't have dcontext, true: though I think that most other routines do?

> Therefore harnessing the getting of drcontext out of the branch where set_client_error_code() would be called results in a slight perf hit.

Not sure I follow: the caller of set_client_error_code() only passes dcontext if they have it readily available in a local; else they pass NULL and this function looks it up for them.

Seems fine to leave it as it is w/o taking in the dcontext: since internal we can always change it later if we end up with a whole bunch of calls who have the dcontext.

core/globals.h
derekbruening: OK sure

core/unix/os.c
johnfxgalea: Ah, I see. So you mean enable the passing of drcontext as a param but if it is set to NULL, have set_client_error_code() call get_thread_private_dcontext().

johnfxgalea: Right, but I assume keeping the declaration in os_shared.h right? Otherwise, os.c cannot use it.

johnfxgalea: Done.

johnfxgalea: Done.

core/unix/os.c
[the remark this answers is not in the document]
johnfxgalea: Done.

core/unix/os.c
johnfxgalea: Actually, I am not sure how I can include set_client_error_code() in instrument.c and still use the function to set error codes in os.c and ntdll.c. Essentially, the function needs to have the same scope as get_thread_private_dcontext(). I tried moving the implementation to instrument.c and leaving the declaration in the os_shared.h but got a linking error on Appveyor.

I could not find a .c file that shares code between different OS's and is of lower dependency and instrument.c

derekbruening: > Ah, I see. So you mean enable the passing of drcontext as a param but if it is set to NULL, have set_client_error_code() call get_thread_private_dcontext()?

Yes, and to also do so for GLOBAL_DCONTEXT so callers don't have to.

derekbruening: In the old days, a client_data_t operation like this would be declared in instrument.h and implemented in instrument.c.  os.c does include instrument.h.  If deeper parts of the core with no existing client interaction like ntdll.c needed to be used, a different layer would exist between ntdll.c and os.c which would use a different non-CLIENT_INTERFACE way to indicate errors, and the API function would convert that.

That said, these days nobody ever makes a non-CLIENT_INTERFACE build, we just live with whatever perf impact the define has, and we should probably clean up and simplify the code by removing the define altogether: #2971.

But does the Windows impl need to be in ntdll.c?  You could just put it in os.c and if os.c needs some ntdll data it can call into ntdll.c, and then have it in instrument.h and instrument.c  like the rest of the core client operations.

johnfxgalea: Okay, understood. Thank you.

johnfxgalea:
@derekbruening I think it is ready. I added the check for GLOBAL_DCONTEXT and moved the implementation to instrument.c. No need for a final check, but let me know if you want to do so before I merge.

Issue #4299, opened 2020-05-15
Bug in jcxz/jecxz detection code


nextsilicon-itay-bookstein:
**Describe the bug**
Running x64 code containing jecxz (and, to my understanding, x86 code containing jcxz) leads to incorrect behavior/crashes due to subtly broken pattern-detection code in `core/arch/x86/instr.c`. In particular, when running an application that uses libomp under debug drrun, the following assert is hit because the generated jecxz+jmp+jmp are tagged as meta instructions, and therefore are not detected as the second exit-cti:
https://github.com/DynamoRIO/dynamorio/blob/d275a9aee4e2260cda0b7ce3f5e47f6f1b668320/core/arch/interp.c#L6187-L6189

The particular piece of code (instruction, even) that triggered this when running under DynamoRIO is this:
https://github.com/llvm/llvm-project/blob/3a76b8a538c0d20efe823f03a133c778e842270e/openmp/runtime/src/z_Linux_asm.S#L1112

**To Reproduce**
I'm not sure about the precise _minimal_ conditions under which the problem reproduces, because when I attempted to yank out the above assembly routine to a test app and run it, I did not encounter the bug. The repro I debugged involved running an application that uses OpenMP against a locally installed llvm-7 libomp.so. I suspect that I need to somehow trigger cbr inversion for that particular branch for the assert/crash to happen. But I've already got a fix and will open a PR attached to this issue.

The problem (ASSERT_NOT_REACHED) reproduces under vanilla drrun without any clients.
I used -debug -loglevel 4 to diagnose the issue.

**Expected behavior**
No ASSERTs, application runs to completion successfully.

**Versions**
ASSERTs in freshly compiled (14/05/2020) master.

derekbruening:
Thank you for the detailed filing and PR.  Xref #508, 8ab9344b23b2053c71e7241027e14dc54a1744bb which put in the addr prefix handling but apparently only for pc==NULL.

nextsilicon-itay-bookstein:
Glad to have helped :)

derekbruening:
Ideally a test would be added but it sounds like it was difficult to arrange so I went ahead and marked PR #4300 as fixing the bug.

PR #4300, opened 2020-05-15
i#4299: Fix cti_short_rewrite detection and inversion for addr-prefixed jcxz/jecxz

nextsilicon-itay-bookstein:
In x86, jcxz has an 0x67 prefix. In x64, jecxz has an 0x67 prefix.
The code in instr_is_cti_short_rewrite() does not take that into
account when the pc argument is not NULL.
The code in instr_invert_cbr does not take that into account when
verifying and inverting the cbr, using wrong offsets in that case.
This fix checks for the 0x67 prefix in both unaddressed cases so
that the checks against the instruction bytes pass as they should.

Issue: #4299


derekbruening:
run arm tests

PR #4302, opened 2020-05-19
i#4301: Fix x86-32 Linux stack alignment before dynamo_process_exit

### LINKED ISSUE #4301: Call dynamorio_process_exit is not 16 byte aligned (Linux x86-32)
Hi!

I run my x86-32 DynamoRIO client on Linux (Release build). I get unaligned stack here:
```
0xf7eb81f3 <cat_done_saving_dstack+16>  call   0xf7ddfaf0 <dynamo_process_exit>

#13 0xf7e3fc26 in instrument_exit_event () at /data/predicate_solving/dynamorio/core/lib/instrument.c:871
#14 0xf7dcbdd0 in dynamo_process_exit () at /data/predicate_solving/dynamorio/core/dynamo.c:1554
#15 0xf7ea41f8 in cat_done_saving_dstack ()
#16 0xf7ea4257 in cat_have_lock ()
```

This leads to further crash in my client due to XMM usage.
```
0xf663d332 <_ZN15prime_generator22process_next_k_numbersEy+290> movaps XMMWORD PTR [ebp-0xe8],xmm3
```

SweetVishnya:
Closes #4301

### DIFF
--- a/core/arch/x86/x86.asm
+++ b/core/arch/x86/x86.asm
@@ -567,7 +567,7 @@ GLOBAL_LABEL(cleanup_and_terminate:)
         mov      [4*ARG_SZ + REG_XBP], ARG4
 #else
         mov      REG_XBP, REG_XSP
-# if defined(MACOS) && !defined(X64)
+# ifdef UNIX
         lea      REG_XSP, [-3*ARG_SZ + REG_XSP] /* maintain align-16: offset retaddr */
 # endif
 #endif
@@ -602,7 +602,7 @@ cat_done_saving_dstack:
         /* avoid sygate sysenter version as our stack may be static const at
          * that point, caller will take care of sygate hack */
         CALLC0(GLOBAL_REF(get_cleanup_and_terminate_global_do_syscall_entry))
-#if defined(MACOS) && !defined(X64)
+#if defined(UNIX) && !defined(X64)
         lea      REG_XSP, [-2*ARG_SZ + REG_XSP] /* maintain align-16 w/ 2 pushes below */
 #endif
         push     REG_XBX /* 16-byte aligned again */
@@ -649,7 +649,7 @@ cat_have_lock:
         mov      REG_XSI, [2*ARG_SZ + REG_XBP]  /* sysnum */
         pop      REG_XAX             /* syscall */
         pop      REG_XCX             /* dstack */
-#if defined(MACOS) && !defined(X64)
+#if defined(UNIX) && !defined(X64)
         lea      REG_XSP, [2*ARG_SZ + REG_XSP] /* undo align-16 lea from above */
 #endif
         mov      REG_XBX, REG_XBP /* save for arg access after swapping stacks */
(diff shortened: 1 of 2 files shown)


core/arch/x86/x86.asm
```diff
@@ -604,6 +604,9 @@ cat_done_saving_dstack:
         CALLC0(GLOBAL_REF(get_cleanup_and_terminate_global_do_syscall_entry))
 #if defined(MACOS) && !defined(X64)
         lea      REG_XSP, [-2*ARG_SZ + REG_XSP] /* maintain align-16 w/ 2 pushes below */
+#endif
+#if defined(LINUX) && !defined(X64)
+        lea      REG_XSP, [-ARG_SZ + REG_XSP] /* maintain align-16 w/ 2 pushes below */
```
derekbruening, line 609: If you could help me understand: why is Linux different from Mac?  At a glance aren't they the same below with 2 pushes?

core/arch/x86/x86.asm
```diff
@@ -649,6 +652,9 @@ cat_have_lock:
         mov      REG_XSI, [2*ARG_SZ + REG_XBP]  /* sysnum */
         pop      REG_XAX             /* syscall */
         pop      REG_XCX             /* dstack */
+#if defined(LINUX) && !defined(X64)
+        lea      REG_XSP, [ARG_SZ + REG_XSP] /* undo align-16 lea from above */
+#endif
```
derekbruening, line 657: I wonder if you could add an exit event that calls check_stack_alignment() to suite/tests/client-interface/cleancall.dll.c and confirm it crashes without this fix? Then we'll have a regression test.

SweetVishnya: How can I run this specific test?

core/arch/x86/x86.asm
SweetVishnya: It seems to me that stack is misaligned for previous call too. For instance, `get_cleanup_and_terminate_global_do_syscall_entry`. I need further investigation.

SweetVishnya: I will continue on this pull request after some sleep.

core/arch/x86/x86.asm
derekbruening: `ctest -V -R 'client\.cleancall$'`

core/arch/x86/x86.asm
derekbruening: Sure, np

SweetVishnya:
Done.

core/arch/x86/x86.asm
```diff
@@ -567,7 +567,7 @@ GLOBAL_LABEL(cleanup_and_terminate:)
         mov      [4*ARG_SZ + REG_XBP], ARG4
 #else
         mov      REG_XBP, REG_XSP
-# if defined(MACOS) && !defined(X64)
+# ifdef UNIX
```
derekbruening, line 570: So this changes the 64-bit behavior: is that intentional?  You found that the 64-bit alignment was off too?

core/arch/x86/x86.asm
```diff
@@ -1791,11 +1791,11 @@ GLOBAL_LABEL(dr_setjmp:)
         /* PR 206278: for try/except we need to save the signal mask */
         mov      REG_XDX, ARG1
         push     REG_XDX /* preserve */
-# if defined(MACOS) && !defined(X64)
+# ifndef X64
```
derekbruening, line 1794: This and the #if below used to be Mac-only and unlike all the other changes in this PR here it's expanded to Windows too: is that intentional?  32-bit Windows has no stack alignment beyond 4 in its ABI.  I expected all the changes to be s/MACOS/UNIX/?

core/arch/x86/x86.asm
SweetVishnya: This affects only 32-bit, it is the `#else` of `#ifdef X64`.

core/arch/x86/x86.asm
SweetVishnya: this affects only UNIX, see `#ifdef UNIX` above

SweetVishnya:
So, mainly I just exploit nested ifdefs. No need to recheck UNIX. Also, I fixed Windows CI.

AssadHashmi:
run arm tests

core/arch/x86/x86.asm
derekbruening: OK, you can see how from the PR diff view it looks suspect.

core/arch/x86/x86.asm
derekbruening: OK.

SweetVishnya:
OK, I will do so in future.

Fix for Windows CI was adding explicit (void) in function arguments.

PR #4303, opened 2020-05-28
Remove redundant libgcc dependency on 64 bit targets

yury-khrustalev:
Bits from libgcc are not used on AArch64 and x86_64 but are still included in the build and may be linked statically into resulting binary. To avoid any licencing questions on these targets it's easier to just remove this part from the build where it is not used anyway.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -198,7 +198,10 @@ if (UNIX)
   # These are needed for ARM per i#1566 above, but are also needed in some cases
   # for x86, such as 32-bit manipulation of 64-bit integers under clang where
   # __moddi3 is needed.
-  set(CORE_SRCS ${CORE_SRCS} ../third_party/libgcc/udivmoddi4.c)
+  if (ARM OR (X86 AND NOT X64))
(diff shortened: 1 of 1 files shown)


johnfxgalea:
@yury-khrustalev Thank you for the PR.

So that was linked to address any missing __moddi3 (and possibly other variants?) Can this be removed for 64-bit  @derekbruening?

derekbruening:
> @yury-khrustalev Thank you for the PR.
> 
> So that was linked to address any missing __moddi3 (and possibly other variants?) I think this can be removed for 64-bit @derekbruening?

Yes AFAIK this should be fine.

derekbruening:
run arm tests

johnfxgalea:
@yury-khrustalev Thanks!

derekbruening:
> To avoid any licencing questions on these targets 

Note that the original logic for inclusion was that the GPL GCC Runtime Library Exception applies to these libgcc snippets.

AssadHashmi:
> run arm tests

Added yury-khrustalev to AArch64 pre-commit test ACL.

PR #4305, opened 2020-05-29
i#4304: Fix AArch64 asm reachability

### LINKED ISSUE #4304: AArch64 static DR link error: relocation out of range for unexpected_return
```
ld.lld: error: ..._core_arch_aarch64_aarch64.o:(.text+0x2C): relocation R_AARCH64_CONDBR19 out of range: 1508388 is not in [-1048576, 1048575]; references unexpected_return
>>> defined in ..._core_drlibc_drlibc_xarch.o
```

I can solve by moving `unexpected_return` into aarch64.asm.  It is in fact not referenced in drlibc for A64 (only for x86).

derekbruening:
Moves unexpected_return() from drlibc to the core's asm file for
AArch64, to avoid a branch reachability link error when linking DR
statically into large apps.  unexpected_return() is not used in drlibc
for AArch64.

Fixes #4304

### DIFF
--- a/core/arch/aarch64/aarch64.asm
+++ b/core/arch/aarch64/aarch64.asm
@@ -100,6 +100,23 @@ START_FILE
 DECL_EXTERN(dr_setjmp_sigmask)
 #endif
 
+DECL_EXTERN(d_r_internal_error)
+
+/* For debugging: report an error if the function called by call_switch_stack()
+ * unexpectedly returns.  Also used elsewhere.
+ */
+        DECLARE_FUNC(unexpected_return)
+GLOBAL_LABEL(unexpected_return:)
+        CALLC3(GLOBAL_REF(d_r_internal_error), HEX(0), HEX(0), HEX(0))
+        /* d_r_internal_error normally never returns */
+        /* Infinite loop is intentional.  Can we do better in release build?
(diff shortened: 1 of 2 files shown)


derekbruening:
We are now including an x86 issue in #4304 with a different linker relocation error where we used a different workaround of directly building the drlibc files into the shared dynamorio library: but this solution here seems cleaner than repeating that for the static dynamorio library.

derekbruening:
Unfortunately I don't think there's any easy way to reproduce in our github tests: it requires a large target.  We have tests already of using the static libdynamorio and linking it into apps, but those don't hit this problem b/c they fit in the 1MB branch reach.  We'd have to either check in a large test app or somehow generate a large test app with macros or sthg.  Not sure either is worth the effort.  Will put a comment about this in the commit.

xdje42:
Regarding "check in a large test app or ...".
One possibility is:
```
asm ("\n\
	.text\n\
	.global foo\n\
foo:\n\
        ret\n\
	.space 0x100000\n\
");
```
suitably augmented for any needed cleanliness/portability/etc.
[The point being it's easy enough to write a function that consumes that much space. Not sure what was meant by "generate a large test app with macros or sthg".]

PR #4306, opened 2020-05-30
Add cmake avx2 detect function

johnfxgalea:
Adds a cmake helper function that determines whether or not the current processor (at compile time) supports AVX2 (and indirectly BMI2) instructions.


johnfxgalea:
Do you think I should add a separate function for BMI2 or rely on the heuristic that AVX2 processors "typically" also support BMI2?

johnfxgalea:
> I don't know enough about when BMI would not be there. This seems fine: it's just for tests so if there's a problem it's not like it's going to break a user who has deployed a tool based on DR.

But these utility functions are also exposed so that dr clients and external tools can use them. (not just tests). Well, at least, that is how I am using them.

derekbruening:
> > I don't know enough about when BMI would not be there. This seems fine: it's just for tests so if there's a problem it's not like it's going to break a user who has deployed a tool based on DR.
> 
> But these utility functions are also exposed so that dr clients and external tools can use them. (not just tests). Well, at least, that is how I am using them.

OK, but it's still a build time decision, so it seems like it's not going to impact an end user who is not set up to diagnose a problem.

johnfxgalea:
Yes, I agree. Thank you for looking at the PR!

PR #4307, opened 2020-05-31
i4134: Add a pre duplication callback

johnfxgalea:
Enhances drmgr with the support for a new callback to be internally registered by drbbdup. Such callbacks are triggered prior to the duplication process, giving the opportunity for drbbdup users to make any changes/analysis to the basic block before copy generation.

Issue: #4134


ext/drmgr/drmgr_priv.h
```diff
@@ -94,6 +95,26 @@ DR_EXPORT
 bool
 drmgr_unregister_bbdup_event();
 
+DR_EXPORT
+/* Registers a callback that is triggered prior basic block duplication. This gives
```
derekbruening, line 99: prior to

johnfxgalea:
Thank you for the review. Just a wild thought but I am now wondering whether such a call-back should be offered (exposed) by drmgr instead of drbbdup to improve flexibility. Essentially, do we want extensions to become "drbbdup" aware, without depending (linking drbbdup). Could it be that other extensions require basic block manipulation prior to duplication and the ability of registering such a callback via drmgr (instead of drbbdup) makes things easier? Currently, app2app events are triggered for each copy after duplication. hmm

derekbruening:
> Thank you for the review. Just a wild thought but I am now wondering whether such a call-back should be offered (exposed) by drmgr instead of drbbdup to improve flexibility. Essentially, do we want extensions to become "drbbdup" aware, without depending (linking drbbdup). Could it be that other extensions require basic block manipulation prior to duplication and the ability of registering such a callback via drmgr (instead of drbbdup) makes things easier? Currently, app2app events are triggered for each copy after duplication. hmm

Wait: how are app2app after duplication when I thought there was an analysis event that decided how many copies to make?

johnfxgalea:
I extended the interface to allow different basic block duplications to also have different modifications and not just insertion.

By integrating drbbdup with drmgr,  extensions are able to function regardless of whether or not dups are being generated. 

Look at: https://github.com/DynamoRIO/dynamorio/blob/fbe8507d13d19095503783936987fbf39ec0cd67/ext/drmgr/drmgr.c#L1007

ext/drmgr/drmgr_priv.h
johnfxgalea: Done.

johnfxgalea:
@derekbruening I seem to be hitting the __chkstk crash again. (ref: https://github.com/DynamoRIO/dynamorio/pull/4281#issuecomment-623204437) .

I already tried to address the error by reducing EVENTS_STACK_SZ down to 12.
Hmm, I am a bit reluctant to add  #pragma check_stack(off). Maybe I should try a size of 10?

derekbruening:
> @derekbruening I seem to be hitting the __chkstk crash again. (ref: [#4281 (comment)](https://github.com/DynamoRIO/dynamorio/pull/4281#issuecomment-623204437)) .
> 
> I already tried to address the error by reducing EVENTS_STACK_SZ down to 12.
> Hmm, I am a bit reluctant to add #pragma check_stack(off). Maybe I should try a size of 10?

So it's using that much stack?  I would suggest:
+ Confirm it disappears with a higher -stack_size
+ Examine the state in the debugger at the __chkstk crash and list the frame size for each stack frame so we can understand where all the space is going and whether it's just one culprit function we can address or whether the stack size is just too small

johnfxgalea:
Okay, so reducing the buff size on the stack seems to have averted the error. Hmm, checking with a debugger is going to be tricky for me because I do not have a windows machine at hand.

derekbruening:
> Okay, so reducing the buff size on the stack seems to have averted the error. Hmm, checking with a debugger is going to be tricky for me because I do not have a windows machine at hand.

Understood.  Though this is worrisome: it seems like it's on the verge of overflow and if just one small (these entries are small right?) entry is the tipping point on this small test client, we would expect real clients to overflow all the time?

johnfxgalea:
I'm going dump some sizes using dr_fprintf just to get an idea how big these cb entries are. Therefore, no need to check in.

derekbruening:
I will have some time later this week where I could run the test on Windows and try to look in the debugger.

derekbruening:
> @derekbruening I seem to be hitting the __chkstk crash again. (ref: [#4281 (comment)](https://github.com/DynamoRIO/dynamorio/pull/4281#issuecomment-623204437)) .

Revisiting this and looking at the prior one in #4281 which says:

> error LNK2019: unresolved external symbol __chkstk referenced in function drmgr_bb_event

So that was a build-time error.  By "__chkstk crash" here I thought you meant a runtime crash where __chkstk went off the end of the stack?  Do you mean a build-time unresolved symbol?

derekbruening:
What happens is, for any stack allocation larger than one 4096 page, it inserts __chkstk to ensure the code doesn't skip over the guard page.  There's only one guard page, so if you access 4097 bytes from the top of the stack you can crash and not extend the stack (for a normal guard-page-extended stack).  __chkstk just goes and touches each page to the end of your huge stack alloc.  When you shrink the array count you're just getting the frame size down under 4096.

derekbruening:
That said, 4K is pretty big for a DR frame, given the entire stack itself is only 56K.

johnfxgalea:
No, not runtime crash but this error:

 _unresolved external symbol __chkstk referenced in function drmgr_bb_event_

johnfxgalea:
So some background info: Initially EVENTS_STACK_SZ was 16, I reduced it down to 15 for #4281 and issue was resolved. When I added a new cb list in this PR, the issue was resolved only when I dropped it down to size 10.

derekbruening:
So it's the `local_cb_info_t local_info`, right?  That is larger than it was before the recent refactorings for drbbdup?  It might be best to move it to the heap and live w/ the alloc cost?

johnfxgalea:
I am wondering whether or not local_cb_info_t is the actual culprit. Is there something else that is hogging up the stack, and we are resolving this issue indirectly by tweaking local_cb_info_t?

johnfxgalea:
I guess we can leave it at 10 and shift to allocation if problems are encountered in the future?

derekbruening:
> I am wondering whether or not local_cb_info_t is the actual culprit. Is there something else that is hogging up the stack, and we are resolving this issue indirectly by tweaking local_cb_info_t?

What else would it be?  There are not very many locals in drmgr_bb_event:

```C
    dr_emit_flags_t res = DR_EMIT_DEFAULT;
    local_cb_info_t local_info;
    void **pair_data = NULL, **quartet_data = NULL;
    per_thread_t *pt = (per_thread_t *)drmgr_get_tls_field(drcontext, our_tls_idx);

    bool is_dups = false;
```

It seems pretty clear that `local_info` is the only thing there that's going to take up more than a pointer-sized stack slot.

derekbruening:
OK I fired up windbg and here we have the sizes for this PR branch:

```
drmgr!drmgr_bb_event:
00007ff6`7dc53c10 44884c2420      mov     byte ptr [rsp+20h],r9b
00007ff6`7dc53c15 4c89442418      mov     qword ptr [rsp+18h],r8
00007ff6`7dc53c1a 4889542410      mov     qword ptr [rsp+10h],rdx
00007ff6`7dc53c1f 48894c2408      mov     qword ptr [rsp+8],rcx
00007ff6`7dc53c24 4881ec180e0000  sub     rsp,0E18h

0:000> dt -v local_cb_info_t
drmgr!local_cb_info_t
struct _local_ctx_t, 17 elements, 0xd98 bytes
   +0x000 iter_app2app     : struct _cb_list_t, 7 elements, 0x38 bytes
   +0x038 iter_insert      : struct _cb_list_t, 7 elements, 0x38 bytes
   +0x070 iter_instru      : struct _cb_list_t, 7 elements, 0x38 bytes
   +0x0a8 app2app          : [10] struct _cb_entry_t, 5 elements, 0x50 bytes
   +0x3c8 insert           : [10] struct _cb_entry_t, 5 elements, 0x50 bytes
   +0x6e8 instru           : [10] struct _cb_entry_t, 5 elements, 0x50 bytes
   +0xa08 iter_opcode_insert : Ptr64 to struct _cb_list_t, 7 elements, 0x38 bytes
   +0xa10 was_opcode_instrum_registered : Char
   +0xa14 pair_count       : Int4B
   +0xa18 quartet_count    : Int4B
   +0xa1c is_bbdup_enabled : Char
   +0xa20 bbdup_duplicate_cb : Ptr64 to     char 
   +0xa28 bbdup_extract_cb : Ptr64 to     _instr_list_t* 
   +0xa30 bbdup_stitch_cb  : Ptr64 to     void 
   +0xa38 bbdup_insert_encoding_cb : Ptr64 to     void 
   +0xa40 iter_pre_bbdup   : struct _cb_list_t, 7 elements, 0x38 bytes
   +0xa78 pre_bbdup        : [10] struct _cb_entry_t, 5 elements, 0x50 bytes
```

derekbruening:
So the frame is 0xe18==3608 bytes, with 0xd98 ==3480 bytes coming from the local_cb_info_t struct.  For each increase in EVENTS_STACK_SZ we have another 0x50*4==320 bytes.  So +2 to EVENTS_STACK_SIZE and frame will hit 4096.

johnfxgalea:
Okay thanks for checking.

> What else would it be? There are not very many locals in drmgr_bb_event

I thought the cause of the error could also potentially stem from other nested stack frames not local to drmgr_bb_event. However, your debug print shows this is unlikely: 0xd98 is 3480 bytes in size which is quite big.

derekbruening:
> Okay thanks for checking.
> 
> > What else would it be? There are not very many locals in drmgr_bb_event
> 
> I thought the cause of the error could also potentially stem from other nested stack frames not local to drmgr_bb_event. However, your debug print shows this is unlikely: 0xd98 is 3480 bytes in size which is quite big.

No, see the explanation for __chkstk: it is *local* to *one stack frame* and has nothing to do with callers or total stack usage.  It is to ensure that a single frame's stack increase doesn't skip over the guard page.

derekbruening:
If we had a runtime crash in __chkstk, that would be due to total stack usage of all the callers.  But the build-time error for the compiler trying ot insert a __chkstk call is just that one frame.

johnfxgalea:
Okay, understood. Thanks for explaining.

derekbruening:
Here is the compiler flag to control this, /Gs: https://docs.microsoft.com/en-us/cpp/build/reference/gs-control-stack-checking-calls?view=vs-2019  So we could turn it off by setting a larger size than 4096.  Or we could move this struct to the stack.  Since we've already got locks in there and other costs and probably the extra thread-local heap alloc won't cost that much I would lean toward moving the big struct to the heap.

derekbruening:
Can you remind me what `EVENTS_STACK_SZ` is a limit on: the number of events in one phase from all the clients/extensions?

derekbruening:
Is `EVENTS_STACK_SZ` an absolute limit or just an optimization?  cblist_create_local seems to heap-alloc if it doesn't fit in the inlined array.

johnfxgalea:
It's an optimisation to prevent heap allocation. If all registered cbs of an event fit within the list allocated on the stack, then thread-local storage allocation is avoided.

johnfxgalea:
> Can you remind me what `EVENTS_STACK_SZ` is a limit on: the number of events in one phase from all the clients/extensions?

Yes, all clients and extensions.

derekbruening:
Could drmgr_bb_event be on a callstack that goes and calls third-party private libraries?  That's where we're most likely to overflow the 56K stack: sthg like a chain of dbghelp.dll function calls.

derekbruening:
> It's an optimisation to prevent heap allocation.

Hmm, so if we move the whole struct to the heap we've thwarted it partway: I guess we would still avoid the 4 allocs for the 4 separate arrays.

Given that it's an optimization, what if we shrunk it even further to reduce stack usage and kept it how it is on the stack?  What is a typical size in a big multi-extension client?  Maybe run Dr. Memory and see?  I would think it wouldn't be higher than 8?

johnfxgalea:
> Could drmgr_bb_event be on a callstack that goes and calls third-party private libraries? That's where we're most likely to overflow the 56K stack: sthg like a chain of dbghelp.dll function calls.

I think so. Ultimately, drmgr_bb_event will be on the callstack of user-defined callback functions. These can use third-party libraries.

derekbruening:
Could even have diff sizes: app2app seems likely to be smaller than insert and instru.

Was this local stack usage almost this big before the drbbdup refactor that created this struct?

derekbruening:
> Was this local stack usage almost this big before the drbbdup refactor that created this struct?

If it was always close to this size then maybe just leaving it on the stack is fine, maybe shrinking the size to 8 if we've never seen higher than 8.  Add comments by the struct def, EVENTS_STACK_SZ, and function so we remember all of this years from now?

johnfxgalea:
Okay, I'll check how many cbs are registered by Dr. Memory and modify accordingly. Thanks again for checking!

johnfxgalea:
A static count of the number of drmgr_register_bb_instrumentation_event() found in drmemory's code code is **3**. This is excluding extensions. Hmm, I say we leave EVENTS_STACK_SZ  as 10, and lower it down further if we encounter any problems in the future.

I'll go ahead and add comment.

johnfxgalea:
Done; I added the comment. I'll go ahead with the merge unless there are any pending issues.

Issue #3348, opened 2019-01-17
Rename all global symbols to avoid conflicts when used as a static library


derekbruening:
Xref #975.  We're using DR as a static library more and more these days, and the current solution for hiding its global symbols, running `objcopy --localize-hidden`, is a fragile post-link step that conflicts with other build steps such as lld's ICF and which can break when used in complex build configurations.

This issue is a proposal to rename all of DR's global symbols (i.e., everything not file-static) to dramatically reduce the risk of symbol conflicts when statically linked into large applications.

We would leave alone all of the symbols exported as part of the API.  Unfortunately these have a variety of prefixes but at least they have prefixes:

```
$ nm lib64/release/libdynamorio_static.a | grep ' T ' | awk '{print $3}' | perl -p -e 's/^([a-zA-Z0-9]+)_.*/\1/' | sort | uniq -c
     10 decode
      5 disassemble
    311 dr
      3 dynamorio
      3 get
    184 instr
     35 instrlist
     93 opnd
     21 proc
     26 reg
      1 set
      1 __wrap_calloc
      1 __wrap_free
      1 __wrap_malloc
      1 __wrap_realloc
$ nm lib64/release/libdynamorio_static.a | grep ' T ' | awk '{print $3}' | grep '^[gs]et'
get_application_short_name
get_register_name
get_x86_mode
set_x86_mode
```

Moving on to the non-API routines, there are thousands of them, including functions and variables.
Some thoughts on renaming rules:

"dr_": a simple prefix, except today it means "exported for API" and we have
several instances of "dr_foo" exported and "foo" internally: e.g.,  internal "safe_read" and interface "dr_safe_read".  So how do we rename the internal one?

"dri_": where "i" means internal.

Sthg specific to the module?  "drinterp_xxx", "drvm_xxx", "drxl8_xxx", etc.

"dr__" or "__dr_": these violate the C/C++ standards which say no double underscores anywhere, no leading single underscore plus capital letter anywhere, and no leading underscore in global symbols.

Could do a single underscore suffix: arch_init_(), clean_call_opt_exit_(),
add_dynamo_vm_area_(), executable_areas_.  More of a minimal change.
It would mean "global but not in API".  Could combine it with a "dr_" prefix.

derekbruening:
[...]

Issue #3315, opened 2018-12-15
link error: duplicate symbol: memcpy


derekbruening:
The private version of memcpy and other routines in x86_shared.asm are linked into drfrontendlib which is statically linked into various "regular" applications.  Some users have reported duplicate symbol errors with some toolchains: details will be added here.  We can likely solve by adding `WEAK(memcpy)`, etc. into x86_shared.asm.

derekbruening:
@hgreving2304 -- I remember you hit this somewhere?  Could you supply details on what platform/toolchain it was?

hgreving2304:
@derekbruening , I think that was in Google3 when the -nostdlib compiler (driver, not linker) switch was missing.

derekbruening:
This is the plan:

Core DR's copies of memset and memcpy were placed into drhelper for
#1504, to avoid glibc-versioned symbols in drrun and other shipped
binaries.  They are linked into the static drfrontendlib and
drinjectlib libraries.  They are also in core DR when built as a
static library.  In these static library cases they can conflict with
symbols of the same name elsewhere in the linking application.

To solve this, we first separate memset and memcpy from drhelper into
their own library, drmemfuncs.  We simply avoid linking it into static
core DR, and into drdecode.  For static core DR we assume that
whatever versions the application or glibc provides will be
re-entrant.

We can't avoid linking it into drfrontendlib and drinjectlib (we hit
the glibc versioning problem), but we reduce the chance of conflicts
with these libraries by marking memset and memcpy as weak.

derekbruening:
There is a downside to having static DR use glibc memcpy.  It lets sanitizers or other interceptors override
DR's memcpy.  However, we already have interoperability concerns with sanitizers, so I was assuming we could live with that.

TACIXAT:
Running into this issue on Windows.

PR #4309, opened 2020-06-02
i#3315,i#3348: Move mem{cpy,set,move} into memfuncs on AArchXX

derekbruening:
Provides isolation of mem{cpy,set,move} in the DR static library on
AArchXX.  Moves the assembly mem{cpy,set} into a separate memfuncs.asm
for arm and aarch64, and links in memmove.c.  Marks the
__mem{cpy,set}_chk aliases as weak.  This brings AArchXX into parity
with x86 on library symbol isolation.

Removes the exceptions for AArchXX in the CMake_symbol_check test.
Confirms the build-time test fails without the code changes here:

  CMake Error at xxx/core/CMake_symbol_check.cmake:98 (message):
    *** Error:
    xxx/build_a64_dbg_tests/lib64/debug/libdynamorio_static_nohide.a
    contains a likely-to-conflict symbol: 4279: 00000000002cea3c 0 FUNC GLOBAL
    HIDDEN 248 memcpy

Issue: #3315, #3348

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -90,6 +90,7 @@ if (UNIX)
     "_memfuncs" "" "${asm_deps}")
   add_library(drmemfuncs STATIC ${memfuncs_asm_src} lib/memmove.c)
   add_gen_events_deps(drmemfuncs)
+  add_dependencies(drmemfuncs api_headers)
 endif ()
 
 # i#1409: to share core libc-ish code with non-core, we use the "drlibc" library.
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -1,5 +1,5 @@
 # **********************************************************
-# Copyright (c) 2010-2019 Google, Inc.    All rights reserved.
+# Copyright (c) 2010-2020 Google, Inc.    All rights reserved.
 # Copyright (c) 2009-2010 VMware, Inc.    All rights reserved.
 # **********************************************************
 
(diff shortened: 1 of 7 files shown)


core/CMakeLists.txt
```diff
@@ -80,13 +80,13 @@ endif ()
 add_asm_target(arch/pre_inject_asm.asm preinject_asm_src preinject_asm_tgt ""
   "-DNOT_DYNAMORIO_CORE_PROPER" "${asm_deps}")
 
-if (UNIX AND X86)
```
xyzsam, line 83: Just a reminder that there is an outstanding bug on the X86 macro definition causing conflicts in google3.

derekbruening:
> I think you're missing memfuncs.asm for arm in this commit - I only see deletions.

Wow, thank you -- never added locally.  Hmm, now I'm wondering how the AArch64 tests passed -- ok and now that I added them we have them failing :(.  Local AArch64 build was good.  Investigating.

derekbruening:
The A64 failure here is strange: I am theorizing a build race due a missing dependence on api_headers.  And I guess all the x86 builds just never hit it.

For A64 succeeding after I removed mem* and didn't add it back: there is a check that libdynamorio has zero libc imports that I would expect to fail with the compiler substituting libc versions.  Maybe it inlined everything?

derekbruening:
> For A64 succeeding after I removed mem* and didn't add it back: there is a check that libdynamorio has zero libc imports that I would expect to fail with the compiler substituting libc versions. Maybe it inlined everything?

Yeah I guess it would fail first due to not finding the referenced asm file :)  And looking at the output:

http://139.178.83.194:8080/job/DynamoRIO-AArch64-Precommit/1382/console

```
CMake Error: Cannot determine link language for target "drmemfuncs".
CMake Error in core/CMakeLists.txt:
  Exporting the target "drmemfuncs" is not allowed since its linker language
  cannot be determined
Command exited with the value: 1
Error(s) when configuring the project
```

So it did indeed fail.  But the wrapper script fails to detect the config-time failure as an overall error (too loose in ignoring cross-arch build attempts in multi-build setup?) and it showed up green.  Not good.  It seems familiar: maybe I'm thinking of Mac #4022?

derekbruening:
I filed #4312 on the misleading green on config failure.  And now AArch64 is green and I double-checked and it's a true green.  So I think we're all set here for this PR.
PTAL.

xyzsam:
Looks like appveyor is still failing?

derekbruening:
> Looks like appveyor is still failing?

Looks like a flake that has started becoming more common: drwrap-test-detach under #4058.  Should be added to the flake list since probably nobody will have time to try to fix it (if someone can reproduce it: most of these Appveyor flakes I can't repro on my recent-Win10 machine: Appveyor I believe is still on Win10-1607).  I clicked "re-run" on appveyor just to see how flaky this is.  Looks like @johnfxgalea hit it on a recent PR as well.

derekbruening:
I have PR #4313 marking it as a flake.

derekbruening:
And PR #4314 fixes the false green.

johnfxgalea:
> Looks like @johnfxgalea hit it on a recent PR as well.

Issue #3562, opened 2019-04-19
instruction tracing tool


rkgithubs:
I see that even though there are two sample program for instruction tracing (basic/IA specific); there is no instruction tracing tool. There are two tools for function tracing (system calls and library calls). I found these two sample code talking about instruction trace collection but there is 
http://dynamorio.org/docs/samples/instrace_simple.c
http://dynamorio.org/docs/samples/instrace_x86.c

derekbruening:
drmemtrace includes both the instruction stream and data stream and has a bunch of tooling around it with drcachesim.  It also supports just gathering the instruction stream with an online L1 cache filter.  Maybe the simplest thing would be to add an option for an unfiltered instruction-only trace there.

PR #4310, opened 2020-06-02
i3562: Add option for instruction-only trace collection to drcachesim.

abhinav92003:
Add option to drcachesim to include only instruction fetch entries and omit data entries from trace.

Issue: #3562

### DIFF
--- a/clients/drcachesim/common/options.cpp
+++ b/clients/drcachesim/common/options.cpp
@@ -177,6 +177,12 @@ droption_t<bytesize_t> op_L0D_size(
     "Must be a power of 2 and a multiple of -line_size, unless it is set to 0, "
     "which disables data entries from appearing in the trace.");
 
+droption_t<bool> op_inst_only_trace(
+    DROPTION_SCOPE_CLIENT, "inst_only_trace", false,
+    "Include only instruction fetch entries in trace",
+    "If -inst_only_trace, only instruction fetch entries are included in the "
+    "trace and data entries are omitted.");
+
 droption_t<bool> op_coherence(
     DROPTION_SCOPE_FRONTEND, "coherence", false, "Model coherence for private caches",
     "Writes to cache lines will invalidate other private caches that hold that line.");
--- a/clients/drcachesim/tracer/tracer.cpp
+++ b/clients/drcachesim/tracer/tracer.cpp
@@ -903,6 +903,9 @@ instrument_memref(void *drcontext, user_data_t *ud, instrlist_t *ilist, instr_t 
                   reg_id_t reg_ptr, int adjust, instr_t *app, opnd_t ref, int ref_index,
                   bool write, dr_pred_type_t pred)
 {
+    if (op_inst_only_trace.get_value()) {
+      return adjust;
+    }
     instr_t *skip = INSTR_CREATE_label(drcontext);
     reg_id_t reg_third = DR_REG_NULL;
     if (op_L0_filter.get_value()) {
--- a/suite/tests/CMakeLists.txt
+++ b/suite/tests/CMakeLists.txt
@@ -2874,6 +2874,7 @@ endif ()
       torunonly_drcachesim(filter-simple ${ci_shared_app} "-L0_filter" "")
       torunonly_drcachesim(filter-no-i ${ci_shared_app} "-L0_filter -L0I_size 0" "")
       torunonly_drcachesim(filter-no-d ${ci_shared_app} "-L0_filter -L0D_size 0" "")
+      torunonly_drcachesim(filter-inst-only ${ci_shared_app} "-inst_only_trace" "")
 
       torunonly_drcachesim(delay-simple ${ci_shared_app}
         "-trace_after_instrs 20000 -exit_after_tracing 10000" "")
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -148,6 +148,7 @@ Further non-compatibility-affecting changes include:
  - Added drmgr_register_opcode_instrumentation_event() and
    drmgr_unregister_opcode_instrumentation_event() so that drmgr supports
    opcode event instrumentation.
+ - Added -inst_only_trace option to drcachesim.
 
 **************************************************
 <hr>
--- a/clients/drcachesim/common/options.h
+++ b/clients/drcachesim/common/options.h
@@ -79,6 +79,7 @@ extern droption_t<std::string> op_LL_miss_file;
 extern droption_t<bytesize_t> op_L0I_size;
 extern droption_t<bool> op_L0_filter;
 extern droption_t<bytesize_t> op_L0D_size;
+extern droption_t<bool> op_inst_only_trace;
 extern droption_t<bool> op_coherence;
 extern droption_t<bool> op_use_physical;
 extern droption_t<unsigned int> op_virt2phys_freq;
--- a/clients/drcachesim/tests/filter-inst-only.templatex
+++ b/clients/drcachesim/tests/filter-inst-only.templatex
@@ -1,0 +1,23 @@
+Hello, world!
+---- <application exited with code 0> ----
+Cache simulation results:
+Core #0 \(1 thread\(s\)\)
+  L1I stats:
+    Hits:                         *[0-9,\.]*
+    Misses:                       *[0-9,\.]*
+    Invalidations:                *0
+.*   Miss rate:                   *[1-9][0-9][,\.]..%
+  L1D stats:
+    Hits:                                0
+    Misses:                              0
+    Invalidations:                       0
+Core #1 \(0 thread\(s\)\)
+Core #2 \(0 thread\(s\)\)
+Core #3 \(0 thread\(s\)\)
+LL stats:
+    Hits:                         *[0-9,\.]*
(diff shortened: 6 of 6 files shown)


AssadHashmi:
Added @abhinav92003 to AArch64 pre-commit ACL.

AssadHashmi:
run arm tests

clients/drcachesim/common/options.cpp
```diff
@@ -177,6 +177,12 @@ droption_t<bytesize_t> op_L0D_size(
     "Must be a power of 2 and a multiple of -line_size, unless it is set to 0, "
     "which disables data entries from appearing in the trace.");
 
+droption_t<bool> op_inst_only_trace(
```
derekbruening, line 180: Given that existing options involving instructions use `instr` to abbreviate it (`op_trace_after_instrs`, `op_online_instr_types`) I would vote for consistency here: so `op_instr_only_trace`.

suite/tests/CMakeLists.txt
```diff
@@ -2874,6 +2874,7 @@ endif ()
       torunonly_drcachesim(filter-simple ${ci_shared_app} "-L0_filter" "")
       torunonly_drcachesim(filter-no-i ${ci_shared_app} "-L0_filter -L0I_size 0" "")
       torunonly_drcachesim(filter-no-d ${ci_shared_app} "-L0_filter -L0D_size 0" "")
+      torunonly_drcachesim(filter-inst-only ${ci_shared_app} "-inst_only_trace" "")
```
derekbruening, line 2877: Please also add an offline-trace test to make sure post-processing of raw traces works when there are no data references (raw2trace may have some issues).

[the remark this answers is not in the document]
abhinav92003: Renamed it to instr-only-trace.

clients/drcachesim/common/options.cpp
abhinav92003: Done.

suite/tests/CMakeLists.txt
abhinav92003: It seems that when -offline is present, the trace somehow continues to have data references. Checking.

abhinav92003: Offline trace collection is fixed now. Also, added the suggested test.

derekbruening:
>￼ abhinav92003 force-pushed the i3562-inst-only-trace branch from 4f88bdf to 0722987

Please don't force-push to shared branches: it destroys the code history and review comment history, and makes reviewing difficult by not giving the reviewer the ability to see just what has changed since the last review.  We have warnings about this on multiple wiki pages: https://github.com/DynamoRIO/dynamorio/wiki/Contributing#contributing-to-dynamorio, https://github.com/DynamoRIO/dynamorio/wiki/Code-Reviews#responding-to-review-comments-and-submitting-code-updates, https://github.com/DynamoRIO/dynamorio/wiki/Code-Reviews#submitting-a-change-as-a-non-project-member.  Yet it still happens, although it should not be done for git shared branches in general and not just our project.  Is there anywhere else we can put warnings to help new contributors realize it should not be done?

abhinav92003:
> > ￼ abhinav92003 force-pushed the i3562-inst-only-trace branch from [4f88bdf](https://github.com/DynamoRIO/dynamorio/commit/4f88bdf841009f1d9d1c75f17e4932ae60b60209) to [0722987](https://github.com/DynamoRIO/dynamorio/commit/07229873332e622341eb2ae0a469af6a690f6170)
> 
> Please don't force-push to shared branches: it destroys the code history and review comment history, and makes reviewing difficult by not giving the reviewer the ability to see just what has changed since the last review. We have warnings about this on multiple wiki pages: https://github.com/DynamoRIO/dynamorio/wiki/Contributing#contributing-to-dynamorio, https://github.com/DynamoRIO/dynamorio/wiki/Code-Reviews#responding-to-review-comments-and-submitting-code-updates, https://github.com/DynamoRIO/dynamorio/wiki/Code-Reviews#submitting-a-change-as-a-non-project-member. Yet it still happens, although it should not be done for git shared branches in general and not just our project. Is there anywhere else we can put warnings to help new contributors realize it should not be done?

Sorry about that! I'll make sure to avoid this in future.

If it helps with the review: in this instance the force-push only changed the previous commit message, the code was unchanged,

derekbruening:
> If it helps with the review: in this instance the force-push only changed the previous commit message, the code was unchanged,

OK, though Github still does not show a button for viewing "Changes since last review", b/c the force breaks the history.

Just a heads up that I'm OOO today so may not finish a review until tomorrow.

abhinav92003:
> OK, though Github still does not show a button for viewing "Changes since last review", b/c the force breaks the history.
Thanks for explaining. I'll be more careful.

> Just a heads up that I'm OOO today so may not finish a review until tomorrow.
No problem. Thanks.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1480,7 +1483,8 @@ init_thread_in_process(void *drcontext)
         NOTIFY(2, "Created thread trace file %s\n", buf);
 
         /* Write initial headers at the top of the first buffer. */
-        offline_file_type_t file_type = op_L0_filter.get_value()
+        offline_file_type_t file_type =
+            op_L0_filter.get_value() || op_instr_only_trace.get_value()
             ? OFFLINE_FILE_TYPE_FILTERED
```
derekbruening, line 1488: I'm worried that conflating a filtered trace with a no-data trace will cause problems, as they are not the same.  A filtered trace inserts a separate instruction entry for every insruction, while a no-data trace just has the block tags and requires reconstruction of every single instruction in raw2trace.  Are we sure this ends up working properly in raw2trace?  Even if so, we may want a separate flag.

Actually I'm not really sure how a filtered no-data trace works in raw2trace: it looks like it will still go into append_memref()?  And hit the "Missing memref from predication, 0-iter repstr, or filter" case?  Is that what happens with this PR here as well (run `-verbose 4`)?

derekbruening: It would be more efficient to skip the memrefs altogether.  Add a new type attribute for no data and use it to skip the memrefs?  (And set it for filtered no-data too?)

abhinav92003: > it looks like it will still go into append_memref()? And hit the "Missing memref from predication, 0-iter repstr, or filter" case?

Yes to both. For each memory instruction src+dst we still go into append_memref(), but return without actually appending any memref.


> I'm worried that conflating a filtered trace with a no-data trace will cause problems, as they are not the same. A filtered trace inserts a separate instruction entry for every insruction, while a no-data trace just has the block tags and requires reconstruction of every single instruction in raw2trace. Are we sure this ends up working properly in raw2trace? Even if so, we may want a separate flag.

By setting the file type to OFFLINE_FILE_TYPE_FILTERED, we're disabling address elision [here](https://github.com/DynamoRIO/dynamorio/blob/8551f131d97490203c2fedfbef2da9314a3ea45a/clients/drcachesim/tracer/raw2trace.h#L773), which causes the "Missing memref..." path to trigger.

> It would be more efficient to skip the memrefs altogether. Add a new type attribute for no data and use it to skip the memrefs? (And set it for filtered no-data too?)

Do you mean by adding another field in [raw2trace_thread_data_t](https://github.com/DynamoRIO/dynamorio/blob/8551f131d97490203c2fedfbef2da9314a3ea45a/clients/drcachesim/tracer/raw2trace.h#L1402) to indicate absence of memrefs?

derekbruening: > > it looks like it will still go into append_memref()? And hit the "Missing memref from predication, 0-iter repstr, or filter" case?
> 
> Yes to both. For each memory instruction src+dst we still go into append_memref(), but return without actually appending any memref.
> 
> > I'm worried that conflating a filtered trace with a no-data trace will cause problems, as they are not the same. A filtered trace inserts a separate instruction entry for every insruction, while a no-data trace just has the block tags and requires reconstruction of every single instruction in raw2trace. Are we sure this ends up working properly in raw2trace? Even if so, we may want a separate flag.
> 
> By setting the file type to OFFLINE_FILE_TYPE_FILTERED, we're disabling address elision [here](https://github.com/DynamoRIO/dynamorio/blob/8551f131d97490203c2fedfbef2da9314a3ea45a/clients/drcachesim/tracer/raw2trace.h#L773), which causes the "Missing memref..." path to trigger.

Right, but the worry is that a future change to raw2trace will take an action on actual filtered traces which have a very different structure from this trace and break these.

> 
> > It would be more efficient to skip the memrefs altogether. Add a new type attribute for no data and use it to skip the memrefs? (And set it for filtered no-data too?)
> 
> Do you mean by adding another field in [raw2trace_thread_data_t](https://github.com/DynamoRIO/dynamorio/blob/8551f131d97490203c2fedfbef2da9314a3ea45a/clients/drcachesim/tracer/raw2trace.h#L1402) to indicate absence of memrefs?

No, I mean adding a new offline_file_type_t to more accurately describe this: a trace with no data refs.  And using it for filtered instr-only traces too.  Then raw2trace can make better decisions internally, and we have more accurate labels.

PR #4311, opened 2020-06-02
Add check for XSP in dr_restore_reg and dr_save_reg on AArch64

yury-khrustalev:
This check is helpful for AArch64 target which does not support loading into and storing from XSP.
This change allows to fail early with a meaningful error message: for a debug build it will look like this:

```
<Application /usr/bin/ls (68183) DynamoRIO usage error : dr_save_reg: store from XSP is not supported>
<Usage error: dr_save_reg: store from XSP is not supported (...dynamorio/core/lib/instrument.c, line 5807)
```

### DIFF
--- a/core/lib/instrument.c
+++ b/core/lib/instrument.c
@@ -5803,9 +5803,9 @@ dr_save_reg(void *drcontext, instrlist_t *ilist, instr_t *where, reg_id_t reg,
 
     CLIENT_ASSERT(reg_is_pointer_sized(reg), "dr_save_reg requires pointer-sized gpr");
 
-#ifdef AARCH64
+#    ifdef AARCH64
     CLIENT_ASSERT(reg != DR_REG_XSP, "dr_save_reg: store from XSP is not supported");
-#endif
+#    endif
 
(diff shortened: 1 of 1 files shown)


core/lib/instrument.c
```diff
@@ -5803,6 +5803,10 @@ dr_save_reg(void *drcontext, instrlist_t *ilist, instr_t *where, reg_id_t reg,
 
     CLIENT_ASSERT(reg_is_pointer_sized(reg), "dr_save_reg requires pointer-sized gpr");
 
+#    ifdef AARCH64
```
johnfxgalea, line 5806: I would test against X86, i.e, #ifndef X86. AFAIR, I think this check also applies for T32 instructions as well.

johnfxgalea: Double checked how drreg does it: https://github.com/DynamoRIO/dynamorio/blob/4964f9e063ef7daea574dfd1b3c2cd10d4b15592/ext/drreg/drreg.c#L827

johnfxgalea:
Thank you for the PR!

Issue #1684, opened 2015-04-14
add cross-arch support to drdecodelib


derekbruening:
Today we have ifdefs selecting whether the decoder and encoder support
either ARM or x86.  For a nice standalone decoder we should have support
for both in the same build.

derekbruening:
Alternatively, supporting building one or the other on either host architecture would also go a long way and be a nice mostly-there solution.

derekbruening:
For #4318 we'd want cross-arch support in full DR and not just drdecodelib.  I think it's not much extra work to do so as the bulk of the work would have to be done for drdecodelib anyway.

derekbruening:
The goal is to compile the decodelib sources for all architectures into
every DR library build: so e.g. the x86 build has code to generate, decode,
encode, and disassemble ARM and AArch64 instructions.

The main challenge is duplicate symbols:

+ Functions: various top-level routines like decode() and various per-arch
  routines like instr_is_call_arch().
  We can change these w/o any interface changes by having a new top-level
  decode() check the ISA mode.  Perhaps we can short-circuit that extra
  overhead for bb building decode+gen+encode.

+ Opcodes and INSTR_CREATE_ and OPND_ macros.
  Changing these to have arch prefixes or sthg is going to be a major
  compatibility break and we'd probably have to go to 9.0?

  So we'd have sthg like:
  `OP_add` => `OP_X86_add`, `OP_ARM_add`, `OP_AARCH64_add`?
  `INSTR_CREATE_add` => `INSTR_CREATE_X86_add`, `INSTR_CREATE_ARM_add`, `INSTR_CREATE_AARCH64_add`?

  We'd add the "_X86" to every single one and not just the ones whose
  names clash I assume.

  Would we add a `DR_` prefix while at it for better namespace separation?

  The symbol lengths start to get long is one downside.  Having some
  assembly-feel layer on top to generate code w/o the clunky and long macro
  and operand creation names (xref: err can't find it...thought we had it
  filed somewhere) would help.

  To shrink the long `INSTR_CREATE_AARCH64_add` and add `DR_`:
  `DR_CREATE_X86_add`, `DR_CREATE_A32_add`, `DR_CREATE_A64_add`?
  And `XINST_CREATE_add` becomes `DR_XCREATE_add`??

derekbruening:
@johnfxgalea , @AssadHashmi : looking for some feedback on the proposal in the prior comment re: interface changes and names.

[...]

PR #4321, opened 2020-06-12
i#1684 xarch-IR: Move IR files from core/arch to core/ir

derekbruening:
Moves all of the IR-related files (instruction generation, encoding,
decoding, disassembly, instructions, operands, instruction lists) from
core/arch to core/ir, mirroring the arch-specific subdirectories under
core/ir.  This is a code cleanup step toward properly isolating the
drdecode library, as well as moving us toward the ability to build for
a separate target architecture from the host architecture and
eventually perhaps building in multiple target architectures in the
same binary for decoding and IR manipulation.

Also renames mangle_utils.c to ir_utils.c to better fit its purpose
and location.

Issue: #1684


johnfxgalea:
I'm wondering whether there is also any documentation that needs to be updated?

johnfxgalea:
> I'm wondering whether there is also any documentation that needs to be updated?

Could be that there is none since these changes relate to internal code, but I'll do a quick grep just in case.

derekbruening:
> > I'm wondering whether there is also any documentation that needs to be updated?
> 
> Could be that there is none since these changes relate to internal code, but I'll do a quick grep just in case.

Thanks.  I did do some grepping and updating comments in addition to code but I did not do exhaustive searches, nor did I look in the wiki.

derekbruening:
The wiki just has `arch/` in ARM-Port.md, which contains historical design discussions and I'm inclined to leave it, and in JIT-Optimization.md which is a branch off an older version and so it should be left as well.

derekbruening:
I'm not seeing anything in `api/`.  I don't recall any docs in general talking about internal source code layout.

johnfxgalea:
When I did a grep, I found: 

```
./tools/x86opnums.pl:35:### enum in src/arch/*/opcode.h
./core/perscache.c:859:/* These decode-and-instr-using routines could go in arch/ as they assume that direct
./core/os_shared.h:1254:/* FIXME: if we add more opcodes maybe should be exported by arch/ */
./core/unix/os_exports.h:72: * Keep this consistent w/ the TLS_SEG_OPCODE define in arch/instr.h
```

derekbruening:
Done:
+ ./tools/x86opnums.pl
+ ./core/os_shared.h
+ ./core/unix/os_exports.h:
+ ./core/vmareas.c
+ ./core/ir/decode.h

Unchanged:
+ ./core/perscache.c: that's DR-managed-execution-oriented code so it should remain arch/

derekbruening:
Appveyor failures are burst_* which are known issues (unfortunately can't repro locally) and flaky: re-running to confirm, though last commit was just comments so almost no chance caused by this since 1st commit was green.

derekbruening:
PTAL

PR #4322, opened 2020-06-12
i#1684 xarch-IR: Remove mangle.c from drdecode

derekbruening:
Moves the handful of functions from mangle.c used by drdecode (things
that are difficult to untangle from the IR functions) into ir_utils.c
Renames the top-level ir_utils.c to ir_utils_shared.c.  Adds
ir_utils.h.

Issue: #1684

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -238,6 +238,8 @@ set(DECODER_SRCS
   ir/${ARCH_NAME}/encode.c
   ir/disassemble_shared.c
   ir/${ARCH_NAME}/disassemble.c
+  ir/ir_utils_shared.c
+  ir/${ARCH_NAME_SHARED}/ir_utils.c
   )
 if (X86)
   set(DECODER_SRCS ${DECODER_SRCS}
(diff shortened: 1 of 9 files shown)


johnfxgalea:
> (things that are difficult to untangle from the IR functions) 

Wait, maybe I misunderstood, but is there code that should not ideally be in ir_utils.c and refactoring is needed? If so, please add a comment.

johnfxgalea:
> > (things that are difficult to untangle from the IR functions)
> 
> Wait, maybe I misunderstood, but is there code that should not ideally be in ir_utils.c and refactoring is needed? If so, please add a comment.

Just to add, it seems odd that remangle_short_rewrite is moved away from mangle.c just based off their names. Include a comment or perhaps consider changing its name to better reflect why it is included in  ir_utils.c

derekbruening:
> > > (things that are difficult to untangle from the IR functions)
> > 
> > 
> > Wait, maybe I misunderstood, but is there code that should not ideally be in ir_utils.c and refactoring is needed? If so, please add a comment.
> 
> Just to add, it seems odd that remangle_short_rewrite is moved away from mangle.c just based off their names. Include a comment or perhaps consider changing its name to better reflect why it is included in ir_utils.c

Yeah, it's not ideal: but this is better than drdecode building *part* of mangle.c.  There are already a bunch of DR-execution-control things embedded in the IR code which this joins.  It will be a lot of work to properly untangle them: planning to live with it for now.

I added a comment to the two remangle_short_rewrite functions.

Issue #4318, opened 2020-06-09
Add cross-architecture drcachesim trace analysis and postprocessing support


derekbruening:
It is useful to be able to analyze and post-process drcachesim offline traces gathered on one architecture on a different platform: e.g., to trace an ARM application and pull the trace onto an x86 machine for analysis.

Most of the analysis code is already cross-platform, but the decoder for tools like opcode_mix is the key missing piece.  We would need something like #1684 to continue using DR's decoder there, but we'd need it in full DR and not just drdecodelib (because raw2trace uses full DR and it is complex to downgrade it).

derekbruening:
Pasting from PR #4325 :

Here's the key test in the new Travis job for drcachesim:
https://travis-ci.com/github/DynamoRIO/dynamorio/jobs/353366354
```
3109    Start 6: code_api|tool.drcacheoff.altbindir
3110
31116: Test command: /home/travis/build/DynamoRIO/dynamorio/build_aarch64-on-x86/bin64/runstats "-s" "90" "-killpg" "-silent" "-env" "LD_LIBRARY_PATH" "/home/travis/build/DynamoRIO/dynamorio/build_aarch64-on-x86/lib64/debug:/home/travis/build/DynamoRIO/dynamorio/build_aarch64-on-x86/ext/lib64/debug:" "-env" "DYNAMORIO_OPTIONS" "-stderr_mask 0xC -dumpcore_mask 0 -code_api" "/home/travis/build/DynamoRIO/dynamorio/build_aarch64-on-x86/clients/bin64/drcachesim" "-indir" "/home/travis/build/DynamoRIO/dynamorio/build_aarch64-on-x86/drmemtrace.threadsig.aarch64" "-simulator_type" "opcode_mix" "-alt_module_dir" "/home/travis/build/DynamoRIO/dynamorio/clients/drcachesim/tests/drmemtrace.threadsig.aarch64.raw" "-module_file" "/home/travis/build/DynamoRIO/dynamorio/build_aarch64-on-x86/drmemtrace.threadsig.aarch64/raw/modules.log"
31165/6 Test #5: code_api|api.drdecode ................   Passed    0.00 sec
31176: Opcode mix tool results:
31186:          125442 : total executed instructions
31196:           15742 :     bcond
31206:           15165 :       stp
31216:           13670 :      subs
31226:           12240 :       add
31236:            8286 :       and
31246:            6059 :        xx
...
```

[...]

PR #4330, opened 2020-06-22
i#4318 xarch memtrace: Add library search dir

derekbruening:
Adds a new option -alt_module_dir which specifies an alternate
directory in which to look for libraries in a modules.log offline
drcachesim file.  Hooks the option up to the opcode_mix and view tools
and to raw2trace (both its standalone launcher and drcachesim frontend
integration).

This will be useful in general, but is needed for adding tests with
checked-in raw files and binaries, where the checked-in modules.log
has a hardcoded path that will never match all test environments.

Adds a test that runs both raw2trace and opcode_mix on a new
checked-in aarch64 raw trace (this test will evolve for future
cross-arch testing, which is the reason for using aarch64).  The
binaries are stripped, with the two libraries "fake" binaries:
libmemtrace.so is not executed, and libdynamorio.so is only used for
dynamorio_sigreturn, so these are synthetic binaries (created from asm
files) to keep them smaller.  The tests deliberately fail on x86 and
match an error message that will improve when architecture tags are
added.

Issue: #4318


derekbruening:
The modules.log is the wrong format for Windows (need extra columns for checksum, etc. there).  I'm just going to make the test UNIX-only.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -439,6 +441,7 @@ class module_mapper_t final {
     byte *last_map_base_ = nullptr;
 
     uint verbosity_ = 0;
+    std::string alt_dir_;
```
snehasish, line 444: s/alt_dir_/alt_module_dir_ to be consistent with the rest of the changes.

clients/drcachesim/common/options.cpp
```diff
@@ -94,6 +94,12 @@ droption_t<std::string> op_module_file(
     "If the file is named modules.log and is in the same directory as the trace file, "
     "or a raw/ subdirectory below the trace file, this parameter can be omitted.");
 
+droption_t<std::string> op_alt_mod_dir(
```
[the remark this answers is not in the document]
derekbruening: SGTM.  Done.

clients/drcachesim/tracer/raw2trace.h
derekbruening: Done.

Issue #1345, opened 2014-11-28
mixed-mode *nix support: 64-bit + 32-bit on *nix


derekbruening:
_From [derek.br...@gmail.com](https://code.google.com/u/117968039472581148324/) on January 02, 2014 01:08:26_

Xref issue #49 , issue #828 , issue #865 , etc. for Windows mixed-mode and for mixed-mode
client issues.

On both Mac and Linux, mixing libraries not officially supported.  Also on
both, there's no WOW-like user-mode layer: 32-bit just talks directly to
the kernel.  We have yet to see any app that mixes modes.  For now we do
not plan to support mixed-mode except on Windows.  We certainly don't plan
to support mixing 32-bit and 64-bit libraries: we are currently hardcoding
the ELF and Mach-O struct definitions for the current arch.

_Original issue: http://code.google.com/p/dynamorio/issues/detail?id=1345_

derekbruening:
This is blocking cross-architecture support for drdecode and drcachesim, both for a single build with host!=target of a different bitwidth, and the future goal a single multi-arch build: #1684, #4318.

PR #4331, opened 2020-06-23
i#4318 xarch memtrace: Add arch tags to memtraces

derekbruening:
Adds architecture tags to raw traces, propagated to final traces in a
new marker type.  Adds reader support for the new type.

Adds a check for unhandled architectures to raw2trace.

Adds checks for unhandled architectures to the opcode_mix and view
tools.  To reach that point, relaxes the ELF header checks to allow
libraries of the same bitwidth but different architectures (i#1345
prevents allowing all libraries), so we can map in the other-arch
libraries.

Updates the existing aarch64 raw trace files to contain architecture
tags, and updates the x86 failure message.

Cross-architecture handling of these same files was tested manually
with a prototype which will be committed in the future.

Issue: #4318, #1345


derekbruening:
I haven't done a chain of PR's on Github before: interesting that Travis is not being run here.  Because it's not a pull to master?  I thought our config asked for all PR's, and the limit to master was just pushes, not PR comparisons.  In any case, I planned to change the target branch to master once the base PR #4330 is in, so Travis will run at that point.  The idea is to get parallel reviews.  Local tests pass.

abhinav92003:
Just wanted to let you know that I'm OOO today. I might not be able to get to this until tomorrow.

derekbruening:
> Just wanted to let you know that I'm OOO today (Tuesday) also. I might not be able to get to this until tomorrow.

NP, will switch reviewers.

clients/drcachesim/tools/view.cpp
```diff
@@ -109,6 +109,19 @@ view_t::process_memref(const memref_t &memref)
...
     if (memref.marker.type == TRACE_TYPE_MARKER) {
         switch (memref.marker.marker_type) {
+        case TRACE_MARKER_TYPE_FILETYPE:
+            if (TESTANY(OFFLINE_FILE_TYPE_ARCH_ALL, memref.marker.marker_value) &&
+                !TESTANY(IF_X86_ELSE(IF_X64_ELSE(OFFLINE_FILE_TYPE_ARCH_X86_64,
+                                                 OFFLINE_FILE_TYPE_ARCH_X86_32),
+                                     IF_X64_ELSE(OFFLINE_FILE_TYPE_ARCH_AARCH64,
+                                                 OFFLINE_FILE_TYPE_ARCH_ARM32)),
+                         memref.marker.marker_value)) {
+                error_string_ = std::string("Architecture mismatch: trace recorded on ") +
```
xyzsam, line 119: It's worth clarifying that this error string is about bitwidth mismatch by adding what the expected architecture/bitwidth should be. It took me awhile to parse those macros.

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -1077,15 +1079,27 @@ trace_metadata_reader_t::is_thread_start(const offline_entry_t *entry,
...
         return false;
     }
+    if (TESTANY(OFFLINE_FILE_TYPE_ARCH_ALL, type) &&
+        !TESTANY(IF_X86_ELSE(IF_X64_ELSE(OFFLINE_FILE_TYPE_ARCH_X86_64,
+                                         OFFLINE_FILE_TYPE_ARCH_X86_32),
+                             IF_X64_ELSE(OFFLINE_FILE_TYPE_ARCH_AARCH64,
+                                         OFFLINE_FILE_TYPE_ARCH_ARM32)),
+                 type)) {
+        std::stringstream ss;
+        ss << "Architecture mismatch: trace recorded on " << trace_arch_string(type);
```
xyzsam, line 1099: Same comment about error string here.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1496,6 +1496,12 @@ init_thread_in_process(void *drcontext)
             file_type = static_cast<offline_file_type_t>(
                 file_type | OFFLINE_FILE_TYPE_INSTRUCTION_ONLY);
         }
+        file_type = static_cast<offline_file_type_t>(
+            file_type |
+            IF_X86_ELSE(
```
xyzsam, line 1501: This macro has appeared several times now, is it worth sticking into a header?

clients/drcachesim/tools/view.cpp
derekbruening: OK I added the architecture the tool was compiled for to the output.

clients/drcachesim/tracer/raw2trace.cpp
derekbruening: Done.  And for opcode_mix.

clients/drcachesim/tracer/tracer.cpp
derekbruening: Done.

derekbruening:
Hmm, the A64 failure is odd: it has close to double the counts it should.  Is Jenkins not using a fresh build directory and we have the old .raw files too?  I thought the cmake scripts cleared the dirs out though.  I can't repro the failure on a local A64 machine: passes there.  I'll try an explicit file clear.

PR #4333, opened 2020-06-23
Fix NULL pointer dereference.

slackito:
Although this is a "benign" dereference (it is only used to compute the address of a member of a struct and no actual memory access is attempted), this is still undefined behavior and it causes problems when using tools like UndefinedBehaviorSanitizer.


derekbruening:
run arm tests

derekbruening:
@AssadHashmi -- "run arm tests" doesn't seem to be triggering Jenkins here

derekbruening:
> @AssadHashmi -- "run arm tests" doesn't seem to be triggering Jenkins here

Scratch that: it did run!  It took a while to update here is all: even a reload didn't show it running at all where normally it shows the status immediately.

PR #4340, opened 2020-06-27
Fixed bug where multiple drwrap_skip_call calls would fail

missimer:
wrap_level was not being decremented when a function call is skipped, this
eventually lead to the MAX_WRAP_NESTING limit being hit

### DIFF
--- a/ext/drwrap/drwrap.c
+++ b/ext/drwrap/drwrap.c
@@ -1994,6 +1994,7 @@ drwrap_in_callee(void *arg1, reg_t xsp _IF_NOT_X86(reg_t lr))
     if (pt->skip[pt->wrap_level]) {
         /* drwrap_skip_call already adjusted the stack and pc */
         /* ensure we have DR_MC_ALL */
+        pt->wrap_level--;
         dr_redirect_execution(drwrap_get_mcontext_internal((void *)&wrapcxt, DR_MC_ALL));
         ASSERT(false, "dr_redirect_execution should not return");
     }


missimer:
First time opening a PR to this project, hit this bug while using DynamoRIO to skip a large number of function calls.  Seems to address: https://github.com/DynamoRIO/dynamorio/issues/2298

If there is anything I can add to the PR please let me know.

derekbruening:
> First time opening a PR to this project, hit this bug while using DynamoRIO to skip a large number of function calls. Seems to address: #2298

Thank you for the patch!

> If there is anything I can add to the PR please let me know.

The ideal thing would be to have a test that fails without the fix and passes with it.  The main drwrap tests are in suite/tests/client-interface/drwrap-test.dll.c and its app code: maybe new wrap targets with lots of skips could be added?  Or a new separate test if that's easier.

Also, it seems that drwrap-test is failing on Travis and Appveyor: so there may be something not quite right with the fix?  Can you reproduce the failure locally?  The CI links have the test output if not.

missimer:
I did find the related tests in drwrap-test.dll.c and I have an idea on how to expose the bug via the test.  I'm having difficulty running the tests locally but I haven't had time to investigate a lot, currently have a deadline and I've worked around the bug by using the replace interface of drwrap.

Apologize for asking here but is there a simple way to run just that one failing test, that why when I get the tests working locally I can save my self some time while I try to find the cause of the failure.

I will probable be able to look into this in a 2/3 weeks.  If you prefer I can close the PR until it passes on the CI.

derekbruening:
> I did find the related tests in drwrap-test.dll.c and I have an idea on how to expose the bug via the test. I'm having difficulty running the tests locally but I haven't had time to investigate a lot, currently have a deadline and I've worked around the bug by using the replace interface of drwrap.
> 
> Apologize for asking here but is there a simple way to run just that one failing test, that why when I get the tests working locally I can save my self some time while I try to find the cause of the failure.

See https://github.com/DynamoRIO/dynamorio/wiki/Test-Suite#building-and-running-tests or 
https://cmake.org/cmake/help/latest/manual/ctest.1.html
```
ctest -V -R drwrap-test\$
```

> 
> I will probable be able to look into this in a 2/3 weeks. If you prefer I can close the PR until it passes on the CI.

No need to close; just put in a PTAL comment when you get time to update it.

Issue #4328, opened 2020-06-22
AArch64 prefetch and flush instructions are not marked by drmemtrace


derekbruening:
drcachesim's tracer is missing AArch64 checks for opcodes in instru_t::instr_to_prefetch_type and instru_t::instr_is_flush.

abhinav92003:
I noticed something odd in how the x86 `clflush` instruction is handled by the tracer. 

It is handled by `instrument_memref`, same as other instructions with memory references. Therefore, the L0 filter is also applied to it. This means that it may be filtered out of the trace. I verified this by adding multiple `clflush` in a pure asm app. 
https://github.com/DynamoRIO/dynamorio/blob/cdc24cb13514cca50012dc9397b446323ddd62c5/clients/drcachesim/tracer/tracer.cpp#L902



Without `-L0_filter` all of them show up:
```
rm -r drmemtrace.*;  ./bin64/drrun -t drcachesim  -offline a.out; ./bin64/drrun -t drcachesim -indir drmemtrace* -simulator_type view
<marker: timestamp 13239920657418705>
<marker: tid 375365 on core 15>
  0x0000000000401000  0f ae 3c 25 34 10 40 clflush 0x00401034
                      00
  0x0000000000401008  0f ae 3c 25 34 10 40 clflush 0x00401034
                      00
  0x0000000000401010  0f ae 3c 25 34 10 40 clflush 0x00401034
                      00
  0x0000000000401018  0f ae 3c 25 34 10 40 clflush 0x00401034
                      00
  0x0000000000401020  0f ae 3c 25 34 10 40 clflush 0x00401034
                      00
  0x0000000000401028  48 c7 c0 3c 00 00 00 mov    $0x0000003c, %rax
  0x000000000040102f  48 31 ff             xor    %rdi, %rdi
  0x0000000000401032  0f 05                syscall
<marker: timestamp 13239920657418786>
<marker: tid 375365 on core 15>
View tool results:
              8 : total disassembled instructions
```

With `-L0_filter` only one shows up:
```
rm -r drmemtrace.*;  ./bin64/drrun -t drcachesim -L0_filter -L0D_size 0      -offline a.out; ./bin64/drrun -t drcachesim -indir drmemtrace* -simulator_type view
<marker: timestamp 13239920381948985>
<marker: tid 374727 on core 15>
  0x0000000000401000  0f ae 3c 25 34 10 40 clflush 0x00401034
                      00
[...]

Issue #4296, opened 2020-05-13
HANG drcachesim online across setarch execve


derekbruening:
This surprised me b/c I thought we had multi-process tests including execve, so there is likely something particular about this specific execve?

I just wanted to disable ASLR for the app and it hangs:

```
$ bin64/drrun -t drcachesim -- setarch `uname -m` -R suite/tests/bin/simple_app
<Starting application /usr/bin/setarch (6420)>
<Initial options = -no_dynamic_options -client_lib '/home/bruening/dr/git/build_x64_dbg_tests/bin64/../clients/lib64/debug/libdrmemtrace.so;0;' -code_api -stack_size 56K -signal_stack_size 32K -max_elide_jmp 0 -max_elide_call 0 -early_inject -emulate_brk -no_inline_ignored_syscalls -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct >
<Paste into GDB to debug DynamoRIO clients:
set confirm off
add-symbol-file ...>
<curiosity: rex.w on OPSZ_6_irex10_short4!>
<get_memory_info mismatch! (can happen if os combines entries in /proc/pid/maps)
	os says: 0x00007f8597254000-0x00007f8597277000 prot=0x00000001
	cache says: 0x00007f8597254000-0x00007f859730d000 prot=0x00000001
>
<(1+x) Handling our fault in a TRY at 0x00007f8597a8d93b>
<-- execve suite/tests/bin/simple_app -->
<Starting application /home/bruening/dr/git/build_x64_dbg_tests/suite/tests/bin/simple_app (6420)>
<Initial options = -client_lib '/home/bruening/dr/git/build_x64_dbg_tests/bin64/../clients/lib64/debug/libdrmemtrace.so;0;' -code_api -stack_size 56K -signal_stack_size 32K -max_elide_jmp 0 -max_elide_call 0 -early_inject -emulate_brk -no_inline_ignored_syscalls -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct >
<Paste into GDB to debug DynamoRIO clients:
set confirm off
add-symbol-file ..>
<curiosity: rex.w on OPSZ_6_irex10_short4!>
<get_memory_info mismatch! (can happen if os combines entries in /proc/pid/maps)
	os says: 0x00007ffff75df000-0x00007ffff7602000 prot=0x00000001
	cache says: 0x00007ffff75df000-0x00007ffff7698000 prot=0x00000001
>
[...]

Issue #2007, opened 2016-09-20
AArch64 build failures with static DR


derekbruening:
#975 made static DR first-class-supported and it builds on its own and in small startstop test apps, but linking with drmemtrace it fails to link:

https://travis-ci.org/DynamoRIO/dynamorio/jobs/161201683

```
Linking CXX executable ../bin64/tool.drcacheoff.burst_static
../../lib64/debug/libdynamorio_static.a(libdynamorio_static.o): In function `check_param_bounds':
/home/travis/build/DynamoRIO/dynamorio/core/options.c:561:(.text+0x108c4): .relocation truncated to fit: R_AARCH64_LD64_GOT_LO12_NC against `main_logfile'
/home/travis/build/DynamoRIO/dynamorio/core/options.c:561:(.text+0x10934): relocation truncated to fit: R_AARCH64_LD64_GOT_LO12_NC against `main_logfile'
../../lib64/debug/libdynamorio_static.a(libdynamorio_static.o): In function `update_dynamic_options':
/home/travis/build/DynamoRIO/dynamorio/core/optionsx.h:221:(.text+0x339f8): relocation truncated to fit: R_AARCH64_LD64_GOT_LO12_NC against `main_logfile'
/home/travis/build/DynamoRIO/dynamorio/core/optionsx.h:227:(.text+0x33b28): relocation truncated to fit: R_AARCH64_LD64_GOT_LO12_NC against `main_logfile'
/home/travis/build/DynamoRIO/dynamorio/core/optionsx.h:228:(.text+0x33c58): relocation truncated to fit: R_AARCH64_LD64_GOT_LO12_NC against `main_logfile'
/home/travis/build/DynamoRIO/dynamorio/core/optionsx.h:229:(.text+0x33d88): relocation truncated to fit: R_AARCH64_LD64_GOT_LO12_NC against `main_logfile'
/home/travis/build/DynamoRIO/dynamorio/core/optionsx.h:230:(.text+0x3!3eb8): relocation truncated to fit: R_AARCH64_LD64_GOT_LO12_NC against `main_logfile'
/home/travis/build/DynamoRIO/dynamorio/core/optionsx.h:231:(.text+0x33fe8): relocation truncated to fit: R_AARCH64_LD64_GOT_LO12_NC against `main_logfile'
[...]

PR #4341, opened 2020-06-29
i#4328 AArch64 prefetch and flush instrs: Add trace types for prefetch instructions

abhinav92003:
Adds new trace types for AArch64 prfm and prfum instructions. These prefetch instructions have multiple variants based on:
- prefetch target: this may be the L1, L2 or L3 cache
- prefetch policy: either temporal or non-temporal
- type: prefetch may be for load, instruction or store

Also makes changes to account for these new prefetch trace types in drcachesim.

Issue: #4328


derekbruening:
> Install missing cross compiler for DYNAMORIO_A64_ON_X86_ONLY.

Please revert this part -- the whole point of the a64-on-x86 is to analyze aarch64 traces *on x86*.  There should be *no cross-compiler*.  See #4318, #1684, PR #4325.

clients/drcachesim/tests/prefetch-aarch64-basic-counts.templatex
```diff
@@ -0,0 +1,32 @@
...
+     .* total function id markers
+     .* total function return address markers
+     .* total function argument markers
+     .* total function return value markers
+     .* total other markers
+Thread .* counts:
+     .* \(fetched\) instructions
+     .* unique \(fetched\) instructions
+     .* non-fetched instructions
+          62 prefetches
```
abhinav92003, line 23: Some 26 prefetches occur independently of the prefetches issued explicitly in the test. 

This makes the total expected prefetches in prefetch-aarch64-basic-counts = 36+26 = 62.

Also, in the other test (prefetch-basic-counts), this makes the total expected prefetches = 8+26 = 34

abhinav92003:
Travis is failing for macOS: https://travis-ci.com/github/DynamoRIO/dynamorio/jobs/355707609. I couldn't find any useful information to debug this in the log, except that it exited with code 2. @derekbruening any idea what may be the cause?

derekbruening:
> Travis is failing for macOS: https://travis-ci.com/github/DynamoRIO/dynamorio/jobs/355707609. I couldn't find any useful information to debug this in the log, except that it exited with code 2. @derekbruening any idea what may be the cause?

See #4223.  Somehow when the Mac build/tests fail the output is truncated.  Maybe a Travis bug.  Do you have a Mac you could repro on?

suite/tests/client-interface/prefetch_aarch64_app.c
```diff
@@ -0,0 +1,112 @@
+/* **********************************************************
```
derekbruening, line 1: Normally test apps dedicated to drcachesim tests live in clients/drcachesim/tests/ (e.g., clients/drcachesim/tests/stride_benchmark.cpp).

suite/tests/client-interface/prefetch_app.c
```diff
@@ -0,0 +1,60 @@
...
+ * DAMAGE.
+ */
+
+#include <stdio.h>
+int
+main(void)
+{
+    int d = 1;
+
+    // Prefetch for read.
```
derekbruening, line 39: Do we need two separate test apps?  What if we put all the aarch64 asm inside this one and simplified the suite.

suite/tests/client-interface/prefetch_aarch64_app.c
```diff
@@ -0,0 +1,112 @@
...
+}
+
+#if defined(__aarch64__)
+void
+prfm(void)
+{
+    int d = 1;
+
+    // prfm for read.
+    __asm__ __volatile__("prfm\tpldl1keep,[%0]" : : "r"(d));
```
derekbruening, line 47: This tab seems a little unorthodox?  I would think space(s) would be more normal and more readable and more consistent with all of DR's other asm.

clients/drcachesim/tests/prefetch-aarch64-basic-counts.templatex
derekbruening: > Some 26 prefetches occur independently of the prefetches issued explicitly in the test.

That is going to make this a flaky test as it is likely quite dependent on the particular compiler and/or system libraries.

There are 3 potential solutions:
1) Make the test app pure asm so there is no compiler-added or system library code.  This means it won't run on Windows -- but we can live with that.
2) Use a checked-in offline trace (with no online test).
3) Make the regex handle variations.

abhinav92003: I used a combination of (2) and (3). I modified the existing regex to verify only that prefetch instruction count is non-zero. And I also checked-in an offline trace collected on AArch64 that verifies the exact values.

About (1), I tried using a very simple asm code, but even that contains some 14 prefetches. Were you suggesting a different way?
```
abhinavas@tx1:~/dr/build/i4328-1$ cat a.s
.global main
main:
abhinavas@tx1:~/dr/build/i4328-1$ gcc -O0 a.s
abhinavas@tx1:~/dr/build/i4328-1$ ./a.out
abhinavas@tx1:~/dr/build/i4328-1$ rm -r drmemtrace.*;  ./bin64/drrun -t drcachesim  -offline ./a.out; ./bin64/drrun -t drcachesim -indir drmemtrace* -simulator_type basic_counts
Basic counts tool results:
Total counts:
       69762 total (fetched) instructions
        6531 total unique (fetched) instructions
           0 total non-fetched instructions
          14 total prefetches
       17886 total data loads
        7626 total data stores
           1 total threads
          56 total scheduling markers
           0 total transfer markers
           0 total function id markers
           0 total function return address markers
           0 total function argument markers
           0 total function return value markers
           1 total other markers
Thread 11801 counts:
       69762 (fetched) instructions
        6531 unique (fetched) instructions
           0 non-fetched instructions
          14 prefetches
       17886 data loads
        7626 data stores
          56 scheduling markers
           0 transfer markers
           0 function id markers
           0 function return address markers
           0 function argument markers
           0 function return value markers
           1 other markers
abhinavas@tx1:~/dr/build/i4328-1$ 
```

clients/drcachesim/common/trace_entry.h
```diff
@@ -163,6 +163,49 @@ typedef enum {
      */
     TRACE_TYPE_INSTR_SYSENTER,
 
+    /** AArch64 specific prefetch trace entries.
+     *  Relative order of these enum values is relevant to the handling of AArch64
+     * prefetch instructions.
+     */
+    TRACE_TYPE_PREFETCH_PLDL1KEEP, /**< An AArch64 load prefetch to level 1 of the cache.
```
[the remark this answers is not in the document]
abhinav92003: I renamed the enum strings to be architecture agnostic. 

> If this is the same as TRACE_TYPE_PREFETCH_READ let's just use that.

AArch64 prefetches have more variations than x86 or ARM -- {read, instr, write} x {L1, L2, L3} x {temporal, non-temporal}. So we'll need to add new enum values.

> I see that some of the existing prefetches do mention ARM or x86, but IMHO we should try to remove that too.

Do you mean we should remove the existing enum values and reuse these new ones for all architectures?

clients/drcachesim/common/trace_entry.h
```diff
@@ -163,6 +163,49 @@ typedef enum {
...
+                                    */
+    TRACE_TYPE_PREFETCH_PLDL1STRM, /**< An AArch64 streaming load prefetch to level 1 of
+                                      the cache. */
+    TRACE_TYPE_PREFETCH_PLDL2KEEP, /**< An AArch64 load prefetch to level 2 of the cache.
```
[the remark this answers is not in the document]
abhinav92003: Shortened the comments to fit in a single line.

clients/drcachesim/common/trace_entry.h
```diff
@@ -163,6 +163,49 @@ typedef enum {
      */
     TRACE_TYPE_INSTR_SYSENTER,
 
+    /** AArch64 specific prefetch trace entries.
```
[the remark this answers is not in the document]
abhinav92003: Checked the HTML output -- Doxygen does append them. Replaced them with single line comments.

clients/drcachesim/tracer/instru.cpp
```diff
@@ -78,6 +78,51 @@ instru_t::instr_to_instr_type(instr_t *instr, bool repstr_expanded)
...
+    unsigned short op_type = get_aarch64_prefetch_op_type(prfop);
+    DR_ASSERT_MSG(target_policy < 6 && op_type < 3,
+                  "Unsupported AArch64 prefetch operation.");
+    switch (op_type) {
+    case 0: // prefetch for load
+        return TRACE_TYPE_PREFETCH_PLDL1KEEP + target_policy;
+    case 1: // prefetch for instruction
+        return TRACE_TYPE_PREFETCH_PLIL1KEEP + target_policy;
+    case 2: // prefetch for store
+        return TRACE_TYPE_PREFETCH_PSTL1KEEP + target_policy;
```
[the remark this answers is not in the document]
abhinav92003: Added the comment.

suite/tests/CMakeLists.txt
```diff
@@ -652,6 +652,20 @@ set(ci_shared_app simple_app)
 set(ci_shared_app_src client-interface/${ci_shared_app}.c)
 add_exe(${ci_shared_app} ${ci_shared_app_src})
 
+if (CMAKE_COMPILER_IS_GNUCC)
+  # An app with prefetch instructions.
```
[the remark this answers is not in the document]
abhinav92003: Moved it to later where the test is setup.

suite/tests/client-interface/prefetch_aarch64_app.c
abhinav92003: Moved the apps to clients/drcachesim/tests

suite/tests/client-interface/prefetch_aarch64_app.c
abhinav92003: Replaced it with space.

clients/drcachesim/tracer/instru.cpp
```diff
@@ -78,6 +78,51 @@ instru_t::instr_to_instr_type(instr_t *instr, bool repstr_expanded)
     return TRACE_TYPE_INSTR;
 }
 
+#ifdef AARCH64
+unsigned short
```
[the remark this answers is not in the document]
abhinav92003: Made them static members of instru_t.

suite/tests/CMakeLists.txt
```diff
@@ -3108,6 +3122,15 @@ if (CLIENT_INTERFACE)
 
       torunonly_drcachesim(instr-only-trace ${ci_shared_app} "-instr_only_trace" "")
 
+      if (CMAKE_COMPILER_IS_GNUCC)
```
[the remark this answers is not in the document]
abhinav92003: The test uses __builtin_prefetch() which is available only in gcc

suite/tests/CMakeLists.txt
```diff
@@ -652,6 +652,20 @@ set(ci_shared_app simple_app)
 set(ci_shared_app_src client-interface/${ci_shared_app}.c)
 add_exe(${ci_shared_app} ${ci_shared_app_src})
 
+if (CMAKE_COMPILER_IS_GNUCC)
```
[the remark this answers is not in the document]
abhinav92003: Added comment to explain the restriction.

clients/drcachesim/common/trace_entry.h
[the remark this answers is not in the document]
abhinav92003: Changed this to a set of single line comments.

suite/tests/client-interface/prefetch_app.c
abhinav92003: I thought it better to separate them as they have different domains -- one contains AArch64-specific assembly and the other a GCC-specific function (__builtin_prefetch).

clients/drcachesim/common/trace_entry.h
```diff
@@ -163,6 +163,30 @@ typedef enum {
      */
     TRACE_TYPE_INSTR_SYSENTER,
 
+    // Trace entry types for prefetch instructions.
+    // Relative order of these enum values is relevant to the handling of prefetch
+    // instructions.
+    TRACE_TYPE_PREFETCH_READ_L1,    /**< Load prefetch to L1 cache. */
```
abhinav92003, line 169: @derekbruening Do you prefer to qualify this identifier with a _TE suffix denoting 'temporal prefetch'?

clients/drcachesim/tests/prefetch-aarch64-basic-counts.templatex
derekbruening: > About (1), I tried using a very simple asm code, but even that contains some 14 prefetches. Were you suggesting a different way?

Build with absolutely no startup code so the *only* code is the asm code and there will be nothing else but the precise instructions you place there.  See e.g. suite/tests/common/allasm_aarch64_cache.asm and related files which are built `-nostartfiles -nodefaultlibs -static`.

clients/drcachesim/common/trace_entry.h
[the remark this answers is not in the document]
derekbruening: > > If this is the same as TRACE_TYPE_PREFETCH_READ let's just use that.
> 
> AArch64 prefetches have more variations than x86 or ARM -- {read, instr, write} x {L1, L2, L3} x {temporal, non-temporal}. So we'll need to add new enum values.

Right, but only new ones for the ones that don't already exist.
 
> > I see that some of the existing prefetches do mention ARM or x86, but IMHO we should try to remove that too.
> 
> Do you mean we should remove the existing enum values and reuse these new ones for all architectures?

Not remove, but change the doc description and add an arch-agnostic alias enum value.  Could be a separate PR (but this would be chained onto it if the aarch64 types overlap).

suite/tests/CMakeLists.txt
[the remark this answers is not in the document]
derekbruening: __builtin_prefetch is not just in gcc: it is also in clang.

clients/drcachesim/common/trace_entry.h
derekbruening: I would vote for only labeling the non-temporal.

abhinav92003: Ack. I'll let it be as it is.

clients/drcachesim/tests/prefetch-aarch64-basic-counts.templatex
abhinav92003: Thanks. I was able to build some asm code without any of those 'default' prefetches.

In this PR, do you suggest we switch to a pure asm test app suggested in (1), instead of the tests added currently?

derekbruening: I think a pure asm test is superior to a regex checking for non-zero.  You could also have a custom-analysis test that checks for the different types of prefetches.  Another method of controlling the precise count is to use attach+detach around the traced code.

abhinav92003: attach+detach seems interesting, but based on the latest update on #1551, I think start/stop support is not implemented yet for ARM. I'll add a pure asm test instead.

About the custom analysis test, this will be only for offline trace right? To use it for online, we'll need to add it as another simulator_type in drcachesim. I'm assuming we don't want to do that only for this test.

clients/drcachesim/common/trace_entry.h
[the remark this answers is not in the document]
abhinav92003: I'm not sure if we can share the same set of enums across all architectures.

> Right, but only new ones for the ones that don't already exist.

The existing enums describe prefetches with a limited set of properties (only the target cache level or operation type). For AArch64, we need enums that describe all three properties: operation type, target cache level, locality (streaming/temporal). If I understand correctly, we cannot use any of the existing enums for AArch64.

Maybe we can use some of the new enums for x86/ARM. ARM's `TRACE_TYPE_PREFETCH_READ` PLD instruction can perhaps be deprecated in favour of `TRACE_TYPE_PREFETCH_READ_L1`. But for some others, the mapping is not straight-forward - like `TRACE_TYPE_PREFETCHT2` doesn't specify whether there's intent to write or not.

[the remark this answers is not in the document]
derekbruening: Agreed, if an existing one does not line up we would keep it separate and make a new one.  But if they are identical semantics, IMHO sharing is the best thing, as it will simplify trace analysis and simulation tools.

clients/drcachesim/common/trace_entry.h
```diff
@@ -73,10 +73,16 @@ typedef enum {
...
     TRACE_TYPE_PREFETCH, /**< A general prefetch to the level 1 data cache. */
     // X86 specific prefetch
-    TRACE_TYPE_PREFETCHT0,  /**< An x86 prefetch to all levels of the cache. */
-    TRACE_TYPE_PREFETCHT1,  /**< An x86 prefetch to level 1 of the cache. */
-    TRACE_TYPE_PREFETCHT2,  /**< An x86 prefetch to level 2 of the cache. */
-    TRACE_TYPE_PREFETCHNTA, /**< An x86 non-temporal prefetch. */
+    TRACE_TYPE_PREFETCHT0, /**< An x86 prefetch to all levels of the cache. */
+    TRACE_TYPE_PREFETCH_READ_L1 =
+        TRACE_TYPE_PREFETCHT0, /**< Load prefetch to L1 cache. */
+    TRACE_TYPE_PREFETCHT1,     /**< An x86 prefetch to level 2 cache and higher. */
```
abhinav92003, line 79: Corrected comments for TRACE_TYPE_PREFETCHT1 and TRACE_TYPE_PREFETCHT2 based on [1].

[1]: https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-instruction-set-reference-manual-325383.pdf 4-402 Vol. 2B

clients/drcachesim/common/trace_entry.h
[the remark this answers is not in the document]
abhinav92003: I added some changes to address this comment. It seems complicated to get an architecture independent set of trace types for prefetch instructions.

About 32-bit ARM prefetch instructions (PLD, PLDW and PLI), ARM ISA does not specify the target cache level[1]. I did find some references that mention a target cache level[2][3], but it seems implementation defined. As we don't know the target cache level, the existing ARM-specific prefetch trace types don't seem to map to any of the new generic ones.

About x86, the PREFETCHTX instrs map to TRACE_TYPE_PREFETCH_READ_LX trace types. (Based on [4], these prefetches do _not_ anticipate write.)
However, PREFETCHW (which anticipates write) can prefetch into L1 _or_ L2[5], so that also doesn't map directly to any one of the new trace type. 

[1]: https://static.docs.arm.com/100076/0100/arm_instruction_set_reference_guide_100076_0100_00_en.pdf C2-272
[2]: https://developer.arm.com/documentation/ddi0488/d/level-1-memory-system/l1-data-memory-system/preload-instruction-behavior
[3]: https://developer.arm.com/documentation/ddi0535/b/l1-memory-system/data-prefetching/pld--pldw--and-pli-instructions
[4]:  https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-instruction-set-reference-manual-325383.pdf 4-402 Vol. 2B
[5]: https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-instruction-set-reference-manual-325383.pdf 4-404 Vol. 2B

clients/drcachesim/tests/prefetch-aarch64-basic-counts.templatex
abhinav92003: Added a custom analyzer test for a pure asm app that checks each of the new prefetch trace types.

clients/drcachesim/common/trace_entry.h
```diff
@@ -73,10 +73,16 @@ typedef enum {
...
+    TRACE_TYPE_PREFETCH_READ_L2 =
+        TRACE_TYPE_PREFETCHT1, /**< Load prefetch to L2 cache. */
+    TRACE_TYPE_PREFETCHT2,     /**< An x86 prefetch to level 3 cache and higher. */
+    TRACE_TYPE_PREFETCH_READ_L3 =
+        TRACE_TYPE_PREFETCHT2, /**< Load prefetch to L3 cache. */
+    TRACE_TYPE_PREFETCHNTA,    /**< An x86 non-temporal prefetch. */
```
derekbruening, line 85: Please comment what makes this different from the other level-specific non-temporal prefetches.

clients/drcachesim/tests/offline-allasm-aarch64-prefetch-counts.templatex
```diff
@@ -0,0 +1,50 @@
...
+           5 prefetch_instr_l2
+           4 prefetch_instr_l2_nt
+           3 prefetch_instr_l3
+           2 prefetch_instr_l3_nt
+           5 prefetch_write_l1
+           4 prefetch_write_l1_nt
+           3 prefetch_write_l2
+           2 prefetch_write_l2_nt
+           7 prefetch_write_l3
+           6 prefetch_write_l3_nt
```
derekbruening, line 50: Missing newline on final line

clients/drcachesim/tests/offline-builtin-prefetch-basic-counts.templatex
```diff
@@ -0,0 +1,39 @@
...
+#endif
+     .* data loads
+     .* data stores
+     .* scheduling markers
+     .* transfer markers
+     .* function id markers
+     .* function return address markers
+     .* function argument markers
+     .* function return value markers
+     .* other markers
```
derekbruening, line 39: Missing newline on final line

clients/drcachesim/tests/prefetch_analyzer_launcher.cpp
```diff
@@ -0,0 +1,84 @@
...
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+/* Standalone prefetch analyzer tool launcher for file traces. */
```
derekbruening, line 33: Hmm, I think I suggested originally putting the analyzer into a static start/stop test, which would have the whole test all in one place and not require multiple binaries to be built, with no separate special launcher or separate analyzer.  Did you end up deciding against that?  What was the reasoning?  We currently have tests that run raw2trace after gathering: is there a blocker to running an analyzer there?  I would think the analyzer would have fewer issues than raw2trace which has some link complexities on Windows.

suite/tests/CMakeLists.txt
```diff
@@ -3331,6 +3353,12 @@ if (CLIENT_INTERFACE)
       torunonly_drcacheoff(instr-only-trace ${ci_shared_app} "-instr_only_trace" "" "")
       torunonly_drcacheoff(filter-and-instr-only-trace ${ci_shared_app} "-instr_only_trace -L0_filter" "" "")
 
+      # __builtin_prefetch used in the test is available only on gcc.
```
derekbruening, line 3356: Repeating comment from before: no, it is not limited to gcc.  It is supported by clang as well.

abhinav92003: Oops. Sorry I didn't address this before. __builtin_prefetch is not available in MSVC. I changed the condition to reflect that.

clients/drcachesim/tests/offline-builtin-prefetch-basic-counts.templatex
abhinav92003: Added.

clients/drcachesim/tests/offline-allasm-aarch64-prefetch-counts.templatex
abhinav92003: Added.

clients/drcachesim/common/trace_entry.h
abhinav92003: [1] doesn't specify what cache level prefetchnta targets -- only that it "—prefetch data into non-temporal cache structure and into a location close to the processor, minimizing cache pollution.". Added that to a comment here.

[1]: https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-instruction-set-reference-manual-325383.pdf 4-402 Vol. 2B

clients/drcachesim/tests/prefetch_analyzer_launcher.cpp
abhinav92003: Based on the latest update on #1551, I think start/stop support is not implemented yet for ARM. As discussed offline, existing burst_* tests that use the start/stop api are also not enabled for AArch64 yet due to #2007.

third_party/dynamorio/trunk/suite/tests/CMakeLists.txt:3408
`# FIXME i#2007: fails to link on A64`

clients/drcachesim/common/trace_entry.h
```diff
@@ -163,6 +163,49 @@ typedef enum {
      */
     TRACE_TYPE_INSTR_SYSENTER,
 
+    /** AArch64 specific prefetch trace entries.
+     *  Relative order of these enum values is relevant to the handling of AArch64
+     * prefetch instructions.
+     */
+    TRACE_TYPE_PREFETCH_PLDL1KEEP, /**< An AArch64 load prefetch to level 1 of the cache.
+                                    */
+    TRACE_TYPE_PREFETCH_PLDL1STRM, /**< An AArch64 streaming load prefetch to level 1 of
```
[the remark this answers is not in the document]
abhinav92003: Renamed to make the enums architecture-agnostic. More related discussion in the other comment thread: https://github.com/DynamoRIO/dynamorio/pull/4341#discussion_r447799402

clients/drcachesim/common/trace_entry.h
```diff
@@ -71,12 +71,22 @@ typedef enum {
     TRACE_TYPE_READ,  /**< A data load. */
     TRACE_TYPE_WRITE, /**< A data store. */
 
-    TRACE_TYPE_PREFETCH, /**< A general prefetch to the level 1 data cache. */
+    TRACE_TYPE_PREFETCH, /**< A general prefetch. */
```
abhinav92003, line 74: I updated the comment for TRACE_TYPE_PREFETCH too. IIUC, it was intended to be a 'default' prefetch trace type, for operations that don't fall under any other existing type.

clients/drcachesim/tracer/instru.cpp
```diff
@@ -78,6 +78,72 @@ instru_t::instr_to_instr_type(instr_t *instr, bool repstr_expanded)
...
+        switch (target_policy) {
+        case 0b000: return TRACE_TYPE_PREFETCH_WRITE_L1;
+        case 0b001: return TRACE_TYPE_PREFETCH_WRITE_L1_NT;
+        case 0b010: return TRACE_TYPE_PREFETCH_WRITE_L2;
+        case 0b011: return TRACE_TYPE_PREFETCH_WRITE_L2_NT;
+        case 0b100: return TRACE_TYPE_PREFETCH_WRITE_L3;
+        case 0b101: return TRACE_TYPE_PREFETCH_WRITE_L3_NT;
+        }
+    }
+    DR_ASSERT_MSG(false, "Unsupported AArch64 prefetch operation.");
```
abhinav92003, line 142: [1] mentions that there may be some encodings not accessible using prfop. I'm not sure how these work. Do you suggest we simply map them to the default `TRACE_TYPE_PREFETCH`, instead of throwing an error?

[1] https://static.docs.arm.com/ddi0596/a/DDI_0596_ARM_a64_instruction_set_architecture.pdf Page 829

clients/drcachesim/tests/prefetch_analyzer_launcher.cpp
derekbruening: Hmm, ok, the #1551 is 32-bit arm.  I recall running attach/detach on aarch64 and it working -- but yes on the Travis setup there was some link issue (fixed on machines we control w/ a binutils tweak).  Sure, since you already have this let's go with this.

clients/drcachesim/tracer/instru.cpp
derekbruening: > there may be some encodings not accessible using prfop

You mean 0b110 and 0b111?

Pros and cons depending on the use case.  If we did map to the default we should have a warning or debug-build-only error.  I guess I would lean toward release build mapping to generic and debug throwing an error for something like this.

clients/drcachesim/common/trace_entry.h
derekbruening: OK, sure.

clients/drcachesim/tracer/instru.cpp
abhinav92003: 0bxx110, 0bxx111 and also 0b11xxx. 

I made changes so that the assert applies only for DEBUG builds. The offline trace test prints the error message and exits as expected. But the online test doesn't exit after printing and needs to be terminated manually. Is this expected?
`ASSERT FAILURE: /home/abhinavas/dr/src/i4328-1/clients/drcachesim/tracer/instru.cpp:144: false (Unsupported AArch64 prefetch operation.)`

derekbruening: > But the online test doesn't exit after printing and needs to be terminated manually.

You mean the test app exits, but the simulator process doesn't detect the exit?  I think we try to close the pipe on an exit but there are definitely still some outstanding hangs, such as #4296.  Maybe worth a quick check to ensure the pipe is closed -- do we need a custom abort.

abhinav92003: I used `top` to check which process has trouble terminating.
```
 ./bin64/drrun -t drcachesim -simulator_type basic_counts -- suite/tests/bin/allasm_aarch64_prefetch
```
`allasm_aarch64_prefetch` app terminates, but `drcachesim` doesn't

But the following workaround helps. It exits as expected.
```
./bin64/drrun -t drcachesim  -dr_ops -unsafe_build_ldstex -simulator_type basic_counts -- setarch `uname -m` -R    suite/tests/bin/allasm_aarch64_prefetch
```
I confirmed that we need both `-dr_ops -unsafe_build_ldstex`  and the `setarch`

derekbruening: The `-unsafe_build_ldstex` is #1698 and that would eliminate a hang in the app process, which is not what you saw: you saw a hang in the simulator process (drcachesim).

The `setarch` is strange: that is the opposite of #4296 where launching under `setarch` *caused* a hang.  On your machine does #4296 not reproduce as described in the bug?

abhinav92003: In this case, the hang occurs even without the `setarch`, so perhaps this is a different issue?

Another observation:
#4296 suggests running `drrun` under the `setarch`. However, in this case, the following hangs as well
```
setarch `uname -m` -R ./bin64/drrun -t drcachesim  -simulator_type basic_counts -- suite/tests/bin/allasm_aarch64_prefetch
```

abhinav92003: The following patch allows to reproduce this issue with the allasm_aarch64_prefetch app on debug builds.

```
clients/drcachesim/tests/allasm_aarch64_prefetch.asm
@@ -163,6 +163,7 @@ _start:
         prfum    pstl3strm, [x1]
         prfum    pstl3strm, [x1]
 
+        prfum    28, [x1]
         // Exit.
         mov      w0, #1            // stdout
         adr      x1, helloworld
```

abhinav92003: > On your machine does #4296 not reproduce as described in the bug?

On my machine, the issue reproduces as described. However, the workaround doesn't. All of the following hang, with the app process and `drcachesim` still showing up in `top`:
```
bin64/drrun -t drcachesim -- suite/tests/bin/simple_app
bin64/drrun -t drcachesim -- setarch `uname -m` -R suite/tests/bin/simple_app
setarch `uname -m` -R bin64/drrun -t drcachesim --  suite/tests/bin/simple_app
```

The following also hangs, but neither drcachesim nor the app show up in `top`. Weird.
```
bin64/drrun -t drcachesim  -dr_ops -unsafe_build_ldstex  -- setarch `uname -m` -R suite/tests/bin/simple_app
```

However, the following works:
```
setarch `uname -m` -R bin64/drrun -t drcachesim  -dr_ops -unsafe_build_ldstex  --  suite/tests/bin/simple_app
```

derekbruening: Make sure to delete any stale pipe file before any subsequent run: that will cause a hang.  Also, when you say "in top", are you only looking at the top N sorted by cpu usage?  Be sure to look at all processes through whatever method.

abhinav92003: > The following also hangs, but neither drcachesim nor the app show up in top. Weird.
> 
> bin64/drrun -t drcachesim  -dr_ops -unsafe_build_ldstex  -- setarch `uname -m` -R suite/tests/bin/simple_app

`ps -x` indeed shows both `drcachesim` and the app when this command is run. Thanks.


I ran each of the commands in my last comment and verified that `/tmp/drcachesimpipe` gets deleted after I kill the previous process. Same behaviour.
There is a `/tmp/drtestpipe` that's not owned by me on the system, but that is used only by `ctest` so shouldn't be the cause here.

abhinav92003: This issue seems different than #4296 -- it comes up on assert failure only, and without `setarch`. Do we investigate further here or handle it in a separate issue?

derekbruening: +1 for separate.

abhinav92003: Raised #4376 for this bug.

PR #4342, opened 2020-06-30
i#4318 xarch memtrace: Bump version and add message for arch tags

derekbruening:
PR #4331 commit 5d49675c added a new marker with the architecture
early in a trace, enough to break the file_reader_t assumptions.  The
file reader was updated, but it's best to also bump the version for
old binary code.  Additionally, we provide a separate distinct message
about a version mismatch (otherwise the error message from the old
reader is the same as without the version bump).

Issue: #4318

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -136,10 +136,6 @@ jobs:
       compiler: gcc
       env: DYNAMORIO_CROSS_ANDROID_ONLY=yes DYNAMORIO_ANDROID_TOOLCHAIN='/tmp/android-gcc-arm-ndk-10e' DEPLOY=yes EXTRA_ARGS=32_only
 
-before_install:
-  # workaround for https://github.com/travis-ci/travis-ci/issues/8973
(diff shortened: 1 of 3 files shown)


derekbruening:
This will just provide a better error message on a future version bump.  I think users having to update their readers to read new versions is reasonable and cleaner than jumping through hoops to make new trace features readable by old readers.  Keeping old traces readable by new readers is important though and we are maintaining that.

derekbruening:
Hmm Mac is the failure and its output is truncated: xref #4223, and PR #4341 hitting that, and @abhinav92003 's workaround found in the Travis issue tracker: https://github.com/DynamoRIO/dynamorio/pull/4341/commits/18c2a9d006c407a09a50484a45052012c4f3eb23

derekbruening:
Since this PR is sort of time sensitive with the version bumping that should have gone in with the prior commit I'm going to put it in with Mac red.  I tried to build Mac locally but the local machine had XCode upgraded and now doesn't build: I filed #4344.  Grrr.

Issue #4344, opened 2020-06-30
DR fails to configure with recent XCode on OSX due to standalone cpp problems


derekbruening:
```
CMake Error at CMakeLists.txt:1343 (message):
  ***
  /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cpp
  failed: ***

  In file included from
  /.../dr/git/build_x64_dbg/configure_temp.h:150:

  In file included from
  /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/sys/types.h:184:


  In file included from
  /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/sys/_types/_fd_def.h:32:


  In file included from
  /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/Availability.h:257:


  
  /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/AvailabilityInternal.h:147:2:
  error: #endif without #if

  #endif /* __ENABLE_LEGACY_IPHONE_AVAILABILITY */

   ^
```

```
$ cpp --version
Apple clang version 11.0.3 (clang-1103.0.32.62)
Target: x86_64-apple-darwin19.5.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

Others have hit this and it seems like /usr/bin/cpp with this xcode just
can't handle system headers b/c it passes -traditional-cpp to clang;
https://stackoverflow.com/questions/42031921/unterminated-unconditional-directive-error-when-preprocessing-c-code

Stated solution is to either install own cpp, or use compiler driver instead
of standalone cpp.  I don't know that the latter is an option for us as we
pass many particular options to cpp that IIRC are not exposed by the
compiler driver.

derekbruening:
I tried negating the `-traditional-cpp` but failed:
```
clang: error: unknown argument: '-no-traditional-cpp'
clang: error: unknown argument: '-notraditional-cpp'
```

I tried using `CMAKE_C_COMPILER` instead of `CMAKE_CPP` and hit an include path problem:
```
[...]

PR #4345, opened 2020-07-01
i#4223: Fix truncated log issue for macOS on Travis

abhinav92003:
Sleep for 10 seconds after script run/failure to allow some time for Travis to capture the complete log.  It was used as a workaround also at travis-ci/travis-ci#6018.

Also attempted another workaround (using fcntl) from travis-ci/travis-ci#8973 but it didn't work as expected.

Fixes: #4223

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -136,10 +136,6 @@ jobs:
       compiler: gcc
       env: DYNAMORIO_CROSS_ANDROID_ONLY=yes DYNAMORIO_ANDROID_TOOLCHAIN='/tmp/android-gcc-arm-ndk-10e' DEPLOY=yes EXTRA_ARGS=32_only
 
-before_install:
-  # workaround for https://github.com/travis-ci/travis-ci/issues/8973
-  - python -c "import fcntl; fcntl.fcntl(1, fcntl.F_SETFL, 0)"
-
 # For C/C++ there is no default install, so we set "install", not "before_install".
 install:
   - uname -a
--- a/.travis.yml
+++ b/.travis.yml
@@ -136,6 +136,10 @@ jobs:
       compiler: gcc
       env: DYNAMORIO_CROSS_ANDROID_ONLY=yes DYNAMORIO_ANDROID_TOOLCHAIN='/tmp/android-gcc-arm-ndk-10e' DEPLOY=yes EXTRA_ARGS=32_only
 
+before_install:
+  # workaround for https://github.com/travis-ci/travis-ci/issues/8973
+  - python -c "import fcntl; fcntl.fcntl(1, fcntl.F_SETFL, 0)"
+
 # For C/C++ there is no default install, so we set "install", not "before_install".
 install:
   - uname -a
--- a/.travis.yml
+++ b/.travis.yml
@@ -106,7 +106,7 @@ jobs:
       os: osx
       # gcc on Travis claims to not be CMAKE_COMPILER_IS_GNUCC so we only run clang.
       compiler: clang
-      env: DYNAMORIO_CROSS_AARCHXX_LINUX_ONLY=no DEPLOY=no EXTRA_ARGS=64_only
+      env: DYNAMORIO_CROSS_AARCHXX_LINUX_ONLY=no DEPLOY=no EXTRA_ARGS=64_only SLEEP_AFTER_SCRIPT_OR_FAILURE=yes
     # AArch64 drdecode and drmemtrace on x86:
     - if: type != cron AND env(TRAVIS_EVENT_TYPE) != cron
       os: linux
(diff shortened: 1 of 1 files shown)


abhinav92003:
@derekbruening Based on travis-ci/travis-ci#6018, the macOS build log gets truncated because the VM is terminated too early. The log looks complete now: https://travis-ci.com/github/DynamoRIO/dynamorio/jobs/356357628 (I reran Travis to make sure). I admit it doesn't seem an ideal fix, but it might be helpful for now.

Also, about commit cced012, more info that I saw in the log was just a coincidence -- after I reran Travis, the additional log lines disappeared.

.travis.yml
```diff
@@ -188,6 +184,14 @@ install:
 script:
   - travis_wait 45 suite/runsuite_wrapper.pl travis $EXTRA_ARGS
 
+after_script:
+  - sleep 10
+  - echo "--end--"
+
+after_failure:
+  - sleep 10
+  - echo "--end--"
```
derekbruening, line 193: Could these be made specific to the Mac job only?

derekbruening, line 193: A comment explaining why these are here would be good.

derekbruening: Does it not work to put this inside the job matrix and so local to the Mac job w/o needing a global var to connect them?

abhinav92003: Here, `sleep` keeps the VM around long enough for all logs to be captured. As build is done after the job matrix is set up, there aren't any pending logs at that point, so IIUC moving `sleep` to the job matrix won't work.

abhinav92003: Done.

.travis.yml
```diff
@@ -184,6 +184,20 @@ install:
...
+  # Travis logs for OSX builds are truncated because the VM exits early.
+  # Sleep to allow Travis to capture complete log. More context on
+  # DynamoRIO/dynamorio#4223.
+  - test $SLEEP_AFTER_SCRIPT_OR_FAILURE = "yes" && sleep 10
+  - echo "--end--"
+
+after_failure:
+  # Travis logs for OSX builds are truncated because the VM exits early.
+  # Sleep to allow Travis to capture complete log. More context on
+  # DynamoRIO/dynamorio#4223.
```
derekbruening, line 197: Please update #4223 with the Travis bug link and what you tried and whether it worked (the fcntl, and the sleep).

.travis.yml
derekbruening: Making sure we're talking about the same thing: I mean putting an `after_script` and `after_failure` config setting inside the Mac job entry, not some other way of adding a sleep.  You're saying Travis won't accept `after_script` inside the matrix for a job-specific config?

.travis.yml
abhinav92003: Done.

.travis.yml
abhinav92003: Sorry, I misunderstood your suggestion. Made the changes. Thanks.

derekbruening: NP.  Thanks for fixing: IMHO it's cleaner this way.

.travis.yml
```diff
@@ -107,6 +107,19 @@ jobs:
       # gcc on Travis claims to not be CMAKE_COMPILER_IS_GNUCC so we only run clang.
       compiler: clang
       env: DYNAMORIO_CROSS_AARCHXX_LINUX_ONLY=no DEPLOY=no EXTRA_ARGS=64_only
+      after_script:
+        # Travis logs for OSX builds are truncated because the VM exits early.
+        # Sleep to allow Travis to capture complete log. More context on
+        # DynamoRIO/dynamorio#4223.
```
derekbruening, line 113: nit: Our convention (from before the move to github) has been `i#4223`; ditto below.  That is what I search for in the code.

abhinav92003: Fixed it. Thanks.

derekbruening:
Maybe upgrading the compiler (see `osx_image: xcode11.3` in .travis.yml) will solve the compiler OOM here -- OTOH it might hit more problems (like #4344).

derekbruening:
> Maybe upgrading the compiler (see `osx_image: xcode11.3` in .travis.yml) will solve the compiler OOM here -- OTOH it might hit more problems (like #4344).

Maybe file a separate issue on the OOM.  You'd think our Mac VM's would be pegged at XCode 11.3: so what changed to cause this compiler failure?  It appeared in between our own commits, right, so it was not triggered by a change we made?

abhinav92003:
> > Maybe upgrading the compiler (see `osx_image: xcode11.3` in .travis.yml) will solve the compiler OOM here -- OTOH it might hit more problems (like #4344).
> 
> Maybe file a separate issue on the OOM. You'd think our Mac VM's would be pegged at XCode 11.3: so what changed to cause this compiler failure? It appeared in between our own commits, right, so it was not triggered by a change we made?

The failure seems to appear between #4339 and #4342, so it's not caused by us. Sure I'll file a separate issue and submit this one first.

abhinav92003:
> The failure seems to appear between #4339 and #4342, so it's not caused by us. Sure I'll file a separate issue and submit this one first.

Filed #4347

Issue #4349, opened 2020-07-01
api.thread_churn test failing on Windows due to reattach problems


derekbruening:
My new test api.thread_churn for #4334 is failing on Windows due to some re-attach issues:

+ .data is protected causing a crash on init
+ failing to take over threads: 80 such messages on the re-attach:
```
<Detaching from process, entering final cleanup>
<Starting application D:\derek\dr\git\build_x64_dbg_tests\suite\tests\bin\api.thread_churn.exe (7644)>
<Running on newer-than-this-build "Microsoft Windows 10-1909 x64">
<Early threads found>
<Initial options = -no_dynamic_options -loglevel 2 -code_api -probe_api -msgbox_mask 0 -dumpcore_mask 125 -stderr_mask 15 -stack_size 56K -max_elide_jmp 0 -max_elide_call 0 -no_inline_ignored_syscalls -staged -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct >
<CURIOSITY : (0) && "failed to take over a thread!" in file D:\derek\dr\git\src\core\win32\os.c line 2715
version 8.0.18444, custom build
-no_dynamic_options -loglevel 2 -code_api -probe_api -msgbox_mask 0 -dumpcore_mask 125 -stderr_mask 15 -stack_size 56K -max_elide_jmp 0 -max_elide_call 0 -no_inline_ignored_syscalls -staged -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct >
<CURIOSITY : (0) && "failed to take over a thread!" in file D:\derek\dr\git\src\core\win32\os.c line 2715
version 8.0.18444, custom build
-no_dynamic_options -loglevel 2 -code_api -probe_api -msgbox_mask 0 -dumpcore_mask 125 -stderr_mask 15 -stack_size 56K -max_elide_jmp 0 -max_elide_call 0 -no_inline_ignored_syscalls -staged -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct >
<...>
```

I tried to use logging to diagnose but hit another:
+ log file not being opened
  Xref #2661 on ops appending: but here the option is not processed at all, leaving d_r_stats->loglevel 0.

It's not clear what's going on.  The test makes 10 threads in the first burst.  Is some state left over?  Fixing logging would make it much easier to diagnose.  For now I'm disabling the test on Windows.

Issue #271, opened 2014-11-27
CRASH  instrument_post_syscall() race w/ another thread


derekbruening:
_From [derek.br...@gmail.com](https://code.google.com/u/117968039472581148324/) on March 11, 2010 21:59:29_

one thread terminates the process, but in release build we don't synch w/
the other threads.  for CLIENT_INTERFACE we go and call
instrument_thread_exit() for the other threads, though, and can hit crashes
like this one, where dcontext->client_data is set to NULL for another
thread that then de-refs it.  even for non-CI there can be races that
result in crashes if other threads refer to global data structs that get
cleaned up: though we don't do that much cleanup in release build.

There are two problems to be solved:

A) A thread can execute beyond its own exit event, causing correctness
   problems in clients

B) A thread can execute beyond its data structures or global data
   structures being freed, causing crashes

I took a multi-pronged approach to solving both:

1) if no thread exit event exists then A is not possible.
   avoid B by skipping instrument_thread_exit().
   if other threads alive at process exit event, client should do synch
   when combining data.  then we don't care what happens after exit event
   since client persistent data is all written out so who cares what other
   threads do between then and actual process death.

   actually we do fragment_exit(), etc. before client exit event,
   so other threads can crash: the change for `#2` about not reporting such
   crashes will help.  we could do the send-out-signals even if no thread
   exit event: not going to bother though, clients can ask for full synch
   if nec.

2) the default option when a thread exit event exists is best-effort
   for both A and B but no guarantees for either:
- for A: attempt to prevent other threads from executing beyond their
   own exit events: on linux, send custom terminate signal (but don't
   wait for it), on Windows call NtTerminateProcess(0)
   we just live w/ risk of thread holding a lock since very few locks are
   used beyond this point.
- for B, the solution to A helps.  I also made two further changes:
  - don't free DR data in instrument_thread_exit() to reduce incidence
    of crashes
  - do not report SIGSEGV/exception when dynamo_exited=true and
    multiple threads exist.
    this could potentially mask other bugs: we'll live with that.
    A change as part of implementing that last one on Windows:
  - with -private_loader, ntdll hooks are not removed on release build exit
    so we will squelch exit-time crashes

3) -synch_at_exit option to guarantee a solution to both A and B.
   APi routine dr_request_synchronized_exit() also sets this.
   I documented this in the API docs.
   if client needs guarantee that thread exit event will never be followed
   by execution by that thread then it should set this option.

[...]

PR #4350, opened 2020-07-01
i#4334: Fix leaks on thread exit

### LINKED ISSUE #4334: Test application on RHEL 7.8 with DynamoRio-Linux 8.0.0 runs out of memory and program gets aborted
We have a 64-bit test application for RHEL 7.8 that can run natively for more 24+ hours. But when it is run under DynamoRio-Linux 8.0.0 without any client, it gets aborted by the RHEL OS in ~10 minutes,

	$  /opt/dr-8.0.0-1/bin64/drrun -version
	drrun version 8.0.0 -- build 1
	$ date ; /opt/dr-8.0.0-1/bin64/drrun -- ./execute ; date
	Wed Jun 24 17:36:13 EDT 2020
	server: got connection from 127.0.0.1 port 59154
	Wed Jun 24 17:49:14 EDT 2020

When run under the DynamoRio-Linux 7.1.0 release without any client, it runs out of memory in about 1 hour.

We have examined the memory usage using the "pmap" command at different time intervals. When the application is run natively, the total memory usage does not change. But under, DR 8.0.0 or DR 7.1.0, we notice increasing total memory usage with the time.

We would appreciate any help you can provide in the debug of this problem.

derekbruening:
These leaks are all cleaned up at process exit, which is why the
existing exit-time leak checks don't notice anything.  The global
block list lets us clean them up later, and hence a classic
reachability-based leak detector would not catch them either: so
perhaps they should be called "reachable accumulations" instead of
"leaks".

Fixes a memory leak on thread exit in all builds:
+ The private reachable_heap units were not being freed

Fixes a number of memory leaks on thread exit in release build:
+ Adds individual private fragment deletion on thread exit, to free
  stubs if -separate_private_stubs is on.
  Turns off -separate_private_stubs by default, to avoid this for
  shared caches where there are few benefits to separating private stubs.
  -thread_private turns it back on where the benefits probably outweight
  the thread exit costs.
+ Moves the privload_tls_exit call to release build too to properly
  unmap the TLS.
+ Moves the sigaltstack free to release build to properly free it.
+ Moves freeing of "local unprotected" heap to the release-build path,
  since it is actually global!  This is done for kstats, stats, and
  clone_tls.
+ Moves client_data_t, client_todo_list_t, and thread-private
  client_flush_req_t to be PROTECTED to make it thread-local and thus
  not need freeing (the free is avoided for i#271).

Adds an -rstats_to_stderr dump point when DR terminates for OOM to
make it much easier to diagnose what the problem is.

Adds a new api.thread_churn test which attaches twice, once with a few
threads and once with many threads, and confirms memory usage has not
gone up.  The 5 rstats on peak vmm block sizes are added to dr_stats_t
to facilitate this.

I tried to get api.thread_churn to work on Windows but hit a number of
issues with reattach.  I solved two of them before bailing and making
the test UNIX-only for now and filed the rest as #4349.
+ Unprotect .data on reattach
+ Clear the logfile flag on detach

Issue: #4334, #271, #4349
Fixes #4334

### DIFF
--- a/core/lib/instrument.c
+++ b/core/lib/instrument.c
@@ -1345,8 +1345,11 @@ void
 instrument_client_thread_init(dcontext_t *dcontext, bool client_thread)
 {
     if (dcontext->client_data == NULL) {
+        /* We use PROTECTED partly to keep it local (unprotected "local" heap is
+         * global and adds complexity to thread exit: i#271).
+         */
         dcontext->client_data =
-            HEAP_TYPE_ALLOC(dcontext, client_data_t, ACCT_OTHER, UNPROTECTED);
+            HEAP_TYPE_ALLOC(dcontext, client_data_t, ACCT_OTHER, PROTECTED);
         memset(dcontext->client_data, 0x0, sizeof(client_data_t));
 
 #    ifdef CLIENT_SIDELINE
@@ -1462,7 +1465,7 @@ instrument_thread_exit(dcontext_t *dcontext)
 #    endif
 
 #    ifdef DEBUG
-    /* PR 470957: avoid racy crashes by not freeing in release build */
+    /* i#271: avoid racy crashes by not freeing in release build. */
 
 #        ifdef CLIENT_SIDELINE
     DELETE_LOCK(dcontext->client_data->sideline_mutex);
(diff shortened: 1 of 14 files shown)


derekbruening:
The failures look like flakes: fib-conflict (#3406) and client.winxfer (#4058)

core/lib/instrument.c
```diff
@@ -1462,7 +1465,7 @@ instrument_thread_exit(dcontext_t *dcontext)
 #    endif
 
 #    ifdef DEBUG
-    /* PR 470957: avoid racy crashes by not freeing in release build */
+    /* i#271: avoid racy crashes by not freeing in release build. */
```
johnfxgalea, line 1468: Your commit message suggests that these changes are done for release builds but here we have an ifdef DEBUG.

https://github.com/DynamoRIO/dynamorio/pull/4350#issue-443017097

derekbruening: Actually the commit message lists 3 other different places where the freeing is moved to release build, and has a separate list item for this one: "Moves client_data_t, client_todo_list_t, and thread-private client_flush_req_t to be PROTECTED to make it thread-local and thus not need freeing (the free is avoided for i#271).".  I will make it clearer in the message that this stays DEBUG-only.

PR #4353, opened 2020-07-02
i#4335: Fix out of memory error due to allocation of more than 2GB in unreachable heap

### LINKED ISSUE #4335: Allocation of more than 2GB in unreachable heap causes out of memory errror
# How to reproduce

This reproduces on both Linux x86-32 and x86-64.

Just allocate more than 2GB from `dr_client_main` with `malloc`:

```
    int *tmp = 0;
    int cnt = 0;
    for (int i = 0; i != 4 * 96086; ++i)
    {
        tmp = (int *) malloc(30000);
        ++cnt;
        std::cout << cnt << std::endl;
        for (int j = 0; j != 30000 / sizeof(int); ++j)
        {
            tmp[j] = i * 43;
        }
    }
    std::cout << "Do not optimize out malloc " << tmp[20] << std::endl;
```

Run debug:

```
dynamorio/bin64/drrun -debug -c client.so -- /bin/ls
```

Get an error:

```
<Out of vmheap reservation - reserving 256KB. Falling back onto OS allocation>
<CURIOSITY : (0) && "Out of vmheap reservation" in file /home/vishnya/dynamorio/core/heap.c line 1614
<Out of memory - cannot reserve 256KB. Trying to recover.>
<OOM Status: R 0x0000000000000001>
```

# What goes wrong

When unreachable memory size reaches first 2GB it tries to ask more memory from system. But [this](https://github.com/DynamoRIO/dynamorio/blob/master/core/heap.c#L2825) code fails and leads to out of memory.

SweetVishnya:
Closes #4335

### DIFF
--- a/core/heap.c
+++ b/core/heap.c
@@ -1545,7 +1545,7 @@ vmm_heap_reserve(size_t size, heap_error_code_t *error_code, bool executable,
             });
             reached_beyond_vmm();
 #ifdef X64
-            if (which & VMM_REACHABLE) {
+            if (TEST(VMM_REACHABLE, which) || REACHABLE_HEAP()) {
                 /* PR 215395, make sure allocation satisfies heap reachability
                  * contraints */
                 p = os_heap_reserve_in_region(
--- a/core/heap.c
+++ b/core/heap.c
@@ -1517,7 +1517,7 @@ schedule_low_on_memory_event_trigger()
 /* Reserve virtual address space without committing swap space for it */
 static vm_addr_t
 vmm_heap_reserve(size_t size, heap_error_code_t *error_code, bool executable,
-                 which_vmm_t which)
+                 which_vmm_t which, bool reachable)
 {
     vm_addr_t p;
     vm_heap_t *vmh = vmheap_for_which(which);
@@ -1545,14 +1545,18 @@ vmm_heap_reserve(size_t size, heap_error_code_t *error_code, bool executable,
             });
             reached_beyond_vmm();
 #ifdef X64
-            /* PR 215395, make sure allocation satisfies heap reachability contraints */
-            p = os_heap_reserve_in_region(
-                (void *)ALIGN_FORWARD(heap_allowable_region_start, PAGE_SIZE),
-                (void *)ALIGN_BACKWARD(heap_allowable_region_end, PAGE_SIZE), size,
-                error_code, executable);
-            /* ensure future heap allocations are reachable from this allocation */
-            if (p != NULL)
-                request_region_be_heap_reachable(p, size);
+            if (reachable) {
+                /* PR 215395, make sure allocation satisfies heap reachability
+                 * contraints */
+                p = os_heap_reserve_in_region(
+                    (void *)ALIGN_FORWARD(heap_allowable_region_start, PAGE_SIZE),
+                    (void *)ALIGN_BACKWARD(heap_allowable_region_end, PAGE_SIZE), size,
+                    error_code, executable);
+                /* ensure future heap allocations are reachable from this allocation */
+                if (p != NULL)
+                    request_region_be_heap_reachable(p, size);
+            } else
+                p = os_heap_reserve(NULL, size, error_code, executable);
 #else
             p = os_heap_reserve(NULL, size, error_code, executable);
 #endif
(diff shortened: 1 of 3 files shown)


core/heap.c
```diff
@@ -1517,7 +1517,7 @@ schedule_low_on_memory_event_trigger()
 /* Reserve virtual address space without committing swap space for it */
 static vm_addr_t
 vmm_heap_reserve(size_t size, heap_error_code_t *error_code, bool executable,
-                 which_vmm_t which)
+                 which_vmm_t which, bool reachable)
```
derekbruening, line 1520: This parameter shouldn't be necessary: instead test `which` for `VMM_REACHABLE`.

core/heap.c
```diff
@@ -1545,14 +1545,18 @@ vmm_heap_reserve(size_t size, heap_error_code_t *error_code, bool executable,
...
+                 * contraints */
+                p = os_heap_reserve_in_region(
+                    (void *)ALIGN_FORWARD(heap_allowable_region_start, PAGE_SIZE),
+                    (void *)ALIGN_BACKWARD(heap_allowable_region_end, PAGE_SIZE), size,
+                    error_code, executable);
+                /* ensure future heap allocations are reachable from this allocation */
+                if (p != NULL)
+                    request_region_be_heap_reachable(p, size);
+            } else
+                p = os_heap_reserve(NULL, size, error_code, executable);
```
derekbruening, line 1559: Would it be feasible to add a test for this?  For the test you put in the issue: the concern with making that a regression test is whether it's going to cause performance issues.  On Linux, for large mallocs I guess only the header page will be touched and thus backed by physical memory...but on Windows won't the entire thing be backed?  Maybe make it a (64-bit) Linux-only test?

SweetVishnya: Added Linux x64 only test.

core/heap.c
```diff
@@ -1545,14 +1545,18 @@ vmm_heap_reserve(size_t size, heap_error_code_t *error_code, bool executable,
...
 #ifdef X64
-            /* PR 215395, make sure allocation satisfies heap reachability contraints */
-            p = os_heap_reserve_in_region(
-                (void *)ALIGN_FORWARD(heap_allowable_region_start, PAGE_SIZE),
-                (void *)ALIGN_BACKWARD(heap_allowable_region_end, PAGE_SIZE), size,
-                error_code, executable);
-            /* ensure future heap allocations are reachable from this allocation */
-            if (p != NULL)
-                request_region_be_heap_reachable(p, size);
+            if (which & VMM_REACHABLE) {
```
derekbruening, line 1548: style: Please use `TEST(VMM_REACHABLE, which)` as the existing bit-checking code does

suite/tests/client-interface/cleancall.dll.c
```diff
@@ -600,6 +600,14 @@ static void
 app_exit_event(void)
 {
     check_stack_alignment();
+
+#if defined(LINUX) && defined(X86_64)
+    /* i#4335: Test allocation of more than 2.8GB in unreachable heap */
+    for (int i = 0; i != 50; ++i)
+    {
```
derekbruening, line 607: Looks like a clang-format style violation: should turn Travis red.

suite/tests/client-interface/cleancall.dll.c
```diff
@@ -600,6 +600,14 @@ static void
 app_exit_event(void)
 {
     check_stack_alignment();
+
+#if defined(LINUX) && defined(X86_64)
+    /* i#4335: Test allocation of more than 2.8GB in unreachable heap */
```
derekbruening, line 605: I would say this fits better in the alloc.dll.c test as it is unrelated to clean calls?

johnfxgalea:
Just checking the PR, looks good to me as well.  

However, I was not aware global_alloc can return unreachable heap memory? 

Look at the docs:
https://dynamorio.org/dynamorio_docs/dr__tools_8h.html#abb7e20e83f7b8e1b65428e45c8ab4211

derekbruening:
> Just checking the PR, looks good to me as well.
> 
> However, I was not aware dr_global_alloc can return unreachable heap memory?
> 
> Look at the docs:
> https://dynamorio.org/dynamorio_docs/dr__tools_8h.html#abb7e20e83f7b8e1b65428e45c8ab4211

dr_global_alloc uses reachable memory, but malloc does *not* use dr_global_alloc: it uses plain DR heap which is unreachable.

johnfxgalea:
Thanks for making that clear. I had a look again at the code, and confirm that it does not use dr_global_alloc.

PR #4354, opened 2020-07-02
Experimental: Revert to state where Travis was green for mac build

abhinav92003:
Experimental only: Reverting to a previous state where Travis was green, at commit 047a25f.

When I reran build on the existing Travis page for commit 047a25f, the build succeeded again: https://travis-ci.com/github/DynamoRIO/dynamorio/jobs/354686738. I want to check if it still succeeds on a new PR. This may help in figuring out what configuration is different then vs now.


abhinav92003:
Travis macOS builds are passing again, of their own accord. As shown here as well, all status are green now.

PR #4096, opened 2020-02-14
Try to diagnose Travis Mac failure that does not reproduce locally

derekbruening:

PR #4149, opened 2020-03-02
i#3917: Fix gcc 9.2.1 build errors

derekbruening:
Fixes a duplicate sigcontext define and adds a missing override in
drgui to address issues building with gcc 9.2.1

Fixes #3917

PR #4140, opened 2020-02-27
i#4134: DrBBDup: basic block duplication

johnfxgalea:
The first PR concerning DrBBDup. It includes the main functionality of the extension along with tests. The header file includes documentation.

Issue #4134

PR #4355, opened 2020-07-03
Update xcode version from the default 9.4 to 11.3

abhinav92003:
Currently the default xcode9.4 is being used for macOS builds because osx_image setting is not present in jobs section. This somehow wasn't an issue when it was initially added in #4096.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,6 @@ branches:
 # We use Xenial.
 sudo: required
 dist: xenial
-osx_image: xcode11.3
 
 language:
   - c
@@ -104,6 +103,7 @@ jobs:
     # 64-bit OSX build with clang and run tests:
     - if: type != cron AND env(TRAVIS_EVENT_TYPE) != cron
       os: osx
+      osx_image: xcode11.3
       # gcc on Travis claims to not be CMAKE_COMPILER_IS_GNUCC so we only run clang.
       compiler: clang
       env: DYNAMORIO_CROSS_AARCHXX_LINUX_ONLY=no DEPLOY=no EXTRA_ARGS=64_only


abhinav92003:
For more context, copying @derekbruening's comments from other PRs:

> I put in the 11.3 upgrade in February and it certainly took effect back then as you can see in PR #4096 https://travis-ci.com/github/DynamoRIO/dynamorio/jobs/287132339
> 
> ```
> InstalledDir: /Applications/Xcode-11.3.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
> ```
> 
> There were failures with the default, so you would think we'd notice if it regressed at some point back to 9.4.


> I checked a few builds and it looks like Travis made some change which caused the regression between our #4149 and #4150 on Mar 1-2:
> 
> * #4140 11.3  Mar 1
> * #4149 11.3  Mar 1
> * #4150 9.4   Mar 2

.travis.yml
```diff
@@ -104,6 +103,7 @@ jobs:
     # 64-bit OSX build with clang and run tests:
     - if: type != cron AND env(TRAVIS_EVENT_TYPE) != cron
       os: osx
+      osx_image: xcode11.3
```
derekbruening, line 106: This is weird that this would be required here, while `dist: xenial` at top level applies to all jobs.

abhinav92003: Going by the example at https://docs.travis-ci.com/user/multi-os/#example-multi-os-build-matrix, both `dist` and `osx_image` should be part of the `jobs` section. I see that `xenial` is also the default linux dist (https://docs.travis-ci.com/user/reference/linux/#default); maybe we're getting our expected build env because of that. I can verify this in a separate PR.

Verified that we're indeed using xcode11.3 now: https://travis-ci.com/github/DynamoRIO/dynamorio/jobs/357194591.

derekbruening: Not sure it's that clear: there are other examples which have a mix of global and jobs.include (and many examples with global plus jobs.exclude): https://docs.travis-ci.com/user/build-matrix/#explicitly-included-jobs-inherit-the-first-value-in-the-array

Mixing seems nicer: a global default with a few local overrides; avoid duplicating multiple settings in every single job.  I wonder if the xcode regression on Mar 2 was really global vs jobs.include or something else specific to osx.

`xenial` was not the default when the setting was put in place, of course.

derekbruening: This also brings up other settings: `sudo`, `language`, etc.  The example I cited above has language settings both globally and in the matrix.  It seems odd that the os ones would have to be per-job but not the language.

Issue #4356, opened 2020-07-03
Support forging OS signals on Linux


yury-khrustalev:
**Users list discussion**
Original discussion: https://groups.google.com/forum/#!topic/dynamorio-users/kHAyNgzvXuA.

**Description**
This feature request is about supporting raising a signal from code inserted by a clean call. This signal should appear as a signal
naturally raised by OS for the application or by the application itself. This should be correctly detected y debuggers (the point of fault should be found at the specified PC of the instrumented application code). This ticket is limited to scope of Linux / MacOS support.

**How it should look**

```C++
// API
void dr_forge_signal(app_pc pc, int sig);

// Usage
app_pc pc = ...;
int sig = SIGILL;
dr_forge_signal(pc, sig);
```

**Poor man's solution**
A dummy solution is to invoke `int raise(int sig);` from `signal.h`, however this is not "expected" by DynamoRIO core and is not handled properly (especially if DR is built in Debug mode), plus there is no way to deliver correct PC to a debugger.

**Existing functionality**
There already is function `os_forge_exception` which does something similar, but is not exposed via API.

PR #4357, opened 2020-07-03
i#4356 Linux: function to forge OS signal

yury-khrustalev:
Implements #4356.
This patch allows to emulate an OS signal with some number from a function invoked by a clean call. The debugger does not yet point to the intended PC, but functionally it works. I re-use existing internal function "os_forge_exception" and expose its part as a new API function "dr_forge_signal".

Please advise how to better test this new function.

### DIFF
--- a/core/lib/instrument_api.h
+++ b/core/lib/instrument_api.h
@@ -1676,6 +1676,16 @@ dr_standalone_exit(void);
  * UTILITY ROUTINES
  */
 
+#    ifdef UNIX
+DR_API
+/**
+ * Forge (emulate) OS signal with a given number at the specified PC.
+ * \note Only Unix.
+ */
+void
+dr_forge_signal(app_pc target_pc, int signal);
+#    endif
+
 #    ifdef WINDOWS
 /**
  * If \p x is false, displays a message about an assertion failure
--- a/core/lib/instrument_api.h
+++ b/core/lib/instrument_api.h
@@ -1707,13 +1707,14 @@ DR_API
 /**
  * Forge (emulate) OS signal with a given number at the specified PC.
  * \note Only Unix.
+ * \note Can be used only from a clean call.
  * \note This function is not supposed to return on success therefore
  * care should be taken to avoid any stale state or unreleased resources
- * remaining when the functions is called.
+ * remaining when the function is called.
  * \return If successful, this function does not return.
  */
 void
-dr_forge_signal(app_pc target_pc, int sig);
+dr_forge_signal(app_pc target_pc, int sig, dr_mcontext_t *mcontext);
 #    endif /* UNIX */
 
 DR_API
--- a/core/lib/instrument_api.h
+++ b/core/lib/instrument_api.h
@@ -1676,16 +1676,6 @@ dr_standalone_exit(void);
  * UTILITY ROUTINES
  */
 
-#    ifdef UNIX
-DR_API
-/**
- * Forge (emulate) OS signal with a given number at the specified PC.
- * \note Only Unix.
- */
-void
-dr_forge_signal(app_pc target_pc, int signal);
-#    endif
-
 #    ifdef WINDOWS
 /**
  * If \p x is false, displays a message about an assertion failure
@@ -1711,6 +1701,20 @@ dr_forge_signal(app_pc target_pc, int signal);
 #    define DR_ASSERT(x) DR_ASSERT_MSG(x, "")
 
 /* DR_API EXPORT END */
+
+#    ifdef UNIX
+DR_API
+/**
+ * Forge (emulate) OS signal with a given number at the specified PC.
+ * \note Only Unix.
+ * \note This function is not supposed to return on success therefore
+ * care should be taken to avoid any stale state or unreleased resources
+ * remaining when the functions is called.
+ * \return If successful, this function does not return.
+ */
+void
+dr_forge_signal(app_pc target_pc, int sig);
+#    endif /* UNIX */
 
 DR_API
 /** Returns true if all DynamoRIO caches are thread private. */
--- a/core/unix/signal.c
+++ b/core/unix/signal.c
@@ -6448,12 +6448,20 @@ os_forge_exception(app_pc target_pc, dr_exception_type_t type)
         sig = SIGSEGV;
         break;
     }
-    dr_forge_signal(target_pc, sig);
+    dcontext_t *dcontext = get_thread_private_dcontext();
+    priv_forge_signal(target_pc, sig, get_mcontext(dcontext));
     ASSERT_NOT_REACHED();
 }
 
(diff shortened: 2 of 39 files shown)


johnfxgalea:
Thanks for the PR!

Can you please add a test - maybe forge a signal and check it by registering a signal event please?

johnfxgalea:
Also, please add a reference to the issue in the commit message (as stated in the wiki). In other words, include "Fixes: #4356"

Thank you.

core/lib/instrument_api.h
```diff
@@ -1676,6 +1676,16 @@ dr_standalone_exit(void);
  * UTILITY ROUTINES
  */
 
+#    ifdef UNIX
+DR_API
+/**
+ * Forge (emulate) OS signal with a given number at the specified PC.
+ * \note Only Unix.
```
derekbruening, line 1683: Please document that this function does not return when successful (maybe return a bool like dr_redirect_execution?).  That means it cannot be called from random places where a lock might be held or something, which also needs to be documented: probably using the same limits as dr_redirect_execution() would be safe.

core/unix/signal.c
```diff
@@ -6457,6 +6448,21 @@ os_forge_exception(app_pc target_pc, dr_exception_type_t type)
         sig = SIGSEGV;
         break;
     }
+    dr_forge_signal(target_pc, sig);
+}
```
derekbruening, line 6452: Add `ASSERT_NOT_REACHED()`

core/unix/signal.c
```diff
@@ -6457,6 +6448,21 @@ os_forge_exception(app_pc target_pc, dr_exception_type_t type)
...
+
+void
+dr_forge_signal(app_pc target_pc, int sig) {
+    dcontext_t *dcontext = get_thread_private_dcontext();
```
derekbruening, line 6456: The code below uses the mcontext stored in the dcontext.  That is not the current application state when in a clean call (the state is on the dstack then).  Plus, the user may want to set more than just the pc.  I think the API should take in the `dr_mcontext_t` to use.

core/lib/instrument_api.h
yury-khrustalev: The documentation for `dr_redirect_execution` does not specify any particular limitations related to no return if the call is successful. I will describe that this function is not supposed to return and that care should be taken to avoid any stale state or unreleased resources.

core/unix/signal.c
yury-khrustalev: The code uses `get_mcontext(dcontext)` which returns `priv_mcontext_t`, not `dr_mcontext_t`. Also, we cannot support both `dcontext_t *dcontext` and `dr_mcontext_t* mc` (or `priv_mcontext_t* mc`) as input parameters since we would have to deal with all sorts of combinations of those two (most of which would probably be meaningless).
Did you mean that API should take in `dcontext_t *dcontext`?

core/unix/signal.c
yury-khrustalev: Done

yury-khrustalev:
Could someone please explain the build error in [this log](https://travis-ci.com/github/DynamoRIO/dynamorio/jobs/366985009)? The code compiles correctly when I try to use the function `dr_forge_signal`, but for `vps-debug-internal-32` it does not compile.

core/lib/instrument_api.h
derekbruening: The documentation for dr_redirect_execution says explicitly "if successful, does not return" (the last sentence in the docs: https://dynamorio.org/dynamorio_docs/dr__ir__utils_8h.html#a9f612cf9d0ca2025b23d30b6f5bb0231)

derekbruening: OK I missed the context (email just shows yours, and my initial comment above is from a month ago which I have forgotten): you're replying to limitations on when dr_redirect_execution can be called.  The docs explicitly say it can only be called from a clean call or a Windows exception event.

core/unix/signal.c
derekbruening: It should be like the other API routines: the exposed interface function should take in the public type dr_mcontext_t and convert it to the private type priv_mcontext_t before passing to an internal function.

derekbruening:
> Could someone please explain the build error in [this log](https://travis-ci.com/github/DynamoRIO/dynamorio/jobs/366985009)? The code compiles correctly when I try to use the function `dr_forge_signal`, but for `vps-debug-internal-32` it does not compile.

Didn't look at the log, but the vps build is not defining CLIENT_INTERFACE which is where most build issues with it come from.  #2971 will remove the define and the vps build but that's not done yet so for now ifdefs have to be used.

core/unix/signal.c
yury-khrustalev: Thank you for the explanation about `dr_mcontext_t` vs `priv_mcontext_t`. I can see how this should work. Originally, the `os_forge_exception` internal function obtained `dcontext_t` from `get_thread_private_dcontext()` and then used it to get `priv_mcontext_t`. If we expose `dr_mcontext_t` as a separate parameter, should we also expose `dcontext_t` as a separate one?

Addition of `dr_mcontext_t` to the new API function results in more changes to the existing internal function `os_forge_exception`. It becomes inconvenient to use the API function from here (we would need to call `dr_get_mcontext()` to obtain public mcontext which will later be converted into private mcontext).

To satisfy both requirements I should add another internal function which will take `priv_mcontext_t` and will be used from both `os_forge_exception` and `dr_forge_signal`.

yury-khrustalev:
Some builds have failed with DynamoRIO debug check failure:
```
340: <Application /home/travis/build/DynamoRIO/dynamorio/build_debug-internal-32/suite/tests/bin/api.forge_signal (23444).  Internal Error: DynamoRIO debug check failure: /home/travis/build/DynamoRIO/dynamorio/core/stats.c:359 ks->node[ks->depth - 1].var == pv && "stop not matching TOS"
```

Could you please help me with understanding what it means?

derekbruening:
> Some builds have failed with DynamoRIO debug check failure:
> 
> ```
> 340: <Application /home/travis/build/DynamoRIO/dynamorio/build_debug-internal-32/suite/tests/bin/api.forge_signal (23444).  Internal Error: DynamoRIO debug check failure: /home/travis/build/DynamoRIO/dynamorio/core/stats.c:359 ks->node[ks->depth - 1].var == pv && "stop not matching TOS"
> ```
> 
> Could you please help me with understanding what it means?

The `KSTART`..`KSTOP*` directives are mismatched: probably a `KSTOP` of some sort is missing from this API path.

yury-khrustalev:
I have to postpone this as the implementation may require much more time.

derekbruening:
OK.  It would be all right to leave it open until you're able to get back to it.

PR #4358, opened 2020-07-03
Experimental: Change top-level dist setting to bionic.

abhinav92003:
This is to verify whether setting dist at top-level has an effect at all. If it doesn't, we'll see Travis continue to use the default xenial build environment.
 In case of osx_image, the top-level setting had to be moved to the jobs section for it to really take effect.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,7 @@ branches:
 
 # We use Xenial.
 sudo: required
-dist: xenial
+dist: bionic
 
 language:
   - c


abhinav92003:
In https://github.com/DynamoRIO/dynamorio/runs/834429806, `dist` is set to `bionic`. Seems that the top-level setting indeed has an effect for `dist`, unlike `osx_image`.
```
  "dist": "bionic",
```

PR #4360, opened 2020-07-06
i#4318 xarch memtrace: Make alt lib dir take precedence

derekbruening:
Moves the -alt_module_dir for finding binaries during trace
post-processing to take precedence over the recorded path, to allow
avoiding problems when an identical-seeming path exists on the
processing machine (e.g., system libraries) on UNIX where we have no
checksum or other library distinguisher.

Tested manually on a case where a system library in the same path as
on the recorded machine caused failure.

Issue: #4318


derekbruening:
> `read_and_map_modules` is getting pretty complicated, it may be worth documenting the various strategies (and sub-strategies) as a function comment or refactoring them out as separate functions altogether.

Added a function comment.

Issue #4352, opened 2020-07-02
APP CRASH: (8.0 Master <custom_multithread_program) win32/syscall.c:thread_handle_to_pid() Thread Handle Leak on Windows


dkim-virsec:
**Describe the bug**

In running a hello-world type multi-threaded program, I saw that when running this application (source attached) under DR, thread handles are leaked when the source application uses CreateThread on Windows platforms.

**To Reproduce**

1. Compile the multithread-example.cpp shown here:
```
#include <windows.h>
#include <iostream>

using namespace std;

DWORD WINAPI thread1(LPVOID lpParameter)
{
    unsigned int& size = *((unsigned int*)lpParameter);
    for (int i = 0; i < 20000000 * (rand()%100); i++)
    {
        if ((i%100000000 == 0))
        {
            cout << "thread 1 " << i << endl;
        }
    }
    return 0;
}

DWORD WINAPI thread2(LPVOID lpParameter)
{
    unsigned int& size = *((unsigned int*)lpParameter);
    for (int i = 0; i < 2000000000; i++)
    {
        if ((i % 10000000 == 0))
        {
            cout << "thread 2 " << i << endl;
        }
    }
    return 0;
}

int main(int argc, char* argv[])
{
    unsigned int s1 = 10;
    unsigned int s2 = 20;
    DWORD myThreadID[2];
    
    HANDLE myHandle[2];

    for (int i = 0; i < 2; i++)
    {
        myHandle[i] = CreateThread(0, 0, thread1, &s1, 0, myThreadID+i);
    }
    //system("pause");
    for (int i = 0; i < 2; i++)
    {
        WaitForSingleObject(myHandle[i], INFINITE);
    }
    for (int i = 0; i < 2; i++)
    {
        cout << CloseHandle(myHandle[i]) << endl;
    }
    system("pause");
    return 0;
}

```
2. When the execution is paused after CloseHandle() is called, view the open handles through any debugging tool (I used ProcessExplorer64.exe -> view handles) when running this application natively vs under DR. 
3. When running under DR, there are extra thread handles that are not closed properly.
The cause is found in win32/syscall.c:thread_handle_to_pid():

**Versions**
 - What version of DynamoRIO are you using? 8.0 Master Build
 - Does the latest build from
https://github.com/DynamoRIO/dynamorio/wiki/Latest-Build solve the problem? No
[...]

PR #4361, opened 2020-07-06
i#4352: Thread handle Leak

dkim-virsec:
Fixes the thread handle leak when thread_handle_to_pid() in win32/syscall.c is called.

Fixes: #4352


derekbruening:
run arm tests

Issue #4326, opened 2020-06-22
ARM32 opcode_count returns always zero


bernhardklein:
Hello together,

thank you for your engagement in the project. I'm new to this project and it is exiting.
However I run into some trouble and maybe some of you can help me out.

When I run a command like:
`./bin32/drrun -c api/bin/libopcode_count.so -opcode 36 -- ls`
I always get:
`0/0 instructions executed`
executed on a arm device, a Raspberry Pi 3B or Pi4. I tried a lot of different opcodes it always looks the same. 
If I instead run it on a a x86 based machine I get plausible results.

Is the opcode_count sample not yet supported for arm devices? Or am I doing something wrong?

Thank you very much!

Best Regards Bernhard

AssadHashmi:
Hi @bernhardklein, which build version of DynamoRIO are you using?

As a sanity test, can you run the following and check that a non-zero value appears for the instruction count:
`./bin32/drrun -c api/bin/libinscount.so -- ls`

Can you run the following and paste the output? We can check whether opcode 36 (cmp) appears in the list at the end of the run:
`./bin32/drrun -c api/bin/libopcodes.so -opcode 36 -- ls`

bernhardklein:
Hi @AssadHashmi thank you for your super fast help :-)

At the moment I work with git commit 8551f131d97490203c2fedfbef2da9314a3ea45a. However, I got the same results with the prebuild DynamoRIO-ARM-Linux-EABIHF-8.0.1841 release.

`libinscount` seems to work, it reports: `613880 instructions executed`

the result of `./bin32/drrun -c api/bin/libopcodes.so -opcode 36 -- ls`:
`Client opcodes is running
Inconsistency detected by ld.so: do-rel.h: 116: elf_dynamic_do_Rel: Assertion map->l_info[VERSYMIDX (DT_VERSYM)] != NULL' failed!
Top 15 opcode execution counts in 32-bit ARM mode:
         10 : stmdb
         10 : uqsub8
         11 : lsl
         13 : rsb
         15 : clz
         15 : mvn
         16 : bl
         50 : sub
         55 : subs
         69 : orr
         77 : str
        132 : ldrb
        133 : ldr
        159 : add
        166 : mov
        312 : b
        463 : cmp`

Yes cmp appears, so I would assume that it should work with `opcode_count` as well.

Thank you!

AssadHashmi:
[...]

PR #4367, opened 2020-07-13
i#4326: Bug fix for global counter

johnfxgalea:
Fixes a bug that was resulting in an incorrect global instr count.

The PR also replaces the sample's existing sanity test with a more comprehensive regression test. 

Issue: #4326 #4372


api/samples/opcode_count.cpp
```diff
@@ -101,6 +101,15 @@ event_opcode_instruction(void *drcontext, void *tag, instrlist_t *bb, instr_t *i
     return DR_EMIT_DEFAULT;
 }
 
+static dr_emit_flags_t
+event_bb_analysis(void *drcontext, void *tag, instrlist_t *bb, bool for_trace,
+                  bool translating, OUT void **user_data)
+{
+    intptr_t bb_size = (intptr_t)drx_instrlist_app_size(bb);
```
derekbruening, line 108: Hmm, `ptr_int_t` is what DR defines and uses.  Looks like there are existing uses of `intptr_t`...mostly from you :)  I guess it's the more standard name externally in stdint.h so for a sample it seems reasonable.

johnfxgalea: Would it be worthwhile to do the switch to intptr_t? Don't really have a strong opinion tbh.

johnfxgalea: ^ I mean throughout the DR code base - of course via a separate issue.

johnfxgalea: ^ I mean throughout the DR code base - of course via a separate issue.

johnfxgalea:
> Should there be a test that checks actual output instead of just a crash sanity test?

The problem is that ensuring stable and exact execution counts are tricky. hmm

derekbruening:
> > Should there be a test that checks actual output instead of just a crash sanity test?
> 
> The problem is that the process of ensuring stable and exact execution counts is tricky. hmm

A pure-asm UNIX test has precise count control.  None of the other samples have such tests though so it's not just a problem of this sample.

api/samples/opcode_count.cpp
derekbruening: No I would leave core DR with its own types.  IIRC stdint.h was *not* part of Visual Studio until rather recently which was part of the reason to have our own types.

derekbruening:
> > > Should there be a test that checks actual output instead of just a crash sanity test?
> > 
> > 
> > The problem is that the process of ensuring stable and exact execution counts is tricky. hmm
> 
> A pure-asm UNIX test has precise count control. None of the other samples have such tests though so it's not just a problem of this sample.

The other way to have precise counts is to have an attach-detach test so the instrumentation runs on just a controlled window which could be pure asm if necessary even on Windows.

johnfxgalea:
@derekbruening Okay, with a pure-asm test that should work. I'll try to get this in when I have some free time.

johnfxgalea:
> > > > Should there be a test that checks actual output instead of just a crash sanity test?
> > > 
> > > 
> > > The problem is that the process of ensuring stable and exact execution counts is tricky. hmm
> > 
> > 
> > A pure-asm UNIX test has precise count control. None of the other samples have such tests though so it's not just a problem of this sample.
> 
> The other way to have precise counts is to have an attach-detach test so the instrumentation runs on just a controlled window which could be pure asm if necessary even on Windows.

@derekbruening I gave it a shot. LMK what you think.

johnfxgalea:
We need to decide whether we are going to keep the basic sanity checks. We could also instead compile two versions of the sample, one that shows results (required for the so called comprehensive test) and another that does not show results (for  the sanity test).

derekbruening:
> We need to decide whether we are going to keep the basic sanity checks. We could also instead compile two versions of the sample, one that shows results (required for the so called comprehensive test) and another that does not show results (for the sanity test).

IIRC the result showing on Windows pops up a msgbox, which is not very conducive to automated testing, so there we would probably want a way to switch to console messages.

johnfxgalea:
> > We need to decide whether we are going to keep the basic sanity checks. We could also instead compile two versions of the sample, one that shows results (required for the so called comprehensive test) and another that does not show results (for the sanity test).
> 
> IIRC the result showing on Windows pops up a msgbox, which is not very conducive to automated testing, so there we would probably want a way to switch to console messages.

What about Linux please?

derekbruening:
> > > We need to decide whether we are going to keep the basic sanity checks. We could also instead compile two versions of the sample, one that shows results (required for the so called comprehensive test) and another that does not show results (for the sanity test).
> > 
> > 
> > IIRC the result showing on Windows pops up a msgbox, which is not very conducive to automated testing, so there we would probably want a way to switch to console messages.
> 
> What about Linux please?

I would vote for removing the sanity check if the functionality test is a superset (even though more tests is generally good, fast CI is also good, and if it's not adding anything I would remove it).

api/samples/CMakeLists.txt
```diff
@@ -150,13 +150,20 @@ if (OPTIMIZE_SAMPLES)
...
   add_library(${name} SHARED ${source_file_list})
   configure_DynamoRIO_client(${name})
   foreach (ext ${extension_list})
     use_DynamoRIO_extension(${name} ${ext})
   endforeach (ext)
 
+  # Some samples actually have more tests than just simple sanity tests.
+  # The show_results is a passed flag that indicates that we should override
+  # SHOW_RESULTS_DEFAULT (if is turned off) for this sample.
+  if (NOT SHOW_RESULTS_DEFAULT AND show_results)
```
derekbruening, line 163: This should look at SHOW_RESULTS, not SHOW_RESULTS_DEFAULT, right?  I.e., at the actual value.  Is the goal of checking that just to avoid a duplicate -DSHOW_RESULTS?

api/samples/CMakeLists.txt
```diff
@@ -204,54 +211,54 @@ configure_DynamoRIO_global(OFF ON)
...
-add_sample_client(opcode_count    "opcode_count.cpp"    "drmgr;drreg;drx;droption")
+add_sample_client(bbbuf       "bbbuf.c"         "drmgr;drreg;drx" OFF)
+add_sample_client(bbcount     "bbcount.c"       "drmgr;drreg;drx" OFF)
+add_sample_client(bbsize      "bbsize.c"        "drmgr;drx" OFF)
+add_sample_client(div         "div.c"           "drmgr" OFF)
+add_sample_client(empty       "empty.c"         "" OFF)
+add_sample_client(memtrace_simple "memtrace_simple.c;utils.c" "drmgr;drreg;drutil;drx" OFF)
+add_sample_client(memval_simple   "memval_simple.c;utils.c"   "drmgr;drreg;drutil;drx" OFF)
+add_sample_client(instrace_simple "instrace_simple.c;utils.c" "drmgr;drreg;drx" OFF)
+add_sample_client(opcode_count    "opcode_count.cpp"    "drmgr;drreg;drx;droption" ON)
```
derekbruening, line 222: Hmm, multiple thoughts here:

1. This ignores the user's SHOW_RESULTS option which seems undesirable -- although that option should maybe rarely be used and nobody should rely on it (should we mark it ADVANCED?) (except Windows where annoying popups could make it a reasonable user option -- see next tiem).  I guess I'm just complaining and not suggesting a solution: label SHOW_RESULTS and not affecting all samples?  And mark it ADVANCED and make it not really part of the user-facing choices or sthg.

2. This will cause popups on Windows which can hang automated tests.  Maybe long-term what we want is to get rid of SHOW_OPTIONS, have all samples print to the console, add functionality-testing tests for all samples, and have a USE_POPUPS or sthg option for Windows where console printing doesn't work for GUI apps and pre-Win8 where there are all kinds of console printing problems.

suite/tests/CMakeLists.txt
```diff
@@ -3011,8 +3011,18 @@ if (CLIENT_INTERFACE)
                 client-interface/memval-test.c "" "" "")
           endif(X86 AND (NOT X64) AND UNIX)
         elseif (sample STREQUAL "opcode_count")
-          torunonly_ci(sample.${sample} common.${control_flags} ${sample}
-            common/${control_flags}.c "-opcode 5" "" "")
+          # We do not do a simple sanity test for opcode_count, but a normal test
+          # that checks the sample's output.
+          # FIXME i#4372: The test is currently only supported on X86 32-bit.
+          # The test program is implemented in pure asm (for UNIX).
```
derekbruening, line 3017: So should we do the sanity test on other platforms then?

suite/tests/samples/opcode_count-test.asm
```diff
@@ -0,0 +1,39 @@
...
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+.global _start
+
+_start:
```
derekbruening, line 35: Hmm: we copy the samples into an exported dir for install packages.  Please ensure that copy skips this test file.

api/samples/CMakeLists.txt
johnfxgalea: Actually, what we want is that we always show results, when building tests. Therefore, I should test BUILD_TESTS AND show_results. I just used SHOW_RESULTS_DEFAULT as a proxy which perhaps caused confusion.

suite/tests/samples/opcode_count-test.asm
johnfxgalea: This test is in the test folder, does DR export copy tests? I'll do an install and check if it is there nonetheless.

api/samples/CMakeLists.txt
johnfxgalea: Yes, it does ignore the SHOW_RESULTS option when building tests. This seems reasonable no? Otherwise we cannot set the up, or do you mean if SHOW_RESULTS is false, we switch back to just sanity tests?

suite/tests/samples/opcode_count-test.asm
johnfxgalea: I did a search for the test in the install director (_find . -name \*opcode\*_)  and did not find anything.

johnfxgalea:
Yay, all green! @derekbruening please take a look when you get a chance. Sorry for the delay; it was a busy week.

suite/tests/samples/opcode_count-test.asm
derekbruening: Huh, I think I misread the path and assumed it was api/samples/ (since there is no suite/tests/samples).  Resolving.

api/samples/CMakeLists.txt
derekbruening: I guess I'm seeing downsides in:
+ The extra add_sample_client param, which is made public to the user, right?  This is our sample CMakeLists.txt for users.
+ Having SHOW_RESULTS overridden in a potentially confusing way: just some samples, and then just UNIX, and not lining up with which samples have tests: this test is 32-bit only, but this sample will print results in 64-bit.

Here is my proposal:
+ Create two build executable targets for each sample: one public-facing with SHOW_RESULTS on, and another that's for testing purposes.  Then we also avoid samples being silent in a build dir with BUILD_TESTS on.  As for tests printing or not: I guess as we add functional tests and remove sanity tests (is that what we decided?) we flip them to print vs not print.  Long-term the public ones and testing ones are identical on UNIX probably, and on Windows it's popup vs console printing.

johnfxgalea: Okay, I agree, will make the code simpler with fewer branches. So in this case, we need to ensure that the second build is indeed not exported to release packages.

johnfxgalea: >  Long-term the public ones and testing ones are identical on UNIX probably

Do you mean eventually we will move away from the SHOW_RESULTS option and always print the results?

derekbruening: > Do you mean eventually we will move away from the SHOW_RESULTS option and always print the results?

If we have a separate build target for tests, we could get rid of the SHOW_RESULTS CMake option, yes.  The only reason I could see to have an option would be for the Windows popups which can be annoying: but these are just samples and people are supposed to edit them so maybe no reason to go adding runtime or build options for that.

johnfxgalea:
@derekbruening Done, PTAL.

api/samples/CMakeLists.txt
```diff
@@ -172,6 +172,20 @@ function (add_sample_client name source_file_list extension_list)
     "${sample_list};${name}")                                     # NON-PUBLIC
   set(srcs ${srcs} ${source_file_list} PARENT_SCOPE)              # NON-PUBLIC
   set(tgts ${tgts} ${name} PARENT_SCOPE)                          # NON-PUBLIC
+
+  # Some samples actually have more tests than just simple sanity tests. These
```
derekbruening, line 176: As mentioned before, this should all be hidden, so `# NON-PUBLIC`

api/samples/CMakeLists.txt
```diff
@@ -204,54 +214,61 @@ configure_DynamoRIO_global(OFF ON)
...
-add_sample_client(bbbuf       "bbbuf.c"         "drmgr;drreg;drx")
-add_sample_client(bbcount     "bbcount.c"       "drmgr;drreg;drx")
-add_sample_client(bbsize      "bbsize.c"        "drmgr;drx")
-add_sample_client(div         "div.c"           "drmgr")
-add_sample_client(empty       "empty.c"         "")
-add_sample_client(memtrace_simple "memtrace_simple.c;utils.c" "drmgr;drreg;drutil;drx")
-add_sample_client(memval_simple   "memval_simple.c;utils.c"   "drmgr;drreg;drutil;drx")
-add_sample_client(instrace_simple "instrace_simple.c;utils.c" "drmgr;drreg;drx")
-add_sample_client(opcode_count    "opcode_count.cpp"    "drmgr;drreg;drx;droption")
+# FIXME i#4326: Add additional tests for samples that go beyond sanity checks.
```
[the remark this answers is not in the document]
derekbruening: `# NON-PUBLIC`

api/samples/CMakeLists.txt
```diff
@@ -212,7 +227,13 @@ add_sample_client(empty       "empty.c"         "")
 add_sample_client(memtrace_simple "memtrace_simple.c;utils.c" "drmgr;drreg;drutil;drx")
 add_sample_client(memval_simple   "memval_simple.c;utils.c"   "drmgr;drreg;drutil;drx")
 add_sample_client(instrace_simple "instrace_simple.c;utils.c" "drmgr;drreg;drx")
-add_sample_client(opcode_count    "opcode_count.cpp"    "drmgr;drreg;drx;droption")
+# FIXME i#4326: We currently only support additional test for the opcode_count sample on
+# X86 32-bit UNIX. We should port to other archs.
+if (X86 AND (NOT X64) AND UNIX)
+  add_sample_client(opcode_count    "opcode_count.cpp"    "drmgr;drreg;drx;droption")
```
derekbruening, line 233: These are now identical and this if() would go away?

api/samples/CMakeLists.txt
```diff
@@ -172,6 +172,20 @@ function (add_sample_client name source_file_list extension_list)
...
+  # tests require that results are shown. In order to avoid short-circuiting
+  # the SHOW_RESULTS option, we build an additional sample (which shows results)
+  # for testing purposes.
+  if (BUILD_TESTS)
+    add_library(${name}_test SHARED ${source_file_list})
+    configure_DynamoRIO_client(${name}_test)
+    foreach (ext ${extension_list})
+      use_DynamoRIO_extension(${name}_test ${ext})
+    endforeach (ext)
+    target_compile_definitions(${name}_test PRIVATE -DSHOW_RESULTS)
```
derekbruening, line 186: But today we want the _test version to *not* have SHOW_RESULTS for our sanity tests of running an app and having zero output (long-term yes we want output for _test, but console and not popup for Windows, and functional tests for each sample, but we're not there yet).

How about this:
+ The _test target does *not* have SHOW_RESULTS: i.e., it's silent
+ Today's sanity tests of these samples use the _test target
+ Your new opcode_count functional test uses the regular (non _test) target and is only on UNIX?
+ Add comments (file issue?) about removing the SHOW_RESULTS CMake var (we only need a cpp define), converting the others to use functional tests, and changing _test to replace the Windows popup with console printing

derekbruening: Wait so your version still has BUILD_TESTS turning off SHOW_RESULTS so it's reversed from mine?

api/samples/CMakeLists.txt
johnfxgalea: Do you mean, I start the comment with NON-PUBLIC?

derekbruening: Not sure what the script stripping these looks for precisely: whatever works for it.  See how the lines above have that marker?  Some CMake code strips such lines out when coping this for export (yeah kind of hacky).

derekbruening: I would think the same suffix would look best and should match the script.

api/samples/CMakeLists.txt
johnfxgalea: Yes, it is reversed.

derekbruening: I would vote for reversing what this PR has: it is more work (remove the disabling or whatnot and tweak the sanity tests) but it would be nice to have the regular sample show results in a build dir w/ tests (I've been annoyed by that before).

johnfxgalea: It has to be reversed. If we set the main exported sample to always show results when building tests, we will ignore the user's SHOW_RESULTS option.

derekbruening: Or if you want to put that under an issue which also covers removing SHOW_RESULTS and longer-term stuff that seems ok.

suite/tests/CMakeLists.txt
johnfxgalea: Done.

api/samples/CMakeLists.txt
johnfxgalea: Done.

api/samples/CMakeLists.txt
johnfxgalea: Done.

api/samples/CMakeLists.txt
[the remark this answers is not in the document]
johnfxgalea: Done.

api/samples/CMakeLists.txt
```diff
@@ -150,13 +150,23 @@ if (OPTIMIZE_SAMPLES)
...
     use_DynamoRIO_extension(${name} ${ext})
   endforeach (ext)
 
+  # Some samples actually have more tests than just simple sanity tests.
+  # The show_results is a passed flag that indicates that we should override
+  # SHOW_RESULTS and always print results. Otherwise, if not result is output,
+  # we cannot run the tests which go beyond simple sanity checks.
+  # Moreover, we do not short-circuit SHOW_RESULTS for Windows, because pop-ups
+  # will hang our automated tests.
+  if (BUILD_TESTS AND UNIX AND show_results)
```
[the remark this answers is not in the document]
johnfxgalea: Done (now irrelevant given the code changes for the following review).

api/samples/CMakeLists.txt
derekbruening: > we will ignore the user's SHOW_RESULTS option.

I want to get rid of that option -- but sure, a separate issue + PR.

I think this needs a comment (ideally w/ issue ref) that just enabling SHOW_RESULTS is not enough for testing: we have to change the Windows popup to be a console print.

johnfxgalea: Okay, I filed https://github.com/DynamoRIO/dynamorio/issues/4392

api/samples/CMakeLists.txt
johnfxgalea: Done.

johnfxgalea:
@derekbruening PTAL. Let me know if there are other major issues. I'm not sure, but do I have to make # NON-PUBLIC for the building of the duplicated tests as well? I take it we do not want to make our CMAKE code for building tests public no? Should I revert?

derekbruening:
> @derekbruening PTAL. Let me know if there are other major issues. I'm not sure, but do I have to make # NON-PUBLIC for the building of the duplicated tests as well? I take it we do not want to make our CMAKE code for building tests public no? 

The public CMakeLists.txt for the samples is what we present to users as a recommended build file, so we'd like to keep it simple, is the idea.

johnfxgalea:
> > @derekbruening PTAL. Let me know if there are other major issues. I'm not sure, but do I have to make # NON-PUBLIC for the building of the duplicated tests as well? I take it we do not want to make our CMAKE code for building tests public no?
> 
> The public CMakeLists.txt for the samples is what we present to users as a recommended build file, so we'd like to keep it simple, is the idea.

Okay, I don't think a user needs to know about our tests. I think this is good to merge of okay with you?

api/samples/CMakeLists.txt
```diff
@@ -172,6 +172,24 @@ function (add_sample_client name source_file_list extension_list)
     "${sample_list};${name}")                                     # NON-PUBLIC
   set(srcs ${srcs} ${source_file_list} PARENT_SCOPE)              # NON-PUBLIC
   set(tgts ${tgts} ${name} PARENT_SCOPE)                          # NON-PUBLIC
+
+  # Some samples actually have more tests than just simple sanity # NON-PUBLIC
+  # tests. These tests require that results are shown. In order   # NON-PUBLIC
+  # to avoid short-circuiting the SHOW_RESULTS option, we build   # NON-PUBLIC
+  # an additional sample (which shows results) for testing        # NON-PUBLIC
+  # purposes.                                                     # NON-PUBLIC
+  # FIXME i#4372 i#4392: Ultimately, when we remove all sanity    # NON-PUBLIC
```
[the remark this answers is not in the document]
johnfxgalea: I think separation is okay; one focuses on adding tests and the other concerns removing SHOW_RESULTS.

PR #4368, opened 2020-07-13
i#4318 xarch memtrace: Detect view arch mismatch earlier

derekbruening:
For the view tool, the original wrong-arch check was placed *after*
the entry skipping for -skip_refs, which can result in never checking
the architecture.  We fix that here by moving the check up front.

Issue: #4318


derekbruening:
Thanks for the review!

PR #4371, opened 2020-07-16
Fix compilation on old Linux which does not have EM_ARM/EM_AARCH64 symbols

SweetVishnya:
Hi!

I tried to build DynamoRIO on CentOS 6.9. I failed because old Linux distribution does not have EM_AARCH64 symbol which leads to compilation errrors. This merge request fixes compilation on CentOS 6.9 and maybe other old distros.

### DIFF
--- a/core/unix/module_elf.h
+++ b/core/unix/module_elf.h
@@ -193,7 +193,7 @@
 #    endif                 /* ANDROID */
 #endif                     /* X86/ARM */
 
-/* Define ARM ELF machine types to support compiling on old Linux distors. */
+/* Define ARM ELF machine types to support compiling on old Linux distros. */
 #ifndef EM_ARM
 #    define EM_ARM 40
 #endif
--- a/core/unix/module_elf.h
+++ b/core/unix/module_elf.h
@@ -195,10 +195,10 @@
 
 /* Define ARM ELF machine types to support compiling on old Linux distors. */
 #ifndef EM_ARM
-#define EM_ARM 40
+#   define EM_ARM 40
 #endif
 #ifndef EM_AARCH64
-#define EM_AARCH64 183
+#   define EM_AARCH64 183
 #endif
 
 bool
--- a/core/unix/module_elf.h
+++ b/core/unix/module_elf.h
@@ -193,6 +193,14 @@
 #    endif                 /* ANDROID */
 #endif                     /* X86/ARM */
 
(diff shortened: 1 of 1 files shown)


core/drlibc/drlibc_module_elf.c
```diff
@@ -115,9 +115,9 @@ is_elf_so_header_common(app_pc base, size_t size, bool memory)
             (memory && elf_header.e_ehsize != sizeof(ELF_HEADER_TYPE)) ||
             (memory &&
 #    ifdef X64
-             elf_header.e_machine != EM_X86_64 && elf_header.e_machine != EM_AARCH64
+             elf_header.e_machine != IF_AARCHXX_ELSE(EM_AARCH64, EM_X86_64)
```
derekbruening, line 118: This breaks cross-arch support where we want to map AArch64 libraries on an x86 machine (basically this undoes recent changes to this code).  Probably best to instead add a define for `EM_AARCH64` if it's not already defined to module_elf.h.

SweetVishnya: Maybe should check `#ifdef EM_AARCH64` and add code if defined?

SweetVishnya: It is not defined in `module_elf.h`. We can defined it there if it is a known constant.

SweetVishnya: So, what would be the best choice?

derekbruening: I think you must have missed the 2nd sentence of my comment?

SweetVishnya: Sorry, I will add a define to module_elf.h.

SweetVishnya: Done.

core/unix/module_elf.h
```diff
@@ -193,6 +193,14 @@
 #    endif                 /* ANDROID */
 #endif                     /* X86/ARM */
 
+/* Define ARM ELF machine types to support compiling on old Linux distors. */
```
derekbruening, line 196: nit: spelling: "distros"

derekbruening:
Thank you for the patch.  Just a spelling nit; otherwise LGTM.

SweetVishnya:
Fixed spelling.

AssadHashmi:
run arm tests

derekbruening:
Thanks again for the patch.

Issue #2390, opened 2017-04-27
Reduce overhead of indirect branch on AArch64


egrimley:
It is well known that indirect branches are a major factor in the performance of dynamic binary translation, optimisation and instrumentation (see, for example, "Optimizing Indirect Branches in Dynamic Binary Translators", 2016), although in the case of dynamic binary instrumentation the overhead of instrumentation may dominate everything else.

DynamoRIO's handling of indirect branches on AArch64 is currently rather inefficient; we have so far not paid much attention to performance. This issue documents the current implementation of indirect branches and lists a few improvements that may be worth implementing. Some of these things may also be applicable to other architectures.

The sequence of instructions executed under DynamoRIO to simulate a single RET instruction in the original application typically looks like this:
```
   0x4c7ea954:  str     x2, [x28,#16]           # Save X2 to TLS.
   0x4c7ea958:  mov     x2, x30                 # Move target address to X2.
   0x4c7ea95c:  b       0x4c7ea960              # Gratuitous branch to following instr.
   0x4c7ea960:  stp     x0, x1, [x28]           # Save X0 and X1 to TLS.
   0x4c7ea964:  mov     x0, #0xae08             # Load value to identify this location;
   0x4c7ea968:  movk    x0, #0x4c81, lsl #16    # in this case we could load the value
   0x4c7ea96c:  movk    x0, #0x0, lsl #32       # with just 2 instrs but we always use 4.
   0x4c7ea970:  movk    x0, #0x0, lsl #48       #
   0x4c7ea974:  ldr     x1, [x28,#120]          # We use LDR+BR to get to look-up code
   0x4c7ea978:  br      x1                      # even when it is in range of direct B.

   0x4c521400:  str     x0, [x28,#24]           # Save X0 to TLS ... again!
   0x4c521404:  ldp     x1, x0, [x28,#168]      # Load hash mask and base from TLS. The
   0x4c521408:  and     x1, x1, x2              # hash_mask is 0x3f: could use immediate?
   0x4c52140c:  add     x1, x0, x1, lsl #4      # Shift depends on ibl_hash_func_offset.
   0x4c521410:  ldr     x0, [x1]                # Load app addr from hash table; use LDP?

   0x4c521414:  cbz     x0, 0x4c521440          # Why check for zero before match?
   0x4c521418:  sub     x0, x0, x2              # Could use EOR here.
   0x4c52141c:  cbnz    x0, 0x4c521438          # Check keys match: no, so branch.

[...]

Issue #4273, opened 2020-04-21
ASSERT_NOT_IMPLEMENTED aarch64/emit_utils.c:190


derekbruening:
Running debug drcachesim offline on a proprietary app:
```
$ bin64/drrun -debug -msgbox_mask 12 -stderr_mask 15 -t drcachesim -dr_debug -offline -max_trace_size 10M -- <app and args>
...
<Application xxx (nnn).  Internal Error: DynamoRIO debug check failure: Not implemented @src/core/arch/aarch64/emit_utils.c:190 (0)
(Error occurred @152412 frags)
```

```
#5  0x0000fffec740ee90 in d_r_internal_error (
    file=0xfffec76fa9c0 "Not implemented @src/core/arch/aarch64/emit_utils.c", line=190, expr=0xfffec76fa9b8 "(0)")
    at src/core/utils.c:176
#6  0x0000fffec7557c2c in patch_stub (f=0xfffdfa86ea98, stub_pc=0xfffe7fc17214 "\200\a", target_pc=0xfffe77c16cec "\201\233", hot_patch=true)
    at src/core/arch/aarch64/emit_utils.c:190
#7  0x0000fffec754c2a0 in link_direct_exit (dcontext=0xfffdf78d8660, f=0xfffdfa86ea98, l=0xfffdfa86ead8, targetf=0xfffdf7ae6ec8, hot_patch=true)
    at src/core/arch/emit_utils_shared.c:466
#8  0x0000fffec73e32a0 in link_branch (dcontext=0xfffdf78d8660, f=0xfffdfa86ea98, l=0xfffdfa86ead8, targetf=0xfffdf7ae6ec8, hot_patch=true)
    at src/core/link.c:1166
#9  0x0000fffec73e8638 in link_fragment_outgoing (dcontext=0xfffdf78d8660, f=0xfffdfa86ea98, new_fragment=true)
    at src/core/link.c:1798
#10 0x0000fffec73e9f48 in link_new_fragment (dcontext=0xfffdf78d8660, f=0xfffdfa86ea98) at src/core/link.c:1980
#11 0x0000fffec740d1ac in emit_fragment_common (dcontext=0xfffdf78d8660, tag=0xfffec1b6cc20 "\375{\276\251\364O\001\251\364\003", ilist=0xfffdf7990ae8, 
    flags=16777216, vmlist=0xfffdf78644b8, link_fragment=true, add_to_htable=true, replace_fragment=0x0)
[...]

PR #4375, opened 2020-07-18

merckhung:
model (+7%).

Issue: #2390


core/arch/aarch64/emit_utils.c
```diff
@@ -717,29 +716,33 @@ emit_indirect_branch_lookup(dcontext_t *dc, generated_code_t *code, byte *pc,
...
+    APP(&ilist,
+        INSTR_CREATE_and(dc, opnd_create_reg(DR_REG_X1), opnd_create_reg(DR_REG_X1),
+                         opnd_create_reg(DR_REG_X2)));
+    /* dmb ishld to prevent memory-access instructions from being reordered across the
+     * barrier. DMB results generally in better performances compared to LDAR in the case.
+     */
     APP(&ilist,
-        INSTR_CREATE_ldar(dc, opnd_create_reg(DR_REG_X1),
-                          OPND_CREATE_MEMPTR(DR_REG_X1, 0)));
+        instr_create_0dst_1src(dc, OP_dmb, OPND_CREATE_INT(/* DR_DMB_ISHLD */ 9)));
```
derekbruening, line 739: We'd prefer to use INSTR_CREATE_macros: ARM and x86 are complete but it seems AArch64 is missing some.  Could you add the ones you need?  IIRC they are manually done instead of automated like the ARM ones are.

derekbruening, line 739: Could we use `DR_DMB_ISHLD` directly?

core/arch/aarch64/emit_utils.c
```diff
@@ -750,11 +753,13 @@ emit_indirect_branch_lookup(dcontext_t *dc, generated_code_t *code, byte *pc,
     /* x1 now holds the fragment_entry_t* in the hashtable. */
     APP(&ilist, load_tag);
     /* Load tag from fragment_entry_t* in the hashtable to x0. */
-    /* ldr x0, [x1, #tag_fragment_offset] */
-    APP(&ilist,
-        INSTR_CREATE_ldr(
-            dc, opnd_create_reg(DR_REG_X0),
-            OPND_CREATE_MEMPTR(DR_REG_X1, offsetof(fragment_entry_t, tag_fragment))));
+    /* ldr x0, [x1], #tag_fragment_offset */
+    instr_t *instr = instr_create_2dst_3src(
```
derekbruening, line 757: We'd again prefer an INSTR_CREATE_.

core/arch/aarch64/emit_utils.c
```diff
@@ -717,29 +716,33 @@ emit_indirect_branch_lookup(dcontext_t *dc, generated_code_t *code, byte *pc,
...
      */
-
     /* Spill x0. */
-    APP(&ilist, instr_create_save_to_tls(dc, DR_REG_R0, TLS_REG3_SLOT));
-    /* Load-acquire hash mask.  We need a load-acquire to ensure we see updates
-     * properly; the corresponding store-release is in update_lookuptable_tls().
+    APP(&ilist, instr_create_save_to_tls(dc, DR_REG_X0, TLS_REG3_SLOT));
+    /* Load hash mask before loading hash_table, with DMB to prevent memory-access
+     * instructions from being reordered. A corresponding store-release is in
+     * update_lookuptable_tls() to ensure a new hast_table will be written after a
```
derekbruening, line 723: typo: hash_table

derekbruening:
> I took a quick look. Probably best for an AArch64 expert and/or those who wrote the original code to take a deeper look.

Requesting @egrimley to take a look (or delegate)

AssadHashmi:
run arm tests

AssadHashmi:
run arm tests

core/arch/aarch64/emit_utils.c
merckhung: Got it. I have created INSTR_CREATE_dmb macro for AArch64 in the 2nd commit.

merckhung: Sure thing. I have made a copy in AArch64's instr_create.h instead of including ARM's header, to avoid macro definition conflicts across architectures.

core/arch/aarch64/emit_utils.c
merckhung: I see. INSTR_CREATE_ldr_imm has been created to serve this purpose.

core/arch/aarch64/emit_utils.c
merckhung: Corrected. Thank you!

derekbruening:
@AssadHashmi -- trying to get a review from an ARM expert

AssadHashmi:
> @AssadHashmi -- trying to get a review from an ARM expert

Based on my (limited) experience the change looks functionally correct.I can't comment on performance

core/ir/aarch64/instr_create.h
```diff
@@ -2514,5 +2520,28 @@
...
     instr_create_1dst_3src(dc, OP_umull2, Rd, Rm, Rn, width)
 
+/**
+ * Creates a DMB instruction.
+ * \param dc      The void * dcontext used to allocate memory for the instr_t.
+ * \param imm     The integer constant opnd_t operand.
+ */
+#define INSTR_CREATE_dmb(dc, imm) instr_create_0dst_1src((dc), OP_dmb, (imm))
+
+/** Immediate values for INSTR_CREATE_dmb(). */
```
derekbruening, line 2530: This looks like an exact copy of the enum in arm/.  To avoid duplicating code, please instead move it to a shared header.  Maybe instr_create_shared.h, or maybe a new ir/aarchxx/instr_create_aarchxx.h, or possibly instr.h or opnd.h.

core/ir/aarch64/instr_create.h
```diff
@@ -2514,5 +2520,28 @@
 #define INSTR_CREATE_umull2_vector(dc, Rd, Rm, Rn, width) \
     instr_create_1dst_3src(dc, OP_umull2, Rd, Rm, Rn, width)
 
+/**
+ * Creates a DMB instruction.
+ * \param dc      The void * dcontext used to allocate memory for the instr_t.
+ * \param imm     The integer constant opnd_t operand.
```
derekbruening, line 2526: Point at one of the DR_DMB_OSHLD values, using `#` to make a doxygen link, so the user knows there are named constants.

core/ir/opnd.h
```diff
@@ -1702,7 +1702,7 @@ DR_API
 opnd_t
 opnd_create_far_base_disp_ex(reg_id_t seg, reg_id_t base_reg, reg_id_t index_reg,
                              int scale, int disp, opnd_size_t size, bool encode_zero_disp,
-                             bool force_full_disp, bool disp_short_addr);
+                             bool force_full_disp, bool disp_short_addr, bool post_index);
```
derekbruening, line 1705: This is a breaking interface change.  This needs to be accompanied by documentation changes announcing the break.  Alternatively, instead of changing this function, a separate opnd_set_zero_offset_pre_index() could be added, which would be better for compatibility.

derekbruening: I guess opnd_set_zero_offset_post_index() since the default is pre.

core/ir/aarch64/instr_create.h
```diff
@@ -539,6 +539,12 @@
 #define INSTR_CREATE_ldp(dc, rt1, rt2, mem) \
     instr_create_2dst_1src(dc, OP_ldp, rt1, rt2, mem)
 #define INSTR_CREATE_ldr(dc, Rd, mem) instr_create_1dst_1src((dc), OP_ldr, (Rd), (mem))
+#define INSTR_CREATE_ldr_imm(dc, Rt, Rn, imm, post)                                      \
```
derekbruening, line 542: Not sure why this needs the "post": that is just an encoding detail that doesn't affect semantics, and it only applies to a zero offset, right?

core/ir/aarch64/instr_create.h
derekbruening: (Wondering what happens to the duplicated enum values in the Doxygen generated docs?  Is there a warning from Doxygen?)

core/ir/aarch64/instr_create.h
derekbruening: My comment is based on only passing this to opnd_create_far_base_disp_ex().  See the comment for the pre_index field in opnd.h.  For creating semantically-different variants of the instruction itself, those are going to change the other operands too and not just the memop, right?  Not sure what is currently done for other A64 instrs, but for A32 we have different variants of the macro itself for different write-back options.

core/ir/aarch64/instr_create.h
merckhung: Got it. I will change the code accordingly. It will take a while on my end, but I will submit a version by the end of Wed. I didn't look at the generated Doxygen docs, let me check it out. I don't remember it warns, perhaps I didn't notice one back then.

merckhung: I have had a look of those aforementioned files. Thank you!
instr.h: defines architectural values WRT instruction encoding/decoding.
opnd.h: defines architectural values WRT operand and register encoding/decoding.
instr_create_shared.h: seems like defining architecturally independent macros and definitions.

For the DMB macro and its IMM definitions, I feel creating a new ir/aarchxx/instr_create_aarchxx.h is a way to go.

Because DMB's IMM enum doesn't share the exactly same semantics with the other arch (e.g. x86), so the instr_create_shared.h is ruled out;

instr.h and opnd.h are more related to architectural definitions, rather than instruction-level definitions. So I think I will go with creating a new ir/aarchxx/instr_create_aarchxx.h and move both ARM and AArch64's DMB macro and IMM enum to the new file. If no objection, then I will go with this path forward. Thank you.

core/ir/aarch64/instr_create.h
merckhung: I should've read the explanation of the "pre_index" bitfield. Thank you.
I did gdb tracing and found pre_index is hard-coded to true, so I figured the way to workaround it is to post-process it (manually changing to false, after encoded).

Today, I've tried to passing a negative offset to opnd_create_far_base_disp_ex() and I found it doesn't encode correctly.
Report: Internal Error: DynamoRIO debug check failure: /home/merck/Code/dynamorio_orig/core/arch/emit_utils_shared.c:1061 nxt_pc != NULL

** I will look into it tomorrow to see what actually happens, but I guess the AArch64 encoder doesn't catch it.

I realized that when a pre_index LDR (immediate) is encoded the instr_t will be handled by encode_opndsgen_38400400() in encode_gen.h (line: 8085). The AArch64 encoder tries each possible encoding functions until one of those catches it.

LDR (immediate) differs pre- or post-index by looking at bit 11 and 12.
When it comes to encoding a post-index instruction, the encoder will fall through to next encode_opndsgen_38400c00() function (bit11 = 1), and the instr_t is supposed to be caught by it.
In the first commit of this pull request, I found toggling the pre_index bitfield does affect how AArch64 encodes LDR (immediate). But passing a negative offset to opnd_create_far_base_disp_ex() doesn't seem be encoded correctly in AArch64.

Let me have a look tmr, and I will come back with a better change for reviews. Thank you!

merckhung: BTW, thank you for laying out the design principles.

derekbruening: > Today, I've tried passing a negative offset to opnd_create_far_base_disp_ex() and I found it doesn't encode correctly.
> Report: Internal Error: DynamoRIO debug check failure: /home/merck/Code/dynamorio_orig/core/arch/emit_utils_shared.c:1061 nxt_pc != NULL

I remember something off in the A64 branch linking code: #4273, but that's a different assert.  @AssadHashmi might know more?

core/ir/aarch64/instr_create.h
derekbruening: ir/aarchxx/instr_create_aarchxx.h seems good to me.

core/ir/aarch64/instr_create.h
merckhung: I've done some investigations about the post-LDR encoding problem.

Summary:
0) I think the explanation of pre_index bit-field in the opnd.h is described in terms of A64 decoding.
   For recording a decoded LDR is pre- or post-index, it uses the pre_index bit-field to distinguish it.
   The value of the pre_index will be written by A64 decoder according to the input app's code sequence,
   and the JIT is supposed to emit a correct LDR version by looking at the value of the pre_index field.
   
   It's not meant to say a post-index LDR cannot have an offset other than 0, but both pre- and post-index
   LDR could both have a zero offset encoded. So the pre_index field is needed for distinguishing them in DR.

   Therefore, for *manually* encoding an LDR in the DR program itself, the pre_index field is hard-coded to *true*.
   So if a developer calls OPND_CREATE_MEMPTR()->opnd_create_far_base_disp_ex() in DR.
   The JIT will never have a chance to emit a correct post-LDR instr.

Tl;Dr
1) Only LDR/STR family uses encode_opnd_mem9/encode_opnd_mem9post/encode_opnd_mem9_bytes functions to encode the signed offset imm9 field (bit 20:12). If a change happens in those will only affect LDR/STR family.

2) In the encode_opnd_mem9post() and the following encode_opnd_mem9_bytes() function, if post==true, then the disp is requested to be 0, otherwise it will fail the encoding (codec.c, line: 662). *This is where it fails.*
    if (post ? (disp != 0) : (disp < -256 || disp > 255))
        return false;

3) Based on point #1, regardless of the indexing direction, per A64 SPEC. both pre- and post-index LDR/STR support encoding a signed 0 imm9 offset (AARM, p. 1000). But in the point #2, current impl. limits an input of post==true to have a zero disp and allows post==false to have a disp of range from -256 to +255. *This impl. is contradicting with the A64 SPEC.*

4) Because the imm9 field's type is a signed integer. If passing an negative offset to opnd_create_far_base_disp_ex() function for the purpose of determining the pre- or post-index attribute of an LDR/STR encoding will result in prohibiting negative offsets to be correctly encoded in the further pre-index LDR/STR.
*Therefore, will not go this way, but will think about setting FLAG_NEGATIVE and converting value to positive*

5) The later AArch64 instruction encoding process doesn't determine a pre- or post-index LDR by directly looking at the pre_index bit-field. Instead, the A64 encoder tries each possible LDR encoding functions until one of those reports successful. *A bit inefficient in A64 encoding process.* Furthermore, if a preceding encoding function returns TRUE, it may cascadingly block a later *more correct* one. This actually complicates the constructions of opnd_t variables, because the relationship among those opnd_t variables determines the selection of the later A64 encoding functions. Rather than simply determining by looking at the values of individually opnd_t vars at first.

Conclusion:
6) I will study more about ARM's implementation (NEGATIVE FLAG, positive conversion), before I implement a change for A64. Will come back with a more considerate commit that is more aligned with current ARM port.

core/ir/aarch64/instr_create.h
merckhung: I have created instr_create_aarchxx.h in the latest commit, but I am struggling with making it appear in the Doxygen htmls.
Do you know how to make instr_create_aarchxx.h show up in the doxygen updates? Thank you.

core/ir/aarch64/instr_create.h
merckhung: Done in the latest commit, but the Doxygen is not generated correctly.

core/ir/opnd.h
merckhung: Done the change accordingly in the latest commit.

core/ir/aarch64/instr_create.h
merckhung: Reverted the changes of opnd_create_far_base_disp_ex(). Refined INSTR_CREATE_ldr_imm() macros to align with DR design (moved away from AArch64 mnemonics style to DR style). If the developers want a post-indexed LDR, they are supposed to call opnd_set_zero_offset_pre_index() on the just generated memop to set the pre_index field from true to false (indicating post-indexed), see an example in core/arch/aarch64/emit_utils.c.

merckhung:
Close this pull request since I messed up with my fork with a force-push. A clean and new pull request has been created at #4390 4390.

core/ir/aarch64/instr_create.h
derekbruening: It needs to be added to the file list in genapi.pl.

Issue #4329, opened 2020-06-22
AArch64: Implement effective address of DC ZVA cache zero operation


mwvantol:
While running the drcachesim component, AArch64 data cache operations such as DC ZVA are not marked as making a memory access. However, these are often used in userspace in for example bzero() implementations to efficiently zero out memory.

derekbruening:
Presumably this is an AArch64 decoder bug where `DC ZVA` has the wrong operand type and so is seen by tools like drcachesim's tracer as not accessing memory.  @AssadHashmi hoping you can take a look?

mwvantol:
There are also several other cache operations that can execute in EL0/userspace. These would be to invalidate or clean (write back) data from the caches;
IC IVAU

DC CVAC
DC CVAU
DC CVAP
DC CIVAC

Note that all these instructions are variants of the SYS instruction, depending on the immediate value used (and there are many more that are privileged). More valid variations exist when certain architecture extensions are present, such as v8.5 memory tagging, which adds variants of these to deal with the data and tags.

I only found a mention of IC IVAU in the DynamoRIO source code as it needs to deal with that for self-modifying code. I haven't seen any of the above (except DC ZVA) in the workloads I'm looking at so far, so I don't know how DynamoRIO behaves around them. DC ZVA is certainly the most commonly used.

AssadHashmi:
Hi @mwvantol, do you have a small test case which replicates the DC ZVA omission?

mwvantol:
I don't have anything at hand, but will try to create one, and also include the other cacheops I listed above.

mwvantol:
I've written and ran the following small program wrapped with the C runtime; you'll also find DC ZVA's in the C runtime initialization code. I confirmed that for none of the IC/DC instructions I see the effective memory address on the drcachesim tracing interface.
```
    .arch armv8.1-a
    .file   "icdc-test.c"
    
    .text
    .align 2
    .global main

main:
    adr x1, flushaddr
    dc zva, x1
    ic ivau, x1
    dc cvac, x1
    dc cvau, x1
    dc civac, x1

    ret


    .data
    .align 10
    .type   flushaddr, %object
    .size   flushaddr, 8

flushaddr:
    .xword 0x1

```
Please find the binary attached, or compile the above yourself by simply passing it as a .s file to GCC.
[icdc_test.zip](https://github.com/DynamoRIO/dynamorio/files/4819115/icdc_test.zip)

derekbruening:
[...]

Issue #4382, opened 2020-07-27
Add DR_DISASM_AARCH64 standard disassembly style


derekbruening:
Xref #4381 

This covers adding `-syntax_aarch64` and `DR_DISASM_AARCH64` support for disassembly in standard style for AArch64.  (Using `DR_DISASM_ARM` results in no operands being printed at all: #4381).

derekbruening:
For now I'm lumping opcode alias issues here as well: we can separate if desired.

For an instruction in the aarch64 trace in the new trace samples repository https://github.com/DynamoRIO/drmemtrace_samples, DR calls it `subs`:

```
$ bin64/drrun -t drcachesim -indir drmemtrace_samples/threadsig.aarch64/drmemtrace.threadsig.11451.1365.dir -simulator_type view -module_file drmemtrace_samples/threadsig.aarch64/drmemtrace.threadsig.11451.1365.dir/raw/modules.log -alt_module_dir drmemtrace_samples/threadsig.aarch64/binaries
...
  0x0000000000402e4c  6b0300bf   subs   %w5 %w3 lsl $0x00 -> %wzr

$ clients/bin64/drdisas 6b0300bf
 6b0300bf   subs   %w5 %w3 lsl $0x00 -> %wzr
```

But everybody else calls it `cmp`:
```
$ echo "0xbf 0x00 0x03 0x6b" | /usr/bin/llvm-mc-8 -arch aarch64 --disassemble
	cmp	w5, w3
$ aarch64-linux-gnu-objdump -d threadsig 
  402e4c:       6b0300bf        cmp     w5, w3
```

In the manual they are listed as aliases of each other, so it seems to be a disassembly style issue: hence lumping under here for now.

AssadHashmi:
Xref #4386 Ideally we should disassemble SYS instructions to their aliases, `DC`, `IC` etc.
e.g. `SYS #3, C7, C4, #1, Xt` should disassemble as `DC ZVA, Xt`

derekbruening:
> Xref #4386 Ideally we should disassemble SYS instructions to their aliases, `DC`, `IC` etc.
> e.g. `SYS #3, C7, C4, #1, Xt` should disassemble as `DC ZVA, Xt`

The proposal is to split SYS into separate opcodes for each sub-type, which first needs #4393.

derekbruening:
[...]

PR #4386, opened 2020-07-29
i#4329: Implement effective address of DC cache operations

AssadHashmi:
This adds and tests the SYS encoding which supports the DC,
IC, AT and TLB1 instruction aliases.

Issue: #4329


derekbruening:
> Issue: #4328

Looks like a typo: #4329

derekbruening:
(Please assign a reviewer when ready: will assume still a draft until then)

AssadHashmi:
> (Please assign a reviewer when ready: will assume still a draft until then)

Will do. What is the next step in fixing this for drcachesim? AIUI SYS has to indicate a memory access is being made, but the operands are immediates and a register.

AssadHashmi:
AIUI there are 2 things that we need to do when decoding:
1 Check the SYS operands and determine if the resulting operation will touch memory.
2 Mark the register operand as an address operand if 1) is true.

1 can be done in `codec.c` decode functions.
But for 2) how does dr mark a register as a memory operand at the decoder level, i.e. not from the API/IR level?

derekbruening:
> > (Please assign a reviewer when ready: will assume still a draft until then)
> 
> Will do. What is the next step in fixing this for drcachesim? AIUI SYS has to indicate a memory access is being made, but the operands are immediates and a register.
> 
> I guess for certain combinations of immediates we have to mark the register as an addressing operand?

Right, it would be a base register of an addressing operand.  DC ZVA will even trigger faults and watchpoints, right?  So it seems like a regular memory operand.

derekbruening:
> AIUI there are 2 things that we need to do when decoding:
> 1 Check the SYS operands and determine if the resulting operation will touch memory.
> 2 Mark the register operand as an address operand if 1) is true.
> 
> 1 can be done in `codec.c` decode functions.
> But for 2) how does dr mark a register as a memory operand at the decoder level, i.e. not from the API/IR level?

The decoder should create a base-disp operand instead of a register operand.  Just like it would do for a regular memory load where the instruction encoding specifies the base register.

core/ir/aarch64/codec.c
```diff
@@ -1532,18 +1560,32 @@ encode_opnd_cond(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_out)
...
-    return decode_opnd_int(5, 14, false, 0, OPSZ_2, 0, enc, opnd);
+    return decode_opnd_int(16, 3, false, 0, OPSZ_3b, 0, enc, opnd);
 }
 
 static inline bool
-encode_opnd_sysops(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_out)
+encode_opnd_sys_op1(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_out)
 {
-    return encode_opnd_int(5, 14, false, 0, 0, opnd, enc_out);
+    return encode_opnd_int(16, 3, false, 0, 0, opnd, enc_out);
```
derekbruening, line 1588: (OK, confirmed from manual: op1 is 16+3, cn is 12+4, cm is 8+4, op2 is 5+3.)

AssadHashmi:
> The decoder should create a base-disp operand instead of a register operand. Just like it would do for a regular memory load where the instruction encoding specifies the base register.

Ok, I'll use the usual method of creating a base-disp operand instead of a register (I asked because I thought there might be another way of marking operands internally, below the API functions).

@mwvantol has kindly put together the list of `DC` ops which touch memory in the v8.0 spec:
```
CVAC
CVAU
CVAP
CIVAC
ZVA
```

core/ir/aarch64/instr_create.h
```diff
@@ -556,6 +556,10 @@
     instr_create_1dst_1src((dc), OP_mrs, (Xt), (sysreg))
 #define INSTR_CREATE_msr(dc, sysreg, Xt) \
     instr_create_1dst_1src((dc), OP_msr, (sysreg), (Xt))
+#define INSTR_CREATE_sys(dc, op1, CRn, CRm, op2, Rn) \
+    instr_create_0dst_5src(dc, OP_sys, op1, CRn, CRm, op2, Rn)
```
derekbruening, line 560: We want `DC` instructions to have a memory operand, not a plain register.  That could be done either by having encoding variants of `OP_sys`: some take a register, some a base-disp (and the encoder extracts the base, just like for `OP_ldr`).  Or, we could split the opcodes, having `OP_dc` or `OP_sys_dc` or sthg.  Either way, we probably want `INSTR_CREATE_dc` which takes in a base-disp operand.

core/ir/instr.h
```diff
@@ -2589,6 +2589,16 @@ instr_t *
 instr_create_0dst_4src(dcontext_t *dcontext, int opcode, opnd_t src1, opnd_t src2,
                        opnd_t src3, opnd_t src4);
 
+DR_API
+/**
+ * Convenience routine that returns an initialized instr_t allocated
```
derekbruening, line 2594: I know the others are missing it too: but including a `#` (so `#instr_t`) will add a Doxygen link.

suite/tests/api/dis-a64.txt
```diff
@@ -1304,16 +1304,21 @@ d5033f5f : clrex                          : clrex  $0x0f
...
 d5100000 : msr    s2_0_c0_c0_0, x0        : msr    %x0 $0x0000
 d51b4201 : msr    nzcv, x1                : msr    %x1 -> %nzcv
 d51b4402 : msr    fpcr, x2                : msr    %x2 -> %fpcr
 d51b4423 : msr    fpsr, x3                : msr    %x3 -> %fpsr
 d51bd044 : msr    tpidr_el0, x4           : msr    %x4 -> %tpidr_el0
 d51fffff : msr    s3_7_c15_c15_7, xzr     : msr    %xzr $0x7fff
-d5280000 : sysl   x0, #0, C0, C0, #0      : sys    $0x0000 -> %x0
-d52fffff : sysl   xzr, #7, C15, C15, #7   : sys    $0x3fff -> %xzr
+d5280000 : sysl   x0, #0, C0, C0, #0      : sysl   $0x00 $0x00 $0x00 $0x00 -> %x0
+d52fffff : sysl   xzr, #7, C15, C15, #7   : sysl   $0x07 $0x0f $0x0f $0x07 -> %xzr
```
derekbruening, line 1321: Whichever of these is `dc`: I guess we leave the preferred-alias issue to #4382.  Alternatively, for some aliases, we could have actually-separate opcodes (see above as well).

suite/tests/api/ir_aarch64.c
```diff
@@ -3960,6 +3960,71 @@ test_asimddiff(void *dc)
...
+
+    /* SYS #<op1>, <Cn>, <Cm>, #<op2>{, <Xt>}
+     *
+     * Data cache operations are aliases of SYS:
+     * DC <dc_op>, <Xt>
+     * is equivalent to
+     * SYS #<op1>, C7, <Cm>, #<op2>, <Xt>
+     */
+
+    /* DC ZVA, Xt => SYS #3, C7, C4, #1, Xt */
```
derekbruening, line 3977: It would be more convenient for the user, for (relatively) frequently constructed instructions like this, to have dedicated constructors (`INSTR_CREATE_dc_zva`?) or named constants for the immediates.

derekbruening, line 3977: Also, on the analysis side: for something like #4328, drcachesim needs to identify cache flush instructions.  That would similarly be easier without checking hardcoded numeric values for 4 different immediates.  Some kind of separate opcode, or at least named constants, would be nicer.

This is somewhat similar, though more significant, to the prefetch type distinction, which @abhinav92003 recently put in and had to do decoding of immediates in drcachesim: https://github.com/DynamoRIO/dynamorio/blob/master/clients/drcachesim/tracer/instru.cpp#L82.  IMHO it would be nicer to have DR provide such decoding and not force clients to look at immediate bits.  It could again be separate opcodes or at least named constants (xref DR_DMB_OSHLD, etc. where arm/insr_create.h has named constants for `OP_dmb`).

derekbruening: Note that there is precedent in DR for splitting opcodes where convenient for tools and not strictly matching ISA manuals.  Assembler mnemonics are using `DC ZVA` and not `SYS 0xxx 0xx 0xxx 0xxx` anyway so separate opcodes need not be seen as divergent from standard practices; though separating only in INSTR_CREATE and providing named constants is another way to go.

derekbruening:
> > The decoder should create a base-disp operand instead of a register operand. Just like it would do for a regular memory load where the instruction encoding specifies the base register.
> 
> Ok, I'll use the usual method of creating a base-disp operand instead of a register (I asked because I thought there might be another way of marking operands internally, below the API functions).
> 
> @mwvantol has kindly put together the list of `DC` ops which touch memory in the v8.0 spec:
> 
> ```
> CVAC
> CVAU
> CVAP
> CIVAC
> ZVA
> ```

`DC IVAC` also?

The others (the set/way and all) will still need to be recognized by drcachesim to label as cache flushes so any help in identifying them would be good as well.

derekbruening:
> The others (the set/way and all) will still need to be recognized by drcachesim to label as cache flushes so any help in identifying them would be good as well.

WDYT about the separate opcode approach?  `OP_dc` (or `OP_sys_dc` to make it a subset of `sys`)?  It would make it easier for clients.

mwvantol:
> > > The decoder should create a base-disp operand instead of a register operand. Just like it would do for a regular memory load where the instruction encoding specifies the base register.
> > 
> > 
> > Ok, I'll use the usual method of creating a base-disp operand instead of a register (I asked because I thought there might be another way of marking operands internally, below the API functions).
> > @mwvantol has kindly put together the list of `DC` ops which touch memory in the v8.0 spec:
> > ```
> > CVAC
> > CVAU
> > CVAP
> > CIVAC
> > ZVA
> > ```
> 
> `DC IVAC` also?
> 
> The others (the set/way and all) will still need to be recognized by drcachesim to label as cache flushes so any help in identifying them would be good as well.

That's a good question. DC IVAC is EL1 and above only. I assumed we only needed to support all legal operations that can execute at EL0. If we want to also support EL1 operations, there's a whole new can of worms with AT and TLBI variants that operate on addresses.

derekbruening:
> > > > The decoder should create a base-disp operand instead of a register operand. Just like it would do for a regular memory load where the instruction encoding specifies the base register.
> > > 
> > > 
> > > Ok, I'll use the usual method of creating a base-disp operand instead of a register (I asked because I thought there might be another way of marking operands internally, below the API functions).
> > > @mwvantol has kindly put together the list of `DC` ops which touch memory in the v8.0 spec:
> > > ```
> > > CVAC
> > > CVAU
> > > CVAP
> > > CIVAC
> > > ZVA
> > > ```
> > 
> > 
> > `DC IVAC` also?
> > The others (the set/way and all) will still need to be recognized by drcachesim to label as cache flushes so any help in identifying them would be good as well.
> 
> That's a good question. DC IVAC is EL1 and above only. I assumed we only needed to support all legal operations that can execute at EL0. If we want to also support EL1 operations, there's a whole new can of worms with AT and TLBI variants that operate on addresses.

For DR executing an app in nearly all uses cases today, we would not see EL1 instructions, but we would like to have the decoder properly decode all instructions for use as a standalone decoder, and possible future use on privileged code (such as in https://github.com/DynamoRIO/drk).  So could be punted to later but we would still like to have it.

suite/tests/api/ir_aarch64.c
AssadHashmi: Good point about divergence!

Just to be clear, do you mean having e.g. `OP_dc_zva` as an opcode rather than having `OP_sys_dc` as an opcode with constructors like `INSTR_CREATE_dc_zva` adding ZVA as operands?

AssadHashmi:
> WDYT about the separate opcode approach? `OP_dc` (or `OP_sys_dc` to make it a subset of `sys`)? It would make it easier for clients.

That's a balanced approach.
I was wondering if we could have `OP_dc_zva`, handling `zva` as part of the `dc` opcode rather than a parameter to it?

suite/tests/api/ir_aarch64.c
derekbruening: I'm not sure what the best choice is.  I would definitely vote for separating out `dc` from `sys`: I think there are few downsides and lots of benefits there.  For separating further in the opcodes: `OP_dc_zva`, `OP_dc_cvac`, etc.: I kind of like that as it seems to match other ISA's better, with each opcode a distinct semantic operation.  But then I'd want to also split `OP_prfm` up, and that also brings up `OP_dmb`.  If there were just one "sub-opcode" enum of named constants, like for `OP_dmb` with `DR_DMB_OSHLD`, etc., that might be a good compromise?  So `OP_dc` with `DR_DC_ZVA`, etc., with a client dispatching on a single immediate operand?  The creation would then be `INSTR_CREATE_dc` taking in a single immediate.  But completely separate opcodes doesn't seem so bad either.

derekbruening:
> I was wondering if we could have `OP_dc_zva`, handling `zva` as part of the `dc` opcode rather than a parameter to it?

Discussion is here: https://github.com/DynamoRIO/dynamorio/pull/4386/files/ebc67095e1c6c428d2013f7758eb36ac7fe1313e#r462356954

suite/tests/api/ir_aarch64.c
derekbruening: Maybe get more opinions?  @johnfxgalea has written a bunch of clients; so has @zhaoqin in the past at least.

mwvantol: The tricky bit here is that DC ZVA is a bit of an odd one out, as it is not really a cache maintentance operation (but a store zero operation). Perhaps you'd want some OP_dc_flush and OP_dc_zva then to separate between the two categories?

zhaoqin: Do not know much about details or context here, just my two cents here.

IMHO, the opcode should be the semantic behavior of the instruction, and the operand should be the affected target, e.g., data type: memory, reg, immed, data size: 1 byte, 2 bytes etc.

Instruction like OP_dmb with the DR_DMB_* enum is a bit strange to me. Those enums are more like a semantic policy.
It is like we have an OP_shift instruction and the operand is left/right/arith/logic.

So I would prefer separate opcode for each semantic behavior.

johnfxgalea: Personally, I would keep the IR as close to the ISA as possible. If A64 sees DC as a separate single opcode, then I think DR should do the same. For instance, on Intel, there is the VPCMPW instruction. This instruction takes an imm operand which defines how the mask should be constructed, e.g., if the value is 0, do an equality comparison, if the value is 1, do a less then comparison. In this case, DynamoRIO does not have a separate opcode for each case.

mwvantol: DC, as well as AT, CFP, CPP, DVP, IC and TLBI are all aliases of the SYS instruction at specific indexes that the SYS operates on. The Arm ARM states that the use of the alias mnemonics is preferred for SYS instructions at those indexes. However, that would still give you a single 'DC' instruction, and not a way to distinguish between DC ZVA and the other DC flush operations.

johnfxgalea: Oh, I am in agreement with separating DC from SYS. I am talking about whether further opcodes like OP_dc_zva and OP_dc_cvac should be added.

Don't hold a strong opinion, just my two cents :p

derekbruening: In the past we generally followed what @zhaoqin and I are espousing about the opcode defining semantics and the operands defining only the degree/scale/size and which registers/addresses.  IMHO that makes it easier to write analysis code. For 32-bit ARM, we did things like splitting up `VST1` into `OP_vst1_{8,16,32,64,lane_8,lane_16,lane_32}`.  Looks like we weren't 100% consistent though: we did not split up `DMB`, following the ISA there with a semantic immediate left exposed in the IR, probably b/c we just didn't consider it closely in the initial port.

Re: vpcmpw: looks like it's AVX-512 which brought in a ton of complex opcodes.  I think it was more that it was easiest to just add the ISA opcodes and that the semantics were not closely examined (many of those SIMD ops are complex), than that there was any conscious choice to not split it up.   I was not aware of what that opcode did until it was pointed out here.

So I guess it's your decision @AssadHashmi.  We have votes for each of these:

1. `OP_dc` with an immediate (and enum `DR_DC_ZVA`, `DR_DC_CVAC` or somesuch)
2. `OP_dc_zva`, plus `OP_dc_flush` with an immediate + enum
3. `OP_dc_zva`, `OP_dc_cvac`, etc.

AssadHashmi: Initially I was in agreement with @johnfxgalea in that the IR should remain faithful to the ISA as far as possible and actually started doing that change yesterday. Having slept on it though, I can see that `OP_dc_zva`, `OP_dc_cvac`, etc are an improvement in terms of usability.

In this case we have a `SYS` instruction with 5 operands which has aliases in instructions like `DC` and `IC` which have 2 operands 1 of which actually tells the instruction what to do. So, 4 of `SYS`'s operands are permutations which map to a particular behaviour of `DC`, `IC` etc. To me, that justifies having OP_dc_zva, OP_dc_cvac, etc as opcodes in their own right. **In this case**.

derekbruening: SGTM.  Wondering what @johnfxgalea thinks about the drcachesim code for `OP_prfm` where right now we have just the ISA opcode with immediates the client has to go and decode: https://github.com/DynamoRIO/dynamorio/blob/master/clients/drcachesim/tracer/instru.cpp#L82.  I'm not a fan of a client having to do that: it feels like DR only decoded it half-way.  Splitting the opcode would make 18 separate `OP_prfm_read_l1`, `OP_prfm_read_l1_nt`, etc. and the client has to check for it being in a range to figure out whether it's a `PRFM`, but a range check is simpler than decoding encoded bits.  I suppose leaving the single `OP_prfm` and providing a named enum where `DR_PRFM_READ_L1_NT`, etc. are set to equal their encoded 5-bit immediate value avoids the bit decoding as well.

johnfxgalea: I agree that adding opcodes based on semantics is way better for clients. I only raised that issue b/c I want the IR to be consistent. For instance, another example relates to pshuf instructions... In an ideal world, are we going to have a different opcode for each shuffle combination?

johnfxgalea: Also what if we have separate opcodes and the user just wants to know if an instruction is OP_prfm (regardless of the semantics). Is DR going to have helper functions that take instructions, retrieves the opcodes, and tests those opcodes against each sub (semantic-aware) opcode?

derekbruening: Right, there is a tradeoff.  Yes, for the split opcode grouping into the ISA opcode, a helper would be good.  So either way there's a helper: either split opcodes with a helper to group, or a single opcode with a helper (maybe just enum for source operand integer) to obtain the sub-opcode, and there's a judgement call on where to draw the line on semantic difference consistency: prefetch types vs cache management types.  Anyway, to move this PR forward, I filed a separate #4388 on `PRFM`.

AssadHashmi: Folks, sadly our discussion turns out to be academic!

The AArch64 codec is not able to generate decode logic for multiple encodings with the same opcode. This means the policy of IR faithfully following the ISA is **built in** to the implementation.

So, not only can we **not** have `OP_dc_zve()` etc but we can't even have `OP_dc()` !

I've just pushed the latest change :-/

AssadHashmi:
With commit 9475f9c the following tests have regressed:
```
         78 - code_api|security-linux.trampoline (Failed)
         80 - code_api|security-common.codemod (Failed)
         83 - code_api|security-common.TestAllocWE (Failed)
         99 - code_api|client.timer (Failed)
        118 - code_api|client.drreg-end-restore (Failed)
        162 - code_api|tool.drcachesim.delay-simple (Failed)
        183 - code_api|tool.drcacheoff.func_view_heap (Failed)
```

I'm guessing these failures are related to `OP_sys` changes which aren't reflected in `instr_is_icache_op()` and `mangle_icache_op()`?

Nothing obvious stands out at this stage. I've not got much experience of DR's SMC implementation.

suite/tests/api/ir_aarch64.c
derekbruening: > The AArch64 codec is not able to generate decode logic for multiple encodings with the same opcode.

Not sure I follow: aren't we trying to have separate opcodes for separate encodings?  You mean there are some `SYS` encodings that overlap with the `DC` ones we're separating in some way the codec generator doesn't support today?  We can change the codec or the generator code.

mwvantol:
> With commit [9475f9c](https://github.com/DynamoRIO/dynamorio/commit/9475f9c701a319754b673b0f048f3980688a38f7) the following tests have regressed:
> 
> ```
>          78 - code_api|security-linux.trampoline (Failed)
>          80 - code_api|security-common.codemod (Failed)
>          83 - code_api|security-common.TestAllocWE (Failed)
>          99 - code_api|client.timer (Failed)
>         118 - code_api|client.drreg-end-restore (Failed)
>         162 - code_api|tool.drcachesim.delay-simple (Failed)
>         183 - code_api|tool.drcacheoff.func_view_heap (Failed)
> ```
> 
> I'm guessing these failures are related to `OP_sys` changes which aren't reflected in `instr_is_icache_op()` and `mangle_icache_op()`?
> 
> Nothing obvious stands out at this stage. I've not got much experience of DR's SMC implementation.

I've tested Assad's fix with the test program I had created earlier, and it turns out that while the DC operations now report the memory accesses correctly (and are shown on the drcachesim tracing interface as 0-byte accesses), it has broken the IC operation to the point of segfaulting DynamoRIO. That's probably also the case of the failing tests mentioned above.

AssadHashmi:
The latest commit (https://github.com/DynamoRIO/dynamorio/pull/4386/commits/b66ec31037030beb7a27796840adf5e9fffaa52b) fixes @mwvantol 's test case crash and some of the unit test regressions.
3 test still failing:
```
        118 - code_api|client.drreg-end-restore (Failed)
        162 - code_api|tool.drcachesim.delay-simple (Failed)
        183 - code_api|tool.drcacheoff.func_view_heap (Failed)
```

mwvantol:
> The latest commit ([b66ec31](https://github.com/DynamoRIO/dynamorio/commit/b66ec31037030beb7a27796840adf5e9fffaa52b)) fixes @mwvantol 's test case crash and some of the unit test regressions.
> 3 test still failing:
> 
> ```
>         118 - code_api|client.drreg-end-restore (Failed)
>         162 - code_api|tool.drcachesim.delay-simple (Failed)
>         183 - code_api|tool.drcacheoff.func_view_heap (Failed)
> ```

I've just tested it and indeed it fixes the IC issue. Also the IC instruction's memory address shows up on the drcachesim tracing interface now.

suite/tests/api/ir_aarch64.c
```diff
@@ -3960,6 +3960,59 @@ test_asimddiff(void *dc)
...
+    instr = INSTR_CREATE_sys(
+        dc, opnd_create_base_disp_aarch64(DR_REG_X0, DR_REG_NULL, 0, false, 0, 0, OPSZ_0),
```
derekbruening, line 3978: So this OPSZ_0 would become the named constant from above.

core/ir/aarch64/codec.c
```diff
@@ -1060,6 +1060,28 @@ encode_opnd_x0(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_out)
     return encode_opnd_wxn(true, false, 0, opnd, enc_out);
 }
 
+/* memx0: memory operand with no offset used as memref for SYS */
+
+static inline bool
+decode_opnd_memx0(uint enc, int opcode, byte *pc, OUT opnd_t *opnd)
+{
+    *opnd = opnd_create_base_disp(decode_reg(extract_uint(enc, 0, 5), true, false),
+                                  DR_REG_NULL, 0, 0, OPSZ_0);
```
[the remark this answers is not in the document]
derekbruening: I think the logic before was that something like `OP_lea`, which just constructs a linear address and does not do anything with it related to memory, should be `OPSZ_0`, but other things that actually affect memory in some way would have some substantial size.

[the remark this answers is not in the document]
mwvantol: > What size are AArch64 prefetches using?

They appear to be using a size of 1, or at least reported in the drcachesim traces I've looked at.

core/ir/aarch64/codec.c
```diff
@@ -1060,6 +1060,28 @@ encode_opnd_x0(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_out)
...
+}
+
+static inline bool
+encode_opnd_memx0(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_out)
+{
+    uint xn;
+    bool is_x;
+    if (!opnd_is_base_disp(opnd) || !encode_reg(&xn, &is_x, opnd_get_base(opnd), false) ||
+        !is_x)
+        return false;
```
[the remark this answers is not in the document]
AssadHashmi: Good point. I've added more checks.

core/arch/aarchxx/mangle.c
```diff
@@ -2834,7 +2834,7 @@ mangle_icache_op(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
 {
     int opc = instr_get_opcode(instr);
     if (opc == OP_sys) {
-        reg_id_t xt = opnd_get_reg(instr_get_src(instr, 1));
+        reg_id_t xt = opnd_get_base(instr_get_dst(instr, 0));
```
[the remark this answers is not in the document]
AssadHashmi: Yes, in fact it should've been a source operand from the start. I was getting things confused with SYSL.

core/ir/aarch64/codec.c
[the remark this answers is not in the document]
AssadHashmi: Agreed, it should be OPSZ_1.

suite/tests/api/ir_aarch64.c
AssadHashmi: Done in latest commit https://github.com/DynamoRIO/dynamorio/pull/4386/commits/cd48ffae460dddcad2e004c28ceabff542b7342c

suite/tests/api/dis-a64.txt
```diff
@@ -1304,16 +1304,19 @@ d5033f5f : clrex                          : clrex  $0x0f
 d5033f9f : dsb    sy                      : dsb    $0x0f
 d5033fbf : dmb    sy                      : dmb    $0x0f
 d5033fdf : isb                            : isb    $0x0f
-d5080000 : sys    #0, C0, C0, #0, x0      : sys    $0x0000 %x0
-d50fffff : sys    #7, C15, C15, #7        : sys    $0x3fff %xzr
+d5080000 : sys    #0, C0, C0, #0, x0      : sys    $0x0000 -> (%x0)
+d50fffff : sys    #7, C15, C15, #7        : sys    $0x3fff -> (%xzr)
+d50b7420 : sys    #3,  C7,  C4, #1, x0    : sys    $0x1ba1 -> (%x0)
```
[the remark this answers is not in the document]
AssadHashmi: Added comment and example to https://github.com/DynamoRIO/dynamorio/issues/4382#issuecomment-668720692

AssadHashmi:
3 regressions with the latest fix (https://github.com/DynamoRIO/dynamorio/pull/4386/commits/cd48ffae460dddcad2e004c28ceabff542b7342c):
```
code_api|security-linux.trampoline
code_api|security-common.TestAllocWE
code_api|security-common.codemod
```
@derekbruening any ideas what could cause these?

AssadHashmi:
> 3 regressions with the latest fix ([cd48ffa](https://github.com/DynamoRIO/dynamorio/commit/cd48ffae460dddcad2e004c28ceabff542b7342c)):
> 
> ```
> code_api|security-linux.trampoline
> code_api|security-common.TestAllocWE
> code_api|security-common.codemod
> ```
> 
> @derekbruening any ideas what could cause these?

Ignore the above. I think I have a fix.

derekbruening:
> > 3 regressions with the latest fix ([cd48ffa](https://github.com/DynamoRIO/dynamorio/commit/cd48ffae460dddcad2e004c28ceabff542b7342c)):
> > ```
> > code_api|security-linux.trampoline
> > code_api|security-common.TestAllocWE
> > code_api|security-common.codemod
> > ```
> > 
> > 
> > @derekbruening any ideas what could cause these?
> 
> Ignore the above. I think I have a fix.

OK, they all modify code so I would suspect cache consistency detection problems at a high level w/o looking at details.

AssadHashmi:
> For the longer-term preferred fix of splitting the opcodes, if you could file an issue on that work and describe what it entails (new codec format? just script changes?) Xref the related #4388.

Created #4393. It covers 3 major aspects of of the codec which can be broken out if required when work starts.

core/ir/aarch64/codec.c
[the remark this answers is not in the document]
derekbruening: I filed #4400 and #4401 where after further thinking I'm proposing changing from a 1 bytesize.

Issue #4379, opened 2020-07-24
Add module_mapper_t interfaces to iterate modules without mapping, and to write them out


derekbruening:
We have several use cases of manipulating module data for offline drcachesim traces, and since the format includes binary blobs, it is best to provide proper interfaces rather than having hacky scripts make assumptions about the format.

What's missing, specifically, to allow offline rewriting of module data, is:
+ A module_mapper_t interface to iterate modules without mapping.
+ A module_mapper_t interface to write out module data after it's been changed.
+ A way to specify the "print_cb" for custom data (whether the one used with drmemtrace_custom_module_data or a different one).

derekbruening:
Looks like module_mapper_t::create is already doing the 1st item above for us.

PR #4389, opened 2020-07-30
i#4379: Add module_mapper_t::write_module_data

derekbruening:
Adds a method for re-processing drcachesim offline trace module data
by adding module_mapper_t::write_module_data().

Adds a test case to the burst_replace test.

Fixes: #4379


derekbruening:
This is the missing interface that should let us cleanly produce new modules.log files: e.g., removing custom fields that were added during tracing.

derekbruening:
Right, I plan to do that after I use this for converting in the other
direction.

PR #4390, opened 2020-08-02

merckhung:
model (+7%).

A new opnd_set_zero_offset_post_index() function is created to set
the pre_index field of a memop operand to false (post-indexed).

INSTR_CREATE_ldr_imm macro is created to take Rd, memop, and imm
operands (DR way), instead of Rd, Rt, Imm (assembly mnemonics).
The opnd_set_zero_offset_post_index() on a memop allows selecting a
post-indexed encoding if needed (pre-indexed by default).

If the developers want a post-indexed LDR, they are supposed to call
opnd_set_zero_offset_pre_index() on the just generated memop to set
the pre_index field from true to false (indicating post-indexed),
see an example in core/arch/aarch64/emit_utils.c

core/ir/aarchxx/instr_create_aarchxx.h is created to share instr.
macros between AArch64 and ARM ports. INSTR_CREATE_dmb and its enums
are moved into the new file.

Issue: #2390


merckhung:
I've been struggling with making the newly created core/ir/aarchxx/instr_create_aarchxx.h appear in the Doxygen, hope to get some help.

merckhung:
I have fixed the doxygen generation problem for the newly created instr_create_aarchxx.h in this new and clean pull request.
Please kindly review the new approach again. Thank you.

Duplicate of #4375

merckhung:
The build failure of "continuous-integration/appveyor/pr — AppVeyor build failed" looks like a CI infrastructure problem.
The doxygen-1.8.17.windows.x64.bin.zip file is no longer available from the specified website.

appveyor DownloadFile http://doxygen.nl/files/doxygen-1.8.17.windows.x64.bin.zip
Error downloading remote file: One or more errors occurred.
Inner Exception: Remote server returned 404: Not Found
Command exited with code 2

derekbruening:
> Please kindly review the new approach again.

Will do.  A little behind: will try to get to it soon.

core/arch/aarch64/emit_utils.c
```diff
@@ -750,11 +750,12 @@ emit_indirect_branch_lookup(dcontext_t *dc, generated_code_t *code, byte *pc,
...
     APP(&ilist, load_tag);
     /* Load tag from fragment_entry_t* in the hashtable to x0. */
-    /* ldr x0, [x1, #tag_fragment_offset] */
+    /* ldr x0, [x1], #tag_fragment_offset */
     APP(&ilist,
-        INSTR_CREATE_ldr(
+        INSTR_CREATE_ldr_imm(
             dc, opnd_create_reg(DR_REG_X0),
-            OPND_CREATE_MEMPTR(DR_REG_X1, offsetof(fragment_entry_t, tag_fragment))));
+            opnd_set_zero_offset_post_index(OPND_CREATE_MEMPTR(DR_REG_X1, 0)),
```
derekbruening, line 757: This isn't making sense to me: the immediate/displacement for the writeback is not zero but is `sizeof(fragment_entry_t)`, and the separate immediate operand has to match the memory operand displacement.  I would think that for a non-zero immediate, the encoder will figure out whether it's a pre-indexed or post-indexed add by looking at the memory operand to see whether the displacement there is zero: if it's zero, it's post-indexed; if it's non-zero, it's pre-indexed.  The flag `pre_index` inside the memory operand is only needed if you have a 0 immediate (i.e., you're adding 0 to the base reg) and the encoder can't tell which is which and will pick the post-indexed and you really care about the actual raw bits of the encoding so you want the pre-indexed encoding (both are semantically the same).

We need help from the ARM folks for their encoder: @AssadHashmi if you could clarify: is that how the AArch64 encoder works to figure out pre vs post indexed?  IIRC that's what the ARM encoder does.  So this instruction should *not* need to set any special flags on the memory operand (and a flag on the memory operand should not be affecting whether there's a 2nd destination, the base reg).

core/ir/aarch64/instr_create.h
```diff
@@ -2532,5 +2533,16 @@ enum {
...
 
+/**
+ * Creates an LDR (immediate) instruction.
+ * \param dc      The void * dcontext used to allocate memory for the instr_t.
+ * \param Rd      The output register.
+ * \param mem     The source memory opnd.
+ * \param imm     The integer constant opnd_t operand.
+ */
+#define INSTR_CREATE_ldr_imm(dc, Rd, mem, imm)                                      \
+    instr_create_2dst_3src((dc), OP_ldr, (Rd), opnd_create_reg(opnd_get_base(mem)), \
```
derekbruening, line 2544: See above: I would expect to not have a 2nd destination based on the name, which in the manual just means it has a displacement.  I'm suggesting using `INSTR_CREATE_ldr_wbimm` to match ARM and indicate it's a "writeback" operation which modifies the base reg (and making that clear in the doxygen above).  @AssadHashmi seem reasonable?

core/ir/opnd.h
```diff
@@ -2316,6 +2316,16 @@ DR_API
  */
 bool
 opnd_set_index_extend(opnd_t *opnd, dr_extend_type_t extend, bool scaled);
+
+DR_API
+/**
+ * Assumes \p opnd is a base+disp memory reference.
+ * Sets the index register to be post-indexed.
```
derekbruening, line 2323: See the discussion above where this really does not change the semantic behavior: all it does is pick the post-indexed encoding for the case of a 0 displacement+immediate where the two encodings are semantically the same, for (rare) cases where someone cares about the encoding.  The comments should make that clear.  @AssadHashmi if you could confirm my interpretation of the `pre_index` field.

core/ir/aarchxx/instr_create_aarchxx.h
```diff
@@ -0,0 +1,51 @@
...
+#define INSTR_CREATE_AARCHXX_H 1
+
+#include "../instr_create_shared.h"
+#include "instr.h"
+
+/* DR_API EXPORT TOFILE dr_ir_macros_aarchxx.h */
+/* DR_API EXPORT BEGIN */
+
+/**
+ * @file dr_ir_macros_aarchxx.h
```
derekbruening, line 15: This new file "dr_ir_macros_aarchxx.h" is not included anywhere in the exported headers, so this change will break external clients (and our internal tests of the exported headers must not have any clients using DMB).  This file needs a #include inside core/ir/instr_create_shared.h by the other includes of dr_ir_macros_*.

core/arch/aarch64/emit_utils.c
```diff
@@ -763,50 +764,40 @@ emit_indirect_branch_lookup(dcontext_t *dc, generated_code_t *code, byte *pc,
...
     /* br x0
      * (keep in sync with instr_is_ibl_hit_jump())
      */
     APP(&ilist, INSTR_CREATE_br(dc, opnd_create_reg(DR_REG_X0)));
 
-    APP(&ilist, try_next);
-
-    /* Try next entry, in case of collision. No wraparound check is needed
-     * because of the sentinel at the end.
-     * ldr x0, [x1, #tag_fragment_offset]! */
```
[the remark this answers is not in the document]
derekbruening: I wish Github had a "no action requested" option.  Just talking here: no action.  Resolving.

core/arch/aarch64/emit_utils.c
AssadHashmi: Yes, you shouldn't need to do anything special to the memory operand, the decoder/encoder take care of such things, e.g. see `mem9post` mask in `codec.txt`
Is there any reason why `opnd_create_base_disp_aarch64()` is not being used here?

core/ir/aarch64/instr_create.h
AssadHashmi: Yes

core/ir/opnd.h
AssadHashmi: Yes, as above pre/post index are taken care of by the codec.

core/arch/aarch64/emit_utils.c
merckhung: Yes, I had noticed the existence of the opnd_create_base_disp_aarch64() and it's used only once by decoder and is used by a couple of unit tests. I probably misunderstood the design of codec a bit. Thank you for the comments, I will make a change to let the encoder take care of the pre-/post-indexed selection.

core/ir/aarch64/instr_create.h
merckhung: Will do. Thanks for the feedback.

core/arch/aarch64/emit_utils.c
```diff
@@ -750,11 +750,12 @@ emit_indirect_branch_lookup(dcontext_t *dc, generated_code_t *code, byte *pc,
     /* x1 now holds the fragment_entry_t* in the hashtable. */
     APP(&ilist, load_tag);
     /* Load tag from fragment_entry_t* in the hashtable to x0. */
-    /* ldr x0, [x1, #tag_fragment_offset] */
+    /* ldr x0, [x1], #tag_fragment_offset */
     APP(&ilist,
-        INSTR_CREATE_ldr(
+        INSTR_CREATE_ldr_imm(
```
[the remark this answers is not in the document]
merckhung: Okay, will do!

core/arch/aarch64/emit_utils.c
```diff
@@ -750,11 +750,12 @@ emit_indirect_branch_lookup(dcontext_t *dc, generated_code_t *code, byte *pc,
     /* x1 now holds the fragment_entry_t* in the hashtable. */
     APP(&ilist, load_tag);
     /* Load tag from fragment_entry_t* in the hashtable to x0. */
-    /* ldr x0, [x1, #tag_fragment_offset] */
+    /* ldr x0, [x1], #tag_fragment_offset */
```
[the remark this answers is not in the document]
merckhung: My bad, when I refactored the code again, I literally copied the original comments and forgot to change it. Thanks for the catch.

core/arch/aarch64/emit_utils.c
```diff
@@ -822,7 +813,7 @@ emit_indirect_branch_lookup(dcontext_t *dc, generated_code_t *code, byte *pc,
             XINST_CREATE_sub(dc, opnd_create_reg(DR_REG_X0), OPND_CREATE_INT8(1)));
         /* cbnz x0, miss */
         APP(&ilist,
-            INSTR_CREATE_cbnz(dc, opnd_create_instr(miss), opnd_create_reg(DR_REG_R0)));
+            INSTR_CREATE_cbnz(dc, opnd_create_instr(miss), opnd_create_reg(DR_REG_X0)));
```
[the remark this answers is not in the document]
merckhung: In the next commit, I will revert those. I agree with you, it's not necessary to change them here. In addition, this type of change has nothing to do with the purpose of this pull-request. I will revert it for now, until we find any other more strong reason to do so in the future.

core/arch/aarch64/emit_utils.c
```diff
@@ -717,29 +716,30 @@ emit_indirect_branch_lookup(dcontext_t *dc, generated_code_t *code, byte *pc,
      *     x2: app's x2
      *     TLS_REG1_SLOT: app's x0 (recovered by fragment_prefix)
      */
-
     /* Spill x0. */
-    APP(&ilist, instr_create_save_to_tls(dc, DR_REG_R0, TLS_REG3_SLOT));
-    /* Load-acquire hash mask.  We need a load-acquire to ensure we see updates
-     * properly; the corresponding store-release is in update_lookuptable_tls().
+    APP(&ilist, instr_create_save_to_tls(dc, DR_REG_X0, TLS_REG3_SLOT));
+    /* Load hash mask before loading hash_table, with DMB to prevent memory-access
```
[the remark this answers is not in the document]
merckhung: Will do!

merckhung:
I will catch up and update the patch next week, a bit delayed due to the perf cycle began last week plus performing some job duty this week.

Issue #4377, opened 2020-07-20
x86 IR has instances of missing or extra implicit operands not caught by tests: pabsb


johnfxgalea:
INSTR_CREATE_pabsb incorrectly creates a 1 dst, 2 src IR instruction. PABSB and variants have only one source and therefore  instr_create_1dst_1src should be used.

Currrent code:

`#define INSTR_CREATE_pabsb(dc, d, s) instr_create_1dst_2src((dc), OP_pabsb, (d), (s), (d))
`

Ref: https://www.felixcloutier.com/x86/pabsb:pabsw:pabsd:pabsq

johnfxgalea:
INSTR_CREATE_vmovlpd and variants incorrectly creates an instr via instr_create_1dst_1src, when it should use instr_create_1dst_2src.

```
#define INSTR_CREATE_vmovlps(dc, d, s) instr_create_1dst_1src((dc), OP_vmovlps, (d), (s))
#define INSTR_CREATE_vmovlpd(dc, d, s) instr_create_1dst_1src((dc), OP_vmovlpd, (d), (s))
```

Ref: https://www.felixcloutier.com/x86/movlpd

derekbruening:
The IR tests would fail if the INSTR_CREATE macros did not match the decoding tables: so the decoding tables must be wrong for these.  And that means the tests against other decoders must not include cases of these instrutions?  But I do see them in those tests:

```
third_party/binutils/test_decenc/drdecode_decenc_x86.expect: c4 e2 79 1c f4       vpabsb %xmm4, %xmm6
third_party/binutils/test_decenc/drdecode_decenc_x86.expect: c4 e2 79 1c 21       vpabsb (%ecx), %xmm4
```

Is it that we use Intel syntax for the binutils comparison and so it will never catch an extra implicit operand?

johnfxgalea:
@derekbruening I didn't check vpabsb yet, only pabsb so not sure. Will let you know when I get some time.

derekbruening:
Meant to do what you had reported but the grep found vpabsb too.  pabsb is the same: certainly present in the tests (just a handful pasted here...there are more):
```
third_party/binutils/test_decenc/drdecode_decenc_x86.expect: 66 0f 38 1c f4       pabsb  %xmm4, %xmm6
third_party/binutils/test_decenc/drdecode_decenc_x86.expect: 66 0f 38 1c 21       pabsb  (%ecx), %xmm4
[...]

PR #4391, opened 2020-08-03
i#4377: Arranges pabs instructions to have only one source.

johnfxgalea:
A bug fix that sets pabsb and variants to only have one destination and one source.

Fixes: #4377


johnfxgalea:
PTAL

johnfxgalea:
> I guess the api.ir test doesn't change b/c it only deals with _explicit_ operands, and the binutils tests don't change for the same reason.

That's right.

Issue #4337, opened 2020-06-26
AArch64 decoder bug on 128-bit SIMD variant of STUR/LDUR instructions


mwvantol:
While working with drcachesim we discovered that 128-bit SIMD variants of the STUR/LDUR instructions report an incorrect memory access size of 1 byte instead of the expected 16 bytes. These are quadword load/stores to the 128-bit FP/SIMD registers.

I noticed that the STUR/LDUR instructions are flagged as 'mem9' in the decoder template in core/arch/aarch64/codec.txt, where 'mem9' is defined as:
`??---------xxxxxxxxx--xxxxx-----  mem9       # gets size from 31:30`

However, according to the Arm ARM bits [23:22] form an additional opc field that is also used to determine the size (in addition to the size field in [31:30]), though effectively only bit 23 matters here;

> 8-bit variant
> Applies when size == 00 && opc == 00 .
> STUR <Bt>, [<Xn|SP>{, #<simm>}]
> 16-bit variant
> Applies when size == 01 && opc == 00 .
> STUR <Ht>, [<Xn|SP>{, #<simm>}]
> 32-bit variant
> Applies when size == 10 && opc == 00 .
> STUR <St>, [<Xn|SP>{, #<simm>}]
> 64-bit variant
> Applies when size == 11 && opc == 00 .
> STUR <Dt>, [<Xn|SP>{, #<simm>}]
> 128-bit variant
> Applies when size == 00 && opc == 10 .
> STUR <Qt>, [<Xn|SP>{, #<simm>}]

This also shows that if the 'opc' field is ignored, the 128-bit variant maps onto the 8-bit variant, which appears to correspond with the incorrect size I see reported. I don't know if this can potentially affect applications, but it certainly affects the cache model behavior.

nuviaram:
We are attempting to leverage Dynamo Rio for instruction tracing for chip-level / cache simulations. This issue is blocking us. Request help fixing this bug.

PR #4395, opened 2020-08-06
i#4337: Fix STUR/LDUR bugs

AssadHashmi:
This fixes bugs in the 128-bit SIMD variant of STUR/LDUR instructions (#4337) and adds extra IR and disassembly tests.

### DIFF
--- a/core/ir/aarch64/codec.txt
+++ b/core/ir/aarch64/codec.txt
@@ -425,33 +425,38 @@ x0110111xxxxxxxxxxxxxxxxxxxxxxxx  tbnz   tbz
 
 ## Load/store register (unscaled immediate)
 
-00111000000xxxxxxxxx00xxxxxxxxxx  sturb  mem9 : w0  # STURB Wt,[Xn,#simm]
+# Base
+10111000000xxxxxxxxx00xxxxxxxxxx  stur   mem9 : w0
+11111000000xxxxxxxxx00xxxxxxxxxx  stur   mem9 : x0
+00111000000xxxxxxxxx00xxxxxxxxxx  sturb  mem9 : w0
+01111000000xxxxxxxxx00xxxxxxxxxx  sturh  mem9 : w0
+
+# SIMD
+00111100000xxxxxxxxx00xxxxxxxxxx  stur   mem9 : b0
+01111100000xxxxxxxxx00xxxxxxxxxx  stur   mem9 : h0
+10111100000xxxxxxxxx00xxxxxxxxxx  stur   mem9 : s0
+11111100000xxxxxxxxx00xxxxxxxxxx  stur   mem9 : d0
+00111100100xxxxxxxxx00xxxxxxxxxx  stur   mem9q : q0
+
+# Base
+10111000010xxxxxxxxx00xxxxxxxxxx  ldur   w0 : mem9
+11111000010xxxxxxxxx00xxxxxxxxxx  ldur   x0 : mem9
 00111000010xxxxxxxxx00xxxxxxxxxx  ldurb  w0 : mem9
+01111000010xxxxxxxxx00xxxxxxxxxx  ldurh  w0 : mem9
 00111000100xxxxxxxxx00xxxxxxxxxx  ldursb x0 : mem9
 00111000110xxxxxxxxx00xxxxxxxxxx  ldursb w0 : mem9
-00111100000xxxxxxxxx00xxxxxxxxxx  stur   mem9 : b0
-00111100010xxxxxxxxx00xxxxxxxxxx  ldur   b0 : mem9
-00111100100xxxxxxxxx00xxxxxxxxxx  stur   mem9 : q0
-00111100110xxxxxxxxx00xxxxxxxxxx  ldur   q0 : mem9
-
-01111000000xxxxxxxxx00xxxxxxxxxx  sturh  mem9 : w0
-01111000010xxxxxxxxx00xxxxxxxxxx  ldurh  w0 : mem9
 01111000100xxxxxxxxx00xxxxxxxxxx  ldursh x0 : mem9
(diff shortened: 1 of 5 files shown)


AssadHashmi:
AppVeyor failures which seem completely unrelated to this patch:
```
[00:16:49] 97% tests passed, 6 tests failed out of 233                    
[00:16:49]                                                                
[00:17:38] Total Test time (real) = 398.94 sec                            
[00:17:38]                                                                
[00:17:38] The following tests FAILED:                                    
[00:17:38]      124 - code_api|client.winxfer (Failed)                    
[00:17:38]      142 - code_api|client.drmgr-test (Failed)                 
[00:17:38]      162 - code_api|client.pcache-use (Failed)                 
[00:17:38]      218 - code_api|client.drwrap-test-detach (Failed)         
[00:17:38]      220 - code_api|api.startstop (Failed)                     
[00:17:38]      221 - code_api|api.detach (Failed)  
```

core/ir/aarch64/codec.txt
```diff
@@ -425,33 +425,38 @@ x0110111xxxxxxxxxxxxxxxxxxxxxxxx  tbnz   tbz
...
+
+# SIMD
 00111100000xxxxxxxxx00xxxxxxxxxx  stur   mem9 : b0
-00111100010xxxxxxxxx00xxxxxxxxxx  ldur   b0 : mem9
-00111100100xxxxxxxxx00xxxxxxxxxx  stur   mem9 : q0
-00111100110xxxxxxxxx00xxxxxxxxxx  ldur   q0 : mem9
+01111100000xxxxxxxxx00xxxxxxxxxx  stur   mem9 : h0
+10111100000xxxxxxxxx00xxxxxxxxxx  stur   mem9 : s0
+11111100000xxxxxxxxx00xxxxxxxxxx  stur   mem9 : d0
+00111100100xxxxxxxxx00xxxxxxxxxx  stur   mem9q : q0
```
derekbruening, line 439: It is hard to know what is being fixed here from the diff: it looks like a bunch of lines are just being rearranged.  The commit description mentions 128-bit which I believe would be "q0": does that mean the two "q0" lines are all that changed here?  And the change is s/mem9/mem9q/?  Looking at the bug description: yes the bug is about the memory size.  Maybe add a few words to the commit description ("s/mem9/mem9q/" or sthg) making it easier to know what this is actually changing since it's hard to see from the diff?

suite/tests/api/ir_aarch64.expect
```diff
@@ -41,6 +41,27 @@ ldar   (%x1)[8byte] -> %x0
...
+stur   %b11 -> +0xff(%x11)[1byte]
+stur   %h12 -> (%x12)[2byte]
+stur   %h13 -> +0xff(%x13)[2byte]
+stur   %s14 -> (%x14)[4byte]
+stur   %s15 -> +0xff(%x15)[4byte]
+stur   %d16 -> (%x16)[8byte]
+stur   %d17 -> +0xff(%x17)[8byte]
+stur   %q18 -> (%x18)[16byte]
+stur   %q19 -> +0xff(%x19)[16byte]
+test_ldur_stur complete
```
derekbruening, line 64: Is there a separate issue on adding tests vs a third-party AArch64 decoder, like we have for x86?  That may help to find more IR bugs.  Although if the only way to compare is the standard disassembly output, that wouldn't catch this bug, since memory sizes are not printed in normal output formats.

derekbruening:
> AppVeyor failures which seem completely unrelated to this patch:
> 
> ```
> [00:16:49] 97% tests passed, 6 tests failed out of 233                    
> [00:16:49]                                                                
> [00:17:38] Total Test time (real) = 398.94 sec                            
> [00:17:38]                                                                
> [00:17:38] The following tests FAILED:                                    
> [00:17:38]      124 - code_api|client.winxfer (Failed)                    
> [00:17:38]      142 - code_api|client.drmgr-test (Failed)                 
> [00:17:38]      162 - code_api|client.pcache-use (Failed)                 
> [00:17:38]      218 - code_api|client.drwrap-test-detach (Failed)         
> [00:17:38]      220 - code_api|api.startstop (Failed)                     
> [00:17:38]      221 - code_api|api.detach (Failed)  
> ```

Please look at the very bottom where all but one of those are thrown out because they are on the flaky list: that one is code_api|client.winxfer.  Please link to the issue covering a test being ignored, or file one if there isn't one.  In this case that is under #4058, in particular https://github.com/DynamoRIO/dynamorio/issues/4058#issuecomment-651255163 pointed out there's an output ordering flakiness which hopefully someone will fix.

core/ir/aarch64/codec.txt
AssadHashmi: Yes, the fix is to use the correct mask: mem9q. The re-arrangement was to put the encodings in organised groups and conventional sequences to make maintenance and future changes less error prone. It was quite difficult to read.

suite/tests/api/ir_aarch64.expect
AssadHashmi: There's no separate issue. It's not clear to me what the intention for such an issue would be. Is it to let others exercise the IR by writing their own use-cases/tests?

derekbruening: No, the idea is to add ground-truth tests of DR's decoder/encoder/IR to DR's test suite.  We have internal-consistency tests (that the decoder, encoder, fast decoder/size decoder (for x86), INSTR_CREATE, etc. all agree with each other) for AArch64, but nothing checking that those components aren't all wrong.  For x86 we have a bunch of tests comparing the decoder to other 3rd-party decoders: comparing to binutils and to udis86.

AssadHashmi: Ah right! Raised #4396.

derekbruening: Thanks!

Issue #3339, opened 2019-01-11
Re-encoding instructions changes encoding type


rlyerly:
Decoding an instruction into DynamoRIO's IR format and then directly re-encoding back to machine code may change the machine-level encoding type.  This happens even if the instruction is never modified using the IR APIs.  Below is an example of dumping a function's instructions using DynamoRIO as a standalone decoder/re-encoder:

    byte *start, *end; // Start & end address of memory-mapped function
    byte *real; // Real virtual address of function
    byte *prev;
    ... (retrieve real, start & end) ...
    while(start < end) {
      instr_t *instr = instr_create(GLOBAL_DCONTEXT);
      instr_init(GLOBAL_DCONTEXT, instr);
      prev = start;
      start = decode_from_copy(GLOBAL_DCONTEXT, start, real, instr);
      std::cout << "Instruction size: " << instr_length(GLOBAL_DCONTEXT, instr) << std::endl;
      disassemble_with_bytes(GLOBAL_DCONTEXT, real, STDERR);
      instr_free(GLOBAL_DCONTEXT, instr);
      real += start - prev;
    }

The while-loop decodes all instructions in a function (mapped into memory from an on-disk binary) and prints their sizes using `instr_length()`.  This which forces re-encoding the instruction to determine its size since the loop uses `decode_from_copy()` to decode instructions.  This code produces the following output for a given binary:

    Instruction size: 10
     48 8b 04 25 d0 35 92 mov    0x009235d0[8byte] -> %rax
     00

The instruction's original size is 8 bytes, but DynamoRIO's re-encoding process changes the machine-level encoding so that it is now 10 bytes.  According to the discussion [here](https://groups.google.com/forum/#!topic/DynamoRIO-Users/F7zbNtF6zdc), this is because DynamoRIO walks an encoding template from specialized to general encoding types.  In this particular situation, DynamoRIO found a more specialized encoding for the instruction versus what was emitted by the compiler.  The instruction's change in size is a side-effect of changing the encoding.

Being able to control the encoding types may provide more flexibility for users, especially for instances where the user explicit control.  For example, when using DynamoRIO as a standalone decoder/re-encoder users may not want to change the code size as it invalidates control flow targets.  Exposing encoding controls may lead to finicky APIs, however, especially for encoding- or user-specified restrictions.  For example, what if the user requests an encoding type that is not compatible with the instruction's operands?

[...]

PR #4397, opened 2020-08-07
i#4328: Handle AArch64 cache flush operations in drmemtrace.

abhinav92003:
Maps variants of AArch64 flush instructions to existing instr and data flush trace types. Also adds count of instr and data flush operations to basic_counts simulator.

Issue: #4328

### DIFF
--- a/clients/drcachesim/tests/allasm-aarch64-prefetch-basic-counts.templatex
+++ b/clients/drcachesim/tests/allasm-aarch64-prefetch-basic-counts.templatex
@@ -8,8 +8,8 @@ Total counts:
           81 total prefetches
      .* total data loads
      .* total data stores
-           0 total icache flushes
-           0 total dcache flushes
+     .* total icache flushes
+     .* total dcache flushes
            1 total threads
      .* total scheduling markers
      .* total transfer markers
@@ -25,8 +25,8 @@ Thread .* counts:
           81 prefetches
      .* data loads
      .* data stores
-           0 icache flushes
-           0 dcache flushes
+     .* icache flushes
+     .* dcache flushes
      .* scheduling markers
      .* transfer markers
      .* function id markers
--- a/clients/drcachesim/tests/allasm-aarch64-prefetch-basic-counts.templatex
+++ b/clients/drcachesim/tests/allasm-aarch64-prefetch-basic-counts.templatex
@@ -8,6 +8,8 @@ Total counts:
           81 total prefetches
      .* total data loads
      .* total data stores
+           0 total icache flushes
+           0 total dcache flushes
            1 total threads
      .* total scheduling markers
      .* total transfer markers
@@ -23,6 +25,8 @@ Thread .* counts:
           81 prefetches
      .* data loads
      .* data stores
+           0 icache flushes
+           0 dcache flushes
      .* scheduling markers
      .* transfer markers
      .* function id markers
--- a/clients/drcachesim/tests/allasm_aarch64_flush.asm
+++ b/clients/drcachesim/tests/allasm_aarch64_flush.asm
@@ -1,0 +1,69 @@
+/* **********************************************************
+ * Copyright (c) 2020 Google, Inc. All rights reserved.
+ * Copyright (c) 2016 ARM Limited. All rights reserved.
+ * **********************************************************/
+
+/*
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice,
+ *   this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of ARM Limited nor the names of its contributors may be
+ *   used to endorse or promote products derived from this software without
+ *   specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL ARM LIMITED OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+        .global  _start
+
+        .align   6
+_start:
+        // Align stack pointer and get some space.
+        mov      x0, sp
+        bic      x0, x0, #63
+        mov      x1, x0        // x1 is top of region
+        sub      x0, x0, #1024 // x0 is bottom of region
+        mov      sp, x0
+
+        adr      x0, helloworld
+        adr      x1, .
+
+        // data cache flush operations.
+        dc       civac, x0
+        dc       cvac, x0
+        dc       cvau, x0
+
+        // instruction cache flush operations.
+        ic       ivau, x1
+
+        // Exit.
+        mov      w0, #1            // stdout
+        adr      x1, helloworld
+        mov      w2, #14           // sizeof(helloworld)
+        mov      w8, #64           // SYS_write
+        svc      #0
+        mov      w0, #0            // status
+        mov      w8, #94           // SYS_exit_group
+        svc      #0
+
+        .data
+        .align   6
+helloworld:
+        .ascii   "Hello, world!\n"
--- a/clients/drcachesim/tests/allasm_aarch64_flush.asm
+++ b/clients/drcachesim/tests/allasm_aarch64_flush.asm
@@ -45,12 +45,12 @@ _start:
         adr      x0, helloworld
         adr      x1, .
 
-        // data cache flush operations.
+        // Data cache flush operations.
         dc       civac, x0
         dc       cvac, x0
         dc       cvau, x0
 
-        // instruction cache flush operations.
+        // Instruction cache flush operations.
         ic       ivau, x1
 
         // Exit.
--- a/clients/drcachesim/tests/burst_flush_aarch64.cpp
+++ b/clients/drcachesim/tests/burst_flush_aarch64.cpp
@@ -1,0 +1,110 @@
+/* **********************************************************
+ * Copyright (c) 2020 Google, Inc.  All rights reserved.
+ * **********************************************************/
+
+/*
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice,
+ *   this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of Google, Inc. nor the names of its contributors may be
+ *   used to endorse or promote products derived from this software without
+ *   specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL GOOGLE, INC. OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+/* This application links in drmemtrace_static and acquires a trace during
+ * a "burst" of execution in the middle of the application.  It then detaches.
+ */
+#include "dr_api.h"
+#include <assert.h>
+#include <iostream>
+#include <signal.h>
+
+bool
+my_setenv(const char *var, const char *value)
+{
+#ifdef UNIX
+    return setenv(var, value, 1 /*override*/) == 0;
+#else
+    return SetEnvironmentVariable(var, value) == TRUE;
+#endif
+}
+
+void
+sigill_handler(int signal)
+{
+    dr_app_stop_and_cleanup();
+    std::cerr << "all done\n";
+    _Exit(0);
+}
+
+static void
+do_some_work()
+{
+    int d = 0;
+    // Expected to send SIGILL.
+    // Control will be transferred to sigill_handler
+    __asm__ __volatile__("dc ivac, %0" : : "r"(d));
+}
+
+int
+main(int argc, const char *argv[])
+{
+    signal(SIGILL, sigill_handler);
+    if (!my_setenv("DYNAMORIO_OPTIONS", "-client_lib ';;-offline'"))
+        std::cerr << "failed to set env var!\n";
+
+    std::cerr << "pre-DR init\n";
+    dr_app_setup();
+    assert(!dr_app_running_under_dynamorio());
+    std::cerr << "pre-DR start\n";
+    dr_app_start();
+    assert(dr_app_running_under_dynamorio());
+    do_some_work();
+
+    return 0;
+}
+
+/* FIXME i#2099: the weak symbol is not supported on Windows. */
+#if defined(UNIX) && defined(TEST_APP_DR_CLIENT_MAIN)
+#    ifdef __cplusplus
+extern "C" {
+#    endif
+
+/* Test if the drmemtrace_client_main() in drmemtrace will be called. */
+DR_EXPORT WEAK void
+drmemtrace_client_main(client_id_t id, int argc, const char *argv[])
+{
+    std::cerr << "wrong drmemtrace_client_main\n";
+}
+
+/* This dr_client_main should be called instead of the one in tracer.cpp */
+DR_EXPORT void
+dr_client_main(client_id_t id, int argc, const char *argv[])
+{
+    std::cerr << "app dr_client_main\n";
+    drmemtrace_client_main(id, argc, argv);
+}
+
+#    ifdef __cplusplus
+}
+#    endif
+#endif /* UNIX && TEST_APP_DR_CLIENT_MAIN */
--- a/clients/drcachesim/tests/burst_flush_aarch64.cpp
+++ b/clients/drcachesim/tests/burst_flush_aarch64.cpp
@@ -35,10 +35,13 @@
  * a "burst" of execution in the middle of the application.  It then detaches.
  */
 #include "dr_api.h"
+#include "../../../suite/tests/client_tools.h"
 #include <assert.h>
 #include <iostream>
 #include <signal.h>
 #include <setjmp.h>
+
+#define ASSERT_NOT_REACHED() ASSERT_MSG(false, "Shouldn't be reached")
 
 static sigjmp_buf mark;
 static int handled_sigill_count = 0;
@@ -54,15 +57,19 @@ sigill_handler(int signal)
 {
     handled_sigill_count++;
     siglongjmp(mark, handled_sigill_count);
-    exit(-1);
+    ASSERT_NOT_REACHED();
 }
 
+/* Attempts to execute the privileged 'dc ivac' instruction.
+ * This will raise a SIGILL. Caller must register a SIGILL handler before
+ * invoking this function.
+ */
 static void
 dc_ivac()
 {
     int d = 0;
     // Expected to raise SIGILL.
-    // Control will be transferred to sigill_handler
+    // Control will be transferred to sigill_handler.
     __asm__ __volatile__("dc ivac, %0" : : "r"(&d));
 }
 
@@ -93,7 +100,9 @@ do_some_work()
     handled_sigill_count = 0;
     int i = sigsetjmp(mark, 1);
     switch (i) {
-    case 0: dc_ivac();
+    case 0:
+        dc_ivac();
+        // TODO i#4406: Test other privileged dc and ic flush instructions too.
     }
     return;
 }
--- a/clients/drcachesim/tests/burst_flush_aarch64.cpp
+++ b/clients/drcachesim/tests/burst_flush_aarch64.cpp
@@ -69,7 +69,9 @@ int
 main(int argc, const char *argv[])
 {
     signal(SIGILL, sigill_handler);
-    if (!my_setenv("DYNAMORIO_OPTIONS", "-signal_stack_size 64K -client_lib ';;-offline'"))
+    if (!my_setenv("DYNAMORIO_OPTIONS",
+                   "-stderr_mask 0xc -signal_stack_size 64K "
+                   "-client_lib ';;-offline'"))
         std::cerr << "failed to set env var!\n";
 
     std::cerr << "pre-DR init\n";
--- a/clients/drcachesim/tests/burst_flush_aarch64.cpp
+++ b/clients/drcachesim/tests/burst_flush_aarch64.cpp
@@ -1,3 +1,4 @@
+
 /* **********************************************************
  * Copyright (c) 2020 Google, Inc.  All rights reserved.
  * **********************************************************/
@@ -37,32 +38,64 @@
 #include <assert.h>
 #include <iostream>
 #include <signal.h>
+#include <setjmp.h>
+
+static sigjmp_buf mark;
+static int handled_sigill_count = 0;
 
 bool
 my_setenv(const char *var, const char *value)
 {
-#ifdef UNIX
     return setenv(var, value, 1 /*override*/) == 0;
-#else
-    return SetEnvironmentVariable(var, value) == TRUE;
-#endif
 }
 
 void
 sigill_handler(int signal)
 {
-    dr_app_stop_and_cleanup();
-    std::cerr << "all done\n";
-    _Exit(0);
+    handled_sigill_count++;
+    siglongjmp(mark, handled_sigill_count);
+    exit(-1);
+}
+
+static void
+dc_ivac()
+{
+    int d = 0;
+    // Expected to raise SIGILL.
+    // Control will be transferred to sigill_handler
+    __asm__ __volatile__("dc ivac, %0" : : "r"(&d));
+}
+
+static void
+dc_unprivileged_flush()
+{
+    int d = 0;
+    __asm__ __volatile__("dc civac, %0" : : "r"(&d));
+    __asm__ __volatile__("dc cvac , %0" : : "r"(&d));
+    __asm__ __volatile__("dc cvau , %0" : : "r"(&d));
+}
+
+static void
+ic_unprivileged_flush()
+{
+    __asm__ __volatile__("ic ivau , %0" : : "r"(ic_unprivileged_flush));
 }
 
 static void
 do_some_work()
 {
-    int d = 0;
-    // Expected to send SIGILL.
-    // Control will be transferred to sigill_handler
-    __asm__ __volatile__("dc ivac, %0" : : "r"(d));
+    dc_unprivileged_flush();
+    ic_unprivileged_flush();
+
+    // Testing privileged instructions requires handling SIGILL.
+    // We use sigsetjmp/siglongjmp to resume execution after handling the
+    // signal.
+    handled_sigill_count = 0;
+    int i = sigsetjmp(mark, 1);
+    switch (i) {
+    case 0: dc_ivac();
+    }
+    return;
 }
 
 int
@@ -70,6 +103,8 @@ main(int argc, const char *argv[])
 {
     signal(SIGILL, sigill_handler);
     if (!my_setenv("DYNAMORIO_OPTIONS",
+                   // XXX i#4425: Fix debug-build stack overflow issue and
+                   // remove custom signal_stack_size below.
                    "-stderr_mask 0xc -signal_stack_size 64K "
                    "-client_lib ';;-offline'"))
         std::cerr << "failed to set env var!\n";
@@ -81,32 +116,7 @@ main(int argc, const char *argv[])
     dr_app_start();
     assert(dr_app_running_under_dynamorio());
     do_some_work();
-
+    dr_app_stop_and_cleanup();
+    std::cerr << "all done\n";
     return 0;
 }
-
-/* FIXME i#2099: the weak symbol is not supported on Windows. */
-#if defined(UNIX) && defined(TEST_APP_DR_CLIENT_MAIN)
-#    ifdef __cplusplus
-extern "C" {
-#    endif
-
-/* Test if the drmemtrace_client_main() in drmemtrace will be called. */
-DR_EXPORT WEAK void
-drmemtrace_client_main(client_id_t id, int argc, const char *argv[])
-{
-    std::cerr << "wrong drmemtrace_client_main\n";
-}
-
-/* This dr_client_main should be called instead of the one in tracer.cpp */
-DR_EXPORT void
-dr_client_main(client_id_t id, int argc, const char *argv[])
-{
-    std::cerr << "app dr_client_main\n";
-    drmemtrace_client_main(id, argc, argv);
-}
-
-#    ifdef __cplusplus
-}
-#    endif
-#endif /* UNIX && TEST_APP_DR_CLIENT_MAIN */
--- a/clients/drcachesim/tests/burst_flush_aarch64.cpp
+++ b/clients/drcachesim/tests/burst_flush_aarch64.cpp
@@ -69,7 +69,7 @@ int
 main(int argc, const char *argv[])
 {
     signal(SIGILL, sigill_handler);
-    if (!my_setenv("DYNAMORIO_OPTIONS", "-client_lib ';;-offline'"))
+    if (!my_setenv("DYNAMORIO_OPTIONS", "-signal_stack_size 64K -client_lib ';;-offline'"))
         std::cerr << "failed to set env var!\n";
 
     std::cerr << "pre-DR init\n";
--- a/clients/drcachesim/tracer/instru.cpp
+++ b/clients/drcachesim/tracer/instru.cpp
@@ -200,7 +200,7 @@ instru_t::is_aarch64_dcache_flush_op(instr_t *instr)
     if (instr_get_opcode(instr) != OP_sys)
         return false;
     switch (opnd_get_immed_int(instr_get_src(instr, 0))) {
-    // i#4328: Handle privileged dcache operations.
+    // TODO i#4406: Handle all privileged dcache operations.
     case DR_DC_IVAC:
     case DR_DC_CVAU:
     case DR_DC_CIVAC:
--- a/clients/drcachesim/tracer/instru.cpp
+++ b/clients/drcachesim/tracer/instru.cpp
@@ -188,6 +188,7 @@ instru_t::is_aarch64_icache_flush_op(instr_t *instr)
     if (instr_get_opcode(instr) != OP_sys)
         return false;
     switch (opnd_get_immed_int(instr_get_src(instr, 0))) {
+    // i#4328: Handle privileged icache operations.
     case DR_IC_IVAU: return true;
     }
     return false;
@@ -199,6 +200,8 @@ instru_t::is_aarch64_dcache_flush_op(instr_t *instr)
     if (instr_get_opcode(instr) != OP_sys)
         return false;
     switch (opnd_get_immed_int(instr_get_src(instr, 0))) {
+    // i#4328: Handle privileged dcache operations.
+    case DR_DC_IVAC:
     case DR_DC_CVAU:
     case DR_DC_CIVAC:
     case DR_DC_CVAC: return true;
--- a/clients/drcachesim/tracer/instru.cpp
+++ b/clients/drcachesim/tracer/instru.cpp
@@ -188,7 +188,7 @@ instru_t::is_aarch64_icache_flush_op(instr_t *instr)
     if (instr_get_opcode(instr) != OP_sys)
         return false;
     switch (opnd_get_immed_int(instr_get_src(instr, 0))) {
-    // i#4328: Handle privileged icache operations.
+    // TODO i#4406: Handle privileged icache operations.
     case DR_IC_IVAU: return true;
     }
     return false;
--- a/clients/drcachesim/tracer/instru.cpp
+++ b/clients/drcachesim/tracer/instru.cpp
@@ -181,6 +181,32 @@ instru_t::instr_to_prefetch_type(instr_t *instr)
     }
 }
 
+#ifdef AARCH64
+bool
+instru_t::is_aarch64_icache_flush_op(instr_t *instr)
+{
+    if (instr_get_opcode(instr) != OP_sys)
+        return false;
+    switch (opnd_get_immed_int(instr_get_src(instr, 0))) {
+    case DR_IC_IVAU: return true;
+    }
+    return false;
+}
+
+bool
+instru_t::is_aarch64_dcache_flush_op(instr_t *instr)
+{
+    if (instr_get_opcode(instr) != OP_sys)
+        return false;
+    switch (opnd_get_immed_int(instr_get_src(instr, 0))) {
+    case DR_DC_CVAU:
+    case DR_DC_CIVAC:
+    case DR_DC_CVAC: return true;
+    }
+    return false;
+}
+#endif
+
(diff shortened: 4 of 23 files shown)


abhinav92003:
There are some other cache flush operations recognised by DR that are not handled in this PR. https://github.com/DynamoRIO/dynamorio/blob/ce6ee1867817f2fdd5629cd1ccfa9511dffdef80/core/ir/aarch64/instr_create.h#L119
```
dc cisw
dc isw
dc csw
dc ivac
ic iallu
ic ialluis
```

These are defined only at EL1 and above (Eg https://developer.arm.com/docs/ddi0595/h/aarch64-system-instructions/ic-iallu).

Based on the discussion at https://github.com/DynamoRIO/dynamorio/pull/4386#issuecomment-665727841, I understand that we don't intend to support these yet, right? @derekbruening

derekbruening:
> Based on the discussion at [#4386 (comment)](https://github.com/DynamoRIO/dynamorio/pull/4386#issuecomment-665727841), I understand that we don't intend to support these yet, right?

We decode them and if it's easy to handle them in drcachesim we should just do it IMHO.  Is there complexity in handling them?  The virtual address ones should be easy: is it just the sets-ways ones that don't fit the current scheme?

clients/drcachesim/tracer/instru.cpp
```diff
@@ -181,17 +181,67 @@ instru_t::instr_to_prefetch_type(instr_t *instr)
...
+
+bool
+instru_t::is_aarch64_dcache_flush_op(instr_t *instr)
+{
+    if (instr_get_opcode(instr) != OP_sys)
+        return false;
+    switch (opnd_get_immed_int(instr_get_src(instr, 0))) {
+    case DR_DC_CVAU:
+    case DR_DC_CIVAC:
+    case DR_DC_CVAC: return true;
```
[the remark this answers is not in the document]
abhinav92003: Unlike other DC operations handled here, `DC ZVA` does not invalidate or clear the cache line. IIUC, it's not a cache flush operation.

 It just zeroes out a region of memory. So maybe it can be treated as a store?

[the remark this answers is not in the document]
derekbruening: It does seem like it could be treated as a regular store.  But it needs a proper start address (back-align its operand) and size (cache line size).  The easiest thing for a client would be to have that in the IR operand, instead of the encoded address: which bring up issues of re-encoding fidelity (xref #3339).  Grrr, another source of complexity.  I filed #4400.

On a similar topic: I'm thinking the size of the cache operations (here, and OP_clflush) should match OP_invlpg and have a 0 size, and that we should change opc_is_not_a_real_memory_load() to just look for zero-sized memory operand?  I think most tools (taint tracking is always the example I think of) would not want to think of these as reading a byte of memory: they should not interact w/ the taint of data I would think.  I filed #4401.

abhinav92003:
> > Based on the discussion at [#4386 (comment)](https://github.com/DynamoRIO/dynamorio/pull/4386#issuecomment-665727841), I understand that we don't intend to support these yet, right?
> 
> We decode them and if it's easy to handle them in drcachesim we should just do it IMHO. Is there complexity in handling them? The virtual address ones should be easy: is it just the sets-ways ones that don't fit the current scheme?

I'm not sure how to write test app for the EL1 and above operations. Including them in the existing test app gives the following as it's not defined at EL0.
```
Program received signal SIGILL, Illegal instruction.
_start () at fl.s:27
27	        ic iallu
```

derekbruening:
> > > Based on the discussion at [#4386 (comment)](https://github.com/DynamoRIO/dynamorio/pull/4386#issuecomment-665727841), I understand that we don't intend to support these yet, right?
> > 
> > 
> > We decode them and if it's easy to handle them in drcachesim we should just do it IMHO. Is there complexity in handling them? The virtual address ones should be easy: is it just the sets-ways ones that don't fit the current scheme?
> 
> I'm not sure how to write test app for the EL1 and above operations. Including them in the existing test app gives the following as it's not defined at EL0.
> 
> ```
> Program received signal SIGILL, Illegal instruction.
> _start () at fl.s:27
> 27	        ic iallu
> ```
> 
> Same for the remaining VA one:
> 
> ```
> Program received signal SIGILL, Illegal instruction.
> _start () at fl.s:23
> 23	         dc ivac, x0
> ```

The instrumentation is before the instruction so we would still include it in the trace, so you could wrap each in a try or longjmp or whatnot, for generation. (I think there's an issue on trying to not have it in the trace if the fault is a front-end fault...or maybe not...does not seem a high priority trace correctness issue.)

For the processing side, a synthetic trace could include these.

Or we could live with having a few opcodes with no regression test.

clients/drcachesim/tests/allasm_aarch64_flush.asm
```diff
@@ -0,0 +1,69 @@
...
+
+        adr      x0, helloworld
+        adr      x1, .
+
+        // data cache flush operations.
+        dc       civac, x0
+        dc       cvac, x0
+        dc       cvau, x0
+
+        // instruction cache flush operations.
```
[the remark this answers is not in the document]
abhinav92003: > The instrumentation is before the instruction so we would still include it in the trace, so you could wrap each in a try or longjmp or whatnot, for generation. (I think there's an issue on trying to not have it in the trace if the fault is a front-end fault...or maybe not...does not seem a high priority trace correctness issue.)
> 
> For the processing side, a synthetic trace could include these.
> 
> Or we could live with having a few opcodes with no regression test.

I verified locally that `DC_IVAC` gets counted as expected by DR (even though it causes a front-end fault), same as other VA-based flush instrs.

clients/drcachesim/tests/allasm_aarch64_flush.asm
```diff
@@ -0,0 +1,69 @@
...
+        mov      x0, sp
+        bic      x0, x0, #63
+        mov      x1, x0        // x1 is top of region
+        sub      x0, x0, #1024 // x0 is bottom of region
+        mov      sp, x0
+
+        adr      x0, helloworld
+        adr      x1, .
+
+        // data cache flush operations.
```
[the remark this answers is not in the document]
abhinav92003: Done.

clients/drcachesim/tests/allasm_aarch64_flush.asm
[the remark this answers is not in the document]
abhinav92003: Done.

clients/drcachesim/tests/allasm_aarch64_flush.asm
```diff
@@ -0,0 +1,69 @@
...
+        dc       civac, x0
+        dc       cvac, x0
+        dc       cvau, x0
```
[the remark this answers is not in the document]
abhinav92003: As discussed in other comment thread, `DC ZVA` is not a cache flush instruction. Will handle it in a separate PR.

clients/drcachesim/tracer/instru.cpp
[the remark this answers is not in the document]
abhinav92003: As this PR is for cache flush instructions, I'll create a new issue and PR for handling `DC ZVA`.

clients/drcachesim/tracer/instru.cpp
```diff
@@ -181,17 +181,67 @@ instru_t::instr_to_prefetch_type(instr_t *instr)
...
 }
 
+#ifdef AARCH64
+bool
+instru_t::is_aarch64_icache_flush_op(instr_t *instr)
+{
+    if (instr_get_opcode(instr) != OP_sys)
+        return false;
+    switch (opnd_get_immed_int(instr_get_src(instr, 0))) {
+    case DR_IC_IVAU: return true;
```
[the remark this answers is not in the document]
abhinav92003: I added `DR_DC_IVAC` here. Others will probably require more changes as they either don't have any operand (like `IC_IALLU`) or the operand is not a memref (like `DC_CSW`), and therefore they won't be handled by `instrument_memref` at all currently.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1268,7 +1274,7 @@ template <typename T> class trace_converter_t {
                 buf->type = instr->prefetch_type();
                 buf->size = 1;
             } else if (instr->is_flush()) {
-                buf->type = TRACE_TYPE_DATA_FLUSH;
+                buf->type = instr->flush_type();
```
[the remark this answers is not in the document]
abhinav92003: Done.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1268,7 +1274,7 @@ template <typename T> class trace_converter_t {
...
                 buf->size = (ushort)opnd_size_in_bytes(opnd_get_size(memref.opnd));
```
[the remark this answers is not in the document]
abhinav92003: Is this an issue for the VA based AArch64 flushes? Or only for the set/way ones, which aren't handled right now anyway.

abhinav92003:
> The instrumentation is before the instruction so we would still include it in the trace, so you could wrap each in a try or longjmp or whatnot, for generation. (I think there's an issue on trying to not have it in the trace if the fault is a front-end fault...or maybe not...does not seem a high priority trace correctness issue.)
> 
> For the processing side, a synthetic trace could include these.
> 
> Or we could live with having a few opcodes with no regression test.

I verified locally that DC_IVAC gets counted as expected by DR (even though it causes a front-end fault), same as other VA-based flush instrs.

clients/drcachesim/tracer/raw2trace.h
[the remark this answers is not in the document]
derekbruening: All of them.  x86 CLFLUSH may have the wrong size too?  Maybe only the ARM syscall in online mode (the original drcachesim target) has it right.

derekbruening:
> > The instrumentation is before the instruction so we would still include it in the trace, so you could wrap each in a try or longjmp or whatnot, for generation. (I think there's an issue on trying to not have it in the trace if the fault is a front-end fault...or maybe not...does not seem a high priority trace correctness issue.)
> > For the processing side, a synthetic trace could include these.
> > Or we could live with having a few opcodes with no regression test.
> 
> I verified locally that DC_IVAC gets counted as expected by DR (even though it causes a front-end fault), same as other VA-based flush instrs.

Just a manual check b/c your tests are all pure-asm?  If your tests are C/C++ it should be simple to add these, continuing past the SIGILLs.

clients/drcachesim/tracer/instru.cpp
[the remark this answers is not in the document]
derekbruening: There needs to be new trace entry type for the set-way ones.  If you could file an issue on that (and xref here).

clients/drcachesim/tracer/raw2trace.h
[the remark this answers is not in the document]
abhinav92003: `size` for x86 `clflush` is being set to 1 currently, yes. IIUC, for all currently added flush instructions, this needs to be set to the cache line size right? Is that available to DR?

clients/drcachesim/tracer/instru.cpp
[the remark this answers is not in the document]
abhinav92003: Opened #4406 for non-address cache flush operations.

clients/drcachesim/tracer/instru.cpp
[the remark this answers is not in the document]
abhinav92003: I can continue on `DC ZVA` work under #4400 itself.

clients/drcachesim/tracer/raw2trace.h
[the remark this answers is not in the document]
derekbruening: Yes: https://dynamorio.org/dynamorio_docs/dr__proc_8h.html#ad2add9b8aab5b68cb2362a1f53428453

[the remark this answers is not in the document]
abhinav92003: Oops, thanks for the pointer. Made the change.

abhinav92003:
> Just a manual check b/c your tests are all pure-asm? If your tests are C/C++ it should be simple to add these, continuing past the SIGILLs.

Ia added a C++ test that uses a SIGILL handler to continue past the invalid instruction. It passes in release build, but fails in debug with the following log:
```
abhinavas@tx1:~/dr/build/i4328-2_debug$ ctest -VV -R 'burst_flush'
UpdateCTestConfiguration  from :/home/abhinavas/dr/build/i4328-2_debug/DartConfiguration.tcl
Parse Config file:/home/abhinavas/dr/build/i4328-2_debug/DartConfiguration.tcl
UpdateCTestConfiguration  from :/home/abhinavas/dr/build/i4328-2_debug/DartConfiguration.tcl
Parse Config file:/home/abhinavas/dr/build/i4328-2_debug/DartConfiguration.tcl
Test project /home/abhinavas/dr/build/i4328-2_debug
Constructing a list of tests
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 186
    Start 186: code_api|tool.drcacheoff.burst_flush_aarch64

186: Test command: /usr/bin/cmake "-D" "precmd=foreach@/usr/bin/cmake@-E@remove_directory@drmemtrace.tool.drcacheoff.burst_flush_aarch64.*.dir" "-D" "cmd=/home/abhinavas/dr/build/i4328-2_debug/clients/bin64/tool.drcacheoff.burst_flush_aarch64" "-D" "postcmd=firstglob@/home/abhinavas/dr/build/i4328-2_debug/clients/bin64/drcachesim@-indir@drmemtrace.tool.drcacheoff.burst_flush_aarch64.*.dir@-simulator_type@basic_counts" "-D" "postcmd2=" "-D" "postcmd3=" "-D" "cmp=/home/abhinavas/dr/build/i4328-2_debug/suite/tests/offline-burst_flush_aarch64.expect" "-D" "code=" "-P" "/home/abhinavas/dr/src/i4328-2/suite/tests/runmulti.cmake"
186: Test timeout computed to be: 1500
186: Running |/usr/bin/cmake;-E;remove_directory /home/abhinavas/dr/build/i4328-2_debug/suite/tests/drmemtrace.tool.drcacheoff.burst_flush_aarch64.63075.3663.dir|
186: Running cmd |/home/abhinavas/dr/build/i4328-2_debug/clients/bin64/tool.drcacheoff.burst_flush_aarch64|
186: CMake Error at /home/abhinavas/dr/src/i4328-2/suite/tests/runmulti.cmake:106 (message):
186:   *** cmd failed (255): pre-DR init
186: 
186:   <unable to determine lib path for cross-arch execve>
186: 
186:   <Starting application
186:   /home/abhinavas/dr/build/i4328-2_debug/clients/bin64/tool.drcacheoff.burst_flush_aarch64
186:   (63088)>
186: 
186:   <Initial options = -client_lib ';;-offline' -stack_size 56K
186:   -signal_stack_size 32K -max_elide_jmp 0 -max_elide_call 0
186:   -no_inline_ignored_syscalls -native_exec_default_list ''
186:   -no_native_exec_managed_code -no_indcall2direct >
186: 
186:   pre-DR start
186: 
186:   <Application
186:   /home/abhinavas/dr/build/i4328-2_debug/clients/bin64/tool.drcacheoff.burst_flush_aarch64
186:   (63088).  DynamoRIO Cache Simulator Tracer internal crash at PC
186:   0x0000aaaaae4909fc.  Please report this at http://dynamorio.org/issues.
186:   Program aborted.
186: 
186:   Received SIGSEGV at client library pc 0x0000aaaaae4909fc in thread 63088
186: 
186:   Base: 0x0000aaaaae3ae000
186: 
186:   Registers:	eflags=0x0000000060000000
186: 
186:   version 8.0.18491, custom build
186: 
186:   -client_lib ';;-offline' -stack_size 56K -signal_stack_size 32K
186:   -max_elide_jmp 0 -max_elide_call 0 -no_inline_ignored_syscalls
186:   -native_exec_default_list '' -no_native_exec_managed_code
186:   -no_indcall2direct
186: 
186:   0x0000ffff176f3180 0x0000aaaaae6c40f8
186: 
186:   0x0000ffff176f55b0 0x0000aaaaae6c3d00
186: 
186:   0x0000ffff176f55c0 0x0000aaaaae6ccf38
186: 
186:   0x0000ffff176f5600 0x0000aaaaae6ccfd4
186: 
186:   0x0000ffff176f5640 0x0000aaaaae6e335c
186: 
186:   0x0000ffff176f5670 0x0000aaaaae6ebd78
186: 
186:   0x0000ffff176f56a0 0x0000aaaaae6e83d0
186: 
186:   0x0000ffff176f6860 0x0000aaaaae6eb57c
186: 
186:   0x0000ffff176f8bc0 0x0000aaaaae6c40f8
186: 
186:   0x0000ffff176faff0 0x0000aaaaae6ea90c
186: 
186:   0x0000ffff176f3180 0x0000aaaaae6c40f8
186: 
186:   0x0000ffff176f55b0 0x0000aaaaae6c3d00
186: 
186:   0x0000ffff176f55c0 0x0000aaaaae6ccf38
186: 
186:   0x0000ffff176f5600 0x0000aaaaae6ccfd4
186: 
186:   0x0000ffff176f5640 0x0000aaaaae6e335c>
186: 
186:   ***
186: 
186: Call Stack (most recent call first):
186:   /home/abhinavas/dr/src/i4328-2/suite/tests/runmulti.cmake:115 (process_cmdline)
186: 
186: 
1/1 Test #186: code_api|tool.drcacheoff.burst_flush_aarch64 ...***Failed    0.40 sec

Total Test time (real) =   1.18 sec

The following tests FAILED:
	186 - code_api|tool.drcacheoff.burst_flush_aarch64 (Failed)
Errors while running CTest
```

abhinav92003:
> I added a C++ test that uses a SIGILL handler to continue past the invalid instruction. It passes in release build, but fails in debug.

Running the test app (built with debug on) under gdb shows that for some reason the signal handler is not handling the SIGILL:
```
(gdb) run
Starting program: /home/abhinavas/dr/build/i4328-2_debug/clients/bin64/tool.drcacheoff.burst_flush_aarch64 
pre-DR init
<unable to determine lib path for cross-arch execve>
<Starting application /home/abhinavas/dr/build/i4328-2_debug/clients/bin64/tool.drcacheoff.burst_flush_aarch64 (68545)>
<Initial options = -client_lib ';;-offline' -stack_size 56K -signal_stack_size 32K -max_elide_jmp 0 -max_elide_call 0 -no_inline_ignored_syscalls -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct >
pre-DR start

Program received signal SIGILL, Illegal instruction.
0x0000aaaa6aaba448 in ?? ()
(gdb) q
```
Not sure how building with the debug option leads to this behaviour.

derekbruening:
> > I added a C++ test that uses a SIGILL handler to continue past the invalid instruction. It passes in release build, but fails in debug.
> 
> Running the test app (built with debug on) under gdb shows that for some reason the signal handler is not handling the SIGILL:
> 
> ```
> (gdb) run
> Starting program: /home/abhinavas/dr/build/i4328-2_debug/clients/bin64/tool.drcacheoff.burst_flush_aarch64 
> pre-DR init
> <unable to determine lib path for cross-arch execve>
> <Starting application /home/abhinavas/dr/build/i4328-2_debug/clients/bin64/tool.drcacheoff.burst_flush_aarch64 (68545)>
> <Initial options = -client_lib ';;-offline' -stack_size 56K -signal_stack_size 32K -max_elide_jmp 0 -max_elide_call 0 -no_inline_ignored_syscalls -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct >
> pre-DR start
> 
> Program received signal SIGILL, Illegal instruction.
> 0x0000aaaa6aaba448 in ?? ()
> (gdb) q
> ```
> 
> Not sure how building with the debug option leads to this behaviour.

From just the pasted output, I don't see any evidence that the app's signal handler is not doing what you expect?  gdb will still receive the first-chance signal, and DR will then pass it to the app.

abhinav92003:
> From just the pasted output, I don't see any evidence that the app's signal handler is not doing what you expect? gdb will still receive the first-chance signal, and DR will then pass it to the app.

If the sigill_handler was indeed working as intended, the app should've simply exit. Also, the `Program received signal SIGILL, Illegal instruction.` error is something I've seen when there's no SIGILL handler registered.

derekbruening:
> > From just the pasted output, I don't see any evidence that the app's signal handler is not doing what you expect? gdb will still receive the first-chance signal, and DR will then pass it to the app.
> 
> If the sigill_handler was indeed working as intended, the app should've simply exit. Also, the `Program received signal SIGILL, Illegal instruction.` error is something I've seen when there's no SIGILL handler registered.

OK, I was responding to what you pasted only, which was just that gdb saw a SIGILL, which is expected.

derekbruening:
> Also, the `Program received signal SIGILL, Illegal instruction.` error is something I've seen when there's no SIGILL handler registered.

If you mean *gdb* printing that, no, that has nothing to do with whether the app has a handler registered.

abhinav92003:
Sorry for the delay in getting back to this.

> If you mean _gdb_ printing that, no, that has nothing to do with whether the app has a handler registered.

Thanks. When I continue past the SIGILL (which is indeed handled as expected), there are some SIGSEGVs in drcachesim:
```
<Application /home/abhinavas/dr/build/i4328-2_debug/clients/bin64/tool.drcacheoff.burst_flush_aarch64 (85394).  DynamoRIO Cache Simulator Tracer internal crash at PC 0x0000aaaaaab8c9fc.  Please report this at http://dynamorio.org/issues.  Program aborted.
Received SIGSEGV at client library pc 0x0000aaaaaab8c9fc in thread 85394
```

Full details:

```
(gdb) r
Starting program: /home/abhinavas/dr/build/i4328-2_debug/clients/bin64/tool.drcacheoff.burst_flush_aarch64 
pre-DR init
<unable to determine lib path for cross-arch execve>
<Starting application /home/abhinavas/dr/build/i4328-2_debug/clients/bin64/tool.drcacheoff.burst_flush_aarch64 (85394)>
<Initial options = -client_lib ';;-offline' -stack_size 56K -signal_stack_size 32K -max_elide_jmp 0 -max_elide_call 0 -no_inline_ignored_syscalls -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct >
pre-DR start

Program received signal SIGILL, Illegal instruction.
0x0000aaaa6aaba448 in ?? ()
(gdb) backtrace
#0  0x0000aaaa6aaba448 in ?? ()
#1  0x0000000100000000 in ?? ()
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
(gdb) x/10i 0x0000aaaa6aaba448
=> 0xaaaa6aaba448:	dc	ivac, x0
   0xaaaa6aaba44c:	nop
   0xaaaa6aaba450:	add	sp, sp, #0x10
   0xaaaa6aaba454:	ldr	x1, [x28,#320]
   0xaaaa6aaba458:	ldr	x1, [x1]
   0xaaaa6aaba45c:	cbz	x1, 0xaaaa6aaba4e4
   0xaaaa6aaba460:	str	x0, [x28]
   0xaaaa6aaba464:	ldr	x0, [x28,#56]
   0xaaaa6aaba468:	str	x1, [x0]
   0xaaaa6aaba46c:	mov	x1, sp
(gdb) c
Continuing.

Program received signal SIGSEGV, Segmentation fault.
safe_read_asm_pre () at /home/abhinavas/dr/src/i4328-2/core/arch/aarch64/aarch64.asm:416
416	        strb     w3, [ARG1]
(gdb) backtrace
#0  safe_read_asm_pre () at /home/abhinavas/dr/src/i4328-2/core/arch/aarch64/aarch64.asm:416
#1  0x0000aaaaaadbfd00 in safe_read_fast (base=0xaafcfdc0, size=8, out_buf=0xffff37cab698, bytes_read=0x0) at /home/abhinavas/dr/src/i4328-2/core/arch/x86_code.c:430
#2  0x0000aaaaaadc8f38 in safe_read_ex (base=0xaafcfdc0, size=8, out_buf=0xffff37cab698, bytes_read=0x0) at /home/abhinavas/dr/src/i4328-2/core/unix/os.c:4573
#3  0x0000aaaaaadc8fd4 in d_r_safe_read (base=0xaafcfdc0, size=8, out_buf=0xffff37cab698) at /home/abhinavas/dr/src/i4328-2/core/unix/os.c:4592
#4  0x0000aaaaaaddf35c in sig_has_restorer (info=0xffff37c90148, sig=4) at /home/abhinavas/dr/src/i4328-2/core/unix/signal.c:2821
#5  0x0000aaaaaade7d78 in execute_handler_from_cache (dcontext=0xffff37c52000, sig=4, our_frame=0xffff37cafda0, sc_orig=0xffff37cada10, f=0xffff37c696c8, access_address=0x0) at /home/abhinavas/dr/src/i4328-2/core/unix/signal.c:5464
#6  0x0000aaaaaade43d0 in record_pending_signal (dcontext=0xffff37c52000, sig=4, ucxt=0xffff37cafe20, frame=0xffff37cafda0, forged=false, access_address=0x0) at /home/abhinavas/dr/src/i4328-2/core/unix/signal.c:4364
#7  0x0000aaaaaade757c in master_signal_handler_C (sig=4, siginfo=0xffff37cafda0, ucxt=0xffff37cafe20, xsp=0xffff37cafda0 "\004") at /home/abhinavas/dr/src/i4328-2/core/unix/signal.c:5309
#8  0x0000aaaaaadc00f8 in dynamorio_clone () at /home/abhinavas/dr/src/i4328-2/core/arch/aarch64/aarch64.asm:547
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
(gdb) c
Continuing.

Program received signal SIGSEGV, Segmentation fault.
0x0000aaaaaab8c9fc in d_r_notify (priority=SYSLOG_NONE, internal=false, synch=false, substitution_num=0, prefix=0x0, fmt=0x0) at /home/abhinavas/dr/src/i4328-2/core/utils.c:1884
1884	{
(gdb) backtrace
#0  0x0000aaaaaab8c9fc in d_r_notify (priority=SYSLOG_NONE, internal=false, synch=false, substitution_num=0, prefix=0x0, fmt=0x0) at /home/abhinavas/dr/src/i4328-2/core/utils.c:1884
#1  0x0000aaaaaade690c in master_signal_handler_C (sig=11, siginfo=0xffff37caa360, ucxt=0xffff37caa3e0, xsp=0xffff37caa360 "\v") at /home/abhinavas/dr/src/i4328-2/core/unix/signal.c:5126
#2  0x0000aaaaaadc00f8 in dynamorio_clone () at /home/abhinavas/dr/src/i4328-2/core/arch/aarch64/aarch64.asm:547
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
(gdb) c
Continuing.
<Application /home/abhinavas/dr/build/i4328-2_debug/clients/bin64/tool.drcacheoff.burst_flush_aarch64 (85394).  DynamoRIO Cache Simulator Tracer internal crash at PC 0x0000aaaaaab8c9fc.  Please report this at http://dynamorio.org/issues.  Program aborted.
Received SIGSEGV at client library pc 0x0000aaaaaab8c9fc in thread 85394
Base: 0x0000aaaaaaaaa000
Registers:	eflags=0x0000000060000000
version 8.0.18491, custom build
-client_lib ';;-offline' -stack_size 56K -signal_stack_size 32K -max_elide_jmp 0 -max_elide_call 0 -no_inline_ignored_syscalls -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct 
0x0000ffff37ca9180 0x0000aaaaaadc00f8
0x0000ffff37cab5b0 0x0000aaaaaadbfd00
0x0000ffff37cab5c0 0x0000aaaaaadc8f38
0x0000ffff37cab600 0x0000aaaaaadc8fd4
0x0000ffff37cab640 0x0000aaaaaaddf35c
0x0000ffff37cab670 0x0000aaaaaade7d78
0x0000ffff37cab6a0 0x0000aaaaaade43d0
0x0000ffff37cac860 0x0000aaaaaade757c
0x0000ffff37caebc0 0x0000aaaaaadc00f8
0x0000ffff37cb0ff0 0x0000aaaaaade690c
0x0000ffff37ca9180 0x0000aaaaaadc00f8
0x0000ffff37cab5b0 0x0000aaaaaadbfd00
0x0000ffff37cab5c0 0x0000aaaaaadc8f38
0x0000ffff37cab600 0x0000aaaaaadc8fd4
0x0000ffff37cab640 0x0000aaaaaaddf35c>
[Inferior 1 (process 85394) exited with code 0377]
(gdb) 


```

Unlike SIGILL, the backtrace for the SIGSEGVs doesn't go all the way back it seems.

derekbruening:
gdb is not very good at backtraces, esp w/ DR's separate stack.  Xref https://github.com/DynamoRIO/dynamorio/wiki/Debugging#call-stacks

The crash inside `d_r_notify` is weird: gdb thinks `fmt=0x0` but that's a literal so how could that be?  Either gdb is confused or sthg weird happened.  That's a printout for a safe_read, presumably the one that was just hit.  I would examine the code sequence and state there in more detail.

abhinav92003:
> gdb is not very good at backtraces, esp w/ DR's separate stack. Xref https://github.com/DynamoRIO/dynamorio/wiki/Debugging#call-stacks
> 
> The crash inside `d_r_notify` is weird: gdb thinks `fmt=0x0` but that's a literal so how could that be? Either gdb is confused or sthg weird happened. That's a printout for a safe_read, presumably the one that was just hit. I would examine the code sequence and state there in more detail.

The crashing SIGSEGV was caused by stack overflow in `d_r_notify` invoked at https://github.com/DynamoRIO/dynamorio/blob/70be2df2c17c4a55167715e2b3e95fdd3ff9ba19/core/unix/signal.c#L5126. As shown below, it occurred when register values were being pushed on the stack. 

```
5126	            SYSLOG_INTERNAL_WARNING_ONCE("(1+x) Handling our fault in a TRY at " PFX, pc);
(gdb) n
 
Program received signal SIGSEGV, Segmentation fault.
0x0000aaaaaab8ca74 in d_r_notify (priority=SYSLOG_NONE, internal=false, synch=false, substitution_num=0, prefix=0x0, fmt=0x0) at /home/abhinavas/dr/src/i4328-2/core/utils.c:1884
1884	{
(gdb) disassemble
Dump of assembler code for function d_r_notify:
   0x0000aaaaaab8ca70 <+0>:	sub	sp, sp, #0x900
=> 0x0000aaaaaab8ca74 <+4>:	stp	x29, x30, [sp,#-16]!
   0x0000aaaaaab8ca78 <+8>:	mov	x29, sp
   0x0000aaaaaab8ca7c <+12>:	str	w0, [x29,#76]
```

I was able to resolve it by increasing `signal_stack_size` for this test.

derekbruening:
> I was able to resolve it by increasing signal_stack_size for this test.

Note that a stack overflow on the dstack has dedicated detection code with a nice clear message to make debugging easy, but for the signal stack that goes out the window.

We need to figure out why this was hit here, when it appears to be nothing more than a two-level-deep signal frame nest, which you would think would happen on other tests as well and on real apps.  How does the signal frame size and default stack size compare to x86?  I would think the stack size would be larger if the machine has 64K pages (due to the default adjustments for page size).

Probably a separate issue should be filed.

abhinav92003:
> How does the signal frame size and default stack size compare to x86? I would think the stack size would be larger if the machine has 64K pages (due to the default adjustments for page size).

Page size seems to be =4K for both x86 and AArch64.
```
Breakpoint 1, adjust_defaults_for_page_size (options=0xaaaaaaf35000 <dynamo_options>) at /home/abhinavas/dr/src/i4328-2/core/options.c:255
255	    uint page_size = (uint)PAGE_SIZE;
(gdb) n
258	    if (page_size == 4096)
(gdb) print page_size
$1 = 4096
```
Due to this `adjust_defaults_for_page_size` doesn't make any adjustments.

Also, the final value computed for `signal_stack_size` after adjustments in `options_enable_code_api_dependences`, is 32K in both x86 and AArch64.

> Probably a separate issue should be filed.

Created #4425 for this.

abhinav92003:
> Created #4425 for this.
@derekbruening this PR should be good to go though right?

derekbruening:
> Also, the final value computed for signal_stack_size after adjustments in options_enable_code_api_dependences, is 32K in both x86 and AArch64.

It looks like it should be at least 56K:
```C
    options->stack_size = MAX(options->stack_size, 56 * 1024);
```

clients/drcachesim/tests/burst_flush_aarch64.cpp
```diff
@@ -0,0 +1,112 @@
...
+                   "-client_lib ';;-offline'"))
+        std::cerr << "failed to set env var!\n";
+
+    std::cerr << "pre-DR init\n";
+    dr_app_setup();
+    assert(!dr_app_running_under_dynamorio());
+    std::cerr << "pre-DR start\n";
+    dr_app_start();
+    assert(dr_app_running_under_dynamorio());
+    do_some_work();
```
derekbruening, line 83: This looks like a bug to the reader: there's no stop, and no comment explaining why not.

clients/drcachesim/tests/burst_flush_aarch64.cpp
```diff
@@ -0,0 +1,112 @@
...
+    return SetEnvironmentVariable(var, value) == TRUE;
+#endif
+}
+
+void
+sigill_handler(int signal)
+{
+    dr_app_stop_and_cleanup();
+    std::cerr << "all done\n";
+    _Exit(0);
```
derekbruening, line 56: This makes the test non-extensible: we can't add more cases; we can't combine it with other tests in one executable (to save CI time); etc.  IMHO this should use siglongjmp to continue, like all our other tests with privileged instructions.

clients/drcachesim/tests/burst_flush_aarch64.cpp
```diff
@@ -0,0 +1,112 @@
...
+    std::cerr << "pre-DR start\n";
+    dr_app_start();
+    assert(dr_app_running_under_dynamorio());
+    do_some_work();
+
+    return 0;
+}
+
+/* FIXME i#2099: the weak symbol is not supported on Windows. */
+#if defined(UNIX) && defined(TEST_APP_DR_CLIENT_MAIN)
```
derekbruening, line 89: This is not being used here so this sequence should just be removed.

clients/drcachesim/tests/burst_flush_aarch64.cpp
```diff
@@ -0,0 +1,112 @@
...
+#    ifdef __cplusplus
+extern "C" {
+#    endif
+
+/* Test if the drmemtrace_client_main() in drmemtrace will be called. */
```
derekbruening, line 94: I don't think we need another test of weak symbols: seems like this should be a regular export.

clients/drcachesim/tests/burst_flush_aarch64.cpp
```diff
@@ -0,0 +1,112 @@
...
+    _Exit(0);
+}
+
+static void
+do_some_work()
+{
+    int d = 0;
+    // Expected to send SIGILL.
+    // Control will be transferred to sigill_handler
+    __asm__ __volatile__("dc ivac, %0" : : "r"(d));
```
derekbruening, line 65: Do we need both the allasm and this test?  We could just put the 5 or so asm instructions here and delete the allasm test.  Move them into a single sequence between the start and stop with no function call if we want to reduce the (low-seeming) risk of the compiler putting its own flush instructions in a function call.

clients/drcachesim/tracer/instru.cpp
```diff
@@ -181,17 +181,70 @@ instru_t::instr_to_prefetch_type(instr_t *instr)
...
+#ifdef AARCH64
+bool
+instru_t::is_aarch64_icache_flush_op(instr_t *instr)
+{
+    if (instr_get_opcode(instr) != OP_sys)
+        return false;
+    switch (opnd_get_immed_int(instr_get_src(instr, 0))) {
+    // i#4328: Handle privileged icache operations.
+    case DR_IC_IVAU: return true;
+    }
```
derekbruening, line 193: Add a `TODO i#4406: xxxxxx` comment for the missing handling code here (else the reader thinks this is a bug b/c it looks like you think you handled them all).

abhinav92003:
> It looks like it should be at least 56K:

Shouldn't we be looking at `signal_stack_size` as the stack overflow was during signal handling? Increasing `signal_stack_size` indeed fixed the issue. For this option, the following sets it to 32K currently.
```
    options->signal_stack_size = MAX(options->signal_stack_size, 32 * 1024);
```

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1268,8 +1274,10 @@ template <typename T> class trace_converter_t {
                 buf->type = instr->prefetch_type();
                 buf->size = 1;
             } else if (instr->is_flush()) {
-                buf->type = TRACE_TYPE_DATA_FLUSH;
-                buf->size = (ushort)opnd_size_in_bytes(opnd_get_size(memref.opnd));
+                buf->type = instr->flush_type();
+                // i#4398: Handle flush sizes larger than ushort.
```
derekbruening, line 1278: style: as is this is just a descriptive comment of the code.  It needs a `TODO` or at least `XXX` so we know this is referring to unfinished work.  Ditto elsewhere.

suite/tests/CMakeLists.txt
```diff
@@ -2090,6 +2090,18 @@ if (DR_HOST_AARCH64)
     # with some tool chains (such as SUSE GCC 6.2) results in an executable that
     # is not properly static. Adding --no-export-dynamic avoids the problem.
     "-Wl,--no-export-dynamic")
+
+  add_exe(allasm_aarch64_flush
+    ${PROJECT_SOURCE_DIR}/clients/drcachesim/tests/allasm_aarch64_flush.asm
+    "-early_inject" "")
```
derekbruening, line 2096: early_inject is the default so no need to have it here (and we might drop support for late)

clients/drcachesim/tests/burst_flush_aarch64.cpp
```diff
@@ -0,0 +1,112 @@
...
+}
+
+int
+main(int argc, const char *argv[])
+{
+    signal(SIGILL, sigill_handler);
+    if (!my_setenv("DYNAMORIO_OPTIONS",
+                   "-stderr_mask 0xc -signal_stack_size 64K "
```
derekbruening, line 73: This `-signal_stack_size` needs explanation and needs a XXX comment for removing it as soon as we resolve the issue.

clients/drcachesim/tests/allasm-aarch64-prefetch-basic-counts.templatex
```diff
@@ -8,6 +8,8 @@ Total counts:
           81 total prefetches
      .* total data loads
      .* total data stores
+           0 total icache flushes
```
derekbruening, line 11: All these 0's may lead to flaky tests if some ld.so or crt startup code on some system has a flush instruction: it may be better to put .* like the other entries (maybe even .* for both lines and not just the numbers, to also cover future basic_counts additions for rare things none of these tests will have so we don't have to go update all these files again).

derekbruening:
> > It looks like it should be at least 56K:
> 
> Shouldn't we be looking at `signal_stack_size` as the stack overflow was during signal handling? Increasing `signal_stack_size` indeed fixed the issue. For this option, the following sets it to 32K currently.
> 
> ```
>     options->signal_stack_size = MAX(options->signal_stack_size, 32 * 1024);
> ```

Ah right, glanced at the wrong line.

clients/drcachesim/tests/burst_flush_aarch64.cpp
abhinav92003: Restructured. `dr_app_stop_and_cleanup` is in `main` now.

clients/drcachesim/tests/burst_flush_aarch64.cpp
abhinav92003: Modified test to use sigsetjmp/siglongjmp to resume execution.

I did something similar locally before, by calling `do_some_work` from `sigill_handler` repeatedly to resume execution. But I didn't push that for review as currently we have only a single instruction in the test. But yeah, it makes sense to do it now and make it easier for future contributors.

clients/drcachesim/tests/burst_flush_aarch64.cpp
abhinav92003: Removed.

clients/drcachesim/tests/burst_flush_aarch64.cpp
abhinav92003: Removed.

clients/drcachesim/tests/burst_flush_aarch64.cpp
abhinav92003: I deleted the offline version of the allasm test. But we still need the online version to verify the changes in clients/drcachesim/tracer/instru_online.cpp (as burst_flush_aarch64.cpp is only for offline)

clients/drcachesim/tracer/instru.cpp
abhinav92003: Done.

clients/drcachesim/tracer/raw2trace.h
abhinav92003: Done.

suite/tests/CMakeLists.txt

clients/drcachesim/tests/burst_flush_aarch64.cpp
abhinav92003: Done.

clients/drcachesim/tests/allasm-aarch64-prefetch-basic-counts.templatex
abhinav92003: I put .* to match all numbers. 

Position of new entries may be different; adding .* here preemptively might not cover that.

clients/drcachesim/tests/burst_flush_aarch64.cpp
```diff
@@ -37,39 +38,73 @@
...
 }
 
 static void
-do_some_work()
+dc_ivac()
 {
     int d = 0;
-    // Expected to send SIGILL.
+    // Expected to raise SIGILL.
     // Control will be transferred to sigill_handler
```
derekbruening, line 65: style nit: missing ending `.`

clients/drcachesim/tests/burst_flush_aarch64.cpp
```diff
@@ -37,39 +38,73 @@
...
-    std::cerr << "all done\n";
-    _Exit(0);
+    handled_sigill_count++;
+    siglongjmp(mark, handled_sigill_count);
+    exit(-1);
 }
 
 static void
-do_some_work()
+dc_ivac()
```
derekbruening, line 61: Either the longjmp should be inside this function, or there should be a comment here saying that this function raises SIGILL and it's the caller's responsibility to handle it.

clients/drcachesim/tests/burst_flush_aarch64.cpp
```diff
@@ -37,39 +38,73 @@
...
+{
+    dc_unprivileged_flush();
+    ic_unprivileged_flush();
+
+    // Testing privileged instructions requires handling SIGILL.
+    // We use sigsetjmp/siglongjmp to resume execution after handling the
+    // signal.
+    handled_sigill_count = 0;
+    int i = sigsetjmp(mark, 1);
+    switch (i) {
```
derekbruening, line 95: Either have a TODO comment about adding more cases, or this should just be an `if` for just one case.

abhinav92003: Added a TODO referencing #4406.

clients/drcachesim/tests/burst_flush_aarch64.cpp
abhinav92003: Added the comment.

clients/drcachesim/tests/burst_flush_aarch64.cpp
abhinav92003: Fixed.

clients/drcachesim/tests/burst_flush_aarch64.cpp
```diff
@@ -37,39 +38,73 @@
...
 
 void
 sigill_handler(int signal)
 {
-    dr_app_stop_and_cleanup();
-    std::cerr << "all done\n";
-    _Exit(0);
+    handled_sigill_count++;
+    siglongjmp(mark, handled_sigill_count);
+    exit(-1);
```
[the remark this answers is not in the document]
abhinav92003: Done.

Issue #4399, opened 2020-08-09
Add new ASSERT that only activates in debug builds.


johnfxgalea:
This a feature request to make a version of DR_ASSERT that automatically turns void during release builds.

Currently, DR_ASSERT and DR_ASSERT_MSG cannot be easily turned off for release builds. Unlike for <assert.h>, there is no "NDEBUG" version that turns DR asserts off. Therefore, a lot of code, particularly DR extensions, repeatedly resort to defining a new version of ASSERT to resolve the issue:

#ifdef DEBUG
#    define ASSERT(x, msg) DR_ASSERT_MSG(x, msg)
#else
#    define ASSERT(x, msg)
#endif

Add DR_ASSERT_DEBUG that is controllable with respect to the DEBUG def.

derekbruening:
There must have been some discussion long ago over having DR_ASSERT in release build; it does seem odd since "assert" usually means "debug only".

PR #4402, opened 2020-08-11
i#4399: Adds Debug Assertion

johnfxgalea:
Adds a DR_DEBUG_ASSERT which asserts only when DEBUG is defined.

Fixes: #4399

### DIFF
--- a/core/lib/instrument_api.h
+++ b/core/lib/instrument_api.h
@@ -1700,6 +1700,15 @@ dr_standalone_exit(void);
  */
 #    define DR_ASSERT(x) DR_ASSERT_MSG(x, "")
 
+/**
+ * The assertion is only set if DEBUG is defined.
+ */
+#    ifdef DEBUG
+#        define DR_DEBUG_ASSERT(x, msg) DR_ASSERT_MSG(x, msg)
+#    else
+#        define DR_DEBUG_ASSERT(x, msg)
+#    endif
+
 /* DR_API EXPORT END */
 
 DR_API
--- a/suite/tests/client-interface/drmgr-test.dll.c
+++ b/suite/tests/client-interface/drmgr-test.dll.c
@@ -343,6 +343,14 @@ dr_init(client_id_t id)
     CHECK(ok, "drmgr_unregister_kernel_xfer_event failed");
     ok = drmgr_register_kernel_xfer_event_ex(event_kernel_xfer, &priority);
     CHECK(ok, "drmgr_register_kernel_xfer_event_ex failed");
+
+    /* A quick check to ensure DR_DEBUG_ASSERT works on extensions. */
+#ifdef DEBUG
+    DR_DEBUG_ASSERT(true "should not fail");
+#else
+    /* Should not trigger failed assertion. */
+    DR_DEBUG_ASSERT(false, "should not fail");
+#endif
 }
 
 static void


johnfxgalea:
I'll modify the extensions to use this assertion once we agree on its format.

johnfxgalea:
run arm tests

johnfxgalea:
huh, appveyor seems to be unavailable?

johnfxgalea:
run arm tests

johnfxgalea:
Not sure if I am doing something wrong or if a compiler bug is present, but if I use the #ifdef #else, I get a compilation error. Separating the checks seems to work.

@derekbruening Any idea what is going on here please?

derekbruening:
> Not sure if I am doing something wrong or if a compiler bug is present, but if I use the #ifdef #else, I get a compilation error. Separating the checks seems to work.
> 
> @derekbruening Any idea what is going on here please?

Both genapi.pl and doxygen go through headers and interpret ifdefs: both have lists of what to consider defined; genapi has lists of what to take both sides of.  Probably it's one of those; probably genapi.

johnfxgalea:
@derekbruening Not sure if you want to take a preliminary look before I start changing the code of the extensions please. No rush.

derekbruening:
Should we deprecate DR_ASSERT, add an alias DR_CHECK or DR_RELEASE_ASSERT, recommend new code use that, and change all our own existing code to use that?  So then we'd have DR_DEBUG_ASSERT and DR_RELEASE_ASSERT (or DR_CHECK or sthg) and there's no confusion over "ASSERT" being in release build.

johnfxgalea:
> DR_ASSERT

Do you mean that DR_ASSERT, as a name, is potentially confusing to users and that is the reason why you want to deprecate it?

derekbruening:
> > DR_ASSERT
> 
> Do you mean that DR_ASSERT, as a name, is potentially confusing to users and that is the reason why you want to deprecate it?

Right: I think generally people expect an "assert" to be debug-only.  Or is that not always the case?

johnfxgalea:
Thinking about it now is to adjust slightly my PR to adopt the <assert.h> model, i.e., keep DR_ASSERT (and remove my DR_DEBUG_ASSERT) and toggle whether it asserts or does nothing via -DDR_NDEBUG

johnfxgalea:
> Thinking about it now, one way is to adjust slightly my PR to adopt the <assert.h> model, i.e., keep DR_ASSERT (and remove my DR_DEBUG_ASSERT) and toggle whether it asserts or does nothing via -DDR_NDEBUG

Ofc, the downside is that we break backwards compatibility.

derekbruening:
> Thinking about it now, one way is to adjust slightly my PR to adopt the <assert.h> model, i.e., keep DR_ASSERT (and remove my DR_DEBUG_ASSERT) and toggle whether it asserts or does nothing via -DDR_NDEBUG

Hmm, I originally assumed we'd have to leave the existing behavior alone, for compatibility: tools might rely on release-build checks at this point.  I think you would have to *not* turn on -DDR_NDEBUG for DEBUG=off in cmake then, right?  Or we could have a version check or a DR variable (like DynamoRIO_REG_COMPATIBILITY) to have it on for DEBUG=off for all recompiled code (but then someone just rebuilding would hit it).

johnfxgalea:
We treat DR_NDEBUG as a cflag option that is only set via the command-line. If the user does not set it, backwards compatibility of DR_ASSERT is maintained.

derekbruening:
> How about we treat DR_NDEBUG as a cflag option that is only set via the command-line. If the user does not set it when invoking cmake .., backwards compatibility of DR_ASSERT is maintained.

OK, and we audit all the uses in our own code base and replace ones we want in release with the release version, and we add DR_NDEBUG to package.cmake and runsuite.cmake, and document it of course?

johnfxgalea:
> replace ones we want in release with the release version,

Oh, so DR does rely on DR_ASSERT for release builds? That's right, I think I can recall some places, particularly in init functions associated with exts.

johnfxgalea:
> > How about we treat DR_NDEBUG as a cflag option that is only set via the command-line. If the user does not set it when invoking cmake .., backwards compatibility of DR_ASSERT is maintained.
> 
> OK, and we audit all the uses in our own code base and replace ones we want in release with the release version, and we add DR_NDEBUG to package.cmake and runsuite.cmake, and document it of course?

So just to confirm that we are on the same page please; we are okay with breaking the backwards compatibility of DR_ASSERT on new releases right? or do you think there is a neater way of doing this?

derekbruening:
> > > How about we treat DR_NDEBUG as a cflag option that is only set via the command-line. If the user does not set it when invoking cmake .., backwards compatibility of DR_ASSERT is maintained.
> > 
> > 
> > OK, and we audit all the uses in our own code base and replace ones we want in release with the release version, and we add DR_NDEBUG to package.cmake and runsuite.cmake, and document it of course?
> 
> So just to confirm that we are on the same page please; we are okay with breaking the backwards compatibility of DR_ASSERT on new releases right? or do you think there is a neater way of doing this?

By "new releases" you just mean our own uses of DR_ASSERT in the binaries we include in new release packages, and *not* changing behavior of other users who take our new release and recompile their own code?

johnfxgalea:
> > > > How about we treat DR_NDEBUG as a cflag option that is only set via the command-line. If the user does not set it when invoking cmake .., backwards compatibility of DR_ASSERT is maintained.
> > > 
> > > 
> > > OK, and we audit all the uses in our own code base and replace ones we want in release with the release version, and we add DR_NDEBUG to package.cmake and runsuite.cmake, and document it of course?
> > 
> > 
> > So just to confirm that we are on the same page please; we are okay with breaking the backwards compatibility of DR_ASSERT on new releases right? or do you think there is a neater way of doing this?
> 
> By "new releases" you just mean our own uses of DR_ASSERT in the binaries we include in new release packages, and _not_ changing behavior of other users who take our new release and recompile their own code?

Yes. Like you said before, we'll need to audit our use of DR_ASSERT.

derekbruening:
> > > > > How about we treat DR_NDEBUG as a cflag option that is only set via the command-line. If the user does not set it when invoking cmake .., backwards compatibility of DR_ASSERT is maintained.
> > > > 
> > > > 
> > > > OK, and we audit all the uses in our own code base and replace ones we want in release with the release version, and we add DR_NDEBUG to package.cmake and runsuite.cmake, and document it of course?
> > > 
> > > 
> > > So just to confirm that we are on the same page please; we are okay with breaking the backwards compatibility of DR_ASSERT on new releases right? or do you think there is a neater way of doing this?
> > 
> > 
> > By "new releases" you just mean our own uses of DR_ASSERT in the binaries we include in new release packages, and _not_ changing behavior of other users who take our new release and recompile their own code?
> 
> Yes. Like you said before, we'll need to audit our use of DR_ASSERT.

Yes, changing how some errors are reported in some of our own ext/ libraries seems fine.

derekbruening:
@johnfxgalea just going through the still-open PR's: this still seems like a useful change as I keep being surprised at these assert macros being enabled in release build; did you have plans to revive this?

Issue #4403, opened 2020-08-12
drmemtrace address elision ignoring non-memref writes, leading to incorrect addresses


derekbruening:
PR #3977 for #2001 put in elision of recorded addresses we can deduce in post-processing.  But it turns out that non-memory-accessing instructions are not properly considered when looking at register mods.  The tool.drcacheoff.burst_traceopts test which compares an optimized to un-optimized trace is just missing this pattern of modification.

I made a tool to measure how often this happens and the ratio on most apps is about 1 in 500 data addresses being wrong (typically they'll be off by some small constant depending on the reg mod in between).

Issue #975, opened 2014-11-28
re-add support for building as a static library


derekbruening:
_From [bruen...@google.com](https://code.google.com/u/109494838902877177630/) on November 13, 2012 14:24:59_

long, long ago in a galaxy far, far away, DR supported being built as a static library.  we now want to revive that support.

issues we need to address include:
- code that looks for DR library bounds
- code that makes .data read-only
- may be ok to leave the extra data sections

_Original issue: http://code.google.com/p/dynamorio/issues/detail?id=975_

derekbruening:
We're re-reviving this!

My proposal is to throw out the top-level STATIC_LIBRARY build type and instead have a side-by-side dynamorio_static library target that we create in every build, just like we do with the extension libraries.  This will make it easier to maintain and test.

derekbruening:
Since it's a pain to set DR runtime options, we should just commit to -code_api being on by default for dynamorio_static.

derekbruening:
I'm leaning toward combining proposals 1 and 2 above: we enable -code_api and assume that client code could be run at any time, even if there's no dr_init, but we also look for and call dr_init and dr_client_main (aside: should we drop dr_init for static to reduce collision risk with the app?)

derekbruening:
*** TODO for non-dr_init model, client will not receive init-time module events

Just document it: client should use module iterator after initializing DR.

*** TODO drop dr_init support for static b/c of risk of collision?

cs/ shows no exact match, but some close ones

*** TODO how set DR options

Can set env var

*** TODO how set client options

If client is not invoked by DR, options come from app.
If client has a dr_client_main: simplest to again have options come from
the app.

derekbruening:
I wish github had an auto-link feature other than on closing an issue.

Here is what I've done:

commit 4f70937d024cba812be5d93330a2c896207f3383

    i#975 static DR: rename create_thread to avoid name conflicts
    
    On Windows we have no solution to hide non-exported symbols in
    dynamorio_static like we do on Linux.  In our own tests the only conflict
    is create_thread.  We rename the core's to our_create_thread.  We have yet
    to find a great solution to this general issue on Windows.
    
commit 8db618085fbe331f58709a6e5fade485eaca87cc

    i#975 static DR: refactor STATIC_LIBRARY to a side-by-side target
    
    This is a big refactoring of the STATIC_LIBRARY top-level CMake
    configuration and separate build into instead using a side-by-side
    dynamorio_static target.
    
[...]

PR #4410, opened 2020-08-17
i#975 static DR: Add isolated GPR set class

derekbruening:
A std::unordered_set, even using dr_allocator_t, raises transparency
risks when statically linked on Windows (from lock functions and other
non-allocator resources).  We thus create our own resource-isolated
class to track GPR register inclusion for drmemtrace elision
optimizations.

Adds unit tests of the set class to burst_traceopts.

Further tested manually by ensuring the same number of elided
addresses is seen as with std::unordered_set in the burst_traceopts
test:
  [drmemtrace]: Reconstructed 622 elided addresses.

Even further tested by ensuring that raw2trace using the new set
correctly operates on a raw file that was created using
std::unordered_set.

Issue: #975, #4403

### DIFF
--- a/clients/drcachesim/tests/burst_traceopts.cpp
+++ b/clients/drcachesim/tests/burst_traceopts.cpp
@@ -215,7 +215,7 @@ reg_id_set_unit_tests()
     reg_id_set_t set;
     assert(set.begin() == set.end());
     // Test non-GPR.
-    auto insert_res = set.insert(DR_REG_XMM0);
+    auto insert_res = set.insert(IF_X86_ELSE(DR_REG_XMM0, DR_REG_Q0));
     assert(insert_res.first == set.end());
     assert(!insert_res.second);
     assert(set.begin() == set.end());
--- a/clients/drcachesim/tests/burst_traceopts.cpp
+++ b/clients/drcachesim/tests/burst_traceopts.cpp
@@ -50,6 +50,10 @@
 #    include <stdlib.h>
 #    include <string.h>
 
+/* Unit tests for classes used for trace optimizations. */
+void
+reg_id_set_unit_tests();
+
 /* Asm routines. */
 extern "C" {
 void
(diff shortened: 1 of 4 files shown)


clients/drcachesim/tests/burst_traceopts.cpp
```diff
@@ -209,9 +209,53 @@ gather_trace(const std::string &tracer_ops, const std::string &out_subdir)
...
+    assert(!insert_res.second);
+    assert(set.begin() == set.end());
+    // Test adding a GPR.
+    insert_res = set.insert(DR_REG_START_GPR + 1);
+    assert(insert_res.first != set.end());
+    assert(insert_res.second);
+    assert(set.begin() != set.end());
+    // Test find not-present.
+    auto find_res = set.find(DR_REG_START_GPR);
+    assert(find_res == set.end());
```
abhinav92003, line 229: Add an assert statement to show that `*find_res == DR_REG_INVALID` in this case. Adding a comment over `operator*` to document this will also be helpful.

clients/drcachesim/tests/burst_traceopts.cpp
```diff
@@ -209,9 +209,53 @@ gather_trace(const std::string &tracer_ops, const std::string &out_subdir)
...
+            if (*iter == DR_REG_START_GPR + 4)
+                found_next = true;
+            ++iter;
+        }
+    }
+    assert(found_next);
+    assert(set.find(DR_REG_START_GPR + 1) == set.end());
+    assert(set.find(DR_REG_START_GPR + 4) != set.end());
+    assert(*set.find(DR_REG_START_GPR + 4) == DR_REG_START_GPR + 4);
+}
```
abhinav92003, line 252: Perhaps add a test to show that a reg added multiple times, but removed only once, is indeed removed from the set.

clients/drcachesim/tests/burst_traceopts.cpp
```diff
@@ -209,9 +209,53 @@ gather_trace(const std::string &tracer_ops, const std::string &out_subdir)
...
+
 int
 main(int argc, const char *argv[])
 {
+    reg_id_set_unit_tests();
```
abhinav92003, line 257: Will it be useful to move this to a separate test file? I'm not aware how other such custom classes are tested currently.

derekbruening: core unit tests have a framework, but there aren't really any parallels to this outside the core.  This test covers trace opts, and this class is used for the opts, so it is a logical fit as part of the same set of tests.  With the upstream CI having limited resources, and in particular Appveyor being rather slow and each new separate executable launching slowly on Windows and affecting developer productivity, it's better for project usability to have tests combined into fewer executables (like googletest or other frameworks do, combining related tests into a single executable).  IMHO placing this here as part of the related tests is no loss in modularity and a win in the CI.

clients/drcachesim/tests/burst_traceopts.cpp
derekbruening: Done.

clients/drcachesim/tests/burst_traceopts.cpp
abhinav92003: Ack, thanks for explaining.

derekbruening:
PTAL

clients/drcachesim/tests/burst_traceopts.cpp
derekbruening: Actually a separate file but linked into this test would work and may be nicer for separating out the class test -- I will try that.

derekbruening: There is also clients/drcachesim/tests/drcachesim_unit_tests.cpp.  I moved this code into clients/drcachesim/tests/reg_id_set_unit_tests.cpp.  It is nicer that way.  +1.

derekbruening:
Appveyor is winxfer again, presumably output ordering: https://github.com/DynamoRIO/dynamorio/issues/4058#issuecomment-651255163.  Someone should fix that, or put it on the flaky list...

derekbruening:
Since nobody else is volunteering I made #4411.

PR #4412, opened 2020-08-18
i#1578 arm detach: Fix sigreturn crash by setting xsp

derekbruening:
Fixes a crash on detach on aarchxx by setting xsp to the proper value
for SYS_rt_sigreturn to find the signal frame for resuming app state.

Tested manually on aarch64 the api.detach test in gdb.  Unfortunately
there are further bugs, so the test cannot be enabled yet.

Issue: #1578

### DIFF
--- a/core/unix/signal.c
+++ b/core/unix/signal.c
@@ -7193,7 +7193,6 @@ typedef struct _sig_detach_info_t {
 #endif
 } sig_detach_info_t;
 
-/* xsp is only set for X86 */
 static void
 notify_and_jmp_without_stack(KSYNCH_TYPE *notify_var, byte *continuation, byte *xsp)
 {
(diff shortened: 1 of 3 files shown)


core/unix/signal.c
```diff
@@ -7232,18 +7231,22 @@ notify_and_jmp_without_stack(KSYNCH_TYPE *notify_var, byte *continuation, byte *
...
         asm("b dynamorio_condvar_wake_and_jmp");
 #endif
     } else {
         ksynch_set_value(notify_var, 1);
 #ifdef DR_HOST_NOT_TARGET
         ASSERT_NOT_REACHED();
 #elif defined(X86)
-        asm("mov %0, %%" ASM_XSP : : "m"(xsp));
         asm("mov %0, %%" ASM_XAX : : "m"(continuation));
+        asm("mov %0, %%" ASM_XSP : : "m"(xsp));
```
abhinav92003, line 7244: Why was this reordering required for x86?

abhinav92003 (approved):

core/unix/signal.c
derekbruening: That's me worried that gcc will use xsp to fetch the parameter (e.g., 32-bit, or maybe it put a copy on the stack and it's not still in a register and there's no frame pointer, etc.), and so it's safest to mess up xsp last.  I'll add a comment.

derekbruening: Done.

PR #4417, opened 2020-08-24
i#4335 OOM: Fix beyond-vmm reachability constraints bug

### LINKED ISSUE #4335: Allocation of more than 2GB in unreachable heap causes out of memory errror

derekbruening:
Fixes a bug where beyond-vmm memory was subject to reachability
constraints even when it did not have VMM_REACHABLE set.

Adds a test by running client.alloc's current OOM test with
-no_enable_reset, which hits OOM and so fails without this fix.

Fixes #4335

### DIFF
--- a/suite/tests/CMakeLists.txt
+++ b/suite/tests/CMakeLists.txt
@@ -2151,6 +2151,9 @@ if (CLIENT_INTERFACE)
       # For the subtest "targeting upper 2GB of low 4GB" we move -vm_base out of the
       # way.  We also raise the checklevel so we get memory filling for client allocs.
       "-no_vm_base_near_app -vm_base 0x120000000 -checklevel 3" "")
+    # A second test for i#4335's OOM despite a reset.  We leave checklevel for CI speed.
+    tobuild_ci(client.alloc-noreset client-interface/alloc.c ""
+      "-no_enable_reset -no_vm_base_near_app -vm_base 0x120000000" "")
     # XXX i#1312, i#3504: The test doesn't currently compile with CFLAGS_AVX512 but
     # attempts to test features based on it. The compile flag needs to be added.
     tobuild_ci(client.cleancall client-interface/cleancall.c "" "" "")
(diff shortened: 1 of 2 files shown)


suite/tests/CMakeLists.txt
```diff
@@ -2151,6 +2151,9 @@ if (CLIENT_INTERFACE)
       # For the subtest "targeting upper 2GB of low 4GB" we move -vm_base out of the
       # way.  We also raise the checklevel so we get memory filling for client allocs.
       "-no_vm_base_near_app -vm_base 0x120000000 -checklevel 3" "")
+    # A second test for i#4335's OOM despite a reset.  We leave checklevel for CI speed.
+    tobuild_ci(client.alloc-noreset client-interface/alloc.c ""
+      "-no_enable_reset -no_vm_base_near_app -vm_base 0x120000000" "")
```
abhinav92003, line 2156: I see some tests that explicitly turn on the `enable_reset` option. Is this option on or off by default?

Issue #4418, opened 2020-08-25
Support -vmheap_size > 2G


derekbruening:
Today, simply due to using the same static bitmap size for both vmcode and vmheap, we limit -vmheap_size to 2G.  We would like to allow increasing its size, to avoid things like #2115, to allow using guard pages with larger page sizes (xref #4415).

Xref #2115: beyond-vmheap slowdowns allocating memory
Xref #3570 on -vm_size 2G
Xref #4415 on parceling out sub-page pieces

PR #4419, opened 2020-08-25
i#4418 vmheap: Change sized options to use pointer-sized types

derekbruening:
Changes the "uint_size" option type to be pointer-sized, to better
handle larger values we are now wanting for things like -vmheap_size,
especially on ARM machines with larger page sizes.

Adds some sanity tests to ensure all the parsing and printing code was
updated.

Issue: #4418


derekbruening:
Friendly ping

PR #4429, opened 2020-08-31
i#4423: Split VMM stats into reachable vs unreachable

### LINKED ISSUE #4423: split "Peak vmm blocks" stats into vmcode and vmheap
The stats, esp the rstats, for vmm blocks should be updated to reflect which are in vmcode and which in vmheap, since that split (to be fair, the split was done recently, and the stats were added long long ago) is critical to understanding DR memory pressure:
```
                          Peak vmm blocks for heap :             14679
                         Peak vmm blocks for cache :              5139
                         Peak vmm blocks for stack :              9039
                  Peak vmm blocks for special heap :              4518
                  Peak vmm blocks for special mmap :              9043
```

derekbruening:
Given the vmcode vs vmheap split and completely different size limits
on the two, the VMM stats need to be split to be useful.  Here we
split the heap, special heap, and special mmap stats into reachable
and unreachable versions.  We also split the public versions in
dr_statistics_t, renaming and reordering fields, which is not
considered a compatibility break because we are in between releases.

Additionally tested on a large AArch64 app where the stats used to
look something like this:
                          Peak vmm blocks for heap :             14679
                         Peak vmm blocks for cache :              5139
                         Peak vmm blocks for stack :              9039
                  Peak vmm blocks for special heap :              4518
                  Peak vmm blocks for special mmap :              9043
Now they are more useful as this (different params so not perfect
numeric comparison) to see that vmcode is using 470M while vmheap is
at 2120M:
              Peak vmm blocks for unreachable heap :              8401
                         Peak vmm blocks for stack :             14586
      Peak vmm blocks for unreachable special heap :              3646
      Peak vmm blocks for unreachable special mmap :              7292
                Peak vmm blocks for reachable heap :              3650
                         Peak vmm blocks for cache :              3863
        Peak vmm blocks for reachable special mmap :                 5

Fixes #4423

### DIFF
--- a/core/lib/statsx.h
+++ b/core/lib/statsx.h
@@ -974,16 +974,28 @@ STATS_DEF("Peak total unallocatable free space", peak_total_wasted_vsize)
 STATS_DEF("Number of unaligned allocations (TEB's etc.)", unaligned_allocations)
 STATS_DEF("Peak unaligned allocations", peak_unaligned_allocations)
 #endif
-RSTATS_DEF("Current vmm blocks for heap", vmm_blocks_heap)
-RSTATS_DEF("Peak vmm blocks for heap", peak_vmm_blocks_heap)
-RSTATS_DEF("Current vmm blocks for cache", vmm_blocks_cache)
-RSTATS_DEF("Peak vmm blocks for cache", peak_vmm_blocks_cache)
+RSTATS_DEF("Current vmm blocks for unreachable heap", vmm_blocks_unreach_heap)
+RSTATS_DEF("Peak vmm blocks for unreachable heap", peak_vmm_blocks_unreach_heap)
 RSTATS_DEF("Current vmm blocks for stack", vmm_blocks_stack)
 RSTATS_DEF("Peak vmm blocks for stack", peak_vmm_blocks_stack)
-RSTATS_DEF("Current vmm blocks for special heap", vmm_blocks_special_heap)
-RSTATS_DEF("Peak vmm blocks for special heap", peak_vmm_blocks_special_heap)
-RSTATS_DEF("Current vmm blocks for special mmap", vmm_blocks_special_mmap)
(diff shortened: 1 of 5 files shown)


core/lib/statsx.h
```diff
@@ -974,16 +974,28 @@ STATS_DEF("Peak total unallocatable free space", peak_total_wasted_vsize)
 STATS_DEF("Number of unaligned allocations (TEB's etc.)", unaligned_allocations)
 STATS_DEF("Peak unaligned allocations", peak_unaligned_allocations)
 #endif
-RSTATS_DEF("Current vmm blocks for heap", vmm_blocks_heap)
-RSTATS_DEF("Peak vmm blocks for heap", peak_vmm_blocks_heap)
-RSTATS_DEF("Current vmm blocks for cache", vmm_blocks_cache)
-RSTATS_DEF("Peak vmm blocks for cache", peak_vmm_blocks_cache)
+RSTATS_DEF("Current vmm blocks for unreachable heap", vmm_blocks_unreach_heap)
+RSTATS_DEF("Peak vmm blocks for unreachable heap", peak_vmm_blocks_unreach_heap)
 RSTATS_DEF("Current vmm blocks for stack", vmm_blocks_stack)
```
abhinav92003, line 979: Similar to dr_stats_t, is it better to explicitly qualify stack and cache stats identifiers with reach/unreach for clarity. E.g. use vmm_blocks_unreach_stack instead of vmm_blocks_stack. 

Or if brevity is more important, maybe remove the reach/unreach qualifier from both.

derekbruening: Originally, my thought was that it would be more confusing to qualify since it would imply there's such a thing as the opposite reachability: hence the names here.  When updating dr_stats_t: I thought a user who does not know that stacks are unreachable and caches are reachable would want the qualification.  I still sort of like that split.  But maybe making them the same is better.  I'll qualify the internal too.

Issue #4424, opened 2020-08-27
Tune default options, esp. -guard_pages, for large-page machines, esp. AArch64


derekbruening:
On an AArch64 machine with 64K pages, the default guard pages around each heap, cache, and stack unit end up occupying a huge portion of the vmm, since many units are 64K or not much larger, making the guard space 2x as much as the used space.  For vmheap, with #4418, we now have 8G by default and maybe extra address space doesn't matter much since we can now arbitrarily increase the size, and every machine should have overcommit where untouched pages are (nearly) free.  But for vmcode, we have a 2G hard limit, and there this extra space can be problematic.

This option covers:
- Considering changing the meaning of -guard_pages to not mean around every unit, but in between *some* units, where we place them in between every Nth vmm alloc.  A tradeoff between the security/debugging benefit of guard pages and the space occupied.  Note that this may also be necessary for #4415 if the vmm starts handing out sub-page units, where it cannot then put guard pages around them but only around every N of them.
- Tuning the various heap and cache unit size parameters for large page sizes.  Today adjust_defaults_for_page_size() tries to do some of this but does not touch the cache sizes and should be re-evaluated.

For per-thread data structures, large-page wastage can really add up.  E.g., the stack, signal stack, heap, and sigpending allocs are often <64K (much less for heap and sigpending), yet today they can end up occupying 196K each, leading to a huge amount of sparsely used address space for an app with thousands of threads.  That's where we may want #4415.

derekbruening:
Xref #2607 on factoring guard pages out of internal sizing games for cache and heap units.

I'm thinking of something like this:

1) First, decouple guard page use from unit size by implementing #2607
   where guard pages are ignored in heap and cache unit sizing, to make it
   easier to have conditional guard pages b/c fewer places look at the
   guard page option.

   As part of this (or maybe a separate earlier PR) change Windows to use
   4K VMM blocks to make it consistent w/ UNIX and make adding guard pages
   to a 64K unit only cost the page sizes on a 4K-page system.

   We could say that we don't care about Windows 64K granularity: large vmheap, so
   pretty much always in reservation -- but the cases where we exceed the
   vmm are exactly where we want precision sizing for efficiency and guard
   pages for stability/security.  OTOH we could say we're desperate there
   and being inefficient is ok and we don't want to sacrifice the common,
   tuned cases for corner cases.

2) Use a VMM_GUARDED which_vmm_t flag as a way to have per-alloc control
[...]

PR #4430, opened 2020-09-01
i#2607: Externalize guard pages from units

### LINKED ISSUE #2607: remove guard page counts from heap and cache runtime option values
For heap and cache unit sizes specified in runtime options, we take guard pages
out of the requested size: so asking for 64K gives only 56K of usable space for
4K pages.  The original logic was tuned for Windows without -vm_reserve where
the OS allocation granularity matters and we don't want to waste space.

On UNIX, however, with the new 4K (or page size) vmm blocks, and with
-vm_reserve covering most allocations at least for smaller applications, the OS
granularity is less important: xref #2597.

Having the guards included makes it difficult to tune the default sizes based on
actual usage (and even more so when guard pages are sometimes turned off).  This
isssue covers making the heap and cache sizes like the cache sizes where the
guard pages are added on top of the requested size.  (In #2592 I removed the
debug-build STACK_GUARD_PAGE which was removing a page from the given stack size
to make a guard page: now it matches release where what you ask for is the
usable size you get, for stacks.)

derekbruening:
Decouples guard page use from cache and heap unit size.

Changes the Windows -vmm_block_size to the page size so that there is
nothing special about 64K for 4K pages.  The logic is that 99% of our
usage stays within the VMM reservations these days, so it is not worth
the complexity of dealing with 64K sizing for 4K pages.

Removes the guard page adjustments to cache and heap units.  Now the
specified sizes are the actual usable sizes, with any guard pages
added externally and not considered as part of the unit size.

Adjusts the default heap unit sizes down by 8K to maintain the same
sizing as before.

Leaves the default cache unit sizes how they were.  The sub-64K did
*not* have 8K subtracted and remain small.  The 64K will now occupy
72K, but shrinking to 56K is complicated by the quadrupling scheme
from 4K to 16K to 64K which previously ended up at 56K.

Issue: #2607, #4424
Fixes #2607

### DIFF
--- a/core/heap.c
+++ b/core/heap.c
@@ -855,7 +855,7 @@ vmm_place_vmcode(vm_heap_t *vmh, size_t size, heap_error_code_t *error_code)
                      get_random_offset(DYNAMO_OPTION(vm_max_offset) /
                                        DYNAMO_OPTION(vmm_block_size)) *
                          DYNAMO_OPTION(vmm_block_size));
-        preferred = ALIGN_FORWARD(preferred, DYNAMO_OPTION(vmm_block_size));
+        preferred = ALIGN_FORWARD(preferred, OS_ALLOC_GRANULARITY);
         /* overflow check: w/ vm_base shouldn't happen so debug-only check */
         ASSERT(!POINTER_OVERFLOW_ON_ADD(preferred, size));
         /* let's assume a single chunk is sufficient to reserve */
@@ -1887,6 +1887,7 @@ vmm_heap_alloc(size_t size, uint prot, heap_error_code_t *error_code, which_vmm_
 void
 vmm_heap_init()
 {
+    IF_WINDOWS(ASSERT(ALIGNED(OS_ALLOC_GRANULARITY, DYNAMO_OPTION(vmm_block_size))));
 #ifdef X64
     /* add reachable regions before we allocate the heap, xref PR 215395 */
     /* i#774, i#901: we no longer need the DR library nor ntdll.dll to be
--- a/core/heap.c
+++ b/core/heap.c
@@ -1887,7 +1887,6 @@ vmm_heap_alloc(size_t size, uint prot, heap_error_code_t *error_code, which_vmm_
 void
 vmm_heap_init()
 {
-    IF_WINDOWS(ASSERT(DYNAMO_OPTION(vmm_block_size) == OS_ALLOC_GRANULARITY));
 #ifdef X64
     /* add reachable regions before we allocate the heap, xref PR 215395 */
     /* i#774, i#901: we no longer need the DR library nor ntdll.dll to be
--- a/core/heap.c
+++ b/core/heap.c
@@ -166,31 +166,11 @@ DECLARE_NEVERPROT_VAR(static bool out_of_vmheap_once, false);
 #define MEMSET_HEADER(p, value) VARIABLE_SIZE(p) = HEAP_TO_PTR_UINT(value)
 #define GET_VARIABLE_ALLOCATION_SIZE(p) (VARIABLE_SIZE(p) + HEADER_SIZE)
 
-/* heap is allocated in units
- * we start out with a small unit, then each additional unit we
- * need doubles in size, up to a maximum, we default to 32kb initial size
- * (24kb useful with guard pages), max size defaults to 64kb (56kb useful with
- * guard pages), we keep the max small to save memory, it doesn't seem to be
- * perf hit! Though with guard pages we are wasting quite a bit of reserved
- * (though not committed) space */
-/* the only big things global heap is used for are pc sampling
- * hash table and sideline sampling hash table -- if none of those
- * are in use, 16KB should be plenty, we default to 32kb since guard
- * pages are on by default (gives 24kb useful) max size is same as for
- * normal heap units.
- */
-/* the old defaults were 32kb (usable) for initial thread private and 16kb
- * (usable) for initial global, changed to simplify the logic for allocating
- * in multiples of the os allocation granularity.  The new defaults prob.
- * make more sense with the shared cache then the old did anyways.
- */
-/* restrictions -
- * any guard pages are included in the size, size must be > UNITOVERHEAD
- * for best performance sizes should be of the form
- * 2^n * page_size (where n is a positve integer) and max should be a multiple
- * of the os allocation granularity so that with enough doublings we are
- * reserving memory in multiples of the allocation granularity and not wasting
- * any virtual address space (beyond our guard pages)
+/* The heap is allocated in units.
+ * We start out with a small unit. Then each additional unit we
+ * need doubles in size, up to a maximum.
+ * We keep the initial units small for thread-private heaps, since with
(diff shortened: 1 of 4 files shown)


abhinav92003:
There seem to be some Windows `DynamoRIO debug check failure` issues on Appveyor that need fixing. Continuing my review of other changes in the meantime.

core/heap.c
```diff
@@ -166,31 +166,11 @@ DECLARE_NEVERPROT_VAR(static bool out_of_vmheap_once, false);
...
-/* heap is allocated in units
- * we start out with a small unit, then each additional unit we
- * need doubles in size, up to a maximum, we default to 32kb initial size
- * (24kb useful with guard pages), max size defaults to 64kb (56kb useful with
- * guard pages), we keep the max small to save memory, it doesn't seem to be
- * perf hit! Though with guard pages we are wasting quite a bit of reserved
- * (though not committed) space */
-/* the only big things global heap is used for are pc sampling
- * hash table and sideline sampling hash table -- if none of those
- * are in use, 16KB should be plenty, we default to 32kb since guard
```
abhinav92003, line 178: This removed comment mentions that 16KB initial global heap is sufficient, but we defaulted to 32KB because of guard pages. Since now guard page is not part of the heap size, perhaps we can use 16KB as the initial global heap size now?

core/heap.c
```diff
@@ -166,31 +166,11 @@ DECLARE_NEVERPROT_VAR(static bool out_of_vmheap_once, false);
...
- */
-/* the old defaults were 32kb (usable) for initial thread private and 16kb
- * (usable) for initial global, changed to simplify the logic for allocating
- * in multiples of the os allocation granularity.  The new defaults prob.
- * make more sense with the shared cache then the old did anyways.
- */
-/* restrictions -
- * any guard pages are included in the size, size must be > UNITOVERHEAD
- * for best performance sizes should be of the form
- * 2^n * page_size (where n is a positve integer) and max should be a multiple
```
abhinav92003, line 190: initial_heap_unit_size is set to 24 * 1024 in this PR. This old comment suggests that might not be optimal.

core/heap.c
derekbruening: No, this is a very old, stale comment from the days of thread-private-only caches.  Today there are many things in the global heap.

core/heap.c
derekbruening: initial_heap_unit_size is the same in this PR as it was before: 24K plus two guard pages if enabled.  The doubling behavior will work out b/c it will hit the max and take that value no matter what the doubling would result in.  But, we could shrink it to the power-of-2 16K, which might be better anyway: these days the per-thread units are rarely used (well, for -disable_traces anyway, and for many-threaded apps even with traces the traces are likely to be built by other threads) and should be kept small.  I tried to avoid changing the actual usable size in the PR to minimize changes: but maybe shrinking to 16K for more "natural" doubling is good.

derekbruening:
Looks like the vmm itself is aligned to the vmm block size instead of the OS granularity so that needs to change.  If it's more than a few lines maybe it is best to split that to a separate PR.

core/heap.c
derekbruening: Given that the global and the nonpers init sizes are 24K too, I'd rather have fewer changes in this PR so I left this as 24K.

derekbruening:
PTAL.  The Windows fix was a one-liner (well in local test spot-checks anyway: we'll see if Appveyor is different) so leaving it in this PR.

core/heap.c
abhinav92003: Sure, makes sense to keep changes in each PR minimal. 

Maybe a TODO if you think the additional change might be valuable in future.

derekbruening: > Maybe a TODO if you think the additional change might be valuable in future.

No, the 2^n under the max is not important and does not really help: better to save space.

derekbruening:
Was there a requested change?  If not, normally an Approval would be expected (w/ or w/o a few minor tweaks: see the review conventions where we do not go to a fixed point of zero comments).

derekbruening:
The increased cache unit sizes (64K usable, vs the old 56K usable but listed as 64K b/c of the guards) seemed to result in a non-deterministic assert on 4 of the drcachesim tests on Appveyor.  I filed it as #4433, but I could not reproduce it locally, and even on Appveyor it's a <50% repro.  I put the sizes back to 56K (so explicit 56K in option list) as a workaround: so now no usable sizes change here.

derekbruening:
Well, on an AArch64 machine with 64K pages, shrinking to 56K results in this assert:
```
#1  0x0000fffff7c4e96c in d_r_internal_error (file=0xfffff7efa028 "src/core/fcache.c", line=1331, 
    expr=0xfffff7efa910 "ALIGNED(size, PAGE_SIZE)") at src/core/utils.c:176
#2  0x0000fffff7bf8d78 in fcache_create_unit (dcontext=0xfffdb3b54560, cache=0xfffdb3db1fa0, pc=0x0, size=57344)
    at src/core/fcache.c:1331
#3  0x0000fffff7bfb934 in fcache_cache_init (dcontext=0xfffdb3b54560, flags=0, initial_unit=true) at src/core/fcache.c:1594
#4  0x0000fffff7c185c8 in get_cache_for_new_fragment (dcontext=0xfffdb3b54560, f=0xfffdb3db1f00) at src/core/fcache.c:3627
#5  0x0000fffff7c186f4 in fcache_add_fragment (dcontext=0xfffdb3b54560, f=0xfffdb3db1f00) at src/core/fcache.c:3644
#6  0x0000fffff7bd4c70 in fragment_create (dcontext=0xfffdb3b54560, tag=0xfffff7b27a18 "\"|@\223\b\006\200\322\341\003", body_size=32, direct_exits=2, indirect_exits=0, 
    exits_size=56, flags=4096) at src/core/fragment.c:2447
#7  0x0000fffff7c4a7d0 in emit_fragment_common (dcontext=0xfffdb3b54560, tag=0xfffff7b27a18 "\"|@\223\b\006\200\322\341\003", ilist=0xfffdb3da0be0, flags=4096, 
    vmlist=0xfffdb3c744d0, link_fragment=true, add_to_htable=true, replace_fragment=0x0) at src/core/emit.c:581
#8  0x0000fffff7c4e338 in emit_fragment_ex (dcontext=0xfffdb3b54560, tag=0xfffff7b27a18 "\"|@\223\b\006\200\322\341\003", ilist=0xfffdb3da0be0, flags=4096, 
    vmlist=0xfffdb3c744d0, link=true, visible=true) at src/core/emit.c:864
#9  0x0000fffff7e64a78 in build_basic_block_fragment (dcontext=0xfffdb3b54560, start=0xfffff7b27a18 "\"|@\223\b\006\200\322\341\003", initial_flags=0, link=true, 
    visible=true, for_trace=false, unmangled_ilist=0x0) at src/core/arch/interp.c:5167
#10 0x0000fffff7c373b4 in d_r_dispatch (dcontext=0xfffdb3b54560) at src/core/dispatch.c:214
```

#4424 will fix this (not sure how to do a dedicated fix...it would look like much of #4424).

PR #4432, opened 2020-09-02
i#2607: debug appveyor assert

derekbruening:
Trying to debug an assert seen in PR #4430


derekbruening:
Filed #4433 on the assert.  Not going after it now (apparently it's been a latent issue in the code for a long time): closing this PR.

Issue #4400, opened 2020-08-11
Use the actual written address range in DC ZVA's memory operand


derekbruening:
Unlike the other DC cache operations, DC ZVA is not a cache management
operation but more of a regular store instruction.  But clients analyzing
it need its actual start address (back-aligned) and size (cache line size).
The easiest thing for a client would be to have that in the IR operand,
instead of the encoded address: though that brings up issues of re-encoding
fidelity (xref #3339).

abhinav92003:
Another complexity with the existing handling of `DC ZVA`: 
it is currently counted as a load operation, as specified by the codec here: https://github.com/DynamoRIO/dynamorio/blob/d179619df14df7762651d38405c25adbc8b5602a/core/ir/aarch64/codec.txt#L282
I tried adding another codec rule, specific to `DC ZVA`, that considers `memx0` as a destination instead. Something like
```110101010000101101110100001xxxxx  sys    memx0 : sysops``` 
but the pattern overlaps with the existing pattern for `sys`

We might have to either:
1. improve the codec to allow overlapping patterns: in case of ambiguity, give precedence to the first matching one.
2. split the pattern for sys into 127 multiple patterns, one for each of the variants of sys.(https://developer.arm.com/docs/ddi0595/h/aarch64-system-instructions)

derekbruening:
> We might have to either:
> 
> 1. improve the codec to allow overlapping patterns: in case of ambiguity, give precedence to the first matching one.
> 2. split the pattern for sys into multiple mutually exclusive patterns for its variants.(127 variants: https://developer.arm.com/docs/ddi0595/h/aarch64-system-instructions)

Improving the codec is #4393, and splitting up SYS is a key motivator for doing so.

abhinav92003:
> The easiest thing for a client would be to have that in the IR operand,
instead of the encoded address

Just checking my understanding: this IR change would require adding a new member in `_opnd_t` to denote back-alignment, here: https://github.com/DynamoRIO/dynamorio/blob/0762e11b8b5f3f5e8851f0d2f699b7f80a446d49/core/ir/opnd.h#L1413 
and then use this in https://github.com/DynamoRIO/dynamorio/blob/0762e11b8b5f3f5e8851f0d2f699b7f80a446d49/ext/drutil/drutil.c#L385 during the address calculation right?

Also,
[...]

PR #4440, opened 2020-09-08
i#4400: Back align DC ZVA memory reference before storing

abhinav92003:
Issue: #4400

### DIFF
--- a/clients/drcachesim/tests/allasm_aarch64_dc_zva.asm
+++ b/clients/drcachesim/tests/allasm_aarch64_dc_zva.asm
@@ -1,0 +1,89 @@
+/* **********************************************************
+ * Copyright (c) 2020 Google, Inc. All rights reserved.
+ * Copyright (c) 2016 ARM Limited. All rights reserved.
+ * **********************************************************/
+
+/*
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice,
+ *   this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of ARM Limited nor the names of its contributors may be
+ *   used to endorse or promote products derived from this software without
+ *   specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL ARM LIMITED OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+        .global  _start
+
+        .align   7
+_start:
+        // Align stack pointer and get some space.
+        mov      x0, sp
+        bic      x0, x0, #63
+        mov      x1, x0            // x1 is top of region
+        sub      x0, x0, #1024     // x0 is bottom of region
+        mov      sp, x0
+
+        adr      x1, second_str
+        dc       zva, x1
+
+        mov      w0, #1            // stdout
+        mov      w8, #64           // SYS_write
+        adr      x1, first_str
+        mov      w2, #13           // sizeof(first_str)
+        svc      #0
+
+        mov      w0, #1            // stdout
+        mov      w8, #64           // SYS_write
+        adr      x1, second_str
+        mov      w2, #14           // sizeof(second_str)
+        svc      #0
+
+        mov      w0, #1            // stdout
+        mov      w8, #64           // SYS_write
+        adr      x1, third_str
+        mov      w2, #13           // sizeof(third_str)
+        svc      #0
+
+        mov      w0, #1            // stdout
+        mov      w8, #64           // SYS_write
+        adr      x1, helloworld
+        mov      w2, #14           // sizeof(helloworld)
+        svc      #0
+
+
+        // Exit.
+        mov      w0, #0            // status
+        mov      w8, #94           // SYS_exit_group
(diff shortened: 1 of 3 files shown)


abhinav92003:
I verified locally that the `dc zva` memref is indeed back-aligned after these changes. I used the `-LL_miss_file` option to look at the memory addresses, as shown below.
Online:
```
abhinavas@tx1:~/dr/build/i4400-1$ ./bin64/drrun -t drcachesim  -simulator_type cache -LL_miss_file online_{no}backalign.gz  -- suite/tests/bin/allasm_aarch64_dc_zva 
abhinavas@tx1:~/dr/build/i4400-1$ gunzip online_{no}backalign.gz 
abhinavas@tx1:~/dr/build/i4400-1$ diff online_nobackalign online_backalign
2c2
< 0x400198,0x41020d
---
> 0x400198,0x410200
```
Offline:
```
abhinavas@tx1:~/dr/build/i4400-1$ rm -r drmemtrace.*;  ./bin64/drrun -t drcachesim       -offline suite/tests/bin/allasm_aarch64_dc_zva ; ./bin64/drrun -t drcachesim -indir drmemtrace* -simulator_type cache -LL_miss_file offline_{no}backalign.gz
abhinavas@tx1:~/dr/build/i4400-1$ gunzip offline_{no}backalign.gz 
abhinavas@tx1:~/dr/build/i4400-1$ diff offline_nobackalign offline_backalign 
2c2
< 0x400198,0x41020d
---
> 0x400198,0x410200
```
The instruction at `0x400198` is the `dc zva`, as expected
`0x0000000000400198  d50b7421   sys    $0x1ba1 (%x1)[1byte]`

But it seems tricky to write a test for it. Any suggestions?

derekbruening:
> But it seems tricky to write a test for it. Any suggestions?

Make an offline trace test that contains its own trace analyzer and decodes the instructions in the trace looking for DC ZVA and checking its address.  This decoding is easy b/c it's the same address space (see e.g. clients/drcachesim/tests/burst_traceopts.cpp which has its own analyzer).  I would add an analyzer to burst_flush_aarch64, add DC ZVA to its assembly, and have the analyzer check both the flush types and the ZVA.

clients/drcachesim/tests/allasm_aarch64_dc_zva.asm
```diff
@@ -0,0 +1,89 @@
+/* **********************************************************
```
derekbruening, line 1: The commit description is missing a body (https://github.com/DynamoRIO/dynamorio/wiki/Code-Reviews#commit-messages) (even if it's relatively redundant with the title)

clients/drcachesim/tests/allasm_aarch64_dc_zva.asm
```diff
@@ -0,0 +1,89 @@
...
+        mov      w0, #1            // stdout
+        mov      w8, #64           // SYS_write
+        adr      x1, second_str
+        mov      w2, #14           // sizeof(second_str)
+        svc      #0
+
+        mov      w0, #1            // stdout
+        mov      w8, #64           // SYS_write
+        adr      x1, third_str
+        mov      w2, #13           // sizeof(third_str)
```
derekbruening, line 63: What are all these strings for?  Comments would help.

clients/drcachesim/tests/allasm_aarch64_dc_zva.asm
```diff
@@ -0,0 +1,89 @@
...
+        svc      #0
+
+        mov      w0, #1            // stdout
+        mov      w8, #64           // SYS_write
```
derekbruening, line 67: The expected output file is missing?

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -918,8 +949,30 @@ instrument_memref(void *drcontext, user_data_t *ud, instrlist_t *ilist, instr_t
...
         insert_load_buf_ptr(drcontext, ilist, where, reg_ptr);
-    adjust = instru->instrument_memref(drcontext, ilist, where, reg_ptr, adjust, app, ref,
-                                       ref_index, write, pred);
+
+#ifdef AARCH64
+    // TODO i#4400: Use a new operand type to back-align zeroed address, instead of
+    // doing it in drcachesim.
+    if (is_dc_zva_instr(app)) {
+        reg_id_t reg_ba_addr;
+        if (drreg_reserve_register(drcontext, ilist, where, NULL, &reg_ba_addr) !=
```
derekbruening, line 958: This is failing to use the predicate set inside instru->instrument_memref.  I suppose A64 has no predicates, but it feels like it's not the model way to do it: someone will mirror this for some A32 code and fail to notice the bug, assuming this code here has no flaws.  Plus, if it could be elided, this skips all of that code: with the same arguments.  It seems best to be inside the instru_t classes and not out here.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -918,8 +949,30 @@ instrument_memref(void *drcontext, user_data_t *ud, instrlist_t *ilist, instr_t
...
+            DRREG_SUCCESS) {
+            FATAL("Fatal error: failed to reserve register for back-aligned DC ZVA "
+                  "memory reference\n");
+        }
+        opnd_t back_aligned_memref = back_align_dc_zva_memref_opnd(
+            drcontext, ilist, where, ref, proc_get_cache_line_size(), reg_ba_addr);
+        adjust = instru->instrument_memref(drcontext, ilist, where, reg_ptr, adjust, app,
+                                           back_aligned_memref, ref_index, write, pred);
+
+        if (drreg_unreserve_register(drcontext, ilist, where, reg_ba_addr) !=
```
derekbruening, line 968: For offline, this is all unnecessary extra register spilling and an and-ing: this is most optimally done in post-processing.  (If there were no runtime cost, then yes better to share the same code for online and offline; but given how common this instruction seems to be, we do not want to pay for all this extra runtime cost when it's so easy to avoid it).

In fact, given that this is supposed to be a temporary fix, and that our usage cares less about online, you could argue for putting the workaround *only* in post-processing and having online live with the bug until the proper fix is in place (no need for a separate .asm test then either).

suite/tests/CMakeLists.txt
```diff
@@ -2101,6 +2101,17 @@ if (DR_HOST_AARCH64)
...
+
+  add_exe(allasm_aarch64_dc_zva
+    ${PROJECT_SOURCE_DIR}/clients/drcachesim/tests/allasm_aarch64_dc_zva.asm "" "")
+  set_target_properties(allasm_aarch64_dc_zva PROPERTIES LINKER_LANGUAGE C)
+  append_link_flags(allasm_aarch64_dc_zva
+    "-nostartfiles -nodefaultlibs -static")
+  append_property_string(TARGET allasm_aarch64_dc_zva LINK_FLAGS
+    # The default value of CMAKE_SHARED_LIBRARY_LINK_C_FLAGS has -rdynamic, which
+    # with some tool chains (such as SUSE GCC 6.2) results in an executable that
+    # is not properly static. Adding --no-export-dynamic avoids the problem.
```
derekbruening, line 2113: This long comment is repeated multiple times: probably best to have the duplicates point at the master rather than copy-pasting the entire thing in many places.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -898,6 +898,37 @@ insert_filter_addr(void *drcontext, instrlist_t *ilist, instr_t *where, user_dat
     return reg_idx;
 }
 
+#ifdef AARCH64
+static bool
+is_dc_zva_instr(instr_t *instr)
+{
+    return instr_get_opcode(instr) == OP_sys &&
```
derekbruening, line 905: I would expect a XXX or TODO i# comment about this being non-ideal and that we want a proper opcode (I guess the IR back-aligning supercedes this...even so...)

clients/drcachesim/tracer/tracer.cpp
abhinav92003: Oh okay, makes sense. As that implementation is completely different than this PR, I've opened a new one #4444 for that.

abhinav92003:
Closing this PR in favour of #4444. As discussed in comments on this PR, we're okay with a solution that covers only offline traces, as it's simpler and temporary anyway.

Issue #3320, opened 2018-12-21
ASSERT type_is_instr in multiple drcachesim tests (tool.reuse_distance,tool.drcachesim.warmup-zeros,tool.drcachesim.missfile) on appveyor


derekbruening:
The 32-bit version failed once:
https://ci.appveyor.com/project/DynamoRIO/dynamorio/builds/21162519
```
[00:09:36] test 213
[00:09:36]         Start 213: code_api|tool.reuse_distance
[00:09:36] 
[00:09:36] 213: Test command: C:\projects\dynamorio\build\build_debug-internal-32\bin32\drrun.exe "-s" "90" "-quiet" "-debug" "-use_dll" "C:/projects/dynamorio/build/build_debug-internal-32/lib32/debug/dynamorio.dll" "-exit0" "-stderr_mask" "0xC" "-msgbox_mask" "0" "-dumpcore_mask" "0x7d" "-staged" "-code_api" "-t" "drcachesim" "-ipc_name" "drtestpipe_reuse_distance" "-simulator_type" "reuse_distance" "-reuse_distance_threshold" "256" "--" "C:/projects/dynamorio/build/build_debug-internal-32/suite/tests/bin/simple_app.exe"
[00:09:36] 213: Test timeout computed to be: 600
[00:09:36] 213: Hello, world!
[00:09:36] 213: Assertion failed: type_is_instr(cur_ref.instr.type), file ..\..\clients\drcachesim\reader\reader.cpp, line 132
```

derekbruening:
Happened again:
https://ci.appveyor.com/project/DynamoRIO/dynamorio/builds/22313260

```
00:06:34] test 213
[00:06:34]         Start 213: code_api|tool.reuse_distance
[00:06:34] 
[00:06:34] 213: Test command: C:\projects\dynamorio\build\build_debug-internal-32\bin32\drrun.exe "-s" "90" "-quiet" "-debug" "-use_dll" "C:/projects/dynamorio/build/build_debug-internal-32/lib32/debug/dynamorio.dll" "-exit0" "-stderr_mask" "0xC" "-msgbox_mask" "0" "-dumpcore_mask" "0x7d" "-staged" "-code_api" "-t" "drcachesim" "-ipc_name" "drtestpipe_reuse_distance" "-simulator_type" "reuse_distance" "-reuse_distance_threshold" "256" "--" "C:/projects/dynamorio/build/build_debug-internal-32/suite/tests/bin/simple_app.exe"
[...]

PR #4441, opened 2020-09-09
i#4424: Adjust guard pages and unit sizes for >4K pages

### LINKED ISSUE #4424: Tune default options, esp. -guard_pages, for large-page machines, esp. AArch64

derekbruening:
Adds a new option -per_thread_guard_pages controlling whether
per-thread allocations have guard pages.  This option is turned off by
default for >4K pages, removing guard pages from thread-private
initial cache units, all thread-private heap units, thread-private
signal-pending special heap units, stacks, TLS segments (DR and
privlib), and thread-private gencode.  This saves a lot of space on
many-threaded apps, while maintaining guards throughout the VMM from
shared units, which are used far more often than private these days.

Implements the option with a new VMM flag VMM_PER_THREAD, added to the
reservation calls for the above list of unit types.

Additionally, for >4K pages, tunes the default unit sizes to take into
account the guard changes.  Increases the shared unit sizes further
for a better ratio of content to guard pages.

Tested on a large app on a 64K-page machine which originally hit OOM
without "-no_guard_pages" before (along with other now-fixed bugs:
i#4335; i#4418).  Here are some stats for a smaller version of the app
that is easier to capture statistics for:

-no_guard_pages:
              Peak threads under DynamoRIO control :               152
                              Threads ever created :               701
              Peak vmm blocks for unreachable heap :               914
                         Peak vmm blocks for stack :               610
      Peak vmm blocks for unreachable special heap :               152
      Peak vmm blocks for unreachable special mmap :               304
                Peak vmm blocks for reachable heap :               161
                         Peak vmm blocks for cache :               368
        Peak vmm blocks for reachable special mmap :                 5
            Peak vmm virtual memory in use (bytes) :         164757504

Here's this PR with -per_thread_guard_pages (i.e., overriding the new
default: so matching the HEAD defaults):
              Peak threads under DynamoRIO control :               157
                              Threads ever created :               706
              Peak vmm blocks for unreachable heap :              1690
                         Peak vmm blocks for stack :               945
      Peak vmm blocks for unreachable special heap :               471
      Peak vmm blocks for unreachable special mmap :               942
                Peak vmm blocks for reachable heap :               482
                         Peak vmm blocks for cache :               741
        Peak vmm blocks for reachable special mmap :                 7
            Peak vmm virtual memory in use (bytes) :         345899008

Vs the new defaults in this PR:
              Peak threads under DynamoRIO control :               141
                              Threads ever created :               701
              Peak vmm blocks for unreachable heap :               884
                         Peak vmm blocks for stack :               566
      Peak vmm blocks for unreachable special heap :               141
      Peak vmm blocks for unreachable special mmap :               282
                Peak vmm blocks for reachable heap :               152
                         Peak vmm blocks for cache :               411
        Peak vmm blocks for reachable special mmap :                 7
            Peak vmm virtual memory in use (bytes) :         160104448

Fixes #4424

### DIFF
--- a/core/heap.c
+++ b/core/heap.c
@@ -1917,7 +1917,7 @@ vmh_exit(vm_heap_t *vmh, bool contains_stacks)
         uint perstack =
             (uint)(ALIGN_FORWARD_UINT(
                        DYNAMO_OPTION(stack_size) +
-                           (DYNAMO_OPTION(guard_pages)
+                           (DYNAMO_OPTION(per_thread_guard_pages)
                                 ? (2 * PAGE_SIZE)
                                 : (DYNAMO_OPTION(stack_guard_pages) ? PAGE_SIZE : 0)),
                        DYNAMO_OPTION(vmm_block_size)) /
@@ -2802,7 +2802,9 @@ get_guarded_real_memory(size_t reserve_size, size_t commit_size, uint prot, bool
     heap_error_code_t error_code;
     bool try_vmm = true;
     ASSERT(reserve_size >= commit_size);
-    if (!guarded || !dynamo_options.guard_pages) {
+    if (!guarded ||
+        (TEST(VMM_PER_THREAD, which) && !DYNAMO_OPTION(per_thread_guard_pages)) ||
+        !DYNAMO_OPTION(guard_pages)) {
         if (reserve_size == commit_size)
             return get_real_memory(reserve_size, prot, add_vm, which _IF_DEBUG(comment));
         guard_size = 0;
@@ -2898,7 +2900,9 @@ static void
 release_guarded_real_memory(vm_addr_t p, size_t size, bool remove_vm, bool guarded,
                             which_vmm_t which)
 {
-    if (!guarded || !dynamo_options.guard_pages) {
+    if (!guarded ||
+        (TEST(VMM_PER_THREAD, which) && !DYNAMO_OPTION(per_thread_guard_pages)) ||
+        !dynamo_options.guard_pages) {
         release_real_memory(p, size, remove_vm, which);
         return;
     }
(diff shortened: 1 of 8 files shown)


derekbruening:
Appveyor failure looks like a flake: drcachesim.coherence test output flakiness.  Will search for existing issue/file new one.

derekbruening:
The coherence failure is the type_is_instr assert: #3320

core/heap.c
```diff
@@ -3335,6 +3341,7 @@ heap_vmareas_synch_units()
 {
     heap_unit_t *u, *next;
     /* make sure to add guard page on each side, as well */
+    // NOCHECK
```
abhinav92003, line 3344: What does this NOCHECK comment signify?

abhinav92003 (approved):

core/heap.c
derekbruening: Ah, I missed that, thank you.  That function needs a logic update for the new guard rules.  Fixed and removed the marker comment.

abhinav92003 (approved):

Issue #1582, opened 2014-12-31
add -native_exec support for ARM


derekbruening:
Split from issue #1551 

We need to implement all the glue code, including back_from_native_retstubs, to support -native_exec on ARM.

derekbruening:
I put in some support for going native and not coming back.  One major pain point is the TLS swapping.  os_should_swap_state():
```
    /* FIXME i#1852: this should return true, but there is a lot of complexity
     * getting os_switch_seg_to_context() to do the right then when called
     * at main thread init, secondary thread init, early and late injection,
     * and thread exit, since it is fragile with its writes to app TLS.
     */
```

Issue #3543, opened 2019-04-15
Is privload_call_entry really necessary?


summershrimp:
In order to running dynamorio on android9-arm64, we change `privload_call_lib_func`
```patch
diff --git a/core/unix/loader.c b/core/unix/loader.c
index 89d5d31c..00c3519e 100644
--- a/core/unix/loader.c
+++ b/core/unix/loader.c
@@ -911,7 +911,7 @@ privload_call_lib_func(fp_t func)
      */
     dummy_argv[0] = dummy_str;
     dummy_argv[1] = NULL;
-    func(1, dummy_argv, our_environ);
+    //func(1, dummy_argv, our_environ);
 }

 bool
```
Firstly, we thought programs with `init` or `init_array` would behave abnormal, but a few testcases shows that It's working perfectly. Then we thought the module load callback would behave abnormal, but it also works. 

The question is, what's really happening when call init functions from dynamorio privload? What would failed if we don't call init functions from dynamorio privload?

PS. Testing with libinscount.so with patch below.
```patch
+++ b/api/samples/inscount.cpp
@@ -85,6 +85,14 @@ static dr_emit_flags_t
 event_app_instruction(void *drcontext, void *tag, instrlist_t *bb, instr_t *inst,
                       bool for_trace, bool translating, void *user_data);

+static void
+event_module_load(void *drcontext, const module_data_t *info, bool loaded)
+{
+    const char *module_name = dr_module_preferred_name(info);
+    dr_fprintf(STDOUT, "Module Name: %s\n", module_name);
+
+}
+
 DR_EXPORT void
 dr_client_main(client_id_t id, int argc, const char *argv[])
 {
@@ -108,6 +116,7 @@ dr_client_main(client_id_t id, int argc, const char *argv[])
     dr_register_exit_event(event_exit);
     drmgr_register_bb_instrumentation_event(event_bb_analysis, event_app_instruction,
                                             NULL);
+    drmgr_register_module_load_event(event_module_load);

     /* make it easy to tell, by looking at log file, which client executed */
     dr_log(NULL, DR_LOG_ALL, 1, "Client 'inscount' initializing\n");
```

Testing program:

bad_case.cpp:
```
// g++ bad_case.cpp -o bad_case.so -fPIC -shared 
#include <cassert>
#include <cstdio>
#include <sys/types.h>
#include <sys/stat.h>
#include <fcntl.h>
#include <unistd.h>

unsigned long global_magic;

[...]

PR #4443, opened 2020-09-10
i#1578 arm detach: Restore TLS register on detach

derekbruening:
Sets os_should_swap_state() to true.
Adds support for os_swap_context() twice in a row with the same state.
Removes os_swap_context_go_native().

The api.detach test now passes on AArch64.

Enables all of the api.* and api.static_* tests for AArchXX, except
api.startstop( crash) and api.static_prepop (hang) which will be fixed
separately.

Fixes a build error in the api.static_crash test: a missing header.

Issue: #1578, #1582


derekbruening:
Will update the description to include not enabling api.static_noclient and api.thread_churn.

derekbruening:
The A64 failure is odd: it lists `code_api|common.allasm_aarch64_isa` as a failure but it passed in the output above.

suite/tests/CMakeLists.txt
```diff
@@ -2776,40 +2776,41 @@ if (CLIENT_INTERFACE)
     endif ()
   endif (ARM)
 
-  if (NOT ARM AND NOT AARCH64) # FIXME i#1578: fix detach on ARM/AArch64
+  if (NOT AARCHXX) # TODO i#1578: Crashing on AArch64.
```
abhinav92003, line 2779: Is the crash mentioned here related to the `XXX i#1578` comment in `core/unix/os.c`?

abhinav92003 (approved):

suite/tests/CMakeLists.txt
abhinav92003: Oh that comment is just for pure-asm apps. So probably not.

derekbruening: The os.c comment about no-libc apps?  No, that was affecting the allasm tests which have no TLS of their own b/c they don't use any libraries.

derekbruening:
> The A64 failure is odd: it lists `code_api|common.allasm_aarch64_isa` as a failure but it passed in the output above.

I figured it out: the release build test passed, and the web interface truncated the start of the log.  On expanding it we see the failure in debug.  It's the same as what I fixed with the os.c no-libc check for having a TLS segment set up, but this test puts a garbage value in the thread register (just as part of testing the ISA) so we have to rule that out too.

core/unix/os.c
```diff
@@ -2698,12 +2705,7 @@ os_should_swap_state(void)
     return (INTERNAL_OPTION(mangle_app_seg) &&
             IF_CLIENT_INTERFACE_ELSE(INTERNAL_OPTION(private_loader), false));
 #elif defined(AARCHXX)
-    /* FIXME i#1582: this should return true, but there is a lot of complexity
-     * getting os_switch_seg_to_context() to do the right then when called
-     * at main thread init, secondary thread init, early and late injection,
-     * and thread exit, since it is fragile with its writes to app TLS.
-     */
-    return false;
+    return IF_CLIENT_INTERFACE_ELSE(INTERNAL_OPTION(private_loader), false);
```
petrochenkov, line 2708: @derekbruening 
I found this change to be the source of a regression on Android - an application run under DR (without any extra options or clients) is now failing with "Cannot correctly handle received signal 11 in thread X.".

`INTERNAL_OPTION(private_loader)` here reports `true` on Android, but private loader doesn't really work there (cc https://github.com/DynamoRIO/dynamorio/issues/3543), and in this particular case something breaks if it's considered working.

I guess, since things generally work on Android, considering `INTERNAL_OPTION(private_loader)` true on Android is not *too* harmful, but apparently there's an unknown set of cases where it is indeed harmful.
Not sure what is the best course of action here, besides actually implementing the private loader.

derekbruening: Hmm, the real problem is that there is no CI testing for Android: #4639.  The code reflects the state as of several years ago, when the private loader did work.  Like you said, the best solution would be to get it working with the latest Android, though if you have a quick fix that improves things in the short term please submit a PR.

Unfortunately while the handful of regular DR maintainers are happy to provide advice and consulting for Android we are not likely to be able to spend time on actual coding work targeting Android.  Several people have expressed interest in Android support; hopefully someone will contribute some time.

derekbruening: @petrochenkov if you could file a new issue on the regression; and ideally submit a PR with a quick fix if it's blocking things.

petrochenkov: The quick fix that seems to work in practice is to continue returning `false` in the `defined(ANDROID)` case, unfortunately I cannot submit it without an approval.

derekbruening: > The quick fix that seems to work in practice is to continue returning false in the defined(ANDROID) case, unfortunately I cannot submit it without an approval.

You mean some approval process from your employer I assume.

PR #4447, opened 2020-09-16
i#4318 xarch memtrace: Add README for testdata

derekbruening:
Adds a README about how to re-generate the testdata added in PR #4330
commit 58a911a1.

Issue: #4318

### DIFF
--- a/clients/drcachesim/tests/drmemtrace.threadsig.aarch64.raw/README
+++ b/clients/drcachesim/tests/drmemtrace.threadsig.aarch64.raw/README
@@ -1,0 +1,50 @@
+How to re-generate the drmemtrace.threadsig.aarch64.raw files
+
+The application run is "threadsig" from https://github.com/DynamoRIO/drmemtrace_samples/blob/master/threadsig.aarch64/threadsig.cp
+
+To save repository space, build it with -Os and -static on an AArch64
+machine or cross-compile it:
+
+  $ g++ -o threadsig -Os threadsig.cpp -lpthread -static
+  $ strip threadsig
+
+Now run it with 4 threads and 2000 iterations, and possibly
+-unsafe_build_ldstex depending on your architecture:
+
+  $ bin64/drrun -unsafe_build_ldstex -t drcachesim -offline -- ~/threadsig 4 2000
+
+Compress the raw files:
+
+  $ gzip `ls -1td drmemtrace.threadsig.*.dir|head -1`/raw/*.raw
+
+Now copy the new data files and the threadsig and libdynamorio.so.
+
+Strip libdynamorio.so, though this may not decrease it much:
+
+  $ strip lib64/release/libdynamorio.so
+
+There is no need to change the checked-in libdrmemtrace.so.  It is not
+executed.  I created it by just copying a pure-asm tiny ELF binary:
+e.g., suite/tests/bin/common.allasm_aarch64_isa.
+
+A smaller libdynamorio.so I created by compiling this file
+allasm_svc_a64.s targeting dynamorio_sigreturn (the only code executed
+in libdynamorio.so) to be at address 0x510890:
+
+  --------------------------------------------------
+  .text
+  .globl _start
+  .type _start, @function
+
+          .align   6
+  _start:
+          .skip 0x110790, 0
+          mov      w8, #0x8b
+          svc      #0
+  --------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


clients/drcachesim/tests/drmemtrace.threadsig.aarch64.raw/README
```diff
@@ -0,0 +1,50 @@
+How to re-generate the drmemtrace.threadsig.aarch64.raw files
+
+The application run is "threadsig" from https://github.com/DynamoRIO/drmemtrace_samples/blob/master/threadsig.aarch64/threadsig.cp
```
abhinav92003, line 3: threadsig.cpp (missed one letter at the end)

clients/drcachesim/tests/drmemtrace.threadsig.aarch64.raw/README
```diff
@@ -0,0 +1,50 @@
...
+
+To save repository space, build it with -Os and -static on an AArch64
```
abhinav92003, line 5: There is a binary present in the repo: https://github.com/DynamoRIO/drmemtrace_samples/blob/master/threadsig.aarch64/binaries/threadsig, which is same as the one checked in at dynamorio/clients/drcachesim/tests/drmemtrace.threadsig.aarch64.raw/threadsig. Is it ok to just use that one? Or should I recompile and regenerate the trace?

clients/drcachesim/tests/drmemtrace.threadsig.aarch64.raw/README
```diff
@@ -0,0 +1,50 @@
...
+
+  $ gzip `ls -1td drmemtrace.threadsig.*.dir|head -1`/raw/*.raw
+
+Now copy the new data files and the threadsig and libdynamorio.so.
+
+Strip libdynamorio.so, though this may not decrease it much:
+
+  $ strip lib64/release/libdynamorio.so
+
+There is no need to change the checked-in libdrmemtrace.so.  It is not
```
abhinav92003, line 26: Curious why dynamorio includes it in the list of modules if it is not executed?

clients/drcachesim/tests/drmemtrace.threadsig.aarch64.raw/README
```diff
@@ -0,0 +1,50 @@
...
+executed.  I created it by just copying a pure-asm tiny ELF binary:
+e.g., suite/tests/bin/common.allasm_aarch64_isa.
+
+A smaller libdynamorio.so I created by compiling this file
```
abhinav92003, line 30: The part below seems optional. Perhaps, mark it as such?

abhinav92003 (approved):

clients/drcachesim/tests/drmemtrace.threadsig.aarch64.raw/README
derekbruening: You mean, why is a client library in the list of app libraries?  And why just the top-level client and none of the extension libraries it uses?  Some client libraries do have the app execute some of their code (such as when using drwrap_replace).  The client lib is mapped early, and its deps loaded later, and the memory walk finds it, and we deemed it better to include it than omit it for the code reason.  There is an issue on it IIRC.

clients/drcachesim/tests/drmemtrace.threadsig.aarch64.raw/README
derekbruening: Hmm, yeah, just running that one seems fine.  I will update the README to give that option.

clients/drcachesim/tests/drmemtrace.threadsig.aarch64.raw/README
derekbruening: It does save 300K so I was waffling whether to push it.  WDYT?

clients/drcachesim/tests/drmemtrace.threadsig.aarch64.raw/README
abhinav92003: The output binary after following these steps is exactly the same.

clients/drcachesim/tests/drmemtrace.threadsig.aarch64.raw/README
derekbruening: Probably it's too fiddly and complex to synthesize (have to go find dynamorio_sigreturn address, etc.) or make a tool; will mark optional and we will live with the 300K.

Issue #4446, opened 2020-09-16
[AArch64] proc_get_cache_line_size returns incorrect size


abhinav92003:
On my AArch64 machine, `proc_get_cache_line_size` returns 32, whereas `cat /sys/devices/system/cpu/cpu0/cache/index0/coherency_line_size` shows 128. 

proc_get_cache_line_size needs to be ported to give correct cache line size for AArch64.

abhinav92003:
Unfortunately, `getconf LEVEL1_DCACHE_LINESIZE` doesn't give this information on AArch64.

We can probably read DminLine from the ctr_el0 system register to get this.
https://developer.arm.com/docs/ddi0595/h/aarch64-system-registers/ctr_el0

derekbruening:
@AssadHashmi is DminLine the best solution here?  DR can read that from userspace?

AssadHashmi:
> @AssadHashmi is DminLine the best solution here? DR can read that from userspace?

Yes, using the MRS instruction.

PR #4449, opened 2020-09-17
i#4446: Obtain AArch64 cache line size using ctr_el0.

abhinav92003:
Uses the CTR_EL0 system register to obtain cache line size in proc_get_cache_line_size on AArch64.

Fixes: #4446

### DIFF
--- a/core/arch/aarch64/proc.c
+++ b/core/arch/aarch64/proc.c
@@ -34,25 +34,9 @@
 #include "proc.h"
 #include "instr.h"
 
-// The whitepaper below documents AArch64 words being 32 bits wide.
-// https://developer.arm.com/-/media/Files/pdf/
-// graphics-and-multimedia/Porting%20to%20ARM%2064-bit.pdf
-#define AARCH64_WORD_SIZE_BYTES 4
-
 static int num_simd_saved;
 static int num_simd_registers;
 static int num_opmask_registers;
-
-void
-set_cache_line_size_using_ctr_el0()
-{
-    uint64 ctr_el0;
-    __asm__ volatile("MRS %0, CTR_EL0" : "=r"(ctr_el0));
-    // Bits [19:16] are Log2 of the number of words in the smallest cache line
-    // of all the data caches and unified caches.
-    // https://developer.arm.com/docs/ddi0595/h/aarch64-system-registers/ctr_el0
-    cache_line_size = (1 << ((ctr_el0 >> 16) & 0xf)) * AARCH64_WORD_SIZE_BYTES;
-}
 
 void
 proc_init_arch(void)
@@ -61,7 +45,8 @@ proc_init_arch(void)
     num_simd_registers = MCXT_NUM_SIMD_SLOTS;
     num_opmask_registers = MCXT_NUM_OPMASK_SLOTS;
 
-    set_cache_line_size_using_ctr_el0();
+    set_cache_line_size_using_ctr_el0(/* dcache_line_size= */ &cache_line_size,
+                                      /* icache_line_size= */ NULL);
 
     /* FIXME i#1569: NYI */
 }
--- a/core/arch/aarch64/proc.c
+++ b/core/arch/aarch64/proc.c
@@ -45,8 +45,14 @@ proc_init_arch(void)
     num_simd_registers = MCXT_NUM_SIMD_SLOTS;
     num_opmask_registers = MCXT_NUM_OPMASK_SLOTS;
 
-    set_cache_line_size_using_ctr_el0(/* dcache_line_size= */ &cache_line_size,
-                                      /* icache_line_size= */ NULL);
+    size_t dcache_line_size;
+    // When DR_HOST_NOT_TARGET, get_cache_line_size returns false and does
+    // not set any value in given args. In that case, we do not want to
+    // overwrite the default cache_line_size.
+    if (get_cache_line_size(&dcache_line_size,
+                            /* icache_line_size= */ NULL)) {
+        cache_line_size = dcache_line_size;
+    }
 
     /* FIXME i#1569: NYI */
 }
--- a/core/arch/aarch64/proc.c
+++ b/core/arch/aarch64/proc.c
@@ -45,10 +45,8 @@ proc_init_arch(void)
     num_simd_registers = MCXT_NUM_SIMD_SLOTS;
     num_opmask_registers = MCXT_NUM_OPMASK_SLOTS;
 
-#ifndef DR_HOST_NOT_TARGET
     set_cache_line_size_using_ctr_el0(/* dcache_line_size= */ &cache_line_size,
                                       /* icache_line_size= */ NULL);
-#endif
 
     /* FIXME i#1569: NYI */
 }
--- a/core/arch/aarch64/proc.c
+++ b/core/arch/aarch64/proc.c
@@ -34,9 +34,25 @@
 #include "proc.h"
 #include "instr.h"
 
+// The whitepaper below documents AArch64 words being 32 bits wide.
+// https://developer.arm.com/-/media/Files/pdf/
+// graphics-and-multimedia/Porting%20to%20ARM%2064-bit.pdf
+#define AARCH64_WORD_SIZE_BYTES 4
+
 static int num_simd_saved;
 static int num_simd_registers;
 static int num_opmask_registers;
+
+void
+set_cache_line_size_using_ctr_el0()
+{
+    uint64 ctr_el0;
+    __asm__ volatile("MRS %0, CTR_EL0" : "=r"(ctr_el0));
+    // Bits [19:16] are Log2 of the number of words in the smallest cache line
+    // of all the data caches and unified caches.
+    // https://developer.arm.com/docs/ddi0595/h/aarch64-system-registers/ctr_el0
+    cache_line_size = (1 << ((ctr_el0 >> 16) & 0xf)) * AARCH64_WORD_SIZE_BYTES;
+}
 
 void
 proc_init_arch(void)
@@ -44,6 +60,8 @@ proc_init_arch(void)
     num_simd_saved = MCXT_NUM_SIMD_SLOTS;
     num_simd_registers = MCXT_NUM_SIMD_SLOTS;
     num_opmask_registers = MCXT_NUM_OPMASK_SLOTS;
+
+    set_cache_line_size_using_ctr_el0();
 
     /* FIXME i#1569: NYI */
 }
--- a/core/arch/aarch64/proc.c
+++ b/core/arch/aarch64/proc.c
@@ -45,8 +45,10 @@ proc_init_arch(void)
     num_simd_registers = MCXT_NUM_SIMD_SLOTS;
     num_opmask_registers = MCXT_NUM_OPMASK_SLOTS;
 
+#ifndef DR_HOST_NOT_TARGET
     set_cache_line_size_using_ctr_el0(/* dcache_line_size= */ &cache_line_size,
                                       /* icache_line_size= */ NULL);
+#endif
 
     /* FIXME i#1569: NYI */
 }
--- a/core/arch/aarch64/proc.c
+++ b/core/arch/aarch64/proc.c
@@ -45,13 +45,12 @@ proc_init_arch(void)
     num_simd_registers = MCXT_NUM_SIMD_SLOTS;
     num_opmask_registers = MCXT_NUM_OPMASK_SLOTS;
 
-    size_t dcache_line_size;
-    // When DR_HOST_NOT_TARGET, get_cache_line_size returns false and does
-    // not set any value in given args. In that case, we do not want to
-    // overwrite the default cache_line_size.
-    if (get_cache_line_size(&dcache_line_size,
-                            /* icache_line_size= */ NULL)) {
-        cache_line_size = dcache_line_size;
+    /* When DR_HOST_NOT_TARGET, get_cache_line_size returns false and does
+     * not set any value in given args.
+     */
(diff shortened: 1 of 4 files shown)


abhinav92003:
I verified locally that `proc_get_cache_line_size` indeed returns 128 now on tx1. Also on tx1, we get:
```
abhinavas@tx1:~/dr/build/i4446-1$ cat /sys/devices/system/cpu/cpu0/cache/index0/coherency_line_size 
128
```

core/arch/aarch64/proc.c
```diff
@@ -34,17 +34,35 @@
...
+
 static int num_simd_saved;
 static int num_simd_registers;
 static int num_opmask_registers;
 
+void
+set_cache_line_size_using_ctr_el0()
+{
+    uint64 ctr_el0;
+    __asm__ volatile("MRS %0, CTR_EL0" : "=r"(ctr_el0));
```
derekbruening, line 50: style nit: our other inline asm all uses lower-case

core/arch/aarch64/proc.c
```diff
@@ -34,17 +34,35 @@
...
+    // Bits [19:16] are Log2 of the number of words in the smallest cache line
+    // of all the data caches and unified caches.
+    // https://developer.arm.com/docs/ddi0595/h/aarch64-system-registers/ctr_el0
+    cache_line_size = (1 << ((ctr_el0 >> 16) & 0xf)) * AARCH64_WORD_SIZE_BYTES;
```
derekbruening, line 54: Searching for "ctr_el0": looks like we already have code to compute the cache line size, in drlibc.c.  Maybe we should share that code instead of having it in two places.

AssadHashmi (approved):
The implementation looks fine, allowing for Derek's comments.
What is the caller/use-case for `set_cache_line_size_using_ctr_el0()`?
Will it always be an internal non-API function?

core/arch/aarch64/proc.c
```diff
@@ -34,17 +34,35 @@
...
+// The whitepaper below documents AArch64 words being 32 bits wide.
+// https://developer.arm.com/-/media/Files/pdf/
+// graphics-and-multimedia/Porting%20to%20ARM%2064-bit.pdf
+#define AARCH64_WORD_SIZE_BYTES 4
+
 static int num_simd_saved;
 static int num_simd_registers;
 static int num_opmask_registers;
 
+void
```
derekbruening, line 46: As Assad pointed out: this should be static, and should probably have a note that it will only work at init time (it will crash later trying to write to .data)

derekbruening: However, this will likely go away and be replaced with sthg like an assignment to the var inside the init function to a call to a drlibc function.

abhinav92003: Removed this code and using the drlibc function now.

core/arch/aarch64/proc.c
abhinav92003: Moved to drlibc

core/arch/aarch64/proc.c
abhinav92003: Done.

abhinav92003:
I'm seeing a cross-compile issue on the [DYNAMORIO_A64_ON_X86_ONLY](https://travis-ci.com/github/DynamoRIO/dynamorio/jobs/386369139) build

```
/home/travis/build/DynamoRIO/dynamorio/core/drlibc/drlibc.c: Assembler messages:
/home/travis/build/DynamoRIO/dynamorio/core/drlibc/drlibc.c:175: Error: no such instruction: `mrs %rax,ctr_el0'
```

but when I build using the following on an x86 machine, it works fine.
```
cmake  -DCMAKE_TOOLCHAIN_FILE=./../../src/i4446-1/make/toolchain-arm64.cmake ./../../src/i4446-1
make -j
```

It's weird also because drlib.c already had this code (I just moved it around) and with no build failure.

derekbruening:
> I'm seeing a cross-compile issue on the [DYNAMORIO_A64_ON_X86_ONLY](https://travis-ci.com/github/DynamoRIO/dynamorio/jobs/386369139) build
> 
> ```
> /home/travis/build/DynamoRIO/dynamorio/core/drlibc/drlibc.c: Assembler messages:
> /home/travis/build/DynamoRIO/dynamorio/core/drlibc/drlibc.c:175: Error: no such instruction: `mrs %rax,ctr_el0'
> ```
> 
> but when I build using the following on an x86 machine, it works fine.
> 
> ```
> cmake  -DCMAKE_TOOLCHAIN_FILE=./../../src/i4446-1/make/toolchain-arm64.cmake ./../../src/i4446-1
> make -j
> ```

Those are quite different builds: the first builds x86 binaries containing aarch64 decode/encode code, while the second builds "regular" aarch64 binaries with host==target==aarch64 decode/encode code.

> 
> It's weird also because drlib.c already had this code (I just moved it around) and with no build failure.

Are you sure you didn't accidentally change a DR_HOST_NOT_TARGET guard or something?  `mrs` is illegal asm for A64_ON_X86: you have to use x86 asm there.

abhinav92003:
> Those are quite different builds: the first builds x86 binaries containing aarch64 decode/encode code, while the second builds "regular" aarch64 binaries with host==target==aarch64 decode/encode code.

Ah right. I got the former mixed up with DYNAMORIO_CROSS_AARCHXX_LINUX_ONLY, which is the one that builds the "regular" aarch64 binaries.

> Are you sure you didn't accidentally change a DR_HOST_NOT_TARGET guard or something? `mrs` is illegal asm for A64_ON_X86: you have to use x86 asm there.

Didn't change any guard. 

I'm trying to verify if there's some compiler optimisation in play here; something that skips a method if it doesn't find a caller. I'm trying to put the call-site under a !DR_HOST_NOT_TARGET guard.

abhinav92003:
> > Are you sure you didn't accidentally change a DR_HOST_NOT_TARGET guard or something? `mrs` is illegal asm for A64_ON_X86: you have to use x86 asm there.
> 
> Didn't change any guard.

Ah right. The previous one was actually under a guard. Oops. I'll commit the fix.

abhinav92003:
In the aarch64-on-x86 build, the AArch64 `proc_init_arch` seems to be invoked, instead of the x86 one[1]. This means that `proc_get_cache_line_size` will return the default (32; set in core/arch/proc_shared.c), instead of the actual cache line size in the x86 machine used for post-processing. 

[1]: see test failure in 949c9c6

derekbruening:
> In the aarch64-on-x86 build, the AArch64 `proc_init_arch` seems to be invoked, instead of the x86 one[

Generally it's trying to make as much as possible be the target machine and not the host machine.  You can kind of see it both ways: for some instrumentation tools you would want the target hardware params, while for some execution optimizations you would want the host.  Of course DR is not a true cross-ISA binary translator, so these middle grounds are just gray areas with no correct answer at this point.

core/arch/aarch64/proc.c
```diff
@@ -45,6 +45,9 @@ proc_init_arch(void)
     num_simd_registers = MCXT_NUM_SIMD_SLOTS;
     num_opmask_registers = MCXT_NUM_OPMASK_SLOTS;
 
+    set_cache_line_size_using_ctr_el0(/* dcache_line_size= */ &cache_line_size,
```
derekbruening, line 48: I would expect this to be named *get* not *set*: it's a query; it's not setting some persistent state.

derekbruening, line 48: Also, why have "using_ctr_el0" in the name?  I don't think the caller should care about the implementation: that should be hidden behind an abstraction barrier.

core/drlibc/drlibc.c
```diff
@@ -164,6 +155,32 @@ clear_icache(void *beg, void *end)
...
+     * https://developer.arm.com/-/media/Files/pdf/
+     * graphics-and-multimedia/Porting%20to%20ARM%2064-bit.pdf
+     */
+    if (cache_info == 0)
+        __asm__ __volatile__("mrs %0, ctr_el0" : "=r"(cache_info));
+    if (dcache_line_size != NULL)
+        *dcache_line_size = 4 << (cache_info >> 16 & 0xf);
+    if (icache_line_size != NULL)
+        *icache_line_size = 4 << (cache_info & 0xf);
+#    endif
```
derekbruening, line 182: It just does nothing for host!=target?  No return value?  That is not what a caller is going to expect: seems like it will lead to uninitialized variables and subsequent hard to debug bugs.

core/drlibc/drlibc.c
```diff
@@ -164,6 +155,32 @@ clear_icache(void *beg, void *end)
...
+     * CTR_EL0 [19:16] : Log2 of number of 4-byte words in smallest dcache line
+     * CTR_EL0 [3:0]   : Log2 of number of 4-byte words in smallest icache line
+     * https://developer.arm.com/docs/ddi0595/h/aarch64-system-registers/ctr_el0
+     *
+     * Also, the whitepaper below documents AArch64 words being 32 bits wide.
+     * https://developer.arm.com/-/media/Files/pdf/
+     * graphics-and-multimedia/Porting%20to%20ARM%2064-bit.pdf
+     */
+    if (cache_info == 0)
+        __asm__ __volatile__("mrs %0, ctr_el0" : "=r"(cache_info));
```
derekbruening, line 177: Hmm, two issues with the static var if it's not called at init time: racy; will crash if .data is read-only.  Best to have a comment stating that this is required to be called at init time when linked into DR.

core/arch/aarch64/proc.c
abhinav92003: Done. I called it `set_` when it was writing the persistent `cache_line_size` var. But forgot to rename it later on.

abhinav92003: Modified the name.

core/drlibc/drlibc.c
abhinav92003: Added the comment.

core/drlibc/drlibc.c
abhinav92003: I modified it to return false when no value is set, and asserted the return value to be true in existing call at `clear_icache`.

core/drlibc/drlibc.c
abhinav92003: As the variable was marked static even before (when it was part of clear_icache), I assume this condition is already satisfied.

core/drlibc/drlibc.c
derekbruening: Stepping back, is having this leaf function fail the right solution?  For x86-on-a64 (not yet fully supported), won't we have a similar issue where proc_init runs `cpuid`?  Shouldn't the solution be at a higher level?

This goes back to the earlier discussion.  Maybe all our current uses case should use the host.   For raw2trace: asking for the cache line size is a fallback and is expected to be the host.  Unfortunately it's a complex topic: things like proc_num_simd_registers(), if you're trying to generate some cross-arch instrumentation, you'd want to be for the target.

derekbruening: Since this is a complex, unclear topic, for now let's go with your solution but with a XXX comment pointing at #1684.

core/arch/aarch64/proc.c
```diff
@@ -45,8 +45,14 @@ proc_init_arch(void)
     num_simd_registers = MCXT_NUM_SIMD_SLOTS;
     num_opmask_registers = MCXT_NUM_OPMASK_SLOTS;
 
-    set_cache_line_size_using_ctr_el0(/* dcache_line_size= */ &cache_line_size,
-                                      /* icache_line_size= */ NULL);
+    size_t dcache_line_size;
+    // When DR_HOST_NOT_TARGET, get_cache_line_size returns false and does
+    // not set any value in given args. In that case, we do not want to
```
derekbruening, line 50: But if it doesn't set the arg, we can just pass in cache_line_size.  No need for a separate local.

core/drlibc/drlibc.c
```diff
@@ -129,7 +129,11 @@ clear_icache(void *beg, void *end)
     if (beg_uint >= end_uint)
         return;
 
-    set_cache_line_size_using_ctr_el0(&dcache_line_size, &icache_line_size);
+    if (!get_cache_line_size(&dcache_line_size, &icache_line_size)) {
+        // We don't expect get_cache_line_size to return false, as this code is
+        // invoked only when (not DR_HOST_NOT_TARGET).
+        ASSERT(false);
```
derekbruening, line 135: ASSERT_NOT_REACHED

core/drlibc/drlibc.c
```diff
@@ -129,7 +129,11 @@ clear_icache(void *beg, void *end)
     if (beg_uint >= end_uint)
         return;
 
-    set_cache_line_size_using_ctr_el0(&dcache_line_size, &icache_line_size);
+    if (!get_cache_line_size(&dcache_line_size, &icache_line_size)) {
+        // We don't expect get_cache_line_size to return false, as this code is
```
derekbruening, line 133: We use C-style comments in .c files.

core/drlibc/drlibc.c
abhinav92003: Done.

core/drlibc/drlibc.c
abhinav92003: Done.

core/arch/aarch64/proc.c
abhinav92003: Done.

For some reason, I thought it safer to do it that way. But I guess it's okay.

core/drlibc/drlibc.c
abhinav92003: Done. Added the comment.

derekbruening (approved):

PR #4454, opened 2020-09-28
i#4083 func view: Allow for 7-digit thread id in func_view tests

derekbruening:
After an upgdate, my machine has 7-digit thread id's, causing the
func_view tests to fail.  Here I widen the printed field and update
the expected output for the stored trace.

Issue: #4083


abhinav92003 (approved):

derekbruening:
The Travis failure is on Mac and is rather puzzling.  I'm assuming it's unrelated to this PR and it's a Travis XCode upgrade or sthg?  But Mac built for PR #4453 just prior to this.
```
debug-internal-64: **** 4 build errors ****
	/.../dynamorio/build_debug-internal-64/suite/tests/decode.c_asm.asm:554: error: mismatch in operand sizes
```

derekbruening:
I re-ran just the Mac job and the error went away...?!?  Well unfortunately we do not have resources to figure out strange non-reproducible issues like that.  Moving on.

PR #4463, opened 2020-09-30
i#4459: Do not mangle app TLS ops for -no_private_loader

### LINKED ISSUE #4459: ARM: static DR should *not* emulate in mangle_reads_thread_register()
Statically-linked-into-the-app DR has the native app value in the thread register, so it should not mangle reads and writes of that register.  The code today does mangle it.

derekbruening:
Disables emulation of thread register reads or writes by the
application when -no_private_loader is in place, such as with
statically linked DR.  Such emulation is unnecessary as the
thread register holds native values in such cases.

Fixes #4459

### DIFF
--- a/core/arch/mangle_shared.c
+++ b/core/arch/mangle_shared.c
@@ -1603,14 +1603,16 @@ d_r_mangle(dcontext_t *dcontext, instrlist_t *ilist, uint *flags INOUT, bool man
 #endif /* X64 || ARM */
 
 #ifdef AARCHXX
-        if (!instr_is_meta(instr) && instr_reads_thread_register(instr)) {
+        if (!instr_is_meta(instr) && instr_reads_thread_register(instr) &&
+            IF_CLIENT_INTERFACE_ELSE(INTERNAL_OPTION(private_loader), false)) {
             next_instr = mangle_reads_thread_register(dcontext, ilist, instr, next_instr);
             continue;
         }
 #endif /* ARM || AARCH64 */
 
 #ifdef AARCH64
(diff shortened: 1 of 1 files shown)


derekbruening:
This is not a correctness issue, but a simplicity and efficiency issue, so there's no test to be added.

derekbruening:
Hmm the A64 failure is one of the tests just added yesterday which passed several times in a row then:
```
204: /var/lib/jenkins/.jenkins/workspace/DynamoRIO-AArch64-Precommit/build/build_debug-internal-64/suite/tests/bin/api.static_signal: relocation error: /var/lib/jenkins/.jenkins/workspace/DynamoRIO-AArch64-Precommit/build/build_debug-internal-64/suite/tests/bin/api.static_signal: symbol pthread_join, version GLIBC_2.17 not defined in file libpthread.so.0 with link time reference
```

That's weird: shouldn't building and running on the same machine never hit such a problem?  And why would it be non-deterministic?

derekbruening:
@AssadHashmi : any idea on this pthread issue?  Seems like a toolchain problem.  This is just a regular executable linking libpthread.

derekbruening:
run arm tests

derekbruening:
Hmm it went away on a re-run.  Well if we see it again we'll file an issue.

AssadHashmi:
> @AssadHashmi : any idea on this pthread issue? Seems like a toolchain problem. This is just a regular executable linking libpthread.

I've not seen this behaviour. Possibly a toolchain version issue?

PR #4466, opened 2020-09-30
i#4457: Fix clobbered register on detach

### LINKED ISSUE #4457: AArch64 CRASH post-detach: x0 clobbered, resulting in bad arg for stats in dr_app_stop_and_cleanup_with_stats
If we change the call to dr_app_stop_and_cleanup() to instead call dr_app_stop_and_cleanup_with_stats() in the api.detach test, we hit a crash in stats_get_snapshot().

Looks like x0 is clobbered on detach.

Hmm: ARM has emit_fcache_enter_gonative() but A64 does not!
So stolen reg value is not restored.
&stats is in x0 as 1st (only) arg to dr_app_stop_and_cleanup_with_stats().
Then fcache_enter uses x0 to store the target PC, clobbering it.

Looks like yet another case of the AArch64 port stopping early and not finishing known pieces, causing
problems later and wasted time diagnosing symptoms.

derekbruening:
Generates a special fcache_enter_gonative routine for AArch64 (now it
matches other architectures) to restore the stolen register and to
avoid clobbering x0 on detach.

The new generated code required adding negative displacement support
to XINST_CREATE_{load,store} and tweaking some other defines.

Since on Aarch64 we can't jump through memory or write to the PC, for
now we assume we're at an ABI boundary and we clobber a caller-saved
register.

Augments the api.detach test to use the _with_stats variant, which
reproduces the original crash without the fix.

Fixes #4457

### DIFF
--- a/core/arch/arch.c
+++ b/core/arch/arch.c
@@ -161,7 +161,7 @@ dump_emitted_routines(dcontext_t *dcontext, file_t file, const char *code_descri
                 print_file(file, "fcache_return:\n");
             else if (last_pc == code->do_syscall)
                 print_file(file, "do_syscall:\n");
-#    ifdef ARM
+#    ifdef AARCHXX
             else if (last_pc == code->fcache_enter_gonative)
                 print_file(file, "fcache_enter_gonative:\n");
 #    endif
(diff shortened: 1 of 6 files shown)


derekbruening:
Hmm static_signal printed "Got SIGSEGV' 2x in a row.  Let's see if it reproduces.

derekbruening:
run arm tests

derekbruening:
The static_signal issues are under https://github.com/DynamoRIO/dynamorio/issues/1578#issuecomment-701721626

abhinav92003 (approved):

Issue #4456, opened 2020-09-29
Travis Mac build failing non-deterministically on decode test


derekbruening:
The Travis Mac build suddenly failed in PR #4454.  That error went away on a re-run.  Then it happened again in PR #4455 :

https://travis-ci.com/github/DynamoRIO/dynamorio/jobs/392490440
```
====> FAILURE in debug-internal-64 <====
debug-internal-64: **** 4 build errors ****
	/.../dynamorio/build_debug-internal-64/suite/tests/decode.c_asm.asm:554: error: mismatch in operand sizes
	make[2]: *** [suite/tests/CMakeFiles/common.decode.dir/decode.c_asm.asm.o] Error 1
	make[1]: *** [suite/tests/CMakeFiles/common.decode.dir/all] Error 2
	make: *** [all] Error 2
```

The failure is unrelated to the PR's.  It is strange that such a failure wouldn't be repeatable.

Issue #4408, opened 2020-08-16
[Aarch64] float instruction `fmov` unrecognized, src and dst go wrong，float register value not saved


Eliana-Xie:
**Describe the bug**
[linux aarch64] cannot recognize `fmov` instruction，`instr_num_srcs` and `instr_num_dsts`  result error
 
**To Reproduce**
1. test.c
``` c
float a = 3.0;
printf("%f", a);
```
  40062c:	1e211000 	fmov	s0, #3.000000000000000000e+00
  400630:	bd001fa0 	str	s0, [x29, #28]
  400634:	bd401fa0 	ldr	s0, [x29, #28]
  400638:	1e22c000 	fcvt	d0, s0

2. ./../bin64/drrun -c bin/libinstrace_simple.so -- ./float_test

modify `instrument_instr` funtion

```
    num = instr_num_srcs(where);
    printf(" src: %d ", num);
    for (i = 0; i < num; ++i)
    {
        insert_save_src(drcontext, ilist, where, reg_ptr, reg_tmp, instr_get_src(where, i), i);
    }
    
    num = instr_num_dsts(where);
    printf(" dst: %d ", num);
    for (i = 0; i < num; ++i)
    {
        insert_save_dst(drcontext, ilist, where, reg_ptr, reg_tmp, instr_get_dst(where, i), i);
    }
```
add `insert_save_src` function
```
static void
insert_save_src(void *drcontext, instrlist_t *ilist, instr_t *where, reg_id_t base, reg_id_t scratch, opnd_t opnd, int i)
{
    if(opnd_is_reg(opnd)) {
        reg_id_t regid = opnd_get_reg(opnd);
        if(regid >= DR_REG_S0 && regid <= DR_REG_B31){
            printf("reg:%d\n", regid);
            opnd_t dst_addr = opnd_create_base_disp(base, DR_REG_NULL, 0, offsetof(ins_ref_t, src[i]), reg_get_size(regid));
            instr_t* instr = instr_create_1dst_1src(drcontext, OP_str, dst_addr, opnd);
            char disas_instr[196];
            instr_disassemble_to_buffer(drcontext, instr, disas_instr, 196);
            printf("instruction: %s\n", disas_instr);
            instrlist_meta_preinsert(ilist, where, instr);
        }
    }
}
```

3. Exact output or incorrect behavior.

pc: 40062c  src: 5  dst: 4 
pc: 400630  src: 1 reg:132
[...]

Issue #2629, opened 2017-09-08
Implement save_fpstate on AArch64


fhahn:
It is not implemented at the moment. We should also add a test that requires save_fpstate, because it looks like there is non for AArch64 at the moment.

xref #1569

egrimley:
But what should `save_fpstate` do on AArch64? The comment in AArch64's `proc_save_fpstate` says:
```
    /* All registers are saved by insert_push_all_registers so nothing extra
     * needs to be saved here.
     */
```
And there's the comment in proc.h:
```
/* On ARM/AArch64 proc_save_fpstate saves nothing, so use the smallest
 * legal size for an array.
 */
# define DR_FPSTATE_BUF_SIZE 1
```
Clearly `save_fpstate` is not the same thing as `proc_save_fpstate`, but perhaps the solution for `save_fpstate` is to replace `ASSERT_NOT_IMPLEMENTED` with `ASSERT_NOT_REACHED`: perhaps the function is required by an API but should not be called on ARM/AArch64?

egrimley:
On ARM and AArch64 the only use of `save_fpstate` is in `thread_set_self_context`, which (on any architecture) is only used by `check_wait_at_safe_spot`, sometimes via `thread_set_self_mcontext`.

abhinav92003:
> On ARM and AArch64 the only use of `save_fpstate` is in `thread_set_self_context`, which (on any architecture) is only used by `check_wait_at_safe_spot`, sometimes via `thread_set_self_mcontext`.

@egrimley Is this just an observation of the existing code's state, or are you pointing out that we do need the `save_fpstate` in `thread_set_self_context`.

For context: I'm working on PR #4507 and we're trying to figure out whether we need this call or not for AArch64 (see comment thread on PR).

derekbruening:
Both `save_fpstate` and `proc_save_fpstate` only exist because the x87 floating point registers and status words are not kept in the `dr_mcontext_t`.  `save_fpstate` is used to update the x87 state when a signal is delayed and the state stored by the kernel needs updating in case the app ran x87 instructions in the delay window; `proc_save_fpstate` is for when DR or a client uses x87 instructions and keeps that use from clobbering the native x87 state which is what the app uses.

But, on arm and aarch64, the entire register state is in `dr_mcontext_t`, including all the SIMD registers which are used for floating-point operations.   Right?  So I don't think we need either `save_fpstate` nor `proc_save_fpstate` on any aarchxx architecture.

[...]

PR #4467, opened 2020-10-01
i#4408 Add AArch64 immediate FMOV instructions

AssadHashmi:
Adds vector and scalar immediate FMOV instructions for AArch64.


AssadHashmi:
Hi @derekbruening, this patch fails the Travis GCC `DYNAMORIO_A64_ON_X86_ONLY=yes DEPLOY=no` tests with:
```
/home/travis/build/DynamoRIO/dynamorio/core/ir/aarch64/codec.c: In function decode_opnd_fpimm8:
/home/travis/build/DynamoRIO/dynamorio/core/ir/aarch64/codec.c:1597:9: error: unknown type name __fp16
         __fp16 f;
```
Is this a cross-compiler issue or something to do with building a standalone decoder supporting both x86 and Arm (https://github.com/DynamoRIO/dynamorio/issues/1684)?
I tried explictly defining the fp16 compiler flag (`-march=armv8-a+fp16`) as a check but that didn't work.

derekbruening:
> Hi @derekbruening, this patch fails the Travis GCC `DYNAMORIO_A64_ON_X86_ONLY=yes DEPLOY=no` tests with:
> 
> ```
> /home/travis/build/DynamoRIO/dynamorio/core/ir/aarch64/codec.c: In function decode_opnd_fpimm8:
> /home/travis/build/DynamoRIO/dynamorio/core/ir/aarch64/codec.c:1597:9: error: unknown type name __fp16
>          __fp16 f;
> ```
> 
> Is this a cross-compiler issue or something to do with building a standalone decoder supporting both x86 and Arm (#1684)?

#1684, but mostly #4318 which is the real use case: processing arm traces on x86 machines.

> I tried explictly defining the fp16 compiler flag (`-march=armv8-a+fp16`) as a check but that didn't work.

Follow https://github.com/DynamoRIO/dynamorio/wiki/How-To-Build#building-aarchxx-tooling-to-execute-on-x86 to build locally.  This builds x86 binaries, so you would not want -march to point at arm.  You would need a type known to the compiler for building on x86.

core/ir/aarch64/codec.c
```diff
@@ -1556,6 +1557,113 @@ encode_opnd_cond(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_out)
...
+     * imm64 = Replicate(imm32, 2);
+     *       = a:~b:bbbbb:cdefgh:0000000000000000000 a:~b:bbbbb:cdefgh:0000000000000000000
+     *
+     * datasize = if Q == '1' then 128 else 64;
+     * imm = Replicate(imm64, datasize DIV 64);
+     *     = imm64       (Q=0)
+     *     = imm64:imm64 (Q=1)
+     */
+    union {
+        __fp16 f;
```
derekbruening, line 1597: As discussed, this needs some way to construct a type known to the x86 compiler for a64-on-x86.  Not sure what the best solution is: build a 32-bit float from the pieces manually; find some half-float support in gcc and clang; sthg.

core/ir/aarch64/codec.c
```diff
@@ -1556,6 +1557,113 @@ encode_opnd_cond(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_out)
...
+     * 0x1    0    0    0     b
+     * 0x0    8    0    0     c
+     * 0x0    7    c    0     defgh
+     */
+    union {
+        __fp16 f;
+        ushort i;
+    } fpv;
+
+    if (opnd_is_immed_float(opnd)) {
```
derekbruening, line 1645: suggest: invert and have `if (!opnd_is_immed_float(opnd)) return false;` so the rest of the function is not indented and thus easier to read

core/ir/aarch64/instr_create.h
```diff
@@ -2544,5 +2544,24 @@ enum {
 #define INSTR_CREATE_umull2_vector(dc, Rd, Rm, Rn, width) \
     instr_create_1dst_3src(dc, OP_umull2, Rd, Rm, Rn, width)
 
+/**
+ * Creates a FMOV immediate to vector floating point move instruction.
```
derekbruening, line 2548: nit: s/a/an/ (unless you pronounce it "fuh-move"?  I guess I see it as "eff-move")

core/ir/aarch64/instr_create.h
```diff
@@ -2544,5 +2544,24 @@ enum {
...
+ * \param Rd      The output vector register.
+ * \param f       The source immediate floating point opnd.
+ * \param width   The output vector element width. Use either OPND_CREATE_HALF()
+ *                or OPND_CREATE_SINGLE().
+ */
+#define INSTR_CREATE_fmov_vector_imm(dc, Rd, f, width) \
+    instr_create_1dst_2src(dc, OP_fmov, Rd, f, width)
+
+/**
+ * Creates a FMOV immediate to scalar floating point move instruction.
```
derekbruening, line 2559: Ditto

core/ir/aarch64/codec.c
```diff
@@ -1744,6 +1852,143 @@ encode_opnd_z16(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_out)
...
+        uint64_t not_b = ((b == 0) ? 1 : 0);
+        uint64_t bbbbbbbb = ((b == 0) ? 0 : 0xff);
+        uint64_t c = imm & 0x20;
+        uint64_t defgh = imm & 0x1f;
+
+        uint64_t imm64 =
+            (a << 56) | (not_b << 62) | (bbbbbbbb << 55) | (c << 48) | (defgh << 48);
+
+        fpv.i = imm64;
+        *opnd = opnd_create_immed_float(fpv.f);
```
derekbruening, line 1919: Hmm, opnd_create_immed_float only takes a float, so this will lose precision from the double.  It seems we need to add a double value to opnd_t.  For 64-bit, replacing the existing float with double should result in no size change to opnd_t: for 32-bit I guess it will, but if no ISA has double immediates for 32-bit we could have an ifdef for X64?

suite/tests/api/ir_aarch64.c
```diff
@@ -458,6 +458,189 @@ test_fmov_general(void *dc)
...
+
+    /* FMOV <Vd>.<T>, #<imm> (v8.2 half-precision)
+     * 16 bit floating-point values encoded in instruction's 8 bit field, so
+     * there is a fixed, limited set of floating-point values which can be set,
+     * see 'Table C2-2 Floating-point constant values' in section 'Modified
+     * immediate constants in A64 floating-point instructions.' of the Arm
+     * Reference Manual.
+     */
+    instr = INSTR_CREATE_fmov_vector_imm(
+        dc, opnd_create_reg(DR_REG_Q1), opnd_create_immed_float(1.0), OPND_CREATE_HALF());
```
derekbruening, line 475: nit suggest: Use `1.0f` for a closer literal type to what is present, so the user doesn't expect double precision, and we don't have unnecessary casting?  Ditto below for the non-doubles (and see above for adding double support).

core/ir/aarch64/codec.c
AssadHashmi: After looking at half-precision support on GCC and Clang/LLVM for X86 and considering various options including intrinsics for conversion, I decided to build 32-bit float pieces manually. The variation in h/w support and relatively immature compiler support suggests that doing this manually for now is the best low-maintenance approach.

core/ir/aarch64/codec.c
AssadHashmi: Good point, done.

core/ir/aarch64/instr_create.h
AssadHashmi: Done.

core/ir/aarch64/instr_create.h
AssadHashmi: Done.

core/ir/aarch64/codec.c
AssadHashmi: Done.

suite/tests/api/ir_aarch64.c
AssadHashmi: Done.

AssadHashmi: run arm tests

AssadHashmi:
run arm tests

AssadHashmi:
Hi @derekbruening, the AArch64 precommit is consistently failing `code_api|client.drwrap-test-detach`.
It was passing before I merged up with the https://github.com/DynamoRIO/dynamorio/pull/4476 change, committed yesterday.

Also getting consistent appveyor regressions for `code_api|win32` and `code_api|security-win32` which weren't happening before. Not sure is this is a Windows test machine issue or related to https://github.com/DynamoRIO/dynamorio/pull/4476

derekbruening:
> Hi @derekbruening, the AArch64 precommit is consistently failing `code_api|client.drwrap-test-detach`.
> It was passing before I merged up with the #4476 change, committed yesterday.

That test was never run on a64 until PR #4470 enabled it Friday so we don't have much history for it.  #4476 seems unlikely to affect it -- it only changed a different set of tests.  Is it possible to ssh to the Jenkins machine and run client.drwrap-test-detach 100x to see the failure rate?

> Also getting consistent appveyor regressions for `code_api|win32` and `code_api|security-win32` which weren't happening before. Not sure is this is a Windows test machine issue or related to #4476

The appveyor log for this PR looks like everything is timing out and the run did not finish?  There's no summary and it seems to be mid-run with lots of 90s results: 90s is the drrun timeout.

derekbruening:
> > Also getting consistent appveyor regressions for `code_api|win32` and `code_api|security-win32` which weren't happening before. Not sure is this is a Windows test machine issue or related to #4476
> 
> The appveyor log for this PR looks like everything is timing out and the run did not finish? There's no summary and it seems to be mid-run with lots of 90s results: 90s is the drrun timeout.

Could you point to the appveyor page where you're seeing just those two (I think you're missing part of the name of the first one) tests failing?

What I see at https://ci.appveyor.com/project/DynamoRIO/dynamorio/history is that this PR was green for "Merge branch 'master' into i4408-fmov-imm-fix", but then for "i#4408 Add half-precision fixes for X86-64 and address review comments" it's red with a 1-hour duration which almost always means a timeout (that's the Appveyor absolute time limit) -- and the next 2 runs are also 1 hour, both "i#4408 Fix code layout/format non-conformances" and "Merge branch 'master' into i4408-fmov-imm-fix".  That seems to point at something in "i#4408 Add half-precision fixes for X86-64 and address review comments" causing hangs on Windows?

AssadHashmi:
> Could you point to the appveyor page where you're seeing just those two (I think you're missing part of the name of the first one) tests failing?

I was looking at https://ci.appveyor.com/project/DynamoRIO/dynamorio/builds/35600738
It's difficult to sift through it on a Web GUI so I downloaded the full log as text and searched for the failures locally.

> That seems to point at something in "i#4408 Add half-precision fixes for X86-64 and address review comments" causing hangs on Windows?

The only thing I can thing of in that commit which may cause hangs is that adding the `double` type to `opnd_t` changed its size causing a knock-on effect somewhere. Or perhaps adding a build for X86 host?

core/ir/aarch64/codec.c
```diff
@@ -1594,8 +1594,14 @@ decode_opnd_fpimm8(uint enc, int opcode, byte *pc, OUT opnd_t *opnd)
      *     = imm64:imm64 (Q=1)
      */
     union {
+#ifdef DR_HOST_AARCH64
         __fp16 f;
-        ushort i;
+        uint16_t i;
+#else
+        /* For platforms on which 16 bit (half-precision) FP not yet available */
```
derekbruening, line 1601: style nit: missing end punctuation `.` and missing complete sentence grammar: `is`

AssadHashmi:
> Or perhaps adding a build for X86 host?

I meant adding `#ifdef DR_HOST_AARCH64`

core/ir/aarch64/codec.c
```diff
@@ -1594,8 +1594,14 @@ decode_opnd_fpimm8(uint enc, int opcode, byte *pc, OUT opnd_t *opnd)
      *     = imm64:imm64 (Q=1)
      */
     union {
+#ifdef DR_HOST_AARCH64
```
derekbruening, line 1597: I think it would be nicer to have this be `#ifdef HAVE_HALF_FLOAT`, which matches other platform capabilities.  That define could be like these, set from cmake and configure.h:
```
make/configure.cmake.h:#cmakedefine HAVE_FVISIBILITY
make/configure.cmake.h:#cmakedefine HAVE_TYPELIMITS_CONTROL
make/configure.cmake.h:#cmakedefine HAVE_RSEQ
```

core/ir/aarch64/codec.c
```diff
@@ -1638,30 +1651,50 @@ encode_opnd_fpimm8(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_ou
      * 0x0    7    c    0     defgh
      */
     union {
+#ifdef DR_HOST_AARCH64
         __fp16 f;
-        ushort i;
+        uint16_t i;
+#else
+        /* For platforms on which 16 bit (half-precision) FP not yet available */
```
derekbruening, line 1658: Ditto

core/ir/opnd.h
```diff
@@ -1443,6 +1444,7 @@ enum {
     NULL_kind,
     IMMED_INTEGER_kind,
     IMMED_FLOAT_kind,
+    IMMED_DOUBLE_kind,
```
derekbruening, line 1447: Inserting in the middle is a binary compatibility break in this exposed enum.  This would have to be documented as such and the version numbers jumped including the oldest compatible.  Better to insert at the end.

core/ir/opnd.h
```diff
@@ -1535,6 +1537,15 @@ DR_API
...
 opnd_create_immed_float(float f);
 
+DR_API
+/**
+ * Returns an immediate double operand with value \p d.
+ * The caller's code should use proc_save_fpstate() or be inside a
+ * clean call that has requested to preserve the floating-point state.
+ */
+opnd_t
+opnd_create_immed_double(double d);
```
derekbruening, line 1547: Please add to the changelist in release.dox this new API function.

core/ir/opnd.h
```diff
@@ -1535,6 +1537,15 @@ DR_API
 opnd_t
 opnd_create_immed_float(float f);
 
+DR_API
+/**
+ * Returns an immediate double operand with value \p d.
+ * The caller's code should use proc_save_fpstate() or be inside a
+ * clean call that has requested to preserve the floating-point state.
```
derekbruening, line 1544: I know the float one says this, but it is a little stale these days: for x86 we always save the SIMD state and only don't auto-save the legacy x87 state.  For AArchXX aren't we also always saving any floating-point state?  So I would say tweak the docs to make that part conditional on legacy x87 floating point (on x86 modern compilers almost always used the SIMD floating point support and not the old x87).

suite/tests/api/ir_aarch64.c
```diff
@@ -536,55 +540,55 @@ test_fmov_scalar(void *dc)
...
     test_instr_encoding(dc, OP_fmov, instr);
 
     instr = INSTR_CREATE_fmov_scalar_imm(dc, opnd_create_reg(DR_REG_S0),
-                                         opnd_create_immed_float(0.2109375));
+                                         opnd_create_immed_float(0.2109375f));
     test_instr_encoding(dc, OP_fmov, instr);
 
     instr = INSTR_CREATE_fmov_scalar_imm(dc, opnd_create_reg(DR_REG_S0),
-                                         opnd_create_immed_float(31.0));
+                                         opnd_create_immed_float(31.0f));
```
derekbruening, line 591: None of these need the new opnd_create_immed_double?  Is there any code that calls that new routine here as at least a sanity check?

derekbruening:
> > Could you point to the appveyor page where you're seeing just those two (I think you're missing part of the name of the first one) tests failing?
> 
> I was looking at https://ci.appveyor.com/project/DynamoRIO/dynamorio/builds/35600738
> It's difficult to sift through it on a Web GUI so I downloaded the full log as text and searched for the failures locally.

Every single test there is failing except a handful of tests that don't run an app in the code cache (the unit tests).  The suite doesn't come close to finishing (only starts 74 tests out of hundreds in the first build, never gets to 2nd build) and is killed after 1 hour.  That's actually a good sign: better for everything to fail and then it should be an easy to figure out problem as opposed to just a few tests mysteriously failing.

> 
> > That seems to point at something in "i#4408 Add half-precision fixes for X86-64 and address review comments" causing hangs on Windows?
> 
> The only thing I can thing of in that commit which may cause hangs is that adding the `double` type to `opnd_t` changed its size causing a knock-on effect somewhere. Or perhaps adding a build for X86 host?

You can actually log in to an appveyor VM and interactively try to debug: http://www.appveyor.com/docs/how-to/rdp-to-build-worker

AssadHashmi:
> Is it possible to ssh to the Jenkins machine and run client.drwrap-test-detach 100x to see the failure rate?

It passes sometimes and fails others. Can't give a figure yet as it's still running due to the 90s timeout when it fails.
I built with with `cmake -DDEBUG=ON -DBUILD_TESTS=ON` and running with `ctest -V -I 196,196`.

derekbruening:
> > Is it possible to ssh to the Jenkins machine and run client.drwrap-test-detach 100x to see the failure rate?
> 
> It passes sometimes and fails others. Can't give a figure yet as it's still running due to the 90s timeout when it fails.
> I built with with `cmake -DDEBUG=ON -DBUILD_TESTS=ON` and running with `ctest -V -I 196,196`.

Given that several other detach tests still fail including some with a hang (api.static_prepop hangs; api.startstop crashes), despite several detach fixes I put in place, it's probably an underlying a64 detach bug unrelated to any PR since it was enabled.

AssadHashmi:
> Is it possible to ssh to the Jenkins machine and run client.drwrap-test-detach 100x to see the failure rate?

28 out of 100 runs failed.

derekbruening:
I tried to help by building this branch on my Win10 2004 machine but the tests are passing there: not reproducing the hang unfortunately.

AssadHashmi:
> I tried to help by building this branch on my Win10 2004 machine but the tests are passing there: not reproducing the hang unfortunately.

Thanks, I'll try a few more things tomorrow.

core/ir/aarch64/codec.c
AssadHashmi: Done.

core/ir/aarch64/codec.c
AssadHashmi: Done.

core/ir/opnd.h
AssadHashmi: Done.

core/ir/opnd.h
AssadHashmi: Done.

core/ir/opnd.h
AssadHashmi: > For AArchXX aren't we also always saving any floating-point state?

It's not clear from this issue whether that is the case: https://github.com/DynamoRIO/dynamorio/issues/2629

core/ir/opnd_shared.c
```diff
@@ -389,6 +389,22 @@ opnd_create_immed_float(float i)
...
 
+/* NOTE: requires caller to be under PRESERVE_FLOATING_POINT_STATE */
+opnd_t
+opnd_create_immed_double(double i)
+{
+    opnd_t opnd;
+    opnd.kind = IMMED_DOUBLE_kind;
+    /* note that manipulating floats and doubles is dangerous - see case 4360
+     * even this copy can end up using fp load/store instrs and could
+     * trigger a pending fp exception (i#386)
```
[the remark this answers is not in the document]
AssadHashmi: Done.

suite/tests/api/ir_aarch64.c
AssadHashmi: Good call. Added `double` tests which flushed out encoding and decoder errors.

AssadHashmi:
> > Is it possible to ssh to the Jenkins machine and run client.drwrap-test-detach 100x to see the failure rate?
> 
> 28 out of 100 runs failed.

The https://github.com/DynamoRIO/dynamorio/pull/4480 PR fixes the AArch64 precommit test regressions.

AssadHashmi:
Hi @derekbruening, regarding the Travis regressions https://api.travis-ci.com/v3/job/398223863/log.txt, is this expected?
```
fatal: unable to access 'https://github.com/Homebrew/homebrew-core/': transfer closed with outstanding read data remaining

Error: Fetching /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core failed!
```
Is this related to the addition of a `double` type to `opnd_t` and related changes? If so, odd that it should only happen on macOS. Is there an `OPSZ_` somewhere which needs setting for `double`s on macOS only?
```

debug-internal-64: **** 4 build errors ****

	/.../dynamorio/build_debug-internal-64/suite/tests/decode.c_asm.asm:554: error: mismatch in operand sizes

	make[2]: *** [suite/tests/CMakeFiles/common.decode.dir/decode.c_asm.asm.o] Error 1

	make[1]: *** [suite/tests/CMakeFiles/common.decode.dir/all] Error 2

	make: *** [all] Error 2
```

AssadHashmi:
> 	/.../dynamorio/build_debug-internal-64/suite/tests/decode.c_asm.asm:554: error: mismatch in operand sizes

This appears first in this commit https://github.com/DynamoRIO/dynamorio/pull/4467/commits/1c52b833109d24a4caf76dc7891ed00f9a85b2fd which also added more cases of `IMMED_DOUBLE_kind` and related `double` handling.

derekbruening:
> Hi @derekbruening, regarding the Travis regressions https://api.travis-ci.com/v3/job/398223863/log.txt, is this expected?
> 
> ```
> fatal: unable to access 'https://github.com/Homebrew/homebrew-core/': transfer closed with outstanding read data remaining
> 
> Error: Fetching /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core failed!
> ```

No, seems like a Travis issue with their network, or maybe some other transient infrastructure issue between Travis and Github or sthg.

> 
> Is this related to the addition of a `double` type to `opnd_t` and related changes? If so, odd that it should only happen on macOS. Is there an `OPSZ_` somewhere which needs setting for `double`s on macOS only?
> 
> ```
> 
> debug-internal-64: **** 4 build errors ****
> 
> 	/.../dynamorio/build_debug-internal-64/suite/tests/decode.c_asm.asm:554: error: mismatch in operand sizes
> 
> 	make[2]: *** [suite/tests/CMakeFiles/common.decode.dir/decode.c_asm.asm.o] Error 1
> 
> 	make[1]: *** [suite/tests/CMakeFiles/common.decode.dir/all] Error 2
> 
> 	make: *** [all] Error 2
> ```

Should show up in a tracker search: #4456

AssadHashmi:
@derekbruening wrote https://github.com/DynamoRIO/dynamorio/pull/4467#issuecomment-704998187:
> Every single test there is failing except a handful of tests that don't run an app in the code cache (the unit tests). The suite doesn't come close to finishing (only starts 74 tests out of hundreds in the first build, never gets to 2nd build) and is killed after 1 hour. That's actually a good sign: better for everything to fail and then it should be an easy to figure out problem as opposed to just a few tests mysteriously failing.

Right! Coupled with the fact that AArch64 precommit and AArch64 decode on X86 work, suggests a Windows specific issue.

Looking at the commit at which appveyor first started failing (2043bafc i#4408 Add half-precision fixes for X86-64 and address review comments) the only change which could affect Windows is the addition of a new `double` member in the `value` union in `_opnd_t`:
```
$ git diff 8bbb8419 2043bafc core/ir/opnd.h
diff --git a/core/ir/opnd.h b/core/ir/opnd.h
index 6d43bd7c..25ad7e1e 100644
--- a/core/ir/opnd.h
+++ b/core/ir/opnd.h
@@ -1360,7 +1360,8 @@ struct _opnd_t {
             int low;  /* IMMED_INTEGER_kind with DR_OPND_MULTI_PART */
             int high; /* IMMED_INTEGER_kind with DR_OPND_MULTI_PART */
         } immed_int_multi_part;
-        float immed_float; /* IMMED_FLOAT_kind */
+        float immed_float;   /* IMMED_FLOAT_kind */
+        double immed_double; /* IMMED_DOUBLE_kind */
```

This assert in `d_r_arch_init()` does not fire which implies the size of `_opnd_t` is not the cause:
```
    ASSERT(sizeof(opnd_t) == EXPECTED_SIZEOF_OPND);
```

However, there are two comments in the `_opnd_t` definition which describe previous workarounds for differences between the way `gcc` and `cl` compilers lay out, pad and align structs and unions in memory.

IMHO it is likely that the failure is caused by these differences. Can you shed any light on what could be the cause based on previous workarounds? 

> I tried to help by building this branch on my Win10 2004 machine but the tests are passing there: not reproducing the hang unfortunately.

Are there `cl` and/or library versions differences between your machine and appveyor's Windows VM which could account for the discrepancy?

derekbruening:
> However, there are two comments in the `_opnd_t` definition which describe previous workarounds for differences between the way `gcc` and `cl` compilers lay out, pad and align structs and unions in memory.
> 
> IMHO it is likely that the failure is caused by these differences. Can you shed any light on what could be the cause based on previous workarounds?

I see comments on the bitfields is all: but if those were not real bitfields I would think the sizeof assert would fire.

I would still suggest using the remote desktop access pointed at above to get a callstack of the hang.  It might point directly at the issue. 

> 
> Are there `cl` and/or library versions differences between your machine and appveyor's Windows VM which could account for the discrepancy?

The toolchain should be identical or very close.  The OS is different: Appveyor is way back on 1607 or sthg and mine is 2004. 
 For the toolchain, both are VS2017.  AFAIK there aren't major sub-versions of the compiler but there could be small differences from different updates since installation.  Looks like mine is `Microsoft (R) C/C++ Optimizing Compiler Version 19.12.25835 for x64` and Appveyor has...looking...`-- The CXX compiler identification is MSVC 19.16.27035.0`.  But I would find it surprising if such a difference changed how a struct is laid out.

AssadHashmi:
> I would still suggest using the remote desktop access pointed at above to get a callstack of the hang. It might point directly at the issue.

I'm having problems following the instructions at https://www.appveyor.com/docs/how-to/rdp-to-build-worker/

When I log in to AppVeyor, DynamoRIO appears as an account in my profile's **Security Settings** with **Role**->**Committer**.
dynamorio and drmemory appear in the **Projects** tab as well but the **Environments** tab is empty so I can't set any environment variable, like `APPVEYOR_RDP_PASSWORD`.
The **Global YAML configuration** is empty, unless all I need to insert  is `environment`, `init` or `on_finish` ?

I tried to add our standard `.appveyor.yml` into the **Global YAML configuration** section but a permissions error appears.

This is the first time I've tried to RDP into an AppVeyor VM.
Is there some initialisation I have to do in my AppVeyor config or the PR branch's `.appveyor.yml`?

derekbruening:
> > I would still suggest using the remote desktop access pointed at above to get a callstack of the hang. It might point directly at the issue.
> 
> I'm having problems following the instructions at https://www.appveyor.com/docs/how-to/rdp-to-build-worker/
> 
> When I log in to AppVeyor, DynamoRIO appears as an account in my profile's **Security Settings** with **Role**->**Committer**.
> dynamorio and drmemory appear in the **Projects** tab as well but the **Environments** tab is empty so I can't set any environment variable, like `APPVEYOR_RDP_PASSWORD`.
> The **Global YAML configuration** is empty, unless all I need to insert is `environment`, `init` or `on_finish` ?
> 
> I tried to add our standard `.appveyor.yml` into the **Global YAML configuration** section but a permissions error appears.
> 
> This is the first time I've tried to RDP into an AppVeyor VM.
> Is there some initialisation I have to do in my AppVeyor config or the PR branch's `.appveyor.yml`?

There is no need to log in to appveyor at all or make any changes through a web interface.
Simply put something like this in your branch's `.appveyor.yml`:

```
environment:
  APPVEYOR_RDP_PASSWORD: Diagnose-90s-Timeouts

init:
  - ps: iex ((new-object net.webclient).DownloadString('https://raw.githubusercontent.com/appveyor/ci/master/scripts/enable-rdp.ps1'))
```

Then look for the IP, port, and username in the build log and point an RDP client like `xfreerdp` at them.
IIRC it gives you one hour on the VM.  Use the build log to find the directory where all the binaries are.

AssadHashmi:
> There is no need to log in to appveyor at all or make any changes through a web interface.
> Simply put something like this in your branch's `.appveyor.yml`:

Thanks @derekbruening, that's nice and simple!

So, ALL failures trigger the `opnd_t` sizeof assert in `d_r_arch_init()`:
```
ASSERT(sizeof(opnd_t) == EXPECTED_SIZEOF_OPND);
```
e.g.
```
common.decode.exe.4300.00000000.ldmp:DynamoRIO debug check failure: ..\..\core\arch\arch.c:663 sizeof(opnd_t) == EXPECTED_SIZEOF_OPND
```
Based on the comment header for:
```
#define EXPECTED_SIZEOF_OPND (3 * sizeof(uint) IF_X64(+4 /*struct size padding*/))
```
I guess that `cl`'s padding and/or alignment dealing with the addition of a new `double` type to the union increases its size rather than fitting it in as `gcc` must have done.

derekbruening:
> > There is no need to log in to appveyor at all or make any changes through a web interface.
> > Simply put something like this in your branch's `.appveyor.yml`:
> 
> Thanks @derekbruening, that's nice and simple!

I'll look into adding it to the wiki pages on testing.

> So, ALL failures trigger the `opnd_t` sizeof assert in `d_r_arch_init()`:
> 
> ```
> ASSERT(sizeof(opnd_t) == EXPECTED_SIZEOF_OPND);
> ```

And the assert is very early and so some deadlock or early reporting problem hangs the process instead of printing it out visibly in the log?  I recall problems like that with very early asserts.

derekbruening:
> > > There is no need to log in to appveyor at all or make any changes through a web interface.
> > > Simply put something like this in your branch's `.appveyor.yml`:
> > 
> > 
> > Thanks @derekbruening, that's nice and simple!
> 
> I'll look into adding it to the wiki pages on testing.

https://github.com/DynamoRIO/dynamorio/wiki/Test-Suite#remote-login-to-appveyor

AssadHashmi:
Hi @derekbruening, I RDP'd into the AppVeyor VM to debug the `opnd_t` size problem on Windows with some standalone code which I built and tested on AArch64 Linux.

Using `PATH`s from passing AppVeyor log output, for `cl.exe`:
```
PS C:\Users\tmp> $env:Path += ";C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023\bin\Hostx86\x86"
```
and for `vcvarsall.bat`:
```
PS C:\Users\tmp> $env:Path += ";C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build"
PS C:\Users\tmp> vcvarsall.bat x86
**********************************************************************
** Visual Studio 2017 Developer Command Prompt v15.9.19
** Copyright (c) 2017 Microsoft Corporation
**********************************************************************
[vcvarsall.bat] Environment initialized for: 'x86'
```
Building the test case:
```
PS C:\Users\tmp> cl.exe opnd.c
Microsoft (R) C/C++ Optimizing Compiler Version 19.16.27035 for x86
Copyright (C) Microsoft Corporation.  All rights reserved.

opnd.c
opnd.c(1): fatal error C1034: stdio.h: no include path set
PS C:\Users\tmp>
```
Is there some other initialisation which needs to be done?
AIUI `vcvars` batch files are the standard way of setting up command-line use of `cl.exe`.

derekbruening:
> Hi @derekbruening, I RDP'd into the AppVeyor VM to debug the `opnd_t` size problem on Windows with some standalone code which I built and tested on AArch64 Linux.
> 
> Using `PATH`s from passing AppVeyor log output, for `cl.exe`:

You would want to have the vcvars file set the PATH since there are multiple dirs to include, not just the cl.exe one.

> PS C:\Users\tmp> vcvarsall.bat x86

I am not a powershell expert, but this may be launching a sub-process cmd.exe, and so the env var changes are local to that process and once you're back in the powershell parent process there are no changes?  (Also, I think this asks for 32-bit: you want "amd64" and not "x86" I think.)  Running the .bat file within cmd would keep the env var changes.

The simplest thing for one-time manual use is to use the start menu to open up a cmd shell from the Visual Studio 2017 menu items for a native 64-bit build environment.

derekbruening:
Is your plan to tweak the opnd_t padding for VS, or just to put the double field under an ifdef so Windows doesn't see it?  I'm assuming the former based on your posts.  I'm now wondering if the x86 code is wrong to always use single-precision float: I think the x87 floating-point instructions are always using the 80-bit format, so `double` would be better than `float`: so we do want `double` for Windows, so if it's a simple padding tweak to make it work let's go for that.  I filed #4488.

AssadHashmi:
> Is your plan to tweak the opnd_t padding for VS, or just to put the double field under an ifdef so Windows doesn't see it?

If it's a relatively simple low-risk tweak for VS, I'll do it, otherwise I'll just have to `#ifdef`.
(I've spent more time on this PR than scheduled.)

AssadHashmi:
> I am not a powershell expert, but this may be launching a sub-process cmd.exe

You are right @derekbruening , that seems to be the case. Using native `cmd` shell works.

> PS C:\Users\tmp> vcvarsall.bat x86
> you want "amd64" and not "x86"

Yes. I misread the AppVeyor log output.

Unfortunately I failed to find out in the time available and a toolchain I'm not familiar with what `cl.exe` is doing differently which causes the `sizeof` assert to fire. So, ommitting `immed_double` from `opnd_t` for Windows in this PR.

Ready for final review.

core/ir/opnd_shared.c
```diff
@@ -432,6 +461,19 @@ opnd_get_immed_float(opnd_t opnd)
     return opnd.value.immed_float;
 }
 
+#ifndef WINDOWS
+/* Type double currently not included for Windows because sizeof(opnd_t) does
+ * not equal EXPECTED_SIZEOF_OPND, triggering the ASSERT in d_r_arch_init().
```
derekbruening, line 466: Always point at an issue with a TODO or XXX comment for unfinished/non-permanent situations.  Could put under #4488.

core/ir/aarch64/codec.c
derekbruening: Re-iterating request to have a capability rather than a hardcoded arch.

core/ir/opnd_shared.c
derekbruening: I see you do have #4488 in one spot.  I think it's best if it's in each of these comments: there are a number of them.

core/ir/opnd_shared.c
```diff
@@ -1321,6 +1367,13 @@ opnd_same(opnd_t op1, opnd_t op2)
     case IMMED_FLOAT_kind:
         /* avoid any fp instrs (xref i#386) */
         return *(int *)(&op1.value.immed_float) == *(int *)(&op2.value.immed_float);
+#ifndef WINDOWS
+        /* Type double currently not included for Windows because sizeof(opnd_t) does
+         * not equal EXPECTED_SIZEOF_OPND, triggering the ASSERT in d_r_arch_init().
+         */
+    case IMMED_DOUBLE_kind:
+        return *(long *)(&op1.value.immed_double) == *(long *)(&op2.value.immed_double);
```
derekbruening, line 1375: `long` is going to have problems as it is 32-bit on Windows and could be elsewhere as well.  Sthg like `int64` would be better.

core/ir/opnd_shared.c
AssadHashmi: Done.

core/ir/opnd_shared.c
AssadHashmi: Done.

core/ir/aarch64/codec.c
AssadHashmi: Done.

derekbruening (approved):

derekbruening:
I think this forgot to put a ref to #4408 in the commit body: adding link here

PR #4484, opened 2020-10-10
GIVE ME BACK MY OPTIMIZATIONS!

jessielesbian:


derekbruening (commented):
If you'd like to experiment with these old dynamic optimizations, I think it would be best to do local experiments first.  You could make a project branch here if you like, or just do local changes until there's some evidence that these will provide some value.  The code is most likely bit-rotted and it will take more than this ifdef change to get it working, and then you'd do experimentation to see whether it has any effect on modern compiled code and modern applications, and decide whether to instead apply it to instrumentation.  Incremental commits are good, though, but in this case I think a few more steps to see what's ahead would be warranted before going back and committing small pieces; or as mentioned use a project branch.

 See also my reply here: https://groups.google.com/forum/#!topic/dynamorio-users/wF4rzW7mq88

Issue #2006, opened 2016-09-15
generalize drcachesim to support arbitrary trace analysis tools


derekbruening:
We plan to generalize drcachesim to support two types of trace analysis tools:
1.  Tools that are cache-agnostic, will not interact with the cache or TLB simulator, and will operate on the memory stream in the trace on their own.  These will simply use the tracer and the trace reader but not the simulator.
2.  Tools built on top of the cache or TLB simulator that will analyze additional aspects of the trace in relation to cache misses and other simulator features.  For these, the simulator will become a platform that may provide an event API.

derekbruening:
Xref #1703, #1729

derekbruening:
We've put a bunch of work into this:

+ b11c9068 i#2006 drcachesim: add class analyzer
+ fcd2c4a9 i#2006 generalize drcachesim: split reader from simulator
+ 2038b647 i#2006 generalize drcachesim: add analysis_tool_t
+ 141824d0 i#2006 generalize drcachesim: add the histogram tool
+ 19fe03b1 i#2006 generalize drcachesim: add drmodtrack_dump_buf
+ dc738447 i#2006 generalize drcachesim: add DRX_FILE_SKIP_OPEN for skip file open
+ e48dbd9b i#2006 generalize drcachesim: expand trace entry docs
+ 96a6614d i#2006 generalize drcachesim: refactor memref_t to use a union
+ d444c131 i#2006 generalize drcachesim: fix Windows build warning
+ 4fc4ac4d i#2006 generalize drcachesim: add branch type info
+ 8617acc9 i#2006 generalize drcachesim: use placement new
+ 018a5f73 i#2006 generalize drcachesim: add the reuse-distance tool
+ 4d7d5422 i#2006 generalize drcachesim: support customized file operation
+ 2f7bebcc i#2006 generalize drcachesim: set drmemtrace dr_client_main as a weak symbol
+ 12112de7 i#2006 generalize drcachesim: add extern "C" for drmemtrace_client_main on Linux
+ 65561fa3 i#2006 generalize drcachesim: omit log msg when file ops replaced (#2177)
+ 6c75e1fe i#2006 generalize drcachesim: add custom field support to drmodtrack (#2212)
+ 1e70a65e i#2006 generalize drcachesim: add custom module processing (#2214)
+ 370a9b20 i#2006 generalize drcachesim: improve custom field support (#2244)
[...]

PR #4490, opened 2020-10-16
i#2006 drcachesim: Support config files on other filesystems

derekbruening:
Changes the drcachesim configuration file interface to use a
std::istream rather than a file path string, to support non-standard
filesystems.

Tested with a proprietary filesystem library.

Issue: #2006

### DIFF
--- a/clients/drcachesim/simulator/cache_simulator.cpp
+++ b/clients/drcachesim/simulator/cache_simulator.cpp
@@ -60,7 +60,15 @@ cache_simulator_create(const cache_simulator_knobs_t &knobs)
 analysis_tool_t *
 cache_simulator_create(const std::string &config_file)
 {
-    return new cache_simulator_t(config_file);
+    std::ifstream fin;
+    fin.open(config_file);
+    if (!fin.is_open()) {
+        ERRMSG("Failed to open the config file '%s'\n", config_file.c_str());
+        return nullptr;
+    }
+    analysis_tool_t *sim = new cache_simulator_t(&fin);
+    fin.close();
+    return sim;
 }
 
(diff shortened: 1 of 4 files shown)


derekbruening:
The appveyor failure is a flaky test: it's client.winxfer, under #4058.  I already tried to relax the output constraints in PR #4411 but apparently there is more nondeterminism in there.  I put a comment in the issue for now.

clients/drcachesim/simulator/cache_simulator.cpp
```diff
@@ -60,7 +60,15 @@ cache_simulator_create(const cache_simulator_knobs_t &knobs)
...
 {
-    return new cache_simulator_t(config_file);
+    std::ifstream fin;
+    fin.open(config_file);
+    if (!fin.is_open()) {
+        ERRMSG("Failed to open the config file '%s'\n", config_file.c_str());
+        return nullptr;
+    }
+    analysis_tool_t *sim = new cache_simulator_t(&fin);
+    fin.close();
```
xyzsam, line 70: Would it be better to use a unique_ptr so the cache_simulator itself can own the std::ifstream and close it on its own? I was a bit surprised to see the ifstream closed so soon after the constructor call.

xyzsam (approved):

clients/drcachesim/simulator/cache_simulator.cpp
derekbruening: I like the idea but std::istream does not have a close() method, and std::ifstream is not simple to work with for our proprietary i/o library...per offline discussion going to live w/ the pointer.

Issue #1369, opened 2014-11-28
ASSERT (drmgr-test) meta-instr faulted?  must set translation field and handle fault!


derekbruening:
_From [bruen...@google.com](https://code.google.com/u/109494838902877177630/) on February 19, 2014 10:57:54_

This has happened at least twice in the last year or so when I've run the
test suite on win7  I can't reproduce it when run manually.

```
72/88 Testing: code_api|client.drmgr-test
72/88 Test: code_api|client.drmgr-test
Command: "D:/derek/dr/build_suite/build_debug-internal-64/bin64/drrun.exe" "-s" "60" "-quiet" "-exit0" "-ops" "-stderr_mask 0xC -msgbox_mask 0 -dumpcore_mask 0xfd -staged  -code_api -client_lib 'D:/derek/dr/build_suite/build_debug-internal-64/suite/tests/bin/client.drmgr-test.dll.dll;0;'  -code_api " "-use_dll" "D:/derek/dr/build_suite/build_debug-internal-64/lib64/debug/dynamorio.dll" "D:/derek/dr/build_suite/build_debug-internal-64/suite/tests/bin/client.drmgr-test.exe"
Directory: D:/derek/dr/build_suite/build_debug-internal-64/suite/tests
"code_api|client.drmgr-test" start time: Jul 11 15:26 Eastern Daylight Time
Output:
----------------------------------------------------------
About to create thread
in wnd_callback 0x0000000000000024 0
in wnd_callback 0x0000000000000081 0
in wnd_callback 0x0000000000000083 0
in wnd_callback 0x0000000000000001 0
in wnd_callback 0x0000000000008001 3 0
About to crash
<Application D:\derek\dr\build_suite\build_debug-internal-64\suite\tests\bin\client.drmgr-test.exe (8408) DynamoRIO usage error : meta-instr faulted?  must set translation field and handle fault!>
<Usage error: meta-instr faulted?  must set translation field and handle fault! (D:\derek\dr\git\src\core\x86\arch.c, line 2924)
version 3.2.1458, custom build
-client_lib 'D:/derek/dr/build_suite/build_debug-internal-64/suite/tests/bin/client.drmgr-test.dll.dll;0;' -code_api -probe_api -msgbox_mask 0 -dumpcore_mask 253 -stderr_mask 12 -stack_size >
<end of output>
Test time =   1.92 sec
----------------------------------------------------------
Test Fail Reason:
[...]

Issue #4460, opened 2020-09-29
AArch64 CRASH from some stolen register handling bug


derekbruening:
On a proprietary app we are seeing a crash where the app reads from the address 0x100.
The bug goes away with `-steal_reg 29` and seems to be some kind of stolen register mangling bug.

The problematic code involves reading from the thread reg into x28:
```
   0xaaaab55ec634 +116:    mrs     x28, tpidr_el0
   0xaaaab55ec638 +120:    mov     w8, #0x1                        // #1
   0xaaaab55ec63c +124:    ldr     w10, [x20]
   0xaaaab55ec640 +128:    cbz     w10, 0xaaaab55ec660 +160
   0xaaaab55ec644 +132:    mov     w9, w10
   0xaaaab55ec648 +136:    sub     w11, w10, #0x1
   0xaaaab55ec64c +140:    casa    w9, w11, [x20]
   0xaaaab55ec650 +144:    cmp     w9, w10
   0xaaaab55ec654 +148:    b.eq    0xaaaab55ec730 +368  // b.none
   0xaaaab55ec658 +152:    mov     w10, w9
   0xaaaab55ec65c +156:    cbnz    w9, 0xaaaab55ec648 +136
   0xaaaab55ec660 +160:    tbnz    w8, #0, 0xaaaab55ec68c +204
```

This is DR linked statically into the app, so the thread reg holds the app value.
It looks like DR mangles the thread reg when static, which it shouldn't (filed #4459) but fixing that doesn't eliminate the problem here.

I don't know where the mangling bug is: I'm using the workaround to work around it.

derekbruening:
I think the bug could be in DR mangling or drreg spill/restore/barrier or tracer instrumentation.

derekbruening:
Not making much progress:

I tried to make a standalone small test with blocks like that above but cannot reproduce a problem.

Spot-checks of DR mangling and drcachesim tracing code looks correct via visual inspection.

Trying to use `-reset_at_fragment_count NNN` (fixed for ... hmm not filed: part of PR #4491) plus `-steal_reg_at_reset N` (enabled in the code for A64) on the big app to zero in on the trigger: hitting what seem like problems in reset itself on A64.

derekbruening:
[...]

PR #4491, opened 2020-10-20
i#1369: Use synch flush callback to enable drcachesim tracing.

abhinav92003:
DR translates a fault in the code cache to a fault at the corresponding application address. This is done using ilist reconstruction for the fragment where the fault occurred.

But, this does not work as expected when the DR client changes instrumentation during execution; currently, drcachesim does this to enable tracing after -trace_after_instrs. The reconstructed basic block gets the new instrumentation whereas the one in code cache has the old one. This causes issues during fault handling.

In the current drcachesim case, it appears as though a meta-instr has faulted because the reconstructed ilist has a meta-instr at the code cache fault pc. This issue may manifest differently if the basic block with the new instrumentation is smaller than the old one (unlike the drcachesim 'meta-instr faulted' case) and the faulting address lies beyond the end of the new instrumented basic block. We may see an ASSERT_NOT_REACHED due to the ilist walk ending before the faulting code cache pc was found in the reconstructed ilist.

In the existing code, drcachesim attempts to avoid this by flushing old fragments using dr_unlink_flush_region after it switches to the tracing instrumentation. However, due to the flush being asynch, there's a race and the flush does not complete in time.

This PR adds support for a callback in the synchronous dr_flush_region API. The callback is executed after the flush but before the threads are resumed.

Using the dr_flush_region callback to change drcachesim instrumentation ensures that old instrumentation is not applied after the flush and the new one is not applied before.

Issue: #1369

### DIFF
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -173,6 +173,9 @@ Further non-compatibility-affecting changes include:
  - Added opnd_create_immed_double(), opnd_get_immed_double() and
    opnd_is_immed_double() to enable the creation and handling of double
    precision floating-point operands.
+ - Added dr_flush_region_ex API that accepts a callback to be executed after synch
+   flush but before the threads are resumed. The existing dr_flush_region API
+   is modified to invoke dr_flush_region_ex with a NULL callback.
 
 **************************************************
 <hr>
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -141,6 +141,10 @@ changes:
    member fields of classes are now following a consistent style with
    an underscore suffix.  References to renamed fields will need to be
    updated.
+ - Changed the signature of the dr_flush_region() instrument API, in order to
+   support a callback function which is executed after completion of flush but
+   before threads are resumed. Existing client source will need to be modified
+   to pass a NULL value for the flush_completion_callback arg.
 
 The changes between version \DR_VERSION and 8.0.0 include the following minor
 compatibility changes:
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -173,10 +173,6 @@ Further non-compatibility-affecting changes include:
  - Added opnd_create_immed_double(), opnd_get_immed_double() and
    opnd_is_immed_double() to enable the creation and handling of double
    precision floating-point operands.
- - Added -reset_at_created_thread_count option to allow triggering reset when
-   created thread count reaches the given value. This is unlike
-   reset_at_nth_thread which checks the existing thread count. This option is
-   currently used to make Linux reset tests more robust.
 
 **************************************************
 <hr>
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -141,10 +141,6 @@ changes:
    member fields of classes are now following a consistent style with
    an underscore suffix.  References to renamed fields will need to be
    updated.
- - Changed the signature of the dr_flush_region() instrument API, in order to
-   support a callback function which is executed after completion of flush but
-   before threads are resumed. Existing client source will need to be modified
-   to pass a NULL value for the flush_completion_callback arg.
 
 The changes between version \DR_VERSION and 8.0.0 include the following minor
 compatibility changes:
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -145,8 +145,6 @@ changes:
    support a callback function which is executed after completion of flush but
    before threads are resumed. Existing client source will need to be modified
    to pass a NULL value for the flush_completion_callback arg.
- - Replaces the -reset_at_fragment_count option with -reset_at_thread_count. This
-   was required to make the Linux reset tests more robust.
 
 The changes between version \DR_VERSION and 8.0.0 include the following minor
 compatibility changes:
@@ -179,6 +177,9 @@ Further non-compatibility-affecting changes include:
  - Added opnd_create_immed_double(), opnd_get_immed_double() and
    opnd_is_immed_double() to enable the creation and handling of double
    precision floating-point operands.
+ - Added -reset_at_thread_count option to allow triggering reset when created thread
+   count reaches the given value. This option is currently used to make Linux reset
+   tests more robust.
 
 **************************************************
 <hr>
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -174,8 +174,9 @@ Further non-compatibility-affecting changes include:
    opnd_is_immed_double() to enable the creation and handling of double
    precision floating-point operands.
  - Added -reset_at_created_thread_count option to allow triggering reset when
-   created thread count reaches the given value. This option is currently used to
-   make Linux reset tests more robust.
+   created thread count reaches the given value. This is unlike
+   reset_at_nth_thread which checks the existing thread count. This option is
+   currently used to make Linux reset tests more robust.
 
 **************************************************
 <hr>
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -145,6 +145,8 @@ changes:
    support a callback function which is executed after completion of flush but
    before threads are resumed. Existing client source will need to be modified
    to pass a NULL value for the flush_completion_callback arg.
+ - Replaces the -reset_at_fragment_count option with -reset_at_thread_count. This
+   was required to make the Linux reset tests more robust.
 
 The changes between version \DR_VERSION and 8.0.0 include the following minor
 compatibility changes:
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -173,9 +173,9 @@ Further non-compatibility-affecting changes include:
  - Added opnd_create_immed_double(), opnd_get_immed_double() and
    opnd_is_immed_double() to enable the creation and handling of double
    precision floating-point operands.
- - Added -reset_at_thread_count option to allow triggering reset when created thread
-   count reaches the given value. This option is currently used to make Linux reset
-   tests more robust.
+ - Added -reset_at_created_thread_count option to allow triggering reset when
+   created thread count reaches the given value. This option is currently used to
+   make Linux reset tests more robust.
 
 **************************************************
 <hr>
--- a/api/samples/cbr.c
+++ b/api/samples/cbr.c
@@ -251,7 +251,7 @@ at_taken(app_pc src, app_pc targ)
     /* Since the flush will remove the fragment we're already in,
      * redirect execution to the target address.
      */
-    dr_flush_region(src, 1);
+    dr_flush_region(src, 1, NULL);
     dr_get_mcontext(drcontext, &mcontext);
     mcontext.pc = targ;
     dr_redirect_execution(&mcontext);
@@ -280,7 +280,7 @@ at_not_taken(app_pc src, app_pc fall)
     /* Since the flush will remove the fragment we're already in,
      * redirect execution to the fallthrough address.
      */
-    dr_flush_region(src, 1);
+    dr_flush_region(src, 1, NULL);
     dr_get_mcontext(drcontext, &mcontext);
     mcontext.pc = fall;
     dr_redirect_execution(&mcontext);
--- a/api/samples/cbr.c
+++ b/api/samples/cbr.c
@@ -232,10 +232,9 @@ insert(hash_table_t table, app_pc addr, cbr_state_t state)
 static void
 at_taken(app_pc src, app_pc targ)
 {
-    dr_mcontext_t mcontext = {
-        sizeof(mcontext),
-        DR_MC_ALL,
-    };
+    dr_mcontext_t mcontext;
+    mcontext.size = sizeof(mcontext);
+    mcontext.flags = DR_MC_ALL;
     void *drcontext = dr_get_current_drcontext();
 
     /*
@@ -261,10 +260,9 @@ at_taken(app_pc src, app_pc targ)
 static void
 at_not_taken(app_pc src, app_pc fall)
 {
-    dr_mcontext_t mcontext = {
-        sizeof(mcontext),
-        DR_MC_ALL,
-    };
+    dr_mcontext_t mcontext;
+    mcontext.size = sizeof(mcontext);
+    mcontext.flags = DR_MC_ALL;
     void *drcontext = dr_get_current_drcontext();
 
     /*
--- a/api/samples/cbr.c
+++ b/api/samples/cbr.c
@@ -251,7 +251,7 @@ at_taken(app_pc src, app_pc targ)
     /* Since the flush will remove the fragment we're already in,
      * redirect execution to the target address.
      */
-    dr_flush_region(src, 1, NULL);
+    dr_flush_region(src, 1);
     dr_get_mcontext(drcontext, &mcontext);
     mcontext.pc = targ;
     dr_redirect_execution(&mcontext);
@@ -280,7 +280,7 @@ at_not_taken(app_pc src, app_pc fall)
     /* Since the flush will remove the fragment we're already in,
      * redirect execution to the fallthrough address.
      */
-    dr_flush_region(src, 1, NULL);
+    dr_flush_region(src, 1);
     dr_get_mcontext(drcontext, &mcontext);
     mcontext.pc = fall;
     dr_redirect_execution(&mcontext);
--- a/api/samples/cbr.c
+++ b/api/samples/cbr.c
@@ -232,6 +232,11 @@ insert(hash_table_t table, app_pc addr, cbr_state_t state)
 static void
 at_taken(app_pc src, app_pc targ)
 {
+    /*
+     * We've found that the time taken to zero a large struct shows up as
+     * noticeable overhead for optimized clients. So, instead of using partial
+     * struct initialization, we set the fields required individually.
+     */
     dr_mcontext_t mcontext;
     mcontext.size = sizeof(mcontext);
     mcontext.flags = DR_MC_ALL;
@@ -260,6 +265,11 @@ at_taken(app_pc src, app_pc targ)
 static void
 at_not_taken(app_pc src, app_pc fall)
 {
+    /*
+     * We've found that the time taken to zero a large struct shows up as
+     * noticeable overhead for optimized clients. So, instead of using partial
+     * struct initialization, we set the fields required individually.
+     */
     dr_mcontext_t mcontext;
     mcontext.size = sizeof(mcontext);
     mcontext.flags = DR_MC_ALL;
--- a/clients/drcachesim/tracer/tracer.cpp
+++ b/clients/drcachesim/tracer/tracer.cpp
@@ -1323,7 +1323,7 @@ static void *schedule_tracing_lock;
 #ifdef X86_64
 #    define DELAYED_CHECK_INLINED 1
 #else
-// XXX: we don't have the inlining implemented yet.
+// XXX, i#4487: we don't have the inlining implemented yet.
 #endif
 
 static dr_emit_flags_t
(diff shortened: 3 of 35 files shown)


abhinav92003:
PR in progress...

Local run of a proprietary app shows that some threads seg fault after the flush completes.

abhinav92003:
Ah, so the segfaulting threads were redirected to the reset exit stub by DR at [1]. The `SIGSEGV` occurs in the exit stub shown below (see [2] for DR code that emits this).

```
Thread xxx received signal SIGSEGV, Segmentation fault.
0x0000000046fc2458 in ?? ()
(gdb) x/50i 0x46fc2458
=> 0x46fc2458:	stp	x0, x1, [x28]
   0x46fc245c:	mov	x0, #0x4c20                	// #19488
   0x46fc2460:	movk	x0, #0xf7f0, lsl #16
   0x46fc2464:	movk	x0, #0xffff, lsl #32
   0x46fc2468:	ldr	x1, [x28, #64]
   0x46fc246c:	br	x1
   ...
```
The exit stub expects the stolen reg to be set up already, but it doesn't seem to be -- see `x28` below.
```
(gdb) i r 
x0             0xfffffffffffffffc  -4
x1             0xfffff45bcc80      281474781400192
...
x28            0x30000             196608
...
pc             0x46fc2458          0x46fc2458
```

After adding `arch_mcontext_reset_stolen_reg` the `SIGSEGV` goes away. But the current implementation does it only for `INTERNAL_OPTION(steal_reg_at_reset) != 0` (zero is the default).
https://github.com/DynamoRIO/dynamorio/blob/9bedb9c72d5808785e28d20df3ed1c932b555ac9/core/synch.c#L1749

Trying to check whether the stolen reg should have been set up already somewhere else, for the `steal_reg_at_reset==0` (default) case.

[1]: `translate_from_synchall_to_dispatch` https://github.com/DynamoRIO/dynamorio/blob/9bedb9c72d5808785e28d20df3ed1c932b555ac9/core/synch.c#L1816

[2]: exit stub emitting code:  https://github.com/DynamoRIO/dynamorio/blob/9bedb9c72d5808785e28d20df3ed1c932b555ac9/core/arch/aarch64/emit_utils.c#L121

derekbruening:
`steal_reg_at_reset` is a diagnostic option we added in the past for identifying and debugging stolen reg mangling bugs on 32-bit arm.

abhinav92003:
This PR seems to resolve the issue. I'll mark the issue fixed in the final commit message.

core/synch.c
```diff
@@ -1749,6 +1749,10 @@ translate_from_synchall_to_dispatch(thread_record_t *tr, thread_synch_state_t sy
                 arch_mcontext_reset_stolen_reg(dcontext, mc);
             }
         });
+        IF_AARCHXX({
+            set_stolen_reg_val(mc, (reg_t)os_get_dr_tls_base(dcontext));
```
abhinav92003, line 1753: Do we  need to save the existing value of the stolen reg somehow?

derekbruening: A client xl8 could have restored a value there so it seems yes.  (I wonder if the signal redirect does that properly.)

abhinav92003:
> The exit stub expects the stolen reg to be set up already, but it doesn't seem to be 

`linux.thread-reset` is capable of reproducing the above segfault-after-reset. BUT the current `reset_at_fragment_count` arg used in the test suite doesn't allow it to. As it is, the reset happens much before the child thread (the test has just 2 threads) starts. But the test did fail when I manually adjusted `reset_at_fragment_count` to make the reset happen when both threads are active.

Manually adjusting `reset_at_fragment_count` in the test suite doesn't seem like a good idea though, as the ideal value may change in future without us noticing, and also it seems to be different on A64 and x86. I tried change the option to `reset_at_every_nth_fragment_count` and tried with a low value like 10, but that leads to another debug assert failure:
```
<Application dynamorio/suite/tests/bin/linux.thread (38747).  Internal Error: DynamoRIO debug check failure: Not implemented @dynamorio/core/unix/signal_linux_aarch64.c:53 (0)
```

derekbruening:
> Manually adjusting `reset_at_fragment_count` in the test suite doesn't seem like a good idea though, as the ideal value may change in future without us noticing, and also it seems to be different on A64 and x86.

Other options:
- `-reset_at_nth_thread`.
- We could add an annotation to trigger a reset programmatically from the app.  However, annotations aren't implemented for A64 (another missing feature...)
- We could use a nudge of type reset, requested from a client or possibly forked helper app.

core/synch.c
abhinav92003: `transfer_from_sig_handler_to_fcache_return` doesn't seem to save the existing stolen reg value before writing it.
https://github.com/DynamoRIO/dynamorio/blob/6dea75659134018d0796e7c923b62db03fcc6cf2/core/unix/signal.c#L3426

Is it because the spill slot will continue to have that value, and can be restored later?
https://github.com/DynamoRIO/dynamorio/blob/6dea75659134018d0796e7c923b62db03fcc6cf2/core/translate.c#L1062

abhinav92003:
> Other options:
> 
> * `-reset_at_nth_thread`.

Thanks, I used this suggestion.

I also removed the existing `reset_at_fragment_count` flag as it seemed to me that it is useful only for the reset tests. As it was added before the initial open source import of DR, I cannot verify the context in which it was added. I added a release note for this. 

Let me know if you think we should retain the flag.

derekbruening:
> I also removed the existing `reset_at_fragment_count` flag as it seemed to me that it is useful only for the reset tests. As it was added before the initial open source import of DR, I cannot verify the context in which it was added. I added a release note for this.

I remember using it to binary-search which blocks trigger a bug: esp when combined with -steal_reg_at_reset.

abhinav92003:
> I remember using it to binary-search which blocks trigger a bug: esp when combined with -steal_reg_at_reset.

Oh okay, better to keep it then. I added it back now.

derekbruening:
> > I remember using it to binary-search which blocks trigger a bug: esp when combined with -steal_reg_at_reset.
> 
> Oh okay, better to keep it then. I added it back now.

While having tons of options does make it harder to test and validate all the code paths and I agree that removing unused ones is a good cleanup, I think this option will be useful: in fact, since I'm having such a hard time reproducing #4460 in an isolated test, the mentioned binary-search approach on the bigger app may be my next step (I'll s/ARM/AARCHXX/ for -steal_reg_at_reset), since it went away stealing x28 but was there for x29.

api/docs/release.dox
```diff
@@ -141,6 +141,10 @@ changes:
    member fields of classes are now following a consistent style with
    an underscore suffix.  References to renamed fields will need to be
    updated.
+ - Changed the signature of the dr_flush_region() instrument API, in order to
```
derekbruening, line 144: Generally we have avoided such breaks, esp for binary compatibility breaks like this (new DR with old client will use garbage callback value and probably crash).  We instead add an _ex version, so `dr_flush_region_ex()` (or we use a struct with a size for graceful field appending).

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1316,13 +1316,14 @@ event_kernel_xfer(void *drcontext, const dr_kernel_xfer_info_t *info)
...
-static void *enable_tracing_lock;
+static bool tracing_enabled;
+static volatile bool tracing_scheduled;
+static void *schedule_tracing_lock;
 
 #ifdef X86_64
 #    define DELAYED_CHECK_INLINED 1
 #else
-// XXX: we don't have the inlining implemented yet.
+// XXX, i#4487: we don't have the inlining implemented yet.
```
derekbruening, line 1326: nit: style convention has no comma there

core/lib/instrument_api.h
```diff
@@ -6228,11 +6228,14 @@ DR_API
...
  *
+ * \note Use flush_completion_callback to specify logic to be executed after the flush
+ * and before the threads are resumed.
+ *
  * \note As currently implemented, dr_delay_flush_region() with no completion callback
  * routine specified can be substantially more performant.
  */
 bool
-dr_flush_region(app_pc start, size_t size);
+dr_flush_region(app_pc start, size_t size, void (*flush_completion_callback)());
```
derekbruening, line 6238: suggest: Consider having a user_data param to the callback.  We've regretted not having them for all the DR event callbacks and in other places (xref ... some issue, will find it, plus internal issue ...).  It makes it nicer for the user to avoid global/static data.

core/synch.c
derekbruening: No, the client xl8 cb is called during xl8 and its result is not kept anywhere else.  It is likely rare for a client to mess w/ the app's stolen reg (most tools are observational only; drreg avoids using the stolen reg; etc.) but it seems like a real theoretical issue to me.  How about we ignore it here and match signal.c and I file an issue on it and we fix it in both places later if we all agree about it.

I filed #4495.

core/synch.c
```diff
@@ -1749,6 +1749,12 @@ translate_from_synchall_to_dispatch(thread_record_t *tr, thread_synch_state_t sy
...
```
derekbruening, line 1753: I wonder if this should be fixed in a separate PR since it is a separate bug.

suite/tests/linux/thread.c
```diff
@@ -83,22 +83,28 @@ static struct timespec sleeptime;
...
-    assert(child > -1);
-
-    /* waste some time */
-    nanosleep(&sleeptime, NULL);
-
-    child_exit = true;
-    /* we want deterministic printf ordering */
-    while (!child_done)
+    // We need to create multiple threads to verify correctness of reset,
+    // which requires more than one threads.
```
derekbruening, line 87: grammar: s/threads/thread/

suite/tests/linux/thread.c
```diff
@@ -83,22 +83,28 @@ static struct timespec sleeptime;
...
+    // XXX: Using -reset_at_thread_count 2 causes an ASSERT failure, so we
```
derekbruening, line 88: What assert is this?  We should file an issue if it's a pre-existing thing.  Does this happen with -reset_at_nth_thread?

derekbruening (changes requested):
Sorry for the delay -- somehow I thought a review had not yet been requested and it was still in an exploratory phase.

core/synch.c
derekbruening: If you could add a XXX or TODO comment about #4495.

suite/tests/linux/thread.c
abhinav92003: Yes, it occurs with `-reset_at_nth_thread 2` too.

```
Internal Error: DynamoRIO debug check failure: Not implemented @dynamorio/core/unix/signal_linux_aarch64.c:53 (0)
```

I'm having a hard time getting a stack trace  :/. For some reason, gdb doesn't break at `d_r_internal_error` or `save_fpstate`, even though it recognises them when setting the breakpoints.

Anyway, here's a stack trace I constructed manually:

```
save_fpstate at core/unix/signal_linux_aarch64.c:53
thread_set_self_context at core/unix/signal.c:2684
thread_set_self_mcontext at core/unix/signal.c:2771
check_wait_at_safe_spot at core/synch.c:770
dispatch_enter_fcache at core/dispatch.c:436
```
Let me try getting a proper stack trace and I'll file an issue for this.

api/docs/release.dox
```diff
@@ -173,6 +177,9 @@ Further non-compatibility-affecting changes include:
  - Added opnd_create_immed_double(), opnd_get_immed_double() and
    opnd_is_immed_double() to enable the creation and handling of double
    precision floating-point operands.
+ - Added -reset_at_thread_count option to allow triggering reset when created thread
+   count reaches the given value. This option is currently used to make Linux reset
+   tests more robust.
```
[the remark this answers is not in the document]
abhinav92003: Oh, I misunderstood your previous comment to suggest that we can add an option like that. I should've tried searching for `reset_at_nth_thread` first.

I tried using `reset_at_nth_thread` now. With a value of 2, it gets the assert failure (discussion in other comment thread). With a value higher than that, it doesn't ever reset, because the test has atmost 2 active threads at any point (we do not want the test to have more active threads than that, to keep the output ordering deterministic).

core/synch.c
abhinav92003: Without this part, this PR would be incomplete as the large app crashes without the properly set up stolen reg. 

If this part is a step in the right direction and you feel there's more work to be done on stolen reg handling here, I can add the TODO with #4495 and #4460 both.

derekbruening: The separate PR would go in before this one of course.  But this is a direct bug fix of arm synchall, which is a separate thing from the new callback.  IHMO the arm synchall bug should have its own issue filed, and its own separate fix.

suite/tests/linux/thread.c
derekbruening: > For some reason, gdb doesn't break at d_r_internal_error or save_fpstate, even though it recognises them when setting the breakpoints.

Maybe you're setting them in drrun rather than in the application?

core/synch.c
derekbruening: I don't think this is related to #4460: there is no evidence that it is, right?  There's no synchall in the #4460 scenario.  I've also put this fix in a local tree and #4460 still reproduces.

Maybe this synchall stolen reg restore could be bundled into #4495 and we leave it open for the further fix for synchall and for signals, rather than a separate issue.

core/synch.c
abhinav92003: Done.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1379,28 +1380,46 @@ enable_tracing_instrumentation()
...
-    if (do_flush && !dr_unlink_flush_region(NULL, ~0UL))
+    dr_mutex_unlock(schedule_tracing_lock);
+
+    if (do_flush) {
+        if (!dr_flush_region(NULL, ~0UL, change_instrumentation_callback))
+            DR_ASSERT(false);
+
+        dr_mcontext_t mcontext = {
+            sizeof(mcontext),
+            DR_MC_ALL,
```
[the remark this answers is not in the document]
abhinav92003: Done. Some other code also used the partial struct initialisation; changed those too.

api/docs/release.dox
abhinav92003: I added `dr_flush_region_ex` with the additional args, but didn't use a struct as I don't think new args would be added frequently.

clients/drcachesim/tracer/tracer.cpp
abhinav92003: Done.

core/lib/instrument_api.h
abhinav92003: Done.

suite/tests/linux/thread.c
abhinav92003: Done.

core/synch.c
abhinav92003: > The separate PR would go in before this one of course. But this is a direct bug fix of arm synchall, which is a separate thing from the new callback. IHMO the arm synchall bug should have its own issue filed, and its own separate fix.

Oh that makes sense. I can file a separate issue and PR for this.

Earlier, I thought you were suggesting leaving out the stolen reg restore completely for now, perhaps because we haven't yet figured out the stolen reg bug completely (pending #4460 and #4495)

> Maybe this synchall stolen reg restore could be bundled into #4495 and we leave it open for the further fix for synchall and for signals, rather than a separate issue.

I think it's better to fix the known part in a separate PR, as suggested above.

api/docs/release.dox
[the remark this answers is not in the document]
abhinav92003: I had to keep the new option, because `reset_at_nth_thread` doesn't fail without the stolen reg fix (due to the reason above). But I renamed it to `reset_at_created_thread_count` to highlight the difference from the existing option. Also added a comment with #4496 to remove the new option when we fix the assert failure with `reset_at_nth_thread 2`.

core/synch.c
abhinav92003: Separated out the stolen reg changes in PR #4498.

abhinav92003:
I separated out the stolen reg changes in PR #4498. After that is submitted, will update this PR's branch so that it contains just the synch flush callback related changes.

core/synch.c
abhinav92003: Done.

core/synch.c
derekbruening: > > Maybe this synchall stolen reg restore could be bundled into #4495 and we leave it open for the further fix for synchall and for signals, rather than a separate issue.
> 
> I think it's better to fix the known part in a separate PR, as suggested above.

FTR #4495 is an issue: I meant bundling the two synchall stolen reg problems into the same issue (fixed with separate PR's).  But a separate issue is fine too.

api/samples/cbr.c
```diff
@@ -232,10 +232,9 @@ insert(hash_table_t table, app_pc addr, cbr_state_t state)
 static void
 at_taken(app_pc src, app_pc targ)
 {
-    dr_mcontext_t mcontext = {
-        sizeof(mcontext),
-        DR_MC_ALL,
-    };
+    dr_mcontext_t mcontext;
```
derekbruening, line 235: These are good cleanups: I didn't realize there were so many like this in the samples and tests.  But given that there are quite of few of these mcontext changes I would separate these into their own PR (no need for an issue: iX branch) since they are logically distinct.  Cleaner history, simpler revert paths, etc.

api/docs/release.dox
```diff
@@ -173,6 +173,10 @@ Further non-compatibility-affecting changes include:
  - Added opnd_create_immed_double(), opnd_get_immed_double() and
    opnd_is_immed_double() to enable the creation and handling of double
    precision floating-point operands.
+ - Added -reset_at_created_thread_count option to allow triggering reset when
```
derekbruening, line 176: dr_flush_region_ex() should be listed here.

derekbruening, line 176: Please also add to api/docs/bt.dox which has a section on adaptive instrumentation where a statement about the callback being the way to do phased instrumentation in a race-free manner would be good.  (Search for dr_flush_region).

derekbruening, line 176: Just an official reminder to remove the PR #4498 code, to have it listed in the review comments for a "Done" etc. (not that github does a good job of listing which comments are not resolved).

derekbruening (approved):

api/samples/cbr.c
abhinav92003: Makes sense. Sent PR #4499 for review.

core/synch.c
abhinav92003: I ended up filing a separate issue for the two. #4497 (loading the stolen reg before reset) will need to be resolved more immediately (as it's actually causing some crashes) vs #4495 (saving the translated stolen reg value) needs more discussion and may need more time.

api/docs/release.dox
abhinav92003: Done.

abhinav92003: Done.

suite/tests/linux/thread.c
abhinav92003: Further discussion in #4496

clients/drcachesim/tracer/tracer.cpp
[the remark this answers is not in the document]
abhinav92003: Moved these changes to PR #4499

api/docs/release.dox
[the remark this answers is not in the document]
abhinav92003: Moved these changes to #4498

api/docs/release.dox
abhinav92003: Removed #4498 code and updated branch which removed the submitted #4499 from this PR. Resolved those comments too.

PR #4492, opened 2020-10-20
Fix leaks in drcachesim config file check

derekbruening:
The drcachesim config file checking code doesn't free two temporary
arrays.  We fix that here by switching to vectors.


derekbruening:
Decided to file an issue to cover this.  Will update the description on merge: #4493.

abhinav92003 (approved):

Issue #4497, opened 2020-10-28
SIGSEGV due to wrong stolen reg value after synchall


abhinav92003:
While working on PR #4491 which adds a synchall to drcachesim while enabling tracing, I encountered a crash due to the stolen reg having a value that doesn't seem like an address.

The segfaulting threads were redirected to the reset exit stub by DR at [1]. The `SIGSEGV` occurs in the exit stub shown below.

```
Thread xxx received signal SIGSEGV, Segmentation fault.
0x0000000046fc2458 in ?? ()
(gdb) x/50i 0x46fc2458
=> 0x46fc2458:	stp	x0, x1, [x28]
   0x46fc245c:	mov	x0, #0x4c20                	// #19488
   0x46fc2460:	movk	x0, #0xf7f0, lsl #16
   0x46fc2464:	movk	x0, #0xffff, lsl #32
   0x46fc2468:	ldr	x1, [x28, #64]
   0x46fc246c:	br	x1
   ...
```
The exit stub expects the stolen reg to be set up already, but it doesn't seem to be -- see `x28` below.
```
(gdb) i r 
x0             0xfffffffffffffffc  -4
x1             0xfffff45bcc80      281474781400192
...
x28            0x30000             196608
...
pc             0x46fc2458          0x46fc2458
```

The issue was fixed by adding a stolen reg store. Also there's a bug in the reset tests `linux.thread-reset` and `linux.clone-reset`, due to which they don't fail on AArch64 when they should.

Filing and fixing these separately from the PR #4491.

[1]: `translate_from_synchall_to_dispatch` https://github.com/DynamoRIO/dynamorio/blob/9bedb9c72d5808785e28d20df3ed1c932b555ac9/core/synch.c#L1816

Issue #4496, opened 2020-10-28
ASSERT failure using `reset_at_nth_thread 2` on test app


abhinav92003:
**Describe the bug**
DR crashes with an assert failure with the `-reset_at_nth_thread 2` option. The assert failure is different on AArch64 and x86.

**To Reproduce**
`bin64/drrun -enable_reset -reset_at_nth_thread 2  -- suite/tests/bin/linux.thread`

**Screenshots or Pasted Text**
On AArch64:
```
<proactive reset @ 1309 fragments>
Sideline thread started
<Application build/dynamorio/suite/tests/bin/linux.thread (33389).  Internal Error: DynamoRIO debug check failure: Not implemented @src/dynamorio/core/unix/signal_linux_aarch64.c:53 (0)
(Error occurred @1327 frags)
version 8.0.18562, custom build
-no_dynamic_options -code_api -stack_size 56K -signal_stack_size 32K -max_elide_jmp 0 -max_elide_call 0 -enable_reset -reset_at_nth_thread 2 -early_inject -emulate_brk -no_inline_ignored_syscalls -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct >
```

On x86:
```
 <proactive reset @ 1736 fragments>
<Application build/dynamorio/suite/tests/bin/linux.thread (2261478).  Internal Error: DynamoRIO debug check failure: src/dynamorio/core/dispatch.c:751 wherewasi == DR_WHERE_FCACHE || wherewasi == DR_WHERE_TRAMPOLINE || wherewasi == DR_WHERE_APP || (dcontext->go_native && wherewasi == DR_WHERE_DISPATCH)
(Error occurred @1736 frags)
version 8.0.18563, custom build
-no_dynamic_options -code_api -stack_size 56K -signal_stack_size 32K -max_elide_jmp 0 -max_elide_call 0 -reset_at_nth_thread 2 -early_inject -emulate_brk -no_inline_ignored_syscalls -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct >
```

**Versions**
Working on PR #4491 

**Additional context**
The same issue comes up with `-reset_at_created_thread_count 2` (this is a new flag tried while working on the above PR; this flag resets when created thread count reaches 2, unlike `reset_at_nth_thread` which checks the active thread count). But the failure does not come up with `-reset_at_created_thread_count 3`.
(Note that `reset_at_nth_thread 3` doesn't invoke the failing path at all)

[...]

PR #4498, opened 2020-10-28
i#4497: Add missing stolen reg load after synchall for AArch64.

abhinav92003:
Loads stolen reg with TLS base address before entering exit stub after synchall.

Fixes `linux.thread-reset` and `linux.clone-reset` tests which should fail on AArch64 due to missing stolen reg load but don't. Adds a new option to `reset_at_created_thread_count` and use that instead of `reset_at_fragment_count`. This was required because the latter requires to be tuned to make sure that reset happens when multiple threads are active; otherwise, the failing path is not invoked at all. Also, the ideal `reset_at_fragment_count` value is different for x86 and AArch64. Using `reset_at_created_thread_count` makes these tests more robust.

Fixes: #4497

### DIFF
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -173,10 +173,6 @@ Further non-compatibility-affecting changes include:
  - Added opnd_create_immed_double(), opnd_get_immed_double() and
    opnd_is_immed_double() to enable the creation and handling of double
    precision floating-point operands.
- - Added -reset_at_created_thread_count option to allow triggering reset when
-   created thread count reaches the given value. This is unlike
-   reset_at_nth_thread which checks the existing thread count. This option is
-   currently used to make Linux reset tests more robust.
 
 **************************************************
 <hr>
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -173,6 +173,10 @@ Further non-compatibility-affecting changes include:
  - Added opnd_create_immed_double(), opnd_get_immed_double() and
    opnd_is_immed_double() to enable the creation and handling of double
    precision floating-point operands.
+ - Added -reset_at_created_thread_count option to allow triggering reset when
+   created thread count reaches the given value. This is unlike
+   reset_at_nth_thread which checks the existing thread count. This option is
+   currently used to make Linux reset tests more robust.
 
 **************************************************
 <hr>
--- a/core/synch.c
+++ b/core/synch.c
@@ -1749,6 +1749,14 @@ translate_from_synchall_to_dispatch(thread_record_t *tr, thread_synch_state_t sy
                 arch_mcontext_reset_stolen_reg(dcontext, mc);
             }
         });
+        IF_AARCHXX({
+            // XXX i#4495: Consider saving stolen reg's application value.
+            set_stolen_reg_val(mc, (reg_t)os_get_dr_tls_base(dcontext));
+            // XXX: This path is tested by linux.thread-reset and linux.clone-reset.
+            // We just haven't run those on ARM yet.
+            IF_ARM(ASSERT_NOT_TESTED());
+        });
+
         /* We send all threads, regardless of whether was in DR or not, to
          * re-interp from translated cxt, to avoid having to handle stale
          * local state problems if we simply resumed.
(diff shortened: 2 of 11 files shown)


abhinav92003:
This PR contains the stolen reg changes separated out from PR #4491.

api/docs/release.dox
```diff
@@ -173,6 +173,10 @@ Further non-compatibility-affecting changes include:
  - Added opnd_create_immed_double(), opnd_get_immed_double() and
    opnd_is_immed_double() to enable the creation and handling of double
    precision floating-point operands.
+ - Added -reset_at_created_thread_count option to allow triggering reset when
```
derekbruening, line 176: Still confused: I am not seeing why we have to add a whole new option that is so similar to the existing -reset_at_nth_thread.  The use case here is test apps that we control so we can make the thread count exactly what we want.  I'm also not really understanding the difference: this seems like a duplicate option.

core/synch.c
```diff
@@ -1749,6 +1749,14 @@ translate_from_synchall_to_dispatch(thread_record_t *tr, thread_synch_state_t sy
                 arch_mcontext_reset_stolen_reg(dcontext, mc);
             }
         });
+        IF_AARCHXX({
+            // XXX i#4495: Consider saving stolen reg's application value.
```
derekbruening, line 1753: Include the key word "translated".  Otherwise the reader will find this confusing.

core/synch.c
```diff
@@ -1749,6 +1749,14 @@ translate_from_synchall_to_dispatch(thread_record_t *tr, thread_synch_state_t sy
...
+            set_stolen_reg_val(mc, (reg_t)os_get_dr_tls_base(dcontext));
```
derekbruening, line 1754: I think this code will make more sense if it's down after we point pc at get_reset_exit_stub.  It's directly related to that step.  If we changed our strategy to do sthg different we might not want this at new code at all: so best to keep it coupled to that.  I would also include a comment stating *why* we have to do this.

core/unix/signal.c
```diff
@@ -3423,6 +3423,7 @@ transfer_from_sig_handler_to_fcache_return(dcontext_t *dcontext, kernel_ucontext
      * from DR's handler.
      */
     ASSERT(get_sigcxt_stolen_reg(sc) != (reg_t)*get_dr_tls_base_addr());
+    // XXX i#4495: Consider saving stolen reg's application value.
```
derekbruening, line 3426: I would again differentiate the *translated* value.

api/docs/release.dox
abhinav92003: PR #4491 has a comment thread on this. Basically, `-reset_at_nth_thread n` triggers reset when the **active** thread count reaches `n`. Because of #4496, we cannot use n=2. We also cannot use n>2 because `linux.thread-reset` has at most two threads active at any given time (to keep synchronisation simple for deterministic print ordering).

So, we need `-reset_at_created_thread_count n` which triggers reset when the **created** thread count reaches `n`. Note that #4496 affects this option with n=2 too. But the good thing about this option is that we can use n=3 to reproduce the reset issue.

abhinav92003: Hmm.. I can probably put a workaround in suite/tests/linux/thread.c. Let me try.

derekbruening: This doesn't seem like the best solution: adding a whole new duplicate option just to work around an assert we want to fix anyway and to avoid a 3rd thread in a tiny test app.

abhinav92003: I modified the reset tests to have 3 concurrent threads so that we can use `-reset_at_nth_thread 3`. But the tests seem flaky now: `linux.thread-reset` fails 3/100 times, `linux.clone-reset` fails 29/100 times. 

This is surprising because previously, with `-reset_at_created_thread_count 3`, I didn't see any flakiness in 100 runs (all pass). But maybe the same race was present then too.

core/synch.c
abhinav92003: Done.

core/synch.c
abhinav92003: Done.

core/unix/signal.c

api/docs/release.dox
abhinav92003: So before, both the tests passed 100/100 times. Not sure what's causing the flakiness now (maybe something to do with the fact that we have more concurrent threads running now than before). 

I think it's better to just fix the missing pieces in #4496 first. Then if the flakiness is still relevant, I can get back to it

derekbruening (approved):

abhinav92003:
Closing this PR in favour of #4507, which uses a better approach to test the fix.

PR #4499, opened 2020-10-29
Optimise mcontext init by avoiding excessive zeroing.

abhinav92003:
Instead of using partial struct initialisation, simply initialise the required fields.

### DIFF
--- a/api/samples/cbr.c
+++ b/api/samples/cbr.c
@@ -232,10 +232,9 @@ insert(hash_table_t table, app_pc addr, cbr_state_t state)
 static void
 at_taken(app_pc src, app_pc targ)
 {
-    dr_mcontext_t mcontext = {
-        sizeof(mcontext),
-        DR_MC_ALL,
-    };
+    dr_mcontext_t mcontext;
+    mcontext.size = sizeof(mcontext);
+    mcontext.flags = DR_MC_ALL;
     void *drcontext = dr_get_current_drcontext();
 
     /*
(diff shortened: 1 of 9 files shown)


api/samples/cbr.c
```diff
@@ -232,10 +232,9 @@ insert(hash_table_t table, app_pc addr, cbr_state_t state)
 static void
 at_taken(app_pc src, app_pc targ)
 {
-    dr_mcontext_t mcontext = {
-        sizeof(mcontext),
-        DR_MC_ALL,
-    };
+    dr_mcontext_t mcontext;
```
derekbruening, line 235: optional: Do you think a reader might wonder why we're doing it this way?  If it were a small struct using {} is simpler.  We've found that the time taken to zero the large struct shows up as noticeable overhead for optimized clients.  Maybe a simple comment to make that clear?

derekbruening, line 235: commit message style: No "IX: " in the title (just on the branch name) -- so the title would just start with "Opt..."

derekbruening (approved):
Thank you for the cleanup.  We discovered the zeroing of the big struct was a perf hit a while back on Dr. Memory and somehow I thought we had already changed all our samples to set good examples.

api/samples/cbr.c
abhinav92003: Done. Added the comment in just `api/samples/cbr.c`, not in the (many) tests.

abhinav92003: Thanks. Done.

Issue #2440, opened 2017-05-26
AArch64 is missing XINST_CREATE and INSTR_CREATE macros


derekbruening:
Split from #1569.  AArch64 is missing a number of XINST_CREATE macros, and nearly all of its INSTR_CREATE macros.  This is hindering tool development: e.g., I'm adding some instrumentation to drcachesim that does an AND operation yet there's no XINST_CREATE_and_s.  This makes it difficult for new tool code to be cross-platform.  It would be best to avoid new code having to be disabled for AArch64.

egrimley:
Presumably the XINST_CREATE macros will be written by hand and are required more urgently than the INSTR_CREATE macros, which could be generated automatically.

There seem to be 27 XINST_CREATE macros for x86 and ARM. AArch64 currently has 19: compared to the other architectures, there are 9 missing, and one extra one, presumably added in error. So `XINST_CREATE_and` should be removed.

derekbruening:
Agreed, but the missing INSTR_CREATE are also causing some pain now: e.g. I'm adding a right-shift XINST_CREATE to use in drcachesim and it would be much easier if there were an INSTR_CREATE_lsr that auto-added whatever extend or shift extra params I want to ignore -- we don't want devs to have to either study the A64 manual and encoder or bail on supporting A64.

egrimley:
In A64, LSR (immediate) is an alias of UBFM, and LSR (register) is an alias of LSRV. So an XINST_CREATE_lsr would definitely be helpful.

c-rhodes:
XINST_CREATE macros were added for AArch64 in 3c49cc0.

fhahn:
@prakharbahuguna is working on this. I could not assign it to him, so I assigned myself to make sure it's not unassigned any more.

derekbruening:
Hit another case today where I want to put in a pre-index pair push but there is no support for that so I had to do this:
```
        instr_create_2dst_4src(dc, OP_stp,
                               opnd_create_base_disp(DR_REG_XSP, DR_REG_NULL, 0,
                                                     -2 * (int)sizeof(void *), OPSZ_16),
                               opnd_create_reg(DR_REG_XSP), opnd_create_reg(DR_REG_R29),
                               opnd_create_reg(DR_REG_LR), opnd_create_reg(DR_REG_XSP),
                               OPND_CREATE_INT8(-2 * (int)sizeof(void *))));
```

derekbruening:
[...]

Issue #4393, opened 2020-08-05
AArch64 decoder/encoder improvements


AssadHashmi:
The current method of generating decode/encode functions based on the `codec.txt` bitmap format has limitations in terms of:
- Flexibility when specifying instruction aliases and variants, e.g. `DC` and `IC` based on `SYS`, and `PRFM`. See #4388 and discussion at https://github.com/DynamoRIO/dynamorio/pull/4386#discussion_r462356954.
- Extensibility when new versions of the ISA are released, e.g. SVE.
- More automation in order to generate as much decode/encode logic as possible from an ISA spec.

The proposal is to use the publicly available XML specifications to replace `codec.txt` as the source from which to generate decode/encode functionality. Useful links:
https://alastairreid.github.io/ARM-v8a-xml-release/
https://developer.arm.com/architectures/instruction-sets/base-isas/a64
https://github.com/alastairreid/mra_tools

fhahn:
There are some script to auto-generate parts of codec.txt, tests and instr_create.h macros from an ISA description on the branch: https://github.com/DynamoRIO/dynamorio/commits/project-aarch64-generate-patterns

Not sure how well they aged though.

Issue #4388, opened 2020-07-30
[AArch64] Split OP_prfm opcode, or provide named constants for the sub-opcode immediate operand


derekbruening:
See the discussion at https://github.com/DynamoRIO/dynamorio/pull/4386#discussion_r462356954 regarding opcode philosophy of separate opcodes for separate semantics.

This issue covers either splitting the OP_prfm opcode into `OP_prfm_read_l1`, `OP_prfm_read_l1_nt`, etc., with a helper to determine whether an opcode is a `PRFM`, or leaving the single OP_prfm and providing a named enum where `DR_PRFM_READ_L1_NT`, etc. are set to equal their encoded 5-bit immediate value in the source operand.  Both approaches avoid the bit decoding  currently at https://github.com/DynamoRIO/dynamorio/blob/master/clients/drcachesim/tracer/instru.cpp#L82 and we'd want to update that code with the result.  We have pros and cons of both approaches; splitting the opcodes is a compatibilty break as well.  Whoever takes on the work gets to pick the approach I suppose.

derekbruening:
It's sounding like we have more votes for the single `OP_prfm` here: so we're considering the `DC` cache management types to be semantically different enough for separate opcodes, but not the prefetch types (nor the various SIMD operation variations we don't split inside the same opcode).

derekbruening:
Xref #4393 which is a pre-req for opcode splitting.

PR #4500, opened 2020-10-29
i#2440: Macros for creating conditional instructions

yury-khrustalev:
Add macros to create conditional instructions
Add functions conditional predicate operands
Add tests to verify the added functionality

Issue: #2440

### DIFF
--- a/core/ir/aarch64/instr_create.h
+++ b/core/ir/aarch64/instr_create.h
@@ -2613,7 +2613,7 @@ enum {
  *                (use opnd_create_cond(val) to create the operand).
  */
 #define INSTR_CREATE_cinv(dc, Rd, Rn, cond) \
-    instr_create_1dst_3src(dc, OP_csinv, Rd, Rn, Rn,  opnd_invert_cond(cond))
+    instr_create_1dst_3src(dc, OP_csinv, Rd, Rn, Rn, opnd_invert_cond(cond))
 
 /**
  * Creates a CNEG (Conditional Negate) instruction.
@@ -2645,9 +2645,9 @@ enum {
  * \param cond    A 4-bit immediate value for condition
  *                (use opnd_create_cond(val) to create the operand).
  */
-#define INSTR_CREATE_cset(dc, Rd, cond) \
-    instr_create_1dst_3src(dc, OP_csinc, Rd, \
-    OPND_CREATE_ZR(Rd), OPND_CREATE_ZR(Rd), opnd_invert_cond(cond))
+#define INSTR_CREATE_cset(dc, Rd, cond)                                              \
+    instr_create_1dst_3src(dc, OP_csinc, Rd, OPND_CREATE_ZR(Rd), OPND_CREATE_ZR(Rd), \
+                           opnd_invert_cond(cond))
 
 /**
  * Creates a CSETM (Conditional Set Mask) instruction.
@@ -2656,9 +2656,9 @@ enum {
  * \param cond    A 4-bit immediate value for condition
  *                (use opnd_create_cond(val) to create the operand).
  */
-#define INSTR_CREATE_csetm(dc, Rd, cond) \
-    instr_create_1dst_3src(dc, OP_csinv, Rd, \
-    OPND_CREATE_ZR(Rd), OPND_CREATE_ZR(Rd), opnd_invert_cond(cond))
+#define INSTR_CREATE_csetm(dc, Rd, cond)                                             \
+    instr_create_1dst_3src(dc, OP_csinv, Rd, OPND_CREATE_ZR(Rd), OPND_CREATE_ZR(Rd), \
+                           opnd_invert_cond(cond))
 
 /**
  * Creates a CSINC (Conditional Select Increment) instruction.
--- a/core/ir/aarch64/instr_create.h
+++ b/core/ir/aarch64/instr_create.h
@@ -2569,10 +2569,8 @@ enum {
  * Creates a CCMN (Conditional Compare Negative) instruction.
  * \param dc      The void * dcontext used to allocate memory for the instr_t.
  * \param Rn      The source register.
- * \param Op      Either a 5-bit immediate (use opnd_create_immed_uint(val, OPSZ_5b)
- *                to create the operand) or a source register.
- * \param nzcv    The 4 bit flag bit specifier
- *                (use opnd_create_immed_uint(val, OPSZ_4b) to create the operand).
+ * \param Op      Either a 5-bit immediate or a source register.
+ * \param nzcv    The 4 bit flag bit specifier.
  * \param cond    A 4-bit immediate value for condition
  *                (use opnd_create_cond(val) to create the operand).
  */
@@ -2583,10 +2581,8 @@ enum {
  * Creates a CCMP (Conditional Compare) instruction.
  * \param dc      The void * dcontext used to allocate memory for the instr_t.
  * \param Rn      The source register.
- * \param Op      Either a 5-bit immediate (use opnd_create_immed_uint(val, OPSZ_5b)
- *                to create the operand) or a source register.
- * \param nzcv    The 4 bit flag bit specifier
- *                (use opnd_create_immed_uint(val, OPSZ_4b) to create the operand).
+ * \param Op      Either a 5-bit immediate or a source register.
+ * \param nzcv    The 4 bit flag bit specifier.
  * \param cond    A 4-bit immediate value for condition
  *                (use opnd_create_cond(val) to create the operand).
  */
--- a/core/ir/aarch64/instr_create.h
+++ b/core/ir/aarch64/instr_create.h
@@ -2565,5 +2565,136 @@ enum {
  */
 #define INSTR_CREATE_fmov_scalar_imm(dc, Rd, f) instr_create_1dst_1src(dc, OP_fmov, Rd, f)
 
+/**
+ * Creates a CCMN (Conditional Compare Negative) instruction.
+ * \param dc      The void * dcontext used to allocate memory for the instr_t.
+ * \param Rn      The source register.
+ * \param Op      Either a 5-bit immediate (use opnd_create_immed_uint(val, OPSZ_5b)
+ *                to create the operand) or a source register.
+ * \param nzcv    The 4 bit flag bit specifier
+ *                (use opnd_create_immed_uint(val, OPSZ_4b) to create the operand).
+ * \param cond    A 4-bit immediate value for condition
+ *                (use opnd_create_cond(val) to create the operand).
+ */
+#define INSTR_CREATE_ccmn(dc, Rn, Op, nzcv, cond) \
+    instr_create_0dst_4src(dc, OP_ccmn, Rn, Op, nzcv, cond)
+
+/**
+ * Creates a CCMP (Conditional Compare) instruction.
+ * \param dc      The void * dcontext used to allocate memory for the instr_t.
+ * \param Rn      The source register.
+ * \param Op      Either a 5-bit immediate (use opnd_create_immed_uint(val, OPSZ_5b)
+ *                to create the operand) or a source register.
+ * \param nzcv    The 4 bit flag bit specifier
+ *                (use opnd_create_immed_uint(val, OPSZ_4b) to create the operand).
+ * \param cond    A 4-bit immediate value for condition
+ *                (use opnd_create_cond(val) to create the operand).
+ */
+#define INSTR_CREATE_ccmp(dc, Rn, Op, nzcv, cond) \
+    instr_create_0dst_4src(dc, OP_ccmp, Rn, Op, nzcv, cond)
+
+/**
+ * Creates a CINC (Conditional Increment) instruction.
+ * \param dc      The void * dcontext used to allocate memory for the instr_t.
+ * \param Rd      The output register.
+ * \param Rn      The input register.
+ * \param cond    A 4-bit immediate value for condition
+ *                (use opnd_create_cond(val) to create the operand).
+ */
+#define INSTR_CREATE_cinc(dc, Rd, Rn, cond) \
+    instr_create_1dst_3src(dc, OP_csinc, Rd, Rn, Rn, opnd_invert_cond(cond))
+
+/**
+ * Creates a CINV (Conditional Invert) instruction.
+ * \param dc      The void * dcontext used to allocate memory for the instr_t.
+ * \param Rd      The output register.
+ * \param Rn      The input register.
+ * \param cond    A 4-bit immediate value for condition
+ *                (use opnd_create_cond(val) to create the operand).
+ */
+#define INSTR_CREATE_cinv(dc, Rd, Rn, cond) \
+    instr_create_1dst_3src(dc, OP_csinv, Rd, Rn, Rn,  opnd_invert_cond(cond))
+
+/**
+ * Creates a CNEG (Conditional Negate) instruction.
+ * \param dc      The void * dcontext used to allocate memory for the instr_t.
+ * \param Rd      The output register.
+ * \param Rn      The input register.
+ * \param cond    A 4-bit immediate value for condition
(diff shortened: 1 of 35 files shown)


yury-khrustalev:
I'm not sure some of the build failures are related to the patch, can you please have a look?

```
The CMAKE_C_COMPILER:

    arm-linux-androideabi-gcc

  is not a full path and was not found in the PATH.
```

core/ir/opnd_shared.c
```diff
@@ -743,7 +743,85 @@ opnd_create_base_disp_aarch64(reg_id_t base_reg, reg_id_t index_reg,
         CLIENT_ASSERT(false, "opnd_create_base_disp_aarch64: invalid extend type");
     return opnd;
 }
-#endif
+
+opnd_t
+opnd_create_cond(int cond)
+{
+    /* FIXME i#1569: Move definition of dr_pred_type_t to opnd.h for AArch64
```
AssadHashmi, line 750: Please create an issue separate from the AArch64 master issue #1569 to track this.

core/ir/opnd_shared.c
```diff
@@ -743,7 +743,85 @@ opnd_create_base_disp_aarch64(reg_id_t base_reg, reg_id_t index_reg,
...
+    default:
+        CLIENT_ASSERT(false, "invalid condition constant");
+        return opnd_create_null();
+    }
+}
+
+int
+opnd_get_cond(opnd_t opnd)
+{
+    /* FIXME i#1569: Move definition of dr_pred_type_t to opnd.h for AArch64
```
AssadHashmi, line 778: As above.

suite/tests/api/ir_aarch64.c
```diff
@@ -4332,6 +4367,661 @@ test_sys(void *dc)
...
+                            DR_PRED_GT, DR_PRED_LE, DR_PRED_AL, DR_PRED_NV };
+
+static void
+test_ccmn_imm(void *dc)
+{
+    /* ccmn <Xn>, #<imm>, #<nzcv>, <cond> */
+    for (reg_id_t rn = DR_REG_X0; rn <= DR_REG_XZR; rn++) {
+        if (rn == DR_REG_XSP) {
+            continue;
+        }
```
AssadHashmi, line 4382: Does the code format checker insist on having braces for single-line `if` statements?

core/ir/opnd_shared.c
yury-khrustalev: Created: #4502.

suite/tests/api/ir_aarch64.c
yury-khrustalev: I'm not sure if code formatter insists on it. I believe using braces improves readability of code, so I'd rather keep it like this.

core/ir/opnd.h
```diff
@@ -1777,6 +1777,34 @@ opnd_t
...
+
+DR_API
+/**
+ * Assumes the operand is an unsigned immediate 4-bit integer and returns an
+ * operand which contains inverted condition (see #dr_pred_type_t).
+ * \note Does not support AL and NV conditions.
+ * \note AArch64-only.
+ */
+opnd_t
+opnd_invert_cond(opnd_t opnd);
```
derekbruening, line 1806: I think this needs some thought as to how to integrate with existing interfaces and code, especially with ARM code.

The API already has `instr_invert_predicate`. I would agree it seems less of an instr-level operation, but it would be confusing to have two different sets of interfaces.

The API also already has predicates passed as dr_pred_type_t to things like `XINST_CREATE_jump_cond` in arm/instr_create.h.  If a new type of operand, even if it's an immediate underneath is created, it would be best to use it everywhere.  Whatever is decided, it seems best to have the same approach for all platforms, rather than special AArch64-only functions that don't match ARM functions yet do the same thing.  That makes for confusing interfaces and complex cross-platform clients.

derekbruening (changes requested):

core/ir/opnd.h
yury-khrustalev: It would be a very big change to move all the condition-related interfaces that currently exist on the level of instructions to the level of operands. I don't think I know enough about other platforms to come up with a uniform solution, and I don't know if such change actually makes sense for x86 (after all `instr_invert_predicate` was create for some reason).

Anyway, I'd like to do it in steps rather than in a single big pull request which is very difficult to merge. At least some guidance, suggestions or ideas would be helpful.

I would start with #4502, but again as a separate patch.

yury-khrustalev: Right now I could remove `opnd_invert_cond` and use `instr_invert_predicate` around `instr_create_...` to flip the operand, however I'm not sure if `instr_invert_predicate` is actually implemented for AArch64.

derekbruening: See https://dynamorio.org/dynamorio_docs/API_BT.html#sec_predication which lists one key current use of predicates/conditions:
+ An instruction-level modifier that is not an operand: `instr_{get,set}_predicate`; `INSTR_PRED` short form.  While perhaps this could have been an operand (along with other things like condition code effects: tradeoffs with matching ISA conventions), it is not.  Yet we want to be able to invert these predicates too.  The current `instr_invert_predicate` does not actually operate on `instr_t`: it operates on `dr_pred_type_t` and thus operates on the instr-level predicates as well as operand-level.

Another current use, similar to the macros added here:
+ `XINST_CREATE_jump_cond()`.  This, on x86, arm, and aarch64, takes in a `dr_pred_type_t`-typed value.  I would think that the signature/operation of this instruction should match things like the `INSTR_CREATE_ccmn` added here.  One shouldn't take a `dr_pred_type_t` while the other takes an `opnd_t`.  Maybe the argument could be made that we should change the `XINST_CREATE_jump_cond` to take an `opnd_t` of this new type: though note that the condition turns into different opcodes on x86, the instruction-level predicate field on arm, and an operand on aarch64, so they are not all treating it as an operand.  See also other high-level discussions on what should constitute an operand versus an opcode or sub-opcode: https://github.com/DynamoRIO/dynamorio/pull/4386#discussion_r462356954.  Perhaps these aarch64 comparison should be using separate opcodes, rather than operands that change behavior which goes against what operands are supposed to do.  Having a single opcode doing all these different comparisons is sort of like having one opcode "arithmetic" with an operand that says whether it's an add or a subtract or whatnot, instead of separate add and subtract opcodes.

derekbruening: Continuing in the thread of the last few sentences: xref other issues where we want to split up the current aarch64 opcodes which have behavior-controlling integer operands which are perhaps better as separate opcodes rather than operands: #4388, #4393.

core/ir/aarch64/instr_create.h
```diff
@@ -2565,5 +2565,136 @@ enum {
  */
 #define INSTR_CREATE_fmov_scalar_imm(dc, Rd, f) instr_create_1dst_1src(dc, OP_fmov, Rd, f)
 
+/**
+ * Creates a CCMN (Conditional Compare Negative) instruction.
+ * \param dc      The void * dcontext used to allocate memory for the instr_t.
+ * \param Rn      The source register.
+ * \param Op      Either a 5-bit immediate (use opnd_create_immed_uint(val, OPSZ_5b)
```
derekbruening, line 2572: Does this really need to be created with the `OPSZ_5b` size?  Could we change that and have the encoder handle any declared size?  Unless there are multiple encoding variants that differ in immediate sizes, the encoder shouldn't care so long as the value fits in 5 bits?  It would be best for usability to not require anyone to specify all these precise bit counts.  If a bit count does have to specified, there should be a named constant like `OPSZ_ccmn` or sthg like x86 has for `OPSZ_xlat`, `OPSZ_lgdt`, etc., and in addition maybe this macro should take just the integer and supply the size or set the size of the opnd before passing it on or sthg.

core/ir/opnd.h
yury-khrustalev: > One shouldn't take a dr_pred_type_t while the other takes an opnd_t.

I think it would be much better if all `INSTR_CREATE_` macros would take values instead of operands (e.g. `DR_REG_` constants instead of `opnd_create_reg(...)`), but most of such macros take `opnd_t`, that is why I implemented new macros in such a way -- to follow the common approach. I believe such choice was made for better portability of macros across different architectures. So, I agree that `XINST_CREATE_jump_cond` is an exception, from this point of view.

Conditional operands in AArch64 work differently from other platforms, that is why I thought having an operand-level function for AArch64 was a better idea than extending `instr_invert_predicate` (I suggest to discuss it in #4504).

Maybe we shouldn't make inherently different architectures conform to some common notion of how instructions "should" work?

core/ir/aarch64/instr_create.h
yury-khrustalev: For these instructions we need an unsigned integer immediate operand. The existing API for creating such operands (`opnd_create_immed_uint`) requires to specify an `opnd_size_t` value. How do you suggest to create an operand without using `opnd_create_immed_uint`? Or do you suggest to change `opnd_create_immed_uint`?

derekbruening: Allow any size.  Or additionally maybe add `opnd_create_immed_uint_unsized` and internally it sets OPSZ_4 or sthg and the a64 encoder ignores the size since it has no encoding decisions based on size.

yury-khrustalev: I thought it is helpful to give an example of how to create a correct operand. It would be better if this macro took just an integer value, but most of `INSTR_CREATE_` macros accept operands. I can remove `OPSZ_5b` from the doc string.

core/ir/opnd.h
derekbruening: > Maybe we shouldn't make inherently different architectures conform to some common notion of how instructions "should" work?

The whole point of an IR is to abstract away underlying machine encoding quirks.  It is a major burden on client writers if you impose on them different IR conventions for different architectures.  Plus, this predicate here is not anything inherently different from other architectures: what's different is that the ISA *manual* presents it as an operand rather than part of the opcode, while other architectures would present it as part of the opcode with zero changes in how it works.  Yes, some things are fundamentally different, like universal predication on ARM, but we should make all the other things as similar as we can in the IR.

I would vote for not having any "magic behavior modifiers" as integer operands anywhere in DR's IR: they should be part of the opcode.  The only integer operand should be an actual integer value, not an encoding of how to change the behavior.  The behavior comes from the opcode, not the operands.  So this CCMN and CCMP would have their opcodes split up.  Just like the SYS opcode should be split up as discussed elsewhere.  Just like on ARM we split up the SIMD opcodes and have `OP_vld2_lane_8` instead of just one `OP_vld2` with magic operands encoding behavior and then special operand types for each of those integer-encoded operands.

@AssadHashmi WDYT about splitting up CCMN and CCMP?

derekbruening: > I think it would be much better if all INSTR_CREATE_ macros would take values instead of operands (e.g. DR_REG_ constants instead of opnd_create_reg(...)),

This breaks down for complex operands like memory references.  Some instructions can take either a register or a memory reference for the same operand slot.

core/ir/aarch64/instr_create.h
derekbruening: > I can remove OPSZ_5b from the doc string.

SGTM.  The other suggestions (like an _unsized version) are separate/larger scope from this PR.  I'm pretty sure the a64 encoder today ignores the size completely, so the user could pass OPSZ_NA or sthg.

core/ir/opnd.h
yury-khrustalev: > WDYT about splitting up CCMN and CCMP?

Why only these instructions? All conditional instructions in AArch64 (all mentioned in this PR) have `cond` operand.

Also, in my opinion, inventing non-documented opcodes may create extra confusion. If some notion is not common to the rest of the architectures it's not the reason to disallow low-level access _as documented_. I think in DynamoRIO, when we work at the level of instructions and operands, we don't really have the benefit of "write once run everywhere". In my understanding, the point of the tool is to allow working on the lowest possible level.

yury-khrustalev: >  The behavior comes from the opcode, not the operands.

How about compare and swap instructions (e.g. CAS)?  The behaviour depends on the value of the register being equal to the value loaded from memory. It is similar to the conditional instructions, although in the latter case the comparison has happened _before_.

algrant-arm: It might be useful to think of the pseudocode that specifies instruction behaviors. CCMP is specified by one pseudocode function, and the operands are inputs to that function. That single behavior function acts uniformly on all values of the operands. So it makes sense to treat it as one instruction, rather than single out one operand and split it into one instruction for each value of that operand.

The situation with SYS instructions is very different (similar to CDP instructions in 32-bit Arm) - they really are separate functions, selected by a function code, and it does make much more sense to split them out, at least when their specific functions are being differently handled by DynamoRIO.

AssadHashmi: > @AssadHashmi WDYT about splitting up CCMN and CCMP?

If I understand correctly we have 3 choices:

1 - Create a new operand for `cond`.
2 - Split out the opcodes into specific `cond` types, e.g. `ccmp_eq`, `ccmp_ne` etc.
3 - Use existing `dr_pred_type_t` as in `XINST_CREATE_jump_cond` (alias for `INSTR_CREATE_bcond`).

1 requires a non-trivial amount of work for a type which has essentially the same intent as `dr_pred_type_t`.

2 is not consistent with AArch64's existing codec. We may run into problems as I did in https://github.com/DynamoRIO/dynamorio/pull/4386

3 would be my choice. It makes use of an existing type with the same intent which has already been implemented and does not involve a lot of codec reworking.
e.g.
```
#define INSTR_CREATE_ccmp(dc, Rn, Op, nzcv, cond) \
    (INSTR_PRED(instr_create_0dst_3src(dc, OP_ccmp, Rn, Op, nzcv), (cond)))
```
Which users can call with:
```
INSTR_CREATE_ccmp(dc, opnd_create_reg(DR_REG_X1),
        opnd_create_immed_int(0b01010, OPSZ_5b),
        opnd_create_immed_int(0b1111, OPSZ_4b), DR_PRED_EQ);
```

derekbruening: @AssadHashmi highlights what is perhaps the main issue here: using a simple and consistent set of abstractions and types in the IR.  If we take each encoded behavior immediate integer that AArch64 presents as an operand (but I would call a sub-opcode and another ISA might well make part of the opcode, such as x86 conditional moves) and place it as a separate operand in DR's IR, we then need to make a new operand type with a new set of rules and enums and handling code and introduce a new concept to tool writers for what could logically fit into existing concepts.  We could end up with a dozen new operand types without much benefit but incurring non-trivial costs.  Using an existing IR concept keeps the IR simpler and cleaner and more consistent.

> we don't really have the benefit of "write once run everywhere".

We can make it as easy as possible, though.  It does make a difference: each IR choice that lets a tool or analysis library be more platform-agnostic makes tool writing easier.

> In my understanding, the point of the tool is to allow working on the lowest possible level.

None of any of the discussion here has any bearing on whether a tool writer can generate any precise machine code desired: we're only talking about how to represent ISA concepts in the DR IR.  None of the choices affect *whether* the ISA concept can be represented or whether the tool writer can generate it.  And while the IR should be more abstracted and cross-platform, we can and do have layers that are closer to the assembly conventions of that ISA: the INSTR_CREATE_ macros try to look more like assembly, disassembly has both an abstracted DR format and an ISA format, and we'd love to have an even closer layer where assembler strings are converted into IR (can't seem to find it in the tracker...#4510) when writing code for a specific machine.  But code analyzing and manipulating at the IR level should have as few types and special operand concepts as possible.

yury-khrustalev:
To sum up the discussion, I would like to freeze this PR until I have time to submit a patch for #4504 (or someone else does it).
The idea is to use `instr_invert_predicate` function instead of `opnd_invert_cond` proposed here (which will be remove from the patch). I hope this is OK.

Issue #1569, opened 2014-12-12
add AArch64 support: master issue


derekbruening:
Split from issue #1551 as our initial focus is just AArch32.

toshipiazza:
In api/samples/CMakeLists.txt I noticed that bbbuf.c is the only sample listed as not yet ported from AArch32 to AArch64. However #1982 modifies bbbuf.c to use drx_buf, to which AArch64 support was added (not by me).

Would someone with an AArch64 setup be able to verify it works and modify CMakelists.txt as appropriate?

fhahn:
bbbuf was enabled on AArch64 a while ago by fa7e5c55d9a2526ae566b1e082ec23033b705196

toshipiazza:
`drx_buf_insert_buf_memcpy()` and the related tests in `drx_buf-test.dll.c` were disabled for Aarch64 because of #2440. However, it looks like this was partially fixed by https://github.com/DynamoRIO/dynamorio/commit/3c49cc083dde66d47a58c29f8ba1d7996954dc68, enough so that the relevant `XINST_CREATE_` macros are now there. Would anyone with an Aarch64 build be able to build/verify `drx_buf_insert_buf_memcpy()` works on Aarch64?

fhahn:
@toshipiazza `drx_buf-test.dll.c` fails during execution, with all `#ifndef AARCH64` removed #2520

derekbruening:
Unfortunately we didn't have auto-links for partial commits so I'm pasting in the many commits contributing to the port so far:

2017-07-20 c908cddd i#1569 AArch64: Make encodings easier to specify. (#2549)
2017-07-12 654f7b2e i#1569 AArch64: Fix bug in encoding (SIMD structure load/store). (#2527)
2017-07-12 8d4d5edc i#1569 AArch64: Initialise variables in encoder functions. (#2528)
2017-07-11 69913cb2 i#1569 AArch64: In encoder/decoder, use all bits provided by pattern. (#2517)
2017-07-10 aec49770 i#1569 AArch64: Follow-up to 428b896: Use opndset for ADR and ADRP. (#2516)
2017-07-10 bc483ace i#1569 AArch64: Follow-up to 3275b81: check X30 opnd in encoder. (#2515)
2017-06-21 662d0940 i#1569 AArch64: Fix uninitialized variable error when building with GCC 7 on AArch64. (#2480)
[...]

PR #4503, opened 2020-10-29
i#1569 AArch64: Enable -steal_reg_at_reset for AArch64

derekbruening:
Enables the diagnostic option -steal_reg_at_reset for AArch64,
generalizing the existing ARM code.

Switches -reset_at_fragment_count to work in release build by using
the sum of the release stats num_bbs and num_traces.

Adds a release-build syslog for an informational notification when any
reset occurs.

Adds a test of -steal_reg_at_reset.

Issue: #1569

### DIFF
--- a/core/synch.c
+++ b/core/synch.c
@@ -1817,6 +1817,7 @@ translate_from_synchall_to_dispatch(thread_record_t *tr, thread_synch_state_t sy
         LOG(GLOBAL, LOG_CACHE, 2, "\tsent to reset exit stub " PFX "\n", mc->pc);
         /* TODO i#4497: Replace with the official fix from PR#4498. */
         IF_AARCHXX(set_stolen_reg_val(mc, (reg_t)os_get_dr_tls_base(dcontext)));
+        IF_AARCHXX(ASSERT_NOT_TESTED()); /* PR#4498 will improve test coverage. */
 #ifdef WINDOWS
         /* i#25: we could have interrupted thread in DR, where has priv fls data
          * in TEB, and fcache_return blindly copies into app fls: so swap to app
--- a/core/synch.c
+++ b/core/synch.c
@@ -1714,7 +1714,7 @@ translate_from_synchall_to_dispatch(thread_record_t *tr, thread_synch_state_t sy
             free_cxt = false;
         }
 #endif
-        IF_ARM({
+        IF_AARCHXX({
             if (INTERNAL_OPTION(steal_reg_at_reset) != 0) {
                 /* We don't want to translate, just update the stolen reg values */
                 arch_mcontext_reset_stolen_reg(dcontext, mc);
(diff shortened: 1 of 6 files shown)


core/synch.c
```diff
@@ -1815,6 +1815,8 @@ translate_from_synchall_to_dispatch(thread_record_t *tr, thread_synch_state_t sy
          */
         mc->pc = (app_pc)get_reset_exit_stub(dcontext);
         LOG(GLOBAL, LOG_CACHE, 2, "\tsent to reset exit stub " PFX "\n", mc->pc);
+        /* TODO i#4497: Replace with the official fix from PR#4498. */
```
abhinav92003, line 1818: Well, the line below is already the same fix used in PR#4498 too.

suite/tests/CMakeLists.txt
```diff
@@ -4073,6 +4073,17 @@ if (UNIX)
...
+  if (AARCHXX)
+    # Test our diagnostic option -steal_reg_at_reset, which is also a stress
+    # test of reset and register stealing.
+    if (ARM)
+      set(new_reg 9)
+    else ()
+      set(new_reg 29)
+    endif ()
+    torunonly(linux.steal-switch linux.thread linux/thread.c
+      "-enable_reset -reset_at_fragment_count 100 -steal_reg_at_reset ${new_reg}" "")
```
abhinav92003, line 4085: This test may have the same issue like `linux.thread-reset` and `linux.clone-reset`, with triggering reset at the right time. A reset event will execute `translate_from_synchall_to_dispatch` (therefore the new code added in `core/synch.c`) only if >1 threads are active. So, the `n` in `-reset_at_fragment_count n` needs to be hand-tuned. I'm trying to fix this issue in these tests in #4498 and #4496).

Does the test fail without the stolen reg store in core/synch.c?

derekbruening: For my purposes I don't care about the synchall path: I just want to replace the entire code cache and generated code and stubs that were using r28 with new code using r29.  Any fragment count will do.  If the test were to be improved, it would be to add app code that checks all register values, but I don't think it's worth the effort for a sanity check test that just makes sure the base option continues to work.

core/synch.c
derekbruening: I figured PR #4498 would go in before this one.  Though at this point maybe not.  Are you suggesting to remove this fix if the tests on the tx1 and jenkins don't hit it and assume PR #4498 will be in before anyone happens to hit it on the new test on a new platform?

abhinav92003: PR #4498 is currently blocked on #4496 (found another `ASSERT_NOT_IMPLEMENTED` for AArch64 after the `save_fpstate` one).

It should be okay to include the fix in this PR if it helps avoid anyone else hitting and debugging it. Though, do add an `ASSERT_NOT_TESTED` for #4498 work on improving the tests to hit this path.

abhinav92003 (approved):

core/synch.c
derekbruening: OK, I'm ok either way.  I'll leave this line plus an ASSERT_NOT_TESTED as requested.

abhinav92003: Would be good to xref `Issue #4497` in the final commit description.

derekbruening: Will do.

Issue #2100, opened 2016-12-09
tools/libdynamorio.so-gdb.py causes "RuntimeError: Cannot access memory at address 0x712fd867"


thorstent:
I'm running DynamoRIO on a fresh Ubuntu 16.10 64bit VM in VirtualBox (Host is Windows).

I am using DynamoRIO 6.2 binary release. But the problem persists with the latest nightly 6.2.17144-0x12112de

My ~/.gdbinit file constists of these two lines:

```
add-auto-load-safe-path /home/ttarrach/Downloads/DynamoRIO-Linux-6.2.0-2/lib64/debug/libdynamorio.so.6.2-gdb.py
add-auto-load-safe-path /home/ttarrach/Downloads/DynamoRIO-Linux-6.2.17144-0x12112de/lib64/debug/libdynamorio.so.6.2-gdb.py
```

The minimal steps to reproduce this problem are:
1. start GDB: `~/Downloads/DynamoRIO-Linux-6.2.0-2$ gdb --args bin64/drrun -debug -- ls`
2. type `run` into the GDB command prompt

Here is the complete output:

```
GNU gdb (Ubuntu 7.11.90.20161005-0ubuntu1) 7.11.90.20161005-git
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from bin64/drrun...Reading symbols from /home/ttarrach/Downloads/DynamoRIO-Linux-6.2.0-2/bin64/drrun.debug...done.
done.
(gdb) run
Starting program: /home/ttarrach/Downloads/DynamoRIO-Linux-6.2.0-2/bin64/drrun -debug -- ls
process 2728 is executing new program: /home/ttarrach/Downloads/DynamoRIO-Linux-6.2.0-2/lib64/debug/libdynamorio.so.6.2
Loading gdb scripts for debugging DynamoRIO...
[...]

PR #4505, opened 2020-10-30
Add gdb scripts for memquery and drsymload

derekbruening:
Adds two gdb python scripts I've developed that may be useful to others:

1) drsymload: loads DR symbols regardless of gdb's current state,
which may include having DR symbols at the wrong address.  It does
this by reading /proc/self/maps and running objdump on
libdynamorio.so.  Ideally this would be integrated into a revived
libdynamorio.so-gdb.py: that's part of #2100.

2) memquery: prints the /proc/self/maps line for a given address.
I'm still shocked gdb doesn't provide such a command natively.

Issue: #2100

### DIFF
--- a/tools/gdb-scripts/gdb-drsymload.py
+++ b/tools/gdb-scripts/gdb-drsymload.py
@@ -1,0 +1,84 @@
+#!/usr/bin/env python
+
+# **********************************************************
+# Copyright (c) 2020 Google, Inc.    All rights reserved.
+# **********************************************************
+
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# * Redistributions of source code must retain the above copyright notice,
+#   this list of conditions and the following disclaimer.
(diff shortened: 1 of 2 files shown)


tools/gdb-scripts/gdb-drsymload.py
```diff
@@ -0,0 +1,84 @@
...
+from pathlib import Path
+import os.path
+
+class DRSymLoadCommand (gdb.Command):
+    """Loads symbols for libdynamorio.so"""
+    def __init__ (self):
+        super (DRSymLoadCommand, self).__init__ ("drsymload",
+                                                gdb.COMMAND_DATA,
+                                                gdb.COMPLETE_FILENAME)
+    def invoke(self, arg, from_tty):
```
abhinav92003, line 57: May be better to name them unused_arg and unused_from_tty, to document that they are unused intentially..

abhinav92003 (approved):

PR #4506, opened 2020-10-30
Fix uninitialized var in drmemtrace offline instru

derekbruening:
While investigating #4460 I found that reg_ptr_used in
insert_save_addr is uninitialized locally and insert_obtain_addr only
writes it when true, leaving it uninitialized for the false case: thus
we may re-instate the buffer pointer in cases where we don't need to.
I believe this is only a performance issue.

Issue: #4460


abhinav92003 (approved):

derekbruening:
I was wrong here: drutil_insert_get_mem_addr_ex() sets *scratch_used in all cases.

Issue #4488, opened 2020-10-15
x87 floating point immediates should be double precision


derekbruening:
DR's decoder uses single-precision floats for the x87 immediates from instructions like `fldl2t`.  However, these instructions push double extended-precision format values for the 80-bit x87.  It seems best for DR to represent them as `double` rather than `float`.  (I don't know that it's worth trying to use `long double` or something for a 10-byte format where supported.)

Xref #4408 where for AArch64 we are adding `double` to `opnd_t`.

derekbruening:
This issue also covers potentially changing the layout of 32-bit ARM and x86 opnd_t.  We could decide to break binary compability, have the opnd_t size increase to 16 bytes with a 4-byte pad to align the "value" union containing the double to the proper 8 bytes, change the black box struct layout, change the heap allocator expected size if there was a 12-byte size class, and bump the version to 8.90 (with the next 9.0).

We'd do this for just being cleaner, but also to avoid problems in case any ARM or other instructions involving doubles require 8-byte alignment: is that the case?

PR #4507, opened 2020-11-01
i#4496: Fix reset issues on AArch64 and x86

### LINKED ISSUE #4496: ASSERT failure using `reset_at_nth_thread 2` on test app

### LINKED ISSUE #4497: SIGSEGV due to wrong stolen reg value after synchall

abhinav92003:
Changes reset trigger for `linux.thread-reset` and `linux.clone-reset` tests to `-reset_at_nth_thread`. To increase coverage of these tests, we need to trigger reset at a time when multiple threads are active. This is difficult to do using the existing` -reset_at_fragment_count` option in tests, which may need to be tuned regularly to its ideal value, which also differs based on platform.

Implements some missing AArch64 pieces for `dynamorio_sigreturn`.

Relaxes dispatch assert to allow threads to enter dispatch after reset. The current condition allows it only when sending threads native.

Fixes #4496
Fixes #4497

### DIFF
--- a/core/dispatch.c
+++ b/core/dispatch.c
@@ -752,9 +752,10 @@ dispatch_enter_dynamorio(dcontext_t *dcontext)
            wherewasi == DR_WHERE_APP ||
            /* If the thread was waiting at check_wait_at_safe_point when getting
             * suspended, we were in dispatch (ref i#3427). We will be here after the
-            * thread's context is being reset before sending it native.
+            * thread's context is being reset proactively (due to some -reset_at_*
+            * option) or before sending it native.
             */
-           (dcontext->go_native && wherewasi == DR_WHERE_DISPATCH));
+           wherewasi == DR_WHERE_DISPATCH);
     dcontext->whereami = DR_WHERE_DISPATCH;
     ASSERT_LOCAL_HEAP_UNPROTECTED(dcontext);
     ASSERT(check_should_be_protected(DATASEC_RARELY_PROT));
--- a/core/dispatch.c
+++ b/core/dispatch.c
@@ -755,7 +755,8 @@ dispatch_enter_dynamorio(dcontext_t *dcontext)
             * thread's context is being reset proactively (due to some -reset_at_*
             * option) or before sending it native.
             */
-           wherewasi == DR_WHERE_DISPATCH);
+           ((dcontext->go_native || dcontext->last_exit == get_reset_linkstub()) &&
+            wherewasi == DR_WHERE_DISPATCH));
     dcontext->whereami = DR_WHERE_DISPATCH;
     ASSERT_LOCAL_HEAP_UNPROTECTED(dcontext);
     ASSERT(check_should_be_protected(DATASEC_RARELY_PROT));
--- a/core/synch.c
+++ b/core/synch.c
@@ -1820,10 +1820,11 @@ translate_from_synchall_to_dispatch(thread_record_t *tr, thread_synch_state_t sy
          * translate_mcontext.
          */
         IF_AARCHXX({
-            // XXX i#4495: Consider saving stolen reg's translated application value.
+            /* XXX i#4495: Save stolen reg's translated application value. */
             set_stolen_reg_val(mc, (reg_t)os_get_dr_tls_base(dcontext));
-            // XXX: This path is tested by linux.thread-reset and linux.clone-reset.
-            // We just haven't run those on ARM yet.
+            /* XXX: This path is tested by linux.thread-reset and linux.clone-reset.
+             * We just haven't run those on ARM yet.
+             */
             IF_ARM(ASSERT_NOT_TESTED());
         });
 #ifdef WINDOWS
--- a/core/synch.c
+++ b/core/synch.c
@@ -1815,9 +1815,17 @@ translate_from_synchall_to_dispatch(thread_record_t *tr, thread_synch_state_t sy
          */
         mc->pc = (app_pc)get_reset_exit_stub(dcontext);
         LOG(GLOBAL, LOG_CACHE, 2, "\tsent to reset exit stub " PFX "\n", mc->pc);
-        /* TODO i#4497: Replace with the official fix from PR#4498. */
-        IF_AARCHXX(set_stolen_reg_val(mc, (reg_t)os_get_dr_tls_base(dcontext)));
-        IF_AARCHXX(ASSERT_NOT_TESTED()); /* PR#4498 will improve test coverage. */
+        /* The reset exit stub expects the stolen reg to contain the TLS base address.
+         * But the stolen reg was restored to the application value during
+         * translate_mcontext.
+         */
+        IF_AARCHXX({
+            // XXX i#4495: Consider saving stolen reg's translated application value.
+            set_stolen_reg_val(mc, (reg_t)os_get_dr_tls_base(dcontext));
+            // XXX: This path is tested by linux.thread-reset and linux.clone-reset.
+            // We just haven't run those on ARM yet.
+            IF_ARM(ASSERT_NOT_TESTED());
+        });
 #ifdef WINDOWS
         /* i#25: we could have interrupted thread in DR, where has priv fls data
          * in TEB, and fcache_return blindly copies into app fls: so swap to app
--- a/core/unix/signal.c
+++ b/core/unix/signal.c
@@ -2681,7 +2681,10 @@ thread_set_self_context(void *cxt)
 #    endif /* X86 */
     frame.uc.uc_mcontext = *sc;
 #endif
+    IF_ARM(ASSERT_NOT_TESTED());
+#if defined(LINUX) && defined(X86)
     save_fpstate(dcontext, &frame);
+#endif
     /* The kernel calls do_sigaltstack on sys_rt_sigreturn primarily to ensure
      * the frame is ok, but the side effect is we can mess up our own altstack
      * settings if we're not careful.  Having invalid ss_size looks good for
@@ -2695,7 +2698,8 @@ thread_set_self_context(void *cxt)
     DOLOG(LOG_ASYNCH, 3,
           { dump_sigcontext(dcontext, get_sigcontext_from_rt_frame(&frame)); });
     /* set up xsp to point at &frame + sizeof(char*) */
-    xsp_for_sigreturn = ((app_pc)&frame) + sizeof(char *);
+    /* For x86 only, we need to skip pretcode while setting xsp. */
+    xsp_for_sigreturn = ((app_pc)&frame) IF_X86(+ sizeof(char *));
 #ifdef X86
     asm("mov  %0, %%" ASM_XSP : : "m"(xsp_for_sigreturn));
 #    ifdef MACOS
@@ -2710,7 +2714,8 @@ thread_set_self_context(void *cxt)
     asm("jmp  *%" ASM_XCX);
 #    endif /* MACOS/LINUX */
 #elif defined(AARCH64)
-    ASSERT_NOT_IMPLEMENTED(false); /* FIXME i#1569 */
+    asm("mov  " ASM_XSP ", %0" : : "r"(xsp_for_sigreturn));
+    asm("b    dynamorio_sigreturn");
 #elif defined(ARM)
     asm("ldr  " ASM_XSP ", %0" : : "m"(xsp_for_sigreturn));
     asm("b    dynamorio_sigreturn");
@@ -3423,6 +3428,7 @@ transfer_from_sig_handler_to_fcache_return(dcontext_t *dcontext, kernel_ucontext
      * from DR's handler.
      */
     ASSERT(get_sigcxt_stolen_reg(sc) != (reg_t)*get_dr_tls_base_addr());
+    // XXX i#4495: Consider saving stolen reg's translated application value.
     set_sigcxt_stolen_reg(sc, (reg_t)*get_dr_tls_base_addr());
 #    ifndef AARCH64
     /* We're going to our fcache_return gencode which uses DEFAULT_ISA_MODE */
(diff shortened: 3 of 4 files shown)


core/unix/signal.c
```diff
@@ -2676,12 +2675,16 @@ thread_set_self_context(void *cxt)
...
 #ifdef LINUX
 #    ifdef X86
+    dcontext_t *dcontext = get_thread_private_dcontext();
     byte *xstate = get_and_initialize_xstate_buffer(dcontext);
     frame.uc.uc_mcontext.fpstate = &((kernel_xstate_t *)xstate)->fpstate;
 #    endif /* X86 */
     frame.uc.uc_mcontext = *sc;
 #endif
+    IF_ARM(ASSERT_NOT_TESTED());
+#if defined(LINUX) && defined(X86)
```
abhinav92003, line 2685: @AssadHashmi Do we need any special handling for fpstate on AArch64/ARM?

Copying over the [conversation](https://github.com/DynamoRIO/dynamorio/issues/4496#issuecomment-718945863) from #4496. 

> > abhinav92003: @derekbruening About the `ASSERT_NOT_IMPLEMENTED` in `save_fpstate` for AArch64: Do we need special handling for fpstate on AArch64/ARM?
> 
> derekbruening: I don't think so: dr_mcontext.simd[] should hold all the direct values (i.e., it's not like x86 where the x87 state is not in dr_mcontext_t _at all_). Maybe there are status registers or sthg though? I remember this coming up recently in some issue or PR and Assad was not sure -- I can't find the issue though: recent ones are #4467, #4488.
> 
> @AssadHashmi -- if you could comment.

core/dispatch.c
```diff
@@ -752,9 +752,10 @@ dispatch_enter_dynamorio(dcontext_t *dcontext)
            wherewasi == DR_WHERE_APP ||
            /* If the thread was waiting at check_wait_at_safe_point when getting
             * suspended, we were in dispatch (ref i#3427). We will be here after the
-            * thread's context is being reset before sending it native.
+            * thread's context is being reset proactively (due to some -reset_at_*
+            * option) or before sending it native.
             */
-           (dcontext->go_native && wherewasi == DR_WHERE_DISPATCH));
+           wherewasi == DR_WHERE_DISPATCH);
```
derekbruening, line 758: Could this instead keep the `go_native` and add to it "or the last exit was the special reset exit"?

core/unix/signal.c
```diff
@@ -2676,12 +2675,16 @@ thread_set_self_context(void *cxt)
     memset(&frame, 0, sizeof(frame));
 #ifdef LINUX
 #    ifdef X86
+    dcontext_t *dcontext = get_thread_private_dcontext();
```
derekbruening, line 2678: It seems fragile to put this decl in a different ifdef set from the save_fpstate usage below.

core/unix/signal.c
derekbruening, line 2685: The prior code called this for non-Linux too (e.g., MacOS) so this does not seem like a safe change.

core/unix/signal.c
```diff
@@ -2709,12 +2715,13 @@ thread_set_self_context(void *cxt)
     asm("mov  %0, %%" ASM_XCX : : "m"(asm_jmp_tgt));
     asm("jmp  *%" ASM_XCX);
 #    endif /* MACOS/LINUX */
-#elif defined(AARCH64)
-    ASSERT_NOT_IMPLEMENTED(false); /* FIXME i#1569 */
+#elif defined(AARCH64) && !defined(DR_HOST_NOT_TARGET)
```
derekbruening, line 2718: This additional host check should not just be added to one of these arches: the behavior should be the same for other host-target combinations.  Instead, in general either A) DR_HOST_<arch> should be used as the condition for each of asm block or B) We don't support host!=target here since we only support non-code-execution use (decoding; standalone lib) and there should be a check for that with ASSERT_NOT_REACHED just like elsewhere in this file (e.g., notify_and_jmp_without_stack()).

core/synch.c
```diff
@@ -1815,9 +1815,17 @@ translate_from_synchall_to_dispatch(thread_record_t *tr, thread_synch_state_t sy
...
         LOG(GLOBAL, LOG_CACHE, 2, "\tsent to reset exit stub " PFX "\n", mc->pc);
-        /* TODO i#4497: Replace with the official fix from PR#4498. */
-        IF_AARCHXX(set_stolen_reg_val(mc, (reg_t)os_get_dr_tls_base(dcontext)));
-        IF_AARCHXX(ASSERT_NOT_TESTED()); /* PR#4498 will improve test coverage. */
+        /* The reset exit stub expects the stolen reg to contain the TLS base address.
+         * But the stolen reg was restored to the application value during
+         * translate_mcontext.
+         */
+        IF_AARCHXX({
+            // XXX i#4495: Consider saving stolen reg's translated application value.
```
derekbruening, line 1823: style nit: It does say at https://github.com/DynamoRIO/dynamorio/wiki/Code-Style#commenting-conventions to prefer `/* */` to `//` in C code though we could vote to change that.

I would use something stronger than "Consider" since this seems to be a correctness bug.

Ditto below.

core/unix/signal.c
abhinav92003: There's an `ASSERT_NOT_REACHED` after this ladder, but +1 on making it clearer like `notify_and_jmp_without_stack`.

core/unix/signal.c
abhinav92003: Non-Linux seems to skip `save_fpstate` at other spots too.
https://github.com/DynamoRIO/dynamorio/blob/420adb01d466e134037b1db6ed7023749c3f8034/core/unix/signal.c#L5575
Do we need a fix there as well?

derekbruening: Huh, that's weird.  I would expect Mac on x86 to need the same thing, yes, for the x87 state which is not in the mcontext and so can't be updated from there.  Wondering if changing that will hit the assert on the mac tests though.

core/dispatch.c
abhinav92003: Done.

core/unix/signal.c

core/synch.c
abhinav92003: Oh sorry about that. Done.

core/unix/signal.c
abhinav92003: Separated the declaration into its own ifdef.

core/unix/signal.c
abhinav92003: macOS doesn't implement `save_fpstate`
https://github.com/DynamoRIO/dynamorio/blob/a5ab87c9c24f176a474dc1b7d3cb4cc4fe4d8ed4/core/unix/signal_macos.c#L145

I modified the preprocessor guard to allow `save_fpstate` for macOS builds, but macOS still passes on Travis: https://travis-ci.com/github/DynamoRIO/dynamorio/jobs/425890837.

abhinav92003: In a bunch of spots in `signal.c`, I see that fpstate handling is done only for `X86 && LINUX` (not including macOS). Do we need to re-evaluate this, perhaps as part of a separate issue?

derekbruening: The signal frame extended state layout is different: maybe you're seeing code specific to how Linux lays it out.  The general step of updating the fpstate on a delayed signal to have the new delivery spot's state is what is common and that should only be in a couple of places calling save_fpstate.   I'm guessing you're looking at different code that is not related to that step.

core/unix/signal.c
```diff
@@ -2692,11 +2697,16 @@ thread_set_self_context(void *cxt)
                         sizeof(frame.uc.uc_sigmask));
     LOG(THREAD_GET, LOG_ASYNCH, 2, "thread_set_self_context: pc=" PFX "\n", sc->SC_XIP);
     LOG(THREAD_GET, LOG_ASYNCH, 3, "full sigcontext\n");
-    DOLOG(LOG_ASYNCH, 3,
-          { dump_sigcontext(dcontext, get_sigcontext_from_rt_frame(&frame)); });
+    DOLOG(LOG_ASYNCH, 3, {
+        dcontext_t *dcontext = get_thread_private_dcontext();
```
derekbruening, line 2701: This is shadowing the var of the same name in the outer scope: please fix.  Surprising the compiler doesn't warn about it (MSVC complains about such things).  Maybe best to just put it back how it was, with dcontext in all builds.  Your goal was to remove the call to gtpd for arm?  Is it not removed by the compiler in a release build?

derekbruening (approved):

core/unix/signal.c
abhinav92003: All usages of `dcontext` in this function are conditional (on either some environment or build type) now, unlike before this PR. So it's slightly tricky to avoid unused var errors.

For this case, it seems that `DOLOG` is defined only under a complex combination (`INTERNAL && DEBUG && !STANDALONE_DECODER`). So I figured it's best to let there be a separate declaration here; I'll just change the identifier so that it doesn't shadow the one in outer scope.

core/unix/signal.c
abhinav92003: Oh okay. 

The other `save_fpstate` instance[1] in `signal.c` is invoked during `receive_pending_signal`, so we do expect that to be done for macOS too right? 

About this instance of `save_fpstate` not being hit in macOS tests: actually the reset tests don't seem to be enabled for macOS. Do we want them to be?

[1]: https://github.com/DynamoRIO/dynamorio/blob/420adb01d466e134037b1db6ed7023749c3f8034/core/unix/signal.c#L5575

core/unix/signal.c

core/unix/signal.c
AssadHashmi: Apologies for the late response @abhinav92003 and @derekbruening, this slipped under the mat.

> Do we need special handling for fpstate on AArch64/ARM?

Not in this case AFAICT.

For future reference, the full implementation of `save_fpstate` for AArch64 is tracked by https://github.com/DynamoRIO/dynamorio/issues/2629

abhinav92003: Discussed offline. 

Added an XXX comment for now for the other `save_fpstate`. Will investigate in a separate PR if we need any changes there.

AssadHashmi (approved):

core/unix/signal.c
abhinav92003: Thanks Assad and Derek. (Some more discussion on #2629).

Issue #4508, opened 2020-11-01
Dynamorio not loading stolen register value into mcontext on AArch64


gregcawthorne:
On AArch64 the value of the stolen register isn't loaded into the mcontext object for later instrumentation, due to an ifdef only allowing this functionality on ARM (non AArch64), in instrument.c, dr_get_mcontext_priv().

This causes errors when addresses need to be calculated based on the value of contents the stolen register.

Issue is present on the current latest build: cronbuild-8.0.18565

Fix is to change this one line ifdef to #ifdef AARCHXX
https://github.com/DynamoRIO/dynamorio/blob/420adb01d466e134037b1db6ed7023749c3f8034/core/lib/instrument.c#L6802

Will submit the fix very soon.

derekbruening:
My test for #4495 is hitting this as it wants to test the stolen register in a clean call.

Issue #4495, opened 2020-10-27
Translated stolen register app value lost on signal/synchall control transfer


derekbruening:
When translating a thread context for synchall or on a signal, we invoke the client fault handler and arrive at the app's proper value for the stolen register in our local mcontext.  We then use the same mcontext to get back into DR through an exit stub, but for that transition we need to put DR's TLS pointer back into the stolen register.  That clobbers the translated value.  If all client handlers did not touch the value, no harm is done since the TLS slot value is taken as the app value; but if a client were to need to restore a value there, we would get the wrong value in the end.  This should be rare, since drreg avoids using the stolen register.

derekbruening:
Taking this on as it is looking like it may be related to #4460 in that both seem to involve the TLS stolen reg slot across signals and translations.

derekbruening:
For the signal path, it seems to only affect the handler itself: a longjmp out of there should restore properly to the setjmp value, and a re-execute will work since the sigcontext has the right value.  It seems less likely that the handler will assume the value from the faulting context, but it is possible.

For the synchall path, this is a more serious issue and less of a corner case.

Issue #4370, opened 2020-07-16
CMake 3.18.0 build of master is broken (Ubuntu 16.04/18.04)


SweetVishnya:
commit 12496fe8bb2636fb8df66bca211469b967795414

```
sudo docker run --rm -it ubuntu:16.04 /bin/bash
apt update && apt upgrade -y && apt install -y git build-essential wget
wget https://github.com/Kitware/CMake/releases/download/v3.18.0/cmake-3.18.0-Linux-x86_64.sh
sh cmake-3.18.0-Linux-x86_64.sh --prefix=/ --exclude-subdir --skip-license
git clone https://github.com/DynamoRIO/dynamorio.git
cd dynamorio
mkdir build
cd build
cmake ..
make
```

I get the following compilation error.

```
Scanning dependencies of target dynamorio_static
[  4%] Building C object core/CMakeFiles/dynamorio_static.dir/options.c.o
[  4%] Building C object core/CMakeFiles/dynamorio_static.dir/dynamo.c.o
[  4%] Building C object core/CMakeFiles/dynamorio_static.dir/fragment.c.o
[  4%] Building C object core/CMakeFiles/dynamorio_static.dir/fcache.c.o
[  5%] Building C object core/CMakeFiles/dynamorio_static.dir/link.c.o
[  5%] Building C object core/CMakeFiles/dynamorio_static.dir/dispatch.c.o
[  5%] Building C object core/CMakeFiles/dynamorio_static.dir/emit.c.o
[  5%] Building C object core/CMakeFiles/dynamorio_static.dir/utils.c.o
[  5%] Building C object core/CMakeFiles/dynamorio_static.dir/config.c.o
[  6%] Building C object core/CMakeFiles/dynamorio_static.dir/stats.c.o
[  6%] Building C object core/CMakeFiles/dynamorio_static.dir/heap.c.o
[  6%] Building C object core/CMakeFiles/dynamorio_static.dir/monitor.c.o
[  6%] Building C object core/CMakeFiles/dynamorio_static.dir/perfctr.c.o
[  7%] Building C object core/CMakeFiles/dynamorio_static.dir/vmareas.c.o
[  7%] Building C object core/CMakeFiles/dynamorio_static.dir/rct.c.o
[  7%] Building C object core/CMakeFiles/dynamorio_static.dir/hotpatch.c.o
[  7%] Building C object core/CMakeFiles/dynamorio_static.dir/hashtable.c.o
[...]

Issue #2433, opened 2017-05-16
opnd_compute_address() is not fully implemented on ARM or AArch64


egrimley:
On ARM `opnd_compute_address_priv()` does not take account of `DR_OPND_NEGATED`, and on AArch64 the extension and shift are ignored. The API function `opnd_compute_address()` is not tested.

derekbruening:
Dr. Memory uses opnd_compute_address() extensively.  However there is no test inside DR's own test suite.

AssadHashmi:
#4529

PR #4513, opened 2020-11-02
i#4508: Stolen register not loaded in dr_get_mcontext on AArch64

gregcawthorne:
i#4508 Dynamorio not loading stolen register value into mcontext on AArch64.

On AArch64 the value of the stolen register isn't loaded into the mcontext object for later instrumentation.

This is due to an ifdef only allowing this functionality on ARM (non AArch64), in core/lib/instrument.c, dr_get_mcontext_priv().

This causes errors when addresses need to be calculated based on the value of contents the stolen register.

Fix is a simple update of an #ifdef from #ifdef ARM to ifdef AARCHXX


AssadHashmi:
run arm tests

AssadHashmi:
Added @gregcawthorne to Jenkins ACL for AArch64 pre-commit tests.

core/lib/instrument.c
```diff
@@ -6799,7 +6799,7 @@ dr_get_mcontext_priv(dcontext_t *dcontext, dr_mcontext_t *dmc, priv_mcontext_t *
     else if (TEST(DR_MC_CONTROL, dmc->flags))
```
derekbruening, line 6799: (Putting commit title/description comment here to get an action item that can be marked Done and a separate thread.)

style nit 1: Title should have a colon after the issue number.

style nit 2: Description should have a final line that is either `Fixes #4508` or `Issue: #4508` (the former here).

style nit 3: suggestion: include the key `dr_get_mcontext` in the title, and remove the sort of implied "Dynamorio"  (should be "DynamoRIO", or "DR" for short) (it will also then fit in the github issue title); maybe sthg like `i#4508: Stolen register not loaded in dr_get_mcontext on AArch64`.

derekbruening (commented):
Thank you for submitting the patch.  The code looks fine; just some nits/suggestions on the commit description.

derekbruening:
Looks like the `client.drwrap-test-callconv` and `client.drwrap-test` tests fail on A64 with this patch with asserts which is surprising: can you reproduce that locally?

gregcawthorne:
Hi.

Yes I am having the same problem.

The error is the same for both of them.

128: <Application /home/***/dynamorio/build/suite/tests/bin/client.drwrap-test-callconv (49428) DynamoRIO usage error : meta-instr faulted?  must set translation field and handle fault!>
128: <Usage error: meta-instr faulted?  must set translation field and handle fault! (/home/***/dynamorio/core/translate.c, line 909)

I will have to dig into this, this week.

Had been using in my drmemory work, but hadn't run the dynamoRIO tests.

derekbruening:
PR #4491 very recently fixed an instance of that assert and maybe is not in your branch so it's worth an update to check, though I don't think the causes of the assert that it fixed (can be caused by multiple underlying issues) would show up on the drwrap tests.

core/lib/instrument.c
```diff
@@ -6799,7 +6799,7 @@ dr_get_mcontext_priv(dcontext_t *dcontext, dr_mcontext_t *dmc, priv_mcontext_t *
...
         dmc->xsp = get_mcontext(dcontext)->xsp;
 
-#ifdef ARM
+#ifdef AARCHXX
     if (TEST(DR_MC_INTEGER, dmc->flags)) {
```
derekbruening, line 6803: I happened to need this locally and it turns out there is a bug here: `dmc` can be NULL, so this should be:
```
    if (mc != NULL || TEST(DR_MC_INTEGER, dmc->flags)) {
```
I hit a crash with just this patch as-is due to this.

core/lib/instrument.c
```diff
@@ -6799,7 +6799,7 @@ dr_get_mcontext_priv(dcontext_t *dcontext, dr_mcontext_t *dmc, priv_mcontext_t *
...
         /* get the stolen register's app value */
         if (mc != NULL) {
```
derekbruening, line 6805: Also, there seem to be 2 instances of the exact same `s/ARM/AARCHXX/` issue in `dr_set_mcontext` which could easily be part of this same PR.

derekbruening, line 6805: It would also be nice to have a test that fails without each of the get and set fixes here (not to pile too much work on this...)

core/lib/instrument.c
gregcawthorne: Ah thanks,
I presume you mean:
`if (dmc != NULL || TEST(DR_MC_INTEGER, dmc->flags)) {`
?

core/lib/instrument.c
gregcawthorne: I should be able to create a test that mimic the original bug I found. Hopefully that will catch the absence of the other fix too, but will have to look into it.

core/lib/instrument.c
derekbruening: No, if `mc` is non-NULL, then there are no flags and everything is filled in and `dmc` is ignored and should be NULL.  Only if `mc` is NULL is `dmc` guaranteed to be non-NULL.

gregcawthorne: Ah okay that makes sense thanks!

derekbruening:
Adding link to issue so it shows up there: #4508

derekbruening:
My test for #4495 is hitting this as it wants to test the stolen register in a clean call so I'm slightly blocked on it.  @gregcawthorne -- when I put in the dr_set_mcontext fix too, the drwrap-test failures go away.

gregcawthorne:
Ah okay that's good to know about the test failures going away, thanks.

Will confirm and push an updated version.

Will also look into making a test and see about getting that in tomorrow.

core/lib/instrument.c
gregcawthorne: For the description change, would you consider me force pushing a single commit with all the changes bad form. Or is it best to always do it incrementally and squash the commits before the merge.

derekbruening: Never force push.  The merge is always a squash with an edited description: edit it there.  Please remove all the star lines and create a single message at that time as instructed in the wiki.

gregcawthorne:
Hi.

Update here.

Been putting together the test for this. At first I tried to make an api test, but now have tried to make a client test.

I have made one which calls dr_get_mcontext, dr_get_mcontext and opnd_compute_address, which seemed to pass with the fixes applied, but then I removed the fixes and it still passed.

I looked into it and its because these ifs return true from my test.

[https://github.com/DynamoRIO/dynamorio/blob/8f142e8909736b154d0ed09809c602b4497dac4a/core/lib/instrument.c#L6774
](url)
It's the dcontext->client_data->mcontext_in_dcontext part which return true here.


[https://github.com/DynamoRIO/dynamorio/blob/8f142e8909736b154d0ed09809c602b4497dac4a/core/lib/instrument.c#L6851](url)

This possibly explains how it hasn't been hit in testing previously.

 I am currently looking into how to make them return false for the sake of the test. One handicap seems to be how I can't seem to call into a lot of the nitty gritty methods and definitions directly from an external client, but that's by design probably. It should be possible to get this test to fail of course as DrMemory was and that is a client.

derekbruening:
> I have made one which calls dr_get_mcontext, dr_get_mcontext and opnd_compute_address, which seemed to pass with the fixes applied, but then I removed the fixes and it still passed.
> 
> I looked into it and its because these ifs return true from my test.
> 
> [https://github.com/DynamoRIO/dynamorio/blob/8f142e8909736b154d0ed09809c602b4497dac4a/core/lib/instrument.c#L6774](url)

You mean your test calls `dr_get_mcontext` from a syscall event?  If you call it from a clean call from the cache you should hit the bug.  I hit it in every clean call from the cache I've added recently to tests: e.g., suite/tests/client-interface/stolen-reg.dll.c.

derekbruening:
You could consider adding checks to suite/tests/client-interface/stolen-reg.{c,dll.c} rather than making a separate test if desired: that test is meant to detect stolen register bugs, after all, which this is an instance of, and I already hit it there when I checked the value in the clean callee.

gregcawthorne:
Ah I didn't see those tests. They look comprehensive and not sure if my test would add much.

The only thing I was thinking is that isn't there an option to change the stolen reg from x28 to something else?

Perhaps the tests should find the stolen reg and then check based on that?

derekbruening:
> Ah I didn't see those tests. They look comprehensive and not sure if my test would add much.

But stolen-reg.* is not comprehensive enough: it is not currently detecting this bug.

> The only thing I was thinking is that isn't there an option to change the stolen reg from x28 to something else?
> 
> Perhaps the tests should find the stolen reg and then check based on that?

Yes, that would be good, or at least a check for dr_get_stolen_reg() equaling x28 in the client main so we'll catch the test silently no longer testing what it should if we change the default stolen reg.

> But that should probably be a different commit.

If you could add it as part of adding the clean call `dr_get_mcontext` check that would be great.

derekbruening:
> Ah I didn't see those tests. 

I just added the stolen-reg test last week which is why you didn't see it :)

gregcawthorne:
Okay thanks. I will crack on!

suite/tests/client-interface/set-stolen-reg-mcontext.c
```diff
@@ -0,0 +1,40 @@
...
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+#include <stdio.h>
+
+int main(void){
+    // just an instruction to detect in *.dll.c
+    __asm__ __volatile__("mov x27, #1" : : : "x27");
```
gregcawthorne, line 38: I wasn't sure about checking for he stolen reg value in the app, as this fix is specifically about how the mcontext is maintained. For me I don't think the issue wast the app having a corrupt x28 register in its execution, but rather the client not knowing what the true value was.

suite/tests/client-interface/set-stolen-reg-mcontext.dll.c
```diff
@@ -0,0 +1,122 @@
...
+
+    for (instr = instrlist_first(bb); instr != NULL; instr = next_instr) {
+        next_instr = instr_get_next(instr);
+
+        reg_t imm1 = 0;
+        if (instr_is_mov_constant(instr, &imm1) &&
+                opnd_is_reg(instr_get_dst(instr, 0)) &&
+                opnd_get_reg(instr_get_dst(instr, 0)) == DR_REG_R27 && imm1 == 1) {
+
+            dr_insert_clean_call(
```
gregcawthorne, line 105: Currently the fails with the fix in instrument.c, dr_set_mcontext reverted, due to a translation related error in recreate_app_state_from_ilist. However it passes with he fix and the recreate_app_state_from_ilist code path isn't hit with he fix. Currently trying to work out why this is the case.

suite/tests/client-interface/set-stolen-reg-mcontext.c
derekbruening: The suggestion to possibly have the app check the value is for dr_set_mcontext where the client is changing the app value.  It is admittedly not a very normal thing for a real tool to do, but a cooperative app in a test is ok.  Another suggestion is to do what the existing cleancall tests do: consecutive clean calls where one does a dr_set_mcontext and the next does a dr_get_mcontext and looks for the new value.

gregcawthorne: I have implemented the latter here. Can do a cooperative app-client test instead if that's best.

abhinav92003:
@gregcawthorne This issue (#4508) is also causing `code_api|tool.drcachesim.delay-simple` test failures on AArch64 (#4571). We had to disable that test recently to unblock various other contributors. It'd be best to get the fix submitted soon, so that we can re-enable the test (not good to keep it disabled for long, as other bugs may creep in).

gregcawthorne:
Have most of the tests passing now.

I seem to be having the same x86 test issue as #2433 is currently having, but only for the osx test.

Will look into the AArch64 Precommit failure tomorrow.

I tried to 'resolve conflicts' with he web UI and it said 'resolve commit and merge', so I hung fire as I wasn't sure if that was a merge into my dev branch or not.

gregcawthorne:
I just resolved the merge conflicts, taking in the new changes from master, and all tests groups are failing now. I am sure were the cause of this is coming from currently.

derekbruening:
https://github.com/DynamoRIO/dynamorio/issues/4370#issuecomment-738172341

abhinav92003:
> I just resolved the merge conflicts, taking in the new changes from master, and all tests groups are failing now. I am sure were the cause of this is coming from currently.

Those failures were due to #4370, which is fixed now. Please update this branch to trigger the CI again.

gregcawthorne:
run tests

derekbruening:
I thought the plan was to just add a couple of checks to the existing stolen-reg test?

gregcawthorne:
@derekbruening Will merge it into stolen reg then. Wasn't sure which way for definite, but was focusing on getting it fully passing first anyway.

derekbruening:
If possible, please test offline to avoid so many pushes, to avoid stressing the CI resources: alternatively, please cancel the old ones.  These are all stacked up Appveyor which is globally serialized and is now tied up on just old commits for this PR for the next 3 to 4 hours: https://ci.appveyor.com/project/DynamoRIO/dynamorio/history

Please see https://github.com/DynamoRIO/dynamorio/wiki/Test-Suite#ci-resources-and-appveyor-global-serialization

gregcawthorne:
Hi okay will do.

They were part formatting errors which I can test locally, part aarch32 errors which I am not currently set up to test as easily.

I naively hoped a new build would auto cancel the previous builds work.

Will try and cancel them now.

Thanks for the heads up

Greg

derekbruening:
> @derekbruening Will merge it into stolen reg then. Wasn't sure which way for definite, but was focusing on getting it fully passing first anyway.

Sure.  Note that I did not look at the patch yet -- if the test is large and you prefer it separate that makes sense; I just remembered you saying it would fit into the existing test, and me also thinking it would be a natural fit there, but then seeing several commits here getting a separate test to build/work that might be avoided.

gregcawthorne:
I have merged it into the original stolen-reg.dll.c test now anyway :)

derekbruening:
> Hi okay will do. They were part formatting errors which I can test locally, part aarch32 errors which I am not currently set up to test as easily. I naively hoped a new build would auto cancel the previous builds work. Will try and cancel them now. Thanks for the heads up.

Travis would auto-cancel if a new commit was within a few seconds of a prior one.  Not sure about Github Actions -- @abhinav92003 do you recall seeing anything in the docs or notice anything there: it would be nice to auto-cancel old runs there.  I would say if there's a new commit at all before the old run is done, cancel it -- but not sure Github supports that.

derekbruening:
Looks like it might be possible to use someone else's support for auto-cancel out of the box: https://github.community/t/is-it-possible-to-cancel-previous-jobs-automatically-when-the-pull-request-is-updated/16779

I'll file a feature request issue on it, if someone gets spare time (not sure how likely that is...)

derekbruening:
> Looks like it might be possible to use someone else's support for auto-cancel out of the box: https://github.community/t/is-it-possible-to-cancel-previous-jobs-automatically-when-the-pull-request-is-updated/16779
> 
> I'll file a feature request issue on it, if someone gets spare time (not sure how likely that is...)

#4585

derekbruening:
> add appveyor RDP login

Note that I canceled all the appveyor builds: just making sure you're not investigating a cancel or a flaky test: which test are you looking at on appveyor?

gregcawthorne:
I just wanted to re-trigger the tests to get a 100% pass and I thought might as well add those lines you linked to in case it was useful in future.

derekbruening:
Is this ready to review?  Github does not have a very good interface for "whose turn it is" so it is hard to tell: normally a "PTAL" comment helps when the code writer is finished, along with replying to all of the reviewer comments with "Done" or sthg.

gregcawthorne:
Yes I believe could do with a review.

I still need to make the changes to the commit message you said, but that is to be done in the squash and merge I believe so didn't want to put 'Done'.

suite/tests/client-interface/stolen-reg.dll.c
```diff
@@ -67,6 +67,59 @@ do_flush(app_pc next_pc)
...
+// Storing the original stolen reg before mconext set.
+static ptr_int_t orig_value = 0;
+
+// Random value to detect after mcontext set.
+static ptr_int_t test_value = 7;
+
+static void
+load_stolen_reg_to_mcontext()
+{
+    fprintf(stderr, "load_stolen_reg_to_mcontext entered\n");
```
derekbruening, line 79: Use `dr_fprintf(STDERR` in clients.

derekbruening, line 79: suggest: Avoid duplication to avoid stale values when one changes: use `%s` and `__FUNCTION__`.

suite/tests/client-interface/stolen-reg.dll.c
```diff
@@ -67,6 +67,59 @@ do_flush(app_pc next_pc)
...
+    dr_mcontext_t mc;
+    mc.size = sizeof(mc);
+    mc.flags = DR_MC_ALL;
+
+    fprintf(stderr, "fetching TLS\n");
+
+    dr_get_mcontext(drcontext, &mc);
+
+    fprintf(stderr, "mc->stolen_reg after = " IF_ARM_ELSE("%d", "%ld") "\n",
+            mc.IF_ARM_ELSE(r10, r28));
```
derekbruening, line 93: I think it is better to have a DR_ASSERT or otherwise explicit check here that r10==test_value.  That makes the code easier to read (it's clear what we expect) and is generally better than an implied, hard-to-figure-out assertion through the "golden output" check of the output file.

suite/tests/client-interface/stolen-reg.dll.c
```diff
@@ -67,6 +67,59 @@ do_flush(app_pc next_pc)
...
+
+    mc.IF_ARM_ELSE(r10, r28) = orig_value;
+
+    dr_set_mcontext(drcontext, &mc);
+}
+
+static void
+save_stolen_reg_to_tls()
```
derekbruening, line 101: nit suggest: The client shouldn't really be aware of where DR is putting the stolen reg value, whether it's in TLS or somewhere in the private mcontext.  This is a black box test using public interfaces: so maybe s/save_stolen_reg_to_tls/change_stolen_reg_value/ and the other "read_and_restore_stolen_reg_value" or something.

derekbruening (commented):
The basic content looks good but I have a number of suggestions that I think will improve the test.

derekbruening:
We are again hitting this -- let's get it in.  I think I may just go make these suggested tweaks to get it merged.

.appveyor.yml
```diff
@@ -130,3 +130,9 @@ deploy:
   on:
     branch: master
     appveyor_repo_tag: true
+
+environment:
+  APPVEYOR_RDP_PASSWORD: MyPasswordToDiagnoseBug12345
```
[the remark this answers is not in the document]
derekbruening: Reverted.

suite/tests/client-interface/stolen-reg.dll.c
```diff
@@ -133,6 +199,20 @@ DR_EXPORT
 void
 dr_init(client_id_t id)
 {
+    // Stop test failing silently if we ever change the sotlen reg value.
```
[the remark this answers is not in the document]
derekbruening: Done.

suite/tests/client-interface/stolen-reg.dll.c
```diff
@@ -133,6 +199,20 @@ DR_EXPORT
...
+#elif defined(ARM)
+    if (dr_get_stolen_reg() != DR_REG_R10) {
+#else
+#    error Unsupported arch
+#endif
+        printf("ERROR: stolen reg value has changed, this test needs to be updated");
+        DR_ASSERT(false);
+#ifdef AARCHXX
+    }
+#endif
```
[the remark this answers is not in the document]
derekbruening: Done.

suite/tests/client-interface/stolen-reg.dll.c
```diff
@@ -120,11 +173,24 @@ bb_event(void *drcontext, void *tag, instrlist_t *bb, bool for_trace, bool trans
...
             next_instr != NULL && instr_is_nop(next_instr) && next_next_instr != NULL &&
             instr_is_nop(next_next_instr)) {
+
             dr_insert_clean_call(
                 drcontext, bb, next_next_instr, (void *)do_flush, false /*fpstate */, 1,
                 OPND_CREATE_INTPTR((ptr_uint_t)instr_get_app_pc(next_next_instr)));
             break;
         }
+
+        // Look for mov <stolen-reg>, #1.
```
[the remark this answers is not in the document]
derekbruening: Switched to a rarer sentinel value.

suite/tests/client-interface/stolen-reg.dll.c
```diff
@@ -133,6 +199,20 @@ DR_EXPORT
...
 {
+    // Stop test failing silently if we ever change the sotlen reg value.
+#ifdef AARCH64
+    if (dr_get_stolen_reg() != DR_REG_R28) {
+#elif defined(ARM)
+    if (dr_get_stolen_reg() != DR_REG_R10) {
+#else
+#    error Unsupported arch
+#endif
+        printf("ERROR: stolen reg value has changed, this test needs to be updated");
```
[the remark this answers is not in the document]
derekbruening: Done.

suite/tests/client-interface/stolen-reg.dll.c
```diff
@@ -120,11 +173,24 @@ bb_event(void *drcontext, void *tag, instrlist_t *bb, bool for_trace, bool trans
             opnd_get_reg(instr_get_dst(instr, 0)) == dr_get_stolen_reg() &&
             next_instr != NULL && instr_is_nop(next_instr) && next_next_instr != NULL &&
             instr_is_nop(next_next_instr)) {
+
```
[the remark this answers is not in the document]
derekbruening: Removed.

suite/tests/client-interface/stolen-reg.dll.c
derekbruening: Done.

derekbruening: Done.

suite/tests/client-interface/stolen-reg.dll.c
derekbruening: Added the assert.  Just left the golden output though to expedite the changes.

suite/tests/client-interface/stolen-reg.dll.c
derekbruening: Done.

derekbruening:
Hmm, `0xdead` assembled in my local 32-bit arm cross-compile.

derekbruening:
As mentioned, I made the requested changes myself to expedite this as we have been waiting for it too long and it is blocking a64 development.  Merging.

derekbruening (approved):

PR #4521, opened 2020-11-05
i#4520: Restore ARM opnd_t layout

### LINKED ISSUE #4520: ASSERT in every app on ARM: opnd_t size changed
I assume this is a regression from PR #4467 3a408a034b841d036c53d972eff0c9cbe75d7cbe.  Lack of ARM CI tests is really an issue.
```
95: <Application /home/derek/dr/git/build_dbg_tests/suite/tests/bin/client.signal (16492).  Internal Error: DynamoRIO debug check failure: /home/derek/dr/git/src/core/arch/arch.c:668 sizeof(opnd_t) == EXPECTED_SIZEOF_OPND
```

derekbruening:
Avoids padding due to 8-byte-alignment of the new opnd_t.immed_double
field, to keep the layout how it was before PR #4467, avoiding an
init-time assert and problems matching the "black box" exported opnd_t
struct.

Fixes #4520

### DIFF
--- a/core/ir/opnd.h
+++ b/core/ir/opnd.h
@@ -1366,7 +1366,11 @@ struct _opnd_t {
          * Currently not included for Windows because sizeof(opnd_t) does not
          * equal EXPECTED_SIZEOF_OPND, triggering the ASSERT in d_r_arch_init().
          */
-        double immed_double; /* IMMED_DOUBLE_kind */
+        /* For 32-bit ARM we keep alignment at 4 to avoid changing the opnd_t shape.
+         * Marking this field as packed seems to do it and avoids other changes
+         * that might occur if packing the whole struct.
+         */
+        double immed_double IF_ARM(__attribute__((__packed__))); /* IMMED_DOUBLE_kind */
 #    endif
         /* PR 225937: today we provide no way of specifying a 16-bit immediate
          * (encoded as a data16 prefix, which also implies a 16-bit EIP,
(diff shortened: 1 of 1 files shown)


AssadHashmi (commented):
I guess we don't need to align explicitly as the compiler will take care of it.

AssadHashmi (approved):

derekbruening:
> I guess we don't need to align explicitly as the compiler will take care of it.

To match the binary interface's struct layout (and heap allocator's tuned size), we'd want to set alignment to 4 for the double.  An __aligned__ attribute will only set a minimum and so would be a nop since the other fields are already at 4.

One question is: are there ARM instructions involving doubles that require 8-byte alignment and so would fail with this packed 4-byte-aligned-only double field in opnd_t?  We could decide to break binary compability, have the opnd_t size increase to 16 bytes with a 4-byte pad to align the "value" union containing the double, change the black box struct layout, change the heap allocator expected size if there was a 12-byte size class, and bump the version to 8.90 (with the next 9.0).

derekbruening:
I don't see any use of floating-point immediates at all in our ARM decoder/encoder (is that expected?): so maybe it doesn't matter, assuming we're not missing something.  Maybe it's best to just put this simple change in now to fix the completely broken ARM, and we can either keep #4520 open or close that and have #4488 cover potential 32-bit ARM/x86 changes to the opnd_t layout.

AssadHashmi:
> I don't see any use of floating-point immediates at all in our ARM decoder/encoder (is that expected?)

Yes.

> we can either keep #4520 open or close that and have #4488 cover potential 32-bit ARM/x86 changes to the opnd_t layout.

Agree to close  #4520 and leave #4488 open for future work.

Issue #4522, opened 2020-11-05
ASSERT wrong isa mode on signal redirect and delivery on ARM: regression from kernel xfer event feature


derekbruening:
I'm porting the client.signal test to ARM and it hits this assert:
```
Redirected
Sending SIGUSR2
signal event 8 sig=12
kernel_xfer_event: type 9, sig 12
<Application /home/derek/dr/git/build_dbg_tests/suite/tests/bin/client.signal (1540).  Internal Error: DynamoRIO debug check failure: /home/derek/dr/git/src/core/dispatch.c:491 dr_get_isa_mode(dcontext) == FRAG_ISA_MODE(targetf->flags) IF_X64(|| (dr_get_isa_mode(dcontext) == DR_ISA_IA32 && !FRAG_IS_32(targetf->flags) && DYNAMO_OPTION(x86_to_x64)))
```

Looks like the kernel_xfer_event additional calls are re-calling
canonicalize_pc_target() which can't be called twice: the ISA mode is lost
then.

Issue #4474, opened 2020-10-06
Bridge testing gap on AArch64


derekbruening:
I'm filing this to help organize and drive work to finish big missing pieces on AArch64 in the test suite. There are about 140 tests on x86 that are not ported to AArch64, resulting in a big gap in coverage.  As we run larger A64 apps we're running into many bugs, at least some of which would have been caught and fixed much more easily on our test suite than on a  large, complex, custom app.

Xref the old master porting issue #1569 for AArch64 (and i#1551 for ARM). 

My main initial focus is drcachesim, and there we have two big drivers of missing tests:
- 11 tests missing because of toolchain linking issue #2007 
- 4 tests missing because of lack of annotation support #1672, but we don't really need annotations: we could tweak the test app

derekbruening:
Hmm actually #4516 is supposed fixed.

Issue #4518, opened 2020-11-04
CRASH on AArch64 in sigcontext_to_mcontext_simd() in client.signal test


derekbruening:
The client.signal test, after porting to a64, hits this crash:

```
Thread 1 "client.signal" received signal SIGSEGV, Segmentation fault.
0x000000007135abc4 in sigcontext_to_mcontext_simd (mc=0xfffd7402b380, sc_full=0xfffd7402b290) at /home/derek/dr/src/core/unix/signal_linux_aarch64.c:73
73	    ASSERT(fpc->head.magic == FPSIMD_MAGIC);
(gdb) bt
#0  0x000000007135abc4 in sigcontext_to_mcontext_simd (mc=0xfffd7402b380, sc_full=0xfffd7402b290)
    at /home/derek/dr/src/core/unix/signal_linux_aarch64.c:73
#1  0x00000000713329cc in sigcontext_to_mcontext (mc=0xfffd7402b380, sc_full=0xfffd7402b290, flags=DR_MC_ALL)
    at /home/derek/dr/src/core/unix/signal.c:2438
#2  0x000000007131a410 in os_context_to_mcontext (dmc=0xfffd7402b370, mc=0x0, osc=...) at /home/derek/dr/src/core/unix/os.c:3744
#3  0x00000000711f7e24 in dr_get_mcontext_priv (dcontext=0xfffd74005080, dmc=0xfffd7402b370, mc=0x0) at /home/derek/dr/src/core/lib/instrument.c:6745
#4  0x00000000711f8108 in dr_get_mcontext (drcontext=0xfffd74005080, dmc=0xfffd7402b370) at /home/derek/dr/src/core/lib/instrument.c:6827
#5  0x0000ffff73f904d0 in kernel_xfer_event (drcontext=0xfffd74005080, info=0xfffd7402ba68)
    at /home/derek/dr/src/suite/tests/client-interface/signal.dll.c:75
#6  0x00000000711ec608 in instrument_kernel_xfer (dcontext=0xfffd74005080, type=DR_XFER_SIGNAL_RETURN, source_os_cxt=..., source_dmc=0x0, 
    source_mc=0xfffd74005080, target_pc=0xffffb7c38c98 <chain_fragments_for_flush+2744> "\037\374?\261B", target_xsp=281474976707616, 
    target_os_cxt=..., target_mc=0x0, sig=23) at /home/derek/dr/src/core/lib/instrument.c:2243
[...]

PR #4524, opened 2020-11-05
i#4516: Add client thread support for AArchXX

### LINKED ISSUE #4516: AArch64 missing support for client threads
The AArch64 port #1569 failed to implement a number of features.  This one covers client threads, which is blocking enabling tests for #4460 (also xref #4474).

derekbruening:
Adds client thread support to AArchXX.
Implements dynamorio_clone for both ARM and AArch64 and adds necessary
TLS swapping, including handling initial tpidr* values pointing at
read-only memory.

Ports the client.signal test's assembly to a64 and arm.  Enables the
client.signal test on aarchxx.  The test covers the client thread support
here, as well as general client signal handling and manipulating.
Tested manually on arm with the #4522 fix.

Issue: #4474, #4516, #4522
Fixes #4516

### DIFF
--- a/core/arch/aarch64/aarch64.asm
+++ b/core/arch/aarch64/aarch64.asm
@@ -515,10 +515,21 @@ GLOBAL_LABEL(_dynamorio_runtime_resolve:)
 #endif /* UNIX */
 
 #ifdef LINUX
-
+/* thread_id_t dynamorio_clone(uint flags, byte *newsp, void *ptid, void *tls,
+ *                             void *ctid, void (*func)(void))
+ */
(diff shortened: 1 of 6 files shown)


derekbruening:
Hmm, the test fails on Jenkins.  It passed on the TX1 packet.net machine.  Maybe the review will spot something.  Otherwise may need access to Jenkins.

derekbruening:
Ah, I think I know what the Jenkins crash is: it's #4518, whose fix I had bundled into the tree I tested on the tx1.  let's update and see if that fixes it.

abhinav92003 (approved):

Issue #4436, opened 2020-09-03
client.alloc-noreset test failure: "memory iteration failed"


derekbruening:
I saw this once in PR #4435 but on re-running it did not reproduce.  I
could not reproduce it locally.  It seems rather unlikely to be related to
that PR.

https://ci.appveyor.com/project/DynamoRIO/dynamorio/builds/35001136
```
[00:07:44]         Start  93: code_api|client.alloc-noreset
[00:07:44] 
[00:07:44] 93: Test command: C:\projects\dynamorio\build\build_debug-internal-32\bin32\drrun.exe "-s" "90" "-quiet" "-debug" "-use_dll" "C:/projects/dynamorio/build/build_debug-internal-32/lib32/debug/dynamorio.dll" "-exit0" "-stderr_mask" "0xC" "-msgbox_mask" "0" "-dumpcore_mask" "0x7d" "-staged" "-no_enable_reset" "-no_vm_base_near_app" "-vm_base" "0x120000000" "-vmheap_size" "2G" "-code_api" "-c" "C:/projects/dynamorio/build/build_debug-internal-32/suite/tests/bin/client.alloc-noreset.dll.dll" "--" "C:/projects/dynamorio/build/build_debug-internal-32/suite/tests/bin/client.alloc-noreset.exe"
[00:07:44] 93: Test timeout computed to be: 600
[00:07:44] 93: thank you for testing the client interface
[00:07:44] 93:   testing global memory alloc...success
[00:07:44] 93:   testing nonheap memory alloc...success
[00:07:44] 93:   testing custom memory alloc....success
[00:07:44] 93:   testing custom windows alloc....success
[00:07:44] 93: error: memory iteration failed
[00:07:44] 93:   testing alignment....success
[00:07:44] 93: thread initialization:
[00:07:44] 93:   testing local memory alloc....success
[00:07:44] 93:   testing global memory alloc...success
[00:07:44] 93: code cache:
[00:07:44] 93:   testing local memory alloc....success
[00:07:44] 93:   testing global memory alloc...success
[00:07:44] 93:   testing nonheap memory alloc...success
[00:07:44] 93:   testing raw memory alloc...success
[00:07:44] 93:   testing query pretend-write....success
[00:07:44] 93: wrote to app code page successfully
[00:07:44] 93: all done

[00:17:05]         Start  79: code_api|client.alloc-noreset
[00:17:05] 
[...]

PR #4526, opened 2020-11-11
i#4495: Preserve translated stolen register

### LINKED ISSUE #4495: Translated stolen register app value lost on signal/synchall control transfer

derekbruening:
Currently, a stolen register value that is translated by a client is
discarded both in a synchall and a synchronous signal.  We fix both
paths here.

Adds a new test client.stolen-reg which tests both a synchronous fault
signal translation path as well as a synchall translation path via a
synchall flush.  The test was confirmed to fail without either of the
fixes in place.

Ensures the new test passes on 32-bit ARM.  This exercised the
translate_from_synchall_to_dispatch() path and found some ARM bugs:
the reset exit stub is A32 mode, so we need to change the mode for the
suspended sigcontext; yet thread_set_mcontext() did not support that.
That is all fixed now and verified.

Fixes #4495

### DIFF
--- a/suite/tests/client-interface/stolen-reg.c
+++ b/suite/tests/client-interface/stolen-reg.c
@@ -104,7 +104,7 @@ cause_sigsegv(void)
 {
     uintptr_t val = 0;
     /* Generate SIGSEGV with a sentinel in the stolen reg.
-     * This precise instruction sequence is matched by the client.
+     * This precise instruction sequence is matched by the stolen-reg.dll.c client.
      */
 #if defined(AARCH64)
     __asm__ __volatile__("mov x28, #" STRINGIFY(STOLEN_REG_SENTINEL) "\n\t"
@@ -131,7 +131,7 @@ cause_sigsegv(void)
 THREAD_FUNC_RETURN_TYPE
 thread_func(void *arg)
 {
-    /* The client looks for this exact sequence of instructions. */
+    /* The stolen-reg.dll.c client looks for this exact sequence of instructions. */
 #if defined(AARCH64)
     __asm__ __volatile__("mov x28, #" STRINGIFY(STOLEN_REG_SENTINEL) "\n\t"
                                                                      "nop\n\t"
@@ -182,7 +182,7 @@ main(int argc, char **argv)
 
     /* Now test synchall from another thread (the initiating thread does not
      * hit the i#4495 issue).
-     * The client looks for this exact sequence of instructions.
+     * The stolen-reg.dll.c client looks for this exact sequence of instructions.
      */
 #if defined(AARCH64)
     __asm__ __volatile__("mov x28, #" STRINGIFY(STOLEN_REG_SENTINEL) "\n\t"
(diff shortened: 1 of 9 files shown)


suite/tests/client-interface/stolen-reg.dll.c
```diff
@@ -0,0 +1,138 @@
...
+{
+    dr_fprintf(STDERR, "Performing synchall flush\n");
+    if (!dr_flush_region(NULL, ~0UL))
+        DR_ASSERT(false);
+    void *drcontext = dr_get_current_drcontext();
+    dr_mcontext_t mcontext;
+    mcontext.size = sizeof(mcontext);
+    mcontext.flags = DR_MC_ALL;
+    dr_get_mcontext(drcontext, &mcontext);
+    mcontext.pc = dr_app_pc_as_jump_target(dr_get_isa_mode(drcontext), next_pc);
```
derekbruening, line 65: PR #4507 has a bug where it fails to decorate the PC in a similar redirect for the new flush in tracer.cpp.  I'll submit a separate PR to fix that.

derekbruening:
Maybe the ARM mode fixes could be separated: but I'm not willing to make a separate test so they'd share this one.

derekbruening:
(Note that the appveyor failure is alloc-noreset == #4436)

suite/tests/client-interface/stolen-reg.dll.c
abhinav92003: I suppose you  mean PR #4491?

suite/tests/client-interface/stolen-reg.c
```diff
@@ -0,0 +1,218 @@
...
+        SIGLONGJMP(mark, 1);
+    }
+}
+
+static uintptr_t
+cause_sigsegv(void)
+{
+    uintptr_t val = 0;
+    /* Generate SIGSEGV with a sentinel in the stolen reg.
+     * This precise instruction sequence is matched by the client.
```
abhinav92003, line 107: Maybe add that the client code is in suite/tests/client-interface/stolen-reg.dll.c.

abhinav92003 (approved):

suite/tests/client-interface/stolen-reg.c
derekbruening: Every suite/tests/client-interface test has the app as foo.c and the cilent as foo.dll.c, so I would say it's not necessary in these tests, but I suppose an explicit comment doesn't hurt.  Done.

suite/tests/client-interface/stolen-reg.dll.c
derekbruening: I filed #4527 on this.

derekbruening: Woah, somehow I missed your pointer to #4491 -- yes, correct.  I'll update PR #4528 on merge.

PR #4528, opened 2020-11-11
i#4527: Set ISA mode for drcachesim flush redirect PC

### LINKED ISSUE #4527: APP CRASH: drcachesim delayed tracing broken on 32-bit ARM
Commit 72275ad4c52ef9d04bc4194e637c4f1516685b3d for #4496 via PR #4507 broke ARM delayed tracing:

```
$ bin32/drrun -t drcachesim -trace_after_instrs 20000 -- suite/tests/bin/simple_app
...
Hit delay threshold: enabling tracing.
<Application tried to execute from unreadable memory 0xb71db0a2.
This may be a result of an unsuccessful attack or a potential application vulnerability.>
<Application /home/derek/dr/git/build_dbg_tests/suite/tests/bin/simple_app (1075).  Application exception at PC 0xb71db0a2.  
Signal 11 delivered to application as default action.
Callstack:
	0xb71db0a2  
	0x0001e916  
	0x00000000  
>
```

It's due to an incorrect ISA mode on the dr_redirect_execution from the new synchall flush.

derekbruening:
Fixes drcachesim delayed tracing which was broken by commit 72275ad
for #4496 via PR #4507.  The fix is to decorate the target PC to
indicate the ISA mode.

Adds a reminder to the dr_flush_region_ex() documentation where it
talks about using dr_redirect_execution() to set the ISA mode.

Tested on ARM:
$ bin32/drrun -t drcachesim -trace_after_instrs 20000 -- suite/tests/bin/simple_app

Fixes #4527

### DIFF
--- a/core/lib/instrument_api.h
+++ b/core/lib/instrument_api.h
@@ -5896,6 +5896,12 @@ DR_API
  * The flags field of \p context must contain DR_MC_ALL; using a partial set
  * of fields is not suported.
  *
+ * For 32-bit ARM, be sure to use dr_app_pc_as_jump_target() to properly set the ISA
+ * mode for the continuation pc if it was obtained from instr_get_app_pc() or a
+ * similar source rather than from dr_get_proc_address().  This will set the least
+ * significant bit of the mcontext pc field to 1 when in Thumb mode
+ * (#DR_ISA_ARM_THUMB), \ref sec_thumb for more information.
+ *
  * \note dr_get_mcontext() can be used to get the register state (except pc)
  * saved in dr_insert_clean_call() or dr_prepare_for_call().
  *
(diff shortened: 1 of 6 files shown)


core/lib/instrument_api.h
```diff
@@ -6219,6 +6219,8 @@ DR_API
  * have been flushed. The context to use can be obtained via
  * dr_get_mcontext() with the exception of the pc to continue at which must be passed as
  * an argument to the callout (see instr_get_app_pc()) or otherwise determined.
+ * For 32-bit ARM, be sure to use dr_app_pc_as_jump_target() to properly set the ISA
```
abhinav92003, line 6222: This comment should probably be at `dr_redirect_execution` too, as the continuation pc is passed to that method.

I see other example usages where this is not done currently. Eg https://github.com/DynamoRIO/dynamorio/blob/c1be309c8f21f9eabd05b759378ee097e1af83b0/api/samples/cbr.c#L261

abhinav92003 (approved):
Thanks for the fix!

core/lib/instrument_api.h
derekbruening: dr_redirect_execution already has a comment about this, but perhaps it could be reworded.

derekbruening: Moved the doc to dr_redirect_execution, merged with what was already there, and raised it up rather than being a lower-down note.

Updated a sample, bbdup, and several tests.

derekbruening:
Appveyor is alloc-noreset again -- seems to be getting more common: #4436.

PR #4531, opened 2020-11-12
i#4297: Fix operand sizes for vcvtps2ph

khuey:
Fixes incorrect operand sizes for the vcvtps2ph instruction. 

Issue: #4297


khuey:
idk how to fix these failures. The test suite says "instr_encode error: no encoding found (see log)" but I don't see a log.

khuey:
Alright that seems to have done it (note to self, running locally with `ctest -R api.ir -VV` gets the output).

The ps2ph/ph2ps use half-width registers (and memory operands) in both their VEX and EVEX forms. This fixes the decoding table and the tests.

AssadHashmi:
run arm tests

derekbruening:
Was there an issue on this?  I don't see an issue number in the title or description.  It sounds familiar.

@johnfxgalea -- are you able to review as the triager?

khuey:
I did not file an issue.

derekbruening:
#4297 is the one I am thinking of: vcvtps2pd, plus apparently the binutils comparison test has the wrong data in it (due to a prior bug in binutils?).  I guess here binutils does not distinguish partial register sizes, nor any memory sizes, and so this change would not show up in the binutils comparison test?

khuey:
Yeah it doesn't look like the binutils test distinguishes (if it used Intel syntax we'd expect things like YMMWORD PTR->XMMWORD PTR changes but AT&T syntax doesn't have that).

johnfxgalea:
#4377 is somewhat relevant but #4297  fits more nicely.

khuey:
#4297 is about another instruction that this does not fix though.

johnfxgalea:
I was going to generalize that issue to accommodate a reference for this fix and avoid adding another similar issue page. Don't mind adding another issue though if you prefer.

johnfxgalea (approved):
LGTM, thanks for the fix!

khuey:
Feel free to do what you think is best I just wanted to be clear that there are multiple size-changing instructions with issues :)

johnfxgalea:
@khuey Merged, thanks again for the PR!

Issue #3775, opened 2019-08-08
Incorrect operand order between SSE and AVX


YValeri:
There is currently a problem regarding the way DynamoRIO manages AVX and SSE floating point instructions, explained in this post : (https://groups.google.com/forum/#!topic/DynamoRIO-Users/EwjJLo-fBdo). For AVX, the Intel standard (https://software.intel.com/sites/default/files/managed/39/c5/325462-sdm-vol-1-2abcd-3abcd.pdf) specifies that for the instruction are as followed :
"instr destination, source0, source1", so all the operands are explicitely given. However, for SSE, the standard differs a bit : "instr destination(=source0), source1", so the destination and first source are the same, the instructions are done in-place.
Regarding DynamoRIO's way of doing, it is the correct one for AVX, when using instr_get_src(instr, 0), we get the source0 of the instruction, and instr_get_src(instr, 1) gives the source1 (and instr_get_dst works fine aswell). On the contrary, for SSE, since the first operand is a destination AND source at the same time, DynamoRIO considers the destination as an explicit operand, and the source as implicit. That implies that the source is put at the last position in the instruction's sources, so instr_get_src(instr, 0) actually gives back the source1, while instr_get_src(instr, 1) gives source0.
From a simple logic point, this is pretty strange, but mostly, it is really problematic regarding instruction overload because one would have to manage SSE and AVX instructions differently because of DynamoRIO, not because of the Intel standard. Also, since no floating point operation is commutative, this problem actually change the values of the results which is extremely problematic.
Many instructions are impacted, here is a non limited list :
SSE -> {add|mul|sub|div}{s|p}{s|d}
AVX -> v{add|mul|sub|div}{s|p}{s|d}

Moreover, as stated in the same post, this is even more obvious for FMA instructions, because Intel manages these the same way as for SSE instructions : "instr destination(=source0), source1, source2". However, the instr for FMA is very specific for Intel, because the order of FMA is specified in the OP_code. So for instance, we can have the following :
"VFMADD231SS xmm0, xmm1, xmm2" -> multiply xmm1 by xmm2, add xmm0, put the result in xmm0.
Since the source0 is implicit, DynamoRIO places it at the end of the sources, so in third position, meaning that if don't specifically for this case, we would get the following :
[...]

Issue #2816, opened 2018-01-29
ASSERT: missing decode type in core/arch/x86/disassemble.c:227


abenkhadra:
In order to log the actual instructions observed at runtime, one can enable loglevel 3 by issuing the following command:

```
drrun -debug -loglevel 3  -c any_client -- any_app
```
This command runs without problems dumping the instructions in DR's own syntax. However, changing the disassembly syntax to `intel` or `att`, by adding command line option `-syntax_intel` or `-syntax_att` respectively,  would trigger the above mentioned assertion which can be found [here](https://github.com/DynamoRIO/dynamorio/blob/master/core/arch/x86/disassemble.c#L228).

I've debugged this issue a bit and it seems like the assertion is triggered by instruction `syscall` which is called by `ld-linux-x86-64.so.2`. The instruction `syscall` does not have a matching `optype`. The issue is independent of the client used or the instrumented application since it's triggered by the loader.

I'm using DR 6.2.0 on an  x86_64 machine running Ubuntu 16.04

derekbruening:
We would welcome a patch if you have one.

Issue #1683, opened 2015-04-13
support disassembly in ARM assembler format


derekbruening:
We want to add -syntax_arm which would match the standard ARM assembler
format, initially for our own use in tests that compare to other decoders,
but also for general user use.

derekbruening:
I'm documenting a design decision here:

```
*** TODO hide writeback implicit opnds for non-reg-list instrs

    /* Writeback implicit operands for non-list instrs */
    if ((nonlist_store || nonlist_load) &&
        /* Base reg is always last dst, and implicit srcs are after main srcs */
        ((dst && *idx == max-1) || (!dst && *idx >= tostore))) {
        opnd_t memop = nonlist_store ? instr_get_dst(instr, 0) : instr_get_src(instr, 0);
        CLIENT_ASSERT(opnd_is_base_disp(memop), "internal disasm error");
        /* We want to hide:
         *   1) Base reg as dst
         *   2) Base reg as src
         *   3) Disp as src, if present in memop
         *   4) Index reg as src, if present in memop
         *   5) Index shift type + amount, if present in memop
         * TODO: how distinguish base from index?  The order varies!
         * + Could reorder so wb base always last: always want to hide for -syntax_arm.
         *   That requires changing xop_shift.
         * + Could reorder so wb base always first: but that requires more
         *   xop_ entries for the different immeds, negative index, etc.
         * => "i#1683: writeback implicit base reg opnd must be last for -syntax_arm disasm"
```

derekbruening:
Having just written a bunch of code to deduce the ARM format, which covers
most of the general cases (we're going to need more code for some
out-of-order-operand opcodes and some special-case dst-as-src implicit
operands) I measured the perf on release build api.dis and it seems that,
maybe not surprisingly when building a 100MB file, the printing dominates
and the ARM format which does more work up front is faster b/c it prints
less:

```
# ls -sh /tmp/randombits 
7.1M /tmp/randombits
# for ((i=0; i<5; i++)); do LD_LIBRARY_PATH=lib32/release /usr/bin/time suite/tests/bin/api.dis /tmp/randombits -arm > /tmp/xx; done; for ((i=0; i<5; i++)); do DYNAMORIO_OPTIONS=-syntax_arm LD_LIBRARY_PATH=lib32/release /usr/bin/time suite/tests/bin/api.dis /tmp/randombits -arm > /tmp/xx2; done
[...]

PR #4533, opened 2020-11-14
i#4297: fix operand size for vcvtudq2pd

khuey:


johnfxgalea:
run arm tests

core/ir/x86/decode_table.c
```diff
@@ -7665,7 +7665,7 @@ const instr_info_t evex_W_extensions[][2] = {
     {OP_vcvtudq2ps, 0xf20f7a10, "vcvtudq2ps", Ve, xx, KEw, We, xx, mrm|evex|ttfv, x, END_LIST},
     {OP_vcvtuqq2ps, 0xf20f7a50, "vcvtuqq2ps", Ve, xx, KEb, We, xx, mrm|evex|ttfv, x, END_LIST},
   }, { /* evex_W_ext 61 */
-    {OP_vcvtudq2pd, 0xf30f7a10, "vcvtudq2pd", Ve, xx, KEb, We, xx, mrm|evex|tthv, x, END_LIST},
+    {OP_vcvtudq2pd, 0xf30f7a10, "vcvtudq2pd", Ve, xx, KEb, Wh_e, xx, mrm|evex|tthv, x, END_LIST},
```
johnfxgalea, line 7668: But vcvtudq2pd can also make use of the full XMM register for the following case: **VCVTUDQ2PD ymm1{k1}{z}, xmm2/m128/m32bcst**

See: https://www.officedaytime.com/simd512e/simdimg/convert.php?f=vcvtudq2pd

Isn't this the same problem as the X instruction, which you mentioned on the issue page? (https://github.com/DynamoRIO/dynamorio/issues/4297#issuecomment-727074492)

> * VEX vcvtdq2pd this one is weird, the source register is always an XMM (whether it's half or all is controlled by VEX.L). Is that really covered by Wvq_dq?

I suspect the solution is going to have two entries in the decode table for these two cases, but tbh, I never encountered this problem before either. @derekbruening Did you face such an issue in the past please?

johnfxgalea:
Add a short description to the PR, please. Also mention issue number at the end of the description, i.e., Issue: #XXXX or Fixes: #XXXX (the latter only if a PR closes the issue completely, which in this case does not).

core/ir/x86/decode_table.c
derekbruening: I'm less familiar w/ all the AVX-512 stuff and am just now looking up the "m32bcst": broadcast.  So the source can also be a 32-bit memory location that is broadcast into each element of the dest register.  The table does not seem to take that into account: there's only one entry, and neither `We` nor `Wh_e` allow that.  In decode.c I see mention of broadcast but it's only used for disp scaling?  Maybe I'm missing sthg but it's looking like the mem sizes for broadcasts are all wrong?  @hgreving2304 may know more.

hgreving2304: It's been a while but I'd say the m32bcst size was missed.

johnfxgalea: Hey, @hgreving2304 Thanks for your reply.

Perhaps we can punt m32bcst for now and not keep @khuey hanging unless @khuey wants to address it?

W.R.T We vs. Wh_e, was We used to cover the two source cases where the XMM register is partially and completely used? Or do you think we need to add another entry to the table for the instruction?

khuey: For m32bcst do we need to add another entry to the table? Is there an example instruction somewhere that *does* handle broadcasts appropriately?

khuey: To answer my own question.

```
khuey@minbar:~/dev/scratch$ cat test.s
	vfmadd213ps (%rdi), %zmm15, %zmm16
	vfmadd213ps (%rdi){1to16}, %zmm15, %zmm16
	
khuey@minbar:~/dev/scratch$ as test.s
khuey@minbar:~/dev/scratch$ objdump -d a.out

a.out:     file format elf64-x86-64


Disassembly of section .text:

0000000000000000 <.text>:
   0:	62 e2 05 48 a8 07    	vfmadd213ps (%rdi),%zmm15,%zmm16
   6:	62 e2 05 58 a8 07    	vfmadd213ps (%rdi){1to16},%zmm15,%zmm16
khuey@minbar:~/dev/scratch$ ~/dev/dynamorio/obj/clients/bin64/drdisas 62 e2 05 48 a8 07
 62 e2 05 48 a8 07    vfmadd213ps zmm16 {k0}, zmm15, [rdi]
khuey@minbar:~/dev/scratch$ ~/dev/dynamorio/obj/clients/bin64/drdisas 62 e2 05 58 a8 07
 62 e2 05 58 a8 07    vfmadd213ps zmm16 {k0}, zmm15, [rdi]
```

It appears that DR's support for the broadcast flag is incomplete.

AssadHashmi:
run arm tests

AssadHashmi:
run arm tests

core/ir/x86/decode_table.c
derekbruening: Yes, let's separate broadcast.  I filed #4534 on that.

derekbruening: @johnfxgalea -- I'm trying to reconcile the URL you're citing showing a full xmm reg with the Intel manual.  My version of the manual has some clear typos in it.  Checking a newer one.

derekbruening: Well the latest version has the same typos (mixing up xmm vs ymm in part of the description) but has the same {xmm,xmm,ymm} source pattern as in that other page, though the typo made me wonder whether to trust that.  The other text does point to the {x,x,y} pattern as well though.  I had thought that VEX.L would be half of a ymm (so 4 32-bit source ints) and VEX.L' would be half of a zmm (so 8 32-bit source ints).  Maybe confirm experimentally?  And then make sure the register picker for ymm or xmm does the right thing (I think it's based on the size though); check all the other uses of `Wh_e` == `OPSZ_half_16_vex32_evex64`: are more (or all?) really supposed to be {x,x,y} instead of {x,y,z}?

Answering the question about a new entry in the table: that could be done, but since nothing else changes a new (or redefined) size type would be sufficient.

johnfxgalea: By any chance, is your manual similar to https://www.felixcloutier.com/x86/vcvtudq2pd ? I noticed odd details also there.

khuey: The Intel manuals are riddled with errors. Just as one example, the narrowest version VPMOVQD is described as having an m128 output while the essentially equivalent VPMOVSQD is correctly described as having an m64 output.

The GNU assembler/disassembler do agree that the correct register sequence for half-width operands is xmm,xmm,ymm though, and they refuse to accept a ymm for the 128 bit instructions.

khuey: VPMOVQB, per the manual, takes XMM registers for each variant. So it seems that the half,quarter,eighth width operands in EVEX should be XMM registers if they fit.

If we agree I'll do that in a separate PR and then rebase this on top of it.

derekbruening: > By any chance, is your manual similar to https://www.felixcloutier.com/x86/vcvtudq2pd ? I noticed odd details also there.

I am looking at the official PDF from Intel ("Intel 64 and IA-32 Architectures Software Developer’s Manual").

Agreed, there are have always been some errors in the Intel manuals.

Since this rises above just a size and affects the register name that is listed, the binutils tests should be covering these: why haven't they complained if we think something is ymm when it's really xmm?

khuey: Even for VEX prefixes binutils uses XMM if it fits

```khuey@minbar:~/dev/scratch$ objdump -d a.out

a.out:     file format elf64-x86-64


Disassembly of section .text:

0000000000000000 <.text>:
   0:   c4 e3 79 1d ca 01       vcvtps2ph $0x1,%xmm1,%xmm2
   6:   c4 e3 7d 1d ca 01       vcvtps2ph $0x1,%ymm1,%xmm2```

Were the decoded instructions in drdecode_decenc_x86_64.expect actually taken from binutils?

johnfxgalea: There is no binutil test for vcvtudq2pd; I searched in drdecode_decenc_x86.expect

https://raw.githubusercontent.com/DynamoRIO/dynamorio/master/third_party/binutils/test_decenc/drdecode_decenc_x86.expect

derekbruening: > Were the decoded instructions in drdecode_decenc_x86_64.expect actually taken from binutils?

@hgreving2304 set them up.  Looking at 4e99a45734d3ec97e388516596315e0e9dd7c829: rather than being binutils output tweaked to match DR's imperfect Intel syntax it looks like they are DR output hand-checked.  Hmm.  So as the commit log says they are not golden references.  I wonder if we can correct that and take the binutils output and tweak whatever pieces of syntax are different.

johnfxgalea: Is there an issue raised regarding: "DR's imperfect Intel syntax"? I tried to search for it. If in the future we fix DR's intel syntax, do we solve the problem?

derekbruening: > Is there an issue raised regarding: "DR's imperfect Intel syntax"? I tried to search for it. If in the future we fix DR's intel syntax, do we solve the problem?

I thought there were several individual issues: #344, #2816, #3775.  DR has no first-class support for disassembly and tries to deduce everything, when it might be better to go store format strings for each opcode in the tables: though that has its own downsides.  Xref ARM #1683 and AArch64 #4382 which have the same problems.

Yes, you would think we could massage the binutils output to match DR's: that's what we did with the `suite/tests/api/dis-udis86-randtest.raw` output.  @hgreving2304 were they so far output it was just not feasible to do so and that's why you took the DR output?

johnfxgalea: Okay, so action points for this PR are as follows: 

- [ ] Add a new size type to denote that the operand may be partial or a full xmm.
- [ ] Add a check for vcvtudq2pd in binutils-based tests. 

@derekbruening @khuey lmk if you agree

khuey: > Add a new size type to denote that the operand may be partial or a full xmm.

Are there any examples of this size operand where that's *not* the case though? It looks to me like the entire class is just wrong.

derekbruening: I made a new `Component-IR` label and took a few mins to label existing issues which should make it easier to look at decode-related outstanding issues: https://github.com/DynamoRIO/dynamorio/labels/Component-IR

khuey: @johnfxgalea Having looked at the Intel manual some more I believe all fractional SIMD registers are rendered as XMM, unless they are 256 bits in which case they're rendered as YMM.

Also I think it would make sense to update the binutils tests and bring over all the AVX512/etc instructions which are missing at once rather than cherry pick vcvtudq2pd here.

So I propose to do neither of your action items before merging :)

johnfxgalea: > @johnfxgalea I believe all fractional SIMD registers are rendered as XMM, unless they are 256 bits in which case they're rendered as YMM.

Okay, thanks for checking.

> Also I think it would make sense to update the binutils tests and bring over all the AVX512/etc instructions which are missing at once rather than cherry pick vcvtudq2pd here.

Okay. It would have been nice a test from binutils to assert the source operand pattern raised  [here](https://github.com/DynamoRIO/dynamorio/pull/4533/files/1b69f7eae27d84112ec5c264d11f52880aff941d#r524395430).

johnfxgalea (approved):

core/ir/x86/decode_table.c
khuey: > > Also I think it would make sense to update the binutils tests and bring over all the AVX512/etc instructions which are missing at once rather than cherry pick vcvtudq2pd here.
> 
> Okay. It would have been nice to have a test from binutils to assert the source operand pattern raised [here](https://github.com/DynamoRIO/dynamorio/pull/4533/files/1b69f7eae27d84112ec5c264d11f52880aff941d#r524395430).

I don't disagree, I just think we should pull in all the AVX-512 instruction tests from binutils at once.

Issue #4534, opened 2020-11-16
AVX-512 broadcast memory operand sizes incorrect


derekbruening:
DR's IR is not representing AVX-512 broadcast memory operand sizes correctly.  It ignores the broadcast bit EVEX.b and portrays the size as though there were no broadcast.  This was noticed here: https://github.com/DynamoRIO/dynamorio/pull/4533#discussion_r523430591

khuey:
There are several issues here:

- Instructions that do not accept EVEX.b at all (e.g. vpaddb and everything else in the E4.nb/E4NF.nb subclasses) should fail to decode
- Instructions that accept EVEX.b to enable broadcasting (e.g. vorpd) need appropriate memory operand sizing when EVEX.b=1.
- Instructions that accept EVEX.b to control rounding (e.g. vsqrtss) need the current behavior.

And note that there are instructions that fall into categories 2 and 3 depending on whether the relevant operand is a memory operand or a register operand (e.g. vsubpd).

Handling category 1 instructions can be done by adding a REQUIRES_EVEX_B_0 flag (a la REQUIRES_VEX_L_0).

The broadcast operand size depends on the lane width of the operation so there are full size operations with both 64 and 32 bit broadcasts (e.g. vcvtps2dq and vcvtpd2qq) as well as half size operations with 32 bit broadcasts (e.g. vcvtps2pd). Adding new OPSZ codes for these would require 3 additions (e.g. OPSZ_16_vex32_evex64_bcst64, OPSZ_16_vex32_evex64_bcst32, and OPSZ_half_16_vex32_evex64_bcst32). So we could do that and then teach the relevant code to check for EVEX.b when encoding/decoding these OPSZs.

Thoughts?

derekbruening:
SGTM.  The only code I saw that looks at the broadcast bit is for the compressed displacement scaling (and I wonder if that would not be necessary if the size were correct, b/c the scaling is the memory size (always?)) so it sounds right that the other pieces you mention are all missing.  Thanks for looking into the details!

Issue #4537, opened 2020-11-18
OPSZ_8x16 is in the wrong place in size_names


khuey:
This looks like an oversight in 099ba138c45ddd4a9603fb3447b7a03eeb53884b

Issue #4536, opened 2020-11-18
reqL1 appears twice in the decode table entry for kandnw


khuey:

derekbruening:
Note that sthg like this that does not actually affect the code (or similar things like typos in comments or sthg) does not necessarily need an issue filed: one suggestion would be to just post the fix PR on an iX-* branch.

khuey:
Ok I'll keep that in mind in the future.

PR #4538, opened 2020-11-18
Various fixes to EVEX.b handling

khuey:
Largely for #4534. I'd like to get feedback on this before I do the rest of the bcst operands.

One thing I'd be particularly interested in is where the tests for instructions that _don't_ decode are.


derekbruening:
It seems odd to file separate issues on the small problems (#4536, #4537) but then not fix them separately and instead bundle them into something else: I would expect a separate PR per filed issue.

derekbruening:
> One thing I'd be particularly interested in is where the tests for instructions that don't decode are.

The api.dis test may be the only current one trying to cover a variety of encodings.  The api.ir* tests have some particular tests (e.g., encodings invalid in 32-bit mode but valid in 64-bit).  (There are others testing fault handling for illegal instrs but they do not cover many encodings.)

khuey:
Alright this looks good. The various pieces are self explanatory, I think. Throwing the vpermb patch in the middle there was a mistake in retrospect but disentangling it would be rather painful because of the decode table renumbering.

khuey:
Because of #4555 I think we need a different approach here.

PR #4543, opened 2020-11-19
i#4542: Allow XINST_CREATE_sub with sp source on A64

### LINKED ISSUE #4542: XINST_CREATE_sub fails to take sp as first source on A64
On A64 you can encode `sub` with `sp` as the first source, yet `XINST_CREATE_sub` fails to encode:

```
XNST_CREATE_sub(dcontext, opnd_create_reg(DR_REG_X1), opnd_create_reg(DR_REG_SP),
                             opnd_create_reg(DR_REG_X2));
=>
<Internal Error: Failed to encode instruction: 'sub    %sp %x2 lsl $0x0000000000000000 -> %x1'
```

It's because only the extended-register form takes `sp`, while the macro hardcodes the shifted-register form.  We should have it pick the extended.

derekbruening:
Adds a fix allowing the 'sp' register as the first source operand of
an ADD or SUB by using the extended-register rather than
shifted-register form on AArch64.

Adds a test which fails without the fix.

Fixes #4542

### DIFF
--- a/core/ir/aarch64/instr_create.h
+++ b/core/ir/aarch64/instr_create.h
@@ -466,8 +466,10 @@ enum {
  */
 
 /** \cond disabled_until_i4106_is_fixed */
-#define INSTR_CREATE_add(dc, rd, rn, rm_or_imm) \
-    INSTR_CREATE_add_shift(dc, rd, rn, rm_or_imm, OPND_CREATE_LSL(), OPND_CREATE_INT(0))
+#define INSTR_CREATE_add(dc, rd, rn, rm_or_imm)                                     \
+    /* _extend supports sp in rn, so prefer it. */                                  \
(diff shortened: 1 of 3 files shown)


derekbruening:
Assad is out today -- @abhinav92003 if you could take a look?

abhinav92003:
LGTM.

The issue is only when the second operand is a reg right? The followed works for me in the current impl:  `sub    %sp $0x0000000000000008 lsl $0x0000000000000000 -> %sp`

abhinav92003 (approved):

derekbruening:
> The issue is only when the second operand is a reg right? The followed works for me in the current impl: `sub %sp $0x0000000000000008 lsl $0x0000000000000000 -> %sp`

Yes that is right.

Issue #1698, opened 2015-05-22
ldrex..strex pair constraints challenge instrumentation and even core operation


derekbruening:
A ldrx..strex pair has some constraints that make inserting
instrumentation in between, or even core DR operation, challenging:

From the ARMv8 manual, section E2.10.5:

  An implementation might clear an exclusive monitor between the LoadExcl
  instruction and the StoreExcl, instruction without any application-related
  cause. For example, this might happen because of cache evictions.  Software
  must, in any single thread of execution, avoid having any explicit memory
  accesses or cache maintenance instructions between the LoadExcl instruction
  and the associated StoreExcl instruction.

  Implementations can benefit from keeping the LoadExcl and StoreExcl
  operations close together in a single thread of execution. This minimizes
  the likelihood of the exclusive monitor state being cleared between the
  LoadExcl instruction and the StoreExcl instruction. Therefore, for best
  performance, ARM strongly recommends a limit of 128 bytes between
  LoadExcl and StoreExcl instructions in a single thread of execution.

One-time switches back to dispatch (during code discovery, or synchall or
something) should be fine, because the ldrex..strex code has to loop and
be prepared to failed a few times.  The problems are all related to
inserted memory operations that execute every single time.

Thus we have problems with:

+ Stolen register mangling in between or on ldrex/strex themselves

+ Inserted instrumentation

There could be multiple cbr's in between ldrex and strex, so do we give up
on getting into a single bb?  Should we study sequences we see in the wild
to see if having some assumptions would make the problem drastically
easier?  Should we have a special trace head trigger and stopp supporting
-disable_traces?

One suggestion is adding a check for each instrumentation memory reference
to skip if it's in the middle of ldrex/strex: but this gets tricky without
allowing memory refs itself.

zhaoqin:
It looks like the hardware is forgiving if there is only a few memory reference in between, (based on my experiment with memtrace instrumentation, inline instrumentation is fine, but large clean call would cause problem), so I would not worry about the stolen reg handling.

Just for memtrace, if we assume there is no memory reference in between and the hardware is forgiving, we can just not insert the clean call.

derekbruening:
We need to know what "only a few" means, and how it varies with different hardware: let's not build something that only runs on one machine.  It could be that just a few memory refs that happen to collide on cache lines will cause a hang, so it's not clear yet that we can consider inline instrumentation or even stolen reg mangling to be safe.

algr:
[...]

PR #4544, opened 2020-11-19
i#1698 ldstex: Add exclusive memop create and query support

derekbruening:
i#1698 ldstex: Add exclusive memop create and query support

Adds a new API function instr_is_exclusive_load().  Moves the existing
private implementation (for -unsafe_build_ldstex) to become a public
function on AArch64.

Adds new instruction creation macros for exclusive load, store, and
clear opcodes.

Adds tests to api.ir for instr_is_exclusive_* and the new creation
macros.  This necessitated switching api.ir to use runcmp.cmake, since
its output is now too large for a CMake regular expression match.

Issue: #1698

### DIFF
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -176,6 +176,7 @@ Further non-compatibility-affecting changes include:
  - Added dr_flush_region_ex API that accepts a callback to be executed after synch
    flush but before the threads are resumed. The existing dr_flush_region API
    is modified to invoke dr_flush_region_ex with a NULL callback.
+ - Added instr_is_exclusive_load().
 
 **************************************************
 <hr>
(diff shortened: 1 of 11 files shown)


derekbruening:
Note: this is chained on PR #4543.  Please select just the 2nd commit in the diff view when reviewing.

derekbruening:
> Note: this is chained on PR #4543. Please select just the 2nd commit in the diff view when reviewing.

I should have made this PR vs that branch instead of master for a true chain -- but these are small diffs so I think it should be doable to review by selecting just this commit.

derekbruening:
Going to do a proper PR vs that branch, which will help b/c that branch needs a fix.

PR #4545, opened 2020-11-19
i#1698 ldstex: Add XINST_CREATE_{load,store}_pair

derekbruening:
Adds new XINST_CREATE_{load,store}_pair utilities for AArchXX, driven
by cross-platform needs in exclusive monitor handling.

Adds A64 sanity tests.

Issue: #1698

Issue #4532, opened 2020-11-13
Remove superfluous non-operands from LDXR, STXR, and related instructions on AArch64


derekbruening:
On AArch64 we have superfluous operands in the exclusive load and store instructions in DR's IR:

```
  0x0000ffff7d459b24  885f7c01   ldxr   (%x0)[4byte] $0x1f $0x1f -> %w1
  0x0000ffff7d459af0  885ffc01   ldaxr  (%x0)[4byte] $0x1f $0x1f -> %w1

  0x0000ffff7d459afc  88027c04   stxr   %w4 $0x1f -> (%x0)[4byte] %w2
  0x0000ffff7d459b28  8803fc02   stlxr  %w2 $0x1f -> (%x0)[4byte] %w3
```

These `0x1f` integers seem to be just part of the encoding and aren't actually operands.  I would expect them to not show up in the IR at all.

derekbruening:
CLREX shouldn't have this 0 either:
```
 +52   m4 @0x0000fffd47084ff8  d503305f   clrex  $0x0000000000000000
```

derekbruening:
There are still 2 references to `i#4532` in the code base: I think PR #5234 failed to remove/address those.

Also during the merge of PR #5234 the `Fixes` line was deleted from its final commit description: I guess it was the line in the PR itself that still managed to auto-close this?

derekbruening:
CLREX still has an immediate int operand: `clrex  $0x0000000000000000`.  The code points at this issue:
```
core/ir/aarch64/instr_create_api.h:/* TODO i#4532: Remove this superfluous operand. */
core/ir/aarch64/instr_create_api.h-#define INSTR_CREATE_clrex(dc) instr_create_0dst_1src(dc, OP_clrex, OPND_CREATE_INT(0))
```
So I think this should be re-opened.

joshua-warburton:
CLREX is a different case to the other instructions in that it does actually have an optional imm argument, but that argument is ignored, unlike the other superfluous arguments which were there to handle the potentially variable "sticky bits." It also defaults to a value of 15 for the imm if it is not set

This is similar to udf, which also has an ignored, optional operand.

Considering this, we should have two variants in order to comply with the spec:
11010101000000110011xxxx01011111  n   58        clrex                : imm4
11010101000000110011111101011111  n   58        clrex

derekbruening:
[...]

PR #4546, opened 2020-11-19
i#1698 ldstex: Add exclusive memop create and query support

derekbruening:
Adds a new API function instr_is_exclusive_load().  Moves the existing
private implementation (for -unsafe_build_ldstex) to become a public
function on AArch64.

Adds new instruction creation macros for exclusive load, store, and
clear opcodes.

Adds tests to api.ir for instr_is_exclusive_* and the new creation
macros.  This necessitated switching api.ir to use runcmp.cmake, since
its output is now too large for a CMake regular expression match.

Issue: #1698


derekbruening:
OK trying this chained PR again vs the PR #4543.

derekbruening:
Hmm the diff has some of the base in it.  Maybe if I had merged the fix a different way.  Github doesn't have great support for chained PR's.  In any case it's still quite reviewable IMHO; will update and ensure green before merge of course.  Let me know if it's not easy to look at.

derekbruening:
@abhinav92003 if you could take this one over from Assad as well.  #4545 is chained on top of this so this needs to go in before it.

suite/tests/api/ir_aarch64.c
```diff
@@ -4342,6 +4342,84 @@ test_sys(void *dc)
...
+                               OPND_CREATE_MEM8(DR_REG_X1, 0));
+    ASSERT(instr_is_exclusive_load(instr));
+    test_instr_encoding(dc, OP_ldxrb, instr);
+    instr = INSTR_CREATE_ldxrh(dc, opnd_create_reg(DR_REG_W0),
+                               OPND_CREATE_MEM16(DR_REG_X1, 0));
+    ASSERT(instr_is_exclusive_load(instr));
+    test_instr_encoding(dc, OP_ldxrh, instr);
+    instr = INSTR_CREATE_ldxp(dc, opnd_create_reg(DR_REG_W0), opnd_create_reg(DR_REG_W1),
+                              OPND_CREATE_MEM64(DR_REG_X2, 0));
+    ASSERT(instr_is_exclusive_load(instr));
```
abhinav92003, line 4364: Perhaps also test the 64-bit version of `ldxp` and `stxp`?

abhinav92003 (approved):

suite/tests/api/ir_aarch64.c
derekbruening: Done.

derekbruening:
Ah, the tests are failing on clrex, which had a fix (it has a fake source!  #4532 covers removing) that got separated when I made this separate branch for this PR separate from the main #1698 tree.  Fixed now.

Issue #125, opened 2014-11-27
re-enable tests: linux.vfork, linux.vfork-fib, win32.debugger, security-common.selfmod-big


derekbruening:
_From [derek.br...@gmail.com](https://code.google.com/u/117968039472581148324/) on April 13, 2009 09:57:06_

several tests were disabled in the past.  this case covers re-investigating
whether we can enable them now.

xref for vfork:
PR 207477 - Determina Case 2999: Linux regression failures: segfault and
vfork-fib
PR 209506 - Determina Case 5028: linux: re-enable vfork and signal nightly
tests once they don't hang/ruin machines

_Original issue: http://code.google.com/p/dynamorio/issues/detail?id=125_

PR #4548, opened 2020-11-19
i#4537: Move OPSZ_8x16 to the correct location in size_names

### LINKED ISSUE #4537: OPSZ_8x16 is in the wrong place in size_names

khuey:
This appears to be an oversight from when it was added. It got added to the very end of the list after the fractional register sizes, when it should not have been.

Fixes #4537

### DIFF
--- a/core/ir/decode_shared.c
+++ b/core/ir/decode_shared.c
@@ -150,6 +150,7 @@ const char *const size_names[] = {
     "OPSZ_32_of_64",
     "OPSZ_4_of_32_evex64",
     "OPSZ_8_of_32_evex64",
+    "OPSZ_8x16",
     "OPSZ_1_of_4",
     "OPSZ_2_of_4",
     "OPSZ_1_of_8",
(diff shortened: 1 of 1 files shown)


khuey:
I suppose it's possible somehow but I'm skeptical that the code_api|client.winxfer failure here is related to my change.

The log also says `127/234 Test #125: code_api|client.winxfer ......................................   Passed    2.30 sec` so I have no idea what to make of that.

derekbruening:
> I suppose it's possible somehow but I'm skeptical that the code_api|client.winxfer failure here is related to my change.
> 
> The log also says `127/234 Test #125: code_api|client.winxfer ...................................... Passed 2.30 sec` so I have no idea what to make of that.

That's the 32-bit one.  The 64-bit one failed:
```
[00:17:25] 110/254 Test #109: code_api|client.winxfer ......................................***Failed  Required regular expression not found. Regex=[^(kernel_
xfer_event: type 2
...
```
It is #4058 flaky output ordering which I tried to fix once already but it still fails sometimes...grrr.

derekbruening (approved):

Issue #4549, opened 2020-11-19
Migrate ASAP from Travis CI as it no longer offers free OSS CI and will stop working very soon


derekbruening:
Announced here: https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

We have been burning through our 10K credit for the last two weeks.  When it runs out everything will just stop working unless we sign up for a paid account, but those do not look nearly as nice as what we had: $69/month gets just one concurrent job!

There seems to be a way to apply for a certain number of free OSS minutes every month but it sounds like we'd have to re-apply each month and it is not clear what the chance is of getting them nor whether we'd get enough to cover our usage.

I can't seem to find a way to view the current credit, so it's not clear when we'll run out, but I would imagine any minute.

Xref #4131 which lists possible alternatives of Github Actions and CircleCI.  Or Appveyor in a reversal of that issue.

derekbruening:
Github Actions looks like the best choice as it has high concurrency limits and seems to have similar features to (the old) Travis.

https://docs.github.com/en/free-pro-team@latest/actions/reference/usage-limits-billing-and-administration
20 concurrent jobs, incl 5 macos

Migration docs:
https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/migrating-from-travis-ci-to-github-actions

derekbruening:
Here's a quick tasklist for migration from Travis CI to Github Actions to help track progress and help parallelize the effort.  Maybe when someone starts on one of these they can edit this comment and put their name at the end of its line or sthg?  Each of these includes investigation into the options and syntax plus testing.  Maybe each of us working on it would make a branch with changes and we can merge them at some point.

Base setup:
- [ ] Xenial
- [ ] trigger on PR or on master push
- [ ] notify dynamorio-devs on failures
- [ ] log access

Duplicate each of the 7 current Travis test jobs and ensure the right
software is installed (different VM's have different base software) and the
tests are green (different OS/libs/etc. can cause different failures):
- [ ] 32-bit Linux gcc + tests
- [ ] 64-bit Linux gcc + tests
- [ ] aarchxx cross-compile
- [ ] android arm cross-compile
- [ ] 64-bit Linux clang-9 + format check
- [ ] 64-bit OSX clang + tests
- [ ] A64-on-x86 gcc + tests

Ditto for the package cronjobs:
- [ ] x86
- [ ] aarch64
- [ ] arm
- [ ] Android

Other package steps:
[...]

PR #4552, opened 2020-11-19
i#4540: Preserve extended opnd attributes in opnd_replace_reg()

### LINKED ISSUE #4540: opnd_replace_reg does not preserve negated/shifted/extended flags
opnd_replace_reg() creates a new operand that throws away all the flags from the original.  These aarchxx flags indicate negation, shifting, and extension.

derekbruening:
Adds preservation of AArch64 extended reg and base-disp attributes in
opnd_replace_reg().  Adds some AArch64 basic tests.

Fixes #4540

### DIFF
--- a/core/ir/opnd_shared.c
+++ b/core/ir/opnd_shared.c
@@ -1240,7 +1240,9 @@ opnd_replace_reg(opnd_t *opnd, reg_id_t old_reg, reg_id_t new_reg)
 
     case REG_kind:
         if (old_reg == opnd_get_reg(*opnd)) {
-            *opnd = opnd_create_reg(new_reg);
+            *opnd = opnd_create_reg_ex(
+                new_reg, opnd_is_reg_partial(*opnd) ? opnd_get_size(*opnd) : 0,
+                opnd_get_flags(*opnd));
             return true;
         }
         return false;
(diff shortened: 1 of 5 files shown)


suite/tests/api/ir_aarch64.c
```diff
@@ -4447,6 +4447,47 @@ test_xinst(void *dc)
...
+    found = opnd_replace_reg(&op, DR_REG_W7, DR_REG_W1);
+    ASSERT(!found);
+    found = opnd_replace_reg(&op, DR_REG_X7, DR_REG_X1);
+    ASSERT(found);
+    ASSERT(opnd_get_base(op) == DR_REG_X1);
+    ASSERT(opnd_get_flags(op) == DR_OPND_EXTENDED);
+    extend = opnd_get_index_extend(op, &scaled, &amount);
+    ASSERT(extend == DR_EXTEND_SXTX && scaled && amount == 3);
+
+    /* XXX: test other routines like opnd_defines_use() */
```
AssadHashmi, line 4488: Perhaps also test with:
- other register operand properties, specifically shift and negate
- index register when creating opnd_create_base_disp_aarch64()

core/ir/opnd_shared.c
```diff
@@ -1255,8 +1257,10 @@ opnd_replace_reg(opnd_t *opnd, reg_id_t old_reg, reg_id_t new_reg)
             reg_id_t i = (old_reg == oi) ? new_reg : oi;
             int d = opnd_get_disp(*opnd);
 #if defined(AARCH64)
```
AssadHashmi, line 1259: Should we also handle index reg shifts?

AssadHashmi (commented):
Minor point: web API documentation for `opnd_replace_reg()` doesn't say anything about return value.

derekbruening:
> Minor point: web API documentation for `opnd_replace_reg()` doesn't say anything about return value.

I thought I fixed that.   Checking: ok it's in the follow-up PR #4553.

core/ir/opnd_shared.c
AssadHashmi: > Should we also handle index reg shifts?

Doh! Scratch that, I just checked `opnd_get_index_extend()`.

AssadHashmi (approved):

suite/tests/api/ir_aarch64.c
derekbruening: Added a test of base-disp with an index reg and replacing the index reg.

For shift and negate: the ir_arm.c existing test is already testing those for A32.  Just added to the comment for A64.

derekbruening:
We hit the Travis limit: no more Travis runs.  #4549 covers migrating.  Not worried about this one as the 1st commit was green and the next modified only A64 code; going to merge it.

PR #4553, opened 2020-11-19
i#4541: Add instr_replace_reg_resize() and opnd_replace_reg_resize()

### LINKED ISSUE #4541: add instr_replace_reg() which covers overlapping regs and resizes to match
Today we have opnd_replace_reg(), which only replaces an exact match, and does not resize to match (as it is typically used to replace one reg with a differently sized version of itself).

This is a request for an instr-level routine to replace which cover overlapping regs and resizes to match.  This would initially be used for #1698.  I would want to be able to say "replace x28 with x0" and have it modify `strex w28, x1, [x2]` to become `strex w0, x1, [x2]`.

derekbruening:
Adds new API routines instr_replace_reg_resize() and
opnd_replace_reg_resize() which replace instances of an overlapping
register with a new register of the same size.

Adds x86 and a64 tests to the api.ir test.

Fixes #4541

### DIFF
--- a/core/ir/opnd.h
+++ b/core/ir/opnd.h
@@ -2756,6 +2756,8 @@ DR_API
 /**
  * Assumes that both \p old_reg and \p new_reg are DR_REG_ constants.
  * Replaces all occurrences of \p old_reg in \p *opnd with \p new_reg.
+ * Only replaces exact matches (use opnd_replace_reg_resize() to match
+ * size variants).
  * Returns whether it replaced anything.
  */
 bool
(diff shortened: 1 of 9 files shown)


core/ir/opnd.h
```diff
@@ -2756,10 +2756,20 @@ DR_API
 /**
  * Assumes that both \p old_reg and \p new_reg are DR_REG_ constants.
  * Replaces all occurrences of \p old_reg in \p *opnd with \p new_reg.
+ * Returns whether it replaced anything.
```
abhinav92003, line 2759: Maybe clarify explicitly that `opnd_replace_reg` will not replace size variants, and point to `opnd_replace_reg_resize`.

abhinav92003 (approved):

core/ir/opnd.h
derekbruening: Done.

derekbruening:
Appveyor is alloc-noreset #4436 and winxfer (again!) #4058.
Travis is down: #4549.  Ran api.ir on x86 manually.

PR #4559, opened 2020-11-24
i4549 Github Action: Add CI test for clang-format, osx and a64-on-x86

abhinav92003:
Migrates remaining continuous integration testing to Github Actions.

Creates multiple GA workflows by platform: ci-x86 (for 32- and 64-bit), ci-aarchxx (for aarchxx cross compile, android arm cross compile, and a64-on-x86), ci-osx (for 64-bit), and ci-clang-format. Multiple smaller workflows are better than a single monolithic one; particularly, this allows re-running a smaller set of jobs instead of all of them (e.g. when re-running due to flaky tests).

Enables GA CI on all push events, which includes commits to all branches, including master. 

Adds some comments to previously added test suites too.

This PR also documents some testing done to verify this GA CI setup. Some bugs were added on purpose to various tests, to verify if the corresponding CI(s) fails, which they did.

Issue: #4549

### DIFF
--- a/.github/workflows/ci-test-suites.yml
+++ b/.github/workflows/ci-test-suites.yml
@@ -98,7 +98,7 @@ jobs:
         DYNAMORIO_CROSS_AARCHXX_LINUX_ONLY: no
 
   # AArchXX cross-compile with gcc, no tests:
-  archxx-cross-compile:
+  aarchxx-cross-compile:
     runs-on: ubuntu-16.04
 
     steps:
--- a/.github/workflows/ci-test-suites.yml
+++ b/.github/workflows/ci-test-suites.yml
@@ -62,7 +62,7 @@ jobs:
     # But fetching multiple branches isn't supported yet: actions/checkout#214
     # Pending PR that adds this support actions/checkout#155
     # TODO i#4549: When necessary support is available, remove/replace the
-    # workaround here and below.
+    # workaround here and from every job below too.
     - run: git fetch --no-tags --depth=1 origin master
 
     # Install multilib for non-cross-compiling Linux build:
--- a/.github/workflows/ci-test-suites.yml
+++ b/.github/workflows/ci-test-suites.yml
@@ -28,14 +28,12 @@
 # OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
 # DAMAGE.
 
-# Github Actions workflow for Continuous Integration testing on Linux.
-# TODO i#4549: This is a work-in-progress and not enabled currently.
+# Github Actions workflow for Continuous Integration testing on Linux and MacOS.
 
-name: ci-linux
+name: ci-test-suites
 on:
   # Run on pushes to any repository branch.
-  # TODO i#4549: Uncomment when Github Actions integration complete.
-  # push:
+  push:
 
   # Manual trigger using the Actions page. May remove when integration complete.
   workflow_dispatch:
@@ -50,6 +48,7 @@ defaults:
   # https://github.community/t/reusing-sharing-inheriting-steps-between-jobs-declarations
 
 jobs:
+  # 32-bit Linux build with gcc and run tests:
   x86:
     runs-on: ubuntu-16.04
 
@@ -65,6 +64,7 @@ jobs:
     # TODO i#4549: Remove/replace this workaround when necessary support is added.
     - run: git fetch --no-tags --depth=1 origin master
 
+    # Install multilib for non-cross-compiling Linux build:
     - name: Create Build Environment
       run: |
         sudo apt-get -y install doxygen vera++ cmake zlib1g-dev libsnappy-dev \
@@ -76,6 +76,7 @@ jobs:
       env:
         DYNAMORIO_CROSS_AARCHXX_LINUX_ONLY: no
 
+  # 64-bit Linux build with gcc and run tests:
   x86-64:
     runs-on: ubuntu-16.04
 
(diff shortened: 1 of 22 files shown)


derekbruening:
Still says "archxx-cross-compile"?

abhinav92003:
> Still says "archxx-cross-compile"?

Oops. The old version must've sneaked in somehow while working on this PR. Will fix it.

As part of verifying correctness of the migration to GA CI, I'm trying to induce some test failures by intentionally adding some bugs. Will add details here and mark the PR ready for review once that is done.

.github/workflows/ci-test-suites.yml
```diff
@@ -28,14 +28,12 @@
 # OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
 # DAMAGE.
 
-# Github Actions workflow for Continuous Integration testing on Linux.
-# TODO i#4549: This is a work-in-progress and not enabled currently.
+# Github Actions workflow for Continuous Integration testing on Linux and MacOS.
```
abhinav92003, line 31: @derekbruening Note that all the test suites are part of one workflow. One drawback of this setup is that any re-run will start all the test suites again (unlike Travis where we could re-run individual suites), which seems a waste of resources. We could alternatively split this into 4 workflows, one for x86, all arm, macos, and clang-format.

derekbruening: Is there any downside to having 4 workflows instead of 1?

.github/workflows/ci-test-suites.yml
```diff
@@ -142,3 +148,74 @@ jobs:
         DYNAMORIO_CROSS_ANDROID_ONLY: yes
         DYNAMORIO_ANDROID_TOOLCHAIN: /tmp/android-gcc-arm-ndk-10e
       run: ./suite/runsuite_wrapper.pl travis
+
+  # 64-bit Linux build with clang, no tests (runsuite.cmake disables the tests),
+  # install and require clang-format.
+  # We used to build 32-bit with clang but have decided that is not worth
+  # the Travis resources: 64-bit hits most clang-only warnings and 64-bit
```
derekbruening, line 155: s/Travis/CI/

.github/workflows/ci-test-suites.yml
```diff
@@ -65,6 +64,7 @@ jobs:
     # TODO i#4549: Remove/replace this workaround when necessary support is added.
```
derekbruening, line 64: Maybe add a note that this is present in every job below so someone doesn't just remove this one.

derekbruening (approved):

.github/workflows/ci-test-suites.yml
abhinav92003: I couldn't find anything else other than code duplication, which wouldn't be that much anyway.

If there are multiple workflows for some event, all of them will be triggered on the event (https://github.community/t/execution-order-for-multiple-workflows-in-one-repo/116780). There's a limit on number of concurrent jobs (https://docs.github.com/en/free-pro-team@latest/actions/reference/usage-limits-billing-and-administration#usage-limits), but that will remain the same in both multiple or single workflow design.

.github/workflows/ci-test-suites.yml
abhinav92003: Done.

.github/workflows/ci-test-suites.yml
abhinav92003: Done.

.github/workflows/ci-test-suites.yml
abhinav92003: I've split it into multiple smaller workflows. PTAL if this seems fine.

abhinav92003: Perhaps the names of individual jobs can be simplified: 'osx-64bit' -> '64-bit', because the workflow name provides the remaining context. But may be better to keep them verbose, so that they make sense even out of context.

derekbruening: I would keep the job name as is, but perhaps `s/ci-test-suite-/ci-/`, so `ci-osx / osx-64bit`.  "ci" kind of implies a test suite.  Or `s/ci-test-suite-/suite/` for `suite-osx / osx-64bit`.  Deployment jobs would probably not have 'ci" or "suite" in the name.

I would also rename the "x86" job to "x86-32".  Maybe "osx-x86-64" to match (the implied "linux-").

derekbruening (approved):

.github/workflows/ci-test-suites.yml
abhinav92003: Done. Thanks.

abhinav92003:
@derekbruening As Travis is yellow, I'm not able to merge this PR myself. Can you please do it? I have updated the final commit message in the PR's description at the top.

After this is submitted, we can remove Travis, and add the new CI as a blocker for submission in repo settings.

derekbruening:
> After this is submitted, we can remove Travis, and add the new CI as a blocker for submission in repo settings.

Done.

derekbruening:
> > After this is submitted, we can remove Travis, and add the new CI as a blocker for submission in repo settings.
> 
> Done.

Well, I didn't *remove* Travis configs or setup, I just removed it as required to merge.  I figure we'll wait a bit before removing it entirely.

abhinav92003:
> Well, I didn't _remove_ Travis configs or setup, I just removed it as required to merge. I figure we'll wait a bit before removing it entirely.

Sure makes sense, that's what I intended to say too.

PR #4561, opened 2020-11-24
i#4560: Handle AArch64 zero registers consistently

### LINKED ISSUE #4560: aarch64 XZR considered GPR but WZR is not and is not handled in multiple API routines
Working on #1698 I'm hitting problems trying to use XZR.
It looks like reg_is_gpr() says the DR_REG_XZR is in fact a GPR, while
DR_REG_WZR is not: looks like a bug.  The DR_REG_START_* defines have
neither as a GPR.

Plus we have failure to handle XZR in several register routines:
```
    <Application /home/derek/dr/build/suite/tests/bin/client.ldstex (68163) DynamoRIO usage error : reg_64_to_32: passed non-64-bit reg>
<Usage error: reg_64_to_32: passed non-64-bit reg (/home/derek/dr/src/core/ir/opnd_shared.c, line 2220)

#4  0x00000000710c3ea0 in external_error (file=0x713d6ab0 "/home/derek/dr/src/core/ir/opnd_shared.c", line=2220, 
    msg=0x713d7a20 "reg_64_to_32: passed non-64-bit reg") at /home/derek/dr/src/core/utils.c:201
#5  0x00000000712283a4 in reg_64_to_32 (reg=34) at /home/derek/dr/src/core/ir/opnd_shared.c:2219
#6  0x0000000071228518 in reg_resize_to_opsz (reg=34, sz=6 '\006') at /home/derek/dr/src/core/ir/opnd_shared.c:2329
#7  0x0000000071234e14 in instr_create_save_to_tls (dcontext=0xfffd74004080, reg=34, offs=168) at /home/derek/dr/src/core/ir/instr_shared.c:3538
#8  0x0000000071314b84 in mangle_exclusive_monitor_op (dcontext=0xfffd74004080, ilist=0xfffd7402dbe0, instr=0xfffd7402f608, next_instr=0xfffd7402fda0)
    at /home/derek/dr/src/core/arch/aarchxx/mangle.c:3383
#9  0x00000000713066a8 in d_r_mangle (dcontext=0xfffd74004080, ilist=0xfffd7402dbe0, flags=0xfffd7402ae08, mangle_calls=true, record_translation=false)
```

At first I thought it would be simpler to consider the zero regs as GPR's,
which simplifies reg_get_size(), reg_resize_to_opsz(), etc. to not need
special-case handling.
OTOH drreg doesn't want to try to hand out xzr as a scratch register: so
keeping them non-GPR is probably best.
So the action items are to fix reg_is_gpr() to not include ZXR, and to add
special handling to the reg size routines..

derekbruening:
Fixes reg_is_gpr() to return false for both XZR and WZR.

Adds handling of the two zero register names to the register resize routines.

Adds some tests to api.ir.

Fixes #4560

### DIFF
--- a/core/ir/aarch64/instr.c
+++ b/core/ir/aarch64/instr.c
@@ -332,7 +332,8 @@ instr_predicate_is_cond(dr_pred_type_t pred)
 bool
 reg_is_gpr(reg_id_t reg)
 {
-    return (DR_REG_X0 <= reg && reg <= DR_REG_WSP);
+    return (reg >= DR_REG_START_64 && reg <= DR_REG_STOP_64) ||
+        (reg >= DR_REG_START_32 && reg <= DR_REG_STOP_32);
 }
 
 bool
--- a/core/ir/opnd_shared.c
+++ b/core/ir/opnd_shared.c
@@ -2208,6 +2208,10 @@ reg_32_to_8(reg_id_t reg)
 reg_id_t
 reg_32_to_64(reg_id_t reg)
 {
+#    ifdef AARCH64
+    if (reg == DR_REG_WZR)
+        return DR_REG_XZR;
+#    endif
     CLIENT_ASSERT(reg >= REG_START_32 && reg <= REG_STOP_32,
                   "reg_32_to_64: passed non-32-bit reg");
     return (reg - REG_START_32) + REG_START_64;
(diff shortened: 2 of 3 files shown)


AssadHashmi (approved):

derekbruening:
Ran api.ir manually.

PR #4570, opened 2020-12-01
i#4549 Github Actions: Trigger on pull requests

derekbruening:
Changes the Github Actions trigger from pushes on all branches, which
fails to trigger on a pull request from a forked repository, to pull
requests and pushes on master.

Issue: #4549

### DIFF
--- a/.github/workflows/ci-aarchxx.yml
+++ b/.github/workflows/ci-aarchxx.yml
@@ -38,6 +38,7 @@ on:
     branches:
       - master
   pull_request:
+    types: [opened, reopened, synchronize, edited]
 
   # Manual trigger using the Actions page. May remove when integration complete.
   workflow_dispatch:
--- a/.github/workflows/ci-aarchxx.yml
+++ b/.github/workflows/ci-aarchxx.yml
@@ -32,8 +32,12 @@
 
 name: ci-aarchxx
 on:
-  # Run on pushes to any repository branch.
+  # Run on pull requests and pushes to master.
+  # We do not run on pushes to feature branches: pull requests cover those.
   push:
+    branches:
+      - master
+  pull_request:
 
   # Manual trigger using the Actions page. May remove when integration complete.
   workflow_dispatch:
--- a/.github/workflows/ci-clang-format.yml
+++ b/.github/workflows/ci-clang-format.yml
@@ -32,8 +32,12 @@
 
 name: ci-clang-format
 on:
-  # Run on pushes to any repository branch.
+  # Run on pull requests and pushes to master.
+  # We do not run on pushes to feature branches: pull requests cover those.
   push:
+    branches:
+      - master
+  pull_request:
 
   # Manual trigger using the Actions page. May remove when integration complete.
   workflow_dispatch:
--- a/.github/workflows/ci-clang-format.yml
+++ b/.github/workflows/ci-clang-format.yml
@@ -38,6 +38,7 @@ on:
     branches:
       - master
   pull_request:
+    types: [opened, reopened, synchronize, edited]
 
   # Manual trigger using the Actions page. May remove when integration complete.
   workflow_dispatch:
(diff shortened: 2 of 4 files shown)


.github/workflows/ci-aarchxx.yml
```diff
@@ -32,8 +32,12 @@
 
 name: ci-aarchxx
 on:
-  # Run on pushes to any repository branch.
+  # Run on pull requests and pushes to master.
+  # We do not run on pushes to feature branches: pull requests cover those.
   push:
+    branches:
+      - master
+  pull_request:
```
abhinav92003, line 40: The `pull_request` event is triggered by various sub-events; perhaps we don't need all of them? https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request

Also, could you try pushing a commit to this PR's branch, to check whether new pushes to existing PR indeed trigger the CI? (it could be covered under the `edited` event, but not sure)

derekbruening: I looked at those docs while creating this change and it was a little confusing.  It says "By default, a workflow only runs when a pull_request's activity type is opened, synchronize, or reopened. To trigger workflows for more activity types, use the types keyword."  But it doesn't clearly define the types: it points at the REST docs which also does not define them.  I was hoping the default would match what both Travis and Appveyor have, i.e., creation plus pushes, but maybe not.

Searching others shows that "opened, synchronize, and edited" are a common combo:

https://github.com/search?q=pull_request+extension%3A.yml+path%3A.github%2Fworkflows&type=Code

So maybe "opened" means the initial creation, "synchronize" means update by merging in master, and "edited" means a push?  Posting this response here does not seem to count as "edited".

derekbruening: I think that's what we want: the default plus "edited".  It did trigger on my new push.  I suppose with this level of control we could remove "edited" and have a trigger on all pushes to all branches if we want to run it on non-pull-request branches but I'm not sure that's what we want: and that would not be in sync with Appveyor.  Maybe once we've removed Appveyor if we switch to GA for Windows we can revisit at that point; for now I think it's best to mirror what we had before.

abhinav92003 (approved):

derekbruening:
A64 failure is #4571.

abhinav92003:
> Posting this response here [in a review comment thread] does not seem to count as "edited".

@derekbruening On PR #4579 I noticed that adding a PR-wide comment (like this one; instead of a review comment thread) triggers the CI again.

derekbruening:
> > Posting this response here [in a review comment thread] does not seem to count as "edited".
> 
> @derekbruening On PR #4579 I noticed that adding a PR-wide comment (like this one; instead of a review comment thread) or a title change also triggers the CI again.

If it were limited to sthg like the "run arm tests" that could be a feature but on every top-level comment is too much.  Probably we should remove the 'edited' and trigger on pushes to all branches and just live with extra runs on collaborative in-progress branches (which are rare for us anyway).

derekbruening:
I wonder if we remove "edited" if we will no longer have a trigger on a PR from a forked repo when the forked repo has a push update, though?

derekbruening:
There's a new `pull_request_target` event but it's not clear how to trigger that for forked repos but not for intra-repo-branch PR's.

derekbruening:
> There's a new `pull_request_target` event but it's not clear how to trigger that for forked repos but not for intra-repo-branch PR's.

I.e., I want my PR to be able to tweak the CI files.  Maybe it could do so by first replacing `pull_request_target` with `pull_request`?

But even so: we'd still have to live with triggers on a forked PR on every edit, if we did `edited` for the _target one to cover forks, making it behave differently from intra-PR's.

derekbruening:
OK I think `synchronize` covers a push to the forked repo: "updated due to a new push in the branch that the pull request is tracking" from https://github.blog/2011-10-19-all-of-the-hooks/.

derekbruening:
#4580

Issue #4571, opened 2020-12-01
drcachesim.delay-simple test failing on A64


derekbruening:
The delay-simple test is now failing every time on A64:

```
====> FAILURE in debug-internal-64 <====
	code_api|tool.drcachesim.delay-simple =>  Application /var/lib/jenkins/.jenkins/workspace/DynamoRIO-AArch64-Precommit/build/build_debug-internal-64/suite/tests/bin/simple_app (16772).  Internal Error: DynamoRIO debug check failure: /var/lib/jenkins/.jenkins/workspace/DynamoRIO-AArch64-Precommit/core/heap.c:1961 IF_WINDOWS(doing_detach ||) vmh-num_free_blocks == vmh-num_blocks - unfreed_blocks || ((ever_beyond_vmm IF_WINDOWS(|| get_os_version() = WINDOWS_VERSION_8_1))  vmh-num_free_blocks = vmh-num_blocks - unfreed_blocks) 
	code_api|api.static_sideline_FLAKY 

====> FAILURE in release-external-64 <====
release-external-64: 51 tests passed, **** 2 tests failed, of which 1 were flaky, but ignoring 0 for i#2417: ****
	code_api|tool.drcachesim.delay-simple 
```

@abhinav92003 since you recently change how that works (the flush type, etc.) could you take a look?  I think the leak assert reproduces on the TX1.

Xref analysis of when it started failing: https://github.com/DynamoRIO/dynamorio/pull/4529#issuecomment-734293283

abhinav92003:
The timing of the failures is definitely weird. 

PR #4491 changed the flush type, but there are lots of successes after 1783 (the Jenkins invocation for that PR) on http://139.178.83.194:8080/job/DynamoRIO-AArch64-Precommit/buildTimeTrend. Also PR #4559 (1870  on the Jenkins dash) shouldn't really have affected Jenkins at all, but it seems to have.

I'll trying to reproduce the issue on Tx1.

abhinav92003:
I confirmed locally that the failures in `tool.drcachesim.delay-simple` indeed start at commit 893c06c2605eb8423ef4a4b33185d0fef9c75257.  Not sure why it didn't show up on the Jenkins run for PR #4491 though (http://139.178.83.194:8080/job/DynamoRIO-AArch64-Precommit/1783/consoleFull).

derekbruening:
[...]

PR #4573, opened 2020-12-01
i#4058 flaky tests: Relax winxfer output regex

derekbruening:
Relaxes the client.winxfer test golden output regex further based on
observed output on Appveyor.

Issue: #4058


derekbruening:
A64 is #4571.

PR #4574, opened 2020-12-01
i#4436 alloc-noreset: Mark client.alloc-noreset as flaky

derekbruening:
Adds client.alloc-noreset to the auto-ignore list.  Although we'd like
to fix the memory iteration issue, we can't reproduce it locally, and
it is making the CI red often enough to be a pain.

Issue: #4436


derekbruening:
The x86 failure is filed as i#4575.
A64 is #4571.

PR #4577, opened 2020-12-02
Implement AVX-512 broadcasts

khuey:
This is a long series. I've tried to break it up into discrete semi-meaningful pieces but it's still going to be tough going for whoever reviews it.

I originally thought that we could do this with flags but the {sae}/{er} stuff (coming later) threw a big wrench in that. Some instructions have as many as three variants:

- The normal EVEX.L/EVEX.LL controlled SIMD register/memory operand of appropriate size case.
- The broadcast variant, which has destinations controlled by the EVEX prefix but takes a fixed size memory operand.
- The {sae}/{er} variant which does not use the EVEX prefix at all but is instead a fixed size (the highest the instruction offers, but notably in some cases it is not a ZMM register e.g. any -SS instruction).

The latter two are distinguishable only by the modrm byte, not by anything in the EVEX prefix, so when we add {sae}/{er} support some of the entries added here will become lookups into `mod_extensions`.

Everything here is tested with API tests and I've also run the binutils test suite (not included in this PR, since it also requires {sae}/{er}) and verified that all of their cases produce output (whether they produce the correct output is not verified, but roundtripping the binutils cases through the decoder/encoder did find some issues).

Open questions:
* Should we rename `evex_W_extensions` to `evex_WB_extensions` or something similar?
* Should we apply the OPCODE_TWOBYTES stealing to VEX prefixed instructions too and reclaim a bit for them, even though we have no current use for them.


khuey:
I'm just tagging @derekbruening and he can redirect it as necessary :)

derekbruening:
Unfortunately @hgreving2304 who did the original AVX-512 and is the most knowledgeable on the topic is no longer on the project and unlikely to be able to help.  I think only @johnfxgalea and myself are candidates for a good review...and I would have to look up these details like what "{sae}/{er}" means.  So I think a review may be delayed until someone has some time...please have patience.

khuey:
Alright, let me know if there's anything I can do to make that more tractable (including explaining things like the {sae}/{er} stuff)

johnfxgalea:
I might take a look at the weekend.

core/ir/x86/decode.c
```diff
@@ -1121,7 +1121,8 @@ read_instruction(byte *pc, byte *orig_pc, const instr_info_t **ret_info,
     } else if (info->type == EVEX_W_EXT) {
         /* discard old info, get new one */
         int code = (int)info->code;
-        int idx = (TEST(PREFIX_REX_W, di->prefixes) ? 1 : 0);
+        int idx = (TEST(PREFIX_REX_W, di->prefixes) ? 2 : 0) +
+            (TEST(PREFIX_EVEX_b, di->prefixes) ? 1 : 0);
```
derekbruening, line 1125: Yes, I would rename EVEX_W_EXT and evex_W_extensions.

The alternative is to split off a separate evex_b_extensions with another table lookup.  If the number of evex_W_extensions entries that care about evex.b is small, that may be nicer.  (If doing that, and if evex.b does make them invalid, we could do that via a REQUIRES_EVEX_B flag.)

core/ir/x86/decode_private.h
```diff
@@ -144,15 +144,17 @@ enum {
  * 2nd = 2nd byte of opcode (if there are 2)
  * 3rd (ls) = split into nibbles
  *   1st nibble (ms) = if bit 1 (OPCODE_TWOBYTES) set, opcode has 2 bytes
+ *                       if REQUIRES_EVEX then this bit instead means that this
+ *                       instruction must have evex.b set
```
derekbruening, line 148: I would expect this to instead be a flag like REQUIRES_VEX_L0?

In fact I'm not remembering why we have what seems a hacky overload of these opcode bits for the vex.W, etc. below: do we not have access to the flags field when we need that info?

core/ir/x86/encode.c
```diff
@@ -2835,7 +2838,7 @@ instr_encode_arch(dcontext_t *dcontext, instr_t *instr, byte *copy_pc, byte *fin
             info->opcode);
         info = get_next_instr_info(info);
         /* stop when hit end of list or when hit extra operand tables (OP_CONTD) */
-        if (info == NULL || info->type == OP_CONTD) {
```
derekbruening, line 2838: Why was OP_CONTD removed?

suite/tests/api/ir_x86_3args_avx512_evex_mask.h
```diff
@@ -545,16 +545,28 @@ OPCODE(vcvtps2pd_zhik7zhi, vcvtps2pd, vcvtps2pd_mask, X64_ONLY, REGARG(ZMM16), R
        REGARG(ZMM16))
 OPCODE(vcvtpd2ps_xlok0ld, vcvtpd2ps, vcvtpd2ps_mask, 0, REGARG(XMM0), REGARG(K0),
        MEMARG(OPSZ_16))
+OPCODE(vcvtpd2ps_xlok0bcst, vcvtpd2ps, vcvtpd2ps_mask, 0, REGARG(XMM0), REGARG(K0),
```
derekbruening, line 548: Could we also enable the binutils tests that cover these at the same time?

core/ir/x86/decode_table.c
```diff
@@ -1328,154 +1328,155 @@ const instr_info_t * const op_instr[] =
...
-    /* OP_vbroadcasti32x4 */  &evex_W_extensions[152][0],
-    /* OP_vbroadcasti32x8 */  &evex_W_extensions[153][0],
-    /* OP_vbroadcasti64x2 */  &evex_W_extensions[152][1],
-    /* OP_vbroadcasti64x4 */  &evex_W_extensions[153][1],
-    /* OP_vcompresspd     */  &evex_W_extensions[156][1],
-    /* OP_vcompressps     */  &evex_W_extensions[156][0],
-    /* OP_vcvtpd2qq       */  &evex_W_extensions[46][1],
-    /* OP_vcvtpd2udq      */  &evex_W_extensions[47][1],
-    /* OP_vcvtpd2uqq      */  &evex_W_extensions[48][1],
+    /* OP_valignd         */  &evex_W_extensions[155][0],
```
derekbruening, line 1331: Smaller diff if these are not shifted?  What was the reason to shift?

core/ir/x86/decode_table.c
```diff
@@ -3141,7 +3142,7 @@ const instr_info_t prefix_extensions[][12] = {
     {OP_movsd,  0xf20f1110, "movsd",  Wsd, xx, Vsd, xx, xx, mrm, x, END_LIST},
     {OP_vmovups,   0x0f1110, "vmovups", Wvs, xx, Vvs, xx, xx, mrm|vex, x, tevexw[0][0]},
     {MOD_EXT,    0xf30f1110, "(mod ext 10)", xx, xx, xx, xx, xx, mrm|vex, x, 10},
-    {OP_vmovupd, 0x660f1110, "vmovupd", Wvd, xx, Vvd, xx, xx, mrm|vex, x, tevexw[2][1]},
+    {OP_vmovupd, 0x660f1110, "vmovupd", Wvd, xx, Vvd, xx, xx, mrm|vex, x, tevexw[2][2]},
```
derekbruening, line 3145: Just trying a sanity check: so this was pointing at the entry for rex.w being set: so this requires rex.w to be set?  Should that mean its opcode should have OPCODE_MODRM set, which would be 0x40?  But it does not (only has 0x10)?  Am I missing something?

core/ir/x86/decode_table.c
```diff
@@ -7450,632 +7439,1029 @@ const instr_info_t vex_W_extensions[][2] = {
...
+    {OP_vmovq, 0x660f7e50, "vmovq", Ey, xx, Vq_dq, xx, xx, mrm|vex, x, tevexw[136][2]},
   },
 };
 
 /****************************************************************************
  * Instructions that differ depending on evex.W.
  * Index is evex.W value
  */
-const instr_info_t evex_W_extensions[][2] = {
+const instr_info_t evex_W_extensions[][4] = {
```
derekbruening, line 7451: The comment above has to be updated.
It should explain the indices.  So it is -w-b,-w+b,+w-b,+w+b?

core/ir/x86/decode_table.c
```diff
@@ -7450,632 +7439,1029 @@ const instr_info_t vex_W_extensions[][2] = {
...
   {    /* evex_W_ext 0 */
-    {OP_vmovups, 0x0f1010,"vmovups", Ves,xx,KEd,Wes,xx,mrm|evex|ttfvm,x,tevexw[1][0]},
+    {OP_vmovups, 0x0f1000,"vmovups", Ves,xx,KEd,Wes,xx,mrm|evex|ttfvm,x,tevexw[1][0]},
+    {INVALID, 0x0f1010,"(bad)", xx,xx,xx,xx,xx,no,x,NA},
```
derekbruening, line 7454: So this opcode really becomes invalid if evex.b is set?

core/ir/x86/decode.c
khuey: Ultimately 88 out of 265 entries don't vary at all on evex.b.

core/ir/x86/decode_private.h
khuey: Those flags are only used for decoding, not encoding.

core/ir/x86/encode.c
khuey: There are now OP_CONTD entries that are not END_LIST.

suite/tests/api/ir_x86_3args_avx512_evex_mask.h
khuey: Perhaps, though it'd be rather tedious.

core/ir/x86/decode_table.c
khuey: There was no reason other than putting vpermb next to vpermw.

core/ir/x86/decode_table.c
khuey: This doesn't actually require REX.W=1 with the VEX prefix, it's a WIG instruction.

core/ir/x86/decode_table.c
khuey: Yes, that's correct.

core/ir/x86/decode_table.c
khuey: Yes. The EVEX prefixed vmovups belongs to exception class E4.nb which #UD if EVEX.b=1.

core/ir/x86/decode_table.c
khuey: I do wonder if there are tests somewhere for these instructions with C4 VEX prefixes and both values of W testing that the W is in fact IGnored. The binutils tests I see via git grep for vmovupd are all C5 VEX encodings.

khuey: Seem to decode ok at least.

khuey@minbar:~/dev/dynamorio/obj$ ./clients/bin64/drdisas -syntax dr c5 fd 10 f4
 c5 fd 10 f4          vmovupd %ymm4 -> %ymm6
khuey@minbar:~/dev/dynamorio/obj$ ./clients/bin64/drdisas -syntax dr c4 e1 7d 10 f4
 c4 e1 7d 10 f4       vmovupd %ymm4 -> %ymm6
khuey@minbar:~/dev/dynamorio/obj$ ./clients/bin64/drdisas -syntax dr c4 e1 fd 10 f4
 c4 e1 fd 10 f4       vmovupd %ymm4 -> %ymm6

suite/tests/api/ir_x86_3args_avx512_evex_mask.h
derekbruening: Just to clarify: we'd want to enable them in any case eventually: you mean it would be a lot of work to enable a subset of them and you plan to enable all of them all at once later?

khuey: Right

core/ir/x86/decode.c
derekbruening: OK, so the vast majority need all.  OK, seems reasonable how you have it.

core/ir/x86/decode_private.h
derekbruening: Hmm, it seems like we could use them for encoding, right?  Maybe at least add a comment about switching to that in the future, which seems a little cleaner.

core/ir/x86/encode.c
derekbruening: OK.

core/ir/x86/decode_table.c
derekbruening: OK.

derekbruening (approved):
Seems reasonable.  I could see not doing the shift for a simpler diff and simpler history. but it's already there so maybe just leave it.

core/ir/x86/decode.c
khuey: Is EVEX_WB_EXT and evex_WB_extensions fine? or do you have other preferred names?

derekbruening: > Is EVEX_WB_EXT and evex_WB_extensions fine?

SGTM

khuey:
> 
> 
> Seems reasonable. I could see not doing the shift for a simpler diff and simpler history. but it's already there so maybe just leave it.

Yeah I think it'd be more work than it's worth to untangle it at this point.

Other review comments addressed.

derekbruening:
Thanks for taking the time to improve the decoder.  You now have commit privileges so you are able do the squash-and-merge: just please clean up the commit message (remove all those lines starting with *, etc. from the individual intermediate commit titles.: see https://github.com/DynamoRIO/dynamorio/wiki/Code-Reviews#merging-into-master-and-cleaning-up-the-final-commit-message).

khuey:
Alright, hopefully I did that correctly!

PR #4579, opened 2020-12-02
i#4571: Fix drcachesim delay-simple test on AArch64.

abhinav92003:
Add missing stolen reg value in mcontext.

PR #4513 is already working on setting the stolen reg value. This PR is just to verify whether this also fixes the delay-simple test. I verified that it is indeed fixed locally, but this test was somewhat flaky on Jenkins CI, so need to verify that. 

Fixes: #4571

### DIFF
--- a/core/lib/instrument.c
+++ b/core/lib/instrument.c
@@ -6833,7 +6833,7 @@ dr_set_mcontext(void *drcontext, dr_mcontext_t *context)
 {
     priv_mcontext_t *state;
     dcontext_t *dcontext = (dcontext_t *)drcontext;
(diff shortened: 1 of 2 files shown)


abhinav92003:
The `delay-simple` failure indeed goes away with this: http://139.178.83.194:8080/job/DynamoRIO-AArch64-Precommit/1899/console 

`client.drwrap-test-callconv` and `client.drwrap-test` have started failing though, but I see that PR  #4513 is already working on that.

abhinav92003:
Fixed by PR #4513.

PR #4580, opened 2020-12-02
i#4549 Github Actions: Trigger on all pushes, not on edits

derekbruening:
Changes the Github Actions trigger to remove the pull request "edited"
action, which is triggered on any top-level comment, title changes,
etc. and is thus more than we want.  Instead we trigger on a push to
any branch, plus pull request creation and "synchronize" which seems
to cover code updates to the pull request branch even when in a forked
repo.

Issue: #4549


derekbruening:
Unfortunately it's not de-duplicating: it's running everything twice, once for a push trigger and once for a pull_request trigger.  Plus, we seem to have yet more flaky tests: I already spent some time yesterday on some and filed several issues.  Seems to be flakier than the Travis runs.

abhinav92003:
> Unfortunately it's not de-duplicating: it's running everything twice, once for a push trigger and once for a pull_request trigger.

I assume we don't want this PR yet then?

derekbruening:
> > Unfortunately it's not de-duplicating: it's running everything twice, once for a push trigger and once for a pull_request trigger.
> 
> I assume we don't want this PR yet then?

Any suggestion?  We can either add logic to de-dup (I see some examples), or we can try removing PR creation and keeping just "synchronize" and hope that's only later changes, or just live w/ the "edit" we have now.

derekbruening:
> > > Unfortunately it's not de-duplicating: it's running everything twice, once for a push trigger and once for a pull_request trigger.
> > 
> > 
> > I assume we don't want this PR yet then?
> 
> Any suggestion? We can either add logic to de-dup (I see some examples), or we can try removing PR creation and keeping just "synchronize" and hope that's only later changes, or just live w/ the "edit" we have now.

"synchronize" being later changes would still duplicate intra-repo-branch pushes though.

We could change "push" back to being master-only, and keep the other changes here of removing "edit": maybe that'll do it.

abhinav92003 (approved):

derekbruening:
Even though some other PR's don't seem to trigger runs on top-level comments, we don't want runs on title edits, so we still want this.

PR #4583, opened 2020-12-04
i#4370 cmake 3.19: Do no clobber client defines

derekbruening:
Changes the absolute set of COMPILE_DEFINITIONS from PR #4582 for
clients to an append, to avoid breaking clients who set their own
defines (like Dr. Memory).

Tested on Dr. Memory, which fails to configure without this fix when
using the latest DR.

Issue: #4370


derekbruening:
Small bug fix; putting in TBR.

PR #4584, opened 2020-12-04
i#4370 cmake 3.19: Use proper target lib for rpath

derekbruening:
PR #4582 hardcoded "dynamorio" as the library to link to, breaking
users (like Dr. Memory) who rely on rpath linking to layers of their
own libraries.  We fix that here.

Tested on Dr. Memory, where drfuzz tests fail without this fix when
using the latest DR.

Issue: #4370


derekbruening:
Ideally there'd be a DR test that would catch this.  It would take layers of libs each stored in a different directory.  Hmm.  May bail on it due to complexity and live with DrMem catching this; trying to fix DrMem GithubActions and this is a prereq.

abhinav92003 (approved):

Issue #4131, opened 2020-02-24
Investigate Appveyor alternatives for more parallel builds: Github Actions; Travis; CircleCI


derekbruening:
Appveyor's big downside is the single sequential job serialized across both dynamorio and drmemory repositories and across PR's and master merges.  One single job is now down to ~23-24 mins thanks to #4059 which is not that different from a Travis invocation, but the lack of parallelism does cause pain.

Possible alternatives that now include Windows (all quite recently):
+ Github Actions
+ Travis: looks like it has only VS2017 and very few packages and is in early stages
+ CircleCI

derekbruening:
Xref #4549 where we almost certainly need to migrate from Travis rather urgently: and Appveyor is a possibility, in a reversal of this issue.

derekbruening:
Assigning to me since I started some work on Dr. Memory's Windows CI on Github Actions which overlaps with the work for DR: https://github.com/DynamoRIO/drmemory/pull/2326

derekbruening:
If we successfully add auto-cancellation (#4585) to Github Actions that could be another big reason to move from Appveyor, avoiding the current pain: https://github.com/DynamoRIO/dynamorio/wiki/Test-Suite#ci-resources-and-appveyor-global-serialization

derekbruening:
I removed the Appveyor webhook and swapped to the GA jobs as the ones that are required to merge.

derekbruening:
We now have every single CI job under ~15 mins (64-bit Windows is sometimes 16 mins or so).  It's pretty nice.  Add in no serialization across PR's or repos, and auto-cancellation of prior jobs, and it's really nice.

Unfortunately there are still a number of tests failing on Windows which are just on an ignore list, but that's how it was with Appveyor when we moved to Win10 VS2017.  Hopefully someone will get some time to fix a few of them.  I think it would be best to file separate issues instead of lumping them all under here or #4058 so I will take that as an action item to file them all separately (the set failing on GA is slightly different from Appveyor).

derekbruening:
Separate issues for test failures include #2246, #4616, #4617, #4618, #4621, #4622.
That covers the highest priority ones.  Closing this now.  Leaving #4058 open until all of them are separated out or fixed.

derekbruening:
#4619 also.

PR #4586, opened 2020-12-05
i#4131: Set up Windows Github Actions

derekbruening:
Adds two separate jobs on Windows, 32-bit and 64-bit.
We target a VS2017 Win10-1607 image to match what we had on Appveyor.

Modifies the runsuite_wrapper.pl script to *not* fork, since there is
no Cygwin perl available by default.  Instead, we tee to a file.

This is based on the similar work in progress for Dr. Memory:
https://github.com/DynamoRIO/drmemory/pull/2326

Issue: #4131


abhinav92003 (approved):

.github/workflows/ci-windows.yml
```diff
@@ -0,0 +1,169 @@
...
+      working-directory: ${{ github.workspace }}
+      run: |
+        echo ------ Setting up paths ------
+        7z x c:\projects\install\ninja.zip -oc:\projects\install\ninja > nul
+        set PATH=c:\projects\install\ninja;%PATH%
+        7z x c:\projects\install\doxygen.zip -oc:\projects\install\doxygen > nul
+        set PATH=c:\projects\install\doxygen;%PATH%
+        call "C:/Program Files (x86)/Microsoft Visual Studio/2017/Enterprise/VC/Auxiliary/Build/vcvars32.bat"
+        echo ------ Running suite ------
+        echo PATH is "%PATH%"
```
abhinav92003, line 167: These seem to be temporary debug statements. Perhaps they need to be removed?

derekbruening: Sorry, just saw this after merging (took a while to edit the merge message).  Yes, but it is very useful and doesn't take up much space and was in fact present in the Appveyor config the whole time as well.  Could be removed but since it was there before I included it here.

PR #4598, opened 2020-12-09
i#4597: Fix drconfig error reading debug status

### LINKED ISSUE #4597: drconfig reports incorrect debug information due to buffer re-use
This was discovered here: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!topic/dynamorio-users/J0boO-VEZxo

Code inspection shows what looks like a bug where the path to compare to the debug path is clobbered by reusing the same buffer to read a numeric parameter here:
https://github.com/DynamoRIO/dynamorio/blob/master/libutil/dr_config.c#L1346
Thus it will always print Debug=no with that bug.

derekbruening:
Fixes a bug where drconfiglib clobbers the buffer holding the library
path before it reads it to check for debug vs release build.

Adds a test of debug to libutil.drconfig_test which fails without this
fix.

Fixes #4597

### DIFF
--- a/libutil/dr_config.c
+++ b/libutil/dr_config.c
@@ -1341,6 +1341,15 @@ read_process_policy(IF_REG_ELSE(ConfigGroup *proc_policy, FILE *f),
         return;
     }
 
+    if (debug != NULL) {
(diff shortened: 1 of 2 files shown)


johnfxgalea (approved):
LGTM

Issue #4585, opened 2020-12-05
add auto-cancellation of prior CI jobs when a PR is updated


derekbruening:
In #4549 we moved from Travis CI to Github Actions.  Travis would auto-cancel if a new commit was within a few seconds of a prior one.  Appveyor does not do this, and it is a huge pain: see https://github.com/DynamoRIO/dynamorio/wiki/Test-Suite#ci-resources-and-appveyor-global-serialization

Looks like it might be possible to use someone else's support for auto-cancel out of the box: https://github.community/t/is-it-possible-to-cancel-previous-jobs-automatically-when-the-pull-request-is-updated/16779

If we also implement #4131 and move from Appveyor to Github Actions, maybe we could finally be free of queued up jobs and wasted CI resources.

derekbruening:
Final action item here is to update the wiki to remove or edit the part about manually cancelling CI jobs: I'll take care of that now.

PR #4599, opened 2020-12-09
i#4585: Set up auto-cancellation of prior CI jobs

derekbruening:
i#4585: Set up auto-cancellation of any prior in-progress Github Actions jobs for a PR when new jobs are triggered for that PR


derekbruening:
It seems to work.  I ensured it would not cancel a simultaneous job running for a different PR.  I have a conditional to prevent cancellation for pushes to master since we want to run all of those.

derekbruening:
The commit description will be:

i#4585: Set up auto-cancellation of prior CI jobs

Sets up auto-cancellation of any prior in-progress Github Actions jobs for a PR when new jobs are triggered for that PR.  Jobs for pushes to master are never canceled.

Fixes #4585

derekbruening:
One downside to having these separate workflows is that it sends an email for each when it cancels them.  So the PR author will get 4 emails every time they push to a PR whose tests are in-progress.  But that seems a small price to pay.

abhinav92003 (approved):

Issue #4593, opened 2020-12-08
32-bit client.drwrap-test-detach failed


derekbruening:
https://github.com/DynamoRIO/dynamorio/pull/4592/checks?check_run_id=1515392239
```
2020-12-08T05:41:40.3045321Z 323/348 Test #321: code_api|client.drwrap-test-detach ...............................***Failed  Required regular expression not found. Regex=[^in dr_client_main
2020-12-08T05:41:40.3047360Z client done
2020-12-08T05:41:40.3048072Z app done
2020-12-08T05:41:40.3048637Z $
2020-12-08T05:41:40.3049180Z ]  0.77 sec
2020-12-08T05:41:40.3049843Z in dr_client_main
2020-12-08T05:41:40.3050529Z client done
```

derekbruening:
Failed again:
https://github.com/DynamoRIO/dynamorio/pull/4600/checks?check_run_id=1532203276
```
2020-12-10T17:23:19.3668930Z 324/348 Test #321: code_api|client.drwrap-test-detach ...............................***Failed  Required regular expression not found. Regex=[^in dr_client_main
2020-12-10T17:23:19.3669721Z client done
2020-12-10T17:23:19.3670044Z app done
2020-12-10T17:23:19.3670321Z $
2020-12-10T17:23:19.3670587Z ]  0.93 sec
2020-12-10T17:23:19.3670911Z in dr_client_main
2020-12-10T17:23:19.3671390Z <ERROR: master_signal_handler with no siginfo (i#26?): tid=20838, sig=11>
2020-12-10T17:23:19.3672814Z <Application /home/runner/work/dynamorio/dynamorio/build_debug-internal-32/suite/tests/bin/client.drwrap-test-detach (20837). Cannot correctly handle received signal 11 in thread 20838.>
```

khuey:
https://github.com/DynamoRIO/dynamorio/pull/4533/checks?check_run_id=1534369244

derekbruening:
Happened on GA CI in PR #4624.

PR #4600, opened 2020-12-10
i#1698 ldstex2cas: Add comments on mismatched pair handling

derekbruening:
Adds comments explaining why we're leaving DR's strict equality checks
on the base address and size between the exclusive-load and
exclusive-store.

Issue: #1698

### DIFF
--- a/core/arch/aarchxx/mangle.c
+++ b/core/arch/aarchxx/mangle.c
@@ -3206,6 +3206,14 @@ mangle_exclusive_load(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
             (instr_is_exclusive_load(in) || instr_get_opcode(in) == OP_clrex))
             break;
         if (instr_is_app(in) && instr_is_exclusive_store(in)) {
+            /* Warn on a mismatched pair. */
+            if (opnd_get_size(instr_get_dst(in, 0)) !=
+                opnd_get_size(instr_get_src(instr, 0))) {
+                /* See comment below about CONSTRAINED UNPREDICTABLE. */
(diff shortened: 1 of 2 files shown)


AssadHashmi (approved):

AssadHashmi:
One minor point. Perhaps `Constrained Unpredictable` should be all upper case to match the reference manual.
This emphasises that It's part of a set of well-defined architectural behaviours signified by all being upper case, like `UNDEFINED` and `UNPREDICTABLE`.

derekbruening:
> One minor point. Perhaps `Constrained Unpredictable` should be all upper case to match the reference manual.
> This emphasises that It's part of a set of well-defined architectural behaviours signified by all being upper case, like `UNDEFINED` and `UNPREDICTABLE`.

Done.

derekbruening:
> That said, the PR should attempt if possible to include enough debug/trace for detecting failures due to mismatch so that it's relatively easy to spot from users' reports that that is the cause.

I added a debug build syslog-once for an easily-detected same-block size-mismatch case.  For cross-block cases we'd have to add a special exit or clean call to every stex mangling which I'm not sure I want to put in.  The symptoms should be a hang, which should be relatively easy to spot and diagnose (as opposed to silent data corruption or mysterious crashes thousands of instructions later, anyway).

> Perhaps collect CPU implementer and CPU part from /proc/cpuinfo to identify the implementation? 

We can add that to the debug build up-front logging, and possibly to the crash message: though a hang won't have such a message.  In any case I would do that in a separate PR.

derekbruening:
drwrap-test-detach failed on 32-bit again: #4593

Issue #1921, opened 2016-04-03
add proper signal support while threads are native


derekbruening:
xref #38

DR_EMIT_GO_NATIVE is considered experimental on Linux and one feature missing is:

```
    /* For proper native execution with re-takeover we need to propagate
     * signals to app handlers while native.  For now we do not support re-takeover
     * and we give up our handles via signal_remove_handlers().
     */
```

derekbruening:
I'm considering this issue to also cover signal support for start/stop where all threads are either native or under DR.  Because handling signals for just one thread being native is difficult (we aren't tracking syscalls so we'll have to query for the proper handler and mask) the plan for start/stop whole-process is to remove our handler while the app is native.

derekbruening:
The api.detach_spawn test I'm trying to add for #2601 is revealing a bunch of races, some of which overlap with this issue.

Today there are race windows where we haven't removed our handler completely: e.g., during detach we remove the alarm handlers after the synchall before we let any thread go native, but we leave other signal handlers.

All of this has corner cases with no guarantees.  For alarms we can be reasonably confident that the app's handler will come right back but that may not be true for other signals.  This affects us whether we keep our handler and implement the feature covered here, sending native control to the app's handler, or whether we drop our handler and let the kernel send native control to the app's handler.  The only way to get guarantees is to delay all signals which gets complicated when trying to clean up and detach.

For doing native delivery, the kernel mask should be native, so we should be able to just undo the master_signal_handler_C frame and jump to the app's handler if it exists: if not we'll have to decide what to do.  If the native action is to terminate and we're during detach it'd be best to wait so that any client can run exit time code.

One option is to not let any thread go native until we've cleaned them all up.  Then we remove our handler and then we let them all go at once.  Performance will be worse though with 600 threads stacking up for much longer than before.

hgreving2304:
xref #2941

derekbruening:
I think it's time to close this.  Nearly everything is covered.  There are a few cases that are not, such as too-small sigaltstack for the extra space we take to copy the frame, or default actions: but those are going to exit anyway.  Let's file separate issues on anything that becomes important enough to spend time on; I think we can certainly live with the current shortcomings.

PR #4603, opened 2020-12-10
i#1921 native sig: Add handling for signals for temp-native threads

derekbruening:
Adds a new execute_native_handler() feature where DR delivers a signal
to a currently-native thread.  For a native thread, a signal frame is
set up on the appropriate stack, blocked signals are set natively
rather than for emulation, and control is directly sent to the
application's handler.

Adds handling of asynchronous signals by not recording them for
delivery later but delivering immediately.

Adds a test that uses the client interface DR_EMIT_GO_NATIVE to send a
thread native.  The app then generates a synchronous and an
asynchronous signal.

Initial attempts are made to also handle a pure-native thread with no
dcontex, such as late in detach, but that portion is not fully
finished and will be completed separately.  Handling deault-action
signals is also not fully tested and likely still has some issues to
be worked out.

Fixes several small bugs in the go-native path hit by the test:
+ Moves dynamo_thread_not_under_dynamo() to dispatch_enter_native() to
  avoid problems unlocking bb_building_lock on the path back through
  dispatch.
+ Relaxes a TRY_EXCEPT assert to support a currently-native thread
  with no dcontext in TLS.
+ Suppresses a curiosity from a still-temporarily-native thread at exit.

Issue: #1921

### DIFF
--- a/core/arch/interp.c
+++ b/core/arch/interp.c
@@ -2766,7 +2766,7 @@ client_process_bb(dcontext_t *dcontext, build_bb_t *bb)
         /* we leverage the existing native_exec mechanism */
         dcontext->native_exec_postsyscall = bb->start_pc;
         dcontext->next_tag = BACK_TO_NATIVE_AFTER_SYSCALL;
-        dynamo_thread_not_under_dynamo(dcontext);
+        /* dynamo_thread_not_under_dynamo() will be called in dispatch_enter_native(). */
         return false;
     }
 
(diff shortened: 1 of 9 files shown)


derekbruening:
Huh, works locally.  Looks like compiler complaining about a decl ordering issue or sthg, plus OSX which has a different struct.  Shouldn't affect the review -- will tweak to get green later.

abhinav92003 (approved):

PR #4606, opened 2020-12-12
i#4585 CI autocancel: Add missing auto-cancel to 3 jobs

derekbruening:
PR #4599 failed to add auto-cancel to jobs beyond the first in each
workflow, thus missing 3 jobs.  We correct that here.

Issue: #4585


derekbruening:
Huh, strangely appveyor was never triggered??  Not sure why.

PR #4612, opened 2020-12-15
i#4549 GA CI: Add release package building and uploading

derekbruening:
Adds a new GA workflow triggered via cron Fridays 9pm or manually with
version and build input fields.

Uses 4 parallel jobs to build 5 packages: Linux, AArch64, ARM,
Android, and Windows.  Each job uploads its binaries as artifacts,
allowing sharing between jobs (having each build job racily create and
upload to the same release failed: there is no check for an existing
release in the provided scripts).

Then a separate job waits for the build jobs to finish, creates a
release (which includes adding a tag), downloads the artifacts, and
uploads them to the release.

Adds a step to the Linux package job to deploy the html documentation
to our Github Pages docs site.  Relies on a personal access token stored
as a secret in the dynamorio repository for access to dynamorio_docs.

Renames the TRAVIS_EVENT_TYPE env var used to communicate with our
build scripts to to CI_TARGET=package.  Removes the APPVEYOR_REPO_TAG
env var.

Issue: #4549

### DIFF
--- a/.github/workflows/ci-package.yml
+++ b/.github/workflows/ci-package.yml
@@ -1,0 +1,437 @@
+# **********************************************************
+# Copyright (c) 2020 Google, Inc.  All rights reserved.
+# **********************************************************
+
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# * Redistributions of source code must retain the above copyright notice,
+#   this list of conditions and the following disclaimer.
+#
+# * Redistributions in binary form must reproduce the above copyright notice,
+#   this list of conditions and the following disclaimer in the documentation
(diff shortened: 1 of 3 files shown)


derekbruening:
So I'm leaning toward checking this in to master and only then testing it once it shows up in Actions for the following reasons:

+ to test in a forked repo gets complex with the docs deployment and may not properly test the separate personal access token and permissions for the docs repo
+ the cron build isn't until Fri so we have plenty of time to get it working
+ I give it a 75% chance of "just working" because it is based on the Dr. Memory one which (after many iterations...) works -- could easily be small errors in there though that are hard to catch by humans

abhinav92003 (approved):

PR #4614, opened 2020-12-15
i#4131 Windows CI: Remove Appveyor config file

derekbruening:
Now that we have switched our Windows CI to Github Actions we no
longer need an Appveyor config file.

Issue: #4131


abhinav92003:
Thanks for the migration! GA seems a big improvement over Appveyor -- quick runs and no serialisation.

abhinav92003 (approved):

derekbruening:
Failure is win64 client.drwrap-test-detach which is already on the win32 ignore list -- and it looks like it used to be on the win64 one under #4058 but I didn't see it fail in 8 or so runs in a row on GA so I removed it.  I guess we need to re-add it.  I'll do that separately.

derekbruening:
#4623 adds the promised ignore entry.

PR #4630, opened 2020-12-16
iX add platform-independent macros for call ind

abhinav92003:
Adds `XINST_CREATE_call_ind` macros for ARM, AArch64 and x86. They point to the respective instrs on the platforms for transferring control to address in given register operand.

### DIFF
--- a/core/ir/aarch64/instr_create.h
+++ b/core/ir/aarch64/instr_create.h
@@ -479,8 +479,8 @@ enum {
 /**
  * This platform-independent macro creates an instr_t for an indirect call instr.
  * \param dc  The void * dcontext used to allocate memory for the instr_t.
- * \param t   The opnd_t explicit source operand for the instruction. This should
- * be a reg_id_t with the address of the subroutine.
+ * \param r   The opnd_t explicit source operand for the instruction. This should
+ * be a reg_id_t operand with the address of the subroutine.
  */
 #define XINST_CREATE_call_ind(dc, r) INSTR_CREATE_blr(dc, r)
 
--- a/core/ir/aarch64/instr_create.h
+++ b/core/ir/aarch64/instr_create.h
@@ -476,6 +476,14 @@ enum {
  */
 #define XINST_CREATE_nop(dc) INSTR_CREATE_nop(dc)
 
+/**
+ * This platform-independent macro creates an instr_t for an indirect call instr.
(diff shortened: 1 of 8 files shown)


core/ir/x86/instr_create.h
```diff
@@ -466,6 +466,14 @@
...
  */
 #define XINST_CREATE_nop(dc) INSTR_CREATE_nop(dc)
+
+/**
+ * This platform-independent macro creates an instr_t for an indirect call instr.
+ * \param dc  The void * dcontext used to allocate memory for the instr_t.
+ * \param r   The opnd_t explicit source operand for the instruction. This should
+ * be a reg_id_t operand with the address of the subroutine.
+ */
+#define XINST_CREATE_call_ind(dc, r) INSTR_CREATE_call_ind(dc, r)
```
derekbruening, line 476: Please add a case to each version of test_xinst()

core/ir/aarch64/instr_create.h
```diff
@@ -476,6 +476,14 @@ enum {
  */
 #define XINST_CREATE_nop(dc) INSTR_CREATE_nop(dc)
 
+/**
+ * This platform-independent macro creates an instr_t for an indirect call instr.
+ * \param dc  The void * dcontext used to allocate memory for the instr_t.
+ * \param r   The opnd_t explicit source operand for the instruction. This should
+ * be a reg_id_t operand with the address of the subroutine.
+ */
+#define XINST_CREATE_call_ind(dc, r) INSTR_CREATE_blr(dc, r)
```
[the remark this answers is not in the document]
abhinav92003: Done.

core/ir/x86/instr_create.h
abhinav92003: Oops -- didn't notice those tests. Thanks.

It seems that x86 doesn't verify the exact disassembled instr, unlike AArch64. For now, I've maintained the status quo, but if you'd like me to change that in this PR, let me know.

Also, for ARM, as we don't have an ARM machine, I haven't added the expected test result yet. Added a TODO for that.

derekbruening: > It seems that x86 doesn't verify the exact disassembled instr

Huh, did not know that.  Would be nice to add it but not high priority: if it encodes and decodes then it should match a real opcode encoding and the other IR tests cover anything else.

derekbruening (approved):

PR #4631, opened 2020-12-17
i#4549 GA: Configure failure notifs to dynamorio-devs.

abhinav92003:
Uses the dawidd6/action-send-mail@v2 action to send notifications via email to dynamorio-devs for all failures in CI or package workflows.

This is the final piece in the migration to Github Actions.

Fixes: #4549

### DIFF
--- a/.github/workflows/ci-aarchxx.yml
+++ b/.github/workflows/ci-aarchxx.yml
@@ -99,7 +99,7 @@ jobs:
         to: dynamorio-devs@googlegroups.com
         from: Github Action CI
 
-  # Android ARM cross-compile with gcc, no tests:
+  # Android ARM cross-compile with gcc, no tests
   android-arm-cross-compile:
     runs-on: ubuntu-16.04
 
--- a/.github/workflows/ci-aarchxx.yml
+++ b/.github/workflows/ci-aarchxx.yml
@@ -99,7 +99,7 @@ jobs:
         to: dynamorio-devs@googlegroups.com
         from: Github Action CI
 
-  # Android ARM cross-compile with gcc, no tests
+  # Android ARM cross-compile with gcc, no tests:
   android-arm-cross-compile:
     runs-on: ubuntu-16.04
 
--- a/.github/workflows/ci-aarchxx.yml
+++ b/.github/workflows/ci-aarchxx.yml
@@ -75,6 +75,28 @@ jobs:
       env:
         DYNAMORIO_CROSS_AARCHXX_LINUX_ONLY: yes
 
+    - name: Send failure mail to dynamorio-devs
+      if: failure()
+      uses: dawidd6/action-send-mail@v2
+      with:
+        server_address: smtp.gmail.com
+        server_port: 465
+        username: ${{secrets.MAIL_USERNAME}}
+        password: ${{secrets.MAIL_PASSWORD}}
+        subject: Github Actions CI FAILED on ${{github.repository}} ${{github.ref}}
+        body: |
+          Github Actions CI workflow run FAILED!
+          Workflow: ${{github.workflow}}/aarchxx-cross-compile
+          Repository: ${{github.repository}}
+          Branch ref: ${{github.ref}}
+          SHA: ${{github.sha}}
+          Triggering actor: ${{github.actor}}
+          Triggering event: ${{github.event_name}}
+          Run Id: ${{github.run_id}}
+          See more details on github.com/abhinav92003/dynamorio/actions/runs/${{github.run_id}}
+        to: dynamorio-devs@googlegroups.com
+        from: Github Action CI
+
   # Android ARM cross-compile with gcc, no tests:
   android-arm-cross-compile:
     runs-on: ubuntu-16.04
@@ -112,6 +134,28 @@ jobs:
         DYNAMORIO_ANDROID_TOOLCHAIN: /tmp/android-gcc-arm-ndk-10e
       run: ./suite/runsuite_wrapper.pl automated_ci
 
+    - name: Send failure mail to dynamorio-devs
+      if: failure()
+      uses: dawidd6/action-send-mail@v2
+      with:
+        server_address: smtp.gmail.com
+        server_port: 465
+        username: ${{secrets.MAIL_USERNAME}}
+        password: ${{secrets.MAIL_PASSWORD}}
+        subject: Github Actions CI FAILED on ${{github.repository}} ${{github.ref}}
+        body: |
+          Github Actions CI workflow run FAILED!
+          Workflow: ${{github.workflow}}/android-arm-cross-compile
+          Repository: ${{github.repository}}
+          Branch ref: ${{github.ref}}
+          SHA: ${{github.sha}}
+          Triggering actor: ${{github.actor}}
+          Triggering event: ${{github.event_name}}
+          Run Id: ${{github.run_id}}
+          See more details on github.com/abhinav92003/dynamorio/actions/runs/${{github.run_id}}
+        to: dynamorio-devs@googlegroups.com
+        from: Github Action CI
+
   # AArch64 drdecode and drmemtrace on x86:
   a64-on-x86:
     runs-on: ubuntu-16.04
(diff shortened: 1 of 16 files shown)


abhinav92003:
@derekbruening Will sync offline to add the two new Github secrets (for username and password of the email account) to the DynamoRIO repo settings. 

One caveat of this approach, for PRs from forked repos:
> Secrets are not passed to workflows that are triggered by a pull request from a fork.

I haven't tested behaviour on forked repo PRs yet, but it probably means that we won't get these notification emails for CI failures on those.
But the test suites still run as expected. Note that the Github secret is currently not available in the DynamoRIO repo; but in the above commits, the failed workflows do have complete logs from failed tests, and the successful workflows aren't affected.

.github/workflows/ci-osx.yml
```diff
@@ -78,3 +78,25 @@ jobs:
         # To find available versions, add the following as a step above:
         #  - run: ls -l /Applications
         DEVELOPER_DIR: /Applications/Xcode_11.3.1.app/Contents/Developer
+
+    - name: Send failure mail to dynamorio-devs
```
abhinav92003, line 82: I don't really like this duplication... but even if we create our own custom action with the common stuff, we'll remove only 7/21 of these lines (will need to keep the ones that access env vars, and the ones required to invoke the new action).

derekbruening: We already have a lot of duplication among our 6 workflow files and it's sort of a pain to add something common b/c you have to edit 6 files...but OTOH hopefully we'll not have to touch them again for years and it's prob not worth a lot of effort to share.

derekbruening:
> I haven't tested behaviour on forked repo PRs yet, but it probably means that we won't get these notification emails for CI failures on those.

We want the list emails for pushes to master (not important for PR failures), and I would assume that a forked repo will always be merged by a project member into master and that merge will trigger the email?

.github/workflows/ci-aarchxx.yml
```diff
@@ -75,6 +75,28 @@ jobs:
...
 
+    - name: Send failure mail to dynamorio-devs
+      if: failure()
+      uses: dawidd6/action-send-mail@v2
+      with:
+        server_address: smtp.gmail.com
+        server_port: 465
+        username: ${{secrets.NOTIFICATION_EMAIL_USERNAME}}
+        password: ${{secrets.NOTIFICATION_EMAIL_PASSWORD}}
+        subject: Github Actions CI FAILED on ${{github.repository}} ${{github.ref}}
```
derekbruening, line 86: I would say include `github.event_name` so we know whether it's `pull request` or whatever the string is for a cronjob at a glance.

Maybe include github.workflow in the title too so we can tell at a glance whether it's osx or windows or what?

.github/workflows/ci-aarchxx.yml
```diff
@@ -75,6 +75,28 @@ jobs:
       env:
         DYNAMORIO_CROSS_AARCHXX_LINUX_ONLY: yes
 
+    - name: Send failure mail to dynamorio-devs
+      if: failure()
```
derekbruening, line 79: Hmm, in the past we did have list emails on all PR failures.  I guess we can set up filters if we want to ignore PR failures (a reason to put `event_name` in the subject).  Or should we limit emails to cron jobs and master pushes?

core/arch/aarchxx/emit_utils.c
```diff
@@ -50,7 +50,7 @@ insert_load_dr_tls_base(dcontext_t *dcontext, instrlist_t *ilist, instr_t *where
     /* Load TLS base from user-mode thread pointer/ID register:
      * mrs reg_base, tpidr_el0
      */
-    PRE(ilist, where,
+    PRE(ilist, whre,
```
derekbruening, line 53: This and below are supposed to be temporary do-not-check-in I assume -- I would suggest always having a comment saying that as a safe practice.

abhinav92003:
> We want the list emails for pushes to master (not important for PR failures), and I would assume that a forked repo will always be merged by a project member into master and that merge will trigger the email?

The master merge done by a project member should trigger the email on failure, I believe.

I'm going to try to test our assumptions about forked repo PRs but creating such a PR with i4549-failure-email-notifs as target.

Thanks for adding the GIthub Secrets. Will update the PR when testing is complete and ready for review.

abhinav92003:
Observations from testing for forked repo PRs; see PR #4634 which was forked from abhinav92003/dynamorio and merged into DynamoRIO/dynamorio:i4549-failure-email-notifs for testing:
- Secrets from DynamoRIO/dynamorio are not available to any forked repo PRs. This means that forked repo PRs won't be able to send failure notif emails (see CI logs on PR #4634); unless the forked repo also defines the same Secret variables as DynamoRIO/dynamorio, in which case that other email address will be used. (See other comment thread: This is probably okay as we don't want to send failure notif emails on non-master branches anyway.) We should rename our Secret vars to have a DYNAMORIO_ prefix.

- If CI succeeds, the email notif issue doesn't hinder green-ness of the checks.

- When the PR is merged to a DR branch, it does send failure notif emails (see commit b9344bc above).

.github/workflows/ci-aarchxx.yml
```diff
@@ -75,6 +75,30 @@ jobs:
       env:
         DYNAMORIO_CROSS_AARCHXX_LINUX_ONLY: yes
 
+    - name: Send failure mail to dynamorio-devs
+      if: failure() && github.ref == 'refs/heads/master'
+      uses: dawidd6/action-send-mail@v2
+      with:
+        server_address: smtp.gmail.com
+        server_port: 465
+        username: ${{secrets.DYNAMORIO_NOTIFICATION_EMAIL_USERNAME}}
```
abhinav92003, line 84: @derekbruening Can you please rename the Secret vars? It is better to have a DYNAMORIO_ prefix on these variables to reduce chances of conflict with a forked repo's Secret vars. See comment on  main PR thread for details.

core/arch/aarchxx/emit_utils.c
abhinav92003: Will do in future. Thanks.

abhinav92003: Reverted these temporary changes.

.github/workflows/ci-aarchxx.yml
abhinav92003: I've added a condition to send email only for failures on master branch (I assume the cron jobs should also satisfy that condition as they are also on the  master branch). That seems better than folks having to add email filters. Failures on master branch will probably be what others are mostly interested in. And the author of the PR gets emails directly from Github for CI triggered by them.

.github/workflows/ci-aarchxx.yml
abhinav92003: Done.

abhinav92003:
Some other observations:
- Commit SHA for the synchronize event (which we use for triggering CI on PR branch pushes) is different from the hash of the commit that triggered the synchronize event.
E.g., following is an email to dynamorio-devs. Note that the SHA doesn't match any commit on PR #4631.
```
Github Actions CI workflow run FAILED!
Workflow: ci-aarchxx/a64-on-x86
Repository: DynamoRIO/dynamorio
Branch ref: refs/pull/4631/merge
SHA: dbb985c6121274ae0e30bda2fc80c0afea062d18
Triggering actor: abhinav92003
Triggering event: pull_request
Run Id: 431475676
See more details on github.com/DynamoRIO/dynamorio/actions/runs/431475676
```

- clang-format doesn't seem to fail after the PR is pushed to master, even though it did fail on the PR. Found this while testing on my forked repo; note the failure on https://github.com/abhinav92003/dynamorio/pull/1, but not on the corresponding merge commit. @derekbruening this is probably because clang-format looks only at the changed lines wrt base branch?

- The automatic emails sent by Github Actions (the ones titled "[DynamoRIO/dynamorio] PR run failed: ci-xxx") has dynamorio@noreply.github.com as its recipient. But I don't see such emails for any PR other than my own (@derekbruening can you please confirm?) . So, there will be duplication of failure  notif email for the PR author.

.github/workflows/ci-aarchxx.yml
derekbruening: They can't be renamed.  I created new ones with the new names and deleted the old ones.

derekbruening:
> * clang-format doesn't seem to fail after the PR is pushed to master, even though it did fail on the PR. Found this while testing on my forked repo; note the failure on [abhinav92003#1](https://github.com/abhinav92003/dynamorio/pull/1), but not on the corresponding merge commit. @derekbruening this is probably because clang-format looks only at the changed lines wrt base branch?

Sounds right -- yes, it does a "git diff -U0 origin/master".

> * The automatic emails sent by Github Actions to individuals (the ones titled "[DynamoRIO/dynamorio] PR run failed: ci-xxx") has [dynamorio@noreply.github.com](mailto:dynamorio@noreply.github.com) as its recipient. But I don't see such emails for any PR other than my own (@derekbruening can you please confirm?) . So, there will be duplication of failure notif email for the PR author if they are subscribed to dynamorio-devs too.

My "PR run failed" and "Run failed" (manually triggered) emails are `from: Derek Bruening <notifications@github.com>`, to the same recipient you cite, and cc-ed to `ci_activity@noreply.github.com`.  Like you I have not received any from other people's PR's.  Duplication seems fine: we're talking about merges to master, not PR failures, so they should be relatively rare.

derekbruening (approved):

abhinav92003:
Verified using a simple cron config that it also has `github.ref == 'refs/heads/master'`.

abhinav92003:
run arm tests

PR #4635, opened 2020-12-21
i#4549 GA CI: Fix cronbuild version errors and allow manual

derekbruening:
The cronbuild version was missing the major and minor: that was on the
tag only.  We fix that here.

Adds a source checkout to the release job, needed to run "git log"
to compute the version.

Allows manual cronbuilds via a blank version, so we can actually test
them on demand, and create them separately from the scheduled time.

Applies input build numbers to cronbuilds too so we can manually test
them the same day as a scheduled build.

Moves the scheduled cronbuild to 2am UTC for 9pm EST.

Issue: #4549

### DIFF
--- a/.github/workflows/ci-package.yml
+++ b/.github/workflows/ci-package.yml
@@ -32,18 +32,16 @@
 
 name: ci-package
 on:
-  # Our weekly cronbuild: 9pm on Fridays.
-  # Presumably this is only triggered by this .yml file on the master branch.
+  # Our weekly cronbuild: 9pm EST on Fridays.
   schedule:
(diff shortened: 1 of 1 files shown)


derekbruening:
Making the input version required was a mistake: it wouldn't allow it to be empty and there was thus no way to manually test or create a cronbuild (as opposed to a release build).  Here we fix that, along with fixing two bugs that prevented the scheduled cronbuild from working.  Unfortunately we have to merge to master to change the required-non-empty, but then we can do a manual test of that path in PR's and fix any other issues there.

abhinav92003 (approved):

PR #4636, opened 2020-12-23
i#1921 native sig: Deliver signals to unmanaged threads

derekbruening:
When a signal arrives in an completely unmanaged thread with no
dcontext, typically because DR is detaching, we now deliver the signal
if the application has a handler for it.  This requires adding support
for no dcontext to several parts of the frame setup code even beyond
what was added in PR #4603 for temporarily-native threads.

We have to save the app's handler when we detach a thread so we know
where to send a native signal.  Full support is complex when we're
cleaning up and have no dynamic storage, so we use a single global
handler per signal.  We detect whether multiple handlers are in
operation in this single DR instance (quite rare: only custom
non-pthread clones have this behavior) and in that case we abort like
before on a native signal.  Adds ATOMIC_READ_1BYTE() to complement the
existing atomic operations for a cleaner read of the new multi-handler
flag.

Delivering the frame often overlaps with DR's frame and even DR's
stack usage while delivering, if the app has no sigaltstack.  We add
logic to detect this overlap and avoid clobbering the stack memory.

Alarm signals are still dropped, since they can arrive mid-thread-init
when it is even harder to deliver.

Adds a new test api.detach_signal which creates 10 threads who all sit
in a loop sending 4 different alternating signals (SIGSEGV, SIGBUS,
SIGURG, SIGALRM) while the main thread attaches and then detaches.
When run in debug build, many many signals arrive in post-detach
threads, since detach takes a while to do debug cleanup, exercising
the new code.

Adds a new RSTAT for native signals so we can identify when this
happens in release build.  Exports the stat to the API and uses it to
check that at least some signals were delivered natively in the new
test.

Removes the fatal error on a signal arriving with no dcontext.  But,
non-ignore default signal actions when no handler is present are still
not fully handled, along with multi-sighand-processes as mentioned,
and the fatal error remains in those cases.  For default actions,
since the process is going to terminate anyway, the only shortcoming
of this is whether a core is generated and whether the proper process
exit code is raised.

Issue: #1921


derekbruening:
Test failures are minor, of two types: missing UNIX ifdefs for Windows; and an assert in the atomic macros that should look at the target.

core/arch/arch_exports.h
```diff
@@ -327,9 +327,13 @@ emit_detach_callback_final_jmp(dcontext_t *dcontext,
 /* note that the microsoft compiler will not enregister variables across asm
  * blocks that touch those registers, so don't need to worry about clobbering
  * eax and ebx */
+#    define ATOMIC_1BYTE_READ(addr_src, addr_res)               \
+        do {                                                    \
+            *(BYTE *)(addr_res) = *(volatile BYTE *)(addr_src); \
```
abhinav92003, line 332: Why are there two strategies of atomic reads: the asm block and the volatile pointer deref?

core/utils.h
```diff
@@ -1223,7 +1226,7 @@ bitmap_check_consistency(bitmap_t b, uint bitmap_size, uint expect_free);
 #    endif /* INTERNAL */
 #    define THREAD          \
         ((dcontext == NULL) \
-             ? INVALID_FILE \
```
abhinav92003, line 1226: Are we concerned about any case where dcontext isn't present but we wrongly assume that it's because DR is detaching?

abhinav92003 (approved):

core/utils.h
derekbruening: It could be a `native_exec_list` library who created a new thread or something, so it does not have to be a detach (though in today's most common uses detach is most frequent) but it seems to me that printing extra log statements from any native thread into the global log file is always better than having log statements dropped on the ground: these are verbose messages for diagnostics after all.

core/arch/arch_exports.h
derekbruening: Oh, the separate Windows and Linux ones for x86?  The Visual Studio 64-bit compiler does not support *any* inline assembly is the reason.  I suppose we could have the C code for Linux too?  I was just following the existing split.  Many of the more complex operations can't be written in C so only a few could be shared like that (they use special intrinsics for Visual Studio).  I guess I would lean toward just leaving it unless you feel strongly?

abhinav92003: I was just curious about the difference. We can continue following the existing split.

core/utils.h
abhinav92003: Sure, makes sense.

Issue #4504, opened 2020-10-30
Implement instr_invert_predicate for AArch64


yury-khrustalev:
Arm-32-only API function `instr_invert_predicate` needs to be implemented for AArch64 to support working with conditional operands.

Issue #4000, opened 2019-12-31
doxgyen failing to install on Appveyor


derekbruening:
We have had many problems installing chocolately packages on Appveyor, causing us to downgrade installation failures to just warnings we continue past.  We noticed recently that doxygen has been failing to install every time:

https://ci.appveyor.com/project/DynamoRIO/dynamorio/builds/29690836?fullLog=true

```
choco install --timeout 600 -y doxygen.portable imagemagick.tool ghostscript || true
Chocolatey v0.10.8
Installing the following packages:
doxygen.portable;imagemagick.tool;ghostscript
By installing you accept licenses for the packages.
Progress: Downloading doxygen.portable 1.8.11... 100%

doxygen.portable v1.8.11 [Approved] - Likely broken for FOSS users (due to download location changes)
doxygen.portable package files install completed. Performing other installation steps.
Attempt to get headers for http://ftp.stack.nl/pub/users/dimitri/doxygen-1.8.11.windows.x64.bin.zip failed.
  The remote file either doesn't exist, is unauthorized, or is forbidden for url 'http://ftp.stack.nl/pub/users/dimitri/doxygen-1.8.11.windows.x64.bin.zip'. Exception calling "GetResponse" with "0" argument(s): "The remote name could not be resolved: 'ftp.stack.nl'"
Downloading doxygen.portable 64 bit
  from 'http://ftp.stack.nl/pub/users/dimitri/doxygen-1.8.11.windows.x64.bin.zip'
ERROR: The remote file either doesn't exist, is unauthorized, or is forbidden for url 'http://ftp.stack.nl/pub/users/dimitri/doxygen-1.8.11.windows.x64.bin.zip'. Exception calling "GetResponse" with "0" argument(s): "The remote name could not be resolved: 'ftp.stack.nl'" 
This package is likely not broken for licensed users - see https://chocolatey.org/docs/features-private-cdn.
The install of doxygen.portable was NOT successful.
Error while running 'C:\ProgramData\chocolatey\lib\doxygen.portable\tools\chocolateyInstall.ps1'.
 See log for details.
```

We could try the `doxygen` package instead of `doxygen.portable`.

derekbruening:
Xref #2616

derekbruening:
This is blocking #2861, #3194, https://github.com/DynamoRIO/drmemory/issues/2255 where we need the docs for package building.  We may also have to solve the ghostscript and fig2dev problems.

derekbruening:
[...]

PR #4637, opened 2020-12-29
i#4504: Implement instr_invert_predicate for AArch64

yury-khrustalev:
Add implementation of `instr_invert_predicate` for AArch64.
This change is required to complete #4500 to allow creation of conditional instructions.

Fixes: #4504


yury-khrustalev:
VS builds failed with "The remote server returned an error: (404) Not Found." error, could you please restart the builds, thank you!

derekbruening:
> VS builds failed with "The remote server returned an error: (404) Not Found." error, could you please restart the builds, thank you!

Please see https://docs.github.com/en/free-pro-team@latest/actions/managing-workflow-runs/re-running-a-workflow

yury-khrustalev:
> Please see https://docs.github.com/en/free-pro-team@latest/actions/managing-workflow-runs/re-running-a-workflow

@derekbruening, thank you, I know how to use GitHub actions, but I don't have "Re-run all jobs" button in this repository (probably due to not having permissions or something along these lines).

yury-khrustalev:
Also, it looks like there is a problem with URL used to download Doxygen for windows builds.

derekbruening:
> Also, it looks like there is a problem with URL used to download Doxygen for windows builds.

If the doxygen URL changed, then a re-run won't fix it.  If you have the new URL, submit a separate PR to fix it?

yury-khrustalev:
The URL which is no longer available is

    http://doxygen.nl/files/doxygen-1.8.19.windows.x64.bin.zip

Probably version 1.8.19 is considered too old. The next version 1.8.20 is [available](https://doxygen.nl/files/doxygen-1.8.20.windows.x64.bin.zip), though. Or the most recent version 1.9.0 can be used instead (I'm not sure why 1.8.19 version is currently used in particular):

    https://doxygen.nl/files/doxygen-1.9.0.windows.x64.bin.zip

>  If you have the new URL, submit a separate PR to fix it?

If I submitted a new PR, the builds would obviously fail and the PR wouldn't be able to merge. I think it's easier for a maintainer to fix the URL directly.

yury-khrustalev:
>  If you have the new URL, submit a separate PR to fix it?

Was there a reason why version 1.8.19 is used? Would it not work with 1.9.0? I simply don't have appropriate devices to test.

derekbruening:
> > If you have the new URL, submit a separate PR to fix it?
> 
> Was there a reason why version 1.8.19 is used? Would it not work with 1.9.0?

I recall using chocolatey to download the latest w/o having to specify its version but chocolately was flaky: looks like #4000.  So we switched to a hardcoded version to solve the choco problems and probably just put in the latest version at the time, which then suffers from breaking when the link disappears.  Suggestions for a better way are welcome.

> I simply don't have appropriate devices to test.

I think anyone would test it the same way as you: make a PR and see if Windows turns green on the PR.  Or wait and someone else will hit it and have to fix it to make progress.

yury-khrustalev:
@derekbruening, could you please review this when you have time. This comes from the discussion in #4500 and is required to rework that change according to what we decided. Thank you.

core/ir/instr.h
```diff
@@ -1351,17 +1351,22 @@ DR_API
...
 
+#ifdef AARCHXX
 DR_API
 /**
  * Returns the DR_PRED_ constant that represents the opposite condition
  * from \p pred.  A valid conditional branch predicate must be passed (i.e.,
- * not #DR_PRED_NONE, DR_PRED_AL, or DR_PRED_OP).
- * \note ARM-only.
+ * not #DR_PRED_NONE, DR_PRED_AL, or DR_PRED_OP for ARM and not #DR_PRED_NONE,
+ * DR_PRED_AL, or DR_PRED_NV for AArch64).
```
derekbruening, line 1362: nit: missing doxygen links for these two (the `#`)
Looks like the original ARM is also missing them for 2 entries: if you could add those too.

derekbruening (approved):
(I see there's no existing test for the ARM version but it could be nice to add some sanity check to ir_aarch64.c or sthg.)

derekbruening:
> Fixes: #4504 by adding implementation of instr_invert_predicate for AArch64.

The Fixes has to be on a line by itself.  See https://github.com/DynamoRIO/dynamorio/wiki/Code-Reviews#commit-messages where we always put it at the bottom.

core/ir/instr.h
yury-khrustalev: This is fixed now.

Issue #1911, opened 2016-03-24
ARM machine_cache_sync is not thread-safe


egrimley:
According to sections B2.6.5 (AArch64) and E2.6.5 (AArch32) of the ARM ARM, when one CPU writes instructions that will be executed by another CPU, a number of operations must be performed on the first CPU (these are implemented by machine_cache_sync), and the second CPU must execute an ISB instruction when the modified instructions are observable and before it executes them. There does not appear to be any mechanism in place to ensure that the second CPU executes an ISB.

I do not know how likely it is that this non-conformance to the architecture specification will cause problems in practice. It is perhaps very unlikely. Also, I cannot see any efficient and easy way of forcing an ISB onto every CPU/thread.

derekbruening:
Remember to use the labels

derekbruening:
We've had some discussions about this elsewhere: some private email threads, and https://github.com/DynamoRIO/dynamorio/pull/4638/files#r549664851, and possibly elsewhere.

Trying to summarize some of the current thoughts/concerns:

While the manual guarantees we can modify certain opcodes (B , BL , BRK , HVC , ISB , NOP , SMC , and SVC: see B2.2.5) with all threads seeing either the original or the modified, it makes no statement about the modified being seen in any guaranteed time limit.  This seems to imply that we actually cannot modify any instructions and have other threads see the change promptly without forcing all threads to execute an ISB.

That then implies that all direct branch linking (both of the exit cti and in the stub), changing the jumps around an inlined syscall, etc. are all unsafe.

For direct linking, we could probably recoup the performance loss of always using indirect transfers by implementing -enable_traces.

How would we solve inlined syscalls?  A conditional branch based on data we change?

If all known current hardware does in fact provide a timeliness guarantee, or in practice OS interrupts provide them as @egrimley wondered, we could live with theoretically unsound code.  Perhaps we could try to request that the specification be changed.

derekbruening:
> unsafe.

Clarifying here: for plain direct links (and not inlined syscalls where there is clearly a problem) the problem is subsequent code thinking the fragment is linked or unlinked and taking the wrong action based on that; no time limits anymore on "unlink flushing"; etc.  Maybe if we didn't use unlink flushing things might work out in some cases, and we might have to run the code with simulated infinite time to see the new code version to really know all the problems, but my intuition is that multiple things beyond unlink flushing would break.

derekbruening:
[...]

PR #4638, opened 2020-12-29
i#4273: Implement far fragment linking on AArch64.

abhinav92003:
Adds support for linking fragments arbitrarily far away on AArch64. Uses indirect branch to a stored address, and changes the stored address between the fcache-return routine address and the linked fragment address.

Modifies the fragment prefix to restore both x0 and x1. Also stores both register values in respective slots in flows that use the fragment prefix.

Detailed design: https://github.com/DynamoRIO/dynamorio/wiki/Linking-Far-Fragments-on-AArch64; Option 4b is implemented here. It increases stub size by 12B for the data-slot (extra 4B for alignment).

Fixes: #4273


core/arch/aarch64/emit_utils.c
```diff
@@ -124,11 +136,28 @@ insert_exit_stub_other_flags(dcontext_t *dcontext, fragment_t *f, linkstub_t *l,
...
+         * lots of places expect the stub size to be fixed.
+         */
+        for (uint j = 0; j < num_nops_needed; j++)
+            *pc++ = NOP_INST;
+        /* The final slot is a data slot, which will hold the address of either
+         * the fcache-return routine or the linked fragment.
+         */
+        ASSERT(sizeof(app_pc) == 8);
+        pc += DIRECT_EXIT_STUB_DATA_SZ / sizeof(uint);
+        /* We start off with the fcache-return routine address in the slot. */
```
abhinav92003, line 155: This works under the assumption that the `fcache-return` routine is shared between all threads. @derekbruening Would you consider that okay to assume? If not, the alternative is Option 4a at https://github.com/DynamoRIO/dynamorio/wiki/Linking-Far-Fragments-on-AArch64.

core/arch/aarch64/emit_utils.c
```diff
@@ -181,36 +207,22 @@ exit_cti_reaches_target(dcontext_t *dcontext, fragment_t *f, linkstub_t *l,
 void
 patch_stub(fragment_t *f, cache_pc stub_pc, cache_pc target_pc, bool hot_patch)
 {
-    /* Compute offset as unsigned, modulo arithmetic. */
-    ptr_uint_t off = (ptr_uint_t)target_pc - (ptr_uint_t)stub_pc;
-    if (off + 0x8000000 < 0x10000000) {
```
abhinav92003, line 186: @derekbruening For now, I've removed the direct branch patching, as that is not within the architecture specifications. The alternative is to keep it and accept that it isn't strictly okay.

As discussed, I can work on gathering some numbers to help us understand how important this is.

derekbruening: You mean, patching a CBZ, CBNZ, TBZ, or TBNZ?  Patching a B or BL is still explicitly on the opcode mod list.  Probably B is the most common in any case?

derekbruening: (BL presumably turns into B after mangling)

abhinav92003: I mean that I've removed the `STP <-> B` mangling that's done by `patch_stub` for near fragment linking. 

Per the manual, both opcodes _before_ and _after_ need to be part of the opcode mod list, of which `STP` is not a part.

We could do `NOP <-> B` by adding an additional `NOP` as the first stub instruction, but the only guarantee we get is that the executed instruction will be _either one of them, not the latest one_. 

To guarantee that only the latest instruction is executed, we'd still need a sync, which also includes all other threads executing an `ISB` after [`clear_icache`](https://github.com/DynamoRIO/dynamorio/blob/57784023fc35b5aad7108c444ab79fc820276543/core/drlibc/drlibc.c#L106); I don't think we have any way to do that right now (#1911).

derekbruening: So you mean not the exit cti direct branch patching, but the intermediate patching that went through the stub.  But the exit cti patching could involve a CBZ.  And does this mean that the intermediate is all gone and now there's nothing but exit cti or indirect branch?

core/arch/aarch64/emit_utils.c
derekbruening: > This works under the assumption that the fcache-return routine is shared between all threads. @derekbruening Would you consider that okay to assume?

In general the "gencode" comes in both thread-private and thread-shared varieties.  But any one fragment, and its stubs, will only ever use one or the other, depending on whether the fragment is private or shared.  (Plus, IIRC 64-bit for both x86 and aarch doesn't have private gencode at all anyway?)

core/arch/aarch64/emit_utils.c
derekbruening: Maybe file (or add to an issue) sthg about the inlined syscall skipping.  If we have no good solution there, we might decide to live with a lack of theoretical guarantee if no known hardware is like that, and perhaps we could try to get ARM to change the spec.

derekbruening: I put some comments in #1911 about the inlined skipping and these other larger concerns.

core/arch/aarch64/emit_utils.c
```diff
@@ -102,6 +102,18 @@ insert_mov_imm(uint *pc, reg_id_t dst, ptr_int_t val)
     return pc;
 }
 
+/* Returns addr for the target_pc data slot of the given stub. The slot starts at the
+ * 8-byte aligned region in the 12-byte slot reserved in the stub.
+ */
+cache_pc
```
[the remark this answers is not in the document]
abhinav92003: Done.

core/arch/aarch64/emit_utils.c
```diff
@@ -124,11 +136,28 @@ insert_exit_stub_other_flags(dcontext_t *dcontext, fragment_t *f, linkstub_t *l,
         /* mov x0, ... */
         pc = insert_mov_imm(pc, DR_REG_X0, (ptr_int_t)l);
         num_nops_needed = 4 - (pc - write_stub_pc - 1);
-        /* ldr x1, [x(stolen), #(offs)] */
-        *pc++ = (0xf9400000 | 1 | (dr_reg_stolen - DR_REG_X0) << 5 |
-                 get_fcache_return_tls_offs(dcontext, f->flags) >> 3 << 10);
+        uint target_pc_slot_offs = (uint *)get_target_pc_slot(f, stub_pc) - pc;
+        /* ldr x1, [pc, target_pc_slot_offs * AARCH64_INSTR_SIZE] */
+        *pc++ = (0x58000000 | (DR_REG_X1 - DR_REG_X0) | target_pc_slot_offs << 5);
```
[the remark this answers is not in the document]
abhinav92003: Will do for the final commit message. It's there in the PR's description already.

core/arch/aarch64/emit_utils.c
```diff
@@ -102,6 +102,18 @@ insert_mov_imm(uint *pc, reg_id_t dst, ptr_int_t val)
     return pc;
 }
 
+/* Returns addr for the target_pc data slot of the given stub. The slot starts at the
```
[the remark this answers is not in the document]
abhinav92003: Done.

core/arch/aarch64/emit_utils.c
```diff
@@ -144,17 +173,14 @@ insert_exit_stub_other_flags(dcontext_t *dcontext, fragment_t *f, linkstub_t *l,
                  get_ibl_entry_tls_offs(dcontext, exit_target) >> 3 << 10);
         /* br x1 */
         *pc++ = BR_X1_INST;
+        /* Fill up with NOPs, depending on how many instructions we needed to move
+         * the immediate in a registers. Ideally we would skip adding NOPs, but
```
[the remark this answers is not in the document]
abhinav92003: Done.

core/arch/aarch64/emit_utils.c
```diff
@@ -744,8 +757,11 @@ emit_indirect_branch_lookup(dcontext_t *dc, generated_code_t *code, byte *pc,
         INSTR_CREATE_ldp(dc, opnd_create_reg(DR_REG_X0), opnd_create_reg(DR_REG_X2),
                          opnd_create_base_disp(dr_reg_stolen, DR_REG_NULL, 0,
                                                TLS_REG0_SLOT, OPSZ_16)));
-    /* Store x0 in TLS_REG1_SLOT as requied in the fragment prefix. */
-    APP(&ilist, instr_create_save_to_tls(dc, DR_REG_R0, TLS_REG1_SLOT));
+    /* Store x0, x1 in TLS_REG0_SLOT,TLS_REG1_SLOT as required by the fragment
+     * prefix.
+     */
+    APP(&ilist, instr_create_save_to_tls(dc, DR_REG_X0, TLS_REG0_SLOT));
+    APP(&ilist, instr_create_save_to_tls(dc, DR_REG_X2, TLS_REG1_SLOT));
```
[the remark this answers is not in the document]
abhinav92003: I realised that this isn't needed at all actually. Removed this and some other restores that will be performed by the fragment prefix later.

core/arch/arch.h
```diff
@@ -723,7 +723,7 @@ mangle_writes_thread_register(dcontext_t *dcontext, instrlist_t *ilist, instr_t
 /* offsets within local_state_t used for specific scratch purposes */
 enum {
     /* ok for this guy to overlap w/ others since he is pre-cache */
-    FCACHE_ENTER_TARGET_SLOT = TLS_REG0_SLOT,
+    FCACHE_ENTER_TARGET_SLOT = TLS_REG2_SLOT,
```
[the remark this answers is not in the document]
abhinav92003: I think the comment says that other slots can use the one that `FCACHE_ENTER_TARGET_SLOT` does.

So instead of replacing, I just added that `FCACHE_ENTER_TARGET_SLOT` cannot use reg0 and reg1 slots.

core/arch/aarch64/emit_utils.c
```diff
@@ -181,36 +207,22 @@ exit_cti_reaches_target(dcontext_t *dcontext, fragment_t *f, linkstub_t *l,
...
-        *(uint *)vmcode_get_writable_addr(stub_pc) =
-            (0x14000000 | (0x03ffffff & off >> 2));
-        if (hot_patch)
-            machine_cache_sync(stub_pc, stub_pc + 4, true);
-        return;
-    }
-    /* We must use an indirect branch. */
-    ASSERT_NOT_IMPLEMENTED(false); /* FIXME i#1569 */
+    byte *target_pc_slot = get_target_pc_slot(f, stub_pc);
+    *(uint64 *)target_pc_slot = (ptr_uint_t)target_pc;
```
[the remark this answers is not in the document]
abhinav92003: Done.

core/arch/aarch64/emit_utils.c
```diff
@@ -181,36 +207,22 @@ exit_cti_reaches_target(dcontext_t *dcontext, fragment_t *f, linkstub_t *l,
...
 }
 
 bool
-stub_is_patched(fragment_t *f, cache_pc stub_pc)
+stub_is_patched(dcontext_t *dcontext, fragment_t *f, cache_pc stub_pc)
 {
-    ASSERT_NOT_IMPLEMENTED(false); /* FIXME i#1569 */
-    return false;
+    uint *addr = (uint *)get_target_pc_slot(f, stub_pc);
```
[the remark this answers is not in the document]
abhinav92003: Done.

core/arch/arch_exports.h
```diff
@@ -1880,8 +1880,12 @@ fill_with_nops(dr_isa_mode_t isa_mode, byte *addr, size_t size);
 #        define AARCH64_INSTR_SIZE 4
 #        define FRAGMENT_BASE_PREFIX_SIZE(flags) AARCH64_INSTR_SIZE
 #        define DIRECT_EXIT_STUB_SIZE(flags) \
-            (7 * AARCH64_INSTR_SIZE) /* see insert_exit_stub_other_flags */
-#        define DIRECT_EXIT_STUB_DATA_SZ 0
+            (10 * AARCH64_INSTR_SIZE) /* see insert_exit_stub_other_flags */
+/* Size of data slot used to store address of linked fragment or fcache return routine.
+ * We reserve 12 bytes for the 8 byte address, so that we can store it in an 8-byte
+ * aligned address. This is required for atomicity of write operations.
```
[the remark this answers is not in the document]
abhinav92003: Yeah, we can also save another 4x bytes when the linkstub move takes less than 4 movs (currently we fill the remaining with nop). But the comment there notes that lots of places expect the stub size to be fixed.

core/arch/emit_utils_shared.c
```diff
@@ -464,7 +466,7 @@ link_direct_exit(dcontext_t *dcontext, fragment_t *f, linkstub_t *l, fragment_t
          * this stub-requiring scheme.
```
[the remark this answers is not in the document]
abhinav92003: Done.

core/arch/emit_utils_shared.c
```diff
@@ -2148,15 +2150,18 @@ emit_fcache_enter_common(dcontext_t *dcontext, generated_code_t *code, byte *pc,
...
+    /* Put app's X0,X1 in TLS_REG0_SLOT,TLS_REG1_SLOT: */
+    /* ldp x0, x1, [x5] */
     APP(&ilist,
-        XINST_CREATE_load(dcontext, opnd_create_reg(DR_REG_X0),
-                          OPND_CREATE_MEMPTR(DR_REG_X5, 0)));
+        XINST_CREATE_load_pair(
+            dcontext, opnd_create_reg(DR_REG_X0), opnd_create_reg(DR_REG_X1),
+            opnd_create_base_disp(DR_REG_X5, DR_REG_NULL, 0, 0, OPSZ_16)));
+
     /* str x0, [x28, #8] */
```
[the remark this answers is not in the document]
abhinav92003: Done.

core/arch/emit_utils_shared.c
```diff
@@ -4791,10 +4796,10 @@ emit_do_syscall_common(dcontext_t *dcontext, generated_code_t *code, byte *pc,
 
 #ifdef AARCH64
     /* We will call this from handle_system_call, so need prefix on AArch64. */
-    APP(&ilist,
-        instr_create_restore_from_tls(dcontext, ENTRY_PC_REG, ENTRY_PC_SPILL_SLOT));
+    APP(&ilist, instr_create_restore_from_tls(dcontext, DR_REG_X0, TLS_REG0_SLOT));
+    APP(&ilist, instr_create_restore_from_tls(dcontext, DR_REG_X1, TLS_REG1_SLOT));
```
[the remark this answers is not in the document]
abhinav92003: Done.

core/arch/emit_utils_shared.c
```diff
@@ -4921,7 +4926,7 @@ emit_fcache_enter_gonative(dcontext_t *dcontext, generated_code_t *code, byte *p
     /* get target PC */
     APP(&ilist,
         XINST_CREATE_load(dcontext, opnd_create_reg(DR_REG_R0),
-                          OPND_CREATE_MEMPTR(dr_reg_stolen, 0)));
+                          OPND_CREATE_MEMPTR(dr_reg_stolen, 16)));
```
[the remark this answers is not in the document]
abhinav92003: This is required because the next pc was stored in `FCACHE_ENTER_TARGET_SLOT` (`TLS_REG2_SLOT`) by `append_setup_fcache_target`.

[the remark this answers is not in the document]
abhinav92003: Added a comment for this.

core/ir/disassemble_shared.c
```diff
@@ -1409,10 +1409,11 @@ common_disassemble_fragment(dcontext_t *dcontext, fragment_t *f_in, file_t outfi
             pc = (cache_pc)disassemble_with_bytes(dcontext, (byte *)pc, outfile);
         }
         if (LINKSTUB_DIRECT(l->flags) && DIRECT_EXIT_STUB_DATA_SZ > 0) {
-            ASSERT(DIRECT_EXIT_STUB_DATA_SZ == sizeof(cache_pc));
-            if (stub_is_patched(f, EXIT_STUB_PC(dcontext, f, l))) {
+            ASSERT(DIRECT_EXIT_STUB_DATA_SZ == sizeof(cache_pc) IF_AARCH64(+4));
```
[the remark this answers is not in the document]
abhinav92003: Done.

core/arch/aarch64/emit_utils.c
```diff
@@ -181,36 +207,22 @@ exit_cti_reaches_target(dcontext_t *dcontext, fragment_t *f, linkstub_t *l,
...
+unpatch_stub(dcontext_t *dcontext, fragment_t *f, cache_pc stub_pc, bool hot_patch)
 {
-    /* Restore the stp x0, x1, [x28] instruction. */
-    *(uint *)vmcode_get_writable_addr(stub_pc) =
-        (0xa9000000 | 0 | 1 << 10 | (dr_reg_stolen - DR_REG_X0) << 5 |
-         TLS_REG0_SLOT >> 3 << 15);
-    if (hot_patch)
-        machine_cache_sync(stub_pc, stub_pc + AARCH64_INSTR_SIZE, true);
+    byte *target_pc_slot = get_target_pc_slot(f, stub_pc);
+    *(uint64 *)target_pc_slot = (ptr_uint_t)fcache_return_routine(dcontext);
```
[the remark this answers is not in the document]
abhinav92003: Done.

core/arch/aarch64/emit_utils.c
```diff
@@ -184,33 +273,63 @@ patch_stub(fragment_t *f, cache_pc stub_pc, cache_pc target_pc, bool hot_patch)
...
+
+    ptr_uint_t target_pc;
+    ATOMIC_8BYTE_ALIGNED_READ(get_target_pc_slot(f, stub_pc), &target_pc);
+    /* linked to far fragment. */
+    return target_pc != (ptr_uint_t)fcache_return_routine(dcontext);
 }
 
 void
-unpatch_stub(fragment_t *f, cache_pc stub_pc, bool hot_patch)
+unpatch_stub(dcontext_t *dcontext, fragment_t *f, cache_pc stub_pc, bool hot_patch)
```
abhinav92003, line 316: This implementation consists of two operations to unpatch, but at any given time only one patching strategy would be used. So, one of these two operations would really be a no-op.

abhinav92003:
I'll try to run these changes on a large application to verify if the far fragment linking strategy works.

The fact that commit c07b6db (the one where only the far fragment linking strategy was enabled) has all tests green makes me believe that it does work. But I guess more testing is always good so I'll do it anyway.

core/arch/aarch64/emit_utils.c
abhinav92003: Yes AArch64 has shared fragments (hence all the consistency related issues too). I added an ASSERT as discussed, to verify that the `fcache-return` routine from thread dcontext is same as the one from `GLOBAL_DCONTEXT`. This seems a strong assumption for this design, so I think it'll be good to add asserts.

core/arch/aarch64/emit_utils.c
abhinav92003: The inlined syscall skip patching is relevant only to the thread that's receiving the signal. So we only need to worry about making the patched code visible to the current processor element.

The current implementation calls machine_cache_sync[1] after patching, which is also specified by this blogpost[2] on ARM website. (Though [2] also mentions some branch predictor ops that we don't do today)

I suppose it is still possible that the thread is moved to some other processor element, in which case it'll execute stale code.

I'll add these details to #1911 for now, as we don't know for sure if this is a concern.

[1]: https://github.com/DynamoRIO/dynamorio/blob/b3bc4dfad113575cdb89730168c1113d090c93b8/core/arch/mangle_shared.c#L816

[2]: https://community.arm.com/developer/ip-products/processors/b/processors-ip-blog/posts/caches-and-self-modifying-code

core/arch/emit_utils_shared.c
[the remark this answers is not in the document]
derekbruening: We have explicit functions for loading TLS slots which make it clear which slot it is: is there a reason this couldn't use one of those?  At least it should use the named offsets and not a numeric literal.

core/arch/aarch64/emit_utils.c
```diff
@@ -50,6 +50,69 @@
...
+ *         br   x1
+ *         <target_fragment_prefix>
+ *
+ * To ensure atomicity of <target> patching, the data slot must be 8-byte
+ * aligned. We do this by reserving 12 bytes for the data slot and using the
+ * appropriate offset in ldr for the 8-byte aligned 8 byte region within it.
+ *
+ * For complete design details, see the following wiki
+ * https://github.com/DynamoRIO/dynamorio/wiki/Linking-Far-Fragments-on-AArch64
+ */
```
derekbruening, line 114: Nice, I think this improves readability for future code maintainers (including future versions of ourselves).

core/arch/aarch64/emit_utils.c
derekbruening: > Yes AArch64 has shared fragments

Note that "gencode" in DR refers to just the synthetic code pointed at by generated_code_t (so fcache_{enter,return} and the IBL and variants), not to fragments.  X64 (so x86-64 and A64)'s private fragments use shared gencode.

core/arch/aarch64/emit_utils.c
```diff
@@ -153,6 +216,10 @@ insert_exit_stub_other_flags(dcontext_t *dcontext, fragment_t *f, linkstub_t *l,
         ASSERT(sizeof(app_pc) == 8);
         pc += DIRECT_EXIT_STUB_DATA_SZ / sizeof(uint);
         /* We start off with the fcache-return routine address in the slot. */
+        /* AArch64 uses shared fragments, so fcache_return gencode should be
+         * shared too.
```
derekbruening, line 220: This is inaccurate: the presence of shared fragments does not mean there are no private fragments.  For ease of mangling, selfmod fragments are thread-private; temp fragments during trace building are thread-private; there used to be some cases of syscall-ending fragments that were thread-private.

The key fact is that A64's private fragments use shared gencode (and in fact A64 never generates private gencode).

core/arch/aarch64/emit_utils.c
```diff
@@ -205,24 +272,76 @@ exit_cti_reaches_target(dcontext_t *dcontext, fragment_t *f, linkstub_t *l,
...
+         */
+        *(uint *)vmcode_get_writable_addr(stub_pc) =
+            (0x14000000 | (0x03ffffff & off >> 2));
+        if (hot_patch)
+            machine_cache_sync(stub_pc, stub_pc + 4, true);
+        return;
+    }
+    /* target_pc is a far fragment. We must use an indirect branch. Note that the indirect
+     * branch needs to be to the fragment prefix, as we need to restore the clobbered
+     * regs. */
```
derekbruening, line 296: nit: prefer `*/` on own line (though this may be a losing battle since there's no clang-format rule to enforce it...could be convinced to remove the rule)

core/arch/aarch64/emit_utils.c
```diff
@@ -205,24 +272,76 @@ exit_cti_reaches_target(dcontext_t *dcontext, fragment_t *f, linkstub_t *l,
...
+    return;
 }
 
 bool
 stub_is_patched(dcontext_t *dcontext, fragment_t *f, cache_pc stub_pc)
 {
-    uint *addr = (uint *)get_target_pc_slot(f, stub_pc);
-    return addr != (uint *)fcache_return_routine(dcontext);
+    uint enc;
+    ATOMIC_4BYTE_ALIGNED_READ(vmcode_get_writable_addr(stub_pc), &enc);
```
derekbruening, line 308: No need to convert to writable to read it?

core/arch/aarch64/emit_utils.c
```diff
@@ -205,24 +272,76 @@ exit_cti_reaches_target(dcontext_t *dcontext, fragment_t *f, linkstub_t *l,
...
+    /* linked to far fragment. */
+    return target_pc != (ptr_uint_t)fcache_return_routine(dcontext);
 }
 
 void
 unpatch_stub(dcontext_t *dcontext, fragment_t *f, cache_pc stub_pc, bool hot_patch)
 {
+    /* This implementation consists of two operations to unpatch, but at any given time
+     * only one patching strategy would be used. So, one of these two operations would
+     * really be a no-op. */
```
derekbruening, line 324: Should we read the current state to figure out whether each one needs to be done?  Generally a read and conditional to skip a write is worthwhile, performance-wise.

core/arch/emit_utils_shared.c
```diff
@@ -1897,12 +1903,13 @@ append_jmp_to_fcache_target(dcontext_t *dcontext, instrlist_t *ilist,
...
-             * ldr x0, [x28]
+            /* Load next_tag from FCACHE_ENTER_TARGET_SLOT, stored by
+             * append_setup_fcache_target.
              */
             APP(ilist,
-                XINST_CREATE_load(dcontext, opnd_create_reg(DR_REG_X0),
-                                  OPND_CREATE_MEMPTR(dr_reg_stolen, 16)));
+                XINST_CREATE_load(
+                    dcontext, opnd_create_reg(DR_REG_X0),
+                    OPND_CREATE_MEMPTR(dr_reg_stolen, FCACHE_ENTER_TARGET_SLOT)));
```
derekbruening, line 1912: This could be instr_create_restore_from_tls()?

core/arch/emit_utils_shared.c
```diff
@@ -4923,10 +4934,12 @@ emit_fcache_enter_gonative(dcontext_t *dcontext, generated_code_t *code, byte *p
...
         XINST_CREATE_store(dcontext, OPND_CREATE_MEMPTR(DR_REG_SP, -XSP_SZ),
                            opnd_create_reg(DR_REG_R0)));
-    /* get target PC */
+    /* Load target PC from FCACHE_ENTER_TARGET_SLOT, stored by
+     * by append_setup_fcache_target.
+     */
     APP(&ilist,
         XINST_CREATE_load(dcontext, opnd_create_reg(DR_REG_R0),
-                          OPND_CREATE_MEMPTR(dr_reg_stolen, 16)));
+                          OPND_CREATE_MEMPTR(dr_reg_stolen, FCACHE_ENTER_TARGET_SLOT)));
```
derekbruening, line 4942: This could be instr_create_restore_from_tls()?

derekbruening (approved):

core/arch/emit_utils_shared.c
abhinav92003: Done.

core/arch/emit_utils_shared.c
abhinav92003: Done.

core/arch/aarch64/emit_utils.c
abhinav92003: Thanks, I added a read to pick which write needs to be done.

Then I also found that unpatch_stub is called blindly in some cases. So I made both writes conditional on their respective conditions.

core/arch/aarch64/emit_utils.c
abhinav92003: Done.

core/arch/aarch64/emit_utils.c
abhinav92003: I made the change. `*/` on its own line looks better somehow.

core/arch/aarch64/emit_utils.c
abhinav92003: Thanks. Corrected the comment.

core/arch/aarch64/emit_utils.c
abhinav92003: Definitely. Thanks!

core/arch/aarch64/emit_utils.c
abhinav92003: Resolved in other comment thread.

core/arch/emit_utils_shared.c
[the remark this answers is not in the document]
abhinav92003: Used instr_create_restore_from_tls and the named offset constants.

derekbruening:
(Waiting for "PTAL"...they really should make it easier to know whose turn it is.)

abhinav92003:
> (Waiting for "PTAL"...they really should make it easier to know whose turn it is.)

I'm still working on running a large app under DR so the new path gets invoked (an alternative is to disable the intermediate-fragment linking optimisation and run the test suite).

Sorry for the confusion on whose turn it was. Github allows requesting re-reviews; I usually just do that when the PR is ready again.

derekbruening:
> Github allows requesting re-reviews; I usually just do that when the PR is ready again.

Good idea: let's make that the convention.  Maybe add it to the wiki on code review workflow.

abhinav92003:
> Good idea: let's make that the convention. Maybe add it to the wiki on code review workflow.

Done, added it to https://github.com/DynamoRIO/dynamorio/wiki/Code-Reviews#responding-to-review-comments-and-submitting-code-updates.

abhinav92003:
> I'm still working on running a large app under DR so the new path gets invoked

I tested far fragment linking on a proprietary app, and also on the `ibl-stress` test which I ported to AArch64. Unfortunately, the `ibl-stress` options required to invoke the far fragment linking path on AArch64 cannot be used in CI; they require ~12 mins to run. But I've enabled the `ibl-stress` test on AArch64 anyway in this PR, as it improves coverage. I added instructions in `ibl-stress.c` on how to use it to locally invoke/test far fragment linking in AArch64.

core/arch/aarch64/emit_utils.c
abhinav92003: Let me know if you think there are any atomicity concerns with the current implementation of `unpatch_stub`.

core/arch/aarch64/emit_utils.c
abhinav92003: Yes, got it. Thanks.

core/arch/aarch64/emit_utils.c
```diff
@@ -199,7 +199,8 @@ insert_exit_stub_other_flags(dcontext_t *dcontext, fragment_t *f, linkstub_t *l,
         /* mov x0, ... */
         pc = insert_mov_imm(pc, DR_REG_X0, (ptr_int_t)l);
         num_nops_needed = 4 - (pc - write_stub_pc - 1);
-        uint target_pc_slot_offs = (uint *)get_target_pc_slot(f, stub_pc) - pc;
+        uint *target_pc_slot = (uint *)get_target_pc_slot(f, stub_pc);
+        uint target_pc_slot_offs = target_pc_slot - pc;
```
derekbruening, line 203: Maybe ASSERT(pc < target_pc_slot) so it's clear there's no underflow

core/arch/aarch64/emit_utils.c
```diff
@@ -199,7 +199,8 @@ insert_exit_stub_other_flags(dcontext_t *dcontext, fragment_t *f, linkstub_t *l,
         /* mov x0, ... */
         pc = insert_mov_imm(pc, DR_REG_X0, (ptr_int_t)l);
         num_nops_needed = 4 - (pc - write_stub_pc - 1);
-        uint target_pc_slot_offs = (uint *)get_target_pc_slot(f, stub_pc) - pc;
+        uint *target_pc_slot = (uint *)get_target_pc_slot(f, stub_pc);
```
derekbruening, line 202: The `uint *` type is confusing: the pc is 8 bytes, not 4.  It makes this look like a bug where it assumes the pc fits in a uint.

core/arch/aarch64/emit_utils.c
```diff
@@ -213,15 +214,15 @@ insert_exit_stub_other_flags(dcontext_t *dcontext, fragment_t *f, linkstub_t *l,
         /* The final slot is a data slot, which will hold the address of either
          * the fcache-return routine or the linked fragment.
          */
+        ASSERT(pc == target_pc_slot || pc + 1 == target_pc_slot);
```
derekbruening, line 217: The +1 case is alignment padding?

core/arch/aarch64/emit_utils.c
```diff
@@ -213,15 +214,15 @@ insert_exit_stub_other_flags(dcontext_t *dcontext, fragment_t *f, linkstub_t *l,
...
         /* We start off with the fcache-return routine address in the slot. */
-        /* AArch64 uses shared fragments, so fcache_return gencode should be
-         * shared too.
+        /* AArch64 uses shared gencode. So, fcache_return routine address should be
+         * same, no matter which thread creates/unpatches the stub.
          */
         ASSERT(fcache_return_routine(dcontext) == fcache_return_routine(GLOBAL_DCONTEXT));
-        *(uint64 *)get_target_pc_slot(f, stub_pc) =
-            (ptr_uint_t)fcache_return_routine(dcontext);
+        *(uint64 *)target_pc_slot = (ptr_uint_t)fcache_return_routine(dcontext);
```
derekbruening, line 225: See comment about treating target_pc_slot as uint*: this uint64* makes more sense.  But it seems odd here to have uint64* on one side, but ptr_uint_t on the other.

core/arch/aarch64/emit_utils.c
```diff
@@ -293,55 +294,70 @@ patch_stub(fragment_t *f, cache_pc stub_pc, cache_pc target_pc, cache_pc target_
...
+static bool
+stub_is_patched_for_intermediate_fragment_link(dcontext_t *dcontext, cache_pc stub_pc)
 {
     uint enc;
-    ATOMIC_4BYTE_ALIGNED_READ(vmcode_get_writable_addr(stub_pc), &enc);
-    /* linked to intermediate fragment. */
-    if ((enc & 0x14000000) == 0x14000000)
-        return true;
+    ATOMIC_4BYTE_ALIGNED_READ(stub_pc, &enc);
+    return (enc & 0x14000000) == 0x14000000;
```
derekbruening, line 311: A named constant would help to know where this magic value comes from.

core/arch/aarch64/emit_utils.c
```diff
@@ -293,55 +294,70 @@ patch_stub(fragment_t *f, cache_pc stub_pc, cache_pc target_pc, cache_pc target_
...
 {
-    /* This implementation consists of two operations to unpatch, but at any given time
-     * only one patching strategy would be used. So, one of these two operations would
-     * really be a no-op. */
-    /* For near fragment link: Restore the stp x0, x1, [x(stolen), #(offs)]
-     * i#1911: Patching unconditional branch to some arbitrary instruction
-     * is theoretically not sound. Architectural specifications do not
-     * guarantee safe behaviour or any bound on when the change will be
-     * visible to other process elements.
+    /* At any time, atmost one patching strategy will be in effect: the one for
```
derekbruening, line 333: at most

core/arch/aarch64/emit_utils.c
```diff
@@ -293,55 +294,70 @@ patch_stub(fragment_t *f, cache_pc stub_pc, cache_pc target_pc, cache_pc target_
...
-    /* We set hot_patch to false as we are not modifying code. */
-    ATOMIC_8BYTE_ALIGNED_WRITE(target_pc_slot,
-                               (ptr_uint_t)fcache_return_routine(dcontext),
-                               /*hot_patch=*/false);
+    if (stub_is_patched_for_intermediate_fragment_link(dcontext, stub_pc)) {
+        /* Restore the stp x0, x1, [x(stolen), #(offs)]
+         * i#1911: Patching unconditional branch to some arbitrary instruction
+         * is theoretically not sound. Architectural specifications do not
+         * guarantee safe behaviour or any bound on when the change will be
+         * visible to other process elements.
```
derekbruening, line 341: "processing elements" (else it sounds like an OS construct)

core/arch/aarch64/emit_utils.c
```diff
@@ -293,55 +294,70 @@ patch_stub(fragment_t *f, cache_pc stub_pc, cache_pc target_pc, cache_pc target_
...
+         */
+        *(uint *)vmcode_get_writable_addr(stub_pc) =
+            (0xa9000000 | 0 | 1 << 10 | (dr_reg_stolen - DR_REG_X0) << 5 |
+             TLS_REG0_SLOT >> 3 << 15);
+        if (hot_patch)
+            machine_cache_sync(stub_pc, stub_pc + AARCH64_INSTR_SIZE, true);
+    }
+    if (stub_is_patched_for_far_fragment_link(dcontext, f, stub_pc)) {
```
derekbruening, line 349: So this should be an `else if`?

suite/tests/api/ibl-stress.c
```diff
@@ -37,10 +37,19 @@
...
-#ifdef LARGER_TEST
+#if defined(LARGER_TEST)
 /* 20K sequences gives us ~150K bbs. */
 #    define NUM_SEQUENCES 20000
 #    define NUM_THREADS 16
+#elif defined(TEST_FAR_LINK_AARCH64)
+/* These values trigger far fragment linking path on AArch64. With this option,
+ * the test is too big to enable in CI. The following changes are required to
+ * the regular ctest options:
+ * - set timeout to 15 mins, instead of the default 90 seconds
```
derekbruening, line 48: You can have these changes automatically made, right?  No reason to require manual changes.

derekbruening (approved):

core/arch/aarch64/emit_utils.c
abhinav92003: Done.

core/arch/aarch64/emit_utils.c
abhinav92003: Made it a `ptr_uint_t*` type.

core/arch/aarch64/emit_utils.c
abhinav92003: Yes.

abhinav92003: I added some clarification in the comment above.

core/arch/aarch64/emit_utils.c
abhinav92003: LHS is `ptr_uint_t*` too now.

core/arch/aarch64/emit_utils.c
abhinav92003: I added a comment about the opcode, like many other spots in this file.

core/arch/aarch64/emit_utils.c
abhinav92003: Done.

core/arch/aarch64/emit_utils.c
abhinav92003: Done.

core/arch/aarch64/emit_utils.c
abhinav92003: Done. Oops, thanks.

suite/tests/api/ibl-stress.c
abhinav92003: The current support for marking tests as short/long seems to only allow applying different DR options, and it's tricky to skip long tests in the regular suite. https://github.com/DynamoRIO/dynamorio/blob/eb57ba9555cfd19fa6889a2ba764aba9a524d39d/suite/tests/CMakeLists.txt#L99.

So I created a new instance of `ibl-stress` that is compiled with `-DTEST_FAR_LINK_AARCH64`. I marked it with `_LONG` suffix and added logic to skip such marked long tests in `is_test_enabled_by_name`, based on the `TEST_LONG` flag.

abhinav92003:
'client.stolen-reg` failure seems flaky -- passes locally, also in previous commit.

derekbruening:
> `client.stolen-reg` failure seems flaky -- passes locally, also in previous commit.

If you could file an issue (if there isn't one already).

suite/tests/CMakeLists.txt
```diff
@@ -2895,6 +2897,13 @@ if (CLIENT_INTERFACE)
     if (AARCH64)
       tobuild_api(api.ibl-stress api/ibl-stress.c
         "-disable_traces -shared_bb_ibt_tables -shared_ibt_table_bb_init 16 ${checklevel}" "" OFF OFF)
+
+      tobuild_api(api.ibl-stress-aarch64-far-link_LONG api/ibl-stress.c
+        "-disable_traces -shared_bb_ibt_tables -shared_ibt_table_bb_init 20 ${checklevel}" "" OFF OFF)
+      append_property_string(TARGET api.ibl-stress-aarch64-far-link_LONG COMPILE_FLAGS "-DTEST_FAR_LINK_AARCH64")
```
derekbruening, line 2903: style: Line too long

derekbruening (approved):

suite/tests/api/ibl-stress.c
derekbruening: Yes I think there was no prior concept of labeling an individual test.  Looks like a good solution.

core/arch/aarch64/emit_utils.c
derekbruening: I think it makes sense: cache sync for tweaking code (xref the separate threads on that but those will be dealt w/ separately as discussed), store-release for changing data.

suite/tests/CMakeLists.txt
abhinav92003: Done.

Issue #4641, opened 2020-12-29
CRASH from race on detach in LOG debug-build code


derekbruening:
While testing a fix for #3535, running debug build I hit this crash maybe once in 100 runs:

```
#4  0x00007fe6e8f590a5 in getchar () at getchar.c:37
#5  0x00005640f8054db3 in handle_signal (signal=11, siginfo=0x7fe6e56cff30, ucxt=0x7fe6e56cfe00) at /home/bruening/dr/git/src/suite/tests/api/detach_signal.cpp:105
#6  <signal handler called>
#7  0x00007fe6e93d1c72 in master_signal_handler_C (sig=7, siginfo=0x7fe6e56d0770, ucxt=0x7fe6e56d0640, xsp=0x7fe6e56d0638 ";\016\071\351\346\177") at /home/bruening/dr/git/src/core/unix/signal.c:5107

   0x00007fe6e93d1c59 <+1004>:	lea    0x161388(%rip),%rax        # 0x7fe6e9532fe8 <d_r_stats>
   0x00007fe6e93d1c60 <+1011>:	mov    (%rax),%rax
   0x00007fe6e93d1c63 <+1014>:	test   %rax,%rax
   0x00007fe6e93d1c66 <+1017>:	je     0x7fe6e93d1ccc <master_signal_handler_C+1119>
   0x00007fe6e93d1c68 <+1019>:	lea    0x161379(%rip),%rax        # 0x7fe6e9532fe8 <d_r_stats>
   0x00007fe6e93d1c6f <+1026>:	mov    (%rax),%rax
==>
   0x00007fe6e93d1c72 <+1029>:	mov    0x218(%rax),%eax
   0x00007fe6e93d1c78 <+1035>:	test   %eax,%eax
   0x00007fe6e93d1c7a <+1037>:	je     0x7fe6e93d1ccc <master_signal_handler_C+1119>
   0x00007fe6e93d1c7c <+1039>:	lea    0x161365(%rip),%rax        # 0x7fe6e9532fe8 <d_r_stats>
   0x00007fe6e93d1c83 <+1046>:	mov    (%rax),%rax
   0x00007fe6e93d1c86 <+1049>:	mov    0x214(%rax),%eax
   0x00007fe6e93d1c8c <+1055>:	and    $0x10,%eax
   0x00007fe6e93d1c8f <+1058>:	test   %eax,%eax
[...]

PR #4642, opened 2020-12-29
i#3535: Solve detach handler race

### LINKED ISSUE #3535: safe_read magic unstable during detach from threaded app.
Following happens with a small reproducer that we produced off Google code.

1) DR is attached to app running threaded code. The app is sending PROF signals (but type of signal doesn't matter, as long as it is asynchronous and process-directed).
2) The app starts a detach via start/stop API (with_cleanup() path).
3) Detach syncs and all threads go native. At this point, the thread's TLS (DynamoRIO's) has been removed.
4) Main thread is not done cleaning up, signal handlers are still DR's.
5) PROF signal comes in, child thread receives it. Signal handler calls safe_read_tls_magic().
<=>
6) Corner case: main thread removes signal handlers and restores the app's handlers.
7) safe_read_tls_magic() segfaults because no TLS.
8) SIGSEGV is delivered to app.

derekbruening:
Eliminates a race on detach where the detaching thread removes DR's
SIGSEGV handler while a now-native detached thread is in the middle of
having a signal delivered (natively) and invokes a TLS magic field
safe read, whose SIGSEGV then goes to the application.

The detaching thread is the one doing all the real cleanup, so we
simply avoid any safe reads or TLS for detaching threads by recording
the detacher's ID when we start the detach process.  This var is not
cleared until re-init, so we have no race with the end of detach.

Tested on api.detach_signal with the forthcoming signal mask checks,
which trigger when the handler is invoked for a DR signal instead of
an app-generated signal.  Without this fix, the test fails easily:
about 1 in 5 runs in debug build.  With this fix, it succeeds 200x in
a loop.  I still see one type of crash in debug build, a rare race
where d_r_stats is set to NULL in between the check and use of a
LOG(), but that is limited to debug and is beyond the scope of this
fix and is much lower priority: I filed it as #4641.

Fixes #3535

### DIFF
--- a/core/dynamo.c
+++ b/core/dynamo.c
@@ -627,6 +627,11 @@ dynamorio_app_init(void)
             sideline_init();
 #endif
 
+        /* We can't clear this on detach like other vars b/c we need native threads
+         * to continue to avoid safe_read_tls_magic() in is_thread_tls_initialized().
+         * So we clear it on (re-)init.
+         */
+        detacher_tid = INVALID_THREAD_ID;
         /* thread-specific initialization for the first thread we inject in
          * (in a race with injected threads, sometimes it is not the primary thread)
          */
--- a/core/globals.h
+++ b/core/globals.h
@@ -533,6 +533,7 @@ extern bool dynamo_all_threads_synched; /* are all other threads suspended safel
  * go through the app interface.
  */
 extern bool doing_detach;
+extern thread_id_t detacher_tid;
 
 extern event_t dr_app_started;
 extern event_t dr_attach_finished;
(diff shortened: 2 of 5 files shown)


derekbruening:
To make it even more clear: I added signal mask checking to the api.detach_signal test and that's where I reproduced #3535 and tested the fix.  That test, covering this PR, will come in separately once this PR is in.

abhinav92003 (approved):

abhinav92003:
`ci-windows` is showing some weird error; it's there on other PRs too.

derekbruening:
> `ci-windows` is showing some weird error; it's there on other PRs too.

@yury-khrustalev identified it as the doxygen download link being broken:
https://github.com/DynamoRIO/dynamorio/pull/4637#issuecomment-752144193

PR #4643, opened 2020-12-30
Use a stable URL for downloading Windows doxygen

derekbruening:
The doxygen.nl URL breaks every few months when they put out their
latest release and remove the old release.  We switch to their
sourceforge site which archives old releases, so we won't break in the
near future.  (Xref #4000 where we used to use chocolatey to avoid
specifying the precise version: but choco was flaky.)


abhinav92003:
A different failure in ci-windows now it seems.

abhinav92003 (approved):

PR #4644, opened 2020-12-30
i#1921 native sig: Properly set native signal mask

derekbruening:
Adds proper signal mask setting for native signals, which requires
adding handler-masked signals to the DR frame's mask instead of
setting the mask in the kernel.

Adds tests to api.detach_signal.  These mask tests also detect when a
DR-caused crash from a DR bug during detach shows up in the app's
handler post-detach due to detach races, including from #3535.

Issue: #1921, #3535


derekbruening:
Windows failure is #3320:
```
2020-12-30T06:35:20.6127202Z 205: Invalid trace entry type 23 before a bundle
2020-12-30T06:35:20.6181667Z 205: Assertion failed: type_is_instr(cur_ref_.instr.type) || cur_ref_.instr.type == TRACE_TYPE_INSTR_NO_FETCH, file ..\clients\drcachesim\reader\reader.cpp, line 155
```

abhinav92003 (approved):

PR #4646, opened 2021-01-01
More operand size work

khuey:
This fixes a couple more cases from #4297. Should be self-explanatory.


johnfxgalea (approved):
Awesome! Thanks for adding more fixes. LGTM.

johnfxgalea:
Just elaborate the commit message on what this PR actually does please, i.e., fixes operand sizes for vcvtps2pd etc. This makes the commit history more readable. Thank you.

Issue #1615, opened 2015-03-04
pre-xl8 raw_mcontext for signals holds post-xl8 simd state


derekbruening:
The dr_restore_state_info_t.raw_mcontext is supposed to hold the
pre-translation machine state in a dr_mcontext_t.  However, the
signal-handling code only keeps a copy of the sigcontext_t prior to
translation.  For x86, the simd state is shared between the pre- and
post- translation contexts.  Thus if a client modifies simd state on
translation, the wrong values will be in raw_mcontext.

PR #3893, opened 2019-10-12
i#3844: Extend drreg for other types of spills

johnfxgalea:
Adds supports to drreg for XMM spillage and also a framework for other types of register spills.

ARM not supported yet.

Issue: #3844

PR #4649, opened 2021-01-02
i#1615: Maintain pre translation values of SIMD registers

johnfxgalea:
The PR copies SIMD values prior to translation when handling signals. This won't result in corrupting SIMD values with post-translation values.

Two helper functions are introduced. These include copy_sigcontext() and restore_orig_sigcontext(). The former copies pre-translation values of the sigcontext, including SIMD regs, while the latter restores a frame's sigcontext with pre-translation values

Issue: #1615


sapostolakis:
Hey @johnfxgalea. Is there any issue that is blocking this PR?

johnfxgalea:
@sapostolakis I would like to finish off https://github.com/DynamoRIO/dynamorio/pull/3893 so I can integrate sophisticated tests for this PR.

sapostolakis:
Sounds good.

Issue #4567, opened 2020-11-30
AT&T disassembly of an instruction with a single AVX 512 mask source asserts


khuey:
```
./clients/bin64/drdisas -syntax intel 62 f2 fe 48 28 c9
 62 f2 fe 48 28 c9    vpmovm2w zmm1 {k1}
./clients/bin64/drdisas -syntax att 62 f2 fe 48 28 c9
Aborted (core dumped)
...
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7bda859 in __GI_abort () at abort.c:79
#2  0x00005555555b0b9a in external_error (file=0x5555555cfdb8 "/home/khuey/dev/dynamorio/core/ir/x86/disassemble.c", line=238, msg=0x5555555cfe4e "missing decode type")
    at /home/khuey/dev/dynamorio/core/ir/decodelib.c:132
#3  0x00005555555ade95 in opnd_disassemble_noimplicit (buf=0x7fffffffdc40 " 62 f2 fe 48 28 c9    vpmovm2w %zmm1 {", bufsz=228, sofar=0x7fffffffdc38, dcontext=0xffffffffffffffff,
    instr=0x7fffffffdb50, optype=0 '\000', opnd=..., prev=false, multiple_encodings=false, dst=false, idx=0x7fffffffda14) at /home/khuey/dev/dynamorio/core/ir/x86/disassemble.c:238
#4  0x00005555555ac705 in instr_disassemble_opnds_noimplicit (buf=0x7fffffffdc40 " 62 f2 fe 48 28 c9    vpmovm2w %zmm1 {", bufsz=228, sofar=0x7fffffffdc38, dcontext=0xffffffffffffffff,
    instr=0x7fffffffdb50) at /home/khuey/dev/dynamorio/core/ir/disassemble_shared.c:976
#5  0x00005555555ad0e3 in internal_instr_disassemble (buf=0x7fffffffdc40 " 62 f2 fe 48 28 c9    vpmovm2w %zmm1 {", bufsz=228, sofar=0x7fffffffdc38, dcontext=0xffffffffffffffff,
    instr=0x7fffffffdb50) at /home/khuey/dev/dynamorio/core/ir/disassemble_shared.c:1139
[...]

PR #4651, opened 2021-01-02
Don't use libc's stderr in libdrdecode

khuey:
This fixes a regression introduced in #4567.

### DIFF
--- a/core/ir/decodelib.c
+++ b/core/ir/decodelib.c
@@ -47,11 +47,15 @@
 /* initialize to all zeros.  disassemble_set_syntax will write to it. */
 options_t dynamo_options;
 
+#undef STDERR
+
 /* support use of STD* macros so user doesn't have to use "stdout->_fileno" */
 #    ifdef UNIX
 file_t our_stdout = STDOUT_FILENO;
 file_t our_stderr = STDERR_FILENO;
 file_t our_stdin = STDIN_FILENO;
+
+#define STDERR (our_stderr)
 #    endif
 
 #    ifdef WINDOWS
@@ -72,6 +76,8 @@ dr_get_stdin_file(void)
 {
     return GetStdHandle(STD_INPUT_HANDLE);
 }
+
+#define STDERR (dr_get_stderr_file())
 #    endif
 
 static uint vendor = VENDOR_INTEL; /* default */
--- a/core/ir/decodelib.c
+++ b/core/ir/decodelib.c
@@ -128,7 +128,7 @@ get_thread_private_dcontext(void)
 void
 external_error(const char *file, int line, const char *msg)
 {
-    fprintf(stderr, "Usage error: %s (%s, line %d)\n", msg, file, line);
+    print_file(STDERR, "Usage error: %s (%s, line %d)\n", msg, file, line);
     abort();
 }
 
--- a/core/ir/decodelib.c
+++ b/core/ir/decodelib.c
@@ -47,7 +47,7 @@
 /* initialize to all zeros.  disassemble_set_syntax will write to it. */
 options_t dynamo_options;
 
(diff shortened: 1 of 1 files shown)


khuey:
Hrm, that fixes the Unix variant platforms but Windows requires something else apparently.

core/ir/decodelib.c
```diff
@@ -128,7 +128,7 @@ get_thread_private_dcontext(void)
 void
 external_error(const char *file, int line, const char *msg)
 {
-    fprintf(stderr, "Usage error: %s (%s, line %d)\n", msg, file, line);
+    print_file(STDERR, "Usage error: %s (%s, line %d)\n", msg, file, line);
```
derekbruening, line 131: Not sure drdecode links utils.c which has print_file.  The `STDERR` should work b/c it turns into either `our_stderr` or `dr_get_stderr_handle` and both are defined up above in this file.  Probably this should be `s/print_file/fprintf/`?

derekbruening: Never mind, STDERR is the fileno and not the FILE*.  Plus, print_file is defined in this very file.

derekbruening (approved):

core/ir/decodelib.c
```diff
@@ -128,7 +128,7 @@ get_thread_private_dcontext(void)
 void
```
derekbruening, line 128: The commit title and description are not following the conventions in multiple respects:

+ The title should start with the issue number
+ The description should describe what is being fixed (yes, duplicating the title)
+ There should be a Fixes line at the bottom

Please see https://github.com/DynamoRIO/dynamorio/wiki/Code-Reviews#commit-messages

derekbruening (changes requested):

khuey:
Thanks for approving this I guess but I still don't know how to make it build on Windows :)

derekbruening:
Pasting the error would make it easier so we don't have to go click on the logs too.  Needs get_stderr_handle, etc. equivalent to the dr_ ones right?

khuey:
```
 ====> FAILURE in debug-internal-32 <====
debug-internal-32: **** 2 build errors ****
	drdecode.lib(decodelib.c.obj) : error LNK2019: unresolved external symbol _get_stderr_handle referenced in function _external_error
	clients\bin32\drdisas.exe : fatal error LNK1120: 1 unresolved externals
```

khuey:
I guess I could s/STDERR/dr_get_stderr_file()/ perhaps?

derekbruening:
globals_shared.h has `define STDERR (dr_get_stderr_file())` which is in decodelib.c, but globals.h (internal) has `#    define STDERR (file_t) get_stderr_handle()` which is not in decodelib.c -- maybe it needs both or an alias or whatnot?

derekbruening:
I would add get_stderr_handle which calls dr_get_stderr_handle or sthg

derekbruening:
> I guess I could s/STDERR/dr_get_stderr_file()/ perhaps?

Also seems good -- I guess the internal define is being used for decodelib itself and so there is no other reason for the non dr_ versions.  External users of the lib are going to have the external define and thus need the dr_ version.

khuey:
> 
> 
> > I guess I could s/STDERR/dr_get_stderr_file()/ perhaps?
> 
> Also seems good -- I guess the internal define is being used for decodelib itself and so there is no other reason for the non dr_ versions. External users of the lib are going to have the external define and thus need the dr_ version.

Doing this would require defining dr_get_stderr_file for the Unix platforms as its also windows only.

khuey:
Alright how about that?

derekbruening (approved):
Works for me.  Don't forget the commit title + description conventions.

Issue #607, opened 2014-11-28
drinjectlib early injection support


derekbruening:
_From [bruen...@google.com](https://code.google.com/u/109494838902877177630/) on November 12, 2011 15:04:00_

this was PR 211367

easiest w/ earliest injection => issue #234 prereq

drinject today doesn't use early_inject (and changing it do so is non-trivial since it will require alt. methods of gathering the relevant addresses, will be easier once we have earliest injection)

coming in late causes problems and we end up telling people to create a parent process and inject from there as a workaround

_Original issue: http://code.google.com/p/dynamorio/issues/detail?id=607_

Issue #147, opened 2014-11-27
client interactions with cross-arch execve


derekbruening:
_From [derek.br...@gmail.com](https://code.google.com/u/117968039472581148324/) on May 23, 2009 10:55:03_

With debug build and a 32-bit client, if an app does execve to a 64-bit
app, we'll kill the process with a usage error due to the 32-bit client
mismatch:

<-- parent 16846 forked child 16903 -->
<-- execve /bin/sh -->
<Application sh (16903) DynamoRIO usage error :
        Error opening instrumentation library
/work/dr/tot/exports_drmemory/libdrmemory.so:
        /work/dr/tot/exports_drmemory/libdrmemory.so: wrong ELF class:
ELFCLASS32>

We should at least provide an option to turn this into a non-fatal error,
to enable using debug build in such situations. In my usage case the execve
of 64-bit sh is followed by execve of 32-bit perlbmk, and I'm fine with my
client not being loaded into sh, but I do want to be loaded into perlbmk.
This is what happens today with a release build, with a non-fatal message:

<Application sh (18109). Unable to load client library:
        Error opening instrumentation library
/work/dr/tot/exports_drmemory/libdrmemory.so:
        /work/dr/tot/exports_drmemory/libdrmemory.so: wrong ELF class:
ELFCLASS32.>
<Starting application sh (18109)>
<Starting application perlbmk (18110)>

We should consider what model we want to support for clients in such scenarios:
specify separate 32-bit and 64-bit clients?
Xref all the similar Windows issues: PR 240257, PR 254193, issue #49.
Those Windows cases also focus on mixed-mode processes.

_Original issue: http://code.google.com/p/dynamorio/issues/detail?id=147_

derekbruening:
I think there are two possible solutions and that we should do both?

+ Add a new option to specify the other-arch path to the client(s).  -client32, -client64, and shortcuts -c32 and -c64?  So you could do `bin64/drrun -c bin64/client64.so -c32 bin32/client32.so` or `bin32/drrun -c bin/client32.so -c64 bin64/client64.so` or `bin64/drrun -c64 bin64/client64.so -c32 bin32/client32.so`.

+ Have the pre-execve code try some standard substitutions for the existing client paths like `s/bin32/bin64/;s/bin/bin64/;s/lib32/lib64;s/lib/lib64;`.

@johnfxgalea wondering if you've ever hit this?  It seems surprising that nobody has really hit this in all these years.

[...]

Issue #49, opened 2014-11-27
simultaneous 32-bit and 64-bit library support for x64 DR controlling WOW64 app


derekbruening:
_From [derek.br...@gmail.com](https://code.google.com/u/117968039472581148324/) on February 24, 2009 14:14:20_

Today 32-bit DR can control the 32-bit parts of a WOW64 app but to see all
of the code including the emulation layer we want 64-bit DR able to run the
whole mixed-mode app.  Some of the capabilities here also apply to Linux
mixed-mod apps but those are much, much rarer.

here is my list of cases that will eventually be separately filed here:
- PR 240257: support 32-bit clients on WOW64?  how mix 32 and 64 bit code?
  like Pin, give stream to separate clients?  or give 32-bit code to
  64-bit client?
- PR 253431: [wow64] simultaneous 32-bit and 64-bit dll support in 64-bit DR
- PR 314367: re-enable x64 DR controlling WOW64 process once it works
- PR 272553: [x64] late injection must switch from kernel32 to ntdll for
  wow64 children
- PR 271317: preserve cs changes from far ctis and iret
- PR 283895: [x64][correctness][performance] for x86 code use separate
  x86 ibl tables and compacted or separate tls
- PR 283152: support high bit preservation across mode changes
- PR 284029: [x64] support syscalls in x86 mode
  TODO: reg_spill_dcontext_offs(reg_id_t reg):
     /\* Use REG_E?? instead of REG_X?? to eventually support 32-bit code
  spills in 
  - mixed 64-bit/32-bit execution. */
- PR 269595: WOW64 context translation failing when at our own
  post-syscall point
- PR 254193: [x64] inject into different-architecture child: x64 to
  WOW64, WOW64 to x64
  => long-term we'll only support 64-bit-DR in WOW64 following (PR 253431)
- PR 253943: [x64] support sysenter
- PR 255555: [x64] 32-bit drinject options for launching 64-bit exe
  how know if ilist is 32-bit: instr_get_x86_mode() on each instr, or
  can assume if 1st then whole is same?  shouldn't matter for most ops:
  IR is rich enough and cross-platform enough

_Original issue: http://code.google.com/p/dynamorio/issues/detail?id=49_

derekbruening:
We never got as far as running clients: for clients there are issues with static things like DR_REG_X* defines.  For that, xref inside DR's IBL generation:

[...]

PR #4653, opened 2021-01-04
i#803: Cross-arch Windows injection

### LINKED ISSUE #803: [x64] inject into different-architecture child: x64 to WOW64, WOW64 to x64
_From [bruen...@google.com](https://code.google.com/u/109494838902877177630/) on June 12, 2012 09:52:13_

this was PR 254193

xref issue #20 : handle execve of different-architecture child: 32 to 64, 64 to 32
xref issue #147 /PR 408318: client interactions with cross-arch execve

xref PR 240257: support 32-bit clients on WOW64?  how mix 32 and 64 bit code?
xref issue #49 /PR 253431
xref issue #381 /PR 215423
(xref PR 249112: [x64] inject into child process)
(xref issue #142 /PR 251677: [x64] early inject into child process)

-----------------------------------------
old notes which may not all apply:

For deciding which DR to inject, probably should just have 64-bit config
take precedence: if 32-bit config is also present, too bad, it gets
ignored.

For the child-doesn't-equal parent cases we'll have to have the other
architecture's bootstrap code to inject into the child (or for 64-to-32
fool the loader into loading the 32-bit lib if go early and run 64-bit
bootstrap code): will require a cross-compiled asm routine, or gencode, for
inject-into-thread.

Perhaps worth reconsidering what we really want to bother to support here
(PR 240257).

One plan is to never support 32-bit DR following into 64-bit child.
Long-term we'll only support 64-bit-DR in WOW64 following children (PR
253431).  Only when a client hasn't been ported will we support 32-bit DR
in a WOW64 process,
So if dr64 is in a process that launches a wow64 child foo.exe where
foo.exe has only a wow64 config, our stance is that we won't inject into
the child (since we don't bother to support cross-arch child injection) and
you'll have to rely on AppInit.

_Original issue: http://code.google.com/p/dynamorio/issues/detail?id=803_

derekbruening:
Adds a long-missing feature: following into a Windows child process of
a different bitwidth.

Switches injection from DR and from drinjectlib (including drrun and
drinject) to use -early_inject_map.  This was most easily done by
turning on -early_inject by default as well.  However, the
-early_inject_location default is INJECT_LOCATION_ImageEntry, which is
the same late takeover point as with thread injection.  Switching all
injection over to map-from-the-parent simplifies cross-arch following,
as well as making it easier to shift the takeover point to an earlier
spot in the future.  This is a step toward #607 by switching
drinjectlib to use map injection; the takeover point, as mentioned, is
still the image entry.

Adds an -inject_x64 option to inject a 64-bit DR lib into a 32-bit
child from a 64-bit parent, but this option is only sketched out and
is not fully supported yet: #49 covers adding tests and official
support.

Adds library swapping code to find the other-bitwidth library, which
assumes a parallel directory structure.  Add a new fatal error if the
library for a child is not found.

To support generating code for all 3 child-parent cases (same-same,
32-64, and 64-32), and in particular for 32-64, switches the small
gencode sequence for -early_inject_map from using IR to using raw
bytes.  A multi-arch encoder (#1684) would help but we would need
cross-bitwidth support there, which is not on the horizon.  Fixes what
look like bugs in the original gencode generation along the way
(s/pc/cur_local_pos/ and s/local_code_buf/remote_code_buf/): it's not
clear how it worked before.

Adds support for several system calls from a 32-bit parent to a 64-bit
child where the desired NtWow64* system call does not exist.  We use
switch_modes_and_call() for NtProtectVirtualMemory and
NtQueryVirtualMemory.

Changes all types in the injection code to handle 64-bit addresses in
32-bit code.  Adds UNICODE_STRING_32 and
RTL_USER_PROCESS_PARAMETERS_32 for handling 32-bit structures from
64-bit parents.  Similarly, adds RTL_USER_PROCESS_PARAMETERS_64 and
PROCESS_BASIC_INFORMATION64.

Adds get_process_imgname_cmdline() capability for 64-bit remote from 32-bit.

Adds get_remote_proc_address() and uses it to look up
dynamorio_earliest_init_takeover() in a child DR.

Finds the remote ntdll base via a remote query memory walk plus remote
image header parsing.  This requires adding a switch_modes_and_call()
version of NtQueryVirtualMemory (also mentioned above), which needs
64-bit args: so we refactor switch_modes_and_call() to take in a
struct of all 64-bit fields for the args.

Fixes a few bugs in other routines to properly get the image name and
image entry for 32-bit children of 64-bit parents.

Updates environment variable propagation code to handle a 32-bit
parent and a 64-bit child.  Updates a 64-bit parent and 32-bit child
to insert the variables into the 32-bit PEB (64-bit does no good),
which requires finding the 32-bit PEB.  This is done via the 32-bit
TEB, using a hack due to what seems like a kernel bug where it has the
TebBaseAddress 0x2000 too low.

Makes environment variable propagation failures fatal and visible,
unlike previously where errors would just result in silently letting
the child run natively.  Turns some other prior soft errors into fatal
errors on child takeover.

Moves environment variable propagation to post-CreateUserProcess
instead of waiting for ResumeThread, which avoids having to get the
thread context (for which we have no other-bitwidth support) to figure
out whether it's the first thread in the process or not.  We bail on
propagation for pre-Vista where we'd have to wait for ResumeThred.

Generalizes the other-bitwidth Visual Studio toolchain environment
variable setting for use in a new build-and-test other-bitwidth test
which builds dynamorio and the large_options client (to ensure options
are propagated to children; and it has convenient init and exit time
prints) for the other bitwidth, arranges parallel lib dirs, and runs
the other client

Issue: #803, #147, #607, #49
Fixes #803

### DIFF
--- a/core/win32/ntdll.h
+++ b/core/win32/ntdll.h
@@ -1,5 +1,5 @@
 /* **********************************************************
- * Copyright (c) 2011-2018 Google, Inc.  All rights reserved.
+ * Copyright (c) 2011-2021 Google, Inc.  All rights reserved.
  * Copyright (c) 2003-2010 VMware, Inc.  All rights reserved.
  * **********************************************************/
 
@@ -95,7 +95,21 @@
 #    define ATTACH_PARENT_PROCESS ((DWORD)-1)
 #endif
 
-#ifndef X64
+#ifdef X64
+typedef struct _UNICODE_STRING_32 {
+    /* Length field is size in bytes not counting final 0 */
+    USHORT Length;
+    USHORT MaximumLength;
+    uint Buffer;
+} UNICODE_STRING_32;
+
+typedef struct _RTL_USER_PROCESS_PARAMETERS_32 {
+    uint Reserved[14];
+    UNICODE_STRING_32 ImagePathName;
+    UNICODE_STRING_32 CommandLine;
+    uint Environment;
+} RTL_USER_PROCESS_PARAMETERS_32, *PRTL_USER_PROCESS_PARAMETERS_32;
+#else
 typedef struct ALIGN_VAR(8) _UNICODE_STRING_64 {
     /* Length field is size in bytes not counting final 0 */
     USHORT Length;
@@ -109,6 +123,14 @@ typedef struct ALIGN_VAR(8) _UNICODE_STRING_64 {
         uint64 Buffer64;
     } u;
 } UNICODE_STRING_64;
+
+typedef struct _RTL_USER_PROCESS_PARAMETERS_64 {
+    BYTE Reserved1[16];
+    uint64 Reserved2[10];
+    UNICODE_STRING_64 ImagePathName;
+    UNICODE_STRING_64 CommandLine;
+    uint64 Environment;
+} RTL_USER_PROCESS_PARAMETERS_64, *PRTL_USER_PROCESS_PARAMETERS_64;
 #endif
 
 /* from DDK2003SP1/3790.1830/inc/ddk/wnet/ntddk.h */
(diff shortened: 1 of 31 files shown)


derekbruening:
The diff is rather large and I was not expecting anyone to be able to do a detailed review as we don't really have any Windows experts as regular contributors anymore, but if anyone wants to take a look please go ahead.  Windows support is currently not in a great place and here I'm just trying to fill in some big missing pieces as best-effort improvements.

core/win32/ntdll.h
```diff
@@ -95,7 +95,21 @@
 #    define ATTACH_PARENT_PROCESS ((DWORD)-1)
 #endif
 
-#ifndef X64
+#ifdef X64
+typedef struct _UNICODE_STRING_32 {
+    /* Length field is size in bytes not counting final 0 */
```
johnfxgalea, line 100: Nit: Full stop

core/win32/ntdll_shared.c
```diff
@@ -72,7 +72,52 @@
...
+bool
+nt_write_virtual_memory(HANDLE process, void *base, const void *buffer,
+                        size_t buffer_length, size_t *bytes_written);
+
+#ifndef X64 /* Around most of the rest of the file. */
+
+#    if !defined(NOT_DYNAMORIO_CORE) && !defined(NOT_DYNAMORIO_CORE_PROPER)
+#        define UNPROT_IF_INIT()                                                 \
+            do {                                                                 \
+                /* The first call may not be during init so we have to unprot */ \
```
johnfxgalea, line 89: Nit: Full stop

johnfxgalea (approved):
Admittedly, a very light and quick review, but let me know if there is something particular you want me to have a closer look at.

derekbruening:
The Windows failures all seem to be flaky tests: grrr.

I also see that the sourceforge doxygen link for the older version is suddenly broken, and unfotunately it does not show up as a fatal error: the suite just doesn't build the docs.  I guess I will file a new issue.  Sigh -- so hard to keep stuff working.

In any case, that's all separate from the changes in this PR.  The win32.xarch test is passing beautifully on GA CI.

derekbruening:
> Admittedly, a very light and quick review, but let me know if there is something particular you want me to have a closer look at.

Thanks!  I guess my biggest concern is that this will break something we don't have good tests for, like AppInit injection (theoretically it shouldn't), or somehow cause issues on some variant of OS and app type (Windows 7, pre-Vista, Windows service, or graphical apps, or user switches, or sthg) that isn't going to show up on the GA CI: but it seems ok to put it in and rely on users filing issues if sthg breaks (and ideally adding new tests...).  Given that none of us is able to spend much time on Windows I think all we can do is put it in as best-effort support.  It certainly works well on my local machine and as shown here on GA and seems a very good step forward (users have asked for cross-arch injection before; will ask one of them to test it).

core/win32/ntdll.h
derekbruening: Done.

core/win32/ntdll_shared.c
derekbruening: Done.

derekbruening:
> The Windows failures all seem to be flaky tests: grrr.

We have burst_static #4486, vmareas.c assert in broadfun #4077, and win32.dotnet -- wait, that one actually may be a real bug: the image entry takeover point might be a problem there.  Good thing we actually have a test for that.  Investigating.

derekbruening:
So hooking the image entry for takeover worked on everything except .NET; hooking the thread start fixes .NET but seems to lead to strange instability in a handful of tests, including asserts about thread exit: maybe that hook is hit by early extra threads (though you'd think DR would just take over earlier in that case).  The weirdest is tool.cpuid.exe claiming the hook was written but not actually writing it and thus running native: no idea how that is happening, nor how to figure out further w/o walking through the kernel code.

Going to try one more thing before bailing and picking this up at some future point: setting the thread context, like thread injection does.

derekbruening:
OK it's green.  Just a final tweak which should not change stability.

Issue #4658, opened 2021-01-06
[AArch64] Fragment prefix size ASSERT failure with -shared_bb_ibt_tables


abhinav92003:
Running `simple_app` under DR with `-shared_bb_ibt_tables` option leads to the following assert failure:
```
$ bin64/drrun -shared_bb_ibt_tables   -- suite/tests/bin/simple_app
<Starting application dynamorio/suite/tests/bin/simple_app (95016)>
<Initial options = -no_dynamic_options -code_api -stack_size 56K -signal_stack_size 32K -max_elide_jmp 0 -max_elide_call 0 -shared_bb_ibt_tables -early_inject -emulate_brk -no_inline_ignored_syscalls -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct >

<Application dynamorio/suite/tests/bin/simple_app (95016).  Internal Error: DynamoRIO debug check failure: dynamorio/core/arch/aarch64/emit_utils.c:386 f->prefix_size == fragment_prefix_size(f->flags)
```
This is because [`fragment_prefix_size`](https://github.com/DynamoRIO/dynamorio/blob/6ec20cfd2b3c1eb986ec36a8d97d0caca93402a0/core/arch/emit_utils_shared.c#L1295) returns zero for some cases.

The assert failure goes away with `-bb_prefixes`. Probably just need to set default to true for AArch64, where fragments do have a prefix.

PR #4659, opened 2021-01-06
i#4658: Fix fragment_prefix_size on AArch64.

abhinav92003:
`fragment_prefix_size` returns an incorrect value on AArch64 when `-shared_bb_ibt_tables` is set. To fix this, we set the default value of `-bb_prefixes` to true on AArch64.

Fixes: #4658

### DIFF
--- a/core/options.c
+++ b/core/options.c
@@ -1234,6 +1234,7 @@ check_option_compatibility_helper(int recurse_count)
     bool changed_options = false;
 #    if defined(CLIENT_INTERFACE) && defined(AARCH64)
     if (!DYNAMO_OPTION(bb_prefixes)) {
+        USAGE_ERROR("bb_prefixes must be true on AArch64");
         dynamo_options.bb_prefixes = true;
         changed_options = true;
     }
--- a/core/options.c
+++ b/core/options.c
@@ -1232,6 +1232,12 @@ static bool
 check_option_compatibility_helper(int recurse_count)
 {
     bool changed_options = false;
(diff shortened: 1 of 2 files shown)


core/optionsx.h
```diff
@@ -497,7 +497,8 @@ OPTION_COMMAND(bool, opt_memory, false, "opt_memory",
                "enable memory savings at potential loss in performance", STATIC,
                OP_PCACHE_NOP)
 
-PC_OPTION_INTERNAL(bool, bb_prefixes, "give all bbs a prefix")
+PC_OPTION_DEFAULT_INTERNAL(bool, bb_prefixes, IF_AARCH64_ELSE(true, false),
```
derekbruening, line 500: This option cannot be disabled for A64, so it seems best to add an options.c rule to prevent that.

abhinav92003: Done.

core/options.c
```diff
@@ -1232,6 +1232,12 @@ static bool
 check_option_compatibility_helper(int recurse_count)
 {
     bool changed_options = false;
+#    if defined(CLIENT_INTERFACE) && defined(AARCH64)
+    if (!DYNAMO_OPTION(bb_prefixes)) {
+        dynamo_options.bb_prefixes = true;
```
derekbruening, line 1237: I think a USAGE_ERROR is called for here to let the user know, like in the other cases.

derekbruening (approved):

core/options.c
abhinav92003: Done.

PR #4661, opened 2021-01-07
i#4640: Pass static "client" faults to the app

### LINKED ISSUE #4640: Static-linked DR makes all SIGSEGV in exe fatal, thwarting app or client handlers
DR's signal handler has a check for SIGSEGV or SIGBUS:
```
        if (CLIENTS_EXIST() && is_in_client_lib(pc)) {
```

The problem is, that condition is always true for statically-linked DR.  The code goes on to raise a fatal error with no chance of continuing, which breaks an app or client with a handler and expected SIGSEGV or SIGBUS.

I think we just have to disable that for static DR and live without a nice crash message as we have no real way to distinguish further.

derekbruening:
A fault in the executable that has a statically-linked client and DR
is currently reported as a client crash, even though we have no way to
distinguish client code from app code in such a situation.  We change
that here to invoke execute_native_handler(), which solves problems
where the application has a fault handler in place and a tool or other
mechanism that generates faults during DR init or other points.

I tried to reproduce this by making a static version of
api.detach_signal but it requires a fault-generating allocator
replacement (hit by droption and other places that use the system
allocator) or something similar, which is non-trivial to replicate in
a small test.  I did test on the original proprietary application.

Fixes #4640

### DIFF
--- a/core/unix/signal.c
+++ b/core/unix/signal.c
@@ -5285,6 +5285,14 @@ master_signal_handler_C(byte *xsp)
             if (is_write && !is_couldbelinking(dcontext) && OWN_NO_LOCKS(dcontext) &&
                 check_for_modified_code(dcontext, pc, ucxt, target, true /*native*/))
                 break;
+#    ifdef STATIC_LIBRARY
+            /* i#4640: We cannot distinguish the client from DR or the app.  Though
+             * it's rare, original app instructions can come here for some app
+             * setups for static DR during init.  We send to the app handler.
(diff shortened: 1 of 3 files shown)


abhinav92003 (approved):

derekbruening:
Ah, the failures are api.static_crash, due to needing #1921 init-time-signal fixes I had in the same development tree when doing most of my testing.  I'll put those in first and then merge here.

Issue #4652, opened 2021-01-03
APP-CRASH pytorch (AVX2) illegal instruction under just drrun


andrewbartolo:
Hi all,

In the spirit of the recent AVX fixes, I've come across what seems to be another issue with AVX2 under just drrun (no client).

My execution environment:
CPU: Xeon E5-2683 v4 (Broadwell, AVX2)
Hypervisor: Proxmox (Debian 10) KVM in cpu-passthrough mode (CPUID shows native CPU in VM)
VM OS: Debian 11 (also tried with Ubuntu 20.04)
DR build: bbf41a8 (latest)
Runtime language: Python 3.9.1
Application: PyTorch 1.7.1

The application works fine without drrun. However, when I invoke `drrun -- python3 resnet-evaluate.py`, drrun stops with `Illegal instruction`. The application also works fine under drrun if I set KVM to emulate a Sandy Bridge (AVX1) CPU instead, as PyTorch auto-detects CPUID features and then runs with the highest-supported level. I'm unsure if PyTorch falls back to AVX1 or all the way back to SSE instructions on Sandy Bridge, but it no longer throws `Illegal instruction`.

I built DR in debug mode, and got the following backtrace:

```
Signal 4 delivered to application as default action.
Callstack:
	0x00007f0c5c661990   <.../lib/python3.9/site-packages/torch/lib/libtorch_cpu.so+0x3a6d990>
	0x00007f0c59916894   <.../lib/python3.9/site-packages/torch/lib/libtorch_cpu.so+0xd22894>
	0x00007f0c59916d85   <.../lib/python3.9/site-packages/torch/lib/libtorch_cpu.so+0xd22d85>
	0x00007f0c59909b47   <.../lib/python3.9/site-packages/torch/lib/libtorch_cpu.so+0xd15b47>
	0x0000000000000000  
	0x0000000000000001  
	0x00007f0c1b514100   <.../lib/python3.9/site-packages/torch/lib/libc10.so+0x3f100>
	0x002378b8058b4808  
>
<Stopping application /usr/bin/python3.9 (17212)>
Illegal instruction
```

After objdumping libtorch_cpu.so, I found:

```
3a6d985:	0c 0a 
3a6d987:	c5 fc 59 c2          	vmulps %ymm2,%ymm0,%ymm0
3a6d98b:	c4 e2 55 98 c3       	vfmadd132ps %ymm3,%ymm5,%ymm0
[...]

PR #4667, opened 2021-01-09
Stop searching for additional prefixes in decode_cti after a c5 VEX prefix is found

khuey:
When decoding the two-byte c5 VEX prefix, currently the second byte of the prefix is not skipped and is instead treated as its own prefix. If the second byte is, say, c4, that will trigger the three-byte c4 VEX prefix code path which will cause decode_cti to fail spuriously. Since VEX prefixes are always the final prefix, just copy what we do for c4 VEX and EVEX prefixes and stop looking for additional prefixes after we see a c5 VEX prefix.

### DIFF
--- a/core/ir/x86/decode_fast.c
+++ b/core/ir/x86/decode_fast.c
@@ -1359,6 +1359,9 @@ decode_cti(dcontext_t *dcontext, byte *pc, instr_t *instr)
                 /* VEX 2-byte prefix implies 0x0f opcode */
                 byte0 = 0x0f;
                 byte1 = *(pc + prefixes);
+                /* There are no prefixes after vex. */
+                pc = start_pc + prefixes;
+                i = prefixes;
                 break;
             case EVEX_PREFIX_OPCODE:
                 instr_set_prefix_flag(instr, PREFIX_EVEX);
--- a/suite/tests/api/ir_x86.c
+++ b/suite/tests/api/ir_x86.c
@@ -888,7 +888,7 @@ test_cti_prefixes(void *dc)
     ASSERT(decode_next_pc(dc, buf) == (byte *)&buf[6]);
 
 #if defined(DEBUG) && defined(BUILD_TESTS)
-    /* PR 4667, a c5 vex encoding where the second byte is c4
+    /* Issue 4652, a c5 vex encoding where the second byte is c4
      *   c5 c4 54 2d 68 1d 0c 0a vandps 0xa0c1d68(%rip),%ymm7,%ymm5
      */
     buf[0] = 0xc5;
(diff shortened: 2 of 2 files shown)


core/ir/x86/decode_fast.c
```diff
@@ -1359,6 +1359,9 @@ decode_cti(dcontext_t *dcontext, byte *pc, instr_t *instr)
                 /* VEX 2-byte prefix implies 0x0f opcode */
                 byte0 = 0x0f;
                 byte1 = *(pc + prefixes);
+                /* There are no prefixes after vex. */
+                pc = start_pc + prefixes;
+                i = prefixes;
```
derekbruening, line 1364: Thanks for looking into this.  The fix makes sense.  I'm wondering why none of the api.ir tests hit this -- I guess none happen to have the right byte sequence.  Could we add a test that hits this?  Even just one case in api.ir would be good to have a regression test.

khuey: Nothing in apir.ir or in the binutils tests has the right byte sequence (I tried adding a decode_cti check to the latter but it doesn't trip).

derekbruening: Add a vandps entry to api.ir with the same operands as the culprit in #4652 (looks like i#4652 can go in the title and have 
Fixes here)?  `vandps 0xa0c1d68(%rip),%ymm7,%ymm5`

khuey: Alright that should do it.

suite/tests/api/ir_x86.c
```diff
@@ -886,6 +886,31 @@ test_cti_prefixes(void *dc)
...
+     *   c5 c4 54 2d 68 1d 0c 0a vandps 0xa0c1d68(%rip),%ymm7,%ymm5
+     */
+    buf[0] = 0xc5;
+    buf[1] = 0xc4;
+    buf[2] = 0x54;
+    buf[3] = 0x2d;
+    buf[4] = 0x68;
+    buf[5] = 0x1d;
+    buf[6] = 0x0c;
+    buf[7] = 0x0a;
```
derekbruening, line 901: OK.  I was thinking of a singe line that's a variant of one of these:
```
suite/tests/api/ir_x86_3args_avx.h:OPCODE(vandps_256, vandps, vandps, 0, REGARG(YMM0), REGARG(YMM1), MEMARG(OPSZ_32))
```
But I didn't look at what it takes to make it 0xc4: are there multiple encoding choices and DR doesn't ever pick the 0xc4 one?

In any case this should be sufficient.  Note that normally we would be point at the issue #4652 rather than the fixing PR.

derekbruening (approved):
Thanks again for taking the time to look into this.

suite/tests/api/ir_x86.c
khuey: Ah I didn't try that.

Replaced the PR # with the Issue #.

Issue #4669, opened 2021-01-12
Translation failures and asserts from client calling dr_app_pc_from_cache_pc on bb prefixes and rseq mangling


derekbruening:
We have a client who has an itimer which calls dr_app_pc_from_cache_pc() on the PC on each sample.  This results many instances of a curiosity, and some of an assert:
```
<CURIOSITY : tdcontext != get_thread_private_dcontext() || (((OPTION_IS_INTERNAL_stress_recreate_pc)) ? ((((void)(((dynamo_options.checklevel >= (1)) && !(!((OPTION_IS_STRING_stress_recreate_pc)) || (((&options_lock)->num_readers > 0) || self_owns_write_lock(&options_lock)))) ? (d_r_internal_error("third_party/dynamorio/trunk/core/translate.c", 907, "!((OPTION_IS_STRING_stress_recreate_pc)) || READWRITE_LOCK_HELD(&options_lock)"), 0) : 0)), dynamo_options.stress_recreate_pc)) : ((((dynamo_options.checklevel >= (1)) && !((0))) ? (d_r_internal_error("non-internal option argument " "stress_

Internal Error: DynamoRIO debug check failure: core/translate.c:991 tdcontext != get_thread_private_dcontext() || INTERNAL_OPTION(stress_recreate_pc) || TEST(FRAG_SELFMOD_SANDBOXED, flags) || TEST(FRAG_WAS_DELETED, flags)
```

Investigating we see that the translation points are always either in bb prefixes (we're running -disable_traces, but on arm every bb has a prefix) or in rseq mangling code.

Action items:
+ Relax these asserts for client requests: these aren't synchronous faults.
+ Ideally, add full xl8 support for all parts of rseq mangling and for prefixes. It improves performance (fewer synchall retries) and helps things like these client samples.
+ Print out DR's base at the top of these assert message callstacks so we can symbolize them more easily.

derekbruening:
For adding translation to prefixes I'm considering 3 approaches:

1) xl8 code emits raw to buffer, decodes, and adds instrs to list.

2) Change how prefixes are emitted in first place to IR instead of raw bytes. This would result in simpler and cleaner prefix code. But it is a bigger change and carries risk of destabilizing sthg since it affects what a recreated ilist looks like; it is definitely more work; it has a perf impact on emit by adding a bunch of encoding, which esp w/ no tool will be a noticeable bb build time increase over a bb which only needs to encode its final jump.

[...]

PR #4672, opened 2021-01-13
i#4669 xl8 failures: No asserts for client xl8 requests

derekbruening:
Applies the existing client_data->is_translating debug flag to avoid a
curiosity and an assert on translation failure stemming from client
use of dr_app_pc_from_cache_pc() on a sampling itimer.

Augments the client.timer test to run with -disable_traces and use a
lot of indirect branches to trigger a curiosity on xl8 failure on
every run when the iterations are increased (a little too high for our
small regression tests though).  With this fix, the curiosity never
occurs.

Issue: #4669

### DIFF
--- a/suite/tests/client-interface/timer.dll.c
+++ b/suite/tests/client-interface/timer.dll.c
@@ -1,5 +1,5 @@
 /* **********************************************************
- * Copyright (c) 2018 Google, Inc.  All rights reserved.
+ * Copyright (c) 2018-2021 Google, Inc.  All rights reserved.
  * Copyright (c) 2010 VMware, Inc.  All rights reserved.
  * **********************************************************/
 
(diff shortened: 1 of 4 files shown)


derekbruening:
coherence test failure is a flake: #4673.

suite/tests/client-interface/timer.dll.c
```diff
@@ -49,6 +49,11 @@
  */
 static int buckets[DR_WHERE_LAST];
 
+/* Currently only written to but we may add checks on this once i#4669 failures
```
abhinav92003, line 52: Worth adding an `ASSERT_CURIOSITY` for now?

abhinav92003 (approved):

suite/tests/client-interface/timer.dll.c
derekbruening: You mean a client version of the core's ASSERT_CURIOSITY?  I think that will just lead to a flaky test if we print something out while we have known failures, right?

abhinav92003: Oh okay. I don't know whether the `stderr_mask` can filter out those messages, or if we want to do that, or if it's worth it.

derekbruening: -stderr_mask filters messages from DR.  For this client we would have to make the expected output file ignore it or something.  Since I was thinking of fixing these failures soon I'm inclined to leave it.

Issue #4487, opened 2020-10-14
[AArch64] Inline instr counting for trace_after_instrs in drcachesim


abhinav92003:
Today for AArch64, instr counting for the `-trace_after_instrs` option in drcachesim is not inlined. Instead it's done using a clean call. https://github.com/DynamoRIO/dynamorio/blob/5cbe8113c919b557f7bfc3daafffc9b1fef7938c/clients/drcachesim/tracer/tracer.cpp#L1430

We should inline it for better performance.

derekbruening:
This involves implementing drx_insert_counter_update() for 64-bit counters for AArchXX which was previously under the general #1551.

PR #4677, opened 2021-01-18
i#4487: inline instr count for trace_after_instrs in drcachesim, AArch64

sapostolakis:
Inlines instruction counting for the `-trace_after_instrs` option in drcachesim on
AArch64 to avoid doing a clean call and thus improve performance.
Involves enabling `drx_insert_counter_update()` for 64-bit counters for AArch64.

Extends two tests with checks for 64-bit counter updates on AArch64.

Fixes: #4487


derekbruening:
@AssadHashmi -- If you could add @sapostolakis to the list for auto-triggering the aarch64 Jenkins.  Though I thought any project member would trigger?

derekbruening:
run arm tests

ext/drx/drx.c
```diff
@@ -556,6 +558,7 @@ drx_insert_counter_update(void *drcontext, instrlist_t *ilist, instr_t *where,
      * address being near this one, and add to reg1 instead of
      * taking 2 instrs to load it fresh.
      */
+    /* update counter */
```
derekbruening, line 561: Hmm, it seems like for our use case at least we would want release-acquire semantics to ensure writes are visible promptly and in order to other threads, even though we're ok not having a pure atomic fetch-and-add and thus having some races.  I think the existing 32-bit arm code would want that too.  I would vote for just putting in the release-acquire, though you could argue for a 3rd API mode to let clients use completely relaxed models.  This makes the racy counter update behave like the x86 one, which at least matches my intuition for how it should operate.

So I'm voting for `ldar` and `stlr` for a64 and `dmb ish;str` and `ldr;dmb ish` for a32, and mentioning the release-acquire in the docs.

derekbruening: Or are there common cases of clients incrementing thread-private values and they don't want to pay for barriers?  Maybe get some more opinions: @johnfxgalea?

AssadHashmi:
> @AssadHashmi -- If you could add @sapostolakis to the list for auto-triggering the aarch64 Jenkins.

Added @sapostolakis to AArch64 CI Jenkins ACL.

> Though I thought any project member would trigger?

Not yet. It's a manual process ATM.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1469,20 +1471,58 @@ event_delay_app_instruction(void *drcontext, void *tag, instrlist_t *bb, instr_t
...
+                                   (dr_spill_slot_t)(SPILL_SLOT_MAX + 1), &instr_count,
+                                   num_instrs, DRX_COUNTER_64BIT))
+        DR_ASSERT(false);
+
+    reg_id_t scratch1, scratch2, scratch3;
+    if (drreg_reserve_register(drcontext, bb, instr, NULL, &scratch1) != DRREG_SUCCESS ||
+        drreg_reserve_register(drcontext, bb, instr, NULL, &scratch2) != DRREG_SUCCESS ||
+        drreg_reserve_register(drcontext, bb, instr, NULL, &scratch3) != DRREG_SUCCESS)
+        FATAL("Fatal error: failed to reserve scratch registers");
+    dr_save_arith_flags_to_reg(drcontext, bb, instr, scratch3);
```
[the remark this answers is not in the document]
sapostolakis: From an initial exploration, I found that the problem with drreg_reserve_aflags() is that at some point (for a long run where `-trace_after_instrs` is set to a very large value) the aflags are not restored. By forcing aflags to restore in drreg_insert_restore_all()  (essentially by replacing the if statement in https://github.com/DynamoRIO/dynamorio/blob/f5016906b5699773acf52dd7f5147da89f10ae12/ext/drreg/drreg.c#L411 with  just `if (!pt->aflags.native)`) , the issue is resolved and I can use drreg_reserve_aflags(). Now the question is at what point the aflags should have been restored and didn't, and which of the conditions in lines 412-418 should have been true and weren't.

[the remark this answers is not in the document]
johnfxgalea: What is the problem exactly? Do you get a crash? From my experience, this issue often arises more due to a bug exposed by the instructions in a particular basic block, rather than the length of the trace. For example, drreg mistakenly expects a given instruction to overwrite eflags, and therefore restoration is not done. 

Such problems can be a bit of a pain to debug. If you do get a crash, you can ask DR to output logs (with log level 3 or above), and inspect the few basic blocks right before the crash to maybe find anything suspicious.

ext/drx/drx.c
johnfxgalea: I could see cases where one does not need barriers for thread-private counters, but given the simplicity of increments, one can easily go around this problem by not using this API. Besides, the 3rd API mode that you suggested earlier might also accept a flag to avoid barriers if we need this in the future.

johnfxgalea: Maybe add a new flag as DRX_COUNTER_LOCK is for atomic behaviour.

clients/drcachesim/tracer/tracer.cpp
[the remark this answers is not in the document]
sapostolakis: true the error is irrelevant to the length of the trace. The bug exhibited in a big workload after million of instructions probably because of some specific sequence of instructions that is not that common. I have not been able to trigger a similar issue in smaller benchmarks or any test.

The problem is that the program takes at some point some unexpected control flow and is eventually led to an application-level assertion down the line. I looked at the logs but they are too long and there is nothing suspicious at the last few basic blocks. Probably the clobbered aflags affect something much earlier than the end. I cannot pinpoint where things went wrong, but I might be able to find out where the aflags were not restored and look into those cases. That said, so far I have been more successful looking at the drreg code and trying to reason about its correctness rather than investigating the problematic run.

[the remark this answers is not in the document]
derekbruening: Comparing the block tags between a bad and good run, w/ ASLR or other nondet sources turned off, is one method we have used to find divergence points, but that still requires walking backward from the differing conditional branch to the block where the bad flags were produced: but for flags this should be quite close, as opposed to GPR's.

Here, adding flags checks to ensure the app flags value is the same before the instru and at the next sink point (app flags read or end of block) should help.  Xref #4678 on library support to make such things easier.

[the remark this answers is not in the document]
sapostolakis: Got it. I will try comparing the block tags to figure out the issue.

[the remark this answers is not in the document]
sapostolakis: It took some time but this approach worked. There were various points of divergence without any noticeable error related to flags. Divergence was caused by content differences in general-purpose registers. Thankfully, I noticed several basic blocks before a point of divergence unrecognized instructions (`xx`) that executed before the flags restoration. If any of these instructions read/wrote the flags then the application would behave incorrectly. I filed an issue. See #4687 for more info. With the #4688 fix, the issue is resolved for AArch64.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1320,10 +1320,10 @@ static bool tracing_enabled;
 static volatile bool tracing_scheduled;
 static void *schedule_tracing_lock;
 
-#ifdef X86_64
+#if defined(X86_64) || defined(AARCH64)
 #    define DELAYED_CHECK_INLINED 1
 #else
-// XXX i#4487: we don't have the inlining implemented yet.
+/* XXX we don't have the inlining implemented yet for ARM_32 */
```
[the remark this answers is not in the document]
sapostolakis: done.

clients/drcachesim/tracer/tracer.cpp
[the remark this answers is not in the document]
sapostolakis: With the #4688 fix merged, I was able to use `drreg_reserve_aflags()`.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1469,20 +1471,58 @@ event_delay_app_instruction(void *drcontext, void *tag, instrlist_t *bb, instr_t
...
+        drreg_unreserve_register(drcontext, bb, instr, scratch3) != DRREG_SUCCESS)
+        DR_ASSERT(false);
+#        endif
 #    else
 #        error NYI
 #    endif
 #else
-    // XXX: drx_insert_counter_update doesn't support 64-bit, and there's no
-    // XINST_CREATE_load_8bytes.  For now we pay the cost of a clean call every time.
+    /* XXX: drx_insert_counter_update doesn't support 64-bit counters for ARM_32
```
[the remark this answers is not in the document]
sapostolakis: actually i386 supports 64-bit counter updates but not inlining of check_instr_count_threshold. Clarified the missing part for i386 and addressed the nit.

ext/drx/drx.c
```diff
@@ -528,8 +528,10 @@ drx_insert_counter_update(void *drcontext, instrlist_t *ilist, instr_t *where,
         }
     }
 #elif defined(AARCHXX)
+#    ifdef ARM_32
```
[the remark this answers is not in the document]
sapostolakis: done.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1469,20 +1471,58 @@ event_delay_app_instruction(void *drcontext, void *tag, instrlist_t *bb, instr_t
...
+                              OPND_CREATE_MEMPTR(scratch1, 0)));
+    instrlist_insert_mov_immed_ptrsz(drcontext, op_trace_after_instrs.get_value(),
+                                     opnd_create_reg(scratch1), bb, instr, NULL, NULL);
+    MINSERT(bb, instr,
+            XINST_CREATE_cmp(drcontext, opnd_create_reg(scratch2),
+                             opnd_create_reg(scratch1)));
+    MINSERT(bb, instr,
+            XINST_CREATE_jump_cond(drcontext, DR_PRED_LT, opnd_create_instr(skip_call)));
+#        endif
+
```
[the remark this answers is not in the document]
sapostolakis: Added restoring of scratch registers and aflags on the clean call path.

ext/drx/drx.c
sapostolakis: as suggested, updated counter updates for 32-bit and 64-bit arm with release-acquire semantics and added a flag for optional completely relaxed model. Also, updated the description in the header with info about the new flag.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1448,12 +1448,16 @@ event_delay_app_instruction(void *drcontext, void *tag, instrlist_t *bb, instr_t
...
+    instr_t *skip_call = INSTR_CREATE_label(drcontext);
+#        ifdef X86_64
     if (!drx_insert_counter_update(drcontext, bb, instr,
                                    (dr_spill_slot_t)(SPILL_SLOT_MAX + 1) /*use drmgr*/,
                                    &instr_count, num_instrs, DRX_COUNTER_64BIT))
         DR_ASSERT(false);
-    instr_t *skip_call = INSTR_CREATE_label(drcontext);
+
+    if (drreg_reserve_aflags(drcontext, bb, instr) != DRREG_SUCCESS)
+        FATAL("Fatal error: failed to reserve aflags");
```
sapostolakis, line 1460: Reserve aflags before clobbering cmp instruction on x86.

ext/drx/drx.h
```diff
@@ -132,8 +132,9 @@ drx_aflags_are_dead(instr_t *where);
 
 /** Flags for \p drx_insert_counter_update */
 enum {
-    DRX_COUNTER_64BIT = 0x01, /**< 64-bit counter is used for update. */
-    DRX_COUNTER_LOCK = 0x10,  /**< Counter update is atomic. */
+    DRX_COUNTER_64BIT = 0x01,    /**< 64-bit counter is used for update. */
+    DRX_COUNTER_LOCK = 0x10,     /**< Counter update is atomic. */
+    DRX_COUNTER_REL_ACQ = 0x100, /**< Release-acquire semantics for counter update. */
```
derekbruening, line 137: Hmm, looks like DRX_COUNTER_LOCK was added with a mistaken not-next-bit value.  Best to leave it for compatibilty, but may as well take 0x02 for the new one.

derekbruening, line 137: Please add the new feature to the feature changelist at https://github.com/DynamoRIO/dynamorio/blob/master/api/docs/release.dox#L192

ext/drx/drx.h
```diff
@@ -159,8 +160,10 @@ DR_EXPORT
  * \note The counter update is racy (i.e., not synchronized among threads)
  * unless #DRX_COUNTER_LOCK is specified in \p flags. When #DRX_COUNTER_LOCK
  * is set, the instrumentation may fail if a 64-bit counter is updated in
- * a 32-bit application or the counter crosses cache lines. Currently,
- * #DRX_COUNTER_LOCK is not yet supported on ARM.
+ * a 32-bit application or the counter crosses cache lines. Currently, #DRX_COUNTER_LOCK
+ * is not yet supported on ARM. For ARM, if #DRX_COUNTER_REL_ACQ is specified in \p flags,
```
derekbruening, line 164: Maybe best to s/ARM/AArchXX/ so it's clear it's both 32-bit and 64-bit ("ARM" means just 32-bit in many contexts, though not in others so it can be ambiguous)?

suite/tests/client-interface/drxmgr-test.dll.c
```diff
@@ -69,6 +75,10 @@ event_app_instruction(void *drcontext, void *tag, instrlist_t *bb, instr_t *inst
                               IF_NOT_X86_(SPILL_SLOT_MAX + 1) & counterA, 1, 0);
     drx_insert_counter_update(drcontext, bb, inst, SPILL_SLOT_MAX + 1,
                               IF_NOT_X86_(SPILL_SLOT_MAX + 1) & counterB, 3, 0);
+#if defined(AARCH64)
+    drx_insert_counter_update(drcontext, bb, inst, SPILL_SLOT_MAX + 1, SPILL_SLOT_MAX + 1,
+                              &counterC, 3, DRX_COUNTER_64BIT);
```
derekbruening, line 80: Please add at least one test using DRX_COUNTER_REL_ACQ to exercise that path in some fashion.

ext/drx/drx.c
```diff
@@ -556,16 +558,46 @@ drx_insert_counter_update(void *drcontext, instrlist_t *ilist, instr_t *where,
      * address being near this one, and add to reg1 instead of
      * taking 2 instrs to load it fresh.
      */
+    /* update counter */
```
derekbruening, line 561: nit: In new comments use capitalization, punctuation, and complete sentences (you may see old comments not following this; it is a style change more recent than some old code)

clients/drcachesim/tracer/tracer.cpp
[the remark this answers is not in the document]
derekbruening: Hmm yeah I guess it does take some effort for the 64-bit inlined memop.  OK.

suite/tests/client-interface/drxmgr-test.dll.c
sapostolakis: yep, sorry about that. I should have added a test case for that. 
Added a few additional cases for both 32-bit and 64-bit arm to exercise all the available combinations of flags.

ext/drx/drx.c
sapostolakis: Got it.

ext/drx/drx.h
sapostolakis: Done.

ext/drx/drx.h
sapostolakis: Done.

sapostolakis: Yeah it was weird but I thought I should follow the convention. I changed the new flag to 0x02.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1469,20 +1473,76 @@ event_delay_app_instruction(void *drcontext, void *tag, instrlist_t *bb, instr_t
...
+                                     opnd_create_reg(scratch1), bb, instr, NULL, NULL);
+    MINSERT(bb, instr,
+            XINST_CREATE_cmp(drcontext, opnd_create_reg(scratch2),
+                             opnd_create_reg(scratch1)));
+    MINSERT(bb, instr,
+            XINST_CREATE_jump_cond(drcontext, DR_PRED_LT, opnd_create_instr(skip_call)));
+#        endif
+
+    /* hit_instr_count_threshold does not always return. Restore scratch registers and
+     * aflags.*/
```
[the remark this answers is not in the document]
sapostolakis: Done.

api/docs/release.dox
```diff
@@ -190,6 +190,8 @@ Further non-compatibility-affecting changes include:
  - Added instr_is_exclusive_load().
  - Added XINST_CREATE_store_pair() and XINST_CREATE_load_pair().
  - Added instr_replace_reg_resize() and opnd_replace_reg_resize().
+ - Added DRX_COUNTER_REL_ACQ flag to optionally enable release-acquire semantics for
```
derekbruening, line 193: Use a `#` prefix (so `#DRX_COUNTER_REL_ACQ`) to have Doxygen generate an html link (`()` does it for functions but need `#` for enums) as well as serving as a sanity check that there's no typo (Doxygen will generate an error if there is no such symbol with that name).

derekbruening (approved):

api/docs/release.dox
sapostolakis: I see. Done.

Issue #4687, opened 2021-01-23
[APP CRASH, AArch64] aflags not properly restored using drreg API due to incomplete decoder


sapostolakis:
Use of drreg's API ( `drreg_reserve_aflags()` and `drreg_unreserve_aflags()`) to save and restore aflags on AArch64 clobbered the aflags for an application. The control flow under DR was different than the control flow on a native run leading to an unexpected application-level assertion. The bug exhibited after a couple of millions of instructions and DR was run with these arguments: `-unsafe_build_ldstex -disable_traces -vm_size 2G -no_enable_reset -t drcachesim -offline -trace_after_instrs 3000G -exit_after_tracing 20G`.

As an optimization, drreg performs restoration of aflags as late as possible to minimize spills/fills. This optimization works fine for all the recognized flags-sensitive (read/write flags) instructions in the application that I checked. However, for AArch64 the decoder is incomplete and there are unrecognized instructions decoded as instances of a generic instruction, OP_xx. It seems that the placement of the aflags restoration by drreg does not take into account these unrecognized instructions and there are multiple cases like the following:

```
m4  ...   subs   ...
L3  ...   xx     ...
m4  ...   msr    %x1 -> %nzcv
 ```

In this example, DR clobbers the aflags (with `subs`), then an unrecognized instruction of the application (`xx`) that may read/write aflags operates on the clobbered aflags, and then the flags are overwritten (`msr`) nullifying any effect of the unrecognized instruction on the flags. Instead, the flags should have been restored prior to the `xx` instruction.

derekbruening:
So this is in some sense a duplicate of #2626.  Having an incomplete decoder was presumed to be a temporary condition.  Since the missing instructions could cause many other problems in liveness and other analyses of GPR's (including drreg problems with GPR's akin to the aflags here), of SIMD registers for clean call state preservation analysis as well as drreg, and are a non-starter for taint analysis which needs to know every operand: it seems #2626 should have its priority raised.

Issue #2626, opened 2017-08-30
Finish AArch64 encoder/decoder


egrimley:
https://github.com/DynamoRIO/dynamorio/wiki/AArch64-Port: "At the end of 2016, DynamoRIO's encoder/decoder handles all the load/store instructions, including load/store of FP/SIMD registers, and all the instructions that do not operate on FP/SIMD registers, up to ARMv8.2."

The non-load/store FP/SIMD instructions should be added.

See #1569, #2440 and #2443.

egrimley:
I'll just mention something here that I don't think is stated anywhere else:

The `OP_xx` trick, decoding an unrecognised instruction in a defensive way, guessing that it might read and write the general-purpose registers referenced in the four places in the instruction word where the number of a general-purpose register typically appears, should perhaps be retained, at least as on option, even when the decoder is believed to be complete. It would let DynamoRIO correctly handle, in many cases, instructions that have been added to the architecture since DynamoRIO's decoder was last updated.

fhahn:
I pushed a first version of the script I used to generate everything for #2951 to the [project-aarch64-generate-patterns](https://github.com/DynamoRIO/dynamorio/tree/project-aarch64-generate-patterns) branch. For completing support for NEON, we should extend those scripts to generate everything for the missing instructions.

AssadHashmi:
This issue describes one user's experiences #3700 
(Initial user group posting https://groups.google.com/forum/#!topic/dynamorio-users/F5Lu4iNtiJk)

derekbruening:
Xref #3044 on SVE

derekbruening:
Xref testing improvements: #2443, #4396

derekbruening:
Xref improving the underlying design to add more flexibility for things like splitting opcodes: #4393

derekbruening:
Missing flags effects on unknown instructions are causing mysterious crashes requiring time to track down: #4687.  Missing GPR effects will cause similar problems.  It seems the missing instructions in the decoder are now a blocker to further scaling up our AArch64 usage.  Raising priority.

derekbruening:
[...]

PR #4688, opened 2021-01-23
i#4687: conservatively handle aflags for unrecognized (OP_xx) instructions on AArch64

sapostolakis:
Conservatively restore aflags before unrecognized instructions in case they read/write the flags.

Fixes: #4687


johnfxgalea:
Just a quick thought, can unrecognized instructions also affect badly the restoration of gprs, and not just aflags?

derekbruening:
> Just a quick thought, can unrecognized instructions also affect badly the restoration of gprs, and not just aflags? For example, failing to restore a register before the execution of an unrecognised instruction that uses that same register as a source?

Yes, and for some uses cases missing SIMD usage would cause problems (such as clean call optimizations: though maybe today a64 does not perform those?  as well as in other types of tools).  One mitigating factor may be that the decoder of OP_xx looks at the typical register operand slots and fills in what GPR's it predicts the instruction might use?  I'm basing that on https://github.com/DynamoRIO/dynamorio/issues/2626#issuecomment-351976969 from @egrimley -- I have not checked whether these OP_xx insructions actually have any operands?  Do they?

derekbruening:
Ideally the fix here would be to add the missing instructions to the decoder (#2626).  Perhaps we should try to understand whether there is an ETA on that in order to figure out how to proceed.  We would like to avoid having to spend time tracking down any more mysterious bugs that turn out to be #2626 again.  I believe @AssadHashmi planned to work on #4393 and maybe the plan was to wait for that before adding missing opcodes?

ext/drreg/drreg.c
```diff
@@ -410,7 +410,12 @@ drreg_insert_restore_all(void *drcontext, instrlist_t *bb, instr_t *inst,
     uint aflags = (uint)(ptr_uint_t)drvector_get_entry(&pt->aflags.live, pt->live_idx);
     if (!pt->aflags.native &&
         (force_restore ||
-         TESTANY(EFLAGS_READ_ARITH, instr_get_eflags(inst, DR_QUERY_DEFAULT)) ||
+         /* On AArch64, unrecognized instructions are decoded as instances of a generic
+          * instruction, OP_xx. For these instructions conservatively restore aflags in
+          * case they read/write the aflags.
+          */
+         IF_AARCH64(instr_get_opcode(inst) == OP_xx ||)
```
derekbruening, line 417: Given that all of the missing opcodes are vector instructions, I would expect a very small fraction to read or write flags (or GPR's for that matter).  I wonder if adding the couple of opcodes that affect flags and GPR's would not be much work.  Though it's certainly more work than this workaround.  Let's wait for advice from those more familiar with the a64 decoder: @AssadHashmi?

ext/drreg/drreg.c
```diff
@@ -410,7 +410,12 @@ drreg_insert_restore_all(void *drcontext, instrlist_t *bb, instr_t *inst,
     uint aflags = (uint)(ptr_uint_t)drvector_get_entry(&pt->aflags.live, pt->live_idx);
     if (!pt->aflags.native &&
         (force_restore ||
-         TESTANY(EFLAGS_READ_ARITH, instr_get_eflags(inst, DR_QUERY_DEFAULT)) ||
+         /* On AArch64, unrecognized instructions are decoded as instances of a generic
+          * instruction, OP_xx. For these instructions conservatively restore aflags in
+          * case they read/write the aflags.
+          */
```
derekbruening, line 416: A workaround like this should be labeled as a temporary thing we do not want to leave in the code base long-term: so I would expect a XXX or TODO comment with an issue number covering the work to put in the permanent fix.

ext/drreg/drreg.c
```diff
@@ -410,7 +410,12 @@ drreg_insert_restore_all(void *drcontext, instrlist_t *bb, instr_t *inst,
...
+         IF_AARCH64(instr_get_opcode(inst) == OP_xx ||)
+             TESTANY(EFLAGS_READ_ARITH, instr_get_eflags(inst, DR_QUERY_DEFAULT)) ||
```
derekbruening, line 418: This is only handling reading flags: there are also points where drreg needs to update a spilled aflags value after an app instruction *writes* to the flags.  That point would need a workaround as well.

sapostolakis: I thought that this workaround handles writing flags as well. My understanding is that even if an unrecognized application instruction (`xx`) writes to the flags, this workaround will force the flags to be restored before `xx`, effectively preventing insertion of a restore after `xx` that overwrites the write to the flags by `xx`. If there is another save/restore pair by drreg after the 'xx' inst, it will not change the application view of the flags. Are you thinking of something else or am I missing something?

ext/drreg/drreg.c
sapostolakis: I added a XXX comment linking to issue #2626 in case we decide to go ahead with this workaround till the decoder is more complete.

ext/drreg/drreg.c
derekbruening: Restoring is not unreserving.  drreg will assume that the app's flags value is still stored in TLS after an app read but not write of the flags, which will be incorrect after `xx` writes to the flags and changes that value.  drreg will proceed to use the wrong value when restoring on future reads or unreserves.

sapostolakis: I see.

ext/drreg/drreg.c
AssadHashmi: > Given that all of the missing opcodes are vector instructions, I would expect a very small fraction to read or write flags (or GPR's for that matter). I wonder if adding the couple of opcodes that affect flags and GPR's would not be much work. Though it's certainly more work than this workaround. Let's wait for advice from those more familiar with the a64 decoder: @AssadHashmi?

The resourcing planned to address AArch64 decoder gaps has been diverted elsewhere at the present time. Reluctantly, I'd advise carrying on with the workaround. This is far from ideal for AArch64 support but we currently have to work within limited constraints. Hopefully the situation will change later in the year.

derekbruening: Thinking more about the implications here: workarounds will have to added to multiple places and not just drreg.  The drcachesim tracer does GPR analysis to elide same-base addresses.  The memval tool looks at register usage.  Dr. Memory, being ported now, looks at everything.  (The core does flags analysis for fragment prefixes and stay-on-trace comparisons -- though I believe these have no problem today due to flags-less IBL and no trace support yet for AArch64.)

@AssadHashmi is it the case that OP_xx insructions have GPR operands in place as hinted at here https://github.com/DynamoRIO/dynamorio/issues/2626#issuecomment-351976969?  Does that reliably cover all GPR usage in such instructions?

AssadHashmi: > @AssadHashmi is it the case that `OP_xx` instructions have GPR operands in place as hinted at here #2626 (comment)?

@derekbruening yes I believe that is the case. @egrimley can confirm.

> Does that reliably cover all GPR usage in such instructions?

The majority, if not all, `OP_xx` instructions are SIMD. Of those, GPR usage is restricted to an index register. Based on that you should be reliably covered. There may be cases of non-SIMD `OP_xx` but even then you should be ok.

sapostolakis:
As per @derekbruening suggestion I added aflags as read and written to every OP_xx at the point of decoding. This addresses the issue in #4687 and should address any other undiscovered issue regarding usage of aflags by unrecognized instructions.

core/ir/aarch64/codec.c
```diff
@@ -3488,6 +3488,16 @@ decode_common(dcontext_t *dcontext, byte *pc, byte *orig_pc, instr_t *instr)
          */
         eflags |= EFLAGS_WRITE_NZCV;
     }
+
```
derekbruening, line 3491: Ah, so right up there it lists OPfcsel, OP_fccmp, etc.: probably you hit one of those.

Optional: s/1569/2626/ for those two flags cases right above (1569 was the general container; everything at this point should move to a specific issue) (this interface won't let me attach a comment up there).

core/ir/aarch64/codec.c
```diff
@@ -3488,6 +3488,16 @@ decode_common(dcontext_t *dcontext, byte *pc, byte *orig_pc, instr_t *instr)
...
+    /* XXX i#2626: Until the decoder for AArch64 covers all the instructions that
+     * read/write aflags, as a workaround conservatively assume that all OP_xx
+     * instructions (i.e., unrecognized insructions) may read/write aflags
```
derekbruening, line 3494: nit: missing trailing `.`

derekbruening (approved):

core/ir/aarch64/codec.c
sapostolakis: done

core/ir/aarch64/codec.c
sapostolakis: done

Issue #4665, opened 2021-01-09
[AArch64] Crash with low shared_ibt_table_bb_init value


abhinav92003:
While working on PR #4638, I found that `ibl-stress.c` with -`DTEST_FAR_LINK_AARCH64` on AArch64 crashes with `-shared_ibt_table_bb_init` 16 and 18, but works with 20. Similar behaviour was observed on a proprietary app where `-shared_ibt_table_bb_init` had to be raised to avoid a crash. This was observed to happen with large apps only, which is why` -DTEST_FAR_LINK_AARCH64` is required to reproduce the crash.

abhinav92003:
```
$ dynamorio/bin64/runstats "-s" "9000" "-killpg" "-silent" "-env" "LD_LIBRARY_PATH" "dynamorio/lib64/debug:dynamorio/ext/lib64/debug:" "-env" "DYNAMORIO_OPTIONS" "-dumpcore_mask 0 -disable_traces -shared_bb_ibt_tables -shared_ibt_table_bb_init 16  -checklevel 0  -code_api" "dynamorio/suite/tests/bin/api.ibl-stress"
<Starting application dynamorio/suite/tests/bin/api.ibl-stress (34983)>
<Initial options = -checklevel 0 -code_api -stack_size 56K -signal_stack_size 32K -max_elide_jmp 0 -max_elide_call 0 -shared_bb_ibt_tables -shared_ibt_table_bb_init 16 -no_inline_ignored_syscalls -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct >
<pre-building code cache from 1050001 tags>
<all_memory_areas is missing regions including 0x0000fffd9561a000-0x0000fffd986be000>
<Application tried to execute from unreadable memory 0xf9402381f9400b82.
This may be a result of an unsuccessful attack or a potential application vulnerability.>
<Application dynamorio/suite/tests/bin/api.ibl-stress (34983).  Application exception at PC 0xf9402381f9400b82.  
Signal 11 delivered to application as default action.
Callstack:
	0xf9402381f9400b82  
	0x0000ffff98c7a0a0   </lib/aarch64-linux-gnu/libpthread-2.24.so+0x70a0>
	0x0000ffff98788eac   </lib/aarch64-linux-gnu/libc-2.24.so+0xc7eac>
>
<Stopping application dynamorio/suite/tests/bin/api.ibl-stress (34983)>
[...]

PR #4699, opened 2021-01-29
i#4665: Fix crash after IBT resize on AArch64

abhinav92003:
Adds missing store for `next_tag` and `last_exit` linkstub in indirect branch lookups after IBT resize. In this case, the IBT payloads are replaced with the `target_delete_entry` gencode address to induce a cache exit, and safely nullify the table.

Removes the `-shared_ibt_table_bb_init` workaround for `ibl-stress` test on AArch64, which was required to avoid the buggy IBT resize.

Verified locally that the `ibl-stress-aarch64-far-link_LONG` test also works without the workaround after this fix. Manual local verification was required as this long test is not enabled in the regular suite.

Fixes: #4665


abhinav92003:
There seem to be a couple unrelated issues with CI runs:
- Jenkins is not responding, even to web requests. Is it down?
- `ci-windows` is showing an error `Exception calling "DownloadFile" with "2" argument(s): "Unable to connect to the remote server"`. Didn't go away on retrying.

abhinav92003:
run arm tests

core/arch/aarch64/emit_utils.c
```diff
@@ -947,16 +956,23 @@ emit_indirect_branch_lookup(dcontext_t *dc, generated_code_t *code, byte *pc,
...
+    /* Get back next_tag's value previously saved in r1 by lookup code. This will be
+     * stored in dcontext below.
+     */
+    APP(&ilist,
+        XINST_CREATE_move(dc, opnd_create_reg(DR_REG_R2), opnd_create_reg(DR_REG_R1)));
+
+    APP(&ilist, miss);
+
+    /* Recover the dcontext->last_exit to x0 */
+    APP(&ilist, instr_create_restore_from_tls(dc, DR_REG_R0, TLS_REG3_SLOT));
```
abhinav92003, line 974: I see there's a special linkstub `linkstub_ibl_deleted` that's used by x86 code for cases [when linkstub isn't present](https://github.com/DynamoRIO/dynamorio/blob/e5ac96b4886785546d4b02a1dbcdd132d0c1954b/core/arch/x86/emit_utils.c#L2687). But here as we do have the last linkstub saved in `TLS_REG3_SLOT`. I'm just restoring that.

abhinav92003:
> There seem to be a couple unrelated issues with CI runs:

Both of them went away on their own.

core/arch/aarch64/emit_utils.c
```diff
@@ -891,14 +891,23 @@ emit_indirect_branch_lookup(dcontext_t *dc, generated_code_t *code, byte *pc,
...
-    /* Recover app's original x2. */
-    APP(&ilist, instr_create_restore_from_tls(dc, DR_REG_R2, TLS_REG2_SLOT));
-
     /* ldr x0, [x1, #start_pc_fragment_offset] */
     APP(&ilist,
         INSTR_CREATE_ldr(dc, opnd_create_reg(DR_REG_X0),
                          OPND_CREATE_MEMPTR(
                              DR_REG_X1, offsetof(fragment_entry_t, start_pc_fragment))));
+
+    /* Save next_tag value from r2 to r1. We may need to store it into dcontext in the
```
derekbruening, line 900: This code was modeled on ARM I would guess: did you check whether ARM has the same problem?  At first glance, it looks like it does.

derekbruening, line 900: Instead of paying an extra instruction in the hit path, it seems better to do what x86 does and recover the tag from the table pointer (which is in r1) in the target_delete path.

core/arch/aarch64/emit_utils.c
derekbruening: x86 actually never uses a linkstub for indirect branches (-indirect_stubs is off; exit cti goes straight to IBL) but aarchxx with reachability problems will probably always have a stub.

abhinav92003: I meant what linkstub address is set in `dcontext->last_exit`. I see that on AArch64, `ibl-stress` passes no matter if I set it to the actual last linkstub or `linkstub_ibl_deleted`. But as the former is available here, I use that.

core/arch/aarch64/emit_utils.c
abhinav92003: It seems so too, but I haven't tested it on an ARM machine. The ibl-stress test isn't ported to ARM yet, and I don't have access to an ARM machine. I added the same fix to arm/emit_utils with an ASSERT_NOT_TESTED. Does that work for now?

abhinav92003: Done.

derekbruening: SGTM

core/arch/aarch64/emit_utils.c
derekbruening: What are the consequences of having a real instead of special linkstub?  Without the special linkstub, the stat num_exits_ibl_deleted won't be incremented.  Are there any asserts or warnings about not knowing why DR exited the cache?

derekbruening (approved):
The only question is whether we'd prefer to have the special linkstub telling us why we exited the cache, rather than the real linkstub.

I think the real fragment will still find the right reason it might end up in fragment_add_ibl_target()?  The `reason = "shared IBT table resize";` line.  So is it just the exit stat `num_exits_ibl_deleted` and the LOG about why it exited?

core/arch/aarch64/emit_utils.c
abhinav92003: Didn't find any relevant asserts or warnings when using the special linkstub. Made the change to use it, and I see proper stats for `Fcache exits, deleted but hit in ibl` now.

abhinav92003:
> The only question is whether we'd prefer to have the special linkstub telling us why we exited the cache, rather than the real linkstub.
> 
> I think the real fragment will still find the right reason it might end up in fragment_add_ibl_target()? The `reason = "shared IBT table resize";` line. So is it just the exit stat `num_exits_ibl_deleted` and the LOG about why it exited?

I made the change to use `&linkstub_ibl_deleted`, and am seeing stats now.
```
abhinavas@tx1:~/dr/build/dynamorio/logs/api.ibl-stress.74563.00000000$ grep -ir "Extra exits due to shared table resizing" .
./api.ibl-stress.0.74563.html:          Extra exits due to shared table resizing :              2198
./api.ibl-stress.0.74563.html:          Extra exits due to shared table resizing :              5697
./api.ibl-stress.0.74563.html:          Extra exits due to shared table resizing :              5697
./api.ibl-stress.0.74563.html:          Extra exits due to shared table resizing :              5697
./api.ibl-stress.0.74563.html:          Extra exits due to shared table resizing :              5697
./api.ibl-stress.0.74563.html:          Extra exits due to shared table resizing :              5697
./api.ibl-stress.0.74563.html:          Extra exits due to shared table resizing :              5697
./api.ibl-stress.0.74563.html:          Extra exits due to shared table resizing :              5697
./api.ibl-stress.0.74563.html:          Extra exits due to shared table resizing :              5697
./api.ibl-stress.0.74563.html:          Extra exits due to shared table resizing :              5698
./api.ibl-stress.0.74563.html:          Extra exits due to shared table resizing :              5698
./api.ibl-stress.0.74563.html:          Extra exits due to shared table resizing :              5698
./log.0.74563.html:          Extra exits due to shared table resizing (thread):                 1
./log.3.74566.html:          Extra exits due to shared table resizing (thread):               761
./log.7.74570.html:          Extra exits due to shared table resizing (thread):               771
./log.6.74568.html:          Extra exits due to shared table resizing (thread):               540
./log.1.74564.html:          Extra exits due to shared table resizing (thread):               766
./log.2.74565.html:          Extra exits due to shared table resizing (thread):               747
./log.4.74567.html:          Extra exits due to shared table resizing (thread):               764
./log.5.74569.html:          Extra exits due to shared table resizing (thread):               596
./log.8.74571.html:          Extra exits due to shared table resizing (thread):               752
abhinavas@tx1:~/dr/build/dynamorio/logs/api.ibl-stress.74563.00000000$ grep -ir "Fcache exits, deleted but hit in ibl " .
./api.ibl-stress.0.74563.html:              Fcache exits, deleted but hit in ibl :                57
./api.ibl-stress.0.74563.html:              Fcache exits, deleted but hit in ibl :                57
./api.ibl-stress.0.74563.html:              Fcache exits, deleted but hit in ibl :                57
./log.0.74563.html:              Fcache exits, deleted but hit in ibl (thread):                57
```
By their string, it seems that both of these counters should be the same. I guess the former in addition includes the times when the tag wasn't found in the IBL at all during a resize, right?

I made the same change speculatively to ARM as well. I see another bug though in ARM, and made some changes for that too. Could you please take a look again?

core/arch/arm/emit_utils.c
```diff
@@ -980,10 +980,26 @@ emit_indirect_branch_lookup(dcontext_t *dc, generated_code_t *code, byte *pc,
...
+    APP(&ilist,
+        INSTR_CREATE_ldr(
+            dc, opnd_create_reg(DR_REG_R2),
+            OPND_CREATE_MEMPTR(DR_REG_R1, offsetof(fragment_entry_t, tag_fragment))));
+    instrlist_insert_mov_immed_ptrsz(dc, (ptr_uint_t)get_ibl_deleted_linkstub(),
+                                     opnd_create_reg(DR_REG_R1), &ilist, NULL, NULL,
+                                     NULL);
+    APP(&ilist, instr_create_restore_from_tls(dc, DR_REG_R0, TLS_REG0_SLOT));
+
     /* Unlink path: entry from stub */
```
abhinav92003, line 1003: For the `Unlink path` case, it seems that `fcache_return` will see the app's r1 value in r0, instead of the last linkstub address. (see L1011 -> 1017 -> 1019) 
@derekbruening

core/arch/arm/emit_utils.c
```diff
@@ -980,10 +980,26 @@ emit_indirect_branch_lookup(dcontext_t *dc, generated_code_t *code, byte *pc,
...
+     */
+    ASSERT_NOT_TESTED();
+    APP(&ilist,
+        INSTR_CREATE_ldr(
+            dc, opnd_create_reg(DR_REG_R2),
+            OPND_CREATE_MEMPTR(DR_REG_R1, offsetof(fragment_entry_t, tag_fragment))));
+    instrlist_insert_mov_immed_ptrsz(dc, (ptr_uint_t)get_ibl_deleted_linkstub(),
+                                     opnd_create_reg(DR_REG_R1), &ilist, NULL, NULL,
+                                     NULL);
+    APP(&ilist, instr_create_restore_from_tls(dc, DR_REG_R0, TLS_REG0_SLOT));
```
abhinav92003, line 1001: At this point, `r0` should contain the `target_delete_entry` address, set by the IBL hit path. L1009 would overrwrite app's `r0` value, unless this new line at L1001 is added.

core/arch/arm/emit_utils.c
abhinav92003: Haven't made any fix for this yet.

derekbruening:
> By their string, it seems that both of these counters should be the same. I guess the former in addition includes the times when the tag wasn't found in the IBL at all during a resize (ibl miss), right?

So is the former when a thread has the old mask and the new table and so misses, and the latter is a thread has the old table which was nullified?

core/arch/arm/emit_utils.c
derekbruening: It looks correct I think.
Stub spills r1->r1_slot, r2->r2_slot; puts linkstub into r1, target into r2.
Unlink then spills r0->r0_slot, r1(linkstub) -> r3_slot.
Then on the shared miss path, it moves the linkstub from r3_slot back to r1.
Then it moves the linkstub from r1 into r0.
Then it restores r1_slot->r1, moves target from r2 into ntext_tag, restores r2.

derekbruening: Oh you don't mean entering at the unlink entry point: you mean taking the target delete path and falling through?

abhinav92003: Oh right. `r1` at this point has the linkstub already, not the app value. (Maybe I kept thinking `r0` because of AArch64). Thanks.

core/arch/arm/emit_utils.c
derekbruening: Yes, this makes sense: the shared unlink: sequence will save r0.

core/arch/arm/emit_utils.c
abhinav92003: For the target delete path and falling through, I already made a fix above, by loading `linkstub_ibl_deleted` in r1.

core/arch/arm/emit_utils.c
derekbruening: Should I take this branch and try to run a sanity check on my armbook?

derekbruening: Is there a test you suggest running?  ibl-stress is not ported?

abhinav92003: `ibl-stress` is one option, but it isn't ported yet to ARM. Another option is to set `-shared_ibt_table_bb_init` to a very low value, and try with whatever large app is available.

abhinav92003: Testing on some ARM machine would be great. Thanks for the offer. I was a bit uneasy putting in  many changes in ARM without testing them.

abhinav92003: Let me quickly try and check if any test suite app can trigger this path with a low shared_ibt_table_bb_init.

abhinav92003: You can use these. I think they are available on ARM.
```
dynamorio/bin64/drrun "-s" "90" "-quiet" "-debug" "-killpg"  "-dumpcore_mask" "0" "-enable_reset" "-reset_at_nth_thread" "2" "-code_api" "-disable_traces" "-shared_bb_ibt_tables" "-shared_ibt_table_bb_init" "2" "--" "dynamorio/suite/tests/bin/linux.clone"
```

```
dynamorio/bin64/drrun "-s" "90" "-quiet" "-debug" "-killpg"  "-dumpcore_mask" "0" "-code_api" "-disable_traces" "-shared_bb_ibt_tables" "-shared_ibt_table_bb_init" "2" "--" "dynamorio/suite/tests/bin/linux.signest"
```

abhinav92003: These are the `linux.clone-reset` and `linux.signest` tests respectively, with their options modified to include `-disable_traces -shared_bb_ibt_tables -shared_ibt_table_bb_init 2` in `vmap_run_list` [SHORT](https://github.com/DynamoRIO/dynamorio/blob/a9be2914cd1971f7d4292101a21ff3b18fcf877b/suite/tests/CMakeLists.txt#L101) configuration.

derekbruening: The first crashes with reset enabled, but that is listed as a known ARM problem.  Without reset it runs and it does hit the ASSERT_NOT_TESTED (though I guess that's on generation so any app at init will hit it):

```
> bin32/drrun -disable_traces -loglevel 1 "-shared_bb_ibt_tables" "-shared_ibt_table_bb_init" "2" "--" "suite/tests/bin/linux.clone"
<log dir=/home/derek/dr/git/build_dbg_tests/bin32/../logs/linux.clone.28638.00000000>
<Starting application /home/derek/dr/git/build_dbg_tests/suite/tests/bin/linux.clone (28638)>
<Not tested @/home/derek/dr/git/src/core/arch/arm/emit_utils.c:993>
<Initial options = -no_dynamic_options -loglevel 1 -code_api -stack_size 56K -signal_stack_size 32K -max_elide_jmp 0 -max_elide_call 0 -shared_bb_ibt_tables -shared_ibt_table_bb_init 2 -early_inject -emulate_brk -no_inline_ignored_syscalls -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct >
<(1+x) Handling our fault in a TRY at 0x711e60e0>
<Paste into GDB to debug DynamoRIO clients:
set confirm off
add-symbol-file '/home/derek/dr/git/build_dbg_tests/lib32/debug/libdynamorio.so' 0x71024368
>
Sideline thread started
i = 2500000
i = 5000000
i = 7500000
i = 10000000
i = 12500000
i = 15000000
i = 17500000
i = 20000000
i = 22500000
i = 25000000
Sideline thread finished
Waiting for child to exit
Child has exited
Sideline thread started
i = 2500000
i = 5000000
i = 7500000
i = 10000000
i = 12500000
i = 15000000
i = 17500000
i = 20000000
i = 22500000
i = 25000000
Sideline thread finished
Waiting for child to exit
Child has exited
<Stopping application /home/derek/dr/git/build_dbg_tests/suite/tests/bin/linux.clone (28638)>
```

It has no exit hits but it's probably too small with too few threads:
```
$ egrep 'table resiz|deleted but hit' `ls -1td logs/*0|head -1`/li*
```

derekbruening: The other hits a problem that does not occur without the ibt table options so may indicate a problem with this PR?

```
> bin32/drrun -disable_traces -loglevel 1 "-shared_bb_ibt_tables" "-shared_ibt_table_bb_init" "2" "--" "suite/tests/bin/linux.signest"
<log dir=/home/derek/dr/git/build_dbg_tests/bin32/../logs/linux.signest.29007.00000000>
<Starting application /home/derek/dr/git/build_dbg_tests/suite/tests/bin/linux.signest (29007)>
<Not tested @/home/derek/dr/git/src/core/arch/arm/emit_utils.c:993>
<Initial options = -no_dynamic_options -loglevel 1 -code_api -stack_size 56K -signal_stack_size 32K -max_elide_jmp 0 -max_elide_call 0 -shared_bb_ibt_tables -shared_ibt_table_bb_init 2 -early_inject -emulate_brk -no_inline_ignored_syscalls -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct >
<(1+x) Handling our fault in a TRY at 0x711e60e0>
<Paste into GDB to debug DynamoRIO clients:
set confirm off
add-symbol-file '/home/derek/dr/git/build_dbg_tests/lib32/debug/libdynamorio.so' 0x71024368
>
sending 2 signals
in handler 42
in handler 42
<get_memory_info mismatch! (can happen if os combines entries in /proc/pid/maps)
	os says: 0xb6475000-0xb6675000 prot=0x00000000
	cache says: 0xb6475000-0xb6676000 prot=0x00000000
>
<Application /home/derek/dr/git/build_dbg_tests/suite/tests/bin/linux.signest (29007).  Internal Error: DynamoRIO debug check failure: /home/derek/dr/git/src/core/unix/signal.c:5452 syscall_signal || safe_is_in_fcache(dcontext, pc, (byte *)sc->SC_XSP)
(Error occurred @2526 frags)
version 8.0.18659, custom build
-no_dynamic_options -loglevel 1 -code_api -stack_size 56K -signal_stack_size 32K -max_elide_jmp 0 -max_elide_call 0 -shared_bb_ibt_tables -shared_ibt_table_bb_init 2 -early_inject -emulate_brk -no_inline_ignored_syscalls -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct 
/home/derek/dr/git/build_dbg_tests/lib32/debug/libdynamorio.so=0x71000000>
```

derekbruening: Trying to analyze this crash...the gdb on this machine is old and having symbol trouble.  The big-little core is also missing some kernel support like watchpoints IIRC.

abhinav92003: About `linux.clone`, without reset it doesn't hit the bug, so a successful run there doesn't tell us anything. `Fcache exits, deleted but hit in ibl` should also be zero for this run.

abhinav92003: About the `linux.signest` crash, does it also occur with a high value of `-shared_ibt_table_bb_init`? (which shouldn't invoke the new code, but still use shared ibt)

derekbruening: linux.signest does not crash with the default value for `-shared_ibt_table_bb_init`.  In gdb: it does seem to hit the target_delete case and goes to that IBL code sequence.  I think the bug is that our gencode is Thumb but it does a `bx` without LSB set.

derekbruening: OK after fixing the Thumb LSB issue we hit a crash that is a bug in this PR:

```
Thread 1 "linux.signest" received signal SIGSEGV, Segmentation fault.
0x41e1d35e in ?? ()
1: x/i $pc
=> 0x41e1d35e:	ldr	r2, [r1, #0]
```

r1 was cleared on the hit path up above:
```
1: x/i $pc
=> 0x41e1d334:	sub.w	r0, r0, r2
(gdb) 
0x41e1d338 in ?? ()
1: x/i $pc
=> 0x41e1d338:	cbnz	r0, 0x41e1d346
(gdb) 
0x41e1d33a in ?? ()
1: x/i $pc
=> 0x41e1d33a:	ldr	r0, [r1, #4]
(gdb) 
0x41e1d33c in ?? ()
1: x/i $pc
=> 0x41e1d33c:	ldr.w	r1, [r10, #4]
(gdb) 
0x41e1d340 in ?? ()
1: x/i $pc
=> 0x41e1d340:	ldr.w	r2, [r10, #8]
(gdb) 
0x41e1d344 in ?? ()
1: x/i $pc
=> 0x41e1d344:	bx	r0
```

Hmm.  Looking at x86: it doesn't load the target into rcx; it jumps through memory `jmp 0x08(%rcx)`, so rcx still holds the table pointer.  So it looks like I was wrong: we don't have the table pointer still available after the hit jump for aarchxx??

derekbruening: So it is there for a64, so that does hold; it's just a32 that does not have it.

derekbruening: So for a32 we would either:

1) Re-arrange the IBL to have the table pointer in r0 (swapping with r1 for the table load and compare) and change the `bx r0` to `ldr pc, [r0 + 4]`
2) Add a store on the hit path to stick the tag r2 into a TLS slot to be restored on the target_delete path

?

abhinav92003: Ah right. ARM restores app x1 at L936. Sorry, I should've seen that. Let me put in a proper workaround.

abhinav92003: So, in this case, I guess we don't have an option but to store the tag into a TLS slot and take the performance hit on the hit path right?

derekbruening: > So, in this case, I guess we don't have an option but to store the tag into a TLS slot and take the performance hit on the hit path right?

See https://github.com/DynamoRIO/dynamorio/pull/4699#discussion_r568238106 which gives another option which I assume is more performant (that cache line is already in L1 and the load should beat a store) but gets more complex w/ rearranging the IBL.

derekbruening: I pushed my Thumb LSB fix to this branch.  I guess add a `Co-authored-by: ` to the final squash message.

abhinav92003: Thanks Derek for the help in testing this on ARM and the fix too.

I've pushed my fix to this branch; can you please test again? I have implemented option 2 above for now, as it is simpler, given that I don't have access to an ARM machine.

derekbruening: It gets through two target_deletes (shouldn't it have the new table after going back to DR?  not sure I understand it) and then hits an assert on the mode.  Not sure what the cause is: the IBL has insert_mode_change_handling.
```
Entry into F2526(0x00010e18).0x4cb9de90 (A32)(shared)
Exit from fragment deleted but hit in ibl

d_r_dispatch: target = 0x00010d74
Entry into F1383(0x00010d74).0x4caf6f10 (A32)(shared)
Exit from fragment deleted but hit in ibl

d_r_dispatch: target = 0xb6fb90c8
Entry into F1240(0xb6fb90c8).0x4caf42bc (T32)(shared)
SYSLOG_ERROR: Application /home/derek/dr/git/build_dbg_tests/suite/tests/bin/linux.signest (28957).  Internal Error: DynamoRIO debug check failure: /home/derek/dr/git/src/core/dispatch.c:487 dr_get_isa_mode(dcontext) == FRAG_ISA_MODE(targetf->flags) IF_X64(|| (dr_get_isa_mode(dcontext) == DR_ISA_IA32 && !FRAG_IS_32(targetf->flags) && DYNAMO_OPTION(x86_to_x64)))
(Error occurred @2526 frags)
```

abhinav92003: > It gets through two target_deletes (shouldn't it have the new table after going back to DR? not sure I understand it)
You mean the same thread invokes `target_delete_entry` code multiple times? That could happen if the IBL goes through multiple resizes I think.

derekbruening: See the log snippet I pasted at https://github.com/DynamoRIO/dynamorio/pull/4699#discussion_r568296180: there is no resize in between them (a resize prints a line at this loglevel: I see them up above in the log).  There's nothing in between them but a roundtrip from dispatch to the cache.

derekbruening: Ah, I see the source of the assert: insert_mode_change_handling() is called very early in the IBL and it clears the LSB.  Then later in the code after the unlinked: entry point it's called again, but now using the LSB-cleared copy of the tag.  So that needs to be fixed.  Re-loading from the table would fix it since the table has the LSB-set version.  Not sure how else to easily fix it: an inverted form of insert_mode_change_handling(), or storing the original in yet another TLS slot, or something.

abhinav92003: If the issue is a duplicate invocation of `insert_mode_change_handling`, then c42836e may be a possible fix that I pushed to the branch now.

derekbruening: Yes that works:

```
$ bin32/drrun -disable_traces -loglevel 2 "-shared_bb_ibt_tables" "-shared_ibt_table_bb_init" "2" "--" "suite/tests/bin/linux.signest"
<log dir=/home/derek/dr/git/build_dbg_tests/bin32/../logs/linux.signest.4232.00000000>
<Starting application /home/derek/dr/git/build_dbg_tests/suite/tests/bin/linux.signest (4232)>
<Initial options = -no_dynamic_options -loglevel 2 -code_api -stack_size 56K -signal_stack_size 32K -max_elide_jmp 0 -max_elide_call 0 -shared_bb_ibt_tables -shared_ibt_table_bb_init 2 -early_inject -emulate_brk -no_inline_ignored_syscalls -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct >
<(1+x) Handling our fault in a TRY at 0x711e6070>
<Paste into GDB to debug DynamoRIO clients:
set confirm off
add-symbol-file '/home/derek/dr/git/build_dbg_tests/lib32/debug/libdynamorio.so' 0x71024368
>
sending 2 signals
in handler 42
in handler 42
<get_memory_info mismatch! (can happen if os combines entries in /proc/pid/maps)
	os says: 0xb6458000-0xb6658000 prot=0x00000000
	cache says: 0xb6458000-0xb6659000 prot=0x00000000
>
all done
<Stopping application /home/derek/dr/git/build_dbg_tests/suite/tests/bin/linux.signest (4232)>

$ egrep -n 'deleted but hit|bb hashtable resiz' `ls -1td logs/*0|head -1`/lo*
logs/linux.signest.4232.00000000/log.0.4232.html:2819:indjmp_bb hashtable resized at 4 entries from capacity 5 to 9
logs/linux.signest.4232.00000000/log.0.4232.html:3293:ret_bb hashtable resized at 4 entries from capacity 5 to 9
logs/linux.signest.4232.00000000/log.0.4232.html:3638:indjmp_bb hashtable resized at 7 entries from capacity 9 to 17
logs/linux.signest.4232.00000000/log.0.4232.html:4738:ret_bb hashtable resized at 7 entries from capacity 9 to 17
logs/linux.signest.4232.00000000/log.0.4232.html:6598:shared_bb hashtable resized at 564 entries from capacity 1025 to 2049
logs/linux.signest.4232.00000000/log.0.4232.html:7016:ret_bb hashtable resized at 12 entries from capacity 17 to 33
logs/linux.signest.4232.00000000/log.0.4232.html:8330:indcall_bb hashtable resized at 4 entries from capacity 5 to 9
logs/linux.signest.4232.00000000/log.0.4232.html:8703:indjmp_bb hashtable resized at 12 entries from capacity 17 to 33
logs/linux.signest.4232.00000000/log.0.4232.html:9249:ret_bb hashtable resized at 24 entries from capacity 33 to 65
logs/linux.signest.4232.00000000/log.0.4232.html:9831:ret_bb hashtable resized at 46 entries from capacity 65 to 129
logs/linux.signest.4232.00000000/log.0.4232.html:10371:indcall_bb hashtable resized at 7 entries from capacity 9 to 17
logs/linux.signest.4232.00000000/log.0.4232.html:13523:indjmp_bb hashtable resized at 24 entries from capacity 33 to 65
logs/linux.signest.4232.00000000/log.0.4232.html:13934:shared_bb hashtable resized at 1127 entries from capacity 2049 to 4097
logs/linux.signest.4232.00000000/log.0.4232.html:15035:ret_bb hashtable resized at 91 entries from capacity 129 to 257
logs/linux.signest.4232.00000000/log.0.4232.html:18075:indcall_bb hashtable resized at 12 entries from capacity 17 to 33
logs/linux.signest.4232.00000000/log.0.4232.html:20911:Exit from fragment deleted but hit in ibl
logs/linux.signest.4232.00000000/log.0.4232.html:20915:Exit from fragment deleted but hit in ibl
logs/linux.signest.4232.00000000/log.0.4232.html:20919:Exit from fragment deleted but hit in ibl
logs/linux.signest.4232.00000000/log.0.4232.html:20923:Exit from fragment deleted but hit in ibl
logs/linux.signest.4232.00000000/log.0.4232.html:20989:Exit from fragment deleted but hit in ibl
logs/linux.signest.4232.00000000/log.0.4232.html:20993:Exit from fragment deleted but hit in ibl
logs/linux.signest.4232.00000000/log.0.4232.html:20997:Exit from fragment deleted but hit in ibl
logs/linux.signest.4232.00000000/log.0.4232.html:21001:Exit from fragment deleted but hit in ibl
logs/linux.signest.4232.00000000/log.0.4232.html:21034:Exit from fragment deleted but hit in ibl
logs/linux.signest.4232.00000000/log.0.4232.html:21038:Exit from fragment deleted but hit in ibl
logs/linux.signest.4232.00000000/log.0.4232.html:21049:Exit from fragment deleted but hit in ibl
logs/linux.signest.4232.00000000/log.0.4232.html:21053:Exit from fragment deleted but hit in ibl
logs/linux.signest.4232.00000000/log.0.4232.html:21057:Exit from fragment deleted but hit in ibl
logs/linux.signest.4232.00000000/log.0.4232.html:21061:Exit from fragment deleted but hit in ibl
logs/linux.signest.4232.00000000/log.0.4232.html:21065:Exit from fragment deleted but hit in ibl
logs/linux.signest.4232.00000000/log.0.4232.html:21079:Exit from fragment deleted but hit in ibl
logs/linux.signest.4232.00000000/log.0.4232.html:21253:Exit from fragment deleted but hit in ibl
logs/linux.signest.4232.00000000/log.0.4232.html:21540:Exit from fragment deleted but hit in ibl
logs/linux.signest.4232.00000000/log.0.4232.html:21573:Exit from fragment deleted but hit in ibl
logs/linux.signest.4232.00000000/log.0.4232.html:21584:Exit from fragment deleted but hit in ibl
logs/linux.signest.4232.00000000/log.0.4232.html:21595:Exit from fragment deleted but hit in ibl
logs/linux.signest.4232.00000000/log.0.4232.html:21793:Exit from fragment deleted but hit in ibl
logs/linux.signest.4232.00000000/log.0.4232.html:21797:Exit from fragment deleted but hit in ibl
logs/linux.signest.4232.00000000/log.0.4232.html:21974:Exit from fragment deleted but hit in ibl
logs/linux.signest.4232.00000000/log.0.4232.html:21978:Exit from fragment deleted but hit in ibl
logs/linux.signest.4232.00000000/log.0.4232.html:22736:              Fcache exits, deleted but hit in ibl (thread):       25
logs/linux.signest.4232.00000000/log.1.4234.html:2008:indcall_bb hashtable resized at 24 entries from capacity 33 to 65
logs/linux.signest.4232.00000000/log.1.4234.html:4518:shared_bb hashtable resized at 2254 entries from capacity 4097 to 8193
logs/linux.signest.4232.00000000/log.1.4234.html:5513:indjmp_bb hashtable resized at 46 entries from capacity 65 to 129

$ egrep -n 'deleted but hit|resiz' `ls -1td logs/*0|head -1`/li* | tail -3
3771:                                       IBT resizes :       15
3781:          Extra exits due to shared table resizing :        1
3788:              Fcache exits, deleted but hit in ibl :       25
```

derekbruening: I still don't understand why there are so many "Exit from fragment deleted but hit in ibl" when there are no further resizes in between -- does that make sense to you?

derekbruening: The other thread did perform 3 resizes: but we have a lot of exits in a row, far more than that would explain.

abhinav92003: I see those `hit-but-exit`s in a row on AArch64 too.

A possibility is that there's some bug that's causing misses all the time after the resize; maybe some missing update of table/mask? A run of `ibl-stress` with 20k sequences and `-shared_ibt_table_bb_init 2` gives 455 `Exit from fragment deleted but hit in ibl` logs. Given this (perhaps low) number, is such a bug likely? `ibl-stress` has 1 jump reg and 1 call reg per sequence. The exits due to resizing is quite high though.

```
./api.ibl-stress.0.29838.html:          Extra exits due to shared table resizing :             58001
```
```
./api.ibl-stress.0.29838.html:              Fcache exits, deleted but hit in ibl :               455

```

Or maybe the resize took so long that this thread executed `target_delete_entry` multiple times? Given that the thread would need to go through dispatch each time (which would perhaps give more time for the resize to complete), is that likely? 

Logs from ibl-stress run with high sequence count: Notice the bursty-ness, which makes me think maybe the resize did take too long.
```
./log.3.29847.html:1485:Exit from fragment deleted but hit in ibl
./log.3.29847.html:1489:Exit from fragment deleted but hit in ibl
./log.3.29847.html:1493:Exit from fragment deleted but hit in ibl
./log.3.29847.html:1497:Exit from fragment deleted but hit in ibl
./log.3.29847.html:1501:Exit from fragment deleted but hit in ibl
./log.3.29847.html:1505:Exit from fragment deleted but hit in ibl
./log.3.29847.html:1509:Exit from fragment deleted but hit in ibl
./log.3.29847.html:1513:Exit from fragment deleted but hit in ibl
./log.3.29847.html:1517:Exit from fragment deleted but hit in ibl
./log.3.29847.html:1521:Exit from fragment deleted but hit in ibl
./log.3.29847.html:1525:Exit from fragment deleted but hit in ibl
./log.3.29847.html:1529:Exit from fragment deleted but hit in ibl
./log.3.29847.html:1533:Exit from fragment deleted but hit in ibl
./log.3.29847.html:1537:Exit from fragment deleted but hit in ibl
./log.3.29847.html:1549:Exit from fragment deleted but hit in ibl
./log.3.29847.html:1553:Exit from fragment deleted but hit in ibl
./log.3.29847.html:1557:Exit from fragment deleted but hit in ibl
./log.3.29847.html:1561:Exit from fragment deleted but hit in ibl
./log.3.29847.html:1565:Exit from fragment deleted but hit in ibl
./log.3.29847.html:1630:Exit from fragment deleted but hit in ibl

./log.3.29847.html:19138:Exit from fragment deleted but hit in ibl
./log.3.29847.html:19142:Exit from fragment deleted but hit in ibl
./log.3.29847.html:19146:Exit from fragment deleted but hit in ibl

./log.3.29847.html:38570:Exit from fragment deleted but hit in ibl
./log.3.29847.html:38574:Exit from fragment deleted but hit in ibl
./log.3.29847.html:38578:Exit from fragment deleted but hit in ibl
./log.3.29847.html:38582:Exit from fragment deleted but hit in ibl
./log.3.29847.html:38586:Exit from fragment deleted but hit in ibl

./log.3.29847.html:77637:Exit from fragment deleted but hit in ibl
./log.3.29847.html:77641:Exit from fragment deleted but hit in ibl
./log.3.29847.html:77645:Exit from fragment deleted but hit in ibl
```

derekbruening: > Or maybe the resize took so long that this thread executed target_delete_entry multiple times? Given that the thread would need to go through dispatch each time (which would perhaps give more time for the resize to complete), is that likely?

I added an `os_thread_sleep(1000);` to dispatch when it hits the `Exit from fragment deleted but hit in ibl` to give the other thread time to finish a resize: yet I have the same # of such exits in a row.

abhinav92003: > Are you aware of any dispatch flows where dcontext->last_exit is not updated? That could cause duplicate counting and logs.

To check if this is the cause, I set `dcontext->last_exit` to a new linkstub after the log is printed in `dispatch_exit_fcache_stats`, but I still see the exits in a row.

abhinav92003: Perhaps this has something to do with the weak memory model on arm? But then your  `os_thread_sleep` experiment should have shown some effect on the # of repeated logs.

abhinav92003: https://github.com/DynamoRIO/dynamorio/blob/a9be2914cd1971f7d4292101a21ff3b18fcf877b/core/fragment.c#L4217

Here it's documented that a thread's local ptr and state are updated only on IBL miss. Not sure if that includes ibl hits with `target_delete` payload too or not. If it doesn't, that may explain the extra "ibl hits but exit" after resize.

derekbruening: Looks like the special linkstub is not marked indirect, so it doesn't go into that code.  But fixing it results in an assert somewhere else.  Let's separate it out from this PR.  I filed #4704.  I think this PR is good to go.

derekbruening (approved):

core/arch/arm/emit_utils.c
abhinav92003: Okay, I'll go ahead with this PR then. Thanks for flagging #4704.

abhinav92003:
> > By their string, it seems that both of these counters should be the same. I guess the former in addition includes the times when the tag wasn't found in the IBL at all during a resize (ibl miss), right?
> 
> So is the former when a thread has the old mask and the new table and so misses, and the latter is a thread has the old table which was nullified?

The former would count all misses that happen during the resize I think. Not only the ones that would've hit in the old table (which is the latter metric).

PR #4703, opened 2021-02-01
i#2666: Dynamically size CONTEXT buffers

### LINKED ISSUE #2666: Make MAX_CONTEXT_SIZE dynamically sized to future-proof it

derekbruening:
Changes all stack buffers using hardcoded sizes for CONTEXT to instead
use heap-allocated buffers sized to match the kernel's actual CONTEXT
size for the given flags.

Adds a mechanism to query the CONTEXT64 size from 32-bit DR using
switch_modes_and_call().

The one place where we can't easily use heap is
thread_set_self_mcontext() because it doesn't return and we thus do
not have a simple way to free the buffer.  There, we use a large
4096-byte stack buffer, and we check the size needed: if it's larger
than that, we use a lost heap allocation and issue a debug build
warning which should alert someone.  If we end up needing more space
there in the future we may want to store a pointer in the dcontext and
have a special check to free it or something.

Fixes #2666

### DIFF
--- a/core/win32/module_shared.c
+++ b/core/win32/module_shared.c
@@ -1200,6 +1200,28 @@ free_library_64(HANDLE lib)
     return (res >= 0);
 }
 
+size_t
+nt_get_context64_size(void)
+{
+    static size_t context64_size;
+    if (context64_size == 0) {
+        uint64 ntdll64_RtlGetExtendedContextLength;
+        int len = 0;
+        uint64 len_param = (uint64)&len;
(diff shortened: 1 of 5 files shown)


derekbruening:
run arm tests

Issue #4698, opened 2021-01-28
Port api.detach_state to aarch64 and fix any bugs it finds


derekbruening:
Xref #4474 on bridging the testing gap.
Xref #1578 on detach on arm and some bugs already found and fixed there to get api.detach working.
But api.startstop and several other tests still have crashes or hangs.
This issue is about getting api.detach_state ported to aarch64 which may find bugs at the point of occurence that fix the other tests where the symptoms are downstream.

Issue #4486, opened 2020-10-12
ASSERT callback.c:1478 get_app_pc_from_intercept_pc not finding pc on burst_static test


derekbruening:
This happened 2x in a row with PR 4485:

https://ci.appveyor.com/project/DynamoRIO/dynamorio/builds/35693375
https://ci.appveyor.com/project/DynamoRIO/dynamorio/builds/35693164
```
[00:22:23] 219: 
[00:22:23] 218/254 Test #219: code_api|tool.drcacheoff.burst_static ........................***Failed    0.99 sec
[00:22:23] Running cmd |C:/projects/dynamorio/build/build_debug-internal-64/clients/bin64/tool.drcacheoff.burst_static.exe|
[00:22:23] CMake Error at C:/projects/dynamorio/suite/tests/runmulti.cmake:106 (message):
[00:22:23]   *** cmd failed (-1): pre-DR init
[00:22:23] 
[00:22:23]   pre-DR start
[00:22:23] 
[00:22:23]   <Application
[00:22:23]   C:\projects\dynamorio\build\build_debug-internal-64\clients\bin64\tool.drcacheoff.burst_static.exe
[00:22:23]   (4856).  Internal Error: DynamoRIO debug check failure:
[00:22:23]   ..\..\core\win32\callback.c:1478 false
[00:22:23] 
[00:22:23]   (Error occurred @72 frags)
```

I can't reproduce it on win10-2004:
```
% for ((i=0; i<50; i++)); do echo $i; rm -rf drmemt*.dir; clients/bin64/tool.drcacheoff.burst_static.exe; done > OUT 2>&1
% grep Error OUT
% grep -c 'all done' OUT
150
```

It then went away with the s/eax/ebx/ fix: but it's not clear that's really
related or its non-deterministic in some other way, or whether it's
completely unrelated to that PR and is some latent underlying issue.
Without a reproducer we'll file it and see if it happens again.

derekbruening:
Happened again: https://ci.appveyor.com/project/DynamoRIO/dynamorio/builds/36126121
```
21601219:   <Application
21601219:   C:\projects\dynamorio\build\build_debug-internal-64\clients\bin64\tool.drcacheoff.burst_static.exe
21602219:   (5080).  Internal Error: DynamoRIO debug check failure:
21603219:   ..\..\core\win32\callback.c:1478 false
```

derekbruening:
This is happening on GA Server2016 as well:
[...]

PR #4708, opened 2021-02-03
i#4698 detach_state: Port signal + GPR-cache tests to A64

derekbruening:
Enables api.detach_state on AArch64 with some initial the sub-tests enabled:
+ Enables the cross-platform signal state tests on AArch64.
+ Ports the assembly to push and pop all the general-purpose
  registers, fill them with sentinel values, and test them on a detach
  from the code cache.

As part of this, adds some general assembly support:
+ @P@ expands to # in cpp2asm.
+ Adds REG_SP and JUMP_NOT_EQUAL cross-platform asm_defines.
+ Adds a missing dep on detach_state_shared.h for the asm code.

Still left to do for future work:
+ Check SIMD registers.
+ Check flags.
+ Check stack.
+ Detach from DR rather than the cache.

Issue: #4698

### DIFF
--- a/suite/tests/api/detach_state.c
+++ b/suite/tests/api/detach_state.c
@@ -658,16 +658,16 @@ START_FILE
         stp      x4, x5, [sp, #-16]! @N@\
         stp      x2, x3, [sp, #-16]! @N@\
         stp      x0, x1, [sp, #-16]!
-#        define PUSHALL                         \
+#    define PUSHALL \
         PUSH_GPRS
         /* TODO i#4698: Push and check SIMD regs. */
-#        define PUSH_CALLEE_SAVED \
-        stp      x28, x29, [sp, #-16]! @N@\
-        stp      x26, x27, [sp, #-16]! @N@\
-        stp      x24, x25, [sp, #-16]! @N@\
-        stp      x22, x23, [sp, #-16]! @N@\
-        stp      x20, x21, [sp, #-16]! @N@\
-        stp      x19, x30, [sp, #-16]!
+#    define PUSH_CALLEE_SAVED \
+        stp      x29, x30, [sp, #-16]! @N@\
+        stp      x27, x28, [sp, #-16]! @N@\
+        stp      x25, x26, [sp, #-16]! @N@\
+        stp      x23, x24, [sp, #-16]! @N@\
+        stp      x21, x22, [sp, #-16]! @N@\
+        stp      x19, x20, [sp, #-16]!
 #    define POP_GPRS \
         ldp      x0, x1, [sp], #16 @N@\
         ldp      x2, x3, [sp], #16 @N@\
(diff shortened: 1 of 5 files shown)


derekbruening:
The failures are all what looks like a bug in clang-format-diff where it disagrees with clang-format on the `/* clang-format on */` comment.  Not really sure how to work around this: maybe just remove the comment since it's the end of the file anyway.

suite/tests/api/detach_state.c
```diff
@@ -568,12 +627,74 @@ START_FILE
...
+        stp      x16, x17, [sp, #-16]! @N@\
+        stp      x14, x15, [sp, #-16]! @N@\
+        stp      x12, x13, [sp, #-16]! @N@\
+        stp      x10, x11, [sp, #-16]! @N@\
+        stp      x8, x9, [sp, #-16]! @N@\
+        stp      x6, x7, [sp, #-16]! @N@\
+        stp      x4, x5, [sp, #-16]! @N@\
+        stp      x2, x3, [sp, #-16]! @N@\
+        stp      x0, x1, [sp, #-16]!
+#        define PUSHALL                         \
```
abhinav92003, line 661: nit: Indentation can be fixed. Or is that due to clang-format?

suite/tests/api/detach_state.c
```diff
@@ -568,12 +627,74 @@ START_FILE
...
+        PUSH_GPRS
+        /* TODO i#4698: Push and check SIMD regs. */
+#        define PUSH_CALLEE_SAVED \
+        stp      x28, x29, [sp, #-16]! @N@\
+        stp      x26, x27, [sp, #-16]! @N@\
+        stp      x24, x25, [sp, #-16]! @N@\
+        stp      x22, x23, [sp, #-16]! @N@\
+        stp      x20, x21, [sp, #-16]! @N@\
+        stp      x19, x30, [sp, #-16]!
```
abhinav92003, line 670: Curious why x30 is pushed at the end, paired with x19, which seems a bit irregular. Why not push (x29, x30),(x27,x28),...,(x19,x20)? 

We need ` <even,even+1>` only in ARM right?

abhinav92003 (approved):

suite/tests/api/detach_state.c
derekbruening: Fixed, here and below.

The whole asm side of the file has clang-format off b/c it was doing the wrong thing, I think b/c it thinks it's still C code (all of the split-asm files are like that).

suite/tests/api/detach_state.c
derekbruening: I think this was b/c I looked up the callee-saved regs and it said x19..x29.  I then copied the lines from PUSH_GPRS and had x19 left over for a non-pair push.  But later I realized x30 as the link register has to be saved too and so I changed the x19 to a pair push.  I will go ahead and rewrite it to make it in order.

derekbruening:
Failure is burst_static callback.c assert which is #4486

PR #4709, opened 2021-02-04
i#4607: Fix null pointer dereference in drwrap

### LINKED ISSUE #4607: [CRASH] Dynamorio: received signal SIGSEGV while using wrappers.
**Describe the bug**
Dynamorio crashes in function wrapper.

**To Reproduce**
Launch dynamorio client with wrappers on example which uses library to process pictures.
Dynamorio successfully run without any client.

**Expected behavior**
Successful run dynamorio with wrapper.

Launch Dynamorio with dgb and stop on SIGSEGV:
```
(gdb) bt
#0  0x00007f527840cae7 in drwrap_after_callee_func (drcontext=0x7f50789c4080, pt=0x7f52787f74c8, mc=0x7f50791dbd90, level=0, retaddr=0x0, unwind=1 '\001', only_requested_unwind=0 '\000')
    at /home/novikov/Student/git/symbolic_executor/dynamorio/ext/drwrap/drwrap.c:2089
#1  0x00007f527840deee in drwrap_in_callee_check_unwind (drcontext=0x7f50789c4080, pt=0x7f52787f74c8, mc=0x7f50791dbd90) at /home/novikov/Student/git/symbolic_executor/dynamorio/ext/drwrap/drwrap.c:2644
#2  0x00007f527840bfbe in drwrap_in_callee (arg1=0x7f52ba1e90e0, xsp=140721126304792) at /home/novikov/Student/git/symbolic_executor/dynamorio/ext/drwrap/drwrap.c:1902
#3  0x00007f52789b8b05 in ?? ()
#4  0x0000000000000008 in ?? ()
#5  0x0000000000000008 in ?? ()
#6  0x00007f5077c0060e in ?? ()
#7  0x00007ffc30beec18 in ?? ()
#8  0x0000000000000008 in ?? ()
#9  0x00007f5077c00600 in ?? ()
#10 0x00000000000001b9 in ?? ()
#11 0x0000000000000000 in ?? ()
```
when print pt->user_data_pre_cb[level]:
```
(gdb) p pt->user_data_pre_cb[level]
$2 = (void **) 0x0
```

**Versions**
Dynamorio version: master, Dec 3 2020.
Operating system version: Ubuntu 18.04.5 LTS

SweetVishnya:
drwrap_ensure_postcall may not return. Thus, we should decrement
wrap_level, so, it won't be incremented again in drwrap_in_callee. We
reproduce the behavior prior to 9bb0354 commit.

Fixes #4607

### DIFF
--- a/ext/drwrap/drwrap.c
+++ b/ext/drwrap/drwrap.c
@@ -1705,9 +1705,10 @@ set_retaddr_on_stack(reg_t xsp, app_pc value)
     return res;
 }
 
-/* may not return */
+/* May not return. If it doesn't return, it decrements wrap_level and assumes
+ * it's re-starting in-callee. */
 static void
-drwrap_mark_retaddr_for_instru(void *drcontext, app_pc decorated_pc,
+drwrap_mark_retaddr_for_instru(void *drcontext, per_thread_t *pt, app_pc decorated_pc,
                                drwrap_context_t *wrapcxt, bool enabled)
 {
     post_call_entry_t *e;
@@ -1769,6 +1770,10 @@ drwrap_mark_retaddr_for_instru(void *drcontext, app_pc decorated_pc,
             /* ensure we have DR_MC_ALL */
             drwrap_get_mcontext_internal((void *)wrapcxt, DR_MC_ALL);
             wrapcxt->mc->pc = decorated_pc;
+            /* i#4607: We should decrement wrap_level, so, it won't be
+             * incremented again in drwrap_in_callee.
+             */
+            --pt->wrap_level;
             dr_redirect_execution(wrapcxt->mc);
             ASSERT(false, "dr_redirect_execution should not return");
         }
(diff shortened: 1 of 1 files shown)


ext/drwrap/drwrap.c
```diff
@@ -1931,8 +1931,21 @@ drwrap_in_callee(void *arg1, reg_t xsp _IF_NOT_X86(reg_t lr))
                 break; /* we do need a post-call hook */
             }
         }
-        if (intercept_post && wrapcxt.retaddr != NULL)
+        if (intercept_post && wrapcxt.retaddr != NULL) {
+            if (!TEST(DRWRAP_REPLACE_RETADDR, wrap->flags)) {
```
derekbruening, line 1935: Hmm, so this if() is b/c pt->wrap_level is examined inside drwrap_ensure_postcall() if the flag is set (must be why the wrap_level++ was moved in the first place).  It feels a little fragile though if future code inside drwrap_ensure_postcall() is added which looks at wrap_level.

Maybe instead pass pt to drwrap_mark_retaddr_for_instru() and have it decrement right before its call to dr_redirect_execution()?  That has its own fragileness if it's called from somewhere else in the future -- but IMHO it seems nicer (could put a function-level comment explaining that if it doesn't return it decrements wrap_level under the assumption that it's re-starting in-callee).

derekbruening (commented):
Thank you for the pull request.  I have a suggested change.

SweetVishnya:
I updated the code as you suggested.

derekbruening (approved):
LGTM.  Thank you for the patch.  Ideally we would add a regression test to the suite that triggers this code path, but it may not be simple: to hit the flush case the test would have to run the exact post-call block of code before executing the call, which would likely have to be done in assembly.  Let's merge this to get the fix in place; if someone is able to build such a test we will put it in separately.

PR #4715, opened 2021-02-05
i#1578,i#4474 arm detach: Enable api.startstop on arm

derekbruening:
The api.startstop test was crashing previously, but one of the many
recent fixes seems to have addressed that problem.  It now runs to
completion 1000x in a row on AArch64.

Issue: #1578, #4474


derekbruening:
Failure is burst_static and burst_replace: #4486

abhinav92003 (approved):

Issue #4711, opened 2021-02-04
tool.drcachesim.delay-global and tool.drcacheoff.max-global tests failing: did not hit max


derekbruening:
Looks like it didn't hit the max:
https://github.com/DynamoRIO/dynamorio/runs/1833435176?check_suite_focus=true
```
41/80 Test #252: code_api|tool.drcachesim.delay-global ........................***Failed  Required regular expression not found. Regex=[^Hit delay threshold: enabling tracing.
Hit -max_global_trace_refs: disabling tracing.
.*
     Total Number Of iterations   :  3
    ...................................................................
---- <application exited with code 0> ----
Basic counts tool results:
Total counts:
.*
           1 total threads
.*
$
]  0.18 sec
Hit delay threshold: enabling tracing.
---- <application exited with code 139> ----
Basic counts tool results:
```

Same thing for offline:
```
276/357 Test #277: code_api|tool.drcacheoff.max-global ..........................***Failed    0.64 sec
Running cmd |/home/runner/work/dynamorio/dynamorio/build_debug-internal-64/bin64/drrun;-s;90;-quiet;-debug;-killpg;-stderr_mask;0xC;-dumpcore_mask;0;-code_api;-t;drcachesim;-offline;-subdir_prefix;tool.drcacheoff.max-global;-trace_after_instrs;20K;-max_global_trace_refs;10K;--;/home/runner/work/dynamorio/dynamorio/build_debug-internal-64/suite/tests/bin/client.annotation-concurrency;libclient.annotation-concurrency.appdll.so;A;4;64;3|
Running postcmd |/home/runner/work/dynamorio/dynamorio/build_debug-internal-64/clients/bin64/drcachesim;-indir;/home/runner/work/dynamorio/dynamorio/build_debug-internal-64/suite/tests/tool.drcacheoff.max-global.client.annotation-concurrency.22882.5929.dir;-simulator_type;basic_counts|
CMake Error at /home/runner/work/dynamorio/dynamorio/suite/tests/runmulti.cmake:130 (message):
  output |Hit delay threshold: enabling tracing.

  Basic counts tool results:
<...>
  | failed to match expected output |Hit delay threshold: enabling tracing.

  Hit -max_global_trace_refs: disabling tracing.

  .*
```

derekbruening:
Happened again, on 2 runs in a row: https://github.com/DynamoRIO/dynamorio/pull/4721/checks?check_run_id=1841880860

[...]

PR #4722, opened 2021-02-06
i#2666 dyn cxt: Fix early context free

derekbruening:
Fixes a bug in PR #4703 where for post-CreateUserProcess handling it
frees the heap buffer for a CONTEXT before the code is finished using
it.

Also updates the CONTEXT heap buffers from PR #4703 to use
thread-private heap where possible, to avoid locks.

Issue: #2666


derekbruening:
Hmm, hit several issues: the max-global which is getting annoyingly common: #4711; and decode-stress which is now seeming to sometimes hit failures on Linux and Windows: #4723.

Issue #4719, opened 2021-02-05
Add interoperability with QEMU for arm emulation


derekbruening:
To aid in testing on ARM (and AArch64, for that matter) it would be helpful to be able to run DR under the QEMU emulator.  Today that does not work out of the box.  It is not clear whether there are only a handful of surface issues, or whether deeper problems will be hit in both DR and QEMU and we would need significant changes to both.

derekbruening:
Some problems that have been observed include:

A crash in qemu:
```
qemu: uncaught target signal 11 (Segmentation fault) - core dumped 
```

An assert in DR:
```
<Application xxx (zzz).  Internal Error: DynamoRIO debug check failure: core/unix/signal.c:1469 rc == 0 IF_VMX86(|| (sig >= 63 && rc == -EINVAL))
```

Another assert in DR, this one using unmodified versions of everything (the above are local qemu environments; not quite sure what all the local changes are):
```
$ qemu-arm -L /usr/arm-linux-gnueabihf bin32/drrun -- suite/tests/bin/simple_app
<Application /home/bruening/dr/git/build_a32_dbg_tests/suite/tests/bin/simple_app (4017678) DynamoRIO usage error : Failed to read ELF interpreter headers.>
<Usage error: Failed to read ELF interpreter headers. (/home/bruening/dr/git/src/core/unix/loader.c, line 2000)
version 8.0.18663, custom build
-early_inject -emulate_brk 
/home/bruening/dr/git/build_a32_dbg_tests/lib32/debug/libdynamorio.so=0x71000000>
```

derekbruening:
Quick analysis of the final assert above and what happens after it:

DR needs to be passed a root dir for the dynamic loader (just like QEMU's -L).  But we have a problem: we don't init DR and its options until *after* we map the app's interpreter.  The comment says we want the app mapped, and we want dr_get_proc_address() to work. Hmm. Maybe split up the init so we only initialize heap and options first?  Or maybe it would work out to do full init before mapping the interpreter.

Hardcoding a fix there to see what's next, DR initializes but then hangs trying to take over a non-app thread: a QEMU-added thread?

```
<Attached to 1/2 threads in application /home/bruening/dr/git/build_a32_dbg_tests/suite/tests/bin/simple_app (4167127)>
```

derekbruening:
Hardcoding ignoring the other thread, we then hit a problem where the app's loader can't find libc.  But code cache operation seems to work well as it builds 727 blocks:
[...]

PR #4726, opened 2021-02-09
i#4719 qemu: Add xarch_root option for QEMU

derekbruening:
Adds a new option -xarch_root which sets a path that is prepended to:
+ The application executable's interpreter, if the original does not exist.
+ SYS_openat paths, if the original does not exist.
+ System paths ued for loading private libraries: here the prefix is prepended
  before checking whether the original exists.

Splits dynamorio_app_init() into two pieces in order to have the
options set up at the time the loader maps the interpreter, while
avoiding ordering problems with the rest of the initialization.

The new option also auto-sets -ignore_takeover_timeout for
convenience, as that is always needed when running under QEMU.

Manually tested in cross-compile AArchXX setups on a Debian system.
Test suite integration is forthcoming.

Issue: #4719

### DIFF
--- a/core/drlibc/drlibc_module_elf.c
+++ b/core/drlibc/drlibc_module_elf.c
@@ -1,5 +1,5 @@
 /* *******************************************************************************
- * Copyright (c) 2012-2020 Google, Inc.  All rights reserved.
+ * Copyright (c) 2012-2021 Google, Inc.  All rights reserved.
  * Copyright (c) 2011 Massachusetts Institute of Technology  All rights reserved.
  * Copyright (c) 2008-2010 VMware, Inc.  All rights reserved.
  * *******************************************************************************/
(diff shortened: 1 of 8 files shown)


derekbruening:
The failure is the low4G test.  The problem is that the heap_in_lower_4GB option is now set during interp mapping, and it's used in os_map_file to add MAP_32BIT.  DR's vmm is taking up a bunch of the low 2G and it seems the kernel won't put ld.so below it:
```
$ cat /proc/1024986/maps
3fffe000-40006000 rw-p 00000000 00:00 0 
40006000-40007000 ---p 00000000 00:00 0 
40007000-40009000 rw-p 00000000 00:00 0 
40009000-4000f000 ---p 00000000 00:00 0 
4000f000-40010000 rw-p 00000000 00:00 0 
40010000-40017000 ---p 00000000 00:00 0 
40017000-40018000 rw-p 00000000 00:00 0 
40018000-7ffff000 ---p 00000000 00:00 0 
7ffff3b6e000-7ffff3b6f000 r--p 00000000 fe:01 7371509                    /home/bruening/dr/git/build_x64_dbg_tests/suite/tests/bin/simple_app
...
```

Not sure of the cleanest fix...avoid the MAP_32BIT just for this mapping via some special flag or sthg?

abhinav92003 (approved):

derekbruening:
Failures are the delay-global pair #4711.  But, going to wait for the fix for PR #4725 on attaching to large apps before merging anything else including this PR.

derekbruening:
There are enough additions since the review it may be worth another look.

abhinav92003 (approved):

PR #4727, opened 2021-02-09
i#4716 Enable api.static_noclient test on AArch64

AssadHashmi:


derekbruening:
Looks like api.static_noclient failed with the same assertion as in the issue.

derekbruening:
@AssadHashmi -- Going through still-open PR's: looks like this test is still disabled on AArch64.  Maybe you could revive this and enable it?

AssadHashmi:
Gah! Replaced by https://github.com/DynamoRIO/dynamorio/pull/5470.

PR #4729, opened 2021-02-11
Avoid blind syslog during init

derekbruening:
Avoids printing of an internal warning during early initialization for
single-bitwidth setups regardless of -stderr_mask by moving options
init even earlier.

To avoid DR heap init messing up the app's brk setup, moves heap init
out of the options init and into the later half.  This undoes the
early heap init from PR #4726, which is worked around by switching to
a stack buffer for -arch_init.  This seems safer in any case, delaying
heap init and client lib loads until after the app's interpreter is
loaded.

Issue: #4719

### DIFF
--- a/core/config.c
+++ b/core/config.c
@@ -624,21 +624,6 @@ d_r_config_init(void)
     config.u.v = &myvals;
     config_read(&config, NULL, 0, CFG_SFX);
     config_initialized = true;
-
-#    ifndef NOT_DYNAMORIO_CORE_PROPER
-    /* i#1271: to avoid leaving a stale 1config file behind if this process
-     * crashes w/o a clean exit, we give up on re-reading the file and delete
-     * it now.  It's an anonymous file anyway and not meant for manual updates.
-     * The user could override the dynamic_options by re-specifying in
-     * the option string, if desired, and re-create the 1config manually.
-     */
-    if (config.has_1config) {
-        INFO(2, "deleting config file %s", config.fname_app);
-        os_delete_file(config.fname_app);
-        dynamo_options.dynamic_options = false;
-    }
-    /* we ignore otherarch having 1config */
-#    endif
 }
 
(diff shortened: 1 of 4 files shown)


abhinav92003 (approved):

derekbruening:
x86 failures is #4711 delay-global & co. yet again.
The Jenkins a64 failures: do not yet know what that is.

derekbruening:
The Jenkins failures I can reproduce on the tx1: it seems that DR is reloaded every time because `!dynamorio_lib_gap_empty()` and somehow the reload causes it to no longer parse the options with this PR's shift in option parsing: which makes no sense because it should start completely over.  Still trying to figure it out (and why the gap is always non-empty).

derekbruening:
> why the gap is always non-empty).

Here's why: the kernel always put vvar+vdso in between:
```
derek@tx1:~$ cat /proc/9442/maps
ffff82998000-ffff82d96000 r-xp 00000000 08:03 5374901                    /home/derek/dr/build/lib64/debug/libdynamorio.so
ffff82da4000-ffff82da5000 r--p 00000000 00:00 0                          [vvar]
ffff82da5000-ffff82da6000 r-xp 00000000 00:00 0                          [vdso]
ffff82da6000-ffff82dca000 rw-p 003fe000 08:03 5374901                    /home/derek/dr/build/lib64/debug/libdynamorio.so
ffff82dca000-ffff82e02000 rw-p 00000000 00:00 0 
ffffeee05000-ffffeee26000 rw-p 00000000 00:00 0                          [stack]
```

derekbruening:
d_r_config_init() deletes the 1config file is the culprit:
```
deleting config file /home/derek/.dynamorio/linux.mmap.9609.1config64
options_init: incoming |-code_api "-stderr_mask" "0"|
Reloading
options_init: incoming ||
```

PR #4730, opened 2021-02-11
i#4711: Revert x86-related changes in #4677

sapostolakis:
Revert certain x86-related changes from #4677. In particular, we avoid restoring the arithmetic flags and (if used) the scratch register before the call to `hit_instr_count_threshold`, which might not return. 
This fix is necessary but it seems to be causing failures in `drcachesim.delay-global` and `drcacheoff.max-global` on x86-64 CI testing. These failures are not reproducible outside CI testing and are (so far) inexplicable based on manual analysis of the instrumented assembly code (see https://github.com/DynamoRIO/dynamorio/issues/4711#issuecomment-774776247). 
Thus, we are leaving x86_64 as technically broken to keep our tests green until the source of instability is found.

Issue: #4711

### DIFF
--- a/clients/drcachesim/tracer/tracer.cpp
+++ b/clients/drcachesim/tracer/tracer.cpp
@@ -1502,7 +1502,6 @@ event_delay_app_instruction(void *drcontext, void *tag, instrlist_t *bb, instr_t
     /* hit_instr_count_threshold does not always return. Restore scratch registers and
      * aflags.
      */
-
     drreg_statelessly_restore_app_value(drcontext, bb, scratch1, instr, instr, NULL,
                                         NULL);
     drreg_statelessly_restore_app_value(drcontext, bb, scratch2, instr, instr, NULL,
--- a/clients/drcachesim/tracer/tracer.cpp
+++ b/clients/drcachesim/tracer/tracer.cpp
@@ -1326,6 +1326,11 @@ static void *schedule_tracing_lock;
 /* XXX we don't have the inlining implemented yet for 32-bit architectures. */
 #endif
 
+#if defined(X86_64)
+/* FIXME i#4711: Do not restore aflags and register to prevent instability. */
+#    define DISABLED_FOR_BUG_4711 1
+#endif
+
 static dr_emit_flags_t
 event_delay_bb_analysis(void *drcontext, void *tag, instrlist_t *bb, bool for_trace,
                         bool translating, void **user_data);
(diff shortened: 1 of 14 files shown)


sapostolakis:
The x86-64 tests seem to always succeed. In addition to the testing in this PR, I did a dozen manual successful runs of the ci-x86 workflow.

abhinav92003:
If it helps, the test suite can be re-run at https://github.com/DynamoRIO/dynamorio/actions, and also from the details page for any existing CI run. I added this to the Wiki page now -- https://github.com/DynamoRIO/dynamorio/wiki/Test-Suite#continuous-integration-ci-via-github-actions.

sapostolakis:
> If it helps, the test suite can be re-run at https://github.com/DynamoRIO/dynamorio/actions, and also from the details page for any existing CI run. I added this to the Wiki page now -- https://github.com/DynamoRIO/dynamorio/wiki/Test-Suite#continuous-integration-ci-via-github-actions.

Yeah I realized at some point last week that I could manually force runs in github actions.
Good call updating the wiki.

derekbruening:
>￼ sapostolakis requested a review from derekbruening 5 days ago

Hmm, I missed this: because it's still marked as "Draft"?

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1500,19 +1500,10 @@ event_delay_app_instruction(void *drcontext, void *tag, instrlist_t *bb, instr_t
                              opnd_create_reg(scratch1)));
     MINSERT(bb, instr,
             XINST_CREATE_jump_cond(drcontext, DR_PRED_LT, opnd_create_instr(skip_call)));
-#        endif
 
     /* hit_instr_count_threshold does not always return. Restore scratch registers and
      * aflags.
      */
-#        ifdef X86_64
```
derekbruening, line 1508: I'd expect a XXX/FIXME/TODO #4711 comment explaining that this needs to be done for x86 too, but this fix seems to cause instability and so we're leaving x86 as technically broken to keep our tests green until someone is able to figure out what is going on.

derekbruening: I would put this type of explanation in the commit description too, and in the (left open) issue: we're leaving x86 with buggy code b/c the fix seems to cause problems we do not yet understand.

derekbruening (commented):
Please un-mark as Draft.

derekbruening:
Oh I see: you only just now marked as Draft.  And I do see the email.  So I just missed it before.

clients/drcachesim/tracer/tracer.cpp
sapostolakis: Got it. I converted it back to a draft until I finished testing it.

sapostolakis: Done.

sapostolakis:
I did a dozen CI runs and there is no failure for x86_64. So this PR should be good to go.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1506,12 +1506,19 @@ event_delay_app_instruction(void *drcontext, void *tag, instrlist_t *bb, instr_t
      * aflags.
      */
 #        ifdef X86_64
+    /* FIXME i#4711: Need to restore for x86 the arithmetic flags and (if used) the
+     * scratch register before the call to hit_instr_count_threshold. However, this fix
+     * seems to cause instability. So, we're leaving x86 as technically broken
+     * to keep our tests green until the source of instability is found.
+     */
+    /*
```
derekbruening, line 1514: style: No commented-out code.  Use #if, e.g. `#if DISABLED_FOR_BUG_4711`

derekbruening (approved):
Also not sure we want to hold off on merging until https://github.com/DynamoRIO/dynamorio/issues/4731#issuecomment-780127385 is fixed.

clients/drcachesim/tracer/tracer.cpp
sapostolakis: Done.

sapostolakis:
I will wait for #4731 before merging.

PR #4733, opened 2021-02-12
i#4731: Properly handle non-empty segment gaps in raw2trace

### LINKED ISSUE #4731: CRASH in raw2trace module mapping when VDSO in between segments of a library
With a proprietary app we're seeing crashes running opcode_mix.  It turns out to be from ld-2.27.so with segments spanning the VDSO:
```
 24,  24, 0x0000ffffbd3cc000, 0x0000ffffbd3ed000, 0x0000ffffbd3d0900, 0000000000000000, v#1,0, ld-2.27.so
 25,  24, 0x0000ffffbd3fc000, 0x0000ffffbd3fe000, 0x0000ffffbd3d0900, 0000000000020dd0, v#1,0, ld-2.27.so
 26,  24, 0x0000ffffbd40d000, 0x0000ffffbd410000, 0x0000ffffbd3d0900, 0000000000021750, v#1,0, ld-2.27.so
 27,  27, 0x0000ffffbd40c000, 0x0000ffffbd40d000, 0x0000ffffbd40c300, 0000000000000000, v#1,4096
```

```
  last_trace_module_start = 0xffffbd3cc000,
  last_trace_module_size = 0x44000,
  last_mapped_module_start = 0x7ffde94f4000

  modvec_ = std::vector of length 28, capacity 32 = {{            
  <...>
    }, {                        
      path = 0x54bb2dfcd588,  
      orig_base = 0xffffbd3cc000,
      map_base = 0x7ffde94f4000,
      map_size = 0x44000,                           
      is_external = 0x0          
    }, {                        
      path = 0x54bb2dfcd7c0, 
      orig_base = 0xffffbd3cc000,
      map_base = 0x7ffde94f4000, 
      map_size = 0x0,                     
      is_external = 0x0          
    }, {                                            
      path = 0x54bb2dfcd9f8,
      orig_base = 0xffffbd3cc000, 
      map_base = 0x7ffde94f4000, 
      map_size = 0x0,        
      is_external = 0x0 
    }, {                         
      path = 0x54bb2dfcdc30,
      orig_base = 0xffffbd40c000, 
      map_base = 0x54bb39395e94,
      map_size = 0x1000, 
      is_external = 0x1       
    }},                                                           
  last_orig_base_ = 0xffffbd3cc000,
  last_map_size_ = 0x44000,
  last_map_base_ = 0x7ffde94f4000,

(gdb) p/x 0xffffbd3cc000 + 0x44000 
$6 = 0xffffbd410000
```

So it is a bug in raw2trace handling an .so with separated segments with
VDSO in between them.  Surprising, since we put in effort to have
per-segment instead of per-library module entries, and since DR's core has had
several fixes to support things in ELF gaps in years past.

derekbruening:
Switches the module_mapper_t used by raw2trace to explicitly track
individual segments, rather than collapsing them into the single
whole-module mmap, in order to handle objects like vdso allocated in
gaps between the segments.

Tested on a proprietary system where vdso was in the middle of ld.so,
causing the opcode_mix tool to crash.  (Creating a regression test
would be complex, requiring synthesizing ELF files with the required
characteristics.)

Fixes #4731

### DIFF
--- a/clients/drcachesim/tracer/raw2trace.cpp
+++ b/clients/drcachesim/tracer/raw2trace.cpp
@@ -1,5 +1,5 @@
 /* **********************************************************
- * Copyright (c) 2016-2020 Google, Inc.  All rights reserved.
+ * Copyright (c) 2016-2021 Google, Inc.  All rights reserved.
  * **********************************************************/
 
 /*
@@ -167,8 +167,8 @@ module_mapper_t::~module_mapper_t()
     user_free_ = nullptr;
     for (std::vector<module_t>::iterator mvi = modvec_.begin(); mvi != modvec_.end();
          ++mvi) {
-        if (!mvi->is_external && mvi->map_base != NULL && mvi->map_size != 0) {
-            bool ok = dr_unmap_executable_file(mvi->map_base, mvi->map_size);
+        if (!mvi->is_external && mvi->map_seg_base != NULL && mvi->total_map_size != 0) {
+            bool ok = dr_unmap_executable_file(mvi->map_seg_base, mvi->total_map_size);
             if (!ok)
                 WARN("Failed to unmap module %s", mvi->path);
         }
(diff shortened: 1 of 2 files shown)


derekbruening:
Looks like the sourceforge server is down again, making the Windows runs fail.  Maybe we should store a copy of that doxygen package somewhere more reliable: the license should allow it.

abhinav92003 (approved):

derekbruening:
Windows 64-bit failure is burst_static callback assert #4486.

Issue #4425, opened 2020-08-27
Stack overflow during signal handling in AArch64


abhinav92003:
While working on PR #4397, I found that the `burst_flush_aarch64` test crashes with a SIGSEGV when `signal_stack_size = 32K `(which is the value automatically set by DR after adjustment). The crash is due to a stack overflow and is limited to debug builds.

Note that `burst_flush_aarch64` intentionally causes a SIGILL too, which is handled as expected by the test and doesn't cause any crash.

**To Reproduce**
1. Remove custom `-signal_stack_size 64K` in `clients/drcachesim/tests/burst_flush_aarch64.cpp`. 
2. Build with debug on.
3. Run `clients/bin64/tool.drcacheoff.burst_flush_aarch64`

**Expected behavior**
The SIGILL thrown by the test is expected and is handled too. But the crashing SIGSEGV is unexpected.

**Screenshots or Pasted Text**

Details in GDB:
```
(gdb) r
Starting program: /home/abhinavas/dr/build/i4328-2_debug/clients/bin64/tool.drcacheoff.burst_flush_aarch64 
pre-DR init
pre-DR start

Program received signal SIGILL, Illegal instruction.
0x0000aaaa6aaba448 in ?? ()
(gdb) c
Continuing.

Program received signal SIGSEGV, Segmentation fault.
safe_read_asm_pre () at /home/abhinavas/dr/src/i4328-2/core/arch/aarch64/aarch64.asm:416
416	        strb     w3, [ARG1]
(gdb) c
Continuing.

Program received signal SIGSEGV, Segmentation fault.
0x0000aaaaaab8c9fc in d_r_notify (priority=SYSLOG_NONE, internal=false, synch=false, substitution_num=0, prefix=0x0, fmt=0x0) at /home/abhinavas/dr/src/i4328-2/core/utils.c:1884
1884	{
(gdb) disassemble
Dump of assembler code for function d_r_notify:
   0x0000aaaaaab8c9f8 <+0>:	sub	sp, sp, #0x900
=> 0x0000aaaaaab8c9fc <+4>:	stp	x29, x30, [sp,#-16]!
   0x0000aaaaaab8ca00 <+8>:	mov	x29, sp
```
[...]

PR #4735, opened 2021-02-16
i#4425: sigaction instead of signal in burst_aarch64_sys test

sapostolakis:
Prevents stack overflow in signal processing for the burst_aarch64_sys test without increasing the size of the signal stack.
The implementation of `signal` varies across different systems and involves unspecified behavior for various scenarios.
Instead, it is universally recommended to use `sigaction`. 
This signal overflow might not have been a problem for other tests/applications because most of them are using `sigaction`.

Fixes: #4425

### DIFF
--- a/clients/drcachesim/tests/burst_aarch64_sys.cpp
+++ b/clients/drcachesim/tests/burst_aarch64_sys.cpp
@@ -177,9 +177,7 @@ static std::string
 gather_trace()
 {
     if (!my_setenv("DYNAMORIO_OPTIONS",
-                   // XXX i#4425: Fix debug-build stack overflow issue and
-                   // remove custom signal_stack_size below.
-                   "-stderr_mask 0xc -signal_stack_size 64K "
+                   "-stderr_mask 0xc "
                    "-client_lib ';;-offline'"))
         std::cerr << "failed to set env var!\n";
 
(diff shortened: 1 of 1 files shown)


derekbruening:
Per offline discussion:

- DR should handle anything an application does, so even if signal() is not recommended, we would like DR to handle it and not crash.
- It would be good to understand why the restorer is an invalid address for signal().  If a64 signal() is always using restorer to mean something else, or is leaving it uninitialized, it would be best to change DR to not try to read it and thus not fault.
- Nesting could happen for other reasons than a fault on reading the restorer: is SIGUSR2 unblocked in DR's handler?  If so, and only SIGSEGV and SIGBUS are unblocked, then nesting would only be on safe reads at various points (restorer, and IIRC setting up the app frame, and x86 TLS reads, maybe more) or DR bugs, but we would still like to understand whether the current signal stack size is too small for a 2nd signal on just a64, or on x86 too.
- If the stack size is too small for a 2nd signal on a64 and maybe x86 too, can we easily reduce the stack usage of record_pending_signal, or move something to signal-safe special heap?  Or should we increase the default stack size?

sapostolakis:
Issue #4425 was resolved with PRs #4840 and #4888.

PR #4745, opened 2021-02-19
Remove IA-32 from general IR statements

derekbruening:
Removes several cases where the documentation about the IR says
"IA-32" when it means all the ISA'ss we support.


AssadHashmi (approved):
Looks fine.

derekbruening:
Win64 burst_static and burst_replace hit #4486.

Issue #4627, opened 2020-12-16
ASSERT sigaction failure when DR run under QEMU emulating AArch64


derekbruening:
We're seeing this assert at startup on sigaction failing when running DR under QEMU emulating AArch64 in user mode:
```
<Application xxx.  Internal Error: DynamoRIO debug check failure: dynamorio/trunk/core/unix/signal.c:1469 rc == 0 IF_VMX86(|| (sig >= 63 && rc == -EINVAL))
(Error occurred @0 frags)
```

I'll gather more info such as QEMU version, and try to repro with a standalone upstream QEMU.

derekbruening:
We see this with the QEMU on GA CI Ubuntu 16.04 and 20.04: https://github.com/DynamoRIO/dynamorio/pull/4747#issuecomment-782385379
Those are QEMU 2.x and 4.2.

Yet, I cannot reproduce it with older QEMU on my laptop.
I locally built v3.0.0 (took some patching to successfully configure and build it w/ a newer compiler and glibc) and v4.2.0 and neither reproduces the assert:
```
$ git checkout -b v3.0.0 v3.0.0
$ cd ../build_arm
$ ../src/configure --prefix=${PWD} --enable-linux-user --target-list=arm-linux-user
$ make -j
$ cd ~/dr/git/build_a32_dbg_tests
$ ~/extsw/qemu/build_arm/arm-linux-user/qemu-arm -L /usr/arm-linux-gnueabihf bin32/drrun -xarch_root /usr/arm-linux-gnueabihf -- suite/tests/bin/simple_app
replacing interpreter |/lib/ld-linux-armhf.so.3| with |/usr/arm-linux-gnueabihf//lib/ld-linux-armhf.so.3|
<Starting application ~/dr/git/build_a32_dbg_tests/suite/tests/bin/simple_app (688316)>
<Initial options = -no_dynamic_options -xarch_root '/usr/arm-linux-gnueabihf' -code_api -stack_size 56K -signal_stack_size 32K -max_elide_jmp 0 -max_elide_call 0 -early_inject -emulate_brk -no_inline_ignored_syscalls -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct -unsafe_ignore_takeover_timeout -takeover_timeout_ms 0x190 >
<Paste into GDB to debug DynamoRIO clients:
set confirm off
[...]

PR #4747, opened 2021-02-19
i#4719 qemu: Add support for running tests under QEMU

derekbruening:
When cross-compiling, inserts QEMU commands into each test command line.
Increases the test timeouts by 4x to account for emulation overhead.

Adds GA CI support by installing QEMU and enabling running tests for
the AArchXX cross-compilation jobs.  For now, limits the tests to
those marked with a new label RUNS_ON_QEMU, which starts out added to
the ~1/3 of tests that currently pass.  As we get more tests to work
we may want to separate the jobs if they take too much time.

Issue: #4719

### DIFF
--- a/.github/workflows/ci-aarchxx.yml
+++ b/.github/workflows/ci-aarchxx.yml
@@ -48,10 +48,9 @@ defaults:
     shell: bash
 
 jobs:
-  # AArchXX cross-compile with gcc, with some tests run under QEMU.
-  # XXX i#4719: As we add more tests, if this job takes too much time we
-  # should split it up.
-  aarchxx-cross-compile:
+  # AArch64 cross-compile with gcc, with some tests run under QEMU.
+  # We use a more recent Ubuntu for a more recent QEMU.
+  aarch64-cross-compile:
     runs-on: ubuntu-20.04
 
     steps:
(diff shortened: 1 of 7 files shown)


derekbruening:
Hmm, this passes on my local machine, which is a debian variant with qemu 5.2.0.  But here it hits two problems:

```
2021-02-19T20:21:04.9375582Z 6: qemu: Unsupported syscall: 398
2021-02-19T20:21:04.9377281Z 6: <Application /home/runner/work/dynamorio/dynamorio/build_arm-debug-internal-32/suite/tests/bin/common.broadfun (14909).  Internal Error: DynamoRIO debug check failure: /home/runner/work/dynamorio/dynamorio/core/unix/signal.c:1580 rc == 0 IF_VMX86(|| (sig >= 63 && rc == -EINVAL))
```

The assert is #4627 which was also seen on a different machine in the past.  It's sigaction failing, which is weird.  Related to this being an old qemu?  If we can repro we can see what signal it is.

The unknown syscall is a newly seen problem: 398 (or 293 for aarch64) is rseq.  I'm guessing it's an older QEMU since my 5.2.0 doesn't complain.  It doesn't seem to be fatal: though it might cause golden-output tests to fail?  So we may have to get a newer QEMU on the system.

This is what was installed:
```
2021-02-19T20:18:42.4340526Z Setting up qemu-user (1:2.5+dfsg-5ubuntu10.49) ...
2021-02-19T20:18:42.4565040Z Setting up qemu-user-binfmt (1:2.5+dfsg-5ubuntu10.49) ...
```

derekbruening:
GA has Ubuntu 20.04 which has qemu-user 4.2.  I tried it and indeed the rseq message goes away.  But the
sigaction assert is still there!  I guess I'll try to set up a VM to repro and see what is going on.

derekbruening:
OK, now we have green aarch tests under emulation.  It's only 1/3 of the tests but it is a great starting point:
```
arm-debug-internal-32: all 68 tests passed
arm-debug-internal-64: all 76 tests passed
```

This pushes the aarchxx-cross-compile job to 20m so maybe we should split it.

derekbruening:
OK it is all good except the old job name is still listed as required.  I went and edited the required jobs settings but it doesn't seem to be affecting what was already run here so ignore the uncompleted job for this PR.

abhinav92003 (approved):

abhinav92003:
This is great! Thanks for adding Qemu for ARM32. 

So this new suite is `aarch32-cross-compile` right? It's showing yellow.

derekbruening:
> So this new suite is aarch32-cross-compile right? It's showing yellow.

Oh that's the name I gave it in the first commit that split up 32 and 64, but then I changed the name arm-cross-compile to match our internal conventions, and that one is green and taking 10 mins (just like aarch64-cross-compile).  Interesting that it has the aarch32 one marked as required -- at first glance I thought that was the old aarchxx-cross-compile.  Maybe it inherited from that one?  In any case I think a new PR will not show the old name and will be all good.

derekbruening:
> This is great! Thanks for adding Qemu for ARM32.

Thanks for pushing for it -- I had assumed it would be a lot more work than it was.  Though I think to get the other 2/3 of the tests will be a lot of work: I think there's a combination of DR bitrot on ARM, work to get QEMU into command lines, flakiness/slowness exaggerted by QEMU, and QEMU bugs.  But no rush on that: just having these ~70 tests is a great sanity check to avoid completely breaking ARM, and another validation layer for AArch64 beyond Jenkins.

derekbruening:
> > So this new suite is aarch32-cross-compile right? It's showing yellow.
> 
> Oh that's the name I gave it in the first commit that split up 32 and 64, but then I changed the name

Turns out I marked the wrong name as required in the github settings which is why it was still listed.  Fixed now.

Issue #4111, opened 2020-02-16
Merge website, doxygen pages, and wiki?


derekbruening:
Split from #3194 

Most of our user-facing content is prose inside the doxygen pages, leaving just a few pages on the main website. And the two have completely different styles and layout. We could consider merging in one direction or another: we could remove all the non-API content from doxygen control and move it to the Pages and embed the API content in an iframe (maybe remove the navbar, etc. there); or we could go the other way and try to put all the top-level pages inside doxygen auto-generation.

We could also consider removing the wiki and moving its content into the website, to have everything integrated in one place.

derekbruening:
The downside of moving content out of doxygen is that the local docs in a package or local build would then lose that content.  However, I do not know that the typical user looks at the local docs in any case.  It may be better to have nicer online docs.

derekbruening:
@AssadHashmi , @johnfxgalea -- looking for some more opinions.

Also, now that the website is editable by others in the dynamorio.github.io repository, it can be improved by the community: e.g., #4112, or other style changes.

derekbruening:
Another downside of moving the tool docs is that they would no longer live next to their code: e.g., instead of `drcachesim.dox.in` right in the `drcachesim/` code directory, it would live in the separate `dynamorio.github.io` repository.  (I suppose we could have `dynamorio` be a submodule of the docs repo and have jekyll pull out just some targeted tool doc files.)

derekbruening:
An upside is that we could add a search box over all the docs, versus the current search box in the doxygen-generated html which searches only the code tokens and does not search the prose at all.

derekbruening:
I'm leaning toward:

1) Moving all the wiki pages into the website and removing the wiki.

2) Putting in the work to integrate the doxygen output into the jekyll site.  This would involve removing the side frame and having doxygen generate plain html pages, pushing them into a subdir of dynamorio.github.io with a script from Travis, and building up navigation menus using the doxygen file names.  This could allow us to keep the content in both release packages and online.

derekbruening:
My new plan is to duplicate:

+ Keep the wiki, but place a copy of it (updated w/ the weekly cronbuild) inside the web site.
+ Generate both standalone framed doxygen html and doxygen suitable for embeeding.

I put some work into integration of doxygen and put it up on dynamorio.org.

I have a menu system using doxygen's own javascript modified to be extendable (had to scrap their "skip list" indexing and inline all the submenus).

[...]

Issue #4740, opened 2021-02-17
New web documentation search seems not working


SweetVishnya:
Hi!

Could you, please, return the old flavored Doxygen search for DynamoRIO reference web site? The new one fails to find functions.

https://dynamorio.org/docs/index.html

https://dynamorio.org/search?query=dr_mutex_lock

derekbruening:
It does find the page containing the function `dr_mutex_lock`, but it doesn't zero in on the anchor within the page since it currently searches at a page level.  The new search and the new doxygen integration into the top-level framework is a work-in-progress.

The old non-integrated html is still there at https://dynamorio.org/dynamorio_docs/ (just not linked to).

The doxygen search has several major shortcomings:
+ If you don't know what a symbol *starts with* you cannot find it: you cannot search for "lock" and find dr_mutex_lock which is pretty bad.
+ It does not search any of the prose, and we have extensive prose docs beyond just the function/data structure reference.
This will get worse as the wiki content is also integrated into the same site.

If you have experience in simple search solutions for jekyll-based sites and would like to help out that would be appreciated.

Having both searches there would just be confusing I think?  Tweaking the lunr-based search to be over menu entries would be an ideal solution: but while the entries are there in a JS array, the content for each is not: it would have to parse the page and stop at the next menu item's entry point.  Maybe the doxygen can be changed to make every sub-page section or subsection into its own page, which would then have the page list passed to lunr find things, as well as making the site nicer on mobile with smaller pages.

derekbruening:
Xref #4111

derekbruening:
> Having both searches there would just be confusing I think?

Maybe one input string could be passed to both and the results displayed side by side or sthg if the javascript for the two can be arranged that way.

derekbruening:
Copying @johnfxgalea 's comment from https://github.com/DynamoRIO/dynamorio/issues/4111#issuecomment-780737033 over here as it directly relates to this search issue:

Another option might be to move the wiki & website to Docusaurus, which supports document search powered by Algolia DocSearch (or perhaps use this directly). However, I'm not sure if it works on HTML produced by doxygen, and might be an overkill anyways given our current setup.

https://v2.docusaurus.io/docs/search

derekbruening:
@SweetVishnya I generated a set of keywords targeting anchors from the doxygen html patterns: it should be better now, with e.g. dr_mutex_lock's top hit a direct link.

SweetVishnya:
[...]

PR #4748, opened 2021-02-21
i#4111: Generate doxygen suitable for embedding in jekyll

derekbruening:
Creates a second version of the doxygen html output suitable for
embedding in our jekyll-based web site.  This version turns off the
doxygen treeview, but then borrows the javascript menu files from the
treeview version.  We then edit the html to remove the headers, and
edit the menu files for inlining into our web site's single array.

Furthermore, we generate keyword search term data so that searches
using lunr for functions or types result in direct links to their
anchors within their pages.  It generates keyword tokens for each
component of an underscore-separated multi-word type, along with
tokens for each prefix (e.g., so that "dr_mutex" will find
"dr_mutex_*").

Adds installation into a "docs_embed" directory.

Adds deployment logic to package.cmake to copy the embedded html to a
top level "html_embed", along with new steps in the ci-package GA
workflow to take the embedded html files and push them to the
dynamorio.github.io repository.

Issue: #4111

### DIFF
--- a/.github/workflows/ci-package.yml
+++ b/.github/workflows/ci-package.yml
@@ -126,7 +126,7 @@ jobs:
 
     - name: Deploy Embedded Docs
       run: |
-        rsync -av --delete embed_html/ dynamorio.github.io/docs/
+        rsync -av --delete html_embed/ dynamorio.github.io/docs/
         cd dynamorio.github.io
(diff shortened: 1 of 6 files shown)


derekbruening:
This should also reference #4740 as it generates keywords for prefixes like `dr_mutex`.

PR #4751, opened 2021-02-21
i#xxxx: Stub out rseq_is_registered_for_current_thread if HAVE_RSEQ is false

khuey:
HAVE_RSEQ allows disabling all use of the rseq syscall at compile time, except for the function `rseq_is_registered_for_current_thread` which uses rseq to infer whether rseq is in use. If HAVE_RSEQ is false, just stub out `rseq_is_registered_for_current_thread` and always return false.

### DIFF
--- a/core/unix/rseq_linux.c
+++ b/core/unix/rseq_linux.c
@@ -269,6 +269,7 @@ rseq_shared_fragment_flushtime_update(dcontext_t *dcontext)
     rseq_clear_tls_ptr(dcontext);
 }
 
+#ifdef HAVE_RSEQ
 bool
 rseq_is_registered_for_current_thread(void)
 {
@@ -296,6 +297,13 @@ rseq_is_registered_for_current_thread(void)
     }
     return false;
 }
+#else
+bool
+rseq_is_registered_for_current_thread(void)
+{
+    return false;
+}
+#endif
 
 static void
 rseq_analyze_instructions(rseq_region_t *info)


core/unix/rseq_linux.c
```diff
@@ -269,6 +269,7 @@ rseq_shared_fragment_flushtime_update(dcontext_t *dcontext)
     rseq_clear_tls_ptr(dcontext);
 }
 
+#ifdef HAVE_RSEQ
```
derekbruening, line 272: Actually, it looks like HAVE_RSEQ is only used to determine whether the rseq.h header is around, which then only determines whether the regression test is built: it does not affect whether the core handles an application using rseq.  The core rseq mangling code and all the code in rseq_linux.c is always enabled.  I believe the idea is that we want DR built on one machine to work on as many others as possible, rather than requiring a custom build for every variant of kernel feature or whatnot.

What is the top-level problem being solved here?  If it requires disabling or enabling the core handling rseq, probably a separate option should be put in place and maybe HAVE_RSEQ renamed to make it clearer what its consequences are.

khuey: The underlying motivation here is to make it possible to record DR with the rr debugger. rr doesn't support rseq, but it also doesn't just return ENOSYS, it stops the program being recorded and outputs an error about the unsupported syscall.

derekbruening: So maybe as suggested a separate control: a CMake option ENABLE_RSEQ_SUPPORT or sthg, and hopefully it can have as few ifdefs in the code as possible; combined with possibly renaming HAVE_RSEQ to HAVE_RSEQ_HEADER?

khuey:
I made rr just return ENOSYS. https://github.com/rr-debugger/rr/commit/81decaeffe2f786624eec2d039bdb26eab52028a

PR #4752, opened 2021-02-23
i#4627: Work around QEMU signal number bugs

### LINKED ISSUE #4627: ASSERT sigaction failure when DR run under QEMU emulating AArch64

derekbruening:
QEMU prior to 5.0 has a signal number mapping bug where it ends up
failing its app's attempts to invoke sigaction for signals 33 and 64.
We relax DR's assert on sigaction failure when -xarch_root is set for
those numbers, trying to continue for best-effort execution with a
syslog warning.

Tested manually on 4.2.0 and on the forthcoming GA CI QEMU testing in
PR #4747.

Fixes #4627

### DIFF
--- a/core/unix/signal.c
+++ b/core/unix/signal.c
@@ -1571,13 +1571,25 @@ set_handler_and_record_app(dcontext_t *dcontext, thread_sig_info_t *info, int si
 
     /* Arm the signal. */
     rc = sigaction_syscall(sig, act, NULL);
-    ASSERT(rc ==
-           0
-           /* Workaround for PR 223720, which was fixed in ESX4.0 but
-            * is present in ESX3.5 and earlier: vmkernel treats
-            * 63 and 64 as invalid signal numbers.
-            */
-           IF_VMX86(|| (sig >= 63 && rc == -EINVAL)));
+    DODEBUG({
+        /* i#4627: QEMU before 5.0 has a signal mapping bug where 33 and 64
+         * fail.  Better to continue and hope 33 doesn't mess up pthreads.
+         */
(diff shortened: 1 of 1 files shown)


core/unix/signal.c
```diff
@@ -1571,13 +1571,25 @@ set_handler_and_record_app(dcontext_t *dcontext, thread_sig_info_t *info, int si
...
+         */
+        if (rc == -EINVAL && !IS_STRING_OPTION_EMPTY(xarch_root) &&
+            (sig == 33 || sig == 64)) {
+            SYSLOG_INTERNAL_WARNING("Failed to register signal handler for %d: assuming "
+                                    "under QEMU<5.0 and continuing",
+                                    sig);
+        } else {
+            ASSERT(rc ==
+                   0
+                   /* Workaround for PR 223720, which was fixed in ESX4.0 but
```
abhinav92003, line 1586: Out of context question: how do I look up these PR refs? They don't seem to be in the Github repo.

abhinav92003 (approved):

core/unix/signal.c
derekbruening: It's from the old vmware bug tracker unfortunately.  Maybe these lines could just be removed -- maybe a separate cleanup.

PR #4756, opened 2021-02-26
Add EVEX encodings of missing vmovq variants.

### LINKED ISSUE #4755: decoding misses certain EVEX VMOVQ variants
The f30f7e and 660fd6 VMOVQ variants, which don't have a corresponding W=0 VMOVD instruction, are missing their EVEX version in the decoder table.

e.g. `62 61 fd 08 d6 0c 0a` and `62 61 fe 08 7e 0c 0a`

khuey:
These instruction variants, which do not have a EVEX.W=0 vmovd equivalent, were missed when EVEX instructions were added. The XMM/mem forms of these instructions are redundant with existing encodings and can't be tested via the normal encoding API.

Fixes #4755.

### DIFF
--- a/core/ir/x86/decode_table.c
+++ b/core/ir/x86/decode_table.c
@@ -3866,11 +3866,11 @@ const instr_info_t prefix_extensions[][12] = {
     {OP_movd, 0x660f7e10, "movd", Ey, xx, Vd_dq, xx, xx, mrm, x, END_LIST},
     {INVALID, 0xf20f7e10, "(bad)", xx, xx, xx, xx, xx, no, x, NA},
     {INVALID,   0x0f7e10, "(bad)", xx, xx, xx, xx, xx, no, x, NA},
-    {OP_vmovq, 0xf30f7e10, "vmovq", Vdq, xx, Wq_dq, xx, xx, mrm|vex, x, tpe[61][6]},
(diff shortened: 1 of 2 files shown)


khuey:
I wonder if there's a good place to put the sequences from #4755. Conceptually they'd benefit from the same treatment as the binutils decoder test but they're not from binutils.

core/ir/x86/decode_table.c
```diff
@@ -3866,11 +3866,11 @@ const instr_info_t prefix_extensions[][12] = {
     {OP_movd, 0x660f7e10, "movd", Ey, xx, Vd_dq, xx, xx, mrm, x, END_LIST},
```
derekbruening, line 3866: Commit title is missing `i#4755: ` prefix.

derekbruening (approved):

derekbruening:
> I wonder if there's a good place to put the sequences from #4755. Conceptually they'd benefit from the same treatment as the binutils decoder test but they're not from binutils.

Subcase inside api/ir_x86.c?  New test that decode/encodes from asm?

khuey:
These ARM failures are some sort of infrastructure problem, btw.

derekbruening:
> These ARM failures are some sort of infrastructure problem, btw.

Assumed to be transient issue w/ the apt download -- also hit on another PR.  I will override and merge.

PR #4757, opened 2021-02-26
i#4111 web: Deploy web content from docs/ to root dir

derekbruening:
Use permalinks to deploy all the content from the docs/ web subdir
into the top level, rather than having docs/ in every single URL.

Issue: #4111


derekbruening:
Wow that is a lot of failures: but they are all flaky tests, just all at once, plus the qemu download issue.  Nothing related to docs.

We have:
x86-64: cleancallsig meta-instr faulted assert #1369
win-64: burst_static callback.c assert #4486
win-32: simple-config-file: #3320

Issue #4604, opened 2020-12-11
32 bit code_api|linux.thread-reset failed


khuey:
https://github.com/DynamoRIO/dynamorio/runs/1534429151

```test 87
        Start  87: code_api|linux.thread-reset

87: Test command: /home/runner/work/dynamorio/dynamorio/build_debug-internal-32/bin32/drrun "-s" "90" "-quiet" "-debug" "-killpg" "-stderr_mask" "0xC" "-dumpcore_mask" "0" "-enable_reset" "-reset_at_nth_thread" "2" "-code_api" "--" "/home/runner/work/dynamorio/dynamorio/build_debug-internal-32/suite/tests/bin/linux.thread"
87: Test timeout computed to be: 600
87: Sideline thread started
87: <Application /home/runner/work/dynamorio/dynamorio/build_debug-internal-32/suite/tests/bin/linux.thread (19212) DynamoRIO usage error : Is your client invoking raw system calls via vdso sysenter? While such behavior is not recommended and can create problems, it may work with the -sysenter_is_int80 runtime option.>
87: <Usage error: Is your client invoking raw system calls via vdso sysenter? While such behavior is not recommended and can create problems, it may work with the -sysenter_is_int80 runtime option. (/home/runner/work/dynamorio/dynamorio/core/dispatch.c, line 112)
87: version 8.0.18607, custom build
87: -no_dynamic_options -code_api -stderr_mask 12 -stack_size 56K -signal_stack_size 32K -max_elide_jmp 0 -max_elide_call 0 -reset_at_nth_thread 2 -early_inject -emulate_brk -no_inline_ignored_syscalls -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct 
87: 0x4806be98 0xf7c11f86
87: 0x4806bec8 0xf7bf93c6
87: 0x4806bee8 0xf7bfd1ed
87: 0x4806bf58 0xf7bf94ef
87: 0x4806bfe8 0x4808afce
87: 0xffbc6008 0xf7989647>
 88/343 Test  #87: code_api|linux.thread-reset ......................................***Failed  Required regular expression not found. Regex=[^Sideline thread started
i = 2500000
i = 5000000
i = 7500000
i = 10000000
i = 12500000
i = 15000000
i = 17500000
i = 20000000
i = 22500000
i = 25000000
Sideline thread finished
Waiting for child to exit
Child has exited
$
]  0.26 sec
Sideline thread started
[...]

PR #4763, opened 2021-03-01
Fix stale apt version with update

derekbruening:
Attempts to fix the recent aarchxx job failures by running "apt-get
update" prior to install.


abhinav92003 (approved):

derekbruening:
linux32 failure is thread-reset #4604.  Seems like it's time to add that to the flaky list since I assume nobody has time to try to fix it.

PR #4775, opened 2021-03-10
i#4774: Add preferred base to DR and drmodtrack interfaces

### LINKED ISSUE #4474: Bridge testing gap on AArch64

### LINKED ISSUE #4774: Add preferred base to DR and drmodtrack module data
For some use cases, the preferred base of a module is useful.  In our case, our internal symbolization uses the binutils addr2line convention of two different types of PC: either an absolute PC for a non-relocatable executable (i.e., with a non-zero preferred==required base), or an offset from the load base for a position-independent library or executable (which has a zero preferred base for ELF; this design does not seem very cross-platform though as other types of binaries have non-zero preferred bases yet are relocatable).  For this dual PC interface, there is not enough information in the module data recorded during tracing.  This is a feature request to augment the trace module data (which will also require augmenting DR's public module interface) with either the preferred base or a flag indicating whether the module was at its preferred base.

derekbruening:
Adds a preferred_base field to the module_data_t struct.  There is no
compatibility break as DR allocates all such structs.

Adds a preferred_base field to the drmodtrack_info_t struct.  There is
no compatibility break as it has a struct_size field.  Offline
drmodtrack files do incur a version bump here, so an old reader will
fail when reading a file with the new field.  The field is set to a -1
sentinel to aid in determining whether a file has the field or not.
This is also done here with the previously-added offset field.

Adds some sanity checks to the corresponding tests.

Fixes #4774

### DIFF
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -196,6 +196,8 @@ Further non-compatibility-affecting changes include:
  - Added instr_replace_reg_resize() and opnd_replace_reg_resize().
  - Added #DRX_COUNTER_REL_ACQ flag to optionally enable release-acquire semantics for
    drx_insert_counter_update() on AArchXX.
+ - Added a preferred base field to the #_module_data_t struct and to the
+   #_drmodtrack_info_t struct.
 
 **************************************************
 <hr>
--- a/core/lib/instrument.c
+++ b/core/lib/instrument.c
@@ -1,5 +1,5 @@
 /* ******************************************************************************
- * Copyright (c) 2010-2020 Google, Inc.  All rights reserved.
+ * Copyright (c) 2010-2021 Google, Inc.  All rights reserved.
  * Copyright (c) 2010-2011 Massachusetts Institute of Technology  All rights reserved.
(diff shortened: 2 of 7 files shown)


derekbruening:
Windows failure looks like a pre-existing bug that happened to not fail the test: #4777

derekbruening:
PR #4778 is fixing the Windows issue and will be merged into here afterward.

abhinav92003 (approved):

PR #2973, opened 2018-04-29
Revert "Revert (#2940)"

snehasish:
This reverts commit 48db566af975c96f984cc5060e78294ef99e0ce7 which
reverted the changes in PR #2940. The changes pushed caused some apps to
overflow the modidx field (issue #2956). PR #2969 increased the width of
the modidx field. We can now safely revert the revert.

PR #2940, opened 2018-04-18
Split segments and record segment offsets.

snehasish:
This change changes the modules.log format to provide a separate entry
for each segment instead of combining entries for contiguous segments
mapped from each image. The module file version is incremented to 4 and
adds a new field called offseti after the "entry" field.

The file offset information for unix is recorded in
module_add_segment_data and propagated to drmodtrack to be written out
with each segment.

Issue #2939, opened 2018-04-18
Record the physical offset of each segment in modules.log


snehasish:
In order to support internal symbolization tools we would like to record the physical offset of each mapped segment. drmodtrack combines multiple adjacent segments into a single contiguous blob so multiple physical offsets could be mapped to the entry in the modules.log file. Since the modules.log format may already contain multiple entries for non contiguous mapped regions, clients should already handle the case of multiple entries for a single module.

derekbruening:
The #2940 PR for this failed to use this issue number and was committed as 8c90e2dc81eb9b9dfdcf2313bdf91b02f12c4de4 so manually linking and closing here.

PR #4778, opened 2021-03-10
i#4777: Fix uninitialized drmodtrack offset on Windows

### LINKED ISSUE #4777: drmodtrack offset field has uninitialized values on Windows
For #2939 an offset field was added to drmodtrack in PR #2940 (and PR #2973).  However, it is ifdef-ed out for Windows in some places but not in others.  Now, I'm adding a new field for #4474, and drmodtrack-test is failing with uninit bogus printed offset values differing bet online and offline.  Not sure how it passed before: probably just got lucky and the uninit values matched.

It's too late to ifdef it out of the interface so maybe the best solution is to set it to 0 on Windows.

derekbruening:
PR #2940 and #2973 added an offset field to drmodtrack for #2939, but
they ifdef-ed the field in some places but not others, resulting in
uninitialized output.  Since it's already locked into the interface,
we always set it to 0 here and include it in internal structures to
fix the problem.

Tested on drmodtrack-test where the fields were manually confirmed to
no longer contain bogus values, and where the test doesn't fail when
i#4474 adds a new field that shifts the test buffer to avoid the
happens-to-match scenario that masked this bug in that test before.

Issue: #2939, #4474, #4777
Fixes #4777

### DIFF
--- a/ext/drcovlib/modules.c
+++ b/ext/drcovlib/modules.c
@@ -459,8 +459,9 @@ module_table_entry_print(module_entry_t *entry, char *buf, size_t size)
 #endif
     read_entry.path = full_path;
     read_entry.custom = entry->custom;
-    // For unices we record the physical offset from the backing file
-    // (always 0 on Windows).
+    /* For unices we record the physical offset from the backing file
+     *(always 0 on Windows).
+     */
     read_entry.offset = entry->offset;
     return module_read_entry_print(&read_entry, entry->id, buf, size);
 }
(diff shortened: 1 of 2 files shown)


derekbruening:
snehasish is out today so asking sapostolakis

ext/drcovlib/modules.c
```diff
@@ -459,10 +459,9 @@ module_table_entry_print(module_entry_t *entry, char *buf, size_t size)
 #endif
     read_entry.path = full_path;
     read_entry.custom = entry->custom;
-#ifndef WINDOWS
-    // For unices we record the physical offset from the backing file.
+    // For unices we record the physical offset from the backing file
+    // (always 0 on Windows).
```
sapostolakis, line 463: nit: maybe use /* */  instead for the comment.

sapostolakis (approved):

ext/drcovlib/modules.c
derekbruening: That's from snehasish's original code but it is a C file so sure I could tweak it to better match the style.

PR #4779, opened 2021-03-15
Be more explicit in Windows build instructions

moshekaplan:
Be explicit about 32-bit vs 64-bit instructions and use of the correct VS command prompt.


derekbruening:
The test failures: Github has something messed up in their Ubuntu VM's: #4780

derekbruening (approved):
Thank you for the patch!

Issue #4780, opened 2021-03-15
GA CI failures due to stale apt data


derekbruening:
We first saw this on aarchxx with the qemu package and I put in an "apt update" in PR #4763.
I did not put that in every job since it takes a solid 30s and would thus be a 5%+ time increase for every job.
But we're now seeing failures in other places:

Package build failed which is not good: 
https://github.com/DynamoRIO/dynamorio/runs/2100301478?check_suite_focus=true
```
E: Failed to fetch http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu/pool/main/g/gcc-9/libx32gcc1_9.3.0-10ubuntu2~16.04_amd64.deb  404  Not Found [IP: 91.189.95.85 80]
E: Failed to fetch http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu/pool/main/g/gcc-9/lib32gomp1_9.3.0-10ubuntu2~16.04_amd64.deb  404  Not Found [IP: 91.189.95.85 80]
E: Failed to fetch http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu/pool/main/g/gcc-9/libx32gomp1_9.3.0-10ubuntu2~16.04_amd64.deb  404  Not Found [IP: 91.189.95.85 80]
E: Failed to fetch http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu/pool/main/g/gcc-9/lib32itm1_9.3.0-10ubuntu2~16.04_amd64.deb  404  Not Found [IP: 91.189.95.85 80]
E: Failed to fetch http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu/pool/main/g/gcc-9/libx32itm1_9.3.0-10ubuntu2~16.04_amd64.deb  404  Not Found [IP: 91.189.95.85 80]
E: Failed to fetch http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu/pool/main/g/gcc-9/lib32atomic1_9.3.0-10ubuntu2~16.04_amd64.deb  404  Not Found [IP: 91.189.95.85 80]
E: Failed to fetch http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu/pool/main/g/gcc-9/libx32atomic1_9.3.0-10ubuntu2~16.04_amd64.deb  404  Not Found [IP: 91.189.95.85 80]
[...]

PR #4785, opened 2021-03-17
i#4780 apt: Remove multilib workaround

derekbruening:
The underlying issue behind the multilib apt problems has been
resolved so we no longer need the workaround.

Issue: #4780


abhinav92003 (approved):

derekbruening:
Windows failure is a doxygen download failure.  32-bit passed though so something very transient??

Issue #4670, opened 2021-01-12
ASSERT on unlinking for signal-interrupted inlined syscalls


derekbruening:
We hit this assert on a proprietary app in ~3 out of 1000 runs:
```
Internal Error: DynamoRIO debug check failure: third_party/dynamorio/trunk/core/unix/signal.c:4234 info->interrupted == NULL || info->interrupted == f
```
The app is pretty simple; will see if we can reproduce in a github test.

derekbruening:
Briefly explaining what we figured out about the underlying cause: the special ibl transfer mechanism is not unlinkable.  drwrap uses it for its replace-retaddr scheme.

PR #4797, opened 2021-03-21
i#4670: Unlink signal-interrupted special IBL xfer frag

abhinav92003:
Adds check for pending signals in the special ibl xfer trampoline. If there are pending signals, we transfer control to the unlinked ibl entry point instead.

This is required to ensure that we exit the fcache and bound delivery time for signals received while executing fragments that use the special ibl xfer mechanism.

Verified that this fix works on the proprietary x86 app mentioned in #4670. No failure observed in 1000 dbg runs.

Fixes: #4670


core/arch/emit_utils_shared.c
```diff
@@ -5534,6 +5536,56 @@ emit_special_ibl_xfer(dcontext_t *dcontext, byte *pc, generated_code_t *code, ui
...
+                         OPND_CREATE_INT8(0)));
+    insert_shared_restore_dcontext_reg(dcontext, &ilist, NULL);
+    APP(&ilist,
+        XINST_CREATE_jump_cond(dcontext, DR_PRED_GT, opnd_create_pc(ibl_unlinked_tgt)));
+#    elif defined(AARCHXX)
+    instr_t *skip_unlinked_tgt_jump = INSTR_CREATE_label(dcontext);
+    /* Reusing SCRATCH_REG5 which contains dcontext currently. */
+    APP(&ilist,
+        INSTR_CREATE_ldrsb(dcontext, opnd_create_reg(SCRATCH_REG5),
+                           OPND_DC_FIELD(false, dcontext, OPSZ_1, SIGPENDING_OFFSET)));
```
abhinav92003, line 5559: Another option is that we add a new member to `dcontext_t` for `is_thread_unlinking` which can be incremented in all scenarios requiring proactive unlinking-relinking of fragments. But I'm aware of only the pending signals scenario right now for which `signals_pending>0` works just as well.

core/arch/emit_utils_shared.c
```diff
@@ -5534,6 +5536,56 @@ emit_special_ibl_xfer(dcontext_t *dcontext, byte *pc, generated_code_t *code, ui
...
+#ifdef UNIX
+    /* i#4670: Jump to the unlinked IBL target if there are pending signals. This is
+     * required to bound delivery time for signals received while executing fragments
+     * that use the special ibl xfer trampoline, which uses a different (un)linking
+     * mechanism.
+     */
+    insert_shared_get_dcontext(dcontext, &ilist, NULL, true);
+#    ifdef X86
+    APP(&ilist,
+        XINST_CREATE_cmp(dcontext,
```
derekbruening, line 5548: This clobbers the flags but I don't see them saved?  I assume you're not assuming -unsafe_ignore_eflags_ibl (though this is not always an app ib).  Use load+jecxz?  Requires swapping regs.

core/arch/emit_utils_shared.c
```diff
@@ -5534,6 +5536,56 @@ emit_special_ibl_xfer(dcontext_t *dcontext, byte *pc, generated_code_t *code, ui
...
+    APP(&ilist,
+        XINST_CREATE_jump_cond(dcontext, DR_PRED_GT, opnd_create_pc(ibl_unlinked_tgt)));
+#    elif defined(AARCHXX)
+    instr_t *skip_unlinked_tgt_jump = INSTR_CREATE_label(dcontext);
+    /* Reusing SCRATCH_REG5 which contains dcontext currently. */
+    APP(&ilist,
+        INSTR_CREATE_ldrsb(dcontext, opnd_create_reg(SCRATCH_REG5),
+                           OPND_DC_FIELD(false, dcontext, OPSZ_1, SIGPENDING_OFFSET)));
+    APP(&ilist,
+        XINST_CREATE_cmp(dcontext, opnd_create_reg(SCRATCH_REG5), OPND_CREATE_INT8(0)));
```
derekbruening, line 5561: This again clobbers the flags.  Use CBZ instead?

core/arch/emit_utils_shared.c
abhinav92003: Do you mean that we should load `dcontext` into `xcx` (instead of `xdi`) to avoid load/store for a second scratch reg? We can then reuse `ecx` to load pending signals count, and then restore only `xcx`.

abhinav92003: Or is it better to just reuse `insert_shared_get_dcontext` as-is (it has some extra logic for `-protect_mask`) to load `dcontext` into `xdi` and use `xcx` as scratch?

derekbruening: Or load into xdi and xchg with xcx, and xchg back on both paths I guess.  xchg reg-to-reg is fast I believe; only w/ memory is it to be avoided.

core/arch/emit_utils_shared.c
abhinav92003: Done.

core/arch/emit_utils_shared.c
abhinav92003: Done.

core/arch/emit_utils_shared.c
```diff
@@ -5534,6 +5536,56 @@ emit_special_ibl_xfer(dcontext_t *dcontext, byte *pc, generated_code_t *code, ui
         in = instrlist_first(custom_ilist);
     }
 
+#ifdef UNIX
+    /* i#4670: Jump to the unlinked IBL target if there are pending signals. This is
+     * required to bound delivery time for signals received while executing fragments
```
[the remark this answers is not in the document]
abhinav92003: I did 10 runs with and without  the fix of the proprietary app mentioned in i#4670. 



without-fix user seconds | with-fix user seconds
-- | --
109 | 264
230 | 297
246 | 297
308 | 301
316 | 322
329 | 323
334 | 351
338 | 358
354 | 369
383 | 439


There seem to be some variance between runs. So, removing first and last 10% as outliers, we get:
Average old vs new: 307 vs 327 seconds. So there's a 6.5% slowdown.

core/unix/signal.c
```diff
@@ -3900,6 +3900,8 @@ unlink_fragment_for_signal(dcontext_t *dcontext, fragment_t *f,
         if (TEST(FRAG_LINKED_OUTGOING, f->flags)) {
             unlink_fragment_outgoing(dcontext, f);
             changed = true;
+            // i#4670: Fragments that use the special ibl xfer trampoline are unlinked
```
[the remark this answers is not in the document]

core/arch/emit_utils_shared.c
```diff
@@ -5494,10 +5494,12 @@ emit_special_ibl_xfer(dcontext_t *dcontext, byte *pc, generated_code_t *code, ui
     reg_id_t stub_reg = IF_AARCH64_ELSE(SCRATCH_REG0, SCRATCH_REG1);
```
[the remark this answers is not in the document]
abhinav92003: Done. I guess I can remove the before colon part, as this is the only PR required to fix the issue.

core/arch/emit_utils_shared.c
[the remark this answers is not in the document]
abhinav92003: Probably needs more runs for better confidence in the numbers.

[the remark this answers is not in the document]
derekbruening: > I did 10 runs with and without the fix of the proprietary app mentioned in i#4670.

That app is doing heavyweight instrumentation which may mask other overhead?  Best to disable that and have just the drwrap portion.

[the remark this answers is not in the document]
abhinav92003: [drwrap-test-detach](https://github.com/DynamoRIO/dynamorio/blob/d28cd2d711059c00b5de6ec86f7618ce371b2bb1/suite/tests/client-interface/drwrap-test-detach.cpp#L129) seems suitable -- it loops on a function wrapped using `DRWRAP_RETURN_RETADDR`, and has only the wrap overhead, not other instrumentation.

- I noticed that the loop goes on for only [1ms](https://github.com/DynamoRIO/dynamorio/blob/d28cd2d711059c00b5de6ec86f7618ce371b2bb1/suite/tests/client-interface/drwrap-test-detach.cpp#L156), which leads to `pre_count`=`post_count`=0. So it'll be a good idea to increase this to 1s in the test suite. I have added a commit for that to this PR.
- For testing the overhead of this PR, I increased the sleep duration to 5 mins and compared `pre_count` from the before and after versions over 5 runs. Higher is faster.



without-fix pre_count | with-fix pre_count
-- | --
304008602 | 304555102
309676422 | 308248502
309824682 | 309160062
310135622 | 310072462
311061162 | 311737202

Harmonic mean of `pre_count` without fix: 308920630, with fix: 308735986
So there's a 0.06% slowdown after this PR.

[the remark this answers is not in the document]
abhinav92003: Also, I think I should've used the above methodology with the i#4670 proprietary app experiment too (in my earlier comment), because that app also sleeps for some time to allow threads to loop over some code. So overall execution time shouldn't have changed much before and after the change. Probably some other overhead dominated the execution time.

core/arch/emit_utils_shared.c
derekbruening: I think long-term #4804 is the cleanest solution and we shouldn't anticipate having any other fastpath checks for unlinking: we should use regular link/unlink paths for any other cases.

core/arch/emit_utils_shared.c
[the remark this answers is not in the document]
derekbruening: OK, I think it would also be good to compare the tracing metrics on the large proprietary apps that were used when developing the replace-retaddr feature in the first place: we can talk offline about how to do that.

derekbruening (approved):

Issue #4769, opened 2021-03-07
Missing unregistering of drmgr's tls slot in drcachesim's func_trace


sapostolakis:
The `func_trace` in drcachesim registers a thread-local storage (tls) slot during initialization (`func_trace_init()`) but does not unregister this slot during exit (`func_trace_exit()`).  This is problematic in a setting with multiples detaches and re-attaches and no resetting of the taken tls slots, since there is a slot that is not properly freed ("leaked" slot) potentially leading to a false shortage of available tls slots.

derekbruening:
I would argue that the title is not really a bug.  The docs do not require unregistering anything at process exit; like global heap, the general cleanup tears everything down.  The bug is that drmgr_exit() does not reset all of its state for a re-attach (or possibly drmgr_init() does not clear it) with a statically-linked DR, which is a relatively new usage mode that causes problems for global state throughout DR and its libraries.  That state includes more than just whether slots are taken: the drmgr TLS value array is guaranteed to start out at NULL yet it is not reset either, so just having func_trace unregister does not solve the problem, nor solve it with say some 3rd-party client who does not unregister or who relies on the NULL value.

To reiterate, there will never be a problem here with regular DR usage, no matter how many re-attaches occur.  The only issue is with DR statically linked into the app.  The master issue for that is #2157.

I would suggest several action items:

1) Add a regression test that re-attaches with static-DR 100x or more.  The existing static tests only re-attach 3x or so.  Confirm it catches this problem.  The challenge is making it fast enough for the CI suite while still including as many components as possible to test their re-attached state.

2) Add a DR API routine to query whether a detach is in progress, to follow the core's convention of avoiding excess memset on non-detach exit.  The idea is to keep teardown for small processes as fast as possible so DR doesn't have horrible performance when running, say, a series of shell scripts and utilities they launch, all short-lived processes.

2) Audit all the ext libs and add memsets if detaching where necessary.  Unfortunately there's no simple global solution, unless we try to have our own data section and zero it in bulk or something, which would almost certainly still have exceptions and problems.  Xref #2157, #2661, #4319.  Just another messy complexity of static DR, along with lack of resource isolation.

Even if we decided to enforce unregistering of TLS slots and/or other things at exit (which I think requires a documented compatibility break), we would still have all of the other problems and the above would still be needed; in that case we could split into a separate issue perhaps.

sapostolakis:
[...]

PR #4799, opened 2021-03-22
i#4769: Reset global state of ext libs for re-attaches with static-DR.

sapostolakis:
Resets non-previously-reset parts of the global state of several ext libs when detaching is in progress.
Also adds a DR API routine to query whether a detach is in progress.
For testing, adds a regression test that re-attaches with static-DR 100x and triggers #4769.
This test is based on burst_malloc and it takes 16sec on a debug build (much faster than burst_malloc with 100 reattaches mostly thanks to use of -record_dynsym_only option and simpler do_some_work() functionality).

Fixes: #4769


derekbruening:
> This test is based on burst_malloc and it takes 270sec

Is that on your own physical machine?  So probably 2x == nearly 10 mins on a VM?  I guess we should have the actual number from the runs here.  How much did the overall job times increase?  The goal is all jobs <= 15 mins.

sapostolakis:
> > This test is based on burst_malloc and it takes 270sec
> 
> Is that on your own physical machine? So probably 2x == nearly 10 mins on a VM? I guess we should have the actual number from the runs here. How much did the overall job times increase? The goal is all jobs <= 15 mins.

Yes, it was on my own machine (which is a VM though). In CI it took 314.97 sec for x86-64, and 469.63 sec for x86-32. It definitely pushes the time for jobs beyond 15 mins but < 20mins.

ext/drreg/drreg.c
```diff
@@ -2009,8 +2009,13 @@ drreg_exit(void)
...
     }
 
-    /* Support re-attach */
-    memset(&ops, 0, sizeof(ops));
+    /* Support re-attach. */
+    if (dr_doing_detach()) {
+        memset(&ops, 0, sizeof(ops));
+#ifdef DEBUG
+        stats_max_slot = 0;
+#endif
```
derekbruening, line 2017: (OK it does seem like the other ones are set at init time:
    122:static drreg_options_t ops;
    124:static int tls_idx = -1;
    125:static uint tls_slot_offs;
    126:static reg_id_t tls_seg;
    129:static uint stats_max_slot;
   1843:static int drreg_init_count;
   1845:static per_thread_t init_pt;
)

clients/drcachesim/CMakeLists.txt
```diff
@@ -577,6 +577,12 @@ if (BUILD_TESTS)
         COMPILE_FLAGS "-Wno-attribute-alias")
     endif ()
 
+    add_executable(tool.drcacheoff.burst_reattach tests/burst_reattach.cpp)
+    configure_DynamoRIO_static(tool.drcacheoff.burst_reattach)
+    use_DynamoRIO_static_client(tool.drcacheoff.burst_reattach drmemtrace_static)
+    add_win32_flags(tool.drcacheoff.burst_reattach)
+    link_with_pthread(tool.drcacheoff.burst_reattach)
```
derekbruening, line 584: I don't see any use of threads so this looks like a mistake?

core/lib/instrument_api.h
```diff
@@ -6903,4 +6903,11 @@ dr_unregister_persist_patch(bool (*func_patch)(void *drcontext, void *perscxt,
                                                byte *bb_start, size_t bb_size,
                                                void *user_data));
 
+DR_API
+/**
+ * Query whether detach is in progress.
+ */
+bool
+dr_doing_detach(void);
```
derekbruening, line 6911: Maybe the name should have "is" to reflect the bool query and to match dr_is_notify_on, dr_is_thread_native, 
dr_is_tracking_where_am_i.  `dr_is_detaching`?

ext/drwrap/drwrap.c
```diff
@@ -1008,6 +1008,12 @@ drwrap_exit(void)
         postcall_cache[i] = NULL;
     }
     postcall_cache_idx = 0;
+#ifdef WINDOWS
+    sysnum_NtContinue = -1;
+#endif
+    disabled_count = 0;
+    if (dr_doing_detach())
+        memset(&drwrap_stats, 0, sizeof(drwrap_stats_t));
```
derekbruening, line 1016: It seems good to put the postcall_cache array and probably the other stores above inside the same if() right?

clients/drcachesim/tests/burst_reattach.cpp
```diff
@@ -0,0 +1,158 @@
...
+    if (!my_setenv("DYNAMORIO_OPTIONS",
+                   "-stderr_mask 0xc"
+                   " -client_lib '#;;-offline -record_heap"
+                   // Test the low-overhead-wrapping option.
+                   " -record_replace_retaddr"
+                   // Test large values that require two entries.
+                   " -record_function \"malloc|1&return_big_value|1\"'"))
+        std::cerr << "failed to set env var!\n";
+
+    for (int i = 0; i < 100; i++) {
```
derekbruening, line 115: How much of the large runtime is from the do_some_work() size and we can thus easily shrink?

What if we set -max_global_trace_refs or other limits on data written?  Or is that not the bottleneck?

What if we ran -checklevel 1 or 0 -- not sure we want to do that though.

clients/drcachesim/tests/burst_reattach.cpp
```diff
@@ -0,0 +1,158 @@
...
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+/* This application links in drmemtrace_static and acquires a trace during
+ * a "burst" of execution and memory allocations (malloc() and free())
+ * in the middle of the application. It then detaches. Later it re-attaches
+ * and detaches again for several times.
```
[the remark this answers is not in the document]
sapostolakis: Updated it to reflect the purpose of the test.

clients/drcachesim/tests/burst_reattach.cpp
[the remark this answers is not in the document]
sapostolakis: The do_some_work() is shrunk in this one and there are some tweaks in the options (i.e., removal of rstats_to_stderr).  In the latest commit I also removed more options and did more tweaks (removed record_heap and use and recording of return_big_value()). All these help reduce the overhead a bit (27%). If we are ok with the extra overhead we could just increase the reattach count in burst_malloc, or tweak the burst_malloc to reduce its overhead (e.g., simplify its do_some_work() ) . I also thought that this test might evolve and get disabled in some settings, so it might be good to keep it separate.

core/lib/instrument.c
```diff
@@ -8265,4 +8265,14 @@ dr_unregister_persist_patch(bool (*func_patch)(void *drcontext, void *perscxt,
...
 }
 
+DR_API
+bool
+dr_doing_detach(void)
+{
+    if (doing_detach) {
+        return true;
+    }
+    return false;
```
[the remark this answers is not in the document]
sapostolakis: Yep. Much simpler.

core/lib/instrument_api.h
```diff
@@ -6903,4 +6903,11 @@ dr_unregister_persist_patch(bool (*func_patch)(void *drcontext, void *perscxt,
                                                byte *bb_start, size_t bb_size,
                                                void *user_data));
 
+DR_API
+/**
+ * Query whether detach is in progress.
```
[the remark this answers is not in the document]
sapostolakis: Good point. Added some motivation for this API.

core/lib/instrument_api.h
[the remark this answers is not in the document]
sapostolakis: Right. Done.

ext/drcovlib/drcovlib.c
```diff
@@ -499,6 +499,7 @@ drcovlib_exit(void)
     if (!drcov_per_thread) {
         dump_drcov_data(NULL, global_data);
         global_data_destroy(global_data);
+        drcov_per_thread = false;
```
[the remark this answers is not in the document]
sapostolakis: True. Need to reset it to false outside the if() statement.

ext/drmgr/drmgr.c
```diff
@@ -569,6 +569,15 @@ drmgr_exit(void)
     dr_rwlock_destroy(bb_cb_lock);
 
     dr_mutex_destroy(note_lock);
+
+    /* Reset tls and cls arrays to handle reattaches with statically linked
+     * DR (no natural reset of global state), but only in case of detaching
+     * to avoid the memset overhead when there is no chance of reattaching.
+     */
+    if (dr_doing_detach()) {
+        memset(tls_taken, 0, sizeof(tls_taken));
```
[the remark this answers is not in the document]
sapostolakis: Sure.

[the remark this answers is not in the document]
sapostolakis: Moved them.

ext/drreg/drreg.c
sapostolakis: Yep.

core/lib/instrument_api.h
sapostolakis: Done.

clients/drcachesim/CMakeLists.txt
sapostolakis: Good catch. I initially was trying to add more things in this test but I forgot to remove this eventually unused part.

ext/drwrap/drwrap.c
sapostolakis: Sure. I just had inside the if() only stores with a little bit more cost, but I can move the rest there as well.

clients/drcachesim/tests/burst_reattach.cpp
sapostolakis: Compared to the burst_malloc test, I minimized as much as I could the runtime of `do_some_work()` and stats reported (removed` -rstats_to_stderr`) and that resulted in the ~20% reduction that I initially reported. Beyond that I don't see more gains by reducing do_some_work. The bottleneck is the use of `record_function` or `record_heap`. Use of any those options increases the runtime by orders of magnitude. By removing these options and commenting out the assertions for finding malloc and return_big_value, the test only takes 15secs. So, setting up DR and attaching and detaching, and the work done have minimal overhead. Yet, without recording any functions the #4769 issue is not triggered.  Just keeping the recording for `malloc` (and still trigger #4769) but removing `-record_heap` and the recording of `return_big_value` results in 8% reduction in runtime which seems to be worth doing (included this change in the latest commit). Use of `-checklevel 1` additionally reduces runtime by 5% and setting it to 0 yields 12% reduction over not setting checklevel. So, the reduction with checklevel is probably not big enough to justify its use. Use of -max_global_trace_refs with various values (even very small ones) did not have much (if any) impact in performance.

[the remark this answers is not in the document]
derekbruening: If you removed all the function recording, it no longer reproduces the bug?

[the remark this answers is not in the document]
sapostolakis: Yes

derekbruening: As discussed offline, -record_dynsym_only speeds up the test drastically: it seems it was symbol lookup combined with the option accumulation causing many many lookups that caused the slowness.

api/docs/release.dox
```diff
@@ -192,6 +192,7 @@ Further non-compatibility-affecting changes include:
  - Added support for running under QEMU via the \ref op_xarch_root "-xarch_root"
    runtime option which locates guest system libraries and enables workarounds for
    problems with QEMU's threads.
+ - Added dr_is_detaching(), a DR_API to query whether detach is in progress.
```
derekbruening, line 195: nit:  a little unusual to have the header token "DR_API" here?

derekbruening (approved):

api/docs/release.dox
sapostolakis: done

Issue #4787, opened 2021-03-18
File Access Failure in bin32/drrun on NFS


0xIzaak:
**Description and Workaround**
When the DynamoRIO root (unpacked release) is located on an NFS share, the bin32 version of `drrun` will warn about `libdynamorio.so` missing, then error out trying to access the DynamoRIO root directory. The bin64 version works as expected.

Specifically, `dynamorio/drdeploy.c:545` raises the error which originates from a `stat` failure in `dynamorio/dr_frontend_unix.c:81`.

I believe the root cause is that my NFS mount is using 64b inodes but the 32b version of `stat` expects a 32b value and errors. My solution to the problem was to add `add_definitions(-D_FILE_OFFSET_BITS=64)` in the `CMakeLists.txt` and rebuild.

The following link https://www.mjr19.org.uk/sw/inodes64.html was helpful for me understanding the suspected problem.

**To Reproduce**
Running `./bin32/drrun -- myapplication args` in an NFS will terminate with a `cannot access DynamoRIO root directory` error message. This occurs without a client and on all 32b ELF's tried.

**Versions**
This occurs on the latest 8.0.18687 and was also tested on an old 8.0.18439 release. I am running on Ubuntu 18.04.5. My application is 32 bit.

derekbruening:
Thank you for the detailed filing which includes a suggested fix.  I wonder if you could post a pull request with the fix?

PR #4805, opened 2021-03-24
i#4787 Add 64b file offsets to build definitions

0xIzaak:
Resolves issue with the 32b drrun exiting prematurely due to problem `stat`ing large inode numbers.

Fixes: #4787

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,6 +99,10 @@ endif (NOT DEFINED ENV{CXXFLAGS})
 # Don't use the default-Debug build type set for the try-compiles
 set(specified_build_type "${CMAKE_BUILD_TYPE}")
 project(DynamoRIO NONE)
+
+# Enable large file support
+add_definitions(-D_FILE_OFFSET_BITS=64)
+
 if (DEFINED GENERATE_PDBS AND NOT GENERATE_PDBS)
   # i#310: support building over cygwin ssh where we cannot build pdbs.
   # To prevent cmake's try-compile for its working compiler test and


0xIzaak:
This seems to have caused an error in `linux.signalfd` on an invalid file descriptor but it's not obvious how the `_FILE_OFFSET_BITS` macro could affect that. Especially since the passed fd arg is `-1` in the test, this is quite strange to me. Any ideas on how this is getting triggered?

derekbruening:
run arm tests

derekbruening:
(The "run arm tests" triggers the AArch64 run)

derekbruening:
> This seems to have caused an error in linux.signalfd on an invalid file descriptor but it's not obvious how the _FILE_OFFSET_BITS macro could affect that. Since the passed fd arg is -1 in the test, this makes me wonder if signalfd is somehow failing to create the new descriptor. Any ideas on how this is getting triggered?

No, I agree, it seems like it would not affect signalfd at all.  Does it reproduce locally with the signalfd app running native (i.e., not under DR)?

0xIzaak:
Could you guide me on how to build the signalfd app? I've tried running the usual cmake commands in `/dynamorio/suite/tests/` with no luck (and honestly I probably want to test with whatever config the CI system is using).

derekbruening:
> Could you guide me on how to build the signalfd app locally? I've tried running the usual cmake commands in `/dynamorio/suite/tests/` with no luck (and honestly I probably want to test with whatever config the CI system is using).

See https://dynamorio.org/page_test_suite.html#autotoc_md262 -- you probably want `-DDEBUG=ON -DBUILD_TESTS=ON`.

0xIzaak:
Thanks for the link, I'd missed that in the docs. (Notwithstanding possible issues in my local test build) it looks like the error *does* show up in my local env having built with the `-m32` flags. The `/suite/tests/bin/linux.signalfd` runs fine alone but with `bin32/drrun` it yields the same `signalfd failed: Bad file descriptor`.

From looking over the strace I see: the expected mapping to `stat64` in the startup checks, we move past the dr init, a fair bit of thread and signal activity (presumably standard stuff?), and then finally some pipe and file activity. As best as I can tell, it looks like that signal comes from a `dup(-22)` following a failure in `fcntl()` duplicating one of the ends of a pipe fd.

Here's the strace output in case someone more knowledgeable might see something:
[strace_bad_fd.txt](https://github.com/DynamoRIO/dynamorio/files/6205684/strace_bad_fd.txt)

derekbruening:
> As best as I can tell, it looks like that signal comes from a dup(-22) following a failure in fcntl() duplicating one of the ends of a pipe fd.

Does `-steal_fds 0` make the problem go away?

0xIzaak:
`./bin32/drrun -steal_fds 0 -- ./suite/tests/bin/linux.signalfd` still results in `signalfd failed: Bad file descriptor`

derekbruening:
Hmm, signalfd still fails...do not have time to investigate further unfortunately.

Issue #4723, opened 2021-02-06
ASSERT decode-stress translate.c:1094


derekbruening:
This showed up on Windows 32-bit:

https://github.com/DynamoRIO/dynamorio/pull/4722/checks?check_run_id=1846725471
```
2021-02-06T23:18:34.8554457Z 32: Test command: /home/runner/work/dynamorio/dynamorio/build_debug-internal-32/bin32/drrun "-s" "90" "-quiet" "-debug" "-killpg" "-stderr_mask" "0xC" "-dumpcore_mask" "0" "-code_api" "--" "/home/runner/work/dynamorio/dynamorio/build_debug-internal-32/suite/tests/bin/linux.execve-rec"
2021-02-06T23:18:34.8555892Z 32: Test timeout computed to be: 600
2021-02-06T23:18:34.8556319Z 13: Start
2021-02-06T23:18:34.8556816Z 13: Jumping to a sequence of every addr16 modrm byte
2021-02-06T23:18:34.8557360Z 32: under DynamoRIO
2021-02-06T23:18:34.8557728Z 32: Sum(10)
2021-02-06T23:18:34.8558229Z 32: under DynamoRIO
2021-02-06T23:18:34.8558623Z 32: under DynamoRIO
2021-02-06T23:18:34.8559151Z 32: under DynamoRIO
2021-02-06T23:18:34.8559559Z 32: under DynamoRIO
2021-02-06T23:18:34.8561218Z 13: <Application /home/runner/work/dynamorio/dynamorio/build_debug-internal-32/suite/tests/bin/common.decode (18761).  Internal Error: DynamoRIO debug check failure: /home/runner/work/dynamorio/dynamorio/core/translate.c:1094 false
2021-02-06T23:18:34.8562479Z 13: (Error occurred @2432 frags)
2021-02-06T23:18:34.8562940Z 13: version 8.0.18664, custom build
2021-02-06T23:18:34.8564763Z 13: -no_dynamic_options -code_api -stderr_mask 12 -stack_size 56K -signal_stack_size 32K -max_elide_jmp 0 -max_elide_call 0 -early_inject -emulate_brk -no_inline_ignored_syscalls -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct -stress_recreate_pc -stress_recreate_state 
2021-02-06T23:18:34.8566131Z 13: 0x3f7fe298 0xf7c7080f
[...]

Issue #2286, opened 2017-03-11
win32.earlythread failing on Appveyor


derekbruening:
https://ci.appveyor.com/project/DynamoRIO/dynamorio/build/1.0.299/job/p137vc45ie5ybj88

64-bit win32.earlythread started failing on occasion:

```
[00:10:14]  44/193 Test  #43: code_api|win32.earlythread ...................................***Failed  Required regular expression not found.Regex=[^in lib
[00:10:14] ?
[00:10:14] lib returned 4
[00:10:14] ?
[00:10:14] $
[00:10:14] ]  0.13 sec
[00:10:14] 
```

derekbruening:
I can't reproduce this in 1000 iters on win10, or 1000 iters on server12.

toshipiazza:
Seen again here: https://ci.appveyor.com/project/DynamoRIO/dynamorio/build/1.0.827

abhinav92003:
Happened again: https://github.com/DynamoRIO/dynamorio/runs/2569239839?check_suite_focus=true

PR #4806, opened 2021-03-25
i#4736: Fix high-loglevel issues and add test

### LINKED ISSUE #4736: Add high loglevel regression testing by running tiny pure-asm apps
It is not uncommon for new users to run at -loglevel 5 or higher: but regular developers rarely go that high and we have no tests that high, meaning that such verbosity often bitrots and hits asserts or whatnot.  This is a proposal to add some high-loglevel regression tests and keeping them reasonably short and with reasonable disk space by running tiny pure-asm apps (we already have some of those in the suite).

derekbruening:
Fixes several problems with loglevel 5 and above:
+ Fixes a deadlock by avoiding an over-long LOG in a fragile memquery context.
+ Avoids an assert by avoiding an opcode query for a level 0 bundle.
+ Fixes a rank order violation by avoiding diagnostic module name
  lookup if holding fcache_unit_areas->lock.
+ Fixes an infinite recursion by avoiding the same module lookup for the
  loaded_module_areas vector.
+ Fixes a crash when thread_kstats is not yet allocated.
+ Fixes a memory leak at level 5.
+ Effectively disables vmm_dump_map as with 4K blocks it has enormous output now.

Adds a test of loglevel 19 that runs a pure-asm app to keep the time
and space down.  Marks it as part of the QEMU suite.

Fixes #4736

### DIFF
--- a/core/heap.c
+++ b/core/heap.c
@@ -719,8 +719,10 @@ vmm_dump_map(vm_heap_t *vmh)
     bool is_used = bitmap_test(b, 0) == 0;
 
     LOG(GLOBAL, LOG_HEAP, 3, "vmm_dump_map(" PFX ")\n", vmh);
-    /* raw dump first - if you really want binary dump use windbg's dyd */
-    DOLOG(4, LOG_HEAP, {
+    /* We used to do raw dumps but with the shift to 4K blocks, this is just way
+     * too big.  We disable but leave the capability to enable one-off use.
+     */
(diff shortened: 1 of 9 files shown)


derekbruening:
Windows failure is 64-bit win32.earlythread: #2286

abhinav92003 (approved):

derekbruening:
Failure is 32-bit decode-stress translate.c assert: #4723

Issue #4732, opened 2021-02-11
client.winxfer test output does not always match


derekbruening:
This was under #4058 and I put some time into relaxing the output to match, but it is still failing sometimes:

https://github.com/DynamoRIO/dynamorio/pull/4729/checks?check_run_id=1880901026

```
2021-02-11T16:40:01.9664547Z 128: Test command: D:\a\dynamorio\dynamorio\build_debug-internal-32\bin32\drrun.exe "-s" "90" "-quiet" "-debug" "-use_dll" "D:/a/dynamorio/dynamorio/build_debug-internal-32/lib32/debug/dynamorio.dll" "-exit0" "-stderr_mask" "0xC" "-msgbox_mask" "0" "-dumpcore_mask" "0x7d" "-staged" "-code_api" "-c" "D:/a/dynamorio/dynamorio/build_debug-internal-32/suite/tests/bin/client.flush.dll.dll" "--" "D:/a/dynamorio/dynamorio/build_debug-internal-32/suite/tests/bin/client.flush.exe"
2021-02-11T16:40:01.9666562Z 128: Test timeout computed to be: 600
2021-02-11T16:40:01.9666996Z 126: kernel_xfer_event: type 5
2021-02-11T16:40:01.9667410Z 126: in wnd_callback 0x00008001 0 2
2021-02-11T16:40:01.9667826Z 126: kernel_xfer_event: type 6
2021-02-11T16:40:01.9668210Z 126: Got message 0x00008001 1 3
2021-02-11T16:40:01.9668616Z 126: Before _beginthreadex
2021-02-11T16:40:01.9953198Z 126: kernel_xfer_event: type 2
2021-02-11T16:40:02.3778200Z 126: kernel_xfer_event: type 7
2021-02-11T16:40:02.3778795Z 128: options = 
2021-02-11T16:40:02.5687498Z 126: QueueUserAPC returned 1
2021-02-11T16:40:02.5688052Z 126: kernel_xfer_event: type 2
2021-02-11T16:40:02.5688487Z 126: kernel_xfer_event: type 7
2021-02-11T16:40:02.5689063Z 126: SleepEx returned 192
2021-02-11T16:40:02.5689434Z 126: Apc arg = 37
2021-02-11T16:40:02.5689934Z 126: Result = 100
2021-02-11T16:40:02.5754554Z 126: After _beginthreadex
2021-02-11T16:40:02.6090320Z 126: All done
[...]

PR #4815, opened 2021-03-27
i#4111 web: Add direct section searches

derekbruening:
Adds search entries for section headers which target the anchor links
directly.

Issue: #4111


derekbruening:
Failure is flaky 32-bit client.winxfer: #4732

Issue #3092, opened 2018-07-07
[cleanup] refactor public interface headers and eliminate patchwork exports via genapi.pl


derekbruening:
Currently DR's public interface headers are constructed from a patchwork of existing internal headers using the genapi.pl script.  This issue covers refactoring the headers to properly separate the public interface portions into their own headers that can be directly exported, eliminating the need for a script and simplifying the project.

derekbruening:
Xref #3158

derekbruening:
The general approach here is that each exported header should be copied as-is (or possibly with CMake variable expansion, as we do for dr_api.h) from a checked-in header.  No more having a script extract pieces from the middle of headers.

This means that we need to extract the public, exported content from the headers and move it to new entirely-exported files.

For headers also used for internal code, the naming convention is xxx_api.h.  A header with that name must contain only exported content.  A CMake rule `DR_export_header` is used to copy it to the build dir's include/ directory and that copy can rename it: e.g., from `core/module_api.h` to `include/dr_modules.h`.

For headers only used for exported code, we use the same name as the exported header: so `core/lib/dr_events.h`.

derekbruening:
There are a lot of complications and details to deal with here.

E.g., genapi.pl does `s/dcontext_t *\*dcontext/void *drcontext/;`, but w/ the internal header having the full type, the impl files do as well.  So if we make the header void*, the impl file won't match; do we make it void* and add a cast and assume it's zero cost (could cost extra local var slot)?

What about DR_UNS_EXCEPT_TESTS_API for decode_cti?  I think we have to put it in decode_api.h and define that token and just have no doxygen header and no dllexport/visible symbol.

What about AVOID_API_EXPORT?  I've been trying to just reword to be fine to sit in exported header (not in doxygen comment).

What about API_EXPORT_ONLY?  E.g.,  dr_ir_utils.h including dr_ir_instr.h?  Might have to live w/ non-include-what-you-use there.

We'll have to drop not-HAVE_FVISIBILITY support so we can drop genapi.pl -filter and fully get rid of genapi.pl.

derekbruening:
My task list showing all the sources of exported content and what I ended up doing with each:

- [X] core/annotations.h:/* DR_API EXPORT TOFILE dr_annotation.h */
      => new file core/annotations_api.h => dr_annotation.h

- [X] core/lib/dr_config.h:/* DR_API EXPORT TOFILE dr_config.h */
      => cleaned up and exported as-is

[...]

PR #4823, opened 2021-03-30
i#3092 genapi: Split dr_modules.h from dr_tools.h

derekbruening:
Multiple headers fed into dr_tools.h, which is no longer possible with
the simple copy-as-is export model.  We split the module types from
module_shared.h into a new file dr_modules.h, and move module routines
and types from instrument_api.h into the file as well.  The
implementations are also moved from instrument.c to a new file
module_api.c.

Issue: #3092


api/docs/CMake_rundoxygen.cmake
```diff
@@ -148,7 +148,7 @@ execute_process(COMMAND
   ${DOXYGEN_EXECUTABLE}
   RESULT_VARIABLE doxygen_u_result
   ERROR_VARIABLE doxygen_u_error
-  OUTPUT_QUIET
+#  OUTPUT_QUIET
```
abhinav92003, line 151: Do we want to keep this commented out?

abhinav92003 (approved):

api/docs/CMake_rundoxygen.cmake
derekbruening: Nope, this was the mistake I mentioned offline.  Will remove it.

PR #4826, opened 2021-04-01
i#3092 genapi: Split up dr_ir_instr.h

derekbruening:
Renames core/ir/instr_inline.h to core/ir/instr_inline_api.h and makes
it directly copyable to the new exported header dr_ir_instr_inline.h.
To do this, introduces a new define DYNAMORIO_INTERNAL in order to set
macros differently for DR vs clients.

Separates the public parts of core/ir/instr.h (which is most of the
file) to a new file core/ir/instr_api.h, which is exported to
dr_ir_instr.h.  Uses DYNAMORIO_INTERNAL to set inlining and other
flags properly for core vs client use.

Moves the PREFIX_ constants from decode.h to instr_api.h.

Updates dcontext_t* types to void* for exposed functions.

Issue: #3092


derekbruening:
run arm tests

sapostolakis (approved):

derekbruening:
@AssadHashmi -- arm jenkins is not triggering?

AssadHashmi:
> @AssadHashmi -- arm jenkins is not triggering?

@derekbruening Jenkins crashed last night! Back now.

Issue #4830, opened 2021-04-01
GA CI Linux VM's failing: 32-bit toolchain broken; clang-format missing


derekbruening:
The clang-format and both Linux x86 jobs suddenly started failing about an hour ago.

Xref #4780 where 32-bit apt packages were messed up.

Clang-format says:
```
Reading Script: /home/runner/work/dynamorio/dynamorio/suite/runsuite.cmake,automated_ci;64_only;require_format
Detected a CI clang suite: disabling running of tests
CMake Error at /home/runner/work/dynamorio/dynamorio/suite/runsuite.cmake:254 (message):
  FATAL ERROR: clang-format is required but not found!
```

That may be related to this:
https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1604-README.md
```
[Ubuntu] Clang, GCC and Gfortran versions less than 9.x will be removed from images on March, 29
```
Though it still lists clang 9 as being installed in the base.

For the 32-bit the failures look like a messed-up toolchain finding the wrong libraries:

https://github.com/DynamoRIO/dynamorio/pull/4829/checks?check_run_id=2247626216
```
2021-04-01T16:07:58.4961762Z [  2%] Linking C executable bin/linux.infinite
2021-04-01T16:07:58.4962874Z /usr/lib/x86_64-linux-gnu/libdl.so: error adding symbols: File in wrong format
2021-04-01T16:07:58.4964490Z suite/tests/CMakeFiles/linux.infinite.dir/build.make:98: recipe for target 'suite/tests/bin/linux.infinite' failed
2021-04-01T16:07:58.4966284Z CMakeFiles/Makefile2:7501: recipe for target 'suite/tests/CMakeFiles/linux.infinite.dir/all' failed
2021-04-01T16:07:58.4967132Z collect2: error: ld returned 1 exit status

2021-04-01T16:07:58.4971965Z [  2%] Linking C executable bin/client.mbr_instrumentation_segment
2021-04-01T16:07:58.4973167Z /usr/lib/x86_64-linux-gnu/libdl.so: error adding symbols: File in wrong format
2021-04-01T16:07:58.4974711Z suite/tests/CMakeFiles/client.mbr_instrumentation_segment.dir/build.make:98: recipe for target 'suite/tests/bin/client.mbr_instrumentation_segment' failed

2021-04-01T16:09:28.5436729Z [ 58%] Linking CXX executable ../bin32/prefetch_analyzer_launcher
[...]

PR #4831, opened 2021-04-01
i#4830: Work around GA CI VM 32-bit and clang-format toolchain regressions

derekbruening:
WIP: Try Ubuntu 20 for clang w/o the clang-9 env var;
Try Ubuntu 18 for x86 to see if toolchain is better w/o test failures

Issue: #4830

### DIFF
--- a/.github/workflows/ci-clang-format.yml
+++ b/.github/workflows/ci-clang-format.yml
@@ -54,7 +54,7 @@ jobs:
   # the CI resources: 64-bit hits most clang-only warnings and 64-bit
   # is the primary target these days.
   clang-format:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-16.04
 
     steps:
     - uses: actions/checkout@v2
@@ -69,7 +69,8 @@ jobs:
 
     - name: Create Build Environment
       run: |
-        sudo apt-get -y install doxygen vera++ cmake zlib1g-dev libsnappy-dev
+        sudo apt-get -y install doxygen vera++ cmake zlib1g-dev libsnappy-dev \
+          clang-format
 
     - name: Run Suite
       working-directory: ${{ github.workspace }}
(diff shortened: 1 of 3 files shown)


derekbruening:
Exact same failures.  Prob fastest route is to install older gcc -- but it sure seems like we shouldn't have to do that: seems like their gcc-9 is messed up.

derekbruening:
WIth "g++-8 g++-8-multlib" it's broken in another way:

```
   /usr/bin/c++ -m32  CMakeFiles/cmTC_f1708.dir/testCXXCompiler.cxx.o -o cmTC_f1708 
    /usr/bin/ld: skipping incompatible /usr/lib/gcc/x86_64-linux-gnu/5/libstdc++.so when searching for -lstdc++
    /usr/bin/ld: skipping incompatible /usr/lib/gcc/x86_64-linux-gnu/5/libstdc++.a when searching for -lstdc++
    /usr/bin/ld: cannot find -lstdc++
    collect2: error: ld returned 1 exit status
```

Grrr, this shouldn't be this hard.  It shouldn't have broken in the first place: we're fine with gcc 9.

derekbruening:
So the suggested workaround from https://github.com/actions/virtual-environments/issues/2950#issuecomment-812105569 did work: a flaky test failed the 32-bit, and the wrong package name I think the clang-format.  So it seems the 32-bit-toolchain problem is from cmake 3.20.

abhinav92003:
Thanks for the fix! 
Is there any way to subscribe to https://github.com/actions/virtual-environments/labels/Announcement? That should let us know of these required migrations early.

abhinav92003 (approved):

derekbruening:
Looks like I still don't quite have the clang job right -- I think putting the env vars back will do it.

> Is there any way to subscribe to https://github.com/actions/virtual-environments/labels/Announcement? That should let us know of these required migrations early.

Not sure, but if I had seen the title of this one about gcc<9.0 I would have thought it would not affect us.  I wonder if the cmake 3.20 was a different announcement?  But I would have also assumed it wouldn't affect us.  Sill hard to believe cmake by itself would mess it up: wonder if it's a combination of cmake and sthg weird about these VM's.

Issue #4791, opened 2021-03-19
cmake CMP0077 warning


derekbruening:
This is showing up with cmake 3.18.4:
```
CMake Warning (dev) at suite/tests/CMakeLists.txt:55 (option):
  Policy CMP0077 is not set: option() honors normal variables.  Run "cmake
  --help-policy CMP0077" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  For compatibility with older versions of CMake, option is clearing the
  normal variable 'TEST_LONG'.
This warning is for project developers.  Use -Wno-dev to suppress it.
```

derekbruening:
Hoping you can fix this while changing code around this option.

Issue #4800, opened 2021-03-23
Revive the "long suite" and run it on master merges


derekbruening:
We used to run a separate "long suite" of additional tests not on every commit but periodically.
Today that suite is not run and is likely bit-rotted: xref #1807 about failures in the long suite.
Also xref #1793: clean up nightly suite and remove no longer supported configurations.

The suggestion here is to revive the long suite, remove no-longer-supported configurations and tests (overlaps with #1793), and run it on every master merge.  This would pull in important tests like `api.ibl-stress-aarch64-far-link_LONG` which today are not run at all.

We are already running a shorter suite on pull requests than master merges, from #4059 and in particular PR #4079, which uses this code:
```
if (WIN32 AND NOT X64 AND "$ENV{CI_TRIGGER}" STREQUAL "pull_request")
  set(SKIP_LESS_DIFFERENTIATED_TESTS_FOR_PULL_REQUEST ON)
else ()
  set(SKIP_LESS_DIFFERENTIATED_TESTS_FOR_PULL_REQUEST OFF)
endif ()
```

We can use the same CI_TRIGGER env var set by our GA workflows to enable TEST_LONG.

abhinav92003:
I ran the long suite in PR#4832 (https://github.com/DynamoRIO/dynamorio/pull/4832/checks?check_run_id=2255480365), and compared it to a recent master run (https://github.com/DynamoRIO/dynamorio/runs/2255669931) first to see how many additional tests the long suite runs, and whether any fail.

The following show no additional tests:
osx-x86-64
aarch64-cross-compile
aarch64-on-x86


**x86-32:**
regular suite:
```
2021-04-02T18:28:36.5483828Z debug-internal-32: 344 tests passed, **** 3 tests failed, of which 2 were flaky, but ignoring 0 for i#2941: ****
2021-04-02T18:28:36.5484580Z 	code_api|api.static_sideline_FLAKY 
2021-04-02T18:28:36.5485135Z 	code_api|api.static_maps_mixup_novars_FLAKY 
2021-04-02T18:28:36.5485989Z 	code_api|linux.clone-reset 
```
long suite:
```
2021-04-02T17:49:34.1185205Z debug-internal-32: 722 tests passed, **** 8 tests failed, of which 3 were flaky, but ignoring 0 for i#2941: ****
2021-04-02T17:49:34.1185925Z 	code_api|api.detach_spawn_stress_FLAKY 
[...]

Issue #4059, opened 2020-01-27
Speed up VS2017 Appveyor which is taking 45 mins


derekbruening:
VS2013 was taking ~33 mins.

Some of this may be some timeouts.  Some may be a slower compiler?

Build and test times:
```
$ egrep 'Build dire|Test project' Downloads/log\ \(22\).txt 
[00:02:01]    Build directory: C:/projects/dynamorio/build/build_debug-internal-32
[00:06:48] Test project C:/projects/dynamorio/build/build_debug-internal-32
[00:21:01]    Build directory: C:/projects/dynamorio/build/build_debug-internal-64
[00:25:34] Test project C:/projects/dynamorio/build/build_debug-internal-64
[00:35:33]    Build directory: C:/projects/dynamorio/build/build_release-external-32
[00:38:12] Test project C:/projects/dynamorio/build/build_release-external-32
[00:38:12]    Build directory: C:/projects/dynamorio/build/build_release-external-64
[00:44:47] Test project C:/projects/dynamorio/build/build_release-external-64
[00:44:47]    Build directory: C:/projects/dynamorio/build/build_release-external-64
[00:44:47] Command exited with code 1
```

We can probably shrink some of the longer tests: like drcachesim tests running that concurrency
app and taking a minute each.

derekbruening:
The longest-running test by far is nudge_test:
```
[00:19:41] 225/262 Test #128: code_api|client.nudge_test ...................................***Timeout 600.01 sec
```

10 minutes!

I took a look and found a lot of issues:

### #4074: ASSERT in ntdll_redir in infloop on nudge => relax for nudge thread; actually I undid this after the next assert was fixed to *not* swap state

I fixed this.

### #4074: ASSERT in infloop in tls cleanup post-nudge => swapping peb pointer too early

I fixed this.

### #4075: deadlock in infloop when nudge sent: thread_initexit_lock has no owner but lock_requests==4
I failed to figure this one out.  But at least with the changes below we now exit in a reasonable time.

### Add timeouts to all wait loops in runall.cmake

Now we have:
```
Sending nudge -nudge_pid;9816;0;10
CMake Error at D:/derek/dr/git/src/suite/tests/runall.cmake:232 (message):
  Timed out waiting for "done"
```

I also added a TIMEOUT to the execute_process() call.

### Add detection of assert to runall.cmake

```
CMake Error at D:/derek/dr/git/src/suite/tests/runall.cmake:237 (message):
[...]

Issue #1807, opened 2015-10-13
fix failing tests in debug build long suite


derekbruening:
Xref #1793.  There are many, many tests failing in the long suite.  This issue covers fixing them (or removing sets of runs if we decide not to support them anymore) and getting the long suite clean so we can re-enable it.

```
debug-internal-32: 2022 tests passed, **** 81 tests failed, of which 15 were flaky: ****
	code_api|client.thread 
	code_api,opt_memory|linux.child-blacklist 
	code_api,opt_memory|security-common.selfmod =>  Application /work/dr/nightly/run/build_debug-internal-32/suite/tests/bin/security-common.selfmod (26116).  Internal Error: DynamoRIO debug check failure: /work/dr/git/src/core/vmareas.c:2999 (is_coarse  coarse != NULL) || (!is_coarse  coarse == NULL) 
	code_api,opt_memory|client.drwrap-test 
	code_api|linux.signest 
	code_api,opt_speed|linux.persist-use_FLAKY 
	code_api,opt_speed|client.call-retarget 
	code_api,opt_speed|client.count-ctis 
	code_api,opt_speed|client.count-ctis-noopt 
	code_api,opt_speed|client.syscall 
	code_api,opt_speed|client.flush 
	code_api,opt_speed|client.inline 
	code_api,opt_speed|client.drreg-test 
	code_api,opt_speed|client.drwrap-test 
	code_api,opt_speed|client.pcache-use 
	code_api,opt_speed|sample.bbbuf 
	code_api,opt_speed|sample.modxfer 
	code_api,opt_speed|sample.modxfer_app2lib 
	code_api,opt_speed|sample.instrcalls 
	code_api,opt_speed|tool.drcov.fib 
	code_api,thread_private|linux.child-whitelist 
	code_api,thread_private|linux.child-blacklist 
	code_api,thread_private|linux.freeze_FLAKY 
	code_api,thread_private|linux.persist_FLAKY 
	code_api,thread_private|linux.persist-use_FLAKY 
	code_api,thread_private|client.syscall =>  Application /work/dr/nightly/run/build_debug-internal-32/suite/tests/bin/client.syscall (28252).  Internal Error: DynamoRIO debug check failure: /work/dr/git/src/core/fragment.c:5648 thread_owns_no_locks(dcontext) || thread_owns_one_lock(dcontext, thread_initexit_lock) || thread_owns_two_locks(dcontext, thread_initexit_lock, all_threads_synch_lock) 
[...]

PR #4832, opened 2021-04-01
i#4800: Run long test suite on push to master

abhinav92003:
Enable long test suite for push-to-master events, except the AArch64 one.

We want to enable long tests on master merges to improve coverage of the test suite. There are some tests which are never run regularly, which is not good.

Passes CI trigger and branch as environment variables to `runsuite_wrapper.pl`. These are used in CMake script to set `TEST_LONG`, which enables the long test suite.

Also fixes the Policy CMP0077 warning due to using an existing cmake normal variable `TEST_LONG` as a cache variable as well. Adds a new `RUN_LONG_SUITE` variable instead that can be used as a command line option `-DRUN_LONG_SUITE=ON` to cmake, which sets `TEST_LONG` internally.

The AArch64 long suite will be enabled separately by modifying Jenkins config. This PR is for non-AArch64 test suites, run using Github Actions CI only.

Issue: #4800
Fixes: #4791


.github/workflows/ci-clang-format.yml
```diff
@@ -78,6 +78,8 @@ jobs:
         DYNAMORIO_CROSS_AARCHXX_LINUX_ONLY: no
         CC: clang-9
         CXX: clang++-9
+        CI_TRIGGER: ${{ github.event_name }}
```
abhinav92003, line 81: We don't need to pass `CI_TRIGGER` and `CI_BRANCH` in all CI workflows, but it seems a good idea to be consistent and do it anyway.

suite/runsuite_wrapper.pl
```diff
@@ -264,6 +306,63 @@
...
+                'desktop,coarse_freeze_at_exit|common.floatpc' => 1, # i#4845
+                'desktop,coarse_freeze_at_exit|common.floatpc_xl8all' => 1, # i#4845
+                'desktop,coarse_freeze_at_exit|common.getretaddr' => 1, # i#4845
+                'desktop,coarse_freeze_at_exit|common.nativeexec' => 1, # i#4845
+                'desktop,coarse_freeze_at_exit|common.protect-dstack' => 1, # i#4845
+                'desktop,coarse_freeze_at_exit|common.segfault' => 1, # i#4845
+                'finite_shared_bb_cache,cache_shared_bb_regen|common.nativeexec' => 1, # i#4845
+                'finite_shared_trace_cache,cache_shared_trace_regen|common.nativeexec' => 1, # i#4845
                 );
             $issue_no = "#2145";
```
abhinav92003, line 367: I don't think there's a way to map each test with it's own issue number, and IMHO there isn't a strong need either. Let me know though if you think otherwise.

suite/tests/CMakeLists.txt
```diff
@@ -52,7 +52,7 @@ cmake_minimum_required(VERSION 3.7)
 include(../../make/policies.cmake NO_POLICY_SCOPE)
 include(../runsuite_common_pre.cmake) # For _DR_set_VS_bitwidth_env_vars.
 
-option(TEST_LONG "run long set of tests")
+option(RUN_LONG_SUITE "run long set of tests")
```
abhinav92003, line 55: I don't think `TEST_LONG` was documented anywhere. I'll add something for `RUN_LONG_SUITE` once this PR is submitted.

abhinav92003:
See https://github.com/DynamoRIO/dynamorio/runs/2309953156 for the long suite result.

.github/workflows/ci-clang-format.yml
derekbruening: Sure, though it would be nice to have some way to share it...I don't think there is one that's worth the trouble though.

suite/runsuite_wrapper.pl
```diff
@@ -231,6 +231,48 @@
...
+                'prof_pcs|common.decode-bad' => 1, # i#4845
+                'prof_pcs|common.decode' => 1, # i#4845
+                'prof_pcs|common.decode-stress' => 1, # i#4845
+                'prof_pcs|common.eflags' => 1, # i#4845
+                'prof_pcs|common.floatpc' => 1, # i#4845
+                'prof_pcs|common.floatpc_xl8all' => 1, # i#4845
+                'prof_pcs|common.getretaddr' => 1, # i#4845
+                'prof_pcs|common.nativeexec' => 1, # i#4845
+                'prof_pcs|common.protect-dstack' => 1, # i#4845
+                'prof_pcs|common.segfault' => 1, # i#4845
```
derekbruening, line 249: This looks like the entire set of all prof_pcs tests, with the #1807 limit of `^common`.  Should we instead remove Windows from all prof_pcs runs up front (seeing thread_private below also all failing) and file an issue (if there's not one already) on prof_pcs being broken on WIndows, rather than having every single test on a flaky test list?

suite/runsuite_wrapper.pl
```diff
@@ -231,6 +231,48 @@
...
+                'desktop,coarse_freeze_at_exit|common.decode-bad' => 1, # i#4845
+                'desktop,coarse_freeze_at_exit|common.decode' => 1, # i#4845
+                'desktop,coarse_freeze_at_exit|common.decode-stress' => 1, # i#4845
+                'desktop,coarse_freeze_at_exit|common.eflags' => 1, # i#4845
+                'desktop,coarse_freeze_at_exit|common.floatpc' => 1, # i#4845
+                'desktop,coarse_freeze_at_exit|common.floatpc_xl8all' => 1, # i#4845
+                'desktop,coarse_freeze_at_exit|common.getretaddr' => 1, # i#4845
+                'desktop,coarse_freeze_at_exit|common.nativeexec' => 1, # i#4845
+                'desktop,coarse_freeze_at_exit|common.protect-dstack' => 1, # i#4845
+                'desktop,coarse_freeze_at_exit|common.segfault' => 1, # i#4845
```
derekbruening, line 275: This again looks like every single test (probably would all the non-common ones too if the #1807 limit were lifted) and so seems like it should b dealt with at a higher level than the flaky test list?

suite/runsuite_wrapper.pl
```diff
@@ -264,6 +306,63 @@
...
+                'prof_pcs,thread_private|common.decode-bad' => 1, # i#4845
+                'prof_pcs,thread_private|common.decode' => 1, # i#4845
+                'prof_pcs,thread_private|common.decode-stress' => 1, # i#4845
+                'prof_pcs,thread_private|common.eflags' => 1, # i#4845
+                'prof_pcs,thread_private|common.floatpc' => 1, # i#4845
+                'prof_pcs,thread_private|common.floatpc_xl8all' => 1, # i#4845
+                'prof_pcs,thread_private|common.getretaddr' => 1, # i#4845
+                'prof_pcs,thread_private|common.nativeexec' => 1, # i#4845
+                'prof_pcs,thread_private|common.protect-dstack' => 1, # i#4845
+                'prof_pcs,thread_private|common.segfault' => 1, # i#4845
```
derekbruening, line 340: Ditto here and below.

suite/tests/CMakeLists.txt
```diff
@@ -176,6 +176,11 @@ else ()
   set(SKIP_LESS_DIFFERENTIATED_TESTS_FOR_PULL_REQUEST OFF)
 endif ()
 
+# i#4800: Run long test suite on push-to-master events.
+if (RUN_LONG_SUITE OR ("$ENV{CI_TRIGGER}" STREQUAL "push" AND
+                       "$ENV{CI_BRANCH}" STREQUAL "refs/heads/master"))
```
derekbruening, line 181: (Vs the #4059 check which is for !pull_request: I suppose we don't want the long suite on pushes to project or feature branches so it makes sense to be different.)

suite/runsuite_wrapper.pl
abhinav92003: `prof_pcs|common.logstderr` and `prof_pcs,thread_private|common.logstderr` do pass currently. Do we want them enabled?

derekbruening: Oh, so it's not all of them?  That is interesting.  But common.logstderr runs common.broadfun, so how does it pass but common.broadfun fails??   Strange.  I guess it could go either way.  If it's just the one I would still lean toward disabling the whole group.

abhinav92003: yeah, I disabled the whole group. Just testing now to make sure my changes were correct. Will resolve this comment and re-request review when done.

abhinav92003: Done.

suite/tests/CMakeLists.txt
abhinav92003: Right, the current condition works fine if we don't want long suite on pushes to project or feature branches. 

About running long suite on pushes to project branches only: this may help in verifying if things are ok after each reviewed PR is pushed to the project branch, or else one may see a bunch of long suite failures when the project branch is finally merged to master. If we did want that, I wonder how we'd do it. Commits to feature and project branches both is seen as a push event by Github I think. Not sure how'd we differentiate, so that it triggers on the latter only. 

If that's not needed now, we can let this be as-is.

suite/runsuite_wrapper.pl
abhinav92003: Done.

suite/runsuite_wrapper.pl
abhinav92003: Done.

suite/tests/CMakeLists.txt
derekbruening: I would just leave it as-is.

suite/runsuite_wrapper.pl
derekbruening: This is fine.

derekbruening (approved):

PR #4838, opened 2021-04-02
i#3092 genapi: Refactor dr_defines and remove genapi.pl

### LINKED ISSUE #3092: [cleanup] refactor public interface headers and eliminate patchwork exports via genapi.pl

derekbruening:
Refactors the exported content in globals.h, c_defines.h, and
globals_shared.h into a new file core/lib/globals_api.h, which is
exported as dr_defines.h.

Removes the duplicate printf macros in globals_shared.h in favor of
those in globals_api.h.

Renames mcxtx.h to mcxtx_api.h, exports it as dr_mcxtx.h, and resorts
to an ifdef to include the internal name vs the external.

Drops support for non-HAVE_FVISIBILITY.  This allows us to remove
genapi.pl altogether.

Documents the _api.h naming convention.

Fixes #3092


derekbruening:
Windows doesn't know `size_t` but only 32-bit (tested 64-bit locally): not sure how it worked before since the windows.h include was inside API_EXPORT_ONLY.  But it seems best to have that always there.

abhinav92003 (approved):

PR #4839, opened 2021-04-04
Remove two (now tested) cases of ASSERT_NOT_TESTED for AARCH64

sapostolakis:
PR #3003 added handling of a signal interrupting the ibl hit path's jump, but the introduced
client.cleancallsig test could not test the ibl jump on AArchXX leading to the insertion of an
ASSERT_NOT_TESTED() for AARCHXX. Testing on a proprietary app with plain DR (debug build)
on AArch64 hit this case without problems. Thus, we can remove the ASSERT_NOT_TESTED() 
for AARCH64.

The code inserted in PR #4797 was triggered and tested for AARCH64 on a proprietary app so 
remove the ASSERT_NOT_TESTED() for AARCH64 from there as well.

### DIFF
--- a/core/arch/emit_utils_shared.c
+++ b/core/arch/emit_utils_shared.c
@@ -5555,10 +5555,6 @@ emit_special_ibl_xfer(dcontext_t *dcontext, byte *pc, generated_code_t *code, ui
         INSTR_CREATE_cbz(dcontext, opnd_create_instr(skip_unlinked_tgt_jump),
                          opnd_create_reg(SCRATCH_REG5)));
     insert_shared_restore_dcontext_reg(dcontext, &ilist, NULL);
-    /* i#4670: The unlinking case is observed to hit very infrequently on x86. The issue
-     *  or the fix have not been observed on AArchXX yet.
-     */
-    ASSERT_NOT_TESTED();
 #        if defined(AARCH64)
     APP(&ilist,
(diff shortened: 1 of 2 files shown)


core/arch/emit_utils_shared.c
```diff
@@ -5555,17 +5555,17 @@ emit_special_ibl_xfer(dcontext_t *dcontext, byte *pc, generated_code_t *code, ui
         INSTR_CREATE_cbz(dcontext, opnd_create_instr(skip_unlinked_tgt_jump),
```
derekbruening, line 5555: > iX: r

We don't put "iX" in the title: just the branch name.  So `s/iX: r/R/`.

derekbruening, line 5555: > PR #3003 added handling of signal interrupting the ibl hit path's jump, but the introduced

nit: `a signal`

> a
> ASSERT_NOT_TESTED() 

nit: `s/a/an/`

> Testing on proprietary app with plain DR (debug build)

nit: `a proprietary`

derekbruening (approved):

core/arch/emit_utils_shared.c
sapostolakis: Got it.

sapostolakis: done.

Issue #4816, opened 2021-03-29
TBL simulator fails to add tag,pid pair lookup to parent methods


derekbruening:
For #1738 I'm adding a hashtable to speed up cache tag location for high-associativity caches.  Looking at the TLB simulator to see how this change impacts it: it seems broken today, since it imposes the pair of tag,pid as the new lookup key, yet only overrides simulator_t::request!  It seems like invalidate(), contains_tag(), and propagate_eviction() are all just broken for the TLB simulator since they only look up the tag?

Issue #1738, opened 2015-07-15
drcachesim: optimize cache simulator


zhaoqin:
Currently, the cache simulator is ~500x of native execution, the overhead including profiling overhead, communication overhead, but the cache simulator's overhead dominates the overall slowdowns.

One simple optimization is to parallel the cache simulator by splitting the memory into sub-regions and runs a cache simulator for each sub-region.

zhaoqin:
Xref original issue #1703

peterpengwei:
Does multithreading sound like a good solution to alleviate the issue? My initial thought is to assign each cache an independent pthread. The LLC thread contains a pthread mutex for all the I&D caches to arbiter it. If it sounds good, I will start to implement it to see if it helps.

zhaoqin:
No, you should not parallel the cache simulator by assign each cache as an independent pthread.
There would be significant communication overhead dominate the slowdown.

zhaoqin:
The right way should be split the cache into subregion, and each subregion is simulated by one thread.
Fro example, you can use 4 threads to simulate memory reference address from [4N, 4N+cacheline), [4N+cacheline, 4N+2xcacheline), [4N+2xcacheline, 4N+3xcacheline), and [4N+3xcacheline, 4N+4cacheline). By doing that, there would be no communication among the four threads, and should gain the max parallelization.
The potential downside is if the memory reference might concentrate on one or two cache,.

derekbruening:
With larger cache hierarchies and higher associativity (such as simulator a full 2-socket Skylake system) I'm seeing significant time spent walking the ways looking for tags, particularly in invalidate() (this is with coherence turned on as well).  I found that inserting a hashtable (if it's initialized to a large enough starting size) results in a 15% speedup for my setup.  I'll post the PR.

Issue #2056, opened 2016-11-08
raw2trace is slow on large traces


derekbruening:
Split from #1729

On a 67GB bzip2 test raw trace, raw2trace ran for 53 mins and made a 156GB file before running out of disk space.

It took 1:09 to process a 1.1GB raw file and produce a 4GB trace.

We could cache the instr type info to avoid re-decoding on every bb *execution*.

If we do that, we should share it with the analysis tool opcode query impl if we end up adding such a thing.

derekbruening:
I measured different hashtables that simply mapped a PC to an instr_t*.  I disabled the disk write as it added too much variability.

My simple benchmark is post-processing a trace from gzip:
```
$ for ((i=0; i<3; i++)); do /usr/bin/time clients/bin64/drraw2trace -indir drmemtrace.gzip.01022.4725.dir/raw -out /tmp/x; done
```

The results:

always-decode:
20.37user 0.05system 0:20.44elapsed 99%CPU (0avgtext+0avgdata 28052maxresident)k
20.26user 0.04system 0:20.31elapsed 99%CPU (0avgtext+0avgdata 28060maxresident)k
20.29user 0.03system 0:20.34elapsed 99%CPU (0avgtext+0avgdata 28060maxresident)k

std::map.find:
6.96user 0.18system 0:07.15elapsed 99%CPU (0avgtext+0avgdata 7984maxresident)k
6.68user 0.07system 0:06.76elapsed 99%CPU (0avgtext+0avgdata 8112maxresident)k
6.76user 0.04system 0:06.80elapsed 99%CPU (0avgtext+0avgdata 7972maxresident)k

std::map.lower_bound:
6.82user 0.03system 0:06.86elapsed 99%CPU (0avgtext+0avgdata 8012maxresident)k
6.99user 0.03system 0:07.03elapsed 99%CPU (0avgtext+0avgdata 8120maxresident)k
6.88user 0.04system 0:06.93elapsed 99%CPU (0avgtext+0avgdata 7984maxresident)k

std::tr1::unordered_map.find:
6.13user 0.04system 0:06.18elapsed 99%CPU (0avgtext+0avgdata 7768maxresident)k
6.02user 0.04system 0:06.07elapsed 99%CPU (0avgtext+0avgdata 7708maxresident)k
6.10user 0.02system 0:06.14elapsed 99%CPU (0avgtext+0avgdata 7928maxresident)k

c++11 std::unordered_map.find:
[...]

PR #4843, opened 2021-04-07
i#1738 dr$sim perf: Improve tag lookup speed with a hashtable

derekbruening:
Replaces drcachesim's loops over all ways with a hashtable lookup.
For larger cache hierarchies and caches with higher associativity this
increases performance by 15% in cpu-bound tests on offline traces,
when we use a large initial table size to avoid resizes which seem to
outweigh the gains.

The hashtable unfortunately results in a 15% slowdown on simple cache
hierarchies, due to the extra time in erase() and other maintenance
operations outweighing the smaller gains in lookup.  Thus, we make the
default to *not* use a hashtable and use the original linear walk,
providing a method to optionally enable the hashtable.  The cache
simulator enables the hashtables for any 3+-level cache hierarchy with
either coherence or many cores.

Adds coherence to some existing 3-level-hierarchy tests to ensure we
have tests that cover the hashtable path.

The TLB simulator will need to tweak these hashtables: but it looks
like it is already doing the wrong thing in invalidate() and other
simulator_t methods, filed as #4816.

Issue: #1738, #4816

### DIFF
--- a/clients/drcachesim/simulator/cache_simulator.cpp
+++ b/clients/drcachesim/simulator/cache_simulator.cpp
@@ -363,8 +363,11 @@ cache_simulator_t::cache_simulator_t(std::istream *config_file)
         success_ = false;
         return;
     }
-    // TODO add comment and name "32"
-    if (other_caches_.size() > 0 && (knobs_.model_coherence || knobs_.num_cores > 32)) {
+    // For larger hierarchies, especially with coherence, using hashtables
+    // for faster lookups provides performance wins as high as 15%.
+    // However, hashtables can slow down smaller hierarchies, so we only
+    // enable if we anticipate a win.
+    if (other_caches_.size() > 0 && (knobs_.model_coherence || knobs_.num_cores >= 32)) {
         for (auto &cache : all_caches_) {
             cache.second->set_hashtable_use(true);
         }
--- a/clients/drcachesim/simulator/cache_simulator.cpp
+++ b/clients/drcachesim/simulator/cache_simulator.cpp
@@ -1,5 +1,5 @@
 /* **********************************************************
- * Copyright (c) 2015-2020 Google, Inc.  All rights reserved.
+ * Copyright (c) 2015-2021 Google, Inc.  All rights reserved.
  * **********************************************************/
 
 /*
@@ -362,6 +362,12 @@ cache_simulator_t::cache_simulator_t(std::istream *config_file)
         ERRMSG("Usage error: failed to initialize snoop filter.\n");
         success_ = false;
         return;
+    }
+    // TODO add comment and name "32"
+    if (other_caches_.size() > 0 && (knobs_.model_coherence || knobs_.num_cores > 32)) {
+        for (auto &cache : all_caches_) {
+            cache.second->set_hashtable_use(true);
+        }
     }
 }
 
--- a/clients/drcachesim/simulator/caching_device.h
+++ b/clients/drcachesim/simulator/caching_device.h
@@ -36,7 +36,6 @@
 #ifndef _CACHING_DEVICE_H_
 #define _CACHING_DEVICE_H_ 1
 
-#include <iostream> //NOCHECK
 #include <functional>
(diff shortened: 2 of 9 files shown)


clients/drcachesim/simulator/cache_simulator.cpp
```diff
@@ -363,6 +363,15 @@ cache_simulator_t::cache_simulator_t(std::istream *config_file)
         success_ = false;
         return;
     }
+    // For larger hierarchies, especially with coherence, using hashtables
```
xyzsam, line 366: Why is it that we don't enable hashtables through the other constructor which takes knobs instead?

clients/drcachesim/simulator/caching_device.h
```diff
@@ -121,6 +137,33 @@ class caching_device_t {
...
+        if (use_tag2block_table_) {
+            if (block->tag_ != TAG_INVALID)
+                tag2block.erase(block->tag_);
+            tag2block[new_tag] = std::make_pair(block, way);
+        }
+        block->tag_ = new_tag;
+    }
+
+    // Returns the block (and its way) whose tag equals `tag`.
+    // Returns <nullptr,0> is there is no such block.
```
xyzsam, line 163: is -> if

clients/drcachesim/simulator/caching_device.h
```diff
@@ -161,6 +204,16 @@ class caching_device_t {
     addr_t last_tag_;
     int last_way_;
     int last_block_idx_;
+    // Optimization: keep a hashtable for quick lookup of {block,way}
+    // given a tag, if using a large cache hierarchy where serial
+    // walks over the associativity end up as bottlenecks.
+    // We can't easily remove the blocks_ array and replace with just
+    // the hashtable as replace_which_way(), etc. want quick access to
+    // every way for a given line index.
+    std::unordered_map<addr_t, std::pair<caching_device_block_t *, int>,
```
xyzsam, line 213: Two comments:

1. absl containers are open-source, have we considered migrating to them? I wonder how much perf we could get with that. Obviously not for this PR of course.
2. Is it necessary to provide the hash function signature? IIUC `std::hash<addr_t>` is expected to return `std::size_t`.

clients/drcachesim/simulator/tlb.h
```diff
@@ -45,6 +45,10 @@ class tlb_t : public caching_device_t {
     void
     request(const memref_t &memref) override;
 
+    // TODO i#4816: This tag,pid pair lookup needs to be imposed on the parent
```
xyzsam, line 48: What lookup are you referring to? is this a misplaced/stale comment?

xyzsam (approved):

clients/drcachesim/simulator/cache_simulator.cpp
derekbruening: It is not possible to configure more than 2 levels of caches with knobs, so, at least according to my local runs, hashtables would not be a win for any knob config.

clients/drcachesim/simulator/caching_device.h
derekbruening: Done.

clients/drcachesim/simulator/caching_device.h
derekbruening: 1. That is worth considering.  There are other places where the std tables were sub-optimal: I think it was raw2trace.  Here we go: https://github.com/DynamoRIO/dynamorio/issues/2056#issuecomment-333167409.  I filed #4844 on absl.

2. It seems to be required for the hash function as a lambda.

clients/drcachesim/simulator/tlb.h
derekbruening: Will reword: the TLB simulator keys on the tag plus the pid, and when I went to have it disallow use of these hashtables (it inherits from the same cachine_device_t) I found it seems to not consistently use the tag,pid pair in some existing code; hence I filed #4816 and put some comments in several places.

Issue #4842, opened 2021-04-07
Add drcachesim unit tests


derekbruening:
Copying from https://github.com/DynamoRIO/dynamorio/issues/4841#issuecomment-814452409

Unfortunately drcachesim does not have good unit tests today: mostly it has end-to-end tests comparing golden output with broad regexes to account for system differences, which are not ideal. Xref #3195, though simple tests could be created just using short local arrays of memref_t.

This issue covers adding such unit tests to provide better confidence and regression testing of simulator features than the end-to-end broad regexes provide.

derekbruening:
We do have a few unit tests here: https://github.com/DynamoRIO/dynamorio/blob/master/clients/drcachesim/tests/drcachesim_unit_tests.cpp

derekbruening:
Making explicit some unit tests we would like to start with:

+ Testing each cache replacement algorithm.  Xref the LRU bug #4881: make sure the new LRU test catches that bug.
+ Testing coherence

PR #4850, opened 2021-04-15
Fix flaky regex that doesn't match smaller miss rates

abhinav92003:
Modifies miss rate regex to allow matching rates less than 10%, by making the first digit optional.

Due to this issue, `code_api|tool.drcachesim.filter-no-i` failed in `vs2017-32` at github.com/DynamoRIO/dynamorio/actions/runs/750037355.

### DIFF
--- a/clients/drcachesim/tests/filter-no-d.templatex
+++ b/clients/drcachesim/tests/filter-no-d.templatex
@@ -6,7 +6,7 @@ Core #0 \(1 thread\(s\)\)
     Hits:                         *[0-9,\.]*
     Misses:                       *[0-9,\.]*
     Invalidations:                *0
-.*   Miss rate:                   *[1-9][0-9][,\.]..%
+.*   Miss rate:                   *[1-9]?[0-9][,\.]..%
   L1D stats:
     Hits:                                0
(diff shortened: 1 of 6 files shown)


abhinav92003:
It seems that `code_api|tool.drcachesim.filter-no-i` is run only on master merges only, but not as part of the long suite.

derekbruening:
> It seems that `code_api|tool.drcachesim.filter-no-i` in `ci-windows/vs2017-32` is run only on master merges only, but not as part of the long suite.

Probably that's part of the #4059 effort to reduce PR time on 32-bit Windows tests.

derekbruening (approved):
What we should do is make unit tests (filed as #4842) with small controlled synthetic trace sequences: these tests on larger non-deterministic apps have to relax the regexes so much that they are not really testing the details.

abhinav92003:
> What we should do is make unit tests (filed as #4842) with small controlled synthetic trace sequences: these tests on larger non-deterministic apps have to relax the regexes so much that they are not really testing the details.

I agree. There could be more cases where the regexes can be relaxed, but I'm yet to see test failures due to others.

Issue #4316, opened 2020-06-04
Add AArch64 support for restartable sequences


derekbruening:
This covers adding AArch64 support to bring it into parity with x86 for restartable sequence ("rseq") handling which was added as part of #2350.

derekbruening:
Rseq requries a Linux kernel version 4.18+.  Both the TX1 and Jenkins are too old.  I have an internal machine I am currently testing on, but for regression tests it would be best if the Jenkins machine actually had rseq support (the test will still gracefully pass by just doing nothing if rseq is not there).  @AssadHashmi would it be possible to upgrade the Jenkins kernel to 4.18+ for rseq support?

AssadHashmi:
@derekbruening we will shortly be migrating the existing Jenkins to a newer machine: an Ampere Altra!
We would prefer to wait and do such work as part of that migration.
We don't have an accurate ETA yet but the request has been approved and provisioning will be scheduled shortly.
I'll contact them requesting kernel version >4.18.

derekbruening:
Thanks, sounds good.

derekbruening:
I enabled 3 linux.rseq* tests under QEMU because they pass there, but on inspection QEMU is returning ENOSYS for rseq: so they are not testing full functionality.

PR #4854, opened 2021-04-15
i#4316 a64 rseq: Port rseq mangling to AArch64

derekbruening:
Ports rseq mangling to AArch64.  This requires implementing
patch_mov_immed_arch() (done by leveraging existing raw encoding code
from exit stub support) and spilling extra scratch registers in
multiple places.

Expands translation support to cover the new mangling.  This includes
adding proper recognition of an mcontext base load, which was
incorrectly identified as an indirect branch target load on x86.

Ports the heuristic for finding the app's rseq TLS offset on attach
(when we did not see an rseq syscall) for AArch64 to look forward
instead of backward as on x86 where negative segment offsets are used.

Enables the rseq tests for AArch64.
Updates the rseq docs to state that AArch64 is supported.

One final step is to support stores with writeback, which are seen in
real rseq sequences.  That will be done separately.  This completes
porting the x86 support to AArch64.

Issue: #4316

### DIFF
--- a/api/docs/bt.dox
+++ b/api/docs/bt.dox
@@ -1312,7 +1312,7 @@ commit.
 
 This run-twice approach is subject to the following limitations:
 
-- Only x86 is supported for now (no arm or aarch64 support yet).
+- Only x86 and aarch64 are supported for now, and 32-bit x86 is not as well-tested.
 - The application must store an rseq_cs struct for each rseq region in a
   section of its binary named "__rseq_cs", optionally with an "__rseq_cs_ptr_array"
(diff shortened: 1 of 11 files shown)


derekbruening:
WIll add to the commit description:

Tested all 6 rseq tests manually on a machine with a newer kernel.  The Jenkins machine is too old and does not have rseq.  3 rseq tests do run on QEMU so those are enabled but until Jenkins is upgraded we will not have perfect regression tests.

abhinav92003 (approved):

PR #4858, opened 2021-04-16
i#4316 a64 rseq: Nop writeback stores

### LINKED ISSUE #4316: Add AArch64 support for restartable sequences

derekbruening:
Adds nop-ing of writeback stores inside rseq regions by replacing them
with adds to preserve the GPR changes but throw away the memory changes.

Updates the documentation.
Adds a test case.

Fixes #4316


derekbruening:
win32 failure is client.winxfer: looks like flaky output mismatch which is #4732

sapostolakis (approved):

Issue #2985, opened 2018-05-02
Add drx_expand() function for VSIB AVX2/AVX-512 scatter/gather.


sdconsta:
When encountering the following instruction:
```
c4 e2 65 90 14 80       vpgatherdd %ymm3,(%rax,%ymm0,4),%ymm2
```
compiled from the intrinsic
```
__m256i _mm256_i32gather_epi32 (int const* base_addr, __m256i vindex, const int scale)
#include <immintrin.h>
Instruction: vpgatherdd ymm, vm32x, ymm
CPUID Flags: AVX2
```
with either `clang` or `gcc`. DynamoRIO complains that
```
ERROR: Could not find encoding for: lea    (%rax,%ymm0,4) -> %rbp
SYSLOG_ERROR: Application /home/scott/Documents/covert-cpp/build/bin/DynLoader (27816) DynamoRIO usage error : instr_encode error: no encoding found (see log)
SYSLOG_ERROR: Usage error: instr_encode error: no encoding found (see log) (/home/travis/build/DynamoRIO/dynamorio/core/arch/x86/encode.c, line 2423)
version 7.0.17598, custom build
```
It looks like DynamoRIO is expecting the base address to be 4 bytes, but the actual size is 8 bytes:
https://github.com/DynamoRIO/dynamorio/blob/01357f8ae547b4f4ea66d39a3214aed662428a0d/suite/tests/api/ir_x86_3args.h#L384.

derekbruening:
I can't reproduce this.  Decoding:
```
    const byte bytes[] = { 0xc4, 0xe2, 0x65, 0x90, 0x14, 0x80 };
    disassemble_with_info(dcontext, (byte *)bytes, STDERR, false, true);
```
Results in:
```
 c4 e2 65 90 14 80    vpgatherdd (%rax,%ymm0,4)[4byte] %ymm3 -> %ymm2 %ymm3
```
With 7.0.17652 (HEAD at b9726d5fa76be2c3ffe34fcf48ad73cc1f92858c).

I also don't understand this comment:
> It looks like DynamoRIO is expecting the base address to be 4 bytes, but the actual size is 8 bytes:

The memory operand size is 4 bytes (for each vector fetch): why do you think it should be 8 bytes?

Looking at the code I do see some issues: the disasm names do not match the mnemonics for a couple of related instructions (vpgatherqd and vpgatherqq) which I will fix.  But I do not see how you are getting an incorrect decoding?

[...]

Issue #3823, opened 2019-09-09
Support drreg use outside of insertion phase combined with use in insertion phase


derekbruening:
drreg claims to support bare-bones use outside of the insertion phase, and it does, except it does not do the right thing if there are also drreg users in the insertion phase: it will not keep the reservations in the two phases distinct.  The same problem happens with multiple users in the same app2app or instru2instru phase.

If we do want to support use in other phases (the one being hit now is #2985 ), my proposal would be to add disjoint slot spaces or "arenas".  This may have to be done w/ global state to avoid breaking existing interfaces by adding an arena parameter:
```
drreg_set_arena(); 
drreg_reserve_xxx();
drreg_clear_arena();
```
and drreg_init w/ some DRREG_ flag creating a separate arena of the asked-for-number of slots.

We would need to work through interactions w/ regular insertion drreg and make sure it's going to work in other respects -- reg spilled in app2app will be spilled again (nested) but restored to 1st spill on app barrier.    Other client will not be able to get app's native value: ok to live w/?

hgreving2304:
What about, instead of having global state and these setter/clear methods (still not clear whether this can rely on that we are globally synchronized or would need a lock/mutex), we tie the drreg_reserve_register() and drreg_reserve_aflags() etc. calls to phase? Starting with some default phase (NOPHASE) in case drreg is used outside of drmgr (is this even supported?). Any call to drreg_reserve_register() and similar API will tie the slot it is using to the current drmgr phase. In automatic management and optimizations in insertion phase, only slots of NOPHASE or insertion phase would be considered.

Yes this would not support several users in the same phase. But makes it simpler from an API point of view. If somebody would every need this, one could extend the scheme to passing custom phase types. It should just work then and no big changes would need to be made.

derekbruening:
We had an offline discussion and this is the proposal I'm pushing:

Proposal: have drreg-inserted spills+restores added in the app2app phase marked
with a label or flag which drreg treats as an app instruction in its
insertion phase code.  This will then allow app2app use of registers to be
completely isolated from insertion phase: drreg will restore/respill around
them as necessary and re-use the register set in the most optimal way, just
like is done for app code.  (This also has a nice semantic feel to it:
app2app is supposed to be app code :))

drreg should expose a way to query the label, in case some clients want to
analyze reg liveness or sthg in ways that need to know about the app2app
spills.

This does *not* address multiple app2app events using drreg but I'm
suggesting we punt on that for now.

[...]

Issue #3801, opened 2019-08-27
Add a mechanism for clients to store metadata for no-recreate translation


derekbruening:
Today, clients have to decode from cache instructions to deduce how to
restore state (combined with heap-stored metadata they may have kept).  The
DR core, however, uses instrlist recreation metadata as part of its state
updates for mangling expansions, but does not make it easy for clients to
do the same.  This issue covers trying to add a convenience mechanism to
pass data from the bb event w/ translating=true to the restore_state event.

abhinav92003:
Another use case for passing metadata to state restoration:

The current drreg state restore algorithm does not restore the app value of a gpr/aflags if they are dead at the fault instruction. This aligns with general drreg use. However, drreg should restore even dead gprs/aflags when `drreg_options_t.conservative` is set. It is difficult to achieve this without additional metadata to guide state restoration.

derekbruening:
The original description is a little confusing above.  Let's sharpen the focus here on the
no-instrlist case.

Also xref #5005 on drx's scatter-gather translation decode-only 
approach being fragile; drreg's without an
instrlist is also fragile; drbbdup today bails (we'll have to remedy that for
b/254312104,i#5686 but it again will be fragile): without metadata from labels
in the recreated instrlist, clients are faced with extra complexity of decoding
and using heuristics to guess at their own instrumentation.

This issue is about improving the situation by having someone store ilist labels
for no-ilist fragments, just like DR stores pc ranges.

As an aside: note that we could not recreate even if we wanted to for cases
where clients have non-idempotent instrumentation; so even if we tried to
recreate for pending-delete fragments by making a code copy similar to our
selfmod copy we should still have to handle the no-recreate case.

Maybe we can assume that a small % of fragments are non-recreate, though we
should try to keep the storage cost for keeping metadata low.  Maybe all we need
is a list of PC,label_value pairs.

Is the burden on the client for storing that?  Xref complexities of per-bb data
vs bb deletion and tag re-use (esp for pending-delete).  Maybe DR should provide
a user_data pointer (and drmgr multi-plexes or sthg) per fragment?  Or DR
provides a stored translation-specific user_data.

derekbruening:
drbbdup is a good test case here. drbbdup today bails on no-ilist: it does not implement decoding and heuristics to find the bbdup copies. Can easily trigger failure in the new drx-scattergather-bbdup test (for #5686) by having it return DR_EMIT_STORE_TRANSLATIONS.

PR #4863, opened 2021-04-20
i#2985 scatter-gather: Expand scatter-gather instrs in tracing clients

abhinav92003:
Invokes the existing `drx_expand_scatter_gather` to expand any scatter or gather instrs in app2app phase of drcachesim, memval_simple, memtrace_x86 and memtrace_simple clients.

Fixes an issue with detecting `kmovw`, which leads to not enabling saving of zmm regs and opmask regs. Also adds zmm and opmask regs to mcontext dump in logs.

Defers a `get_scatter_gather_info` call after it is determined that the instr is indeed a scatter or gather. This was causing "instr_get_dst: ordinal invalid" issues in some sample clients.

Skips qword index/value variations of scatter-gather on 32-bit version of the scatter-gather tests, as those variations are not supported by drx currently on 32-bit.

Enables drcachesim tests on the existing `client.drx-scattergather` app and a new all-asm app to verify exact basic counts.

Issue: #2985


derekbruening:
We should add drx_expand_scatter_gather() to all of our tracing samples as well: api/samples/memtrace_simple.c, api/samples/memtrace_x86.c, api/samples/memval_simple.c.

abhinav92003:
This PR is still WIP because I see some issues after the initial `lea` assert failure went away.

```
279: <memory leak: Client 4384 bytes not freed>
279: <Application /usr/local/google/home/sharmaabhinav/dr/build/i2985-1/suite/tests/bin/client.drx-scattergather (234580) DynamoRIO usage error : memory leak detected>
```

abhinav92003:
> We should add drx_expand_scatter_gather() to all of our tracing samples as well

Thanks, will do that.

abhinav92003:
Before the memory leak assert failure, there's actually a `SIGSEGV` first
```
<Application dynamorio/suite/tests/bin/client.drx-scattergather (515389).  Application exception at PC 0x00007f47a0529500.  
Signal 11 delivered to application as default action.
...
<Stopping application dynamorio/suite/tests/bin/client.drx-scattergather (515389)>
<memory leak: Client 4384 bytes not freed>
<Application dynamorio/suite/tests/bin/client.drx-scattergather (515389) DynamoRIO usage error : memory leak detected>
<Usage error: memory leak detected (dynamorio/core/heap.c, line 3987)

```
So I'll focus on the `SIGSEGV` instead of the memory leak check.

abhinav92003:
The `vmovdqu32` marked below seg faults due to an invalid `%rcx` value. Note that the `vpgatherdd` and the subsequent `vmovdqu32` are in their own separate fragments (each scatter/gather has it's own basic block -- see note [here](https://github.com/DynamoRIO/dynamorio/blob/c530f52e3797ece0ac023876b65e42a66657a8ad/ext/drx/drx.c#L2275)). I took a look at the latter fragment, and it seems to have correct spills-restores for `%rcx`. I did however find an issue with the former fragment -- the one with the expansion of `vpgatherdd`.
```
Before instrumentation
TAG  0x00007f65f9c464f9
+0    L3 vpgatherdd {%k1} (%rax,%xmm1,4)[4byte] -> %xmm0 %k1 // has its own fragment
+7    L3 vmovdqu32 {%k0} %xmm0 -> (%rcx)[16byte] // SIGSEGV due to invalid %rcx value.
+13   L3 kmovw  %edx -> %k1
+17   L3 vpgatherdd {%k1} (%rax,%ymm1,4)[4byte] -> %ymm0 %k1
+24   L3 vmovdqu32 {%k0} %ymm0 -> 0x10(%rcx)[32byte]
+34   L3 kmovw  %edx -> %k1
+38   L3 vpgatherdd {%k1} (%rax,%zmm1,4)[4byte] -> %zmm0 %k1
+45   L3 vmovdqu32 {%k0} %zmm0 -> 0x30(%rcx)[64byte]
+55   L3 add    $0x0000000000000000 %rsp -> %rsp
+59   L3 pop    %rsp (%rsp)[8byte] -> %r15 %rsp
+61   L3 pop    %rsp (%rsp)[8byte] -> %r14 %rsp
+63   L3 pop    %rsp (%rsp)[8byte] -> %r13 %rsp
+65   L3 pop    %rsp (%rsp)[8byte] -> %r12 %rsp
+67   L3 pop    %rsp (%rsp)[8byte] -> %rbp %rsp
+68   L3 pop    %rsp (%rsp)[8byte] -> %rbx %rsp
+69   L3 ret    %rsp (%rsp)[8byte] -> %rsp
END 0x00007f65f9c464f9
```
Following is the expansion of the `vpgatherdd` above, plus drcachesim instrumentation. I see multiple issues here:
1. spilled value of `%rcx` is clobbered. Possibly due to a conflict with `%drx` spilled by drcachesim. This could be related to #3823, as drx expansion is part of app2app phase, separate from drcachesim.
2. not all addresses read by the gather are stored. I see instrumentation only for the first one.
3. some instrs that are part of the drx expansion are not marked as app instr. Not sure if this is a real issue, but worth investigating.
```
After instrumentation
TAG  0x00007f65f9c464f9
...
+105  m4 mov    %rax -> %gs:0x00000120[8byte]
+116  m4 lahf    -> %ah
+117  m4 seto    -> %al
+120  m4 mov    %rcx -> %gs:0x00000128[8byte] // lost later (see below). Also, Issue 3?: This is added by drx expansion. Not sure why this is not marked as an app instruction. Same as some `jz`s below.
  +129  m4 <label>
  +129  L3 kmovw  %k1 -> %ecx
  +133  L3 test   %ecx $0x00000001
  +139  m4 jz     @0x00007f63b9e02fb0[8byte]
  +145  L3 vextracti32x4 {%k0} $0x00 %zmm1 -> %xmm2
  +152  L3 vpextrd %xmm2 $0x00 -> %ecx
  +158  m4 mov    %rax -> %gs:0x00000118[8byte]
  +169  m4 mov    %gs:0x00000120[8byte] -> %rax
  +180  m4 mov    %rcx -> %gs:0x00000120[8byte]
    +189  m4 mov    %gs:0x00000100[8byte] -> %rcx
    +198  m4 mov    %rdx -> %gs:0x00000128[8byte]  // Issue 1: clobbers %rcx stored in same slot
      +207  m4 mov    $0x0004000a -> (%rcx)[4byte]
      +213  m4 mov    $0x00007f65f9c464f9 -> %rdx
      +223  m4 mov    %rdx -> 0x04(%rcx)[8byte]
      +227  m4 mov    $0x00040011 -> 0x0c(%rcx)[4byte]
      +234  m4 mov    $0x00007f6306070606 -> %rdx
      +244  m4 mov    %rdx -> 0x10(%rcx)[8byte]
      +248  m4 mov    $0x0003000a -> 0x18(%rcx)[4byte]
      +255  m4 mov    $0x00007f65f9c464f9 -> %rdx
      +265  m4 mov    %rdx -> 0x1c(%rcx)[8byte]
      +269  m4 <label>
      +269  m4 mov    %gs:0x00000120[8byte] -> %rcx
      +278  m4 lea    (%rax,%rcx,4) -> %rdx  // this seems to be drcachesim's store of read mem addr
      +282  m4 mov    %gs:0x00000100[8byte] -> %rcx
      +291  m4 mov    %rdx -> 0x28(%rcx)[8byte]
      +295  m4 mov    $0x00040000 -> 0x24(%rcx)[4byte]
      +302  m4 <label>
      +302  m4 lea    0x30(%rcx) -> %rcx
      +306  m4 mov    %rcx -> %gs:0x00000100[8byte]
      +315  m4 <label>
    +315  m4 mov    %gs:0x00000120[8byte] -> %rcx
  +324  m4 mov    %gs:0x00000128[8byte] -> %rdx
  +333  L3 mov    (%rax,%rcx,4)[4byte] -> %ecx
  +336  L4 vextracti32x4 {%k0} $0x00 %zmm0 -> %xmm2
  +343  L4 vpinsrd %xmm2 %ecx $0x00 -> %xmm2
  +349  L4 vinserti32x4 {%k0} $0x00 %zmm0 %xmm2 -> %zmm0
  +356  L4 mov    $0x00000001 -> %ecx
  +361  m4 mov    %rdx -> %gs:0x00000120[8byte]
    +370  m4 kmovw  %k0 -> %edx
    +374  L4 kmovw  %ecx -> %k0
    +378  L4 kandnw %k0 %k1 -> %k1
    +382  m4 kmovw  %edx -> %k0
  +386  m4 mov    %gs:0x00000120[8byte] -> %rdx
  +395  m4 <label>
  +395  L4 kmovw  %k1 -> %ecx
  +399  L4 test   %ecx $0x00000002
  +405  m4 jz     @0x00007f63b9dff8d0[8byte]
  +411  L4 vextracti32x4 {%k0} $0x00 %zmm1 -> %xmm2
  +418  L4 vpextrd %xmm2 $0x01 -> %ecx
  +424  L4 mov    (%rax,%rcx,4)[4byte] -> %ecx // Issue 2: subsequent reads like this one don't seem to have any corresponding drcachesim instrumentation 
  +427  L4 vextracti32x4 {%k0} $0x00 %zmm0 -> %xmm2
  +434  L4 vpinsrd %xmm2 %ecx $0x01 -> %xmm2
  +440  L4 vinserti32x4 {%k0} $0x00 %zmm0 %xmm2 -> %zmm0
  +447  L4 mov    $0x00000002 -> %ecx
  +452  m4 mov    %rdx -> %gs:0x00000120[8byte]
    +461  m4 kmovw  %k0 -> %edx
    +465  L4 kmovw  %ecx -> %k0
    +469  L4 kandnw %k0 %k1 -> %k1
    +473  m4 kmovw  %edx -> %k0
  +477  m4 mov    %gs:0x00000120[8byte] -> %rdx
  +486  m4 <label>
  +486  L4 kmovw  %k1 -> %ecx
  +490  L4 test   %ecx $0x00000004
  +496  m4 jz     @0x00007f63b9e05348[8byte]
...
  +668  L4 kxorq  %k1 %k1 -> %k1
+673  m4 mov    %gs:0x00000128[8byte] -> %rcx  // Issue 1: restores incorrect value to %rcx.
+682  m4 mov    %rax -> %gs:0x00000120[8byte]
+693  m4 mov    %gs:0x00000118[8byte] -> %rax
+704  m4 cmp    %al $0x81
+706  m4 sahf   %ah
+707  m4 mov    %gs:0x00000120[8byte] -> %rax
+718  m4 <label>
+718  L4 jmp    $0x00007f65f9c46500
END 0x00007f65f9c464f9
```

The hypothesis is supported by the fact that `xcx` and `xdx` have the same value `= 0x00007ffe2c0bffff` when the `SIGSEGV` occurs.

abhinav92003:
The `vmovdqu32` marked below seg faults due to an invalid `%rcx` value. Note that the `vpgatherdd` and the subsequent `vmovdqu32` are in their own separate fragments (each scatter/gather has it's own basic block -- see note [here](https://github.com/DynamoRIO/dynamorio/blob/c530f52e3797ece0ac023876b65e42a66657a8ad/ext/drx/drx.c#L2275)). I took a look at the latter fragment, and it seems to have correct spills-restores for `%rcx`. I did however find an issue with the former fragment -- the one with the expansion of `vpgatherdd`.
```
Before instrumentation
TAG  0x00007f65f9c464f9
+0    L3 vpgatherdd {%k1} (%rax,%xmm1,4)[4byte] -> %xmm0 %k1 // has its own fragment
+7    L3 vmovdqu32 {%k0} %xmm0 -> (%rcx)[16byte] // SIGSEGV due to invalid %rcx value.
+13   L3 kmovw  %edx -> %k1
+17   L3 vpgatherdd {%k1} (%rax,%ymm1,4)[4byte] -> %ymm0 %k1
+24   L3 vmovdqu32 {%k0} %ymm0 -> 0x10(%rcx)[32byte]
+34   L3 kmovw  %edx -> %k1
+38   L3 vpgatherdd {%k1} (%rax,%zmm1,4)[4byte] -> %zmm0 %k1
+45   L3 vmovdqu32 {%k0} %zmm0 -> 0x30(%rcx)[64byte]
+55   L3 add    $0x0000000000000000 %rsp -> %rsp
+59   L3 pop    %rsp (%rsp)[8byte] -> %r15 %rsp
+61   L3 pop    %rsp (%rsp)[8byte] -> %r14 %rsp
+63   L3 pop    %rsp (%rsp)[8byte] -> %r13 %rsp
+65   L3 pop    %rsp (%rsp)[8byte] -> %r12 %rsp
+67   L3 pop    %rsp (%rsp)[8byte] -> %rbp %rsp
+68   L3 pop    %rsp (%rsp)[8byte] -> %rbx %rsp
+69   L3 ret    %rsp (%rsp)[8byte] -> %rsp
END 0x00007f65f9c464f9
```
Following is the expansion of the `vpgatherdd` above, plus drcachesim instrumentation. I see multiple issues here:
1. spilled value of `%rcx` is clobbered. Possibly due to a conflict with `%rdx` spilled by drcachesim. This could be related to #3823, as drx expansion is part of app2app phase, separate from drcachesim.
2. not all addresses read by the gather are stored. I see instrumentation only for the first one.
3. some instrs that are part of the drx expansion are not marked as app instr. Not sure if this is a real issue, but worth investigating.

(Indentation in the code below is just to show spills and restores of `%rcx` and `%rdx`)
```
After instrumentation
TAG  0x00007f65f9c464f9
...
+105  m4 mov    %rax -> %gs:0x00000120[8byte]
+116  m4 lahf    -> %ah
+117  m4 seto    -> %al
+120  m4 mov    %rcx -> %gs:0x00000128[8byte] // lost later (see below). Also, Issue 3?: This is added by drx expansion. Not sure why this is not marked as an app instruction. Same as some `jz`s below.
  +129  m4 <label>
  +129  L3 kmovw  %k1 -> %ecx
  +133  L3 test   %ecx $0x00000001
  +139  m4 jz     @0x00007f63b9e02fb0[8byte]
  +145  L3 vextracti32x4 {%k0} $0x00 %zmm1 -> %xmm2
  +152  L3 vpextrd %xmm2 $0x00 -> %ecx
  +158  m4 mov    %rax -> %gs:0x00000118[8byte]
  +169  m4 mov    %gs:0x00000120[8byte] -> %rax
  +180  m4 mov    %rcx -> %gs:0x00000120[8byte]
    +189  m4 mov    %gs:0x00000100[8byte] -> %rcx
    +198  m4 mov    %rdx -> %gs:0x00000128[8byte]  // Issue 1: clobbers %rcx stored in same slot
      +207  m4 mov    $0x0004000a -> (%rcx)[4byte]
      +213  m4 mov    $0x00007f65f9c464f9 -> %rdx
      +223  m4 mov    %rdx -> 0x04(%rcx)[8byte]
      +227  m4 mov    $0x00040011 -> 0x0c(%rcx)[4byte]
      +234  m4 mov    $0x00007f6306070606 -> %rdx
      +244  m4 mov    %rdx -> 0x10(%rcx)[8byte]
      +248  m4 mov    $0x0003000a -> 0x18(%rcx)[4byte]
      +255  m4 mov    $0x00007f65f9c464f9 -> %rdx
      +265  m4 mov    %rdx -> 0x1c(%rcx)[8byte]
      +269  m4 <label>
      +269  m4 mov    %gs:0x00000120[8byte] -> %rcx
      +278  m4 lea    (%rax,%rcx,4) -> %rdx  // this seems to be drcachesim's store of read mem addr
      +282  m4 mov    %gs:0x00000100[8byte] -> %rcx
      +291  m4 mov    %rdx -> 0x28(%rcx)[8byte]
      +295  m4 mov    $0x00040000 -> 0x24(%rcx)[4byte]
      +302  m4 <label>
      +302  m4 lea    0x30(%rcx) -> %rcx
      +306  m4 mov    %rcx -> %gs:0x00000100[8byte]
      +315  m4 <label>
    +315  m4 mov    %gs:0x00000120[8byte] -> %rcx
  +324  m4 mov    %gs:0x00000128[8byte] -> %rdx
  +333  L3 mov    (%rax,%rcx,4)[4byte] -> %ecx
  +336  L4 vextracti32x4 {%k0} $0x00 %zmm0 -> %xmm2
  +343  L4 vpinsrd %xmm2 %ecx $0x00 -> %xmm2
  +349  L4 vinserti32x4 {%k0} $0x00 %zmm0 %xmm2 -> %zmm0
  +356  L4 mov    $0x00000001 -> %ecx
  +361  m4 mov    %rdx -> %gs:0x00000120[8byte]
    +370  m4 kmovw  %k0 -> %edx
    +374  L4 kmovw  %ecx -> %k0
    +378  L4 kandnw %k0 %k1 -> %k1
    +382  m4 kmovw  %edx -> %k0
  +386  m4 mov    %gs:0x00000120[8byte] -> %rdx
  +395  m4 <label>
  +395  L4 kmovw  %k1 -> %ecx
  +399  L4 test   %ecx $0x00000002
  +405  m4 jz     @0x00007f63b9dff8d0[8byte]
  +411  L4 vextracti32x4 {%k0} $0x00 %zmm1 -> %xmm2
  +418  L4 vpextrd %xmm2 $0x01 -> %ecx
  +424  L4 mov    (%rax,%rcx,4)[4byte] -> %ecx // Issue 2: subsequent reads like this one don't seem to have any corresponding drcachesim instrumentation 
  +427  L4 vextracti32x4 {%k0} $0x00 %zmm0 -> %xmm2
  +434  L4 vpinsrd %xmm2 %ecx $0x01 -> %xmm2
  +440  L4 vinserti32x4 {%k0} $0x00 %zmm0 %xmm2 -> %zmm0
  +447  L4 mov    $0x00000002 -> %ecx
  +452  m4 mov    %rdx -> %gs:0x00000120[8byte]
    +461  m4 kmovw  %k0 -> %edx
    +465  L4 kmovw  %ecx -> %k0
    +469  L4 kandnw %k0 %k1 -> %k1
    +473  m4 kmovw  %edx -> %k0
  +477  m4 mov    %gs:0x00000120[8byte] -> %rdx
  +486  m4 <label>
  +486  L4 kmovw  %k1 -> %ecx
  +490  L4 test   %ecx $0x00000004
  +496  m4 jz     @0x00007f63b9e05348[8byte]
...
  +668  L4 kxorq  %k1 %k1 -> %k1
+673  m4 mov    %gs:0x00000128[8byte] -> %rcx  // Issue 1: restores incorrect value to %rcx.
+682  m4 mov    %rax -> %gs:0x00000120[8byte]
+693  m4 mov    %gs:0x00000118[8byte] -> %rax
+704  m4 cmp    %al $0x81
+706  m4 sahf   %ah
+707  m4 mov    %gs:0x00000120[8byte] -> %rax
+718  m4 <label>
+718  L4 jmp    $0x00007f65f9c46500
END 0x00007f65f9c464f9
```

The hypothesis is supported by the fact that `xcx` and `xdx` have the same value `= 0x00007ffe2c0bffff` when the `SIGSEGV` occurs.

abhinav92003:
> We should add drx_expand_scatter_gather() to all of our tracing samples as well: api/samples/memtrace_simple.c, api/samples/memtrace_x86.c, api/samples/memval_simple.c.

Done. I tested that these sample clients are working fine with the `drx-scattergather` app, except the `memval_simple` client which shows the following assert failure:
`ASSERT FAILURE: dynamorio/api/samples/memval_simple.c:150: write_base <= write_ptr ()`. 

Seems like something specific to `memval_simple`, looking into it right now. PTAL remaining PR.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1216,6 +1216,9 @@ event_bb_app2app(void *drcontext, void *tag, instrlist_t *bb, bool for_trace,
         DR_ASSERT(false);
         /* in release build, carry on: we'll just miss per-iter refs */
     }
+    if (!drx_expand_scatter_gather(drcontext, bb, NULL)) {
```
derekbruening, line 1219: Xref my comment at https://github.com/DynamoRIO/dynamorio/issues/2985#issuecomment-826920909 -- I think online traces will have the wrong instr sizes until #4865 is in place (another area where we don't have good detailed testing...all we have are vague "did it print some counts" and no checks on types or sizes).  Prob worth a `XXX #4865` comment for online traces.

ext/drreg/drreg.c
```diff
@@ -173,15 +173,38 @@ get_where_app_pc(instr_t *where)
...
+ * previous instrumentation pass, and tells us that it is not safe to reuse
+ * the slot in the current pass.
+ */
+static bool
+has_pending_slot_usage_by_prior_pass(per_thread_t *pt, instrlist_t *ilist, instr_t *where,
+                                     uint slot)
+{
+    if (!TEST(DRREG_HANDLE_MULTI_PHASE_SLOT_RESERVATIONS, pt->bb_props))
+        return false;
+    for (instr_t *in = where; in != NULL /*sanity*/ && in != instrlist_last(ilist);
```
derekbruening, line 187: Why exclude last?  Why not just check for null?  It seems like a drreg restore could be the last instr?

ext/drreg/drreg.c
```diff
@@ -209,6 +232,9 @@ spill_reg(void *drcontext, per_thread_t *pt, reg_id_t reg, uint slot, instrlist_
         dr_spill_slot_t DR_slot = (dr_spill_slot_t)(slot - ops.num_spill_slots);
         dr_save_reg(drcontext, ilist, where, reg, DR_slot);
     }
+    if (TEST(DRREG_HANDLE_MULTI_PHASE_SLOT_RESERVATIONS, pt->bb_props)) {
+        instr_set_note(instr_get_prev(where), (void *)(ptr_uint_t)slot);
```
derekbruening, line 236: Notes need to have conventions to avoid conflicts.  Use drmgr_reserve_note_range() to get unique values, and only put notes on instructions this code is inserting: I guess that last part is the case here.  It might be worth a comment that this assumes instr_get_prev grabs one of the (potentially multiple for dr_save_reg?) spill instrs to avoid someone placing code in between.

ext/drreg/drreg.h
```diff
@@ -324,6 +324,15 @@ typedef enum {
      * to be handled by the user manually, usually via drreg_restore_all().
      */
     DRREG_USER_RESTORES_AT_BB_END = 0x004,
+
+    /**
+     * Turns on stricter logic to find free register spill slots. This makes sure
+     * that a slot wasn't used to spill some register value in prior instrumentation
+     * passes. An example usage is in drx_expand_scatter_gather which is used in
```
derekbruening, line 331: nit: missing () to make doxygen link

ext/drreg/drreg.c
```diff
@@ -235,6 +261,9 @@ restore_reg(void *drcontext, per_thread_t *pt, reg_id_t reg, uint slot,
         dr_spill_slot_t DR_slot = (dr_spill_slot_t)(slot - ops.num_spill_slots);
         dr_restore_reg(drcontext, ilist, where, reg, DR_slot);
     }
+    if (release && TEST(DRREG_HANDLE_MULTI_PHASE_SLOT_RESERVATIONS, pt->bb_props)) {
+        instr_set_note(instr_get_prev(where), (void *)(ptr_uint_t)slot);
```
derekbruening, line 265: Ditto

ext/drreg/drreg.h
```diff
@@ -324,6 +324,15 @@ typedef enum {
...
+     * the app2app pass and requires spilling of registers to slots that may
+     * conflict with slots used during later instrumentation passes.
+     */
+    DRREG_HANDLE_MULTI_PHASE_SLOT_RESERVATIONS = 0x008,
```
derekbruening, line 335: Should this be on by default, instead of off?  It has a performance cost to walk the ilist every single time, and there could be a problem if code were somewhere inserted beyond the current instr: there are many cases of postinserts, but not by drreg code with unique notes.

I think the biggest issues would be instru2instru phases: but they already have conflict issues and might want this on anyway?  Although actually bb_props is set to 0 in the insertion phase?  **Should that be moved to a late instru2instru event?**

johnfxgalea:
Would it be possible to add a test for drreg please?

derekbruening:
> Would it be possible to add a test for drreg please?

+1.  I could also see splitting the drreg changes out to separate them from the changes to the tracing samples.

abhinav92003:
> > Would it be possible to add a test for drreg please?
> 
> +1. I could also see splitting the drreg changes out to separate them from the changes to the tracing samples.

Sure, I'll split these changes into multiple master pushes, after I've resolved all comments here.

I found two more issues: one was with the logic to detect `zmm`/`opmask` writing instrs, which was broken for `kmovw` (which writes into the `kx` opmask regs). Due to this we were not saving/restoring `zmm` and `kx` regs. Fixed that.

Also ran into some issues due to conflict with the `xmm` spill reg selection logic, which is very naive currently (https://github.com/DynamoRIO/dynamorio/blob/1640ffe36d46ad68af5f1df67898d1ae069b056f/ext/drx/drx.c#L2358). I worked around it by using high-numbered `xmm` regs in the test. I think PR #3893 would remove that difficulty.

ext/drreg/drreg.c
abhinav92003: Thanks for catching this. I assumed usage like the [C++ std containers](https://en.cppreference.com/w/cpp/container/vector/end) where you stop iterating before the last element. But I see now in documentation of [`instrlist_last`](https://github.com/DynamoRIO/dynamorio/blob/1640ffe36d46ad68af5f1df67898d1ae069b056f/core/ir/instrlist_api.h#L143) that it returns the actual last element.

This usage also looks suspicious and may warrant auditing: https://github.com/DynamoRIO/dynamorio/blob/1640ffe36d46ad68af5f1df67898d1ae069b056f/core/vmareas.c#L4750

api/samples/memtrace_simple.c
```diff
@@ -46,6 +46,8 @@
  * This sample illustrates
  * - the use of drutil_expand_rep_string() to expand string loops to obtain
  *   every memory reference,
+ * - the use of drx_expand_scatter_gather() to expand scatter/gather instrs
+ *   into a set of functionally equivalent stores/loads.
```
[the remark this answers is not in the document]
abhinav92003: Done.

api/samples/memtrace_x86.c
```diff
@@ -46,10 +46,13 @@
...
  *
- * Illustrates the use of drutil_expand_rep_string() to expand string
- * loops to obtain every memory reference and of
- * drutil_opnd_mem_size_in_bytes() to obtain the size of OP_enter
- * memory references.
+ * This sample illustrates
+ * - the use of drutil_expand_rep_string() to expand string loops to obtain
+ *   every memory reference,
+ * - the use of drx_expand_scatter_gather() to expand scatter/gather instrs
+ *   into a set of functionally equivalent stores/loads.
```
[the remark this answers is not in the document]
abhinav92003: Done.

api/samples/memval_simple.c
```diff
@@ -44,6 +44,8 @@
  *   written by it,
  * - the use of drutil_expand_rep_string() to expand string loops to obtain
  *   every memory reference,
+ * - the use of drx_expand_scatter_gather() to expand scatter/gather instrs
+ *   into a set of functionally equivalent stores/loads.
```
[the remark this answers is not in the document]
abhinav92003: Done.

clients/drcachesim/tests/scattergather.templatex
```diff
@@ -0,0 +1,79 @@
...
+#endif
+AVX2/AVX-512 scatter/gather checks ok
+---- <application exited with code 0> ----
+Basic counts tool results:
+Total counts:
+     .* total \(fetched\) instructions
+     .* total unique \(fetched\) instructions
+     .* total non-fetched instructions
+     .* total prefetches
+     .* total data loads
```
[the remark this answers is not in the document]
abhinav92003: Added a pure asm test; it's returning a count of just 1 as expected.

ext/drreg/drreg.h
abhinav92003: > Should this be on by default, instead of off? It has a performance cost to walk the ilist every single time

Do you expect multi-phase (or multiple clients in the same phase) use of drreg to be very frequent? If yes, then I agree we should keep it on by default. Otherwise, I'm concerned about the unnecessary overhead. We can keep the note-keeping on by default, and let clients add this flag if they use drreg outside instrumentation phase (like in instru2instru phase). To catch some of these more common cases, we can maybe add the flag ourselves in drreg if we see a usage in these phases.

> Although actually bb_props is set to 0 in the insertion phase? Should that be moved to a late instru2instru event?

What if the client didn't register any phase after insertion?

ext/drreg/drreg.c
derekbruening: That vmareas loop looks like it wants to stop when it hits the final jmp without the loop body looking at it (it's examined afterward), so the !=last seems correct.

ext/drreg/drreg.h
derekbruening: > Do you expect multi-phase (or multiple clients in the same phase) use of drreg to be very frequent? 

Every address tracing client is going to use drx_expand_scatter_gather, which uses drreg in the app2app phase, so such usage will not be uncommon.

> > Although actually bb_props is set to 0 in the insertion phase? Should that be moved to a late instru2instru event?
>
> What if the client didn't register any phase after insertion?

Not sure I follow the question: there would be no harm then in moving the set to 0 later.

clients/drcachesim/tracer/tracer.cpp
abhinav92003: Done.

ext/drreg/drreg.c
abhinav92003: Okay. Thanks for checking. 

Done.

ext/drreg/drreg.h
abhinav92003: Done.

ext/drreg/drreg.h
abhinav92003: >> Do you expect multi-phase (or multiple clients in the same phase) use of drreg to be very frequent?
>
> Every address tracing client is going to use drx_expand_scatter_gather, which uses drreg in the app2app phase, so such usage will not be uncommon.

The flag is already set when `drx_expand_scatter_gather` is invoked. So that is covered.

My comment was about other user-owned DR clients where scatter/gather instrs may not be expected or not important. I don't suppose we need to do anything proactively for those clients right? Or do you mean we should proactively set the flag everywhere, anticipating uses in multiple passes like instru2instru?

>>> Although actually bb_props is set to 0 in the insertion phase? Should that be moved to a late instru2instru event?
>>
>> What if the client didn't register any phase after insertion?
>
> Not sure I follow the question: there would be no harm then in moving the set to 0 later.

If there's no instru2instru event registered by the user (or if there is one but the user doesn't invoke drreg in it), then we won't be able to reset `bb_props` to zero. Do you mean that drreg should register a new low-priority instru2instru event that will get invoked at the very end, which resets `bb_props`?

ext/drreg/drreg.c
abhinav92003: Added the comment, and the proper note usage.

ext/drreg/drreg.c
abhinav92003: > It might be worth a comment that this assumes instr_get_prev grabs one of the (potentially multiple for dr_save_reg?) spill instrs to avoid someone placing code in between.

Done

> Notes need to have conventions to avoid conflicts. Use drmgr_reserve_note_range() to get unique values,

Done; I added it like drmgr does it. But I don't think I understand how note conflicts are avoided by `drmgr_reserve_note_range()`. I only understand that we prevent conflicts between clients on note _value ranges_, but it's still possible for clients (and core DR too) to overwrite note field set by each other right?

ext/drreg/drreg.h
derekbruening: > The flag is already set when drx_expand_scatter_gather is invoked. So that is covered.
>
> My comment was about other user-owned DR clients where scatter/gather instrs may not be expected or not important. I > don't suppose we need to do anything proactively for those clients right? Or do you mean we should proactively set the > flag everywhere, anticipating uses in multiple passes like instru2instru?

If the cost is not very high it seems better to just have it on and avoid subtle hard-to-debug problems in future clients.  OTOH as you say this app2app scatter-gather case is covered already.
If we leave it off by default: could you add clear comments to the drreg docs so that anyone using it in app2app or instru2instru knows to go set this flag?

> Do you mean that drreg should register a new low-priority instru2instru event that will get invoked at the very end, which resets bb_props?

Yes, drreg would register its own event: it couldn't use an event registered by other code even if it wanted to as it has no idea where other event callbacks are.

ext/drreg/drreg.c
abhinav92003: What is the order between callbacks of multiple clients? Based on https://dynamorio.org/page_deploy.html#multi_client, I understand that each client completes all callbacks before letting other clients run. So, does this mean there cannot be any conflicts in the note field itself?

Therefore, we only need to worry about misinterpretation of note values residual from previous clients (which is solved by note value reservation)?

clients/drcachesim/tests/allasm_scattergather.asm
```diff
@@ -0,0 +1,126 @@
...
+        // Exit.
+        mov      rdi, 0           // exit code
+        mov      eax, 231         // SYS_exit_group
+        syscall
+        .data
+        .align   8
+hello:
+       .string "Hello world!\n"
+equal:
+       .string "Equal\n"
```
abhinav92003, line 122: There seem to be some formatting issues here. Will fix it in the next commit.

ext/drreg/drreg.c
derekbruening: Generally it is multiple components of one client, rather than multiple clients.  I.e., multiple libraries like these extension libraries, plus the end client itself.  drmgr adds priorities used to order handlers: see e.g. DRMGR_PRIORITY_INSERT_DRREG_LOW.

Usually note fields are limited to label instructions, and the creator of the label sets the note.  Thus there are no write conflicts because a new note is put in place by creating a new label.  Reserving note ranges lets later consumers of the data iterate the list and read note values of labels and know which means what.  It might be better to use labels here.

abhinav92003: I used a label instr. I found that labels can have data areas, and using that seemed much cleaner way to put this sort of info. So now I use a special note to mark label instrs that contain spill slot usage info in their data area.

ext/drreg/drreg.h
abhinav92003: I'm now saving data about spill slot usage even when the option isn't set. The new spill slot selection logic itself however is still behind the option .Added documentation to `drreg.dox` and `release.dox` about the need to set this option for multi-phase usage of drreg.

> Yes, drreg would register its own event: it couldn't use an event registered by other code even if it wanted to as it has no idea where other event callbacks are.

Done

derekbruening: > I'm now saving data about spill slot usage even when the option isn't set

Should we consider removing the option then for simplicity and ease of testing?

This data would be useful for the restore_state event: if we had #3801.

abhinav92003: > Should we consider removing the option then for simplicity and ease of testing?

The option is still required to enable the stricter, more expensive spill slot selection logic. The only change I made was to save spill slot usage data regardless of the option, so that the logic works even if the option is turned on after first drreg use.

clients/drcachesim/tests/allasm_scattergather.asm
abhinav92003: Done.

abhinav92003:
I sent the drreg changes for review in a separate PR #4887. But I'll need to keep them here as well (until #4887 is merged) as the scatter/gather changes depend on it.

api/docs/release.dox
```diff
@@ -203,12 +203,18 @@ Further non-compatibility-affecting changes include:
    runtime option which locates guest system libraries and enables workarounds for
    problems with QEMU's threads.
  - Added dr_is_detaching(), an API to query whether detach is in progress.
+ - Added DRREG_HANDLE_MULTI_PHASE_SLOT_RESERVATIONS to drreg_bb_properties_t to
```
abhinav92003, line 206: Next push will fix this duplicate entry.

core/ir/x86/instr.c
```diff
@@ -1346,21 +1346,18 @@ instr_may_write_zmm_or_opmask_register(instr_t *instr)
 {
     if (instr_get_prefix_flag(instr, PREFIX_EVEX))
         return true;
-    if (instr_raw_bits_valid(instr)) {
+    if (instr_raw_bits_valid(instr) &&
```
abhinav92003, line 1349: I'm curious about this check: is it that it is significantly faster than the one below? It's certainly not complete, as it misses out on `kmovw`

clients/drcachesim/tests/offline-allasm-scattergather-basic-counts.templatex
```diff
@@ -0,0 +1,41 @@
+Correct
+Hello world!
+Basic counts tool results:
+Total counts:
+     .* total \(fetched\) instructions
```
derekbruening, line 5: I assume the instruction count is way off from incorrectly separately recording all the expansion instructions, right?  Please put a FIXME i#4865.

We should have a precise number here, right, not a .*?  Ditto for the line below.

clients/drcachesim/tests/scattergather.templatex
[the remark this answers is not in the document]
derekbruening: > it's returning a count of just 1 as expected.

Wait: not understanding -- 1 would not be expected, would it?  And what about the FIXME in the output file about the result being wrong?

core/ir/x86/instr.c
```diff
@@ -1346,21 +1346,18 @@ instr_may_write_zmm_or_opmask_register(instr_t *instr)
...
-                    /* If evex, P[3:2] must be 0 and P[10] must be 1. */
-                    TESTALL(MODRM_BYTE(3, 0, 0), instr_get_raw_byte(instr, 1))) {
-                    if (!TEST(0xC, instr_get_raw_byte(instr, 1)) &&
-                        TEST(0x04, instr_get_raw_byte(instr, 2))) {
-                        return true;
-                    }
-                }
+        instr_length(NULL, instr) >= 5 &&
+        instr_get_raw_byte(instr, 0) == EVEX_PREFIX_OPCODE) {
+        if (instr_get_isa_mode(instr) == DR_ISA_AMD64 ||
```
derekbruening, line 1353: What is this change?  Just collapsing the if's into &&?  But why does it stop here?

derekbruening (approved):

core/ir/x86/instr.c
derekbruening: If it's broken maybe just remove it.  For plain DR with no client this would be faster as then there is no decoding required beyond level 1 (just find the size) which is all DR does for non-"interesting" instructions -- but OTOH does core DR care about zmm-writing?  Is this function called by core DR at all?

clients/drcachesim/tests/offline-allasm-scattergather-basic-counts.templatex
abhinav92003: I put the exact counts for reads and stores. Sure, I'll put those for instrs as well.

There's a FIXME already on the load/store instrs; I'll modify it to include instr counts too.

clients/drcachesim/tests/scattergather.templatex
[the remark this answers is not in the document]
abhinav92003: Oh, I meant the bug is there as expected :) -- bad choice of words I guess. Based on the instrumented basic block I saw in https://github.com/DynamoRIO/dynamorio/pull/4863#issuecomment-824560991, I had expected that there's a bug, and that this count would be 1. 

There's already a FIXME in clients/drcachesim/tests/offline-allasm-scattergather-basic-counts.templatex.

core/ir/x86/instr.c
abhinav92003: Seemed slightly more readable to my eyes as the last one was an `||`, but sure I'll collapse this too.

core/ir/x86/instr.c
abhinav92003: Yeah, this is invoked in `client_process_bb`; DR tries to figure out whether there's some AVX instr at play. And if yes, it enables saving/restoring zmm and opmasks.

core/ir/x86/instr.c
abhinav92003: Done. Collapsing partially was quite arbitrary on my part. Maybe I didn't think this part through enough.

core/ir/x86/instr.c
abhinav92003: I'll still need to check if the raw bits check is just an optimisation or if it's functionally different.

abhinav92003:
`code_api|tool.drcacheoff.allasm-scattergather-basic-counts` is being flaky. Will need to figure out why before submitting.

core/ir/x86/instr.c
derekbruening: But `client_process_bb` is only for when a client is present, for which `full_decode` will be set and so this optimization will not be needed.

clients/drcachesim/tests/offline-allasm-scattergather-basic-counts.templatex
abhinav92003: Done.

core/ir/x86/instr.c
abhinav92003: I see. `full_decode` is set here if there are clients: https://github.com/DynamoRIO/dynamorio/blob/c6474fda6ad45c88944dbadd79f870ade46c7ee2/core/arch/interp.c#L5035. So decoded instr is available and can be used directly, instead of the brittle (and slightly complex) bytes check. I'll remove the bytes check.

abhinav92003:
I'm facing https://github.com/mxschmitt/action-tmate/issues/16 when using the tmate web shell. As discussed on the issue, it may be because of timeout. I'll perhaps need to open the web shell quickly after the test fails.

abhinav92003:
I was able to ssh into the Github Actions runner. Here's the issue, a SIGILL on some runners:

```
Program received signal SIGILL, Illegal instruction.
0x00000000310173a4 in ?? ()
(gdb) x/10i 0x00000000310173a4
=> 0x310173a4:  vpcmpequd %xmm12,%xmm10,%k2
   0x310173ab:  kmovw  %k2,%eax
   0x310173af:  cmp    $0x9,%eax
   0x310173b2:  mov    %gs:0x120,%rcx
   0x310173bb:  mov    (%rcx),%rcx
   0x310173be:  jrcxz  0x31017428
   0x310173c0:  movabs %rax,%gs:0x0
   0x310173cb:  movabs %gs:0x20,%rax
   0x310173d6:  mov    %rsp,0x18(%rax)
   0x310173da:  mov    0x948(%rax),%rsp
```

PR #4867, opened 2021-04-21
i#1807: Add flaky long suite tests to ignore list

abhinav92003:
Adds some known flaky tests to ignore list to make master-merge CI green.

Issue: #1807


abhinav92003:
There are some other tests, particularly the variants of `common.nativeexec_bindnow` and `common.decode-stress` that are seen to be failing, with slightly different options each time. Do we want to add those to ignore list already or wait for each different variant to fail multiple times?

derekbruening:
I would add them all as it seems unlikely that the different options are a factor.  Maybe this PR could add a regex match or post-pipe-character match that ignores pre-pipe chars to make that easier to do with a single entry.

Maybe file a separate issue on the MAX_NATIVE_RETSTACK assert?

abhinav92003:
> I would add them all as it seems unlikely that the different options are a factor. Maybe this PR could add a regex match or post-pipe-character match that ignores pre-pipe chars to make that easier to do with a single entry.

I added a quick change that does the latter. 
 
> Maybe file a separate issue on the MAX_NATIVE_RETSTACK assert?

#4868

abhinav92003:
I experimented a bit with making it a perl regex based check, but adding new tests to the ignore list become tricky with many escaped and special chars in each entry. Perhaps not worth playing with regexps if the current solution is acceptable.

abhinav92003:
PTAL. This is ready for review.

derekbruening (approved):

abhinav92003:
`code_api|win32.hookerfirst` started failing in vs2017-32 suddenly 

Adding it to ignore list for now.

Created #4872

derekbruening:
> `code_api|win32.hookerfirst` started failing in vs2017-32 suddenly
> 
> Adding it to ignore list for now.
> 
> Created #4872

Actually it's under #4870: showed up on some earlier PR's.  But it seems to suddenly be happening in every run: from some kind of VM update or sthg probably.

Issue #4848, opened 2021-04-14
AArch64 decoder: Add all GPR-touching instructions


derekbruening:
Splitting sub-pieces off from the master issue #2626 on finishing the AArch64 decoder.  This piece covers ensuring that all instructions with general purpose register operands are fully decoded and encoded.  Xref the instruction lists here: https://github.com/DynamoRIO/dynamorio/issues/2626#issuecomment-771605153

derekbruening:
I found it surprising how many GPR-touching instructions were/are missing from the decoder: those could cause subtle bugs and all kinds of application crashes or abnormalities.  I recall that the `xx` opcodes were supposed to decode GPR slots and try to include GPR registers they touch to avoid such bugs -- has that held for all these SIMD opcodes?

From https://github.com/DynamoRIO/dynamorio/pull/4855#discussion_r614953058
```
$ disasm_a64 1e290000
llvm-mc:   1e290000 fcvtpu w0, s0
capstone:  1e290000 fcvtpu w0, s0
bfd:       1e290000 fcvtpu w0, s0
DynamoRIO: 1e290000 xx $0x1e290000 %x0 %x0 %x0 %x9 -> %x0 %x0 %x0 %x9
$ disasm_a64 1e2903ff
llvm-mc:   1e2903ff fcvtpu wzr, s31
capstone:  1e2903ff fcvtpu wzr, s31
bfd:       1e2903ff fcvtpu wzr, s31
DynamoRIO: 1e2903ff xx $0x1e2903ff %sp %sp %x0 %x9 -> %sp %sp %x0 %x9
```

Here we have wzr/sp and w0 indicated as both read and written which does remove (most/all?) causes of bugs.

AssadHashmi:
Missing opcodes that touch GPR's: 75 of 254 as at 17-06-21.

PR #4869, opened 2021-04-22
i#4848 AArch64 GPR Decode: Add FCVTZS fixed-point

MDevereau:
Add handling of:
- FCVTZS (scalar, fixed-point)
- FCVTZS (vector, fixed-point)(scalar)

Issue: #2626, #4848


derekbruening (approved):

derekbruening:
Windows `win32.hookerfirst` failure seems like a flake but never seen that one before: filed #8470 and re-running.

PR #4875, opened 2021-04-25
i#1312: Adds instr_zeroes_zmmh()

johnfxgalea:
Adds instr_zeroes_zmmh(), which returns true if the instruction clears the upper bits of a ZMM register with zeros.

Updates instr_zeroes_ymmh() to take into account EVEX instructions.

Tests for the above are also included.

Issue: #1312

### DIFF
--- a/core/ir/instr_api.h
+++ b/core/ir/instr_api.h
@@ -1439,11 +1439,18 @@ DR_API
  * the top half while others zero it when writing to the bottom half).
  * This zeroing will occur even if \p instr is predicated (see instr_is_predicated()).
  */
-/* XXX i#1312: For AVX-512, we will want a instr_zeroes_zmmh function as well that also
- * includes the vzeroupper instruction.
- */
 bool
 instr_zeroes_ymmh(instr_t *instr);
+
+DR_API
+/**
+ * Returns true iff \p instr writes to an xmm or ymm register and zeroes the top half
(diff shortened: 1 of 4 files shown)


johnfxgalea:
@khuey Hoping you can take a look at your convenience, please? If you're busy, I perfectly understand and perhaps can try ask someone else to take a look at this PR.

core/ir/instr_api.h
```diff
@@ -1439,12 +1439,19 @@ DR_API
  * the top half while others zero it when writing to the bottom half).
  * This zeroing will occur even if \p instr is predicated (see instr_is_predicated()).
  */
-/* XXX i#1312: For AVX-512, we will want a instr_zeroes_zmmh function as well that also
- * includes the vzeroupper instruction.
```
derekbruening, line 1443: Could you confirm that vzeroupper returns true?

derekbruening (approved):

core/ir/instr_api.h
johnfxgalea: Nice catch! Turns out that the ir for vzeroupper and vzeroall do not include the simd registers as operands. This needs to be fixed, if possible. We need to include them even if there are implicit. Think push instruction which has an implicit operand related to the stack that our ir recognises.

johnfxgalea: Done.

Issue #4256, opened 2020-04-17
ASSERT "ma->start + ma->os_data.alignment == base"


derekbruening:
This happens reliably in a proprietary AArch64 app running on ThunderX2 on a 5.3.12-222-64kb kernel:

```
<CURIOSITY : ma->start + ma->os_data.alignment == base in file dynamorio/core/unix/os.c line 7720
version 7.91.18357, custom build
```

I have not looked into the details; just filing various issues seen with this app and then will go back through them.

Issue #4737, opened 2021-02-17
ELF loader ignores segment alignment


derekbruening:
The ELF loader in drlibc ignores the ElfNN_Phdr.p_align flag entirely: instead it uses the page size only for alignment of the start and end of each segment.  This doesn't matter for most ELF files, since they align their p_vaddr fields.  But it does appear when loading AArch64 ELF files on x86 where the ELF expects 64K pages and aligns its segments to 64K.  (This also raises the problem of how to handle cross-arch page sizes like that when we load cross-arch ELF files for drcachesim trace analysis and other purposes: right now using the host page size is working b/c we only run AArch64 on x86 where x86's page size is *smaller*.)

Getting into the details: it is a little murky as to which loaders honor p_align, and what p_align really means.  I found a case where an ELF file's segments had 2MB p_align, but the kernel there completely ignored the p_align (slightly custom Linux kernel).  DR not matching the kernel in this respect causes major problems when our early injection parses the libdynamorio.so headers for relocation.

egrimley-arm:
There may be a misunderstanding here, perhaps my misunderstanding. The man page says:

> `p_vaddr` should equal `p_offset`, modulo `p_align`.

So in the case of an `ET_EXEC` there's not much the loader can do with `p_align` other than check the condition and stop with an error if it is not satisfied. In practice the kernel seems to ignore `p_align` in an `ET_EXEC`.

However, with an `ET_DYN` the kernel does take account of `p_align` when picking the load address.

I checked this by building the following program with `gcc` on amd64 Ubuntu 22.04:
```
#include <stdio.h>
int main()
{
    printf("%p\n", main);
    return 0;
}
```

Because it's a recent system, the binary I get is an `ET_DYN`, not an `ET_EXEC` as I would get on an older system. The executable segment of the binary has `p_offset`, `p_vaddr` and `p_align` all equal to `0x1000`. When I run the program I get a random number equal to `0x149` modulo `0x1000`.

However, if I edit the binary so that `p_align` of that segment is `0x100000000` then when I run the program I get a random number equal to `0x1149` modulo `0x100000000`.

DynamoRIO's loader does indeed ignore the modified `p_align`.

PR #4876, opened 2021-04-26
i#4256: prevent triggered assertion for duplicated elf_headers on AArch64

sapostolakis:
Prevents triggering an ASSERT_CURIOSITY for duplicated elf headers for debug DR builds 
on AArch64 systems. It is possible for small modules (less than 1 page of disk space) to 
have multiple LOAD data segments. Since these segments are mapped from a single disk 
page they will all have an elf header. The triggered assertion assumed that there can only 
be one data segment. Extends this assertion check to accommodate for more than one 
data segments.

Fixes: #4256

### DIFF
--- a/core/unix/os.c
+++ b/core/unix/os.c
@@ -7834,13 +7834,28 @@ mmap_check_for_module_overlap(app_pc base, size_t size, bool readable, uint64 in
                  * Note, if it is a header of a different module, then we'll not have
                  * an overlap, so we will not hit this case.
                  */
-                ASSERT_CURIOSITY(
-                    ma->start + ma->os_data.alignment ==
-                    base
-                        /* On Mac we walk the dyld module list before the
-                         * address space, so we often hit modules we already
-                         * know about. */
-                        IF_MACOS(|| !dynamo_initialized && ma->start == base));
+#ifdef AARCH64
+                /* For AArch64 systems with 64K pages it is possible for small modules
+                 * (less than 1 page of disk space) to have multiple RW LOADable data
+                 * segments. Since these segments are mapped from a single disk space they
+                 * will all have a elf_header satisfying the check above.
+                 */
+                if (ma->os_data.alignment == 65536) {
+                    size_t offset = base - ma->start;
+                    size_t seg_id = offset / ma->os_data.alignment;
+                    ASSERT_CURIOSITY(seg_id < ma->os_data.num_segments &&
+                                     ma->os_data.segments[seg_id].start == base &&
+                                     ma->os_data.segments[seg_id].prot ==
+                                         (MEMPROT_READ | MEMPROT_WRITE));
+                } else
+#endif
+                    ASSERT_CURIOSITY(
+                        ma->start + ma->os_data.alignment ==
+                        base
+                            /* On Mac we walk the dyld module list before the
+                             * address space, so we often hit modules we already
+                             * know about. */
+                            IF_MACOS(|| !dynamo_initialized && ma->start == base));
             }
         });
     }
--- a/core/unix/os.c
+++ b/core/unix/os.c
@@ -7827,35 +7827,24 @@ mmap_check_for_module_overlap(app_pc base, size_t size, bool readable, uint64 in
                  * disk space, e.g. /usr/lib/httpd/modules/mod_auth_anon.so.  When
                  * such a module is mapped in, the os maps the same disk page twice,
                  * one readonly and one copy-on-write (see pg.  96, Sec 4.4 from
-                 * Linkers and Loaders by John R.  Levine).  This makes the data
-                 * section also satisfy the elf_header check above.  So, if the new
-                 * mmap overlaps an elf_area and it is also a header, then make sure
-                 * the previous page (correcting for alignment) is also a elf_header.
-                 * Note, if it is a header of a different module, then we'll not have
-                 * an overlap, so we will not hit this case.
+                 * Linkers and Loaders by John R.  Levine). It also possible for
+                 * such small modules to have multiple LOAD data segments. Since all
+                 * these segments are mapped from a single disk page they will all have a
+                 * elf_header satisfying the check above. So, if the new mmap overlaps an
+                 * elf_area and it is also a header, then make sure the current segment's
+                 * offset (from the beginning of the backing file) is within the page
+                 * size. Note, if it is a header of a different module, then we'll not
+                 * have an overlap, so we will not hit this case.
                  */
-#ifdef AARCH64
-                /* For AArch64 systems with 64K pages it is possible for small modules
-                 * (less than 1 page of disk space) to have multiple RW LOADable data
-                 * segments. Since these segments are mapped from a single disk space they
-                 * will all have a elf_header satisfying the check above.
-                 */
-                if (ma->os_data.alignment == 65536) {
-                    size_t offset = base - ma->start;
-                    size_t seg_id = offset / ma->os_data.alignment;
-                    ASSERT_CURIOSITY(seg_id < ma->os_data.num_segments &&
-                                     ma->os_data.segments[seg_id].start == base &&
-                                     ma->os_data.segments[seg_id].prot ==
-                                         (MEMPROT_READ | MEMPROT_WRITE));
-                } else
-#endif
-                    ASSERT_CURIOSITY(
-                        ma->start + ma->os_data.alignment ==
-                        base
-                            /* On Mac we walk the dyld module list before the
-                             * address space, so we often hit modules we already
-                             * know about. */
-                            IF_MACOS(|| !dynamo_initialized && ma->start == base));
+                size_t aligned_offset = base - ma->start;
+                size_t seg_id = aligned_offset / ma->os_data.alignment;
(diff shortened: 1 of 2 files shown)


sapostolakis:
Example of a small shared object that was triggering #4256, observed on a AArch64 system with 64K pages:
```
linux-pyb5:~ # readelf -l /usr/grte/v5/lib64/libdl-2.27.so

Elf file type is DYN (Shared object file)
Entry point 0xec0
There are 10 program headers, starting at offset 64

Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  PHDR           0x0000000000000040 0x0000000000000040 0x0000000000000040
                 0x0000000000000230 0x0000000000000230  R      0x8
  INTERP         0x0000000000000270 0x0000000000000270 0x0000000000000270
                 0x0000000000000029 0x0000000000000029  R      0x1
      [Requesting program interpreter: /usr/grte/v5/lib64/ld-linux-aarch64.so.1]
  LOAD           0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000001eb0 0x0000000000001eb0  R E    0x10000
  LOAD           0x0000000000001eb0 0x0000000000011eb0 0x0000000000011eb0
                 0x0000000000000310 0x0000000000000310  RW     0x10000
  LOAD           0x00000000000021c0 0x00000000000221c0 0x00000000000221c0
                 0x0000000000000008 0x0000000000000058  RW     0x10000
  DYNAMIC        0x0000000000001ed0 0x0000000000011ed0 0x0000000000011ed0
                 0x00000000000001e0 0x00000000000001e0  RW     0x8
  GNU_RELRO      0x0000000000001eb0 0x0000000000011eb0 0x0000000000011eb0
                 0x0000000000000310 0x0000000000001150  R      0x1
  GNU_EH_FRAME   0x0000000000000de8 0x0000000000000de8 0x0000000000000de8
                 0x000000000000002c 0x000000000000002c  R      0x4
  GNU_STACK      0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000000000 0x0000000000000000  RW     0x0
  NOTE           0x000000000000029c 0x000000000000029c 0x000000000000029c
                 0x0000000000000040 0x0000000000000040  R      0x4

 Section to Segment mapping:
  Segment Sections...
   00
   01     .interp
   02     .interp .note.ABI-tag .note.gnu.build-id .dynsym .gnu.version .gnu.version_d .gnu.version_r .gnu.hash .dynstr .rela.dyn .rela.plt .rodata .eh_frame_hdr .eh_frame .text .init .fini .plt
   03     .init_array .fini_array .dynamic .got .got.plt
   04     .data .bss
   05     .dynamic
   06     .init_array .fini_array .dynamic .got .got.plt
   07     .eh_frame_hdr
   08
   09     .note.ABI-tag .note.gnu.build-id
```

core/unix/os.c
```diff
@@ -7834,13 +7834,28 @@ mmap_check_for_module_overlap(app_pc base, size_t size, bool readable, uint64 in
...
-                         * address space, so we often hit modules we already
-                         * know about. */
-                        IF_MACOS(|| !dynamo_initialized && ma->start == base));
+#ifdef AARCH64
+                /* For AArch64 systems with 64K pages it is possible for small modules
+                 * (less than 1 page of disk space) to have multiple RW LOADable data
+                 * segments. Since these segments are mapped from a single disk space they
+                 * will all have a elf_header satisfying the check above.
+                 */
+                if (ma->os_data.alignment == 65536) {
```
derekbruening, line 7843: This could happen with 16K pages, or technically 4K, or other sizes: I don't think we should hardcode a particular size.

core/unix/os.c
```diff
@@ -7834,13 +7834,28 @@ mmap_check_for_module_overlap(app_pc base, size_t size, bool readable, uint64 in
...
+                    size_t offset = base - ma->start;
+                    size_t seg_id = offset / ma->os_data.alignment;
```
derekbruening, line 7845: Some ELF files have different alignments for different segments.  I don't think we can assume they are all identical.

core/unix/os.c
```diff
@@ -7834,13 +7834,28 @@ mmap_check_for_module_overlap(app_pc base, size_t size, bool readable, uint64 in
...
+                    ASSERT_CURIOSITY(seg_id < ma->os_data.num_segments &&
+                                     ma->os_data.segments[seg_id].start == base &&
+                                     ma->os_data.segments[seg_id].prot ==
```
derekbruening, line 7848: I think what we want is to walk the segment array until we hit one whose start==base and check that `module_segment_data_t.offset < PAGE_SIZE` for every one through that one.

core/unix/os.c
```diff
@@ -7834,13 +7834,28 @@ mmap_check_for_module_overlap(app_pc base, size_t size, bool readable, uint64 in
...
+                                         (MEMPROT_READ | MEMPROT_WRITE));
+                } else
+#endif
+                    ASSERT_CURIOSITY(
+                        ma->start + ma->os_data.alignment ==
+                        base
```
derekbruening, line 7854: The new check would replace the old one as a more general version, right?  I guess the Mac exception needs to be kept.

derekbruening (changes requested):

core/unix/os.c
sapostolakis: Sure, better not to hardcode since it could happen with other page sizes.

core/unix/os.c
sapostolakis: From my understanding, for every module only one alignment is possible (i.e., `module_area_t.os_module_data_t.alignment`). So, the alignment between all LOAD segments should be the same. Other types of segments can surely have other alignments but these segments are not included in `module_area_t.os_module_data.segments`.

core/unix/os.c
sapostolakis: Yep, that's a better check and what I should have done in the first place. For traversing the array maybe we can avoid it  based on the assumption that the alignment between the segments in the segment array seems to be constant. If my assumption is wrong or checking all the predecessors of the current segment is necessary (the predecessors will be checked on their own when they are mapped to memory and I would think that checking the offset of the current segment would suffice), I will switch to just traversing and checking all of them.

core/unix/os.c
sapostolakis: Yes, we can just keep only one more general check.

core/unix/os.c
derekbruening: > From my understanding, for every module only one alignment is possible

I don't think this is the case: the ELF spec does not say the every segment must have the same alignment.

sapostolakis: Yep not for every segment but for every segment in the segment array in `module_area_t.os_module_data.segments ` only one alignment seems to be possible. Isn't that the case? If it's not the case, should't alignment be part of the `module_segment_t`. My point is based on the DR codebase rather than pointing to an ELF specification. Based on the spec, every segment can have its own alignment as specified in the program header. But it seems that LOAD segments seem to always have the same alignment (in most cases at least). Having LOAD segments with different alignments seem to be causing issues in some settings (e.g., https://github.com/numpy/numpy/issues/16677).

sapostolakis: Maybe we can add a comment in the code of this PR that mentions the assumption that was made based on the current DR code (in case `os_module_data_t` changes) or change the code and move the alignment inside `module_segment_t` so that every segment can have its own alignment (and make no assumptions here about a unified alignment for the segments in the array). As a third option, we can traverse here all segments without alignment assumptions and ignore the assumptions made elsewhere in the DR code.

derekbruening: The computation "seg_id = offset / ma->os_data.alignment" has more problems than just assuming constant alignment: segments are not always adjacent.  There are sometimes gaps in between.  Taking a later segment's base and subtracting the first segment's base does not tell you how many segments are in between.  I think the check here should just do a traversal and abandon the idea of trying to compute the index from the address.  It'll be inside a DOCHECK or whatnot so it won't affect release build performance.

sapostolakis: Great point. I completely missed the scenario of non-adjacent segments. There is also a field `contiguous` in the `os_module_data_t` struct which is meant for this scenario. So, I will do the traversal instead.

core/unix/os.c
sapostolakis: Given the scenario of non-adjustment segment pointed it out in another comment, it is better to do the traversal instead of computing the index of the segment directly from the base address.

core/unix/os.c
derekbruening: Re: DR module code assuming single segment alignment: it does seem like that is the case.  Not ideal since it doesn't seem hard to handle different alignments, but I guess we've lived with it all this time.  There are other areas where DR doesn't do the right thing wrt segment alignment: #4737.

sapostolakis: Should we create a new issue explicitly reporting this single segment alignment assumption, or add some comment in the code to point it out?

derekbruening: However you want to do it -- I see a curiosity assert in module_add_segment_data so we would find out violations in debug build at least.

sapostolakis: I added a comment in the code. Good thing that there is also a check in module_add_segment_data.

derekbruening (approved):

Issue #4878, opened 2021-04-26
aarch64 clean calls over-write whereami field, clobbering signals_pending field


derekbruening:
This surfaces as an assert `info->interrupted == NULL || info->interrupted == f` (seen also in #2328 and #4670 but with different root causes) due to signals_pending suddenly becoming 0 and DR failing to relink a fragment (and failing to deliver a signal):

```
disp signals_pending = 0

d_r_dispatch: target = 0x0000ffffbb1d29bc
Entry into F100338(0x0000ffffbb1d29bc).0x0000aaaa8ce6d740 (shared)

master_signal_handler: thread=976210, sig=26, xsp=0x0000fffd72942da0, retaddr=0x000000000000001a
received alarm 1 @0x0000aaaa8a75bfdc
        dr value is now 0
        master_signal_handler 26 returning now to 0x0000aaaa8a75bfdc


master_signal_handler: thread=976210, sig=27, xsp=0x0000fffd72942da0, retaddr=0x000000000000001b
received alarm 2 @0x0000aaaa8a75bfdc
        app value is now 0
record_pending_signal(27) from gen routine or stub 0x0000aaaa8a75bfdc
        unlinking outgoing for interrupted F2430
        action is not SIG_IGN
3rd pending alarm 27 => dropping 2nd
signals_pending = 1
        master_signal_handler 27 returning now to 0x0000aaaa8a75bfdc


master_signal_handler: thread=976210, sig=27, xsp=0x0000fffd72942da0, retaddr=0x000000000000001b
received alarm 2 @0x0000aaaad3de3388
        app value is now 0
record_pending_signal(27) from DR at pc 0x0000aaaad3de3388
        action is not SIG_IGN
3rd pending alarm 27 => dropping 2nd
signals_pending = 1
        master_signal_handler 27 returning now to 0x0000aaaad3de3388


master_signal_handler: thread=976210, sig=26, xsp=0x0000fffd72942da0, retaddr=0x000000000000001a
received alarm 1 @0x0000aaaad39322e4
        dr value is now 0
        master_signal_handler 26 returning now to 0x0000aaaad39322e4

Exit from F100337(0x0000ffffbb1d29a4).0x0000aaaa8ce6d708 (shared)       
 (block ends with syscall)
Entry into do_syscall to execute a non-ignorable system call
system call 63
Exit from system call
post syscall: sysnum=0x000000000000003f, result=0x0000000000001000 (4096)
disp signals_pending = 0

[...]

PR #4879, opened 2021-04-27
i#4878 signals_pending: Clean up bool-vs-int usage

derekbruening:
Fixes two legacy stores treating signals_pending as a boolean,
discovered while debugging the signals_pending corruption.

Issue: #4878

### DIFF
--- a/core/dispatch.c
+++ b/core/dispatch.c
@@ -714,7 +714,7 @@ dispatch_enter_native(dcontext_t *dcontext)
          * be an alarm signal so we drop it as the simplest solution.
          */
         ASSERT(dcontext->signals_pending);
-        dcontext->signals_pending = false;
+        dcontext->signals_pending = 0;
     } while (true);
 #else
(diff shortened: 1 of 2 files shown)


core/dispatch.c
```diff
@@ -714,7 +714,7 @@ dispatch_enter_native(dcontext_t *dcontext)
          * be an alarm signal so we drop it as the simplest solution.
          */
         ASSERT(dcontext->signals_pending);
```
abhinav92003, line 716: I wonder if usages like this need to be audited too. `signals_pending` can be < 0 if one is being currently handled.

abhinav92003:
Thanks for the cleanup! It was on my todo list but I never got to it.

abhinav92003 (approved):

core/dispatch.c
derekbruening: Done.  This one is >0; all others are !=0.

Issue #4849, opened 2021-04-14
AArch64 decoder: Add all condition code-touching instructions


derekbruening:
Splitting sub-pieces off from the master issue #2626 on finishing the AArch64 decoder.  This piece covers ensuring that all instructions that either read or write the condition codes (NZCVQGE) are fully decoded and encoded.  Xref the instruction lists here: https://github.com/DynamoRIO/dynamorio/issues/2626#issuecomment-771605153.  Xref how missing cases here caused hard-to-debug problems in #4687 and the workaround we put in place which assumes all unknown opcodes both read and write the codes, which adds overhead.  We would like to remove the workaround once we have confidence that all the cases are handled.

AssadHashmi:
Missing opcodes that touch condition codes: 9 of 30 as at 17-06-21.

abhinav92003:
Hit this issue while working on PR #4917. Cannot use the undefined opcode to trigger a fault to verify proper restoration of aflags native val on fault for AArch64. This is because aflags are restored automatically before the undefined opcode.

Issue #4847, opened 2021-04-14
AArch64 decoder: Add all memory-touching instructions 


derekbruening:
Splitting sub-pieces off from the master issue #2626 on finishing the AArch64 decoder.  This piece covers ensuring that all instructions with memory operands are fully decoded and encoded.  Xref the instruction lists here: https://github.com/DynamoRIO/dynamorio/issues/2626#issuecomment-771605153

AssadHashmi:
Hi @derekbruening, do you have any specific instances of memory instructions which are/were failing?

derekbruening:
I do not know of any cases where a memory operand was missed (there are known cases of the decoder decoding the wrong address into the IR and clients having to take extra steps to get the right address: #4400), but it is the kind of thing that would go silently missing and would not be noticed up front, leading to unknowingly inaccurate traces.  Hence we're relying on whatever validation steps can be taken over the decoder.

derekbruening:
If more might be missing, the `DC ZVA` case may point to extra inspection warranted in areas of the ISA where loads or stores might be lurking but disguised as sub-opcodes of system or other instructions.

AssadHashmi:
> Hence we're relying on whatever validation steps can be taken over the decoder.

We're identifying gaps in our test coverage of memory instructions and will shortly be working on PRs to test and fix failures.

AssadHashmi:
Missing opcodes that touch memory: 0 of 48 as at 17-06-21.

derekbruening:
Are `adr` and `adrp` supported?  The tests for them are still disabled in ir_aarch64.c:
```C
#if 0 /* TODO i#4847: add memory touching instructions */
    adr(dcontext);
    adrp(dcontext);
#endif
```

derekbruening:
Also:
```
#if 0 /* TODO i#4847: address memory touching instructions that fail to encode */
        ldr_base_literal(dc);
#endif
```

For #3995 we will use `opnd_create_rel_addr` for the drbbdup case encoding: I'm worried it's not going to encode, given the above disabled test for this operand type.

derekbruening:
We want to have the encoder handle `ABS_ADDR_kind` as well by treating it like `REL_ADDR_kind`: #5295

derekbruening:
> We want to have the encoder handle `ABS_ADDR_kind` as well by treating it like `REL_ADDR_kind`: #5295

[...]

PR #4882, opened 2021-04-27
i#2626 AArch64 v8.0 Decode: Order dis-a64.txt alphabetically

MDevereau:
Ordering dis-a64.txt alphabetically by instruction name will make it
much easier to navigate and detect missing or erroneous tests.

Issue: #2626, #4847, #4848, #4849


MDevereau:
This was done using a python script. Another python script to strip empty lines and comments and count the number of lines remaining was compared to `cat dis-a64.txt | grep -v ^# | grep -v ^\s\*$ | wc -l` both before and after the reformatting to verify that no tests were lost.

AssadHashmi (approved):

Issue #4881, opened 2021-04-27
drcachesim LRU implementation fails to increment counters properly


EagleEyeKestreI:
**Describe the bug**
Hi all,
I read https://github.com/DynamoRIO/dynamorio/blob/master/clients/drcachesim/simulator/cache_lru.cpp , and find LRU implement here doesn't match my knowledge...

Consider this scene, there is a four-associativity cache line, at first the counters of four lines are all 0: (the number following invalid means variable counter_
invalid 0
invalid 0
invalid 0
invalid 0
Then, cache line a enters, according to the logic in cache_lru.cpp, then it becomes:
a 0
invalid 1
invalid 1
invalid 1
Then cache line b enters:
a 1
b 0
invalid 2
invalid 2
Then comes a again, resulting a hit: (invalid counter is still 2 because https://github.com/DynamoRIO/dynamorio/blob/master/clients/drcachesim/simulator/cache_lru.cpp#L49 says if <= cnt then counter++)
a 0
b 1
invalid 2
invalid 2
Then comes two other lines c and d, then the set becomes:
a 2
b 2
c 1
d 0
Then comes line e, it will replace line a according code logic(a and b counter are both 2, and a line index is before b). But the last visit of line a is after line b in this scene, which obeys LRU

is this a bug or there's something I get wrong... Thx very much !

**To Reproduce**
pass

**Expected behavior**
I think in such case b should be replaced?

**Screenshots or Pasted Text**
pass

**Versions**
 - What version of DynamoRIO are you using? newest
 - Does the latest build from https://github.com/DynamoRIO/dynamorio/releases solve the problem? no
- What operating system version are you running on? ("Windows 10" is *not* sufficient: give the release number.) doesn't matter, macOD
- Is your application 32-bit or 64-bit? 64-bit

derekbruening:
Please use the users list for questions: the issue tracker is for known bugs.  It looks like this is a duplicate ofthe users list discussion at https://groups.google.com/g/dynamorio-users/c/AMxxnudnBy0/m/ceESLSf3AQAJ.  It does look like from the discussion there that there is a bug so this can become the issue for that.

EagleEyeKestrel:
Yep the other issue raised in google groups is also me. I have opened a simple pull request for this LRU counter bug. However I don't have experience in writing unit test... Sorry for that.

Issue #3213, opened 2018-10-09
win32.setthreadcontext failed on Appveyor


derekbruening:
https://ci.appveyor.com/project/DynamoRIO/dynamorio/builds/19350530
```
debug-internal-32: 250 tests passed, **** 2 tests failed, but ignoring 1 for i#2145: ****
	(ignore: i#2145) 	code_api|client.pcache-use 
	code_api|win32.setthreadcontext 
```

derekbruening:
Crashed on GA CI:

https://github.com/DynamoRIO/dynamorio/pull/4729/checks?check_run_id=1880901026

```
2021-02-11T16:39:02.9545982Z 59: Test command: D:\a\dynamorio\dynamorio\build_debug-internal-32\bin32\drrun.exe "-s" "90" "-quiet" "-debug" "-use_dll" "D:/a/dynamorio/dynamorio/build_debug-internal-32/lib32/debug/dynamorio.dll" "-exit0" "-stderr_mask" "0xC" "-msgbox_mask" "0" "-dumpcore_mask" "0x7d" "-staged" "-code_api" "--" "D:/a/dynamorio/dynamorio/build_debug-internal-32/suite/tests/bin/win32.setthreadcontext.exe"
2021-02-11T16:39:02.9547826Z 59: Test timeout computed to be: 600
2021-02-11T16:39:03.5463351Z 58: &next_instr recorded
2021-02-11T16:39:03.5672645Z 59: &next_instr recorded
2021-02-11T16:39:03.5790883Z 58: starting thread...
2021-02-11T16:39:03.5791684Z 58: thread suspended.
2021-02-11T16:39:03.5792357Z 59: starting thread...
2021-02-11T16:39:03.5792876Z 59: thread suspended.
2021-02-11T16:39:03.5838040Z 59: control has been redirected.
2021-02-11T16:39:03.6100799Z 59: thread suspended.
2021-02-11T16:39:03.6116055Z 59: ecx was set, exiting
2021-02-11T16:39:03.6121822Z 59: <Early thread failure, no dcontext
2021-02-11T16:39:03.6122682Z 59: >
2021-02-11T16:39:03.6317485Z 59: <Application D:\a\dynamorio\dynamorio\build_debug-internal-32\suite\tests\bin\win32.setthreadcontext.exe (5464).  DynamoRIO internal crash at PC 0x7267f3b3.  Please report this at http://dynamorio.org/issues/.  Program aborted.
[...]

PR #4883, opened 2021-04-27
i#4881: Fix LRU counter bug in drcachesim

EagleEyeKestrel:


EagleEyeKestrel:
@derekbruening Simple fix, expect your review, thanks

clients/drcachesim/simulator/cache_lru.cpp
```diff
@@ -69,7 +69,5 @@ cache_lru_t::replace_which_way(int line_idx)
             max_way = way;
         }
     }
-    // Set to non-zero for later access_update optimization on repeated access
-    get_caching_device_block(line_idx, max_way).counter_ = 1;
```
derekbruening, line 73: See my analysis at https://github.com/DynamoRIO/dynamorio/issues/4881#issuecomment-827660933: removing this line will cause the ==0 short-circuit to always be executed, resulting in all the counters permanently staying at 0, right?

I would vote for my suggestion in the linked comment of initializing the counters to 0..assoc-1.

EagleEyeKestreI: All the counters staying at 0? Why? Considering a 4-line set. When a line comes in, all counters are 0, 0 <= 0, every counter will ++, it will become:

- a 0
- invalid 1
- invalid 1
- invalid 1

You can imagine these lines to be like a linkded-list. When visiting a node, (it will be removed from the middle and inserted at head), every node before it will be put behind a position(index++, if it has a index)(similar to all counter++ which has value <= cnt). And then the new head index becomes 0. Invalid lines counter will always be the max value, we can ignore it.

So where do I make a mistake? In my opinion any initializing value is OK... Could you give me a counter-example?

EagleEyeKestrel: All the counters staying at 0? Why? Considering a 4-line set. When a line comes in, all counters are 0, 0 <= 0, every counter will ++, it will become:

- a 0
- invalid 1
- invalid 1
- invalid 1

You can imagine these lines to be like a linkded-list. When visiting a node, (it will be removed from the middle and inserted at head), every node before it will be put behind a position(index++, if it has a index)(similar to all counter++ which has value <= cnt). And then the new head index becomes 0. Invalid lines counter will always be the max value, we can ignore it.

So where do I make a mistake? In my opinion any initializing value is OK... Could you give me a counter-example?

derekbruening: See cache_lru.cpp lines 44-46:
```C
    // Optimization: return early if it is a repeated access.
    if (cnt == 0)
        return;
```
Counters are all initialized to 0 and this `cnt == 0` branch will be taken every time.  Thus no counter will ever be incremented.

EagleEyeKestrel: oh my fault, I ignore this . Sorry

EagleEyeKestrel: Can we set the initial counter value to 1? I think it can also work. Changing it to 0,1,..associativity - 1 looks a bit complex by changing code.

derekbruening: > Can we set the initial counter value to 1? I think it can also work. 

Intuitively the code seems clearest to me to start out in the steady state of 0..assoc-1 and then there is no separate bootstrapping stage and it is just easier to think about.

> Changing it to 0,1,..associativity - 1 looks a bit complex by changing code.

It should be the same amount of code as changing it to 1: just override `init` and have a very simple loop there.  E.g., see `cache_fifo_t::init`.

clients/drcachesim/simulator/caching_device_block.h
```diff
@@ -53,7 +53,7 @@ class caching_device_block_t {
     // where for the current replacement code we also set the counter at that time.
     caching_device_block_t()
         : tag_(TAG_INVALID)
-        , counter_(0)
+        , counter_(1)
```
derekbruening, line 56: This changes the initial value for all cache types, though, and could break some other type.  Best to only change for LRU.

clients/drcachesim/simulator/cache_lru.cpp
EagleEyeKestrel: Yep. Thanks for guidance. I will push new code later.

clients/drcachesim/simulator/caching_device_block.h
EagleEyeKestrel: Yes. I didn't notice this is a basic class. I have pushed new code.

derekbruening (approved):
Thank you for the patch!  We'll leave a unit test as separate work once infrastructure is in place under #4842.

clients/drcachesim/simulator/cache_lru.cpp
```diff
@@ -37,6 +37,31 @@
 // The count value 0 means the most recent access, and the cache line with the
 // highest counter value will be picked for replacement in replace_which_way.
 
+bool
+cache_lru_t::init(int associativity, int block_size, int total_size,
+                   caching_device_t *parent, caching_device_stats_t *stats,
```
derekbruening, line 42: clang-format test failed: there is an extra space on these lines here

derekbruening: I removed the spaces.

derekbruening:
The behavior change slightly changes the miss counts and breaks the two tests `tool.drcacheoff.legacy` and `tool.drcacheoff.snappy`.  The expected output needs to be updated.

derekbruening:
run arm tests

derekbruening:
> The behavior change slightly changes the miss counts and breaks the two tests `tool.drcacheoff.legacy` and `tool.drcacheoff.snappy`. The expected output needs to be updated.

@EagleEyeKestrel if you could update clients/drcachesim/tests/offline-legacy.templatex and clients/drcachesim/tests/offline-snappy-serial.templatex so that they pass -- easiest by running locally, or look at the log from the failure link above.

EagleEyeKestrel:
> > The behavior change slightly changes the miss counts and breaks the two tests `tool.drcacheoff.legacy` and `tool.drcacheoff.snappy`. The expected output needs to be updated.
> 
> @EagleEyeKestrel if you could update clients/drcachesim/tests/offline-legacy.templatex and clients/drcachesim/tests/offline-snappy-serial.templatex so that they pass -- easiest by running locally, or look at the log from the failure link above.

let me try

clients/drcachesim/tests/offline-snappy-serial.templatex
```diff
@@ -1,5 +1,5 @@
 Cache simulation results:
 .*
 LL stats:
-    Hits:                           58,699
+    Hits:                           78,312
```
derekbruening, line 4: (Wow this is a non-trivial change.  The other "legacy" test had very minor changes.  Wondering if this LRU issue caused other noticeable result differences on other workloads.)

derekbruening (approved):

derekbruening:
run arm tests

clients/drcachesim/tests/offline-snappy-serial.templatex
EagleEyeKestrel: oh it failed again

derekbruening: Actually I think that is an unrelated flaky test...checking

derekbruening: It is #3213 which unfortunately has happened before but nobody has reproduced locally.

derekbruening:
As noted in the thread the failure is win32.setthreadcontext #3213.

derekbruening:
Thanks again for finding and fixing this bug!  Adding better unit tests (#4842) should be a higher priority to catch things like this.

clients/drcachesim/tests/offline-snappy-serial.templatex
EagleEyeKestrel: Read log and see some timeout.

derekbruening: > Read log and see some timeout.

If you could paste a link to what you are seeing?  Only the Windows test was red.

Issue #4884, opened 2021-04-28
Asynch signal arriving in ni-syscall-ending block results in premature and duplicated pre-syscall-handler execution


derekbruening:
If a signal arrives in a block that ends in a non-ignorable syscall, we'll unlink it, and when back in dispatch we check for the syscall *before* the signal.  We do the pre-syscall handling, then try to run the syscall: but enter_fcache sees signals_pending and bails.  We then go run the signal handler, but we never undo the pre-syscall handling.  Once the handler returns we'll go execute the syscall for real, duplicating the pre-syscall handling.

Depending on the syscall, this may not matter, but some syscalls are hard to undo and premature execution and/or duplication might result in issues.

```
Exit from F91(0x0000ffff807775e0).0x0000aaaa75281e70 (shared)   
 (interrupted by delayable signal)
Entry into do_syscall to execute a non-ignorable system call
system call 98
Signal arrived in DR: aborting syscall enter; interrupted 0x0000ffff80777600
receive_pending_signal
        receiving signal 27

<run handler>

receive_pending_signal
        clearing signals_pending flag

d_r_dispatch: target = 0x0000aaaad5b4ac74
Entry into F241(0x0000aaaad5b4ac74).0x0000aaaa755acee0 (shared)
        
Exit from F298(0x0000aaaabe514750).0x0000aaaa755b2258 (shared)  
 (block ends with syscall)
Entry into do_syscall to execute a non-ignorable system call
system call 139
rt_sigreturn()

set next tag to 0x0000ffff80777600, sc->SC_XIP to 0x0000aaaa75276c80
for sigreturn, set sys_param1 to 0x0000aaaad6b4cd40
Exit from asynch event
post-sigreturn: setting xax/r0 to 0x0000aaaad6b4cd40, asynch_target=0x0000ffff80777600
post syscall: sysnum=0x000000000000008b, result=0x0000aaaad6b4cd40 (-692794048)
finished handling system call

d_r_dispatch: target = 0x0000ffff80777600
Entry into F1(0x0000ffff80777600).0x0000aaaa7527a00c (shared)
        
Exit from F1(0x0000ffff80777600).0x0000aaaa7527a0f4 (shared)    
 (block ends with syscall)
Entry into do_syscall to execute a non-ignorable system call
system call 98
Exit from system call
post syscall: sysnum=0x0000000000000062, result=0xffffffffffffff92 (-110)
finished handling system call
```

[...]

PR #4886, opened 2021-04-28
i#4884: Check signal exit before syscall, etc. reasons

derekbruening:
Adds a pending-signal check to avoid attempting to run a syscall for a
signal arriving in the cache, avoiding running pre-syscall handling
only to abort and have complications with undoing or duplicating
actions.

For a signal arriving in DR while executing a syscall, changes DR to
go ahead and do the syscall for all non-ignorable syscalls, again
avoiding complications with undoing or duplicating pre-syscall action.
Previously, this was only done for sigreturn or clone syscalls.
Non-ignorable syscalls generally do not block, so this should not
impact signal delivery delays.

Moves the pending-signal check to be the first thing checked for the
exit reason, for proper exit attribution.

Avoids the IBT add code for a pending-signal exit.

Tested on the large proprietary app where the wrong exit reasons were
first observed.  Adding a regression suite test for these asynchronous
signal conditions is not easy unfortunately.

Issue: #4878, #4884
Fixes: #4884

### DIFF
--- a/core/dispatch.c
+++ b/core/dispatch.c
@@ -908,7 +908,7 @@ dispatch_enter_dynamorio(dcontext_t *dcontext)
          */
         if (exited_due_to_ni_syscall(dcontext) ||
             instrument_invoke_another_syscall(dcontext)) {
-            if (dcontext->signals_pending > 0) {
+            if (IF_UNIX_ELSE(dcontext->signals_pending > 0, false)) {
                 /* Avoid running the pre-handler and aborting the fcache_enter w/o
                  * a good way to undo the pre-handler.
                  */
--- a/core/dispatch.c
+++ b/core/dispatch.c
@@ -723,6 +723,24 @@ dispatch_enter_native(dcontext_t *dcontext)
 }
 
 static void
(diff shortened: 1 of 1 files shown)


derekbruening:
I'm wondering if there is a way to make a test: not sure what it would do.  mprotect or other DR-acted syscalls in a loop with alarm signals and...just make sure no asserts??

abhinav92003:
> I'm wondering if there is a way to make a test: not sure what it would do. mprotect or other DR-acted syscalls in a loop with alarm signals and...just make sure no asserts??

Does `linux.signal_racesys` do just that?

abhinav92003 (approved):

derekbruening:
> > I'm wondering if there is a way to make a test: not sure what it would do. mprotect or other DR-acted syscalls in a loop with alarm signals and...just make sure no asserts??
> 
> Does `linux.signal_racesys` do just that?

Hmm, its syscall is just a sleep though: DR doesn't have to take any action there.  It does cover the code path for parts of the fixes here though which is good: so maybe that's enough combined with the proprietary app manual check of the logs of signal interruptions.

Issue #4885, opened 2021-04-28
note_base uninitialised in drx and drmgr


abhinav92003:
There's a race in initialising `note_base` and `note_base_emul` respectively in `drx` and `drmgr`. Setting default value to -1 and adding an assert reveals this.

```
test 198
    Start 198: code_api|client.drx-test

198: Test command: /usr/local/google/home/sharmaabhinav/dr/build/iX-note-base-race/bin64/drrun "-s" "90" "-quiet" "-debug" "-killpg" "-stderr_mask" "0xC" "-dumpcore_mask" "0" "-code_api" "-c" "/usr/local/google/home/sharmaabhinav/dr/build/iX-note-base-race/suite/tests/bin/libclient.drx-test.dll.so" "--" "/usr/local/google/home/sharmaabhinav/dr/build/iX-note-base-race/suite/tests/bin/client.drx-test"
198: Test timeout computed to be: 1500
198: ASSERT FAILURE: /usr/local/google/home/sharmaabhinav/dr/src/iX-note-base-race/ext/drx/drx.c:92: note_base > -1 (note_base not initialized)
```

derekbruening:
Where is the bug?  Is your client not calling drx_init()?

abhinav92003:
Actually, my bad. My assert statement was wrong  -- asserting `> -1` for a `uint`. I don't see any init issue in `drx` or `drmgr` now.

 Though I haven't seen one happening, but I believe a race is still possible. In `drx_init`, threads after the first one return early, and the first thread does all the init work: https://github.com/DynamoRIO/dynamorio/blob/403122fa9a57964f19b991928ddffb4cc08c97d9/ext/drx/drx.c#L147. If the first thread is suspended by the OS and some other thread tries to get a note val, it may find an uninitialised `note_base`, right?

derekbruening:
> If the first thread is suspended by the OS and some other thread tries to get a note val, it may find an uninitialised `note_base`, right?

You mean, drx_init is not called at single-threaded process init but rather later during multi-thread execution?  That would be a general problem with many initializations throughout DR.  You could make it explicit in the docs to require it during process init.

abhinav92003:
As being discussed in PR #4887, some user called APIs can have some assertions on whether static state has been initialised.

And I'll check the state of documentation for restrictions on when the init methods should be called.

Issue #1163, opened 2014-11-28
ASSERT (DrMem on chromium unit_tests on win8) ntdll.c:1311 first_slot > 0 && last_slot_open_end <= bitmap_size (out of TLS slots)


derekbruening:
_From [bruen...@google.com](https://code.google.com/u/109494838902877177630/) on May 20, 2013 16:05:12_

Dr. Memory runs out of TLS slots on chromium unit_tests on win8.  The fix will likely involve a DR change so I'm filing it in this tracker.

# /usr/bin/time ~/drmemory/git/build_x86_dbg/bin/drmemory.exe -dr_debug -dr_ops "-stderr_mask 15" -light -suppress suppressions.txt -pause_at_assert -batch -dr c:/derek/dr/git/exports -- ./unit_tests.exe --gtest_filter="CancelableTaskTrackerDeathTest.PostFromDifferentThread"

-leaks_only -no_count_leaks works.
-unaddr_only works.

release DR => APP CRASH
debug DR => assert

 # ChildEBP RetAddr  
00 0018e57c 72ed0020 ntdll!ZwRaiseHardError+0xc
01 0018e5d8 72e98eeb dynamorio!nt_messagebox+0x130 [c:\derek\dr\git\src\core\win32\ntdll.c @ 3492]
02 0018e600 72d0281c dynamorio!debugbox+0x4b [c:\derek\dr\git\src\core\win32\os.c @ 4573]
03 0018ee18 72d0226b dynamorio!notify+0x1dc [c:\derek\dr\git\src\core\utils.c @ 1945]
04 0018eebc 72cfe943 dynamorio!report_dynamorio_problem+0x4fb [c:\derek\dr\git\src\core\utils.c @ 2204]
05 0018eff0 72ed72b0 dynamorio!internal_error+0x133 [c:\derek\dr\git\src\core\utils.c @ 183]
06 0018f018 72ed7a89 dynamorio!tls_mark_taken_block_sequence+0x70 [c:\derek\dr\git\src\core\win32\ntdll.c @ 1311]
07 0018f0a4 72ed00f4 dynamorio!tls_alloc_helper+0x769 [c:\derek\dr\git\src\core\win32\ntdll.c @ 1493]
08 0018f0c4 72e953af dynamorio!tls_calloc+0x54 [c:\derek\dr\git\src\core\win32\ntdll.c @ 1559]
[...]

PR #4887, opened 2021-04-29
i#3823 drreg: Avoid GPR spill slot conflicts with earlier phases

abhinav92003:
Adds a stricter free spill slot check in drreg. This is needed when there are
multiple instrumentation passes that use drreg to spill reg values to slots.
In this case, we need to avoid spill slots that are in-use by prior instrumentation
passes. The added logic looks at the remaining instrs in the given ilist after
'where'. Any spill slot usage after 'where' must have been added by a prior
instrumentation pass and not the current one. To avoid spill slot conflicts,
we should avoid using such spill slots with pending usage. This is not completely
optimal as a spill slot may be available for use at some point in the ilist
even if it is used later in the ilist. But we live with this sub-optimality.

Adds new drreg_bb_properties_t (DRREG_HANDLE_MULTI_PHASE_SLOT_RESERVATIONS)
that enables the stricter free spill slot check. This is currently set by
drx_expand_scatter_gather.

Uses data area of label instrs marked with a special note to store information
about spill slot usage, which is used in the new stricter free spill slot check.

Adds documentation about the new option, highlighting that it's necessary to
enable this for multi-phase drreg reg reservations to work as intended, and
that it is not enabled by default due to performance cost.

Delays resetting bb_props to late in the instru2instru stage, in case stages
besides insertion need to read it.

Extends drreg-test to verify that spill slots used in later phases do not
conflict with those from earlier phases.

Issue: #3823


ext/drreg/drreg.c
```diff
@@ -173,15 +184,58 @@ get_where_app_pc(instr_t *where)
...
+        /* We store data about spill slot usage in the data area of a label instr
+         * immediately preceding the usage.
+         */
+        if (instr_is_label(in) &&
+            instr_get_note(in) == (void *)get_drreg_note_val(DRREG_NOTE_SPILL_SLOT_ID)) {
+            dr_instr_label_data_t *data = instr_get_label_data_area(in);
+            ASSERT(data != NULL, "failed to find label's data area");
+            /* Try to continue in release build. */
+            if (data != NULL &&
+                data->data[SPILL_SLOT_ID_LABEL_INSTR_DATA_AREA_OFFSET] ==
```
abhinav92003, line 219: I found a bit late that the existing `drreg_is_instr_spill_or_restore`[1] may be useful for doing this check. Though I don't think using `drreg_is_instr_spill_or_restore` is ideal -- it documents that it "may not be able to distinguish a drreg spill or restore from some other spill or restore", and also there may be more overhead in looking at each instr using `drreg_is_instr_spill_or_restore` vs just looking at label instrs with more readily available spill slot usage data.

[1]: https://github.com/DynamoRIO/dynamorio/blob/dcbaca81b9816a40fddc2b0211475a7c5a9220a3/ext/drreg/drreg.h#L610

ext/drreg/drreg.h
```diff
@@ -274,6 +277,9 @@ DR_EXPORT
  * If called during drmgr's insertion phase, \p where must be the
  * current application instruction.
  *
+ * TODO i#3823: Support multi-phase use. This will require adding support
```
abhinav92003, line 280: Any thoughts on priority for this? I'm leaning towards adding this support after this PR and the scatter/gather one are done.

ext/drreg/drreg.c
derekbruening: This is what I was referring to with the earlier comment and the reference to #3801 where we want to replace the from-cache decode with the recreate IR metadata, for cases of xl8-by-recreate.

api/docs/release.dox
```diff
@@ -203,8 +203,9 @@ Further non-compatibility-affecting changes include:
    runtime option which locates guest system libraries and enables workarounds for
    problems with QEMU's threads.
  - Added dr_is_detaching(), an API to query whether detach is in progress.
- - Added instr_zeroes_zmmh() that returns true if an instruction clears the
```
derekbruening, line 206: Bug: you've deleted this entry!

api/docs/release.dox
```diff
@@ -203,8 +203,9 @@ Further non-compatibility-affecting changes include:
...
-   upper bits of a ZMM register with zeros.
+ - Added DRREG_HANDLE_MULTI_PHASE_SLOT_RESERVATIONS to drreg_bb_properties_t to
```
derekbruening, line 206: Use # on both for doxygen links

ext/drmgr/drmgr.c
```diff
@@ -288,7 +288,7 @@ static void
 drmgr_bb_exit(void);
 
 /* Reserve space for emulation specific note values */
-static ptr_uint_t note_base_emul;
+static ptr_uint_t note_base_emul = 0;
```
derekbruening, line 291: Static vars are guaranteed 0 so this does nothing.

ext/drmgr/drmgr.c
```diff
@@ -3145,6 +3145,7 @@ drmgr_emulation_init(void)
 static ptr_int_t
 get_emul_note_val(int enote_val)
 {
+    ASSERT(note_base_emul > 0, "note_base_emul not initialized");
```
derekbruening, line 3148: How do you know 0 is not a valid value?

We already have an assert in drmgr_emulation_init().  If we really want to add a check here, let's use the same symbolic value instead of 0.  I would also suggest moving such a check to the high-level calls instead of a low-level called-multiple-times function.

ext/drreg/drreg.c
```diff
@@ -173,15 +184,58 @@ get_where_app_pc(instr_t *where)
  * SPILLING AND RESTORING
  */
 
+static ptr_uint_t drreg_note_base = 0;
+
+/* Get note value for a drreg instr note. */
+static ptr_uint_t
+get_drreg_note_val(uint val)
+{
+    ASSERT(drreg_note_base > 0, "drreg_note_base not initialized");
```
derekbruening, line 193: Again, use symbolic constants: you have no guarantee that 0 is not a valid value.  Also, this is called many, many times: I would move such an assert to a higher-level less-frequently-called function that calls this, and mark this as inline.

ext/drreg/drreg.c
derekbruening: I would suggest putting a comment about using this metadata with #3801 in the drreg restore state code.

ext/drreg/drreg.dox
```diff
@@ -42,7 +42,13 @@ instrumentation pass scheme.  The results are most efficient when used in
...
 phase, such as during \p drmgr's instrumentation-to-instrumentation
 transformation phase, or use outside of \p drmgr entirely, is supported,
-but duplicate spills and restores may not be optimized away.
+but duplicate spills and restores may not be optimized away.  When using
+drreg to reserve registers in multiple phases (like app2app or
+instru2instru in addition to the insertion phase), it is important to call
+drreg_set_bb_properties() with #DRREG_HANDLE_MULTI_PHASE_SLOT_RESERVATIONS
+to enable logic that avoids conflicts with spill slots used by previous
+phases.  This may have some performance cost, so we do not turn it on by
+default.
```
derekbruening, line 51: Another issue is that the slots used in app2app are now permanently unavailable for insertion.  Many clients tune their register slot usage precisely (TLS slots are limited commodities on Windows).   Please add a sentence or two here making that fact explicit.

We then need to change the drx_init code to ask for additional *summed* slots, or if we don't want to pay for that for drx in general if nobody is using drx_expand_* then we need to document that clients have to do this in their drreg init.  How many slots does drx_expand_* use?  Should we be worried about too many slots on Windows?
See drx_init which looks like it's only asking for drx_insert_counter_update today and is *not* summing:
```
     * We set do_not_sum_slots to true so that we only ask for *more* slots
     * if the client doesn't ask for any.
```

ext/drreg/drreg.h
derekbruening: Since I see drx_expand_scatter_gather calling drreg_reserve_aflags, it does seem like this is high-ish priority, right?

ext/drreg/drreg.h
```diff
@@ -324,6 +330,15 @@ typedef enum {
      * to be handled by the user manually, usually via drreg_restore_all().
      */
     DRREG_USER_RESTORES_AT_BB_END = 0x004,
+
+    /**
+     * Turns on stricter logic to find free register spill slots. This avoids
+     * conflicts with slots used to spill some register value in prior instrumentation
+     * passes. An example usage is in drx_expand_scatter_gather() which is used in
+     * the app2app pass and requires spilling of registers to slots that may
+     * conflict with slots used during later instrumentation passes.
```
derekbruening, line 339: Add that this requires additional slots

ext/drx/drx.c
```diff
@@ -2297,6 +2304,12 @@ drx_expand_scatter_gather(void *drcontext, instrlist_t *bb, OUT bool *expanded)
         return true;
     if (!instr_is_gather(first_app) && !instr_is_scatter(first_app))
         return true;
+
+    /* We want to avoid spill slot conflicts with later instrumentation passes. */
```
derekbruening, line 2308: See earlier comment: also need to adjust total spill slots requested.  We want to avoid moving into the expensive DR non-direct slots.

ext/drx/drx.c
```diff
@@ -85,8 +85,15 @@ enum {
     DRX_NOTE_AFLAGS_RESTORE_END,
     DRX_NOTE_COUNT,
 };
-static ptr_uint_t note_base;
-#define NOTE_VAL(enum_val) ((void *)(ptr_int_t)(note_base + (enum_val)))
+
+static ptr_uint_t note_base = 0;
```
derekbruening, line 89: Again, this change has no effect.

ext/drx/drx.c
```diff
@@ -85,8 +85,15 @@ enum {
...
+
+static void *
+get_note_val(int val)
+{
+    ASSERT(note_base > 0, "note_base not initialized");
```
derekbruening, line 94: 0 might be a valid value.  I would again move such an assert to the high-level exported functions if the goal is to detect a user who didn't call drx_init at the right time.

api/docs/release.dox
abhinav92003: Ah sorry! Sync issues from when I split off these changes from my old PR. Fixed it.

api/docs/release.dox
abhinav92003: Done.

ext/drmgr/drmgr.c
abhinav92003: Removed. Before, I saw some explicit zero initialisations for some static vars (e.g. in drbbdup.c, instrument.c), so thought of being verbose here.

ext/drmgr/drmgr.c
abhinav92003: > How do you know 0 is not a valid value?

`DRMGR_NOTE_NONE` is zero, also `drmgr_reserve_note_range` uses `DRMGR_NOTE_NONE + 1` as the first free range start: https://github.com/DynamoRIO/dynamorio/blob/b39b85d195b63701e17847b849f4fa4d5c37cce4/ext/drmgr/drmgr.c#L3086

>  I would also suggest moving such a check to the high-level calls instead of a low-level called-multiple-times function.

The get_note_val functions seemed like the right place for the check; it may be out of place in other functions that call get_note_val (like drmgr_insert_emulation_start, drx_restore_arith_flags, etc). If we're worried about performance overhead, then I'm fine with just removing this additional check. I guess I was being paranoid anyway.

ext/drreg/drreg.c
abhinav92003: Removed it.

abhinav92003: Marked it as inline too.

ext/drx/drx.c
abhinav92003: Removed.

ext/drx/drx.c
abhinav92003: Removed this check.

ext/drmgr/drmgr.c
derekbruening: > The get_note_val functions seemed like the right place for the check; it may be out of place in other functions 

I assumed the worry is a user incorrectly not calling dmgr_init during process init.  If that's the worry, then checking that init has been called in the API routines the user calls, like drmgr_insert_emulation_start, seems like the place to put the checks.  That was my logic.

ext/drreg/drreg.dox
abhinav92003: > How many slots does drx_expand_* use?

To find this, I used `drreg_max_slots_used` which keeps track of max spill slots used so far. Before encountering a scatter/gather, it seems to be just 2 (probably from drcachesim/other). The first time a scatter/gather is expanded, it goes up to 4. And subsequently, it goes up to 6 (4 reserved from scatter/gather expansion in app2app, and new 2 from drcachesim/other).
This matches what I see in code (one in `expand_avx512_scatter_gather_update_mask` and 3 in `drx_expand_scatter_gather` itself)

`drx_expand_scatter_gather` already asks for more spill slots
https://github.com/DynamoRIO/dynamorio/blob/9319132573ab39bfe292f6850c69d16b7c71c365/ext/drx/drx.c#L2322
But I'm confused by the total 6 it requests and the comment there. As `do_not_sum_slots` is true, I guess it's specifying the total?

ext/drx/drx.c
abhinav92003: `drx_expand_scatter_gather` already does that.

ext/drreg/drreg.h
abhinav92003: Added.

ext/drreg/drreg.dox
derekbruening: Hmm, that `!expand_scatter_gather_drreg_initialized` code looks suspect:
+ It's racy
+ It calls drreg_init potentially quite late, while drreg_init docs say it must be called before any other drreg routines, and as we were discussing it really should be limited to process init

ext/drreg/drreg.c
abhinav92003: Done.

ext/drreg/drreg.dox
abhinav92003: >  if we don't want to pay for that for drx in general if nobody is using drx_expand_* then we need to document that clients have to do this in their drreg init.

I'm leaning towards paying for it in drx for all cases. It doesn't sound reliable to ask clients to be aware of the instrs they are tracing beforehand.

I see that we have some heuristics in core DR that enable expensive saving/restoring of zmm/opmask regs when we see:
- that the client was compiled with avx flags: static check, could work for us, but may not cover all cases
- some avx instr in a basic block: dynamic check. not known at process init

How big a concern is having an extra 4 spill slots all the time?

derekbruening: > How big a concern is having an extra 4 spill slots all the time?

See https://github.com/DynamoRIO/dynamorio/blob/master/core/lib/dr_tools.h#L1972

See #1163.  Large Windows apps can use up the available slots all by themselves, so DR taking any number can cause apps to fail to run.

We added -alt_teb_tls but it looks like it's used last: maybe we should change that and have it used first to have fewer app conflicts: though it's a little hacky, and on Windows we have late injection so all the libs have already grabbed their slots, so maybe it doesn't matter as much.

I definitely agree that it is much simpler to completely separate the app2app slots have drx request the max at init time.  The only issue is Windows TLS space.

It's not the end of the world to fall back on DR slots: they're just slower beyond the first 3 w/ an extra indirection load.

One option would be to request 4 on UNIX and on Windows fall back on DR slots -- except the app2app phase is going to take all the direct slots and other libs/client slots that might be more perf-critical might end up being the DR slots?

Or just take 4 on Windows too and see if anyone complains.  The biggest use of DR on Windows is drcov: if drx were used by drcov I would be concerned, but it's not.

abhinav92003: > One option would be to request 4 on UNIX and on Windows fall back on DR slots 

This sounds reasonable, as Windows is known to be much more sensitive to available slot count.

> -- except the app2app phase is going to take all the direct slots and other libs/client slots that might be more perf-critical might end up being the DR slots?

The app2app phase is going to take all direct slots only for the scatter/gather instr bbs. Note that scatter/gather instrs have their own separate bb. So the perf overhead should be limited to only these basic blocks.

ext/drreg/drreg.h
abhinav92003: I think supporting multi-phase spill for aflags requires some more work. I have just started looking into it, but unlike other regs, aflags seem to be tied to `AFLAGS_SLOT`.

If you agree that this PR is not in an inconsistent state without the aflag support, I'd prefer to submit this first and work on the aflag support in a new PR. I'm feeling there's too many things in-flight with this PR and PR#4863 still pending.

derekbruening: Agreed on splitting: I meant it's high-ish priority in general, not that it needs to be tied to this PR.

ext/drreg/drreg.dox
abhinav92003: Done.

ext/drreg/drreg.h
abhinav92003: Thanks. Changed this PR's title to mention GPRs specifically and added a new TODO on i#2985 for aflags.

ext/drmgr/drmgr.c
abhinav92003: Sure. Can add some checks as part of #4885.

ext/drreg/drreg.dox
derekbruening: I didn't understand the 6 spill slot issue: does it end up using 6 for scatter-gather (that seems like an awful lot)?

abhinav92003: `drx_expand_scatter_gather` itself seems to be using 4; that's what I see on using `drreg_max_slots_used` to get the value at runtime, and comparing before/after `drx_expand_scatter_gather`.

2 others are being used elsewhere (should be drcachesim insertion phase, like `drx_insert_counter_update`). As insertion and app2app phase cannot share spill slots, that's a total of 6.

ext/drreg/drreg.c
```diff
@@ -173,15 +184,57 @@ get_where_app_pc(instr_t *where)
...
+ * previous instrumentation pass, and tells us that it is not safe to reuse
+ * the slot in the current pass.
+ */
+static bool
+has_pending_slot_usage_by_prior_pass(per_thread_t *pt, instrlist_t *ilist, instr_t *where,
+                                     uint slot)
+{
+    if (!TEST(DRREG_HANDLE_MULTI_PHASE_SLOT_RESERVATIONS, pt->bb_props))
+        return false;
+    for (instr_t *in = where; in != NULL /*sanity*/ && in != instrlist_last(ilist);
```
derekbruening, line 207: Please comment on why this loop does not look at the last instr as this is an unusual iteration.

ext/drx/drx.c
```diff
@@ -129,13 +137,20 @@ bool
 drx_init(void)
 {
     /* drx_insert_counter_update() needs 1 slot on x86 plus the 1 slot
-       drreg uses for aflags, and 2 reg slots on aarch, so 2 on both.
+     * drreg uses for aflags, and 2 reg slots on aarch, so 2 on both.
+     * drx_expand_scatter_gather() needs 4 slots in app2app phase, which
+     * cannot be reused by other phases. So, ideally we should reserve
+     * 6 slots for drx. But we settle with 4 to avoid stealing too many
+     * slots from other clients/libs. When drx needs more for instrumenting
+     * scatter/gather instrs, we fall back on DR slots. As scatter/gather
```
derekbruening, line 145: I forgot about this limitation: DR spill slots are not allowed to be used across application instructions.  I think drreg has logic to not do lazy restores for DR spill slots so there is no correctness issue.  (Though in my head I'm worried about clients who have custom slots they use across app instrs...but those should be disjoint from the slots drreg has reserved...)  But the cost is not just the indirection but the extra restores, which can be significant.  Hmm.  Given that this is limited to scatter-gather single-instr bb's: maybe it is fine and all my intuition-directed misgivings are about other TLS slots and not drreg slots.

ext/drx/drx.c
```diff
@@ -129,13 +137,20 @@ bool
...
+     * instrs are spilt into their own bbs, this effect will be limited.
+     * On Windows however we reserve even fewer slots, as they are
+     * shared with the application and reserving even one slot can result
+     * in failure to initialize for certain applications (e.g. i#1163).
      * We set do_not_sum_slots to true so that we only ask for *more* slots
-     * if the client doesn't ask for any. Another drreg_init() call is made
-     * in case drx_expand_scatter_gather() is called. More slots are reserved
-     * in that case.
+     * if the client doesn't ask for any.
```
derekbruening, line 151: With do_not_sum_slots still set to true, this PR is not doing what it says in this comment: it is not reserving 4 slots separate from the client.  If the client itself wants 4 this PR is doing nothing; if the client wants 3 this PR is reserving just 1 extra.

ext/drx/drx.c
derekbruening: I wonder if we could do some performance measurements to reassure ourselves.

ext/drx/drx.c
abhinav92003: Fixed it. Thanks for catching.

ext/drreg/drreg.c
abhinav92003: (I thought I fixed this before, apparently not). Done.

ext/drx/drx.c
abhinav92003: >  DR spill slots are not allowed to be used across application instructions. I think drreg has logic to not do lazy restores for DR spill slots so there is no correctness issue. 

Another reason for there not being any cross-app instr usage of DR slots here:
drreg restores all slot vals at bb end. As scatter/gather instrs have their own separate bb (with only a single instr), the DR slot would not be used beyond the scatter/gather app instr anyway. 

> But the cost is not just the indirection but the extra restores, which can be significant.
> I wonder if we could do some performance measurements to reassure ourselves.

Checking my understanding here: there are multiple components to the additional cost, most of which apply only for scatter/gather instrs. Which is the one we are concerned about?:
- indirection in use of DR slots: if no scatter/gather instr, no DR slots would need to be used as drx reserves atleast as many TLS slots as before in this PR.
- drx using more TLS slots for scatter/gather expansion, making them less available to other clients: Again, if no scatter/gather instr, drx won't use any more TLS slots than before. So, other clients will see same # of TLS slots via drreg.
- reservation of more TLS slots via `dr_raw_tls_calloc`, making them less available to the application: This cost is present even if there aren't any scatter/gather instrs in the app. 
  - For Windows, there'd be no additional cost as we're requesting the same number of TLS slots as before.
  - For Linux, where we do request more TLS slots now: does this cost apply to Linux too? [1] isn't clear about whether TLS slots are shared with apps on Linux too. It only mentions that they are plentiful on Linux and only have a client-total limit of 64 -- are you concerned about hitting that limit?

About performance measurements: I'm trying to think what the baseline would be. If my above understanding is correct, all additional cost is when we see scatter/gather instrs, which weren't supported before now.



[1]: https://github.com/DynamoRIO/dynamorio/blob/10aed67a4e2d0ecd7272101543873b4ecfab78f9/core/lib/dr_tools.h#L1975

derekbruening: > As scatter/gather instrs have their own separate bb (with only a single instr), the DR slot would not be used beyond the scatter/gather app instr anyway.

Aren't the expanded loads/stores marked as app instrs?  I would assume the bb ends up with a bunch of app instrs, let's say 8.  drx's app2app now takes the direct-TLS slots, and the client sticks in some complex instrumentation in the insertion phase which now has to restore and re-spill around each of those 8 app instrs: so if it's using 3 slots that's 24 extra loads and stores.  OTOH the scatter expansion is pretty heavyweight already.

Mostly I recall going to great lengths to tune things to not use DR slots, and problems when that tuning failed.  Dr. Memory is the biggest most complex client we have, so I would say let's test on it: but its switch to use drreg is not finished (and may never be...).  I just "have a bad feeling".  But not enough concrete details to hold this up.

derekbruening: > For Windows, there'd be no additional cost as we're requesting the same number of TLS slots as before.

No, we're asking for 2 additional for clients who are already using 2 (which most are), b/c of the change to do_not_sum_slots.  Perhaps that should remain true for Windows.

derekbruening (approved):

ext/drx/drx.c
abhinav92003: > Aren't the expanded loads/stores marked as app instrs? 

Oh you're right. So drreg's handling to skip lazy restore for DR slots saves the day then.

> No, we're asking for 2 additional for clients who are already using 2 (which most are), b/c of the change to do_not_sum_slots. Perhaps that should remain true for Windows.

Done

ext/drx/drx.c
```diff
@@ -129,13 +137,20 @@ bool
...
+     * drx_expand_scatter_gather() needs 4 slots in app2app phase, which
+     * cannot be reused by other phases. So, ideally we should reserve
+     * 6 slots for drx. But we settle with 4 to avoid stealing too many
+     * slots from other clients/libs. When drx needs more for instrumenting
+     * scatter/gather instrs, we fall back on DR slots. As scatter/gather
+     * instrs are spilt into their own bbs, this effect will be limited.
+     * On Windows however we reserve even fewer slots, as they are
+     * shared with the application and reserving even one slot can result
+     * in failure to initialize for certain applications (e.g. i#1163).
+     * We set do_not_sum_slots to false so that we get atleast as many
```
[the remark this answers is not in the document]
abhinav92003: I suppose you mean "at least" (with the space)? Done.

ext/drx/drx.c
abhinav92003: Discussed offline that we'll need to observe perf impact and fine-tune slot reservations if needed. There are unknowns in its impact on Windows and large clients like DrMemory. Good to submit for now though.

ext/drreg/drreg.dox
abhinav92003: Discussed offline that we'll need to observe perf impact and fine-tune slot reservations if needed. There are unknowns in its impact on Windows and large clients like DrMemory. Good to submit for now though.

PR #4888, opened 2021-04-29
i#4425: save signal stack space with deeper call stack

sapostolakis:
Avoids redundant stack allocations during signal handling for copies of sigcontext_t
by hiding them within callsites.

Issue: #4425

### DIFF
--- a/core/unix/signal.c
+++ b/core/unix/signal.c
@@ -3752,6 +3752,29 @@ handle_client_action_from_cache(dcontext_t *dcontext, int sig, dr_signal_action_
     return true;
 }
 
+static bool
+handle_sent_to_client_signal(dcontext_t *dcontext, int sig, sigframe_rt_t *frame,
+                             sigcontext_t *sc, sigcontext_t *sc_orig,
+                             byte *access_address, bool blocked, fragment_t *fragment,
+                             bool no_deliver)
+{
+    /* make a copy before send_signal_to_client() tweaks it */
+    sigcontext_t sc_interrupted = *sc;
+    dr_signal_action_t action = send_signal_to_client(dcontext, sig, frame, sc_orig,
+                                                      access_address, blocked, fragment);
+    if (blocked) {
+        /* For blocked signal early event we disallow BYPASS (xref i#182/PR 449996) */
+        CLIENT_ASSERT(action != DR_SIGNAL_BYPASS, "cannot bypass a blocked signal event");
+    } else {
+        if (no_deliver && action == DR_SIGNAL_DELIVER) {
+            /* for delivery do not handle here*/
+            return true;
+        }
+    }
+    return handle_client_action_from_cache(dcontext, sig, action, frame, sc_orig,
+                                           &sc_interrupted, blocked);
+}
+
 static void
 abort_on_fault(dcontext_t *dcontext, uint dumpcore_flag, app_pc pc, byte *target, int sig,
                sigframe_rt_t *frame, const char *prefix, const char *signame,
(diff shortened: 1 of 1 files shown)


core/unix/signal.c
```diff
@@ -3752,6 +3752,29 @@ handle_client_action_from_cache(dcontext_t *dcontext, int sig, dr_signal_action_
     return true;
 }
 
+static bool
+handle_sent_to_client_signal(dcontext_t *dcontext, int sig, sigframe_rt_t *frame,
```
derekbruening, line 3756: nit: The function name sounds like the signal was already sent to the client before this function was called

core/unix/signal.c
```diff
@@ -5548,12 +5561,9 @@ execute_handler_from_cache(dcontext_t *dcontext, int sig, sigframe_rt_t *our_fra
...
-    dr_signal_action_t action = send_signal_to_client(
-        dcontext, sig, our_frame, sc_orig, access_address, false /*not blocked*/, f);
-    if (!handle_client_action_from_cache(dcontext, sig, action, our_frame, sc_orig,
-                                         &sc_interrupted, false /*!blocked*/))
+    if (!handle_sent_to_client_signal(dcontext, sig, our_frame, sc, sc_orig,
+                                      access_address, false /*not blocked*/, f,
+                                      false /*handle deliver*/))
         return false;
 
     LOG(THREAD, LOG_ASYNCH, 2, "execute_handler_from_cache for signal %d\n", sig);
```
derekbruening, line 5569: You said this PR shrinks the master_signal_handler_C space: but there's no change here to that function.  One of these must be inlined into it?

core/unix/signal.c
```diff
@@ -4553,20 +4576,14 @@ record_pending_signal(dcontext_t *dcontext, int sig, kernel_ucontext_t *ucxt,
          */
         if (blocked && !forged && !can_always_delay[sig] &&
             safe_is_in_fcache(dcontext, pc, xsp)) {
-            dr_signal_action_t action;
             /* cache the fragment since pclookup is expensive for coarse (i#658) */
             f = fragment_pclookup(dcontext, (cache_pc)sc->SC_XIP, &wrapper);
             sc_orig = *sc;
             translate_sigcontext(dcontext, ucxt, true /*shouldn't fail*/, f);
-            /* make a copy before send_signal_to_client() tweaks it */
-            sigcontext_t sc_interrupted = *sc;
```
derekbruening, line 4562: So the two big structs for record_pending_signal are sc_orig and sc_interrupted.  sc_interrupted was in two places: was the compiler already sharing the space for those?  Hopefully, but I've seen compilers fail to do so before.  That leaves the two sc_orig and sc_interrupted.  This PR avoids sc_interrupted space for code paths that don't send to the client immediately: those are if blocked and synchronous, or if in client gencode.  But it doesn't change the stack usage for those two cases.  I guess the two sc_orig and sc_interrupted are both needed simultaneously?  Is the sc_interrupted copy really needed?  That makes 3 copies: sc_orig==pre-translate; *sc==in the frame; sc_interrupted==translated and..err..would have to study the code.

Looking at send_signal_to_client: it says:
```C
    /* It's safe to allocate since we do not send signals that interrupt DR.
     * With priv_mcontext_t x2 that's a little big for stack alloc.
     */
    si.mcontext = heap_alloc(dcontext, sizeof(*si.mcontext) HEAPACCT(ACCT_OTHER));
    si.raw_mcontext = heap_alloc(dcontext, sizeof(*si.raw_mcontext) HEAPACCT(ACCT_OTHER));
```

Given that, we could heap-allocate this sc_interrupted?  And avoid its stack space for all paths.

core/unix/signal.c
sapostolakis: There are code changes in master_signal_handler_C.

derekbruening: Ah missed the function label in the diff on the first examination.  OK so that does reduce the usage on all record_ paths.

core/unix/signal.c
sapostolakis: The space for multiple instances of sc_interrupted is not shared. There is a separate stack allocation for each of them. Essentially stack was allocated for up to 3x sc_interrupted structs unnecessarily. This PR limits the scope of these allocations forcing at most one of them to be active (and free the stack space immediately after use, assuming there will be no inlining).  

Surely, it might be the case that we do not need sc_interrupted in the first place and more optis are coming.

derekbruening: OK, so you're suggesting moving heap-allocating sc_interrupted into a separate PR.  OK.

derekbruening (approved):

core/unix/signal.c
sapostolakis: I guess it is a bit misleading. Would `send_signal_to_client_and_handle_action()` be better? I am terrible with function names.

derekbruening: `send_signal_to_client_and_handle_action()` SGTM

core/unix/signal.c
```diff
@@ -3752,6 +3752,29 @@ handle_client_action_from_cache(dcontext_t *dcontext, int sig, dr_signal_action_
     return true;
 }
 
+static bool
+handle_sent_to_client_signal(dcontext_t *dcontext, int sig, sigframe_rt_t *frame,
+                             sigcontext_t *sc, sigcontext_t *sc_orig,
+                             byte *access_address, bool blocked, fragment_t *fragment,
+                             bool no_deliver)
+{
+    /* make a copy before send_signal_to_client() tweaks it */
```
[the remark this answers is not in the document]
sapostolakis: Done.

PR #3960, opened 2019-11-21
i#3926: Do not clone label callbacks.

hgreving2304:
Label callbacks are called every time an instruction is destroyed if the instruction is
a label instruction and the field is valid. When cloning an instruction, we can't blindly
copy the label callback field causing it to be called when the cloned instruction is
destroyed. We are adding a note to the docs of instr_clone() and prevent the field from
being copied.

Fixes #3926
Issue: #3962

Issue #3173, opened 2018-09-18
Improve testing of emulation API functions


AssadHashmi:
The implementation of an emulation API requires better test coverage:
- A selection of sample observational clients running with an emulation client.
- Support for x86 and ARM (currently on AArch64).
- More complex emulation clients.

#3112 
#3104 
#3168

PR #4889, opened 2021-04-29
i#3962 label cb: Add instr_clear_label_callback()

### LINKED ISSUE #3962: Label callback uses and instruction cloning inconsistent.
For label instructions with allow the raw bytes field to be re-used for a label callback pointer. If the pointer is valid, it is called when the label instruction is destroyed. This was causing problems when the instruction is cloned, xref https://github.com/DynamoRIO/dynamorio/pull/3960.

We are fixing the hot immediate issue by explicitly not copying the label callback pointer. However, the deeper issue remains about the higher level semantics. Is the cloned instruction passed to the client and the pointer is now invalid? Do we need to make a deep copy? Or should it be transferred? Whatever the solution is, it needs to avoid having two instructions with the same valid pointer since the callback should be unique to an instruction and should only be called once when this instruction is destroyed.

derekbruening:
PR #3960 added a call to instr_set_label_callback() to set it to NULL
from instr_clone(), but if the callback is non-NULL an assert fires in
that case.  This only normally happens with emulation labels that turn
into traces, which happens to not occur in our very few tests of
emulation labels (#3173 covers adding more tests).

We fix that by adding instr_clear_label_callback() here and using that
from instr_clone(), since these callbacks are a little different from
other values and it feels best to not clear them using the set
routine.

A test is added by putting a loop around a scatter-gather expansion,
triggering trace creation.  I confirmed that the assert does fire
without this fix with the loop in place.

Fixes #3962

### DIFF
--- a/suite/tests/client-interface/drx-scattergather.c
+++ b/suite/tests/client-interface/drx-scattergather.c
@@ -503,10 +503,15 @@ test_avx2_avx512_scatter_gather(void)
         return false;
 #    endif
 #    ifdef __AVX__
-    if (!test_avx2_gather(test_avx2_vpgatherdd, ref_sparse_test_buf,
-                          ref_idx32_val32_xmm_ymm_zmm, test_idx32_vec,
-                          output_xmm_ymm_zmm))
-        return false;
+    /* Run in a loop to trigger trace creation and stress things like cloning
+     * (i#3962).
+     */
+    for (int i = 0; i < 100; i++) {
+        if (!test_avx2_gather(test_avx2_vpgatherdd, ref_sparse_test_buf,
+                              ref_idx32_val32_xmm_ymm_zmm, test_idx32_vec,
+                              output_xmm_ymm_zmm))
+            return false;
+    }
     if (!test_avx2_gather(test_avx2_vgatherdps, ref_sparse_test_buf,
                           ref_idx32_val32_xmm_ymm_zmm, test_idx32_vec,
                           output_xmm_ymm_zmm))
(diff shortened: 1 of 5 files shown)


suite/tests/client-interface/drx-scattergather.c
```diff
@@ -503,10 +503,15 @@ test_avx2_avx512_scatter_gather(void)
         return false;
 #    endif
 #    ifdef __AVX__
-    if (!test_avx2_gather(test_avx2_vpgatherdd, ref_sparse_test_buf,
-                          ref_idx32_val32_xmm_ymm_zmm, test_idx32_vec,
-                          output_xmm_ymm_zmm))
-        return false;
+    /* Run in a loop to trigger trace creation and stress things like cloning
```
johnfxgalea, line 506: Are we sure this will always trigger trace creation? I don't want the test to be flaky.

johnfxgalea (approved):

suite/tests/client-interface/drx-scattergather.c
derekbruening: Failing to make a trace wouldn't make the test fail: the output would be the same.  It just might not stress the assert path.  But, yes I think this should always make a trace, unless the default threshold were raised from 50, which is not likely, and even then the test would not be flaky as in sometimes failing.

johnfxgalea: Indeed, okay with me then. At least it won't fail I guess :)

derekbruening: Once the other PR #4877 goes in, we'll have emulation labels with callbacks for rep string loops, and a lot of those will show up in traces, so I think we'll have good coverage: so I don't consider this loop critical (plus I think years from now this will still cover the trace path).

PR #4900, opened 2021-05-11
i#4847 AArch64 v8.0 memory decode: Expand GPR LDR and STR tests

MDevereau:
Add IR and disassembly tests for the GPR versions of LDR and STR
Failed encodings discovered:
- LDR literal (32/64bit variant)

Issue: #2626, #4847


AssadHashmi:
Looks like the patch has failed a64-on-x86 IR tests
https://github.com/DynamoRIO/dynamorio/pull/4900/checks?check_run_id=2558122135

@MDevereau can you replicate the failure locally on an x86_64 machine?

@derekbruening any clues as to how it's possible for IR tests to pass on AArch64 but fail on x86_64?
The codec and IR construction are purely C bit twiddling, no low-level architecture sensitive code.
I see no relevant build warnings in the logs either.

MDevereau:
@AssadHashmi I've replicated the issue locally on a x86-64 machine.

MDevereau:
It seems like it's failing on all negative immediate integers that get sign-extended.

```
    int reg_32[] = { DR_REG_W0, DR_REG_W16, DR_REG_W30 };
    int reg_64[] = { DR_REG_X0, DR_REG_X16, DR_REG_X30 };
    int reg_dest[] = { DR_REG_X1, DR_REG_X17, DR_REG_X30 };
    int value[] = { 129, 255, -256, 170, 85, -86, -171 };

    for (int i = 0; i < 3; i++) {
        for (int l = 0; l < 6; l++) {
            instr_t *instr = INSTR_CREATE_ldr_imm(
                dc, opnd_create_reg(reg_32[i]), opnd_create_reg(reg_dest[i]),
                opnd_create_base_disp_aarch64(reg_dest[i], DR_REG_NULL, 0, false, 0, 0,
                                              OPSZ_4),
                OPND_CREATE_INT(value[l]));
            test_instr_encoding(dc, OP_ldr, instr);
```
The negative numbers in value[] will produce immediate patters such as ```ldr (%x1)[4byte] %x1 $0xffffffffffffff00 -> %w0 %x1```. The tests pass fine when these negatives are removed

derekbruening:
For a64-on-x86 problems can come from DR_HOST_AARCH64 vs AARCH64 but since we made the default the target instead of the host it is much more likely to hit execution problems than to hit IR problems when DR_HOST_NOT_TARGET.

AssadHashmi (approved):

PR #4908, opened 2021-05-12
i#4893: Delay function tracing with address tracing

### LINKED ISSUE #4893: drmemtrace gathers function trace data before hitting -trace_after_instrs trigger
The -trace_after_instrs delayed tracing feature fails to delay function tracing, which gathers and stores data the whole time.  I hit this when running with a -trace_after_instrs threshold that was never reached, yet a whole bunch of trace data was stored.

derekbruening:
In drcachesim, the -trace_after_instrs delayed-tracing feature fails
to delay function tracing.  We fix that here by leaving the function
wrapping in place from the start and simply ignoring i/o until the
trigger point.

Delaying the function wrapping is complicated by being unable to hold
locks during the whole module load event.  Locks are needed to delay
that event, as we need to enable it and then iterate existing modules.
This is why we instead opt to not delay function wrapping.

Adds documentation to drwrap_wrap* that event-blocking locks cannot be
held.

Adds a new test that delays all tracing and ensures there are zero
function markers.

Fixes #4893

### DIFF
--- a/clients/drcachesim/tracer/tracer.cpp
+++ b/clients/drcachesim/tracer/tracer.cpp
@@ -568,12 +568,20 @@ clean_call(void)
  * Tracing instrumentation.
  */
 
+static int tracing_enabled;
+
 static void
 append_marker_seg_base(void *drcontext, func_trace_entry_vector_t *vec)
 {
     per_thread_t *data = (per_thread_t *)drmgr_get_tls_field(drcontext, tls_idx);
     if (BUF_PTR(data->seg_base) == NULL)
         return; /* This thread was filtered out. */
+    // i#4893: While it would be lower overhead to delay function tracing's drwrap
+    // calls rather than solely skipping i/o, the lock restrictions around drwrap
+    // make it difficult to delay func_trace's module load event.  This is the
+    // simplest way to delay function tracing output.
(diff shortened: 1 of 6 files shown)


sapostolakis (approved):

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1436,7 +1443,7 @@ enable_tracing_instrumentation()
             event_bb_instru2instru, &memtrace_pri))
         DR_ASSERT(false);
     dr_register_filter_syscall_event(event_filter_syscall);
-    tracing_enabled = true;
+    dr_atomic_add32_return_sum(&tracing_enabled, 1);
```
abhinav92003, line 1446: Is there any benefit to using `dr_atomic_store32` instead of `dr_atomic_add32_return_sum`? Former may be less expensive I guess.

abhinav92003: As `tracing_enabled` is just a boolean that makes a one-way transition to `true`, may be better to use former.

derekbruening: Switched to store.

PR #4909, opened 2021-05-12
i#2985 scatter-gather: Fix scratch mask restoration on fault during scatter

abhinav92003:
Adds a missing state transition that was causing restoration of the wrong value to scratch mask `k0`. This was resulting in a ` ERROR: expected k0 == 0xffff, but is 0xd360` error in the `drx-scattergather` app.
    
Postpones updating trace buffer ptr in the `memval_simple` sample client to post-write. This is to mitigate cases when the write buffer doesn't get written to due to the app write segfaulting, which results in no matching entry in the write buffer for the trace buffer entry. This causes an assert failure later `write_base <= write_ptr` in `trace_fault`.

Adds a test that runs the `drx-scattergather` app under the `memval_simple` client to verify the above two fixes.
    
Issue: #2985


api/samples/memval_simple.c
```diff
@@ -219,8 +219,7 @@ instrument_mem(void *drcontext, instrlist_t *ilist, instr_t *where, opnd_t ref)
     size = (ushort)drutil_opnd_mem_size_in_bytes(ref, where);
```
derekbruening, line 219: > Adds a missing state transition that was causing restoration of wrong value to scratch mask k0.

nit: *the* wrong value

api/samples/memval_simple.c
```diff
@@ -219,8 +219,7 @@ instrument_mem(void *drcontext, instrlist_t *ilist, instr_t *where, opnd_t ref)
...
     drx_buf_insert_buf_store(drcontext, trace_buffer, ilist, where, reg_ptr, reg_tmp,
                              OPND_CREATE_INT16(size), OPSZ_2, offsetof(mem_ref_t, size));
-    drx_buf_insert_update_buf_ptr(drcontext, trace_buffer, ilist, where, reg_ptr,
-                                  DR_REG_NULL, sizeof(mem_ref_t));
+    /* postpone updating trace_buffer ptr to post-write, in case the write segfaults */
```
derekbruening, line 222: nit: Prefer capitalization and punctuation for comments.

suite/tests/CMakeLists.txt
```diff
@@ -3079,11 +3083,21 @@ if (BUILD_SAMPLES)
         endif (UNIX)
       elseif (sample STREQUAL "memval_simple")
         # i#2449: only witnessed crash on 32-bit linux
-        if (X86 AND (NOT X64) AND UNIX)
-          tobuild(client.memval-test client-interface/memval-test.c)
-          torunonly_ci(sample.${sample} client.memval-test ${sample}
-              client-interface/memval-test.c "" "" "")
-        endif(X86 AND (NOT X64) AND UNIX)
+        if (X86 AND UNIX)
+          if (NOT X64)
```
derekbruening, line 3087: Does this memval-test run on 64-bit?  If so it seems best to just remove the `NOT X64` and the `i#2449` comment.

api/samples/memval_simple.c
derekbruening, line 222: But instrument_mem() is called for loads too, and this change means the buffer pointer is never advanced for loads?

api/samples/memval_simple.c
```diff
@@ -280,6 +279,10 @@ instrument_post_write(void *drcontext, instrlist_t *ilist, instr_t *where, opnd_
     drx_buf_insert_buf_memcpy(drcontext, write_buffer, ilist, where, reg_ptr, reg_addr,
                               stride);
 
+    drx_buf_insert_load_buf_ptr(drcontext, trace_buffer, ilist, where, reg_ptr);
```
derekbruening, line 282: The drx_buf_insert_buf_memcpy docs say it updates the buffer pointer -- does it clobber reg_ptr too and that's why it has to be reloaded?  But I don't understand: updating the pointer for the address, which was written pre-app-instr, would have to be done *before* storing the value here: else the value will clobber the address, and then this extra advance will just have uninitialized data in the buffer.

api/samples/memval_simple.c
derekbruening, line 222: I'm not sure I understand: is it the *app* write faulting, or the *trace buffer write* faulting?

abhinav92003: Done.

suite/tests/CMakeLists.txt
abhinav92003: It works. Done.

api/samples/memval_simple.c
abhinav92003: This[1] loop calls `instrument_mem` only for dests, so seems like it's only for stores? Also this comment[2] documents that `memval_simple` dumps only write addresses.

[1]: https://github.com/DynamoRIO/dynamorio/blob/cf9d6a95262015581a5184e75ce599cc66ac4df4/api/samples/memval_simple.c#L356
[2]: https://github.com/DynamoRIO/dynamorio/blob/cf9d6a95262015581a5184e75ce599cc66ac4df4/api/samples/memval_simple.c#L36

abhinav92003: The app write is faulting. There are some writes[1] in `drx-scattergather.c` that intentionally `SIGSEGV` to verify whether the scratch mask register is restored upon a fault.

[1]: https://github.com/DynamoRIO/dynamorio/blob/cf9d6a95262015581a5184e75ce599cc66ac4df4/suite/tests/client-interface/drx-scattergather.c#L569

api/samples/memval_simple.c
abhinav92003: Done.

api/samples/memval_simple.c
abhinav92003: There are two different buffers here -- `trace_buffer` that stores addr, size, type of the write, and `write_buffer` that stores the actual written value. The former is written before the app write by `instrument_mem`, whereas the latter is written after the app write by `instrument_post_write`.

> does it clobber reg_ptr too and that's why it has to be reloaded?

Before L282, `reg_ptr` has the `write_buffer` ptr, but we need the `trace_buffer` ptr, so we reload.

> else the value will clobber the address, and then this extra advance will just have uninitialized data in the buffer.

The written value is stored in a different buffer, so it won't clobber the address.

abhinav92003: By delaying the advancement of the `trace_buffer` ptr, in a way we delay committing the <addr, type, size> written by `instrument_mem`. So if the app write fails for some reason and the app recovers from the error (like `drx-scattergather.c` does), there'll be no mismatch between `trace_buffer` and `write_buffer`, and the DR client will be able to continue on.

derekbruening: This adds an extra load for a very rare path: most app stores do not fault.  If the buffer pointer discrepancy can be handled in the fault handler, that seems a superior solution, rather than forcing every single non-faulting store to pay extra overhead, right?

api/samples/memval_simple.c
derekbruening: I see.  Maybe this should be called instrument_pre_write() rather than instrument_mem(): that would make it clearer.

derekbruening: OK, some of these buffers use faults to handle filling up, so just making sure I understand.  The comment does imply it's the app, but maybe making it even clearer could help.

api/samples/memval_simple.c
abhinav92003: I agree. I considered handling this case in a fault handler, but decided against the extra complexity in this sample client. If performance is a bigger concern here, sure I can use the fault handler strategy.

api/samples/memval_simple.c
abhinav92003: Done.

abhinav92003: Done. Renamed.

abhinav92003: Will push with later commits.

api/samples/memval_simple.c
derekbruening: How much more complex is it?  It looks like this client already has a fault handler.  Oh wait -- that's the trace buffer fault, not the app fault.  OK, I was thinking it would just be one line in the existing handler, but it would take a whole new event.  Yeah, doing it this way with a comment that this is extra common path overhead that could be avoided with an app fault handler seems fine.

derekbruening (approved):

api/samples/memval_simple.c
abhinav92003: Thanks. Done.

Issue #4865, opened 2021-04-21
Update drcachesim and sample tracers to record original, not emulation, instruction sequences


derekbruening:
With #3112 we have support for marking emulation sequences where the original application code is distinct from what is executed on the machine.  In addition to emulation for features from different generations of architectures, we plan to have emulation markers for rep string expansion and scatter-gather expansion.  This issue covers changing the drcachesim/drmemtrace tracer and all our sample tracers to record the original, rather than emulation, sequences.  We already have hardcoded special cases for rep string expansion in drcachesim; we should be able to eliminate that and replace with general handling.

derekbruening:
For rep str expansion and scatter gather expansion, our tools should look at the original instruction for instruction tracing, but the emulation sequence for recording data addresses (and/or values) in traces.

But does that hold for all emulation sequences?  There aren't many examples to look at (xref #3173).  @AssadHashmi does this generalization seem reasonable for all of the various emulations used by ArmIE, or would most of them want the *original* instruction's data address?  And if so do we need a flag indicating whether an emulation sequence is a tool-simplification-expansion versus a true emulation?

derekbruening:
Note that the rep str and scatter gather expansions are today using the translation field to point at the original insruction, which makes instruction tracing mostly do the right thing: the original instruction address, with the emulation data addresses/values.  However, the rep str expansion ends up with 3 app instructions instead of 1, messing up instruction counters; plus I believe it has the wrong instruction size.  This is why we wanted to use the emulation markers there.

derekbruening:
(The instruction counting problem shows up in things like #4857 and today are solved with hardcoded checks for a rep str expansion.)

AssadHashmi:
> But does that hold for all emulation sequences? There aren't many examples to look at (xref #3173). @AssadHashmi does this generalization seem reasonable for all of the various emulations used by ArmIE,  

@derekbruening to clarify, when you say
> or would most of them want the _original_ instruction's data address?

Do you mean the instruction to be emulated and its address as it appears in the subject binary execution space? If so, the answer is yes.

> And if so do we need a flag indicating whether an emulation sequence is a tool-simplification-expansion versus a true emulation?

Does this mean an emulation sequence of native instructions will either include instrumentation instructions or it won't as indicated by the flag?

derekbruening:
> > And if so do we need a flag indicating whether an emulation sequence is a tool-simplification-expansion versus a true emulation?
>
[...]

Issue #2051, opened 2016-11-01
drcachesim incorrectly counts each rep string iter as an ifetch


derekbruening:
Xref #2011

The original rep string loop only has one ifetch for the whole loop, while the drutil-expanded instru in drcachesim has an ifetch per iteration.

This will be easy to solve for offline, but harder for online: in fact it seems that some kind of explicit iter count check is needed.

derekbruening:
Offline now looks like this:

```
[drmemtrace]: Appending 1 instrs in bb 0x00007fb352d124ba in mod 3 +0x1a4ba = /usr/lib64/ld-2.21.so
 0x00007f946082f4ba rep stos %al %rdi %rcx -> %es:(%rdi)[1byte] %rdi %rcx
[drmemtrace]: Appended memref to 0x00007f945f97a028
[drmemtrace]: Appending 1 instrs in bb 0x00007fb352d124ba in mod 3 +0x1a4ba = /usr/lib64/ld-2.21.so
[drmemtrace]: Skipping instr fetch for 0x00007fb352d124ba
[drmemtrace]: Appended memref to 0x00007f945f97a029
[drmemtrace]: Appending 1 instrs in bb 0x00007fb352d124ba in mod 3 +0x1a4ba = /usr/lib64/ld-2.21.so
[drmemtrace]: Skipping instr fetch for 0x00007fb352d124ba
[drmemtrace]: Appended memref to 0x00007f945f97a02a
```

derekbruening:
+ 514e90d i#2051 drcachesim repstr: skip repstr ifetch for offline

derekbruening:
We'd like to support feeding our traces to core simulators as well as cache simulators.
But, while a cache simulator expects a single instr fetch followed by N data
refs, a core simulator expects an instr with each iteration.

Middle ground would be to include the count with the first rep instr,
though this complicates the currently-unified trace format.

There could be a thread switch in the middle of a rep loop?  Yes, but
raw2trace will insert a rep instr when the thread resumes.

Since it’s easier to ignore the instr if it matches the prior instr than to
look ahead, we could go back to including the instr in each iteration and
force cache simulators to add special handling.

derekbruening:
I looked at some other tools to see how they handle rep string instructions.  I made a tiny app with no crt that just executes some assembly:

```
# To build:
# as -o allasm_rep64.o allasm_rep64.s && gcc -static -o allasm_rep64 allasm_rep64.o -nostartfiles
.text
[...]

PR #4912, opened 2021-05-14
i#2985 scatter-gather: Fix drcachesim and raw2trace issues.

abhinav92003:
Fixes issues in drcachesim and raw2trace because of which some
instructions and memrefs in the scatter/gather expansion were
not being instrumented. This led to lower than expected
basic_counts for data loads and stores.

In the scatter/gather expansion, all instrs have the same app
pc (this is unlike rep str expansion where the stringop has
the original app pc, and others have a fake one). So we need
to continue instrumenting even if we see an instr with app pc
same as the last one we saw.

Modifies raw2trace to properly detect scatter/gather memrefs:
when we see a scatter/gather instr, we keep reading entries
until we encounter a non-memref entry. The alternative to this
was to add a pc entry for each mov instr (to fulfil raw2trace's
requirement of always having a preceding pc entry for each
memref), but that may have higher overhead.

Another tricky behavior of scatter/gather expansion is that
the first non-label instruction in the expansion is a non-
app memref, which needs to be instrumented to output pc entry
(for scatter/gather completely suppressed due to  mask, we
still want a pc entry). But as we don't want a memref entry for
that, had to add a special check for it.

Figures out pc for scatter/gather instr using fragment tag,
instead of the passed instr. As the first instr in the expanded
scatter/gather is a non-app instr, we cannot use that to
get the app pc. 

Hard-codes type and length for scatter/gather instrs in
online traces. Type and length cannot be found out using
the passed-in instr (which is from the expansion), so we
use the original instr at given pc to get those details.

Disables instr bundling for scatter/gather expansions, so
that we can choose to not output a duplicate pc entry for
some instrs in the online trace.

Adds drcachesim online test for allasm scatter gather app.

Adds an allasm app that uses repstr, and a drcachesim offline
and online test for this.

Issue: #2985

### DIFF
--- a/clients/drcachesim/tests/allasm-repstr-basic-counts.templatex
+++ b/clients/drcachesim/tests/allasm-repstr-basic-counts.templatex
@@ -4,6 +4,7 @@ Basic counts tool results:
 Total counts:
           14 total \(fetched\) instructions
           14 total unique \(fetched\) instructions
+/* One for each movs after the first one. */
            4 total non-fetched instructions
            0 total prefetches
            5 total data loads
--- a/clients/drcachesim/tests/allasm-repstr-basic-counts.templatex
+++ b/clients/drcachesim/tests/allasm-repstr-basic-counts.templatex
@@ -1,14 +1,14 @@
-Correct
-Hello world!
+Adios world!
+---- <application exited with code 0> ----
 Basic counts tool results:
 Total counts:
-          43 total \(fetched\) instructions
-          43 total unique \(fetched\) instructions
-/* One for each store/load in expanded scatter/gather seq. */
-           6 total non-fetched instructions
+          14 total \(fetched\) instructions
+          14 total unique \(fetched\) instructions
+/* One for each movs after the first one */
+           4 total non-fetched instructions
            0 total prefetches
-           3 total data loads
-           3 total data stores
+           5 total data loads
+           5 total data stores
            0 total icache flushes
            0 total dcache flushes
            1 total threads
@@ -20,12 +20,12 @@ Total counts:
      .* total function return value markers
      .* total other markers
 Thread .* counts:
-          43 \(fetched\) instructions
-          43 unique \(fetched\) instructions
-           6 non-fetched instructions
+          14 \(fetched\) instructions
+          14 unique \(fetched\) instructions
+           4 non-fetched instructions
            0 prefetches
-           3 data loads
-           3 data stores
+           5 data loads
+           5 data stores
            0 icache flushes
            0 dcache flushes
      .* scheduling markers
--- a/clients/drcachesim/tests/allasm-repstr-basic-counts.templatex
+++ b/clients/drcachesim/tests/allasm-repstr-basic-counts.templatex
@@ -4,8 +4,7 @@ Basic counts tool results:
 Total counts:
           14 total \(fetched\) instructions
           14 total unique \(fetched\) instructions
-/* One for each movs after the first one */
-           4 total non-fetched instructions
+           0 total non-fetched instructions
            0 total prefetches
            5 total data loads
            5 total data stores
@@ -22,7 +21,7 @@ Total counts:
 Thread .* counts:
           14 \(fetched\) instructions
           14 unique \(fetched\) instructions
-           4 non-fetched instructions
+           0 non-fetched instructions
            0 prefetches
            5 data loads
            5 data stores
--- a/clients/drcachesim/tests/allasm-repstr-basic-counts.templatex
+++ b/clients/drcachesim/tests/allasm-repstr-basic-counts.templatex
@@ -4,7 +4,7 @@ Basic counts tool results:
 Total counts:
           14 total \(fetched\) instructions
           14 total unique \(fetched\) instructions
-           0 total non-fetched instructions
+           4 total non-fetched instructions
            0 total prefetches
            5 total data loads
            5 total data stores
@@ -21,7 +21,7 @@ Total counts:
 Thread .* counts:
           14 \(fetched\) instructions
           14 unique \(fetched\) instructions
-           0 non-fetched instructions
+           4 non-fetched instructions
            0 prefetches
            5 data loads
            5 data stores
--- a/clients/drcachesim/tracer/instru_offline.cpp
+++ b/clients/drcachesim/tracer/instru_offline.cpp
@@ -757,13 +757,11 @@ offline_instru_t::identify_elidable_addresses(void *drcontext, instrlist_t *ilis
         // identifying elision opportunities. We can possibly provide a consistent
         // view by expanding the instr in raw2trace (e.g. using
         // drx_expand_scatter_gather) when building the ilist.
-#ifdef X86
-        // TODO i#3837: Scatter/gather support NYI on ARM/AArch64.
-        if (drutil_instr_is_stringop_loop(instr) ||
-            (instr_is_scatter(instr) || instr_is_gather(instr))) {
+        if (drutil_instr_is_stringop_loop(instr)
+            // TODO i#3837: Scatter/gather support NYI on ARM/AArch64.
+            IF_X86(|| instr_is_scatter(instr) || instr_is_gather(instr))) {
             return;
         }
-#endif
         if (drmgr_is_emulation_start(instr) || drmgr_is_emulation_end(instr)) {
             return;
         }
--- a/clients/drcachesim/tracer/instru_offline.cpp
+++ b/clients/drcachesim/tracer/instru_offline.cpp
@@ -587,8 +587,7 @@ offline_instru_t::instrument_memref(void *drcontext, instrlist_t *ilist, instr_t
 int
 offline_instru_t::instrument_instr(void *drcontext, void *tag, void **bb_field,
                                    instrlist_t *ilist, instr_t *where, reg_id_t reg_ptr,
-                                   int adjust, instr_t *app, bool scatter_gather_expanded,
-                                   bool repstr_expanded)
+                                   int adjust, instr_t *app)
 {
     app_pc pc;
     reg_id_t reg_tmp;
@@ -600,7 +599,10 @@ offline_instru_t::instrument_instr(void *drcontext, void *tag, void **bb_field,
     } else {
         // This routine is called for the first instr in the expanded scatter/gather
         // sequence, which turns out to be a non-app instr which doesn't have an app pc.
+        // To verify this, we have an assert in tracer.cpp where we have info on whether
+        // the current bb has an expanded scatter/gather instr.
         // XXX: For repstr do we want tag insted of skipping rep prefix?
+        bool scatter_gather_expanded = !instr_is_app(where);
         pc = scatter_gather_expanded ? dr_fragment_app_pc(tag) : instr_get_app_pc(app);
     }
     drreg_status_t res =
--- a/clients/drcachesim/tracer/instru_offline.cpp
+++ b/clients/drcachesim/tracer/instru_offline.cpp
@@ -628,8 +628,7 @@ offline_instru_t::instrument_ibundle(void *drcontext, instrlist_t *ilist, instr_
 
 void
 offline_instru_t::bb_analysis(void *drcontext, void *tag, void **bb_field,
-                              instrlist_t *ilist, bool repstr_expanded,
-                              bool scatter_gather_expanded)
+                              instrlist_t *ilist, bool repstr_expanded)
 {
     instr_t *instr;
     ptr_uint_t count = 0;
@@ -656,8 +655,7 @@ offline_instru_t::bb_analysis(void *drcontext, void *tag, void **bb_field,
     }
     *bb_field = (void *)count;
 
-    identify_elidable_addresses(drcontext, ilist, OFFLINE_FILE_VERSION,
-                                scatter_gather_expanded);
+    identify_elidable_addresses(drcontext, ilist, OFFLINE_FILE_VERSION);
 }
 
 bool
@@ -740,16 +738,8 @@ offline_instru_t::label_marks_elidable(instr_t *instr, OUT int *opnd_index,
 
 void
 offline_instru_t::identify_elidable_addresses(void *drcontext, instrlist_t *ilist,
-                                              int version, bool has_scatter_gather_instr)
+                                              int version)
 {
-    // XXX: We turn off address elision for bbs containing scatter/gather. As
-    // instru_offline and raw2trace see different instrs in scatter gather bbs
-    // (expanded seq vs original app instr), there may be mismatches in identifying
-    // elision opportunities. We can possibly provide a consistent view by using
-    // drx_expand_scatter_gather in raw2trace when building the ilist.
-    if (has_scatter_gather_instr) {
-        return;
-    }
     // Analysis for eliding redundant addresses we can reconstruct during
     // post-processing.
     if (disable_optimizations_)
@@ -758,6 +748,24 @@ offline_instru_t::identify_elidable_addresses(void *drcontext, instrlist_t *ilis
     if (memref_needs_full_info_)
         return;
     reg_id_set_t saw_base;
+    for (instr_t *instr = instrlist_first(ilist); instr != NULL;
+         instr = instr_get_next(instr)) {
+        // XXX: We turn off address elision for bbs containing emulation sequences
+        // or instrs that are expanded into emulation sequences like scatter/gather
+        // and rep stringop. As instru_offline and raw2trace see different instrs in
+        // these bbs (expanded seq vs original app instr), there may be mismatches in
+        // identifying elision opportunities. We can possibly provide a consistent
+        // view by expanding the instr in raw2trace (e.g. using
+        // drx_expand_scatter_gather) when building the ilist.
+        if (drutil_instr_is_stringop_loop(instr)
+            // TODO i#3837: Scatter/gather support NYI on ARM/AArch64.
+            IF_X86(|| instr_is_scatter(instr) || instr_is_gather(instr))) {
+            return;
+        }
+        if (drmgr_is_emulation_start(instr) || drmgr_is_emulation_end(instr)) {
+            return;
+        }
+    }
     for (instr_t *instr = instrlist_first_app(ilist); instr != NULL;
          instr = instr_get_next_app(instr)) {
         // For now we bail at predication.
--- a/clients/drcachesim/tracer/instru_offline.cpp
+++ b/clients/drcachesim/tracer/instru_offline.cpp
@@ -599,11 +599,11 @@ offline_instru_t::instrument_instr(void *drcontext, void *tag, void **bb_field,
     } else {
         // This routine is called for the first instr in the expanded scatter/gather
         // sequence, which turns out to be a non-app instr which doesn't have an app pc.
-        // To verify this, we have an assert in tracer.cpp where we have info on whether
-        // the current bb has an expanded scatter/gather instr.
+        // XXX i#4865: Refactor tracer so that the first expanded scatter/gather
+        // sequence instr that we instrument doesn't need to be a non-app instr.
         // XXX: For repstr do we want tag insted of skipping rep prefix?
-        bool scatter_gather_expanded = !instr_is_app(where);
-        pc = scatter_gather_expanded ? dr_fragment_app_pc(tag) : instr_get_app_pc(app);
+        DR_ASSERT(instr_is_app(where) || drmgr_is_first_nonlabel_instr(drcontext, where));
+        pc = instr_is_app(where) ? instr_get_app_pc(app) : dr_fragment_app_pc(tag);
     }
     drreg_status_t res =
         drreg_reserve_register(drcontext, ilist, where, reg_vector_, &reg_tmp);
--- a/clients/drcachesim/tracer/instru_offline.cpp
+++ b/clients/drcachesim/tracer/instru_offline.cpp
@@ -757,11 +757,13 @@ offline_instru_t::identify_elidable_addresses(void *drcontext, instrlist_t *ilis
         // identifying elision opportunities. We can possibly provide a consistent
         // view by expanding the instr in raw2trace (e.g. using
         // drx_expand_scatter_gather) when building the ilist.
-        if (drutil_instr_is_stringop_loop(instr)
-            // TODO i#3837: Scatter/gather support NYI on ARM/AArch64.
-            IF_X86(|| instr_is_scatter(instr) || instr_is_gather(instr))) {
+#ifdef X86
+        // TODO i#3837: Scatter/gather support NYI on ARM/AArch64.
+        if (drutil_instr_is_stringop_loop(instr) ||
+            (instr_is_scatter(instr) || instr_is_gather(instr))) {
             return;
         }
+#endif
         if (drmgr_is_emulation_start(instr) || drmgr_is_emulation_end(instr)) {
             return;
         }
--- a/clients/drcachesim/tracer/instru_offline.cpp
+++ b/clients/drcachesim/tracer/instru_offline.cpp
@@ -587,7 +587,8 @@ offline_instru_t::instrument_memref(void *drcontext, instrlist_t *ilist, instr_t
 int
 offline_instru_t::instrument_instr(void *drcontext, void *tag, void **bb_field,
                                    instrlist_t *ilist, instr_t *where, reg_id_t reg_ptr,
-                                   int adjust, instr_t *app)
+                                   int adjust, instr_t *app, bool scatter_gather_expanded,
+                                   bool repstr_expanded)
 {
     app_pc pc;
     reg_id_t reg_tmp;
@@ -597,8 +598,10 @@ offline_instru_t::instrument_instr(void *drcontext, void *tag, void **bb_field,
             return adjust;
         pc = dr_fragment_app_pc(tag);
     } else {
+        // This routine is called for the first instr in the expanded scatter/gather
+        // sequence, which turns out to be a non-app instr which doesn't have an app pc.
         // XXX: For repstr do we want tag insted of skipping rep prefix?
-        pc = instr_get_app_pc(app);
+        pc = scatter_gather_expanded ? dr_fragment_app_pc(tag) : instr_get_app_pc(app);
     }
     drreg_status_t res =
         drreg_reserve_register(drcontext, ilist, where, reg_vector_, &reg_tmp);
--- a/clients/drcachesim/tracer/instru_offline.cpp
+++ b/clients/drcachesim/tracer/instru_offline.cpp
@@ -628,7 +628,8 @@ offline_instru_t::instrument_ibundle(void *drcontext, instrlist_t *ilist, instr_
 
 void
 offline_instru_t::bb_analysis(void *drcontext, void *tag, void **bb_field,
-                              instrlist_t *ilist, bool repstr_expanded)
+                              instrlist_t *ilist, bool repstr_expanded,
+                              bool scatter_gather_expanded)
 {
     instr_t *instr;
     ptr_uint_t count = 0;
@@ -655,7 +656,8 @@ offline_instru_t::bb_analysis(void *drcontext, void *tag, void **bb_field,
     }
     *bb_field = (void *)count;
 
-    identify_elidable_addresses(drcontext, ilist, OFFLINE_FILE_VERSION);
+    identify_elidable_addresses(drcontext, ilist, OFFLINE_FILE_VERSION,
+                                scatter_gather_expanded);
 }
 
 bool
@@ -738,8 +740,16 @@ offline_instru_t::label_marks_elidable(instr_t *instr, OUT int *opnd_index,
 
 void
 offline_instru_t::identify_elidable_addresses(void *drcontext, instrlist_t *ilist,
-                                              int version)
+                                              int version, bool has_scatter_gather_instr)
 {
+    // XXX: We turn off address elision for bbs containing scatter/gather. As
+    // instru_offline and raw2trace see different instrs in scatter gather bbs
+    // (expanded seq vs original app instr), there may be mismatches in identifying
+    // elision opportunities. We can possibly provide a consistent view by using
+    // drx_expand_scatter_gather in raw2trace when building the ilist.
+    if (has_scatter_gather_instr) {
+        return;
+    }
     // Analysis for eliding redundant addresses we can reconstruct during
     // post-processing.
     if (disable_optimizations_)
--- a/clients/drcachesim/tracer/instru_online.cpp
+++ b/clients/drcachesim/tracer/instru_online.cpp
@@ -318,7 +318,7 @@ online_instru_t::instrument_instr(void *drcontext, void *tag, void **bb_field,
 {
     // We cannot rely on the given instr's app_pc in case of expanded repstr and
     // scatter/gather sequences. In both cases, this routine is called for the instr
-    // at the top of the bb.
+    // at the top of the expanded sequence (which has its own separate bb).
     // - in the repstr expansion, this is a jrcxz instr with a fake app pc.
     // - in the scatter/gather expansion, this is a non-app reg spill instr, which
     // doesn't have an app_pc set.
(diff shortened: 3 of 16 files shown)


derekbruening:
Before putting more work into this maybe we should discuss how this interacts with #4865 which is in progress and is going to end up with a single instruction and many memrefs:

```event_bb_insert(instr_t *exec_instr) {
  emulated_instr_t emulation;
  if (drmgr_in_emulation_region(&emulation)) {
    if (TEST(DR_EMULATE_FIRST_INSTR, emulation.flags)) {
      record_instr_fetch(emulation.instr);
      if (TEST(DR_EMULATE_DATA, emulation.flags))
        record_data_addresses(emulation.instr);
    } // Else skip further instr fecthes until outside emulation region.
    if (instr_is_app(exec_instr) && !TEST(DR_EMULATE_DATA, emulation.flags))
      record_data_addresses(exec_instr);
  } else if (instr_is_app(exec_instr)) {
    record_instr_fetch(exec_instr);
    record_data_addresses(exec_instr);
  }
}
```

Does the #4865 proposal need to change?  Let's discuss general solutions.

derekbruening:
> Does the #4865 proposal need to change? Let's discuss general solutions.

Let's put the discussion in the issue: https://github.com/DynamoRIO/dynamorio/issues/4865#issuecomment-841498987

clients/drcachesim/tests/allasm-repstr-basic-counts.templatex
```diff
@@ -0,0 +1,37 @@
+Adios world!
+---- <application exited with code 0> ----
+Basic counts tool results:
+Total counts:
+          14 total \(fetched\) instructions
+          14 total unique \(fetched\) instructions
+/* One for each movs after the first one. */
```
abhinav92003, line 7: Unfortunately, in the repstr case, we cannot avoid the pc entry added by each iteration of the stringop (hence the 4 non-fetched instrs here). All instrs in the expanded repstr seq are in a loop, so the pc entry will be output each time.

This is unlike scatter/gather expansion where there's no loop, and we could choose to not instrument mov.

clients/drcachesim/tracer/instru_offline.cpp
```diff
@@ -597,8 +598,10 @@ offline_instru_t::instrument_instr(void *drcontext, void *tag, void **bb_field,
             return adjust;
         pc = dr_fragment_app_pc(tag);
     } else {
+        // This routine is called for the first instr in the expanded scatter/gather
+        // sequence, which turns out to be a non-app instr which doesn't have an app pc.
         // XXX: For repstr do we want tag insted of skipping rep prefix?
-        pc = instr_get_app_pc(app);
+        pc = scatter_gather_expanded ? dr_fragment_app_pc(tag) : instr_get_app_pc(app);
```
abhinav92003, line 604: The instrs added by drreg in `drx_expand_scatter_gather` are not marked as app instrs today. I suppose if we do mark them as app and set their translation, we won't need this special logic here. It seems intuitive that they should be app instrs, as they are added by the app2app phase.

But marking them as app raises another question: should reg spills and restores in the expanded sequence be counted as app memref as well?

derekbruening: drreg spill instrs should not be app instrs.  We already have non-app instrs in app2app from repstr expansion and we're just living w/ the fact that clients should check for app vs tool in analysis and insertion phase.

clients/drcachesim/tests/allasm-repstr-basic-counts.templatex
derekbruening: ?  These statistics are on the *final* trace which must have an entry for each iteration of the loop: we can't skip them in the final trace.  Whether the raw trace has all of them or not is a different matter.  Xref #4914.

clients/drcachesim/tracer/instru_offline.cpp
```diff
@@ -597,8 +598,10 @@ offline_instru_t::instrument_instr(void *drcontext, void *tag, void **bb_field,
             return adjust;
         pc = dr_fragment_app_pc(tag);
     } else {
+        // This routine is called for the first instr in the expanded scatter/gather
+        // sequence, which turns out to be a non-app instr which doesn't have an app pc.
```
derekbruening, line 602: It seems best to require that `app` be an app instr, rather than passing in flags and having special cases here.

clients/drcachesim/tracer/instru_online.cpp
```diff
@@ -313,20 +313,34 @@ online_instru_t::instrument_memref(void *drcontext, instrlist_t *ilist, instr_t
...
+                                  bool repstr_expanded)
 {
-    bool repstr_expanded = *bb_field != 0; // Avoid cl warning C4800.
-    app_pc pc = repstr_expanded ? dr_fragment_app_pc(tag) : instr_get_app_pc(app);
+    // We cannot rely on the given instr's app_pc in case of expanded repstr and
+    // scatter/gather sequences. In both cases, this routine is called for the instr
+    // at the top of the expanded sequence (which has its own separate bb).
+    // - in the repstr expansion, this is a jrcxz instr with a fake app pc.
+    // - in the scatter/gather expansion, this is a non-app reg spill instr, which
+    // doesn't have an app_pc set.
```
derekbruening, line 324: Rather than passing in all these booleans and having special cases here, it seems better to only call this for app instrs.

suite/tests/CMakeLists.txt
```diff
@@ -2624,23 +2624,35 @@ if (X86)
...
+        add_exe(allasm_scattergather
+          ${PROJECT_SOURCE_DIR}/clients/drcachesim/tests/allasm_scattergather.asm
+          "-early_inject" "")
+        set_target_properties(allasm_scattergather PROPERTIES LINKER_LANGUAGE C)
+        append_link_flags(allasm_scattergather
+          "-nostartfiles -nodefaultlibs -static")
+        append_property_string(TARGET allasm_scattergather LINK_FLAGS
+          # The default value of CMAKE_SHARED_LIBRARY_LINK_C_FLAGS has -rdynamic, which
+          # with some tool chains (such as SUSE GCC 6.2) results in an executable that
+          # is not properly static. Adding --no-export-dynamic avoids the problem.
```
derekbruening, line 2638: How about a reference rather than full 3-line copy-and-paste; or a function/macro to share this common sequence.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1073,33 +1080,66 @@ template <typename T> class trace_converter_t {
...
-                        break;
+                if (instr->is_scatter_or_gather()) {
+                    // The instr should either load or store, but not both. Also,
+                    // it should have a single src or dest operand.
+                    DR_ASSERT(instr->num_mem_srcs() + instr->num_mem_dests() == 1);
+                    bool is_scatter = instr->num_mem_dests() == 1;
+                    bool reached_end_of_memrefs = false;
+                    // For expanded scatter/gather instrs, we do not have prior knowledge
+                    // of the number of store/load memrefs that will be present. So we
+                    // continue reading entries until we find a non-memref entry.
```
derekbruening, line 1091: Comment that this only works because the scatter/gather is the only instruction in the block: we assume that drx_expand_* isolates it to a block by itself.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1073,33 +1080,66 @@ template <typename T> class trace_converter_t {
...
+                    while (!reached_end_of_memrefs) {
+                        // XXX: Add sanity check for max count of store/load memrefs
+                        // possible for a given scatter/gather instr.
+                        if (is_scatter) {
+                            error =
+                                append_memref(tls, &buf, instr, instr->mem_dest_at(0),
```
derekbruening, line 1097: This is passing the same memref_summary_t for each one.  Are we sure that won't break something?    What if we add further fields in the future: would that ever trip over this corner case?  Worth comments here and in memref_summary_t.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1073,33 +1080,66 @@ template <typename T> class trace_converter_t {
...
-                    if (!error.empty())
-                        return error;
-                    error = handle_kernel_interrupt_and_markers(
-                        tls, &buf, cur_modoffs, instr->length(), instrs_are_separate,
-                        &interrupted);
-                    if (!error.empty())
-                        return error;
-                    if (interrupted)
-                        break;
+                if (instr->is_scatter_or_gather()) {
```
derekbruening, line 1083: This is a large increase in the size of this function.  Could we try one of these:

1) Split something out into a separate function.

2) Instead of this big separate if-else, try to have uniform loops somehow -- not sure it's better.

derekbruening (commented):
It seems cleaner to change the top-of-block PC instru to be an app instr than to add all these boolean parameters and special casing to instru_offline?

clients/drcachesim/tests/allasm-repstr-basic-counts.templatex
abhinav92003: > These statistics are on the final trace which must have an entry for each iteration of the loop:

Note that the `Total non-fetched instructions` stat is for instrs, not memrefs.

My understanding is that the final trace would contain only a single `rep movs` instr entry. The raw trace would contain multiple pc entries all pointing to the `rep movs` pc; all of these entries are initially marked as `TRACE_TYPE_INSTR_MAYBE_FETCH`). Later only one of these will be fetched while constructing the final trace [here](https://github.com/DynamoRIO/dynamorio/blob/dcbaca81b9816a40fddc2b0211475a7c5a9220a3/clients/drcachesim/tracer/raw2trace.h#L1035), and others will be marked as [`TRACE_TYPE_INSTR_NO_FETCH`](https://github.com/DynamoRIO/dynamorio/blob/c530f52e3797ece0ac023876b65e42a66657a8ad/clients/drcachesim/common/trace_entry.h#L164).

clients/drcachesim/tracer/instru_online.cpp
abhinav92003: Even if we call this only for app instrs, we'd still need `scatter_gather_expanded` and `repstr_expanded`, to be able to set `type` correctly.

Before this PR, we used `*bb_field != 0` to determine whether this bb is a repstr expansion, which obviated the need to pass in these booleans. But now that we have multiple type of expansions, we need to pass more data in (either as these boolean params, or maybe as a struct or something).

clients/drcachesim/tests/allasm-repstr-basic-counts.templatex
derekbruening: The final trace will have a separate instruction entry for each iteration of the rep string loop.  This is the whole point for the existence of "non-fetched" instructions.

clients/drcachesim/tracer/raw2trace.h
derekbruening: I'm worried about interactions with identify_elidable_addresses.  If the tracer is looking at the expanded scatter memrefs and decides to elide something, raw2trace has to match that.

clients/drcachesim/tracer/instru_offline.cpp
abhinav92003: That might introduce some more special cases in tracer.cpp.

Here are the first few instrs in the expansion:
```
mov     %rax -> %gs:0x000000e8[8byte] // first instr in expansion. the reg spill.
lahf    -> %ah
seto    -> %al
<label>
mov    %rcx -> %gs:0x000000f0[8byte]
<label>
mov    %rdx -> %gs:0x000000f8[8byte]
<label>
kmovw  %k1 -> %ecx // first app instr in expansion.
test   %ecx $0x00000001
jz     @0x00007f7aed7b8060[8byte]
...
```
I assume you mean that we should just skip instrumenting the first `mov` (the reg spill), and instrument `kmovw` instead (which is an app instr). It is indeed a suitable candidate as it's guaranteed to execute in 0-iter cases too. But, this'd mean sacrificing the always-instrument-first-instr invariant (look for `drmgr_is_first_nonlabel_instr` in tracer.cpp), which may get a bit tricky. I can try.

clients/drcachesim/tests/allasm-repstr-basic-counts.templatex
derekbruening: The point it, it's a conscious decision to have a separate instruction entry for each iteration of a rep string loop.  It is unrelated to how the raw trace is recorded or whether the instrumentation can easily avoid an instruction entry: post-processing could strip them all out if we wanted.  The final trace can look how we want it to look.

It also seems like a scatter/gather is basically a loop: it is not an atomic instruction.  It has an iteration over the mask bits.  If interrupted in the middle, it does not go back and re-execute the original instruction: it resumes in the middle with whatever last mask bit it finished with.

clients/drcachesim/tracer/instru_online.cpp
derekbruening: Looking: yes, online instru does special-case rep strings to deal with the pc+1 translation.  But it shouldn't.  That special case should disappear with #4865.

clients/drcachesim/tests/allasm-repstr-basic-counts.templatex
derekbruening: Note that we ended up with mixed feelings about the non-fetched instructions and whether they should be there or not.

For scatter-gather: is the expansion uniform?  Could it be re-written as a loop?  Would that shrink it significantly?

abhinav92003: Thanks for clarifying. I think I misunderstood what "non fetched instrs" meant (I see now that they're NOT "instr entries that are ignored while making the final trace"). I verified using `opcode_mix` and `view` that there are indeed 5 instances of `rep movs` in the final trace.

> It also seems like a scatter/gather is basically a loop: it is not an atomic instruction. It has an iteration over the mask bits. If interrupted in the middle, it does not go back and re-execute the original instruction: it resumes in the middle with whatever last mask bit it finished with.

So I assume we want the same behaviour for scatter/gather as what we have for rep stringop? -- as many scatter/gather instrs in the final trace as the number of set bits.

derekbruening: > So I assume we want the same behaviour for scatter/gather as what we have for rep stringop? -- as many scatter/gather instrs in the final trace as the number of set bits.

I'm not sure: even for rep string, for many cases we don't want the extra instructions.  E.g., perfctrs only count it as 1 instruction, and cache simulators should probably count it as 1 fetch.  Xref https://github.com/DynamoRIO/dynamorio/issues/2051#issuecomment-348601827.  Looking back, I kind of wish we had never made the "non fetched instrs" and had just one instr entry followed by just memrefs.

abhinav92003: Having a single instr and multiple memrefs makes sense to me -- that's the behaviour I assumed initially too. What was the rationale behind having multiple instances of the rep stringop in the instr stream?

derekbruening: I think originally it was easiest to implement with an instr per iter, and that matched other cache simulators like Cachegrind and Pin's icache tool -- but I think both of those do similar rep string rewrites into explicit loops and maybe don't have good reasons to have a fetch per iter.  Only later did we look at it more closely with core simulations and comparing dr$sim to a wider range of simulators.  We seem to have assumed that a core simulator would want the extra instruction entries: https://github.com/DynamoRIO/dynamorio/issues/2051#issuecomment-348713878.  But, I'm not sure that's really true.  I suspect we could remove them and core simulators would be fine.  The issue now would be backward compatiblity: we'd have to label it a breakage.

clients/drcachesim/tracer/instru_offline.cpp
derekbruening: Right, that tracer logic is complex.  This also interacts with #4865 where I was assuming I'd have to tackle that same logic and try to refactor it.

clients/drcachesim/tests/allasm-repstr-basic-counts.templatex
abhinav92003: I see. So this PR can do the right thing and let there be one instr entry + following memref(s).

The rep stringop case will need to be fixed. I'll probably create a new issue, or add it to the i#2985 task list. As it's a breaking change, I'd prefer to do it in new PR atleast.

clients/drcachesim/tracer/instru_offline.cpp
derekbruening: There does seem to be a gulf between the ideal of only app instrs being visible, or using first_app instead of first_nonlabel, and reality with messy sequences from app2app.

What if instead of passing in the scatter-gather the code instead checked for !instr_is_app and in that case it used the tag, for this PR: and assuming we can clean up the iteration in the tracer for #4865, maybe we can remove that later, but as it doesn't need extra bool params it feels like a cleaner less intrusive temporary (hopefully) solution.

clients/drcachesim/tests/allasm-repstr-basic-counts.templatex
derekbruening: Seems reasonable to file a new issue on re-considering "non-fetched instrs".  We would want to ask all concerned parties, including third parties running traces through simulators, for input to make sure it's an unused feature before removing it.

abhinav92003: > For scatter-gather: is the expansion uniform? 

Not sure what you mean by uniform?

> Could it be re-written as a loop? Would that shrink it significantly?

Initially I had thoughts of re-writing it as a loop, which would shrink the expansion size by a factor of number of store/load ops in the scatter/gather. 

But that may be independent of this discussion. To avoid multiple instr entries, we can find one instruction that's before the loop (in case of rep stringop, maybe we can add a dummy nop), and instrument that to output the pc entry.

abhinav92003: Though we may run into https://github.com/DynamoRIO/dynamorio/pull/4912#discussion_r633798543 if we instrument any non-app instr (like the dummy nop I suggested).

abhinav92003: > Not sure what you mean by uniform?

Oh you mean whether the instrs in each iteration are same so that it can be made into a loop. I haven't really done it yet, so not sure if there are any unknown barriers to doing that. I think we can do it independently of this PR. I'll add it to the i#2985 task list.

derekbruening: Right.  A complication is that the looping backward branch must be an app instr for DR to make it into a loop.  We could augment drmgr_fix_app_ctis() or sthg.

clients/drcachesim/tracer/raw2trace.h
abhinav92003: But this logic (of scanning memrefs following a scatter/gather instr until we reach a non-memref) doesn't really depend on scatter/gather being in its own block, does it? The important thing is that we can never have memrefs on their own, without being preceded by a pc entry, which is true always. So even if there were other app instrs following/before the scatter/gather expansion in the same block, the logic would still work.

derekbruening: > The important thing is that we can never have memrefs on their own, without being preceded by a pc entry, which is true always.

No, this is not true.  I think you are conflating raw traces with final traces.  Raw traces have a single PC entry for the whole block, followed by memref entries for all the instructions in the blcok, without any more PC entries in between.  That's why raw2trace has to go figure out what memref goes with which instruction.

clients/drcachesim/tracer/raw2trace.h
abhinav92003: > Split something out into a separate function.

Did this. It did help a bit I think.

clients/drcachesim/tracer/raw2trace.h
abhinav92003: I think I was. I understand now. Added  the comment about separate bb being important for this logic to work.

suite/tests/CMakeLists.txt

abhinav92003: Added a macro.

clients/drcachesim/tracer/instru_offline.cpp
abhinav92003: Dropped the idea of marking reg spill instrs as app.

clients/drcachesim/tracer/instru_offline.cpp
abhinav92003: Removed the two extra bool params from `instrument_instr`. Switched to `!instr_is_app` to check whether the given bb has scatter/gather expanded. Added an assert check to verify that `!instr_is_app` iff `ud->scatter_gather`.

abhinav92003: Also, to find whether any repstr was expanded, reverted to the old check.

clients/drcachesim/tracer/instru_online.cpp
abhinav92003: Reverted the newly added booleans. This PR now keeps the existing logic for repstr expansion.

clients/drcachesim/tracer/raw2trace.h
abhinav92003: > This is passing the same memref_summary_t for each one. Are we sure that won't break something? 

My rationale was that: if we had a separate pc entry preceding each memref, the `memref_summary_t` object for each memref would still be identical because it's based on the same instr at the same pc entry. So passing the same `memref_summary_t`  object should be equivalent.

> Worth comments here and in memref_summary_t.

Added.

abhinav92003: > I'm worried about interactions with identify_elidable_addresses

Good point. Can we just disable elision for scatter gather bbs?

clients/drcachesim/tests/allasm-repstr-basic-counts.templatex
abhinav92003: https://github.com/DynamoRIO/dynamorio/pull/4912#discussion_r634043217: reverted extra bools and switched to old logic for repstr detection and mitigation.

#4915 for changing behaviour for repstr to single instr entry.

clients/drcachesim/tracer/raw2trace.h
derekbruening: > My rationale was that: if we had a separate pc entry preceding each memref, the memref_summary_t object for each memref would still be identical because it's based on the same instr at the same pc entry. So passing the same memref_summary_t object should be equivalent.

It's based on the same instr right now in raw2trace b/c it's not expanding the scatter/gather.  But the instrumentation is not based on a single instr: it's based on a series of separate loads/stores.  And the elision is as well.  So we have a discrepancy and potential for problems.

> > I'm worried about interactions with identify_elidable_addresses
>
> Good point. Can we just disable elision for scatter gather bbs?

I think we should disable for now.  A possible solution is to call drx_expand_scatter_gather in raw2trace (making sure it can run statically); I would disable with an XXX comment about maybe doing this.  Are the addresses possibly reconstructable or are the index values too runtime-dependent?

clients/drcachesim/tracer/instru_offline.cpp
abhinav92003: Hmm.. found an issue when I fixed the assert statement to verify that `!app` happens if-and-only-if `scatter_gather`: `DR_ASSERT((ud->scatter_gather && !instr_is_app(instr)) ||
                  (!ud->scatter_gather && instr_is_app(instr)));`

Hitting the assert in `burst_malloc` for a non-app `nop` instr that is the first (and only) instruction in a bb.

derekbruening: Why do we care whether !app coincides with scatter-gather?  I don't think we do.

derekbruening: We are just assuming that !app corresponds to the first instr in the block, and we assume the app pc version of the tag corresponds to that.

abhinav92003: I was hoping that `!instr_is_app <=> ud->scatter_gather`: this is a stricter assumption that would've increased my confidence in the temporary workaround we're using. But it seems that `!instr_is_app` also happens in case of the `nop` at `replace_retaddr_sentinel`.

Anyway, I replaced the assert with: `instr_is_app || drmgr_is_first_nonlabel_instr` which should work for now. As discussed, for `!instr_is_app`, we get pc from the bb tag, and also decode pc to get the instr's size.

clients/drcachesim/tracer/raw2trace.h
abhinav92003: Disabled address elision for scatter/gather bbs.

abhinav92003: `vpscatterdd [arr + xmm11*4]{k1}, xmm10`

The memrefs do share `arr`, but the index values in `xmm11` could be anything.

clients/drcachesim/tracer/instru_offline.cpp
```diff
@@ -733,8 +740,16 @@ offline_instru_t::label_marks_elidable(instr_t *instr, OUT int *opnd_index,
 
 void
 offline_instru_t::identify_elidable_addresses(void *drcontext, instrlist_t *ilist,
-                                              int version)
-{
+                                              int version, bool has_scatter_gather_instr)
+{
+    // XXX: We turn off address elision for bbs containing scatter/gather. As
```
derekbruening, line 745: How about generalizing: check for emulation labels and disallow inside any emulation region.  Then remove the extra bool parameters.  Any emulation will have this same issue, not just scatter/gather.

For the call from raw2trace: maybe add an explicit check for an (unexpanded) scatter/gather or rep-string in the loop below?  Instead of the caller figuring it out and passing in a bool.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -849,16 +863,26 @@ template <typename T> class trace_converter_t {
         // state needed to reconstruct elided addresses.
         instrlist_t *ilist = instrlist_create(dcontext_);
         app_pc pc = start_pc;
+        bool has_scatter_gather_instr = false;
         for (uint count = 0; count < instr_count; ++count) {
             instr_t *inst = instr_create(dcontext_);
             app_pc next_pc = decode(dcontext_, pc, inst);
+#ifdef X86
+            // NYI on ARM/AArch64.
```
derekbruening, line 871: Is there an issue to reference?

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -958,6 +982,21 @@ template <typename T> class trace_converter_t {
...
+                   OUT bool *interrupted)
+    {
+        std::string error = append_memref(tls, buf_in, instr, memref, write, reg_vals,
+                                          reached_end_of_memrefs);
+        if (!error.empty())
+            return error;
+        error = handle_kernel_interrupt_and_markers(
+            tls, buf_in, cur_modoffs, instr->length(), instrs_are_separate, interrupted);
+        return error;
+    }
```
derekbruening, line 999: nit: Normally we separate functions with blank lines.  I thought clang-format enforced that?

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1073,33 +1112,55 @@ template <typename T> class trace_converter_t {
...
-                        break;
+                if (instr->is_scatter_or_gather()) {
+                    // The instr should either load or store, but not both. Also,
+                    // it should have a single src or dest operand.
+                    DR_ASSERT(instr->num_mem_srcs() + instr->num_mem_dests() == 1);
+                    bool is_scatter = instr->num_mem_dests() == 1;
+                    bool reached_end_of_memrefs = false;
+                    // For expanded scatter/gather instrs, we do not have prior knowledge
+                    // of the number of store/load memrefs that will be present. So we
+                    // continue reading entries until we find a non-memref entry.
```
derekbruening, line 1123: Is it possible there's a marker in the middle, or some other non-memref, that causes a false break?  I suppose handle_kernel_interrupt_and_markers() should consume all the markers: though IIRC it bails back to the main loop in some cases.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1193,12 +1205,18 @@ event_app_instruction(void *drcontext, void *tag, instrlist_t *bb, instr_t *inst
...
         adjust = instrument_instr(drcontext, tag, ud, bb, instr, reg_ptr, adjust, instr);
     }
     ud->last_app_pc = instr_get_app_pc(instr);
 
-    if (is_memref) {
+    // For the scatter/gather expansion, the first instr turns out to be a non-app
+    // memref (a reg spill). So, we reach here for a non-app instruction and we
+    // don't want to add a memref for that.
+    bool is_first_instr_in_scatter_gather_expansion =
+        ud->scatter_gather && drmgr_is_first_nonlabel_instr(drcontext, instr);
```
derekbruening, line 1218: Should this also check it's a non-app instr?  What if something changes (the expansion or the logic here to skip non-app instrs up above) -- we won't want to skip an app instr here.  This could would ideally be changed alongside a change like that: maybe a xref to #4865.

derekbruening (approved):

clients/drcachesim/tracer/raw2trace.h
abhinav92003: From my reading, there are two cases where `handle_kernel_interrupt_and_markers` bails back to the main loop.
- when there's an interrupt marker for some other module offset. This doesn't seem like it could happen in the middle of the scatter/gather memrefs, as otherwise the module offset would be the current one's (in which case, it would be processed by `handle_kernel_interrupt_and_markers`) https://github.com/DynamoRIO/dynamorio/blob/dcbaca81b9816a40fddc2b0211475a7c5a9220a3/clients/drcachesim/tracer/raw2trace.h#L1184
- or there's markers other than the interrupt marker. https://github.com/DynamoRIO/dynamorio/blob/dcbaca81b9816a40fddc2b0211475a7c5a9220a3/clients/drcachesim/tracer/raw2trace.h#L1192
I had a look at the marker list, and I don't think any of those can occur in the middle of scatter/gather memrefs: 
kernel xfer is placed only before sigret, there are many that occur only in header (timestamp, cpu Id, file type), func tracing related ones can't happen in the middle of an instr, instr count comes only in footer.

clients/drcachesim/tracer/instru_offline.cpp
abhinav92003: Removed the extra boolean param and added a check for emulation labels (for the call from instru_offline), and a check for scatter/gather/repstr (for the call from raw2trace).

I had to add a new loop, because the current one looks only at app instrs, and the emulation labels are non-app. Is it worth avoiding this extra loop during raw2trace, where it would be sufficient to look at the app instrs? I guess raw2trace overhead isn't as important as that of instru_offline, so maybe it's okay to have the extra loop.

clients/drcachesim/tracer/raw2trace.h
abhinav92003: i#3837. Added.

clients/drcachesim/tracer/raw2trace.h
abhinav92003: clang-format didn't throw any error. I fixed it though.

clients/drcachesim/tracer/tracer.cpp
abhinav92003: Done.

abhinav92003:
Made some changes to resolve comments. PTAL?

clients/drcachesim/tracer/instru_offline.cpp
derekbruening: raw2trace overhead does matter: we had to put effort into the various caching and parallelism schemes.  But this info collected is cached, right?  So this would just be called once per static block (not per execution), and there it shouldn't matter.

clients/drcachesim/tracer/raw2trace.h
derekbruening: OK.

clients/drcachesim/tracer/instru_offline.cpp
abhinav92003: Yes, it seems to be cached. `identify_elidable_addresses` is called only once per bb in instru_offline, and the `analyze_elidable_addresses` call in raw2trace is made only if instr summary doesn't exist already.

PR #4917, opened 2021-05-24
i#3823 multi-phase drreg: Avoid aflags slot conflict

abhinav92003:
Extends drreg support for multi-phase use by avoiding aflags spill slot conflict across multiple phases.

Removes existing restriction of aflags spill slot always being the slot 0. Allows selection of aflags spill slot at spills, and releasing of the slot at unreservations. Unlike gprs, we need to reset `pt->aflags.slot` everytime we release the slot, because this field is used to determine whether we are spilling/restoring aflags in various routines.

Adjusts fault restoration logic in `drreg_event_restore_state_without_ilist` to detect aflags spill slot dynamically. This is not perfect as we need the ilist for more robust restoration, which is done in `drreg_event_restore_state_with_ilist` if ilist is available.

Adds drreg test to verify absence of aflags spill slot conflict when aflags are reserved in app2app and insertion phases.

Also adds drreg test to verify correct restoration of aflags on fault, for various cases: when the spill regions are nested, when they overlap (but are not nested), when the aflags value is restored once before the fault, when the native aflags value is spilled to multiple slots with overlapping spill regions, and tests for some restoration cases when the faulting fragment's ilist isn't available.

Issue: #3823
Issue: #2985
Fixes: #3823


abhinav92003:
I'm still trying to root-cause the `drcachesim.filter-simple` test timeout on Windows 32-bit CI. The issue seems to come up only with the `-L0_filter` option. From the log files that I see on the GA test runner, it seems that the test gets stuck somewhere. The last log I see is from drreg, in a fragment with an expanded rep str instr. Maybe the condition codes are getting messed up somehow and the rep str loop goes on infinitely? But I would expect to see the "after instrumentation" log atleast in that case.

abhinav92003:
Huh.. the `filter-simple` failure went away this time: https://github.com/DynamoRIO/dynamorio/pull/4917/checks?check_run_id=2853187626

ext/drreg/drreg.c
```diff
@@ -1217,9 +1233,11 @@ drreg_statelessly_restore_app_value(void *drcontext, instrlist_t *ilist, reg_id_
...
 #ifdef X86
     if (reg != DR_REG_NULL && pt->aflags.xchg == reg) {
-        pt->slot_use[AFLAGS_SLOT] = DR_REG_XAX; /* appease assert */
-        restore_reg(drcontext, pt, DR_REG_XAX, AFLAGS_SLOT, ilist, where_respill, false);
-        pt->slot_use[AFLAGS_SLOT] = DR_REG_NULL;
+        ASSERT(pt->aflags.slot != MAX_SPILLS, "Aflags slot not reserved");
+        pt->slot_use[pt->aflags.slot] = DR_REG_XAX; /* appease assert */
+        restore_reg(drcontext, pt, DR_REG_XAX, pt->aflags.slot, ilist, where_respill,
+                    false);
+        pt->slot_use[pt->aflags.slot] = DR_REG_NULL;
```
derekbruening, line 1240: Should this be calling reset_aflags_spill_slot?  If not please comment as to why it's different from the other cases.

ext/drreg/drreg.c
```diff
@@ -1292,7 +1310,9 @@ drreg_reservation_info(void *drcontext, reg_id_t reg, opnd_t *opnd OUT,
     };
     per_thread_t *pt = get_tls_data(drcontext);
     drreg_status_t res;
-    if (reg < DR_REG_START_GPR || reg > DR_REG_STOP_GPR || !pt->reg[GPR_IDX(reg)].in_use)
+    if (!(reg >= DR_REG_START_GPR && reg <= DR_REG_STOP_GPR &&
+          pt->reg[GPR_IDX(reg)].in_use) &&
+        !(reg == DR_REG_NULL && pt->aflags.in_use))
```
derekbruening, line 1315: Please update the docs, which today only say that DR_REG_NULL for flags can be passed to the _ex version, not the base version.

derekbruening: Also, if this now matches the _ex version, this check can be removed and the check in _ex relied upon.

ext/drreg/drreg.c
```diff
@@ -2024,6 +2095,11 @@ drreg_event_restore_state_with_ilist(void *drcontext, bool restore_memory,
      * that do not have their native app value.
      */
     reg_id_t spill_slot_to_reg[MAX_SPILLS];
+    /* Note that we do not need to track multiple spill regs for aflags. On x86
+     * aflags can be spilled only to xax using lahf. On AArchXX, while they can
+     * be read into any register using mrs, but they are always spilled to a
```
derekbruening, line 2100: grammar: s/but// (or s/while//)

suite/tests/client-interface/drreg-test.c
```diff
@@ -67,6 +67,18 @@ void
 test_asm_faultK();
 void
 test_asm_faultL();
+void
+test_asm_faultM();
```
derekbruening, line 71: Should these have descriptive names rather than just letters, with so many?

suite/tests/client-interface/drreg-test.c
```diff
@@ -431,6 +543,77 @@ main(int argc, const char *argv[])
         test_asm_faultL();
     }
 
+    /* For some aflags restore tests below we do not use SIGILL to raise the
+     * fault. This is because the undefined instr on AArchXX is assumed to
+     * read aflags, and therefore restores aflags automatically. So the
```
derekbruening, line 548: Hmm but this is because the decoder is incomplete?  Seems worth mentioning in the issue on that so ARM is aware, and xref the issue number here.

suite/tests/client-interface/drreg-test.c
```diff
@@ -1254,6 +1463,414 @@ GLOBAL_LABEL(FUNCNAME:)
...
+        jmp      test24
+     test24:
+        mov      TEST_REG_ASM, DRREG_TEST_24_ASM
+        mov      TEST_REG_ASM, DRREG_TEST_24_ASM
+        mov      ah, DRREG_TEST_AFLAGS_ASM
+        sahf
+
+        mov      TEST_REG2_ASM, 1
+        /* Read aflags so that it is restored once. */
+        lahf
```
derekbruening, line 1630: Instead of the app using LAHF and SAHF in all of these tests, it seems better to use instructions that read/write OF as well in at least some of these tests, to exercise that code path?

derekbruening (approved):

ext/drreg/drreg.c
abhinav92003: No, because we don't want to modify state in `drreg_statelessly_restore_app_value`; I also see that we pass `release=false` to `restore_reg` for the same reason probably.

Sure, I'll add a comment.

derekbruening: But then slot_use and aflags.slot don't match.

But wait -- this is all inside "pt->aflags.xchg == reg" -- so is there a slot in the first place?  You added an assert there: does it not fire?  Looks like there should *not* be a slot, and it temporarily clobbers the slot to appease restore_reg asserts, and this set to null is just undoing that temp clobber.  I misread it as a true clear, which everywhere else uses the new helper.

derekbruening: Something seems wrong: either the assert, or the code which clobbers what the assert is asserting.

abhinav92003: Yeah I was just looking at that to confirm -- it does seem weird. I checked that none of our current tests exercise this path.

abhinav92003: So it is possible for `pt->aflags.xchg == xax`  (aflags in xax) **and** `pt->aflags.slot != max_spills` (aflags spilled to slot) to be true at the same time. This happens as an intermediate state in `drreg_statelessly_restore_app_value`, when it is invoked for xax holding the spilled aflags value. When this happens, it invokes `drreg_move_aflags_from_reg` with `stateful=false`. This spills aflags to some slot at `where_restore`, but doesn't clear `pt->aflags.xchg`. And then in the `restore_reg` call, aflags is spilled again to xax at `where_respill`.

What we want for this PR is to simply add the extra book-keeping required for aflags slot -- change the `restore_reg` call to release the slot, and also call `reset_aflags_spill_slot`.

Added a test that invokes this path.

ext/drreg/drreg.c
abhinav92003: > Please update the docs

Done

> Also, if this now matches the _ex version, this check can be removed and the check in _ex relied upon.

The _ex version has a different check. Unlike this one, it doesn't seem to care whether the queried reg is actually reserved or not. So keeping this check separate.

ext/drreg/drreg.c
abhinav92003: Done.

suite/tests/client-interface/drreg-test.c
abhinav92003: Done. Added the comment on i#4849 too.

suite/tests/client-interface/drreg-test.c
abhinav92003: Done.

suite/tests/client-interface/drreg-test.c
abhinav92003: Done.

abhinav92003:
Made some other small changes to tests -- used named constants for instrumentation markers; in nested spill region tests, made spill regions **strictly** nested (seemed cleaner than overlapping beginning/end instrs); and removed a wrong drreg use (aflags write without reservation) in `test_asm_fault_restore_multi_phase_native_aflags_spilled_twice`.

ext/drreg/drreg.c
```diff
@@ -68,14 +68,6 @@
 /* This should be pretty hard to exceed as there aren't this many GPRs */
 #define MAX_SPILLS (SPILL_SLOT_MAX + 8)
 
-/* We store data about spill slot usage in the data area of a label instr. */
-#define SPILL_SLOT_ID_LABEL_INSTR_DATA_AREA_OFFSET 0
```
derekbruening: This was separately removed: looks like you merged w/ master, and the "View changes since last review" is showing these unrelated changes?  It used to be that nearly all code review systems would do that and it was best to never merge until after the final review was done; in the just the last couple of years code review systems got better at hiding unrelated changes, but that seems to have failed here...not sure how to get a view of just new changes -- do you see a way to do it?

abhinav92003: I merged early this time because I wanted to test with changes from a related master commit. But I probably should have done that locally and waited for the review to complete before pushing it.

> not sure how to get a view of just new changes -- do you see a way to do it?

Only idea I have is looking at individual commits from the last review. But nothing that would put all changes in one view, minus the master merge noise.

ext/drreg/drreg.c
```diff
@@ -1173,11 +1173,16 @@ drreg_statelessly_restore_app_value(void *drcontext, instrlist_t *ilist, reg_id_
          */
 #ifdef X86
     if (reg != DR_REG_NULL && pt->aflags.xchg == reg) {
+        /* drreg_restore_app_value invokes drreg_move_aflags_from_reg with stateful=false
+         * above. This spills aflags to some slot, regardless of whether they are live or
+         * not. Also, even though aflags are spilled to a slot, pt->aflags.xchg is not
+         * cleared. Below, we restore aflags to xax, and forget the spill slot.
+         */
         ASSERT(pt->aflags.slot != MAX_SPILLS, "Aflags slot not reserved");
         pt->slot_use[pt->aflags.slot] = DR_REG_XAX; /* appease assert */
```
derekbruening, line 1182: Why is this needed?  spil_reg set slot_use for us -- at least at the where_restore point.

ext/drreg/drreg.c
```diff
@@ -1173,11 +1173,16 @@ drreg_statelessly_restore_app_value(void *drcontext, instrlist_t *ilist, reg_id_
...
         restore_reg(drcontext, pt, DR_REG_XAX, pt->aflags.slot, ilist, where_respill,
-                    false);
-        pt->slot_use[pt->aflags.slot] = DR_REG_NULL;
+                    /*release=*/true);
+        reset_aflags_spill_slot(pt);
```
derekbruening, line 1185: This doesn't make sense to me: this is a stateless change; there is an assert on line 1181 that aflags.slot is set, so clearing it here will cause that to fire if this routine is called again.

Also, passing release=true is duplicating the clear of slot_use in reset_aflags_spill_slot.  And again, it feels wrong to have release=true for a stateless change.  I'm not sure I remember all the invariants: but shouldn't slot_use[aflags.slot] always be set if aflags are reserved and we should not give up that slot, or else someone else could take it and we'd have aflags.slot pointing to someone else's slot?  If for aflags-in-xax we don't normally have a slot: then why is the assert at line 1181 there?

suite/tests/client-interface/drreg-test.c
```diff
@@ -44,41 +44,41 @@
...
+test_asm_fault_restore_aflags_in_slot();
 void
-test_asm_faultC();
+test_asm_fault_restore_ignore_3rd_dr_tls_slot();
 void
-test_asm_faultD();
+test_asm_fault_restore_non_public_dr_slot();
 void
-test_asm_faultE();
+test_asm_fault_restore_non_public_dr_slot_rip_rel_addr_in_reg();
```
derekbruening, line 55: I like these names much better than the letters.

ext/drreg/drreg.c
```diff
@@ -1173,11 +1173,16 @@ drreg_statelessly_restore_app_value(void *drcontext, instrlist_t *ilist, reg_id_
          */
 #ifdef X86
     if (reg != DR_REG_NULL && pt->aflags.xchg == reg) {
+        /* drreg_restore_app_value invokes drreg_move_aflags_from_reg with stateful=false
+         * above. This spills aflags to some slot, regardless of whether they are live or
+         * not. Also, even though aflags are spilled to a slot, pt->aflags.xchg is not
+         * cleared. Below, we restore aflags to xax, and forget the spill slot.
```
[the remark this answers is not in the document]
abhinav92003: Isn't it unusual that both `pt->aflags.xchg != DR_REG_NULL` and `pt->aflags.slot != MAX_SPILLS` here? I thought that warranted some explanation, that the spill slot is used only for the duration of this routine.

ext/drreg/drreg.c
abhinav92003: > there is an assert on line 1181 that aflags.slot is set, so clearing it here will cause that to fire if this routine is called again.

Note that when we enter `drreg_statelessly_restore_app_value`, `aflags.slot` is **not** already set. It is set by `drreg_restore_app_value` on L1166, and then cleared at L1185. So, in the second call to `drreg_statelessly_restore_app_value`, it will be spilled again before we get to this assert.

I modified the test to call `drreg_statelessly_restore_app_value` twice consecutively, and it worked fine. In the bb, I see two pairs of spills/restores, as expected:

```
 +28   L3 @0x00007fa0fc407cf8  48 c7 c3 01 00 00 00 mov    $0x0000000000000001 -> %rbx
 +35   m4 @0x00007fa0fc407fe0  65 48 a3 e8 00 00 00 mov    %rax -> %gs:0x000000e8[8byte]
                               00 00 00 00
 +46   m4 @0x00007fa0fc408060  65 48 a1 e8 00 00 00 mov    %gs:0x000000e8[8byte] -> %rax
                               00 00 00 00
 +57   m4 @0x00007fa0fc3a71a8  65 48 a3 e8 00 00 00 mov    %rax -> %gs:0x000000e8[8byte]
                               00 00 00 00
 +68   m4 @0x00007fa0fc3a84d0  65 48 a1 e8 00 00 00 mov    %gs:0x000000e8[8byte] -> %rax
                               00 00 00 00
 +79   L3 @0x00007fa0fc407d78  48 c7 c3 02 00 00 00 mov    $0x0000000000000002 -> %rbx
```

Will update the test in the next push.

> And again, it feels wrong to have release=true for a stateless change.

We are not modifying any relevant drreg state (state that existed before the call to `drreg_statelessly_restore_app_value`). We are not releasing any slot that was in use prior to the call. The spill of aflags to a slot and its subsequent restore are local to `drreg_statelessly_restore_app_value` and is required so that we can restore xax between `where_restore` and `where_respill`.

> Also, passing release=true is duplicating the clear of slot_use in reset_aflags_spill_slot. 

Yes, passing `release=true` is redundant. But that's true for today. Tomorrow, if someone adds additional bookkeeping to update in restore_reg, we want this to still work.  As we really do intend to release the slot here, I figured its better to do the redundant thing.

> shouldn't slot_use[aflags.slot] always be set if aflags are reserved and we should not give up that slot, or else someone else could take it and we'd have aflags.slot pointing to someone else's slot

If aflags are in xax, then we don't want to hold any slot (as an optimisation). In this case, `aflags.slot = MAX_SPILLS` so we wouldn't try to read that slot.

Q: Based on the documentation of `drreg_statelessly_restore_app_value`,  I assumed that there are no app instrs between `where_respill` and `where_restore`. Is that right?
```
This routine is meant for use with instrumentation that uses separate control flow
 * paths, such as a fastpath and a slowpath, where the slowpath needs access to the
 * full application state yet must retain scratch register parity with the fastpath.
 * 
```

If not then the aflags slot used temporarily here may be overwritten by instrumentation added on the intervening app instr. Which means it's better to lock in a slot for aflags whenever it is reserved.

> If for aflags-in-xax we don't normally have a slot: then why is the assert at line 1181 there?

This is normally true, yes. But (as described above) in case of `drreg_statelessly_restore_app_value`, we have `aflags.xchg=DR_REG_XAX` (even though xax holds its native value at this point) and `aflags.slot!=MAX_SPILLS` both. The slot was reserved by `drreg_statelessly_restore_app_value` at L1166. This seemingly inconsistent intermediate state is local to `drreg_statelessly_restore_app_value`.

suite/tests/client-interface/drreg-test.c
abhinav92003: Yes; made verification easier too.

ext/drreg/drreg.c
abhinav92003: was trying not to touch this code, thinking that there may be some corner case that needs it. But I don't see why this assert was needed in the first place. I removed it in a clean client, and it didn't hit any assert.

ext/drreg/drreg.c
derekbruening: > Note that when we enter drreg_statelessly_restore_app_value, aflags.slot is not already set. It is set by drreg_restore_app_value on L1166

Ah, that's what I was missing.  So if aflags.xchg is set and aflags are in xax, we don't have a slot reserved.  Let me re-examine the code w/ that info.

ext/drreg/drreg.c
abhinav92003: Oh, I think some assert may be hit if aflags were spilled using some reg other than xax (which may happen if xax was already spilled before we reserved aflags). I can write a test for this, and maybe workaround so that we always spill from xax (otherwise the restore logic also becomes more complicated).

abhinav92003: This is the case I was referring to in previous comment: if aflags were spilled using some other reg: https://github.com/DynamoRIO/dynamorio/blob/27e2f830e209d3c2aca725628b6b453ed5d6ad68/ext/drreg/drreg.c#L1441

ext/drreg/drreg.c
derekbruening: > Q: Based on the documentation of drreg_statelessly_restore_app_value, I assumed that there are no app instrs between where_respill and where_restore. Is that right?

This was added primarily for Dr. Memory for the still-not-finished attempt to replace its existing scratch register management system with drreg.  Here are the two calls to this API function:
https://github.com/DynamoRIO/drmemory/blob/i1795-use-drreg/drmemory/slowpath.c#L1084

The post_jmp there is across a "lean call".  So no app instrs in between.  I think most use cases would be similar: internal control flow added for different instrumentation cases, all handling a single app instr,

ext/drreg/drreg.c
derekbruening: Maybe spill_reg wasn't called at the time this assert was added.

ext/drreg/drreg.c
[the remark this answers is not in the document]
derekbruening: Yes, combined w/ the conversation below, I understand now.  I think the wording here could have helped me: maybe reword explicitly say that the spill slot is temporary and local to this function?

ext/drreg/drreg.c
derekbruening: For the xax_swap case: you mean the lazy unrestore could have that slot still holding the other reg.  But...it was spilled to a non-aflags slot, so should that xax_swap code go clear the aflags slot?

derekbruening (approved):

ext/drreg/drreg.c
abhinav92003: > Oh, I think some assert may be hit if aflags were spilled using some reg other than xax

Actually, this cannot happen. The `drreg_statelessly_restore_app_value` aflags spill via `drreg_restore_app_value(xax)` is always done using xax:
https://github.com/DynamoRIO/dynamorio/blob/27e2f830e209d3c2aca725628b6b453ed5d6ad68/ext/drreg/drreg.c#L1355

ext/drreg/drreg.c
abhinav92003: I'm still a bit worried about the tmp aflags slot colliding with some drreg reservation by instrumentation between `where_restore` and `where_respill`, which seems possible right? Or should we just document that there shouldn't be any?

abhinav92003:
Add another test: for the case when xax is already reserved and in-use when we try to reserve aflags. In this case, the existing code used a swap xax reg to write aflags to slot. Changed it so that it's always xax that is used to write aflags to slot -- which is simpler for the restore logic. Also added a test for this case, which failed before the fix.

ext/drreg/drreg.c
abhinav92003: > For the xax_swap case: you mean the lazy unrestore could have that slot still holding the other reg. But...it was spilled to a non-aflags slot, so should that xax_swap code go clear the aflags slot?

Not sure I understand this scenario. 

Just to be clear, the `xax_swap` case (which happens when aflags are being spilled at a time when xax is already reserved and in-use) and the mysterious "appease assert" statement are not related (they seemed to be to me initially, but as I said in my last comment, they aren't).

Still can't figure out why the "appease assert" stmt was needed; based on what I read in code, it doesn't seem like it could be an issue. Removed it.

ext/drreg/drreg.c
abhinav92003: I mean, should we just document that a client shouldn't add any drreg reservation between `where_restore` and `where_respill`?

derekbruening: It is up to the client to keep the reg reserved w/ a single reservation across the two points.  Could make that clearer in the docs.

abhinav92003: Added comment in drreg.h also to describe this limitation.

I agree on just documenting this limitation. The other option would be to let go of this optimisation of aflags not reserving a spill slot when not needed, which could be expensive.

ext/drreg/drreg.c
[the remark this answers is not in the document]
abhinav92003: Done.

abhinav92003:
@derekbruening Is this PR good to go? Just making sure, as there are a couple open comment threads and some new commits (Apologies for the multiple round of reviews).

derekbruening:
I thought I had approved it after the 2nd review?  Was there a 3rd review request?  I must have missed that.  Looking now.

ext/drreg/drreg.c
```diff
@@ -1173,13 +1173,26 @@ drreg_statelessly_restore_app_value(void *drcontext, instrlist_t *ilist, reg_id_
...
-         * not. Also, even though aflags are spilled to a slot, pt->aflags.xchg is not
-         * cleared. Below, we restore aflags to xax, and forget the spill slot.
+        ASSERT(reg == DR_REG_XAX, "xax is the only x86 reg that may have spilled aflags");
+        /* If reg has aflags, they need to be spilled from reg to slot at `where_restore`,
+         * and then restored from the slot to reg at `where_respill`. This is done locally
+         * to this routine, and doesn't affect any existing drreg state (so this routine
+         * continues to be 'stateless'). This spill slot is used only between
+         * `where_restore` and `where_respill`; we assume there's no new drreg reservation
+         * by the client in between these, which could spill to this temp slot and clobber
+         * native aflags value. The alternative to this limitation is to keep a slot aside
```
derekbruening, line 1183: nit: s/native/the native/

ext/drreg/drreg.c
```diff
@@ -1173,13 +1173,26 @@ drreg_statelessly_restore_app_value(void *drcontext, instrlist_t *ilist, reg_id_
...
+         * for aflags whenever aflags are reserved (currently we do not reserve any spill
+         * slot for aflags when they can be stored in xax), which may reduce number of
```
derekbruening, line 1185: nit: s/number/the number/

ext/drreg/drreg.c
```diff
@@ -1462,14 +1475,18 @@ drreg_spill_aflags(void *drcontext, instrlist_t *ilist, instr_t *where, per_thre
...
-        PRE(ilist, where,
-            INSTR_CREATE_xchg(drcontext, opnd_create_reg(xax_swap),
-                              opnd_create_reg(DR_REG_XAX)));
         ASSERT(pt->aflags.slot == MAX_SPILLS, "aflags slot not reset");
         if (pt->aflags.slot == MAX_SPILLS) {
             pt->aflags.slot = find_free_slot(drcontext, pt, ilist, where);
         }
-        spill_reg(drcontext, pt, xax_swap, pt->aflags.slot, ilist, where);
+        /* We want to use xax only to spill aflags. This is to make drreg state
+         * restoration a bit simpler. So, we first spill aflags to slot using xax,
```
derekbruening, line 1483: nit: a/slot/a slot/

ext/drreg/drreg.h
```diff
@@ -488,6 +488,11 @@ DR_EXPORT
  * whether instructions were inserted at \p where_restore and \p where_respill,
  * respectively.
  *
+ * For correct operation on x86 in the case when aflags are in xax and this routine
+ * is invoked to get app value of xax, there shouldn't be any new reservation between
+ * \p where_restore and \p where_respill that may write to a spill slot and clobber
+ * the tmp slot used in this routine.
```
derekbruening, line 494: nit: s/tmp/temporary/

suite/tests/client-interface/drreg-test.c
```diff
@@ -255,6 +257,17 @@ handle_signal12(int signal, siginfo_t *siginfo, ucontext_t *ucxt)
     SIGLONGJMP(mark, 1);
 }
 
+static void
+handle_signal13(int signal, siginfo_t *siginfo, ucontext_t *ucxt)
```
derekbruening, line 261: Should each numeric suffix on the handlers be replaced with the descriptive suffix for the corresponding test_ routine?

derekbruening (approved):

suite/tests/client-interface/drreg-test.c
abhinav92003: Done. Some names could be a bit long because many signal handlers handle two tests.

ext/drreg/drreg.h
abhinav92003: Done.

ext/drreg/drreg.c
abhinav92003: Done.

ext/drreg/drreg.c
abhinav92003: Done.

ext/drreg/drreg.c
abhinav92003: Done.

Issue #2417, opened 2017-05-05
Fix flaky tests on AArch64


fhahn:
Currently, the following tests are flaky on some AArch64 hardware

     code_api|pthreads.pthreads_fork_FLAKY
     code_api|linux.sigsuspend
     code_api|pthreads.pthreads_exit
     code_api|security-common.TestMemProtChg_FLAKY
     code_api|pthreads.ptsig_FLAKY

egrimley:
I have the impression that `code_api|security-common.TestMemProtChg_FLAKY` isn't "flaky": it seems to pass or fail consistently depending on the system.

johnfxgalea:
Ref:  security-common.TestMemProtChg_FLAKY failed:
https://github.com/DynamoRIO/dynamorio/pull/3893/checks

AssadHashmi:
The following tests fail on the new Jenkins CI host, an Ampere Altra:

code_api|tool.drcachesim.phys #4922 
code_api|tool.drcacheoff.rseq #4924 
code_api|api.rseq #4923 
code_api|tool.drcachesim.threads #4928 
code_api|tool.drcachesim.TLB-threads #4928

joshua-warburton:
Additionally, drcachesim.coherence times out on many different aarch64 hosts. It intermittently times out in the precommit of this repo and seems to always time out on our aws-based ci.

derekbruening:
The drcachesim.coherence test, like #4928 and #4954, only times out in release build

derekbruening:
drcachesim.coherence is fixed by the same fix for #4928: PR #5375

xdje42:
Data point: code_api|tool.drcachesim.TLB-threads failed here:
https://github.com/DynamoRIO/dynamorio/actions/runs/10929966711/job/30341854185#step:7:49658

joshua-warburton:
> > I've just observed `code_api|tool.drcachesim.TLB-threads` failing with this output:
> > ```
> > 242:     ...................................................................
> > 242: Invalid trace entry type thread_exit (23) before a bundle
> > 242: drmemtrace_launcher: /.../dynamorio/clients/drcachesim/reader/reader.cpp:262: virtual bool dynamorio::drmemtrace::reader_t::process_input_entry(): Assertion `type_is_instr(cur_ref_.instr.type) || cur_ref_.instr.type == TRACE_TYPE_INSTR_NO_FETCH' failed.
[...]

Issue #4924, opened 2021-05-26
Fix tool.drcacheoff.rseq test regression


AssadHashmi:
Migrating the AArch64 Jenkins CI to an Ampere Altera machine caused a regression in the `tool.drcacheoff.rseq` test:

`  *** postcmd failed (1): ERROR: Usage error: Unknown option: -test_mode`

AssadHashmi:
The usage error above has been fixed by #4929 but the test now fails with `failed to match expected output`.

derekbruening:
I believe GA CI Linux is hitting this same failure on Ubuntu 20.04 in release x86-64, and this is blocking our upgrade.

derekbruening:
#4953 is the issue covering the upgrade

derekbruening:
It looks like it's just not passing -test_mode in release and that's why it
doesn't print that it passed the invariants.
I think the only reason is I was using assert(), but no more after PR #5077.

Xref where I was going to add a -simulator_type=invariant_checker for #5076.
But here we want to run two simulators at once: so leave the -test_mode for
that and just enable for release build?  Slight extra overhead to check for
the flag in analyzer_multi but should be fine.

PR #4929, opened 2021-05-27
i#2417 Suppress regressions caused by CI migration

AssadHashmi:
The following tests fail on the new Jenkins CI host, an Ampere Altra:
code_api|tool.drcachesim.phys            i4922
code_api|tool.drcacheoff.rseq            i4924
code_api|api.rseq                        i4923
code_api|tool.drcachesim.threads         i4928
code_api|tool.drcachesim.TLB-threads     i4928

This patch adds them to the 'ignore' list for now.
Issues have been raised to track fixes:

Issue: #2417

### DIFF
--- a/suite/runsuite_wrapper.pl
+++ b/suite/runsuite_wrapper.pl
@@ -313,6 +313,11 @@ for (my $i = 0; $i <= $#lines; ++$i) {
                                    'code_api|linux.fib-conflict' => 1,
                                    'code_api|linux.fib-conflict-early' => 1,
                                    'code_api|linux.mangle_asynch' => 1,
+                                   'code_api|tool.drcachesim.phys' => 1, # i#4922
+                                   'code_api|tool.drcacheoff.rseq' => 1, # i#4924
+                                   'code_api|api.rseq' => 1, # i#4923
+                                   'code_api|tool.drcachesim.TLB-threads' => 1, # i#4928
+                                   'code_api|tool.drcachesim.threads' => 1, # i#4928
                                    'code_api,tracedump_text,tracedump_origins,syntax_intel|common.loglevel' => 1, # i#1807
                                    );
             if ($is_32) {
--- a/suite/tests/CMakeLists.txt
+++ b/suite/tests/CMakeLists.txt
@@ -3542,7 +3542,7 @@ if (BUILD_CLIENTS)
     endif ()
 
     if (LINUX AND X64 AND HAVE_RSEQ)
-      torunonly_drcacheoff(rseq linux.rseq "" "@-test_mode" "")
+      torunonly_drcacheoff(rseq linux.rseq "" "${test_mode_flag}" "")
     endif ()
 
     if (AARCH64)


suite/runsuite_wrapper.pl
```diff
@@ -313,6 +313,11 @@
                                    'code_api|linux.fib-conflict' => 1,
                                    'code_api|linux.fib-conflict-early' => 1,
                                    'code_api|linux.mangle_asynch' => 1,
+                                   'code_api|tool.drcachesim.phys' => 1, # i#4922
+                                   'code_api|tool.drcacheoff.rseq' => 1, # i#4924
```
derekbruening, line 317: This can be removed since the test mode fix in this PR fixes it?

derekbruening (approved):

suite/runsuite_wrapper.pl
derekbruening: This test would then be removed from the commit message too?

AssadHashmi: > This can be removed since the test mode fix in this PR fixes it?

The test now runs with the `test_mode_flag` fix but the test fails with `failed to match expected output`. Updated #4924.

derekbruening:
Looks like this change broke the tool.drcacheoff.rseq test on x86 by removing the `@` preceding the tool options, causing the glob to include "-test_mode".

Why didn't this fail the GA CI suite?  Do those VM's not have the rseq header or something?

derekbruening:
> Looks like this change broke the tool.drcacheoff.rseq test on x86 by removing the `@` preceding the tool options, causing the glob to include "-test_mode".
> 
> Why didn't this fail the GA CI suite? Do those VM's not have the rseq header or something?

Indeed:
https://github.com/DynamoRIO/dynamorio/runs/2855772012?check_suite_focus=true
```
-- Looking for linux/rseq.h - not found
```

Not sure how to solve that as it's from just the libc package:
```
$ dpkg -S /usr/include/linux/rseq.h
linux-libc-dev:amd64: /usr/include/linux/rseq.h
```

Given that it's just a kernel feature and has no library: we only need that one header and could copy it manually somehow.

derekbruening:
Oh we're still using Ubuntu 16.04:
https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1604-README.md
Linux kernel version: 4.15.0-1113-azure

And we need 4.18 for rseq.  So we'd have to upgrade to 18.04.  IIRC we have some tests that turn flaky there?  Anyway will file a separate issue on that.

derekbruening:
Filed #4953

AssadHashmi:
> And we need 4.18 for rseq. So we'd have to upgrade to 18.04. IIRC we have some tests that turn flaky there?

Yes. We have found the following to be flaky on 18.04 AArch64:
```
code_api|tool.drcachesim.coherence and 
code_api|tool.drcachesim.threads-with-config-file
```

PR #4932, opened 2021-06-01
iX: demonstrate broken aflags and reg restore on fault

abhinav92003:
Demonstrates issues with drreg restore state logic on faults:
- restore-on-fault for aflags is broken when they are not spilled to any slot, but simply stored in xax
- restore-on-fault for gpr is broken when the gpr is read by an app instr before fault


AssadHashmi:
@abhinav92003 the AArch64 precommit failure is an out-of-date workaround on Jenkins which has been fixed now. Future commits should pass.

abhinav92003:
Fixed in #4944.

Issue #4678, opened 2021-01-19
Library for systematic machine state comparisons around any instrumentation sequence


derekbruening:
Today, many of our tests of instrumentation simply run it on a small application.  Yet if the instrumentation corrupts part of the application's machine state, the application may not notice it at all.  What we'd like is to perform systematic, exhaustive machine state comparisons across the instrumentation sequence.  We do have some tests like this: e.g., the `api.detach_state` test.  The idea here is to make some kind of library that makes it easy to add such tests everywhere.  Perhaps instead of the application itself checking its own state, it would be easier to apply (since different features require different applications for whatever they're targeting) if the checks are in client code that acquires the application mcontext.  That won't work for detach where there is no client on the other side, and it won't catch bugs in dr_get_mcontext() itself (like #4508), but it would catch many types of bugs.

derekbruening:
One complication is lazy restores: where does the library check that each app value matches the pre-instru value?  The library would have to implement some of the same lazy conditions as drreg, adding complexity and thus possible errors in the library itself.

derekbruening:
I'm pasting some notes from discussions about extending drstatecmp to help catch bugs with state restoration for both faults and asynchronous thread redirection such as synchall for flushing (which should help catch detach bugs too).  Since this is not contained to a single block, the current drstatecmp design would have to be augmented.

Bug across fault:
Won't reach end of block state cmp.
Want to update start-of-block state snapshot w/ state right before fault:
but can't in instrumented run b/c of lazy drreg state.
What if ran non-instru *first*, update state before every instr that
might fault, stop exec before fault, then run instru copy,
have state restore event run after drreg's and compare drreg's restored
state to the pre-fault snapshot?
Have to be able to identify faults beforehand.
If drreg messes up state and app doesn't fault on instru exec: will
detect in end-of-block snapshot not matching pre-fault snapshot.
Can limit to (synthesized) apps w/ annotations for faults: maybe
better to have fault checks on limited custom apps, than weaker checks
on arbitrary apps.

PR #4938, opened 2021-06-07
i#4678: DrStateCmp, a library for machine state comparison.

sapostolakis:
Implements drstatecmp, a library for machine state comparison for detecting instrumentation-induced clobbering of application state. This implementation supports checks for side-effect-free basic blocks (with the exception of the terminating instruction that could be a function call with side effects). The check involves executing two versions of these basic blocks, one with instrumentation and one without, and then comparing the resulting state to detect bugs. When a mismatch is detected, by default, the disassembled basic block that involved the bug will be emitted.

Introduces a new fifth instrumentation phase in drmgr (called post-instrumentation) to enable the drstatecmp library to perform the  duplication and the state comparison after all other instrumentation passes and without breaking any drmgr instrumentation contract (e.g., by adding app instructions in the instru2instru phase).

Illustrates the use of the drstatecmp library by providing a sample client.  This sample client introduces an instrumentation bug that is caught by drstatecmp. This sample client also shows how to specify an user-defined  callback to be invoked on state comparison mismatches.

Adds a fuzzing app for AArch64 that automatically generates from the decoding files a random sequence of side-effect-free and non-branch instructions with fuzzed operands for stress-testing clients with drstatecmp. This enables automatic testing of even newly inserted instructions in the decoder. 

Adds an option in drcachesim to enable drstatecmp and an AArch64 test that enables this option along with delayed tracing for the fuzzing app. Testing with an older DR version showed that this test would have eventually automatically detected #5023  (instrumentation-clobbering of aflags). 

Split up into 4 smaller PRs: #5059, #5090, #5107, #5125.

Issue: #4678


johnfxgalea:
I had a quick look at the PR! Thank you.

Not sure, but I think drbbdup can help you with your the bb copying, and facilitate the support of x86 32.

derekbruening:
> Not sure, but I think drbbdup can help you with your bb copying, and facilitate the support of x86 32.

I believe I mentioned drbbdup up front -- @sapostolakis you had reasons why it wouldn't work well in this case?  May be worth comments in the code on that front as well for future readers who see a multi-copy block and might wonder "why not drbbdup".

sapostolakis:
> > Not sure, but I think drbbdup can help you with your bb copying, and facilitate the support of x86 32.
> 
> I believe I mentioned drbbdup up front -- @sapostolakis you had reasons why it wouldn't work well in this case? May be worth comments in the code on that front as well for future readers who see a multi-copy block and might wonder "why not drbbdup".

Yep, I looked into drbbdup. It was well documented and it was very useful to understand how some things worked and get some ideas about duplication. I ended up not using it since just the duplication of the basic block is relatively simple (and a small part of the code, especially once more complete checking is added) and the use case is different enough that it cannot easily be used as it is (or with minimum changes).  In particular, drbbdup dispatches control according to runtime conditions, so one case is executed every time. Instead, I want to execute both the original basic block and the copy basic block. Besides, the biggest complexity of drstatecmp is the state comparison parts, and I do not have use for much of drbbdup's functionality. Finally, I also need to have ARM support so that would require extending and testing drbbdup for that as well.

For the x86-32 the issue is in the comparison part which is done in a clean call. eax seems to mismatch. Apart from the state comparison part, all the rest of the code is currently arch agnostic.

api/samples/statecmp.c
```diff
@@ -0,0 +1,55 @@
...
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+/* Illustrates using the drstatecmp extension.
+ *
```
derekbruening, line 34: This is not finished, yet the code doesn't say that: generally it's best to have TODO or other comments to make that clear.

What do you envision having here?  This should have some sample instrumentation to be tested or sthg.

I'm wondering whether this doesn't fit into api/samples/ which is for simple example tools, while drstatecmp is for comprehensive, sophisticated testing of instrumentation libraries and large tools.

ext/drstatecmp/drstatecmp.dox
```diff
@@ -0,0 +1,146 @@
...
+
+Side-effect-free basic blocks are duplicated and executed twice, once with the full
+instrumentation sequence and once with no instrumentation except for a state comparison at
+the end. Essentially, we check whether the machine state at the end of the basic block is
+the same with and without instrumentation code. This check should catch any state
+clobbering by instrumentation.
+
+The side-effect-free basic blocks are duplicated at drmgr's app2app stage. The copy of
+each such basic block is appended immediately after the original basic block. Labels are
+inserted in the beginning of the original basic block (ORIG_BB) and its copy (COPY_BB) to
```
derekbruening, line 75: Use `\p ORIG_BB` or sthg to indicate it's a literal code term.  Ditto for the others.

ext/drstatecmp/drstatecmp.dox
```diff
@@ -0,0 +1,146 @@
...
+allow identification in later stages. The terminating instruction of the basic block (if
+any) is only kept in the copy and is preceded by a TERM label. A jump to the TERM label is
+inserted at the end the original basic block to skip the duplicated bb.
+
+The insert instrumentation stage inserts the actual checks. First, the original basic
+block is instrumented. In the beginning of the original basic block, the machine state is
+saved. This state will be used for state restoration before the re-execution of the basic
+block. At the end of the original block, code is inserted to save the state for later
+comparison and restore the machine state to the state before executing this bb. This stage
```
derekbruening, line 84: s/this bb/the other copy/ or sthg

ext/drstatecmp/drstatecmp.dox
```diff
@@ -0,0 +1,146 @@
...
+also removes the previously inserted jump instr to the TERM label and let it fall through
```
derekbruening, line 85: grammar: s/let/lets/

derekbruening: Not understanding the inserted-and-then-removed jump instr: this is all static operation so why insert at all if it will jsut be removed?

ext/drstatecmp/drstatecmp.dox
```diff
@@ -0,0 +1,146 @@
...
+to the instructions of the copy of the basic block. Then, the copy of the basic block is
+instrumented. At the end of the copy, a clean call is inserted that compares the current
+machine state with the machine state that was saved at the end of the original basic
+block. Finally, all previously-inserted (from other passes) instrumentation instructions
```
derekbruening, line 89: You mean from other app2app actions like rep string expansion?  How do you know how to undo such expansion?  It seems like you would instead need to have a high priority app2app callback to copy before anyone changed the code?

ext/drstatecmp/drstatecmp.dox
```diff
@@ -0,0 +1,146 @@
...
+\subsection sec_drstatecmp_checks_side_effects Basic Blocks with Side Effects
+
+Checking basic blocks with side effects is not yet implemented but this section serves
+as a high-level design overview.
+
+Basic blocks with side-effects cannot be executed twice (at least not without a lot of
+extra complexity to monitor and restore memory state). In this case, the checks are
+inserted throughout the basic block whenever needed taking into account the app
+instructions, existing instrumentation, and the lazy restoration by drreg. As a result,
+incomplete-decoder bugs and lazy-condition-ddreg bugs will not be detected.
```
derekbruening, line 102: "ddreg"

ext/drstatecmp/drstatecmp.dox
```diff
@@ -0,0 +1,146 @@
...
+
+Instructions with side effects include instructions that write to memory, interrupts,
+and system calls. To avoid inter-procedural analysis, all function calls are
+conservatively assumed to have side effects.
```
derekbruening, line 106: I think this last sentence can be removed (function calls are are not embedded; they'll become separate blocks for each piece)

ext/drstatecmp/drstatecmp.dox
```diff
@@ -0,0 +1,146 @@
...
+
+\section sec_drstatecmp_instr_state_sharing Instrumentation State Sharing
+
+The instrumentation stages communicate through the user_data parameter. Space for this
+data is allocated in thread-local storage in the app2app stage and deallocated at the
+end of the insert instrumentation stage. The app2app stage specifies in the user_data
+whether a basic block is side-effect free or not. The analysis stage is not used except
+for passing over the user_data to the insert instrumentation stage. In the latter stage,
+the user_data is also used to track when the instrumentation moves from the original
+basic block to its copy.
```
derekbruening, line 126: Hmm, this is more of an internal implementation detail.  I would expect these docs to focus more on how to use this library: like how to create a test from an existing complex client.  I want to create a version that includes these extra state comparisons for my test.  Do I have to use ifdefs to do two separate builds to avoid the checks in my release version?  Or is there a convenient flag where drstatecmp can be disabled without having to be ifdef-ed out?  If I link with it but place the drstatecmp_init call under a test-only flag in my client: will that do it, or will there still be overhead?

What apps should be executed for the test?  Up to the user presumably.

How are comparison failures reported?  Does the user get to decide whether the process is killed or whether there's just a printout or a callback to some user-controlled action?

ext/drstatecmp/drstatecmp.dox
```diff
@@ -0,0 +1,146 @@
...
+
+\section sec_drstatecmp_machine_state_saved Machine States Saved
+
+During execution, thread-local storage is used to keep two machine states for each
+side-effect basic block. One is the state of the original basic block before its
```
derekbruening, line 131: side-effect-free?

ext/drstatecmp/drstatecmp.h
```diff
@@ -0,0 +1,99 @@
...
+    DRSTATECMP_ERROR_NOT_INITIALIZED,     /* Operation failed: not initialized. */
+} drstatecmp_status_t;
+
+/***************************************************************************
+ * INIT
+ */
+
+/**
+ * Priorities of drmgr instrumentation passes used by drstatecmp.  Users
+ * of drstatecmp can use the name DRMGR_PRIORITY_NAME_DRSTATECMP in the
```
derekbruening, line 64: Use `#` for a doxygen link: `#DRMGR_PRIORITY_NAME_DRSTATECMP`

ext/drstatecmp/drstatecmp.h
```diff
@@ -0,0 +1,99 @@
...
+ * drmgr_priority_t.before field or can use these numeric priorities
```
derekbruening, line 65: Ditto: `#drmgr_priority_t.`

ext/drstatecmp/drstatecmp.h
```diff
@@ -0,0 +1,99 @@
...
+ * \addtogroup drstatecmp Machine State Comparison Library
+ */
+/*@{*/ /* begin doxygen group */
+
+/** Success code for each drstatecmp operation. */
+typedef enum {
+    DRSTATECMP_SUCCESS,                   /* Operation succeeded. */
+    DRSTATECMP_ERROR,                     /* Operation failed. */
+    DRSTATECMP_ERROR_ALREADY_INITIALISED, /* DRSTATECMP can only be initialised once. */
+    DRSTATECMP_ERROR_NOT_INITIALIZED,     /* Operation failed: not initialized. */
```
derekbruening, line 55: Be consistent: you have INITIALISED on one line (British spelling) and INITIALIZED on the next, in the code and the comments.

ext/drstatecmp/drstatecmp.h
```diff
@@ -0,0 +1,99 @@
...
+#endif
+
+/**
+ * \addtogroup drstatecmp Machine State Comparison Library
+ */
+/*@{*/ /* begin doxygen group */
+
+/** Success code for each drstatecmp operation. */
+typedef enum {
+    DRSTATECMP_SUCCESS,                   /* Operation succeeded. */
```
derekbruening, line 52: These should all be Doxygen comments `/**< */`

ext/drstatecmp/drstatecmp.h
```diff
@@ -0,0 +1,99 @@
...
+    DRSTATECMP_ERROR,                     /* Operation failed. */
+    DRSTATECMP_ERROR_ALREADY_INITIALISED, /* DRSTATECMP can only be initialised once. */
```
derekbruening, line 54: nit: why all caps in the comment when no other docs use all caps?

derekbruening (commented):
Didn't look at drstatecmp.c yet: comments on the rest of the files.

ext/drstatecmp/drstatecmp.c
```diff
@@ -0,0 +1,572 @@
...
+    /* Change mode to process the copy bb once we encounter the COPY_BB label.
+     * Save the state at the end of this bb for later comparison, and remove the
+     * previously inserted jmp instr to the TERM label of the copy bb and let it fall
+     * through to the instructions of the copy bb (that will be instrumentation-free).
+     * Finally, restore the machine state to the state before executing the original
+     * version of this bb (allows us to re-execute this bb).
+     */
+    else if (match_label_val(instr, DRSTATECMP_LABEL_COPY_BB)) {
+        data->orig_bb_mode = false;
+        instr_t *prev_instr = instr_get_prev_app(instr);
```
derekbruening, line 445: A sanity check that prev_instr is not NULL and is what you expect would be good

ext/drstatecmp/drstatecmp.c
```diff
@@ -0,0 +1,572 @@
...
+    bool side_effect_free; /* Denotes whether the current bb has side-effects. */
+    bool orig_bb_mode; /* True when processing the instructions of the original version of
+                          a side-effect-free bb. Becomes false when starting processing
+                           the copies of these instructions*/
+} drstatecmp_user_data_t;
+
+typedef struct {
+    dr_mcontext_t saved_state_for_restore; /* Last saved machine state for restoration. */
+    dr_mcontext_t saved_state_for_cmp;     /* Last saved machine state for comparison. */
+} drstatecmp_saved_states;
```
derekbruening, line 55: style: `_t` suffix for type names

ext/drstatecmp/drstatecmp.c
```diff
@@ -0,0 +1,572 @@
...
+#    define ASSERT(x, msg) DR_ASSERT_MSG(x, msg)
+#else
+#    define ASSERT(x, msg)
+#endif
+
+typedef struct {
+    bool side_effect_free; /* Denotes whether the current bb has side-effects. */
+    bool orig_bb_mode; /* True when processing the instructions of the original version of
+                          a side-effect-free bb. Becomes false when starting processing
+                           the copies of these instructions*/
```
derekbruening, line 49: style: * on each line

style:  missing end `.`

ext/drstatecmp/drstatecmp.c
```diff
@@ -0,0 +1,572 @@
...
+    return instr_get_note(instr) == (void *)get_label_val(label_type);
+}
+
+/* Create and insert bb labels. */
+static instr_t *
+drstatecmp_insert_labels(void *drcontext, instrlist_t *ilist, instr_t *where,
+                         drstatecmp_label_t label_type, bool preinsert)
+{
+    instr_t *label = INSTR_CREATE_label(drcontext);
+    instr_set_meta(label);
```
derekbruening, line 96: instrlist_meta_preinsert does this, so this is redundant

ext/drstatecmp/drstatecmp.c
```diff
@@ -0,0 +1,572 @@
...
+/* Compare instr label against a label_type. Returns true if they match. */
+static inline bool
+match_label_val(instr_t *instr, drstatecmp_label_t label_type)
+{
+    return instr_get_note(instr) == (void *)get_label_val(label_type);
+}
+
+/* Create and insert bb labels. */
+static instr_t *
+drstatecmp_insert_labels(void *drcontext, instrlist_t *ilist, instr_t *where,
```
derekbruening, line 92: Remove plural from name since it inserts *one* label?

ext/drstatecmp/drstatecmp.c
```diff
@@ -0,0 +1,572 @@
...
+    return label;
+}
+
+/* Returns whether or not instr may have side effects. */
+static bool
+drstatecmp_may_have_side_effects_instr(instr_t *instr)
+{
+    /* Instructions with side effects include instructions that write to memory,
+     * interrupts, and syscalls. To avoid inter-procedural analysis, all function calls
+     * are conservatively assumed to have side effects.
```
derekbruening, line 111: The function call comment can be removed.

ext/drstatecmp/drstatecmp.c
```diff
@@ -0,0 +1,572 @@
...
+     */
+    return instr_writes_memory(instr) || instr_is_interrupt(instr) ||
+        instr_is_syscall(instr) || instr_is_call(instr);
```
derekbruening, line 114: The instr_is_call can be removed: it should be just like any other branch for the purposes here

ext/drstatecmp/drstatecmp.c
```diff
@@ -0,0 +1,572 @@
...
+    drstatecmp_user_data_t *data = (drstatecmp_user_data_t *)dr_thread_alloc(
+        drcontext, sizeof(drstatecmp_user_data_t));
+    memset(data, 0, sizeof(*data));
+    *user_data = (void *)data;
+
+    /* Determine whether the basic block is free of side effects. */
+    bool side_effect_free = true;
+    for (instr_t *inst = instrlist_first_app(bb); inst != NULL;
+         inst = instr_get_next_app(inst)) {
+        side_effect_free &= !drstatecmp_may_have_side_effects_instr(inst);
```
derekbruening, line 137: Avoid bitwise operators for bools, especially in C where there's no true bool type: safest to use logical &&

ext/drstatecmp/drstatecmp.c
```diff
@@ -0,0 +1,572 @@
...
+     * COPY_BB:
+     *   instr1
+     *   instr2
+     * TERM:
+     *   term_instr
+     *
+     */
+
+    /* Create a clone of bb. */
+    instrlist_t *copy_bb = instrlist_clone(drcontext, bb);
```
derekbruening, line 178: If sthg like rep str or scatter expansion went first and inserted emulation labels with stored data, cloning becomes a problem as the clone has the labels but they are missing some data (IIRC the data is set to NULL).  Looks like the late priority will have this go after those expansions.  Better to go first?

ext/drstatecmp/drstatecmp.c
```diff
@@ -0,0 +1,572 @@
...
+                                              DRSTATECMP_LABEL_TERM, /*preinsert=*/true);
+    } else {
+        term_label = drstatecmp_insert_labels(drcontext, copy_bb, term_inst_copy_bb,
+                                              DRSTATECMP_LABEL_TERM, /*preinsert=*/false);
+    }
+    /* Need an operand for the jmp instr targeting this label. */
+    opnd_t term_label_opnd = opnd_create_instr(term_label);
+
+    /* Replace the terminating instruction of the original bb with a jmp to the
+     * terminating instruction of the bb clone (if any).
```
derekbruening, line 199: What is the purpose of this added jump?  Other instrumentation libraries/clients are not likely to look at it and conclude that it skips the copy and then ignore the copy: they'll assume linear control flow and blindly instrument every instruction in the list, ignoring the fact that there's a jump here.

ext/drstatecmp/drstatecmp.c
```diff
@@ -0,0 +1,572 @@
...
+    instr_t *jmp_term = XINST_CREATE_jump(drcontext, term_label_opnd);
+    if (instr_is_cti(term_inst) || instr_is_return(term_inst)) {
+        instrlist_replace(bb, term_inst, jmp_term);
+        instr_destroy(drcontext, term_inst);
+    } else {
+        /* If there is no control transfer, append a jmp at the end of the bb. */
+        instrlist_postinsert(bb, term_inst, jmp_term);
+    }
+
+    /* Append the instructions of the bb copy to the original bb. */
```
derekbruening, line 211: But this will present the copy to all the instrumentation libraries and clients.  So drreg will consider all the reg uses in the copy when deciding which scratch regs to use, which will warp its behavior so it won't match what it would do w/o a copy.  IIRC its liveness analysis does consider jumps?  Is that why a jump was added?  Please add comments if so.  But its reg use counts will ignore the jump.  Plus other libraries/clients frequently count app instrs or take actions across app instrs: the most common is putting a clean call or batched action at the end of the block, which would mean inside this copy.  Then below you remove all the instrumentation: so the resulting block will not contain the instrumentation in the original part that would have been there.

I think this copy has to be hidden from the instrumentation, and only added in the insru2instru phase or something like that.

derekbruening (commented):
Finishing off the initial review.

ext/drstatecmp/drstatecmp.c
derekbruening: Hmm, I was thinking it would be good to have the original rather than expanded version (i.e., single scatter instruction; single rep movs insruction), so going first was always good.  But for true emulation where the original won't even run on the underlying hardware: you'd want to go after that.

derekbruening: Actually for the rep movs I guess you couldn't take the original which would loop all N iters at once and compare to the expanded version which would do just one iter for each state cmp.

ext/drstatecmp/drstatecmp.c
```diff
@@ -0,0 +1,572 @@
...
+        (drstatecmp_saved_states *)drmgr_get_tls_field(drcontext, tls_idx);
+
+    dr_mcontext_t *mc_instrumented = &pt->saved_state_for_cmp;
+    dr_mcontext_t mc_expected;
+    mc_expected.size = sizeof(mc_expected);
+    mc_expected.flags = DR_MC_ALL;
+    dr_get_mcontext(drcontext, &mc_expected);
+
+#ifdef X86
+    drstatecmp_check_gpr_value("xdi", mc_instrumented->xdi, mc_expected.xdi);
```
johnfxgalea, line 339: I would add all these debug checks to a helper function.

ext/drstatecmp/drstatecmp.c
```diff
@@ -0,0 +1,572 @@
...
+     * Save the state at the end of this bb for later comparison, and remove the
+     * previously inserted jmp instr to the TERM label of the copy bb and let it fall
+     * through to the instructions of the copy bb (that will be instrumentation-free).
+     * Finally, restore the machine state to the state before executing the original
+     * version of this bb (allows us to re-execute this bb).
+     */
+    else if (match_label_val(instr, DRSTATECMP_LABEL_COPY_BB)) {
+        data->orig_bb_mode = false;
+        instr_t *prev_instr = instr_get_prev_app(instr);
+        instrlist_remove(bb, prev_instr);
```
johnfxgalea, line 446: If I am not mistaken, you cannot remove app instructions (inserted at app2app) during the insertion phase. This breaks DR's contract.

Moreover, there could have been an application code analysis pass (so between app2app and insertion) that takes into account the jmp.

johnfxgalea (commented):
My major concern is the removal of the jmp instr during the instrumentation phase.

ext/drstatecmp/drstatecmp.c
johnfxgalea: @derekbruening Without the jmp, one can have tools, like drreg, which will consider the orig and copy blocks as one whole basic block and therefore perform incorrect analysis (for drreg's case liveness). Think of a situation where a register used in the orig block is erroneously considered dead only because the copy block writes to it further down the line.

Edit: I just read that you raised a similar point further down in your review.

ext/drstatecmp/drstatecmp.c
derekbruening: @johnfxgalea -- Any thoughts on the best way to hide the extra copy from instrumentation?  It seems like either using the drbbdup approach with a drmgr-involved modified interface to hide the other copies, or adding it in the instru2instru phase (or I suppose insertion w/ very low priority) are the options: with the latter technically breaking rules on adding app instructions.

sapostolakis: Yep, the jump was added to allow drreg to do liveness analysis correctly but you are right that in many cases linear control flow is assumed and the copy will affect various instrumentation passes. To me doing everything in the instru2instru phase seems to be the best choice. Besides the main motivation for the comparison library is to check for clobbering by instrumentation. So, the safest option is to add the checks at the very end without influencing any other instrumentation. The code will also be simpler, since there is bb-level scope in instru2instru and everything will done in a single pass.

johnfxgalea: instru2instru does seem simple, but I don't really like the idea of breaking drmgr's instrumentation contract with the addition of app instructions, especially if this becomes a common occurrence in the future. Then again, other solutions require more work.

If performance is not a major problem, an alternative approach might be to rely on re-direction, i.e., with the use of drbbdup,  first execute the instrumented case. Then rely on a clean call, which is placed at the end, to set a flag and redirect execution again to the same basic block. The active flag will instead lead drbbdup to dispatch to the uninstrumented case (which will only include instrumentation to reset the flag).

sapostolakis: Performance is not that critical for now (although it might be in the future if it proves useful and it is used beyond test apps), but code simplicity is. The point of this lib is to detect bugs in other DR components while it remains bug-free itself. If it is too complex or relies on other extensions (and thus inherit their complexities and potential bugs) then its usefulness decreases. Bugs in drstatecmp or its deps could lead to time-wasting false positives (incorrect detection of mismatch that needs to be investigated manually) or false negatives (undetected clobbering). I would also count as detrimental any intrusive effect of drstatecmp or its deps on the behavior of other DR components.

johnfxgalea: All great points, and a big fan of simplicity.

> while it remains bug-free itself

The only issue I see is that going against the specification of drmgr is a bug in itself.

We could document the violating use of the instru2instru phase really well, stressing that DrStateCmp is just a debugging tool, but if we can find a better way to avoid this exception to the rule, that would be great! In general, exceptions to rules could potentially lead to nasty bugs or maintenance issues in the future, and this is something we should avoid.

Keeping in mind points on simplicity, how about we introduce a new final special drmgr phase (maybe call it `bb2debug`?), which is suited for DrStateCmp and other debugging extensions that we might create in the future? The phase will be the the last phase considered by drmgr and allows for flexible instrumentation, including app instr. Its introduction appears that it won't break drmgr rules.

sapostolakis: Having a fifth drmgr phase might be a good idea. The fourth phase is for instrumentation optimizations and there will be a fifth phase that is for instrumentation debugging. Seems appropriate. I am fine with more work as long as DrStateCmp itself remains simple.

johnfxgalea: Yeah! If we do include the fifth stage, I'm not sure if we should document it solely for use by so called "debugging" extensions or allow other "normal" extensions to possibly register it as well. There could be some small differences in the implementation. If we go for phase 5 to be strictly for debugging tools, then we could potentially have an internal header file, and leave drmgr.h unchanged.

Maybe @derekbruening also has an opinion on this?

derekbruening: My first reaction is to be hesitant to change the core phases: I started implementing a split of the insertion phase and ended up bailing because of complexities with drbbdup and other existing users of the current phases (filed aborted attempt as #4947).  But a completely separate phase that doesn't touch the existing key phases of app2app-analysis-insertion is probably going to have minimal disruption so it seems fine in that respect.

I kind of like the drbbdup re-execute idea at a high level: drbbdup is an existing framework supporting construction of two separate versions, which is exactly what we want here.  But having to do the full redirect to chain two versions, which drbbdup wasn't designed for, makes it much less appealing.

> If we go for phase 5 to be strictly for debugging tools, then we could potentially have an internal header file, and leave drmgr.h unchanged.

Could you elaborate on this?  I assumed we would want a public phase 5 since we want these debugging tools to be first-class libraries usable with new instrumentation schemes.

johnfxgalea: Put simply, we need to decide whether we should have access to phase 5 via drmgr_priv.h or drmgr.h. If we decide that phase 5 is solely intended for debugging tools, maybe drmgr_priv.h would be sufficient. If we envisage phase 5 to be used more generally, then drmgr.h ?

derekbruening: OK I could see separating, or documenting inside drmgr.h that it is meant for only certain uses.  But even if separated it would still be a published header using exported drmgr library functions, right?  So it wouldn't really be private/internal: just documented as for narrow use cases.

johnfxgalea: But even if separated it would still be a published header using exported drmgr library functions, right? 

Yep, but generally users should include drmgr.h and not drmgr_priv.h.

derekbruening: OK, though as a nit I would have a different name, since _private.h in DR means an interface not exported at all outside a module/library.

sapostolakis:
Added a 5th instrumentation phase in drmgr, as discussed, and updated drstatecmp to use it. 
Converted this PR to a draft since, before further reviewing, I need to refine it a bit, address all previous review comments, and update the documentation regarding the newly introduced instrumentation phase.

ext/drstatecmp/drstatecmp.h
sapostolakis: Fixed.

ext/drstatecmp/drstatecmp.c
sapostolakis: Done.

ext/drstatecmp/drstatecmp.c
sapostolakis: Removed.

ext/drstatecmp/drstatecmp.c
sapostolakis: Removed.

ext/drstatecmp/drstatecmp.c
sapostolakis: Fixed.

ext/drstatecmp/drstatecmp.c
sapostolakis: Deleted code.

ext/drstatecmp/drstatecmp.c
sapostolakis: Fixed.

ext/drstatecmp/drstatecmp.c
sapostolakis: Deleted code.

ext/drstatecmp/drstatecmp.c
```diff
@@ -0,0 +1,572 @@
...
+        if (data->orig_bb_mode) {
+            drstatecmp_insert_instrument_bb_orig_side_effect_free(drcontext, bb, instr,
+                                                                  data);
+        } else {
+            drstatecmp_insert_instrument_bb_copy_side_effect_free(drcontext, bb, instr);
+        }
+    }
+
+    /* Free the user_data when we are done with this bb. */
+    if (instr == instrlist_last(bb))
```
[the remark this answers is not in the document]
sapostolakis: Deleted code.

ext/drstatecmp/drstatecmp.c
```diff
@@ -0,0 +1,572 @@
...
+        drstatecmp_compare_state(drcontext, bb, instr);
+        return;
+    }
+
+    /* Remove all instrumentation code in the bb copy to detect any
+     * unintended clobbering by instrumentation.
+     */
+    if (!instr_is_app(instr)) {
+        instrlist_remove(bb, instr);
+        instr_destroy(drcontext, instr);
```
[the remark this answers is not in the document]
sapostolakis: Fixed with new implementation.

ext/drstatecmp/drstatecmp.c
```diff
@@ -0,0 +1,572 @@
...
+    /* Compare current state at the end of the bb with the saved state at the
+     * end of the original (instrumented) bb.
+     */
+    if (match_label_val(instr, DRSTATECMP_LABEL_TERM)) {
+        drstatecmp_compare_state(drcontext, bb, instr);
+        return;
+    }
+
+    /* Remove all instrumentation code in the bb copy to detect any
+     * unintended clobbering by instrumentation.
```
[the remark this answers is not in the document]
sapostolakis: Fixed.

ext/drstatecmp/drstatecmp.c
sapostolakis: Handle app2app transformations with new implementation.

ext/drstatecmp/drstatecmp.c
sapostolakis: The jump is not used anymore.

ext/drstatecmp/drstatecmp.c
sapostolakis: Addressed by moving all changes to the new final phase.

ext/drstatecmp/drstatecmp.c
```diff
@@ -0,0 +1,606 @@
...
+    drstatecmp_user_data_t *data = (drstatecmp_user_data_t *)user_data;
+
+    if (!data->side_effect_free_bb)
+        return DR_EMIT_DEFAULT;
+
+    for (instr_t *inst = instrlist_first(bb); inst != NULL; inst = instr_get_next(inst)) {
+        if (drmgr_is_emulation_start(inst)) {
+            emulated_instr_t emulated_inst;
+            drmgr_get_emulated_instr_data(inst, &emulated_inst);
+            if (!(emulated_inst.flags & DR_EMULATE_INSTR_ONLY)) {
```
derekbruening, line 181: We use `TEST`

ext/drstatecmp/drstatecmp.c
```diff
@@ -0,0 +1,606 @@
...
+        instr_is_syscall(instr);
+}
+
+/****************************************************************************
+ * APPLICATION-TO-APPLICATION PHASE
+ *
+ * Save a pre-app2app copy of side-effect-free basic blocks.
+ * It is assumed that this pass has the highest priority among all app2app passes
+ * and thus it is able to capture the pre-app2app state.
+ *
```
derekbruening, line 124: nit: empty line at unusual place

ext/drstatecmp/drstatecmp.c
```diff
@@ -0,0 +1,606 @@
...
+    for (instr_t *inst = instrlist_first(bb); inst != NULL; inst = instr_get_next(inst)) {
+        if (drmgr_is_emulation_start(inst)) {
+            emulated_instr_t emulated_inst;
+            drmgr_get_emulated_instr_data(inst, &emulated_inst);
+            if (!(emulated_inst.flags & DR_EMULATE_INSTR_ONLY)) {
+                /* Emulation sequence required for re-execution and golden bb copy should
+                 * be the post-app2app copy.
+                 */
+                instrlist_clear_and_destroy(drcontext, data->pre_app2app_bb);
+                data->golden_bb_copy = instrlist_clone(drcontext, bb);
```
derekbruening, line 186: Best to have a test case for this path

ext/drstatecmp/drstatecmp.c
```diff
@@ -0,0 +1,606 @@
...
+     *   instr1
+     *   instr2
+     *
+     * TERM:
+     *   term_instr
+     *
+     */
+
+    /* Get an instrumentation-free copy of the bb which is the golden copy kept
+     * in the user_data*/
```
derekbruening, line 254: style: missing space and newline

ext/drstatecmp/drstatecmp.c
```diff
@@ -0,0 +1,606 @@
...
+                                                    /*preinsert=*/true);
+    labels->copy_bb_start =
+        drstatecmp_insert_label(drcontext, copy_bb, instrlist_first(copy_bb),
+                                DRSTATECMP_LABEL_COPY_BB, /*preinsert=*/true);
+    /* Insert the TERM label before the terminating instruction or after the
+     * last instruction if the bb falls through.
+     */
+    instr_t *term_inst_copy_bb = instrlist_last_app(copy_bb);
+    bool preinsert = true;
+    if (!instr_is_cti(term_inst_copy_bb) && !instr_is_return(term_inst_copy_bb))
```
derekbruening, line 269: "cti" should include return

ext/drstatecmp/drstatecmp.c
```diff
@@ -0,0 +1,606 @@
...
+    drstatecmp_restore_state(drcontext, bb, labels->copy_bb_start);
+
+    /* Compare the state at the end of the copy bb (uninstrumented) with the saved state
+     * at the end of the original (instrumented) bb to detect clobbering by the
+     * instrumentation.
+     */
+    drstatecmp_compare_state(drcontext, bb, labels->term);
+}
+
+/* Duplicate the side-effect basic block for re-execution and add saving/restoring of
```
derekbruening, line 493: side-effect free

derekbruening (commented):
I think the only major issue is adding a test.

Not sure if @johnfxgalea wanted to also take a look?

ext/drstatecmp/drstatecmp.c
```diff
@@ -0,0 +1,606 @@
...
+    mcontext->size = sizeof(*mcontext);
+    mcontext->flags = DR_MC_ALL;
+    dr_get_mcontext(drcontext, mcontext);
+}
+
+static void
+drstatecmp_save_state(void *drcontext, instrlist_t *bb, instr_t *instr, bool for_cmp)
+{
+    dr_insert_clean_call(drcontext, bb, instr, (void *)drstatecmp_save_state_call,
+                         false /*fpstate */, 1, OPND_CREATE_INT32((int)for_cmp));
```
johnfxgalea, line 313: Nit: Add space for fpstate comment.

johnfxgalea (approved):

johnfxgalea:
Apart from @derekbruening 's comments, just one nit from my end.

ext/drmgr/drmgr.h
```diff
@@ -515,6 +518,42 @@ bool
...
                                               int opcode);
 
+DR_EXPORT
+/**
+ * Registers a callback function for the fifth instrumentation stage:
+ * post-instrumentation analysis and transformations on each
+ * basic block.  drmgr will call \p func as the fifth of five
+ * instrumentation stages for each dynamic application basic block.
+ * Post-instrumentation passes are allowed to insert both meta and
+ * not non-meta instructions, and are intended for debugging prior
```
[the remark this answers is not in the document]
sapostolakis: Fixed.

ext/drstatecmp/drstatecmp.dox
```diff
@@ -0,0 +1,136 @@
...
+\p CMakeLists.txt file:
+
+\code use_DynamoRIO_extension(clientname drstatecmp) \endcode
+
+That will automatically set up the include path and library dependence.
+
+The initialization routine \p drstatecmp_init() enables the insertion of the machine state
+checks. When drstatecmp is linked but the client does not invoke \p drstatecmp_init(),
+there are no inserted checks and no overhead to the client. Thus, a client can wrap the
+initialization of drstatecmp with a flag or a ifdef and invoke it only on debug builds.
```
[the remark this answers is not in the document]
sapostolakis: Done.

ext/drstatecmp/drstatecmp.dox
```diff
@@ -0,0 +1,136 @@
...
+
+Beyond instrumentation bugs, drstatecmp can also catch bugs introduced in the
+application-to-application phase. To that end, instead of using for re-execution the
+post-app2app-phase version of each side-effect-free basic block (i.e.,
+instrumentation-free copy but with potentially modified application instructions), it uses
+the pre-app2app-phase version (contains only the original app instructions), unless any of
+original app instructions requires emulation (i.e., not executable).
+
+The duplication of side-effect-free basic blocks and the checks insertion occurs at
+drmgr's post-instrumentation stage. A instrumentation-free copy of each side-effect-free
```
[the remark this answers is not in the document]
sapostolakis: Done.

ext/drstatecmp/drstatecmp.dox
```diff
@@ -0,0 +1,136 @@
...
+basic block is appended immediately after the original/instrumented version of the basic
```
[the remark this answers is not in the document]
sapostolakis: Done.

ext/drstatecmp/drstatecmp.dox
```diff
@@ -0,0 +1,136 @@
...
+block. Before the execution of the original basic block, its machine state is saved. This
```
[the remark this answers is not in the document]
sapostolakis: Yes, just using instrumented is clearer.

ext/drstatecmp/drstatecmp.c
sapostolakis: Done.

ext/drstatecmp/drstatecmp.c
sapostolakis: Done.

ext/drstatecmp/drstatecmp.c
sapostolakis: You are right. Double-checked it and indeed instr_is_cti() checks for instr_is_mbr_arch() which is a superset of instr_is_return().

ext/drstatecmp/drstatecmp.c
sapostolakis: Done.

ext/drstatecmp/drstatecmp.c
```diff
@@ -0,0 +1,606 @@
...
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+#include "dr_defines.h"
```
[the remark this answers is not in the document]
sapostolakis: Removed it.

ext/drstatecmp/drstatecmp.c
sapostolakis: Removed.

ext/drstatecmp/drstatecmp.c
sapostolakis: Done.

sapostolakis (commented):
Fixed remaining issues except for adding a test outside the sample client. Will add tomorrow.

sapostolakis:
Actually code_api|sample.statecmp failed on AArch64, so I will need to fix the sample client for AArch64 as well.

sapostolakis:
The AArch64 error and testing with drreg revealed some corner cases leading to false positives. drreg will not bother to spill (and later restore) a dead register. So, instrumentation is allowed to clobber without care a register that will be written by an app instr later in the basic block without being read first. This is mostly okay for drstatecmp since at the end of the basic block the register will contain the written app value and the checks will work out. However, if the app instr that overwrites the register is the terminating instruction of the basic block, then it is not re-executed and precedes the checks, leading to a "false" positive. It is not really a false alarm since there is in fact clobbering but it is expected and not harmful so this alarm is undesirable. In the AArch64 example, the terminating example was a 'blr' instr which sets the link register (x30), but this register was used as a scratch register and it was never spilled/restored by drreg since it will be overwritten at the end of the basic block without being read before that by any other app instr.

sapostolakis:
Another source of false positives are side-effect-free instructions that yield on re-executions different results with the same register state input. One such instruction is rdtsc. A re-execution will lead to a different result triggering a false alarm. Could be seen as "non-deterministic" results. Sadly, x86 should have a lot of opcodes with the same issue.

sapostolakis:
Added a comprehensive fuzzing app for AArch64 that automatically generates from the decoding files (codec.txt) a random sequence of side-effect-free and non-branch instructions with fuzzed operands for stress-testing clients with drstatecmp. 

A test that uses this fuzzing app with delayed tracing and enabled drstatecmp has also been added for drcachesim.

By default the app generates a basic block with 10k random instructions. 9-10% of these instructions  (while being decodable) end up leading to SIGILL and are thus skipped and 1.5-2% of produced encoding cannot be decoded. Therefore around 9k instructions are eventually used each time. To avoid flakiness, I added a handle for SIGSEGV and SIGFPE although it should be rare to see such errors, especially after limiting the Rd to 0-15 gpr.
The test takes 7sec for release and 25sec for a debug AArch64 build.

sapostolakis:
Adding @AssadHashmi as a reviewer since this PR involves changes in codec.py/c and is related to decoding efforts and improved AArch64 testing. Updated the PR description to provide more context.

suite/tests/client-interface/drx-scattergather.dll.c
```diff
@@ -330,7 +330,7 @@ dr_init(client_id_t id)
     res = drreg_init(&ops);
     CHECK(res == DRREG_SUCCESS, "drreg_init failed");
     dr_register_exit_event(event_exit);
-    ok = drmgr_register_bb_instrumentation_ex_event(event_bb_app2app, event_bb_analysis,
-                                                    event_app_instruction, NULL, NULL);
+    ok = drmgr_register_bb_instrumentation_ex_event(
+        event_bb_app2app, event_bb_analysis, event_app_instruction, NULL, NULL, NULL);
     CHECK(ok, "drmgr register bb failed");
 }
```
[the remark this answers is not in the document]
sapostolakis: It now has two tests. One with the sample client for all archs and one more comprehensive for AArch64 with the fuzzing app and drcachesim.

ext/drstatecmp/drstatecmp.c
sapostolakis: Should be exercised now with the fuzz app.

core/ir/aarch64/codec.py
```diff
@@ -338,6 +338,39 @@ def generate_opcode_names(patterns):
...
+         '} opcode_opnd_pair_t;',
+         '',
+         'const opcode_opnd_pair_t fuzz_opcode_opnd_pairs[] = {']
+    # Exclude instructions with side-effects and branch instructions.
+    # Particularly,  exclude: i) Load/Stores (opcode: x1x0);
+    # ii) Branches, Exception Generating and System instructions (opcode: 101x);
+    # iii) SVE memory (opcode: 0010 and 1 for the most significant bit).
+    # The allowed instructions include all the Data Processing instructions (including
+    # the FP and SIMD ones) and the non-memory SVE instructions.
+    excluded_opcodes = re.compile('....1.0|...101.|1..0010')
```
AssadHashmi, line 359: AIUI the MSB is only 1 for the SME (Scalable Matrix Extension) instructions.

AssadHashmi (approved):

core/ir/aarch64/codec.py
sapostolakis: Thanks for taking a look.
For finding the opcodes to exclude I relied on the Armv8 documentation of the encodings (https://developer.arm.com/documentation/ddi0596/2021-06/Index-by-Encoding/SVE-encodings?lang=en). In this documentation, SVE encodings with MSB 1 are marked as SVE memory. I assume that the newly introduced SME in Armv9 would also involve memory ops. If there are non-memory-involving ops in there that could be included in the fuzz app, we can always relax this exclusion in the future.

AssadHashmi: You are right, my bad, I was looking at old documentation :-/

derekbruening:
This is an extremely large diff (making it hard to review (more likely for review to miss pieces, burden on reviewer to find large chunk of time, etc.), harder to revert just pieces of it if sthg goes wrong: just harder to deal with): `git diff origin/master..` is 4751 lines!   We consider 1500 lines to be large: https://dynamorio.org/page_code_reviews.html#autotoc_md110

Ordinarily I would request that it be split up and would not even attempt to review it as one: the drcachesim portion should be a separate PR; the drmgr 5th stage could go in first as its own PR w/o any drstatecmp code.  Given the circumstances though I'm making my way through it; part-way at the moment.

sapostolakis:
> Ordinarily I would request that it be split up and would not even attempt to review it as one: the drcachesim portion should be a separate PR; the drmgr 5th stage could go in first as its own PR w/o any drstatecmp code. Given the circumstances though I'm making my way through it; part-way at the moment.

I think it is around 1.8K lines but I agree it is too big and I appreciate you taking the time to review it anyway. It has been reviewed incrementally and older parts have not changed but it has been a long time since then. As changes were requested, I kept updating this PR but I should have spawn new PRs for some of the changes, especially for the new drmgr stage that is quite separate from this library. Even if you get through this PR, I do not mind splitting it up for maintainability reasons (easier to revert part of it etc.).

derekbruening:
> I think it is around 1.8K lines 

FTR, I've always measured as the unified diff format line count: 

```
$ git checkout i4678-add-state-comp-lib
$ git diff origin/master.. | wc
   4751   34499  259491
```

But it looks like this branch is not merged w/ the latest master...so I think we want:
```
$ git diff $(git merge-base i4678-add-state-comp-lib origin/master).. | wc
   2451   11809  102941
```

sapostolakis:
From the github GUI I was seeing +1,735 −86  as this PR diff. But in any case, big PR. I will spawn some new PRs to make it easier to maintain (and review if you have not finished looking into this one).

I also still need to fix the windows failure. I think adding the 5th instrumentation stage led to the missing __chkstk for building drmgr. So, another reason to split it is to isolate the windows issue and resolve it.

ext/drstatecmp/drstatecmp.c
```diff
@@ -0,0 +1,691 @@
...
+}
+
+/****************************************************************************
+ * ANALYSIS PHASE
+ *
+ * The analysis phase determines which copy of each side-effect bb should be used as the
+ * golden copy for comparison. There are two options: i) pre-app2app-phase copy, where
+ * the code just contains the original app instructions, and ii) post-app2app-phase copy.
+ * The first option can catch bugs in app2app passes and it is selected unless any of
+ * original app instructions requires emulation (true emulation). In that case, the
```
derekbruening, line 181: nit: s/original/the original/

ext/drstatecmp/drstatecmp.c
```diff
@@ -0,0 +1,691 @@
...
+drstatecmp_report_error(const char *msg, void *tag)
+{
+    if (ops.error_callback != NULL) {
+        (*ops.error_callback)(msg, tag);
+    } else {
+        byte *bb_start_pc = dr_fragment_app_pc(tag);
+        void *drcontext = dr_get_current_drcontext();
+        instrlist_t *bb = decode_as_bb(drcontext, bb_start_pc);
+        dr_fprintf(STDERR, "Application basic block where mismatch detected: \n");
+        instrlist_disassemble(drcontext, bb_start_pc, bb, STDERR);
```
derekbruening, line 384: Having the mcontext values would also be useful: how about a TODO or XXX comment for now.

ext/drstatecmp/drstatecmp.dox
```diff
@@ -0,0 +1,137 @@
...
+drmgr's post-instrumentation stage. An instrumentation-free copy of each side-effect-free
+basic block is appended immediately after the instrumented version of the basic block.
+Before the execution of the instrumented basic block, its machine state is saved. This
+state will be used for state restoration before the re-execution of the basic block. At
+the end of the instrumented basic block, code is inserted to save the current state for
+later comparison and then restore the machine state to the state before executing the
+instrumented basic block. At the end of the instrumentation-free copy of the basic block,
+a clean call is inserted that compares the current machine state of the
+instrumentation-free basic block with the machine state that was saved at the end of the
+instrumented basic block. An user-supplied callback is invoked in case of any state
```
derekbruening, line 93: nit: s/An user/A user/

suite/tests/CMakeLists.txt
```diff
@@ -3302,6 +3307,11 @@ if (BUILD_CLIENTS)
         "-simulator_type miss_analyzer -miss_count_threshold 5000 -miss_frac_threshold 0.25" "")
     endif ()
 
+    # Stress-test drcachesim with drstatecmp checks.
+    if (AARCH64)
```
derekbruening, line 3311: I know the fuzz app is a64-only -- but isn't this test available for x86?

suite/tests/client-interface/drstatecmp-fuzz-app.c
```diff
@@ -0,0 +1,218 @@
...
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+/* Fuzzing application to stress-test DR with the drstatecmp library.
+ * Only AArch64 is currently supported.
```
derekbruening, line 34: Add ifdefs to `#error` if AARCH64 is not set

suite/tests/client-interface/drstatecmp-fuzz-app.c
```diff
@@ -0,0 +1,218 @@
...
+    instr_disassemble(GLOBAL_DCONTEXT, instr, STDERR);
+    fprintf(stderr, "\n");
+#endif
+}
+
+static byte *
+append_instr(instr_t *instr, byte *encode_pc)
+{
+    print_instr_pc(instr, encode_pc);
+    byte *nxt_pc = instr_encode(GLOBAL_DCONTEXT, instr, encode_pc);
```
derekbruening, line 88: Check the return value to ensure it encoded successfully

suite/tests/client-interface/drstatecmp-fuzz-app.c
```diff
@@ -0,0 +1,218 @@
...
+    fprintf(stderr, "%p: ", encode_pc);
+    instr_disassemble(GLOBAL_DCONTEXT, instr, STDERR);
+    fprintf(stderr, "\n");
+#endif
+}
+
+static byte *
+append_instr(instr_t *instr, byte *encode_pc)
+{
+    print_instr_pc(instr, encode_pc);
```
derekbruening, line 87: nit: Better to avoid the call if verbose is turned off

suite/tests/client-interface/drstatecmp-fuzz-app.c
```diff
@@ -0,0 +1,218 @@
...
+    encode_pc = append_instr(XINST_CREATE_return(GLOBAL_DCONTEXT), encode_pc);
+    assert(encode_pc && encode_pc <= generated_code + code_size);
+    protect_mem(generated_code, code_size, ALLOW_EXEC | ALLOW_READ);
+}
+
+int
+main(void)
+{
+    /* Produce fuzzing application code. */
+    srand(time(NULL));
```
derekbruening, line 181: Add an XXX comment about taking in the seed as a parameter, and printing it out on an error, to reproduce the exact behavior in a subsequent run.

suite/tests/client-interface/drstatecmp-fuzz-app.c
```diff
@@ -0,0 +1,218 @@
...
+ * DAMAGE.
+ */
+
+/* Fuzzing application to stress-test DR with the drstatecmp library.
+ * Only AArch64 is currently supported.
+ */
+
+#include "configure.h"
+#include "dr_api.h"
+#include "opcode_opnd_pairs.h"
```
derekbruening, line 39: Maybe worth a comment here that this is a special file generated from the codec just for this fuzzer

suite/tests/client-interface/drstatecmp-fuzz-app.c
```diff
@@ -0,0 +1,218 @@
...
+generate_encoded_inst(void)
+{
+    /* Pick one of the available (side-effect-free and non-branch) opcodes
+     * and randomize the non-fixed bits.
+     */
+    uint32_t opcode_pick = rand() % FUZZ_INST_CNT;
+    const opcode_opnd_pair_t opcode_opnd_pair = fuzz_opcode_opnd_pairs[opcode_pick];
+
+    uint32_t opnd_mask = opcode_opnd_pair.opnd;
+    /* Avoid registers 16-31 for Rd (special registers and callee-saved registers). */
```
derekbruening, line 114: I think this needs a little more discussion: this reduces the space of errors that will be found.  If you could add a comment discussing the benefits and drawbacks and how you might handle these with additional state saving or sthg.

derekbruening (approved):
May be nice to split this into separate PR's as discussed.

suite/tests/client-interface/drstatecmp-fuzz-app.c
sapostolakis: I added a comment on the reasoning of this constraint and how to relax it.

suite/tests/client-interface/drstatecmp-fuzz-app.c
sapostolakis: Done.

suite/tests/client-interface/drstatecmp-fuzz-app.c
sapostolakis: Done.

suite/tests/client-interface/drstatecmp-fuzz-app.c
sapostolakis: Done.

suite/tests/client-interface/drstatecmp-fuzz-app.c
sapostolakis: Checked with assertion two lines down but moved it immediately after to make it clearer.

suite/tests/client-interface/drstatecmp-fuzz-app.c
sapostolakis: Done.

suite/tests/CMakeLists.txt
sapostolakis: drcachesim with drstatecmp can tested for x86 for a different app. 
I added another drcachesim test with drstatecmp for the the ci_shared_app (so available for all archs).
Btw the code_api|sample.statecmp test is available for all archs.

ext/drstatecmp/drstatecmp.dox
sapostolakis: Done.

ext/drstatecmp/drstatecmp.c
sapostolakis: Done.

ext/drstatecmp/drstatecmp.c
sapostolakis: Done.

ext/drmgr/drmgr.h
```diff
@@ -515,6 +515,79 @@ bool
...
+ * five of them.  See the documentation for drmgr_register_bb_app2app_event(),
+ * drmgr_register_bb_instrumentation_event(), drmgr_register_bb_instru2instru_event(), and
+ * drmgr_register_bb_post_instru_event() for further details of each pass. The
+ * aforementioned routines are identical to this with the exception of the extra \p
+ * user_data parameter, which is an OUT parameter to the \p app2app_func and passed in to
+ * the four subsequent callbacks. The \p priority param can be NULL, in which case a
+ * default priority is used.
+ */
+bool
+drmgr_register_bb_post_instru_ex_event(drmgr_app2app_ex_cb_t app2app_func,
```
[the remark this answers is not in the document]
sapostolakis: Oh I didn't realize the naming convention there. Changed it to drmgr_register_bb_instrumentation_all_events() and ditto for unregister.

core/ir/aarch64/codec.py
```diff
@@ -338,6 +338,39 @@ def generate_opcode_names(patterns):
           '#endif /* OPCODE_NAMES_H */']
     return '\n'.join(c) + '\n'
 
+def generate_opcode_opnd_pairs(patterns):
```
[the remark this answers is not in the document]
sapostolakis: Done.

api/samples/statecmp.c
```diff
@@ -0,0 +1,180 @@
...
+    drreg_options_t drreg_ops = { sizeof(drreg_ops), 1 /*max slots needed: aflags*/,
+                                  false };
+    dr_set_client_name("DynamoRIO Sample Client 'statecmp'",
+                       "http://dynamorio.org/issues");
+    /* Make it easy to tell, by looking at log file, which client executed. */
+    dr_log(NULL, DR_LOG_ALL, 1, "Client 'statecmp' initializing\n");
+    /* To enable state comparison checks by the drstatecmp library, a client simply needs
+     * to initially invoke drstatecmp_init() and drstatecmp_exit() on exit.
+     * The invocation of drstatecmp_init() registers callbacks that insert
+     * machine state comparison checks in the code. An user-provided callback is
```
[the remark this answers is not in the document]
sapostolakis: Done.

api/samples/statecmp.c
```diff
@@ -0,0 +1,180 @@
...
+/* Invoked by drstatecmp when a state comparison fails. */
+static void
+error_callback(const char *msg, void *tag)
+{
+    error_detected = 1;
+    DR_ASSERT_MSG(!strcmp(msg, "xflags"), msg);
+}
+
+static bool
+bb_may_have_side_effects(void *drcontext, instrlist_t *bb)
```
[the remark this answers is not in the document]
sapostolakis: Just to clarify, no other client will need to duplicate drstatecmp internal logic to use drstatecmp. This function does not specify where checks are enabled but it is just an artifact of creating a buggy client that does not lead to a flaky test. This client introduces clobbering of aflags with its instrumentation. To avoid a seg fault or any form of weird behavior, I had to restrain the client to only introduce bugs in basic blocks that drstatecmp could test.  
That said, it is worth exporting this functionality to avoid the redundancy here (and getting out-of-sync with the test client). I also renamed it to drstatecmp_bb_checks_enabled() for future-proofing as suggested.

api/samples/statecmp.c
```diff
@@ -0,0 +1,180 @@
...
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+/* Illustrates use of the drstatecmp extension by a sample client.
+ * This sample client introduces an instrumentation bug that is caught by
+ * drstatecmp. This sample client also shows how to specify an user-defined
```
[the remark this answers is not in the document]
sapostolakis: Done.

sapostolakis:
For x86 in DR an undefined effect on eflags is considered to write to eflags since programs shouldn't be reading eflags after an undefined modification to them. However, drstatecmp was checking the architectural state through re-execution and thus was catching clobbering of the flags by instrumentation in cases where drreg relied on undef behavior to consider the flags to be dead. Avoided checking the flags in cases with dead flags to avoid these "false" positives.

sapostolakis:
Split the new instrumentation phase to a new separate PR (https://github.com/DynamoRIO/dynamorio/pull/5059).

sapostolakis:
As discussed, this PR has been split up into 4 smaller PRs: #5059, #5090, #5107, #5125.

Issue #2210, opened 2017-02-22
AArchXX clean calls handle far too few use cases, blocking tool development


derekbruening:
The AArchXX implementation of insert_parameter_preparation() is missing some key functionality:

```
/* FIXME i#1551, i#1569: we only implement naive parameter preparation,
 * where args are all regs or immeds and do not conflict with param regs.
 */
```

This is a big enough missing port feature to call for a separate issue.  This is blocking tool development on AArchXX, even simple samples: https://github.com/DynamoRIO/dynamorio/pull/2174

Xref original issues #1551, #1569

derekbruening:
This shouldn't have been closed as handling memory operands as arguments is not yet finished.

egrimley:
(Why did it get closed? A bug in GitHub?)

egrimley:
Memory operands will need a somewhat different approach as there will be cases (even if they occur very rarely in practice) in which it is necessary to spill and restore registers during the marshalling: imagine (31 * 31) arguments of the form [Xn, Xm], for example.

We should start by specifying exactly which types of memory operands should be accepted.

derekbruening:
> (Why did it get closed? A bug in GitHub?)

Your commit message closed it by saying "Fixes #2210"

derekbruening:
It seems fine to disallow hard-to-handle memory operands if they're unlikely to be used much by clients.

egrimley:
Really? I can't see those words on this page: https://github.com/DynamoRIO/dynamorio/commit/e06883cbc9515828379109c0ff20695e71b86cc8

derekbruening:
> > > (Why did it get closed? A bug in GitHub?)

> > Your commit message closed it by saying "Fixes #2210"

> Really? I can't see those words on this page: e06883c

Hmm, right, it's not there in the final squash-and-merge commit, only in the first one in the pull request.  Maybe worth filing a github bug.

PR #4946, opened 2021-06-14
mem-ref-for-clean-calls-aarch64 initial

gregcawthorne:
This patch adds new functionality to support using memory references are arguments to clean calls in DR.

I fully expect some issues from upstream to be reported regarding this. I wasn't sure of the best approach to integrating some aspects of this code, in terms of style and locality so didn't over think it in the first attempt.

### DIFF
--- a/core/arch/aarchxx/mangle.c
+++ b/core/arch/aarchxx/mangle.c
@@ -671,7 +671,7 @@ insert_parameter_preparation(dcontext_t *dcontext, instrlist_t *ilist, instr_t *
                            reg_get_size(opnd_get_reg(args[i])) == OPSZ_PTR) ||
                           opnd_is_base_disp(args[i]),
                       "insert_parameter_preparation: bad argument type");
-        ASSERT_NOT_IMPLEMENTED(!opnd_is_base_disp(args[i])); /* FIXME i#2210 */
+        IF_ARM(ASSERT_NOT_IMPLEMENTED(!opnd_is_base_disp(args[i])); /* FIXME i#2210 */)
     }
 
     /* The strategy here is to first set up the arguments that can be set up
@@ -777,7 +777,9 @@ insert_parameter_preparation(dcontext_t *dcontext, instrlist_t *ilist, instr_t *
     for (i = 0; i < num_regs; i++) {
         if (opnd_is_immed_int(args[i]))
             regs[i] = -1;
-        else {
+        else if (opnd_is_base_disp(args[i])){
+            regs[i] = -3;
+        } else {
             reg_id_t reg = opnd_get_reg(args[i]);
             regs[i] = -2;
             for (j = 0; j < NUM_REGPARM; j++) {
@@ -817,6 +819,9 @@ insert_parameter_preparation(dcontext_t *dcontext, instrlist_t *ilist, instr_t *
                     PRE(ilist, instr,
                         instr_create_restore_from_dc_via_reg(
                             dcontext, d_r_regparms[i], d_r_regparms[i], XSP_OFFSET));
+                } else if (regs[i] == -3){
+                    mov_ldr_aarch64(dcontext, ilist, instr,
+                                           opnd_create_reg(d_r_regparms[i]), args[i]);
                 } else {
                     PRE(ilist, instr,
                         XINST_CREATE_move(dcontext, opnd_create_reg(d_r_regparms[i]),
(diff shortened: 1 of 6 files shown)


gregcawthorne:
I see there are some linting errors.

I shall address these.

core/arch/aarchxx/mangle.c
```diff
@@ -777,7 +777,9 @@ insert_parameter_preparation(dcontext_t *dcontext, instrlist_t *ilist, instr_t *
     for (i = 0; i < num_regs; i++) {
         if (opnd_is_immed_int(args[i]))
             regs[i] = -1;
-        else {
+        else if (opnd_is_base_disp(args[i])) {
+            regs[i] = -3;
```
AssadHashmi, line 781: What does -3 mean? Comment block above goes up to -2.

gregcawthorne: -3 is opnd_is_base_disp type (previously unsupported) as described in the condition

gregcawthorne: Will add to the comment

AssadHashmi (commented):
> This patch adds new functionality to support using memory references are arguments to clean calls in DR.

Which issue is the PR related to? #2210?

Our convention is to add the related issue at the bottom of the first comment when the PR is created then include in final commit message when merging, e.g.
```
This patch adds new functionality to support using memory references are arguments to clean calls in DR.

Issue: #2210
```
And at the start of the PR title, e.g.
`i#2210 mem-ref-for-clean-calls-aarch64 initial #4946`

derekbruening:
> > This patch adds new functionality to support using memory references are arguments to clean calls in DR.
> 
> Which issue is the PR related to? #2210?
> 
> Our convention is to add the related issue at the bottom of the first comment when the PR is created then include in final commit message when merging, e.g.
> 
> ```
> This patch adds new functionality to support using memory references are arguments to clean calls in DR.
> 
> Issue: #2210
> ```

+1, I found this confusing because memory references are already supported as clean call arguments in general: please make it clear that this is for AArch64 only where this support was missing.

gregcawthorne:
Sorry, yes the issue was #2210

For aarch64 it is my understanding memory references are not supported in clean calls.

Thanks Assad I will go through those issues.

core/arch/aarchxx/mangle.c
johnfxgalea: I think it would be better if these were enums or at least macro defs to aid readability.

core/ir/aarch64/instr.c
```diff
@@ -329,6 +329,14 @@ instr_predicate_is_cond(dr_pred_type_t pred)
     return pred != DR_PRED_NONE && pred != DR_PRED_AL && pred != DR_PRED_NV;
 }
 
+/* Segmenting not available on AArch64
```
johnfxgalea, line 332: Nit: Having */ on same line looks nicer since the comment is a one liner.

core/ir/aarch64/instr_create_api.h
```diff
@@ -464,6 +529,21 @@ enum {
...
+ * \param Rd The destination register opnd_t operand.
+ * \param Rn The source register opnd_t operand.
+ * \param rm_or_imm The second input register or immediate.
+ */
+// TODO: this imm path might not be 100% right
+#define INSTR_CREATE_lsl(dc, Rd, Rn, rm_or_imm)                           \
+    (opnd_is_reg(rm_or_imm)                                               \
+         ? instr_create_1dst_2src((dc), OP_lslv, (Rd), (Rn), (rm_or_imm)) \
+         : instr_create_1dst_3src((dc), OP_extr, (Rd), (Rn),              \
+                                  OPND_CREATE_ZR(Rn)                      \
```
johnfxgalea, line 544: is there a missing `,` here ?

core/ir/aarchxx/ir_utils.c
```diff
@@ -162,6 +162,198 @@ convert_to_near_rel_arch(dcontext_t *dcontext, instrlist_t *ilist, instr_t *inst
...
+}
+
+/* Idea is to add or subtract a mem addresses disp,
+ * into its base register, in the case the range is too far
+ * for the instruction without the use of a another index register.
+ */
+void
+mem_access_fixer(void *drcontext, instrlist_t *bb, instr_t *inst, opnd_t mem, bool before)
+{
+    // If it has an index reg we dont have a disp?
```
johnfxgalea, line 182: Nit: Do not use `//` but `/*` and `*/` instead.

core/lib/dr_ir_utils.h
```diff
@@ -998,6 +998,34 @@ bool
...
+ * with operand size detection and overcomes a displacement being
+ * too far for to be encoded in a single instruction,
+ * without the use of a another index register.
+ */
+DR_API
+void
+mov_ldr_aarch64(void *drcontext, instrlist_t *bb, instr_t *inst, opnd_t dst, opnd_t src);
+
+/* Instruments a branch to a specific value, allowing for the value
+ * operand to be supplied inside a register, immidiate,
```
johnfxgalea, line 1022: Typo in comment.

core/ir/opnd_shared.c
```diff
@@ -1997,7 +1997,8 @@ reg_get_value_ex(reg_id_t reg, dr_mcontext_t *mc, OUT byte *val)
         *(reg_t *)val = regval;
     }
 #else
-    CLIENT_ASSERT(false, "NYI i#1551");
+    reg_t regval = reg_get_value(reg, mc);
```
johnfxgalea, line 2000: With some small refactoring of the function, you can share this code with those between lines 1996-1997.

johnfxgalea (changes requested):
Thanks for the PR! It would be great if you could add tests for the new aarch64 support!

PR #4949, opened 2021-06-18
i#3823 multi-phase drreg: Remove slot id labels.

abhinav92003:
Modifies free spill slot selection logic to use is_our_spill_or_restore instead
of the labels with spill slot use information added for this purpose. We do not
need the extra information in the latter and can simply use the former routine.

Removes the extra spill slot use metadata added in form of labels from the
instrlist.

Also adds documentation about possibility of DR slot conflicts if DR APIs are
mixed with drreg ones.

Issue: #3823

### DIFF
--- a/ext/drreg/drreg.dox
+++ b/ext/drreg/drreg.dox
@@ -53,6 +53,11 @@ less available in future phases; the current logic skips over a slot if
 there's a usage found anywhere later in the bb added by any previous phase.
 So it requires additional spill slots as well.
 
+Note that DR provides APIs to directly access TLS slots: like
+\p dr_insert_write_raw_tls(), \p dr_insert_write_tls_field(). These are
+also used by drreg internally. It is unsafe to mix usages of drreg and
+these direct APIs as it may cause conflicts in slot usages.
+
 CAUTION: this extension is still a work in progress.  Not all pieces are
 implemented yet, and the interface may be in flux until the details are
 finalized.
--- a/ext/drreg/drreg.dox
+++ b/ext/drreg/drreg.dox
@@ -53,14 +53,16 @@ less available in future phases; the current logic skips over a slot if
 there's a usage found anywhere later in the bb added by any previous phase.
 So it requires additional spill slots as well.
 
-Note that DR provides APIs to directly access TLS slots: like
-\p dr_insert_write_raw_tls(), \p dr_insert_write_tls_field(). These are
-also used by drreg internally. It is unsafe to mix usages of drreg and
-these direct APIs as it may cause conflicts in slot usages.
-
-CAUTION: this extension is still a work in progress.  Not all pieces are
-implemented yet, and the interface may be in flux until the details are
-finalized.
+Note that DR provides other APIs to save or restore regs, like \p
+dr_save_reg() and \p dr_restore_reg(), and the corresponding \p
+dr_read_saved_reg() and \p dr_write_saved_reg(). They use DR spill slots.
+drreg may also use DR spill slots if not given enough dedicated slots via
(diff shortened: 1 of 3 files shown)


abhinav92003:
AArch64 failure for `code_api|tool.drcachesim.threads-with-config-file` seems flaky; works fine locally. Will re-run.

abhinav92003:
run arm tests

ext/drreg/drreg.dox
```diff
@@ -53,6 +53,11 @@ less available in future phases; the current logic skips over a slot if
 there's a usage found anywhere later in the bb added by any previous phase.
 So it requires additional spill slots as well.
 
+Note that DR provides APIs to directly access TLS slots: like
+\p dr_insert_write_raw_tls(), \p dr_insert_write_tls_field(). These are
```
derekbruening, line 57: This is incorrect: dr_insert_write_tls_field uses the single client field which is unrelated to drreg (except via drmgr multiplexing it to provide drmgr_register_tls_field).

derekbruening: dr_insert_write_raw_tls is perfectly safe to use, and many clients do use it, for slot ranges reserved by dr_raw_tls_calloc.  Only if someone used a slot range they did not actually reserve, which doesn't seem like we need to call it out here.

ext/drreg/drreg.dox
```diff
@@ -53,6 +53,11 @@ less available in future phases; the current logic skips over a slot if
...
+also used by drreg internally. It is unsafe to mix usages of drreg and
+these direct APIs as it may cause conflicts in slot usages.
+
 CAUTION: this extension is still a work in progress.  Not all pieces are
```
derekbruening, line 61: I think we can remove this.

ext/drreg/drreg.dox
derekbruening: The problem is if drreg is not given enough of its own slots via `#drreg_options_t.num_spill_slots` and so ends up using DR spill slots, and a client simultaneously directly uses DR's spill slots via dr_save_reg()/dr_restore_reg() (and the corresponding dr_read_saved_reg()/dr_write_saved_reg()).

ext/drreg/drreg.h
```diff
@@ -117,9 +117,12 @@ typedef struct _drreg_options_t {
      * requested from DR via dr_raw_tls_calloc().  Any slots needed beyond
      * this number will use DR's base slots, which are not allowed to be
      * used across application instructions.  DR's slots are also more
-     * expensive to access (beyond the first few).
-     * This number should be computed as one plus the number of
-     * simultaneously used general-purpose register spill slots, as
+     * expensive to access (beyond the first few). DR's base slots may
+     * also be used by core DR or other clients, which may cause
```
derekbruening, line 121: Core DR can be removed here: that's the source of the "not allowed to be used across application instructions"; beyond that there will be no conflict.

derekbruening: For "other clients": I would stress that the problem is when using a client or library that doesn't use drreg and instead directly using DR register spill slots (dr_save_reg, etc.); i.e., all cooperating client components should be using drreg.

abhinav92003: Done.

ext/drreg/drreg.dox
abhinav92003: Thanks. Done.

ext/drreg/drreg.dox
abhinav92003: Done.

derekbruening (approved):

PR #4956, opened 2021-06-18
i#4719 qemu: Replace WFI with WFE for QEMU interop

derekbruening:
QEMU crashes when executing the WFI AArchXX instruction.
DR uses WFI in its spinlock loops.
We replace WFI with WFE here, which is roughly a superset
of WFI and should work similarly for us while not
breaking QEMU.

This fixes over a dozen tests that previously failed under QEMU with
"unhandled CPU exception 0x10001".  Here we add 17 tests on AArch64
and 14 tests on AArch32 to the RUNS_ON_QEMU label.

Issue: #4719


derekbruening:
Well unfortunately nearly all the ARM QEMU tests that pass locally still fail on GA CI with various asserts.  The difference may be the QEMU version, or something else.  I guess I'll have to remove those for now.  For A64: at least nearly all of those also pass on GA CI.

derekbruening:
We're down to just 2 arm additions, and 12 aarch64.  If we can repro and fix the memcache assert we can re-add most of these but I couldn't reproduce it locally even with the same qemu version as GA CI:
```
2021-06-18T21:54:04.7387119Z 65: <Application /home/runner/work/dynamorio/dynamorio/build_arm-debug-internal/suite/tests/bin/linux.eintr (13785).  Internal Error: DynamoRIO debug check failure: /home/runner/work/dynamorio/dynamorio/core/unix/memcache.c:382 found
```

abhinav92003:
> If we can repro and fix the memcache assert we can re-add most of these but I couldn't reproduce it locally even with the same qemu version as GA CI:

You can try using tmate to ssh into the runner and get more logs out: https://github.com/marketplace/actions/debugging-with-tmate

abhinav92003 (approved):

derekbruening:
> > If we can repro and fix the memcache assert we can re-add most of these but I couldn't reproduce it locally even with the same qemu version as GA CI:
> 
> You can try using tmate to ssh into the runner and get more logs out: https://github.com/marketplace/actions/debugging-with-tmate

Let's separate that out; for now would like to get this 6-month-old fix in.  If there's not already an issue on the memcache assert (thought it was at least listed under a qemu issue) will file a new one.

derekbruening:
https://github.com/DynamoRIO/dynamorio/issues/4719#issuecomment-1015896276

Issue #4954, opened 2021-06-18
tool.drcachesim.threads-with-config-file occasional timeout on A64 Jenkins


derekbruening:
Hit a timeout in PR #4952 on tool.drcachesim.threads-with-config-file: took 90.00s:
http://139.178.84.19:8080/job/DynamoRIO-AArch64-Precommit/98/consoleFull

```
$
] 90.00 sec

    -------------------------------------------------------------------
     Performance for solving AX=B Linear Equation using Jacobi method
     Running on DynamoRIO
     Client version (null)
    ...................................................................

     Matrix Size :  64
     Threads     :  4


     Started iteration 1 of the computation...

     Finished computing current solution distance in mode 0.
     Mode changed to 0.

     Started iteration 2 of the computation...

     Finished computing current solution distance in mode 0.
     Mode changed to 0.

     Started iteration 3 of the computation...
```

It ends there, in iteration 3.

This happened on PR #4949 too: http://139.178.84.19:8080/job/DynamoRIO-AArch64-Precommit/95/

It passed on re-running and took very little time:
```
41/50 Test #177: code_api|tool.drcachesim.threads-with-config-file ...............   Passed    0.30 sec
```

So it seems like it's not just always close the time limit and it's maybe an actual hang or something?

derekbruening:
The coherence test hit the same hang.  It runs the same app so likely the same problem:

http://139.178.84.19:8080/job/DynamoRIO-AArch64-Precommit/126/consoleFull
```
50/50 Test #178: code_api|tool.drcachesim.coherence ..............................***Timeout 150.02 sec

    -------------------------------------------------------------------
     Performance for solving AX=B Linear Equation using Jacobi method
     Running on DynamoRIO
     Client version (null)
    ...................................................................

     Matrix Size :  64
     Threads     :  4


     Started iteration 1 of the computation...

     Finished computing current solution distance in mode 0.
     Mode changed to 0.

     Started iteration 2 of the computation...
```

johnfxgalea:
We hit this again: http://139.178.84.19:8080/job/DynamoRIO-AArch64-Precommit/147/console

derekbruening:
Xref #3971 where coherence hung on Windows.

Hung again on AArch64 Jenkins: http://139.178.84.19:8080/job/DynamoRIO-AArch64-Precommit/161/consoleFull

abhinav92003:
[...]

PR #4961, opened 2021-06-22
i#4857 icount: Include instr count in filtered trace unit header

### LINKED ISSUE #4857: Feature Request:  Add a count for total number of instruction for filtered trace
When traced in filtered mode, the total number of records does not equal to the number of instructions executed during tracing.   It would be very useful to have accurate instruction count for filtered trace. At the minimum, it should provide a total number of instructions executed during the tracing period.   Additional details such as # of instructions per thread, or # of instructions since last record (i.e. for every filtered miss) would great with the understanding that these will add additional overhead in speed and size.

This feature would be extremely useful to try to correlate filtered trace with unfiltered trace and would enable filtered trace be more useful for cache studies.

derekbruening:
Adds the instruction count to the buffer header for drcachesim
filtered traces, instead of just at thread exit, to aid in correlating
filtered and unfiltered traces.

Updates the documentation and the trace_invariants test.

Augments the view tool to print out these markers.
Further verified manually on a larger trace with the view tool:

    <marker: filetype 65>
    <marker: cache line size 64>
    <marker: timestamp 13268850698443940>
    <marker: tid 2539348 on core 1>
    <marker: instruction count 33326>
      0x00007f6f64314090  48 89 e7             mov    %rsp, %rdi
      0x00007f6f6432dce0  b8 0c 00 00 00       mov    $0x0000000c, %eax
    <marker: timestamp 13268850698727951>
    <marker: tid 2539348 on core 1>
    <marker: instruction count 40130>
      0x00007f6f6432c906  c7 44 24 3c 00 00 00 movl   $0x00000000, 0x3c(%rsp)
    <...>
      0x00007f6f63cc3670  4c 8b 05 f9 27 0f 00 mov    <rel> 0x00007f6f63db5e70, %r8
    <marker: timestamp 13268850699555438>
    <marker: tid 2539348 on core 1>
    <marker: instruction count 126115>

Fixes #4857

### DIFF
--- a/clients/drcachesim/common/options.cpp
+++ b/clients/drcachesim/common/options.cpp
@@ -167,7 +167,8 @@ droption_t<bool> op_L0_filter(
     "that miss in this initial cache.  This cache is direct-mapped with sizes equal to "
     "-L0I_size and -L0D_size.  It uses virtual addresses regardless of -use_physical. "
     "The dynamic (pre-filtered) per-thread instruction count is tracked and supplied "
-    "via a #TRACE_MARKER_TYPE_INSTRUCTION_COUNT marker at thread exit.");
+    "via a #TRACE_MARKER_TYPE_INSTRUCTION_COUNT marker at thread buffer boundaries "
+    "and at thread exit.");
 
 droption_t<bytesize_t> op_L0I_size(
     DROPTION_SCOPE_CLIENT, "L0I_size", 32 * 1024U,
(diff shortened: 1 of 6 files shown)


derekbruening:
Jenkins failures: threads-with-config file again #4954 plus coherence which also timed out!  Added that to #4954 as it's the same test app, so likely the same bug.

sapostolakis (approved):

derekbruening:
run arm tests

Issue #4965, opened 2021-06-24
drbbdup diagram


johnfxgalea:
The diagram included in the documentation of drbbdup is no longer being shown.

Doc: https://dynamorio.org/page_drbbdup.html
Image: https://github.com/DynamoRIO/dynamorio/blob/master/api/docs/images/drbbdup_image.svg

derekbruening:
`Failed to load resource: the server responded with a status of 404 ()`

@johnfxgalea IIRC there was some shift to put images into a subdir but it was a manual process: if you could add the image to the images/ dir of dynamorio.github.io that should solve it

johnfxgalea:
@derekbruening Does the diagram need to be removed from the main repo?

derekbruening:
> @derekbruening Does the diagram need to be removed from the main repo?

No, I think doxygen will complain if it's not there.  (Yes, non-ideal duplication...maybe moving out of images/ would solve?  IMHO not worth messing with at this point.)

johnfxgalea:
https://github.com/DynamoRIO/dynamorio.github.io/pull/24

derekbruening:
I ran the ci-docs Action to regenerate ahead of next week's cronjob so the live page is now fixed.

PR #4971, opened 2021-06-26
i#4965: Add SVG data path modification.

johnfxgalea:
Adds a regex command for docs to find SVG data in /images.

Fixes: #4965


johnfxgalea:
We could have avoided this PR if we just simply made the drbbdup image a **.png**, but **.svg** has its benefits too, so I guess it is worthwhile.

derekbruening (approved):

johnfxgalea:
run arm tests

PR #4972, opened 2021-06-28
Fix quoting mishap in build-from-source docs

derekbruening:
Apparently Doxygen's markdown parser cannot handle a single backtic
inside a triple-backtic quote block.  This was causing a messed-up
sequence of what is quoted and what is not and a missed final section
on the build-from-source web page.  We fix it here by switching to $()
instead of backtics in a bash function; we also add a named link to
the DrGUI build directions since Dr. Memory links to them.

### DIFF
--- a/api/docs/building.dox
+++ b/api/docs/building.dox
@@ -821,7 +821,7 @@ Here is a sample bash function to make building easier within Cygwin:
 
 ```
 function build {
-    config=`grep ^CMAKE_CONFIGURATION_TYPES:S CMakeCache.txt | sed 's/^.*=//'`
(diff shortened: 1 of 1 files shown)


sapostolakis (approved):

derekbruening:
coherence test hung again on AArch64 which is disturbing: #4954

derekbruening:
Note that my local doxygen 1.9.1 does properly handle the backtix -- so it could be that we could avoid this type of problem (seeing similar issues w/ the recent Dr. Memory docs where local looks good and GA is bad) if we installed the very latest.  The ci-docs workflows use ubunut-20.04 from GA but there doxygen is just 1.8.17.

Issue #4963, opened 2021-06-23
32-bit tool.drcacheoff.burst_threads fails with "Cannot correctly handle received signal 11"


derekbruening:
This happened 2x in a row on GA CI in PR #4957:

```
2021-06-22T22:54:46.2847759Z 295: CMake Error at /home/runner/work/dynamorio/dynamorio/suite/tests/runmulti.cmake:106 (message):
2021-06-22T22:54:46.2848637Z 295:   *** cmd failed (40): pre-DR init
2021-06-22T22:54:46.2849042Z 295: 
2021-06-22T22:54:46.2849496Z 295:   pre-DR start
2021-06-22T22:54:46.2849856Z 295: 
2021-06-22T22:54:46.2850331Z 295:   pre-DR detach
2021-06-22T22:54:46.2850663Z 295: 
2021-06-22T22:54:46.2851124Z 295:   pre-DR init
2021-06-22T22:54:46.2851475Z 295: 
2021-06-22T22:54:46.2851915Z 295:   pre-DR start
2021-06-22T22:54:46.2852271Z 295: 
2021-06-22T22:54:46.2852741Z 295:   pre-DR detach
2021-06-22T22:54:46.2853069Z 295: 
2021-06-22T22:54:46.2853418Z 295:   <Application
2021-06-22T22:54:46.2854440Z 295:   /home/runner/work/dynamorio/dynamorio/build_debug-internal-32/clients/bin32/tool.drcacheoff.burst_threads
2021-06-22T22:54:46.2855393Z 295:   (21574).  Cannot correctly handle received signal 11 in thread 21575:
2021-06-22T22:54:46.2855986Z 295:   default action in native thread.>
```

In my Ubuntu 16.04 VM: I can reproduce this in TOT, so unrelated to that PR.  It reproduces 5x out of 100.

derekbruening:
@abhinav92003 -- I believe this is related to PR #4944.  I reproduced with -loglevel 3 in this VM and the thread that crashes has a bunch of drreg restoration during the detach.  I then rolled back to 4b6d1ca39fc870cd6f20618f65f5d999eb80e734, the commit prior to PR #4944, and in 100 runs there is no failure, when previously I was seeing 5 failures per 100.  That's not conclusive: I'll do another 1000 runs in the VM for better confirmation.  But it lends weight to the theory.

Are you able to reproduce on any machine you have access to?

abhinav92003:
[...]

PR #4976, opened 2021-06-29
Simplify drreg state restoration code.

abhinav92003:
Rearranges some code to obviate the need for the *_now variables. Also separates out the common instr-is-meta check to an outer if-condition.

This is in preparation for sending the i#4963 fix for review. Split out this harmless refactor to make review easier for that fix.

Issue: #4963

### DIFF
--- a/ext/drreg/drreg.c
+++ b/ext/drreg/drreg.c
@@ -2146,50 +2146,52 @@ drreg_event_restore_state_with_ilist(void *drcontext, bool restore_memory,
             }
         }
 
-        /* Update book-keeping for gpr/aflags spill slot or aflags spill reg. */
-        if (!instr_is_app(inst) &&
-            is_our_spill_or_restore(drcontext, inst, &spill, &reg, &slot, &offs)) {
-            if (spill) {
-                /* Slot used in current instr may have a previous spilled gpr/aflags
-                 * value that may already have been restored. Forget that.
-                 */
-                spill_slot_to_reg[slot] = DR_REG_NULL;
-                if (aflags_spill_reg == reg) {
-                    ASSERT(!gpr_native[GPR_IDX(aflags_spill_reg)],
-                           "reg with aflags cannot be native");
-                    spill_slot_to_reg[slot] = AFLAGS_ALIAS_REG;
-                } else if (gpr_native[GPR_IDX(reg)]) {
-                    spill_slot_to_reg[slot] = reg;
+        /* Update book-keeping for gpr/aflags spill slot and aflags spill reg. Mark
+         * as native the gprs/aflags that were restored. All spills/restores are meta
+         * instrs.
+         */
+        if (!instr_is_app(inst)) {
+            if (is_our_spill_or_restore(drcontext, inst, &spill, &reg, &slot, &offs)) {
+                if (spill) {
+                    /* Slot used in current instr may have a previous spilled gpr/aflags
+                     * value that may already have been restored. Forget that.
+                     */
+                    spill_slot_to_reg[slot] = DR_REG_NULL;
+                    if (aflags_spill_reg == reg) {
+                        ASSERT(!gpr_native[GPR_IDX(aflags_spill_reg)],
+                               "reg with aflags cannot be native");
+                        spill_slot_to_reg[slot] = AFLAGS_ALIAS_REG;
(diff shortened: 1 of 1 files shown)


abhinav92003:
AArch64 failure is `code_api|tool.drcachesim.threads-with-config-file`#4954.

abhinav92003:
x86-32 failure is `code_api|tool.drcacheoff.burst_threads`, which is #4963 itself.

ext/drreg/drreg.c
```diff
@@ -2107,80 +2107,26 @@ drreg_event_restore_state_with_ilist(void *drcontext, bool restore_memory,
...
-            }
-        }
-
         /* Mark gprs/aflags as containing native or non-native value based on whether
          * the current instr is an app or tool write, respectively. An example of the
          * latter is a restore for a spilled tool value, which may happen in the
          * multi-phase nested reservation case.
-         * If the current instr is a tool write that restores an app value (as tracked
-         * by the various *_now bools), we skip marking the gpr/aflags as non-native.
+         * Later we check whether the current instr restores native app value for any
```
derekbruening, line 2114: nit: s/native/the native/

ext/drreg/drreg.c
```diff
@@ -2107,80 +2107,26 @@ drreg_event_restore_state_with_ilist(void *drcontext, bool restore_memory,
             last_opcode = instr_get_opcode(inst);
```
derekbruening, line 2107: > This is in preparation for sending the i#4963 fix for review. Split out this harmless refactor to make review easier for that fix.

Add `Issue: #4963`

ext/drreg/drreg.c
```diff
@@ -2107,80 +2107,26 @@ drreg_event_restore_state_with_ilist(void *drcontext, bool restore_memory,
...
-        }
-
         /* Mark gprs/aflags as containing native or non-native value based on whether
          * the current instr is an app or tool write, respectively. An example of the
          * latter is a restore for a spilled tool value, which may happen in the
          * multi-phase nested reservation case.
-         * If the current instr is a tool write that restores an app value (as tracked
-         * by the various *_now bools), we skip marking the gpr/aflags as non-native.
+         * Later we check whether the current instr restores native app value for any
+         * gpr/aflags, and if it does we mark the written gpr/aflags as native.
```
derekbruening, line 2115: This sounds like a behavior change, and not just refactoring?

Before: "we skip marking as non-native" for "a tool write that restores".
Now: we *do* mark as native.

Yet the PR description just says it's a (non-semantic-changing) refactor?

ext/drreg/drreg.c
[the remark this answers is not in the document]
abhinav92003: Done.

ext/drreg/drreg.c
abhinav92003: Done.

ext/drreg/drreg.c
abhinav92003: I intentionally omitted it, because this PR isn't strictly related to that issue. It's some code simplification that makes sense even otherwise.

Added the issue# to PR description anyway.

ext/drreg/drreg.c
abhinav92003: The final outcome is the same. Although we do mark a restored gpr as non-native initially, but the "is this a restore?" check (which is moved to later in this PR) will eventually mark it as native; this is described in a comment there. This is simpler as it avoids any need of the bool flag (the *_now variables). There is no correctness issue due to this reordering because for restore instrs, `gpr_native` is not read.

Before:
<mark restored gpr as native; set flag F>
<unless F set, if meta instr, mark all dest gpr as non-native>

Now:
<if meta instr, mark all dest gpr as non-native>
<mark restored gpr as native>

abhinav92003: Similar reasoning shows that the final outcome for other vars (`aflags_native` and `aflags_reg`) is also the same.

derekbruening: OK.

derekbruening (approved):

abhinav92003:
run arm tests

abhinav92003:
`code_api|tool.drcachesim.threads-with-config-file` failure on AArch64 again.

abhinav92003:
run arm tests

PR #4987, opened 2021-07-01
i#4963 drreg state restoration: observe non-drreg spills

abhinav92003:
Makes drreg aware of some non-drreg spills/restores that use non-drreg TLS slots
or mcontext slots. While scanning through the ilist during state restoration, if
the current instr is not a drreg spill/restore but is one to a non-drreg TLS or
mcontext slot, we update our book-keeping to correctly reflect whether the
restored gpr is native or not. Even though we do not need to restore non-drreg
spills, we still need to keep track of them. This is required to properly
differentiate between tool value and app value spills (a spill is an app value
spill if the gpr contains its native value at that point), which is particularly
important if there's multi-phase drreg use.

If drreg doesn't understand some non-drreg restore, it will mark the gpr/aflags
non-native. Any later drreg spills will then be incorrectly assumed to be tool
value spills. This was the root-cause of i#4963.

Detects spills to mcontext slots by keeping track of whether any gpr holds the
mcontext base, using the now public API instr_is_load_mcontext_base. Also adds a
new API (instr_is_reg_spill_or_restore_ex) that takes in the gpr that currently
has the mcontext base and also returns whether the given instr is an mcontext
spill,

Adds tests to verify correct restoration of gpr and aflags native value when
the gpr/aflags were spilled/restored previously using dr_save_reg/dr_restore_reg.
Also adds a test where we instrument a clean call, and then reserve the gpr that
was used to hold the mcontext base during the clean call instrumentation. This
reproduces the i#4963 issue.

Verified that this indeed fixes the `tool.drcacheoff.burst_threads` failure. No
failure observed in 500 runs on a Github Actions runner.

Detecting spills/restores using stack will be done separately. Stack spills are
seen in some cases of clean call instrumentation on AArch64.

Issue: #4963


core/ir/instr_api.h
```diff
@@ -2337,6 +2337,11 @@ DR_API
 instr_t *
 instr_create_pusha(void *drcontext);
 
+DR_API
+/** Returns whether \p instr loads mcontext base to some reg. */
+bool
+instr_is_load_mcontext_base(instr_t *instr);
```
abhinav92003, line 2343: Is it okay to expose this information in a public API? If not, then the alternative is that we pass the ilist to `instr_is_reg_spill_or_restore_ex` and let it figure out which reg contains mcontext base.

ext/drreg/drreg.c
```diff
@@ -2047,43 +2063,78 @@ drreg_event_restore_state_with_ilist(void *drcontext, bool restore_memory,
...
+    /* Tracks whether a gpr/aflags has its native app value or not. */
     bool gpr_native[DR_NUM_GPR_REGS];
     bool aflags_native;
-    int last_opcode = 0;
+    /* Tracks the offset of each gpr's spill slot. This is for non-drreg spills
+     * only. We do not restore these spills. This is only to help us keep track
+     * of the native gpr value. For simplicity, we assume that non-drreg
+     * instrumentation will spill a gpr only to a single slot (no multiple
+     * simultaneous spills).
+     * XXX: Handling multiple simultaneous spills requires keeping track of
```
abhinav92003, line 2106: I can probably use `hashtable_t` for this. But to me, this assumption seemed reasonable -- that there wouldn't be multiple simultaneous spills of app value using `dr_save_reg` (In drreg, this happens due to multi phase use only). What do you think?

ext/drreg/drreg.c
```diff
@@ -2140,17 +2201,33 @@ drreg_event_restore_state_with_ilist(void *drcontext, bool restore_memory,
...
+         * mark as native the gprs/aflags that were restored by drreg. All spills/restores
+         * are meta instrs.
          */
         if (!instr_is_app(inst)) {
+            /* There are multiple types of instrs that may spill or restore the gpr/aflags
+             * native value:
+             * - drreg spill or restore: detected using is_our_spill_or_restore
+             * - non-drreg spill or restore to mcontext slot or TLS slot not reserved by
+             *   drreg: if above is false, detected using instr_is_reg_spill_or_restore_ex
+             * - non-drreg spill or restore to stack: e.g. this may happen in some cases
```
abhinav92003, line 2220: Detecting offset for stack spills (future work; not in this PR) seems tricky. For x86, it may perhaps be easier (keep a relative offset which starts at zero, which is incremented when we see a `push` and decremented on `pop`, and use that as offset when we see a stack spill), but for AArch64 it's much more complex. We'd need to look for ldp/stp with not just the stack pointer, but any reg that's copied from sp (something like `mov x0 <- sp+10` means that stores to `x0` are also essentially spills to stack), and also more complex addressing modes.

As commented, I see stack spills happening in some cases for clean call instrumentation on AArch64.

abhinav92003: This doesn't seem like a blocker in fixing i#4963 though. Also, the current restore logic works fine for the clean call in drcachesim.

suite/tests/client-interface/drreg-test.dll.c
```diff
@@ -369,6 +369,28 @@ check_const_ne(ptr_int_t reg, ptr_int_t val)
     CHECK(reg != val, "register value not preserved");
 }
 
+static void
+empty_clean_call()
+{
+    /* TODO: fix memory leak when the following statement is removed and this
```
abhinav92003, line 375: Not sure if this is a known issue. If not, I'll file a new one.

ext/drreg/drreg.c
abhinav92003: Decided to tackle stack spills in a separate PR.

abhinav92003:
The fact that drreg's state restoration logic needs to recognise non-drreg spills/restores is a bit non-ideal. But, I think most users would use drreg for making their reg/aflags reservations, and even if they do mix APIs like `dr_save_reg`/`dr_restore_reg` with drreg, it should be okay as we can recognise these just fine.

The only problem I see is with DR-internal code using other special ways to spills regs, like:
- using stack: discussed more in other comments 
- maybe other weird spills like spilling to some other reg temporarily? If this seems like an issue, we can easily augment our logic to detect reg to reg app value spills.

derekbruening:
> The fact that drreg's state restoration logic needs to recognise non-drreg spills/restores is a bit non-ideal. But, I think most users would use drreg for making their reg/aflags reservations, and even if they do mix APIs like `dr_save_reg`/`dr_restore_reg` with drreg, it should be okay as we can recognise these just fine.
> 
> The only problem I see is with DR-internal code using other special ways to spills regs, like:
> 
> * using stack: discussed more in other comments
> * maybe other weird spills like spilling to some other reg temporarily? If this seems like an issue, we can easily augment our logic to detect reg to reg app value spills.

The approach I would expect drreg to take is to look only for its own spills and restores and not have any need to try to recognize other spills.  If drreg never used DR slots, that approach should work, and it would never need to try to figure out someone else's (like core DR's) spills, right?  The only reason it's trying to look at *everything* is for the case where it doesn't have enough slots and it falls back on DR slots?  Should we consider dropping support for drreg using DR slots (limited slots on Windows are the concern, or mistakes in clients computing how many they need: so not a simple decision)?

core/ir/instr_api.h
```diff
@@ -2337,6 +2337,11 @@ DR_API
 instr_t *
```
derekbruening, line 2337: > spill,

s/,/./

core/ir/instr_api.h
```diff
@@ -2337,6 +2337,11 @@ DR_API
...
 instr_create_pusha(void *drcontext);
```
derekbruening, line 2338: > Adds tests to verify correct restoration of gpr and aflags native value when

s/value/values/

core/ir/instr_shared.c
```diff
@@ -3733,21 +3744,40 @@ instr_is_reg_spill_or_restore_ex(void *drcontext, instr_t *instr, bool DR_only,
     return false;
 }
 
+DR_API
+bool
+instr_is_load_mcontext_base(instr_t *inst)
+{
+    if (instr_get_opcode(inst) != OP_load || !opnd_is_base_disp(instr_get_src(inst, 0)))
```
derekbruening, line 3751: This is not strict enough: for x86 it should check for a *far* base disp with the TLS segment; for arm, the stolen reg base.

derekbruening (commented):
I started looking at the code but I keep getting hung up on whether there's another way to do this...sorry, probably won't finish the review promptly; may spill to tomorrow.

core/ir/instr_api.h
```diff
@@ -2337,6 +2337,11 @@ DR_API
...
 
+DR_API
+/** Returns whether \p instr loads mcontext base to some reg. */
+bool
+instr_is_load_mcontext_base(instr_t *instr);
+
```
derekbruening, line 2344: Re-starting a thread here where we have better threading and tracking:

Could we avoid drreg having to recognize *all* register moves, including DR's,  with just a couple of label instructions added to the ilist at slot-changing points or something?  Trying to figure out what the alternative are.

abhinav92003:
> I started looking at the code but I keep getting hung up on whether there's another way to do this...sorry, probably won't finish the review promptly; may spill to tomorrow.

That's fine. I'm testing a slightly different approach that doesn't need to recognize non-drreg spills/restores. I'll request a review again when ready.

core/ir/instr_api.h
```diff
@@ -2337,6 +2337,11 @@ DR_API
 instr_t *
 instr_create_pusha(void *drcontext);
 
+DR_API
+/** Returns whether \p instr loads mcontext base to some reg. */
```
derekbruening, line 2341: nit: I think this needs to explain what "mcontext" is and why this would be useful and where; plus it should include articles (s/mcontext/the mcontext/).

core/ir/instr_api.h
```diff
@@ -2351,6 +2356,16 @@ bool
 instr_is_reg_spill_or_restore(void *drcontext, instr_t *instr, bool *tls OUT,
                               bool *spill OUT, reg_id_t *reg OUT, uint *offs OUT);
 
+DR_API
+/**
+ * Same as \p instr_is_reg_spill_or_restore but also considers spills/restores using
+ * mcontext base in given reg.
```
derekbruening, line 2362: Ditto: is this "dr_mcontext_t"; when would this happen; how would the user know what reg to pass?

core/lib/instrument.c
```diff
@@ -5720,14 +5720,17 @@ dr_restore_arith_flags_from_reg(void *drcontext, instrlist_t *ilist, instr_t *wh
     CLIENT_ASSERT(reg == DR_REG_XAX,
                   "only xax should be used for save arith flags in X86");
     /* flag restoring code:
-     *   add 0x7f,%al
+     *   cmp %al $0x81
```
derekbruening, line 5723: Should OP_add be removed from instr_is_trace_cmp() then?

ext/drreg/drreg.c
```diff
@@ -79,6 +79,14 @@
  */
 #define AFLAGS_ALIAS_REG (DR_REG_STOP_GPR + 1)
 
+/* Used to indicate that "no offset" is being used. */
```
derekbruening, line 82: Used where?  Only in gpr_to_non_drreg_slot_offs[]?  Move the def down next to that var?

ext/drreg/drreg.c
```diff
@@ -79,6 +79,14 @@
...
+#define NO_OFFS 0xffffffff
+
+/* Value used as base to help differentiate the offset for a TLS and mcontext
+ * spill added by non-drreg routines.
+ */
+#define NON_DRREG_TLS_SPILL_OFFS_BASE 0xf00000
```
derekbruening, line 88: How is this not conflicting?  Are we assuming non-sentinel offsets are never this high?

ext/drreg/drreg.c
```diff
@@ -2047,43 +2063,78 @@ drreg_event_restore_state_with_ilist(void *drcontext, bool restore_memory,
...
     reg_id_t spill_slot_to_reg[MAX_SPILLS];
-    /* Note that we do not need to track multiple spill regs for aflags. On x86
+    /* Tracks which gpr contains the aflags app value.
+     * Note that we do not need to track multiple spill regs for aflags. On x86
      * aflags can be spilled only to xax using lahf. On AArchXX, while they can
      * be read into any register using mrs, they are always spilled to a slot.
      */
     reg_id_t aflags_spill_reg;
+    /* Tracks which gpr contains the mcontext base. This is required to identify
+     * mcontext spills/restores.
```
derekbruening, line 2095: For 32-bit thread-private, there is no base: an absolute address is used.

The mcontext is used for regular spills, instead of TLS, in some cases: so "mcontext spills" might be confusing, since instr_is_reg_spill_or_restore will identify those in such cases.  (Never happens for 64-bit x86.)

ext/drreg/drreg.c
derekbruening: Wait: drreg multi phase has multiple simulataneous spills of the same reg?  Is that if app2app holds a reg across an app instr, or has lazy restores (it doesn't IIRC), or insert phase instruments an app2app-added meta instr?

suite/tests/client-interface/drreg-test.dll.c
derekbruening: Best to put an issue number by any TODO comment.

core/ir/instr_api.h
abhinav92003: I tried to include some relevant examples below.

At the very least, we need to be aware of app instrs. Consider this case where we have multiple restores (which happens when we restore the reg before app read instrs):

example1:
```
<spill reg r1 to s1>
<restore reg r1 from s1>
<restore reg r1 from s1>
<crash>
```

Now, after the crash, do we restore r1 from s1 or not? If s1 still has r1’s app value, we can restore it (and live with a redundant restore in some cases), but if r1 was written by some app instruction before the crash, then s1 is stale and we should not restore. So we definitely need to look at app instrs. Whenever we have an app write, we forget all spill slots for that reg.

The next question is: can we avoid looking at meta instrs and the need to recognize non-drreg spills. In the current drreg restore logic, this is needed to be able to differentiate between tool val spills and app val spills. Without knowing whether a reg contains its native value currently, it’s hard to do that. We have used other heuristics in the past: like checking whether the reg was already spilled to some slot, and if so, assuming that the next spill is a tool val spill, but that has its limitations -- e.g. in some cases we may need to remember the second spill slot, like example2 below.

example2:
```
<app2app spills app r1 to s1>
<insertion phase spills app r1 to s2>
<app2app restores app r1 from s1>
<insertion phase writes some tool val to r1>
<app2app spills app r2 to s1>
// at this point app r1 is only present in s2
```



For aflags, this becomes a tougher problem. Because we don’t even keep the same slot at re-spills for aflags, or a slot at all (if spilled to xax). So we cannot even rely on each phase using the same slot always.

Coming back to the current issue, which is that gpr_native tracking is not perfect (as we may miss many patterns of spills/restores). I think we can use some heuristics like: if at a drreg spill, the reg is not native and and also not spilled to any drreg slot, then we assume that the gpr must be native right then. But I’m trying to think of a way to handle cases where we do have a spill slot at a spill but it is forgotten later.

example3:
```
<app2app spills app r1 to s1>
<app2app restores app r1 from s1>
<non-drreg r1 spill>
<non-drreg r1 restore> // now we wrongly think that r1 is !native
<insertion phase spills (app/tool?? Drreg won’t know) r1 to s2>
<insertion phase writes some tool val to r1>
<app2app spills app r2 to s1>
// at this point app r1 is only present in s2
```

derekbruening:
```
<spill reg r1 to s1>
<restore reg r1 from s1>
<restore reg r1 from s1>
<crash>
```

How does this sequence happen?  If the restores are before an app read, wouldn't they be accompanied by a tool spill and restore to a different slot?  So we can infer that the 1st refer is a lazy restore and r1 now has the app value?  And the 2nd restore can't happen w/o some intervening action?

abhinav92003: > If the restores are before an app read, wouldn't they either end a reservation or else be accompanied by a tool spill and restore to a different slot?

Yes, in this case, there is a tool val spill before the app read instr and after the app read instr a tool val restore. I omitted that here for brevity as the main question is still valid ("should we restore r1 from s1 after crash").

> So we can infer that the 1st refer is a lazy restore and r1 has the app value afterward?

Yes, the current logic does understand that r1 has the app value after a restore from s1. In this case, after the app read is complete, r1's tool value is restored, so we do need to restore r1's app value at crash.

> 2nd restore seems like it couldn't happen since r1 is not reserved?

r1 is still reserved. It was restored once only for the duration of the app read instr.


For reference: following is the instrumented sequence for `test_asm_fault_restore_gpr_restored_for_read`:

```
TAG  0x00007fb23802a183
 +0    L3 @0x00007faff81015c8  48 c7 c2 16 f1 00 00 mov    $0x000000000000f116 -> %rdx
 +7    L3 @0x00007faff80fdeb8  48 c7 c2 16 f1 00 00 mov    $0x000000000000f116 -> %rdx
 +14   m4 @0x00007faff80fffe8                       <label>
 +14   m4 @0x00007faff8176330  65 48 89 14 25 e0 00 mov    %rdx -> %gs:0x000000e0[8byte] // app spill
                               00 00
 +23   m4 @0x00007faff80e2768  48 c7 c2 cd ab 00 00 mov    $0x000000000000abcd -> %rdx // tool write
 +30   L3 @0x00007faff80e26e8  90                   nop
 +31   m4 @0x00007faff80dd8b8  65 48 89 14 25 e8 00 mov    %rdx -> %gs:0x000000e8[8byte] // tool spill
                               00 00
 +40   m4 @0x00007faff80ffee8  65 48 8b 14 25 e0 00 mov    %gs:0x000000e0[8byte] -> %rdx // app restore
                               00 00
 +49   L3 @0x00007faff80ff158  48 01 d3             add    %rdx %rbx -> %rbx
 +52   m4 @0x00007faff8107590  65 48 8b 14 25 e8 00 mov    %gs:0x000000e8[8byte] -> %rdx // tool restore
                               00 00
 +61   L3 @0x00007faff816bec8  48 c7 c1 00 00 00 00 mov    $0x0000000000000000 -> %rcx
 +68   L3 @0x00007faff8174990  48 8b 09             mov    (%rcx)[8byte] -> %rcx
 +71   m4 @0x00007faff8103360  65 48 8b 14 25 e0 00 mov    %gs:0x000000e0[8byte] -> %rdx // app restore
                               00 00
 +80   L4 @0x00007faff8108738  e9 7c 1e f3 3f       jmp    $0x00007fb23802a1a1
END 0x00007fb23802a183
```

abhinav92003: One of the complexities is that: we cannot know for sure which restore for a reg ends the reservation. Besides the final restore, there could be restores for app read (like the one above, which could arguably be discovered based on the extra tool spill and restore around it, but not sure if hard-coding that pattern is okay), and restores by the user to get the app value.

derekbruening: > I omitted that here for brevity

I would claim that makes the question invalid: we would never see the sequence listed.  We would have the full info.

> be discovered based on the extra tool spill and restore around it, but not sure if hard-coding that pattern is okay

Since that is *our own pattern that we control*, then yes it seems ok IMHO.

> restores by the user to get the app value.

Ah this is pointing at something problematic.  So here is the example where we can't tell whether a restore ends the reservation: the user requesting a restore.  If this is the problem, maybe we could use a label here to help us identify it.

abhinav92003: > Since that is our own pattern that we control, then yes it seems ok IMHO.

Such patterns may be harder to detect if there are interleaving instrumentation between them added e.g. by the instru2instru phase.

Maybe we can use a combination of hardcoding patterns and adding labels to disambiguate restores and figure out where the gpr reservation ends, and then use the old logic of using the spill slot Id to differentiate between app and tool spills. I still need to think through some cases to convince myself. 

Aflags is a different story though. The most important difference being that it doesn't guarantee that re-spills will use the same slot, or any slot at all (if spilled to xax). So it's harder to differentiate between app aflags spill and tool aflags spill.

In my mind, my current approach of tracking the native value seemed better as other drreg logic can be completely decoupled from the restore state event.

ext/drreg/drreg.c
```diff
@@ -2047,43 +2063,78 @@ drreg_event_restore_state_with_ilist(void *drcontext, bool restore_memory,
      * different slot) after app instr write, multiple reg restores due to app
      * instr read, aflags spilled to reg instead of slot (for x86 only), lazy
      * restores, multi-phase use of drreg leading to nested or overlapping
```
abhinav92003, line 2065: Starting an independent thread for this: How are multiple restore state events handled? If multiple clients and core DR had altered some machine state (e.g. spilled some gpr to some location that they own), then all of them will try to restore it in their respective restore state events. Does the order in which we invoke the restore state events guarantee that the gpr's final restored value will be the app value?

derekbruening: DR always goes first and restores state that was changed in its mangling or instrumentation it owns like clean calls.  Clients/extensions are in order of the callback registration or drmgr priority.

core/ir/instr_api.h
abhinav92003: Update: I'm trying out an approach where, instead of walking the faulting fragment ilist from the start till the faulting instr, I walk it back from the end to the faulting instr. Finding the matching spill for a restore is turning out much simpler than finding the matching restore to a spill. This approach seems to work fine on the tests I have currently. Will send it over for review in a new PR when ready.

derekbruening: Does the reverse walk make it easier to distinguish tool-value spills and user-requested restores?  I guess the final restore in the block always ends the reservation so that is what makes it easier?  Or is that always true: thinking of cases where a reg is live but then an app write makes it dead later, but there would still be a restore before the app read that made it live and that would be after all other restores.

abhinav92003: > Does the reverse walk make it easier to distinguish tool-value spills 

We want to see only app value restores and spills in this approach, and "jump" over tool value spills/restores: starting with the last restore[1] (which may be the end of block app val restore or one for an app read), we jump to its matching previous spill which is also guaranteed to be an app val spill. This goes on: any restore before this must also be an app value restore (or an app val write[2]), and so on. We backtrack until we find an app val spill-restore pair that crosses the faulting instr and use that slot to restore.

> user-requested restores

I see that `drreg_get_app_value` restores may be stateful if the reg is not in-use by drreg. So we don't want to ignore user restores every time. In any case, I think my current approach won't be affected by non-stateful restores: it doesn't care why a restore was added, just that it provided the app value for an app read (or end of block).

> I guess the final restore in the block always ends the reservation so that is what makes it easier?

Yes, but
[1]: I'm still debating whether I should start the reverse walk at the final restore in the block or the restore before the first app read after the faulting instr (if any). The latter feels better, but I'm unable to come up with a proof test case.

> thinking of cases where a reg is live but then an app write makes it dead later, but there would still be a restore before the app read that made it live and that would be after all other restores.

Right, there must be some app read, before the app write that makes it dead; or no app write at all in which case the reg is live beyond the block.


[2]: if the first app instr that accesses a reg after a fault is a write to that reg, then the reg is dead and doesn't need restoration. This shouldn't need special handling as we don't expect there to be any spill-restore pairs across an app write.

abhinav92003: > we don't expect there to be any spill-restore pairs across an app write.

Hmm.. this happens in the app2app scatter/gather expansion. We reserve aflags at the top and unreserve it at end, and there are `test` instrs (that write aflags) in the middle marked as app. On a separate note: it's weird the the corresponding `jz`s are not marked as app too; why the former but not the latter is marked as app?

I think this is an example that motivates starting the walk from the end of ilist, instead of first app read, so that we don't miss restore instrs. There may be unconventional uses in app2app phase that don't restore before app read or re-spill after app write.

abhinav92003: > There may be unconventional uses in app2app phase that don't restore before app read or re-spill after app write.

But then if some client makes such unconventional use, I think it's their responsibility to restore state on fault, right? And drx does that for the scatter/gather expansion.

derekbruening: > On a separate note: it's weird the the corresponding jzs are not marked as app too; why the former but not the latter is marked as app?

A branch marked as app must end the block.  (DR then mangles it to have an exit stub and link it to other blocks.)

abhinav92003:
Dropping this PR in favour of #5000 that implements a superior approach.

PR #4988, opened 2021-07-01
i#2626: AArch64 v8.0 decode: Add SHA- instructions to codec

MDevereau:
Adds the following instructions to the codec:
- sha1c
- sha1h
- sha1m
- sha1p
- sha1su0
- sha1su1
- sha256h
- sha256h2
- sha256su0
- sha256su1

Issue: #2626

### DIFF
--- a/core/ir/aarch64/codec.txt
+++ b/core/ir/aarch64/codec.txt
@@ -1433,3 +1433,14 @@ x001111001000010xxxxxxxxxxxxxxxx     scvtf     d0 : wx5 scale
 0x001110xx100000001010xxxxxxxxxx     saddlp    dq0 : dq5 bhsd_sz
 0000111100xxxxxx101001xxxxxxxxxx     sshll      d0 : d5 immhb
 0100111100xxxxxx101001xxxxxxxxxx     sshll2     q0 : q5 immhb
+
+01011110000xxxxx000000xxxxxxxxxx     sha1c     q0 : s5 d16
+0101111000101000000010xxxxxxxxxx     sha1h     s0 : s5
+01011110000xxxxx001000xxxxxxxxxx     sha1m     q0 : s5 d16
+01011110000xxxxx000100xxxxxxxxxx     sha1p     q0 : s5 d16
+01011110000xxxxx001100xxxxxxxxxx     sha1su0   d0 : d5 d16
+0101111000101000000110xxxxxxxxxx     sha1su1   d0 : d5
+01011110000xxxxx010000xxxxxxxxxx     sha256h   q0 : q5 d16
+01011110000xxxxx010100xxxxxxxxxx     sha256h2  q0 : q5 d16
+0101111000101000001010xxxxxxxxxx     sha256su0 d0 : d5
+01011110000xxxxx011000xxxxxxxxxx     sha256su1 d0 : d5 d16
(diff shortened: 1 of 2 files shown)


MDevereau:
run arm tests

AssadHashmi (approved):

core/ir/aarch64/codec.txt
```diff
@@ -1433,3 +1433,14 @@ x001111001000010xxxxxxxxxxxxxxxx     scvtf     d0 : wx5 scale
...
+01011110000xxxxx000000xxxxxxxxxx     sha1c     q0 : s5 d16
+0101111000101000000010xxxxxxxxxx     sha1h     s0 : s5
+01011110000xxxxx001000xxxxxxxxxx     sha1m     q0 : s5 d16
+01011110000xxxxx000100xxxxxxxxxx     sha1p     q0 : s5 d16
+01011110000xxxxx001100xxxxxxxxxx     sha1su0   d0 : d5 d16
+0101111000101000000110xxxxxxxxxx     sha1su1   d0 : d5
+01011110000xxxxx010000xxxxxxxxxx     sha256h   q0 : q5 d16
+01011110000xxxxx010100xxxxxxxxxx     sha256h2  q0 : q5 d16
+0101111000101000001010xxxxxxxxxx     sha256su0 d0 : d5
+01011110000xxxxx011000xxxxxxxxxx     sha256su1 d0 : d5 d16
```
davidben, line 1446: Are you sure these are correct? I may be misunderstanding this system (trying to figure out how to add SHA512 instructions), but it seems every `d` register should have been a `q` register. Checking just the first few:

[SHA1C](https://developer.arm.com/documentation/ddi0602/2021-12/SIMD-FP-Instructions/SHA1C--SHA1-hash-update--choose--)
The PR decodes Rd as `q`, Rn as `s`, and Rm as `d`. However, Rm is described as `<Vm>.4S`, not `<Vm>.2S`, which totals 128-bit, not 64-bit.

[SHA1H](https://developer.arm.com/documentation/ddi0602/2021-12/SIMD-FP-Instructions/SHA1H--SHA1-fixed-rotate-)
This looks correct. The PR decodes Rd and Rn as `s`, which matches the documentation. Indeed, the pseudocode only reads 32 bits.

[SHA1M](https://developer.arm.com/documentation/ddi0602/2021-12/SIMD-FP-Instructions/SHA1M--SHA1-hash-update--majority--)
The PR decodes Rd as `q`, Rn as `s`, and Rm as `d`. However, Rm is described as `<Vm>.4S`, not `<Vm>.2S`, which totals 128-bit, not 64-bit.

[SHA1P](https://developer.arm.com/documentation/ddi0602/2021-12/SIMD-FP-Instructions/SHA1P--SHA1-hash-update--parity--)
The PR decodes Rd as `q`, Rn as `s`, and Rm as `d`. However, Rm is described as `<Vm>.4S`, not `<Vm>.2S`, which totals 128-bit, not 64-bit.

If this is right, the other instructions probably also need a re-check. Is it possible all the `V` registers were mistranscribed in this and other PRs?

joshua-warburton: Your logic is correct. However the other thing that is missing here is that there is no operand to describe the element size, which in this case is always an `s` (`0x02`)

davidben: I guess, with only one size, whether the element size is part of the instruction, or just implicit in the operation, is probably a matter of interpretation. :-) I figured it was the latter, since it wasn't included here.

PR #4989, opened 2021-07-01
Fixing support for inject_x64 injection inside WOW64 processes

N0fix:
PR for [issue#4990](https://github.com/DynamoRIO/dynamorio/issues/4990).

Those changes fixes support for inject_x64 injection on WoW64 processes, which allows support for mixed mode code, see [this](https://groups.google.com/g/dynamorio-users/c/rhEpslerwf8)

Note that allocation of vmheap fails upon initializing dynamoRIO 64 on WoW64 processes. Thus, we need to pass `-reachable_heap` to avoid having to make this allocation. 

We still need to have a proper support on `drrun64` to inject natively without having to use `create_process.exe`.

Example command line that works : 
```
bin64\drrun.exe -reachable_heap -inject_x64 -c .\clientdll.dll -- bin64\create_process.exe .\helloworld32.exe
```
As we need to specify `reachable_heap` I am afraid that `win32.mixedmode` test will be needing some tweaks.

Changes features : 

- Saving `eax` register that holds routine address for `RtlUserThreadStart` before mode switch, and restore it on mode switch
- Fixing far jmp to switch to x64 mode on injection
- Fixing env variable argument propagation 

## EDIT

`-reachable_heap` should not be required anymore since vmheap injection issues has been fixed in [this](https://github.com/DynamoRIO/dynamorio/pull/4989/commits/898b8c25b36f4ae55e5305d2fc16776b6cbb8218) commit.


core/win32/inject.c
```diff
@@ -1083,20 +1082,25 @@ generate_switch_mode_jmp_to_hook(HANDLE phandle, byte *local_code_buf,
     instr_t *restore_esp =
         INSTR_CREATE_mov_ld(GDC, opnd_create_reg(REG_ESP),
                             OPND_CREATE_MEM32(REG_NULL, (int)(size_t)mode_switch_data));
-
+	byte* eaxpos =(byte*) ((size_t)mode_switch_data) + 8;
```
derekbruening, line 1085: This does not satisfy the style conventions.  Best to use clang-format locally.  The clang-format GA CI run will fail unless this matches our style.

derekbruening: Why +8?  Maybe should have a named offset or even better define a struct.

core/win32/inject.c
```diff
@@ -1083,20 +1082,25 @@ generate_switch_mode_jmp_to_hook(HANDLE phandle, byte *local_code_buf,
...
+	// Restoring eax back, which is storing routine address that needs to be passed to RtlUserStartThread
```
derekbruening, line 1086: Where is eax clobbered, such that we need to restore it?  Not sure I understand this.

core/win32/inject.c
```diff
@@ -1185,7 +1189,10 @@ inject_gencode_mapped_helper(HANDLE phandle, char *dynamo_path, uint64 hook_loca
     byte *mode_switch_buf = NULL;
     byte *mode_switch_data = NULL;
     size_t switch_code_sz = PAGE_SIZE;
-    static const size_t switch_data_sz = SWITCH_MODE_DATA_SIZE;
+    size_t switch_data_sz = SWITCH_MODE_DATA_SIZE;
+    if (x86_code && DYNAMO_OPTION(inject_x64)) {
+	switch_data_sz *= 3; // we need space for saving ESP and EAX
```
derekbruening, line 1194: Not following why 3x?  Just +4?

core/win32/syscall.c
```diff
@@ -2961,7 +2964,12 @@ pre_system_call(dcontext_t *dcontext)
     reg_t *param_base = pre_system_call_param_base(mc);
     dr_where_am_i_t old_whereami = dcontext->whereami;
     dcontext->whereami = DR_WHERE_SYSCALL_HANDLER;
-    IF_X64(ASSERT_TRUNCATE(sysnum, int, mc->xax));
+    // FIXME
+    // commenting this ASSERT since x64 client injected in
+    // WOW64 binaries will be using mc->rax as comparaison
+    // thus making the assert fail, while the syscall still
+    // being valid.
```
derekbruening, line 2971: A) Who is setting the upper bits?

B) You mean the Windows kernel ignores the upper bits?  So the assert is flawed?

core/win32/syscall.c
```diff
@@ -1744,7 +1744,10 @@ propagate_options_via_env_vars(dcontext_t *dcontext, HANDLE process_handle,
      * model without requiring setting up config files for children.
      */
     uint64 peb;
-    bool peb_is_32 = is_32bit_process(process_handle);
+    bool peb_is_32 = is_32bit_process(process_handle)
+	    // if x64 client targeting WOW64 we need to
```
derekbruening, line 1748: nit: Modern style is to prefer capitalized and punctuated whole-sentence comments.

derekbruening (commented):
Thank you for working on this and contributing the patch.  I'm not sure I understand some of it though.

core/win32/inject.c
```diff
@@ -1065,7 +1065,6 @@ inject_gencode_at_ldr(HANDLE phandle, char *dynamo_path, uint inject_location,
  * process, mode_switch_buf_sz is maximum size for switch code, and
```
derekbruening, line 1065: > Note that allocation of vmheap fails upon initializing dynamoRIO 64 on WoW64 processes. Thus, we need to pass -reachable_heap to avoid having to make this allocation.

This should be solved by changing the default `vmheap_size` to be much smaller than 8GB for x64 DR inside WOW64.

N0fix: This is done in my last push.

core/win32/syscall.c
N0fix: I meant that the assert always fails due to the size of mc->rax being compared with wrong sizes, no matter if the syscall is valid or not.

core/win32/inject.c
N0fix: Fixed with +4 and proper name on last commit.

core/win32/inject.c
N0fix: I have no idea how/why/where EAX is getting clobbered in the process yet, but it is.

core/win32/inject.c
N0fix: +4 is fine. Defining a struct might be a bit overkill. I made variable in last commit.

core/win32/inject.c
derekbruening: Best to document that fact in a comment so future readers know what the status is.

core/win32/syscall.c
derekbruening: I don't understand "the assert always fails due to the size of mc->rax being compared with wrong sizes"?

core/heap.c
```diff
@@ -1901,8 +1901,13 @@ vmm_heap_init()
     if (DYNAMO_OPTION(vm_reserve)) {
         vmm_heap_unit_init(&heapmgt->vmcode, DYNAMO_OPTION(vm_size), true, "vmcode");
         if (!REACHABLE_HEAP()) {
-            vmm_heap_unit_init(&heapmgt->vmheap, DYNAMO_OPTION(vmheap_size), false,
-                               "vmheap");
+            if (DYNAMO_OPTION(inject_x64)){
```
derekbruening, line 1904: This should also test for this being a WOW64 process.  This option would be set in 64-bit parents too so by itself it does not tell you it's WOW64.

core/optionsx.h
```diff
@@ -1595,7 +1595,12 @@ OPTION_DEFAULT(uint_size, vmheap_size, IF_X64_ELSE(8192ULL, 128) * 1024 * 1024,
                 * for which we need more than 256MB.
                 */
                "capacity of virtual memory region reserved for unreachable heap")
-
+OPTION_DEFAULT(uint_size, vmheap_size_wow64, 128 * 1024 * 1024,
```
derekbruening, line 1598: Probably best inside #ifdef WINDOWS

core/win32/inject.c
```diff
@@ -1082,11 +1082,12 @@ generate_switch_mode_jmp_to_hook(HANDLE phandle, byte *local_code_buf,
...
         INSTR_CREATE_mov_ld(GDC, opnd_create_reg(REG_ESP),
                             OPND_CREATE_MEM32(REG_NULL, (int)(size_t)mode_switch_data));
-	byte* eaxpos =(byte*) ((size_t)mode_switch_data) + 8;
-	// Restoring eax back, which is storing routine address that needs to be passed to RtlUserStartThread
-	instr_t *restore_eax = INSTR_CREATE_mov_ld(GDC, opnd_create_reg(REG_EAX), 
-		OPND_CREATE_MEM32(REG_NULL, (int)(size_t)eaxpos));
-	
+    const byte *eax_saved_offset = (byte *)((size_t)mode_switch_data) + 4;
+    // Restoring eax back, which is storing routine address that needs to be passed to
+    // RtlUserStartThread
```
derekbruening, line 1087: From other comment thread: would be good to explain that we do not know why eax is clobbered in a `XXX` comment since this may indicate some bug or problem somewhere else.

derekbruening:
All the tests are failing: looks like b/c of tabs:
```
ERROR: diff contains tabs:
  +	 // if x64 client targeting WOW64 we need to
```

The style convention is to not have any tabs.

core/heap.c
N0fix: No idea about how to do that without messing with function's arguments, so I left a message here for now.

core/optionsx.h
N0fix: Done

core/win32/inject.c
N0fix: Done

core/win32/syscall.c
N0fix: Trying to print erroneous syscalls (syscalls that makes the assert trigger) on WoW64 + inject_x64, I can see that `mc->xax` value is `ffffffff00031054`. I assume that this comes from a conversion somehow representing the syscall with a signed number, which outputs a value that is wrong if you display it. Since `mc->xax` is only used as a sysnumber and type converted to `int`, this is not an issue, but this still makes `ASSERT_TRUNCATE` (more precisely `CHECK_TRUNCATE_TYPE_int`) fail.

derekbruening: There would not be a widening conversion for an application register.  App register values have to be precisely preserved.  The machine context structure stores 64-bit values and prints them as such.  If the top bits there are 1's, either the app set them to 1's, or somehow there is a bug where DR corrupted the register value.  Is this for 32-bit app code?  If it is 64-bit, we can't put it past Windows making some change to have the top bits mean something -- though in the past 64-bit code wrote to `eax`, zeroing the top bits.

core/heap.c
derekbruening: Use `is_wow64_process(NT_CURRENT_PROCESS)`

derekbruening:
Tests are all still failing with clang-format style violations:

```
 ERROR: diff contains trailing spaces:
  + // XXX : We should also be testing for target to be wow64 along with

-- Changes are not formatted properly:
```

N0fix:
> Tests are all still failing with clang-format style violations:
> 
> ```
>  ERROR: diff contains trailing spaces:
>   + // XXX : We should also be testing for target to be wow64 along with
> 
> -- Changes are not formatted properly:
> ```

Fixed in last commit

derekbruening:
There are more -- I would suggest checking the entire diff, or better following the development setup `make/git/devsetup.sh` https://dynamorio.org/page_workflow.html#autotoc_md296 which checks for trailing spaces and other things at local commit time.

```
  ERROR: diff contains trailing spaces:
  + // If x64 client targeting WOW64 we need to
```

There are also clang-format failures -- see the ci-clang-format test.  I would suggest setting up clang-format in your editor.

N0fix:
> There are more -- I would suggest checking the entire diff, or better following the development setup `make/git/devsetup.sh` https://dynamorio.org/page_workflow.html#autotoc_md296 which checks for trailing spaces and other things at local commit time.
> 
> ```
>   ERROR: diff contains trailing spaces:
>   + // If x64 client targeting WOW64 we need to
> ```
> 
> There are also clang-format failures -- see the ci-clang-format test. I would suggest setting up clang-format in your editor.

Just removed other trailing spaces.
I was looking for your dev setup, thanks for the links, I will take a look into it if this merge fails.

core/heap.c
N0fix: Fixed

core/heap.c
```diff
@@ -1901,12 +1901,20 @@ vmm_heap_init()
     if (DYNAMO_OPTION(vm_reserve)) {
         vmm_heap_unit_init(&heapmgt->vmcode, DYNAMO_OPTION(vm_size), true, "vmcode");
         if (!REACHABLE_HEAP()) {
-            vmm_heap_unit_init(&heapmgt->vmheap, DYNAMO_OPTION(vmheap_size), false,
-                               "vmheap");
+            vmm_heap_unit_init(
+                &heapmgt->vmheap,
+#ifdef WINDOWS
+                (DYNAMO_OPTION(inject_x64) && is_wow64_process(NT_CURRENT_PROCESS))
```
derekbruening, line 1907: Probably safer to remove the inject_x64 check and have IF_X64 instead since we don't care about injection options, just whether we're a 64-bit DR in a WOW64 process.

IF_WINDOWS_ELSE and IF_X64_ELSE could be used to avoid whole-line ifdefs.

core/win32/syscall.c
```diff
@@ -2961,7 +2964,13 @@ pre_system_call(dcontext_t *dcontext)
...
     dr_where_am_i_t old_whereami = dcontext->whereami;
     dcontext->whereami = DR_WHERE_SYSCALL_HANDLER;
-    IF_X64(ASSERT_TRUNCATE(sysnum, int, mc->xax));
+    /* FIXME
+     * Commenting this ASSERT since x64 client injected in
+     * WOW64 binaries will be using mc->rax as comparaison
+     * thus making the assert fail, no matter if the
+     * syscall is correct or not.
+     */
+    // IF_X64(ASSERT_TRUNCATE(sysnum, int, mc->xax));
```
derekbruening, line 2973: The style guide disallows commented-out code (https://dynamorio.org/page_code_style.html#autotoc_md125).  It is also best to have an issue to reference with any FIXME or XXX comment.  If we don't understand this yet I think we want a ref to #49 and to disable based on just this situation.

```C
/* XXX i#49: mc->rax's top bits are non-zero in 32-bit mode for reasons we do not yet understand.
 * For now we disable the assert for mixed-mode.
 */
IF_X64(ASSERT(is_wow64_process(NT_CURRENT_PROCESS) || CHECK_TRUNCATE_TYPE_int(mc->xax)))    
```

derekbruening (approved):

N0fix:
Not sure about what is failing in Windows build. Looks like a CI error that is not coming from my code.

derekbruening:
> Not sure about what is failing in Windows build. Looks like a CI error that is not coming from my code.

I think it's a sourceforge server error causing doxygen to not get downloaded.  Probably re-running is the first thing to try: I clicked re-run.

derekbruening:
run arm tests

derekbruening:
@AssadHashmi -- is there any way to enable existing contributors to trigger AArch64 Jenkins runs from a forked PR?  It seems there's no way to do that now for a PR from a forked PR created by a new contributor.

derekbruening:
Going to merge this w/o the Jenkins for now: the master run will catch issues.  Cross-compile for these code changes gives high confidence.  Once @AssadHashmi is back we'll try to figure out a better long term solution.

AssadHashmi:
add to whitelist

AssadHashmi:
ok to test

AssadHashmi:
test this please

derekbruening:
run arm tests

derekbruening:
I think once it's merged nothing will run

AssadHashmi:
> I think once it's merged nothing will run

You're right @derekbruening.
`add to whitelist` worked for a new contributor @mat-kalinowski on his first PR https://github.com/DynamoRIO/dynamorio/pull/5020#issuecomment-883589296

N0fix:
An `encoding cti error` is occurring inside `generate_switch_mode_jmp_to_hook` on latest builds (from source or cronbuild, doesn't matter). I have been looking at commits since this PR, and I did not find something that would have interfere with this PR at first glance.

N0fix:
I can confirm that mixed code injection is not working anymore on latest cronbuilds (18324 and 18331). Well, it doesn't work since cronbuild release. I will check if this comes from my side, maybe I missed something. 
The line I mentioned above is *not* modified in my PR source code, and my PR source code have successful injection to x86 code.

derekbruening:
A regression test should be enabled to avoid breakage.

N0fix:
> A regression test should be enabled to avoid breakage.

Applying modification from this PR to #803 works perfectly well, if we add :
```C
			if (is_32bit_process(phandle) && DYNAMO_OPTION(inject_x64)) {
				target_64 = false;
			}
```
that I did forget in my PR.

The syscall trouble must be coming from some other commit since this PR. Would you know which commit might have mess with syscall handling since then?

N0fix:
Trouble starts from cronbuild-8.0.18663, we need to check commits between 18655 and 18663.

PR #5002, opened 2021-07-09
i#4848: AArch64 v8.2 Decode: FCVTMU FCVTNU SYSL CMLT

MDevereau:
Adds FCVTMU FCVTNU SYSL and CMLT to the codec.

Issue: #4848, #2626


MDevereau:
run arm tests

MDevereau:
@derekbruening Could you please review this patch as Assad is away? Only patches that include changes to codec.c block the addition of further instructions, so I can keep the patch frequency to a minimum and work on further instructions locally and get Assad to do the large volume of non codec.c changes on return.

derekbruening (approved):
I spot-checked one encoding.  Hopefully there's a mechanism to verify vs an external reference or other decoder.

PR #5003, opened 2021-07-09
i4489 drcachesim: support any associativity

algrant-arm:
drcachesim used to support only power-of-2 associativity even though it's not uncommon for L1 caches to be 3-way, 5-way, or sometimes 24-way for larger caches. Most changes were in option validation and related messages. The use of assoc_bits_ was removed. Shifts and masks were replaced by multiplication (as cheap as shifting, on modern CPUs) and a test. There are no division or modulo operators on the speed path.

This also seemed like a good opportunity to change the misnamed blocks_per_set_ variable to blocks_per_way_ (blocks per set is the same as associativity).

Reworked pull request to fix clang-format issues.

### DIFF
--- a/clients/drcachesim/simulator/cache_fifo.cpp
+++ b/clients/drcachesim/simulator/cache_fifo.cpp
@@ -56,8 +56,8 @@ cache_fifo_t::init(int associativity, int block_size, int total_size,
 
     // Create a replacement pointer for each set, and
     // initialize it to point to the first block.
-    for (int i = 0; i < blocks_per_set_; i++) {
-        get_caching_device_block(i << assoc_bits_, 0).counter_ = 1;
+    for (int i = 0; i < blocks_per_way_; i++) {
+        get_caching_device_block_scaled(i, 0).counter_ = 1;
     }
     return true;
 }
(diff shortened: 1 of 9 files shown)


clients/drcachesim/simulator/cache_fifo.cpp
```diff
@@ -56,8 +56,8 @@ cache_fifo_t::init(int associativity, int block_size, int total_size,
 
     // Create a replacement pointer for each set, and
     // initialize it to point to the first block.
-    for (int i = 0; i < blocks_per_set_; i++) {
-        get_caching_device_block(i << assoc_bits_, 0).counter_ = 1;
+    for (int i = 0; i < blocks_per_way_; i++) {
+        get_caching_device_block_scaled(i, 0).counter_ = 1;
```
abhinav92003, line 60: I'm thinking whether we really need a helper method here; maybe just inline the 1-line implementation of get_caching_device_block_scaled?

clients/drcachesim/simulator/cache_simulator.cpp
```diff
@@ -107,9 +107,9 @@ cache_simulator_t::cache_simulator_t(const cache_simulator_knobs_t &knobs)
     if (!llc->init(knobs_.LL_assoc, (int)knobs_.line_size, (int)knobs_.LL_size, NULL,
                    new cache_stats_t(knobs_.LL_miss_file, warmup_enabled_))) {
         error_string_ =
-            "Usage error: failed to initialize LL cache.  Ensure sizes and "
-            "associativity are powers of 2, that the total size is a multiple "
-            "of the line size, and that any miss file path is writable.";
+            "Usage error: failed to initialize LL cache.  Ensure line size is "
+            "a multiple of 2, that the total size is a multiple of line size "
+            "and associativity, and that any miss file path is writable.";
```
abhinav92003, line 112: Also `total_size / (associativity*line_size)` needs to be a power of 2.

Issue #3068, opened 2018-06-26
drcachesim: add structured simulation results


derekbruening:
This is a feature request to add an interface for retrieving simulation
results as structured data when using drcachesim as a library, to avoid
having to print out statistics and then parse them back in.

This will be immediately useful in the drcachesim_unit_tests.

derekbruening:
Xref #4842 on drcachesim unit tests

derekbruening:
There is discussion here on the best way to export these statistics: https://groups.google.com/g/dynamorio-users/c/sPidMM2KA1E/m/R3gxzEpCAAAJ

derekbruening:
We now have structured results for drcachesim's cache statistics and for basic_counts.  Leaving this open for adding structured results to our other analyzers like reuse_distance.

Issue #4343, opened 2020-06-30
[cleanup] use namespaces in C++ code, in particular drcachesim


derekbruening:
This is proposing adopting general principles of using namespaces in our C++ code (particularly drcachesim, our biggest C++ project) for better name isolation and organization.

derekbruening:
We have settled on `dynamorio::drmemtrace` which is in place in the tracer code; it still needs to be added to the analyzer code.

derekbruening:
The plan is to go ahead and break source compatibility and change the existing headers, prior to the 10.0.0 release.

PR #5020, opened 2021-07-20
i#3068: API for accessing cache model metrics

mat-kalinowski:
This patch contains implementation of the API for accessing separate
metrics gathered during the simulation.

Functions for accessing metrics were added for all caches in the
default configuration: L1I, L1D and LL. Also additional functions
for getting configuration options of the cache_simulator_t.

Issue: #3068

### DIFF
--- a/clients/drcachesim/simulator/cache_simulator.cpp
+++ b/clients/drcachesim/simulator/cache_simulator.cpp
@@ -607,7 +607,7 @@ cache_simulator_t::get_l1_dcache_name(unsigned int core) const
 }
 
 const caching_device_stats_t *
-cache_simulator_t::get_cache_stats(const std::string& cache_type) const
+cache_simulator_t::get_cache_stats(const std::string &cache_type) const
 {
     auto pos = all_caches_.find(cache_type);
 
--- a/clients/drcachesim/simulator/cache_simulator.cpp
+++ b/clients/drcachesim/simulator/cache_simulator.cpp
@@ -595,17 +595,16 @@ cache_simulator_t::print_results()
 }
 
 int_least64_t
-cache_simulator_t::get_cache_metric(unsigned level, metric_name_t metric, unsigned core,
+cache_simulator_t::get_cache_metric(metric_name_t metric, unsigned level, unsigned core,
                                     cache_split_t split) const
 {
     caching_device_t *curr_cache;
 
     if (core >= knobs_.num_cores) {
-        ERRMSG("Wrong core number: %u.\n", core);
-        return 0;
+        return STATS_ERROR_WRONG_CORE_NUMBER;
     }
 
-    if (split == DATA) {
+    if (split == cache_split_t::DATA) {
         curr_cache = l1_dcaches_[core];
     } else {
         curr_cache = l1_icaches_[core];
@@ -615,16 +614,14 @@ cache_simulator_t::get_cache_metric(unsigned level, metric_name_t metric, unsign
         caching_device_t *parent = curr_cache->get_parent();
 
         if (parent == NULL) {
-            ERRMSG("Wrong cache level: %u.\n", core);
-            return 0;
+            return STATS_ERROR_WRONG_CACHE_LEVEL;
         }
         curr_cache = parent;
     }
     caching_device_stats_t *stats = curr_cache->get_stats();
 
     if (stats == NULL) {
-        ERRMSG("Cache doesn't support statistics counting.");
-        return 0;
+        return STATS_ERROR_NO_CACHE_STATS;
     }
 
     return stats->get_metric(metric);
--- a/clients/drcachesim/simulator/cache_simulator.cpp
+++ b/clients/drcachesim/simulator/cache_simulator.cpp
@@ -594,6 +594,91 @@ cache_simulator_t::print_results()
     return true;
 }
 
+inline std::string
+cache_simulator_t::get_l1_icache_name(unsigned int core) const {
+    return "L1_I_Cache_" + std::to_string(core);
+}
+
+inline std::string
+cache_simulator_t::get_l1_dcache_name(unsigned int core)  const {
+    return "L1_D_Cache_" + std::to_string(core);
+}
+
+const caching_device_stats_t *
+cache_simulator_t::get_cache_stats(const std::string& cache_type) const {
+    auto pos = all_caches_.find(cache_type);
+
+    if (pos != all_caches_.end()) {
+        return pos->second->get_stats();
+    } else {
+        ERRMSG("Couldn't get statistics. Wrong cache type.\n");
+        return nullptr;
+    }
+}
+
+int_least64_t
+cache_simulator_t::get_l1i_metric(metric_name_t metric, unsigned int core) const
+{
+    if (core >= knobs_.num_cores) {
+        ERRMSG("Wrong core number: %u.\n", core);
+        return 0;
+    }
+
+    const auto *stats = get_cache_stats(get_l1_icache_name(core));
+
+    if (!stats) {
+        return 0;
+    }
+
+    return stats->get_metric(metric);
+}
+
+int_least64_t
+cache_simulator_t::get_l1d_metric(metric_name_t metric, unsigned int core) const
+{
+    if (core >= knobs_.num_cores) {
+        ERRMSG("Wrong core number: %u.\n", core);
+        return 0;
+    }
+
+    const auto *stats = get_cache_stats(get_l1_dcache_name(core));
+
+    if (!stats) {
+        return 0;
+    }
+
+    return stats->get_metric(metric);
+}
+
+int_least64_t
+cache_simulator_t::get_ll_metric(metric_name_t metric) const
+{
+    const auto *stats = get_cache_stats(ll_cache_name);
+
+    if (!stats) {
+        return 0;
+    }
+
+    return stats->get_metric(metric);
+}
+
+unsigned int
+cache_simulator_t::get_core_num() const {
+    return knobs_.num_cores;
+}
+
+bool
+cache_simulator_t::is_prefetcher_used() const
+{
+    return knobs_.data_prefetcher != PREFETCH_POLICY_NONE;
+}
+
+bool
+cache_simulator_t::is_cache_coherent() const
+{
+    return knobs_.model_coherence;
+}
+
 cache_t *
 cache_simulator_t::create_cache(const std::string &policy)
 {
--- a/clients/drcachesim/simulator/cache_simulator.cpp
+++ b/clients/drcachesim/simulator/cache_simulator.cpp
@@ -595,17 +595,20 @@ cache_simulator_t::print_results()
 }
 
 inline std::string
-cache_simulator_t::get_l1_icache_name(unsigned int core) const {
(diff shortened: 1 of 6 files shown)


clients/drcachesim/simulator/cache_simulator.h
```diff
@@ -60,12 +60,26 @@ class cache_simulator_t : public simulator_t {
     bool
     print_results() override;
 
+    int_least64_t
+    get_l1i_metric(metric_name_t metric, unsigned int core) const;
+    int_least64_t
+    get_l1d_metric(metric_name_t metric, unsigned int core) const;
+    int_least64_t
+    get_ll_metric(metric_name_t metric) const;
```
derekbruening, line 68: In our usage we routinely have a 3-level cache with a per-core L2 in between L1{I,D} and LLC (created with a config file).  Possibilities:

+ Add a get_l2_metric().  To take in a core index, it would have to only support direct parents of L1* and so would not support arbitrary cache layouts.
+ Export an iterator over all caches (split into l1, llc, and "other"==l2 separate iterators?) which returns some kind of key (the cache name?) to then pass back to query a metric.
+ Take in the cache by name (the name from the config file) -- less usable by itself without an iterator.

clients/drcachesim/simulator/cache_simulator.h
```diff
@@ -60,12 +60,26 @@ class cache_simulator_t : public simulator_t {
...
+
     // Exposed to make it easy to test
     bool
     check_warmed_up();
     uint64_t
     remaining_sim_refs() const;
 
+    unsigned int
+    get_core_num() const;
```
derekbruening, line 77: Since cache_simulator_knobs_t is already public -- should the knobs be returned?  IIRC a config file setup does set the key knobs fields?

clients/drcachesim/simulator/caching_device_stats.h
```diff
@@ -115,6 +141,10 @@ class caching_device_stats_t {
     // Print out write invalidations if cache is coherent.
     bool is_coherent_;
 
+    // References to the properties with statistics are held in the map with the
+    // statistic name as the key. Sample map element: {HITS, num_hits_}
+    std::map<metric_name_t, int_least64_t &> stats_map_;
+
```
derekbruening, line 147: Please add some kind of test.  Maybe in drcachesim_unit_tests or a new similar test?  Xref #4842.

derekbruening (commented):
Thank you for the contribution.  I have a few suggestions.

AssadHashmi:
add to whitelist

derekbruening:
Looks like #4954 was hit on A64 Jenkins -- maybe should be added to the known-flaky list.

derekbruening:
run arm tests

clients/drcachesim/simulator/cache_simulator.h
mat-kalinowski: Thanks for the suggestions! Sorry for the delay, I had few days off. I like the flexibility of accessing stats in arbitrary cache model and also having really simple interface. I thought about having common function for all caches, something like `int_least64_t get_cache_metric(unsigned level, unsigned core, metric_t metric)`
Does separation for instruction and data caches makes sense only for L1 cache in any model ?

derekbruening: I think the current code only supports i vs d at L1 but if it's easy not to assume that here that might help future support.

> something like `int_least64_t get_cache_metric(unsigned level, unsigned core, metric_t metric)`

Would this support asymmetric layouts?  I guess it would if you always walk from the core up `level` levels.

clients/drcachesim/simulator/caching_device_stats.h
mat-kalinowski: I have added unit test in the `drcachesim_unit_tests`. I couldn't test `Invalidates` metric properly because of the bug/issue that I described here: https://github.com/DynamoRIO/dynamorio/issues/5031

clients/drcachesim/simulator/cache_simulator.h
mat-kalinowski: Yes, knobs fields are set by config file setup. I have added suitable method.

clients/drcachesim/simulator/cache_simulator.h
mat-kalinowski: I tried to make this code as flexible as possible. Current approach should work with split cache of the higher level in simple models, when given higher level split cache is an ancestor of L1 (for example L1i -> L2i).

mat-kalinowski: Can you provide example of some simple asymmetric layout ? My current approach is to walk up from L1 cache and I think it should work properly.

derekbruening: > Can you provide example of some simple asymmetric layout ? My current approach is to walk up from L1 cache and I think it should work properly.

Never used or seen anything too weird with dr$sim, was just speculating about possible corner cases.  Such a walk should work with any layout though it seems.

mat-kalinowski:
Right now I am handling errors like this: 
```
if (parent == NULL) {
    ERRMSG("Wrong cache level: %u.\n", core);
    return 0;
}
```
I thought also about returning -1 without printing error message. Which approach you think is better? Also do you have some comments about newest commit with changes ?

derekbruening:
> Right now I am handling errors like this:
> 
> ```
> if (parent == NULL) {
>     ERRMSG("Wrong cache level: %u.\n", core);
>     return 0;
> }
> ```
> 
> I thought also about returning -1 without printing error message. Which approach you think is better? 

For get_cache_metric() I would think of it as a library routine likely to be used by third-party code and as such I'd return an error rather than printing an error.

> Also do you have some comments about newest commit with changes ?

Oh, was waiting for a (re-)review request (https://dynamorio.org/page_code_reviews.html#autotoc_md114 "When the requested changes have been pushed, you can ask for a re-review from the reviewers. This can be done by clicking the re-review button next to the reviewers' name on the right sidebar on the pull request page.")

clients/drcachesim/simulator/cache_simulator.h
```diff
@@ -43,6 +43,8 @@
 #include "cache.h"
 #include "snoop_filter.h"
 
+enum cache_split_t { DATA, INSTRUCTION };
```
derekbruening, line 46: Safer to put inside the class and not in the global namespace (xref #4343).  Or see other suggestions for the enum below.

clients/drcachesim/simulator/cache_simulator.h
```diff
@@ -60,12 +62,19 @@ class cache_simulator_t : public simulator_t {
     bool
     print_results() override;
 
+    int_least64_t
+    get_cache_metric(unsigned level, metric_name_t metric, unsigned core = 0,
```
derekbruening, line 66: I would re-order the parameters with the metric first and grouping the 3 identifiers of which cache (level, core, split) together.

clients/drcachesim/simulator/caching_device_stats.h
```diff
@@ -49,6 +50,20 @@ enum invalidation_type_t {
     INVALIDATION_COHERENCE,
 };
 
+enum metric_name_t {
```
derekbruening, line 53: I know the existing code has global scope enums but we're trying to avoid that going forward (xref #4343), and here we have very simple names that could conflict.  We do require C++11 so this could be a scoped enum (`typedef enum class metric_name_t`) where it would require the prefix `metric_name_t::HITS`; or at least use a prefix like the existing enums so `METRIC_HITS`; or something, to avoid global symbols with very simple names like `HITS` which could conflict (this code is statically linked into some very large code bases).

clients/drcachesim/simulator/cache_simulator.cpp
```diff
@@ -594,6 +594,48 @@ cache_simulator_t::print_results()
...
 
+int_least64_t
+cache_simulator_t::get_cache_metric(unsigned level, metric_name_t metric, unsigned core,
+                                    cache_split_t split) const
+{
+    caching_device_t *curr_cache;
+
+    if (core >= knobs_.num_cores) {
+        ERRMSG("Wrong core number: %u.\n", core);
+        return 0;
```
derekbruening, line 605: Xref the top-level thread: I would vote for returning an error.  Could go so far as error enum types; or just return -1 (document that's the error return code) either w/ or w/o the print (could see the print useful if there's no errno; could see it annoying linked into a large system that maybe doesn't know the cache layout and is passing all possible levels and using return code to find the bounds of the layout -- there an errno would be nice).

Ditto below.

clients/drcachesim/simulator/caching_device_stats.h
derekbruening: OK, best to put a `XXX i#5031: ` or `TODO i#5031: ` comment in the test about adding that metric.

derekbruening (commented):
Thank you for adding the test.  Just a few comments mostly about names and then should be good to go.

mat-kalinowski:
I've changed enum types to the `enum class` to prevent polluting global namespaces. I've implemented `stats_error_t` as a plain enum and added `STATS_` prefix because I needed to return it as an int from the `get_cache_metric` function.

clients/drcachesim/simulator/cache_simulator.h
```diff
@@ -63,8 +75,8 @@ class cache_simulator_t : public simulator_t {
     print_results() override;
 
     int_least64_t
-    get_cache_metric(unsigned level, metric_name_t metric, unsigned core = 0,
-                     cache_split_t split = DATA) const;
+    get_cache_metric(metric_name_t metric, unsigned level, unsigned core = 0,
```
derekbruening, line 78: Please comment that a negative return value is an error and is of type stats_error_t

clients/drcachesim/simulator/cache_simulator.h
```diff
@@ -42,8 +42,20 @@
 #include "cache_stats.h"
 #include "cache.h"
 #include "snoop_filter.h"
+#include <limits.h>
 
-enum cache_split_t { DATA, INSTRUCTION };
+enum class cache_split_t { DATA, INSTRUCTION };
+
+// Error codes returned when passing wrong parameters to the function for
+// accessing cache statistics.
```
derekbruening, line 50: Use the name `get_cache_metric` rather than a description so a reader knows precisely which one

derekbruening (approved):
I just have some comment suggestions, otherwise looks good.  Thank you for the contribution.

mat-kalinowski:
run arm tests

derekbruening:
Thanks again for the contribution.

One nit: For future merges please remove the duplicate title line (xref https://dynamorio.org/page_code_reviews.html#autotoc_md117):

```
commit 9574fca502ee03661dfd788f8744a8881f19eb63

    i#3068: API for accessing cache model metrics (#5020)
    
    i#3068: API for accessing cache model metrics
    
    This patch contains implementation of the API for accessing separate
    metrics gathered during the simulation.
    
    Functions for accessing metrics were added for all caches in the
    default configuration: L1I, L1D and LL. Also additional functions
    for getting configuration options of the cache_simulator_t.
    
    Issue: #3068
```

PR #5012, opened 2021-07-15
i#2626: AArch64 v8.2 Decode: Add missing instructions, part 1

MDevereau:
Adds the following instructions to the codec:
- CMLE
- FCMLE
- FCMPE
- FCVTXN
- FCVTXN2

Issue: #2626

PR #4966, opened 2021-06-24
i#2626: AArch64 decode: Add FCSEL FMAXV FMINV FCMP FCCMP FCCMPE

AssadHashmi:
Adds the following instructions to the codec:

FCSEL Hd, Hn, Hm, cond 
FCSEL Sd, Sn, Sm, cond 
FCSEL Dd, Dn, Dm, cond 
FMAXV Vd, Vn.T
FMINV Vd, Vn.T
FCMP Hn, Hm
FCMP Hn, #0.0
FCMP Sn, Sm
FCMP Sn, #0.0
FCMP Dn, Dm
FCMP Dn, #0.0
FCCMP Hn, Hm, #nzcv, cond
FCCMP Sn, Sm, #nzcv, cond
FCCMP Dn, Dm, #nzcv, cond
FCCMPE Hn, Hm, #nzcv, cond
FCCMPE Sn, Sm, #nzcv, cond
FCCMPE Dn, Dm, #nzcv, cond

Issue: #2626

PR #5024, opened 2021-07-22
i#5023: Mark FCSEL, FC*MP* as reading/writing aflags

### LINKED ISSUE #5023: APP CRASH: Regression from AArch64 decoder additions: FCSEL and FC*MP* not marked as reading/writing flags, resulting in mysterious crashes
PR #4966 added the FCSEL FCMP FCCMP FCCMPE opcodes, and PR #5012 added FCMPE.  But they failed to update this code in codec.c:
```C
         /* FIXME i#2626: When handled by decoder, add:
          * opc == OP_fcsel
          */
         eflags |= EFLAGS_READ_NZCV;

         /* FIXME i#2626: When handled by decoder, add:
          * opc == OP_fccmp || opc == OP_fccmpe || opc == OP_fcmp || opc == OP_fcmpe
          */
         eflags |= EFLAGS_WRITE_NZCV;
```

And since the opcodes are now decoded properly, the workaround from PR #4688 where OP_xx is marked as reading and writing the flags *no longer applies*.  This leads to incorrect aflags behavior where these opcodes are not marked as reading or writing.   Which leads to instrumentation corrupting application state in subtle ways, leading to strange application behavior and difficult-to-debug crashes.

We hit this on a large application.  It took me several days to figure out the general area of the culprit and track it down to these opcodes.  We've been down this road before, with difficult to debug problems coming from decoder bugs: that's what led to PR #4688's workaround.  @AssadHashmi  -- Are there further tests we could add?  Further sanity checks to catch this type of regression ahead of time?  It's not clear there's anything simple, unfortunately.

Should NZCV reading/writing be part of the codec entry instead of separate?  That at least avoids two separate places that can get out of sync.

derekbruening:
Fixes a regression from PR #4966 and PR #5012 where FCSEL and FC*MP*
opcodes were added to the decoder but the flag reading/writing code
was not updated for them, meaning they no longer went into the OP_xx
workaround, resulting in incorrect aflags attributes and resulting
instrumentation bugs.

Tested on the large application where app crashes were first seen.
It's not clear how to easily add a regression test for this kind of
thing for all opcodes.

Fixes #5023

### DIFF
--- a/core/ir/aarch64/codec.c
+++ b/core/ir/aarch64/codec.c
@@ -1,5 +1,5 @@
 /* **********************************************************
- * Copyright (c) 2017-2020 Google, Inc.  All rights reserved.
+ * Copyright (c) 2017-2021 Google, Inc.  All rights reserved.
  * Copyright (c) 2016 ARM Limited. All rights reserved.
  * **********************************************************/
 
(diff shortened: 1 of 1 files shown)


AssadHashmi:
My bad for not enabling NZCV r/w...even when there's comments pointing to the instruction opcodes! :-(
Thanks for fixing.

AssadHashmi (approved):

Issue #5026, opened 2021-07-26
drmemtrace instruction delay counting 20x slowdown on multi-threaded AArch64 apps


derekbruening:
Xref #4487 where inlined instruction counting was added for drmemtrace's delay feature which greatly improved single-threaded performance.

However, running on a large multi-threaded AArch64 app, I'm seeing huge slowdowns from instruction counting: on the order of 20x versus plain DR!

I tried some experiments:
+ Count down and use TBZ to avoid aflags clobbering: doesn't make much of a difference so it's not the bottleneck, but we should keep this approach and it should help once the bottleneck is gone.
+ Remove drx_insert_counter_update() to avoid re-acquiring the address and value: again not much difference vs the big bottleneck.
+ s/ldar/ldr;s/stlr/str => 2x faster, so the atomicity is part of the problem, but not the whole story.
+ Use thread-local counters => nearly all of the 20x is gone (just a quick experiment here w/o any conditional; final perf will be a little higher) so the fundamental issue is contention on the single global counter.

My proposal is to use thread-private counters, count down, use TBZ to see when they hit 0, and only then update the global value and see if it hit the threshold.  Maybe a thread would update the global every 1K or 10K instructions.  We'd use this mode for any large delay values and we'd document it won't hit it exactly which should be fine when asking to delay 4 billion instructions.  For small values (say, <10M) we'd use the current change-global-every-time approach.

derekbruening:
I implemented the thread-local, with threads adding to the global every 10K instrs.  It is *much* faster: now 14% slowdown on this app vs plain DR, instead of 20x!!

However, testing on a small many-threaded app shows the actual instr count when tracing starts after the flush can be quite a bit larger than the trigger:
```
$ bin64/drrun -t drcachesim -trace_after_instrs 30M -- ~/dr/test/threadsig 20 1000000
<...>
T297628 added 10001 => 31430502
T297628 added 10012 => 31440514
T297628 added 10007 => 31450521
+******* T297628 added 10001 => 31460522
+******* T297627 added 10001 => 31470523
+******* T297627 added 10012 => 31480535
<...>
+******* T297639 added 10001 => 67624549
+******* T297639 added 10012 => 67634561
+******* T297639 added 10007 => 67644568
Hit delay threshold: enabling tracing.
```

There the other threads executed another 30M instrs by the time the flush finishes and we start tracing!

We frequently reach 1.5x or 2x the trigger:
```
[...]

PR #5027, opened 2021-07-26
i#5026 icount perf: Avoid flags for arm drmemtrace

derekbruening:
Switches the instruction counting inlined code for AArch64
drmemtrace's delayed-by-instruction-count feature to count down and
use TBZ, avoiding use of the arithmetic flags.

Adds INSTR_CREATE_ support for TBZ and TBNZ.

Fixes missing support for subtracting in drx_insert_counter_update()
for AArch64 (fatal encoder errors are raised without this).
Adds a drx test for subtracting.

Issue: #5026

### DIFF
--- a/clients/drcachesim/tracer/tracer.cpp
+++ b/clients/drcachesim/tracer/tracer.cpp
@@ -1467,8 +1467,8 @@ enable_delay_instrumentation()
         DR_ASSERT(false);
     schedule_tracing_lock = dr_mutex_create();
 #if defined(AARCH64) && defined(DELAYED_CHECK_INLINED)
-    /* By counting down we can avoid clobbering aflags in our comparison. */
-    instr_count = op_trace_after_instrs.get_value();
+    /* By counting down to < 0, we can avoid clobbering aflags in our comparison. */
+    instr_count = op_trace_after_instrs.get_value() - 1;
 #endif
 }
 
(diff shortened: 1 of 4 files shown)


clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1466,6 +1466,10 @@ enable_delay_instrumentation()
             event_delay_bb_analysis, event_delay_app_instruction, &memtrace_pri))
         DR_ASSERT(false);
     schedule_tracing_lock = dr_mutex_create();
+#if defined(AARCH64) && defined(DELAYED_CHECK_INLINED)
+    /* By counting down we can avoid clobbering aflags in our comparison. */
+    instr_count = op_trace_after_instrs.get_value();
```
sapostolakis, line 1471: This will hit the threshold after op_trace_after_instrs.get_value() + 1 instructions, since it calls the hit_instr_count_threshold when the counter becomes negative (top bit turns to 1). If I am not missing something, this should change to instr_count = op_trace_after_instrs.get_value() - 1;

sapostolakis (approved):

clients/drcachesim/tracer/tracer.cpp
derekbruening: Right, changed to -1.  There is no test of the precise count -- not sure it's worth trying to make one, since this isn't used as a precise trigger in tracing use cases.

sapostolakis: Yeah I do not think it is worth making a test for precise count. For practical purposes even the one-off does not matter but it was worth fixing since I noticed it.

PR #5028, opened 2021-07-27
i#5026 icount perf: Use per-thread counters

### LINKED ISSUE #5026: drmemtrace instruction delay counting 20x slowdown on multi-threaded AArch64 apps

derekbruening:
For drmemtrace -trace_after_instrs, if the threshold is large (>10M),
we use per-thread counters and only check the global value every 10K
instructions in each thread.  The approximate result is fine for these
use cases.

Tested on the original app where the global counter contention results
in a 20x slowdown over plain DR.  The new scheme has a 14% slowdown
over plain DR on this app, a massive speedup.

Fixes #5026

### DIFF
--- a/clients/drcachesim/common/options.cpp
+++ b/clients/drcachesim/common/options.cpp
@@ -241,8 +241,14 @@ droption_t<bytesize_t> op_trace_after_instrs(
     "Do not start tracing until N instructions",
     "If non-zero, this causes tracing to be suppressed until this many dynamic "
     "instruction "
-    "executions are observed.  At that point, regular tracing is put into place.  Use "
-    "-max_trace_size to set a limit on the subsequent trace length.");
+    "executions are observed.  At that point, regular tracing is put into place. "
+    "The threshold should be considered approximate, especially for larger values. "
+    "Switching to regular tracing takes some amount of time during which other "
+    "threads than the one that triggered the switch can continue to execute, "
+    "resulting in a larger count of executed instructions before tracing actually "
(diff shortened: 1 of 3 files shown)


sapostolakis (approved):
This approximation seems very reasonable given the massive perf gains.

PR #5029, opened 2021-07-28
i#5021: Set frozen timestamp at drmemtrace max size

### LINKED ISSUE #5021: drmemtrace -max_global_trace_refs fails to flush threads when the max is hit
The -max_global_trace_refs feature in drcachesim's drmemtrace fails to flush thread buffers when the max is hit.  Since execution continues, what happens is that much later when the app exits we then have thread exits and unflushed data in thread buffers emitted into the trace with timestamps from that much later time, producing a confusing time gap in the trace.

For our use case, what we really want is to detach when the max is hit, but we need #2644 for that.  Barring that, we could put in a synchall when the max is hit and go and flush and exit all the thread buffers: if we add logic for accessing another thread's buffer.  Another choice is to record the timestamp when the max is hit and use that when the app finally exits.

If we implement -detach_after_tracing: would we make -max_global_trace_refs an alias of it (with compatibility breakage notes), or go and fix it in addition??

derekbruening:
Adds a new drmemtrace mechanism to set a frozen timestamp for all
future entries, to avoid huge time gaps when -max_global_trace_refs is
reached but existing thread buffers and exits are not emitted until
much later when the app exits.

Adding a small regression test seems difficult without flakiness as
this involved real-time gaps.  Tested manually:

Pre-fix we see a 2s gap before the thread exit:

    $ ninja && ctest -V -R max-global
    $ bin64/drrun -t drcachesim -simulator_type view -indir suite/tests/tool.drcacheoff.max-global*.dir 2>&1 | grep timestamp
    T3427537 <marker: timestamp 13271354817417504>
    T3427537 <marker: timestamp 13271354817418955>
    T3427537 <marker: timestamp 13271354817624561>
    T3427537 <marker: timestamp 13271354817657186>
    T3427537 <marker: timestamp 13271354817659466>
    T3427537 <marker: timestamp 13271354817664998>
    T3427537 <marker: timestamp 13271354817667972>
    T3427537 <marker: timestamp 13271354817701187>
    T3427537 <marker: timestamp 13271354819175717>

After adding a frozen timestamp:

    $ bin64/drrun -t drcachesim -simulator_type view -indir suite/tests/tool.drcacheoff.max-global*.dir 2>&1 | grep timestamp
    T3429223 <marker: timestamp 13271355614195474>
    T3429223 <marker: timestamp 13271355614196983>
    T3429223 <marker: timestamp 13271355614399149>
    T3429223 <marker: timestamp 13271355614432167>
    T3429223 <marker: timestamp 13271355614434675>
    T3429223 <marker: timestamp 13271355614438892>
    T3429223 <marker: timestamp 13271355614441260>
    T3429223 <marker: timestamp 13271355614474587>
    T3429223 <marker: timestamp 13271355614475843>

Fixes #5021

### DIFF
--- a/clients/drcachesim/tracer/instru.h
+++ b/clients/drcachesim/tracer/instru.h
@@ -218,6 +218,11 @@ public:
     // This is a per-buffer-writeout header.
     virtual int
     append_unit_header(byte *buf_ptr, thread_id_t tid) = 0;
+    virtual void
+    set_frozen_timestamp(uint64 timestamp)
+    {
+        frozen_timestamp_ = timestamp;
+    }
 
     // These insert inlined code to add an entry into the trace buffer.
     virtual int
@@ -277,6 +282,7 @@ protected:
     bool memref_needs_full_info_;
     drvector_t *reg_vector_;
     bool disable_optimizations_;
+    uint64 frozen_timestamp_ = 0;
 
 private:
     instru_t()
--- a/clients/drcachesim/tracer/instru.h
+++ b/clients/drcachesim/tracer/instru.h
@@ -282,6 +282,8 @@ protected:
     bool memref_needs_full_info_;
     drvector_t *reg_vector_;
     bool disable_optimizations_;
+    // Stores a timestamp to use for all future unit headers.  This is meant for
+    // avoiding time gaps for max-limit scenarios (i#5021).
     uint64 frozen_timestamp_ = 0;
 
 private:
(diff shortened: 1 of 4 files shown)


clients/drcachesim/tracer/instru.h
```diff
@@ -277,6 +282,7 @@ class instru_t {
     bool memref_needs_full_info_;
     drvector_t *reg_vector_;
     bool disable_optimizations_;
+    uint64 frozen_timestamp_ = 0;
```
abhinav92003, line 285: Maybe add a comment here on this field as well.

abhinav92003 (approved):

clients/drcachesim/tracer/instru.h
derekbruening: Done.

PR #5032, opened 2021-08-02
i#4990: Cross injection x64 to x86 failing (#803) 

N0fix:
Small edits to fix troubles we talked about [here](https://github.com/DynamoRIO/dynamorio/pull/4989#issuecomment-890041292).
This is adding a `DYNAMO_OPTION(inject_x64)` on commit [i#4700](https://github.com/DynamoRIO/dynamorio/commit/c5a368d278b95f48f557625a0710a91f886299b5), due to troubles with syscalls on x64->Wow64 injection using `-inject_x64` option.


N0fix:
Not sure why CI vs2017-32 is failing. I guess this is a CI issue.

core/arch/mangle_shared.c
```diff
@@ -1824,7 +1824,7 @@ find_syscall_num(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr)
 
     if (prev == NULL) {
 #if defined(WINDOWS) && defined(X64)
-        if (get_os_version() >= WINDOWS_VERSION_10_1511) {
+        if (get_os_version() >= WINDOWS_VERSION_10_1511 && !DYNAMO_OPTION(inject_x64)) {
```
derekbruening, line 1827: I don't understand: this same code will be there when in 64-bit mode: why disable it?

core/arch/mangle_shared.c
```diff
@@ -1824,7 +1824,7 @@ find_syscall_num(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr)
```
derekbruening, line 1824: nit: Please follow the commit title and body formatting conventions

derekbruening, line 1824: Please add a regression test: any sanity test of -inject_x64 will do.

core/arch/mangle_shared.c
N0fix: This code messes up with syscall handling, prompting an error on some syscalls when running WoW64 child injected from x64 parent. Disabling this part prevented the wrong deduction of syscall I guess. I did not dive into debugging to understand why this part messes up with syscall on x64->WoW64, but it sure does.

derekbruening: It seems like it needs to check the current mode being 64-bit, since the existing behavior will be wanted for 64-bit mode: so it would depend on the mode rather than the -inject_x64 option.

core/arch/mangle_shared.c
N0fix: I will do.

N0fix: I will do.

N0fix:
Messed up commit message in latest commit, I will fix this tomorrow, alongside other CI issues.

core/arch/mangle_shared.c
derekbruening: (Please mark comments as Resolved when done.)

core/win32/inject.c
```diff
@@ -1571,7 +1571,7 @@ inject_into_new_process(HANDLE phandle, HANDLE thandle, char *dynamo_path, bool
 #endif
         }
         if (hook_location == 0) {
-            bool target_64 = !x86_code IF_X64(|| DYNAMO_OPTION(inject_x64));
+            bool target_64 = !x86_code;
```
derekbruening, line 1574: Wait: I thought `x86_code` was whether the target executable was 32-bit, and that -inject_x64 wanted to take over in the 64-bit ntdll and so we wanted the original code here?

suite/tests/CMakeLists.txt
```diff
@@ -4511,6 +4511,71 @@ else (UNIX)
       PASS_REGULAR_EXPRESSION ".*${expect}[ \n]*$")
   endif (X86)
 
+
+  # Cross-arch injection x64 to x86 test using -inject_x64.
+  if(X64)
```
derekbruening, line 4516: style nit: missing space

core/arch/mangle_shared.c
```diff
@@ -1824,7 +1824,8 @@ find_syscall_num(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr)
 
     if (prev == NULL) {
 #if defined(WINDOWS) && defined(X64)
-        if (get_os_version() >= WINDOWS_VERSION_10_1511) {
+        if (get_os_version() >= WINDOWS_VERSION_10_1511 &&
+            !is_wow64_process(NT_CURRENT_PROCESS)) {
```
derekbruening, line 1828: Repeating comment from prior review: when in 64-bit mode we still want this, so we want to check the current mode: since it's passed instrs, the instr mode makes the most sense: `instr_get_isa_mode`.

suite/tests/CMakeLists.txt
```diff
@@ -4511,6 +4511,71 @@ else (UNIX)
...
+    set(inject_x64_dir ${CMAKE_CURRENT_BINARY_DIR}/inject_x64)
+    set(test_bindir suite/tests/bin)
+    file(MAKE_DIRECTORY ${inject_x64_dir})
+    # Unfortunately, while CMake 3.13+ supports create_symlink on Windows, that
+    # only works if the user has the right privileges, so we can't set up
+    # lib dir symlinks like we do for unix cross-arch build-and-test.
+    # We have to make a copy in order to have parallel dirs with the two bitwidths.
```
derekbruening, line 4523: Instead of duplicating everything verbatim, including these long cmoments, it would be nice to share code and comments to reduce maintenance burden and future staleness when one is changed w/o the other.  If you're not comfortable enough w/ CMake to make a shared function or macro, at least add a `XXX` comment up above about sharing all this duplication.

derekbruening (commented):
Not sure I understand the target_64 change in inject.c?

PR #5035, opened 2021-08-04
i5033: Fix intermittent crash in api.drdecode aarch64 test

### LINKED ISSUE #5033: CRASH: code_api|api.drdecode SEGFAULT/SIGILLs intermittently on aarch64 armv8.0
When run on certain v8.0 hardware, the api.decode test will segfault or SIGILL intermittently. The test harness eats the error message but when the binary is run independently, the fault will appear. Digging into this it appears that the following line in the test triggers the crash
```
uint written = ((uint(*)(void))generated_code)();
```

This is running assembly created by the lines above:
```
instrlist_t *ilist = instrlist_create(GD);
    instr_t *callee = INSTR_CREATE_label(GD);
    instrlist_append(
        ilist,
        XINST_CREATE_move(GD, opnd_create_reg(DR_REG_X1), opnd_create_reg(DR_REG_LR)));
    instrlist_insert_mov_instr_addr(GD, callee, (byte *)ilist, opnd_create_reg(DR_REG_X0),
                                    ilist, NULL, NULL, NULL);
    instrlist_append(ilist, INSTR_CREATE_blr(GD, opnd_create_reg(DR_REG_X0)));
    instrlist_append(ilist, INSTR_CREATE_ret(GD, opnd_create_reg(DR_REG_X1)));
    instrlist_append(ilist, callee);
    instrlist_insert_mov_immed_ptrsz(GD, 0xdeadbeef, opnd_create_reg(DR_REG_X0), ilist,
                                     NULL, NULL, NULL);
    instrlist_append(ilist, XINST_CREATE_return(GD));
```

I modified the test to print out the instructions that are being generated, which seem to be fine
```
orr    %xzr %x30 lsl $0x0000000000000000 -> %x1
movz   @0x0000aaab256572c0[2byte] lsl $0x00 -> %x0
movk   %x0 @0x0000aaab256572c0[2byte] lsl $0x10 -> %x0
movk   %x0 @0x0000aaab256572c0[2byte] lsl $0x20 -> %x0
blr    %x0 -> %x30
ret    %x1
<label>
movz   $0xbeef lsl $0x00 -> %x0
movk   %x0 $0xdead lsl $0x10 -> %x0
ret    %x30
```

Oddly enough, if the binary is moved to an arm v8.2 machine, it runs just fine.

Looking at the dump, it appears the SIGILL comes from
```
Program terminated with signal SIGILL, Illegal instruction.
#0  0x0000ffff885d3000 in ?? ()
```

```
0xffff885d3000      mov x1, x30
```

**To Reproduce**
1. Build the unit tests in debug mode
2. Run `./suite/tests/bin/api.drdecode` until you hit a segfault/sigill on an arm v8.0 machine

**Expected behaviour**
Unit test passes

**Versions**
 - most recent version of upstream/master
 - Ubuntu 18.04.5 LTS
 - 64-bit

joshua-warburton:
This patch fixes an intermittent issue when the CI tests are run on AArch64 V8.0 machines. Depending on the compiler used to build the tests, SIGILLs or SEGFAULTS are triggered when generated code is run before the generated code is flushed from the cache. 

This change adds an explicit cache flush in the test that has been exhibiting this flaky behaviour.

Fixes #5033

### DIFF
--- a/suite/tests/api/drdecode_aarch64.c
+++ b/suite/tests/api/drdecode_aarch64.c
@@ -82,6 +82,8 @@ test_mov_instr_addr(void)
     instrlist_encode(GD, ilist, generated_code, true);
     protect_mem(generated_code, gencode_max_size, ALLOW_EXEC | ALLOW_READ);
 
+    __builtin___clear_cache(generated_code, generated_code + gencode_max_size);
+
     uint written = ((uint(*)(void))generated_code)();
     ASSERT(written == 0xdeadbeef);
 
--- a/suite/tests/api/drdecode_aarch64.c
+++ b/suite/tests/api/drdecode_aarch64.c
@@ -82,7 +82,7 @@ test_mov_instr_addr(void)
     instrlist_encode(GD, ilist, generated_code, true);
     protect_mem(generated_code, gencode_max_size, ALLOW_EXEC | ALLOW_READ);
 
-    // Make sure to flush the cache so v8.0 versions of aarch64 avoid stale icache
+    // Make sure to flush the cache to avoid stale icache
     // values which can lead to SEGFAULTs or SIGILLS on the subsequent attempted
     // execution (i#5033)
     __builtin___clear_cache(generated_code, generated_code + gencode_max_size);
--- a/suite/tests/api/drdecode_aarch64.c
+++ b/suite/tests/api/drdecode_aarch64.c
@@ -82,6 +82,9 @@ test_mov_instr_addr(void)
     instrlist_encode(GD, ilist, generated_code, true);
     protect_mem(generated_code, gencode_max_size, ALLOW_EXEC | ALLOW_READ);
 
+    // Make sure to flush the cache so v8.0 versions of aarch64 avoid stale icache
+    // values which can lead to SEGFAULTs or SIGILLS on the subsequent attempted
+    // execution (i#5033)
     __builtin___clear_cache(generated_code, generated_code + gencode_max_size);
 
     uint written = ((uint(*)(void))generated_code)();
--- a/suite/tests/api/drdecode_aarch64.c
+++ b/suite/tests/api/drdecode_aarch64.c
@@ -82,9 +82,10 @@ test_mov_instr_addr(void)
     instrlist_encode(GD, ilist, generated_code, true);
     protect_mem(generated_code, gencode_max_size, ALLOW_EXEC | ALLOW_READ);
 
-    // Make sure to flush the cache to avoid stale icache
-    // values which can lead to SEGFAULTs or SIGILLS on the subsequent attempted
-    // execution (i#5033)
+    /* Make sure to flush the cache to avoid stale icache values which
+     * can lead to SEGFAULTs or SIGILLS on the subsequent attempted
+     * execution (i#5033)
+     */
     __builtin___clear_cache(generated_code, generated_code + gencode_max_size);
 
     uint written = ((uint(*)(void))generated_code)();


suite/tests/api/drdecode_aarch64.c
```diff
@@ -82,6 +82,8 @@ test_mov_instr_addr(void)
     instrlist_encode(GD, ilist, generated_code, true);
     protect_mem(generated_code, gencode_max_size, ALLOW_EXEC | ALLOW_READ);
 
+    __builtin___clear_cache(generated_code, generated_code + gencode_max_size);
```
abhinav92003, line 85: `suite/tests/api/ibl-stress.c` also dynamically generates code. Perhaps we should add the same fix there as well.

suite/tests/api/drdecode_aarch64.c
```diff
@@ -82,6 +82,8 @@ test_mov_instr_addr(void)
     instrlist_encode(GD, ilist, generated_code, true);
```
abhinav92003, line 82: Regarding the PR's description and title: please take a look at the section on Commit Messages at https://dynamorio.org/page_code_reviews.html. Particularly, please shorten the PR's title, and add more details about the bug and the issue number to the description.

AssadHashmi:
add to whitelist

abhinav92003 (commented):
Thanks for root-causing and fixing the issue! I left some comments on the PR.

suite/tests/api/drdecode_aarch64.c
joshua-warburton: I had a look there, and as it's not immediately trying to execute the code it generates and we haven't seen an issue there in any of the runs we've run.

abhinav92003: Maybe add a comment in `drdecode_aarch64.c` about when we need to flush the icache.

> as it's not immediately trying to execute the code it generates

So can the bug still manifest in `ibl-stress` if the icache is big enough that it continued to have the stale instrs?

suite/tests/api/drdecode_aarch64.c
abhinav92003: Thanks for the changes. Note that conventions also suggest a `#` in the issue number in the title, i.e. "i#5033: Fix intermittent..."

suite/tests/api/drdecode_aarch64.c
```diff
@@ -82,6 +82,11 @@ test_mov_instr_addr(void)
...
     protect_mem(generated_code, gencode_max_size, ALLOW_EXEC | ALLOW_READ);
 
+    // Make sure to flush the cache so v8.0 versions of aarch64 avoid stale icache
```
abhinav92003, line 85: Based on commenting conventions at https://dynamorio.org/page_code_style.html, please use /* */ comments in this C file.

abhinav92003 (commented):
Thanks for the changes, and apologies for the late re-review. We typically use the "request re-review" button to avoid delays[1]. Please also mark comments as "Done"/"Complete" when they are resolved. 

[1]: Responding to Review Comments section at https://dynamorio.org/page_code_reviews.html

abhinav92003 (approved):
LGTM. Before merging this PR, please take a look at the section on final commit message at https://dynamorio.org/page_code_reviews.html.

joshua-warburton:
slight change in the wording of the comment as the stale cache issue isn't dependent on the aarch64 version but optional implementation details.

abhinav92003 (approved):
LGTM. Please mark the pending comments as resolved if complete.

Issue #4407, opened 2020-08-14
[CRASH] Stack smashing with 32-bit target binary on AMD CPU


ogimgd:
**Describe the bug**

Dynamorio crashes when I try to run 32bit binary with it.
The strange thing here is that it's not reproducible on ANY machine. It does work on my workstation, but it crashes on server with AMD EPYC 7702.

**To Reproduce**

Here is a minimal example:

```
#include <stdio.h>

int main() {
  printf("Hello ");
  printf(" World\n");
  return 0;
}
```

That's it. Just compile it in 32 bit mode:

```
gcc -g test.c -m32 -o test_prog
```

And run it like that:

```
./bin32/drrun -debug -c ./api/bin/libbbcount.so -- test_prog
```
And here is the crash:

```
<Starting application /home/gulakov/dynamorio/build/test_prog (166564)>
<Initial options = -no_dynamic_options -client_lib '/home/gulakov/dynamorio/build/api/bin/libbbcount.so;0;' -client_lib32 '/home/gulakov/dynamorio/build/api/bin/libbbcount.so;0;' -code_api -stack_size 56K -signal_stack_size 32K -max_elide_jmp 0 -max_elide_call 0 -early_inject -emulate_brk -no_inline_ignored_syscalls -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct >
<Paste into GDB to debug DynamoRIO clients:
set confirm off
add-symbol-file '/home/gulakov/dynamorio/build/api/bin/libbbcount.so' 0xf7b7f970
add-symbol-file '/home/gulakov/dynamorio/build/lib32/debug/libdynamorio.so' 0xf7bccff0
add-symbol-file '/home/gulakov/dynamorio/build/ext/lib32/debug/libdrx.so' 0xf7b700d0
add-symbol-file '/home/gulakov/dynamorio/build/ext/lib32/debug/libdrreg.so' 0xf7b62080
add-symbol-file '/home/gulakov/dynamorio/build/ext/lib32/debug/libdrmgr.so' 0xf7b520e0
add-symbol-file '/lib32/libc.so.6' 0xf798d610
add-symbol-file '/lib/ld-linux.so.2' 0xf794cab0
>
Client bbcount is running
*** stack smashing detected ***: <unknown> terminated
[1]    166564 abort      ./bin32/drrun -debug -c ./api/bin/libbbcount.so -- test_prog
```

**Expected behavior**

'Hello world' in the output and normal exit.

[...]

PR #5037, opened 2021-08-04
Fixed 4407 issue

JivanH:
Fixed https://github.com/DynamoRIO/dynamorio/issues/4407# issue


core/arch/emit_utils_shared.c
```diff
@@ -5120,7 +5120,11 @@ update_syscalls(dcontext_t *dcontext)
     generated_code_t *code = THREAD_GENCODE(dcontext);
     protect_generated_code(code, WRITABLE);
     pc = get_do_syscall_entry(dcontext);
-    update_syscall(dcontext, pc);
+    if (IF_X64_ELSE(true,
+                    !(get_syscall_method() == SYSCALL_METHOD_SYSCALL &&
```
derekbruening, line 5124: So the proposal is to have the global syscalls remain `int 0x80`, rather than figuring out what the problem is with 32-bit `syscall`?  Please at least include a comment explaining this, that the actual problem is still unknown.

But what about:
+ The `syscall` instruction clobbers `ecx` IIRC -- so wouldn't the registers holding parameters be different?  Is the 2nd param not in ecx but somewhere else??
+ What about inlined syscalls inside a fragment in the code cache?  They will be left as `syscall`.  Does the same problem that affects the global generated code affect them, and thus this is not a complete solution?

SweetVishnya:
`drrun` still crashes on program that reads from file.

file_read.c:
```
#include <sys/types.h>
#include <sys/stat.h>
#include <fcntl.h>
#include <unistd.h>
#include <stdio.h>
#include <stdlib.h>

int main(int argc, char** argv)
{
    int j = 0;
    char buffer[4];
    int fd = open(argv[1], O_RDONLY, S_IRWXU | S_IRWXG | S_IRWXO);

    if (fd < 0) return 1;

    read(fd, buffer, 4);

    if (buffer[0] == 'b')
    {
        printf("Path 1T\n");
    }
    else
    {
        printf("Path 1F\n");
    }

    if (buffer[1] == 'a')
    {
        printf("Path 2T\n");
    }
    else
    {
        printf("Path 2F\n");
    }

    if (buffer[2] == 'd')
    {
        printf("Path 3T\n");
    }
    else
    {
        printf("Path 3F\n");
    }

    if (buffer[3] == '!')
    {
        printf("Path 4T\n");
    }
    else
    {
        printf("Path 4F\n");
    }

    return 0;
}
```

Steps to reproduce:

```
gcc -m32 file_read.c -o file_read
echo xxxx > input && dynamorio/bin32/drrun -- ./file_read input
```

Program crashes and does not print anything.

JivanH:
I changed my code.

The problem was in hook_vsyscall. 
AMD `__kernel_vsyscall` uses syscall instruction instead of `sysenter` and hook_vsyscall() method does not call.

core/dispatch.c
```diff
@@ -1803,20 +1806,23 @@ adjust_syscall_continuation(dcontext_t *dcontext)
...
         }
 #    endif
-    } else if (vsyscall_syscall_end_pc != NULL &&
-               /* PR 341469: 32-bit apps (LOL64) on AMD hardware have
-                * OP_syscall in a vsyscall page
-                */
-               get_syscall_method() != SYSCALL_METHOD_SYSCALL) {
+    } else if (vsyscall_syscall_end_pc != NULL) {
+        /* PR 341469: 32-bit apps (LOL64) on AMD hardware have
+         * OP_syscall in a vsyscall page
```
derekbruening, line 1811: But this comment went with the exclusion for the method being SYSCALL -- so it doesn't make sense stranded here now?

core/dispatch.c
```diff
@@ -1803,20 +1806,23 @@ adjust_syscall_continuation(dcontext_t *dcontext)
...
         /* If we fail to hook we currently bail out to int; but we then
          * need to manually jump to the sysenter return point.
          * Once we have PR 288330 we can remove this.
          */
-        if (dcontext->asynch_target == vsyscall_syscall_end_pc) {
-            ASSERT(vsyscall_sysenter_return_pc != NULL);
-            dcontext->asynch_target = vsyscall_sysenter_return_pc;
-            LOG(THREAD, LOG_SYSCALLS, 3, "%s: asynch_target => " PFX "\n", __FUNCTION__,
-                dcontext->asynch_target);
+        if (IF_X86_32((syscall_method_is_syscall && cpu_info.vendor == VENDOR_AMD) ||)
```
derekbruening, line 1818: I don't follow: the previous code knew about AMD (see the comment above) and deliberately excluded SYSCALL -- but now this PR adds logic to include SYSCALL for AMD which is the opposite.  If the original was wrong: why keep the `!syscall_method_is_syscall` at all?

If you could elaborate on what the problem is and how this solves it -- is vsyscall_sysenter_return_pc even set properly for SYSCALL (or is it only set for SYSENTER)?

core/dispatch.c
JivanH: I think we should remove this comment. On AMD vsyscall page has OP_syscall and we should take care of it.
May be comparison was wrong and we should update `dcontext->asynch_target` when **syscall_method_is_syscall** (on all hardwares)?

core/dispatch.c
JivanH: The problem is in hooking.

DR hooks **__kernel_vsyscall** method, but in the case when it contains `sysenter` instruction. 
It is correct on Intel machines. But on my AMD machine ___kernel_vsyscall_ contains `syscall` instead of `sysenter` and DR do not hook it.


```
____AMD____
  <__kernel_vsyscall>          push   %ecx
  <__kernel_vsyscall+1>        push   %edx
  <__kernel_vsyscall+2>        push   %ebp 
  <__kernel_vsyscall+3>        mov    %ecx,%ebp
  <__kernel_vsyscall+5>        syscall
  <__kernel_vsyscall+7>        int    $0x80
  <__kernel_vsyscall+9>        pop    %ebp
  <__kernel_vsyscall+10>       pop    %edx
  <__kernel_vsyscall+11>       pop    %ecx
  <__kernel_vsyscall+12>       ret


____ Intel_____
  <__kernel_vsyscall>          push   %ecx
  <__kernel_vsyscall+1>        push   %edx
  <__kernel_vsyscall+2>        push   %ebp 
  <__kernel_vsyscall+3>        mov    %esp,%ebp
  <__kernel_vsyscall+5>        sysenter
  <__kernel_vsyscall+7>        int    $0x80
  <__kernel_vsyscall+9>        pop    %ebp
  <__kernel_vsyscall+10>       pop    %edx
  <__kernel_vsyscall+11>       pop    %ecx
  <__kernel_vsyscall+12>       ret

```

https://github.com/DynamoRIO/dynamorio/blob/3d775872c261c11d4f9ca1d621981301cd929ceb/core/arch/arch.c#L3257-L3266

Hooking occurs when DR checks system call instructions. hook_vsyscall() method is under condition and can be executed only when OP is `sysenter`.
I think it is wrong and DR should hooks vsyscall page when run on AMD and OP is syscall.

SweetVishnya:
With this patch applied all our internal tests succeed.

derekbruening:
Xref #5085.
Is that related to what this is trying to fix?  There we have:
```
adjust_syscall_continuation: asynch_target => 0xf7b65549
finished handling system call

d_r_dispatch: target = 0xf7b65549

interp: start_pc = 0x00000000
```

So somewhere the PC turns into 0.

JivanH:
https://github.com/DynamoRIO/dynamorio/pull/5087
Unfortunentally your patch does not fix the problem in AMD.

core/dispatch.c
derekbruening: I suggest using logs and the debugger to understand what is going wrong, and then we can design and understand an appropriate fix.

FTR here is what happened in #5085:
```
Breakpoint 2, adjust_syscall_continuation (dcontext=0x4036b040) at /home/bruening/dr/git/src/core/dispatch.c:1817
1817	            dcontext->asynch_target = vsyscall_sysenter_return_pc;
(gdb) bt
#0  adjust_syscall_continuation (dcontext=0x4036b040) at /home/bruening/dr/git/src/core/dispatch.c:1817
#1  0xf7cb3370 in handle_post_system_call (dcontext=0x4036b040) at /home/bruening/dr/git/src/core/dispatch.c:2203
#2  0xf7caa782 in dispatch_enter_dynamorio (dcontext=0x4036b040) at /home/bruening/dr/git/src/core/dispatch.c:892
#3  0xf7ca63a8 in d_r_dispatch (dcontext=0x4036b040) at /home/bruening/dr/git/src/core/dispatch.c:160
(gdb) p get_syscall_method()
$1 = 1
(gdb) x/4i dcontext->next_tag
   0x403b5240:	int    $0x80
   0x403b5242:	mov    %eax,0x4036b05c
   0x403b5247:	mov    $0xf7ecd42c,%eax
   0x403b524c:	jmp    0x403b4f40
(gdb) x/4i dcontext->asynch_target
   0xf7c0e547:	int    $0x80
   0xf7c0e549:	pop    %ebp
   0xf7c0e54a:	pop    %edx
   0xf7c0e54b:	pop    %ecx
(gdb) p vsyscall_sysenter_return_pc
$3 = (app_pc) 0xf7c0e549 "]ZYÐ\220\220\220\215\264&"
(gdb) p vsyscall_syscall_end_pc
$4 = (app_pc) 0xf7c0e547 "̀]ZYÐ\220\220\220\215\264&"
(gdb) p vsyscall_sysenter_displaced_pc
$12 = (app_pc) 0x0
```

The vsyscall code in vdso:
```
   0xf7fcf540 <__kernel_vsyscall>:	push   %ecx
   0xf7fcf541 <__kernel_vsyscall+1>:	push   %edx
   0xf7fcf542 <__kernel_vsyscall+2>:	push   %ebp
   0xf7fcf543 <__kernel_vsyscall+3>:	mov    %esp,%ebp
   0xf7fcf545 <__kernel_vsyscall+5>:	sysenter 
   0xf7fcf547 <__kernel_vsyscall+7>:	int    $0x80
   0xf7fcf549 <__kernel_vsyscall+9>:	pop    %ebp
   0xf7fcf54a <__kernel_vsyscall+10>:	pop    %edx
   0xf7fcf54b <__kernel_vsyscall+11>:	pop    %ecx
   0xf7fcf54c <__kernel_vsyscall+12>:	ret    
```

So it has the `int` in the slot after the `sysenter`, and `adjust_syscall_continuation` is needed to skip over that for the sysenter case where the hook failed.

JivanH: Instead of using `sysenter` AMD architecture uses `syscall` (on LOL64 mode).


> The problem is in hooking.
> 
> DR hooks **__kernel_vsyscall** method, but in the case when it contains `sysenter` instruction.
> It is correct on Intel machines. But on my AMD machine ___kernel_vsyscall_ contains `syscall` instead of `sysenter` and DR do not hook it.
> 
> ```
> ____AMD____
>   <__kernel_vsyscall>          push   %ecx
>   <__kernel_vsyscall+1>        push   %edx
>   <__kernel_vsyscall+2>        push   %ebp 
>   <__kernel_vsyscall+3>        mov    %ecx,%ebp
>   <__kernel_vsyscall+5>        syscall
>   <__kernel_vsyscall+7>        int    $0x80
>   <__kernel_vsyscall+9>        pop    %ebp
>   <__kernel_vsyscall+10>       pop    %edx
>   <__kernel_vsyscall+11>       pop    %ecx
>   <__kernel_vsyscall+12>       ret
> 
> 
> ____ Intel_____
>   <__kernel_vsyscall>          push   %ecx
>   <__kernel_vsyscall+1>        push   %edx
>   <__kernel_vsyscall+2>        push   %ebp 
>   <__kernel_vsyscall+3>        mov    %esp,%ebp
>   <__kernel_vsyscall+5>        sysenter
>   <__kernel_vsyscall+7>        int    $0x80
>   <__kernel_vsyscall+9>        pop    %ebp
>   <__kernel_vsyscall+10>       pop    %edx
>   <__kernel_vsyscall+11>       pop    %ecx
>   <__kernel_vsyscall+12>       ret
> ```
> 
> 




Here is the result.

```
Breakpoint 1, adjust_syscall_continuation (dcontext=0x412ac040) at /root/dynamorio/dynamorio/core/dispatch.c:1784
1784	    bool syscall_method_is_syscall = get_syscall_method() == SYSCALL_METHOD_SYSCALL;
(gdb) disass 0xf7c11060, +20
Dump of assembler code from 0xf7c11060 to 0xf7c11074:
   0xf7c11060:	push   %ecx
   0xf7c11061:	push   %edx
   0xf7c11062:	push   %ebp
   0xf7c11063:	mov    %ecx,%ebp
   0xf7c11065:	syscall 
   0xf7c11067:	int    $0x80
   0xf7c11069:	jmp    0x4127fa42
   0xf7c1106e:	pop    %ebp
   0xf7c1106f:	pop    %edx
   0xf7c11070:	pop    %ecx
   0xf7c11071:	ret    
   0xf7c11072:	mov    $0x26,%ah
End of assembler dump.
(gdb) bt
#0  adjust_syscall_continuation (dcontext=0x412ac040) at /root/dynamorio/dynamorio/core/dispatch.c:1820
#1  0xf7cb596f in handle_post_system_call (dcontext=0x412ac040) at /root/dynamorio/dynamorio/core/dispatch.c:2209
#2  0xf7cacd55 in dispatch_enter_dynamorio (dcontext=0x412ac040) at /root/dynamorio/dynamorio/core/dispatch.c:892
#3  0xf7ca8955 in d_r_dispatch (dcontext=0x412ac040) at /root/dynamorio/dynamorio/core/dispatch.c:160
#4  0x4127efce in ?? ()
Backtrace stopped: previous frame inner to this frame (corrupt stack?)
(gdb) x/4i dcontext->next_tag
   0x412f6240:	syscall 
   0x412f6242:	mov    %eax,0x412ac05c
   0x412f6247:	mov    $0xf7ecebec,%eax
   0x412f624c:	jmp    0x412f5f40
(gdb) x/4i dcontext->asynch_target
   0xf7c11067:	int    $0x80
   0xf7c11069:	jmp    0x4127fa42
   0xf7c1106e:	pop    %ebp
   0xf7c1106f:	pop    %edx
(gdb) p vsyscall_sysenter_return_pc
$4 = (app_pc) 0xf7c11069 "\351\324\351fI]ZYô&"
(gdb) p vsyscall_syscall_end_pc
$5 = (app_pc) 0xf7c11067 "̀\351\324\351fI]ZYô&"
(gdb) p vsyscall_sysenter_displaced_pc
$6 = (app_pc) 0xf7c1106e "]ZYô&"

```

In my case PC is not 0x0, but points after `jmp`,  Have you some idea?

derekbruening: The vsyscall sequence doesn't make sense to me: how does it work natively (w/o DR)?
```
   0xf7c11065:	syscall 
   0xf7c11067:	int    $0x80
```
`syscall` stores the PC right afterward in `rcx` and `sysret` jumps there: so it would execute `syscall` and then execute `int 0x80` right afterward which can't be right.

Also, what is that `jmp    0x4127fa42`?  What is at that target address?  That also looks suspect.  What does this sequence look like natively?

JivanH:
```
  <__kernel_vsyscall>          push   %ecx
  <__kernel_vsyscall+1>        push   %edx
  <__kernel_vsyscall+2>        push   %ebp 
  <__kernel_vsyscall+3>        mov    %ecx,%ebp
  <__kernel_vsyscall+5>        syscall
  <__kernel_vsyscall+7>        int    $0x80
  <__kernel_vsyscall+9>        pop    %ebp
  <__kernel_vsyscall+10>       pop    %edx
  <__kernel_vsyscall+11>       pop    %ecx
  <__kernel_vsyscall+12>       ret
```

This is the original body of `__kernel_vsyscall` (without DR). After `syscall` execute `int $0x80` and my program works fine. So I think it is right.

`jmp 0x4127fa42` instruction inserted in hook_vsyscall() method and target address calculated through **after_do_shared_syscall_addr()**. From that address make jump to code generated by **append_fcache_return_common()**

JivanH: Please describe the meaning of mangling in DR, or give some link. The code generated after mangling is different from code in debbuger.

derekbruening: > After syscall execute int $0x80 and my program works fine. So I think it is right.

But why would it execute two system call gateways for one request?  That does not make sense to me.  Under the debugger where does control go after the `syscall`?  I find it hard to believe it would then go back into the kernel again.

derekbruening: I would guess that the kernel adds 2 to `ecx` prior to the `sysret` and it comes back *after* the `int` or something like that.

JivanH: >  Under the debugger where does control go after the syscall


I set breakpoint on `int` and `pop` instructions. Debugger stopped on `pop`.
Screenshot from 2021-09-15 00-07-43

JivanH: > But I thought DR would not try to insert a hook unless the syscall method was sysenter?

You are right, DR on master branch does not insert hook for `syscall` instraction. I added that feature, only on AMD and LOL64 DR insert hook.

Maybe we little bit misunderstood each other :-)

JivanH: > I would guess that the kernel adds 2 to ecx prior to the sysret and it comes back after the int or something like that.

I agree with you, but what is meaning to generate code which will not be executed at all?

derekbruening: Given this behavior where `syscall` does *not* return to the subsequent PC, we need to understand:

+ Does this only happen for the `syscall` in the VDSO or other `syscall` instances?  Are there instances of `syscall` inlined into libc or elsewhere?  What do they do?
+ What happens with DR generating its own `syscall` instructions?  Does the kernel add 2 to the PC, or does it set the PC to the absolute PC of the VDSO post-`syscall`-post-`int` location?

If the kernel is setting the absolute return PC of every `syscall` for 32-bit AMD, then it's behaving like `sysenter` and indeed DR needs treat it like `sysenter`: it needs to hook the vsyscall code and it has to be prepared for control to jump after any `syscall` it generates.

JivanH: I wrote assembler code with `syscall` instruction after execution specified system function kernel went to *__kernel_vsyscall* and my program crashed. 

The same behavior in DR. DR inserts `syscall` instruction through _handle_system_call()_ method. After execution `syscall` kernel goes to *__kernel_vsyscall* and cannot go back for dispatch and DR crushes.
Through hook, I add `jmp` to the instruction after `syscall` inserted in handle time, which guarantees the correct routine.

derekbruening: Thank you for investigating so we can better understand this.  Given this behavior, where the kernel turns `syscall` from returning to the subsequent PC into an abnormal control flow sequence like `sysenter`, then I agree that DR needs to hook the vsyscall sequence and needs to treat `syscall` like `sysenter` for this AMD 32-bit case.

JivanH: Nice.
Should I make some changes in my patch? Is it already fit to merge?

core/arch/arch.c
```diff
@@ -3254,7 +3254,10 @@ check_syscall_method(dcontext_t *dcontext, instr_t *instr)
         ASSERT(get_syscall_method() == SYSCALL_METHOD_UNINITIALIZED ||
                get_syscall_method() == SYSCALL_METHOD_INT);
 #    ifdef LINUX
-        if (new_method == SYSCALL_METHOD_SYSENTER) {
+        if (new_method ==
+            SYSCALL_METHOD_SYSENTER IF_X86_32(||
+                                              (new_method == SYSCALL_METHOD_SYSCALL &&
```
derekbruening, line 3259: Please add a comment explaining as this looks like an error w/o knowing about this weird kernel behavior.

core/arch/arch.c
```diff
@@ -3254,7 +3254,10 @@ check_syscall_method(dcontext_t *dcontext, instr_t *instr)
...
+                                               cpu_info.vendor == VENDOR_AMD))) {
```
derekbruening, line 3260: Please update other places that assume syscall behaves "normally": like does_syscall_ret_to_callsite().  Not sure if there are others.

Do DR's own syscalls, both the static ones in drlibc_x86.asm (are they still `int 0x80` here?) and generated (is it again using `int 0x80` and never `syscall`??  Please double-check) work??  I could imagine lots of problems with the weird `syscall` behavior.

core/dispatch.c
derekbruening: OK, so you are removing this comment, but please add a new one before line 1818 below to explain the AMD check there.

core/arch/arch.c
JivanH: > Please update other places that assume syscall behaves "normally": like does_syscall_ret_to_callsite(). Not sure if there are others.

I am sorry but do not understand, what should update and where?

> Do DR's own syscalls, both the static ones in drlibc_x86.asm 

In drlibc_x86.asm file, there is no case for 32-bit where used `syscall` instruction.
Here is a fragment from that static code.
Screenshot from 2021-09-18 00-17-38

core/arch/arch.c
JivanH: Added

core/arch/arch.c
derekbruening: > > Please update other places that assume syscall behaves "normally": like does_syscall_ret_to_callsite(). Not sure if there are others.
> 
> I am sorry but do not understand, what should update and where?

As mentioned, the `does_syscall_ret_to_callsite()` function needs updating, since this `syscall` certainly does not return.

JivanH: Updated

core/arch/arch.c
```diff
@@ -3254,7 +3254,14 @@ check_syscall_method(dcontext_t *dcontext, instr_t *instr)
         ASSERT(get_syscall_method() == SYSCALL_METHOD_UNINITIALIZED ||
                get_syscall_method() == SYSCALL_METHOD_INT);
 #    ifdef LINUX
-        if (new_method == SYSCALL_METHOD_SYSENTER) {
+
+        /* PR 5037: 32-bit apps on AMD machines have OP_syscall instruction in
```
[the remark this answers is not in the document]
JivanH: I will take this into account for the next collaboration.

It is strange, my repo is public.
Maybe the button grayed out because I not synced with upstream. Already I have done it. Please try again.

If the button will still inactive then you can send me the text and I will change it.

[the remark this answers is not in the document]
derekbruening: "Edit file" is still grayed out.  Please change this comment to something like:

```C
/* i#4407: An OP_syscall instruction on 32-bit AMD returns to a hardcoded vsyscall PC no
 * matter where it is.  Thus we must hook the vsyscall just like we do for OP_sysenter.
 */
```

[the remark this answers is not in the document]
JivanH: Changed

derekbruening (approved):

core/arch/arch.c
```diff
@@ -3273,7 +3273,14 @@ check_syscall_method(dcontext_t *dcontext, instr_t *instr)
         ASSERT(get_syscall_method() == SYSCALL_METHOD_UNINITIALIZED ||
                get_syscall_method() == SYSCALL_METHOD_INT);
 #    ifdef LINUX
-        if (new_method == SYSCALL_METHOD_SYSENTER) {
+
+        /* i#4407: An OP_syscall instruction on 32-bit AMD returns to a hardcoded vsyscall PC no
+         * matter where it is. Thus we must hook the vsyscall just like we do for OP_sysenter.
```
derekbruening, line 3278: Looks like it's failing clang-format b/c the line's too long so it has to wrap it:
```
-        /* i#4407: An OP_syscall instruction on 32-bit AMD returns to a hardcoded vsyscall PC no
-         * matter where it is. Thus we must hook the vsyscall just like we do for OP_sysenter.
+        /* i#4407: An OP_syscall instruction on 32-bit AMD returns to a hardcoded vsyscall
+         * PC no matter where it is. Thus we must hook the vsyscall just like we do for
+         * OP_sysenter.
```

JivanH: Fixed

derekbruening:
Hmm is that a new Jenkins flake?  Filed #5120.

derekbruening:
run arm tests

derekbruening:
Merged.  Thank you for the PR and for following it through to finalization.

JivanH:
> Merged. Thank you for the PR and for following it through to finalization.

Excellent. I think our collaboration will be continued.

Issue #5009, opened 2021-07-14
Determine coverage of DR test suites


abhinav92003:
It'd be good to determine coverage of our test suites. This may help in guiding efforts for improving stability of DR.

We can use tools like https://github.com/linux-test-project/lcov, or even drcov, though there may be issues in running drcov on DR itself.

abhinav92003:
I did something similar to https://github.com/codecov/example-cpp11-cmake/blob/master/CMakeLists.txt and added `--coverage` to link and build flags in [dynamorio/CMakeLists.txt](https://github.com/DynamoRIO/dynamorio/blob/master/CMakeLists.txt). That gives me:
```
CMake Error at /usr/local/google/home/sharmaabhinav/dr/src/i5009-1/core/CMake_readelf.cmake:100 (message):
  *** Error:
  /usr/local/google/home/sharmaabhinav/dr/build/i5009-1/lib64/debug/libdynamorio.so
  has undefined symbol: GLOBAL DEFAULT UND fdopen



make[2]: *** [core/CMakeFiles/dynamorio.dir/build.make:1319: lib64/debug/libdynamorio.so] Error 1
make[2]: *** Deleting file 'lib64/debug/libdynamorio.so'
make[1]: *** [CMakeFiles/Makefile2:944: core/CMakeFiles/dynamorio.dir/all] Error 2
```
If I disable the dependency checks by setting `check_deps` OFF, it builds DR (didn't include tests yet though) successfully, but even plain DR doesn't work:
```
$ ./bin64/drrun -- ls
ERROR: failed to relocate DynamoRIO!
Please file an issue at http://dynamorio.org/issues.
```

abhinav92003:
Something to think about later: we can integrate with https://codecov.io/ for coverage reports with CI runs. They seem to have some interesting free features for public open source repos.

derekbruening:
libdynamorio can't have any dependencies: looks like you're pulling in libc among other things.  Maybe try static DR for a prototype, but if the libc operations are not all at init/exit time there would still be trouble from lack of isolation from the app.

PR #5040, opened 2021-08-09
i#5009: Functionality for counting compulsory misses

mat-kalinowski:
This patch contains implementation of the functionality for counting
compulsory misses as a separate metric.

In this implementation information about all the already accessed memory
regions are stored. If the cache miss happens and block which given address
refers to was never accessed we are counting it as a compulsory miss.

Issue: #5009


mat-kalinowski:
run arm tests

derekbruening:
Thank you for the PR -- I will try to get to it tomorrow

mat-kalinowski:
Sure, thanks

clients/drcachesim/simulator/caching_device_stats.h
```diff
@@ -64,9 +67,88 @@ enum class metric_name_t {
...
+    {
+        if (!IS_POWER_OF_2(block_size)) {
+            ERRMSG("Block size should be a power of 2.");
+            return;
+        }
+        int block_size_bits = compute_log2(block_size);
+        block_size_mask_ = ~((1 << block_size_bits) - 1);
+    }
+
+    // Takes non-alligned address and inserts bound consisting of the nearest multiples
```
derekbruening, line 88: spelling: "alligned"

clients/drcachesim/simulator/caching_device_stats.h
```diff
@@ -64,9 +67,88 @@ enum class metric_name_t {
...
+            } else if (addr >= bound.beg && addr < bound.end) {
+                return std::make_pair(true, i);
+            }
+            i++;
+        }
+        return std::make_pair(false, bounds.size());
+    }
+
+private:
+    std::vector<bound> bounds;
```
derekbruening, line 144: So this is an interval tree, right?  I would suggest storing it in a tree which should perform much better (as it is it looks like it will perform a whole lot of vector shifts).  You should be able to use std::map (esp std::map::upper_bound).

If this is changed to a tree does the overhead still show up enough for this to be off by default?

clients/drcachesim/simulator/caching_device_stats.cpp
```diff
@@ -69,6 +73,7 @@ caching_device_stats_t::caching_device_stats_t(const std::string &miss_file,
     stats_map_.emplace(metric_name_t::MISSES, num_misses_);
     stats_map_.emplace(metric_name_t::HITS_AT_RESET, num_hits_at_reset_);
     stats_map_.emplace(metric_name_t::MISSES_AT_RESET, num_misses_at_reset_);
+    stats_map_.emplace(metric_name_t::COMPULSORY_MISSES, num_compulsory_misses_);
```
derekbruening, line 76: How about also printing this out as a sub-category of misses in the cache simulator?

derekbruening: Please add a test: augment the unit tests, and/or add an end-to-end test w/ golden output.

clients/drcachesim/simulator/caching_device_stats.h
mat-kalinowski: I changed implementation to use std::map, it is way more effective indeed. The overhead is small enough that I turned this on by default. Thanks for the good suggestion!

clients/drcachesim/simulator/caching_device_stats.cpp
mat-kalinowski: I have added suitable print and unit test.

clients/drcachesim/simulator/caching_device_stats.h
derekbruening: (The convention is to reply "Done" or somesuch and mark as Resolved)

mat-kalinowski: Sure, I will remember

mat-kalinowski:
I see that some tests are failing, because expected output changed. We have additional metric printed and some .templatex files don't have this metric included in the regex. Is there an easy way instead of manually changing all the files ? I see that mentioned files under clients/drcachesim/tests need adjustments, are there some other places ? Thanks

derekbruening:
> I see that some tests are failing, because expected output changed. We have additional metric printed and some .templatex files don't have this metric included in the regex. Is there an easy way instead of manually changing all the files ?

Unfortunately there is no automated script -- you would have to script it yourself with sed or whatnot.

> I see that mentioned files under clients/drcachesim/tests need adjustments, are there some other places ? Thanks

No it should only be the expected output files in that directory.

clients/drcachesim/simulator/caching_device_stats.h
```diff
@@ -99,49 +98,55 @@ class access_count_t {
...
+            return std::make_pair(true, prev_it.base());
+        } else {
+            return std::make_pair(false, next_it);
         }
-        return std::make_pair(false, bounds.size());
     }
 
 private:
-    std::vector<bound> bounds;
+    // Bounds are members of the the std::map. The beginning of the bound is stored
```
derekbruening, line 147: nit: "the the"

suite/tests/client-interface/annotation-concurrency.appdll.c
```diff
@@ -59,7 +59,6 @@ EXPORT void
 jacobi_init(int matrix_size, bool enable_annotations)
 {
     invoke_annotations = enable_annotations;
-
```
derekbruening, line 62: Best to revert this to avoid superfluous unrelated noise in the diff.

derekbruening (approved):

clients/drcachesim/simulator/caching_device_stats.h
mat-kalinowski: I have fixed the comment

suite/tests/client-interface/annotation-concurrency.appdll.c
mat-kalinowski: Right, I didn't notice that, thanks. I reverted this change.

Issue #2974, opened 2018-04-29
add trace support for AArch64


fhahn:
Similar to #1668 for AArch64.  There is a bunch of x86-specific trace building code that needs to be ported.

fhahn:
@egrimley  shared a draft patch at #2442

Vincent-lau:
Hi I am working on a project to add trace support for AArch64, I am wondering if you have any suggestions on what would be a good starting point?

AssadHashmi:
> Hi I am working on a project to add trace support for AArch64, I am wondering if you have any suggestions on what would be a good starting point?

Have a look at the draft patch https://github.com/DynamoRIO/dynamorio/pull/2442 and see if it can be updated to the latest master.

Vincent-lau:
Hi, I am planning to submit a patch to add trace support for AArch64, but I am not yet a member of DynamoRIO, shall I submit a few small patches by forking the repo before asking to be added as the member?

AssadHashmi:
> Hi, I am planning to submit a patch to add trace support for AArch64, but I am not yet a member of DynamoRIO, shall I submit a few small patches by forking the repo before asking to be added as the member?

There's no need to fork, I've added you as a contributor. You should receive an invite message from GitHub.
Please follow the advice on contributing here https://dynamorio.org/page_code_reviews.html
Thanks for contributing, we look forward to your pull request.

derekbruening:
Documenting a proposal that eliminates the internal control flow of fixup sequences, which break the linear assumptions of translation:

Pasting from https://github.com/DynamoRIO/dynamorio/pull/5045/files/b9ada873ecc0b86128169aae0bd7ad648c0f921b..fc70e80bad8a66b7f417c27673fa5659e96ff1ef#r697566139

One proposal is to change the i.b. comparison to use x2 instead of x0 as the temp reg holding the expected target immediate. Then the only thing missing is copying from the actual target reg into x2, and maybe that could be done locally. It would use a taken forward branch on the match which maybe is not ideal but certainly simple:

```
     str x2, TLS_REG2_SLOT
     mov x2, #trace_next_target
     eor x2, x2, jump_target
     cbz x2, match
     mov x2, jump_target
     b ibl_routine_regular_exit_stub
match:
     ldr x2, TLS_REG2_SLOT
```

Then we eliminate this special-case trace_exit_label.

[...]

PR #2442, opened 2017-05-26
i#1569 AArch64: Implement trace optimisation.

egrimley:
The optimisation is not yet enabled by default, though some programs
run faster, and correctly, with "-enable_traces".

This is based on work done by Kevin Zhou in July 2016.

Change-Id: I6ab3bf0a2f48b25002f1586ff1b85e6232b4357d

Issue #4680, opened 2021-01-20
Port -stress_recreate_state to AArchXX


derekbruening:
The -stress_recreate_state and decode-stress tests are useful validators of translation code.  This issue covers porting them to AArchXX.

derekbruening:
1st problem:
```
derek@tx1:~/dr/build$ make -j dynamorio && bin64/drrun -stress_recreate_state -- suite/tests/bin/simple_app
<Application /home/derek/dr/build/suite/tests/bin/simple_app (5413).  Internal Error: DynamoRIO debug check failure: /home/derek/dr/src/core/translate.c:984 instr_is_meta(inst)
(Error occurred @963 frags)
```

Looks like the mangling of a cbr that uses the stolen reg loses its xl8.
At -loglevel 4 we hit an earlier, similar assert:
```
bb ilist before mangling:
TAG  0x0000ffff7ef45770
 +0    L3 @0x0000fffd3af97ea8  3500007c   cbnz   $0x0000ffff7ef4577c %w28
 +4    L4 @0x0000fffd3af97dd8  14000000   b      $0x0000ffff7ef45774
END 0x0000ffff7ef45770

bb ilist after mangling:
TAG  0x0000ffff7ef45770
 +0    m4 @0x0000fffd3af9a180  f9000380   str    %x0 -> (%x28)[8byte]
 +4    m4 @0x0000fffd3af97e40  f9401b80   ldr    +0x30(%x28)[8byte] -> %x0
 +8    m4 @0x0000fffd3af98128  34000000   cbz    @0x0000fffd3af992d8[8byte] %w0
 +12   m4 @0x0000fffd3af997e0  f9400380   ldr    (%x28)[8byte] -> %x0
 +16   L4 @0x0000fffd3af97ea8  14000000   b      $0x0000ffff7ef4577c
 +20   m4 @0x0000fffd3af992d8  14000000   <label>
 +20   m4 @0x0000fffd3af99190  f9400380   ldr    (%x28)[8byte] -> %x0
 +24   L4 @0x0000fffd3af97dd8  14000000   b      $0x0000ffff7ef45774
END 0x0000ffff7ef45770


done building bb instrlist *********************

vm_area_remove_fragment: entry 0x0000fffd3af444d0
ilist for recreation:
TAG  0x0000ffff7ef45770
 +0    m4 @0x0000fffd3af9a180  f9000380   str    %x0 -> (%x28)[8byte]
[...]

PR #5045, opened 2021-08-11
i#2974 trace support for AArch64, part 2: added trace implementation

Vincent-lau:
This patch incorporated changes from patch #2442, and fixed some corner cases
where the assumption was incorrect and caused the program to crash.

Trace support is not yet enabled by default, but can be enabled with "-enable_traces".

A large part of this work was based on work done by Kevin Zhou. Please see branch
i1569-trace and patch #2442 for the original version.

Issues: #1569, #2974


Vincent-lau:
The results are 
```
debug-internal-64: 242 tests passed, **** 5 tests failed, of which 1 were flaky, but ignoring 4 for i#2417: ****
	(ignore: i#2417) 	code_api|linux.fib-conflict 
	(ignore: i#2417) 	code_api|linux.fib-conflict-early 
	(ignore: i#2417) 	code_api|tool.drcachesim.phys 
	(ignore: i#2417) 	code_api|api.rseq 
	code_api|api.static_sideline_FLAKY 

====> FAILURE in release-external-64 <====
release-external-64: 48 tests passed, **** 3 tests failed, but ignoring 2 for i#2417: ****
	(ignore: i#2417) 	code_api|tool.drcachesim.phys 
	(ignore: i#2417) 	code_api|tool.drcacheoff.rseq 
	code_api|tool.drcachesim.coherence 
```
which I believe are flaky tests by #2417

AssadHashmi:
> which I believe are flaky tests by #2417

Yes AIUI. We need to add that to the `FLAKY_` list.

AssadHashmi:
run arm tests

core/arch/interp.c
```diff
@@ -5771,6 +5771,189 @@ mangle_x64_ib_in_trace(dcontext_t *dcontext, instrlist_t *trace, instr_t *target
...
+#    define STR_IMM_ENC 0xf9000000
+#    define SAVE_IBL_REG_ENC                                                       \
+        ((STR_IMM_ENC | 16 >> 3 << 10) | (dr_reg_stolen - DR_REG_START_GPR) << 5 | \
+         (IBL_TARGET_REG - DR_REG_START_GPR))
+#    define MOV_IBL_REG_ENC_MASK 0xff00ffff
+#    define MOV_IBL_REG_MASK 0x001f0000
+#    define MOV_IBL_REG_ENC (0xaa0003e0 | (IBL_TARGET_REG - DR_REG_START_GPR))
+
+static reg_id_t
+check_and_remove_patched_ibl(dcontext_t *dcontext, instrlist_t *trace, instr_t *targeter)
```
abhinav92003, line 5796: `dcontext` seems unused here. Do we need it?

core/arch/interp.c
```diff
@@ -5771,6 +5771,189 @@ mangle_x64_ib_in_trace(dcontext_t *dcontext, instrlist_t *trace, instr_t *target
...
+     */
+    instr_t *prev = instr_get_prev(targeter);
+    if (!instr_is_load_tls(prev))
+        return prev;
+    instr_t *prev_prev = instr_get_prev(prev);
+    if (!instr_is_cbr_stolen(prev_prev))
+        return prev;
+    // dr_printf("==============now we confirm ========================\n");
+    // monitor_data_t *md = (monitor_data_t *)dcontext->monitor_field;
+    // instr_disassemble(dcontext, targeter, STDOUT);
```
abhinav92003, line 5935: Do we need this commented code? If not, lets remove it.

core/arch/interp.c
```diff
@@ -5781,11 +5964,18 @@ mangle_indirect_branch_in_trace(dcontext_t *dcontext, instrlist_t *trace,
...
-#ifdef X86
+
     instr_t *next = instr_get_next(targeter);
     /* all indirect branches should be ubrs */
     ASSERT(instr_is_ubr(targeter));
     /* expecting basic blocks only */
+    if (!((end_instr != NULL && targeter == end_instr) ||
+          targeter == instrlist_last(trace))) {
+        // dr_printf("end_instr!=NULL %d, targeter==end_instr %d, "
+        //           "targeter==instrlis_last(trace) %d\n",
```
abhinav92003, line 5975: Same here.

core/arch/interp.c
derekbruening: Commented-out code is forbidden in the style guide: https://dynamorio.org/page_code_style.html#autotoc_md125 item 4

core/arch/interp.c
Vincent-lau: Nope, will remove it

core/arch/interp.c
Vincent-lau: done

core/arch/interp.c
Vincent-lau: done

abhinav92003:
Thanks for making the changes. Will take a look today

Vincent-lau:
> Thanks for making the changes. Will take a look today

I just noticed `dcontext` in `check_patch()` is also redundant, I will change it along with any other changes that you might propose.

core/arch/interp.c
```diff
@@ -5771,6 +5771,183 @@ mangle_x64_ib_in_trace(dcontext_t *dcontext, instrlist_t *trace, instr_t *target
...
+check_and_remove_patched_ibl(instrlist_t *trace, instr_t *targeter)
+{
+    int i, j;
+    instr_t *prev = instr_get_prev(targeter);
+    /* XXX: Currently we assume the previous instruction is undecoded. */
+    if (!instr_raw_bits_valid(prev)) {
+        ASSERT_CURIOSITY(false);
+        return DR_REG_NULL;
+    }
+    uint num_instr = prev->length / AARCH64_INSTR_SIZE;
```
abhinav92003, line 5805: I may be missing something here, but isn't prev->length always `AARCH64_INSTR_SIZE` (unless its a label or ldstex)?

core/arch/interp.c
```diff
@@ -5975,6 +6151,88 @@ mangle_indirect_branch_in_trace(dcontext_t *dcontext, instrlist_t *trace,
...
+     *      str x0, TLS_REG0_SLOT
+     *      mov x0, #trace_next_target
+     *      eor x0, x0, jump_target
+     *      cbnz x0, trace_exit (ibl routine)
+     *      ldr x0, TLS_REG0_SLOT
+     */
+
+    /* Check and remove previous two patched instructions. */
+    has_str_ldr = false;
+    jump_target_reg = check_and_remove_patched_ibl(trace, targeter);
```
abhinav92003, line 6175: Maybe rename `check_and_remove_patched_ibl` and `check_patched_ibl` to describe what they do.

core/arch/aarch64/emit_utils.c
```diff
@@ -446,8 +446,14 @@ indirect_linkstub_stub_pc(dcontext_t *dcontext, fragment_t *f, linkstub_t *l)
     cache_pc cti = EXIT_CTI_PC(f, l);
```
abhinav92003, line 446: Can you check if any test in the current suite hits all the new cases introduced by this PR? Also add details of the manual testing you did in the PR's description.

abhinav92003 (commented):
Mid review comments.

core/arch/aarch64/emit_utils.c
Vincent-lau: I haven't checked the test suite yet, but will do.

The manual test I did was to run spec cpu2017 under DR and currently if is able to run all of the benchmarks that can be run without trace enabled on AARCH64.

core/arch/emit_utils_shared.c
```diff
@@ -1294,6 +1294,11 @@ update_indirect_exit_stub(dcontext_t *dcontext, fragment_t *f, linkstub_t *l)
 int
 fragment_prefix_size(uint flags)
 {
+#ifdef AARCH64
+    /* For AArch64, there is no need to save the flags
+     * so we always have the same ibt prefix. */
+    return fragment_ibt_prefix_size(flags);
+#endif
```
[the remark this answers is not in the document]
Vincent-lau: done

core/arch/interp.c
```diff
@@ -5975,6 +6151,88 @@ mangle_indirect_branch_in_trace(dcontext_t *dcontext, instrlist_t *trace,
...
+     * based on indirect target not being equal to next block in trace.
+     * Original ibl lookup:
+     *      str tgt_reg, TLS_REG2_SLOT
+     *      mov tgt_reg, jump_target
+     *      b ibl_routine
+     * Now we rewrite it into:
+     *      str x0, TLS_REG0_SLOT
+     *      mov x0, #trace_next_target
+     *      eor x0, x0, jump_target
+     *      cbnz x0, trace_exit (ibl routine)
```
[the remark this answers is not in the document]
Vincent-lau: You are right, and this is done in `fixup_indirect_trace_exit()` function which is at the bottom of interp.c

core/arch/interp.c
```diff
@@ -5771,6 +5771,183 @@ mangle_x64_ib_in_trace(dcontext_t *dcontext, instrlist_t *trace, instr_t *target
...
+            /* Return jump target register id. */
+            return ((prev_enc & MOV_IBL_REG_MASK) >> 16) + DR_REG_START_GPR;
+        }
+    }
+    return DR_REG_NULL;
+}
+
+/*
+ * checking for the special case where there isn't a str/mov being patched
+ * rather there is a ldr/mov which could happen when for example, the following
```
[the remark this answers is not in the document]
Vincent-lau: done

core/arch/interp.c
```diff
@@ -5975,6 +6151,88 @@ mangle_indirect_branch_in_trace(dcontext_t *dcontext, instrlist_t *trace,
...
+    }
+    LOG(THREAD, LOG_MONITOR, 4, "fixup_last_cti: jump target reg is %s\n",
+        reg_names[jump_target_reg]);
+
+    /* Choose any scratch register except the target reg. */
+    scratch = (jump_target_reg == DR_REG_X0) ? DR_REG_X1 : DR_REG_X0;
+    /* str scratch, TLS_REG0_SLOT */
+    added_size +=
+        tracelist_add(dcontext, trace, next,
+                      instr_create_save_to_tls(dcontext, scratch, TLS_REG0_SLOT));
```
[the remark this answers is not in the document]
Vincent-lau: done

core/arch/interp.c
```diff
@@ -5975,6 +6151,88 @@ mangle_indirect_branch_in_trace(dcontext_t *dcontext, instrlist_t *trace,
...
+    /* If we do stay on the trace, restore x0 and x1. */
+    /* ldr scratch, TLS_REG0_SLOT */
+    added_size +=
+        tracelist_add(dcontext, trace, next,
+                      instr_create_restore_from_tls(dcontext, scratch, TLS_REG0_SLOT));
+    if (has_str_ldr) {
+        /* we need to restore app's x2 in case the branch to trace_exit is not taken */
+        added_size += tracelist_add(
+            dcontext, trace, next,
+            instr_create_restore_from_tls(dcontext, DR_REG_X2, TLS_REG2_SLOT));
```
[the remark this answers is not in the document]
Vincent-lau: done

core/arch/interp.c
```diff
@@ -5975,6 +6151,88 @@ mangle_indirect_branch_in_trace(dcontext_t *dcontext, instrlist_t *trace,
...
+        INSTR_CREATE_cbnz(dcontext, instr_get_target(targeter), opnd_create_reg(scratch));
+    /* Set the exit type from the targeter. */
+    instr_exit_branch_set_type(instr, instr_exit_branch_type(targeter));
+    added_size += tracelist_add(dcontext, trace, next, instr);
+    /* If we do stay on the trace, restore x0 and x1. */
+    /* ldr scratch, TLS_REG0_SLOT */
+    added_size +=
+        tracelist_add(dcontext, trace, next,
+                      instr_create_restore_from_tls(dcontext, scratch, TLS_REG0_SLOT));
+    if (has_str_ldr) {
```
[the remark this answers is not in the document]
Vincent-lau: done

core/arch/interp.c
Vincent-lau: A L0 instruction can contain multiple instructions, I believe?

core/arch/interp.c
Vincent-lau: done

core/arch/aarch64/emit_utils.c
Vincent-lau: I had a brief look at the test suite, and I haven't found any places where the newly introduced code would be explicitly tested (thinking about how a fragment would be mangled into another one with `str/mov` removed). I haven't found any code that tested `mangle_indirect_branch_in_trace()` for x86 either. (but maybe I missed something, if any one knows where the test for x86 trace building is, I would love to know)

There is one test that tested whether a trace is built if some bbs are executed frequently enough, in client.null_instrument.

I am wondering what would be the best way to proceed, wrote tests in assembly that verify the fragment mangling?

core/arch/interp.c
```diff
@@ -8007,3 +8252,157 @@ d_r_emulate(dcontext_t *dcontext, app_pc pc, priv_mcontext_t *mc)
...
+ * before we actually jump to the ibl routine.
+ * We want the indirect hit path to have minimum instructions
+ * and also conform to the protocol of ibl routine
+ * Therefore we append the restore at the end of the trace
+ * after the backward jump to trace head.
+ * For example, the code will be fixed to:
+ *      eor x0, x0, jump_target
+ *      cbnz x0, trace_exit_label
+ *      ...
+ *      b trace_head
```
abhinav92003, line 8280: You mean branch to the next bb in the trace right?

core/arch/interp.c
```diff
@@ -8007,3 +8252,157 @@ d_r_emulate(dcontext_t *dcontext, app_pc pc, priv_mcontext_t *mc)
...
+                prev = instr_get_prev(instr);
+                if (prev->opcode != OP_eor)
+                    return 0;
+
+                ASSERT(instr_num_srcs(prev) == 4 && opnd_is_reg(instr_get_src(prev, 1)));
+                jump_target_reg = opnd_get_reg(instr_get_src(prev, 1));
+
+                if (!ind_target || jump_target_reg == DR_REG_NULL)
+                    return 0;
+                /* Choose any scratch register except the target reg. */
```
abhinav92003, line 8340: So we have as many copies of the `trace_exit:` code as there are indirect branches in the trace? And we need so many copies because we may need different scratch regs for each?

core/arch/interp.c
```diff
@@ -8007,3 +8252,157 @@ d_r_emulate(dcontext_t *dcontext, app_pc pc, priv_mcontext_t *mc)
...
+                /* ldr x0, TLS_REG0_SLOT */
+                instrlist_append(
+                    trace,
+                    instr_create_restore_from_tls(dcontext, scratch, TLS_REG0_SLOT));
+                added_size += AARCH64_INSTR_SIZE;
+                /* if x2 alerady contains the jump_target, then there is no need to store
+                 * it away and load value of jump target into it
+                 */
+                if (jump_target_reg != IBL_TARGET_REG) {
+                    /* str x2, TLS_REG2_SLOT */
```
abhinav92003, line 8353: Better to use the named constant `IBL_TARGET_REG` for x2.

core/arch/aarch64/emit_utils.c
abhinav92003: I couldn't find much testing for traces in x86 either. I think it'll be good to add unit tests though because I see that the implementation sometimes depends on how other mangling (like `mangle_cbr_stolen_reg`) looks like.

> I am wondering what would be the best way to proceed, wrote tests in assembly that verify the fragment mangling?

Yeah, we do have tests that create basic blocks in assembly. E.g. see `drreg-test.c` (we don't need a separate test client for this PR though).

@derekbruening any other ideas? I don't know how we can verify in unit tests if DR indeed created a trace or not. I can only think of the "Trace fragments generated" stats. Do you think it's worth the effort adding unit tests for traces (asking because I couldn't find much testing for x86).

abhinav92003:
Thanks for making the previous changes. In most of my recent comments, I'm just trying to understand some parts of trace building that I'm not familiar with yet, and maybe help improve inline documentation along the way.

core/arch/aarch64/emit_utils.c
abhinav92003: I had an offline discussion with Derek, and it seems that currently we rely mostly on traces being built in the existing tests. This is of course not ideal as it would allow detection of only those regressions that cause crashes. We do want comprehensive tests for trace building that verify corner cases as well, but that's a lot of more work. I've created #5057 for that. For this PR, I think we should document trace creation related stats (like count of traces created and others if they are readily available in DR stats) for SPEC and our current test suite.

core/arch/interp.c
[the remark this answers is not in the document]
Vincent-lau: I get what you mean, but this line of code says `instr =INSTR_CREATE_cbnz(dcontext, instr_get_target(targeter), opnd_create_reg(scratch));` is setting the `trace_exit` to be the same as the target of the `targeter` instruction, which is the `ibl_routine` I suppose? And this `trace_exit` later gets changed into `trace_exit_stub` which is what you said (setting up preconditions and invoke `ibl_routine`). I am happy to change this later on if, maybe in the next commit, if you think this is indeed confusing.

core/arch/interp.c
```diff
@@ -8007,3 +8252,157 @@ d_r_emulate(dcontext_t *dcontext, app_pc pc, priv_mcontext_t *mc)
...
+                /* Modify the target of the cbnz. */
+                instr_set_target(instr, opnd_create_instr(trace_exit_label));
+                indirect_type = instr_exit_branch_type(instr);
+                /* unset exit type */
+                instr->flags &= ~EXIT_CTI_TYPES;
+                instr_set_our_mangling(instr, true);
+
+                /* Retrieve jump target reg from the xor instruction. */
+                prev = instr_get_prev(instr);
+                if (prev->opcode != OP_eor)
```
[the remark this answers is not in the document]
Vincent-lau: I think it should always be an `eor`, have added an `ASSERT_CURIOSITY`

core/arch/interp.c
```diff
@@ -8007,3 +8252,157 @@ d_r_emulate(dcontext_t *dcontext, app_pc pc, priv_mcontext_t *mc)
...
+ *      ...
+ *      b trace_head
+ * trace_exit_label:
+ *      ldr x0, TLS_REG0_SLOT
+ *      str x2, TLS_REG2_SLOT
+ *      mov x2, jump_target
+ *      b ibl_routine
+ */
+int
+fixup_indirect_trace_exit(dcontext_t *dcontext, instrlist_t *trace)
```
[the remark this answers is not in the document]
Vincent-lau: I didn't write this code myself (copied from patch #2442), but I think this is invoked before we emit the trace into code cache. It walks through the trace and mixup all the `trace_exit` to have the preconditions set up.

core/arch/interp.c
```diff
@@ -8007,3 +8252,157 @@ d_r_emulate(dcontext_t *dcontext, app_pc pc, priv_mcontext_t *mc)
...
+                    added_size += AARCH64_INSTR_SIZE;
+                }
+                /* b ibl_target */
+                branch = XINST_CREATE_jump(dcontext, opnd_create_pc(ind_target));
+                instr_exit_branch_set_type(branch, indirect_type);
+                instrlist_append(trace, branch);
+                added_size += AARCH64_INSTR_SIZE;
+            }
+            /* Don't invoke the decoder; only mangled instruction reached here. */
+        } else if (instr->opcode == OP_cbz || instr->opcode == OP_cbnz ||
```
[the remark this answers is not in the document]
Vincent-lau: I think you are right, the second condition is for `cti` that is not an `exit_cti` which here means it is a meta instruction, and is handling the case stolen reg case. Will add a comment.

core/arch/interp.c
```diff
@@ -8007,3 +8252,157 @@ d_r_emulate(dcontext_t *dcontext, app_pc pc, priv_mcontext_t *mc)
...
+                   instr->opcode == OP_tbz || instr->opcode == OP_tbnz) {
+            /* Next instruction should be a LDR_TLS_slot. */
+            instr_t *next = instr_get_next(instr);
+            if (!instr_is_load_tls(next))
```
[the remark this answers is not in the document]
Vincent-lau: I think this is fine, but your suggestion also works.

core/monitor.c
```diff
@@ -1205,6 +1205,7 @@ end_and_emit_trace(dcontext_t *dcontext, fragment_t *cur_f)
         target = opnd_get_pc(instr_get_target(last));
         md->emitted_size -= local_exit_stub_size(dcontext, target, md->trace_flags);
     }
+    IF_AARCH64(md->emitted_size += fixup_indirect_trace_exit(dcontext, trace));
```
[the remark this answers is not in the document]
Vincent-lau: Think so, before we emit the fragment.

core/arch/interp.c
```diff
@@ -8007,3 +8252,157 @@ d_r_emulate(dcontext_t *dcontext, app_pc pc, priv_mcontext_t *mc)
...
+    instr_t *trace_end = instrlist_last(trace);
+
+    LOG(THREAD, LOG_MONITOR, 4, "fixup the indirect trace exit\n");
+
+    for (instr = instrlist_first(trace); instr != trace_end;
+         instr = instr_get_next(instr)) {
+        if (instr_is_exit_cti(instr)) {
+            target = instr_get_branch_target_pc(instr);
+            /* Check for indirect exit. */
+            if (is_indirect_branch_lookup_routine(dcontext, (cache_pc)target)) {
```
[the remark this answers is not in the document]
Vincent-lau: `fixup_indirect_trace_exit()` is invoked each time we want to extend the trace, and it is possible for the new (potentially multiple) bbs to have multiple indirect exits, isn't it? Are you trying to say that once we have scanned some parts of the trace, next time we only need to scan those newly added bbs rather than the entire trace?

core/arch/interp.c
```diff
@@ -8007,3 +8252,157 @@ d_r_emulate(dcontext_t *dcontext, app_pc pc, priv_mcontext_t *mc)
...
+                                                       opnd_create_reg(jump_target_reg)));
+                    added_size += AARCH64_INSTR_SIZE;
+                }
+                /* b ibl_target */
+                branch = XINST_CREATE_jump(dcontext, opnd_create_pc(ind_target));
+                instr_exit_branch_set_type(branch, indirect_type);
+                instrlist_append(trace, branch);
+                added_size += AARCH64_INSTR_SIZE;
+            }
+            /* Don't invoke the decoder; only mangled instruction reached here. */
```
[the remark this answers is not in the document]
Vincent-lau: done

core/arch/interp.c
```diff
@@ -8007,3 +8252,157 @@ d_r_emulate(dcontext_t *dcontext, app_pc pc, priv_mcontext_t *mc)
...
+    trace_exit_label = NULL;
+    instr_t *delete_after = NULL;
+    /* We record the original trace end */
+    instr_t *trace_end = instrlist_last(trace);
+
+    LOG(THREAD, LOG_MONITOR, 4, "fixup the indirect trace exit\n");
+
+    for (instr = instrlist_first(trace); instr != trace_end;
+         instr = instr_get_next(instr)) {
+        if (instr_is_exit_cti(instr)) {
```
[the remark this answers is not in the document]
Vincent-lau: done

core/arch/interp.c
```diff
@@ -8007,3 +8252,157 @@ d_r_emulate(dcontext_t *dcontext, app_pc pc, priv_mcontext_t *mc)
...
+            /* Check for indirect exit. */
+            if (is_indirect_branch_lookup_routine(dcontext, (cache_pc)target)) {
+                /* This branch must be a cbnz, or the last_cti was not fixed up. */
+                if (instr->opcode != OP_cbnz) {
+                    prev = instr_get_prev(instr);
+                    added_size += mangle_indirect_branch_in_trace(
+                        dcontext, trace, instr, target, 0, &delete_after,
+                        instrlist_last(trace));
+                    /* re-check */
+                    instr = prev;
```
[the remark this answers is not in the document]
Vincent-lau: `mangle_indirect_branch_in_trace` will indeed modify the current instruction and I think that is the reason why we are doing the recheck.

core/arch/interp.c
Vincent-lau: I think this `b trace_head` is referring to the last instruction in the trace before any exit stubs, which should be a backward branch instruction, shouldn't it?

core/arch/interp.c
Vincent-lau: > So we have as many copies of the trace_exit: code as there are indirect branches in the trace? 
Yes, each time we have `is_indirect_branch_lookup_routine` true, we should create a new `trace_exit`
> And we need so many copies because we may need different scratch regs for each?
Think so.

core/arch/interp.c
Vincent-lau: ok, so do you want me to change all of the x2's to `IBL_TARGET_REG` around this function? I do notice that in emit_utils.c there is this comment
```
    *     x2: indirect branch target
     *     TLS_REG0_SLOT: app's x0
     *     TLS_REG1_SLOT: app's x1
     *     TLS_REG2_SLOT: app's x2
```
and to stay consistent the comment above `fixup_indirect_trace_exit` also mentions
```
 * The ibl routine requires:
 *     x2: contains indirect branch target
 *     TLS_REG2_SLOT: contains app's x2
 ```
 and I think it's probably ok to use x2 throughout?

**Comment by Vincent-lau** [comment]
Thanks for all the comments, as I said, the original author of this code was not me, and I have to guess the intent of the author  (just as you do). But I have replied to your review with my thoughts.

File: core/arch/interp.c [hunk 25]
```diff
@@ -8007,3 +8252,157 @@ d_r_emulate(dcontext_t *dcontext, app_pc pc, priv_mcontext_t *mc)
...
+         instr = instr_get_next(instr)) {
+        if (instr_is_exit_cti(instr)) {
+            target = instr_get_branch_target_pc(instr);
+            /* Check for indirect exit. */
+            if (is_indirect_branch_lookup_routine(dcontext, (cache_pc)target)) {
+                /* This branch must be a cbnz, or the last_cti was not fixed up. */
+                if (instr->opcode != OP_cbnz) {
+                    prev = instr_get_prev(instr);
+                    added_size += mangle_indirect_branch_in_trace(
+                        dcontext, trace, instr, target, 0, &delete_after,
```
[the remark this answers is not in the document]
Vincent-lau: Yeah, this does not make much sense to me either, `target` is the pc of the ibl routine, which is treated as the next bb's tag here. In fact, I suspect the case where `instr->opcode != OP_cbnz` is not hit at all, so might use an assert as well.

core/arch/aarch64/emit_utils.c
Vincent-lau: Ok, so when you say we need to document trace creation related stats, what exactly do you mean by that? Do we need to have a file somewhere to log all the stats of trace creation, or just put some data in this PR?

core/arch/interp.c
```diff
@@ -8007,3 +8252,157 @@ d_r_emulate(dcontext_t *dcontext, app_pc pc, priv_mcontext_t *mc)
...
+
+                /* Retrieve jump target reg from the xor instruction. */
+                prev = instr_get_prev(instr);
+                if (prev->opcode != OP_eor)
+                    return 0;
+
+                ASSERT(instr_num_srcs(prev) == 4 && opnd_is_reg(instr_get_src(prev, 1)));
+                jump_target_reg = opnd_get_reg(instr_get_src(prev, 1));
+
+                if (!ind_target || jump_target_reg == DR_REG_NULL)
```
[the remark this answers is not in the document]
Vincent-lau: done

Vincent-lau:
One other thing, I am thinking about making the trace support for AArch64 on by default, I am wondering roughly how much work would be needed for that?

core/arch/aarch64/emit_utils.c
abhinav92003: > as in where shall we document it

In this PR's description and the final commit message.

> I guess that we could use macros like GLOBAL_STAT(num_trace) in a call to LOG in the code?

Can you check which stats are already output in log files? We can extract those numbers and put in the PR's description.

core/arch/interp.c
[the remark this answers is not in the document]
abhinav92003: I see. Okay.

core/arch/interp.c
abhinav92003: Sure, I think it's fine.

core/arch/interp.c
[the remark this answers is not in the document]
abhinav92003: This block is to handle the stolen reg case, so the complete condition to detect whether this is the stolen reg case should go into the else-if condition. IMHO that improves readability; we can get `next` again.

core/arch/interp.c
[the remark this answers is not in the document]
abhinav92003: Well, we would need to have multiple copies of `trace_exit:` even if `fixup_indirect_trace_exit` was invoked separately to fix up each exit, right? Because each fixed up indirect branch would still use its own `trace_exit`.

I'm not convinced whether we need to iterate through the complete trace. I think the question to answer is: how many indirect exits do we fix up  each time `fixup_indirect_trace_exit` is invoked. Can it be >1? Does that happen when multiple bbs are added to the trace together?
If we fix up only one exit in each invocation of `fixup_indirect_trace_exit`, then it does seem better to avoid the loop and see if we can get to the ibl routine branch some other way.

Can you add a comment on the loop, clarifying its need.

core/arch/interp.c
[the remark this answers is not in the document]
abhinav92003: Where do we invoke `mangle_indirect_branch_in_trace` then?

It seemed to me that if we enter the `instr_is_exit_cti` case, one of the following will be true:
- it is a fixed conditional jump to `trace_exit`, added during a previous `fixup_indirect_trace_exit`. In this case `is_indirect_branch_lookup_routine` will be false.
- it is a direct branch to another fragment. In this case also `is_indirect_branch_lookup_routine` will be false.
- it is a branch to the ibl lookup routine. In this case `is_indirect_branch_lookup_routine` will be true. Sub-cases:
  - the branch to the ibl lookup routine can either be a cbz added by  `mangle_indirect_branch_in_trace` at L6195; or
  - a `b ibl_routine` that yet needs to be patched `mangle_indirect_branch_in_trace`

core/arch/interp.c
```diff
@@ -8309,15 +8308,8 @@ fixup_indirect_trace_exit(dcontext_t *dcontext, instrlist_t *trace)
             /* Check for indirect exit. */
             if (is_indirect_branch_lookup_routine(dcontext, (cache_pc)target)) {
                 /* This branch must be a cbnz, or the last_cti was not fixed up. */
-                if (instr->opcode != OP_cbnz) {
-                    prev = instr_get_prev(instr);
-                    added_size += mangle_indirect_branch_in_trace(
```
abhinav92003: If we remove this, then `mangle_indirect_branch_in_trace` isn't used at all right?

abhinav92003 (changes requested):

abhinav92003:
> One other thing, I am thinking about making the trace support for AArch64 on by default (after a few more iterations), I am wondering roughly how much work would be needed for that? I would imagine that the bare minimum is to pass all current tests with trace on by default and we might need to close the issue #5057 as well?

Atleast some targeted test coverage would be good, yes. In addition to this, we'd need to run some large internal apps/benchmarks under DR with traces enabled. In the past, we've seen that some issues show up only on large apps, not on unit tests or SPEC.

core/arch/interp.c
Vincent-lau: It is also called in `fixup_last_cti` on L6537 before we add a block to the trace. In `mixup_indirect_trace_exit`, we should expect it to be already fixed up when we iterate over the trace and check for indirect branch lookup routines, unless something has gone wrong.

core/arch/interp.c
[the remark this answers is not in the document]
Vincent-lau: done

core/arch/interp.c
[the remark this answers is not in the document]
Vincent-lau: I think it is indeed possible for multiple exits to be fixed when we invoke `mixup_indirect_trace_exit`. Here is an example trace fragment. I will add comments for clarification.
```
TAG  0x0000ffffad0385d4
 +0    L4 @0x0000fffd691e42c0  f9000380   str    %x0 -> (%x28)[8byte]
 +4    m4 @0x0000fffd691c4588  d29aee80   movz   $0xd774 lsl $0x00 -> %x0
 +8    m4 @0x0000fffd691c4f28  f2b52200   movk   %x0 $0xa910 lsl $0x10 -> %x0
 +12   m4 @0x0000fffd691c3338  f2dfffe0   movk   %x0 $0xffff lsl $0x20 -> %x0
 +16   L4 @0x0000fffd691c5128  ca1e0000   eor    %x0 %x30 lsl $0x00 -> %x0
 +20   L4 @0x0000fffd691e4450  b5000000   cbnz   $0x0000ffff6911ed80 <shared_trace_ibl_ret> %x0
 +24   L4 @0x0000fffd691c65b0  f9400380   ldr    (%x28)[8byte] -> %x0
 +28   L1                      a8c27bfd   ldp    (%sp)[16byte] %sp $0x0000000000000020 -> %x29 %
x30 %sp
 +32   L4 @0x0000fffd691c36f0  f9000380   str    %x0 -> (%x28)[8byte]
 +36   m4 @0x0000fffd691c5970  d29af880   movz   $0xd7c4 lsl $0x00 -> %x0
 +40   m4 @0x0000fffd691c43e0  f2b52200   movk   %x0 $0xa910 lsl $0x10 -> %x0
 +44   m4 @0x0000fffd691c3420  f2dfffe0   movk   %x0 $0xffff lsl $0x20 -> %x0
 +48   L4 @0x0000fffd691e5668  ca1e0000   eor    %x0 %x30 lsl $0x00 -> %x0
 +52   L4 @0x0000fffd691c6158  b5000000   cbnz   $0x0000ffff6911ed80 <shared_trace_ibl_ret> %x0
 +56   L4 @0x0000fffd691c3488  f9400380   ldr    (%x28)[8byte] -> %x0
 +60   L0                      910063e0   add    %sp $0x0018 lsl $0x00 -> %x0
 +64   L0                      91004001   add    %x0 $0x0010 lsl $0x00 -> %x1
 +68   L0                      910063e0   add    %sp $0x0018 lsl $0x00 -> %x0
 +72   L0                      91006000   add    %x0 $0x0018 lsl $0x00 -> %x0
 +76   L0                      d29afb1e   movz   $0xd7d8 lsl $0x00 -> %x30
 +80   L0                      f2b5221e   movk   %x30 $0xa910 lsl $0x10 -> %x30
 +84   L0                      f2dffffe   movk   %x30 $0xffff lsl $0x20 -> %x30
 +88   L4 @0x0000fffd691c5b38  14000000   b      $0x0000ffffa910d748
END 0x0000ffffad0385d4
```

core/arch/interp.c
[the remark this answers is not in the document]
Vincent-lau: In the case when `is_indirect_branch_lookup_routine` is true, we should expect `b ibl_routine` to be already fixed up by call to `mangle_indirect_branch_in_trace` in`fixup_last_cti`. So we should always be expecting a `cbnz` instruction.

core/arch/aarch64/emit_utils.c
Vincent-lau: I had a quick look at the log file generated by DR using `-loglevel 4 -logmask LOG_STATS` and there are some relevant stats such as
```
                 Fragments generated, bb and trace (thread):              1981
                   Basic block fragments generated (thread):              1946
                         Trace fragments generated (thread):                35
               Trace building reset: no trace head (thread):                35
               Number of bbs in all emitted traces (thread):               123
                  Maximum number of bbs in a trace (thread):                11
        Trace wannabes prevented from being traces (thread):              2391
                       Shadowed trace head deleted (thread):                35
                             Trace heads re-marked (thread):                 8
                        Future fragments generated (thread):              2558
                        Shared fragments generated (thread):              1852
                              Shared bbs generated (thread):              1817
                           Shared traces generated (thread):                35

```
And here is the log file if you want to take a look.
[log.0.345299.txt](https://github.com/DynamoRIO/dynamorio/files/7039370/log.0.345299.txt)

Vincent-lau:
> Atleast some targeted test coverage would be good, yes.

I am thinking about adding tests to cover some of the cases introduced in this PR, but not quite sure how to get started. My initial thought is to write a basic block such as 
```
br  %x3
b  ibl_lookup_routine
```
And it would be mangled into 
```
str %x2 -> IBL_TARGET_SLOT
mov %x3 -> %x2
b ibl_lookup_routine
```
And then we expect it to be fixed up as 
```
str  %x0 -> TLS_REG0_SLOT
mov #trace_next_target -> %x0
eor %x0 %x3 -> %x0
cbnz %x0 trace_exit
ldr TLS_REG0_SLOT -> %x0
```

Vincent-lau:
> we'd need to run some large internal apps/benchmarks under DR with traces enabled

Can you suggest some applications to run?

core/arch/aarch64/emit_utils.c
abhinav92003: This looks useful. I suppose the stats in your above comment were for some small test app in the DR suite (as the counts seem low)? It'd be good to document these counts for SPEC as well.

abhinav92003:
> I am thinking about adding tests to cover some of the cases introduced in this PR, but not quite sure how to get started. My initial thought is to write a basic block such as
> 
> ```
> br  %x3
> b  ibl_lookup_routine
> ```

Note that the ` b  ibl_lookup_routine` will not be present in the test app itself; it will be added by DR.

> I problem I have, though, with this idea is that I am not sure if it is actually possible for the compiler to generate such assembly instructions in a bb. On the other hand, if I write tests in C, then I can be sure that it is legitimate code generated by the compiler but can't be sure that it will cover the cases introduced by this PR.

You can add your test basic blocks in assembly, similar to drreg tests that are also written in assembly to verify behaviour for some hand-crafted basic blocks. See [drreg-test.c](https://github.com/DynamoRIO/dynamorio/blob/be78c124eb787601182e3c73f4be8bb859c50ef8/suite/tests/client-interface/drreg-test.c#L876) (the test app for drreg) and [drreg-test.dll.c](https://github.com/DynamoRIO/dynamorio/blob/9822a6515e48446de727af3b58108d35d46ae431/suite/tests/client-interface/drreg-test.dll.c#L164) (the test client for drreg).

I guess you can use `dr_register_trace_event` and `dr_register_end_trace_event` to register callbacks for trace events, and in those callbacks you can verify the instructions in the trace. I have never looked at a trace closely myself, so please verify how the traces look on these events.

core/arch/aarch64/emit_utils.c
Vincent-lau: It is run on a small program, although not on the test suite. Here is some log information on a test suite of SPEC CPU2017
```
                 Fragments generated, bb and trace (thread):             61212
                   Basic block fragments generated (thread):             52009
                         Trace fragments generated (thread):              9203
               Trace building reset: no trace head (thread):              9203
               Number of bbs in all emitted traces (thread):             22522
                  Maximum number of bbs in a trace (thread):                22
        Trace wannabes prevented from being traces (thread):            100694
                       Shadowed trace head deleted (thread):              9203
                             Trace heads re-marked (thread):               118
                        Future fragments generated (thread):             68257
                        Shared fragments generated (thread):             38679
                              Shared bbs generated (thread):             29476
                           Shared traces generated (thread):              9203

```
And this is the stats on the test common.broadfun
```
              Fragments generated, bb and trace (thread):              1902
                   Basic block fragments generated (thread):              1856
                         Trace fragments generated (thread):                46
               Trace building reset: no trace head (thread):                46
               Number of bbs in all emitted traces (thread):               136
                  Maximum number of bbs in a trace (thread):                11
        Trace wannabes prevented from being traces (thread):              2622
                       Shadowed trace head deleted (thread):                46
                             Trace heads re-marked (thread):                 8
                        Future fragments generated (thread):              2422
                        Shared fragments generated (thread):              1761
                              Shared bbs generated (thread):              1715
                           Shared traces generated (thread):                46

```

Vincent-lau:
Thanks for the suggestions on writing the test, can you also take a look at my latest commit and see if there is any problem with it?

core/arch/interp.c
```diff
@@ -5771,6 +5770,170 @@ mangle_x64_ib_in_trace(dcontext_t *dcontext, instrlist_t *trace, instr_t *target
...
+ *    ldr x0, [x28]
+ *    b target            <- delete after
+ * fall:
+ *    ldr x0, [x28]
+ *    b fall_target
+ *    ...
+ * If we delete all code after "b target", then the "fall" path would
+ * be lost. Therefore we need to append the fall path at the end of
+ * the trace as a fake exit stub. Swapping them might be dangerous since
+ * a stub trace may be created on both paths.
```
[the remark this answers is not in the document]
derekbruening: > what would happen if target_cache is pointing to one of these mangled instructions 

The uncompleted mangled app instruction will be re-executed, so all state changes have to be rolled back: restore spilled registers, etc.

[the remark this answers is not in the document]
derekbruening: > > Are you interested in implementing #5062?
> 
> How long do you estimate this would take? I am currently on the final week of an internship where I try to port the trace support for a64. Obviously I can continue contributing after the internship but I am just trying to get a sense of how long this is going to take.

I don't know for sure -- it might be quite straightforward:
+ go enable things permanently that are currently gated on monitor_data_t.pass_to_client like extend_unmangled_ilist's work
+ change internal_extend_trace to estimate the size from the unmangled list (maybe w/ some fudge factor for mangling: for new scheme only used to keep trace below max sizes I would assume)
+ delete monitor_data_t.trace_buf, extend_trace, etc.
+ try to delete decode_fragment: check other callers

But there might be some extra things to deal with once in the details.

Maybe better to finish this PR off as a deliverable -- but the missing translation support is an issue.  Could revert the optimization and have sub-optiimal code that matches x86 with the spill-and-move left there and a simple exit stub and complete translation support, with a comment that we can improve the perf later.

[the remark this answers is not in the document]
Vincent-lau: > The uncompleted mangled app instruction will be re-executed, so all state changes have to be rolled back: restore spilled registers, etc.

In that case is it OK to include these instructions in the recreated ilist
```
 * trace_exit_label:
 *      ldr x0, TLS_REG0_SLOT
 *      str x2, TLS_REG2_SLOT
 *      mov x2, jump_target
 *      b ibl_routine
 ```
 And update the translation to restore the app state if necessary?

File: core/arch/interp.c (hunk 25 above)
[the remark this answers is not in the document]
Comment (derekbruening) [reply core/arch/interp.c]: You mean line 8468 I assume.  Adding link: PR #2442.

Agreed this doesn't look right and replacing the whole if() with an assert seems reasonable: unless we start adding traces to traces or many-i.b.-containing blocks (may have been done at one point but against client rules now) it seems this should not happen.

File: core/arch/aarch64/emit_utils.c (hunk 6 above)
Comment (derekbruening) [reply core/arch/aarch64/emit_utils.c:446]: With the other tests seems fine.

**Comment by derekbruening** [comment]
Seeing unresolved comments: please resolve when addressed

File: core/arch/interp.c [hunk 29]
```diff
@@ -8342,31 +8343,36 @@ fixup_indirect_trace_exit(dcontext_t *dcontext, instrlist_t *trace)
                 /* Add the trace exit label. */
                 instrlist_append(trace, trace_exit_label);
                 /* ldr x0, TLS_REG0_SLOT */
-                instrlist_append(
-                    trace,
-                    instr_create_restore_from_tls(dcontext, scratch, TLS_REG0_SLOT));
+                new_instr =
+                    instr_create_restore_from_tls(dcontext, scratch, TLS_REG0_SLOT);
+                instr_set_translation(new_instr, instr_get_translation(instr));
+                instrlist_append(trace, new_instr);
```
derekbruening, line 8349: You can use INSTR_XL8 to make all of these a single statement again if desired

core/arch/interp.c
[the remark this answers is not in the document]
derekbruening: > In that case is it OK to include these instructions in the recreated ilist (currently they are not)

Wait -- why are they not in the recreation of the trace?  That is not good.  It should go through the same mangling steps.

core/arch/interp.c
```diff
@@ -5771,6 +5771,172 @@ mangle_x64_ib_in_trace(dcontext_t *dcontext, instrlist_t *trace, instr_t *target
...
+ *    b target            <- delete after
+ * fall:
+ *    ldr x0, [x28]
+ *    b fall_target
+ *    ...
+ * If we delete all code after "b target", then the "fall" path would
+ * be lost. Therefore we need to append the fall path at the end of
+ * the trace as a fake exit stub. Swapping them might be dangerous since
+ * a stub trace may be created on both paths.
+ */
```
derekbruening, line 5883: Please add a `XXX i#5062` comment about being able to delete this whole function once we switch to the simpler #5062 approach.

core/arch/interp.c
```diff
@@ -5771,6 +5771,172 @@ mangle_x64_ib_in_trace(dcontext_t *dcontext, instrlist_t *trace, instr_t *target
...
+
```
derekbruening, line 5884: Please include `TODO i#xxxx` comments (with the real issue number) for the missing pieces: in particular, adding translation support for all the new mangling code.  (Ideally it would be in this PR but for making progress here w/ a new feature in pieces ok to not have all at once so long as what's missing is clearly documented not just in the commit description but in the code too).  I would put such comments both here and in translate.c.

core/arch/interp.c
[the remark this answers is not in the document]
Vincent-lau: Because these instructions are added in `fixup_indirect_trace_exit` which is not called when `recreate_fragment_ilist` runs. I have added a call in `recreate_fragment_ilist` so that these instructions are included in the recreated instrlist as well.

core/arch/interp.c
Vincent-lau: done

core/arch/interp.c
Vincent-lau: done

core/arch/interp.c
Vincent-lau: I feel this there isn't any translation support needed for the new mangling code. The new mangling code includes:
```
     * Original ibl lookup:
     *      str tgt_reg, TLS_REG2_SLOT
     *      mov tgt_reg, jump_target
     *      b ibl_routine
     * Now we rewrite it into:
     *      str x0, TLS_REG0_SLOT
     *      mov x0, #trace_next_target
     *      eor x0, x0, jump_target
     *      cbnz x0, trace_exit (ibl routine)
     *      ldr x0, TLS_REG0_SLOT
     */
   ```
   and
   ```
    * For example, the code will be fixed to:
 *      eor x0, x0, jump_target
 *      cbnz x0, trace_exit_label
 *      ...
 *      b trace_head
 * trace_exit_label:
 *      ldr x0, TLS_REG0_SLOT
 *      str x2, TLS_REG2_SLOT
 *      mov x2, jump_target
 *      b ibl_routine
   ```
which are just basic spilling of registers and perform some calculation after spilling, I think? What sort of support do you think would be needed for these new mangling instructions?

Comment (Vincent-lau) [reply core/arch/interp.c:5884]: The issue I mentioned previously about `recreate_fragment_ilist` not being able to find an instruction was purely because the new mangled instructions added by `fixup_indirect_trace_exit` is not added when creating the fragment ilist.

Comment (derekbruening) [reply core/arch/interp.c:5887]: Please run with -stress_recreate_state and verify there are no translation failures or warnings about unknown mangling sequences.

Comment (derekbruening) [reply core/arch/interp.c:5887]: Looks like the "unsupported mangle instr" message is at `DOLOG(2` -- I thought it was more visible esp with -stress_recreate_state.  Please run with `-loglevel 2 -stress_recreate_state` and ensure there are no "unsupported mangle instr" messages in the log files.

File: core/arch/interp.c [hunk 32]
```diff
@@ -5421,6 +5420,8 @@ recreate_fragment_ilist(dcontext_t *dcontext, byte *pc,
             instrlist_destroy(dcontext, bb);
         }
 
+        IF_AARCH64(fixup_indirect_trace_exit(dcontext, ilist));
```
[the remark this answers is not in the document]
Vincent-lau: If I understand correctly, you are saying that once we switch to #5062, we are able to put this call only inside `mangle_trace()` and now we still need to have a special call here. I will add a comment about that.

core/arch/arch_exports.h
```diff
@@ -2196,6 +2196,11 @@ int
 append_trace_speculate_last_ibl(dcontext_t *dcontext, instrlist_t *trace,
                                 app_pc speculate_next_tag, bool record_translation);
 
+#ifdef AARCH64
+int
+fixup_indirect_trace_exit(dcontext_t *dcontext, instrlist_t *trace);
```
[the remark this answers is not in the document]
Vincent-lau: done

core/arch/interp.c
Vincent-lau: I think there are already quite a lot of unsupported mangle instr even without trace turned. I was not able to find any issues mentioning this so I opened a new one #5069. Feel free to point out if there are already issues that are talking about this problem.

I will also add a inline comment that references this translation support issue.

Vincent-lau: I added a comment in translate.c referencing issue #5069 but I didn't add any in interp.c because I believe all those mangled instructions added by trace support will be covered by the condition `instr_is_trace_cmp` and never reach the `DOLOG(2`. And again, correct me if I am wrong, but I don't think we have introduced any instructions that require additional support.

derekbruening: In general we really prefer linear control flow and a lot of code assumes linear control flow, including translation.  This linear requirement used to be a strong invariant but various issues have required adding limited internal control flow.  This PR is adding internal control flow with different spill/restore logic on different paths, which is going to break translate.c's assumptions on linear control flow.  There's a spill that's not restored until trace_exit_label and the translate code will see "spill x0; restore x0; restore x0" and probably freak out -- when you ran with stress did it assert?  I think it focuses on "our mangling" regions -- I would double-check that the trace_exit_label code is properly marked as "our mangling", and that the stress code walks the list all the way to the end including these trace_exit_label sequences.  This trace_exit_label complexity is going to cause headaches with translation and require special support.  IMHO alternatives should be strongly considered such as the proposal I had earlier.

derekbruening: > I think there are already quite a lot of unsupported mangle instr even without trace enabled. I was not able to find any issues mentioning this so I opened a new one #5069. Feel free to point out if there are already issues that are talking about this problem.

This is surprising since I put in work earlier this year to enable -stress_recreate_state, fix a number of translation problems on aarch64, and enable a regression test: #4680.  Taking a look.

derekbruening: To make progress let's merge this but please first add a comment in the code and in the commit message about these separate trace_exit_label sequences breaking translation and how either special support needs to be added or (the better choice IMHO: best to avoid internal control flow) they need to be removed and a different scheme used such as the one proposed at https://github.com/DynamoRIO/dynamorio/pull/5045/files/b9ada873ecc0b86128169aae0bd7ad648c0f921b..fc70e80bad8a66b7f417c27673fa5659e96ff1ef#r697566139

derekbruening (approved):

core/arch/aarch64/emit_utils.c
```diff
@@ -370,14 +370,14 @@ patch_branch(dr_isa_mode_t isa_mode, cache_pc branch_pc, cache_pc target_pc,
     ASSERT(ALIGNED(branch_pc, 4) && ALIGNED(target_pc, 4));
     if ((enc & 0xfc000000) == 0x14000000) { /* B */
         ASSERT(off + 0x8000000 < 0x10000000);
-        *pc_writable = (0x14000000 | (0x03ffffff & off >> 2));
+        *pc_writable = (0x14000000 | (0x03ffffff & (off >> 2)));
     } else if ((enc & 0xff000010) == 0x54000000 ||
                (enc & 0x7e000000) == 0x34000000) { /* B.cond, CBNZ, CBZ */
-        ASSERT(off + 0x40000 < 0x80000);
-        *pc_writable = (enc & 0xff00001f) | (0x00ffffe0 & off >> 2 << 5);
+        ASSERT(off + 0x100000 < 0x200000);
```
derekbruening, line 376: Wait -- where did this change come from?  This seems unrelated to trace building and seems like it could affect basic execution, unlike the rest of this PR which is all under the off-by-default trace option.  Please separate this into its own PR and probably ask @AssadHashmi to review.

derekbruening (changes requested):

core/arch/aarch64/emit_utils.c
Vincent-lau: This is an arithmetic error that I found which will fail the test if I use -enable_traces for all the tests (which is not yet turned on in this patch, I simply tried that offline). Shall I undo this change and make a separate PR for this?

core/arch/interp.c
Vincent-lau: > the better choice IMHO: best to avoid internal control flow

I agree with you that a linear control flow is preferable, will add a comment in the code for `trace_exit_label

core/arch/interp.c
Vincent-lau: @abhinav92003 Are you happy with this explanation? If so can you approve this change you have requested?

Vincent-lau:
It seems that I need to add a lot in the final final commit message 😄 , might as well include it here before I merge this into the master branch just in case I forgot something.

```
i#2974 trace support for AArch64, part 2: implementation of the main part of support 

This patch incorporated changes from patch #2442, and fixed some corner cases
where the assumption was incorrect and caused the program to crash.

Trace support is not yet enabled by default, but can be enabled with "-enable_traces".

This commit introduces internal control flow by adding a trace_exit_label in
fixup_indirect_trace_exit, which might break code that assumes linear control
flow. Either special support is needed for this trace_exit_label or alternative
schemes using linear control.

Co-authored-by: Kevin Zhou <Kevin.Zhou@arm.com>

Issues: #1569, #2974
```

core/arch/interp.c
Vincent-lau: > This is surprising since I put in work earlier this year to enable -stress_recreate_state, fix a number of translation problems on aarch64, and enable a regression test: #4680. Taking a look.

I am guessing it's because these instructions are just reported with a log message, which did not catch your attention while you were fixing them?

derekbruening: I recall looking at such log messages -- but I think those memories are from rseq translation work #4316 and the #4680 is as you said: focusing on asserts and fatal problems.

core/arch/aarch64/emit_utils.c
```diff
@@ -370,14 +370,14 @@ patch_branch(dr_isa_mode_t isa_mode, cache_pc branch_pc, cache_pc target_pc,
     ASSERT(ALIGNED(branch_pc, 4) && ALIGNED(target_pc, 4));
```
derekbruening: >  i#2974 trace support for AArch64, part 18: Adding comment on preferring
linear control flow

Note that the ", part N: " is a convention for merges to master: i.e., this PR being merged will be part 1; the next PR for trace building merged into master would be part 2 (not part 19 or sthg: the PR-internal ones don't count toward the master merges).

derekbruening (approved):
Thank you for the PR, and taking the time to go through what ended up being a long review process!

core/arch/aarch64/emit_utils.c
Vincent-lau: sure, that's why I put part 2 in my final commit message above

core/arch/arch_exports.h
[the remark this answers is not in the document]
abhinav92003: Instead of 'here', maybe mention `end_and_emit_trace`.

abhinav92003 (approved):
Thanks for the discussion and addressing all comments!

Issue #4943, opened 2021-06-09
ASSERT (opnd_create_immed_int: value too large for 8-bit size) in offline tracer when running with L0_filter and if application has basic block longer than 256 insts


prasun3:
**Describe the bug**
The assert occurs if we have a basic block longer than 256 instructions. The assert goes away if I change `OPND_CREATE_INT8(ud->bb_instr_count)));` to `OPND_CREATE_INT16(ud->bb_instr_count)));`

```
   1051        if (op_L0_filter.get_value()) {
   1052            // Count dynamic instructions per thread.
   1053            // It is too expensive to increment per instruction, so we increment once
   1054            // per block by the instruction count for that block.
   1055            if (drmgr_is_first_nonlabel_instr(drcontext, app)) {
   1056                // On x86 we could do this in one instruction if we clobber the flags: but
   1057                // then we'd have to preserve the flags before our same-line skip in
   1058                // insert_filter_addr().
   1059                dr_insert_read_raw_tls(drcontext, ilist, where, tls_seg,
   1060                                       tls_offs + sizeof(void *) * MEMTRACE_TLS_OFFS_ICOUNT,
   1061                                       reg_ptr);
   1062                MINSERT(ilist, where,
  >1063                        XINST_CREATE_add(drcontext, opnd_create_reg(reg_ptr),
   1064                                         OPND_CREATE_INT8(ud->bb_instr_count)));
   1065                dr_insert_write_raw_tls(drcontext, ilist, where, tls_seg,
   1066                                        tls_offs + sizeof(void *) * MEMTRACE_TLS_OFFS_ICOUNT,
   1067                                        reg_ptr);
   1068            }

(gdb) p ud->bb_instr_count
$5 = 257
```
**To Reproduce**
Run offline tracer with the L0_filter option and an application that has a basic block longer than 256 instructions.

`drrun -t drcachesim -offline -L0_filter -- <app>`

**Expected behavior**
The assert should not fail.

**Screenshots or Pasted Text**

```
opnd_create_immed_int: value too large for 8-bit size
```

**Versions**
 - What version of DynamoRIO are you using? cronbuild-8.0.18752
 - Does the latest build from https://github.com/DynamoRIO/dynamorio/releases solve the problem? No
- What operating system version are you running on? Ubuntu 18.04.5 LTS
- Is your application 32-bit or 64-bit? ELF 64-bit LSB executable

**Additional context**
None

derekbruening:
[...]

PR #5046, opened 2021-08-12
Fix for assert in offline tracer with L0_filter

prasun3:
We see the assert 'ASSERT (opnd_create_immed_int: value too large for 8-bit
size)' because basic blocks larger than max_bb_instrs(256) are generated (see #4943)
This change fixes the condition check.

### DIFF
--- a/core/arch/interp.c
+++ b/core/arch/interp.c
@@ -3908,7 +3908,7 @@ build_bb_ilist(dcontext_t *dcontext, build_bb_t *bb)
             BBPRINT(bb, 3, "reached end pc " PFX ", stopping\n", bb->stop_pc);
             break;
         }
-        if (total_instrs > DYNAMO_OPTION(max_bb_instrs)) {
+        if (total_instrs == DYNAMO_OPTION(max_bb_instrs) - 1) {
             /* this could be an enormous basic block, or it could
              * be some degenerate infinite-loop case like a call
              * to a function that calls exit() and then calls itself,
--- a/core/arch/interp.c
+++ b/core/arch/interp.c
@@ -3908,7 +3908,7 @@ build_bb_ilist(dcontext_t *dcontext, build_bb_t *bb)
             BBPRINT(bb, 3, "reached end pc " PFX ", stopping\n", bb->stop_pc);
             break;
         }
-        if (total_instrs == DYNAMO_OPTION(max_bb_instrs) - 1) {
+        if (total_instrs > DYNAMO_OPTION(max_bb_instrs)) {
             /* this could be an enormous basic block, or it could
              * be some degenerate infinite-loop case like a call
              * to a function that calls exit() and then calls itself,


core/arch/interp.c
```diff
@@ -3908,7 +3908,7 @@ build_bb_ilist(dcontext_t *dcontext, build_bb_t *bb)
             BBPRINT(bb, 3, "reached end pc " PFX ", stopping\n", bb->stop_pc);
             break;
         }
-        if (total_instrs > DYNAMO_OPTION(max_bb_instrs)) {
+        if (total_instrs == DYNAMO_OPTION(max_bb_instrs) - 1) {
```
derekbruening, line 3911: So this is a swing of 2: but I think we only need to reduce by 1?  The artificial jump added on truncation is not passed to clients: it's like other mangling added later.  So if `-max_bb_instrs 4` means 4 app instrs passed to clients, we'd want ==, not ==-1, right?

Plus, if we did want -1 and to stop at 3 instead of 4, this is not sufficient: the inner loop above currently has `total_instrs <= cur_max_bb_instrs`.

derekbruening: I see a bug in a prior change: this should use `cur_max_bb_instrs` to handle single-step, rather than reading the option.

derekbruening: Should we add a sanity check at checklevel 1 or sthg (`DOCHECK(1, ...)`) in client_process_bb() that counts the app instrs and ensures the count is <= DYNAMO_OPTION(max_bb_instrs)?

prasun3: > 
> 
> So this is a swing of 2: but I think we only need to reduce by 1? The artificial jump added on truncation is not passed to clients: it's like other mangling added later. So if `-max_bb_instrs 4` means 4 app instrs passed to clients, we'd want ==, not ==-1, right?
> 
> Plus, if we did want -1 and to stop at 3 instead of 4, this is not sufficient: the inner loop above currently has `total_instrs <= cur_max_bb_instrs`.

With the default (-max_bb_instrs 256), the block size was going to 257 but OPND_CREATE_INT8 can only handle 255. I now realize that 256 is not a hard limit and the user could specify a higher value. In that case, is it better to replace the OPND_CREATE_INT8 with OPND_CREATE_INT16?

derekbruening: > is it better to replace the OPND_CREATE_INT8 with OPND_CREATE_INT16?

Yes I agree that should be done since IIRC the encoder will still use the 1-byte immed if it fits.

derekbruening:
> prasun3 closed this 6 hours ago

But we also want to fix the fencepost error with -max_bb_instrs

prasun3:
> 
> 
> > prasun3 closed this 6 hours ago
> 
> But we also want to fix the fencepost error with -max_bb_instrs

I have opened #5056 to track this issue

Issue #4396, opened 2020-08-06
Add third-part verification tests for AArch64 decoder/encoder


AssadHashmi:
The current test set for the AArch64 decoder/encoder checks internal consistency but not correctness against an external third-party reference decoder/encoder. For AArch64, GNU's binutils and/or QEMU could be used for such cross-checking.

The x86 decoder/encoder already does this using binutils and udis86.

This issue was raised as a result of discussions in #4395 , see https://github.com/DynamoRIO/dynamorio/pull/4395#discussion_r466503441.

derekbruening:
Is the dis-a64 test today using binutils objdump in some fashion at least (I guess up to a human to notice if its operands are different from DR's though?)?

AssadHashmi:
> Is the dis-a64 test today using binutils objdump in some fashion at least (I guess up to a human to notice if its operands are different from DR's though?)?

Yes, currently we use GCC+bunutils objdump in `dis-a64.txt` tests to establish that we decode to the correct opcode. We then re-encode from the IR signature (added as part of the test) to check the hex encoding created matches GCC's. This doesn't check the correctness of the signature in terms of instruction interpretation, just that the specified IR signature string generates the same hex encoding as was decoded.

The plan is to close the loop when we complete the encoder tests checking that the IR signature created by macros in `ir_aarch64.c` matches the signatures in `dis-a64.txt` generating the same hex encoding as GCC's in `dis-a64.txt`.

Is that adequate?

derekbruening:
> Yes, currently we use GCC+bunutils objdump in dis-a64.txt tests to establish that we decode to the correct opcode.

The opcode matching is by a human when dis-a64.txt is updated?

> We then re-encode from the IR signature (added as part of the test) to check the hex encoding created matches GCC's. 

Not sure I follow "from the IR signature": do you mean the disassembly?  I believe the test decodes from the 4-byte hex encoding into IR and then encodes from IR to hex: so the disassembly strings in the file are not used by the test?

Also, from a glance at the test: I don't see the original raw bits being invalidated, so I would expect the re-encoding to just copy the original raw bits and not do any encoding from the IR?  Or does #5176 prevent this (I do see a check in instr_encode_arch() for the raw bits being valid turning into a memcpy but maybe they are never valid because of #5176)?  I would expect the test to explicitly mark the raw bits invalid to ensure full encoding is triggered.

> Is that adequate?

[...]

Issue #2443, opened 2017-05-26
add exhaustive AArch64 decoder/encoder/macro consistency tests


derekbruening:
For x86 we have exhaustive tests of every INSTR_CREATE and XINST_CREATE
macro with self-consistency tests that the decoder and encoder agree on
everything.

For ARM we are still missing this: #1686.

This issue covers adding such tests for AArch64.

Without these tests, it is fragile to add new XINST_CREATE macros: we have
no idea whether they work.

derekbruening:
Xref #2440

egrimley:
Do the tests of XINST_CREATE macros have to be architecture-specific? It might be better if they weren't. At the moment it is perhaps not always clear what kinds of operand are allowed (register or immediate), what range of immediate values, and so on.

Note that we already have "api.dis-a64" for testing the A64 encoder and decoder, without the macros.

derekbruening:
Sure, some of the XINST could be shared.

The api.dis test is a decoding and disassembly test, comparing the DR decoder/disassembler to other disassemblers using a moderately large, relatively random binary input and comparing the output to what other tools agree on.  At least that's what it is for both x86 and ARM.

The api.ir test that this issue covers is an internal self-consistency test that systematically goes through every INSTR_CREATE macro with every type variation and ensures it can encode and be decoded back and that all varieties of decoder agree (fast and slow).  It says nothing about agreeing with a ground truth: only internal consistency, but it systematically covers the full IR, whereas the decode-this-binary api.dis test only covers what happens to be in the binary input.

egrimley:
The test "api.dis-a64" has a big set of cases (dis-a64.txt) and checks encoder and decoder (dis-a64.c). It doesn't check different varieties of decoder because those aren't implemented yet (i#2374).

brettcoon:
Another benefit of having at least one test for every INST_CREATE macro is that it provides example invocations.  Could have used that today.

PR #5047, opened 2021-08-12
i#2626: AArch64 v8.0 decode: Fix indexed vector instructions

AssadHashmi:
In addition to fixing broken indexed vector instructions,
this patch also adds:
```
FMLS <Vd>.<T>, <Vn>.<T>, <Vm>.<Ts>[<index>]
FMUL <V><d>, <V><n>, <Vm>.<Ts>[<index>]
FMUL <Vd>.<T>, <Vn>.<T>, <Vm>.<Ts>[<index>]
```
Issue: #2626

### DIFF
--- a/core/ir/aarch64/codec.c
+++ b/core/ir/aarch64/codec.c
@@ -1878,19 +1878,6 @@ encode_opnd_sysops(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_ou
     return encode_opnd_int(5, 14, false, 0, 0, opnd, enc_out);
 }
 
-/* dq16_idx_lhm: imm4 from bits 16-20, the lower 4 bits of register Rm with idx_lhm */
-static inline bool
-decode_opnd_dq16_idx_lhm(uint enc, int opcode, byte *pc, OUT opnd_t *opnd)
-{
-    return decode_opnd_int(16, 4, false, 0, OPSZ_4b, 0, enc, opnd);
-}
-
-static inline bool
-encode_opnd_dq16_idx_lhm(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_out)
-{
-    return encode_opnd_int(16, 4, false, 0, 0, opnd, enc_out);
-}
-
 /* sysreg: system register, operand of MRS/MSR */
 
 static inline bool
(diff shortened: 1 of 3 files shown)


core/ir/aarch64/codec.c
```diff
@@ -2385,49 +2372,44 @@ encode_opnd_x16immvs(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_
 static inline bool
 decode_opnd_vindex_H(uint enc, int opcode, byte *pc, OUT opnd_t *opnd)
 {
-    uint bits = (enc >> 11 & 1) << 2 | (enc >> 21 & 1) << 1 | (enc >> 20 & 1);
-    *opnd = opnd_create_immed_int(bits, OPSZ_2b);
+    /* Example encoding:
+     * FMLA <Vd>.<T>, <Vn>.<T>, <Vm>.H[<index>]
+     * 3322222222221111111111
```
derekbruening, line 2377: What is this 33222...?
Oh, is it the tens digit for the offsets below?  Hmm, ok.  Slightly confusing, along with the `z` on its own line (I figured out that it's part of `sz`) -- but not sure I have a much better idea: 2 digits for each perhaps but prob should just leave it.

core/ir/aarch64/codec.txt
```diff
@@ -1097,20 +1104,23 @@ x101101011000000000101xxxxxxxxxx  n  cls     wx0 : wx5
 0x001110xx1xxxxx100001xxxxxxxxxx  n     add       dq0 : dq5 dq16 bhsd_sz
 0x001110xx1xxxxx100011xxxxxxxxxx  n     cmtst     dq0 : dq5 dq16 bhsd_sz
 0x001110xx1xxxxx100101xxxxxxxxxx  n     mla       dq0 : dq0 dq5 dq16 bhs_sz
-0x101111xxxxxxxx0000x0xxxxxxxxxx  n     mla       dq0 : dq5 dq16_idx_lhm bhsd_sz idx_lhm
+0x1011111xxxxxxx0000x0xxxxxxxxxx  n     mla       dq0 : dq5 dq16 vindex_SD sd_sz
+0x10111101xxxxxx0000x0xxxxxxxxxx  n     mla       dq0 : dq5 dq16_h_sz vindex_H h_sz
```
derekbruening, line 1108: So the bug being fixed was that the size was wrong for this variation?
How was the bug identified?  Could we capture how it was identified and make a test?

Comparing to external decoders does catch it just by disasm comparison:
```
$ disasm_a64 2f620820 
llvm-mc:   2f620820 mla v0.4h, v1.4h, v2.h[6]
capstone:  2f620820 mla v0.4h, v1.4h, v2.h[6]
bfd:       2f620820 mla v0.4h, v1.4h, v2.h[6]
DynamoRIO: 2f620820 mla %d1 $0x02 $0x01 $0x06 -> %d0
```

That's #4396.

derekbruening (approved):

core/ir/aarch64/codec.c
AssadHashmi: This was an attempt at commenting the `H:L:M` bit-twiddling for clarity and future maintenance.

I've used the same bit numbering scheme in `encode_opnd_fpimm8()`, `decode_opnd_fpimm13()` and other places but that divides the encoding into nibbles which is more obvious as a bit numbering comment I guess:
https://github.com/DynamoRIO/dynamorio/blob/d998d2bf371a53fe8c196c1152c75b5fa0c356ee/core/ir/aarch64/codec.c#L1821
https://github.com/DynamoRIO/dynamorio/blob/d998d2bf371a53fe8c196c1152c75b5fa0c356ee/core/ir/aarch64/codec.c#L2594

I can't think of a better way of commenting this sort of information in `ASCII` text.

core/ir/aarch64/codec.txt
AssadHashmi: > So the bug being fixed was that the size was wrong for this variation?

No, it's a few things, the main one that the operand which should be a source register is an immediate int, e.g.
```
2f620820 : mla v0.4h, v1.4h, v2.h[6]                : mla    %d1 $0x02 $0x01 $0x06 -> %d0
2f620820 : mla v0.4h, v1.4h, v2.h[6]                : mla    %d1 %d2 $0x06 $0x01 -> %d0
                                                                 ^^^
```
The other problems with the existing implementation were:
- Use of new fields (`dq16_idx_lhm`, `idx_lhm`) incorrectly duplicating existing ones (`vindex_H`, `vindex_SD`, `dq16_h_sz`)
- Returning 2 bit integer operand for half-precision element indices rather than 3 bit in `decode_opnd_vindex_H()`.
  https://github.com/DynamoRIO/dynamorio/blob/d998d2bf371a53fe8c196c1152c75b5fa0c356ee/core/ir/aarch64/codec.c#L2389

> How was the bug identified?

I noticed the new fields when I returned from holiday but couldn't look at them properly until adding the new indexed vector `FMLS` and `FMUL` instructions.

> Could we capture how it was identified and make a test?

I think these problems would have been identified when adding IR creation tests in `suite/tests/api/ir_aarch64.c`, https://github.com/DynamoRIO/dynamorio/issues/2443, but we're currently focussing on decode rather than encode.

Where did you build `disasm_a64` from? I can't see it in the DR tree. Looks handy!

derekbruening:
```
$ type disasm_a64
disasm_a64 is a function
disasm_a64 () 
{ 
    bytes=`echo $1 | sed 's/\(..\)/0x\1 /g' | awk '{printf "%s %s %s %s\n", $4, $3, $2, $1}'`;
    echo "$bytes" | /usr/bin/llvm-mc -arch aarch64 --disassemble | grep --color=auto -v '\.text' | xargs echo "llvm-mc:   $1";
    echo "$bytes" | ~/bin/capstone -aarch64 - | cut -c 9-100 | xargs echo "capstone: ";
    echo -e ".global _start\n.align 6\n_start: .word 0x$1\n" > /tmp/_ddd.asm;
    aarch64-linux-gnu-as -g -o /tmp/_ddd.o /tmp/_ddd.asm || return;
    aarch64-linux-gnu-ld -static -o /tmp/_ddd /tmp/_ddd.o || return;
    aarch64-linux-gnu-objdump -D -d /tmp/_ddd | grep --color=auto -A 1 _start | grep --color=auto -v '_start>:' | cut -c 11-100 | xargs echo "bfd:      ";
    ~/dr/git/build_x64_tgt_a64_dbg/clients/bin64/drdisas ${bytes} | xargs echo "DynamoRIO:"
}
```

(For capstone I created a front-end program myself.  I can try to share it somewhere.)

Issue #4673, opened 2021-01-13
tool.drcachesim.coherence test failed due to output mismatch


derekbruening:
Happened on GA CI:
https://github.com/DynamoRIO/dynamorio/pull/4672/checks?check_run_id=1697584999

It looks like some minor output mismatch:
```

2021-01-13T20:18:50.3548316Z 45/79 Test #260: code_api|tool.drcachesim.coherence ...........................***Failed  Required regular expression not found. Regex=[^
2021-01-13T20:18:50.3550127Z     -------------------------------------------------------------------
2021-01-13T20:18:50.3551085Z      Performance for solving AX=B Linear Equation using Jacobi method
2021-01-13T20:18:50.3551884Z      Running on DynamoRIO
2021-01-13T20:18:50.3552456Z      Client version .*
2021-01-13T20:18:50.3553016Z     ...................................................................
2021-01-13T20:18:50.3553418Z 
2021-01-13T20:18:50.3553896Z      Matrix Size :  64
2021-01-13T20:18:50.3554409Z      Threads     :  4
2021-01-13T20:18:50.3554779Z 
2021-01-13T20:18:50.3555104Z 
2021-01-13T20:18:50.3555724Z      Started iteration 1 of the computation...
2021-01-13T20:18:50.3556245Z 
2021-01-13T20:18:50.3556918Z      Finished computing current solution distance in mode 0.
2021-01-13T20:18:50.3557643Z      Mode changed to 0.
2021-01-13T20:18:50.3558034Z 
2021-01-13T20:18:50.3558634Z      Started iteration 2 of the computation...
2021-01-13T20:18:50.3559152Z 
2021-01-13T20:18:50.3559817Z      Finished computing current solution distance in mode 0.
2021-01-13T20:18:50.3560541Z      Mode changed to 0.
2021-01-13T20:18:50.3561127Z 
2021-01-13T20:18:50.3561765Z      Started iteration 3 of the computation...
2021-01-13T20:18:50.3562285Z 
2021-01-13T20:18:50.3562942Z      Finished computing current solution distance in mode 0.
2021-01-13T20:18:50.3563653Z      Mode changed to 0.
2021-01-13T20:18:50.3564046Z 
2021-01-13T20:18:50.3564372Z 
2021-01-13T20:18:50.3564924Z      The Jacobi Method For AX=B .........DONE
2021-01-13T20:18:50.3565597Z      Total Number Of iterations   :  3
[...]

Issue #4167, opened 2020-03-06
tool.drcachesim.coherence test failed on Travis: "Invalid trace entry type 23 before a bundle"


derekbruening:
This happened once on Travis:
https://travis-ci.com/DynamoRIO/dynamorio/jobs/295374252
```
Invalid trace entry type 23 before a bundle
```

23 is (+1 here b/c 1-based):
```
$ grep '^ *TRACE_TYPE' clients/drcachesim/common/trace_entry.h | grep -n . | grep 24
24:    TRACE_TYPE_THREAD_EXIT, /**< A thread exit. */
```

derekbruening:
Happened on `code_api|tool.drcachesim.threads-with-config-file` at http://139.178.82.61:8080/job/DynamoRIO-AArch64-Precommit/2003/consoleFull

brettcoon:
The same issue seems to have happened on `code_api|tool.drcachesim.TLB-threads` at [http://139.178.82.61:8080/job/DynamoRIO-AArch64-Precommit/2744/consoleFull](http://139.178.82.61:8080/job/DynamoRIO-AArch64-Precommit/2744/consoleFull)

derekbruening:
I believe this is the same as #3320.  Closing.

Issue #3971, opened 2019-11-26
tool.drcachesim.coherence timeout on appveyor


derekbruening:
This happened once on Appveyor:

https://ci.appveyor.com/project/DynamoRIO/dynamorio/builds/29130585

```
[00:12:07]         Start 215: code_api|tool.drcachesim.coherence
[00:12:07] 215: Test command: C:\projects\dynamorio\build\build_debug-internal-32\bin32\drrun.exe "-s" "150" "-quiet" "-debug" "-use_dll" "C:/projects/dynamorio/build/build_debug-internal-32/lib32/debug/dynamorio.dll" "-exit0" "-stderr_mask" "0xC" "-msgbox_mask" "0" "-dumpcore_mask" "0x7d" "-staged" "-code_api" "-t" "drcachesim" "-ipc_name" "drtestpipe_coherence" "-coherence" "--" "C:/projects/dynamorio/build/build_debug-internal-32/suite/tests/bin/client.annotation-concurrency.exe" "client.annotation-concurrency.appdll.dll" "A" "4"
[00:12:07] 215: Test timeout computed to be: 150
[00:12:07] 215: 
[00:12:07] 215:     -------------------------------------------------------------------
[00:12:07] 215:      Performance for solving AX=B Linear Equation using Jacobi method
[00:12:07] 215:      Running on DynamoRIO
[00:12:07] 215:      Client version (null)
[00:12:07] 215:     ...................................................................
[00:12:07] 215: 
[00:12:07] 215:      Matrix Size :  512
[00:12:07] 215:      Threads     :  4
[00:12:07] 215: 
[00:12:07] 215: 
[00:12:07] 215:      Started iteration 1 of the computation...
[00:12:07] 215: 
[00:12:07] 215:      Finished computing current solution distance in mode 0.
[00:12:07] 215:      Mode changed to 0.
[00:12:07] 215: 
[00:12:07] 215:      Started iteration 2 of the computation...
[00:12:07] 215: 
[00:12:07] 215:      Finished computing current solution distance in mode 0.
[00:12:07] 215:      Mode changed to 0.
[00:12:07] 215: 
[00:12:07] 215:      Started iteration 3 of the computation...
[00:12:07] 215: 
[00:12:07] 215:      Finished computing current solution distance in mode 0.
[00:12:07] 215:      Mode changed to 0.
[00:12:31] 215: 
[00:12:31] 215:      Started iteration 4 of the computation...
[00:12:31] 215: 
[...]

PR #5049, opened 2021-08-12
i#2417: Add drcachesim.coherence to the aarch64 flaky list

joshua-warburton:
This test always fails on certain types of aarch64 hardware.

### DIFF
--- a/suite/runsuite_wrapper.pl
+++ b/suite/runsuite_wrapper.pl
@@ -321,6 +321,7 @@ for (my $i = 0; $i <= $#lines; ++$i) {
                                    'code_api|tool.drcachesim.threads' => 1, # i#4928
                                    'code_api,tracedump_text,tracedump_origins,syntax_intel|common.loglevel' => 1, # i#1807
                                    'code_api|tool.drcachesim.threads-with-config-file' => 1, # i#4954
+                                   'code_api|tool.drcachesim.coherence' => 1, # i#2417
                                    );
             if ($is_32) {
                 $issue_no = "#2416";


suite/runsuite_wrapper.pl
```diff
@@ -321,6 +321,7 @@
                                    'code_api|tool.drcachesim.threads' => 1, # i#4928
                                    'code_api,tracedump_text,tracedump_origins,syntax_intel|common.loglevel' => 1, # i#1807
                                    'code_api|tool.drcachesim.threads-with-config-file' => 1, # i#4954
+                                   'code_api|tool.drcachesim.coherence' => 1, # i#2417
```
derekbruening, line 324: I don't see this test mentioned in #2417.  If you could look at the reason it's failing and paste that output into the issue (timeout, output mismatch, assert, crash).  Or is this #4673, or #4167, or #3971?

derekbruening (approved):

suite/runsuite_wrapper.pl
derekbruening: In general please reply to each review request ("Done", or somesuch) and mark as Resolve if resolved.  It looks like this was ignored -- if there's a disagreement please reply and we can discuss.

joshua-warburton: Sorry about that, thought I had replied. 

It appears to be hitting the timeout repeatedly.

derekbruening: Please add that information to #2417

joshua-warburton: added

PR #5058, opened 2021-08-19
i#2626: AArch64 v8.0 decode: add frintz and cmgt

joshua-warburton:
This patch adds
`FRINTZ <Vd>.<T>, <Vn>.<T>`
`CMGT <V><d>, <V><n>, <V><m>`

issue: #2626


AssadHashmi (approved):

petrochenkov:
@joshua-warburton @AssadHashmi 
Could you double-check correctness of the encodings added here?
I bisected an application crash under DR to this commit (although the crash is probabilistic, so I'm not entirely sure).

joshua-warburton:
I've double checked this against the v8 spec and nothing seems to be amiss. The tests cover these encodings well, so I would be surprised if the issue was in this commit. You may be hitting a deeper issue in the codec.

If you do manage to create a reproducer we're more than happy to look at it.

petrochenkov:
False alarm, I've observed the same crash without this change.
Apparently the crash depends on some random unknown factors that happened to correlate with my bisection process yesterday.

PR #5059, opened 2021-08-23
i#4678: a new meta-instrumentation phase in drmgr

sapostolakis:
Introduces a new fifth instrumentation phase in drmgr (called meta-instrumentation)
that executes after the insertion of instrumentation and instrumentation optimizations. 
Its primary purpose is to enable debugging of instrumentation sequences and detection 
of instrumentation-induced bugs.

Issue: #4678


ext/drmgr/drmgr.h
```diff
@@ -216,6 +216,7 @@ typedef enum {
     DRMGR_PHASE_ANALYSIS,      /**< Currently in the analysis phase. */
     DRMGR_PHASE_INSERTION,     /**< Currently in the instrumentation insertion phase. */
     DRMGR_PHASE_INSTRU2INSTRU, /**< Currently in the instru2instru phase. */
+    DRMGR_PHASE_POST_INSTRU,   /**< Currently in the post instrumentation phase. */
```
abhinav92003, line 219: I'm rethinking whether we should name it "post instru". Post-instrumentation could refer to any phase after `DRMGR_PHASE_INSERTION`. Maybe a name that describes the phase itself, rather than its placement would be better.

ext/drmgr/drmgr.h
```diff
@@ -515,6 +515,79 @@ bool
...
+ * five of them.  See the documentation for drmgr_register_bb_app2app_event(),
+ * drmgr_register_bb_instrumentation_event(), drmgr_register_bb_instru2instru_event(), and
+ * drmgr_register_bb_post_instru_event() for further details of each pass. The
+ * aforementioned routines are identical to this with the exception of the extra \p
+ * user_data parameter, which is an OUT parameter to the \p app2app_func and passed in to
+ * the four subsequent callbacks. The \p priority param can be NULL, in which case a
+ * default priority is used.
+ */
+bool
+drmgr_register_bb_instrumentation_all_events(drmgr_app2app_ex_cb_t app2app_func,
```
abhinav92003, line 567: I'm skeptical about naming this "*_all_events". If we add another phase in future, it will be confusing and we wouldn't be able to rename later due to issues with binary compatibility.

ext/drmgr/drmgr.h
```diff
@@ -216,6 +216,7 @@ typedef enum {
     DRMGR_PHASE_ANALYSIS,      /**< Currently in the analysis phase. */
     DRMGR_PHASE_INSERTION,     /**< Currently in the instrumentation insertion phase. */
     DRMGR_PHASE_INSTRU2INSTRU, /**< Currently in the instru2instru phase. */
+    DRMGR_PHASE_POST_INSTRU,   /**< Currently in the post instrumentation phase. */
 } drmgr_bb_phase_t;
```
abhinav92003, line 220: Can you also add unit tests in drmgr-test?

ext/drmgr/drmgr.h
```diff
@@ -515,6 +515,79 @@ bool
...
+ * user_data parameter passed among them all, enabling data sharing for all
+ * five of them.  See the documentation for drmgr_register_bb_app2app_event(),
+ * drmgr_register_bb_instrumentation_event(), drmgr_register_bb_instru2instru_event(), and
+ * drmgr_register_bb_post_instru_event() for further details of each pass. The
+ * aforementioned routines are identical to this with the exception of the extra \p
+ * user_data parameter, which is an OUT parameter to the \p app2app_func and passed in to
+ * the four subsequent callbacks. The \p priority param can be NULL, in which case a
+ * default priority is used.
+ */
+bool
```
abhinav92003, line 566: Do we need `drmgr_register_bb_instrumentation_all_events`? If the user wants the fifth event, maybe they can call `drmgr_register_bb_post_instru_event` separately on their own?

ext/drmgr/drmgr.h
abhinav92003: Maybe instru_debug or something?

abhinav92003 (commented):
Thanks for the PR. Some early comments.

ext/drmgr/drmgr.c
```diff
@@ -1668,6 +1705,75 @@ drmgr_unregister_opcode_instrumentation_event(drmgr_opcode_insertion_cb_t func,
...
+            app2app_func, analysis_func, insertion_func, instru2instru_func, priority))
+        return false;
+
+    if (post_instru_func == NULL)
+        return false;
+
+    bool ok = true;
+    if (post_instru_func != NULL) {
+        ok = drmgr_bb_cb_add(&cblist_post_instru, (void *)post_instru_func, NULL,
+                             priority, NULL /* no user data */,
```
abhinav92003, line 1746: You mentioned that we need `drmgr_register_bb_instrumentation_all_events` as we want to pass user data between all five stages. But here it says "no user data". Am I understanding something wrong?

ext/drmgr/drmgr.h
abhinav92003: > Went with post-instru because in early review stages someone mentioned that it might be more than just for debugging, 

Can you remind me what additional uses we thought the additional phase can support? Maybe we can name "meta instru" or something.

ext/drmgr/drmgr.h
abhinav92003: > Derek also mentioned the convention to use an _ex suffix when we later extend an API we've already released. So, there is a process to adjust for released API.

I think the _ex suffix allows us to add a new version of a function, for supporting a different signature. But the old versions still have to stay, for binary compatibility (we cannot remove the old functions or users using it in their clients will see errors). So the `drmgr_register_bb_instrumentation_all_events` with 5 callbacks cannot be removed, which can be confusing because "all events" may not always mean 5 callbacks.

Another possible way to add `drmgr_register_bb_instrumentation_all_events` is by taking as argument a struct for all callbacks. This way, even if we add a new phase in future, the struct arg will never need to change, we'll just add a new member to the end of the struct.

derekbruening: +1 to using a struct: it makes backward compatibility much easier for function argument changes.

> Not sure if the struct would work 

See how we use structs throughout the API: a size field as the first field provides the versioning and we can then append fields in future versions without breaking anything.

sapostolakis: Right, a size field would resolve the concern I had. Changed it to use a struct.

suite/tests/client-interface/drmgr-test.dll.c
```diff
@@ -761,6 +795,21 @@ event_bb4_instru2instru(void *drcontext, void *tag, instrlist_t *bb, bool for_tr
     return DR_EMIT_DEFAULT;
 }
 
+static dr_emit_flags_t
+event_bb5_post_instru(void *drcontext, void *tag, instrlist_t *bb, bool for_trace,
+                      bool translating, void *user_data)
+{
+    CHECK(user_data == (void *)((ptr_uint_t)tag + 1), "user data not preserved");
```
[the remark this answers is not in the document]
abhinav92003: > but cannot test the counter in the event_exit

Instead of using a counter in `user_data`, maybe we can increment some global counter in each event? If there's some bug due to which the meta_instru callback isn't being invoked, it'd be good if the unit tests can detect that.

abhinav92003 (approved):
LGTM. As this PR introduces a big user-visible feature, maybe wait for Derek's review too before submitting.

derekbruening:
> maybe wait for Derek's review too before submitting.

Seeing unresolved prior comment threads so not everything is settled?

abhinav92003:
> > maybe wait for Derek's review too before submitting.
> 
> Seeing unresolved prior comment threads so not everything is settled?

Resolved the ones that are settled.

ext/drmgr/drmgr.h
derekbruening: Resolve?

ext/drmgr/drmgr.h
```diff
@@ -209,13 +209,30 @@ typedef struct _drmgr_priority_t {
     int priority;
 } drmgr_priority_t;
 
+/** Specifies the callbacks when registering all \p drmgr's events */
```
derekbruening, line 212: drmgr mediates many more events than just basic block instrumentation.  s/events/basic block instrumentation events/ or sthg.

ext/drmgr/drmgr.h
```diff
@@ -209,13 +209,30 @@ typedef struct _drmgr_priority_t {
...
+typedef struct _drmgr_events_t {
```
derekbruening, line 213: Similarly s/events/instru_events/ or sthg

ext/drmgr/drmgr.h
```diff
@@ -209,13 +209,30 @@ typedef struct _drmgr_priority_t {
...
+    /** The size of the drmgr_events_t struct */
```
derekbruening, line 214: nit: Use end punctuation on comments esp doxygen user-facing ones like these

ext/drmgr/drmgr.h
```diff
@@ -209,13 +209,30 @@ typedef struct _drmgr_priority_t {
...
+    size_t struct_size;
+    /** Callback for the app2app event */
```
derekbruening, line 216: Maybe link to the main docs for each of these?

derekbruening (approved):
(Only skimmed some parts since already reviewed)

ext/drmgr/drmgr.h
sapostolakis: Done.

ext/drmgr/drmgr.h
sapostolakis: Done.

ext/drmgr/drmgr.h
sapostolakis: Done.

suite/tests/client-interface/drmgr-test.dll.c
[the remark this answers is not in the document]
sapostolakis: Sure. Added a global counter and checked on exit.

ext/drmgr/drmgr.h
sapostolakis: Yep resolved. Use of the \*all_events\* is now okay given the use of the struct that allows for backward compatibility.

ext/drmgr/drmgr.h
sapostolakis: Not sure which docs you refer to. I would only think of https://dynamorio.org/page_drmgr.html#sec_drmgr_stages as a generic reference to the stages and I cannot think of any documentation for each individual stage. From other code, I would only think to point to the individual register calls (e.g., drmgr_register_bb_app2app_event)

derekbruening: I'm thinking of the individual register calls.  But if you think these are well enough documented in the function that takes in this struct then could leave as is.

sapostolakis: Thanks for quickly clarifying. The function that uses the struct indeed points to the documentation for the individual register calls. So that should probably be sufficient.

PR #5068, opened 2021-09-01
i#5067: Do not skip version in view tool

### LINKED ISSUE #5067: drcachesim view tool skips version with -skip_refs
The drcachesim view tool incorrectly skips over key metadata like the version when -skip_refs is used.  This causes it to misinterpret trace data as shown here where it thinks the kernel xfer value is an offset rather than absolute:

```
$ bin64/drrun -t drcachesim -simulator_type view -sim_refs 10 -indir drmemtrace.linux.infloop.*.dir/ 
T444096 <marker: version 3>

$ bin64/drrun -t drcachesim -simulator_type view -skip_refs 44000111 -indir drmemtrace.linux.infloop.*.dir/ 
T444096   0x00007f3e83b0b7f2  48 83 45 f0 01       addq   $0x01, -0x10(%rbp)
T444096     read  8 byte(s) @ 0x7ffc4c720190
T444096 <marker: kernel xfer from module offset +0x7f3e83b0b7f2 to handler>
```

derekbruening:
When -skip_refs is passed, the view tool still needs to look at the
initial metadata to obtain the trace version and file type.

Tested:
Before:
  $ bin64/drrun -t drcachesim -offline -- suite/tests/bin/linux.signest
  $ bin64/drrun -t drcachesim -simulator_type view -skip_refs 100 -indir drmemtrace.linux.signest.*.dir/ 2>&1 | egrep 'version|kernel xfer'
  T1602962 <marker: version -1>
  T1602964 <marker: version -1>
  T1602964 <marker: kernel xfer from module offset +0x7fc58e2bd212 to handler>
  T1602964 <marker: kernel xfer from module offset +0x7fc58e2bd212 to handler>
  T1602964 <marker: kernel xfer from module offset +0x7fc58e2bd212 to handler>
  T1602964 <marker: kernel xfer from module offset +0x7fc58e2b87b2 to handler>
After:
  $ bin64/drrun -t drcachesim -simulator_type view -skip_refs 100 -indir drmemtrace.linux.signest.*.dir/ 2>&1 | egrep 'version|kernel xfer'
  T1602962 <marker: version 3>
  T1602964 <marker: version 3>
  T1602964 <marker: kernel xfer from 0x7fc58e2bd212 to handler>
  T1602964 <marker: kernel xfer from 0x7fc58e2bd212 to handler>
  T1602964 <marker: kernel xfer from 0x7fc58e2bd212 to handler>
  T1602964 <marker: kernel xfer from 0x7fc58e2b87b2 to handler>

Fixes #5067

### DIFF
--- a/clients/drcachesim/tools/view.cpp
+++ b/clients/drcachesim/tools/view.cpp
@@ -103,31 +103,8 @@ view_t::initialize()
 bool
 view_t::process_memref(const memref_t &memref)
 {
-    if (instr_count_ < knob_skip_refs_ ||
-        instr_count_ >= (knob_skip_refs_ + knob_sim_refs_)) {
-        if (type_is_instr(memref.instr.type) ||
-            memref.data.type == TRACE_TYPE_INSTR_NO_FETCH)
-            ++instr_count_;
-        return true;
-    }
-
-    if (memref.instr.tid != 0) {
-        if (prev_tid_ != -1 && prev_tid_ != memref.instr.tid)
-            std::cerr << "------------------------------------------------------------\n";
-        prev_tid_ = memref.instr.tid;
-        std::cerr << "T" << memref.instr.tid << " ";
-    }
-
+    // Even for -skip_refs we need to process the up-front version and type.
     if (memref.marker.type == TRACE_TYPE_MARKER) {
-        if (memref.marker.tid != 0 &&
-            printed_header_.find(memref.marker.tid) == printed_header_.end()) {
-            printed_header_.insert(memref.marker.tid);
-            std::cerr << "<marker: version " << trace_version_ << ">\n";
-            std::cerr << "T" << memref.instr.tid << " ";
-            std::cerr << "<marker: filetype 0x" << std::hex << prev_filetype_ << std::dec
-                      << ">\n";
-            std::cerr << "T" << memref.instr.tid << " ";
-        }
         switch (memref.marker.marker_type) {
         case TRACE_MARKER_TYPE_VERSION:
             // We delay printing until we know the tid.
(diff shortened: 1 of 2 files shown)


clients/drcachesim/tools/view.cpp
```diff
@@ -118,37 +146,26 @@ view_t::process_memref(const memref_t &memref)
         std::cerr << "T" << memref.instr.tid << " ";
     }
 
-    if (memref.marker.type == TRACE_TYPE_MARKER) {
-        if (memref.marker.tid != 0 &&
-            printed_header_.find(memref.marker.tid) == printed_header_.end()) {
-            printed_header_.insert(memref.marker.tid);
+    if (memref.marker.type == TRACE_TYPE_MARKER ||
+        (knob_sim_refs_ > 0 && instr_count_ == knob_skip_refs_)) {
```
abhinav92003, line 150: Why did we need this additional condition? Is it for the case when we don't have any markers in the [knob_skip_refs_, knob_skip_refs_ + knob_sim_refs_) instr range?

clients/drcachesim/tools/view.cpp
```diff
@@ -118,37 +146,26 @@ view_t::process_memref(const memref_t &memref)
...
+        if (memref.instr.tid != 0 &&
```
abhinav92003, line 151: Why did we have to change `memref.marker.tid` to `memref.instr.tid`? I thought that we can use any `memref.*.tid` to get the tid as it has the same location in all structs in the `_memref_t` union.

clients/drcachesim/tools/view.cpp
```diff
@@ -103,6 +103,34 @@ view_t::initialize()
...
+            if (trace_version_ == -1)
+                trace_version_ = static_cast<int>(memref.marker.marker_value);
+            else if (trace_version_ != static_cast<int>(memref.marker.marker_value)) {
+                error_string_ = std::string("Version mismatch across files");
+                return false;
+            }
+            break;
+        case TRACE_MARKER_TYPE_FILETYPE:
+            // We delay printing until we know the tid.
+            prev_filetype_ = memref.marker.marker_value;
```
abhinav92003, line 120: Why don't we have a check like above that fails if a different file type is found later? I thought we can have only one file type in a trace.

abhinav92003 (approved):

clients/drcachesim/tools/view.cpp
derekbruening: Yes, exactly.  Adding a coment.

clients/drcachesim/tools/view.cpp
derekbruening: Just for consistency: the other .tid references in this code block are using instr.  Maybe keeping these and changing those others to marker is better though since the if() here will often match a marker -- making that change.

clients/drcachesim/tools/view.cpp
derekbruening: You mean like the version check?  Done.

abhinav92003: Yes, thanks.

Issue #5076, opened 2021-09-02
Generalize and parallelize trace_invariants test into a first-class tool


derekbruening:
The drcachesim trace_invariants test performs sanity checks for various properties of traces, ensuring we have no problems in collection or post-processing that leave PC gaps or violate the promises we make to core simulators.  What we'd like to do is be able to run it on large traces by generalizing and speeding it up:

- Add parallel operation support.  It can still check the thread transition invariants by looking for timestamps and assuming another thread could interleave there.
- Generalize error reporting to be customizable instead of using `assert`.
- Report further details on where the error was found: maybe the instruction count for use with the `view` tool and `-skip_refs`.
- Add single-thread support to the `view` tool to speed up locating the error?  So we'd report the instruction count for that thread which is all we'd have for parallel operation anyway.
- Generalize or put under options the current checks that rely on annotations inside the `signal_invariants` and `filter_asm` test binaries.
- Add checks for nested signals (to catch #4998) under an option since some apps do expect them.

derekbruening:
+ Move from tests/ to tools/?

derekbruening:
This is now complete.  We have incorporated the new first-class tool into regular analysis runs and it is identifying a number of invariant violations on routine large traces which need to be investigated; but those errors will be filed separately.

PR #5077, opened 2021-09-02
i#5076 trace invariants, part 1: Generalize error reporting

derekbruening:
Replaces the use of assert() with a virtual function report_if_false()
so that subclasses can set up custom reporting.

Generalizes the includes for easier use in other build frameworks.

Issue: #5076

### DIFF
--- a/clients/drcachesim/tests/trace_invariants.h
+++ b/clients/drcachesim/tests/trace_invariants.h
@@ -36,8 +36,8 @@
 #ifndef _TRACE_INVARIANTS_H_
 #define _TRACE_INVARIANTS_H_ 1
 
-#include "../analysis_tool.h"
-#include "../common/memref.h"
+#include "analysis_tool.h"
+#include "memref.h"
 #include <stack>
 #include <unordered_map>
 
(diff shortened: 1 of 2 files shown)


clients/drcachesim/tests/trace_invariants.h
```diff
@@ -36,8 +36,8 @@
 #ifndef _TRACE_INVARIANTS_H_
 #define _TRACE_INVARIANTS_H_ 1
 
-#include "../analysis_tool.h"
```
abhinav92003, line 39: Did this file move? And that's why changing the paths to the included files works? The Github review UI doesn't show a move here though (don't remember how that looks like).

abhinav92003 (approved):

clients/drcachesim/tests/trace_invariants.h
derekbruening: Nothing moved: the build has include dir flags that point to all the directories with headers, so the compiler finds this without the explicit path (which makes it hard to embed or use in other build systems).

Issue #5070, opened 2021-09-01
CRASH: AArch64 decoder cannot handle certain combinations of encoding definitions.


AssadHashmi:
The `codec.py` decode generator is flawed in that it can't handle certain encodings due to an overlap between operands and opcodes.

Specifically, cases where the **ONLY** difference between two encoding definitions is some part of an operand field and some part of an opcode field.

For example, when adding:
`0x0011110xxxxxxx111111xxxxxxxxxx  n     fcvtzs    dq0 : dq5 sd_sz immhb`

The following encoding overlap error is detected:
```
$ ./codec.py `pwd`/codec.txt `pwd` 
Traceback (most recent call last):
  File "./codec.py", line 489, in <module>
    main()
  File "./codec.py", line 477, in main
    consistency_check(patterns, opndtab)
  File "./codec.py", line 420, in consistency_check
    pattern_to_str(*patterns[i])))
Exception: Overlapping patterns:
0x00111100000xxx111111xxxxxxxxxx fmov dq0 : fpimm8 h_sz
0x0011110xxxxxxx111111xxxxxxxxxx fcvtzs dq0 : dq5 sd_sz immhb
```

Similarly when adding:
`0x00111100000xxxxxxx01xxxxxxxxxx  n     movi     dq0 : imm8 cmode4`
```
$ ./codec.py `pwd`/codec.txt `pwd` 
Traceback (most recent call last):
  File "./codec.py", line 489, in <module>
    main()
  File "./codec.py", line 477, in main
    consistency_check(patterns, opndtab)
  File "./codec.py", line 420, in consistency_check
    pattern_to_str(*patterns[i])))
Exception: Overlapping patterns:
0x00111100000xxxxxxx01xxxxxxxxxx movi dq0 : imm8 cmode4
0x00111100000xxxxxx101xxxxxxxxxx orr dq0 : imm8 cmode3
```
`codec.py` needs to be able to generate correct decode logic for such overlaps.

AssadHashmi:
Further testing and problems adding new instructions has shown that further work is needed for a more robust and scalable solution.

PR #5082, opened 2021-09-09
i#5070: AArch64 decode: fixes generated decoder function names

AssadHashmi:
This patch fixes the inability of codec.py to correctly
generate decode function names with sufficient resolution
to deal with variants of the same opcode.

Issues: #5070, #2626


joshua-warburton (approved):

AssadHashmi:
run arm tests

PR #5084, opened 2021-09-09
i#2626: AArch64 v8.0 decode: Add SQSHL, SHL and SLI

joshua-warburton:
This patch implements the instructions:
`SQSHL <V><d>, <V><n>, #<shift>`
`SHL <Vd>.<T>, <Vn>.<T>, #<shift>`
`SLI <Vd>.<T>, <Vn>.<T>, #<shift>`

To implement these instructions an additional
operand was necessary to allow a zero-indexed
shift. The appropriate encode tests for an
instruction using this immhb_0shf operand have
been added.

Issue: #2626


AssadHashmi (changes requested):
There's a lot of duplication/common functionality in the `decode_opnd_immhb_` and `encode_opnd_immhb_` functions. Can you factor this out into one function and call that?

joshua-warburton:
> There's a lot of duplication/common functionality in the `decode_opnd_immhb_` and `encode_opnd_immhb_` functions. Can you factor this out into one function and call that?

Done

AssadHashmi (approved):

Issue #5085, opened 2021-09-09
APP CRASH on sysenter after vsyscall hook fails in 32-bit apps on Ubuntu20


derekbruening:
Running any Dr. Memory test app, including tests/hello, and presumably any 32-bit app at all (but then why doesn't every single test fail in the DR and DrM suites??), we have an app crash:

```
bruening@ubuntu:~/drmemory/git/build_x86_dbg_tests$ dynamorio/bin32/drrun -- tests/hello
<Starting application /home/bruening/drmemory/git/build_x86_dbg_tests/tests/hello (35741)>
<Initial options = -no_dynamic_options -code_api -stack_size 56K -signal_stack_size 32K -max_elide_jmp 0 -max_elide_call 0 -early_inject -emulate_brk -no_inline_ignored_syscalls -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct >
<Paste into GDB to debug DynamoRIO clients:
set confirm off
add-symbol-file '/home/bruening/drmemory/git/build_x86_dbg_tests/dynamorio/lib32/debug/libdynamorio.so' 0xf7b9f000
>
<spurious rep/repne prefix @0xf7b48dd0 (f3 0f 1e fb): >
<(1+x) Handling our fault in a TRY at 0xf7db56fd>
<Application tried to execute from unreadable memory 0x00000000.
This may be a result of an unsuccessful attack or a potential application vulnerability.>
<Application /home/bruening/drmemory/git/build_x86_dbg_tests/tests/hello (35741).  Application exception at PC 0xf7b7e549.  
Signal 11 delivered to application as default action.
Callstack:
	0xf7b7e549   <[vdso]+0x549>
>
<Stopping application /home/bruening/drmemory/git/build_x86_dbg_tests/tests/hello (35741)>
Segmentation fault (core dumped)
```

```
interp: start_pc = 0xf7a251dd
check_thread_vm_area: pc = 0xf7a251dd
prepend_entry_to_fraglist: putting fragment @0xf7a251dd (shared) on vmarea 0xf794e000-0xf7aa6000
check_thread_vm_area: check_stop = 0xf7aa6000
  0xf7a251dd  8b 4c 24 20          mov    0x20(%esp)[4byte] -> %ecx
  0xf7a251e1  b8 27 01 00 00       mov    $0x00000127 -> %eax
  0xf7a251e6  bb 9c ff ff ff       mov    $0xffffff9c -> %ebx
[...]

PR #5087, opened 2021-09-10
i#5085 vsys hook: Handle hook failure in SA_RESTART adjustment

derekbruening:
Adds handling of the vsyscall hook failing by avoiding trying to
set the PC to the (zero) displacement PC in such a case.

Tested on a 5.11 kernel where 32-bit hello,world under DR crashes
without this fix but works with the fix.

Issue #5085

### DIFF
--- a/core/arch/interp.c
+++ b/core/arch/interp.c
@@ -244,8 +244,14 @@ init_build_bb(build_bb_t *bb, app_pc start_pc, bool app_interp, bool for_cache,
      * whose fall-through hits our hook.  We avoid interpreting our own hook
      * by shifting it to the displaced pc.
      */
-    if (DYNAMO_OPTION(hook_vsyscall) && start_pc == vsyscall_sysenter_return_pc)
-        start_pc = vsyscall_sysenter_displaced_pc;
+    if (DYNAMO_OPTION(hook_vsyscall) && start_pc == vsyscall_sysenter_return_pc) {
+        if (vsyscall_sysenter_displaced_pc != NULL)
+            start_pc = vsyscall_sysenter_displaced_pc;
+        else {
+            /* Our hook must have failed. */
+            ASSERT(should_syscall_method_be_sysenter());
+        }
+    }
 #endif
     bb->check_vm_area = true;
     bb->start_pc = start_pc;


derekbruening:
Unfortunately, adding a `-no_hook_vsyscall` test does not reproduce the problem on a different machine: it relies on a specific config that's not easy to add as a regression test.

core/arch/interp.c
```diff
@@ -244,8 +244,14 @@ init_build_bb(build_bb_t *bb, app_pc start_pc, bool app_interp, bool for_cache,
...
      * by shifting it to the displaced pc.
      */
-    if (DYNAMO_OPTION(hook_vsyscall) && start_pc == vsyscall_sysenter_return_pc)
-        start_pc = vsyscall_sysenter_displaced_pc;
+    if (DYNAMO_OPTION(hook_vsyscall) && start_pc == vsyscall_sysenter_return_pc) {
+        if (vsyscall_sysenter_displaced_pc != NULL)
+            start_pc = vsyscall_sysenter_displaced_pc;
+        else {
+            /* Our hook must have failed. */
+            ASSERT(should_syscall_method_be_sysenter());
```
abhinav92003, line 252: So the 32-bit tests on the new kernel hit this case? Won't they fail in debug build?

core/arch/interp.c
```diff
@@ -244,8 +244,14 @@ init_build_bb(build_bb_t *bb, app_pc start_pc, bool app_interp, bool for_cache,
      * whose fall-through hits our hook.  We avoid interpreting our own hook
```
abhinav92003, line 244: This solves only the test hello world app right? Do any signal tests work? I assume no.

abhinav92003 (approved):

core/arch/interp.c
derekbruening: This solves every single app crashing (i.e., every single test failing).  Some tests still fail for other reasons like many (not all) of the signal-related tests -- probably matching GA CI.

core/arch/interp.c
derekbruening: Not sure I follow: why would they fail?  I'm testing debug build and it works for the simple tests like broadfun.

abhinav92003: Sorry, please ignore. I think I misread the assert.

PR #5090, opened 2021-09-12
i#4678: DrStateCmp, a library for machine state comparison.

sapostolakis:
Implements drstatecmp, a library for machine state comparison for detecting 
instrumentation-induced clobbering of application state. 

This implementation supports checks for side-effect-free basic blocks (with the 
exception of the terminating instruction that could be a function call with side effects). 
The check involves executing two versions of these basic blocks, one with 
instrumentation and one without, and then comparing the resulting state to detect bugs. 
When a mismatch is detected, by default, the disassembled basic block that involved 
the bug will be emitted.

Illustrates the use of the drstatecmp library by providing a sample client. 
This sample client introduces an instrumentation bug that is caught by drstatecmp. 
This sample client also shows how to specify a user-defined callback to be invoked on 
state comparison mismatches.

Originally part of #4938.

Issue: #4678


api/docs/release.dox
```diff
@@ -232,6 +232,8 @@ Further non-compatibility-affecting changes include:
    insertion of instrumentation and instrumentation optimizations. Its primary purpose
```
derekbruening, line 232: PR description:
> Implements drstatecmp, a library for machine state comparison for detecting instrumentation-induced clobbering of

nit: This is a really wide line width to have inserted newlines: I think people would expect either 80-char or maybe 90-char, or else no newlines at all.

api/docs/release.dox
```diff
@@ -232,6 +232,8 @@ Further non-compatibility-affecting changes include:
...
    is to enable debugging of instrumentation sequences and detection of
```
derekbruening, line 233: PR description again:
> that is caught by drstatecmp. This sample client also shows how to specify an user-defined callback to be invoked on state

grammar nit: s/an user/a user/

api/docs/release.dox
```diff
@@ -232,6 +232,8 @@ Further non-compatibility-affecting changes include:
...
    instrumentation-induced bugs.
```
derekbruening, line 234: Would it be better to have used the original PR #4938?  Is that PR being split into any further pieces or is this the rest of it after the drmgr PR?  Then we would have our review history here, and presumably Github would give us a nice "view changes since last review' or at least we could select from which commits to view and not have to re-review a whole lot of code we've already reviewed...with this I feel like I have to go manually check out both of your branches and diff them myself to see what to look at.

sapostolakis: I would have used the original PR but there are two more PRs that will come out of #4938. One with the addition of an option in drcachesim to use drstatecmp and a drcachesim delay test, and another PR for the fuzz-app and the test that comes with it.

sapostolakis: In terms of diff, this PR has essentially a copy paste of the drstatecmp files from #4938 apart from changing the calls to the modified drmgr all-events registration function (changed in #5059 and thus post-#4938).

api/docs/release.dox
sapostolakis: You are right it is too long. Changed it

api/docs/release.dox
sapostolakis: Right.

derekbruening (approved):
LGTM based on this:

> In terms of diff, this PR has essentially a copy paste of the drstatecmp files from #4938 apart from changing the calls to the modified drmgr all-events registration function (changed in #5059 and thus post-#4938).

api/docs/release.dox
derekbruening: OK, SGTM.

PR #5071, opened 2021-09-01
i#5070: AArch64 decode: fixes codec.py to handle overlapping encodings

AssadHashmi:
This patch fixes the inability of codec.py to generate decode
logic for certain combinations of encoding definitions.

Specifically, cases where the **ONLY** difference between two
encoding definitions is some part of an operand field and
some part of an opcode field. Examples are:
```
0x00111100000xxx111111xxxxxxxxxx fmov dq0 : fpimm8 h_sz
0x0011110xxxxxxx111111xxxxxxxxxx fcvtzs dq0 : dq5 sd_sz immhb
```
and
```
0x00111100000xxxxxxx01xxxxxxxxxx movi dq0 : imm8 cmode4
0x00111100000xxxxxx101xxxxxxxxxx orr dq0 : imm8 cmode3
```
Issues: #5070, #2626

PR #5092, opened 2021-09-15
i#5070: AArch64 decode: add operand constraint mask type: +

joshua-warburton:
Previous patches to deal with operand/opcode overlaps
(#5071, #5082, #5091) did not handle cases in which an opcode
matches the operand value. In such cases decode logic
generated by codec.py could not distinguish between input
encodings of different opcodes.

This patch adds the + character to operand masks which
requires that at least one of the bits in masks marked +
must be set for that instruction to be considered matching.

An example is the following where bits 19->22 are 0000 for
fmov but cannot be 0000 for fcvtzs as specified by the ISA:

0x00111100000xxx111111xxxxxxxxxx fmov dq0 : fpimm8 h_sz
0x0011110xxxxxxx111111xxxxxxxxxx fcvtzs dq0 : dq5 bhsd_immh_sz immhb_fxp

For which the immh:immb operand mask is now:
---------++++xxx---------------- immhb_fxp

Issues: #5070, #2626


make/aarch64_check_codec_order.py
```diff
@@ -73,22 +73,24 @@ def check(l1, l2):
             print('Lines {} differ: \n  >  {}\n  <  {}'.format(i, a, b))
         sys.exit(1)
 
+def normilise_plus(string):
```
AssadHashmi, line 76: Typo, should be `normalise`.

core/ir/aarch64/codec.txt
```diff
@@ -38,13 +38,14 @@
...
-# Each operand type is listed here with a mask, consisting of '-', '?' and 'x',
-# showing how the bits are used by the encoder/decoder functions,
+# Each operand type is listed here with a mask, consisting of '-', '?', 'x' and
+# '+', showing how the bits are used by the encoder/decoder functions,
 # {de,en}code_opnd_OPNDTYPE, in "codec.c". The operands are ordered by pattern.
 #
 # '-' : Bit is not handled by operand encoder or decoder.
 # '?' : Bit is read by encoder and decoder but not generated by encoder.
 # 'x' : Bit is read by decoder and generated by encoder.
+# '+' : As 'x', but at lest one of the bits marked must be set
```
AssadHashmi, line 48: Typo `lest`

AssadHashmi (changes requested):

AssadHashmi:
For maintenance and future debug reasons, the final commit message needs to be consistent with related patches, ideally referring to them and should have an example of what is being fixed
```
i#5070: AArch64 decode: add operand constraint mask type (+)

Previous patches to deal with operand/opcode overlaps
(#5071,#5082,#5091) did not handle cases in which an opcode
matches the operand value. In such cases decode logic
generated by codec.py could not distinguish between input
encodings of different opcodes.

This patch adds the + character to operand masks which
requires that at least one of the bits in masks marked +
must be set for that instruction to be considered matching.

An example is the following where bits 19->22 are 0000 for
fmov but cannot be 0000 for fcvtzs as specified by the ISA:

0x00111100000xxx111111xxxxxxxxxx fmov dq0 : fpimm8 h_sz
0x0011110xxxxxxx111111xxxxxxxxxx fcvtzs dq0 : dq5 bhsd_immh_sz immhb_fxp

For which the immh:immb operand mask is now:
---------++++xxx---------------- immhb_fxp

Issues: #5070, #2626
```

AssadHashmi (approved):

PR #5094, opened 2021-09-16
i#2243: Support non-first-segment-code in drcov

### LINKED ISSUE #2243: drcov2lcov fails to handle code in ELF segments past a gap
Xref #2213.  drmodtrack now supports ELF libraries with gaps, but while
drcov2lcov will now not think one library covers other libraries inside it,
it will miss code present in other than the first segment.  It creates a
hashtable by path and stores the size, which is now just the 1st segment
size.  We can't simply change the table to store by start address, though,
b/c we'd have to redo enumerate_line_info which walks the htable and calls
drsym_enumerate_lines() on each path, passing a single bitmap.  It will
take some work to support multiple segments.

derekbruening:
Adds support to drcov and drcov2lcov for code in a non-initial module
segment, and for gaps between segments.

Changes drcov's output to uses the module segment offset, rather than
the module base offset.  This better supports modules with code
beyond the first segment and with gaps between segments.

Updates drcov2lcov to use a separate bitmap table per segment, rather
than per module.

Fixes #2243

### DIFF
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -169,6 +169,9 @@ compatibility changes:
    drmgr_in_emulation_region(): #DR_EMULATE_IS_FIRST_INSTR.  This flag addition preserves
    binary compatibility, but source code that did not zero the structure could end
    up with an uninitialized flags field when calling drmgr_insert_emulation_start().
+ - drcov's output now uses a module segment offset, rather than a module base offset.
+   This better supports modules with code beyond the first segment and with
+   gaps between segments.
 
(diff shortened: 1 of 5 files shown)


derekbruening:
Windows fails b/c the doxygen download fails: the downloaded file is html saying `We're sorry -- the Sourceforge site is currently in Disaster Recovery mode.  Please check back later.`.  Will re-start the Windows runs tomorrow.

abhinav92003 (approved):

PR #5096, opened 2021-09-16
i#2626: AArch64 v8.0 decode: fix cmode vector shift fields

AssadHashmi:
This fixes the existing cmode3 handling and adds cmode4.

Also fixes:
```
BIC <Vd>.<T>, #<imm8>{, LSL #<amount>}
ORR <Vd>.<T>, #<imm8>{, LSL #<amount>}
```
and adds:
```
MOVI <Vd>.<T>, #<imm8>{, LSL #<amount>}
MOVI <Vd>.<T>, #<imm8>, MSL #<amount>
MVNI <Vd>.<T>, #<imm8>{, LSL #<amount>}
MVNI <Vd>.<T>, #<imm8>, MSL #<amount>
```
Issue: #2626


AssadHashmi:
Requesting review from @derekbruening in the absence of @joshua-warburton.

core/ir/aarch64/codec.c
```diff
@@ -1039,6 +1039,33 @@ encode_opnd_s_const_sz(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *en
...
+{
+    /* cmode size amounts
+     * 1110  8    0
+     * 10x0  16   0,8
+     * 0xx0  32   0,8,16,24
+     * 110x  32   8,16 (MSL, shifting ones)
+     */
+    int esize = 8;
+    int shift = 0;
+    uint64 sz_shft = ((uint64)esize << 32) | shift;
```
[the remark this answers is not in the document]
derekbruening: I guess I don't understand this: if this is the size==8 row of the table, isn't the shift amount zero?  So why is this 8<<32?

core/ir/aarch64/codec.c
```diff
@@ -1039,6 +1039,33 @@ encode_opnd_s_const_sz(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *en
...
+    *opnd = opnd_create_immed_int64(sz_shft, OPSZ_8);
+    return true;
+}
+
+static inline bool
+encode_opnd_cmode4_b_sz(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_out)
+{
+    int esize = 8;
```
derekbruening, line 1063: const

core/ir/aarch64/codec.c
```diff
@@ -1484,6 +1511,55 @@ encode_opnd_imm4(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_out)
...
+static inline bool
+encode_opnd_cmode4_s_sz_msl(uint enc, int opcode, byte *pc, opnd_t opnd,
+                            OUT uint *enc_out)
+{
+    if (!opnd_is_immed_int64(opnd))
+        return false;
+
+    int64 sz_shft = opnd_get_immed_int64(opnd);
+    int shift = (int)(sz_shft & 0xFFFFFFFF);
+    if ((shift & 0x10000000) == 0) // MSL bit
```
derekbruening, line 1542: We use TEST to check bits

core/ir/aarch64/codec.c
```diff
@@ -1484,6 +1511,55 @@ encode_opnd_imm4(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_out)
...
+        return false;
+    shift = shift & 0xFF;
+    int esize = (int)(sz_shft >> 32);
```
derekbruening, line 1545: Wait: you're storing the element size in the upper bits?  Is that an intra-DR decision and unrelated to the ISA?  Please document this.  So we're taking two different integers and merging them to store as a single operand?  Would it be simpler to use two separate operands, and that would avoid this int64 opnd type?

core/ir/opnd_shared.c
```diff
@@ -365,7 +365,6 @@ opnd_create_immed_int64(int64 i, opnd_size_t size)
 {
     opnd_t opnd;
     opnd.kind = IMMED_INTEGER_kind;
-    IF_X64(CLIENT_ASSERT(false, "32-bit only"));
```
derekbruening, line 368: But if we enable this for 64-bit builds it gets confusing as to why int64 would be used vs int.  There is a comment in the user-facing docs in opnd_api.h that says this is 32-bit only: that should match this assert.  But see above: do we need to use this?  This PR is for AArch64, for which opnd_create_immed_int() is pointer-sized: so are you thinking of re-using some of this code for AArch32 or something and that's why this _int64 is being used?  I'm confused.

core/ir/aarch64/codec.c
```diff
@@ -1039,6 +1039,33 @@ encode_opnd_s_const_sz(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *en
...
 
+/* cmode4_b_sz : Operand for byte elements' shift amount
+ */
+static inline bool
+decode_opnd_cmode4_b_sz(uint enc, int opcode, byte *pc, OUT opnd_t *opnd)
+{
+    /* cmode size amounts
+     * 1110  8    0
+     * 10x0  16   0,8
+     * 0xx0  32   0,8,16,24
```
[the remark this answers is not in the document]
AssadHashmi: Yes, I put the whole `cmode` table comment in all the `cmode` handlers. I take your point it can be confusing. Removed entries in this and other `cmode` handlers.

Also omitted to check `cmode` encoding for byte elements because the `codec.txt` definition was out-of-date :-( Now fixed.

core/ir/aarch64/codec.c
```diff
@@ -1039,6 +1039,33 @@ encode_opnd_s_const_sz(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *en
...
+     * 110x  32   8,16 (MSL, shifting ones)
```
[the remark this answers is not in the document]
AssadHashmi: As this row has been removed, the `MSL` explanation is in the relevant function below, `decode_opnd_cmode4_s_sz_msl()`.

core/ir/aarch64/codec.c
```diff
@@ -1039,6 +1039,33 @@ encode_opnd_s_const_sz(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *en
...
+     */
+    int esize = 8;
```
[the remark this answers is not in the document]
AssadHashmi: Yes, done for this and other `cmode` handlers.

core/ir/aarch64/codec.c
```diff
@@ -1039,6 +1039,33 @@ encode_opnd_s_const_sz(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *en
...
+    int shift = 0;
```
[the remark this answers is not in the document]
AssadHashmi: `shift` removed. It was initially part of a formula but as code evolved lost its commonality.

core/ir/aarch64/codec.c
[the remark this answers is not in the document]
AssadHashmi: > This is a constant: best to mark esize as const, and this var as well.

`const` modifiers added in all `cmode` handlers.

> I guess I don't understand this: if this is the size==8 row of the table, isn't the shift amount zero? So why is this 8<<32?

size and shift amount encapsulated in one int operand, see my reply below.

core/ir/aarch64/codec.c
AssadHashmi: Done.

core/ir/aarch64/codec.c
```diff
@@ -1484,6 +1511,55 @@ encode_opnd_imm4(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_out)
...
+{
+    /* cmode size amounts
+     * 1110  8    0
+     * 10x0  16   0,8
+     * 0xx0  32   0,8,16,24
+     * 110x  32   8,16 (MSL, shifting ones)
+     */
+    int cmode4 = extract_uint(enc, 12, 1);
+    int esize = 32;
+    int shift = ((cmode4 == 0) ? 8 : 16) | (1 << 28);
```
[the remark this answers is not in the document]
AssadHashmi: > I'm not following how this matches the table: so this is the bottom 2 rows, or just the bottom row? Maybe indicate that in the comment.

Just the bottom row, 32 but `MSL` rather than `LSL`. Removed other entries and added explanation for `MSL`. The `MSL` operator stands for 'Modified Shift Left', a left shift which shifts ones instead of zeros into the low order bits, unlike the more common/default `LSL` (Logical Shift Left) which shifts in zeros.

> Where does the 1<<28 come from? Shouldn't this just be the 8 or 16 if it's the bottom row?

This is the element size, 32 bits, because size and shift amount encapsulated in one int operand, see my reply below.

core/ir/aarch64/codec.c
AssadHashmi: Gah, yes! Done.

core/ir/aarch64/codec.c
AssadHashmi: Yes, agreed. This is a one-off for this particular type of encoding and temporary. Further work is required on the codec generator to be able to specify different numbers of operands and will form part of upcoming development work on codec generator improvements and preparing for future work.

core/ir/opnd_shared.c
AssadHashmi: Gah, of course! Not your confusion, mine. Reverted change and now using int operand not int64 in `cmode` functions.

AssadHashmi:
run arm tests

derekbruening:
(Note that our convention is to re-request a review when ready for another look)

core/ir/aarch64/codec.c
AssadHashmi: Added comments which explain size/shift amount operand and reference to #4393.

core/ir/aarch64/codec.c
```diff
@@ -1484,6 +1484,58 @@ encode_opnd_imm4(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_out)
...
+     * This is an MSL (Modified Shift Left). Unlike an LSL (Logical Shift
+     * Left), this left shift shifts ones instead of zeros into the low order
+     * bits.
+     *
+     * The element size and shift amount are stored as two 32 bit numbers in
+     * sz_shft. This is a workaround until issue #4393 is addressed.
+     */
+    const int cmode4 = extract_uint(enc, 12, 1);
+    const int size = 32;
+    const int shift = ((cmode4 == 0) ? 8 : 16) | (1 << 28);
```
derekbruening, line 1503: Where did this value 28 come from?  I see it called the "MSL bit" below: is that in the ISA spec or something synthesized here?  In either case, a named constant (#define or const static int) would help readability: so s/28/CMODE4_MSL_BIT/ or sthg.

core/ir/aarch64/codec.c
```diff
@@ -1484,6 +1484,58 @@ encode_opnd_imm4(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_out)
...
+static inline bool
+encode_opnd_cmode4_s_sz_msl(uint enc, int opcode, byte *pc, opnd_t opnd,
+                            OUT uint *enc_out)
+{
+    if (!opnd_is_immed_int(opnd))
+        return false;
+
+    int64 sz_shft = opnd_get_immed_int(opnd);
+    int shift = (int)(sz_shft & 0xffffffff);
+    if (!TEST(1U << 28, shift)) // MSL bit should be set
```
derekbruening, line 1518: nit: `1 << 28` above vs `1U << 28` here

core/ir/aarch64/codec.c
```diff
@@ -1484,6 +1484,58 @@ encode_opnd_imm4(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_out)
...
+        return false;
+    shift &= 0xff;
```
derekbruening, line 1520: Is this just removing the MSL bit?  Clearer to use &= ~CMODE4_MSL_BIT

core/ir/aarch64/codec.c
```diff
@@ -1505,6 +1557,51 @@ encode_opnd_extam(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_out
...
+}
+
+static inline bool
+encode_opnd_cmode_h_sz(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_out)
+{
+    if (!opnd_is_immed_int(opnd))
+        return false;
+
+    const int64 sz_shft = opnd_get_immed_int(opnd);
+    const int shift = (int)(sz_shft & 0xFF);
```
derekbruening, line 1586: Hmm here's the one-byte limit again: should it instead return false on a too-large value?  Or is this again looking for CMODE4_MSL_BIT -- but if so why is the code here not checking that bit explicitly like the others?

core/ir/aarch64/codec.c
```diff
@@ -1484,6 +1484,58 @@ encode_opnd_imm4(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_out)
...
+decode_opnd_cmode4_s_sz_msl(uint enc, int opcode, byte *pc, OUT opnd_t *opnd)
+{
+    /* cmode size shift amounts
+     * 110x  32   8,16
+     * This is an MSL (Modified Shift Left). Unlike an LSL (Logical Shift
+     * Left), this left shift shifts ones instead of zeros into the low order
+     * bits.
+     *
+     * The element size and shift amount are stored as two 32 bit numbers in
+     * sz_shft. This is a workaround until issue #4393 is addressed.
```
derekbruening, line 1499: The existing convention in the code base is to use `i#4393` so probably best to stick with that so someone searching for it won't miss this ref.  Ditto below.

derekbruening (approved):

core/ir/aarch64/codec.c
AssadHashmi: Done.

core/ir/aarch64/codec.c
AssadHashmi: Made consistent with `1U`

core/ir/aarch64/codec.c
AssadHashmi: This is just picking out the shift amount from the lowest byte. Seems to be the simplest way.

core/ir/aarch64/codec.c
AssadHashmi: As above, just picking out the shift amount from the lowest byte. Seems to be the simplest way.

core/ir/aarch64/codec.c
AssadHashmi: Done.

AssadHashmi:
run arm tests

core/ir/aarch64/codec.c
derekbruening: But it's not in the lowest byte: it's the bottom 32 bits, with the top 32 bits being the size, according to the comment above.

derekbruening: It seems there is complex structure here: it's not just two 32-bit values then; it seems to be a one-byte value, a special bit, and a size.  Maybe a struct should be defined to avoid all this bit manipulation.

Issue #4953, opened 2021-06-18
Upgrade GA CI Linux VM's to Ubuntu 18.04 for rseq testing and other features


derekbruening:
We're using the old Ubuntu 16.04 in our GA CI suite and it's too old to support rseq (4.18 is needed).
Xref https://github.com/DynamoRIO/dynamorio/pull/4929#issuecomment-864151571, https://github.com/DynamoRIO/dynamorio/pull/4929#issuecomment-864160943

This issue covers upgrading to at least 18.04.  IIRC we tried in the past and hit flaky tests and other problems?  We may hit things like #4830.

derekbruening:
Looks like we'll be forced to upgrade soon:
```
The ubuntu-16.04 environment is deprecated and will be removed on September 20, 2021. Migrate to ubuntu-latest instead. For more details see https://github.com/actions/virtual-environments/issues/32
```

derekbruening:
We decided that we should probably go straight to 20.04 rather than 18.04.

abhinav92003:
Based on https://github.com/actions/virtual-environments/issues/3287#issue-870886726, there will be two short brownout periods where Ubuntu 16.04 builds will fail.

```
The brownouts are scheduled for the following dates and times:

September 6, 2021 5:00pm UTC – 10:00pm UTC
September 14, 2021 5:00pm UTC – 10:00pm UTC
```

derekbruening:
The api.detach_state 64-bit failure:

```
<Detaching from application /home/bruening/dr/git/build_x64_dbg_tests/suite/tests/bin/api.detach_state (59264)>

Thread 2 "api.detach_stat" received signal SIGUSR2, User defined signal 2.
<Detaching from process, entering final cleanup>

Thread 2 "api.detach_stat" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff7913700 (LWP 59268)]
0x00007ffff79adf4d in _IO_str_init_static_internal (sf=sf@entry=0x7ffff7912928, ptr=ptr@entry=0x7ffff7912bb8 "", size=size@entry=15, pstart=pstart@entry=0x7ffff7912bb8 "") at strops.c:41
41	strops.c: No such file or directory.
(gdb) x/10i $pc
=> 0x7ffff79adf4d <_IO_str_init_static_internal+61>:	movaps %xmm0,(%rsp)
[...]

PR #5097, opened 2021-09-16
i#4953 ubuntu20: Update package building to use Ubuntu 20

derekbruening:
Upgrades the VM's used for package building from Ubuntu 16.04 to
Ubuntu 20.04.

Issue: #4953

### DIFF
--- a/.github/workflows/ci-package.yml
+++ b/.github/workflows/ci-package.yml
@@ -55,7 +55,7 @@ jobs:
   ###########################################################################
   # Linux x86 tarball with 64-bit and 32-bit builds:
   linux-x86:
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-20.04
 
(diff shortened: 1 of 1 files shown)


derekbruening:
Running a test package build now: https://github.com/DynamoRIO/dynamorio/actions/runs/1242068139

derekbruening:
The package succeeded: https://github.com/DynamoRIO/dynamorio/releases/tag/cronbuild-8.0.18886
I downloaded the Linux tarball and ran the commands in the README as a sanity check on my laptop and it worked; a better test would be to run on an old Linux with old libc, since that's likely where issues would come in from building on a new Linux with a new libc.  But we don't have enough developer resources to be so thorough; I would say let's rely on users on old machines to complain to figure this out.

abhinav92003 (approved):

PR #5098, opened 2021-09-16
i#4953: Upgrade some AArchXX CI jobs to use Ubuntu 20.04

abhinav92003:
Upgrades Ubuntu version in android-arm-cross-compile and a64-on-x86 CI jobs on Github to 20.04. We currently use 16.04 which is going to be deprecated on Github next week.

Issue: #4953

### DIFF
--- a/.github/workflows/ci-aarchxx.yml
+++ b/.github/workflows/ci-aarchxx.yml
@@ -160,7 +160,7 @@ jobs:
 
   # Android ARM cross-compile with gcc, no tests:
   android-arm-cross-compile:
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-20.04
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-aarchxx.yml
```diff
@@ -160,7 +160,7 @@ jobs:
```
derekbruening, line 160: > We currently use 18.04 which is going to be deprecated on Github next week.

You mean 16.04.

derekbruening, line 160: > Upgrades Ubuntu version in AArchXX CI jobs on Github to 20.04. 

So the cross-compile ones that run tests on QEMU were already on 20.04, which I did to get the latest QEMU.  So it may be best to update this and maybe the title: this is only updating Android and a64-on-x86.

derekbruening (approved):

.github/workflows/ci-aarchxx.yml
abhinav92003: Oops. Done.

abhinav92003: I updated details in the description. But made only a small change in title, to keep it short.

Issue #5101, opened 2021-09-16
ASSERT is_readable_without_exception in pcache-use test on Ubuntu 20


derekbruening:
Split from #4953.

Can repro on laptop:
```
$ rm logs/dpc-124212/*; bin64/drrun -persist -no_use_persisted -no_coarse_disk_merge -no_coarse_lone_merge -c suite/tests/bin/libclient.pcache.dll.so -- suite/tests/bin/client.pcache
$ bin64/drrun -persist -no_coarse_disk_merge -no_coarse_lone_merge -c suite/tests/bin/libclient.pcache.dll.so -- suite/tests/bin/client.pcache
```

The libc segments are all messed up with some pieces having privileges removed:
```
(gdb) bt
#0  report_dynamorio_problem (dcontext=0x0, dumpcore_flag=8, exception_addr=0x0, report_ebp=0x0, 
    fmt=0x7ffff7ec5810 "DynamoRIO debug check failure: %s:%d %s\n(Error occurred @%d frags in tid %d)") at /home/bruening/dr/git/src/core/utils.c:2107
#1  0x00007ffff7c5f457 in d_r_internal_error (file=0x7ffff7ee5780 "/home/bruening/dr/git/src/core/module_list.c", line=657, 
    expr=0x7ffff7ee5b70 "is_readable_without_exception(intersection_start, intersection_len)") at /home/bruening/dr/git/src/core/utils.c:179
#2  0x00007ffff7d3f7b3 in ensure_section_readable (module_base=0x7ffdf75f8000 "\177ELF\002\001\001\003", seg_start=0x7ffdf761d000 "\377\065\002\220\031", 
    seg_len=1355776, seg_chars=5, old_prot=0x7ffdf7be9ff4, view_start=0x7ffdf75f8000 "\177ELF\002\001\001\003", view_len=1855488)
    at /home/bruening/dr/git/src/core/module_list.c:657
#3  0x00007ffff7d3fe0a in module_calculate_digest (digest=0x7ffdf7bea258, module_base=0x7ffdf75f8000 "\177ELF\002\001\001\003", module_size=1855488, 
    full_digest=false, short_digest=true, short_digest_size=4096, sec_char_include=4294967293, sec_char_exclude=2) at /home/bruening/dr/git/src/core/module_list.c:848
[...]

PR #5103, opened 2021-09-17
i#5101 pcache: Fix is-readable assert on pcache loading

derekbruening:
Newer ld.so mprotects segments spanning a gap as no-access before it
maps any contents.  This can mean that these segments are no-access in
all_memory_areas at the time we process a loaded executable segment,
resulting in an assert when we try to compute the md5.

To solve this, we query the OS and not all_memory_areas (which is not
updated until after the pcache is loaded).  We also limit our md5 to
only executable segments, rather than non-writable segments as before,
to avoid trying to read the not-yet-mapped and no-access readonly-data
segment (previously we were computing on the wrong content in any case
before it was mapped).

Issue: #5101

### DIFF
--- a/core/arch/emit_utils_shared.c
+++ b/core/arch/emit_utils_shared.c
@@ -683,8 +683,8 @@ entrance_stub_target_tag(cache_pc stub, coarse_info_t *info)
      */
     if (info == NULL)
         info = get_stub_coarse_info(stub);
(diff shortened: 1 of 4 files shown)


derekbruening:
(The shift bug fix will be a separate PR after this is in; seemed cleaner that way (that's the revert commit).)

abhinav92003 (approved):

PR #5104, opened 2021-09-17
i#5101 pcache: Fix incorrect shift bounds

derekbruening:
Fix a bug where the coarse base shift is not applied due to checking
the module base plus region size, instead of the persisted-region
base, which when the text segment is not at the module base is not the
same thing.

Tested on the pcache-use test, which raises an application crash on
recent Linux distros without this fix.

Fixes #5101

PR #5105, opened 2021-09-17
i#4953 ubuntu20: Upgrade the x86-64 job to Ubuntu 20

derekbruening:
Upgrades the x86-64 test suite job to Ubnutu 20 now that all failures
have been addressed.
    
Issue: #4953


derekbruening:
The pcache commit will disappear from this PR after updating to include PR #5104.  Wanted to test it out on GA CI ahead of time to ensure it matched local tests.

abhinav92003 (approved):

PR #5111, opened 2021-09-20
i#4953 ubuntu20: Fix bad cast in raw2trace

derekbruening:
Fixes a raw2trace cast from 32-bit app_pc to 64-bit int that was
incorrectly sign-extending, to fix the histogram test's invariant
check failure.

Issue: #4953


clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1030,8 +1030,9 @@ template <typename T> class trace_converter_t {
             TESTANY(OFFLINE_FILE_TYPE_FILTERED, impl()->get_file_type(tls));
```
abhinav92003, line 1030: Also remove the ignore-list entry: https://github.com/DynamoRIO/dynamorio/blob/02b88bb247899af74f95a42b1f512c0a64070686/suite/runsuite_wrapper.pl#L348

derekbruening: Done.

abhinav92003 (approved):

PR #5113, opened 2021-09-21
i#4953 ubuntu20: Fix issues on common.decode test in x86-32

abhinav92003:
Fixes some issues related to preservation of clobbered regs in the
common.decode test, seen on the x86-32 bit suite on latest Ubuntu versions.

Preserves stack reg in a separate data var on the test_modrm16 test.

Preserves callee saved regs in the test_avx512_vex test.

Issue: #4953


abhinav92003:
Oops. My local workspace had some already committed stuff too. Will raise a new clean PR.

PR #5114, opened 2021-09-21
i#4953 ubuntu20: Fix issues on common.decode test in x86-32.

abhinav92003:
Fixes some issues related to preservation of regs in the common.decode
test, seen on the x86-32 bit suite on latest Ubuntu versions.

Preserves stack reg in a separate data var on the test_modrm16 test.

Preserves callee saved regs in the test_avx512_vex test.

Removes common.decode from ignore list. Verified locally that there was no
failure in 1000 runs.

Issue: #4953


abhinav92003:
Closing in favour of #5117.

PR #5117, opened 2021-09-22
i#4953 ubuntu20: Fix issues on common.decode test in x86-32

abhinav92003:
Fixes some issues related to preservation of regs in the common.decode
test, seen on the x86-32 bit suite on latest Ubuntu versions.

Avoids clobbering the stack reg in the `test_modrm16` test by saving it to
a scratch reg before executing the test modrm instr that overwrites it, and
restoring before `ret`. Without this, the `test_modrm16` invocations that
write to `esp` would crash at `ret`.

Removes `sigsetjmp` before all `test_modrm16` which is not required
now as it isn't expected to crash anymore; this also obviates the need to
use a separate signal stack on UNIX. In the earlier setup of this test,
even if some iteration of the modrm tests crashed due to an unexpected
issue, the test would go on to the next iteration. This could have hidden
some new issues from showing up in tests.

Fixes an issue in `mangle_seg_reg`, to avoid picking the same reg as
scratch for the 16-bit offset and the app TLS base offset, and avoid
sequences like the following:
```
mov    %eax -> %fs:0x00[4byte]
mov    %fs:0x48[4byte] -> %eax
mov    %eax -> %fs:0x00[4byte]
addr16 lea    (%bx,%si) -> %eax
mov    (%eax,%eax)[4byte] -> %ebx
```

Preserves callee saved regs in the `test_avx512_vex` test, which was
causing some crashes as well.

Removes unnecessary buf argument to `test_avx512_vex` which is unused.

Adds some more logs, for some helpful verbosity.

Removes common.decode from ignore list. Verified locally that there was no
failure in 1000 runs on 32-bit x86.

Issue: #4953


suite/tests/common/decode.c
```diff
@@ -145,6 +145,90 @@ actual_call_target(void)
...
+            } else {
+                /* Use eax to save esp in prolog and restore in epilog. */
+                buf[j * EACH_SEQ_SIZE + PROLOG_START + 1] = 0xe0; /* esp -> eax */
+                buf[j * EACH_SEQ_SIZE + EPILOG_START + 1] = 0xc4; /* eax -> esp */
+            }
+        } else {
+            /* Prolog and epilog are not needed for non-xsp dest, as we can
+             * handle other regs being clobbered.
+             */
+            buf[j * EACH_SEQ_SIZE + PROLOG_START + 0] = NOP_ENC;
```
derekbruening, line 198: So these nops are only here for each entry to have the exact same size and simplify sizing the buffer up front?  Maybe worth a comment.

derekbruening (approved):

suite/tests/common/decode.c
abhinav92003: It's already there on L169, at the beginning of `construct_modrm_test_buf`.

abhinav92003: Added the part about it helping to simplify buf size up front.

PR #5122, opened 2021-09-25
Fix wrong link (instrcalls sample)

Liblor:
While browsing the documentation I noticed a broken link due to a typo.


derekbruening (approved):
Thank you for the fix!

Issue #5121, opened 2021-09-25
Multi-phase drreg use hides app value from later phases


abhinav92003:
Today, DR supports multi-phase use for drreg, which means that clients can use drreg to get scratch regs in insertion phase and other phases as well.

#3823 added support to avoid spill slot conflicts in multi-phase use. However, different phases can still get the same scratch reg. In such "nested" spill regions, `drreg_get_app_value` won't work as expected in later phases, as it will give the value of the reg in the earlier phase, not the app value. This should be fixed, and documented until then.

Another related note:
Drreg support in non-insertion phases is bare bones -- e.g. it does not automatically restore app value before app read, or re-spill app value after app writes. This is usually fine because non-insertion phase is pretty limited. Today, we do that to expand scatter/gather instructions in app2app phase. But we should document this restriction about non-insertion phase use.

derekbruening:
This seems like a significant problem: not the drreg_get_app_value(), which is rare, but the automated restore before app reads, which is quite common and easily corrupts and crashes the app if not done correctly every time.

Does x86 scatter/gather expansion never reserve a register that's used in its own expansion, and that's what makes this problem not occur today?  If so, is that just because the expansion only uses a small set of registers and so there are enough unused ones available, even in 32-bit?

derekbruening:
> This seems like a significant problem: not the drreg_get_app_value(), which is rare, but the automated restore before app reads, which is quite common and easily corrupts and crashes the app if not done correctly every time.

Expanding on this: even if I reserve a reg in app2app across several instructions and I carefully restore the app value manually before app instrs that read that register, that's not enough.  The insertion phase could later use that same register, but have the wrong value stored in its slot, and think it's restoring the app value as part of its automated pass before that app read but instead it's undoing my manual restore and it's writing my tool value into that register.

abhinav92003:
>  I carefully restore the app value manually before app instrs that read that register, that's not enough. The insertion phase could later use that same register, but have the wrong value stored in its slot

Note that the insertion phase will treat the app2app restore as an app write, and update the reg's value in its slot. So the insertion phase slot will contain the actual app value after the app2app restore.

Correct drreg usage in app2app phase requires: manually restoring the app value before app reads, and manually re-spilling app value after app writes. My understanding is that if these are done correctly, then there shouldn't be any issue. If you have a case in mind, an example would help.

abhinav92003:
[...]

PR #5124, opened 2021-09-25
i#5121: Add test for insertion phase respill after app2app restore

abhinav92003:
Adds a new drreg test to show another multi-phase use scenario.
This demonstrates that the insertion phase updates the reg value
in its own slot after the app2app phase restores it to the original
app value before an app read. This results in overwriting the app2app
meta value that was present in the insertion phase's slot previously.
This original app value is then restored correctly by insertion phase
before subsequent app read instrs.

Instrumented basic block for the new test#38, after app2app and
insertion phase (each drreg op is annotated):
https://gist.github.com/abhinav92003/6a663cb19fda01838451bbe41b329bdc

Issue: #5121


ext/drreg/drreg.dox
```diff
@@ -42,7 +42,11 @@ instrumentation pass scheme.  The results are most efficient when used in
 the \p drmgr instrumentation insertion phase.  Use outside of the insertion
 phase, such as during \p drmgr's instrumentation-to-instrumentation
 transformation phase, or use outside of \p drmgr entirely, is supported,
-but duplicate spills and restores may not be optimized away.  When using
+but duplicate spills and restores may not be optimized away.  Also, for
+non-insertion phases, the responsibility to restore a spilled app value
+before an app read, and to re-spill the app value after an app write, lies
```
derekbruening, line 47: So we would also want a manual restore before a label with note DR_NOTE_ANNOTATION which my PR should have added to the docs: if you could add that here.

ext/drreg/drreg.dox
```diff
@@ -42,7 +42,11 @@ instrumentation pass scheme.  The results are most efficient when used in
...
+with the user; in the insertion phase, this is done automatically, but it
+needs to be done manually in non-insertion phases. When using
```
derekbruening, line 49: Maybe update my `XXX i#5121: multi-phase` comment in drreg_insert_restore_all() as well.

derekbruening (approved):

ext/drreg/drreg.dox
abhinav92003: Yes, a manual restore is required in app2app before that label.

Note that the insertion phase will re-spill the above restored value to its slot only if it sees that the value is live. This is of course always true when there's an app read coming up. For user prompted restores (like drreg_get_app_value or drreg_statelessly_restore_app_value), app2app needs to add a dummy read -- see PR #5126 for an example. Is anything like that required for the annotation case too?

derekbruening: I believe that having every app2app drreg usage that crosses app instructions look for the annotation label and restore prior to that label is sufficient.

ext/drreg/drreg.dox
abhinav92003: Done. PTAL.

ext/drreg/drreg.dox
abhinav92003: Done. PTAL.

ext/drreg/drreg.c
```diff
@@ -498,10 +498,20 @@ drreg_insert_restore_all(void *drcontext, instrlist_t *bb, instr_t *inst,
...
-     * XXX i#5121: multi-phase use causes correctness issues here!  Our restore
-     * in the insertion phase here could actually restore the tool value from the
-     * app2app phase.  A better solution is needed for multi-phase, or perhaps
-     * disallowing the use of the same register.
+     * XXX i#5121: For regs that are in spilled state at this instr also during the
+     * app2app phase, the following does not restore the app value, it restores the
+     * app2app meta value instead. This is expected and correct, because the drreg use
+     * in insertion phase adds spills/restores around the drreg spills/restores from
+     * previous phases like they were app instrs. So, the app2app meta value is no
+     * different than the app value for the insertion phase.
```
derekbruening, line 506: I found this confusing: it would only be the app2app meta value if the app2app phase does *not* restore before the read.  The last sentence only applies if the app2app phase restored before the read.

ext/drreg/drreg.c
```diff
@@ -498,10 +498,20 @@ drreg_insert_restore_all(void *drcontext, instrlist_t *bb, instr_t *inst,
...
+     *
+     * For cases where we really want the app value at some point in the insertion phase,
+     * e.g. for use with annotations, or for user-added calls to drreg_get_app_value and
```
derekbruening, line 509: But the app reading the register is even more important than annotations or drreg_get_app_value, and the user needs to restore for that too: so I also found this confusing.  The last sentence of this paragraph does say that but I don't understand why it's not considered a case "where we really want the app value".

ext/drreg/drreg.dox
```diff
@@ -44,7 +44,8 @@ phase, such as during \p drmgr's instrumentation-to-instrumentation
 transformation phase, or use outside of \p drmgr entirely, is supported,
 but duplicate spills and restores may not be optimized away.  Also, for
 non-insertion phases, the responsibility to restore a spilled app value
-before an app read, and to re-spill the app value after an app write, lies
+before certain instructions like an app read or a label with the note
+DR_NOTE_ANNOTATION, and to re-spill the app value after an app write, lies
```
derekbruening, line 48: Use `#` to get a link

ext/drreg/drreg.c
abhinav92003: Oh I meant "where the instrumentation wants the app value"

ext/drreg/drreg.c
abhinav92003: >  it would only be the app2app meta value if the app2app phase does not restore before the read

This comment starts with "For regs that are in spilled state at this instr also during the app2app phase", so it's talking about the not-restored case.

> The last sentence only applies if the app2app phase restored before the read.

I reworded it. Is it better now?

ext/drreg/drreg.dox
abhinav92003: Done.

ext/drreg/drreg.c
```diff
@@ -500,16 +500,16 @@ drreg_insert_restore_all(void *drcontext, instrlist_t *bb, instr_t *inst,
...
      * XXX i#5121: For regs that are in spilled state at this instr also during the
      * app2app phase, the following does not restore the app value, it restores the
-     * app2app meta value instead. This is expected and correct, because the drreg use
-     * in insertion phase adds spills/restores around the drreg spills/restores from
-     * previous phases like they were app instrs. So, the app2app meta value is no
-     * different than the app value for the insertion phase.
+     * app2app meta value instead. This is expected and correct, because drreg in
+     * insertion phase adds spills/restores around the drreg spills/restores from the
+     * previous phases like they were app instrs. So, the app2app meta value is treated
+     * like an app value by the insertion phase.
```
derekbruening, line 506: I still find this confusing, in the order it talks about things and the "expected and correct" for the non-ideal case of not having app2app restore.  This is the place in the code where drreg is adding spills/restores, but the comment implies it's somewhere else.

I would suggest:

XXX i#5121: Any app2app cross-app-instr use should manually restore app values prior to app reads.  That interacts well with the insertion phase since we treat those app2app restores as app writes and end up with the app value here.  If the app2app phase has a regsiter in a spilled state at this instr and did not restore the app value, the following insertion phase code will not restore the app value but instead will restore the app2app meta value.

ext/drreg/drreg.c
```diff
@@ -500,16 +500,16 @@ drreg_insert_restore_all(void *drcontext, instrlist_t *bb, instr_t *inst,
...
      *
-     * For cases where we really want the app value at some point in the insertion phase,
-     * e.g. for use with annotations, or for user-added calls to drreg_get_app_value and
-     * drreg_statelessly_restore_app_value to work as expected, the user needs to restore
-     * the app value at the same point in the app2app phase also. This aligns with other
-     * similar expectations from the user for non-insertion phase use of drreg, like
+     * For cases where instrumentation wants the app value at some point in the insertion,
```
derekbruening, line 508: s/,//

ext/drreg/drreg.c
```diff
@@ -500,16 +500,16 @@ drreg_insert_restore_all(void *drcontext, instrlist_t *bb, instr_t *inst,
...
+     * phase e.g. for use with annotations, or for user-added calls to drreg_get_app_value
```
derekbruening, line 509: run-on: how about () around the e.g. phrase or at least a , before e.g.

ext/drreg/drreg.c
```diff
@@ -500,16 +500,16 @@ drreg_insert_restore_all(void *drcontext, instrlist_t *bb, instr_t *inst,
...
+     * and drreg_statelessly_restore_app_value to work as expected, the user needs to
+     * restore the app value at the same point in the app2app phase also. This aligns with
+     * other similar expectations from the user for non-insertion phase use of drreg, like
      * manually restoring/re-spilling around app reads and writes, which happens
```
derekbruening, line 513: This was just talked about: I would delete this whole sentence (else it is confusing as it implies what the first paragraph talked about is unrelated).

ext/drreg/drreg.c
abhinav92003: Done.

ext/drreg/drreg.c
abhinav92003: Done.

ext/drreg/drreg.c
abhinav92003: Done.

ext/drreg/drreg.c
abhinav92003: Done. Thanks for the detailed suggestion! Couldn't figure out the best words here on my own!

ext/drreg/drreg.c
abhinav92003: Resolved in newer comment thread.

ext/drreg/drreg.c
abhinav92003: Resolved in a later comment thread.

PR #5125, opened 2021-09-25
i#4678: AArch64 fuzz app for drstatecmp.

sapostolakis:
Adds a fuzzing app for AArch64 that automatically generates from the decoding
files a random sequence of side-effect-free and non-branch instructions with 
fuzzed operands for stress-testing clients with drstatecmp. 
This enables automatic testing of even newly inserted instructions in the decoder.

Adds a drcachesim test with delayed tracing for the fuzzing app. 
Testing with an older DR version showed that this test would have eventually 
automatically detected #5023 (instrumentation-clobbering of aflags).

Originally part of #4938. 

Issue: #4678

### DIFF
--- a/core/ir/aarch64/codec.py
+++ b/core/ir/aarch64/codec.py
@@ -115,15 +115,15 @@ def generate_decoder(patterns, opndsettab, opndtab, opc_props):
     # Recursive function to generate nested conditionals in main decoder.
     def gen(c, pats, depth):
         def reorder_key(t):
-            f, v, m, t = t
-            return (m, t, f, v)
+            f, v, m, t, p = t
+            return (m, t, f, v, p)
 
         def indent_append(text):
             c.append('{}{}'.format("    " * depth, text))
 
         if len(pats) < 4:
             else_str = ""
-            for opcode_bits, opnd_bits, m, t in sorted(pats, key=reorder_key):
+            for opcode_bits, opnd_bits, m, t, p in sorted(pats, key=reorder_key):
 
                 not_zero_mask = 0
                 try:
@@ -180,7 +180,7 @@ def generate_decoder(patterns, opndsettab, opndtab, opc_props):
         for b in range(N):
             x0 = 0
             x1 = 0
-            for (f, v, _, _) in pats:
+            for (f, v, _, _, _) in pats:
                 if (1 << b) & (~f | v):
                     x0 += 1
                 if (1 << b) & (f | v):
@@ -193,7 +193,7 @@ def generate_decoder(patterns, opndsettab, opndtab, opc_props):
         pats0 = []
         pats1 = []
         for p in pats:
-            (f, v, _, _) = p
+            (f, v, _, _, _) = p
             if (1 << best_b) & (~f | v):
                 pats0.append(p)
             if (1 << best_b) & (f | v):
@@ -284,7 +284,7 @@ def generate_encoder(patterns, opndsettab, opndtab):
         c.append('')
     case = dict()
     for p in patterns:
-        (b, m, mn, f) = p
+        (b, m, mn, f, _) = p
         if not mn in case:
             case[mn] = []
         case[mn].append(p)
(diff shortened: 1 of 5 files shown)


suite/tests/client-interface/drstatecmp-fuzz-app.c
```diff
@@ -0,0 +1,238 @@
...
+}
+
+static byte *
+generate_inst(byte *encode_pc)
+{
+    /* Pick a random side-effect-free and non-branch instruction. */
+    uint32_t encoded_inst = generate_encoded_inst();
+    byte encoded_inst_bytes[4];
+    for (int i = 0; i < 4; ++i, encoded_inst >>= 8)
+        encoded_inst_bytes[i] = (byte)(encoded_inst & 0xff);
```
derekbruening, line 148: But the encoding is already in `encoded_inst` in little-endian form: this byte array and loop can be removed and we can just point at `&encoded_inst`.

core/ir/aarch64/codec.py
```diff
@@ -390,6 +390,42 @@ def generate_opcode_names(patterns):
...
+def generate_opcode_opnd_pairs(patterns):
+    c = ['#ifndef OPCODE_OPND_PAIRS_H',
+         '#define OPCODE_OPND_PAIRS_H 1',
+         '',
+         '#include <stdint.h>',
+         '',
+         'typedef struct {',
+         '  uint32_t opcode;',
+         '  uint32_t opnd;',
+         '} opcode_opnd_pair_t;',
```
derekbruening, line 405: suggest: use a dr_ (and DR_ for `#define DR_FUZZ_INST_CNT`) prefix for all global types and variables to avoid namespace collisions.  We do have some legacy types that don't have a prefix, but going forward we should try to avoid problems.  We have had a number of collision issues in the past with statically linking DR into large applications that also have other third-party C libraries linked in.

suite/tests/client-interface/drstatecmp-fuzz-app.c
```diff
@@ -0,0 +1,238 @@
...
+        (byte *)allocate_mem(code_size, ALLOW_EXEC | ALLOW_READ | ALLOW_WRITE);
+    assert(generated_code != NULL);
+
+    /* Synthesize code which includes a lot of side-effect-free instructions. Only one
+     * basic block is created (linear control flow). To test clobbering of arithmetic
+     * flags conditionally-executed instructions are included.
+     */
+    byte *encode_pc = generated_code;
+    for (int i = 0; i < NUM_INSTS; ++i) {
+        encode_pc = generate_inst(encode_pc);
```
derekbruening, line 179: But if check_decoded_inst() finds an invalid inst, encode_pc will not move forward, and we'll end up with fewer than NUM_INSTS instructions here.  Better to count actually encoded instrs.

suite/tests/client-interface/drstatecmp-fuzz-app.c
```diff
@@ -0,0 +1,238 @@
...
+    generate_code();
+
+    /* Handle execution of illegal instructions that were decodable.
+     * (fairly common).
+     */
+    struct sigaction act_ill;
+    memset(&act_ill, 0, sizeof(act_ill));
+    act_ill.sa_sigaction = (void (*)(int, siginfo_t *, void *))sigill_handler;
+    act_ill.sa_flags = SA_SIGINFO;
+    sigaction(SIGILL, &act_ill, NULL);
```
derekbruening, line 207: This is already including tools.h so this whole sequence can be replaced with intercept_signal.

suite/tests/client-interface/drstatecmp-fuzz-app.c
```diff
@@ -0,0 +1,238 @@
...
+
+    /* Handle seg faults and floating-point exceptions caused by the fuzzed insts
+     * (rarely occur).
+     */
+    struct sigaction act_segv_fpe;
+    memset(&act_segv_fpe, 0, sizeof(act_segv_fpe));
+    act_segv_fpe.sa_sigaction = (void (*)(int, siginfo_t *, void *))sig_segv_fpe_handler;
+    act_segv_fpe.sa_flags = SA_SIGINFO;
+    sigaction(SIGSEGV, &act_segv_fpe, NULL);
```
derekbruening, line 216: Ditto

derekbruening (approved):

core/ir/aarch64/codec.py
sapostolakis: Done.

suite/tests/client-interface/drstatecmp-fuzz-app.c
sapostolakis: Right. Wasn't sure about the endianness and did the byte array to be safe but it is simpler this way. Changed.

suite/tests/client-interface/drstatecmp-fuzz-app.c
sapostolakis: Now compute actual count of encoded insts and changed the check with this count.

suite/tests/client-interface/drstatecmp-fuzz-app.c
sapostolakis: Done.

suite/tests/client-interface/drstatecmp-fuzz-app.c
sapostolakis: Done.

Issue #5131, opened 2021-09-29
Possibly unhandled system calls in Linux kernel 5.10


abhinav92003:
Found some unhandled system calls while updated Ubuntu CI to 20.04 in #4953.

On my local system, which is Linux kernel 5.10.46, following system calls are new:

x86-64:
```
+#    define __NR_pidfd_send_signal 424
+#    define __NR_io_uring_setup 425
+#    define __NR_io_uring_enter 426
+#    define __NR_io_uring_register 427
+#    define __NR_open_tree 428
+#    define __NR_move_mount 429
+#    define __NR_fsopen 430
+#    define __NR_fsconfig 431
+#    define __NR_fsmount 432
+#    define __NR_fspick 433
+#    define __NR_pidfd_open 434
+#    define __NR_clone3 435
+#    define __NR_close_range 436
+#    define __NR_openat2 437
+#    define __NR_pidfd_getfd 438
+#    define __NR_faccessat2 439
+#    define __NR_process_madvise 440
```

x86-32:
```
+#    define __NR_semget 393
+#    define __NR_semctl 394
+#    define __NR_shmget 395
+#    define __NR_shmctl 396
+#    define __NR_shmat 397
+#    define __NR_shmdt 398
+#    define __NR_msgget 399
+#    define __NR_msgsnd 400
+#    define __NR_msgrcv 401
+#    define __NR_msgctl 402
+#    define __NR_clock_gettime64 403
+#    define __NR_clock_settime64 404
+#    define __NR_clock_adjtime64 405
+#    define __NR_clock_getres_time64 406
+#    define __NR_clock_nanosleep_time64 407
+#    define __NR_timer_gettime64 408
+#    define __NR_timer_settime64 409
+#    define __NR_timerfd_gettime64 410
+#    define __NR_timerfd_settime64 411
+#    define __NR_utimensat_time64 412
+#    define __NR_pselect6_time64 413
+#    define __NR_ppoll_time64 414
+#    define __NR_io_pgetevents_time64 416
+#    define __NR_recvmmsg_time64 417
+#    define __NR_mq_timedsend_time64 418
+#    define __NR_mq_timedreceive_time64 419
+#    define __NR_semtimedop_time64 420
+#    define __NR_rt_sigtimedwait_time64 421
+#    define __NR_futex_time64 422
+#    define __NR_sched_rr_get_interval_time64 423
+#    define __NR_pidfd_send_signal 424
+#    define __NR_io_uring_setup 425
[...]

PR #5132, opened 2021-09-29
i#4953 Ubuntu20: Fix linux.syscall_pwait hang.

abhinav92003:
Fixes hang at the syscall ppoll_time64 by handling it same as ppoll.
The former is the year-2038-safe variant of the latter, that allows
a wider field for timestamp, as the year 2038 will see a unix
timestamp greater than 2^31.

Adds new syscalls up to Linux kernel 5.10.46.

Adds a warning for new syscalls other than ppoll_time64 that are
not handled yet and have a similarly named variant that needs
special handling. E.g. SYS_clone has special handling, so added
a warning for SYS_clone3.

Also upgrades Ubuntu version for the x86-32 bit GA CI to 20.04.

Issue: #5131
Fixes: #4953


core/unix/include/syscall_linux_x86.h
```diff
@@ -356,10 +358,29 @@
 #    define __NR_statx 332
 #    define __NR_io_pgetevents 333
 #    define __NR_rseq 334
+#    define __NR_pidfd_send_signal 424
```
derekbruening, line 361: This gap looks suspicious: but on inspection it seems to be deliberate.

core/unix/os.c
```diff
@@ -5019,6 +5034,9 @@ ignorable_system_call_normalized(int num)
 #    endif
     case SYS_readlinkat: return !DYNAMO_OPTION(early_inject);
 #endif
+#ifdef SYS_openat2
+    case SYS_openat2: return IS_STRING_OPTION_EMPTY(xarch_root);
```
derekbruening, line 5038: Hmm should the warning then be under -xarch_root too?

core/unix/os.c
```diff
@@ -8637,13 +8688,19 @@ post_system_call(dcontext_t *dcontext)
     }
 #endif
 #ifdef LINUX
+#    ifdef SYS_ppoll_time64
+    case SYS_ppoll_time64:
```
derekbruening, line 8692: Add a test to suite/tests/linux/syscall_pwait.cpp (maybe just the raw version, a la `psyscall_raw_ppoll`) under ifdef SYS_ppoll_time64?

derekbruening (approved):

core/unix/include/syscall_linux_x86.h
abhinav92003: `__NR_pidfd_send_signal` onwards syscall number assignments seem to align on 32- and 64-bit, and the space is perhaps for arch-specific syscall. https://github.com/torvalds/linux/commit/0d6040d4681735dfc47565de288525de405a5c99

core/unix/os.c
abhinav92003: Yeah, there should be. Done.

core/unix/os.c
abhinav92003: Done.

core/unix/os.c
abhinav92003: Wait, this the syscall is ignorable if xarch_root is not set right? That is the case this PR doesn't print warning for currently. 
If xarch_root is set, the pre and post handling will print the warning anyway.

derekbruening: The xarch_root path manipulation seems to be very targeted to what the loader uses: it is only looking at openat and not at open.  So if the loader switched to using open it would also find the wrong lib w/o a warning: but these are easy errors to find since you'd get the wrong arch and see the error right away; plus these are common syscalls so we don't want to intercept (and avoid inlining) if we can avoid it.

core/unix/os.c
```diff
@@ -5035,7 +5035,12 @@ ignorable_system_call_normalized(int num)
     case SYS_readlinkat: return !DYNAMO_OPTION(early_inject);
 #endif
 #ifdef SYS_openat2
-    case SYS_openat2: return IS_STRING_OPTION_EMPTY(xarch_root);
+    case SYS_openat2:
+        if (!IS_STRING_OPTION_EMPTY(xarch_root))
+            return false;
+        SYSLOG_INTERNAL_WARNING_ONCE(
+            "WARNING i#5131: possibly unhandled system call openat2");
```
derekbruening, line 5042: I think this should be removed: we don't care about open and so are unlikely to care about openat2.

derekbruening (approved):

core/unix/os.c
abhinav92003: Okay so no warning then?

core/unix/os.c
abhinav92003: But we do care about `SYS_openat`, which is why I assumed we should atleast warn for `SYS_openat2`. Xref the discussion in the comment thread: IIUC we expect any issues with openat2 handling to be clearly visible, and that is why we want to remove the warning?

core/unix/os.c
derekbruening: I don't have a strong opinion for the -xarch_root case, could go either way.

core/unix/os.c
derekbruening: If -xarch_root is not set I don't think there should be any warning.

core/unix/os.c
abhinav92003: Removed it.

core/unix/os.c
abhinav92003: Removed it.

PR #5134, opened 2021-09-30
i#5133: Call __libc_eary_init for private libc

### LINKED ISSUE #5133: Private libc is not initialized on Ubuntu 21
This was discovered in DynamoRIO/drmemory#2425

The upshot is that in glibc-2.32-403, ld.so now calls a secret routine `__libc_early_init` to initialize some key pieces of libc, rather than using regular ELF constructors as in the past.  This can easily result in tool crashes referencing NULL pointers.  We'll have to mirror this hardcoded special behavior in DR's loader.

derekbruening:
Mirrors the glibc 2.32+ behavior where ld.so has a hardcoded call to
"__libc_eary_init" before it calls the regular ELF constructors.

This fixes crashes in clients on Ubuntu 21.
Tested in Dr. Memory on Ubuntu 21.

Fixes #5133

### DIFF
--- a/core/unix/loader.c
+++ b/core/unix/loader.c
@@ -136,6 +136,7 @@ stdfile_t **privmod_stdin;
 #define LIBC_STDOUT_NAME "stdout"
 #define LIBC_STDERR_NAME "stderr"
 #define LIBC_STDIN_NAME "stdin"
+#define LIBC_EARLY_INIT_NAME "__libc_early_init"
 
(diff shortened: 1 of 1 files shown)


derekbruening:
Windows failures are sourceforge yet again.  Maybe we should look into checking the doxygen binary into our repo or something.

abhinav92003 (approved):

PR #5136, opened 2021-09-30
i#2243 drcov segments: Increase drov output format version

derekbruening:
PR #5094 added support for non-first-segment-code in drcov by changing
the drcov output format.  It should have bumped its file format
version; we do that here.

drcov2lcov prints a warning but continues if passed the prior version,
as we expect it to work for cases that worked before: code in the
first segment.  An old build of drcov2lcov would have failed on the
different module table format in any case; now it gives a better error
message about a version change.

Reduces the warning level to 1 for the fatal warnings opening files to
make them more visible.

Tested on a version 2 file from an 8.0 run on a +rx-first-segment application:
    $ clients/bin64/drcov2lcov -input ~/dr/releases/DynamoRIO-Linux-8.0.0-1/drcov*.log && grep -A 5 hello coverage.info
    [6/6] Generating embed/html/index.html
    [DRCOV2LCOV] WARNING(1): No output file name specified: using default coverage.info
    [DRCOV2LCOV] INFO(1):    Reading input files...
    [DRCOV2LCOV] WARNING(1): File is in legacy version 2 format: only code in the first segment of each module will be reported
    [DRCOV2LCOV] INFO(1):    Enumerating line info...
    [DRCOV2LCOV] WARNING(1): Failed to enumerate lines for [vdso]
    [DRCOV2LCOV] WARNING(1): Failed to enumerate lines for [vdso]
    [DRCOV2LCOV] INFO(1):    Writing output file...
    SF:/home/derek/hello.c
    DA:4,1
    DA:5,1
    DA:6,1
    DA:7,1
    DA:8,1

Issue: #2243

### DIFF
--- a/clients/drcov/postprocess/drcov2lcov.cpp
+++ b/clients/drcov/postprocess/drcov2lcov.cpp
@@ -882,7 +882,7 @@ read_file_header(const char *buf)
         return NULL;
     }
     if (version != DRCOV_VERSION) {
-        if (version == 2) {
+        if (version == DRCOV_VERSION_MODULE_OFFSETS) {
             WARN(1,
                  "File is in legacy version 2 format: only code in the first "
                  "segment of each module will be reported\n");
--- a/clients/drcov/postprocess/drcov2lcov.cpp
+++ b/clients/drcov/postprocess/drcov2lcov.cpp
@@ -774,7 +774,7 @@ read_module_list(const char *buf, module_table_t ***tables, uint *num_mods)
     /* module table header */
     if (drmodtrack_offline_read(INVALID_FILE, buf, &buf, &handle, num_mods) !=
         DRCOVLIB_SUCCESS) {
-        WARN(2, "Failed to read module table");
+        WARN(1, "Failed to read module table");
         return NULL;
     }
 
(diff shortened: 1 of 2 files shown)


derekbruening:
Win64 failure is drwrap-test output mismatch: #4983.

clients/drcov/postprocess/drcov2lcov.cpp
```diff
@@ -878,24 +878,30 @@ read_file_header(const char *buf)
...
     if (dr_sscanf(buf, "DRCOV VERSION: %u\n", &version) != 1) {
-        WARN(2, "Failed to read version number");
+        WARN(1, "Failed to read version number");
         return NULL;
     }
     if (version != DRCOV_VERSION) {
-        WARN(2, "Version mismatch: file version %d vs tool version %d\n", version,
-             DRCOV_VERSION);
-        return NULL;
+        if (version == 2) {
```
abhinav92003, line 885: Maybe add a named constant for '2', describing that drcov version (similar to the OFFLINE_FILE_VERSION* constants in drmemtrace/trace_entry.h.

abhinav92003 (approved):

clients/drcov/postprocess/drcov2lcov.cpp
derekbruening: Done.

PR #5138, opened 2021-10-01
i#5076 trace invariants, part 4: Add unit tests

derekbruening:
Adds targeted testing for the trace invariants checker tool for checks
that are not tested by the signal_invariants app's prefetch and
handler markers.

Issue: #5076

### DIFF
--- a/clients/drcachesim/tests/invariant_checker_test.cpp
+++ b/clients/drcachesim/tests/invariant_checker_test.cpp
@@ -138,7 +138,7 @@ check_branch_target_after_branch()
         }
         if (checker.errors.size() != 1 ||
             checker.errors[0] != "Branch target not immediately after branch") {
-            std::cerr << "Failed to catch bad branch target poisition\n";
+            std::cerr << "Failed to catch bad branch target position\n";
             return false;
         }
     }
@@ -170,6 +170,23 @@ check_branch_target_after_branch()
 bool
 check_sane_control_flow()
 {
+    // Positive test: branches.
+    {
+        checker_no_abort_t checker(true /*offline*/);
+        std::vector<memref_t> memrefs = {
+            gen_instr(1, 1),   gen_branch(1, 2),  gen_instr(1, 3), // Not taken.
+            gen_branch(1, 4),  gen_instr(1, 101),                  // Taken.
+            gen_instr(1, 102),
+        };
+        for (const auto &memref : memrefs) {
+            checker.process_memref(memref);
+        }
+        for (const auto &error : checker.errors) {
+            std::cerr << "Unexpected error: " << error << "\n";
+        }
+        if (!checker.errors.empty())
+            return false;
+    }
     // Negative simple test.
     {
         checker_no_abort_t checker(true /*offline*/);
--- a/clients/drcachesim/tests/invariant_checker_test.cpp
+++ b/clients/drcachesim/tests/invariant_checker_test.cpp
@@ -1,0 +1,327 @@
+/* **********************************************************
+ * Copyright (c) 2021 Google, LLC  All rights reserved.
+ * **********************************************************/
+
+/*
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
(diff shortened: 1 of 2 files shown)


derekbruening:
32-bit x86 failure is api.startstop #4604

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -0,0 +1,331 @@
...
+            gen_branch(1, 2),
+            gen_marker(2, TRACE_MARKER_TYPE_TIMESTAMP, 0),
+            gen_instr(2, 1),
+        };
+        for (const auto &memref : memrefs) {
+            checker.process_memref(memref);
+        }
+        if (checker.errors.size() != 1 ||
+            checker.errors[0] != "Branch target not immediately after branch") {
+            std::cerr << "Failed to catch bad branch target poisition\n";
```
abhinav92003, line 141: spelling: 'position'

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -0,0 +1,331 @@
...
+
+bool
+check_branch_target_after_branch()
+{
+    // Positive simple test.
+    {
+        checker_no_abort_t checker(true /*offline*/);
+        std::vector<memref_t> memrefs = {
+            gen_instr(1, 1), gen_branch(1, 2),
+            gen_instr(1, 3), gen_marker(2, TRACE_MARKER_TYPE_TIMESTAMP, 0),
```
abhinav92003, line 115: So here the branch is not taken (as the next instruction has pc = prev_pc + 1). Can we add a positive test for a taken branch, where the next instruction pc is the target of the branch?

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -0,0 +1,331 @@
...
+            return false;
+        }
+    }
+    // Invariant relaxed for thread exit or signal.
+    {
+        checker_no_abort_t checker(true /*offline*/);
+        std::vector<memref_t> memrefs = {
+            gen_marker(3, TRACE_MARKER_TYPE_CACHE_LINE_SIZE, 64),
+            gen_branch(3, 2),
+            gen_exit(3),
```
abhinav92003, line 151: Curious when this happens: abrupt exit after a branch. Shouldn't there be other instructions leading up to an exit system call?

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -0,0 +1,331 @@
...
+    // Return to recorded interruption point.
+    {
+        checker_no_abort_t checker(true /*offline*/);
+        std::vector<memref_t> memrefs = {
+            gen_instr(1, 1),
+            gen_marker(1, TRACE_MARKER_TYPE_KERNEL_EVENT, 2),
+            gen_instr(1, 101),
+            // XXX: This marker value is actually not guaranteed, yet the checker
+            // requires it and the view tool prints it.
+            gen_marker(1, TRACE_MARKER_TYPE_KERNEL_XFER, 102),
```
abhinav92003, line 241: Will another test case for nested signals be useful?

abhinav92003 (approved):

clients/drcachesim/tests/invariant_checker_test.cpp
derekbruening: If we traced the entire application lifetime, yes, but for detach the exit point can be anywhere our signal arrived: and since we delay asynch signals to block end it is almost always after a branch in practice.

clients/drcachesim/tests/invariant_checker_test.cpp
derekbruening: Yes, the plan is to add a nested signals check: but right now that check does not exist in invariant_checker (partly b/c it is really hard in the presence of longjmp).  That will be a future PR.

clients/drcachesim/tests/invariant_checker_test.cpp
derekbruening: Fixed.

Issue #5129, opened 2021-09-27
relink_special_ibl_xfer() may overwrite first instruction of target block on AArch64 (CRASH)


AssadHashmi:
This bug was spotted by a user and posted at https://groups.google.com/g/dynamorio-users/c/Ai5fFN9Q6iE

AArch64 uses two instructions to patch an IBL: `LDR` followed by `BR`.

However, both AArch32 and x86 use one jump instruction and the IBL patching functions assume only a one instruction jump patch. This can result in AArch64 applications crashing as summarised below.

Code generated by `emit_special_ibl_xfer()` on AArch64 (using `-loglevel 4`):
```
client_ibl_xfer:
. . .
  0x0000000041e424b8  f9403f81   ldr    +0x78(%x28)[8byte] -> %x1
  0x0000000041e424bc  d61f0020   br     %x1
clean_call_save:
  0x0000000041e424c0  a90007e0   stp    %x0 %x1 -> (%sp)[16byte]
. . .
```
When `relink_special_ibl_xfer()` is called the `BR` instruction of the `LDR`/`BR` pair overwrites:
```
. . .
  0x0000000041e424b8  f9403f81   ldr    +0x78(%x28)[8byte] -> %x1     <-- ldr    +0x78(%x28)[8byte] -> %x1
  0x0000000041e424bc  d61f0020   br     %x1                           <-- ldr    +0x78(%x28)[8byte] -> %x1
clean_call_save:
  0x0000000041e424c0  a90007e0   stp    %x0 %x1 -> (%sp)[16byte]      <-- br     %x1     <-- STP overwritten!
. . .
```

We have not seen this bug cause failure before due to `NOP` padding used for e.g. cache alignment which depends on the machine's cache line size:
```
client_ibl_xfer:
. . .
  0x00000000310124ac  f9403f81   ldr    +0x78(%x28)[8byte] -> %x1  <-- ldr    +0x78(%x28)[8byte] -> %x1
  0x00000000310124b0  d61f0020   br     %x1                        <-- ldr    +0x78(%x28)[8byte] -> %x1
  0x00000000310124b4  d503201f   nop                               <-- br     %x1
  0x00000000310124b8  d503201f   nop
  0x00000000310124bc  d503201f   nop
clean_call_save:
  0x00000000310124c0  a90007e0   stp    %x0 %x1 -> (%sp)[16byte]   <-- NOT overwritten because of NOP padding above.
. . .
```

PR #5142, opened 2021-10-05
i#5129 Fix client_ibl_xfer to avoid overwriting next block

AssadHashmi:
Fixes the case on AArch64 where the LDR+BR jump instructions in the
client_ibl_xfer block can overwrite the first instruction of the next
generated block. This happens when relink_special_ibl_xfer() is called
AND the cache line alignment places the LDR+BR at the end of
client_ibl_xfer.

Issue: #5129

### DIFF
--- a/core/arch/emit_utils_shared.c
+++ b/core/arch/emit_utils_shared.c
@@ -5483,7 +5483,7 @@ emit_special_ibl_xfer(dcontext_t *dcontext, byte *pc, generated_code_t *code, ui
      */
     reg_id_t stub_reg = IF_AARCH64_ELSE(SCRATCH_REG0, SCRATCH_REG1);
     ushort stub_slot = IF_AARCH64_ELSE(TLS_REG0_SLOT, TLS_REG1_SLOT);
-    IF_X86(size_t len;)
+    size_t len;
     byte *ibl_linked_tgt = special_ibl_xfer_tgt(dcontext, code, IBL_LINKED, ibl_type);
     byte *ibl_unlinked_tgt = special_ibl_xfer_tgt(dcontext, code, IBL_UNLINKED, ibl_type);
     bool absolute = !code->thread_shared;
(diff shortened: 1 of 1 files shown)


core/arch/emit_utils_shared.c
```diff
@@ -5625,6 +5625,20 @@ emit_special_ibl_xfer(dcontext_t *dcontext, byte *pc, generated_code_t *code, ui
             dcontext, opnd_create_reg(SCRATCH_REG1),
             OPND_TLS_FIELD(get_ibl_entry_tls_offs(dcontext, ibl_linked_tgt))));
     APP(&ilist, XINST_CREATE_jump_reg(dcontext, opnd_create_reg(SCRATCH_REG1)));
+    /* Unlike X86 and ARM/AArch32 which use 1 instruction for an indirect jump,
+     * AArch64 requires 2 instructions: LDR+BR. In the case where cache line
+     * alignment places LDR+BR at the end of the IBL transfer stub, a relink
+     * (see relink_special_ibl_xfer) will result in the BR instruction
+     * overwriting the first instruction of the next stub, (currently
+     * clean_call_save). This code checks if LDR+BR are at the end and adds a
+     * NOP to trigger padding for cache line alignment when emitted.
```
derekbruening, line 5634: But this will still turn LDR;BR;NOP into LDR;LDR;BR which does not seem ideal to have an extra LDR to the same reg.  It seems cleaner to fix the root of the problem: point the special_ibl_unlink_offs at the LDR instead of the BR, right?  And have the relink routine only write the LDR since the BR does not need to change.

derekbruening: A separate topic: relink_special_ibl_xfer is modifying a LDR instruction which is not on the list of opcodes that do not require synchronization with other threads when changing: so this code change method is technically unsafe.  We modified the aarch64 exit stubs to only change either data or opcodes on the list; probably we should do the same thing here.  Could we change a TLS value instead of changing the LDR to point to different TLS slots?  Currently this is under #1911.

derekbruening (changes requested):

derekbruening:
> @derekbruening , firstly is the special_ibl_unlink_offs fix correct?

Sorry for the delay, I was out Friday.  No, the change looks like it would break other updates.

PR #5143, opened 2021-10-05
i#5131 new linux syscalls: Handle clone3

abhinav92003:
Adds  support for the clone3 syscall, similar to existing support  for the clone
syscall.  The functionality provided by  clone3  is a superset of clone.  It uses
a struct to encapsulate all args. Also, the stack address is expected to be the
lowest one in the stack (inclusive);  this is unlike clone where it is expected to
be the highest one (non-inclusive).

Extends the existing linux.clone test to verify behaviour on clone3 as well.

Issue: #5131


abhinav92003:
Will give this another self-review pass tomorrow before re-requesting review.

PR #5152, opened 2021-10-08
cmake: add support for DESTDIR install relocation

fmoessbauer:
CMake might change the value of CMAKE_INSTALL_PREFIX on the fly at
install time in case the install is relocated via DESTDIR.

This is only supported in case the paths of the install target are
encoded as relative path.
These paths are anyways located relative to the CMAKE_INSTALL_PREFIX,
so manually specifying the prefix there is not required.

This patch enables a user to install multiple DynamoRIO versions
simultaneously and to switch between them by using tools like xstow.
It also helps as a preparation to package DynamoRIO for Linux
distributions like Debian.


derekbruening:
run arm tests

derekbruening:
Thank you for the contribution.

Issue #2414, opened 2017-05-04
[Feature Request] Getting stack trace when exception


illera88:
Hi,
There is no way (I know of) to get the stack trace when a exception is raised. This would be very helpful to collect information about an exception or crash.

Can this be done from within DynamoRio? I can help implementing it. Can I get pointed to where it should be implemented? 

Cheers

derekbruening:
Please clarify:

+ What kind of exception: in the app, in DR, in a client?  They are handled very differently.

+ What kind of stack trace: print raw addresses, print symbolic names, programmatic output fed to other code, assume frame pointers or have stack scanning and other features, etc.

Xref https://github.com/DynamoRIO/drmemory/issues/823

illera88:
- In a client
- I want to get the same stack trace that you get when running a program that crashes through GDB for example

derekbruening:
Getting a callstack in arbitrary code can be quite challenging: just look at all of the many issues in the Dr. Memory tracker on problems and challenges getting callstacks on Windows.  Linux is generally simpler, and the more code you control compilation of the easier it is, so it is easier for client code than general app code (though clients can easily call off into Windows private libraries).

On a crash in DR we do not know what state anything is in and can't reliably run much code at all.  A client crash is a little safer but it could still have corrupted state needed to do things like try to match what a debugger does, which on Windows would involve calling into dbghelp.dll.

The current simple DR frame pointer walk could be post-processed:
Xref #973
Xref #1470

Clients could be given a control point and could from there use an extension or sthg:
Xref #50

illera88:
First of all thank you for your time. 
Second, I didn't understand your first question regarding where the crash happens. My bad. I want to get the stack trace of a crash in the instrumented application. I said in the client and I meant in the instrumented binary. 

In the client I'm using `drmgr_register_exception_event` or `drmgr_register_signal_event` (windows and linux) and once the exception callback is called I want to get the stack trace from it (from the target binary, not from the DR client).

Sorry for the confusion. 
Cheers

derekbruening:
Your earlier users list emails were also talking about a report on a DR or tool crash, so this is rather confusing.  If you're looking for an app callstack, I don't see that it has much relationship to an exception or fault.  This just sounds like a request for a general app callstack feature and is unrelated to whether an app exception just happened.

derekbruening:
[...]

PR #5154, opened 2021-10-11
i#2414 drcallstack: Add libunwind-based callstack walking

derekbruening:
Adds a new extension "drcallstack" which provides callstack walking
facilities.  This initial implementation adds a libunwind-based
implementation and targets only Linux for now.

Adds an interface to walk one step at a time over callstack frames.
The implementation converts the dr_mcontext_t into libunwind's context
structure and invokes the libunwind step API.

Getting libunwind to work requires several steps:

+ Ignore libpthread exports when importing any symbol that does not start with
  "pthread".  Otherwise libunwind crashes using __errno_location from libpthread
  instead of from libc.  We add a warning when a private libpthread is loaded
  to help diagnose any other potential problems (xref #956).

+ Have dl_iterate_phdr operate on app libraries instead of private libraries.
  This is done with a new flag and logic in the redirection code, with the
  flag set for libraries named "libunwind*".

A statically-linked libunwind is not supported.

Updates drwrap to set the mcontext pc field to simplify usage.

Adds a test and documentation.

Adds a sample client showing how to use this library.

Issue: #2414, #956

### DIFF
--- a/ext/CMakeLists.txt
+++ b/ext/CMakeLists.txt
@@ -126,7 +126,8 @@ if (LINUX) # TODO i#2414: Port to Windows, Mac, and Android.
   else ()
     # Require libunwind.h for automated testing.
     # The TEST_SUITE var is set for packages too, to ensure we publish drcallstack.
-    if (TEST_SUITE)
+    # We do not install for Android nor a64-on-x86 in the suite.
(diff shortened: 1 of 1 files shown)


derekbruening:
PTAL at the workflow changes especially.  The 32-bit one is very frustrating that we have to go those lengths just to installed libunwind-dev:i386 on their setup (vanilla Ubuntu installs fine w/ the default repos).

PR #5157, opened 2021-10-13
Match param name in function declaration and definition.

abhinav92003:
Fixes Clang-Tidy error due to mismatch in a parameter name in
create_clone_record definition and declaration.


derekbruening:
> tool.drcachesim.delay-global failed, output stops unexpectedly, did it crash? In any case, it is definitely unrelated to this change.

Please file an issue if there is not already one.

abhinav92003:
> Please file an issue if there is not already one.

Filed one for the A64 `delay-global` failure. vs2017-64 `earlythread` failure is an existing issue linked above.

Issue #5161, opened 2021-10-14
Collapse drmgr into the core and unify the event interface


derekbruening:
Xref #4128 where having key libraries separate from the core causes headaches, and #5160.

The DR API has evolved to the point where all the powerful features rely on drmgr and there's really no reason to support non-drmgr use anymore.  Having it separate adds complications that may outweigh the "keep the core simple" goal.

The idea here would be to move drmgr into the core and remove all the layered events, and probably give all events a priority and user_data (user_data is key to better C++ interaction to avoid requiring static functions).  So we'd get rid of the plain basic block event, e.g.

Issue #5160, opened 2021-10-14
Move annotation handling to a client?


derekbruening:
For #4128 we want drreg to know about all clean calls, yet annotation handler clean calls are inserted during mangling and are thus invisible to clients.  The solution there for now is for drreg to look for the DR_NOTE_ANNOTATION label that is added in the app2app phase.

We could move just the clean call insertion to a client in the insertion phase, or try to move the annotation insruction sequence identification: but that may be harder to move.

Alternatively, we could collapse drmgr into the core, and then move the handler clean call to the drmgr insertion phase.

EIther approach would eliminate the drreg special-casing of annotation clean calls, and make them visible to instru2instru and other client actions.

PR #5164, opened 2021-10-14
i#4128: Add cleancall reg read/write flags

### LINKED ISSUE #4128: Call drreg_restore_all prior to clean calls
In order to for clean calls to have access to a machine context that contains original register/flag values, we need to call drreg's drreg_restore_all(). Essentially, the issue is that due to drreg's lazy restoration, registers may not have their original values upon invoking a clean call.

drreg_restore_app_values() is not appropriate as it only works on operands.

derekbruening:
Adds new dr_cleancall_save_t flags which are required for proper
interaction between clean calls and drreg:
DR_CLEANCALL_READS_APP_CONTEXT must be set for dr_get_mcontext() to
obtain the proper values, and #DR_CLEANCALL_WRITES_APP_CONTEXT must be
set to ensure that dr_set_mcontext() is persistent.

Adds a clean call insertion event to enable drreg to know about clean
calls at the time they are inserted.  dr_insert_clean_call_ex()
invokes the callback and passes the flags to drreg, who then treats
the clean call as an app instruction.

For annotations, for now we leave drreg looking for the annotation
label (possible future changes #5160 or #5161 would eliminate this
special case).

dr_insert_{cbr,ubr,mbr,call}_instrumentation() always set both labels.

drwrap always sets both labels for pre and post callbacks.

Updates uses throughout our tests and samples to use the new flags as
appropriate.

Adds a new dedicated test client.drwrap-drreg-test which tests both a
drwrap call and a direct clean call.

Fixes a missing drwrap cache invalidation on module unload that the
new test uncovers.

Fixes #4128

### DIFF
--- a/ext/drreg/drreg.c
+++ b/ext/drreg/drreg.c
@@ -1592,8 +1592,13 @@ drreg_spill_aflags(void *drcontext, instrlist_t *ilist, instr_t *where, per_thre
                                pt->live_idx) == REG_LIVE) {
             spill_reg(drcontext, pt, DR_REG_XAX, xax_slot, ilist, where);
             pt->reg[DR_REG_XAX - DR_REG_START_GPR].ever_spilled = true;
-        } else
+        } else {
+            /* XXX: Re-analyzing this: we shouldn't need this slot?
+             * drreg_move_aflags_from_reg() undoes it for this xax-is-dead case; can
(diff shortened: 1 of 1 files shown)


derekbruening:
Unfortunately it is looking like #4711 is reproducing

PR #5170, opened 2021-10-20
Fix drreg ever_spilled inaccuracy

derekbruening:
Fixes a drreg bug in drreg_spill_aflags() which set xax's ever_spilled
flag to true even if it wasn't spilled because it was dead.  This
caused requests to get the app value of xax to return success and to
move aflags out of xax and into a slot, but without restoring the app
value.  The fixed behavior is to fail up front if xax is dead and so
leave aflags in xax.

Updates the drreg tests that rely on getting the xax value moving
aflags into a slot to use a live xax rather than a dead xax.

This was discovered while adding PR #5164 for #4128.

Issue: #4128


abhinav92003:
Taking some more time to review this as I'm trying to convince myself of some stuff locally.

Issue #3844, opened 2019-09-17
Provide support in drreg to spill something other than GPRs, cross-app instructions.


hgreving2304:
Specifically currently hitting this issue is #2985. The case needs some capability to spill - in this case - xmm and a mask register value, cross-app instruction (hence mcontext can not be used). This issue covers providing a method to spill into the TLS data that drreg already has, managed by drmgr.

Interface TBD:
// Specifically allow for reserving xmm and mask registers? Also needs an unreserve method.
drreg_reserve_xmm_register() + drreg_reserve_mask_register()?
// Instructing drreg to spill arbitrary number of based from opnd_t? Also needs a restore method.
drreg_reserve_spill_bytes()?
// Others ideas?

Case #2985 needs this cross-app, hence no automatic management is needed. Should new API be written to provide automatic spill/restore management when called from insertion phase?

johnfxgalea:
I already implemented this functionality but haven't yet cleaned it up for a PR. I can handle this issue, should you wish, but I only have some time during the weekend.

hgreving2304:
Always great to have contributions like that. We do need it rather sooner than later. Are you able to share what you have, maybe sneak-peek style, so we could evaluate if that's what we want and we go from there?

johnfxgalea:
An old version is available publicly [here](https://github.com/johnfxgalea/dynamorio/blob/final_xmm_spill/ext/drreg/drreg.c). My private fork which I use for my PhD has contains bug fixes and improv performance. I'll try and get a branch in during the weekend.

hgreving2304:
Ok. We also need to be able to spill sth other than xmm. For example specifically, AVX-512 mask registers (k0, k1, ... k7). Maybe consider having API that is able to spill an arbitrary register opnd_t? I am assigning this issue to you for now.

johnfxgalea:
Hello @hgreving2304 - I created a branch with drreg capable of spilling and restoring XMM registers. Shall we take a step-by-step approach and focus on getting the branch merged and then see what your needs are specifically? If so, I'll include tests now before any further API changes.

johnfxgalea:
Pushed the branch now.

hgreving2304:
Hi can I look at the branch somewhere? Didn't see. The problem with making this specific to xmm is that I was hoping we can somehow make it more [general](https://github.com/DynamoRIO/dynamorio/issues/3844#issuecomment-532757148). Not sure if that's the best approach but have you thought about that in your design?

[...]

PR #5182, opened 2021-11-01
i#2985 scatter-gather: Preserve scratch xmm.

abhinav92003:
Preserves the scratch xmm register by spilling it before the scatter/gather
instr expansion, and restoring it after. We use memory obtained from the
thread-private heap, using dr_thread_alloc, as the spill slot.

Reserves TLS slot for every thread in drx, to store the allocated pointer to the
xmm spill slot.

Extends drx's state restore to also restore the app value of the spilled xmm
register.

An alternate approach was to extend drreg to make it capable of spilling xmm
regs too, but that was overkill as the scatter/gather expansion requires only
a small subset of that functionality that is very easily implemented as a set of
manual spill and restore. When that support is available in drreg (#3844) we
can use it here too; added a TODO.

Also extends the pure-asm scatter/gather test to verify that the scratch reg
(which is xmm0 in the test, given that we always select the lowest-numbered
available xmm reg as scratch) value is preserved across the scatter/gather
instrs.

Also fixes an issue with comparison of xmm regs in the test. `cmpss` writes
the result of comparison to the first reg, not the aflags. (We had replaced
`vpcmpud` with `cmpss` in 106bf952d9c7aca020d52a03e2f3adc33cb208b7 due to some `SIGILL` issues.)

Issues: #2985

### DIFF
--- a/ext/drx/drx.c
+++ b/ext/drx/drx.c
@@ -2832,7 +2832,7 @@ restore_spilled_xmm_value(drx_state_machine_params_t *params)
     ASSERT(params->spilled_xmm_slot_addr != NULL,
            "No spill address recorded for the app xmm value");
     ASSERT(params->spilled_xmm != DR_REG_NULL && reg_is_strictly_xmm(params->spilled_xmm),
-           "no spilled xmm recorded");
+           "No spilled xmm reg recorded");
     void *spilled_xmm_val = params->spilled_xmm_slot_addr;
     for (int i = 0; i < XMM_REG_SIZE; i++) {
         xmm_val[i] = *((byte *)spilled_xmm_val + i);
@@ -2900,9 +2900,8 @@ drx_avx2_gather_sequence_state_machine(void *drcontext,
     switch (params->detect_state) {
     /* First we detect the spill address for the scratch xmm reg. */
     case DRX_DETECT_RESTORE_AVX2_GATHER_EVENT_STATE_0: {
-        ASSERT(params->spilled_xmm_slot_addr == NULL &&
-                   params->spilled_xmm_slot_addr_reg == DR_REG_NULL,
-               "Spill address for xmm must be undetermined yet");
+        params->spilled_xmm_slot_addr_reg = DR_REG_NULL;
+        params->spilled_xmm_slot_addr = NULL;
         ptr_int_t spilled_xmm_slot_addr;
         if (instr_is_mov_constant(&params->inst, &spilled_xmm_slot_addr) &&
             opnd_is_reg(instr_get_dst(&params->inst, 0)) &&
@@ -2931,8 +2930,6 @@ drx_avx2_gather_sequence_state_machine(void *drcontext,
             advance_state(DRX_DETECT_RESTORE_AVX2_GATHER_EVENT_STATE_2, params);
             break;
         }
-        params->spilled_xmm_slot_addr_reg = DR_REG_NULL;
-        params->spilled_xmm_slot_addr = NULL;
         skip_unknown_instr_inc(DRX_DETECT_RESTORE_AVX2_GATHER_EVENT_STATE_0, params);
         break;
     /* We come back to DRX_DETECT_RESTORE_AVX2_GATHER_EVENT_STATE_2 for each
@@ -2952,9 +2949,7 @@ drx_avx2_gather_sequence_state_machine(void *drcontext,
                 break;
             }
         }
-        /* We don't need to ignore any instructions here, because we are already in
-         * DRX_DETECT_RESTORE_AVX2_GATHER_EVENT_STATE_2.
-         */
+        skip_unknown_instr_inc(DRX_DETECT_RESTORE_AVX2_GATHER_EVENT_STATE_0, params);
         break;
     case DRX_DETECT_RESTORE_AVX2_GATHER_EVENT_STATE_3:
         ASSERT(params->the_scratch_xmm != DR_REG_NULL,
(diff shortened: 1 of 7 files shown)


ext/drx/drx.c
```diff
@@ -126,10 +137,42 @@ drx_event_restore_state(void *drcontext, bool restore_memory,
...
  * INIT
  */
 
+#ifdef PLATFORM_SUPPORTS_SCATTER_GATHER
+static void
+drx_thread_init(void *drcontext)
+{
+    per_thread_t *pt = (per_thread_t *)dr_thread_alloc(drcontext, sizeof(*pt));
+    pt->scratch_xmm_spill_slot = dr_thread_alloc(
+        dr_get_current_drcontext(), XMM_REG_SIZE + 31 /* for 32-byte alignment */);
```
derekbruening, line 162: `drcontext` (else can have mismatches and problems)

derekbruening: Add XMM_ALIGNMENT and use for all the 31 (-1) and 32 values

ext/drx/drx.c
```diff
@@ -2358,7 +2416,22 @@ drx_expand_scatter_gather(void *drcontext, instrlist_t *bb, OUT bool *expanded)
             scratch_xmm != reg_resize_to_opsz(sg_info.gather_dst_reg, OPSZ_16))
             break;
     }
-    /* FIXME i#2985: spill scratch_xmm using a future drreg extension for simd. */
+    /* Spill the scratch xmm.
+     * TODO i#3844: drreg does not support spilling xmm regs yet, so we do it ourselves.
+     * When that support is available, replace the following with the required drreg API
+     * calls.
```
derekbruening, line 2422: What about restore_state?  I don't see any change here: are thread redirects disallowed while scratch_xmm is clobbered?

ext/drx/drx.c
abhinav92003: Done.

ext/drx/drx.c
abhinav92003: Added.

ext/drx/drx.c
```diff
@@ -2791,6 +2825,21 @@ skip_unknown_instr_inc(int reset_state, drx_state_machine_params_t *params)
...
+restore_spilled_xmm_value(drx_state_machine_params_t *params)
+{
+    byte xmm_val[XMM_REG_SIZE];
+    ASSERT(params->spilled_xmm_slot_addr != NULL,
+           "No spill address recorded for the app xmm value");
+    ASSERT(params->spilled_xmm != DR_REG_NULL && reg_is_strictly_xmm(params->spilled_xmm),
+           "No spilled xmm reg recorded");
+    void *spilled_xmm_val = params->spilled_xmm_slot_addr;
+    for (int i = 0; i < XMM_REG_SIZE; i++) {
+        xmm_val[i] = *((byte *)spilled_xmm_val + i);
```
derekbruening, line 2838: Doing dwords at a time or using memcpy should be faster

ext/drx/drx.c
```diff
@@ -2846,23 +2898,60 @@ drx_avx2_gather_sequence_state_machine(void *drcontext,
...
+    }
+    case DRX_DETECT_RESTORE_AVX2_GATHER_EVENT_STATE_1:
+        ASSERT(params->spilled_xmm == DR_REG_NULL,
+               "Spilled xmm reg must be undetermined yet");
+        ASSERT(params->spilled_xmm_slot_addr != NULL &&
+                   params->spilled_xmm_slot_addr_reg != DR_REG_NULL,
+               "xmm spill address must be determined already");
+        if (instr_get_opcode(&params->inst) == OP_vmovdqa &&
+            opnd_is_memory_reference(instr_get_dst(&params->inst, 0)) &&
+            opnd_get_base(instr_get_dst(&params->inst, 0)) ==
```
derekbruening, line 2924: opnd_is_memory_reference() returns true for opnd types that don't have bases: safer to change it to opnd_is_base_disp()

ext/drx/drx.c
```diff
@@ -3086,23 +3182,60 @@ drx_avx512_scatter_sequence_state_machine(void *drcontext,
...
+        break;
+    }
+    case DRX_DETECT_RESTORE_AVX512_SCATTER_EVENT_STATE_1:
+        ASSERT(params->spilled_xmm == DR_REG_NULL,
+               "Spilled xmm reg must be undetermined yet");
+        ASSERT(params->spilled_xmm_slot_addr != NULL &&
+                   params->spilled_xmm_slot_addr_reg != DR_REG_NULL,
+               "xmm spill address must be determined already");
+        if (instr_get_opcode(&params->inst) == OP_vmovdqa &&
+            opnd_is_memory_reference(instr_get_dst(&params->inst, 0)) &&
```
derekbruening, line 3207: Ditto.

ext/drx/drx.c
```diff
@@ -3334,23 +3472,60 @@ drx_avx512_gather_sequence_state_machine(void *drcontext,
...
+        break;
+    }
+    case DRX_DETECT_RESTORE_AVX512_GATHER_EVENT_STATE_1:
+        ASSERT(params->spilled_xmm == DR_REG_NULL,
+               "Spilled xmm reg must be undetermined yet");
+        ASSERT(params->spilled_xmm_slot_addr != NULL &&
+                   params->spilled_xmm_slot_addr_reg != DR_REG_NULL,
+               "xmm spill address must be determined already");
+        if (instr_get_opcode(&params->inst) == OP_vmovdqa &&
+            opnd_is_memory_reference(instr_get_dst(&params->inst, 0)) &&
```
derekbruening, line 3497: Ditto.

derekbruening (approved):

ext/drx/drx.c
abhinav92003: Done.

ext/drx/drx.c
abhinav92003: Done.

ext/drx/drx.c
abhinav92003: Done.

ext/drx/drx.c
abhinav92003: Done.

Issue #5166, opened 2021-10-15
drcachesim.drstatecmp-fuzz test failed on A64 with invalid instruction


derekbruening:
Looks like an error in the instruction synthesis?

http://139.178.84.19:8080/job/DynamoRIO-AArch64-Precommit/824/consoleFull
```
252/255 Test #192: code_api|tool.drcachesim.drstatecmp-fuzz ........................***Failed  Required regular expression not found. Regex=[^Generate code
Hit delay threshold: enabling tracing.
Execute generated code
All done
---- <application exited with code 0> ----
.*
$
]  9.14 sec
Generate code
Hit delay threshold: enabling tracing.
Usage error: opnd_create_reg: invalid register (/var/lib/jenkins/workspace/DynamoRIO-AArch64-Precommit/core/ir/instr_inline_api.h, line 209)
```

Issue #2374, opened 2017-04-21
AArch64 decoder should have option to get just opcode and flag usage


egrimley:
This option would be used by decode_opcode() and decode_eflags_usage(). A full decode of operands adds significant overhead.

PR #5183, opened 2021-11-01
i#5166: test if random inst is decodeable before fully decoding in drstatecmp-fuzz test

sapostolakis:
Fixes a raised assertion in drstatecmp-fuzz test when an invalid register
is passed to opnd_create_reg. The solution is to avoid fully decoding the
random instruction but first test whether it is valid and supported by
the decoder.

Fixes: #5166

### DIFF
--- a/core/ir/decode_api.h
+++ b/core/ir/decode_api.h
@@ -57,6 +57,16 @@ DR_API
 byte *
 decode_eflags_usage(void *drcontext, byte *pc, uint *usage, dr_opnd_query_flags_t flags);
 
+#ifdef AARCH64
+DR_API
+/**
+ * Attempts to decode the instruction at address \p pc to determine if the
+ * instruction is supported by the decoder.
+ * Returns false on decoding an invalid or unsupported instruction.
+ */
+decode_supported_op(void *drcontext, byte *pc);
+#endif /* AARCH64 */
+
 DR_API
 /**
  * Decodes the instruction at address \p pc into \p instr, filling in the
--- a/suite/tests/client-interface/drstatecmp-fuzz-app.c
+++ b/suite/tests/client-interface/drstatecmp-fuzz-app.c
@@ -142,20 +142,22 @@ generate_inst(byte *encode_pc, size_t *skipped_insts)
 {
     /* Pick a random side-effect-free and non-branch instruction. */
     uint32_t encoded_inst = generate_encoded_inst();
+    byte *encoded_inst_bytes = (byte *)&encoded_inst;
 
     /* Try to decode the randomized encoding. */
     instr_t *decoded_inst = instr_create(GLOBAL_DCONTEXT);
-    byte *nxt_pc = decode(GLOBAL_DCONTEXT, (byte *)&encoded_inst, decoded_inst);
-    /* XXX: Ideally the decoder would report as erroneous any encoding leading to SIGILL.
-     * Currently, several valid decodings are illegal instructions.
-     */
-    if (nxt_pc != NULL && check_decoded_inst(decoded_inst)) {
-        encode_pc = append_instr(decoded_inst, encode_pc);
-    } else {
-        (*skipped_insts)++;
-        instr_destroy(GLOBAL_DCONTEXT, decoded_inst);
+    if (decode_supported_op(GLOBAL_DCONTEXT, encoded_inst_bytes)) {
+        byte *nxt_pc = decode(GLOBAL_DCONTEXT, encoded_inst_bytes, decoded_inst);
+        /* XXX: Ideally the decoder would report as erroneous any encoding leading to
+         * SIGILL. Currently, several valid decodings are illegal instructions.
+         */
+        if (nxt_pc != NULL && check_decoded_inst(decoded_inst)) {
+            encode_pc = append_instr(decoded_inst, encode_pc);
+            return encode_pc;
+        }
     }
-
+    (*skipped_insts)++;
+    instr_destroy(GLOBAL_DCONTEXT, decoded_inst);
     return encode_pc;
 }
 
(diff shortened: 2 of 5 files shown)


sapostolakis:
The assertion in #5166 seems to be caused by calls to `opnd_create_reg` in `decode_common`. 
There needs to be a more lightweight decoding API for AArch64 to address this. Not sure whether what I did is ideal. Seems that more decode API was exposed in the past (saw some comments regarding decoding levels).

derekbruening:
> The assertion in #5166 seems to be caused by calls to `opnd_create_reg` in `decode_common`. There needs to be a more lightweight decoding API for AArch64 to address this. Not sure whether what I did is ideal. Seems that more decode API was exposed in the past (saw some comments regarding decoding levels).

Are you looking for #2374?

suite/tests/client-interface/drstatecmp-fuzz-app.c
```diff
@@ -142,20 +142,22 @@ generate_inst(byte *encode_pc, size_t *skipped_insts)
...
-     */
-    if (nxt_pc != NULL && check_decoded_inst(decoded_inst)) {
-        encode_pc = append_instr(decoded_inst, encode_pc);
-    } else {
-        (*skipped_insts)++;
-        instr_destroy(GLOBAL_DCONTEXT, decoded_inst);
+    if (decode_supported_op(GLOBAL_DCONTEXT, encoded_inst_bytes)) {
+        byte *nxt_pc = decode(GLOBAL_DCONTEXT, encoded_inst_bytes, decoded_inst);
+        /* XXX: Ideally the decoder would report as erroneous any encoding leading to
+         * SIGILL. Currently, several valid decodings are illegal instructions.
```
derekbruening, line 152: There are deliberate opcodes that raise SIGILL: UDF being the prime example.  Probably an explicit check for UDF should appear here?

core/ir/decode_api.h
```diff
@@ -57,6 +57,17 @@ DR_API
 byte *
 decode_eflags_usage(void *drcontext, byte *pc, uint *usage, dr_opnd_query_flags_t flags);
 
+#ifdef AARCH64
+DR_API
+/**
+ * Attempts to decode the instruction at address \p pc to determine if the
+ * instruction is supported by the decoder.
+ * Returns false on decoding an invalid or unsupported instruction.
```
derekbruening, line 65: This does not seem a necessary interface addition: an invalid instruction will already return NULL and set OP_INVALID; an unsupported instruction will have OP_xx.

Is the instruction causing the test failure OP_xx?

sapostolakis: The problem is with both unsupported and invalid ops, although I think the aarch64 decoder will classify both as OP_xx. In [decode_common](https://github.com/DynamoRIO/dynamorio/blob/0496c7fe38a1fb69c89a94ea989cd1da07e515c7/core/ir/aarch64/codec.c#L4524) if the decoder returns false (due to an invalid or an unsupported inst)  then it will still populate various fields of the instr and call opnd_create_reg. This is fine for just an unsupported instruction that is well-formed, but with a completely random and invalid instruction, this code could cause the assertion you reported.

sapostolakis:
> > The assertion in #5166 seems to be caused by calls to `opnd_create_reg` in `decode_common`. There needs to be a more lightweight decoding API for AArch64 to address this. Not sure whether what I did is ideal. Seems that more decode API was exposed in the past (saw some comments regarding decoding levels).
> 
> Are you looking for #2374?

#2374 wants a more efficient way to partially decode. For x86 there is support for partial decoding, while for AArch64 even the code for `decode_eflags_usage` calls full decoding (decode_common) and `decode_opcode` is not supported (just the opcode would not be enough for this issue) . For this issue, since the problem is within `decode_common`, I would need to call `decoder` instead of the `decode_common`. Ideally, one could change the actual decoder and check the validity of the instruction, but this definitely needs a lot of work and might be even hard to check the whole instruction for validity without properly decoding.

suite/tests/client-interface/drstatecmp-fuzz-app.c
sapostolakis: good idea. will add.

core/ir/decode_api.h
derekbruening: I see: so that is a bug in the decoder.  But how does it get an invalid reg value: wouldn't it use 4 bits and any 4-bit value is ok, or is this reg 31 not being ok somewhere?  It seems best to fix this in the decoder: is it a small fix?

sapostolakis: I couldn't reproduce the issue and without a callstack it is unclear who called `opnd_create_reg` leading to the assertion. However, we know that it is during generating the random code, and it should be when decoding the random inst. So, it is inside `decode_common`, either in the call to the [decoder](https://github.com/DynamoRIO/dynamorio/blob/0496c7fe38a1fb69c89a94ea989cd1da07e515c7/core/ir/aarch64/codec.c#L4524) or in the code immediately after. The decoder calls various functions to create reg operands that are defined in [codec.c](https://github.com/DynamoRIO/dynamorio/blob/980099e42d7ce0cec55800aeefb0357d9274e98b/core/ir/aarch64/codec.c). I checked all the uses of `opnd_create_reg` in `codec.c` and none of them seem to be able to produce a DR_REG_INVALID or > DR_REG_LAST_ENUM  . The only weird case is in the `decode_common` where the decoder returns false and  instead of using `extract_uint(enc, pos, n)` as in all other places it directly does the shift and the mask without forcing an uint. It does not seem likely but is it possible that the result of this shift and mask could be interpreted as a 5-bit negative number or can some subcomputation be interpreted as signed? Probably worth changing this part in any case to conform with the norm in codec.c.

PR #5185, opened 2021-11-02
i#5076 trace invariants, part 7: View single threads

derekbruening:
Augments the view tool to show a single thread when the new option
"-only_thread <tid>" is passed.  This is done in parallel mode for
efficiency.

Updates the view test with a simple thread filter test.

Issue: #5076

### DIFF
--- a/clients/drcachesim/common/options.cpp
+++ b/clients/drcachesim/common/options.cpp
@@ -374,6 +374,12 @@ droption_t<std::string> op_tracer_ops(
     "(For internal use: sweeps up tracer options)",
     "This is an internal option that sweeps up other options to pass to the tracer.");
 
+droption_t<int>
+    op_only_thread(DROPTION_SCOPE_FRONTEND, "only_thread", 0,
+                   "Only analyze this thread (0 means all)",
+                   "For simulator types that support it, limits analyis to the single "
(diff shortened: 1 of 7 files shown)


derekbruening:
32-bit error is a weird git fetch issue: presumably transient; re-running.

clients/drcachesim/common/options.cpp
```diff
@@ -374,6 +374,12 @@ droption_t<std::string> op_tracer_ops(
     "(For internal use: sweeps up tracer options)",
     "This is an internal option that sweeps up other options to pass to the tracer.");
 
+droption_t<int>
+    op_only_thread(DROPTION_SCOPE_FRONTEND, "only_thread", 0,
```
abhinav92003, line 378: Is there an easy way for the user to find out which tids are available?

abhinav92003 (approved):

clients/drcachesim/common/options.cpp
derekbruening: One specific use case is a thread id printed out by the invariant checker tool.  Other sources include from the file names of the thread files, from the basic_counts tool, or an unfiltered view run.  I think there are lots of sources and use cases.

PR #5186, opened 2021-11-03
i#5076 trace invariants, part 8: Report error location

derekbruening:
Tracks the thread and memref count for reporting with each error
found, for use with the view tool to see the context around the error.

Adds location checking to the invariant test.

Issue: #5076

### DIFF
--- a/clients/drcachesim/tests/invariant_checker_test.cpp
+++ b/clients/drcachesim/tests/invariant_checker_test.cpp
@@ -53,7 +53,7 @@ public:
     }
     std::vector<std::string> errors;
     std::vector<memref_tid_t> error_tids;
-    std::vector<uint64_t> error_refs;
+    std::vector<uint64_t> error_refs; // Memref count (ordinal) at error point.
 
 protected:
     void
--- a/clients/drcachesim/tests/invariant_checker_test.cpp
+++ b/clients/drcachesim/tests/invariant_checker_test.cpp
@@ -52,19 +52,26 @@ public:
     {
     }
     std::vector<std::string> errors;
+    std::vector<memref_tid_t> error_tids;
+    std::vector<uint64_t> error_refs;
 
 protected:
     void
-    report_if_false(bool condition, const std::string &message) override
+    report_if_false(per_shard_t *shard, bool condition,
+                    const std::string &message) override
     {
-        if (!condition)
+        if (!condition) {
             errors.push_back(message);
+            error_tids.push_back(shard->tid);
+            error_refs.push_back(shard->ref_count);
+        }
     }
 };
 
 /* Assumes there are at most 3 threads with tids 1, 2, and 3 in memrefs. */
 bool
 run_checker(const std::vector<memref_t> &memrefs, bool expect_error,
+            memref_tid_t error_tid = 0, uint64_t error_refs = 0,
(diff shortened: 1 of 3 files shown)


clients/drcachesim/tools/invariant_checker.h
```diff
@@ -91,12 +91,15 @@ class invariant_checker_t : public analysis_tool_t {
         bool found_instr_count_marker_ = false;
         uint64_t last_instr_count_marker_ = 0;
         std::string error;
+        // Track the location of errors.
+        memref_tid_t tid = -1;
+        uint64_t ref_count = 0;
```
abhinav92003, line 96: This seems more like a `ref_id`/`ref_ordinal` than `ref_count`

abhinav92003 (approved):

clients/drcachesim/tools/invariant_checker.h
derekbruening: I don't follow: it is not an "id".  It is the count of memref trace entries to this point in this thread.

abhinav92003: Right, so `ref_ordinal` makes the most sense then. IMHO, "ordinal" is clearer than "count", as the latter may be confused to mean the total count or something.

derekbruening: But this is the per-thread counter that is used to count the references: this is not a stored snapshot for one error.  So this request, to me, implies that every counter variable would have to be named "ordinal" as it's not the total count until the very end when all the increments are done.

abhinav92003: Probably okay to keep it `ref_count`.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -52,19 +52,26 @@ class checker_no_abort_t : public invariant_checker_t {
     {
     }
     std::vector<std::string> errors;
+    std::vector<memref_tid_t> error_tids;
+    std::vector<uint64_t> error_refs;
```
abhinav92003, line 56: Maybe add a comment to say that these refs should be interpreted as ordinals of the per-thread memrefs where some error was found.

abhinav92003 (approved):

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -52,19 +52,26 @@ class checker_no_abort_t : public invariant_checker_t {
...
```
abhinav92003, line 57: Does the view tool currently print at regular intervals the count of memrefs already printed? If not, will be good to add it; it'll make finding the memrefs from `error_refs` easiers.

clients/drcachesim/tests/invariant_checker_test.cpp
derekbruening: Done.

clients/drcachesim/tests/invariant_checker_test.cpp
derekbruening: No.  Probably it would be a per-line prefix, like line numbers in less -- not sure it would need an option to turn off since the numbers can get quite large.  This would be a separate PR.

PR #5188, opened 2021-11-03
i#5131 new Linux syscalls: fix clone3 struct handling

abhinav92003:
Fixes DR's handling of the clone_args struct used by the app. While creating our
own copy of the app's clone_args object, we need to allocate as much space as
specified by the app in the clone3 syscall's args. We should not use the size of
DR's internal struct copy (clone3_syscall_args_t) for this, as it may contain
less/more fields than the one used by the app.

Replaces memcpy with d_r_safe_read to properly handle the case where
reading of the user-provided clone args faults. Adds checks to match kernel's
behaviour for EINVAL and EFAULT.

Also fixes an existing bug where we were fixing syscall params and freeing mem
in the fork case as well.

Fixes handling for the !CLONE_VM case in clone3, to match what we do for the
same in clone.

Adds test for the !CLONE_VM case, and for some cases where clone3 should
return EINVAL or EFAULT.

Performs some refactoring: makes asm code in linux.clone test more verbose by
using explicit move/load/store  to set up registers for the syscall; use separate
routine for pre-syscall handling of clone3; use DR's copy of clone args and saved
flags everywhere, instead of reading the app clone args without a safe-read.

Issue: #5131


core/unix/signal.c
```diff
@@ -674,9 +674,14 @@ create_clone_record(dcontext_t *dcontext, reg_t *app_thread_xsp, app_pc thread_f
...
+                     *
+                     * These args are not really clone3_syscall_args_t*. The app
+                     * uses the clone_args provided by the kernel, which may have
+                     * more fields than DR's internal struct (clone3_syscall_args_t)
+                     * We also don't want to use the kernel's clone_args struct here,
+                     * because it may not be available on all LINUX versions.
                      */
-                    clone3_syscall_args_t *dr_clone_args,
-                    clone3_syscall_args_t *app_clone_args)
+                    void *dr_clone_args, void *app_clone_args)
```
abhinav92003, line 684: @derekbruening Looking for some feedback on this. Calling these `clone3_syscall_args_t*` doesn't seem correct, though it would make some of the code below look better.

Other DR-internal copies of kernel structs usually are meant to be kept in sync with their kernel version. `clone3_syscall_args_t` doesn't need to be. This is why I thought maybe it shouldn't be an alias for `clone_args`.

derekbruening: See above: I would use the type.  It is a standard practice for these sized structs to have varying fields off the end without having to throw away type info and have a formless blob.

core/unix/signal.c
```diff
@@ -712,8 +717,9 @@ create_clone_record(dcontext_t *dcontext, reg_t *app_thread_xsp)
              * cl_args->stack. But we expect the highest (non-inclusive)
              * in the clone record's app_thread_xsp.
              */
-            record->app_thread_xsp = app_clone_args->stack + app_clone_args->stack_size;
-            record->clone_flags = app_clone_args->flags;
+            record->app_thread_xsp = ((clone3_syscall_args_t *)app_clone_args)->stack +
```
derekbruening, line 720: These field accesses should check the size to confirm they are present.

core/unix/os.c
```diff
@@ -6935,8 +6934,10 @@ pre_system_call(dcontext_t *dcontext)
                 ", exit_signal = 0x" HEX64_FORMAT_STRING
                 ", stack = 0x" HEX64_FORMAT_STRING ", stack_size = 0x" HEX64_FORMAT_STRING
                 "\n",
-                cl_args->flags, cl_args->exit_signal, cl_args->stack,
-                cl_args->stack_size);
+                ((clone3_syscall_args_t *)cl_args)->flags,
```
derekbruening, line 6937: This is where the size should be checked: on field accesses to confirm these fields are present

core/unix/include/clone3.h
```diff
@@ -24,10 +29,6 @@ typedef struct {
     __aligned_u64 exit_signal;
     __aligned_u64 stack;
     __aligned_u64 stack_size;
-    __aligned_u64 tls;
```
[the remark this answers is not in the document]
abhinav92003: This struct doesn't need to be kept in sync with the latest kernel, unless some new field is added that DR needs to be aware of. Is there an advantage to keeping these fields that DR doesn't really care about? Is it completeness? In future, more fields can be added by the kernel here which would make this struct incomplete anyway. So, keeping only these here without committing to keep the struct in sync seems neither here nor there.

core/unix/signal.c
abhinav92003: I don't think we need to worry about fields like `stack`, `stack_size` and `flags` being absent. I cannot imagine `clone3` without these essential args. Also, the [old struct that Assad observed](https://github.com/DynamoRIO/dynamorio/issues/5131#issuecomment-958971904) also has these fields.

abhinav92003: If in future DR starts accessing some new field added by the kernel, we should add the size check for those accesses. But it doesn't seem necessary for these ever-present fields.

abhinav92003: The earliest version of clone_args, in v5.3: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=8f6ccf6159aed1f04c6d179f61f6fb2691261e84.

It has all fields till tls.

core/unix/os.c
abhinav92003: See reply in comment thread above.

core/unix/os.c
```diff
@@ -6921,9 +6921,8 @@ pre_system_call(dcontext_t *dcontext)
          */
         uint64_t flags;
         if (dcontext->sys_num == SYS_clone3) {
-            clone3_syscall_args_t *cl_args = (clone3_syscall_args_t *)sys_param(
-                dcontext, SYSCALL_PARAM_CLONE3_CLONE_ARGS);
-            flags = cl_args->flags;
+            void *cl_args = (void *)sys_param(dcontext, SYSCALL_PARAM_CLONE3_CLONE_ARGS);
```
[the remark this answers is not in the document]
abhinav92003: > The downsides of losing the type and having to cast to access fields outweighs the downsides of accidentally referencing a field beyond the size.

I see your point. I'll revert the part that removes the struct type.

core/unix/include/clone3.h
```diff
@@ -15,7 +15,12 @@
```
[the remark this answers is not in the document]
abhinav92003: Done.

core/unix/signal.c

core/unix/os.c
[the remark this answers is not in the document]

core/unix/include/clone3.h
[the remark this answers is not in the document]
derekbruening: My logic is:

A) We have to handle varying sizes anyway, so there is no downside to including fields we don't actively read in our code.
B) If we need to access a new field added below these, we'll have to add them all anyway later.
C) I'd prefer having something here that's closer to what recent kernels have, for completeness and easier understanding.
D) These fields are not going to change so they will not get stale.  New fields will be appended is all.

Basically I don't see a downside to including, I don't see an upside to deliberately removing, and there are modest upsides to keeping.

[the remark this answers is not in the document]
abhinav92003: Keeping it closer to recent kernels and having to update anyway in future does make sense. Reverted these changes.

core/unix/os.c
```diff
@@ -6988,12 +6988,21 @@ pre_system_call(dcontext_t *dcontext)
...
+                 * syscall's args.
+                 */
+                clone3_syscall_args_t *dr_clone_args =
+                    (clone3_syscall_args_t *)heap_alloc(
+                        dcontext, app_clone_args_size HEAPACCT(ACCT_OTHER));
                 clone3_syscall_args_t *app_clone_args =
                     (clone3_syscall_args_t *)sys_param(dcontext,
                                                        SYSCALL_PARAM_CLONE3_CLONE_ARGS);
-                memcpy(dr_clone_args, app_clone_args, sizeof(clone3_syscall_args_t));
+                memcpy(dr_clone_args, app_clone_args, app_clone_args_size);
```
derekbruening, line 7005: As discussed offline, let's make this a d_r_safe_read() to avoid DR crashing if an app passes bogus memory to the kernel.

core/unix/os.c
```diff
@@ -6988,12 +6988,30 @@ pre_system_call(dcontext_t *dcontext)
...
+                        dcontext, app_clone_args_size HEAPACCT(ACCT_OTHER));
                 clone3_syscall_args_t *app_clone_args =
                     (clone3_syscall_args_t *)sys_param(dcontext,
                                                        SYSCALL_PARAM_CLONE3_CLONE_ARGS);
-                memcpy(dr_clone_args, app_clone_args, sizeof(clone3_syscall_args_t));
+                if (!d_r_safe_read(app_clone_args, app_clone_args_size, dr_clone_args)) {
+                    LOG(THREAD, LOG_SYSCALLS, 2,
+                        "\treturning EFAULT to app for clone3\n");
+                    set_failure_return_val(dcontext, EFAULT);
+                    DODEBUG({ dcontext->expect_last_syscall_to_fail = true; });
```
abhinav92003, line 7009: I see some other code where `expect_last_syscall_to_fail` is set to `true` *and* `execute_syscall` to `false`. I see that `expect_last_syscall_to_fail` is read in `post_system_call`, which doesn't execute in case of `!execute_syscall`. Then why do we need to set `expect_last_syscall_to_fail` at all?

There are some other uses of `expect_last_syscall_to_fail` in `core/win32/syscall.c`, but I assume they are separate.

core/unix/os.c

core/unix/os.c
derekbruening: It does seem to be technically unnecessary.

core/unix/os.c
```diff
@@ -8612,8 +8630,10 @@ post_system_call(dcontext_t *dcontext)
              * post-syscall natively anyway, there's no need to restore here.
              */
 #    endif
-            HEAP_TYPE_FREE(dcontext, dcontext->sys_param1, clone3_syscall_args_t,
-                           ACCT_OTHER, PROTECTED);
+            uint app_clone_args_size =
+                (uint)sys_param(dcontext, SYSCALL_PARAM_CLONE3_CLONE_ARGS_SIZE);
```
derekbruening, line 8634: Generally we avoid sys_param() in post-syscall b/c some registers (and sometimes memory) are clobbered and it's easy to make a mistake with which ones are not for different system call gateways: so we always store in the cache vars sys_paramX to avoid problems.

core/unix/os.c
```diff
@@ -6988,12 +6988,30 @@ pre_system_call(dcontext_t *dcontext)
...
+                if (!d_r_safe_read(app_clone_args, app_clone_args_size, dr_clone_args)) {
+                    LOG(THREAD, LOG_SYSCALLS, 2,
+                        "\treturning EFAULT to app for clone3\n");
+                    set_failure_return_val(dcontext, EFAULT);
+                    DODEBUG({ dcontext->expect_last_syscall_to_fail = true; });
+                    execute_syscall = false;
+                    heap_free(dcontext, dr_clone_args,
+                              app_clone_args_size HEAPACCT(ACCT_OTHER));
+                    break;
+                }
```
derekbruening, line 7014: I'm still not sure we can ignore the size: if I write an app that sets the size to very small so that it does not cover the fields we read, presumably the kernel will return EINVAL instead of EFAULT, and DR will crash (well it will read bogus memory so prob won't crash: will assume syscall will succeed and might mess some things up; do we handle syscall failing when we thought it would succeed (that's a tough topic in some cases though)?  Always best to anticipate failure accurately.

derekbruening (approved):

core/unix/signal.c
abhinav92003: Now we return `EINVAL` if the clone args size passed by the app is less than the required.

core/unix/os.c
abhinav92003: Resolved above.

core/unix/os.c
abhinav92003: Used `sys_param`.

core/unix/os.c
abhinav92003: Yes, the kernel has a check for size, and if it is less than the ver0 size of clone_args, it returns EINVAL: https://elixir.bootlin.com/linux/v5.15/source/kernel/fork.c#L2725. I made changes to match the kernel's EINVAL and EFAULT behaviour. This also guarantees that the fields that we need (stack, stack_size, flags etc) will be accessible.

core/unix/os.c
abhinav92003: I guess there's no harm in keeping it when the others have it. In case we're missing something.

core/unix/os.c
```diff
@@ -8612,8 +8653,9 @@ post_system_call(dcontext_t *dcontext)
              * post-syscall natively anyway, there's no need to restore here.
              */
 #    endif
-            HEAP_TYPE_FREE(dcontext, dcontext->sys_param1, clone3_syscall_args_t,
-                           ACCT_OTHER, PROTECTED);
+            uint app_clone_args_size = (uint)dcontext->sys_param2;
+            heap_free(dcontext, (clone3_syscall_args_t *)dcontext->sys_param1,
```
abhinav92003, line 8657: @derekbruening Checking as I'm not completely sure: for the CLONE_VM case, it's probably not sufficient to free this memory just in the parent, and we need to do the same in the child too right?

core/unix/os.c
abhinav92003: Also added a test for it.

abhinav92003:
The Windows failures are unrelated, but looks like they need some intervention: `The remote server returned an error: (401) Unauthorized.`

core/unix/os.c
abhinav92003: Oh that would be required in the `!CLONE_VM` case.

abhinav92003: Hmm.. let me fix this now.

abhinav92003: Done.

suite/tests/linux/clone.c
```diff
@@ -287,7 +294,17 @@ create_thread_clone3(void (*fcn)(void), void **stack, bool share_sighand)
...
+    assert(errno == EFAULT);
+
+    ret = make_clone3_syscall((void *)0x123 /* bogus address */,
+                              sizeof(struct clone_args), fcn);
+    assert(errno == EFAULT);
+
+    ret = make_clone3_syscall(&cl_args, CLONE_ARGS_SIZE_MIN_POSSIBLE - 1, fcn);
+    assert(errno == EINVAL);
+
+    ret = make_clone3_syscall(&cl_args, sizeof(struct clone_args), fcn);
```
derekbruening, line 307: Seems good to have a test of fork (i.e. w/o CLONE_VM) using clone3 too.

suite/tests/linux/clone.c
```diff
@@ -215,19 +220,19 @@ create_thread(int (*fcn)(void *), void *arg, void **stack, bool share_sighand)
...
 #        ifdef X64
     asm volatile("syscall\n\t"
                  "test %%rax, %%rax\n\t"
                  "jnz parent\n\t"
                  "call *%%rdx\n\t"
                  "parent:\n\t"
                  : "=a"(ret)
-                 : "0"(SYS_clone3), "D"(clone_args), "S"(sizeof(struct clone_args)),
-                   "d"(fcn)
+                 : "0"(SYS_clone3), "D"(clone_args), "S"(clone_args_size), "d"(fcn)
```
derekbruening, line 235: (I never learned all these codes: I assume they are loading into specific registers up front.  Personally I find explicit setup and named parameters the easiest to read like at https://github.com/DynamoRIO/dynamorio/blob/master/suite/tests/linux/rseq.c#L184)

derekbruening (approved):

core/unix/os.c
```diff
@@ -6920,23 +6920,68 @@ pre_system_call(dcontext_t *dcontext)
...
+            app_clone_args_size =
+                (uint)sys_param(dcontext, SYSCALL_PARAM_CLONE3_CLONE_ARGS_SIZE);
+            if (app_clone_args_size < CLONE_ARGS_SIZE_VER0) {
+                LOG(THREAD, LOG_SYSCALLS, 2, "\treturning EINVAL to app for clone3\n");
+                set_failure_return_val(dcontext, EINVAL);
+                DODEBUG({ dcontext->expect_last_syscall_to_fail = true; });
+                execute_syscall = false;
+                break;
+            }
+            app_clone_args = (clone3_syscall_args_t *)sys_param(
```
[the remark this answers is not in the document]

core/unix/os.c
```diff
@@ -6920,23 +6920,68 @@ pre_system_call(dcontext_t *dcontext)
...
                 dcontext, SYSCALL_PARAM_CLONE3_CLONE_ARGS);
-            flags = cl_args->flags;
-            /* Save for post_system_call. As clone3 flags are 64-bit, even in 32-bit
+            if (app_clone_args == NULL) {
```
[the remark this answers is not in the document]
abhinav92003: Done. I think I added it before to avoid a NULL deref, but forgot that it's redundant in the final version.

core/unix/os.c
```diff
@@ -6920,23 +6920,68 @@ pre_system_call(dcontext_t *dcontext)
...
+            dcontext->sys_param1 = (reg_t)dr_clone_args;
+            dcontext->sys_param2 = (reg_t)app_clone_args_size;
             LOG(THREAD, LOG_SYSCALLS, 2,
                 "syscall: clone3 with args: flags = 0x" HEX64_FORMAT_STRING
                 ", exit_signal = 0x" HEX64_FORMAT_STRING
                 ", stack = 0x" HEX64_FORMAT_STRING ", stack_size = 0x" HEX64_FORMAT_STRING
                 "\n",
-                cl_args->flags, cl_args->exit_signal, cl_args->stack,
-                cl_args->stack_size);
+                app_clone_args->flags, app_clone_args->exit_signal, app_clone_args->stack,
```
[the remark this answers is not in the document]
abhinav92003: DOne.

core/unix/os.c
```diff
@@ -6920,23 +6920,68 @@ pre_system_call(dcontext_t *dcontext)
...
+            if (!d_r_safe_read(app_clone_args, app_clone_args_size, dr_clone_args)) {
+                LOG(THREAD, LOG_SYSCALLS, 2, "\treturning EFAULT to app for clone3\n");
+                set_failure_return_val(dcontext, EFAULT);
+                DODEBUG({ dcontext->expect_last_syscall_to_fail = true; });
+                execute_syscall = false;
+                heap_free(dcontext, dr_clone_args,
+                          app_clone_args_size HEAPACCT(ACCT_OTHER));
+                break;
+            }
+            flags = app_clone_args->flags;
```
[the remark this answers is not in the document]

Curious about how this case works though. You mean the app might be intercepting faults to the page and after the first fault due to the safe read, it may mark it inaccessible?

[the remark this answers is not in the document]
derekbruening: It may not be likely, and it may be difficult to write an app that hits the case (one example would be an app that has a thread that sits in a loop swapping the page protections between no-access and readable, and DR's safe read happens to occur when the page is readable) but it's a general TOCTOU error we should avoid to assume a page is accessible now just because it was a little while ago.  Besides, we have the copy, so there is no need to read the original.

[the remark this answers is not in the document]
abhinav92003: Where it was straight-forward, I changed the code to use the DR copy of clone args, instead of the app's. But there are some cases where it is not straight-forward. For `!CLONE_VM`, we do create a copy of the clone args to do the safe-read check (the check is required to read the `flags` field), but we free the copy pre-syscall. It is complex to keep it till post-syscall, as then we'd need to free it in the child and parent both.

There's also this check in `d_r_dispatch`; it seems like it might happen for the child too. We do not want to access the DR copy of clone args in the child thread, as there'll be a race with the parent free-ing it.
https://github.com/DynamoRIO/dynamorio/blob/eb2a333f859d5a9b89bf6d9dbb77f9f4b372811d/core/dispatch.c#L141

If the scenario you mentioned is unlikely, maybe it is better to avoid creating more complex code paths, and just live with multiple reads of the app clone args. Let me know what you think.

suite/tests/linux/clone.c
abhinav92003: Done.

suite/tests/linux/clone.c
abhinav92003: Done.

suite/tests/linux/clone.c
abhinav92003: I don't think there was a test without CLONE_VM for clone either. At least not in Linux.clone. Added that too.

core/unix/os.c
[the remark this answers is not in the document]
derekbruening: I think we should avoid de-referencing app memory unsafely.  If we have to read it again, we should use another safe read.  Or we should store the info we need.  For was_thread_create_syscall, we just need the flags.  We could store that.  For !CLONE_VM I'm not sure what we need.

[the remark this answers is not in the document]
abhinav92003: > Or we should store the info we need. For was_thread_create_syscall, we just need the flags.

flags is all we need from the struct post-syscall, and storing just that would indeed be the simplest. Unfortunately the `sys_paramX` slots in `dcontext_t` are of type `reg_t`, but we need 8 bytes for storing the flags on 32-bit systems as well.

We could perhaps split the flags across two `sys_paramX` variables. This option will use up all of the 5 `sys_paramX` slots in total. 
sys_param0: app_clone_flags, to restore in the sys param reg post-syscall
sys_param1: dr_clone_flags, for freeing post-syscall
sys_param2: app_clone_flags_size, for freeing dr_clone_flags post-syscall
sys_param3, sys_param4: flags split across two vars.

> If we have to read it again, we should use another safe read

This is the other option. But we would need to pay for an extra safe read of the whole struct. I guess the previous option of splitting the flags is better.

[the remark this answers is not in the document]
abhinav92003: Made the change. PTAL.

If in future we need to save anything else from `clone_args` for post-syscall use, we may need to resort to the second option of using more safe-reads.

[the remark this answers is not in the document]
derekbruening: Safe reads that do not fault are cheap so it is not a problem to do more of them.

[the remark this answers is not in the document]
abhinav92003: It's probably better to re-use the saved flags if possible, instead of doing a safe read each time and handling the possibility that the read may fault.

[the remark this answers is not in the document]
abhinav92003: So, for now I'll keep the current approach of using sys_param3 and sys_param4 for storing the 64-bit flags.

core/unix/os_private.h
```diff
@@ -114,6 +114,7 @@
 
 #define SYSCALL_PARAM_CLONE3_CLONE_ARGS 0
 #define SYSCALL_PARAM_CLONE3_CLONE_ARGS_SIZE 1
+#define CLONE3_FLAGS_4_BYTE_MASK 0xffffffff
```
derekbruening, line 117: (See comment in os.c.)

core/unix/os_exports.h
```diff
@@ -351,7 +351,7 @@ extern app_pc vsyscall_sysenter_displaced_pc;
 #define VSYSCALL_PAGE_MAPS_NAME "[vdso]"
 
 bool
-is_thread_create_syscall(dcontext_t *dcontext);
+is_thread_create_syscall(dcontext_t *dcontext _IF_LINUX(void *maybe_clone_args));
```
derekbruening, line 354: This is no longer exported so it should not be in os_exports.h anymore.

derekbruening (approved):

core/dispatch.c
```diff
@@ -2031,7 +2031,12 @@ handle_system_call(dcontext_t *dcontext)
         /* FIXME: move into some routine inside unix/?
          * if so, move #include of sys/syscall.h too
          */
-        if (is_thread_create_syscall(dcontext)) {
+        /* We use was_thread_create_syscall even though the syscall has not really
+         * happened yet. THis is because, for the clone3 syscall, we want to avoid
```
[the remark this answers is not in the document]
abhinav92003: Done.

core/unix/os.c
```diff
@@ -5346,29 +5346,46 @@ is_thread_create_syscall_helper(ptr_uint_t sysnum, reg_t sys_param0)
...
+     */
+    if (sysnum == SYS_clone3) {
+        ASSERT(maybe_clone_args != NULL);
+        flags = ((clone3_syscall_args_t *)maybe_clone_args)->flags;
+    }
+#endif
+    return is_thread_create_syscall_helper(sysnum, flags);
+}
+
+ptr_uint_t
```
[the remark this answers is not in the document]
abhinav92003: Renamed and made static.

core/unix/os.c
```diff
@@ -6551,6 +6568,162 @@ os_switch_seg_to_context(dcontext_t *dcontext, reg_id_t seg, bool to_app)
...
+        dcontext->sys_param0 = (reg_t)app_clone_args;
+        /* For freeing the allocated memory. */
+        dcontext->sys_param1 = (reg_t)dr_clone_args;
+        dcontext->sys_param2 = (reg_t)app_clone_args_size;
+        /* clone3 flags are 64-bit even on 32-bit systems. So we need to split them across
+         * two reg_t vars on 32-bit. We do it on 64-bit systems as well for simpler code.
+         */
+        dcontext->sys_param3 = (reg_t)(flags & CLONE3_FLAGS_4_BYTE_MASK);
+        ASSERT(((flags >> 32) & ~CLONE3_FLAGS_4_BYTE_MASK) == 0);
+        dcontext->sys_param4 = (reg_t)((flags >> 32));
```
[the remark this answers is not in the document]
abhinav92003: The one for `>>` can be dropped. I thought it makes it a bit better to read. Dropping it anyway now.

core/unix/os_exports.h
abhinav92003: Done.

core/unix/os_private.h
abhinav92003: Done.

core/unix/os.c
```diff
@@ -6551,6 +6568,162 @@ os_switch_seg_to_context(dcontext_t *dcontext, reg_id_t seg, bool to_app)
...
+         * post-syscall.
+         */
+        dcontext->sys_param0 = (reg_t)app_clone_args;
+        /* For freeing the allocated memory. */
+        dcontext->sys_param1 = (reg_t)dr_clone_args;
+        dcontext->sys_param2 = (reg_t)app_clone_args_size;
+        /* clone3 flags are 64-bit even on 32-bit systems. So we need to split them across
+         * two reg_t vars on 32-bit. We do it on 64-bit systems as well for simpler code.
+         */
+        dcontext->sys_param3 = (reg_t)(flags & CLONE3_FLAGS_4_BYTE_MASK);
```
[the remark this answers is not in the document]
abhinav92003: Done.

abhinav92003:
Thanks for the multiple rounds of review. It seems there were some cases that should've been handled when I added this support initially, like the `!CLONE_VM`, and the `EINVAL`/`EFAULT` cases.

abhinav92003:
Glad that those cases are handled properly now.

derekbruening:
The other issue is: would glibc try clone3 and if it gets back ENOSYS it falls back to clone?  Xref discussion at https://github.com/DynamoRIO/dynamorio/issues/5131#issuecomment-963101835 and https://github.com/DynamoRIO/dynamorio/issues/5131#issuecomment-963269395.  What happens if DR sees a clone3 that's going to fail with ENOSYS: does DR take some non-undoable action?  Is DR even trying to undo a clone syscall that fails (we do try to many syscalls)?

abhinav92003:
> #5131 (comment)

Added a TODO for that for now.

PR #5189, opened 2021-11-04
i#5076 trace invariants: Report parallel view error

derekbruening:
Fixes an omission from part 7 (PR #5185) where the view tools runs in
parallel but fails to report error strings for that mode.

Tested manually on a wrong-archtecture trace.

Issue: #5076


derekbruening:
Realizing it may be worth noting the view tool change in the changelog.  Adding here as a 2nd view tool update.

abhinav92003 (approved):

PR #5191, opened 2021-11-04
i#5190: drcachesim child stats include all lower levels

### LINKED ISSUE #5190: drcachesim child cache stats should include all lower levels
The child (==lower level) cache stats in drcachesim today are only for the immediately lower level.  This is a holdover from the original implementation which didn't support other than 2 levels.  For multiple levels, one would expect the child stats to include the sum of all the lower levels.

derekbruening:
The child (==lower level) cache stats in drcachesim today are only for
the immediately lower level. This is a holdover from the original
implementation which didn't support other than 2 levels. For multiple
levels, one would expect the child stats to include the sum of all the
lower levels.  We fix that here and add a unit test.

The difference is clear in the config file test:
  LLC stats:
     Child hits:                        147
=>
  LLC stats:
     Child hits:                    174,695

But to avoid flakiness we do not update its regex and rely on the unit
test.

Fixes #5190

### DIFF
--- a/clients/drcachesim/simulator/caching_device.cpp
+++ b/clients/drcachesim/simulator/caching_device.cpp
@@ -140,9 +140,7 @@ caching_device_t::request(const memref_t &memref_in)
         caching_device_block_t *cache_block =
             &get_caching_device_block(last_block_idx_, last_way_);
         assert(tag != TAG_INVALID && tag == cache_block->tag_);
-        stats_->access(memref_in, true /*hit*/, cache_block);
-        for (caching_device_t *up = parent_; up != nullptr; up = up->parent_)
-            up->stats_->child_access(memref_in, true, cache_block);
+        record_access_stats(memref_in, true /*hit*/, cache_block);
         access_update(last_block_idx_, last_way_);
         return;
     }
@@ -161,9 +159,7 @@ caching_device_t::request(const memref_t &memref_in)
             // Access is a hit.
             caching_device_block_t *cache_block = block_way.first;
             way = block_way.second;
-            stats_->access(memref, true /*hit*/, cache_block);
-            for (caching_device_t *up = parent_; up != nullptr; up = up->parent_)
-                up->stats_->child_access(memref, true, cache_block);
+            record_access_stats(memref, true /*hit*/, cache_block);
             if (coherent_cache_ && memref.data.type == TRACE_TYPE_WRITE) {
                 // On a hit, we must notify the snoop filter of the write or propagate
                 // the write to a snooped cache.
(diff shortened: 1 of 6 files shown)


clients/drcachesim/tests/drcachesim_unit_tests.cpp
```diff
@@ -235,6 +235,64 @@ unit_test_compulsory_misses()
...
+
+    const int num_accesses = 16;
+    for (int i = 0; i < num_accesses; i++) {
+        ref.data.addr = 64 + i;
+        if (!cache_sim.process_memref(ref)) {
+            std::cerr << "drcachesim unit_test_child_hits failed: "
+                      << cache_sim.get_error_string() << "\n";
+            exit(1);
+        }
+    }
```
abhinav92003, line 291: The cache line size is presumably 64. So `assert(num_accesses < 64)` too?

clients/drcachesim/tests/drcachesim_unit_tests.cpp
```diff
@@ -235,6 +235,64 @@ unit_test_compulsory_misses()
...
+    assert(cache_sim.get_cache_metric(metric_name_t::CHILD_HITS, 3, 0) ==
```
abhinav92003, line 292: For a slightly stronger check, maybe add  `&& cache_sim.get_cache_metric(metric_name_t::CHILD_HITS, 1, 0)`

clients/drcachesim/simulator/caching_device.cpp
```diff
@@ -185,7 +185,8 @@ caching_device_t::request(const memref_t &memref_in)
             missed = true;
             // If no parent we assume we get the data from main memory
             if (parent_ != NULL) {
-                parent_->stats_->child_access(memref, false, cache_block);
+                for (caching_device_t *up = parent_; up != nullptr; up = up->parent_)
```
abhinav92003, line 188: (Optional) Worth extracting this loop into a separate routine? `update_child_access`?

abhinav92003 (approved):

clients/drcachesim/tests/drcachesim_unit_tests.cpp
derekbruening: See the `line_size` right above.  Added a comment.  An assert seems redundant.  If something is messed up the stats will be off in any case.

clients/drcachesim/tests/drcachesim_unit_tests.cpp
derekbruening: Added several more stats asserts.

clients/drcachesim/simulator/caching_device.cpp
derekbruening: Done as part of changing the logic to propagate hits all the way but misses just one level.

derekbruening: Unfortunately there's no way to test the miss logic b/c dr$sim does not care about child misses: it's something we want for external uses that subclasses this.  Prob sufficient to have just external tests.

PR #5193, opened 2021-11-05
Upgrade GA CI ninja to 1.10.2

derekbruening:
We're hitting problems downloading 1.7.2 and it's good to upgrade in
any case.  This udpates the ninja in our CI jobs to 1.10.2.


abhinav92003 (approved):

derekbruening:
Seems to be working as the download succeeded and the build is ongoing.

PR #5194, opened 2021-11-05
i#5076 trace invariants: Fix concurrent thread writes

derekbruening:
Moves two invariant_checker_t data fields that are written
concurrently (though we expect their values to be constant across
threads) into per-thread data to avoid data race tools complaining.

Tested under ThreadSanitizer.

Issue: #5076


derekbruening:
win64 failure is drcachesim.coherence type_is_instr #3320 which nobody has been able to repro or figure out.

abhinav92003 (approved):

Issue #5123, opened 2021-09-25
ASSERT api.detach_state vmareas.c:8308 !ok || !TEST(MEMPROT_WRITE, prot2) || TEST(FRAG_SELFMOD_SANDBOXED, *flags) || !INTERNAL_OPTION(hw_cache_consistency)


derekbruening:
This happened once on GA CI:

https://github.com/DynamoRIO/dynamorio/pull/5119/checks?check_run_id=3708972532

```
<Application /home/runner/work/dynamorio/dynamorio/build_debug-internal-64/suite/tests/bin/api.detach_state (22501).  Internal Error: DynamoRIO debug check failure: /home/runner/work/dynamorio/dynamorio/core/vmareas.c:8308 !ok || !TEST(MEMPROT_WRITE, prot2) || TEST(FRAG_SELFMOD_SANDBOXED, *flags) || !INTERNAL_OPTION(hw_cache_consistency)
(Error occurred @136 frags in tid 22503)
```

Xref this assert showing up in various options of common.fib in #1807

derekbruening:
This is happening frequently in the long suite: 3 out of 35 runs which is enough redness that we may have to mark it flaky.  That's bad though as we don't want to lose this valuable test.

derekbruening:
I'm going to add it to the ignore list for the long suite only as a compromise short-term solution

derekbruening:
I ran 1000x on my local machine and did not reproduce any failure

jackgallagher-arm:
This error seems to be appearing much more frequently in the CI since I merged #7743 but I have been unable to reproduce it locally on x86 or AArch64.

PR #5201, opened 2021-11-11
Udpate transparency docs to reflect modern support

derekbruening:
Updates the docs on transparency to reflect that we have a private
loader and do support many typical uses of clients using external
libraries.


derekbruening:
x64 failure is detach_state #5123 vmareas.c assert

abhinav92003 (approved):

PR #5206, opened 2021-11-15
Fix missing reg encoding info for aarch64

JerryYouxin:
The register information (including `reg_get_size()` and register name `reg_names`) is lacked for AArch64, including, `DR_REG_CNTVCT_EL0`, `DR_REG_NZCV`, `DR_REG_FPCR`, and `DR_REG_FPSR`. This lack of information will cause client failure as follows (e.g., client tool called `reg_get_size(DR_REG_CNTVCT_EL0)`):

```
...
<Application /usr/bin/ls (60103) DynamoRIO usage error : reg_get_size: invalid register>
<Usage error: reg_get_size: invalid register (/home/yx/dynamorio/core/ir/opnd_shared.c, line 2538)
...
```

I have already added this information via this PR and added a CI test to early-detect similar information lacks for possible future extensions. These CI tests are listed as `code_api|client.reg_size_test`.


core/ir/aarch64/encode.c
```diff
@@ -44,42 +44,42 @@
```
derekbruening, line 44: Grammar errors: s/lacked/missing/ in the title and description

core/ir/opnd_shared.c
```diff
@@ -2531,10 +2531,16 @@ reg_get_size(reg_id_t reg)
         return OPSZ_SCALABLE;
     if (reg >= DR_REG_P0 && reg <= DR_REG_P15)
         return OPSZ_SCALABLE_PRED;
+    if (reg == DR_REG_CNTVCT_EL0)
+        return OPSZ_8;
+    if (reg >= DR_REG_NZCV && reg <= DR_REG_FPSR)
+        return OPSZ_8;
```
derekbruening, line 2537: Probably better to ask @AssadHashmi or another AArch64 expert for a review rather than me -- @AssadHashmi if you could confirm that these status registers are 64-bit despite having only a few fields?

suite/tests/client-interface/reg_size_test.dll.c
```diff
@@ -0,0 +1,38 @@
...
+#endif
+
+DR_EXPORT void
+dr_client_main(client_id_t id, int argc, const char *argv[])
+{
+    dr_set_client_name("DynamoRIO Sample Client 'test_reg_size'",
+                       "http://dynamorio.org/issues");
+
+    opnd_size_t res;
+    for (int i = DR_REG_NULL + 1; i < DR_REG_LAST_VALID_ENUM; ++i) {
```
derekbruening, line 14: Fencepost error since the last is inclusive: `s/</<=/`

suite/tests/client-interface/reg_size_test.dll.c
```diff
@@ -0,0 +1,38 @@
...
+        if (i == DR_REG_INVALID)
+            continue;
+        if (get_register_name(i)[0] == '\0') {
```
derekbruening, line 17: So this is not testing that there's a name: it's trying to skip reserved-and-unused enum values.  I think this is worth a comment as it's not clear why this is here as at first you'd think it's an error to have no name.

derekbruening: So there is no test for names and the missing "cntvct_el0" would not have been detected by this test.  It would read off the end of the array and return garbage.  How about checking that the size of the returned string is small and that each character is ascii alphanumeric?

suite/tests/client-interface/reg_size_test.dll.c
```diff
@@ -0,0 +1,38 @@
...
+            continue;
+        }
+#if defined(X86) && !defined(X64)
+        if (i >= REG_START_64 && i <= REG_STOP_64)
+            continue;
+        if (i >= REG_START_x64_8 && i <= REG_STOP_x64_8)
+            continue;
+        if (i >= DR_REG_STOP_XMM && i <= DR_REG_START_YMM)
```
derekbruening, line 25: (Hmm I guess there is no cleaner way to do this.)

suite/tests/client-interface/reg_size_test.dll.c
```diff
@@ -0,0 +1,38 @@
...
+            continue;
+        if (i >= DR_REG_STOP_YMM && i <= DR_REG_START_ZMM)
+            continue;
+        if (i >= DR_REG_STOP_ZMM && i <= RESERVED_ZMM)
```
derekbruening, line 29: (RESERVED_ZMM is not really public (not documented).  But for this test I guess it's ok.)

suite/tests/client-interface/reg_size_test.dll.c
```diff
@@ -0,0 +1,38 @@
...
+            continue;
+#endif
+        res = reg_get_size(i);
```
derekbruening, line 32: The return value needs to be checked for `OPSZ_NA`: otherwise this test is useless in release build!

suite/tests/client-interface/reg_size_test.dll.c
```diff
@@ -0,0 +1,38 @@
+#include "dr_api.h"
+
+#ifdef __cplusplus
+extern "C" {
```
derekbruening, line 4: This can be removed: this is a C file.

core/ir/opnd_shared.c
AssadHashmi: @derekbruening @JerryYouxin: yes, they're 64 bits and `INSTR_CREATE_mrs()`/`INSTR_CREATE_msr()` are used to access the specified user-space (EL0) fields.

core/ir/aarch64/encode.c
JerryYouxin: Fixed

core/ir/opnd_shared.c
JerryYouxin: I found these register size info from ARM Developer website: [CNTVCT_EL0](https://developer.arm.com/documentation/ddi0595/2020-12/AArch64-Registers/CNTVCT-EL0--Counter-timer-Virtual-Count-register?lang=en), [NZCV](https://developer.arm.com/documentation/ddi0595/2020-12/AArch64-Registers/NZCV--Condition-Flags?lang=en), [FPCR](https://developer.arm.com/documentation/ddi0595/2020-12/AArch64-Registers/FPCR--Floating-point-Control-Register?lang=en), [FPSR](https://developer.arm.com/documentation/ddi0595/2020-12/AArch64-Registers/FPSR--Floating-point-Status-Register?lang=en)

suite/tests/client-interface/reg_size_test.dll.c
JerryYouxin: I tried to add tests of register names in the last commit by checking the size and contents as you suggested.

suite/tests/client-interface/reg_size_test.dll.c
JerryYouxin: fixed

suite/tests/client-interface/reg_size_test.dll.c
JerryYouxin: fixed

suite/tests/client-interface/reg_size_test.dll.c
JerryYouxin: Now I use RESERVED_XXX macro to skip all reserved register enums to avoid checking register name/size of X86 registers.

derekbruening:
Note that we use re-review requests to know whose "turn" it is: https://dynamorio.org/page_code_reviews.html#autotoc_md114
"When the requested changes have been pushed, you can ask for a re-review from the reviewers. This can be done by clicking the re-review button next to the reviewers' name on the right sidebar on the pull request page."

core/ir/aarch64/encode.c
derekbruening: > JerryYouxin changed the title from `Fix lacked reg encoding info for aarch64` to `Fixs lacked missing reg encoding info for aarch64`

Maybe you're not familiar with the UNIX sed/perl/bash syntax `s/replace-this/with-this/`?  The ungrammatical part is `lacked` and it should be replaced with `missing`.  We now also have another grammar error `Fixs`.

suite/tests/client-interface/reg_size_test.dll.c
derekbruening: (Looks like this was done so I will resolve it.  The convention is for you to reply Done.)

suite/tests/client-interface/reg_size_test.dll.c
derekbruening: LGTM

derekbruening (approved):
Thank you for the patch!

core/ir/aarch64/encode.c
derekbruening: I will fix the title and description.

PR #5208, opened 2021-11-16
Use decode_from_copy in opcode_mix tool

derekbruening:
Updates the opcode_mix tool to use decode_from_copy() as a better
example of the proper way to decode offline trace instructions which
are mapped to a different address than their original location during
tracing.


derekbruening:
Windows failure is strange:
```
Error copying file "/.../dynamorio/build_debug-internal-32/lib32/debug/../drconfiglib.dll" to "/.../dynamorio/build_debug-internal-32/suite/tests/bin/drconfiglib.dll".
```
Re-running.  Presumably unrelated as this tiny change is not going to affect building.

abhinav92003 (approved):

PR #5217, opened 2021-11-18
i#5131 new linux syscalls: Fix aarch64 dup3 and close_range handling

abhinav92003:
Fixes an existing bug in dup3 handling on AArch64. As SYS_dup2 is not
available on AArch64, SYS_dup3 was not being included in the switch in
pre-syscall handling due to a misplaced conditional compile statement. Note
that on AArch64 dup2 automatically uses dup3 internally, so the bug affected
also the dup2 calls in the client.file_io test.

Fixes close_range handling on AArch64 by updating the syscall defines from a
newer header file that does define __NR_close_range. Defines the
SYS_close_range constant for AArch64.

Enables the client.file_io test on AArchXX. Replaces the inline asm code to
unmask divide-by-zero exceptions, with a glibc wrapper. Also adds a test for
dup3. The clone_range part of this test is enabled on AArchXX but hasn't been
run yet, because our CI machine does not have kernel >= 5.9.

Also, on AArch64, unmasking of float-point exceptions in the setup of the
client.file_io test may be ignored as trapping exceptions is optional.
https://code.woboq.org/userspace/glibc/sysdeps/aarch64/fpu/feenablxcpt.c.html#37

Issue: #5131

### DIFF
--- a/core/unix/os.c
+++ b/core/unix/os.c
@@ -7675,24 +7675,26 @@ pre_system_call(dcontext_t *dcontext)
 #endif
         break;
     }
-#ifdef SYS_dup2
-    case SYS_dup2:
+#if defined(SYS_dup2) || defined(LINUX)
         IF_LINUX(case SYS_dup3:)
-        {
-            file_t newfd = (file_t)sys_param(dcontext, 1);
-            if (fd_is_dr_owned(newfd) || fd_is_in_private_range(newfd)) {
-                SYSLOG_INTERNAL_WARNING_ONCE("app trying to dup-close DR file(s)");
-                LOG(THREAD, LOG_TOP | LOG_SYSCALLS, 1,
-                    "WARNING: app trying to dup2/dup3 to %d.  Disallowing.\n", newfd);
-                if (DYNAMO_OPTION(fail_on_stolen_fds)) {
-                    set_failure_return_val(dcontext, EBADF);
-                    DODEBUG({ dcontext->expect_last_syscall_to_fail = true; });
-                } else
-                    set_success_return_val(dcontext, 0);
-                execute_syscall = false;
-            }
-            break;
+#    ifdef SYS_dup2
+    case SYS_dup2:
+#    endif
+    {
+        file_t newfd = (file_t)sys_param(dcontext, 1);
+        if (fd_is_dr_owned(newfd) || fd_is_in_private_range(newfd)) {
+            SYSLOG_INTERNAL_WARNING_ONCE("app trying to dup-close DR file(s)");
+            LOG(THREAD, LOG_TOP | LOG_SYSCALLS, 1,
+                "WARNING: app trying to dup2/dup3 to %d.  Disallowing.\n", newfd);
+            if (DYNAMO_OPTION(fail_on_stolen_fds)) {
+                set_failure_return_val(dcontext, EBADF);
+                DODEBUG({ dcontext->expect_last_syscall_to_fail = true; });
+            } else
+                set_success_return_val(dcontext, 0);
+            execute_syscall = false;
         }
+        break;
+    }
 #endif
 
 #ifdef MACOS
--- a/core/unix/os.c
+++ b/core/unix/os.c
@@ -7679,8 +7679,10 @@ pre_system_call(dcontext_t *dcontext)
 #endif
         break;
     }
-#if defined(SYS_dup2) || defined(LINUX)
-        IF_LINUX(case SYS_dup3:)
+#if defined(SYS_dup2) || defined(SYS_dup3)
+#    ifdef SYS_dup3
+    case SYS_dup3:
+#    endif
 #    ifdef SYS_dup2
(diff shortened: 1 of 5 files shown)


suite/tests/linux/clone.c
```diff
@@ -245,10 +245,6 @@ make_clone3_syscall(struct clone_args *clone_args, uint clone_args_size,
 #        endif
 #    elif defined(AARCH64)
     uint64 clone_args_size_64 = (uint64)clone_args_size;
-    /* Do not add code between above declarations and their use below.
```
abhinav92003, line 248: Found this stray comment. It's unrelated to this PR, but I thought I'll clean it up.

suite/tests/client-interface/file_io.c
```diff
@@ -270,9 +308,32 @@ main()
      */
 #ifdef WINDOWS
     _control87(_MCW_EM & (~_EM_ZERODIVIDE), _MCW_EM);
+#elif defined(LINUX)
+    if (feenableexcept(FE_DIVBYZERO) == -1) {
+#    ifdef AARCH64
+        /* This call may return EPERM on AArch64.
+         * https://code.woboq.org/userspace/glibc/sysdeps/aarch64/fpu/feenablxcpt.c.html#37
```
abhinav92003, line 315: I see the same on https://www.qnx.com/developers/docs/7.0.0/#com.qnx.doc.neutrino.lib_ref/topic/f/feenableexcept.html. Maybe worth adding Assad to understand this limitation better?

abhinav92003: Also on https://code.woboq.org/userspace/glibc/sysdeps/aarch64/fpu/feenablxcpt.c.html#37 as mentioned in the PR' description.

suite/tests/client-interface/file_io.c
```diff
@@ -99,24 +108,53 @@ main()
...
     /* Test dup/close for stolen FDs. */
     for (int i = rlimit.rlim_max; i < rlimit.rlim_max + DR_STEAL_FDS; i++) {
-        if (dup2(0, i) != -1 || errno != EBADF)
-            fprintf(stderr, "Expected dup2 to return EBADF for stolen FD %d\n", i);
+        if (IF_LINUX_ELSE(i % 2 == 0, true)) {
+            /* Note that on AArch64 dup2 is not available and is automatically replaced
+             * with dup3.
+             */
+            if (dup2(0, i) != -1 || errno != EBADF)
+                fprintf(stderr, "Expected dup2 to return EBADF for stolen FD %d\n", i);
```
derekbruening, line 116: Generally we use print() to avoid flushing issues with stderr on Windows.  Matters less for output we don't check for but best to be consistent and not have someome use this as a model for output-compared printing.

suite/tests/client-interface/file_io.c
derekbruening: So if the AArch64 hardware doesn't support floating-point traps, then this will return EPERM, and we do not fail the test.  But we don't then have a test on hardware that does have traps; we could then have something untested and not realize it.  Not sure failing the test is good either.  It's similar to rseq or close_range or other things: we manually make sure our test machines have the feature, and we have the tests gracefully skip that if not present.

derekbruening (approved):

suite/tests/client-interface/file_io.c
abhinav92003: This test uses `fprintf(stderr,...` everywhere. I didn't put much thought into it and just assumed that there was some  reason for it. Do you suggest changing it everywhere in this test?

core/unix/os.c
```diff
@@ -7675,24 +7679,26 @@ pre_system_call(dcontext_t *dcontext)
 #endif
         break;
     }
-#ifdef SYS_dup2
-    case SYS_dup2:
+#if defined(SYS_dup2) || defined(LINUX)
```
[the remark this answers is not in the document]

suite/tests/client-interface/file_io.c
derekbruening: Yes I would replace them all with print().  Xref https://dynamorio.org/page_test_suite.html#autotoc_md272

abhinav92003: Done.

suite/tests/client-interface/file_io.c
abhinav92003: @AssadHashmi Is this something that can be solved by updating the kernel of the Jenkins machine? The first link says it's a limitation of the hardware (unmasking of floating-point exceptions being ignored), which makes me think updating the kernel might not help.

On a related note, there are some syscalls that aren't supported by the current kernel there, notably `close_range`. Not sure if it's easy to update the kernel on that machine. It is 5.4 I think, which is ~1.5 years old.

abhinav92003: I guess we can follow up offline. Resolving for now.

PR #5220, opened 2021-11-19
i#5131 new linux syscalls: fail early for clone3 enosys

abhinav92003:
Adds support to detect at init time whether the clone3 syscall is supported by
the system or not. Use this to fail early if a clone3 syscall is made and avoid
doing the pre-syscall work.

Adds a test that attempts the same clone3 syscall as the os.c code and checks
that it returns EINVAL on systems that support clone3 and ENOSYS on systems
that don't.

Manually verified on a system with Linux kernel 4.15 (that does not have the
clone3 syscall) that DR's pre-syscall handling of clone3 indeed returns early
with this fix.

Fixes the size of the result var for the syscall in the test, which should be 64-bit
on 64-bit systems.

Issue: #5131, #5221


core/unix/os.c
```diff
@@ -482,6 +482,11 @@ our_libc_errno_loc(void)
  */
 typedef int *(*errno_loc_t)(void);
 
+#ifdef LINUX
+/* Stores whether clone3 is unsupported on the system we're running on. */
+static bool is_clone3_enosys = false;
```
abhinav92003, line 487: I considered making this a bitmap, but didn't as it's not needed right now. But I'm open to changing.

derekbruening: I think most existing cases are DR using a syscall, not the app: e..g, SYS_memfd_create.  That could conceivably be moved to init time instead of every call to os_create_memory_file().  Maybe that could be done here; or a comment could be added.

Looking at other new syscalls the app uses: rseq has only post-syscall handling.

core/unix/os.c
```diff
@@ -838,6 +843,72 @@ get_uname(void)
...
+#    if defined(ARM)
+    /* TODO i#5221: Implement for ARM. */
+#    else
+    long result;
+    /* We know that clone3 fails with EINVAL with these args. */
+    ulong clone_args_size = 0;
+    void *clone_args = NULL;
+#        ifdef X86
+#            ifdef X64
+    asm volatile("mov %[sys_clone3], %%rax\n\t"
```
derekbruening, line 859: Replace all of this assembly with a single dynamorio_syscall() (i.e., we already have per-arch asm to make syscalls; no need to duplicate it here).

core/unix/os.c
abhinav92003: I added a comment for now, to not block the release work on cleanup.

suite/tests/linux/clone.c
```diff
@@ -204,14 +226,15 @@ create_thread(int (*fcn)(void *), void *arg, void **stack, bool share_sighand,
...
-#    ifdef X86
-#        ifdef X64
-    uint64 clone_args_size_64 = (uint64)clone_args_size;
+#ifdef SYS_clone3
+    assert(CLONE3_SYSCALL_NUM == SYS_clone3);
+#endif
+    long result;
+#ifdef X86
+#    ifdef X64
     asm volatile("mov %[sys_clone3], %%rax\n\t"
```
derekbruening, line 238: In fact I remember now that we deliberately export dynamorio_syscall to tests inside drlibc so it could be used here as well to simplify this code.

derekbruening (approved):

suite/tests/linux/clone.c
abhinav92003: It's better to use inline asm for this `clone3` call. If we use `dynamorio_syscall`, the child thread will also return from it and will try to use some variables local to `make_clone3_syscall` (like storing the ret value in `result`). This is fragile in the `CLONE_VM` case as it may cause a crash if enough space is not left on the new stack for such accesses. Xref https://github.com/DynamoRIO/dynamorio/pull/5143#discussion_r726347743.

It was okay to use `dynamorio_syscall` in the os.c code because we do not expect the child thread to be successfully created there, so there won't ever be any such issues.

derekbruening: Looking at what dr_create_client_thread() does: it uses dynamorio_clone().  Is that in drlibc?

derekbruening: I suppose it doesn't use clone3 though, at least now in the current impl.

abhinav92003: > Looking at what dr_create_client_thread() does: it uses dynamorio_clone()

I see that `dynamorio_clone` also redirects the child thread in asm code itself, probably due to the same reason (avoiding return-to-caller for child thread) https://github.com/DynamoRIO/dynamorio/blob/75d7b48c5843d312a29f1afeea86fc6643fb7531/core/arch/x86/x86.asm#L1594

> I suppose it doesn't use clone3 though, at least now in the current impl

Right, so we cannot use it here.

abhinav92003: Assuming this is resolved.

Issue #5145, opened 2021-10-05
Release DynamoRIO v9.0.0


abhinav92003:
From offline discussion with @derekbruening: we want to release a new version of DR soon. It has been a while since the last release, which was in April 2020. We have made several stability fixes since then, particularly some related to memory management options. As some users hit issues with OOM recently (see discussions on https://groups.google.com/g/dynamorio-users), possibly due to sub-optimal defaults, it'll be good to create a new release.

Task list for some more fixes to include in the release is as follows. We can add more to this list if needed.

- [X] #5144 (@AssadHashmi)
- [x] #5131 (@abhinav92003)
- [X] #4128 (@derekbruening)
- [X] #2414 (@derekbruening)
- [X] #5190 (@derekbruening)
- [X] samples.dox errors + missing samples (@derekbruening)
- [X] transparency library wording (@derekbruening)
- [X] broken pubs links (@derekbruening)
- [X] #5198 (@derekbruening)
- [X] #4958 (@derekbruening)
- [X] #2891
- [X] mark -attach as experimental in drrun output and deployment docs (done in PR #5227)
- [ ] #2626
- [ ] #4400
- [ ] #4532
- [x] #5233
- [x] DynamoRIO/drmemory#2440
- [x] #5054
- [x] Add AArch64 opcode enum reorderings to changelog
- [x] #2985

Nice to have:
- [ ] #38 for AArch64

derekbruening:
I may try to implement a libunwind version of #2414 but would not be able to put SEH64 in there in time so it should not gate the release since it would not be an all-platform feature.

derekbruening:
The other thing is that contributors added externally-triggered attach support (Window: #725 PR #5075; Linux: #38  PR #5019) which is a major feature that we could list for 9.0, but there is no test at all for Linux which makes it hard to claim it works.

@M3m3M4n if you could add an attach test we could then announce attach support for the 9.0 release.

@OrBenPorath if you get a chance to move the Windows attach test off the flaky list that would be great.

derekbruening:
Xref #1995

Should fix #2891 docs minor issues

M3m3M4n:
I'm quite tight on time until the end of the year. Will try to add the tests in my free time.
Also I think #5054 need to be fixed first, I haven't read too much into DR core aside from attaching part, so need more time to iron things out.

derekbruening:
[...]

PR #5232, opened 2021-11-29
i#5145 v9.0: Update version number to 9.0

derekbruening:
Updates the default version number in the sources and package commands
to 9.0.

Updates the oldest compatible version to 9.0.

Issue: #5145


derekbruening:
x64 failure is detach_state #5123 -- re-running just in case.

abhinav92003 (approved):

derekbruening:
The plan is to wait for #4400 and #4532 for the release so I'll hold off merging this.

derekbruening:
x64 failure is again detach_state #5123.  Not sure what's up with Windows.  Will have to re-run on merging w/ master in any case.

PR #5234, opened 2021-11-30
i#4532: Remove superfluous non-operands from AArch64 memory instructions

joshua-warburton:
Several AArch64 memory related instructions have various
bits designated as "soft bits." This means that the spec
suggests a value of 1 for them, but it is allowed for the
bits to vary from that.

Previously these bits were represented as operands, but
that caused stray values to show up in the IR. This change
introduces a way to represent soft bits in codec.txt which
tells the decoder to ignore them and the encoder to set them
to be 1.

This change also updates the encoding macros to remove the
unnecessary operands and updates the appropriate tests. The
decode tests have been modified to allow the re-encoding to
differ from the original encoding. This is because the soft
bits can be values not strictly dictated by the spec but as
we will always re-encode to spec, the re-encoding will be
different to the initial encoding.

Fixes: #4532


core/ir/aarch64/codec.py
```diff
@@ -143,80 +152,87 @@ def indent_append(text):
...
+            bit_not_set_or_variable = 0
+            bit_set_or_variable = 0
+            for p in pats:
+                # In how many patterns is this bit not set or included
+                # in the variable bits
+                if (1 << switch_bit) & (~p.opcode_bits | p.opnd_bits | p.high_soft_bits):
+                    bit_not_set_or_variable += 1
+                # how many patterns have this b set and or in the variable
+                # bits
+                if (1 << switch_bit) & (p.opcode_bits | p.opnd_bits| p.high_soft_bits):
```
derekbruening, line 221: nit: `s/bits|/bits |/`

suite/tests/api/dis-a64.c
```diff
@@ -121,41 +123,73 @@ error(const char *line, size_t len, const char *s)
...
+    uint decode_length;
+    uint expected_enc = 0;
+    if (fields == 3) {
+        decode_start = field_start[2];
+        decode_length = field_end[2] - field_start[2];
+    } else {
+        // Sometimes we don't expect the encoded value to match the initial value,
+        // such as when there are bits that are allowed to vary in the spec (soft bits).
+        // In these cases there will be an additional field at position 3 with an
+        // encoding, wind forward looking for a :, making sure we note the current
```
derekbruening, line 181: writing nit: `s/encoding, wind/encoding.  Wind/` plus add ending `.`

suite/tests/api/dis-a64.txt
```diff
@@ -32,9 +32,20 @@
...
 # The script "dis-a64.pl" can generate a draft version of this file.
 
-# Field 1: Encoding.
-# Field 2: Disassembly produced by objdump (not used for testing).
-# Field 3: Disassembly produced by DynamoRIO.
+# This file contains colon-separated fields that are used to
+# test the decoder.
+
+# The first field contains the hex encoding of the instruction and
+# its operands
```
derekbruening, line 39: nit: add ending `.` for all of these

suite/tests/api/ir_aarch64.expect
```diff
@@ -839,23 +839,23 @@ sys    $0x1bd9 (%x29)[1byte]
...
+stxrb  %w0 -> (%x1)[1byte] %w2
+stxrh  %w0 -> (%x1)[2byte] %w2
 stxp   %w0 %w1 -> (%x2)[8byte] %w3
 stxp   %x0 %x1 -> (%x2)[16byte] %w3
-stlxr  %x0 $0x000000000000001f -> (%x1)[8byte] %w2
-stlxrb %w0 $0x000000000000001f -> (%x1)[1byte] %w2
-stlxrh %w0 $0x000000000000001f -> (%x1)[2byte] %w2
+stlxr  %x0 -> (%x1)[8byte] %w2
+stlxrb %w0 -> (%x1)[1byte] %w2
+stlxrh %w0 -> (%x1)[2byte] %w2
```
derekbruening, line 858: There's no `smulh` or `umulh` here?  Do we expect that?  Maybe not everything is printed so those opcode changes in  `dis-a64.txt` are tested without any output here.

derekbruening (approved):
Thanks for the cleanup.  I don't think any changes are required in any tools since the memory operand and register operand placements didn't change.

AssadHashmi (approved):
Patch reviewed internally prior to PR submission.

suite/tests/api/ir_aarch64.expect
joshua-warburton: I only updated the existing encoding tests, and I don't think smulh or umulh had any so they won't appear in this file. Dis-a64.txt doesn't use an expect to check its output

PR #5235, opened 2021-12-01
i#4400: AArch64 v8.0 codec: add cache management instructions

AssadHashmi:
This patch splits out DC and IC instruction aliases from
the SYS instruction and makes DC ZVA a store instruction.

Issues: #4400, #4382, #4329, #2626


AssadHashmi:
@derekbreuning, @abhinav92003: In addition to addressing #4329, #4382 this patch also makes DC ZVA into a store instruction which wasn't possible before (see https://github.com/DynamoRIO/dynamorio/issues/4400#issuecomment-686089419).
What else needs to be done for #4400? Is this adequate for the v9.0 release?

clients/drcachesim/tracer/instru.cpp
```diff
@@ -185,36 +185,30 @@ instru_t::instr_to_prefetch_type(instr_t *instr)
...
+    case OP_dc_civac:
+    case OP_dc_cvac: return true;
     }
     return false;
 }
 
 bool
 instru_t::is_aarch64_dc_zva_instr(instr_t *instr)
 {
-    // TODO i#4393: Split OP_sys into multiple opcodes based on the operation.
```
derekbruening, line 215: Should this PR mention #4393 too?

clients/drcachesim/tracer/instru.cpp
```diff
@@ -185,36 +185,30 @@ instru_t::instr_to_prefetch_type(instr_t *instr)
...
-    return instr_get_opcode(instr) == OP_sys &&
-        opnd_get_immed_int(instr_get_src(instr, 0)) == DR_DC_ZVA;
+    return instr_get_opcode(instr) == OP_dc_zva;
```
derekbruening, line 211: Please also remove the #4400 workaround from raw2trace.h.
However: we actually can't remove it nor the helper code like instr_summary_t::is_aarch64_dc_zva since we need to handle traces gathered with a tracer prior to this PR.
This PR needs to bump the offline raw trace version (the final trace_version_t is unchanged) OFFLINE_FILE_VERSION and add conditional use of the #4400 workaround for older versions.

core/ir/aarch64/instr_create_api.h
```diff
@@ -108,20 +108,6 @@
 /** Create an operand specifying LSL, the default shift type when there is no shift. */
 #define OPND_CREATE_LSL() opnd_add_flags(OPND_CREATE_INT(DR_SHIFT_LSL), DR_OPND_IS_SHIFT)
 
-/** INSTR_CREATE_sys() operations are specified by the following immediates. */
-enum {
-    DR_DC_ZVA = 0x1ba1,    /**< Zero dcache by address. */
```
derekbruening, line 113: I guess this was added after the last official release so this removal does not need to be mentioned in the changelog as a compatibility break.

core/ir/aarch64/instr_create_api.h
```diff
@@ -738,6 +724,100 @@ enum {
...
+ *             No alignment restrictions apply to this VA.
+ */
+#define INSTR_CREATE_dc_ivac(dc, Rn) instr_create_0dst_1src(dc, OP_dc_ivac, Rn)
+
+/**
+ * Creates a DC ZVA instruction to Zero data cache by Virtual Address.
+ * Zeroes a naturally aligned block of N bytes, where N is identified in
+ * DCZID_EL0 system register.
+ * \param dc   The void * dcontext used to allocate memory for the #instr_t.
+ * \param Rn   The input register containing the virtual address to use.
```
derekbruening, line 792: But this is not a register: it is a memory operand, and its size is not OPSZ_sys but rather a cache line (the N above).  The docs here imply the macro will auto-convert it to a memory reference of the right size: but the test below passes in a memory reference (of the wrong size).

Similarly for all the other operations here that take in a virtual address.

clients/drcachesim/tracer/instru.cpp
```diff
@@ -185,36 +185,30 @@ instru_t::instr_to_prefetch_type(instr_t *instr)
 bool
```
derekbruening, line 185: The description says #4382 but I don't see any disasm changes here (if there were: I would vote for a separate PR) -- is that a typo?

core/ir/aarch64/codec.txt
```diff
@@ -294,1336 +294,1347 @@ x----------xxxxx----------------  wx16       # W/X register (or WZR/XZR)
...
-0x001110xx0xxxxx010110xxxxxxxxxx  n   558       uzp2            dq0 : dq5 dq16 bhsd_sz
-11010101000000110010000001011111  n   559        wfe                :
-11010101000000110010000001111111  n   560        wfi                :
-11010101000000110010000011111111  n   561    xpaclri                :
-00001110xx100001001010xxxxxxxxxx  n   562        xtn             d0 : d5 bhsd_sz
-01001110xx100001001010xxxxxxxxxx  n   563       xtn2             q0 : q5 bhsd_sz
-11010101000000110010000000111111  n   564      yield                :
-0x001110xx0xxxxx001110xxxxxxxxxx  n   565       zip1            dq0 : dq5 dq16 bhsd_sz
-0x001110xx0xxxxx011110xxxxxxxxxx  n   566       zip2            dq0 : dq5 dq16 bhsd_sz
+0101111011100000101110xxxxxxxxxx  n   6           abs             d0 : d5
```
derekbruening, line 303: Can't tell what changed here: an extra space?  Normally the diff would identify that as the only change on the line: are there other changes?

core/ir/aarch64/codec.txt
```diff
@@ -294,1336 +294,1347 @@ x----------xxxxx----------------  wx16       # W/X register (or WZR/XZR)
...
+00011010110xxxxx010110xxxxxxxxxx  n   73      crc32cw             w0 : w5 w16
+10011010110xxxxx010111xxxxxxxxxx  n   74      crc32cx             w0 : w5 x16
+00011010110xxxxx010001xxxxxxxxxx  n   75       crc32h             w0 : w5 w16
+00011010110xxxxx010010xxxxxxxxxx  n   76       crc32w             w0 : w5 w16
+10011010110xxxxx010011xxxxxxxxxx  n   77       crc32x             w0 : w5 x16
+x0011010100xxxxxxxxx00xxxxxxxxxx  r   78         csel            wx0 : wx5 wx16 cond
+x0011010100xxxxxxxxx01xxxxxxxxxx  r   79        csinc            wx0 : wx5 wx16 cond
+x1011010100xxxxxxxxx00xxxxxxxxxx  r   80        csinv            wx0 : wx5 wx16 cond
+x1011010100xxxxxxxxx01xxxxxxxxxx  r   81        csneg            wx0 : wx5 wx16 cond
+110101010000100001111110010xxxxx  n   569     dc_cisw                : x0
```
derekbruening, line 420: OK it wasn't easy but I found what really changed: these were inserted.  So this was possible w/o changing to xml as #4393 talked about?  Not even a change to the python script was needed?

derekbruening (changes requested):

core/ir/aarch64/instr_create_api.h
derekbruening: This also brings up the question of who back-aligns the address.  So this PR in its current state is mostly splitting the opcode (#4393) -- but #4400 is about making the operand reflect the actual memory range affected, which means aligning it and setting the size to the cache line size.  But xref the discussion in #4400 about how to get the cache line size for things like host != target (e.g., we run the AArch64 decoder on x86 machines).

clients/drcachesim/tracer/instru.cpp
abhinav92003: I don't think we can remove the #4400 workaround yet, as this PR doesn't back-align the dc zva address in any new way. When we do that, we should also add a test to verify back-alignment in online traces.

core/ir/aarch64/codec.txt
abhinav92003: > So this was possible w/o changing to xml as #4393 talked about? Not even a change to the python script was needed?

You mean why this PR doesn't make any python script changes to ensure ordering? The ordering related stuff was already handled by 2c4f1e5a30a98ba3061e78d358ed3b60377f8ce6, I think. IIUC this PR needs to only add new entries in the codec, which is what it does.

core/ir/aarch64/codec.txt
```diff
@@ -294,1336 +294,1347 @@ x----------xxxxx----------------  wx16       # W/X register (or WZR/XZR)
...
+11111000111xxxxx100000xxxxxxxxxx  n   480       swpal        x0 mem0 : x16 mem0
+00111000111xxxxx100000xxxxxxxxxx  n   481      swpalb        w0 mem0 : w16 mem0
+01111000111xxxxx100000xxxxxxxxxx  n   482      swpalh        w0 mem0 : w16 mem0
+00111000001xxxxx100000xxxxxxxxxx  n   483        swpb        w0 mem0 : w16 mem0
+01111000001xxxxx100000xxxxxxxxxx  n   484        swph        w0 mem0 : w16 mem0
+10111000011xxxxx100000xxxxxxxxxx  n   485        swpl        w0 mem0 : w16 mem0
+11111000011xxxxx100000xxxxxxxxxx  n   485        swpl        x0 mem0 : x16 mem0
+00111000011xxxxx100000xxxxxxxxxx  n   486       swplb        w0 mem0 : w16 mem0
+01111000011xxxxx100000xxxxxxxxxx  n   487       swplh        w0 mem0 : w16 mem0
+1101010100001xxxxxxxxxxxxxxxxxxx  n   488         sys                : sysops memx0
```
abhinav92003, line 1495: So we continue to have this entry. Is that for all variants of `sys` other than `dc` or `ic` (which are handled by a different entry above)?

I may have missed previous decoder changes that allow such overlapping entries in the codec (the bits of `sys` and `dc` match except the `x`s). What's the logic now? When there are multiple matching entries, is the first matching entry used?

core/ir/aarch64/codec.txt
```diff
@@ -294,1336 +294,1347 @@ x----------xxxxx----------------  wx16       # W/X register (or WZR/XZR)
...
+x1011010100xxxxxxxxx00xxxxxxxxxx  r   80        csinv            wx0 : wx5 wx16 cond
+x1011010100xxxxxxxxx01xxxxxxxxxx  r   81        csneg            wx0 : wx5 wx16 cond
+110101010000100001111110010xxxxx  n   569     dc_cisw                : x0
+110101010000101101111110001xxxxx  n   570    dc_civac                : memx0
+110101010000100001111010010xxxxx  n   571      dc_csw                : x0
+110101010000101101111010001xxxxx  n   572     dc_cvac                : memx0
+110101010000101101111011001xxxxx  n   573     dc_cvau                : memx0
+110101010000100001110110010xxxxx  n   574      dc_isw                : x0
+110101010000100001110110001xxxxx  n   575     dc_ivac                : memx0
+110101010000101101110100001xxxxx  n   568      dc_zva          memx0 :
```
abhinav92003, line 427: Adding to Derek's comment in instr_create_api.h, I guess we need to use a new type of operand here instead of `memx0`. One that also back-aligns the address.

core/ir/aarch64/instr_create_api.h
abhinav92003: > and its size is not OPSZ_sys but rather a cache line

So do we add new `OPSZ_n` constants, for all known cache line size values as `n`?

> how to get the cache line size for things like host != target

From Derek's comment on #4400: "I guess we would make this a decoder setting like the ISA mode (dr_get_isa_mode()) and add routines for setting it": Can we set the cache line size for the decoder by reading it from the `TRACE_MARKER_TYPE_CACHE_LINE_SIZE` marker whenever we see it? Or is the  marker seen too late?

abhinav92003 (approved):

core/ir/aarch64/codec.txt
derekbruening: #4393 was filed about scrapping this codec input format and switching to xml input in order to provide enough flexibility for splitting SYS into sub-opcodes.  But here it seems we're able to do that split in this codec format.  Was it a prior augmentation that gave us that capability?

core/ir/aarch64/instr_create_api.h
derekbruening: For drcachesim offline use, yes we could use the marker in the trace.  But we have other uses of the decoder, and we provide drdecodelib for others to use.

From an offline discussion:
+ We could add a new OPSZ_CACHE_LINE_SIZE to avoid having to ensure we have OPSZ_ values for all possible cache line sizes.
+ opnd_size_in_bytes() would then translate OPSZ_CACHE_LINE_SIZE to a value.  For live DR it would use the already-queried line size; for drdecode we would add a new global attribute storing a line size.
+ The new global attribute would have a default of the host size and we'd add init code to get that?  Hopefully that won't be painful to share the existing code for that with drdecode.
+ Users of drdecode doing cross-arch would have to call a set routine on the global attribute to get the right size.  Maybe the default could be the most likely AArch64 value.

core/ir/aarch64/instr_create_api.h
AssadHashmi: Yes.

clients/drcachesim/tracer/instru.cpp
AssadHashmi: > The description says #4382 but I don't see any disasm changes here (if there were: I would vote for a separate PR) -- is that a typo?

The reason I added #4382 is because there was discussion of splitting out SYS:
https://github.com/DynamoRIO/dynamorio/issues/4382#issuecomment-664596105
https://github.com/DynamoRIO/dynamorio/issues/4382#issuecomment-668720692

core/ir/aarch64/codec.txt
AssadHashmi: > So this was possible w/o changing to xml as #4393 talked about?

Yes, the comments in #4393 were optimistic about using XML for v8.0 and the comment about SYS/DC/IC was more of an example of what could be done rather than what we would do in the current codec. The plan is to use XML for Neoverse V1 and N2.

> Not even a change to the python script was needed?

No change to the script, but `codec.txt` change has to be run through `codecsort.py` which re-orders the new instructions and in so doing can re-align fields which results in space changes. Not ideal but I use the `Hide whitespace` option in the `Diff view` confing on GitHub for `codec.txt`.

> The ordering related stuff was already handled by 2c4f1e5, I think. IIUC this PR needs to only add new entries in the codec, which is what it does.

Yes.

As an aside, we need to review the approach proposed by #4393 and decide how to make the most of automating future instruction support. That is a discussion for #4393 we should have in January 2022, or offline as it is a siginificant decision with relatively long-term consequences.

clients/drcachesim/tracer/instru.cpp
AssadHashmi: No, see later comment https://github.com/DynamoRIO/dynamorio/pull/5235#discussion_r761123121

core/ir/aarch64/codec.txt
AssadHashmi: >So we continue to have this entry. Is that for all variants of sys other than dc or ic (which are handled by a different entry above)?

Yes, all variants of SYS other than DC or IC.

> I may have missed previous decoder changes that allow such overlapping entries in the codec (the bits of sys and dc match except the xs). What's the logic now? When there are multiple matching entries, is the first matching entry used?

The codec can now resolve conflicts to the precise instruction from the encoding regardless of its postion in `codec.txt`.

The reason the overlaps were problematic was because resolution was based on the fact that operand overlaps using `x` allowed for ANY combination of bits. But similar encodings are designed such that operands can only have a limited number of bit patterns. The `+` operand mask was introduced in #5092 to resolve such conflicts.

core/ir/aarch64/instr_create_api.h
AssadHashmi: > But this is not a register: it is a memory operand...

This is my understanding:
- The instruction takes a 64 bit register `<Xt>` which should contain a VA: `DC <dc_op>, <Xt>`
- The codec defines `<Xt>` as `memx0`:
`110101010000101101110100001xxxxx  n   568      dc_zva          memx0 :`
- Which creates a memory reference to the VA in `<Xt>` when decoded:
`opnd_create_base_disp(decode_reg(extract_uint(enc, 0, 5), true, false), DR_REG_NULL, 0, 0, OPSZ_sys);`

Disregarding `OPSZ_sys` for now, is that wrong?
What should it be, an absolute address using `opnd_create_abs_addr()` ?

>...and its size is not `OPSZ_sys` but rather a cache line (the N above).

This implies the DC ZVA macro should have an extra size input operand:
`#define INSTR_CREATE_dc_zva(dc, Rn, cline_sz) instr_create_1dst_1src(dc, OP_dc_zva, Rn, cline_sz)`
Which is converted to the `opnd_size_t` type.

But that introduces an input IR operand which is not part of the encoding definition. It should really be a fixed global size like the proposed `OPSZ_CACHE_LINE_SIZE`, e.g.
`opnd_create_base_disp(decode_reg(extract_uint(enc, 0, 5), true, false), DR_REG_NULL, 0, 0, opnd_size_in_bytes(OPSZ_CACHE_LINE_SIZE));`

> The docs here imply the macro will auto-convert it to a memory reference of the right size: but the test below passes in a memory reference (of the wrong size).

Just realised tests in `ir_aarch64.c` are wrong because they do an unnecessary `opnd_create_base_disp()` conversion on the register before calling `INSTR_CREATE_dc_zva()`!

core/ir/aarch64/codec.txt
AssadHashmi: See later comment https://github.com/DynamoRIO/dynamorio/pull/5235#discussion_r761123121

core/ir/aarch64/codec.txt
AssadHashmi: Probably. I need to understand what I'm doing wrong as noted here (https://github.com/DynamoRIO/dynamorio/pull/5235#discussion_r761130002) before changing `memx0`.

AssadHashmi:
@derekbruening @abhinav92003 re-requesting review though no code changes.
I've posted responses and questions to the issues raised.

clients/drcachesim/tracer/instru.cpp
derekbruening: OK.

core/ir/aarch64/instr_create_api.h
derekbruening: OK, it seems reasonable to keep the creation macro looking like assembly and thus taking in a bare register.  But the codec is not involved in direct synthesis of an instruction at the IR level: we need the implementation of the creation macro to turn that register into a memory operand with the cache line size (our vote is OPSZ_CACHE_LINE_SIZE).

The memory operand being a base-disp makes sense.

I think the one part we haven't thought through is the back-alignment: my comments in #4400 about back-aligning at decode time only works for an address that's a literal immediate: with a dynamic value in a register maybe the only way it could work is if it is a completely new type of operand?  Alternatively, we could leave it as a base-disp, and rely on everyone using `drutil_insert_get_mem_addr()` and having it notice the OPSZ_CACHE_LINE_SIZE and we document that that size implies back-alignment and the drutil code inserts a back-alignment (can be done w/o touching the flags on aarch64).

clients/drcachesim/tracer/instru.cpp
derekbruening: OK.

core/ir/aarch64/codec.txt
derekbruening: Ack.

core/ir/aarch64/instr_create_api.h
derekbruening: We have the dr_shift_type_t and dr_extend_type_t modifiers on index operands today: we could have something like that as a modifier on the base register to back-align it, w/o requiring a whole new opnd type: just a modifier on a base-disp.

derekbruening: dr_opnd_flags_t is another candidate

derekbruening: Given that dr_extend_type_t is hardcoded for index registers (`opnd_get_index_extend()`) we should have something new. 

 It could be a single bit like the existing `opnd_t.value.addr.scaled` or a new `dr_base_adjust_type_t`.  There are 3 bits free for AArch64 in `opnd_t.value.addr`.  I would vote for a single bit like `opnd_t.value.addr.base_aliigned`.  It would need new setter and getter functions.

derekbruening: So my vote is for:
- [ ] A new flag `opnd_t.value.addr.base_aligned` with new setter and getter functions
- [ ] A new size `OPSZ_CACHE_LINE_SIZE` with new global drdecode state for use outside of managed mode and resolution in `opnd_size_in_bytes()`
- [ ] A new codec type that sets the flag and sets `OPSZ_CACHE_LINE_SIZE`
- [ ] The creation macros turn the passed-in bare register into a base-disp that sets the flag and sets `OPSZ_CACHE_LINE_SIZE`
- [ ] A bump in the offline trace version with conditional handling of the existing workaround
- [ ] Adding support for the new flag to `drutil_insert_get_mem_addr()`

core/ir/aarch64/codec.txt
abhinav92003: Ack.

core/ir/aarch64/instr_create_api.h
AssadHashmi: > OK, it seems reasonable to keep the creation macro looking like assembly and thus taking in a bare register. But the codec is not involved in direct synthesis of an instruction at the IR level: we need the implementation of the creation macro to turn that register into a memory operand with the cache line size (our vote is OPSZ_CACHE_LINE_SIZE).

I've done this change but have a place holder for the memory operand size temporarily set to OPSZ_512.

There is 1 issue which needs a decision so I can implement and 1 I need to understand before implementing:

1) OPSZ_CACHE_LINE_SIZE
The Arm spec for DC ZVA says:
`Zero data cache by address. Zeroes a naturally aligned block of N bytes, where the size of N is identified in DCZID_EL0.`

`DCZID_EL0` is a system register the 4 least significant bits hold the block size:
`Log2 of the block size in words. The maximum size supported is 2KB (value == 9).`

The spec refers to block size rather than cache line size. What shall we call it?
`OPSZ_CACHE_LINE_SIZE`
`OPSZ_DC_ZVA_BLOCK_SIZE`
`Something else`

Where is the best place to initialise `OPSZ_!_SIZE`?

What should it be initialised to?
A fixed value? 2K bytes is the maximum size.
The value read from `DCZID_EL0` using `MRS <Xt>, DCZID_EL0` on startup?

2) What is the purpose of back-aligining the address? Is it a case of aligning to some previous memory boundary determined by cache line?
The Arm spec says:
`There is no alignment restriction on the address within the block of N bytes that is used.`
I guess the key word here is "_within_". We will backalign the _whole_ block to some boundary?

I noticed the `ALIGN_BACKWARD` macro called in many places. Does the proposed solution using `opnd_t.value.addr.base_aligned` flag boil down to using that macro?

derekbruening: > 1. OPSZ_CACHE_LINE_SIZE

Given that other instructions like `DC GVA` and `DC GZVA` (looks like we need to add handling to treat this like `ZVA`) also use `DCZID_EL0` I would not name this with just `ZVA`.

I could see two alternatives:

A. Try to be cross-platform (maybe RISC-V has something that would need this, or some other ISA or an extension to an existing ISA we support) and use the general name `OPSZ_CACHE_LINE_SIZE`.  At init time, read `DCZID_EL0` and if it's not equal to the dcache line size then we fatally abort (would that ever happen?)

B. Have a more ISA-specific name `OPSZ_CACHE_ZERO` or sthg and set it to `DCZID_EL0`.

> 2. What is the purpose of back-aligining the address?

Tools need to know the precise memory range affected by an instruction.  `DC ZVA x0` with `x0 == my-addr` is not very useful for tools: tracers, simulator, taint tracers, etc. all want to know what actual address range is being touched.  If the IR or its provided utilities like `drutil_insert_get_mem_addr()` don't convert `my-addr` into the actual `[back-aligned-my-addr, back-aligned-my-addr + block-size)`, then every single tool has to add special code to handle this one opcode, which is not good.  From the point of view of most tools, this is just a store instruction, and we want regular things like a memory store operand to able to be handled by general code without special casing for the opcode.

> I noticed the ALIGN_BACKWARD macro called in many places. Does the proposed solution using opnd_t.value.addr.base_aligned flag boil down to using that macro?

Since the actual value of the source register is dynamic, `drutil_insert_get_mem_addr()` cannot use that macro: it will have to generate and insert raw instruction(s) to do the alignment.

derekbruening: Given that `OPSZ` already implies "size" maybe `OPSZ_CACHE_LINE` is better if we go that name direction.

Re: "line" vs "block": they seem fairly interchangeable; I think the dr$sim sources also use "block" quite a bit.

AssadHashmi: I've had a go at implementing `OPSZ_CACHE_LINE`.

The value for `OPSZ_CACHE_LINE` in `dis-a64.txt` and `ir_aarch64.expect` is currently hard-coded. I don't think any changes are required in the disassembler, but we do need to modify the harness for codec testing. Is that something that should be done for v9.0 (#5145)?

Will start on the back-alignment mods now.

suite/tests/api/dis-a64.txt
```diff
@@ -13562,69 +13562,69 @@ d5087a5c : dc csw, x28                    : dc_csw %x28
...
-d50b7b36 : dc cvau, x22                   : dc_cvau (%x22)[1byte]
-d50b7b37 : dc cvau, x23                   : dc_cvau (%x23)[1byte]
-d50b7b38 : dc cvau, x24                   : dc_cvau (%x24)[1byte]
-d50b7b39 : dc cvau, x25                   : dc_cvau (%x25)[1byte]
-d50b7b3a : dc cvau, x26                   : dc_cvau (%x26)[1byte]
-d50b7b3b : dc cvau, x27                   : dc_cvau (%x27)[1byte]
-d50b7b3c : dc cvau, x28                   : dc_cvau (%x28)[1byte]
-d50b7b3d : dc cvau, x29                   : dc_cvau (%x29)[1byte]
-d50b7b3e : dc cvau, x30                   : dc_cvau (%x30)[1byte]
+d50b7a20 : dc cvac, x0                    : dc_cvac (%x0)[64byte]
```
abhinav92003, line 13565: Does this mean that we're assuming the cache line size to be 64 bytes in this test? Will it fail if run on a system with a different line size? Maybe add a comment.

core/arch/proc_shared.c
```diff
@@ -133,6 +133,11 @@ proc_init(void)
         global_heap_free(buf, PAGE_SIZE HEAPACCT(ACCT_OTHER));
         os_close(cpuinfo);
     }
+    CLIENT_ASSERT(dcache_zva_size > 0, "invalid dcache block size for zeroing");
```
abhinav92003, line 136: Will this fail in the DR_HOST_NOT_TARGET case? Do we want to maybe use a default to prevent this and other failures due to not getting the `dcache_zva_size`?

core/arch/proc.h
```diff
@@ -46,6 +46,7 @@
 
 /* exported for efficient access */
 extern size_t cache_line_size;
+extern size_t dcache_zva_size;
```
abhinav92003, line 49: Maybe note in a comment why we want to keep two separate variables for these, even though we expect them to be the same.

core/ir/aarch64/instr_create_api.h
```diff
@@ -739,7 +740,11 @@
  * \param Rn   The input register containing the virtual address to use.
  *             No alignment restrictions apply to this VA.
  */
-#define INSTR_CREATE_dc_civac(dc, Rn) instr_create_0dst_1src(dc, OP_dc_civac, Rn)
+#define INSTR_CREATE_dc_civac(dc, Rn)                                                   \
+    instr_create_0dst_1src(dc, OP_dc_civac,                                             \
+                           opnd_create_base_disp_aarch64(opnd_get_reg(Rn), DR_REG_NULL, \
+                                                         0, false, 0, 0,                \
+                                                         OPSZ_CACHE_LINE()))
```
abhinav92003, line 747: I see that there are `OPSZ_*` variables for 32, 64, 128 and 512. So we don't need to define any others that could be a possible cache line size on some systems.

suite/tests/api/dis-a64.txt
AssadHashmi: > Does this mean that we're assuming the cache line size to be 64 bytes in this test? Will it fail if run on a system with a different line size? 

Yes. I don't think any changes are required in the disassembler, but we do need to modify the harness for codec testing. I'll add a comment if the change doesn't need to be done for v9.0.

core/arch/proc_shared.c
AssadHashmi: > Will this fail in the DR_HOST_NOT_TARGET case?

Yes, it's failing:
```
aarch64-on-x86: 3 tests passed, **** 5 tests failed: ****
        tool.drcacheoff.raw2trace_unit_tests 
        code_api|api.ir 
        code_api|api.ir_negative 
        code_api|tool.drcacheoff.altbindir 
        code_api|api.ir-static
```
What needs to be done to run aarch64-on-x86? Is it just a case of setting `dcache_zva_size` to the host's `cache_line_size`?

abhinav92003: @derekbruening FYI

We discussed an API to set `dcache_zva_size` to a given value, that clients can use if they have that information. E.g. drcachesim can set it based on the value of `TRACE_MARKER_TYPE_CACHE_LINE_SIZE`.

At init time, it would be reasonable to set it to the host's cache line size.

AssadHashmi: > At init time, it would be reasonable to set it to the host's cache line size.

I've committed a fix for the failing `aarch64-on-x86` tests: `tool.drcacheoff.raw2trace_unit_tests` and `tool.drcacheoff.altbindir` now pass. The fix simply sets `dcache_zva_size = cache_line_size` if `DR_HOST_NOT_TARGET`.

`api.ir` fails because the hard-coded size in `ir_aarch64.expect` is `[64byte]` while the runtime value is `[32byte]`
But the runtime value is wrong for my x86 box:
```
$ cat /sys/devices/system/cpu/cpu0/cache/index0/coherency_line_size 
64
$ 

$ getconf LEVEL1_DCACHE_LINESIZE
64
asshas01@man-06:~$
```
So I would expect `get_processor_specific_info()` in `core/arch/x86/proc.c` to also set `cache_line_size` to 64.

My x86's `/proc/cpuinfo` says:
```
processor       : 55
vendor_id       : GenuineIntel
cpu family      : 6
model           : 79
model name      : Intel(R) Xeon(R) CPU E5-2690 v4 @ 2.60GHz
. . .
cache size      : 35840 KB
. . .
clflush size    : 64
cache_alignment : 64
```

derekbruening: > `api.ir` fails because the hard-coded size in `ir_aarch64.expect` is `[64byte]` while the runtime value is `[32byte]` But the runtime value is wrong for my x86 box:
> ...
> So I would expect `get_processor_specific_info()` in `core/arch/x86/proc.c` to also set `cache_line_size` to 64.

I would guess that for host != target regular DR is not used and this is just the drdecode library, which probably does not run all the regular managed-mode initialization code and so never runs the proc.c code.  Is there *any* initialization for a static library version of drdecode?  The API does not require any init call IIRC.

AssadHashmi: > probably does not run all the regular managed-mode initialization code and so never runs the proc.c code.

Yes, `get_cache_line_size()` and `get_dcache_zero_blk_size()` return false because `DR_HOST_NOT_TARGET`, and so `CTR_EL0` or `DCZID_EL0` can't be read.

Unlike x86, the smallest line size for AArch64 is 64. The latest commit sets `dcache_zva_size` and `cache_line_size` to 64 bytes for `DR_HOST_NOT_TARGET`.

The `api.ir-static` tests are are still failing. Not sure why the base disp doesn't appear at all in test output:
```
$ diff -y suite/tests/ir_aarch64.expect suite/tests/ir_aarch64.expect-out | grep \| | head -3
dc_zva  -> (%x0)[64byte]                                      | dc_zva  -> (%x0)
dc_zva  -> (%x1)[64byte]                                      | dc_zva  -> (%x1)
dc_zva  -> (%x2)[64byte]                                      | dc_zva  -> (%x2)
$
```

derekbruening: You mean it's not printing the size `[64byte]`?  Some kind of disassembly mask/option difference?

core/ir/opnd_shared.c
```diff
@@ -744,6 +744,14 @@ opnd_create_base_disp_aarch64(reg_id_t base_reg, reg_id_t index_reg,
         CLIENT_ASSERT(false, "opnd_create_base_disp_aarch64: invalid extend type");
     return opnd;
 }
+
+size_t dcache_zva_size;
```
derekbruening, line 748: style: Put globals at the top of the file (or after a banner comment).
But: shouldn't this be adjacent to cache_line_size which is in drlibc.c?

core/ir/opnd_api.h
```diff
@@ -1872,6 +1869,11 @@ opnd_t
 opnd_create_base_disp_aarch64(reg_id_t base_reg, reg_id_t index_reg,
                               dr_extend_type_t extend_type, bool scaled, int disp,
                               dr_opnd_flags_t flags, opnd_size_t size);
+
+DR_API
+/** Returns the data cache block size in bytes for zeroing the cache. */
+size_t
+proc_get_dcache_zva_size(void);
```
derekbruening, line 1876: A proc_ function should be in proc_api.h.

core/ir/opnd_shared.c
```diff
@@ -744,6 +744,14 @@ opnd_create_base_disp_aarch64(reg_id_t base_reg, reg_id_t index_reg,
...
+
+size_t
+proc_get_dcache_zva_size(void)
```
derekbruening, line 751: A proc_ function should be in proc.c.

suite/tests/api/ir_aarch64.c
```diff
@@ -5336,8 +5336,23 @@ test_asimddiff(void *dc)
...
+/* Macro expansion generates sequence of tests for creating cache instructions
+ * which are aliases of SYS instruction variants:
+ * DC ZVA, Xt        DC CVAC, Xt        DC CVAU, Xt        DC CIVAC Xt
+ * DC IVAC Xt        DC ISW Xt          DC CSW Xt          DC CISW Xt
+ * IC IVAU, Xt       IC IALLU           IC IALLUIS
+ */
+
+#define SYS_CACHE(cache, op)                                             \
+    do {                                                                 \
+        for (int x = DR_REG_X1; x < DR_REG_XSP; x++) {                   \
```
derekbruening, line 5348: Use DR_REG_START_GPR and DR_REG_STOP_GPR

suite/tests/api/ir_aarch64.c
```diff
@@ -5350,30 +5365,47 @@ test_sys(void *dc)
      */
 
     /* DC ZVA, Xt => SYS #3, C7, C4, #1, Xt */
-    instr = INSTR_CREATE_sys(
-        dc, opnd_create_immed_int(DR_DC_ZVA, OPSZ_2),
-        opnd_create_base_disp_aarch64(DR_REG_X0, DR_REG_NULL, 0, false, 0, 0, OPSZ_sys));
-
-    test_instr_encoding(dc, OP_sys, instr);
+    instr = INSTR_CREATE_dc_zva(dc, opnd_create_reg(DR_REG_X0));
```
derekbruening, line 5368: This is identical to one iteration of the SYS_CACHE loop?  So why have it separate?

suite/tests/api/ir_aarch64.c
```diff
@@ -5336,8 +5336,23 @@ test_asimddiff(void *dc)
...
 }
 
+/* Macro expansion generates sequence of tests for creating cache instructions
+ * which are aliases of SYS instruction variants:
+ * DC ZVA, Xt        DC CVAC, Xt        DC CVAU, Xt        DC CIVAC Xt
+ * DC IVAC Xt        DC ISW Xt          DC CSW Xt          DC CISW Xt
+ * IC IVAU, Xt       IC IALLU           IC IALLUIS
+ */
+
+#define SYS_CACHE(cache, op)                                             \
```
derekbruening, line 5346: A name that reflects the loop may help: TEST_SYS_CACHE_ALL_REGS or sthg

core/ir/aarch64/instr_create_api.h
```diff
@@ -108,20 +108,7 @@
...
-    DR_DC_CVAC = 0x1bd1,   /**< Clean dcache to point of coherency. */
-    DR_DC_CSW = 0x3d2,     /**< Clean dcache by set/way. */
-    DR_DC_CVAU = 0x1bd9,   /**< Clean dcache to point of unification. */
-    DR_DC_CIVAC = 0x1bf1,  /**< Clean and invalidate dcache to point of coherency. */
-    DR_DC_CISW = 0x3f2,    /**< Clean and invalidate dcache by set/way. */
-    DR_IC_IALLUIS = 0x388, /**< Invalidate icaches in ISD to point of unification. */
-    DR_IC_IALLU = 0x3a8,   /**< Invalidate icaches to point of unification. */
-    DR_IC_IVAU = 0x1ba9 /**< Invalidate icache by address to point of unification. */,
-};
+#define OPSZ_CACHE_LINE() opnd_size_from_bytes(proc_get_dcache_zva_size())
```
derekbruening, line 111: This is incorrect: operand sizes are enum values, not integer counts of bytes. `OPSZ_CACHE_LINE` needs to be an enum value, and opnd_size_in_bytes() is what resolves its actual size.  See my checklist where I said "resolution in opnd_size_in_bytes()".

derekbruening:
For the 9.0 release #5145: we're leaning toward just putting it out and not waiting for this.

Looking at the changes here if they came in after 9.0:
this PR removes DR_DC_ZVA, etc. which would become a compatibility
change; new flag opnd_t.value.addr.base_aligned should mostly be hidden by handling
in drutil_insert_get_mem_addr(); new size OPSZ_CACHE_LINE_SIZE shouldn't affect
compatibility; drcachesim version bump will be backward-compatible.  So it's only if
a tool besides drcachesim is already breaking SYS down and looking at the enums
manually.  And for DC ZVA, if the tool is using drutil_insert_get_mem_addr(), the new
DR should work fine.  So the likely breakage would be a tool doing the equivalent of
instru_t::is_aarch64_dcache_flush_op().  I think we're ok calling that kind of
breakage just 9.1 instead of 10 as a minor compatibility break.

abhinav92003:
> this PR removes DR_DC_ZVA, etc. which would become a compatibility
change; 

Is it possible to split the `SYS` opcode now, and do the remaining minor compatibility changes later in 9.1?

derekbruening:
> > this PR removes DR_DC_ZVA, etc. which would become a compatibility
> > change;
> 
> Is it possible to split the `SYS` opcode now, and do the remaining minor compatibility changes later in 9.1?

+1

Then post-9.0 we only change the DC ZVA operan which may not break anything since tools using drutil_insert_get_mem_addr() would auto-magically work even if their current workarounds were no longer triggered.

@AssadHashmi is that doable

AssadHashmi:
> For the 9.0 release #5145: we're leaning toward just putting it out and not waiting for this.

Apologies for being slow with this change @derekbruening and @abhinav92003  :-/

> > this PR removes DR_DC_ZVA, etc. which would become a compatibility
> > change;
> 
> 
> Is it possible to split the `SYS` opcode now, and do the remaining minor compatibility changes later in 9.1?

Sure. Just to be clear, I'm assuming you mean just the new `dc_` and `ic_` opcodes without the `OPSZ_CACHE_LINE` and its related changes?

derekbruening:
> Just to be clear, I'm assuming you mean just the new dc_ and ic_ opcodes without the OPSZ_CACHE_LINE and its related changes?

Right, leaving the operands exactly how they were with SYS so the current DC ZVA workarounds still work.  (With a TODO i#4400 comment about improving the operands later.)

clients/drcachesim/tracer/instru.cpp
AssadHashmi: To be resolved in post-v9.0 patch.

core/ir/aarch64/codec.txt
AssadHashmi: To be addressed and resolved in a post-v9.0 patch.

core/ir/aarch64/instr_create_api.h
```diff
@@ -738,6 +725,124 @@ enum {
...
+ * Creates a DC CIVAC instruction to Clean and Invalidate data cache by
+ * Virtual Address to point of Coherency.
+ * \param dc   The void * dcontext used to allocate memory for the #instr_t.
+ * \param Rn   The input register containing the virtual address to use.
+ *             No alignment restrictions apply to this VA.
+ */
+#define INSTR_CREATE_dc_civac(dc, Rn)                                                   \
+    instr_create_0dst_1src(dc, OP_dc_civac,                                             \
+                           opnd_create_base_disp_aarch64(opnd_get_reg(Rn), DR_REG_NULL, \
+                                                         0, false, 0, 0,                \
```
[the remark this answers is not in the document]
AssadHashmi: To be addressed and resolved in a post-v9.0 patch.

core/ir/aarch64/instr_create_api.h
AssadHashmi: To be addressed and resolved in a post-v9.0 patch.

suite/tests/api/ir_aarch64.c
AssadHashmi: Done.

suite/tests/api/ir_aarch64.c
AssadHashmi: These are longhand/explicit `INSTR_CREATE_` example calls which are handy for future maintainers and users, rather than being wrapped in a test macro.

suite/tests/api/ir_aarch64.c
derekbruening: These START and STOP are meant for this type of iteration: they would work well here, right?

suite/tests/api/ir_aarch64.c
derekbruening: OK.

derekbruening (approved):

derekbruening (approved):

abhinav92003 (approved):

clients/drcachesim/tracer/instru.cpp
```diff
@@ -185,36 +185,30 @@ instru_t::instr_to_prefetch_type(instr_t *instr)
...
 instru_t::is_aarch64_icache_flush_op(instr_t *instr)
 {
-    if (instr_get_opcode(instr) != OP_sys)
-        return false;
-    switch (opnd_get_immed_int(instr_get_src(instr, 0))) {
+    switch (instr_get_opcode(instr)) {
     // TODO i#4406: Handle privileged icache operations.
-    case DR_IC_IVAU: return true;
+    case OP_ic_ivau: return true;
```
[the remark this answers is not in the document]
AssadHashmi: > Was OP_sys added before the prior 8.0 release?

Yes.

> If so, please add a changelog entry in api/docs/release.dox about this opcode change.

Done.

suite/tests/api/ir_aarch64.c
AssadHashmi: Done.

AssadHashmi:
I'm pretty sure that the `tool.drcachesim.phys` regression is not caused by faulty encoding of `STUR`, see explanation at https://github.com/DynamoRIO/dynamorio/issues/4922#issuecomment-995023193.

@derekbruening should I merge this patch despite the above regressions?

derekbruening:
> I'm pretty sure that the `tool.drcachesim.phys` regression is not caused by faulty encoding of `STUR`, see explanation at [#4922 (comment)](https://github.com/DynamoRIO/dynamorio/issues/4922#issuecomment-995023193).
> 
> @derekbruening should I merge this patch despite the above regressions?

So these tests `tool.drcachesim.phys` and `tool.drcachesim.coherence` were already flaky and failing sometimes (always?) on Jenkins?  I see both on the list in runsuite_wrapper.pl: phys marked as #4922; coherence just as #2417.  I.e., is this not really a regression and the failures were already there?

derekbruening:
It looks like the test failure that is making Jenkins red is `code_api|security-common.TestAllocWE`: the phys and coherence tests failures were ignored.  TestAllocWE failed in the most recent run and the one right before your post about phys and coherence.

derekbruening:
I think TestAllocWE is testing detection of code modification: could this PR be failing to update core DR's handling of cache flush instructions?

derekbruening:
```
core/ir/aarch64/instr.c:instr_is_icache_op(instr_t *instr)
core/ir/aarch64/instr.c-{
core/ir/aarch64/instr.c-    int opc = instr_get_opcode(instr);
core/ir/aarch64/instr.c-#define SYS_ARG_IC_IVAU 0x1ba9
core/ir/aarch64/instr.c-    if (opc == OP_sys && opnd_get_immed_int(instr_get_src(instr, 0)) == SYS_ARG_IC_IVAU)
```

Please do further searches to make sure there are no other hidden uses that weren't using the enum for the SYS variants.

AssadHashmi:
> So these tests tool.drcachesim.phys and tool.drcachesim.coherence were already flaky and failing sometimes (always?) on Jenkins? I see both on the list in runsuite_wrapper.pl: phys marked as #4922; coherence just as #2417. I.e., is this not really a regression and the failures were already there?

Yes, my bad for not checking properly.

> I think TestAllocWE is testing detection of code modification: could this PR be failing to update core DR's handling of cache flush instructions?

Yes, I searched for instances of `OP_sys` which use a qualifying number for specific cache instructions and found only the instance you highlighted: `0x1ba9` in `instr_is_icache_op()`. Updated `instr.c` and `mangle.c` to use `OP_ic_ivau` instead. `code_api|security-common.TestAllocWE` now passing.

Issue #5247, opened 2021-12-13
SPIN exclusive load dest matching store result


derekbruening:
Xref #5245 which is another spin coming from -ldstex2cas.

For #1698 we have -ldstex2cas (see https://dynamorio.org/page_ldstex.html). It records what was loaded, and then before the exclusive store it re-loads and if the result does not match it loops.

Unfortunately it does not handle an exclusive load whose destination matches the exclusive store result when the sequence is mangled using the optimized same-block approach:

```
      1:
        ldaxp    w2, w1, [x0]
        add      w4, w1, #0x1
        add      w5, w2, #0x1
        stlxp    w1, w4, w5, [x0]
        cbnz     w1, 1b
```

That just loops forever as it is mangled into:

```
bb ilist after mangling:
TAG  0x0000ffff86e355b8
 +0    L4 @0x0000fffd46ede918  887f8402   ldaxp  (%x0)[8byte] -> %w2 %w1
 +4    m4 @0x0000fffd46edcac8  887f8402   <label>
 +4    L3 @0x0000fffd46edfce8  11000424   add    %w1 $0x0001 lsl $0x00 -> %w4
 +8    L3 @0x0000fffd46edda78  11000445   add    %w2 $0x0001 lsl $0x00 -> %w5
 +12   m4 @0x0000fffd46edf4f0  f9000f83   str    %x3 -> +0x18(%x28)[8byte]
 +16   m4 @0x0000fffd46edd3d0  887f8c01   ldaxp  (%x0)[8byte] -> %w1 %w3
 +20   m4 @0x0000fffd46edf5a0  cb226021   sub    %x1 %x2 uxtx $0x0000000000000000 -> %x1
 +24   m4 @0x0000fffd46edf090  b5000001   cbnz   @0x0000fffd46edb5b0[8byte] %x1
 +28   m4 @0x0000fffd46edea50  cb216061   sub    %x3 %x1 uxtx $0x0000000000000000 -> %x1
 +32   m4 @0x0000fffd46ede8b0  b5000001   cbnz   @0x0000fffd46edb5b0[8byte] %x1
 +36   L3 @0x0000fffd46edf7a0  88219404   stlxp  %w4 %w5 -> (%x0)[8byte] %w1
 +40   m4 @0x0000fffd46eddb48  14000000   b      @0x0000fffd46edd288[8byte]
 +44   m4 @0x0000fffd46edb5b0  14000000   <label>
[...]

PR #5248, opened 2021-12-13
i#5245: Handle exclusive loads into the zero register

### LINKED ISSUE #5245: SPIN exclusive load to zero register
For #1698 we have -ldstex2cas (see https://dynamorio.org/page_ldstex.html).  It records what was loaded, and then before the exclusive store it re-loads and if the result does not match it loops.

Unfortunately it does not handle a discarded result to the zero register:
```
      1:
        ldaxp    xzr, x1, [x0]
        add      x2, x1, #0x1
        stlxp    w3, xzr, x2, [x0]
        cbnz     w3, 1b
```

That just loops forever as it is mangled into:
```
 +0    L4 @0x0000fffd6f51f6a0  c87f841f   ldaxp  (%x0)[16byte] -> %xzr %x1
 +4    m4 @0x0000fffd6f51f310  f9000b82   str    %x2 -> +0x10(%x28)[8byte]
 +8    m4 @0x0000fffd6f51f3d8  a90aff80   stp    %x0 %xzr -> +0xa8(%x28)[16byte]
 +12   m4 @0x0000fffd6f51f470  d2800102   movz   $0x0000000000000008 lsl $0x0000000000000000 -> %x2
 +16   m4 @0x0000fffd6f51f5a0  a90b8b81   stp    %x1 %x2 -> +0xb8(%x28)[16byte]
 +20   m4 @0x0000fffd6f51f620  f9400b82   ldr    +0x10(%x28)[8byte] -> %x2

 +220  L3 @0x0000fffd6f51fc80  91000422   add    %x1 $0x0001 lsl $0x00 -> %x2
 +224  m4 @0x0000fffd6f51f720  f9000781   str    %x1 -> +0x08(%x28)[8byte]
 +228  m4 @0x0000fffd6f51f7a0  f9001384   str    %x4 -> +0x20(%x28)[8byte]
 +232  m4 @0x0000fffd6f51f820  f9001785   str    %x5 -> +0x28(%x28)[8byte]
 +236  m4 @0x0000fffd6f51f8e8  f9405781   ldr    +0xa8(%x28)[8byte] -> %x1
 +240  m4 @0x0000fffd6f51f968  cb216003   sub    %x0 %x1 uxtx $0x0000000000000000 -> %x3
 +244  m4 @0x0000fffd6f51f9e8  b5000003   cbnz   @0x0000fffd6f51ded8[8byte] %x3
 +248  m4 @0x0000fffd6f51fa68  f9406381   ldr    +0xc0(%x28)[8byte] -> %x1
 +252  m4 @0x0000fffd6f51fb30  d1002023   sub    %x1 $0x0000000000000008 lsl $0x0000000000000000 -> %x3
 +256  m4 @0x0000fffd6f51e0b8  b5000003   cbnz   @0x0000fffd6f51ded8[8byte] %x3
 +260  m4 @0x0000fffd6f51b7e8  f9405b81   ldr    +0xb0(%x28)[8byte] -> %x1   <===== loads the stored xzr
 +264  m4 @0x0000fffd6f51ae40  f9405f84   ldr    +0xb8(%x28)[8byte] -> %x4
 +268  m4 @0x0000fffd6f51b630  c87f9403   ldaxp  (%x0)[16byte] -> %x3 %x5  <===== loads an actual maybe non-zero value
 +272  m4 @0x0000fffd6f51add8  cb216063   sub    %x3 %x1 uxtx $0x0000000000000000 -> %x3  <==== likely to not be the same
 +276  m4 @0x0000fffd6f51e188  b5000003   cbnz   @0x0000fffd6f51ded8[8byte] %x3
 +280  m4 @0x0000fffd6f51e120  cb2460a3   sub    %x5 %x4 uxtx $0x0000000000000000 -> %x3
 +284  m4 @0x0000fffd6f51e3d8  b5000003   cbnz   @0x0000fffd6f51ded8[8byte] %x3
 +288  L3 @0x0000fffd6f51dc30  c823881f   stlxp  %xzr %x2 -> (%x0)[16byte] %w3
 +292  m4 @0x0000fffd6f51b930  14000000   b      @0x0000fffd6f51bfd0[8byte]
 +296  m4 @0x0000fffd6f51ded8  14000000   <label>
 +296  m4 @0x0000fffd6f51ac30  d503305f   clrex  $0x0000000000000000
 +300  m3 @0x0000fffd6f51c4a8  c823881f   stlxp  %xzr %x2 -> (%x0)[16byte] %w3
 +304  m4 @0x0000fffd6f51bfd0  d503305f   <label>
 +304  m4 @0x0000fffd6f51ce30  f9400781   ldr    +0x08(%x28)[8byte] -> %x1
 +308  m4 @0x0000fffd6f51d040  f9401384   ldr    +0x20(%x28)[8byte] -> %x4
 +312  m4 @0x0000fffd6f51dae0  f9401785   ldr    +0x28(%x28)[8byte] -> %x5
 +316  L3 @0x0000fffd6f51d208  35ffffa3   cbnz   $0x0000ffffaf4756d8 %w3
 +320  L4 @0x0000fffd6f51f290  14000000   b      $0x0000ffffaf4756e8
```

We need to add special-casing of `xzr` as an exclusive load target register.

derekbruening:
Adds handling for AArch64 exclusive loads that discard a value by
targeting the zero register.  Without this fix, we loop forever in the
store since we load the real value and compare it to the zero we
recorded.  Our fix is to replace the zero register with a scratch
register so that we record the actual value at load time.  For the
same-block optimized case, we instead simply skip the value comparison
check at the store.

Adds a test of every size and destination zero variant.  This test
spins (at multiple points) without the fix.

Fixes #5245

### DIFF
--- a/core/arch/aarchxx/mangle.c
+++ b/core/arch/aarchxx/mangle.c
@@ -3253,8 +3253,11 @@ mangle_exclusive_load(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
         if (instr_writes_to_reg(in, value_reg, DR_QUERY_INCLUDE_ALL) ||
             (value2_reg != DR_REG_NULL &&
              instr_writes_to_reg(in, value2_reg, DR_QUERY_INCLUDE_ALL)) ||
-            instr_writes_to_reg(in, base_reg, DR_QUERY_INCLUDE_ALL))
+            instr_writes_to_reg(in, base_reg, DR_QUERY_INCLUDE_ALL)) {
+            LOG(THREAD, LOG_INTERP, 4,
+                "Value clobbered => not using same-block ldex-stex mangling\n");
             break;
+        }
     }
     /* If the ldex uses the stolen reg, we do not swap around it as we normally do,
      * since we have a bunch of TLS refs inside that would then have a non-standard
@@ -3272,8 +3275,8 @@ mangle_exclusive_load(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
              */
             stex_in_same_block = false;
         }
-        swap_reg = pick_scratch_reg(dcontext, instr, false, DR_REG_NULL, DR_REG_NULL,
-                                    DR_REG_NULL, &swap_slot, &swap_restore);
+        swap_reg = pick_scratch_reg(dcontext, instr, DR_REG_NULL, DR_REG_NULL,
+                                    DR_REG_NULL, false, &swap_slot, &swap_restore);
         insert_save_to_tls_if_necessary(dcontext, ilist, instr, swap_reg, swap_slot);
         if (instr_reads_from_reg(instr, dr_reg_stolen, DR_QUERY_DEFAULT)) {
             PRE(ilist, instr,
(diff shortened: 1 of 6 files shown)


derekbruening:
Test manually on an ARM system.

suite/tests/client-interface/ldstex.c
```diff
@@ -843,6 +881,64 @@ GLOBAL_LABEL(FUNCNAME:)
         END_FUNC(FUNCNAME)
 #undef FUNCNAME
 
+#ifdef AARCH64
+/* int ldstex_inc32_with_xzr(two_counters_t *counter)
+ */
```
abhinav92003, line 886: These don't test the case where the exclusive load and store are in different blocks. Do we test that path using the clean-calls case below? (which thwarts single-block optimizations for ldstex2cas)

suite/tests/client-interface/ldstex.c
```diff
@@ -140,6 +151,18 @@ test_shapes(void)
     added = ldstex_inc_shapes(&ctr);
     if (ctr != 42 + added)
         print("Error in ldstex_inc_shapes: %d\n", ctr);
+
+#    ifdef AARCH64
+    my_var.counter2 = 117;
+    added = ldstex_inc32_with_xzr(&my_var);
+    /* We zero both and only the 2nd's add sticks. */
+    if (my_var.counter2 != added)
```
abhinav92003, line 159: Check that `counter1 == 0` too.

abhinav92003 (approved):

core/arch/aarchxx/mangle.c
```diff
@@ -3286,6 +3289,39 @@ mangle_exclusive_load(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
...
+    reg_id_t xzr_repl = DR_REG_NULL, xzr2_repl = DR_REG_NULL;
+#ifdef AARCH64
+    /* If the ldex loads into the zero register, we need to instead get the real
+     * value so our compare at the stex will succeed (otherwise we will loop
+     * forever: i#5245).  For same-block we statically skip the compare.
+     */
+    if (!stex_in_same_block && value_reg == DR_REG_XZR) {
+        xzr_repl = pick_scratch_reg(dcontext, instr, swap_reg, DR_REG_NULL, DR_REG_NULL,
+                                    true, &xzr_slot, &xzr_restore);
+        insert_save_to_tls_if_necessary(dcontext, ilist, instr, xzr_repl, xzr_slot);
```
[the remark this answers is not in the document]
derekbruening: Hmm, no it is not: yet it is saved in several places in these functions (but not restored) for that case.  Fixed.

core/arch/aarchxx/mangle.c
```diff
@@ -3391,6 +3427,10 @@ mangle_exclusive_load(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
                 instr_create_restore_from_tls(dcontext, swap_reg, swap_slot));
         }
     }
+    if (xzr_restore)
```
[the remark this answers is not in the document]
derekbruening: Not at L3325.  Possibly at 3370, but not for the A64 base==sp case: so it gets complicated.  I don't think it saves anything to do it early: the slot count is fixed (these are DR's core scratch slots) and we are not out of them here (we do run out for a case in #5247).  So it would add complexity for no gain: so I would prefer to leave it.

suite/tests/client-interface/ldstex.c
derekbruening: Done.

suite/tests/client-interface/ldstex.c
derekbruening: Yes -- added a comment.  The slowpath is the same regardless of whether actually in separate blocks.

PR #5249, opened 2021-12-13
i#5247: Handle exclusive loads sharing store result

### LINKED ISSUE #5247: SPIN exclusive load dest matching store result

derekbruening:
Adds handling for AArch64 exclusive loads whose destination equals the
result register of the paired exclusive store.  For optimized
same-block handling, without this fix our value comparison loops
forever as it clobbers the original value by using the dead store
result register.

The fix is to use a scratch register, but this fails for stolen
register cases: thus we bail to the slowpath for those.  We also have
to re-order the compare to ensure we look at the matching case first,
as the compare must write to the same result register.

Adds test cases to the ldstex test.
Tested manually on an ARM machine.

Fixes #5247


core/arch/aarchxx/mangle.c
```diff
@@ -3043,7 +3043,13 @@ create_ldax_from_stex(dcontext_t *dcontext, instr_t *strex, reg_id_t *dest_reg O
...
      */
     opnd_size_t opsz = opnd_get_size(instr_get_src(strex, 0));
-    *dest_reg = reg_resize_to_opsz(opnd_get_reg(instr_get_dst(strex, 1)), opsz);
+    /* The store dest reg could equal a load dest reg, in which case the caller must
+     * pass us a scratch reg.
+     */
+    if (*dest_reg == DR_REG_NULL)
+        *dest_reg = reg_resize_to_opsz(opnd_get_reg(instr_get_dst(strex, 1)), opsz);
+    else
+        *dest_reg = reg_resize_to_opsz(*dest_reg, opsz);
```
abhinav92003, line 3052: Instead of using the dest_reg if provided, shouldn't we check for the load-sharing-store-result case?

core/arch/aarchxx/mangle.c
```diff
@@ -3238,6 +3244,14 @@ mangle_exclusive_load(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
                 opnd_get_size(instr_get_src(in, 0)) ==
                     opnd_get_size(instr_get_dst(instr, 0)) &&
                 opnd_get_base(instr_get_dst(in, 0)) == base_reg &&
+                /* pick_scratch_reg() only takes 3 conflicts, so we push a pair with
```
abhinav92003, line 3247: I'm thinking if we should use the un-optimized sequence for all load-shares-store-result cases. May be better to avoid adding more complexity.

core/arch/aarchxx/mangle.c
```diff
@@ -3529,16 +3543,36 @@ mangle_exclusive_store(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
...
+    bool compare_second_first = false;
     if (ldex_in_same_block) {
         /* We aren't saving the flags so we can only handle Thumb mode with CBNZ. */
         IF_ARM(ASSERT(is_cbnz_available(dcontext, reg_res)));
+        if (reg_to_pointer_sized(reg_res) == reg_orig_ld_val ||
+            reg_to_pointer_sized(reg_res) == reg_orig_ld_val2) {
+            /* We can't use the store res in the synthetic load if it has a value. */
+            scratch = pick_scratch_reg(dcontext, instr, swap_reg, reg_orig_ld_val,
+                                       reg_orig_ld_val2,
+                                       /*dead_reg_ok=*/false, &slot, &should_restore);
```
abhinav92003, line 3557: Why isn't a dead reg okay here?

core/arch/aarchxx/mangle.c
```diff
@@ -3529,16 +3543,36 @@ mangle_exclusive_store(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
...
+            insert_save_to_tls_if_necessary(dcontext, ilist, instr, scratch, slot);
+            reg_new_ld_val = scratch;
+            if (reg_to_pointer_sized(reg_res) == reg_orig_ld_val2)
```
abhinav92003, line 3560: Could both the loaded values be the reg_res? Or will that give a SIGILL? ( ldp r1,r1,[x] gives a SIGILL IIRC)

core/arch/aarchxx/mangle.c
```diff
@@ -3529,16 +3543,36 @@ mangle_exclusive_store(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
...
+                compare_second_first = true;
+        }
         if (is_pair) {
```
abhinav92003, line 3563: Trying to understand which case this is. Here, we have `ldex_in_same_block` and `is_pair`. There's no load-sharing-store-dest overlap though right? But I see we're still picking a scratch reg below.

suite/tests/client-interface/ldstex.c
```diff
@@ -492,7 +513,25 @@ GLOBAL_LABEL(FUNCNAME:)
...
         bne      2b
-        mov      r0, #2
+        /* Test store-res == load-dest (i#5247). */
+        push     {r4, r5, r10, r11}
+      3:
+        ldaexd   r2, r3, [r0]
+        add      r4, r2, #0x1
+        add      r5, r3, #0x1
+        stlexd   r2, r4, r5, [r0]
+        cmp      r1, #0
```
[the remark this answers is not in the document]
derekbruening: Yes, this went through some iterations b/c ARM requires a pair to start with an even-numbered register: looks like the final version was messed up.  Fixed.

suite/tests/client-interface/ldstex.c
```diff
@@ -492,7 +513,25 @@ GLOBAL_LABEL(FUNCNAME:)
...
+        bne      3b
+      4:
+        ldaexd   r10, r11, [r0] /* Test stolen reg. */
+        add      r4, r10, #0x1
+        add      r5, r11, #0x1
+        stlexd   r10, r4, r5, [r0]
+        cmp      r1, #0
```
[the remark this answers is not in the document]
derekbruening: Fixed.

core/arch/aarchxx/mangle.c
```diff
@@ -3529,16 +3543,36 @@ mangle_exclusive_store(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
...
     if (ldex_in_same_block) {
         /* We aren't saving the flags so we can only handle Thumb mode with CBNZ. */
         IF_ARM(ASSERT(is_cbnz_available(dcontext, reg_res)));
+        if (reg_to_pointer_sized(reg_res) == reg_orig_ld_val ||
+            reg_to_pointer_sized(reg_res) == reg_orig_ld_val2) {
+            /* We can't use the store res in the synthetic load if it has a value. */
+            scratch = pick_scratch_reg(dcontext, instr, swap_reg, reg_orig_ld_val,
+                                       reg_orig_ld_val2,
+                                       /*dead_reg_ok=*/false, &slot, &should_restore);
+            insert_save_to_tls_if_necessary(dcontext, ilist, instr, scratch, slot);
```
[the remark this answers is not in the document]
derekbruening: Done.

core/arch/aarchxx/mangle.c
derekbruening: No, see the comment on the line above: the caller has to pass us a scratch reg for that case.

core/arch/aarchxx/mangle.c
derekbruening: I considered this, but rejected it since this does not seem to be a rare occurrence: every single time there's a store-release pair, the load pair has one as XZR and the other as the store's destination.  I think the compiler really wants to use XZR for both, but making both load dests identical is undefined behavior, so it picks a dead register.  So it seems that this is not uncommon, and it's always in a single block, so it seems worth avoiding all the extra memory operations of the slowpath.

derekbruening: Added a comment.

core/arch/aarchxx/mangle.c
derekbruening: This is the same as all the existing scratch reg picking: because the compares have to write to the store res (so the mismatch case has the right result), and that would clobber a subsequent compare's sources.  Added a comment up by the general scratch reg discussion.

core/arch/aarchxx/mangle.c
derekbruening: Identical pair destinations is undefined behavior (and SIGILL on the Jenkins machine).

core/arch/aarchxx/mangle.c
derekbruening: Two registers are needed for the value comparison: the store result is used for one, so a scratch is only needed for a pair.  (Added comment about this case of a dead reg working out b/c it is always compared immediately and so does not affect a subsequent compare source.)

core/arch/aarchxx/mangle.c
abhinav92003: I mean: if it is the load-sharing-store-result case, assert that the caller provided a scratch reg in `dest_reg`; and if it is not that case, don't use the scratch even if it was provided by the user. What would happen currently in the load-sharing-store-result case if the caller didn't provide the scratch reg?

core/arch/aarchxx/mangle.c
```diff
@@ -3525,20 +3545,47 @@ mangle_exclusive_store(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
...
+            if (should_restore)
+                insert_save_to_tls_if_necessary(dcontext, ilist, instr, scratch, slot);
+            reg_new_ld_val = scratch;
+            if (reg_to_pointer_sized(reg_res) == reg_orig_ld_val2)
+                compare_second_first = true;
+        }
         if (is_pair) {
             /* We do need one scratch reg for the value comparison. */
-            scratch3 = pick_scratch_reg(dcontext, instr, swap_reg, reg_orig_ld_val,
+            /* pick_scratch_reg() only takes 3 conflicts, so we push a pair with
```
abhinav92003, line 3578: So, swap_reg and scratch both cannot be DR_REG_NULL. Because the stolen reg case and load-shares-store-dest case is pushed to the unoptimised path.

Maybe make this comment more verbose. Remind the reader that scratch!=DR_REG_NULL for the load-shares-store-dest  case and swap_reg!=DR_REG_NULL for the stolen reg case. Maybe also rename those variables (swap_reg -> stolen_reg_swap_reg, or something).

core/arch/aarchxx/mangle.c
abhinav92003: Ah right, the only change here is the do_not_pick reg. No new scratch.

abhinav92003 (approved):

core/arch/aarchxx/mangle.c
abhinav92003: Thanks.

core/arch/aarchxx/mangle.c
derekbruening: This function does not know what the load looks like so it doesn't know whether it's the load-sharing-store-result case: it is passed only the store, and the caller does not know what the load is for slowpath cases.

> What would happen currently in the load-sharing-store-result case if the caller didn't provide the scratch reg?

Probably an infinite loop.

core/arch/aarchxx/mangle.c
derekbruening: Extended comment.

In my mind 'swap_reg' implied the stolen reg as that is the case where we swap in general when mangling (not just for ldstex), but I could see that not being the case for other reader so I did s/swap_reg/stolen_swap_reg/.

core/arch/aarchxx/mangle.c
derekbruening: So summarizing: an assert could only go in the caller for the optimized case; but that's where the caller is already picking the scratch reg so it wouldn't help.  Unfortunately I don't think more can be done.  This code is getting complex but simplifying by dropping some of these cases has its own downsides...I think we have to live with it.  A virtual scratch register layer could help but has its own problems.

Issue #5254, opened 2021-12-15
rt_sigprocmask not failing for bad address


abhinav92003:
DR's handling of `rt_sigprocmask` doesn't fail even if a bad address is passed. `handle_sigprocmask` indeed checks whether the passed address is valid: https://github.com/DynamoRIO/dynamorio/blob/223213a2f254640cd1433a48cf0bebd20f828274/core/unix/signal.c#L2306, but it returns success anyway: https://github.com/DynamoRIO/dynamorio/blob/223213a2f254640cd1433a48cf0bebd20f828274/core/unix/os.c#L7349. This is causing an assertion to fail in a proprietary app.

abhinav92003:
Example code that uses this syscall to check whether a given address is valid:  https://github.com/abseil/abseil-cpp/blob/master/absl/debugging/internal/address_is_readable.cc#L85

abhinav92003:
Keeping this open as there's still some work left on handling the TODOs added by #5255.

PR #5255, opened 2021-12-16
i#5254: handle rt_sigprocmask error cases.

abhinav92003:
Handles the EINVAL and EFAULT cases in rt_sigprocmask. Some code
uses this syscall to check whether a given address is valid;
e.g. https://github.com/abseil/abseil-cpp/blob/master/absl/debugging/internal/address_is_readable.cc#L85
This requires DR to return EFAULT as expected by the app.

Also adds tests for some error and success cases of rt_sigprocmask.

Fixes: #5254

### DIFF
--- a/core/unix/os.c
+++ b/core/unix/os.c
@@ -7341,10 +7341,11 @@ pre_system_call(dcontext_t *dcontext)
         dcontext->sys_param0 = sys_param(dcontext, 0);
         dcontext->sys_param1 = sys_param(dcontext, 1);
         dcontext->sys_param2 = sys_param(dcontext, 2);
-        /* SYS_sigprocmask on MacOS does not have a size arg. So we use the kernel_sigset_t
-         * size instead.
+        /* SYS_sigprocmask on MacOS does not have a size arg. So we use the
+         * kernel_sigset_t size instead.
          */
-        dcontext->sys_param3 = IF_MACOS_ELSE(sizeof(kernel_sigset_t), sys_param(dcontext, 3));
+        dcontext->sys_param3 =
+            IF_MACOS_ELSE(sizeof(kernel_sigset_t), sys_param(dcontext, 3));
         execute_syscall = handle_sigprocmask(dcontext, (int)sys_param(dcontext, 0),
                                              (kernel_sigset_t *)sys_param(dcontext, 1),
                                              (kernel_sigset_t *)sys_param(dcontext, 2),
--- a/core/unix/os.c
+++ b/core/unix/os.c
@@ -7344,12 +7344,13 @@ pre_system_call(dcontext_t *dcontext)
         /* SYS_sigprocmask on MacOS does not have a size arg. So we use the
          * kernel_sigset_t size instead.
          */
-        dcontext->sys_param3 =
-            IF_MACOS_ELSE(sizeof(kernel_sigset_t), sys_param(dcontext, 3));
+        size_t sigsetsize =
+            (size_t)IF_MACOS_ELSE(sizeof(kernel_sigset_t), sys_param(dcontext, 3));
+        dcontext->sys_param3 = (reg_t)sigsetsize;
         execute_syscall = handle_sigprocmask(dcontext, (int)sys_param(dcontext, 0),
                                              (kernel_sigset_t *)sys_param(dcontext, 1),
                                              (kernel_sigset_t *)sys_param(dcontext, 2),
-                                             (size_t)sys_param(dcontext, 3), &error_code);
+                                             sigsetsize, &error_code);
         if (!execute_syscall) {
             if (error_code == 0)
                 set_success_return_val(dcontext, 0);
--- a/core/unix/os.c
+++ b/core/unix/os.c
@@ -7337,6 +7337,7 @@ pre_system_call(dcontext_t *dcontext)
              size_t sigsetsize)
          */
         /* we also need access to the params in post_system_call */
+        uint errno_val = 0;
         dcontext->sys_param0 = sys_param(dcontext, 0);
         dcontext->sys_param1 = sys_param(dcontext, 1);
         dcontext->sys_param2 = sys_param(dcontext, 2);
@@ -7344,9 +7345,13 @@ pre_system_call(dcontext_t *dcontext)
         execute_syscall = handle_sigprocmask(dcontext, (int)sys_param(dcontext, 0),
                                              (kernel_sigset_t *)sys_param(dcontext, 1),
                                              (kernel_sigset_t *)sys_param(dcontext, 2),
-                                             (size_t)sys_param(dcontext, 3));
-        if (!execute_syscall)
-            set_success_return_val(dcontext, 0);
+                                             (size_t)sys_param(dcontext, 3), &errno_val);
+        if (!execute_syscall) {
+            if (errno_val == 0)
+                set_success_return_val(dcontext, 0);
+            else
+                set_failure_return_val(dcontext, errno_val);
+        }
         break;
     }
 #ifdef MACOS
--- a/core/unix/os.c
+++ b/core/unix/os.c
@@ -7331,6 +7331,7 @@ pre_system_call(dcontext_t *dcontext)
         break;
     }
     case IF_MACOS_ELSE(SYS_sigprocmask, SYS_rt_sigprocmask): { /* 175 */
+        /* TODO i#5256: Fx this path and enable linux.sigaction on MacOS. */
         /* in /usr/src/linux/kernel/signal.c:
            asmlinkage long
            sys_rt_sigprocmask(int how, sigset_t *set, sigset_t *oset,
--- a/core/unix/os.c
+++ b/core/unix/os.c
@@ -7341,7 +7341,10 @@ pre_system_call(dcontext_t *dcontext)
         dcontext->sys_param0 = sys_param(dcontext, 0);
         dcontext->sys_param1 = sys_param(dcontext, 1);
         dcontext->sys_param2 = sys_param(dcontext, 2);
-        dcontext->sys_param3 = sys_param(dcontext, 3);
+        /* SYS_sigprocmask on MacOS does not have a size arg. So we use the sigset_t
+         * size instead.
+         */
+        dcontext->sys_param3 = IF_MACOS_ELSE(sizeof(sigset_t), sys_param(dcontext, 3));
         execute_syscall = handle_sigprocmask(dcontext, (int)sys_param(dcontext, 0),
                                              (kernel_sigset_t *)sys_param(dcontext, 1),
                                              (kernel_sigset_t *)sys_param(dcontext, 2),
--- a/core/unix/os.c
+++ b/core/unix/os.c
@@ -7337,7 +7337,7 @@ pre_system_call(dcontext_t *dcontext)
              size_t sigsetsize)
          */
         /* we also need access to the params in post_system_call */
-        uint errno_val = 0;
+        uint error_code = 0;
         dcontext->sys_param0 = sys_param(dcontext, 0);
         dcontext->sys_param1 = sys_param(dcontext, 1);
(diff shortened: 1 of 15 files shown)


suite/tests/linux/sigaction.c
```diff
@@ -115,6 +115,43 @@ set_sigaction_handler(int sig, void *action)
...
+    assert(errno == EINVAL);
+
+    /* Success. */
+    assert(syscall(SYS_rt_sigprocmask, ~0, NULL, NULL, 8) == 0);
+    assert(syscall(SYS_rt_sigprocmask, SIG_SETMASK, &new, NULL, 8) == 0);
+    assert(syscall(SYS_rt_sigprocmask, ~0, NULL, &old, 8) == 0);
+    assert(new == old);
+
+    /* Restore original sigprocmask. */
+    assert(syscall(SYS_rt_sigprocmask, SIG_SETMASK, &original, NULL, 8) == 0);
```
derekbruening, line 151: Please add parallel SYS_sigprocmask (for 32-bit Linux and for 64-bit Mac) tests as well: can probably generalize and use the same function w/ the syscall number parameterized.

suite/tests/linux/sigaction.c
```diff
@@ -115,6 +115,43 @@ set_sigaction_handler(int sig, void *action)
...
+    /* EFAULT cases. */
+    assert(syscall(SYS_rt_sigprocmask, ~0, 0x123, NULL, 8) == -1);
+    assert(errno == EFAULT);
+    assert(syscall(SYS_rt_sigprocmask, ~0, NULL, 0x123, 8) == -1);
+    assert(errno == EFAULT);
+    assert(syscall(SYS_rt_sigprocmask, SIG_BLOCK, 0x123, NULL, 8) == -1);
+    assert(errno == EFAULT);
+    assert(syscall(SYS_rt_sigprocmask, SIG_BLOCK, NULL, 0x123, 8) == -1);
+    assert(errno == EFAULT);
+
```
derekbruening, line 137: Add a test where the old set is readable but not writable.

core/unix/signal.c
```diff
@@ -2293,17 +2293,29 @@ check_signals_pending(dcontext_t *dcontext, thread_sig_info_t *info)
...
+        (oset != NULL && !d_r_safe_read(oset, sizeof(safe_old_set), &safe_old_set))) {
+        if (errno_val != NULL)
+            *errno_val = EFAULT;
+        return false;
+    }
+    if (app_set != NULL && !(how >= SIG_BLOCK && how <= SIG_SETMASK)) {
+        if (errno_val != NULL)
+            *errno_val = EINVAL;
+        return false;
+    }
```
derekbruening, line 2311: What about an invalid sigsetsize

suite/tests/linux/sigaction.c
```diff
@@ -115,6 +115,43 @@ set_sigaction_handler(int sig, void *action)
...
+    /* EINVAL cases. */
+    assert(syscall(SYS_rt_sigprocmask, ~0, &new, NULL, 8) == -1);
+    assert(errno == EINVAL);
+    assert(syscall(SYS_rt_sigprocmask, SIG_SETMASK + 1, &new, NULL, 8) == -1);
+    assert(errno == EINVAL);
+
```
derekbruening, line 143: Add a test of an invalid size.

suite/tests/linux/sigaction.c
```diff
@@ -115,6 +115,43 @@ set_sigaction_handler(int sig, void *action)
...
 
+#if !defined(MACOS)
+static void
+test_rt_sigprocmask()
+{
+    uint64 new = 0xf00d, old, original;
+    /* Save original sigprocmask. Both return the current sigprocmask. */
+    assert(syscall(SYS_rt_sigprocmask, SIG_SETMASK, NULL, &original,
+                   /*sizeof(kernel_sigset_t)*/ 8) == 0);
+    assert(syscall(SYS_rt_sigprocmask, ~0, NULL, &original, 8) == 0);
```
derekbruening, line 126: A test with both new and old sets being NULL would be good too.

core/unix/os.c
```diff
@@ -7337,16 +7337,21 @@ pre_system_call(dcontext_t *dcontext)
              size_t sigsetsize)
          */
         /* we also need access to the params in post_system_call */
+        uint errno_val = 0;
```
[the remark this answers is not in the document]
abhinav92003: errno_value is used many other places as well. I changed the uses in this PR for now.

core/unix/os_private.h
```diff
@@ -350,7 +350,7 @@ handle_sigaltstack(dcontext_t *dcontext, const stack_t *stack, stack_t *old_stac
 
 bool
 handle_sigprocmask(dcontext_t *dcontext, int how, kernel_sigset_t *set,
-                   kernel_sigset_t *oset, size_t sigsetsize);
+                   kernel_sigset_t *oset, size_t sigsetsize, uint *errno_val);
```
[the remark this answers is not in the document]
abhinav92003: Done.

suite/tests/linux/sigaction.c
abhinav92003: Added test for SYS_sigprocmask in mac. The 32-bit SYS_sigprocmask apparently does not have any handling currently. Can add that in a later PR.

suite/tests/linux/sigaction.c
abhinav92003: Done.

core/unix/signal.c
abhinav92003: Will need to research a bit more about what kernel expects in the size arg. I'm leaning towards submitting this fix for a bad address first, so that the immediate issue is resolved. Added a TODO to check bad size arg too.

suite/tests/linux/sigaction.c
abhinav92003: See above.

suite/tests/linux/sigaction.c
abhinav92003: Already there below in the "success" section.

core/unix/signal.c
```diff
@@ -2293,17 +2293,29 @@ check_signals_pending(dcontext_t *dcontext, thread_sig_info_t *info)
...
 handle_sigprocmask(dcontext_t *dcontext, int how, kernel_sigset_t *app_set,
-                   kernel_sigset_t *oset, size_t sigsetsize)
+                   kernel_sigset_t *oset, size_t sigsetsize, uint *errno_val)
 {
     thread_sig_info_t *info = (thread_sig_info_t *)dcontext->signal_field;
     int i;
-    kernel_sigset_t safe_set;
+    kernel_sigset_t safe_set, safe_old_set;
+    if ((app_set != NULL && !d_r_safe_read(app_set, sizeof(safe_set), &safe_set)) ||
+        (oset != NULL && !d_r_safe_read(oset, sizeof(safe_old_set), &safe_old_set))) {
```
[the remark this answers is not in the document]
abhinav92003: Added check for write privileges for old set, thanks for pointing it out.

core/unix/signal.c
derekbruening: I think we have to return EINVAL for anything but the size we're using now since we won't properly handle it anyway.  We can delay handling some other size if the kernel supports some other size (doubt it).

abhinav92003:
MacOS is seeing some failures now:
```
	code_api|linux.signal0000 
	code_api|linux.signal0010 
	code_api|linux.signal0100 
	code_api|linux.signal0110 
```
Also, I cannot see `linux.sigaction` running in the MacOS suite.

core/unix/signal.c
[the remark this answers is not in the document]
abhinav92003: For the other cases (future write attempts failing etc) I guess there's no way to guarantee success of future writes. Added a comment about that. Also added a TODO to re-use the read sigset instead of making the kernel read the same memory again. Will handle it in a follow up PR so that we can get this one submitted.

core/unix/signal.c
abhinav92003: I'm seeing something weird with MacOS and the size. sigprocmask on MacOS also uses the same path. However, sigprocmask does NOT have a 4th arg for sigsetsize. The current code was anyway extracting the 4th arg and using it as size. When I changed it to use sizeof(sigset_t) for mac, other test started failing (signal0000, etc). So that existing bug needs to be fixed.

abhinav92003: Discussed that Mac and others are all using kernel_sigset_t size and ignoring the passed in size. For non-Mac, made changes to check that the passed-in size is each to sizeof(kernel_sigset_t).

suite/tests/linux/sigaction.c
abhinav92003: Added the test.

core/unix/signal.c
abhinav92003: The sigaction test is still failing on Mac. I tried tmate to debug it on the GA runner: it seems that the pre and post syscall handling isn't even being invoked for sigprocmask. I reverted the mac related changes in test. But keeping the changes in os.c and signal.c code. It still performs the same thing as before for mac, uses sizeof(kernel_sigset_t) for the safe read, and an additional safe-write check with the same size.

abhinav92003: I created i#5256 for this issue on mac.

core/unix/signal.c
```diff
@@ -2293,20 +2293,43 @@ check_signals_pending(dcontext_t *dcontext, thread_sig_info_t *info)
...
-        (oset != NULL && !d_r_safe_read(oset, sizeof(safe_old_set), &safe_old_set))) {
-        if (errno_val != NULL)
-            *errno_val = EFAULT;
+    /* Some code uses this syscall to check whether the given address is
+     * readable. E.g.
+     * github.com/abseil/abseil-cpp/blob/master/absl/debugging/internal/
+     * address_is_readable.cc#L85
+     * Those uses as well as our checks below don't guarantee that the given
+     * address will _remain_ readable or writable, even if they succeed now.
+     * TODO i#5255: DR can atleast pass the safe_set to the actual syscall
```
derekbruening, line 2307: nit: "at least"

core/unix/signal.c
```diff
@@ -2293,20 +2293,43 @@ check_signals_pending(dcontext_t *dcontext, thread_sig_info_t *info)
...
     }
+    if (app_set != NULL && !d_r_safe_read(app_set, sizeof(safe_set), &safe_set)) {
+        if (error_code != NULL)
+            *error_code = EFAULT;
+        return false;
+    }
+    if (oset != NULL) {
+        /* Old sigset should be writable too. */
+        if (!d_r_safe_read(oset, sizeof(safe_old_set), &safe_old_set) ||
+            !safe_write_ex(oset, sizeof(safe_old_set), &safe_old_set, NULL)) {
```
derekbruening, line 2324: I wonder if there is some corner case like failing with EINVAL below where the kernel doesn't clobber the old set but we do here.

derekbruening (approved):

core/unix/signal.c
abhinav92003: The kernel implementation checks for size EINVAL first, then new sigset EFAULT, then 'how' EINVAL, and then old sigset EFAULT. In fact, the abseil code relies on the 'how' EINVAL check happening after the new sigset EFAULT check.

We currently do the two EFAULT checks together. I'll move the old sigset check to last 

https://elixir.bootlin.com/linux/latest/source/kernel/signal.c#L3162
https://elixir.bootlin.com/linux/latest/source/kernel/signal.c#L3076

abhinav92003: And also some more tests to verify this order of checks

abhinav92003:
run arm tests

core/unix/signal.c
abhinav92003: Done.

core/unix/signal.c

Issue #5256, opened 2021-12-16
sigprocmask handling not being invoked on Mac


abhinav92003:
While trying to debug the `linux.sigaction` test on Mac for PR #5255, which uses SYS_sigprocmask (instead of SYS_rt_sigprocmask), I found that the pre and post syscall handling wasn't being invoked at all. This needs further investigation. When done, the `linux.sigaction` test should be enabled on MacOS like PR #5255 did before 2e44926df8a869f723676d6d065da36e64b53b6a.

abhinav92003:
I see that the `find_syscall_num` call in `bb_process_syscall` returns 0x2000000
https://github.com/DynamoRIO/dynamorio/blob/117b00d1832f4d8c75c358ea18019714613f74ba/core/arch/interp.c#L1927.
`os_normalized_sysnum` then maps this to 0: https://github.com/DynamoRIO/dynamorio/blob/117b00d1832f4d8c75c358ea18019714613f74ba/core/unix/os.c#L4884

```
logdir/linux.sigaction.81232.00000000//log.0.191436.html-interp: start_pc = 0x00007fff6f6c69f0
logdir/linux.sigaction.81232.00000000//log.0.191436.html-check_thread_vm_area: pc = 0x00007fff6f6c69f0
logdir/linux.sigaction.81232.00000000//log.0.191436.html-prepend_entry_to_fraglist: putting fragment @0x00007fff6f6c69f0 (shared) on vmarea 0x00007fff2810e000-0x00007fff7b8ca000
logdir/linux.sigaction.81232.00000000//log.0.191436.html-check_thread_vm_area: check_stop = 0x00007fff7b8ca000
logdir/linux.sigaction.81232.00000000//log.0.191436.html-  0x00007fff6f6c69f0  b8 00 00 00 02       mov    $0x02000000 -> %eax
logdir/linux.sigaction.81232.00000000//log.0.191436.html-  0x00007fff6f6c69f5  49 89 ca             mov    %rcx -> %r10
logdir/linux.sigaction.81232.00000000//log.0.191436.html-  0x00007fff6f6c69f8  0f 05                syscall  -> %rcx
logdir/linux.sigaction.81232.00000000//log.0.191436.html-interp: syscall @ 0x00007fff6f6c69f8
logdir/linux.sigaction.81232.00000000//log.0.191436.html-find_syscall_num: found syscall number write: 33554432
[...]

Issue #5257, opened 2021-12-17
Include native run in some tests to verify kernel behaviour


abhinav92003:
For some tests, it makes sense to do a native run of the test app to make sure that the behaviour expected from the kernel (or other external entities emulated by DR) is correct. E.g. in `linux.sigaction` we have asserts that check the expected errno from the `rt_sigprocmask` syscall in various error scenaiors. It'll be good to verify that the kernel's behaviour hasn't changed by running the test app natively too, before running it under DR.

If we don't do the native run, we cannot be sure if what DR is doing is correct (e.g. returning EINVAL instead of EFAULT)

PR #5258, opened 2021-12-17
i#5256 sigprocmask: enable linux.sigaction on MacOS

abhinav92003:
Fixes and enables the linux.sigaction test on MacOS.

On MacOS, sigprocmask has different behaviour than Linux. It
does not fail for an unreadable or unwriteable old sigset. Also,
it does not need a size arg.

Fixes terminate_via_kill on MacOS. We need to add a constant
for Unix syscalls on Mac. Otherwise it results in a hang due to
the debug infinite loop at global_do_syscall_syscall when we
attempt to SIGTERM the process in linux.sigaction.

Fixes is_sys_kill detection on Mac. Things seem to work without
this, but the latest Mac documentation shows that the method
we use for non-MacOS (si_code <= 0) would work on Mac too:
https://www.unix.com/man-page/osx/3HEAD/siginfo.h

Fixes sigaction mask comparison in the linux.sigaction test.

Fixes a corner case where a bad old sigset fails the syscall with
EFAULT, _but_ still succeeds in setting the new sigset.

Also adds a native execution test for linux.sigaction. This is to
verify kernel's behaviour in the CI.

Adds a -no_intercept_all_signals test for linux.sigaction.This is to
verify write faults in the path where DR has to manually modify
the old sigset returned by the kernel in sigprocmask.

Issue: #5257, #5254
Fixes: #5256


abhinav92003:
I'm seeing a hang in the `linux.sigaction` test on MacOS: https://gist.github.com/abhinav92003/3e59ff8fa71872d398890ff37293dd2f. It works fine till the end, but hangs at the SIGTERM. It also hangs when I remove everything but the `kill(SIGTERM...` from the test.

The test ends up succeeding after 90s but that's just because of the timeout.

It works fine when the test is run with `-no_code_api -no_intercept_all_signals`.

abhinav92003:
The problem seems to be that the `is_sys_kill` call at https://github.com/DynamoRIO/dynamorio/blob/686f1a118c045c74f6a8446aef5508ad9cddd6fc/core/unix/signal.c#L6343
does not return true on Mac.

On Linux, the SIGTERM is recognised as a user-generated signal at
https://github.com/DynamoRIO/dynamorio/blob/686f1a118c045c74f6a8446aef5508ad9cddd6fc/core/unix/signal.c#L4851. So it returns true early.
The second condition in `is_sys_kill` also returns false because `is_at_do_syscall` is false, but the syscall is a `SYS_kill`.

abhinav92003:
I made a fix to `is_sys_kill` to allow detecting user-sent signals on Mac, but then I realised that it doesn't make a difference because the if-block executes even if `is_sys_kill` returns false.

Following the bug more, I see that this may be the loop it's hitting:
https://github.com/DynamoRIO/dynamorio/blob/686f1a118c045c74f6a8446aef5508ad9cddd6fc/core/arch/x86/x86.asm#L842

abhinav92003:
run arm tests

core/unix/signal.c
```diff
@@ -6172,7 +6205,7 @@ terminate_via_kill(dcontext_t *dcontext)
      * this may not exit all threads in the address space
      */
     block_cleanup_and_terminate(
-        dcontext, SYS_kill,
+        dcontext, IF_MACOS(SYSCALL_NUM_MARKER_UNIX +) SYS_kill,
```
derekbruening, line 6208: This is 64-bit Mac only (though 32-bit support should probably be dropped)

derekbruening, line 6208: There are multiple calls to *cleanup_and_terminate.  How are the others working on 64-bit?  Here just one is being changed.  Better to adjust the syscall number inside cleanup_and_terminate (and assume it's BSD and not Mach or Machdep) so all calls work and we continue using SYS_ constants.

suite/tests/CMakeLists.txt
```diff
@@ -99,6 +99,8 @@ endif (UNIX)
 set(main_run_list
   # our main configuration
   "SHORT::-code_api"
+  # no_intercept_all_signals tests.
+  "SHORT::ONLY::signal_intercept$::-no_code_api -no_intercept_all_signals"
```
derekbruening, line 103: Put a #5261 comment about removing this option

derekbruening, line 103: This should be UNIX-only right?

suite/tests/CMakeLists.txt
```diff
@@ -4141,7 +4143,10 @@ if (UNIX)
   tobuild(linux.sigplain101 linux/sigplain101.c)
   tobuild(linux.sigplain110 linux/sigplain110.c)
   tobuild(linux.sigplain111 linux/sigplain111.c)
-  tobuild(linux.sigaction   linux/sigaction.c)
+  tobuild(linux.sigaction.signal_intercept   linux/sigaction.c)
```
derekbruening, line 4146: Not sure this needs this suffix?  I would think the "native" suffix on the new test would be sufficient.

suite/tests/linux/sigaction.c
```diff
@@ -78,9 +78,13 @@ test_query(int sig)
     memset((void *)&old_act, 0xff, sizeof(old_act));
     rc = sigaction(sig, NULL, &old_act);
     assert(rc == 0 && old_act.sa_sigaction == first_act.sa_sigaction &&
-           /* The flags do not match due to SA_RESTORER. */
-           /* The rest of mask is uninit stack values from the libc wrapper. */
+    /* The flags do not match due to SA_RESTORER. */
+    /* The rest of mask is uninit stack values from the libc wrapper. */
+#if defined(MACOS)
+           *(int *)&old_act.sa_mask == *(int *)&first_act.sa_mask);
```
derekbruening, line 84: A typedef up top (or maybe IF_MACOS_ELSE) might be more readable.

suite/tests/linux/sigaction.c
```diff
@@ -115,56 +123,96 @@ set_sigaction_handler(int sig, void *action)
     assert(rc == 0);
 }
 
-#if !defined(MACOS)
+static int
+make_sigprocmask(int how, void *sigset, void *oldsigset, int size)
+{
+#if defined(MACOS)
+    /* XXX: Couldn't get the raw syscall to work on Mac. */
```
derekbruening, line 130: With the BSD top bit set?

core/unix/include/syscall_mach.h
```diff
@@ -38,6 +38,7 @@
 #define _SYSCALL_MACH_H_ 1
 
 #define SYSCALL_NUM_MARKER_MACH 0x1000000
+#define SYSCALL_NUM_MARKER_UNIX 0x2000000
```
derekbruening, line 41: Would be good to update os_normalized_sysnum() (and elsewhere?) to use the named constant.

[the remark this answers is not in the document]
abhinav92003: Done.

core/unix/os.c
```diff
@@ -8927,9 +8926,12 @@ post_system_call(dcontext_t *dcontext)
              size_t sigsetsize)
          */
         /* FIXME i#148: Handle syscall failure. */
```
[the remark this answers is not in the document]
abhinav92003: I'm not sure. I think this comments refers to failures returned by the kernel. If that happens perhaps we just need to propagate the same failure to the app, and for -no_intercept_all_signals revert the sig_set to its app value.

Given that we already checked all EFAULT and EINVAL cases already, I don't know in what case this syscall would return failure.

core/unix/signal.c
```diff
@@ -2411,37 +2415,66 @@ handle_sigprocmask(dcontext_t *dcontext, int how, kernel_sigset_t *app_set,
...
-        /* Restore app memory */
+        /* Restore app memory.
+         * XXX i#1187: See comment in handle_sigprocmask about read-only
+         * app_set. Ideally, we would replace app_set with our own copy
+         * pre-syscall so that we only need to restore a pointer to the original
+         * app copy post-syscall.
+         * In case of a write failure below, we do not return an EFAULT as it would
+         * be incorrect to expect app_set to be writeable. We try continuing and
+         * return a cloberred app_set to the app. This is low priority as
+         * -intercept_all_signals is true by default.
```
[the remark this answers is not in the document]
abhinav92003: Do you recommend reverting any part of this PR? Maybe not add more tests for -no_intercept_all_signals or something?

core/unix/signal.c
abhinav92003: Done.

abhinav92003: Before, I wasn't sure if we want to assume the syscall arg to cleanup_and_terminate to be BSD always. I agree that would be better. Done.

core/unix/signal.c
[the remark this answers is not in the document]
abhinav92003: Saw your comment below about the i#5261 TODO. Got it; not reverting anything yet.

suite/tests/CMakeLists.txt
abhinav92003: My thought was to suffix all the tests that need a `-no_intercept_all_signals` variant with `.signal_intercept`.  Are you suggesting using `sigaction` as the regex for the new suite in main_run_list?

core/unix/signal.c
derekbruening: Xref some mach ones in x86.asm -- one is dynamorio_sys_exit which might be the only concern if someone passed that mach number.

suite/tests/CMakeLists.txt
abhinav92003: I assumed the `.signal_intercept` suffix check would filter out non-UNIX already. Do you recommend adding a "UNIX::" option here (it doesn't exist yet)?

suite/tests/linux/sigaction.c
abhinav92003: Done.

suite/tests/linux/sigaction.c
abhinav92003: I tried it, but it didn't work natively either. "Invalid syscall" was the error IIRC.

core/unix/include/syscall_mach.h
abhinav92003: Done.

suite/tests/CMakeLists.txt
abhinav92003: Done.

core/unix/signal.c
```diff
@@ -2411,37 +2415,66 @@ handle_sigprocmask(dcontext_t *dcontext, int how, kernel_sigset_t *app_set,
...
+         * - On non-Mac systems, if oset was initially writable in our
+         *   handle_sigprocmask check, but is unwriteable now, we want to
+         *   return EFAULT.
+         * - On Mac systems, where a bad oset does not cause EFAULT,
+         *   we want to fail the write silently.
+         */
+        if (DYNAMO_OPTION(intercept_all_signals)) {
+            if (!safe_write_ex(oset, sizeof(*oset), &info->pre_syscall_app_sigblocked,
+                               NULL))
+                return IF_MACOS_ELSE(0, EFAULT);
```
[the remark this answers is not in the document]
abhinav92003: This revealed an interesting behaviour.

Even though the syscall with EFAULT due to bad old sigset, it still succeeds in setting the new mask. 
See https://elixir.bootlin.com/linux/latest/source/kernel/signal.c#L3181. oset is checked only after the new mask has been set.

Added this case to tests.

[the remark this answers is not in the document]
abhinav92003: So, we don't need to do anything about the two problems above. We replicate native behaviour.

core/unix/signal.c
```diff
@@ -2411,37 +2415,66 @@ handle_sigprocmask(dcontext_t *dcontext, int how, kernel_sigset_t *app_set,
...
+                        }
+                    }
+                    write_fault = false;
+                },
+                {
+                    /* EXCEPT */
+                    /* nothing: write_fault is already true */
+                });
+            if (write_fault)
+                return IF_MACOS_ELSE(0, EFAULT);
```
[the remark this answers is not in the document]
abhinav92003: See above.

core/unix/signal.c
abhinav92003: The usages in x86.asm seem fine, as the syscall is decided statically, so we know whether they are bsd/mach.

```
        mov      eax, MACH_thread_self_trap
# ifdef X64
        or       eax, SYSCALL_NUM_MARKER_MACH
# else
```

```
        mov      eax, SYS_bsdthread_terminate
        or       eax, SYSCALL_NUM_MARKER_BSD
```

```
        mov      eax, MACH_semaphore_signal_all_trap
        or       eax, SYSCALL_NUM_MARKER_MACH
```

dynamorio_syscall in `core/drlibc/drlibc_x86.asm` blindly assumes BSD though.

suite/tests/CMakeLists.txt
derekbruening: Right but it seems nice to label it explicitly as OS-specific.  There's `LIN` which looks like is actually UNIX since it doesn't look for Linux but rather not-WIN32.

suite/tests/CMakeLists.txt
derekbruening: I guess I'm imagining the added test config being short-lived if we remove -no_intercept_all_signals soon and so the value of new naming seems diminished and we might want to change it back soon in any case.

abhinav92003: Makes sense. Removed the new config. I'm not using the test name itself (`sigaction$`) for the new suite.

abhinav92003: > I'm not using the test name itself (sigaction$) for the new suite.

s/not/now

suite/tests/CMakeLists.txt

suite/tests/linux/sigaction.c
abhinav92003: Unfortunately, gdb on the GA runner is showing some errors so I'm not able to figure out what the wrapper is doing.

derekbruening: A lot easier if you have access to a local Mac.  Note that gdb is not supported on Mac these days and you're supposed to use lldb -- but lldb has a bug continuing past signals.  I remember trying to build my own lldb with the bug patched but failing to get it to build and getting frustrated.

core/unix/signal.c
```diff
@@ -2408,7 +2395,10 @@ handle_sigprocmask(dcontext_t *dcontext, int how, kernel_sigset_t *app_set,
         check_signals_pending(dcontext, info);
     }
     if (!execute_syscall) {
-        handle_post_sigprocmask(dcontext, how, app_set, oset, sigsetsize);
+        int status = handle_post_sigprocmask(dcontext, how, app_set, oset, sigsetsize);
```
derekbruening, line 2398: But we still have the problems pointed out in the prior review, don't we?  If the syscall fails b/c of a bad old-sigset, the whole thing should abort, right?  So we need to undo the changes to info->app_sigblocked, and we might need to check old-sigset before calling check_signals_pending if that's a non-undoable change -- or we have to convince ourselves by looking at the ordering of things that we don't actually deliver a signal there and it's a functional nop (but wastes time in dispatch looking for a signal to deliver when they're all blocked).

suite/tests/linux/sigaction.c
```diff
@@ -207,9 +207,19 @@ test_sigprocmask()
     assert(make_sigprocmask(SIG_SETMASK + 1, &new, NULL, 8) == -1);
     assert(errno == EINVAL);
     /* Bad 'how' EINVAL gets reported before bad old sigset EFAULT. */
-    assert(make_sigprocmask(~0, &new, (void *)0x123, 8) == -1);
+    assert(make_sigprocmask(~0, &new, fault_addr, 8) == -1);
     assert(errno == EINVAL);
 
+    /* EFAULT due to bad old sigset still sets the new mask. */
```
derekbruening, line 213: Really?  Wow: I thought in general syscalls should be single all-or-nothing transactions, and that a failure return value meant total failure, not some complex half-success.  This does not seem like a good thing for the kernel to do IMHO.  Please document this in the DR handling code because this is not what I would expect.

core/unix/signal.c
abhinav92003: > But we still have the problems pointed out in the prior review, don't we? If the syscall fails b/c of a bad old-sigset, the whole thing should abort, right?

As I mentioned in the previous comment thread, we don't need to abort. The syscall has this weird non-atomic behavior where the new sigset still takes hold, even if the syscall fails with EFAULT due to bad old sigset.  

> So we need to undo the changes to info->app_sigblocked,

No need to undo. The changes are made successfully natively too. 

As DR's record of the signals blocked in the app is correct, I don't think we need any changes in how we handle signals.

core/unix/signal.c
```diff
@@ -2321,19 +2321,6 @@ handle_sigprocmask(dcontext_t *dcontext, int how, kernel_sigset_t *app_set,
             *error_code = EINVAL;
         return false;
     }
-    /* On MacOS, sigprocmask does not fail if the old sigset is not readable
-     * or not writeable.
-     */
-    if (IF_MACOS_ELSE(false, oset != NULL)) {
-        kernel_sigset_t safe_old_set;
-        /* Old sigset should be writable too. */
```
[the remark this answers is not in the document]
abhinav92003: > for the -no_intercept_all_signals case we do the wrong thing for a bad old-sigset but that we're going to remove support for that config.

Even for `-no_intercept_all_signals`, the current implementation does the right thing; all tests pass. 

For a bad old-sigset and `-no_intercept_all_signals`, the pre syscall checks will all pass, the syscall will be invoked and will fail on Linux (will succeed on Mac, because no bad-old-sigset failures happen on Mac), then the post-syscall handling will be invoked which will write the old-sigset in a TRY and will fault and return EFAULT again (no error on Mac, EFAULT on Linux).

core/unix/os.c
```diff
@@ -4880,7 +4880,7 @@ os_normalized_sysnum(int num_raw, instr_t *gateway, dcontext_t *dcontext)
         }
     }
 #    ifdef X64
-    if (num_raw >> 24 == 0x2)
+    if (num_raw & SYSCALL_NUM_MARKER_BSD)
```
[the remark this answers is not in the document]
abhinav92003: Done.

suite/tests/linux/sigaction.c
abhinav92003: Done.

suite/tests/linux/sigaction.c
abhinav92003: I'm keeping the wrapper for this PR. But i'll try finding what's wrong and change it in a future PR  if I find.

core/unix/signal.c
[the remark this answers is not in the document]
abhinav92003: Note that in the above case, we need the syscall to be invoked. So that the kernel can update the app signal mask (sigprocmask updates the signal masks even when old-sigset is bad). Added a comment for this.

core/unix/signal.c
derekbruening: OK, my only request then would be to document this kernel behavior in the DR code and not just the test.

core/unix/signal.c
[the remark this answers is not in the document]
derekbruening: Ack

suite/tests/linux/sigaction.c
derekbruening: Ack

core/unix/signal.c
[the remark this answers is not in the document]
derekbruening: OK, if we document the kernel behavior.

derekbruening (approved):

core/unix/signal.c
abhinav92003: Yes, documented already.

PR #5262, opened 2021-12-20
Fix typo in deployment docs

derekbruening:
Fixes a reference to "Dr. Memory" which should be "DynamoRIO" in the
deployment docs.


derekbruening:
win32 failure is client.winxfer output mismatch: #4732

abhinav92003 (approved):

Issue #2267, opened 2017-03-06
x64 common.floatpc_xl8all fails non-deterministically with a truncated pc


derekbruening:
Splitting from #2145 as it does not look like this is easily solved.

I can repro the common.floatpc_xl8all failure seen on Appveyor on my local
win10 and win8.1 VM's every so often: a few times every 1000 iters.

```
% for ((i=0; i<1000; i++)); do echo $i; bin64/drrun -translate_fpu_pc -- suite/tests/bin/common.floatpc.exe; done 2>&1 | tee OUT
```
2 instances

At -loglevel 2, 1 instance out of 1000:

```
688
<log dir=C:\derek\dr\git\build_x64_dbg_tests\logs\common.floatpc.exe.6776.00000001>
<Starting application C:\derek\dr\git\build_x64_dbg_tests\suite\tests\bin\common.floatpc.exe (6776)>
<Initial options = -no_dynamic_options -loglevel 2 -code_api -probe_api -stack_size 56K -max_elide_jmp 0 -max_elide_call 0 -no_inline_ignored_syscalls -translate_fpu_pc -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct -pad_jmps_mark_no_trace >
FXSAVE64 intra is correctly handled
FXSAVE64 inter is **incorrectly** handled
FXSAVE intra is correctly handled
FXSAVE inter is correctly handled
<Stopping application C:\derek\dr\git\build_x64_dbg_tests\suite\tests\bin\common.floatpc.exe (6776)>
```

Comparing success and failure log files relevant lines:

Success:
```
Entry into F3218(0x00007ff6e956118b).0x00007ff6697d15c8 (shared)
Exit from F3218(0x00007ff6e956118b).0x00007ff6697d15e8 (shared) [common.floatpc.exe]

Entry into F3219(0x00007ff6e95611a4).0x00007ff6697d15f8 (shared)
Exit from F3219(0x00007ff6e95611a4).0x00007ff6697d163c (shared) [common.floatpc.exe]
 (target 0x00007ff6e95611b7 not in cache)
float_pc_update: fp state 0x000000719070fbb0
recreate_app_pc -- translating from pc=0x00007ff6697d15d8
recreate_app : pc is in F3218(0x00007ff6e956118b)
ilist for recreation:
TAG  0x00007ff6e956118b
 +0    L3                      48 8b c1             mov    %rcx -> %rax
[...]

Issue #2052, opened 2016-11-01
ASSERT signal.c:4224 in prof_pcs nativeexec test


derekbruening:
http://dynamorio.org/CDash/testDetails.php?test=144053&build=13931
Test: prof_pcs,thread_private|common.nativeexec_exe (Failed) 
Build: debug-internal-32 (X64.Fedora22.gcc531.Ancalagon) on 2016-11-01 03:15:29
Repository revision: 2c6ae55eab665aad7b68cbf23376a1b46a3713a0

```
calling via IAT-style call
nativeexec.dll:import_me1(57)
calling via PLT-style call
nativeexec.dll:import_me2(37)
calling via funky ind call
nativeexec.dll:import_me3(17)
calling nested native
nativeexec.exe:print_int(42)
calling cross-module unwinder
before longjmp, 1
after longjmp
calling indirect ret_imm
 -> 40
calling loop_test
<Application /work/dr/nightly/run/build_debug-internal-32/suite/tests/bin/common.nativeexec (25332).  Internal Error: DynamoRIO debug check failure: /work/dr/nightly/src/core/unix/signal.c:4224 tr == NULL || tr->under_dynamo_control || IS_CLIENT_THREAD(dcontext)
(Error occurred @609 frags)
version 6.2.17106, custom build
-no_dynamic_options -prof_pcs -code_api -stderr_mask 12 -stack_size 56K -max_elide_jmp 0 -max_elide_call 0 -no_shared_bbs -no_shared_traces -thread_private -no_private_ib_in_tls -no_atomic_inlined_linking -no_shared_trace_ibl_routine -no_finite_bb_cache -no_finite_trace_cache -no_inline_ignored_syscalls -native_exec_defa
0x505d2950 0x710e5a52
0x505d2a84 0x712700eb
0x505d2af0 0x7123fa68
0xff832c5c 0x711b670d
0xff832c78 0x711b6d8c
0xff832c94 0x7123fb3c
0xff832dfc 0x08049474
0xff832e08 0x7123faa7>
```

derekbruening:
Again:

http://dynamorio.org/CDash/testDetails.php?test=201116&build=18510

PR #5264, opened 2021-12-20
i#1807 long failures: List 8 long suite tests as flaky

derekbruening:
Adds the following tests to the ignore list in order to keep the long
suite on master merges greener:

+ win64 'common.floatpc_xl8all' #2267
+ win64 'code_api|tool.drcachesim.simple-config-file' #1807
+ lin32 'code_api,opt_speed|common.fib' #1807
+ lin32 'prof_pcs|common.nativeexec_exe_opt' #2052
+ lin64 'code_api,opt_speed|common.floatpc_xl8all' #1807

Only ignored in the long suite since these are tests we don't want to
completely break:

+ lin32 'code_api|api.startstop' #4604
+ lin64 'code_api|api.detach_state' #5123
+ lin64 'code_api|client.cleancallsig' #1807

Issue: #1807, #2267, #2052, #4604, #5123

### DIFF
--- a/suite/runsuite_wrapper.pl
+++ b/suite/runsuite_wrapper.pl
@@ -49,6 +49,7 @@ use File::Basename;
 my $mydir = dirname(abs_path($0));
 my $is_CI = 0;
 my $is_aarchxx = $Config{archname} =~ /(aarch64)|(arm)/;
+my $is_long = $ENV{'CI_TRIGGER'} eq 'push' && $ENV{'CI_BRANCH'} eq 'refs/heads/master';
 
 # Forward args to runsuite.cmake:
 my $args = '';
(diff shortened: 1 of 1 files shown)


derekbruening:
Hmm, this PR ignores lin32 api.startstop failures in the long suite only: but here in the short suite we hit #4604 again.  Should we put it on the ignore list for PR's too for lin32??

abhinav92003:
> Hmm, this PR ignores lin32 api.startstop failures in the long suite only: but here in the short suite we hit #4604 again. Should we put it on the ignore list for PR's too for lin32??

As the PR description says, this is one of the important ones. So if there's flakiness in the PR suite, let's not ignore that. Better to ignore only in the long suite.

abhinav92003:
For now, I guess just a re-run might do.

Do you have any guesses on the root-cause? If this is high priority, I can try solving it this/next week.

abhinav92003 (approved):

derekbruening:
> For now, I guess just a re-run might do.

In progress.

> Do you have any guesses on the root-cause? If this is high priority, I can try solving it this/next week.

Actually for this one I reproduced on my old laptop and meant to try to debug it but never had time:
https://github.com/DynamoRIO/dynamorio/issues/4604#issuecomment-924446463
Have not yet tried on the new one.
Sure, if you can reproduce, that would be good.  #5123 is the other one I'm worried about, but I can't repro that one.

Issue #4366, opened 2020-07-09
Replace insensitive option and variable names


derekbruening:
Both DR and Dr. Memory have options and variables with the terms "whitelist" and "blacklist".  This option covers replacing them with "allowlist" and "blocklist".

derekbruening:
Dr. Memory is done as linked above.  DR remains.

derekbruening:
There is also the Jenkins auto-add phrase: @AssadHashmi if you could change that to "add to allowlist"

AssadHashmi:
> There is also the Jenkins auto-add phrase: @AssadHashmi if you could change that to "add to allowlist"

Done.

PR #5267, opened 2021-12-23
i#5266: Support CFLAGS

### LINKED ISSUE #5266: Support CFLAGS
I am trying to figure out how to pass additional CFLAGS when building libraries.
The only option I found that works now is overriding CC to include a compiler with the flags:
```
(rm -rf dynamorio-test && mkdir -p dynamorio-test && cd dynamorio-test && \
CC="gcc -fsanitize=address" cmake -GNinja ../third_party/dynamorio && \
ninja libdrdecode.a && \
objdump -d lib64/libdrdecode.a | grep 0x7fff8000 | wc -l)
4141
```
But the following options don't work 
```
(rm -rf dynamorio-test && mkdir -p dynamorio-test && cd dynamorio-test && \
CFLAGS="-fsanitize=address" cmake -GNinja ../third_party/dynamorio && \
ninja libdrdecode.a && \
objdump -d lib64/libdrdecode.a | grep 0x7fff8000 | wc -l)
0
```
and:
```
(rm -rf dynamorio-test && mkdir -p dynamorio-test && cd dynamorio-test && \
cmake -GNinja -DCMAKE_C_FLAGS="-fsanitize=address" ../third_party/dynamorio && \
ninja libdrdecode.a && \
objdump -d lib64/libdrdecode.a | grep 0x7fff8000 | wc -l)
0
```

Overriding `CC` is quite inconvenient when combined with cross-compilation/toolchains.
I see docs mention setting `CFLAGS=-m32`. I am not sure if this broken, or wasn't supported, or I am holding it wrong...

dvyukov:
Append env CFLAGS to compiler invocations.
This allows users to build custom configurations (e.g. asan/msan).

Fixes #5266


CMakeLists.txt
```diff
@@ -1157,16 +1157,16 @@ endif (UNIX)
 #   Release => -O3 -NDEBUG
 # Right now we only support gcc and cl but could change in future
 if (DEBUG)
-  set(CMAKE_C_FLAGS "${BASE_CFLAGS} ${BASE_CONLY_FLAGS} ${DBG} ${DBG_OPT}")
-  set(CMAKE_CXX_FLAGS "${BASE_CFLAGS} ${BASE_CXXONLY_FLAGS} ${DBG} ${DBG_OPT}")
+  set(CMAKE_C_FLAGS "${BASE_CFLAGS} ${BASE_CONLY_FLAGS} ${DBG} ${DBG_OPT} $ENV{CFLAGS}")
```
abhinav92003, line 1160: I think it's better to define a new environment variable for these additional options. `ADDITIONAL_CFLAGS` maybe. This flag can then be set before invoking CMake. Does that still solve your use-case?

CMakeLists.txt
```diff
@@ -1157,16 +1157,16 @@ endif (UNIX)
 #   Release => -O3 -NDEBUG
```
abhinav92003, line 1157: Re the PR title `i#4400: Support CFLAGS`

The issue number needs to be corrected.

abhinav92003, line 1157: I'm confused by the i#5266 comment. Did this solution work for you? If not, I  notice that we reset the `CFLAGS` env var here:
https://github.com/DynamoRIO/dynamorio/blob/27193828261be587d06bcbbca3eabc49726d441f/suite/runsuite.cmake#L413
I haven't verified if this is the cause; can you check?

CMakeLists.txt
dvyukov: Yes, it would solve my use-case, but why leave CFLAGS broken and introduce a new variable that nobody knows about?

CMakeLists.txt
dvyukov: Ooops. But how? Derek asked me to not force push, so I can't fix up an existing commit...

dvyukov: What comment? Yes, this solution works perfectly for me.
I don't think anything in suite/runsuite.cmake will anyhow affect things. I am building just libdrdecode.a not the tests.

abhinav92003: Editing the title at the top of the PR page would be sufficient. No need to fix it in the commit.

CMakeLists.txt
abhinav92003: I was just concerned about our build picking up any flags from the env var that we don't intend to. Using a new env var and also documenting it in `api/docs/building.dox` would be the safest I think.

CMakeLists.txt
dvyukov: Done

CMakeLists.txt
abhinav92003: Derek's comment on the issue (https://github.com/DynamoRIO/dynamorio/issues/5266#issuecomment-1001689348) seems to agree with using a new environment variable. Can you make the change?

derekbruening: I think using `CFLAGS` seems good since that's the method for passing flags like `-m32` today.  Since it is an explicit user-set list of flags, if the user includes flags that are not compatible with DR (or DrMemory) it is a user error, as opposed to the default sets of initial flags which we need to carefully handle.  We would assume the user knows what they are doing for `CFLAGS`.

CMakeLists.txt
```diff
@@ -1157,16 +1157,16 @@ endif (UNIX)
...
 # Right now we only support gcc and cl but could change in future
 if (DEBUG)
-  set(CMAKE_C_FLAGS "${BASE_CFLAGS} ${BASE_CONLY_FLAGS} ${DBG} ${DBG_OPT}")
-  set(CMAKE_CXX_FLAGS "${BASE_CFLAGS} ${BASE_CXXONLY_FLAGS} ${DBG} ${DBG_OPT}")
+  set(CMAKE_C_FLAGS "${BASE_CFLAGS} ${BASE_CONLY_FLAGS} ${DBG} ${DBG_OPT} $ENV{CFLAGS}")
+  set(CMAKE_CXX_FLAGS "${BASE_CFLAGS} ${BASE_CXXONLY_FLAGS} ${DBG} ${DBG_OPT} $ENV{CFLAGS}")
```
derekbruening, line 1161: This should be `CXXFLAGS`.  (We dup CFLAGS into that if unset for convenience, but a separate `CXXFLAGS` is the convention assumed by CMake so we should follow that.)

Ditto below.

CMakeLists.txt
derekbruening, line 1160: With this in place maybe we should remove the manual addition of `-m32` at line 786 (else it will be on the list twice).

dvyukov: Done

dvyukov: Oh, now I see how `CFLAGS=-m32` works today even that we don't add whole `CFLAGS` to the compilation flags.

dvyukov (commented):
PTAL

CMakeLists.txt
```diff
@@ -780,10 +780,8 @@ if (UNIX)
   # now that the user must set non-default -mNN we don't nec. need these:
```
derekbruening, line 780: I believe this comment is now stale and can be removed.

derekbruening (approved):

CMakeLists.txt
abhinav92003: Reminder to address this comment.

abhinav92003 (approved):

abhinav92003:
run arm tests

abhinav92003:
add to whitelist

abhinav92003:
Despite the above comment, I don't see dvyukov added to the Jenkins whitelist. Perhaps it needs an Admin's comment (either @derekbruening or Assad)

derekbruening:
Note that this should be changed to "add to allowlist" per #4366: @AssadHashmi if you could make that change

derekbruening:
add to whitelist

derekbruening:
run arm tests

CMakeLists.txt
dvyukov: Done.

AssadHashmi:
> Note that this should be changed to "add to allowlist" per #4366: @AssadHashmi if you could make that change

Done.

PR #5270, opened 2022-01-05
i#5269: Upgrade to clang-format-9

### LINKED ISSUE #5269: update to clang-format-9
We are currently using clang-format-6 but it is getting difficult to maintain that on new systems.  This is a proposal to update to clang-format-9.  Some files do have drastic differences with version 9, such as optionsx.h, so we'd do a single reformat across the code base.

derekbruening:
Updates the official clang-format version from 6.0 to 9.0.  Re-formats
all of the .h, .c, and .cpp files, which results in mostly minor
changes, but substantial changes in core/optionsx.h and
core/lib/globals_api.h.

Fixes #5269


clients/drcachesim/analysis_tool.h
```diff
@@ -90,8 +90,8 @@ class analysis_tool_t {
      * error message.
```
abhinav92003, line 90: Update year range in Copyright notice.

clients/drcachesim/tests/heap_test.cpp
```diff
@@ -71,36 +71,36 @@ test_operators()
             int x = 0;
```
abhinav92003, line 71: Update year range in Copyright notice.

core/annotations.c
```diff
@@ -352,12 +352,13 @@ instrument_annotation(dcontext_t *dcontext, IN OUT app_pc *start_pc,
 #    endif
```
abhinav92003, line 352: Update year range in Copyright notice, and elsewhere too.

abhinav92003 (approved):

clients/drcachesim/analysis_tool.h
derekbruening: Done for all files changed here.

clients/drcachesim/tests/heap_test.cpp
derekbruening: Done for all files changed here.

core/annotations.c
derekbruening: Done for all files changed here.

PR #5271, opened 2022-01-06
Add alias support to droption

derekbruening:
Adds support for multiple names for one option.
Adds a simple test.

This is needed for renaming a drcpusim option while keeping legacy
support for #4366.

Issue: #4366

### DIFF
--- a/suite/tests/CMakeLists.txt
+++ b/suite/tests/CMakeLists.txt
@@ -1,5 +1,5 @@
 # **********************************************************
-# Copyright (c) 2010-2021 Google, Inc.    All rights reserved.
+# Copyright (c) 2010-2022 Google, Inc.    All rights reserved.
 # Copyright (c) 2009-2010 VMware, Inc.    All rights reserved.
 # Copyright (c) 2016      ARM Limited.    All rights reserved.
 # **********************************************************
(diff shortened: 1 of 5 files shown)


derekbruening:
Trying to resolve these confusing C++ overload issues with certain compilers (OSX, Android).

suite/tests/CMakeLists.txt
```diff
@@ -2532,7 +2532,7 @@ torunonly_ci(client.blackbox ${ci_shared_app} client.blackbox.dll.A
   "" "" "")
 
 tobuild_ci(client.option_parse client-interface/option_parse.cpp
```
abhinav92003, line 2534: Same as what you did for -x (set it to 3 then override to 4), maybe do the same for z (use -no_z first, then override to true using -z_alias)

abhinav92003 (approved):

suite/tests/CMakeLists.txt
derekbruening: -z is a string: do you mean to add an alias for a bool option (like `-flag`)?

abhinav92003: Oh z is not a bool. Maybe do it for some bool option too

derekbruening: I'll assume that's what you meant.  Adding 2 aliases so will test 3 names.

derekbruening:
Failure is drcachesim.invariants type_is_instr assert #3320

PR #5272, opened 2022-01-07
i#4366: Rename filter lists

derekbruening:
Renames filter lists:
+ s/blacklist/blocklist/
+ s/whitelist/allowlist/

This renames a public-facing drcpusim option:
+ s/-blacklist/-blocklist/
The legacy option is still supported using a droption alias.
Adds a test of both option names.

This also renames some non-publicized DR options:
+ s/whitelist_company_names_default/allowlist_company_names_default/
+ s/whitelist_company_names/allowlist_company_names/
It also renames DR variables like DYNAMORIO_VAR_APP_PROCESS_ALLOWLIST_ID.
We do not bother to provide aliases to support the old names for these
options and variables as they are not documented or commonly used anymore
and maybe should be removed in some future cleanup.

Issue #4366


clients/drcpusim/options.cpp
```diff
@@ -127,10 +127,13 @@ droption_t<bool> op_allow_prefetchw(
     "processors, while they do not officially support it, will turn it into a NOP. "
     "As it is commonly seen on Windows, by default drcpusim does not complain about it.");
 
-droption_t<std::string> op_blacklist(
-    DROPTION_SCOPE_CLIENT, "blacklist", IF_WINDOWS_ELSE("ntdll.dll", ""),
-    ":-separated list of libs to ignore.",
-    "The blacklist is a :-separated list of library names for which violations "
+droption_t<std::string> op_blocklist(
+    // We support the legacy name as an alias for compatibility.
+    // We explicitly specity the vector type to avoid ambiguity with iterators
```
abhinav92003, line 132: You mean otherwise it may be considered an `std::pair`?

abhinav92003 (approved):

clients/drcpusim/options.cpp
derekbruening: No, it considers the 2 char* to define a start and end iterator.

Issue #3805, opened 2019-08-30
Add VS2019 support


baptistapedro:
````find_program(CMAKE_ASM_COMPILER ml64.exe HINTS "${cl_path}" DOC "path to assembler")
    if ("${CMAKE_GENERATOR}" MATCHES "NMake Makefiles")
      set(CMAKE_ASM_COMPILER "ml64.exe")
    endif()
  else (X64)
    find_program(CMAKE_ASM_COMPILER ml.exe HINTS "${cl_path}" DOC "path to assembler")
    if ("${CMAKE_GENERATOR}" MATCHES "NMake Makefiles")
      set(CMAKE_ASM_COMPILER "ml.exe")
    endif()
  endif (X64)
  if (NOT CMAKE_ASM_COMPILER)
    message(FATAL_ERROR "assembler not found: required to build")
  endif (NOT CMAKE_ASM_COMPILER)
  message(STATUS "Found assembler:````


I am doing cmake -G"Visual Studio 16 Win64" ..
and I happen to have an error on the exact line   if (NOT CMAKE_ASM_COMPILER)
    message(FATAL_ERROR "assembler not found: required to build")

I am not sure why it doesnt find the compiler since I'm using the VS with all Build Tools installed, plus the cmake should do it by yourself?

**Comment by derekbruening** [comment]
> I am doing cmake -G"Visual Studio 16 Win64" ..

"Visual Studio 16" is VS2019 which is not supported.  Please see https://github.com/DynamoRIO/dynamorio/wiki/How-To-Build, #2166, and #2924.  I suppose there's no issue yet on VS2019 support but VS2017 #2924 should be addressed first, and there is some ongoing work on that.

**Comment by derekbruening** [comment]
Changing into a feature request to add VS2019 support.  As mentioned, please do not start on this without synchronizing with VS2017 #2924 work first.

**Comment by m-0day** [comment]
Does this support VS 2019 now? I am getting the same error on build...thx

**Comment by derekbruening** [comment]
No, VS2017 should build but has not yet been added to Appveyor.  If you would like to add VS2019 support we are happy to review a PR.

**Comment by firodj** [comment]
I try to install VS2017 on my new installed VM, but currently there is only VS2019 installer so I trying a combination a bit:

### VS2019 with VS2017 (v141 | v14.16):

1. Start command line to use cl.exe from v14.16 not VS 2019 (v142 | v14.24)

```
[...]

Issue #5195, opened 2021-11-07
Upgrade GA CI from windows-2016 which is deprecated and will be removed 3/15/22


derekbruening:
Github gives us this message:

The windows-2016 environment is deprecated and will be removed on March 15, 2022. Migrate to windows-latest instead. For more details see https://github.com/actions/virtual-environments/issues/4312

This issue covers updating for our DR and DrMemory jobs.

derekbruening:
Looks like the GA CI windows-2019 has VS2019 so we'd need #3805

abhinav92003:
Test failures and their immediate causes:

`client.drsyms-testgcc`
expected:
```
Failed to unmangle _ZN10linked_ptrIN12CrxInstaller14WhitelistEntryEE4copyIS1_EEvPKS_IT_E
got correct overflow
Failed to unmangle std::operator<<<<<std::char_traits<char, truncated
got correct overflow
finished unmangling.
enumerating with DRSYM_LEAVE_MANGLED
enumerating with DRSYM_DEMANGLE
enumerating with DRSYM_DEMANGLE_FULL
found drsyms-test.appdll.cpp
stack trace:
drsyms-test\.appdll\.cpp:60!dll_public(\(\))?
overloaded: 1
overloaded: 2
overloaded: 4
app num_calls: 7
got back 4
got back 4
```
actual:
```
Failed to unmangle _ZN10linked_ptrIN12CrxInstaller14WhitelistEntryEE4copyIS1_EEvPKS_IT_E
got correct overflow
Unexpected unmangling:
 actual: testing::internal::ActionResultHolder<void>::PerformAction<void >(class testing::Action<void > const &,class std::tr1::tuple<class FilePath const &,class FilePath const &,class base::DictionaryValue const *,class Extension const *,void,void,void,void,void,void> const &)
 expected: testing::internal::ActionResultHolder<void>::PerformAction<void (class FilePath const &,class FilePath const &,class base::DictionaryValue const *,class Extension const *)>(class testing::Action<void (class FilePath const &,class FilePath const &,class base::DictionaryValue const *,class Extension const *)> const &,class std::tr1::tuple<class FilePath const &,class FilePath const &,class base::DictionaryValue const *,class Extension const *,void,void,void,void,void,void> const &)
Failed to unmangle std::operator<<<<<std::char_traits<char, truncated
got correct overflow
finished unmangling.
enumerating with DRSYM_LEAVE_MANGLED
[...]

PR #5274, opened 2022-01-11
i#3805 msvc2019: add support for Microsoft Visual Studio 2019 and Tools for Visual Studio 2019

StephanvanSchaik:
This addresses issue #3805 as follows:

- Remove old detection code for CMake versions older than 2.6.3 to detect `cl_path`. This was hindering CMake from finding the correct `cl_path`.
- Ensure `cl_path` is defined before `cpp2asm_support.cmake` gets included, such that `ml.exe` and `ml64.exe` for x86 and x64 respectively are found in the correct path.
- Specify `cl_path` as a hint when trying to find `lib.exe` and `dumpbin.exe`.
- The Windows SDK that is being used may be installed into `C:\Program Files (x86)\Windows Kits\10`, of which the installation path is covered by a different registry key, namely `HKEY_LOCAL_MACHINE\Software\Microsoft\Windows Kits\Installed Roots\KitsRoot10`. Check if the registry key points to an existing path, and use it as `kits_dir` if it does and then use that to derive the correct `sdk_bindir`. Otherwise fall back to the old method of detecting `sdk_bin` and `sdk_bindir`.
- When MFC is installed, this will try building drstats. Since drstats has hardcoded paths to `Uuid.Lib`, we now simply use `kits_dir` to specify the correct path to `Uuid.Lib`.

I have tested this with Tools for Visual Studio 2019 as follows:

```
git clone https://github.com/StephanvanSchaik/dynamorio -b i3805-msvc2019
cd dynamorio
mkdir build
cd build
cmake ..
cmd.exe
call "C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Auxiliary\Build\vcvarsall.bat" amd64
msbuild.exe DynamoRIO.sln
```

This builds everything successfully. However, do note that you should not have Windows 11 SDK (version 10.0.22000.0) installed. Trying to build DynamoRIO against the Windows 11 SDK will fail with rc.exe emitting error RC2188 due to line 253 in `C:\Program Files (x86)\Windows Kits\10\Include\10.0.22000.0\um\winnt.h`. See also: https://bugs.python.org/issue45220 (EDIT: the fix for this is to ensure `_AMD64_` or `_X86_` are defined when building).

### DIFF
--- a/.github/workflows/ci-windows.yml
+++ b/.github/workflows/ci-windows.yml
@@ -40,7 +40,7 @@ on:
     branches:
       - master
   pull_request:
-    types: [opened, reopened, synchroni]
+    types: [opened, reopened, synchronize]
 
   # Manual trigger using the Actions page. May remove when integration complete.
   workflow_dispatch:
--- a/.github/workflows/ci-windows.yml
+++ b/.github/workflows/ci-windows.yml
@@ -40,7 +40,7 @@ on:
     branches:
       - master
   pull_request:
-    types: [opened, reopened, synchronize]
+    types: [opened, reopened, synchroni]
 
   # Manual trigger using the Actions page. May remove when integration complete.
   workflow_dispatch:
@@ -246,7 +246,7 @@ jobs:
         from: Github Action CI
 
   ###########################################################################
-  # 32-bit and 64-bit VS2019 release builds:
+  # 32-bit and 64-bit VS2019 debug builds:
   vs2019-builds:
     runs-on: windows-2019
 
--- a/.github/workflows/ci-windows.yml
+++ b/.github/workflows/ci-windows.yml
@@ -244,3 +244,68 @@ jobs:
           See more details on github.com/DynamoRIO/dynamorio/actions/runs/${{github.run_id}}
         to: dynamorio-devs@googlegroups.com
         from: Github Action CI
+
+  ###########################################################################
+  # 32-bit and 64-bit VS2019 release builds:
+  vs2019-builds:
+    runs-on: windows-2019
+
+    steps:
+    - uses: actions/checkout@v2
+
+    # Cancel any prior runs for a PR (but do not cancel master branch runs).
+    - uses: n1hility/cancel-previous-runs@v2
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+      if: ${{ github.event_name == 'pull_request' }}
+
+    - name: Fetch Sources
+      run: git fetch --no-tags --depth=1 origin master
+
+    - name: Download Packages
+      shell: powershell
+      run: |
+        md c:\projects\install
+        (New-Object System.Net.WebClient).DownloadFile("https://github.com/ninja-build/ninja/releases/download/v1.10.2/ninja-win.zip", "c:\projects\install\ninja.zip")
+        (New-Object System.Net.WebClient).DownloadFile("https://sourceforge.net/projects/doxygen/files/rel-1.8.19/doxygen-1.8.19.windows.x64.bin.zip", "c:\projects\install\doxygen.zip")
+
+    - name: Run Suite
+      working-directory: ${{ github.workspace }}
+      run: |
+        echo ------ Setting up paths ------
+        7z x c:\projects\install\ninja.zip -oc:\projects\install\ninja > nul
+        set PATH=c:\projects\install\ninja;%PATH%
+        7z x c:\projects\install\doxygen.zip -oc:\projects\install\doxygen > nul
+        set PATH=c:\projects\install\doxygen;%PATH%
+        call "C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/Auxiliary/Build/vcvars32.bat"
+        echo ------ Running suite ------
+        echo PATH is "%PATH%"
+        echo Running in directory "%CD%"
+        perl suite/runsuite_wrapper.pl automated_ci use_ninja debug_only build_only
+      env:
+        CI_TRIGGER: ${{ github.event_name }}
+        CI_BRANCH: ${{ github.ref }}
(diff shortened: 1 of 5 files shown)


derekbruening:
Thank you for working on this -- this will be needed for #5195

derekbruening:
Adding a GA CI job that uses VS2019 might be the best way to have a test for this as part of this PR.  To focus on building (probably some tests will be flaky on the change of platform; could try running them first: maybe we'll get lucky and they'll be green) and separate test success to #5195, add `build_only` to the runsuite_wrapper args.

I.e., please add a windows-2019 job (probably need 2, one for 32-bit and one for 64-bit).  This could be inside the same file `.github/workflows/ci-windows.yml`: copy the existing ones and add `build_only`.

StephanvanSchaik:
@derekbruening You are welcome!

* I have updated my branch with upstream.
* I have added a workflow for vs2019-builds that builds successfully now and that currently passes the `build_only` argument when running `runsuite_wrapper`.
* I have fixed an issue where either `_AMD64_` or `_X86_` need to be defined on Microsoft Windows when using the Windows SDK 10.0.22000.x but weren't because either the generator did not define them (e.g. the Ninja generator has this issue), or because CMake did not pass the definitions to rc.exe (for the Microsoft Visual Studio generators).

.github/workflows/ci-windows.yml
```diff
@@ -244,3 +244,68 @@ jobs:
           See more details on github.com/DynamoRIO/dynamorio/actions/runs/${{github.run_id}}
         to: dynamorio-devs@googlegroups.com
         from: Github Action CI
+
+  ###########################################################################
+  # 32-bit and 64-bit VS2019 release builds:
```
derekbruening, line 249: Looks like this is release-build-only so it's skipping a bunch of debug code which might contain warnings or issues in vs2019.  Removing the `nontest_only` below should make it build both; or maybe better to replace `nontest_only` with `debug_only` to cover the most code in one job (else this job might become the slowest one) (we'll get release build covered with the switch in #5195).

ext/drsyms/CMakeLists.txt
```diff
@@ -70,7 +70,7 @@ set(DynamoRIO_USE_LIBC ON)
 if (WIN32)
   # XXX: if we add any more of these .lib files we should share this code
   # (currently we have make/ntdll_imports.cmake and here).
-  find_program(LIB_EXECUTABLE lib.exe DOC "path to lib.exe")
+  find_program(LIB_EXECUTABLE lib.exe HINTS ${cl_path} DOC "path to lib.exe")
```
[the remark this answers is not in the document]
StephanvanSchaik: Good catch.

derekbruening:
> StephanvanSchaik force-pushed the i3805-msvc2019 branch from 9120d63 to d570f1d 15 hours ago

For the future, please don't force-push: it breaks the viewing of recent changes (important for reviewing), breaks the comment history, etc.  This is documented on multiple pages such as https://dynamorio.org/page_contributing.html

.github/workflows/ci-windows.yml
derekbruening: nit: It still says "release" here

derekbruening (approved):

.github/workflows/ci-windows.yml
derekbruening: I edited it myself.

derekbruening:
add to allowlist

derekbruening:
run arm tests

StephanvanSchaik:
> > StephanvanSchaik force-pushed the i3805-msvc2019 branch from [9120d63](https://github.com/DynamoRIO/dynamorio/commit/9120d63aca45a9ee281dfd0ddfa923abcee971b4) to [d570f1d](https://github.com/DynamoRIO/dynamorio/commit/d570f1d2c350fa54f8e0d8eac0672da08c50cee7) 15 hours ago
> 
> For the future, please don't force-push: it breaks the viewing of recent changes (important for reviewing), breaks the comment history, etc. This is documented on multiple pages such as https://dynamorio.org/page_contributing.html

@derekbruening Sorry about that, I am a bit used to the Linux kernel reviewing process where you just e-mail patch sets, and essentially just re-open the pull requests for later versions to keep the commits clean. One of the last projects I worked with on Github instead prefered having the same PR open. If instead you prefer additional commits to address the reviews, then I am happy to do that too.

Also sorry about missing it in the contribution page, I did skim through that page to get an overall idea.

derekbruening:
> > > StephanvanSchaik force-pushed the i3805-msvc2019 branch from [9120d63](https://github.com/DynamoRIO/dynamorio/commit/9120d63aca45a9ee281dfd0ddfa923abcee971b4) to [d570f1d](https://github.com/DynamoRIO/dynamorio/commit/d570f1d2c350fa54f8e0d8eac0672da08c50cee7) 15 hours ago
> > 
> > 
> > For the future, please don't force-push: it breaks the viewing of recent changes (important for reviewing), breaks the comment history, etc. This is documented on multiple pages such as https://dynamorio.org/page_contributing.html
> 
> @derekbruening Sorry about that, I am a bit used to the Linux kernel reviewing process where you just e-mail patch sets, and essentially just re-open the pull requests for later versions to keep the commits clean. One of the last projects I worked with on Github instead prefered having the same PR open. If instead you prefer additional commits to address the reviews, then I am happy to do that too.
> 
> Also sorry about missing it in the contribution page, I did skim through that page to get an overall idea.

No problem, different systems can be confusing.  In our case we do a squash-and-merge at the end so the final merge into master is cleaned up into a single commit and single nice description at that point, letting us keep the history with a series of commits during the review process.

derekbruening:
Merged!  Thank you again for the contribution.

PR #5277, opened 2022-01-12
Fix add new module failed when mmap  a partial map on Android.

Iv4n550:
Fix module_walk_program_headers failed when mmap a partial map on Android.
So add module_is_partial_map check before call os_add_new_app_module as other platforms do.

### DIFF
--- a/core/unix/os.c
+++ b/core/unix/os.c
@@ -8340,21 +8340,7 @@ process_mmap(dcontext_t *dcontext, app_pc base, size_t size, uint prot,
                 * read, so pass size=0 to use a safe_read.
                 */
                module_is_header(base, 0)) {
-#ifdef ANDROID
-        /* The Android loader's initial all-segment-covering mmap is anonymous */
-        dr_mem_info_t info;
-        if (query_memory_ex_from_os((byte *)ALIGN_FORWARD(base + size, PAGE_SIZE),
-                                    &info) &&
-            info.prot == MEMPROT_NONE && info.type == DR_MEMTYPE_DATA) {
-            LOG(THREAD, LOG_SYSCALLS, 4, "mmap " PFX ": Android elf\n", base);
-            image = true;
-            DODEBUG({ map_type = "ELF SO"; });
-            os_add_new_app_module(dcontext, true /*at_map*/, base,
-                                  /* pass segment size, not whole module size */
-                                  size, memprot);
-        } else
-#endif
-            if (module_is_partial_map(base, size, memprot)) {
+        if (module_is_partial_map(base, size, memprot)) {
             /* i#1240: App might read first page of ELF header using mmap, which
              * might accidentally be treated as a module load. Heuristically
              * distinguish this by saying that if this is the first mmap for an ELF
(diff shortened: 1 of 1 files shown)


core/unix/os.c
```diff
@@ -8363,10 +8349,25 @@ process_mmap(dcontext_t *dcontext, app_pc base, size_t size, uint prot,
...
+                                        &info) &&
+                info.prot == MEMPROT_NONE && info.type == DR_MEMTYPE_DATA) {
+                LOG(THREAD, LOG_SYSCALLS, 4, "mmap " PFX ": Android elf\n", base);
+                image = true;
+                DODEBUG({ map_type = "ELF SO"; });
+                os_add_new_app_module(dcontext, true /*at_map*/, base,
+                                      /* pass segment size, not whole module size */
+                                      size, memprot);
+            }
+#else
```
derekbruening, line 8365: This code below should be enabled for Android as well for the per-segment mappings, just like in the original code.  There should be an `else` for this for Android, just like before.

derekbruening:
> Iv4n550 force-pushed the iX-fix-android-partial-map-filter branch from b000539 to 5e805bf 12 hours ago

For the future, please don't force-push: it breaks the viewing of just the recent changes (important for reviewing: I'd like to easily see only changes since my last review -- but now the button Github provides shows nothing because of the force-push), often breaks the comment history ties to the code, etc. This is documented on multiple pages such as https://dynamorio.org/page_contributing.html

core/unix/os.c
derekbruening: Again for the future: please mark each suggestion as "Done" if addressed.

derekbruening (approved):
Thank you for the patch.  It would be even better to add some kind of regression test of this scenario but that can be tricky to set up and we don't have good Android testing in general...

derekbruening:
add to allowlist

derekbruening:
run arm tests

petrochenkov:
@Iv4n550
How did you test this change?
It actually broke DR on Android.
Libraries like libc.so, libdl.so are no longer registered, not reported in module init events, etc.

cc @derekbruening

Iv4n550:
> @Iv4n550 How did you test this change? It actually broke DR on Android. Libraries like libc.so, libdl.so are no longer registered, not reported in module init events, etc. (Tested on Android 10.)
>
> cc @derekbruening



> @Iv4n550 How did you test this change? It actually broke DR on Android. Libraries like libc.so, libdl.so are no longer registered, not reported in module init events, etc. (Tested on Android 10.)
>
> cc @derekbruening

Hi, I have noticed this problem. Because the android loader parses the ELF file's load-header firstly, then mmap every single loadable segment separately in [[ElfReader::ReserveAddressSpace](https://android.googlesource.com/platform/bionic/+/master/linker/linker_phdr.cpp#640)](https://android.googlesource.com/platform/bionic/+/master/linker/linker_phdr.cpp#640) , so we may not wait for a true result in module_is_partial_map.

derekbruening:
It would be very helpful if someone currently using DR on Android could add regression tests on an Android emulator: #5388.

petrochenkov:
I'd suggest reverting this in the meantime.

derekbruening:
Yes, if this is a regression, please revert.

PR #5278, opened 2022-01-13
i#2626: AArch64 v8 decode: Add validating tests for GPR instructions

joshua-warburton:
As part of #2626, exhaustive tests for the instructions that use GPR
registers have been generated. Some of these tests have revealed issues
with the current implementation of several of these instructions. This
patch is the first of several and contains all of the tests that validate
existing instructions as correct.

issues: #2626

### DIFF
--- a/suite/tests/api/dis-a64.c
+++ b/suite/tests/api/dis-a64.c
@@ -195,7 +195,10 @@ do_line(void *dc, const char *line, size_t len, bool verbose, bool *failed)
 int
 run_test(void *dc, const char *file, bool verbose)
 {
+    bool test_failed = false;
     bool failed = false;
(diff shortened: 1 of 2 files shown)


AssadHashmi (approved):

derekbruening:
> Some of these tests have revealed issues
> with the current implementation of several of these instructions.

Is there/will there be a list of these known problems?  We have some stability issues on AArch64 and it would be good to know about potential possible causes.

AssadHashmi:
> Is there/will there be a list of these known problems? We have some stability issues on AArch64 and it would be good to know about potential possible causes.

Fix PRs will be submitted for known bugs today and through next week.

Issue #5130, opened 2021-09-28
build failure on Ubuntu21 gcc10.3: drrun has too-recent import: GLOBAL DEFAULT UND stat@GLIBC_2.33


derekbruening:
While reproducing DynamoRIO/drmemory#2425 I made an Ubuntu 21.04 VM and it fails to build all of DR:

```
CMake Error at /home/derek/drmemory/git/src/dynamorio/core/CMake_readelf.cmake:146 (message):
  *** Error: /home/derek/drmemory/git/build_x64_dbg/dynamorio/bin64/drrun has
  too-recent import: GLOBAL DEFAULT UND stat@GLIBC_2.33 (3)
```

derekbruening:
Ended up filing this in the DrM tracker accidentally but will leave that open to cover importing the fix: DynamoRIO/drmemory#2429

derekbruening:
Xref the original #1504.  

Possible action items are:

+ Bump the minimum version where drrun is supported (libdynamorio.so itself will run just about anywhere so this is just about the launcher).

+ Have a separate standard for a release package that is expected to work on random old distros versus someone making a one-off build only for local use.  Maybe via the package-building var that is being set somewhere IIRC.

derekbruening:
@mikaellenz if you are interested perhaps you could take a stab at B) -- sthg like have package.cmake set a cmake var and the launcher of this import check passes that to the sub-script which disable this check for non-packages.

mikaellenz:
Hello @derekbruening !
Based on some posts online, I found  lib32-gcc-libs is missing, so I installed libs (maybe it was absolutely unnecessary.

After I got following warning:
/mnt/ramdisk/drmemory/drsyscall/drsyscall_linux.c: In function ‘handle_semctl’:
/mnt/ramdisk/drmemory/drsyscall/drsyscall_linux.c:979:9: error: ‘arg_val’ is used uninitialized [-Werror=uninitialized]
  979 |     arg = *(union semun *) &arg_val;
      |     ~~^~~~~~~~~~~~~~~~~~~~~~~
/mnt/ramdisk/drmemory/drsyscall/drsyscall_linux.c:973:15: note: ‘arg_val’ declared here
  973 |     ptr_int_t arg_val;
      |               ^~~~~


I fixed warning using following edit (I hope it won't crash running drmemory):
   
     ptr_int_t arg_val=0;

[...]

PR #5280, opened 2022-01-14
i#5130: improve glibc import checks and make them optional

StephanvanSchaik:
This addresses issue #5130 in the following ways:

- Improve the glibc import checking code by using a regex to match all imports, and then for each import extract the version and use CMake's version comparison functionality to match the version. This allows us to write the desired glibc version as a string such as "2.4", instead of having to write a regular expression that matches everything greater than 2.4 or in general the desired version.
- Introduce a `BUILD_PACKAGE` option that is `ON` by default, but can be turned off when invoking `cmake` as follows: `cmake -DBUILD_PACKAGE=OFF ..`. This is important for rolling release distributions where glibc 2.32 is currently stable and these checks prevent drrun from building.
- Bump the glibc version to 2.17. After checking multiple distributions with long support times (typically Debian and CentOS, and Ubuntu LTS being a close one), I found that the oldest versions that are still supported as of writing are Debian Stretch, CentOS 7, RHEL 7, and Ubuntu 16.04 LTS (Xenial). These ship glibc [2.24](https://packages.debian.org/source/stretch/glibc), glibc [2.17](https://centos.pkgs.org/7/centos-x86_64/glibc-2.17-317.el7.x86_64.rpm.html) glibc [2.17](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/7.0_release_notes/sect-red_hat_enterprise_linux-7.0_release_notes-compiler_and_tools-glibc) and glibc [2.23](https://launchpad.net/ubuntu/xenial/+source/glibc) respectively. According to the [release page of glibc](https://ftp.gnu.org/gnu/glibc/) the release date of glibc 2.17 is 2012-12-25.

In addition, I am keeping around the glibc version separation between AArch64 and non-AArch64, just in case they diverge again.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -143,6 +143,11 @@ endif (UNIX)
 
 # many are core-specific
 
+# whether this build of DynamoRIO is meant for distribution, in which case
+# glibc should not have any too-recent imports. See core/CMake_readelf.cmake
+# for more information.
+option(BUILD_PACKAGE "build DynamoRIO for packaging purposes by performing glibc checks" ON)
+
 # configurations that also have defines
 option(DRSTATS_DEMO "build DRstats without the no-longer-supported Run, etc. controls" ON)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -145,8 +145,9 @@ endif (UNIX)
 
 # whether this build of DynamoRIO is meant for distribution, in which case
 # glibc should not have any too-recent imports. See core/CMake_readelf.cmake
-# for more information.
-option(BUILD_PACKAGE "build DynamoRIO for packaging purposes by performing glibc checks" ON)
+# for more information. This is off by default, but gets set by
+# make/package.cmake.
+option(BUILD_PACKAGE "build DynamoRIO for packaging purposes by performing glibc checks" OFF)
 
 # configurations that also have defines
 option(DRSTATS_DEMO "build DRstats without the no-longer-supported Run, etc. controls" ON)
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -131,7 +131,6 @@ changes:
    from source is straight-forward to do on rolling release Linux
    distributions, and enabled by make/package.cmake when building a
    distributable version of DynamoRIO.
- - (Nothing so far: this is a placeholder.)
 
 Further non-compatibility-affecting changes include:
  - Added alias support to droption.
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -126,6 +126,11 @@ clients.
 
 The changes between version \DR_VERSION and 9.0.0 include the following compatibility
 changes:
+ - Introduced a new CMake option called BUILD_PACKAGE to skip glibc
+   compatibility checks. This is off by default such that building DynamoRIO
+   from source is straight-forward to do on rolling release Linux
+   distributions, and enabled by make/package.cmake when building a
+   distributable version of DynamoRIO.
  - (Nothing so far: this is a placeholder.)
 
 Further non-compatibility-affecting changes include:
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -661,7 +661,7 @@ if (UNIX)
         ARGS -D lib_fileloc=${CMAKE_CURRENT_BINARY_DIR}/${locvar_name}
              -D check_textrel=${check_textrel}
              -D check_deps=${check_deps}
-             -D check_libc=OFF
+             -D check_libc=${BUILD_PACKAGE}
              -D check_interp=ON
              -D READELF_EXECUTABLE=${READELF_EXECUTABLE}
              -P ${CMAKE_CURRENT_SOURCE_DIR}/CMake_readelf.cmake
@@ -931,7 +931,7 @@ if (UNIX)
       ARGS -D lib_fileloc=${CMAKE_CURRENT_BINARY_DIR}/${locvar_name}
            -D check_textrel=ON
            -D check_deps=OFF
-           -D check_libc=ON
+           -D check_libc=${BUILD_PACKAGE}
            -D check_interp=ON
            -D READELF_EXECUTABLE=${READELF_EXECUTABLE}
            -P ${CMAKE_CURRENT_SOURCE_DIR}/CMake_readelf.cmake
--- a/core/CMake_readelf.cmake
+++ b/core/CMake_readelf.cmake
@@ -145,7 +145,7 @@ if (check_libc)
     set (glibc_version_regexp " GLOBAL [ A-Z]* UND [^\n]*@GLIBC_([0-9]+\\.[0-9]+)")
     string(REGEX MATCHALL "${glibc_version_regexp}" imports "${string}")
 
-    foreach(import ${imports})
+    foreach (import ${imports})
       string(REGEX MATCH "${glibc_version_regexp}" match "${import}")
       set(version ${CMAKE_MATCH_1})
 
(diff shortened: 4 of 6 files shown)


core/CMake_readelf.cmake
```diff
@@ -127,23 +127,36 @@ if (check_libc)
...
+    # portability without going to extremes with a fixed toolchain (xref i#1504):
+    execute_process(COMMAND
+      ${READELF_EXECUTABLE} -h ${${lib_file}}
+      OUTPUT_VARIABLE file_header_result
+      )
+
+    if (file_header_result MATCHES "AArch64")
+      set (glibc_version "2.17")
+    else ()
+      set (glibc_version "2.17")
```
derekbruening, line 142: If you could duplicate the comment here from the description where you said you checked Ubuntu 16.04 LTS, etc. to come up with this number I think that will help future maintainers to know how to find what to increase this to in the future.

core/CMake_readelf.cmake
```diff
@@ -127,23 +127,36 @@ if (check_libc)
...
+    endif ()
+
+    set (glibc_version_regexp " GLOBAL [ A-Z]* UND [^\n]*@GLIBC_([0-9]+\\.[0-9]+)")
+    string(REGEX MATCHALL "${glibc_version_regexp}" imports "${string}")
+
+    foreach(import ${imports})
```
derekbruening, line 148: nit: s/foreach(/foreach (/

CMakeLists.txt
```diff
@@ -143,6 +143,11 @@ endif (UNIX)
 
 # many are core-specific
 
+# whether this build of DynamoRIO is meant for distribution, in which case
+# glibc should not have any too-recent imports. See core/CMake_readelf.cmake
+# for more information.
+option(BUILD_PACKAGE "build DynamoRIO for packaging purposes by performing glibc checks" ON)
```
derekbruening, line 149: Given https://groups.google.com/g/dynamorio-users/c/Y2cUFCgESHA/m/GLq46qGnBgAJ and others hitting this I'm thinking this should be OFF by default and turned on by our package.cmake.  The logic is that a user obtaining our sources should be able to build for the local machine without having to say "-DBUILD_PACKAGE=OFF"; put the onus on those building published packages to set a non-default value as those users/scripts are already setting various things -- keep the new user "just build this locally" simple.  (I don't think breaking the current default is a significant concern -- maybe add a note to the changelog in api/docs/release.dox.)

derekbruening (commented):
Thank for taking on this issue.  I have a few suggestions and then we should be able to merge this.

core/CMake_readelf.cmake
```diff
@@ -127,23 +127,36 @@ if (check_libc)
...
-  endif ()
-  string(REGEX MATCH " GLOBAL [ A-Z]* UND [^\n]*@GLIBC_${glibc_version_regexp}"
-    has_recent "${string}")
-  if (has_recent)
-    string(REGEX MATCH " GLOBAL DEFAULT  UND [^\n]*@GLIBC_${glibc_version_regexp}[^\n]*\n"
-      symname "${string}")
-    message(FATAL_ERROR "*** Error: ${${lib_file}} has too-recent import: ${symname}")
+
+  if (BUILD_PACKAGE)
+    # Avoid dependences beyond glibc 2.4 (2.17 on AArch64) for maximum backward
```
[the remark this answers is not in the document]
derekbruening: Meant sthg like "oldest well-supported version", not "recent"

core/CMake_readelf.cmake
```diff
@@ -127,23 +127,36 @@ if (check_libc)
...
+    # portability without going to extremes with a fixed toolchain (xref i#1504):
+    execute_process(COMMAND
+      ${READELF_EXECUTABLE} -h ${${lib_file}}
+      OUTPUT_VARIABLE file_header_result
+      )
+
+    if (file_header_result MATCHES "AArch64")
```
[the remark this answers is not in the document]
StephanvanSchaik: Yes, I checked to be sure, and from what I found the AArch64 flavors of Debian Stretch, RHEL 7 and Ubuntu 16.04 LTS are the same. In fact, they just use the same glibc for all supported architectures for that particular version of their distribution.

api/docs/release.dox
```diff
@@ -126,6 +126,11 @@ clients.
 
 The changes between version \DR_VERSION and 9.0.0 include the following compatibility
 changes:
+ - Introduced a new CMake option called BUILD_PACKAGE to skip glibc
+   compatibility checks. This is off by default such that building DynamoRIO
+   from source is straight-forward to do on rolling release Linux
+   distributions, and enabled by make/package.cmake when building a
+   distributable version of DynamoRIO.
  - (Nothing so far: this is a placeholder.)
```
derekbruening, line 134: This line should now be removed.

make/package.cmake
```diff
@@ -44,6 +44,9 @@
 cmake_minimum_required(VERSION 3.7)
 set(CTEST_SOURCE_DIRECTORY "${CTEST_SCRIPT_DIRECTORY}/..")
 
+# ensure that we check glibc for version incompatibilities.
+SET(BUILD_PACKAGE ON CACHE BOOL "build DynamoRIO for packaging purposes by performing glibc checks")
```
derekbruening, line 48: This actually won't set it in the builds done by the ctest_build() step we use: it needs to be inside base_cache which is set down at line 162 or so.

It would be nice to have a check that would catch this failing to be set: but I don't know that there's a simple way to do it b/c how would we know we're doing a build that wants the check.

core/CMake_readelf.cmake
```diff
@@ -129,23 +129,31 @@ if (check_libc)
   endif (readelf_result OR readelf_error)
 
   if (BUILD_PACKAGE)
```
derekbruening, line 131: This CMake script is launched separately and so variables in the configure process, like BUILD_PACKAGE, will not be set here: they would have to be passed in explicitly on the command line.  But, we already have a check_libc parameter, so I think we just update the callers who are now passing ON to instead pass the value of BUILD_PACKAGE.

core/CMake_readelf.cmake
```diff
@@ -129,23 +129,31 @@ if (check_libc)
...
-    endif ()
+    # To determine the minimum version of glibc that we should support for packaging, we should
+    # check the Linux distributions that are known not to be rolling releases and offer long
+    # support, and then check the oldest option available that is not yet EOL. As of writing,
+    # these are:
+    #  * Debian Stretch, which is on glibc 2.24
+    #  * CentOS 7/RHEL 7, which is on glibc 2.17
+    #  * Ubuntu 16.04 LTS (Xenial), which is on glibc 2.23
+    #
+    # Therefore, we want to support at least glibc 2.17.
```
derekbruening, line 147: (Nice, looks good.)

StephanvanSchaik:
Thanks for the feedback and pointers. I tested it on my machine with glibc 2.32, where:

- Invoking `cmake ..` just works out of the box.
- Invoking `cmake -DBUILD_PACKAGE=ON ..` errors out on the too-recent glibc imports.
- Invoking `ctest -S package.cmake,build=1` also errors out on the too-recent glibc imports.

core/CMakeLists.txt
```diff
@@ -661,7 +661,7 @@ if (UNIX)
         ARGS -D lib_fileloc=${CMAKE_CURRENT_BINARY_DIR}/${locvar_name}
              -D check_textrel=${check_textrel}
              -D check_deps=${check_deps}
-             -D check_libc=OFF
+             -D check_libc=${BUILD_PACKAGE}
```
derekbruening, line 664: This was always off before because there are no glibc imports for libdynamorio.  But I guess there's no harm in having it on for BUILD_PACKAGE=ON.

derekbruening (approved):

derekbruening:
(Generally we'd put "Fixes #5130" at the bottom of the final commit message which will auto-close the issue.)

PR #5281, opened 2022-01-14
add support for the encls/enclu/enclv instructions

StephanvanSchaik:
This PR adds support for the `encls`, `enclu` and `enclv` instructions as described by [chapter 41 "SGX Instruction References"](https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-vol-3d-part-4-manual.pdf) in volume 3D of the Intel 64 and IA-32 Architectures Software Developer's Manual. These are used in **s**upervisor mode, **u**ser mode and **V**MM mode respectively.

These instructions are somewhat similar to `cpuid`, in that they accept all have a leaf function that can be specified through the `eax` register. However, unlike `cpuid`, the `ebx`, `ecx` and `edx` can both be used as inputs and outputs depending on the specific leaf function.

While Intel SGX is usable on a select set of Intel CPUs, it is usable both in 32-bit and 64-bit mode on those CPUs.

This PR is as follows:

* Implement `encls`, `enclu` and `enclv` in the decoder for x86.
* Add `INSTR_CREATE` macros for `encls`, `enclu` and `enclv`.
* Add test cases for `encls`, `enclu` and `enclv`.

### DIFF
--- a/core/ir/x86/instr_create_api.h
+++ b/core/ir/x86/instr_create_api.h
@@ -5039,6 +5039,26 @@
                            opnd_create_reg(DR_REG_EBX), opnd_create_reg(DR_REG_ECX), \
                            opnd_create_reg(DR_REG_EDX), opnd_create_reg(DR_REG_EAX), \
                            opnd_create_reg(DR_REG_ECX))
+
+/* 3 implicit destinations, 1 source */
+#define INSTR_CREATE_encls(dc)                                                       \
+    instr_create_4dst_4src((dc), OP_encls, opnd_create_reg(DR_REG_EAX),              \
+                           opnd_create_reg(DR_REG_EBX), opnd_create_reg(DR_REG_ECX), \
+                           opnd_create_reg(DR_REG_EDX), opnd_create_reg(DR_REG_EAX), \
+                           opnd_create_reg(DR_REG_EBX), opnd_create_reg(DR_REG_ECX), \
+                           opnd_create_reg(DR_REG_EDX))
+#define INSTR_CREATE_enclu(dc)                                                       \
+    instr_create_4dst_4src((dc), OP_enclu, opnd_create_reg(DR_REG_EAX),              \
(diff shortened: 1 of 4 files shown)


core/ir/x86/instr_create_api.h
```diff
@@ -5039,6 +5039,26 @@
                            opnd_create_reg(DR_REG_EBX), opnd_create_reg(DR_REG_ECX), \
                            opnd_create_reg(DR_REG_EDX), opnd_create_reg(DR_REG_EAX), \
                            opnd_create_reg(DR_REG_ECX))
+
+/* 3 implicit destinations, 1 source */
```
abhinav92003, line 5043: Can you please explain this further? If I'm not wrong, there seem to be 4 dests (eax, ebx, ecx, edx) and 4 srcs (same regs).

abhinav92003 (commented):
I'm not familiar enough with the x86 decoder, so I'm adding @johnfxgalea. John, can you help review?

core/ir/x86/instr_create_api.h
StephanvanSchaik: I wasn't sure what to write there as I was thinking the same, but then the `cpuid` has the same comment, even though `ecx` can also be a source for `cpuid` (for subleafs). Thanks for confirming though, I will update that comment, but I think the `cpuid` one might need to be corrected too.

johnfxgalea (approved):
LGTM. Thanks for the SGX instructions!

core/ir/x86/instr_create_api.h
johnfxgalea: Indeed, cpuid has 2 implicit sources as well, although ecx is leveraged only in specific cases.

Would you mind fixing the stale comment for cpuid? I think adding the change to this PR and mentioning it in the commit message would suffice, since it requires small modification.

johnfxgalea:
run arm tests

Issue #12, opened 2014-11-27
tests: suite/tests/* object files and results/ should go in top-level build/


derekbruening:
_From [derek.br...@gmail.com](https://code.google.com/u/117968039472581148324/) on February 15, 2009 15:43:32_

xref issue #6 it would be nice to clean up /suite/tests/\* object files to go into
something like /build/tests/

we could put the results/ dir there as well

_Original issue: http://code.google.com/p/dynamorio/issues/detail?id=12_

PR #5282, opened 2022-01-14
i#2626: AArch64 v8 decode: Fix incorrect decode of add/s, sub/s, tb(n)z

joshua-warburton:
This patch fixes two underlying issues:
 * Several of the instruction variants were decoding to an x register
   when they should have been decoding to a w register
 * LSL #12 was being decoded as 0x10 (16) rather than 0x0c (12)

issues: #2626


joshua-warburton:
run arm tests

PR #5283, opened 2022-01-14
i#2626: AArch64 v8 decode: Fix incorrect decode of ldrb

joshua-warburton:
This patch fixes two problems with ldrb:
 * Some variants were decoding to an x register
   when they should have been decoding to a w register
 * The "extension" LSL #0 was being decoded as uxtx #0

issues: #2626

### DIFF
--- a/core/ir/aarch64/codec.c
+++ b/core/ir/aarch64/codec.c
@@ -707,11 +707,13 @@ decode_opnd_memreg_size(opnd_size_t size, uint enc, OUT opnd_t *opnd)
     dr_extend_type_t extend;
     switch (enc >> 13 & 7) {
     case 0b010: extend = DR_EXTEND_UXTW; break;
-    case 0b011: extend = DR_EXTEND_UXTX; break; // Actually LSL
+    // Alias for LSL. LSL preferred in disassembly.
+    case 0b011: extend = DR_EXTEND_UXTX; break;
     case 0b110: extend = DR_EXTEND_SXTW; break;
     case 0b111: extend = DR_EXTEND_SXTX; break;
     default: return false;
     }
+
     *opnd = opnd_create_base_disp_aarch64(
         decode_reg(enc >> 5 & 31, true, true),
         decode_reg(enc >> 16 & 31, TEST(1U << 13, enc), false), extend,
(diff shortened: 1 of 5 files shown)


core/ir/aarch64/codec.c
```diff
@@ -704,9 +704,18 @@ decode_opnd_memreg_size(opnd_size_t size, uint enc, OUT opnd_t *opnd)
 {
     if (!TEST(1U << 14, enc))
         return false;
-    *opnd = opnd_create_base_disp_aarch64(decode_reg(enc >> 5 & 31, true, true),
-                                          decode_reg(enc >> 16 & 31, true, false),
-                                          enc >> 13 & 7, TEST(1U << 12, enc), 0, 0, size);
+    dr_extend_type_t extend;
+    switch (enc >> 13 & 7) {
+    case 0b010: extend = DR_EXTEND_UXTW; break;
+    case 0b011: extend = DR_EXTEND_UXTX; break; // Actually LSL
```
AssadHashmi, line 710: This comment is should be:
`// Alias for LSL. LSL preferred in disassembly.`

AssadHashmi (changes requested):

AssadHashmi (approved):

petrochenkov:
@joshua-warburton @derekbruening 
This change caused some regressions.

Not all places in DR are prepared to encounter a `w` register in a load index context.
For example, the register comparison with a stolen register in `drutil_insert_get_mem_addr_arm` can fail incorrectly now.

I'm not sure what other places besides `drutil_insert_get_mem_addr_arm` may be affected.

petrochenkov:
>What is the nature of the failure in release build? It gets the wrong address but silently?

Yes, by "comparison fails" I mean this branch is not reached
```c
        } else if (index == stolen) {
                                       scratch_used);
        }
```
since `w` registers are never identified as stolen.

It doesn't result in a crash, just silently results in incorrect instrumentation.
We noticed this pretty late by encountering "impossible" load addresses in traces written by instrumentation.

petrochenkov:
>could you file new issue in the tracker on this?

Done: https://github.com/DynamoRIO/dynamorio/issues/5498

derekbruening:
Trying to make it clear what the changes here involve: all issues with the scaling/extension of index registers are purely cosmetic; there were not problems with decoding and computing the scale/extension, just with the disassembly.
See also https://github.com/DynamoRIO/dynamorio/issues/5498#issuecomment-1138952432

So the only address computation bug was with the index being treated as X instead of W.

Issue #3862, opened 2019-09-30
Private Linux loader should read /etc/ld.so.config


derekbruening:
Split from #460.  Currently DR's private loader has a hardcoded list of paths for a simple emulation of reading /etc/ld.so.conf.  This has caused some problems on new platforms.  We may want to just implement reading of the config file and its includes.

Issue #460, opened 2014-11-28
Library search for Linux private loader


derekbruening:
_From [qin.zhao@gmail.com](https://code.google.com/u/qin.zhao@gmail.com/) on May 22, 2011 17:02:23_

When private loader loads client library and their dependent libraries, the process of search them is fairly complicated.
DR currently does not support DT_RPATH, and /etc/ld.so.cache search.

_Original issue: http://code.google.com/p/dynamorio/issues/detail?id=460_

derekbruening:
DT_RUNPATH support was added in 3fe091a09a0f04945d88cc4863d321ecb8aa012a

derekbruening:
The Linux loader is complete enough to close this initial issue in favor of specific, targeting bug or feature issues.  #3862 covers reading /etc/ld.so.config.

PR #5285, opened 2022-01-17
i#3862: parse /etc/ld.so.conf

StephanvanSchaik:
This addresses issue #3862 (originally #460) by implementing code that parses the search paths in `/etc/ld.so.conf` for the private loader on Linux.

The `/etc/ld.so.conf` file consists of one or more lines, where a line may be a search path, an include directive to include another configuration file or where a line may start with '#' if it is a comment. The include directives support globbing (see `man 3 glob` and `man 7 glob` for more information).

Some notes about compatibility:

- Alpine is a musl-based distribution, and the dynamic linker used by Alpine instead may look for`/etc/ld-musl-$(ARCH).path`. For instance, on x86-64, it would look for `/etc/ld-musl-x86_64.path`. The format seems to be slightly different from `/etc/ld.so.conf`. If the file is not present, musl simply defaults to "/lib:/usr/local/lib:/usr/lib". See also: https://www.musl-libc.org/doc/1.0.0/manual.html
- Android uses bionic as its libc and instead looks for `/system/etc/ld.config.txt`. See also: https://android.googlesource.com/platform/bionic/+/master/linker/ld.config.format.md
- While the FreeBSD man pages do seem to mention the existence of `/etc/ld.so.conf`, the default seems to be that the dynamic linker on FreeBSD looks for the `ldconfig_paths` and `ldconfig32_paths` keys in `/etc/defaults/rc.conf`.
- Mac OS X does not seem to have `/etc/ld.so.conf` either, but `man dyld` seems to provide some useful hints.

We probably want to open new issues referring to #3862 and #460 for the other cases.

This PR implements the following:

- The `strncasecmp` function to match the `include` directive. On second thought, `strncmp` may work well enough, but it doesn't hurt to be case-insensitive.
- The `fnmatch` function, which is required to implement `glob`. The `fnmatch` lets us match a given string against a pattern including character ranges (e.g. `[a-z]`), the any character (denoted as `?`) and wildcards (denoted as '*'). Furthermore, multiple wildcards may be present.
- The `glob` function which uses `fnmatch` function and the Unix-specific directory iterator code that is currently present in DynamoRIO to match all found filenames against a given pattern.
- The `getdelim` and `getline` functions which help us parse the config files line-by-line.
- The `privload_parse_ld_conf` function which parses `/etc/ld.so.conf` and any included config files and constructs an array of search paths by piggybacking on the `glob` function, which can be used to append search paths to an array of search paths.
- Finally, the `privload_locate` function has been extended to use the `privload_parse_ld_conf`, and otherwise falls back to the fixed array of system paths (see compatibility notes).

While the `ld.so.conf` parsing itself is straight-forward on any POSIX system due to the existence of `glob` and `getline`, DynamoRIO originally did not provide these functions as the private loader cannot rely on libc, and it would be helpful to have a thorough review of the implementation of these functions.

I have tested this on:
 - My desktop running Gentoo where `libgcc_s.so.1` is located in /usr/lib/gcc/x86_64-pc-linux-gnu/11.2.0` by running: `./bin64/drrun -c ./api/bin/libempty.so -- ls`. Before these changes, it was necessary to specify the correct path using the `LD_LIBRARY_PATH` environment variable.
 - Ubuntu 20.04 LTS by running: `./bin64/drrun -c /api/bin/libempty.so -- ls` to be sure that there are no regressions introduced by this PR.

### DIFF
--- a/core/globals.h
+++ b/core/globals.h
@@ -1103,6 +1103,8 @@ utf16_to_utf8_size(const wchar_t *src, size_t max_chars,
 #    define tolower d_r_tolower
 #    undef strcasecmp
 #    define strcasecmp d_r_strcasecmp
+#    undef strncasecmp
+#    define strncasecmp d_r_strncasecmp
 #    undef strtoul
 #    define strtoul d_r_strtoul
 #endif
@@ -1132,6 +1134,8 @@ int
 tolower(int c);
 int
 strcasecmp(const char *left, const char *right);
+int
+strncasecmp(const char *left, const char *right, size_t n);
 unsigned long
 strtoul(const char *str, char **end, int base);
 
--- a/core/unix/loader.c
+++ b/core/unix/loader.c
@@ -834,6 +834,110 @@ privload_search_rpath(privmod_t *mod, bool runpath, const char *name,
     return false;
 }
 
+bool
+privload_parse_ld_conf(glob_t *search_paths, const char *config_path)
+{
+    file_t file;
+    char *line = NULL;
+    size_t len = 0;
+    ssize_t nread;
+
+    /* Ensure the arguments are not NULL. */
+    if (!search_paths || !config_path) {
+        return false;
+    }
+
+    /* Open the config file as read-only. */
+    file = os_open(config_path, OS_OPEN_READ);
+
+    if (file == INVALID_FILE) {
+        return false;
+    }
+
+    while ((nread = os_getline(&line, &len, file)) >= 0) {
+        /* Remove the newline. */
+        char *end = strchr(line, '\n');
+
+        if (end) {
+            *end = '\0';
+        }
+
+        /* Ignore comments. */
+        if (line[0] == '#') {
+            continue;
+        }
+
+        /* Handle the include directive by recursively parsing the included
+         * config file.
+         */
+        size_t nbytes = strlen("include");
+
+        if (nread < nbytes) {
+            nbytes = nread;
+        }
+
+        if (strncasecmp(line, "include", nbytes) == 0) {
+            glob_t includes = { 0 };
+
+            /* Locate the path that should be included. */
+            char *include_path = line + nbytes;
+
+            /* Make sure there is some form of whitespace after the include
+             * directive.
+             */
+            include_path = strchr(include_path, ' ');
+
+            if (!include_path) {
+                continue;
+            }
+
+            while (*include_path && *include_path == ' ') {
+                ++include_path;
+            }
+
+            /* We have enough room in the string to prefix it with "/etc/". */
+            size_t count = strlen("/etc/");
+            strncpy(include_path - count, "/etc/", count);
+
+            /* Glob the path. */
+            if (os_glob(include_path, 0, NULL, &includes) < 0) {
+                continue;
+            }
+
+            /* Recursively parse the included config files. */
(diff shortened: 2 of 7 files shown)


derekbruening:
Thank you for the patch.   Was out yesterday; will try to to look today.

StephanvanSchaik:
No problem! I realized yesterday this might need some more work from my side since the `glob` function is hard to get correct, so I am going to try to get some fixes out to this PR. I also realized that the libstl_test.so is a much better test since it tries to pull `libgcc_s.so.1` from a different path on my system, whereas `libempty.so` does not use `libgcc_s.so.1` at all.

core/unix/os_private.h
```diff
@@ -269,6 +269,55 @@ os_walk_address_space(memquery_iter_t *iter, bool add_modules);
...
+};
+
+typedef struct os_glob_t os_glob_t;
+
+#ifdef GLOB_MARK
+#    undef GLOB_MARK
+#endif
+#define GLOB_MARK (1 << 0)
+
+#ifdef GLOB_DOOFFS
```
derekbruening, line 285: Wondering whether instead of an identical interface to libc we should provide just what we need, which is a narrower interface and thus easier to test and maintain.  We only need to glob directories so we could not take any options and only provide a directory-globbing routine.  If in the future we needed further glob functionality we could at that point switch back to something like this (since it's all internal and not part of DR's exposed API).  WDYT?

derekbruening (commented):
Didn't finish but here are some early comments.

core/unix/loader.c
```diff
@@ -834,6 +834,110 @@ privload_search_rpath(privmod_t *mod, bool runpath, const char *name,
...
+
+            if (!include_path) {
+                continue;
+            }
+
+            while (*include_path && *include_path == ' ') {
+                ++include_path;
+            }
+
+            /* We have enough room in the string to prefix it with "/etc/". */
```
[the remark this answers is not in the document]
StephanvanSchaik: OK, sounds like I need to check whether it is an absolute path or a relative path. On my system, it seems to be a relative path where it assumes the current directory is `/etc/`.

core/unix/os_private.h
StephanvanSchaik: I think this makes sense since `glob` has a number of options that we do not need and that are not trivial to implement like not setting `GLOB_NOSORT`, in which case we need to call `qsort` (DynamoRIO might already have this and I simply missed it).

In addition, I want to keep the number of libc functions that we have to implement to a minimum, since some of them can be hard to get right and both `fnmatch` and `glob` are already non-trivial but unfortunately necessary to deal with `/etc/ld.so.conf`.

core/globals.h
```diff
@@ -1132,8 +1136,12 @@ int
 tolower(int c);
 int
 strcasecmp(const char *left, const char *right);
+int
+strncasecmp(const char *left, const char *right, size_t n);
 unsigned long
 strtoul(const char *str, char **end, int base);
+int
+fnmatch(const char *pattern, const char *string, int flags);
```
derekbruening, line 1144: It would be good to add some unit tests of the new utilities.  See how the existing ones are done: mostly small functions embedded in some files: e.g., see unit_test_io() in core/io.c inside STANDALONE_UNIT_TEST.

PR #5287, opened 2022-01-19
i#2626: AArch64 v8 decode: Fix missing MSR variant and sysreg decode

joshua-warburton:
This patch adds the missing pstate MSR variant to the decoder
and populates the DR_REG enum with all the non-privileged
system registers so they can be resolved to their symbolic
names

issues: #2626


core/ir/aarch64/encode.c
```diff
@@ -43,45 +43,83 @@
...
-    "b28",         "b29",        "b30", "b31", "nzcv", "fpcr", "fpsr", "tpidr_el0",
-    "tpidrro_el0", "z0",         "z1",  "z2",  "q3",   "z4",   "z5",   "z6",
-    "z7",          "z8",         "z9",  "z10", "z11",  "z12",  "z13",  "z14",
-    "z15",         "z16",        "z17", "z18", "z19",  "z20",  "z21",  "z22",
-    "z23",         "z24",        "z25", "z26", "z27",  "z28",  "z29",  "z30",
-    "z31",         "p0",         "p1",  "p2",  "p3",   "p4",   "p5",   "p6",
-    "p7",          "p8",         "p9",  "p10", "p11",  "p12",  "p13",  "p14",
-    "p15",         "cntvct_el0",
+    "<NULL>", "<invalid>",
+    /* X registers */
```
AssadHashmi, line 49: These comments don't add much to the code.
I would either leave a blank line separator or update with a bit more info, e.g. 64 bit GPR, 128 bit vector, 16 bit FP in vector reg, scalable vector etc.

AssadHashmi (approved):

PR #5294, opened 2022-01-27
i#4134 drbbdup: Use spill slot enum

derekbruening:
Changes hardcoded slot numbers to use the corresponding symbolic enum
value instead.

Re-orders a too-late dead register analysis.

Fixes a typo in the drbbdup docs.

Issue: #4134

### DIFF
--- a/ext/drbbdup/drbbdup.c
+++ b/ext/drbbdup/drbbdup.c
@@ -1,5 +1,5 @@
 /* **********************************************************
- * Copyright (c) 2013-2020 Google, Inc.   All rights reserved.
+ * Copyright (c) 2013-2022 Google, Inc.   All rights reserved.
  * **********************************************************/
 
 /*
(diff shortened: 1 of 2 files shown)


johnfxgalea (approved):

johnfxgalea:
Nit: Maybe change `drbbdup_spill_register()` and `drbbdup_restore_register()` to take type `drbbdup_thread_slots_t`, instead of `int`, for index passed as param.

ext/drbbdup/drbbdup.c
```diff
@@ -712,16 +714,20 @@ drbbdup_encode_runtime_case(void *drcontext, drbbdup_per_thread *pt, void *tag,
      * dispatcher.
      */
     drreg_are_aflags_dead(drcontext, where, &manager->are_flags_dead);
-    if (!manager->is_scratch_reg_dead)
```
derekbruening, line 715: So this was a mistake before causing the scratch to *always* be spilled, right?  There wasn't some subtle reason this was checked before drreg_is_register_dead was called?

johnfxgalea: Nice catch! No reason.

derekbruening:
> Nit: Maybe change `drbbdup_spill_register()` and `drbbdup_restore_register()` to take type `drbbdup_thread_slots_t`, instead of `int`, for index passed as param.

Done.

Issue #4937, opened 2021-06-02
Translation storing interface fails to notify all parties


derekbruening:
Today, if any client or library requests DR_EMIT_STORE_TRANSLATIONS, DR will not use recreation to translate and will instead rely on persisted data from block creation time.  However, all the other clients and libraries may not be aware that one of them requested to store translations, and will be relying on recreation.  It seems that the store-translations approach is broken in the presence of multiple components.

Probably the simplest solution is to add a new callback that DR invokes after the basic block callback when someone requested to store translations.  Then all components can treat the new callback kind of like a recreation, but they would persist storage of data for translation.

We should also try to provide a way to associate persisted data with a callback by a user_data pointer or something: though we'll have to handle each component having its own pointer.  This avoids each party having to keep its own hashtable and worry about lifetime issues with deletion happening independently of a new block at the same tag (xref various issues with that in the past).

derekbruening:
Xref #3801

derekbruening:
Note that storing instead of recreating happens for other reasons:

There are fragments where recreation is impossible: because the original
code is lost or might have changed and we didn't save a copy (modified code
we didn't mark as selfmod (we save a copy for marked-selfmod)) (this is
what -safe_translate_flushed is supposed to solve but it needs work so
today by default we still try to recreate them which is unsafe); because
instrumentation is not repeatable and the client returned
DR_EMIT_STORE_TRANSLATIONS; or the option -store_translations was
set.

PR #5296, opened 2022-01-27
i#2985 scatter-gather: add design doc.

### LINKED ISSUE #2985: Add drx_expand() function for VSIB AVX2/AVX-512 scatter/gather.

abhinav92003:
Adds a design doc with details of the various pieces
involved in adding support for scatter-gather instrs.

Fixes #2985


api/docs/scatter_gather_emulation.dox
```diff
@@ -0,0 +1,230 @@
...
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+/**
+ ****************************************************************************
+\page page_scatter_gather_emulation Emulating x86 Scatter and Gather Instructions
```
derekbruening, line 35: The basic approach applies to AArchXX as well: so maybe remove x86 from the title and up front in the doc say the details are for x86 but we expect the same approach to work on other platforms?  Or leave it in the title and still mention it in the doc.

api/docs/scatter_gather_emulation.dox
```diff
@@ -0,0 +1,230 @@
...
+The x86 gather and scatter instructions were introduced in the AVX2 and AVX512 instruction set extensions. They allow loading or storing a subset of elements in a vector from/to multiple non-contiguous addresses.
+
+AVX2 has only gather instructions, no scatter instructions, whereas AVX512 has both. AVX2 is limited to 256-bit length vectors, whereas AVX512 has 512-bit support. Both support masking of individual memory accesses, using either a special mask register in AVX512 or another vector in AVX2.
+
+Examples of these instructions are (in DR’s IR):
+```
+vpgatherdd %rax(,%ymm11,4)[4byte] %ymm13 -> %ymm12 %ymm13
+```
+
+Above is an AVX2 gather instruction that reads 32-bit doublewords into the 256-bit `ymm12` vector from addresses generated by adding the base address in `rax` to the corresponding index elements in `ymm11`, conditionally based on the masks in `ymm13`. The mask is set to zero on completion, unless the instruction causes an exception.
```
derekbruening, line 50: Not just zero on completion: it zeroes each bit as it carries out the corresponding load/store, and supports interruption with partial effects and seamless restart due to the mask state update.  I think this incremental state update and interruption behavior is important and should be mentioned.

api/docs/scatter_gather_emulation.dox
```diff
@@ -0,0 +1,230 @@
...
+
+```
+vpscatterdd {%k1} %xmm10 -> %rax(,%xmm11,4)[4byte] %k1
+```
+
+Above is an AVX512 scatter instruction that writes 32-bit doublewords from the 128-bit `xmm10` vector to addresses generated by adding the base address in `rax` to the corresponding index element in `xmm11`, conditionally based on the mask reg `k1`. The mask register is set to zero on completion, unless the instruction causes an exception.
```
derekbruening, line 56: `s/reg /register /` (below as well)

api/docs/scatter_gather_emulation.dox
```diff
@@ -0,0 +1,230 @@
...
+
+
+# Problem Statement
+
+Scatter and gather instructions pose a challenge to DynamoRIO clients that observe memory addresses (like drcachesim, the DR client that collects memory address and control flow traces). They are complex to handle because:
```
derekbruening, line 61: I would also give broad examples of tool classes people are familiar with: taint tracking tools; address tracing tools

api/docs/scatter_gather_emulation.dox
```diff
@@ -0,0 +1,230 @@
...
+
+-  A single gather or scatter instruction loads from or stores to multiple addresses
+-  Accessed addresses may be non-contiguous
+-  Each access is conditional based on some mask
+
+DynamoRIO clients only see the scatter or gather instruction and need to do more work to extract all accessed addresses. This is unlike regular scalar loads or stores, where the accessed address is readily available. The goal of this work is to make it easier for DR clients to observe these addresses. We achieve this by expanding the scatter and gather instructions into a functionally equivalent sequence of scalar stores and loads. This way, DR clients will see regular store and load instructions which they can instrument as usual.
```
derekbruening, line 67: It may be worth mentioning that this is similar to the problem of string loop instructions, which DR solves by unrolling: so this approach of expanding/unrolling is established as a good way of solving these many-memrefs-in-one-instruction problems.

api/docs/scatter_gather_emulation.dox
```diff
@@ -0,0 +1,230 @@
...
+# Design
+
+This required the addition of new support in various DynamoRIO components, like drreg, drx, drmgr and core DR. Multiple contributors worked on designing and implementing the required changes.
+
+
+## Scatter/gather Instruction Expansion
+
+Owner: [Hendrik Greving](https://github.com/hgreving2304)
+
+As described above, we can simplify work for DR clients by replacing the scatter and gather instructions with a functionally equivalent sequence of scalar stores and loads. The expanded sequence is the unrolled version of the following loop:
```
derekbruening, line 80: s/the ... instructions/each ... instruction/
(avoid ambiguity over replacing several scatters at once)

api/docs/scatter_gather_emulation.dox
```diff
@@ -0,0 +1,230 @@
...
+for i = 0, 1, ..., (num_accesses-1), do
+  extract mask for the ith access from mask reg or mask vector
+  if mask is set, then
+    extract ith element of index vector
+    compute address = base + ith index element
+    if instr_is_gather, then
+      load data from address into a scalar reg
+      insert scalar data into destination vector
+    else // instr_is_scatter
+      Extract scalar data from source vector to scalar reg
```
derekbruening, line 92: nit: These 2 lines start w/ capital letters; none of the others do.  Pick one style and apply it consistently.

api/docs/scatter_gather_emulation.dox
```diff
@@ -0,0 +1,230 @@
...
+      Store data from scalar reg to address
+    done
+    clear ith mask in mask reg or mask vector
+  done
+done
+```
+
+Due to the x86 ISA, the extraction/insertion of the scalar value from/to the vector may involve multiple steps, e.g. to extract a 32-bit scalar value from a 512-bit `zmm` reg, we first need to extract a 128-bit `xmm` from it.
+
+drmgr in DR provides multiple phases of instrumentation. This expansion is done in the first phase known as app2app. As the name suggests, this phase is intended to transform app instructions to equivalent instructions. For simplicity, we also separate out the scatter or gather instructions from its basic block and create a separate fragment with only the expanded sequence. The logic for expanding scatter and gather instructions is implemented in the drx extension library as [`drx_expand_scatter_gather`](https://github.com/DynamoRIO/dynamorio/blob/eb5d5af8e3444912c9f3f70e5ebf7969252ee4d6/ext/drx/drx.h#L538), and can be used by any client that needs it, including drcachesim. This support was added by [commit](https://github.com/DynamoRIO/dynamorio/commit/4359ef134e47942004c09db04b54593579763186).
```
derekbruening, line 102: s/This/Our/
(avoid confusion over "This expansion" being sthg drmgr does)

(It would be easier to comment on particular sentences if these lines were word wrapped instead of one giant 500 char line)

api/docs/scatter_gather_emulation.dox
```diff
@@ -0,0 +1,230 @@
...
+
+As an example, the following are the expansions of some instructions.
+
+Expansion for `vpgatherdd`: https://gist.github.com/abhinav92003/637c07bac785428550f05c18319f80c2
```
derekbruening, line 106: Seems better to inline this here and be self-contained than rely on external content when it's just text

api/docs/scatter_gather_emulation.dox
```diff
@@ -0,0 +1,230 @@
...
+
+Expansion for `vpscatterdd`: https://gist.github.com/abhinav92003/4a78ec4ad1621cbf6ca905a6774cb389
+
+We also need to make sure that if some load or store faults, we restore the application state properly. While the spilled registers obtained from drreg are restored by the drreg state restoration logic, drx still needs to restore the scratch mask registers that are spilled manually to a GPR and ensure that the bit for the faulting access is cleared too. In case a fault or some other state restore event like DR detach happens, we walk the expanded sequence using a state machine till we reach the fault pc, keeping track of the state that needs to be restored ([commit](https://github.com/DynamoRIO/dynamorio/commit/de511a72197e9b04a9732f33622fe7bc7fd12623), [commit](https://github.com/DynamoRIO/dynamorio/commit/e54913b4845794b96e8682ba6a214d54e10741a0)).
```
derekbruening, line 110: The mask bits for the completed memrefs have to be cleared, but not the faulting one?

api/docs/scatter_gather_emulation.dox
```diff
@@ -0,0 +1,230 @@
...
+
+
+## Drreg Support For Multi-phase Reservations
+
+Owner: [Abhinav Sharma](https://github.com/abhinav92003)
+
+Upstream issue: [DynamoRIO/dynamorio#3823](https://github.com/DynamoRIO/dynamorio/issues/3823)
+
+Drreg is DynamoRIO’s register reservation framework. It allows users to reserve a reg to use as scratch. Internally, drreg automatically performs the following functions so that the user does not need to. Drreg
```
derekbruening, line 119: As noted above s/reg /register /

api/docs/scatter_gather_emulation.dox
```diff
@@ -0,0 +1,230 @@
...
+- keeps all required book-keeping like the spill slot to spilled register mapping
+- restores spilled registers to their application value before they are read by an application instruction; also, it re-spills the spilled registers if they are written by an application instruction.
+- performs application state restoration on state restore events like encountering an application fault, and DR detach.
+
+While expanding a scatter or gather instruction in the app2app phase, we need a scratch register to hold the scalar values and masks. In later phases (like the insertion or the instru2instru phase), drcachesim and other DR clients may also use drreg to get scratch registers for their instrumentation.
+
+Drreg initially supported only single phase use and did not keep any state between phases. To support multi-phase use of drreg, we need to solve the following:
```
derekbruening, line 126: s/need/needed/

api/docs/scatter_gather_emulation.dox
```diff
@@ -0,0 +1,230 @@
...
+- avoid spill slot conflict across multiple phases: multi-phase use can potentially lead to spill slot conflicts if the same slot is selected in multiple phases. This may clobber the spilled application value and cause the application to crash or otherwise fail.
+- allow aflags spill to any slot: drreg hardcodes the aflags spill slot as the zero-th slot, to simplify some logic. To support the ability to spill aflags in multiple phases, drreg should be able to use any spill slot for aflags.
+- application state restore logic: on a state restore event, we should be able to figure out which slot contains each spilled register's app value. This is complicated by the fact that registers may be spilled by instrumentation added by multiple phases, and the spill regions may overlap which causes the spilled application value to be moved between spill slots.
+
+We explored the following ideas to avoid spill slot conflicts in drreg:
+
+#### Disjoint slot spaces or arenas
```
derekbruening, line 133: The sections jump from ## to ####.  Best to add a ### in between, parallel to the ### below.

api/docs/scatter_gather_emulation.dox
```diff
@@ -0,0 +1,230 @@
...
+As an example, the following are the expansions of some instructions.
+
+Expansion for `vpgatherdd`: https://gist.github.com/abhinav92003/637c07bac785428550f05c18319f80c2
+
+Expansion for `vpscatterdd`: https://gist.github.com/abhinav92003/4a78ec4ad1621cbf6ca905a6774cb389
+
+We also need to make sure that if some load or store faults, we restore the application state properly. While the spilled registers obtained from drreg are restored by the drreg state restoration logic, drx still needs to restore the scratch mask registers that are spilled manually to a GPR and ensure that the bit for the faulting access is cleared too. In case a fault or some other state restore event like DR detach happens, we walk the expanded sequence using a state machine till we reach the fault pc, keeping track of the state that needs to be restored ([commit](https://github.com/DynamoRIO/dynamorio/commit/de511a72197e9b04a9732f33622fe7bc7fd12623), [commit](https://github.com/DynamoRIO/dynamorio/commit/e54913b4845794b96e8682ba6a214d54e10741a0)).
+
+
+## Drreg Support For Multi-phase Reservations
```
derekbruening, line 113: On first reading, the transition to this section is abrupt and it seems like it doesn't belong in this document.  Some context right up front in this section, and maybe at the end of the prior section too to transition, and maybe also in this section title, would all help: why does this matter for scatter expansion?

api/docs/scatter_gather_emulation.dox
derekbruening, line 126: Make it more clear that the single phase supported was insertion phase (o/w it sounds like it could be any of them).  There was non-insertion-phase support (the key functions worked), there was just no explicit support for bad interactions between them; the non-insertion phase used was assumed to be instru2instru or sthg where the client had to be careful.

api/docs/scatter_gather_emulation.dox
```diff
@@ -0,0 +1,230 @@
...
+- application state restore logic: on a state restore event, we should be able to figure out which slot contains each spilled register's app value. This is complicated by the fact that registers may be spilled by instrumentation added by multiple phases, and the spill regions may overlap which causes the spilled application value to be moved between spill slots.
+
+We explored the following ideas to avoid spill slot conflicts in drreg:
+
+#### Disjoint slot spaces or arenas
+We can ask drreg to create slot spaces or arenas at init time, which are assigned disjoint spill slots. When reserving a register, the user passes in a “space/arena Id '' to instruct drreg to pick free slots only from that arena. This requires keeping some global drreg state. This also requires the user to guess the best configuration for assigning slots to the arenas, and passing the correct arena Id before each reservation. It may artifically make some spill slots unavailable for use, thereby reducing efficiency.
```
derekbruening, line 134: Are those quotes right: ` “space/arena Id '' `

derekbruening, line 134: spelling: "artifically"

api/docs/scatter_gather_emulation.dox
```diff
@@ -0,0 +1,230 @@
...
+
+### State Restoration For Drreg
+
+Owner: [Abhinav Sharma](https://github.com/abhinav92003)
+
+Upstream issue: [DynamoRIO/dynamorio#3823](https://github.com/DynamoRIO/dynamorio/issues/3823), [DynamoRIO/dynamorio#3801](https://github.com/DynamoRIO/dynamorio/issues/3801)
+
+On a state restore event, drreg should be able to restore all spilled registers to their application values.
+
+Unfortunately, when a state restore event happens, we only have the encoded fragment, and none of the drreg state, like the reg to spill slot mappings. We need to reconstruct this state based on the faulting pc and the encoded fragment.
```
derekbruening, line 153: We could store state (have to do this anyway for fragments where recreation is not possible)

api/docs/scatter_gather_emulation.dox
```diff
@@ -0,0 +1,230 @@
...
+
+
+It is complex to determine which registers need to be restored and from which spill slot. This is because drreg automatically adds spill and restore instructions to handle various complex cases like automatic re-spilling of reserved registers after their application write instruction, and automatic restore of reserved registers before their application read instruction. Drreg also uses various optimisations like lazy restores for application value in case the register is reserved again. This is even more complex for aflags, for which spill and restore are a two-step process (spilling aflags involves reading aflags into a register using `lahf` and then writing that register to a spill slot; restoring aflags involves reading aflags from its spill slot to a register, and then writing aflags from that register using `sahf`). In some cases, aflags are even kept in a register as an optimisation.
```
derekbruening, line 156: s/value/values/

derekbruening, line 156: 3-step process to get OF

api/docs/scatter_gather_emulation.dox
```diff
@@ -0,0 +1,230 @@
...
+
+Additionally, in multi-phase use, a register may be spilled by multiple phases, with a separate spill slot for each phase. The application value for the register may reside in one or more spill slots, and may also move between spill slots based on how the spill regions from different phases overlap. See various tricky scenarios in [drreg-test.c](https://github.com/DynamoRIO/dynamorio/blob/f1d496b451eaa6e9aaff7125617030164c6cfdff/suite/tests/client-interface/drreg-test.c#L637).
+
+We explored two ways to adapt drreg’s state restoration logic to multi-phase use. This also fixed some known existing issues with drreg: [Dynamorio/dynamorio#4933](https://github.com/DynamoRIO/dynamorio/issues/4933), [DynamoRIO/dynamorio#4939](https://github.com/DynamoRIO/dynamorio/issues/4939).
```
derekbruening, line 160: (Not requesting a change: just noting that I wouldn't normally expect to see references to past now-fixed bugs in a design doc.)

api/docs/scatter_gather_emulation.dox
```diff
@@ -0,0 +1,230 @@
...
+
+#### Track app values as they are moved between slots and regs
```
derekbruening, line 162: registers

api/docs/scatter_gather_emulation.dox
```diff
@@ -0,0 +1,230 @@
...
+
+#### Preferred: Pairing restores with spills (instead of the other way)
+
+The key observation behind this approach is that it is easier to find the matching spill for a given restore, than to find the matching restore for a given spill. This is because there may be other restores besides the final restore, e.g. restores for app read, user prompted restores, etc. This makes it hard to find exactly where the spill region for a register/aflags ends.  Additional complexities include the fact that aflags re-spills may not use the same slot, which makes differentiating spills from multiple phases difficult.
+
+Each restore must have a matching spill. Based on this observation, we scan the faulting fragment from end to beginning, matching register restores to their spills. When we reach the faulting instruction, any restore for which we did not see the matching spill yet must be performed by the drreg state restoration. This was implemented by ([commit](https://github.com/DynamoRIO/dynamorio/commit/be78c124eb787601182e3c73f4be8bb859c50ef8)).
```
derekbruening, line 178: What about a register with two spills (the 2nd is after an app write) and one restore, with the fault between the 2 spills?  Won't this approach incorrectly fail to restore the register?

derekbruening (approved):

ext/drx/drx.c
```diff
@@ -2370,6 +2370,9 @@ expand_gather_load_scalar_value(void *drcontext, instrlist_t *bb, instr_t *sg_in
  * Finally, clear the entire mask register, even
  * the parts that are not used as a mask:
  *   kxorq           %k1 %k1 -> %k1
+ *
+ * For more design details see the following wiki
```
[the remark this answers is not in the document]
abhinav92003: Done

api/docs/scatter_gather_emulation.dox
abhinav92003: Done

api/docs/scatter_gather_emulation.dox
abhinav92003: Done

api/docs/scatter_gather_emulation.dox
abhinav92003: Done.

abhinav92003: Ah. Copying from docs made it weird. Fixed.

api/docs/scatter_gather_emulation.dox
abhinav92003: I used "####" because I didn't want this showing up as a huge heading.

api/docs/scatter_gather_emulation.dox
abhinav92003: Done

api/docs/scatter_gather_emulation.dox
abhinav92003: Done

api/docs/scatter_gather_emulation.dox
abhinav92003: Done

api/docs/scatter_gather_emulation.dox
abhinav92003: Done

api/docs/scatter_gather_emulation.dox
abhinav92003: Done

api/docs/scatter_gather_emulation.dox
abhinav92003: Done

api/docs/scatter_gather_emulation.dox
abhinav92003: Kept in the title, but mentioned in the doc that we expect it to work on other platforms, like AArch64 SVE instructions.

api/docs/scatter_gather_emulation.dox
```diff
@@ -0,0 +1,230 @@
...
+The scatter and gather expansions requires a scratch `xmm` register, for which we need
+the capability to spill and restore vector registers. Following are the design
+choices:
+
+- Extend drreg to support reservation for vector registers.
+  [DynamoRIO/dynamorio#3844](https://github.com/DynamoRIO/dynamorio/issues/3844)
+  aims to add this support.
+
+- Use custom spill and restore logic in drx. We can do this by reserving memory in
+TLS to use as spill slot.
```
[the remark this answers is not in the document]
abhinav92003: Done

api/docs/scatter_gather_emulation.dox
```diff
@@ -0,0 +1,230 @@
...
+
+
+Some observations about this use-case for vector reservation:
+- We need to spill only one vector register, so we do not need sophisticated spill slot management logic.
+- The spilled vector register will not need to be restored for app reads, or re-spilled after app writes. Note that we will not encounter any application instructions that use the spilled vector register, because it needs to be spilled only for the duration of the expanded scatter or gather sequence.
+
+Extending drreg to support vector spilling is a complex task. Given the above observations, the current use case does not justify the effort. Therefore, we choose to implement custom spill logic in drx ([commit](https://github.com/DynamoRIO/dynamorio/commit/88cba2817fef7a4d4ba3e8c2375784ea4165c133), [commit](https://github.com/DynamoRIO/dynamorio/commit/84bf9288a12e8e38abab04d4d8273cc3226fa13c)).
```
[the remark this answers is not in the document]
abhinav92003: Chose is better; in line with other similar statements "We implemented", "This was implemented", "We dropped this approach".

api/docs/scatter_gather_emulation.dox
abhinav92003: Done.

api/docs/scatter_gather_emulation.dox
abhinav92003: Done.

Also split these lines so that they are not as long. Should have done it before to make review easier. I'm afraid there will be a large diff due to this fix.

derekbruening:
I would still add a ### to keep a strict hierarchy. What does the contents
show without it?

derekbruening:
The sidebar menu I mean: not sure what the JavaScript does with missing
levels.

api/docs/scatter_gather_emulation.dox
abhinav92003: Done

api/docs/scatter_gather_emulation.dox
abhinav92003: Done

api/docs/scatter_gather_emulation.dox
abhinav92003: That's right; the approach will not restore that reg. However, as the reg is dead before the app write, it does not affect app correctness, unless the app deliberately inspects that reg I guess.

api/docs/scatter_gather_emulation.dox
abhinav92003: Probably -- though would we store it for all fragments? That would increase memory usage a lot I think.

Not adding it to this doc though, unless that was what you were suggesting.

api/docs/scatter_gather_emulation.dox
abhinav92003: Done

api/docs/scatter_gather_emulation.dox
derekbruening: Prior replies were on phone to email so showed up below; for completeness duplicating here:

I would still add a ### up above to keep a strict hierarchy and not have all this content in the "intro" for the ## instead of a full-fledged ### parallel to the one below.  What does the contents show without it, especially the sidebar menu: not sure what the JavaScript does with missing levels.  Please make sure the menus are sane.

api/docs/scatter_gather_emulation.dox
abhinav92003: Based on this comment
https://github.com/DynamoRIO/dynamorio/blob/faf502728968e8d003b912b6771b3df5ed9bc7aa/ext/drx/drx.c#L3701
I believe the faulting access's bit needs to be zeroed.

derekbruening: But how do you re-execute the instruction then?  You would want to re-try the not-completed previously-faulted access, but if the bit is already zeroed it won't even try it on the re-execution.

api/docs/scatter_gather_emulation.dox
abhinav92003: Done

api/docs/scatter_gather_emulation.dox
abhinav92003: yeah, I'm confused by this too. I tried this out natively and I don't see the faulting mask bit being zeroed. I'm checking to see if I mis-understood that comment.

api/docs/scatter_gather_emulation.dox
abhinav92003: I removed the markdown heading and just made that text bold. It's better to not have this in the sidebar menu.

api/docs/scatter_gather_emulation.dox
abhinav92003: If I recall correctly, this is not a limitation of the multi-phase drreg design, but only of the implementation. The current restore logic without ilist (`drreg_event_restore_state_without_ilist`) is somewhat stale (doesn't handle multi-phase reservations), but it can be made to be very similar to `drreg_event_restore_state_with_ilist`. The only thing we'll be  missing out on is detection of false spills and restore, which is a general limitation, not related to the scatter/gather sequence.

I have created #5303 for this. Given that it is a bug that we intend to fix (I'll try to find some time soon for this), maybe it's not worth adding it to the design limitation?

api/docs/scatter_gather_emulation.dox
abhinav92003: Still looking into this. Pushed the other changes for review for now.

abhinav92003:
Sorry for the large diff. That's because I split up lines to avoid too-long lines; should definitely have done this before the last review.

I'm still looking into the "mask bit for the faulting access" thread. Pushed the remaining changes for review for now.

api/docs/scatter_gather_emulation.dox
derekbruening: Agreed, the without_ilist support is fairly good, so typical translation would probably work for drreg after all.  Also xref #4937 for general stored-xl8 support.  OK.

api/docs/scatter_gather_emulation.dox
derekbruening: If drreg_options_t.conservative is turned on, drreg will then do the wrong thing here.  This should probably be documented in drreg.h, and mentioned here.  (Not considered likely to affect most apps of course.)

api/docs/scatter_gather_emulation.dox
abhinav92003: Ah, figured this out. The code at the above comment and other similar comments in this file (e.g. the one at https://github.com/DynamoRIO/dynamorio/blob/faf502728968e8d003b912b6771b3df5ed9bc7aa/ext/drx/drx.c#L3174) applies only when there's a fault between L180 and L187 in the doc I just added (non-inclusive). This is when the load completed but there was a fault before we could clear the mask.

I don't think this can happen in any instruction in the expanded sequence itself (there's no memory instruction there), but maybe some other client's instrumentation or other async state restore event may trigger this path.

I'll fix the documentation to clarify this.

abhinav92003: Done.

api/docs/scatter_gather_emulation.dox
```diff
@@ -0,0 +1,645 @@
...
+-  Each access is conditional based on some mask
+
+DynamoRIO clients only see the scatter or gather instruction and need to do more
+work to extract all accessed addresses. This is unlike regular scalar loads or
+stores, where the accessed address is readily available. The goal of this work is
+to make it easier for DR clients to observe these addresses. We achieve this by
+expanding the scatter and gather instructions into a functionally equivalent
+sequence of scalar stores and loads. This way, DR clients will see regular store
+and load instructions which they can instrument as usual. This is similar to what
+DR does for repeat string operations (like `rep`, `repnz`): convert it into a
```
derekbruening, line 94: A complete opcode like `rep movs` seems clearer than just the prefix.

api/docs/scatter_gather_emulation.dox
```diff
@@ -0,0 +1,645 @@
...
+registers for the expansion. The GPR scratch registers are obtained using
+drreg, whereas the scratch `zmm` register and the scratch mask register are
+obtained by manually spilling them.
+
+We need to make sure that if some load or store faults, we restore the
+application state properly. While the spilled registers obtained from drreg are
+restored by the drreg state restoration logic, drx still needs to restore the
+scratch mask register that is spilled manually to a GPR, and the scratch `zmm` register
+that is spilled manually to a drx spill slot. We also need to ensure that the bit
+for the faulting access is cleared too. In case a fault or some other state restore
```
derekbruening, line 383: This still looks incorrect

api/docs/scatter_gather_emulation.dox
```diff
@@ -0,0 +1,645 @@
...
+Each restore must have a matching spill. Based on this observation, we scan the
+faulting fragment from end to beginning, matching register restores to their
+spills. When we reach the faulting instruction, any restore for which we did not
+see the matching spill yet must be performed by the drreg state restoration. This
+was implemented by
+([commit](https://github.com/DynamoRIO/dynamorio/commit/be78c124eb787601182e3c73f4be8bb859c50ef8)).
+
+This algorithm does not need to be aware of non-drreg methods of spilling/restoring
+registers. Note that, like other drreg design, this method does not restore a
+spilled register's application value if it is dead at the faulting instruction.
```
derekbruening, line 564: This is only true if `drreg_options_t.conservative` is set to false.  If it is true, this approach is then incorrect.  This should be documented in drreg.h and updated here.

api/docs/scatter_gather_emulation.dox
abhinav92003: This seems very hard to do without additional metadata. Added a comment to #3801, changed the doc and added the comment to drreg.h too. Sad to see an unhandled case (even though less likely to matter) in drreg state restoration. I'm finding it very tricky to get all corner cases working perfectly without additional  metadata.

api/docs/scatter_gather_emulation.dox
abhinav92003: Pushed that code just now.

api/docs/scatter_gather_emulation.dox
abhinav92003: The latest push fixed it.

api/docs/scatter_gather_emulation.dox
abhinav92003: Fixed.

api/docs/scatter_gather_emulation.dox
abhinav92003: By mistake, I used `git commit --amend` instead of `git commit`, without any only-local commit. That caused a mismatch with the origin branch. Had to manually put the new changes back.

abhinav92003:
Pushed the latest code. PTAL.

derekbruening (approved):

Issue #5302, opened 2022-01-28
ASSERT opnd_disassemble: invalid opnd type on rel-addr opnd on aarch32


derekbruening:
Running a drbbdup test that creates a rel-addr opnd on aarch32:
```
SYSLOG_ERROR: Application suite/tests/bin/simple_app (2316) DynamoRIO usage error : opnd_disassemble: invalid opnd type
```

It looks like internal_opnd_disassemble() incorrectly has REL_ADDR_kind handling under ifdef X64 instead of X64 || ARM.

Xref #5295

PR #5305, opened 2022-01-28
i#5304: Ignore reachability for instr_is_encoding_possible()

### LINKED ISSUE #5304: instr_is_encoding_possible fails on unreachable pc-relative operands but takes no target
While adding a test for #5302 on arm, I hit this:
```
ASSERT FAILURE: /home/derek/dr/git/src/suite/tests/api/ir_arm.c:80: instr_is_encoding_possible(instr)
```

This is for the instruction in test_pcrel() in ir_arm.c.

The problem is that  instr_is_encoding_possible() does not take in a target location for encoding, so it can't know whether pc-relative operands will reach.  It has to relax those reachability checks.

derekbruening:
On ARM, instr_is_encoding_possible() was checking reachabilty, yet it
took no target encode address, causing it to fail for any but very
small addresses.  We remove the reachability checks to solve this.

The test for #5302 in a forthcoming PR adds a regression test for
this.

Fixes #5304

### DIFF
--- a/core/ir/encode_shared.c
+++ b/core/ir/encode_shared.c
@@ -1,5 +1,5 @@
 /* **********************************************************
- * Copyright (c) 2011-2021 Google, Inc.  All rights reserved.
+ * Copyright (c) 2011-2022 Google, Inc.  All rights reserved.
  * Copyright (c) 2001-2010 VMware, Inc.  All rights reserved.
  * **********************************************************/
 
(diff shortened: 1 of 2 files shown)


abhinav92003 (approved):

core/ir/encode_shared.c
```diff
@@ -99,6 +99,7 @@ get_encoding_info(instr_t *instr)
     const instr_info_t *info = instr_get_instr_info(instr);
     decode_info_t di;
     decode_info_init_for_instr(&di, instr);
+    IF_AARCHXX(di.check_reachable = false;)
```
abhinav92003, line 102: From the PR description it looks like this PR is only for ARM, but here it's changing for AArch64 too.

abhinav92003: Commenting as a review comment to make it visible on front PR page.

abhinav92003 (approved):

core/ir/encode_shared.c
derekbruening: Changing the description in the final merge.

PR #5306, opened 2022-01-28
i#5302: Fix a32 assert on rel-addr opnd disassembly

### LINKED ISSUE #5302: ASSERT opnd_disassemble: invalid opnd type on rel-addr opnd on aarch32

derekbruening:
Adds correct ifdef ARM handling of REL_ADDR_kind operands when
disassembling, and in legacy loadtoconst code.

Adds a test to api.ir (confirmed it fails without this fix).

Fixes other latent issues in api.ir on arm which unfortunately does
not run on QEMU yet and so is not run automatically in today's CI.
Tested manually on an arm system.

Fixes #5302


suite/tests/api/ir_arm.c
```diff
@@ -81,7 +81,10 @@ test_instr_encoding(void *dc, uint opcode, instr_t *instr)
     pc = instr_encode(dc, instr, buf);
```
abhinav92003, line 81: Assert pc != NULL? It was checked in the old version of `test_pcrel`.

suite/tests/api/ir_arm.c
```diff
@@ -111,28 +114,26 @@ test_pred(void *dc)
...
 test_pcrel(void *dc)
 {
-    byte *pc;
     instr_t *inst;
     inst = INSTR_CREATE_ldr(dc, opnd_create_reg(DR_REG_R0),
                             opnd_create_rel_addr((void *)&buf[128], OPSZ_PTR));
-    pc = instr_encode(dc, inst, buf);
-    /* (gdb) x/2i buf+1
-     *    0x120b9 <buf+1>:     ldr     r0, [pc, #124]  ; (0x12138 <buf+128>)
+    /* On decoding our rel-addr opnd turns into a base-disp:
```
abhinav92003, line 132: Is it worth constructing an instruction with the expected base-disp operand and checking equality of `inst`  with that?

abhinav92003 (approved):

suite/tests/api/ir_arm.c
derekbruening: Done.

suite/tests/api/ir_arm.c
derekbruening: Duplicating the code from test_instr_encoding?  Maybe it's worth it to avoid the bool param.  OK, done.

abhinav92003: Not only to avoid the bool param. But also for a stronger test.

PR #5307, opened 2022-01-29
Only check for AVX instructions on x86

davidben:
On Arm, the configure process outputs warnings for not finding AVX support, but that's to be expected on Arm.

-- WARNING: Compiler or processor do not support AVX. Skipping tests
-- WARNING: Compiler or processor do not support AVX2. Skipping tests
-- WARNING: Compiler or processor do not support AVX-512. Skipping tests

Only probe for AVX on x86 platforms, which avoids the warnings elsewhere.


derekbruening:
add to allowlist

derekbruening:
run arm tests

derekbruening (approved):

derekbruening:
run arm tests

derekbruening:
@AssadHashmi I think I did the trigger correctly but the Jenkins run never started?

PR #5308, opened 2022-01-29
i#2626: add Armv8.2 SHA-512 instructions to the decoder

davidben:
I mostly did this by mimicking the other SHA instructions, so I'm not sure if I did this right. Test cases constructed by running them through `as` and then `objdump -D`.

(Context is that we were integrating some assembly for this instruction into BoringSSL, and I noticed neither valgrind nor drmemory supported the instruction. Looks like drmemory doesn't work yet on aarch64 anyway and I'm not sure what the consequences of `OP_xx` are in the first place. But I figure transcribing the instruction encoding was easy enough to do.)

References:
https://developer.arm.com/documentation/ddi0602/2021-12/SIMD-FP-Instructions/SHA512H--SHA512-Hash-update-part-1-
https://developer.arm.com/documentation/ddi0602/2021-12/SIMD-FP-Instructions/SHA512H2--SHA512-Hash-update-part-2-
https://developer.arm.com/documentation/ddi0602/2021-12/SIMD-FP-Instructions/SHA512SU0--SHA512-Schedule-Update-0-
https://developer.arm.com/documentation/ddi0602/2021-12/SIMD-FP-Instructions/SHA512SU1--SHA512-Schedule-Update-1-

Issue: #2626

### DIFF
--- a/core/ir/aarch64/codec.txt
+++ b/core/ir/aarch64/codec.txt
@@ -1161,10 +1161,10 @@ x0011010110xxxxx000011xxxxxxxxxx  n   363        sdiv            wx0 : wx5 wx16
 01011110000xxxxx010100xxxxxxxxxx  n   374    sha256h2             q0 : q5 q16 s_const_sz
 0101111000101000001010xxxxxxxxxx  n   375   sha256su0             q0 : q5 s_const_sz
 01011110000xxxxx011000xxxxxxxxxx  n   376   sha256su1             q0 : q5 q16 s_const_sz
-11001110011xxxxx100000xxxxxxxxxx  n   579     sha512h             q0 : q5 q16 d_const_sz
-11001110011xxxxx100001xxxxxxxxxx  n   580    sha512h2             q0 : q5 q16 d_const_sz
-1100111011000000100000xxxxxxxxxx  n   581   sha512su0             q0 : q5 d_const_sz
-11001110011xxxxx100010xxxxxxxxxx  n   582   sha512su1             q0 : q5 q16 d_const_sz
+11001110011xxxxx100000xxxxxxxxxx  n   579     sha512h             q0 : q5 q16 d_const_sz # v8.2
+11001110011xxxxx100001xxxxxxxxxx  n   580    sha512h2             q0 : q5 q16 d_const_sz # v8.2
+1100111011000000100000xxxxxxxxxx  n   581   sha512su0             q0 : q5 d_const_sz # v8.2
+11001110011xxxxx100010xxxxxxxxxx  n   582   sha512su1             q0 : q5 q16 d_const_sz # v8.2
 0x001110xx1xxxxx000001xxxxxxxxxx  n   377       shadd            dq0 : dq5 dq16 bhs_sz
 0101111101xxxxxx010101xxxxxxxxxx  n   378         shl             d0 : d5 immhb_0shf
 0x0011110xxxxxxx010101xxxxxxxxxx  n   378         shl            dq0 : dq5 bhsd_immh_sz immhb_0shf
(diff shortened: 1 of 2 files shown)


derekbruening:

@AssadHashmi hoping you can take a look, or triage further.

davidben:
> Please add the comment string # v8.2 after each new defintion in codec.txt before merging.

Done.

> There are upcoming changes to prepare the codec for new versions of the AArch64 ISA and such strings will help when splitting up defintions by version.

Oh! Is this going to figure into something like drcpusim? Would it be possible to teach the decode which instructions are gated on FEAT_WHATEVER? I see there are functions like `instr_is_sse3` for x86. On x86, we (BoringSSL) use Intel SDE to verify cpuid checks (i.e. simulate non-existence of an instruction) and also emulate instructions not on our CI's host machines. If we could do something similar for Arm, that'd be really useful.

AssadHashmi:
> > There are upcoming changes to prepare the codec for new versions of the AArch64 ISA and such strings will help when splitting up defintions by version.
> 
> Oh! Is this going to figure into something like drcpusim? Would it be possible to teach the decode which instructions are gated on FEAT_WHATEVER? I see there are functions like `instr_is_sse3` for x86. On x86, we (BoringSSL) use Intel SDE to verify cpuid checks (i.e. simulate non-existence of an instruction) 

Yes, eventually. The first step is to split up `codec.txt` and get the codec generator to create cleanly separated decode/encode files based on v8.x (`codec_v80.txt`, `codec_v81.txt`, `codec_v82.txt`), v9.x etc. Later patches will read the `ID_AA64ISAR0_EL1` register at startup establishing which `FEAT_`s are supported on the h/w.

AssadHashmi:
> On x86, we (BoringSSL) use Intel SDE to verify cpuid checks (i.e. simulate non-existence of an instruction) and also emulate instructions not on our CI's host machines. If we could do something similar for Arm, that'd be really useful.

Sorry, I think you meant that BoringSSL emulates instructions not on your CI's host machines didn't you? After DynamoRIO tells you which instructions are not supported on the h/w.

davidben:
> Sorry, I think you meant that BoringSSL emulates instructions not on your CI's host machines didn't you? After DynamoRIO tells you which instructions are not supported on the h/w.

No, we don't emulate instructions in our tooling. The situation is we have a lot of code that uses such-and-such new instruction, and gates it on a CPU capability check. We would like to test two things:

1. We correctly gated the right functions on the right CPU capability checks. If we test on, say, a machine with AVX, we won't notice if we forgot to gate the AVX code on the AVX bit in CPUID.
2. The code using the new instructions is correct. If we test on, say, a machine without AVX, we won't notice if the AVX code has a bug.

Intel SDE solves both those problems for us. We can tell it what CPU to mimic and it'll both simulate a lack of instructions for (1) and it'll emulate instructions newer than the host hardware for (2). We require both capabilities to adequately test.

As I understand it, drcpusim, if it worked on Arm, would cover (1) but not (2). That would already be a big improvement, although I'm hoping DynamoRIO could be used to also address (2), perhaps in drcpusim itself. (Emulating NEON is probably too large of a target, but things like FEAT_SHA512 should be doable.)

derekbruening:
> No, we don't emulate instructions in our tooling. The situation is we have a lot of code that uses such-and-such new instruction, and gates it on a CPU capability check. We would like to test two things:
> 
> 1. We correctly gated the right functions on the right CPU capability checks. If we test on, say, a machine with AVX, we won't notice if we forgot to gate the AVX code on the AVX bit in CPUID.
> 2. The code using the new instructions is correct. If we test on, say, a machine without AVX, we won't notice if the AVX code has a bug.
> 
> Intel SDE solves both those problems for us. We can tell it what CPU to mimic and it'll both simulate a lack of instructions for (1) and it'll emulate instructions newer than the host hardware for (2). We require both capabilities to adequately test.
> 
> As I understand it, drcpusim, if it worked on Arm, would cover (1) but not (2). That would already be a big improvement, although I'm hoping DynamoRIO could be used to also address (2), perhaps in drcpusim itself. (Emulating NEON is probably too large of a target, but things like FEAT_SHA512 should be doable.)

I filed #5311 on adding AArch64 support to the existing drcpusim, and #5312 on your suggestion of extending drcpusim to emulate missing features.

davidben:
Looks like this PR slipped through, but then it was done in https://github.com/DynamoRIO/dynamorio/pull/5589 some 8 months later, so we can probably close this now.

PR #5310, opened 2022-01-31
i#1698 ldstex: Document core DR memop insertion

derekbruening:
Documents several more cases where core DR, even without any tool,
inserts memory operations between an application's load-exclusive and
store-exclusive: far direct links; trace building; indirect branches.

Issue: #1698

### DIFF
--- a/api/docs/ldstex.dox
+++ b/api/docs/ldstex.dox
@@ -106,17 +106,26 @@ discussed below.
 
 For sequences where the load-exclusive is in a different block than the corresponding
 store-exclusive, there are multiple scenarios where DynamoRIO can insert loads and stores
-in between the two.  First, for linking two blocks together, if the blocks are too far
-apart for a single branch instruction to reach, DynamoRIO jumps through an exit stub
-which contains loads and stores (see \ref page_aarch64_far).  Second, for building
-traces, DynamoRIO unlinks each block before execution while discovering the sequence of
-blocks to use for the new traces.  It assumes that unlinking does not perturb the
-application.  However, the unlink path goes through a stub with memory operations and
-then through DynamoRIO system code with a large number of loads and stores.  Third,
-indirect branches load entries from a hashtable, so an indirect branch between a
-load-exclusive and store-exclusive, while perhaps unlikely to occur in most code, is
-another case where even without a tool adding additional instruction the core DynamoRIO
-engine has a potential problem.
+in between the two.  First, an unlinked block will execute many loads and stores on its
+path into DynamoRIO and inside DynamoRIO system code.  Thus, on first execution of a pair
+of split load-exclusive and store-exclusive blocks, the store-exclusive will likely fail
+due to all of the operations in between while building its block.  Even once the two
+blocks are linked, though, if the blocks are too far apart for a single branch
+instruction to reach, DynamoRIO jumps through an exit stub which contains loads and
+stores in order to reach the target block (see \ref page_aarch64_far).
+
+The operations after an unlinked block also affect trace building.  During trace
+discovery, DynamoRIO unlinks each block before execution to identify the sequence of
+blocks to use for the new trace.  It assumes that unlinking does not perturb the
+application.  However, the unlink path's memory operations are likely to cause monitor
+failure, leading the trace stopping after the store-exclusive block and not capturing the
+hot path in the primary trace.  Secondary traces on the exits should still include the
+proper code in the final set of traces, but with sub-optimal layout.
+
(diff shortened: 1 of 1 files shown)


api/docs/ldstex.dox
```diff
@@ -89,7 +89,34 @@ The application must be prepared for monitor failure (which can also occur due t
...
+tool instrumentation and stolen register handling can in some cases plan ahead and insert
+all added memory references before or after the entire sequence (with complications where
+tool instrumentation wants to mirror execution flow, but if there is a fault or
+predicated execution it can be challenging to instrument too far away from the target
+instruction).  However, these sequences frequently include branches and cross blocks, as
+discussed below.
+
+For sequences where the load-exclusive is in a different block than the corresponding
+store-exclusive, there are multiple scenarios where DynamoRIO can insert loads and stores
+in between the two.  First, for linking two blocks together, if the blocks are too far
```
abhinav92003, line 109: Even unlinked fragments have a memory read, to load the fcache return routine's address from the memory. Though in that case, there would be many other memory accesses too (from fcache enter and return).

Is it worth mentioning that such unlinked fragments with a direct branch will probably be linked after the first run, therefore will fail just once?

api/docs/ldstex.dox
```diff
@@ -89,7 +89,34 @@ The application must be prepared for monitor failure (which can also occur due t
...
+apart for a single branch instruction to reach, DynamoRIO jumps through an exit stub
+which contains loads and stores (see \ref page_aarch64_far).  Second, for building
+traces, DynamoRIO unlinks each block before execution while discovering the sequence of
+blocks to use for the new traces.  It assumes that unlinking does not perturb the
+application.  However, the unlink path goes through a stub with memory operations and
```
abhinav92003, line 114: Unlinking would cause only the current iteration to fail -- does it mitigate this 2nd issue at all?

abhinav92003 (approved):

api/docs/ldstex.dox
derekbruening: Will work it into revised trace discussion.

api/docs/ldstex.dox
derekbruening: The current iteration is what is enshrined in the trace, so a primary trace will never capture the monitor success path.  This means the trace will always stop at the stex.  Exits from primary traces are marked as trace heads for making secondary traces so the success path should still become part of the set of traces, but with sub-optimal layout.  Rewrote to reflect this.

api/docs/ldstex.dox
```diff
@@ -106,17 +106,26 @@ discussed below.
...
-engine has a potential problem.
+in between the two.  First, an unlinked block will execute many loads and stores on its
+path into DynamoRIO and inside DynamoRIO system code.  Thus, on first execution of a pair
+of split load-exclusive and store-exclusive blocks, the store-exclusive will likely fail
+due to all of the operations in between while building its block.  Even once the two
+blocks are linked, though, if the blocks are too far apart for a single branch
+instruction to reach, DynamoRIO jumps through an exit stub which contains loads and
+stores in order to reach the target block (see \ref page_aarch64_far).
+
+The operations after an unlinked block also affect trace building.  During trace
```
abhinav92003, line 117: How about "... operations after unlinking a block also affect ... "? "unlinked block" may also mean a block that is simply not linked (like in the previous paragraph).

abhinav92003 (approved):

api/docs/ldstex.dox
derekbruening: Not following: this is referring to any unlinked block.  "operations after unlinking a block" sounds like either events happening inside the agent that did the unlocking rather than the block that got unlocked, or it implies events that only happen once when a block is marked unlinked, rather than every time an unlinked bock is executed.

abhinav92003: > rather than the block that got unlocked

I agree that this is what we want to say. I only was suggesting that we should make it clear that we're talking about a block that got unlinked now, rather than an "unlinked block" which may also mean a block that was simply never linked (like the usage in the previous paragraph).

derekbruening: But this lead-in sentence is referring to the previous paragraph's discussion about blocks in an unlinked state in general, and even for the subsequent trace building discussion it could include a block that was just created and not yet linked, if trace discovery goes onto a new path.

Not sure it's worth further time...inclined to leave it and move on if that's ok.

abhinav92003: Sounds good

Issue #5295, opened 2022-01-27
AArch64 opnd_create_abs_addr() succeeds but opnd_is_memory_reference() returns false


derekbruening:
For #4134 and #3995 I'm porting drbbdup to arm and its tests call `opnd_create_abs_addr()` to create the `runtime_case_opnd`.  DR lets it create an absolute addr opnd of type `ABS_ADDR_kind` on AArch64; but then `opnd_is_memory_reference` returns false on AArch64 because the check for `opnd_is_abs_addr()` is limited to `IF_X86_64`.  Basically, DR isn't sure it wants to support absolute address operands for AArch64, since the addressing modes there do not have that sort of thing.  Maybe users are expected to use `opnd_create_rel_addr()` instead (AArch64 does have `#define OPND_CREATE_ABSMEM(addr, size) opnd_create_rel_addr(addr, size)`): but the docs for  `opnd_create_abs_addr()` say it will auto-convert into a rel addr.

The rest of the code for abs addr handling in opnd_shared.c is just gated by `X64`, so `opnd_is_memory_reference` looks like an anomaly.  So maybe the solution is to fix `opnd_is_memory_reference` to return true, and ensure the AArch64 and ARM encoders handle `ABS_ADDR_kind` and treat it just like rel-addr (the alternative of converting to rel-addr at creation time might confuse users who would query their own just-created abs-addr opnd and it would return false for `opnd_is_abs_addr()`).

I went to check whether the AArch64 encoder might already handle abs-addr: but it looks like it doesn't even handle rel-addr!  See 
https://github.com/DynamoRIO/dynamorio/issues/4847#issuecomment-1023433975

derekbruening:
For 32-bit ARM: from a quick glance at the encoder, it looks like it does not think abs-addr is valid and expects only rel-addr.

derekbruening:
@abhinav92003 is voting for making abs-addr x86-only.  Looking at 32-bit arm implementation notes: I do not see any mention of it so I assume there was no real plan for it; so making abs-addr operands x86-only seems reasonable.

Issue #1834, opened 2015-12-10
better support for clients de-referencing far or PC-relative app memory values: add drutil_insert_get_mem_value()?


derekbruening:
Split from #1823

This seems like a more general problem than just an issue with
dr_insert_mbr_instrumentation(). While often inserted code that
de-references app memory is app code, this is not always the case, and we
want to support tool code doing so. Similarly to the stolen register on
ARM, we should at least document that the burden falls on clients to detect
and handle far refs on Linux rather than blindly copying opnds. We should
also add convenience routines if it is too awkward to work around this
using regular API routines.

A similar problem exists with PC-relative memory references: if a tool
blindly copies the operand, is DR supposed to try and mangle it,
potentially spilling a register and maybe messing up the tool's
instrumentation assumptions?

One proposal would be a drutil_insert_get_mem_value() routine in an
extension that handles the corner cases.

derekbruening:
** TODO on amd64 today tools rely on DR mangling pc-rel instrs, which could fail

If we have mangle_rel_addr() only operate on app instrs, these all fail:

```
debug-internal-64: 219 tests passed, **** 5 tests failed: ****
        code_api|client.count-ctis 
        code_api|client.count-ctis-noopt 
        code_api|sample.modxfer 
        code_api|sample.modxfer_app2lib 
        code_api|sample.instrcalls 

<Application /work/dr/build_suite/build_debug-internal-64/suite/tests/bin/client.count-ctis (5413) DynamoRIO usage error : encode error: rip-relative reference out of 32-bit reach>
```

For count-ctis-noopt:

```
interp: start_pc = 0x00007f4b1f4c8aa0
  0x00007f4b1f4c8aa0  ff 25 22 05 22 00    jmp    <rel> 0x00007f4b1f6e8fc8[8byte]

bb ilist before mangling:
TAG  0x00007f4b1f4c8aa0
 +0    m4 @0x000000004e78b120  65 48 89 0c 25 18 00 mov    %rcx -> %gs:0x18[8byte]
                               00 00
 +9    m4 @0x000000004e78b1a0  48 8b 0d c1 e5 fc d0 mov    <rel> 0x00007f4b1f6e8fc8[8byte] -> %rcx
```

It relies on DR re-relativizing the app's PC-rel mem ref to pass as an arg to a clean call.

This could fail if the cache is not 32-bit-reachable from the app code
(which is not guaranteed).

[...]

PR #5298, opened 2022-01-28
i#5297: Relax a64 jmp-to-label encoding failures

derekbruening:
Removes a requirement for an aligned target when encoding without
checking reachability, to allow a branch to a label whose note is set
to a small integer during instrumentation.

Removes a debug warning about an encoding failure for
!assert_reachable for the same reason: a branch to a label with a
small note value.

Adds a regression test case to api.ir.  Confirmed that it fails
without these fixes.

Fixes #5297

PR #5317, opened 2022-02-02
i#4134 drbbdup: Port drbbdup to AArchXX

derekbruening:
Enables building drbbdup and its tests on AArchXX.

Adds case comparison, conditional branch, and TLS access code for both
AArch64 and ARM.  Uses optimizations for comparing small case values;
a second scratch register is needed for larger values.

Ports the tests by using OPND_CREATE_ABSMEM (xref #5295).  To make
these pc-relative case encoding loads work on AArchXX in the presence
of AArch64 encoder bugs and to avoid a more-complex interface, drbbdup
auto-inserts loading the target address into its scratch register to
avoid reachability problems (xref #1834 on having DR do this during
mangling).

Avoids storing block translations if there is no dynamic handling,
improving robustness and reducing memory usage.

Adds the tests to the passes-under-QEMU list for cross-compilation.
Additionally, I tested manually on native ARM and AArch64 machines.

Issue: #4134, #5295, #1834, #3995

### DIFF
--- a/ext/drbbdup/drbbdup.c
+++ b/ext/drbbdup/drbbdup.c
@@ -802,7 +802,6 @@ drbbdup_insert_compare_encoding_and_branch(void *drcontext, instrlist_t *bb,
                                            reg_id_t reg_encoding, bool jmp_if_equal,
                                            instr_t *jmp_label)
 {
-#define MINSERT instrlist_meta_preinsert
 #ifdef X86
 #    ifdef X86_64
     /* XXX: We could use an immediate for small values.
--- a/ext/drbbdup/drbbdup.c
+++ b/ext/drbbdup/drbbdup.c
@@ -809,14 +809,16 @@ drbbdup_insert_compare_encoding_and_branch(void *drcontext, instrlist_t *bb,
      * We could also use jecxz and avoid flags altogether for zero values.
      */
     opnd_t opnd = opnd_create_abs_addr(&current_case->encoding, OPSZ_PTR);
-    MINSERT(bb, where, XINST_CREATE_cmp(drcontext, opnd, opnd_create_reg(reg_encoding)));
+    instrlist_meta_preinsert(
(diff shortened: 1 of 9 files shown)


derekbruening:
This currently includes PR #5298. Hoping to have that merged to master soon and then will update here; did not seem worth it to make PR off of that branch.

ext/drbbdup/drbbdup.c
```diff
@@ -751,31 +776,115 @@ drbbdup_encode_runtime_case(void *drcontext, drbbdup_per_thread *pt, void *tag,
...
-#ifdef X86_64
 static void
-drbbdup_insert_compare_encoding(void *drcontext, instrlist_t *bb, instr_t *where,
-                                drbbdup_case_t *current_case, reg_id_t reg_encoding)
+drbbdup_insert_compare_encoding_and_branch(void *drcontext, instrlist_t *bb,
+                                           instr_t *where, drbbdup_case_t *current_case,
+                                           reg_id_t reg_encoding, bool jmp_if_equal,
+                                           instr_t *jmp_label)
 {
+#define MINSERT instrlist_meta_preinsert
```
johnfxgalea, line 805: Nit: (me being super picky here!!) It seems odd, in terms of style, that this definition is here and not universally used throughout the file. I would remove the definition in this PR, and then have another PR that solely introduces `MINSERT` and replaces all occurrences of  `instrlist_meta_preinsert` with the definition in  drbbdup.c.

johnfxgalea (approved):
Great! This is awesome to see!

Just one optional nit, but otherwise LGTM!

ext/drbbdup/drbbdup.c
derekbruening: That's fair: consistency is good.  Done.  (Agreed that macros are generally not preferred, but with these long lines and with more and more instructions being inserted it can help to make them more concise and get closer to an assembly listing (xref #4510).)

PR #5321, opened 2022-02-02
i#4134 drbbdup: Add acquire semantics to case loading

derekbruening:
Updates drbbdup_options_t.atomic_load_encoding to provide acquire
semantics, which is all that's generally needed for a load of a flag.

Implements the option on AArchXX (nothing to do on x86).

Adds a sanity test to one of the existing tests, ensuring the code
generated at least executes succesfully in one thread.

Issue: #4134

### DIFF
--- a/ext/drbbdup/drbbdup.c
+++ b/ext/drbbdup/drbbdup.c
@@ -775,7 +775,6 @@ drbbdup_encode_runtime_case(void *drcontext, drbbdup_per_thread *pt, void *tag,
     }
 
     /* Load the encoding to the scratch register. */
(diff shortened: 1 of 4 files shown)


johnfxgalea (approved):

johnfxgalea:
> Your commit message does not mention anything about the modifications to drx.

But the mods are insignificant...

derekbruening:
> > Your commit message does not mention anything about the modifications to drx.
> 
> But the mods are insignificant...

Whoops, merged before I saw this.  Should have mentioned adding the official fail on trying to get LOCK on aarch64.

PR #5322, opened 2022-02-02
i#4134 drbbdup: Avoid 2nd scratch via max_case_encoding

derekbruening:
Adds a new drbbdup option max_case_encoding.  When the value is small,
we can avoid a 2nd scratch register for aarchxx, since we can fit all
encodings in immediates inside compare instructions.

Adds a sanity check.
Confirmed visually in the debug logs that the 2nd scratch is gone:
Before:
```
  --------------------------------------------------
  after instrumentation:
  TAG  0x0000ffff868340c0
   +0    m4 @0x0000fffd428950e8  f900b781   str    %x1 -> +0x0168(%x28)[8byte]
   +4    m4 @0x0000fffd42894da0  d53b4200   mrs    %nzcv -> %x0
   +8    m4 @0x0000fffd42894cd8  f900af80   str    %x0 -> +0x0158(%x28)[8byte]
   +12   m4 @0x0000fffd42894c58  d28c1000   movz   $0x6080 lsl $0x00 -> %x0
   +16   m4 @0x0000fffd42894bd8  f2a85000   movk   %x0 $0x4280 lsl $0x10 -> %x0
   +20   m4 @0x0000fffd42894b10  f2dfffe0   movk   %x0 $0xffff lsl $0x20 -> %x0
   +24   m4 @0x0000fffd42894a48  f9400000   ldr    (%x0)[8byte] -> %x0
   +28   m4 @0x0000fffd42894e20  f9400000   <label>
   +28   m4 @0x0000fffd42894980  f100041f   subs   %x0 $0x0000000000000001 lsl $0x0000000000000000 -> %xzr
   +32   m4 @0x0000fffd42894900  54000001   b.ne   @0x0000fffd42894fa0[8byte]
  --------------------------------------------------
```
After:
```
  --------------------------------------------------
  after instrumentation:
  TAG  0x0000ffff9aedf0c0
   +0    m4 @0x0000fffd56f400e8  d53b4200   mrs    %nzcv -> %x0
   +4    m4 @0x0000fffd56f3fda0  f900af80   str    %x0 -> +0x0158(%x28)[8byte]
   +8    m4 @0x0000fffd56f3fcd8  d2821000   movz   $0x1080 lsl $0x00 -> %x0
   +12   m4 @0x0000fffd56f3fc58  f2aadd60   movk   %x0 $0x56eb lsl $0x10 -> %x0
   +16   m4 @0x0000fffd56f3fbd8  f2dfffe0   movk   %x0 $0xffff lsl $0x20 -> %x0
   +20   m4 @0x0000fffd56f3fb10  f9400000   ldr    (%x0)[8byte] -> %x0
   +24   m4 @0x0000fffd56f3fe20  f9400000   <label>
   +24   m4 @0x0000fffd56f3fa48  f100041f   subs   %x0 $0x0000000000000001 lsl $0x0000000000000000 -> %xzr
   +28   m4 @0x0000fffd56f3f980  54000001   b.ne   @0x0000fffd56f3ffa0[8byte]
  --------------------------------------------------
```

Issue: #4134

### DIFF
--- a/ext/drbbdup/drbbdup.c
+++ b/ext/drbbdup/drbbdup.c
@@ -39,6 +39,7 @@
 #include <string.h>
 #include <stdint.h>
 #include <stddef.h>
+#include <limits.h>
 #include "../ext_utils.h"
 
 #undef drmgr_is_first_instr
@@ -813,12 +814,17 @@ drbbdup_insert_compare_encoding_and_branch(void *drcontext, instrlist_t *bb,
 {
 #ifdef X86
 #    ifdef X86_64
-    /* XXX: We could use an immediate for small values.
-     * We could also use jecxz and avoid flags altogether for zero values.
-     */
-    opnd_t opnd = opnd_create_abs_addr(&current_case->encoding, OPSZ_PTR);
-    instrlist_meta_preinsert(
-        bb, where, XINST_CREATE_cmp(drcontext, opnd, opnd_create_reg(reg_encoding)));
+    /* XXX: We could use jecxz and avoid flags altogether for zero values. */
+    if (current_case->encoding <= INT_MAX) {
+        /* It fits in an immediate so we can avoid the load. */
+        opnd_t opnd = opnd_create_immed_uint(current_case->encoding, OPSZ_4);
+        instrlist_meta_preinsert(
+            bb, where, XINST_CREATE_cmp(drcontext, opnd_create_reg(reg_encoding), opnd));
+    } else {
+        opnd_t opnd = opnd_create_abs_addr(&current_case->encoding, OPSZ_PTR);
+        instrlist_meta_preinsert(
+            bb, where, XINST_CREATE_cmp(drcontext, opnd, opnd_create_reg(reg_encoding)));
+    }
 #    elif defined(X86_32)
     opnd_t opnd = opnd_create_immed_uint(current_case->encoding, OPSZ_PTR);
     instrlist_meta_preinsert(
--- a/ext/drbbdup/drbbdup.c
+++ b/ext/drbbdup/drbbdup.c
@@ -70,7 +70,7 @@
 #define TABLE_SIZE 65536 /* Must be a power of 2 to perform efficient mod. */
 
 #ifdef AARCHXX
-#    define MAX_IMMED_IN_CMP 256
+#    define MAX_IMMED_IN_CMP 255
 #endif
 
 typedef enum {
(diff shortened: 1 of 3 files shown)


johnfxgalea:
@derekbruening Is this ready for review?

derekbruening:
> @derekbruening Is this ready for review?

Yes, forgot to request it.  Thanks for noticing.

ext/drbbdup/drbbdup.c
```diff
@@ -735,14 +743,15 @@ drbbdup_encode_runtime_case(void *drcontext, drbbdup_per_thread *pt, void *tag,
...
-    drreg_is_register_dead(drcontext, DRBBDUP_SCRATCH_REG2, where,
-                           &manager->is_scratch_reg2_dead);
-    if (!manager->is_scratch_reg2_dead) {
-        /* TODO i#4134: Add a user-set flag promising that runtime case values are
-         * always <256 so we know we don't need this 2nd scratch register.
-         */
-        drbbdup_spill_register(drcontext, bb, where, DRBBDUP_SCRATCH_REG2_SLOT,
-                               DRBBDUP_SCRATCH_REG2);
+    if (opts.max_case_encoding > 0 && opts.max_case_encoding < MAX_IMMED_IN_CMP)
+        manager->is_scratch_reg2_dead = true; /* Avoid restore, since not used. */
```
johnfxgalea, line 747: I personally would have introduced a new flag that denotes whether reg2 is in use rather that piggybacking on the dead flag. It makes the code more readable and maybe less error-prone  as now the dead flag can be true when it actual fact the register is not.

johnfxgalea:
You can punt the x86_64 optimisation for `drbbdup_insert_compare_encoding_and_branch` to another PR, but it is also suitable here as it also leverages the assumption of a max encoding case (32-bit size in this case).

derekbruening:
> You can punt the x86_64 optimisation for `drbbdup_insert_compare_encoding_and_branch` to another PR, but it is also suitable here as it also leverages the assumption of a max encoding case (32-bit size in this case).

You mean adding the use of an immediate instead of a load?  OK.  Doesn't it not depend on the max encoding: it only depends on the current case value, which never changes?

johnfxgalea:
> > You can punt the x86_64 optimisation for `drbbdup_insert_compare_encoding_and_branch` to another PR, but it is also suitable here as it also leverages the assumption of a max encoding case (32-bit size in this case).
> 
> You mean adding the use of an immediate instead of a load? OK. Doesn't it not depend on the max encoding: it only depends on the current case value, which never changes?



> > You can punt the x86_64 optimisation for `drbbdup_insert_compare_encoding_and_branch` to another PR, but it is also suitable here as it also leverages the assumption of a max encoding case (32-bit size in this case).
> 
> You mean adding the use of an immediate instead of a load? OK. Doesn't it not depend on the max encoding: it only depends on the current case value, which never changes?

Yeah. Okay, max encoding is not needed for that optimisation; we can just check the size of each encoding.

ext/drbbdup/drbbdup.c
derekbruening: Done.

johnfxgalea:
LGTM. One thing which I noticed is that `drbbdup_prepare_redirect` does not restore reg2 before redirection when handling dynamic cases. Shouldn't reg2 be restored there?

derekbruening:
> LGTM. One thing which I noticed is that `drbbdup_prepare_redirect` does not restore reg2 before redirection when handling dynamic cases. Shouldn't reg2 be restored there?

Good catch.  Done.

Note that you didn't actually mark "Approve" (the "LGTM" string doesn't do it in this system).

johnfxgalea (approved):

Issue #5328, opened 2022-02-04
dynamorio.org docx not recent


abhinav92003:
A documentation page I submitted last week was not picked up by the most recent docs release: dynamorio.org/page_scatter_gather_emulation.html

I also found that the release notes page (https://dynamorio.org/release_notes.html) is not updated with this commit from Dec 2021 https://github.com/DynamoRIO/dynamorio/commit/bd060e5f8545962da3bed134fb64d7ef148d7a54#

Looking at the last three runs of ci-docx: https://github.com/DynamoRIO/dynamorio/actions/workflows/ci-docs.yml, I see the same commit hash in all:
```
Submodule path 'dynamorio': checked out '5e27a98213a18575ade3c3e089dc1cd0dacdc696'
```

This is from this commit in Dec 2021: https://github.com/DynamoRIO/dynamorio/commit/5e27a98213a18575ade3c3e089dc1cd0dacdc696

abhinav92003:
I'm seeing this error in the "Get version" stage of the workflow:

```
lline 6: [: : integer expression expected
```
in this line of the script:
https://github.com/DynamoRIO/dynamorio/blob/4d8c8372681367051c091ca5659780a396ac1c4f/.github/workflows/ci-docs.yml#L95

I see that the `build` variable should default to zero:
https://github.com/DynamoRIO/dynamorio/blob/4d8c8372681367051c091ca5659780a396ac1c4f/.github/workflows/ci-docs.yml#L51

But in the expanded script, it shows blank:
```
if test -z ""; then
    export VERSION_NUMBER=8.0.$((`git log -n 1 --format=%ct` / (60*60*2[4](https://github.com/DynamoRIO/dynamorio/runs/5061555949?check_suite_focus=true#step:5:4))))
  else
    export VERSION_NUMBER=
  fi
  if [ "" -ne 0 ]; then
    export VERSION_NUMBER="${VERSION_NUMBER}-"
  fi
  echo "::set-output name=version_number::${VERSION_NUMBER}"
```

abhinav92003:
For `page_scatter_gather_emulation` I actually found this error: https://github.com/DynamoRIO/dynamorio.github.io/runs/5061627854?check_suite_focus=true

> I also found that the release notes page (https://dynamorio.org/release_notes.html) is not updated with this commit from Dec 2021 https://github.com/DynamoRIO/dynamorio/commit/bd060e5f8545962da3bed134fb64d7ef148d7a54

[...]

PR #5330, opened 2022-02-07
i#5328: Fix broken docs release.

abhinav92003:
Fixes the dynamorio.org docs release by addressing
the unexpected Jekyll tag string "{%" that is present
as part of the scatter and gather instruction
disassembly.

Adds the raw-endraw pair around "{%" and "%}" to
prevent them from being interpreted as tag-start
and tag-end respectively.

Also removes the old workaround of adding raw-endraw
in docx itself, from drdisas.dox.

Tested locally that this fix works.

Fixes: #5328

### DIFF
--- a/api/docs/CMake_rundoxygen.cmake
+++ b/api/docs/CMake_rundoxygen.cmake
@@ -186,11 +186,13 @@ if (embeddable)
         "<object type=\"image/svg+xml\" data=\"/images/"
         string "${string}")
     # Ensure that strings that look like Jekyll tags are not interpreted as such.
-    # We add the Jekyll raw-endraw tags around "{%" and "%}" separately.
-    string(REGEX REPLACE "{%" "jekyll_raw_tag{%jekyll_endraw_tag" string "${string}")
-    string(REGEX REPLACE "%}" "jekyll_raw_tag%}jekyll_endraw_tag" string "${string}")
-    string(REGEX REPLACE "jekyll_raw_tag" "{% raw %}" string "${string}")
-    string(REGEX REPLACE "jekyll_endraw_tag" "{% endraw %}" string "${string}")
+    # We add the Jekyll raw-endraw tags around "{%" and "%}" separately. We use
+    # a random suffix for jekyll_raw_tag and jekyll_endraw_tag to reduce the
+    # possibility of a conflict with an actual use of these strings.
+    string(REGEX REPLACE "{%" "jekyll_raw_tag_f00d{%jekyll_endraw_tag_f00d" string "${string}")
+    string(REGEX REPLACE "%}" "jekyll_raw_tag_f00d%}jekyll_endraw_tag_f00d" string "${string}")
+    string(REGEX REPLACE "jekyll_raw_tag_f00d" "{% raw %}" string "${string}")
+    string(REGEX REPLACE "jekyll_endraw_tag_f00d" "{% endraw %}" string "${string}")
     # Collect type info for keyword searches with direct anchor links (else the
     # search finds the page but the user then has to manually search within the long
     # page -- and there are no doygen options to split each onto its own page).
--- a/api/docs/CMake_rundoxygen.cmake
+++ b/api/docs/CMake_rundoxygen.cmake
@@ -186,6 +186,10 @@ if (embeddable)
         "<object type=\"image/svg+xml\" data=\"/images/"
         string "${string}")
 
+    string(REGEX REPLACE "{%" "jekyll_raw_tag{%jekyll_endraw_tag" string "${string}")
+    string(REGEX REPLACE "%}" "jekyll_raw_tag%}jekyll_endraw_tag" string "${string}")
+    string(REGEX REPLACE "jekyll_raw_tag" "{% raw %}" string "${string}")
+    string(REGEX REPLACE "jekyll_endraw_tag" "{% endraw %}" string "${string}")
     # Collect type info for keyword searches with direct anchor links (else the
     # search finds the page but the user then has to manually search within the long
     # page -- and there are no doygen options to split each onto its own page).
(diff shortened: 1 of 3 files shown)


api/docs/scatter_gather_emulation.dox
```diff
@@ -61,7 +61,7 @@ be complete.
 
 
 ```
-vpscatterdd {%k1} %xmm10 -> %rax(,%xmm11,4)[4byte] %k1
+{%raw %}vpscatterdd {%k1} %xmm10 -> %rax(,%xmm11,4)[4byte] %k1{%endraw %}
```
derekbruening, line 64: nit: All our Liquid tags have a space between the `%` and the tag name.  The same style with a space is what they use in all the Liquid docs I've seen.

derekbruening, line 64: For the direct (not for embedding in Jekyll: `api/docs/html` rather than `api/docs/embed/html`) html output for this doc, what does this look like?  I assume it has this `{% raw %}` there.  Would an alternative solution of `&lt;%k1}` solve the Jekyll issue while also looking fine in the direct html output?

abhinav92003: > what does this look like? I assume it has this {% raw %} there.

It shows the `{%raw %}` and `{%endraw %}` as is. I saw this before but assumed it's okay as the website contains the Jekyll one only. And we had an existing use of this workaround: https://github.com/DynamoRIO/dynamorio/blob/c99bcafc7e9057e0adb83fc42784a2fb1220e27e/clients/drdisas/drdisas.dox#L45

> Would an alternative solution of &lt;%k1} solve the Jekyll issue while also looking fine in the direct html output?

I can try.

derekbruening: Whatever choice we make let's make that drdisas one do the same thing as the ones here.

abhinav92003: > Would an alternative solution of &lt;%k1} solve the Jekyll issue while also looking fine in the direct html output?

It didn't. Perhaps because this is in a code block, the '&' itself gets converted to &amp. So the html output shows `&gt;`.

Also, I assume you meant `&#123;` for `{`.

abhinav92003: Added the space.

abhinav92003: Do we use the `api/docs/html` output for anything? Or is it okay for it to have the `{% raw %}` tag.

derekbruening: Yes, that output is placed into the release package.  If you untar/unzip any of our release packages it is right there for local browsing.  If we had a Windows installer (instead of just zip file) like Dr. Memory does it would create a "Start Menu" link that points to that html output.

derekbruening: Maybe the best solution is a regex replacement that auto-adds `{% raw %}...{% endraw %}` around `{%.*}` in the existing cmake code that creates the embedded docs.  Then the regular html is fine and nobody will add future disassembly and hit this same problem like may well happen if we rely on a manual fix each time.

abhinav92003: Done. Instead of adding `{% raw %}...{% endraw %}` around `{%.*}`, I added changes to add `{% raw %}...{% endraw %}` around `{%` and `%}` separately. I think this is better because `{%` and `%}` both can occur independently (here we have only the former though). Also, `{` and `}` alone are not problematic afaik.

abhinav92003:
Made the changes. PTAL.

api/docs/CMake_rundoxygen.cmake
```diff
@@ -185,7 +185,12 @@ if (embeddable)
         "<object type=\"image/svg\\+xml\" data=\""
         "<object type=\"image/svg+xml\" data=\"/images/"
         string "${string}")
-
+    # Ensure that strings that look like Jekyll tags are not interpreted as such.
+    # We add the Jekyll raw-endraw tags around "{%" and "%}" separately.
+    string(REGEX REPLACE "{%" "jekyll_raw_tag{%jekyll_endraw_tag" string "${string}")
+    string(REGEX REPLACE "%}" "jekyll_raw_tag%}jekyll_endraw_tag" string "${string}")
+    string(REGEX REPLACE "jekyll_raw_tag" "{% raw %}" string "${string}")
```
abhinav92003, line 192: Thinking whether I should throw a random number in there `jekyll_raw_tag_1234` or something. So that it doesn't match with actual text if we ever add it to doc.

api/docs/scatter_gather_emulation.dox
abhinav92003: So this approach also assumes that we won't have any actual use of Jekyll tags in the file. I suppose that's okay?

derekbruening: I think so, yes.

api/docs/CMake_rundoxygen.cmake
derekbruening: OK, or a bunch of underscores or sthg.

derekbruening (approved):

api/docs/CMake_rundoxygen.cmake
abhinav92003: Done.

abhinav92003:
So another option is to use `render_with_liquid: false` to disable Liquid entirely for a document: https://jekyllrb.com/docs/liquid/tags/. But I see that we have a tag usage here: https://github.com/DynamoRIO/dynamorio/blob/deeca56bc76c6d458f0592597cee5b52c7307085/api/docs/CMake_rundoxygen.cmake#L261

So we might not want to disable Liquid parsing entirely. This PR's strategy is better.

PR #5337, opened 2022-02-09
i#2626: Fix operand clashes in AArch64 codec

AssadHashmi:
The current functions for encoding and decoding
vector element size operands using immediate int
enums clashes with operands representing actual
immediate ints.

Issue: #2626


joshua-warburton (approved):
Looks good for now, probably worth having a deeper look into the element size operand when  encoding is properly tested

PR #5341, opened 2022-02-09
i#5336: Add missing pcprofile whereami category

### LINKED ISSUE #5336: ASSERT "unknown whereAmI 6" core/unix/pcprofile.c:563
This has happened twice in the master-merge tests:

https://github.com/DynamoRIO/dynamorio/issues/1807#issuecomment-999296015
`prof_pcs|common.decode`
```
ERROR: unknown whereAmI 6
<Application /home/runner/work/dynamorio/dynamorio/build_debug-internal-64/suite/tests/bin/common.decode (41356). Internal Error: DynamoRIO debug check failure: /home/runner/work/dynamorio/dynamorio/core/unix/pcprofile.c:563 false
```
https://github.com/DynamoRIO/dynamorio/runs/4602849193

Happened again: https://github.com/DynamoRIO/dynamorio/runs/4986838918

I believe this should be easy to fix.

derekbruening:
The pcprofile printing code was missing DR_WHERE_TRAMPOLINE, leading
to an assert.

I was not able to reproduce the assert locally so the new case is
technically not tested.

Fixes #5336

### DIFF
--- a/core/unix/pcprofile.c
+++ b/core/unix/pcprofile.c
@@ -1,5 +1,5 @@
 /* **********************************************************
- * Copyright (c) 2011-2021 Google, Inc.  All rights reserved.
+ * Copyright (c) 2011-2022 Google, Inc.  All rights reserved.
  * Copyright (c) 2001-2010 VMware, Inc.  All rights reserved.
  * **********************************************************/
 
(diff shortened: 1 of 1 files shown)


core/unix/pcprofile.c
```diff
@@ -548,10 +548,12 @@ pcprofile_results(thread_pc_info_t *info)
                     print_file(info->file, " dispatch\n");
                 } else if (e->whereami == DR_WHERE_MONITOR) {
                     print_file(info->file, " monitor\n");
-                } else if (e->whereami == DR_WHERE_SIGNAL_HANDLER) {
-                    print_file(info->file, " signal handler\n");
                 } else if (e->whereami == DR_WHERE_SYSCALL_HANDLER) {
                     print_file(info->file, " syscall handler\n");
+                } else if (e->whereami == DR_WHERE_SIGNAL_HANDLER) {
+                    print_file(info->file, " signal handler\n");
```
abhinav92003, line 554: I see that the order between `DR_WHERE_SYSCALL_HANDLER` and `DR_WHERE_SIGNAL_HANDLER` has changed too. Is that intended?

abhinav92003 (approved):

core/unix/pcprofile.c
derekbruening: Yes, I made it match the enum order, which makes it easier to see whether anything is missing.

derekbruening: If this were the only dispatch I would use a case and the compiler could tell us if we missed something, but two of the cases are large and handled separately so I abandoned that plan.  Plus I think that only works if we have no default case.

PR #5344, opened 2022-02-10
Remove .DS_Store and add to .gitignore

derekbruening:
Removes a file unintentionally added in commit 9de4dfbe and adds
.DS_Store to .gitignore to prevent this in the future.


abhinav92003 (approved):

abhinav92003:
Thanks for removing that .DS_Store (maybe I did a `git add *` and it added it to the commit), and for adding it to gitignore.

PR #5347, opened 2022-02-10
i#5345: Add missing SYSCALL and SYSRET operands

### LINKED ISSUE #5345: SYSCALL instruction fails to list r11 as a destination
DR's decoder only lists rcx as a destination for SYSCALL, while in fact it clobbers r11 to store the flags:
```
  {OP_syscall, 0x0f0510, "syscall", xcx, xx, xx, xx, xx, no, x, NA}, /* AMD/x64 only */
```

Similarly, SYSRET does not list r11 as a source, nor rcx for that matter:
```
  {OP_sysret, 0x0f0710, "sysret", xx, xx, xx, xx, xx, no, x, NA}, /* AMD/x64 only */
```

derekbruening:
Adds missing r11 destination for SYSCALL.
Adds missing rcx and r11 sources for SYSRET.
Updates the creation macros.

Fixes #5345

### DIFF
--- a/core/ir/x86/decode_table.c
+++ b/core/ir/x86/decode_table.c
@@ -1,5 +1,5 @@
 /* **********************************************************
- * Copyright (c) 2011-2021 Google, Inc.  All rights reserved.
+ * Copyright (c) 2011-2022 Google, Inc.  All rights reserved.
  * Copyright (c) 2001-2010 VMware, Inc.  All rights reserved.
  * **********************************************************/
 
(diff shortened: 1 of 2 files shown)


core/arch/emit_utils_shared.c
```diff
@@ -3557,8 +3557,10 @@ create_syscall_instr(dcontext_t *dcontext)
         return create_int_syscall_instr(dcontext);
     } else if (method == SYSCALL_METHOD_SYSENTER) {
         return INSTR_CREATE_sysenter(dcontext);
+#    ifdef X64
```
abhinav92003, line 3560: Maybe add this change to the PR description explicitly?

suite/tests/api/ir_x86_0args.h
```diff
@@ -1,5 +1,5 @@
 /* **********************************************************
- * Copyright (c) 2011-2016 Google, Inc.  All rights reserved.
+ * Copyright (c) 2011-2022 Google, Inc.  All rights reserved.
```
abhinav92003, line 2: Just a header update?

abhinav92003 (approved):

suite/tests/api/ir_x86_0args.h
derekbruening: Ah, I changed this file to mark them X64_ONLY, but in fact they are valid in 32-bit mode on AMD processors, so I reverted those changes but failed to revert the date.  Reverting it now.

core/arch/emit_utils_shared.c
derekbruening: Done.  Actually -- I think we can revert this, since this is valid for 32-bit AMD.

PR #5349, opened 2022-02-11
i#1034: Eliminate signed char assumptions

### LINKED ISSUE #1034: DynamoRIO assumes that char is signed
_From [rnk@google.com](https://code.google.com/u/rnk@google.com/) on December 18, 2012 10:36:19_

I ran into this because the Google build tool sets -funsigned-char by default.  It would be nice if DR was portable with respect to this setting, but it's a minor issue.  I'm mostly reporting this to document the issue in case anyone else runs into it.  It manifests as failures to encode and decode int 0x80 and other instructions.

If we decide to fix this, I would propose adding an sbyte typedef and using it where appropriate.

_Original issue: http://code.google.com/p/dynamorio/issues/detail?id=1034_

derekbruening:
Updates pieces of the code that used "char" and assumed it was signed
to use "sbyte" instead.

Adds "-funsigned-char" to the base compiler flags to really ensure the
code base is free of assumptions and to further ensure it is
interoperable with other toolchains that use "-funsigned-char".

Adds a note to the style guide.

Tweaks the drdecode_x86 test to fail without the fixes here in the x86
decoder.

Fixes #1034

### DIFF
--- a/core/utils.c
+++ b/core/utils.c
@@ -2546,7 +2546,7 @@ strcasecmp_with_wildcards(const char *regexp, const char *consider)
             return -1;
         } else if (*consider == '\0')
             return 1;
-        ASSERT(*regexp != EOF && *consider != EOF);
+        ASSERT((sbyte)*regexp != EOF && (sbyte)*consider != EOF);
         cr = (char)tolower(*regexp);
         cc = (char)tolower(*consider);
         if (cr != '?' && cr != cc) {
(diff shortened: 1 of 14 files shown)


core/utils.c
```diff
@@ -2546,7 +2546,7 @@ strcasecmp_with_wildcards(const char *regexp, const char *consider)
             return -1;
         } else if (*consider == '\0')
             return 1;
-        ASSERT(*regexp != EOF && *consider != EOF);
+        ASSERT((sbyte)*regexp != EOF && (sbyte)*consider != EOF);
```
abhinav92003, line 2549: Wondering why this change was required.

In any case, I think EOF is an int (https://www.gnu.org/software/libc/manual/html_node/EOF-and-Errors.html), so also wondering why we cast the other operand to an `sbyte`

abhinav92003 (approved):

core/utils.c
derekbruening: EOF is defined as -1.  *regexp is `char` which is unsigned so if it's `0xff` and we cast to `int` it becomes 255, not -1.  So we have to first cast to a signed type before widening.

Issue #5329, opened 2022-02-05
ASSERT tool.drcachesim.scattergather test: tracer.cpp:394: towrite <= ipc_pipe.get_atomic_write_size() && towrite > 0


derekbruening:
tool.drcachesim.scattergather failed:

https://github.com/DynamoRIO/dynamorio/runs/5077131960?check_suite_focus=true
```
ASSERT FAILURE: /home/runner/work/dynamorio/dynamorio/clients/drcachesim/tracer/tracer.cpp:394: towrite <= ipc_pipe.get_atomic_write_size() && towrite > 0 ()
```

I don't remember ever seeing that assert before on any drcachesim test.

derekbruening:
This was the 32-bit x86 test

derekbruening:
Happened again: https://github.com/DynamoRIO/dynamorio/runs/5131569686

abhinav92003:
I ran the 32-bit test 1000 times  on my machine, and couldn't reproduce the failure:

```
$ ctest -VV -R 'tool.drcachesim.scattergather' --repeat-until-fail 1000
...
1/1 Test #284: code_api|tool.drcachesim.scattergather ...   Passed    1.51 sec
 
The following tests passed:
	code_api|tool.drcachesim.scattergather
 
100% tests passed, 0 tests failed out of 1
 
Total Test time (real) = 1547.39 sec
```

I can try running it on the Github Actions runner using `tmate`. In the past, we've had failures which reproduce only on GA.

abhinav92003:
I ran this 1000 times on a GA runner, no failure encountered. Maybe it reproduces on only some runners? If that is so, I'll need to try this multiple times.

```
ctest -VV -R 'tool.drcachesim.scattergather' --repeat-until-fail 1000
...
1/1 Test #286: code_api|tool.drcachesim.scattergather ...   Passed    2.06 sec

The following tests passed:
        code_api|tool.drcachesim.scattergather

100% tests passed, 0 tests failed out of 1

Total Test time (real) = 2024.27 sec
```

dolanzhao:
My PR #5515 failed to pass test 286 multiple times on the x86_32 action.
https://github.com/DynamoRIO/dynamorio/runs/7048743130?check_suite_focus=true

abhinav92003:
> Looks like the following tests are failing deterministically on my machine (and Derek's) now:

[...]

PR #5351, opened 2022-02-12
i#5329: debug tool.drcachesim.scattergather fail.

abhinav92003:
Add tmate to help reproduce this failure on a GA runner.

Issue: #5329


derekbruening:
@abhinav92003 is this still valid?

abhinav92003:
The failure that this PR tries to reproduce seems to happen only on some GA runners (https://github.com/DynamoRIO/dynamorio/issues/5329#issuecomment-1037255523). This makes it hard to debug. It'd be good to keep this PR open in case I pick this up again.

PR #5357, opened 2022-02-15
i#5356: Add drwrap control inversion option

### LINKED ISSUE #5356: Add drwrap mode that inverts control for use inside a drbbdup case
Split from #3995

Pasting from there:

I guess this raises a larger issue: we want the drwrap clean call to only
be inserted for one of our drbbdup cases. But with no control over drwrap,
we can't arrange that, and drwrap is going to go insert its clean call at
the start before the drbbdup case dispatch I would guess.

How to solve?

If drbbdup were integrated inside drmgr: would non-bbdup-aware users of
drmgr like drwrap be invoked for the default case insertion and not the
other cases? So the user can either have wrapping for just one case (has
to be default)? Provide an option where user can pick one case, or all
cases?

Or, drwrap has to turn its control model around and have the user call
"do_drwrap_instru" from its insertion event, with drwrap not registering
instru events (but still registering modload, etc.)?

Leaning toward the latter: add a drwrap global flag.

However, we have several issues implementing this:

+ There is no app2app event inside drbbdup!
  => For now, don't support drwrap_replace.
+ The "for_trace" and "translating" params are not passed
  => For now, document that drwrap doesn't need them
+ The dr_emit_flags_t return value is not propagated
  => For now, document that when wrapping and not replacing it always
  returns DR_EMIT_DEFAULT

We would have to augment drbbdup to support drwrap_replace which needs app2app
and under some options needs to return something other than DR_EMIT_DEFAULT.

derekbruening:
Adds a new global flag to drwrap which inverts control such that the
client must explicitly invoke drwrap's app2app and instrumentation
handlers.  This lets a client use drbbdup with drwrap only applying to
a subset of the cases.

Given missing pieces in drbbdup, only drwrap wrapping is supported
this way: not drwrap replacing.  (For full support, we would need to
add app2app event support inside drbbdup, as well as support for
returning other than DR_EMIT_DEFAULT for the various instrumentation
handlers.)

Adds a new test drbbdup-drwrap-test which wraps a function in only one
of its drbbdup cases.

The new test includes a check for drwrap_set_global_flags() being
called after drwrap_init() by having a
drwrap_init;drwrap_exit;drwrap_init sequence.  This revealed several
missing unregister calls in drwrap_exit, which we fix here.  These are
not assumed to affect full re-attach since drmgr resets its state
completely; they cause a use-after-free here because drwrap but not
drmgr was reset and we had two thread exit handlers.

Issue: #3995, #5356
Fixes #5356

### DIFF
--- a/ext/drwrap/drwrap.c
+++ b/ext/drwrap/drwrap.c
@@ -946,9 +946,11 @@ drwrap_init(void)
     drmgr_init();
     if (!drmgr_register_bb_app2app_event(drwrap_event_bb_app2app, &pri_replace))
         return false;
-    if (!drmgr_register_bb_instrumentation_event(drwrap_event_bb_analysis,
-                                                 drwrap_event_bb_insert, &pri_insert))
-        return false;
+    if (!TEST(DRWRAP_INVERT_CONTROL, global_flags)) {
+        if (!drmgr_register_bb_instrumentation_event(drwrap_event_bb_analysis,
+                                                     drwrap_event_bb_insert, &pri_insert))
+            return false;
+    }
     if (!drmgr_register_restore_state_ex_event(drwrap_event_restore_state_ex))
         return false;
 
@@ -963,7 +965,9 @@ drwrap_init(void)
                       false /*!str_dup*/, false /*!synch*/, post_call_entry_free, NULL,
                       NULL);
     post_call_rwlock = dr_rwlock_create();
-    wrap_lock = dr_recurlock_create();
+    /* This lock may have been set up by drwrap_set_global_flags() (in this thread). */
+    if (wrap_lock == NULL)
+        wrap_lock = dr_recurlock_create();
     drmgr_register_module_unload_event(drwrap_event_module_unload);
 
     tls_idx = drmgr_register_tls_field();
@@ -1005,12 +1009,27 @@ drwrap_exit(void)
     if (count != 0)
         return;
 
+    if (!TEST(DRWRAP_INVERT_CONTROL, global_flags)) {
+        if (!drmgr_unregister_bb_instrumentation_event(drwrap_event_bb_analysis))
+            ASSERT(false, "failed to unregister in drwrap_exit");
+    }
     if (!drmgr_unregister_bb_app2app_event(drwrap_event_bb_app2app) ||
-        !drmgr_unregister_bb_instrumentation_event(drwrap_event_bb_analysis) ||
         !drmgr_unregister_restore_state_ex_event(drwrap_event_restore_state_ex) ||
         !drmgr_unregister_module_unload_event(drwrap_event_module_unload) ||
+        !drmgr_unregister_thread_init_event(drwrap_thread_init) ||
+        !drmgr_unregister_thread_exit_event(drwrap_thread_exit) ||
         !drmgr_unregister_tls_field(tls_idx))
         ASSERT(false, "failed to unregister in drwrap_exit");
+
+#ifdef WINDOWS
+    if (sysnum_NtContinue != -1) {
+        if (!dr_unregister_filter_syscall_event(drwrap_event_filter_syscall) ||
+            !drmgr_unregister_pre_syscall_event(drwrap_event_pre_syscall))
+            ASSERT(false, "failed to unregister in drwrap_exit");
+    }
+    if (!drmgr_unregister_exception_event_ex(drwrap_event_exception))
+        ASSERT(false, "failed to unregister in drwrap_exit");
+#endif
 
     if (dr_is_detaching()) {
         memset(&drwrap_stats, 0, sizeof(drwrap_stats_t));
@@ -1030,6 +1049,7 @@ drwrap_exit(void)
     hashtable_delete(&post_call_table);
     dr_rwlock_destroy(post_call_rwlock);
     dr_recurlock_destroy(wrap_lock);
+    wrap_lock = NULL; /* For re-attach early drwrap_set_global_flags(). */
     drmgr_exit();
 
     while (post_call_notify_list != NULL) {
@@ -1085,7 +1105,21 @@ drwrap_set_global_flags(drwrap_global_flags_t flags)
 {
     drwrap_global_flags_t old_flags;
     bool res;
+    /* This can be called prior to drwrap_init().
+     * We only support this being done in the single DR-initialization thread.
+     */
+    if (wrap_lock == NULL) {
+        ASSERT(dr_atomic_load32(&drwrap_init_count) == 0,
+               "Unsupported race between drwrap_init() and drwrap_set_global_flags()");
+        wrap_lock = dr_recurlock_create();
+    }
     dr_recurlock_lock(wrap_lock);
+    if (dr_atomic_load32(&drwrap_init_count) > 0 &&
+        /* After drwrap_init() was called, control inversion cannot be changed. */
+        (flags & DRWRAP_INVERT_CONTROL) != (global_flags & DRWRAP_INVERT_CONTROL)) {
+        dr_recurlock_unlock(wrap_lock);
+        return false;
+    }
     /* if anyone asks for safe, be safe.
      * since today the only 2 flags ask for safe, we can accomplish that
      * by simply or-ing in each request.
@@ -1097,6 +1131,38 @@ drwrap_set_global_flags(drwrap_global_flags_t flags)
     res = (global_flags != old_flags);
     dr_recurlock_unlock(wrap_lock);
     return res;
+}
+
+DR_EXPORT
+dr_emit_flags_t
+drwrap_invoke_app2app(void *drcontext, void *tag, instrlist_t *bb, bool for_trace,
+                      bool translating)
+{
+    ASSERT(TEST(DRWRAP_INVERT_CONTROL, global_flags),
+           "must set DRWRAP_INVERT_CONTROL to call drwrap_invoke_app2app");
+    return drwrap_event_bb_app2app(drcontext, tag, bb, for_trace, translating);
+}
+
+DR_EXPORT
+dr_emit_flags_t
+drwrap_invoke_analysis(void *drcontext, void *tag, instrlist_t *bb, bool for_trace,
+                       bool translating, OUT void **user_data)
+{
+    ASSERT(TEST(DRWRAP_INVERT_CONTROL, global_flags),
+           "must set DRWRAP_INVERT_CONTROL to call drwrap_invoke_app2app");
+    return drwrap_event_bb_analysis(drcontext, tag, bb, for_trace, translating,
+                                    user_data);
+}
+
+DR_EXPORT
+dr_emit_flags_t
+drwrap_invoke_insert(void *drcontext, void *tag, instrlist_t *bb, instr_t *inst,
+                     bool for_trace, bool translating, void *user_data)
+{
+    ASSERT(TEST(DRWRAP_INVERT_CONTROL, global_flags),
+           "must set DRWRAP_INVERT_CONTROL to call drwrap_invoke_app2app");
+    return drwrap_event_bb_insert(drcontext, tag, bb, inst, for_trace, translating,
+                                  user_data);
 }
 
 /***************************************************************************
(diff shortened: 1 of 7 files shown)


derekbruening:
On integration into dr$sim I see we need drwrap's insert event to take a distinct 'where' parameter.  I have that implemented in another tree; cherry-picking here.

derekbruening:
Both failures are flakes: api.startstop #4604; Windows has the weird file copying issue.

suite/tests/client-interface/drbbdup-drwrap-test.dll.c
```diff
@@ -0,0 +1,187 @@
...
+}
+
+static void
+event_instrument_instr(void *drcontext, void *tag, instrlist_t *bb, instr_t *instr,
+                       instr_t *where, uintptr_t encoding, void *user_data,
+                       void *orig_analysis_data, void *case_analysis_data)
+{
+    if (encoding == 1) {
+        /* drwrap guarantees it doesn't need "for_trace", "translating", or
+         * a non-DR_EMIT_DEFAULT return value for drwrap_wrap().
```
abhinav92003, line 106: s/drwrap_wrap()/drwrap_invoke_insert()

suite/tests/client-interface/drbbdup-drwrap-test.dll.c
```diff
@@ -0,0 +1,187 @@
...
+                    /* Time to switch modes. */
+                    case_encoding = 1;
+                }
+                consec_nop_count = 0;
+            }
+        }
+    }
+    if (encoding == 1) {
+        /* drwrap guarantees it doesn't need "for_trace", "translating", or
+         * a non-DR_EMIT_DEFAULT return value for drwrap_wrap().
```
abhinav92003, line 92: s/drwrap_wrap()/drwrap_invoke_analysis()

ext/drwrap/drwrap.c
```diff
@@ -885,6 +885,11 @@ static dr_emit_flags_t
 drwrap_event_bb_insert(void *drcontext, void *tag, instrlist_t *bb, instr_t *inst,
                        bool for_trace, bool translating, void *user_data);
 
+static dr_emit_flags_t
+drwrap_event_bb_insert_where(void *drcontext, void *tag, instrlist_t *bb, instr_t *inst,
```
abhinav92003, line 889: Though this is not an API, it'll be helpful to add a comment for this and `drwrap_event_bb_insert`.

ext/drwrap/drwrap.c
```diff
@@ -2300,8 +2361,9 @@ drwrap_event_bb_analysis(void *drcontext, void *tag, instrlist_t *bb, bool for_t
 }
 
 static dr_emit_flags_t
-drwrap_event_bb_insert(void *drcontext, void *tag, instrlist_t *bb, instr_t *inst,
-                       bool for_trace, bool translating, void *user_data)
+drwrap_event_bb_insert_where(void *drcontext, void *tag, instrlist_t *bb, instr_t *inst,
```
abhinav92003, line 2364: I don't understand why we need to support separate `inst` and `where`. As "wrapping" implies adding instrumentation just-before and just-after, wouldn't it be wrong to pass different values for inst (the instructionto be wrapped) and where (where instrumentation has to be placed)?

I see that drbbdup would pass different values for `inst` and `where` in this case:
https://github.com/DynamoRIO/dynamorio/blob/4d8c8372681367051c091ca5659780a396ac1c4f/ext/drbbdup/drbbdup.c#L1272

Is this the use-case? Worth adding a comment to explain it.

ext/drwrap/drwrap.c
```diff
@@ -946,9 +951,11 @@ drwrap_init(void)
     drmgr_init();
     if (!drmgr_register_bb_app2app_event(drwrap_event_bb_app2app, &pri_replace))
```
abhinav92003, line 952: I see that "replace" is performed by `drwrap_event_bb_app2app`. So, the fact that we don't conditionally register `drwrap_event_bb_app2app` means that drwrap will perform "replace" even if `DRWRAP_INVERT_CONTROL` is set. So, control is inverted only for "wrap" but not "replace"? That can be confusing.

How about we drop support for "replace" if `DRWRAP_INVERT_CONTROL` is set? It seems less likely that the user will want such mix-and-match anyway.

ext/drwrap/drwrap.c
```diff
@@ -1085,7 +1110,21 @@ drwrap_set_global_flags(drwrap_global_flags_t flags)
 {
     drwrap_global_flags_t old_flags;
     bool res;
+    /* This can be called prior to drwrap_init().
+     * We only support this being done in the single DR-initialization thread.
+     */
+    if (wrap_lock == NULL) {
+        ASSERT(dr_atomic_load32(&drwrap_init_count) == 0,
+               "Unsupported race between drwrap_init() and drwrap_set_global_flags()");
+        wrap_lock = dr_recurlock_create();
```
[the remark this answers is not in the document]
derekbruening: It's not.  Just like it wouldn't be if a user called drwap_init without a later drwrap_exit.

ext/drwrap/drwrap.h
```diff
@@ -78,6 +81,40 @@ DR_EXPORT
...
+
+DR_EXPORT
+/**
+ * When #drwrap_global_flags_t #DRWRAP_INVERT_CONTROL is set, the user
+ * must call this function from a drmgr analysis event handler
+ * (typically registered with
+ * drmgr_register_bb_instrumentation_event()).  For best results the
+ * calling handler should use a priority of
+ * #DRMGR_PRIORITY_INSERT_DRWRAP or the equivalent ordering within drbbdup.
+ * Currently, the "for_trace" and "translating" parameters are ignored, and
```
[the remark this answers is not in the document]
derekbruening: Without this, we couldn't safely use drwrap from drbbdup unless we add for_trace and translating parameters to the drbbdup callbacks (and the return value while at it, which is needed right now anyway: https://github.com/DynamoRIO/dynamorio/issues/5356#issuecomment-1041126899.  I probably *have* to go add the return value; that's an interface addition/change socould add the others while at it perhaps).

suite/tests/client-interface/drbbdup-drwrap-test.dll.c
derekbruening: No this means for using wrapping (as opposed to replacing)

ext/drwrap/drwrap.c
derekbruening: > As "wrapping" implies adding instrumentation just-before and just-after, wouldn't it be wrong to pass different values for inst (the instructionto be wrapped) and where (where instrumentation has to be placed)?

There are several cases these days where the instruction to monitor and where to put the monitoring are different.  One is emulation, and for that things like the drmemtrace tracer code already have some separation of the two (e.g., in instrument_memref()).  Another is drbbdup, which can't clone the block-final instr if it's a branch or syscall -- so it has just one copy after all the duplicated sequences, but it has to put the monitoring code up inside each duplicated case sequence.

ext/drwrap/drwrap.h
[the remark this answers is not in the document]
derekbruening: Bit the bullet and went and added new versions of the drbbdup callbacks in PR #5361 so now this can go away.

ext/drwrap/drwrap.h
```diff
@@ -78,6 +81,40 @@ DR_EXPORT
...
+
+DR_EXPORT
+/**
+ * When #drwrap_global_flags_t #DRWRAP_INVERT_CONTROL is set, the user
+ * must call this function from a drmgr insertion event handler
+ * (typically registered with
+ * drmgr_register_bb_instrumentation_event()).  For best results the
+ * calling handler should use a priority of
+ * #DRMGR_PRIORITY_INSERT_DRWRAP or the equivalent ordering within drbbdup.
+ * Currently, the "for_trace" and "translating" parameters are ignored, and
```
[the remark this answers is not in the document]
derekbruening: Ditto.

ext/drwrap/drwrap.c
```diff
@@ -1030,6 +1054,7 @@ drwrap_exit(void)
     hashtable_delete(&post_call_table);
     dr_rwlock_destroy(post_call_rwlock);
     dr_recurlock_destroy(wrap_lock);
+    wrap_lock = NULL; /* For re-attach early drwrap_set_global_flags(). */
```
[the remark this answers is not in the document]
derekbruening: Done.

ext/drwrap/drwrap.c
```diff
@@ -1085,7 +1110,21 @@ drwrap_set_global_flags(drwrap_global_flags_t flags)
 {
     drwrap_global_flags_t old_flags;
     bool res;
+    /* This can be called prior to drwrap_init().
```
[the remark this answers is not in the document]
derekbruening: Added a comment in the header re-iterating to be careful with interoperability of the invert-control global change.

ext/drwrap/drwrap.h
```diff
@@ -66,6 +66,9 @@ DR_EXPORT
  * normally) but each call must be paired with a corresponding call to
  * drwrap_exit().
  *
+ * Some drwrap behavior must be set by calling drwrap_set_global_flags() *before*
```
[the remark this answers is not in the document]
derekbruening: Done.

suite/tests/client-interface/drbbdup-drwrap-test.dll.c
```diff
@@ -0,0 +1,187 @@
...
+ * in only a subset of the cases.
+ */
+
+#include "dr_api.h"
+#include "drmgr.h"
+#include "drbbdup.h"
+#include "drwrap.h"
+#include <string.h>
+
+#define CHECK(x, msg)                                                                \
```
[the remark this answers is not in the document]
derekbruening: There are many of these.  Made a separate #5363.

suite/tests/client-interface/drbbdup-drwrap-test.dll.c
derekbruening: Now obsolete in any case.

suite/tests/client-interface/drbbdup-drwrap-test.dll.c
derekbruening: Now obsolete in any case.

ext/drwrap/drwrap.c
derekbruening: Done.

ext/drwrap/drwrap.c
derekbruening: Done.

ext/drwrap/drwrap.h
```diff
@@ -751,6 +788,16 @@ typedef enum {
...
      */
     DRWRAP_FAST_CLEANCALLS = 0x08,
+    /**
+     * This flag must only be set before calling drwrap_init().  If set, drwrap will not
+     * register for the drmgr analysis or insertion events.  The user must instead
+     * explicitly invoke drwrap_invoke_analysis() and drwrap_invoke_insert() from its
+     * own handler for those events.  This "inverted control" mode is provided for
+     * better compatibility with drbbdup where the user wishes to only perform wrapping
+     * in a subset of the drbbdup cases.  Only wrapping is supported this way, not
+     * replacing, at this time.
```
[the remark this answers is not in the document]
derekbruening: Replacing is no longer supported at all.

derekbruening:
Waiting on PR #5363 before updating here to avoid that diff...or could update here now and those changes could be ignored (would have to make new PR to compare vs that branch).

For https://github.com/DynamoRIO/dynamorio/issues/5356#issuecomment-1041568407 I plan to do that as a separate PR.

abhinav92003:
> Waiting on PR #5363 before updating here to avoid that diff...or could update here now and those changes could be ignored (would have to make new PR to compare vs that branch).
> 
> For [#5356 (comment)](https://github.com/DynamoRIO/dynamorio/issues/5356#issuecomment-1041568407) I plan to do that as a separate PR.

Approved.

ext/drwrap/drwrap.c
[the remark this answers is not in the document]
abhinav92003: Makes sense.

ext/drwrap/drwrap.c
abhinav92003: I see. Worth documenting this in a comment.

derekbruening: Done.

abhinav92003 (approved):

PR #5358, opened 2022-02-16
i#3995: Add lazy initialisation

johnfxgalea:
Add code that lazily initialises the code cache for dynamic handling.

Issue: #3995

### DIFF
--- a/ext/drbbdup/drbbdup.c
+++ b/ext/drbbdup/drbbdup.c
@@ -158,6 +158,9 @@ drbbdup_encoding_already_included(drbbdup_manager_t *manager, uintptr_t encoding
 static void
 drbbdup_handle_new_case();
 
+static app_pc
+init_fp_cache(void (*clean_call_func)());
+
 static uintptr_t *
 drbbdup_get_tls_raw_slot_addr(drbbdup_thread_slots_t slot_idx)
 {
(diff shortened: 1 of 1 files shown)


ext/drbbdup/drbbdup.c
```diff
@@ -450,6 +456,9 @@ drbbdup_duplicate_phase(void *drcontext, void *tag, instrlist_t *bb, bool for_tr
                 stats.no_dynamic_handling_count++;
             dr_mutex_unlock(stat_mutex);
         }
+        if (manager->enable_dynamic_handling && new_case_cache_pc == NULL) {
+            new_case_cache_pc = init_fp_cache(drbbdup_handle_new_case);
```
derekbruening, line 460: This is racy.  Are you relying on DR's global bb building lock?  I don't think that's a good idea: we may get rid of that lock and the API does not guarantee it.

Even the read of the global var without a load-acquire might technically never get the new global value.  Should be ok to leave plain loads here and have a load-acquire always done in init_fp_cache (or, use stdatomic.h).

johnfxgalea: There is the dr_rwlock_write_lock() invocation.

derekbruening: Oh ok -- it wasn't visible in the context diff and I only looked at init_fp_cache().  It seems best to add a comment to init_fp_cache() saying it assumes/requires the caller to provide synchronization?

PR #5359, opened 2022-02-16
i#4134: Add atomic increment for drbbdup inits

johnfxgalea:
Makes increment/decrement operations atomic, using dr_atomic_add32_return_sum(), when tracking number of drbbdup inits. (Note, atm, only one drbbdup init is allowed).

Issue: #4134

### DIFF
--- a/ext/drbbdup/drbbdup.h
+++ b/ext/drbbdup/drbbdup.h
@@ -64,10 +64,11 @@ typedef enum {
     DRBBDUP_ERROR_INVALID_OPND,            /**< Operation failed: invalid case opnd. */
     DRBBDUP_ERROR_CASE_ALREADY_REGISTERED, /**< Operation failed: already registered. */
     DRBBDUP_ERROR_CASE_LIMIT_REACHED,      /**< Operation failed: case limit reached. */
-    DRBBDUP_ERROR_ALREADY_INITIALISED,     /**< DRBBDUP can only be initialised once. */
-    DRBBDUP_ERROR,                         /**< Operation failed. */
-    DRBBDUP_ERROR_UNSET_FEATURE,           /**< Operation failed: feature not set. */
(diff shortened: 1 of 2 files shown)


derekbruening (approved):

ext/drbbdup/drbbdup.h
```diff
@@ -64,10 +64,11 @@ typedef enum {
     DRBBDUP_ERROR_INVALID_OPND,            /**< Operation failed: invalid case opnd. */
     DRBBDUP_ERROR_CASE_ALREADY_REGISTERED, /**< Operation failed: already registered. */
     DRBBDUP_ERROR_CASE_LIMIT_REACHED,      /**< Operation failed: case limit reached. */
-    DRBBDUP_ERROR_ALREADY_INITIALISED,     /**< DRBBDUP can only be initialised once. */
-    DRBBDUP_ERROR,                         /**< Operation failed. */
-    DRBBDUP_ERROR_UNSET_FEATURE,           /**< Operation failed: feature not set. */
-    DRBBDUP_ERROR_NOT_INITIALIZED,         /**< Operation failed: not initialized. */
+    DRBBDUP_ERROR_ALREADY_INITIALISED, /**< DRBBDUP can only be initialised once. This is
```
abhinav92003, line 67: The comment indentation seems off. But the tests didn't complain, so i assume this is clang.

abhinav92003 (approved):

ext/drbbdup/drbbdup.h
johnfxgalea: Yeah, it is clang-format.

derekbruening: Note that officially we would prefer to move this to a whole-line beforehand comment to make it "prettier"; the same for the comment on line 1758 of drbbdup.c where officially we prefer the `*/` on its own line.  But since there is no way to make clang-format enforce that I've given up on that as a style rule.

PR #5361, opened 2022-02-16
i#5356 bbdup-wrap: Add extended drbbdup instrumentation callback

derekbruening:
Adds a new field to the drbbdup options that points at an
instrumentation handler that takes in the "for_trace" and
"translating" parameters and returns dr_emit_flags_t.  Either the
original or this one must be set.

Adds a simple sanity check of the parameters.  Testing the return
value is difficult, unfortunately.

Issue: #5356, #3995, #4134

### DIFF
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -129,7 +129,7 @@ changes:
  - Nothing yet (placeholder).
 
 Further non-compatibility-affecting changes include:
- - Added a new field instrument_instr_ex to #drbbdup_options_t.
+ - Added new fields analyze_case_ex and instrument_instr_ex to #drbbdup_options_t.
 
 The changes between version 9.0.1 and 9.0.0 include the following compatibility
 changes:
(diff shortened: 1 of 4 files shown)


johnfxgalea:
@derekbruening Is this ready to be reviewed?

derekbruening:
> @derekbruening Is this ready to be reviewed?

Yes.  The arm failure is b/c traces are (still) not implemented on arm: will make that check x86-only.

johnfxgalea (approved):
LGTM

derekbruening:
Wondering whether we should do the same thing for the analyze_case callback, since in PR #5357 I'm having to call the drwrap drmgr analyze event from the drbbdup one.  drwrap doesn't do anything in its analysis event...so either:

- A: Remove drwrap's analyze event
- B: Add an analyze_case_ex here which takes the extra params
- C: Keep the docs promising drwrap ignores the params: this seems the worst choice

johnfxgalea:
A and B both sound good to me for now, with B seeming the best long term solution.

derekbruening:
I went ahead and put in B in this same PR since it is so similar to the existing changes.  Did you want to take a look?

johnfxgalea (approved):

Issue #5362, opened 2022-02-16
Crash due to stack overflow caused by null sanitizer


abhinav92003:
Crashes were observed on a proprietary app in a build where the null sanitizer was enabled. This check causes a huge ~7x increase in code, and also in stack usage as each pointer is stored and retrieved from a local on the stack as part of an is-null check. This increases stack space usage on the DR stack (primary and signal both), and the app stack too when using the app start/stop interface.

We can probably disable the null sanitizer for the functions that are problematic (e.g. `dump_global_stats` which has a  lot of pointer derefereces (9 for each stat), which leads to a huge 127KB frame.

derekbruening:
It's sounding like this issue was not actually fixed?

abhinav92003:
Right. 

The fixes made in PR #5364 reduce stack space requirement by a lot (~147KB for `dump_local_stats`, ~127KB for `dump_global_stats`), but we still had to increase stack provisioning to 64K in the proprietary app to make it work. So other users will also need to increase stack space provisioning if they try to run DR with null sanitizer. The defaults for `stack_size` and `signal_stack_size` are 24K: https://github.com/DynamoRIO/dynamorio/blob/f417f41ae358fe5c4f5ce28e49124d86d6e0109a/core/optionsx.h#L828

derekbruening:
> The defaults for stack_size and signal_stack_size are 24K:

That's for plain DR.  With a client the default -stack_size is 56K and -signal_stack_size 32K.

derekbruening:
Historically we always tried to have the debug build stack sizes no larger than release build, to give confidence the release-build sizes were enough: if debug build with larger frame sizes and more tests and checks has no problem, then release build should have no problem.  This is why I've been reluctant to increase just debug.  But if we have to choose between:

A) Make debug stack sizes larger than release
B) Declare no support for the null sanitizer in core DR and pass a compiler flag to turn it off

The choice is less clear?

OTOH: Maybe the null sanitizer value is not all that high: a null de-ref will crash immediately in any case so what is it giving us?  Maybe it is not worth compromising on anything to support it.

PR #5364, opened 2022-02-17
i#5362: Disable null sanitizer for stats dumping functions.

abhinav92003:
dump_global_stats and dump_thread_stats print stats to logs. They have multiple
pointer dereferences for each stat. The null sanitizer in UBSan adds an is-null
check before each pointer dereference. As part of this, it stores and retrieves
the pointer from a location on the current stack frame. Each pointer dereference
uses a different location on the stack. This causes a huge increase in the
frame size for functions like dump_global_stats and dump_thread_stats (127 KB
and 147 KB respectively observed) which have a lot of pointer dereferences. We
annotate such functions with the added annotation to disable the null sanitizer
check.
https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#available-checks

Tried adding the annotation to various other functions too, but the gains were
relatively low:
dump_global_stats: 127KB to 8B
dump_local_stats: 147KB to 16B
build_bb_ilist: 12.8KB to 1KB
instr_encode_arch: 2.3KB to 0.5KB
common_heap_alloc: 5KB to 0.7KB
recreate_app_state_internal: 8.6KB to 5.1KB

Also fixes a null pointer dereference in an assert statement.

The null sanitizer fix is more relevant for when we want DR logs, as the major
stack space savings come from the dump_*_stats functions.

Fixes: #5362

### DIFF
--- a/core/lib/globals_api.h
+++ b/core/lib/globals_api.h
@@ -156,6 +156,16 @@ typedef char bool;
 #    define INLINE_FORCED inline
 #    define WEAK __attribute__((weak))
 #    define NOINLINE __attribute__((noinline))
+/* The null sanitizer adds is-null checks for pointer dereferences. As part
+ * of this, it stores and retrieves pointers from the stack frame. Each
+ * pointer dereference uses a different stack location. So, if there are too
+ * many pointer dereferences in a function (like dump_global_stats and
+ * dump_thread_stats) it will increase the size of the stack frame a lot
+ * (127KB and 147KB respectively for the mentioned examples) when the null
+ * sanitizer is enabled. The following macro lets us disable this check for
+ * methods annotated with it.
+ */
+#    define DISABLE_NULL_SANITIZER __attribute__((no_sanitize("null")))
 #endif
 
 /* We want a consistent size so we stay away from MAX_PATH.
--- a/core/utils.h
+++ b/core/utils.h
@@ -1427,7 +1427,8 @@ extern mutex_t do_threshold_mutex;
          */                                                                            \
         ASSERT(                                                                        \
             (try_pointer) == &global_try_except || (try_pointer) == NULL ||            \
-            (try_pointer) == &get_thread_private_dcontext()->try_except ||             \
+            (get_thread_private_dcontext() != NULL &&                                  \
+             (try_pointer) == &get_thread_private_dcontext()->try_except) ||           \
             (try_pointer) == /* A currently-native thread: */                          \
                 &thread_lookup(IF_UNIX_ELSE(get_sys_thread_id(), d_r_get_thread_id())) \
                      ->dcontext->try_except);                                          \
--- a/core/utils.c
+++ b/core/utils.c
@@ -3042,6 +3042,7 @@ stats_thread_exit(dcontext_t *dcontext)
     }
 }
 
+DISABLE_NULL_SANITIZER
 void
 dump_thread_stats(dcontext_t *dcontext, bool raw)
 {
@@ -3100,6 +3101,7 @@ dump_thread_stats(dcontext_t *dcontext, bool raw)
 #    endif
 }
 
+DISABLE_NULL_SANITIZER
 void
 dump_global_stats(bool raw)
 {


abhinav92003:
Created #5366 for the AArch64 failure. Ran it 100 times locally, and it worked fine. Seems flaky, unrelated to this PR.

core/lib/globals_api.h
```diff
@@ -156,6 +157,27 @@ typedef char bool;
 #    define INLINE_FORCED inline
 #    define WEAK __attribute__((weak))
 #    define NOINLINE __attribute__((noinline))
+/* As per https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005fattribute.html,
+ * we need to first check whether __has_attribute is defined, to ensure
+ * portability. We saw issues in the Android-ARM cross compile without this.
+ */
+#    if defined __has_attribute
+#        if __has_attribute(__no_sanitize__) && defined(USE_FNOSANITIZE_NULL)
```
derekbruening, line 165: As the _api.h suffix denotes, this is a public header we give to clients.  These defines seem like something we only need internally.  I think it would be better to not put all this in this exposed header and instead keep it in globals.h or another internal header.

make/configure.cmake.h
```diff
@@ -264,6 +265,10 @@
 # ifdef HAVE_FVISIBILITY
 #  define USE_VISIBILITY_ATTRIBUTES
 # endif
+
+# ifdef HAVE_FNOSANITIZE_NULL
+#  define USE_FNOSANITIZE_NULL
```
derekbruening, line 270: We could remove this define and just use the HAVE_ one above: seems simpler.  I'm guessing the USE_VISIBILITY_ATTRIBUTES was already there with that name historically and the conversion to cmake ended up with the extra define; I don't think we should mirror that for something new.

CMakeLists.txt
```diff
@@ -1029,6 +1029,13 @@ if (UNIX)
     set(HAVE_FVISIBILITY ON)
   endif (VISIBILITY)
 
+  # Check for -fno-sanitize=null
+  CHECK_C_COMPILER_FLAG("-fno-sanitize=null" HAVE_FNOSANITIZE_NULL_FLAG)
+
+  if (HAVE_FNOSANITIZE_NULL_FLAG)
+    set(HAVE_FNOSANITIZE_NULL ON)
```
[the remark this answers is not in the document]
abhinav92003: `HAVE_FNOSANITIZE_NULL_FLAG` is a bool. I thought it's better if the CMake variable we export is either defined or undefined.

core/lib/globals_api.h
abhinav92003: > These defines seem like something we only need internally

Well, I thought `DISABLE_NULL_SANITIZER` could be useful to clients/extension libraries as well.

core/utils.h
```diff
@@ -1427,7 +1427,8 @@ extern mutex_t do_threshold_mutex;
          */                                                                            \
         ASSERT(                                                                        \
             (try_pointer) == &global_try_except || (try_pointer) == NULL ||            \
-            (try_pointer) == &get_thread_private_dcontext()->try_except ||             \
+            (get_thread_private_dcontext() != NULL &&                                  \
+             (try_pointer) == &get_thread_private_dcontext()->try_except) ||           \
```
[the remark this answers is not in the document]
abhinav92003: When the null sanitizer is enabled, this raises an exception.

make/configure.cmake.h
abhinav92003: I thought the purpose of creating a new symbol was to limit the usages of `HAVE_FNOSANITIZE_NULL` (which is shared with cmake) to this `configure.cmake.h` file.

abhinav92003: I see that we don't do the same for other symbols in `configure.cmake.h`. Okay, makes sense to not create a new one here.

core/lib/globals_api.h
abhinav92003: Moved it to `globals.h`.

make/configure.cmake.h
abhinav92003: Done.

core/utils.h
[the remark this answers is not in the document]
abhinav92003: The crash without this fix:
```
Thread 2 "<redacted>" received signal SIGILL, Illegal instruction.
[Switching to Thread 0x7ffff793d700 (LWP 1006714)]
0x0000555572607565 in copy_frame_to_stack (dcontext=0x7ffdaedd5840, info=0x7ffdaee6d148, sig=12, frame=0x7ffdaee88378, sp=0x7ffff793b168 "P\207\347\367\377\177", from_pending=0 '\000') at third_party/dynamorio/trunk/core/unix/signal.c:3558
3558	        TRY_EXCEPT(

```
Disassembly for this part, notice the null check for the result of `get_thread_private_dcontext`:
```
   0x55557260754f <copy_frame_to_stack+1423>:	callq  0x5555725811c0 <get_thread_private_dcontext>
   0x555572607554 <copy_frame_to_stack+1428>:	mov    %rax,-0x188(%rbp)
   0x55557260755b <copy_frame_to_stack+1435>:	cmp    $0x0,%rax
   0x55557260755f <copy_frame_to_stack+1439>:	jne    0x55557260756a <copy_frame_to_stack+1450>
=> 0x555572607565 <copy_frame_to_stack+1445>:	ud1    0x16(%eax),%eax
   0x55557260756a <copy_frame_to_stack+1450>:	mov    -0x190(%rbp),%rax

```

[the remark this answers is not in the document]
abhinav92003: From the disassembly above, I see that you are correct about the result of `get_thread_private_dcontext()` not getting dereferenced, but the null check is still applied to it it seems.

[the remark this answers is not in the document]
abhinav92003: Added a comment to clarify this.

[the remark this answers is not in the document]
derekbruening: Looks like a bug in the null sanitizer to me.

core/globals.h
```diff
@@ -109,6 +109,32 @@
 #    define DYNAMORIO_EXPORT
 #endif
 
+#ifdef WINDOWS
```
derekbruening, line 112: This seems out of place splitting up the import/export defines here (some above and some below): how about moving it at least below DR_UNS_API

CMakeLists.txt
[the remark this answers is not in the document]
abhinav92003: For the symbol used in code, I think we want it to be either defined or undefined. This means we cannot use the result of CHECK_C_COMPILER_FLAG directly, which is a bool. We'll need to use a second variable somewhere. So you're suggesting that I add the second var in `make/configure.cmake.h` instead of here?

core/utils.h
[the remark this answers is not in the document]
abhinav92003: Probably. We'll still need to add this fix though. The comment should help readers.

CMakeLists.txt
```diff
@@ -1029,6 +1029,13 @@ if (UNIX)
     set(HAVE_FVISIBILITY ON)
   endif (VISIBILITY)
 
+  # Check for -fno-sanitize=null
+  CHECK_C_COMPILER_FLAG("-fno-sanitize=null" HAVE_FNOSANITIZE_NULL)
+
+  if (HAVE_FNOSANITIZE_NULL)
+    set(USE_FNOSANITIZE_NULL ON)
```
derekbruening, line 1036: Where did the unresolved comment here go?  Still pushing to remove this redundant variable.

Generally we want ON or OFF for cmake variables. The #cmakedefines operate with that. We don't normally want to deal with cmake vars that might be undefined.

Other uses of CHECK_C_COMPILER_FLAG do not turn around and make a new variable.

core/utils.h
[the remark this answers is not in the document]
derekbruening: If this were release-build critical-path code I would push to disable the broken sanitizer rather than add unnecessary overhead.  Since it's debug this seem fine.

CMakeLists.txt
abhinav92003: Hmmm. I see `HAVE_TYPELIMITS_CONTROL` is actually used directly. Perhaps I'm wrong in understanding the result type of `CHECK_C_COMPILER_FLAG`. Making the change.

CMakeLists.txt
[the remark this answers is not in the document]
derekbruening: If you mean in C code, yes it has to be separate.  Having a new cmake variable here has no bearing on that: C code doesn't see this `HAVE_FNOSANITIZE_NULL` variable.  I don't see any benefit to this 2nd cmake variable.

I would expect this to be like `HAVE_TYPELIMITS_CONTROL`.

CMakeLists.txt
derekbruening: `HAVE_TYPELIMITS_CONTROL` has a `#cmakedefine` in configure.h which turns it into a C preprocessor define and lets C code use it.  `#cmakedefine` acts based on CMake boolean rules: if `OFF` or `*-NOTFOUND` it is omitted; else it is defined.

CMakeLists.txt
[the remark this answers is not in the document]
abhinav92003: Just a single symbol everywhere, works. Like `HAVE_TYPELIMITS_CONTROL`.

core/globals.h
abhinav92003: Done.

CMakeLists.txt
abhinav92003: Now I use just a single symbol everywhere.

> if OFF or *-NOTFOUND it is omitted; else it is defined.

I had some trouble figuring this out. To test the case when the feature is absent, I set the CHECK_C_COMPILER_FLAG arg to a flag I know isn't supported (some random string), and did a `make -j`. But the HAVE_FNOSANITIZE_NULL constant was still getting defined. This caused me to wrongly assume that if the CHECK_C_COMPILER_FLAG result is false, it would still define the macro, but maybe to "false" or something.

I later figured out that, for my CHECK_C_COMPILER_FLAG change to take effect, I need to actually delete the build dir and run the `cmake` and `make -j` commands again! (just `make clean` or `cmake` doesn't do it).

derekbruening (approved):

Issue #5366, opened 2022-02-17
[AArch64] Release-build hang in tool.drcacheoff.max-global 


abhinav92003:
tool.drcacheoff.max-global failed on AArch64 in a CI run on PR #5364. I tried running it 100 times locally and it worked fine. Looking at logs, they seem to be cut short. Maybe it crashed?
Logs: http://139.178.84.19:8080/job/DynamoRIO-AArch64-Precommit/1334/console

derekbruening:
This is a timeout (90s is the limit) and so likely a hang:
```
66/66 Test #212: code_api|tool.drcacheoff.max-global .............................***Failed   90.03 sec
```

This might be the plain-DR hang I found in https://github.com/DynamoRIO/dynamorio/issues/4928#issuecomment-1043154222 which was the cause of dr$sim -offline hangs in my runs.  With PR #5367 -offline does not hang (ran 20K in a row) so I would be willing to bet that this is fixed by that PR.

Issue #4928, opened 2021-05-27
tool.drcachesim threads tests fail on AArch64


AssadHashmi:
Migrating the AArch64 Jenkins CI to an Ampere Altera machine caused a regression in the `tool.drcachesim.threads` and `tool.drcachesim.TLB-threads` tests failing with timeout messages.

derekbruening:
I think there are multiple hangs here, in fact.  I found one with plain DR and fixed it but dr$sim online still hangs.  Will add more details later.  The one I found is related to #2502.

derekbruening:
Looks like the libc write() contains calls to pthread cancel code.  This pthread code must eventually involve a futex, and it must interact with application resources (we aren't currently able to separate pthread state or support a private libpthread: #956).

The simplest solution is to have the private loader redirect `write` to `os_write`.  May was well do the same for `read`; and may as well redirect `open` to `os_open_protected` (thought we were already doing that: I guess it was just `dr_open_file`) for file descriptor isolation.

With that in place: there is no hang!

I also tested other related issues:

+ #4928: drcachesim.threads and drcachesim.TLB-threads
+ #2417: drcachesim.coherence
+ #4954: drcachesim.threads-with-config-file

```
derek@dynamorio:~/dr/build_rel$ for i in sim.threads\$ sim.TLB-threads sim.coherence sim.threads-with; do echo $i; ctest --repeat-until-fail 20000 -R $i > RUN-$i 2>&1; done
sim.threads$
sim.TLB-threads
sim.coherence
sim.threads-with
derek@dynamorio:~/dr/build_rel$ grep -c Passed RUN-*
RUN-sim.coherence:20000
RUN-sim.threads$:20000
RUN-sim.threads-with:20000
RUN-sim.TLB-threads:20000
derek@dynamorio:~/dr/build_rel$ grep failed RUN-*
RUN-sim.coherence:100% tests passed, 0 tests failed out of 1
RUN-sim.threads$:100% tests passed, 0 tests failed out of 1
RUN-sim.threads-with:100% tests passed, 0 tests failed out of 1
RUN-sim.TLB-threads:100% tests passed, 0 tests failed out of 1
```

Looks like PR #5367 plus this `write` redirect solve all the hangs!

edeiana:
code_api|tool.drcachesim.TLB-threads flaky on #6805

PR #5367, opened 2022-02-17
i#2502 locks: Fix bug in ATOMIC_MAX on AArchXX

derekbruening:
Adds a missing load-acquire in a loop in ATOMIC_MAX that causes hangs
in RSTATS_ADD_PEAK.

Tested on the client.annotation-concurrency test on AArch64 where
without this fix it hangs under release-build plain DR every ~100
runs.  With the fix, there is no hang in 20K runs.

Issue: #2502, #4928


derekbruening:
I am highly confident that this fixes #5366.  Without this fix I saw hangs in dr$sim -offline every ~100-300 runs; with the fix, no hangs in 20K runs.  That is similar to the plain DR rates.  dr$sim online however also has a different, more frequent hang that #4928 covers: have not figured that one out yet.

Issue #4279, opened 2020-05-01
Please follow the guidelines in our wiki at https://github.com/DynamoRIO/dynamorio/wiki/Bug-Reporting and use one of the CRASH, APP CRASH, HANG, or ASSERT keywords.


Desperado17:
I tested DynamoRIO-AArch64-Linux-8.0.0-1 and drcov with a large application of ca. 150 threads. The program seems to get stuck at certain points. Attaching with gdb shows that the program still runs but contends a mutex that has to do with fragment writing/reading:

DynamoRIO-AArch64-Linux-8.0.0-1

/media/mp000/DynamoRIO-AArch64-Linux-8.0.0-1/bin64/drrun -root /media/mp000/DynamoRIO-AArch64-Linux-8.0.0-1 -t drcov -logdir /media/mp000/ -dump_binary -- /bin/myprogram
<Application /bin/busybox (4023). DrCov internal crash at PC 0x00000000710c6694. Please report this at http://dynamorio.org/issues. Program aborted.
Received SIGSEGV at pc 0x00000000710c6694 in thread 4023
Base: 0x0000000071000000
Registers: eflags=0x0000000040000000
version 8.0.0, build 1

The readers ( dozens ):
Thread 42 (LWP 2650):
#0 0x000000007112fd08 in dynamorio_syscall ()
#1 0x000000007105b2a8 in d_r_read_lock (rw=0x7f0fefe5ac) at /home/travis/build/DynamoRIO/dynamorio/core/utils.c:1185
#2 0x0000000071048730 in fragment_lookup_type (lookup_flags=15, tag=0x7f0c28431c "\342\003", dcontext=0x7f110859e0) at /home/travis/build/DynamoRIO/dynamorio/core/fragment.c:2738
#3 fragment_lookup (tag=0x7f0c28431c "\342\003", dcontext=0x7f110859e0) at /home/travis/build/DynamoRIO/dynamorio/core/fragment.c:2756
#4 fragment_lookup_fine_and_coarse (dcontext=dcontext@entry=0x7f110859e0, tag=0x7f0c28431c "\342\003", wrapper=wrapper@entry=0x7f10c44fc8, last_exit=0x7f10e09a48) at /home/travis/build/DynamoRIO/dynamorio/core/fragment.c:8109
[...]

PR #5370, opened 2022-02-17
i#2502 locks: Add acquires to bare loads in lock routines

derekbruening:
Adds load-acquire semantics to a number of bare loads (i.e., not
inside a mutex) in DR's various lock routines.

This includes a key missing load-acquire in the spinlock loop in
d_r_mutex_lock_app() which might help explain some performance issues
such as #4279.

Since some of these are in the utils.h header, and we have ordering
dependencies between utils.h and arch_exports.h, this required
splitting the atomic defines out of arch/arch_exports.h into a new
file arch/atomic_exports.h.  Since those atomic defines use ASSERT
which references dynamo_options, this further required splitting the
option struct out of options.h into options_struct.h.

Issue: #2502, #4928, #4279


derekbruening:
The header code that was moved was not modified.  All the not-just-moved modifications (the inserted acquires) are in utils.c and utils.h.

Unfortunately there isn't a simple way to test these things.  I ran dr$sim -offline thousands of times on the Jenkins machine and there are no hangs so this doesn't make anything worse; there is still a hang in dr$sim online (#4928).  I don't have a setup to easily test performance losses or gains; we'll try to ask on #4279 to do that once this is in to see if it helped there.

abhinav92003 (approved):

PR #5372, opened 2022-02-17
Fix crash with disabled global stats.

abhinav92003:
When global stats are disabled, we do not keep fragment count. This
causes us to try triggering a reset even when reset_at_fragment_count
is zero.

Failure without this fix:
In debug build:
```
 ./bin64/drrun -no_global_stats  -- suite/tests/bin/allasm_scattergather

<Application dynamorio/suite/tests/bin/allasm_scattergather (1281891).  Internal Error: DynamoRIO debug check failure: dynamorio/core/fragment.c:2376 INTERNAL_OPTION(reset_at_fragment_count) != 0

```

In release build:
```
 ./bin64/drrun -no_global_rstats  -- suite/tests/bin/allasm_scattergather

<Goes into a reset infinite loop>

```

Also adds a long suite test that runs the simple app without global stats.
Tested locally using
```
cmake -DDEBUG=ON   -DBUILD_TESTS=ON -DRUN_LONG_SUITE=ON ..
```


derekbruening:
@abhinav92003 is this still valid?

abhinav92003:
> @abhinav92003 is this still valid?

Well, the core issue is probably still there. I can't remember why I abandoned working on this PR; probably got distracted by whatever I was originally working on. It'll be good to fix this; I'll resume my work here.

core/fragment.c
```diff
@@ -2369,7 +2369,7 @@ fragment_create(dcontext_t *dcontext, app_pc tag, int body_size, int direct_exit
         if (!fragment_lookup_deleted(dcontext, tag) && !TEST(FRAG_COARSE_GRAIN, flags))
             STATS_INC(num_unique_fragments);
     });
-    if (d_r_stats != NULL &&
+    if (GLOBAL_STATS_ON() &&
```
derekbruening, line 2372: Maybe this PR should add a long-suite test that disables global stats.

derekbruening (approved):

core/fragment.c
abhinav92003: Done.

derekbruening (approved):

PR #5374, opened 2022-02-18
i#5356 bbdup-drwrap: Add drwrap cleanup-only support

derekbruening:
To best handle wrapping functions with drwrap in only some drbbdup
instrumentation cases, drwrap now provides a cleanup-only instrumentation
feature which is called in drbbdup cases without wrapping.  This
ensures that key cleanup like restoring replaced return addresses is
performed without worrying about sychronously clearing state in all
threads on a drbbdup case transition.

To avoid having 2 analysis functions on top of the now 2 insert
functions, removes the analysis code completely as it was empty.

Adds a test which mis-behaves without this feature.

Issue: #5356, #3995


derekbruening:
Removing the analysis routines means the user_data is always NULL.  Maybe those params should just be removed from the drwrap_invoke_insert_* functions?  drwrap doesn't use them.

ext/drwrap/drwrap.c
```diff
@@ -2345,9 +2348,22 @@ drwrap_after_callee(app_pc retaddr, reg_t xsp)
     }
 }
 
+static void
+drwrap_after_callee_cleanup_only(app_pc retaddr, reg_t xsp)
+{
+    void *drcontext = dr_get_current_drcontext();
+    per_thread_t *pt = (per_thread_t *)drmgr_get_tls_field(drcontext, tls_idx);
+    /* To reduce overhead for non-drbbdup cases we avoid an extra arg to a common
```
abhinav92003, line 2356: Wondering why we didn't add the arg, it's just `drwrap_after_callee_func` and `drwrap_after_callee` right?

ext/drwrap/drwrap.h
```diff
@@ -88,34 +88,41 @@ drwrap_exit(void);
...
- * drmgr_register_bb_instrumentation_event() or if using drbbdup with
- * the instrument_instr_ex field in #drbbdup_options_t).  It is up to the user
- * to control the ordering, since the priority #DRMGR_PRIORITY_INSERT_DRWRAP
- * will not apply.
+ * When #drwrap_global_flags_t #DRWRAP_INVERT_CONTROL is set and the
+ * user is using multiple types of instrumentation such as with
+ * drbbdup, the user must call this function for each instrumentation
+ * case where function wrapping should not be enabled.  This will
+ * insert required cleanup for instrumentation cases changing in the
+ * middle of a wrapped function.  It will not invoke any wrapped
```
abhinav92003, line 114: Maybe add that there's an opportunity for drbbdup to switch instrumentation  modes because the post-call callback is in a different fragment.

suite/tests/client-interface/drbbdup-drwrap-test.appdll.c
```diff
@@ -56,7 +48,33 @@ run_tests(void)
...
+}
+
+void
+run_tests(void)
+{
+    /* First, a regular pre-and-post-wrapped instance. */
+    print("first wrapme returned %d\n", wrapme(1));
+    /* Now we'll have a pre but not post b/c we'll swap in the middle. */
     print("second wrapme returned %d\n", wrapme(2));
+    /* Now we swap back and will have no pre or post. */
```
abhinav92003, line 71: So when we swap first, we only have a pre, but not post. When we swap back, we don't have any; I'm wondering if a user might expect a post, because the instrumentation changed during the wrapped-func.

I'm also wondering if there can be cases where the user is keeping some state between the pre and post callbacks, and whether it's better to always do both or none. It may be difficult to do that though.

suite/tests/client-interface/drbbdup-drwrap-test.dll.c
```diff
@@ -125,7 +127,12 @@ event_module_load(void *drcontext, const module_data_t *mod, bool loaded)
         return;
     app_pc target = (app_pc)dr_get_proc_address(mod->handle, "wrapme");
     CHECK(target != NULL, "cannot find lib export");
-    bool res = drwrap_wrap(target, wrap_pre, wrap_post);
+    int flags = 0;
+#if 0 /* XXX i#5356: Enable once drbbdup handles the no-app-pc sentinel block. */
```
abhinav92003, line 131: IIRC `DRWRAP_REPLACE_RETADDR` is an important feature. Maybe change the XXX to a TODO?

suite/tests/client-interface/drbbdup-drwrap-test.appdll.c
```diff
@@ -56,7 +48,33 @@ run_tests(void)
...
+    print("third wrapme returned %d\n", wrapme(2));
+    /* If we did not clean up drwrap state, we would see a pre and *two* posts as drwrap
+     * tries to catch up to the interrupted state.  With the proper cleanup, we have
```
abhinav92003, line 74: Do you mean there would be two posts triggered by the fourth wrapme below? Trying to understand how drwrap would remember to "catch up". (Didn't investigate too much on this, so  maybe I'm  missing something obvious).

ext/drwrap/drwrap.c
```diff
@@ -1122,7 +1120,8 @@ drwrap_set_global_flags(drwrap_global_flags_t flags)
     dr_recurlock_lock(wrap_lock);
     if (dr_atomic_load32(&drwrap_init_count) > 0 &&
         /* After drwrap_init() was called, control inversion cannot be changed. */
-        (flags & DRWRAP_INVERT_CONTROL) != (global_flags & DRWRAP_INVERT_CONTROL)) {
+        TEST(DRWRAP_INVERT_CONTROL, flags) &&
```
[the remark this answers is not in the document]
derekbruening: The original was incorrect, because drwrap_set_global_flags() only adds new flags: it does not replace the set of flags with the passed-in value but just ORs the passed-in value.  (Maybe it should have been called drwrap_add_global_flag() and a separate drwrap_remove_global_flag() added; right now you can never remove a flag.)

ext/drwrap/drwrap.c
```diff
@@ -877,19 +879,15 @@ static dr_emit_flags_t
 drwrap_event_bb_app2app(void *drcontext, void *tag, instrlist_t *bb, bool for_trace,
                         bool translating);
 
-static dr_emit_flags_t
-drwrap_event_bb_analysis(void *drcontext, void *tag, instrlist_t *bb, bool for_trace,
```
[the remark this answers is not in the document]
derekbruening: You mean the drwrap_invoke_analysis?  That was present in the code base for just a few days in between releases so we are not worried about breaking users.

ext/drwrap/drwrap.c
derekbruening: Overhead is important for these calls; would prefer to not add anything to the common case if we can help it.

ext/drwrap/drwrap.c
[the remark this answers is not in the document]
derekbruening: Added a comment to drwrap_set_global_flags() to clarify this.

ext/drwrap/drwrap.h
```diff
@@ -796,11 +803,14 @@ typedef enum {
     DRWRAP_FAST_CLEANCALLS = 0x08,
     /**
      * This flag must only be set before calling drwrap_init().  If set, drwrap will not
-     * register for the drmgr analysis or insertion events.  The user must instead
-     * explicitly invoke drwrap_invoke_analysis() and drwrap_invoke_insert() from its
-     * own handler for those events.  This "inverted control" mode is provided for
+     * register for the drmgr insertion event.  The user must instead explicitly invoke
+     * drwrap_invoke_insert() or and drwrap_invoke_insert_cleanup_only() from its own
```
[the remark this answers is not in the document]
derekbruening: Done.

api/docs/release.dox
```diff
@@ -130,6 +130,8 @@ changes:
```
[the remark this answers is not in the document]
derekbruening: s/drwrap provides/drwrap now provides/

ext/drwrap/drwrap.c
```diff
@@ -877,19 +879,15 @@ static dr_emit_flags_t
...
-                         bool translating, OUT void **user_data);
-
 static dr_emit_flags_t
 drwrap_event_bb_insert(void *drcontext, void *tag, instrlist_t *bb, instr_t *inst,
                        bool for_trace, bool translating, void *user_data);
 
-/* This version takes a separate "instr" and "where" for use with drbbdup. */
+/* This version takes a separate "instr" and "where", + cleanup_only, for drbbdup use. */
```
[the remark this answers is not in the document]
derekbruening: Done.  (Now goes onto next line.)

ext/drwrap/drwrap.c
```diff
@@ -1141,24 +1140,25 @@ drwrap_set_global_flags(drwrap_global_flags_t flags)
...
-drwrap_invoke_analysis(void *drcontext, void *tag, instrlist_t *bb, bool for_trace,
-                       bool translating, OUT void **user_data)
+drwrap_invoke_insert(void *drcontext, void *tag, instrlist_t *bb, instr_t *inst,
+                     instr_t *where, bool for_trace, bool translating, void *user_data)
 {
     ASSERT(TEST(DRWRAP_INVERT_CONTROL, global_flags),
-           "must set DRWRAP_INVERT_CONTROL to call drwrap_invoke_analysis");
-    return drwrap_event_bb_analysis(drcontext, tag, bb, for_trace, translating,
-                                    user_data);
+           "must set DRWRAP_INVERT_CONTROL to call drwrap_invoke_app2app");
```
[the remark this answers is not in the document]
derekbruening: Done, and the other one.

ext/drwrap/drwrap.h
derekbruening: I don't follow?  Do you mean a different drbbdup case, rather than a different fragment?

suite/tests/client-interface/drbbdup-drwrap-test.dll.c
derekbruening: Done, but probably a moot point since days later there is a PR to solve it which might make it in before this one: PR #5381.

suite/tests/client-interface/drbbdup-drwrap-test.appdll.c
derekbruening: On a post-call point, drwrap always checks the stack value to see whether a call was skipped because of a tail-call or a longjmp or other actions.  That is how it "catches up": it invokes the post-call handlers for post-call points that implicitly happened on the skip.

suite/tests/client-interface/drbbdup-drwrap-test.appdll.c
derekbruening: It is not feasible to do none since you wouldn't know in the pre whether the phase is going to change.

For a user who wants the post to happen regardless of the phase, they would just not use this inverted control mode.  (Not 100% sure everything works though with drwrap using regular drmgr compare to the user using drbbdup but I think it does: drwrap will inserts its calls *before* the drbbdup dispatch so they'll happen for all cases.  I don't think it's worth our time right now to go make sure that works; wait for someone who needs it.)

ext/drwrap/drwrap.c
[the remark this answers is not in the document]
derekbruening: That check is broken.  Users today, including our own code, make multiple calls to drwrap_set_global_flags() adding one new flag at a time without specifying the other flags in each call.  Nobody is passing the full set they expect to be there, so expecting one particular flag to be in every call is always going to fail over and over.

derekbruening:
> > Removing the analysis routines means the user_data is always NULL. Maybe those params should just be removed from the drwrap_invoke_insert_* functions? drwrap doesn't use them.
> 
> drwrap passes it to the callbacks in `drwrap_in_callee`. It also makes the effort to keep a separate copy of it for each wrap level. I wonder if we're removing a useful feature by not supporting analysis functions and user data. I haven't thought of an affected use-case yet though. Is there a migration plan for users that did use the analysis routines?

No, those user_data parameters are different: those are set by user pre-wrap routines and passed to user post-wrap routines, for keeping state across them.

No migration plan is needed because the inverted control invocation of drwrap's analysis routine was only added a few days ago.

I'm thinking of leaving the user_data parems to drwrap_invoke_insert* and documenting that they are ignored.  My logic is that if we did want to add back invocation of an analysis routine we then wouldn't have to change these signatures. though I think it's unlikely.

ext/drwrap/drwrap.c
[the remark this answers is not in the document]
abhinav92003: > adding one new flag at a time without specifying the other flags in each call. Nobody is passing the full set they expect to be there,

Oh okay.

ext/drwrap/drwrap.h
```diff
@@ -88,34 +88,41 @@ drwrap_exit(void);
...
+ * since the priority #DRMGR_PRIORITY_INSERT_DRWRAP will not apply.
  * The separate "where" handles cases such as with drbbdup's final app
  * instruction (which cannot be duplicated into each case) or with
  * emulation where the instruction "inst" to monitor is distinct from
  * the location "where" to insert instrumentation.
  */
 dr_emit_flags_t
-drwrap_invoke_insert(void *drcontext, void *tag, instrlist_t *bb, instr_t *inst,
-                     instr_t *where, bool for_trace, bool translating, void *user_data);
+drwrap_invoke_insert_cleanup_only(void *drcontext, void *tag, instrlist_t *bb,
```
abhinav92003, line 123: So the cleanup code is added always, even in the drbbdup instrumentation case where drwrap didn't wrap. I see that this works fine when the drbbdup case gets changed in the middle of a wrapped function; it allows us to cleanup the changes we made to wrap the function. But, is it okay to cleanup if we didn't wrap at all?

ext/drwrap/drwrap.h
abhinav92003: Just describing the problem in some more detail:

The drbbdup instrumentation case changes happen at the fragment boundary. As cleanup is part of a different fragment, the instrumentation case may change before we can complete cleanup.

abhinav92003 (approved):

ext/drwrap/drwrap.h
abhinav92003: Feel free to ignore if being more verbose doesn't add value here.

derekbruening: If the case changes to don't-wrap and then changes back to wrap all inside a wrapped callee: the full post-wrap callback plus cleanup would then be called.  I think that is what we want.

ext/drwrap/drwrap.h
derekbruening: Yes, the drwrap post-call handling checks whether it's in a target function or not.  It has to, to handle some other control flow reaching the same basic block that a retaddr targets.

ext/drwrap/drwrap.h
derekbruening: Adding a sentence about that.

PR #5377, opened 2022-02-18
i#5376: Fix drwrap reattach failing to clear flags

### LINKED ISSUE #5376: drwrap fails to clear its flags on re-attach
For statically-linked drwrap, on re-attach its `global_flags` is not reset, causing problems if a test checks the return value of `drwrap_set_global_flags` (or wants to use different flags).

Hit on the drcachesim burst tests with #3995.

derekbruening:
Fixes a drwrap reattach bug by resetting global_flags to 0.

Adds a test case to the drbbdup-drwrap-test.

Fixes #5376


derekbruening:
For load balancing: @johnfxgalea if you have a chance to look at this one

PR #5378, opened 2022-02-18
i#5368: Replace disrespectful terms "master" and "slave"

### LINKED ISSUE #5368: Replace disrespectful terms "Master" and "Slave"
This is to identify and replace the term "Master" from the codebase/documentation where possible.

bete0:
Replaces "master" and "slave" with "main" (or "manager") and "worker"
respectively where possible. Also renames files containing these terms
with the corresponding inclusive terms.

Fixes #5368

### DIFF
--- a/ext/drbbdup/drbbdup.c
+++ b/ext/drbbdup/drbbdup.c
@@ -30,7 +30,6 @@
  * DAMAGE.
  */
 
-#include "dr_defines.h"
 #include "dr_api.h"
 #include "drmgr.h"
 #include "drreg.h"


core/unix/signal.c
```diff
@@ -7760,7 +7760,7 @@ sig_detach_go_native(sig_detach_info_t *info)
     ASSERT_NOT_REACHED();
 }
 
-/* Sets this (slave) thread to detach by directly returning from the signal. */
+/* Sets this (worker) thread to detach by directly returning from the signal. */
```
derekbruening, line 7763: Hmm, to me "worker" implies a thread pool or something of threads dedicated to a task, while this is an app thread but is just not the one orchestrating the detach. Maybe "secondary" or "follower" instead?

ext/drmgr/CMakeLists.txt
```diff
@@ -70,7 +70,7 @@ configure_drmgr_target(drmgr)
 
 # Must be shared to ensure only one copy, since can be used by
 # the client and by multiple other libraries, some of which may
-# be shared: the whole point is to have one master drmgr.
+# be shared: the whole point is to have one main drmgr.
```
derekbruening, line 73: How about "unique"

ext/drreg/drreg.c
```diff
@@ -2478,7 +2478,7 @@ drreg_init(drreg_options_t *ops_in)
         return DRREG_ERROR_INVALID_PARAMETER;
 
     /* Instead of allowing only one drreg_init() and all other components to be
-     * passed in scratch regs by a master, which is not always an easy-to-use
+     * passed in scratch regs by a main, which is not always an easy-to-use
```
derekbruening, line 2481: Hmm to me it sounds off -- the original seems off too.  I would do `s/by a main//`

core/synch.c
```diff
@@ -2248,7 +2248,7 @@ detach_on_permanent_stack(bool internal, bool do_cleanup, dr_stats_t *drstats)
         dynamo_thread_exit_pre_client(my_dcontext, my_tr->id);
     }
 
-    LOG(GLOBAL, LOG_ALL, 1, "Detach: Letting slave threads go native\n");
+    LOG(GLOBAL, LOG_ALL, 1, "Detach: Letting worker threads go native\n");
```
derekbruening, line 2251: Like below, I would replace this with whatever is chosen for detach there: "secondary" or "follower"

derekbruening (approved):
Just a few suggestions for different replacements in some cases.

core/unix/signal.c
bete0: Done; changed to "secondary".

ext/drmgr/CMakeLists.txt
bete0: Done.

ext/drreg/drreg.c
bete0: Done.

core/synch.c
bete0: Done; changed to "secondary"

johnfxgalea:
We should also change the DR's default branch to main, modifying CI scripts accordingly.

derekbruening:
> > We should also change DR's default branch to main, modifying CI scripts accordingly (maybe separate PR?).
> 
> Agreed. We have had a discussion to do that in a way that won't negatively affect contributors and users.

I believe this complicates outstanding pull requests and other users and would have to be done carefully and probably with announcements.  I assumed we would at least first do a cleanup pass through the open pull requests (needs a cleanup pass in any case).

Issue #409, opened 2014-11-28
add event for app code modification or for code region being flushed


derekbruening:
_From [bruen...@google.com](https://code.google.com/u/109494838902877177630/) on March 08, 2011 10:13:46_

for clients to update data tied to application code, the module unload
event covers the most common cases, but for other code changes the
fragment_delete event is not usable enough.  it only passes the tag, it's
not always safe to decode forward in a no-longer-read-only region, and
clients don't necessarily know where the fragment ends, especially if a
trace.

DR should raise a new event on each whole region being flushed, for every
flush, regardless of the source of the flush, since code anywhere in the
region could change after removal from executable_areas (if writable,
anyway, but could be made writable later).

this would help solve these Dr. Memory issues:
issue #260: remove xl8_sharing_table and ignore_unaddr_table entries on code mod
issue #114: remove post_call_table and call_site_table entries on code mod

_Original issue: http://code.google.com/p/dynamorio/issues/detail?id=409_

PR #5381, opened 2022-02-19
i#5456: Add drbbdup support for meta bbs

johnfxgalea:
Adds drbbdup support for basic blocks consisting solely of meta instructions. The hash-table manager no longer uses  PCs of the first app instructions inside basic blocks as keys. Instead, tags of basic blocks are now used.

The PR also adds tests. 

Issue: #5356

### DIFF
--- a/ext/drbbdup/drbbdup.c
+++ b/ext/drbbdup/drbbdup.c
@@ -112,9 +112,7 @@ typedef struct {
     bool is_scratch_reg2_needed;
     bool is_scratch_reg2_dead; /* If _needed, is DRBBDUP_SCRATCH_REG2 dead at start. */
 #endif
-    bool is_gen;   /* Denotes whether a new bb copy is dynamically being generated. */
-    bool for_meta; /* Denotes whether this manager is for bbs that consist only of meta
-                      instructions. */
+    bool is_gen; /* Denotes whether a new bb copy is dynamically being generated. */
     drbbdup_case_t default_case;
     drbbdup_case_t *cases; /* Is NULL if enable_dup is not set. */
 } drbbdup_manager_t;
@@ -126,8 +124,6 @@ typedef enum {
 } drbbdup_label_t;
 
 typedef struct {
-    drbbdup_manager_t
-        *manager;   /* Manager corresponding to bb currently being instrumented. */
     int case_index; /* Used to keep track of the current case during insertion. */
     void *orig_analysis_data;        /* Analysis data accessible for all cases. */
     void *default_analysis_data;     /* Analysis data specific to default case. */
@@ -220,27 +216,10 @@ drbbdup_restore_register(void *drcontext, instrlist_t *ilist, instr_t *where,
 static bool
 drbbdup_is_special_instr(instr_t *instr)
 {
-    return instr_is_syscall(instr) || instr_is_cti(instr) || instr_is_ubr(instr) ||
-        instr_is_interrupt(instr);
+    return instr != NULL &&
+        (instr_is_syscall(instr) || instr_is_cti(instr) || instr_is_ubr(instr) ||
+         instr_is_interrupt(instr));
 }
-
-/****************************************************************************
- * DEBUG
- *
- * Contains debug functions.
- */
-
-#ifdef DEBUG
-void
-drbbdup_debug_check_for_meta(drbbdup_manager_t *manager, instrlist_t *bb)
-{
-    if (!manager->for_meta) {
-        app_pc pc = instr_get_app_pc(drbbdup_first_app(bb));
-        ASSERT(pc != NULL, "pc cannot be NULL");
-        ASSERT(hashtable_lookup(&manager_table, pc) != NULL, "manager cannot be NULL");
-    }
-}
-#endif
 
 /****************************************************************************
  * DUPlICATION PHASE
@@ -303,7 +282,7 @@ drbbdup_add_copy(void *drcontext, instrlist_t *bb, instrlist_t *orig_bb)
 
 /* Creates a manager, which contains book-keeping data for a fragment. */
 static drbbdup_manager_t *
-drbbdup_create_manager(void *drcontext, void *tag, instrlist_t *bb, bool for_meta)
+drbbdup_create_manager(void *drcontext, void *tag, instrlist_t *bb)
 {
     drbbdup_manager_t *manager = dr_global_alloc(sizeof(drbbdup_manager_t));
     memset(manager, 0, sizeof(drbbdup_manager_t));
@@ -316,8 +295,6 @@ drbbdup_create_manager(void *drcontext, void *tag, instrlist_t *bb, bool for_met
     manager->enable_dup = true;
     manager->enable_dynamic_handling = true;
     manager->is_gen = false;
-    manager->for_meta = for_meta; /* This is a temporary manager associated for bbs only
-                                     containing meta instructions. */
     manager->scratch_reg = DRBBDUP_SCRATCH_REG;
 
     ASSERT(opts.set_up_bb_dups != NULL, "set up call-back cannot be NULL");
@@ -336,13 +313,6 @@ drbbdup_create_manager(void *drcontext, void *tag, instrlist_t *bb, bool for_met
     DR_ASSERT_MSG(!manager->enable_dynamic_handling,
                   "dynamic case generation is not yet supported");
 
-    if (for_meta) {
-        /* XXX: For now we do not support dynamic handling for-meta-type managers. */
-        DR_ASSERT_MSG(
-            !manager->enable_dynamic_handling,
-            "dynamic case generation is not yet possible for bbs consisting solely "
-            "of meta instructions");
-    }
     /* Check whether user wants copies for this particular bb. */
     if (!manager->enable_dup && manager->cases != NULL) {
         /* Multiple cases not wanted. Destroy cases. */
@@ -406,7 +376,7 @@ drbbdup_set_up_copies(void *drcontext, instrlist_t *bb, drbbdup_manager_t *manag
      * This is done so that we do not copy such instructions and abide by DR rules.
      */
     instr_t *last = instrlist_last_app(original);
-    if (drbbdup_is_special_instr(last)) {
+    if (last != NULL && drbbdup_is_special_instr(last)) {
         instrlist_remove(original, last);
         instr_destroy(drcontext, last);
     }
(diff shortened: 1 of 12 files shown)


suite/tests/CMakeLists.txt
```diff
@@ -2579,6 +2579,51 @@ use_DynamoRIO_extension(client.drbbdup-test.dll drmgr)
...
+
+  add_library(client.drbbdup-meta-label-bb-test.dll SHARED  client-interface/drbbdup-meta-label-bb-test.dll.c)
+  setup_test_client_dll_basics(client.drbbdup-meta-label-bb-test.dll)
+  use_DynamoRIO_extension(client.drbbdup-meta-label-bb-test.dll drmgr)
+  use_DynamoRIO_extension(client.drbbdup-meta-label-bb-test.dll drreg)
+  use_DynamoRIO_extension(client.drbbdup-meta-label-bb-test.dll drbbdup)
+
+  torunonly_ci(client.drbbdup-meta-label-bb-test client.drbbdup-meta-label-bb-test
+    client.drbbdup-meta-label-bb-test.dll client-interface/drbbdup-meta-label-bb-test.asm "" "-max_bb_instrs 0" "")
+endif (X86 AND LINUX)
```
derekbruening, line 2595: Make a macro or function to share these three code blocks?  They look identical except the test name.

suite/tests/client-interface/drbbdup-empty-bb-test.asm
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+.section .text
+        .global main
+main:
+        or eax,eax
+        nop
+        nop
+        nop
+        mov eax, 1
```
derekbruening, line 40: Add a comment about which SYS_ this is.

derekbruening, line 40: Syscall 1 is exit for 32-bit only; it's write for 64-bit.  Is this test 32-bit only?  It looks enabled for 64-bit, where I assume it crashes running junk code after the int 0x80?

suite/tests/client-interface/drbbdup-empty-bb-test.dll.c
```diff
@@ -0,0 +1,300 @@
...
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+/* Tests the drbbdup extension. */
```
derekbruening, line 33: Add what in particular it is testing

suite/tests/client-interface/drbbdup-empty-bb-test.dll.c
```diff
@@ -0,0 +1,300 @@
...
+    CHECK(res == DRBBDUP_SUCCESS, "failed to check whether instr is start");
+    CHECK(is_first, "must be first");
+
+    res = drbbdup_is_first_nonlabel_instr(drcontext, instr, &is_first_nonlabel);
+    CHECK(res == DRBBDUP_SUCCESS, "failed to check whether instr is first non-label");
+    CHECK(is_first_nonlabel, "must be first non-label");
+
+    res = drbbdup_is_last_instr(drcontext, instr, &is_last);
+    CHECK(res == DRBBDUP_SUCCESS, "failed to check whether instr is last");
+    CHECK(is_last, "must be last");
```
derekbruening, line 232: I don't understand: this seems to assume that every block has either zero instrs or a single instr?  That doesn't seem right.

suite/tests/client-interface/drbbdup-empty-bb-test.dll.c
```diff
@@ -0,0 +1,300 @@
...
+{
+    return instrlist_first(bb) == NULL;
+}
+
+static dr_emit_flags_t
+remove_app_instr(void *drcontext, void *tag, instrlist_t *bb, bool for_trace,
+                 bool translating)
+{
+    instr_t *first_instr = instrlist_first_app(bb);
+    CHECK(instr_get_next(first_instr) == NULL, "must just be 1 instr");
```
derekbruening, line 72: It took a while to understand how this works: this relies on passing -max_bb_instrs 0.  Please add a comment about that, maybe here and at the top of the file.

suite/tests/client-interface/drbbdup-empty-bb-test.dll.c
derekbruening: OK now I understand: it relies on -max_bb_instrs 0.  Please add another comment here.

suite/tests/client-interface/drbbdup-meta-label-bb-test.dll.c
```diff
@@ -0,0 +1,321 @@
...
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+/* Tests the drbbdup extension. */
```
derekbruening, line 33: The same requests from above: explain what this is doing.  It assumes -max_bb_instrs 0, it assumes the particular asm app with some nops which it turns into labels to stress drbbdup with blocks containing only a label.

suite/tests/client-interface/drbbdup-meta-label-bb-test.dll.c
```diff
@@ -0,0 +1,321 @@
...
+                                          TEST_PRIORITY_APP2APP_NAME, NULL, NULL,
+                                          TEST_PRIORITY_APP2APP };
+
+    bool app2app_res =
+        drmgr_register_bb_app2app_event(remove_app_instr, &app2app_priority);
+    CHECK(app2app_res, "app2app failed");
+
+    drbbdup_status_t res = drbbdup_init(&opts);
+    CHECK(res == DRBBDUP_SUCCESS, "drbbdup init failed");
+    dr_register_exit_event(event_exit);
```
derekbruening, line 320: The whole file is very similar to the other 2 new tests here.  Is there a way to share code?  That would also make it easier to understand these tests, to highlight what matters among the hundreds of identical lines.

Another option would be to merge all 3 tests into one test and replace the first app nop with nothing, the 2nd with a label, and the 3rd with a meta nop, covering all the cases.  Presumably the couple of other changes throughout the file can be updated to accomodate that.  I think that would make it easier to maintain than these 3 separate mostly-duplicated files.   I vote for that.  WDYT?

ext/drbbdup/drbbdup.c
```diff
@@ -423,20 +441,14 @@ static dr_emit_flags_t
...
 
     drbbdup_manager_t *manager =
-        (drbbdup_manager_t *)hashtable_lookup(&manager_table, pc);
+        (drbbdup_manager_t *)hashtable_lookup(&manager_table, tag);
 
-    if (!for_trace && manager != NULL && !manager->is_gen) {
+    if (!for_trace && !translating && manager != NULL && !manager->is_gen) {
         /* Remove existing invalid book-keeping data. */
-        hashtable_remove(&manager_table, pc);
+        hashtable_remove(&manager_table, tag);
```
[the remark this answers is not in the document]
johnfxgalea: Yes it is trying to handle app code mods.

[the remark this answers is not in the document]
johnfxgalea: > What about thread-private caches: this will throw away on any block duplicated among threads?

We assume global cache for now. drbbdup does not support thread-private caches well atm.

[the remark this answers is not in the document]
johnfxgalea: @derekbruening By any chance, do you have a better idea for detecting app code mods?

ext/drbbdup/drbbdup.c
```diff
@@ -423,20 +441,14 @@ static dr_emit_flags_t
 drbbdup_duplicate_phase(void *drcontext, void *tag, instrlist_t *bb, bool for_trace,
                         bool translating)
 {
-    if (translating)
-        return DR_EMIT_DEFAULT;
```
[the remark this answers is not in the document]
johnfxgalea: Yes, exactly.

suite/tests/client-interface/drbbdup-meta-label-bb-test.dll.c
johnfxgalea: So my rationale was to keep tests as self-contained as possible, but you are right in saying there is a lot of code duplication.

johnfxgalea: I'll see what I can do.

suite/tests/client-interface/drbbdup-empty-bb-test.dll.c
johnfxgalea: Sorry, about that! Will add a comment to make it clearer.

suite/tests/client-interface/drbbdup-empty-bb-test.asm
johnfxgalea: Done.

johnfxgalea: Done.

suite/tests/client-interface/drbbdup-empty-bb-test.dll.c
johnfxgalea: Done.

suite/tests/client-interface/drbbdup-meta-label-bb-test.dll.c
johnfxgalea: Done.

ext/drbbdup/drbbdup.c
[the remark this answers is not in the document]
derekbruening: > @derekbruening By any chance, do you have a better idea for detecting app code mods?

Having something like #409 would be good.  We don't have that today though, and with races limiting dr_register_delete_event(), just looking for duplicate tags is a reasonable approach.  IIRC Dr. Memory is doing something like this too.  Its comment says it handle thread-private; haven't re-examined it enough to explain that well.  https://github.com/DynamoRIO/drmemory/blob/master/drmemory/spill.c#L1755

derekbruening:
Also, assuming PR #5374 goes in first, please then remove the `#if 0` around `DRWRAP_REPLACE_RETADDR` in suite/tests/client-interface/drbbdup-drwrap-test.dll.c: xref https://github.com/DynamoRIO/dynamorio/pull/5374/files/0398bf7f35f870614e5c1caefe24883d90186876#r812125576

suite/tests/CMakeLists.txt
johnfxgalea: Done

suite/tests/client-interface/drbbdup-empty-bb-test.dll.c
johnfxgalea: Done.

suite/tests/client-interface/drbbdup-meta-label-bb-test.dll.c
johnfxgalea: I shortened the test-cases so that they are easier to read.

johnfxgalea:
> Also, assuming PR #5374 goes in first, please then remove the `#if 0` around `DRWRAP_REPLACE_RETADDR` in suite/tests/client-interface/drbbdup-drwrap-test.dll.c: xref https://github.com/DynamoRIO/dynamorio/pull/5374/files/0398bf7f35f870614e5c1caefe24883d90186876#r812125576

Done.

suite/tests/client-interface/drbbdup-empty-bb-test.dll.c
johnfxgalea: Done.

johnfxgalea:
@derekbruening PTAL. If all is well, please go ahead with the merge.

derekbruening (approved):

petrochenkov:
@johnfxgalea @derekbruening 
This change somehow broke a client that doesn't even use drbbdup (but uses drmgr, drwrap and drreg), now it crashes with SIGSEGV in some instrumentation tracing data of store instructions, possibly inside a drwrapp'ed function.

I continue investigating, maybe it's some client's issue uncovered by this change, but perhaps you have some ideas about what could go wrong here.

johnfxgalea:
@petrochenkov Kindly provide further debug details when available so we can determine the root cause and proceed accordingly.

petrochenkov:
@johnfxgalea 
Sorry for the noise, there's nothing wrong with this commit, it just happen to coincide with the tag `cronbuild-9.0.19046`, and we use cronbuild versions for dealing with breaking changes in DynamoRIO, and some version checking incorrectly used this version, and it triggered a behavior change in the client that would be correct for a different future cronbuild version, but not for this version.

PR #5392, opened 2022-03-04
i#5390: Mark non-cloned final drbbdup instr as emulation

### LINKED ISSUE #5390: drbbdup fails to interoperate with drmgr emulation API
For #3995 I'm integrating drbbdup with drmemtrace, the tracer for drcachesim.
But drmemtrace uses the drmgr emulation support:
```C
instr_t *instr_fetch = drmgr_orig_app_instr_for_fetch(drcontext);
instr_t *instr_operands = drmgr_orig_app_instr_for_operands(drcontext);
```
Those routines are not supported by drbbdup, which splits the `where` to insert from the app instr being instrumented, to handle the inability to clone a block-final branch or syscall.
This causes drmemtrace to instrument the wrong instruction.

Xref past discussions on possibly integrating drbbdup with drmgr.
A possible simpler solution is to add a drmgr API to set the current app instruction.
Or, could we re-implement the drbbdup `where` vs `instr` split to instead use the emulation API itself?
That is one of the intended uses of the emulation API, for app instr rewrites that ease instrumentation (such as rep string or scatter-gather expansion): it is not only for pure emulation.

derekbruening:
Adds emulation markers for each non-cloned final drbbdup instruction
to ensure that drmgr_orig_app_instr_for_fetch() and
drmgr_orig_app_instr_for_operands() work properly with drbbdup.

Adds a sanity test that those emulation queries never see drbbdup's
jump to a label.  The test fails without the fix here.

Fixes #5390

### DIFF
--- a/ext/drbbdup/drbbdup.c
+++ b/ext/drbbdup/drbbdup.c
@@ -416,7 +416,26 @@ drbbdup_set_up_copies(void *drcontext, instrlist_t *bb, drbbdup_manager_t *manag
     for (i = start; i >= 0; i--) {
         /* Prepend a jmp targeting the EXIT label. */
         instr_t *jmp_exit = XINST_CREATE_jump(drcontext, exit_label_opnd);
-        instrlist_preinsert(bb, instrlist_first(bb), jmp_exit);
+        instr_t *first = instrlist_first(bb);
+        last = instrlist_last(bb);
+        if (drbbdup_is_special_instr(last)) {
+            /* Mark this jmp as emulating the final special instr to ensure that
+             * drmgr_orig_app_instr_for_fetch() and drmgr_orig_app_instr_for_operands()
+             * work properly (without this those two will look at this jmp since drmgr
+             * does not have a "where" vs "instr" split).
+             * XXX i#5390: Integrating drbbdup into drmgr is another way to solve this.
+             */
+            emulated_instr_t emulated_instr;
+            emulated_instr.size = sizeof(emulated_instr);
+            emulated_instr.pc = instr_get_app_pc(last);
+            emulated_instr.instr = instr_clone(drcontext, last); /* Freed by label. */
+            emulated_instr.flags = 0;
+            drmgr_insert_emulation_start(drcontext, bb, first, &emulated_instr);
+            instrlist_preinsert(bb, first, jmp_exit);
+            drmgr_insert_emulation_end(drcontext, bb, first);
+        } else {
+            instrlist_preinsert(bb, instrlist_first(bb), jmp_exit);
+        }
 
         /* Prepend a copy. */
         drbbdup_add_copy(drcontext, bb, original);
@@ -558,7 +577,11 @@ drbbdup_is_at_end(instr_t *check_instr)
 
     if (instr_is_cti(check_instr)) {
         instr_t *next_instr = instr_get_next(check_instr);
-        return drbbdup_is_at_label(next_instr, DRBBDUP_LABEL_START);
+        return drbbdup_is_at_label(next_instr, DRBBDUP_LABEL_START) ||
+            /* There may be an emulation endpoint label in between. */
+            (next_instr != NULL && instr_get_next(next_instr) != NULL &&
+             instr_is_label(next_instr) &&
+             drbbdup_is_at_label(instr_get_next(next_instr), DRBBDUP_LABEL_START));
     }
 
     return false;
@@ -1389,6 +1412,7 @@ drbbdup_instrument_dups(void *drcontext, void *tag, instrlist_t *bb, instr_t *in
     } else if (drbbdup_is_at_end(instr)) {
         /* Handle last special instruction (if present). */
         if (is_last_special) {
+            // NOCHECK
             flags = drbbdup_instrument_instr(drcontext, tag, bb, last, instr, for_trace,
                                              translating, pt, manager);
             if (pt->case_index == DRBBDUP_DEFAULT_INDEX) {
--- a/ext/drbbdup/drbbdup.c
+++ b/ext/drbbdup/drbbdup.c
@@ -343,6 +343,29 @@ drbbdup_destroy_manager(void *manager_opaque)
     dr_global_free(manager, sizeof(drbbdup_manager_t));
 }
 
+static bool
+drbbdup_ilist_has_cti(instrlist_t *bb)
+{
+    for (instr_t *inst = instrlist_first(bb); inst != NULL; inst = instr_get_next(inst)) {
+        if (instr_is_cti(inst))
+            return true;
+    }
+    return false;
+}
+
+static bool
+drbbdup_ilist_has_unending_emulation(instrlist_t *bb)
+{
+    emulated_instr_t emul_info = { sizeof(emul_info), 0 };
+    for (instr_t *inst = instrlist_first(bb); inst != NULL; inst = instr_get_next(inst)) {
+        if (drmgr_is_emulation_start(inst) &&
+            drmgr_get_emulated_instr_data(inst, &emul_info) &&
+            TEST(DR_EMULATE_REST_OF_BLOCK, emul_info.flags))
+            return true;
+    }
+    return false;
+}
+
 /* Transforms the bb to contain additional copies (within the same fragment). */
 static void
 drbbdup_set_up_copies(void *drcontext, instrlist_t *bb, drbbdup_manager_t *manager)
@@ -373,18 +396,8 @@ drbbdup_set_up_copies(void *drcontext, instrlist_t *bb, drbbdup_manager_t *manag
      * Note, we add jmp instructions here and DR will set them to meta automatically.
      */
 
-    bool has_rest_of_block_emulation = false;
-    bool has_prior_control_flow = false;
-    emulated_instr_t emul_info = { sizeof(emul_info), 0 };
-    for (instr_t *inst = instrlist_first(bb); inst != NULL; inst = instr_get_next(inst)) {
-        if (instr_is_cti(inst))
-            has_prior_control_flow = true;
-        if (drmgr_is_emulation_start(inst) &&
-            drmgr_get_emulated_instr_data(inst, &emul_info) &&
-            TEST(DR_EMULATE_REST_OF_BLOCK, emul_info.flags)) {
-            has_rest_of_block_emulation = true;
-        }
-    }
+    bool has_rest_of_block_emulation = drbbdup_ilist_has_unending_emulation(bb);
+    bool has_prior_control_flow = drbbdup_ilist_has_cti(bb);
 
     /* We create a duplication here to keep track of original bb. */
     instrlist_t *original = instrlist_clone(drcontext, bb);
@@ -615,7 +628,7 @@ drbbdup_is_at_end(instr_t *check_instr)
 }
 
 static bool
-drbbdup_is_our_emulation_marker(instr_t *check_instr)
+drbbdup_is_exit_jmp_emulation_marker(instr_t *check_instr)
 {
     if (check_instr == NULL)
         return false;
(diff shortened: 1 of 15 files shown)


derekbruening:
Test failure is known #5329 32-bit scattergather assert on towrite.

ext/drbbdup/drbbdup.c
```diff
@@ -1389,6 +1412,7 @@ drbbdup_instrument_dups(void *drcontext, void *tag, instrlist_t *bb, instr_t *in
     } else if (drbbdup_is_at_end(instr)) {
         /* Handle last special instruction (if present). */
         if (is_last_special) {
+            // NOCHECK
```
johnfxgalea, line 1415: What does NOCHECK mean here, please? I am only familiar with this term in an SQL context.

ext/drbbdup/drbbdup.c
```diff
@@ -558,7 +577,11 @@ drbbdup_is_at_end(instr_t *check_instr)
 
     if (instr_is_cti(check_instr)) {
         instr_t *next_instr = instr_get_next(check_instr);
-        return drbbdup_is_at_label(next_instr, DRBBDUP_LABEL_START);
+        return drbbdup_is_at_label(next_instr, DRBBDUP_LABEL_START) ||
+            /* There may be an emulation endpoint label in between. */
+            (next_instr != NULL && instr_get_next(next_instr) != NULL &&
+             instr_is_label(next_instr) &&
```
johnfxgalea, line 583: Is there a way to add to the condition that the label is actually an emulation endpoint label, and not just check it is a generic label?

ext/drbbdup/drbbdup.c
derekbruening: My mistake: that is my marker for something that needs attention and it should have been removed after adding the handling.  Removed.

suite/tests/client-interface/drbbdup-meta-nop-bb-test.dll.c
```diff
@@ -193,17 +193,24 @@ instrument_instr(void *drcontext, void *tag, instrlist_t *bb, instr_t *instr,
     if (is_cur_all_meta)
         CHECK(is_meta_nop(instr), "should be the meta nop instruction");
 
+    bool has_emul_labels = drmgr_is_emulation_start(instr) || instr_is_syscall(instr) ||
```
johnfxgalea, line 196: I do not get these changes. Why would emulation affect results of drbbdup_is_first_nonlabel_instr, drbbdup_is_first_instr, etc, and therefore we need to edge them out in these test-cases? Should drbbdup internally  handle the emulation of the jmp exit instruction? From the perspective of the client, the jmp exit instruction is invisible.

derekbruening: The emulation labels are passed to the insert-instrumentation callbacks: so this code which expects just one instruction sees 3 instructions for the syscall in the asm: label;syscall;label.

ext/drbbdup/drbbdup.c
derekbruening: Yes we can use drmgr_is_emulation_{start,end}.  Will put that in.  But first there's a crash in PR #5393 -- first making sure that is not related to more that needs to be done here before pushing changes.

suite/tests/client-interface/drbbdup-test.dll.c
```diff
@@ -214,6 +214,18 @@ instrument_instr(void *drcontext, void *tag, instrlist_t *bb, instr_t *instr,
         CHECK(is_first_nonlabel, "should be first non label");
     }
 
+    /* Ensure the drmgr emulation API works for a final special instr where "instr" !=
+     * "where".  The emulation shows up on the label prior to the "where" for the last
+     * instr so we can't compare emul to instr; instead we make sure the emul returned
+     * is never a jump to a label.
+     */
```
johnfxgalea, line 221: What about other emulation done by drbbdup clients which include a jump to label? Will confusion arise? I think it would be best to have a drbbdup client also do some from of emulation and then check there are no mismatches

suite/tests/client-interface/drbbdup-meta-nop-bb-test.dll.c
johnfxgalea: Why should drbbdup pass them to the client? I guess this is an indicator for client to call drmgr_orig_app_instr_for_fetch(), but can't the client just call it anyways and check for nullness?

johnfxgalea: Note that drbbdup_is_last_instr() should return true when the special instruction is passed. If I pass the emulated instruction to drbbdup_is_last_instr(), does it return true?

suite/tests/client-interface/drbbdup-test.dll.c
derekbruening: I don't understand: this test is using this check for a jump-label as a sanity check for base drbbdup.  A regular client is not going to have a check like this: it is specific to this test.  So if someone added emulation that somehow pointed at a jump-label (ignoring that that doesn't really make sense: why would being-emulated app code have a jump to a label), yes it would thwart this test-specific check: but that is not relevant; that is not part of this test.

The separate topic of ensuring that the drbbdup emulation markers interoperate with other emulation markers: I was going to say that since the drbbdup emulation markers are around its own synthetic instruction it is not possible to overlap with any other emulation, but there is DR_EMULATE_REST_OF_BLOCK -- for that drbbdup should probably not insert any markers at all.  But that's where the drmemtrace errors came from, rep string, which sets that flag: so now it is confusing as to why the rest-of-block flag didn't solve things there.  Plus, this bring up whether the rest-of-block inside drmgr then messes up the subsequent drbbdup cases: a rest-of-block in the first case then applies to *all* the app instructions in all the subsequent cases??  Another drbbdup incompatibility it seems like...

derekbruening: > But that's where the drmemtrace errors came from, rep string, which sets that flag: so now it is confusing as to why the rest-of-block flag didn't solve things there.

Scratch that, that's wrong: the emulation errors were coming from every block-final branch, not rep string.  The rep string problem is the crash in PR #5393 which seems to be unrelated to emulation markers but is instead from something to do with drreg control flow.

derekbruening: For DR_EMULATE_REST_OF_BLOCK: we can solve that by having drbbdup insert an emulation end label at the end of each case.  drmgr will end a region if it sees that, even if the flag is set.

johnfxgalea: > I don't understand: this test is using this check for a jump-label as a sanity check for base drbbdup. A regular client is not going to have a check like this: it is specific to this test. So if someone added emulation that somehow pointed at a jump-label (ignoring that that doesn't really make sense: why would being-emulated app code have a jump to a label), yes it would thwart this test-specific check: but that is not relevant; that is not part of this test.

Yeah, sorry about that (I had my security hat on); I was trying to get to the interpolation concern and was think of how would I extend this test-case to break those checks.

johnfxgalea: For DR_EMULATE_REST_OF_BLOCK: we can solve that by having drbbdup insert an emulation end label at the end of each case.  drmgr will end a region if it sees that, even if the flag is set.

Assuming that reasoning over the end label is control-flow sensitive, I think that should work.

suite/tests/client-interface/drbbdup-meta-nop-bb-test.dll.c
derekbruening: OK you are saying can drbbdup hide the emulation markers from the client for the drbbdup-controlled instru insertion callbacks, and only have drmgr see them for the drmgr emulation queries.  Yes that seems doable.

derekbruening: And the pt->last_instr needs to skip over the emulation labels.  Done.

ext/drbbdup/drbbdup.c
```diff
@@ -597,6 +661,21 @@ drbbdup_next_end(instr_t *instr)
...
+ * emulation labels, steps back to point at the first of those: i.e., this is the
+ * start of the end sequence.  Returns NULL upon failure.
+ */
+static instr_t *
+drbbdup_next_end_initial(instr_t *instr)
+{
+    while (instr != NULL && !drbbdup_is_at_end(instr))
+        instr = instr_get_next(instr);
+    if (drbbdup_is_our_emulation_marker(instr_get_prev(instr)))
+        instr = instr_get_prev(instr);
```
johnfxgalea, line 674: Assert that this is an emu start label instr.

ext/drbbdup/drbbdup.c
```diff
@@ -1349,7 +1430,9 @@ drbbdup_instrument_dups(void *drcontext, void *tag, instrlist_t *bb, instr_t *in
         if (is_last_special) {
             pt->last_instr = last;
         } else {
-            instr_t *prev = instr_get_prev(end_instr);
+            instr_t *prev = instr_get_prev(end_initial);
+            if (drbbdup_is_our_emulation_marker(prev))
```
johnfxgalea, line 1434: I don't get this branch. From my reading `end_initial` should be the start emu label, and we called instr_get_prev() to get the prior instr in the line before. Why do we have this branch. May you add a comment explaining it.

johnfxgalea (approved):

ext/drbbdup/drbbdup.c
```diff
@@ -373,6 +373,19 @@ drbbdup_set_up_copies(void *drcontext, instrlist_t *bb, drbbdup_manager_t *manag
      * Note, we add jmp instructions here and DR will set them to meta automatically.
      */
 
+    bool has_rest_of_block_emulation = false;
```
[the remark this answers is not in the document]
derekbruening: Done.  Two separate functions means two separate ilist walks but it is better code so seems worthwhile.

ext/drbbdup/drbbdup.c
```diff
@@ -597,6 +661,21 @@ drbbdup_next_end(instr_t *instr)
...
+/* Iterates forward to the end of the next bb copy.  If the end instruction has
+ * emulation labels, steps back to point at the first of those: i.e., this is the
+ * start of the end sequence.  Returns NULL upon failure.
+ */
+static instr_t *
+drbbdup_next_end_initial(instr_t *instr)
+{
+    while (instr != NULL && !drbbdup_is_at_end(instr))
+        instr = instr_get_next(instr);
+    if (drbbdup_is_our_emulation_marker(instr_get_prev(instr)))
```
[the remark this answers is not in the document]
derekbruening: Done.

ext/drbbdup/drbbdup.c
```diff
@@ -558,12 +604,28 @@ drbbdup_is_at_end(instr_t *check_instr)
...
+            (next_instr != NULL && instr_get_next(next_instr) != NULL &&
+             drmgr_is_emulation_end(next_instr) &&
+             drbbdup_is_at_label(instr_get_next(next_instr), DRBBDUP_LABEL_START));
     }
 
     return false;
 }
 
+static bool
+drbbdup_is_our_emulation_marker(instr_t *check_instr)
```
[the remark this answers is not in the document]
derekbruening: Done.

ext/drbbdup/drbbdup.c
derekbruening: Done.

ext/drbbdup/drbbdup.c
derekbruening: Hmm...looks like this was left from a prior version where there was no separate `end_initial`.  Unfortunately I'm getting sloppy here it seems, trying to get this all tied up too quickly to get the #3995 going.  Removing this if().

derekbruening:
I put in a little more time and was able to repro the #5398 crash.  I'm going to remove the code here that fixes that in favor of a separate PR with that fix and the test augmentation.

PR #5405, opened 2022-03-09
i#4842 drcachesim unit tests: Add cache_lru unit test

bete0:
This change adds unit tests for clients/drcachesim/simulators/cache_lru.cpp.
These unit tests catch issues reported in i#4881 which was tested by
reverting the fix for that issue #4883.
    
Issue: #4842, #4881

### DIFF
--- a/clients/drcachesim/CMakeLists.txt
+++ b/clients/drcachesim/CMakeLists.txt
@@ -499,7 +499,8 @@ endif ()
 # Be sure to give the targets qualified test names ("tool.drcache*...").
 
 if (BUILD_TESTS)
-  add_executable(tool.drcachesim.unit_tests tests/drcachesim_unit_tests.cpp tests/cache_replacement_policy_unit_test.cpp)
+  add_executable(tool.drcachesim.unit_tests tests/drcachesim_unit_tests.cpp
+    tests/cache_replacement_policy_unit_test.cpp)
   if (ZLIB_FOUND)
     target_link_libraries(tool.drcachesim.unit_tests drmemtrace_simulator
       drmemtrace_static drmemtrace_analyzer ${ZLIB_LIBRARIES})
--- a/clients/drcachesim/CMakeLists.txt
+++ b/clients/drcachesim/CMakeLists.txt
@@ -499,6 +499,17 @@ endif ()
 # Be sure to give the targets qualified test names ("tool.drcache*...").
 
 if (BUILD_TESTS)
+  add_executable(tool.drcachesim.cache_lru_unit_tests tests/cache_lru_unit_tests.cpp)
+  if (ZLIB_FOUND)
+    target_link_libraries(tool.drcachesim.cache_lru_unit_tests drmemtrace_simulator
+      drmemtrace_static drmemtrace_analyzer ${ZLIB_LIBRARIES})
+  else ()
+    target_link_libraries(tool.drcachesim.cache_lru_unit_tests drmemtrace_simulator
+      drmemtrace_static drmemtrace_analyzer)
+  endif ()
+  add_win32_flags(tool.drcachesim.cache_lru_unit_tests)
+  add_test(NAME tool.drcachesim.cache_lru_unit_tests
+           COMMAND tool.drcachesim.cache_lru_unit_tests)
   add_executable(tool.drcachesim.unit_tests tests/drcachesim_unit_tests.cpp)
   if (ZLIB_FOUND)
     target_link_libraries(tool.drcachesim.unit_tests drmemtrace_simulator
--- a/clients/drcachesim/CMakeLists.txt
+++ b/clients/drcachesim/CMakeLists.txt
@@ -87,6 +87,7 @@ macro (add_exported_library name type)
   add_dependencies(${name} api_headers)
 endmacro ()
 
+add_exported_library(drcachesim_cache_replacement_policy_test STATIC tests/cache_replacement_policy_unit_test.cpp)
 add_exported_library(drmemtrace_reuse_distance STATIC tools/reuse_distance.cpp)
 add_exported_library(drmemtrace_histogram STATIC tools/histogram.cpp)
 add_exported_library(drmemtrace_reuse_time STATIC tools/reuse_time.cpp)
@@ -499,23 +500,12 @@ endif ()
 # Be sure to give the targets qualified test names ("tool.drcache*...").
 
 if (BUILD_TESTS)
-  add_executable(tool.drcachesim.cache_lru_unit_tests tests/cache_lru_unit_tests.cpp)
-  if (ZLIB_FOUND)
-    target_link_libraries(tool.drcachesim.cache_lru_unit_tests drmemtrace_simulator
-      drmemtrace_static drmemtrace_analyzer ${ZLIB_LIBRARIES})
-  else ()
-    target_link_libraries(tool.drcachesim.cache_lru_unit_tests drmemtrace_simulator
-      drmemtrace_static drmemtrace_analyzer)
-  endif ()
-  add_win32_flags(tool.drcachesim.cache_lru_unit_tests)
-  add_test(NAME tool.drcachesim.cache_lru_unit_tests
-           COMMAND tool.drcachesim.cache_lru_unit_tests)
   add_executable(tool.drcachesim.unit_tests tests/drcachesim_unit_tests.cpp)
   if (ZLIB_FOUND)
-    target_link_libraries(tool.drcachesim.unit_tests drmemtrace_simulator
+    target_link_libraries(tool.drcachesim.unit_tests drcachesim_cache_replacement_policy_test drmemtrace_simulator
       drmemtrace_static drmemtrace_analyzer ${ZLIB_LIBRARIES})
   else ()
-    target_link_libraries(tool.drcachesim.unit_tests drmemtrace_simulator
+    target_link_libraries(tool.drcachesim.unit_tests drcachesim_cache_replacement_policy_test drmemtrace_simulator
       drmemtrace_static drmemtrace_analyzer)
   endif ()
   add_win32_flags(tool.drcachesim.unit_tests)
--- a/clients/drcachesim/CMakeLists.txt
+++ b/clients/drcachesim/CMakeLists.txt
@@ -87,7 +87,6 @@ macro (add_exported_library name type)
   add_dependencies(${name} api_headers)
 endmacro ()
 
-add_exported_library(drcachesim_cache_replacement_policy_test STATIC tests/cache_replacement_policy_unit_test.cpp)
 add_exported_library(drmemtrace_reuse_distance STATIC tools/reuse_distance.cpp)
 add_exported_library(drmemtrace_histogram STATIC tools/histogram.cpp)
 add_exported_library(drmemtrace_reuse_time STATIC tools/reuse_time.cpp)
@@ -500,12 +499,12 @@ endif ()
 # Be sure to give the targets qualified test names ("tool.drcache*...").
 
 if (BUILD_TESTS)
-  add_executable(tool.drcachesim.unit_tests tests/drcachesim_unit_tests.cpp)
+  add_executable(tool.drcachesim.unit_tests tests/drcachesim_unit_tests.cpp tests/cache_replacement_policy_unit_test.cpp)
   if (ZLIB_FOUND)
-    target_link_libraries(tool.drcachesim.unit_tests drcachesim_cache_replacement_policy_test drmemtrace_simulator
+    target_link_libraries(tool.drcachesim.unit_tests drmemtrace_simulator
       drmemtrace_static drmemtrace_analyzer ${ZLIB_LIBRARIES})
   else ()
-    target_link_libraries(tool.drcachesim.unit_tests drcachesim_cache_replacement_policy_test drmemtrace_simulator
+    target_link_libraries(tool.drcachesim.unit_tests drmemtrace_simulator
       drmemtrace_static drmemtrace_analyzer)
   endif ()
   add_win32_flags(tool.drcachesim.unit_tests)
--- a/clients/drcachesim/tests/cache_lru_unit_tests.cpp
+++ b/clients/drcachesim/tests/cache_lru_unit_tests.cpp
@@ -1,5 +1,5 @@
 /* **********************************************************
- * Copyright (c) 2016-2021 Google, Inc.  All rights reserved.
+ * Copyright (c) 2016-2022 Google, Inc.  All rights reserved.
  * **********************************************************/
 
 /*
@@ -75,7 +75,7 @@ public:
             exit(1);
         }
 
-        // Access the cacheclines in set 1 in the following sequence and ensure the
+        // Access the cache lines in set 1 in the following sequence and ensure the
         // counters are updated properly.
         access_update(1, 6);
         access_update(1, 1);
--- a/clients/drcachesim/tests/cache_lru_unit_tests.cpp
+++ b/clients/drcachesim/tests/cache_lru_unit_tests.cpp
@@ -1,0 +1,106 @@
+/* **********************************************************
+ * Copyright (c) 2016-2021 Google, Inc.  All rights reserved.
+ * **********************************************************/
+
+/*
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice,
+ *   this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of Google, Inc. nor the names of its contributors may be
+ *   used to endorse or promote products derived from this software without
+ *   specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL GOOGLE, INC. OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+// Unit tests for cache_lru
+#include <iostream>
+#include <assert.h>
+#include "simulator/cache_lru.h"
+
+struct cache_config_t {
+    int associativity = 4;
+    int line_size = 32;
+    int total_size = 256;
+    caching_device_t *parent = nullptr;
+    caching_device_stats_t *stats = new cache_stats_t(line_size, "", true);
+    prefetcher_t *prefetcher = nullptr;
+    bool inclusive = true;
+    bool coherent_cache = true;
+    int id = -1;
+    snoop_filter_t *snoop_filter = nullptr;
+};
+
+static cache_config_t
+make_test_configs(int associativity, int line_size, int total_size)
+{
+    cache_config_t config;
+    config.associativity = associativity;
+    config.line_size = line_size;
+    config.total_size = total_size;
+    return config;
+}
+
+class cache_lru_test_t : public cache_lru_t {
+public:
+    void
+    unit_test_access_update()
+    {
+        // Create and initialize an 8-way set associative cache with line size of 32 and
+        // total size of 1024 bytes.
+        cache_config_t config_two = make_test_configs(8, 32, 1024);
+        if (!init(config_two.associativity, config_two.line_size, config_two.total_size,
+                  config_two.parent, config_two.stats, config_two.prefetcher,
+                  config_two.inclusive, config_two.coherent_cache, config_two.id,
+                  config_two.snoop_filter)) {
+            std::cerr << "LRU cache failed to initialize"
+                      << "\n";
+            exit(1);
+        }
+
+        // Access the cacheclines in set 1 in the following sequence and ensure the
+        // counters are updated properly.
+        access_update(1, 6);
+        access_update(1, 1);
+        access_update(1, 2);
+        access_update(1, 3);
+        access_update(1, 4);
+        access_update(1, 5);
+        access_update(1, 0);
+        access_update(1, 7);
+
+        assert(get_caching_device_block(1, 0).counter_ == 1);
+        assert(get_caching_device_block(1, 1).counter_ == 6);
+        assert(get_caching_device_block(1, 2).counter_ == 5);
+        assert(get_caching_device_block(1, 3).counter_ == 4);
+        assert(get_caching_device_block(1, 4).counter_ == 3);
+        assert(get_caching_device_block(1, 5).counter_ == 2);
+        assert(get_caching_device_block(1, 6).counter_ == 7);
+        assert(get_caching_device_block(1, 7).counter_ == 0);
+    }
+};
+
+int
+main(int argc, const char *argv[])
+{
+    cache_lru_test_t cache_lru;
+    cache_lru.unit_test_access_update();
+    return 0;
+}
--- a/clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
+++ b/clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
@@ -39,6 +39,40 @@
 class cache_lru_test_t : public cache_lru_t {
 public:
     void
+    unit_test_replace_which_way()
+    {
+        // Create and initialize a 4-way set associative cache with line size of 32 and
+        // total size of 256 bytes.
+        initialize_cache(4, 32, 256);
+
+        // Access the cache line in the following fashion. This sequence follows the
+        // sequence shown in https://github.com/DynamoRIO/dynamorio/issues/4881.
+        const int ADDRESS_A = 0;
+        const int ADDRESS_B = 64;
+        const int ADDRESS_C = 128;
+        const int ADDRESS_D = 192;
+        const int ADDRESS_E = 72;
+
+        assert(get_block_index(ADDRESS_A) == get_block_index(ADDRESS_B));
+        assert(get_block_index(ADDRESS_B) == get_block_index(ADDRESS_C));
+        assert(get_block_index(ADDRESS_C) == get_block_index(ADDRESS_D));
+        assert(get_block_index(ADDRESS_D) == get_block_index(ADDRESS_E));
+
+        // Lower-case letter shows the least recently used way.
+        access_and_check_lru(ADDRESS_A, 1); // A X X X
+        access_and_check_lru(ADDRESS_B, 2); // A B X X
+        access_and_check_lru(ADDRESS_C, 3); // A B C X
+        access_and_check_lru(ADDRESS_D, 0); // a B C D
+        access_and_check_lru(ADDRESS_A, 1); // A b C D
+        access_and_check_lru(ADDRESS_A, 1); // A b C D
+        access_and_check_lru(ADDRESS_A, 1); // A b C D
+        access_and_check_lru(ADDRESS_E, 2); // A E c D
+
+        // XXX i#4842: Add more test sequences.
+    }
+
+private:
+    void
     initialize_cache(int associativity, int line_size, int total_size)
     {
         caching_device_stats_t *stats = new cache_stats_t(line_size, "", true);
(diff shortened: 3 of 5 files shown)


clients/drcachesim/CMakeLists.txt
```diff
@@ -499,6 +499,17 @@ endif ()
 # Be sure to give the targets qualified test names ("tool.drcache*...").
 
 if (BUILD_TESTS)
+  add_executable(tool.drcachesim.cache_lru_unit_tests tests/cache_lru_unit_tests.cpp)
```
derekbruening, line 502: I would vote for not having a separate executable: a separate file makes sense but we can link all the unit tests together into one executable.  I'm not sure I see any advantage of separate executables: we're always going to run all of them.  There are a number of downsides: extra build code complexity; extra build time and extra test time increasing the CI time.

clients/drcachesim/CMakeLists.txt
```diff
@@ -499,6 +499,17 @@ endif ()
 # Be sure to give the targets qualified test names ("tool.drcache*...").
```
derekbruening, line 499: > No description provided.

Please fix the PR title and description.  Normally you'd squash local commits into one prior to making the PR (once the branch is shared don't edit shared commits -- but clean up private commits prior to sharing) which would then auto-populate the PR title and description with what we want.

clients/drcachesim/tests/cache_lru_unit_tests.cpp
```diff
@@ -0,0 +1,106 @@
...
+        access_update(1, 6);
+        access_update(1, 1);
+        access_update(1, 2);
+        access_update(1, 3);
+        access_update(1, 4);
+        access_update(1, 5);
+        access_update(1, 0);
+        access_update(1, 7);
+
+        assert(get_caching_device_block(1, 0).counter_ == 1);
```
derekbruening, line 89: I think it would be more robust and maintainable to have a black-box test that doesn't know/care about *how* LRU is implemented with this private counter: it would just makes sure the least-recently-used line is replaced.  So it would ask replace_which_way() and ensure it got the right answer, rather than making assumptions about how LRU is implemented using this internal counter variable.

Then we ensure that what matters, replace_which_way(), is operating how we want it to: which this test here doesn't actually test, since we could have perfect counter udpate code but a bug in replace_which_way and this wouldn't find it.

Plus, we could then swap to a different internal counter algorithm for LRU and have the same test, as it would focus on the interface: the contract this LRU module is providing.

clients/drcachesim/CMakeLists.txt
bete0: Done.

clients/drcachesim/CMakeLists.txt
bete0: Done.

clients/drcachesim/tests/cache_lru_unit_tests.cpp
bete0: When testing out this change with CL https://github.com/DynamoRIO/dynamorio/pull/4883 reverted I realized `replace_which_way()` could pass (returns the `way` with highest counter) even though the counters are not set properly in the first place. For example consider the 8-way set associative cache used in this test. With https://github.com/DynamoRIO/dynamorio/pull/4883 reverted, when accessing the ways sequentially 0 through 7, the counter values will be as follows:

| way | counter |
| ----|-------- |
|  0  |    2    |
|  1  |    2    |
|  2  |    2    |
|  3  |    2    |
|  4  |    2    |
|  5  |    2    |
|  6  |    1    |
|  7  |    0    |

As opposed to the following correct ones:
| way | counter |
| ----|-------- |
|  0  |    7    |
|  1  |    6    |
|  2  |    5    |
|  3  |    4    |
|  4  |    3    |
|  5  |    2    |
|  6  |    1    |
|  7  |    0    |

That is obviously wrong but `replace_which_way()` would still return the correct way to be replaced (0). The reason for that is even though the actual counter is wrong, it is still the highest count. This masks the bug fixed by https://github.com/DynamoRIO/dynamorio/pull/4883 fixed so decided to keep the counter test in there as well which this unit test would catch as expected.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,102 @@
...
+    assert(get_caching_device_block(0, 0).counter_ == 7);
+    assert(get_caching_device_block(0, 1).counter_ == 6);
+    assert(get_caching_device_block(0, 2).counter_ == 5);
+    assert(get_caching_device_block(0, 3).counter_ == 4);
+    assert(get_caching_device_block(0, 4).counter_ == 3);
+    assert(get_caching_device_block(0, 5).counter_ == 2);
+    assert(get_caching_device_block(0, 6).counter_ == 1);
+    assert(get_caching_device_block(0, 7).counter_ == 0);
+
+    assert(replace_which_way(0) == 0);
```
derekbruening, line 94: I expected to see the sequence from #4881, since that is part of the motivation for prioritizing more unit tests, with a confirmation that removing that fix results in the test failing.

derekbruening, line 94: I also expected to see a variety of other sequences to give more confidence and cover more corner cases.  This is just one simple test case.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+    memref_t ref;
+    ref.data.type = TRACE_TYPE_READ;
+    ref.data.size = 1;
+    for (int i = 0; i < 256; i++) {
+        ref.data.addr = i;
+        request(ref);
+    }
+
+    assert(get_caching_device_block(0, 0).counter_ == 7);
```
derekbruening, line 85: I would put comments explaining why you're looking into the internal state and making assumptions on the implementation details, since the reader might expect a purely black-box test (I think I would still vote for that).

clients/drcachesim/tests/cache_replacement_policy_unit_test.h
```diff
@@ -0,0 +1,47 @@
...
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+#ifndef _CACHE_REPL_POLICY_UNIT_TESTS_
+#define _CACHE_REPL_POLICY_UNIT_TESTS_ 1
+
+#include "simulator/cache_lru.h"
+
+class cache_lru_test_t : public cache_lru_t {
```
derekbruening, line 38: This class seems private to its test: no reason to be exported in a header?

clients/drcachesim/tests/cache_lru_unit_tests.cpp
derekbruening: I would expect to test the actual sequence from #4881 that exposes the bug?  Isn't the problem that the test in this PR only tests one simple sequence and it's not a sequence that actually hits the bug?

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
bete0: Added more test cases.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
bete0: Done. Removed the counter assertions.

clients/drcachesim/tests/cache_replacement_policy_unit_test.h
bete0: Done.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
bete0: Done. Added a test case that correspond to https://github.com/DynamoRIO/dynamorio/issues/4881

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,102 @@
...
+ * DAMAGE.
+ */
+
+// Unit tests for cache replacement policies
+#include <iostream>
+#include <assert.h>
+#include "cache_replacement_policy_unit_test.h"
+#include "simulator/cache_lru.h"
+
+struct cache_config_t {
```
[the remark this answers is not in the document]
bete0: `cache_lru::init` does not have default values for these configs. 

from `https://github.com/DynamoRIO/dynamorio/blob/master/clients/drcachesim/simulator/cache_lru.cpp`
```:
bool
cache_lru_t::init(int associativity, int block_size, int total_size,
                  caching_device_t *parent, caching_device_stats_t *stats,
                  prefetcher_t *prefetcher, bool inclusive, bool coherent_cache, int id,
                  snoop_filter_t *snoop_filter,
                  const std::vector<caching_device_t *> &children)
{
```

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,102 @@
...
+    config.total_size = total_size;
+    return config;
+}
+
+void
+cache_lru_test_t::unit_test_replace_which_way()
+{
+    // Create and initialize an 8-way set associative cache with line size of 32 and
+    // total size of 256 bytes.
+    cache_config_t config_two = make_test_configs(8, 32, 256);
```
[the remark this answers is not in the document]
bete0: Done.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,102 @@
...
+    caching_device_stats_t *stats = new cache_stats_t(line_size, "", true);
+    prefetcher_t *prefetcher = nullptr;
+    bool inclusive = true;
+    bool coherent_cache = true;
+    int id = -1;
+    snoop_filter_t *snoop_filter = nullptr;
+};
+
+static cache_config_t
+make_test_configs(int associativity, int line_size, int total_size)
```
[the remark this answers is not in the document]
bete0: Remove "s".

clients/drcachesim/CMakeLists.txt
```diff
@@ -87,6 +87,7 @@ macro (add_exported_library name type)
   add_dependencies(${name} api_headers)
 endmacro ()
 
+add_exported_library(drcachesim_cache_replacement_policy_test STATIC tests/cache_replacement_policy_unit_test.cpp)
```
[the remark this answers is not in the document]
bete0: Done.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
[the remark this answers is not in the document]
derekbruening: It does have defaults.  Look at the header file.

derekbruening:
(Remember to re-request review when ready; that is how reviewers know to take another look)

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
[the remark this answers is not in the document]
bete0: Done. Removed the struct.

derekbruening:
I noticed that each additional commit here seems to have an identical commit title and description to the first: it makes it difficult to figure out what each one is changing.  Normally each would describe specifically what it is doing even if it's just "Update comments per reviewer requests".

clients/drcachesim/CMakeLists.txt
```diff
@@ -499,7 +499,7 @@ endif ()
 # Be sure to give the targets qualified test names ("tool.drcache*...").
 
 if (BUILD_TESTS)
-  add_executable(tool.drcachesim.unit_tests tests/drcachesim_unit_tests.cpp)
+  add_executable(tool.drcachesim.unit_tests tests/drcachesim_unit_tests.cpp tests/cache_replacement_policy_unit_test.cpp)
```
derekbruening, line 502: style: Split the long line

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,163 @@
...
+// Unit tests for cache replacement policies
+#include <iostream>
+#include <assert.h>
+#include "cache_replacement_policy_unit_test.h"
+#include "simulator/cache_lru.h"
+
+class cache_lru_test_t : public cache_lru_t {
+public:
+    void
+    initialize_cache(int way, int line_size, int total_size)
```
derekbruening, line 42: nit: "way" means one specific selection (e.g., as passed to access_update()), while this is really the count of how many ways there are, or the associativity as all the parent classes call it.  So "associativity" for consistency, or "ways" or "way_count", seem better name choices.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,163 @@
...
+    {
+        caching_device_t *parent = nullptr;
+        prefetcher_t *prefetcher = nullptr;
```
derekbruening, line 45: nit: Do these need named local variables?

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,163 @@
...
+            exit(1);
+        }
+    }
+
+    void
+    unit_test_replace_which_way()
+    {
+        // Create and initialize a 4-way set associative cache with line size of 32 and
+        // total size of 128 bytes.
+        // 0            32          64         96
```
derekbruening, line 59: Not sure what these are labeling?

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,163 @@
...
+                      << "\n";
+            exit(1);
+        }
+    }
+
+    void
+    unit_test_replace_which_way()
+    {
+        // Create and initialize a 4-way set associative cache with line size of 32 and
+        // total size of 128 bytes.
```
derekbruening, line 58: It's 256 below.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,163 @@
...
+        // 0            32          64         96
+        //  ----------------------------------------------
+        // |   way 0    |   way 1   |   way 2   |  way 3  |
+        // -----------------------------------------------
```
derekbruening, line 62: Not sure what the picture's intent is: did you mean to also draw 2 rows?  For 256 bytes in a cache with 32-byte lines, that's just 8 entries that will fit, and at 4 ways each, that's just 2 rows.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,163 @@
...
+        ref_1.data.type = TRACE_TYPE_READ;
+        ref_1.data.size = 1;
+        for (int i = 0; i < 256; i++) {
+            ref_1.data.addr = i;
+            request(ref_1);
+        }
+        // Access the ways in the following fashion. This sequence follows the sequence
+        // shown in https://github.com/DynamoRIO/dynamorio/issues/4881.
+        // Access the ways in 3("a"), 0("b"), 1("c"), 2("b") order.
+        ref_1.data.addr = 192; // Access way 3 ("a" in issue 4881).
```
derekbruening, line 75: I don't understand: why would this be way 3?  The cache is completely empty and so the first addition would go into probably way 0 -- looks like 0 from the code.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,163 @@
...
+        request(ref_1);
+
+        ref_1.data.addr = 0; // Access way 0 ("b" in issue 4881).
```
derekbruening, line 78: So would this be the same row as 192==0xc0: drop bottom 5 bits for 32-byte line; just 2 rows so 6th bit is all that matters which is 0 for both: looks like yes?  Could we add an assert by calling compute_block_idx()?

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,163 @@
...
+        request(ref_1);
+
+        ref_1.data.addr = 64; // Access way 1 ("c" in issue 4881).
```
derekbruening, line 81: Not following: I would expect the 1st four accesses to the same row to fill the ways in order 0,1,2,3 from the code.  Am I missing something?

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+        ref_1.data.addr = 128; // Access way 2 ("d" in issue 4881).
+        request(ref_1);
+
+        // Way 3 ("a") should be replaced.
+        assert(replace_which_way(0) == 3);
+
+        // Create and initialize an 8-way set associative cache with line size of 32 and
+        // total size of 256 bytes.
+        initialize_cache(8, 32, 256);
```
derekbruening, line 92: Is it supported to call init() again?  Doesn't this leak memory (the stats object, for one)?  Seems better to either invalidate the whole cache, or make a whole new instance of the class.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,163 @@
...
+        memref_t ref_2;
+        ref_2.data.type = TRACE_TYPE_READ;
+        ref_2.data.size = 1;
+        for (int i = 0; i < 256; i++) {
+            ref_2.data.addr = i;
+            request(ref_2);
+        }
+
+        // Access the ways in 1, 0, 2, 3, 4, 5, 7, 6 order.
+        ref_2.data.addr = 32; // Access way 1.
```
derekbruening, line 103: Same comment as above: I do not understand this.  How do you know the LRU will give you way 1 here?  Ditto below.

derekbruening (changes requested):

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
bete0: Done.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
bete0: Thanks for the catch. Based on our offline discussion the order 0, 1, 2, 3 is the correct access and we can't know which way we want to access.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
bete0: Done.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
bete0: Thanks for the catch. Done.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
bete0: Done. Removed it altogether.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
bete0: Done. Removed them.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
bete0: Done. Renamed it `associativity`.

clients/drcachesim/CMakeLists.txt
bete0: Done.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+                      << "\n";
+            exit(1);
+        }
+    }
+
+    void
+    unit_test_replace_which_way()
+    {
+        // Create and initialize a 4-way set associative cache with line size of 32 and
+        // total size of 256 bytes.
```
derekbruening, line 56: suggest: Instead of this separate comment which duplicates the numbers (and thus can get out of sync and require changing in two places), use parameter name comments:
```C
initialize_cache(/*associativity=*/4, /*line_size=*/32, /*total_size=*/256);
```

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+        initialize_cache(4, 32, 256);
+
+        memref_t ref_1;
+        ref_1.data.type = TRACE_TYPE_READ;
+        ref_1.data.size = 1;
+        // Access the cache line in the following fashion. This sequence follows the
+        // sequence shown in https://github.com/DynamoRIO/dynamorio/issues/4881.
```
derekbruening, line 63: The convention in the code base is to use "i#4881" to refer to an issue.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+        // Access the first row.
+        ref_1.data.addr = 0;
+        request(ref_1); // This accesses "a" in issue 4881.
+
+        ref_1.data.addr = 64;
```
derekbruening, line 68: I think it's worth asserting that this is the same index as address 0 via compute_block_idx to make it clear to the reader and as a sanity check.  Maybe it's simpler to use addresses 0,1,2,3 which are more clearly the same index.  See also the comment below about naming them.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+        //  way 0 ("a" in issue 4881): 3
+        //  way 1 ("b" in issue 4881): 2
+        //  way 2 ("c" in issue 4881): 1
+        //  way 3 ("d" in issue 4881): 0
+        // At this point way 0 ("a") has the highest counter so it should be replaced by
+        // the LRU policy.
+        assert(replace_which_way(0) == 0);
+
+        ref_1.data.addr = 0;
+        request(ref_1); // This replaces way 0 ("a").
```
derekbruening, line 87: No, this won't replace anything: this is a hit as its address matches what's there.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+        // At this point way 0 has been replaced(accessed) and way 1 has the highest
```
derekbruening, line 88: No, it wasn't replaced.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+        request(ref_1); // This accesses "d" in issue 4881.
+
+        // After the above accesses, the counters for each way will be as follows:
+        //  way 0 ("a" in issue 4881): 3
+        //  way 1 ("b" in issue 4881): 2
+        //  way 2 ("c" in issue 4881): 1
+        //  way 3 ("d" in issue 4881): 0
+        // At this point way 0 ("a") has the highest counter so it should be replaced by
+        // the LRU policy.
+        assert(replace_which_way(0) == 0);
```
derekbruening, line 84: Maybe this should be replace_which_way(compute_block_idx(0)) to again make it clear and avoid confusion over whether you're passing it an address vs an index.  (Even better `s/0/ADDRESS_A//`.)

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
derekbruening, line 88: > "way 1 has the highest counter"

Not sure thinking in terms of the counters inside the implementation is the right way to approach this test: this goes back to the black box test suggestion.  Which line is the least recently used is what we care about, not how that is computed internally.  If we invert the counters internally so 0 is old and 3 is new, it will still work properly, and the test should still apply.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+        request(ref_1); // This replaces way 0 ("a").
+        // At this point way 0 has been replaced(accessed) and way 1 has the highest
+        // counter.
+        //  way 0 ("a" in issue 4881): 0
+        //  way 1 ("b" in issue 4881): 3
+        //  way 2 ("c" in issue 4881): 2
+        //  way 3 ("d" in issue 4881): 1
+        assert(replace_which_way(0) == 1);
+
+        ref_1.data.addr = 64;
```
derekbruening, line 96: A symbolic name seems good: ADDRESS_B, and assert that all of them have the same block index.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+        ref_1.data.addr = 192;
+        request(ref_1); // This replaces way 3 ("d").
+        // At this point way 3 has been replaced(accessed) and way 0 has the highest
+        // counter.
+        //  way 0 ("a" in issue 4881): 3
+        //  way 1 ("b" in issue 4881): 2
+        //  way 2 ("c" in issue 4881): 1
+        //  way 3 ("d" in issue 4881): 0
+        assert(replace_which_way(0) == 0);
+    }
```
derekbruening, line 125: More test scenarios would be good.  The above does not actually test replacing anything: so testing that the right thing happens when new addresses come in and evict prior data would be good, e.g.  Other corner cases you can think of would be good too.  As separate scenarios with a new cache might be best.

Or if you prefer we could commit just the #4881 sequence first and add more separately, but please put in a `XXX i#4842: Add more test sequences` comment since otherwise the reader will look at this and assume that the author thinks this is complete.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+        ref_1.data.addr = 64;
+        request(ref_1); // This replaces way 1 ("b").
+        // At this point way 1 has been replaced(accessed) and way 2 has the highest
+        // counter.
+        //  way 0 ("a" in issue 4881): 1
+        //  way 1 ("b" in issue 4881): 0
+        //  way 2 ("c" in issue 4881): 3
+        //  way 3 ("d" in issue 4881): 2
+        assert(replace_which_way(0) == 2);
```
derekbruening, line 104: suggest: Something more concise that makes it easy to add more test sequences?  What if a helper created a memref, called request, and checked that afterward the right way was considered the least recently used.  So it might look like this with lower-case showing which one is least recently used:

```C
access_and_check_lru(ADDRESS_A, 1); // A x X X
access_and_check_lru(ADDRESS_B, 2); // A B x X
access_and_check_lru(ADDRESS_C, 3); // A B C x
access_and_check_lru(ADDRESS_D, 0); // a B C D
access_and_check_lru(ADDRESS_A, 1); // A b C D
access_and_check_lru(ADDRESS_A, 1); // A b C D
access_and_check_lru(ADDRESS_A, 1); // A b C D
access_and_check_lru(ADDRESS_E, 2); // A E c D
```

Or if something else makes more sense to you.  I actually find these long comment sequences in the current code hard to scan and quickly see the state changes that matter and that something more concise is easier to read and step through mentally, and also seems to make it simpler to add a bunch more test sequences.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
bete0: Understood. Updated the comments and added helper methods.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
bete0: Done.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
bete0: Added the comment to add more sequences in subsequent commits.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
bete0: Done. Added `access_and_check_lru` helper method.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,109 @@
...
+    int
+    get_block_index(int addr)
+    {
+        addr_t tag = compute_tag(addr);
+        int block_idx = compute_block_idx(tag);
+        return block_idx;
+    }
+
+    void
+    access_and_check_lru(int addr, int way_to_replace)
```
derekbruening, line 93: Make it clear that the 2nd param is *after* adding addr to the cache, and that it's just a check -- as it is doesn't it sound like this function is going to actively replace something?  So rename to `expected_replacement_way_after_access` or something.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,109 @@
...
+        access_and_check_lru(ADDRESS_A, 1); // A X X X
+        access_and_check_lru(ADDRESS_B, 2); // A B X X
+        access_and_check_lru(ADDRESS_C, 3); // A B C X
+        access_and_check_lru(ADDRESS_D, 0); // a B C D
+        access_and_check_lru(ADDRESS_A, 1); // A b C D
+        access_and_check_lru(ADDRESS_A, 1); // A b C D
+        access_and_check_lru(ADDRESS_A, 1); // A b C D
+        access_and_check_lru(ADDRESS_E, 2); // A E c D
+
+        // XXX i#4842: Add more test sequences.
```
derekbruening, line 69: It's not clear to me how new tests are going to fit in this existing code: they'll want a brand-new cache, or to invalidate the current one.  Their tests should probably be in separate functions.  Think about how to better structure this, probably renaming this function to reflect what it focuses on and moving the initialize_cache() to be called just once or something?  Or moving the test code to not be inside a cache instance so we can easily create a new cache for each test sequence: does this test code only need public functions?  What is your plan for more tests?

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+class cache_lru_test_t : public cache_lru_t {
+public:
+    void
+    unit_test_replace_which_way()
+    {
+        initialize_cache(/*associativity=*/4, /*line_size=*/32, /*total_size=*/256);
+
+        // Access the cache line in the following fashion. This sequence follows the
+        // sequence shown in https://github.com/DynamoRIO/dynamorio/issues/4881.
```
derekbruening, line 47: Repeating: the convention is "i#4881" rather than a URL

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,109 @@
...
+        const int ADDRESS_D = 192;
+        const int ADDRESS_E = 72;
+
+        assert(get_block_index(ADDRESS_A) == get_block_index(ADDRESS_B));
+        assert(get_block_index(ADDRESS_B) == get_block_index(ADDRESS_C));
+        assert(get_block_index(ADDRESS_C) == get_block_index(ADDRESS_D));
+        assert(get_block_index(ADDRESS_D) == get_block_index(ADDRESS_E));
+
+        // Lower-case letter shows the least recently used way.
+        access_and_check_lru(ADDRESS_A, 1); // A X X X
```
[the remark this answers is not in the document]
bete0: Done.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,109 @@
...
+class cache_lru_test_t : public cache_lru_t {
+public:
+    void
+    unit_test_replace_which_way()
+    {
+        initialize_cache(/*associativity=*/4, /*line_size=*/32, /*total_size=*/256);
+
+        // Access the cache line in the following fashion. This sequence follows the
+        // sequence shown in https://github.com/DynamoRIO/dynamorio/issues/4881.
+        const int ADDRESS_A = 0;
```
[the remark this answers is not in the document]
bete0: Done. Changed to `addr_t`.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
bete0: Done. Renamed it to `expected_replacement_way_after_access`

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
bete0: Restructured it. Moved the test sequence outside of `cache_lru_test_t` class. That class is standalone and can be instantiated by caches of varying sizes implemented outside similar to `unit_test_cache_lru_four_way()`.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
bete0: Done.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+class cache_lru_test_t : public cache_lru_t {
+public:
+    void
+    initialize_cache(int associativity, int line_size, int total_size)
+    {
+        caching_device_stats_t *stats = new cache_stats_t(line_size, "", true);
+        if (!init(associativity, line_size, total_size, nullptr, stats, nullptr)) {
+            std::cerr << "LRU cache failed to initialize"
+                      << "\n";
```
derekbruening, line 47: nit: Seems like it should be part of same string on prior line

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -0,0 +1,113 @@
...
+    void
+    access_and_check_lru(const addr_t addr,
+                         const int expected_replacement_way_after_access)
+    {
+        memref_t ref;
+        ref.data.type = TRACE_TYPE_READ;
+        ref.data.size = 1;
+        ref.data.addr = addr;
+        request(ref);
+        assert(replace_which_way(get_block_index(addr)) ==
```
derekbruening, line 69: We've been burned before using the provided assert and then running release-build tests where assert is compiled out and all the checks disappear.  We could use our own check, or could undefine NDEBUG.  Actually I see we have precedent for that:
```
suite/tests/linux/rseq.c:#undef NDEBUG
suite/tests/linux/rseq.c-#include <assert.h>
```
Maybe you could just insert the same thing before assert.h (and in the other unit test file).

derekbruening (approved):
Looks good.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
bete0: Done.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
bete0: `clang-format` moves it to the next line so better to stick with that?

derekbruening: ?  That does not sound right: the string is not long enough to be split: "LRU cache failed to initialize\n"

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
bete0: Done.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
bete0: Done.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
bete0: Done. Added assertion for block index.

clients/drcachesim/CMakeLists.txt
```diff
@@ -499,7 +499,8 @@ endif ()
 # Be sure to give the targets qualified test names ("tool.drcache*...").
```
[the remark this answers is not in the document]
bete0: Done.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
bete0: Done.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
bete0: Done. Removed it altogether.

clients/drcachesim/CMakeLists.txt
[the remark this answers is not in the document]
bete0: Done.

PR #5414, opened 2022-03-10
i#5195 win2019: Fix except-execution test

derekbruening:
Updates the security-win32.except-execution test to not make a raw
thread with a custom stack, which fails on GA CI win2019.  Running the
test code on the initial thread as-is works fine, for regular DR
instrumentation modes.

To avoid headaches with tabs and trailing spaces, updates the context
dumping to avoid those, and updates other templates that use that.

Issue: #5195

### DIFF
--- a/.github/workflows/ci-windows.yml
+++ b/.github/workflows/ci-windows.yml
@@ -1,5 +1,5 @@
 # **********************************************************
-# Copyright (c) 2020-2021 Google, Inc.  All rights reserved.
+# Copyright (c) 2020-2022 Google, Inc.  All rights reserved.
 # **********************************************************
 
 # Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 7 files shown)


derekbruening:
OK confirmed it's fixed for VS2019:
```
debug-internal-32: 243 tests passed, **** 6 tests failed, but ignoring 4 for i#2145: ****
	(ignore: i#2145) 	code_api|win32.hookerfirst 
	(ignore: i#2145) 	code_api|client.drx-test 
	code_api|client.tls 
	code_api|client.drsyms-test 
	(ignore: i#2145) 	code_api|client.pcache-use 
	(ignore: i#2145) 	code_api|client.drwrap-test-detach 
```

derekbruening:
I'll revert the ci-windows.yml and runsuite.cmake changes right before merging.  All jobs will fail then b/c of the tabs in the old file (but at least any future changes won't have to deal with that) and I'll have to do an override-merge.

abhinav92003 (approved):

Issue #5411, opened 2022-03-10
Eliminate drbbdup bookkeeping for single-case-forever uses


derekbruening:
With drbbdup inside drmemtrace (#3995) we're seeing drbbdup use
an extra ~35MB ("Peak vmm virtual memory in use" rstat) for ~150K blocks;
so ~140MB for ~600K blocks.  This is not horrible, but is not insignificant, 
and is unnecessary: if the user promises to never use more
than just one case (which memtrace can certainly do based on its own
runtime options at init time) drbbdup shouldn't need any bookkeeping.

johnfxgalea:
This seems okay with me. Note, it would require setting up dup options again for translations and traces, which might be expensive with respect to the user's use-case.

derekbruening:
> Note, it would require setting up dup options again for translations and traces, which might be expensive with respect to the user's use-case.

I'm not sure I understand: if you could spell this out?

derekbruening:
It is now clear that this overhead is the cause for out-of-memory failures we're seeing in a proprietary app: it has 2.5M blocks, so that's ~583MB extra just from drbbdup tracking even when duplication is turned off.  We had vmcode size set to 1G, and client heap is reachable for backward compatibility: meaning that drbbdup's bookkeeping took up over half the limited reachable space.  We can bump to 2G, but that's the absolute max and it precludes -vm_base_near_app and so incurs overheads.

Maybe we should change the client heap default, or change drbbdup to explicitly request unreachable heap.

johnfxgalea:
> > Note, it would require setting up dup options again for translations and traces, which might be expensive with respect to the user's use-case.
> 
> I'm not sure I understand: if you could spell this out?

If I understood you correctly, you would like to remove drbbdup_manager_t structures, which are associated with basic blocks, to improve memory performance. This is possible, provided dynamic handling is not enabled (although I believe I can optimise the structure's memory footprint when enabled as well).

However, this requires setting up a temporary manager, which in turn invokes a user callback, upon every basic block event, including for translations and traces. Currently, drbbdup limits this with the use of a hash-table, and only invokes the callback if the lookup fails.

johnfxgalea:
BTW, to be clear, the point I mentioned is not a blocker; just an observation.

derekbruening:
[...]

PR #5417, opened 2022-03-12
i#5411 bbdup mem: Use unreachable heap for per-block data

derekbruening:
Switch drbbdup's per-block bookkeeping to use unreachable heap instead
of reachable.  This data can really add up, taking up half a GB or
more on large applications -- which really eats into or exceeds the
limited reachable space.

Quick small local test:
Before:
```
                   Basic block fragments generated :              5169
                         Trace fragments generated :               527
              Peak vmm blocks for unreachable heap :               641
                Peak vmm blocks for reachable heap :               336
```
After:
```
                   Basic block fragments generated :              5169
                         Trace fragments generated :               527
              Peak vmm blocks for unreachable heap :               757
                Peak vmm blocks for reachable heap :               230
```
Issue: #5411

### DIFF
--- a/ext/drbbdup/drbbdup.c
+++ b/ext/drbbdup/drbbdup.c
@@ -288,13 +288,20 @@ drbbdup_add_copy(void *drcontext, instrlist_t *bb, instrlist_t *orig_bb)
 static drbbdup_manager_t *
 drbbdup_create_manager(void *drcontext, void *tag, instrlist_t *bb)
 {
-    drbbdup_manager_t *manager = dr_global_alloc(sizeof(drbbdup_manager_t));
+    /* This per-block memory can add up: for 2.5M basic blocks we can take up >512M
+     * of space, which if it's in the limited-size vmcode region is a problem.
+     * We thus explicitly request unreachable heap.
+     * XXX: Maybe DR should break compatibility and change the default.
+     */
+    drbbdup_manager_t *manager = dr_custom_alloc(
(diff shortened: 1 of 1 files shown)


ext/drbbdup/drbbdup.c
```diff
@@ -288,13 +288,20 @@ drbbdup_add_copy(void *drcontext, instrlist_t *bb, instrlist_t *orig_bb)
 static drbbdup_manager_t *
 drbbdup_create_manager(void *drcontext, void *tag, instrlist_t *bb)
 {
-    drbbdup_manager_t *manager = dr_global_alloc(sizeof(drbbdup_manager_t));
+    /* This per-block memory can add up: for 2.5M basic blocks we can take up >512M
+     * of space, which if it's in the limited-size vmcode region is a problem.
+     * We thus explicitly request unreachable heap.
+     * XXX: Maybe DR should break compatibility and change the default.
+     */
+    drbbdup_manager_t *manager = dr_custom_alloc(
```
johnfxgalea, line 296: Why don't you pass drcontext here?

derekbruening: I did in the first iteration but that API routine actually yells at you if you don't pass NULL for global heap.

johnfxgalea: Nit: Maybe add a comment to indicate we want to use the global heap.

johnfxgalea (approved):

ext/drbbdup/drbbdup.c
derekbruening: Done.

PR #5418, opened 2022-03-14
i#5195 win2019: Improve dbghelp.dll searching

derekbruening:
Searches both Program Files and Program Files (x86) for dbghelp.dll to
fix problems where 32-bit fails to find an SDK or VS version and ends
up using the system version which DR"s private loader has trouble
with.

Adds a CMake status message listing all the dghelp.dll paths found, to
help diagnose selection issues.

Issue: #5195


derekbruening:
win64 drcachesim.invariants failure is the type_is_instr assert: #3320

abhinav92003 (approved):

PR #5419, opened 2022-03-14
i#5411 bbdup mem: Shrink per-thread heap usage

derekbruening:
Moves all drbbdup per-thread memory to unreachable heap.

Moves the drbbdup dynamic case hit table to be dynamically allocated
based on a new drbbdup option "never_enable_dynamic_handling" not
being set.  This eliminates 128K per thread for drmemtrace, which
makes a big difference on large applications.

Sanity test:
Before:
```
  $ bin64/drrun -rstats_to_stderr -t drcachesim -offline -max_global_trace_refs 10K -- suite/tests/bin/client.annotation-concurrency libclient.annotation-concurrency.appdll.so A 4 64 3
              Peak threads under DynamoRIO control :                 5
              Peak vmm blocks for unreachable heap :               767
                Peak vmm blocks for reachable heap :               726
            Peak vmm virtual memory in use (bytes) :          10293248
```
After:
```
              Peak threads under DynamoRIO control :                 5
              Peak vmm blocks for unreachable heap :               736
                Peak vmm blocks for reachable heap :               250
            Peak vmm virtual memory in use (bytes) :           8183808
```
Issue: #5411

### DIFF
--- a/ext/drbbdup/drbbdup.c
+++ b/ext/drbbdup/drbbdup.c
@@ -1288,6 +1288,8 @@ drbbdup_insert_dynamic_handling(void *drcontext, void *tag, instrlist_t *bb,
 
     ASSERT(new_case_cache_pc != NULL,
            "new case cache for dynamic handling must be already initialised.");
+    DR_ASSERT_MSG(!opts.never_enable_dynamic_handling,
+                  "should not reach here if dynamic cases were disabled globally");
 
     /* Check whether case limit has not been reached. */
     if (drbbdup_do_dynamic_handling(manager)) {
@@ -1807,6 +1809,9 @@ drbbdup_handle_new_case()
     drbbdup_per_thread *pt =
         (drbbdup_per_thread *)drmgr_get_tls_field(drcontext, tls_idx);
 
+    DR_ASSERT_MSG(!opts.never_enable_dynamic_handling,
+                  "should not reach here if dynamic cases were disabled globally");
+
     /* Must use DR_MC_ALL due to dr_redirect_execution. */
     dr_mcontext_t mcontext;
     mcontext.size = sizeof(mcontext);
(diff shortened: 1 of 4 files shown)


ext/drbbdup/drbbdup.c
```diff
@@ -2024,9 +2043,13 @@ drbbdup_thread_exit(void *drcontext)
     if (is_thread_private)
         hashtable_delete(&pt->manager_table);
 
-    dr_thread_free(drcontext, pt->case_analysis_data,
+    dr_custom_free(drcontext, DR_ALLOC_THREAD_PRIVATE, pt->case_analysis_data,
                    sizeof(void *) * opts.non_default_case_limit);
-    dr_thread_free(drcontext, pt, sizeof(drbbdup_per_thread));
+    if (pt->hit_counts != NULL) {
+        dr_custom_free(drcontext, DR_ALLOC_THREAD_PRIVATE, pt->hit_counts,
```
johnfxgalea, line 2049: Add an assert here that opts.never_enable_dynamic_handling should be false.

johnfxgalea (approved):
Can you sprinkle some asserts in the code-base where opts.never_enable_dynamic_handling should be false.  Maybe somewhere close to the beginning of the following functions:

- init_fp_cache()
- drbbdup_handle_new_case()
- drbbdup_insert_dynamic_handling()

ext/drbbdup/drbbdup.c
derekbruening: Done.

derekbruening:
> Can you sprinkle some asserts in the code-base where opts.never_enable_dynamic_handling should be false. Maybe somewhere close to the beginning of the following functions:
> 
> * init_fp_cache()
> * drbbdup_handle_new_case()
> * drbbdup_insert_dynamic_handling()

Done.

derekbruening:
win64 failure is burst_traceopts/replaceall hitting #4486

PR #5420, opened 2022-03-14
i#5195 win2019: Relax drsyms-test output ordering

derekbruening:
Relaxes the ordering of symbols found through the dbghelp iterator to
handle orderings seen with VS2019.

This has the client.drsyms-test (and client.drsyms-testgcc) tests now
passing for both 32-bit and 64-bit on my VS2019 local setup.
Unfortunately the GA CI setup has further problems.

Issue: #5195


derekbruening:
Failure is api.detach-state #5123

abhinav92003 (approved):

Issue #5421, opened 2022-03-14
dual_map_file not protected from being closed by app


abhinav92003:
Today when running with `-satisfy_w_xor_x` we do not add the `dual_map_file` to the DR FD table. This means that the app can successfully close the file, which causes issues.

abhinav92003:
This has been observed to manifest as an OOM
```
Out of memory.  Program aborted.  Source E, type 0x0000000000000022, code 0x0000000000000000.>
```

PR #5422, opened 2022-03-14
i#5421: Prevent app from closing dual_map_file.

abhinav92003:
Adds the `dual_map_file` to the DR `fd_table`. This is to prevent the app from
closing this FD, which manifests as an OOM on an extending commit.

Unfortunately, due to constraints on the order of invocation of
`vmm_heap_unit_init` and `d_r_os_init` in `dynamorio_app_init_part_two_finalize`,
the `fd_table` is not initialized when the `dual_map_file` is created. This means that
`fd_table_add` will not really add it to the `fd_table`. So, we remember the
`dual_map_file_fd` and add it to `fd_table` when we create it.

Modifies the handling for the global log file to use the same mechanism, instead of
adding it explicitly in the DEBUG build.

Also adds a `-satisfy_w_xor_x` variant for the `client.file_io` test which attempts to
close DR FDs. It fails without adding `dual_map_file` to `fd_table`.

Also adds the same variant for the `common.loglevel` test where both, the global
log file and the `dual_map_file`, are recorded for  later `fd_table_add`.

Fixes: #5421

### DIFF
--- a/core/unix/os.c
+++ b/core/unix/os.c
@@ -335,10 +335,10 @@ static generic_table_t *fd_table;
  * XXX: try refactoring init code so that we don't have to track dual_map_file_fd
  * and add it later.
  */
-static int dual_map_file_fd;
-#ifdef DEBUG
+#define MAX_FD_ADD_PRE_HEAP 2
+static int fd_add_pre_heap[MAX_FD_ADD_PRE_HEAP];
+static int fd_add_pre_heap_flags[MAX_FD_ADD_PRE_HEAP];
 static int num_fd_add_pre_heap;
-#endif
 
 #ifdef LINUX
 /* i#1004: brk emulation */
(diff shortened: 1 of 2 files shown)


core/unix/os.c
```diff
@@ -1364,7 +1380,6 @@ os_slow_exit(void)
 
     if (doing_detach) {
         vsyscall_page_start = NULL;
-        IF_DEBUG(num_fd_add_pre_heap = 0;)
```
derekbruening, line 1367: So this should be back to 0 and that's why this isn't needed?  Maybe add an assert to make this clearer.

derekbruening (approved):

core/unix/os.c
abhinav92003: Done.

PR #5436, opened 2022-03-29
Added missed unlock in case of failed binary search

kuhanov:
There are some cases with null start and end app_pc-s during profiling java on AArch64.
As result, we caught hang in release build.
Kirill


derekbruening (approved):
Thank you for the fix

PR #5443, opened 2022-04-04
i#5250: Add alt-bitwidth DR lib path config param

### LINKED ISSUE #5250: Obtain other-bitwidth libdynamorio from the configuration
Driven by https://github.com/DynamoRIO/dynamorio/pull/5172#issuecomment-991684246

This is a feature request to obtain the path used for the other-bitwidth libdynamorio for cross-arch execve from the configuration, instead of editing the current self path using build-time path assumptions.

Today cross-arch execve (a 32-bit process launching a 64-bit binary, or vice versa) uses `dynamorio_alt_arch_path`, which is set up in `get_dynamo_library_bounds()` at init time using the "lib32" and "lib64" paths set in the build system (`INSTALL_LIB_{X86,X64}`) to edit the path of the current self library.  That hardcoded path edit causes problems with package installs.

The proposal here is to obtain the other-bitwidth path from the configuration: the same place we get the options and log directory.

We would add something like `DYNAMORIO_VAR_AUTOINJECT{32,64}`, have drrun set both for every app just like it sets `DYNAMORIO_VAR_RUNUNDER` and other parameters, and add both to `env_to_propagate` for execve so that child processes inherit the configuration in environment variables (overridden by config files of course).  The existing config file defaults could be used for systemwide or package installs to set the value.

derekbruening:
Adds a new config parameter DYNAMORIO_ALTINJECT for specifying the
path to the alternate-bitwidth DR library for cross-bitwidth execve.

If the parameter is set, DR uses it; otherwise it falls back to the
current hardcoded path assumptions.

drrun and friends automatically set the parameter if the other
bitwidth library is found.

Adds an automated test by augmenting the "linux.execve{32,64}" test
from a 64-bit build which does a test-and-build.  The renamed
"linux.xarch" test now does one build (for efficiency) and runs a
script which executes both (32<->64) directions for each of the
existing assumed-layout strategy and the new
config-points-at-alt-bitwidth strategy.

Also tested manually:
$ bin64/drrun -v -use_alt_dll ~/dr/releases/DynamoRIO-Linux-9.0.0/lib32/debug/libdynamorio.so -trace_threshold 51 -- suite/tests/bin/linux.execve64 ~/dr/git/build_x86_dbg_tests/suite/tests/bin/linux.execve-sub32

Fixes #5250


derekbruening:
@fmoessbauer this PR adds a config entry to specify the other-bitwidth library.

core/config.h
```diff
@@ -46,6 +46,9 @@ d_r_config_init(void);
 void
 d_r_config_exit(void);
 
+bool
```
bete0, line 49: nit: do we need this function in header as opposed to making it static? I don't see it being used outside of `config.c`.

libutil/dr_config.c
```diff
@@ -1495,6 +1495,73 @@ dr_registered_process_iterator_stop(dr_registered_process_iterator_t *iter)
...
+#ifndef PARAMS_IN_REGISTRY
+    /* shift rest of file up, overwriting old value, so we can append new value */
+    read_config_ex(f, DYNAMORIO_VAR_AUTOINJECT, NULL, 0, true);
+#endif
+    convert_to_tchar(wpath, dr_lib_path, MAXIMUM_PATH);
+    NULL_TERMINATE_BUFFER(wpath);
+    status = write_config_param(IF_REG_ELSE(proc_policy, f),
+                                PARAM_STR(DYNAMORIO_VAR_AUTOINJECT), wpath);
+    DO_ASSERT(status == DR_SUCCESS);
+    if (status != DR_SUCCESS)
```
bete0, line 1529: `if (status != DR_SUCCESS)` We assert with on `DO_ASSERT(status == DR_SUCCESS)` so if that assert passes, there won't be a need to check `if status != DR_SUCCESS`, right?

libutil/dr_config.c
```diff
@@ -1495,6 +1495,73 @@ dr_registered_process_iterator_stop(dr_registered_process_iterator_t *iter)
...
+#ifndef PARAMS_IN_REGISTRY
+        /* shift rest of file up, overwriting old value, so we can append new value */
+        read_config_ex(f, DYNAMORIO_VAR_ALTINJECT, NULL, 0, true);
+#endif
+        convert_to_tchar(wpath, dr_alt_lib_path, MAXIMUM_PATH);
+        NULL_TERMINATE_BUFFER(wpath);
+        status = write_config_param(IF_REG_ELSE(proc_policy, f),
+                                    PARAM_STR(DYNAMORIO_VAR_ALTINJECT), wpath);
+        DO_ASSERT(status == DR_SUCCESS);
+        if (status != DR_SUCCESS)
```
bete0, line 1542: Same comment as line 1529.

```
DO_ASSERT(status == DR_SUCCESS);
return status;
```

core/unix/os.c
```diff
@@ -9319,32 +9332,71 @@ get_dynamo_library_bounds(void)
...
+        LOG(GLOBAL, LOG_VMAREAS, 1, PRODUCT_NAME " alt arch filepath: %s\n",
+            dynamorio_alt_arch_filepath);
+        LOG(GLOBAL, LOG_VMAREAS, 1, PRODUCT_NAME " alt arch path: %s\n",
+            dynamorio_alt_arch_path);
+    } else {
+        /* Construct a path under assumptions of build-time path names. */
+        strncpy(dynamorio_alt_arch_path, dynamorio_library_path,
+                BUFFER_SIZE_ELEMENTS(dynamorio_alt_arch_path));
+        /* Assumption: libdir name is not repeated elsewhere in path */
+        libdir =
```
bete0, line 9379: nit:
```
char *libdr = 
    strstr(dynamorio_alt_arch_path, IF_X64_ELSE(DR_LIBDIR_X64, DR_LIBDIR_X86));
```
and delete line 9355?

core/config.h
derekbruening: unix/os.c has `ASSERT(d_r_config_initialized())`

libutil/dr_config.c
derekbruening: This does seem odd: it's a pattern elsewhere in the file.  Since it handles the failure case by returning status it seems the assert should be removed.  Probably it was meant as a something like core's ASSERT_CURIOSITY.  I'm removing the assert.

libutil/dr_config.c
derekbruening: Removing the assert.

core/unix/os.c
derekbruening: Done.

bete0 (approved):

fmoessbauer:
@derekbruening : Thanks for this feature. With this now implemented, I'll have a second look at the packaging PR.

fmoessbauer:
I just confirmed that this makes it possible to inject from a system-wide install using just a wrapper:
`./bin64/drrun -use_dll <...> -use_alt_dll <...> -- /bin/echo "test"`

This also works together with tools, but there a warning is emitted in case the alt-arch version of a client is not available. This is fixed in #5543

PR #5450, opened 2022-04-06
i#3995 multi-window: Write separate raw files per window

derekbruening:
Adds a drmemtrace feature under a new on-by-default -split_windows
option to create a separate subdirectory with a separate set of raw
files per traced window.  This avoids disk space issues with a single
file, and splitting at the raw stage is relatively simple for regular
drmemtrace usage (though not as simple for external users of the file
i/o redirection).

Files in raw/window.NNNN/ subdirectories are mirrored in
trace/window.NNNN/ subdirectories upon being post-processed.
Post-processing handles just the first window by default; the others
must be explicitly passed as input directories in separate
post-processing invocations.

This changes the non-window behavior to not create an output file
until tracing starts, which necessitated changing the
tool.drcacheoff.delay-func test to check for no output files as a
slightly different type of test.

Adds a test of split-file offline windows.

Fixes an infinite loop bug in raw2trace hit when a file is truncated:
hit while the windows were buggy and missing footers.

Issue: #3995


clients/drcachesim/common/directory_iterator.cpp
```diff
@@ -128,8 +129,24 @@ directory_iterator_t::is_directory(const std::string &path)
 }
 
 bool
-directory_iterator_t::create_directory(const std::string &path)
+directory_iterator_t::create_directory(const std::string &path_in)
 {
-    drfront_status_t res = drfront_create_dir(path.c_str());
-    return res == DRFRONT_SUCCESS;
+    std::string path = path_in;
```
bete0, line 134: nit: Do we need this local variable? Can't we use `path_in` directly?

bete0 (approved):

clients/drcachesim/common/directory_iterator.cpp
derekbruening: path_in is const so the Windows replacement fails without a writable local copy.

Issue #5427, opened 2022-03-16
Add snappy compression to drmemtrace output


derekbruening:
Xref https://github.com/DynamoRIO/dynamorio/issues/2001#issue-175552118 where zlib with Z_BEST_SPEED was tried on the memtrace_x86 sample and was faster than writing uncompressed to an HDD but slower than an SSD (which was why it was not committed, though maybe it should have been under an option).

Xref #790 where general write-compressed-file support for all clients is requested.

This issue is about adding snappy compression for the direct output of drmemtrace.  Snappy should out-perform zlib Z_BEST_SPEED (though with worse compression) and should be a net win for an SSD as well.

This issue can also cover generalizing this snappy compression into a library and to use it by default in the tracing samples.

In addition to improving overall performance, this will also shrink the disk space needed to store raw files for offline traces.

For online traces: this still should be a net win, shrinking the pipe throughput.  We already have an offline file reader for snappy; we'd need to use that for the online reader.

derekbruening:
I implemented snappy compression on drmemtrace raw files and did a quick perf test on spec2k6 mcf test on my laptop which has an SSD:
```
$ /usr/bin/time bin64/drrun -t drcachesim -offline -raw_compress none -- ~/spec2006/mcf-test/mcf_base.gcc-64bit ~/spec2006/mcf-test/inp.in 
5.85user 7.31system 0:18.31elapsed 71%CPU (0avgtext+0avgdata 167772maxresident)k
1656inputs+28450072outputs (11major+40110minor)pagefaults 0swaps
$ ls -sh ~/dr/git/build_x64_rel_tests/drmemtrace.mcf_base.gcc-64bit.528525.5144.dir/raw
total 14G
14G drmemtrace.mcf_base.gcc-64bit.528525.7489.raw    0 funclist.log  12K modules.log

$ /usr/bin/time bin64/drrun -t drcachesim -offline -- ~/spec2006/mcf-test/mcf_base.gcc-64bit ~/spec2006/mcf-test/inp.in 
53.32user 2.89system 0:56.77elapsed 99%CPU (0avgtext+0avgdata 167848maxresident)k
0inputs+7528296outputs (0major+40156minor)pagefaults 0swaps
$ ls -sh ~/dr/git/build_x64_rel_tests/drmemtrace.mcf_base.gcc-64bit.527083.0507.dir/raw/
total 3.6G
[...]

PR #5451, opened 2022-04-06
i#5427 compress: Add snappy_nocrc drmemtrace option

derekbruening:
Adds a new drmemtrace option -raw_compress snappy_nocrc.
This requests snappy compression but omitting checksums.
This is 2.5x faster than snappy with checksums.
This currently uses local-only snappy chunk types; we should
propose adding them to the public spec.

Adds a sanity test.

Issue: #5427

### DIFF
--- a/clients/drcachesim/reader/snappy_file_reader.cpp
+++ b/clients/drcachesim/reader/snappy_file_reader.cpp
@@ -74,13 +74,18 @@ snappy_reader_t::check_magic()
 bool
 snappy_reader_t::read_data_chunk(uint32_t size, chunk_type_t type)
 {
-    char *buf =
-        (type == COMPRESSED_DATA) ? compressed_buf_.data() : uncompressed_buf_.data();
-    size_t max_size = (type == COMPRESSED_DATA) ? max_compressed_size_ : max_block_size_;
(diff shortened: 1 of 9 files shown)


clients/drcachesim/reader/snappy_file_reader.cpp
```diff
@@ -110,21 +114,23 @@ snappy_reader_t::read_data_chunk(uint32_t size, chunk_type_t type)
...
     }
 
-    size_t dummy;
-    uint32_t checksum = mask_crc32(src_->Peek(&dummy), src_->Available());
-    if (checksum != read_checksum) {
-        ERRMSG("Checksum failure on snappy block. Want %x, got %x\n", read_checksum,
-               checksum);
-        return false;
+    if (has_checksum) {
+        size_t dummy;
```
bete0, line 127: nit: disrespectful term

bete0 (approved):

clients/drcachesim/reader/snappy_file_reader.cpp
derekbruening: Replaced with "ignored" in this pre-existing code.

Issue #3044, opened 2018-06-09
Add encoder/decoder support for Arm's Scalable Vector Extension


fhahn:
Armv8-a's SVE is a big extension requiring a few changes to the AArch64 backend. At first, I think we need new register types for scalable vector (Z) and predicate (P) registers as well as support for the new SVE encodings. 

https://developer.arm.com/docs/ddi0584/latest/arm-architecture-reference-manual-supplement-the-scalable-vector-extension-sve-for-armv8-a

Xref #2626

AssadHashmi:
Currently we can handle 10 SVE opcodes (out of 314) in the codec. Now that we are preparing to implement the rest of SVE, the notion of a variable vector register length in DynamoRIO needs to be addressed. Specifically, how will clients and tools like DrMemory and drcachesim work correctly on SVE hardware with different SVE vector lengths?

Currently we have `OPSZ_SCALABLE` and `OPSZ_SCALABLE_PRED` for SVE vector and predicate registers returned by `reg_get_size()`. This needs to be converted to an `OPSZ_` representing the SVE vector and predicate register size on the currently executing AArch64 implementation, i.e. one of:
```
OPSZ_128 OPSZ_256 OPSZ_384 OPSZ_512 OPSZ_640 OPSZ_768 OPSZ_896 OPSZ_1024 OPSZ_1152 OPSZ_1280 OPSZ_1408 OPSZ_1536 OPSZ_1664 OPSZ_1792 OPSZ_1920 OPSZ_2048
```
One approach would be for DynamoRIO to execute the `RDVL` instruction (https://developer.arm.com/documentation/ddi0602/2022-06/SVE-Instructions/RDVL--Read-multiple-of-vector-register-size-to-scalar-register-) on startup so that when `reg_get_size()` is called for SVE vector or predicate registers, the appropriate `OPSZ_` is returned.

@derekbruening is this enough to cover all of DynamoRIO's clients/tools requirements? Is there another approach?

derekbruening:
Having the actual register size in the operand in the IR at runtime does seem necessary but maybe not sufficient.  Tools like DrMemory which are shadowing registers may want to know the maximum register sizes at initialization time to optimize their shadow allocations.  Maybe an API routine like dr_mcontext_zmm_fields_valid() where a client can ask which parts of the mcontext are live for the current hardware, with some way to refer to the different simd vector sizes?

[...]

PR #5453, opened 2022-04-08
i#4393 New AArch64 codec implementation

AssadHashmi:
This patch is an example of the code which will be generated from a
machine readable specification (MRS) to decode and encode AArch64
instructions from v8.1 onwards.

It is provided for review and discussion purposes, in order to resolve
any issues which may arise and to make visible what changes to expect.

This patch does not include the MRS, the parser or generator, just an
example of the target code which we intend to generate, based on the
v8.1 SQRDMLAH instruction.

Issue: #4393, #3044, #2626


derekbruening:
It seems reasonable at a glance.  I'm not sure exactly how to evaluate: nothing jumps out as problematic.  If we'll get auto-generation of INSTR_CREATE_ macros and easier splitting of opcodes it sounds good to me.

Issue #38, opened 2014-11-27
attach injection on Linux


derekbruening:
_From [derek.br...@gmail.com](https://code.google.com/u/117968039472581148324/) on February 24, 2009 10:01:18_

this was PR 204490

focusing on Linux as Windows has issues that are best solved with a kernel
driver, unless we think we can rely on backward decoding heuristics or
don't mind losing control for a while

we do have some issues on Linux:
- ability to suspend threads before we have control of them: we will
  probably rely on ptrace (xref issue #37 )
- determining the state of sharing of CLONE_\* among threads: should
  probably use a modify-and-observe approach

_Original issue: http://code.google.com/p/dynamorio/issues/detail?id=38_

derekbruening:
Trying to summarize the status of attach:

Attaching to an already-running process is not yet officially supported.  There is a
prototype implementation on Linux, along with fully-implemented attach as
well as full detach through the start/stop API, but we have not had the
resources to finish off a ptrace implementation (#38).

On Windows there are non-trivial technical obstacles (#725).

What was implemented was ptrace injection into a fresh process (#37):
75597c3a888ea1f4314a45c57ac6c22e026aba77

As well as taking over all threads at DR init time (for #722):
5e052f7cb256945cbb0397b67aa97f34fd17f6f2

Rrecent work has made the start/stop API's attach and thread takeover more
robust, and has added an internally-triggered detach on Linux
(959211659d07f35e356dc9aac0f451f389da5e73
and several later refactorings and tweaks), which should make it easier to
implement ptrace-based attach and detach.

The remaining work would be extending the ptrace fresh injection to
attaching to an existing process, and perhaps some details around segment
stealing.

illera88:
When is the attach feature expected to be completed and ready to be used in Linux systems? 

Thank you

derekbruening:
If you'd like to contribute to this feature it may be best to first coordinate with:
https://groups.google.com/forum/#!searchin/dynamorio-devs/iannillo%7Csort:relevance/dynamorio-devs/xFejqJpHET4/swlaCygnBwAJ

M3m3M4n:
(Xref #5019) Comment added per [request](https://github.com/DynamoRIO/dynamorio/pull/5019#discussion_r688641351)

After we ptrace a process, it is stopped with SIGSTOP. This is the point where we inject our code, but...
[...]

PR #5460, opened 2022-04-12
i#38 attach: Enable attach on AArch64

derekbruening:
Implements missing functionality for ptrace attach on AArch64 and
AArch32: generated code sequences were previously x86-only;
-skip_syscall handling only supported x86; and AArch64 does not
support PTRACE_POKEUSER or PTRACE_PEEKUSER.

For AArch32, Thumb vs Arm mode require multiple steps: clearing LSB to
point at the path used as data via a call; switching to Arm mode for
DR's _start; setting the LSB of the initial app PC.

For AArch32, additionally fixes an encoder error where the opcode is
queried before copying a needs-no-encoding instruction.  This is
required for the instruction used to hold data for injection.

Tweaks the disassembler to leave a level 0 instr alone, again to
better handle the data-only insruction used for injection.

Enables the client.attach test on AArch64 and AArch32.  For AArch32,
it needs -skip_syscall.  Long-term we want that on by default
everywhere but we want explicit tests that hit it on all platforms
first.

Issue: #38

### DIFF
--- a/core/unix/injector.c
+++ b/core/unix/injector.c
@@ -909,7 +909,7 @@ enum { REG_PC_OFFSET = offsetof(struct USER_REGS_TYPE, REG_PC_FIELD) };
 static inline size_t
 system_call_length(dr_isa_mode_t mode)
 {
-#    ifdef X86
+#    if defined(X86)
     ASSERT(INT_LENGTH == SYSCALL_LENGTH);
     ASSERT(SYSENTER_LENGTH == SYSCALL_LENGTH);
     return SYSCALL_LENGTH;
@@ -922,7 +922,7 @@ system_call_length(dr_isa_mode_t mode)
 #    endif
 }
 
-#    ifdef X86
+#    if defined(X86)
 /* X86s are little endian */
 enum { SYSCALL_AS_SHORT = 0x050f, SYSENTER_AS_SHORT = 0x340f, INT80_AS_SHORT = 0x80cd };
 #    elif defined(AARCH64)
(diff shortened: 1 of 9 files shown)


derekbruening:
Will add to commit description:

Tested manually on an AArch32 machine.  Unfortunately the client.attach test is not trival to set up under QEMU with its multiple command lines and background processes so that is left as beyond the scope of this PR and part of #4719.

core/unix/injector.c
```diff
@@ -1222,24 +1293,40 @@ injectee_run_get_retval(dr_inject_info_t *info, void *dc, instrlist_t *ilist)
...
+    if (!info->wait_syscall)
+        offset = system_call_length(app_mode);
+#    ifdef AARCH64
+    /* POKEUSER is not available. */
+    ptr_int_t saved_pc = regs.REG_PC_FIELD;
+    regs.REG_PC_FIELD = (ptr_int_t)(pc + offset);
+    r = our_ptrace_setregs(info->pid, &regs);
+    if (r < 0)
+        return r;
+    regs.REG_PC_FIELD = saved_pc;
```
bete0, line 1306: nit: If this line is not part of the if/else block, move it up before `if` or down after the `else`?

core/unix/injector.c
```diff
@@ -1169,6 +1225,22 @@ injectee_run_get_retval(dr_inject_info_t *info, void *dc, instrlist_t *ilist)
...
+    /* Get register state before executing the shellcode. */
+    r = our_ptrace_getregs(info->pid, &regs);
+    if (r < 0)
+        return r;
+
+#    ifdef X86
+    app_mode = IF_X64_ELSE(DR_ISA_AMD64, DR_ISA_IA32);
+#    elif defined(AARCH64)
+    app_mode = DR_ISA_ARM_A64;
+#    elif defined(ARM)
```
bete0, line 1239: nit: inconsistency between using #ifdef and defined in the same block.

bete0 (approved):

core/unix/injector.c
derekbruening: This is part of this `if`, restoring the app's pc for the post-execution restore at the end of the function.

core/unix/injector.c
derekbruening: Fixed, here and in several other similar additions in this PR.

PR #5461, opened 2022-04-12
i#5458: Switch suspend signal to SIGFPE

### LINKED ISSUE #5458: Thread takeover on attach fails b/c SIGUSR2 is blocked: switch to another signal?
Attaching (via ptrace #38 or for statically-linked DR) to a process that has masked most non-fatal signals fails to take over the rest of the app threads.  We could try to use ptrace to take them over but that is difficult for the static-link case.  Or we could switch from SIGUSR2 to a signal less likely to be masked, like SIGFPE.  We would distinguish from a synchronous signal by looking at si->code (set as far back as 2.2 kernel; some other siginfo fields were unreliable back then but not this one) and other fields.

E.g., this is hit attaching to mysqld, which blocks all non-fatal signals.  The ptrace attach succeeds but then DR's takeover times out and fails.

derekbruening:
Changes the signal that DR uses to suspend a thread from SIGUSR2,
which is sometimes blocked by the app at attach time, to SIGFPE, which
as a fatal normally-synchronous signal is less likely to be blocked.

Manually tested on an attach to mysqld which failed with SIGUSR2 but
succeeds now.

Fixes #5458

### DIFF
--- a/api/docs/debugging.dox
+++ b/api/docs/debugging.dox
@@ -1,5 +1,5 @@
 /* ******************************************************************************
- * Copyright (c) 2010-2021 Google, Inc.  All rights reserved.
+ * Copyright (c) 2010-2022 Google, Inc.  All rights reserved.
  * ******************************************************************************/
 
 /*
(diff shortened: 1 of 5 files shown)


derekbruening:
Unfortunately it's looking like QEMU does not handle DR sending SIGFPE via SYS_kill.

bete0 (approved):

Issue #5456, opened 2022-04-11
Calling `cache_fifo::replace_which_way()` directly results in a change in FIFO cache state


bete0:
While working on a unit test for `clients/drcachesim/simulator/cache_fifo.cpp` (https://github.com/DynamoRIO/dynamorio/pull/5454), I discovered that calling `cache_fifo::replace_which_way()` right after accessing a cache line (by calling `request()`) would update the cache state. 

For the cache replacement unit tests, we validate the expected `way` is replaced after an access by first sending a `request()` and then checking the `way` to be replaced on the next access matches the expected way as shown below (from https://github.com/DynamoRIO/dynamorio/pull/5454):
```
 void
 access_and_check_fifo(const addr_t addr,
                          const int expected_replacement_way_after_access)
 {
        memref_t ref;
        ref.data.type = TRACE_TYPE_READ;
        ref.data.size = 1;
        ref.data.addr = addr;
        request(ref);
        assert(replace_which_way(get_block_index(addr)) ==
                                 expected_replacement_way_after_access);
}
```
However calling `replace_which_way()` changes the state of the FIFO cache by updating the `way` to be replaced which results in the wrong `way` to be replaced on the next access.

This is not necessarily a bug in the FIFO cache replacement implementation but makes testing the implementation harder and potentially could result in unexpected behavior if `cache_fifo::replace_which_way()` is called directly somewhere else.

1. Checkout PR https://github.com/DynamoRIO/dynamorio/pull/5454, build and run drcachesim unit test and observe the test fails
2. Update the test cases from https://github.com/DynamoRIO/dynamorio/pull/5454/files#diff-b7acbebc334f2800e156dcb14d7ad91ea8b243fee66b9cf3011430d2605081b9R165-R168 as follows and observe the test passes
```
  cache_fifo_test.access_and_check_fifo(ADDRESS_A, 1); // A x X X
  cache_fifo_test.access_and_check_fifo(ADDRESS_B, 3); // A B x X
  cache_fifo_test.access_and_check_fifo(ADDRESS_C, 1); // A B C x
  cache_fifo_test.access_and_check_fifo(ADDRESS_D, 3); // a B C D
```

PR #5454, opened 2022-04-08
i#4842 drcachesim unit test: Add cache_fifo unit test

bete0:
Adds unit tests for FIFO cache simulator

Adds a method get_next_way_to_replace() that just returns the victim 'way' 
without a side effect on the cache state. This makes unit testing the cache
replacement policies convenient and avoids unexpected behavior. 

Adds cache_policy_test_t template for testing multiple cache replacement policies. 

Adds get_block_index() method that returns the block index for a given
addr_t address.
    
Issue: #4842, #5456

PR #5465, opened 2022-04-15
i#5456 cache_fifo: Add a method that returns next way to be replaced

bete0:
Adds a non-state changing method get_next_way_to_replace() that
just returns the victim 'way' without updating the cache state. This
makes unit testing the FIFO cache replacement policy convenient.

Issue: #5456

### DIFF
--- a/clients/drcachesim/simulator/cache_fifo.h
+++ b/clients/drcachesim/simulator/cache_fifo.h
@@ -52,6 +52,8 @@ protected:
     access_update(int line_idx, int way) override;
     int
     replace_which_way(int line_idx) override;
+    int
+    get_next_way_to_replace(int block_idx);
 };
 
 #endif /* _CACHE_FIFO_H_ */
--- a/clients/drcachesim/simulator/cache_fifo.cpp
+++ b/clients/drcachesim/simulator/cache_fifo.cpp
@@ -86,3 +86,14 @@ cache_fifo_t::replace_which_way(int block_idx)
     }
     return -1;
 }
+
+int
+cache_fifo_t::get_next_way_to_replace(int block_idx)
+{
+    for (int i = 0; i < associativity_; i++) {
+        if (get_caching_device_block(block_idx, i).counter_ == 1) {
+            return i;
+        }
+    }
+    return -1;
+}


clients/drcachesim/simulator/cache_fifo.h
```diff
@@ -52,6 +52,8 @@ class cache_fifo_t : public cache_t {
     access_update(int line_idx, int way) override;
     int
     replace_which_way(int line_idx) override;
+    int
+    get_next_way_to_replace(int block_idx);
```
derekbruening, line 56: Declare as const

derekbruening, line 56: It seems worth a comment explaining why there are 2 functions which seem to do the same thing.  The const will help, but a comment will additionally make it clear.

derekbruening, line 56: There's no use of this function.  Generally we'd like a test or use of a new function in the same PR: it seems unusual to just add a function without anything calling it.  Maybe it is best to add this code to the test PR which presumably uses this function.

Or, if replace_which_way can have side effects, maybe this const query should be in the base class, and the LRU test should use it instead of replace_which_way, in this PR.

bete0: It's being used by this draft PR: https://github.com/DynamoRIO/dynamorio/pull/5454 but agree it makes sense to put it in that PR in the base class. Will do that.

bete0: Actually it can't be in the base class as each cache replacement algorithm's replace policy is different.

bete0: > Or, if replace_which_way can have side effects, maybe this const query should be in the base class, and the LRU test should use it instead of replace_which_way, in this PR.

`replace_which_way()` implementation in LRU cache doesn't have a side effect as of the current LRU cache test cases.

derekbruening: > Actually it can't be in the base class as each cache replacement algorithm's replace policy is different.

This *implementation* wouldn't be in the base class, but a virtual method would be, and each subclass would override it with their particular implementation.  All tests would then call this new const method, rather than the one that can have side effects, as the cleaner and safer way to query.

bete0: Understood. Will make that change in PR https://github.com/DynamoRIO/dynamorio/pull/5454.

PR #5466, opened 2022-04-16
i#5455 path-ignore-directory: Do not try to execute directories

### LINKED ISSUE #5455: USAGE ERROR when a directory named like an executable is present in the PATH variable [Minor bug]
**Describe the bug**
When there is a directory named like an executable in one of the locations the PATH variable points to, normally it is ignored and a real executable is searched instead.
For example if there is a directory called `vim` in a user defined PATH directory (like `~/.bin`) it is ignored and the real vim application will be run.
DynamoRIO does not handle this case and will try to run the directory resulting in an error message like this:
`./drrun -- vim`
`<Application /home/derek/.bin/vim (543) DynamoRIO usage error : Failed to read app ELF headers.  Check path and architecture.>`

This can be confusing because just typing `vim` without `./drrun` does work in this case.

**To Reproduce**
1. Have a directory named like an executable in one of the PATH variables
2. Run `./drrun -- [name of the executable/directory]`

Please also answer these questions:
 - What happens when you run without any client?
 Same
 - What happens when you run with debug build ("-debug" flag to drrun/drconfig/drinject)?
 Same

**Expected behavior**
DynamoRIO should ignore directories and traverse through the other possible locations in the PATH variable.

**Versions**
6e88738383e78c0be9c3dda373e43f311425b862
 No
- What operating system version are you running on? 
Manjaro Linux and Debian 11
- Is your application 32-bit or 64-bit?
64 bit

Additional context
This does also happen with non-executable files in `PATH`

derchr:
I changed the required access mode to `DRFRONT_EXEC` to ciruumvent trying to execute files that aren't executable.
A new check in `drfront_searchenv()` is introduced to determine if the path in option is in fact a directory and should not be executed.

In this implementiation we are in the akward situation of calling `stat()` twice (first time in `drfront_access()` and then in `drfront_searchenv()`). An alternative approach would be to add a new variant for `drfront_access_mode_t` called `DRFRONT_ISDIR`. Though `ISDIR` isn't really an access mode.

So what would the preferable method be?

```
Fixes a bug where DynamoRIO tries to execute a directory or a file that
is not marked as executable when it is present in a directory contained
in the PATH environment variable.

Fixes #5455
```

### DIFF
--- a/libutil/dr_frontend_unix.c
+++ b/libutil/dr_frontend_unix.c
@@ -137,6 +137,8 @@ drfront_searchenv(const char *fname, const char *env_var, OUT char *full_path,
     char realpath_buf[PATH_MAX]; /* realpath hardcodes its buffer length */
     drfront_status_t status_check = DRFRONT_ERROR;
     bool access_ret = false;
+    int r;
+    struct stat st;
 
     if (ret == NULL)
         return DRFRONT_ERROR_INVALID_PARAMETER;
@@ -166,15 +168,21 @@ drfront_searchenv(const char *fname, const char *env_var, OUT char *full_path,
         NULL_TERMINATE_BUFFER(tmp);
         /* realpath checks for existence too. */
         if (realpath(tmp, realpath_buf) != NULL) {
-            status_check = drfront_access(realpath_buf, DRFRONT_EXIST, &access_ret);
+            status_check = drfront_access(realpath_buf, DRFRONT_EXEC, &access_ret);
             if (status_check != DRFRONT_SUCCESS) {
                 *ret = false;
                 return status_check;
             } else if (access_ret == true) {
-                *ret = true;
-                snprintf(full_path, full_path_size, "%s", realpath_buf);
-                full_path[full_path_size - 1] = '\0';
-                return DRFRONT_SUCCESS;
+                // XXX: An other option to prevent calling stat() twice
+                // could be a new variant DRFRONT_NOTDIR for drfront_access_mode_t
+                // that drfront_access() then takes into account.
+                r = stat(realpath_buf, &st);
+                if (r != -1 && !S_ISDIR(st.st_mode)) {
+                    *ret = true;
+                    snprintf(full_path, full_path_size, "%s", realpath_buf);
+                    full_path[full_path_size - 1] = '\0';
+                    return DRFRONT_SUCCESS;
+                }
             }
         }
         cur = next + 1;
--- a/suite/tests/libutil/frontend_test.c
+++ b/suite/tests/libutil/frontend_test.c
@@ -1,5 +1,5 @@
 /* **********************************************************
- * Copyright (c) 2014-2019 Google, Inc.  All rights reserved.
+ * Copyright (c) 2014-2022 Google, Inc.  All rights reserved.
  * Copyright (c) 2003 VMware, Inc.  All rights reserved.
  * **********************************************************/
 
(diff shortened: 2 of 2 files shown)


libutil/dr_frontend_unix.c
```diff
@@ -166,15 +168,18 @@ drfront_searchenv(const char *fname, const char *env_var, OUT char *full_path,
...
+            status_check = drfront_access(realpath_buf, DRFRONT_EXEC, &access_ret);
             if (status_check != DRFRONT_SUCCESS) {
                 *ret = false;
                 return status_check;
             } else if (access_ret == true) {
-                *ret = true;
-                snprintf(full_path, full_path_size, "%s", realpath_buf);
-                full_path[full_path_size - 1] = '\0';
-                return DRFRONT_SUCCESS;
+                r = stat(realpath_buf, &st);
```
derekbruening, line 176: > In this implementiation we are in the akward situation of calling stat() twice (first time in drfront_access() and then in drfront_searchenv()). An alternative approach would be to add a new variant for drfront_access_mode_t called DRFRONT_ISDIR. Though ISDIR isn't really an access mode.
> 
> So what would the preferable method be?

Hmm, adding `DRFRONT_ISDIR = 0x08` or something could break existing code that wants to include directories: it would have to be negated `DRFRONT_NOTDIR = 0x08` so the unset behavior is the legacy behavior: and then it would be less useful b/c it couldn't be used to find a directory, only to exclude them...so maybe `drfront_access_ex` is better which looks at `DRFRONT_ISDIR` while `drfront_access` ignores it.  Hmm.  I could also live with the extra `stat`.  How about just a `XXX` comment here about possibly avoiding the extra `stat` by either the `DRFRONT_NOTDIR` or the  `DRFRONT_ISDIR` plus `drfront_access_ex`.

libutil/dr_frontend_unix.c
```diff
@@ -166,15 +168,18 @@ drfront_searchenv(const char *fname, const char *env_var, OUT char *full_path,
...
+                if (r != -1 && !S_ISDIR(st.st_mode)) {
+                    *ret = true;
+                    snprintf(full_path, full_path_size, "%s", realpath_buf);
+                    full_path[full_path_size - 1] = '\0';
+                    return DRFRONT_SUCCESS;
+                }
```
derekbruening, line 182: Could we include a test in suite/tests/libutil/frontend_test.c?  It already creates a dir: could add that to PATH with setenv and then test whether it shows up in searchenv?

libutil/dr_frontend_unix.c
derchr: I added a comment.

libutil/dr_frontend_unix.c
derchr: I added a test case that tests this. It's a bit convoluted though because the current directory is handled differently.
I'm not sure if it could be written in a bit more elegant way.

derekbruening:
> derchr force-pushed the i5455-path-ignore-directory branch from a9105cf to 27b7b35 Compare4 hours ago

Please don't force-push: it ruins the history.  Now the reviewer can't see just the new changes.  We put warnings against force-pushing on several docs pages (https://dynamorio.org/page_contributing.html, https://dynamorio.org/page_code_reviews.html).  The squash-and-merge will clean up all commits at the end.

libutil/dr_frontend_unix.c
derekbruening: Looks reasonable.

suite/tests/libutil/frontend_test.c
```diff
@@ -53,11 +58,22 @@ main()
         return -1;
     }
 
-    res = drfront_remove_dir("test_dir");
+    path = getenv("PATH");
+    strncpy(path_buf, "test_dir:", 9);
+    strncat(path_buf, path, 512 - 10);
```
derekbruening, line 63: nit: Better to use `sizeof(path_buf)` instead of repeating the 512 so they'll stay in sync; even better to use our BUFFER_SIZE_ELEMENTS macro, but that may not be available w/ the current headers here.

derekbruening (approved):

suite/tests/libutil/frontend_test.c
derekbruening: Well I tried to tweak this and get the test to run on Windows but there are still Windows and Mac issues and I have no more time.

derekbruening:
OK got it green and merged.

derchr:
Thank you for merging and taking your time.

PR #5468, opened 2022-04-18
Fix bitmap_find_free_sequence() result check

maksqwe:
"index < 0" always false because index is unsigned integer


derekbruening (approved):
Thank you for the patch.

Issue #2311, opened 2017-03-28
process-directed signals incorrectly delivered to threads blocking them


derekbruening:
DR by default intercepts all signals and keeps them all unblocked in all
threads.  It then emulates the kernel upon receipt of a signal, deciding
whether to deliver it or queue it as a pending signal, based on the
receiving thread's mask.  However, for process-directed signals, delivery
is supposed to only consider those threads that have the signal unblocked.
Since DR will not deliver a signal to a different thread from the one that
received it, DR fails to properly emulate the kernel.

To handle all scenarios, including clients using alarm signals, we can't
rely on installing the app's mask and must emulate the kernel.  We'll have
to examine other threads and deliver the signal to one of them.  This may
not be simple to do efficiently yet race-free.

egrimley:
Could we insist (in a future version of the API) that clients not use alarm signals, except via DR's API?

derekbruening:
> Could we insist (in a future version of the API) that clients not use alarm signals, except via DR's API?

Maybe you mean only supporting client alarm signals in sideline threads,
which would be a restriction of the current API.

If we only support clients using alarm signals in sideline threads could we
set the alarm signal block state natively?  But what about timer_create
where *any* signal could be an alarm signal, and what about non-timer
signals sent to the whole process hitting this same blocking emulation
problem.  Certainly we can't have the app's block state for SIGSEGV or
SIGUSR2 be the kernel state.

egrimley:
Life would be much simpler if we could install the app's mask, except perhaps for a small number of signal numbers that we would choose to leave unblocked whatever the app says. We could do that if all client timers were set up using DR's API so that we could force them to use real-time signal number 49, say, which the app is unlikely to be using, though we could still attempt to distinguish app uses of signal 49 from our uses of signal 49 to provide a second line of defence. (Of course the number "49" would have to be chosen at run time as the set of available signal numbers potentially depends on the kernel.) Are there other reasons for not installing the app's mask?

derekbruening:
Using a real-time number is not possible on Mac OSX, AFAIK.  On Linux it requires kernel 2.6+.  (Most of the DR signal handling was written before 2.6 was in widespread use.)  There, as you mention, the corner case still exists where emulation is needed or some complex swapping of signal numbers.  If we have to write emulation for OSX it becomes a question of simpler overall cross-platform code vs OS-specific internally-simpler code.

derekbruening:
[...]

PR #5472, opened 2022-04-21
i#2311 sigmask: Reroute blocked signal if unblocked thread exists

derekbruening:
Adds a solution to the problem of the real signal mask having a signal
unblocked in a thread where the app has it blocked, while another app
thread has it unblocked, when the kernel delivers a group-wide signal.

To avoid checking all threads on every blocked signal, which requires
heavyweight locks that can't be acquired from our signal handler, we
maintain a count of unmasked threads for each signal number.  This is
shared across the handler group and is updated using atomic
operations.

On arrival of a signal blocked by the app, we first detect whether it
is whole-group-directed, by examining the si_code values.  If it is,
and if the unmasked thread count is > 0, we want to re-route it: but
we can't acquire the locks here, so we mark it unblocked, unlink the
interrupted fragment (if any), and head back through dispatch with the
signals_pending flag set.

Once in the pending-signal code from dispatch, we grab the
thread_initexit_lock and walk the threads, looking for a thread with
this signal unblocked.  For synchronization with other threads' masks,
we can't have a mutex used on every access as we need lock-free reads
from our signal handler.  Instead, we have a mutex used only on writes
(only done by the owning thread) and on reads from other threads (just
this rerouting code).

Once we find a target thread, we send it a new signal and drop the
current one.  This will properly interrupt any syscall.  This should
be an asynchronous signal, so a new later copy should be fine.

We handle sigqueue's extra value by using SYS_rt_tgsigqueueinfo where
available.

Adds a new test linux.sigmask which tests both kill() and sigqueue().
Enables previously-disabled alarms in the api.static_signal to further
test the fix.

Future improvements include keeping the actual mask matching the app's
mask, if DR and clients don't care about that signal: which will
require an API addition.

Issue: #2311

### DIFF
--- a/core/unix/os.c
+++ b/core/unix/os.c
@@ -3529,6 +3529,23 @@ thread_signal(process_id_t pid, thread_id_t tid, int signum)
 #endif
 }
 
+/* This is not available on all platforms/kernels and so may fail. */
+bool
+thread_signal_queue(process_id_t pid, thread_id_t tid, int signum, void *value)
+{
+#ifdef MACOS
+    return false;
+#else
+    kernel_siginfo_t info;
+    memset(&info, 0, sizeof(info));
+    info.si_signo = signum;
+    info.si_code = SI_QUEUE;
+    info.si_value.sival_ptr = value;
+    /* SYS_rt_sigqueueinfo is on 2.6.31+ so we expect failure on older kernels. */
+    return dynamorio_syscall(SYS_rt_tgsigqueueinfo, 4, pid, tid, signum, &info) == 0;
+#endif
+}
+
 static bool
 known_thread_signal(thread_record_t *tr, int signum)
 {
@@ -5018,6 +5035,7 @@ ignorable_system_call_normalized(int num)
 #    endif
     case SYS_rt_sigtimedwait:
     case SYS_rt_sigqueueinfo:
+    case SYS_rt_tgsigqueueinfo:
     case SYS_rt_sigsuspend:
 #    ifdef SYS_signalfd
     case SYS_signalfd:
@@ -7564,6 +7582,7 @@ pre_system_call(dcontext_t *dcontext)
 #    endif
     case SYS_rt_sigtimedwait: /* 177 */
     case SYS_rt_sigqueueinfo: /* 178 */
+    case SYS_rt_tgsigqueueinfo:
 #endif
     case IF_MACOS_ELSE(SYS_sigpending, SYS_rt_sigpending): { /* 176 */
         /* FIXME i#92: handle all of these syscalls! */
--- a/core/unix/signal.c
+++ b/core/unix/signal.c
@@ -302,7 +302,7 @@ static void
 execute_default_from_dispatch(dcontext_t *dcontext, int sig, sigframe_rt_t *frame);
 
 static bool
-reroute_to_unmasked_thread(dcontext_t *dcontext, kernel_siginfo_t *siginfo, int sig);
+reroute_to_unmasked_thread(dcontext_t *dcontext, sigframe_rt_t *frame, int sig);
 
 static bool
 handle_alarm(dcontext_t *dcontext, int sig, kernel_ucontext_t *ucxt);
@@ -4679,7 +4679,7 @@ record_pending_signal(dcontext_t *dcontext, int sig, kernel_ucontext_t *ucxt,
              * can call this now.
              */
             ASSERT_NOT_TESTED();
-            handled = reroute_to_unmasked_thread(dcontext, &frame->info, sig);
+            handled = reroute_to_unmasked_thread(dcontext, frame, sig);
             if (!handled)
                 blocked = true;
         } else {
@@ -6716,8 +6716,13 @@ execute_default_from_dispatch(dcontext_t *dcontext, int sig, sigframe_rt_t *fram
 }
 
 static bool
-reroute_to_unmasked_thread(dcontext_t *dcontext, kernel_siginfo_t *siginfo, int sig)
+reroute_to_unmasked_thread(dcontext_t *dcontext, sigframe_rt_t *frame, int sig)
 {
+    kernel_siginfo_t *siginfo = &frame->info;
+    sigcontext_t *sc = get_sigcontext_from_rt_frame(frame);
+    if (!signal_is_process_wide(dcontext, siginfo, (byte *)sc->SC_XIP,
+                                (byte *)sc->SC_XSP))
+        return false;
     bool found = false;
     thread_record_t **trecs;
     int num_threads;
@@ -6738,7 +6743,7 @@ reroute_to_unmasked_thread(dcontext_t *dcontext, kernel_siginfo_t *siginfo, int 
         d_r_mutex_lock(&tgt_info->sigblocked_lock);
         if (!kernel_sigismember(&tgt_info->app_sigblocked, sig)) {
             LOG(THREAD, LOG_ASYNCH, 2,
-                "re-routing signal %d to thread " TIDFMT " where it is unblocked\n", sig,
+                "rerouting signal %d to thread " TIDFMT " where it is unblocked\n", sig,
                 trecs[i]->id);
             /* It is simpler to send a new signal to the target thread, rather than
              * trying to copy this same signal frame over there.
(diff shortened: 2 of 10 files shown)


derekbruening:
To add to commit description:

The added test hangs at both the kill and sigqueue points without those fixes.

core/unix/signal.c
```diff
@@ -1036,6 +1120,12 @@ signal_thread_inherit(dcontext_t *dcontext, void *clone_record)
             info->sighand->is_shared = true;
             d_r_mutex_lock(&info->sighand->lock);
             info->sighand->refcount++;
+            for (i = 1; i <= MAX_SIGNUM; i++) {
+                if (!EMULATE_SIGMASK(info, i))
+                    continue;
+                if (!kernel_sigismember(&info->app_sigblocked, i))
+                    ATOMIC_INC(int, info->sighand->threads_unmasked[i]);
```
abhinav92003, line 1127: Does this mean we keep a common count for the cloned child and the parent? But the parent and child are different processes, so signals shouldn't be rerouted between them right?

core/unix/signal.c
```diff
@@ -2227,18 +2362,41 @@ signal_swap_mask(dcontext_t *dcontext, bool to_app)
...
+         * redo with the latest blocked signals.
+         */
+        for (int i = 1; i <= MAX_SIGNUM; i++) {
+            if (!EMULATE_SIGMASK(info, i))
+                continue;
+            if (kernel_sigismember(&info->app_sigblocked, i))
+                ATOMIC_INC(int, info->sighand->threads_unmasked[i]);
+        }
+        d_r_mutex_lock(&info->sigblocked_lock);
         unblock_all_signals(&info->app_sigblocked);
```
abhinav92003, line 2385: Can you explain this else-block, and the case when it is invoked? Why do we need to increment and then decrement the unmasked thread count.

core/unix/signal.c
```diff
@@ -2227,18 +2362,41 @@ signal_swap_mask(dcontext_t *dcontext, bool to_app)
...
             return;
         }
+        d_r_mutex_lock(&info->sigblocked_lock);
         sigprocmask_syscall(SIG_SETMASK, &info->app_sigblocked, NULL,
                             sizeof(info->app_sigblocked));
+        d_r_mutex_unlock(&info->sigblocked_lock);
     } else {
+        /* Normally we get here just once at takeover, but if we have threads
+         * going native we don't consider them unmasked as we can't rely on their
+         * mask not changing: so we undo their counts when the come back here and
```
abhinav92003, line 2375: s/the/they/

abhinav92003 (approved):

core/unix/os.c
```diff
@@ -3529,6 +3529,23 @@ thread_signal(process_id_t pid, thread_id_t tid, int signum)
...
+#ifdef MACOS
+    return false;
+#else
+    kernel_siginfo_t info;
+    memset(&info, 0, sizeof(info));
+    info.si_signo = signum;
+    info.si_code = SI_QUEUE;
+    info.si_value.sival_ptr = value;
+    /* SYS_rt_sigqueueinfo is on 2.6.31+ so we expect failure on older kernels. */
+    return dynamorio_syscall(SYS_rt_tgsigqueueinfo, 4, pid, tid, signum, &info) == 0;
```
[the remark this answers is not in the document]
derekbruening: That is the parameter count (since complex to use vararg in asm).

suite/tests/linux/sigmask.c
```diff
@@ -0,0 +1,153 @@
+/* **********************************************************
+ * Copyright (c) 2016-2022 Google, Inc.  All rights reserved.
```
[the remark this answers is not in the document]
derekbruening: I started out by copying an existing file so there are likely lines here from earlier years so I just kept the range.

suite/tests/linux/sigmask.expect
```diff
@@ -0,0 +1,6 @@
+sending 10
+in handler for signal 10
+sending 28 with value
+in handler for signal 28 from -1 value 0xdeadbeef
```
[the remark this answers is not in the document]
derekbruening: SI_QUEUE == -1

core/unix/signal_private.h
```diff
@@ -392,6 +392,14 @@ typedef struct _sighand_info_t {
...
     kernel_sigaction_t *action[SIGARRAY_SIZE];
     bool we_intercept[SIGARRAY_SIZE];
+    /* For handling masked-for-app-but-not-for-DR signals.  Any time we receive
+     * a signal in a thread for which it is blocked, we need to know whether it was
+     * a "process"-wide signal and whether some other thread has it unblocked.
+     * To avoid heavyweight locks every time, we keep an atomic-access counter of
+     * unmasked threads for each signal, for the CLONE_SIGHAND group (typically
+     * whole process).
+     */
+    int threads_unmasked[SIGARRAY_SIZE];
```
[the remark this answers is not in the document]
derekbruening: Yes, it is potentially fragile.  The assert at exit helped and with the logging I added I was able to debug several issues without too much trouble, but it will always be a little messy.  But, I couldn't come up with a cleaner solution so I think this is the best we have for now.

Added a comment by app_sigblocked.

core/unix/signal.c
```diff
@@ -4457,20 +4639,59 @@ record_pending_signal(dcontext_t *dcontext, int sig, kernel_ucontext_t *ucxt,
...
+             * By definition as a process-wide signal this should be asynchronous
+             * and delayable.
+             */
+            LOG(THREAD, LOG_ASYNCH, 2,
+                "will reroute process-wide signal to unblocked thread\n");
+            blocked = false;
+            reroute = true;
+        }
+    }
+    if (blocked) {
```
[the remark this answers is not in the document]
derekbruening: Yes, because setting blocked to false requires entering other conditions below.  Added a comment.

core/unix/signal.c
```diff
@@ -952,7 +1033,10 @@ signal_info_exit_sigaction(dcontext_t *dcontext, thread_sig_info_t *info,
     memset(&act, 0, sizeof(act));
     act.handler = (handler_t)SIG_DFL;
     kernel_sigemptyset(&act.mask); /* does mask matter for SIG_DFL? */
+    DOLOG(4, LOG_ASYNCH, dump_unmasked(dcontext, __FUNCTION__););
     for (i = 1; i <= MAX_SIGNUM; i++) {
+        /* No synch here b/c we're exiting. */
+        ASSERT(info->sighand->threads_unmasked[i] == 0);
```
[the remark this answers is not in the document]
derekbruening: Yes it was very useful.

core/unix/signal.c
derekbruening: This code is for two entities that share a signal handler table, created by SYS_clone with CLONE_SIGHAND.  By definition this is the group that will share signals and should have signals rerouted between them.  It used to be you could create weird mixtures of sharing with separate address spaces but identical handlers but they've locked that down in recent kernels, so nowadays this is always two "threads".

core/unix/signal.c
derekbruening: Done.

core/unix/signal.c
derekbruening: This line is setting app_sigblocked to a new absolute value.  The only way to update the counts is to first remove all the counts from our old value of app_sigblocked and then re-increment for the new values.  This is also how we do it for sigprocmask with SIG_SETMASK.  I added a sentence making it clear that unblock_all_signals *writes* to app_sigblocked.

abhinav92003: Thanks. I was thinking only about takeover, but from your comment I understand that this is also required for threads that were previously under DR control, went back native, and then again under DR control. That could be caused by multiple calls to dr_app_start I suppose. From code I see this can also happen on a "call to a native module"?

core/unix/os.c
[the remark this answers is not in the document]
abhinav92003: Ah. Should've realised that myself.

core/unix/signal.c
derekbruening: There is a -native_exec feature which was used to run .NET and Java natively as a stopgap to get apps working that had just a little dynamic scripting.  It has lots of caveats of course.

PR #5476, opened 2022-04-27
i#5474: Eliminate stat glibc 2.33 dependence

### LINKED ISSUE #5474: Reduce glibc dependences in drrun to operate on older distros
While the core libdynamorio.so has no library dependences, drrun does.  Building drrun on a more recent machine and then running on something older like Debian 11 (glibc 2.31) results in:

```
/build_a64_rel/bin64/drrun: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /build_a64_rel/bin64/drrun)
```

This issue covers trying to improve binary interoperability of drrun.

derekbruening:
Eliminates a dependence on glibc 2.33 coming from the use of stat() in
drfrontendlib.  This is done by using the dr_stat_syscall() from drlibc.

Fixes #5474

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -611,6 +611,11 @@ if (NOT UNIX)
     add_definitions(-D_X86_)
   endif (X64)
 endif (NOT UNIX)
+if (UNIX)
+  # Ensure we can export dr_stat_syscall() from drlibc and get everyone to agree
+  # that struct stat64 is the stat struct.
+  add_definitions(-D_LARGEFILE64_SOURCE)
+endif ()
 
 # Set up assembly support and CMAKE_CPP.
 # Note that in cmake < 2.6.4, I had to fix a bug in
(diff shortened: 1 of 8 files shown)


derekbruening:
I have mixed feelings here:

+ It is not ideal to mix a raw syscall in there...but really it's the abysmal support for targeting older systems provided by Linux toolchains that's the problem.  There should be a simple linker flag for this.  The method of using .symvar IMHO is worse than the raw syscall.

+ It would be nice to have a test: enable the glibc 2.17 check all the time, for our own uses -- but that seems bad for others building DR for local use.  So we leave this w/o a test.  A one-off change once every few years to make our own local binary copying easier isn't so bad.

We'd have to solve this in a few years when our package builder here hit this same stat problem: so may as well solve it now while it helps our own local use cases.

abhinav92003 (approved):

PR #5478, opened 2022-05-02
i#5427 compress: Add zlib output to drmemtrace offline raw files

derekbruening:
Adds support for compressing drmemtrace offline raw trace files with
zlib as they are written out by the tracer.  Two forms of zlib
compression are supported: zlib or gzip.  Both use Z_BEST_SPEED, but
the extra headers make gzip slower.  These can be enabled by setting
-raw_compress to "zlib" or "gzip".

Adds a reader for zlib.

Fixes compression interactions with window files: setup and teardown
was not being done on a symmetric per-file basis.

For each of zlib and gzip, adds a test without windows and a test with
windows to ensure file teardown coincides with compression teardown.

Issue: #5427


abhinav92003 (approved):

derekbruening:
win32 failure is setthreadcontext #3213

PR #5479, opened 2022-05-03
i#5427 compress: Add lz4 drmemtrace raw compression

derekbruening:
Adds a new -raw_compress lz4 option.  lz4 out-performs even
snappy_nocrc, and it has knobs to support a range of i/o constraints.
However, it has no allocator parameterization, so it cannot be used
when drmemtrace is statically linked into the app.

Adds an lz4 reader and tests.

Updates the build docs and the GA CI installation to include
liblz4-dev.  Fixes the cross-compile jobs to also properly include
zlib and libsnappy.

Issue: #5427


derekbruening:
I didn't create this as a chained PR vs #5478 b/c IIRC Github wasn't running all the tests for non-vs-master PR's; I need to test the GA CI cross-compile compression lib changes here.

clients/drcachesim/common/lz4_istream.h
```diff
@@ -0,0 +1,146 @@
...
+        return *gptr();
+    }
+    std::iostream::pos_type
+    seekoff(std::iostream::off_type off, std::ios_base::seekdir dir,
+            std::ios_base::openmode which = std::ios_base::in) override
+    {
+        if (dir == std::ios_base::cur &&
+            ((off >= 0 && gptr() + off < egptr()) ||
+             (off < 0 && gptr() + off >= eback())))
+            gbump(off);
```
bete0, line 114: Should this return 0 after the call to gbump since gbump is void?

bete0 (approved):

clients/drcachesim/common/lz4_istream.h
derekbruening: The return value is the current position, which will not be 0 (unless the caller seek-ed to 0): the gptr() - eback() looks correct since gbump should update gptr.

PR #5485, opened 2022-05-05
i#5427 compress: Turn -raw_compress lz4 on by default

derekbruening:
For non-static drmemtrace, -raw_compress is now set to lz4 by default,
as that is nearly always faster than no compression or any other
compression scheme we've tried.

If lz4 is not enabled in the build, and again drmemtrace is not
static, snappy_nocrc is the default.

Otherwise, compression is disabled by default, just like today.

Inverts the raw-lz4 tests to become raw-none tests.

Updates the documentation.

Issue: #5427

### DIFF
--- a/suite/tests/CMakeLists.txt
+++ b/suite/tests/CMakeLists.txt
@@ -3584,11 +3584,9 @@ if (BUILD_CLIENTS)
       torunonly_drcacheoff(raw-gzip ${ci_shared_app} "-raw_compress gzip" "" "")
       set(tool.drcacheoff.raw-gzip_expectbase "offline-simple")
     endif ()
-    if (liblz4)
-      # lz4 is on by default so we test no compression here.
-      torunonly_drcacheoff(raw-none ${ci_shared_app} "-raw_compress none" "" "")
-      set(tool.drcacheoff.raw-none_expectbase "offline-simple")
-    endif ()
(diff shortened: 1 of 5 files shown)


suite/tests/CMakeLists.txt
```diff
@@ -3585,8 +3585,9 @@ if (BUILD_CLIENTS)
       set(tool.drcacheoff.raw-gzip_expectbase "offline-simple")
     endif ()
     if (liblz4)
```
abhinav92003, line 3587: If we're testing no compression, we don't need the liblz4 guard.

suite/tests/CMakeLists.txt
```diff
@@ -3657,10 +3658,11 @@ if (BUILD_CLIENTS)
       set(tool.drcacheoff.windows-gzip_expectbase "offline-windows-simple")
     endif ()
     if (liblz4)
```
abhinav92003, line 3660: Same here.

abhinav92003 (approved):

suite/tests/CMakeLists.txt
derekbruening: Done.

suite/tests/CMakeLists.txt
derekbruening: Done.

PR #5491, opened 2022-05-11
i#5474 Add FEATURE_ detection for AArch64

AssadHashmi:
This patch reads from the ID_AA64ISAR0_EL1 features register and sets
the following if supported:
FEATURE_AES    FEATURE_PMULL FEATURE_SHA1    FEATURE_SHA256
FEATURE_SHA512 FEATURE_CRC32 FEATURE_LSE     FEATURE_RDM
FEATURE_SM3    FEATURE_SM4   FEATURE_DotProd FEATURE_FHM
FEATURE_FlagM2 FEATURE_RNG   FEATURE_SHA3    FEATURE_FlagM

Example output in logfile:
Processor features:
 ID_AA64ISAR0_EL1 = 0x0000000000011120
   Processor has FEATURE_AES
   Processor has FEATURE_PMULL
   Processor has FEATURE_SHA1
   Processor has FEATURE_SHA256
   Processor has FEATURE_CRC32

ID_AA64ISAR1_EL1 and ID_AA64PFR0_EL1 are also read. The features for
these will be implemented in the next patch.

Issues: #5474, #1569

### DIFF
--- a/core/arch/aarch64/aarch64.asm
+++ b/core/arch/aarch64/aarch64.asm
@@ -115,12 +115,17 @@ GLOBAL_LABEL(unexpected_return:)
         JUMP  GLOBAL_REF(unexpected_return)
         END_FUNC(unexpected_return)
 
-/* All CPU ID registers are accessible only in privileged modes. */
-        DECLARE_FUNC(cpuid_supported)
-GLOBAL_LABEL(cpuid_supported:)
-        mov      w0, #0
+/* bool mrs_id_reg_supported(void)
+ * Checks for support of the MRS instr by attempting to read Instruction Set
+ * Attribute Register 0. Some older kernels on v8.0 systems do not support
+ * this, raising a SIGILL.
+ */
+        DECLARE_FUNC(mrs_id_reg_supported)
+GLOBAL_LABEL(mrs_id_reg_supported:)
+        mrs     x0, ID_AA64ISAR0_EL1
+        mov     w0, #1
         ret
-        END_FUNC(cpuid_supported)
+        END_FUNC(mrs_id_reg_supported)
 
 /* void call_switch_stack(void *func_arg,             // REG_X0
  *                        byte *stack,                // REG_X1
--- a/core/arch/aarch64/proc.c
+++ b/core/arch/aarch64/proc.c
@@ -38,12 +38,59 @@ static int num_simd_saved;
 static int num_simd_registers;
 static int num_opmask_registers;
 
+const static int num_feature_registers = sizeof(features_t) / sizeof(uint64);
+
+#define MRS(REG, IDX)                                                            \
+    do {                                                                         \
+        if (IDX > (num_feature_registers - 1))                                   \
+            CLIENT_ASSERT(false, "Reading undefined AArch64 feature register!"); \
+        asm("mrs %0, " #REG : "=r"(isa_features[IDX]));                          \
+    } while (0);
+
+void
+read_feature_regs(uint64 isa_features[])
+{
+    MRS(ID_AA64ISAR0_EL1, AA64ISAR0);
+    MRS(ID_AA64ISAR1_EL1, AA64ISAR1);
+    MRS(ID_AA64PFR0_EL1, AA64PFR0);
+}
+
+static void
+get_processor_specific_info(void)
+{
+    uint64 isa_features[num_feature_registers];
+
+    /* FIXME i#5474: Catch and handle SIGILL if MRS not supported.
+     * Placeholder for some older kernels on v8.0 systems which do not support
+     * this, raising a SIGILL.
+     */
+    if (!mrs_id_reg_supported()) {
+        ASSERT_CURIOSITY(false && "MRS instruction unsupported");
+        SYSLOG_INTERNAL_WARNING("MRS instruction unsupported");
+        return;
+    }
+
+    /* Reads instruction attribute and preocessor feature registers
+     * ID_AA64ISAR0_EL1, ID_AA64ISAR1_EL1 and ID_AA64PFR0_EL1.
+     */
+    read_feature_regs(isa_features);
+    cpu_info.features.flags_aa64isar0 = isa_features[AA64ISAR0];
+    cpu_info.features.flags_aa64isar1 = isa_features[AA64ISAR1];
+    cpu_info.features.flags_aa64pfr0 = isa_features[AA64PFR0];
+}
+
+#define LOG_FEATURE(feature)       \
+    if (proc_has_feature(feature)) \
+        LOG(GLOBAL, LOG_TOP, 1, "   Processor has " #feature "\n");
+
 void
 proc_init_arch(void)
 {
     num_simd_saved = MCXT_NUM_SIMD_SLOTS;
     num_simd_registers = MCXT_NUM_SIMD_SLOTS;
     num_opmask_registers = MCXT_NUM_OPMASK_SLOTS;
+
+    get_processor_specific_info();
 
     /* When DR_HOST_NOT_TARGET, get_cache_line_size returns false and does
      * not set any value in given args.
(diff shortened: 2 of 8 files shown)


core/arch/aarch64/proc.c
```diff
@@ -38,6 +38,55 @@ static int num_simd_saved;
...
 
+#ifndef DR_HOST_NOT_TARGET
+
+const static int num_feature_registers = sizeof(features_t) / sizeof(uint64);
+
+#    define MRS(REG, IDX)                                                            \
+        do {                                                                         \
+            if (IDX > (num_feature_registers - 1))                                   \
+                CLIENT_ASSERT(false, "Reading undefined AArch64 feature register!"); \
+            asm("mrs %0, " #REG : "=r"(isa_features[IDX]));                          \
```
derekbruening, line 49: Better to avoid this implicit parameter: maybe define this inside read_feature_regs()?  Or pass this in as an explicit argument.

core/arch/aarch64/aarch64.asm
```diff
@@ -115,12 +115,17 @@ GLOBAL_LABEL(unexpected_return:)
...
         END_FUNC(unexpected_return)
 
-/* All CPU ID registers are accessible only in privileged modes. */
-        DECLARE_FUNC(cpuid_supported)
-GLOBAL_LABEL(cpuid_supported:)
-        mov      w0, #0
+/* bool mrs_id_reg_supported(void)
+ * Checks for support of the MRS instr by attempting to read Instruction Set
+ * Attribute Register 0. Some older kernels on v8.0 systems do not support
+ * this, raising a SIGILL.
```
derekbruening, line 121: Hmm, I think DR is using `mrs` in multiple other places already?

core/arch/aarch64/proc.c
```diff
@@ -53,14 +102,87 @@ proc_init_arch(void)
         LOG(GLOBAL, LOG_TOP, 1, "Unable to obtain cache line size");
     }
 
-    /* FIXME i#1569: NYI */
+#ifndef DR_HOST_NOT_TARGET
+    get_processor_specific_info();
+
+    if (d_r_stats->loglevel > 0 && (d_r_stats->logmask & LOG_TOP) != 0) {
```
derekbruening, line 108: Use `DOLOG(1, LOG_TOP, {})`

core/arch/proc_api.h
```diff
@@ -255,6 +265,48 @@ typedef enum {
...
+ *   supported at all but in some cases 15 means a feature is not supported at
+ *   all, NSFLAG.
+ */
+#    define DEF_FEAT(FREG, NIBPOS, FVAL, NSFLAG) \
+        ((ushort)((NSFLAG << 15) | (FREG << 8) | (NIBPOS << 4) | FVAL))
+#    define GET_FEAT_REG(FEATURE) (feature_reg_idx_t)((((ushort)FEATURE) & 0x7F00) >> 8)
+#    define GET_FEAT_NIBPOS(FEATURE) ((((ushort)FEATURE) & 0x00F0) >> 4)
+#    define GET_FEAT_VAL(FEATURE) (((ushort)FEATURE) & 0x000F)
+#    define GET_FEAT_NSFLAG(FEATURE) ((((ushort)FEATURE) & 0x8000) >> 15)
+typedef enum {
```
derekbruening, line 284: Missing doxygen comment for the enum.

core/arch/proc_api.h
```diff
@@ -255,6 +265,48 @@ typedef enum {
...
+ * - There are multiple 64 bit system registers for features storage only, FREG.
+ * - Each register is divided into nibbles representing a feature, NIBPOS.
+ * - The value of a nibble represents a certain level of support for that feature, FVAL.
+ * - The values can range from 0 to 15. In most cases 0 means a feature is not
+ *   supported at all but in some cases 15 means a feature is not supported at
+ *   all, NSFLAG.
+ */
+#    define DEF_FEAT(FREG, NIBPOS, FVAL, NSFLAG) \
+        ((ushort)((NSFLAG << 15) | (FREG << 8) | (NIBPOS << 4) | FVAL))
+#    define GET_FEAT_REG(FEATURE) (feature_reg_idx_t)((((ushort)FEATURE) & 0x7F00) >> 8)
```
derekbruening, line 280: Normally the _api.h files only contain public, doxygen-documented items.  This and the next few should probably be in proc.h or just proc.c instead.

core/arch/arch_exports.h
```diff
@@ -609,6 +609,10 @@ cleanup_and_terminate(dcontext_t *dcontext, int sysnum, ptr_uint_t sys_arg1,
 
 bool
 cpuid_supported(void);
+#ifdef AARCHXX
+bool
+mrs_id_reg_supported(void);
```
derekbruening, line 614: This is only used inside arch/ so no need to be in arch_exports.h: can be in arch.h

core/arch/proc_api.h
```diff
@@ -158,7 +159,16 @@ typedef struct {
     uint ext_flags_ecx;  /**< extended feature flags stored in ecx */
     uint sext_flags_ebx; /**< structured extended feature flags stored in ebx */
 } features_t;
+#elif defined(AARCHXX)
+typedef struct {
```
derekbruening, line 163: Missing doxygen doc for the struct.

core/arch/proc_api.h
```diff
@@ -158,7 +159,16 @@ typedef struct {
...
+    uint64 flags_aa64isar0; /**< feature flags stored in ID_AA64ISAR0_EL1 */
+    uint64 flags_aa64isar1; /**< feature flags stored in ID_AA64ISAR1_EL1 */
+    uint64 flags_aa64pfr0;  /**< feature flags stored in ID_AA64PFR0_EL1 */
+} features_t;
+typedef enum { AA64ISAR0 = 0, AA64ISAR1 = 1, AA64PFR0 = 2 } feature_reg_idx_t;
```
derekbruening, line 168: If this is not exported, best to move to a private header or impl file and not keep it in an _api.h file.  If it is exported (doesn't look like it), it needs doxygen docs for the enum and each member.

core/arch/proc_api.h
```diff
@@ -347,8 +399,9 @@ proc_has_feature(feature_bit_t feature);
 
 DR_API
 /**
- * Returns all 4 32-bit feature values.  Use proc_has_feature to test
- * for specific features.
+ * Returns all 4 32-bit feature values on X86 and architectural feature
+ * registers' values on AArch64. Use proc_has_feature to test for specific
```
derekbruening, line 403: Include () to get a doxygen link.

derekbruening, line 403: I know we don't currently have systematic x86 tests for features (I just see a few uses in the clean call test) but it would be good to add something: could we add some kind of sanity test that asking about features works?

core/arch/aarch64/proc.c
AssadHashmi: Done.

core/arch/aarch64/aarch64.asm
AssadHashmi: Most of the time the usual status registers are being read: NZCV, FPCR and FPSR.
The other two are CTR_EL0 and TPIDR_EL0 both in user-space at EL0.
Before this patch, this was not a problem, but now we are reading registers at EL1.

That comment is too vague. It should make clear that that the Linux kernel enables MRS to read a certain set of specified registers above exception level EL0 using a trap-and-emulate mechanism. Comment updated.

core/arch/aarch64/proc.c
AssadHashmi: Done.

core/arch/proc_api.h
AssadHashmi: Done.

core/arch/arch_exports.h
AssadHashmi: Done.

core/arch/proc_api.h
AssadHashmi: Moved to proc.c.

core/arch/proc_api.h
AssadHashmi: Done.

core/arch/proc_api.h
AssadHashmi: Done.

core/arch/proc_api.h
AssadHashmi: I don't know how to do this without #including a private header in proc_api.h
Any advice?

core/arch/proc_api.h
AssadHashmi: Do you mean client tests?

Currently the LOG_FEATURE macro puts out text in the log file for debug builds which exercises the register access and feature bits interpretation code.

core/arch/proc_api.h
derekbruening: Looking more closely: ok, you mean that this is used directly in the exposed enum values below, as in ` FEATURE_AES = DEF_FEAT(AA64ISAR0, 1, 1, 0), `.  I guess it does need to be here then.  What does the generated doxygen look like?  I guess it has DEF_FEAT and AA64ISAR0 but they aren't links.

derekbruening: One option would be to doxygen-document them so the user can see them in the html.  Or it doesn't seem that bad to have them not documented: can live w/ the negative of not having the html show the components and could possibly change them later.

core/arch/proc_api.h
derekbruening: Right, a client using the API routines seems like a good addition.

core/arch/proc_api.h
AssadHashmi: > What does the generated doxygen look like? I guess it has DEF_FEAT and AA64ISAR0 but they aren't links.

I can't see AArch64 `feature_bit_t` output in the HTML.

I built with
`cmake -DDEBUG=ON -DBUILD_TESTS=ON -DINTERNAL=ON -DBUILD_DOCS=ON`
and the X86 enum `feature_bit_t` list appeared in full but no AArch64.

Have I built it wrong? Or is there some Doxygen/C pre-processor magic which needs doing?

core/arch/proc_api.h
AssadHashmi: Added client-interface/features.dll.c unit test.

core/arch/proc_api.h
derekbruening: I think it's the use of `#elif`.  See `PREDEFINED` in API.doxy where it has all OS and arch defines set: so if avoid else in the _api.h headers we get it all in any build.

I see at least one comment about this:
```
core/ir/instr_api.h:/* We resist using #elif here because otherwise doxygen will be unable to
```

But I see a bunch of other uses of elses: what we need is a lint rule on this. I filed #5496 on this.

suite/tests/client-interface/features.dll.c
```diff
@@ -0,0 +1,99 @@
...
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+/* Test AArch64 ISA features from a client. Reads 'Features' string from
+ * /proc/cpuinfo then checks to see if proc_has_feature() correctly identifies
+ * the FEATURE_<x> as supported.
```
derekbruening, line 35: OK, so this is a good focused test for these features for aarch64.  How about a TODO comment about adding x86 testing as well.

derekbruening (approved):

core/arch/proc_api.h
AssadHashmi: Removing the `#elif` does result in both X86 and AArch64 documentation being generated but it merges the X86 and AArch64 HTML comments into single blocks. It also confuses FEATURE_AES because it's common to both. I've renamed the AArch64 version to FEATURE_AESX.

I've re-worded the Doxygen comments to make more sense when rendered to HTML as well as adding X86 and AArch64 labels wherever needed to make it clearer for users. Not ideal but I think will do for now.

> Or it doesn't seem that bad to have them not documented: can live w/ the negative of not having the html show the components and could possibly change them later.

Yes, I would prefer to wait.

derekbruening:
Turns out this broke our internal build which sets `-Wvla` and in C a `const int` is not a compile-time constant.  I'll fix it and add `-Wvla` to the cmake flags.

Issue #4610, opened 2020-12-14
VS2017 running out of memory compiling ir_x86.c after recent AVX-512 additions


derekbruening:
This happened with VS2013 too in the past: #3992.  We split up some #included files to solve it then, but they have grown again.

The recent AVX-512 broadcast additions for #4534 in PR #4577 seem to have pushed it over the edge: but it seems non-deterministic since several prior builds on GA CI Server2016 VS2017 worked.

https://github.com/DynamoRIO/dynamorio/pull/4586/checks?check_run_id=1553077661
```
2020-12-14T21:09:44.3614062Z FAILED: suite/tests/CMakeFiles/api.ir.dir/api/ir_x86.c.obj 
2020-12-14T21:09:44.3616541Z C:\PROGRA~2\MICROS~1\2017\ENTERP~1\VC\Tools\MSVC\1416~1.270\bin\Hostx64\x64\cl.exe  /nologo -DDR_HOST_ -DDR_REG_ENUM_COMPATIBILITY -DDYNAMORIO_STANDALONE -DX86_64 -I. -I..\suite\tests -I..\core\arch -Icmake\..\include -I..\core\drlibc -I..\core\lib -Iinclude\annotations -Isuite\tests\annotations -Iext\include -DSTANDALONE -DUSE_DYNAMO  /nologo /MP /GF /FS /GS- /MTd  /Zi /Od /W3 /WX -DNIGHTLY_REGRESSION -DNOT_DYNAMORIO_CORE /showIncludes /Fosuite\tests\CMakeFiles\api.ir.dir\api\ir_x86.c.obj /Fdsuite\tests\CMakeFiles\api.ir.dir\ /FS -c ..\suite\tests\api\ir_x86.c
2020-12-14T21:09:44.3618960Z d:\a\dynamorio\dynamorio\suite\tests\api\ir_x86.c(469) : fatal error C1002: compiler is out of heap space in pass 2
2020-12-14T21:09:44.3620066Z d:\a\dynamorio\dynamorio\suite\tests\api\ir_x86.c(458) : fatal error C1002: compiler is out of heap space in pass 2
2020-12-14T21:09:44.8225895Z ninja: build stopped: subcommand failed.
```

Compare these to the sizes listed in #3992:
```
$ wc suite/tests/api/ir*.h | sort -n
    65    453   3684 suite/tests/api/ir_x86_2args_avx512_vex.h
[...]

PR #5497, opened 2022-05-22
i#5383: build on macos m1

heyitsanthony:
Good morning dynamo devs.

This PR adds support for running a "hello world" program (printf and read()) on my M1 macbook. It is by no means perfect but it's enough of a first cut that it can be refined / stabilized for more cases. My apologies for the large-ish first time contributor patch set. 

Most of the changes in this PR are related to:
* macos aarch64 TLS
* untangling macos64 = x86 / aarch64 = linux assumptions
* macho symbols / minor assembler differences
* MacOS's MAP_JIT flag
* v8.3 PAC pointers / arm64e

Rough edges:
* segfaults on heap cleanup when exiting in debug builds (release builds are OK)
* returning from signals not seriously attempted


AssadHashmi:
Added @heyitsanthony to AArch64 Jenkins CI user list.

derekbruening:

Looks like all the redness for test failures is from tabs which are checked before actually running any tests:
```
  ERROR: diff contains tabs:
  +	message(FATAL_ERROR "nasm is too old: no 64-bit support")
```

Not sure if we have a volunteer for a detailed review?

bete0:
@[AssadHashmi](https://github.com/AssadHashmi) Would you like to review this?

AssadHashmi:
> @[AssadHashmi](https://github.com/AssadHashmi) Would you like to review this?

I'll have a look at this but not sure when I'll be able to do it. I'll aim for a review at the start of next week.

bete0:
Thank you @AssadHashmi.

derekbruening:
@AssadHashmi have you started looking at this?  If not I will take a look.

@heyitsanthony for future updates please use regular pushes rather than force-pushes as the forces mess up the review history and make it hard for reviewers to see just what's changed (our docs do request no forces).  The final squash-and-merge will clean everything up into one commit with one description.

core/arch/arch.c
```diff
@@ -3426,15 +3428,15 @@ dr_mcontext_to_priv_mcontext(priv_mcontext_t *dst, dr_mcontext_t *src)
     if (src->size > sizeof(dr_mcontext_t))
         return false;
     if (TESTALL(DR_MC_ALL, src->flags) && src->size == sizeof(dr_mcontext_t)) {
-        *dst = *(priv_mcontext_t *)(&MCXT_FIRST_REG_FIELD(src));
+        *dst = *(priv_mcontext_t *)(MCXT_FIRST_REG_FIELD(src));
```
derekbruening, line 3431: These changes seem unnecessary.

core/arch/arch.h
```diff
@@ -1336,9 +1336,20 @@ new_thread_setup(priv_mcontext_t *mc);
 #    ifdef MACOS
 void
 new_bsdthread_setup(priv_mcontext_t *mc);
+// Enable writing to MAP_JIT pages.
+#        define pthread_jit_write() pthread_jit_write_protect_np(false)
```
derekbruening, line 1340: style: use all caps for macros.  Ditto below.l

core/arch/arch_exports.h
```diff
@@ -1674,8 +1674,23 @@ _dynamorio_runtime_resolve(void);
 #    define APP_PARAM(mc, offs) APP_PARAM_##offs(mc)
 #endif /* X86/ARM */
 
-#define MCXT_SYSNUM_REG(mc) ((mc)->IF_X86_ELSE(xax, IF_ARM_ELSE(r7, r8)))
-#define MCXT_FIRST_REG_FIELD(mc) ((mc)->IF_X86_ELSE(xdi, r0))
+#if !(defined(MACOS) && defined(AARCH64))
+#    define MCXT_SYSNUM_REG(mc) ((mc)->IF_X86_ELSE(xax, IF_ARM_ELSE(r7, r8)))
+#    define MCXT_SET_SYSNUM_REG(mc, sysnr) \
+        ((mc)->IF_X86_ELSE(xax, IF_ARM_ELSE(r7, r8))) = (sysnr)
+#    define MCXT_FIRST_REG_FIELD(mc) &((mc)->IF_X86_ELSE(xdi, r0))
```
derekbruening, line 1681: As noted above it seems unnecessary to change this to take the address; that also seems misleading as the name does not sound like an address.

core/arch/arch_exports.h
```diff
@@ -1674,8 +1674,23 @@ _dynamorio_runtime_resolve(void);
...
+#else
+// syscall() sets r16 to 0 and uses r0 for sysnr; libsystem_c uses r16
```
derekbruening, line 1683: You mean the library function syscall()?  DR goes straight to the kernel gateway if it can and doesn't try to deal with library routines.  I don't understand this though: does the kernel have multiple conventions?  Or is DR intercepting multiple library routines?

core/arch/arch_exports.h
```diff
@@ -1674,8 +1674,23 @@ _dynamorio_runtime_resolve(void);
...
+#    define MCXT_SYSNUM_REG(mc) ((mc)->r16 ? (mc)->r16 : (mc)->r0)
```
derekbruening, line 1684: style: We also use explicit bools so `r16 != 0`; ditto below

core/arch/asm_defines.asm
```diff
@@ -99,23 +99,71 @@
 # define START_DATA .data
 # define START_FILE .text
 # define END_FILE /* nothing */
+
+#if defined(MACOS) && defined(AARCH64)
```
derekbruening, line 103: Unfortunately we don't have clang-format on this file (not sure why not?  It's all cpp despite the name .asm; should try to add) which would auto-indent for us.  This should be indented and the ones below as well.

core/arch/atomic_exports.h
```diff
@@ -389,13 +389,14 @@ atomic_add_exchange_int64(volatile int64 *var, int64 value)
...
-            do {                                                     \
-                /* We use "load-acquire" to add a barrier. */        \
-                __asm__ __volatile__("ldarb w0, [%0]; strb w0, [%1]" \
-                                     :                               \
-                                     : "r"(addr_src), "r"(addr_res)  \
-                                     : "w0", "memory");              \
+#        define ATOMIC_1BYTE_READ(addr_src, addr_res)               \
+            do {                                                    \
+                /* We use "load-acquire" to add a barrier. */       \
+                __asm__ __volatile__("ldarb w0, [%0]\n\t"           \
```
derekbruening, line 395: The assembler on mac won't accept a ; between instructions?

core/arch/arch.h
```diff
@@ -1336,9 +1336,20 @@ new_thread_setup(priv_mcontext_t *mc);
...
+// Enable executing MAP_JIT pages.
```
derekbruening, line 1341: Explain that this is per-thread as the general intuition is that it would be process-wide and DR would have to serialize all the threads.

core/unix/os.c
```diff
@@ -3123,6 +3153,9 @@ os_raw_mem_alloc(void *preferred, size_t size, uint prot, uint flags,
     uint os_prot = memprot_to_osprot(prot);
     uint os_flags =
         MAP_PRIVATE | MAP_ANONYMOUS | (TEST(RAW_ALLOC_32BIT, flags) ? MAP_32BIT : 0);
+#if defined(MACOS) && defined(AARCH64)
+    os_flags |= MAP_JIT;
```
derekbruening, line 3157: This should only be for +x memory, shouldn't it?

core/unix/os.c
```diff
@@ -3314,15 +3347,16 @@ os_heap_reserve(void *preferred, size_t size, heap_error_code_t *error_code,
     /* should only be used on aligned pieces */
     ASSERT(size > 0 && ALIGNED(size, PAGE_SIZE));
     ASSERT(error_code != NULL);
+    uint os_flags = MAP_PRIVATE |
+        MAP_ANONYMOUS IF_X64(| (DYNAMO_OPTION(heap_in_lower_4GB) ? MAP_32BIT : 0));
+#if defined(MACOS) && defined(AARCH64)
+    os_flags |= MAP_JIT;
```
derekbruening, line 3353: This should only be for +x memory, shouldn't it?

core/arch/arch.c
```diff
@@ -497,6 +497,8 @@ static void shared_gencode_init(IF_X86_64_ELSE(gencode_mode_t gencode_mode, void
     bool x86_to_x64_mode = false;
 #endif
 
+    pthread_jit_write();
```
derekbruening, line 500: Not sure these calls are everywhere they should be, nor paired with the counterpart at all the right places.  I don't see a write call for building a new block?  Is that the vmcode_get_writable_addr()?  But that's called many times...probably better to call the write just once at a higher level?  How about some XXX i#5383 comments for each of these calls in this PR explaining that they may not be ideally placed or paired or exhaustive and an audit should be done to improve.

core/dispatch.c
```diff
@@ -562,6 +562,7 @@ enter_fcache(dcontext_t *dcontext, fcache_enter_func_t entry, cache_pc pc)
 #endif
 
     dcontext->whereami = DR_WHERE_FCACHE;
+    pthread_jit_read();
```
derekbruening, line 565: See the general comment above on these.  Better to have write calls around block building and linking paths rather than assuming all paths might have written, with a debug query here to make sure no paths were missed?  Is this call expensive?

derekbruening (commented):
Made it through the first 15 files to dispatch.c; sending out comments so far to get the ball rolling.

core/ir/aarch64/instr_create_api.h
```diff
@@ -3376,4 +3376,11 @@
  */
 #define INSTR_CREATE_fmov_scalar_imm(dc, Rd, f) instr_create_1dst_1src(dc, OP_fmov, Rd, f)
 
+/**
+ * Creates a XPACI instruction.
+ * \param dc      The void * dcontext used to allocate memory for the instr_t.
+ * \param Rd      Register with PAC bits to remove.
+ */
+#define INSTR_CREATE_xpaci(dc, Rd) instr_create_0dst_1src((dc), OP_xpaci, (Rd))
```
derekbruening, line 3384: Where are all the other new opcodes added?  Though xref #2440 and long-term these should be auto-generated.

core/options.c
```diff
@@ -1127,7 +1127,7 @@ options_enable_code_api_dependences(options_t *options)
      * logic in heap_mmap_reserve_post_stack() to handle sharing the
      * tail end of a multi-64K-region stack.
      */
-    options->stack_size = MAX(options->stack_size, 56 * 1024);
+    options->stack_size = MAX(options->stack_size, IF_MACOSA64_ELSE(48, 56) * 1024);
```
derekbruening, line 1130: What is the reason for this?  A comment would be good, to explain.

core/options.c
```diff
@@ -1393,7 +1402,7 @@ check_option_compatibility_helper(int recurse_count)
         dynamo_options.shared_traces = false;
         changed_options = true;
     }
-#            ifdef X64
+#            if defined(X64) && !(defined(MACOS) && defined(AARCH64))
```
derekbruening, line 1405: But isn't HAVE_TLS defined for Mac?

core/optionsx.h
```diff
@@ -833,7 +833,8 @@ OPTION_DEFAULT(uint_size, stack_size,
                 * 32KB is the max that will still allow sharing per-thread
                 * gencode in the same 64KB alloc as the stack on Windows.
                 */
-               24 * 1024, "size of thread-private stacks, in KB")
+               IF_MACOSA64_ELSE(32 * 1024, 24 * 1024),
```
derekbruening, line 836: Another one that needs a comment

core/optionsx.h
```diff
@@ -1158,9 +1159,11 @@ OPTION_DEFAULT(uint, max_trace_bbs, 128, "maximum number of basic blocks in a tr
 /* FIXME i#3522: re-enable SELFPROT_DATA_RARE on linux */
 OPTION_DEFAULT(
     uint, protect_mask,
-    IF_STATIC_LIBRARY_ELSE(0x100 /*SELFPROT_GENCODE*/,
-                           IF_WINDOWS_ELSE(0x101 /*SELFPROT_DATA_RARE|SELFPROT_GENCODE*/,
-                                           0x100 /*SELFPROT_GENCODE*/)),
+    IF_STATIC_LIBRARY_ELSE(
+        0x100 /*SELFPROT_GENCODE*/,
+        IF_MACOSA64_ELSE(0,
```
derekbruening, line 1164: There's really no support for changing +rwx memory to +rx?  I thought that's what the jit_write_* calls you added did?

core/unix/loader.c
```diff
@@ -2195,4 +2195,15 @@ privload_early_inject(void **sp, byte *old_libdr_base, size_t old_libdr_size)
 #    endif /* !defined(STANDALONE_UNIT_TEST) && !defined(STATIC_LIBRARY) */
 #else
 /* XXX i#1285: implement MacOS private loader */
+void
+relocate_dynamorio(byte *dr_map, size_t dr_size, byte *sp)
+{
+    abort();
```
derekbruening, line 2201: Please include TODO comments with issue numbers for anything not implemented, and use ASSERT_NOT_IMPLEMENTED (maybe followed by os_terminate if really worried about release build), not libc routines.  Ditto below.

core/unix/os.c
```diff
@@ -132,9 +132,15 @@ typedef struct rlimit64 rlimit64_t;
  */
 #define MCXT_SYSCALL_RES(mc) ((mc)->IF_X86_ELSE(xax, r0))
 #if defined(DR_HOST_AARCH64)
-#    define READ_TP_TO_R3_DISP_IN_R2    \
-        "mrs " ASM_R3 ", tpidr_el0\n\t" \
-        "ldr " ASM_R3 ", [" ASM_R3 ", " ASM_R2 "] \n\t"
+#    if defined(MACOS)
+#        define READ_TP_TO_R3_DISP_IN_R2      \
+            "mrs " ASM_R3 ", tpidrro_el0\n\t" \
```
derekbruening, line 137: The only difference I see is `tpidrro_el0` instead of `tpidr_el0`.  It seems like that should be a change made in the  TLS headers with the usage here parameterized.

core/unix/os.c
```diff
@@ -1291,12 +1297,21 @@ query_time_seconds(void)
...
+
 /* milliseconds since 1601 */
 uint64
 query_time_millis()
 {
     struct timeval current_time;
+#if !(defined(MACOS) && defined(AARCH64))
     uint64 val = dynamorio_syscall(SYS_gettimeofday, 2, &current_time, NULL);
+#else
+    uint64 val = gettimeofday(&current_time, NULL);
```
derekbruening, line 1312: As above, library routines cause problems.  Please include a XXX i#NNN comment about this being temporary.

core/unix/os.c
```diff
@@ -1312,7 +1327,7 @@ query_time_millis()
         res += UTC_TO_EPOCH_SECONDS * 1000;
         return res;
     } else {
-        ASSERT_NOT_REACHED();
+        // ASSERT_NOT_REACHED();
```
derekbruening, line 1330: style violation: no commented-out code

core/ir/aarch64/codec.c
```diff
@@ -1648,6 +1648,30 @@ encode_opnd_op2(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_out)
     return encode_opnd_int(5, 3, false, 0, 0, opnd, enc_out);
 }
 
+/* x5p8 : X register, add 8 */
```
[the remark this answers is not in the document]
AssadHashmi: This looks like it decodes the `Xn` register not to `n` but to `n+8`.
It's used by `blraaz` and `autdb` in `codec_v83.txt` below.
But looking at those instrs in the manual, there appears to be no restriction to the `n` in `Xn`.
They should use the existing `decode_opnd_x5()` and `encode_opnd_x5()` AFAICT.

core/ir/aarch64/codec_v83.txt
```diff
@@ -0,0 +1,53 @@
...
+# Instruction definitions:
+
+# retaa, retab
+11010110010111110000101111111111 n 579 reta :
+11010110010111110000111111111111 n 579 reta :
+11010101000000110010001101111111 n 580 pacibsp :
+11010101000000110010001111111111 n 581 autibsp :
+1101011000111111000010xxxxx11111 n 582 blraaz impx30 : x5
+# blraa
+110101110011111100001001xxxxxxxx n 582 blraaz impx30 : x5p8 x0
```
AssadHashmi, line 44: Which variant of the `blra*` instructions is this? I can't find one with 3 operands.

core/ir/aarch64/codec_v83.txt
```diff
@@ -0,0 +1,53 @@
...
+1101011100011111000010xxxxxxxxxx n 583 braa  : x5 x0
+110110101100000100100011111xxxxx n 584 paciza : x0
+110110101100000101000011111xxxxx n 585 xpaci : x0
+110110101100000101000111111xxxxx n 586 xpacd : x0
+1101101011000001000010xxxxxxxxxx n 587 pacda : x0 x5
+1101101011000001000000xxxxxxxxxx n 588 pacia : x0 x5
+1101101011000001000001xxxxxxxxxx n 589 pacib : x0 x5
+110110101100000100011101xxxxxxxx n 590 autdb : x5p8 x0
```
AssadHashmi, line 52: I think `x5p8` should be `x5`, see comment in `codec.c`.

core/ir/aarch64/instr_create_api.h
AssadHashmi: `codec.py` will autogenerate the new opcodes from the instr definitions in `codec_v83.txt` and add to the `opcodes_*.h` files, placed in the build directory.

core/unix/os_public.h
```diff
@@ -134,41 +140,48 @@ typedef kernel_sigcontext_t sigcontext_t;
...
-#    define SC_R25 SC_FIELD(regs[25])
-#    define SC_R26 SC_FIELD(regs[26])
-#    define SC_R27 SC_FIELD(regs[27])
-#    define SC_R28 SC_FIELD(regs[28])
-#    define SC_FP SC_FIELD(regs[29])
-#    define SC_LR SC_FIELD(regs[30])
 #    define SC_XSP SC_FIELD(sp)
-#    define SC_XFLAGS SC_FIELD(pstate)
-#    define SC_SYSNUM_REG SC_FIELD(regs[8])
+#    define SC_SYSNUM_REG SC_FIELD_AARCH64(8)
```
derekbruening, line 184: This should be 16 for mac.  Fixed in #5610.

ext/CMakeLists.txt
```diff
@@ -112,7 +112,9 @@ add_subdirectory(drx)
 add_subdirectory(drwrap)
 add_subdirectory(drreg)
 add_subdirectory(drbbdup)
-add_subdirectory(drsyms)
+if (NOT (MACOS AND AARCH64))
+  add_subdirectory(drsyms)
```
derekbruening, line 116: Should have a TODO comment here: added.

ext/drwrap/drwrap_asm_aarch64.asm
```diff
@@ -39,9 +39,8 @@
 
 START_FILE
 
-#define FUNCNAME replace_native_xfer
-        DECLARE_FUNC(FUNCNAME)
-GLOBAL_LABEL(FUNCNAME:)
+        DECLARE_FUNC(replace_native_xfer)
```
derekbruening, line 42: What is the reason for the changes in this file?

core/unix/signal.c
```diff
@@ -4174,30 +4189,34 @@ abort_on_fault(dcontext_t *dcontext, uint dumpcore_flag, app_pc pc, byte *target
         dumpcore_flag = 0;
 #endif
 
+#if !(defined(MACOS) && defined(AARCH64))
```
derekbruening, line 4192: Re-enabling and removing the abort() w/o an explanation for this change.

core/unix/signal.c
```diff
@@ -4567,6 +4586,9 @@ find_next_fragment_from_gencode(dcontext_t *dcontext, sigcontext_t *sc)
             instr_free(dcontext, &instr);
         }
 #ifdef AARCHXX
+#    if defined(MACOS)
+        abort();
```
derekbruening, line 4590: Changed to ASSERT_NOT_IMPLEMENTED.

core/options.c
derekbruening: Reverting, since there's no explanation.

core/optionsx.h
derekbruening: Added an XXX comment at least.

core/optionsx.h
derekbruening: Added an XXX comment.

derekbruening (commented):
I went through everything and applied changes to PR #4610 in the interest of making progress on M1.

derekbruening:
I merged this as PR #5610 with local fixes to all my review comments.  It does not build on my M1, and the tests don't configure, but hopefully it's just a few warnings in the core code.

Issue #5499, opened 2022-05-24
ASSERT !bitmap_test in -satisfy_w_xor_x on an app that forks


derekbruening:
For #5495 I'm adding a fork test to drcachesim -offline and it fails under -satisfy_w_xor_x for two reasons: first, there's an assert on an unclosed file, which I'm fixing in my #5495 PR.  The other I did not have time to figure out so I'm filing here and disabling the test for -satisfy_w_xor_x.  There is something wrong in the across-fork remappings wrt the vmm metadata that only shows up with a tool using more heap than plain DR:

```
$ rm -rf drmemtrace.*.dir; bin64/drrun -loglevel 2 -satisfy_w_xor_x -t drcachesim -offline -- suite/tests/bin/linux.fork
<log dir=/home/bruening/dr/git/build_x64_dbg_tests/bin64/../logs/linux.fork.66462.00000000>
<Starting application /home/bruening/dr/git/build_x64_dbg_tests/suite/tests/bin/linux.fork (66462)>
<Initial options ...>
<Paste into GDB...>
<curiosity: rex.w on OPSZ_6_irex10_short4!>
<spurious rep/repne prefix @0x00007f7c242ed280 (f3 0f 1e fa): >
parent is running under DynamoRIO
<-- parent 66462 forked child 66465 -->
<log dir=/home/bruening/dr/git/build_x64_dbg_tests/bin64/../logs/linux.fork.66465.00000000>
parent waiting for child
child is running under DynamoRIO
<Stopping application /home/bruening/dr/git/build_x64_dbg_tests/suite/tests/bin/linux.fork (66465)>
<Application /home/bruening/dr/git/build_x64_dbg_tests/suite/tests/bin/linux.fork (66465).  Internal Error: DynamoRIO debug check failure: /home/bruening/dr/git/src/core/heap.c:1113 bitmap_are_reserved_blocks(vmh->blocks, vmh->num_blocks, vmm_addr_to_block(vmh, p), (uint)(size / DYNAMO_OPTION(vmm_block_size)))
(Error occurred @3693 frags in tid 66465)
```

```
#0  report_dynamorio_problem (dcontext=0x0, dumpcore_flag=8, exception_addr=0x0, report_ebp=0x0, 
    fmt=0x7ffff7ec3018 "DynamoRIO debug check failure: %s:%d %s\n(Error occurred @%d frags in tid %d)") at /home/bruening/dr/git/src/core/utils.c:2114
[...]

Issue #4660, opened 2021-01-07
Avoid standard heap in droption and drcachesim init


derekbruening:
While typical shared-library usage of DR isolates a client's heap from the app, when DR and clients are statically linked into the application there is unfortunately no way to isolate.  We're already using placement new at runtime in drcachesim to support static usage.  We'd been assuming that application heap usage at client init time was relatively safe, for internally-triggered attach where the init thread is at a clean ABI point.  However, we have hit issues with application heap and tools that replace the allocator with a scheme involving SIGSEGV for detecting bugs.  During attach, the init thread then raises SIGSEGV, running into signal delivery problems under #1921.  While we plan to solve all the #1921 cases in any case, it may be worth providing app-heap-free options, for possible ptrace-attach with static DR or other cases where worse problems arise, or cases where clients need to parse options post-init-time.

Xref https://github.com/DynamoRIO/dynamorio/issues/4640#issuecomment-755728478 where using std::string with a custom allocator that invokes dr_global_alloc for all droption strings does not work with static DR, due to the static initializers that run natively at process init time.  An allocator that looks at dynamo_heap_initialized does not work either, as it doesn't know how to free.

A possible solution is to store only literal char* types at init time and only construct std::string at parse time.

Beyond droption, func_trace has a std::vector and std::set which can easily be made to use a custom allocator.

We may want to make a C++ allocator that uses dr_global_heap part of a library exported by DR.  Or bundled with droption.

PR #5500, opened 2022-05-24
i#5495: Fix drcachesim fork bug

### LINKED ISSUE #5495: drcachesim offline fails on fork child with error about file creation
I first hit this tracing SPEC2017 perlbench.
I can reproduce it easily:
```
$ bin64/drrun -t drcachesim -offline -- suite/tests/bin/linux.fork
parent is running under DynamoRIO
<-- parent 564285 forked child 564286 -->
Failed to create new thread file for window ./drmemtrace.linux.fork.564286.0539.dir/raw/drmemtrace.linux.fork.564286.7156.raw.lz4
```

derekbruening:
Fixes a bug where drcachesim -offline complained about an open file
across a fork.  This bug blocked tracing of SPECCPU perlbench.

Adds a new test of drcachesim -offline with an app that forks.  As is,
this matched the -satisfy_w_xor_x test regex, and I tried to make it
work with that option by fixing a file close bug here.  However, I did
not have time to figure out another bug which I filed as #5499.  I
thus tightened the "fork" regex to exclude this test from
-satisfy_w_xor_x.

Fixes #5495

### DIFF
--- a/clients/drcachesim/tests/offline-fork.templatex
+++ b/clients/drcachesim/tests/offline-fork.templatex
@@ -1,0 +1,26 @@
+Hello, world!
+Cache simulation results:
+Core #0 \(1 thread\(s\)\)
+  L1I stats:
+    Hits:                         *[0-9,\.]*
+    Misses:                       *[0-9,\.]*
+    Compulsory misses:            *[0-9,\.]*
+    Invalidations:                *0
+.*    Miss rate:                        [0-3][,\.]..%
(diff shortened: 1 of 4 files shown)


derekbruening:
Grrr, the 32-bit test hits heap use from droption's C++ classes which is #4660 and doesn't have a simple solution.

abhinav92003 (approved):

Issue #4014, opened 2020-01-13
Add support for -use_physical with drcachesim -offline


derekbruening:
-use_physical is only supported for online drcachesim today, but that is not made clear in the docs nor in actual usage.  The tool lets you run `-offline -use_physical` and ends up with a post-processed trace with physical data addresses but virtual PC fetches, due to an accident.  `instr_offline_t::get_entry_addr` blindly treats the entry as a data ref, so it tries to
translate bogus addresses composed of module id + offset encodings.  
And indeed we get warnings at -verbose 1+:
```
virtual2physical translation failure for <10,  0, 0x0004000c00000f30>
virtual 140727803080432 => physical 3824
virtual2physical translation failure for <10,  0, 0x002c000c00001d30>
virtual2physical translation failure for <10,  0, 0x0016000c00001d85>
virtual2physical translation failure for <10,  0, 0x0004000c00001de3>
virtual2physical translation failure for <10,  0, 0x000a000c00001dd3>
virtual 140346642070176 => physical 2720
virtual 140346642067080 => physical 3720
virtual2physical translation failure for <10,  0, 0x0004000c00001de3>
...
```
Those are the instruction entries:
```
$ od -t x8 -A x `ls -1td drmemtrace.threadsig.*.dir | head -1`/raw/*.raw | head
000000 c000000000000003 4000000000030385
000010 6000000000030385 802efaa01299a30c
000020 c203000000000003 2004000c00000f30
000030 0000000000000ef0 202c000c00001d30
000040 0000000000000ee8 0000000000000ee0
000050 0000000000000ed8 0000000000000ed0
000060 0000000000000ec8 0000000000000ec0
000070 2016000c00001d85 2004000c00001de3
000080 200a000c00001dd3 0000000000000aa0
000090 0000000000000e88 2004000c00001de3
```

So we have two action items.  First, we should have the front-end refuse to combine `-use_physical` and `-offline` for now, and update the docs to reflect this.

Next, we need to decide whether to try to support this combination.  Is it possible that future Linux distros will all shut down access to pagemap?  Although for research purposes running as sudo or something may still be feasible.

How would we support this?  We'd either have to store extra info for each data ref plus
an entry for each instr like for DGC (#2062), or have a custom solution just for offline
where each block PC has 2 entries: one virtual and one physical.  The former will not work
well w/ tools that want operands (like opcode_mix or micro-arch simulators): but that would be the same for DGC, so we might need an option to store the full instr bytes.

[...]

PR #5506, opened 2022-05-27
i#4014 dr$sim phys: Avoid malloc in physaddr_t

derekbruening:
Switches from std::unordered_map in physaddr_t to a drcontainers
hashtable to avoid malloc and make things safe for statically-linked
drmemtrace.

Similarly, switches from std::ostringstream to dr_snprintf in
physaddr_t::init() to avoid malloc.

Tested on a multi-threaded app which hits the post-init malloc warning
without both fixes (test will be added in a forthcoming PR: it cannot
be added now as physaddr_t is not thread-safe yet).

Issue: #4014

### DIFF
--- a/clients/drcachesim/tracer/physaddr.cpp
+++ b/clients/drcachesim/tracer/physaddr.cpp
@@ -69,13 +69,17 @@ physaddr_t::physaddr_t()
     , count_(0)
 #endif
 {
-    // Nothing else.
-    // No destructor needed either: ifstream closes when destroyed.
+#ifdef LINUX
+    v2p_.table = nullptr;
+#endif
 }
 
 physaddr_t::~physaddr_t()
 {
-    hashtable_delete(&v2p_);
+#ifdef LINUX
+    if (v2p_.table != nullptr)
+        hashtable_delete(&v2p_);
+#endif
 }
 
 bool
--- a/clients/drcachesim/tracer/physaddr.cpp
+++ b/clients/drcachesim/tracer/physaddr.cpp
@@ -40,6 +40,7 @@
 #endif
 #include "physaddr.h"
 #include "../common/options.h"
+#include "../common/utils.h"
 
 #if defined(X86_32) || defined(ARM_32)
 #    define IF_X64_ELSE(x, y) y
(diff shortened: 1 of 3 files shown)


derekbruening:
The Jenkins failure is almost certainly unrelated.  Filed as #5507.

derekbruening:
run arm tests

clients/drcachesim/tracer/physaddr.cpp
```diff
@@ -69,22 +69,37 @@ physaddr_t::physaddr_t()
...
+    // Some threads may not do much, so start out small.
+    constexpr int V2P_INITIAL_BITS = 9;
+    hashtable_init_ex(&v2p_, V2P_INITIAL_BITS, HASH_INTPTR, /*strdup=*/false,
+                      /*synch=*/false, nullptr, nullptr, nullptr);
+
+    // We avoid std::ostringstream to avoid malloc use for static linking.
+    constexpr int MAX_PAGEMAP_FNAME = 64;
+    char fname[MAX_PAGEMAP_FNAME];
+    dr_snprintf(fname, sizeof(fname), "/proc/%d/pagemap", getpid());
+    fname[MAX_PAGEMAP_FNAME - 1] = '\0';
```
abhinav92003, line 98: Is this correct? Don't we want to null-terminate after the "/proc/%d/pagemap",  instead of at `MAX_PAGEMAP_FNAME-1`?

clients/drcachesim/tracer/physaddr.cpp
```diff
@@ -142,7 +161,9 @@ physaddr_t::virtual2physical(addr_t virt)
         std::cerr << "virtual " << virt << " => physical "
                   << (last_ppage_ + PAGE_OFFS(virt)) << std::endl;
     }
-    v2p_[vpage] = last_ppage_;
+    // Store 0 as 1 since 0 means no entry.
+    hashtable_add(&v2p_, reinterpret_cast<void *>(vpage),
+                  reinterpret_cast<void *>(last_ppage_ == 0 ? 1 : last_ppage_));
```
abhinav92003, line 166: Maybe #define a named constant for 1.

abhinav92003 (approved):

clients/drcachesim/tracer/physaddr.cpp
derekbruening: This is standard defensive programming with these string routines that do not always null-terminate (if the buffer fills up): in core/ this is NULL_TERMINATE_BUFFER.

derekbruening: We have a style rule to always null-terminate a buffer after an snprintf or related operation.

clients/drcachesim/tracer/physaddr.cpp
derekbruening: Done.

clients/drcachesim/tracer/physaddr.cpp
derekbruening: Actually I see drcachesim's utils.h has NULL_TERMINATE_BUFFER.  Changed to use it here.

PR #5508, opened 2022-05-27
i#4014 dr$sim phys: Make physaddr_t per-thread

derekbruening:
The physaddr_t class is not thread-safe and was previously used racily
in the drmemtrace code.  We fix that by creating a separate instance
per thread.  A test with multiple threads is added.

Issue: #4014

### DIFF
--- a/clients/drcachesim/tracer/physaddr.h
+++ b/clients/drcachesim/tracer/physaddr.h
@@ -57,8 +57,12 @@ private:
 #ifdef LINUX
     addr_t last_vpage_;
     addr_t last_ppage_;
-    // XXX: An app with thousands of threads might hit open file limits.
-    // Sharing the descriptor would require locks, however.
+    // TODO i#4014: An app with thousands of threads might hit open file limits,
+    // and even a hundred threads will use up DR's private FD limit and push
+    // other files into potential app conflicts.
+    // Sharing the descriptor would require locks, however.  Evaluating
+    // how best to do that (maybe the caching will reduce the contention enough)
+    // is future work.
     int fd_;
     // We would use std::unordered_map, but that is not compatible with
     // statically linking drmemtrace into an app.
--- a/clients/drcachesim/tracer/physaddr.h
+++ b/clients/drcachesim/tracer/physaddr.h
@@ -42,6 +42,8 @@
 #include "hashtable.h"
 #include "../common/trace_entry.h"
 
+// This class is not thread-safe: the caller should create a separate instance
+// per thread.
 class physaddr_t {
 public:
     physaddr_t();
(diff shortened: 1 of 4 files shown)


clients/drcachesim/tracer/tracer.cpp
```diff
@@ -2662,7 +2671,7 @@ event_thread_init(void *drcontext)
 {
     per_thread_t *data = (per_thread_t *)dr_thread_alloc(drcontext, sizeof(per_thread_t));
     DR_ASSERT(data != NULL);
-    memset(data, 0, sizeof(*data));
+    *data = {};
```
abhinav92003, line 2674: Why was this change needed?

clients/drcachesim/tracer/physaddr.h
```diff
@@ -52,10 +54,11 @@ class physaddr_t {
     virtual2physical(addr_t virt);
 
 private:
-    // Assumed to be single-threaded
 #ifdef LINUX
     addr_t last_vpage_;
     addr_t last_ppage_;
+    // XXX: An app with thousands of threads might hit open file limits.
```
abhinav92003, line 60: Does this use one of DR's private FDs? I don't think so, as the implementation doesn't use `dr_file_open`. It probably should right? But then I'd be concerned about exhausting the small number of available private FDs (96). A lock would cause too much contention I suppose, so that's not the solution...

clients/drcachesim/tracer/tracer.cpp
derekbruening: The compiler warns about zeroing a no-longer-POD struct: this is how you're supposed to zero w/o zeroing the class (and potentially clobbering vtable).

clients/drcachesim/tracer/physaddr.h
derekbruening: It does use a private FD: "open" is redirected to "os_open_protected".  So a many-threaded app will indeed use that up.  Hmm.

abhinav92003: > "open" is redirected to "os_open_protected"

I thought only the app's `open`s are redirected.

derekbruening: "Redirected" means pointing to a different function, done by the private loader to client and private libraries: https://github.com/DynamoRIO/dynamorio/blob/master/core/unix/loader.c#L1438

For the app using SYS_open: I don't think DR does anything there.  Looking: just some debug stuff and some emulation stuff.

abhinav92003: > For the app using SYS_open

Oh right, got it.

> "Redirected" means pointing to a different function, done by the private loader to client and private libraries:

Okay.

derekbruening: I think for now I will augment the comment to talk about using up the private FDs but not try to solve it yet; the goal here is to get simple apps/benchmarks working.  We'll list scaling up the thousands of threads as a separate issue to solve.

derekbruening: Expanded the comment pointing to #4014.

clients/drcachesim/tracer/tracer.cpp
derekbruening: Added a comment.

abhinav92003 (approved):

Issue #5498, opened 2022-05-24
ARM: `drutil_insert_get_mem_addr` no longer works correctly on AArch64


petrochenkov:
https://github.com/DynamoRIO/dynamorio/pull/5283 introduced a change to decoding of load instructions, as a result of that change load index registers can be decoded as `w` registers (as opposed to `x` registers as previously).

Not all places in DR are prepared to encounter a `w` register in a load index context.
For example, this logic in `drutil_insert_get_mem_addr_arm`
```c
        reg_id_t index = opnd_get_index(memref);

        // ...

        reg_id_t stolen = dr_get_stolen_reg();

        // ...

        } else if (index == stolen) {
            index = replace_stolen_reg(drcontext, bb, where, memref, dst, scratch,
                                       scratch_used);
        }
```
now doesn't recognize `w` `index` registers as stolen, because `dr_get_stolen_reg()` typically returns an `x` register.

This results in instrumentation being silently incorrect and `drutil_insert_get_mem_addr_arm` producing incorrect addresses.

Some other places besides `drutil_insert_get_mem_addr_arm` may also be affected by this change.

derekbruening:
This may have resulted in incorrect addresses in AArch64 traces gathered since Jan 18.
Is one way to check whether any instruction matches `ldr.*w28` or `str.*w28`?

derekbruening:
This is rather disturbing.  What can we do to avoid such regressions in the future?

+ Add more comprehensive regression tests for drutil_insert_get_mem_addr covering stolen register corner cases

+ Add checks for unmapped addresses to the invariant_checker trace analyzer: but this requires having metadata on the valid mapped regions, which are not gathered by the Github drmemtrace code today.   This is no guarantee as the computed incorrect address could happen to land in a mapped region.

derekbruening:
> Is one way to check whether any instruction matches `ldr.*w28` or `str.*w28`?

The `w28` must be an index, so it would be `ldr.*, w28` or `str.*, w28` for arm-style disassembly -- right?  Planning to search through objdump output as an initial check.

AssadHashmi:
> This is rather disturbing. What can we do to avoid such regressions in the future?

It is! Adding more comprehensive regression tests for `drutil_insert_get_mem_addr()` covering stolen register corner cases is definitely something we should do.

[...]

PR #5511, opened 2022-05-31
i#5498 Fix and test for indexed memory instrs using W28

AssadHashmi:
i#5498 Fix and test for indexed memory instrs using W28

- Handle W28 in drutil_insert_get_mem_addr_arm() if it's used as index
  register.
- Add client test case using W28 index register in memory
  instructions.

Issue: #5498

### DIFF
--- a/clients/drcachesim/common/utils.h
+++ b/clients/drcachesim/common/utils.h
@@ -62,6 +62,13 @@
 
 #define ALIGN_FORWARD(x, alignment) \
     ((((ptr_uint_t)x) + ((alignment)-1)) & (~((ptr_uint_t)(alignment)-1)))
+#define ALIGN_BACKWARD(x, alignment) (((ptr_uint_t)x) & (~((ptr_uint_t)(alignment)-1)))
+
+#define NOTIFY(level, ...)                     \
+    do {                                       \
+        if (op_verbose.get_value() >= (level)) \
+            dr_fprintf(STDERR, __VA_ARGS__);   \
+    } while (0)
 
 #define BOOLS_MATCH(b1, b2) (!!(b1) == !!(b2))
 
--- a/clients/drcachesim/launcher.cpp
+++ b/clients/drcachesim/launcher.cpp
@@ -1,5 +1,5 @@
 /* **********************************************************
- * Copyright (c) 2015-2020 Google, Inc.  All rights reserved.
+ * Copyright (c) 2015-2022 Google, Inc.  All rights reserved.
  * **********************************************************/
 
 /*
@@ -67,6 +67,7 @@
         exit(1);                                            \
     } while (0)
 
+#undef NOTIFY
 #define NOTIFY(level, prefix, msg, ...)                              \
     do {                                                             \
         if (op_verbose.get_value() >= level) {                       \
--- a/clients/drcachesim/tracer/physaddr.cpp
+++ b/clients/drcachesim/tracer/physaddr.cpp
@@ -30,8 +30,6 @@
  * DAMAGE.
  */
 
-#include <iostream>
-#include <sstream>
 #ifdef LINUX
 #    include <sys/types.h>
 #    include <unistd.h>
@@ -56,15 +54,13 @@
 #    define PAGEMAP_VALID 0x8000000000000000
 #    define PAGEMAP_SWAP 0x4000000000000000
 #    define PAGEMAP_PFN 0x007fffffffffffff
-#    define PAGE_BITS 12 // XXX i#1703: handle large pages
-#    define PAGE_START(addr) ((addr) & (~((1 << PAGE_BITS) - 1)))
-#    define PAGE_OFFS(addr) ((addr) & ((1 << PAGE_BITS) - 1))
 static const addr_t PAGE_INVALID = (addr_t)-1;
 #endif
 
 physaddr_t::physaddr_t()
 #ifdef LINUX
-    : last_vpage_(PAGE_INVALID)
+    : page_size_(dr_page_size())
+    , last_vpage_(PAGE_INVALID)
     , last_ppage_(PAGE_INVALID)
     , fd_(-1)
     , count_(0)
@@ -72,6 +68,14 @@ physaddr_t::physaddr_t()
 {
 #ifdef LINUX
     v2p_.table = nullptr;
+
+    page_bits_ = 0;
+    size_t temp = page_size_;
+    while (temp > 1) {
+        ++page_bits_;
+        temp >>= 1;
+    }
+    NOTIFY(1, "Page size: %zu; bits: %d\n", page_size_, page_bits_);
 #endif
 }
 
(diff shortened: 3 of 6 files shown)


suite/tests/client-interface/stolen-reg-index.appdll.c
```diff
@@ -0,0 +1,96 @@
...
+{
+    run_tests();
+    return 0;
+}
+
+#else /* asm code *************************************************************/
+#    include "asm_defines.asm"
+/* clang-format off */
+START_FILE
+#define FUNCNAME indexed_mem_test
```
derekbruening, line 73: I was going to say that with just one asm file and without needing Windows support, this could be an inline asm function added to drutil-test.c to leverage whatever address verification is there too, rather than a separate file.  But: it looks like today's drutil-test.c is drmgr-test.c.  Huh.  That's only to get the rep str asm code I assume: it doesn't need the other stuff.  Maybe a new drutil-test.c should be made that has the rep str for x86 and stolen reg for arm.

ext/drutil/drutil.c
```diff
@@ -422,6 +422,11 @@ drutil_insert_get_mem_addr_arm(void *drcontext, instrlist_t *bb, instr_t *where,
             disp = -disp;
             negated = !negated;
         }
+#    ifdef AARCH64
+        /* FIXME i#5498 */
+        if (index == DR_REG_W28)
+            return false;
```
derekbruening, line 428: Mixed feelings here:

- On the one hand, this seems the right thing to do, putting this in ASAP rather than having silently incorrect instrumentation.
- On the other hand, drmemtrace turns this false return value into a fatal error that kills the process -- and after some analysis last week we found that the highest % of addresses with a w28 index never got above 0.08% in any of our traces, which for many analysis purposes (aggregate statistics over various simulations, e.g.) we can live with more easily than we can live with the tracer just dying on automated tests and jobs where it used to successfully execute.

Is the real fix complex?  At first glance it seems like it may be quite simple?  Expand the index to pointer-sized before comparing; shrink the return value of replace_stolen_reg()?  A 2-line change?  I didn't spend very long looking at it so maybe I missed something.  If it is simple we might prefer to go straight to the fix.

suite/tests/client-interface/stolen-reg-index.dll.c
```diff
@@ -0,0 +1,137 @@
...
+        DRREG_SUCCESS) {
+        DR_ASSERT(false);
+    }
+
+    if (drreg_reserve_register(drcontext, ilist, instr, NULL, &reg_ptr) !=
+        DRREG_SUCCESS) {
+        DR_ASSERT(false);
+    }
+
+    if (!drutil_insert_get_mem_addr(drcontext, ilist, instr, mref, reg_tmp, reg_ptr)) {
```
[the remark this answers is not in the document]
AssadHashmi: > Please add a XXX comment or something.

Added XXX comment. Don't know how such tests can validate against addresses which can change.

ext/drutil/drutil.c
AssadHashmi: > Mixed feelings here...

Same here, I just put this in as a placeholder and a check to see which of our tests regressed.

> after some analysis last week we found that the highest % of addresses with a w28 index never got above 0.08% in any of our traces

That's a relief. I didn't think the % would be high, but that's much lower than I'd hoped for.

> Is the real fix complex? At first glance it seems like it may be quite simple?

I've put a fix in now. I was concerned about violating transparency but after some thought it's not a problem.

AssadHashmi:
There's only one other call to replace_stolen_reg() which doesn't need special handling. The base register will always be 64 bits.

This issue again highlights the question about how to improve DR and its test coverage in cases where bugs remain hidden by virtue of silent failure.

suite/tests/CMakeLists.txt
```diff
@@ -2557,6 +2557,13 @@ set(client.option_parse_client_ops_islist ON)
 
 if (AARCH64)
   tobuild_ci(client.features client-interface/features.c "" "" "")
+
+  tobuild_appdll(client.stolen-reg-index client-interface/stolen-reg-index.c)
```
derekbruening, line 2561: No longer needs the _appdll

suite/tests/client-interface/stolen-reg-index.c
```diff
@@ -0,0 +1,92 @@
...
+void
+indexed_mem_test(int *val);
+
+int
+main(int argc, char *argv[])
+{
+    int value = 41;
+    indexed_mem_test(&value);
+    if (value != 42) {
+        fprintf(stderr, "indexed_mem_test() failed with %d, expected 42.\n", value);
```
derekbruening, line 53: Our convention is to always use print() in tests which goes to stderr to interleave reliably with DR's stderr output, and is flushed everywhere (incl Windows cygwin).  Ditto below.

suite/tests/client-interface/stolen-reg-index.dll.c
```diff
@@ -0,0 +1,142 @@
...
+    return DR_EMIT_DEFAULT;
+}
+
+DR_EXPORT void
+dr_client_main(client_id_t id, int argc, const char *argv[])
+{
+    drreg_options_t ops = { sizeof(ops), 4 /*max slots needed*/, false };
+    if (!drmgr_init() || !drutil_init() || drreg_init(&ops) != DRREG_SUCCESS)
+        DR_ASSERT(false);
+
```
derekbruening, line 138: Assert that x28 is the stolen register in case we change the default.
Other tests do this today:
```
suite/tests/client-interface/drreg-test.dll.c:        CHECK(dr_get_stolen_reg() == TEST_REG_STOLEN, "stolen reg doesn't match");
```

derekbruening (approved):

ext/drutil/drutil.c
```diff
@@ -430,6 +443,15 @@ drutil_insert_get_mem_addr_arm(void *drcontext, instrlist_t *bb, instr_t *where,
         } else if (index == stolen) {
             index = replace_stolen_reg(drcontext, bb, where, memref, dst, scratch,
                                        scratch_used);
+#    ifdef AARCH64
+            /* Narrow replaced index register if it was W28 before
+             * replace_stolen_reg() call.
+             */
+            if (is_index_w28) {
+                index = DR_REG_START_32 + (index - DR_REG_START_64);
```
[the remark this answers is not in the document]
AssadHashmi: Done.

suite/tests/CMakeLists.txt
AssadHashmi: Done.

suite/tests/client-interface/stolen-reg-index.c
AssadHashmi: Done.

derekbruening:
This PR actually failed to fix the bug: https://github.com/DynamoRIO/dynamorio/issues/5498#issuecomment-1242470308

PR #5513, opened 2022-06-02
i#4014 dr$sim phys: Add different page size support

derekbruening:
Adds support for different page sizes from 4K for drmemtrace's
-use_physical.

Adds hugepage support, which does not require anything special: just
updated the comment.

Improves the error reporting for physical translation.

Tested on an AArch64 machine with 64K pages.

Issue: #4014

### DIFF
--- a/clients/drcachesim/tracer/tracer.cpp
+++ b/clients/drcachesim/tracer/tracer.cpp
@@ -89,12 +89,6 @@ drmemtrace_client_main(client_id_t id, int argc, const char *argv[]);
  */
 DR_DISALLOW_UNSAFE_STATIC
 
-#define NOTIFY(level, ...)                     \
-    do {                                       \
-        if (op_verbose.get_value() >= (level)) \
-            dr_fprintf(STDERR, __VA_ARGS__);   \
-    } while (0)
-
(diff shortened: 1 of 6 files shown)


clients/drcachesim/tracer/tracer.cpp
```diff
@@ -974,12 +968,11 @@ memtrace(void *drcontext, bool skip_size_cap)
                 if (have_phys && op_use_physical.get_value() &&
                     type != TRACE_TYPE_THREAD && type != TRACE_TYPE_THREAD_EXIT &&
                     type != TRACE_TYPE_PID) {
-                    addr_t virt = instru->get_entry_addr(mem_ref);
-                    addr_t phys = data->physaddr.virtual2physical(virt);
                     DR_ASSERT(type != TRACE_TYPE_INSTR_BUNDLE);
-                    if (phys != 0)
```
abhinav92003, line 980: We should document at `virtual2physical` that zero is a valid translation.

abhinav92003 (approved):

clients/drcachesim/tracer/tracer.cpp
derekbruening: Done.  Also mentioned high values due to holes.

Issue #5517, opened 2022-06-06
Add libipt as a submodule of dynamorio


dolanzhao:
Adding submodules in general brings up a bunch of auxiliary support we probably want.  I wonder if we should separate this out as its own PR.  I assume we want the submodule to be initialized all the time?  If so, things we would want:

+ We have to tell GA CI to init submodules: `submodules: true` arg to checkout step
+ Add submodule update to `make/git/devsetup.sh`
+ Add submodule update to `make/git/git_pullall.sh`
+ Potentially add to how-to-build docs if not completely automated
+ Potentially add a pre-commit hook to prevent the common mistake of accidentally rolling back a recent submodule update when switching branches

_Originally posted by @derekbruening in https://github.com/DynamoRIO/dynamorio/pull/5515#discussion_r889369065_

derekbruening:
Not sure this needed to be its own issue?  The reason libipt is desired is for dr$sim kernel traces #5505.

Issue #5505, opened 2022-05-26
Support kernel tracing in drcachesim for system calls


abhinav92003:
This issue is to track the design and implementation for the support to trace kernel instructions and data addresses for system calls in  drcachesim.

The current plan is to use Intel-PT to collect system call instruction traces in the drcachesim tracer. We can start kernel tracing by using perf_events_open + ioctl in a pre-syscall event, and end it using ioctl in a post-syscall event.

We can then convert the PT trace to a drcachesim trace using libipt, and interleave it into the user-mode trace gathered by drcachesim, in raw2trace.

An approximation of the data addresses accessed by the kernel can be reconstructed using some heuristics and metadata collected by drcachesim (e.g. system call parameters). There are some interesting research articles that implement similar techniques: https://ieeexplore.ieee.org/document/9065578.

prasun3:
Purely software based approach (like some prior academic work) not being considered?

https://sorav.compiler.ai/pubs/btkernel.pdf
http://www.cs.toronto.edu/~peter/feiner_asplos_2012.pdf

abhinav92003:
We do not want to assume the capability of modifying the kernel or loading a kernel module at runtime.

algr:
It ought to be possible to do this on Arm as well using cs_etm events; the equivalent decoder library is https://github.com/Linaro/OpenCSD. Would it be possible for this to be done in a way where the main logic is cross-architecture, and only the bits that are specific to PT (resp ETM) are in the Intel (resp Arm) parts?

abhinav92003:
Yes, we do plan to use the same hooks in the drmemtrace tracer and raw2trace that we use today for Intel-PT traces to also support ETM traces. We're blocked on getting access to required hardware with ETM capability though.

rejohn:
Is this feature fully supported now?

abhinav92003:
The basic support for collecting Intel-PT traces during system calls is present. We also have some unit tests but they're not run in our Github Actions test suite because they need a machine with Intel-PT support. You can run them manually on a suitable machine by following the instructions at https://dynamorio.org/page_drpttracer.html#sec_unit_tests.

There are still some missing pieces and rough edges in this feature though:
- PT traces are instruction-only. Our plan was to reconstruct an address stream that preserves some useful properties of the original (e.g., the stack distance profile), much like DWT from HPCA 2020 (https://ieeexplore.ieee.org/document/9065578).
[...]

PR #5515, opened 2022-06-03
i#5505 kernel trace: Add drpt2trace to decode PT traces

dolanzhao:
Adds a PT raw traces decoding library and a command-line tool for testing it:
- Adds drpt2ir. It is a static library that can decode PT raw traces. The next PR will support converting PT raw traces to DynamoRIO's IR(instr_t).
- Adds drpt2trace. It is a command-line tool that invokes drpt2ir and is used to test drpt2ir's decoding APIs. The current version can decode the PT raw traces and count the number of instructions.
- Updates clients/drcachesim/CMakeLists.txt to support building drpt2ir and drpt2trace.
- Updates suite/tests/CMakeLists.txt to add a simple test for drpt2trace. The simple trace data is recorded by perf.

Issue: #5505

PR #5518, opened 2022-06-06
i#5505 kernel memtrace: Add submodule libipt & Initialize submodules all the time

dolanzhao:
- Add libipt as a submodule
- Tell GA CI to init submodules: submodules: true arg to checkout step
- Add submodule update to make/git/devsetup.sh
- Add submodule update to make/git/git_pullall.sh
- Update how-to-build docs

Issue: #5505


derekbruening:
> [#5515 (comment)](https://github.com/DynamoRIO/dynamorio/pull/5515#discussion_r889369065)
> 
> > Potentially add a pre-commit hook to prevent the common mistake of accidentally rolling back a recent submodule update when switching branches
> 
> @derekbruening Could you explain this point in details?

If you check out the drmemory repo and do "git grep submodule" you will see all the support it has: it is part of make/git/hook-pre-commit.sh that looks useful for the DR repo.

api/docs/building.dox
```diff
@@ -47,7 +47,7 @@ To build DynamoRIO on Linux, use the following commands as a guide.  This builds
   # other distributions (in particular, use "cmake3" for Ubuntu Trusty).
   $ sudo apt-get install cmake g++ g++-multilib doxygen git zlib1g-dev libunwind-dev libsnappy-dev liblz4-dev
   # Get sources.
-  $ git clone https://github.com/DynamoRIO/dynamorio.git
+  $ git clone --recurse https://github.com/DynamoRIO/dynamorio.git
```
derekbruening, line 50: nit: I think the actual option is either --recursive or --recurse-submodules?  I guess it is allowing your abbreviation but maybe better to use the real name.

.github/workflows/ci-package.yml
```diff
@@ -59,6 +59,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
+        submodules: true
```
derekbruening, line 63: The checkout step is used in all the other .yml files too

make/git/git_pullall.sh
```diff
@@ -44,6 +44,8 @@ if test -z "${has_remote}"; then
 else
     echo "First, updating with rebase from remote ${branch}."
     git pull --rebase origin ${branch}
+    echo "\nSecond, updating submodules."
```
derekbruening, line 47: This will print a literal `\n`.  Use -e to interpret newlines.

make/git/git_pullall.sh
```diff
@@ -44,6 +44,8 @@ if test -z "${has_remote}"; then
 else
```
derekbruening, line 44: The submodule update should be in the if case above too

make/git/git_pullall.sh
```diff
@@ -44,6 +44,8 @@ if test -z "${has_remote}"; then
...
     echo "First, updating with rebase from remote ${branch}."
     git pull --rebase origin ${branch}
+    echo "\nSecond, updating submodules."
+    git submodule update --init --recursive
```
derekbruening, line 48: Hmm, drmemory does this after the pull below; I think that is what we want, to reflect the remote main branch changes

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "third_party/libipt"]
+    path = third_party/libipt
+    url = https://github.com/intel/libipt.git
```
derekbruening, line 3: I had assumed that libipt was not provided in the main distro repos: but it does seem to be there as `libipt-dev`.  It would be simpler to use that package, just like we do with many other libraries (e.g., for dr$sim compression: libsnappy, lz4, zlib).  Is there some very recent interface that is not part of that package that is the reason to build it from sources?

dolanzhao: Yes. libipt-dev only contains libipt.so. It can not decode perf recode trace data directly. Currently, drptpostprocessor uses the libipt sideband library (libipt-sb.so) to help handle context switches across images. So we need to build it from the source file. drptpostprocessor is a general solution for perf record data decoder. Using the sideband APIs is a quick solution because we don't need to write more code to handle sideband messages.
But for the data generated by our recorder, the kernel trace instructions may all be decoded from kcore, and we may only need libipt.so. In this case, libipt-dev might meet our requirements.

dolanzhao: I discussed this question with @abhinav92003 this morning. We think for the current version, there are three reasons that we need to add libipt as a submodule.
- The first reason is what I mentioned above, we need drptpostprocessor to handle trace data recorded by perf, using libipt-sb.so's API is an easy solution. 
- The second reason is except for the image switch, we may need to handle other information like timestamps in the future. 
- The last reason is we may need to handle the image switch when decoding kernel traces.  
So do you agree that we can include it first and if we don't need it we can remove it in the future.

derekbruening: OK, seems reasonable.  This is a good reason to isolate all the submodule parts in a separate PR so it can be reverted on its own.

.github/workflows/ci-aarchxx.yml
```diff
@@ -55,6 +55,8 @@ jobs:
```
derekbruening, line 55: The PR title:

> i#5517 add submodule libipt: Add libipt & Initialize submodules all the time #5518

The convention is to only use the short issue description before the colon for issue numbers addresses with multiple PR's.  Since you've made this single issue just for this submodule, with presumably a single PR addressing it, we would expect to not have that before-colon issue description.  (Or, put this PR under the #5505 issue and close #5517 as a duplicate/part of #5505.  If not doing that: please add a ref to #5505 in the Issue: list in the commit description.)

api/docs/building.dox
dolanzhao: Done.

.github/workflows/ci-package.yml
dolanzhao: Done.

make/git/git_pullall.sh
dolanzhao: Done.

make/git/git_pullall.sh
dolanzhao: Done.

make/git/git_pullall.sh
dolanzhao: Done.

.github/workflows/ci-aarchxx.yml
dolanzhao: Done.

make/git/hook-pre-commit.sh
```diff
@@ -57,4 +57,30 @@ if [ -z "${existing}" ]; then
     printf "Please double-check that it is configured properly.\n"
 fi
 
+# Prevent mistaken submodule commits by checking whether committing
+# submodule + other stuff and if so bail if the submodule version is older.
+# We want to allow a submodule-only rollback.
+# XXX: this relies on several unix utilities which we assume are available
+# on Windows.
```
derekbruening, line 64: Looks like a close copy from the drmemory version: but there's no really simple way to share unfortunately.  Maybe add a XXX comment that this is mostly a verbatim copy and if we set up some way to share (dedicated separate repository that's a submodule in both that provides shared config files or sthg?) we should be sure to share this too.

derekbruening (approved):

make/git/hook-pre-commit.sh
dolanzhao: Done.

abhinav92003:
Thanks for addressing the comments! In addition to commenting "done" in each conversation (which you did), please also click on the "resolve conversation" so that it doesn't show up anymore.

dolanzhao:
> Thanks for addressing the comments! In addition to commenting "done" in each conversation (which you did), please also click on the "resolve conversation" button so that it doesn't show up anymore.

OK. Thanks!

api/docs/building.dox
```diff
@@ -47,7 +47,7 @@ To build DynamoRIO on Linux, use the following commands as a guide.  This builds
   # other distributions (in particular, use "cmake3" for Ubuntu Trusty).
   $ sudo apt-get install cmake g++ g++-multilib doxygen git zlib1g-dev libunwind-dev libsnappy-dev liblz4-dev
   # Get sources.
-  $ git clone https://github.com/DynamoRIO/dynamorio.git
+  $ git clone --recursive https://github.com/DynamoRIO/dynamorio.git
```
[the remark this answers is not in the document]
dolanzhao: Done.

abhinav92003:
run arm tests

PR #5521, opened 2022-06-08
i#4014 dr$sim phys: Improve v2p cache and hash performance

derekbruening:
Switches from the drcontainers hashtable to the new open-address
hashtable provided by DR.  This is 2x to 3x faster due to the reduced
dereferences from the inlined data.

Increases the last-value cache from one entry to an array of 8
entries.  This was found to provide improved performance on small
benchmarks.

Measured on bzip2 local and SPEC2006 runs as they are short enough to
allow interactive experimentation.  -use_physical still incurs a ~2.5x
slowdown, but it was 9x before these changes.  The bottleneck is no
longer the hashtable but is now spread across all the address
iteration and querying code.

Issue: #4014

### DIFF
--- a/clients/drcachesim/tracer/physaddr.cpp
+++ b/clients/drcachesim/tracer/physaddr.cpp
@@ -60,14 +60,15 @@ static const addr_t PAGE_INVALID = (addr_t)-1;
 physaddr_t::physaddr_t()
 #ifdef LINUX
     : page_size_(dr_page_size())
-    , last_vpage_(PAGE_INVALID)
-    , last_ppage_(PAGE_INVALID)
+    , cache_idx_(0)
     , fd_(-1)
+    , v2p_(nullptr)
     , count_(0)
 #endif
 {
 #ifdef LINUX
-    v2p_.table = nullptr;
+    memset(last_vpage_, static_cast<char>(PAGE_INVALID), sizeof(last_vpage_));
+    memset(last_ppage_, static_cast<char>(PAGE_INVALID), sizeof(last_ppage_));
 
     page_bits_ = 0;
     size_t temp = page_size_;
@@ -82,8 +83,12 @@ physaddr_t::physaddr_t()
 physaddr_t::~physaddr_t()
 {
 #ifdef LINUX
-    if (v2p_.table != nullptr)
-        hashtable_delete(&v2p_);
+    NOTIFY(1,
+           "physaddr: hit cache: " UINT64_FORMAT_STRING
+           ", hit table " UINT64_FORMAT_STRING ", miss " UINT64_FORMAT_STRING "\n",
+           num_hit_cache_, num_hit_table_, num_miss_);
+    if (v2p_ != nullptr)
+        dr_hashtable_destroy(dr_get_current_drcontext(), v2p_);
 #endif
 }
 
@@ -93,8 +98,12 @@ physaddr_t::init()
 #ifdef LINUX
     // Some threads may not do much, so start out small.
     constexpr int V2P_INITIAL_BITS = 9;
-    hashtable_init_ex(&v2p_, V2P_INITIAL_BITS, HASH_INTPTR, /*strdup=*/false,
-                      /*synch=*/false, nullptr, nullptr, nullptr);
+    // The hashtable lookup performance is important.
+    // A non-open-address hashtable is about 3x slower due to the extra
+    // loads, and higher resize thresholds are also slower.
+    // With the setup here, the hashtable lookup is no longer the bottleneck.
+    v2p_ = dr_hashtable_create(dr_get_current_drcontext(), V2P_INITIAL_BITS, 20,
+                               /*synch=*/false, nullptr);
 
     // We avoid std::ostringstream to avoid malloc use for static linking.
     constexpr int MAX_PAGEMAP_FNAME = 64;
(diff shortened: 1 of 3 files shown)


clients/drcachesim/tracer/physaddr.cpp
```diff
@@ -60,14 +60,15 @@ static const addr_t PAGE_INVALID = (addr_t)-1;
...
-    , last_ppage_(PAGE_INVALID)
+    , cache_idx_(0)
     , fd_(-1)
+    , v2p_(nullptr)
     , count_(0)
 #endif
 {
 #ifdef LINUX
-    v2p_.table = nullptr;
+    memset(last_vpage_, static_cast<char>(PAGE_INVALID), sizeof(last_vpage_));
```
abhinav92003, line 70: Does this work as expected? It will set the first `NUM_CACHE` bytes of `last_vpage_` to -1 right?

clients/drcachesim/tracer/physaddr.cpp
```diff
@@ -93,8 +98,12 @@ physaddr_t::init()
 #ifdef LINUX
     // Some threads may not do much, so start out small.
     constexpr int V2P_INITIAL_BITS = 9;
-    hashtable_init_ex(&v2p_, V2P_INITIAL_BITS, HASH_INTPTR, /*strdup=*/false,
-                      /*synch=*/false, nullptr, nullptr, nullptr);
+    // The hashtable lookup performance is important.
+    // A non-open-address hashtable is about 3x slower due to the extra
```
abhinav92003, line 102: Curious about the terminology, why not just say closed address hashtable instead of non-open address hashtable? Maybe the latter includes some types of hashtable addressing schemes I'm not aware of?

clients/drcachesim/tracer/physaddr.cpp
```diff
@@ -93,8 +98,12 @@ physaddr_t::init()
...
+    // loads, and higher resize thresholds are also slower.
```
abhinav92003, line 103: Aren't the number of loads the same? It's just that the loads have better locality, right?

clients/drcachesim/tracer/physaddr.cpp
```diff
@@ -130,35 +140,41 @@ physaddr_t::virtual2physical(addr_t virt, OUT addr_t *phys, OUT bool *from_cache
         // XXX i#4014: Provide a similar option that doesn't flush and just checks
         // whether mappings have changed?
         use_cache = false;
-        last_vpage_ = PAGE_INVALID;
-        hashtable_clear(&v2p_);
+        memset(last_vpage_, static_cast<char>(PAGE_INVALID), sizeof(last_vpage_));
```
abhinav92003, line 143: So we don't reset `last_ppage_` because it won't be used if we reset `last_vpage`? Perhaps worth a comment.

clients/drcachesim/tracer/physaddr.cpp
```diff
@@ -130,35 +140,41 @@ physaddr_t::virtual2physical(addr_t virt, OUT addr_t *phys, OUT bool *from_cache
...
                 ppage = 0;
-            last_vpage_ = vpage;
-            last_ppage_ = ppage;
             if (from_cache != nullptr)
                 *from_cache = true;
-            *phys = last_ppage_ + page_offs(virt);
+            *phys = ppage + page_offs(virt);
+            last_vpage_[cache_idx_] = vpage;
+            last_ppage_[cache_idx_] = ppage;
+            cache_idx_ = (cache_idx_ + 1) % NUM_CACHE;
```
abhinav92003, line 172: I guess we're fine with FIFO instead of LRU here.

abhinav92003 (approved):
Sorry for the delay!

clients/drcachesim/tracer/physaddr.cpp
derekbruening: No, the regular table stores a pointer to the payload it has to de-reference, so there is an extra load.  The open address here has the data inlined.

clients/drcachesim/tracer/physaddr.cpp
derekbruening: I think I just never hear anyone say "closed address" and so I don't use the term: in my head it's always been open address or "not".

clients/drcachesim/tracer/physaddr.cpp
derekbruening: sizeof applied to an array is the total size of all the elements.  So it will set `NUM_CACHE*sizeof(last_vpage_[0])` to -1.

clients/drcachesim/tracer/physaddr.cpp
derekbruening: Yes, didn't seem worth extra complexity; didn't measure it though.

clients/drcachesim/tracer/physaddr.cpp
derekbruening: Yes.  Done.

clients/drcachesim/tracer/physaddr.cpp
derekbruening: Changed it to say "closed-address".

clients/drcachesim/tracer/physaddr.cpp
derekbruening: Added a mention of the inlined payload to the comment.

clients/drcachesim/tracer/physaddr.cpp
abhinav92003: argh. I searched for size of and just read array::size. Sorry for the noise.

derekbruening: If each byte has all the bits set (==-1), then the word will also have all the bits set and match PAGE_INVALID which is -1 cast to `addr_t==uintptr_t`.

Issue #5514, opened 2022-06-03
HANG tool.drcachesim.phys-threads once on GA CI


derekbruening:
The new tool.drcachesim.phys-threads test failed once in a PR with a hang:

https://github.com/DynamoRIO/dynamorio/runs/6716249823?check_suite_focus=true
```
] 90.00 sec

    -------------------------------------------------------------------
     Performance for solving AX=B Linear Equation using Jacobi method
     Running on DynamoRIO
     Client version (null)
    ...................................................................

     Matrix Size :  64
     Threads     :  4
```

Stopping at that early point in the output and then timing out sure looks like a hang.

Happened just the once.

Issue #5507, opened 2022-05-27
ASSERT tool.drcacheoff.windows-zlib signal.c:5729 dcontext != GLOBAL_DCONTEXT


derekbruening:
This happened once in PR #5506:

http://139.178.82.61:8080/job/DynamoRIO-AArch64-Precommit/1606/consoleFull
```
232:   *** postcmd failed (255): <Application
232:   /var/lib/jenkins/workspace/DynamoRIO-AArch64-Precommit/build/build_debug-internal-64/clients/bin64/drcachesim
232:   (1817526).  Internal Error: DynamoRIO debug check failure:
232:   /var/lib/jenkins/workspace/DynamoRIO-AArch64-Precommit/core/unix/signal.c:5729
232:   dcontext != GLOBAL_DCONTEXT
232: 
232:   (Error occurred @0 frags in tid 1817612)
```

joshua-warburton:
I believe I've found a reproducer for this. Spamming the test seems to cause this to fail about 5% of the time, but as this is a crash in the postcmd I managed to extract an output that causes it to always fail. I've attached this in a tar and you can reproduce by:
```
tar -xzf ../i5507-reproduce.tar.gz
drcachesim -indir i5507-reproduce -simulator_type basic_counts
```
[i5507-reproduce.tar.gz](https://github.com/DynamoRIO/dynamorio/files/10840380/i5507-reproduce.tar.gz)

I'll also submit a patch marking this as flaky

PR #5523, opened 2022-06-08
Document Linx cpu id versus socket split

derekbruening:
Documents which bits hold the cpu number versus the socket/node number
on Linux for the drmemtrace marker #TRACE_MARKER_TYPE_CPU_ID.


derekbruening:
Grrr we're going to need more cleanup work: I'm hitting flaky tests quite frequently these days.  phys-threads failed again: #5514.  Jenkins failed too in `code_api|tool.drcacheoff.raw-zlib `: signal assert #5507.

abhinav92003 (approved):

PR #5526, opened 2022-06-11
Fixed instructions for building 64-bit VS 2019

prasun3:
Also changed the description to have 64-bit as default.


derekbruening (approved):
Thank you for the PR!

Issue #5527, opened 2022-06-11
Android build failing due to NDK setup problems


derekbruening:
This started happening yesterday and has caused every Android build to fail, including the weekly package:

https://github.com/DynamoRIO/dynamorio/actions/runs/2478571712

https://github.com/DynamoRIO/dynamorio/runs/6840580519?check_suite_focus=true

```
...
2022-06-11 03:14:38 (67.2 MB/s) - ‘android-ndk-r10e-linux-x86_64.zip’ saved [1110915721/1110915721]
ERROR: Your ANDROID_NDK_ROOT variable does not point to a valid directory.
Error: Process completed with exit code 1.
```

abhinav92003:
@dolanzhao also hit this on #5515.

derekbruening:
Android support is not in great shape in general: there are private loader problems on Android 9 #3543 and Android 10 #3683.  (#5437's loader refactoring proposal may help there.)

There is DR-on-Android interest out there and there have been some PR's trying to improve things: maybe we should ask the community to help maintain by getting Android regression tests working.  The current tiny set of regular maintainers just do not have the bandwidth to maintain Android and it would simplify things to drop Android support, unless more maintainers step up.

derekbruening:
I was going to mark the Android build as not required for PR merging -- but it's just part of the arm cross-compile job.  Maybe we should split it out.

PR #5532, opened 2022-06-15
i#5527: fix android build broken by preset ANDROID_NDK_ROOT

dolanzhao:
Fixes the problem of NDK setup failing due to ANDROID_NDK_ROOT pointing to the wrong path. The GitHub VM system preset ANDROID_NDK_ROOT to a location which does not include the necessary resources that the setup process needs. We set the ANDROID_NDK_ROOT to the right location before running the setup script. 

Fixed: #5527

### DIFF
--- a/.github/workflows/ci-aarchxx.yml
+++ b/.github/workflows/ci-aarchxx.yml
@@ -209,6 +209,7 @@ jobs:
         cd /tmp
         wget https://dl.google.com/android/repository/android-ndk-r10e-linux-x86_64.zip
         unzip -q android-ndk-r10e-linux-x86_64.zip
+        export ANDROID_NDK_ROOT=/tmp/android-ndk-r10e
         android-ndk-r10e/build/tools/make-standalone-toolchain.sh --arch=arm \
           --toolchain=arm-linux-androideabi-4.9 --platform=android-21 \
           --install-dir=/tmp/android-gcc-arm-ndk-10e
--- a/.github/workflows/ci-package.yml
+++ b/.github/workflows/ci-package.yml
@@ -356,6 +356,7 @@ jobs:
         cd /tmp
         wget https://dl.google.com/android/repository/android-ndk-r10e-linux-x86_64.zip
         unzip -q android-ndk-r10e-linux-x86_64.zip
+        export ANDROID_NDK_ROOT=/tmp/android-ndk-r10e
         android-ndk-r10e/build/tools/make-standalone-toolchain.sh --arch=arm \
           --toolchain=arm-linux-androideabi-4.9 --platform=android-21 \
           --install-dir=/tmp/android-gcc-arm-ndk-10e


.github/workflows/ci-aarchxx.yml
```diff
@@ -209,6 +209,7 @@ jobs:
         cd /tmp
         wget https://dl.google.com/android/repository/android-ndk-r10e-linux-x86_64.zip
         unzip -q android-ndk-r10e-linux-x86_64.zip
+        export ANDROID_NDK_ROOT=/tmp/android-ndk-r10e
```
derekbruening, line 212: Please fix the package build one as well (unfortunately these command sequences are just duplicated; maybe we could make little shell scripts, or maybe there's an include or function mechanism in .yml files?  not sure):
```
$ git grep android-ndk-r10e-linux-x86_64
.github/workflows/ci-aarchxx.yml:        wget https://dl.google.com/android/repository/android-ndk-r10e-linux-x86_64.zip
.github/workflows/ci-aarchxx.yml:        unzip -q android-ndk-r10e-linux-x86_64.zip
.github/workflows/ci-package.yml:        wget https://dl.google.com/android/repository/android-ndk-r10e-linux-x86_64.zip
.github/workflows/ci-package.yml:        unzip -q android-ndk-r10e-linux-x86_64.zip
api/docs/building.dox:wget https://dl.google.com/android/repository/android-ndk-r10e-linux-x86_64.zip
api/docs/building.dox:unzip android-ndk-r10e-linux-x86_64.zip
```
Probably don't need to update the building.dox one.

.github/workflows/ci-aarchxx.yml
```diff
@@ -209,6 +209,7 @@ jobs:
         cd /tmp
```
derekbruening, line 209: > Fixes the problem of NDK setup failing due to ANDROID_NDK_ROOT point a wrong position. The GitHub VM system 

grammar: s/point a wrong position/pointing to the wrong path/

> preset ANDROID_NDK_ROOT to a location which not include the necessary resources that the setup process needs. 

grammar: s/not/does not/

dolanzhao: Done.

.github/workflows/ci-aarchxx.yml
derekbruening: (To clarify: I'm not requesting that you solve the command duplication: if we did find a way to solve it, we would do a separate PR, and it's not high priority right now.)

.github/workflows/ci-aarchxx.yml
abhinav92003, line 209: About the PR title: let's try to change it to something like the following shorter version, so that it doesn't get wrapped by Github UI. 

`i#5527: fix android build broken by preset ANDROID_NDK_ROOT`

The remaining details can then be included in the PR description instead

.github/workflows/ci-aarchxx.yml
dolanzhao: Look at this [answer](https://stackoverflow.com/a/15437697). It said YML files don't support the include mechanism.
And if I want to split this section into an script, which directory do you want to put it? make/.github, .github/scripts or tools/ci-scripts?

.github/workflows/ci-aarchxx.yml
dolanzhao: Done.

.github/workflows/ci-aarchxx.yml
dolanzhao: Got it. Done.

derekbruening (approved):

abhinav92003 (approved):

PR #5535, opened 2022-06-16
Fix maybe-uninitialized compiler warnings

fmoessbauer:
This minimal fix ensures that the cur_range_*_fd variables
are always initialized to make it compatible with the gcc compiler
flag -Wmaybe-uninitialized.

Technically, this does not fix any bug but supresses the warning
as the compiler cannot see through the for loop that changes the
variables value.

### DIFF
--- a/core/unix/os.c
+++ b/core/unix/os.c
@@ -7748,7 +7748,8 @@ pre_system_call(dcontext_t *dcontext)
             DODEBUG({ dcontext->expect_last_syscall_to_fail = true; });
             break;
         }
-        uint cur_range_first_fd, cur_range_last_fd;
+        uint cur_range_first_fd = 0;
+        uint cur_range_last_fd = 0;
         bool cur_range_valid = false;
         int ret = 0;
         for (int i = first_fd; i <= last_fd; i++) {


abhinav92003 (approved):

core/unix/os.c
```diff
@@ -7748,7 +7748,8 @@ pre_system_call(dcontext_t *dcontext)
             DODEBUG({ dcontext->expect_last_syscall_to_fail = true; });
```
abhinav92003, line 7748: nit: in the PR title 
s/fix/Fix/

abhinav92003 (approved):

core/unix/os.c
fmoessbauer: Done

fmoessbauer:
@abhinav92003 Any further changes required?

abhinav92003:
> @abhinav92003 Any further changes required?

Nope, that's it.

I see that you are already a member of the "Committers" group, so you should have commit privileges in this repository. You can merge this PR into master now. Remember to follow our workflow on cleaning up the final commit message: https://dynamorio.org/page_code_reviews.html#autotoc_md116.

In case you run into any issues, let me know. Thanks for the contribution!

PR #5536, opened 2022-06-17
i#4014 dr$sim phys: Handle page-spanning accesses

derekbruening:
Adds handling of page-spanning instruction fetch and data accesses.
For offline traces these are heuristics as we do not know the precise
sizes at tracing time.  Since an extra physical page value is not
harmful, the heuristics emit a second page if there is any decent
change that it is reached.

Adds a data page span test by switching the drcacheoff.phys test to
use allasm_repstr and adding such an access there.  An instruction
cross is more difficult to synthesize.

Issue: #4014

### DIFF
--- a/clients/drcachesim/tests/allasm-repstr-basic-counts.templatex
+++ b/clients/drcachesim/tests/allasm-repstr-basic-counts.templatex
@@ -11,11 +11,11 @@ Adios world!
 ---- <application exited with code 0> ----
 Basic counts tool results:
 Total counts:
-          95 total \(fetched\) instructions
-          23 total unique \(fetched\) instructions
+          98 total \(fetched\) instructions
+          26 total unique \(fetched\) instructions
            4 total non-fetched instructions
            0 total prefetches
-           5 total data loads
+           7 total data loads
            5 total data stores
            0 total icache flushes
            0 total dcache flushes
@@ -28,11 +28,11 @@ Total counts:
            0 total function return value markers
            4 total other markers
 Thread [0-9]* counts:
-          95 \(fetched\) instructions
-          23 unique \(fetched\) instructions
+          98 \(fetched\) instructions
+          26 unique \(fetched\) instructions
            4 non-fetched instructions
            0 prefetches
-           5 data loads
+           7 data loads
            5 data stores
            0 icache flushes
            0 dcache flushes
--- a/clients/drcachesim/tests/allasm_repstr.asm
+++ b/clients/drcachesim/tests/allasm_repstr.asm
@@ -55,8 +55,8 @@ _start:
 
         // Test page-spanning accesses.
         lea      rcx, page_str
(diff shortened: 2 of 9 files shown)


clients/drcachesim/tests/allasm_repstr.asm
```diff
@@ -53,6 +53,13 @@ _start:
         cld
         rep      movsb
 
+        // Test page-spanning accesses.
+        lea      rcx, page_str
+        // Somehow the assembler is adding 4 to the displacements here
```
abhinav92003, line 58: That's interesting. So rcx and rcx-4 will read the same address?

clients/drcachesim/tests/allasm_repstr.asm
```diff
@@ -76,3 +83,6 @@ hello_str:
         .string  "Hello world!\n"
 bye_str:
         .string  "Adios\n"
+        .align   4096
```
abhinav92003, line 86: Add comment that we need this for the multi-page check.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -987,6 +987,91 @@ output_buffer(void *drcontext, per_thread_t *data, byte *buf_base, byte *buf_ptr
     return current_num_refs;
 }
 
+static byte *
+process_entry_for_physaddr(void *drcontext, per_thread_t *data, size_t header_size,
+                           byte *v2p_ptr, byte *mem_ref, addr_t virt, trace_type_t type,
+                           bool *emitted, size_t *skip)
+{
```
abhinav92003, line 994: Check emitted != NULL.

abhinav92003 (approved):

clients/drcachesim/tracer/tracer.cpp
derekbruening: Done, along with skip.

clients/drcachesim/tests/allasm_repstr.asm
derekbruening: Done, along w/ assumption on 4K pages.

clients/drcachesim/tests/allasm_repstr.asm
derekbruening: No, they would read addresses 4 bytes apart, but +4 from what was written.  It is very strange.  It adds the load size to the displacement.  Happens with GNU as 2.38.  Added the version to the comment.  It must be some kind of bug, maybe with the Intel syntax flag; may try to see if it's known offline.

clients/drcachesim/tests/allasm-repstr-basic-counts.templatex
```diff
@@ -11,11 +11,11 @@ Adios world!
 ---- <application exited with code 0> ----
```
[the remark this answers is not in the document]
derekbruening: Will do on merge.

PR #5540, opened 2022-06-21
Minor formatting fixes in the view tool

abhinav92003:
Separates out header printing to a separate routine that can
be invoked by sub-classes instead of duplicating the print logic.

Sets width of instr/data bytes to 2, so that double-digit byte
memrefs are formatted the same as single-digit ones.

### DIFF
--- a/clients/drcachesim/tools/view.h
+++ b/clients/drcachesim/tools/view.h
@@ -80,6 +80,14 @@ protected:
     should_skip(const memref_t &memref);
 
     inline void
+    print_header()
+    {
+        std::cerr << std::setw(9) << "Output format:\n<record#>"
+                  << ": T<tid> <record details>\n"
+                  << "------------------------------------------------------------\n";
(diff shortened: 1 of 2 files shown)


clients/drcachesim/tools/view.h
```diff
@@ -79,6 +79,14 @@ class view_t : public analysis_tool_t {
     bool
     should_skip(const memref_t &memref);
 
+    inline void
+    print_header()
```
derekbruening, line 83: The description says you want to override this in subclasses: so it needs to be virtual.

derekbruening (approved):

clients/drcachesim/tools/view.h
abhinav92003: Not override, just invoke. E.g. if the sub-class overrides the (virtual) initialize routine, it will invoke this print_header routine by itself.

Issue #5534, opened 2022-06-16
Return of local addr in fragment_add_ibl_target


fmoessbauer:
With gcc 10 and `-Werror=return-local-addr` it reports that a local reference is returned in fragment_add_ibl_target.

For me it looks like this is an actual bug:

1. Returned reference to `fragment_t wrapper` stack-variable: https://github.com/DynamoRIO/dynamorio/blob/master/core/fragment.c#L4106
2. Ref passed to `fragment_coarse_lookup_wrapper` as arg: https://github.com/DynamoRIO/dynamorio/blob/master/core/fragment.c#L4111
3. Ref to wrapper returned from `fragment_coarse_lookup_wrapper`: https://github.com/DynamoRIO/dynamorio/blob/master/core/fragment.c#L8052
4. Ref returned from fragment_add_ibl_target: https://github.com/DynamoRIO/dynamorio/blob/master/core/fragment.c#L4139

I did not check if this is actually a realistic code path, but theoretically it should be valid and IMO should be fixed by making a copy of the wrapper. Note, that this path can only be reached in some configurations (with `defined(RETURN_AFTER_CALL) || defined(RCT_IND_BRANCH)`).

**To Reproduce**

Compile with gcc 10 and `-Wall -Werror=return-local-addr`

**Versions**
 Current master (c3b73b0224c88b656da067819cf41787ee98fa3f) / 9.0.19146

Note: found while working on #5153

abhinav92003:
Thanks for the report!

I see that the existing usages of `fragment_add_ibl_target` don't use the returned fragment ptr at all. We can just change the routine to a void return type for now. If we ever need it to return an actual value, we can allocate a fragment then.

abhinav92003:
Do you have a local fix? Can you send it as a PR for review?

PR #5542, opened 2022-06-22
Fix return of local addr in fragment_add_ibl_target

fmoessbauer:
This fixes the return of a local stack addr by making the
function void. This is an internal API and the return value
is nowhere used across DynamoRIO.

closes: #5534

### DIFF
--- a/core/fragment.c
+++ b/core/fragment.c
@@ -4098,7 +4098,7 @@ fragment_add_ibl_target_helper(dcontext_t *dcontext, fragment_t *f,
 }
 
 /* IBL targeted fragments per branch type */
-fragment_t *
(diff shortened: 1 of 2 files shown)


core/fragment.c
```diff
@@ -4098,7 +4098,7 @@ fragment_add_ibl_target_helper(dcontext_t *dcontext, fragment_t *f,
 }
```
abhinav92003, line 4098: nit in the PR title: we generally don't have `fix:` prefixes in the title.

s/fix: return/Fix return/

abhinav92003 (approved):
Thank you for the fix!

Issue #5153, opened 2021-10-11
Simplify packaging of DynamoRIO


fmoessbauer:
**Add a pointer to any prior users list discussion.**

This issue is a follow up of the discussion started in !5152.

**Is your feature request related to a problem? Please describe.**

Currently, packaging of DynamoRIO for OS distributions like Debian is not possible (without patches) due to a list of incompatibilities. The following list should help to keep track of the current status and changes that will be required.

**Describe the solution you'd like**

The following aspects should be changed:

- CMake
  - [ ] Add support for install-relocation via DESTDIR env variable !5152
  - [ ] libdynamorio-dev: Make CMake scripts of DynamoRIO work with system install paths
- drrun
  - [ ] rework injection logic to support other directory structures as well (e.g. `/usr/lib/<multiarch-triplet>/dynamorio.so`, config in `/etc/dynamorio`, etc...)
  - [ ] Find installed dynamorio tools / clients via conf.d style integration (each new client drops a file in `/etc/dynamorio/tools.d/`). This makes it possible for third-party tools that are packaged as well to easily integrate into DR.
- Libraries
  - [ ] Check if the `libdynamorio.so` needs the execution bit or if there are other strategies. Installing a library with execution bit set is discouraged or even forbidden for some distros.
  - [ ] Add SO version, describing ABI compatibility

**Describe alternatives you've considered**

It is possible to work around some limitations (like expected directory layout of drrun) by using wrappers and ignoring checks. However, this is neither a clean solution, nor sustainable.

**Additional context**

- Debian policies: https://www.debian.org/doc/debian-policy/index.html
- Debian guide ["How do I add a new package to the archive?"](https://wiki.debian.org/DebianMentorsFaq#How_do_I_add_a_new_package_to_the_archive.3F)
- please note, I'm not a Debian "sponsor", so I cannot integrate that directly

derekbruening:
We pull in and run the Dr. Memory CMake code (triggered by an option though so not required) to include it in our packages and it would need some of the same types of changes to be included in a distro package.

fmoessbauer:
When building the current version of DynamoRIO in a `schroot` (via Debian sbuilder), the build breaks at the following step. When building the exact same configuration outside of the chroot, the build finishes correctly.

```
[...]

PR #5543, opened 2022-06-24
Only enable alt-arch tools/clients when available

fmoessbauer:
This patch adds a check when parsing the tool config files:
Clients for alternative archs (e.g. for cross-exec) injection
are only appended if the path to the client lib actually
exists.

This is a preparation for system-wide deployments, where this
check has to be performed at runtime as users might only
install the dr library for a single architecture.

Previously having invalid paths in the tool config file
worked if the respective client was not requested, but led to a warning.
With this patch, the missing client is simply not registered at all.

Issue: #5153

### DIFF
--- a/tools/drdeploy.c
+++ b/tools/drdeploy.c
@@ -976,7 +976,6 @@ read_tool_file(const char *toolname, const char *dr_root, const char *dr_toolcon
             alt_client[alt_size - 1] = '\0';
             if (!does_file_exist(alt_client)) {
                 alt_client[0] = '\0';
-                continue;
             }
             if (native_path[0] != '\0') {
                 add_extra_option(tool_ops, tool_ops_size, tool_ops_sofar, "\"%s\"",
@@ -1003,7 +1002,6 @@ read_tool_file(const char *toolname, const char *dr_root, const char *dr_toolcon
                     alt_size);
             if (!does_file_exist(alt_client)) {
                 alt_client[0] = '\0';
-                continue;
             }
             if (native_path[0] != '\0') {
                 add_extra_option(tool_ops, tool_ops_size, tool_ops_sofar, "\"%s\"",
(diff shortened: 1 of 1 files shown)


fmoessbauer:
This is a pre-condition for #5153

tools/drdeploy.c
```diff
@@ -970,6 +970,10 @@ read_tool_file(const char *toolname, const char *dr_root, dr_platform_t dr_platf
             _snprintf(alt_client, alt_size, "%s/%s", dr_root,
```
abhinav92003, line 970: nit in the PR title: s/only/Only/

abhinav92003, line 970: In the PR description, please include `Issue: #5153`, as this PR is relevant to that issue.

abhinav92003 (approved):

tools/drdeploy.c
```diff
@@ -970,6 +970,10 @@ read_tool_file(const char *toolname, const char *dr_root, dr_platform_t dr_platf
...
                       line + strlen(IF_X64_ELSE("CLIENT32_REL=", "CLIENT64_REL=")));
```
derekbruening, line 971: PR description:
> worked if the respective client was not requested, but lead to a warning.

nit: s/lead/led/

fmoessbauer:
@abhinav92003 For me it looks like the CI failures are unrelated. Do you have some issues with flaky tests?

abhinav92003:
> @abhinav92003 For me it looks like the CI failures are unrelated. Do you have some issues with flaky tests?

Looks like some tests are consistently  failing in this PR, which are passing in others (like #5549). The error seems to be:
```
ERROR: unable to inject
```
May be worth investigating. It's possible that the Github Actions runner has some special setup that we should handle. Check out the section on `Debugging Tests on Github Actions Runner`: https://dynamorio.org/page_test_suite.html

fmoessbauer:
Found it: We cannot simply "continue" in case the alt-client is not found but have to run the `add_extra_option` as well. This is required for drcachesim. Luckily the CI caught that.

PR #5548, opened 2022-06-26
Add support to overwrite default tool config dir

fmoessbauer:
This patch adds a new parameter -tool_dir to specify in which folder tool configuration files should be looked up.
This is a preparation for issue #5153.

### DIFF
--- a/tools/drdeploy.c
+++ b/tools/drdeploy.c
@@ -407,6 +407,9 @@ read_tool_list(const char *dr_root, dr_platform_t dr_platform)
     char list_file[MAXIMUM_PATH];
     size_t sofar = 0;
     const char *arch = IF_X64_ELSE("64", "32");
+    /* clear global tool list on re-read */
+    tool_list[0] = '\0';
+
     if (dr_platform == DR_PLATFORM_32BIT)
         arch = "32";
     else if (dr_platform == DR_PLATFORM_64BIT)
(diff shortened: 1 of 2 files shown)


tools/drdeploy.c
```diff
@@ -1404,7 +1417,15 @@ _tmain(int argc, TCHAR *targv[])
             /* support -dr_home alias used by script */
             strcmp(argv[i], "-dr_home") == 0) {
             dr_root = argv[++i];
-            read_tool_list(dr_root, dr_platform);
+            /* modify default toolconfig dir only.
```
derekbruening, line 1420: style nits to prefer capitalized complete sentences (old code doesn't always follow it but we'd like new code to do so):

s/modify/Modify/; s/the default/the default./

tools/drdeploy.c
```diff
@@ -318,6 +318,7 @@ const char *options_list_str =
 #    endif
     "       -use_dll <dll>     Inject given dll instead of configured DR dll.\n"
     "       -use_alt_dll <dll> Use the given dll as the alternate-bitwidth DR dll.\n"
+    "       -toolsdir <dir>    Directory containing tool configuration files.\n"
```
derekbruening, line 321: Hmm, is "-tooldir" or "-tool_dir" better?  Subjective I guess; singular ("tool" instead of "tools") sounds better to me b/c of English grammar preferring singular in compounds (e.g., "dogwalker" not "dogswalker") but if you strongly prefer "-toolsdir" go with that.

derekbruening (approved):

PR #5549, opened 2022-06-27
Rename nudgeunix to drnudgeunix

fmoessbauer:
This patch renames the nudgeunix tool to drnudgeunix to be
named more consistently with the rest of DynamoRIOs frontend tooling.
This helps for packaging as nudgeunix is somehow non-specific.

For backward compatibility, a symlink with the old name is added.

Related to issue #5153

### DIFF
--- a/tools/drdeploy.c
+++ b/tools/drdeploy.c
@@ -265,8 +265,8 @@ const char *options_list_str =
     "                          wait.  A value of 0 means don't wait for nudges to\n"
     "                          complete."
 #    else  /* WINDOWS */
-    /* FIXME i#840: integrate nudgeunix into drconfig on Unix */
-    "Note: please use the nudgeunix tool to nudge processes on Unix.\n";
+    /* FIXME i#840: integrate drnudgeunix into drconfig on Unix */
+    "Note: please use the drnudgeunix tool to nudge processes on Unix.\n";
 #    endif /* !WINDOWS */
 #else      /* DRCONFIG */
     "       -no_wait           Return immediately: do not wait for application exit.\n"
--- a/api/docs/deployment.dox
+++ b/api/docs/deployment.dox
@@ -376,9 +376,9 @@ than invoking the script directly:
 
 To \ref sec_comm "nudge" a process with pid \c targetpid running under
 DynamoRIO and pass argument "5" to the nudge callback, use the \c
-nudgeunix tool:
+drnudgeunix tool:
 \code
-bin32/nudgeunix -pid targetpid -client 0 5
+bin32/drnudgeunix -pid targetpid -client 0 5
 \endcode
 This will result in a nudge event with argument=5 delivered to the
 client callback registered with dr_register_nudge_event() in the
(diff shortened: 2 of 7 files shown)


derekbruening:
I don't remember why this is, but drrun has a -nudge option for Windows -- yet for Linux nudging is in a separate program.  Should we remove this separate drnudgeunix and put its functionality into drrun to match Windows?

Happy to rename for now if we want to file an issue on that for later.

tools/drdeploy.c
```diff
@@ -265,8 +265,8 @@ const char *options_list_str =
     "                          wait.  A value of 0 means don't wait for nudges to\n"
     "                          complete."
 #    else  /* WINDOWS */
-    /* FIXME i#840: integrate nudgeunix into drconfig on Unix */
-    "Note: please use the nudgeunix tool to nudge processes on Unix.\n";
+    /* FIXME i#840: integrate drnudgeunix into drconfig on Unix */
```
derekbruening, line 268: Ah here's an already-filed issue on my top-level comment: ok let's leave it for that issue.
No action requested.  (Githhub should add a type of comment that is already resolved.)

tools/drdeploy.c
```diff
@@ -265,8 +265,8 @@ const char *options_list_str =
...
+    "Note: please use the drnudgeunix tool to nudge processes on Unix.\n";
```
derekbruening, line 269: I'm seeing a bunch of other references that need updating, right?
```
$ git grep nudgeunix
api/docs/deployment.dox:nudgeunix tool:
api/docs/deployment.dox:bin32/nudgeunix -pid targetpid -client 0 5
core/lib/dr_events.h: * nudgeunix tool on Linux.  A client in this process can use
core/unix/nudgesig.c: * shared between core & tools/nudgeunix.c
core/unix/nudgesig.c:/* shared with tools/nudgeunix.c */
suite/tests/CMakeLists.txt:        get_target_path_for_execution(toolbindir nudgeunix "${location_suffix}")
suite/tests/runall.cmake:# * nudge = arguments to nudgeunix or drconfig
suite/tests/runall.cmake:  set(nudge_cmd nudgeunix)
suite/tests/runall.cmake:  # nudgeunix and drconfig have different syntax:
tools/CMakeLists.txt:    add_executable(nudgeunix nudgeunix.c ${PROJECT_SOURCE_DIR}/core/unix/nudgesig.c)
tools/CMakeLists.txt:    DR_install(TARGETS nudgeunix DESTINATION ${INSTALL_BIN})
tools/drdeploy.c:    /* FIXME i#840: integrate nudgeunix into drconfig on Unix */
tools/drdeploy.c:    "Note: please use the nudgeunix tool to nudge processes on Unix.\n";
tools/nudgeunix.c: * nudgeunix.c: nudges a target Linux process
tools/nudgeunix.c:    "usage: nudgeunix [-help] [-v] [-pid <pid>] [-type <type>] [-client <ID> <arg>]\n"
tools/nudgeunix.c:            printf("nudgeunix version %s -- build %d\n", STRINGIFY(VERSION_NUMBER),
```

fmoessbauer: This grep is run on a different source tree. But you are right. I still missed a couple of references.

fmoessbauer:
@derekbruening I just renamed all references, except for the filename (`nudgeunix.c`). Renaming this as well adds only little value but destroys the git history.

derekbruening (approved):

PR #5551, opened 2022-06-28
Build libutil when clients are built to fix drcachesim dependency

fmoessbauer:
Before this patch, libutil was not built in case building of
BUILD_TOOLS and BUILD_DRSTATS was disabled. This lead to a
link-time error as drcachesim (a client) depends on drfrontendlib.

This is now fixed by also building libutil when enabling BUILD_CLIENTS.


derekbruening (approved):

abhinav92003 (approved):
Apologies -- I missed this notification.

PR #5552, opened 2022-06-29
i#5505 kernel trace: Support octal and hex values in droption

dolanzhao:
- Enhance the conversion of integer options to support parsing octal or hexadecimal format strings
- Added negative format string checking before unsigned integer option's parsing
- Added test cases in the  test client 'code_api|client.option_parse'. The new test cases use integer options to parse octal and hexadecimal format strings and unsigned integers to parse the negative format strings.

Issue: #5505

### DIFF
--- a/ext/droption/droption.h
+++ b/ext/droption/droption.h
@@ -715,7 +715,7 @@ inline bool
 droption_t<unsigned long>::convert_from_string(const std::string s)
 {
     // Checks if the first non-space character of a string is a negative sign. If it is,
-    // it is invalid.
+    // the input string is invalid.
     for (size_t i = 0; i < s.size(); i++) {
         // XXX: The function isspace() only identify ASCII whitespace characters. We
         // should implement one function that can support non-ASCII whitespace characters.
@@ -735,7 +735,7 @@ inline bool
 droption_t<unsigned long long>::convert_from_string(const std::string s)
 {
     // Checks if the first non-space character of a string is a negative sign. If it is,
-    // it is invalid.
+    // the input string is invalid.
     for (size_t i = 0; i < s.size(); i++) {
         // XXX: The function isspace() only identify ASCII whitespace characters. We
         // should implement one function that can support non-ASCII whitespace characters.
--- a/ext/droption/droption.h
+++ b/ext/droption/droption.h
@@ -371,6 +371,8 @@ protected:
     convert_from_string(const std::string s1, const std::string s2) = 0;
     virtual bool
     clamp_value() = 0;
+    virtual bool
+    is_negative_input(const std::string &s) = 0;
     virtual std::string
     default_as_string() const = 0;
     virtual void
@@ -571,6 +573,23 @@ protected:
     }
 
     bool
+    is_negative_input(const std::string &s) override
+    {
+        // Checks if the first non-space character of a string is a negative sign.
+        for (size_t i = 0; i < s.size(); i++) {
+            // XXX: The function isspace() only identify ASCII whitespace characters. We
+            // should implement one function that can support non-ASCII whitespace
+            // characters.
+            if (isspace(s[i]))
+                continue;
+            if (s[i] == '-')
+                return true;
+            break;
+        }
+        return false;
+    }
+
+    bool
     option_takes_arg() const override;
     bool
     option_takes_2args() const override;
@@ -691,8 +710,7 @@ inline bool
 droption_t<long long>::convert_from_string(const std::string s)
 {
     errno = 0;
-    // If we set 0 as the base, strtoll() will automatically identify the base like
-    // strtol().
+    // If we set 0 as the base, strtoll() will automatically identify the base.
     value_ = strtoll(s.c_str(), NULL, 0);
     return errno == 0;
 }
@@ -714,17 +732,8 @@ template <>
 inline bool
 droption_t<unsigned long>::convert_from_string(const std::string s)
 {
-    // Checks if the first non-space character of a string is a negative sign. If it is,
-    // the input string is invalid.
-    for (size_t i = 0; i < s.size(); i++) {
-        // XXX: The function isspace() only identify ASCII whitespace characters. We
-        // should implement one function that can support non-ASCII whitespace characters.
-        if (isspace(s[i]))
-            continue;
-        if (s[i] == '-')
-            return false;
-        break;
-    }
+    if (is_negative_input(s))
+        return false;
 
     errno = 0;
     value_ = strtoul(s.c_str(), NULL, 0);
@@ -734,17 +743,8 @@ template <>
 inline bool
 droption_t<unsigned long long>::convert_from_string(const std::string s)
 {
-    // Checks if the first non-space character of a string is a negative sign. If it is,
-    // the input string is invalid.
-    for (size_t i = 0; i < s.size(); i++) {
-        // XXX: The function isspace() only identify ASCII whitespace characters. We
-        // should implement one function that can support non-ASCII whitespace characters.
-        if (isspace(s[i]))
-            continue;
-        if (s[i] == '-')
-            return false;
-        break;
-    }
+    if (is_negative_input(s))
+        return false;
 
     errno = 0;
     value_ = strtoull(s.c_str(), NULL, 0);
--- a/ext/droption/droption.h
+++ b/ext/droption/droption.h
@@ -47,6 +47,7 @@
 #include <limits.h>
 #include <stdint.h>
 #include <errno.h>
+#include <ctype.h>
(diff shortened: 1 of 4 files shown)


suite/tests/client-interface/option_parse.dll.cpp
```diff
@@ -198,8 +246,42 @@ dr_client_main(client_id_t client_id, int argc, const char *argv[])
...
+
+    // Test parsing hexadecimal format numeric strings.
+    dr_fprintf(STDERR, "param xi = %d\n", op_xi.get_value());
+    dr_fprintf(STDERR, "param xl = %ld\n", op_xl.get_value());
+    dr_fprintf(STDERR, "param xll = %lld\n", op_xll.get_value());
+    dr_fprintf(STDERR, "param xu = %u\n", op_xu.get_value());
+    dr_fprintf(STDERR, "param xul = %lu\n", op_xul.get_value());
+    dr_fprintf(STDERR, "param xull = %llu\n", op_xull.get_value());
+
+    // Test parsing a string of negetive digits. These tests are used to check whether the
```
abhinav92003, line 270: s/negetive/negative/

dolanzhao: Done.

ext/droption/droption.h
```diff
@@ -680,24 +683,25 @@ inline bool
...
     return errno == 0;
 }
 template <>
 inline bool
 droption_t<long long>::convert_from_string(const std::string s)
 {
     errno = 0;
-    value_ = strtoll(s.c_str(), NULL, 10);
+    // If we set 0 as the base, strtoll() will automatically identify the base like
+    // strtol().
```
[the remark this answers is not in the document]
dolanzhao: Done.

ext/droption/droption.h
```diff
@@ -710,25 +714,40 @@ template <>
...
 {
-    errno = 0;
-    long input = strtol(s.c_str(), NULL, 10);
-    if (input >= 0)
-        value_ = (unsigned long)input;
-    else
-        return false;
+    // Checks if the first non-space character of a string is a negative sign. If it is,
+    // the input string is invalid.
+    for (size_t i = 0; i < s.size(); i++) {
```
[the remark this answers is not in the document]
dolanzhao: Done.

ext/droption/droption.h
```diff
@@ -570,6 +572,23 @@ template <typename T> class droption_t : public droption_parser_t {
         return true;
     }
 
+    bool
+    is_negative_input(const std::string &s) override
+    {
+        // Checks if the first non-space character of a string is a negative sign.
+        for (size_t i = 0; i < s.size(); i++) {
+            // XXX: The function isspace() only identify ASCII whitespace characters. We
```
abhinav92003, line 580: s/identify/identifies/

ext/droption/droption.h
```diff
@@ -570,6 +572,23 @@ template <typename T> class droption_t : public droption_parser_t {
...
+            // should implement one function that can support non-ASCII whitespace
+            // characters.
```
abhinav92003, line 582: s/characters/characters too/

ext/droption/droption.h
```diff
@@ -570,6 +572,23 @@ template <typename T> class droption_t : public droption_parser_t {
         return true;
     }
 
+    bool
+    is_negative_input(const std::string &s) override
```
abhinav92003, line 576: s/is_negative_input/is_negative/

ext/droption/droption.h
```diff
@@ -371,6 +371,8 @@ class droption_parser_t {
     convert_from_string(const std::string s1, const std::string s2) = 0;
     virtual bool
     clamp_value() = 0;
+    virtual bool
+    is_negative_input(const std::string &s) = 0;
```
abhinav92003, line 375: Does this routine need to be virtual? We can probably make it static, right?

abhinav92003 (approved):

ext/droption/droption.h
dolanzhao: Done.

ext/droption/droption.h
dolanzhao: Done.

ext/droption/droption.h
dolanzhao: Done.

ext/droption/droption.h
dolanzhao: Done.

abhinav92003 (approved):

ext/droption/droption.h
```diff
@@ -570,6 +570,23 @@ template <typename T> class droption_t : public droption_parser_t {
         return true;
     }
 
+    static bool
+    is_negative(const std::string &s)
+    {
+        // Checks if the first non-space character of a string is a negative sign.
+        for (size_t i = 0; i < s.size(); i++) {
+            // XXX: The function isspace() only identifies ASCII whitespace characters. We
```
derekbruening, line 578: isspace() checks for whitespace in the current locale and I believe if you set the locale to something like "en_US.UTF8" it will correctly return true for a unicode non-ascii whitespace -- are you seeing contradictory information?  Or are you saying that even if the locale is "C" or "POSIX" we'd prefer this to recognize a unicode non-ascii space -- but if the world has agreed that these locales should not be aware of non-ascii spaces then I think it's fine if we behave the same.

ext/droption/droption.h
```diff
@@ -665,7 +682,10 @@ inline bool
 droption_t<int>::convert_from_string(const std::string s)
 {
     errno = 0;
-    long input = strtol(s.c_str(), NULL, 10);
+    // If we set 0 as the base, strtol() will automatically identify the base of the
+    // number to convert. By default, it will assume the number to be converted is
+    // decimal, and number starting with 0 or 0x is assumed to be octal or hexadecimal.
```
derekbruening, line 687: grammar nit: s/number/a number/

derekbruening, line 687: nit: s/./, respectively./

derekbruening (approved):

ext/droption/droption.h
dolanzhao: Done.

dolanzhao: Done.

ext/droption/droption.h
dolanzhao: Yes, the comment is contradictory. After I discussed this with @abhinav92003, he advised me to see the source code of strtoxx() and how it skips whitespaces. I found there are two solutions: 
(1) gcc uses ISSPACE() which is defined in [safe-ctype.h](https://github.com/gcc-mirror/gcc/blob/976196b3332d4f56248e7a1401e5e373501598e4/include/safe-ctype.h#L98) to skips only ASCII whitespace. 
(2) For [glibc](https://github.com/bminor/glibc/blob/a1bdd81664aa681364da368154c48501db249df9/stdlib/strtol_l.c#L150), [Apple libc](https://opensource.apple.com/source/Libc/Libc-167/stdlib.subproj/strtoul.c.auto.html), and [Android libc](https://android.googlesource.com/platform/bionic/+/a27d2baa/libc/stdlib/strtoul.c) directly use isspace() to identify whitespace and ignore the locale. 
And I didn't find how windows' strtoxx() skip whitespaces.

Do you agree that we keep the current implementation and add a comment that we ignore locale effects?

derekbruening: In general the glibc routines, including isspace(), look at the locale, so the locale is not being ignored: so I don't follow your references to it ignoring the locale.  The issue is that isspace() takes as input just an unsigned char and so it can't handle a UTF-8 multi-byte char.  We would have to convert to UTF-16 and use iswspace() or std::isspace on wchar_t.
(Or maybe std::skipws handles UTF-8.)  This is more complex than I originally thought: while we have the code in DR to do the conversion it does not seem worth the effort at this point.  So yes I agree with keeping the current implementation, but please update the comment as this does not seem to be a locale issue but instead is the requirement that we convert to wchar_t to check one character.

dolanzhao: OK

ext/droption/droption.h
```diff
@@ -47,6 +47,7 @@
 #include <limits.h>
```
abhinav92003, line 47: For the PR title, let's try to shorten it:

How about: "i#5505 kernel trace: Support octal and hex values in droption"

We can move other details to the PR description.

abhinav92003 (approved):
Thanks for making the suggested changes!

ext/droption/droption.h
dolanzhao: Done.

PR #5555, opened 2022-07-06
i#5505 kernel trace: Return trace as instrlist_t in drpt2ir

dolanzhao:
Updates library drpt2ir to support converting libipt's IR format to DynamoRIO's IR format(instr_t).
And updates the command line client drpt2trace to invoke the new conversion function of drpt2ir and add disassembly printing in the status output function.
Also, updates the test case to verify the new features added in drpt2ir and drpt2trace.

Issue: #5505


clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -294,21 +300,26 @@ main(int argc, const char *argv[])
...
         return FAILURE;
     }
-    if (ptconverter->convert() != PT2IR_CONV_SUCCESS) {
-        std::cerr << CLIENT_NAME << ": failed to convert pt raw data to DR IR."
+    instrlist_t *ilist = nullptr;
+    pt2ir_convert_status_t status = ptconverter->convert(&ilist);
+    if (status != PT2IR_CONV_SUCCESS) {
+        std::cerr << CLIENT_NAME << ": failed to convert PT raw trace to DR IR."
                   << std::endl;
+        instrlist_clear_and_destroy(GLOBAL_DCONTEXT, ilist);
```
abhinav92003, line 314: In case of failure, it's better if the convert routine itself destroys the ilist.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -191,10 +191,16 @@ static droption_t<unsigned long long> op_sb_kernel_start(
  */
 
 static void
-print_stats(IN pt2ir_t *pt_converter)
+print_stats(IN instrlist_t *ilist)
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -191,10 +191,16 @@ static droption_t<unsigned long long> op_sb_kernel_start(
...
 {
-    std::cout << "Number of Instructions: " << pt_converter->get_instr_count()
-              << std::endl;
+    instr_t *instr = instrlist_first(ilist);
+    uint64_t count = 0;
+    while (instr != NULL) {
+        count++;
+        instr = instr_get_next(instr);
+    }
+    instrlist_disassemble(GLOBAL_DCONTEXT, 0, ilist, STDOUT);
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
dolanzhao: Done.

clients/drcachesim/drpt2trace/test_simple.expect
```diff
@@ -0,0 +1,10 @@
+TAG  0x0000000000000000
+ +0    L2                      ba 0e 00 00 00       mov    $0x0000000e -> %edx
```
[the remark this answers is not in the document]
dolanzhao: Rewrite an X64 version asm code.

clients/drcachesim/drpt2trace/pt2ir.h
```diff
@@ -240,23 +246,12 @@ class pt2ir_t {
      * Returns pt2ir_convert_status_t.
      * \note The convert function performs two processes: (1) decode the PT raw trace into
      * libipt's IR format pt_insn; (2) convert pt_insn into the DynamoRIO's IR format
-     * instr_t. If the convertion is successful, the function returns PT2IR_CONV_SUCCESS.
-     * Otherwise, the function returns the corresponding error code.
+     * instr_t and append it to ilist. If the convertion is successful, the function
+     * returns PT2IR_CONV_SUCCESS. Otherwise, the function returns the corresponding error
+     * code.
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -294,21 +304,23 @@ main(int argc, const char *argv[])
...
         std::cerr << CLIENT_NAME << ": failed to initialize pt2ir_t." << std::endl;
         return FAILURE;
     }
-    if (ptconverter->convert() != PT2IR_CONV_SUCCESS) {
-        std::cerr << CLIENT_NAME << ": failed to convert pt raw data to DR IR."
+    instrlist_t *ilist = nullptr;
+    pt2ir_convert_status_t status = ptconverter->convert(&ilist);
+    if (status != PT2IR_CONV_SUCCESS) {
+        std::cerr << CLIENT_NAME << ": failed to convert PT raw trace to DR IR."
                   << std::endl;
```
abhinav92003, line 317: Also print the status code.

clients/drcachesim/drpt2trace/pt2ir.cpp
```diff
@@ -284,30 +289,23 @@ pt2ir_t::convert()
                 return PT2IR_CONV_ERROR_DECODE_NEXT_INSTR;
```
abhinav92003, line 289: We need to do `instrlist_clear_and_destroy` for each of these error cases, right?

clients/drcachesim/drpt2trace/pt2ir.h
```diff
@@ -237,26 +243,17 @@ class pt2ir_t {
...
-     * instr_t. If the convertion is successful, the function returns PT2IR_CONV_SUCCESS.
-     * Otherwise, the function returns the corresponding error code.
+     * Returns pt2ir_convert_status_t. If the convertion is successful, the function
+     * returns PT2IR_CONV_SUCCESS. Otherwise, the function returns the corresponding error
+     * code.
+     * \note The convert function performs two processes:
+     * (1) decode the PT raw trace into libipt's IR format pt_insn;
+     * (2) convert pt_insn into the DynamoRIO's IR format instr_t and append it to ilist.
+     * \note The caller does not need to initialize ilist. But if the convertion is
+     * successful, the caller need to destory the ilist.
```
abhinav92003, line 253: s/need to destory/needs to destroy/

abhinav92003 (approved):

clients/drcachesim/drpt2trace/pt2ir.cpp
dolanzhao: Done.

clients/drcachesim/drpt2trace/pt2ir.h
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -64,9 +64,9 @@
 /* A collection of options. */
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
dolanzhao: Done.

clients/drcachesim/drpt2trace/pt2ir.cpp
```diff
@@ -281,33 +290,27 @@ pt2ir_t::convert()
...
-            pt_insn_list_.push_back(insn);
-
-            /* TODO i#5505: Use drdecode to decode insn(pt_insn) to instr_t. */
+            /* Use drdecode to decode insn(pt_insn) to instr_t. */
+            instr_t *instr = instr_create(GLOBAL_DCONTEXT);
+            instr_init(GLOBAL_DCONTEXT, instr);
+            decode(GLOBAL_DCONTEXT, insn.raw, instr);
+            instr_set_translation(instr, (app_pc)insn.ip);
+            instr_allocate_raw_bits(GLOBAL_DCONTEXT, instr, insn.size);
+            if (instr_get_opcode(instr) == OP_INVALID) {
```
abhinav92003, line 303: We can use instr_valid instead.

abhinav92003 (approved):

abhinav92003:
run arm tests

clients/drcachesim/drpt2trace/pt2ir.cpp
derekbruening: Drive-by too-late comment, but for future cases: the C++ way would be to create a local object whose destruction will call instrlist_clear_and_destroy() and let the compiler auto-insert the call to it on every return path.

abhinav92003: Do you mean something like `absl::Cleanup`? Is that an equivalent construct available in DR? Also, we want to _not_ destroy the instrlist in the happy-case.

derekbruening: > Do you mean something like absl::Cleanup? Is that an equivalent construct available in DR?

Yes, like absl::Cleanup; no there is no provided class but is simple to make a local struct/class (similar things are done in other places like https://github.com/DynamoRIO/dynamorio/blob/master/clients/drcachesim/tools/opcode_mix.h#L98).

>  Also, we want to not destroy the instrlist in the happy-case.

Oh ok, did not realize this: would take some more logic there so not as clearly a win but can still avoid common errors of forgetting to destroy on some future-added error return.

Issue #5167, opened 2021-10-15
x86 linux.signal0001 test failed with output ordering differences


derekbruening:
Looks like non-deterministic output ordering.  I don't think we've seen this before which is weird.

https://github.com/DynamoRIO/dynamorio/pull/5164/checks?check_run_id=3910404056
```
2021-10-15T21:06:08.1963378Z  46/356 Test  #45: code_api|linux.signal0001 ........................................***Failed  Required regular expression not found. Regex=[^Sending SIGUSR2
2021-10-15T21:06:08.1964069Z Sending SIGUSR1
2021-10-15T21:06:08.1964433Z in signal handler
2021-10-15T21:06:08.1964764Z Got SIGUSR1
2021-10-15T21:06:08.1965115Z Sending SIGRTMAX
2021-10-15T21:06:08.1965482Z in signal handler
2021-10-15T21:06:08.1965818Z Got SIGRTMAX
2021-10-15T21:06:08.1966182Z Generating SIGSEGV
2021-10-15T21:06:08.1966568Z in signal handler
2021-10-15T21:06:08.1966898Z Got SIGSEGV
2021-10-15T21:06:08.1967201Z 1750007\.722714
2021-10-15T21:06:08.1967531Z Got some timer hits!
2021-10-15T21:06:08.1967852Z $
2021-10-15T21:06:08.1968118Z ]  0.56 sec
2021-10-15T21:06:08.1968444Z Sending SIGUSR2
2021-10-15T21:06:08.1968805Z in signal handler
2021-10-15T21:06:08.1969138Z Got SIGUSR1
2021-10-15T21:06:08.1969483Z Sending SIGRTMAX
2021-10-15T21:06:08.1969851Z in signal handler
2021-10-15T21:06:08.1970185Z Got SIGRTMAX
2021-10-15T21:06:08.1970549Z Generating SIGSEGV
2021-10-15T21:06:08.1970929Z in signal handler
2021-10-15T21:06:08.1971269Z Got SIGSEGV
2021-10-15T21:06:08.1971558Z 1750007.722714
2021-10-15T21:06:08.1971887Z Got some timer hits!
```

derekbruening:
Just had a failure on arm-cross-compile but this time it's missing the numeric result:
https://github.com/DynamoRIO/dynamorio/runs/7241572690?check_suite_focus=true
```
[...]

PR #5556, opened 2022-07-07
i#5554 CRASH: DR ELF code does not check for NULL soname fix

jweezy24:
Updated module elf file to catch when pd->soname is NULL.
Bug in issue i#5554 is no longer there.


core/unix/module_elf.c
```diff
@@ -1528,7 +1535,7 @@ module_relocate_symbol(ELF_REL_TYPE *rel, os_privmod_data_t *pd, bool is_rela)
 
     res = module_lookup_symbol(sym, pd);
     LOG(GLOBAL, LOG_LOADER, 3, "symbol lookup for %s %p\n", name, res);
-    if (res == NULL && ELF_ST_BIND(sym->st_info) != STB_WEAK) {
+    if (res == NULL && ELF_ST_BIND(sym->st_info) != STB_WEAK && pd->soname != NULL) {
```
derekbruening, line 1538: We would still want a warning about undefined symbols, even if it isn't able to give a library name (I guess we don't have access to file_name here), so I would suggest entering this if() and just using `pd->soname == NULL ? "<empty soname>" : pd->soname` or something in the SYSLOG.

derekbruening (commented):
Thank you for the PR and for testing it on your test case.
I just have one suggestion.

derekbruening (approved):
Thanks for the quick response.  I'll merge this in.

derekbruening:
Hmm there is a test failure: it seems to be a flake #5167.

Issue #5475, opened 2022-04-27
Add feature name (FEAT_) support to AArch64 instructions


AssadHashmi:
AArch64 instructions are characterised by a feature name starting with `FEAT_`, see:
https://developer.arm.com/downloads/-/exploration-tools/feature-names-for-a-profile

Implementations of the AArch64 architecture are more accurately described by `FEAT_` feature names, than by vX.Y versions, see the explanation at https://github.com/DynamoRIO/dynamorio/pull/5334#issuecomment-1033999742

This issue will clarify how `FEAT_` should be implemented. The following is proposed:
- A new enum `dr_isa_feat_t` similar to `dr_isa_mode_t` but for AArch64 only.
- API accessor functions `dr_get_isa_feat()` and `instr_set_isa_feat()`.
- The AArch64 codec will specify the `FEAT_` for all instructions after base v8.0.
- The system registers `ID_AA64ISAR0_EL1`, `ID_AA64ISAR1_EL1` and `ID_AA64PFR0_EL1` will be read on DynamoRIO startup in order to establish which `FEAT_`s are implemented by the h/w. During instruction decode and encode, the instruction's `FEAT_` will be checked against the `FEAT_` list implemented by the h/w, delivering an error message if the `FEAT_` is not implemented.

derekbruening:
For improved unsupported/invalid instruction support, xref #431 and #1000

For the features: today we have `proc_has_feature()` and the `feature_bit_t` enum of `FEATURE_` values https://dynamorio.org/dr__proc_8h.html#a5e226019d68bd13050692da392f6c4bf.  What would be the relation between that and the AArch64 features?  Should the AArch64 ones be `FEATURE_` instead of `FEAT_` to match?  Would `proc_has_feature()` be deprecated and `dr_isa_feat_t` take its place on x86 too?

There is no x86 connection today between those features and the IR due to early decisions not to include that data in the x86 decode table, which led to the drcpusim tool on x86 using helper functions like instr_is_sse2() which has its own list of opcodes outside of the table.  It might be nice to go back and add feature info into the table eventually, so ideally new interfaces we add should be able to apply to x86 without change.  (It sounds like that is the case other than the relationship to the existing interface.)

AssadHashmi:
> What would be the relation between that and the AArch64 features?

[...]

PR #5557, opened 2022-07-08
i#5475 Add feature checks to AArch64 encoder

AssadHashmi:
This patch implements the codec side of the AArch64 FEATURE_ support
patch, PR5491. A feature name field is added to each opcode in
codec_*.txt files which is used to generate feature checks in the
encoder, e.g.
```
    switch (instr->opcode) {
    case OP_cas:
 #     if !defined(DR_HOST_NOT_TARGET) && !defined(STANDALONE_DECODER)
       if (!proc_has_feature(FEATURE_LSE))
           return ENCFAIL;
 #     endif
       enc = encode_opndsgen_08a07c00_001f03ff(pc, instr,..., di);
       if (enc != ENCFAIL)
           return enc;
```

Issues: #5475, #1569, #2626


core/arch/aarch64/proc.c
```diff
@@ -134,6 +134,7 @@ proc_init_arch(void)
             cpu_info.features.flags_aa64pfr0);
```
derekbruening, line 134: > patch, PR5491. A feature name field is added to each opcode in

`s/PR5491/PR #5491/` so it is auto-linked when viewed on Github.

core/ir/aarch64/codec.py
```diff
@@ -335,14 +342,15 @@ def generate_opndset_encoders(opndsettab, opndtab):
...
-def generate_encoder(patterns, opndsettab, opndtab, curr_isa, next_isa):
+def generate_encoder(patterns, opndsettab, opndtab, opc_props, curr_isa, next_isa):
     c = []
     case = dict()
     for p in patterns:
-        (b, m, mn, f) = p
-        if not mn in case:
-            case[mn] = []
-        case[mn].append(p)
+        (opcode_bits, opnd_bits, opcode, opndset) = p
```
derekbruening, line 349: (These names are much easier to understand than "b, m, mn, f".)

core/ir/aarch64/codec.py
```diff
@@ -532,25 +545,25 @@ def read_codec_file(path):
             for line in (l.split('#')[0].strip() for l in file):
                 if not line:
                     continue
-
-                if re.match('^[01x\^]{32} +[n|r|w|rw|wr|er|ew]+ +[0-9]+ +[a-zA-Z_0-9][a-zA-Z_0-9 ]*:[a-zA-Z_0-9 ]*$', line):
+                #print('{}'.format(line))
```
derekbruening, line 548: Looks like debugging code accidentally left in.

derekbruening (approved):

Issue #5529, opened 2022-06-14
BUILD: Build fails with doxygen issues on AArch64 ubuntu 18.04


joshua-warburton:
Building dynamorio fails during documentation generation on AArch64 ubuntu 18.04 with the following error
```
CMake Error at ./dynamorio/api/docs/CMake_rundoxygen.cmake:154 (message):
  *** /usr/bin/doxygen failed: ***

  ./dynamorio-build/include/dr_tools.h:2691: warning:
  expected whitespace after [ command

  ./dynamorio-build/include/dr_tools.h:2667: warning:
  expected whitespace after [ command

  ./dynamorio-build/include/dr_tools.h:2635: warning:
  expected whitespace after [ command

  ./dynamorio-build/include/dr_tools.h:2657: warning:
  expected whitespace after [ command

  ./dynamorio-build/include/dr_tools.h:2677: warning:
  expected whitespace after [ command

  ./dynamorio-build/include/dr_tools.h:2703: warning:
  expected whitespace after [ command



api/docs/CMakeFiles/htmldocs.dir/build.make:488: recipe for target 'api/docs/html/index.html' failed
```
This is with the default version of doxygen, 1.8.3.

If doxygen is updated to the latest version, we get instead:
```
CMake Error at ./dynamorio/api/docs/CMake_doxyfile.cmake:188 (message):
  /usr/local/bin/doxygen -u failed: This tag has been removed.
```

abhinav92003:
Thanks for reporting the issue. Can you also send a PR for the fix?

PR #5559, opened 2022-07-11
i#5529: Fix incompatible doxygen syntax

joshua-warburton:
Some of the syntax in dr_tools.h was in the format
    @param[name]  desc
which can cause issues with certain versions of doxygen.
This patch changes it to the format
    @param[in|out] name desc
which is widely supported

Fixes: #5529


AssadHashmi (approved):

joshua-warburton:
run arm tests

Issue #3544, opened 2019-04-15
What about Risc-V support future ?


eix128:
Hi , as we all know , Risc-V popularity is rising more in time.
Will Any support be avalible for Risc-V ?

There is already new soc released for risc-v

https://www.sifive.com/boards

AssadHashmi:
Hi, we don't currently have plans for RISC-V support.
However, we do support AArch32 and AArch64 which can be used as reference points for a RISC-V port:
https://github.com/DynamoRIO/dynamorio/wiki/ARM-Port
https://github.com/DynamoRIO/dynamorio/wiki/AArch64-Port

We are a small developer team but welcome and will support anyone or any group who want to add features such as RISC-V. Do you have specific use-cases in mind, i.e. what sort of problems would you hope to address with DynamoRIO if RISC-V was available?

AssadHashmi:
As an aside, I would suggest posting your question on the users list https://groups.google.com/forum/#!forum/DynamoRIO-User

This will be seen by more people than this issues list and we'll get an idea of how many are interested in RISC-V support.

lazyparser:
Hi,

Just FYI that the PLCT Lab is planning to port the DynamoRIO to RISC-V (RV64G)[1]. Not started yet, but we are optimistic that we are able to finish the porting project and contributing back to upstream before the end of 2021.

[1] https://github.com/plctlab/plctlab.github.io/blob/main/PLCT-Roadmap-2021.md#enable-dynamorio-running-on-rv64gc

derekbruening:
That's exciting news!

I have a few suggestions:
+ Please consider a workflow where code changes are sent for review early and often, keeping reviews small and incremental.  This makes the review process much more manageable.  (What we would like to avoid is building up a very large parallel forked code base with no review or merge until it is too large to practically review.)  Initially, existing project members would perform reviews, but once PLCT developers are up to speed they can be added as project members and perform reviews.
+ Naturally we would like to have automated testing of RISC-V added to the Continuous Integration suite: https://github.com/DynamoRIO/dynamorio/wiki/Test-Suite#automated-test-machines.  This could be via Jenkins on a machine hosted somewhere else, or through a Github Actions-provided VM if that can be configured properly.

lazyparser:
Hi @derekbruening 

Thanks for the suggestions. We will send patch set for early review/comments once we got a basic framework. The process of porting is also open (on github, as a fork of the repo).

[...]

PR #5561, opened 2022-07-13
i#3544 RFC: Add basic RISC-V definitions

semihalf-kardach-stanislaw:
This is a first of a set of pull requests whose aim is to enable DynamoRIO compilation for RISC-V architecture.

This change introduces basic architecture specific macro, enum and type definitions that next commits will use.
The first commit is separated out of convenience to highlight a potential problem with handling register references in the generic code with multiple architectures. It also highlights an approach which might help with that.
The second commit contains the meat of the actual definitions added following (up to my understanding) the existing ARM/Aarch64 approach.
This code is not compiled yet (see [my branch](https://github.com/Semihalf/dynamorio/tree/experimental-riscv) for a full compilation codebase reference) and should not impact the current X86 and ARM/Aarch32 functionality.

The "RFC" goal of this pull request is to asses whether this approach is proper or maybe a refactoring is required to better accommodate 3 coexisting architectures.

As noted in [my comment](https://github.com/DynamoRIO/dynamorio/issues/3544#issuecomment-1176782540) in the RISC-V feature request issue, this code is not functional and the goal is to first achieve full compilation first and then work on implementing the functionality.


semihalf-kardach-stanislaw:
@bekcpear, @lazyparser, @vflouris FYI - I'd love to get Your review too.

derekbruening:
add to allowlist

derekbruening:
run arm tests

core/arch/proc_api.h
```diff
@@ -175,6 +180,17 @@ typedef struct {
...
 #endif
+#ifdef RISCV64
+/* FIXME-RISCV: Not implemented */
+/**
+ * For RISC-V64 there are no features readable from userspace. Hence only a
+ * dummy flag is there. May be replaced by actual feature flags in the future.
+ * Used by proc_get_all_feature_bits().
+ */
+typedef struct {
+    uint64 dummy; /**< dummy member to keep size non-0. */
```
derekbruening, line 191: nit: Prefer capitalized sentences in comments, in particular in user-facing Doxygen comments: `s/dummy/Dummy/`.

core/ir/encode_api.h
```diff
@@ -50,6 +50,7 @@ typedef enum _dr_isa_mode_t {
     DR_ISA_ARM_A32,           /**< ARM A32 (AArch32 ARM). */
     DR_ISA_ARM_THUMB,         /**< Thumb (ARM T32). */
     DR_ISA_ARM_A64,           /**< ARM A64 (AArch64). */
+    DR_ISA_RV64IMAFDC,        /**< RISC-V (rv64imafdc). */
```
derekbruening, line 53: Does this mean DR would not support an arch missing one of the letters?  DR itself probably doesn't care about the floating-point support, so could it support r64ima?  Or is it implied that r64imafdc includes r64ima?

core/ir/opnd.h
```diff
@@ -99,7 +99,11 @@ extern const reg_id_t dr_reg_fixer[];
 /* We only normally use r0-r5 but we support more in translation code */
 #    define REG_START_SPILL DR_REG_R0
 #    define REG_STOP_SPILL DR_REG_R10 /* r10 might be used in syscall mangling */
-#endif                                /* X86/ARM */
+#elif defined(RISCV64)
+/* We only normally use r0-r5 but we support more in translation code */
+#    define REG_START_SPILL DR_REG_A0
```
derekbruening, line 104: Wait -- this does not match the spill regs in arch.h!

core/ir/opnd.h
```diff
@@ -280,8 +284,37 @@ enum {
...
+    DR_SYSNUM_REG = DR_REG_A7,
+    REGPARM_0 = DR_REG_A0,
+    REGPARM_1 = DR_REG_A1,
+    REGPARM_2 = DR_REG_A2,
+    REGPARM_3 = DR_REG_A3,
+    REGPARM_4 = DR_REG_A4,
+    REGPARM_5 = DR_REG_A5,
+    REGPARM_6 = DR_REG_A6,
+    REGPARM_7 = DR_REG_A7,
+    NUM_REGPARM = 6,
```
derekbruening, line 297: But if the params go 0..7 isn't that 8?

core/ir/opnd_api.h
```diff
@@ -1124,8 +1124,166 @@ enum {
     /** Platform-independent way to refer to stack pointer. */
     DR_REG_XSP = DR_REG_SP,
 #    endif
+#elif defined(RISCV64)
+    DR_REG_INVALID, /**< Sentinel value indicating an invalid register. */
+    DR_REG_X0,
```
derekbruening, line 1129: We want Doxygen comments for each of these.

core/ir/opnd_api.h
```diff
@@ -1124,8 +1124,166 @@ enum {
...
+    DR_REG_FT8 = DR_REG_F28,
+    DR_REG_FT9 = DR_REG_F29,
+    DR_REG_FT10 = DR_REG_F30,
+    DR_REG_FT11 = DR_REG_F31,
+#    endif /* RISCV_ISA_F || RISCV_ISA_D */
+/* FIXME-RISCV: CCSRs */
+/* FIXME-RISCV: Ifdefs for different extensions */
+
+#    if defined(RISCV_ISA_F) || defined(RISCV_ISA_D)
+    DR_REG_LAST_VALID_ENUM = DR_REG_FCSR, /**< Last valid register enum */
```
derekbruening, line 1268: I know some existing code doesn't do this everywhere but we'd like to have end punctuation in new comments so if you could add a trailing `.` to all of these Doxygen comments.

core/lib/mcxtx_api.h
```diff
@@ -272,4 +272,287 @@
...
+        reg_t ra;
+    };
+    union {
+        reg_t x2;  /**< The x2 register. */
+        reg_t sp;  /**< The stack pointer register.*/
+        reg_t xsp; /**< The platform-independent name for the stack pointer register. */
+    };
+    union {
+        reg_t x3;   /**< The x3 register. */
+        reg_t gp;
```
derekbruening, line 295: I assume we want a Doxygen comment for these alternates.

core/unix/os_exports.h
```diff
@@ -105,7 +105,11 @@
                              */
 #        define LIB_SEG_TLS DR_REG_TPIDRURO /* libc+loader tls */
 #    endif                                  /* 64/32-bit */
-#endif                                      /* X86/ARM */
+#elif defined(RISCV64)
+/* FIXME-RISCV: Check if this is preserved by kernel and safe to use? */
+#    define SEG_TLS DR_REG_S11
```
derekbruening, line 110: See above: for a GPR we'd want it option-controlled and not hardcoded, like we do for AArchXX.

core/unix/os_exports.h
```diff
@@ -295,7 +312,14 @@ disable_env(const char *name);
 #        define DECLARE_DATA_SECTION(name, wx)     \
             asm(".section " name ", \"a" wx "\""); \
             asm(".align 12"); /* 2^12 */
-#    endif                    /* X86/ARM */
+#    elif defined(DR_HOST_RISCV64)
+/* FIXME-RISCV: RFC: Why all archs align those sections to 4K? Page size?
```
derekbruening, line 316: DR does change page protections for different data regions at runtime -- looking at this I wonder if that's broken for >4K pages.  It is only on by default on Windows maybe IIRC.

derekbruening (commented):
Thank you for the PR.  Mostly minor comments.

core/ir/encode_api.h
semihalf-kardach-stanislaw: rv64imafdc includes rv64ima.

However given the flexible and extensible nature of RISC-V spec it could be theoretically possible to create rv64i ISA with ANY combination of other extensions.

Honestly though I'm not sure yet how to handle possible subsets of ISA extensions in this enum.

That said, the spec defines IMAFD as the standard extension for a general purpose CPU. I'm planning for "C" (compressed) because adding it later on may be more work.

core/arch/arch_exports.h
```diff
@@ -1028,7 +1046,37 @@ fill_with_nops(dr_isa_mode_t isa_mode, byte *addr, size_t size);
...
+#    define CTI_PATCH_OFFSET 4
+/* offset of the patchable region from the end of a stub */
+/* FIXME-RISCV: Not implemented */
+#    define EXIT_STUB_PATCH_OFFSET 4
+/* size of the patch to a stub */
+/* FIXME-RISCV: Not implemented */
+#    define EXIT_STUB_PATCH_SIZE 4
+/* the most bytes we'll need to shift a patchable location for -pad_jmps */
+/* FIXME-RISCV: Not implemented */
+#    define MAX_PAD_SIZE 0
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Around 50% of instructions can be compressed when CPU implements C extension. GCC generates mixes the code using compressed and uncompressed instructions (also valid). In theory nothing stops the assembler to have a 4-byte instruction cross the cache line. I haven't verified if GCC guards against that.

core/arch/arch_exports.h
```diff
@@ -1028,7 +1046,37 @@ fill_with_nops(dr_isa_mode_t isa_mode, byte *addr, size_t size);
 /* i#1906: alignment needed for the source address of data to load into the PC */
 #    define PC_LOAD_ADDR_ALIGN 4
 
-#endif /* ARM */
+#elif defined(RISCV64)
+/* FIXME-RISCV: This can be 2B in C. */
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Yes, when CPU implements C extension and only for ~50% of instructions.

core/arch/asm_defines.asm
```diff
@@ -319,6 +321,44 @@ ASSUME fs:_DATA @N@\
...
+# define REG_R4   x4
+# define REG_R5   x5
+# define REG_R6   x6
+# define REG_R7   x7
+# define REG_R8   x8
+# define REG_R9   x9
+# define REG_R10  x10
+# define REG_R11  x11
+# define REG_R12  x12
+# define REG_R13  x13
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: `x10-x17` are argument registers on RISC-V so I need it for defining `ARG1..ARG8` macros below (lines >=506).

core/arch/asm_defines.asm
```diff
@@ -319,6 +321,44 @@ ASSUME fs:_DATA @N@\
...
+# define REG_R23  x23
+# define REG_R24  x24
+# define REG_R25  x25
+# define REG_R26  x26
+# define REG_R27  x27
+# define REG_R28  x28
+# define REG_R29  x29
+# define REG_R30  x30
+# define REG_R31  x31
+#  define SEG_TLS s11 /* keep in sync w/ {unix,win32}/os_exports.h defines */
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Thanks for spotting this. I must have defined it to push compilation forward and in the end it's not necessary.

core/unix/os_exports.h
semihalf-kardach-stanislaw: I'm not sure how to achieve that given that it's a define. Do you suggest refactoring `TLS_REG_ALT` to be a `reg_t` field in `dcontext_t` filled at initialization time?

core/arch/arch.h
```diff
@@ -151,7 +152,29 @@ mixed_mode_enabled(void)
...
+#    define SCRATCH_REG3 DR_REG_T3
+#    define SCRATCH_REG4 DR_REG_T4
+#    define SCRATCH_REG5 DR_REG_T5
+#    define SCRATCH_REG0_OFFS REG0_OFFSET
+#    define SCRATCH_REG1_OFFS REG1_OFFSET
+#    define SCRATCH_REG2_OFFS REG2_OFFSET
+#    define SCRATCH_REG3_OFFS REG3_OFFSET
+#    define SCRATCH_REG4_OFFS REG4_OFFSET
+#    define SCRATCH_REG5_OFFS REG5_OFFSET
+/* FIXME-RISCV: Check is T6 safe to use */
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: `FIXME i#3544` sounds perfect, I'll change it in v2.

core/arch/arch_exports.h
```diff
@@ -220,6 +223,20 @@ typedef struct _local_state_extended_t {
 #    define SCRATCH_REG4 DR_REG_R4
 #    define SCRATCH_REG5 DR_REG_R5
 #    define SCRATCH_REG_LAST SCRATCH_REG5
+#elif defined(RISCV64)
+/* FIXME-RISCV: RFC: Are there any requirements on TLS/scratch registers?
+ * Aarch64 and X86 use a combination of argument, callee-saved and
+ * caller-saved/temporary regs. Can we use temporary registers or rather leave
+ * them to client code?
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Note to self: I'll use argument registers `aX` (alias to `x10-x17`) since those are consecutive and won't cause confusion as using `x1-x4` would (`x0` is a hard-wired-zero register).

core/unix/os_exports.h
derekbruening: I feel like I need a refresher on what this SEG_TLS is really used for on non-x86: when in the code cache we steal a GPR; when in private lib code we have to use whatever regular lib code uses -- but I think that's LIB_SEG_TLS.  So SEG_TLS is not used at all for AArchXX, and presumably the same would be true for RISC-V?  So maybe just set this to DR_REG_INVALID.

core/unix/os_exports.h
derekbruening: Yes, DR has page size computation and dynamic page size handling everywhere else: but it looks like this is a case where it does not do the right thing; but like I said AFAIK it is not enabled anywhere where it has been hit yet.  I filed #5565 on this.  Please include it in this comment; probably can be downgraded to XXX instead of FIXME.

core/unix/os_exports.h
semihalf-kardach-stanislaw: You might be right. `DR_REG_INVALID` sounds OK for now.

core/arch/arch.h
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Done

core/arch/arch_exports.h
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Done

core/arch/arch_exports.h
```diff
@@ -155,6 +155,9 @@ typedef struct _spill_state_t {
     /* These are needed for ldex/stex mangling and A64 icache_op_ic_ivau_asm. */
     reg_t r4, r5;
     reg_t reg_stolen; /* slot for the stolen register */
+#elif defined(RISCV64)
+    /* FIXME-RISCV: Find proper registers to use as scratch. Does it matter? */
+    reg_t t0, t1, t2, t3;
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Done. Changed to `a0-a3`.

core/ir/opnd_api.h
semihalf-kardach-stanislaw: Done

core/arch/asm_defines.asm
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Done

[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Done

core/arch/proc_api.h
semihalf-kardach-stanislaw: Done.

core/ir/opnd.h
semihalf-kardach-stanislaw: Correct. I have removed the 6 and 7.

core/ir/opnd_api.h
semihalf-kardach-stanislaw: Done.

core/lib/mcxtx_api.h
semihalf-kardach-stanislaw: Done.

core/unix/os_exports.h
semihalf-kardach-stanislaw: Done.

core/unix/os_exports.h
semihalf-kardach-stanislaw: Done.

derekbruening (approved):
LGTM

derekbruening:
If you could merge the latest DynamoRIO:master I can merge once the tests re-run.

derekbruening:
Thanks again for the pull request and fixes.  I've merged it with a final message I hope you agree with.  I've invited you to the project such that you could potentially do future squash-and-merges (please see https://dynamorio.org/page_code_reviews.html#autotoc_md116 as editing the final message is always confusing the first time with github squash-and-merge for the title vs the body) and use branches in the dynamorio repo.

semihalf-kardach-stanislaw:
Thanks a lot! I'll post further pull requests from this repo.

derekbruening:
It was implied, but explicitly responding to the "semantic register names" RFC: yes that approach seems good to me.

Issue #2103, opened 2016-12-11
add decoder support for SMAP instructions STAC and CLAC


derekbruening:
Haswell has a new Supervisor Mode Access Prevention (SMAP) feature which
introduces two new instructions, STAC and CLAC, to set and clear a bit in
CR4.

dolanzhao:
(#5505) The kernel instruction traces contain STAC and CLAC instructions. Not sure if ignoring these two types of instructions will affect synthetic kernel memory addresses.

PR #5562, opened 2022-07-13
i#5505 drpt2ir: Support mode without sideband information

dolanzhao:
Update library drpt2ir to support setting the elf image for the whole PT trace.
And updates the command line client drpt2trace to support two conversion modes:
(1) Sideband Mode: the user must provide sideband data and parameters. In this mode, the converter uses sideband decoders to simulate image switches during the conversion. For example, we can use this mode to convert the traces where the instruction bytes are located in multiple images.
(2) Elf Mode: the user needs to provide an elf image for the PT trace. This mode is for cases where the PT trace's instruction bytes belong to one image. So we can use this mode to convert the kernel trace and the short-term user trace where it's likely that we'll not have an image switch.
Also, adds a new test case for the elf mode.

Issue: #5505


clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -74,9 +74,18 @@ static droption_t<std::string>
               "Specifies the file path of the PT raw trace. Please run the "
```
abhinav92003, line 74: Some more details would help in the PR description. Specifically, that the sideband contains events that are relevant only when we have multiple modules, that for kernel traces we expect to have only a single one, and that for user traces we may have multiple if the trace is long enough.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -74,9 +74,18 @@ static droption_t<std::string>
...
               "libipt/script/perf-read-aux.bash script to get PT raw trace file from the "
               "data generated by the perf record command.");
+static droption_t<std::string>
+    op_elf(DROPTION_SCOPE_FRONTEND, "elf", "", "[Optional] Path of the elf file",
```
abhinav92003, line 78: s/Path of/Path to/
same below

s/You must specify/This must be specified/
s/converting short traces/converting traces/
add e.g. kernel-only traces and short user traces.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -74,9 +74,18 @@ static droption_t<std::string>
...
+           "Specifies the file path of the elf file. You must specify when "
+           "converting short traces that don't contain sideband information.");
+static droption_t<unsigned long long>
+    op_elf_base(DROPTION_SCOPE_FRONTEND, "elf_base", 0x0,
+                "[Optional] The base address of the elf file",
+                "Specifies the base address of the elf file.");
```
abhinav92003, line 84: I believe this must be specified if --elf is specified. If so, add it to the description.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -261,6 +264,29 @@ option_init(int argc, const char *argv[])
             return false;
         }
     }
+
+    if (!op_elf.specified() && !op_primary_sb.specified()) {
+        std::cerr << CLIENT_NAME << ": at least one of the options --elf, --primary_sb."
```
abhinav92003, line 269: Incomplete statement? Add "must be specified" in the end.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -261,6 +264,29 @@ option_init(int argc, const char *argv[])
...
+        return false;
+    }
+
+    if (op_primary_sb.specified()) {
+        std::vector<droption_parser_t *> sb_required_op_list;
+        sb_required_op_list.push_back(&op_sb_sample_type);
+        sb_required_op_list.push_back(&op_sb_time_zero);
+        sb_required_op_list.push_back(&op_sb_time_shift);
+        sb_required_op_list.push_back(&op_sb_time_mult);
+        for (auto &op : sb_required_op_list) {
```
abhinav92003, line 281: We check the required_ops only for the op_primary_sb case? We should do it for the op_elf case too right?

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -261,6 +264,29 @@ option_init(int argc, const char *argv[])
             return false;
         }
     }
+
```
abhinav92003, line 267: It would be helpful to add a comment here to describe the two modes: elf and sideband. Also mention in what cases each of them should be used (kernel and short user traces, vs others)

clients/drcachesim/drpt2trace/pt2ir.cpp
```diff
@@ -300,12 +315,18 @@ pt2ir_t::convert(OUT instrlist_t **ilist)
             decode(GLOBAL_DCONTEXT, insn.raw, instr);
             instr_set_translation(instr, (app_pc)insn.ip);
             instr_allocate_raw_bits(GLOBAL_DCONTEXT, instr, insn.size);
+            instr_set_isa_mode(instr,
+                               insn.mode == ptem_32bit ? DR_ISA_IA32 : DR_ISA_AMD64);
             if (!instr_valid(instr)) {
-                ERRMSG("Failed to convert the libipt's IR to Dynamorio's IR.\n");
-                instrlist_clear_and_destroy(GLOBAL_DCONTEXT, *ilist);
-                return PT2IR_CONV_ERROR_DR_IR_CONVERT;
+                /* XXX:If the PT raw data contains some instructions not supported by
```
abhinav92003, line 321: s/XXX:If/XXX: If/

clients/drcachesim/drpt2trace/pt2ir.cpp
```diff
@@ -300,12 +315,18 @@ pt2ir_t::convert(OUT instrlist_t **ilist)
...
+                 * Dynamorio, we will not append them to the output instrlist_t.
+                 * Currently, the PT raw data may contains 'STAC' and 'CLAC' instructions.
+                 * We'd better to support them in the future.
```
abhinav92003, line 324: If we need to relax the failure strategy, I suggest adding the instr to the list. Users of pt2ir can decide what to do with invalid instrs. This also would prevent silent failures.

Also add a SYSLOG_INTERNAL_WARNING_ONCE to say that invalid instrs were found, and also a LOG to print the raw bytes of each bad instr.

suite/tests/CMakeLists.txt
```diff
@@ -4199,9 +4199,22 @@ if (BUILD_CLIENTS)
...
       "../../clients/drcachesim/drpt2trace/test_simple.expect"
-      "" "${drpt2trace_args}" ON OFF)
+      "" "${drpt2trace_sideband_args}" ON OFF)
+    set(drpt2trace_elf_args "-print_trace"
+      "-raw_pt"
+      "${PROJECT_SOURCE_DIR}/clients/drcachesim/drpt2trace/test_simple.raw/pt.bin"
+      "-pt_cpu_family" "6"
+      "-pt_cpu_model" "85"
+      "-pt_cpu_stepping" "4"
+      "-pt_mtc_freq" "3" "-pt_nom_freq" "37"
```
abhinav92003, line 4211: I see there are some args that are common between the elf and sideband case. How about we define a drpt2trace_common_args instead of duplicating the list?

dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
dolanzhao: The elf_base option is a optional in elf mode. How about I change the optional to ELF Mode Optional?

clients/drcachesim/drpt2trace/drpt2trace.cpp
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
dolanzhao: Do you mean to add these message to PR's description or in drpt2trace.cpp?

clients/drcachesim/drpt2trace/pt2ir.cpp
dolanzhao: About "add a LOG to print the raw bytes of each bad instr", do you mean to add them in pt2ir.cpp or drpt2trace.cpp?

clients/drcachesim/drpt2trace/drpt2trace.cpp
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
abhinav92003: In the PR description.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -74,9 +74,18 @@ static droption_t<std::string>
               "Specifies the file path of the PT raw trace. Please run the "
               "libipt/script/perf-read-aux.bash script to get PT raw trace file from the "
               "data generated by the perf record command.");
+static droption_t<std::string>
+    op_elf(DROPTION_SCOPE_FRONTEND, "elf", "", "[Optional] Path of the elf file",
+           "Specifies the file path of the elf file. You must specify when "
+           "converting short traces that don't contain sideband information.");
+static droption_t<unsigned long long>
+    op_elf_base(DROPTION_SCOPE_FRONTEND, "elf_base", 0x0,
+                "[Optional] The base address of the elf file",
```
derekbruening, line 84: What base is this?  The preferred base hardcoded in the file, or the runtime base at which the file was loaded during tracing?  Please clarify in these docs.  I assume the runtime base since the preferred base is available from the file and should not need to be passed in.  But if it's the runtime base, why is it optional?  Do you assume it's loaded at its preferred base otherwise?  If it's PIC/PIE the preferred base will nearly always be 0 so this would be required there.

clients/drcachesim/drpt2trace/pt2ir.cpp
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -74,9 +74,18 @@ static droption_t<std::string>
...
+                "Specifies the base address of the elf file.");
+
 static droption_t<std::string>
     op_primary_sb(DROPTION_SCOPE_FRONTEND, "primary_sb", "",
-                  "[Required] Path of primary sideband stream file",
+                  "[Optional] Path of primary sideband stream file",
                   "Specifies the file path of the primary sideband stream. A primary "
```
derekbruening: Clarify in the these docs that you either need a sideband file or if there isn't one (which should only happen when just one binary was involved, right?) you have to pass that in as --elf.

clients/drcachesim/drpt2trace/drpt2trace.cpp
dolanzhao: Done.

clients/drcachesim/drpt2trace/pt2ir.cpp
```diff
@@ -300,12 +315,18 @@ pt2ir_t::convert(OUT instrlist_t **ilist)
...
             instr_allocate_raw_bits(GLOBAL_DCONTEXT, instr, insn.size);
+            instr_set_isa_mode(instr,
+                               insn.mode == ptem_32bit ? DR_ISA_IA32 : DR_ISA_AMD64);
             if (!instr_valid(instr)) {
-                ERRMSG("Failed to convert the libipt's IR to Dynamorio's IR.\n");
-                instrlist_clear_and_destroy(GLOBAL_DCONTEXT, *ilist);
-                return PT2IR_CONV_ERROR_DR_IR_CONVERT;
+                /* XXX:If the PT raw data contains some instructions not supported by
+                 * Dynamorio, we will not append them to the output instrlist_t.
+                 * Currently, the PT raw data may contains 'STAC' and 'CLAC' instructions.
```
derekbruening: grammar: contain

clients/drcachesim/drpt2trace/pt2ir.h
```diff
@@ -196,6 +196,16 @@ struct pt2ir_config_t {
      */
     std::string raw_file_path;
 
+    /**
+     * The elf file path.
+     */
+    std::string elf_file_path;
+
+    /**
+     * The base address of the elf file.
```
derekbruening, line 205: Clarify which base -- runtime load address during tracing I assume.

clients/drcachesim/drpt2trace/drpt2trace.cpp
derekbruening: So PT does not bother to produce a sideband file if only one module was encountered, right?  That's why passing the module explicitly is required?  I would make that clear.  It would be nicer and more consistent if it would always produce a sideband file.

derekbruening (approved):

clients/drcachesim/drpt2trace/pt2ir.cpp
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
dolanzhao: Done.

clients/drcachesim/drpt2trace/pt2ir.h
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -70,30 +70,48 @@ static droption_t<bool> op_print_trace(DROPTION_SCOPE_FRONTEND, "print_trace", f
...
     op_raw_pt(DROPTION_SCOPE_FRONTEND, "raw_pt", "",
-              "[Required] Path of PT raw trace file",
-              "Specifies the file path of the PT raw trace. Please run the "
+              "[Required] Path to PT raw trace file",
+              "Specifies the file the path of the PT raw trace. Please run the "
               "libipt/script/perf-read-aux.bash script to get PT raw trace file from the "
               "data generated by the perf record command.");
+static droption_t<std::string>
+    op_elf(DROPTION_SCOPE_FRONTEND, "elf", "", "[Optional] Path to the elf file",
+           "Specifies the file the path of the elf file. This must be specified when "
```
abhinav92003, line 79: s/file the path/file path/

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -261,6 +273,49 @@ option_init(int argc, const char *argv[])
             return false;
         }
     }
+
+    /* Because Intel PT doesn't save instruction bytes or memory contents, the converter
+     * needs the instruction bytes for each IPs (Instruction Pointers) to decode PT.
+     * drpt2trace supports two modes to convert:
+     * (1) Sideband Mode: the user must provide sideband data and parameters. In this
+     * mode, the converter uses sideband decoders to simulate image switches during the
+     * conversation. We can use this mode to convert the traces that the instruction bytes
```
abhinav92003, line 282: where the instruction bytes are located

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -261,6 +273,49 @@ option_init(int argc, const char *argv[])
...
+     * located in multiple images.
+     * (2) Elf Mode: the user needs to provide an elf image for the PT. This mode is for
+     * cases where the PT traces' instruction bytes belong to one image. So we can use
```
abhinav92003, line 285: s/traces'/trace's/

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -261,6 +273,49 @@ option_init(int argc, const char *argv[])
...
+     * this mode to convert the kernel trace and the short-term trace.
```
abhinav92003, line 286: short-term user trace where it's likely that we'll not have an image switch.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -261,6 +273,49 @@ option_init(int argc, const char *argv[])
...
+     */
+    if (!op_elf.specified() && !op_primary_sb.specified()) {
+        std::cerr
+            << CLIENT_NAME
+            << ": at least one of the options --elf, --primary_sb must be specified."
```
abhinav92003, line 291: s/at least one of/exactly one of/

and combine the two conditions into one.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -261,6 +273,49 @@ option_init(int argc, const char *argv[])
...
+    if (op_primary_sb.specified()) {
+        std::vector<droption_parser_t *> sb_required_op_list;
+        sb_required_op_list.push_back(&op_sb_sample_type);
+        sb_required_op_list.push_back(&op_sb_time_zero);
+        sb_required_op_list.push_back(&op_sb_time_shift);
+        sb_required_op_list.push_back(&op_sb_time_mult);
+        for (auto &op : sb_required_op_list) {
+            if (!op->specified()) {
+                std::cerr << CLIENT_NAME << ": option " << op->get_name()
+                          << " is required in sideband model." << std::endl;
```
abhinav92003, line 313: s/model/mode/

clients/drcachesim/drpt2trace/pt2ir.cpp
```diff
@@ -300,11 +315,23 @@ pt2ir_t::convert(OUT instrlist_t **ilist)
             decode(GLOBAL_DCONTEXT, insn.raw, instr);
             instr_set_translation(instr, (app_pc)insn.ip);
             instr_allocate_raw_bits(GLOBAL_DCONTEXT, instr, insn.size);
+            instr_set_isa_mode(instr,
+                               insn.mode == ptem_32bit ? DR_ISA_IA32 : DR_ISA_AMD64);
+#ifdef DEBUG
+            /* XXX: Currently, the PT raw data may contain 'STAC' and 'CLAC'
```
abhinav92003, line 321: reference the actual issue number in a TODO i#NNN

clients/drcachesim/drpt2trace/pt2ir.cpp
```diff
@@ -300,11 +315,23 @@ pt2ir_t::convert(OUT instrlist_t **ilist)
...
+             * instructions that are not supported by Dynamorio. We'd better to
```
abhinav92003, line 322: We can skip " We'd better support them in the future."

clients/drcachesim/drpt2trace/pt2ir.cpp
```diff
@@ -300,11 +315,23 @@ pt2ir_t::convert(OUT instrlist_t **ilist)
...
+             * support them in the future.
+             */
             if (!instr_valid(instr)) {
-                ERRMSG("Failed to convert the libipt's IR to Dynamorio's IR.\n");
-                instrlist_clear_and_destroy(GLOBAL_DCONTEXT, *ilist);
-                return PT2IR_CONV_ERROR_DR_IR_CONVERT;
+                /* We print the invalid instruction‘s PC and raw bytes in DEBUG mode.
+                 */
+                dr_fprintf(STDOUT, "0x%" PRIx64 " <INVALID> <raw ", insn.ip);
```
abhinav92003, line 328: @derekbruening do we want to add this to instr library maybe?

clients/drcachesim/drpt2trace/pt2ir.h
```diff
@@ -293,6 +304,9 @@ class pt2ir_t {
     /* The libipt image section cache. */
     struct pt_image_section_cache *pt_iscache_;
 
+    /* The libipt elf file image. */
+    struct pt_image *pt_image_;
```
abhinav92003, line 308: Maybe make the identifier a bit more descriptive. it is currently (almost) same as the struct.

abhinav92003 (approved):

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -70,30 +70,48 @@ static droption_t<bool> op_print_trace(DROPTION_SCOPE_FRONTEND, "print_trace", f
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -70,30 +70,48 @@ static droption_t<bool> op_print_trace(DROPTION_SCOPE_FRONTEND, "print_trace", f
 
 static droption_t<std::string>
     op_raw_pt(DROPTION_SCOPE_FRONTEND, "raw_pt", "",
-              "[Required] Path of PT raw trace file",
-              "Specifies the file path of the PT raw trace. Please run the "
+              "[Required] Path to PT raw trace file",
+              "Specifies the file the path of the PT raw trace. Please run the "
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -70,30 +70,48 @@ static droption_t<bool> op_print_trace(DROPTION_SCOPE_FRONTEND, "print_trace", f
 
 static droption_t<std::string>
     op_raw_pt(DROPTION_SCOPE_FRONTEND, "raw_pt", "",
-              "[Required] Path of PT raw trace file",
-              "Specifies the file path of the PT raw trace. Please run the "
+              "[Required] Path to PT raw trace file",
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -261,6 +273,49 @@ option_init(int argc, const char *argv[])
             return false;
         }
     }
+
+    /* Because Intel PT doesn't save instruction bytes or memory contents, the converter
+     * needs the instruction bytes for each IPs (Instruction Pointers) to decode PT.
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -261,6 +273,49 @@ option_init(int argc, const char *argv[])
...
+     * drpt2trace supports two modes to convert:
+     * (1) Sideband Mode: the user must provide sideband data and parameters. In this
+     * mode, the converter uses sideband decoders to simulate image switches during the
+     * conversation. We can use this mode to convert the traces that the instruction bytes
+     * located in multiple images.
+     * (2) Elf Mode: the user needs to provide an elf image for the PT. This mode is for
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
dolanzhao: Done.

clients/drcachesim/drpt2trace/pt2ir.cpp
dolanzhao: Done.

clients/drcachesim/drpt2trace/pt2ir.cpp
dolanzhao: Done.

clients/drcachesim/drpt2trace/pt2ir.h
dolanzhao: Remove it.

clients/drcachesim/drpt2trace/pt2ir.cpp
abhinav92003: @dolanzhao Can you check if we can use instr_disassemble instead for this case? It may already print the raw bytes, so we probably don't need custom logic for printing just the raw bytes.

abhinav92003: Maybe we can modify instr_disassemble to print the raw bytes in case of invalid instructions (and then use instr_disassemble here). Seems a useful feature to have in general.

dolanzhao: @derekbruening @abhinav92003 I modified the internal_instr_disassemble() to support printing raw bits for invalid instructions. Could you help me review them?

core/ir/disassemble_shared.c
```diff
@@ -1094,6 +1094,18 @@ internal_instr_disassemble(char *buf, size_t bufsz, size_t *sofar INOUT,
 
     if (!instr_valid(instr)) {
         print_to_buffer(buf, bufsz, sofar, "<INVALID>");
+        if (instr_raw_bits_valid(instr)) {
```
derekbruening, line 1097: I'm confused as to how we'd get into this situation: normally for an unknown encoding we would not have valid raw bits because we would not know the instruction length (except for AArch64).  What use case would end up here?

derekbruening: For debugging, the decoder logs some number of raw bytes when it finds an invalid instruction, but it doesn't know the length and just guesses.  Is that what you're trying to do?  I just don't see how the raw bits are valid for that case: I think the decoder does not set the raw bits on a failed decode.

abhinav92003: For the PT case we would know the exact raw bits, but I see now that that may not be the case in general. Maybe it's better to print the raw bits on failure in the pt decoder itself.

abhinav92003: > Are you hitting valid opcodes that DR's decoder is not aware of? 

Yes we're hitting stac and clac from #2103.

>  If we were to put it here also,

We can skip putting it here. We can move it to drpt2ir instead where we do know the instr length.

>  Are you using drdecode and for that maybe the existing logging is stubbed out

Yes we're using drdecode in the drpt2ir library. It'll be better to have a visible error on stderr for decoding failures.

PR #5568, opened 2022-07-19
i#2626: Aarch64 v8.1 decode: Add ldlar, stllr and variants

joshua-warburton:
This patch adds:
LDLAR   <Wt>, [<Xn|SP>]
LDLAR   <Xt>, [<Xn|SP>]
LDLARB  <Wt>, [<Xn|SP>]
LDLARH  <Wt>, [<Xn|SP>]
STLLR   <Wt>, [<Xn|SP>]
STLLR   <Xt>, [<Xn|SP>]
STLLRB  <Wt>, [<Xn|SP>]
STLLRH  <Wt>, [<Xn|SP>]

It includes the disassembly tests, the IR macros and the IR macro tests

Issues: #2626

PR #5563, opened 2022-07-14
i#2626: AArch64 v8.1 codec: add SQRDMLSH

AssadHashmi:
Completes implementation of signed saturating rounding doubling
multiply arithmetic instructions SQRDMLAH and SQRDMULH.

Issue: #2626

PR #5569, opened 2022-07-19
i#3544 RISC-V, part 2: Add opcode and instr macros

semihalf-kardach-stanislaw:
Add minimal opcode, operand and instruction generation macro definitions
to enable the rest of the code to compile.

NOTE:

Issue: #3544

### DIFF
--- a/core/ir/instr_inline_api.h
+++ b/core/ir/instr_inline_api.h
@@ -183,8 +183,27 @@ opnd_is_far_rel_addr(opnd_t opnd)
 {
     return false;
 }
-#        endif
-#    endif /* X64 || ARM */
+#        elif defined(RISCV64)
+#            define OPND_IS_REL_ADDR(op) ((op).kind == REL_ADDR_kind)
+#            define opnd_is_rel_addr OPND_IS_REL_ADDR
+
+INSTR_INLINE
+bool
+opnd_is_near_rel_addr(opnd_t opnd)
+{
+    /* FIXME i#3544: Not implemented */
+    return opnd_is_rel_addr(opnd);
+}
+
+INSTR_INLINE
+bool
+opnd_is_far_rel_addr(opnd_t opnd)
+{
+    /* FIXME i#3544: Not implemented */
+    return false;
+}
+#        endif /* RISCV64 */
+#    endif     /* X64 || ARM */
 
 /* opnd_t constructors */
 
@@ -271,6 +290,11 @@ opnd_create_pc(app_pc pc)
                  "opnd_get_flags called on non-reg non-base-disp non-immed-int opnd")( \
                  opnd)                                                                 \
                  .aux.flags)
+#    elif defined(RISCV64)
+#        define OPND_GET_FLAGS(opnd)                                                     \
+            (CLIENT_ASSERT_(                                                             \
+                opnd_is_reg(opnd) || opnd_is_base_disp(opnd) || opnd_is_immed_int(opnd), \
+                "opnd_get_flags called on non-reg non-base-disp non-immed-int opnd") 0)
 #    endif
 #    define opnd_get_flags OPND_GET_FLAGS
 
@@ -302,7 +326,7 @@ opnd_create_pc(app_pc pc)
                                                            opnd_is_rel_addr(opnd)),     \
                             "opnd_get_segment called on invalid opnd type")(opnd)       \
                  .aux.segment)
-#    elif defined(AARCHXX)
+#    elif defined(AARCHXX) || defined(RISCV64)
 #        define OPND_GET_SEGMENT(opnd) DR_REG_NULL
 #    endif
 #    define opnd_get_segment OPND_GET_SEGMENT
--- a/core/ir/instr_inline_api.h
+++ b/core/ir/instr_inline_api.h
@@ -191,7 +191,6 @@ INSTR_INLINE
 bool
 opnd_is_near_rel_addr(opnd_t opnd)
 {
-    /* FIXME i#3544: Not implemented */
     return opnd_is_rel_addr(opnd);
 }
 
@@ -199,7 +198,7 @@ INSTR_INLINE
 bool
 opnd_is_far_rel_addr(opnd_t opnd)
 {
-    /* FIXME i#3544: Not implemented */
+    /* FIXME i#3544: Decide if this should differentiate between JAL and AUIPC+JALR. */
     return false;
 }
 #        endif /* RISCV64 */
@@ -277,7 +276,7 @@ opnd_create_pc(app_pc pc)
              .value.reg)
 #    define opnd_get_reg OPND_GET_REG
 
-#    ifdef X86
+#    if defined(X86) || defined(RISCV64)
 #        define OPND_GET_FLAGS(opnd)                                                     \
             (CLIENT_ASSERT_(                                                             \
                 opnd_is_reg(opnd) || opnd_is_base_disp(opnd) || opnd_is_immed_int(opnd), \
@@ -290,11 +289,6 @@ opnd_create_pc(app_pc pc)
                  "opnd_get_flags called on non-reg non-base-disp non-immed-int opnd")( \
                  opnd)                                                                 \
                  .aux.flags)
-#    elif defined(RISCV64)
-#        define OPND_GET_FLAGS(opnd)                                                     \
-            (CLIENT_ASSERT_(                                                             \
-                opnd_is_reg(opnd) || opnd_is_base_disp(opnd) || opnd_is_immed_int(opnd), \
-                "opnd_get_flags called on non-reg non-base-disp non-immed-int opnd") 0)
 #    endif
 #    define opnd_get_flags OPND_GET_FLAGS
 
--- a/core/ir/instr_shared.c
+++ b/core/ir/instr_shared.c
@@ -3605,6 +3605,7 @@ instr_raw_is_tls_spill(byte *pc, reg_id_t reg, ushort offs)
     ASSERT_NOT_IMPLEMENTED(false);
     return false;
 #    elif defined(RISCV64)
+    /* FIXME i#3544: Not implemented. */
     ASSERT_NOT_IMPLEMENTED(false);
     return false;
 #    endif /* X86/ARM/RISCV64 */
--- a/core/ir/instr_shared.c
+++ b/core/ir/instr_shared.c
@@ -2656,7 +2656,8 @@ instr_is_cti(instr_t *instr) /* any control-transfer instruction */
 int
 instr_get_interrupt_number(instr_t *instr)
 {
-    CLIENT_ASSERT(instr_get_opcode(instr) == IF_X86_ELSE(OP_int, OP_svc),
+    CLIENT_ASSERT(instr_get_opcode(instr) ==
+                      IF_X86_ELSE(OP_int, IF_RISCV64_ELSE(OP_ecall, OP_svc)),
                   "instr_get_interrupt_number: instr not interrupt");
     if (instr_operands_valid(instr)) {
         ptr_int_t val = opnd_get_immed_int(instr_get_src(instr, 0));
@@ -3603,7 +3604,10 @@ instr_raw_is_tls_spill(byte *pc, reg_id_t reg, ushort offs)
     /* FIXME i#1551, i#1569: NYI on ARM/AArch64 */
     ASSERT_NOT_IMPLEMENTED(false);
     return false;
-#    endif /* X86/ARM */
+#    elif defined(RISCV64)
+    ASSERT_NOT_IMPLEMENTED(false);
+    return false;
+#    endif /* X86/ARM/RISCV64 */
 }
 
 /* this routine may upgrade a level 1 instr */
@@ -3639,6 +3643,10 @@ instr_check_tls_spill_restore(instr_t *instr, bool *spill, reg_id_t *reg, int *o
         opnd_is_abs_base_disp(memop)
 #    elif defined(AARCHXX)
         opnd_is_base_disp(memop) && opnd_get_base(memop) == dr_reg_stolen &&
+        opnd_get_index(memop) == DR_REG_NULL
+#    elif defined(RISCV64)
+        /* FIXME i#3544: Check if valid. */
+        opnd_is_base_disp(memop) && opnd_get_base(memop) == DR_REG_TP &&
         opnd_get_index(memop) == DR_REG_NULL
 #    endif
     ) {
@@ -3693,6 +3701,10 @@ instr_is_tls_xcx_spill(instr_t *instr)
         return instr_is_tls_spill(instr, REG_ECX, MANGLE_XCX_SPILL_SLOT);
 #    elif defined(AARCHXX)
     /* FIXME i#1551, i#1569: NYI on ARM/AArch64 */
+    ASSERT_NOT_IMPLEMENTED(false);
+    return false;
+#    elif defined(RISCV64)
+    /* FIXME i#3544: Not implemented */
     ASSERT_NOT_IMPLEMENTED(false);
     return false;
 #    endif
@@ -3883,7 +3895,11 @@ move_mm_reg_opcode(bool aligned16, bool aligned32)
 #    elif defined(AARCH64)
     ASSERT_NOT_IMPLEMENTED(false); /* FIXME i#1569 */
     return 0;
-#    endif /* X86/ARM */
+#    elif defined(RISCV64)
+    /* FIXME i#3544: Not implemented */
+    ASSERT_NOT_IMPLEMENTED(false);
+    return 0;
+#    endif /* X86/ARM/RISCV64 */
 }
 
 uint
--- a/core/ir/opnd_shared.c
+++ b/core/ir/opnd_shared.c
@@ -1179,6 +1179,8 @@ const reg_id_t d_r_regparms[] = {
 #    ifdef X64
     REGPARM_4,  REGPARM_5, REGPARM_6, REGPARM_7,
 #    endif
+#elif defined(RISCV64)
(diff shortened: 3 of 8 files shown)


semihalf-kardach-stanislaw:
I seem to be getting similar issues test failures even when moving back to commit `74182d892 i#2626: AArch64 v8.1 codec: add SQRDMLSH (#5563)`. I haven't tried moving further yet. All tests I've run on a Ubuntu 20.04 VM using a fresh cloudimg base (setup today).
Additionally all of the code in this PR is either under a `RISCV64` define that's never defined (it will be in the last PR that enables compilation) or in files that aren't added to cmake.
It seems that there are two kinds of faults (both in CI and locally):
```
2022-07-19T12:26:11.8056329Z Running cmd |/home/runner/work/dynamorio/dynamorio/build_debug-internal-64/clients/bin64/tool.drcacheoff.burst_static|
2022-07-19T12:26:11.8057006Z CMake Error at /home/runner/work/dynamorio/dynamorio/suite/tests/runmulti.cmake:108 (message):
2022-07-19T12:26:11.8057585Z   *** cmd failed (255): pre-DR init                
2022-07-19T12:26:11.8057886Z                                                    
2022-07-19T12:26:11.8058087Z   <Application                                     
2022-07-19T12:26:11.8058754Z   /home/runner/work/dynamorio/dynamorio/build_debug-internal-64/clients/bin64/tool.drcacheoff.burst_static
2022-07-19T12:26:11.8059685Z   (24435).  Internal Error: DynamoRIO debug check failure:
2022-07-19T12:26:11.8060180Z   /home/runner/work/dynamorio/dynamorio/core/unix/signal.c:5861
2022-07-19T12:26:11.8060832Z   syscall_signal || safe_is_in_fcache(dcontext, pc, (byte *)sc->SC_XSP)
2022-07-19T12:26:11.8061187Z                                                    
```
or
```
2022-07-19T12:26:12.8356821Z Running cmd |/home/runner/work/dynamorio/dynamorio/build_debug-internal-64/clients/bin64/tool.drcacheoff.burst_traceopts|
2022-07-19T12:26:12.8357405Z CMake Error at /home/runner/work/dynamorio/dynamorio/suite/tests/runmulti.cmake:108 (message):
2022-07-19T12:26:12.8357817Z   *** cmd failed (40): <Application                
2022-07-19T12:26:12.8358421Z   /home/runner/work/dynamorio/dynamorio/build_debug-internal-64/clients/bin64/tool.drcacheoff.burst_traceopts
2022-07-19T12:26:12.8358915Z   (24446).  Cannot correctly handle received signal 11 in thread 24446:
2022-07-19T12:26:12.8359250Z   default action in native thread.>  
```
I have a feeling this is not necessarily due to my pull request but I have no concrete proof yet.

core/ir/opnd_shared.c
```diff
@@ -1273,6 +1275,21 @@ opnd_replace_reg(opnd_t *opnd, reg_id_t old_reg, reg_id_t new_reg)
...
+            /* FIXME i#3544: RISC-V has no support for base + idx * scale + disp.
+             * We could support base + disp as long as disp == +/-1MB.
+             * If needed, instructions with this operand should be transformed
+             * to:
+             *   mul idx, idx, scale # or slli if scale is immediate
+             *   add base, base, idx
+             *   addi base, base, disp
+             */
+            CLIENT_ASSERT(false, "Not implemented");
+            (void)size;
```
abhinav92003, line 1288: What are these four casts for? Please add a comment here, and below as well.

core/ir/riscv64/instr.c
```diff
@@ -0,0 +1,451 @@
...
+}
+
+bool
+instr_is_return(instr_t *instr)
+{
+    /* FIXME i#3544: Check if valid */
+    int opc = instr_get_opcode(instr);
+    opnd_t rd = instr_get_dst(instr, 0);
+    opnd_t rs = instr_get_src(instr, 0);
+    return (opc == OP_jalr && opnd_get_reg(rd) == DR_REG_X0 &&
```
abhinav92003, line 133: So jalr is used for indirect call and return both?

I didn't do an exhaustive search but `jr ra` seems to be used for return from a routine. I may be wrong.

core/ir/riscv64/instr.c
```diff
@@ -0,0 +1,451 @@
...
+            opnd_is_memory_reference(rs) && opnd_get_base(rs) == DR_REG_X1 &&
+            opnd_get_disp(rs) == 0);
+}
+
+bool
+instr_is_cbr_arch(instr_t *instr)
+{
+    ASSERT_NOT_IMPLEMENTED(false);
```
abhinav92003, line 141: Since instr_invert_cbr is implemented, makes sense to implement this one too.

core/ir/riscv64/instr_create_api.h
```diff
@@ -0,0 +1,360 @@
...
+#ifndef _DR_IR_MACROS_RISCV64_H_
+#define _DR_IR_MACROS_RISCV64_H_ 1
+
+/****************************************************************************
+ * Platform-independent INSTR_CREATE_* macros
+ */
+/** @name Platform-independent macros */
+/** @{ */ /* doxygen start group */
+
+/* FIXME i#3544: Check and implement all INSTR_CREATE_* macros. */
```
abhinav92003, line 42: Is the currently added OP_hint equivalent to a no-op? Please add that to this comment (that the following code deliberately uses a no-op for now).

Also, mention "implement all INSTR_CREATE_ and XINST_CREATE_" macros.

core/ir/riscv64/opcode_api.h
```diff
@@ -0,0 +1,89 @@
...
+    /* NULL, */ /**< INVALID opcode */
+    /*   1 */ OP_UNDECODED,
+    /* NULL, */ /**< UNDECODED opcode */
+    /*   2 */ OP_CONTD,
+    /* NULL, */ /**< CONTD opcode */
+    /*   3 */ OP_LABEL,
+    /* NULL, */ /**< LABEL opcode */
+
+    /* FIXME i#3544: Put real opcodes here ans make sure ordering matches HW */
+    /* FIXME i#3544: should load/store width be separate opcode or in the instruction
```
abhinav92003, line 55: Trying to answer this question: does the Risc-v ISA provide separate opcodes to move different widths? If not, then we can probably just encode the width in the operand sizes.

core/ir/riscv64/opcode_api.h
```diff
@@ -0,0 +1,89 @@
...
+       flags? */
+    /* FIXME i#3544: Should pseudo instructions be here too? */
```
abhinav92003, line 57: Do these pseudo instructions have a different encoding than others? If not, we probably don't need to add them here separately. We can however add instruction create macros for them.

@derekbruening

core/ir/riscv64/opcode_api.h
```diff
@@ -0,0 +1,89 @@
...
+    OP_bge,   /**< RISC-V bge opcode. */
+    OP_bgeu,  /**< RISC-V bgeu opcode. */
+    OP_hint,  /**< RISC-V hint opcode. */
+
+    OP_AFTER_LAST,
+    OP_FIRST = OP_hint,          /**< First real opcode. */
+    OP_LAST = OP_AFTER_LAST - 1, /**< Last real opcode. */
+};
+
+/* alternative names */
```
abhinav92003, line 83: I assume we need a fixme here to remind us to replace OP_hint?

abhinav92003:
Thanks for the PR. I left some comments, mostly minor ones and questions as I don't know much about RISC-V yet.

> I have a feeling this is not necessarily due to my pull request but I have no concrete proof yet.

The `tool.drcacheoff.burst_*` failures look suspicious. I see they failed on the x86-64 test run on #5568 and #5562 too (https://github.com/DynamoRIO/dynamorio/runs/7412062010?check_suite_focus=true), so you're probably correct that they're not due to this PR. @derekbruening probably a regression due to some Github workflows internal change?

core/ir/dr_ir_opcodes.h
```diff
@@ -40,6 +40,8 @@
 #    include "dr_ir_opcodes_aarch64.h"
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: That sounds way better. I'll adopt this style with further PRs too.

core/ir/instr_inline_api.h
```diff
@@ -183,8 +183,27 @@ opnd_is_far_rel_addr(opnd_t opnd)
...
-#    endif /* X64 || ARM */
+#        elif defined(RISCV64)
+#            define OPND_IS_REL_ADDR(op) ((op).kind == REL_ADDR_kind)
+#            define opnd_is_rel_addr OPND_IS_REL_ADDR
+
+INSTR_INLINE
+bool
+opnd_is_near_rel_addr(opnd_t opnd)
+{
+    /* FIXME i#3544: Not implemented */
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: I'm a little bit conflicted here to be honest. In theory any PC-relative address can be used for unconditional jumps (so no segments per-se), however depending on the value of the offset the encoder will either have to use one instruction (jumps to `pc +/- 2MiB`):
```asm
jal x1, imm    # x1 will hold the return address (pc + 4).
               # imm is: offset / 2.
```
or two  (jumps beyond that):
```asm
auipc x2, imm     # x2 will hold: pc + imm << 12.
                  # imm is: offset<31:12>.
jalr x1, x2, imm  # x1 will hold the return address.
                  # imm is: offset<11:0>.
```

For now I'll go with the ARM way for the sake of simplicity.

core/ir/instr_inline_api.h
```diff
@@ -271,6 +290,11 @@ opnd_create_pc(app_pc pc)
                  "opnd_get_flags called on non-reg non-base-disp non-immed-int opnd")( \
                  opnd)                                                                 \
                  .aux.flags)
+#    elif defined(RISCV64)
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: At this point I would err on the side of not having any flags and simpler code.

core/ir/instr_shared.c
```diff
@@ -3603,7 +3604,10 @@ instr_raw_is_tls_spill(byte *pc, reg_id_t reg, ushort offs)
     /* FIXME i#1551, i#1569: NYI on ARM/AArch64 */
     ASSERT_NOT_IMPLEMENTED(false);
     return false;
-#    endif /* X86/ARM */
+#    elif defined(RISCV64)
+    ASSERT_NOT_IMPLEMENTED(false);
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Thanks for spotting!

core/ir/opnd_shared.c
semihalf-kardach-stanislaw: Essentially to silent `-Wunused-variable`. I'll add a comment.

core/ir/riscv64/instr.c
```diff
@@ -0,0 +1,451 @@
...
+dr_isa_mode_t
+instr_get_isa_mode(instr_t *instr)
+{
+    return DR_ISA_RV64IMAFDC;
+}
+
+int
+instr_length_arch(dcontext_t *dcontext, instr_t *instr)
+{
+    /* FIXME i#3544: C has shorter instruction length. */
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: I could be more specific here. What **C** stands for is a Compressed Instructions ISA extension. This can compress some instruction instances (depending on the operand values) to a 16bit variants.

core/ir/riscv64/instr.c
```diff
@@ -0,0 +1,451 @@
...
+    }
+    CLIENT_ASSERT(false, "instr_branch_type: unknown opcode");
+    return LINK_INDIRECT;
+}
+
+const char *
+get_opcode_name(int opc)
+{
+    ASSERT_NOT_IMPLEMENTED(false);
+    return "";
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Good idea!

core/ir/riscv64/instr.c
semihalf-kardach-stanislaw: Yes, that's correct. `jr ra` is an alias for `jalr x0, 0(ra)`. It's detailed on top of the page 140 of RISC-V Unprivileged ISA V20191213.

core/ir/riscv64/instr.c
semihalf-kardach-stanislaw: For that matter I should also implement `instr_is_ubr_arch` which is trivial.

core/ir/riscv64/opcode_api.h
semihalf-kardach-stanislaw: The load width is encoded in bits 14:12 of the instruction, the opcode is the same for all integer loads. Only floating point loads have a different opcode but they also store the width of the operand (single, double, quad) in the 14:12 bits. I'll update the comment to clarify that and add `OP_fl / OP_fs` for floating points.

core/ir/riscv64/instr_create_api.h
semihalf-kardach-stanislaw: Done.

core/ir/riscv64/instr.c
```diff
@@ -0,0 +1,451 @@
...
+opc_is_not_a_real_memory_load(int opc)
+{
+    ASSERT_NOT_IMPLEMENTED(false);
+    return opc == OP_l;
+}
+
+uint
+instr_branch_type(instr_t *cti_instr)
+{
+    /* FIXME i#3544: Add parameter checking (inc table 2.1 in spec) */
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Done.

core/ir/riscv64/instr.c
```diff
@@ -0,0 +1,451 @@
...
+{
+    ASSERT_NOT_IMPLEMENTED(false);
+    return false;
+}
+
+bool
+instr_is_nop(instr_t *instr)
+{
+    uint opc = instr_get_opcode(instr);
+    opnd_t rd = instr_get_dst(instr, 0);
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Done

core/ir/riscv64/instr_create_api.h
```diff
@@ -0,0 +1,360 @@
...
+
+#define INSTR_CREATE_ecall(dc) instr_create_0dst_0src(dc, OP_ecall)
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Done.

core/ir/riscv64/opcode_api.h
```diff
@@ -0,0 +1,89 @@
...
+    /*   0 */ OP_INVALID,
+    /* NULL, */ /**< INVALID opcode */
+    /*   1 */ OP_UNDECODED,
+    /* NULL, */ /**< UNDECODED opcode */
+    /*   2 */ OP_CONTD,
+    /* NULL, */ /**< CONTD opcode */
+    /*   3 */ OP_LABEL,
+    /* NULL, */ /**< LABEL opcode */
+
+    /* FIXME i#3544: Put real opcodes here ans make sure ordering matches HW */
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Done.

core/ir/opnd_shared.c
abhinav92003: Oh okay. Thanks.

core/ir/riscv64/instr.c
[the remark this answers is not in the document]
abhinav92003: Might be clearer to just say "Compressed Instructions ISA extension" in the comment.

core/ir/riscv64/opcode_api.h
abhinav92003: I see there are LW, LH, LB etc to load different sized values from memory. Are these just aliases for OP_l?

abhinav92003 (approved):
Looks good to me, thanks for addressing the comments. Let me check the x86-64 failures and see if there's a quick fix. If not, either Derek or I will merge this PR (using admin privileges) since it's not the cause of the regression anyway.

core/ir/riscv64/opcode_api.h
semihalf-kardach-stanislaw: Yes. `L[BHWD]` are encoded in the I-format and have the OPCODE field (bits <6:0>) set to 0000011 and width encoded in the funct3 field (bits <14:12>) as follows:
LB - 000
LH - 001
LW - 010
LD - 011

Stores and FP loads/stores use similar approach.

core/ir/riscv64/instr.c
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: True, Done.

core/ir/riscv64/opcode_api.h
abhinav92003: Thanks.

core/ir/riscv64/opcode_api.h
abhinav92003: Okay, I assume the assembler expands the pseudo instructions into the actual machine-recognizable instructions?

If so, then we definitely shouldn't add pseudo opcodes to this list.

abhinav92003:
> Let me check the x86-64 failures and see if there's a quick fix. If not, either Derek or I will merge this PR (using admin privileges) since it's not the cause of the regression anyway.

We're still working on root-causing #5570. To avoid blocking the risc-v work, I can merge this PR using admin privileges. Can I use the current PR description as the commit message?

semihalf-kardach-stanislaw:
> We're still working on root-causing #5570. To avoid blocking the risc-v work, I can merge this PR using admin privileges. Can I use the current PR description as the commit message?

Definitely, thanks!

Issue #5570, opened 2022-07-20
Regression in tool.drcacheoff.burst* tests on Github Actions


abhinav92003:
The following tests have recently started failing on the Github Actions workflow for x86-64:

```
	code_api|tool.drcacheoff.burst_static 
	code_api|tool.drcacheoff.burst_replace 
	code_api|tool.drcacheoff.burst_replaceall 
	code_api|tool.drcacheoff.burst_noreach 
	code_api|tool.drcacheoff.burst_maps 
	code_api|tool.drcacheoff.burst_traceopts 
	code_api|tool.drcacheoff.burst_threads 
	code_api|tool.drcacheoff.burst_malloc 
	code_api|tool.drcacheoff.burst_reattach 
	code_api|tool.drcacheoff.burst_threadL0filter 
	code_api|tool.drcacheoff.burst_client 
```

This is blocking some PRs, including #5569, #5568 and #5562.

Logs: https://github.com/DynamoRIO/dynamorio/runs/7430951767?check_suite_focus=true

```
2022-07-20T14:16:23.9265812Z 349:   <Application
2022-07-20T14:16:23.9266263Z 349:   /home/runner/work/dynamorio/dynamorio/build_debug-internal-64/clients/bin64/tool.drcacheoff.burst_threads
2022-07-20T14:16:23.9266655Z 349:   (24791).  Internal Error: DynamoRIO debug check failure:
2022-07-20T14:16:23.9266991Z 349:   /home/runner/work/dynamorio/dynamorio/core/unix/signal.c:5861
2022-07-20T14:16:23.9267422Z 349:   syscall_signal || safe_is_in_fcache(dcontext, pc, (byte *)sc->SC_XSP)
```

abhinav92003:
I couldn't reproduce the failure on my machine. All of them pass locally. I'll try using `tmate` to debug the issue on the GA runner.

semihalf-kardach-stanislaw:
Out of curiosity, what versions of Linux kernel, gcc and libc do you have on your machine?  

I have a VM (base image: https://cloud-images.ubuntu.com/focal/20220715/):
* Kernel: `Linux ubuntu 5.4.0-122-generic`
* gcc: `9.4.0-1ubuntu1~20.04.1`
* libc6: `2.31-0ubuntu9.9`
* libunwind8: `1.2.1-9build1`

On this VM I can see failures in the same set of tests with the same message.

abhinav92003:
[...]

PR #5572, opened 2022-07-21
i#3544 RISC-V, part 3: Add sample stubs

semihalf-kardach-stanislaw:
Add definitions required for samples to compile for RISC-V.

NOTE:

Issue: #3544


derekbruening (approved):

derekbruening:
The failures are the recent regression from the GA update: #5570.  Should be ok to merge while that's being worked on.

PR #5575, opened 2022-07-22
i#3544 RISC-V, part 4: Add drcachesim stubs

semihalf-kardach-stanislaw:
Add definitions required for drcachesim to compile for RISC-V.

NOTE:

Issue: #3544


clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1807,6 +1807,9 @@ instrument_clean_call(void *drcontext, instrlist_t *ilist, instr_t *where,
         MINSERT(ilist, where,
                 XINST_CREATE_sub(drcontext, opnd_create_reg(reg_mine),
                                  opnd_create_reg(reg_global)));
+#elif defined(RISCV64)
+        /* FIXME i#3544: Not implemented */
+        DR_ASSERT_MSG(false, "Not implemented on RISC-V");
```
abhinav92003, line 1812: We needed to handle the RISCV64 case here because the `#else` part uses architecture specific macros? It may be that other code needs to be adjusted too. Maybe add a TODO to port to RISCV64 at the top-level comment in this file too.

abhinav92003 (approved):

abhinav92003:
About the failing x86-64 tests on the CI suite: I expect #5570 to be resolved today.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1807,6 +1807,9 @@ instrument_clean_call(void *drcontext, instrlist_t *ilist, instr_t *where,
         MINSERT(ilist, where,
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Correct, only drcachesim did not compile. That to get anything beyond compilation I'll need to modify other clients. But that's for later, once we have full compile and core lib working.

clients/drcachesim/tracer/tracer.cpp
semihalf-kardach-stanislaw: Yes. IIUC `#else` is essentially x86 logic (since it has `lea`).

abhinav92003:
> About the failing x86-64 tests on the CI suite: I expect #5570 to be resolved today.

FYI this is fixed. Please merge this PR to master (you already have the privileges) after cleaning up the final commit message.

Issue #4264, opened 2020-04-19
Use C11 atomics instead of hand-rolled assembly


derekbruening:
Another idea from @algr :

Use C11 atomics to implement our mutexes and other atomic operations, instead of using error-prone (#3956, e.g.) hand-rolled assembly.

I think it's just Windows that blocks this.  We already require C++11 these days, but I didn't think Visual Studio supported many C11 features, including atomics.  I just double-checked VS2017 and indeed there is no stdatomic.h.  I don't have a VS2019 installation handy right now but searching online it does not look promising.  Clang on Windows is a potential avenue but I believe there would be a lot of work to get there (there are many VS-specific compiler and linker flags and pragmas and other things in the DR code base that probably would need changing).  Long-term that may be a good goal which would remove many of these missing language feature problems.

PR #5576, opened 2022-07-22
i#3544 RISC-V, part 5: Add atomic functions

semihalf-kardach-stanislaw:
Add RISC-V implementation of most atomic functions. Several are left
unimplemented (the exchange variants), only containing a non-atomic
logic as a reminder of what the given function should do.

Issue: #3544


core/arch/atomic_exports.h
```diff
@@ -723,7 +723,254 @@ atomic_dec_becomes_zero(volatile int *var)
                                  : "r2")
 #        define SET_IF_NOT_ZERO(flag) SET_FLAG(ne, flag)
 #        define SET_IF_NOT_LESS(flag) SET_FLAG(ge, flag)
-#    endif /* X86/ARM */
+#    elif defined(DR_HOST_RISCV64)
+#        define ATOMIC_1BYTE_READ(addr_src, addr_res)               \
+            do {                                                    \
+                __asm__ __volatile__("lbu %0, 0(%1)\n\t"            \
```
abhinav92003, line 729: Is lbu not atomic on RISC-V for byte reads? Why do we need the fence? 

Same for ATOMIC_1BYTE_WRITE.

core/arch/atomic_exports.h
```diff
@@ -723,7 +723,254 @@ atomic_dec_becomes_zero(volatile int *var)
...
+                                     : "memory");              \
+            } while (0)
+#        define ATOMIC_4BYTE_WRITE(target, value, hot_patch)   \
+            do {                                               \
+                /* Not currently used to write code */         \
+                ASSERT_CURIOSITY(!hot_patch);                  \
+                /* Assuming little endian. */                  \
+                ASSERT(BYTE_ORDER == LITTLE_ENDIAN);           \
+                /* FIXME i#3544: test the assembly!*/          \
+                __asm__ __volatile__("fence iorw,iorw\n\t"     \
```
abhinav92003, line 752: The sequence below is still non-atomic though, right? There are four separate stores, so it's possible for some stores to be visible before the others.

ATOMIC_4BYTE_ALIGNED_WRITE

abhinav92003:
Is this ready for review? (checking since I didn't see a review request)

core/arch/atomic_exports.h
```diff
@@ -723,7 +723,254 @@ atomic_dec_becomes_zero(volatile int *var)
...
+                                     "sb %0, 3(%1)\n\t"        \
+                                     :                         \
+                                     : "r"(value), "r"(target) \
+                                     : "memory");              \
+            } while (0)
+#        define ATOMIC_4BYTE_ALIGNED_WRITE(target, value, hot_patch) \
+            do {                                                     \
+                /* Not currently used to write code */               \
+                ASSERT_CURIOSITY(!hot_patch);                        \
+                __asm__ __volatile__("fence iorw,iorw\n\t"           \
```
abhinav92003, line 768: I see there are fences in almost all of these implementations. How does the fence help? It would only ensure that there's no reordering around it, but not that the read/written data is visible all at once to all threads/processors.

core/arch/atomic_exports.h
```diff
@@ -723,7 +723,254 @@ atomic_dec_becomes_zero(volatile int *var)
...
+                                     "sb %0, 7(%1)\n\t"        \
+                                     :                         \
+                                     : "r"(value), "r"(target) \
+                                     : "memory");              \
+            } while (0)
+#        define ATOMIC_8BYTE_ALIGNED_WRITE(target, value, hot_patch) \
+            do {                                                     \
+                /* Not currently used to write code */               \
+                ASSERT_CURIOSITY(!hot_patch);                        \
+                __asm__ __volatile__("fence iorw,iorw\n\t"           \
```
abhinav92003, line 811: Also assert that the given target is aligned. Same in other ALIGNED macros.

core/arch/atomic_exports.h
```diff
@@ -723,7 +723,254 @@ atomic_dec_becomes_zero(volatile int *var)
...
+                                     : "r"(addr_src), "r"(addr_res) \
+                                     : "memory");                   \
+            } while (0)
+
+#        define ADDI_suffix(sfx, var, val)                               \
+            do {                                                         \
+                /* Result width is handled by load/store instructions */ \
+                __asm__ __volatile__("l" sfx " t0, 0(%0)\n\t"            \
+                                     "add t0, t0, " #val "\n\t"          \
+                                     "fence iorw,iorw\n\t"               \
```
abhinav92003, line 831: How does the fence help make this atomic? Since other processors/threads can still modify the data before the updated value is written.

core/arch/atomic_exports.h
semihalf-kardach-stanislaw: Yes, it is. fence is for preventing re-ordering and fusing in the write buffer. GCC does the same on `atomic_inc(std_bool, bool)`.

core/arch/atomic_exports.h
semihalf-kardach-stanislaw: You're correct, this implementation is wrong. RISC-V atomic operations ISA extension (part of standard set called RV64GC ISA) only works on aligned addresses. I need to figure out how to make effectively compare and swap loop on non-aligned addresses.
That said, IIUC Aarch64 effectively only accepts aligned addresses in `ATOMIC_4BYTE_WRITE` so I could do the same?

semihalf-kardach-stanislaw:
Thanks for your review! I'll be OoO for the next 8 days so I'll get to fixing this PR once I get back.

core/arch/atomic_exports.h
semihalf-kardach-stanislaw: Done. As I require aligned addresses for essentially all atomic operations, I've added asserts to each of them.

core/arch/atomic_exports.h
semihalf-kardach-stanislaw: I've re-worked this to use "Atomic instructions" ISA extension.

semihalf-kardach-stanislaw:
@abhinav92003 I've updated this PR and now it's ready for the review.

core/arch/atomic_exports.h
abhinav92003: Sorry missed your question before. It should be okay to follow what AArch64 does yes.

core/arch/atomic_exports.h
```diff
@@ -723,7 +723,296 @@ atomic_dec_becomes_zero(volatile int *var)
...
+    __asm__ __volatile__("fence   iorw,ow\n\t"
+                         "1: lr.w.aq %0, (%1)\n\t"
+                         "   bne     %0, %2, 2f\n\t"
+                         "   sc.w.aq %0, %3, (%1)\n\t"
+                         "   bnez    %0, 1b\n\t"
+                         "2:"
+                         : "+r"(res)
+                         : "r"(var), "r"(compare), "r"(exchange)
+                         : "memory");
+    return res == 0;
```
abhinav92003, line 941: atomic_compare_exchange_int is supposed to return true if var was equal to compare.

If they were equal, res will have 0 indicating success of the store conditional, so we'll return true, which is correct.

If they weren't equal we'd branch to `2:` and res will have whatever value was in val. If it was zero, we'll incorrectly return true.

core/arch/atomic_exports.h
abhinav92003: For an operation to be atomic, it needs to be visible "all at once". Would that be true even without the fence?

core/arch/atomic_exports.h
semihalf-kardach-stanislaw: I did that. All reads/writes alias to aligned functions and contain align asserts.

abhinav92003 (approved):
I think it'll be good to have another review on this PR. @derekbruening can you help?

abhinav92003:
I've approved but please wait for Derek to have a look too and resolve the pending comments.

core/arch/atomic_exports.h
semihalf-kardach-stanislaw: Yes, there will not be a partial write even without a fence.
Fence (write-barrier) ensures ordering (as you've mentioned) and pushes out any previous write operations from a write buffer.
I'm not sure which memory_order (speaking with `stdatomics.h` terms) is the ATOMIC_4BYTE_ALIGNED_WRITE supposed to model. If the "relaxed" then we don't need a fence here. If "release" then the fence should be here to push out the writes from the critical section.

derekbruening: DR's ATOMIC macros target release-acquire semantics, so not just ensuring "untorn" or non-partial. 
 There is a comment on this somewhere -- here's one at least (should probably be at top of file too, feel free to add there too in this PR to help others): https://github.com/DynamoRIO/dynamorio/blob/master/core/arch/atomic_exports.h#L926

derekbruening: Generally, we want to match what `std::atomic` does for `std::memory_order_release` and `std::memory_order_acquire`, or the C11 equivalent.

derekbruening: Xref #4264

core/arch/atomic_exports.h
derekbruening: Is `atomic_load` w/o specifying the model using `memory_order_seq_cst`?  We would want release-acquire semantics.  I'm guessing there's still a barrier/fence of some sort: but is it a different one?

core/arch/atomic_exports.h
derekbruening: If you could double-check/confirm that all the fences in the PR are appropriate for release/acquire.

core/arch/atomic_exports.h
```diff
@@ -723,7 +723,296 @@ atomic_dec_becomes_zero(volatile int *var)
...
+                                     : "memory");              \
+            } while (0)
+#        define ATOMIC_4BYTE_ALIGNED_WRITE(target, value, hot_patch) \
+            do {                                                     \
+                /* Page 25 of Volume I: RISC-V Unprivileged ISA      \
+                 * V20191213 states that aligned loads/stores are    \
+                 * atomic. Alternatively we could use                \
+                 * amoswap.w zero %0, 0(%1) but that would incur     \
+                 * an extra read we don't need and also requires     \
+                 * alignment.                                        \
```
derekbruening, line 754: We do have alignment for this one: as the name says `ATOMIC_4BYTE_ALIGNED_WRITE`.

With `riscv64-linux-gnu-g++` version 11.2.0, for:
```
std::atomic<int> myatomic;
myatomic.store(1, std::memory_order_release);
```

It does use `amoswap`:
```
 822:   0f50000f                fence   iorw,ow
 826:   08e7a02f                amoswap.w       zero,a4,(a5)
```

Is `amoswap` in every RSIC-V or only if atomics are specified?  We're targeting +A though so maybe it doesn't matter here.

core/arch/atomic_exports.h
semihalf-kardach-stanislaw: GCC's `atomic_load` (without `_explicit`) uses `memory_order_seq_cst` and has 2 fences - before and after the 1byte load. I've implemented atomic load/store functions with acquire/release in mind, hence the fence after the load.

core/arch/atomic_exports.h
semihalf-kardach-stanislaw: `amoswap` is part of "atomic instructions" extension but I am targeting A we could use it.
What I meant by this comment is that 4 byte store is only atomic in case of aligned stores (as mentioned in the spec), therefore using `amoswap` does not give us any benefit because it also needs an aligned address (which we have) and potentially loads a value from the updated memory location. Given that the destination register is `zero` (so hard-wired zero) I guess whether the load actually takes place or not depends on particular HW implementation.
Also I've found a record in GCC's bugzilla which notes that for `atomic_store_explicit` with `relaxed` memory order can be implemented by just a store: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89835. That said I could use the same implementation as GCC's stdatomics if you prefer.

Also an interesting side-note is that the fence in the assembly you've posted above ensures that no writes to normal or I/O memory cannot be reordered before the fence (`fence iorw,ow`). My implementation also keeps the reads from reordering (`fence iorw,iorw`).

core/arch/atomic_exports.h
semihalf-kardach-stanislaw: Confirmed that I have the following semantics:
- `ATOMIC_*_READ/WRITE()` -> acquire/release order.
- The rest follow sequentially consistent order.

derekbruening: > The rest follow sequentially consistent order.

Not sure I follow this: what are "the rest"?

semihalf-kardach-stanislaw: I mean the rest of the arch-dependent atomic function definitions in this file, that is:
`ATOMIC_INC*()`, `ATOMIC_DEC*()`, `ATOMIC_ADD*()`, `atomic_add_exchange*()`, `atomic_compare_exchange*()`, `atomic_exchange()`, `atomic_*_and_test()` and `atomic_dec_becomes_zero()`.

core/arch/atomic_exports.h
```diff
@@ -723,7 +726,296 @@ atomic_dec_becomes_zero(volatile int *var)
...
+            do {                                                    \
+                /* Page 25 of Volume I: RISC-V Unprivileged ISA     \
+                 * V20191213 states that aligned loads/stores are   \
+                 * atomic.                                          \
+                 */                                                 \
+                ASSERT(ALIGNED(addr_src, 4));                       \
+                ASSERT(ALIGNED(addr_res, 4));                       \
+                /* Fence for acquire semantics. */                  \
+                __asm__ __volatile__("lw %0, 0(%1)\n\t"             \
+                                     "fence iorw,iorw"              \
```
derekbruening, line 780: If I do:
```C
static std::atomic<int> myatomic;
...
    if (myatomic.load(std::memory_order_acquire) == 0) {
```
It seems to compile to:
```
 98c:   0ff0000f                fence
 990:   439c                    lw      a5,0(a5)
 992:   0ff0000f                fence
```
With a `fence` before as well as after the load?

Update: that's -O0; compiling -O1 removes the first fence.  Hmm.  Never mind.

core/arch/atomic_exports.h
derekbruening: Those should also only need release-acquire, right?  Though for those there may not be any difference.

semihalf-kardach-stanislaw: AFAIK they are seq_cst in x86 and ARM? I.e. `ATOMIC_INC_int()` on X86 it uses `lock incl %0`, on AArch64 it uses `ldaxr+stlxr` loop (which is load-acquire + op + store+release) and ARM uses similar loop and a `dmb ish` which is a read/write seq_cst barrier on memory in a single share'ability domain (IIRC what that means exactly is implementation specific).

semihalf-kardach-stanislaw: Taking INC+DEC as an example. Yes, if we had INC+DEC follow acquire+release, then the pipeline would have more leeway in running the code out of order. Then again it would mean that only INC->DEC logic is proper for a critical section, which might lead to creating an acquire and release version of INC and DEC. That is a short path to just ditching this header and going for stdatomic. Though one of the reasons of having it might be if we ever need u128 atomic operations. Those might not be intrinsic functions (IIRC on RISC-V they are part of libatomic).

semihalf-kardach-stanislaw:
@derekbruening I'm not sure whether I can mark the last 2 conversations as resolved, hence the re-review request from me.

core/arch/atomic_exports.h
derekbruening: The same sequence might provide seq_cst in some cases, but it is necessary for acquire-release: acquire-release needs the `lock` prefix on x86.  On AArch64 for `std::memory_order_acq_rel` for a `fetch_add`, you end up with `ldaxr..stlxr` .  Similarly on ARM you need the `dmb ish`.  So the sequences for a read-modify-write are similar for acq_rel and seq_cst for the existing arches but I believe the requirement is acq_rel and not seq_cst so if it's different for RISC-V we would want the more relaxed one.

For using stdatomic that is the issue I referenced before: #4264.

core/arch/atomic_exports.h
derekbruening: Let's go with what you have.

derekbruening (approved):

core/arch/atomic_exports.h
semihalf-kardach-stanislaw: I just checked `atomic_fetch_add_explicit()`, `atomic_exchange_explicit()` and `atomic_compare_exchange_strong_explicit()` with both `memory_order_acq_rel` and `memory_order_seq_cst` and both versions are identical for each of the atomic functions (at least in GCC 11.2).

Therefore I think that my approach should be valid. `atomic_compare_exchange*()` follows GCC exactly (since it is a `lr+sc` sequence so I saw no other way of doing it) where as the rest (INC, DEC, etc.) use a simpler implementation. I.e. for `ATOMIC_ADD_int()` I have:
```
amoadd.w.aqrl zero,a1,0(a0)
```
  instead of:
```
fence   iorw,ow
amoadd.w.aq     zero,a1,0(a0)
```
The only difference I can think of is that GCC's version does extra ordering of I/O read/writes before I/O writes with the fence where as my version only orders the type of memory that `var` points to. I don't think the I/O ordering really matters for us, right?
Also my version is one instruction shorter.

derekbruening: Agreed, your version seems sufficient to me.

Issue #3187, opened 2018-09-25
preserve mxcsr in mcontext


derekbruening:
This was "PR 273323: 64-bit DR may need to save and restore MxCsr on context switches" but was never filed in a newer tracker.

Today DR does not preserve mxcsr, neither its volatile nor non-volatile portions.  This is a potential problem in both directions: DR messing up the app, and the app messing up DR, esp with DR code using more and more xmm operations as time goes by and the compilers we build DR with target newer processors.

johnfxgalea:
Would the handling of this issue result in a client not needing to call proc_save_fpstate?

johnfxgalea:
STMXCSR & LDMXCSR incur a typical latency of 5 cycles. I _think_ this would be okay for context-swtich performance?

derekbruening:
For #2629 we have marked all the floating-point state that DR does not preserve as x87-only, implying that clients could use floating-point operations so long as they are not x87: yet DR is not preserving MXCSR, as this issue shows.  DR's explicit fp-saving operations like proc_save_fpstate() are using FXSAVE which it seems does include MXCSR.

So we seem to have two choices:

A) Undo what PR #4596 did: but still clarify to make sure it's an x86-only concern and not an AArchXX concern.  So we'd replace "x87" with "any x86/x87 floating-point", i.e., including XMM/YMM/ZMM floating-point.

B) Add MXCSR to DR's mcontext.  Would that be sufficient to allow XMM fp operations?  We should double-check there aren't other status words.

PR #5577, opened 2022-07-22
i#3544 RISC-V, part 6: Add extension stubs

semihalf-kardach-stanislaw:
Add definitions required for extensions to compile for RISC-V.

NOTE:

Issue: #3544


derekbruening:
I can take this one; please request a review from me when ready.

semihalf-kardach-stanislaw:
Thanks! I'll be off for the next 8 days, so I'll address any comments when I get back.

ext/drreg/drreg.c
```diff
@@ -2069,7 +2069,12 @@ drreg_event_restore_state_without_ilist(void *drcontext, bool restore_memory,
                         __FUNCTION__, pc);
                 }
             }
-        } else if (instr_get_opcode(&inst) == IF_X86_ELSE(OP_lahf, OP_mrs)) {
+        } else if (instr_get_opcode(&inst) ==
+                   IF_X86_ELSE(OP_lahf, IF_RISCV64_ELSE(OP_hint, OP_mrs))) {
```
derekbruening, line 2073: I don't understand why this would look for the HINT opcode and assume it meant a condition code load?  HINT looks like a non-arch-state-affecting opcode which a condition code load would not be.
Plus, RISC-V has no condition codes?  So drreg would never spill them.  So RISC-V should never go into this code block?  Add a local helper function `instr_is_aflags_spill()` which always returns false for RISC-V?

Ditto below.

ext/drwrap/drwrap.c
```diff
@@ -593,6 +597,8 @@ drwrap_get_mcontext_internal(drwrap_context_t *wrapcxt, dr_mcontext_flags_t flag
                  */
 #ifdef AARCHXX
                 wrapcxt->mc->xflags = 0; /*0 is fine for ARM */
+#elif defined(RISCV64)
+                wrapcxt->mc->xflags = 0; /* FIXME i#3544: Check if valid */
```
derekbruening, line 601: If there are no condition codes (or other types of status codes for integer code) there shouldn't even be a field called `xflags`?

semihalf-kardach-stanislaw: I wasn't sure how to solve this. RISC-V indeed does not have condition codes. Instead all conditional branches are arithmetic register comparison based and those are also used for overflow checking.
`xflags` is supposed to be "platform independent condition flags" so question is should it be available on all architectures, just never updated on some? Removing it would add more ifdefs to client code, then again keeping it would give false impression of this field being functional.

Second point is that  there is a status and control register for floating point operations (`fcsr`) on RISC-V. Aarch64 keeps its own floating point registers in the machine context so I believe I should do the same. However currently I am aliasing xflags with it in mcxtx_api.h. Perhaps instead I should add a `dummy` field to mark non-existent condition flags instead?

ext/drreg/drreg.c
semihalf-kardach-stanislaw: This was because I made a fast compilation hack. I'll add aflags spill/restore functions as you suggested.

ext/drwrap/drwrap.c
derekbruening: I would vote for removing the xflags field even if it means adding some ifdefs (in a separate PR).  From separate discussion @abhinav92003 agrees.

Re: floating-point status + control: that should be a separate field.  (AArch64 has FPCR and FPSR in there; x86 might be lacking but should be fixed: #3187).

semihalf-kardach-stanislaw: Done. xflags removal is in #5587.

derekbruening:

ext/drbbdup/drbbdup.c
```diff
@@ -1743,10 +1744,12 @@ drbbdup_prepare_redirect(dr_mcontext_t *mcontext, drbbdup_manager_t *manager,
                          app_pc bb_pc)
 {
     /* Restore flags and scratch reg to their original app values. */
+#if !defined(RISCV64)
```
derekbruening, line 1747: What about the declaration of the `are_flags_dead` field and the handful of other references: should that all be ifdef-ed out as well?  Just this one and not others seems unusual?

derekbruening (approved):

ext/drreg/drreg.c
```diff
@@ -2354,10 +2362,6 @@ drreg_event_restore_state_with_ilist(void *drcontext, bool restore_memory,
 #ifdef X86
```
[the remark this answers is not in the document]
derekbruening: I wonder if the clang-format-diff has no visibility here because it's too far away in the context it looks at and that's why our test passed?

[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Maybe it has something to do with clang-format version? I have 14.0.0 and it indeed catches the indentation. CI I believe has 10.0.0. I did notice that in [another commit](https://github.com/DynamoRIO/dynamorio/pull/5588/commits/7d5d33e4fd872c7e54ca29048d84796b0a5da1bb). Perhaps something similar?

ext/drbbdup/drbbdup.c
semihalf-kardach-stanislaw: I wanted to err on the side of caution and leave the scratch register here. I'm sure I won't use the flags register, though I'm not so sure about the scratch register.

ext/drreg/drreg.c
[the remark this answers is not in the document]
derekbruening: CI was using 9 and I just upgraded it to 12 (the highest easily available on Ubuntu20 I think).

ext/drbbdup/drbbdup.c
derekbruening: > I wanted to err on the side of caution and leave the scratch register here.

Are you referring to `is_scratch_reg_dead`?  That's a separate field: I'm only talking about the other references to `are_aflags_dead` in this and the .h file.

semihalf-kardach-stanislaw: Correct me if I'm wrong but that would mean commenting out `drreg_reserve_aflags()`, `drreg_unreserve_aflags()`, `drreg_aflags_liveness()` and `drreg_are_aflags_dead()`? Then any related functionality such as `drx_aflags_are_dead()` or logic in `tracer.c`?

semihalf-kardach-stanislaw: OK I have disabled this code though I'm not sure if I'm not missing some other APIs.

ext/drreg/drreg.c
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Done.

ext/drreg/drreg.c
semihalf-kardach-stanislaw: Done.

ext/drbbdup/drbbdup.c
derekbruening: No I mean only the `are_aflags_dead` field in the `drbbdup_per_thread` structure in drbbup.c (I thought it was in drbbdup.h too but maybe not).  The PR had an ifdef around one reference to that field, but there are a couple more, so it makes sense to disable all of those.  I didn't mean all the aflags interfaces in other files.

semihalf-kardach-stanislaw: OK, I got confused when you've mentioned `are_aflags_dead` and a header file, whereas the field in `drbbdup_manager_t` is called `are_flags_dead` and it's in the `drbbdup.c` file.
I've reverted interface changes and left out ifdefs around the usage of `drbbdup_manager_t::are_flags_dead`.

semihalf-kardach-stanislaw:
Seems that azure.archive.ubuntu.com is down which causes CI to fail at `apt-get`. I'll re-try tomorrow morning (CET).

ext/drbbdup/drbbdup.c
derekbruening: Sorry, I thought I was being precise with the backtics but I typed it wrong and didn't look closely at the `git grep` to see it was all in the .c file.

derekbruening (approved):

PR #5582, opened 2022-07-28
Add cross-line counting to histogram tool

derekbruening:
Augments the histogram drcachesim tool to count every line a
cross-line access touches.  Previously it would incorrectly only
consider the first line.

Adds a test with synthetic test cases.

### DIFF
--- a/clients/drcachesim/tools/histogram.cpp
+++ b/clients/drcachesim/tools/histogram.cpp
@@ -1,5 +1,5 @@
 /* **********************************************************
- * Copyright (c) 2016-2020 Google, Inc.  All rights reserved.
+ * Copyright (c) 2016-2022 Google, Inc.  All rights reserved.
  * **********************************************************/
 
 /*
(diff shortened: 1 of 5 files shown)


clients/drcachesim/tools/histogram.cpp
```diff
@@ -95,19 +95,39 @@ histogram_t::parallel_shard_exit(void *shard_data)
...
     if (type_is_instr(memref.instr.type) ||
-        memref.instr.type == TRACE_TYPE_PREFETCH_INSTR)
-        ++shard->icache_map[memref.instr.addr >> line_size_bits_];
-    else if (memref.data.type == TRACE_TYPE_READ ||
-             memref.data.type == TRACE_TYPE_WRITE ||
-             // We may potentially handle prefetches differently.
-             // TRACE_TYPE_PREFETCH_INSTR is handled above.
-             type_is_prefetch(memref.data.type))
-        ++shard->dcache_map[memref.data.addr >> line_size_bits_];
+        memref.instr.type == TRACE_TYPE_PREFETCH_INSTR) {
```
bete0, line 112: curious as why we didn't use `type_is_prefetch(memref.instr.type)` instead here?

clients/drcachesim/tools/histogram.cpp
```diff
@@ -135,27 +155,39 @@ cmp(const std::pair<addr_t, uint64_t> &l, const std::pair<addr_t, uint64_t> &r)
...
+        *unique_icache_lines = reduced_.icache_map.size();
+    if (unique_dcache_lines != nullptr)
+        *unique_dcache_lines = reduced_.dcache_map.size();
+    return true;
+}
+
+bool
+histogram_t::print_results()
+{
+    if (!reduce_results())
```
bete0, line 182: `reduce_results()` always returns `true` which makes this check irrelevant unless we update `reduce_results()` with a path that returns `false`.

bete0 (approved):

clients/drcachesim/tools/histogram.cpp
derekbruening: Yes, it is forward-thinking, and it is virtual so a child class can override it and extend it with something that could fail.

clients/drcachesim/tools/histogram.cpp
derekbruening: That would return true for data prefetches also.  We only want to detect instruction prefetches here.

PR #5588, opened 2022-08-04
i#3544 RISC-V, part 8: Add core/ir stubs

semihalf-kardach-stanislaw:
Add stubs to get the IR related code to compile. Most of the code is
non-functional, with exception of the register name list.

Issue: #3544

### DIFF
--- a/core/arch/arch.c
+++ b/core/arch/arch.c
@@ -3444,8 +3444,10 @@ dr_mcontext_to_priv_mcontext(priv_mcontext_t *dst, dr_mcontext_t *src)
         if (TEST(DR_MC_CONTROL, src->flags)) {
             /* XXX i#2710: mc->lr should be under DR_MC_CONTROL */
             dst->xsp = src->xsp;
+#if !defined(RISCV64)
             if (src->size > offsetof(dr_mcontext_t, xflags))
                 dst->xflags = src->xflags;
+#endif
             else
                 return false;
             if (src->size > offsetof(dr_mcontext_t, pc))
@@ -3538,8 +3540,10 @@ priv_mcontext_to_dr_mcontext(dr_mcontext_t *dst, priv_mcontext_t *src)
         }
         if (TEST(DR_MC_CONTROL, dst->flags)) {
             dst->xsp = src->xsp;
+#if !defined(RISCV64)
             if (dst->size > offsetof(dr_mcontext_t, xflags))
                 dst->xflags = src->xflags;
+#endif
             else
                 return false;
             if (dst->size > offsetof(dr_mcontext_t, pc))
@@ -3758,10 +3762,15 @@ dump_mcontext(priv_mcontext_t *context, file_t f, bool dump_xml)
     }
 #endif
 
+#if defined(RISCV64)
+    print_file(f, dump_xml ? "\n\t\tpc=\"" PFX "\" />\n" : "\tpc     = " PFX "\n",
+               context->pc);
+#else
     print_file(f,
                dump_xml ? "\n\t\teflags=\"" PFX "\"\n\t\tpc=\"" PFX "\" />\n"
                         : "\teflags = " PFX "\n\tpc     = " PFX "\n",
                context->xflags, context->pc);
+#endif
 }
 
 #ifdef AARCHXX
--- a/core/arch/arch.c
+++ b/core/arch/arch.c
@@ -3448,6 +3448,7 @@ dr_mcontext_to_priv_mcontext(priv_mcontext_t *dst, dr_mcontext_t *src)
             if (src->size > offsetof(dr_mcontext_t, fcsr))
                 dst->fcsr = src->fcsr;
 #else
+            /* XXX i#5595: AArch64 should handle fpcr and fpsr here. */
             if (src->size > offsetof(dr_mcontext_t, xflags))
                 dst->xflags = src->xflags;
 #endif
@@ -3547,6 +3548,7 @@ priv_mcontext_to_dr_mcontext(dr_mcontext_t *dst, priv_mcontext_t *src)
             if (dst->size > offsetof(dr_mcontext_t, fcsr))
                 dst->fcsr = src->fcsr;
 #else
+            /* XXX i#5595: AArch64 should handle fpcr and fpsr here. */
             if (dst->size > offsetof(dr_mcontext_t, xflags))
                 dst->xflags = src->xflags;
 #endif
--- a/core/arch/arch.c
+++ b/core/arch/arch.c
@@ -3444,7 +3444,10 @@ dr_mcontext_to_priv_mcontext(priv_mcontext_t *dst, dr_mcontext_t *src)
         if (TEST(DR_MC_CONTROL, src->flags)) {
             /* XXX i#2710: mc->lr should be under DR_MC_CONTROL */
             dst->xsp = src->xsp;
-#if !defined(RISCV64)
+#if defined(RISCV64)
+            if (src->size > offsetof(dr_mcontext_t, fcsr))
+                dst->fcsr = src->fcsr;
+#else
             if (src->size > offsetof(dr_mcontext_t, xflags))
                 dst->xflags = src->xflags;
 #endif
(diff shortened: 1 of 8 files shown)


core/arch/arch.c
```diff
@@ -3444,8 +3444,10 @@ dr_mcontext_to_priv_mcontext(priv_mcontext_t *dst, dr_mcontext_t *src)
         if (TEST(DR_MC_CONTROL, src->flags)) {
             /* XXX i#2710: mc->lr should be under DR_MC_CONTROL */
             dst->xsp = src->xsp;
+#if !defined(RISCV64)
             if (src->size > offsetof(dr_mcontext_t, xflags))
                 dst->xflags = src->xflags;
+#endif
```
derekbruening, line 3450: This won't compile for RISCV64 because of the dangling `else`.

derekbruening, line 3450: Hmm, shouldn't riscv fcsr, and aarch64 fpcr and fpsr, be here under DR_MC_CONTROL (and the docs updated)?  Or do we consider them _MULTIMEDIA?  Or are they only accessible via _ALL -- but it looks like _ALL is not supposed to have anything beyond the other flags.  So we have to pick one, and mention it in the flags docs.  If you don't want to deal w/ aarch64 float regs here (presumably there is no test involving them and we'd want to add one so it is some work) maybe file an issue on it?

core/arch/arch.c
```diff
@@ -3538,8 +3540,10 @@ priv_mcontext_to_dr_mcontext(dr_mcontext_t *dst, priv_mcontext_t *src)
         }
         if (TEST(DR_MC_CONTROL, dst->flags)) {
             dst->xsp = src->xsp;
+#if !defined(RISCV64)
             if (dst->size > offsetof(dr_mcontext_t, xflags))
                 dst->xflags = src->xflags;
+#endif
```
derekbruening, line 3546: Same comments: dangling else; missing fcsr.

core/ir/riscv64/decode.c
```diff
@@ -0,0 +1,247 @@
...
+app_pc
+dr_app_pc_as_load_target(dr_isa_mode_t isa_mode, app_pc pc)
+{
+    return pc;
+}
+
+byte *
+decode_eflags_usage(void *drcontext, byte *pc, uint *usage, dr_opnd_query_flags_t flags)
+{
+    ASSERT_NOT_IMPLEMENTED(false);
```
derekbruening, line 66: Generally we put a TODO or XXX comment with the i#NNN for unfinished code.

Ditto below.

core/ir/riscv64/decode.c
```diff
@@ -0,0 +1,247 @@
...
+#ifdef DECODE_UNIT_TEST
+
+#    include "instr_create_shared.h"
+
+int
+main()
+{
+    bool res = true;
+    standalone_init();
+    standalone_exit();
```
derekbruening, line 243: I would expect a `TODO i#3544` comment about adding some tests.

core/ir/riscv64/decode_private.h
```diff
@@ -0,0 +1,39 @@
...
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+#ifndef DECODE_PRIVATE_H
+#define DECODE_PRIVATE_H 1
+
+struct _decode_info_t {
+};
```
derekbruening, line 37: I would expect a `TODO i#3544` comment about adding content.

core/ir/riscv64/encode.c
```diff
@@ -0,0 +1,116 @@
...
+byte *
+copy_and_re_relativize_raw_instr(dcontext_t *dcontext, instr_t *instr, byte *dst_pc,
+                                 byte *final_pc)
+{
+    ASSERT_NOT_IMPLEMENTED(false);
+    return NULL;
+}
+
+void
+append_fcache_enter_prologue(dcontext_t *dcontext, instrlist_t *ilist, bool absolute)
```
derekbruening, line 113: Hmm, this seems out of place in encode.c which is supposed to be pure IR and not including details of DR's context switch sequences?  The others have it in emit_utils.c.

core/lib/globals_api.h
```diff
@@ -780,6 +780,7 @@ typedef enum {
      * On x86, selects the xsp, xflags, and xip fields.
      * On ARM, selects the sp, pc, and cpsr fields.
      * On AArch64, selects the sp, pc, and nzcv fields.
+     * On RISC-V, selects the sp, and pc fields.
```
derekbruening, line 783: See other comment: include fcsr somewhere.

core/arch/arch.c
semihalf-kardach-stanislaw: DR_MC_CONTROL makes sense. I didn't want to disturb Aarch64 and since it wasn't saving FP control registers, I didn't add mine. I'll file an issue.

core/ir/riscv64/decode.c
semihalf-kardach-stanislaw: Sure, I'll add it.
My reasoning for using only `ASSERT_NOT_IMPLEMENTED()` was that it is a RISC-V specific file so the issue comment seemed superfluous. Conversely in generic files I've put both the assert and the comment for the unimplemented RISC-V bits.

derekbruening:
> Nudge for the Aarch64 Jenkins to re-run tests

The string "run arm tests" as a comment does this: no need for a commit.

core/ir/riscv64/decode.c
derekbruening: I'm imagining that some of these will be implemented in this file but some will be left unimplemented b/c they're not exercised on code paths of some initial tests, and a long time later we have most of the file filled in but not all and someone reading it maybe can't tell whether something was deliberately decided to not be needed on RISC-V (though you'd expect a comment about that too) or it was just never looked at beyond the initial skeleton code.  It does add extra lines but IMHO it makes it seem more carefully considered.

derekbruening:

derekbruening:
Normally the PR owner would respond to the review comments: without responses it ends up looking like maybe they were overlooked and not addressed.

semihalf-kardach-stanislaw:
My bad. I'm used to mailing-list style of reviews where it's usually a new patch revision. I'll answer to all comments right away.

core/arch/arch.c
semihalf-kardach-stanislaw: I've added saving of `fcsr`. Haven't touched on Aarch64 logic and opened an issue on that instead.

core/arch/arch.c
semihalf-kardach-stanislaw: Done.

core/ir/riscv64/decode.c
semihalf-kardach-stanislaw: I've added the FIXME comment to all asserts.

core/ir/riscv64/decode.c
semihalf-kardach-stanislaw: Correct, I've added a FIXME above this code.

core/ir/riscv64/decode_private.h
semihalf-kardach-stanislaw: Comment added.

core/ir/riscv64/encode.c
semihalf-kardach-stanislaw: True. I must have pasted this here by mistake. Moved to `emit_utils.c`.

core/arch/arch.c
derekbruening: Could we add a `XXX i#nnn` comment on the missing AArch64 fields here pointing at that new issue?

derekbruening (approved):

core/arch/arch.c
semihalf-kardach-stanislaw: Done.

Issue #5560, opened 2022-07-13
Refactor and split drmemtrace tracer.cpp


derekbruening:
Splitting out from https://github.com/DynamoRIO/dynamorio/issues/3995#issuecomment-1063198769
Which came originally from https://github.com/DynamoRIO/dynamorio/pull/5393#discussion_r820264882

tracer.cpp has grown large and complex.  The idea is to split it up into separate files and possibly classes or other organization for more modularity and easier maintainability.

PR #5593, opened 2022-08-05
i#5560 drmemtrace refactor: Split counting + i/o code

derekbruening:
Separates out two big pieces of tracer.cpp into their own files:
instruction counting into instr_counter.cpp and i/o into output.cpp.
Both still have some dependencies on some of the same TLS and globals
as the rest of the tracer, but shrinking the tracer.cpp file is still
of value, and the interfaces between the files are not huge.  Future
work could further modularize and isolate.

Adds a namespace, especially important now that we have global symbols
shared between files.

Issue: #5560, #4343

### DIFF
--- a/clients/drcachesim/tracer/tracer.h
+++ b/clients/drcachesim/tracer/tracer.h
@@ -51,7 +51,8 @@
 #    include <lz4frame.h>
 #endif
 
-namespace dynamorio::drmemtrace {
+namespace dynamorio {
+namespace drmemtrace {
 
 extern named_pipe_t ipc_pipe;
 // A clean exit via dr_exit_process() is not supported from init code, but
@@ -225,6 +226,7 @@ has_tracing_windows()
     return op_trace_for_instrs.get_value() > 0 || op_retrace_every_instrs.get_value() > 0;
 }
 
-} // namespace dynamorio::drmemtrace
+} // namespace drmemtrace
+} // namespace dynamorio
 
 #endif /* _TRACER_ */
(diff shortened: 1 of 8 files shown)


derekbruening:
This shrinks tracer.cpp from 3405 lines down to 1881 lines.

The diff may not be easy to read: no changes are intended except for adding the namespaces; the rest is pure code movement.

derekbruening:
Windows builds are failing frequently downloading doxygen (again)...

clients/drcachesim/tracer/tracer.h
```diff
@@ -0,0 +1,239 @@
...
+/* XXX we don't have the inlining implemented yet for 32-bit architectures. */
+#endif
+
+#define INSTR_COUNT_LOCAL_UNIT 10000
+
+/***************************************************************************
+ * Buffer writing to disk.
+ */
+
+// NOCHECK move to output.h?
```
abhinav92003, line 175: This was supposed to be addressed.

abhinav92003 (approved):
It's hard to do a more thorough review of the large diff. For more confidence, we should do more automated testing, like tracing on large apps.

clients/drcachesim/tracer/tracer.h
derekbruening: Removed.  Ended up deciding to leave the content where it is.

derekbruening:
32-bit windows failure is setthreadcontext #3213

derekbruening:
I ran on some proprietary tests which are larger than the regression tests here including some larger "import tests" and the results match those before this PR.

PR #5594, opened 2022-08-05
i#5505 kernel trace: Upgrade dr$sim to record syscall's kernel PT

dolanzhao:
Updates drpt2ir and drpttracer and adds a new module syscall_pt_trace that provides an API for syscall's kernel PT tracing:

(1) Updates drpttracer extension. 
Adds drpttracer_create_tracer() and drpttracer_destroy_tracer().
In the original implementation, the client will get a tracer by calling drpttracer_start_tracing() and destroy the tracer when calling drpttracer_end_tracing(). If the client traces multiple times, the tracer's creation and destruction will cause a big overhead. Currently, the client can use drpttracer_create_tracer() to get a tracer and use drpttracer_start_tracing() and drpttracer_end_tracing() to start and stop a tracing processes, then use drpttracer_destory_tracer() to destroy the tracer. In the future implementation, multiple traces can share one tracer, and the overhead is reduced to an acceptable level.

(2) Adds the syscall_pt_trace module for dr$sim to record syscall's kernel PT trace. 
This module wraps a PT tracer as a class and provides APIs for thread-local syscall tracing. Also, it provides APIs to check the currently tracing syscall's sysnum and last recorded syscall's id. In this PR, syscall_pt_trace will create a trace handle for each syscall, making the decoding process simple for the decoder. 
In the current implementation, syscall_pt_trace module can cause significant overhead. A pending optimization can reduce the overhead caused by pttracer initialization. The optimization method is the module use one pttracer handle for all syscalls in one thread and dumps the PT segments into one file. This method needs to upgrade drpt2ir to support decoding PT fragments. We will merge the optimized code after we finish implementing drpt2ir's patch.

(3) Updates dr$sim to support kernel tracing. 
In the current implementation, every monitored thread will create a tracer during the thread start, and dr$sim will start the syscall kernel tracing in the pre-syscall callback and stop the syscall kernel tracing at post-syscall callbacks. And every syscall will dump the PT data and metadata to files. Also, before the process end, dr$sim will use the 'perf record --kcore' command to copy out kcore and kallsyms used in the PT post processer.

(4) Updates drpt2ir. 
The current implementation supports using the metadata file to initiate an instance of pt2ir_config_t and replaces some c pointer parameters with c++ reference parameters.

(5) Updates drpt2trace. 
Adds a new option to support directly reading metadata files to initialize pt2ir_config_t. In the current implementation, drpt2trace will first use a metadata file to init the instance of pt2ir_config_t. Then if the user specifies some options, it will overwrite the corresponding field in the instance of pt2ir_config_t.

(6) Add a new test for dr$sim to check that the kernel PT output is correct.
Defines a new macro for kernel feature check and adds a simple test for dr$sim. The macro will run dr$sim to generate kernel PT and drpt2trace to check whether the generated kernel PT can be decoded.

Issue: #5505


clients/drcachesim/tracer/syscall_pt_trace.cpp
```diff
@@ -0,0 +1,138 @@
...
+#include "drpttracer.h"
+#include "syscall_pt_trace.h"
+
+#ifndef BUILD_PT_TRACER
+#    error "This module requires the drpttracer extension."
+#endif
+
+#define RING_BUFFER_SIZE_SHIFT 8
+#define PT_DATA_FILE_NAME_SUFFIX ".pt"
+#define PT_METADAT_FILE_NAME_SUFFIX ".meta"
```
abhinav92003, line 47: s/PT_METADAT_FILE_NAME_SUFFIX/PT_METADATA_FILE_NAME_SUFFIX/

clients/drcachesim/tracer/syscall_pt_trace.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+syscall_pt_trace_t::syscall_pt_trace_t()
+    : write_file_func_(nullptr)
+    , pttracer_handle_ {}
+    , recorded_syscall_num_(0)
+    , recording_sysnum_(-1)
+    , drcontext_(nullptr)
+    , pt_dir_name_ { '\0' }
```
abhinav92003, line 55: Why not ""?

clients/drcachesim/tracer/syscall_pt_trace.cpp
```diff
@@ -0,0 +1,138 @@
...
+{
+}
+
+syscall_pt_trace_t::~syscall_pt_trace_t()
+{
+}
+
+bool
+syscall_pt_trace_t::init(void *drcontext, char *pt_dir_name, size_t pt_dir_name_size,
```
abhinav92003, line 64: Can we not compute `pt_dir_name_size` inside init?

clients/drcachesim/tracer/syscall_pt_trace.cpp
```diff
@@ -0,0 +1,138 @@
...
+                         ssize_t (*write_file_func)(file_t file, const void *data,
+                                                    size_t count))
+{
+    drcontext_ = drcontext;
+    if (drpttracer_create_tracer(drcontext_, DRPTTRACER_TRACING_ONLY_KERNEL,
+                                 RING_BUFFER_SIZE_SHIFT, RING_BUFFER_SIZE_SHIFT,
```
abhinav92003, line 70: RING_BUFFER_SIZE_SHIFT etc can be declared as constexpr local to this routine if they aren't needed anywhere else.

clients/drcachesim/tracer/syscall_pt_trace.cpp
abhinav92003: Why not use a const std::string for `pt_dir_name`?

clients/drcachesim/tracer/syscall_pt_trace.cpp
```diff
@@ -0,0 +1,138 @@
...
+    ASSERT(data->pt != nullptr, "pt is nullptr");
+    ASSERT(data->pt_size > 0, "pt_size is 0");
+    if (data->pt == nullptr || data->pt_size == 0) {
+        return false;
+    }
+
+    /* Dump PT trace data to file '{thread_id}.{syscall_id}.pt'. */
+    char pt_filename[MAXIMUM_PATH];
+    dr_snprintf(pt_filename, BUFFER_SIZE_ELEMENTS(pt_filename), "%s%s%d.%d%s",
+                pt_dir_name_, DIRSEP, dr_get_thread_id(drcontext_), recorded_syscall_num_,
```
abhinav92003, line 124: Can we just inline DIRSEP and PT_DATA_FILE_NAME_SUFFIX in the format string?

clients/drcachesim/tracer/syscall_pt_trace.cpp
```diff
@@ -0,0 +1,138 @@
...
+    }
+    recording_sysnum_ = -1;
+    recorded_syscall_num_++;
+    return trace_data_dump(output);
+}
+
+bool
+syscall_pt_trace_t::trace_data_dump(drpttracer_output_cleanup_last_t &output)
+{
+    ASSERT(drcontext_ != nullptr, "drcontext_ is nullptr");
```
abhinav92003, line 111: Add TODO comment for buffering PT data.

clients/drcachesim/tracer/syscall_pt_trace.cpp
```diff
@@ -0,0 +1,138 @@
...
+    if (data->pt == nullptr || data->pt_size == 0) {
+        return false;
+    }
+
+    /* Dump PT trace data to file '{thread_id}.{syscall_id}.pt'. */
+    char pt_filename[MAXIMUM_PATH];
+    dr_snprintf(pt_filename, BUFFER_SIZE_ELEMENTS(pt_filename), "%s%s%d.%d%s",
+                pt_dir_name_, DIRSEP, dr_get_thread_id(drcontext_), recorded_syscall_num_,
+                PT_DATA_FILE_NAME_SUFFIX);
+    file_t pt_file = dr_open_file(pt_filename, DR_FILE_WRITE_OVERWRITE);
```
abhinav92003, line 126: Are we really okay with overwriting? Maybe we should use DR_FILE_WRITE_REQUIRE_NEW.

clients/drcachesim/tracer/syscall_pt_trace.h
```diff
@@ -0,0 +1,140 @@
...
+/* The auto cleanup wrapper of pttracer handle.
+ * This can ensure the pttracer handle is cleaned up when it is out of scope.
+ */
+struct drpttracer_handle_cleanup_last_t {
+public:
+    ~drpttracer_handle_cleanup_last_t()
+    {
+        if (handle != nullptr) {
+            void *drcontext = dr_get_current_drcontext();
+            drpttracer_destory_tracer(drcontext, handle);
```
abhinav92003, line 54: s/drpttracer_destory_tracer/drpttracer_destroy_handle/

clients/drcachesim/tracer/syscall_pt_trace.h
```diff
@@ -0,0 +1,140 @@
...
+    ssize_t (*write_file_func_)(file_t file, const void *data, size_t count);
+
+    /* The pttracer handle holded by this instance. */
+    drpttracer_handle_cleanup_last_t pttracer_handle_;
+
+    /* The number of recorded syscall. */
+    int recorded_syscall_num_;
+
+    /* The sysnum of current recording syscall. */
+    int recording_sysnum_;
```
abhinav92003, line 129: cur_recording_sysnum_ is clearer IMHO.

clients/drcachesim/tracer/syscall_pt_trace.h
```diff
@@ -0,0 +1,140 @@
...
+
+    /* The drcontext.
+     * We need ensure pass the same context to all drpttracer's APIs.
+     */
+    void *drcontext_;
+
+    /* The output directory name. */
+    char pt_dir_name_[MAXIMUM_PATH];
```
abhinav92003, line 137: Why not use std::string?

clients/drcachesim/tracer/syscall_pt_trace.h
```diff
@@ -0,0 +1,140 @@
...
+            void *drcontext = dr_get_current_drcontext();
+            drpttracer_destroy_output(drcontext, data);
+            data = nullptr;
+        }
+    }
+    drpttracer_output_t *data = nullptr;
+};
+/* This class is not thread-safe: the caller should create a separate instance per thread.
+ */
+class syscall_pt_trace_t {
```
abhinav92003, line 78: A better name would be syscall_pt_tracer_t

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -2395,9 +2414,56 @@ event_pre_syscall(void *drcontext, int sysnum)
...
+
+    ASSERT(data->syscall_pt_trace.get_recording_sysnum() == sysnum,
+           "last tracing isn't stopped");
+    if (!data->syscall_pt_trace.stop_syscall_pt_trace()) {
+        ASSERT(false, "failed to stop syscall pt trace");
+        return;
+    }
+
+    /* Write a marker to userspace raw trace. */
+    if (tracing_disabled.load(std::memory_order_acquire) != BBDUP_MODE_TRACE)
```
abhinav92003, line 2454: So we gather the syscall trace even if we're not tracing userspace code? Add a comment.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -2902,6 +2974,26 @@ event_thread_exit(void *drcontext)
     if (BUF_PTR(data->seg_base) != NULL) {
         /* This thread was *not* filtered out. */
 
+#ifdef BUILD_PT_TRACER
+        if (op_offline.get_value() && op_enable_kernel_tracing.get_value()) {
+            int recording_sysnum = data->syscall_pt_trace.get_recording_sysnum();
```
abhinav92003, line 2979: So the pre syscall event is invoked before the exit event for SYS_exit?

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -2902,6 +2974,26 @@ event_thread_exit(void *drcontext)
...
+            if (recording_sysnum != SYS_exit && recording_sysnum != SYS_exit_group) {
+                NOTIFY(0, "ERROR: The last recorded syscall of thread T%d is %d.\n",
+                       dr_get_thread_id(drcontext), recording_sysnum);
+                ASSERT(recording_sysnum == SYS_exit || recording_sysnum == SYS_exit_group,
+                       "recording syscall is not exit");
+            }
+
+            if (!data->syscall_pt_trace.stop_syscall_pt_trace()) {
```
abhinav92003, line 2987: We cannot really trace SYS_exit right? We should just skip it in the pre_syscall event.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -2987,6 +3079,32 @@ event_thread_exit(void *drcontext)
...
+        "cp perf.data/kcore_dir/kallsyms %s/ \n"                                     \
+        "chmod 755 -R %s \n"                                                         \
+        "rm -rf perf.data \n"
+#    define SHELLSCRIPT_MAX_LEN 512 + MAXIMUM_PATH * 2
+        char shellscript[SHELLSCRIPT_MAX_LEN];
+        dr_snprintf(shellscript, BUFFER_SIZE_ELEMENTS(shellscript), SHELLSCRIPT_FMT,
+                    kernel_pt_logsubdir, kernel_pt_logsubdir, logdir);
+        NULL_TERMINATE_BUFFER(shellscript);
+        int ret = system(shellscript);
+        // int ret = 0;
```
abhinav92003, line 3099: Remove commented out code.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -2987,6 +3079,32 @@ event_thread_exit(void *drcontext)
...
+        "perf record --kcore -e intel_pt/cyc,noretcomp/k echo '' >/dev/null 2>&1 \n" \
+        "chmod 755 -R perf.data \n"                                                  \
+        "cp perf.data/kcore_dir/kcore %s/ \n"                                        \
+        "cp perf.data/kcore_dir/kallsyms %s/ \n"                                     \
+        "chmod 755 -R %s \n"                                                         \
+        "rm -rf perf.data \n"
+#    define SHELLSCRIPT_MAX_LEN 512 + MAXIMUM_PATH * 2
+        char shellscript[SHELLSCRIPT_MAX_LEN];
+        dr_snprintf(shellscript, BUFFER_SIZE_ELEMENTS(shellscript), SHELLSCRIPT_FMT,
+                    kernel_pt_logsubdir, kernel_pt_logsubdir, logdir);
```
abhinav92003, line 3096: Why is there a "log" in the identifier? It makes it seem like it's for the debug logfiles.

ext/drpttracer/drpttracer.c
```diff
@@ -446,57 +521,64 @@ drpttracer_exit(void)
...
         ASSERT(false, "failed to create pttracer_handle");
         return DRPTTRACER_ERROR_FAILED_TO_CREATE_PTTRACER_HANDLE;
     }
+    *(pttracer_handle_t **)tracer_handle = handle;
+    return DRPTTRACER_SUCCESS;
+}
+
+DR_EXPORT
+drpttracer_status_t
+drpttracer_destory_tracer(IN void *drcontext, INOUT void *tracer_handle)
```
abhinav92003, line 546: s/drpttracer_destory_tracer/drpttracer_destroy_tracer/

abhinav92003 (commented):
Partial review.

clients/drcachesim/CMakeLists.txt
```diff
@@ -301,6 +301,12 @@ macro(add_drmemtrace name type)
     tracer/func_trace.cpp
     ${client_and_sim_srcs}
     )
+  if (BUILD_PT_TRACER)
+    set(drmemtrace_srcs ${drmemtrace_srcs}
+      tracer/syscall_pt_trace.cpp
+    )
+    add_definitions(-DBUILD_PT_TRACER)
```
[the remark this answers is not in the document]
dolanzhao: dr$sim didn't include configure.h. So we need to define it here.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -254,23 +259,11 @@ option_init(int argc, const char *argv[])
         print_usage();
         return false;
     }
-    std::vector<droption_parser_t *> required_op_list;
-    required_op_list.push_back(&op_raw_pt);
-    required_op_list.push_back(&op_pt_cpu_family);
-    required_op_list.push_back(&op_pt_cpu_model);
-    required_op_list.push_back(&op_pt_cpu_stepping);
```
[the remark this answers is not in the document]
dolanzhao: Not yet. If the user not specified this parameters, the decode can success.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -296,23 +289,22 @@ option_init(int argc, const char *argv[])
     }
 
     /* Check if the required options for sideband mode are specified. */
-    if (op_primary_sb.specified()) {
-        std::vector<droption_parser_t *> sb_required_op_list;
-        sb_required_op_list.push_back(&op_sb_sample_type);
```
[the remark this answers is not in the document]
dolanzhao: They are not required arguments.

clients/drcachesim/tracer/tracer.cpp
dolanzhao: Yes.

clients/drcachesim/tracer/tracer.cpp
dolanzhao: We use logsubdir to represent the directory to store user space raw traces . Do we need to follow the same style?

ext/drpttracer/drpttracer.c
dolanzhao: Done.

clients/drcachesim/common/trace_entry.h
```diff
@@ -369,6 +369,11 @@ typedef enum {
      */
     TRACE_MARKER_TYPE_PAGE_SIZE,
 
+    /**
+     * The marker value contains the ID of recorded syscalls.
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -85,6 +85,11 @@ static droption_t<unsigned long long> op_elf_base(
     "This is an optional option in ELF Mode. Specifies the runtime load address of "
     "the elf file. For kernel cases, this always should be 0x0, so it is not required. "
     "But if -elf specified file's runtime load address is not 0x0, it must be set.");
+static droption_t<std::string> op_raw_pt_meta(
+    DROPTION_SCOPE_FRONTEND, "raw_pt_meta", "",
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -296,23 +289,22 @@ option_init(int argc, const char *argv[])
...
+    if (op_primary_sb.specified() && !op_sb_sample_type.specified()) {
+        std::cerr << CLIENT_NAME << ": option " << op_sb_sample_type.get_name()
+                  << " is required in sideband mode." << std::endl;
+        print_usage();
+        return false;
     }
     return true;
 }
+
+#define IF_SPECIFIED_THEN_SET(__OP_VARIABLE__, __SETTED_VARIABLE__) \
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/drpt2trace/pt2ir.h
```diff
@@ -56,6 +58,21 @@
 #    define INOUT // nothing
 #endif
 
+/* The auto cleanup wrapper of instrlist_t.
+ * This can ensure the instant of instrlist_t is cleaned up when it is out of scope.
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/syscall_pt_trace.cpp
dolanzhao: Done.

clients/drcachesim/tracer/syscall_pt_trace.cpp
dolanzhao: I tried using "" to initialize a string in the constructer. It seems DR will crash if I do this. Is this a limit in DR? Or is it an issue of DR? @derekbruening 
The two crash codes are shown in below:

(1) case 1: crash before print
 
``` c++
class syscall_pt_trace_t {
private:
    std::string pt_dir_name_;
};

syscall_pt_trace_t::syscall_pt_trace_t()
    :pt_dir_name_ ("")
{
    // crash before print
    dr_fprintf(stderr, "reach here");
}
```
(2)  case 2: crash after print

``` c++
class syscall_pt_trace_t {
private:
    std::string pt_dir_name_;
};
syscall_pt_trace_t::syscall_pt_trace_t()
{
    dr_fprintf(stderr, "reach here");
    // crash after print
    pt_dir_name_ = "";
}
```

clients/drcachesim/tracer/syscall_pt_trace.cpp
dolanzhao: Done.

clients/drcachesim/tracer/syscall_pt_trace.cpp
dolanzhao: I see in tracer.cpp we didn't inline them. Do we need to keep the uniform style?
https://github.com/DynamoRIO/dynamorio/blob/b503b35cd292e4c0660de3cd6c572d8b44eb378a/clients/drcachesim/tracer/tracer.cpp#L3209

clients/drcachesim/tracer/syscall_pt_trace.cpp
dolanzhao: Done.

clients/drcachesim/tracer/syscall_pt_trace.cpp
dolanzhao: Done.

clients/drcachesim/tracer/syscall_pt_trace.h
dolanzhao: Done.

clients/drcachesim/tracer/syscall_pt_trace.h
dolanzhao: Done.

clients/drcachesim/tracer/tracer.cpp
abhinav92003: I misread the code. sorry.

So we do look at `tracing_disabled` in the pre-syscall event. In this post-syscall event though, we're calling `stop_syscall_pt_trace` before we check it.

clients/drcachesim/tracer/tracer.cpp
dolanzhao: Done.

clients/drcachesim/tracer/syscall_pt_trace.cpp
derekbruening: Please elaborate on "crash": what precisely happens, at what callstack.

`stderr` is an incorrect argument to dr_fprintf which takes file descriptors (like STDERR) not FILE pointers.  Maybe that is what is causing problems: but you'd expect it to just fail; but if you're relying on that to tell you when something happened it would not be reliable.

We do not want to call malloc during tracing (b/c we statically link drmemtrace into apps) so do not use std::string after init time.  Is this used after init time?  There is a check in a test which will fail if malloc is called.

clients/drcachesim/CMakeLists.txt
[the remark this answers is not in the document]
dolanzhao: @derekbruening I try to include configure.h to dr$sim. But it will cause some redefine issues. Could I not include the configrue.h to clients?

clients/drcachesim/tracer/tracer.cpp
dolanzhao: Done.

clients/drcachesim/tracer/tracer.cpp
dolanzhao: Done.

clients/drcachesim/CMakeLists.txt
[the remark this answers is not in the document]
derekbruening: configure_DR_client sets a bunch of defines which overlap so we generally don't use configure.h in clients: it's more for the core.  Maybe what you want is a different header for drcachesim, if there are a bunch of cmake-set defines.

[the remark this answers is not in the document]
dolanzhao: Got it. I only need to add one new cmake-set define for kernel tracing part. So I think using add_definitions is enough.

clients/drcachesim/tracer/syscall_pt_trace.cpp
dolanzhao: Sorry for giving you some confusion. 
The crash occurs when I add a new field in tracer.cpp's per_thread_t, which is an instance of the class containing std::string . During the event_thread_init callback, the code executes to the line where the per_thread_t instance will be initialized. The client will crash. The code is shown as below:
``` c++
class syscall_pt_trace_t {
private:
    std::string test_;
};
syscall_pt_trace_t::syscall_pt_trace_t()
: test_("")
{
}
per_thread_t {
    ...
    syscall_pt_trace_t syscall_pt_trace;
}

static void
event_thread_init(void *drcontext)
{
    per_thread_t *data = (per_thread_t *)dr_thread_alloc(drcontext, sizeof(per_thread_t));
    DR_ASSERT(data != NULL);
    *data = {};  // crash in the this line
    ...
}
```
If I remove the 'test_' in syscall_pt_trace_t it works fine. I know that the initialization of std::string invokes some allocation function. Do you think that's why the crash happened?

derekbruening: A class does not need explicit initialization to its default value: that is done automatically, so this `test_("")` is superfluous.

I would not expect that line to crash: as the comment says, it should be properly handled wrt classes inside the struct.  As with any crash in our own C++ code, look at the callstack to understand what is going on: where exactly is the crash point?  I would expect an error in zeroing that struct to only cause a crash later when accessing the vtable of a class field or something.  A crash on the zeroing line itself implies a read-only page or something??  Again, look at the callstack: no need to speculate when we can get real data.

ext/drpttracer/drpttracer.c
```diff
@@ -517,47 +599,67 @@ drpttracer_end_tracing(IN void *drcontext, INOUT void *tracer_handle,
         return DRPTTRACER_ERROR_FAILED_TO_STOP_TRACING;
     }
 
+    /* Check if the PT data in the aux ring buffer is overwritten. */
+    uint64_t pt_size = handle->header->aux_head - handle->header->aux_tail;
+    if (pt_size > handle->header->aux_size) {
+        ERRMSG("the size of PT trace data(%" PRIu64
```
derekbruening, line 605: DR provides cross-platform format string codes (e.g. `UINT64_FORMAT_CODE`) which we generally use in existing code.  That was provided before these new C99 codes were implemented in Visual Studio.  Maybe it is ok to use them now since we're on VS2019 which I believe (finally) implements all of C99?

ext/drpttracer/drpttracer.dox
```diff
@@ -88,37 +85,48 @@ the buffer size is 2^pt_size_shift * PAGE_SIZE.
...
-To stop collecting the trace, the client needs to pass the trace handle to
-drpttracer_end_tracing(). When the tracing is stopped, the stop function will allocate an
-instance of #drpttracer_output_t, and copy the PT data, sideband data, and metadata to a
-#drpttracer_output_t instance which is returned to the instance to the client. The client
-can dump the data in the output to different files for offline post-processing. The user
-can use the library pt2ir_t in drcachesim to convert the PT trace to DynamoRIO's
-IR.
-
-Additionally, if drpttracer_end_tracing() detect an overflow, it will return an error
+which may cause issues in pt2ir_t decoding.
```
derekbruening, line 96: Use #

ext/drpttracer/drpttracer.dox
```diff
@@ -88,37 +85,48 @@ the buffer size is 2^pt_size_shift * PAGE_SIZE.
...
+
+After creating a pttracer handle successfully, the client can start or stop a tracing
+session by calling drpttracer_start_tracing() and drpttracer_stop_tracing().
+
+ - drpttracer_start_tracing()
+ - drpttracer_stop_tracing()
+
+The client must pass an invaild trace handle to the start and stop function. And the
```
derekbruening, line 104: "invalid" is spelled wrong: but I don't think that's the word you mean -- noone should pass an invalid anything.  Do you mean "valid"?

ext/drpttracer/drpttracer.dox
```diff
@@ -88,37 +85,48 @@ the buffer size is 2^pt_size_shift * PAGE_SIZE.
...
+client can use one handle for multiple tracing sessions. When the tracing is stopped,
+the stop function will allocate an instance of #drpttracer_output_t, and copy the PT data,
+sideband data, and metadata to a #drpttracer_output_t instance which is returned to the
+client. If drpttracer_end_tracing() detect an overflow, it will return an error
```
derekbruening, line 108: grammar: s/detect/detects/

ext/drpttracer/drpttracer.dox
```diff
@@ -88,37 +85,48 @@ the buffer size is 2^pt_size_shift * PAGE_SIZE.
...
 
  - Return #DRPTTRACER_ERROR_OVERWRITTEN_SIDEBAND_DATA if the sideband data is
 overwritten.
 
-Also, to avoid memory leak, the client must call the following function to destroy the
-output instance after using it:
+After all the tracing sessions are stopped, the client needs to call
+the following function to free the pttracer handle:
+
+ - drpttracer_destory_handle()
```
derekbruening, line 119: spelling: destroy

ext/drpttracer/drpttracer.dox
```diff
@@ -88,37 +85,48 @@ the buffer size is 2^pt_size_shift * PAGE_SIZE.
...
+
+Additionally, to avoid memory leak, the client must call the following function to destroy
+the output instances after using them:
 
  - drpttracer_destroy_output()
 
+Also, the client can dump the data in the output to different files for offline
+post-processing. The user can use the library pt2ir_t in drcachesim to convert the PT
```
derekbruening, line 127: Use #

clients/drcachesim/tracer/syscall_pt_trace.cpp
```diff
@@ -0,0 +1,144 @@
...
+{
+    ASSERT(drcontext_ != nullptr, "drcontext_ is nullptr");
+    ASSERT(pttracer_handle_.handle != nullptr, "pttracer_handle_.handle is nullptr");
+
+    drpttracer_output_cleanup_last_t output;
+    if (drpttracer_stop_tracing(drcontext_, pttracer_handle_.handle, &output.data) !=
+        DRPTTRACER_SUCCESS) {
+        return false;
+    }
+    pttracer_handle_.~drpttracer_handle_cleanup_last_t();
```
derekbruening, line 103: This destroys the handle: but the .dox file says the same handle can be used across multiple sessions.  So one of these is incorrect.

clients/drcachesim/tracer/syscall_pt_trace.cpp
```diff
@@ -0,0 +1,144 @@
...
+    /* TODO i#5505: To reduce the overhead caused by IO, we better use a buffer to store
+     * multiple PT trace data. And only dump the buffer when it is full.
+     */
+
+    /* Dump PT trace data to file '{thread_id}.{syscall_id}.pt'. */
+    char pt_filename[MAXIMUM_PATH];
+    dr_snprintf(pt_filename, BUFFER_SIZE_ELEMENTS(pt_filename), "%s%s%d.%d%s",
+                pt_dir_name_, DIRSEP, dr_get_thread_id(drcontext_), recorded_syscall_num_,
+                PT_DATA_FILE_NAME_SUFFIX);
+    file_t pt_file = dr_open_file(pt_filename, DR_FILE_WRITE_REQUIRE_NEW);
```
derekbruening, line 132: We can't use this here: all i/o including opening files must be redirected through the pointers registered with drmemtrace.

clients/drcachesim/tracer/syscall_pt_trace.cpp
```diff
@@ -0,0 +1,144 @@
...
+    write_file_func_(pt_file, data->pt, data->pt_size);
+    dr_close_file(pt_file);
```
derekbruening, line 134: Ditto.  And below.

clients/drcachesim/common/options.cpp
```diff
@@ -640,3 +640,11 @@ droption_t<bool> op_enable_drstatecmp(
     DROPTION_SCOPE_CLIENT, "enable_drstatecmp", false, "Enable the drstatecmp library.",
     "When true, this option enables the drstatecmp library that performs state "
     "comparisons to detect instrumentation-induced bugs due to state clobbering.");
+
+#ifdef BUILD_PT_TRACER
+droption_t<bool> op_enable_kernel_tracing(
```
derekbruening, line 645: Please add entries to api/docs/release.dox on this option (and on the drpttrace extension library if that was not added before).  Label as still experimental.

clients/drcachesim/common/options.cpp
```diff
@@ -640,3 +640,11 @@ droption_t<bool> op_enable_drstatecmp(
...
+    DROPTION_SCOPE_ALL, "enable_kernel_tracing", false, "Enable Kernel PT tracing.",
+    "By default, offline tracing only records userspace trace. If this option is "
```
derekbruening, line 647: s/userspace/a userspace/

clients/drcachesim/common/options.cpp
```diff
@@ -640,3 +640,11 @@ droption_t<bool> op_enable_drstatecmp(
...
+    "enabled, offline tracing will record syscall's Kernel PT and write every syscall's "
```
derekbruening, line 648: s/record/record each/

derekbruening, line 648: "PT" is an Intel-specific feature and not everyone will know what  it is.  Explain that it only works on Intel processors that support the feature.

clients/drcachesim/common/trace_entry.h
```diff
@@ -369,6 +369,12 @@ typedef enum {
...
```
derekbruening, line 373: The plural "syscalls" makes this confusing: multiple are in a single marker?  Reword to "This marker is emitted prior to each system call when -enable_kernel_tracing is specified.  The marker value contains the system call identifier."

(Note that we may want to emit this when not doing kernel tracing too but we'll tweak the wording if we implement that.)

clients/drcachesim/drpt2trace/pt2ir.cpp
```diff
@@ -243,7 +247,6 @@ pt2ir_t::convert(OUT instrlist_t **ilist)
             if (status == -pte_eos)
                 break;
             dx_decoding_error(status, "sync error", insn.ip);
-            instrlist_clear_and_destroy(GLOBAL_DCONTEXT, *ilist);
```
derekbruening, line 246: The destructor of a pass-by-reference parameter is *not* called when the function exits, so this cannot be removed.  Ditto below.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1206,9 +1222,64 @@ event_pre_syscall(void *drcontext, int sysnum)
 #endif
     if (file_ops_func.handoff_buf == NULL)
         process_and_output_buffer(drcontext, false);
+#ifdef BUILD_PT_TRACER
+    if (op_offline.get_value() && op_enable_kernel_tracing.get_value()) {
+        /* The post syscall callback of SYS_exit and SYS_exit_group can't be triggered. So
+         * we skip recording the kernel PT of these syscalls.
+         */
+        if (sysnum == SYS_exit || sysnum == SYS_exit_group) {
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1206,9 +1222,64 @@ event_pre_syscall(void *drcontext, int sysnum)
...
+
+    /* Write a marker to userspace raw trace. */
+    if (BUF_PTR(data->seg_base) == NULL)
+        return; /* This thread was filtered out. */
+    trace_marker_type_t marker_type = TRACE_MARKER_TYPE_SYSCALL_ID;
+    uintptr_t marker_val = data->syscall_pt_trace.get_last_recorded_syscall_id();
+    BUF_PTR(data->seg_base) +=
+        instru->append_marker(BUF_PTR(data->seg_base), marker_type, marker_val);
+    if (file_ops_func.handoff_buf == NULL)
+        process_and_output_buffer(drcontext, false);
```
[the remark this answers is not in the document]
dolanzhao: Removed the flush code.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1352,6 +1429,21 @@ event_thread_exit(void *drcontext)
     if (BUF_PTR(data->seg_base) != NULL) {
         /* This thread was *not* filtered out. */
 
+#ifdef BUILD_PT_TRACER
+        if (op_offline.get_value() && op_enable_kernel_tracing.get_value()) {
+            int cur_recording_sysnum = data->syscall_pt_trace.get_cur_recording_sysnum();
+            if (cur_recording_sysnum != -1) {
+                NOTIFY(0,
+                       "ERROR: The last recorded syscall %d of thread T%d isn't be "
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1406,6 +1498,34 @@ event_thread_exit(void *drcontext)
...
+        /* TODO i#5505: Using the perf command to get kcore and kallsyms is not a good
+         * solution. We need to implement the kcore_copy() in syscall_pt_trace_t.
+         */
+        /* Dump kcore and kallsyms to {kernel_pt_logsubdir}. */
+#    define SHELLSCRIPT_FMT                                                          \
+        "perf record --kcore -e intel_pt/cyc,noretcomp/k echo '' >/dev/null 2>&1 \n" \
+        "chmod 755 -R perf.data \n"                                                  \
+        "cp perf.data/kcore_dir/kcore %s/ \n"                                        \
+        "cp perf.data/kcore_dir/kallsyms %s/ \n"                                     \
+        "chmod 755 -R %s \n"                                                         \
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1406,6 +1498,34 @@ event_thread_exit(void *drcontext)
...
 {
+#ifdef BUILD_PT_TRACER
+    if (op_offline.get_value() && op_enable_kernel_tracing.get_value()) {
+        drpttracer_exit();
+        /* TODO i#5505: Using the perf command to get kcore and kallsyms is not a good
+         * solution. We need to implement the kcore_copy() in syscall_pt_trace_t.
+         */
+        /* Dump kcore and kallsyms to {kernel_pt_logsubdir}. */
+#    define SHELLSCRIPT_FMT                                                          \
+        "perf record --kcore -e intel_pt/cyc,noretcomp/k echo '' >/dev/null 2>&1 \n" \
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/syscall_pt_trace.h
```diff
@@ -0,0 +1,140 @@
...
+
+private:
+    /* Dump PT trace and the metadata to files. */
+    bool
+    trace_data_dump(drpttracer_output_cleanup_last_t &output);
+
+    /* The shared file write function. */
+    ssize_t (*write_file_func_)(file_t file, const void *data, size_t count);
+
+    /* The pttracer handle holded by this instance. */
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/syscall_pt_trace.h
```diff
@@ -0,0 +1,140 @@
...
+/* The auto cleanup wrapper of pttracer handle.
+ * This can ensure the pttracer handle is cleaned up when it is out of scope.
+ */
+struct drpttracer_handle_cleanup_last_t {
+public:
+    ~drpttracer_handle_cleanup_last_t()
+    {
+        if (handle != nullptr) {
+            void *drcontext = dr_get_current_drcontext();
+            drpttracer_destory_handle(drcontext, handle);
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/syscall_pt_trace.h
```diff
@@ -0,0 +1,140 @@
...
+
+/* The auto cleanup wrapper of pttracer handle.
+ * This can ensure the pttracer handle is cleaned up when it is out of scope.
+ */
+struct drpttracer_handle_cleanup_last_t {
+public:
+    ~drpttracer_handle_cleanup_last_t()
+    {
+        if (handle != nullptr) {
+            void *drcontext = dr_get_current_drcontext();
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/syscall_pt_trace.h
```diff
@@ -0,0 +1,140 @@
...
+#include <cstddef>
+
+#include "../common/utils.h"
+#include "dr_api.h"
+#include "drpttracer.h"
+
+/* The auto cleanup wrapper of pttracer handle.
+ * This can ensure the pttracer handle is cleaned up when it is out of scope.
+ */
+struct drpttracer_handle_cleanup_last_t {
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/syscall_pt_trace.cpp
```diff
@@ -0,0 +1,144 @@
...
+    ASSERT(drcontext_ != nullptr, "drcontext_ is nullptr");
+    ASSERT(pttracer_handle_.handle != nullptr, "pttracer_handle_.handle is nullptr");
+
+    drpttracer_output_cleanup_last_t output;
+    if (drpttracer_stop_tracing(drcontext_, pttracer_handle_.handle, &output.data) !=
+        DRPTTRACER_SUCCESS) {
+        return false;
+    }
+    pttracer_handle_.~drpttracer_handle_cleanup_last_t();
+    pttracer_handle_ = {};
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/syscall_pt_trace.h
```diff
@@ -0,0 +1,140 @@
...
+ */
+class syscall_pt_trace_t {
+public:
+    syscall_pt_trace_t();
+
+    ~syscall_pt_trace_t();
+
+    /* Initialize the syscall_pt_trace_t instance for current thread.
+     * The instance will dump the kernel PT trace for every syscall. So we need to pass
+     * the output directory and the file write function to the caller.
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/syscall_pt_trace.h
```diff
@@ -0,0 +1,140 @@
...
+    stop_syscall_pt_trace();
+
+    /* Get the sysnum of current recording syscall. */
+    int
+    get_cur_recording_sysnum()
+    {
+        return cur_recording_sysnum_;
+    }
+
+    /* Get the record id of the last recorded syscall. */
```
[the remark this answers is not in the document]
dolanzhao: Add more details about what the id represents.

clients/drcachesim/tracer/syscall_pt_trace.h
```diff
@@ -0,0 +1,140 @@
...
+    trace_data_dump(drpttracer_output_cleanup_last_t &output);
+
+    /* The shared file write function. */
+    ssize_t (*write_file_func_)(file_t file, const void *data, size_t count);
+
+    /* The pttracer handle holded by this instance. */
+    drpttracer_handle_cleanup_last_t pttracer_handle_;
+
+    /* The number of recorded syscall. */
+    int recorded_syscall_num_;
```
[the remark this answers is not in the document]
dolanzhao: Done.

ext/drpttracer/drpttracer.c
```diff
@@ -123,6 +123,52 @@ read_file_to_buf(IN const char *filename, OUT char **buf, OUT uint64_t *buf_size
     return true;
 }
 
+/*
+ * This function is used to copy data from ring buffer to a output buffer.
```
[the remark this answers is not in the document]
dolanzhao: Done.

ext/drpttracer/drpttracer.c
```diff
@@ -123,6 +123,52 @@ read_file_to_buf(IN const char *filename, OUT char **buf, OUT uint64_t *buf_size
...
+    if (ring_buf_base == NULL) {
+        ASSERT(false, "try to read from NULL buffer");
+        return;
+    }
+    uint64_t data_size = head - tail;
+    if (data_size > ring_buf_size) {
+        ASSERT(
+            false,
+            "data size is larger than the ring buffer size, and old data is overwritten");
+        return;
```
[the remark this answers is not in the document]
dolanzhao: Done.

ext/drpttracer/drpttracer.c
```diff
@@ -446,57 +521,64 @@ drpttracer_exit(void)
...
         ASSERT(false, "failed to create pttracer_handle");
         return DRPTTRACER_ERROR_FAILED_TO_CREATE_PTTRACER_HANDLE;
     }
+    *(pttracer_handle_t **)tracer_handle = handle;
+    return DRPTTRACER_SUCCESS;
+}
+
+DR_EXPORT
+drpttracer_status_t
+drpttracer_destory_handle(IN void *drcontext, INOUT void *tracer_handle)
```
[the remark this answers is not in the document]
dolanzhao: Done.

ext/drpttracer/drpttracer.c
dolanzhao: Done.

ext/drpttracer/drpttracer.dox
dolanzhao: Done.

ext/drpttracer/drpttracer.dox
dolanzhao: Done.

ext/drpttracer/drpttracer.dox
dolanzhao: Done.

clients/drcachesim/tracer/syscall_pt_trace.cpp
dolanzhao: Curretly, using drpttracer_start_tracing() and drpttracer_stop_tracing() will generate a PT fragment, which may not contains PSB packets. So drpt2ir can't to decode the fragment successful. In future, if I upgrade drpt2ir to support decoding the PT fragment, I will update this part.

clients/drcachesim/common/options.cpp
dolanzhao: Done.

clients/drcachesim/common/options.cpp
dolanzhao: Done.

dolanzhao: Done.

clients/drcachesim/common/trace_entry.h
dolanzhao: Done.

clients/drcachesim/drpt2trace/pt2ir.cpp
dolanzhao: Done.

clients/drcachesim/tracer/syscall_pt_trace.cpp
dolanzhao: Done.

clients/drcachesim/tracer/syscall_pt_trace.cpp
dolanzhao: Done.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1206,9 +1222,64 @@ event_pre_syscall(void *drcontext, int sysnum)
 #endif
     if (file_ops_func.handoff_buf == NULL)
         process_and_output_buffer(drcontext, false);
+#ifdef BUILD_PT_TRACER
+    if (op_offline.get_value() && op_enable_kernel_tracing.get_value()) {
+        /* The post syscall callback of SYS_exit and SYS_exit_group can't be triggered. So
+         * we skip recording the kernel PT of these syscalls.
```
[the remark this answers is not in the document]
dolanzhao: I didn't find any introduction about how to implement this feature. If I found it, I will update this part.

clients/drcachesim/common/options.cpp
dolanzhao: Done.

ext/drpttracer/drpttracer.dox
dolanzhao: Done.

ext/drpttracer/drpttracer.dox
dolanzhao: Done.

clients/drcachesim/tracer/syscall_pt_trace.cpp
```diff
@@ -0,0 +1,200 @@
...
+{
+    ASSERT(drcontext_ != nullptr, "drcontext_ is nullptr");
+    ASSERT(pttracer_handle_.handle != nullptr, "pttracer_handle_.handle is nullptr");
+
+    drpttracer_output_autoclean_t output = { drcontext_, nullptr };
+    if (drpttracer_stop_tracing(drcontext_, pttracer_handle_.handle, &output.data) !=
+        DRPTTRACER_SUCCESS) {
+        return false;
+    }
+    pttracer_handle_.reset();
```
abhinav92003, line 116: Add the TODO on optimizing here too.

api/docs/CMake_doxyfile.cmake
```diff
@@ -56,6 +56,8 @@ endforeach (dir)
```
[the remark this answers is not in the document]
abhinav92003: Also update the PR description with details of the pending optimization.

api/docs/release.dox
```diff
@@ -156,6 +156,13 @@ Further non-compatibility-affecting changes include:
    online filtering for only instruction or only data entries respectively. The
    old option -L0_filter is deprecated but still supported for backward
    compatibility. It simply sets both the new options.
+ - Added a new DR extension, namely "drpttracer", which rovides clients with tracing
```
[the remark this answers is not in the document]
dolanzhao: Done.

api/docs/release.dox
```diff
@@ -156,6 +156,13 @@ Further non-compatibility-affecting changes include:
...
+   functionality via Intel's PT instruction tracing feature. This feature is still
+   experimental.
```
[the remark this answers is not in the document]
dolanzhao: Done.

api/docs/release.dox
```diff
@@ -156,6 +156,13 @@ Further non-compatibility-affecting changes include:
...
+ - Added new drmemtrace options -enable_kernel_tracing that allow recording each
+   syscall's Kernel PT and write every syscall's PT and metadata to files in
+   -outdir/kernel.raw/ for later offline analysis. This feature is still experimental
+   and depended on the Intel CPU that supports Intel@ Processor Trace.
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/common/options.cpp
```diff
@@ -640,3 +640,13 @@ droption_t<bool> op_enable_drstatecmp(
...
     "When true, this option enables the drstatecmp library that performs state "
     "comparisons to detect instrumentation-induced bugs due to state clobbering.");
+
+#ifdef BUILD_PT_TRACER
+droption_t<bool> op_enable_kernel_tracing(
+    DROPTION_SCOPE_ALL, "enable_kernel_tracing", false, "Enable Kernel Intel PT tracing.",
+    "By default, offline tracing only records a userspace trace. If this option is "
+    "enabled, offline tracing will record each syscall's Kernel PT and write every "
+    "syscall's PT and metadata to files in -outdir/kernel.raw/ for later offline "
+    "analysis. And this feature is depended on the Intel CPU that supports Intel@ "
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/common/trace_entry.h
```diff
@@ -369,6 +369,12 @@ typedef enum {
      */
     TRACE_MARKER_TYPE_PAGE_SIZE,
 
+    /**
+     * This marker is emitted prior to each system call when -enable_kernel_tracing is
+     * specified. The marker value contains the system call identifier.
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/syscall_pt_trace.h
```diff
@@ -0,0 +1,177 @@
...
+    {
+        if (drcontext != nullptr && data != nullptr) {
+            void *drcontext = dr_get_current_drcontext();
+            drpttracer_destroy_output(drcontext, data);
+            data = nullptr;
+        }
+    }
+    void *drcontext = nullptr;
+    drpttracer_output_t *data = nullptr;
+};
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/syscall_pt_trace.h
```diff
@@ -0,0 +1,177 @@
...
+struct drpttracer_output_autoclean_t {
+public:
+    drpttracer_output_autoclean_t(void *drcontext, drpttracer_output_t *data)
+        : drcontext { drcontext }
+        , data { data }
+    {
+    }
+    ~drpttracer_output_autoclean_t()
+    {
+        if (drcontext != nullptr && data != nullptr) {
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1217,9 +1230,60 @@ event_pre_syscall(void *drcontext, int sysnum)
 #endif
     if (file_ops_func.handoff_buf == NULL)
         process_and_output_buffer(drcontext, false);
+#ifdef BUILD_PT_TRACER
+    if (op_offline.get_value() && op_enable_kernel_tracing.get_value()) {
+        if (!syscall_pt_trace_t::is_syscall_pt_trace_enabled(sysnum)) {
+            return true;
+        }
+
+        if (data->syscall_pt_trace.get_cur_recording_sysnum() != -1) {
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1217,9 +1230,60 @@ event_pre_syscall(void *drcontext, int sysnum)
...
     return true;
 }
 
+static void
+event_post_syscall(void *drcontext, int sysnum)
+{
+#ifdef BUILD_PT_TRACER
+    if (tracing_disabled.load(std::memory_order_acquire) != BBDUP_MODE_TRACE)
+        return;
+    if (!op_offline.get_value() || !op_enable_kernel_tracing.get_value())
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1217,9 +1230,60 @@ event_pre_syscall(void *drcontext, int sysnum)
...
+
+    per_thread_t *data = (per_thread_t *)drmgr_get_tls_field(drcontext, tls_idx);
+
+    if (data->syscall_pt_trace.get_cur_recording_sysnum() == -1) {
+        ASSERT(false, "last syscall is not traced");
+        return;
+    }
+
+    ASSERT(data->syscall_pt_trace.get_cur_recording_sysnum() == sysnum,
+           "last tracing isn't stopped");
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1363,6 +1434,21 @@ event_thread_exit(void *drcontext)
     if (BUF_PTR(data->seg_base) != NULL) {
         /* This thread was *not* filtered out. */
 
+#ifdef BUILD_PT_TRACER
+        if (op_offline.get_value() && op_enable_kernel_tracing.get_value()) {
+            int cur_recording_sysnum = data->syscall_pt_trace.get_cur_recording_sysnum();
+            if (cur_recording_sysnum != -1) {
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1363,6 +1434,21 @@ event_thread_exit(void *drcontext)
...
+                NOTIFY(0,
+                       "ERROR: The last recorded syscall %d of thread T%d wasn't be "
+                       "stopped.\n",
+                       cur_recording_sysnum, dr_get_thread_id(drcontext));
+                ASSERT(cur_recording_sysnum, "syscall recording is not stopped");
+                FATAL("failed to stop syscall pt trace(sysnum=%d)\n",
```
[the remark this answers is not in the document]
dolanzhao: Done.

ext/drpttracer/drpttracer.c
```diff
@@ -123,6 +130,60 @@ read_file_to_buf(IN const char *filename, OUT char **buf, OUT uint64_t *buf_size
...
+    READ_RING_BUFFER_ERROR,
+    READ_RING_BUFFER_ERROR_INVALID_PARAMETER,
+    READ_RING_BUFFER_ERROR_OLD_DATA_OVERWRITTEN
+} read_ring_buf_status_t;
+
+/*
+ * This function is used to copy data from ring buffer to an output buffer.
+ * ring buffer:
+ *  if head_offs > tail_offs:
+ *   vaild data = [tail_offs, head_offs]
```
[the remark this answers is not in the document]
dolanzhao: Done.

ext/drpttracer/drpttracer.dox
```diff
@@ -88,37 +85,47 @@ the buffer size is 2^pt_size_shift * PAGE_SIZE.
...
-status code:
+which may cause issues in #pt2ir_t decoding.
+
+After creating a pttracer handle successfully, the client can start or stop a tracing
+session by calling drpttracer_start_tracing() and drpttracer_stop_tracing().
+
+ - drpttracer_start_tracing()
+ - drpttracer_stop_tracing()
+
+The client must pass an valid trace handle to the start and stop function. And the client
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/syscall_pt_trace.cpp
dolanzhao: Done.

api/docs/CMake_doxyfile.cmake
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/syscall_pt_trace.cpp
```diff
@@ -113,6 +114,10 @@ syscall_pt_trace_t::stop_syscall_pt_trace()
         DRPTTRACER_SUCCESS) {
         return false;
     }
+
+    /* TODO i#5505: If we can share the same pttracer handle for all syscalls, we can
+     * avoid to reset the handle.
```
abhinav92003, line 119: s/to reset/resetting/

abhinav92003 (approved):

clients/drcachesim/tracer/syscall_pt_trace.cpp
dolanzhao: Done.

Issue #2062, opened 2016-11-09
add non-module and vdso support to offline traces


derekbruening:
Split from #1729 

What about DGC from a JIT, or vsyscall?

We need to store the type (read/write/prefetch*) and size for a memref, and
the size of every single ifetch.

For memrefs, we may have to use double entries, with the first one having
an escape type and the address and the second having the full type and
size.

For ifetch, perhaps we can fit multiple sizes in one entry.  Modidx will be
a sentinel, so we have the 45 modoffs bits.  We will want 4 bits for each
instr size (ok to assume 16-byte max instr, even though technically there
can be 17 and an illegal instr could be even bigger) so we can fit the
sizes of 11 instrs there.  We'll zero the rest.

The vdso shows up as a module but is harder to decode.
We have two choices for vdso:
1) Treat as non-module and record type info in the trace.  The tracer will need
   to look for "[vdso]".
2) Assume raw2trace is run on a machine w/ the same vdso code and try to
   decode from its own vdso via special code in raw2trace.

derekbruening:
3) Dump full contents of vdso somewhere and store that path in the module list

derekbruening:
fd63caa7286f2f0dd239cbd8de07a264d5a4e82c from PR #2706 added vdso content dumping and post-processing support.  

General DGC is still unsupported for offline.

prasun3:
Can you give me some pointers as to what would be needed to get this working?

Also, should we indicate (via warnings etc) that jitted code is not supported in offline traces. Right now the only indication I can see is this message in debug builds:

`SYSLOG_INTERNAL_WARNING_ONCE("writing to executable region.");`

derekbruening:
That `writing to executable region` is from the core, not the tracer.
It does seem surprising that the tracer doesn't print any warning, and the post-processor has a verbose-3 warning is all:
```
$ git grep -A 5 '#2062'
clients/drcachesim/tracer/instru_offline.cpp:        // FIXME i#2062: add non-module support.  The plan for instrs is to have
clients/drcachesim/tracer/instru_offline.cpp-        // one entry w/ the start abs pc, and subsequent entries that pack the instr
clients/drcachesim/tracer/instru_offline.cpp-        // length for 10 instrs, 4 bits each, into a pc.modoffs field.  We will
clients/drcachesim/tracer/instru_offline.cpp-        // also need to store the type (read/write/prefetch*) and size for the
[...]

PR #5599, opened 2022-08-08
i#2062 memtrace nonmod: Add test for generated code

derekbruening:
Adds a drmemtrace test which generates code outside of any module.

Adds a proper warning about this scenario at level 0 (with a do-once
to avoid spew) to both the tracer and raw2trace (previously it would
silently be traced without any indication of a problem, and the
raw2trace warning was at level 1).

Issue: #2062

### DIFF
--- a/.github/workflows/ci-clang-format.yml
+++ b/.github/workflows/ci-clang-format.yml
@@ -72,7 +72,7 @@ jobs:
     - name: Create Build Environment
       run: |
         sudo apt-get -y install doxygen vera++ zlib1g-dev libsnappy-dev \
-          liblz4-dev clang-format-9 libunwind-dev
+          liblz4-dev clang-format-12 libunwind-dev
 
     - name: Run Suite
       working-directory: ${{ github.workspace }}
--- a/clients/drcachesim/tracer/tracer.cpp
+++ b/clients/drcachesim/tracer/tracer.cpp
@@ -169,6 +169,17 @@ clean_call(void)
     process_and_output_buffer(drcontext, false);
 }
 
+void
+instru_notify(uint level, const char *fmt, ...)
+{
+    if (op_verbose.get_value() < level)
+        return;
+    va_list args;
+    va_start(args, fmt);
+    dr_vfprintf(STDERR, fmt, args);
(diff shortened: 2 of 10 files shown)


.github/workflows/ci-clang-format.yml
```diff
@@ -72,7 +72,7 @@ jobs:
     - name: Create Build Environment
       run: |
         sudo apt-get -y install doxygen vera++ zlib1g-dev libsnappy-dev \
-          liblz4-dev clang-format-9 libunwind-dev
+          liblz4-dev clang-format-12 libunwind-dev
```
abhinav92003, line 75: Mention this upgrade in the PR's description too?

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -169,6 +169,17 @@ clean_call(void)
     process_and_output_buffer(drcontext, false);
 }
 
+void
+instru_notify(uint level, const char *fmt, ...)
```
abhinav92003, line 173: Curious why this couldn't just be a part of instru_offline.cpp

abhinav92003 (approved):

clients/drcachesim/tracer/tracer.cpp
derekbruening: Today instru_offline.cpp has some nice isolation from the rest of the tracer: it does not include the options.h or utils.h and is a pure library that does not share any of the configuration or reporting infrastructure specific to the tracer.  It seemed nice to preserve that.  It is used by raw2trace as well and raw2trace consequently does not have to pull in the tracer's options/logging code and deal with it or have to stub it out like we have to do with libraries we've made from the core because we don't have great isolation everywhere (xref those weak symbol complexities).

.github/workflows/ci-clang-format.yml
derekbruening: +1, will do.  I considered doing a separate PR.  One of the files changed here didn't have the function return type on its own line in clang-format-9 and my laptop has 14 and can't easily install 9 so it seemed best to just upgrade.

PR #5601, opened 2022-08-09
i#2626: AArch64 v8.2 codec: Update fcsel to use a predicate

joshua-warburton:
This patch updates fcsel so that it sets the condition
parameter as a predicate and joins it to the mnemonic
when printing as debug. Also added are tests and
instr_create macros.

Issue: #2626


joshua-warburton:
run arm tests

joshua-warburton:
run arm tests

PR #5605, opened 2022-08-11
i#3544 RISC-V, part 9: Add arch files

semihalf-kardach-stanislaw:
Add files with arch-specific symbols and definitions required to compile
for RISC-V.
Those definitions are used in multiple DynamoRIO modules but splitting
them across multiple commits would likely involve merge dependencies.
Therefore it seemed more straightforward to keep all of the files under
core/arch/riscv64 in a single commit.

NOTE:

Issue: #3544


abhinav92003:
Since Derek is on vacation today, I can help review this one.

core/arch/riscv64/clean_call_opt.c
```diff
@@ -0,0 +1,71 @@
+#include "../globals.h"
```
abhinav92003, line 1: This is missing the copyright banner.

core/arch/riscv64/mangle.c
```diff
@@ -0,0 +1,219 @@
...
+{
+    /* FIXME i#3544: Not implemented */
+    ASSERT_NOT_IMPLEMENTED(false);
+    return false;
+}
+
+/*###########################################################################
+ *###########################################################################
+ *
+ *   M A N G L I N G   R O U T I N E S
```
abhinav92003, line 110: Modify to exclude the extra spaces between characters. So that it fits in with other DR code's style.

core/arch/riscv64/proc.c
```diff
@@ -0,0 +1,257 @@
...
+    size_t sz;
+
+    if (dcache_line_size) {
+        ATOMIC_8BYTE_ALIGNED_READ(&dcache_line, &sz);
+        if (sz == 0) {
+            sz = read_cache_line(d_cache_fname);
+            if (sz > 0)
+                ATOMIC_8BYTE_ALIGNED_WRITE(&dcache_line, &sz, false);
+        }
+        ATOMIC_8BYTE_ALIGNED_READ(&dcache_line, &sz);
```
abhinav92003, line 93: We don't need this. When sz != 0 at L88, we already have the required value. When sz == 0, sz was set to the value returned by read_cache_line already.

core/arch/riscv64/proc.c
```diff
@@ -0,0 +1,257 @@
...
+        if (sz != 0)
+            *dcache_line_size = sz;
+    }
+
+    if (icache_line_size) {
+        if (icache_line == 0) {
```
abhinav92003, line 99: Do we need atomic reads/writes here?

core/arch/riscv64/proc.c
```diff
@@ -0,0 +1,257 @@
...
+            sz = read_cache_line(i_cache_fname);
+            if (sz > 0)
+                icache_line = sz;
+        }
+        if (icache_line != 0)
+            *icache_line_size = icache_line;
+    }
+
+    return true;
```
abhinav92003, line 108: We should return false when the read cache line size was zero, and we skipped writing it to the result.

core/arch/riscv64/riscv64.asm
```diff
@@ -0,0 +1,316 @@
...
+GLOBAL_LABEL(call_switch_stack:)
+/* FIXME i#3544: Not implemented */
+        ret
+        END_FUNC(call_switch_stack)
+
+/*
+ * Calls the specified function 'func' after switching to the DR stack
+ * for the thread corresponding to 'drcontext'.
+ * Passes in 8 arguments.  Uses the C calling convention, so 'func' will work
+ * just fine even if if takes fewer than 8 args.
```
abhinav92003, line 70: s/even if if/even if it/

core/arch/riscv64/clean_call_opt.c
semihalf-kardach-stanislaw: Done.

core/arch/riscv64/mangle.c
semihalf-kardach-stanislaw: Done. I've copied this straight from x86 and aarchxx versions of this file. Those will also need aligning I guess.

core/arch/riscv64/proc.c
semihalf-kardach-stanislaw: Correct.

core/arch/riscv64/proc.c
semihalf-kardach-stanislaw: Not at all. I've missed fixing the dcache logic when I updated icache to normal writes.

core/arch/riscv64/proc.c
semihalf-kardach-stanislaw: Done. I still return true if both output pointers were NULL (even though it seems a very degraded case).

core/arch/riscv64/riscv64.asm
semihalf-kardach-stanislaw: Done.

abhinav92003 (approved):

core/arch/riscv64/mangle.c
abhinav92003: We can optionally fix those old files in this PR, but I'm fine either way.

semihalf-kardach-stanislaw: Done. Better do it now than forget in the future. :)

abhinav92003: Thanks!

abhinav92003:
The x86-32 failure is a known flake: #5329. Should go away on rerunning the suite.

PR #5608, opened 2022-08-12
i#3544 RISC-V, part 10: Add various core stubs

semihalf-kardach-stanislaw:
Add a minimal implementation of symbols required to get the core of
DynamoRIO compiling.
This changeset contains code from multiple modules and was not separated
as it would create merge dependencies for code charing the same file.
Additionally those definitions are stubs for compilation purposes, so
there doesn't seem to be much to be gained from the separation.

NOTE:
  was to achieve compilation. Some of the trivial logic has been
  implemented though. Proper implementation of stubs will follow in
  later commits.

Issue: #3544


core/unix/os_public.h
```diff
@@ -169,6 +169,9 @@ typedef kernel_sigcontext_t sigcontext_t;
 #    define SC_XSP SC_FIELD(sp)
 #    define SC_XFLAGS SC_FIELD(pstate)
 #    define SC_SYSNUM_REG SC_FIELD(regs[8])
+/* FIXME i#3544: RFC: Why is SC_RETURN_REG not defined for AArch64? It would
```
abhinav92003, line 172: Re RFC: I think x0 to x7 are used to return the result in the AArch64 calling convention: https://developer.arm.com/documentation/den0024/a/The-ABI-for-ARM-64-bit-Architecture/Register-use-in-the-AArch64-Procedure-Call-Standard/Parameters-in-general-purpose-registers. So we don't have a single register to define SC_RETURN_REG.

core/unix/signal.c
```diff
@@ -3946,18 +4031,24 @@ transfer_from_sig_handler_to_fcache_return(dcontext_t *dcontext, kernel_ucontext
 
 #if defined(X64) || defined(ARM)
     /* x64 always uses shared gencode */
-    dcontext->local_state->spill_space.IF_X86_ELSE(xax, r0) =
-        sc->IF_X86_ELSE(SC_XAX, SC_R0);
+    /* FIXME i#3544: RFC: Perhaps IF_<arch>_ELSE ladder should be replaced by
```
abhinav92003, line 4034: I guess we do have things like SC_SYSNUM_REG and SC_RETURN_REG where it makes sense. Is there one you suggest here?

core/unix/os.c
```diff
@@ -2454,12 +2518,19 @@ os_thread_init(dcontext_t *dcontext, void *os_data)
     }
 #endif
 
+    /* FIXME i#3544: RFC: this log essentially prints DR TLS registers but
```
abhinav92003, line 2521: By "keeping order consistent" you mean between the two log statements?

The current modifications you made look fine. Not sure what the suggestion meant to improve?

core/arch/asm_defines.asm
```diff
@@ -851,6 +855,28 @@ ASSUME fs:_DATA @N@\
         mov      ARG2, p2   @N@\
         mov      ARG1, p1   @N@\
         blx      callee
+#elif defined(RISCV64)
+# define CALLC0(callee)    \
+        call      callee
+/* FIXME i#3544: Handle p1..4 being registers instead of immediates. */
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: No. It's a RISC architecture so opcode defines the whole encoding. That's why there is `ADD` for reg+reg add and `ADDI` for reg+imm.

core/arch/asm_defines.asm
```diff
@@ -903,6 +929,16 @@ ASSUME fs:_DATA @N@\
 # endif
 # define INC(reg) add reg, reg, POUND 1
 # define DEC(reg) sub reg, reg, POUND 1
+#elif defined(RISCV64)
+# define REG_SCRATCH0 REG_A0
+# define REG_SCRATCH1 REG_A1
+# define REG_SCRATCH2 REG_A2
+# define JUMP     j
+/* FIXME i#3544: RFC: There is no state register with compare flags. Emulate? */
+# define JUMP_NOT_EQUAL     bnez x5,
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Not really. Looking at the code again this macro is only used in the `thread_check_gprs_from_cache` test in `suite/tests/api/detach_state.c`:
```
check_gprs_from_cache_spin:
        CHECK_SIDELINE_EXIT
        JUMP_NOT_EQUAL check_gprs_from_cache_spin
        PUSHALL
```
The `CHECK_SIDELINE_EXIT` is defined in the same file.
Given that I see 2 solutions:
1. Remove `JUMP_NOT_EQUAL` from `asm_defines.asm` and define it straight in `suite/tests/api/detach_state.c`. This will remove the issue for repo-internal code.
2. Modify the `JUMP_NOT_EQUAL -> JUMP_NOT_EQUAL(reg_lhs, reg_rhs, label)`.

Both will be problematic if `asm_defines.asm` is an API file for extensions. If it is though, I think we're missing a `COMPARE(lhs,rhs,cmp)` macro.

I'll go with the first approach for now.

core/arch/emit_utils_shared.c
```diff
@@ -1001,7 +1005,7 @@ relocate_patch_list(dcontext_t *dcontext, patch_list_t *patch, instrlist_t *ilis
                 if (opnd_is_near_base_disp(opnd)) {
                     /* indirect through XDI and update displacement */
                     opnd_set_disp(&opnd, (int)patch->entry[cur].value_location_offset);
-                    opnd_replace_reg(&opnd, REG_NULL, SCRATCH_REG5 /*xdi/r5*/);
+                    opnd_replace_reg(&opnd, REG_NULL, SCRATCH_REG5 /*xdi/r5/a5*/);
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Yes, it was defined in part 1 of the RISC-V pull requests. More precisely:
```
#    define SCRATCH_REG5 DR_REG_A5
```

core/unix/signal.c
```diff
@@ -2890,6 +2928,44 @@ mcontext_to_sigcontext(sig_full_cxt_t *sc_full, priv_mcontext_t *mc,
...
+        sc->SC_FIELD(sc_regs.s1) = mc->s1;
+        sc->SC_FIELD(sc_regs.a0) = mc->a0;
+        sc->SC_FIELD(sc_regs.a1) = mc->a1;
+        sc->SC_FIELD(sc_regs.a2) = mc->a2;
+        sc->SC_FIELD(sc_regs.a3) = mc->a3;
+        sc->SC_FIELD(sc_regs.a4) = mc->a4;
+        sc->SC_FIELD(sc_regs.a5) = mc->a5;
+        sc->SC_FIELD(sc_regs.a6) = mc->a6;
+        sc->SC_FIELD(sc_regs.a7) = mc->a7;
+        sc->SC_FIELD(sc_regs.s2) = mc->s2;
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: I was following the order in `struct user_regs_struct` from `sigcontext.h`.

core/drlibc/drlibc_riscv64.asm
```diff
@@ -0,0 +1,67 @@
...
+/*
+ * ptr_int_t dynamorio_syscall(uint sysnum, uint num_args, ...);
+ *
+ * Linux riscv64 system call:
+ * - a7: syscall number
+ * - a0..a5: syscall arguments
+ */
+        DECLARE_FUNC(dynamorio_syscall)
+GLOBAL_LABEL(dynamorio_syscall:)
+        mv      a7,a0
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Nice catch, thanks!

core/arch/interp.c
```diff
@@ -6555,7 +6571,8 @@ append_trace_speculate_last_ibl(dcontext_t *dcontext, instrlist_t *trace,
 
 #ifdef HASHTABLE_STATISTICS
     DOSTATS({
-        reg_id_t reg = IF_X86_ELSE(REG_XCX, DR_REG_R2);
+        /* FIXME i#3544: RFC: Is this SCRATCH_REG2 used on IBL or XCX/R2? */
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: OK, then this whole `IF_*_ELSE` ladder can be replaced by `SCRATCH_REG2`.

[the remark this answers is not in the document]
semihalf-kardach-stanislaw: OK, I'll replace the `IF_*_ELSE` ladders then. However there is a function called `instr_is_tls_xcx_spill()` which probably should be renamed to `instr_is_tls_scratch_reg2_spill()`?

core/arch/interp.c
```diff
@@ -6643,7 +6660,8 @@ append_ib_trace_last_ibl_exit_stat(dcontext_t *dcontext, instrlist_t *trace,
 
     instr_t *inst = instrlist_last(trace); /* currently only relevant to last CTI */
     instr_t *where = inst;                 /* preinsert before exit CTI */
-    reg_id_t reg = IF_X86_ELSE(REG_XCX, DR_REG_R2);
+    /* FIXME i#3544: RFC: Is this SCRATCH_REG2 used on IBL or XCX/R2? */
+    reg_id_t reg = IF_X86_ELSE(REG_XCX, IF_RISCV64_ELSE(DR_REG_A2, DR_REG_R2));
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Done.

core/translate.c
```diff
@@ -1972,10 +1978,13 @@ stress_test_recreate_state(dcontext_t *dcontext, fragment_t *f, instrlist_t *ili
             }
 
             if (spill_ibreg_outstanding_offs != UINT_MAX) {
-                mc.IF_X86_ELSE(xcx, r2) =
+                /* FIXME i#3544: RFC: Perhaps we need a union field for IBL register in
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Not necessarily. Consider how LR is defined [in `mxatx_api.h`](https://github.com/DynamoRIO/dynamorio/blob/master/core/lib/mcxtx_api.h#L87):
```c
    union {
        reg_t r30; /**< The r30 register. \note For 64-bit DR builds only. */
        reg_t lr;  /**< The link register. */
    }; /**< The anonymous union of alternative names for r30/lr register. */
```
What I suggest is to have a similar union around `xcx` and `r2`, so for X86:
```c
    union {
        reg_t xcx; /**< The platform-independent name for full rcx/ecx register. */
        reg_t IF_X64_ELSE(rcx, ecx); /**< The platform-dependent name for
                                          rcx/ecx register. */
        reg_t ibl; /**< The platform-independent name for the Indirect Branch Lookup target register */
    }; /**< The anonymous union of alternative names for rcx/ecx register. */
```
For ARM:
```c
    union {
        reg_t r2;   /**< The r2 register. */
        reg_t ibl; /**< The platform-independent name for the Indirect Branch Lookup target register */
    }; /**< The anonymous union of alternative names for r2 register. */
```
And for RISC-V:
```c
    union {
        reg_t x12; /**< The x12 register. */
        reg_t a2;  /**< The 3rd argument register. */
        reg_t ibl; /**< The platform-independent name for the Indirect Branch Lookup target register */
    };  /**< The anonymous union of alternative names for the x12/a2 register. */
```

Using the above line 1975 in `translate.c` would look like this:
```c
mc.ibl = (reg_t)d_r_get_tls(spill_ibreg_outstanding_offs) + 1;
```
I'll adapt to this change in the next version to show exactly what I mean.

core/unix/os.c
```diff
@@ -1639,7 +1639,67 @@ os_timeout(int time_in_milliseconds)
...
+                                 : "=r"(var)                           \
+                                 : "i"(DR_TLS_BASE_OFFSET), "i"(imm)); \
+        } while (0)
+#    define WRITE_TLS_SLOT(offs, var)                                           \
+        do {                                                                    \
+            IF_NOT_HAVE_TLS(ASSERT_NOT_REACHED());                              \
+            ASSERT(sizeof(var) == sizeof(void *));                              \
+            __asm__ __volatile__("ld t0, %0(tp) \n\t"                           \
+                                 "add t0, t0, %2\n\t"                           \
+                                 "sd %1, 0(t0) \n\t"                            \
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Yes. RISC-V does not have a store instruction that takes a displacement/offset register, only a 12 bit immediate value.

core/unix/os.c
```diff
@@ -1639,7 +1639,67 @@ os_timeout(int time_in_milliseconds)
...
+                                 : "i"(DR_TLS_BASE_OFFSET), "r"(var), "r"(offs) \
+                                 : "memory", "t0");                             \
+        } while (0)
+#    define READ_TLS_SLOT(offs, var)                                    \
+        do {                                                            \
+            IF_NOT_HAVE_TLS(ASSERT_NOT_REACHED());                      \
+            ASSERT(sizeof(var) == sizeof(void *));                      \
+            __asm__ __volatile__("ld %0, %1(tp) \n\t"                   \
+                                 "add %0, %0, %2\n\t"                   \
+                                 "ld %0, 0(%0) \n\t"                    \
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Ditto.

core/unix/os.c
semihalf-kardach-stanislaw: > By "keeping order consistent" you mean between the two log statements?

Yes. Current code will always print `gs` first and `fs` second. If that is not required, then in `os_exports.h` we could have:
```c
#ifdef X86
#    if defined(MACOS64)
// ...
#        define STR_SEG "gs"
#        define STR_LIB_SEG "fs"
#    elif defined(X64)
// ...
#        define STR_SEG "gs"
#        define STR_LIB_SEG "fs"
#    else
// ...
#        define STR_SEG "fs"
#        define STR_LIB_SEG "gs"
#    endif
#elif defined(AARCHXX)
#    ifdef X64
// ...
#        define STR_SEG "tpidruro"
#        define STR_LIB_SEG "tpidrurw"
#    else
// ...
#        define STR_SEG "tpidrurw"
#        define STR_LIB_SEG "tpidruro"
#    endif                                  /* 64/32-bit */
#elif defined(RISCV64)
// ...
#    define STR_SEG "tp"
#    define STR_LIB_SEG "s11"
#endif /* X86/ARM */
```
And in `os.c` this logs would be:
```c
    LOG(THREAD, LOG_THREADS, 1, "post-TLS-setup, cur %s base is " PFX "\n",
        STR_SEG, get_segment_base(SEG_TLS));
    LOG(THREAD, LOG_THREADS, 1, "post-TLS-setup, cur %s base is " PFX "\n",
        STR_LIB_SEG, get_segment_base(LIB_SEG_TLS));
```

core/unix/os_public.h
semihalf-kardach-stanislaw: Right, but those defines are for system calls and Linux's `syscall()` returns `long`, which is **usually** within word size. Note that `SC_RETURN_REG` is defined for `ARM` (line 195), even though the procedure call ABI specifies that [several registers may be used to return values](https://developer.arm.com/documentation/den0013/d/Application-Binary-Interfaces/Procedure-Call-Standard/Returning-results). In general the return value is written into the first of the return registers (`x0` on ARM, `a0` on RISC-V) when it fits (so anything up to word size) and the rest goes to the next register (i.e. 128bit return would go to `x0` and `x1`).

core/arch/interp.c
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Ditto

core/arch/interp.c
```diff
@@ -7616,10 +7634,14 @@ decode_fragment(dcontext_t *dcontext, fragment_t *f, byte *buf, /*IN/OUT*/ uint
                         raw_start_pc = prev_pc;
                     }
                     /* now append our new xcx save */
-                    instrlist_append(ilist,
-                                     instr_create_save_to_dcontext(
-                                         dcontext, IF_X86_ELSE(REG_XCX, DR_REG_R2),
-                                         IF_X86_ELSE(XCX_OFFSET, R2_OFFSET)));
+                    /* FIXME i#3544: RFC: Is this SCRATCH_REG2 used on IBL or XCX/R2? */
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Ditto.

core/translate.c
```diff
@@ -407,6 +408,9 @@ translate_walk_track_post_instr(dcontext_t *tdcontext, instr_t *inst,
               (!opnd_is_pc(instr_get_target(inst)) ||
                (opnd_get_pc(instr_get_target(inst)) >= walk->start_cache &&
                 opnd_get_pc(instr_get_target(inst)) < walk->end_cache))))
+#elif defined(RISCV64)
+            /* FIXME i#3544: Not implemented */
+            true
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Done. I've also changed the condition to `false` to make it more apparent this code is not implemented.

core/unix/signal.c
semihalf-kardach-stanislaw: Since return value in case of a function or syscall follows the same ABI (x86 - xax, ARM - R0, RISC-V - A0), I've added a `retval` register alias in mcxtx and spill_space. Please let me know if you prefer the union approach or rather a define based one as I've described in [one of the fix commits](https://github.com/DynamoRIO/dynamorio/pull/5608/commits/51162039cb20613a4baeca89d64af858eb6da63b).

core/unix/signal.c
```diff
@@ -4392,7 +4496,8 @@ adjust_syscall_for_restart(dcontext_t *dcontext, thread_sig_info_t *info, int si
     byte *pc = (byte *)sc->SC_XIP;
     int sys_inst_len;
 
-    if (sc->IF_X86_ELSE(SC_XAX, SC_R0) != -EINTR) {
+    /* FIXME i#3544: RFC: Why not use SC_RETURN_REG below? */
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Ditto.

semihalf-kardach-stanislaw:
@abhinav92003 I've left the conversations which contain open topics unresolved so that you can indicate whether my approach is correct/acceptable or not.

core/arch/asm_defines.asm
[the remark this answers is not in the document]
abhinav92003: Okay, the FIXME should suffice for now then.

core/translate.c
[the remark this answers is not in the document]
abhinav92003: I think core/lib/mcxtx_api.h is expected to define the machine's context only, independent of how DR is using it. By adding the definitions you suggested, we're including details about how DR is using the available regs to store the IBL target.

IMHO defining the `IBL(mc)` macro (in source outside the header) is a better way to accomplish code simplicity.

core/unix/os_public.h
abhinav92003: Okay, I see on https://man7.org/linux/man-pages/man2/syscall.2.html that even x86 technically has two return regs, but os_public.h still defines SC_RETURN_REG.

Defining it for AArch64 sounds good to me.

core/arch/interp.c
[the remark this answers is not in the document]
abhinav92003: I see `instr_is_tls_xcx_spill` is currently not implemented on non-x86. We can defer the rename to when we do implement it.

core/unix/signal.c
abhinav92003: Same as the other comment, I believe it'll be better to keep those structs independent of DR's implementation details.

core/translate.c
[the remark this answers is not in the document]
abhinav92003: Since this came up in another comment thread, I think it'll be good to get @derekbruening 's opinion. I believe it'll be better to keep structs like spill_state_t and mcxtx independent of DR's implementation details.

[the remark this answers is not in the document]
semihalf-kardach-stanislaw: I agree that macros will leave the structs more readable at the first glance and also containing only HW-dependent fields.
It'll be then easier to move the arch-specific functional macros to `core/arch/<arch>/some_header.h`.

core/unix/os_exports.h
```diff
@@ -78,16 +78,22 @@
 #    if defined(MACOS64)
 #        define SEG_TLS SEG_GS     /* DR is sharing the app's segment. */
 #        define LIB_SEG_TLS SEG_GS /* libc+loader tls */
+#        define STR_SEG "gs"
+#        define STR_LIB_SEG "fs"
```
abhinav92003, line 82: Based on LIB_SEG_TLS, shouldn't this be gs?

core/unix/os_exports.h
```diff
@@ -99,16 +105,22 @@
...
-#    endif                                  /* 64/32-bit */
+#        define STR_SEG "tpidrurw"
+#        define STR_LIB_SEG "tpidruro"
+#    endif /* 64/32-bit */
 #elif defined(RISCV64)
 /* FIXME i#3544: Not used on RISC-V, so set to invalid. Check if this is true. */
 #    define SEG_TLS DR_REG_INVALID
 #    define LIB_SEG_TLS DR_REG_TP
+#    define STR_SEG "tp"
+#    define STR_LIB_SEG "s11"
```
abhinav92003, line 123: Shouldn't this be "tp" based on the above definition for LIB_SEG_TLS? And STR_SEG should be "invalid" or something?

core/unix/os.c
abhinav92003: Sounds good in general. See comments though.

core/arch/asm_defines.asm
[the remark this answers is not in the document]
abhinav92003: This change looks good, though I'm trying to understand whether asm_defines is exported to users outside DR or not. If it is, then this change is compatibility breaking and we should preserve the existing JUMP_NOT_EQUAL. 

I see drmemory uses asm_defines directly: https://github.com/DynamoRIO/drmemory/blob/a887f796321b1d29b14f617d856a9f157f018c51/CMakeLists.txt#L1022.


Maybe the safest way is to define a new JUMP_NOT_EQUAL macro in asm_defines that takes the extra args for RISC-V and preserve the existing definitions for other archs.

core/unix/os_exports.h
semihalf-kardach-stanislaw: True, it should be gs. Funnily enough previous version of this log used "fs" and I presume it's a mistake.

core/unix/os_exports.h
semihalf-kardach-stanislaw: Yes, it should be. Plus I should fix the `read_thread_register()` and `get_segment_base()` to adapt to this.

core/arch/asm_defines.asm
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Done. RISC-V now has a separate `JUMP_NOT_EQUAL` and the code in `detach_state.c`.

core/unix/os.c
semihalf-kardach-stanislaw: Done.

core/unix/os_public.h
semihalf-kardach-stanislaw: Done.

core/unix/signal.c
semihalf-kardach-stanislaw: Done.

core/arch/arch.h
```diff
@@ -119,6 +119,9 @@ mixed_mode_enabled(void)
 #    define SCRATCH_REG4_OFFS XSI_OFFSET
 #    define SCRATCH_REG5_OFFS XDI_OFFSET
 #    define CALL_SCRATCH_REG DR_REG_R11
+#    define MC_IBL_REG xcx
+#    define MC_RETVAL_REG xax
```
abhinav92003, line 123: Wondering why we have two definitions for the same thing; can we have just one `RETVAL_REG`?

core/unix/tls.h
```diff
@@ -184,6 +184,17 @@ read_thread_register(reg_id_t reg)
         ASSERT_NOT_REACHED();
         return 0;
     }
+#elif defined(RISCV64)
+    ptr_uint_t sel;
+    if (reg == DR_REG_TP) {
+        asm volatile("mv %0, tp" : "=r"(sel));
+    } else if (reg == DR_REG_INVALID) {
```
abhinav92003, line 191: It's a bit confusing to handle DR_REG_INVALID (which is really meant to denote an error) this way. Ideally the caller shouldn't even invoke this routine with this reg.

abhinav92003 (approved):

core/arch/arch.h
semihalf-kardach-stanislaw: I wanted to name it in a similar fashion to `SC_RETURN_REG`, hence the shortened struct name as a prefix to indicate which struct this should be used with.

abhinav92003: Okay, we can keep it as is.

core/unix/tls.h
semihalf-kardach-stanislaw: Actually `DR_REG_INVALID` is a valid choice for `read_thread_register()` on RISC-V. The reason is here: https://github.com/DynamoRIO/dynamorio/pull/5561#discussion_r921392505.

More precisely `read_thread_register()` is/should be called with either `TLS_REG_LIB` or `TLS_REG_ALT`.
I use `DR_REG_INVALID` to express that RISC-V has no alternative TLS register. (also `TLS_REG_ALT` is supposed to be X86-only AFAIK). I chose that (following a suggestion from Derek) as it is a more explicit statement than choosing a dummy register and never using it.

semihalf-kardach-stanislaw: To add some context: AArch64 uses [`DR_REG_TPIDRRO_EL0`](https://developer.arm.com/documentation/ddi0601/2020-12/AArch64-Registers/TPIDRRO-EL0--EL0-Read-Only-Software-Thread-ID-Register) for `TLS_REG_ALT` which is a read-only register for userspace (EL0) that may hold thread information for the kernel software (EL1).

abhinav92003: Okay.

PR #5610, opened 2022-08-13
i#5383 mac a64: Build on M1

derekbruening:
Testing a local repo branch that pull in heyitsanthony:i5383-m1 from PR #5497.


derekbruening:
@AssadHashmi a heads up that I was planning to merge this which is PR #5497 cloned into our repo for editing.  To give @heyitsanthony credit it is simplest to commit the whole PR even though otherwise we'd split it up into smaller pieces.  This includes the codec_v83.txt which adds some pointer auth instructions: but doesn't add tests.  I put in TODO comments about adding tests.

derekbruening:
> @AssadHashmi a heads up that I was planning to merge this which is PR #5497 cloned into our repo for editing. To give @heyitsanthony credit it is simplest to commit the whole PR even though otherwise we'd split it up into smaller pieces. This includes the codec_v83.txt which adds some pointer auth instructions: but doesn't add tests. I put in TODO comments about adding tests.

I guess I'm asking whether these FEAT_PAuth opcodes were already on your roadmap to add and so the full set with tests would be coming along at some future point.

derekbruening:
This PR does not succeed in building core DR for me -- and the tests don't configure.  But let's be incremental and not make this any larger than it already is.

AssadHashmi:
> I guess I'm asking whether these FEAT_PAuth opcodes were already on your roadmap to add and so the full set with tests would be coming along at some future point.

Yes, these type of new FEAT_ures, their opcodes and tests are on the roadmap.

PR #5622, opened 2022-08-19
i#5621: Fix fdtable assert in standalone mode

### LINKED ISSUE #5621: ASSERT dynamo_exited in fd_table_remove
If I pass `-satisfy_w_xor_x` to the tool.drcacheoff.gencode test, I hit an assert `dynamo_exited` in fd_table_remove().  The test does a full DR detach followed by dr_standalone_init() + dr_standalone_exit().  The full detach clears dynamo_exited for re-attach, and the standalone exit closes a file due to the `-satisfy_w_xor_x` mappings and hits the assert.

I think just adding `|| standalone_library` to the assert is the right fix.

derekbruening:
In standalone mode after a detach with -satisfy_w_xor_x, an overactive
assert is hit.  We relax it here.

Augments the tool.drcacheoff.gencode test to cover this case.

Fixes #5621

### DIFF
--- a/clients/drcachesim/tests/burst_gencode.cpp
+++ b/clients/drcachesim/tests/burst_gencode.cpp
@@ -201,7 +201,7 @@ static std::string
 gather_trace()
 {
     if (!my_setenv("DYNAMORIO_OPTIONS",
-#ifdef LINUX
+#if defined(LINUX) && defined(X64)
                    // We pass -satisfy_w_xor_x to further stress that option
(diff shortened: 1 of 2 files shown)


abhinav92003 (approved):

derekbruening:
aarch64 failure is #4167 "Invalid trace entry type 23 before a bundle"

derekbruening:
windows failure is #3320 assert type_is_instr in reuse_distance test

Issue #5538, opened 2022-06-21
Add fast-forward feature to skip N instructions in drmemtrace analysis tools


derekbruening:
This is a feature request to make it easier to simulate a subset of a long
trace. Today, this would be done by running a trace splitter offline. The
proposal is to support skipping forward N instructions by seeking in the
trace file, to make it usable online during simulation.

The records are fixed-size, but the instruction type density is not
uniform, so we'd have to do something like embed markers with instruction
counts every N records so the seeking can find the proper instruction
boundary.  We could limit the fast-forward jumps to every N instructions.

There are several features which may not interact well with this feature,
where we want to only emit non-changing data once early in the trace and
assume a trace reader can cache the data:

- Instruction encodings (#5520) only included for the first dynamic instance of a static
  instruction, to save space.  One solution is to re-emit these at the fast-foward
  point.

- Physical address translation markers (#4014), which are only emitted on the first
  instance of a new page (or when mappings change, if that can be detected).
  Again, a solution would be to re-emit these at the fast-forward "checkpoints".

derekbruening:
A related feature is having an instruction count in the view tool in addition to the total record count used today, and corresponding -skip_instrs and -sim_instrs to go with today's -skip_refs and -sim_refs (for view and simulator tools).

Also, as part of this feature we should probably solve #4915 / #4948: what about non-fetched instrs?

derekbruening:
There is a bug in how reader_t is skipping the duplicate top-of-chunk timestamp,cpu header pair: it is assuming single-thread operation and completely fails for serial mode.  We saw this in a larger traces with serial mode and it can be reproduced in a small trace where the headers are skipped when there is no chunk:
```
        8        0: T3 <marker: timestamp 1001>
        9        0: T3 <marker: tid 3 on core 2>
       10        1: T3 ifetch       4 byte(s) @ 0x000000000000002a non-branch
       11        2: T3 ifetch       4 byte(s) @ 0x000000000000002a non-branch
------------------------------------------------------------
       12        3: T7 ifetch       4 byte(s) @ 0x000000000000002a non-branch
       13        4: T7 ifetch       4 byte(s) @ 0x000000000000002a non-branch
------------------------------------------------------------
       14        5: T3 ifetch       4 byte(s) @ 0x000000000000002a non-branch
------------------------------------------------------------
       15        5: T7 <marker: timestamp 1004>
       16        5: T7 <marker: tid 7 on core 3>
[...]

PR #5624, opened 2022-08-23
i#5505 kernel trace: Add drir2trace to convert DR's IR to trace entries.

dolanzhao:
Adds a new module kernel_image to dump kcore's code segments and a new library drir2trace to convert DR's IR to trace entries.

(1) Adds a new module kernel_image to dump kcore's code segments.
This module can read '/proc/modules' and '/proc/kallsyms' to get all module's information. Then it bases on the module's information to locate all code segments in '/proc/kcore', dump all code segments to 'kimage', and store all code segment's metadata, such as offset, length, and base, to 'kimage.metadata'.

(2) Updates drpt2ir to support getting raw bits of every instruction from 'kimage' and 'kimage.metadata'.
drpt2ir uses two functions to load the kernel code segments. The load_kcore() can get code segments from 'kcore' generated by the Linux perf command. And the load_kernel_image() can get code segments from 'kimage' generated by the kernel_image module.

(3)Updates drpt2trace to support a new working mode: DR Mode.
Adds a new option "-mode" to drpt2trace to let the user choose the working mode of drpt2trace. 
In the current implementation, drpt2trace supports three modes to convert: SIDEBAND Mode(-mode SIDEBAND), ELF Mode(-mode ELF) and DR Mode(-mode DR). And the "DR" mode is used to check whether the syscall PT trace generated by drcachesim is correct.

(4) Adds a new library drir2trace.
This library converts DR's IR format to a list of trace_entry_t instances. 
We only convert DR's IR format to TRACE_TYPE_INSTR type instances in the current implementation.

(5) Updates raw2trace to support PT raw data converting
In the current implementation, raw2trace invokes drpt2ir and drir2trace to convert every syscall's PT raw data to a list of trace_entry_t instances. And it writes all these trace_entry_t instances to the trace file.

(6) Updates the test used to check if dr$sim's kernel PT output is correct.
In the current implementation, the test executes drcachesim to analyze the output userspace and kernel space trace first, and then it uses drpt2trace to check whether the first syscall's PT raw data is correct.

Issue: #5505


derekbruening:
This is on the large side -- seeing all the steps listed it maybe it could have been split up?  Not requesting now as it is already posted, but in the future please consider splitting into smaller pieces.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -213,25 +234,13 @@ static droption_t<unsigned long long> op_sb_kernel_start(
...
-        instr = instr_get_next(instr);
-    }
-
     if (op_print_trace.specified()) {
         /* Print the disassemble code of the trace. */
-        instrlist_disassemble(GLOBAL_DCONTEXT, 0, ilist.data, STDOUT);
+        instrlist_disassemble(GLOBAL_DCONTEXT, 0, drir.data, STDOUT);
     }
-    std::cout << "Number of Instructions: " << count << std::endl;
+    std::cout << "Number of Instructions: " << entries.size() << std::endl;
```
derekbruening, line 243: Once we add encoding entries this will not be the instruction count.  Maybe better to keep the count loop or change this to say "entries" instead of instructions or something.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -791,6 +795,15 @@ template <typename T> class trace_converter_t {
...
+#ifdef BUILD_PT_POST_PROCESSOR
+                    if (!impl()->write(tls, buf_base,
+                                       reinterpret_cast<trace_entry_t *>(buf)))
+                        return "Failed to write to output file";
+                    append_syscall_pt_trace(tls, tid, marker_val);
+                    buf_base = impl()->get_write_buffer(tls);
+                    buf = reinterpret_cast<byte *>(buf_base);
+#endif
                 }
                 impl()->log(3, "Appended marker type %u value " PIFX "\n",
```
derekbruening, line 808: This needs to be *before* the PT data

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1442,6 +1455,75 @@ template <typename T> class trace_converter_t {
...
+        config.raw_file_path = impl()->get_syscall_pt_trace_dir(tls) + "/" +
+            std::to_string(thread_id) + "." + std::to_string(syscall_id) +
+            PT_DATA_FILE_NAME_SUFFIX;
+        config.kernel_image_path =
+            impl()->get_syscall_pt_trace_dir(tls) + "/" + KERNEL_IMAGE_FILE_NAME;
+        config.kernel_image_metadata_path =
+            impl()->get_syscall_pt_trace_dir(tls) + "/" + KERNEL_IMAGE_METADATA_FILE_NAME;
+        config.init_with_metadata(config.raw_file_path + PT_METADAT_FILE_NAME_SUFFIX);
+
+        std::unique_ptr<pt2ir_t> ptconverter(new pt2ir_t());
```
derekbruening, line 1477: This can just be a local var on the stack -- I don't see why this needs to be on the heap?

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1442,6 +1455,75 @@ template <typename T> class trace_converter_t {
...
+        if (!ptconverter->init(config)) {
+            error = "Failed to initialize pt2ir";
```
derekbruening, line 1479: Just return the string: no need to set a named variable.  Ditto below.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1442,6 +1455,75 @@ template <typename T> class trace_converter_t {
...
+            error = "Failed to convert DR IR to trace entries";
+            return error;
+        }
+
+        trace_entry_t *buf_base = impl()->get_write_buffer(tls);
+        byte *buf = reinterpret_cast<byte *>(buf_base);
+        for (auto &entry : entries) {
+            trace_entry_t *buf_entry = reinterpret_cast<trace_entry_t *>(buf);
+            *buf_entry = entry;
+            buf += sizeof(trace_entry_t);
```
derekbruening, line 1502: This will also need to count instructions for #5538 which I am working on now, so this will add a complication there: please add a TODO comment.

derekbruening, line 1502: A trace analyzer may assert that load/store instructions have data address entries immediately afterward.  We need a solution to indicate what to expect.  The offline file type mentioned above helps: but it would be better to delimit the instr-only regions.  A new marker type at the start and end?  A general type that doesn't care about kernel but is about instr-only, to support using PT for windows of user code too.  Or it shouldn't really be instr-only as we plan to synthesize data entries: but on that note, do we consider this instr-only state a temporary non-published state (and so should have comments here making that clear) and we only want to export PT sequences that have synthesized data entries?

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1791,6 +1875,7 @@ class raw2trace_t : public trace_converter_t<raw2trace_t> {
         uint64 last_decode_modidx;
         uint64 last_decode_modoffs;
         block_summary_t *last_block_summary;
+        std::string syscall_pt_trace_dir;
```
derekbruening, line 1878: We don't need to duplicate this in every thread

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1442,6 +1455,75 @@ template <typename T> class trace_converter_t {
...
+#    define PT_DATA_FILE_NAME_SUFFIX ".pt"
+#    define PT_METADAT_FILE_NAME_SUFFIX ".metadata"
+#    define KERNEL_IMAGE_FILE_NAME "kimage"
+#    define KERNEL_IMAGE_METADATA_FILE_NAME "kimage.metadata"
+    std::string
+    append_syscall_pt_trace(void *tls, thread_id_t thread_id, uint64_t syscall_id)
+    {
+        std::string error = "";
+        pt2ir_config_t config = {};
+        config.raw_file_path = impl()->get_syscall_pt_trace_dir(tls) + "/" +
```
derekbruening, line 1468: raw2trace cannot open files, to support use in environments where library-based filesystems are used.  Any files used need to be passed in as istreams or some similar mechanism, or if all the files involved are relatively small we could require that they be copied to local paths where a unix file open will work (we do that today for module_mapper_t binaries, and for the encoding file we pass in a file_t).  But the actual PT trace data is likely going to get large, so I do not think requiring local files is a viable solution.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1633,6 +1715,7 @@ class raw2trace_t : public trace_converter_t<raw2trace_t> {
                 const std::vector<std::ostream *> &out_files,
                 file_t encoding_file = INVALID_FILE, void *dcontext = nullptr,
                 unsigned int verbosity = 0, int worker_count = -1,
+                const std::string &syscall_pt_trace_dir = "",
```
derekbruening, line 1718: Seems like this should either be appended to avoid breaking existing uses or inserted at a meaningful place: it seems unusual to insert it here.  I would append it.  But see below on this model of raw2trace opening files and assuming they are local probably not being workable: we probably need a different interface here, passing in istreams.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1521,9 +1522,13 @@ event_exit(void)
...
         drpttracer_exit();
-        /* Dump kcore and kallsyms to {kernel_pt_logsubdir}. */
+        /* Dump kcore to kimage and kimage.metadata, and store the two file to
+         * {kernel_pt_logsubdir}.
+         * The Kernel dump function is only invoked at processes exit on the Linux X86_64
+         * platform.
+         */
         if (!syscall_pt_trace_t::kernel_image_dump(kernel_pt_logsubdir)) {
-            NOTIFY(0, "WARNING: failed to run shellscript to dump kernel image\n");
+            NOTIFY(0, "WARNING: failed to dump kernel image\n");
```
derekbruening, line 1531: Just like for raw2trace, tracer can't directly write to files: we need indirection to support library-based filesystems.  This file dumping needs to go through the write_file indirection.

derekbruening (commented):
Did not look at the kernel_image files.

Issue #5629, opened 2022-08-24
linux.clone failing on some environments


abhinav92003:
The linux.clone test today verifies DR's handling of SYS_clone3. The checks the test performs are based on whether or not SYS_clone3 is defined: if it is not defined, we check for ENOSYS, but if it is, we check for the actual result. It isn't reliable to assume presence or absence of kernel syscall support based on whether the glibc version defines the corresponding SYS_ constant.

PR #5630, opened 2022-08-24
i#5629: Fix expected output in linux.clone test.

abhinav92003:
Accepts EINVAL as a valid response from clone3 even if SYS_clone3
is not defined by glibc. This is to avoid incorrectly failing the
test on systems where the kernel supports clone3 but glibc version
doesn't define the SYS_clone3 constant.

Fixes: #5629


derekbruening (approved):

abhinav92003:
run arm tests

Issue #5383, opened 2022-02-23
Add MacOS on AArch64 support


derekbruening:
Xref https://groups.google.com/d/msgid/dynamorio-users/01e7e991-675c-4ff4-85f3-e4745e8fecfbn%40googlegroups.com?utm_medium=email&utm_source=footer

This issue covers adding support for DR on MacOS on AArch64 such as M1.

derekbruening:
The Github MacOS runners are shifting to AArch64 so the priority here is a little higher.
They still have x86 in macos-14-large: https://github.com/actions/runner-images?tab=readme-ov-file#available-images, https://github.com/actions/runner-images/issues/9741.
Xref #7720.

PR #5632, opened 2022-08-29
i#5383 mac a64, part 2: Fix compiler warnings

derekbruening:
Fixes many unused-variable warnings when building the core and
extension libraries.  This enables a config without tests or docs to
build.  The drsyms libraries are missing so clients/ is not yet built.

Issue: #5383

### DIFF
--- a/core/arch/interp.c
+++ b/core/arch/interp.c
@@ -5504,7 +5504,7 @@ recreate_fragment_done:
         *alloc_res = alloc;
     if (f_res == NULL && alloc)
         fragment_free(dcontext, f);
-    DODEBUG(ok =) dr_set_isa_mode(dcontext, old_mode, NULL);
+    DEBUG_DECLARE(ok =) dr_set_isa_mode(dcontext, old_mode, NULL);
     ASSERT(ok);
     return ilist;
 }
--- a/core/arch/interp.c
+++ b/core/arch/interp.c
@@ -4656,7 +4656,7 @@ static void
 build_native_exec_bb(dcontext_t *dcontext, build_bb_t *bb)
 {
     instr_t *in;
-    opnd_t jmp_tgt;
+    opnd_t jmp_tgt IF_AARCH64(UNUSED);
 #if defined(X86) && defined(X64)
     bool reachable = rel32_reachable_from_vmcode(bb->start_pc);
 #endif
@@ -5294,7 +5294,7 @@ recreate_fragment_ilist(dcontext_t *dcontext, byte *pc,
     fragment_t *f;
     uint flags = 0;
     instrlist_t *ilist;
-    bool alloc = false, ok;
+    bool alloc = false;
     monitor_data_t md = {
         0,
     };
@@ -5335,7 +5335,8 @@ recreate_fragment_ilist(dcontext_t *dcontext, byte *pc,
     }
 
     /* Recreate in same mode as original fragment */
-    ok = dr_set_isa_mode(dcontext, FRAG_ISA_MODE(f->flags), &old_mode);
+    DEBUG_DECLARE(bool ok =)
+    dr_set_isa_mode(dcontext, FRAG_ISA_MODE(f->flags), &old_mode);
     ASSERT(ok);
 
     if ((f->flags & FRAG_IS_TRACE) == 0) {
@@ -5503,7 +5504,7 @@ recreate_fragment_done:
         *alloc_res = alloc;
     if (f_res == NULL && alloc)
         fragment_free(dcontext, f);
-    ok = dr_set_isa_mode(dcontext, old_mode, NULL);
+    DODEBUG(ok =) dr_set_isa_mode(dcontext, old_mode, NULL);
     ASSERT(ok);
     return ilist;
 }
@@ -7247,7 +7248,9 @@ decode_fragment(dcontext_t *dcontext, fragment_t *f, byte *buf, /*IN/OUT*/ uint 
     bool coarse_elided_ubrs = false;
     dr_isa_mode_t old_mode;
     /* for decoding and get_ibl routines we need the dcontext mode set */
-    bool ok = dr_set_isa_mode(dcontext, FRAG_ISA_MODE(f->flags), &old_mode);
+    DEBUG_DECLARE(bool ok =)
+    dr_set_isa_mode(dcontext, FRAG_ISA_MODE(f->flags), &old_mode);
+    ASSERT(ok);
     /* i#1494: Decoding a code fragment from code cache, decode_fragment
      * may mess up the 32-bit/64-bit mode in -x86_to_x64 because 32-bit
      * application code is encoded as 64-bit code fragments into the code cache.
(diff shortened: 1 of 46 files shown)


core/arch/interp.c
```diff
@@ -4656,7 +4656,7 @@ static void
 build_native_exec_bb(dcontext_t *dcontext, build_bb_t *bb)
 {
     instr_t *in;
-    opnd_t jmp_tgt;
+    opnd_t jmp_tgt IF_AARCH64(UNUSED);
```
abhinav92003, line 4659: Isn't this used on L4748 below for AArch64?

```
jmp_tgt = opnd_create_pc(bb->start_pc);
```

core/arch/interp.c
```diff
@@ -7247,7 +7248,9 @@ decode_fragment(dcontext_t *dcontext, fragment_t *f, byte *buf, /*IN/OUT*/ uint
     bool coarse_elided_ubrs = false;
     dr_isa_mode_t old_mode;
     /* for decoding and get_ibl routines we need the dcontext mode set */
-    bool ok = dr_set_isa_mode(dcontext, FRAG_ISA_MODE(f->flags), &old_mode);
+    DEBUG_DECLARE(bool ok =)
```
abhinav92003, line 7251: Maybe we should have a macro for such patterns where the returned boolean is supposed to be checked only in debug mode.

core/ir/aarch64/codec.c
```diff
@@ -1323,6 +1323,7 @@ encode_opnd_b_const_sz(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *en
     return false;
 }
 
+#if 0  /* Currently unused. */
 /* h_const_sz: Operand size for half (16-bit) vector elements
```
abhinav92003, line 1327: If these routines are not exported, isn't it fine to remove them? Or do we think they might be used in the near future?

core/ir/instr_shared.c
```diff
@@ -1490,15 +1490,15 @@ instr_decode_cti(dcontext_t *dcontext, instr_t *instr)
      */
     if (!instr_opcode_valid(instr) ||
         (instr_is_cti(instr) && !instr_operands_valid(instr))) {
-        byte *next_pc;
+        DEBUG_EXT_DECLARE(byte * next_pc;)
```
abhinav92003, line 1493: DEBUG_EXT_DECLARE doesn't seem to do anything that's "declaration" related. Why not just IF_DEBUG/DODEBUG?

core/utils.c
```diff
@@ -1864,12 +1864,9 @@ d_r_notify(syslog_event_type_t priority, bool internal, bool synch,
            const char *fmt, ...)
 {
     char msgbuf[MAX_LOG_LENGTH];
-    int size;
     va_list ap;
     va_start(ap, fmt);
-    /* FIXME : the vsnprintf call is not needed in the most common case where
```
abhinav92003, line 1870: This fixme is fixed now?

core/utils.h
```diff
@@ -1567,15 +1567,21 @@ enum { LONGJMP_EXCEPTION = 1 };
...
 /* Use to shut up the compiler about an unused variable when the
  * alternative is a painful modification of more src code. Our
  * standard is to use this macro just after the variable is declared
  * and to use it judiciously.
  */
-#define UNUSED_VARIABLE(pv)           \
-    {                                 \
-        void *unused_pv = (void *)pv; \
-        unused_pv = NULL;             \
+#define UNUSED_VARIABLE(pv)                  \
```
abhinav92003, line 1581: Why do we need UNUSED_VARIABLE and UNUSED both?

Edit: Oh okay. We use the former to not have to put too many pre-processor defines on the declaration. Whereas the latter is used when the variable is unused in all or only a small number of environments.

core/unix/os.c
```diff
@@ -4702,7 +4699,7 @@ void
 os_syslog(syslog_event_type_t priority, uint message_id, uint substitutions_num,
           va_list args)
 {
-    int native_priority;
+    int native_priority UNUSED;
```
abhinav92003, line 4702: What does the native_priority var here achieve?

core/unix/os.c
```diff
@@ -6039,7 +6035,7 @@ handle_execve(dcontext_t *dcontext)
      */
     char *fname;
     bool x64 = IF_X64_ELSE(true, false);
-    bool expect_to_fail = false;
+    bool expect_to_fail UNUSED = false;
```
abhinav92003, line 6038: But it's used on L6151:
```
if (should_inject && DYNAMO_OPTION(early_inject) && !expect_to_fail) {
```


Is unused a no-op if it is actually used?

abhinav92003 (approved):

core/utils.h
abhinav92003: s/only a small number of environments/simple set of environments/

core/arch/interp.c
derekbruening: It is only set there: never read after that for a64.  This clang 14 is very aggressive at these warnings: as you see it found tons of such vars.

core/ir/aarch64/codec.c
derekbruening: Let's ask the owners: @AssadHashmi the latest clang (14) flags a bunch of codec.c functions as unused -- were these used in the past and are now truly dead, or  are these planned to be used in the future?

core/ir/instr_shared.c
derekbruening: I thought DEBUG_EXT_DECLARE was for CLIENT_ASSERT which is subtly different in which places it's enabled vs DEBUG -- at least that's what the comment says:
```C
/* For use in CLIENT_ASSERT or elsewhere that exists even if
 * STANDALONE_DECODER is defined.
 */
```

core/utils.h
derekbruening: Also I think we didn't find a cross-platform compiler marking that worked well in the past: we were avoiding pragmas in VS

core/unix/os.c
derekbruening: Didn't want to deal with the unfinished code: you're asking to delete the function body code that just sets native_priority and never does anything with it?

core/unix/os.c
derekbruening: Used only for LINUX, not OSX.  Yes, it's a nop if it's actually used, so doesn't hurt LINUX build.

core/arch/interp.c
derekbruening: Could be confusing though if you had:
```C
DEBUG_CHECK_RET_TRUE(dr_set_isa_mode(dcontext, FRAG_ISA_MODE(f->flags), &old_mode));
```
Which the reader will probably think is only executed at all in debug build.
Would have to name it ALWAYS_RUN_BUT_CHECK_RETVAL_IN_DEBUG() or sthg?

core/ir/aarch64/codec.c
derekbruening: No reply yet: I guess they can always be restored from version control so I will remove them.

core/arch/interp.c
derekbruening: Just going to leave the code as-is.

core/ir/instr_shared.c
derekbruening: Sticking with what is intended for use in this situation.

core/unix/os.c
derekbruening: Going to leave it.

core/ir/aarch64/codec.c
derekbruening: Removing them breaks the aarch64_check_codec_order.py script which statically examines this code and compares to opnd_defs.txt.  So we'd have to remove lines from opnd_defs.txt too.  I'm going to put them back and defer to @AssadHashmi on a longer-term fix.

Issue #2112, opened 2016-12-14
static_signal failing with strange output (non-det)


derekbruening:
Failed once on CDash:

http://dynamorio.org/CDash/testDetails.php?test=168921&build=15987

```
Required regular expression not found.Regex=[^Sending SIGUSR1 pre-DR-init
Got SIGUSR1
pre-DR init
in dr_client_main
Sending SIGUSR1 pre-DR-start
Got SIGUSR1
pre-DR start
Sending SIGUSR1 under DR
Got SIGUSR1
pre-DR stop
Saw some bb events
Saw 1 signal\(s\)
Sending SIGUSR1 post-DR-stop
Got SIGUSR1
all done
$
]

Test output
Sending SIGUSR1 pre-DR-init
Got SIGUSR1
pre-DR init
in dr_client_main
Sending SIGUSR1 pre-DR-start
Got SIGUSR1
pre-DR start
Sending SIGUSR1 under DR
Got SIGUSR1
Saw some bb events
Saw 2 signal(s)
```

derekbruening:
Failed on Travis: https://travis-ci.org/DynamoRIO/dynamorio/jobs/216403717

derekbruening:
I believe the recent Travis failures are not from the original code and thus the original CDash issue filed here but from the failure to see any alarm signals from d6d4faa417f5dc7b78c574bc5517d95ab33bfa3c which did not provide any guarantee that they'd be hit.

PR #5643, opened 2022-09-08
i#5636: Fix elided duplicate timestamps

### LINKED ISSUE #5636: drcachesim.invariants test failing sometimes on win32

derekbruening:
PR #5633 added code to skip the duplicate timestamps at the top of each chunk: but its logic assumed there would never be two legitimate timestamps with identical values.  That does happen, in particular in our online-drcachesim tests on Windows.  This resulted in the invariant_checker not seeing some timestamp entries, causing its exception for non-fetched instrs across thread switches to not apply and resulting in invariant error reports.

We fix this by skipping the first timestamp in each chunk by instruction count instead.  We'll want the insruction and chunk counts for #5538 and I was about to add those fields in any case.

Fixes #5636

### DIFF
--- a/clients/drcachesim/reader/reader.h
+++ b/clients/drcachesim/reader/reader.h
@@ -143,7 +143,9 @@ private:
     addr_t prev_instr_addr_ = 0;
     int bundle_idx_ = 0;
     std::unordered_map<memref_tid_t, memref_pid_t> tid2pid_;
-    uint64_t last_timestamp_ = 0;
+    uint64_t cur_instr_count_ = 0;
+    uint64_t chunk_instr_count_ = 0;
+    uint64_t last_timestamp_instr_count_ = 0;
     bool skip_next_cpu_ = false;
 };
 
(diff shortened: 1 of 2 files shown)


clients/drcachesim/reader/reader.h
```diff
@@ -143,7 +143,9 @@ class reader_t : public std::iterator<std::input_iterator_tag, memref_t> {
     addr_t prev_instr_addr_ = 0;
     int bundle_idx_ = 0;
     std::unordered_map<memref_tid_t, memref_pid_t> tid2pid_;
-    uint64_t last_timestamp_ = 0;
+    uint64_t cur_instr_count_ = 0;
+    uint64_t chunk_instr_count_ = 0;
```
abhinav92003, line 147: As this is expected to be a constant once assigned, I wonder if we should name it as kChunkInstrCount or something.

abhinav92003 (approved):

clients/drcachesim/reader/reader.h
derekbruening: That naming convention is not in place anywhere in dr$sim or DR in general: all-caps normally indicates a constant, but that doesn't feel right here.  I'll just add a comment.

derekbruening:
Failure is 32-bit x86 static_signal timeout: #2112

PR #5647, opened 2022-09-12
i#5644 Apply EVEX compressed disp when segment prefixes are present

### LINKED ISSUE #5644: Incorrect address offset decoding
**Describe the bug**

Decoding `2e 2e 67 62 01 c5 00 c4 62 21 00` with drdecode gives:
```
addr32 vpinsrw %xmm23[14byte] %cs:0x21(%r10d)[2byte] $0x00 -> %xmm28
```

But actual execution on CPU and other decoders suggest that the address offset is 0x42 rather than 0x21:
```
echo -en "\x2e\x2e\x67\x62\x01\xc5\x00\xc4\x62\x21\x00" > /tmp/bin | objdump -mi386 -Mx86-64 -b binary -D /tmp/bin
cs cs vpinsrw $0x0,0x42(%r10d),%xmm23,%xmm28
```

```
echo -en "0x2e 0x2e 0x67 0x62 0x01 0xc5 0x00 0xc4 0x62 0x21 0x00" | llvm-mc --disassemble -
	vpinsrw	$0, %cs:66(%r10d), %xmm23, %xmm28
```

**Expected behavior**

Decode address with 0x42 offset.

**Versions**
 6e69f2a90b55870e849718223815c2d23fba0871

Did not try, but there are only 3 commits that don't look relevant.

khuey:
The current behavior is simply a bug, the compressed displacement adjustment needs to be performed whether segment prefixes are present or not.

Fixes #5644


core/ir/x86/decode.c
```diff
@@ -1761,6 +1761,10 @@ decode_modrm(decode_info_t *di, byte optype, opnd_size_t opsize, opnd_t *reg_opn
         }
         force_full_disp = !needs_full_disp && di->has_disp && disp >= INT8_MIN &&
             disp <= INT8_MAX && di->mod == 2;
+        if (di->evex_encoded) {
```
derekbruening, line 1764: May as well merge this into the end of the same if condition a few lines up at 1756-1761

derekbruening (approved):

core/ir/x86/decode.c
derekbruening: Or does the force_full_disp computation need to use the un-compressed-scaled disp?

khuey: We can merge them. force_full_disp is actually mutually exclusive (because it requires di->mod == 2 and this EVEX compressed disp requires di->mod == 1).

PR #5648, opened 2022-09-12
i5645: Make the last rep/repne prefix "win"

### LINKED ISSUE #5447: TZCNT is incorrectly decoded
Decoding this byte sequence: `f3 f2 0f bc f2`
DynamoRIO: `tzcnt %edx -> %esi`
Capstone:     `bsf	esi, edx`
llvm-mc: `invalid instruction encoding`
objdump:
```
   0:	f3 f2 0f bc          	repz (bad) 
   4:	f2                   	repnz
```

Intel manual says TZCNT encodings are:
```
  F3 0F BC /r
  F3 REX.W 0F BC /r
```

Executing this instruction with different values of RDX always yields 0 in RSI, so effectively it seems to work as some kind of non-trapping nop (?).

If it's an invalid instruction for my use case it would be more useful if DynamoRIO would reject to decode it (or decode to something that would agree with CPU behavior, e.g. OP_nop).

On commit 6e88738383e78c0be9c3dda373e43f311425b862.

### LINKED ISSUE #5645: Incorrect decoding of PSHUFLW/PSHUFHW
**Describe the bug**

Decoding `4d 49 41 f3 f3 f2 4b 0f 70 76 00 ff` with drdecode gives:
```
pshufhw 0x00(%r14)[16byte] $0xff -> %xmm6
```
executing this instruction with %r14 pointing to memory region with 0x00, 0x01, 0x02, ... bytes should result in:
`XMM6=0x0f0e0f0e0f0e0f0e706050403020100`
But executing it on a CPU gives:
`XMM6=0x0f0e0d0c0b0a09080706070607060706`

Decoding the instruction with other decoders suggests it's a different instruction:
```
echo -en " 0x4d 0x49 0x41 0xf3 0xf3 0xf2 0x4b 0x0f 0x70 0x76 0x00 0xff" | llvm-mc --disassemble -
	rep
	pshuflw	$255, (%r14), %xmm6             # xmm6 = mem[3,3,3,3,4,5,6,7]
```

```
echo -en "\x4d\x49\x41\xf3\xf3\xf2\x4b\x0f\x70\x76\x00\xff" > /tmp/bin | objdump -mi386 -Mx86-64 -b binary -D /tmp/bin
00000000 <.data>:
   0:	4d                   	rex.WRB
   1:	49                   	rex.WB
   2:	41                   	rex.B
   3:	f3 f3 f2 4b 0f 70 76 	repz repz rex.WXB pshuflw $0xff,0x0(%r14),%xmm6
   a:	00 ff 
```

PSHUFLW would indeed produce the result that agrees with actual CPU execution. 

**Expected behavior**

Decode the instruction as PSHUFLW.

**Versions**
6e69f2a90b55870e849718223815c2d23fba0871

Did not try, but there are only 3 commits that don't look relevant.

khuey:
x86 allows multiple prefixes to be present on an instruction. Although not documented in the Intel SDM, empirical testing shows that the last rep/repne prefix "wins", since it makes no sense for both prefixes to be present. Since these prefixes are used at times to select between instructions (e.g. anything with PREFIX_EXT in the decoder table) faithfully replicating this behavior is important to ensure that byte sequences with unnecessary leading prefixes are correctly decoded.

Fixes #5645
Fixes #5447

### DIFF
--- a/core/ir/x86/decode.c
+++ b/core/ir/x86/decode.c
@@ -880,6 +880,7 @@ read_prefix_ext(const instr_info_t *info, decode_info_t *di)
     int code = (int)info->code;
     /* The order here matters: rep, then repne, then data (i#2431). */
     int idx = (di->rep_prefix ? 1 : (di->repne_prefix ? 3 : (di->data_prefix ? 2 : 0)));
+    ASSERT(!(di->rep_prefix && di->repne_prefix));
     if (di->vex_encoded)
         idx += 4;
     else if (di->evex_encoded)
@@ -1014,9 +1015,11 @@ read_instruction(byte *pc, byte *orig_pc, const instr_info_t **ret_info,
             if (info->code == PREFIX_REP) {
                 /* see if used as part of opcode before considering prefix */
                 di->rep_prefix = true;
+                di->repne_prefix = false;
             } else if (info->code == PREFIX_REPNE) {
                 /* see if used as part of opcode before considering prefix */
                 di->repne_prefix = true;
+                di->rep_prefix = false;
             } else if (REG_START_SEGMENT <= info->code &&
                        info->code <= REG_STOP_SEGMENT) {
                 CLIENT_ASSERT_TRUNCATE(di->seg_override, ushort, info->code,
--- a/suite/tests/api/ir_x86.c
+++ b/suite/tests/api/ir_x86.c
@@ -2728,6 +2728,30 @@ test_evex_compressed_disp_with_segment_prefix(void *dc)
 #endif
 }
 
+static void
+test_extra_leading_prefixes(void *dc)
+{
+#ifdef X64
+    byte *pc;
+    const byte b1[] = { 0xf3, 0xf2, 0x4b, 0x0f, 0x70, 0x76, 0x00, 0xff };
+    const byte b2[] = { 0xf3, 0xf2, 0x0f, 0xbc, 0xf2 };
+    char dbuf[512];
+    int len;
+
+    pc =
+        disassemble_to_buffer(dc, (byte *)b1, (byte *)b1, false /*no pc*/,
+                              false /*no bytes*/, dbuf, BUFFER_SIZE_ELEMENTS(dbuf), &len);
+    ASSERT(pc == &b1[0] + sizeof(b1));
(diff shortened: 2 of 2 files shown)


khuey:
Hmm, this fails on 32 bit because of

```diff
--- /home/khuey/dev/dynamorio/obj-32/suite/tests/drdecode_decenc_x86.expect-out 2022-09-12 16:54:00.936150403 -0700
+++ /home/khuey/dev/dynamorio/obj-32/suite/tests/drdecode_decenc_x86.expect-expect      2022-09-12 16:54:00.948150414 -0700
@@ -76144,9 +76144,9 @@
  0f 54 f2             andps  %xmm2, %xmm6
  66 0f 11 22          movupd %xmm4, (%edx)
  f2 67 66 0f 11 22    data16 addr16 movsd  %xmm4, (%bp,%si)
  f3 66 0f 11 22       data16 movss  %xmm4, (%edx)
- f3 67 f2 66 0f 11 22 data16 addr16 movsd  %xmm4, (%bp,%si)
+ f3 67 f2 66 0f 11 22 data16 addr16 movss  %xmm4, (%bp,%si)
  f3 66 3e 0f 11 22    data16 movss  %xmm4, %ds:(%edx)
  f2 66 36 0f 11 22    data16 movsd  %xmm4, %ss:(%edx)
  22 f2                and    %dl, %dh
  66 3e 36 0f 11 22    movupd %xmm4, %ss:(%edx)
```

So perhaps this approach isn't correct after all.

khuey:
Ah, no, the *expectations* are wrong here.

derekbruening:
> Ah, no, the _expectations_ are wrong here.

Unfortunately I think at some point some of these test expected outputs were taken from the DR output and human-verified?

derekbruening (approved):
Thank you for the quick fix!

nit: s/i5645/i#5645/ in the title.

khuey:
> > Ah, no, the _expectations_ are wrong here.
> 
> Unfortunately I think at some point some of these test expected outputs were taken from the DR output and human-verified?

Yeah that doesn't surprise me, going through them all by hand would be excruciating.

dvyukov:
Thanks!

Issue #5493, opened 2022-05-12
CRASH tool.drcacheoff.windows-zlib test on AArch64


derekbruening:
This happened once:

http://139.178.82.61:8080/job/DynamoRIO-AArch64-Precommit/1574/consoleFull

```
71/79 Test #227: code_api|tool.drcacheoff.windows-zlib ...........................***Failed    0.63 sec
Running cmd |/var/lib/jenkins/workspace/DynamoRIO-AArch64-Precommit/build/build_release-external-64/bin64/drrun;-s;90;-quiet;-killpg;-stderr_mask;0xC;-dumpcore_mask;0;-code_api;-t;drcachesim;-offline;-subdir_prefix;tool.drcacheoff.windows-zlib;-raw_compress;zlib;-no_split_windows;-trace_after_instrs;20K;-trace_for_instrs;5K;-retrace_every_instrs;35K;--;/var/lib/jenkins/workspace/DynamoRIO-AArch64-Precommit/build/build_release-external-64/suite/tests/bin/simple_app|
Running postcmd |/var/lib/jenkins/workspace/DynamoRIO-AArch64-Precommit/build/build_release-external-64/clients/bin64/drcachesim;-indir;/var/lib/jenkins/workspace/DynamoRIO-AArch64-Precommit/build/build_release-external-64/suite/tests/tool.drcacheoff.windows-zlib.simple_app.2994852.1587.dir;-simulator_type;basic_counts|
CMake Error at /var/lib/jenkins/workspace/DynamoRIO-AArch64-Precommit/suite/tests/runmulti.cmake:106 (message):
  *** postcmd failed (Segmentation fault): ***

Call Stack (most recent call first):
  /var/lib/jenkins/workspace/DynamoRIO-AArch64-Precommit/suite/tests/runmulti.cmake:118 (process_cmdline)
```

dolanzhao:
The PR #5555 failed to pass the tool.drcacheoff.windows-zlib test on CI AArch64 too.
http://139.178.82.61:8080/job/DynamoRIO-AArch64-Precommit/1773/console

abhinav92003:
Saw a segfault on #5581 for this test: http://139.178.82.61:8080/job/DynamoRIO-AArch64-Precommit/1867/console

PR #5652, opened 2022-09-15
i#5651: Switch DR's suspend signal to SIGILL with sigqueueinfo

### LINKED ISSUE #5651: gdb has poor support for DR's new takeover signal SIGSTKFLT
For #5458 we switched from SIGUSR2 to SIGSTKFLT.  However, gdb does not handle this signal well at all:

```
Thread 2 "tool.drcacheoff" received signal ?, Unknown signal.
(gdb) p $_siginfo
$1 = {
  si_signo = 16,
(gdb) c
Continuing.
warning: Signal ? does not exist on this system.

(gdb) handle 16 nostop noprint pass
Only signals 1-15 are valid as numeric signals.
Use "info signals" for a list of symbolic signals.
(gdb) handle SIGSTKFLT nostop noprint pass
Unrecognized or ambiguous flag word: "SIGSTKFLT".
```

I tried `handle all nostop noprint pass` and it didn't apply to signal 16!
So we have to hit `c` a hundred times to make progress which is not good.

I saw this in gdb version 9 and 10.

I propose we switch to SIGPWR which gdb does support.  I tested it with gdb and with QEMU and QEMU does not freak out and abort like it does with SIGFPE.

Other candidates are SIGXCPU and SIGXFSZ but the kernel will produce those: SIGPWR seems more obscure.

derekbruening:
Switches the signal DR uses to suspend threads from SIGSTKFLT, which is not supported in gdb, to SIGPWR.  SIGPWR is similarly obscure and never used on Linux, but is supported by gdb, and does not cause QEMU to fail.

Fixes #5651

### DIFF
--- a/api/docs/debugging.dox
+++ b/api/docs/debugging.dox
@@ -82,7 +82,7 @@ On Linux, DR sends itself a SIGILL signal during initialization, in order to mea
 
 Additionally, DR uses various "safe read" strategies where it may raise a SIGSEGV or SIGBUS while examining application memory.  Load symbols and look for `safe_read` variants on the call stack (such as `safe_read_tls_magic`, `safe_read_asm_pre`, `safe_read_fast`) to identify these.  Just like with the init-time SIGILL, simply continue past these.
 
-One final signal used by DR is SIGSTKFLT (SIGFPE on MacOS), which is sent to other threads to suspend or terminate them.
+One final signal used by DR is SIGPWR (SIGFPE on MacOS), which is sent to other threads to suspend or terminate them.
 
 ## Attaching
 
--- a/api/docs/debugging.dox
+++ b/api/docs/debugging.dox
@@ -78,11 +78,9 @@ the shared library that identifies the debug file.
 
 ## Expected Signals
 
-On Linux, DR sends itself a SIGILL signal during initialization, in order to measure the size of the signal frame used by the kernel.  If you are under gdb, simply continue past this signal.
+On Linux, DR sends itself a SIGILL signal during initialization, in order to measure the size of the signal frame used by the kernel.  If you are under gdb, simply continue past this signal.  SIGILL (SIGFPE on MacOS; SIGSTKFLT inside QEMU) is also sent to other threads to suspend or terminate them later on during execution.
 
 Additionally, DR uses various "safe read" strategies where it may raise a SIGSEGV or SIGBUS while examining application memory.  Load symbols and look for `safe_read` variants on the call stack (such as `safe_read_tls_magic`, `safe_read_asm_pre`, `safe_read_fast`) to identify these.  Just like with the init-time SIGILL, simply continue past these.
-
-One final signal used by DR is SIGPWR (SIGFPE on MacOS), which is sent to other threads to suspend or terminate them.
 
 ## Attaching
 
--- a/api/docs/debugging.dox
+++ b/api/docs/debugging.dox
@@ -78,7 +78,7 @@ the shared library that identifies the debug file.
 
 ## Expected Signals
 
-On Linux, DR sends itself a SIGILL signal during initialization, in order to measure the size of the signal frame used by the kernel.  If you are under gdb, simply continue past this signal.  SIGILL (SIGFPE on MacOS; SIGSTKFLT inside QEMU) is also sent to other threads to suspend or terminate them later on during execution.
+On Linux, DR sends itself a SIGILL signal during initialization, in order to measure the size of the signal frame used by the kernel.  If you are under gdb, simply continue past this signal.  SIGILL (SIGFPE on MacOS; SIGSTKFLT inside QEMU where the others crash QEMU (but note the poor support for SIGSTKFLT in gdb)) is also sent to other threads when attaching to a multi-threaded application and to suspend or terminate them later on during execution.
 
 Additionally, DR uses various "safe read" strategies where it may raise a SIGSEGV or SIGBUS while examining application memory.  Load symbols and look for `safe_read` variants on the call stack (such as `safe_read_tls_magic`, `safe_read_asm_pre`, `safe_read_fast`) to identify these.  Just like with the init-time SIGILL, simply continue past these.
 
--- a/core/lib/globals_shared.h
+++ b/core/lib/globals_shared.h
@@ -928,6 +928,7 @@ typedef enum {
  */
 enum {
     NUDGE_IS_INTERNAL = 0x01, /* nudge is internally generated */
+    NUDGE_IS_SUSPEND = 0x02,  /* This is an internal SUSPEND_SIGNAL. */
 #ifdef WINDOWS
     NUDGE_NUDGER_FREE_STACK = 0x02, /* nudger will free the nudge thread's stack so the
                                      * nudge thread itself shouldn't */
--- a/core/lib/globals_shared.h
+++ b/core/lib/globals_shared.h
@@ -923,13 +923,14 @@ typedef enum {
 #define NUDGE_ARG_VERSION_1 1
 #define NUDGE_ARG_CURRENT_VERSION NUDGE_ARG_VERSION_1
 
-/* nudge_arg_t flags
- * On Linux only 2 bits for these
+/* nudge_arg_t bitfield flags.
+ * On UNIX we only have space for 2 bits for these.
  */
 enum {
-    NUDGE_IS_INTERNAL = 0x01, /* nudge is internally generated */
-    NUDGE_IS_SUSPEND = 0x02,  /* This is an internal SUSPEND_SIGNAL. */
-#ifdef WINDOWS
+    NUDGE_IS_INTERNAL = 0x01, /* The nudge is internally generated. */
+#ifdef UNIX
+    NUDGE_IS_SUSPEND = 0x02, /* This is an internal SUSPEND_SIGNAL. */
+#else
     NUDGE_NUDGER_FREE_STACK = 0x02, /* nudger will free the nudge thread's stack so the
                                      * nudge thread itself shouldn't */
     NUDGE_FREE_ARG = 0x04,          /* nudge arg is in a separate allocation and should
@@ -953,9 +954,9 @@ typedef struct {
     uint flags : 2;
     int ignored2; /* siginfo_t.si_code: has meaning to kernel so we avoid using */
 #else
-    uint version;           /* version number for future proofing */
-    uint nudge_action_mask; /* drawn from NUDGE_DEFS above */
-    uint flags;             /* flags drawn from above enum */
+    uint version;                   /* version number for future proofing */
+    uint nudge_action_mask;         /* drawn from NUDGE_DEFS above */
+    uint flags;                     /* flags drawn from above enum */
 #endif
     client_id_t client_id; /* unique ID identifying client */
     uint64 client_arg;     /* argument for a client nudge */
--- a/core/unix/nudgesig.c
+++ b/core/unix/nudgesig.c
@@ -67,7 +67,8 @@ create_nudge_signal_payload(kernel_siginfo_t *info OUT, uint action_mask, uint f
     arg->client_arg = client_arg;
 
     /* ensure nudge_arg_t overlays how we expect it to */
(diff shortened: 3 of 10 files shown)


derekbruening:
Still hitting problems with QEMU limitations: it seems to not honor our siginfo fields sent through SYS_rt_tgsigqueueinfo, clearing si_errno!  Once again going to have to use a special workaround.

derekbruening:
@abhinav92003 PTAL

api/docs/debugging.dox
```diff
@@ -78,12 +78,10 @@ the shared library that identifies the debug file.
 
 ## Expected Signals
 
-On Linux, DR sends itself a SIGILL signal during initialization, in order to measure the size of the signal frame used by the kernel.  If you are under gdb, simply continue past this signal.
+On Linux, DR sends itself a SIGILL signal during initialization, in order to measure the size of the signal frame used by the kernel.  If you are under gdb, simply continue past this signal.  SIGILL (SIGFPE on MacOS; SIGSTKFLT inside QEMU) is also sent to other threads to suspend or terminate them later on during execution.
```
abhinav92003, line 81: Now SIGILL is used also for attach right?

core/unix/nudgesig.c
```diff
@@ -50,7 +50,7 @@
 
 /* shared with tools/nudgeunix.c */
 bool
-create_nudge_signal_payload(kernel_siginfo_t *info OUT, uint action_mask,
+create_nudge_signal_payload(kernel_siginfo_t *info OUT, uint action_mask, uint flags,
                             client_id_t client_id, uint64 client_arg)
 {
```
abhinav92003, line 55: Assert that flags is not more than 2 bits.

core/lib/globals_shared.h
```diff
@@ -928,6 +928,7 @@ typedef enum {
  */
 enum {
     NUDGE_IS_INTERNAL = 0x01, /* nudge is internally generated */
+    NUDGE_IS_SUSPEND = 0x02,  /* This is an internal SUSPEND_SIGNAL. */
 #ifdef WINDOWS
     NUDGE_NUDGER_FREE_STACK = 0x02, /* nudger will free the nudge thread's stack so the
                                      * nudge thread itself shouldn't */
```
abhinav92003, line 934: This goes into nudge_arg_t.flags right? But that field is only 2 bits and here we have 0x04 which won't fit.

core/unix/nudgesig.c
```diff
@@ -80,7 +81,7 @@ send_nudge_signal(process_id_t pid, uint action_mask, client_id_t client_id,
...
     kernel_siginfo_t info;
     int res;
-    if (!create_nudge_signal_payload(&info, action_mask, client_id, client_arg))
+    if (!create_nudge_signal_payload(&info, action_mask, 0, client_id, client_arg))
```
abhinav92003, line 84: There's no value corresponding to zero in nudge_arg_t flags.

api/docs/debugging.dox
abhinav92003, line 81: Also document lack of gdb support on qemu.

core/unix/os.c
```diff
@@ -3714,6 +3758,34 @@ os_thread_sleep(uint64 milliseconds)
...
+    if (is_sigqueueinfo_enosys) {
+        /* We won't be able to distinguish a suspend from a nudge, but we don't
+         * support nudges on old Linux kernels anyway (<2.6.31).
+         */
+        if (tr != NULL)
+            return known_thread_signal(tr, SUSPEND_SIGNAL);
+        else
+            return thread_signal(pid, tid, SUSPEND_SIGNAL);
+    }
+    kernel_siginfo_t info;
```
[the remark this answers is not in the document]
abhinav92003: I see we always set is_sigqueueinfo_enosys for QEMU, so we won't reach here. Add a comment here.

core/unix/signal.c
```diff
@@ -6037,6 +6020,22 @@ main_signal_handler_C(byte *xsp)
 #endif
 
     default: {
+        /* Handle nudges and suspends, which may use the same signal. */
+        if (sig == SUSPEND_SIGNAL) {
+            if (handle_suspend_signal(dcontext, siginfo, ucxt, frame)) {
+                /* i#1921: see comment above */
+                ASSERT(tr == NULL || tr->under_dynamo_control ||
```
abhinav92003, line 6027: Is this assert valid for nudges too?

core/unix/signal.c
```diff
@@ -8329,6 +8328,39 @@ sig_detach(dcontext_t *dcontext, sigframe_rt_t *frame, KSYNCH_TYPE *detached)
...
+    if (siginfo->si_signo != sig_expecting)
+        return false;
+    if (is_sigqueue_supported()) {
+        if (siginfo->si_code != SI_QUEUE || siginfo->si_errno == 0)
+            return false;
+    } else {
+        /* Distinguish from a synchronous app signal by looking for
+         * si_code where <=0 means user-sent.
+         */
+        if (siginfo->si_code > 0)
```
abhinav92003, line 8358: So our signal will have > 0? Do we set it ourselves? if not, how is it set?

core/unix/signal.c
```diff
@@ -8339,15 +8371,20 @@ handle_suspend_signal(dcontext_t *dcontext, kernel_siginfo_t *siginfo,
...
-     * We distinguish further below from the rare case of an app sending
-     * SUSPEND_SIGNAL asynchronously by looking for our particular state settings
-     * that correspond to the use of this signal by DR..
-     */
-    if (siginfo->si_code > 0 || siginfo->si_signo != SUSPEND_SIGNAL)
+    if (!is_signal_for_us(siginfo, SUSPEND_SIGNAL))
         return true; /* pass to app */
 
+    if (is_sigqueue_supported() && SUSPEND_SIGNAL == NUDGESIG_SIGNUM) {
+        nudge_arg_t *arg = (nudge_arg_t *)siginfo;
```
abhinav92003, line 8378: Should we further check is_signal_for_us(siginfo, NUDGESIG_SIGNUM)?

tools/nudgeunix.c
```diff
@@ -140,7 +140,7 @@ main(int argc, const char *argv[])
         return usage();
 
     /* construct the payload */
-    success = create_nudge_signal_payload(&info, action_mask, client_id, client_arg);
+    success = create_nudge_signal_payload(&info, action_mask, 0, client_id, client_arg);
```
abhinav92003, line 143: Is zero a proper value in the flags enum?

abhinav92003 (approved):

core/unix/os.c
```diff
@@ -514,8 +514,11 @@ typedef int *(*errno_loc_t)(void);
 #ifdef LINUX
 /* Stores whether clone3 is unsupported on the system we're running on. */
```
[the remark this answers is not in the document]
derekbruening: Done.

core/unix/os.c
```diff
@@ -892,9 +895,27 @@ detect_unsupported_syscalls()
         dynamorio_syscall(SYS_clone3, 2, NULL /*clone_args*/, 0 /*clone_args_size*/);
     ASSERT(clone3_errno == -ENOSYS || clone3_errno == -EINVAL);
     is_clone3_enosys = clone3_errno == -ENOSYS;
+    int sigqueue_errno = dynamorio_syscall(SYS_rt_tgsigqueueinfo, 4, get_process_id(),
+                                           get_sys_thread_id(), -1, NULL);
+    ASSERT(sigqueue_errno == -ENOSYS || sigqueue_errno == -EINVAL ||
```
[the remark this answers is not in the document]
derekbruening: The NULL raises EFAULT but you could imagine a kernel giving EINVAL for the -1; really we just want it to fail so either way is fine.  Added a comment.

core/unix/os.c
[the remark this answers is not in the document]
derekbruening: Even if the suspend sig != nudge sig, we would prefer to use the payload to better distinguish a suspend sig from an app using that sig: the only reason we wouldn't want to get here is if we can't use sigqueueinfo for some reason, which the if() asking whether it's available seems clear enough I would think.

core/unix/signal.c
```diff
@@ -6037,6 +6020,22 @@ main_signal_handler_C(byte *xsp)
...
+        if (sig == SUSPEND_SIGNAL) {
+            if (handle_suspend_signal(dcontext, siginfo, ucxt, frame)) {
+                /* i#1921: see comment above */
+                ASSERT(tr == NULL || tr->under_dynamo_control ||
+                       IS_CLIENT_THREAD(dcontext));
+                record_pending_signal(dcontext, sig, ucxt, frame, false, NULL);
+            }
+            /* else, don't deliver to app */
+            break;
+        } else if (sig == NUDGESIG_SIGNUM) {
```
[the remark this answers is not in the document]
derekbruening: That feels less safe as a change to how signals are delivered to not use a break would then result in double-handling the same signal.

core/unix/signal.c
```diff
@@ -6037,6 +6020,22 @@ main_signal_handler_C(byte *xsp)
 #endif
 
     default: {
+        /* Handle nudges and suspends, which may use the same signal. */
+        if (sig == SUSPEND_SIGNAL) {
+            if (handle_suspend_signal(dcontext, siginfo, ucxt, frame)) {
```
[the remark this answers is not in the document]
derekbruening: The comment already there at line 6023 says "Handle nudges and suspends, which may use the same signal."  Added another comment at the else.

> Also currently we're taking the same action (record_pending_signal) for nudges and suspend,

That's not the action for an actual nudge or suspend: that's if this signal is *not* one of those and is instead meant for the app, which is what we do for all signals for the app.

core/unix/nudgesig.c
```diff
@@ -62,12 +62,13 @@ create_nudge_signal_payload(kernel_siginfo_t *info OUT, uint action_mask,
     arg = (nudge_arg_t *)info;
     arg->version = NUDGE_ARG_CURRENT_VERSION;
     arg->nudge_action_mask = action_mask;
-    arg->flags = 0;
+    arg->flags = flags;
     arg->client_id = client_id;
     arg->client_arg = client_arg;
 
     /* ensure nudge_arg_t overlays how we expect it to */
```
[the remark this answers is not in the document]
derekbruening: This is used in several different libraries with different error reporting schemes: I think it is better to let them handle the failure which is the return value here.

api/docs/debugging.dox
derekbruening: See the next sentence "SIGILL...is also sent..."

core/unix/nudgesig.c
derekbruening: Done.

core/lib/globals_shared.h
derekbruening: The flags field is 32 bits on Windows.

derekbruening: (There is already a reminder comment for the 2 bits just on Linux right above this enum)

api/docs/debugging.dox
derekbruening: Added a mention.  There are many issues with qemu and I think using gdb on DR on QEMU is already not very workable even w/o SIGSTKFLT.

core/unix/signal.c
derekbruening: Yes, this is already asserted at line 5739 for all signals but with an extra || sig == SUSPEND_SIGNAL b/c we use that on unknown threads.

core/unix/signal.c
derekbruening: The kernel sets this field.

core/unix/signal.c
derekbruening: But if they're the same number we've already called that.

tools/nudgeunix.c
derekbruening: See earlier reply: this is not an enum typed parameter but an integer.

derekbruening:
a64 hit windows-zlib #5493
win64 hit #5655

PR #5657, opened 2022-09-20
i#i5656 raw2trace: Delay function markers

### LINKED ISSUE #5656: Function markers inserted inside a basic block result in invariant error
Function markers inserted inside a basic block resulting in "Function marker should be after a branch" invariant error.

In this case the function markers are inserted inside the basic block. 
```
      874: T265782 ifetch       1 byte(s) @ 0x0000000003c37188 5d                   pop    %rbp
      875: T265782 read         8 byte(s) @ 0x00007f7e11f169a0 by PC 0x0000000003c37188
      876: T265782 ifetch       1 byte(s) @ 0x0000000003c37189 c3                   ret
      877: T265782 read         8 byte(s) @ 0x00007f7e11f169a8 by PC 0x0000000003c37189
      878: T265782 ifetch       3 byte(s) @ 0x0000000003d78868 49 89 c5             mov    %rax, %r13
      879: T265782 ifetch       3 byte(s) @ 0x0000000003d7886b 41 89 d6             mov    %edx, %r14d
      880: T265782 ifetch       5 byte(s) @ 0x0000000003d7886e bf 10 77 00 00       mov    $0x00007710, %edi
      881: T265782 <marker: function #3>
      882: T265782 <marker: function return address 0x3d78878>
      883: T265782 <marker: function argument 0x7710>
      884: T265782 ifetch       5 byte(s) @ 0x0000000003d78873 e8 08 4d f3 ff       call   $0x0000000003cad580
      885: T265782 write        8 byte(s) @ 0x00007f7e11f169a8 by PC 0x0000000003d78873
      886: T265782 ifetch       7 byte(s) @ 0x0000000003cad580 48 81 ff 01 04 00 00 cmp    %rdi, $0x00000401
      887: T265782 ifetch       6 byte(s) @ 0x0000000003cad587 0f 83 86 00 00 00    jnb    $0x0000000003cad613
      888: T265782 ifetch       7 byte(s) @ 0x0000000003cad613 48 81 ff 01 00 04 00 cmp    %rdi, $0x00040001

```
In this case the function markers are inserted between records for a single instruction. 
```
     6642: T805696 ifetch       4 byte(s) @ 0x00000000099d5844 95e8e51e   bl     $0x000000001140ecbc -> %x30
     6643: T805696 ifetch       4 byte(s) @ 0x000000001140ecbc 58000050   ldr    <rel> 0x000000001140ecc4[8byte] -> %x16
     6644: T805696 <marker: function #3>
     6645: T805696 <marker: function return address 0x99d5848>
     6646: T805696 <marker: function argument 0x18>
     6647: T805696 read         8 byte(s) @ 0x000000001140ecc4 by PC 0x000000001140ecbc
     6648: T805696 ifetch       4 byte(s) @ 0x000000001140ecc0 d61f0200   br     %x16
     6649: T805696 ifetch       4 byte(s) @ 0x0000000019413500 aa0003e8   orr    %xzr %x0 lsl $0x00 -> %x8
```

**Expected behavior**
The function markers should not be inserted between an instruction and its memrefs; This can cause issues with tools that do not expect markers amid records for a single instruction.

bete0:
Adds function markers to delayed_branch vector in
raw2trace_thread_data_t so that the markers are not inserted inside a basic block (e.g: between a branch and its memref).

Tested: Run raw2trace on a trace that previously had "Function marker should be after a branch" invariant error and with this change that invariant error goes away.

Fixes #5656


clients/drcachesim/tracer/raw2trace.h
```diff
@@ -797,12 +797,23 @@ template <typename T> class trace_converter_t {
...
+                    if (!error.empty())
+                        return error;
+                    return "";
+                } else {
+                    if (in_entry->extended.valueB == TRACE_MARKER_TYPE_KERNEL_EVENT) {
+                        impl()->log(4, "Signal/exception between bbs\n");
+                    }
+                    impl()->log(3, "Appended marker type %u value " PIFX "\n",
+                                (trace_marker_type_t)in_entry->extended.valueB,
+                                (uintptr_t)in_entry->extended.valueA);
```
derekbruening, line 815: We discussed offline that there should be new synthetic tests for this to augment the existing ones which we looked at but I don't see them in this PR?  They should be part of this same PR.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -797,12 +797,23 @@ template <typename T> class trace_converter_t {
                     return err;
                 buf += trace_metadata_writer_t::write_marker(
                     buf, (trace_marker_type_t)in_entry->extended.valueB, marker_val);
-                if (in_entry->extended.valueB == TRACE_MARKER_TYPE_KERNEL_EVENT) {
-                    impl()->log(4, "Signal/exception between bbs\n");
+                if (in_entry->extended.valueB == TRACE_MARKER_TYPE_FUNC_ARG ||
```
derekbruening, line 800: It seems like any possible intra-block marker needs to be appended by handle_kernel_interrupt_and_markers() so this should include all markers.  Please augment the comments in handle_kernel_interrupt_and_markers() to make this clear as well ("only at block boundaries" comment plus up top?).

derekbruening: But it should not include headers: cpu markers.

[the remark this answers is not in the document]
bete0: Done.

bete0: Done. Let me know if that's what you had in mind.

derekbruening:
Looks like the new unit tests are failing?

bete0:
> Looks like the new unit tests are failing?

commit [1955c0a] fixed the unit test failures but looks like there is an unrelated failure in x86-64 tests (https://github.com/DynamoRIO/dynamorio/actions/runs/3142931057/jobs/5108042408#step:7:32204)

derekbruening:
> commit [[1955c0a](https://github.com/DynamoRIO/dynamorio/commit/1955c0aafbb29e1128094fde5334bb251ae6233b)] fixed the unit test failures but looks like there is an unrelated failure in x86-64 tests (https://github.com/DynamoRIO/dynamorio/actions/runs/3142931057/jobs/5108042408#step:7:32204)

Always cite the bug number covering it; if it is new file a new issue and cite that.

clients/drcachesim/tests/raw2trace_unit_tests.cpp
```diff
@@ -367,6 +382,11 @@ test_marker_placement(void *drcontext)
     raw.push_back(make_marker(TRACE_MARKER_TYPE_FUNC_ID, 0));
     raw.push_back(make_marker(TRACE_MARKER_TYPE_FUNC_RETADDR, 4));
     raw.push_back(make_marker(TRACE_MARKER_TYPE_FUNC_ARG, 2));
+    raw.push_back(make_block(offs_load2, 1));
+    raw.push_back(make_marker(TRACE_MARKER_TYPE_FUNC_ID, 0));
+    raw.push_back(make_marker(TRACE_MARKER_TYPE_FUNC_RETADDR, 4));
+    raw.push_back(make_marker(TRACE_MARKER_TYPE_FUNC_ARG, 2));
+    raw.push_back(make_block(offs_jmp, 1));
```
derekbruening, line 389: I don't think this test is actually testing #5656 since there's nothing for the branch to be delayed after so it will always be after the markers?  If you disable the fix does this test fail?  This does not look sufficient.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -797,12 +797,20 @@ template <typename T> class trace_converter_t {
                     return err;
                 buf += trace_metadata_writer_t::write_marker(
                     buf, (trace_marker_type_t)in_entry->extended.valueB, marker_val);
-                if (in_entry->extended.valueB == TRACE_MARKER_TYPE_KERNEL_EVENT) {
-                    impl()->log(4, "Signal/exception between bbs\n");
+                if (in_entry->extended.valueB != TRACE_MARKER_TYPE_CPU_ID) {
```
derekbruening, line 800: Please add a comment explaining why a cpu id marker is different.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -797,12 +797,20 @@ template <typename T> class trace_converter_t {
...
+                    std::string error = impl()->write_delayed_branches(
```
derekbruening, line 801: Add a comment explaining why we put most markers in the delay queue instead of just outputting them.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -797,12 +797,20 @@ template <typename T> class trace_converter_t {
...
+                        tls, buf_base, reinterpret_cast<trace_entry_t *>(buf));
+                    if (!error.empty())
+                        return error;
+                    return "";
+                } else {
+                    if (in_entry->extended.valueB == TRACE_MARKER_TYPE_KERNEL_EVENT) {
```
derekbruening, line 807: This if is impossible to enter: it needs to be on the other path.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1424,11 +1434,10 @@ template <typename T> class trace_converter_t {
                     // not-inside-a-block main loop.
                 }
             } else {
-                // Other than kernel event markers checked above, markers should be
-                // only at block boundaries, as we cannot figure out where they should go
-                // (and could easily insert them in the middle of this block instead
-                // of between this and the next block, with implicit instructions added).
-                // Thus, we do not append any other markers.
+                // All markers should be only at block boundaries, as we cannot
```
derekbruening, line 1437: This is not accurate: kernel markers can be intra-block.  The original comment should probably be left alone.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1424,11 +1434,10 @@ template <typename T> class trace_converter_t {
...
+                // figure out where they should go (and could easily insert them in the
+                // middle of this block instead of between this and the next block, with
+                // implicit instructions added). Thus, we do not append any other markers.
```
derekbruening, line 1440: I think my prior comment was asking to explain here and in the top-level comment for this function that *all* intra-block markers are inserted by this function (i.e., the higher level process_offline_entry() will never insert a marker intra-block) and all inter-block markers are handled at a higher level (process_offline_entry()) and are never inserted here.

clients/drcachesim/tracer/raw2trace.h
bete0: Done.

clients/drcachesim/tracer/raw2trace.h
bete0: Added this at the top-level comment for the function.

clients/drcachesim/tracer/raw2trace.h
bete0: Done.

clients/drcachesim/tracer/raw2trace.h
bete0: Done.

clients/drcachesim/tests/raw2trace_unit_tests.cpp
bete0: Added a new function `test_delay_markers()` and updated the test case. Thought it would be cleaner to separate it but the common parts for the three test functions could shared. Will do that in a separate CL.

bete0:
> > commit [[1955c0a](https://github.com/DynamoRIO/dynamorio/commit/1955c0aafbb29e1128094fde5334bb251ae6233b)] fixed the unit test failures but looks like there is an unrelated failure in x86-64 tests (https://github.com/DynamoRIO/dynamorio/actions/runs/3142931057/jobs/5108042408#step:7:32204)
> 
> Always cite the bug number covering it; if it is new file a new issue and cite that.

This turned out to be something related to the CL so fixed it.

derekbruening:
(Please re-request review when ready.)

bete0:
> (Please re-request review when ready.)

Looking into some tests failing but is ready for ready.

derekbruening:
> > (Please re-request review when ready.)
> 
> Looking into some tests failing but is ready for ready.

Generally we wait for a green CI before reviewing and for all prior comments to be resolved (except those looking for explicit additional feedback).

clients/drcachesim/tests/raw2trace_unit_tests.cpp
```diff
@@ -327,31 +327,46 @@ test_marker_placement(void *drcontext)
     //    between the instrs in the block.
     // 2) A block with an implicit memref for the first instr, to reproduce i#5620
     //    where markers should wait for the memref (and subsequent implicit instrs).
+    // 3) A block with an implicit memref for the first instr, to reproduce
+    //    i##5656 where markers (along with branches) should be delayed.
```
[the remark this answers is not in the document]
bete0: Updated the test and moved it to a separate function.

bete0:
> > > (Please re-request review when ready.)
> > 
> > 
> > Looking into some tests failing but is ready for ready.
> 
> Generally we wait for a green CI before reviewing and for all prior comments to be resolved (except those looking for explicit additional feedback).

All tests are passing now.

clients/drcachesim/tests/offline-phys.templatex
```diff
@@ -15,15 +15,15 @@ Output format:
...
-        5: T[0-9]+ <marker: page size [0-9]+>
-        6: T[0-9]+ <marker: timestamp [0-9]+>
-        7: T[0-9]+ <marker: tid [0-9]+ on core [0-9]+>
-        8: T[0-9]+ <marker: physical address for following virtual: 0x[0-9a-f][0-9a-f]+>
-        9: T[0-9]+ <marker: virtual address for prior physical: 0x[0-9a-f][0-9a-f]+>
+        5: T[0-9]+ <marker: timestamp [0-9]+>
+        6: T[0-9]+ <marker: tid [0-9]+ on core [0-9]+>
+        7: T[0-9]+ <marker: timestamp [0-9]+>
+        8: T[0-9]+ <marker: tid [0-9]+ on core [0-9]+>
+        9: T[0-9]+ <marker: page size [0-9]+>
```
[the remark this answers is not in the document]
bete0: Are you suggesting we go back to the original approach of delaying only function markers?

clients/drcachesim/tests/raw2trace_unit_tests.cpp
```diff
@@ -447,12 +447,136 @@ test_marker_placement(void *drcontext)
...
+{
+    // Our synthetic test first constructs a list of instructions to be encoded into
+    // a buffer for decoding by raw2trace.
+    instrlist_t *ilist = instrlist_create(drcontext);
+    // raw2trace doesn't like offsets of 0 so we shift with a nop.
+    instr_t *nop = XINST_CREATE_nop(drcontext);
+    instr_t *move =
+        XINST_CREATE_move(drcontext, opnd_create_reg(REG1), opnd_create_reg(REG2));
+#ifdef AARCH64
+    // XXX i#5628: opnd_create_mem_instr is not supported yet on AArch64.
```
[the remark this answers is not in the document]
bete0: Ack. Removed the loads.

clients/drcachesim/tests/raw2trace_unit_tests.cpp
```diff
@@ -447,12 +447,136 @@ test_marker_placement(void *drcontext)
...
+                                      OPND_CREATE_ABSMEM(nop, OPSZ_PTR));
+    instr_t *load2 = INSTR_CREATE_ldr(drcontext, opnd_create_reg(REG1),
+                                      OPND_CREATE_ABSMEM(load1, OPSZ_PTR));
+#else
+    instr_t *load1 = XINST_CREATE_load(drcontext, opnd_create_reg(REG1),
+                                       opnd_create_mem_instr(nop, 0, OPSZ_PTR));
+    instr_t *load2 = XINST_CREATE_load(drcontext, opnd_create_reg(REG1),
+                                       opnd_create_mem_instr(load1, 0, OPSZ_PTR));
+#endif
+    instr_t *jmp = XINST_CREATE_jump(drcontext, opnd_create_instr(load1));
```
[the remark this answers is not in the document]
bete0: Done.

clients/drcachesim/tests/raw2trace_unit_tests.cpp
```diff
@@ -447,12 +447,136 @@ test_marker_placement(void *drcontext)
...
+    instrlist_append(ilist, nop);
+    instrlist_append(ilist, load1);
+    instrlist_append(ilist, load2);
```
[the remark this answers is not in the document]
bete0: Done.

clients/drcachesim/tests/raw2trace_unit_tests.cpp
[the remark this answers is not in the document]
bete0: Removed and put the new test cases in a separate function.

clients/drcachesim/tests/offline-phys.templatex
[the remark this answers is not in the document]
derekbruening: Not sure: maybe categorize the markers and think about future markers?  I wasn't considering header markers but it seemed like other inter-block markers should move.  But we don't want to delay when there was no prior branch delay: maybe that is the solution: don't wait for an instr if there's no reason to.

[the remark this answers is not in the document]
bete0: > But we don't want to delay when there was no prior branch delay

How do we know there was no prior branch delay? Check if `delayed_branch` vec is empty and append the markers only if it is empty?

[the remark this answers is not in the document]
derekbruening: Something like that, probably via an interface?  The general idea of delaying any marker that wasn't handled by the intra-block code when we're going to delay a branch seems nicer to me than trying to have a list of markers and it seems like the only way to handle the v2p ones.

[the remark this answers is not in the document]
bete0: Updated the logic to check for a prior a branch delay before delaying the markers. Also updated the templates now that the other non-function markers went back to their original location.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -2087,6 +2110,10 @@ class raw2trace_t : public trace_converter_t<raw2trace_t> {
 
     // Chunking for seeking support in compressed files.
     uint64_t chunk_instr_count_ = 0;
+
+    // True if there was a prior branch delay. Used to indicate when to also
```
derekbruening, line 2114: I don't understand this logic: what does it mean that a branch was delayed at some point somewhere in the past?  Is this just trying to avoid the start-of-trace headers?  I would expect something specific to the start of the trace then and not a check for branches.  My proposal was to only delay if there is *currently* a delayed branch that has not been emitted yet, which seems the safest and clearest solution IMHO, avoiding delays we don't want regardless of where in the trace we are.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1339,9 +1355,12 @@ template <typename T> class trace_converter_t {
                 // than here (and we are ok bailing on doing this for online traces), so
                 // we handle it in post-processing by delaying a thread-block-final branch
                 // (and its memrefs) to that thread's next block.  This changes the
-                // timestamp of the branch, which we live with.
+                // timestamp of the branch, which we live with. To avoid marker
+                // misplacement (e.g. in the middle of a basic block), we also
+                // delay markers.
                 impl()->log(4, "Delaying %d entries\n", buf - buf_start);
                 error = impl()->write_delayed_branches(tls, buf_start, buf);
+                impl()->prior_branch_delay_ = true;
```
derekbruening, line 1363: The interface is supposed to use functions, not raw fields.

clients/drcachesim/tests/raw2trace_unit_tests.cpp
```diff
@@ -447,12 +447,112 @@ test_marker_placement(void *drcontext)
...
+        check_entry(entries, idx, TRACE_TYPE_INSTR_DIRECT_JUMP, -1) &&
+        // Markers should be delayed along with branches until the end of the
+        // basic block.
+        check_entry(entries, idx, TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_FUNC_ID) &&
+        check_entry(entries, idx, TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_FUNC_RETADDR) &&
+        check_entry(entries, idx, TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_FUNC_ARG) &&
+        check_entry(entries, idx, TRACE_TYPE_ENCODING, -1) &&
+        check_entry(entries, idx, TRACE_TYPE_INSTR, -1) &&
+        check_entry(entries, idx, TRACE_TYPE_THREAD_EXIT, -1) &&
+        check_entry(entries, idx, TRACE_TYPE_FOOTER, -1));
```
derekbruening, line 546: Please add tests of the cases we just discovered:
+ Make sure markers are not delayed across timestamp+cpu headers if there's no branch also being delayed (e.g., the header case -- need the page size or whichever one it was; cache line must be specially treated for this current test to not have failed in the last commit).
+ Make sure markers are not delayed across window boundaries (TRACE_MARKER_TYPE_WINDOW_ID with a new id).

It would also be good to have a sanity test that timestamps and cpu ids are *not* delayed along with branches.

Let's not have future changes accidentally regress and we don't notice until invariants fire on larger traces (as happened in the past): let's cover the cases we've seen in simple regression tests.

clients/drcachesim/tracer/raw2trace.h
bete0: Acknowledged. Updated to use the `delayed_branch` vector instead.

clients/drcachesim/tests/raw2trace_unit_tests.cpp
bete0: Ack. Will add the new test cases in the next commit. Just wanted to make sure the logic above is green.

bete0:
run arm tests

clients/drcachesim/tests/raw2trace_unit_tests.cpp
```diff
@@ -447,12 +447,112 @@ test_marker_placement(void *drcontext)
         check_entry(entries, idx, TRACE_TYPE_FOOTER, -1));
 }
 
+bool
+test_marker_delays(void *drcontext)
```
bete0, line 451: Update test cases based on https://github.com/DynamoRIO/dynamorio/pull/5657#discussion_r989648229

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -1238,6 +1238,13 @@ raw2trace_t::write_delayed_branches(void *tls, const trace_entry_t *start,
     return "";
 }
 
+bool
+raw2trace_t::delayed_branches_exit(void *tls)
```
derekbruening, line 1242: spelling: exist

Ditto on 2 other instances

bete0:
run arm tests

clients/drcachesim/tests/raw2trace_unit_tests.cpp
bete0: Added the suggested test cases.

clients/drcachesim/tests/raw2trace_unit_tests.cpp
```diff
@@ -447,12 +458,158 @@ test_marker_placement(void *drcontext)
...
+        check_entry(entries, idx, TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_TIMESTAMP) &&
+        check_entry(entries, idx, TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_CPU_ID) &&
+        check_entry(entries, idx, TRACE_TYPE_ENCODING, -1) &&
+#ifdef X86_32
+        // An extra encoding entry is needed.
+        check_entry(entries, idx, TRACE_TYPE_ENCODING, -1) &&
+#endif
+        check_entry(entries, idx, TRACE_TYPE_INSTR_DIRECT_JUMP, -1) &&
+        // Markers should be delayed along with branches until the end of the
+        // basic block.
```
derekbruening, line 583: I don't understand "until the end of the basic block": which block?  The branch was delayed from the prior block, and is inserted before the next block.  Which end of block is this referring to?  I suspect this needs to be reworded.

derekbruening, line 583: This is a long sequence, and it is testing 3 different things.  Please add comments labeling each test to make it easier to understand where each test occurs in this sequence.

clients/drcachesim/tests/raw2trace_unit_tests.cpp
```diff
@@ -447,12 +458,158 @@ test_marker_placement(void *drcontext)
...
+    size_t offs_move3 = offs_move2 + instr_length(drcontext, move2);
+    size_t offs_move4 = offs_move3 + instr_length(drcontext, move3);
+
+    // Now we synthesize our raw trace itself, including a valid header sequence.
+    std::vector<offline_entry_t> raw;
+    raw.push_back(make_header());
+    raw.push_back(make_tid());
+    raw.push_back(make_pid());
+    raw.push_back(make_line_size());
+    raw.push_back(make_block(offs_move1, 2));
```
derekbruening, line 511: Comments here on which part belongs to which of tests 1, 2, and 3 would help.  So this is 1: the branch at the end of this block will be delayed until the next block is found: but it should cross the timestmap+cpu below, and carry the 3 func markers with it and not pass over those.  Please put something like this in a comment.

clients/drcachesim/tests/raw2trace_unit_tests.cpp
```diff
@@ -447,12 +458,158 @@ test_marker_placement(void *drcontext)
...
+    raw.push_back(make_timestamp());
+    raw.push_back(make_core());
+    raw.push_back(make_marker(TRACE_MARKER_TYPE_FUNC_ID, 0));
+    raw.push_back(make_marker(TRACE_MARKER_TYPE_FUNC_RETADDR, 4));
+    raw.push_back(make_marker(TRACE_MARKER_TYPE_FUNC_ARG, 2));
+    raw.push_back(make_block(offs_move2, 2));
+    raw.push_back(make_marker(TRACE_MARKER_TYPE_FUNC_ID, 0));
+    raw.push_back(make_marker(TRACE_MARKER_TYPE_FUNC_RETADDR, 4));
+    raw.push_back(make_marker(TRACE_MARKER_TYPE_FUNC_ARG, 2));
```
derekbruening, line 520: OK this must be test 2: func markers w/ no branch followed by timestamp+cpu

clients/drcachesim/tests/raw2trace_unit_tests.cpp
```diff
@@ -199,6 +199,17 @@ make_core()
     return entry;
 }
 
+offline_entry_t
+make_window_id()
```
derekbruening, line 203: This should probably take in the id as otherwise it will not be possible to have 2 new window ids in one sequence.

clients/drcachesim/tests/raw2trace_unit_tests.cpp
```diff
@@ -447,12 +458,158 @@ test_marker_placement(void *drcontext)
...
+        check_entry(entries, idx, TRACE_TYPE_INSTR, -1) &&
+        check_entry(entries, idx, TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_FUNC_ID) &&
+        check_entry(entries, idx, TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_FUNC_RETADDR) &&
+        check_entry(entries, idx, TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_FUNC_ARG) &&
+        check_entry(entries, idx, TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_TIMESTAMP) &&
+        check_entry(entries, idx, TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_CPU_ID) &&
+        check_entry(entries, idx, TRACE_TYPE_ENCODING, -1) &&
+        check_entry(entries, idx, TRACE_TYPE_INSTR, -1) &&
+        check_entry(entries, idx, TRACE_TYPE_ENCODING, -1) &&
+        check_entry(entries, idx, TRACE_TYPE_INSTR, -1) &&
```
derekbruening, line 599: Better to have a branch at the end of this block and ensure it is not delayed past the new window either.

derekbruening (approved):

clients/drcachesim/tests/raw2trace_unit_tests.cpp
bete0: Done. Added clarifying comments.

bete0: Ack. Removed this and added specific comments for the different cases.

Issue #2039, opened 2016-10-20
[drcachesim] trim trace to avoid staggered attach and detach partial-thread anomalies at ends of trace


derekbruening:
Xref #1729 

Due to the asynchronous nature of stopping all the threads on detach, some
may run further and produce more data than we'd like for a bursty trace.  I
already added a new option -max_trace_size to put a cap on the per-thread
trace size, but we probably also want a way to precisely throw out data
past the detach timestamp.

On the burst-threads test, without the -max_trace_size cap, the secondary
threads produce surprisingly large traces: 65MB, when the start/stop thread
only ran 4 iters of its loop.  It may be worth investigating further to
ensure nothing is going wrong there: is it really just a few hundred more
iters that adds up to 65MB??

derekbruening:
This problem occurs on attach as well: the staggered attach means that in an app with thousands of threads the threads taken over early start producing trace data well before the rest of the threads are taken over, resulting in an inaccurate initial picture of the workload.

An alternative to discarding after the fact based on timestamps is to use drbbdup modes to have a non-tracing alternative mode.  Only once attach has taken over all threads would we flip a switch and have the tracing mode start, simultaneously in all threads.  We would use a load-acquire read of the mode flag which is fairly performant.

While this is more overhead than fixing up later, it fits well with schemes that pre-populate the code cache in a lower-overhead no-tracing mode before starting tracing.

There are two ways to trigger the mode switches for statically linked drmemtrace: new DR events for post-attach and pre-detach, or annotations.  Annotations are more flexible as when pre-populating the tool can separate attach from tracing.

We would want to implement annotations for aarchxx i#1672, and solve the raw2trace issues with annotations #4141.

derekbruening:
From https://github.com/DynamoRIO/dynamorio/pull/5704#discussion_r1007187555: we should add an invariant check that timestamps always increase monotonically through each trace.

PR #5660, opened 2022-09-26
i#5659: Handle multiple app copies in rseq mangling

### LINKED ISSUE #5659: rseq mangling does not support multiple app copies from drbbdup
I hit this, like in #5658, when using drbbdup cases with drmemtrace in larger apps that use rseq.

The rseq finalization is what assumes a single sequence: both in its instrlist search and in what it stores per fragment.

derekbruening:
Adds support for multiple app copies in rseq mangling finalization's search for labels to fill in code cache PC's.

Adds support for multiple rseq regions in one fragment_t by expanding the stored rseq_cs per fragment to become a list.

Adds "-trace_after_instrs 5K" to the tool.drcacheoff.rseq test to exercise rseq with drbbdup.

Augments the attach/detach api.rseq test to use drmemtrace, which with the forthcoming #2039 mode switches also hits this multi-rseq case.

Fixes #5659

### DIFF
--- a/core/unix/rseq_linux.c
+++ b/core/unix/rseq_linux.c
@@ -99,7 +99,9 @@ typedef struct _rseq_region_t {
     bool reg_written[DR_NUM_GPR_REGS];
 } rseq_region_t;
 
-/* We need to store potentially multiple rseq_cs per fragment. */
+/* We need to store potentially multiple rseq_cs per fragment when clients
+ * make multiple copies of the app code (e.g., drbbdup).
+ */
 typedef struct _rseq_cs_record_t {
     struct rseq_cs rcs;
     void *alloc_ptr;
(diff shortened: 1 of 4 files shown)


derekbruening:
The a64 failure is #5493 windows-zlib.

core/unix/rseq_linux.c
```diff
@@ -99,7 +99,14 @@ typedef struct _rseq_region_t {
     bool reg_written[DR_NUM_GPR_REGS];
 } rseq_region_t;
 
-/* We need to store a struct rseq_cs per fragment_t.  To avoid the cost of adding a
+/* We need to store potentially multiple rseq_cs per fragment. */
```
abhinav92003, line 102: Maybe comment that the multiple rseq_cs is because of drbbdup, and not because of the app.

abhinav92003 (approved):

core/unix/rseq_linux.c
derekbruening: Done.

Issue #5658, opened 2022-09-26
drbbdup shifts drreg restore before non-cti block end, breaking rseq mangling


derekbruening:
I hit this when using drbbdup for more things in drmemtrace and running larger apps that use rseq.
These app reliably crash with SIGSEGV due to register corruption because drreg-reserved registers
are not restored prior to the copy of the natively-executed app rseq sequence.

rseq mangling says:
```
    /* We assume that by making this a block end, clients will restore app state
     * before this native invocation.
     * TODO i#2350: Take some further action to better guarantee this in the face
     * of future drreg optimizations, etc.  Do we need new interface features, or
     * do we live with a fake app jump or sthg?
     */
```

rseq truncates blocks after stores, so there is no cti or syscall at the end.
drbbdup disables drreg restore before the last instr:
```
    drreg_set_bb_properties(drcontext, DRREG_USER_RESTORES_AT_BB_END);
```
In a non-drbbdup block, drreg inserts *before* the last app instr, while bbdup inserts before
the inserted jump or exit label for the end of the bb copy: so *after* the last app instr,
when there's no ending cti/syscall.

How to solve?
Have DR insert a label "DR_NOTE_SPILL_BARRIER" that drreg looks for?
We already do this for DR_NOTE_ANNOTATION so this type of solution has precedence.
Xref #5160 on potential refactoring to put drreg into the core.

derekbruening:
We're going to leave this issue open to cover adding some tests of drbbdup interactions with drreg as it presents different characteristics than a regular block and drreg might not handle some aspects or might be in danger of regressing if we optimize drreg but don't have regression tests of drbbdup.

PR #5664, opened 2022-09-30
i#5658 rseq drreg: Put native sequence by barrier

derekbruening:
Fixes PR #5663 which fixed drreg restoration for drbbdup but broke it for non-drbbdup.  bb->instr was NULL so the preinsert was an append, and the label as the last instruction moved the drreg restore point to after the native sequence!

Here, we explicitly append to avoid confusion, and we move the native sequence to the point of the barrier instead of at the final app instr, which is where the registers are actually native.

Tested on larger apps using rseq which reliably crashed without this fix.  This fix also seems to fully fix drbbdup usage with drmemtrace for #2039.

Issue: #5658, #2039

### DIFF
--- a/core/arch/interp.c
+++ b/core/arch/interp.c
@@ -3972,6 +3972,10 @@ build_bb_ilist(dcontext_t *dcontext, build_bb_t *bb)
     if (TEST(FRAG_HAS_RSEQ_ENDPOINT, bb->flags)) {
         instr_t *label = INSTR_CREATE_label(dcontext);
         instr_set_note(label, (void *)DR_NOTE_REG_BARRIER);
+        /* We want the label after the final rseq instruction.  We've
+         * truncated the block after that instruction so bb->instr may
+         * be NULL so we append.
+         */
         instrlist_meta_append(bb->ilist, label);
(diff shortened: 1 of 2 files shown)


core/arch/interp.c
```diff
@@ -3972,7 +3972,7 @@ build_bb_ilist(dcontext_t *dcontext, build_bb_t *bb)
     if (TEST(FRAG_HAS_RSEQ_ENDPOINT, bb->flags)) {
         instr_t *label = INSTR_CREATE_label(dcontext);
         instr_set_note(label, (void *)DR_NOTE_REG_BARRIER);
-        instrlist_meta_preinsert(bb->ilist, bb->instr, label);
+        instrlist_meta_append(bb->ilist, label);
```
abhinav92003, line 3975: Worth adding an assert that bb->instr is null?

derekbruening: I think the preinsert was a mistake and we want to postinsert/append regardless to ensure the barrier is after the full 1st instru run of the rseq sequences and right before the 2nd native run.  Added a comment.

PR #5665, opened 2022-09-30
i#3544 riscv64: remove extension defines

semihalf-kardach-stanislaw:
Compile-time defines listing RISC-V extensions are not very useful for DynamoRIO because we may run on a system with a different set of extensions. So it should probably be a responsibility of the client code to figure out whether instructions it would like to use are compatible with the host it is running on.

In the future RISC-V arch code for DynamoRIO should perhaps provide a helper to iterate over all available extensions in runtime. This is not a simple function as the extension set is in fact a string list parsed from the device tree. Therefore it would need some sort of mapping to some enum values for fast evaluation.

Issue #3544


core/arch/riscv64/emit_utils.c
```diff
@@ -33,13 +33,8 @@
 #include "../globals.h"
 #include "arch.h"
 
-#ifdef RISCV_ISA_C
-#    define RAW_NOP_INST 0x0001
-#    define RAW_NOP_INST_SZ 2
-#else
-#    define RAW_NOP_INST 0x00000013
```
derekbruening, line 40: Do we need to keep this define (with a different name) and have DR use it when it dynamically identifies that C is not available?

derekbruening (approved):

core/arch/riscv64/emit_utils.c
semihalf-kardach-stanislaw: It would be safer to do so, yes. I can't find a proof in the spec that compressed instructions should be treated as NOPs on a non-C platform. I'll re-add the define and add an appropriate comment in `fill_with_nops()`.

For context, chapter "26.3 Extensions within fixed-width 32-bit instruction format" only mentions that the Compressed instruction space is available for custom extensions if C-extension is not implemented on a given platform.

Issue #4016, opened 2020-01-14
Implement automated binary re-relativization for AArch32 and AArch64 


derekbruening:
Xref related issues: #731, #3271, #3933 
This issue covers implementing automated binary re-relativization of AArch32 and AArch64 instructions.  We will need a different strategy than the x86 approach of storing the location of a 32-bit immediate, due to the varying immediate encodings.  For now, decode_from_copy() invalidates the raw bits to force a re-encoding as a workaround for not having this feature.

johnfxgalea:
Just checking, but this issue is still only a concern when DR is used in standalone mode right? (in relation to #731) Or are there mangling issues which were recently discovered please?

derekbruening:
This does not impact core DR's basic operations, no.  It impacts someone using decode_from_copy() or other decoding operations, which could be standalone or in a client.  The workaround of forcing re-encode should ensure there are no problems today *except* when we have parts of an instr that are not recorded and not recreated on that re-encode: such as this case which we hit recently: #4017.

derekbruening:
This is causing loss of encoding fidelity for cases like #7132.

PR #5666, opened 2022-09-30
i#3544 riscv64, part 1: add instruction codec

semihalf-kardach-stanislaw:
Add the main source file for instruction codec whose goal is to decode and encode all RISC-V instructions. At the moment only the decoding part is implemented.

The decoding process is based on instr_info_t structures and is separated into two stages:

1. Instruction opcode decoding to retrieve a corresponding rv_instr_info_t. This gives the codec information on the number and type of source and destination operands.
2. Instruction operand decoding. All operands are decoded utilizing the data from instr_info_t and a set of operand decoder functions gathered in the opnd_decoders array indexed by the operand field type.

Note:
   The codec.c file is not yet added to the build system as it relies on a set of generated files with opcode and instr_info_t data. The script for that as well as Instruction Set Listing files will follow in later commits.

Note 2:
   There is currently no support for decoding of translated instructions and hence orig_pc is never used. In fact all base+disp operands utilize opnd_create_mem_instr() which is described as a PC-relative memory reference, so it "should" be resistant to the change in base address.

Issue #3544


semihalf-kardach-stanislaw:
Note to self. The error reported in ci-clang-format run is:
```
debug-internal-64: **** 3 build errors ****
	/.../dynamorio/core/ir/riscv64/codec.c:68: error space_after_call: Extra space after function name: "architectures ".
```
Though at the beginning of the log (line 16) I see `clang-format check passed`. Also the line in question is:
```c
#    error RISC-V codec only supports 64-bit architectures (mask+match -> code).
```
So that is definitely not a "function name". I'll change the line though.

derekbruening:
> 



> Note to self. The error reported in ci-clang-format run is:
> 
> ```
> debug-internal-64: **** 3 build errors ****
> 	/.../dynamorio/core/ir/riscv64/codec.c:68: error space_after_call: Extra space after function name: "architectures ".
> ```
> 
> Though at the beginning of the log (line 16) I see `clang-format check passed`. Also the line in question is:
> 
> ```c
> #    error RISC-V codec only supports 64-bit architectures (mask+match -> code).
> ```
> 
> So that is definitely not a "function name". I'll change the line though.

FYI, this is from vera++ which apparently does not recognize preprocessor lines?  If you wanted to try to tweak the checker instead -- easier to make the change you already made though.

core/ir/riscv64/codec.h
```diff
@@ -0,0 +1,207 @@
...
+
+/**
+ * List of instruction formats handled by the codec.
+ *
+ * Note that enum names have to match ones defined in the codec.py.
+ */
+typedef enum {
+    /* Uncompressed instruction formats */
+    RISCV64_FMT_R = 0,
+    RISCV64_FMT_R4,
```
abhinav92003, line 93: I needed to refer `riscv64_inst_fmt_t` while reviewing #5668. Quick comment: are these defined by the RISC-V ISA? Can you link to the relevant ISA documentation that describes what they mean?

core/ir/riscv64/codec.c
```diff
@@ -0,0 +1,1066 @@
...
+    [RISCV64_FLD_CJ_IMM] = decode_cj_imm_opnd,
+    [RISCV64_FLD_V_L_RS1_DISP] = decode_v_l_rs1_disp_opnd,
+    [RISCV64_FLD_V_S_RS1_DISP] = decode_v_s_rs1_disp_opnd,
+};
+
+static inline rv_instr_info_t *
+match_op_0(int funct, bool rv32, bool rv64)
+{
+    switch (funct) {
+    case 0: return &instr_infos[OP_c_addi4spn];
```
derekbruening, line 776: I'm not sure where these case values are coming from?

core/ir/riscv64/codec.c
```diff
@@ -0,0 +1,1066 @@
...
+    case 1: return &instr_infos[OP_c_fld];
+    case 2: return &instr_infos[OP_c_lw];
+    case 3:
+        if (rv32)
+            return &instr_infos[OP_c_flw];
+        if (rv64)
+            return &instr_infos[OP_c_ld];
+        return NULL;
+    // 4 is reserived.
```
derekbruening, line 785: spelling: reserved

core/ir/riscv64/codec.h
derekbruening: Sec 2.2, 2.3, etc. here seem to list some templates with letter names: https://riscv.org/wp-content/uploads/2017/05/riscv-spec-v2.2.pdf
I assume they match these?

core/ir/riscv64/codec.h
```diff
@@ -0,0 +1,207 @@
...
+#ifndef CODEC_H
+#define CODEC_H 1
+
+#include "decode.h"
+
+/**
+ * @file codec.h
+ * @brief Interface to the RISC-V instruction codec (encoder/decoder).
+ *
+ * Note that encoder does not verify validity of operand values. This is because
```
derekbruening, line 42: nit: "the encoder"

core/ir/riscv64/codec.h
```diff
@@ -0,0 +1,207 @@
...
+ * DAMAGE.
+ */
+
+#ifndef CODEC_H
+#define CODEC_H 1
+
+#include "decode.h"
+
+/**
+ * @file codec.h
```
derekbruening, line 39: See other Doxygen comment.

core/ir/riscv64/codec.c
```diff
@@ -0,0 +1,1066 @@
...
+{
+    if (opc >= ARRAY_SIZE(instr_infos))
+        return NULL;
+    return &instr_infos[opc].nfo;
+}
+
+byte *
+decode_common(dcontext_t *dcontext, byte *pc, byte *orig_pc, instr_t *instr)
+{
+    /* Decode instruction width from the opcode */
```
derekbruening, line 976: style nit: Prefer end punctuation in comments.

core/ir/riscv64/codec.c
```diff
@@ -0,0 +1,1066 @@
...
+    int width = instruction_width(*(uint16_t *)pc);
+    /* Start assuming a compressed instructions. Code memory should be 2b aligned. */
```
derekbruening, line 978: grammar: instruction

core/ir/riscv64/codec.c
```diff
@@ -0,0 +1,1066 @@
...
+
+static rv_instr_info_t *
+get_rvc_instr_info(uint32_t inst, int xlen)
+{
+    /* 0 is an illegal instruction which is often used as a canary */
+    if (inst == 0)
+        return &instr_infos[OP_unimp];
+
+    int op = inst & 0b11;
+    int funct = (inst >> 13) & 0b111;
```
derekbruening, line 906: Could we use GET_FIELD for all of these?

core/ir/riscv64/codec.h
semihalf-kardach-stanislaw: Exactly. Though I could add ASCII-art comments describing the layout to limit the need to jump to the specification document.

core/ir/riscv64/codec.c
```diff
@@ -0,0 +1,1066 @@
...
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+#include <stdint.h>
+
+#include "../globals.h"
+#include "codec.h"
+
+/**
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: I will change to non-doxygen comment prefix. My intention was to document the internal code for future devs.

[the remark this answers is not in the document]
semihalf-kardach-stanislaw: On the other hand I wonder if I should not use doxygen syntax in internal code comments? Even if it's not used for html docs generation, some IDEs may do syntax highlighting on such comments, making them easier to read.
Plus if at any point the internal docs need to be exported (unlikely), we get doxygen for free.

core/ir/riscv64/codec.c
```diff
@@ -0,0 +1,1066 @@
...
+     *              operands, there is always a destination register. Therefore:
+     *   - Destination type (riscv64_fld_t) -> dst1_type
+     *   - 1st source operand (riscv64_fld_t) -> src1_type
+     *   - 2nd source operand (riscv64_fld_t) -> src2_type
+     *   - 3rd source operand (riscv64_fld_t) -> src3_type
+     *   - 4th source operand (riscv64_fld_t) -> dst2_type
+     * - Match - fixed bits of the instruction -> code[63:32]
+     * - Mask - fixed bits mask for encoding validation -> code[31:0]
+     */
+    instr_info_t nfo;
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: The intent was to use as small name as possible since it's a wrapper structure. I can change to `info` for better readability.

core/ir/riscv64/codec.c
```diff
@@ -0,0 +1,1066 @@
...
+ *         the error.
+ */
+typedef bool (*opnd_dec_func_t)(dcontext_t *dc, uint32_t inst, int op_sz, byte *pc,
+                                byte *orig_pc, int idx, instr_t *out);
+
+/**********************************************************
+ * Helper functions.
+ */
+
+#define ARRAY_SIZE(x) sizeof(x) / sizeof(x[0])
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: I have missed the `BUFFER_SIZE_ELEMENTS` macro. I'll use it instead.

core/ir/riscv64/codec.c
semihalf-kardach-stanislaw: They were crafted using tables 16.5–16.7 in the spec. The compressed instruction set is rather small and there are many which alias other instructions so it was simpler to decode compressed instructions explicitly than include them in the search trie created for uncompressed instructions.

core/ir/riscv64/codec.c
semihalf-kardach-stanislaw: Yes and I should have done that.

core/ir/riscv64/codec.c
```diff
@@ -0,0 +1,1066 @@
...
+decode_b_imm_opnd(dcontext_t *dc, uint32_t inst, int op_sz, byte *pc, byte *orig_pc,
+                  int idx, instr_t *out)
+{
+    int32_t imm = BIT(inst, 31) << 12;
+    imm |= BIT(inst, 7) << 11;
+    imm |= GET_FIELD(inst, 30, 25) << 5;
+    imm |= GET_FIELD(inst, 11, 8) << 1;
+    imm = SIGN_EXTEND(imm, 13);
+    /* FIXME i#3544: Should PC-relative jumps targets be encoded as mem_instr or
+     * rather rel_addr?
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: > You mean an indirect jump whose target is stored in memory at a PC-relative address? Does RISC-V have such instructions? I thought its indirect jumps stored their addresses in registers only?

No, I meant instructions such as "JAL" (jump and link) or "BEQZ" (branch if equal zero). Both compute the jump target by adding the immediate encoded in the instruction to the current value of the PC register. So I thought that I should decode the "immediate" operand as either `opnd_create_rel_addr()` or `opnd_create_mem_instr()`.

Though looking at ARM (<=v7) I should just use `opnd_create_base_disp()` with `DR_REG_PC` as base?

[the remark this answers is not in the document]
derekbruening: I see.  Today all arches (x86, aarch64, aarch32) use `opnd_create_pc` to store a direct jump/branch's target as an absolute PC, resolved using the current PC at decode time.  It is implicitly a PC-relative type of operand and follows the general rule of having absolute resolved addresses in operands for easy examination.  When encoding an unchanged instruction there is supposed to be support to auto-magically re-relativize to still point at the same absolute address (instead of maintaining the relative offset from the new encoding position): though there are still some holes there (#4016) as it does not affect core operation.

semihalf-kardach-stanislaw:
run arm tests

semihalf-kardach-stanislaw:
Sorry for pressing the "merge branch" button by mistake. Unfortunate result of a trackpad miss-click.

core/ir/riscv64/codec.h
semihalf-kardach-stanislaw: Done

core/ir/riscv64/codec.h
```diff
@@ -0,0 +1,207 @@
...
+
+/**
+ * Calculate instruction width.
+ *
+ * Returns a negative number on an invalid instruction width.
+ */
+static inline int
+instruction_width(int16_t lower16b)
+{
+    if ((lower16b & 0b11) < 0b11)
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Done

core/ir/riscv64/codec.h
```diff
@@ -0,0 +1,207 @@
...
+        return 2;
+    else if ((lower16b & 0b11100) < 0b11100)
+        return 4;
+    else if ((lower16b & 0b011111) == 0b011111)
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Done

core/ir/riscv64/codec.c
```diff
@@ -0,0 +1,1066 @@
+/* **********************************************************
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Done

core/ir/riscv64/codec.c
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Done

core/ir/riscv64/codec.c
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Done

core/ir/riscv64/codec.c
```diff
@@ -0,0 +1,1066 @@
...
+#include <stdint.h>
+
+#include "../globals.h"
+#include "codec.h"
+
+/**
+ * RISC-V extended instruction information structure.
+ *
+ * Holds extra elements required for encoding/decoding. Since instr_info_t is
+ * 48 bytes large, there is 16 bytes available to a single cache-line (assuming
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Done

core/ir/riscv64/codec.c
```diff
@@ -0,0 +1,1066 @@
...
+#define SIGN_EXTEND(val, val_sz) (((int32_t)(val) << (32 - (val_sz))) >> (32 - (val_sz)))
+
+#define INFO_NDST(opcode) GET_FIELD((opcode), 31, 31);
+#define INFO_NSRC(opcode) GET_FIELD((opcode), 30, 28);
+
+/*
+ * End of helper functions.
+ **********************************************************/
+
+/* Include a generated list of rv_instr_info_t and an acompanying trie structure
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Done

core/ir/riscv64/codec.c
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Done.

core/ir/riscv64/codec.c
semihalf-kardach-stanislaw: Done. I've renamed those functions to better match the spec tables.

core/ir/riscv64/codec.c
semihalf-kardach-stanislaw: Done

core/ir/riscv64/codec.h
semihalf-kardach-stanislaw: Done

core/ir/riscv64/codec.h
semihalf-kardach-stanislaw: Done

core/ir/riscv64/codec.c
semihalf-kardach-stanislaw: Done

core/ir/riscv64/codec.c
semihalf-kardach-stanislaw: Done.

core/ir/riscv64/codec.c
semihalf-kardach-stanislaw: Done

core/ir/riscv64/codec.c
```diff
@@ -0,0 +1,1066 @@
...
+static bool
+decode_none_opnd(dcontext_t *dc, uint32_t inst, int op_sz, byte *pc, byte *orig_pc,
+                 int idx, instr_t *out)
+{
+    ASSERT_NOT_REACHED();
+    return false;
+}
+
+static bool
+decode_rd_opnd(dcontext_t *dc, uint32_t inst, int op_sz, byte *pc, byte *orig_pc, int idx,
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Done. Added layouts for all decode functions.

core/ir/riscv64/codec.c
```diff
@@ -0,0 +1,1413 @@
...
+    case 1: info = match_op_1(funct, funct2, funct3, bit11to7, bit12, rv32, rv64); break;
+    case 2: info = match_op_2(funct, bit11to7, bit6to2, bit12, rv32, rv64); break;
+    }
+
+    if (info == NULL)
+        return NULL;
+    mask = GET_FIELD(info->info.code, 31, 0);
+    match = GET_FIELD(info->info.code, 63, 32);
+
+    if ((inst & mask) != match)
```
derekbruening, line 1275: Maybe it's in one of the other parts, but are there docs on what the `code` field holds?  What is this mask and match?  What do we need to check if we've made it this far and found a table entry?  Is this just a debug sanity check (should it be ifdef DEBUG?) or there are some bits we haven't yet examined and this is a required last step? I can't tell from this code snapshot.  Add a comment here too even if documented somewhere else?

core/ir/riscv64/codec.c
```diff
@@ -0,0 +1,1413 @@
...
+                                                 orig_pc, 0, instr))
+            goto decode_failure;
+    case 0: break;
+    default: ASSERT_NOT_REACHED();
+    }
+
+    if (orig_pc != pc) {
+        /* We do not want to copy when encoding and condone an invalid
+         * relative target.
+         * FIXME i#3544: Add re-relativization support without having to re-encode.
```
derekbruening, line 1395: This will cause re-encoding of every single instr...the FIXME is already there so no suggestion just talking out loud.

derekbruening (approved):

core/ir/riscv64/codec.c
```diff
@@ -0,0 +1,1413 @@
...
+{
+    ASSERT_NOT_REACHED();
+    return false;
+}
+
+/* Decode the destination fixed-point register field:
+ * |31 12|11   7|6      0|
+ * | ... |  rd  | opcode |
+ *        ^----^
+ * Applies to R, R4, I, U and J uncompressed formats.
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: No problem. :)

core/ir/riscv64/codec.c
```diff
@@ -0,0 +1,1413 @@
...
+{
+    int32_t imm = SIGN_EXTEND((BIT(inst, 12) << 5) | GET_FIELD(inst, 6, 2), 6);
+    opnd_t opnd = opnd_create_immed_int(imm, op_sz);
+    instr_set_src(out, idx, opnd);
+    return true;
+}
+
+/* Decode the immediate field of the CB-type format as a pc-relative offset:
+ * |15 13|12        10|9   7|6               2|1      0|
+ * | ... | imm[8|4:3] | ... | imm[7:6|2:1|5] | opcode |
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Apparently this order simplifies the hardware instruction decoder implementation. Whether prioritizing simple instruction decoder implementation is worth it is a matter of debate, at least on comp.arch.

core/ir/riscv64/codec.c
```diff
@@ -0,0 +1,1413 @@
...
+ *
+ * Note that this is a virtual field injected by codec.py into instructions
+ * which share the immediate field type with other non-base+disp instructions.
+ */
+static bool
+decode_v_s_rs1_disp_opnd(dcontext_t *dc, uint32_t inst, int op_sz, byte *pc,
+                         byte *orig_pc, int idx, instr_t *out)
+{
+    reg_t reg = DR_REG_X0 + GET_FIELD(inst, 19, 15);
+    int32_t imm = (GET_FIELD(inst, 31, 25) << 5) | GET_FIELD(inst, 11, 7);
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: That's my next task - enabling compilation of the whole test suite for RISC-V and starting filling it. I think decoder validation tests could be generated from instruction set listing files, including immediate bit decoding validation.

core/ir/riscv64/codec.c
semihalf-kardach-stanislaw: There is a comment on line 57 that states what information is put into which part of instr_info_t.
As for the check, I'll verify tomorrow but I vaguely remember there are some holes in the compressed function encoding that we might fall into if say we get an invalid instruction to decode.

core/ir/riscv64/codec.c
semihalf-kardach-stanislaw: Yeah, that will be painful but I don't feel confident enough to tackle this issue ar this point.

core/ir/riscv64/codec.c
```diff
@@ -0,0 +1,1413 @@
...
+static rv_instr_info_t *
+get_rv_instr_info(uint32_t inst, trie_node_t trie[])
+{
+    rv_instr_info_t *info;
+    uint32_t mask, match;
+    trie_node_t *node;
+    size_t index;
+
+    /* We know the first index into trie straight from the instruction. */
+    index = (inst & 0x7f) + 1;
```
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: Unfortunately this interface is not yet documented as I was planning to only use it locally as Aarch64 and ARM do their decoder generation differently (as far as I can tell with if-else ladders).
Another reason for not making it a separate library is to go away with abstraction costs.
Though agreed, I'll document both the data structure and the algorithm as well as remove magic numbers. If possible I'll put the trie in a separate header with inline or macro operations. As the actual values held by the trie are essentially indices to an array, we might be able to re-use it...provided there would be a script for generating the trie array (which at this point seems very use-case specific).

core/ir/riscv64/codec.c
semihalf-kardach-stanislaw: I've checked the logic again. This should be a sanity check. I'll change it to ASSERT_MESSAGE().

semihalf-kardach-stanislaw: Done.

core/ir/riscv64/codec.c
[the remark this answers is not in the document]
semihalf-kardach-stanislaw: I did separate the trie into a header in the riscv directory.

PR #5670, opened 2022-09-30
i#3544 riscv64, part 5: enable basic instruction decoding

semihalf-kardach-stanislaw:
Add codec.c to the build system and implement basic decode and disassemble functions to enable drdisasm to produce proper output:

$ objdump -d /usr/bin/ls 2>/dev/null | grep -v '\.\.\.' \
          | cut -s -f 2 | ./clients/bin64/drdisas | head
 00017397   auipc  $0x17 -> %t2
 41c30333   sub    %t1 %t3 -> %t1
 8e83be03   ld     -0x0718(%t2)[8byte] -> %t3
 fd430313   addi   %t1 $-0x2c -> %t1
 8e838293   addi   %t2 $-0x718 -> %t0
 00135313   srli   %t1 $0x1 -> %t1
 0082b283   ld     +0x08(%t0)[8byte] -> %t0
 000e0067   jalr   %t3 $0x0 -> %zero
 00017e17   auipc  $0x17 -> %t3
 8d8e3e03   ld     -0x0728(%t3)[8byte] -> %t3

Issue #3544


core/ir/riscv64/disassemble.c
```diff
@@ -53,15 +60,34 @@ void
 opnd_base_disp_scale_disassemble(char *buf, size_t bufsz, size_t *sofar INOUT,
                                  opnd_t opnd)
 {
-    /* FIXME i#3544: Not implemented */
-    ASSERT_NOT_IMPLEMENTED(false);
+    /* There is no scaled addressing on RISC-V */
```
abhinav92003, line 63: Can use ASSERT_NOT_REACHED.

core/ir/riscv64/disassemble.c
```diff
@@ -53,15 +60,34 @@ void
...
+                ? ""
+                : (TEST(DR_DISASM_ARM, DYNAMO_OPTION(disasm_mask)) ? "#" : "$"));
 }
 
 bool
 opnd_disassemble_arch(char *buf, size_t bufsz, size_t *sofar INOUT, opnd_t opnd)
 {
-    /* FIXME i#3544: Not implemented */
-    ASSERT_NOT_IMPLEMENTED(false);
+    switch (opnd.kind) {
```
abhinav92003, line 77: I thought internal_opnd_disassemble (https://github.com/DynamoRIO/dynamorio/blob/820fa4df87c1f77231c258056ed6a721367c1e6d/core/ir/disassemble_shared.c#L537) needs to call opnd_disassemble_arch only for arch-dependent opnds. But the IMMED_INTEGER being printed here can probably be handled by the existing logic there (except the hex print format probably). Do we really need this separate logic?

semihalf-kardach-stanislaw: `internal_opnd_disassemble()` handles negative signed immediate integers only for ARM:
https://github.com/DynamoRIO/dynamorio/blob/820fa4df87c1f77231c258056ed6a721367c1e6d/core/ir/disassemble_shared.c#L546-L552
So I had a choice of either add `RISCV64` to the ifdef in there or put the logic in arch-specific disassemble function.
I preferred the latter because it makes the arch-specific code easier to find. What do you think?

core/ir/riscv64/disassemble.c
semihalf-kardach-stanislaw: Done.

core/ir/riscv64/disassemble.c
abhinav92003: Fair enough. Please add a comment by that `#ifdef ARM` saying that we should we move that to arm/disassemble.c.

abhinav92003 (approved):

core/ir/riscv64/disassemble.c
semihalf-kardach-stanislaw: Done.

Issue #5520, opened 2022-06-07
Add instruction encoding entries to drmemtrace


derekbruening:
Today, the drmemtrace format does not include instruction encodings or even opcodes beyond identifying branch types.  This is sufficient for functional cache simulation, as in drcachesim, but for core simulation the operand dependencies and ideally the opcodes are needed.  The solution today is to preserve the application binaries for the executable and libraries, with support for mapping them in during trace analysis in order to decode the instructions (using the same support built for post-processing offline traces).  Adding the encodings into the trace proper would make the trace self-contained.  It is also the simplest solution for supporting dynamically-generated code that has no binary to query later: https://github.com/DynamoRIO/dynamorio/issues/2062#issuecomment-985707479

prasun3:
Noting two use cases which could be handled differently:
 - DGC code where there is no binary: in this case the opcodes need to be traced at runtime
 - Using traces collected on Windows on Linux in which case the Windows binaries cannot be parsed on Linux: in this case the opcodes may be inserted during raw2trace conversion

derekbruening:
Including the instruction encodings would increase trace files sizes by an estimated 75%.
One option to reduce that is to only include the encoding once for the first dynamic instance of a static instruction (applying only to  unchanging library code, or possibly to jitted code through analysis).  This could complicate skipping part of the trace as in the proposed feature #5538 but we already have precedent for once-only information in physical address markers (#4014) and could solve by re-emitting every so often.

derekbruening:
There are actually a couple of unfinished things:
+ Add a way for tools to know when to invalidate cached decodings
+ Add encodings to online traces, but under an option since the extra trace_entry_t records affect overhead there
+ Add encoding entry count to basic_counts

derekbruening:
Pretty much everything is done except the initial online encoding solution is imperfect:
```
    // TODO i#5520: Currently we emit the encoding again for every dynamic instance
    // of an instruction.  We should record which we've emitted and avoid duplicate
    // instances (the reader caches prior encodings).  For offline we do this
    // separately per thread, which makes knowing when code has changed complex as
    // any per-thread structures would need a global walk across them on a fragment
    // deletion event.  For online, however, we may be able to emit just once globally
    // if the reader is always interleaving all threads: but while that simplifies
    // invalidation/code changes it requires global locks to update the structure.
    // Since encodings are off by default we leave it as emitting every time
    // with corresponding extra overhead for now.
```

PR #5672, opened 2022-10-03
i#5520 memtrace encodings: Add AArch32 support

derekbruening:
Adds support for Thumb versus Arm modes in AArch32 with respect to drmemtrace stored encodings by setting the LSB when storing in the tracer and raw2trace, and by properly setting the base mode to Arm when decoding.

Augments decode_from_copy() to switch locally to Thumb if either the read or original PC has LSB=1 as that better fits most use cases (including reading from a drmemtrace encoding buffer).

Adds AArch32 support to the burst_gencode test.

Fixes 3 outstanding AArch32 issues:

+ Removes a quote in a comment in third_party/libgcc/arm/lib1funcs.S code not handled by old toolchains.

+ Removes an ASSERT_NOT_TESTED path now hit.

+ Fixes an AArch32 tracer bug where a 2nd temp register is needed for a 2nd conditional skip.

Tested by running the drcacheoff.gencode test on an AArch32 machine which now passes.

Issue: #5520


third_party/libgcc/arm/lib1funcs.S
```diff
@@ -1465,7 +1465,8 @@ LSYM(Lover12):
         asr     ah, r2
         sub     r2, #32
         @ If r2 is negative at this point the following step would OR
-        @ the sign bit into all of AL.  That's not what we want...
+        @ <bruening: removing quote on next line since one toolchain trips over it>
+        @ the sign bit into all of AL.  Thats not what we want...
```
abhinav92003, line 1469: s/Thats/That is/
or
s/Thats not what we want/We do not want that/

Is the comment in the above line intended to be kept?

abhinav92003 (approved):

third_party/libgcc/arm/lib1funcs.S
derekbruening: > Is the comment in the above line intended to be kept?

Yes, so future devs aren't confused by the diff vs upstream lib1funcs.S.  Looks like I should mention this or include the diff in ../README.dynamorio too: will do that.

derekbruening: Maybe just the mention in the README is enough and we don't need the embedded comment.

derekbruening:
32-bit failure is scattergather pipe assert #5329

PR #5673, opened 2022-10-04
i#5520 memtrace encodings: Add encoding count to basic_counts

derekbruening:
Although the encoding records are not directly exposed to analysis tools, the encoding_is_new field can be used as a proxy to count them. We do that here for the basic_counts tool.

Issue: #5520


bete0 (approved):

derekbruening:
Failure on the update is a64 windows-zlib crash #5493 (a bug somewhere in our zlib reader)

PR #5674, opened 2022-10-04
i#5520 memtrace encodings: Ensure raw dir is not needed

derekbruening:
Fixes the incorrect attempt by analyzer_interface to synthesize a modules.log directory when running opcode_mix, which failes when there is no raw directory.

Modifies the opcode_mix offline test to remove the raw directory as a sanity check that the embedded instruction encodings are indeed all we need.

Issue: #5520


derekbruening:
32-bit test failure is api.startstop with usage error about vdso: #4604

suite/tests/CMakeLists.txt
```diff
@@ -3749,6 +3751,13 @@ if (BUILD_CLIENTS)
 
     torunonly_drcacheoff(opcode_mix ${ci_shared_app} ""
       "@-simulator_type@opcode_mix" "")
+    # Ensure the tool works without the raw/ subdir.
+    set(${testname_full}_postcmd
```
abhinav92003, line 3755: Does this really work? testname_full is defined inside torunonly_drcacheoff. Are we relying on that? Seems a bit hacky.

abhinav92003 (approved):

suite/tests/CMakeLists.txt
derekbruening: Oh, this is a mistake: better to be the precise name `tool.drcacheoff.opcode_mix_postcmd`.  Fixed, here and the 2 others below.

derekbruening: But yes it happened to work b/c the macro had set that var (macros don't have a separate namespace scope, unlike functions, so the var is visible).

Issue #5635, opened 2022-08-31
CRASH in tool.drcacheoff.raw-zlib test: bug in zlib_istream.h


derekbruening:
The tool.drcacheoff.raw-zlib test crashed on Jenkins but I don't think this
is limited to AArch64:
```
70/84 Test #220: code_api|tool.drcacheoff.raw-zlib ...............................***Failed    0.47 sec
Running cmd |/var/lib/jenkins/workspace/DynamoRIO-AArch64-Precommit/build/build_release-external-64/bin64/drrun;-s;90;-quiet;-killpg;-stderr_mask;0xC;-dumpcore_mask;0;-code_api;-t;drcachesim;-offline;-subdir_prefix;tool.drcacheoff.raw-zlib;-raw_compress;zlib;--;/var/lib/jenkins/workspace/DynamoRIO-AArch64-Precommit/build/build_release-external-64/suite/tests/bin/simple_app|
Running postcmd |/var/lib/jenkins/workspace/DynamoRIO-AArch64-Precommit/build/build_release-external-64/clients/bin64/drcachesim;-indir;/var/lib/jenkins/workspace/DynamoRIO-AArch64-Precommit/build/build_release-external-64/suite/tests/tool.drcacheoff.raw-zlib.simple_app.3622551.1757.dir|
CMake Error at /var/lib/jenkins/workspace/DynamoRIO-AArch64-Precommit/suite/tests/runmulti.cmake:108 (message):
  *** postcmd failed (Segmentation fault): ***
```

On the same machine I managed to reproduce but only on the 2nd try.
Once I had a raw file that reproduces, it is a deterministic crash in raw2trace.

```
 523692 [drmemtrace]: Thread 3631869 timestamp 0x002f461fc314fe80
<...>
 561733 [drmemtrace]: Appending 5 instrs in bb 0xfffd8833c5f0 in mod 7 +0x95f0 = /usr/lib/aarch64-linux-gnu/ld-2.31.so
 561734 [drmemtrace]: Caching last block summary 0xfffd8006d460 for 0xfffd8833c5f0
 561735   0x0000ffff99a9c5f0 ldr    (%x23)[4byte] -> %w1
 561736 [drmemtrace]: Using last block summary 0xfffd8006d460 for 0xfffd8833c5f0
 561737 [drmemtrace]: Appending delayed branch pc=0xffff99a9c5ec for thread 0
[...]

Issue #2040, opened 2016-10-20
static client issues on Windows


derekbruening:
Split from #975

The burst-threads test for #1729 hits this assert on takeover:

```
pre-DR start
<Application D:\derek\dr\build_suite\build_debug-internal-32\clients\bin32\tool.drcacheoff.burst_threads.exe (1936).  Internal Error: DynamoRIO debug check failure: D:\derek\dr\git\src\core\dispatch.c:939 dcontext->app_fls_data == NULL || dcontext->app_fls_data != dcontext->priv_fls_data
```

This may be related to the private loader being disabled?  Or to #2034 mixing threads around?

derekbruening:
I added a setup;detach test to static_noclient and it hits this
app_fls_data assert but only under ctest:

```
% ctest -V -R static_nocli
UpdateCTestConfiguration  from :D:/derek/dr/git/build_x64_dbg_tests/DartConfiguration.tcl
Parse Config file:D:/derek/dr/git/build_x64_dbg_tests/DartConfiguration.tcl
UpdateCTestConfiguration  from :D:/derek/dr/git/build_x64_dbg_tests/DartConfiguration.tcl
Parse Config file:D:/derek/dr/git/build_x64_dbg_tests/DartConfiguration.tcl
Test project D:/derek/dr/git/build_x64_dbg_tests
Constructing a list of tests
Done constructing a list of tests
Checking test dependency graph...
Checking test dependency graph end
test 210
    Start 210: code_api|api.static_noclient

210: Test command: D:\derek\dr\git\build_x64_dbg_tests\bin64\drrun.exe "-s" "90" "-quiet" "-debug" "-use_dll" "D:/derek/dr/git/build_x64_dbg_tests/lib64/debug/dynamorio.dll" "-exit0" "-noinject" "-stderr_mask" "0xC" "-msgbox_mask" "0" "-dumpcore_mask" "0x7d" "-staged" "-code_api" "--" "D:/derek/dr/git/build_x64_dbg_tests/suite/tests/bin/api.static_noclient.exe"
210: Test timeout computed to be: 1500
210: pre-DR init
210: pre-DR start
[...]

PR #5679, opened 2022-10-07
i#2039 trace trim, part 2: Add open_file with tid

derekbruening:
Adds drmemtrace_replace_file_ex_ops() with an expanded file opening function which takes in the thread id and window id, to better support the external file opener with delayed opens due to nop mode and windows.

Adds a test of the function to burst_replaceall.  Adds additional threads to the test and checks that the tid for each thread was seen.

Issue: #2039

### DIFF
--- a/clients/drcachesim/tests/burst_replaceall.cpp
+++ b/clients/drcachesim/tests/burst_replaceall.cpp
@@ -49,14 +49,17 @@
 #include <stdlib.h>
 #include <string.h>
 #include <set>
-#ifndef WINDOWS // XXX i#2040: Static client limitations
+// XXX i#2040: Static client limitations on Windows prevent the thread
+// aspect of this test from working today.  Once that is fixed we can
+// re-enable by removing all of the "ifndef WINDOWS" with this comment:
+#ifndef WINDOWS // XXX i#2040: Disabled until Windows static limits are fixed.
 #    ifdef WINDOWS
 #        include <process.h>
 #    endif
 #endif
 
+#ifndef WINDOWS // XXX i#2040: Disabled until Windows static limits are fixed.
 static void *finished;
-#ifndef WINDOWS // XXX i#2040: Static client limitations
 static constexpr int num_threads = 3;
 static void *started[num_threads];
 #endif
@@ -228,7 +231,7 @@ exit_cb(void *arg)
         dr_close_file(f);
         dr_raw_mem_free(entry->data, entry->alloc_size);
     }
-#ifndef WINDOWS // XXX i#2040: Static client limitations
+#ifndef WINDOWS // XXX i#2040: Disabled until Windows static limits are fixed.
     // Ensure we were passed every app thread tid to our file open function.
     if (tids.size() != num_threads + 1 /*main*/) {
         std::cerr << "Saw " << tids.size() << " threads but expected "
@@ -239,7 +242,7 @@ exit_cb(void *arg)
     drvector_delete(&all_buffers);
 }
 
-#ifndef WINDOWS // XXX i#2040: Static client limitations
+#ifndef WINDOWS // XXX i#2040: Disabled until Windows static limits are fixed.
 #    ifdef WINDOWS
 unsigned int __stdcall
 #    else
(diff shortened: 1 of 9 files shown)


clients/drcachesim/tests/burst_replaceall.cpp
```diff
@@ -42,11 +42,24 @@
...
 #include "drvector.h"
 #include "../../../suite/tests/client_tools.h"
+#include "../../../suite/tests/condvar.h"
 #include <assert.h>
 #include <iostream>
 #include <math.h>
 #include <stdlib.h>
 #include <string.h>
+#include <set>
+#ifndef WINDOWS // XXX i#2040: Static client limitations
```
abhinav92003, line 52: This is confusing. There's an ifndef followed by an ifdef for the same symbol. Is that because of #2040? I don't understand how this works though.

clients/drcachesim/tracer/drmemtrace.h
```diff
@@ -149,6 +169,18 @@ drmemtrace_replace_file_ops(drmemtrace_open_file_func_t open_file_func,
...
                             drmemtrace_create_dir_func_t create_dir_func);
 
+DR_EXPORT
+/**
+ * Identical to drmemtrace_replace_file_ops() except its file open function takes
+ * two extra parameters.
+ */
+drmemtrace_status_t
+drmemtrace_replace_file_ops_ex(drmemtrace_open_file_ex_func_t open_file_ex_func,
+                               drmemtrace_read_file_func_t read_file_func,
```
abhinav92003, line 179: Perhaps we should change the args to this routine to a struct (file_ops_func_t itself)? (to avoid an _ex_ex in future).

clients/drcachesim/tests/burst_replaceall.cpp
```diff
@@ -42,11 +42,24 @@
...
+#    ifdef WINDOWS
+#        include <process.h>
+#    endif
+#endif
+
+static void *finished;
```
abhinav92003, line 58: `finished` can also be conditionally defined for !WINDOWS.

clients/drcachesim/tracer/tracer.h
```diff
@@ -192,7 +193,21 @@ struct file_ops_func_t {
...
+    file_t
+    call_open_file(const char *fname, uint mode_flags, thread_id_t thread_id,
+                   int64 window_id)
+    {
+        if (open_file_ex != nullptr)
+            return open_file_ex(fname, mode_flags, thread_id, window_id);
+        return open_file(fname, mode_flags);
+    }
+    file_t
+    open_file_special(const char *fname, uint mode_flags)
```
abhinav92003, line 205: Maybe change the _special suffix to something that's properly descriptive?

abhinav92003 (approved):

clients/drcachesim/tracer/tracer.h
derekbruening: Do you have a suggestion?  It is not clear how to describe them other than they are not regular trace data files.  These are modules.log, funclist.log, encodings.bin, and all PT files.

clients/drcachesim/tests/burst_replaceall.cpp
derekbruening: The idea was to be able to re-enable for Windows by just removing these marked ifndefs which is why the lines inside may contradict the outside.  What is your suggestion: remove these lines entirely (though if re-enabled someone will have to figure out that this include is missing from the build error)?

abhinav92003: A comment should be enough.

clients/drcachesim/tracer/tracer.h
abhinav92003: I'll downgrade my comment to a nit. We can optionally change it to open_non_thread_file, though even that has its drawbacks (each PT file maps to a single thread).

clients/drcachesim/tests/burst_replaceall.cpp
derekbruening: Done.

clients/drcachesim/tests/burst_replaceall.cpp
derekbruening: Done.

clients/drcachesim/tracer/tracer.h
derekbruening: Thinking about it more: PT should probably go through the regular routine though since we'd want to split it by thread and by window.  So this routine is more geared toward single process-wide files.  I'm renaming to open_process_file() and putting a comment by the PT use that it should be changed.

clients/drcachesim/tracer/drmemtrace.h
derekbruening: +1, done.  The header is extern C though so it has to be a new struct.

derekbruening:
arm hit the raw-zlib crash #5635

derekbruening:
run arm tests

Issue #5638, opened 2022-09-01
API for getting lane width of x86 SIMD operands


khuey:
I have an application that tracks the flow of data through instructions. It would be useful for this to know the lane width of SIMD operands. This data already exists in source form in DynamoRIO (e.g. Xvd vs Xvs entries for operand size in the decoder table) but gets compiled out (because Xvd and Xvs are #defined to the same thing). It's not obvious where to put this: adding variations of the operand size (e.g. OPSZ_16_vex32_ps and OPSZ_16_vex32_pd) would be kind of disruptive. Any thoughts?

derekbruening:
For data flow tools (taint tracking, etc.), the full semantics of each complex SIMD opcode is also needed.  Dr. Memory has to use dedicated code for each opcode to mirror its data movements: https://github.com/DynamoRIO/drmemory/blob/master/drmemory/slowpath_x86.c#L1110

For the x86 DR opcodes, does the lane width vary for the same opcode?
I remember for aarch32 we split the opcodes so you could tell the lane from the opcode.
For aarch64 I'm not sure of the status (@AssadHashmi may know more): SVE is in-progress; not as familiar with NEON aarch64 in DR, but I do recall a number of cases where tools would prefer DR's aarch64 decoder to split opcodes up further than it does.

If the lane width is fixed for each opcode, and you'd still need to dispatch on opcode for the data movement semantics, is the width helping much?  I'm not saying DR should't provide it -- just trying to get a bigger picture for the use cases.  Some kind of data movement by opcode library might be the ultimate solution?

khuey:
I'm not aware of any x86 opcodes that have multiple lane widths, although I wouldn't rule out Intel having snuck something in in the depths of AVX-512 without doing a thorough search. There are some weird things like the broadcast and permute instructions but most SIMD instructions don't allow any crossover between lanes. You'll see there's no code for e.g. vaddps in that drmemory code you linked.

That said simply encoding the lane widths along with the opcodes in my application is something I've considered. It's certainly not hard to do.

khuey:
One thing that is annoying about doing this separately is that there are opcodes where the lane widths are different on the different operands (e.g. cvtps2pd)

derekbruening:
[...]

PR #5681, opened 2022-10-11
i#3044: Add a type to represent aarch64 vectors

joshua-warburton:
Previously in the IR we have represented vectors by a plain H, S, D or Q register combined with a faux-imm added after the last vector to give a hint to its size.

This patch uses the .size field of the opnd struct to include the element size, similar to how partial registers are used for x86. Element Vector registers are differentiated from this by setting the DR_OPND_IS_VECTOR (0x40) bit on the .flag field. This bit is also set by the DR_OPND_IS_EXTEND flag for imms and was chosen as to not extend the size of the flag field while remaining unambiguous.

The new Element Vector operand is printed like z0.b via a check in disassemble_shared.c to set a suffix. Also added are the utility functions opnd_is_element_vector_reg,
opnd_create_reg_element_vector and opnd_get_vector_element_size/

issue: #3044


AssadHashmi:
Hi @derekbruening, I'm happy with this patch to improve the vector<->element binding for AArch64 SVE instructions. Let us know if you have any objections. We would like this to be the format from SVE onwards. Retro-fitting to v8.0/1/2 SIMD (NEON) may happen later.

derekbruening:
Hmm, I don't think we've discussed element sizes before: the #3044 discussions were about the mcontext fields.

IIRC, previously the element size was not in the IR at all (for any arch).  The size field could be smaller than the register capacity but that meant a sub-reg operand where other bits were left unchanged.  Tools that needed the element size (e.g., Dr. Memory) dispatched on opcode.

For this proposal in this PR, my question is: what about an operand that needs both an element size and a sub-register size?  IIRC some of the x86 vector operations would leave some bits unchanged and I think the x86 decoder marks those as sub-register (I assume these are not just leaving exactly half unchanged as there we could use the smaller vector register name?  or maybe we go w/ the name in the ISA convention b/c other operands are on that full size even when it's half...might have to refresh my memory).  Maybe this doesn't happen for SVE?  But if we wanted to adopt this for other ISA's it could be an issue there.  Not sure what the solution is though other than finding other bits to store the element size.

Maybe taking a step back and seeing whether the sub-reg size is always useful is worthwhile: is it always the bottom bits that are used and the top that are unused?  If not then tools have to dispatch on opcode anyway.  Are there identical opcodes that differ only by sub-reg size and then a dispatch is not possible.

derekbruening:
Looking for more opinions: @abhinav92003

joshua-warburton:
For context, I do not believe partial registers are used anywhere in aarch64.

joshua-warburton:
There doesn't appear to be anything that the element size could be union'd with to fulfil those requirements.

However, there is likely some padding space after the first union, kind(8) + size(8) + union of shorts(16) leaves another 32 bits before the 64-bit aligned second union. Putting an element size in there wouldn't increase the overall size of the struct.

derekbruening:
For `REG_kind` for 64-bit pointers, `opnd_t.value` is only using 16 of its 32 bits (for `reg_id_t`), so we could add a size for `REG_kind` there?

joshua-warburton:
Implemented the above in a new patch.

joshua-warburton:
run arm tests

derekbruening:
Turns out there is an issue filed on adding element widths to x86 SIMD operands: https://github.com/DynamoRIO/dynamorio/issues/5638

joshua-warburton:
I believe this is now ready for review, with the suggested change to the storage location @derekbruening @AssadHashmi

core/ir/instr_inline_api.h
```diff
@@ -141,6 +141,13 @@ opnd_is_far_base_disp(opnd_t op)
     return IF_X86_ELSE(op.kind == BASE_DISP_kind && op.aux.segment != DR_REG_NULL, false);
 }
 
+INSTR_INLINE
+bool
+opnd_is_element_vector_reg(opnd_t op)
+{
+    return op.kind == REG_kind && ((op.aux.flags & DR_OPND_IS_VECTOR) != 0);
```
derekbruening, line 148: style: Use TEST

core/ir/opnd_api.h
```diff
@@ -1695,6 +1695,11 @@ typedef enum _dr_opnd_flags_t {
     DR_OPND_EXTENDED = 0x20,
     /** This immediate integer operand should be interpreted as an AArch64 extend type. */
     DR_OPND_IS_EXTEND = 0x40,
+    /** Registers with this flag should be considered vectors and have a subsize
```
derekbruening, line 1698: style: https://dynamorio.org/page_code_style.html#autotoc_md124 says "For comments starting with /∗∗, leave the rest of the first line empty, unless the entire comment is a single line."

core/ir/opnd_api.h
```diff
@@ -1863,6 +1871,15 @@ INSTR_INLINE
 opnd_t
 opnd_create_reg_partial(reg_id_t r, opnd_size_t subsize);
 
+DR_API
+INSTR_INLINE
+/**
+ * Returns a register operand corresponding to a vector
+ * register that uses .size to represent the element size.
```
derekbruening, line 1878: s/.size/\p element_size/

core/ir/opnd_api.h
```diff
@@ -2480,6 +2503,14 @@ DR_API
 void
 opnd_set_size(opnd_t *opnd, opnd_size_t newsize);
 
+DR_API
+/**
+ * Return the element size of \p opnd as a OPSZ_ constant.
+ * Returns OPSZ_NA if \p opnd does not have a valid size.
```
derekbruening, line 2509: Add # to link to the doc: `#OPSZ_NA`

core/ir/opnd_shared.c
```diff
@@ -314,6 +314,15 @@ opnd_set_size(opnd_t *opnd, opnd_size_t newsize)
     }
 }
 
+opnd_size_t
+opnd_get_vector_element_size(opnd_t opnd)
+{
+    if (opnd.kind != REG_kind || ((opnd.aux.flags & DR_OPND_IS_VECTOR) == 0))
```
derekbruening, line 320: style: Use !TEST

core/ir/aarch64/instr_create_api.h
```diff
@@ -1095,6 +1095,19 @@
 #define INSTR_CREATE_add_vector(dc, Rd, Rm, Rn, width) \
     instr_create_1dst_3src(dc, OP_add, Rd, Rm, Rn, width)
 
+/**
+ * Creates a ADD vector instruction.
```
derekbruening, line 1099: s/a/an/

core/ir/opnd_api.h
```diff
@@ -1695,6 +1695,11 @@ typedef enum _dr_opnd_flags_t {
     DR_OPND_EXTENDED = 0x20,
     /** This immediate integer operand should be interpreted as an AArch64 extend type. */
     DR_OPND_IS_EXTEND = 0x40,
+    /** Registers with this flag should be considered vectors and have a subsize
+     *  representing their element size. Doubled up with _IS_EXTEND as that can
+     *  only apply to immd values and saves increasing the size of the containing
+     *  structure. */
+    DR_OPND_IS_VECTOR = 0x40,
```
derekbruening, line 1702: This is the same value as DR_OPND_IS_EXTEND.  Is that deliberate b/c the assumption is one is for ints and one for regs?  If so please make that explicit with a (non-doxygen) comment: otherwise it looks like a bug.  But if we have plenty of space I think it would be better to just use a non-overlapping value (don't recall how many bits we have for these flags).

core/ir/opnd_api.h
```diff
@@ -1770,7 +1775,10 @@ struct _opnd_t {
          * above, to save space.
          */
         instr_t *instr; /* INSTR_kind, FAR_INSTR_kind, and MEM_INSTR_kind */
-        reg_id_t reg;   /* REG_kind */
+        struct {
+            reg_id_t reg;
+            opnd_size_t element_size;
+        } reg_and_element_size; /* REG_kind */
```
derekbruening, line 1781: We do expose this struct for DR_FAST_IR and while clients should be using the inlined accessor routines for source compatibility and the "reg" field should be at the same offset for binary compatibility it still seems worth mentioning in the release notes (as a non-compatibility-affecting change?  or a minor source compatibility change if someone were directly accessing the "reg" field??)

derekbruening (approved):

core/ir/opnd_api.h
```diff
@@ -1770,7 +1775,10 @@ struct _opnd_t {
          * above, to save space.
          */
         instr_t *instr; /* INSTR_kind, FAR_INSTR_kind, and MEM_INSTR_kind */
-        reg_id_t reg;   /* REG_kind */
+        struct {
+            reg_id_t reg;
+            opnd_size_t element_size;
```
derekbruening, line 1780: Add a non-doxygen comment `XXX #5638: Fill in the element size for x86 and aarch32.`?

AssadHashmi:
While reviewing the disassembly changes I noticed an error in `reg_names[]` for SVE register strings. There's a `q3` which should be `z3`:
https://github.com/DynamoRIO/dynamorio/blob/1e82e56202c04484b953f2ff98bac344421e7bfd/core/ir/aarch64/encode.c#L112

Please put the fix in this patch, thanks.

core/ir/instr_inline_api.h
joshua-warburton: TEST isn't actually available in this file and adding the header file it appears in causes quite a lot of problems. 

I'm surprised the TEST at the top of the file doesn't cause issues too

core/ir/opnd_api.h
joshua-warburton: done

core/ir/opnd_api.h
joshua-warburton: done

core/ir/opnd_api.h
joshua-warburton: done

core/ir/aarch64/instr_create_api.h
joshua-warburton: Done

core/ir/opnd_api.h
joshua-warburton: Done

core/ir/opnd_api.h
joshua-warburton: Done

core/ir/disassemble_shared.c
```diff
@@ -630,7 +643,8 @@ internal_opnd_disassemble(char *buf, size_t bufsz, size_t *sofar INOUT,
         break;
     case REG_kind:
         reg_disassemble(buf, bufsz, sofar, opnd_get_reg(opnd), opnd_get_flags(opnd), "",
-                        "");
+                        opnd_is_element_vector_reg(opnd) ? opnd_size_element_suffix(opnd)
```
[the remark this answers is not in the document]
joshua-warburton: done

joshua-warburton:
I believe this is all the resolvable review comments resolved.

AssadHashmi (approved):

derekbruening:
> Also added are the utility functions opnd_is_element_vector_reg,
opnd_create_reg_element_vector and opnd_get_vector_element_size/

Looks like these new functions were not added to the release notes for new features.

PR #5682, opened 2022-10-11
i#5520 memtrace encodings: Add online encodings

### LINKED ISSUE #5520: Add instruction encoding entries to drmemtrace

derekbruening:
Adds a new option -instr_encodings which enables instruction encoding records for online traces.  This is under this off-by-default option as it adds significant overhead to online tracing and is only needed for some tools.

Updates the launcher and docs.

Adds an online opcode_mix test and adds encodings to a basic_counts online test.

Manually tested the view tool with and without -instr_encodings:

  $ bin64/drrun -t drcachesim -instr_encodings -simulator_type view -- suite/tests/bin/allasm_x86_64
      ...
      166: T1718245 ifetch       4 byte(s) @ 0x0000000000401028 48 83 eb 01          sub    $0x0000000000000001 %rbx -> %rbx
      167: T1718245 ifetch       4 byte(s) @ 0x000000000040102c 48 83 fb 00          cmp    %rbx $0x0000000000000000
      168: T1718245 ifetch       2 byte(s) @ 0x0000000000401030 75 d9                jnz    $0x000000000040100b
      ...

  $ bin64/drrun -t drcachesim -simulator_type view -- suite/tests/bin/allasm_x86_64
      ...
      126: T1933188 ifetch       4 byte(s) @ 0x0000000000401028 non-branch
      127: T1933188 ifetch       4 byte(s) @ 0x000000000040102c non-branch
      128: T1933188 ifetch       2 byte(s) @ 0x0000000000401030 conditional jump
      ...

Fixes #5520

### DIFF
--- a/clients/drcachesim/tests/allasm-repstr-basic-counts.templatex
+++ b/clients/drcachesim/tests/allasm-repstr-basic-counts.templatex
@@ -29,7 +29,7 @@ Total counts:
            0 total physical address \+ virtual address marker pairs
            0 total physical address unavailable markers
            4 total other markers
-           0 total encodings
+         102 total encodings
 Thread [0-9]* counts:
           98 \(fetched\) instructions
           26 unique \(fetched\) instructions
@@ -48,4 +48,4 @@ Thread [0-9]* counts:
            0 physical address \+ virtual address marker pairs
            0 physical address unavailable markers
            4 other markers
-           0 encodings
+         102 encodings
--- a/clients/drcachesim/tracer/instru_online.cpp
+++ b/clients/drcachesim/tracer/instru_online.cpp
@@ -33,6 +33,7 @@
 /* instru_online: inserts instrumentation for online traces.
  */
 
+#define NOMINMAX // Avoid windows.h messing up std::min.
 #include "dr_api.h"
 #include "drreg.h"
 #include "drutil.h"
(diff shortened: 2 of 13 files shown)


clients/drcachesim/tracer/output.cpp
```diff
@@ -659,7 +665,9 @@ create_v2p_buffer(per_thread_t *data)
 static bool
 is_ok_to_split_before(trace_type_t type)
 {
-    return type_is_instr(type) || type == TRACE_TYPE_INSTR_MAYBE_FETCH ||
+    return (op_instr_encodings.get_value()
+                ? type == TRACE_TYPE_ENCODING
```
abhinav92003, line 669: May be add a comment that this is based on which is the first entry output for each instrumentation sequence.

clients/drcachesim/tests/allasm-repstr-basic-counts.templatex
```diff
@@ -29,7 +29,7 @@ Total counts:
            0 total physical address \+ virtual address marker pairs
            0 total physical address unavailable markers
            4 total other markers
-           0 total encodings
+         102 total encodings
```
abhinav92003, line 32: 102 encodings for 26 unique instrs? Where are the "extra" ones coming from?

abhinav92003 (approved):

clients/drcachesim/tests/allasm-repstr-basic-counts.templatex
derekbruening: Currently it is emitting encodings for every dynamic instance.  In fact it looks like I did not even put a TODO about omitting for repeats in each thread.  Actually I didn't even think about *how* to do that: hmm.  I don't think it's trivial.  It would have to involve per-thread tables and fragment_delete events -- but if a thread-shared fragment is deleted how do we remove from every thread's table?  I think at this point since online analyzers looking at encodings is not a priority, the best course of action is to put some TODOs in here and maybe leave the issue open and put in this PR as is w/ the extra overhead.  Agreed?

derekbruening: A possible alternative would be to put in a table in each thread and assume no code deletion/changes with TODOs about that.  That would take a little work but it is straightforward; it would reduce overhead at the cost of accuracy.  Online is already slow-ish though and this is an off-by-default option so I would still vote for not doing anything now.

abhinav92003: > Currently it is emitting encodings for every dynamic instance. 

Oh right. But even then it's 98 vs 102; there seem to be 4 extra ones.

> omitting for repeats in each thread.

If we're outputting for each dynamic instance, then it's fine to repeat for each thread. I vote to keep online simple for now, with an XXX comment noting the possible optimizations.

derekbruening: > But even then it's 98 vs 102; there seem to be 4 extra ones.

This is counting total records so my guess is that 4 of them are long and need 2 records each.  I will check to confirm.

clients/drcachesim/tracer/instru_online.cpp
```diff
@@ -373,6 +374,45 @@ online_instru_t::instrument_instr(void *drcontext, void *tag, void *bb_field,
...
+
+    size_t len_left = len;
+    size_t buf_offs = 0;
+    do {
+        size_t len_cur = std::min(len_left, sizeof(((trace_entry_t *)0)->encoding));
+        insert_save_type_and_size(drcontext, ilist, where, reg_ptr, reg_tmp,
+                                  TRACE_TYPE_ENCODING, static_cast<ushort>(len_cur),
+                                  adjust);
+        app_pc immed = *(app_pc *)(buf + buf_offs);
+        insert_save_pc(drcontext, ilist, where, reg_ptr, reg_tmp, immed, adjust);
```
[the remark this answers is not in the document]
derekbruening: Renamed to insert_save_immed taking in a pointer-sized immediate.

clients/drcachesim/tracer/output.cpp
derekbruening: Done.

clients/drcachesim/tests/allasm-repstr-basic-counts.templatex
derekbruening: OK the other 4 are the non-fetches instrs: 98 fetched and 4 non-fetched (the rep loops).

I added a big comment about the repeated encodings and updated the issue.

PR #5684, opened 2022-10-12
i#5683: Detect bb truncation in drstatecmp

### LINKED ISSUE #5683: drstatecmp incorrectly uses pre-app2app ilist when block was truncated in app2app
When I remove -exit_after_tracing from the tool.drcachesim.drstatecmp-delay-simple test it reports a statecmp error:
```
$ bin64/drrun -t drcachesim -trace_after_instrs 20000 -enable_drstatecmp -- suite/tests/bin/simple_app
Application basic block where mismatch detected: 
TAG  0x00007f5f9756a56e
 +0    L3 @0x00007f5d531ae1e8  49 8b 4f 68          mov    0x68(%r15)[8byte] -> %rcx
 +4    L3 @0x00007f5d531a8950  48 8b 72 08          mov    0x08(%rdx)[8byte] -> %rsi
 +8    L3 @0x00007f5d531ac2d0  48 8d 3d 90 d3 01 00 lea    <rel> 0x00007f5f9758790d -> %rdi
 +15   L3 @0x00007f5d531b15e0  48 03 71 08          add    0x08(%rcx)[8byte] %rsi -> %rsi
 +19   L3 @0x00007f5d531b6508  b9 0a 00 00 00       mov    $0x0000000a -> %ecx
 +24   L3 @0x00007f5d531a9220  f3 a6                rep cmps %ds:(%rsi)[1byte] %es:(%rdi)[1byte] %rsi %rdi %rcx -> %rsi %rdi %rcx
 +26   L3 @0x00007f5d531b6b28  0f 97 c2             setnbe  -> %dl
 +29   L3 @0x00007f5d531afeb8  80 da 00             sbb    $0x00 %dl -> %dl
 +32   L3 @0x00007f5d531b1560  84 d2                test   %dl %dl
 +34   L3 @0x00007f5d531aedc0  0f 85 e1 fa ff ff    jnz    $0x00007f5f9756a077
END 0x00007f5f9756a56e

ASSERT FAILURE: ext/drstatecmp/drstatecmp.c:430: false (xdi)
```

Looks like drstatecmp uses the pre-app2app ilist when the block is truncated, so
the reg state result incorrectly doesn't match:

```
after instrumentation:
TAG  0x00007f5f9756a56e
 <...>
 +638  L3 @0x00007f5d531b7990  f3 a6                rep cmps %ds:(%rsi)[1byte] %es:(%rdi)[1byte] %rsi %rdi %rcx -> %rsi %rdi %rcx
 +640  L3 @0x00007f5d531b5b88  0f 97 c2             setnbe  -> %dl
 +643  L3 @0x00007f5d530b3090  80 da 00             sbb    $0x00 %dl -> %dl
 +646  L3 @0x00007f5d531ac250  84 d2                test   %dl %dl
 <...>
END 0x00007f5f9756a56e
```

derekbruening:
Checks for block truncation in drstatecmp so it can use the post-app2app ilist instead of incorrectly using the untruncated version in its state comparison.

Adds a test of drcachesim offline that fails with a false positive without this fix.

Fixes #5683

### DIFF
--- a/suite/tests/CMakeLists.txt
+++ b/suite/tests/CMakeLists.txt
@@ -4102,8 +4102,10 @@ if (BUILD_CLIENTS)
       endif ()
     endif ()
     if (NOT WIN32)
-      torunonly_drcachesim(drstatecmp-delay-simple ${ci_shared_app}
-        "-trace_after_instrs 20000 -exit_after_tracing 10000 -enable_drstatecmp" "")
+      if (NOT ANDROID) # Pipes not working on Android yet (i#1874)
+        torunonly_drcachesim(drstatecmp-delay-simple ${ci_shared_app}
+          "-trace_after_instrs 20000 -exit_after_tracing 10000 -enable_drstatecmp" "")
+      endif ()
       torunonly_drcacheoff(drstatecmp-delay-simple ${ci_shared_app}
         "-trace_after_instrs 20000 -enable_drstatecmp"
         "@-simulator_type@basic_counts" "")
--- a/suite/tests/CMakeLists.txt
+++ b/suite/tests/CMakeLists.txt
@@ -3470,14 +3470,6 @@ if (BUILD_CLIENTS)
         "-simulator_type miss_analyzer -miss_count_threshold 5000 -miss_frac_threshold 0.25" "")
     endif ()
 
-    # Stress-test drcachesim with drstatecmp checks.
-    if (AARCH64)
-      torunonly_drcachesim(drstatecmp-fuzz drstatecmp-fuzz-app "-trace_after_instrs 10M -enable_drstatecmp" "" "")
-    endif ()
-    if (NOT WIN32)
-      torunonly_drcachesim(drstatecmp-delay-simple ${ci_shared_app} "-trace_after_instrs 20000 -exit_after_tracing 10000 -enable_drstatecmp" "")
-    endif (NOT WIN32)
-
     # Test other analysis tools
     macro (torunonly_simtool testname exetgt sim_ops app_args)
       torunonly_ci(tool.${testname} ${exetgt} drcachesim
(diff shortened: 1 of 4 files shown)


suite/tests/CMakeLists.txt
```diff
@@ -4044,6 +4036,19 @@ if (BUILD_CLIENTS)
...
 
+    # Stress-test drcachesim with drstatecmp checks.
+    if (AARCH64)
+      torunonly_drcachesim(drstatecmp-fuzz drstatecmp-fuzz-app
+        "-trace_after_instrs 10M -enable_drstatecmp" "" "")
+    endif ()
+    if (NOT WIN32)
+      torunonly_drcachesim(drstatecmp-delay-simple ${ci_shared_app}
+        "-trace_after_instrs 20000 -exit_after_tracing 10000 -enable_drstatecmp" "")
+      torunonly_drcacheoff(drstatecmp-delay-simple ${ci_shared_app}
```
abhinav92003, line 4047: How does this new test hit the app2app-truncation case? drmemtrace would truncate bbs in app2app only for scatter-gather instrs, right?

suite/tests/CMakeLists.txt
```diff
@@ -4044,6 +4036,19 @@ if (BUILD_CLIENTS)
         "${prefetch_analyzer_path}@-trace_dir@${testname_full}.*.dir/trace")
     endif ()
 
+    # Stress-test drcachesim with drstatecmp checks.
+    if (AARCH64)
+      torunonly_drcachesim(drstatecmp-fuzz drstatecmp-fuzz-app
+        "-trace_after_instrs 10M -enable_drstatecmp" "" "")
+    endif ()
+    if (NOT WIN32)
+      torunonly_drcachesim(drstatecmp-delay-simple ${ci_shared_app}
```
abhinav92003, line 4045: Would be better to rename this old one. Otherwise it looks like both of these tests are supposed to be exactly the same except online vs offline. Maybe drstatecmp-delay-early-exit?

abhinav92003 (approved):

suite/tests/CMakeLists.txt
derekbruening: It truncates rep strings.

derekbruening: I.e., it truncates the block leading up to the rep string to isolate the rep string to its own bb, and truncates that as well.

suite/tests/CMakeLists.txt
derekbruening: I'm actually not sure why this has the exit flag: without the flag it also hits the bug so I wonder if someone added it to mask the bug??  Or to reduce test runtime and never noticed the bug.  Maybe I should just remove the flag.

suite/tests/CMakeLists.txt
abhinav92003: But ci_shared_app (suite/tests/client-interface/simple_app.c) is just a simple "hello world" program. Or does it contain repstr?

suite/tests/CMakeLists.txt
derekbruening: I removed it.  If the tests are green we'll go with that.

derekbruening:
x64 failure is detach_state assert in vmareas.c: #5123

suite/tests/CMakeLists.txt
derekbruening: hello,world runs over 100K instructions in ld.so and crt setup and is very likely to have multiple repstrs in there.  I guess it's not guaranteed if on a platform where the mem* and str* routine ifuncs resolve to an xmm-based scheme.

PR #5685, opened 2022-10-13
i#5520 memtrace encodings: Add new feature slides pdf

derekbruening:
Adds a "News" line to the drmemtrace docs with a link to a new pdf of slides explaining the new embedded instruction encodings along with a preview of the forthcoming seek feature.

Issue: #5520


api/docs/CMake_doxyfile.cmake
```diff
@@ -128,6 +128,11 @@ string(REGEX REPLACE
   "(footer.html)"
   "\"${dest_dir}/\\1\"" string "${string}")
 
+# If we end up with more of these we should generalize and pass them in
+# rather than hardcoding here.
+string(APPEND string
+  "HTML_EXTRA_FILES=${proj_bindir}/clients/drcachesim/docs/new-features-encodings-seek.pdf")
```
abhinav92003, line 134: Just curious why creating a presentation pdf was preferred over maybe a detailed entry in release.dox.

abhinav92003 (approved):

api/docs/CMake_doxyfile.cmake
derekbruening: We had created these slides for presentations already and wanted to place them somewhere they could be referred to later.

Issue #5686, opened 2022-10-17
drbbdup fails to restore state for its register spills


derekbruening:
I'm hitting another issue with drbbdup when we try to use it by default for drmemtrace #2039: drbbdup spills one or two registers and possibly the flags for its dispatch logic, using custom spills (because its control flow doesn't mesh well with drreg).  But it doesn't have a restore state event.  So a detach or other asynch translation event will mess up the app register state.  We are seeing crashes post-detach that we suspect are coming from this.

derekbruening:
Another problem is that drbbdup calls dr_get_dr_segment_base() dynamically to get the TLS base: but for detach where that's called from another thread we have to store the base in TLS so we can look it up from another thread.

derekbruening:
For drbbdup's own spills I have local code restoring them which seems to work well though it is not easy to write a test because unlike other restore-state events there is no app code here and so we will never have a fault: we will only restore state for an asynch translation which is really hard to make deterministic when we can't tweak the code in question to loop or something.

But in a larger proprietary test app I see what look like errors like this one:
```
drbbdup_event_restore_state: restoring rcx at 0x000000004006cae9 (+9) from slot 1 from 0x00007f7844f49e68 to 0x0000000000000028

(gdb) x/18i 0x000000004006cae9-9
   0x4006cae0:  mov    %rcx,%gs:0x120
   0x4006cae9:  movabs $0x55a73d8e3bd0,%rcx
   0x4006caf3:  mov    (%rcx),%rcx
   0x4006caf6:  jrcxz  0x4006cafd
   0x4006caf8:  jmpq   0x4006cb02
   0x4006cafd:  jmpq   0x4006cb18
   0x4006cb02:  mov    %gs:0x120,%rcx
```
That is +9 from the bb (non-ibl) start, which is the bbdup dispatch.  So bbdup has spilled rcx but has not yet written to the local reg.  So the restore logic should have the reg value and the spill slot value as identical.  Yet the spill slot is 0x28.  ?!?!  I thought maybe drreg had messed up the reg value but I don't think so.  No theory as to how the spill slot could be wrong or corrupted somehow.  Stumped.

derekbruening:
For other clients like drx scatter:

So either drbbdup pprovides its own restore events per copy, or it provides a
query for whether at a copy boundary and drx then resets its state machine
(maybe drbbdup_is_first_instr() already does this: but that only works from
[...]

PR #5689, opened 2022-10-18
i#5688: Fix uninitialized state restore field

### LINKED ISSUE #5688: restore state event is passed an uninitialized ilist field for non-fcache cases
instrument_restore_nonfcache_state_prealloc() creates a dr_restore_state_info_t
but leaves its fragment_info.ilist field unintialized.  I hit this in #5686 where my
new bbdup restore event was crashing walking this bogus instruction list.

Could the uninit ilist explain any unknown filed crashes?
Looks like drreg already has a check for cache_start_pc being NULL and it bails on both sides of its ilist==NULL check so it happens to handle any invalid ilist value.

derekbruening:
instrument_restore_nonfcache_state_prealloc() was creating a dr_restore_state_info_t structure but leaving its fragment_info.ilist field unintialized.  We fix that here.

Tested on the forthcoming drbbdup state restore event (#5686) which crashed walking the bogus ilist without this fix.

Fixes #5688

### DIFF
--- a/core/lib/dr_events.h
+++ b/core/lib/dr_events.h
@@ -682,7 +682,8 @@ typedef struct _dr_fault_fragment_info_t {
     /**
      * The start address of the code fragment inside the code cache at
      * the exception/signal/translation interruption point. NULL for interruption
-     * not in the code cache.  Clients are cautioned when examining
+     * not in the code cache (in which case generally only unusual cases of clients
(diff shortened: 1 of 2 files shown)


derekbruening:
clang-format failure is some presumably transient network issue

abhinav92003 (approved):

PR #5692, opened 2022-10-20
Support DT_RELR relative relocation format

MaskRay:
If elf.h defines DT_RELR (glibc 2.36, musl 1.2.4, FreeBSD https://reviews.freebsd.org/D32524), relocate it.
This allows libdynamorio.so to be linked with -z pack-relative-relocs.


core/unix/module_elf.c
```diff
@@ -1616,3 +1618,28 @@ module_relocate_rela(app_pc modbase, os_privmod_data_t *pd, ELF_RELA_TYPE *start
...
+/* This routine is duplicated in privload_relocate_relr for relocating
+ * dynamorio symbols in a bootstrap stage. Any update here should be also
+ * updated in privload_relocate_relr.
+ */
+module_relocate_relr(app_pc modbase, os_privmod_data_t *pd, const ELF_WORD *relr,
+                     size_t size)
+{
+    ELF_ADDR *r_addr = NULL;
+
+    LOG(GLOBAL, LOG_LOADER, 4, "%s walking relr %p-%p\n", relr, (char *)relr + size);
```
derekbruening, line 1632: The %s has no arg: I'm surprised the compiler doesn't warn about mismatched types.  Probably this is meant to pas `__FUNCTION__`.

core/unix/loader.c
```diff
@@ -1123,6 +1123,29 @@ privload_relocate_rela(os_privmod_data_t *opd, ELF_RELA_TYPE *start, ELF_RELA_TY
...
+/* XXX: This routine is called before dynamorio relocation when we are in a
+ * fragile state and thus no globals access or use of ASSERT/LOG/STATS!
+ */
+/* This routine is duplicated from module_relocate_relr for relocating dynamorio. */
+static void
+privload_relocate_relr(os_privmod_data_t *opd, const ELF_WORD *relr, size_t size)
+{
+    ELF_ADDR *r_addr = NULL;
+
+    for (; size; relr++, size -= sizeof(ELF_WORD)) {
```
derekbruening, line 1135: style: We use explicit bools so `size != 0;`

(Ditto on the other copy for all of these.)

core/unix/loader.c
```diff
@@ -1123,6 +1123,29 @@ privload_relocate_rela(os_privmod_data_t *opd, ELF_RELA_TYPE *start, ELF_RELA_TY
...
+        if ((relr[0] & 1) == 0) {
```
derekbruening, line 1136: style: For bit tests we use `!TEST(1, relr[0])`

core/unix/loader.c
```diff
@@ -1123,6 +1123,29 @@ privload_relocate_rela(os_privmod_data_t *opd, ELF_RELA_TYPE *start, ELF_RELA_TY
...
+            r_addr = (ELF_ADDR *)(relr[0] + opd->load_delta);
+            *r_addr++ += opd->load_delta;
+        } else {
+            int i = 0;
+            for (ELF_WORD bitmap = relr[0]; (bitmap >>= 1); i++)
```
derekbruening, line 1141: style: braces around the multi-line body (that is what is failing the tests)

derekbruening (approved):
Thank you for adding this support.  Ideally there'd be some kind of automated unit test but we don't have a framework for that for the existing relocation code either...maybe note that this was tested with whatever clang version in the commit description.

core/unix/module_elf.c
derekbruening: To expedite things I went and edited this directly.  I see two other logs above are also missing the arg; fixed.

core/unix/loader.c
derekbruening: Fixed in both.

core/unix/loader.c
derekbruening: Fixed in both.

core/unix/loader.c
```diff
@@ -1123,6 +1123,29 @@ privload_relocate_rela(os_privmod_data_t *opd, ELF_RELA_TYPE *start, ELF_RELA_TY
...
+                if (bitmap & 1)
```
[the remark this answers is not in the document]
derekbruening: Fixed in both.

derekbruening:
run arm tests

PR #5693, opened 2022-10-20
Add initialization of emulation_info to exclude unavoidable DR_ASSERT.

kuhanov:
Default initialization produces DR_ASSERT each time because drmgr_get_emulated_instr_data retruns false in this case
```
        ...
        } else if (drmgr_is_emulation_start(instr)) {
            bool ok = drmgr_get_emulated_instr_data(instr, &emulation_info);
            DR_ASSERT(ok);
        ...
```
```
DR_EXPORT
bool
drmgr_get_emulated_instr_data(instr_t *instr, emulated_instr_t *emulated)
{
    ASSERT(instr_is_label(instr), "emulation instruction does not have a label");
    ASSERT(drmgr_is_emulation_start(instr), "instruction is not a start emulation label");

    if (emulated->size < offsetof(emulated_instr_t, flags))
        return false;
   ...
```

### DIFF
--- a/clients/drcachesim/tests/burst_gencode.cpp
+++ b/clients/drcachesim/tests/burst_gencode.cpp
@@ -132,12 +132,12 @@ private:
         // instead of the pop into pc which assumes the entry pushed lr?
         instrlist_append(ilist, INSTR_CREATE_bx(dc, opnd_create_reg(DR_REG_LR)));
 #else
-#ifdef X86
+#    ifdef X86
         instrlist_append(ilist,
                          INSTR_CREATE_xor(dc, opnd_create_reg(DR_REG_RCX),
                                           opnd_create_reg(DR_REG_RCX)));
         instrlist_append(ilist, INSTR_CREATE_rep_movs_1(dc));
-#endif
+#    endif
         instrlist_append(ilist, XINST_CREATE_return(dc));
 #endif
 
--- a/clients/drcachesim/tests/burst_gencode.cpp
+++ b/clients/drcachesim/tests/burst_gencode.cpp
@@ -134,7 +134,7 @@ private:
 #else
         instrlist_append(ilist,
                          INSTR_CREATE_xor(dc, opnd_create_reg(DR_REG_RCX),
-                                            opnd_create_reg(DR_REG_RCX)));
+                                          opnd_create_reg(DR_REG_RCX)));
         instrlist_append(ilist, INSTR_CREATE_rep_movs_1(dc));        
         instrlist_append(ilist, XINST_CREATE_return(dc));
 #endif
--- a/clients/drcachesim/tests/burst_gencode.cpp
+++ b/clients/drcachesim/tests/burst_gencode.cpp
@@ -133,7 +133,7 @@ private:
         instrlist_append(ilist, INSTR_CREATE_bx(dc, opnd_create_reg(DR_REG_LR)));
 #else
 #    ifdef X86
-        // Zero-iter rep-movs loop for testing emulation-marked code 
+        // Zero-iter rep-movs loop for testing emulation-marked code
         // written to the encoding file.
         instrlist_append(ilist,
                          INSTR_CREATE_xor(dc, opnd_create_reg(DR_REG_XCX),
--- a/clients/drcachesim/tests/burst_gencode.cpp
+++ b/clients/drcachesim/tests/burst_gencode.cpp
@@ -135,7 +135,7 @@ private:
         instrlist_append(ilist,
                          INSTR_CREATE_xor(dc, opnd_create_reg(DR_REG_RCX),
                                           opnd_create_reg(DR_REG_RCX)));
-        instrlist_append(ilist, INSTR_CREATE_rep_movs_1(dc));        
+        instrlist_append(ilist, INSTR_CREATE_rep_movs_1(dc));
         instrlist_append(ilist, XINST_CREATE_return(dc));
 #endif
 
--- a/clients/drcachesim/tests/burst_gencode.cpp
+++ b/clients/drcachesim/tests/burst_gencode.cpp
@@ -133,11 +133,14 @@ private:
         instrlist_append(ilist, INSTR_CREATE_bx(dc, opnd_create_reg(DR_REG_LR)));
 #else
 #    ifdef X86
+        // Zero-iter rep-movs loop for testing emulation-marked code 
+        // written to the encoding file.
         instrlist_append(ilist,
-                         INSTR_CREATE_xor(dc, opnd_create_reg(DR_REG_RCX),
-                                          opnd_create_reg(DR_REG_RCX)));
+                         INSTR_CREATE_xor(dc, opnd_create_reg(DR_REG_XCX),
+                                          opnd_create_reg(DR_REG_XCX)));
         instrlist_append(ilist, INSTR_CREATE_rep_movs_1(dc));
 #    endif
+        
         instrlist_append(ilist, XINST_CREATE_return(dc));
 #endif
 
--- a/clients/drcachesim/tests/burst_gencode.cpp
+++ b/clients/drcachesim/tests/burst_gencode.cpp
@@ -140,7 +140,7 @@ private:
                                           opnd_create_reg(DR_REG_XCX)));
         instrlist_append(ilist, INSTR_CREATE_rep_movs_1(dc));
 #    endif
-        
+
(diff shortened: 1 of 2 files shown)


clients/drcachesim/tracer/instru_offline.cpp
```diff
@@ -476,7 +476,7 @@ offline_instru_t::record_instr_encodings(void *drcontext, app_pc tag_pc,
     for (instr_t *instr = instrlist_first(ilist); instr != NULL;
          instr = instr_get_next(instr)) {
         instr_t *to_copy = nullptr;
-        emulated_instr_t emulation_info = {};
+        emulated_instr_t emulation_info = { sizeof(emulation_info), 0 };
```
derekbruening, line 479: Thank you for the fix.  I guess the only current test of encodings during tracing is generated code without emulation and there is no test to cover this.  On what application did you catch this?  Would it be feasible to turn a simplified version of it into a small regression test added to this PR?

derekbruening (approved):
On what application did you catch the bug?  Could we turn that into

clients/drcachesim/tracer/instru_offline.cpp
kuhanov: I'm not sure that we have simple reproducer here.
I've just run java profiling without any java workload
./bin64/drrun -sandbox2ro_threshold 0 -ro2sandbox_threshold 0 -disable_traces -t drcachesim -offline -- java

kuhanov: And native binaries without dynamic code didn't produce issue again. ))

derekbruening: Hmm, I wonder where the emulation markers came from in the Java case?  Maybe a generated rep string loop or a generated scatter-gather?  I guess one option is to add some emulation markers to the burst_gencode.cpp test: to avoid having to manually add made-up markers in a new client maybe we can just throw a rep-movs loop in there: even make it zero iters so we don't have to really set up the targets to be real data?  I.e., just add one line under ifdef X86 that appends a rep-movs here https://github.com/DynamoRIO/dynamorio/blob/master/clients/drcachesim/tests/burst_gencode.cpp#L130 and see if that repros the failure?  If that ends up being more work than one line we can bail and just put this fix in but it seems worth some effort for a test.

derekbruening: Sorry that would be two added statements (not lines): have to set rcx to zero for the empty loop.

kuhanov: did you mean something like this?
```
diff --git a/clients/drcachesim/tests/burst_gencode.cpp b/clients/drcachesim/tests/burst_gencode.cpp
index cd1911dbb..eb930f8b4 100644
--- a/clients/drcachesim/tests/burst_gencode.cpp
+++ b/clients/drcachesim/tests/burst_gencode.cpp
@@ -133,6 +133,9 @@ private:
         instrlist_append(ilist, INSTR_CREATE_bx(dc, opnd_create_reg(DR_REG_LR)));
 #else
         instrlist_append(ilist, XINST_CREATE_return(dc));
+        instrlist_append(ilist, XINST_CREATE_xor(dc, opnd_create_reg(DR_REG_RCX),
+            opnd_create_reg(DR_REG_RCX)));
+        instrlist_append(ilist, XINST_CREATE_rep_movs_1(dc));
 #endif

         byte *last_pc = instrlist_encode(dc, ilist, map_, true);

```

derekbruening: It needs to be before the return.  Does that reproduce the bug if the fix is removed?

kuhanov: > drcachesim
```
diff --git a/clients/drcachesim/tests/burst_gencode.cpp b/clients/drcachesim/tests/burst_gencode.cpp
index cd1911dbb..c585ac7d5 100644
--- a/clients/drcachesim/tests/burst_gencode.cpp
+++ b/clients/drcachesim/tests/burst_gencode.cpp
@@ -132,6 +132,10 @@ private:
         // instead of the pop into pc which assumes the entry pushed lr?
         instrlist_append(ilist, INSTR_CREATE_bx(dc, opnd_create_reg(DR_REG_LR)));
 #else
+        instrlist_append(ilist,
+                         INSTR_CREATE_xor(dc, opnd_create_reg(DR_REG_RCX),
+                                            opnd_create_reg(DR_REG_RCX)));
+        instrlist_append(ilist, INSTR_CREATE_rep_movs_1(dc));
         instrlist_append(ilist, XINST_CREATE_return(dc));
 #endif
```

kuhanov: Gencode test result with test patch only
```
kuhanov@kuhanov:~/workspace/drmemtrace_test/dynamorio/build_test$ ctest --output-on-failure  -R  tool.drcacheoff.gencode

Test project /home/kuhanov/workspace/drmemtrace_test/dynamorio/build_test
    Start 325: code_api|tool.drcacheoff.gencode
1/1 Test #325: code_api|tool.drcacheoff.gencode ...***Failed    0.08 sec
Running |/usr/bin/cmake;-E;remove_directory /home/kuhanov/workspace/drmemtrace_test/dynamorio/build_test/suite/tests/drmemtrace.tool.drcacheoff.gencode.2048400.2656.dir|
Running cmd |/home/kuhanov/workspace/drmemtrace_test/dynamorio/build_test/suite/tests/bin/tool.drcacheoff.gencode|
CMake Error at /home/kuhanov/workspace/drmemtrace_test/dynamorio/suite/tests/runmulti.cmake:108 (message):
  *** cmd failed (255): pre-DR init

  pre-DR start

  ASSERT FAILURE:
  /home/kuhanov/workspace/drmemtrace_test/dynamorio/clients/drcachesim/tracer/instru_offline.cpp:485:
  ok ()

  ***

Call Stack (most recent call first):
  /home/kuhanov/workspace/drmemtrace_test/dynamorio/suite/tests/runmulti.cmake:117 (process_cmdline)




Label Time Summary:
RUN_IN_RELEASE    =   0.08 sec*proc (1 test)

Total Test time (real) =   2.60 sec

The following tests FAILED:
        325 - code_api|tool.drcacheoff.gencode (Failed)
Errors while running CTest
```

kuhanov: Gencode test result with test patch and structure initialization
```
kuhanov@kuhanov:~/workspace/drmemtrace_test/dynamorio/build_test$ ctest --output-on-failure  -R  tool.drcacheoff.gencode
Test project /home/kuhanov/workspace/drmemtrace_test/dynamorio/build_test
    Start 325: code_api|tool.drcacheoff.gencode
1/1 Test #325: code_api|tool.drcacheoff.gencode ...   Passed    0.19 sec

100% tests passed, 0 tests failed out of 1

Label Time Summary:
RUN_IN_RELEASE    =   0.19 sec*proc (1 test)

Total Test time (real) =   2.74 sec
```

kuhanov: So, the patched test has the same assert like in java collection.
And patch is fixed it

derekbruening: Nice, thank you for adding the test.

clients/drcachesim/tests/burst_gencode.cpp
```diff
@@ -132,6 +132,12 @@ class code_generator_t {
         // instead of the pop into pc which assumes the entry pushed lr?
         instrlist_append(ilist, INSTR_CREATE_bx(dc, opnd_create_reg(DR_REG_LR)));
 #else
+#    ifdef X86
+        instrlist_append(ilist,
```
derekbruening, line 136: I would add a comment explaining that we're inserting this zero-iter rep string loop to test emulation-marked code written to the encoding file.

derekbruening (approved):

clients/drcachesim/tests/burst_gencode.cpp
```diff
@@ -132,6 +132,12 @@ class code_generator_t {
...
+                         INSTR_CREATE_xor(dc, opnd_create_reg(DR_REG_RCX),
+                                          opnd_create_reg(DR_REG_RCX)));
```
derekbruening, line 138: If you use DR_REG_XCX this will work on 32-bit too

Issue #5437, opened 2022-03-29
SIGFPE in private __libc_early_init in glibc 2.34+


derchr:
**Describe the bug**
This bug may not only affect drcachesim but also drmemory, drcpusim and probably other clients as well.

When I run drcachesim like this: `./drrun -disable_rseq -t drcachesim -offline -- ls` I get a SIGFPE:
`[1]    2736 floating point exception (core dumped)  ./drrun -disable_rseq -t drcachesim -offline -- ls`

**To Reproduce**
1. Pointer to a minimized application: `ls` should work
2. Precise command line for running the application. `./drrun -disable_rseq -t drcachesim -offline -- ls`
3. Exact output or incorrect behavior. See above

Please also answer these questions:
 - What happens when you run without any client?
Without any client works (thanks to -disable_rseq)

 - What happens when you run with debug build ("-debug" flag to drrun/drconfig/drinject)?
 Same behaviour

**Expected behavior**
No crash

**Screenshots or Pasted Text**
<pre>
```
<Initial options = -no_dynamic_options -client_lib '/home/derek/Git/dynamorio/build/bin64/../clients/lib64/debug/libdrmemtrace.so;0;-offline' -client_lib64 '/home/derek/Git/dynamorio/build/bin64/../clients/lib64/debug/libdrmemtrace.so;0;-offline' -client_lib32 '/home/derek/Git/dynamorio/build/bin64/../clients/lib32/debug/libdrmemtrace.so;0;-offline' -code_api -stack_size 56K -signal_stack_size 32K -max_elide_jmp 0 -max_elide_call 0 -early_inject -emulate_brk -no_inline_ignored_syscalls -disable_rseq -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct >

Program received signal SIGFPE, Arithmetic exception.
0x00007ffff770eb89 in ?? ()
(gdb) bt
#0  0x00007ffff770eb89 in ?? ()
#1  0x0000000000800000 in ?? ()
#2  0xffffffffffffffff in ?? ()
#3  0x0000000000000007 in ?? ()
#4  0xabababababababab in ?? ()
#5  0x00007ffff7ff0810 in ?? ()
#6  0x00007ffff7e76f77 in privload_os_finalize (privmod=0x7ffdb3ba77d8)
    at /home/derek/Git/dynamorio/core/unix/loader.c:693
[...]

Issue #5431, opened 2022-03-25
CRASH from new glibc 2.35 rseq on any app (-disable_rseq solves)


derchr:
**Describe the bug**
Running DynamoRIO to instrument any application, a crash occurs.
`./drrun -- grep` (no client at all) results in:
```
<Starting application /usr/bin/grep (331303)>
<Initial options = -no_dynamic_options -code_api -stack_size 56K -signal_stack_size 32K -max_elide_jmp 0 -max_elide_call 0 -early_inject -emulate_brk -no_inline_ignored_syscalls -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct >
<Paste into GDB to debug DynamoRIO clients:
set confirm off
add-symbol-file '/home/derek/Git/dynamorio/build/lib64/debug/libdynamorio.so' 0x00007f972794a000
>
<(1+x) Handling our fault in a TRY at 0x00007f9727bbd98c>
<spurious rep/repne prefix @0x00007f97278a9650 (f3 0f 1e fa): >
<Application /usr/bin/grep (331303).  DynamoRIO internal crash at PC 0x00007f9727c26fda.  Please report this at http://dynamorio.org/issues/.  Program aborted.
Received SIGSEGV at pc 0x00007f9727c26fda in thread 331303
Base: 0x00007f97278ff000
Registers:eax=0x00007f94e391c0a0 ebx=0x0000000000050e27 ecx=0x00007f9727c29258 edx=0x00000000000009a0
        esi=0x00007f94e392b948 edi=0x0000000000001003 esp=0x00007f94e392b9d0 ebp=0x00007f94e392b9f0
        r8 =0x0000000000000000 r9 =0x0000000000000000 r10=0x00007f94e392b948 r11=0x0000000000000246
        r12=0x00007f97238c66d0 r13=0x00007f97238c66c7 r14=0x00007f97238c66cf r15=0x00007f97238c66dd
        eflags=0x0000000000010206
version 9.0.19068, custom build
-no_dynamic_options -code_api -stack_size 56K -signal_stack_size 32K -max_elide_jmp 0 -max_elide_call 0 -early_inject -emulate_brk -no_inline_ignored_syscalls -native_exec_default_list '' -no_native_exec_managed_code -no_indcall2direct 
0x00007f94e392b9f0 0x00007f9727c27254
0x00007f94e392ba10 0x00007f972798076b
0x00007f94e392ba70 0x00007f9727981bad
[...]

PR #5695, opened 2022-10-21
i#5437 glibc2.34: Workaround for SIGFPE

derekbruening:
Adds a workaround for the SIGFPE in glibc 2.34+ __libc_early_init() by setting two ld.so globals located via hardcoded offsets, making this fragile and considered temporary.

Tested on glibc 2.34 where every libc-using client crashes with SIGFPE but they work with this fix.

Adds an Ubuntu22 GA CI run but if we have failures due to other reasons the plan is to drastically shrink the tests run or abandon if it's too much work right now.

Issue: #5437


derekbruening:
The hardcoded offsets here will break on the next glibc update which inserts a var earlier in that globals list -- and I don't know enough about it to know how often that happens, whether vars are always appended, or what.

One option is to decode __libc_early_init and look for offsets there: but we'd look for hardcoded code patterns and that is also fragile.  Maybe there is some simpler exported function to decode. 

We could make runtime options for the offsets so a user could work around this.

2.34 seems to work if we don't even call __libc_early_init (but 2.32 crashes if we don't): we could avoid the call for 2.34, maybe if a decode of the func doesn't match our expected pattern.

OTOH not sure it's worth a ton of work if we think it's better to put that time into a refactor private loading process.

derekbruening:
I'm also looking for reviewer input on what to do with the ubuntu22 test I added here.  We have the following failures (but many passes indicating this fix is working there else nearly every client test would fail):
```
	code_api|common.fib 
	code_api|common.nativeexec 
	code_api|common.nativeexec_retakeover 
	(ignore: i#2941) 	code_api|common.nativeexec_exe 
	(ignore: i#2941) 	code_api|common.nativeexec_retakeover_opt 
	(ignore: i#2941) 	code_api|common.nativeexec_bindnow 
	code_api|common.nativeexec_exe_opt 
	(ignore: i#2941) 	code_api|common.nativeexec_bindnow_opt 
	code_api|security-common.codemod 
	code_api|security-common.decode-bad-stack_FLAKY 
	code_api|security-common.selfmod2 
	code_api|security-common.selfmod-big 
	code_api|security-common.selfmod 
	code_api|security-common.TestAllocWE 
	code_api|security-common.TestMemProtChg_FLAKY 
	code_api|tool.drcachesim.phys_SUDO 
	code_api|tool.drcachesim.phys-threads_SUDO 
	code_api|tool.drcachesim.delay-global 
	code_api|tool.drcachesim.threads 
	code_api|tool.drcachesim.threads-with-config-file 
	code_api|tool.drcachesim.coherence 
	code_api,satisfy_w_xor_x|security-common.selfmod2 
```
Actually, some if not all of those are the rseq bug #5431.

derekbruening:
Actually that list is release build: many more failed on debug; but again they are mostly the rseq #5431.

.github/workflows/ci-x86.yml
```diff
@@ -196,3 +196,66 @@ jobs:
           See more details on github.com/DynamoRIO/dynamorio/actions/runs/${{github.run_id}}
         to: dynamorio-devs@googlegroups.com
         from: Github Action CI
+
+  # Ubuntu22 64-bit Linux build with gcc and run tests:
+  x86-64-ubuntu22:
+    runs-on: ubuntu-22.04
```
abhinav92003, line 202: If this config is same as the other ubuntu one, we can use a run matrix: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

abhinav92003 (approved):

.github/workflows/ci-x86.yml
derekbruening: Done.

derekbruening: There is some negative interaction between the matrix and our auto-cancel: as soon as the first job is done the 2nd is cancelled.  Reverting the matrix with a comment.

tools/drdeploy.c
```diff
@@ -1270,6 +1282,20 @@ _tmain(int argc, TCHAR *targv[])
     /* we re-read the tool list if the root, platform or toolconfig dir change */
     read_tool_list(dr_toolconfig_dir, dr_platform);
 
+#if defined(LINUX) && !defined(ANDROID) && (defined(DRCONFIG) || defined(DRRUN))
+    /* XXX i#5431: Workaround for an rseq issue with glibc 2.35 which makes many
+     * apps fail up front.  We expect to remove this once the real fix is in place.
+     */
+#    include <gnu/libc-version.h>
+    const char *libc_ver = gnu_get_libc_version();
+    if (libc_ver[0] != '\0' && libc_ver[0] >= '2' && libc_ver[1] == '.' &&
```
abhinav92003, line 1291: Didn't pay attention to this before: this condition won't work for all glibc versions >= 2.35, e.g. it would be false for 2.41.

derekbruening: I hope this is all ripped out and replaced with a long-term solution long before that.

abhinav92003: Fair enough.

Issue #5675, opened 2022-10-04
Provide tool to filter drmemtrace offline traces


abhinav92003:
It will be useful to have a tool that can filter out entries that are not needed in an offline drmemtrace trace, after it has been processed by raw2trace.

I imagine this can be implemented as a new drmemtrace analyzer that processes each thread's trace separately. The current trace reader returns a memref_t and hides some entries from the analyzer. We'd need to support reading raw trace_entry_t entries.

abhinav92003:
Now that we have the general infrastructure for filtering in place, we can start working on the required filters. Some that are needed currently:
- positive and negative filter on trace_type_t. "Keep only entries of X type" or "discard all entries of X type"
- split the trace into multiple parts: filter out entries at instr ordinal < X or > X. Currently, we need to run the tool multiple times, once for each "part", since we can output only one trace in each run. Some essential entries, like the trace header/footer, thread id, thread exit might need to be preserved in all "parts".
- filter instr/memory address stream through a simple direct mapped cache.

abhinav92003:
Looks like for traces collected after Oct this year that have instr encodings (Xref #5520), there is a bug in filtering instr records. If we filter out instruction records, then the reader is no longer able to differentiate between TRACE_TYPE_ENCODING entries for different instructions and hits the "Invalid too-large encoding size" error.

```
$ ./clients/bin64/record_filter_launcher --trace_dir drmemtrace.ls.824987.4583.dir/final --output_dir op --remove_trace_types "10,11,12,13,14,15,16"
...
$ ./bin64/drrun -t drcachesim -indir op -simulator_type basic_counts
Invalid too-large encoding size 16 + 2
drcachesim: <dr-src>/clients/drcachesim/reader/reader.cpp:124: virtual bool reader_t::process_input_entry(): Assertion `false' failed.
Aborted
```

We cannot simply also remove all encodings when we're removing instr entries, since the encodings may be required for a later instruction we emit (for filtering done with `-stop_timestamp`). Two ways to solve this:
- in the reader, keep only the last MAX_ENCODING bytes from previously seen TRACE_TYPE_ENCODING entries, and drop the older ones if we find more TRACE_TYPE_ENCODING entries without an instr entry. When we reach an instr entry, look at the expected size and decode that much from the saved encoding entries.
- in the trace filter, cache encoding entries online and emit only if/when that pc is emitted later. May be a bit memory intensive, but would remove the unneeded encodings.

[...]

PR #5697, opened 2022-10-22
i#5675 trace filter: Add basic framework for filtering

abhinav92003:
Adds a tool where trace filtering logic can be implemented. Currently,
it simply writes out everything from the original trace. The actual
filtering logic will follow in next PRs.

Adds trace_entry_file_reader_t that iterates over the raw trace_entry_t
objects in the trace. The current implementation supports reading
gzip'd or uncompressed trace files. Support for filtering zip files will
follow in next PRs.

Issue: #5675


clients/drcachesim/tools/trace_filter.h
```diff
@@ -0,0 +1,145 @@
...
+#define _TRACE_FILTER_H_ 1
+
+#include <memory>
+#include <string>
+#include <vector>
+#include "trace_entry_file_reader.h"
+
+class trace_filter_t {
+public:
+    /**
```
derekbruening, line 47: Two functions here have doxygen comments but the rest don't and there's no file/group header: where do these show up in the html docs?  The other tools don't seem to have doxygen except the create routines (and the abstract base classes).

clients/drcachesim/tools/trace_filter.cpp
```diff
@@ -0,0 +1,217 @@
...
+                               int verbosity)
+    : verbosity_(verbosity)
+    , worker_count_(worker_count)
+    , trace_dir_(trace_dir)
+    , output_dir_(output_dir)
+{
+    if (!init_file_reader_writer(trace_dir, output_dir, verbosity))
+        error_string_ = "Could not initialize trace reader and writer";
+    UNUSED(verbosity_);
+    UNUSED(output_prefix_);
```
derekbruening, line 57: ?  This is not part of this function or class?

clients/drcachesim/tools/trace_filter.cpp
```diff
@@ -0,0 +1,217 @@
...
+                               const std::string &output_dir, int worker_count,
+                               int verbosity)
+    : verbosity_(verbosity)
+    , worker_count_(worker_count)
+    , trace_dir_(trace_dir)
+    , output_dir_(output_dir)
+{
+    if (!init_file_reader_writer(trace_dir, output_dir, verbosity))
+        error_string_ = "Could not initialize trace reader and writer";
+    UNUSED(verbosity_);
```
derekbruening, line 56: Looks unnecessary?

clients/drcachesim/tools/trace_filter.cpp
```diff
@@ -0,0 +1,217 @@
...
+    UNUSED(output_prefix_);
+}
+
+#if defined(HAS_ZLIB)
+static bool
+ends_with(const std::string &str, const std::string &with)
```
derekbruening, line 62: This is directly copied from analyzer.cpp: could we share it

clients/drcachesim/tools/trace_filter.h
```diff
@@ -0,0 +1,145 @@
...
+        FILE_FORMAT_UNKNOWN = 0,
+        FILE_FORMAT_GZIP,
+        // TODO i#5675: Add support for filtering zip files.
+    } file_format_t;
+
+    int verbosity_;
+    int worker_count_;
+    std::string trace_dir_;
+    std::string output_dir_;
+    file_format_t input_file_format;
```
derekbruening, line 134: style: needs _ suffix

clients/drcachesim/tools/trace_filter.cpp
```diff
@@ -0,0 +1,217 @@
...
+    return std::unique_ptr<trace_entry_reader_t>(
+        new default_trace_entry_file_reader_t(path, verbosity));
+}
+
+std::unique_ptr<std::ostream>
+trace_filter_t::get_writer(const std::string &path, int verbosity)
+{
+    std::unique_ptr<std::ostream> ofile;
+#ifdef HAS_ZLIB
+    if (input_file_format == FILE_FORMAT_GZIP) {
```
derekbruening, line 93: This format var is not used anywhere else.  Maybe the reader and writer should be created at the same time to avoid the need to remember this in a separate routine.

clients/drcachesim/tools/trace_filter.cpp
```diff
@@ -0,0 +1,217 @@
...
+    if (worker_count_ <= 0)
+        worker_count_ = std::thread::hardware_concurrency();
+    worker_tasks_.resize(worker_count_);
+    int worker = 0;
+    for (size_t i = 0; i < thread_data_.size(); ++i) {
+        VPRINT(this, 2, "Worker %d assigned trace shard %zd\n", worker, i);
+        worker_tasks_[worker].push_back(&thread_data_[i]);
+        thread_data_[i].worker = worker;
+        worker = (worker + 1) % worker_count_;
+    }
```
derekbruening, line 156: This is all very similar to analyzer_t::init_file_reader, identical in many places.  Maybe this should be a sub-class with just get_reader() overridden to also create an ostream.

clients/drcachesim/tools/trace_filter.cpp
```diff
@@ -0,0 +1,217 @@
...
+    for (std::thread &thread : threads)
+        thread.join();
+    for (auto &tdata : thread_data_) {
+        if (!tdata.error.empty()) {
+            error_string_ = tdata.error;
+            return false;
+        }
+    }
+    return true;
```
derekbruening, line 216: Another function that is basically identical to the analyzer_t code if the serial mode if() is removed.  More evidence to try a subclass.

clients/drcachesim/tools/trace_filter.cpp
```diff
@@ -0,0 +1,217 @@
...
+    for (shard_data_t *tdata : *tasks) {
+        VPRINT(this, 1, "Worker %d starting on trace shard %d\n", tdata->worker,
+               tdata->index);
+        if (!tdata->iter->init()) {
+            tdata->error = "Failed to read from trace" + tdata->trace_file;
+            return;
+        }
+        for (; *tdata->iter != *trace_end_; ++(*tdata->iter)) {
+            const trace_entry_t &entry = **tdata->iter;
+            // TODO i#5675: Filter entries here.
```
derekbruening, line 180: I think we want a framework like for regular trace processing so we can create separate tools rather than hardcoding different filters here.  Even if we anticipate fewer such tools than regular analysis tools, such a setup conforms to the existing analysis tool setup better and seems like it would actually avoid code duplication and so in that sense be simpler in the long run maintenance-wise.

clients/drcachesim/tools/trace_filter.cpp
```diff
@@ -0,0 +1,217 @@
...
+            if (!tdata->writer->write((char *)&entry, sizeof(entry))) {
```
derekbruening, line 181: What if we want to output multiple files?  E.g., we identify 5 regions of interest and want to create separate files for each (as opposed to just inserting markers).  Maybe that's harder to support if we allow multiple filter tools where just one output makes sense so maybe we do not support that but it is worth acknowledging as an explicit design decision.

clients/drcachesim/tests/trace_filter_unit_tests.cpp
```diff
@@ -0,0 +1,132 @@
...
+bool
+test_nop_filter()
+{
+    std::string output_dir = op_tmp_output_dir.get_value() + "/nop_filter";
+    if (!local_create_dir(output_dir.c_str())) {
+        FATAL_ERROR("Failed to create filtered trace output dir %s", output_dir.c_str());
+    }
+    {
+        auto trace_filter = std::unique_ptr<trace_filter_t>(
+            new trace_filter_t(op_trace_dir.get_value(), output_dir));
```
derekbruening, line 103: There is no need for this to be heap-allocated: a local var suffices without any need for a unique_ptr.

clients/drcachesim/tools/basic_counts.cpp
```diff
@@ -282,3 +282,15 @@ basic_counts_t::print_results()
 
     return true;
 }
+
+basic_counts_t::counters_t
+basic_counts_t::get_total_counts()
```
derekbruening, line 287: Hmm, xref #3068

clients/drcachesim/reader/compressed_file_reader.cpp
```diff
@@ -32,6 +32,34 @@
 
 #include "compressed_file_reader.h"
 
+bool
+open_single_file_common(const std::string &path, gzFile &out)
```
derekbruening, line 36: These are not in any class namespace nor are they static.  Let's at least make them static or possibly take a step toward #4343 and add the dynamorio::drmemtrace namespace.

clients/drcachesim/CMakeLists.txt
```diff
@@ -136,6 +136,15 @@ add_exported_library(drmemtrace_func_view STATIC tools/func_view.cpp)
 configure_DynamoRIO_standalone(drmemtrace_opcode_mix)
 configure_DynamoRIO_standalone(drmemtrace_view)
 
+add_exported_library(drmemtrace_trace_filter STATIC tools/trace_filter.cpp
+  reader/reader.cpp
```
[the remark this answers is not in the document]
abhinav92003: On subclass-ing analyzer_t:
- `analyzer_t` works with `reader_t` which returns `memref_t` objects, whereas the trace filter needs to iterate over `trace_entry_t` objects, so it doesn't fit out of the box.
- we don't really need much of the functionality of `analyzer_t`: the support to run multiple tools, serial/parallel mode. One could think of different filters being different tools, but it's much simpler to have them as command-line flags that enable/disable boolean-returning-routines in `trace_filter_t`.
- here, we want to be able to write trace files, which is not typical of analyzers

We could of course change `analyzer_t` to allow us to do all of this, but to  me it seemed simpler to create a standalone tool, with some duplication from `analyzer_t` (some of it could possible be put into a common library I guess), instead of changing `analyzer_t` too much in ways to support two really different paths/use-cases.

clients/drcachesim/tools/trace_filter.cpp
abhinav92003: In some builds, Clang complained about these variables being unused in non-debug mode.

clients/drcachesim/tools/trace_filter.cpp
abhinav92003: See concerns above about sub-classing.

We could put it in a common library, but then `shard_data_t` is different here: has output file and stream objects too.

clients/drcachesim/tools/trace_filter.cpp
abhinav92003: You mean separate tools for filtering? Having a separate sub-class for each kind of filter would probably add boilerplate. I was thinking of adding each filter as a bool-returning-routine that is toggled using a command-line flag. As you noted and as discussed offline previously, regular trace analysis would continue to be on memref_t and this is better off as a standalone tool.

clients/drcachesim/CMakeLists.txt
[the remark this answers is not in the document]
derekbruening: I guess I don't see future filters fitting into the mold of small functions all inside a single file.  That mold seems to only apply to one of the 5+ filters we've thought about.  Various filters we've proposed include region-of-interest extraction, which would likely take in a config file and have to parse it which seems a little much for a small function; truncating endpoints to eliminate staggered starts and ends, which might be most naturally implemented as two passes; applying cache filters, which might want an embedded cache simulator; creating per-core traces, which might want to operate in serial mode.  These do not seem like simple single-function operations.  Some seem like they want both a high-level memref_t iteration and a trace record editing iteration at the same time, actually, and more than one might want serial vs parallel modes.  While we may not implement any of these ourselves as we have either decided to go with alternatives or they aren't high priority, they show that there are a variety of complex tools in this space: they are not all going to be very simple "delete records of static type X" filters that take just a few lines of code deciding whether to keep or throw out each record.  It may not be worth building a too-fancy infrastructure prematurely, but it seems better to keep things open and flexible with separate components with an eye on the future.

If we don't subclass there should at least be comments acknowledging the code duplication.  It still seems that analyzer_t is not really tied to the particular iterator, and its general patterns of opening readers, managing thread pools, running streaming tools are the same for both use cases and it could be subclassed without much change.  Given that the other future filters might each need different framework shapes (memref + trace_entry at once; multiple outputs; etc.) that are more complex to all handle at once, maybe starting with something more specific for just one filter is reasonable; but let's document that we think it will need generalization and refactoring to work with other filters.

Maybe get a 3rd opinion too: maybe @bete0 could take a look or we could have an offline discussion.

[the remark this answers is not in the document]
derekbruening: I forgot another filter we wanted: stripping encoding entries and function trace entries for confidentiality reasons.  That one is a simple filter.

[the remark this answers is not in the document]
abhinav92003: > which seems a little much for a small function

I see the value in passing in a list of `filter_func_t` class objects while constructing `trace_filter_t`, each of which export a bool-returning `should_filter_out(trace_entry_t*)` routine. Taking the filtering logic as an example (which is currently the most complex filter we need), it will certainly make it cleaner.

> truncating endpoints to eliminate staggered starts and ends, which might be most naturally implemented as two passes

Yes, such use cases will likely take two passes. I can imagine that the `filter_func_t` proposed above will then require subclasses to implement an additional `pre_filter_pass()` routine, and `trace_filter_t` will make two loops over the trace.  Multi-pass filters can be left to future work. It's not required currently. 

>  Some seem like they want both a high-level memref_t iteration

trace_filter_t needs to primarily iterate over trace_entry_t. If we do need something in the form of memref_t we can reuse the converter code from reader_t; but sometimes it may not be as simple. I think the current use cases can work with trace_entry_t though.

> one might want serial vs parallel modes.

I couldn't think of any filtering use case that would need serial iteration. My thought was that filters can always treat each thread's trace separately. Do you have any such use case in mind?

> If we don't subclass there should at least be comments acknowledging the code duplication. 

Will do.

>  It still seems that analyzer_t is not really tied to the particular iterator

I agree that there's enough could-be-shared code to want to share more implementation with analyzer_t. I based my decision on the many usages of `reader_t` in the analyzer_t public API  and many in private implementation details too. Also, as I said, the thread shard data needs to be different since we have to also write, and not just read.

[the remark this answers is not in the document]
abhinav92003: I was going to do that per-thread, since it's much faster to not serialize, especially for super long traces.

clients/drcachesim/CMakeLists.txt
```diff
@@ -285,6 +296,10 @@ target_link_libraries(prefetch_analyzer_launcher drmemtrace_analyzer drfrontendl
 use_DynamoRIO_extension(prefetch_analyzer_launcher droption)
 add_dependencies(prefetch_analyzer_launcher api_headers)
 
+add_executable(trace_filter_launcher
+  tools/trace_filter_launcher.cpp)
+target_link_libraries(trace_filter_launcher drmemtrace_trace_filter)
+use_DynamoRIO_extension(trace_filter_launcher droption)
```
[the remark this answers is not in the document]
abhinav92003: Done.

clients/drcachesim/tools/trace_filter_launcher.cpp
```diff
@@ -0,0 +1,89 @@
...
+    op_trace_dir(DROPTION_SCOPE_FRONTEND, "trace_dir", "",
+                 "[Required] Trace input directory",
+                 "Specifies the directory containing the trace files to be filtered.");
+
+static droption_t<std::string>
+    op_output_dir(DROPTION_SCOPE_FRONTEND, "output_dir", "",
+                  "[Required] Output directory for the filtered trace",
+                  "Specifies the directory where the filtered trace will be written.");
+
+droption_t<unsigned int> op_verbose(DROPTION_SCOPE_ALL, "verbose", 0, 0, 64,
```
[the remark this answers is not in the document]
abhinav92003: Done.

clients/drcachesim/tools/trace_filter.h
```diff
@@ -0,0 +1,145 @@
...
+
+#ifndef _TRACE_FILTER_H_
+#define _TRACE_FILTER_H_ 1
+
+#include <memory>
+#include <string>
+#include <vector>
+#include "trace_entry_file_reader.h"
+
+class trace_filter_t {
```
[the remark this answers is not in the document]
abhinav92003: Done.

clients/drcachesim/tools/trace_filter.h
abhinav92003: Removed.

clients/drcachesim/tools/trace_filter.cpp
abhinav92003: Removed.

clients/drcachesim/tools/trace_filter.cpp
abhinav92003: Converted analyzer_t to a template in #5720.

clients/drcachesim/tools/trace_filter.h
abhinav92003: Removed.

clients/drcachesim/tools/trace_filter.cpp
abhinav92003: Removed.

clients/drcachesim/tools/trace_filter.cpp
abhinav92003: Converted analyzer_t to a template in https://github.com/DynamoRIO/dynamorio/pull/5720.

clients/drcachesim/tools/trace_filter.cpp
abhinav92003: Converted analyzer_t to a template in https://github.com/DynamoRIO/dynamorio/pull/5720.

clients/drcachesim/tools/trace_filter.cpp
abhinav92003: Converted analyzer_t to a template in https://github.com/DynamoRIO/dynamorio/pull/5720.

clients/drcachesim/tools/trace_filter.cpp
abhinav92003: Added a comment.

clients/drcachesim/tests/trace_filter_unit_tests.cpp
abhinav92003: Done.

clients/drcachesim/tools/basic_counts.cpp
abhinav92003: Added a comment and to the PR description.

abhinav92003: https://github.com/DynamoRIO/dynamorio/pull/5720.

clients/drcachesim/reader/compressed_file_reader.cpp
abhinav92003: Done.

clients/drcachesim/CMakeLists.txt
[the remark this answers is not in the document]
abhinav92003: Converted analyzer_t to a template in https://github.com/DynamoRIO/dynamorio/pull/5720.

abhinav92003:
Closing this in favour of #5720. I wanted to compare the two code side-by-side so started a new PR.

PR #5698, opened 2022-10-24
i#5505 kernel trace: Add a new module kcore_copy to dump kcore

dolanzhao:
Adds a new module kcore_copy, and updates drpt2ir and drpt2trace.

(1) Adds a new module kcore_copy to dump kcore's code segments.
This module can read '/proc/modules' and '/proc/kallsyms' to get all module's information. Then it bases on the module's information to locate all code segments in '/proc/kcore', and dump all bytes and metadata of code segments to a ELF format file.

(2) Updates drpt2trace to support two working modes: ELF Mode and SIDEBAND Mode.
Adds a new option "-mode" to drpt2trace to let the user choose the working mode of drpt2trace. 
In the current implementation, drpt2trace supports two modes to convert:

* ELF (-mode ELF): the user provides an ELF file that contains all the instruction bytes. So, for example, we can use this mode to convert the kernel trace and the short-term user trace, where it's likely that we'll not have an image switch. 

* SIDEBAND (-mode SIDEBAND): the user must provide sideband data and parameters. In this mode, the converter uses sideband decoders to simulate image switches during the conversion. For example, we can use this mode to convert the traces where the instruction bytes are located in multiple images.

(3)  Updates the test to check if dr$sim's kernel PT output is correct.

Issue: #5505


derekbruening:
I guess my high-level question is one I've asked before: if we're only supporting kernel code from ELF code sections anyway, we should be able to obtain the ELF files offline and not have to do any reading of kcore (which requires privileges and adds overhead) during tracing.  Certainly for stock kernels in Linux distros we can get the files.  Custom kernels and custom drivers complicate things I suppose: is that the reason to push for kcore?

abhinav92003:
> Custom kernels and custom drivers complicate things I suppose: is that the reason to push for kcore?

Additionally, I think it'd be better to have a "complete" kernel trace decoding support in drmemtrace. If the user can obtain the ELF files offline, then they can skip the online kcore dump (we don't have a flag for that yet, but can add it). We have not yet attempted using offline ELF files yet; we can add a dox section on that when we do.

clients/drcachesim/tracer/kimage.h
```diff
@@ -0,0 +1,68 @@
...
+
+/* kimage.h: kimage is a file that contains all kernel code segments. It is used to decode
+ * kernel PT trace. The format of the file is:
+ *  ------------------------------------------------
+ * | An instance of kimage_hdr_t                    |
+ *  ------------------------------------------------
+ * | A table of kimage_code_segment_hdr_t instances |
+ *  ------------------------------------------------
+ * | Binary data for all code segments              |
+ *  ------------------------------------------------
```
derekbruening, line 41: I think using ELF format wouldn't be much more complex in the code (it would simplify it in some ways since this header would disappear) and might be easier to debug since we could use standard ELF tools to examine the binary file.  You already have this format which seems ok but if you hadn't yet written it I would have voted for just using the standard ELF format.

clients/drcachesim/tracer/kcore_copy.cpp
```diff
@@ -0,0 +1,433 @@
...
+        uint64_t addr;
+        if (dr_sscanf(line.c_str(), HEX64_FORMAT_STRING " %*1c %299s [%*99s", &addr,
+                      name) < 2)
+            continue;
+        if (strcmp(name, "_stext") == 0) {
+            if (kernel_module != nullptr) {
+                f.close();
+                return false;
+            }
+            kernel_module = (proc_module_t *)dr_global_alloc(sizeof(proc_module_t));
```
derekbruening, line 305: assert that kernel_module is nullptr

clients/drcachesim/tracer/kcore_copy.cpp
```diff
@@ -0,0 +1,433 @@
...
+            kernel_module->start = addr;
+        } else if (strcmp(name, "_etext") == 0) {
+            if (kernel_module == nullptr) {
+                f.close();
+                return false;
+            }
+            kernel_module->end = addr;
```
derekbruening, line 312: assert that kernel_module is not nullptr

clients/drcachesim/tracer/kcore_copy.cpp
```diff
@@ -0,0 +1,433 @@
...
+            kernel_module->next = modules_;
+            kcore_code_segments_num_++;
+            modules_ = kernel_module;
+            kernel_module = nullptr;
+        }
+    }
+    f.close();
```
derekbruening, line 319: assert that kernel_module is nullptr

clients/drcachesim/tracer/kcore_copy.cpp
```diff
@@ -0,0 +1,433 @@
...
+    uint64_t start;
+
+    /* The length of the code segment. */
+    ssize_t len;
+
+    /* The virtual address of the code segment in memory. */
+    uint64_t vaddr;
+
+    /* The buffer to store the code segment. */
+    char *buf;
```
derekbruening, line 79: Do we need this intermediate copy?  How much data are we talking about?  If it is a lot of data it would be better to just record the offsets and only read it once, in pieces, when we write it out to the output file.

derekbruening:
> When using a PT decoder to decode PT, reading image segments from this format file is faster than reading from an elf file.

I'm skeptical of this: this new file looks pretty similar to an ELF file.  I would expect the speed to be dominated by decoding and whether you have to do a few more header hierarchy steps in reading to not be measurable, so I do not see a meaningful speed advantage with a custom format.  See my review comment, where I would in fact vote for using ELF rather than a custom format if you hadn't already written the code w/ the custom format, b/c it would be not really more complex and not any slower but would be a standard format that is easier to debug and deal with.

clients/drcachesim/tracer/kimage.h
derekbruening: If we output in ELF format would the need for the -image_format and separate handling for this custom format all disappear and everything would be handled by the existing ELF file parsing?  If so, I would vote for that and not using a custom format as I don't see it adding any significant advantage and in fact it is adding complexity by adding this separate type to handle.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -275,31 +293,41 @@ option_init(int argc, const char *argv[])
...
+        /* Check if the required options for ELF mode are specified. */
+        if (!op_image_format.specified() || !op_image.specified()) {
+            std::cerr << CLIENT_NAME << ": option " << op_image_format.get_name()
+                      << " and " << op_image.get_name() << " are required in "
+                      << op_mode.get_value() << " mode." << std::endl;
+            print_usage();
+            return false;
+        }
+    } else if (op_mode.get_value() == "SIDEBAND") {
+        /* Check if the required options for SIDEBAND mode are specified. */
```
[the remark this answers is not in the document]
dolanzhao: drpt2trace fully supports sideband mode. But our online tracer doesn't output sideband data.

dolanzhao:
> > When using a PT decoder to decode PT, reading image segments from this format file is faster than reading from an elf file.
> 
> I'm skeptical of this: this new file looks pretty similar to an ELF file. I would expect the speed to be dominated by decoding and whether you have to do a few more header hierarchy steps in reading to not be measurable, so I do not see a meaningful speed advantage with a custom format. See my review comment, where I would in fact vote for using ELF rather than a custom format if you hadn't already written the code w/ the custom format, b/c it would be not really more complex and not any slower but would be a standard format that is easier to debug and deal with.

dolanzhao:
> > When using a PT decoder to decode PT, reading image segments from this format file is faster than reading from an elf file.
> 
> I'm skeptical of this: this new file looks pretty similar to an ELF file. I would expect the speed to be dominated by decoding and whether you have to do a few more header hierarchy steps in reading to not be measurable, so I do not see a meaningful speed advantage with a custom format. See my review comment, where I would in fact vote for using ELF rather than a custom format if you hadn't already written the code w/ the custom format, b/c it would be not really more complex and not any slower but would be a standard format that is easier to debug and deal with.

OK. I will dump the kernel code segments into an ELF format file.

derekbruening:
> OK. I will dump the kernel code segments into an ELF format file.

That eliminates the separate image config parameter, right?  So overall it makes things simpler?

dolanzhao:
> > OK. I will dump the kernel code segments into an ELF format file.
> 
> That eliminates the separate image config parameter, right? So overall it makes things simpler?

Yes.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -67,29 +67,42 @@ static droption_t<bool> op_help(DROPTION_SCOPE_FRONTEND, "help", false,
...
+    DROPTION_SCOPE_FRONTEND, "image_format", "", "[Optional] The image file's format",
+    "This is a required option in SINGLE_IMAGE Mode. Specifies the format of the image "
+    "file. This tool support two formats: ELF and KIMAGE.");
+static droption_t<unsigned long long> op_image_base(
+    DROPTION_SCOPE_FRONTEND, "image_base", 0x0,
+    "[Optional] The runtime load address of the elf file",
+    "This is an optional option in SINGLE_IMAGE Mode. Specifies the runtime load address "
+    "of "
+    "the elf file. For kernel cases, this always should be 0x0, so it is not required. "
+    "But if -elf specified file's runtime load address is not 0x0, it must be set.");
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -275,31 +293,41 @@ option_init(int argc, const char *argv[])
      * needs the instruction bytes for each IPs (Instruction Pointers) to decode the PT
      * trace.
      * drpt2trace supports two modes to convert:
-     * (1) Sideband Mode: the user must provide sideband data and parameters. In this
+     * (1) SINGLE_IMAGE Mode: the user provides an image file that contains all the
+     * instruction bytes. So, for example, we can use this mode to convert the kernel
+     * trace and the short-term user trace, where it's likely that we'll not have an image
+     * switch. Also, this mode support two types of image format: ELF and KIMAGE. For
+     * KIMAGE,  it is a format generated by drmemtrace and only contains the bytes and
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -67,29 +67,42 @@ static droption_t<bool> op_help(DROPTION_SCOPE_FRONTEND, "help", false,
...
     "generated by the client drcachesim.");
+static droption_t<std::string> op_image(
+    DROPTION_SCOPE_FRONTEND, "image", "", "[Optional] Path to the image file",
+    "This is a required option in SINGLE_IMAGE Mode. Specifies the file path of the "
+    "image file. This must be specified when converting traces that don't contain "
+    "sideband information. e.g. kernel-only traces and short user traces.");
+static droption_t<std::string> op_image_format(
+    DROPTION_SCOPE_FRONTEND, "image_format", "", "[Optional] The image file's format",
+    "This is a required option in SINGLE_IMAGE Mode. Specifies the format of the image "
+    "file. This tool support two formats: ELF and KIMAGE.");
```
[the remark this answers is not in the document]
dolanzhao: Removed. Currently, kcore_copy dumps all kernel code segments to an ELF format kcore.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -275,31 +293,41 @@ option_init(int argc, const char *argv[])
...
-     * for cases where the PT trace's instruction bytes belong to one image. So we can use
-     * this mode to convert the kernel trace and the short-term user trace where it's
-     * likely that we'll not have an image switch.
      */
-    if ((!op_elf.specified() && !op_primary_sb.specified()) ||
-        (op_elf.specified() && op_primary_sb.specified())) {
-        std::cerr << CLIENT_NAME
-                  << ": exactly one of the options --elf, --primary_sb must be specified."
+    if (op_mode.get_value() == "SINGLE_IMAGE") {
+        /* Check if the required options for ELF mode are specified. */
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -327,8 +355,13 @@ main(int argc, const char *argv[])
         config.init_with_metadata(op_raw_pt_metadata.get_value());
     }
     config.raw_file_path = op_raw_pt.get_value();
-    config.elf_file_path = op_elf.get_value();
-    config.elf_base = op_elf_base.get_value();
+    if (op_image_format.get_value() == "ELF") {
+        config.image_format = PT_IMAGE_FORMAT_ELF;
+    } else if (op_image_format.get_value() == "KIMAGE") {
+        config.image_format = PT_IMAGE_FORMAT_KIMAGE;
+    }
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/drpt2trace/pt2ir.cpp
```diff
@@ -394,6 +403,48 @@ pt2ir_t::load_kernel_image(IN std::string &path)
...
+        f.close();
+        return false;
+    }
+    int code_segments_num = kimage_hdr.code_segments_num;
+    for (int i = 0; i < code_segments_num; i++) {
+        kimage_code_segment_hdr_t code_segment_hdr;
+        f.read(reinterpret_cast<char *>(&code_segment_hdr),
+               sizeof(kimage_code_segment_hdr_t));
+        if (f.fail()) {
+            ERRMSG("Failed to read kimage code segment header form kimage file: %s.\n",
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/drpt2trace/pt2ir.h
```diff
@@ -375,7 +390,12 @@ class pt2ir_t {
      * information.
      */
     bool
-    load_kernel_image(IN std::string &path);
+    load_kcore(IN std::string &path);
+
+    /* Load the all code segments in kernel image to pt_insn_decoder's image cache.
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/kcore_copy.cpp
dolanzhao: Done.

clients/drcachesim/tracer/kcore_copy.cpp
dolanzhao: Done.

clients/drcachesim/tracer/kcore_copy.cpp
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
```diff
@@ -275,31 +293,41 @@ option_init(int argc, const char *argv[])
      * needs the instruction bytes for each IPs (Instruction Pointers) to decode the PT
      * trace.
      * drpt2trace supports two modes to convert:
-     * (1) Sideband Mode: the user must provide sideband data and parameters. In this
+     * (1) SINGLE_IMAGE Mode: the user provides an image file that contains all the
+     * instruction bytes. So, for example, we can use this mode to convert the kernel
+     * trace and the short-term user trace, where it's likely that we'll not have an image
```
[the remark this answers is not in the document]
dolanzhao: Currently, we only support ELF format images. And for short-term user trace, if the application loads the single elf image with an offset, they can get the offset value by trace tools and set the -elf_base to the value.

clients/drcachesim/tracer/kimage.h
```diff
@@ -0,0 +1,68 @@
...
+#ifndef _KIMAGE_H_
+#define _KIMAGE_H_ 1
+
+#include <inttypes.h>
+
+/* This struct type defines the file header that stores all kernel code segments. */
+struct kimage_hdr_t {
+    /* The number of code segments in kimage. */
+    uint64_t code_segments_num;
+    /* The offset of the first code segment in . */
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/kcore_copy.h
```diff
@@ -0,0 +1,100 @@
...
+    bool
+    read_kcore();
+
+    /* The module list. */
+    proc_module_t *modules_;
+
+    /* The number of kernel code segments. */
+    int kcore_code_segments_num_;
+
+    /* The kernel code segments list. */
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/kcore_copy.cpp
```diff
@@ -0,0 +1,433 @@
...
+        dr_close_file(from_kallsyms_fd);
+        return false;
+    }
+
+    char buf[1024];
+    ssize_t bytes_read;
+    while ((bytes_read = dr_read_file(from_kallsyms_fd, buf, sizeof(buf))) > 0) {
+        if (dr_write_file(to_kallsyms_fd, buf, bytes_read) != bytes_read) {
+            ASSERT(false, "failed to write kallsyms file");
+            dr_close_file(to_kallsyms_fd);
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/kcore_copy.cpp
```diff
@@ -0,0 +1,433 @@
...
+        return false;
+    }
+
+    Elf64_Ehdr ehdr;
+    if (dr_read_file(fd, &ehdr, sizeof(ehdr)) != sizeof(ehdr)) {
+        ASSERT(false, "failed to read the ehdr of " KCORE_FILE_PATH);
+        dr_close_file(fd);
+        return false;
+    }
+    if (LONG_MAX < ehdr.e_phoff) {
```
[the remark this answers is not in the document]
dolanzhao: Removed.

clients/drcachesim/tracer/kcore_copy.cpp
```diff
@@ -0,0 +1,433 @@
...
+        if (dr_read_file(fd, &phdr, sizeof(phdr)) != sizeof(phdr)) {
+            ASSERT(false, "failed to read the Phdr of " KCORE_FILE_PATH);
+            dr_close_file(fd);
+            return false;
+        }
+
+        if (phdr.p_type != PT_LOAD || phdr.p_filesz == 0)
+            continue;
+
+        proc_module_t *module = modules_;
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/kcore_copy.cpp
```diff
@@ -0,0 +1,433 @@
...
+        if (module == nullptr) {
+            ASSERT(false, "no module found in " MODULES_FILE_PATH);
+            dr_close_file(fd);
+            return false;
+        }
+        while (module != nullptr) {
+            if (module->start >= phdr.p_vaddr &&
+                module->end < phdr.p_vaddr + phdr.p_filesz) {
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/drpt2trace/drpt2trace.cpp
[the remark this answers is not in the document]
derekbruening: For the integrated workflow with drmemtrace the tooling should get the offset for the user: the user should not have to go look up a value that can be automatically obtained.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1609,9 +1610,12 @@ event_exit(void)
 #ifdef BUILD_PT_TRACER
     if (op_offline.get_value() && op_enable_kernel_tracing.get_value()) {
         drpttracer_exit();
-        /* Dump kcore and kallsyms to {kernel_pt_logsubdir}. */
-        if (!syscall_pt_trace_t::kernel_image_dump(kernel_pt_logsubdir)) {
-            NOTIFY(0, "WARNING: failed to run shellscript to dump kernel image\n");
+        /* Copy kcore and kallsyms to {kernel_pt_logsubdir}. */
+        std::unique_ptr<kcore_copy_t> kcore_copy(new kcore_copy_t(
```
derekbruening, line 1614: Looks like there is no reason to put this on the heap: may as well just have a local stack var with no need for `new` or a pointer at all, right?

clients/drcachesim/tracer/kcore_copy.cpp
```diff
@@ -79,8 +76,93 @@ struct proc_kcore_code_segment_t {
     char *buf;
 };
 
-kcore_copy_t::kcore_copy_t()
-    : modules_(nullptr)
+/* The auto close wrapper of file_t.
+ * This can ensure the file is closed when it is out of scope. And it is also allowed to
+ * cutomize file manipulation functions.
```
derekbruening, line 81: spelling: customize

clients/drcachesim/tracer/kcore_copy.cpp
```diff
@@ -79,8 +76,93 @@ struct proc_kcore_code_segment_t {
...
+ */
+struct file_autoclose_t {
```
derekbruening, line 83: style: Seems like this should be a `class` as it has several >1-line functions

clients/drcachesim/tracer/kcore_copy.cpp
```diff
@@ -79,8 +76,93 @@ struct proc_kcore_code_segment_t {
...
+    file_t fd;
```
derekbruening, line 84: As a `class` these member fields should have end in `_`

clients/drcachesim/tracer/kcore_copy.cpp
```diff
@@ -79,8 +76,93 @@ struct proc_kcore_code_segment_t {
...
+    }
+
+    bool
+    is_open()
+    {
+        return fd != INVALID_FILE;
+    }
+
+    bool
+    write(void *buf, size_t count)
```
derekbruening, line 131: `const void *buf`

clients/drcachesim/tracer/kcore_copy.cpp
```diff
@@ -79,8 +76,93 @@ struct proc_kcore_code_segment_t {
...
+        , close_file_func(close_file)
+        , read_file_func(read_file)
+        , write_file_func(write_file)
+        , seek_file_func(seek_file)
+    {
+        ASSERT(open_file_func != NULL, "open_file_func cannot be NULL");
+        ASSERT(close_file_func != NULL, "close_file_func cannot be NULL");
+
+        ASSERT(file_name != NULL, "file_name cannot be NULL");
+        dr_snprintf(file_name_buffer, BUFFER_SIZE_ELEMENTS(file_name_buffer), "%s",
```
derekbruening, line 109: As this is C++ code, a reader would expect a std::string here.  This is avoiding STL data structures b/c it's used in DR client code statically linked into the app, right?  Worth adding a comment about this.

derekbruening (approved):

clients/drcachesim/tracer/kcore_copy.cpp
dolanzhao: Done.

clients/drcachesim/tracer/kcore_copy.cpp
dolanzhao: Done.

clients/drcachesim/tracer/kcore_copy.cpp
dolanzhao: Done.

clients/drcachesim/tracer/kcore_copy.cpp
dolanzhao: Done.

clients/drcachesim/tracer/kcore_copy.cpp
dolanzhao: Done.

clients/drcachesim/tracer/tracer.cpp
dolanzhao: Done.

abhinav92003 (approved):

clients/drcachesim/tracer/kcore_copy.cpp
dolanzhao: I will add an option on the next PR to let the user specify whether to dump kernel binary.

PR #5702, opened 2022-10-25
i#3044: Add release note for new element vectors

joshua-warburton:
issues: #3044

### DIFF
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -176,9 +176,13 @@ Further non-compatibility-affecting changes include:
    the .zip, which sets the granularity of a fast seek.
  - Added dr_register_post_attach_event(), dr_unregister_post_attach_event(),
    dr_register_pre_detach_event(), and dr_unregister_pre_detach_event().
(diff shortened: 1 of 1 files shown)


api/docs/release.dox
```diff
@@ -176,8 +176,12 @@ Further non-compatibility-affecting changes include:
    the .zip, which sets the granularity of a fast seek.
  - Added dr_register_post_attach_event(), dr_unregister_post_attach_event(),
    dr_register_pre_detach_event(), and dr_unregister_pre_detach_event().
- - Added insruction encodings to drmemtrace offline traces.
+ - Added instruction encodings to drmemtrace offline traces.
  - Added drmemtrace_replace_file_ops_ex().
+ - Added a method of representing vector registers with element sizes.
+   They can be created with #opnd_create_reg_element_vector, detected with
```
derekbruening, line 182: Since all of these are functions, empty parens will also trigger a doxygen link and make it clear they're functions (see the examples above): so `opnd_create_reg_element_vector()`.  Ditto below.

derekbruening (approved):

PR #5704, opened 2022-10-25
i#2039 trace trim, part 4: Clamp timestamps

derekbruening:
Moves the timestamp+cpu headers to be added at buffer creation time instead of at buffer output time.

Makes instru_t::frozen_timestamp_ use release-acquire semantics and sets it for -align_endpoints at detach time.

Adds a 'minimum timestamp' recorded at post-attach time.  At buffer output time, we examine the buffer's timestamp and if it was from before the minimum, we replace it with the current (not the minimum) timestamp.

Tested on a large app where these changes result in much tighter timestamps matching the burst period.

Issue: #2039


clients/drcachesim/tracer/output.cpp
```diff
@@ -822,7 +819,9 @@ process_entry_for_physaddr(void *drcontext, per_thread_t *data, size_t header_si
         data->num_phys_markers +=
             output_buffer(drcontext, data, data->v2p_buf, v2p_ptr, header_size);
         v2p_ptr = data->v2p_buf;
-        v2p_ptr += add_buffer_header(drcontext, data, v2p_ptr);
```
abhinav92003, line 825: Remind me why we add a header the v2p buf too?

clients/drcachesim/tracer/output.cpp
```diff
@@ -916,22 +914,51 @@ process_and_output_buffer(void *drcontext, bool skip_size_cap)
...
+        // This is the attach counterpart to instru_t::set_frozen_timestamp(): we place
+        // timestamps at buffer creation, but that can be before we're fully attached.
+        // We update any too-early timestamps to reflect when we actually started
+        // tracing.  (Switching back to timestamps at buffer output is actually
+        // worse as we then have the identical frozen timestamp for all the flushes
+        // during detach, plus they are all on the same cpu too.)
+        size_t stamp_offs =
+            header_size > buf_hdr_slots_size ? header_size - buf_hdr_slots_size : 0;
+        uint64 min_timestamp = attached_timestamp.load(std::memory_order_acquire);
+        bool found = instru->refresh_unit_header_timestamp(data->buf_base + stamp_offs,
```
abhinav92003, line 934: In normal operation, we don't expect the stamp_offs to NOT be the timestamp entry. So maybe it's better to assert found inside refresh_unit_header_timestamp.

clients/drcachesim/tracer/output.cpp
```diff
@@ -445,15 +445,21 @@ open_new_thread_file(void *drcontext, ptr_int_t window_num)
...
+prepend_offline_thread_header(void *drcontext, ptr_int_t window = -1)
 {
     DR_ASSERT(op_offline.get_value());
     /* Write initial headers at the top of the first buffer. */
     per_thread_t *data = (per_thread_t *)drmgr_get_tls_field(drcontext, tls_idx);
     size_t size = reinterpret_cast<offline_instru_t *>(instru)->append_thread_header(
         data->buf_base, dr_get_thread_id(drcontext), get_file_type());
-    BUF_PTR(data->seg_base) = data->buf_base + size + buf_hdr_slots_size;
+    BUF_PTR(data->seg_base) = data->buf_base + size;
+    BUF_PTR(data->seg_base) += append_unit_header(
```
abhinav92003, line 460: Why do we need to add this here? If we don't add it here, then we won't need the if(!prepended) condition checks below.

abhinav92003 (approved):
For some reason I found this PR to be a bit tricky to review :)

clients/drcachesim/tracer/instru.h
```diff
@@ -221,10 +223,15 @@ class instru_t {
     // This is a per-buffer-writeout header.
     virtual int
     append_unit_header(byte *buf_ptr, thread_id_t tid, ptr_int_t window) = 0;
+    // If the entry at buf_ptr is not a timestamp, returns false.
+    // Else, if the timestamp value is < min_timestamp, replaces it with a
+    // new timestamp (ignoring the frozen timestamp).  Returns true either way.
```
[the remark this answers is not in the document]
derekbruening: I didn't like the idea of a bunch of threads with perfectly identical timestamps: that was my logic as it looks artificial.  Do you think we should use the minimum anyway?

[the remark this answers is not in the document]
abhinav92003: Since we are starting tracing for all threads together, it seemed fine to me if all threads had perfectly identical timestamps for their start. Also makes explaining the timestamp entry easier, instead of "it's the timestamp of the first entry for all non-first buffers"

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -427,6 +430,12 @@ event_pre_detach()
 {
     if (align_attach_detach_endpoints()) {
         NOTIFY(1, "Switching to no-tracing mode during detach\n");
+        // Keep all final thread output at the detach timestamp.  We won't add to any
+        // buffers during detach and we now have timestamps added at buffer start
+        // instead of output but a non-omitted thread that is at a wait syscall
```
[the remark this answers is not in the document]
derekbruening: No, I mean a thread that was not completely empty and so was not omitted completely.  But this description doesn't sound right: I remember originally thinking the frozen timestamp would never actually be used, but then I saw a case but I think I mis-interpreted it.  It seems like it would only happen with a new buffer started at exit which would only happen with windows where we're now on a new window and so we make a new buffer for the thread exit.  Will update the comment.

clients/drcachesim/tracer/instru_offline.cpp
```diff
@@ -397,15 +397,26 @@ offline_instru_t::append_unit_header(byte *buf_ptr, thread_id_t tid, ptr_int_t w
...
-    entry->timestamp.usec =
-        frozen_timestamp_ != 0 ? frozen_timestamp_ : instru_t::get_timestamp();
+    uint64 frozen = frozen_timestamp_.load(std::memory_order_acquire);
+    entry->timestamp.usec = frozen != 0 ? frozen : instru_t::get_timestamp();
     new_buf += sizeof(*entry);
     if (window >= 0)
         new_buf += append_marker(new_buf, TRACE_MARKER_TYPE_WINDOW_ID, (uintptr_t)window);
     new_buf += append_marker(new_buf, TRACE_MARKER_TYPE_CPU_ID, instru_t::get_cpu_id());
     return (int)(new_buf - buf_ptr);
 }
```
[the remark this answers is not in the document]
derekbruening: +1.  Added a comment in the issue.

clients/drcachesim/tracer/instru.h
```diff
@@ -38,6 +38,7 @@
```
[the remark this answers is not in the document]
derekbruening: Ah yes I had forgotten about that issue: will mark it as fixed.

clients/drcachesim/tracer/output.cpp
derekbruening: IIRC, once we decided to make it a separate buffer, it is emitted before the regular one so we put a header on it.  With the two separate write-outs potentially not being atomic it seems best to have each buffer have its own header.

clients/drcachesim/tracer/output.cpp
derekbruening: My logic was treating it like a library routine but it is pretty integrated so having this a pre-req and assert seems reasonable.  Done.

clients/drcachesim/tracer/output.cpp
derekbruening: This append needs to be done at some point for every call to this function prepend_offline_thread_header() which is why it is logically bundled here.  If it is moved from here it needs to be added to every call site.  Maybe that is simpler than the conditional: there are only two call sites.  The risk is a future new call that forgets to also add the unit header but maybe that is still worthwhile.  OK, will do.

derekbruening: Actually I end up still needing the prepended return value and check as I don't know whether to add the header to the current pointer or the base here.  Here we could set the pointer to the base before calling set_local_window; up above on an empty buffer though we still need the return value as we again don't know where to write the unit header.  It's less clear whether this is better than before but it's already in there so I guess we'll go with it, after some testing.

PR #5710, opened 2022-10-28
i#5538 memtrace seek, part 4: Check chunk boundaries

derekbruening:
Adds an invariant check that chunk boundaries contain the proper number of instructions.

Fixes off-by-one errors in chunk counts due to not checking at end-of-loop, found by visual inspection (with a local view tool that prints instr counts; that will be committed later) and confirmed to break this new check without the fix in the
tool.drcacheoff.invariant_checker test.

Issue: #5538

### DIFF
--- a/clients/drcachesim/tools/invariant_checker.cpp
+++ b/clients/drcachesim/tools/invariant_checker.cpp
@@ -194,6 +194,8 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
         shard->found_page_size_marker_ = true;
     }
 
+    // Invariant: each chunk's instruction count must be identical and equal to
+    // the value in the top-level marker.
     if (memref.marker.type == TRACE_TYPE_MARKER &&
         memref.marker.marker_type == TRACE_MARKER_TYPE_CHUNK_INSTR_COUNT) {
         shard->chunk_instr_count_ = memref.marker.marker_value;
--- a/clients/drcachesim/tools/invariant_checker.cpp
+++ b/clients/drcachesim/tools/invariant_checker.cpp
@@ -102,6 +102,9 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
     ++shard->ref_count;
     if (shard->tid == -1 && memref.data.tid != 0)
         shard->tid = memref.data.tid;
+    // XXX i#5538: Have the infrastructure provide a common instr and record count.
+    if (type_is_instr(memref.instr.type))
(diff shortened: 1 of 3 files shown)


derekbruening:
run arm tests

derekbruening:
The Jenkins failure is my new check:
```
Trace invariant failure in T4010479 at ref # 93263: Chunk instruction counts are inconsistent
```
Yet that raw dir is gone and I can't reproduce this on the Jenkins machine: I ran the test 1000 times w/o repro; running again 10,000 times now.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -191,6 +194,18 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
         shard->found_page_size_marker_ = true;
     }
 
+    if (memref.marker.type == TRACE_TYPE_MARKER &&
+        memref.marker.marker_type == TRACE_MARKER_TYPE_CHUNK_INSTR_COUNT) {
+        shard->chunk_instr_count_ = memref.marker.marker_value;
+    }
+    if (memref.marker.type == TRACE_TYPE_MARKER &&
```
bete0, line 201: nit: should we add a comment here about this invariant check?

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -1225,6 +1231,14 @@ raw2trace_t::write(void *tls, const trace_entry_t *start, const trace_entry_t *e
                                 reinterpret_cast<const char *>(end) -
                                     reinterpret_cast<const char *>(start)))
         return "Failed to write to output file";
+    // If we're at the end of a block (minus its delayed branch) we need
+    // to split now to avoid going too far by waiting for the next instr.
+    if (tdata->cur_chunk_instr_count >= chunk_instr_count_) {
```
bete0, line 1236: Do we need this if statement if we are asserting that they are `==`? Or is `DEBUG_ASSERT` lets you proceed even when they are not equal?

bete0 (approved):

clients/drcachesim/tools/invariant_checker.cpp
derekbruening: Done.

clients/drcachesim/tracer/raw2trace.cpp
derekbruening: DEBUG_ASSERT is a nop in release build.  The > part is being defensive.

PR #5713, opened 2022-10-31
i#5538 memtrace seek, part 5: Add schedule index files

derekbruening:
Adds two new files generated by raw2trace, each containing 4-entry records of <tid, timestamp, cpuid, instr_count>.  These are used to take a global or per-cpu target instruction count and determine the corresponding instruction count within each software thread, for seeking within each trace file.  One file is globally sorted by timestamp, and the other is a zip file with a separate component for each cpuid holding entries sorted by timestamp for that cpu.

Adds new invariant_checker tests for each file by passing them in for test_mode.  The invariant_checker re-constructs the same sorted sequences using the trace data and confirms it matches the data in the files.  This involves adding a new zipfile_stream_t interface for a simple continuous stream of each zipfile component in turn.

Issue: #5538

### DIFF
--- a/clients/drcachesim/analyzer.cpp
+++ b/clients/drcachesim/analyzer.cpp
@@ -65,17 +65,6 @@ analyzer_t::analyzer_t()
     /* Nothing else: child class needs to initialize. */
 }
 
-#if defined(HAS_SNAPPY) || defined(HAS_ZIP)
-static bool
-ends_with(const std::string &str, const std::string &with)
-{
-    size_t pos = str.rfind(with);
-    if (pos == std::string::npos)
-        return false;
-    return (pos + with.size() == str.size());
-}
-#endif
-
 static std::unique_ptr<reader_t>
 get_reader(const std::string &path, int verbosity)
 {
@@ -98,6 +87,11 @@ get_reader(const std::string &path, int verbosity)
             return nullptr;
         }
         for (; iter != end; ++iter) {
+            const std::string fname = *iter;
+            if (fname == "." || fname == ".." ||
+                starts_with(fname, DRMEMTRACE_SERIAL_SCHEDULE_FILENAME) ||
+                fname == DRMEMTRACE_CPU_SCHEDULE_FILENAME)
+                continue;
 #    ifdef HAS_SNAPPY
             if (ends_with(*iter, ".sz")) {
                 return std::unique_ptr<reader_t>(
@@ -141,7 +135,9 @@ analyzer_t::init_file_reader(const std::string &trace_path, int verbosity)
         }
         for (; iter != end; ++iter) {
             const std::string fname = *iter;
-            if (fname == "." || fname == "..")
+            if (fname == "." || fname == ".." ||
+                starts_with(fname, DRMEMTRACE_SERIAL_SCHEDULE_FILENAME) ||
+                fname == DRMEMTRACE_CPU_SCHEDULE_FILENAME)
                 continue;
             const std::string path = trace_path + DIRSEP + fname;
             std::unique_ptr<reader_t> reader = get_reader(path, verbosity);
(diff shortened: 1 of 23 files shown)


derekbruening:
x64 failure is flaky api.detach_state assert about cache consistency #5123

clients/drcachesim/analyzer.cpp
```diff
@@ -98,6 +87,11 @@ get_reader(const std::string &path, int verbosity)
             return nullptr;
         }
         for (; iter != end; ++iter) {
+            const std::string fname = *iter;
+            if (fname == "." || fname == ".." ||
```
bete0, line 91: Can't we use `starts_with(fname, ".")` and `starts_with(fname, "..")` here?

clients/drcachesim/analyzer_multi.cpp
```diff
@@ -173,9 +185,48 @@ analyzer_multi_t::create_analysis_tools()
     }
     num_tools_ = 1;
     if (op_test_mode.get_value()) {
-        tools_[1] =
-            new invariant_checker_t(op_offline.get_value(), op_verbose.get_value(),
-                                    op_test_mode_name.get_value());
+        if (op_offline.get_value()) {
+            // TODO i#5538: Locate and open the schedule files and pass to the
+            // reader(s) for seeking.  For now we only read them for this test.
```
bete0, line 190: nit: is that extra space at "seeking.  For now..."

clients/drcachesim/common/zipfile_istream.h
```diff
@@ -0,0 +1,128 @@
...
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+/* zipfile_istream_t: provides a std::istream interface for reading components
+ * of a zipfile.  It provides a continuous stream that cycles through all
+ * components.
+ * Supports only limited seeking within the current internal buffer.
```
bete0, line 36: nit: Could this have been in the same line 35?

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -776,17 +776,63 @@ raw2trace_t::do_conversion()
...
+    std::sort(serial.begin(), serial.end(),
+              [](const schedule_entry_t &l, const schedule_entry_t &r) {
+                  return l.timestamp < r.timestamp;
+              });
+    if (serial_schedule_file_ != nullptr) {
+        if (!serial_schedule_file_->write(reinterpret_cast<const char *>(serial.data()),
+                                          serial.size() * sizeof(serial[0])))
+            return "Failed to write to serial schedule file";
+    }
+    if (cpu_schedule_file_ == nullptr)
```
bete0, line 816: Shouldn't this check be at the top of this method perhaps by changing the `&&` to `||` on line 796?

clients/drcachesim/analyzer.cpp
derekbruening: We don't want to match names that start with dot though, like ".bashrc": this is just skipping the special `.` and `..` entries for the current and parent dir, so we want to match the entire string.

clients/drcachesim/analyzer_multi.cpp
derekbruening: Done.  (Though I think a lot of the DR code base uses two spaces between sentences we can tighten it up here as it is not an official style rule I don't think.)

clients/drcachesim/common/zipfile_istream.h
derekbruening: Done.

clients/drcachesim/tracer/raw2trace.cpp
derekbruening: Since both are optional the top check only returns immediately of both are empty.  We then have two optional pieces, one for the serial file above this and one for the cpu file which is guarded by this check.

bete0 (approved):

PR #5721, opened 2022-11-08
i#5538 memtrace seek, part 7: Add instr count to tools

derekbruening:
Adds a new memref_stream_t interface class which provides the record and instruction count to drmemtrace analysis tools.  A pointer to this interface is passed to new extended-argument versions of analysis_tool_t's initialize() and parallel_shard_init() functions, which are now what is called by the analyzer.  The base class implementation of these new functions simply calls the old versions, which are now deprecated but will continue to work.

This new interface is not just for convenience: the tool itself cannot accurately count when the reader skips over records, as will happen with seeking.  The counting must be done in the reader.  (If the tool indeed wants to count only records/instrs that it actually sees, it can continue using its own counters.)

Updates the view tool to use the new interface to obtain the record ordinal, replacing its own counter.  The tool is expanded to print a new column with the instruction ordinal.  The view tool test is updated along with example output in the docs.

Issue: #5538


clients/drcachesim/analysis_tool.h
```diff
@@ -168,20 +183,32 @@ class analysis_tool_t {
...
+     * table if desired (typically for aggregation use in print_results()).  The \p
+     * worker_data is the return value of parallel_worker_init() for the worker thread
+     * who will exclusively operate on this shard.  The \p shard_stream allows tools to
+     * query details of the underlying trace shard during parallel operation; it is
+     * valid only until parallel_shard_exit() is called.  The return value here will be
+     * passed to each invocation of parallel_shard_memref() for that same shard.
+     */
+    virtual void *
+    parallel_shard_init_stream(int shard_index, void *worker_data,
+                               memref_stream_t *shard_stream)
```
abhinav92003, line 208: Xref #5720: the shard_stream interface can probably be used to get the shard_suffix (introduced in #5720) too.

clients/drcachesim/common/memref_stream.h
```diff
@@ -0,0 +1,69 @@
...
+ * @file drmemtrace/memref_stream.h
+ * @brief DrMemtrace interface for obtaining information from analysis
+ * tools on the full stream of memory reference records.
+ */
+
+/**
+ * This is an interface for obtaining information from analysis tools
+ * on the full stream of memory reference records.
+ */
+class memref_stream_t {
```
abhinav92003, line 49: Xref #5720: this is another class that would need to be template-ized. Can we name it something more generic? Or maybe we can just live with "memref" being in the name of the interface everywhere even for the trace_entry_t versions of the analyzer.

abhinav92003 (approved):

clients/drcachesim/analysis_tool.h
derekbruening: I should add to the PR description the other options we discussed in the past: adding more fields to memref_t (but have to append and then these common fields are at different offsets in each type; or prepend and break compat); adding fields to process_memref which seems more intrusive than expanding these init routines.  Having an interface seems most flexible for future features.

clients/drcachesim/common/memref_stream.h
derekbruening: How about "memtrace_stream_t"?  Or "drmemtrace_stream_t".

clients/drcachesim/analysis_tool.h
abhinav92003: Exposing memref_stream_t seems the best way forward. Yes, it'd be good to document those details for posterity.

clients/drcachesim/common/memref_stream.h
derekbruening: Went with `memtrace_stream_t`.

clients/drcachesim/analysis_tool.h
derekbruening: Added notes to memtrace_stream.h and will add to the commit description too.

derekbruening:
a64 failure is windows-zlib #5493

derekbruening:
run arm tests

derekbruening:
run arm tests

derekbruening:
Hit an invariant check failure on chunk instr counts: this was hit once before on a64 and I couldn't reproduce it.  Filed #5724 now.

derekbruening:
run arm tests

PR #5729, opened 2022-11-10
Adjusting documentation

ThiagoAlexsander:
https://stackoverflow.com/questions/28350214/how-to-build-x86-and-or-x64-on-windows-from-command-line-with-cmake


ThiagoAlexsander:
@derekbruening
As requested, here it is

derekbruening (approved):
Thank you for the patch

PR #5731, opened 2022-11-11
i#5538 memtrace seek, part 8: Add skip_thread_instructions()

derekbruening:
Adds a new skip_instructions() reader iterator interface.  It is a linear walk for every type of reader except a chunked zipfile walking a single thread.

Adds a drcachesim command line option -skip_instrs which triggers the analyzer to skip from the start before passing anything to the tool.

Refactors the reader_t++ to provide a process_input_entry to update state while skipping.

Adds a unit test with an added trace file with a small chunk size. The test checks the view output for every skip value from 0 to over double the chunk size.

Leaves several pieces for future work:
+ Recording the record count in each chunk so we have an accurate count after skipping.
+ Presenting global headers skipped over as memtrace_stream_t values that tools can query.
+ Reading the schedule files for serial skipping (or the planned cpu iterator and skipping).
+ Repeating the timestamp+cpu for non-zipfile skipping.

Issue: #5538


derekbruening:
I am not understanding how my code here caused these OSX format string warnings: I thought maybe they upgraded the OSX compiler, but your PR #5711 isn't hitting these, right?  I tried to fix thinking the _APPLE__ define that's in there was incorrectly requesting long long, but the warning claims the type is long long: but it's uint64_t and size_t for the two things being printed.  Going ahead and asking for a review as maybe you can see what I'm missing, or maybe you did hit this in your PR's?

derekbruening:
> I am not understanding how my code here caused these OSX format string warnings

I removed the dr_api.h from skip_unit_tests.cpp which was the cause as it turns out: it was only the include from that file that hit these.  I'll leave the PRIx64 change as that seems an improvement.

derekbruening:
a64 run in the latest is #5493

clients/drcachesim/reader/reader.cpp
```diff
@@ -67,231 +67,251 @@ reader_t::operator++()
...
+    if (have_memref)
+        ++cur_ref_count_;
+    return have_memref;
+}
 
+reader_t &
+reader_t::skip_instructions(uint64_t instruction_count)
+{
+    // This base class has no fast seeking and must do a linear walk.
+    // We have +1 because we need to pass up the memrefs of the final skipped
```
abhinav92003, line 305: "to pass up" means that we also need to skip the memrefs of the last skipped instructions right? I was confused that it might mean "to pass up to the user".

clients/drcachesim/reader/reader.h
```diff
@@ -102,6 +102,14 @@ class reader_t : public std::iterator<std::input_iterator_tag, memref_t>,
     virtual reader_t &
     operator++();
 
+    // Skips records until "count" instruction records have been passed.
+    // This will skip top-level headers for a thread; it is up to the caller
+    // to first observe those before skipping, if needed.  For interleaved-thread
+    // iteration, top-level headers in other threads will be skipped as well
+    // (but generally speaking these are identical to the initial thread).
```
abhinav92003, line 109: Also add that the caller can use the memtrace_stream_t interface to get them later.

clients/drcachesim/reader/zipfile_file_reader.cpp
```diff
@@ -130,4 +131,100 @@ file_reader_t<zipfile_reader_t>::is_complete()
...
+    zipfile_reader_t *zipfile = &input_files_[thread_index];
+    // We assume our unzGoToNextFile loop is plenty performant and we don't need to
+    // know the chunk names to use with a single unzLocateFile.
+    uint64_t stop_count_ = cur_instr_count_ + instruction_count + 1;
+    VPRINT(this, 2,
+           "stop=%" PRIi64 " cur=%" PRIi64 " chunk=%" PRIi64 " est=%" PRIi64 "\n",
+           stop_count_, cur_instr_count_, chunk_instr_count_,
+           cur_instr_count_ +
+               (chunk_instr_count_ - (cur_instr_count_ % chunk_instr_count_)));
+    while (cur_instr_count_ +
```
abhinav92003, line 160: Would be helpful to add a comment that says that the first step is to continue looping until we're in the correct chunk.

clients/drcachesim/reader/zipfile_file_reader.cpp
```diff
@@ -130,4 +131,100 @@ file_reader_t<zipfile_reader_t>::is_complete()
...
+            }
+            return false;
+        }
+        if (unzOpenCurrentFile(zipfile->file) != UNZ_OK)
+            return false;
+        cur_instr_count_ += chunk_instr_count_ - (cur_instr_count_ % chunk_instr_count_);
+        VPRINT(this, 2, "Thread #%zd at %" PRIi64 " instrs at start of new chunk\n",
+               thread_index, cur_instr_count_);
+        VPRINT(this, 2,
+               "zip chunk stop=%" PRIi64 " cur=%" PRIi64 " chunk=%" PRIi64 " est=%" PRIi64
```
abhinav92003, line 179: Is est the instr count at the end of the current chunk? Why "est"?

clients/drcachesim/reader/zipfile_file_reader.cpp
```diff
@@ -130,4 +131,100 @@ file_reader_t<zipfile_reader_t>::is_complete()
...
+    // We need to present a timestamp+cpu so we reset this field so process_input_entry()
+    // will not skip the first pair in this new chunk.
+    last_timestamp_instr_count_ = cur_instr_count_;
+    while (cur_instr_count_ < stop_count_) {
+        if (!read_next_thread_entry(thread_index, &entry_copy_, eof))
+            return false;
+        // We need to pass up memrefs for the final skipped instr, but we don't
+        // want to process_input_entry() on the first unskipped instr so we can
+        // insert the timestamp+cpu first.
+        if (cur_instr_count_ + 1 == stop_count_ &&
```
abhinav92003, line 197: So the loop condition is never false?

clients/drcachesim/tests/skip_unit_tests.cpp
```diff
@@ -0,0 +1,151 @@
...
+        CHECK(iter->init(), "failed to initialize reader");
+        std::unique_ptr<reader_t> iter_end =
+            std::unique_ptr<reader_t>(new zipfile_file_reader_t());
+        // Run the tool.
+        std::unique_ptr<analysis_tool_t> tool =
+            std::unique_ptr<analysis_tool_t>(view_tool_create(
+                "", /*thread=*/0, /*skip_refs=*/0, /*sim_refs=*/view_count, "att"));
+        std::string error = tool->initialize_stream(iter.get());
+        CHECK(error.empty(), error.c_str());
+        iter->skip_instructions(skip_instrs);
```
abhinav92003, line 97: A user may be confused between the skip_refs option and skip_instructions(). It's clear that one is on memrefs and the other on instructions. But maybe also comment somewhere that the former is unoptimized (right now atleast).

clients/drcachesim/tests/skip_unit_tests.cpp
```diff
@@ -0,0 +1,151 @@
...
+        // Check the result.
+        std::string res = capture.str();
+        if (op_verbose.get_value())
+            std::cout << "Got: |" << res << "|\n";
+        CHECK(skip_instrs != 0 ||
+                  res.find("chunk instruction count 20") != std::string::npos,
+              "expecting chunk size of 20 in test trace");
+        std::stringstream res_stream(res);
+        std::string line;
+        std::getline(res_stream, line, '\n');
```
abhinav92003, line 111: Would be helpful to add an example of the expected output to a comment here.

clients/drcachesim/tests/skip_unit_tests.cpp
```diff
@@ -0,0 +1,151 @@
...
+        CHECK(starts_with(line, "Output format"), "missing header");
+        std::getline(res_stream, line, '\n');
+        CHECK(starts_with(line, "<record#> <instr#>"), "missing 2nd header");
+        std::getline(res_stream, line, '\n');
+        CHECK(starts_with(line, "------"), "missing divider line");
+        std::getline(res_stream, line, '\n');
+        std::stringstream expect_stream;
+        expect_stream << skip_instrs << ":";
```
abhinav92003, line 119: Isn't the first instruction after the divider the markers? And after the markers the isntruction? But here we're verifying the instruction first, and then the markers. 

See comment above. Would be helpful to add a comment with an example expected output.

clients/drcachesim/common/options.cpp
```diff
@@ -105,7 +105,7 @@ droption_t<bytesize_t> op_chunk_instr_count(
     // We do not support tiny chunks.  We do not support disabling chunks with a 0
     // value, to simplify testing: although we're still having to support generating
     // non-zip files for !HAS_ZLIB/!HAS_ZIP!
-    bytesize_t(1000),
+    bytesize_t(10),
```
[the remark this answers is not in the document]
abhinav92003: This was probably only to generate the test trace; should probably be reverted now?

clients/drcachesim/reader/zipfile_file_reader.cpp
abhinav92003, line 160: (Optional). An alternative is
```
(cur_instr_count_/chunk_instr_count_ + 1)*chunk_instr_count_ < stop_count_
```
Seemed a bit simpler to me.

clients/drcachesim/reader/reader.cpp
```diff
@@ -67,231 +67,251 @@ reader_t::operator++()
...
+        cur_ref_.marker.tid = cur_tid_;
+        cur_ref_.marker.marker_type = (trace_marker_type_t)input_entry_->size;
+        cur_ref_.marker.marker_value = input_entry_->addr;
+        // Look for timestamp+cpu duplicated from the prior chunk.  Skip them on
+        // a linear walk.  File readers that support seeking will read them
+        // and use them to start post-seek iteration.
+        if (chunk_instr_count_ > 0 &&
+            cur_ref_.marker.marker_type == TRACE_MARKER_TYPE_TIMESTAMP &&
+            cur_instr_count_ / chunk_instr_count_ !=
+                last_timestamp_instr_count_ / chunk_instr_count_) {
```
abhinav92003, line 272: This assumes that each chunk always has a duplicated timestamp+cpu from the previous chunk, which is fine for second-and-onwards chunks. About the first chunk though, we'll have just one timestamp+cpu, which we don't want to skip. I see last_timestamp_instr_count_ is init to 0, so looks like this will indeed skip the (only) timestamp+cpu in the first chunk. Am I missing something? Does this WAI?

clients/drcachesim/reader/zipfile_file_reader.cpp
abhinav92003: Optionally, to have a meaningful ending while-loop condition, we could move read_next_thread_entry to the loop's end, and also have another one before the loop.

clients/drcachesim/reader/zipfile_file_reader.cpp
```diff
@@ -130,4 +131,100 @@ file_reader_t<zipfile_reader_t>::is_complete()
...
+        // records in the unskipped trace: we should instead not print them out
+        // at all, somehow.
+        trace_entry_t instr = entry_copy_;
+        entry_copy_ = timestamp;
+        process_input_entry();
+        queues_[thread_index].push(cpu);
+        queues_[thread_index].push(instr);
+    } else {
+        // We missed the markers somehow; fall back to just process the instr.
+        VPRINT(this, 1, "Skip failed to find both timestamp and cpu\n");
```
abhinav92003, line 226: This could happen if instruction_count < chunk_instr_count_ and we're not skipping from the beginning of the chunk, right?

clients/drcachesim/tests/skip_unit_tests.cpp
```diff
@@ -0,0 +1,151 @@
...
+        std::streambuf *prior = std::cerr.rdbuf(capture.rdbuf());
+        // Open the trace.
+        std::unique_ptr<reader_t> iter = std::unique_ptr<reader_t>(
+            new zipfile_file_reader_t(op_trace_file.get_value()));
+        CHECK(!!iter, "failed to open zipfile");
+        CHECK(iter->init(), "failed to initialize reader");
+        std::unique_ptr<reader_t> iter_end =
+            std::unique_ptr<reader_t>(new zipfile_file_reader_t());
+        // Run the tool.
+        std::unique_ptr<analysis_tool_t> tool =
```
abhinav92003, line 92: This only tests skipping from the beginning of the chunk, not from the middle.

abhinav92003 (approved):

clients/drcachesim/analyzer.h
```diff
@@ -83,7 +84,7 @@ class analyzer_t {
      * The analyzer calls the initialize() function on each tool before use.
      */
```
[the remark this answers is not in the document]
derekbruening: Done, in the long description of op_skip_instrs.

clients/drcachesim/common/options.cpp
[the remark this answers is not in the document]
derekbruening: It was for the test trace though I thought maybe there's little harm in someone wanting a small value: though it would only be useful for testing.  Reverted back to 1000.

clients/drcachesim/common/options.cpp
```diff
@@ -487,6 +487,11 @@ droption_t<int>
                    "For simulator types that support it, limits analyis to the single "
                    "thread with the given identifier.  0 enables all threads.");
 
+droption_t<bytesize_t> op_skip_instrs(DROPTION_SCOPE_FRONTEND, "skip_instrs", 0,
+                                      "Number of instructions to skip",
+                                      "Specifies the number of instructions to skip "
+                                      "in the beginning of the trace analysis.");
```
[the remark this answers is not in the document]
derekbruening: Done -- but now I'm confused on the earlier comment on the analyzer_t constructor: I assume they both wanted the same thing.

clients/drcachesim/reader/reader.cpp
derekbruening: Yes; I did s/pass up/skip/ to clarify.

clients/drcachesim/reader/reader.h
derekbruening: Done, w/ a TODO since that will be done separately.

clients/drcachesim/reader/zipfile_file_reader.cpp
derekbruening: Done.

clients/drcachesim/reader/zipfile_file_reader.cpp
derekbruening: "estimate" but that is not a great name; replaced with "end-of-chunk"

clients/drcachesim/reader/zipfile_file_reader.cpp
derekbruening: Added a comment that the loop condition is not reached.

clients/drcachesim/tests/skip_unit_tests.cpp
derekbruening: Added notes to both of the option descriptions.

clients/drcachesim/tests/skip_unit_tests.cpp
derekbruening: Done, for each line checked.

clients/drcachesim/tests/skip_unit_tests.cpp
derekbruening: This is ensuring the timestamp has the instr count equal to the skip count.  Added a comment.

clients/drcachesim/reader/zipfile_file_reader.cpp
derekbruening: Hmm, could see it either way: leaning toward current b/c it reads as adding to the current.

clients/drcachesim/reader/reader.cpp
derekbruening: Since last_timestamp_instr_count_ is init to 0, cur_instr_count_/chunk_instr_count_ will be 0 which will match last_timestamp_instr_count_ / chunk_instr_count_ as also 0 and so we will not skip, which is what we want.

clients/drcachesim/reader/zipfile_file_reader.cpp
derekbruening: Yes, if the skip linear walk never sees a timestamp b/c it started after the last one before the target.  I'm thinking I will add such support in the next PR as part of adding the repeat-timestamp to the non-zipfile skips too.  Added a comment here and to the skip function decl that it currently is only fully supported from the start.

clients/drcachesim/tests/skip_unit_tests.cpp
derekbruening: Added a TODO to add middle skip testing once we properly duplicate the timstamp,cpu.

clients/drcachesim/common/options.cpp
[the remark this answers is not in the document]
abhinav92003: That one was for a comment by analyzer_t constructor. But it's fine to clarify it only here.

clients/drcachesim/tests/skip_unit_tests.cpp
abhinav92003: But the timestamp will eventually not have the ordinal right? I guess the test can be adapted later.

clients/drcachesim/reader/reader.cpp
abhinav92003: Ah oops. Got it.

derekbruening:
arm failed on linux.signal0101 which is #5167

clients/drcachesim/tests/skip_unit_tests.cpp
derekbruening: We were only going to not list the record ordinal, while this is testing the instruction ordinal (right now anyway).

Issue #5725, opened 2022-11-09
strange decoder and other test failures in some x86-64 GA CI runs


derekbruening:
One x86-64 run in PR #5723 (which changes just a few bbdup asm test files and so should not affect anything else) suddenly had 4 failures.  They went away on a re-run. 

The decoding ones are weird:

https://github.com/DynamoRIO/dynamorio/actions/runs/3430785437/jobs/5718556145

```
385/434 Test #386: code_api|api.dis .............................................***Failed    0.07 sec
-- diff: 1301c1301
< +0x0c7d   66 72 79             jb     0x0000000000000cf9
---
> +0x0c7d   66 72 79             jb     0x0000000010000cf9
1968c1968
< +0x1322   66 7c f6             jl     0x000000000000131b
---
> +0x1322   66 7c f6             jl     0x000000001000131b
2426c2426
< +0x17ea   4c ff 69 f6          jmp    fword ptr [rcx-0x0a]
---
> +0x17ea   4c ff 69 f6          jmp    tbyte ptr [rcx-0x0a]
2995c2995
< +0x1d87   66 77 39             jnbe   0x0000000000001dc3
---
> +0x1d87   66 77 39             jnbe   0x0000000010001dc3

CMake Error at /home/runner/work/dynamorio/dynamorio/suite/tests/runcmp.cmake:86 (message):
  output in
  /home/runner/work/dynamorio/dynamorio/build_debug-internal-64/suite/tests/dis-x64.expect-out
  failed to match expected output in
  /home/runner/work/dynamorio/dynamorio/build_debug-internal-64/suite/tests/dis-x64.expect-expect
```

```
171/434 Test #169: code_api|client.attach_test ..................................***Failed  Required regular expression not found. Regex=[^starting
thank you for testing attach
thread init
event_post_attach
done
$
]  0.24 sec
CMake Error at /home/runner/work/dynamorio/dynamorio/suite/tests/runall.cmake:124 (message):
  *** kill failed (1): kill: (23301): No such process
```

```
393/434 Test #394: code_api|decenc.drdecode_decenc_x86_64 .......................***Failed    1.09 sec
-- diff: 6581,6584c6581,6583
<  66 e9 00 00          data16 jmp    $0x0000000000007b3f
<  89 c3                mov    %eax, %ebx
<  66 e8 00 00          data16 call   $0x0000000000007b45
<  66 e9 00 00          data16 jmp    $0x0000000000007b49
---
[...]

PR #5736, opened 2022-11-14
i#5383 mac a64, part 3: Add drsyms and test configurations

derekbruening:
Adds static libdwarf and libelftc libraries built for Mach-O AArch64. These are not yet fully tested but this is a first pass to enable configuring and building drsyms and components that depend on it such as drcachesim.  This enables using local M1 builds to replicate GA CI OSX build warnings.

Fixes build warnings and issues from things like the oldest OSX supported.

Fixes configuration issues from x86-only tests.  Building the tests does not yet succeed.

Issue: #5383


derekbruening:
x86-32 failure is client.attach_test `No such process` which started appearing on GA CI recently and is lumped under #5725

CMakeLists.txt
```diff
@@ -1826,15 +1824,18 @@ if (UNIX AND X64)
   find_library(liblz4 lz4)
   if (liblz4)
     message(STATUS "Found liblz4: ${liblz4}")
+    if (APPLE)
+      find_path(LZ4_INCLUDE_DIR lz4.h HINTS /usr/local/include /opt/homebrew/include)
+      find_path(LZ4_LIB_DIR liblz4.a HINTS /usr/local/lib /opt/homebrew/lib)
+      include_directories(${LZ4_INCLUDE_DIR})
```
abhinav92003, line 1830: If find_path fails, looks like the result will be lz4.h-NOTFOUND (https://cmake.org/cmake/help/latest/command/find_path.html),  and it'll just fail in include_directories?

abhinav92003 (approved):

CMakeLists.txt
derekbruening: Looks like it (though the result will be LZ4_INCLUDE_DIR-NOTFOUND).  I mirrored the libsnappy lines above.  The include_ and link_ can be wrapped in an if(), here and for libsnappy, with a warning if they fail.  I went ahead and did that with a shared function for both of these.  It is weird that we need the lib path one after find_library().

derekbruening:
(Jenkins is down for planned service for a few days)

PR #5737, opened 2022-11-15
i#5538 memtrace seek, part 9: Add accurate record count

derekbruening:
Adds a reader subclass to raw2trace for computing the memref_t record count for each chunk.  A new marker is inserted in the chunk header which the zipfile skip code uses to obtain the correct ref count when skipping over chunks.

Adds a count suppression feature where the view tool prints 0 for the record count for the synthetic timestamp+cpu added after a seek.

Updates the seek test with a new trace and new expected output.

Issue: #5538


derekbruening:
I keep hitting discrepancies between clang-format and clang-format-diff.  Both are 14 so I don't think it's a version issue.  We've seen this many times in the past as well.  Not sure there is a good solution -- other than changing our clang-format check to run on the full file somehow.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1727,6 +1729,70 @@ template <typename T> class trace_converter_t {
...
+    get_stream_name() const override
+    {
+        return "";
+    }
+    int
+    entry_memref_count(const trace_entry_t *entry)
+    {
+        // Mirror file_reader_t::open_input_files().
+        // In particular, we need to skip TRACE_TYPE_HEADER and to pass the
+        // tid and tid to the reader before the 2 markers in front of them.
```
abhinav92003, line 1762: s/and tid/and pid/

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1727,6 +1729,70 @@ template <typename T> class trace_converter_t {
...
+            if (entry->type == TRACE_TYPE_HEADER)
+                return 0;
+            else if (entry->type == TRACE_TYPE_THREAD) {
+                list_.push_front(*entry);
+                return 0;
+            } else if (entry->type != TRACE_TYPE_PID) {
+                list_.push_back(*entry);
+                return 0;
+            }
+            saw_pid_ = true;
```
abhinav92003, line 1773: It would be less confusing if saw_pid_ is set in the `entry->type != TRACE_TYPE_PID` else-if block above, and  the `saw_pid_` case below (L1785-) adds the currently entry to list_ and processes the whole list (like the current 1774-1783).  Can remove the current 1774-1783 then. Or am I missing some edge case?

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1727,6 +1729,70 @@ template <typename T> class trace_converter_t {
...
+        // In particular, we need to skip TRACE_TYPE_HEADER and to pass the
+        // tid and tid to the reader before the 2 markers in front of them.
+        if (!saw_pid_) {
+            if (entry->type == TRACE_TYPE_HEADER)
+                return 0;
+            else if (entry->type == TRACE_TYPE_THREAD) {
+                list_.push_front(*entry);
+                return 0;
+            } else if (entry->type != TRACE_TYPE_PID) {
+                list_.push_back(*entry);
```
abhinav92003, line 1770: Why do we need to save TRACE_TYPE_THREAD and TRACE_TYPE_PID at all? We know that process_input_entry() returns false for them so they don't need to be counted.

clients/drcachesim/reader/zipfile_file_reader.cpp
```diff
@@ -204,21 +205,30 @@ file_reader_t<zipfile_reader_t>::skip_thread_instructions(size_t thread_index,
...
         if (memref.marker.type == TRACE_TYPE_MARKER) {
-            if (memref.marker.marker_type == TRACE_MARKER_TYPE_TIMESTAMP)
+            if (memref.marker.marker_type == TRACE_MARKER_TYPE_RECORD_ORDINAL) {
+                cur_ref_count_ = memref.marker.marker_value;
+                at_chunk_header = true;
+            } else if (memref.marker.marker_type == TRACE_MARKER_TYPE_TIMESTAMP) {
                 timestamp = entry_copy_;
-            else if (memref.marker.marker_type == TRACE_MARKER_TYPE_CPU_ID)
+                if (at_chunk_header)
+                    cur_ref_count_--; // Invisible.
```
abhinav92003, line 214: So we decrement here because these entries were duplicated from before and aren't really there (invisible), then down below, we have to do it again using suppress_ref_count_ because queues_ would present those entries yet again in read_next_entry. A comment describing why we need both (the decrement and suppress_ref_count_) would help.

abhinav92003 (approved):

clients/drcachesim/common/trace_entry.h
```diff
@@ -408,6 +411,13 @@ typedef enum {
      */
     TRACE_MARKER_TYPE_CHUNK_FOOTER,
 
+    /**
+     * Indicates the record ordinal for this point in the trace.  This is used
+     * to identify the visible record ordinal when skipping over chunks, and can
+     * generally be ignored by tools.
+     */
+    TRACE_MARKER_TYPE_RECORD_ORDINAL,
```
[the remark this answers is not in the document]
derekbruening: At first I was going to hide it but it takes more work to do that than to not hide it since the zipfile reader looks at the memref level for markers, for convenience, and since the default in the reader is to expose all markers.  What confusion do you mean: the +1 when raw2trace emits it?  I can revisit.  The zipfile reader can look at the trace_entry_t directly before process_input_entry() in a separate way from how it looks at the timestamp.

clients/drcachesim/tracer/raw2trace.h
derekbruening: But then the common case (for millions of records: all but the 1st 5) would be to add one entry to an empty list, and then immediately remove that entry from the list.  That doesn't sound very efficient.

clients/drcachesim/reader/zipfile_file_reader.cpp
derekbruening: This would not be needed if this code looked at these at the trace_entry_t level prior to calling process_input_entry -- which it would have to do for the ordinal marker if that is hidden.  Maybe all 3 should be at the trace_entry_t level.  OK I am trying that.

clients/drcachesim/reader/zipfile_file_reader.cpp
```diff
@@ -204,21 +205,30 @@ file_reader_t<zipfile_reader_t>::skip_thread_instructions(size_t thread_index,
...
     if (timestamp.type == TRACE_TYPE_MARKER && cpu.type == TRACE_TYPE_MARKER) {
         // Insert the two markers.
-        // TODO i#5538: These end up with record ordinals that belong to different
-        // records in the unskipped trace: we should instead not print them out
-        // at all, somehow.
         trace_entry_t instr = entry_copy_;
         entry_copy_ = timestamp;
+        // These synthetic entries are not real records in the unskipped trace, so we
+        // do not associate record counts with them.
+        // at all, somehow.
```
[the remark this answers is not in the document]
derekbruening: Fixed.

clients/drcachesim/common/trace_entry.h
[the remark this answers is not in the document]
derekbruening: I am now hiding it and have shifted the zipfile reader to look at trace_entry_t, per the comment below.

clients/drcachesim/tracer/raw2trace.h
derekbruening: Done.

clients/drcachesim/tracer/raw2trace.h
derekbruening: The reader state machine asserts if it doesn't see them.

PR #5738, opened 2022-11-15
i#5675 record filter: Add filter that toggles at given instr count

abhinav92003:
toggle_filter_t starts off either enabled (outputting all entries) or disabled (not outputting any entries), and then toggles when the given instr count is reached.

Adds logic to record_filter that ensures that essential trace entries, like the shard header, unit header, and footers, are not skipped.

Adds a test for toggle filter.

Issue: #5675

### DIFF
--- a/clients/drcachesim/tests/record_filter_unit_tests.cpp
+++ b/clients/drcachesim/tests/record_filter_unit_tests.cpp
@@ -37,6 +37,7 @@
 #include "droption.h"
 #include "tools/basic_counts.h"
 #include "tools/filter/null_filter.h"
+#include "tools/filter/toggle_filter.h"
 #include "tools/filter/record_filter.h"
 
 #include <vector>
@@ -66,6 +67,37 @@ static droption_t<std::string> op_tmp_output_dir(
     "[Required] Output directory for the filtered trace",
     "Specifies the directory where the filtered trace will be written.");
 
+class test_record_filter_t : public dynamorio::drmemtrace::record_filter_t {
+public:
+    test_record_filter_t(const std::vector<record_filter_func_t *> &filters)
+        : record_filter_t("", filters,
+                          /*verbose=*/0)
+    {
+    }
+    std::vector<trace_entry_t>
+    get_output_entries()
+    {
+        return output;
+    }
+
+protected:
+    bool
+    write_trace_entry(dynamorio::drmemtrace::record_filter_t::per_shard_t *shard,
+                      const trace_entry_t &entry) override
+    {
+        output.push_back(entry);
+        return true;
+    }
+    std::unique_ptr<std::ostream>
+    get_writer(per_shard_t *per_shard, memtrace_stream_t *shard_stream) override
+    {
+        return nullptr;
+    }
+
+private:
+    std::vector<trace_entry_t> output;
+};
+
 static bool
 local_create_dir(const char *dir)
 {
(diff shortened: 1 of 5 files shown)


clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -94,7 +126,122 @@ get_basic_counts(const std::string &trace_dir)
...
+        { { TRACE_TYPE_FOOTER, 0, { 32 } }, { true, true } }
+    };
+
+    /* Check each half. */
+    for (int k = 0; k < 2; ++k) {
+        dynamorio::drmemtrace::record_filter_t::record_filter_func_t *toggle_filter =
+            new dynamorio::drmemtrace::toggle_filter_t(SPLIT_AT_INSTR_COUNT, k == 0);
+        test_record_filter_t *record_filter = new test_record_filter_t({ toggle_filter });
+        void *shard_data = record_filter->parallel_shard_init_stream(0, nullptr, nullptr);
+        for (int i = 0; i < (int)entries.size(); ++i) {
```
derekbruening, line 197: Prefer C++ casts in C+ code

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -94,7 +126,122 @@ get_basic_counts(const std::string &trace_dir)
...
+                return false;
+            }
+        }
+        if (!record_filter->parallel_shard_exit(shard_data)) {
+            fprintf(stderr, "Filtering exit failed\n");
+            return false;
+        }
+        std::vector<trace_entry_t> half = record_filter->get_output_entries();
+        delete record_filter;
+        delete toggle_filter;
```
derekbruening, line 209: Better to use unique_ptr and pass to the vector with get() and avoid needing delete

clients/drcachesim/tools/filter/null_filter.h
```diff
@@ -46,7 +46,8 @@ class null_filter_t : public record_filter_t::record_filter_func_t {
         return nullptr;
     }
     bool
-    parallel_shard_filter(const trace_entry_t &entry, void *shard_data) override
+    parallel_shard_filter(const trace_entry_t &entry, void *shard_data,
+                          bool *stop) override
```
derekbruening, line 50: Label OUT

clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -128,23 +137,99 @@ record_filter_t::parallel_shard_error(void *shard_data)
     return per_shard->error;
 }
 
+bool
+record_filter_t::write_trace_entry(per_shard_t *shard, const trace_entry_t &entry)
+{
+    if (!shard->writer->write((char *)&entry, sizeof(entry))) {
```
derekbruening, line 143: Prefer C++ casts

clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -128,23 +137,99 @@ record_filter_t::parallel_shard_error(void *shard_data)
...
 {
     per_shard_t *per_shard = reinterpret_cast<per_shard_t *>(shard_data);
+    ++per_shard->input_entry_count;
     bool output = true;
+    bool all_stop = true;
     for (int i = 0; i < (int)filters_.size(); ++i) {
-        if (!filters_[i]->parallel_shard_filter(entry, filter_shard_data_[i]))
+        bool stop = false;
+        if (!filters_[i]->parallel_shard_filter(entry, per_shard->filter_shard_data[i],
+                                                &stop))
```
derekbruening, line 162: I would expect to not call this if it has already requested stop (but some other tool did not yet) so the tool writer can assume stop means no more calls.

clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -128,23 +137,99 @@ record_filter_t::parallel_shard_error(void *shard_data)
...
+        per_shard->in_shard_header = false;
+    }
+
+    // We output everything in the trace shard's header.
+    if (per_shard->in_shard_header) {
+        output = true;
+    }
+
+    // Keep track of the last omitted unit header so that we can output it if and when
+    // we output a trace_entry_t from this unit.
```
derekbruening, line 181: I am confused as to why we're doing all this work to support splitting up a trace: I thought that was a use case we did not consider high priority and would not spend time supporting in the near future.  I thought we were only going to support tools removing entries which is all we need for our near-term use cases.

clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -128,23 +137,99 @@ record_filter_t::parallel_shard_error(void *shard_data)
...
+            // next unit now.
+            per_shard->last_filtered_unit_header.clear();
+            ANNOTATE_FALLTHROUGH;
+        case TRACE_MARKER_TYPE_WINDOW_ID:
+        case TRACE_MARKER_TYPE_CPU_ID:
+            per_shard->last_filtered_unit_header.push_back(entry);
+        }
+    }
+
+    // Since we're outputing something from this unit, output its header.
```
derekbruening, line 195: What about the chunks: if instructions are removed they will be the wrong size.

clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -128,23 +137,99 @@ record_filter_t::parallel_shard_error(void *shard_data)
...
+                return false;
+        }
+        per_shard->last_filtered_unit_header.clear();
+    }
+
+    if (!output) {
+        if (entry.type == TRACE_TYPE_FOOTER ||
+            (entry.type == TRACE_TYPE_MARKER &&
+             entry.size == TRACE_MARKER_TYPE_CHUNK_FOOTER)) {
+            output = true;
```
derekbruening, line 208: If this is not allowed it should be an error rather than silently ignoring what the tool wants?  Or is it better to assume the tool writer knows what they're doing and let them create a seemingly-invalid trace?  IMHO either of those is preferable to silently ignoring requests.

clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -128,23 +137,99 @@ record_filter_t::parallel_shard_error(void *shard_data)
...
+        } else if (!all_stop && entry.type == TRACE_TYPE_MARKER) {
+            // Markers that may be required for future instr/memrefs, therefore
+            // must be retained if there's a chance we'll output more entries
+            // later. These are not really tied to a unit header, so to save
+            // space we do not output their unit's header unless some other
+            // instr/memref is output from the same unit. We decided against
+            // buffering them since there may be too many of these.
+            // XXX: This list needs to be kept in sync with our trace format.
```
derekbruening, line 216: I'm not sure I understand why we are going to all this work?  What breaks down if we just let the tool remove what it wants?  If it removes all the non-header entries in one unit it seems fine to leave headers with nothing else there.  If it wants to remove markers we should let it; if that breaks the trace it's the tool writer's fault.  It seems unlikely a typical tool would hit problems: isn't it going to be removing cache-filtered loads or encoding entries and nothing else?  And those cases do not need any of this logic?

clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -128,23 +137,99 @@ record_filter_t::parallel_shard_error(void *shard_data)
...
+            switch (entry.size) {
+            case TRACE_MARKER_TYPE_PHYSICAL_ADDRESS:
+            case TRACE_MARKER_TYPE_PHYSICAL_ADDRESS_NOT_AVAILABLE:
+            case TRACE_MARKER_TYPE_VIRTUAL_ADDRESS:
+            case TRACE_MARKER_TYPE_CHUNK_INSTR_COUNT:
+                // TODO i#5675: The value of the TRACE_MARKER_TYPE_CHUNK_INSTR_COUNT
```
derekbruening, line 222: This one seems like what we *should* implement here: this is worth the infrastructure auto-fixing.  I am confused about the value of the other fixups here.

clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -128,23 +137,99 @@ record_filter_t::parallel_shard_error(void *shard_data)
...
+                // marker should be adjusted based on the count of instrs skipped.
+            case TRACE_MARKER_TYPE_CHUNK_FOOTER: output = true;
+            }
+        }
     }
     // XXX i#5675: Currently we support writing to a single output file, but we may
     // want to write to multiple in the same run; e.g. splitting a trace. For now,
     // we can simply run the tool multiple times, but it can be made more efficient.
-    if (output && !per_shard->writer->write((char *)&entry, sizeof(entry))) {
-        per_shard->error = "Failed to write to output file " + per_shard->output_path;
```
derekbruening, line 144: Not sure why this was removed?

clients/drcachesim/tools/filter/record_filter.h
```diff
@@ -73,9 +73,17 @@ class record_filter_t : public record_analysis_tool_t {
          * trace. \p shard_data is same as what was returned by
          * parallel_shard_init(). The given \p entry is included in the result
          * trace iff all provided #record_filter_func_t return true.
+         * The user should set *\p stop to true to indicate that the trace
+         * should be truncated at the current point (plus necessary entries
+         * like the footer etc). This helps the record_filter tool omit
+         * auxillary entries like encodings and v2p entries, which are
```
derekbruening, line 79: auxiliary

clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -128,23 +137,99 @@ record_filter_t::parallel_shard_error(void *shard_data)
...
+    }
+
+    if (!output) {
+        if (entry.type == TRACE_TYPE_FOOTER ||
+            (entry.type == TRACE_TYPE_MARKER &&
+             entry.size == TRACE_MARKER_TYPE_CHUNK_FOOTER)) {
+            output = true;
+        } else if (!all_stop && entry.type == TRACE_TYPE_MARKER) {
+            // Markers that may be required for future instr/memrefs, therefore
+            // must be retained if there's a chance we'll output more entries
```
derekbruening, line 211: What about encoding entries?

clients/drcachesim/tools/filter/toggle_filter.h
```diff
@@ -0,0 +1,94 @@
...
+        return per_shard;
+    }
+    bool
+    parallel_shard_filter(const trace_entry_t &entry, void *shard_data,
+                          bool *stop) override
+    {
+        per_shard_t *per_shard = reinterpret_cast<per_shard_t *>(shard_data);
+        if (type_is_instr((trace_type_t)entry.type)) {
+            ++per_shard->instr_count;
+            if (per_shard->instr_count == instr_count_toggle_) {
```
derekbruening, line 68: No need for this local counter: the stream can be used.

clients/drcachesim/tools/filter/record_filter.cpp
abhinav92003: > What breaks down if we just let the tool remove what it wants?

This was motivated by the toggle_filter where we want to remove everything before or after a certain instr count. Of course, doing just that is not acceptable since it might remove necessary header/footer entries (which is a clear error), or even encoding, v2p entries (which may silently break some stuff). So some of these shouldn't be removed. I ended up generalising these rules to all filters, instead of just the toggle_filter.

About the "let the tool be more responsible" approach: one drawback was that the tool needs to be aware of all internal entries and remember to not filter them out. But, rethinking now, that's what the record_analysis_tools sign up for, so maybe that's fine.

Perhaps this logic should be moved to inside  the toggle_filter instead of the record_filter tool.

> If it removes all the non-header entries in one unit it seems fine to leave headers with nothing else there.

This was to optimize space, which is the primary goal for the current filtering use cases.

clients/drcachesim/tools/filter/record_filter.cpp
abhinav92003: It's a part of write_trace_entry now.

clients/drcachesim/tools/filter/record_filter.cpp
abhinav92003: Done.

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -94,7 +126,122 @@ get_basic_counts(const std::string &trace_dir)
     return counts;
 }
 
-bool
+static void
+print_entry(trace_entry_t entry)
+{
+    fprintf(stderr, "%d:%d:%ld", entry.type, entry.size, entry.addr);
```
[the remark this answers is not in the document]
abhinav92003: Replaced with PRIiPTR

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -94,7 +126,122 @@ get_basic_counts(const std::string &trace_dir)
...
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_TIMESTAMP, { 7 } }, { true, false } },
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_CPU_ID, { 8 } }, { true, false } },
+        { { TRACE_TYPE_INSTR, 4, { 9 } }, { true, false } },
+        { { TRACE_TYPE_WRITE, 4, { 10 } }, { true, false } },
+        { { TRACE_TYPE_INSTR, 4, { 11 } }, { true, false } },
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_VIRTUAL_ADDRESS, { 12 } },
+          { true, true } },
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_PHYSICAL_ADDRESS, { 13 } },
+          { true, true } },
+        { { TRACE_TYPE_READ, 4, { 14 } }, { true, false } },
```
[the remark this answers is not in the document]
abhinav92003: Fixed.

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -94,7 +126,122 @@ get_basic_counts(const std::string &trace_dir)
...
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_TIMESTAMP, { 18 } }, { true, true } },
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_CPU_ID, { 19 } }, { true, true } },
+        { { TRACE_TYPE_INSTR, 4, { 20 } }, { true, false } },
+        /* First half supposed to end here. See SPLIT_AT_INSTR_COUNT. */
+        { { TRACE_TYPE_INSTR, 4, { 21 } }, { false, true } },
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_VIRTUAL_ADDRESS, { 22 } },
+          { false, true } },
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_PHYSICAL_ADDRESS_NOT_AVAILABLE, { 23 } },
+          { false, true } },
+        { { TRACE_TYPE_READ, 4, { 24 } }, { false, true } },
```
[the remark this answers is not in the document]
abhinav92003: Fixed.

clients/drcachesim/tests/record_filter_unit_tests.cpp
abhinav92003: Done.

clients/drcachesim/tools/filter/toggle_filter.h
abhinav92003: Done.

clients/drcachesim/tools/filter/record_filter.cpp
abhinav92003: Removed the stop param now. filter_funcs now will see all entries till the end. filter_funcs now need to own most aspects of filtering and cannot rely on the record_filter framework filling in necessary entries (like the trace footer) on its own.

clients/drcachesim/tests/record_filter_unit_tests.cpp
abhinav92003: Done.

Issue #5724, opened 2022-11-09
invariant_checker failure on a64: "Chunk instruction counts are inconsistent"


derekbruening:
This happened a couple of times in both debug and release builds:

http://139.178.82.61:8080/job/DynamoRIO-AArch64-Precommit/2313/consoleFull

```
262:   *** postcmd failed (Child aborted): Trace invariant failure in T418399 at
262:   ref # 93263: Chunk instruction counts are inconsistent
262: 
262:   ***
262: 
262: Call Stack (most recent call first):
262:   /var/lib/jenkins/workspace/DynamoRIO-AArch64-Precommit/suite/tests/runmulti.cmake:120 (process_cmdline)
262: 
262: 
285/317 Test #262: code_api|tool.drcacheoff.invariant_checker ......................***Failed    1.08
```

Some later instances.  They are always at ref 93263 it seems:

```
 Trace invariant failure in T498512 at
262:   ref # 93263: Chunk instruction counts are inconsistent
```
```
Trace invariant failure in T548743 at
259:   ref # 93263: Chunk instruction counts are inconsistent
```

derekbruening:
When we first saw this I tried to repro by running the test in a loop on the Jenkins machine but failed to repro.

derekbruening:
With the branch from PR #5721 I ran on the Jenkins machine in my own debug build dir:
```
$ ctest --repeat-until-fail 1000 -V -R invariant_checker\$
...
The following tests passed:
	code_api|tool.drcacheoff.invariant_checker

100% tests passed, 0 tests failed out of 1

Total Test time (real) = 789.60 sec
```

derekbruening:
I've found the error: for consecutive delayed branches in the same output buffer split by a chunk boundary, we pre-increment the instr count, but open_new_chunk() clears it to 0 and we lose the count for the 2nd branch.

I've also discovered a related bug: we do not re-emit an encoding for this 2nd branch.  I'm considering that under this same issue.

derekbruening:
Another (less serious) issue is that we'll emit encodings at the very end of the chunk where they're wasted for skips.

derekbruening:
Inserting encodings is messy due to the distance between the generation and the identification of a need for new encodings.  It would also be good to have more tests:

[...]

PR #5739, opened 2022-11-15
i#5538 memtrace seek, part 10: Top-level headers

derekbruening:
Adds cached values of the 5 top-level headers to the metrace_stream_t interface and implements this for the readers.

Adds checks of these values to invariant_checker.

Adds a test with skipped instructions using the skip_unit_tests checked-in trace.

Reverses the order of the initial 2 markers and the tid,pid pair sent to the reader to avoid a 0 tid in tools.  I am surprised this hasn't caused more problems and I thought it was already this fixed way.

Issue: #5538


derekbruening:
x32 failure is api.startstop assert about sysenter #4604

clients/drcachesim/reader/reader.cpp
```diff
@@ -281,11 +272,18 @@ reader_t::process_input_entry()
...
             last_timestamp_instr_count_ = cur_instr_count_;
+        else if (cur_ref_.marker.marker_type == TRACE_MARKER_TYPE_VERSION)
+            version_ = cur_ref_.marker.marker_value;
+        else if (cur_ref_.marker.marker_type == TRACE_MARKER_TYPE_FILETYPE) {
+            filetype_ = cur_ref_.marker.marker_value;
+            if (TESTANY(OFFLINE_FILE_TYPE_ENCODINGS, filetype_))
+                expect_no_encodings_ = false;
+        } else if (cur_ref_.marker.marker_type == TRACE_MARKER_TYPE_CACHE_LINE_SIZE)
+            cache_line_size_ = cur_ref_.marker.marker_value;
+        else if (cur_ref_.marker.marker_type == TRACE_MARKER_TYPE_PAGE_SIZE)
```
abhinav92003, line 283: Where does raw2trace write the TRACE_MARKER_TYPE_PAGE_SIZE marker? Couldn't find it in raw2trace_t::process_header.

clients/drcachesim/reader/file_reader.h
```diff
@@ -121,10 +123,8 @@ template <typename T> class file_reader_t : public reader_t {
             return reader_t::skip_instructions(instruction_count);
         }
         // If the user asks to skip from the very start, we still need to find the chunk
-        // count marker and drain the header queue.
-        // TODO i#5538: Record all of the header values until the first timestamp
-        // and present them as new memtrace_stream_t interfaces.
-        while (chunk_instr_count_ == 0) {
+        // count marker and drain the header queue and populate the stream header values.
+        while (chunk_instr_count_ == 0 || page_size_ == 0) {
```
abhinav92003, line 127: Why do we need to have two conditions here? Either one of the page size or chunk instr marker should be the last one always right?

clients/drcachesim/reader/reader.h
```diff
@@ -171,9 +196,14 @@ class reader_t : public std::iterator<std::input_iterator_tag, memref_t>,
...
     uint64_t cur_ref_count_ = 0;
     uint64_t cur_instr_count_ = 0;
-    uint64_t chunk_instr_count_ = 0; // Unchanging once set to non-zero.
     uint64_t last_timestamp_instr_count_ = 0;
     trace_entry_t *input_entry_ = nullptr;
+    // Remember top-level headers for the memtrace_stream_t interface.
+    uint64_t version_ = 0;
+    uint64_t filetype_ = 0;
+    uint64_t cache_line_size_ = 0;
+    uint64_t chunk_instr_count_ = 0; // Unchanging once set to non-zero.
```
abhinav92003, line 205: The "Unchanging..." comment is true for other fields too right? It's confusing that it's mentioned only for one.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -104,12 +121,26 @@ bool
...
 {
     per_shard_t *shard = reinterpret_cast<per_shard_t *>(shard_data);
-    ++shard->ref_count_;
     if (shard->tid_ == -1 && memref.data.tid != 0)
         shard->tid_ = memref.data.tid;
-    // XXX i#5538: Have the infrastructure provide a common instr and record count.
+    // We check the memtrace_stream_t counts with our own, unless there was an
+    // instr skip from the start where we cannot compare, or we're in a unit
+    // test with no stream interface, or we're in serial mode (since we want
+    // per-shard counts for error reporting; XXX: we could add global local counts).
```
abhinav92003, line 129: "global local"?

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -104,12 +121,26 @@ bool
...
+    ++shard->ref_count_;
     if (type_is_instr(memref.instr.type))
         ++shard->instr_count_;
+    if (shard->instr_count_ <= 1 && !shard->skipped_instrs_ && shard->stream != nullptr &&
```
abhinav92003, line 133: Even when there's an instr skip later in the trace, the invariant checker won't be able to verify the stream counts, right?

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -78,15 +86,24 @@ invariant_checker_t::parallel_shard_supported()
...
 {
     auto per_shard = std::unique_ptr<per_shard_t>(new per_shard_t);
+    per_shard->stream = shard_stream;
     void *res = reinterpret_cast<void *>(per_shard.get());
     std::lock_guard<std::mutex> guard(shard_map_mutex_);
     shard_map_[shard_index] = std::move(per_shard);
     return res;
 }
 
+// We have no stream interface in invariant_checker_test unit tests.
```
abhinav92003, line 100: Instead of adding complexity here, can we just use the stream interface in invariant_checker_test?

abhinav92003 (approved):

clients/drcachesim/tools/invariant_checker.cpp
derekbruening: Do you have a suggestion for how to implement that?  To me it does not look easy.  There is no reader in that test; it uses synthetic sequences and custom process_memref calls.  A reader taking a vector?  Maybe in the end that would be nice; but it sure seems like a lot more work than the tweaks here.

clients/drcachesim/reader/file_reader.h
```diff
@@ -221,6 +221,7 @@ template <typename T> class file_reader_t : public reader_t {
                 return false;
             }
             // Read the meta entries until we hit the pid.
+            std::queue<trace_entry_t> marker_queue;
```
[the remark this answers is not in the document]
derekbruening: Yes, the file has marker,marker,tid,pid but the reader has to fill in the tid+pid fields of memref_t so it is best for it to see the tid+pid first (o/w tools doing per-thread data in serial mode have problems).  Added a comment.

clients/drcachesim/reader/reader.cpp
derekbruening: I think it's in the regular record processing code (along with any further markers added in the future).  raw2trace doesn't need to know its value so it doesn't need to treat it specially.

clients/drcachesim/reader/file_reader.h
derekbruening: Yes just page size; removed the chunk.  It would be nicer to wait for the timestamp but then we've already read that record and it complicates things so unfortunately we have an assumption on the final marker.  Added a comment.

clients/drcachesim/tools/invariant_checker.cpp
derekbruening: Yes that is deliberate; reworded as "add our own global counts to compare to the serial stream"

clients/drcachesim/tools/invariant_checker.cpp
derekbruening: Yes.  Not sure how to detect that though.  Added a comment.

clients/drcachesim/tools/invariant_checker.cpp
derekbruening: Added a comment about possibly refactoring the test.

clients/drcachesim/reader/file_reader.h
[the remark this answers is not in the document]
abhinav92003: We could consider reordering the original order of the fields in raw2trace.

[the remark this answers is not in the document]
derekbruening: That would avoid the reordering multiple pieces of code are having to do right now.  But it makes it hard to change things with a version bump b/c the version is not in the first record so we'd be locked into that format?  There are tradeoffs.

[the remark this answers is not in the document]
abhinav92003: Seems fine to skip trace format change for now. Looks like it needs more thought.

derekbruening:
run arm tests

derekbruening:
Hitting the same attach timeouts it seems on Jenkins: #5740

abhinav92003:
> Hitting the same attach timeouts it seems on Jenkins: #5740

We probably should add it to the ignore-list?

derekbruening:
Re-run is still hitting the attach ones: someone will have to put effort into fixing or adding to ignore list.  Also hit the inv checker on the re-run (not on prior runs): #5724.

PR #5742, opened 2022-11-18
i#5675 record filter: Add cache_filter and type_filter.

abhinav92003:
Adds cache_filter that allows filtering data and/or instr entries by passing the addresses through a cache. Uses the existing drcachesim cache simulator to implement this.

Adds a did_last_access_hit() public API in a sub-class of cache_stats_t to get the status of the last processed access. This is used in cache_filter to determine whether the current trace_entry_t should be output or not.

Adds type_filter that allows filtering trace entries by their type, including markers by their type.

Adds the ability to stop filtering when a certain timestamp is reached. At this point, a new marker TRACE_MARKER_TYPE_FILTER_ENDPOINT is added to help tools figure out the warmup region in the trace. Adds a new get_last_timestamp() public API to memtrace_stream_t to enable this.

Adds a test that verifies cache and type filter operation.

Adds command line options to the record_filter_launcher executable to allow using the cache and type filters.

Issue: #5675


clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -128,23 +143,75 @@ record_filter_t::parallel_shard_error(void *shard_data)
...
+
 bool
 record_filter_t::parallel_shard_memref(void *shard_data, const trace_entry_t &entry)
 {
     per_shard_t *per_shard = reinterpret_cast<per_shard_t *>(shard_data);
+    ++per_shard->input_entry_count;
     bool output = true;
-    for (int i = 0; i < (int)filters_.size(); ++i) {
-        if (!filters_[i]->parallel_shard_filter(entry, filter_shard_data_[i]))
+    if (stop_timestamp_us_ == 0 ||
```
derekbruening, line 164: What does per_shard->enabled mean?  I would expect it to guard all the code here and just exit up front if not enabled.

clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -128,23 +143,75 @@ record_filter_t::parallel_shard_error(void *shard_data)
...
+
+    // Optimize space by outputting the unit header only if we are outputting something
+    // from that unit.
+    if (entry.type == TRACE_TYPE_MARKER) {
+        switch (entry.size) {
+        case TRACE_MARKER_TYPE_TIMESTAMP:
+            // No need to remember the previous unit's header anymore. We're in the
+            // next unit now.
+            // XXX: it may happen that we never output a unit header due to this
+            // optimization. We should ensure that we output it atleast once. We
```
derekbruening, line 190: missing space

derekbruening, line 190: What is the reason to output "at least once"?

clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -159,7 +226,8 @@ record_filter_t::process_memref(const trace_entry_t &memref)
 bool
 record_filter_t::print_results()
 {
-    // TODO i#5675: Print stats about filtered entries.
+    std::cerr << "Outputted " << output_entry_count_ << " entries from "
```
derekbruening, line 229: nit: I think the past tense of "output" is just "output".  Apparently a search shows that "outputted" is also considered an alternative: it sounds wrong to me though; maybe regional dialects.

clients/drcachesim/tools/filter/record_filter.h
```diff
@@ -83,11 +83,23 @@ class record_filter_t : public record_analysis_tool_t {
...
+        }
+
+    protected:
+        std::string error_string_;
     };
 
     record_filter_t(const std::string &output_dir,
                     const std::vector<record_filter_func_t *> &filters,
-                    unsigned int verbose);
+                    uint64_t stop_timestamp_us, unsigned int verbose);
```
derekbruening, line 102: Same comment about the "_us" as above.

clients/drcachesim/tools/filter/type_filter.h
```diff
@@ -0,0 +1,110 @@
...
+    parallel_shard_init(memtrace_stream_t *shard_stream) override
+    {
+        return nullptr;
+    }
+    bool
+    parallel_shard_filter(const trace_entry_t &entry, void *shard_data) override
+    {
+        if (remove_trace_types_.find(static_cast<trace_type_t>(entry.type)) !=
+            remove_trace_types_.end()) {
+            if (entry.type != TRACE_TYPE_MARKER)
```
derekbruening, line 90: I don't understand this: the user is supposed to pass the marker type to the list to be removed even if not all markers will be removed and even though there is a dedicated list of marker sub-types to remove?  It seems more intuitive and simpler to *not* pass the marker type and *not* need this special check here for that use case.  It needs to be documented regardless since as a user I would not expect this behavior.  If I want to remove *all* marker types, I would want to be able to do that by passing the marker type to the remove_trace_types list and *not* have to list every single possible type (thus making my tool not forward-compatible).

clients/drcachesim/tools/record_filter_launcher.cpp
```diff
@@ -59,6 +61,48 @@ static droption_t<unsigned int> op_verbose(DROPTION_SCOPE_ALL, "verbose", 0, 0,
...
+parse_string(std::string s, char sep = ',')
+{
+    size_t pos;
+    if (s.empty())
+        return {};
+    std::vector<T> vec;
+    do {
+        pos = s.find(sep);
+        vec.push_back(static_cast<T>(std::stoi(s.substr(0, pos))));
+        s.erase(0, pos + 1);
```
derekbruening, line 101: Do we need to erase and destroy the string?  We can pass the current position to the other funcs and let the string param be a const which seems more the normal case for string params.

clients/drcachesim/tools/record_filter_launcher.cpp
```diff
@@ -75,17 +119,43 @@ _tmain(int argc, const TCHAR *targv[])
...
+                    op_cache_filter_size.get_value(),
+                    /*filter_data=*/true, /*filter_instrs=*/false)));
+    }
+    if (op_remove_trace_types.specified() || op_remove_marker_types.specified()) {
+        std::vector<trace_type_t> filter_trace_types =
+            parse_string<trace_type_t>(op_remove_trace_types.get_value());
+        std::vector<trace_marker_type_t> filter_marker_types =
+            parse_string<trace_marker_type_t>(op_remove_marker_types.get_value());
+        if (!filter_marker_types.empty())
+            filter_trace_types.push_back(TRACE_TYPE_MARKER);
```
derekbruening, line 141: See above: putting this here seems to just cause confusion and complexity.

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -94,7 +147,176 @@ get_basic_counts(const std::string &trace_dir)
...
+            fprintf(stderr, "Couldn't construct a cache_filter %s",
+                    cache_filter->get_error_string().c_str());
+            return false;
+        }
+        filters.push_back(cache_filter.get());
+
+        auto type_filter =
+            std::unique_ptr<dynamorio::drmemtrace::record_filter_t::record_filter_func_t>(
+                new dynamorio::drmemtrace::type_filter_t(
+                    { TRACE_TYPE_ENCODING, TRACE_TYPE_MARKER },
```
derekbruening, line 233: But we don't want to remove all markers: looks like a bug.
Update: I see special casing to ignore this: see other comments as this is confusing.

derekbruening (approved):

clients/drcachesim/reader/reader.cpp
```diff
@@ -273,9 +273,10 @@ reader_t::process_input_entry()
         } else {
             have_memref = true;
         }
-        if (cur_ref_.marker.marker_type == TRACE_MARKER_TYPE_TIMESTAMP)
+        if (cur_ref_.marker.marker_type == TRACE_MARKER_TYPE_TIMESTAMP) {
             last_timestamp_instr_count_ = cur_instr_count_;
-        else if (cur_ref_.marker.marker_type == TRACE_MARKER_TYPE_VERSION)
+            last_timestamp_ = cur_ref_.marker.marker_value;
```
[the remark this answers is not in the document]
abhinav92003: Done.

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -94,7 +147,176 @@ get_basic_counts(const std::string &trace_dir)
     return counts;
 }
 
-bool
+static void
+print_entry(trace_entry_t entry)
+{
+    fprintf(stderr, "%d:%d:%" PRIxPTR, entry.type, entry.size, entry.addr);
```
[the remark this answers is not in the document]
abhinav92003: Done.

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -66,6 +70,55 @@ static droption_t<std::string> op_tmp_output_dir(
...
+        return last_timestamp_us_;
+    }
+    void
+    set_last_timestamp(uint64_t last_timestamp_us)
+    {
+        last_timestamp_us_ = last_timestamp_us;
+    }
+
+private:
+    uint64_t last_timestamp_us_;
```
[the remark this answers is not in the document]
abhinav92003: Yes, it's microseconds. Removed it.

clients/drcachesim/tools/filter/cache_filter.h
```diff
@@ -0,0 +1,107 @@
...
+        , cache_size_(cache_size)
+        , filter_data_(filter_data)
+        , filter_instrs_(filter_instrs)
+    {
+    }
+    void *
+    parallel_shard_init(memtrace_stream_t *shard_stream) override
+    {
+        per_shard_t *per_shard = new per_shard_t;
+        per_shard->cache = new cache_lru_t;
```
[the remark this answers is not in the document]
abhinav92003: You mean the cache_lru_t can be an auto var? But it needs to survive beyond this routine.

clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -128,23 +143,75 @@ record_filter_t::parallel_shard_error(void *shard_data)
...
 record_filter_t::parallel_shard_memref(void *shard_data, const trace_entry_t &entry)
 {
     per_shard_t *per_shard = reinterpret_cast<per_shard_t *>(shard_data);
+    ++per_shard->input_entry_count;
     bool output = true;
-    for (int i = 0; i < (int)filters_.size(); ++i) {
-        if (!filters_[i]->parallel_shard_filter(entry, filter_shard_data_[i]))
+    if (stop_timestamp_us_ == 0 ||
+        per_shard->shard_stream->get_last_timestamp() < stop_timestamp_us_) {
+        for (int i = 0; i < (int)filters_.size(); ++i) {
```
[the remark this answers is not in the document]
abhinav92003: Ah, missed this one. Done.

clients/drcachesim/tools/filter/record_filter.cpp
abhinav92003: Hmm.. yeah my code here was a bit convoluted wrt the `enabled` var. Fixed it -- used `enabled` to guard the parallel_shard_filter calls.. Though we cannot just exit early from this routine.

clients/drcachesim/tools/filter/record_filter.cpp
abhinav92003: Done.

abhinav92003: Because if there's no timestamp entry at all, some of our assumptions fail about the trace format (e.g. "everything up to the timestamp entry is the shard header").

This is still in line with what we discussed: if a record_filter_func_t wants to filter out timestamp entries, it can still do it. This "at least once" requirement is for the space optimization here.

clients/drcachesim/tools/filter/record_filter.cpp
abhinav92003: Changed it. I was on the fence and made it as such just because of a quick search.

clients/drcachesim/tools/filter/record_filter.h
abhinav92003: Changed.

clients/drcachesim/tools/filter/type_filter.h
abhinav92003: Makes sense. Changed it.

clients/drcachesim/tools/record_filter_launcher.cpp
abhinav92003: Fixed.

clients/drcachesim/tools/record_filter_launcher.cpp
abhinav92003: Using the find(sep, pos) overload now so that we don't have to modify the string.

clients/drcachesim/tests/record_filter_unit_tests.cpp
abhinav92003: Fixed.

clients/drcachesim/common/trace_entry.cpp
```diff
@@ -80,4 +80,5 @@ const char *const trace_type_names[] = {
     "prefetch_write_l2_nt",
     "prefetch_write_l3",
     "prefetch_write_l3_nt",
+    "instr_encoding",
```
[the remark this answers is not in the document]
abhinav92003: Acknowledged.

clients/drcachesim/common/memtrace_stream.h
```diff
@@ -119,4 +125,58 @@ class memtrace_stream_t {
     get_page_size() const = 0;
 };
 
+/**
+ * Implementation of memtrace_stream_t useful for mocks in tests.
```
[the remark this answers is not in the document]
abhinav92003: Acknowledged.

clients/drcachesim/common/trace_entry.h
```diff
@@ -418,6 +418,12 @@ typedef enum {
      */
     TRACE_MARKER_TYPE_RECORD_ORDINAL,
 
+    /**
+     * Indicates a point in the trace where filtering ended.
+     * This is currently added by the record_filter tool to annotate when the
+     * warmup part of the trace ends.
```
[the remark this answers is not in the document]
abhinav92003: Changed it to TRACE_MARKER_TYPE_FILTER_ENDPOINT.

clients/drcachesim/simulator/caching_device_stats.h
```diff
@@ -243,6 +252,7 @@ class caching_device_stats_t {
 #else
     FILE *file_;
 #endif
+    bool last_hit_;
```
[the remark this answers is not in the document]
abhinav92003: Clarified.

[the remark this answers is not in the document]
abhinav92003: I mean I changed the identifier.

clients/drcachesim/tools/filter/cache_filter.h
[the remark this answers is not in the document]
abhinav92003: Ah, you mean in the per_shard_t. Done.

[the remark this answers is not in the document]
abhinav92003: I guess we can change it later if we want to support non-LRU caches.

clients/drcachesim/simulator/caching_device_stats.h
```diff
@@ -192,6 +193,14 @@ class caching_device_stats_t {
         }
     }
 
+    // Returns whether the last access to this cache (not to a child cache)
+    // was a hit.
+    bool
+    did_last_access_hit()
```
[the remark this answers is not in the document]
abhinav92003: Fair enough. Used a subclass now.

clients/drcachesim/tools/filter/cache_filter.h
```diff
@@ -0,0 +1,107 @@
...
+        if (!per_shard->cache->init(cache_associativity_, cache_line_size_, cache_size_,
+                                    nullptr, new cache_stats_t(cache_line_size_))) {
```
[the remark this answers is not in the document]
abhinav92003: Fair enough. Used a subclass now.

derekbruening (approved):

PR #5745, opened 2022-11-18
i#5709: Fix cpu scheduling regression

### LINKED ISSUE #5709: dr$sim -cpu_scheduling ignores migrations
It binds a software thread to the first core it appeared on.  I would expect it to follow migrations and schedule each thread segment on to the appropriate core.

Here we have a single thread that was on cpu 8 initially but then moved to cpu 2, and the results from dr$sim which I would call incorrect:
```
$ bin64/drrun -t drcachesim -indir drmemtrace.*.dir -cpu_scheduling
Cache simulation results:
Core #0 (1 traced CPU(s): #8)
  L1I stats:
    Hits:                          141,938
    Misses:                          1,138
    Compulsory misses:               1,108
    Invalidations:                       0
    Miss rate:                        0.80%
  L1D stats:
    Hits:                           42,491
    Misses:                          1,390
    Compulsory misses:               2,112
    Invalidations:                       0
    Prefetch hits:                     231
    Prefetch misses:                 1,159
    Miss rate:                        3.17%
Core #1 (1 traced CPU(s): #2)
  L1I stats:
    Hits:                                0
    Misses:                              0
    Compulsory misses:                   0
    Invalidations:                       0
  L1D stats:
    Hits:                                0
    Misses:                              0
    Compulsory misses:                   0
    Invalidations:                       0
```

derekbruening:
Fixes a bug where cpu scheduling failed to follow thread migrations due to caching that was added without invalidation on a migration.

Adds a test by checking in a simple_app trace with a migration and ensuring we see both recorded cpus.

Fixes #5709


bete0 (approved):

derekbruening:
x86-64 failure is api.detach_state #5123 vmareas.c assert

PR #5748, opened 2022-11-21
Ignore bb with null instruction count

kuhanov:
In general I have bb with one jmp instruction on java workload
```
TAG  0x0000000000000000
 +0    m4 @0x00007f9f6ecbbc50  65 48 89 0c 25 d8 00 mov    %rcx -> %gs:0x000000d8[8byte]
                               00 00
 +9    m4 @0x00007f9f6ecba030  65 48 8b 0c 25 38 01 mov    %gs:0x00000138[8byte] -> %rcx
                               00 00
 +18   L3 @0x00007f9f6ecbc868  eb fe                jmp    $0x00007f9f65087e50
END 0x0000000000000000
```
In this case `instru_t::count_app_instrs` return the null counts of instructions in list

So, we have assert in converter phase
`append_bb_entries`
```
...
        if (instr_count == 0) {
            // L0 filtering adds a PC entry with a count of 0 prior to each memref.
            skip_icache = true;
            instr_count = 1;
            // We should have set a flag to avoid peeking forward on instr entries
            // based on OFFLINE_FILE_TYPE_FILTERED and
            // OFFLINE_FILE_TYPE_IFILTERED.
            DR_ASSERT(instrs_are_separate);
...
```

Could I just ignore such bb?
Thx, Kirill

### DIFF
--- a/clients/drcachesim/tracer/instru_offline.cpp
+++ b/clients/drcachesim/tracer/instru_offline.cpp
@@ -736,7 +736,8 @@ offline_instru_t::instrument_instr(void *drcontext, void *tag, void *bb_field,
     reg_id_t reg_tmp;
     if (!memref_needs_full_info_) {
         // We write just once per bb, if not filtering.
-        if (per_block->instr_count > MAX_INSTR_COUNT)
+        if (per_block->instr_count > MAX_INSTR_COUNT ||
+            per_block->instr_count == 0)
             return adjust;
         pc = dr_fragment_app_pc(tag);
     } else {


derekbruening:
I would want to understand this better: if this is a synthetic block with no application analog then it seems better for the tracer to not output a bb record at all and raw2trace shouldn't see anything.  Is this a drwrap artificial block?  It looks like a jump to the exposed IBL which is used by drwrap?  Or something else?  What is the application code sequence that led to this?

kuhanov:
> exposed

replace left address column to app_pc
```
TAG  0x0000000000000000
 +0    L3 @0x00007f35cd088f07  49 bb 00 f0 9e 18 38 mov    $0x00007f38189ef000 -> %r11
                               7f 00 00
 +10   L3 @0x00007f35cd088f11  41 85 03             test   (%r11)[4byte] %eax
 +13   L3 @0x00007f35cd088f14  c5 f9 7e d5          vmovd  %xmm2[4byte] -> %ebp
 +17   L3 @0x00007f35cd088f18  4c 89 04 24          mov    %r8 -> (%rsp)[8byte]
 +21   L3 @0x00007f35cd088f1c  89 6c 24 0c          mov    %ebp -> 0x0c(%rsp)[4byte]
 +25   L3 @0x00007f35cd088f20  89 5c 24 08          mov    %ebx -> 0x08(%rsp)[4byte]
 +29   L3 @0x00007f35cd088f24  49 8b f0             mov    %r8 -> %rsi
 +32   L3 @0x00007f35cd088f27  e8 04 d6 ff ff       call   $0x00007f35cd086530 %rsp -> %rsp 0xfffffff8(%rsp)[8byte]
END 0x0000000000000000

TAG  0x0000000000000000
 +0    L3 @0x00007f35cd086530  eb fe                jmp    $0x00007f35cd086530
END 0x0000000000000000

TAG  0x0000000000000000
 +0    L3 @0x00007f35cd086530  e9 eb ea f7 ff       jmp    $0x00007f35cd005020
END 0x0000000000000000

TAG  0x0000000000000000
 +0    L3 @0x00007f35cd005020  55                   push   %rbp %rsp -> %rsp 0xfffffff8(%rsp)[8byte]
 +1    L3 @0x00007f35cd005021  48 8b ec             mov    %rsp -> %rbp
 +4    L3 @0x00007f35cd005024  9c                   pushf  %rsp -> %rsp 0xfffffff8(%rsp)[8byte]
 +5    L3 @0x00007f35cd005025  48 83 ec 08          sub    $0x0000000000000008 %rsp -> %rsp
 +9    L3 @0x00007f35cd005029  48 89 64 24 d8       mov    %rsp -> 0xffffffd8(%rsp)[8byte]
 +14   L3 @0x00007f35cd00502e  48 81 ec 80 00 00 00 sub    $0x0000000000000080 %rsp -> %rsp
 +21   L3 @0x00007f35cd005035  48 89 44 24 78       mov    %rax -> 0x78(%rsp)[8byte]
 +26   L3 @0x00007f35cd00503a  48 89 4c 24 70       mov    %rcx -> 0x70(%rsp)[8byte]
 +31   L3 @0x00007f35cd00503f  48 89 54 24 68       mov    %rdx -> 0x68(%rsp)[8byte]
 +36   L3 @0x00007f35cd005044  48 89 5c 24 60       mov    %rbx -> 0x60(%rsp)[8byte]
 +41   L3 @0x00007f35cd005049  48 89 6c 24 50       mov    %rbp -> 0x50(%rsp)[8byte]
 +46   L3 @0x00007f35cd00504e  48 89 74 24 48       mov    %rsi -> 0x48(%rsp)[8byte]
 +51   L3 @0x00007f35cd005053  48 89 7c 24 40       mov    %rdi -> 0x40(%rsp)[8byte]
 +56   L3 @0x00007f35cd005058  4c 89 44 24 38       mov    %r8 -> 0x38(%rsp)[8byte]
 +61   L3 @0x00007f35cd00505d  4c 89 4c 24 30       mov    %r9 -> 0x30(%rsp)[8byte]
 +66   L3 @0x00007f35cd005062  4c 89 54 24 28       mov    %r10 -> 0x28(%rsp)[8byte]
 +71   L3 @0x00007f35cd005067  4c 89 5c 24 20       mov    %r11 -> 0x20(%rsp)[8byte]
 +76   L3 @0x00007f35cd00506c  4c 89 64 24 18       mov    %r12 -> 0x18(%rsp)[8byte]
 +81   L3 @0x00007f35cd005071  4c 89 6c 24 10       mov    %r13 -> 0x10(%rsp)[8byte]
 +86   L3 @0x00007f35cd005076  4c 89 74 24 08       mov    %r14 -> 0x08(%rsp)[8byte]
 +91   L3 @0x00007f35cd00507b  4c 89 3c 24          mov    %r15 -> (%rsp)[8byte]
 +95   L3 @0x00007f35cd00507f  48 81 ec 00 02 00 00 sub    $0x0000000000000200 %rsp -> %rsp
 +102  L3 @0x00007f35cd005086  48 0f ae 04 24       fxsave64  -> (%rsp)[512byte]
END 0x0000000000000000

```
So there is some infinity loop - jmp to itself
Probably java generated something dynamically and replace this jmp to correct address later
because I could see the next BB handling with jmp only but with some address and count is 1 in this case.
Kirill

kuhanov:
Hi, @derekbruening 
Maybe we don't need ignore such instructions if it is one in the BB like in my case?
```
        if (!in_emulation_region && instr_is_app(inst)) {
            // Hooked native functions end up with an artifical jump whose translation
            // is its target.  We do not want to count these.
            if (!(instr_is_ubr(inst) && opnd_is_pc(instr_get_target(inst)) &&
                  opnd_get_pc(instr_get_target(inst)) == instr_get_app_pc(inst)))
                ++count;
        }
```
Thx, Kirill

derekbruening:
> So there is some infinity loop - jmp to itself
> Probably java generated something dynamically and replace this jmp to correct address later
> because I could see the next BB handling with jmp only but with some address and count is 1 in this case.

But how can it be replaced between the time DR creates the block and executes the block?  There is no app control point in between.  Where in the code are you printing these instrlists?

An actual infinite loop, or self-modifying code with stores prior to the jump which modify the jump, would cause incorrect counts: but the latter case would be an early block exit which I believe is not handled in general.

To tighten the current check you quote for `Hooked native functions` we'd want to get more details on those in order to add more conditions: are they always a single-instruction block with the jump?  I thought DR tried to hide the hook details.

kuhanov:
> To tighten the current check you quote for `Hooked native functions` we'd want to get more details on those in order to add more conditions: are they always a single-instruction block with the jump? I thought DR tried to hide the hook details.

In my case, it was the block from one instruction

derekbruening:
> > To tighten the current check you quote for `Hooked native functions` we'd want to get more details on those in order to add more conditions: are they always a single-instruction block with the jump? I thought DR tried to hide the hook details.
> 
> In my case, it was the block from one instruction

I'm referring to the code you quoted `// Hooked native functions end up with an artifical jump whose translation` which is referring to Windows library functions hooked by DR (maybe also sysenter vsyscall on Linux?)

kuhanov:
> I'm referring to the code you quoted `// Hooked native functions end up with an artifical jump whose translation` which is referring to Windows library functions hooked by DR (maybe also sysenter vsyscall on Linux?)

Hm, my application is on linux and I was inside this condition but counter was not increased because
`(instr_is_ubr(inst) is True and opnd_is_pc(instr_get_target(inst)) is True` so `!(instr_is_ubr(inst) && opnd_is_pc(instr_get_target(inst)) is False`

derekbruening:
> > I'm referring to the code you quoted `// Hooked native functions end up with an artifical jump whose translation` which is referring to Windows library functions hooked by DR (maybe also sysenter vsyscall on Linux?)
> 
> Hm, my application is on linux and I was inside this condition but counter was not increased because `(instr_is_ubr(inst) is True and opnd_is_pc(instr_get_target(inst)) is True` so `!(instr_is_ubr(inst) && opnd_is_pc(instr_get_target(inst)) is False`

Let me try again to explain: the existing code is trying to only match hooked native functions.  You have hit some other case which matches it.  There are 2 issues: A) Make sure we understand your case as it doesn't make sense to me: https://github.com/DynamoRIO/dynamorio/pull/5748#issuecomment-1329117521; B) If we try to change the existing code to not match your case we would want to tighten up the check to only cover the hooked function, so we would want to figure out some other way to identify the hook case.

kuhanov:
> > > I'm referring to the code you quoted `// Hooked native functions end up with an artifical jump whose translation` which is referring to Windows library functions hooked by DR (maybe also sysenter vsyscall on Linux?)
> > 
> > 
> > Hm, my application is on linux and I was inside this condition but counter was not increased because `(instr_is_ubr(inst) is True and opnd_is_pc(instr_get_target(inst)) is True` so `!(instr_is_ubr(inst) && opnd_is_pc(instr_get_target(inst)) is False`
> 
> Let me try again to explain: the existing code is trying to only match hooked native functions. You have hit some other case which matches it. There are 2 issues: A) Make sure we understand your case as it doesn't make sense to me: [#5748 (comment)](https://github.com/DynamoRIO/dynamorio/pull/5748#issuecomment-1329117521); B) If we try to change the existing code to not match your case we would want to tighten up the check to only cover the hooked function, so we would want to figure out some other way to identify the hook case.

Tried to collect more info (bb instrumentation, syscalls and so on ) in debug mode `-debug -logmask 0x000004b8 -loglevel 4`
but issue is not reproduced in debug, no bb with jump to itself at all ((
Kirill

Issue #5763, opened 2022-11-25
Enable tests in the Ubuntu_22 suite


abhinav92003:
Today only a handful tests are run in our Ubuntu-22 x86-64 test suite due to issues like #5437 and other undocumented ones like some failures on the scatter-gather tests: https://github.com/DynamoRIO/dynamorio/blob/aa99f0988358a42d5814b4f91cb5450e64c148cc/suite/tests/CMakeLists.txt#L5401

This issue covers enabling all tests, and also adding a workflow for x86-32.

PR #5764, opened 2022-11-25
i#5763 ubuntu 22: fix gather into xmm test

abhinav92003:
Fixes "gather result does not match" issue in scatter/gather tests. This was
because vpgatherqd and vgatherqps write only the first half of the xmm
reg when the index is an xmm reg, so the remaining part needs to be
explicitly initialized to zero.

With this fix, the scatter/gather tests pass on my local machine. On
Github Ubuntu-22 CI however, there is an unresolved SIGSEGV that
will be resolved separately. Added details to i#5763.

Issue: #5763


abhinav92003:
`sample.memval_simple_scattergather` and `client.drx-scattergather` now are showing a crash on the Ubuntu22 CI whereas `tool.drcachesim.scattergather` is showing #5329.

abhinav92003:
> `sample.memval_simple_scattergather` and `client.drx-scattergather` now are showing a crash on the Ubuntu22 CI whereas `tool.drcachesim.scattergather` is showing #5329.

None of these failures reproduce on my machine.

abhinav92003:
I was able to reproduce the `sample.memval_simple_scattergather` SIGSEGV on a Ubuntu-22 VM. Also found that the crash is likely not related to scatter-gather itself, since the following crashes too:

```
./bin64/drrun -c /home/abhinav92003/dr/build/i5763-1/api/bin/libmemval_simple.so -- suite/tests/bin/simple_app
```

abhinav92003:
It's the `fdopen` inside the `log_stream_from_file` that's crashing with a SIGSEGV:
https://github.com/DynamoRIO/dynamorio/blob/aa99f0988358a42d5814b4f91cb5450e64c148cc/api/samples/memval_simple.c#L434

abhinav92003:
Looking at `dmesg`, the error is:
```
[ 4387.356408] simple_app[70431] vsyscall read attempt denied -- look up the vsyscall kernel parameter if you need a workaround ip:7f7529310758 cs:33 sp:7ffded274bf8 ax:7ffded274cd0 si:ffffffffff600000 di:7ffded274cd0
```

abhinav92003:
Here's the symbolized stacktrack for the crash:

```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff775b253 in __GI___libc_cleanup_push_defer (buffer=buffer@entry=0x7fffffffc3b0) at ./nptl/libc-cleanup.c:30
30      ./nptl/libc-cleanup.c: No such file or directory.
(gdb) bt
#0  0x00007ffff775b253 in __GI___libc_cleanup_push_defer (buffer=buffer@entry=0x7fffffffc3b0) at ./nptl/libc-cleanup.c:30
#1  0x00007ffff77573ce in __GI__IO_link_in (fp=fp@entry=0x7ffdb3c32050) at ./libio/genops.c:92
#2  0x00007ffff7755eae in _IO_new_file_init_internal (fp=fp@entry=0x7ffdb3c32050) at ./libio/fileops.c:113
#3  0x00007ffff774903e in _IO_new_fdopen (fd=fd@entry=1048480, mode=<optimized out>, mode@entry=0x7fffb3b5329a "w") at ./libio/iofdopen.c:139
#4  0x00007fffb3b52bec in log_stream_from_file (f=1048480) at /home/abhinav92003/dr/src/i5763-1/api/samples/utils.c:110
#5  0x00007fffb3b52506 in event_thread_init (drcontext=0x7ffdb3bdf080) at /home/abhinav92003/dr/src/i5763-1/api/samples/memval_simple.c:439
#6  0x00007fffb3ba5ab6 in drmgr_thread_init_event (drcontext=0x7ffdb3bdf080) at /home/abhinav92003/dr/src/i5763-1/ext/drmgr/drmgr.c:2636
#7  0x00007ffff7d5603a in instrument_thread_init (dcontext=0x7ffdb3bdf080, client_thread=false, valid_mc=false) at /home/abhinav92003/dr/src/i5763-1/core/lib/instrument.c:1474
#8  0x00007ffff7d54702 in instrument_init () at /home/abhinav92003/dr/src/i5763-1/core/lib/instrument.c:786
#9  0x00007ffff7bb734a in dynamorio_app_init_part_two_finalize () at /home/abhinav92003/dr/src/i5763-1/core/dynamo.c:716
#10 0x00007ffff7e75f86 in privload_early_inject (sp=0x7fffffffe390, old_libdr_base=0x0, old_libdr_size=1) at /home/abhinav92003/dr/src/i5763-1/core/unix/loader.c:2245
#11 0x00007ffff7e1c263 in reloaded_xfer () at /home/abhinav92003/dr/src/i5763-1/core/arch/x86/x86.asm:1179
#12 0x0000000000000001 in ?? ()
#13 0x00007fffffffe612 in ?? ()
#14 0x0000000000000000 in ?? ()
```

suite/tests/client-interface/drx-scattergather.c
```diff
@@ -791,14 +791,20 @@ DECLARE_FUNC_SEH(FUNCNAME(opcode))                        @N@\
...
+         * is set by the opcode. So we need to
+         * explicitly zero the remaining part.
+         */                                               @N@\
+        vpxor      xmm0, xmm0, xmm0                       @N@\
         opcode     xmm0 {k1}, [REG_XAX + xmm1 * 4]        @N@\
         vmovdqu32  [REG_XCX], xmm0                        @N@\
         kmovw      k1, edx                                @N@\
         opcode     xmm0 {k1}, [REG_XAX + ymm1 * 4]        @N@\
-        vmovdqu32  [REG_XCX + 16], ymm0                   @N@\
+        vmovdqu32  [REG_XCX + 16], xmm0                   @N@\
```
derekbruening, line 804: So this is now storing 128 bits instead of the 256 it was storing -- why?  And why is the next offset still 32 bytes further (+48 from +16) as though this were still storing 256 bits?

derekbruening (approved):

suite/tests/client-interface/drx-scattergather.c
abhinav92003: It should always have been storing 128 bit, since that's all that's generated by the instruction above (note that xmm dest).

About the offset: note that it's an offset into a (xmm + ymm + zmm) sized buffer, which is checked for equality after this asm routine is done (see ref_idx64_val32_xmm_ymm_zmm). The three parts correspond to the result of the three executed opcodes (the ones with xmm, ymm and zmm index regs). In this case, we could defrag ref_idx64_val32_xmm_ymm_zmm to remove the zeroes, but I'm choosing uniformity with other tests.

abhinav92003: Rethinking this... it doesn't hurt to have a stronger check, especially since some vector operations are expected to zero out the larger regs too. Reverted this to ymm and the one below to zmm.

PR #5768, opened 2022-11-30
i#2039 trace trim, part 5: Drop early data

derekbruening:
Part 4 added a minimum timestamp, but it only works for data that is not actually emitted until we are fully attached.  For data emitted before that point, we'll use a too-early timestamp.  We solve that here by dropping such data.

Tested on a large app where these changes solve gaps at the start of the trace due to an early timestamp in one thread.

Issue: #2039


derekbruening:
Update: it turns out this is not needed on the big app in question; there was some other error in prior results.  But this is still a theoretical issue.  It is not easy to make an automated test unfortunately.

abhinav92003 (approved):

PR #5770, opened 2022-12-01
i#2039 trace trim, part 6: Enable for AArchXX

derekbruening:
Since the stability issues with drbbdup on larger apps seem limited to x86, we enable -align_endpoints by default for AArchXX.

Issue: #2039


derekbruening:
x32 failure is scattergather pipe assert #5329

clients/drcachesim/tests/offline-burst_threadfilter.templatex
```diff
@@ -21,7 +21,7 @@ Total counts:
     *[0-9]* total data stores
     *[0-9]* total icache flushes
     *[0-9]* total dcache flushes
-           4 total threads
```
abhinav92003, line 24: Since we have your `burst_replaceall.cpp` fix to make threads wait for attach to complete, won't we surely see all threads? Why did we need to relax this?

clients/drcachesim/tests/offline-burst_replace.templatex
```diff
@@ -6,8 +6,7 @@ create dir .*
 open file .*
 open file .*
 open file .*
-open file .*
```
abhinav92003, line 9: The number of "open file" logs == the number of threads traced, right? So now we have 3, from the previous 4?

clients/drcachesim/CMakeLists.txt
```diff
@@ -41,6 +41,25 @@ else ()
   set(os_name "unix")
```
abhinav92003, line 41: > Since the stability issues with drbbdup on larger apps seem limited to x86, we enable -align_endpoints by default for AArchXX.

So did we do more tests on AArch64 with `-align_endpoints` actually on? And found that it is indeed more stable on AArch64 than X86?

abhinav92003 (approved):

clients/drcachesim/tests/offline-burst_threadfilter.templatex
derekbruening: This is a different test, burst_threadfilter.cpp.  It is more complex and looked harder to add the same type of synchronization so I went with the relax-template approach.

clients/drcachesim/tests/offline-burst_replace.templatex
derekbruening: With lazy opening the 4th is now after the `pre-DR start` line -- but only for arm since align_endpoints is still off for x86!  A bunch of these templates have the preprocessor off so I wasn't sure I could easily add ifdefs (or comments) so I just relaxed it to handle either order.

clients/drcachesim/tests/offline-burst_threadfilter.templatex
abhinav92003: Ah, missed that simple fact. Thanks.

clients/drcachesim/CMakeLists.txt
derekbruening: Yes, a (proprietary) detach stress test has a 100% pass rate on arm with this option on, yet fails consistently ~2% of the time on x86.  A larger (proprietary) app has passed a bunch of times in a row with this option on arm where on x86 it fails about half the time.  I am now focusing on x86-specific aspects of bbdup to try to figure out these stability problems -- haven't found the issue yet.

PR #5772, opened 2022-12-02
i#5675: Fix data race in record_filter.

abhinav92003:
Fixes a race in updating input_entry_count_ and output_entry_count_
in parallel_shard_exit in the record_filter analyzer.

Moves aggregation logic to print_results instead. Like other analyzers,
we now keep track of all shard data pointers and free them in the
destructor instead. We destroy the writer object in parallel_shard_exit
itself though, for symmetry with how it's created (in parallel_shard_stream_init)
and so that users don't have to destroy the record_filter_t object itself
for output files to be written out.

Issue: #5675

### DIFF
--- a/clients/drcachesim/tools/filter/record_filter.cpp
+++ b/clients/drcachesim/tools/filter/record_filter.cpp
@@ -128,6 +128,7 @@ bool
 record_filter_t::parallel_shard_exit(void *shard_data)
 {
     per_shard_t *per_shard = reinterpret_cast<per_shard_t *>(shard_data);
+    std::lock_guard<std::mutex> guard(shard_exit_mutex_);
     input_entry_count_ += per_shard->input_entry_count;
     output_entry_count_ += per_shard->output_entry_count;
     bool res = true;
--- a/clients/drcachesim/tools/filter/record_filter.cpp
+++ b/clients/drcachesim/tools/filter/record_filter.cpp
@@ -67,8 +67,6 @@ record_filter_t::record_filter_t(
     , filters_(std::move(filters))
     , stop_timestamp_(stop_timestamp)
     , verbosity_(verbose)
-    , input_entry_count_(0)
-    , output_entry_count_(0)
 {
     UNUSED(verbosity_);
     UNUSED(output_prefix_);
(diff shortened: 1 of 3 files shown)


abhinav92003:
Created #5774 for the AArch64 failure. It's definitely unrelated to this PR.

clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -128,8 +128,11 @@ bool
 record_filter_t::parallel_shard_exit(void *shard_data)
 {
     per_shard_t *per_shard = reinterpret_cast<per_shard_t *>(shard_data);
-    input_entry_count_ += per_shard->input_entry_count;
-    output_entry_count_ += per_shard->output_entry_count;
+    {
+        std::lock_guard<std::mutex> guard(shard_exit_mutex_);
```
derekbruening, line 132: Aggregation should be done at the end by a single thread, avoiding any need for synchronization points during parallel operation, which can ruin performance with no benefit.  All of our existing tools operate this way: they do not delete per-shard data in shard_exit, keeping it for aggregation in print_results or wherever.

abhinav92003: Synchronization in parallel_shard_exit may not be that bad. Since this lock is acquired only a few times (#shard or #threads-in-trace).

Delaying freeing of per-shard data and performing additional book-keeping for remembering the per_shard ptrs doesn't seem ideal either.

abhinav92003: Though remembering per_shard ptrs has the benefit of also allowing retrieving per-shard outputs.

abhinav92003: Done.

clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -121,21 +122,24 @@ record_filter_t::parallel_shard_init_stream(int shard_index, void *worker_data,
             success_ = false;
         }
     }
+    std::lock_guard<std::mutex> guard(shard_map_mutex_);
```
derekbruening, line 125: (Hmm, forgot that the tools are all using locks on parallel init -- makes the lock on exit not seem so bad.)

derekbruening (approved):

clients/drcachesim/tools/filter/record_filter.cpp
abhinav92003: Tracking shards is helpful as the tool can now produce per-shard results.

PR #5776, opened 2022-12-06
i#3044: AArch64 SVE codec: add LDR and STR

AssadHashmi:
This patch adds the appropriate macros, tests and codec entries to decode and encode the following instructions:
LDR <Zt>, [<Xn|SP>{, #<imm>, MUL VL}]
LDR <Pt>, [<Xn|SP>{, #<imm>, MUL VL}]
STR <Zt>, [<Xn|SP>{, #<imm>, MUL VL}]
STR <Pt>, [<Xn|SP>{, #<imm>, MUL VL}]

Issue: #3044


core/ir/aarch64/codec.c
```diff
@@ -3652,6 +3671,52 @@ encode_opnd_vindex_H(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_
...
+#    define OPSZ_SVE opnd_size_from_bytes(32)
+#else
+/* i3044 TODO: How do we set SVE vector length for offline decode?
+ * A command line option?
+ */
+#    define OPSZ_SVE opnd_size_from_bytes(32)
+#endif
+
+static inline bool
+decode_opnd_svemem_gpr_simm(uint enc, int opcode, byte *pc, OUT opnd_t *opnd)
```
joshua-warburton, line 3691: There are gpr_simm signatures without a mul vl operand, so this should reflect it is mul vl in the name

core/ir/aarch64/codec.c
```diff
@@ -3652,6 +3671,52 @@ encode_opnd_vindex_H(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_
     return true;
 }
 
+/* svemem_gpr_simm: 9 bit signed immediate offset added to base register defined in bits 5
+ * to 9.
+ */
+#if !defined(DR_HOST_NOT_TARGET) && !defined(STANDALONE_DECODER)
```
AssadHashmi, line 3677: Hi @derekbruening, there is an aspect of this SVE memory patch about which I would like clarification and advice.

The number of bytes in memory transfer instructions will depend on the size of the scalable vector on the SVE host. For the purposes of unit testing on non-SVE hardware including x86 we have fixed the vector length to 256 bits. 

This `#ifdef` clause sets `OPSZ_SVE_VL` for:
- use of SVE instructions in tools, clients, tracing etc, a value read from host h/w using `proc_get_vector_length()`, commented out for now (a patch will follow shortly).
- unit testing, a fixed value.

For the latter, how can we set the vector length for offline drdecode?
Some config or command line option or hard coded value?

derekbruening: There is another similar issue with determining a runtime host parameter for drdecode -- let me find it to see if we had a solution.

derekbruening: It was `dc zva`'s size which comes from the cache line size: #4400; PR #5265.

Generally for any of these "global" inputs to an instruction we'd have to add an API routine (similar to `dr_set_isa_mode`) that can be used to set this context for drdecode and we would probably also want a command line option for `drdisas`.  Maybe the only question is whether, if we are starting to accumulate multiple of them, we should combine them into an ISA config struct rather than having completely separate set and get routine?

derekbruening: (The `dr_set_isa_mode` example is similar for ARM vs Thumb)

derekbruening: In PR #5265 we proposed `set_dcache_zva_size()` for that one.  If we would prefer one config struct to make it easier to add more that makes sense to me; so we'd have something like `dr_set_isa_config(dr_isa_config_t *config)` or sthg with a size field so we can append.

derekbruening: Direct link: https://github.com/DynamoRIO/dynamorio/pull/5265/files#r774769132

Probably we leave `dr_set_isa_mode` separate?  Or maybe there is also a mode field in the struct.

derekbruening: @abhinav92003 for further opinions

core/ir/aarch64/codec.h
```diff
@@ -57,4 +57,21 @@ encode_common(byte *pc, instr_t *i, decode_info_t *di);
 #define BITS(_enc, bitmax, bitmin) \
     ((((uint32)(_enc)) >> (bitmin)) & (uint32)MASK((bitmax) - (bitmin) + 1))
 
+#if !defined(DR_HOST_NOT_TARGET) && !defined(STANDALONE_DECODER)
+/* i3044 TODO: Vector length will be read from cpuinfo, e.g.
+ * opnd_size_from_bytes(proc_get_vector_length()));
+ * Setting to fixed size for now in order to pass unit tests.
+ */
+#    define OPSZ_SVE_VL opnd_size_from_bytes(dr_get_sve_vl() / 8)
```
abhinav92003, line 65: Both these #defines are the same: they call dr_get_sve_vl(). Do we need to separate them?

core/ir/decode_shared.c
```diff
@@ -174,6 +174,33 @@ const char *const size_names[] = {
...
+/* AArch64 Scalable Vector Extension's vector length in bits. */
+int sve_veclen;
+int sve_veclens[] = { 128,  256,  384,  512,  640,  768,  896,  1024,
+                      1152, 1280, 1408, 1536, 1664, 1792, 1920, 2048 };
+
+void
+dr_set_sve_vl(int vl)
+{
+    /* i3044 TODO: Vector length will be read from h/w when running on SVE.
+     * CLIENT_ASSERT(!proc_has_feature(FEATURE_SVE),
```
abhinav92003, line 186: Is this CLIENT_ASSERT commented-out because we're hitting it now?

core/ir/decode_shared.c
```diff
@@ -174,6 +174,33 @@ const char *const size_names[] = {
...
 
+/* AArch64 Scalable Vector Extension's vector length in bits. */
+int sve_veclen;
+int sve_veclens[] = { 128,  256,  384,  512,  640,  768,  896,  1024,
+                      1152, 1280, 1408, 1536, 1664, 1792, 1920, 2048 };
+
+void
+dr_set_sve_vl(int vl)
+{
+    /* i3044 TODO: Vector length will be read from h/w when running on SVE.
```
abhinav92003, line 185: s/i3044 TODO/TODO i#3044/

suite/tests/api/ir_aarch64.h
```diff
@@ -127,6 +127,15 @@ test_instr_encoding(void *dc, uint opcode, instr_t *instr, const char *expected)
     return result;
 }
 
+const reg_id_t Xn_six_offset_0[6] = { DR_REG_X0,  DR_REG_X5,  DR_REG_X10,
+                                      DR_REG_X15, DR_REG_X20, DR_REG_X30 };
+const reg_id_t Xn_six_offset_1[6] = { DR_REG_X0,  DR_REG_X6,  DR_REG_X11,
```
abhinav92003, line 132: How do we decide which regs go in each of these Xn_six_offset_* arrays? Do we intend for them to cover all DR_REG_Xn (because they don't currently)?

suite/tests/api/ir_aarch64.h
```diff
@@ -127,6 +127,15 @@ test_instr_encoding(void *dc, uint opcode, instr_t *instr, const char *expected)
...
+                                      DR_REG_X17, DR_REG_X22, DR_REG_X30 };
+const reg_id_t Xn_six_offset_2[6] = { DR_REG_X0,  DR_REG_X7,  DR_REG_X12,
```
abhinav92003, line 134: Xn_six_offset_3 and Xn_six_offset_2 seem unused.

core/ir/aarch64/codec.c
abhinav92003: Should be fine for now with a TODO.

abhinav92003 (approved):

suite/tests/api/ir_aarch64.h
AssadHashmi: Oops! Deleted.

suite/tests/api/ir_aarch64.h
AssadHashmi: These are not exhaustive register sets. They're a selection based on our experience with test coverage.
More will be added later to improve coverage.

core/ir/decode_shared.c
AssadHashmi: Done.

core/ir/decode_shared.c
AssadHashmi: No, it's what will _probably_ be required as part of upcoming patches to enable execution on SVE h/w.
As it's commented out code, I'll delete to avoid confusion.

core/ir/aarch64/codec.h
AssadHashmi: The #define for this #if clause (running on AArch64 and SVE h/w) will change. It's the same as the other #define temporarily so that tests on non-SVE AArch64 h/w pass.

core/globals.h
```diff
@@ -297,6 +297,13 @@ typedef struct _thread_record_t {
 #    define DYNAMORIO_EXPORT DR_APP_API
 #endif
 
+/* AArch64 Scalable Vector Extension's vector length in bits. This depends on
+ * the hardware implementation and can be one of:
+ * 128 256 384 512 640 768 896 1024 1152 1280 1408 1536 1664 1792 1920 2048
```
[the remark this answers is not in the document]
AssadHashmi: Yes, added relevant link to SVE intro at arm.developer.com

core/arch/aarch64/proc.c
```diff
@@ -87,6 +87,22 @@ get_processor_specific_info(void)
     cpu_info.features.flags_aa64pfr0 = isa_features[AA64PFR0];
     cpu_info.features.flags_aa64mmfr1 = isa_features[AA64MMFR1];
     cpu_info.features.flags_aa64dfr0 = isa_features[AA64DFR0];
+
+#        if !defined(DR_HOST_NOT_TARGET) && defined(SVE)
+    /* i3044 TODO: Vector length will be set by reading value from h/w, e.g.
+     * if (proc_has_feature(FEATURE_SVE)) {
```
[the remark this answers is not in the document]
AssadHashmi: I'm happy to remove this commented out code.

core/arch/aarch64/proc.c
```diff
@@ -87,6 +87,22 @@ get_processor_specific_info(void)
     cpu_info.features.flags_aa64pfr0 = isa_features[AA64PFR0];
     cpu_info.features.flags_aa64mmfr1 = isa_features[AA64MMFR1];
     cpu_info.features.flags_aa64dfr0 = isa_features[AA64DFR0];
+
+#        if !defined(DR_HOST_NOT_TARGET) && defined(SVE)
+    /* i3044 TODO: Vector length will be set by reading value from h/w, e.g.
```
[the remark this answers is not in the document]
AssadHashmi: Done.

core/arch/aarch64/proc.c
```diff
@@ -87,6 +87,22 @@ get_processor_specific_info(void)
...
+     * if (proc_has_feature(FEATURE_SVE)) {
+     *     <read length from h/w>
+     *     cpu_info.vector_length = vl;
+     * }
+     * else
```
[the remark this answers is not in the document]
AssadHashmi: Commented out code removed.

core/ir/aarch64/codec.c
AssadHashmi: Done.

derekbruening:
This PR introduced a buffer overflow: #6584

PR #5777, opened 2022-12-06
i#5675 record filter: Adjust trace file type

abhinav92003:
Adjusts the trace's file type based on the entries being filtered out.

Removes the const property for the trace_entry_t argument to
parallel_shard_filter, which allows record_filter_func_t to modify the entry.

Tests operation in the record_filter_unit_tests.

Issue: #5675

### DIFF
--- a/clients/drcachesim/tools/filter/record_filter.h
+++ b/clients/drcachesim/tools/filter/record_filter.h
@@ -70,6 +70,18 @@ public:
         virtual void *
         parallel_shard_init(memtrace_stream_t *shard_stream) = 0;
         /**
+         * Invoked for the TRACE_TYPE_MARKER entry of type
+         * TRACE_MARKER_TYPE_FILETYPE. This routine can make changes to the file
+         * type described by the marker, based on the type of entries it filters out.
+         * The passed marker is not guaranteed to be the original one in the trace
+         * if other filter tools are present, and may include changes made by other
+         * tools. \p partial_trace_filter is set if the trace will be filtered only
+         * partially, e.g. due to stop_timestamp.
+         */
+        virtual void
+        parallel_shard_process_file_type(trace_entry_t &entry, void *shard_data,
+                                         bool partial_trace_filter) = 0;
+        /**
          * Invoked for each #trace_entry_t in the shard. It returns
          * whether or not this \p entry should be included in the result
          * trace. \p shard_data is same as what was returned by
--- a/clients/drcachesim/tools/filter/record_filter.h
+++ b/clients/drcachesim/tools/filter/record_filter.h
@@ -66,30 +66,25 @@ public:
          * any entry. The returned pointer is passed to all invocations of
          * parallel_shard_filter() and parallel_shard_exit().
          * This routine can be used to initialize state for each shard.
+         * \p partial_trace_filter denotes whether the trace will be filtered
+         * only partially, e.g. due to stop_timestamp.
          */
         virtual void *
-        parallel_shard_init(memtrace_stream_t *shard_stream) = 0;
-        /**
-         * Invoked for the TRACE_TYPE_MARKER entry of type
-         * TRACE_MARKER_TYPE_FILETYPE. This routine can make changes to the file
-         * type described by the marker, based on the type of entries it filters out.
-         * The passed marker is not guaranteed to be the original one in the trace
-         * if other filter tools are present, and may include changes made by other
-         * tools. \p partial_trace_filter is set if the trace will be filtered only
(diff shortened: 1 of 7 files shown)


clients/drcachesim/tools/filter/record_filter.h
```diff
@@ -69,6 +69,18 @@ class record_filter_t : public record_analysis_tool_t {
...
+         * Invoked for the TRACE_TYPE_MARKER entry of type
+         * TRACE_MARKER_TYPE_FILETYPE. This routine can make changes to the file
+         * type described by the marker, based on the type of entries it filters out.
+         * The passed marker is not guaranteed to be the original one in the trace
+         * if other filter tools are present, and may include changes made by other
+         * tools. \p partial_trace_filter is set if the trace will be filtered only
+         * partially, e.g. due to stop_timestamp.
+         */
+        virtual void
+        parallel_shard_process_file_type(trace_entry_t &entry, void *shard_data,
```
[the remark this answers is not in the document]
derekbruening: Changing each entry sounds better than a special interface just for the file type.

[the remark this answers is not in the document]
abhinav92003: Done.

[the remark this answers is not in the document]
abhinav92003: PTAL.

clients/drcachesim/tools/filter/type_filter.h
```diff
@@ -78,13 +78,33 @@ class type_filter_t : public record_filter_t::record_filter_func_t {
...
+        return per_shard;
     }
     bool
-    parallel_shard_filter(const trace_entry_t &entry, void *shard_data) override
+    parallel_shard_filter(trace_entry_t &entry, void *shard_data) override
     {
+        per_shard_t *per_shard = reinterpret_cast<per_shard_t *>(shard_data);
+        if (entry.type == TRACE_TYPE_MARKER && entry.size == TRACE_MARKER_TYPE_FILETYPE) {
+            if (TESTANY(entry.addr, OFFLINE_FILE_TYPE_ENCODINGS) &&
+                !per_shard->partial_trace_filter &&
```
derekbruening, line 94: Not sure I understand this: if we filter the first half of a trace and remove all the encodings, don't we still want to remove the encoding flag from the type?  The partial trace will be output as a whole new trace and its type should be the same as when we filter the entire trace, right?

clients/drcachesim/tools/filter/record_filter.h
```diff
@@ -66,18 +66,25 @@ class record_filter_t : public record_analysis_tool_t {
          * any entry. The returned pointer is passed to all invocations of
          * parallel_shard_filter() and parallel_shard_exit().
          * This routine can be used to initialize state for each shard.
+         * \p partial_trace_filter denotes whether the trace will be filtered
+         * only partially, e.g. due to stop_timestamp.
          */
         virtual void *
-        parallel_shard_init(memtrace_stream_t *shard_stream) = 0;
+        parallel_shard_init(memtrace_stream_t *shard_stream,
+                            bool partial_trace_filter) = 0;
```
derekbruening, line 74: Should this be a flag so we can add more later -- but see below: I'm not sure I understand the use case for this.

derekbruening (approved):

clients/drcachesim/tools/filter/type_filter.h
abhinav92003: > if we filter the first half of a trace and remove all the encodings, don't we still want to remove the encoding flag from the type?

Since there still are encodings present in the trace, I thought we should keep the OFFLINE_FILE_TYPE_ENCODINGS bit.

> The partial trace will be output as a whole new trace

No the partial trace is still a part of the overall trace, with the second half still having encodings.

clients/drcachesim/tools/filter/record_filter.h
abhinav92003: Not sure what you mean by flag here?

derekbruening: A bitmask value inside an integer var so we could add more w/o adding more bool params.  Are we sure we won't want more qualifiers later?

abhinav92003: Or even use default args.

derekbruening: I don't feel strongly either/any way.

abhinav92003: I'll leave it to later then, since currently I don't see a requirement.

Issue #5771, opened 2022-12-02
can not detect self modify code range on aarch64 v9


kuaiwei:
**Describe the bug**
see discussion in https://groups.google.com/g/dynamorio-users/c/72wz2q-njCg

When trace java application on arm v9 hardware, DynamoRIO does not know the jit code change. It will cause strange failure. The root cause is ic_ivau instruction is not used in new aarch64 platform for icache flushing.

**To Reproduce**
1. Use a debug build DynamoRIO to run with java command , set the log level to 4
```bash
export DYNAMO_HOME=<Debug build>
export JAVA_HOME=<java install>
$DYNAMO_HOME/bin64/drrun -root $DYNAMO_HOME -verbose -debug -loglevel 4 -c  $DYNAMO_HOME/api/bin/libbbsize.so -- $JAVA_HOME/bin/jps
```
3. check log output to find mangled code stub for isb and ic_ivau
```bash
grep ic_ivau $DYNAMO_HOME/logs/jps*/*
grep isb $DYNAMO_HOME/logs/jps*/*
```
4. Exact output or incorrect behavior.
For ic_ivau instruction, no output.
For isb instruction, we can find log like:
```
jps.292482.00000001/log.1.292483.html:  0x0000ffffa2492554  d5033fdf   isb    $0x0f
jps.292482.00000001/log.1.292483.html: +0    L3 @0x0000fffd5e7c4098  d5033fdf   isb    $0x0f
jps.292482.00000001/log.1.292483.html: +0    L3 @0x0000fffd5e7c4098  d5033fdf   isb    $0x0f
jps.292482.00000001/log.1.292483.html:forward_eflags_analysis: isb    $0x0f
jps.292482.00000001/log.1.292483.html: +0    L3 @0x0000fffd5e7c4098  d5033fdf   isb    $0x0f
jps.292482.00000001/log.1.292483.html:  0x0000ffff5e6a6868  d5033fdf   isb    $0x0f
jps.292482.00000001/log.1.292483.html:  0x0000ffffa0c86824  d5033fdf   isb    $0x0f
```
In log file, the isb mangle code is:

```raw
before instrumentation:
TAG  0x0000ffffa2492400
 +0    L3 @0x0000fffd5e7c2fc8  d5033fdf   isb    $0x0f
END 0x0000ffffa2492400


after instrumentation:
TAG  0x0000ffffa2492400
[...]

PR #5785, opened 2022-12-13
i#5771 mangle mrs ctr.el0 for aarch64

kuaiwei:
ic_ivau is not required for new aarch64 hardware. Dynamorio depends this instruction to detect self modify code region. So the region may be skipped in new hardware.

This fix mangle "mrs ctr.el0" to disable DIC, so ic_ivau can be executed.

JVM sometimes flush empty stubs, check the range before flush fragments

Fix: #5771

### DIFF
--- a/core/arch/aarchxx/mangle.c
+++ b/core/arch/aarchxx/mangle.c
@@ -74,12 +74,15 @@ typedef struct ALIGN_VAR(16) _icache_op_struct_t {
      * consecutive icache lines (which could wrap around the top of memory).
      */
     void *begin, *end;
+    /* cache the ctr el0 at init */
+    ptr_uint_t cached_ctr_el0;
     /* Some space to spill registers. */
     ptr_uint_t spill[2];
 } icache_op_struct_t;
 
 /* Used in aarch64.asm. */
 icache_op_struct_t icache_op_struct;
+#define CTR_DIC_SHIFT 29
 #endif
 
 void
@@ -88,7 +91,19 @@ mangle_arch_init(void)
 #ifdef AARCH64
     /* Check address of "lock" is unaligned. See comment in icache_op_struct_t. */
     ASSERT(!ALIGNED(&icache_op_struct.lock, 16));
+    /* cache the ctr_el0 */
+    ptr_uint_t cache_info;
+    asm volatile ("mrs\t%0, ctr_el0":"=r" (cache_info));
+    LOG(GLOBAL, LOG_INTERP, 2, "cpu cache info: %x\n", cache_info);
+    icache_op_struct.cached_ctr_el0 = cache_info;
 #endif
+}
+
+static bool
+icache_sync_by_hardware() {
+    ASSERT(icache_op_struct.cached_ctr_el0 != 0);
+    /* if CTR_EL0.DIC is set, icache is sync by hardware */
+    return ((icache_op_struct.cached_ctr_el0 >> CTR_DIC_SHIFT) & 0x1) == 0x1;
 }
 
 void
@@ -2977,7 +2992,31 @@ mangle_icache_op(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
         PRE(ilist, instr, /* ldr x0, [x28] */
             instr_create_restore_from_tls(dcontext, DR_REG_X0, TLS_REG0_SLOT));
         /* Leave original instruction. */
-    } else
+    } else if (opc == OP_mrs) {
+        DEBUG_DECLARE(reg_id_t src_reg = opnd_get_reg(instr_get_src(instr, 0)));
+        ASSERT(src_reg == DR_REG_CTR_EL0);
+        if (icache_sync_by_hardware()) {
+            ptr_int_t mangle_ctr_el0 = icache_op_struct.cached_ctr_el0 & (~(0x1l << CTR_DIC_SHIFT));
+            LOG(THREAD, LOG_INTERP, 2, "mangle ctr_el0 as %x -> %x\n",
+                icache_op_struct.cached_ctr_el0, mangle_ctr_el0);
+            reg_id_t reg = opnd_get_reg(instr_get_dst(instr, 0));
+            if (reg != dr_reg_stolen) {
+                insert_mov_immed_arch(dcontext, NULL, NULL, mangle_ctr_el0,
+                                      opnd_create_reg(reg), ilist, instr, NULL, NULL);
+            } else {
+                PRE(ilist, instr, instr_create_save_to_tls(dcontext, DR_REG_X0, TLS_REG0_SLOT));
+                insert_mov_immed_arch(dcontext, NULL, NULL, mangle_ctr_el0,
+                                      opnd_create_reg(DR_REG_X0), ilist, instr, NULL, NULL);
+                PRE(ilist, instr,
+                    instr_create_save_to_tls(dcontext, DR_REG_X0, TLS_REG_STOLEN_SLOT));
+                PRE(ilist, instr,
+                    instr_create_restore_from_tls(dcontext, DR_REG_X0, TLS_REG0_SLOT));
+            }
+            /* remove origin instr */
+            instrlist_remove(ilist, instr);
+            instr_destroy(dcontext, instr);
+        }
+    }else
         ASSERT_NOT_REACHED();
     return next_instr;
 }
--- a/core/dispatch.c
+++ b/core/dispatch.c
@@ -941,9 +941,13 @@ dispatch_enter_dynamorio(dcontext_t *dcontext)
             app_pc begin = (app_pc)dcontext->local_state->spill_space.r2;
             app_pc end = (app_pc)dcontext->local_state->spill_space.r3;
             dcontext->next_tag = (app_pc)dcontext->local_state->spill_space.r4;
-            flush_fragments_from_region(dcontext, begin, end - begin, true,
-                                        NULL /*flush_completion_callback*/,
-                                        NULL /*user_data*/);
+            if (end > begin) {
+              flush_fragments_from_region(dcontext, begin, end - begin, true,
+                                          NULL /*flush_completion_callback*/,
+                                          NULL /*user_data*/);
+            } else {
+              LOG(THREAD, LOG_DISPATCH, 2, "WARNING: skip selfmod region from %llx to %llx, size %d\n", begin, end, end-begin);
+            }
         }
 #endif
 
--- a/core/ir/aarch64/instr.c
+++ b/core/ir/aarch64/instr.c
@@ -288,6 +288,10 @@ instr_is_icache_op(instr_t *instr)
         return true; /* ic ivau, xT */
     if (opc == OP_isb)
         return true; /* isb */
+    if (opc == OP_mrs &&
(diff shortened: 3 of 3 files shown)


AssadHashmi:
run arm tests

kuaiwei:
I can not trigger aarch64 precommit test. May I have volunteer to trigger it.
Thanks

derekbruening:
add to allowlist

derekbruening:
run arm tests

core/arch/aarchxx/mangle.c
```diff
@@ -88,9 +90,33 @@ mangle_arch_init(void)
 #ifdef AARCH64
     /* Check address of "lock" is unaligned. See comment in icache_op_struct_t. */
     ASSERT(!ALIGNED(&icache_op_struct.lock, 16));
+#    ifdef DR_HOST_NOT_TARGET
```
bete0, line 93: nit: The #ifdef should be aligned at the beginning of the line. See line 88.

core/arch/aarchxx/mangle.c
```diff
@@ -88,9 +90,33 @@ mangle_arch_init(void)
...
+#    endif
 #endif
 }
 
+#ifdef AARCH64
+#    define CTR_DIC_SHIFT 29
+static bool
+icache_sync_by_hardware()
+{
+#    ifdef DR_HOST_NOT_TARGET
```
bete0, line 110: Same comment as line 93 here.

bete0:
@AssadHashmi Do you have some bandwidth to take a look at this PR?

core/arch/aarchxx/mangle.c
derekbruening: This is actually the correct style: we indent nested ifdefs in this manner (# still on 1st column).  This is enforced by clang-format so any deviation here would fail the tests (unless inside one of the places where we disable clang-format but this is not one of them).

kuaiwei: Yes, it's enforced by clang-format.

core/arch/aarchxx/mangle.c
kuaiwei: It's enforced by clang-format

AssadHashmi (approved):
Apologies for the delay in reviewing @bete0 and @derekbruening.

Based on my limited understanding of code cache operation, this looks fine to me, nothing stands out as obviously wrong. The worrying thing about this issue is that features like DIC can be changed, cause errors and only found after much debugging, (thank you @kuaiwei).

Perhaps we should identify a set of key parameters in the AArch64 system registers which affect DR's operation on different types of h/w and check them against expected reference values on startup so that we get an early warning of possible causes of faulty operation?

derekbruening:
Should we put this fooling of the app by changing the system register under a runtime option, so that if some app doesn't work properly we can try disabling this?  Even if the app later fails due to cache consistency issues b/c we haven't made that work in such a case for a64 yet, it would be a different error and still prove useful.  Plus, even once we do implement handling hw consistency for a64, we may still want the option of fooling the DIC.

AssadHashmi:
> Should we put this fooling of the app by changing the system register under a runtime option, so that if some app doesn't work properly we can try disabling this? Even if the app later fails due to cache consistency issues b/c we haven't made that work in such a case for a64 yet, it would be a different error and still prove useful. Plus, even once we do implement handling hw consistency for a64, we may still want the option of fooling the DIC.

Yes, I think this is a safe and more flexible approach. A bit like the use of `--unsafe-ldstex` we used to have to handle different AArch64 h/w implementations of the spec.

kuaiwei:
> > Should we put this fooling of the app by changing the system register under a runtime option, so that if some app doesn't work properly we can try disabling this? Even if the app later fails due to cache consistency issues b/c we haven't made that work in such a case for a64 yet, it would be a different error and still prove useful. Plus, even once we do implement handling hw consistency for a64, we may still want the option of fooling the DIC.
> 
> Yes, I think this is a safe and more flexible approach. A bit like the use of `--unsafe-ldstex` we used to have to handle different AArch64 h/w implementations of the spec.

I will add an option to enable/disable it. How about '-unsafe-el0-dic' ?

kuaiwei:
run arm tests

AssadHashmi (approved):

derekbruening:
@kuaiwei making sure you're not waiting on us?  Looks like we did not provide feedback on the name you proposed `-unsafe-el0-dic`: I would suggest adding "disable" in there so maybe `-unsafe_disable_dic`?

kuaiwei:
@AssadHashmi @bete0 Apologize for late response.
I added option "unsafe_fake_el0_dic" in commit https://github.com/DynamoRIO/dynamorio/pull/5785/commits/da7ed6e12037a09d1406b3a47cecbb20d7e6edf9

Because I think we need 3 states (Force 0/Force 1/unchanged) for this option. So I use the name unsafe_fake_el0_dic. Please comments on the change.

derekbruening:
> Please comments on the change.

You should click on the upper right "re-request review" to request another look.  I just did this for you.

kuaiwei:
> > Please comments on the change.
> 
> You should click on the upper right "re-request review" to request another look. I just did this for you.

thanks for your help

derekbruening:
@AssadHashmi has not responded but I can do a review

core/arch/aarchxx/mangle.c
```diff
@@ -74,12 +74,16 @@ typedef struct ALIGN_VAR(16) _icache_op_struct_t {
      * consecutive icache lines (which could wrap around the top of memory).
      */
     void *begin, *end;
+    /* cache the ctr el0 at init */
```
derekbruening, line 77: style nit: UIse capitalization and punctuation in comments, esp whole-line comments.

Ditto at several places below.

core/arch/aarchxx/mangle.c
```diff
@@ -2977,6 +2990,59 @@ mangle_icache_op(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
         PRE(ilist, instr, /* ldr x0, [x28] */
             instr_create_restore_from_tls(dcontext, DR_REG_X0, TLS_REG0_SLOT));
         /* Leave original instruction. */
+    } else if (opc == OP_mrs) {
+        DEBUG_DECLARE(reg_id_t src_reg = opnd_get_reg(instr_get_src(instr, 0)));
+        ASSERT(src_reg == DR_REG_CTR_EL0);
+        ptr_int_t mangle_ctr_el0 = icache_op_struct.cached_ctr_el0;
+        switch (INTERNAL_OPTION(unsafe_fake_el0_dic)) {
+        case 0:
```
derekbruening, line 2998: (We may want an enum for these values: but given that it's hard to share them with optionsx.h and there are only these 3 and 2 match the setting maybe it's fine -- let's leave it.)

core/arch/aarchxx/mangle.c
```diff
@@ -2977,6 +2990,59 @@ mangle_icache_op(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
...
+            mangle_ctr_el0 &= ~(0x1l << CTR_DIC_SHIFT);
+            break;
+        case 1:
+            /* force set as one */
+            mangle_ctr_el0 |= 0x1l << CTR_DIC_SHIFT;
+            break;
+        default: break;
+        }
+        if (mangle_ctr_el0 != icache_op_struct.cached_ctr_el0) {
+            LOG(THREAD, LOG_INTERP, 2, "mangle ctr_el0 as %x -> %x\n",
```
derekbruening, line 3009: Use `0x` to indicate it's being printed as hex

core/arch/mangle_shared.c
```diff
@@ -1640,9 +1640,13 @@ d_r_mangle(dcontext_t *dcontext, instrlist_t *ilist, uint *flags INOUT, bool man
 
 #ifdef AARCH64
         if (instr_is_icache_op(instr) && instr_is_app(instr)) {
-            next_instr = mangle_icache_op(dcontext, ilist, instr, next_instr,
-                                          get_app_instr_xl8(instr) + AARCH64_INSTR_SIZE);
-            continue;
+            instr_t *res =
+                mangle_icache_op(dcontext, ilist, instr, next_instr,
+                                 get_app_instr_xl8(instr) + AARCH64_INSTR_SIZE);
+            if (res) {
```
derekbruening, line 1646: style: explicit booleans: `if (res != NULL)`

core/arch/mangle_shared.c
```diff
@@ -1640,9 +1640,13 @@ d_r_mangle(dcontext_t *dcontext, instrlist_t *ilist, uint *flags INOUT, bool man
...
+                next_instr = res;
+                continue;
+            }
```
derekbruening, line 1649: This is confusing as it doesn't match the usage of the other mangle routines and is in fact the opposite: better to match their style and return next_instr and then the code here can go back to the simpler way it was before.

core/dispatch.c
```diff
@@ -941,9 +941,15 @@ dispatch_enter_dynamorio(dcontext_t *dcontext)
...
-            flush_fragments_from_region(dcontext, begin, end - begin, true,
-                                        NULL /*flush_completion_callback*/,
-                                        NULL /*user_data*/);
+            if (end > begin) {
+                flush_fragments_from_region(dcontext, begin, end - begin, true,
+                                            NULL /*flush_completion_callback*/,
+                                            NULL /*user_data*/);
+            } else {
+                LOG(THREAD, LOG_DISPATCH, 2,
+                    "WARNING: skip selfmod region from %llx to %llx, size %d\n", begin,
```
derekbruening, line 950: Missing 0x again; use %p for pointers.

core/arch/aarchxx/mangle.c
```diff
@@ -74,12 +74,16 @@ typedef struct ALIGN_VAR(16) _icache_op_struct_t {
      * consecutive icache lines (which could wrap around the top of memory).
```
derekbruening, line 74: Could we add some kind of test for this mangling?  A simple app that reads the DIC bit and prints it out or sthg and is run with the different options, in our regression suite so we won't accidentally break this and not notice in the future.

derekbruening (commented):
Some style nits.  Since I'm right here I may edit these to match; but I have no way of testing this.  Please see the comment about adding a simple test: that would be good to have.

core/arch/aarchxx/mangle.c
derekbruening: Done.

core/arch/aarchxx/mangle.c
derekbruening: Switched to PIFX.

core/arch/mangle_shared.c
derekbruening: Deleted this code per comment below.

core/arch/mangle_shared.c
derekbruening: Delete this and revert to how it was.

core/dispatch.c
derekbruening: Done.

derekbruening (commented):
I directly edited the minor things.  Would it be possible for you to add a test?  In suite/tests/common/; as suggested, one app with 2 runs so 2 expect files.

Issue #5784, opened 2022-12-10
drmemtrace: Incorrect kernel xfer marker for signals received at end of wrapped function


abhinav92003:
A proprietary app hit the "Signal handler return point incorrect" invariant error. This happens for signals received at the end of a function that was wrapped with drwrap.

`view` output before the error at `23871501`:
```
 23871110 17417534: T751373 <marker: function #9> // wrapped function starts
 23871111 17417534: T751373 <marker: function return address 0xaaaae88c138c>
...
 23871133 17417549: T751373 ifetch       4 byte(s) @ 0x0000aaaae891ee04 d65f03c0   ret    %x30  // wrapped function ends
 23871134 17417549: T751373 <marker: kernel xfer from 0xaaaae6bf81a8 to handler> // signal arrives: wrong marker value. This should be 0xaaaae88c138c, but it's probably replace_retaddr_sentinel.
... <signal processing>
 23871496 17417807: T751373 <marker: syscall xfer from 0xaaaae6bffcf0> // signal returns
 23871499 17417807: T751373 <marker: function #9> // drwrap post-wrap callback
 23871500 17417807: T751373 <marker: function return value 0x3528260f6028>
 23871501 17417808: T751373 ifetch       4 byte(s) @ 0x0000aaaae88c138c // back to actual app return address
```

`-no_record_replace_retaddr` resolves this, but we should fix it for `-record_replace_retaddr` too.

abhinav92003:
The issue reported in the first comment is actually not resolved yet. #5788 fixes a different type of "Signal handler return point incorrect" error.

Issue #5740, opened 2022-11-17
attach_test and attach_blocking timing out on Jenkins


abhinav92003:
Happened multiple times on retrying: http://139.178.82.61:8080/job/DynamoRIO-AArch64-Precommit/2360/console.

Seems to have started after the recent AArch64 hardware move: https://groups.google.com/g/dynamorio-devs/c/lSaVAoryR1E/m/xTylpzXvDAAJ?utm_medium=email&utm_source=footer.

derekbruening:
Xref attach and detach test failures but with different symptoms on x86-64: #6452, #6536.

derekbruening:
Is this #6558?

derekbruening:
Confirmed: this is #6558.  It works if I enable ptrace_scope.  And that explains why the failures started on a new machine.

PR #5788, opened 2022-12-14
i#5784: Fix false positives in invariant checker

abhinav92003:
Fixes two false positives in the drcachesim invariant checker code. Skips the pre-signal instr check when returning from signals that arrived before any other instr in the trace, or before any other instr since the last kernel xfer (nested signals without any intervening instr). In both these cases we don't have any pre-signal instr whose pc we can compare with the post-signal instr.

Note that we still do not handle the back-to-back signals case where the second signal arrives just after the first one returns, without any intervening instruction. Improved the comment for that.

Adds tests that fail without this fix.

Fixes: #5784

### DIFF
--- a/clients/drcachesim/tests/invariant_checker_test.cpp
+++ b/clients/drcachesim/tests/invariant_checker_test.cpp
@@ -246,6 +246,38 @@ check_kernel_xfer()
         if (!run_checker(memrefs, false))
             return false;
     }
+    // Signal before any instr in the trace.
+    {
+        std::vector<memref_t> memrefs = {
+            // No instr in the beginning here.
+            gen_marker(1, TRACE_MARKER_TYPE_KERNEL_EVENT, 2),
+            gen_instr(1, 101),
+            // XXX: This marker value is actually not guaranteed, yet the checker
+            // requires it and the view tool prints it.
+            gen_marker(1, TRACE_MARKER_TYPE_KERNEL_XFER, 102),
+            gen_instr(1, 2),
+        };
+        if (!run_checker(memrefs, false))
+            return false;
+    }
+    // Nested signals without any intervening instr.
+    {
+        std::vector<memref_t> memrefs = {
+            gen_instr(1, 1),
+            gen_marker(1, TRACE_MARKER_TYPE_KERNEL_EVENT, 2),
+            // No intervening instr here.
+            gen_marker(1, TRACE_MARKER_TYPE_KERNEL_EVENT, 101),
+            gen_instr(1, 201),
+            // XXX: This marker value is actually not guaranteed, yet the checker
+            // requires it and the view tool prints it.
+            gen_marker(1, TRACE_MARKER_TYPE_KERNEL_XFER, 202),
+            gen_instr(1, 101),
+            gen_marker(1, TRACE_MARKER_TYPE_KERNEL_XFER, 102),
+            gen_instr(1, 2),
+        };
+        if (!run_checker(memrefs, false))
+            return false;
+    }
     // Fail to return to recorded interruption point.
     {
         std::vector<memref_t> memrefs = {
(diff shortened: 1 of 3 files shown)


clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -483,8 +487,19 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                         "Kernel event marker value missing");
         if (memref.marker.marker_type == TRACE_MARKER_TYPE_KERNEL_EVENT &&
             // Give up on back-to-back signals.
```
abhinav92003, line 489: @derekbruening I assume back-to-back means

```
kernel xfer
syscall xfer
kernel xfer
...
```

Why do we "give up" on this?

derekbruening: I assume b/c it's complex.  But isn't this a case this PR is trying to solve?  So the code is already trying to avoid raising an error, but fails?

abhinav92003: No this PR is solving nested signals without an intervening instr.

abhinav92003: The added tests show the exact case.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -246,6 +246,38 @@ check_kernel_xfer()
...
+            gen_instr(1, 101),
+            // XXX: This marker value is actually not guaranteed, yet the checker
+            // requires it and the view tool prints it.
+            gen_marker(1, TRACE_MARKER_TYPE_KERNEL_XFER, 102),
+            gen_instr(1, 2),
+        };
+        if (!run_checker(memrefs, false))
+            return false;
+    }
+    // Nested signals without any intervening instr.
```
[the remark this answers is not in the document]
derekbruening: Update: this was from before the offline discussion.

clients/drcachesim/tools/invariant_checker.cpp
derekbruening, line 489: Add a XXX or TODO comment about adding checks for back-to-back?

derekbruening (approved):

clients/drcachesim/tests/invariant_checker_test.cpp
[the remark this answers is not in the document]
abhinav92003: Clarified in the description that that case isn't handled.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -246,6 +246,38 @@ check_kernel_xfer()
...
+    {
+        std::vector<memref_t> memrefs = {
+            gen_instr(1, 1),
+            gen_marker(1, TRACE_MARKER_TYPE_KERNEL_EVENT, 2),
+            // No intervening instr here.
+            gen_marker(1, TRACE_MARKER_TYPE_KERNEL_EVENT, 101),
+            gen_instr(1, 201),
+            // XXX: This marker value is actually not guaranteed, yet the checker
+            // requires it and the view tool prints it.
```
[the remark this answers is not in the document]
abhinav92003: Probably not guaranteed in cases when the app returns from a handler using longjmp or other non-sigreturn means.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: Done.

clients/drcachesim/tests/invariant_checker_test.cpp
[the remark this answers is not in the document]
derekbruening: Oh, I thought you meant the integer value within the marker: you mean the presence of the marker itself.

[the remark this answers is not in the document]
abhinav92003: Right. Btw I copied this comment from elsewhere in the test; did not add it myself.

abhinav92003:
Various non-related issues with the AArch64 CI:

tool.drcacheoff.invariant_checker: inconsistent chunk counts: #5724

attach_test and attach_blocking timeout: #5740

fib-conflict and fib-conflict-early failures: #2417

abhinav92003:
run arm tests

Issue #4232, opened 2020-03-27
Speed up detach by reducing translation retries


derekbruening:
Separating out from #4219 which is about correctness problems.

With the stats from PR #4228 on a large ~600-thread app under drcachesim, I observed ~500 retries (40x more than plain DR) on detach, and a detach time increase of >2x vs plain DR.  This issue trying to reduce that by adding translation handling for more suspend locations so we don't have to retry to suspend.

Xref #37, #38 on ptrace attach/detach which might also speed things up vs signals.

derekbruening:
Xref #4669 showing frequent translation failures on block prefixes (that was -disable_traces) and rseq mangling.  Exit stubs are also missing translation (#4675).

PR #5791, opened 2022-12-15
i#5786: Add precise clean call mangling identification

### LINKED ISSUE #5786: CRASH: tool pc-rel instru mangled by DR can be mistaken as clean call instru and mis-translated
We are seeing crashes when using drbbdup on x86 on detach.  The drbbdup state restore function seems perfectly correct: it is DR that is messing things up.  We finally tracked it down to DR thinking that drbbdup's case load instructions are in fact a clean call setup.  DR then tries to restore the context from the dstack and gets incorrect values.

This is with drmemtrace where the drbbdup case operand is a pc-relative load.  DR has a feature where it will mangle tool-inserted pc-relative references into two-part sequences if they won't reach in their code cache destination.  Thus we end up with something like this (this is with no tool instrumentation aside from drbbdup, so we just have 2 copies of this simple cmp;jnz app block) where diagnostics are marking 3 instrs considered "our_mangling":

```
 +0    m4 @0x00007f6b07b75f70  65 48 89 3c 25 20 01 mov    %rdi -> %gs:0x00000120[8byte]
                               00 00
 +9    m4MANGLING @0x00007f6b0b183178  48 bf 88 34 35 6d 17 mov    $0x000056176d353488 -> %rdi
                               56 00 00
 +19   m4MANGLING @0x00007f6b0b1832c8  48 8b 3f             mov    (%rdi)[8byte] -> %rdi
 +22   m4 @0x00007f6b07b75d90                       <label note=0x000000000000004e>
 +22   m4 @0x00007f6b0b183260  e9 49 00 00 00       jmp    @0x00007f6b0b1830a8[8byte]
 +27   m4 @0x00007f6b0b183040  65 48 8b 3c 25 20 01 mov    %gs:0x00000120[8byte] -> %rdi
                               00 00
 +36   L3 @0x00007f6b07b75ec8  80 7d bc 00          cmp    0xffffffbc(%rbp)[1byte] $0x00
 +40   m4 @0x00007f6b0b183348                       <label note=0x0000000000000001>
 +40   m4 @0x00007f6b07b75c40  e9 4a 00 00 00       jmp    @0x00007f6b07b75cc0[8byte]
 +45   m4 @0x00007f6b07b75d28                       <label note=0x0000000000000002>
 +45   m4 @0x00007f6b0b1830a8                       <label note=0x000000000000004e>
 +45   m4 @0x00007f6b0b1831e0  65 48 8b 3c 25 20 01 mov    %gs:0x00000120[8byte] -> %rdi
                               00 00
 +54   L3 @0x00007f6b0b1833c8  80 7d bc 00          cmp    0xffffffbc(%rbp)[1byte] $0x00
 +58   m4 @0x00007f6b07b75e60                       <label note=0x0000000000000001>
 +58   m4 @0x00007f6b07b75cc0                       <label note=0x000000000000004f>
 +58   m4 @0x00007f6b07b75df8                       <label note=0x0000000000000002>
 +58   L4 @0x00007f6b07b75b98  0f 85 62 53 e2 09    jnz    $0x000056176cab2268
 +64   L4MANGLING @0x00007f6b0b183430  e9 f2 52 e2 09       jmp    $0x000056176cab21f7
```

The first 2 of those are the 2-part case load.

The current translation code has no precise way to identify clean calls and instead it identifies everything else and says that anything in "our_mangling" with no translation must then be a clean call.  But DR mangling tool code breaks this assumption.

We do not want to not mark DR-mangled tool code as not our_mangling as there can be spills and restores in there and we want DR to properly restore those.

We need to either precisely identify mangle_rel_addr() sequences, or precisely identify clean call sequences.  Both are complex.  Since the pc-rel doesn't need any action beyond regular spill tracking, it seems better to identify clean calls.  There are plenty of bits in translation_info_t, but not in instr_t: so I'm thinking of using labels for the instrlist case, and flags for the stored-info case.

derekbruening:
Adds new labels delimiting clean call sequences.
Converts into a translation record flag when storing translations.

Uses the new labels and flag to precisely identify clean call mangling, 
replacing the previous scheme which incorrectly thought mangled tool 
pc-relative was a clean call, resulting in incorrect translations and crashes.

To enable using labels after emitting, stops using instr_t.note to hold encoding
offsets for pc-releative operands.  Adds a new field instr_t.offset which is used for
this purpose.  This leaves note values in place across encodings, which is needed
for new clean call marking labels and also simplifies rseq handling code.

This is a compatibility break and as such we increase the version and
OLDEST_COMPATIBLE_VERSION to 990.

Since we only reserved 16 labels for DR, and we're already breaking compatilibity,
we go ahead and reduce DR_NOTE_FIRST_RESERVED to give DR more reserved labels.

Adds a test case to api.detach_state by adding a client (by converting it to use static DR) which inserts a pc-relative load.  This reproduces the crash on detach, and is fixed with this fix. The added instrumentation caused periodic detach failures which were solved by setting the translation and adding a restore-state event: i#4232 covers trying to improve the situation.

Changing the test to use static DR resulted in some change in some Linux kernel lazy
AVX state restores on sigreturn causing the AVX values set on detach to be ignored.
We solve this by writing the real xstate_bv into the signal frame when setting the
xstate context.

Issue: #5786, #4232
Fixes #5786


derekbruening:
WIll add a ref to #4219 as well which covers xl8 problems within the clean call sequence, not solved here.

core/lib/globals_api.h
```diff
@@ -854,6 +854,13 @@ enum {
      * their application values, as required for DR's internal block mangling.
      */
     DR_NOTE_REG_BARRIER,
+    /* Used for translation from an instruction list.  These apply not only to
+     * client-inserted clean calls but all inserted calls whether inserted by
+     * clients or DR and whether fully clean or not.  This is thus distinct from
+     * DR_NOTE_CLEAN_CALL_END.
+     */
+    DR_NOTE_CALLOUT_SEQUENCE_START,
+    DR_NOTE_CALLOUT_SEQUENCE_END,
```
derekbruening, line 863: It looks like DR_NOTE_FIRST_RESERVED only gives DR 16 reserved note values?  I'm hoping the instr_t.offset change will open up the use of labels for translation support for all the clean all varieties and subsections: but if those all use top-level labels we'll run out.  Should have a single call label and put sub-labels inside the label data area?  Should we change the reservation value to give us a few hundred at least if not thousands -- I really doubt that's much of a compat break, and if we're already bumping the version we could do that now.

derekbruening:
detach_state is failing on the GA CI machine with an interesting ymmh state not being preserved issue which does not reproduce locally.  Going to go ahead and ask for an initial review while I try repro/debug  on other platforms or tmate.  Probably the instr_t.offset would be separated into its own PR: but first let's agree that using labels for translation is the way to go.

CMakeLists.txt
```diff
@@ -567,7 +567,7 @@ endif (EXISTS "${PROJECT_SOURCE_DIR}/.svn")
 
 # N.B.: when updating this, update the default version in ci-package.yml.
 # We should find a way to share (xref i#1565).
-set(VERSION_NUMBER_DEFAULT "9.0.${VERSION_NUMBER_PATCHLEVEL}")
+set(VERSION_NUMBER_DEFAULT "9.90.${VERSION_NUMBER_PATCHLEVEL}")
```
abhinav92003, line 570: Only a minor level bump?

core/arch/mangle_shared.c
```diff
@@ -188,6 +188,10 @@ prepare_for_clean_call(dcontext_t *dcontext, clean_call_info_t *cci, instrlist_t
 {
     uint dstack_offs = 0;
 
+    instr_t *start_label = INSTR_CREATE_label(dcontext);
+    instr_set_note(start_label, (void *)DR_NOTE_CALLOUT_SEQUENCE_START);
```
abhinav92003, line 192: The note doesn't say which sequence starts/ends.

core/ir/instr.h
```diff
@@ -170,9 +170,8 @@ enum {
 #else
     /* Indicates an instruction that's part of the rseq endpoint.  We use this in
      * instrlist_t.flags (sort of the same namespace: INSTR_OUR_MANGLING is used there,
-     * but also EDI_VAL_*) and as a version of DR_NOTE_RSEQ that survives encoding
-     * (seems like we could store notes for labels in another field so they do
-     * in fact survive: a union with instr_t.translation?).
+     * but also EDI_VAL_*); we no longer use it on individual instructions since
+     * the label note field DR_NOTE_RSEQ now survives encoding.
      */
     INSTR_RSEQ_ENDPOINT = 0x01000000,
```
abhinav92003, line 176: Suffix _DEPRECATED.

core/heap.c
```diff
@@ -112,18 +112,17 @@ static const uint BLOCK_SIZES[] = {
     /* 40 dbg / 36 rel: */
     ALIGN_FORWARD(sizeof(fragment_t) + sizeof(indirect_linkstub_t), HEAP_ALIGNMENT),
 #if defined(X64)
-    sizeof(instr_t), /* 104 x64 */
 #    ifdef DEBUG
     sizeof(fragment_t) + sizeof(direct_linkstub_t) +
         sizeof(cbr_fallthrough_linkstub_t), /* 112 dbg x64 / 104 rel x64 */
 #    else
-/* release == instr_t */
+    sizeof(instr_t), /* 112 x64 */
```
abhinav92003, line 119: Do we now have one element less for DEBUG ?

Also the position of `sizeof(instr_t)` is different in BLOCK_SIZES now. Does that matter? I don't see any other adjustment to BLOCK_SIZES reads.

core/lib/globals_api.h
abhinav92003: The only concern I can think of is if some client was reserving a very large range. But I agree it should be fine to extend  DR_NOTE_FIRST_RESERVED to give core DR more. If there is some conflict then `drmgr_reserve_note_range` would inform the user.

suite/tests/api/detach_state.c
```diff
@@ -410,6 +410,111 @@ test_thread_func(void (*asm_func)(void))
...
+    static bool do_once;
+    if (!do_once) {
+        print("in dr_client_main\n");
+        do_once = true;
+    }
+    dr_register_bb_event(event_bb);
+    dr_register_restore_state_ex_event(event_restore);
+    dr_register_exit_event(event_exit);
+
+    /* Try to get an address that is not 32-bit-displacement reachable from
```
abhinav92003, line 507: ASSERT that it's indeed not reachable to avoid the test silently losing utility.

suite/tests/api/detach_state.c
```diff
@@ -410,6 +410,111 @@ test_thread_func(void (*asm_func)(void))
...
+                                               OPND_CREATE_ABSMEM(load_from, OPSZ_PTR)));
+    dr_restore_reg(drcontext, bb, first, DR_REG_XAX, SPILL_SLOT_1);
+    /* XXX i#4232: If we do not give translations, detach can fail as there are so
+     * many no-translation instrs here a thread can end up never landing on a
+     * full-state-translatable PC.  Thus, we set the translation, and provide a
+     * restore-state event.  i#4232 covers are more automated/convenient way to solve
+     * this.
+     */
+    for (instr_t *inst = instrlist_first(bb); inst != first; inst = instr_get_next(inst))
+        instr_set_translation(inst, instr_get_app_pc(first));
```
abhinav92003, line 441: Since this is a test it's not a concern, just trying to understand what this implies: this means each of these instrs are considered a safe spot? And the instrumentation could be interrupted, and if/when we resume, the instrumentation would be repeated too, right?

CMakeLists.txt
```diff
@@ -567,7 +567,7 @@ endif (EXISTS "${PROJECT_SOURCE_DIR}/.svn")
```
abhinav92003, line 567: Reminder to add to the PR description the part about the new offset field.

core/arch/mangle_shared.c
abhinav92003: Maybe say "clean call" explicitly?

abhinav92003 (approved):

CMakeLists.txt
derekbruening: https://dynamorio.org/page_new_release.html#autotoc_md227
"if we break backward compatibility during development, we expect to increment the major number on the next release and we immediately change the minor number to 90. "

core/arch/mangle_shared.c
derekbruening: What do you mean "which sequence starts/ends"?  The note indicates the start of inserted instructions that prepare to do a callout.

core/ir/instr.h
derekbruening: It's still used for instrlist flags which generally use the same flag space.

CMakeLists.txt
derekbruening: Do you think the instr_t.offset portion should be split into its own PR?  I was leaning that way -- or do you think it's fine all as one (less work; but less isolation + harder rollback if it breaks sthg later)?

core/arch/mangle_shared.c
abhinav92003: Oh  I assumed "CALLOUT" to mean that the label is "calling out" something, which led me to ask "what is it calling out".

CMakeLists.txt
abhinav92003: I'd be fine keeping it as one. This PR is not too big. In a different environment I'd suggest splitting (where we're doing regular prod deployments and want to wait between major changes), no need to split to just submit as two parts (unless we want to track both works separately). If we're rolling back, we'd have to roll back both parts anyway.

CMakeLists.txt
abhinav92003: Okay.

derekbruening:
An update here: still trying to figure out the detach_state ymm high bit failure on GA CI.  It does not reproduce in my work workstation; it does reproduce on an Ubuntu22 machine.

I did track down the key details: the issue is that xstate_bv is set to 0x203 on the suspend signals when DR is statically linked into api.detach_state.  That means the AVX state (the high ymm bits) are ignored by sigreturn and end up 0.  This has nothing to do with any other part of this PR: solely with DR being statically linked in.

It is weird: xstate_bv is 0x203 in the 1st SIGSEGV in the main thread for both dynamic and static DR; but for dynamic all subsequent signals have 0x207.  The actual hardware is always 0x207 the entire time.  So it seems like a lazy signal state feature in the kernel but only on some kernels and somehow static DR keeps it lazy: not sure how b/c dynamic doesn't really do anything different wrt AVX with the start/stop interface.

derekbruening:
Adding an AVX instr to the app main thread and sideline thread doesn't change
anything (these were already there in the sideline thread anyway to store the sentinels):
```
    __asm("vmovdqu %%ymm0, %%ymm0" : : : "ymm0");//NOCHECK
```

I still don't understand how it can be so locally lazy: but I solved by adding
what save_xmm() does, setting the signal's xstate_bv to the real value inside
mcontext_to_sigcontext_simd():
```
mcontext_to_sigcontext_simd: setting xstate_bv from 0x0000000000000203 to 0x0000000000000207
```

core/lib/globals_api.h
derekbruening: Done: reduced the value to give DR thousands.  Added a release note about this.

core/heap.c
derekbruening: No.  There are asserts that alert on any invariant violation in the size sequence.

CMakeLists.txt
derekbruening: Ack.

derekbruening: Updated the PR description above.

suite/tests/api/detach_state.c
derekbruening: This is not simple to do -- the best I found was when we have a cache address in event_restore().  Done.

suite/tests/api/detach_state.c
derekbruening: Yes, we'd restart this whole block.

derekbruening:
x64 failure is api.thread_churn assert: have not seen that before.  Does not repro locally.  Likely a flake; re-running; will file after re-run.

derekbruening:
It's thread_churn again.
I found a machine where it reproduces -- the 2nd run does have 10 more unreachable heap blocks.
But, this test passed on the instr_t field and heap bucket size change: it is failing after the final changes none of which seem like they would affect heap usage.  Adding a 3rd run shows zero growth so I do not think this is a real problem:
```
peak unreach heap: 248 A => 258 B => 258 C
```

derekbruening:
If I run -loglevel 1 to get more info it suddenly drops:
```
peak unreach heap: 248 A => 234 B => 234 C
```
Hack-printing the heap category breakdowns shows it getting smaller
```
Total max (not nec. all used simult.):  11018 KB
Total max (not nec. all used simult.):  10980 KB
Total max (not nec. all used simult.):  10980 KB
peak unreach heap: 248 A => 258 B => 258 C
```

core/unix/signal_linux_x86.c
```diff
@@ -595,6 +597,18 @@ mcontext_to_sigcontext_simd(sig_full_cxt_t *sc_full, priv_mcontext_t *mc)
                            YMMH_REG_SIZE);
                 }
             }
+            /* Lazy signal state can stop us from setting the full state.
+             * Switch the signal to full state if necessary.
+             */
+            uint bv_high, bv_low;
```
abhinav92003, line 603: Is this meant to be a temporary solution until we figure out why subsequent signals in static DR don't have the "correct" xstate_bv value?

Some more context in this comment would help future readers.

abhinav92003 (approved):

derekbruening:
I looked closely at the heap usage and it seems that one run happens to push into a new unit and with the double-sizing it ends up with 10 more blocks due to that one larger-capacity unit.  Setting `-initial_global_heap_unit_size 256K` makes that go away: but then we hit this:
```
peak reach cache: 128 A => 144 B
```
This is getting rather frustrating, all the weird failures popping up.

derekbruening:
Searching backward on my local machine: it is the instr_t field and heap bucket commit which triggers the thread_churn failure; yet on GA CI that commit passed and it showed up later, which is unexplained.

derekbruening:
The cache jump from 128 to 144 happens on the 5 to 6 thread transition: still 144 Actually it's on 6 threads!  So if I start with 6 it passes.  Grrr.

api/docs/release.dox
```diff
@@ -136,7 +136,11 @@ changes:
    the register in the structure from struct.value.reg to
    struct.value.reg_and_element_size.reg. The element size can be accessed directly
    via struct.value.reg_and_element_size.element_size.
- - Changed the size of the #instr_t structure by appending a field.
+ - Changed the size of the #instr_t structure by appending a field which is used
+   for relative offsets while encoding.  The note field is no longer modified
+   during encoding.
+ - Reduced the value of #DR_NOTE_FIRST_RESERVED.  This is not expected to cause
+   problems unless clients are directly choosing high note values.
```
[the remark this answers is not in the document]
derekbruening: Done.

core/lib/globals_api.h
```diff
@@ -831,19 +831,30 @@ typedef struct _module_data_t module_data_t;
...
+     */
+    DR_NOTE_CALL_SEQUENCE_START = DR_NOTE_FIRST_RESERVED,
+    DR_NOTE_CALL_SEQUENCE_END,
+    /* **** Add new labels here in the new space in between new and old. ****** */
     /**
      * Identifies an annotation point.  This label will be replaced by a
      * clean call to the registered annotation handler.
      */
-    DR_NOTE_ANNOTATION = DR_NOTE_FIRST_RESERVED + 1,
+    DR_NOTE_ANNOTATION = DR_NOTE_OLD_FIRST_RESERVED + 1,
```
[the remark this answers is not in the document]
derekbruening: I guess so.  Done.

core/unix/signal_linux_x86.c
derekbruening: No, the assumption is there is some real kernel local laziness and we need this fix permanently.  Added more comments.

derekbruening:
thread_churn passes on the one local machine where the first problem reproduced; but it still fails on GA CI.

Plus the Windows runs here seem to be hanging.  This PR has been a nightmare -- and so far all the issues do not seem to be its fault but are other limitations or latent issues or inherent non-determinism that happen to be triggered by these changes.

derekbruening:
The Windows runs get cancelled after 6 hours (I think that's the GA time limit) -- yet the Run Suite step has no logs that can be viewed or downloaded so it is not at all clear what is happening.

derekbruening:
Looking at other PR's, their Windows runs complete as you'd expect after 15-20 mins (though they seem to have some new test failures that appeared a few days ago). Maybe something in this PR is causing hangs and some Github problem is preventing the logs from showing which tests are hanging.

Issue #5734, opened 2022-11-13
tool.drcacheoff.rseq failing on A64 VM with glibc 2.35


abhinav92003:
The `tool.drcacheoff.rseq` test fails due to an invariant error, on an A64 VM with glibc 2.35 (disabled glibc's rseq due to i#5431). The same also happens with the #5711 fix, with and without glibc's rseq. 

```
238:   Trace invariant failure in T-1 at ref # 0: Serial schedule entry count does
238:   not match trace
```

The test works fine on the A64 Jenkins machine (which has an older glibc).

abhinav92003:
Same happens with `-disable_rseq` also.

derekbruening:
For the Jenkins disabling the failures are different and not coming from glibc 2.35:

https://github.com/DynamoRIO/dynamorio/pull/5885#issuecomment-1446371114

> No, it's still on 2.31 and the problem only occurs during postcommit testing, which is very odd.

derekbruening:
> Saw this on x86 too with glibc 2.36.

tool.drcacheoff.rseq passes with glibc 2.36 on x86-64 for me (if I work around the up-front assert failure #5955 with the env var): are you seeing this only once every N runs?

Issue #5733, opened 2022-11-13
api.rseq flakiness on glibc 2.35 A64 due to too long run times


abhinav92003:
The `api.rseq` test has widely varying run times on A64. On the Jenkins machine (which currently has libc 2.31)
```
    Start 307: code_api|api.rseq
    Test #307: code_api|api.rseq ................   Passed    3.06 sec
    Start 307: code_api|api.rseq
    Test #307: code_api|api.rseq ................   Passed    2.56 sec
    Start 307: code_api|api.rseq
    Test #307: code_api|api.rseq ................   Passed    6.20 sec
    Start 307: code_api|api.rseq
    Test #307: code_api|api.rseq ................   Passed   23.19 sec
    Start 307: code_api|api.rseq
    Test #307: code_api|api.rseq ................   Passed   20.61 sec
    Start 307: code_api|api.rseq
    Test #307: code_api|api.rseq ................   Passed   19.14 sec
    Start 307: code_api|api.rseq
    Test #307: code_api|api.rseq ................   Passed    0.30 sec
    Start 307: code_api|api.rseq
    Test #307: code_api|api.rseq ................   Passed    8.32 sec
    Start 307: code_api|api.rseq
    Test #307: code_api|api.rseq ................   Passed    4.98 sec
    Start 307: code_api|api.rseq
1/1 Test #307: code_api|api.rseq ................   Passed    9.42 sec
```

This test traces rseq execution using drmemtrace. The outputs are also really big (and of widely varying sizes):

```
3.1G	suite/tests/drmemtrace.api.rseq.3835685.2325.dir
2.6G	suite/tests/drmemtrace.api.rseq.3835688.0577.dir
5.7G	suite/tests/drmemtrace.api.rseq.3835691.6161.dir
22G	suite/tests/drmemtrace.api.rseq.3835694.4192.dir
20G	suite/tests/drmemtrace.api.rseq.3835697.1669.dir
20G	suite/tests/drmemtrace.api.rseq.3835700.9167.dir
282M	suite/tests/drmemtrace.api.rseq.3835703.5354.dir
8.0G	suite/tests/drmemtrace.api.rseq.3835706.8330.dir
5.1G	suite/tests/drmemtrace.api.rseq.3835709.4015.dir
9.7G	suite/tests/drmemtrace.api.rseq.3835712.2611.dir
```
[...]

PR #5796, opened 2022-12-16
i#5733, i#5734: Minor fixes in tests

abhinav92003:
Adds -max_trace_size to api.rseq to avoid generating too much trace data,
which may cause us to run out of disk space. The issue on varying run times
(and therefore trace data) is not reproducing now on the Jenkins machine,
but with this change, the trace size comes down to 512K from 130M+ which
is better.

Fixes an issue in the invariant checker that causes a SIGSEGV crash during
error reporting when there's no shard stream available. This came up in the
tool.drcacheoff.rseq test during the `check_schedule_data` checks that
happen at the very end in `print_results` of the invariant checker, where
we do not have the shard stream anymore. The invariant error itself is yet
to be debugged (i#5734).

Issue: #5733, #5734

### DIFF
--- a/clients/drcachesim/tools/invariant_checker.cpp
+++ b/clients/drcachesim/tools/invariant_checker.cpp
@@ -549,6 +549,7 @@ invariant_checker_t::check_schedule_data()
     // Check that the scheduling data in the files written by raw2trace match
     // the data in the trace.
     per_shard_t global;
+    // Use a synthetic stream object to allow report_if_false to work normally.
     auto stream = std::unique_ptr<memtrace_stream_t>(new default_memtrace_stream_t());
     global.stream = stream.get();
     std::vector<schedule_entry_t> serial;
--- a/clients/drcachesim/tools/invariant_checker.cpp
+++ b/clients/drcachesim/tools/invariant_checker.cpp
@@ -550,7 +550,8 @@ invariant_checker_t::check_schedule_data()
     // the data in the trace.
     per_shard_t global;
     // Use a synthetic stream object to allow report_if_false to work normally.
-    auto stream = std::unique_ptr<memtrace_stream_t>(new default_memtrace_stream_t());
+    auto stream = std::unique_ptr<memtrace_stream_t>(
+        new default_memtrace_stream_t(&global.ref_count_));
     global.stream = stream.get();
     std::vector<schedule_entry_t> serial;
     std::unordered_map<uint64_t, std::vector<schedule_entry_t>> cpu2sched;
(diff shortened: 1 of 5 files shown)


clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -71,10 +71,11 @@ void
 invariant_checker_t::report_if_false(per_shard_t *shard, bool condition,
                                      const std::string &invariant_name)
 {
+    uint64_t ref_count = shard->stream == nullptr ? shard->ref_count_
```
derekbruening, line 74: This seems like it will mask errors: shard->ref_count_ is not guaranteed to be maintained (e.g., serial) and there would be no indication of a problem except a bogus count.  I think we should keep an invariant that the stream is present on entry here, and set it inside check_schedule_data which is already using a synthetic shard struct.

derekbruening (approved):
Meant to mark approved since the synthetic stream should be straightforward and probably doesn't need another look.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: I don't understand. This is not test code where we can create a test stream object. For a check done in `check_schedule_data` a bogus ref_count is fine since there is no ref count for a non-trace (cpu schedule) file.

abhinav92003: Is your concern about other cases where we do want a shard stream but don't have it. Shouldn't that be asserted elsewhere?

abhinav92003: Or do you mean change the report_if_false interface to accept a ref_count and tid.

derekbruening: > Is your concern about other cases where we do want a shard stream but don't have it. Shouldn't that be asserted elsewhere?

Yes, that is a concern.  It does not seem good to allow stream to be empty but at the same time blindly assuming that ref_count_ is valid when we have cases where it's not (even if today we don't have cases where both are invalid...it just feels fragile).  I'm voting for:

- All `per_shard_t` types have a valid stream object.  If `check_schedule_data` wants to create a fake one, it has to fill in the stream, instead of forcing other functions that take `per_shard_t` to handle its incomplete shard.

abhinav92003: Just to be clear: your suggestion is to pass a test_memtrace_stream_t object from check_schedule_data (no overrides, just let the default get_record_ordinal return zero). This is doable but maybe we should rename test_memtrace_stream_t to default_memtrace_stream_t.

abhinav92003: Done.

derekbruening: Looking at chedk_schedule_data it does update global.ref_count_: maybe that *should* that be part of the error report.

abhinav92003: Added.

derekbruening (approved):

Issue #5802, opened 2022-12-21
file_io tests are failing when run in the CI on windows machines


joshua-warburton:
Across several PRs, the file_ip tests are failing with no seeming relation to the content of the patch, such as here:
https://github.com/DynamoRIO/dynamorio/actions/runs/3746175708/jobs/6361283731

```
tart 116: code_api|client.file_io
2022-12-19T14:48:07.3715131Z 
2022-12-19T14:48:07.3720035Z 116: Test command: D:\a\dynamorio\dynamorio\build_debug-internal-64\bin64\drrun.exe "-s" "90" "-quiet" "-debug" "-use_dll" "D:/a/dynamorio/dynamorio/build_debug-internal-64/lib64/debug/dynamorio.dll" "-exit0" "-stderr_mask" "0xC" "-msgbox_mask" "0" "-dumpcore_mask" "0x7d" "-staged" "-code_api" "-c" "D:/a/dynamorio/dynamorio/build_debug-internal-64/suite/tests/bin/client.file_io.dll.dll" "D:/a/dynamorio/dynamorio/suite/tests/client-interface/file_io_data.txt" "--" "D:/a/dynamorio/dynamorio/build_debug-internal-64/suite/tests/bin/client.file_io.exe"
2022-12-19T14:48:07.3721496Z 116: Working Directory: D:/a/dynamorio/dynamorio/build_debug-internal-64/suite/tests
2022-12-19T14:48:07.3721972Z 116: Test timeout computed to be: 600
2022-12-19T14:48:07.3722327Z 115: About to create thread
2022-12-19T14:48:07.3722677Z 115: kernel_xfer_event: type 2
2022-12-19T14:48:07.3722993Z 116: 0123456789
2022-12-19T14:48:07.3723272Z 116: 01234
2022-12-19T14:48:07.3723546Z 116: abcdefg
2022-12-19T14:48:07.3723805Z 116: x
2022-12-19T14:48:07.3724075Z 116: EOF
2022-12-19T14:48:07.3724344Z 116: 
2022-12-19T14:48:07.3724644Z 115: kernel_xfer_event: type 7
2022-12-19T14:48:07.3724982Z 115: kernel_xfer_event: type 5
2022-12-19T14:48:07.3725315Z 115: kernel_xfer_event: type 6
2022-12-19T14:48:07.3725647Z 115: kernel_xfer_event: type 5
[...]

PR #5803, opened 2022-12-23
i#5802: Disable failing file_io test on windows

joshua-warburton:
Temporarily disable file_io as it has started failing constantly after upgrades
on the windows CI runners. The reason is currently unknown but is tracked 
by #5802.

issue: #5802


suite/runsuite_wrapper.pl
```diff
@@ -243,6 +243,7 @@
                 'code_api,thread_private|common.decode-stress' => 1, # i#1807
```
abhinav92003, line 243: Reminder to add details to the PR description. And s/issues:/Issue:/

https://dynamorio.org/page_code_reviews.html

abhinav92003 (approved):
Thanks for sending the PR. Maybe it is best to just disable the test for now to unblock other PRs.

PR #5805, opened 2022-12-28
Updates to gitignore

prasun3:
Added ctags and cscope files to gitignore


.gitignore
```diff
@@ -8,3 +8,5 @@
 .vscode
 build
 CMakeLists.txt.user
+tags
```
abhinav92003, line 11: Add a comment about the tool that generates these (ctags and cscope?).

abhinav92003 (approved):

.gitignore
prasun3: Done.

PR #5819, opened 2023-01-12
i#4842 Add unit test for snoop filter

bete0:
Adds unit test for the MSI snoop coherence protocol.

Issue: #4842

### DIFF
--- a/clients/drcachesim/simulator/snoop_filter.cpp
+++ b/clients/drcachesim/simulator/snoop_filter.cpp
@@ -1,5 +1,5 @@
 /* **********************************************************
- * Copyright (c) 2015-2020 Google, Inc.  All rights reserved.
+ * Copyright (c) 2015-2023 Google, Inc.  All rights reserved.
  * **********************************************************/
 
 /*
@@ -59,6 +59,7 @@ void
 snoop_filter_t::snoop(addr_t tag, int id, bool is_write)
 {
     coherence_table_entry_t *coherence_entry = &coherence_table_[tag];
+    snoop_filter_stats_.coherence_table = coherence_table_;
     // Initialize new snoop filter entry.
     if (coherence_entry->sharers.empty()) {
         coherence_entry->sharers.resize(num_snooped_caches_, false);
@@ -134,4 +135,13 @@ snoop_filter_t::print_stats(void)
     std::cerr << prefix << std::setw(18) << std::left << "Writebacks:" << std::setw(20)
               << std::right << num_writebacks_ << std::endl;
     std::cerr.imbue(std::locale("C")); // Reset to avoid affecting later prints.
+}
+
+snoop_filter_stats_t
+snoop_filter_t::get_coherence_stats(void)
+{
+    snoop_filter_stats_.num_writes = num_writes_;
+    snoop_filter_stats_.num_writebacks = num_writebacks_;
+    snoop_filter_stats_.num_invalidates = num_invalidates_;
+    return snoop_filter_stats_;
 }
--- a/clients/drcachesim/tests/snoop_filter_test.cpp
+++ b/clients/drcachesim/tests/snoop_filter_test.cpp
@@ -1,0 +1,205 @@
+/* **********************************************************
+ * Copyright (c) 2016-2023 Google, Inc.  All rights reserved.
+ * **********************************************************/
+
+/*
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice,
+ *   this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of Google, Inc. nor the names of its contributors may be
+ *   used to endorse or promote products derived from this software without
+ *   specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL GOOGLE, INC. OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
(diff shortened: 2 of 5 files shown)


clients/drcachesim/tests/snoop_filter_test.cpp
```diff
@@ -0,0 +1,205 @@
...
+    {
+        cache_t *llc = new cache_lru_t;
+        if (!llc->init(/*associatibity=*/8, /*line_size=*/64,
+                       /*total_size=*/1024, nullptr,
+                       new cache_stats_t((int)64, "", true))) {
+            std::string error_string =
+                "Usage error: failed to initialize LL cache.  Ensure sizes and "
+                "associativity are powers of 2, that the total size is a multiple "
+                "of the line size, and that any miss file path is writable.";
+            std::cerr << error_string << "\n";
```
derekbruening, line 82: No reason to create a named string not used anywhere else?

clients/drcachesim/tests/snoop_filter_test.cpp
```diff
@@ -0,0 +1,205 @@
...
+    initialize_caches_and_snoop_filter()
+    {
+        cache_t *llc = new cache_lru_t;
+        if (!llc->init(/*associatibity=*/8, /*line_size=*/64,
+                       /*total_size=*/1024, nullptr,
+                       new cache_stats_t((int)64, "", true))) {
+            std::string error_string =
+                "Usage error: failed to initialize LL cache.  Ensure sizes and "
+                "associativity are powers of 2, that the total size is a multiple "
+                "of the line size, and that any miss file path is writable.";
```
derekbruening, line 81: But your own code here is setting all of these parameters: so this error message doesn't really make sense as the caller/user cannot ensure any of these mentioned things.

clients/drcachesim/tests/snoop_filter_test.cpp
```diff
@@ -0,0 +1,205 @@
...
+            std::cerr << error_string << "\n";
+            return;
```
derekbruening, line 83: This return path leaks `llc` and its stats class.

clients/drcachesim/tests/snoop_filter_test.cpp
```diff
@@ -0,0 +1,205 @@
...
+        }
+    }
+
+    void
+    initialize_caches_and_snoop_filter()
+    {
+        cache_t *llc = new cache_lru_t;
+        if (!llc->init(/*associatibity=*/8, /*line_size=*/64,
+                       /*total_size=*/1024, nullptr,
+                       new cache_stats_t((int)64, "", true))) {
```
derekbruening, line 77: Integer literals are type `int` by default so this cast can be removed.

clients/drcachesim/tests/snoop_filter_test.cpp
```diff
@@ -0,0 +1,205 @@
...
+                new cache_stats_t((int)SNOOPED_FILTER_LINE_SIZE, "", true, true),
+                /*prefetcher=*/nullptr,
+                /*inclusive=*/true, /*coherence_cache=*/true, i, snoop_filter_);
+        }
+
+        if (!snoop_filter_->init(snooped_caches_, num_cores_)) {
+            std::cerr << "Usage error: failed to initialize snoop filter.\n";
+            return;
+        }
+    }
```
derekbruening, line 102: Looks like `llc` and its stats class are *always* leaked.  (The parent pointer stored in each snooped_caches_ entry is not freed.)

clients/drcachesim/tests/snoop_filter_test.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+public:
+    snoop_filter_test_t(const int num_cores)
+        : num_cores_(num_cores)
+    {
+    }
+    ~snoop_filter_test_t()
+    {
+        if (snooped_caches_ != NULL) {
+            delete[] snooped_caches_;
```
derekbruening, line 64: This leaks every single pointer entry as it only frees the array.

How about using smart pointers for all these fields to avoid errors in freeing.

clients/drcachesim/tests/snoop_filter_test.cpp
```diff
@@ -0,0 +1,205 @@
...
+    snoop_filter_test.request_and_check_state(CPU_1, ADDR_A, TRACE_TYPE_READ,
+                                              /*num_writes*/ 0,
+                                              /*num_invalidates*/ 0,
+                                              /*num_writebacks*/ 0,
+                                              /*num_sharers*/ 1, false);
+    snoop_filter_test.request_and_check_state(CPU_0, ADDR_A, TRACE_TYPE_WRITE,
+                                              /*num_writes*/ 1,
+                                              /*num_invalidates*/ 1,
+                                              /*num_writebacks*/ 0,
+                                              /*num_sharers*/ 2, false);
```
derekbruening, line 154: Shouldn't the write invalidate the line in cpu 1 such that there are no sharers and the line is only in cpu 0 in a modified state?  I didn't write this coherence code so I'm less familiar with it: is the sharers vector used in some other way?

clients/drcachesim/tests/snoop_filter_test.cpp
```diff
@@ -0,0 +1,205 @@
...
+    snoop_filter_test.request_and_check_state(CPU_1, ADDR_A, TRACE_TYPE_READ,
+                                              /*num_writes*/ 0,
+                                              /*num_invalidates*/ 0,
+                                              /*num_writebacks*/ 0,
+                                              /*num_sharers*/ 1, false);
+    snoop_filter_test.request_and_check_state(CPU_2, ADDR_A, TRACE_TYPE_WRITE,
+                                              /*num_writes*/ 1,
+                                              /*num_invalidates*/ 2,
+                                              /*num_writebacks*/ 0,
+                                              /*num_sharers*/ 2, false);
```
derekbruening, line 182: I again am confused by the sharers after the write, and after further reads below.

clients/drcachesim/tests/snoop_filter_test.cpp
```diff
@@ -0,0 +1,205 @@
...
+    snoop_filter_test.request_and_check_state(CPU_0, ADDR_A, TRACE_TYPE_WRITE,
+                                              /*num_writes*/ 1,
+                                              /*num_invalidates*/ 1,
+                                              /*num_writebacks*/ 0,
+                                              /*num_sharers*/ 2, false);
+    snoop_filter_test.request_and_check_state(CPU_1, ADDR_A, TRACE_TYPE_READ,
+                                              /*num_writes*/ 1,
+                                              /*num_invalidates*/ 1,
+                                              /*num_writebacks*/ 1,
+                                              /*num_sharers*/ 1, true);
```
derekbruening, line 159: Doesn't cpu 0 still have the line, so there are 2 sharers?

derekbruening (changes requested):

clients/drcachesim/simulator/snoop_filter.cpp
```diff
@@ -135,3 +136,12 @@ snoop_filter_t::print_stats(void)
               << std::right << num_writebacks_ << std::endl;
     std::cerr.imbue(std::locale("C")); // Reset to avoid affecting later prints.
 }
+
+snoop_filter_stats_t
+snoop_filter_t::get_coherence_stats(void)
```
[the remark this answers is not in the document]
bete0: Ack. The style guide (https://dynamorio.org/page_code_style.html) still mentioned to explicitly use the `void` keyword. Will remove remove that suggestion from the guide.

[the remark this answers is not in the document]
derekbruening: It is required for C code so it should be removed but just qualified.

PR #5831, opened 2023-01-19
i#5686 bbdup xl8: Modify xl8 params for other libs

derekbruening:
Moves the drbbdup restore state event to a very early priority and has it modify the restore state parameters to focus on the containing copy of the block where the restore resides.  It modifies the instruction list and the cache start pc.

Adds a new test drx-scattergather-bbdup by adding drbbdup to the existing drx-scattergather test.  This test fails to restore without the fix described above, due to the scatter/gather expansion restore state machine getting confused across multiple copies.

Does not augment drbbdup's restore to handle the no-ilist case: leaves that for the general #3801 mechanism.

Issue: #5686

### DIFF
--- a/suite/tests/client-interface/drx-scattergather-bbdup.dll.c
+++ b/suite/tests/client-interface/drx-scattergather-bbdup.dll.c
@@ -163,9 +163,10 @@ event_bb_analyze_case(void *drcontext, void *tag, instrlist_t *bb, bool for_trac
     } else if (mode == INSTRU_MODE_EXPAND) {
         return event_bb_analysis(drcontext, tag, bb, for_trace, translating,
                                  analysis_data);
-    } else
+    } else {
         DR_ASSERT(false);
-    return DR_EMIT_DEFAULT;
+        return DR_EMIT_DEFAULT;
+    }
 }
 
 static void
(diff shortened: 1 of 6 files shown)


suite/tests/client-interface/drx-scattergather-bbdup.dll.c
```diff
@@ -0,0 +1,422 @@
...
+                      void *orig_analysis_data, void **analysis_data)
+{
+    if (mode == INSTRU_MODE_NOP) {
+        return DR_EMIT_DEFAULT;
+    } else if (mode == INSTRU_MODE_EXPAND) {
+        return event_bb_analysis(drcontext, tag, bb, for_trace, translating,
+                                 analysis_data);
+    } else
+        DR_ASSERT(false);
+    return DR_EMIT_DEFAULT;
```
bete0, line 168: Will it ever get to return DR_EMIT_DEFAULT? We are returning from the `if` and `else if ` clause and asserting in the `else` clause but we can only get to `return DR_EMIT_DEFAULT` if we hit the `else` but we are already asserting there.

bete0 (approved):

suite/tests/client-interface/drx-scattergather-bbdup.dll.c
derekbruening: If the assert is debug-only (common for most assert types) then yes it will reach the return in release build, and the compiler will warn if there's no return there.  DR_ASSERT is actually enabled in release but it still seems safer to have a return after it if we move to a debug-only check.  I will move the return inside the else{}.

PR #5840, opened 2023-01-28
i#5538 memtrace seek, part 11: Fix omitted header problems

derekbruening:
Fixes a bug where reader_t's detection of duplicated timestamp,cpuid headers at the start of a chunk assumed single-threaded mode.  We switch to using a simple per-tid chunk footer trigger.

Adds a test to view_test via a new serial mock which takes in trace_entry_t and allows testing of the interleaving code.  Tests both proper chunk header elision as well as replicating the bug where elision should not happen.

The test revealed a separate bug in the view tool where the version and filetype ordinals, for delaying, were not updated on new threads. That is fixed here as well as otherwise the new tests fail.

Issue: #5538

### DIFF
--- a/clients/drcachesim/tests/view_test.cpp
+++ b/clients/drcachesim/tests/view_test.cpp
@@ -1,5 +1,5 @@
 /* **********************************************************
- * Copyright (c) 2021-2022 Google, LLC  All rights reserved.
+ * Copyright (c) 2021-2023 Google, LLC  All rights reserved.
  * **********************************************************/
 
 /*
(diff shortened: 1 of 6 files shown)


clients/drcachesim/tests/view_test.cpp
```diff
@@ -58,6 +61,34 @@
         }                             \
     } while (0)
 
+// These are for our mock serial reader and must in the same namespace
```
bete0, line 64: typo: `must be`

bete0 (approved):

clients/drcachesim/tests/view_test.cpp
derekbruening: Done.

Issue #5848, opened 2023-02-02
ASSERT drcachesim invariant checker falsely triggers "Branch target not immediately after branch" when branching crosses chunks


lihasgupta:
**Describe the bug**

Consider the following snippet:
```
126673902: T676257 ifetch       7 byte(s) @ 0x0000000010e33e22 80 bd 4f ff ff ff 00 cmp    0xffffff4f(%rbp)[1byte] $0x00
126673903: T676257 read         1 byte(s) @ 0x00007f2c726d5dff by PC 0x0000000010e33e22
126673904: T676257 ifetch       6 byte(s) @ 0x0000000010e33e29 0f 88 24 85 3e 00    js     $0x000000001121c353
126673905: T676257 <marker: chunk footer #8>
126673906: T676257 <marker: timestamp 13311683087087727>
126673907: T676257 <marker: tid 676257 on core 16>
126673908: T676257 ifetch       7 byte(s) @ 0x0000000010e33e2f 48 8d 8d 38 ff ff ff lea    0xffffff38(%rbp) -> %rcx
126673909: T676257 ifetch       4 byte(s) @ 0x0000000010e33e36 48 8b 41 10          mov    0x10(%rcx)[8byte] -> %rax
```

The timestamp at 126673906 is a start-of-chunk timestamp (we get these timestamps at the start of every chunk) and is NOT used for interleaving. All analysis tools should ignore this timestamp. However, the invariant checker triggers the "Branch target not immediately after branch" invariant. This behavior is incorrect.

**To Reproduce**
This can easily be reproduced in a synthetic test case by having some branching, then the end of a chunk, and start of a new chunk. The test case will trigger this invariant even if the branch target does immediately follow the branch.

**Expected behavior**
There should be no invariant error.

PR #5849, opened 2023-02-02
i#5848: Loosen constraint on drcachesim "Branch target not immediately after branch" invariant when branching crosses chunks

lihasgupta:
Consider the following snippet:
```
126673902: T676257 ifetch       7 byte(s) @ 0x0000000010e33e22 80 bd 4f ff ff ff 00 cmp    0xffffff4f(%rbp)[1byte] $0x00
126673903: T676257 read         1 byte(s) @ 0x00007f2c726d5dff by PC 0x0000000010e33e22
126673904: T676257 ifetch       6 byte(s) @ 0x0000000010e33e29 0f 88 24 85 3e 00    js     $0x000000001121c353
126673905: T676257 <marker: chunk footer #8>
126673906: T676257 <marker: timestamp 13311683087087727>
126673907: T676257 <marker: tid 676257 on core 16>
126673908: T676257 ifetch       7 byte(s) @ 0x0000000010e33e2f 48 8d 8d 38 ff ff ff lea    0xffffff38(%rbp) -> %rcx
126673909: T676257 ifetch       4 byte(s) @ 0x0000000010e33e36 48 8b 41 10          mov    0x10(%rcx)[8byte] -> %rax
```

The timestamp at 126673906 is a start-of-chunk timestamp (we get these timestamps at the start of every chunk) and is NOT used for interleaving. All analysis tools should ignore this timestamp. However, the invariant checker triggers the "Branch target not immediately after branch" invariant. This behavior is incorrect.

This PR adds a test case that reproduces this issue, and then it adds some logic to make the timestamp marker at the start of a new chunk be invisible to the invariant_checker. This logic makes the added test case pass.

Issue: #5848

### DIFF
--- a/core/ir/aarch64/codec.c
+++ b/core/ir/aarch64/codec.c
@@ -4272,6 +4272,62 @@ encode_opnd_s16(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_out)
     return encode_opnd_vector_reg(16, 2, opnd, enc_out);
 }
 
+static inline bool
+svemem_gprs_per_element_decode(uint bytes_per_element, aarch64_reg_offset element_size,
+                               uint shift_amount, uint enc, int opcode, byte *pc,
+                               OUT opnd_t *opnd)
+{
+    ASSERT(element_size >= BYTE_REG && element_size <= DOUBLE_REG);
+    ASSERT(bytes_per_element <= (1 << element_size));
+
+    const uint elements = get_elements_in_sve_vector(element_size);
+    const opnd_size_t mem_transfer = opnd_size_from_bytes(bytes_per_element * elements);
+
+    *opnd = opnd_create_base_disp_shift_aarch64(
+        decode_reg(extract_uint(enc, 5, 5), true, true),
+        decode_reg(extract_uint(enc, 16, 5), true, false), DR_EXTEND_UXTX,
+        shift_amount != 0, 0, 0, mem_transfer, shift_amount);
+    return true;
+}
+
+static inline bool
+svemem_gprs_per_element_encode(uint bytes_per_element, aarch64_reg_offset element_size,
+                               uint shift_amount, uint enc, int opcode, byte *pc,
+                               opnd_t opnd, OUT uint *enc_out)
+{
+    ASSERT(element_size >= BYTE_REG && element_size <= DOUBLE_REG);
+    ASSERT(bytes_per_element <= (1 << element_size));
+
+    const uint elements = get_elements_in_sve_vector(element_size);
+    const opnd_size_t mem_transfer = opnd_size_from_bytes(bytes_per_element * elements);
+
+    if (!opnd_is_base_disp(opnd) || opnd_get_size(opnd) != mem_transfer ||
+        opnd_get_disp(opnd) != 0)
+        return false;
+
+    uint rn, rm;
+    bool is_x_register;
+    IF_RETURN_FALSE(!encode_reg(&rn, &is_x_register, opnd_get_base(opnd), true) ||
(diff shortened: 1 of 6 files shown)


clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -177,6 +177,22 @@ check_branch_target_after_branch()
         if (!run_checker(memrefs, false))
             return false;
     }
+    // Invariant should not trigger when it detects a timestamp at the start of a new
+    // chunk.
+    {
+        std::vector<memref_t> memrefs = {
+            gen_instr(1, 1),
+            gen_branch(1, 2),
+            gen_marker(1, TRACE_MARKER_TYPE_CHUNK_INSTR_COUNT, 1),
```
derekbruening, line 186: This is normally a header only present at the very top of a trace.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -177,6 +177,22 @@ check_branch_target_after_branch()
...
+            gen_marker(1, TRACE_MARKER_TYPE_CHUNK_FOOTER, 3),
+            gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 3),
```
derekbruening, line 188: Per the prior comment on reader_t hiding this probably we should only have the footer (which I assume still triggers the false pos).

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -364,9 +364,9 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
             report_if_false(shard,
                             !shard->saw_timestamp_but_no_instr_ ||
                                 // The invariant is relaxed for a signal.
-                                // prev_xfer_marker_ is cleared on an instr, so if set to
-                                // non-sentinel it means it is immediately prior, in
-                                // between prev_instr_ and memref.
+                                // prev_xfer_marker_ is cleared on an instr, so if set
+                                // to non-sentinel it means it is immediately prior,
```
derekbruening, line 368: Does your clang-format have different settings somehow and that is why it ended these lines earlier?

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -454,7 +454,10 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                   << " type " << memref.instr.type << "\n";
     }
     if (memref.marker.type == TRACE_TYPE_MARKER &&
-        memref.marker.marker_type == TRACE_MARKER_TYPE_TIMESTAMP) {
+        memref.marker.marker_type == TRACE_MARKER_TYPE_TIMESTAMP &&
+        // Timestamps at the start of a chunk should be invisible to analysis tools. These
+        // are duplicate timestamps used to faciliate skipping.
+        shard->prev_entry_.marker.marker_type != TRACE_MARKER_TYPE_CHUNK_FOOTER) {
```
derekbruening, line 460: This needs to check that the prev_entry_ is actually a marker before looking at a marker type, else you'll get mismatches.

derekbruening, line 460: See comment above: I believe we can remove this as reader_t hides these.  Plus, given that reader_t hides the dup timestamps, it is possible that the first real entry in the chunk is a timestamp (I don't think the inserted dups are omitted in that case?), and then if you keep this check here you can have a false negative.  I think what we want to do is to allow a chunk footer in between a branch and its target.

derekbruening (changes requested):

PR #5854, opened 2023-02-05
i#5845: Fix format string type for drcov bb count

### LINKED ISSUE #5845: drcov -dump_text table count is sometimes wrong
See https://groups.google.com/d/msgid/dynamorio-users/f9a1c229-4284-483f-8da7-79d120e6d330n%40googlegroups.com?utm_medium=email&utm_source=footer

There appears to be a format string error with a 32-bit output type for a 64-bit entry count for -dump_text: https://github.com/DynamoRIO/dynamorio/blob/master/ext/drcovlib/drcovlib.c#L142
This would not affect binary output.

derekbruening:
The drcov bb count was being printed as a 32-bit integer but passed a 64-bit value for x64/aarch64 builds.  We fix that here to pass a 32-bit value with an overflow check (we do not support >4G blocks in drcov's post-processing today).

Unfortunately there is not a simple way to make a test.  I experimented and actually failed to get the original bug to cause any visible problem even with subsequent specifiers in that same format string, with today's va_list accessors.  Making a test of millions or billions of blocks is non-trivial: the easiest way is to dynamically generate the code, but that is not a great test for drcov in general where we want source lines to map back to; plus it will take a while to run which is not great in the CI suite.

Fixes #5845

### DIFF
--- a/ext/drcovlib/drcovlib.c
+++ b/ext/drcovlib/drcovlib.c
@@ -1,5 +1,5 @@
 /* ***************************************************************************
- * Copyright (c) 2012-2021 Google, Inc.  All rights reserved.
+ * Copyright (c) 2012-2023 Google, Inc.  All rights reserved.
  * ***************************************************************************/
 
 /*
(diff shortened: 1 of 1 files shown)


Desperado17:
Would %lu work for format specifier?

derekbruening:
> Would %lu work for format specifier?

`l` is 32-bit on Windows, but I assume you mean something like the DR-provided `PIFMT` (or possibly`%zu`) for a pointer-size value (which is what the `drtable` size is defined as) to print out a 64-bit count for 64-bit builds.  The problem is that drcov2lcov needs a bunch of changes to handle a 64-bit count and it did not seem worth that effort here since we have never seen anything close to 4 billion blocks: that is an awful lot of code.  Programs can scale to huge amounts of data but huge amounts of code is a different matter.

PR #5856, opened 2023-02-06
Update a couple small document details.

brettcoon:
Fixed the example good vs. bad NULL checks to make them equivalent, and clarified the expected starting directory for the cmake command to enable tests.


api/docs/code_style.dox
```diff
@@ -161,11 +161,11 @@ strcmp() and memcmp(), the use of ! is counter-intuitive.
```
derekbruening, line 161: > iX: Update a couple small document details.

style: While we use `iX-` as a branch name prefix, we don't put it in the commit title.

derekbruening:
add to allowlist

derekbruening (approved):

Issue #5857, opened 2023-02-07
Cleanup: Make C++ constant style consistent and remove CONSTEXPR define


derekbruening:
We have two cleanup issues related to C++ constants:

1. Now that do not support VS2013 and have C++11 everywhere we can remove this in clients/drcachesim/common/utils.h:
```
/* TODO(i#2924): Remove this and others like it once we stop supporting VS2013. */
#if defined(WINDOWS) && _MSC_VER < 1900
#    define CONSTEXPR const /* 'constexpr' not supported */
#else
#    define CONSTEXPR constexpr
#endif
```

2. We should make the constant variable names consistent in drcachesim.  The official style from https://dynamorio.org/page_code_style.html#autotoc_md122 is to use all caps, yet a number of files are using kCamelCase.

PR #5862, opened 2023-02-09
i#5857: Remove CONSTEXPR define

lihasgupta:
Now that we do not support VS2013 and have C++11 everywhere we can remove this in clients/drcachesim/common/utils.h:
```
/* TODO(i#2924): Remove this and others like it once we stop supporting VS2013. */
#if defined(WINDOWS) && _MSC_VER < 1900
#    define CONSTEXPR const /* 'constexpr' not supported */
#else
#    define CONSTEXPR constexpr
#endif
```

This PR removes this.

Issue #5857


derekbruening (approved):

derekbruening:
add to allowlist

PR #5865, opened 2023-02-09
Add unit tests for basic cache behavior.

brettcoon:


clients/drcachesim/tests/drcachesim_unit_tests.cpp
```diff
@@ -306,9 +308,201 @@ LLC {
...
+static int
+generate_1D_accesses(cache_t &cache, addr_t start_address, int step_size, int step_count,
+                     int loop_count = 1)
+{
+    return generate_2D_accesses(cache, start_address, step_size, step_count, 0, 1,
+                                loop_count);
+}
+
+// Helper code to grab a snapshot of cache stats.
+struct CStats {
```
derekbruening, line 349: style: lower-case with underscores and ending in _t for types

clients/drcachesim/tests/drcachesim_unit_tests.cpp
```diff
@@ -306,9 +308,201 @@ LLC {
...
+    return cs;
+}
+
+// Creates and tests LRU caches in a range of assocativities, verifying
+// the associativity works as expected.
+void
+unit_test_cache_associativity()
+{
+    // Range of associativities to be tested.
+    constexpr int MIN_ASSOC = 1;
```
derekbruening, line 371: Seems like `constexpr int` may as well always also be declared `static`

clients/drcachesim/tests/drcachesim_unit_tests.cpp
```diff
@@ -306,9 +308,201 @@ LLC {
...
+    constexpr int BLOCKS_PER_WAY = 16;
+
+    // Only power-of-two associativities are currently supported.
+    for (int assoc = MIN_ASSOC; assoc <= MAX_ASSOC; assoc *= 2) {
+        int total_size = LINE_SIZE * BLOCKS_PER_WAY * assoc;
+        // Test access patterns that stress increasing associativity.
+        for (uint32_t test_assoc = 1; test_assoc <= 2 * assoc; ++test_assoc) {
+            cache_lru_t cache;
+            caching_device_stats_t stats(/*miss_file=*/"", LINE_SIZE);
+            assert(cache.init(assoc, LINE_SIZE, total_size, /*parent=*/nullptr, &stats));
```
derekbruening, line 384: Hmm, `NDEBUG` is explicitly undefined above, but having a required call inside a macro/function that could turn it into a nop makes me nervous: what if someone transplanted this line to another file where asserts were nops in release build.  Seems safer to have the cache.init outside the assert.

clients/drcachesim/tests/drcachesim_unit_tests.cpp
```diff
@@ -306,9 +308,201 @@ LLC {
...
+
+            constexpr int NUM_LOOPS = 3; // Anything >=2 should work.
+            auto read_count =
+                generate_2D_accesses(cache, start_address, LINE_SIZE, BLOCKS_PER_WAY,
+                                     total_size, test_assoc, NUM_LOOPS);
+            CStats c_stats = get_cache_stats(stats);
+            assert(read_count == NUM_LOOPS * BLOCKS_PER_WAY * test_assoc);
+
+            int expected_misses =
+                (test_assoc <= assoc) ? BLOCKS_PER_WAY * test_assoc : read_count;
```
derekbruening, line 396: (OK, because it's LRU.  Where did we request LRU --- ah ok took me a while to see the type `cache_lru_t`; kept looking for a param in the cache.init call.  Worth a comment?  Or maybe it would be obvious to a more careful reader.)

clients/drcachesim/tests/drcachesim_unit_tests.cpp
```diff
@@ -306,9 +308,201 @@ LLC {
...
+    constexpr int LINE_SIZE = 64;
+    constexpr int ASSOCIATIVITY = 2;
+
+    for (int cache_size = MIN_TOTAL_SIZE; cache_size <= MAX_TOTAL_SIZE; cache_size *= 2) {
+        // Access a buffer of increasing size, make sure hits + misses are expected.
+        for (int buffer_size = cache_size / 2; buffer_size < cache_size * 2;
+             buffer_size *= 2) {
+            cache_lru_t cache;
+            caching_device_stats_t stats(/*miss_file=*/"", LINE_SIZE);
+            cache.init(ASSOCIATIVITY, LINE_SIZE, cache_size, /*parent=*/nullptr, &stats);
```
derekbruening, line 428: Seems worth checking the return value.

clients/drcachesim/tests/drcachesim_unit_tests.cpp
```diff
@@ -306,9 +308,201 @@ LLC {
...
+        }
+    }
+}
+
 int
 main(int argc, const char *argv[])
 {
+    unit_test_cache_line_size();
+    unit_test_cache_size();
+    unit_test_cache_associativity();
```
derekbruening, line 505: Hmm, seems nicer to have this order match the function definition order above?

clients/drcachesim/tests/drcachesim_unit_tests.cpp
```diff
@@ -306,9 +308,201 @@ LLC {
...
+        // Stride through the cache at a test line size.  If test line size is
+        // less than actual line size, there will be cache hits.  If test line
+        // size is larger than actual line size, there will be fewer misses than
+        // lines in the cache.
+        for (uint32_t stride = line_size / 2; stride < line_size * 2; stride *= 2) {
+            int cache_line_count = BLOCKS_PER_WAY * ASSOCIATIVITY;
+            int total_cache_size = line_size * cache_line_count;
+            cache_t cache;
+            caching_device_stats_t stats(/*miss_file=*/"", line_size);
+            cache.init(ASSOCIATIVITY, line_size, total_cache_size, /*parent=*/nullptr,
```
derekbruening, line 475: Check return value

derekbruening (approved):
Thank you for adding more tests.

clients/drcachesim/tests/drcachesim_unit_tests.cpp
brettcoon: Done.

clients/drcachesim/tests/drcachesim_unit_tests.cpp
brettcoon: Done.

clients/drcachesim/tests/drcachesim_unit_tests.cpp
brettcoon: Done.

clients/drcachesim/tests/drcachesim_unit_tests.cpp
brettcoon: You're right, it's worth highlighting why the miss count is so predictable.  I added a comment.

clients/drcachesim/tests/drcachesim_unit_tests.cpp
brettcoon: Oops, thought I had fixed all those.  Done.

clients/drcachesim/tests/drcachesim_unit_tests.cpp
brettcoon: Done.

clients/drcachesim/tests/drcachesim_unit_tests.cpp
brettcoon: Done.

derekbruening:
I forgot to request two changes:

> No description provided.

Please always provide a body even if it just duplicates the title.

Also, there is an issue covering adding drcachesim tests: #4842.  Please use it in the title and with an `Issue: #4842` in the body.  Too late to add to the commit itself but there is now a ref in this PR.

PR #5866, opened 2023-02-10
i#3044 AArch64 SVE codec: Add ADR instructions

jackgallagher-arm:
This patch adds the appropriate macros, tests and codec entries to encode the following variants:
```
ADR     <Zd>.D, [<Zn>.D, <Zm>.D, SXTW <amount>]
ADR     <Zd>.D, [<Zn>.D, <Zm>.D, UXTW <amount>]
ADR     <Zd>.<Ts>, [<Zn>.<Ts>, <Zm>.<Ts>, <extend> <amount>]
```

and the required changes to support the use of vector registers in base+disp adress operands.

This required two main changes:

1) Adding element size to base disp operand

 ADR uses Z vector registers for the base and index register so we
 need to be able to specify the element size in the operand.

 This adds an element size field to base+disp operands (AArch64
 only). The following sizes are supported:
     OPSZ_4: Single
     OPSZ_8: Double

 For example, the memory operand for:
```
     adr z0.d, [z1.d, z2.d, lsl 2]
```

 could be created with the call:
```
  opnd_create_vector_base_disp_aarch64(DR_REG_Z1, DR_REG_Z2,
                                       OPSZ_8, // Element size
                                       DR_EXTEND_UXTX, // LSL
                                       true, 0, 0,
                                       OPSZ_0, // Transfers 0 bytes
                                       2); // Shift amount
```
 This will also be needed for SVE scatter/gather instructions.

2) Move DR_REG_Z* < 256

 opnd_t only stores the first 8 bits of the reg_id_t values for the base
 and index, so in order to use a Z register in an address operand we need
 to make sure the DR_REG_Z* constants are < 256.

 While I was there I also added the Z and P registers and one system
 register to the dr_reg_fixer array as they were previously missing.
 The B, H, S, D, Q registers have been changed to map to the Z registers
 because they overlap with the lower 128 bits of the Z registers.

Issues: #3044

### DIFF
--- a/core/ir/opnd_api.h
+++ b/core/ir/opnd_api.h
@@ -650,6 +650,54 @@ enum {
     DR_REG_R15,                               /**< The "r15" register. */
 #    endif
 
+#    ifdef AARCH64
+    /* SVE vector registers */
+    DR_REG_Z0,  /**< The "z0" register. */
+    DR_REG_Z1,  /**< The "z1" register. */
+    DR_REG_Z2,  /**< The "z2" register. */
+    DR_REG_Z3,  /**< The "z3" register. */
+    DR_REG_Z4,  /**< The "z4" register. */
+    DR_REG_Z5,  /**< The "z5" register. */
+    DR_REG_Z6,  /**< The "z6" register. */
+    DR_REG_Z7,  /**< The "z7" register. */
+    DR_REG_Z8,  /**< The "z8" register. */
+    DR_REG_Z9,  /**< The "z9" register. */
+    DR_REG_Z10, /**< The "z10" register. */
+    DR_REG_Z11, /**< The "z11" register. */
+    DR_REG_Z12, /**< The "z12" register. */
+    DR_REG_Z13, /**< The "z13" register. */
+    DR_REG_Z14, /**< The "z14" register. */
+    DR_REG_Z15, /**< The "z15" register. */
+    DR_REG_Z16, /**< The "z16" register. */
+    DR_REG_Z17, /**< The "z17" register. */
+    DR_REG_Z18, /**< The "z18" register. */
+    DR_REG_Z19, /**< The "z19" register. */
+    DR_REG_Z20, /**< The "z20" register. */
+    DR_REG_Z21, /**< The "z21" register. */
+    DR_REG_Z22, /**< The "z22" register. */
+    DR_REG_Z23, /**< The "z23" register. */
+    DR_REG_Z24, /**< The "z24" register. */
+    DR_REG_Z25, /**< The "z25" register. */
+    DR_REG_Z26, /**< The "z26" register. */
+    DR_REG_Z27, /**< The "z27" register. */
+    DR_REG_Z28, /**< The "z28" register. */
+    DR_REG_Z29, /**< The "z29" register. */
+    DR_REG_Z30, /**< The "z30" register. */
+    DR_REG_Z31, /**< The "z31" register. */
+#    endif
+
+/* All registers that can be used in address operands must be before this point.
+ *
+ * Base+disp operands do not store the full reg_id_t value, only the lower
+ * REG_SPECIFIER_BITS, so any register used in addressing must be less than
+ * 1 << REG_SPECIFIER_BITS. This is checked in d_r_arch_init().
+ */
+#    if defined(AARCH64)
+    DR_REG_MAX_ADDRESSING_REG = DR_REG_Z31,
+#    else
+    DR_REG_MAX_ADDRESSING_REG = DR_REG_R15,
+#    endif
+
     /* 128-bit SIMD registers */
     DR_REG_Q0,  /**< The "q0" register. */
     DR_REG_Q1,  /**< The "q1" register. */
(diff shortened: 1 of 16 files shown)


derekbruening:
Requesting opinion from @abhinav92003 as well: for x86 scatter-gather there is no element size in the IR today?  Could it be added using this same mechanism?

derekbruening:
Sorry for the delay -- but I can't seem to find the earlier discussion on element sizes which I'm pretty sure we did discuss.  Could you add points to the issue or PR where we had earlier discussions?

derekbruening:
OK I found #5681 where element size was added to registers via a size field in opnd_t next to the reg value.  That is what I was remembering.  So this is now wanting the element size for a memory operand which can't use the same field that was added before.

core/ir/opnd_api.h
```diff
@@ -632,22 +632,70 @@ enum {
...
+    DR_REG_R11, /**< The "r11" register. */
+    DR_REG_R12, /**< The "r12" register. */
+    DR_REG_R13, /**< The "r13" register. */
+    DR_REG_R14, /**< The "r14" register. */
+    DR_REG_R15, /**< The "r15" register. */
+#    endif
+
+#    ifdef AARCH64
+    /* SVE vector registers */
+    DR_REG_Z0,  /**< The "z0" register. */
```
derekbruening, line 655: This is a binary compatibility break.  Please document this in release.dox.  Officially we should also bump the version again from 9.90 to 9.91 per https://dynamorio.org/page_new_release.html#autotoc_md227

core/ir/opnd_api.h
```diff
@@ -1858,6 +1872,10 @@ struct _opnd_t {
             byte /*bool*/ scaled : 1;
             /* Shift offset amount */
             byte /*uint*/ scaled_value : 3;
+            /* Indicates the element size for vector base and index registers.
+             * This is ignored if the base and index registers are scalar registers.
```
derekbruening, line 1876: The reader will not understand how 1 bit can indicate the size: please point at #ELEMENT_SIZE_SINGLE and its counterpart.

core/ir/opnd_api.h
```diff
@@ -2271,6 +2289,28 @@ opnd_t
 opnd_create_base_disp_aarch64(reg_id_t base_reg, reg_id_t index_reg,
                               dr_extend_type_t extend_type, bool scaled, int disp,
                               dr_opnd_flags_t flags, opnd_size_t size);
+
+DR_API
+/**
+ * Same as opnd_create_base_disp_shift_aarch64 but creates an operand that uses vector
+ * registers for the base and/or index.
+ * At least one of \p base_reg and \p index_reg should be a vector register.
+ * \p element_size indicates the element size for any vector registers used and must be
```
derekbruening, line 2298: To make sure I understand: this PR is not talking about element sizes of loaded or stored regions of memory (that would seem to want a discussion of generalizing into other memory block breakdowns like LDP or x86 PUSHA): this is only talking about the ADR opcode which constructs a whole bunch of addresses from elements in vector registers, right?  Or is there a plan to use this for scatter/gather too?

If this is only for ADR: could we avoid this IR change altogether by *not* considering ADR with vector registers to have a memory operand at all?  If we treat it as just an arithmetic vector operation than we can use the existing register operand element size.  I know x86 LEA today has a memory operand, partly because that's how it's decoded and encoded: but that causes headaches and we have to have exclusions every time we do client operations on memory operands.

derekbruening (commented):
Please see the comment asking whether we can avoid ADR having a memory operand at all.

core/ir/opnd_api.h
jackgallagher-arm: > this PR is not talking about element sizes of loaded or stored regions of memory

Correct. Its the size of the elements in the vector register.

> Or is there a plan to use this for scatter/gather too?

Yes. That is the plan. SVE has scatter/gather instructions that use Z registers for the base and index registers. For example:
```
LD1B   { <Zt>.S }, <Pg>/Z, [<Zn>.S{, #<imm>}]
LDNT1B { <Zt>.S }, <Pg>/Z, [<Zn>.S{, <Xm>}]
ST1B   { <Zt>.D }, <Pg>,   [<Xn|SP>, <Zm>.D, <mod>]
```
We plan to use the same mechanism to encode the element size for any memory operands that use vector registers.

derekbruening: If it will be used for scatter/gather: the question is whether we could use the same implementation for x86 scatter/gather.  Just 1 bit is added here as there are only 2 sizes.  After talking to @abhinav92003 it does look like that would work for today's x86 scatter/gather which also only have 2 sizes.

AssadHashmi: My understanding is that the memory signature, i.e. vector base + scaled vector index, is unique to the `ADR` instruction:
```
[<Zn>.D,<Zm>.D{,LSL #<sh>}]  <-- 64 bit form

[<Zn>.S,<Zm>.S{,LSL #<sh>}]      \__ 32 bit form
[<Zn>.D,<Zm>.D,(S|U)XTW {#<sh>}] /
```
and is not directly supported by SVE's load/store and prefetch instructions. I could be wrong though, @jackgallagher-arm  has been looking at this more recently than me and my knowledge may be out-of-date.

> this PR is not talking about element sizes of loaded or stored regions of memory (that would seem to want a discussion of generalizing into other memory block breakdowns like LDP or x86 PUSHA): this is only talking about the ADR opcode which constructs a whole bunch of addresses from elements in vector registers, right?

Yes, it's not about _actual memory load/stores_, only the _construction of addresses_ for use in later actual memory load/stores.

> If this is only for ADR: could we avoid this IR change altogether by not considering ADR with vector registers to have a memory operand at all?

We thought having a memory operand with zero bytes transferred would be a convenience to the user, indicating the purpose of the operands in the IR as memory addresses. Like the prefetch instructions which indicate a memory address but don't transfer anything. But it looks like in this case that's not what you would like. If so, what should the IR look like? Just plain registers and scaling modifiers? e.g.
```
adr z0.d, [z0.d, z0.d, SXTW] : adr %z0.d,%z0.d,sxtw -> %z0.d
adr z0.d, [z0.d, z0.d, UXTW] : adr %z0.d,%z0.d,uxtw -> %z0.d

adr z0.s, [z0.s, z0.s] : adr %z0.s,%z0.s -> %z0.s
adr z0.s, [z0.s, z0.s, LSL #1] : adr %z0.s,%z0.s,lsl #1 -> %z0.s
adr z31.s, [z31.s, z31.s, LSL #3] : adr %z31.s,%z31.s,lsl #3 -> %z31.s
```

derekbruening: Thinking about it more: since the existing ADR (with GPR's) and ADRP are modeled as memory operands, and LEA on x86 is as well, to be consistent we would have to make a compatibility change and go change all of those as well if we wanted to not have these compute-address opcodes have memory operands.  That may be more than we want to do.  So maybe the solution here is the best way forward.

derekbruening: Not sure whether @abhinav92003 or @khuey have different opinions.

AssadHashmi: > So maybe the solution in this PR is the best way forward.

We are happy with this if you are.

> Not sure whether @abhinav92003 or @khuey have different opinions.

Let us know what you think @abhinav92003 @khuey.

derekbruening: @abhinav92003 has been traveling/out of office.  Let's just go ahead with this PR as-is.

AssadHashmi (approved):

derekbruening:
This one too has "no response for status checks": is it Jenkins which does not seem to be in the list of "13 checks passed"?

Issue #5871, opened 2023-02-14
Add decoding to invariant_checker to confirm branches go to their targets


derekbruening:
Today our invariant checker does not look for PC discontinuities after a branch b/c it has no decoder.

I have a local branch where I started adding decoder use to the checker for looking at ud2a in our signal test but I hit build issues w/ dup syms on Windows plus complexities with the module file and using raw2trace's mod mapper and I abandoned that branch: still have it locally.  Now that we have encodings in the trace itself it will be much simpler (no mod mapper; may still hit link issues).

derekbruening:
I just shared my local branch with the decoder for ud2a (looks like an opcode_mix bug fix too; I will send that one separately): https://github.com/DynamoRIO/dynamorio/compare/master...i5871-add-decoding-to-trace_invariants

derekbruening:
Looks like there is a (pre-existing) oversight here: an uncondtional branch cannot fall through.

derekbruening:
I don't think there's anything missing here: I see `fall_through_allowed` in `invariant_checker_t::check_for_pc_discontinuity()` so it looks like the concern above is handled.  Closing.

PR #5872, opened 2023-02-14
Fix missing free on opcode_mix error path

derekbruening:
Adds a missing instr_free on an opcode_mix error path found in a tree for #5871.


derekbruening:
Not clear how to easily add a test as we'd want leak check enabled to fail w/o this fix...probably just going to put the fix in as an eyeballed-correct one-line change.

derekbruening:
Unfortunately the queue is just waiting for Jenkins which will never run so I have to remove from the queue.

PR #5885, opened 2023-02-24
i#5734: disable rseq test on AArch64 runners

joshua-warburton:
The rseq test fails on various different machines, and in different contexts including during postcommit on the test runner, but not on precommit. Ignoring this by default due to the flakiness

issue: #5734


AssadHashmi (approved):

derekbruening:
#5734 is about glibc 2.35: is the Jenkins machine on that glibc now?

PR #5896, opened 2023-03-07
i#5784: Mitigate drwrap retaddr transparency violation

abhinav92003:
Adds a new drwrap API, drwrap_get_retaddr_if_sentinel(), that allows
mitigation of a transparency violation under the DRWRAP_REPLACE_RETADDR drwrap
strategy where the return address on the stack is replaced with the address of
the internal replace_retaddr_sentinel() routine. This API modifies the passed-in
value to the actual return address of the inner-most nested wrapped function if
the passed-in value is replace_retaddr_sentinel() itself.

Fixes the value of the marker written by the kernel xfer event in drmemtrace by
using the new drwrap_get_retaddr_if_sentinel() API on the mcontext PC before
writing it out to the trace. Before, this caused many invariant errors of type 'Signal
handler return point incorrect' in traces collected on proprietary apps.

Verified on a large proprietary app that this error due to drwrap is fixed now,
whereas there were a few hundred instances before.

Fixes: #5784


ext/drwrap/drwrap.h
```diff
@@ -887,6 +889,18 @@ DR_EXPORT
 bool
 drwrap_get_stats(INOUT drwrap_stats_t *stats);
 
+DR_EXPORT
+/**
+ * If the provided app_pc (\p possibly_sentinel) is indeed the address of the
+ * internal replace_retaddr_sentinel(), this routine replaces it with the actual
```
derekbruening, line 895: replace_retaddr_sentinel is not a public symbol and is not mentioned anywhere else in public docs: probably better to just say "the return address sentinel used to implement #DRWRAP_REPLACE_RETADDR".  Ditto below.

ext/drwrap/drwrap.h
```diff
@@ -416,8 +416,10 @@ typedef enum {
      * However, this does violate transparency, and may cause some applications to fail.
      * In particular, detaching on AArchXX requires scanning the stack to find where the
      * return address was stored, which could conceivably replace an integer or
-     * non-pointer value that happens to match the sentinel used.  Use this at your own
-     * risk.
+     * non-pointer value that happens to match the sentinel used. Also, the transparency
+     * violation may be exposed to the client's kernel xfer event callback if it
```
derekbruening, line 420: Referencing `dr_register_kernel_xfer_event()` should make this clearer

ext/drwrap/drwrap.h
```diff
@@ -887,6 +889,18 @@ DR_EXPORT
...
+/**
+ * If the provided app_pc (\p possibly_sentinel) is indeed the address of the
+ * internal replace_retaddr_sentinel(), this routine replaces it with the actual
+ * return address of the inner-most nested wrapped function. Otherwise, it is a no-op.
+ * This allows mitigation of a transparency violation under the #DRWRAP_REPLACE_RETADDR
+ * strategy where the return address on the stack is replaced with the address of the
+ * internal replace_retaddr_sentinel().
+ */
+void
+drwrap_replace_if_retaddr_sentinel(void *drcontext, INOUT app_pc *possibly_sentinel);
```
derekbruening, line 902: The name is too close to drwrap_replace() and related routines: I think it has to be changed to not start with those 2 words.  `drwrap_update_*` or `drwrap_substitute_*` or sthg?

ext/drwrap/drwrap.c
```diff
@@ -2771,6 +2771,19 @@ drwrap_get_stats(INOUT drwrap_stats_t *stats)
...
+{
+    ASSERT(possibly_sentinel != NULL, "Input cannot be null.");
+    if ((app_pc)replace_retaddr_sentinel != *possibly_sentinel)
+        return;
+    per_thread_t *pt = (per_thread_t *)drmgr_get_tls_field(drcontext, tls_idx);
+    /* If we see the sentinel, we must be inside a wrapped function. */
+    ASSERT(pt != NULL && pt->wrap_level >= 0, "Invalid drwrap state.");
+    *possibly_sentinel = pt->retaddr[pt->wrap_level];
+}
+
```
derekbruening, line 2786: Is there any simple way to add a test either to drwrap or drcachesim?  Might be too complex to set up.

derekbruening (approved):

ext/drwrap/drwrap.h
abhinav92003: Done.

ext/drwrap/drwrap.h
abhinav92003: Done.

ext/drwrap/drwrap.h
abhinav92003: I replaced it with `drwrap_get_retaddr_if_sentinel`: clearly says it only "gets" so no side-effect, also tells what is returned (the retaddr).

abhinav92003: It's also different than the existing routines you pointed out.

ext/drwrap/drwrap.c
abhinav92003: Right. It'll be hard to reliably reproduce the signal-at-wrapped-func-end condition. And the drwrap test would require us to know what the sentinel address is, which isn't available in the public API.

ext/drwrap/drwrap.h
derekbruening: > It's also different than the existing routines you pointed out.

The concern is that there are two different drwrap modes: wrap and replace.  The high-level routines for those start with drwrap_wrap* and drwrap_replace*.  Seeing a function starting with drwrap_replace* makes me think it is related to replace mode, which it is not.  `drwrap_get_retaddr_if_sentinel` SGTM.

abhinav92003: > The concern is that there are two different drwrap modes: wrap and replace.

I  understand.

Issue #5843, opened 2023-01-31
Create simulator interface which provides thread scheduling + speculative fetching


derekbruening:
Rather than having each simulator figure out how to schedule traced software threads onto simulated cores in their own ad hoc way, we would like to provide a scheduler service, which should result in several benefits:

+ Ease of use: a new simulator use case has one less thing to implement
+ Consistency: all simulators can now use the same approach
+ Fill in gaps in trace-based simulation:
  - We can re-schedule threads even when simulating the recorded hardware to deflate context switches increased by tracing overhead
  - We can more easily combine multiple single-workload traces
  - We can provide speculative path fetching using various schemes (from heuristics to additional data recorded during tracing) to help bridge the gap with execution-driven simulation

Xref #5694: provide per-core iterator.
That may become subsumed by this new broader-scope feature.

derekbruening:
There are many design points here; documenting some smaller ones and will probably put the rest in a separate doc:

Lots of little issues with the scheduler -- here is one: the output streams
have to keep their own record counts (b/c they combine multiple inputs).
Yet the inputs sometimes "hide" records like the synthetic headers after a
skip:
```
<--record#-> <--instr#->: <---tid---> <record details>
------------------------------------------------------------
           0          63:      296231 <marker: timestamp 13319413770947393>
           0          63:      296231 <marker: tid 296231 on core 10>
          90          64:      296231 ifetch       4 byte\(s\) @ 0x0000000000401028 48 83 eb 01          sub    \$0x0000000000000001 %rbx -> %rbx
          91          65:      296231 ifetch       4 byte\(s\) @ 0x000000000040102c 48 83 fb 00          cmp    %rbx \$0x0000000000000000
```
The scheduler though does the skip and asks the zipfile reader what the new
record# is so it can update its count and is told 0 so it goes from there
and doesn't see the 90 2 entries later (only queries input on a skip):
```
<--record#-> <--instr#->: <---tid---> <record details>
------------------------------------------------------------
           0          63:      296231 <marker: timestamp 13319413770947393>
           1          63:      296231 <marker: tid 296231 on core 10>
           2          64:      296231 ifetch       4 byte(s) @ 0x0000000000401028 48 83 eb 01          sub    $0x0000000000000001 %rbx -> %rbx
           3          65:      296231 ifetch       4 byte(s) @ 0x000000000040102c 48 83 fb 00          cmp    %rbx $0x0000000000000000
```

What is the best solution?

[...]

PR #5900, opened 2023-03-09
i#5843 scheduler: Refactor readers to be single-input

derekbruening:
Removes multi-input support from file_reader_t and other readers now that the scheduler_t owns that.  Specifically:

+ Removes read_next_thread_entry() and requires that read_next_entry() always check the queue (via a provided helper function).

+ Removes skip_thread_instructions() and refactors the pre-skip header reading and the post-skip walking while remembering timestamps. Places these latter two inside reader_t for use by all readers, with zipfile overriding just the fast skip in the middle and sharing all the other code.  This refactoring and sharing solves the problem of missing timestamps when skipping from the middle.

+ Removes the arrays of data for multiple inputs from file_reader_t and all subclasses.

Updates the view_test to use a scheduler for its multiple-input mock reader.

While at it, removes is_complete().

Issue: #5843, #5538


clients/drcachesim/reader/reader.h
```diff
@@ -182,24 +189,21 @@ class reader_t : public std::iterator<std::input_iterator_tag, memref_t>,
...
     virtual trace_entry_t *
     read_next_entry() = 0;
-    // This first checks the local queue before calling read_next_entry().
-    // in timestamp order.
+    // Returns and removes the entry (nullptr if none) from the local queue.
+    // This should be called by read_next_entry() prior to reading a new record
+    // from the input stream.
     virtual trace_entry_t *
-    read_next_entry_or_queue();
+    read_queue();
```
abhinav92003, line 202: s/read_queue/read_buffered_entry/

Or some other name that makes it clear what the queue is.

clients/drcachesim/reader/file_reader.h
```diff
@@ -107,287 +95,83 @@ template <typename T> class file_reader_t : public reader_t {
...
+        // early marker with the arch tag, and version 2 which only differs wrt
+        // TRACE_MARKER_TYPE_KERNEL_EVENT..
+        if (entry->addr > TRACE_ENTRY_VERSION) {
+            ERRMSG("Cannot handle version #%zu (expect version <= #%u)\n", entry->addr,
+                   TRACE_ENTRY_VERSION);
+            return false;
+        }
+        // Read the meta entries until we hit the pid.
+        // We want to pass the tid+pid to the reader *before* any markers,
+        // even though 2 markers preced the tid+pid in the file.
```
abhinav92003, line 134: s/preced/precede/

clients/drcachesim/reader/file_reader.h
```diff
@@ -107,287 +95,83 @@ template <typename T> class file_reader_t : public reader_t {
...
+        std::queue<trace_entry_t> marker_queue;
+        while ((entry = read_next_entry()) != nullptr) {
+            if (entry->type == TRACE_TYPE_PID) {
+                // We assume the pid entry is the last, right before the timestamp.
+                pid = *entry;
+                break;
+            } else if (entry->type == TRACE_TYPE_THREAD)
+                tid = *entry;
+            else if (entry->type == TRACE_TYPE_MARKER)
+                marker_queue.push(*entry);
```
abhinav92003, line 144: Add to comment which markers we expect here. Looking at `offline_instr_t::append_thread_header` I see only the OFFLINE_EXT_TYPE_HEADER entry before the tid one but no marker.

clients/drcachesim/reader/reader.h
```diff
@@ -110,18 +110,25 @@ class reader_t : public std::iterator<std::input_iterator_tag, memref_t>,
...
+    // timestamp (plus cpuid) is skipped, the most recent skipped timestamp will be
+    // duplicated prior to the target instruction.  Top-level headers will still be
+    // observed.  This generally should call pre_skip_instructions() to observe the
+    // headers, perform any fast skipping, and then should call
+    // skip_instructions_with_timestamp() to properly duplicate the prior timestamp.
     virtual reader_t &
     skip_instructions(uint64_t instruction_count);
 
+    virtual void
+    pre_skip_instructions();
```
abhinav92003, line 123: Why are pre_skip_instructions and skip_instructions_with_timestamp public?

abhinav92003 (approved):

clients/drcachesim/reader/file_reader.h
derekbruening: Done.

clients/drcachesim/reader/file_reader.h
derekbruening: It is legacy traces which have something different: I think they used to have the timestamp before the tid+pid in the raw file.  Updated the comment.

clients/drcachesim/reader/reader.h
derekbruening: Shouldn't be: moved to protected.

Issue #5906, opened 2023-03-13
drbbdup does not restore regs at last instruction of a rep-expanded block


prasun3:
**Describe the bug**
drbbdup does not restore regs at last instruction of a rep-expanded block

**To Reproduce**
See https://github.com/DynamoRIO/dynamorio/pull/5717#issuecomment-1457764812

**Expected behavior**
regs should be restored at the end of the block

**Screenshots or Pasted Text**

**Versions**
 - What version of DynamoRIO are you using? 
   - cronbuild-9.91.19426
 - Does the latest build from https://github.com/DynamoRIO/dynamorio/releases solve the problem?
   - No
- What operating system version are you running on? ("Windows 10" is *not* sufficient: give the release number.)
- Is your application 32-bit or 64-bit?
  - The failure was seen in Linux x86-32, but the problem could happen on other platforms too depending on application/instrumentation code

**Additional context**

Issue #5888, opened 2023-02-28
Windows 32b failure: Error copying file drconfiglib.dll


prasun3:
Seeing this failure in #5717 

```
2023-02-27T04:29:11.5101083Z ==================================================
2023-02-27T04:29:11.5101644Z RESULTS
2023-02-27T04:29:11.5101998Z 
2023-02-27T04:29:11.5102315Z debug-internal-32: **** 1 build errors ****
2023-02-27T04:29:11.5103319Z 	Error copying file "/.../dynamorio/build_debug-internal-32/lib32/debug/../drconfiglib.dll" to "/.../dynamorio/build_debug-internal-32/suite/tests/bin/drconfiglib.dll".
2023-02-27T04:29:11.5103853Z 
2023-02-27T04:29:11.5470204Z Finished running ctest --output-on-failure -VV -S "D:/a/dynamorio/dynamorio/suite/runsuite.cmake,automated_ci;use_ninja;32_only;debug_only"
2023-02-27T04:29:11.5470987Z RESULTS
2023-02-27T04:29:11.5471336Z 
2023-02-27T04:29:11.5471487Z 
2023-02-27T04:29:11.5471769Z ====> FAILURE in debug-internal-32 <====
2023-02-27T04:29:11.5472812Z debug-internal-32: **** 1 build errors ****
2023-02-27T04:29:11.5473596Z 	Error copying file "/.../dynamorio/build_debug-internal-32/lib32/debug/../drconfiglib.dll" to "/.../dynamorio/build_debug-internal-32/suite/tests/bin/drconfiglib.dll".
2023-02-27T04:29:11.5562265Z ##[error]Process completed with exit code 1.
```

derekbruening:
Xref https://github.com/DynamoRIO/dynamorio/issues/1807#issuecomment-1061909208

derekbruening:
Xref #4668

derekbruening:
This looks like a race.  Somewhere a dependency is missing.

xdje42:
Data point: This fails on 64 bit too. Possibly the same race.
https://github.com/DynamoRIO/dynamorio/actions/runs/10894111956/job/30230273364#step:6:1407
```
RESULTS

debug-internal-64: **** 1 build errors ****
	Error copying file "/.../dynamorio/build_debug-internal-64/lib64/debug/../drconfiglib.dll" to "/.../dynamorio/build_debug-internal-64/suite/tests/bin/drconfiglib.dll".

[...]

Issue #5747, opened 2022-11-21
FAILED: 282 - code_api|tool.drcachesim.scattergather (Failed) Required regular expression not found.


prasun3:
**Describe the bug**
The ```code_api|tool.drcachesim.scattergather``` test fails with the error ```Failed  Required regular expression not found.```. On inspection, I noticed that the test output was missing this statement ```AVX2/AVX-512 scatter/gather checks ok```.

Test output
```
282: Test updating the AVX2 gather mask register upon translation events
282: Test restoring the AVX2 gather scratch xmm register upon a fault
282: ---- <application exited with code 139> ----
```
Expected
```
Test updating the AVX2 gather mask register upon translation events
Test restoring the AVX2 gather scratch xmm register upon a fault
AVX2/AVX-512 scatter/gather checks ok
---- <application exited with code 0> ----
```

**To Reproduce**
Run this on an x86-32 build

```ctest -V -R 'tool.drcachesim.scattergather'```

The test passes if the following lines are commented out in ```suite/tests/client-interface/drx-scattergather.c```
```
    intercept_signal(SIGSEGV, (handler_3_t)&signal_handler_check_xmm3, false);
    test_idx32_vec[1] = 0xefffffff;
    if (SIGSETJMP(mark) == 0) {
        test_avx2_restore_gather_scratch_xmm_fault(ref_idx32_val32_xmm_ymm_zmm,
                                                   test_idx32_vec, scratch_xmm_val);
        print("ERROR: Expected a fault\n");
    }
```
Please also answer these questions:

**Expected behavior**
Test should pass

**Screenshots or Pasted Text**

**Versions**
11bf678444ba309e9d1b97f31620d67790bed53d

**Additional context**

PR #5907, opened 2023-03-13
i#5906: restore regs at end of rep-expanded block

prasun3:
drbbdup relies on the emulation start label to detect the end of a block. Expanded rep blocks or blocks not ending in a branch don't have such a label. drbbdup has a special check for the latter (!is_last_special). Add a similar check for the former too.

Issue: #5906


prasun3:
The 32-bit failure is `api.startstop` but it fails on master too on my local run.

```
debug-internal-32: 392 tests passed, **** 5 tests failed, of which 3 were flaky, but ignoring 1 for i#2941: ****
	(ignore: i#2941) 	code_api|common.decode-stress 
	code_api|api.startstop 
	code_api|api.detach_spawn_stress_FLAKY 
	code_api|api.static_maps_mixup_novars_FLAKY 
	code_api|api.static_sideline_FLAKY 
release-external-32: build successful; no tests for this build
```

derekbruening:
Looks like `tool.drcacheoff.burst_threads_counts` failed on aarch64: I don't see an issue on it; is the diff "34 total threads" not matching " [ 1][0-9] total threads"?  If so that looks like a flake but one never seen before: http://139.178.82.61:8080/job/DynamoRIO-AArch64-Precommit/2852/consoleFull

ext/drbbdup/drbbdup.c
```diff
@@ -1590,7 +1590,8 @@ drbbdup_instrument_dups(void *drcontext, void *tag, instrlist_t *bb, instr_t *in
             }
         }
         drreg_restore_all(drcontext, bb, instr);
-    } else if (drbbdup_is_at_end(instr) && !is_last_special) {
+    } else if (drbbdup_is_at_end(instr) &&
+               (!is_last_special || drbbdup_ilist_has_unending_emulation(bb))) {
```
derekbruening, line 1594: Trying to understand this: so `is_at_end_initial` is always true for the end of all block copies, or at least it seems like that from its header comment.

But you said:

> Normal case: drbbdup_is_at_end_initial returns true, so drbbdup_instrument_instr and drreg_restore_all are called
> Basic block just before the rep block: drbbdup_is_at_end_initial is false, drreg_restore_all is called from the !is_last_special case and drbbdup_instrument_instr is called from the final else
> For the rep block: we don't go into the !is_last_special case, and we don't reach the final else because drbbdup_is_exit_jmp_emulation_marker is true

Does that mean this comment for `drbbdup_is_at_end_initial` is wrong?

> If there are no emulation labels this is equivalent to drbbdup_is_at_end().

I'm finding this very confusing: why does the original code on line 1593 have a check `!is_last_special`?

If you've got an understanding of it, maybe you could add and augment comments here and in the headers for these query functions?

prasun3: Yes, that comment looks incorrect to me. You can see that it returns false if drmgr_is_emulation_start returns false https://github.com/DynamoRIO/dynamorio/blob/f59f1b3a2183ba837177d1cac62ce7b706ed555e/ext/drbbdup/drbbdup.c#L667-L685

I think the check exists for handling the scenario that you brought up

> You're saying the drbbdup identification of the end of a copied block relies on an emulation start label, which is not there for rep-str. Wouldn't it also not be there for a block that doesn't end in a branch/syscall? Does such a block also have this problem?

derekbruening: So is the `!is_last_special` trying to detect whether control already went into the case above because `drbbdup_is_at_end_initial` returned true?  It's equating the presence of an emulation start label with a special instr being there (which requires the emulation b/c drbbdup can't put the real instr in each copy).  But if the rep expansion does have `is_last_special` why doesn't it have an emulation label for it?  Looking around: I see this explanation:
https://github.com/DynamoRIO/dynamorio/blob/f59f1b3a2183ba837177d1cac62ce7b706ed555e/ext/drbbdup/drbbdup.c#L474-L475

So it seems that the real intent of the `drbbdup_is_at_end(instr) && xxx` is to capture all cases that do not go into the `drbbdup_is_at_end_initial`, but not to trigger for any cases that do, right?  So I would suggest to not add further conditions to `xxx` which may just make everything more fragile, but to instead have a local boolean in this function which says "inserted_restore_all" and set that in the `*end_initial` case and replace `xxx` with `!inserted_restore_all`.  And please add comments to help a future reader understand all this.

prasun3: So we only want to handle those specific cases, and other cases are handled by the next else condition

https://github.com/DynamoRIO/dynamorio/blob/f59f1b3a2183ba837177d1cac62ce7b706ed555e/ext/drbbdup/drbbdup.c#L1593-L1597

derekbruening: The insertion of drreg_restore_all() is also ignoring/hiding the instruction by not calling drbbdup_instrument_instr so there is no need to hit that next else which says "Ignore": these earlier conditions are also ignoring if they don't hand the instr to drbbdup_instrument_instr.  Could combine all the drbbdup_is_at_end() into one condition with sub-conditions if that seems clearer.

ext/drbbdup/drbbdup.c
```diff
@@ -1590,11 +1589,14 @@ drbbdup_instrument_dups(void *drcontext, void *tag, instrlist_t *bb, instr_t *in
...
-    } else if (drbbdup_is_at_end(instr) &&
-               (!is_last_special || drbbdup_ilist_has_unending_emulation(bb))) {
-        drreg_restore_all(drcontext, bb, instr);
-    } else if (drbbdup_is_at_end(instr) || drbbdup_is_exit_jmp_emulation_marker(instr)) {
-        /* Ignore instruction: hide drbbdup's own markers and the rest of the end. */
+    } else if (drbbdup_is_at_end(instr)) {
+        /* i#5906: if the emulation start label is missing we might still need to restore
+         * registers for blocks that don't end in a branch or for rep-expanded blocks. */
+        if (!is_last_special || drbbdup_ilist_has_unending_emulation(bb))
+            drreg_restore_all(drcontext, bb, instr);
```
derekbruening, line 1596: This still feels fragile: if we add some other case that doesn't trigger drbbdup_is_at_end_initial() it seems easy to forget to, or not realize we have to, update here, and then have more subtle problems requiring complex debugging.  Did you have a reason to not want to take me up on my suggestion of having a boolean `restored_all` that we set to true when we call `drreg_restore_all` for `drbbdup_is_at_end_initial`, and then here for `drbbdup_is_at_end` we call `drreg_restore_all` if `!restored_all`?  That makes the code clearer IMHO (makes it clear we **always** want to restore-all, which is not clear here with these conditions) and less fragile.

ext/drbbdup/drbbdup.c
```diff
@@ -665,8 +665,7 @@ drbbdup_is_at_end(instr_t *check_instr)
 }
 
 /* Returns true if at the start of the end of a bb version: if check_instr is
- * the start emulation label for the inserted jump or the exit label.  If there
- * are no emulation labels this is equivalent to drbbdup_is_at_end().
+ * the start emulation label for the inserted jump or the exit label.
```
derekbruening, line 668: How about adding that this does not return true for certain types of blocks like not ending in a branch/syscall or having unending emulation like repstr.

ext/drbbdup/drbbdup.c
```diff
@@ -1590,10 +1589,14 @@ drbbdup_instrument_dups(void *drcontext, void *tag, instrlist_t *bb, instr_t *in
...
-    } else if (drbbdup_is_at_end(instr) || drbbdup_is_exit_jmp_emulation_marker(instr)) {
-        /* Ignore instruction: hide drbbdup's own markers and the rest of the end. */
+    } else if (drbbdup_is_at_end(instr)) {
+        /* i#5906: if the emulation start label is missing we might still need to restore
+         * registers for blocks that don't end in a branch or for rep-expanded blocks. */
+        if (!is_last_special || drbbdup_ilist_has_unending_emulation(bb))
+            drreg_restore_all(drcontext, bb, instr);
+    } else if (drbbdup_is_exit_jmp_emulation_marker(instr)) {
+        /* Ignore instruction: hide drbbdup's own markers and the rest of the end.
+         * Do not call drreg_restore_all either. */
```
derekbruening, line 1599: Could we add a test for this?  allasm_repstr has rep string instrs and is very small, so it makes a good test target and is already run for several dr$sim options: add a test running allasm_repstr with options that trigger the filter and confirm it fails without this fix and then we have a regression test?

prasun3:
Windows failure is #5888

ext/drbbdup/drbbdup.c
prasun3: Resolving this since we are discussing this below.

ext/drbbdup/drbbdup.c
prasun3: The way the original code was written, I thought there might be cases where we intentionally don't call drreg_restore_all

I have updated it now.

ext/drbbdup/drbbdup.c
prasun3: Added a comment.

ext/drbbdup/drbbdup.c
prasun3: Currently it won't fail because it isn't an issue in the TRACE case (because it is the final copy), and it isn't an issue in the COUNT case (because the instrumentation is smaller)

prasun3:
x86-32 failure is #5747

prasun3:
x86-32 failure https://github.com/DynamoRIO/dynamorio/issues/5747 seen again

ext/drbbdup/drbbdup.c
derekbruening: Not sure I'm following: you're just saying the current test doesn't fail b/c it doesn't do a filter?  But the request is to add a new test that does do a filter to hit the bug and have a regression test for this behavior.

derekbruening (commented):
Approach looks good to me now: just a nit about a simpler initial value, and then adding a test.

ext/drbbdup/drbbdup.c
prasun3: By filter, do you mean TRACE with -L0{I,D}_filter? Because that would not trigger the failure. Or do you mean L0_FILTER? The tests in #5717 depend on this fix, so those could serve as regression tests.

For a test to fail, it needs to touch flags in the non-default case (i.e., COUNT or L0_FILTER)

prasun3:
x86-32 hit both #5747 and #5908

ext/drbbdup/drbbdup.c
```diff
@@ -1551,6 +1554,7 @@ drbbdup_instrument_dups(void *drcontext, void *tag, instrlist_t *bb, instr_t *in
         instr_t *next_bb_label = drbbdup_next_start(end_instr);
         if (next_bb_label == NULL) {
             pt->case_index = DRBBDUP_DEFAULT_INDEX; /* Refer to default. */
+            pt->inserted_restore_all = false;
```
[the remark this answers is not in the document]
prasun3: Fixed. I was looking at case_index field and ended up setting in both the if and else.

ext/drbbdup/drbbdup.c
derekbruening: > probably we could easily extend that to trigger this issue by adding more bbdup cases to the client?

Having at least 3 cases should case the flags to be preserved, hitting this bug?  if so that's a simple augmentation of that existing test and we then have a bbdup regression test?

prasun3:
Bunch of new windows 64-bit failures after merging master. Was this seen anywhere else?

```
2023-03-30T03:08:43.0691006Z The following tests FAILED:
2023-03-30T03:08:43.0691293Z 	 25 - code_api|common.nativeexec (Failed)
2023-03-30T03:08:43.0691807Z 	 65 - code_api|win32.mixedmode_late (Failed)
2023-03-30T03:08:43.0692131Z 	 66 - code_api|win32.mixedmode (Failed)
2023-03-30T03:08:43.0692654Z 	 67 - code_api|win32.x86_to_x64 (Failed)
2023-03-30T03:08:43.0692981Z 	 68 - code_api|win32.x86_to_x64_ibl_opt (Failed)
2023-03-30T03:08:43.0693310Z 	 87 - code_api|client.cleancall (Failed)
2023-03-30T03:08:43.0693617Z 	109 - code_api|client.attach_test (Failed)
2023-03-30T03:08:43.0693931Z 	116 - code_api|client.loader (Failed)
2023-03-30T03:08:43.0694237Z 	118 - code_api|client.file_io (Failed)
2023-03-30T03:08:43.0694560Z 	150 - code_api|client.drx-test (Failed)
2023-03-30T03:08:43.0694891Z 	159 - code_api|client.drsyms-test (Failed)
2023-03-30T03:08:43.0695239Z 	160 - code_api|client.drsyms-testgcc (Failed)
2023-03-30T03:08:43.0695564Z 	164 - code_api|client.pcache-use (Failed)
2023-03-30T03:08:43.0695929Z 	262 - code_api|tool.drcpusim.cpuid-Nehalem (Failed)
2023-03-30T03:08:43.0696272Z 	274 - code_api|api.detach (Failed)
```

ext/drbbdup/drbbdup.c
prasun3: I have added a test based off of drbbdup-emul-test.

suite/tests/client-interface/drbbdup-emul-reg-clobber-test.c
```diff
@@ -0,0 +1,95 @@
...
+          : [x] "=r"(x)
+          : [buf1] "m"(buf1), [buf2] "m"(buf2)
+          : "ecx", "rdi", "rsi", "rax", "memory");
+
+#        else
+
+    uint32_t x = 0x11111111;
+
+    /* Put a magic value in eax to test if it was corrupted. Using eax because
+     * it is used to spill flags.
```
derekbruening, line 74: Switching this file to be a split C-asm file with ASM_CODE_ONLY, etc. would let us have a single asm sequence for 32-bit and 64-bit Linux and Windows: see other files with ASM_CODE_ONLY where the asm uses REG_XAX, etc. for cross-bitwidth names.  I think this is worth doing as it should not be much work (this is a short asm sequence) and it eliminates the duplicate asm and the WIndows TODO items.  We have this same-file infrastructure to make it easy to have cross-platform asm functions w/o a bunch of separate files.

suite/tests/client-interface/drbbdup-emul-reg-clobber-test.dll.c
```diff
@@ -0,0 +1,311 @@
...
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+/* Tests the drbbdup extension's interactions with emulation. */
```
derekbruening, line 33: Please update this to explain what is actually being tested here.

suite/tests/client-interface/drbbdup-emul-reg-clobber-test.dll.c
```diff
@@ -0,0 +1,311 @@
...
+    /* Test drutil rep string expansion interacting with drbbdup. */
+    bool expanded;
+    if (!drutil_expand_rep_string_ex(drcontext, bb, &expanded, NULL))
+        DR_ASSERT(false);
+    if (expanded) {
+        /* We can't overlap ours with drutil's so bow out. */
+        has_rest_of_block_emul = true;
+        return DR_EMIT_DEFAULT;
+    }
+    /* Test handling of DR_EMULATE_REST_OF_BLOCK by inserting it in the middle
```
derekbruening, line 76: This is all an unmodified copy from drbbdup-emul-test: but there is no need to duplicate all the checks done in that test here, right?  Better to remove everything except what's needed to test the issue at hand: then we'll have a clearer, more focused test.  As it is, a reader will look at this and have no idea what's different vs the other test: there aren't even any comments saying how it's different.

suite/tests/client-interface/drbbdup-emul-reg-clobber-test.dll.c
```diff
@@ -0,0 +1,311 @@
...
+    emulated_instr_t emulated_instr;
+    emulated_instr.size = sizeof(emulated_instr);
+    emulated_instr.pc = instr_get_app_pc(mid);
+    emulated_instr.instr = instr_clone(drcontext, mid);
+    emulated_instr.flags = DR_EMULATE_REST_OF_BLOCK;
+    drmgr_insert_emulation_start(drcontext, bb, mid, &emulated_instr);
+    /* XXX i#5400: We'd like to pass a user_data to the instrument event but drbbdup
+     * doesn't support that; instead we rely on being single-threaded.
+     */
+    has_rest_of_block_emul = true;
```
derekbruening, line 103: All of this can be deleted (except the drutil_expand call above), right?

suite/tests/client-interface/drbbdup-emul-reg-clobber-test.dll.c
```diff
@@ -0,0 +1,311 @@
...
+{
+    drbbdup_status_t res;
+
+    CHECK(enable_dups != NULL, "should not be NULL");
+    CHECK(enable_dynamic_handling != NULL, "should not be NULL");
+
+    res = drbbdup_register_case_encoding(drbbdup_ctx, 1);
+    CHECK(res == DRBBDUP_SUCCESS, "failed to register case 1");
+    res = drbbdup_register_case_encoding(drbbdup_ctx, 2);
+    CHECK(res == DRBBDUP_SUCCESS, "failed to register case 2");
```
derekbruening, line 120: I thought the plan was to add a 3rd case to trigger flags reservation in drbbdup.

suite/tests/client-interface/drbbdup-emul-reg-clobber-test.dll.c
```diff
@@ -0,0 +1,311 @@
...
+}
+
+static void
+write_aflags(void *drcontext, instrlist_t *bb, instr_t *inst)
+{
+    instrlist_meta_preinsert(bb, inst,
+                             XINST_CREATE_cmp(drcontext, opnd_create_reg(DR_REG_START_32),
+                                              OPND_CREATE_INT32(0)));
+    instrlist_meta_preinsert(bb, inst,
+                             XINST_CREATE_cmp(drcontext, opnd_create_reg(DR_REG_START_32),
```
derekbruening, line 156: Why a 2nd cmp?

suite/tests/client-interface/drbbdup-emul-reg-clobber-test.dll.c
```diff
@@ -0,0 +1,311 @@
...
+    drbbdup_status_t res;
+    per_block_t *per_block = (per_block_t *)analysis_data;
+
+    res = drbbdup_is_first_instr(drcontext, instr, &is_first);
+    CHECK(res == DRBBDUP_SUCCESS, "failed to check whether instr is first");
+    if (is_first) {
+        per_block->saw_first = true;
+        /* Ensure DR_EMULATE_REST_OF_BLOCK didn't leak through. */
+        const emulated_instr_t *emul_info;
+        if (drmgr_in_emulation_region(drcontext, &emul_info)) {
```
derekbruening, line 175: This can all go away, righ?

suite/tests/client-interface/drbbdup-emul-reg-clobber-test.dll.c
```diff
@@ -0,0 +1,311 @@
...
+             */
+            CHECK(!TEST(DR_EMULATE_REST_OF_BLOCK, emul_info->flags) ||
+                      TEST(DR_EMULATE_INSTR_ONLY, emul_info->flags),
+                  "DR_EMULATE_REST_OF_BLOCK leaked through!");
+        }
+    }
+
+#ifdef X86
+    /* Cause drutil's expanded "rep movs" to crash if drbbdup sets
+     * DRREG_IGNORE_CONTROL_FLOW (i#5398).  We do not spill rcx until after the
```
derekbruening, line 188: This is testing something else and can also be removed, right?

suite/tests/client-interface/drbbdup-emul-reg-clobber-test.dll.c
```diff
@@ -0,0 +1,311 @@
...
+        drreg_set_vector_entry(&rvec, DR_REG_XCX, true);
+        if (drreg_reserve_register(drcontext, bb, where, &rvec, &reg_clobber) !=
+            DRREG_SUCCESS)
+            CHECK(false, "failed to reserve scratch register\n");
+        drvector_delete(&rvec);
+        instrlist_meta_preinsert(bb, where,
+                                 XINST_CREATE_load_int(drcontext,
+                                                       opnd_create_reg(reg_clobber),
+                                                       OPND_CREATE_INT32(-1)));
+        if (drreg_reserve_aflags(drcontext, bb, instr) != DRREG_SUCCESS)
```
derekbruening, line 206: I had to manually compare to the other file to find what was different: please put a comment here as this is a key part of this test and a future reader will have a hard time figuring that out, as I did.  Is this in lieu of the 3rd bbdup case?  Does this depend on the xcx clobber or can we remove that?  In any case the main thing is to add a comment explaining what this is testing.

suite/tests/client-interface/drbbdup-emul-reg-clobber-test.dll.c
```diff
@@ -0,0 +1,311 @@
...
+            CHECK(false, "cannot reserve aflags");
+        write_aflags(drcontext, bb, instr);
+        if (drreg_unreserve_aflags(drcontext, bb, instr) != DRREG_SUCCESS)
+            CHECK(false, "cannot unreserve aflags");
+        if (drreg_unreserve_register(drcontext, bb, where, reg_clobber) != DRREG_SUCCESS)
+            CHECK(false, "failed to unreserve scratch register\n");
+    }
+#endif
+
+    /* Ensure the drmgr emulation API works for a final special instr where "instr" !=
```
derekbruening, line 216: This again can all be removed (the entire rest of this function, right?) instead of being a redundant copy from the other test which makes this test harder to understand.

derekbruening (commented):
Thank you for adding a test.  Mostly I'm asking to pare down the test to only contain the new issue being tested as right now it still has exact copies of everything the other test is checking.

prasun3:
I am not sure why the OSX x86-64 build is failing
```
2023-05-05T05:21:57.5174020Z /Users/runner/work/dynamorio/dynamorio/build_debug-internal-64/suite/tests/drbbdup-emul-reg-clobber-test.c_asm.asm:84: error: invalid combination of opcode and operands
```

The Window 32-bit and 64-bit might be unrelated to the change.

This might be https://github.com/DynamoRIO/dynamorio/issues/3213

```
====> FAILURE in debug-internal-32 <====
code_api|win32.setthreadcontext 
```

```
====> FAILURE in debug-internal-64 <====
code_api|win32.threadchurn 
```

ext/drbbdup/drbbdup.c
```diff
@@ -130,6 +130,8 @@ typedef struct {
     hashtable_t manager_table; /* Maps bbs with book-keeping data (for thread-private
                                   caches only). */
     int case_index; /* Used to keep track of the current case during insertion. */
+    bool inserted_restore_all; /* track if we need to restore regs at the end of the block
```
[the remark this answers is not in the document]
prasun3: Done.

suite/tests/client-interface/drbbdup-emul-reg-clobber-test.c
```diff
@@ -0,0 +1,95 @@
...
+
+int
+main(int argc, char **argv)
+{
+#ifdef X86
+    char buf1[1024];
+    char buf2[1024];
+#    ifdef WINDOWS
+#        ifdef X64
+    // TODO: Implement on Windows.
```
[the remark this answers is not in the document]
prasun3: Removed these changes and replaced the test with ASM_CODE_ONLY.

suite/tests/client-interface/drbbdup-emul-reg-clobber-test.dll.c
prasun3: Done.

suite/tests/client-interface/drbbdup-emul-reg-clobber-test.dll.c
prasun3: Removed these lines.

suite/tests/client-interface/drbbdup-emul-reg-clobber-test.dll.c
prasun3: Deleted.

suite/tests/client-interface/drbbdup-emul-reg-clobber-test.dll.c
prasun3: I believe that there are three cases here: 0 (default), 1, and 2.

Also, the problem occurs when the instrumentation modifies flags in a non-default drbbdup case. The flag reservation by drbbdup does not cause the problem.

suite/tests/client-interface/drbbdup-emul-reg-clobber-test.dll.c
prasun3: Removed.

suite/tests/client-interface/drbbdup-emul-reg-clobber-test.dll.c
prasun3: Removed.

suite/tests/client-interface/drbbdup-emul-reg-clobber-test.dll.c
prasun3: Removed.

suite/tests/client-interface/drbbdup-emul-reg-clobber-test.dll.c
prasun3: Removed all lines that were not needed and added a comment.

suite/tests/client-interface/drbbdup-emul-reg-clobber-test.dll.c
prasun3: Removed.

suite/tests/CMakeLists.txt
```diff
@@ -2714,6 +2714,12 @@ if (NOT RISCV64) # TODO i#3544: Port tests to RISC-V 64
   use_DynamoRIO_extension(client.drbbdup-emul-test.dll drutil)
   use_DynamoRIO_extension(client.drbbdup-emul-test.dll drbbdup)
   use_DynamoRIO_extension(client.drbbdup-emul-test.dll drreg)
+
+  tobuild_ci(client.drbbdup-emul-reg-clobber-test client-interface/drbbdup-emul-reg-clobber-test.c "" "" "")
```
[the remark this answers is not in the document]
prasun3: Done.

suite/tests/client-interface/drbbdup-emul-reg-clobber-test.c
prasun3: Done. The OSX build is failing - I am not sure why. I tried explicitly setting the access size but that didn't change things. Probably some syntax issue but I don't have access to an OSX machine so I am unable to fix this.

prasun3:
Seeing `code_api|client.loader ` failure on vs2019-32 which seems unrelated.

```
====> FAILURE in debug-internal-32 <====
debug-internal-32: 248 tests passed, **** 7 tests failed, but ignoring 6 for i#2145: ****
	(ignore: i#2145) 	code_api|win32.hookerfirst 
	(ignore: i#2145) 	code_api|client.attach_test 
	code_api|client.loader 
	(ignore: i#2145) 	code_api|client.drx-test 
	(ignore: i#2145) 	code_api|client.drsyms-test 
	(ignore: i#2145) 	code_api|client.pcache-use 
	(ignore: i#2145) 	code_api|client.drwrap-test-detach 
```

suite/tests/client-interface/drbbdup-emul-reg-clobber-test.c
```diff
@@ -30,66 +30,64 @@
...
 
-#        endif
-#    endif
+        /* Rep mov that is expanded by the client. It should not clobber the
+         * value in XAX. */
+        mov      REG_XCX, 10
+        rep movsb [REG_XDI], [REG_XSI]
+
+        /* Return the value in the output parameter. */
+#ifdef X64 /* Should ideally use PTRSZ instead of #ifdef but seeing compiler error */
```
derekbruening, line 81: nit: Missing end .

Also: an error on PTRSZ is strange...

derekbruening (approved):
Thank you for updating the test.  LGTM.

derekbruening:
> I am not sure why the OSX x86-64 build is failing
> 
> ```
> 2023-05-05T05:21:57.5174020Z /Users/runner/work/dynamorio/dynamorio/build_debug-internal-64/suite/tests/drbbdup-emul-reg-clobber-test.c_asm.asm:84: error: invalid combination of opcode and operands
> ```

Remove the operands to the rep movsb?  Normally you would not specify them.

prasun3:
x86-32 failure is #5329

suite/tests/client-interface/drbbdup-emul-reg-clobber-test.c
prasun3: I went back to PTRSZ and it worked fine. Maybe I had a typo earlier.

prasun3:
> Remove the operands to the rep movsb? Normally you would not specify them.

This seems to have worked.

PR #5914, opened 2023-03-15
Updates debug printing macro to include if().

brettcoon:
The old DEBUG_VERSOSE(level) macro returns a condition that can be evaluated at runtime for debug actions.  When not in DEBUG mode, the condition always evaluates to false, resulting in unreachable code the compiler will hopefully eliminate, but may cause a warning.

The new IF_DEBUG_VERBOSE(level, action) macro includes the condition evaluation to perform <action> when the condition is met in DEBUG mode.  When not in DEBUG mode, all code including the condition is eliminated, leaving nothing to the compiler.

### DIFF
--- a/clients/drcachesim/tools/reuse_distance.cpp
+++ b/clients/drcachesim/tools/reuse_distance.cpp
@@ -52,11 +52,12 @@ reuse_distance_t::reuse_distance_t(const reuse_distance_knobs_t &knobs)
     : knobs_(knobs)
     , line_size_bits_(compute_log2((int)knobs_.line_size))
 {
-    if (DEBUG_VERBOSE(2)) {
-        std::cerr << "cache line size " << knobs_.line_size << ", "
-                  << "reuse distance threshold " << knobs_.distance_threshold
-                  << ", distance limit " << knobs_.distance_limit << std::endl;
-    }
+    reuse_distance_t::knob_verbose = knobs.verbose;
+    IF_DEBUG_VERBOSE(2,
+                     std::cerr << "cache line size " << knobs_.line_size << ", "
+                               << "reuse distance threshold " << knobs_.distance_threshold
+                               << ", distance limit " << knobs_.distance_limit
+                               << std::endl);
 }
 
 reuse_distance_t::~reuse_distance_t()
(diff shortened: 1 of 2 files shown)


clients/drcachesim/tools/reuse_distance.cpp
```diff
@@ -52,11 +52,12 @@ reuse_distance_t::reuse_distance_t(const reuse_distance_knobs_t &knobs)
...
-        std::cerr << "cache line size " << knobs_.line_size << ", "
-                  << "reuse distance threshold " << knobs_.distance_threshold
-                  << ", distance limit " << knobs_.distance_limit << std::endl;
-    }
+    reuse_distance_t::knob_verbose = knobs.verbose;
+    IF_DEBUG_VERBOSE(2,
+                     std::cerr << "cache line size " << knobs_.line_size << ", "
+                               << "reuse distance threshold " << knobs_.distance_threshold
+                               << ", distance limit " << knobs_.distance_limit
+                               << std::endl);
```
derekbruening, line 60: I would assume we don't really need to flush: this is std:cerr which is not doing buffering anyway.  I think all instances of `std::endl` in this file can be replaced with `\n` which will avoid needless calls to flush().

clients/drcachesim/tools/reuse_distance.h
```diff
@@ -369,20 +372,21 @@ struct line_ref_list_t {
...
-            for (prev = head_; prev != ref; prev = prev->next)
-                ++brute_dist;
-            if (brute_dist != dist) {
-                std::cerr << "Mismatch!  Brute=" << std::dec << brute_dist
-                          << " vs skip=" << dist << "\n";
-                print_list();
-                assert(false);
-            }
-        }
+        IF_DEBUG_VERBOSE(
```
derekbruening, line 375: Hmm, so this is the same as #ifdef DEBUG?

clients/drcachesim/tools/reuse_distance.h
```diff
@@ -49,13 +49,18 @@
...
-// compiler can remove for better performance without going so far as ifdef-ing
-// big code chunks and impairing readability.
+// checks knob_verbose, so for non-debug uses we eliminate it entirely.
+// Example usage:
+//   IF_DEBUG_VERBOSE(1, std::cerr << "This code was executed.\n")
 #ifdef DEBUG
-#    define DEBUG_VERBOSE(level) (reuse_distance_t::knob_verbose >= (level))
+#    define IF_DEBUG_VERBOSE(level, action)                  \
+        do {                                                 \
+            if (reuse_distance_t::knob_verbose >= (level)) { \
```
derekbruening, line 58: This global seems unnecessary: could pass in `this` instead and then better support multiple instances at once.  Could just add a XXX comment about that for future cleanup.

derekbruening (approved):

clients/drcachesim/tools/reuse_distance.cpp
brettcoon: Done.

clients/drcachesim/tools/reuse_distance.h
brettcoon: Yes, but with the option of making it depend on knob_verbose...

PR #5915, opened 2023-03-16
Add input stream queries

derekbruening:
Adds get_input_stream_count() and get_input_stream_name() to the scheduler_t drmemtrace interface.

Adds a test of these to the scheduler unit tests which uses real files and also serves as a test of only_threads for real files, whose code paths are different enough it had a bug which we fix here as well.

Issue: #5843

### DIFF
--- a/clients/drcachesim/tests/scheduler_unit_tests.cpp
+++ b/clients/drcachesim/tests/scheduler_unit_tests.cpp
@@ -367,46 +367,102 @@ test_regions()
 }
 
 static void
-test_only_threads()
+test_only_threads(const char *testdir)
 {
-    static constexpr memref_tid_t TID_A = 42;
-    static constexpr memref_tid_t TID_B = 99;
-    static constexpr memref_tid_t TID_C = 7;
-    std::vector<memref_t> refs_A = {
-        make_instr(TID_A, 50),
-        make_exit(TID_A),
-    };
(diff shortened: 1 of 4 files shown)


clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -367,58 +367,120 @@ test_regions()
...
+             status != scheduler_t::STATUS_EOF; status = stream->next_record(memref)) {
+            assert(status == scheduler_t::STATUS_OK);
+            assert(memref.instr.tid == TID_B);
+        }
+    }
+    // Test with real files as that is a separate code path in the scheduler.
+    // Since 32-bit memref_t is a different size we limit these to 64-bit builds.
+#if (defined(X86_64) || defined(ARM_64)) && defined(HAS_ZLIB) && defined(HAS_SNAPPY)
+
+    {
```
brettcoon, line 417: Just a nit, but why combine these rather than splitting this into two tests (synthetic streams and real files)?

brettcoon (approved):

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: At first they were both going to test only_threads, but now the real-file test includes other things.  OK, split it into a separate test.

PR #5918, opened 2023-03-20
fix a typo: patten -> pattern

lexi-nadia:


core/ir/aarch64/codec.py
```diff
@@ -209,7 +209,7 @@ def indent_append(text):
                         indent_append('    ASSERT(0);')
                 enc_key = fallthrough_instr_id(
                     pattern.opcode, pattern.opcode_bits, pattern.opnd_bits)
-                if enc_key in FALLTHROUGH and patten.opndset == FALLTHROUGH[enc_key].opndset:
+                if enc_key in FALLTHROUGH and pattern.opndset == FALLTHROUGH[enc_key].opndset:
```
derekbruening, line 212: @joshua-warburton (original author), @AssadHashmi -- there must be no tests that trigger this path?  Would it be straightforward to add such a test?

derekbruening (approved):

derekbruening:
run arm tests

derekbruening:
Moving to a top-level comment so we can merge by resolving the conversation:

@joshua-warburton (original author), @AssadHashmi -- there must be no tests that trigger this path? Would it be straightforward to add such a test?

core/ir/aarch64/codec.py
derekbruening: Moved to top-level.

PR #5923, opened 2023-03-22
i#2626: Make dis-a64*.txt test order stable

jackgallagher-arm:
Tests were previously sorted by mnemonic only, so instructions with multiple variants could change order when the the file is added to and re-sorted.
The order should now be stable which should cut back on code review noise.

This also reorders the test files to match the new order used by the script.

Issue: #2626


jackgallagher-arm:
run arm tests

AssadHashmi (approved):

PR #5925, opened 2023-03-23
i#5843 scheduler: Improve region skipping support

derekbruening:
Fixes some fencepost errors in scheduler input region of interest handling.
    
Adds a test of regions of interest which actually contains timestamps,
which is what revealed the errors.

Refactors the scheduler unit tests to use trace_entry_t instead of
memref_t, which is required to properly test the scheduler's input
readers, as that is the record type they operate on.  This results in
no longer needing reader_t::use_prev() which is removed here.

Issue: #5843


clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -698,8 +698,10 @@ scheduler_tmpl_t<RecordType, ReaderType>::advance_region_of_interest(int output_
 {
     uint64_t cur_instr = input.reader->get_instruction_ordinal();
     auto &cur_range = input.regions_of_interest[input.cur_region];
```
abhinav92003, line 700: Assert that input.cur_region is not out of bounds.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -698,8 +698,10 @@ scheduler_tmpl_t<RecordType, ReaderType>::advance_region_of_interest(int output_
...
-    if (cur_range.stop_instruction != 0 && cur_instr > cur_range.stop_instruction) {
+    if (input.in_cur_region && cur_range.stop_instruction != 0 &&
```
abhinav92003, line 701: So this is the case when we just exited an roi? Comment please.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -715,45 +717,32 @@ scheduler_tmpl_t<RecordType, ReaderType>::advance_region_of_interest(int output_
         cur_range = input.regions_of_interest[input.cur_region];
     }
 
-    if (cur_instr >= cur_range.start_instruction)
+    // If we're within one and already skipped, just exit to avoid re-requesting a skip
+    // and making no progress (we're on the inserted timetamp + cpuid and our cur instr
+    // count isn't yet the target).
+    if (input.in_cur_region && cur_instr >= cur_range.start_instruction - 1)
```
abhinav92003, line 723: If the first roi starts at instr 0, then in _cur_region may be false (if it is initialized to false) but we're already in the current region. In that case, should we be returning here?

Also see comment above about overlapping rois.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -698,8 +698,10 @@ scheduler_tmpl_t<RecordType, ReaderType>::advance_region_of_interest(int output_
 {
     uint64_t cur_instr = input.reader->get_instruction_ordinal();
     auto &cur_range = input.regions_of_interest[input.cur_region];
-    if (cur_range.stop_instruction != 0 && cur_instr > cur_range.stop_instruction) {
+    if (input.in_cur_region && cur_range.stop_instruction != 0 &&
+        cur_instr > cur_range.stop_instruction) {
         ++input.cur_region;
+        input.in_cur_region = false;
```
abhinav92003, line 704: Are ROIs mutually exclusive? Is it possible that the next ROI overlaps with the cur one, so that when we exit the cur one, we're in the next ROI and this should still be true?

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -715,45 +717,32 @@ scheduler_tmpl_t<RecordType, ReaderType>::advance_region_of_interest(int output_
...
-    if (cur_instr >= cur_range.start_instruction)
+    // If we're within one and already skipped, just exit to avoid re-requesting a skip
+    // and making no progress (we're on the inserted timetamp + cpuid and our cur instr
+    // count isn't yet the target).
+    if (input.in_cur_region && cur_instr >= cur_range.start_instruction - 1)
         return sched_type_t::STATUS_OK;
 
     VPRINT(this, 2, "skipping from %" PRId64 " to %" PRId64 " instrs for ROI\n",
            cur_instr, cur_range.start_instruction);
+    input.in_cur_region = true;
```
abhinav92003, line 728: nit: If it doesn't break any logic: it'd be better to set this bool only after we actually skip the instrs at L734.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -715,45 +717,32 @@ scheduler_tmpl_t<RecordType, ReaderType>::advance_region_of_interest(int output_
...
     // We assume the queue contains no instrs (else our query of input.reader's
     // instr ordinal would include them and so be incorrect) and that we should
     // thus skip it all.
```
abhinav92003, line 731: Add assert for this ("queue contains no instrs").

abhinav92003 (approved):

clients/drcachesim/reader/reader.h
```diff
@@ -190,18 +190,14 @@ class reader_t : public std::iterator<std::input_iterator_tag, memref_t>,
...
-    // Separated into a virtual method for overriding in test mock readers.
-    virtual void
-    use_prev(trace_entry_t *prev);
     // This updates internal state for the just-read input_entry_.
     // Returns whether a new memref record is now available.
     virtual bool
     process_input_entry();
 
     // Meant to be called from skip_instructions();
     // Looks for headers prior to a skip in case it is from the start of the trace.
```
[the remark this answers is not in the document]
derekbruening: Done.

clients/drcachesim/analyzer.cpp
```diff
@@ -144,8 +144,13 @@ analyzer_tmpl_t<RecordType, ReaderType>::init_scheduler(const std::string &trace
...
     }
     std::vector<typename sched_type_t::range_t> regions;
-    if (skip_instrs_ > 0)
-        regions.emplace_back(skip_instrs_, 0);
+    if (skip_instrs_ > 0) {
+        // TODO i#5843: For serial mode with multiple inputs this is not doing the
+        // right thing: this is skipping in every input stream, while the documented
+        // behavior is supposed to be an output stream skip.  Once we have that
+        // capability in the scheduler we should switch to that.
+        regions.emplace_back(skip_instrs_ + 1, 0);
```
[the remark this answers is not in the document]
derekbruening: The analyzer is the one setting all the scheduler input options so it knows there are no other regions of interest: only this one that it is creating here.

clients/drcachesim/scheduler/scheduler.h
```diff
@@ -712,6 +710,7 @@ template <typename RecordType, typename ReaderType> class scheduler_tmpl_t {
         int priority = 0;
         std::vector<range_t> regions_of_interest;
         int cur_region = 0;
+        bool in_cur_region = false;
```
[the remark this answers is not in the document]
derekbruening: Done.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Done.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Done.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -715,45 +717,32 @@ scheduler_tmpl_t<RecordType, ReaderType>::advance_region_of_interest(int output_
         cur_range = input.regions_of_interest[input.cur_region];
```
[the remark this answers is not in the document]
derekbruening: Not each, after the last one, which may not include the actual thread exit record in the trace, so we have to synthesize it.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Done.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Done.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: No, they can't be.  Added docs saying this and added an up-front check and test cases to the invalid param test.

PR #5928, opened 2023-03-24
i#5843 scheduler: Add simple interleaving

derekbruening:
Adds initial support for MAP_TO_ANY_OUTPUT with multiple outputs. Uses a simple queue of ready-to-schedule inputs and implements an instruction-based scheduling quantum.

Adds a test.

Issue: #5843

### DIFF
--- a/clients/drcachesim/scheduler/scheduler.cpp
+++ b/clients/drcachesim/scheduler/scheduler.cpp
@@ -502,7 +502,7 @@ scheduler_tmpl_t<RecordType, ReaderType>::init(
         for (int i = 0; i < static_cast<int>(inputs_.size()); i++) {
             size_t index = i % output_count;
             if (outputs_[index].input_indices.empty())
-                outputs_[index].cur_input = i;
+                set_cur_input(index, i);
             outputs_[index].input_indices.push_back(i);
             VPRINT(this, 2, "Assigning input #%d to output #%zd\n", i, index);
         }
@@ -515,7 +515,7 @@ scheduler_tmpl_t<RecordType, ReaderType>::init(
         if (output_count != 1)
             return STATUS_ERROR_NOT_IMPLEMENTED;
         if (inputs_.size() == 1) {
-            outputs_[0].cur_input = 0;
+            set_cur_input(0, 0);
         } else {
             // The old file_reader_t interleaving would output the top headers for every
             // thread first and then pick the oldest timestamp once it reached a
@@ -526,11 +526,25 @@ scheduler_tmpl_t<RecordType, ReaderType>::init(
                 return res;
         }
     } else {
-        // TODO i#5843: Implement scheduling onto synthetic cores.
-        // For now we only support a single output stream with no deps.
-        if (options_.deps != DEPENDENCY_IGNORE || output_count != 1)
+        // TODO i#5843: Honor deps when scheduling on synthetic cores.
+        if (options_.deps != DEPENDENCY_IGNORE)
             return STATUS_ERROR_NOT_IMPLEMENTED;
-        outputs_[0].cur_input = 0;
+        // TODO i#5843: Implement time-based quanta.
+        if (options_.quantum_unit != QUANTUM_INSTRUCTIONS)
+            return STATUS_ERROR_NOT_IMPLEMENTED;
+        // Assign initial inputs.
+        // TODO i#5843: Once we support core bindings and priorities we'll want
+        // to consider that here.
+        for (int i = 0; i < static_cast<int>(outputs_.size()); i++) {
+            if (i < static_cast<int>(inputs_.size()))
+                set_cur_input(i, i);
+            else
+                set_cur_input(i, -1);
+        }
+        for (int i = static_cast<int>(outputs_.size());
+             i < static_cast<int>(inputs_.size()); i++) {
+            ready_.push(i);
+        }
     }
     return STATUS_SUCCESS;
 }
@@ -568,8 +582,8 @@ scheduler_tmpl_t<RecordType, ReaderType>::get_initial_timestamps()
             return STATUS_ERROR_INVALID_PARAMETER;
         if (input.next_timestamp < min_time) {
             min_time = input.next_timestamp;
-            // TODO i#5843: Support more than one input (already checked earlier).
-            outputs_[0].cur_input = static_cast<int>(i);
+            // TODO i#5843: Support more than one output (already checked earlier).
+            set_cur_input(0, static_cast<int>(i));
         }
     }
     if (outputs_[0].cur_input >= 0)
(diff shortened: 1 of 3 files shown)


clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -775,20 +789,46 @@ scheduler_tmpl_t<RecordType, ReaderType>::advance_region_of_interest(int output_
     return sched_type_t::STATUS_SKIPPED;
 }
 
+template <typename RecordType, typename ReaderType>
+void
+scheduler_tmpl_t<RecordType, ReaderType>::set_cur_input(int output_ordinal,
+                                                        int input_ordinal)
+{
+    assert(output_ordinal >= 0 && output_ordinal < static_cast<int>(outputs_.size()));
+    assert(input_ordinal < static_cast<int>(inputs_.size()));
```
abhinav92003, line 798: Comment that input_ordinal may be kInvalidInput.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -526,11 +526,25 @@ scheduler_tmpl_t<RecordType, ReaderType>::init(
...
+        if (options_.quantum_unit != QUANTUM_INSTRUCTIONS)
+            return STATUS_ERROR_NOT_IMPLEMENTED;
+        // Assign initial inputs.
+        // TODO i#5843: Once we support core bindings and priorities we'll want
+        // to consider that here.
+        for (int i = 0; i < static_cast<int>(outputs_.size()); i++) {
+            if (i < static_cast<int>(inputs_.size()))
+                set_cur_input(i, i);
+            else
+                set_cur_input(i, -1);
```
abhinav92003, line 542: Use symbolic constant kInvalidInput for -1.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -775,20 +789,46 @@ scheduler_tmpl_t<RecordType, ReaderType>::advance_region_of_interest(int output_
...
 {
-    if (options_.mapping == MAP_TO_ANY_OUTPUT) {
-        // TODO i#5843: Implement synthetic scheduling with instr/time quanta.
-        // These will require locks and central coordination.
-        // For now we only support a single output.
-        return sched_type_t::STATUS_EOF;
-    }
-    int index = outputs_[output_ordinal].cur_input;
+    if (options_.mapping == MAP_TO_ANY_OUTPUT)
+        std::lock_guard<std::mutex> guard(sched_lock_);
```
abhinav92003, line 812: The scope of this lock_guard contains only this declaration. Is it useful?

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -889,11 +929,20 @@ scheduler_tmpl_t<RecordType, ReaderType>::next_record(int output_ordinal,
...
+        bool need_new_input = false;
+        if (options_.mapping == MAP_TO_ANY_OUTPUT &&
+            options_.quantum_unit == QUANTUM_INSTRUCTIONS &&
+            record_type_is_instr(record)) {
+            // TODO i#5843: We also want to swap on blocking syscalls.
+            ++input->instrs_in_quantum;
+            if (input->instrs_in_quantum > options_.quantum_duration)
+                need_new_input = true;
+        }
         if (options_.deps == DEPENDENCY_TIMESTAMPS &&
```
abhinav92003, line 941: This switches input on each timestamp marker? Even if the thread continues on the same core in the trace?

clients/drcachesim/scheduler/scheduler.h
```diff
@@ -830,6 +836,12 @@ template <typename RecordType, typename ReaderType> class scheduler_tmpl_t {
     scheduler_options_t options_;
     std::vector<input_info_t> inputs_;
     std::vector<output_info_t> outputs_;
+    // We use a central lock for global scheduling.  We assume the synchronization
+    // cost is outweighed by the simulator's overhead.  This protects concurrent
+    // access to the inputs_, outputs_, and ready_ fields.
+    std::mutex sched_lock_;
+    // Input indices ready to be scheduled.
+    std::queue<int> ready_;
```
abhinav92003, line 844: We should typedef a type for input ordinal. It'll make APIs and declarations more readable.

clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -82,7 +82,7 @@ class mock_reader_t : public reader_t {
 };
 
 static trace_entry_t
-make_instr(addr_t pc)
+make_instr(addr_t pc, memref_tid_t tid = 1)
```
abhinav92003, line 85: tid unused?  Probably left over from when we used memref_t?

clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -682,6 +682,70 @@ test_real_file_queries_and_filters(const char *testdir)
...
+            sched_as_string[i] += 'A' + (memref.instr.tid - TID_BASE);
+        }
+    }
+    for (int i = 0; i < NUM_OUTPUTS; i++) {
+        std::cerr << "cpu #" << i << " schedule: " << sched_as_string[i] << "\n";
+    }
+    // Hardcoding here for the 2 outputs and 7 inputs.
+    // We expect 3 letter sequences (our quantum) alternating every-other as each
+    // core alternates; with an odd number the 2nd core finishes early.
+    assert(sched_as_string[0] == "AAACCCEEEGGGBBBDDDFFFAAAACCCCEEEEGGGG");
```
abhinav92003, line 745: The later part seems to have 4 letter sequences (AAAACCCCEEEEGGGG). Doesn't look expected. Update comment above if I'm missing something.

abhinav92003 (approved):

clients/drcachesim/scheduler/scheduler.h
derekbruening: Created input_ordinal_t and output_ordinal_t.  This changes the signatures of some previously exported routines but the interfaces is in flux in any case.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Added INVALID_INPUT_ORDINAL and INVALID_OUTPUT_ORDINAL of the new types.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Done.

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: Yes, removed.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: I believe you're thinking of MAP_TO_RECORDED_OUTPUT which is not implemented for other than a single output (for analyzer serial mode): re-creating the recorded schedule on >1 output streams is not yet implemented.

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: Ah good catch: the test needs to only append a letter on an instruction record not other records.  Then it's all 3's.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: No this is not what we want.  Replaced with `std::unique_lock` for a conditional lock that still auto-unlocks on any return path:
```C
    bool need_lock = (options_.mapping == MAP_TO_ANY_OUTPUT);
    auto scoped_lock = need_lock ? std::unique_lock<std::mutex>()
                                 : std::unique_lock<std::mutex>(sched_lock_);
```

derekbruening:
run arm tests

derekbruening:
Hit what looks like a flaky output ordering test on drcacheoff.fork on jenkins: but it's not filed anywhere so re-running and filed #5933.

derekbruening:
x86-32 failure is scattergather pipe assert #5329

Issue #5934, opened 2023-03-29
Create a dedicated invariant check for repeated syscall instructions with the same PCs


bete0:
This is to create a new invariant check type for double syscall instructions that have the same PC address and are currently reported as **_Non-explicit control flow has no marker_** invariant error. This will help in identify this specific invariant error in traces. 
```
  8979242: T141294 ifetch       2 byte(s) @ 0x00007fcf3b9dd9ea 0f 05                syscall  -> %rcx %r11
  8979243: T141294 <marker: timestamp 13311036165328271>
  8979244: T141294 <marker: tid 141294 on core 4194>
  8979245: T141294 ifetch       2 byte(s) @ 0x00007fcf3b9dd9ea 0f 05                syscall  -> %rcx %r11
  8979246: T141294 <marker: timestamp 13311036165328668>
```

derekbruening:
Let's make this cover root-causing and fixing this issue as well and not just adding the check.

derekbruening:
The PR #5945 raw2trace workaround goes too far as it erroneously removes legitimate back-to-back syscalls: #6102

abhinav92003:
Is there anything that sets apart the legitimate back-to-back syscall cases? We'd need to make changes to the invariant checker too accordingly.

derekbruening:
Legitimate cases have to be auto-restart; once we have the number markers we can check that.

PR #5935, opened 2023-03-29
i#5934 Add 'Duplicate syscalls with the same PC' invariant check

bete0:
Adds a dedicated invariant check for duplicate syscalls that have the same PC and reported as 'Non-explicit control flow has no marker'.

Tested with an app that has the 'Non-explicit control flow has no marker' error for duplicate syscalls and after this change, most of the errors classified in that original error type are now reported as 'Duplicate syscall instrs with the same PC'.

Issue: #5934

### DIFF
--- a/clients/drcachesim/tests/invariant_checker_test.cpp
+++ b/clients/drcachesim/tests/invariant_checker_test.cpp
@@ -464,52 +464,6 @@ check_function_markers()
 bool
 check_repeated_syscall_with_same_pc()
 {
-    // Negative: syscalls with the same PC.
-#if defined(X86_64) || defined(X86_32) || defined(ARM_64)
-    {
-        std::vector<memref_t> memrefs = {
-            gen_instr(1, 1),
-#    if defined(X86_64) || defined(X86_32)
-            gen_syscall_encoded(1, 2, { 0x0f, 0x05 }),
-            gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
-            gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
-            gen_syscall_encoded(1, 2, { 0x0f, 0x05 }),
-#    elif defined(ARM_64)
-            gen_syscall_encoded(1, 2, 0xd4000001),
-            gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
-            gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
-            gen_syscall_encoded(1, 2, 0xd4000001),
-#    else
-        // TODO i#5871: Add AArch32 (and RISC-V) encodings.
-#    endif
-        };
-        if (!run_checker(memrefs, true, 1, 5, "Repeated syscall instrs with the same PC",
-                         "Failed to catch repeated syscall instrs with the same PC"))
-            return false;
-    }
-    // Positive test: syscalls with different PCs.
-    {
-        std::vector<memref_t> memrefs = {
-            gen_instr(1, 1),
-#    if defined(X86_64) || defined(X86_32)
-            gen_syscall_encoded(1, 2, { 0x0f, 0x05 }),
-            gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
-            gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
-            gen_syscall_encoded(1, 4, { 0x0f, 0x05 }),
-#    elif defined(ARM_64)
-            gen_syscall_encoded(1, 2, 0xd4000001),
-            gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
-            gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
-            gen_syscall_encoded(1, 6, 0xd4000001),
-#    else
-        // TODO i#5871: Add AArch32 (and RISC-V) encodings.
-#    endif
-        };
-        if (!run_checker(memrefs, false)) {
-            return false;
-        }
-    }
-#endif
     return true;
 }
 
--- a/clients/drcachesim/tests/invariant_checker_test.cpp
+++ b/clients/drcachesim/tests/invariant_checker_test.cpp
@@ -464,39 +464,52 @@ check_function_markers()
 bool
 check_repeated_syscall_with_same_pc()
 {
-      // Negative: syscalls with the same PC.
+    // Negative: syscalls with the same PC.
+#if defined(X86_64) || defined(X86_32) || defined(ARM_64)
     {
         std::vector<memref_t> memrefs = {
             gen_instr(1, 1),
 #    if defined(X86_64) || defined(X86_32)
-            gen_syscall_encoded(1, 2, { 0x0f, 0x05 } ),
-            gen_syscall_encoded(1, 2, { 0x0f, 0x05 } ),
+            gen_syscall_encoded(1, 2, { 0x0f, 0x05 }),
+            gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
+            gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
+            gen_syscall_encoded(1, 2, { 0x0f, 0x05 }),
 #    elif defined(ARM_64)
-            gen_branch_encoded(1, 2, 0x14000e0e),
-            gen_branch_encoded(1, 2, 0x14000e0e),
-#    endif
+            gen_syscall_encoded(1, 2, 0xd4000001),
+            gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
+            gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
+            gen_syscall_encoded(1, 2, 0xd4000001),
+#    else
         // TODO i#5871: Add AArch32 (and RISC-V) encodings.
+#    endif
         };
-        if (!run_checker(memrefs, true, 1, 3, "Repeated syscall instrs with the same PC",
+        if (!run_checker(memrefs, true, 1, 5, "Repeated syscall instrs with the same PC",
                          "Failed to catch repeated syscall instrs with the same PC"))
             return false;
     }
     // Positive test: syscalls with different PCs.
     {
         std::vector<memref_t> memrefs = {
+            gen_instr(1, 1),
 #    if defined(X86_64) || defined(X86_32)
             gen_syscall_encoded(1, 2, { 0x0f, 0x05 }),
+            gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
+            gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
             gen_syscall_encoded(1, 4, { 0x0f, 0x05 }),
 #    elif defined(ARM_64)
-            gen_branch_encoded(1, 2, 0x14000e0e),
-            gen_branch_encoded(1, 4, 0x14000e0e),
-#    endif
+            gen_syscall_encoded(1, 2, 0xd4000001),
+            gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
+            gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
+            gen_syscall_encoded(1, 6, 0xd4000001),
+#    else
         // TODO i#5871: Add AArch32 (and RISC-V) encodings.
+#    endif
         };
         if (!run_checker(memrefs, false)) {
             return false;
         }
     }
+#endif
     return true;
 }
 
--- a/clients/drcachesim/tests/invariant_checker_test.cpp
+++ b/clients/drcachesim/tests/invariant_checker_test.cpp
@@ -530,7 +530,8 @@ int
 main(int argc, const char *argv[])
 {
     if (check_branch_target_after_branch() && check_sane_control_flow() &&
-        check_kernel_xfer() && check_rseq() && check_function_markers()) {
+        check_kernel_xfer() && check_rseq() && check_function_markers() &&
+        check_repeated_syscall_with_same_pc()) {
         std::cerr << "invariant_checker_test passed\n";
         return 0;
     }
--- a/clients/drcachesim/tests/invariant_checker_test.cpp
+++ b/clients/drcachesim/tests/invariant_checker_test.cpp
@@ -464,6 +464,7 @@ check_function_markers()
 bool
 check_repeated_syscall_with_same_pc()
 {
+#ifdef UNIX
     // Negative: syscalls with the same PC.
 #if defined(X86_64) || defined(X86_32) || defined(ARM_64)
     {
@@ -509,6 +510,7 @@ check_repeated_syscall_with_same_pc()
             return false;
         }
     }
+#endif
 #endif
     return true;
 }
--- a/clients/drcachesim/tests/invariant_checker_test.cpp
+++ b/clients/drcachesim/tests/invariant_checker_test.cpp
@@ -466,6 +466,7 @@ check_repeated_syscall_with_same_pc()
 {
     constexpr addr_t ADDR_ONE = 0x7fcf3b9dd9e9;
     constexpr addr_t ADDR_TWO = 0x7fcf3b9dd9eb;
+    constexpr addr_t ADDR_THREE = 0x7fcf3b9dd9ed;
     // Negative: syscalls with the same PC.
 #if defined(X86_64) || defined(X86_32) || defined(ARM_64)
     {
@@ -477,17 +478,19 @@ check_repeated_syscall_with_same_pc()
             gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
             gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
             gen_instr_encoded(1, ADDR_ONE, { 0x0f, 0x05 }),
+            gen_instr_encoded(1 /*tid*/, ADDR_TWO /*pc*/, { 0x01 }),
 #    elif defined(ARM_64)
             gen_instr_encoded(1 /*tid*/, 0 /*pc*/, 0x01),
             gen_instr_encoded(1, ADDR_ONE, 0xd4000001),
             gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
             gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
             gen_instr_encoded(1, ADDR_ONE, 0xd4000001),
+            gen_instr_encoded(1 /*tid*/, ADDR_TWO /*pc*/, 0x01),
 #    else
         // TODO i#5871: Add AArch32 (and RISC-V) encodings.
 #    endif
         };
-        if (!run_checker(memrefs, true, 1, 6, "Repeated syscall instrs with the same PC",
+        if (!run_checker(memrefs, true, 1, 7, "Repeated syscall instrs with the same PC",
                          "Failed to catch repeated syscall instrs with the same PC"))
             return false;
     }
@@ -502,12 +505,14 @@ check_repeated_syscall_with_same_pc()
             gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
             gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
             gen_instr_encoded(1, ADDR_TWO, { 0x0f, 0x05 }),
+            gen_instr_encoded(1 /*tid*/, ADDR_THREE /*pc*/, { 0x01 }),
 #    elif defined(ARM_64)
             gen_instr_encoded(1 /*tid*/, 0 /*pc*/, 0x01),
             gen_instr_encoded(1, ADDR_ONE, 0xd4000001, 2),
             gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
             gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
             gen_instr_encoded(1, ADDR_TWO, 0xd4000001, 2),
+            gen_instr_encoded(1 /*tid*/, ADDR_THREE /*pc*/, 0x01),
 #    else
         // TODO i#5871: Add AArch32 (and RISC-V) encodings.
 #    endif
--- a/clients/drcachesim/tests/invariant_checker_test.cpp
+++ b/clients/drcachesim/tests/invariant_checker_test.cpp
@@ -465,32 +465,30 @@ bool
 check_repeated_syscall_with_same_pc()
 {
     constexpr addr_t ADDR_ONE = 0x7fcf3b9dd9e9;
-    constexpr addr_t ADDR_TWO = 0x7fcf3b9dd9eb;
-    constexpr addr_t ADDR_THREE = 0x7fcf3b9dd9ed;
     // Negative: syscalls with the same PC.
 #if defined(X86_64) || defined(X86_32) || defined(ARM_64)
     {
         std::vector<memref_t> memrefs = {
             gen_marker(1, TRACE_MARKER_TYPE_FILETYPE, OFFLINE_FILE_TYPE_ENCODINGS),
 #    if defined(X86_64) || defined(X86_32)
-            gen_instr_encoded(1 /*tid*/, 0 /*pc*/, { 0x01 }),
-            gen_instr_encoded(1, ADDR_ONE, { 0x0f, 0x05 }),
+            gen_instr_encoded(0, { 0x01 }),
+            gen_instr_encoded(ADDR_ONE, { 0x0f, 0x05 }), // 0x7fcf3b9dd9e9: 0f 05 syscall
             gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
             gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
-            gen_instr_encoded(1, ADDR_ONE, { 0x0f, 0x05 }),
-            gen_instr_encoded(1 /*tid*/, ADDR_TWO /*pc*/, { 0x01 }),
+            gen_instr_encoded(ADDR_ONE, { 0x0f, 0x05 }), // 0x7fcf3b9dd9e9: 0f 05 syscall
 #    elif defined(ARM_64)
-            gen_instr_encoded(1 /*tid*/, 0 /*pc*/, 0x01),
-            gen_instr_encoded(1, ADDR_ONE, 0xd4000001),
+            gen_instr_encoded(0, 0x01),
+            gen_instr_encoded(ADDR_ONE,
+                              0xd4000001), // 0x7fcf3b9dd9e9: 0xd4000001 svc #0x0
             gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
             gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
-            gen_instr_encoded(1, ADDR_ONE, 0xd4000001),
-            gen_instr_encoded(1 /*tid*/, ADDR_TWO /*pc*/, 0x01),
+            gen_instr_encoded(ADDR_ONE,
+                              0xd4000001), // 0x7fcf3b9dd9e9: 0xd4000001 svc #0x0
 #    else
         // TODO i#5871: Add AArch32 (and RISC-V) encodings.
 #    endif
         };
-        if (!run_checker(memrefs, true, 1, 7, "Repeated syscall instrs with the same PC",
+        if (!run_checker(memrefs, true, 1, 6, "Repeated syscall instrs with the same PC",
                          "Failed to catch repeated syscall instrs with the same PC"))
             return false;
     }
@@ -500,19 +498,20 @@ check_repeated_syscall_with_same_pc()
         std::vector<memref_t> memrefs = {
             gen_marker(1, TRACE_MARKER_TYPE_FILETYPE, OFFLINE_FILE_TYPE_ENCODINGS),
 #    if defined(X86_64) || defined(X86_32)
-            gen_instr_encoded(1 /*tid*/, 0 /*pc*/, { 0x01 }),
-            gen_instr_encoded(1, ADDR_ONE, { 0x0f, 0x05 }),
+            gen_instr_encoded(0, { 0x01 }),
+            gen_instr_encoded(ADDR_ONE, { 0x0f, 0x05 }), // 0x7fcf3b9dd9e9: 0f 05 syscall
             gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
             gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
-            gen_instr_encoded(1, ADDR_TWO, { 0x0f, 0x05 }),
-            gen_instr_encoded(1 /*tid*/, ADDR_THREE /*pc*/, { 0x01 }),
+            gen_instr_encoded(ADDR_ONE + 2,
+                              { 0x0f, 0x05 }), // 0x7fcf3b9dd9eb: 0f 05    syscall
 #    elif defined(ARM_64)
-            gen_instr_encoded(1 /*tid*/, 0 /*pc*/, 0x01),
-            gen_instr_encoded(1, ADDR_ONE, 0xd4000001, 2),
+            gen_instr_encoded(0, 0x01),
+            gen_instr_encoded(ADDR_ONE, 0xd4000001,
+                              2), // 0x7fcf3b9dd9e9: 0xd4000001 svc #0x0
             gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
             gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
-            gen_instr_encoded(1, ADDR_TWO, 0xd4000001, 2),
-            gen_instr_encoded(1 /*tid*/, ADDR_THREE /*pc*/, 0x01),
+            gen_instr_encoded(ADDR_ONE + 4, 0xd4000001,
+                              2), // 0x7fcf3b9dd9eb: 0xd4000001 svc #0x0
 #    else
         // TODO i#5871: Add AArch32 (and RISC-V) encodings.
 #    endif
@@ -531,8 +530,7 @@ int
 main(int argc, const char *argv[])
 {
     if (check_branch_target_after_branch() && check_sane_control_flow() &&
-        check_kernel_xfer() && check_rseq() && check_function_markers() &&
-        check_repeated_syscall_with_same_pc()) {
+        check_kernel_xfer() && check_rseq() && check_function_markers()) {
         std::cerr << "invariant_checker_test passed\n";
         return 0;
     }
--- a/clients/drcachesim/tests/invariant_checker_test.cpp
+++ b/clients/drcachesim/tests/invariant_checker_test.cpp
@@ -215,10 +215,8 @@ check_sane_control_flow()
             gen_marker(1, TRACE_MARKER_TYPE_FILETYPE, OFFLINE_FILE_TYPE_ENCODINGS),
 #    if defined(X86_64) || defined(X86_32)
             // 0x74 is "je" with the 2nd byte the offset.
-            gen_instr_encoded(0x71019dba, { 0x20 }),
             gen_branch_encoded(1, 0x71019dbc, { 0x74, 0x32 }),
 #    elif defined(ARM_64)
-            gen_instr_encoded(0x71019db9, 0x20),
             // 71019dbc:   540001a1        b.ne    71019df0 <__executable_start+0x19df0>
             gen_branch_encoded(1, 0x71019dbc, 0x540001a1),
 #    else
@@ -473,13 +471,11 @@ check_repeated_syscall_with_same_pc()
         std::vector<memref_t> memrefs = {
             gen_marker(1, TRACE_MARKER_TYPE_FILETYPE, OFFLINE_FILE_TYPE_ENCODINGS),
 #    if defined(X86_64) || defined(X86_32)
-            gen_instr_encoded(0, { 0x01 }),
             gen_instr_encoded(ADDR_ONE, { 0x0f, 0x05 }), // 0x7fcf3b9dd9e9: 0f 05 syscall
             gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
             gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
             gen_instr_encoded(ADDR_ONE, { 0x0f, 0x05 }), // 0x7fcf3b9dd9e9: 0f 05 syscall
 #    elif defined(ARM_64)
-            gen_instr_encoded(0, 0x01),
             gen_instr_encoded(ADDR_ONE,
                               0xd4000001), // 0x7fcf3b9dd9e9: 0xd4000001 svc #0x0
             gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
@@ -490,7 +486,7 @@ check_repeated_syscall_with_same_pc()
         // TODO i#5871: Add AArch32 (and RISC-V) encodings.
 #    endif
         };
-        if (!run_checker(memrefs, true, 1, 6, "Repeated syscall instrs with the same PC",
+        if (!run_checker(memrefs, true, 1, 5, "Repeated syscall instrs with the same PC",
                          "Failed to catch repeated syscall instrs with the same PC"))
             return false;
     }
@@ -500,16 +496,14 @@ check_repeated_syscall_with_same_pc()
         std::vector<memref_t> memrefs = {
             gen_marker(1, TRACE_MARKER_TYPE_FILETYPE, OFFLINE_FILE_TYPE_ENCODINGS),
 #    if defined(X86_64) || defined(X86_32)
-            gen_instr_encoded(0, { 0x01 }),
             gen_instr_encoded(ADDR_ONE, { 0x0f, 0x05 }), // 0x7fcf3b9dd9e9: 0f 05 syscall
             gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
             gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
             gen_instr_encoded(ADDR_ONE + 2,
                               { 0x0f, 0x05 }), // 0x7fcf3b9dd9eb: 0f 05    syscall
 #    elif defined(ARM_64)
-            gen_instr_encoded(0, 0x01),
             gen_instr_encoded(ADDR_ONE, 0xd4000001,
-                              2), // 0x7fcf3b9dd9e9: 0xd4000001 svc #0x0
+                              2),          // 0x7fcf3b9dd9e9: 0xd4000001 svc #0x0
             gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
             gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
             gen_instr_encoded(ADDR_ONE + 4, 0xd4000001,
--- a/clients/drcachesim/tests/invariant_checker_test.cpp
+++ b/clients/drcachesim/tests/invariant_checker_test.cpp
@@ -466,23 +466,23 @@ check_repeated_syscall_with_same_pc()
 {
 #ifdef UNIX
     // Negative: syscalls with the same PC.
-#if defined(X86_64) || defined(X86_32) || defined(ARM_64)
+#    if defined(X86_64) || defined(X86_32) || defined(ARM_64)
     {
         std::vector<memref_t> memrefs = {
             gen_instr(1, 1),
-#    if defined(X86_64) || defined(X86_32)
+#        if defined(X86_64) || defined(X86_32)
             gen_syscall_encoded(1, 2, { 0x0f, 0x05 }),
             gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
             gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
             gen_syscall_encoded(1, 2, { 0x0f, 0x05 }),
-#    elif defined(ARM_64)
+#        elif defined(ARM_64)
             gen_syscall_encoded(1, 2, 0xd4000001),
             gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
             gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
             gen_syscall_encoded(1, 2, 0xd4000001),
-#    else
+#        else
         // TODO i#5871: Add AArch32 (and RISC-V) encodings.
-#    endif
+#        endif
         };
         if (!run_checker(memrefs, true, 1, 5, "Repeated syscall instrs with the same PC",
                          "Failed to catch repeated syscall instrs with the same PC"))
@@ -492,25 +492,25 @@ check_repeated_syscall_with_same_pc()
     {
         std::vector<memref_t> memrefs = {
             gen_instr(1, 1),
-#    if defined(X86_64) || defined(X86_32)
+#        if defined(X86_64) || defined(X86_32)
             gen_syscall_encoded(1, 2, { 0x0f, 0x05 }),
             gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
             gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
             gen_syscall_encoded(1, 4, { 0x0f, 0x05 }),
-#    elif defined(ARM_64)
+#        elif defined(ARM_64)
             gen_syscall_encoded(1, 2, 0xd4000001),
             gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
             gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
             gen_syscall_encoded(1, 6, 0xd4000001),
-#    else
+#        else
         // TODO i#5871: Add AArch32 (and RISC-V) encodings.
-#    endif
+#        endif
         };
         if (!run_checker(memrefs, false)) {
             return false;
         }
     }
-#endif
+#    endif
 #endif
     return true;
 }
--- a/clients/drcachesim/tests/invariant_checker_test.cpp
+++ b/clients/drcachesim/tests/invariant_checker_test.cpp
@@ -464,25 +464,24 @@ check_function_markers()
 bool
 check_repeated_syscall_with_same_pc()
 {
-#ifdef UNIX
     // Negative: syscalls with the same PC.
-#    if defined(X86_64) || defined(X86_32) || defined(ARM_64)
+#if defined(X86_64) || defined(X86_32) || defined(ARM_64)
     {
         std::vector<memref_t> memrefs = {
             gen_instr(1, 1),
-#        if defined(X86_64) || defined(X86_32)
+#    if defined(X86_64) || defined(X86_32)
             gen_syscall_encoded(1, 2, { 0x0f, 0x05 }),
             gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
             gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
             gen_syscall_encoded(1, 2, { 0x0f, 0x05 }),
-#        elif defined(ARM_64)
+#    elif defined(ARM_64)
             gen_syscall_encoded(1, 2, 0xd4000001),
             gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
             gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
             gen_syscall_encoded(1, 2, 0xd4000001),
-#        else
+#    else
         // TODO i#5871: Add AArch32 (and RISC-V) encodings.
-#        endif
+#    endif
         };
         if (!run_checker(memrefs, true, 1, 5, "Repeated syscall instrs with the same PC",
                          "Failed to catch repeated syscall instrs with the same PC"))
@@ -492,25 +491,24 @@ check_repeated_syscall_with_same_pc()
     {
         std::vector<memref_t> memrefs = {
             gen_instr(1, 1),
-#        if defined(X86_64) || defined(X86_32)
+#    if defined(X86_64) || defined(X86_32)
             gen_syscall_encoded(1, 2, { 0x0f, 0x05 }),
             gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
             gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
             gen_syscall_encoded(1, 4, { 0x0f, 0x05 }),
-#        elif defined(ARM_64)
+#    elif defined(ARM_64)
             gen_syscall_encoded(1, 2, 0xd4000001),
             gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
             gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
             gen_syscall_encoded(1, 6, 0xd4000001),
-#        else
+#    else
         // TODO i#5871: Add AArch32 (and RISC-V) encodings.
-#        endif
+#    endif
         };
         if (!run_checker(memrefs, false)) {
             return false;
         }
     }
-#    endif
 #endif
     return true;
 }
--- a/clients/drcachesim/tests/invariant_checker_test.cpp
+++ b/clients/drcachesim/tests/invariant_checker_test.cpp
@@ -215,8 +215,10 @@ check_sane_control_flow()
             gen_marker(1, TRACE_MARKER_TYPE_FILETYPE, OFFLINE_FILE_TYPE_ENCODINGS),
 #    if defined(X86_64) || defined(X86_32)
             // 0x74 is "je" with the 2nd byte the offset.
+            gen_instr_encoded(0x71019dba, { 0x20 }),
             gen_branch_encoded(1, 0x71019dbc, { 0x74, 0x32 }),
 #    elif defined(ARM_64)
+            gen_instr_encoded(0x71019db9, 0x20),
             // 71019dbc:   540001a1        b.ne    71019df0 <__executable_start+0x19df0>
             gen_branch_encoded(1, 0x71019dbc, 0x540001a1),
 #    else
--- a/clients/drcachesim/tests/invariant_checker_test.cpp
+++ b/clients/drcachesim/tests/invariant_checker_test.cpp
@@ -478,6 +478,7 @@ check_repeated_syscall_with_same_pc()
             gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
             gen_instr_encoded(1, ADDR_ONE, { 0x0f, 0x05 }),
 #    elif defined(ARM_64)
+            gen_instr_encoded(1 /*tid*/, 0 /*pc*/, 0x01),
             gen_instr_encoded(1, ADDR_ONE, 0xd4000001),
             gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
             gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
@@ -503,10 +504,10 @@ check_repeated_syscall_with_same_pc()
             gen_instr_encoded(1, ADDR_TWO, { 0x0f, 0x05 }),
 #    elif defined(ARM_64)
             gen_instr_encoded(1 /*tid*/, 0 /*pc*/, 0x01),
-            gen_instr_encoded(1, ADDR_ONE, 0xd4000001),
+            gen_instr_encoded(1, ADDR_ONE, 0xd4000001, 2),
             gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
             gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
-            gen_instr_encoded(1, ADDR_TWO + 2, 0xd4000001),
+            gen_instr_encoded(1, ADDR_TWO, 0xd4000001, 2),
 #    else
         // TODO i#5871: Add AArch32 (and RISC-V) encodings.
 #    endif
--- a/clients/drcachesim/tests/invariant_checker_test.cpp
+++ b/clients/drcachesim/tests/invariant_checker_test.cpp
@@ -464,7 +464,7 @@ check_function_markers()
 }
 
 bool
-check_double_syscall_with_same_pc()
+check_duplicate_syscall_with_same_pc()
 {
     constexpr addr_t ADDR = 0x7fcf3b9dd9e9;
     // Negative: syscalls with the same PC.
@@ -488,7 +488,7 @@ check_double_syscall_with_same_pc()
         // TODO i#5871: Add AArch32 (and RISC-V) encodings.
 #    endif
         };
-        if (!run_checker(memrefs, true, 1, 5, "Double syscall instrs with the same PC",
+        if (!run_checker(memrefs, true, 1, 5, "Duplicate syscall instrs with the same PC",
                          "Failed to catch double syscall instrs with the same PC"))
             return false;
     }
@@ -528,7 +528,7 @@ main(int argc, const char *argv[])
 {
     if (check_branch_target_after_branch() && check_sane_control_flow() &&
         check_kernel_xfer() && check_rseq() && check_function_markers() &&
-        check_double_syscall_with_same_pc()) {
+        check_duplicate_syscall_with_same_pc()) {
         std::cerr << "invariant_checker_test passed\n";
         return 0;
     }
--- a/clients/drcachesim/tests/invariant_checker_test.cpp
+++ b/clients/drcachesim/tests/invariant_checker_test.cpp
@@ -466,30 +466,22 @@ check_repeated_syscall_with_same_pc()
 {
     constexpr addr_t ADDR_ONE = 0x7fcf3b9dd9e9;
     constexpr addr_t ADDR_TWO = 0x7fcf3b9dd9eb;
-    const std::vector<char> encoding = { 0x01 };
-    // TODO i#5949: Add generic function for creating encoded instrs.
-    memref_t encoded_instr =
-        gen_instr_type(TRACE_TYPE_INSTR, 1 /*tid*/, 0 /*pc*/, 1 /*size*/);
-    encoded_instr.instr.type = TRACE_TYPE_INSTR;
-    memcpy(encoded_instr.instr.encoding, &encoding, sizeof(encoding));
-    encoded_instr.instr.encoding_is_new = true;
-
     // Negative: syscalls with the same PC.
 #if defined(X86_64) || defined(X86_32) || defined(ARM_64)
     {
         std::vector<memref_t> memrefs = {
             gen_marker(1, TRACE_MARKER_TYPE_FILETYPE, OFFLINE_FILE_TYPE_ENCODINGS),
-            encoded_instr,
 #    if defined(X86_64) || defined(X86_32)
-            gen_syscall_encoded(1, ADDR_ONE),
+            gen_instr_encoded(1 /*tid*/, 0 /*pc*/, { 0x01 }),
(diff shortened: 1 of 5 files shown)


clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -409,6 +409,24 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                 instr_free(GLOBAL_DCONTEXT, &instr);
             }
         }
+        if (shard->prev_instr_.instr.addr != 0 /*first*/) {
+            instr_t prev_instr, curr_instr;
+            instr_init(GLOBAL_DCONTEXT, &curr_instr);
+            instr_init(GLOBAL_DCONTEXT, &prev_instr);
+            decode_from_copy(
```
abhinav92003, line 416: Instead of decoding the prev instr again, we can add a bool in per_shard_t for was_prev_instr_syscall and set it at the end of each iteration (and use it in the next iteration).

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -409,6 +409,24 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+                GLOBAL_DCONTEXT, const_cast<app_pc>(shard->prev_instr_.instr.encoding),
+                reinterpret_cast<app_pc>(shard->prev_instr_.instr.addr), &prev_instr);
+            decode_from_copy(GLOBAL_DCONTEXT, const_cast<app_pc>(memref.instr.encoding),
```
abhinav92003, line 419: We're already decoding the cur instr for the "Branch target is not decodeable" check. We should decode only once and use it in both checks.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -440,7 +458,8 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                           TRACE_MARKER_TYPE_KERNEL_XFER)) ||
                     // We expect a gap on a window transition.
                     shard->window_transition_ ||
-                    shard->prev_instr_.instr.type == TRACE_TYPE_INSTR_SYSENTER,
+                    shard->prev_instr_.instr.type == TRACE_TYPE_INSTR_SYSENTER ||
+                    shard->saw_double_syscall_instrs_with_same_pc_,
                 "Non-explicit control flow has no marker");
```
abhinav92003, line 463: So we'll report the new invariant error type and the old one too? I think we need only one.

abhinav92003, line 463: Please add a test.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -440,7 +458,8 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                           TRACE_MARKER_TYPE_KERNEL_XFER)) ||
                     // We expect a gap on a window transition.
                     shard->window_transition_ ||
-                    shard->prev_instr_.instr.type == TRACE_TYPE_INSTR_SYSENTER,
+                    shard->prev_instr_.instr.type == TRACE_TYPE_INSTR_SYSENTER ||
```
abhinav92003, line 461: Why do we need this additional condition here?

abhinav92003 (changes requested):

derekbruening:
I changed the description: s/Fixed/Issue:/ since I made the issue cover root-causing and fixing the bug as well as there was nothing filed on that.

Also note that the title syntax is incorrect: missing colon (and if we have a true fix later there should also be a mini-description before the colon).

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: I read this wrong. Ignore.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -440,7 +458,8 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+                    shard->saw_double_syscall_instrs_with_same_pc_,
```
abhinav92003, line 462: saw_double_syscall_instrs_with_same_pc_ probably doesn't need to be a shard var.

clients/drcachesim/tools/invariant_checker.cpp
bete0: `saw_double_syscall_instrs_with_same_pc_` avoids it from the being reported as both types.

clients/drcachesim/tools/invariant_checker.cpp
bete0: This is to avoid the new repeated syscall type errors to from being also reported as 'Non-explicit control flow...' errors.

clients/drcachesim/tools/invariant_checker.cpp
bete0: The current instr is being decoded inside an else clause in "Branch target is not decodeable" error so it's possible we might not always see it by the time we want to use it for the Repeated syscall check.

abhinav92003: I mean: can we move both `decode_from_copy`s to a common location so that we need to do it only once per iteration?

bete0: Ack on the approach but note that in "Branch target is not decodeable", it's the previous instr that's being decoded.

abhinav92003: How about always decoding the current instr in each iteration, and then assigning it to `shard->decoded_prev_instr` pointer so that the next iteration can use it?

I think it'll be good for overhead to extract the instr decoding to a common location, since we may have more and more checks that require decoding instrs.

clients/drcachesim/tests/memref_gen.h
```diff
@@ -84,7 +84,28 @@ gen_branch_encoded(memref_tid_t tid, addr_t pc, int encoding)
     memref.instr.encoding_is_new = true;
     return memref;
 }
+
+inline memref_t
+gen_syscall_encoded(memref_tid_t tid, addr_t pc, int encoding)
```
abhinav92003, line 89: Do we need to have the `encoding` parameter? It's the same in all invocations.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -409,6 +409,20 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                 instr_free(GLOBAL_DCONTEXT, &instr);
             }
         }
+        bool saw_double_syscall_instrs_with_same_pc = false;
+        if (shard->prev_instr_.instr.addr != 0 /*first*/) {
+            instr_t curr_instr;
+            instr_init(GLOBAL_DCONTEXT, &curr_instr);
```
abhinav92003, line 415: Minor style nit: s/curr_instr/cur_instr/

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -440,7 +454,8 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                           TRACE_MARKER_TYPE_KERNEL_XFER)) ||
                     // We expect a gap on a window transition.
                     shard->window_transition_ ||
-                    shard->prev_instr_.instr.type == TRACE_TYPE_INSTR_SYSENTER,
+                    shard->prev_instr_.instr.type == TRACE_TYPE_INSTR_SYSENTER ||
+                    saw_double_syscall_instrs_with_same_pc,
```
abhinav92003, line 458: Add a comment clarifying that we skip this check because we don't want to double-report.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -409,6 +409,20 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                 instr_free(GLOBAL_DCONTEXT, &instr);
             }
         }
+        bool saw_double_syscall_instrs_with_same_pc = false;
```
abhinav92003, line 412: nit: s/double/repeated/

abhinav92003 (changes requested):

clients/drcachesim/tests/memref_gen.h
derekbruening, line 89: There's nothing here specific to a syscall so the name should perhaps be gen_instr_encoded; for that matter maybe it should take in a trace entry type and we then eliminate the nearly-duplicate gen_branch_encoded and only have these ifdefs and different encoding types in one place.

abhinav92003: I suggested above to remove the encoding parameter let the routine add it on its own, which would make the routine specific to syscalls. Having a dedicated routine for syscalls avoids having to specify the encoding in each call. But I'm fine either way.

derekbruening: I'm guessing in the future we may want to write instruction sequences with encodings for arbitrary code (I'm particularly thinking of testing virtual register dep chains) and won't want to create a "gen_add_with_such_operands" etc. for each variety so I would vote for not having a syscall-specific one (we would have encodings grabbed from IR and turned into a record via sthg like gen_instr_encoded).

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -409,6 +409,20 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+        if (shard->prev_instr_.instr.addr != 0 /*first*/) {
+            instr_t curr_instr;
+            instr_init(GLOBAL_DCONTEXT, &curr_instr);
+            decode_from_copy(GLOBAL_DCONTEXT, const_cast<app_pc>(memref.instr.encoding),
```
abhinav92003, line 416: Re the Windows crash: it should help to check the return value of this call here.

clients/drcachesim/tests/memref_gen.h
bete0: Removed it.

abhinav92003:
Please request re-review using the Github UI when ready.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -461,13 +461,79 @@ check_function_markers()
     return true;
 }
 
+bool
+check_repeated_syscall_with_same_pc()
+{
+    constexpr addr_t ADDR_ONE = 0x7fcf3b9dd9e9;
```
abhinav92003, line 467: The mock addresses can be something simple instead of this hex value.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -461,13 +461,79 @@ check_function_markers()
...
+#    if defined(X86_64) || defined(X86_32)
+            gen_syscall_encoded(1, ADDR_ONE),
+            gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
+            gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
+            gen_syscall_encoded(1, ADDR_TWO),
+#    elif defined(ARM_64)
+            gen_syscall_encoded(1, ADDR_ONE),
+            gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
+            gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
+            gen_syscall_encoded(1, ADDR_TWO + 2),
```
abhinav92003, line 516: Why the `+2`?

abhinav92003, line 516: Looks like addr_one followed by addr_two doesn't trigger the pc discontinuity error. Just curious about why.

clients/drcachesim/tools/invariant_checker.h
```diff
@@ -97,6 +97,7 @@ class invariant_checker_t : public analysis_tool_t {
         int instrs_until_interrupt_ = -1;
         int memrefs_until_interrupt_ = -1;
 #endif
+        bool prev_instr_was_syscall_ = false;
```
abhinav92003, line 100: This is unused.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -376,39 +376,58 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
-                    decode_from_copy(GLOBAL_DCONTEXT, decode_pc,
-                                     reinterpret_cast<app_pc>(trace_pc), &instr);
-                if (next_pc == nullptr || !opnd_is_pc(instr_get_target(&instr))) {
-                    // Neither condition should happen but they could on an invalid
-                    // encoding from raw2trace or the reader so we report an
-                    // invariant rather than asserting.
-                    report_if_false(shard, false, "Branch target is not decodeable");
-                } else {
+            instr_t instr;
+            instr_init(GLOBAL_DCONTEXT, &instr);
```
abhinav92003, line 385: So instr is really the previous instr?

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -376,39 +376,58 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+                        cond_branch_target = reinterpret_cast<addr_t>(
+                            opnd_get_pc(instr_get_target(&instr)));
+                        shard->branch_target_cache[trace_pc] = cond_branch_target;
+                    }
                 }
-                instr_free(GLOBAL_DCONTEXT, &instr);
+            } else {
+                instr_t cur_instr;
+                instr_init(GLOBAL_DCONTEXT, &cur_instr);
+                app_pc curr_pc = decode_from_copy(
```
abhinav92003, line 413: curr_pc is not aptly named. decode_from_copy always returns the pc of the next instr.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -376,39 +376,58 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+                        // Set this flag so that repeated syscalls are not
+                        // double reporeted as other PC discontinuity below.
+                        saw_repeated_syscall_instrs_with_same_pc = true;
+                        report_if_false(shard, false,
+                                        "Repeated syscall instrs with the same PC");
+                    }
+                }
+                instr_free(GLOBAL_DCONTEXT, &cur_instr);
             }
+            instr_free(GLOBAL_DCONTEXT, &instr);
```
abhinav92003, line 428: We're still decoding two instrs for this check, which is not optimal.  See unresolved comment from previous reviews. Instead of decoding the prev instr again, can we decode it only one time (during its own iteration) and then store in `shard` and use it in the next iteration?

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003, line 413: We still have multiple code snippets that are decoding instrs. Like I mentioned in previous review, in each iteration we should be decoding only the cur instr (to reduce decoding overhead) and store it in the `shard` for use in the next iteration.

clients/drcachesim/tests/memref_gen.h
```diff
@@ -84,7 +84,30 @@ gen_branch_encoded(memref_tid_t tid, addr_t pc, int encoding)
...
+    memref_t memref = gen_instr_type(TRACE_TYPE_INSTR, tid, pc);
+    memref.instr.size = 4;
+    memcpy(memref.instr.encoding, &encoding, sizeof(encoding));
+    memref.instr.encoding_is_new = true;
+    return memref;
+}
+
 #elif defined(X86_64) || defined(X86_32)
+inline memref_t
+gen_syscall_encoded(memref_tid_t tid, addr_t pc)
```
abhinav92003, line 101: See unresolved comment about not wanting to add a gen_ routine specific to syscalls.

abhinav92003 (changes requested):

clients/drcachesim/tests/memref_gen.h
bete0: Removed the syscall specific one and added `gen_instr_encoded`.

clients/drcachesim/tests/invariant_checker_test.cpp
bete0: Was using the default size of 1 and that was causing the generic PC discontuity error; added a size paramter to this call.

clients/drcachesim/tools/invariant_checker.cpp
bete0: Right and that's one of the tricky aspects of the decoding logic: We won't be able to know if the "current" instr was a syscall until we process the instr after it. But by the time we get to process the following instr (and discover instr was a syscall), we will have already hit a generic PC discontinuity.

clients/drcachesim/tests/invariant_checker_test.cpp
bete0: When we see repeated syscalls, we also check if they have the same address or not. Only report it if they do have the same addresses. Maybe it should be relaxed to not care about the addresses?

clients/drcachesim/tests/invariant_checker_test.cpp
abhinav92003: Reminder for this unresolved comment.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -461,13 +461,78 @@ check_function_markers()
...
+            gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
+            gen_instr_encoded(1, ADDR_TWO, { 0x0f, 0x05 }),
+            gen_instr_encoded(1 /*tid*/, ADDR_THREE /*pc*/, { 0x01 }),
+#    elif defined(ARM_64)
+            gen_instr_encoded(1 /*tid*/, 0 /*pc*/, 0x01),
+            gen_instr_encoded(1, ADDR_ONE, 0xd4000001, 2),
+            gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
+            gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
+            gen_instr_encoded(1, ADDR_TWO, 0xd4000001, 2),
+            gen_instr_encoded(1 /*tid*/, ADDR_THREE /*pc*/, 0x01),
```
abhinav92003, line 515: We could just use either ADDR_ONE or ADDR_TWO instead of the new address.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -461,13 +461,78 @@ check_function_markers()
...
+    constexpr addr_t ADDR_ONE = 0x7fcf3b9dd9e9;
+    constexpr addr_t ADDR_TWO = 0x7fcf3b9dd9eb;
+    constexpr addr_t ADDR_THREE = 0x7fcf3b9dd9ed;
+    // Negative: syscalls with the same PC.
+#if defined(X86_64) || defined(X86_32) || defined(ARM_64)
+    {
+        std::vector<memref_t> memrefs = {
+            gen_marker(1, TRACE_MARKER_TYPE_FILETYPE, OFFLINE_FILE_TYPE_ENCODINGS),
+#    if defined(X86_64) || defined(X86_32)
+            gen_instr_encoded(1 /*tid*/, 0 /*pc*/, { 0x01 }),
```
abhinav92003, line 476: Since we have the same value of tid for all these calls, I'd suggest we specify a default value in the definition of gen_instr_encoded so that we don't need to specify it repeatedly.

abhinav92003, line 476: For all calls of gen_instr_encoded, also add the human-readable encoding in a comment to the right of the call.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -461,13 +461,78 @@ check_function_markers()
...
+            gen_instr_encoded(1, ADDR_ONE, { 0x0f, 0x05 }),
+            gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
+            gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
+            gen_instr_encoded(1, ADDR_TWO, { 0x0f, 0x05 }),
+            gen_instr_encoded(1 /*tid*/, ADDR_THREE /*pc*/, { 0x01 }),
```
abhinav92003, line 508: The `/*pc*/` is not really needed when the identifier of the variable is descriptive enough to make it clear to the reader what the arg is.

clients/drcachesim/tests/memref_gen.h
```diff
@@ -75,6 +75,15 @@ gen_branch(memref_tid_t tid, addr_t pc)
 // We use these client defines which are the target and so drdecode's target arch.
 #if defined(ARM_64) || defined(ARM_32)
 // Variant for aarchxx encodings.
+inline memref_t
+gen_instr_encoded(memref_tid_t tid, addr_t pc, int encoding, size_t size = 1)
+{
```
abhinav92003, line 80: About the `size` arg: AArch64 has fixed sized instrs.

clients/drcachesim/tests/memref_gen.h
```diff
@@ -84,7 +93,19 @@ gen_branch_encoded(memref_tid_t tid, addr_t pc, int encoding)
     memref.instr.encoding_is_new = true;
     return memref;
 }
+
 #elif defined(X86_64) || defined(X86_32)
+inline memref_t
+gen_instr_encoded(memref_tid_t tid, addr_t pc, const std::vector<char> encoding,
+                  size_t size = 1)
```
abhinav92003, line 100: About the `size` arg: can that be deduced from the encoding itself?

clients/drcachesim/tests/invariant_checker_test.cpp
abhinav92003: > Maybe it should be relaxed to not care about the addresses?

No, it should check the addresses; the invariant we want to check for is the same syscall being repeated.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -376,39 +376,56 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+                            opnd_get_pc(instr_get_target(&instr)));
+                        shard->branch_target_cache[trace_pc] = cond_branch_target;
+                    }
+                }
+            } else {
+                if (next_pc != nullptr) {
+                    if (instr_is_syscall(&instr) &&
+                        shard->prev_instr_addr_ == shard->prev_instr_.instr.addr &&
+                        shard->prev_instr_was_syscall_) {
+                        std::cerr << "Saw double syscall\n";
```
abhinav92003, line 415: Remove std::cerr.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -440,7 +457,9 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                           TRACE_MARKER_TYPE_KERNEL_XFER)) ||
                     // We expect a gap on a window transition.
                     shard->window_transition_ ||
-                    shard->prev_instr_.instr.type == TRACE_TYPE_INSTR_SYSENTER,
+                    shard->prev_instr_.instr.type == TRACE_TYPE_INSTR_SYSENTER ||
+                    saw_repeated_syscall_instrs_with_same_pc ||
+                    shard->prev_instr_was_syscall_,
```
abhinav92003, line 462: I'm confused about this: if the previous instr is a syscall we don't report any discontinuities?

abhinav92003 (changes requested):

clients/drcachesim/tools/invariant_checker.cpp
bete0: Ack. Removed it: After changes with the decoding logic, it is not needed. Previously it was reporting discontinuities at the syscall instr before we detected the double syscall on the following instr.

clients/drcachesim/tests/invariant_checker_test.cpp
bete0: Done.

clients/drcachesim/tests/invariant_checker_test.cpp
bete0: Just kept this one and removed the other ones based on comment below.

clients/drcachesim/tests/invariant_checker_test.cpp
bete0: Done.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -503,17 +495,18 @@ check_repeated_syscall_with_same_pc()
...
             gen_marker(1, TRACE_MARKER_TYPE_FILETYPE, OFFLINE_FILE_TYPE_ENCODINGS),
-            encoded_instr,
 #    if defined(X86_64) || defined(X86_32)
-            gen_syscall_encoded(1, ADDR_ONE),
+            gen_instr_encoded(1 /*tid*/, 0 /*pc*/, { 0x01 }),
+            gen_instr_encoded(1, ADDR_ONE, { 0x0f, 0x05 }),
             gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
             gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
-            gen_syscall_encoded(1, ADDR_TWO),
+            gen_instr_encoded(1, ADDR_TWO, { 0x0f, 0x05 }),
```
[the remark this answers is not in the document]
bete0: Done.

clients/drcachesim/tests/invariant_checker_test.cpp
bete0: Done. Removed it.

clients/drcachesim/tests/invariant_checker_test.cpp
bete0: No, it was a stale var. Removed it.

clients/drcachesim/tests/memref_gen.h
bete0: Ack. Removed the size arg and used 4 in the call to `gen_instr_type`

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -338,6 +338,18 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
     if (type_is_instr(memref.instr.type) ||
         memref.instr.type == TRACE_TYPE_PREFETCH_INSTR ||
         memref.instr.type == TRACE_TYPE_INSTR_NO_FETCH) {
+        // We do not bother to support legacy traces without encodings.
+        bool has_encoding = TESTANY(OFFLINE_FILE_TYPE_ENCODINGS, shard->file_type_);
```
abhinav92003, line 342: s/has_encoding/expect_encoding/

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -338,6 +338,18 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+        addr_t trace_pc = shard->prev_instr_.instr.addr;
```
abhinav92003, line 343: As discussed offline: we should decode the **current** instr in each iteration.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -338,6 +338,18 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+        instr_t cur_instr_decoded;
+        instr_init(GLOBAL_DCONTEXT, &cur_instr_decoded);
+        const app_pc decode_pc = const_cast<app_pc>(shard->prev_instr_.instr.encoding);
+        const app_pc next_pc =
+            decode_from_copy(GLOBAL_DCONTEXT, decode_pc,
+                             reinterpret_cast<app_pc>(trace_pc), &cur_instr_decoded);
```
abhinav92003, line 349: It is confusing that we decode the `prev_instr_`'s encoding into `cur_instr_decoded`. We also don't want that (see comment above).

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -377,36 +389,41 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
             } else {
-                instr_t instr;
-                instr_init(GLOBAL_DCONTEXT, &instr);
-                const app_pc decode_pc =
-                    const_cast<app_pc>(shard->prev_instr_.instr.encoding);
-                const app_pc next_pc =
-                    decode_from_copy(GLOBAL_DCONTEXT, decode_pc,
-                                     reinterpret_cast<app_pc>(trace_pc), &instr);
-                if (next_pc == nullptr || !opnd_is_pc(instr_get_target(&instr))) {
+                if ((shard->prev_instr_decoded_ == nullptr ||
```
abhinav92003, line 400: Remove redundant parenthesis pair.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -484,6 +502,12 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
         }
 #endif
         shard->prev_instr_ = memref;
+        if (next_pc != nullptr) {
+            shard->prev_instr_was_syscall_ = instr_is_syscall(&cur_instr_decoded);
```
abhinav92003, line 506: Do we need `prev_instr_was_syscall_`? It doesn't seem to be read anywhere.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -484,6 +502,12 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+            shard->prev_instr_decoded_ = &cur_instr_decoded;
+            shard->prev_instr_addr_ = shard->prev_instr_.instr.addr;
```
abhinav92003, line 508: Do we need `prev_instr_addr_`? It doesn't seem to be read anywhere.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -484,6 +502,12 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
         }
 #endif
         shard->prev_instr_ = memref;
+        if (next_pc != nullptr) {
+            shard->prev_instr_was_syscall_ = instr_is_syscall(&cur_instr_decoded);
+            shard->prev_instr_decoded_ = &cur_instr_decoded;
```
abhinav92003, line 507: As discussed offline: `cur_instr_decoded` needs to be heap-allocated. `&cur_instr_decoded` will go out of scope when this iteration ends, which means `prev_instr_decoded_` will become a dangling pointer.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -484,6 +502,12 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+            shard->prev_instr_addr_ = shard->prev_instr_.instr.addr;
+            instr_free(GLOBAL_DCONTEXT, &cur_instr_decoded);
```
abhinav92003, line 509: We do not want instr_free for the current instr here since it is still needed in the next iteration. We want it for the prev instr (before we overwrite it with the cur instr above).

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -377,36 +389,41 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
         bool have_cond_branch_target = false;
         addr_t cond_branch_target = 0;
         if (shard->prev_instr_.instr.addr != 0 /*first*/ &&
-            type_is_instr_direct_branch(shard->prev_instr_.instr.type) &&
-            // We do not bother to support legacy traces without encodings.
-            TESTANY(OFFLINE_FILE_TYPE_ENCODINGS, shard->file_type_)) {
-            addr_t trace_pc = shard->prev_instr_.instr.addr;
+            type_is_instr_direct_branch(shard->prev_instr_.instr.type) && has_encoding) {
             if (shard->prev_instr_.instr.encoding_is_new)
                 shard->branch_target_cache.erase(trace_pc);
```
abhinav92003, line 394: In this check, we should use the prev instr's pc (prev_instr_.instr.addr). `trace_pc` erroneously contains that currently (see other comment about wanting to decode the **current** instr in each iteration, which means `trace_pc` should really have the current instr's pc).

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -338,6 +338,18 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
     if (type_is_instr(memref.instr.type) ||
         memref.instr.type == TRACE_TYPE_PREFETCH_INSTR ||
         memref.instr.type == TRACE_TYPE_INSTR_NO_FETCH) {
+        // We do not bother to support legacy traces without encodings.
```
abhinav92003, line 341: This comment is not relevant here. It should be placed near the invariant check where we use `has_encoding`.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -338,6 +338,18 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+        bool has_encoding = TESTANY(OFFLINE_FILE_TYPE_ENCODINGS, shard->file_type_);
+        addr_t trace_pc = shard->prev_instr_.instr.addr;
+        instr_t cur_instr_decoded;
+        instr_init(GLOBAL_DCONTEXT, &cur_instr_decoded);
+        const app_pc decode_pc = const_cast<app_pc>(shard->prev_instr_.instr.encoding);
+        const app_pc next_pc =
+            decode_from_copy(GLOBAL_DCONTEXT, decode_pc,
+                             reinterpret_cast<app_pc>(trace_pc), &cur_instr_decoded);
+        if (next_pc == nullptr) {
+            instr_free(GLOBAL_DCONTEXT, &cur_instr_decoded);
```
abhinav92003, line 351: If decoding fails, we also want to set cur_instr_decoded to nullptr (also see other comment about making it heap-allocated).

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -377,36 +389,41 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+                    cond_branch_target = reinterpret_cast<addr_t>(
+                        opnd_get_pc(instr_get_target(shard->prev_instr_decoded_)));
                     shard->branch_target_cache[trace_pc] = cond_branch_target;
                 }
-                instr_free(GLOBAL_DCONTEXT, &instr);
+            }
+        }
+        bool saw_repeated_syscall_instrs_with_same_pc = false;
+        if (shard->prev_instr_.instr.addr != 0 /*first*/) {
+            if (next_pc != nullptr) {
```
abhinav92003, line 417: Would be more readable to check cur_instr_decoded != nullptr here (see other comment about setting cur_instr_decoded to nullptr if decoding fails).

abhinav92003, line 417: Also prev_instr_decoded_ should not be nullptr.

abhinav92003 (changes requested):

clients/drcachesim/tools/invariant_checker.cpp
bete0: Ack.

clients/drcachesim/tools/invariant_checker.cpp
bete0: Ack. removed it.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -338,17 +338,20 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
-        const app_pc next_pc =
-            decode_from_copy(GLOBAL_DCONTEXT, decode_pc,
-                             reinterpret_cast<app_pc>(trace_pc), &cur_instr_decoded);
-        if (next_pc == nullptr) {
-            instr_free(GLOBAL_DCONTEXT, &cur_instr_decoded);
+        bool expect_encoding = TESTANY(OFFLINE_FILE_TYPE_ENCODINGS, shard->file_type_);
+        addr_t trace_pc = memref.instr.addr;
+        std::unique_ptr<instr_t> cur_instr_decoded(new instr_t());
+        app_pc next_pc = nullptr;
+        if (TESTANY(OFFLINE_FILE_TYPE_ENCODINGS, shard->file_type_)) {
```
abhinav92003, line 345: s/TESTANY(OFFLINE_FILE_TYPE_ENCODINGS, shard->file_type_)/expect_encoding/

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -338,17 +338,20 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
-        const app_pc decode_pc = const_cast<app_pc>(shard->prev_instr_.instr.encoding);
-        const app_pc next_pc =
-            decode_from_copy(GLOBAL_DCONTEXT, decode_pc,
-                             reinterpret_cast<app_pc>(trace_pc), &cur_instr_decoded);
-        if (next_pc == nullptr) {
-            instr_free(GLOBAL_DCONTEXT, &cur_instr_decoded);
+        bool expect_encoding = TESTANY(OFFLINE_FILE_TYPE_ENCODINGS, shard->file_type_);
+        addr_t trace_pc = memref.instr.addr;
+        std::unique_ptr<instr_t> cur_instr_decoded(new instr_t());
+        app_pc next_pc = nullptr;
```
abhinav92003, line 344: Move the next_pc declaration to the narrower if-block scope below.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -338,17 +338,20 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+        if (TESTANY(OFFLINE_FILE_TYPE_ENCODINGS, shard->file_type_)) {
+            instr_init(GLOBAL_DCONTEXT, cur_instr_decoded.get());
+            const app_pc decode_pc = const_cast<app_pc>(memref.instr.encoding);
+            next_pc = decode_from_copy(GLOBAL_DCONTEXT, decode_pc,
+                                       reinterpret_cast<app_pc>(trace_pc),
+                                       cur_instr_decoded.get());
+            if (next_pc == nullptr) {
+                instr_free(GLOBAL_DCONTEXT, cur_instr_decoded.get());
+                cur_instr_decoded = nullptr;
```
abhinav92003, line 353: .reset(nullptr)

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -388,43 +391,39 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
         // by markers.
         bool have_cond_branch_target = false;
         addr_t cond_branch_target = 0;
-        if (shard->prev_instr_.instr.addr != 0 /*first*/ &&
-            type_is_instr_direct_branch(shard->prev_instr_.instr.type) && has_encoding) {
-            if (shard->prev_instr_.instr.encoding_is_new)
+        if (type_is_instr_direct_branch(memref.instr.type) && expect_encoding) {
```
abhinav92003, line 394: Why did we remove the `first` check?

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -388,43 +391,39 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
-                    saw_repeated_syscall_instrs_with_same_pc = true;
-                    report_if_false(shard, false,
-                                    "Repeated syscall instrs with the same PC");
-                }
-            }
+        if (shard->prev_instr_.instr.addr != 0 /*first*/ &&
+            instr_is_syscall(cur_instr_decoded.get()) &&
+            memref.instr.addr == shard->prev_instr_.instr.addr &&
+            instr_is_syscall(shard->prev_instr_decoded_) &&
+            cur_instr_decoded != nullptr && shard->prev_instr_decoded_ != nullptr) {
```
abhinav92003, line 422: These `!= nullptr` checks should be moved to be before the `instr_is_syscall` checks (so that those calls are skipped for nullptrs)

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -215,8 +215,10 @@ check_sane_control_flow()
             gen_marker(1, TRACE_MARKER_TYPE_FILETYPE, OFFLINE_FILE_TYPE_ENCODINGS),
 #    if defined(X86_64) || defined(X86_32)
             // 0x74 is "je" with the 2nd byte the offset.
+            gen_instr_encoded(0x71019dba, { 0x20 }),
```
abhinav92003, line 218: Why do we need these changes in this test?

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -461,13 +463,77 @@ check_function_markers()
...
+check_repeated_syscall_with_same_pc()
+{
+    constexpr addr_t ADDR_ONE = 0x7fcf3b9dd9e9;
+    // Negative: syscalls with the same PC.
+#if defined(X86_64) || defined(X86_32) || defined(ARM_64)
+    {
+        std::vector<memref_t> memrefs = {
+            gen_marker(1, TRACE_MARKER_TYPE_FILETYPE, OFFLINE_FILE_TYPE_ENCODINGS),
+#    if defined(X86_64) || defined(X86_32)
+            gen_instr_encoded(0, { 0x01 }),
```
abhinav92003, line 476: Why do we need this first instr? Why does it not return a pc discontinuity error for the unexplained jump from pc=0 to ADDR_ONE? (Not sure what the decoded instr is; pls add it to the comment like for the others)

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -461,13 +463,77 @@ check_function_markers()
...
+                         "Failed to catch repeated syscall instrs with the same PC"))
+            return false;
+    }
+
+    // Positive test: syscalls with different PCs.
+    {
+        std::vector<memref_t> memrefs = {
+            gen_marker(1, TRACE_MARKER_TYPE_FILETYPE, OFFLINE_FILE_TYPE_ENCODINGS),
+#    if defined(X86_64) || defined(X86_32)
+            gen_instr_encoded(0, { 0x01 }),
```
abhinav92003, line 503: Same here.

clients/drcachesim/tests/memref_gen.h
```diff
@@ -84,7 +93,18 @@ gen_branch_encoded(memref_tid_t tid, addr_t pc, int encoding)
     memref.instr.encoding_is_new = true;
     return memref;
 }
+
 #elif defined(X86_64) || defined(X86_32)
+inline memref_t
+gen_instr_encoded(addr_t pc, const std::vector<char> encoding, memref_tid_t tid = 1)
```
abhinav92003, line 99: Pass the vector by reference to avoid copying.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -376,39 +391,40 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
             } else {
-                instr_t instr;
-                instr_init(GLOBAL_DCONTEXT, &instr);
-                const app_pc decode_pc =
-                    const_cast<app_pc>(shard->prev_instr_.instr.encoding);
-                const app_pc next_pc =
-                    decode_from_copy(GLOBAL_DCONTEXT, decode_pc,
-                                     reinterpret_cast<app_pc>(trace_pc), &instr);
-                if (next_pc == nullptr || !opnd_is_pc(instr_get_target(&instr))) {
+                if (cur_instr_decoded == nullptr ||
```
abhinav92003, line 403: The branch target check should be on the prev instr, not the current one.

clients/drcachesim/tools/invariant_checker.h
```diff
@@ -83,8 +84,10 @@ class invariant_checker_t : public analysis_tool_t {
             last_xfer_marker_.marker.marker_type = TRACE_MARKER_TYPE_VERSION;
         }
         memtrace_stream_t *stream = nullptr;
+
         memref_t prev_entry_ = {};
         memref_t prev_instr_ = {};
+        instr_t *prev_instr_decoded_ = nullptr;
```
abhinav92003, line 90: Why not use a unique_ptr here as well?

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -484,6 +501,9 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
         }
 #endif
         shard->prev_instr_ = memref;
+        if (cur_instr_decoded != nullptr) {
+            shard->prev_instr_decoded_ = cur_instr_decoded.get();
```
abhinav92003, line 505: Unless you use a `std::move`, the object pointed by cur_instr_decoded will be de-allocated when cur_instr_decoded goes out of scope.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: Reminder for this unresolved comment.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: Reminder about this unresolved comment.

abhinav92003 (changes requested):
This PR needs some correctness fixes, so I cannot approve it yet.

I feel there is still some confusion about the checks modified in this PR. Please note that the branch target check should be on the prev instr.

Also there are still some unresolved comments from the last review round.

And there are some changes that are not explained (not sure why they're needed).

clients/drcachesim/tests/invariant_checker_test.cpp
bete0: Removed them.

clients/drcachesim/tests/invariant_checker_test.cpp
bete0: Removed it.

clients/drcachesim/tools/invariant_checker.cpp
bete0: Was doing this change on the current instr and check this check would have been unnecessary in that case but now that we went back to checking the previous instr for decoding check, I have added it back.

clients/drcachesim/tools/invariant_checker.cpp
bete0: Ack.

clients/drcachesim/tools/invariant_checker.cpp
bete0: Done.

clients/drcachesim/tools/invariant_checker.h
bete0: Done.

clients/drcachesim/tests/invariant_checker_test.cpp
bete0: Removed it altogether.

clients/drcachesim/tests/invariant_checker_test.cpp
bete0: Removed this one too.

clients/drcachesim/tools/invariant_checker.cpp
bete0: Removed it.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: I see that the next_pc declaration wasn't moved. Was there any concern in moving it to the if-block?

Please see the section on code reviews here: https://dynamorio.org/page_code_reviews.html#autotoc_md114. It's usually good to have an explicit reply to each comment thread in addition to marking them resolved.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -338,6 +338,21 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
     if (type_is_instr(memref.instr.type) ||
         memref.instr.type == TRACE_TYPE_PREFETCH_INSTR ||
         memref.instr.type == TRACE_TYPE_INSTR_NO_FETCH) {
+        bool expect_encoding = TESTANY(OFFLINE_FILE_TYPE_ENCODINGS, shard->file_type_);
+        std::unique_ptr<instr_t> cur_instr_decoded = nullptr;
+        app_pc next_pc = nullptr;
+        if (expect_encoding) {
+            cur_instr_decoded = std::unique_ptr<instr_t>(new instr_t);
```
abhinav92003, line 345: Or just cur_instr_decoded.reset(new instr_t).

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -338,6 +338,21 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+            instr_init(GLOBAL_DCONTEXT, cur_instr_decoded.get());
+            const app_pc decode_pc = const_cast<app_pc>(memref.instr.encoding);
+            next_pc = decode_from_copy(GLOBAL_DCONTEXT, decode_pc,
```
abhinav92003, line 348: nit: decode_pc can be inlined since it is used only once.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: Reminder to do an instr_free for prev_instr_decoded_ (before replacing it with the cur_instr_decoded).

Again, it'll help to reply to each comment explicitly like our code review guide suggests (https://dynamorio.org/page_code_reviews.html#autotoc_md114).

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -461,13 +461,73 @@ check_function_markers()
...
+                              0xd4000001), // 0x7fcf3b9dd9e9: 0xd4000001 svc #0x0
+            gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
+            gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
+            gen_instr_encoded(ADDR_ONE,
+                              0xd4000001), // 0x7fcf3b9dd9e9: 0xd4000001 svc #0x0
+#    else
+        // TODO i#5871: Add AArch32 (and RISC-V) encodings.
+#    endif
+        };
+        if (!run_checker(memrefs, true, 1, 5, "Repeated syscall instrs with the same PC",
```
abhinav92003, line 489: I think it'll be useful if run_checker also verifies where the error occurred. For now, please add a TODO.

abhinav92003 (changes requested):

abhinav92003:
Thanks for making the changes. Just a few minor suggested changes in my last review.

clients/drcachesim/tests/invariant_checker_test.cpp
derekbruening: Do you mean at which ref count the error occurred?  That's already checked: that's the `5` parameter here.

clients/drcachesim/tools/invariant_checker.cpp
bete0: Done.

clients/drcachesim/tools/invariant_checker.cpp
bete0: Done.

clients/drcachesim/tools/invariant_checker.cpp
bete0: Initially was using `next_pc != nullptr` check down below and it would've gone out of scope but now that we don't use that check, moving it inside the if-block as suggested.

clients/drcachesim/tools/invariant_checker.cpp
bete0: Ack. Please see latest patch where I added:
```
            if (shard->prev_instr_decoded_ != nullptr)
                instr_free(GLOBAL_DCONTEXT, shard->prev_instr_decoded_.get());
```

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -338,6 +338,20 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
     if (type_is_instr(memref.instr.type) ||
         memref.instr.type == TRACE_TYPE_PREFETCH_INSTR ||
         memref.instr.type == TRACE_TYPE_INSTR_NO_FETCH) {
+        bool expect_encoding = TESTANY(OFFLINE_FILE_TYPE_ENCODINGS, shard->file_type_);
+        std::unique_ptr<instr_t> cur_instr_decoded = nullptr;
+        if (expect_encoding && memref.instr.encoding_is_new) {
+            cur_instr_decoded.reset(new instr_t);
+            instr_init(GLOBAL_DCONTEXT, cur_instr_decoded.get());
+            app_pc next_pc = nullptr;
```
abhinav92003, line 346: The declaration of next_pc and the assignment in the next line don't need to be separate.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -374,8 +388,48 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
         }
         // Invariant: non-explicit control flow (i.e., kernel-mediated) is indicated
         // by markers.
-        const std::string non_explicit_flow_violation_msg =
-            check_for_pc_discontinuity(shard, memref);
+        bool have_cond_branch_target = false;
```
abhinav92003, line 391: Both the branch target check and the repeated syscall check should be done inside check_for_pc_discontinuity, since they are also checks for pc discontinuities.

clients/drcachesim/tools/invariant_checker.cpp
bete0: Ack. This was to use the default `nullptr` for the check below but looks like `decode_from_copy` returns NULL on failure. So will combine it as suggested.

clients/drcachesim/tools/invariant_checker.cpp
bete0: Done. Moved them inside the function and passed `cur_instr_decoded` and `expect_encoding` to it.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -417,6 +471,11 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
         }
 #endif
         shard->prev_instr_ = memref;
+        if (cur_instr_decoded != nullptr) {
```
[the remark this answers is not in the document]
bete0: Done.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -374,8 +388,48 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
         // Invariant: non-explicit control flow (i.e., kernel-mediated) is indicated
         // by markers.
-        const std::string non_explicit_flow_violation_msg =
-            check_for_pc_discontinuity(shard, memref);
+        bool have_cond_branch_target = false;
+        addr_t cond_branch_target = 0;
+        if (shard->prev_instr_.instr.addr != 0 /*first*/ &&
+            type_is_instr_direct_branch(shard->prev_instr_.instr.type) &&
+            expect_encoding) {
+            // We do not bother to support legacy traces without encodings.
```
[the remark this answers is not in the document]
bete0: Done. Moved it above `expect_encoding`.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -374,8 +388,48 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
         }
         // Invariant: non-explicit control flow (i.e., kernel-mediated) is indicated
         // by markers.
-        const std::string non_explicit_flow_violation_msg =
-            check_for_pc_discontinuity(shard, memref);
+        bool have_cond_branch_target = false;
+        addr_t cond_branch_target = 0;
+        if (shard->prev_instr_.instr.addr != 0 /*first*/ &&
+            type_is_instr_direct_branch(shard->prev_instr_.instr.type) &&
+            expect_encoding) {
```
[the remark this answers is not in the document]
bete0: Done.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -216,13 +216,14 @@ check_sane_control_flow()
...
             gen_branch_encoded(1, 0x71019dbc, { 0x74, 0x32 }),
+            gen_instr_encoded(0x71019ded, { 0x01 }),
 #    elif defined(ARM_64)
             // 71019dbc:   540001a1        b.ne    71019df0 <__executable_start+0x19df0>
             gen_branch_encoded(1, 0x71019dbc, 0x540001a1),
+            gen_instr_encoded(0x71019ded, 0x01),
 #    else
         // TODO i#5871: Add AArch32 (and RISC-V) encodings.
 #    endif
-            gen_instr(1, 20),
```
bete0, line 225: Changed this to a bigger PC to avoid hitting #5960 on Windows and `TRACE_MARKER_TYPE_FILETYPE, OFFLINE_FILE_TYPE_ENCODINGS` is set, changed it encoded instr.

bete0:
run arm tests

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -663,7 +686,8 @@ invariant_checker_t::check_for_pc_discontinuity(per_shard_t *shard,
                   TRACE_MARKER_TYPE_KERNEL_XFER)) ||
             // We expect a gap on a window transition.
             shard->window_transition_ ||
-            shard->prev_instr_.instr.type == TRACE_TYPE_INSTR_SYSENTER;
+            shard->prev_instr_.instr.type == TRACE_TYPE_INSTR_SYSENTER ||
+            saw_repeated_syscall_instrs_with_same_pc;
```
abhinav92003, line 690: Having repeated syscalls implies an invalid non-branch flow. So it is confusing that valid_nonbranch_flow is set to true when saw_repeated_syscall_instrs_with_same_pc is true.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -596,46 +614,51 @@ invariant_checker_t::print_results()
...
+    bool saw_repeated_syscall_instrs_with_same_pc = false;
+    if (cur_instr_decoded != nullptr && shard->prev_instr_decoded_ != nullptr &&
+        shard->prev_instr_.instr.addr != 0 /*first*/ &&
+        instr_is_syscall(cur_instr_decoded.get()) &&
+        memref.instr.addr == shard->prev_instr_.instr.addr &&
+        instr_is_syscall(shard->prev_instr_decoded_.get())) {
+        // Set this flag so that repeated syscalls are not
+        // double reporeted as other PC discontinuity below.
+        saw_repeated_syscall_instrs_with_same_pc = true;
+        error_msg = "Repeated syscall instrs with the same PC";
```
abhinav92003, line 660: It is confusing that we set the error_msg here but then go on to do the other checks. It'd be better if (similar to the branch target check) we set the error_msg at the end (inside the `if (!valid_nonbranch_flow) {` if block). Also see the other comment on how we're setting `valid_nonbranch_flow`.

abhinav92003 (changes requested):
There are still some correctness concerns and some readability issues. See comments below.

clients/drcachesim/tools/invariant_checker.cpp
bete0: The other existing checks inside that block depend on ` if (shard->prev_instr_.instr.addr != 0 /*first*/) 
` which the Repeated syscall doesn't and the current test sequence only has instrs both of which are syscalls and the double syscall check would be skipped and results in a false negative. 

In an earlier PR I removed an instr before the first syscall in the test sequence after the comments. I can put that instr back infront of the first syscall so that I can put the double syscall check inside ` if (shard->prev_instr_.instr.addr != 0 /*first*/)`

clients/drcachesim/tools/invariant_checker.cpp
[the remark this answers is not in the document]
bete0: Removing the ` if (cur_instr_decoded != nullptr)` on L433 results in a false `"Direct branch does not go to the correct target",  "Failed to catch branch not going to its target"` error.  Investigating why that's happening.

clients/drcachesim/tools/invariant_checker.cpp
bete0: Actually nevermind. That check should still be fine and will move the double syscall check inside that if-block.  I will change the `error_msg` assignment for the `Branch target decode` check back to `report_if_false` and keep it where it is.

clients/drcachesim/tools/invariant_checker.cpp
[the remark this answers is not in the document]
bete0: This is because `shard->prev_instr_decoded_` won't be updated with just a check in the inner loop since we always start from a state where `shard->prev_instr_decode_` was a `nullptr` and we are checking if it != nullptr before updating it. 

```
        if (shard->prev_instr_decoded_ != nullptr) {
            instr_free(GLOBAL_DCONTEXT, shard->prev_instr_decoded_.get());
            shard->prev_instr_decoded_ = std::move(cur_instr_decoded);
        }
```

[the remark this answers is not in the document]
bete0: Nevermind, I misunderstood the comment above. Just removing `(cur_instr_decoded != nullptr` and moving the update outside of the if-block works. 
```
if (shard->prev_instr_decoded_ != nullptr) 
    instr_free(GLOBAL_DCONTEXT, shard->prev_instr_decoded_.get()); 
shard->prev_instr_decoded_ = std::move(cur_instr_decoded);
```

clients/drcachesim/tools/invariant_checker.cpp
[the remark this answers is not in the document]
bete0: Done. Also updated commit title.

bete0: After moving the duplicate syscall check inside the if-block, there is no need for this var. Removed it.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -596,46 +614,51 @@ invariant_checker_t::print_results()
...
 
 std::string
-invariant_checker_t::check_for_pc_discontinuity(per_shard_t *shard,
-                                                const memref_t &memref)
+invariant_checker_t::check_for_pc_discontinuity(
+    per_shard_t *shard, const memref_t &memref,
+    const std::unique_ptr<instr_t> &cur_instr_decoded, const bool expect_encoding)
 {
     std::string error_msg = "";
+    // Invariant: non-explicit control flow (i.e., kernel-mediated) is indicated
```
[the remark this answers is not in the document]
bete0: Moved it back. Stray change from local change.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -338,6 +338,21 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+        if (expect_encoding) {
+            cur_instr_decoded.reset(new instr_t);
+            instr_init(GLOBAL_DCONTEXT, cur_instr_decoded.get());
+            app_pc next_pc = decode_from_copy(
+                GLOBAL_DCONTEXT, const_cast<app_pc>(memref.instr.encoding),
+                reinterpret_cast<app_pc>(memref.instr.addr), cur_instr_decoded.get());
+            if (next_pc == nullptr) {
+                instr_free(GLOBAL_DCONTEXT, cur_instr_decoded.get());
+                cur_instr_decoded.reset(nullptr);
+                instr_free(GLOBAL_DCONTEXT, shard->prev_instr_decoded_.get());
```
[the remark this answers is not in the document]
bete0: Ack. Removed it.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -462,13 +463,73 @@ check_function_markers()
     return true;
 }
 
+bool
+check_repeated_syscall_with_same_pc()
+{
+    constexpr addr_t ADDR_ONE = 0x7fcf3b9dd9e9;
```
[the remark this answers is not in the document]
bete0: Done.

bete0:
run arm tests

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -462,13 +463,72 @@ check_function_markers()
...
+            gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
+            gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
+            gen_instr_encoded(ADDR,
+                              0xd4000001), // 0x7fcf3b9dd9e9: 0xd4000001 svc #0x0
+#    else
+        // TODO i#5871: Add AArch32 (and RISC-V) encodings.
+#    endif
+        };
+        if (!run_checker(memrefs, true, 1, 5, "Duplicate syscall instrs with the same PC",
+                         "Failed to catch double syscall instrs with the same PC"))
```
abhinav92003, line 492: s/double/duplicate/

clients/drcachesim/tests/memref_gen.h
```diff
@@ -84,7 +93,18 @@ gen_branch_encoded(memref_tid_t tid, addr_t pc, int encoding)
     memref.instr.encoding_is_new = true;
     return memref;
 }
+
 #elif defined(X86_64) || defined(X86_32)
+inline memref_t
+gen_instr_encoded(addr_t pc, const std::vector<char> &encoding, memref_tid_t tid = 1)
+{
+    memref_t memref = gen_instr_type(TRACE_TYPE_INSTR, tid, pc, encoding.size());
+    memref.instr.size = encoding.size();
```
abhinav92003, line 102: Redundant assignment. It's already set in gen_instr_type.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -596,46 +612,38 @@ invariant_checker_t::print_results()
...
     bool have_cond_branch_target = false;
     addr_t cond_branch_target = 0;
-
     if (shard->prev_instr_.instr.addr != 0 /*first*/ &&
-        type_is_instr_direct_branch(shard->prev_instr_.instr.type) &&
         // We do not bother to support legacy traces without encodings.
-        TESTANY(OFFLINE_FILE_TYPE_ENCODINGS, shard->file_type_)) {
-        addr_t trace_pc = shard->prev_instr_.instr.addr;
+        expect_encoding && type_is_instr_direct_branch(shard->prev_instr_.instr.type)) {
         if (shard->prev_instr_.instr.encoding_is_new)
```
abhinav92003, line 625: nit: can declare a local var `prev_instr_trace_pc` for shard->prev_instr_.instr.addr since it's used multiple times here.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -678,6 +686,13 @@ invariant_checker_t::check_for_pc_discontinuity(per_shard_t *shard,
                 if (!valid_branch_flow) {
                     error_msg = "Direct branch does not go to the correct target";
                 }
+            } else if (cur_instr_decoded != nullptr &&
+                       shard->prev_instr_decoded_ != nullptr &&
+                       shard->prev_instr_.instr.addr != 0 /*first*/ &&
```
abhinav92003, line 691: We already check `shard->prev_instr_.instr.addr != 0`  in the if-condition at L647, so this condition is not needed again here.

abhinav92003 (approved):
Just a few small nits.

clients/drcachesim/tests/invariant_checker_test.cpp
bete0: Done.

clients/drcachesim/tests/memref_gen.h
bete0: Done.

clients/drcachesim/tools/invariant_checker.cpp
bete0: Done.

clients/drcachesim/tools/invariant_checker.cpp
bete0: Done.

derekbruening:
Reminder to hold off merging until #5971 fixes the instability

bete0:
run arm tests

bete0:
run arm tests

PR #5937, opened 2023-03-29
i#5724 chunk bounds: Fix incorrect instruction counts

derekbruening:
Fixes a too-long chunk after a prior chunk splits up consecutive delayed branches in the same buffer.

Avoids placing wasted encodings at the end of a chunk.

Adds a unit test, and reduces the chunk size for the larger test which causes the original problem to reproduce much more reliably.

Still to fix is to re-emit encodings for the second consecutive branch after the boundary.

Issue: #5724

### DIFF
--- a/clients/drcachesim/tests/raw2trace_unit_tests.cpp
+++ b/clients/drcachesim/tests/raw2trace_unit_tests.cpp
@@ -106,6 +106,39 @@ public:
             std::unique_ptr<module_mapper_t>(new module_mapper_test_t(instrs, drcontext));
         set_modmap_(module_mapper_.get());
     }
+    raw2trace_test_t(const std::vector<std::istream *> &input,
+                     const std::vector<archive_ostream_t *> &output, instrlist_t &instrs,
+                     void *drcontext, uint64_t chunk_instr_count = 10 * 1000 * 1000)
+        : raw2trace_t(nullptr, input, {}, output, INVALID_FILE, nullptr, nullptr,
+                      drcontext,
+                      // The sequences are small so we print everything for easier
+                      // debugging and viewing of what's going on.
+                      4, /*worker_count=*/-1, /*alt_module_dir=*/"", chunk_instr_count)
+    {
+        module_mapper_ =
+            std::unique_ptr<module_mapper_t>(new module_mapper_test_t(instrs, drcontext));
+        set_modmap_(module_mapper_.get());
+    }
+};
+
+class archive_ostream_test_t : public archive_ostream_t {
+public:
+    archive_ostream_test_t()
+        : archive_ostream_t(new std::basic_stringbuf<char, std::char_traits<char>>)
+    {
+    }
+    std::string
+    open_new_component(const std::string &name) override
+    {
+        return "";
+    }
+    std::string
+    str()
+    {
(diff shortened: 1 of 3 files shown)


derekbruening:
Something wrong with the CI triggers?  They're all queued...

clients/drcachesim/tests/raw2trace_unit_tests.cpp
```diff
@@ -619,13 +652,145 @@ test_marker_delays(void *drcontext)
...
+        check_entry(entries, idx, TRACE_TYPE_ENCODING, -1) &&
+        check_entry(entries, idx, TRACE_TYPE_INSTR, -1) &&
+        check_entry(entries, idx, TRACE_TYPE_ENCODING, -1) &&
+#ifdef X86_32
+        // An extra encoding entry is needed.
+        check_entry(entries, idx, TRACE_TYPE_ENCODING, -1) &&
+#endif
+        check_entry(entries, idx, TRACE_TYPE_INSTR_DIRECT_JUMP, -1) &&
+        // Chunk should split the two jumps.
+        check_entry(entries, idx, TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_CHUNK_FOOTER) &&
```
abhinav92003, line 762: Where was this marker placed before this fix?

abhinav92003 (approved):

clients/drcachesim/tests/raw2trace_unit_tests.cpp
```diff
@@ -106,6 +106,39 @@ class raw2trace_test_t : public raw2trace_t {
             std::unique_ptr<module_mapper_t>(new module_mapper_test_t(instrs, drcontext));
         set_modmap_(module_mapper_.get());
     }
+    raw2trace_test_t(const std::vector<std::istream *> &input,
+                     const std::vector<archive_ostream_t *> &output, instrlist_t &instrs,
+                     void *drcontext, uint64_t chunk_instr_count = 10 * 1000 * 1000)
```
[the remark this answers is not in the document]
derekbruening: This is following the order of its parent class raw2trace_t (which has a ton of parameters: maybe should be redone as a struct or sthg at some point).

clients/drcachesim/tests/raw2trace_unit_tests.cpp
derekbruening: It's the next chunk which is incorrect without the fix: it has 3 instrs:
```
14 type: 28 size: 21 val: 0
15 type: 28 size: 22 val: 10
16 type: 28 size: 2 val: 4
17 type: 28 size: 3 val: 0
18 type: 11 size: 5 val: 9
19 type: 47 size: 3 val: 12749640
20 type: 10 size: 3 val: 14
21 type: 47 size: 3 val: 12749640
22 type: 10 size: 3 val: 17      <---- incorrect 3rd instr without fix
23 type: 28 size: 21 val: 1
24 type: 28 size: 22 val: 13
```

abhinav92003: So the issue shows up when the first instruction of the chunk is an instr (and was not properly counted), and that happens particularly when the last chunk ends at a delayed branch?

PR #5945, opened 2023-04-01
i#5934 dup syscall: Skip extraneous syscall in raw2trace.

abhinav92003:
Skips emitting to the final trace the 1-instr duplicate system call block which
has the same pc as the immediately previous system call instr. This will serve
as a workaround until we root-cause and fix the underlying issue in the
drmemtrace tracer.

Adds a unit test for this scenario.

Tested on a proprietary app trace known to have duplicate syscall errors that
this fixes all instances of such errors.

Issue: #5934

### DIFF
--- a/clients/drcachesim/tests/raw2trace_unit_tests.cpp
+++ b/clients/drcachesim/tests/raw2trace_unit_tests.cpp
@@ -936,7 +936,7 @@ test_duplicate_syscalls(void *drcontext)
     instr_t *move1 =
         XINST_CREATE_move(drcontext, opnd_create_reg(REG1), opnd_create_reg(REG2));
     // XXX: Adding an XINST_CREATE_syscall macro will simplify this but there are
-    // complexities (xref create_syscall_instr());
+    // complexities (xref create_syscall_instr()).
 #ifdef X86
     instr_t *sys = INSTR_CREATE_syscall(drcontext);
 #elif defined(AARCHXX)
--- a/clients/drcachesim/tests/raw2trace_unit_tests.cpp
+++ b/clients/drcachesim/tests/raw2trace_unit_tests.cpp
@@ -935,12 +935,16 @@ test_duplicate_syscalls(void *drcontext)
     instr_t *nop = XINST_CREATE_nop(drcontext);
     instr_t *move1 =
         XINST_CREATE_move(drcontext, opnd_create_reg(REG1), opnd_create_reg(REG2));
+    // XXX: Adding an XINST_CREATE_syscall macro will simplify this but there are
+    // complexities (xref create_syscall_instr());
 #ifdef X86
     instr_t *sys = INSTR_CREATE_syscall(drcontext);
 #elif defined(AARCHXX)
     instr_t *sys = INSTR_CREATE_svc(drcontext, opnd_create_immed_int((sbyte)0x0, OPSZ_1));
-#else
+#elif defined(RISCV64)
     instr_t *sys = INSTR_CREATE_ecall(drcontext);
+#else
+    #error Unsupported architecture.
 #endif
     instr_t *move2 =
         XINST_CREATE_move(drcontext, opnd_create_reg(REG2), opnd_create_reg(REG1));
--- a/clients/drcachesim/tests/raw2trace_unit_tests.cpp
+++ b/clients/drcachesim/tests/raw2trace_unit_tests.cpp
@@ -1025,17 +1025,9 @@ test_duplicate_syscalls(void *drcontext)
         check_entry(entries, idx, TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_CPU_ID) &&
         // The move1 instr.
         check_entry(entries, idx, TRACE_TYPE_ENCODING, -1) &&
-#ifdef X86_32
-        // An extra encoding entry is needed.
-        check_entry(entries, idx, TRACE_TYPE_ENCODING, -1) &&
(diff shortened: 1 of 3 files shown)


clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1960,13 +1985,15 @@ class raw2trace_t : public trace_converter_t<raw2trace_t> {
...
         uint64 count_elided = 0;
+        uint64 count_duplicate_syscall = 0;
 
         uint64 cur_chunk_instr_count = 0;
         uint64 cur_chunk_ref_count = 0;
         memref_counter_t memref_counter;
         uint64 chunk_count_ = 0;
         uint64 last_timestamp_ = 0;
         uint last_cpu_ = 0;
+        app_pc last_pc_if_syscall_;
```
derekbruening, line 1996: How about =0 here, matching the surrounding lines and avoiding line 1945

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1273,6 +1281,18 @@ template <typename T> class trace_converter_t {
...
             }
+            // TODO i#5934: This is a workaround for the trace invariant error triggered
+            // by consecutive duplicate system call instructions. Remove this when the
+            // error is fixed in the drmemtrace tracer.
+            if (instr->is_syscall() && impl()->get_last_pc_if_syscall(tls) == pc &&
+                instr_count == 1) {
+                impl()->add_to_statistic(tls, RAW2TRACE_STAT_DUPLICATE_SYSCALL, 1);
+                impl()->log(
+                    3, "Found block with duplicate system call instruction. Skipping.");
+                // Since this instr is in its own block, we're done.
```
derekbruening, line 1292: So the trace will have a timestamp+cpuid with nothing after it before the next timestamp?  I guess there's no invariant on that currently; maybe worth a comment that it's left in a slightly strange state.

clients/drcachesim/tests/raw2trace_unit_tests.cpp
```diff
@@ -925,14 +925,133 @@ test_chunk_encodings(void *drcontext)
...
+    instrlist_t *ilist = instrlist_create(drcontext);
+    // raw2trace doesn't like offsets of 0 so we shift with a nop.
+    instr_t *nop = XINST_CREATE_nop(drcontext);
+    instr_t *move1 =
+        XINST_CREATE_move(drcontext, opnd_create_reg(REG1), opnd_create_reg(REG2));
+#ifdef X86
+    instr_t *sys = INSTR_CREATE_syscall(drcontext);
+#elif defined(AARCHXX)
+    instr_t *sys = INSTR_CREATE_svc(drcontext, opnd_create_immed_int((sbyte)0x0, OPSZ_1));
+#else
```
derekbruening, line 942: Safer to explicitly have RISCV64 and then #error

derekbruening (approved):

clients/drcachesim/tracer/raw2trace.h
abhinav92003: Done.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1293,6 +1313,10 @@ template <typename T> class trace_converter_t {
                 }
             } else
                 impl()->set_prev_instr_rep_string(tls, false);
+            if (instr->is_syscall())
+                impl()->set_last_pc_if_syscall(tls, pc);
```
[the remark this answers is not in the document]
abhinav92003: `pc` may be different than `decode_pc` after the get_instr_summary call, so cannot get rid of it. Replaced it with `orig_pc` here.

clients/drcachesim/tests/raw2trace_unit_tests.cpp
abhinav92003: Done.

clients/drcachesim/tests/raw2trace_unit_tests.cpp
```diff
@@ -925,14 +925,133 @@ test_chunk_encodings(void *drcontext)
...
+    // a buffer for decoding by raw2trace.
+    instrlist_t *ilist = instrlist_create(drcontext);
+    // raw2trace doesn't like offsets of 0 so we shift with a nop.
+    instr_t *nop = XINST_CREATE_nop(drcontext);
+    instr_t *move1 =
+        XINST_CREATE_move(drcontext, opnd_create_reg(REG1), opnd_create_reg(REG2));
+#ifdef X86
+    instr_t *sys = INSTR_CREATE_syscall(drcontext);
+#elif defined(AARCHXX)
+    instr_t *sys = INSTR_CREATE_svc(drcontext, opnd_create_immed_int((sbyte)0x0, OPSZ_1));
```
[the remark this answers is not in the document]
abhinav92003: Done.

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -1032,6 +1036,8 @@ instr_summary_t::construct(void *dcontext, app_pc block_start, INOUT app_pc *pc,
         desc->packed_ |= kIsFlushMask;
     if (instr_is_cti(instr))
         desc->packed_ |= kIsCtiMask;
+    if (instr_is_syscall(instr))
```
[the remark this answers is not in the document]
abhinav92003: Done.

Issue #3385, opened 2019-02-21
ASSERT_CURIOSITY "crashed while walking dynamic header" on AArch64


AssadHashmi:
This issue has been raised as a consequence of https://groups.google.com/forum/#!topic/dynamorio-users/UTWiYoc9TvA

The library load which exposes this failure in the ELF loader is libarmflang.so shipped as part of the Arm HPC compiler 19.0. I think it may be related to https://github.com/DynamoRIO/dynamorio/issues/1589 as the library is quite large (5.2M).

A SIGBUS (not SIGSEGV!) is caused by the strlen() call in the soname check in core/unix/module_elf.c:
```
                /* test string readability while still in try/except
                 * in case we screwed up somewhere or module is
                 * malformed/only partially mapped */
                if (*soname != NULL && strlen(*soname) == -1) {
                    ASSERT_NOT_REACHED();
                }
```
It was reproduced using a simple Fortran test case built with armflang:
```
$ cat hw.f
       program hello
          print *, "Hello World!"
       end program hello
$
$ armflang hw.f -o hw
$
$ ./hw
 Hello World!
$
```
ldd shows which libraries are linked:
```
$ ldd hw
        linux-vdso.so.1 (0x0000ffffb0649000)
        libarmflang.so => /opt/arm/arm-hpc-compiler-19.0_Generic-AArch64_SUSE-12_aarch64-linux/lib/libarmflang.so (0x0000ffffb0243000)
        libomp.so => /opt/arm/arm-hpc-compiler-19.0_Generic-AArch64_SUSE-12_aarch64-linux/lib/libomp.so (0x0000ffffb0165000)
        libm.so.6 => /lib64/libm.so.6 (0x0000ffffb0092000)
        librt.so.1 => /lib64/librt.so.1 (0x0000ffffb0071000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x0000ffffb003c000)
        libc.so.6 => /lib64/libc.so.6 (0x0000ffffafec6000)
        libstdc++.so.6 => /opt/arm/gcc-8.2.0_Generic-AArch64_SUSE-12_aarch64-linux/lib64/libstdc++.so.6 (0x0000ffffafccd000)
        libgcc_s.so.1 => /opt/arm/gcc-8.2.0_Generic-AArch64_SUSE-12_aarch64-linux/lib64/libgcc_s.so.1 (0x0000ffffafc9c000)
[...]

PR #5947, opened 2023-04-02
i#5946: Check whether dynamic info is set in os_module_data_t before using it

### LINKED ISSUE #5946: DynamoRIO crash on dr_get_proc_address() / dr_get_proc_address_ex()
System: ubuntu 20.04
DynamoRIO version: 9.91.19442, but it is reproduced on the latest master.

DynamoRIO crashes on dr_get_proc_address() call during module load event handling of shared library loading.
Crash occurs during initializing dynamic section in os_module_data_t structure. The library itself have this segments:
image

When I made target program to use standard system library (at /usr/lib/x86_64-linux-gnu/) it DynamoRIO works correctly, this library segments are:
image

Debug build of DynamoRIO shows that crash is caused by null bitmask:  
image

Also there are some `<CURIOSITY : (0) && "crashed while walking dynamic header" in file /home/kutz/workspace/sydr/dynamorio/core/unix/module_elf.c line 323` on loading such shared libraries.

This bug is similar to the #3385 and #3525.

apach301:
This PR tries to fix #5946 by performing extra check for dynamic info was initialized in os_module_data_t during module load.

The error is triggered when loaded ELF module have a specific program sections order: `.dynstr`  should be placed in the end. In such case it wouldn't be mapped at the moment of os_module_data_t initialization.
If DR-client would try to use module data later (via dr_get_proc_address for example), DR will crash.

_How to reproduce for testing:_
1) You need a target program with dynamic linking to the shared library (example: )
2) Shared library ELF should have late .dynstr section. This could be achieved by using `patchelf` utility: after a compilation of libtest.so object you need to run `patchelf --set-rpath '$ORIGIN' libtest.so`. To check program sections displacement run `readelf -l libtest.so`
3) The crash is triggered when DR client uses uninitialized os_module_data_t. The easiest way is to call dr_get_proc_address() in module_load_event handler.

Test example with compiled binaries and source code: [example.tar.gz](https://github.com/DynamoRIO/dynamorio/files/11200545/example.tar.gz).
Tested on ubuntu20.04/ubuntu22.04 with clang-14 and patchelf >=0.10

### DIFF
--- a/core/lib/instrument.c
+++ b/core/lib/instrument.c
@@ -1955,6 +1955,14 @@ instrument_module_load_trigger(app_pc pc)
         if (ma != NULL && !TEST(MODULE_LOAD_EVENT, ma->flags)) {
             /* switch to write lock */
             os_get_module_info_unlock();
+
+            /* i#3385: re-initialize dynamic information, because it failed
+             * during the first flat-mmap that loaded the module.
+             */
+            if (!ma->os_data.have_dynamic_info) {
+                os_module_update_dynamic_info(ma->start, ma->end - ma->start, false);
+            }
+
             os_get_module_info_write_lock();
             ma = module_pc_lookup(pc);
             if (ma != NULL && !TEST(MODULE_LOAD_EVENT, ma->flags)) {
--- a/core/lib/instrument.c
+++ b/core/lib/instrument.c
@@ -1958,7 +1958,7 @@ instrument_module_load_trigger(app_pc pc)
 #ifdef LINUX
             /* i#3385: re-try to initialize dynamic information, because
              * it failed during the first flat-mmap that loaded the module.
-             * We doesn't perfrom this if there are no clients, assuming
+             * We don't perform this if there are no clients, assuming
              * DynamoRIO doesn't use os_module_data_t information itself.
              */
             if (!ma->os_data.have_dynamic_info) {
--- a/core/lib/instrument.c
+++ b/core/lib/instrument.c
@@ -1960,6 +1960,9 @@ instrument_module_load_trigger(app_pc pc)
              * it failed during the first flat-mmap that loaded the module.
              * We don't perform this if there are no clients, assuming
              * DynamoRIO doesn't use os_module_data_t information itself.
+             *
+             * XXX: add a regression test later. See PR #5947 for how to
+             * reproduce this situation.
              */
             if (!ma->os_data.have_dynamic_info) {
                 os_module_update_dynamic_info(ma->start, ma->end - ma->start, false);
--- a/core/lib/instrument.c
+++ b/core/lib/instrument.c
@@ -1955,7 +1955,7 @@ instrument_module_load_trigger(app_pc pc)
         if (ma != NULL && !TEST(MODULE_LOAD_EVENT, ma->flags)) {
             /* switch to write lock */
             os_get_module_info_unlock();
-
+#ifdef LINUX
             /* i#3385: re-try to initialize dynamic information, because
              * it failed during the first flat-mmap that loaded the module.
              * We doesn't perfrom this if there are no clients, assuming
@@ -1964,7 +1964,7 @@ instrument_module_load_trigger(app_pc pc)
             if (!ma->os_data.have_dynamic_info) {
                 os_module_update_dynamic_info(ma->start, ma->end - ma->start, false);
             }
-
+#endif
             os_get_module_info_write_lock();
             ma = module_pc_lookup(pc);
             if (ma != NULL && !TEST(MODULE_LOAD_EVENT, ma->flags)) {
--- a/core/lib/instrument.c
+++ b/core/lib/instrument.c
@@ -1956,8 +1956,10 @@ instrument_module_load_trigger(app_pc pc)
             /* switch to write lock */
             os_get_module_info_unlock();
 
-            /* i#3385: re-initialize dynamic information, because it failed
-             * during the first flat-mmap that loaded the module.
+            /* i#3385: re-try to initialize dynamic information, because
+             * it failed during the first flat-mmap that loaded the module.
+             * We doesn't perfrom this if there are no clients, assuming
+             * DynamoRIO doesn't use os_module_data_t information itself.
              */
             if (!ma->os_data.have_dynamic_info) {
                 os_module_update_dynamic_info(ma->start, ma->end - ma->start, false);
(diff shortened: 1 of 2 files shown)


derekbruening (commented):
Thank you for submitting a PR.  Just trying to understand the issue first though.

core/unix/module_elf.c
```diff
@@ -756,6 +756,16 @@ get_proc_address_ex(module_base_t lib, const char *name, bool *is_indirect_code
     os_get_module_info_lock();
     ma = module_pc_lookup((app_pc)lib);
     if (ma != NULL) {
+        /* Check whether os_module_data_t was initialized during
```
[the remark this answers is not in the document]
apach301: Described problem in details in #5946.
Apparently this is the same issue as #3385: dynamic section not initialized because of failed soname check during first flat-mmap point. The problem occurs when loading libraries with .dynstr is in its last +rw segment.

core/lib/instrument.c
```diff
@@ -1955,6 +1955,14 @@ instrument_module_load_trigger(app_pc pc)
         if (ma != NULL && !TEST(MODULE_LOAD_EVENT, ma->flags)) {
             /* switch to write lock */
             os_get_module_info_unlock();
+
+            /* i#3385: re-initialize dynamic information, because it failed
+             * during the first flat-mmap that loaded the module.
+             */
+            if (!ma->os_data.have_dynamic_info) {
+                os_module_update_dynamic_info(ma->start, ma->end - ma->start, false);
```
derekbruening, line 1963: OK, this seems reasonable: at first execution if we don't have the data we try again.  And we'll get this far because the faults on the first instance are all inside try..excepts?

Should we leave the assert curiosities in place?  Probably.

And the SIGBUS you reported is from inside the debugger and it's caught by the try..except and is not fatal, right?

This code is currently wrapped in CLIENTS_EXIST(): do we need this when there are no clients?  Or we assume DR itself doesn't care about this module data?  I assume the latter: maybe worth adding a comment that we deliberately do not bother to re-try if there are no clients.

Another request is to go add a comment by that fault point in that try..except and point to i#3385 and explain that we'll try again in instrument_module_load_trigger() at first execution.

core/lib/instrument.c
```diff
@@ -1955,6 +1955,14 @@ instrument_module_load_trigger(app_pc pc)
...
+            }
```
derekbruening, line 1964: Would it be feasible to add a test for this?  By synthesizing a tiny ELF file with the right characteristics?  If you're not familiar with tooling to generate such a file, that's ok, we can punt on it: but please add a statement to the PR description explaining how you tested this, and maybe add a XXX comment here about possibly adding a regression test later.

derekbruening (commented):
Fix itself looks good, just had some requests to add some comments.

core/lib/instrument.c
apach301: > OK, this seems reasonable: at first execution if we don't have the data we try again. And we'll get this far because the faults on the first instance are all inside try..excepts?

Yes, module parsing is implemented inside try..excepts, so all bugs with unexpected/alternatively formed executables won't crash DR.

> And the SIGBUS you reported is from inside the debugger and it's caught by the try..except and is not fatal, right?

It is debug output for failed test string readablity in parsing header wrapped in try..except. 

> This code is currently wrapped in CLIENTS_EXIST(): do we need this when there are no clients?

I don't know whether DR uses os_nodule_data_t itself. When I first encountered this error, first of all I tried to launch DR with dummy/empty-clients/without-client, and it didn't crashed. So I guess it can be left only for client-usecases.

core/lib/instrument.c
apach301: I'm not able to create a single tiny ELF file with late .dynstr section. I encountered this problem when testing oss-fuzz program: they applied `patchelf --set-rpath` command to shared libs. As a result their dynstr was relocated to the end.

I succeed to reproduce it on a small test-program by making it use a patched libc.so.6 library, but it weights 2.0M. I'm not familiar with DR test system, but to trigger bug a DR client is required that is perform some logic on loaded modules

core/lib/instrument.c
apach301: Updated comments

derekbruening:
> @[apach301](https://github.com/apach301) apach301 [force-pushed](https://github.com/DynamoRIO/dynamorio/compare/2b76f87909ff7cdfab3ca7c353a0bd4916b747bc..9cec9bfbcf2abd38ee08d373540c576ebaa16064) the fix-get-proc-address branch from [2b76f87 ](https://github.com/DynamoRIO/dynamorio/commit/2b76f87909ff7cdfab3ca7c353a0bd4916b747bc)to [9cec9bf ](https://github.com/DynamoRIO/dynamorio/commit/9cec9bfbcf2abd38ee08d373540c576ebaa16064)
[1 hour ago](https://github.com/DynamoRIO/dynamorio/pull/5947#event-8946059096)

Please don't force-push a PR branch: it messes up the review process (there is no "View changes since last review" now; prior comments lose their anchor points; etc.).  We do a squash when we merge so all the commits will be collapsed anyway.

core/lib/instrument.c
derekbruening: There are many test clients that register for module events (see `git grep _register_module_load suite`).  To avoid making a new one you could use `suite/tests/client-interface/events.dll.c`.

So `patchelf --set-rpath` didn't work on a tiny ELF file, only on the larger libc?  2.0M we could live with though.

For how to add the test: an example of a new app that uses an existing already-built events client is the fibers test:
```
  add_exe(client.fibers client-interface/fibers.c)
  torunonly_ci(client.fibers client.fibers client.events_cpp.dll
    client-interface/fibers.c "" "" "")
```

core/lib/instrument.c
```diff
@@ -1955,6 +1955,16 @@ instrument_module_load_trigger(app_pc pc)
         if (ma != NULL && !TEST(MODULE_LOAD_EVENT, ma->flags)) {
             /* switch to write lock */
             os_get_module_info_unlock();
+#ifdef LINUX
+            /* i#3385: re-try to initialize dynamic information, because
+             * it failed during the first flat-mmap that loaded the module.
+             * We doesn't perfrom this if there are no clients, assuming
```
derekbruening, line 1961: grammar nit: s/doesn't/don't/

core/lib/instrument.c
apach301: Couldn't manage how to properly add a test. Provided a detailed description in this PR, attached compiled binaries and test sources, left a XXX comment in DR code.

derekbruening:
run arm tests

core/lib/instrument.c
derekbruening: OK, though unfortunately it is likely no one will have resources later to add the test, we will go with this to get the fix in.

derekbruening (approved):

derekbruening:
Thank you for the root cause analysis and the patch!

Issue #2071, opened 2016-11-16
CRASH in ARM signal_racesys test


derekbruening:
http://dynamorio.org/CDash/testDetails.php?test=153499&build=14701

Test: code_api|linux.signal_racesys (Failed) 
Build: release-external-32 (ARM.Ubuntu1404.gcc484.CAM) on 2016-11-16 05:16:41
Repository revision: c3bd1ca94176c56dff60880ffff3d151061a21ff

```
Exit Value	139
Fail Reason	Required regular expression not found.Regex=[^all done
$
]

Test output
```

Looks like a SIGSEGV not reported by DR?  That would have to be early, or
late after the "all done" which is not printed...unclear how this can happen.

derekbruening:
We're seeing this test fail on x86 32-bit in a similar way, non-deterministically.  It happened on GA CI Ubuntu18.04:

```
76: Test command: /home/runner/work/dynamorio/dynamorio/build_debug-internal-32/bin32/drrun "-s" "90" "-quiet" "-debug" "-killpg" "-stderr_mask" "0xC" "-dumpcore_mask" "0" "-code_api" "--" "/home/runner/work/dynamorio/dynamorio/build_debug-internal-32/suite/tests/bin/linux.signal_racesys"
76: Test timeout computed to be: 600
 77/349 Test  #76: code_api|linux.signal_racesys ....................................***Failed  Required regular expression not found. Regex=[^all done
$
]  0.51 sec

test 77
```

And on my laptop it fails every 200 or 300 runs:
```
$ ninja && ctest --repeat-until-fail 1000 -R racesys
...
    Test #76: code_api|linux.signal_racesys ....   Passed    0.42 sec
    Start 76: code_api|linux.signal_racesys
    Test #76: code_api|linux.signal_racesys ....   Passed    0.42 sec
    Start 76: code_api|linux.signal_racesys
    Test #76: code_api|linux.signal_racesys ....***Failed  Required regular expression not found. Regex=[^all done
$
]  0.33 sec

$ grep -c Testing: Testing/Temporary/LastTest.log  
182
```

PR #5952, opened 2023-04-03
i#5950: Stop skipping at footer

### LINKED ISSUE #5950: ASSERT drmemtrace skip_instrs overshoot in linear walk
The drmemtrace skip_instrs feature correctly returns EOF when given a too-far target that is hit in the chunk loop: but when the linear loop is reached, it has no check for EOF and instead we hit an assert:

```
Unknown trace entry type footer (26)
drcachesim: /usr/local/google/home/bruening/dr/git/src/clients/drcachesim/reader/reader.cpp:312: virtual bool reader_t::process_input_entry(): Assertion `false' failed.
Aborted
```

derekbruening:
Adds an explicit check for -skip_instructions hitting the footer to avoid an assert on trying to process it as a normal entry.

Adds a scheduler test.

Adds a separate analyzer error string for this.
I tested this manually:
    -------------
    $ clients/bin64/drcachesim -simulator_type view -indir drmemtrace.*.dir -skip_instrs 6300 -verbose 2
    ERROR: failed to run analyzer: Too-far -skip_instrs for: drmemtrace.allasm_x86_64.2490594.5358.trace.zip
    -------------

Fixes #5950

### DIFF
--- a/clients/drcachesim/tests/scheduler_unit_tests.cpp
+++ b/clients/drcachesim/tests/scheduler_unit_tests.cpp
@@ -592,7 +592,8 @@ test_regions_too_far()
                          std::unique_ptr<mock_reader_t>(new mock_reader_t()), 1);
 
     std::vector<scheduler_t::range_t> regions;
-    regions.emplace_back(4, 0);
+    // Start beyond the last instruction.
+    regions.emplace_back(3, 0);
 
     scheduler_t scheduler;
     std::vector<scheduler_t::input_workload_t> sched_inputs;
--- a/clients/drcachesim/tests/scheduler_unit_tests.cpp
+++ b/clients/drcachesim/tests/scheduler_unit_tests.cpp
@@ -101,6 +101,14 @@ make_exit(memref_tid_t tid)
 }
 
 static trace_entry_t
+make_footer()
+{
+    trace_entry_t entry;
+    entry.type = TRACE_TYPE_FOOTER;
+    return entry;
+}
+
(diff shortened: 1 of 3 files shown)


derekbruening:
x86-32 failure is api.startstop with vdso message: #4604

clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -562,12 +570,52 @@ test_regions_start()
...
+        make_exit(1),
+        make_footer(),
+        /* clang-format on */
+    };
+    std::vector<scheduler_t::input_reader_t> readers;
+    readers.emplace_back(std::unique_ptr<mock_reader_t>(new mock_reader_t(memrefs)),
+                         std::unique_ptr<mock_reader_t>(new mock_reader_t()), 1);
+
+    std::vector<scheduler_t::range_t> regions;
+    regions.emplace_back(4, 0);
```
abhinav92003, line 595: Would the test continue to work for s/4/3? Since the 3rd instr is the first instr to not exist, it tests the boundary condition.

abhinav92003 (approved):

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: Changed it to 3 (yes test still passes).

derekbruening:
x86-32 failure is signal_racesys which is #2071

PR #5958, opened 2023-04-04
Unique invariant_checker error message for branch target PC discontinuities

lihasgupta:
Currently in invariant_checker we get a generic error message for PC discontinuities: `Non-explicit control flow has no marker`. This is able to catch invariants but these types of invariants often have several different causes. One of these such causes is when a direct branch does not go to the correct target. 

This PR adds a unique error string for this specific kind of invariant: `Direct branch does not go to the correct target`.

### DIFF
--- a/clients/drcachesim/tools/invariant_checker.cpp
+++ b/clients/drcachesim/tools/invariant_checker.cpp
@@ -374,45 +374,10 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
         }
         // Invariant: non-explicit control flow (i.e., kernel-mediated) is indicated
         // by markers.
-        bool have_cond_branch_target = false;
-        addr_t cond_branch_target = 0;
-        if (shard->prev_instr_.instr.addr != 0 /*first*/ &&
-            type_is_instr_direct_branch(shard->prev_instr_.instr.type) &&
-            // We do not bother to support legacy traces without encodings.
-            TESTANY(OFFLINE_FILE_TYPE_ENCODINGS, shard->file_type_)) {
-            addr_t trace_pc = shard->prev_instr_.instr.addr;
-            if (shard->prev_instr_.instr.encoding_is_new)
-                shard->branch_target_cache.erase(trace_pc);
-            auto cached = shard->branch_target_cache.find(trace_pc);
-            if (cached != shard->branch_target_cache.end()) {
-                have_cond_branch_target = true;
-                cond_branch_target = cached->second;
-            } else {
-                instr_t instr;
-                instr_init(GLOBAL_DCONTEXT, &instr);
-                const app_pc decode_pc =
-                    const_cast<app_pc>(shard->prev_instr_.instr.encoding);
-                const app_pc next_pc =
-                    decode_from_copy(GLOBAL_DCONTEXT, decode_pc,
-                                     reinterpret_cast<app_pc>(trace_pc), &instr);
-                if (next_pc == nullptr || !opnd_is_pc(instr_get_target(&instr))) {
-                    // Neither condition should happen but they could on an invalid
-                    // encoding from raw2trace or the reader so we report an
-                    // invariant rather than asserting.
-                    report_if_false(shard, false, "Branch target is not decodeable");
-                } else {
-                    have_cond_branch_target = true;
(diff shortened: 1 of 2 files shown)


clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -410,40 +410,58 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
             }
         }
         if (shard->prev_instr_.instr.addr != 0 /*first*/) {
-            report_if_false(
-                shard, // Filtered.
+
```
derekbruening, line 413: I find that these blank lines disrupt the existing grouping: see how in this file each check has a comment and a block of code with no blank lines inside the block for that check but with blank lines separating each check (see e.g. line 258 and line 278) and that's how they're grouped to help in visually seeing each check.  This particular check has gotten very large though.  It would probably be best to move this all to a helper function, or maybe turn the header comment at line 375 into a banner (i.e., whole line of *'s) and have a blank line before it and another banner line at the bottom to better group this and then we can have separators inside (the function seems the better choice).

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -661,3 +629,61 @@ invariant_checker_t::print_results()
     std::cerr << "Trace invariant checks passed\n";
     return true;
 }
+
+std::string
```
derekbruening, line 633: Preserve the comment  "Invariant: non-explicit control flow (i.e., kernel-mediated) is indicated by markers." somewhere: right now there is no mention of markers (nor of "non-explicit") except suddenly in the error message.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -661,3 +629,61 @@ invariant_checker_t::print_results()
...
+invariant_checker_t::pc_discontinuity_error_msg(void *shard_data, const memref_t &memref,
```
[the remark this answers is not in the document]
derekbruening: The function name makes me think this takes in an enum of error types and turns it into a string, rather than identifying whether there's an error.  How about "check_for_pc_discontinuity()" or something else active.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -661,3 +629,61 @@ invariant_checker_t::print_results()
...
+            // mismatch.
+            if (type_is_instr_branch(shard->prev_instr_.instr.type)) {
+                error_msg =
+                    ( // Indirect branches we cannot check.
+                        !type_is_instr_direct_branch(shard->prev_instr_.instr.type) ||
+                        // Conditional fall-through hits the regular case above.
+                        !have_cond_branch_target ||
+                        memref.instr.addr == cond_branch_target)
+                    ? ""
+                    : "Direct branch does not go to the correct target";
```
derekbruening, line 681: I think it would be easier to read if this were just an if() setting the non-empty (don't need else b/c set to empty up front) instead of a ?: operator.

derekbruening (commented):
Just some nits/minor suggestions.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -410,8 +410,8 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
             }
         }
         // Check if we have any invariant violations caused by PC discontinuities.
```
derekbruening, line 412: Now that I look at the prior code and the `// Invariant: ` comment indicating what we're checking up on line 375: I would expect the whole decoding sequence and identifying the branch target to be in the helper function.  (There will be some conflict with Nahome's decoding changes, but based purely on this diff, all this decoding is used only for this invariant so I would expect it in the function for this invariant).

derekbruening (approved):

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -225,7 +225,8 @@ check_sane_control_flow()
             gen_instr(1, 20),
```
derekbruening, line 225: Please follow the commit description conventions:

+ `i#NNN: ` title prefix
+ `Issue: NNNN` line at the end of the body

PR #5961, opened 2023-04-06
i#4062: Remove trace_converter_t to simplify raw2trace_t

### LINKED ISSUE #4062: refactor raw2trace into a single implementation
#3129 refactored drcachesim's offline trace post-processing to allow for custom implementations.  However, after trace sharding #3230 we are no longer using the refactored interface.  It may be cleanest to undo the interface split and merge raw2trace back into a single implementation.

derekbruening:
Removes the separate template class trace_converter_t and moves its contents into the raw2trace_t class.  We no longer need this separation, and in fact it is hindering our efforts to add new capabilities to raw2trace_t.

Fixes #4062


abhinav92003 (approved):

derekbruening:
x86-32 failure is scattergather atomic write assert #5329

Issue #5953, opened 2023-04-03
rseq side exit in native run results in PC discontinuity in drmemtrace


derekbruening:
With our run-twice solution, drmemtrace always records the full rseq sequence, yet the native execution might take a side branch and not complete execution (this is distinct from an rseq abort due to a kernel event).  When that happens, the next block executed is not the logically next block after the rseq sequence.  This results in a PC discontinuity in the resulting trace.

This is compounded by the possibility of a signal arriving in between the side exit and next block.

This is our plan to solve this:

the client needs to know when a side exit from the native
rseq execution happened. This is a regular exit cti which will be linked up. We want to
keep overhead small: even if these are sort of rare it's still hot code in
general. Maybe DR can add one store to a TLS field on each side exit (via local control
flow during mangling) storing the branch's PC, and the client identifies all exits of
side exits (DR either passes all regions up front for each module and client maps to
blocks, or DR adds a label to the top of exit target blocks) and adds a check looking
for this field (which DR exports a query for) and adds a marker (tracer itself can't
roll back b/c for long rseq region could have already written out erroneous instrs)
whose value is the PC of the exit branch and raw2trace sees that and rolls back.

raw2trace now needs to be able to go backward multiple blocks; plus a signal could
arrive before the side exit target block. Should DR tell the tracer about all rseq
regions? A label for the start block? And the tracer adds a marker; then raw2trace
buffers all rseq blocks. Plus a label for the end too.

So 4 new labels added by DR:

1. Rseq region entry at top of block (DR arranges for start of block)
2. Rseq region end before committing store (to avoid hiccups with label as final instr in
   block): call it a committing store label. Tracer adds marker after store.
3. Rseq region side entry target block, at top of block
4. Rseq abort handler, at top of block

Tracer converts labels to markers for raw2trace.

New DR API:

Query side exit PC TLS field: clear field at this point.
Also instrumentation-level opnd API so can inline a check and then clean call to write
marker. Clean call required to call API query to clear value.

Raw2trace:

If see rseq entry marker, start buffering. Flush buffer at rseq end marker (moved to
after store by tracer) or side exit marker or abort handler marker. 
If see rseq side exit marker, roll back in buffer until marker value PC found.
If signal came in: raw2trace would not have flushed buffer yet: so this would naturally
handle the signal interruption.

derekbruening:
[...]

PR #5962, opened 2023-04-06
i#5953 rseq side: Share code in raw2trace_unit_tests

derekbruening:
Shares the chunk of code running raw2trace on a vector of offline entries and producing a vector of trace_entry_t, which was previously duplicated in every test.

Issue: #5953

### DIFF
--- a/clients/drcachesim/tests/raw2trace_unit_tests.cpp
+++ b/clients/drcachesim/tests/raw2trace_unit_tests.cpp
@@ -274,9 +274,72 @@ check_entry(std::vector<trace_entry_t> &entries, int &idx, unsigned short expect
     return true;
 }
 
+// Takes ownership of ilist and destroys it.
+bool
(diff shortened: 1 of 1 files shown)


derekbruening:
Failure is a flake: tool.drcacheoff.burst_threads_counts failed with flaky output #5963

clients/drcachesim/tests/raw2trace_unit_tests.cpp
```diff
@@ -274,9 +274,72 @@ check_entry(std::vector<trace_entry_t> &entries, int &idx, unsigned short expect
     return true;
 }
 
+// Takes ownership of ilist and destroys it.
+bool
+run_raw2trace(void *drcontext, std::vector<offline_entry_t> raw, instrlist_t *ilist,
```
abhinav92003, line 279: Add const for `raw`.

abhinav92003 (approved):

PR #5964, opened 2023-04-07
Move test_module_mapper_t to raw2trace.h for code sharing

abhinav92003:
Refactors test code to move common logic to raw2trace.h to enable easier
code sharing among tests.

### DIFF
--- a/clients/drcachesim/tests/raw2trace_unit_tests.cpp
+++ b/clients/drcachesim/tests/raw2trace_unit_tests.cpp
@@ -68,31 +68,6 @@
 #    error Unsupported arch
 #endif
 
-// Subclasses module_mapper_t and replaces the module loading with a
-// buffer of encoded instr_t.
-class module_mapper_test_t : public module_mapper_t {
-public:
-    module_mapper_test_t(instrlist_t &instrs, void *drcontext)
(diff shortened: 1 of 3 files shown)


clients/drcachesim/tests/raw2trace_unit_tests.cpp
```diff
@@ -105,8 +80,8 @@ class raw2trace_test_t : public raw2trace_t {
                       // debugging and viewing of what's going on.
                       4)
     {
-        module_mapper_ =
-            std::unique_ptr<module_mapper_t>(new module_mapper_test_t(instrs, drcontext));
+        module_mapper_ = std::unique_ptr<module_mapper_t>(
+            new test_module_mapper_t(&instrs, drcontext));
```
derekbruening, line 84: (It's weird that this re-formatted when the name is the same length.)

derekbruening (approved):

clients/drcachesim/tests/raw2trace_unit_tests.cpp
abhinav92003: `instrs` was changed to `&instrs` so the line length increased.

Issue #5912, opened 2023-03-14
PC discontinuity reported as incorrect signal handler return point by invariant checker


abhinav92003:
Today we do not check PC discontinuities between an instr followed by a "kernel xfer" marker. If there is a discontinuity, it is reported as a "signal handler return point incorrect" invariant error later when the signal returns. E.g. the following

```
    28156078    20615075:       79758 ifetch       5 byte(s) @ 0x000000001a593862 66 45 89 0c fc       data16 mov    %r9w -> (%r12,%rdi,8)[2byte]
    28156079    20615075:       79758 write        2 byte(s) @ 0x0000177219ac0060 by PC 0x000000001a593862

// PC discontinuity in the kernel xfer marker address and prev instr. Not reported by invariant checker as such.
    28156080    20615075:       79758 <marker: kernel xfer from 0x1a59386d to handler>
...
    28156767    20615515:       79758 <marker: syscall xfer from 0x7ff30d21c1c9>

// Invariant error reported here due to mismatch with the pre_kernel_xfer instr (even though it is same as the kernel xfer marker address).
    28156770    20615516:       79758 ifetch       3 byte(s) @ 0x000000001a59386d 48 31 c0             xor    %rax %rax -> %rax
```

Such errors should be reported as what they are to avoid confusion.

derekbruening:
Offline discussion concluded we want to do this:

A) Add our regular PC transition check to the pre-signal instruction fetch transition to the signal marker value for the interrupted PC

B) Change the signal check to compare the handler return point to the marker value, not to the pre-signal instruction fetch

Issue #5490, opened 2022-05-11
Add branch target address to drmemtrace traces for use without instruction encodings


derekbruening:
When a drmemtrace and associated application binaries are available to supply
the instruction encodings, the targets of non-taken branches are obtained
from the encodings (and targets of taken branches from the subsequent PC -- which
has complications across intervening kernel transfers, solved with special kernel
transfer fields, and across thread interleavings, solved by tweaking those in raw2trace for
offline at least).  However, the binaries not always available (e.g., the Google workload traces).

Adding the target of every conditional branch could be done in raw2trace. 
The main cost would be a size increase. If the average block is 5 instrs + 2 addrs then
this would be a 14% increase scaled down a little since not all branches
are conditional, but still a non-trivial increase of likely >10% by this
un-measured back-of-envelope estimate anyway.

Presumably we would implement this in post-processing, waiting for the next
block, rather than having online instrumentation deduce the target before
the branch occurs.

PR #5965, opened 2023-04-07
i#5912: Add caching and clean up PC discontinuity checks

lihasgupta:
Removes redundant `pre_signal_flow_continuity` check in the invariant_checker.

Adds unit tests to check for PC discontinuities in the transitions between instructions and kernel xfer markers

Adds caching around the decoding, to prevent repeated calls to the expensive `decode_from_copy()` function.

Refactors the invariant_checker to use `instr_info_t` structs which group together memref's and their corresponding decodings.

Fixes: #5912
Fixes: #6006

### DIFF
--- a/core/ir/riscv64/encode.c
+++ b/core/ir/riscv64/encode.c
@@ -102,6 +102,7 @@ instr_encode_arch(dcontext_t *dcontext, instr_t *instr, byte *copy_pc, byte *fin
 {
     decode_info_t di;
     uint enc;
+    int instr_length;
 
     if (has_instr_opnds != NULL) {
         *has_instr_opnds = false;
@@ -135,8 +136,15 @@ instr_encode_arch(dcontext_t *dcontext, instr_t *instr, byte *copy_pc, byte *fin
         });
         return NULL;
     }
-    *(uint *)copy_pc = enc;
-    return copy_pc + 4;
+    instr_length = instr_length_arch(dcontext, instr);
+    if (instr_length == RISCV64_INSTR_COMPRESSED_SIZE) {
+        *(ushort *)copy_pc = (ushort)enc;
+    } else {
+        ASSERT(instr_length == RISCV64_INSTR_SIZE);
+        *(uint *)copy_pc = enc;
+    }
+
+    return copy_pc + instr_length;
 }
 
 byte *


clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -514,6 +514,22 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+            // running online.
+            (knob_test_name_ != "kernel_xfer_app" || knob_offline_) &&
+            // All TRACE_MARKER_TYPE_RSEQ_ABORT markers are followd by
+            // TRACE_MARKER_TYPE_KERNEL_EVENT. We only want to consider the
+            // TRACE_MARKER_TYPE_KERNEL_EVENT markers that do not occur directly after
+            // an RSEQ abort marker.
+            !(shard->prev_entry_.data.type == TRACE_TYPE_MARKER &&
+              shard->prev_entry_.marker.marker_type == TRACE_MARKER_TYPE_RSEQ_ABORT)) {
+            const std::string pc_discontinuity_error_string =
+                check_for_pc_discontinuity(shard, memref, nullptr, false);
```
abhinav92003, line 529: Do we want a different error string for such signal marker discontinuities?

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -696,15 +720,17 @@ invariant_checker_t::check_for_pc_discontinuity(
             TESTANY(OFFLINE_FILE_TYPE_FILTERED | OFFLINE_FILE_TYPE_IFILTERED,
                     shard->file_type_) ||
             // Regular fall-through.
```
abhinav92003, line 722: FYI, there's an additional condition that needs to be added here to check PC discontinuity for signals. It was added by the recently submitted PR #5985 (see `shard->last_signal_context_.xfer_int_pc == 0`).

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003, line 529: The new checks for signals are also being done today as the "signal handler return point incorrect" check. Do we plan to remove the redundant checks?

abhinav92003 (changes requested):

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: Actually the `shard->last_signal_context_.xfer_int_pc == 0` check is not part of the continuity check, so that's not needed here. See other comment about the `pre_signal_flow_continuity` checks though.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -764,10 +781,12 @@ invariant_checker_t::check_for_pc_discontinuity(
...
-            (prev_instr_trace_pc + shard->prev_instr_.instr.size == memref.instr.addr) ||
+            (prev_instr_trace_pc + shard->prev_instr_.instr.size ==
+             current_memref_addr) ||
             // String loop.
-            (prev_instr_trace_pc == memref.instr.addr &&
-             (memref.instr.type == TRACE_TYPE_INSTR_NO_FETCH ||
+            (prev_instr_trace_pc == current_memref_addr &&
```
abhinav92003, line 787: There are a bunch of conditions related to signal continuity. See `pre_signal_flow_continuity`. Let's make sure all of them are reflected here.

https://github.com/DynamoRIO/dynamorio/blob/0e86f7948bd7a791c9f628e2a0932cc5c08de88b/clients/drcachesim/tools/invariant_checker.cpp#L455

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -275,28 +275,30 @@ check_sane_control_flow()
         if (!run_checker(memrefs, false))
             return false;
     }
-    // PC discontinuity from instr to kernel_xfer event marker value.
+    // Positive test (PC discontinuity): Transition from instr to kernel_xfer event
```
abhinav92003, line 280: IMHO the terminology used in existing tests is confusing: we use "negative" to mean presence of PC discontinuity. I think we should either change that as well, or just conform with the existing convention for now here too.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -824,7 +824,10 @@ invariant_checker_t::check_for_pc_discontinuity(
                        instr_is_syscall(shard->prev_instr_decoded_->data)) {
                 error_msg = "Duplicate syscall instrs with the same PC";
             } else {
-                error_msg = "Non-explicit control flow has no marker";
+                // If we reach this block, we know there is a PC discontinuity.
+                error_msg = memref_is_kernel_event_marker
```
abhinav92003, line 828: May be better to let the caller decide what error message to show.

abhinav92003 (changes requested):

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -793,19 +812,22 @@ invariant_checker_t::check_for_pc_discontinuity(
...
+                    !have_cond_branch_target || current_memref_addr == cond_branch_target;
 
                 if (!valid_branch_flow) {
                     error_msg = "Direct branch does not go to the correct target";
                 }
             } else if (cur_instr_decoded != nullptr &&
                        shard->prev_instr_decoded_ != nullptr &&
                        instr_is_syscall(cur_instr_decoded->data) &&
-                       memref.instr.addr == prev_instr_trace_pc &&
+                       current_memref_addr == prev_instr_trace_pc &&
```
abhinav92003, line 823: For signals, we should be comparing `last_instr_in_cur_context_.instr.addr` (not prev_instr_.instr.addr) with current_memref_addr.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -871,10 +896,11 @@ invariant_checker_t::check_for_pc_discontinuity(
             TESTANY(OFFLINE_FILE_TYPE_FILTERED | OFFLINE_FILE_TYPE_IFILTERED,
                     shard->file_type_) ||
             // Regular fall-through.
-            (prev_instr_trace_pc + shard->prev_instr_.instr.size == memref.instr.addr) ||
+            (prev_instr.instr.addr + prev_instr.instr.size == current_memref_addr) ||
             // String loop.
-            (prev_instr_trace_pc == memref.instr.addr &&
-             (memref.instr.type == TRACE_TYPE_INSTR_NO_FETCH ||
+            (prev_instr.instr.addr == current_memref_addr &&
+             (memref_is_kernel_event_marker ||
```
abhinav92003, line 902: As discussed offline, we should not try to add to this "string loop" condition. There are other parts to this condition that apply only to the string loop case

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -837,13 +849,26 @@ invariant_checker_t::check_for_pc_discontinuity(
...
+#ifdef UNIX
+    // Identify whether the current memref is a marker instead of an instruction. This
+    // will be handled differently.
+    if (memref.marker.type == TRACE_TYPE_MARKER &&
+        memref.marker.marker_type == TRACE_MARKER_TYPE_KERNEL_EVENT) {
+        memref_is_kernel_event_marker = true;
+        prev_instr = shard->last_instr_in_cur_context_;
+    }
+#endif
+    const addr_t current_memref_addr =
```
abhinav92003, line 863: (Optional, but recommended) Better to avoid this alias. It makes it a little harder to understand the conditions below. If we use either "memref.marker.marker_value" or "memref.instr.addr" directly, it's easily apparent what the condition expects the memref to be (a marker or an instr, respectively).

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -303,6 +303,73 @@ check_sane_control_flow()
...
+        };
+        if (!run_checker(memrefs, false)) {
+            return false;
+        }
+    }
+    // Positive test: We should skip the check if there is no instruction before the
+    // kernel event. i.e. The previous instr address is zero.
+    {
+        std::vector<memref_t> memrefs = {
+            gen_instr(1, 0),
```
abhinav92003, line 339: If you're checking for the case where there's no prev instr, then this get_instr shouldn't be there. The zero is an internal default, so better to not create an actual instr with that pc.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -303,6 +303,73 @@ check_sane_control_flow()
...
+            gen_instr(1, 101),
+            gen_marker(1, TRACE_MARKER_TYPE_KERNEL_XFER, 102),
+            gen_instr(1, 2),
+        };
+        if (!run_checker(memrefs, false)) {
+            return false;
+        }
+    }
+#endif
+
```
abhinav92003, line 372: Can we also add tests when the last instr before the signal is a branch? And for when there was an rseq abort in the last signal context?

abhinav92003 (changes requested):

abhinav92003 (changes requested):

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -837,13 +850,24 @@ invariant_checker_t::check_for_pc_discontinuity(
...
+#endif
+
+    if (prev_instr.instr.addr != 0 /*first*/ &&
         // We do not bother to support legacy traces without encodings.
         expect_encoding && type_is_instr_direct_branch(shard->prev_instr_.instr.type)) {
         if (shard->prev_instr_.instr.encoding_is_new)
-            shard->branch_target_cache.erase(prev_instr_trace_pc);
-        auto cached = shard->branch_target_cache.find(prev_instr_trace_pc);
+            shard->branch_target_cache.erase(prev_instr.instr.addr);
+        auto cached = shard->branch_target_cache.find(prev_instr.instr.addr);
```
derekbruening, line 870: I had to go add a call to check_for_pc_discontinuity at a kernel event to my code for #6213 and #5490 which overlaps with this PR.  One issue I was about to fix was shard->prev_instr_decoded_ but I figured I should avoid too much overlap -- but I don't see it being fixed here.  Here is the TODO I have in my code:

            // TODO i#5912: This prev_instr_decoded_ isn't always the same as the
            // decoding of the passed-in prev_instr, for signals.  What we should do is
            // create a new instr_info_t struct with fields for the memref_t plus the
            // from-decoding attributes is_syscall, writes_memory, and branch_target, and
            // fill it in at decode time.  We can then store it as prev_instr_ and in
            // pre_signal_instr and last_instr_in_cur_context_ and pass it here as
            // prev_instr.  We can then delete this branch_target_cache and
            // prev_instr_decoded_.  For now we live with this inaccuracy as it is
            // rare to not match and we leave the cleanup for i#5912.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -871,15 +896,11 @@ invariant_checker_t::check_for_pc_discontinuity(
...
             // Regular fall-through.
-            (prev_instr_trace_pc + shard->prev_instr_.instr.size == memref.instr.addr) ||
-            // String loop.
-            (prev_instr_trace_pc == memref.instr.addr &&
-             (memref.instr.type == TRACE_TYPE_INSTR_NO_FETCH ||
-              // Online incorrectly marks the 1st string instr across a thread
-              // switch as fetched.
-              // TODO i#4915, #4948: Eliminate non-fetched and remove the
-              // underlying instrs altogether, which would fix this for us.
-              (!knob_offline_ && shard->saw_timestamp_but_no_instr_))) ||
```
derekbruening, line 882: For the prev_xfer_marker relaxation below:

        // TODO i#5912: Isn't this relaxation too loose since we really only want
        // to relax if a kernel event happened immediately prior, while prev_xfer_marker_
        // could be many records back?

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -345,6 +345,144 @@ check_sane_control_flow()
         if (!run_checker(memrefs, false))
             return false;
     }
+#ifdef UNIX
+    // Negative test (PC discontinuity): Transition from instr to kernel_xfer event
```
abhinav92003, line 349: Let's just use the new notation "correct"/"incorrect" which is clearer than "positive"/"negative".

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -930,7 +933,11 @@ invariant_checker_t::check_for_pc_discontinuity(
     std::string error_msg = "";
     bool have_branch_target = false;
     addr_t branch_target = 0;
-    const addr_t prev_instr_trace_pc = prev_instr.instr.addr;
+    addr_t prev_instr_trace_pc = prev_instr.instr.addr;
```
abhinav92003, line 936: Use the `#else` block to assign the other value.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -701,7 +701,10 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                         memref.marker.marker_value, nullptr,
                         TESTANY(OFFLINE_FILE_TYPE_ENCODINGS, shard->file_type_),
                         /*at_kernel_event=*/true);
-                    report_if_false(shard, discontinuity.empty(), discontinuity);
+                    const std::string error_msg_suffix =
+                        "Discontinuity between instruction and kernel event marker";
```
abhinav92003, line 705: Let's try to abbreviate this message a bit. I'm concerned about storage costs. Maybe just add "@kernel_event marker"?

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -930,7 +932,12 @@ invariant_checker_t::check_for_pc_discontinuity(
     std::string error_msg = "";
     bool have_branch_target = false;
     addr_t branch_target = 0;
-    const addr_t prev_instr_trace_pc = prev_instr.instr.addr;
+#ifdef UNIX
+    addr_t prev_instr_trace_pc = shard->last_instr_in_cur_context_.instr.addr;
```
abhinav92003, line 936: Add a comment explaining why it needs to be last_instr_in_cur_context_: because we want to check continuity with the last instruction at the same signal nesting depth as the current instr (and not the previous executed instruction which may be from an inner nested signal if a signal just ended).

abhinav92003 (approved):

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -176,7 +176,7 @@ bool
 check_branch_target_after_branch()
```
abhinav92003, line 176: Please resolve all existing comments.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -930,7 +932,16 @@ invariant_checker_t::check_for_pc_discontinuity(
...
+#ifdef UNIX
+    // This needs to be last_instr_in_cur_context_ because we want to check continuity
+    // with the last instruction at the same signal nesting depth as the current instr
+    // instead of the previous executed instr (which may be from an inner nested signal if
+    // a signal just ended).
+    addr_t prev_instr_trace_pc = shard->last_instr_in_cur_context_.instr.addr;
+#else
+    addr_t prev_instr_trace_pc = prev_instr.instr.addr;
+#endif
+
```
derekbruening, line 944: What about the 4 different i#5912 comments in this file: I assumed they would be resolved by this PR.  E.g., the prev_instr_decode_ cache: I thought that was broken for for the signal checks being added here.

derekbruening (changes requested):

clients/drcachesim/tools/invariant_checker.cpp
derekbruening: I'm not sure I understand why this branch_target_cache was added in the first place: we always had prev_instr_decoded so we always had the direct target right there.  What we want is a cache of the decoding to avoid calling decode_from_copy() on every single instruction: that's the expensive step.

abhinav92003: > What we want is a cache of the decoding to avoid calling decode_from_copy() on every single instruction:

Yes we should add that now; it'd require updating our current logic of free-ing the `instr_t`s too. I'm a little concerned about memory usage going up due to tracking many `instr_t`, but we can test that experimentally on a large trace.

clients/drcachesim/tests/invariant_checker_test.cpp
derekbruening, line 176: x86-32 failed in what looks like something related to this PR:
```
356:   *** postcmd2 failed (Subprocess aborted): Trace invariant failure in T23655
356:   at ref # 438464 (1 instrs since timestamp 2057181375): Signal handler
356:   return point incorrect
356: 
356:   ***
355/415 Test #356: code_api|tool.histogram.offline ..................................***Failed    4.96 sec
```

clients/drcachesim/tools/invariant_checker.h
```diff
@@ -125,13 +125,20 @@ class invariant_checker_t : public analysis_tool_t {
...
-        std::unique_ptr<instr_autoclean_t> prev_instr_decoded_ = nullptr;
         memref_t prev_xfer_marker_ = {}; // Cleared on seeing an instr.
         memref_t last_xfer_marker_ = {}; // Not cleared: just the prior xfer marker.
         uintptr_t prev_func_id_ = 0;
         addr_t last_retaddr_ = 0;
         uintptr_t trace_version_ = 0;
+        struct instr_info_t {
+            memref_t instr;
+            // Decoding related attributes.
+            bool has_decoded = false;
```
derekbruening, line 136: Is there a clearer name: not sure what this means, and it sounds ungrammatical.  "was_decoded"??  But we're decoding every instr, aren't we?  Is this supposed to be a "valid" field indicating whether the other fields are valid?

clients/drcachesim/tools/invariant_checker.h
```diff
@@ -125,13 +125,20 @@ class invariant_checker_t : public analysis_tool_t {
...
+            bool is_syscall = false;
+            bool writes_memory = false;
+            opnd_t branch_target;
```
derekbruening, line 139: Don't we want an address, not an opnd_t?  This is only for direct branches, right?

clients/drcachesim/tools/invariant_checker.h
```diff
@@ -211,9 +215,10 @@ class invariant_checker_t : public analysis_tool_t {
     // for such violations.
     std::string
     check_for_pc_discontinuity(
-        per_shard_t *shard, const memref_t &memref, const memref_t &prev_instr,
-        addr_t cur_pc, const std::unique_ptr<instr_autoclean_t> &cur_instr_decoded,
-        bool expect_encoding, bool at_kernel_event);
+        per_shard_t *shard, const memref_t &memref,
+        const per_shard_t::instr_info_t &prev_instr_info, addr_t cur_pc,
+        const std::unique_ptr<instr_autoclean_t> &cur_instr_decoded, bool expect_encoding,
```
derekbruening, line 220: I would expect the current instr to be the same info type as the prev instr.

clients/drcachesim/tools/invariant_checker.h
```diff
@@ -125,13 +125,20 @@ class invariant_checker_t : public analysis_tool_t {
...
         memref_t prev_entry_ = {};
-        memref_t prev_instr_ = {};
-        std::unique_ptr<instr_autoclean_t> prev_instr_decoded_ = nullptr;
         memref_t prev_xfer_marker_ = {}; // Cleared on seeing an instr.
         memref_t last_xfer_marker_ = {}; // Not cleared: just the prior xfer marker.
         uintptr_t prev_func_id_ = 0;
         addr_t last_retaddr_ = 0;
         uintptr_t trace_version_ = 0;
+        struct instr_info_t {
+            memref_t instr;
```
derekbruening, line 134: `memref` seems like a clearer name; `instr` sounds like `instr_t` or something.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -530,11 +531,6 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
         // Invariant: non-explicit control flow (i.e., kernel-mediated) is indicated
         // by markers.  Checks are relaxed if a kernel event occurred so prev_instr_
         // is sufficient for now.
-        // XXX i#5912: For adding checks across every part of a signal handler we
-        // should change this to last_instr_in_cur_context_ and remove/adjust the
```
derekbruening, line 533: But this doesn't seem to have been done: it's still passing prev_instr_ instead of last_instr_in_cur_context_.  Should we merge the two, for UNIX, to avoid confusion in the code?

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -581,9 +577,7 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                 // TODO i#5912: since we have the branch decoding now, we can handle
                 // this case.
                 type_is_instr_branch(
-                    shard->last_signal_context_.pre_signal_instr.instr.type) ||
-                shard->last_signal_context_.pre_signal_instr.instr.type ==
-                    TRACE_TYPE_INSTR_SYSENTER;
```
derekbruening, line 585: Why was this removed?

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -682,7 +696,7 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                 shard->last_signal_context_ = { 0, {}, false };
                 // We have not seen any instr in the outermost scope that we just
                 // discovered.
-                shard->last_instr_in_cur_context_ = {};
+                shard->last_instr_in_cur_context_.instr = {};
```
derekbruening, line 699: But if the "has_decoded" (see comment about renaming) is what determines whether this is valid shouldn't we clear that?

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -696,7 +710,7 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                 // signal should be same as the pre-signal instr for the first one.
                 // Here we restore last_instr_in_cur_context_ to the last instr we
                 // saw *in the same nesting depth* before the first signal.
-                shard->last_instr_in_cur_context_ =
+                shard->last_instr_in_cur_context_.instr =
```
derekbruening, line 713: But this needs to set the other fields, or invalidate them.  pre_signal_instr should probably be an instr info type.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -706,21 +720,24 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                 shard->prev_entry_.marker.marker_type == TRACE_MARKER_TYPE_RSEQ_ABORT) {
                 saw_rseq_abort = true;
             } else {
-                // TODO i#3937: We need to exclude this check for "kernel_xfer_app" when
-                // running online. Online traces are missing instructions after sigreturns
-                // which causes the PC discontinuity checks to be violated.
-                if (knob_test_name_ != "kernel_xfer_app" || knob_offline_) {
```
derekbruening: Hmm, why was this checking the test name for offline?  Was this a bug?

derekbruening: Wait I think there is something weird about your sequence of commits: I'm looking at changes since my last review, but the code on the right is what's checked in.  So this is some kind of merge fix where something wasn't merged right before?

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -1026,14 +1025,11 @@ invariant_checker_t::check_for_pc_discontinuity(
             }
             if (have_branch_target && branch_target != cur_pc)
                 error_msg = "Branch does not go to the correct target";
-        } else if (cur_instr_decoded != nullptr &&
-                   shard->prev_instr_decoded_ != nullptr &&
+        } else if (cur_instr_decoded != nullptr && prev_instr_info.has_decoded &&
                    instr_is_syscall(cur_instr_decoded->data) &&
```
derekbruening, line 1029: This should be using the new instr info.  Xref the other comment on expecting the cur instr to also use the new type.

derekbruening (commented):
The cur instr, and not just the prev instr, should use the new type.

clients/drcachesim/tools/invariant_checker.cpp
derekbruening: But both conditions have to be true, so removing this is not correct and will cause failures on 32-bit -- is that why the 32-bit test failed?  Did you investigate that failure?

derekbruening:
Trying to bring some clarity to this PR:

Since it was created back in April, the actual check for PC discont. between pre-signal-instr and interrupted-marker was added separately by #6219 as it was needed for branch target trace changes.

What was left after that I believe is 3 things:

1. Add more tests on that particular check (left out of #6219 deliberately as they belonged here)
2. Fix the corner case of the wrong prev instr by using the prev-instr-in-context instead of prev-instr (which has several side effects like replacing that branch cache to have the right prev instr target)
3. Check the marker on handler return instead of the pre-signal-instr: per https://github.com/DynamoRIO/dynamorio/issues/5912#issuecomment-1500503835

The PR description should be updated to reflect that it is not actually adding new checks for pre-signal to interrupted-marker as they are already in place.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -345,6 +345,142 @@ check_sane_control_flow()
...
+            gen_marker(TID, TRACE_MARKER_TYPE_RSEQ_ABORT, 40),
+            gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_EVENT, 40),
+            // We get a signal after the RSEQ abort.
+            gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_EVENT, 4),
+        };
+        if (!run_checker(memrefs, false)) {
+            return false;
+        }
+    }
+    // Correct test: Branch before signal.
```
abhinav92003, line 428: Add here that this is correct only because the branch doesn't have an encoding with it.

clients/drcachesim/tools/invariant_checker.cpp
derekbruening: No, if I'm on the instr right after a sigreturn, then prev_instr will not equal last_*.  My suggestion as you saw was to merge them and not have them be two different things: do we ever actually want prev_instr?  (I'm not sure: you would have to quickly check the code.) 

It's been so long I don't remember what is already being checked and what this PR is adding.  For the instr after a sigreturn: have we already checked that transition from pre-interruption to interruption point?  Plus the resumption could be a repeat for a synchronous signal or auto-restart syscall: so it is special.  If that's the argument to do nothing then let's make that explicit with a comment.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -779,19 +805,22 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                 shard->prev_entry_.marker.marker_type == TRACE_MARKER_TYPE_RSEQ_ABORT) {
                 saw_rseq_abort = true;
             } else {
-                if (type_is_instr(shard->last_instr_in_cur_context_.instr.type) &&
+                if (type_is_instr(shard->last_instr_in_cur_context_.memref.instr.type) &&
                     !shard->saw_rseq_abort_ &&
                     // XXX i#3937: Online traces do not place signal markers properly,
                     // so we can't precisely check for continuity there.
                     knob_offline_) {
+                    per_shard_t::instr_info_t instr_info {};
```
derekbruening, line 813: No need for `{}`.

clients/drcachesim/tools/invariant_checker.h
```diff
@@ -126,19 +126,33 @@ class invariant_checker_t : public analysis_tool_t {
...
         uintptr_t trace_version_ = 0;
+        // Struct to store decoding related attributes.
+        struct decoding_info_t {
+            bool has_valid_decoding = false;
+            bool is_syscall = false;
+            bool writes_memory = false;
+            bool is_predicated = false;
+            uint num_memory_read_access = 0;
+            uint num_memory_write_access = 0;
+            addr_t branch_target;
```
derekbruening, line 143: = 0

clients/drcachesim/tools/invariant_checker.h
```diff
@@ -126,19 +126,33 @@ class invariant_checker_t : public analysis_tool_t {
...
+        };
+        struct instr_info_t {
+            memref_t memref;
+            decoding_info_t decoding;
+        };
+        std::unordered_map<app_pc, decoding_info_t> decode_cache_;
+        instr_info_t prev_instr_ = {};
```
derekbruening, line 150: See other comments: you have default values for all the fields so no need to zero it out.

clients/drcachesim/tools/invariant_checker.h
```diff
@@ -126,19 +126,33 @@ class invariant_checker_t : public analysis_tool_t {
...
+            bool has_valid_decoding = false;
+            bool is_syscall = false;
+            bool writes_memory = false;
+            bool is_predicated = false;
+            uint num_memory_read_access = 0;
+            uint num_memory_write_access = 0;
+            addr_t branch_target;
+        };
+        struct instr_info_t {
+            memref_t memref;
```
derekbruening, line 146: I would zero this one out `= {}` and then the default constructor is all we need for this struct.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -1062,46 +1092,31 @@ invariant_checker_t::print_results()
 
 std::string
 invariant_checker_t::check_for_pc_discontinuity(
-    per_shard_t *shard, const memref_t &memref, const memref_t &prev_instr, addr_t cur_pc,
-    const std::unique_ptr<instr_autoclean_t> &cur_instr_decoded, bool expect_encoding,
+    per_shard_t *shard, const per_shard_t::instr_info_t &prev_instr_info,
+    const per_shard_t::instr_info_t &cur_instr_info, bool expect_encoding,
```
derekbruening, line 1096: But at a kernel event there is no current instruction: we're at a marker.  It seems a little weird to pass in "cur_instr_info" for a non-instruction.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -1062,46 +1092,31 @@ invariant_checker_t::print_results()
...
     bool at_kernel_event)
 {
+    const memref_t prev_instr = prev_instr_info.memref;
     std::string error_msg = "";
     bool have_branch_target = false;
     addr_t branch_target = 0;
-    const addr_t prev_instr_trace_pc = prev_instr.instr.addr;
-    if (prev_instr_trace_pc == 0 /*first*/)
+    addr_t prev_instr_trace_pc = prev_instr.instr.addr;
+    const addr_t cur_pc = at_kernel_event ? cur_instr_info.memref.marker.marker_value
```
derekbruening, line 1104: This is an assumption that should probably be documented (previously this cur_pc was passed in which seemed a little cleaner) -- that "cur_instr_info" really is a marker if "at_kernel_event".

clients/drcachesim/CMakeLists.txt
```diff
@@ -171,6 +171,7 @@ target_link_libraries(drmemtrace_invariant_checker drdecode)
 
 configure_DynamoRIO_standalone(drmemtrace_opcode_mix)
```
derekbruening, line 172: In the description:

> Fixes the corner case of the wrong previous instruction by using last_instr_in_cur_context_ instead of prev_instr_ when dealing with kernel xfers.

What line # has that change?  I don't see it.

derekbruening (commented):
I am confused as to what this PR is actually changing: it is definitely adding caching, and it adds tests, but I don't see any other change to the checker?  I thought we wanted to change the checks but all I see is some refactoring which doesn't seem to actually change any checks?

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -355,6 +355,207 @@ check_sane_control_flow()
...
+        instrlist_append(ilist, nop);
+        instrlist_append(ilist, move);
+        static constexpr addr_t BASE_ADDR = 0x123450;
+        std::vector<memref_with_IR_t> memref_setup = {
+            { gen_marker(TID, TRACE_MARKER_TYPE_VERSION, TRACE_ENTRY_VERSION_BRANCH_INFO),
+              nullptr },
+            { gen_marker(TID, TRACE_MARKER_TYPE_FILETYPE, OFFLINE_FILE_TYPE_ENCODINGS),
+              nullptr },
+            { gen_instr_type(TRACE_TYPE_INSTR_TAKEN_JUMP, TID), cbr_to_move },
+            { gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_EVENT, 0), move },
```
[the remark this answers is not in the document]
derekbruening: OK so add_encodings_to_memrefs() replaces the 0 marker value with the PC of `move`.  How about `s/0/kWillBeReplaced/` or a comment by the 0 to avoid future readers being confused?

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -620,39 +651,35 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                 memref.instr.addr == shard->last_signal_context_.xfer_int_pc ||
```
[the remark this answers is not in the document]
derekbruening: Per offline discussion: the PR description of adding a check should be removed, since this check is already here.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -620,39 +651,35 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
-                    shard->last_signal_context_.pre_signal_instr.instr.addr ||
+                    shard->last_signal_context_.pre_signal_instr.memref.instr.addr ||
                 // Asynch will go to the subsequent instr.
                 memref.instr.addr ==
-                    shard->last_signal_context_.pre_signal_instr.instr.addr +
-                        shard->last_signal_context_.pre_signal_instr.instr.size ||
-                // Too hard to figure out branch targets.  We have the
-                // last_signal_context_.xfer_int_pc though.
-                // TODO i#5912: since we have the branch decoding now, we can handle
-                // this case.
```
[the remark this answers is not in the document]
derekbruening: From offline discussion: we want to remove this entire `pre_signal_flow_continuity` bool since this A to B check is already being done (and more completely, by checking branches) elsewhere.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -355,6 +355,207 @@ check_sane_control_flow()
...
+            { gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_XFER, 0), nop },
```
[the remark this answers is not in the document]
derekbruening: Like the one above, the 0 is replaced by the nop PC, but that is still incorrect (plus here there are no instrs in the signal handler body) -- so this is a check we are completely missing, that the _XFER marker equals the prev instr fall-through (always a syscall).  OK to add this check in a separate PR; but put a comment here that we know this is not being checked yet.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -507,45 +508,81 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+                instr_noalloc_init(drcontext_, &noalloc);
+                instr_t *noalloc_instr = instr_from_noalloc(&noalloc);
+                app_pc next_pc = decode_from_copy(
+                    drcontext_, const_cast<app_pc>(memref.instr.encoding),
+                    reinterpret_cast<app_pc>(memref.instr.addr), noalloc_instr);
+                if (next_pc == nullptr) {
+                    noalloc_instr = nullptr;
+                }
+                // Add decoding attributes to cur_instr_info.
+                if (noalloc_instr != nullptr) {
```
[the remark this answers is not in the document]
derekbruening: I meant: remove the `if (noalloc_instr != nullptr)`.  Since `noalloc_instr` is non-null by construction before the decode_from_copy call, and decode_from_copy cannot set it to null, it seems weird to convert the return value from decode_from_copy into it being null.  We still need to check that the decoding succeeded.  I was suggesting to put all the valid decoding code inside `if (next_pc != nullptr)`.  (Certainly the new code is incorrect as it will mark it valid even if the decoding fails.)

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -627,8 +623,7 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                 "Branch target not immediately after branch");
         }
         // Invariant: non-explicit control flow (i.e., kernel-mediated) is indicated
-        // by markers.  Checks are relaxed if a kernel event occurred so prev_instr_
-        // is sufficient for now.
+        // by markers.
```
derekbruening, line 626: But it does seem worth explaining why prev_instr_ is used here instead of the context one: because after a signal the interruption and resumption checks are done elsewhere and this call turns into a nop essentially (right?).

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -653,29 +648,9 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
-                    shard->last_signal_context_.pre_signal_instr.memref.instr.addr +
-                        shard->last_signal_context_.pre_signal_instr.memref.instr.size ||
-                type_is_instr_branch(
-                    shard->last_signal_context_.pre_signal_instr.memref.instr.type) ||
-                shard->last_signal_context_.pre_signal_instr.memref.instr.type ==
-                    TRACE_TYPE_INSTR_SYSENTER;
             report_if_false(
                 shard,
-                (kernel_event_marker_equality && pre_signal_flow_continuity) ||
+                (kernel_event_marker_equality) ||
```
derekbruening, line 653: nit: superfluous parentheses

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -810,7 +785,7 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                     // XXX i#3937: Online traces do not place signal markers properly,
                     // so we can't precisely check for continuity there.
                     knob_offline_) {
-                    per_shard_t::instr_info_t instr_info {};
+                    per_shard_t::instr_info_t instr_info;
```
derekbruening, line 788: memref_info as this is not an instr, it is a marker.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -1093,16 +1068,17 @@ invariant_checker_t::print_results()
...
 {
     const memref_t prev_instr = prev_instr_info.memref;
     std::string error_msg = "";
     bool have_branch_target = false;
     addr_t branch_target = 0;
-    addr_t prev_instr_trace_pc = prev_instr.instr.addr;
-    const addr_t cur_pc = at_kernel_event ? cur_instr_info.memref.marker.marker_value
-                                          : cur_instr_info.memref.instr.addr;
+    const addr_t prev_instr_trace_pc = prev_instr.instr.addr;
+    // cur_memref_info is a marker (not an instruction) if at_kernel_event is true.
```
derekbruening, line 1079: (I preferred passing in the cur_pc with the logic in the caller where it seems natural and not needing this logic inside here but not strongly enough to request a change; the at_kernel_event is still needed in any case.)

clients/drcachesim/tools/invariant_checker.h
```diff
@@ -140,14 +140,15 @@ class invariant_checker_t : public analysis_tool_t {
...
             uint num_memory_read_access = 0;
             uint num_memory_write_access = 0;
-            addr_t branch_target;
+            addr_t branch_target = 0;
         };
         struct instr_info_t {
-            memref_t memref;
-            decoding_info_t decoding;
+            memref_t memref = {};
+            decoding_info_t decoding = {};
```
derekbruening, line 147: But you have default values for all the fields so why set those and then do a separate memset to zero?

derekbruening: Yes, you need the memref memset: that is what I pointed out should be added in the prior review.  It is a C struct.  What you don't need is the decoding_info_t which is a C++ struct with every field initialized in its default constructor -- that is why I have the comment on this line, not the line above it.

clients/drcachesim/CMakeLists.txt
```diff
@@ -171,6 +171,7 @@ target_link_libraries(drmemtrace_invariant_checker drdecode)
```
derekbruening, line 171: Please update the PR description:

> Fixes the corner case of the wrong previous instruction by using last_instr_in_cur_context_ instead of prev_instr_ when dealing with kernel xfers.

I don't see this change.  This line should be removed I assume.

> Adds unit tests to check for PC discontinuities in the transitions between instructions and kernel xfer markers
> Adds caching around the decoding, to prevent repeated calls to the expensive decode_from_copy() function.

These 2 are not changes "to how the invariant checker checks for discontinuities between kernel xfer markers": please rearrange.

Probably worth mentioning that the big redundant pre_signal_whatever_it_was check is removed.

> Issue: https://github.com/DynamoRIO/dynamorio/issues/5912

Does this fix that issue?

derekbruening (commented):
Just the PR description left.

clients/drcachesim/CMakeLists.txt
derekbruening, line 171: Also, doesn't this fix #6006?  That should be in the PR description.

derekbruening, line 171: > i#5912: Modifications to PC discontinuity checks

"Modifications" is a little general: how about "Add caching and clean up PC discontinuity checks"?

derekbruening (approved):

PR #5970, opened 2023-04-12
i#4041: Use rseq handler PC instead of commit PC

### LINKED ISSUE #4041: trace signal placement in rseq region is wrong due to abort handler redirect
Gathering an offline drcachesim trace on the rseq test, for #4019, shows a signal marker too late -- it should be right after the ud2a:

```
$ bin64/drrun -t drcachesim -offline -- suite/tests/bin/linux.rseq
$ bin64/drrun -t drcachesim -simulator_type view -indir drmemtrace.linux.rseq.123467.3754.dir/
=>
  0x00007f7a6c9e78f3  0f 0b                ud2a
  0x00007f7a6c9e78f5  83 05 4c 38 00 00 01 addl   $0x01, <rel> 0x00007f7a6c9eb148
<marker: kernel xfer to handler>
<marker: timestamp 13223619762485105>
<marker: tid 123467 on core 0>
  0x00007f7a6c9e785c  55                   push   %rbp
<...>
<marker: syscall xfer>
<marker: timestamp 13223619762492787>
<marker: tid 123467 on core 0>
  0x00007f7a6c9e7901  83 05 44 38 00 00 01 addl   $0x01, <rel> 0x00007f7a6c9eb14c
  0x00007f7a6c9e7908  c6 05 41 38 00 00 00 movb   $0x00, <rel> 0x00007f7a6c9eb150
```

This is caught by the invariants test, which I was hoping to add as a regression test for #4019:
```
$ bin64/drrun -t drcachesim -offline -- suite/tests/bin/linux.rseq
$ bin64/drrun -t drcachesim -test_mode -indir drmemtrace.linux.rseq.123467.3754.dir/
drcachesim: /home/bruening/dr/git/src/clients/drcachesim/tests/trace_invariants.cpp:143: virtual bool trace_invariants_t::process_memref(const memref_t&): Assertion `memref.instr.tid != pre_signal_instr.instr.tid || memref.instr.addr == pre_signal_instr.instr.addr || memref.instr.addr == pre_signal_instr.instr.addr + pre_signal_instr.instr.size || memref.instr.addr == app_handler_pc || type_is_instr_branch(pre_signal_instr.instr.type) || pre_signal_instr.instr.type == TRACE_TYPE_INSTR_SYSENTER' failed.
Aborted
```

Looks like the recorded offset is wrong:
```
[drmemtrace]: Appending 2 instrs in bb 0x7fcb228a48f3 in mod 5 +0x18f3 = /home/bruening/dr/git/build_x64_dbg_tests/suite/tests/bin/linux.rseq
  0x00007fdb501da8f3 ud2a
[drmemtrace]: Checking whether reached signal/exception +0x1901 vs cur +0x18f3
  0x00007fdb501da8f5 add    $0x00000001 <rel> 0x00007fdb501de198[4byte] -> <rel> 0x00007fdb501de198[4byte]
[drmemtrace]: Checking whether reached signal/exception +0x1901 vs cur +0x18f5
[drmemtrace]: Filling in elided rip-rel addr with 0x7fdb501de198
[drmemtrace]: Appended memref type 0 size 4 to 0x7fdb501de198
[drmemtrace]: Checking whether reached signal/exception +0x1901 vs cur +0x18fc
[drmemtrace]: Filling in elided rip-rel addr with 0x7fdb501de198
[drmemtrace]: Appended memref type 1 size 4 to 0x7fdb501de198
[drmemtrace]: Checking whether reached signal/exception +0x1901 vs cur +0x18fc
[drmemtrace]: Signal/exception between bbs
[drmemtrace]: Appended marker type 0 value 0x1901
```

Ah here we go:
```
recreate_app : looking for 0x00007fdb103cf90b in frag @ 0x00007fdb103cf8d4 (tag 0x00007fdb501da8f3)
recreate_app -- found valid state pc 0x00007fdb501da8f3
        translation 0x00007fdb501da8f3 is post-walk 0x0000000000000000 so not fixing xsp
recreate_app: moving 0x00007fdb501da8f3 inside rseq region to handler 0x00007fdb501da901
recreate_app -- found ok pc 0x00007fdb501da901
```

Hmm, how can we solve that?  The kernel does not record the actual signal
PC (unlike rseq v1), so DR does not either, and thus the tracer does not
know when the signal was triggered.

derekbruening:
The Linux kernel presents the rseq abort handler PC as the interrupted (really, the resume point) for a signal generated while inside an rseq region.  Yet DR was modifying that to instead show clients the rseq committing store PC.  Similarly, DR presented that commit PC for its rseq abort event if an rseq region aborted through some other non-signal means like a thread migration.  Since this is a little misleading and doesn't match what the app sees, and results in the signal interrutped PC not matching the later resume PC, we change that behavior here to have both signal and rseq events use the handler PC. There is no way to obtain the precise interruption PC from the kernel so we follow that lead and do not try to present it from DR.

Furthemore, DR was not previously generating an rseq abort event if a signal was the cause of the abort.  We change that here, raising the event prior to the signal event.

Adds a new drmemtrace label TRACE_MARKER_TYPE_RSEQ_HANDLER to communicate the handler PC to raw2trace.  Updates raw2trace to use this to detect an rseq abort.

Adds a new test case to raw2trace_unit_tests.

Updates the api.rseq test.

Fixes #4041

### DIFF
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -157,7 +157,8 @@ changes:
    decode "retaa" and "retab". These instructions will now encode and decode correctly
    as "retaa" and "retab".
  - Changed the interrupted PC for #DR_XFER_RSEQ_ABORT and a signal generated during
-   an rseq region to be the abort handler rather than the committing store.
+   an rseq region to be the abort handler rather than the committing store,
+   matching the kernel behavior.
  - Added a #DR_XFER_RSEQ_ABORT event for a signal generated during an rseq region.
 
 Further non-compatibility-affecting changes include:
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -227,6 +227,8 @@ Further non-compatibility-affecting changes include:
    query key attributes of each input trace.
  - Added instr_create_1dst_6src() convenience function that returns an instr_t
    with one destination and six sources.
+ - Added a new label to help in handling "rseq" (Linux restartable sequence) regions:
+   #DR_NOTE_RSEQ_ENTRY.
 
 **************************************************
 <hr>
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -156,6 +156,9 @@ changes:
    AArch64 instruction and "reta" entries in the AArch64 codec were being used to
    decode "retaa" and "retab". These instructions will now encode and decode correctly
    as "retaa" and "retab".
+ - Changed the interrupted PC for #DR_XFER_RSEQ_ABORT and a signal generated during
+   an rseq region to be the abort handler rather than the committing store.
+ - Added a #DR_XFER_RSEQ_ABORT event for a signal generated during an rseq region.
 
 Further non-compatibility-affecting changes include:
  - Added AArchXX support for attaching to a running process.
--- a/clients/drcachesim/common/trace_entry.h
+++ b/clients/drcachesim/common/trace_entry.h
@@ -348,7 +348,9 @@ typedef enum {
     /**
      * Serves to further identify #TRACE_MARKER_TYPE_KERNEL_EVENT as a
      * restartable sequence abort handler.  This will always be immediately followed
-     * by #TRACE_MARKER_TYPE_KERNEL_EVENT.
+     * by #TRACE_MARKER_TYPE_KERNEL_EVENT.  The marker value holds the continuation
+     * program counter, which is the restartable sequence abort handler.  (The precise
+     * interrupted point inside the sequence is not provided by the kernel.)
      */
     TRACE_MARKER_TYPE_RSEQ_ABORT,
 
--- a/clients/drcachesim/common/trace_entry.h
+++ b/clients/drcachesim/common/trace_entry.h
@@ -1,5 +1,5 @@
 /* **********************************************************
- * Copyright (c) 2015-2022 Google, Inc.  All rights reserved.
+ * Copyright (c) 2015-2023 Google, Inc.  All rights reserved.
  * **********************************************************/
 
 /*
@@ -424,6 +424,13 @@ typedef enum {
      * warmup part of the trace ends.
      */
     TRACE_MARKER_TYPE_FILTER_ENDPOINT,
+
+    /**
+     * Indicates the start of an "rseq" (Linux restartable sequence) region.  The marker
+     * value holds the end PC of the region (this is the PC after the committing store).
+     */
+    TRACE_MARKER_TYPE_RSEQ_ENTRY,
+
     // ...
     // These values are reserved for future built-in marker types.
     // ...
@@ -482,6 +489,16 @@ type_has_address(const trace_type_t type)
 {
     return type_is_instr(type) || type == TRACE_TYPE_INSTR_NO_FETCH ||
         type == TRACE_TYPE_INSTR_MAYBE_FETCH || type_is_prefetch(type) ||
+        type == TRACE_TYPE_READ || type == TRACE_TYPE_WRITE ||
+        type == TRACE_TYPE_INSTR_FLUSH || type == TRACE_TYPE_INSTR_FLUSH_END ||
+        type == TRACE_TYPE_DATA_FLUSH || type == TRACE_TYPE_DATA_FLUSH_END;
+}
+
+/** Returns whether the type represents an operand of an instruction. */
+static inline bool
+type_is_data(const trace_type_t type)
+{
+    return type == TRACE_TYPE_INSTR_MAYBE_FETCH || type_is_prefetch(type) ||
         type == TRACE_TYPE_READ || type == TRACE_TYPE_WRITE ||
         type == TRACE_TYPE_INSTR_FLUSH || type == TRACE_TYPE_INSTR_FLUSH_END ||
         type == TRACE_TYPE_DATA_FLUSH || type == TRACE_TYPE_DATA_FLUSH_END;
--- a/clients/drcachesim/common/trace_entry.h
+++ b/clients/drcachesim/common/trace_entry.h
@@ -438,12 +438,6 @@ typedef enum {
      */
     TRACE_MARKER_TYPE_RSEQ_ENTRY,
 
-    /**
-     * Indicates the PC of the abort handler for the following rseq region.  Always
-     * follows #TRACE_MARKER_TYPE_RSEQ_ENTRY.
-     */
-    TRACE_MARKER_TYPE_RSEQ_HANDLER,
-
     // ...
     // These values are reserved for future built-in marker types.
     // ...
--- a/clients/drcachesim/common/trace_entry.h
+++ b/clients/drcachesim/common/trace_entry.h
@@ -431,6 +431,12 @@ typedef enum {
      */
     TRACE_MARKER_TYPE_RSEQ_ENTRY,
 
+    /**
+     * Indicates the PC of the abort handler for the following rseq region.  Always
+     * follows #TRACE_MARKER_TYPE_RSEQ_ENTRY.
+     */
+    TRACE_MARKER_TYPE_RSEQ_HANDLER,
+
     // ...
     // These values are reserved for future built-in marker types.
     // ...
--- a/clients/drcachesim/common/trace_entry.h
+++ b/clients/drcachesim/common/trace_entry.h
@@ -425,6 +425,11 @@ typedef enum {
      */
     TRACE_MARKER_TYPE_FILTER_ENDPOINT,
 
+    // We use one marker at the start whose data is the end, instead of a separate
+    // marker at the end PC, as this seems easier for users to process as they can plan
+    // ahead. Also, when the rseq aborted, if we had the marker at the committing store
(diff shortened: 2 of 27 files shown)


clients/drcachesim/common/trace_entry.h
```diff
@@ -431,6 +431,12 @@ typedef enum {
      */
     TRACE_MARKER_TYPE_RSEQ_ENTRY,
```
abhinav92003, line 433: Update docx of TRACE_MARKER_TYPE_RSEQ_ABORT to specify what the address in the marker represents.

clients/drcachesim/common/trace_entry.h
```diff
@@ -431,6 +431,12 @@ typedef enum {
...
 
+    /**
+     * Indicates the PC of the abort handler for the following rseq region.  Always
+     * follows #TRACE_MARKER_TYPE_RSEQ_ENTRY.
+     */
+    TRACE_MARKER_TYPE_RSEQ_HANDLER,
```
abhinav92003, line 438: > There is no way to obtain the precise interruption PC from the kernel

So for all types of rseq abort (signal, thread migration etc), the pc shown by the kernel and DR is the abort handler pc.

I'm thinking why we need to specify the handler address separately when TRACE_MARKER_TYPE_RSEQ_ABORT already has it. Maybe the invariant checker should just make sure that all instances of TRACE_MARKER_TYPE_RSEQ_ABORT have the same address.

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -1832,12 +1845,11 @@ raw2trace_t::adjust_and_emit_rseq_buffer(raw2trace_thread_data_t *tdata, addr_t
...
-    // We actually won't see the handler PC as the rseq abort event has the
-    // commit PC and we have it call here and don't wait for a subsequent
-    // instr (or signal interruption) to find the handler PC; so here it's
-    // not really the "next_pc" but the "exit_pc".
-    else if (next_pc == tdata->rseq_commit_pc_) {
+    } else if ((at_abort_marker && next_pc == tdata->rseq_handler_pc_) ||
+               // Old traces have the commit PC instead of the handler in the abort and
+               // signal events (so not really the "next_pc" but the "exit_pc") and
+               // don't have abort events on a signal aborting an rseq region.
+               next_pc == tdata->rseq_commit_pc_) {
```
abhinav92003, line 1852: Add `&& !rseq_ever_saw_entry_` to the second one?

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -1832,12 +1845,11 @@ raw2trace_t::adjust_and_emit_rseq_buffer(raw2trace_thread_data_t *tdata, addr_t
         // up on distinguishing a side exit whose target is the end PC fall-through from
         // a completion.
         log(4, "Rseq completed normally\n");
-    }
-    // We actually won't see the handler PC as the rseq abort event has the
-    // commit PC and we have it call here and don't wait for a subsequent
-    // instr (or signal interruption) to find the handler PC; so here it's
-    // not really the "next_pc" but the "exit_pc".
-    else if (next_pc == tdata->rseq_commit_pc_) {
+    } else if ((at_abort_marker && next_pc == tdata->rseq_handler_pc_) ||
```
abhinav92003, line 1848: Is this why we want to have TRACE_MARKER_TYPE_RSEQ_HANDLER? Instead of having just `at_abort_marker` here?

suite/tests/linux/rseq.c
```diff
@@ -1123,9 +1127,15 @@ kernel_xfer_event(void *drcontext, const dr_kernel_xfer_info_t *info)
...
+     * For a signal it won't be as signal handler arguments will be set in
+     * registers and a sigaltstack may be set.
+     */
+    if (info->type == DR_XFER_RSEQ_ABORT && in_rseq_migration_test) {
+        /* The interrupted context should be identical, including the pc now
+         * that DR passes the abort handler PC just like the kernel does.
+         */
+        assert(info->source_mcontext->pc == mc.pc);
 #    ifdef X86
         assert(info->source_mcontext->xax == mc.xax);
```
abhinav92003, line 1140: Now that the pc is also the same, does the source_mcontext struct layout allow us to use a memcmp on x86 like we do on AArch64?

suite/tests/linux/rseq.c
```diff
@@ -1141,10 +1151,9 @@ kernel_xfer_event(void *drcontext, const dr_kernel_xfer_info_t *info)
 #        error Unsupported arch
 #    endif
 #    ifdef DEBUG /* See above: special code in core/ is DEBUG-only> */
-        /* Check that the interrupted PC for the true abort case is *prior* to the
-         * committing store.
-         */
-        assert(info->source_mcontext->pc == (app_pc)test_rseq_native_abort_pre_commit);
+        /* Check that the interrupted PC for the true abort case is the handler. */
+        assert(!in_rseq_migration_test ||
```
abhinav92003, line 1155: Since we're inside the if-body, in_rseq_migration_test is true always.

suite/tests/linux/rseq.c
```diff
@@ -1141,10 +1151,9 @@ kernel_xfer_event(void *drcontext, const dr_kernel_xfer_info_t *info)
...
+               info->source_mcontext->pc == (app_pc)test_rseq_native_abort_handler);
 #    endif
     }
```
abhinav92003, line 1158: For !in_rseq_migration_test also, can we check info->source_mcontext->pc == (app_pc)test_rseq_native_abort_handler?

clients/drcachesim/common/trace_entry.h
abhinav92003, line 433: TRACE_MARKER_TYPE_RSEQ_ABORT is always followed by TRACE_MARKER_TYPE_KERNEL_EVENT. What address does the kernel event marker have?

abhinav92003 (approved):

api/docs/release.dox
```diff
@@ -156,6 +156,9 @@ changes:
    AArch64 instruction and "reta" entries in the AArch64 codec were being used to
    decode "retaa" and "retab". These instructions will now encode and decode correctly
    as "retaa" and "retab".
+ - Changed the interrupted PC for #DR_XFER_RSEQ_ABORT and a signal generated during
+   an rseq region to be the abort handler rather than the committing store.
```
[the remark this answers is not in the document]
derekbruening: Done.

clients/drcachesim/common/trace_entry.h
derekbruening: Done.

clients/drcachesim/common/trace_entry.h
derekbruening: Hmm, I did not consider this: I think you are right and we could eliminate this new marker.  Let me try to adjust the code for that.

clients/drcachesim/common/trace_entry.h
derekbruening: The same abort handler PC.  It does seem redundant; but simpler to leave it I think than possibly break someone by changing (though we're already changing the abort marker value...)

suite/tests/linux/rseq.c
derekbruening: Removed.

suite/tests/linux/rseq.c
derekbruening: No, each test has its own abort handler.  We'd have to add a global label exported from the asm for each test and then have a big || here or have another global saying which test it was...doesn't seem worth it.

derekbruening: Or a global var pointing at which handler should be expected...but then have to write it from the asm, or again export it.  IMHO the check for one test is enough.

derekbruening: Added a comment on this.

suite/tests/linux/rseq.c
derekbruening: Yes, can compare from xdi up through pc.  Done.

clients/drcachesim/tracer/raw2trace.cpp
derekbruening: The problem is there's a window between PR #5967 and this one: I guess it's very narrow since it looks like this will be merged right after but at the time I wasn't sure.  Maybe we can not support that PR without this one: I might have to update a unit test for that.  Trying it.

clients/drcachesim/tracer/raw2trace.cpp
derekbruening: Right but if we're at the abort marker we have the handler value as you pointed out so trying the earlier suggestion.

clients/drcachesim/common/trace_entry.h
derekbruening: Confirmed: the new marker can be eliminated.  Done.

clients/drcachesim/tracer/raw2trace.cpp
derekbruening: Done.

PR #5976, opened 2023-04-14
i#5951: Make -only_thread a top-level analyzer option

### LINKED ISSUE #5951: drmemtrace -only_thread should open just one file
Today the -only_thread flag is only passed to the view tool, with all other threads still opened and passed through the tool.  With the new scheduler interface's thread input filter we should be able to turn -only_thread into a filter such that just one trace file is opened. We should perhaps document this behavior in case multiple tools are specified: that -only_thread is a global filter.

derekbruening:
Applies -only_thread at the analyzer framework level, setting a thread filter for the scheduler.  This results in just one thread file being opened.

Removes the only_thread knob from the view tool.

Tested:
Before:
  $ bin64/drrun -t drcachesim -verbose 1 -simulator_type view -indir drmemtrace.*.dir -only_thread 2557808 -sim_refs 20 2>&1 | grep -c 'Opened reader'
  9
After:
  $ bin64/drrun -t drcachesim -verbose 1 -simulator_type view -indir drmemtrace.*.dir -only_thread 2557808 -sim_refs 20 2>&1 | grep -c 'Opened reader'
  0

Fixes #5951

### DIFF
--- a/clients/drcachesim/analyzer.cpp
+++ b/clients/drcachesim/analyzer.cpp
@@ -136,6 +136,7 @@ analyzer_tmpl_t<RecordType, ReaderType>::analyzer_tmpl_t()
 template <typename RecordType, typename ReaderType>
 bool
 analyzer_tmpl_t<RecordType, ReaderType>::init_scheduler(const std::string &trace_path,
+                                                        memref_tid_t only_thread,
                                                         int verbosity)
 {
     verbosity_ = verbosity;
@@ -152,6 +153,9 @@ analyzer_tmpl_t<RecordType, ReaderType>::init_scheduler(const std::string &trace
         regions.emplace_back(skip_instrs_ + 1, 0);
     }
     typename sched_type_t::input_workload_t workload(trace_path, regions);
+    if (only_thread != INVALID_THREAD_ID) {
+        workload.only_threads.insert(only_thread);
+    }
     return init_scheduler_common(workload);
 }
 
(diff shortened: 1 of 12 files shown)


derekbruening:
x86-32 failure is scattergather #5329

derekbruening:
@AssadHashmi the Jenkins trigger doesn't seem to be working: is that related to the recent work you mentioned on Actions?

clients/drcachesim/analyzer.cpp
```diff
@@ -136,6 +136,7 @@ analyzer_tmpl_t<RecordType, ReaderType>::analyzer_tmpl_t()
 template <typename RecordType, typename ReaderType>
```
abhinav92003, line 136: Should also add a test that exercises the new option.

abhinav92003 (approved):
Thanks for this PR. I had some hacky changes in a local client to run invariant checker on only a particular thread. But this simplifies it.

clients/drcachesim/analyzer.cpp
derekbruening: Added a test of basic_counts with only_thread on the checked-in threadsig trace.

AssadHashmi:
> @AssadHashmi the Jenkins trigger doesn't seem to be working: is that related to the recent work you mentioned on Actions?

Not that I know of. Jenkins may have gone down temporarily but seems to be back up and running again.

Issue #5954, opened 2023-04-03
rseq not rolled back in drmemtrace post-processing due to a timestamp marker


derekbruening:
With our run-twice solution, drmemtrace always records the full rseq sequence, and we roll back the committing store on an abort.  But the logic in raw2trace to detect the abort is thwarted if a signal arrived before we execute the rseq abort handler.  This is a not-uncommon occurrence in our proprietary app.

The rseq-delimited buffer in raw2trace proposed for #5953 should solve this.

derekbruening:
Our analysis was wrong: it is actually a timestamp before the abort marker (due to happening to fill up the buffer at that point) that thwarts the check.

PR #5978, opened 2023-04-15
i#5980: Fix 3 rseq exit/signal issues

### LINKED ISSUE #5980: raw2trace failing with "Failed to identify rseq commit index"
After the #5953 and #4041 fixes, large apps with a lot of rseq are failing to postprocess with "Failed to identify rseq commit index".

While investigating several other issues were found:

1. Side exits in instrumented runs are not detecting, leading to unbounded rseq buffer growth.

2. Asynch signal in instrumented run: this was not being adjusted so the interruption/continuation PC is the abort handler. Adds a test to linux.rseq.

3. A double-abort was seen:
```
    76394602    60025222:       96026 <marker: rseq abort from 0x2473ecd4 to handler>
    76394603    60025222:       96026 <marker: kernel xfer from module offset +0x2473ecd4 to handler>
    76394604    60025222:       96026 <marker: timestamp 13325925760303605>
    76394605    60025222:       96026 <marker: tid 96026 on core 14>
    76394606    60025222:       96026 <marker: rseq abort from 0x138531e0 to handler>
    76394607    60025222:       96026 <marker: kernel xfer from module offset +0x138531e0 to handler>
    76394608    60025222:       96026 <marker: timestamp 13325925760303617>
    76394609    60025222:       96026 <marker: tid 96026 on core 14>
    76394610    60025222:       96026 <marker: kernel xfer from module offset +0x2473ecd4 to handler>
    76394611    60025222:       96026 <marker: timestamp 13325925760303624>
    76394612    60025222:       96026 <marker: tid 96026 on core 14>
    76394613    60025223:       96026 ifetch       
<...>
    76394968    60025476:       96026 <marker: timestamp 13325925760303635>
    76394969    60025476:       96026 <marker: tid 96026 on core 14>
    76394970    60025476:       96026 <marker: syscall xfer from module offset +0x12da6b90>
    76394971    60025476:       96026 <marker: timestamp 13325925760303641>
    76394972    60025476:       96026 <marker: tid 96026 on core 14>
    76394973    60025477:       96026 ifetch
```
Theory:
+ asynch signal came in while in rseq region
+ native rseq aborted and DR generated rseq+xfer@0x2473ecd4
+ delivery of pending asynch signal didn't require xl8 so the last xl8 was rseq so the signal resulted in rseq+xfer@committing-store plus xfer@next-pc==abort-handler

If the signal arrived while in the native rseq: it caused the abort. But it could have arrived earlier in the instru portion of the same fragment.

Better to remove the 1st abort+xfer marker pair (presenting as though signal caused abort: which is only true some of the time), or to remove the 2nd abort pair (presenting as though the signal arrived after the abort when about to run the abort handler)? It is easier to implement the second choice (since the signal "last xl8" is fragile). Though the current docs say any abort in native execution, signal or scheduler, doesn't have the interrupted PC: so this easier path still matches thatTry to handle double-abort: couldn't repro in test though so the fix of disabling a later signal from reporting an abort when we report an abort fragment exit is not fully vetted.

derekbruening:
Fixes 3 more problems with rseq in DR and in drmemtraces related to exits and signals:

1) Side exit in instrumented run: avoids unbounded buffer growth by detecting by looking at the region bounds.  Adds a buffer large size sanity check.  Adds a test to linux.rseq.

2) Try to handle double-abort: couldn't repro in test though so the fix of disabling a later signal from reporting an abort when we report an abort fragment exit is not fully vetted.

3) Asynch signal in instrumented run: this was not being adjusted so the interruption/continuation PC is the abort handler.  Adds a test to linux.rseq.

Adds more invariant checks around rseq region exits.

Issue: #5953, #4041, #5954, #5980
Fixes #5980

### DIFF
--- a/clients/drcachesim/tools/invariant_checker.cpp
+++ b/clients/drcachesim/tools/invariant_checker.cpp
@@ -207,7 +207,33 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
     }
     if (memref.marker.type == TRACE_TYPE_MARKER &&
         memref.marker.marker_type == TRACE_MARKER_TYPE_RSEQ_ENTRY) {
+        shard->in_rseq_region_ = true;
+        shard->rseq_start_pc_ = 0;
         shard->rseq_end_pc_ = memref.marker.marker_value;
+    } else if (shard->in_rseq_region_) {
+        if (type_is_instr(memref.instr.type)) {
+            if (shard->rseq_start_pc_ == 0)
+                shard->rseq_start_pc_ = memref.instr.addr;
+            if (memref.instr.addr + memref.instr.size == shard->rseq_end_pc_) {
+                // Completed normally.
+                shard->in_rseq_region_ = false;
+            } else if (memref.instr.addr >= shard->rseq_start_pc_ &&
+                       memref.instr.addr < shard->rseq_end_pc_) {
+                // Still in the region.
+            } else {
+                // We should see an abort marker or a side exit if we leave the region.
+                report_if_false(shard,
+                                type_is_instr_branch(shard->prev_instr_.instr.type),
+                                "Rseq region exit requires marker, branch, or commit");
+                shard->in_rseq_region_ = false;
+            }
+        } else {
+            report_if_false(shard,
+                            memref.marker.type != TRACE_TYPE_MARKER ||
+                                memref.marker.marker_type !=
+                                    TRACE_MARKER_TYPE_KERNEL_EVENT,
+                            "Signal in rseq region should have abort marker");
+        }
     }
     if (memref.marker.type == TRACE_TYPE_MARKER &&
         memref.marker.marker_type == TRACE_MARKER_TYPE_RSEQ_ABORT) {
@@ -221,6 +247,7 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                                 shard->rseq_end_pc_ ||
                             shard->prev_instr_.instr.addr == memref.marker.marker_value,
                         "Rseq post-abort instruction not rolled back");
+        shard->in_rseq_region_ = false;
     }
 #endif
 
--- a/clients/drcachesim/tools/invariant_checker.cpp
+++ b/clients/drcachesim/tools/invariant_checker.cpp
@@ -231,7 +231,9 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
             report_if_false(shard,
                             memref.marker.type != TRACE_TYPE_MARKER ||
                                 memref.marker.marker_type !=
-                                    TRACE_MARKER_TYPE_KERNEL_EVENT,
+                                    TRACE_MARKER_TYPE_KERNEL_EVENT ||
+                                // Side exit.
+                                type_is_instr_branch(shard->prev_instr_.instr.type),
                             "Signal in rseq region should have abort marker");
         }
     }
--- a/clients/drcachesim/tracer/raw2trace.cpp
+++ b/clients/drcachesim/tracer/raw2trace.cpp
@@ -585,6 +585,9 @@ raw2trace_t::process_offline_entry(raw2trace_thread_data_t *tdata,
             } else if (in_entry->extended.valueB == TRACE_MARKER_TYPE_RSEQ_ENTRY) {
                 if (tdata->rseq_want_rollback_) {
                     if (tdata->rseq_buffering_enabled_) {
+                        // Our rollback schemes do the minimal rollback: for a side
+                        // exit, taking the last branch.  This means we don't need the
+                        // prior iterations in the buffer.
                         log(4, "Rseq was already buffered: assuming loop; emitting\n");
                         err = adjust_and_emit_rseq_buffer(tdata, marker_val);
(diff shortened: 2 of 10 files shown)


derekbruening:
x64 drcachesim.threads failure looks like a flake: will file new issue and re-run if not already there (not seeing on quick search)

derekbruening:
Maybe I should file a new issue or multiple issues on these instead of putting under the existing

suite/tests/linux/rseq.c
```diff
@@ -812,6 +812,126 @@ test_rseq_native_abort(void)
...
+        "movq %%rcx, %%xmm1\n\t"
+
+        /* Restartable sequence. */
+        "2:\n\t"
+        /* Increase xmm0 every time.  DR (currently) won't restore xmm inputs
+         * to rseq sequences, nor does it detect that it needs to.
+         */
+        "paddq %%xmm1,%%xmm0\n\t"
+        "movq %%xmm0, %%rax\n\t"
+        /* We raise the signal on the instrumented run and never come back. */
```
abhinav92003, line 847: Add "in the instrumented run" to the test call-site in `main` and `instru` to the test function name.

Edit: oh so this test actually does the same thing for the native and instrumented run.

What about a case where there's a signal in the instrumented run but the native run completes properly? (I assume we indeed do a native run even if the instrumented one is interrupted by a signal or a side exit, right?)

suite/tests/linux/rseq.c
```diff
@@ -812,6 +812,126 @@ test_rseq_native_abort(void)
...
+        /* clang-format on */
+
+        : [rseq_cs] "=m"(reg_rseq->rseq_cs), [restarts] "=m"(restarts)
+        : [signum_alarm] "i"(SIGALRM), [sysnum_kill] "i"(SYS_kill)
+        : "x0", "x1", "x2", "x8", "q0", "q1", "memory");
+#    else
+#        error Unsupported arch
+#    endif
+    /* This is expected to fail on a native run where restarts will be 0. */
+    assert(restarts > 0);
```
abhinav92003, line 931: Check `commits` too? Like the other tests.

suite/tests/linux/rseq.c
```diff
@@ -927,6 +1047,119 @@ test_rseq_side_exit(void)
...
+
+        /* Restartable sequence. */
+        "2:\n\t"
+        /* Increase xmm0 every time.  DR (currently) won't restore xmm inputs
+         * to rseq sequences, nor does it detect that it needs to.
+         */
+        "paddq %%xmm1,%%xmm0\n\t"
+        "movq %%xmm0, %%rax\n\t"
+        /* Take a side exit in the 1st run == instrumented run. */
+        "cmp $2, %%rax\n\t"
```
abhinav92003, line 1079: Wouldn't rax be `$1` in the first run?

abhinav92003 (approved):

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -207,7 +207,33 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
     }
```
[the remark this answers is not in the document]
abhinav92003: Ignore.

derekbruening:
> x64 drcachesim.threads failure looks like a flake: will file new issue and re-run if not already there (not seeing on quick search)

This was already filed: #4167 "Invalid entry type 23 before a bundle" but here on drcachesim.threads instead of coherence

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -1268,9 +1275,15 @@ raw2trace_t::append_bb_entries(raw2trace_thread_data_t *tdata,
...
+            addr_t instr_pc = reinterpret_cast<addr_t>(orig_pc);
             if (tdata->rseq_past_end_) {
-                error =
-                    adjust_and_emit_rseq_buffer(tdata, reinterpret_cast<addr_t>(orig_pc));
+                error = adjust_and_emit_rseq_buffer(tdata, instr_pc);
+                if (!error.empty())
+                    return error;
+            } else if (instr_pc < tdata->rseq_start_pc_ ||
+                       instr_pc >= tdata->rseq_end_pc_) {
+                log(4, "Hit rseq instrumented exit to 0x%zx\n", orig_pc);
```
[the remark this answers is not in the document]
derekbruening: Done.

suite/tests/linux/rseq.c
derekbruening: Actually most tests do not have a commits var: just the side exit tests.  I suppose it could be added here.  Done.

suite/tests/linux/rseq.c
derekbruening: Yes, it's a jne, not a je (b/c copied from the inverted one).  Changed to $1 and je to be more straightforward.

suite/tests/linux/rseq.c
abhinav92003: I'm trying to understand whether this can be a source of transparency violations: do rseq aborts from signals become significantly more likely during instrumented runs? Is it likely to have a degenerate case where the rseq region is never able to complete instrumented execution, because of repeated rseq aborts caused by signals during instrumented run?

Issue #5984, opened 2023-04-19
Invariant checker false positive for back-to-back signals


abhinav92003:
The drmemtrace invariant checker currently produces a false positive "Signal handler return point incorrect" error on back-to-back signals without any intervening instructions, nested within another signal handler.

Example case coming up in a PR in a few minutes.

PR #5985, opened 2023-04-19
i#5984: Fix trace invariant false positive on back-to-back signals

abhinav92003:
Moves tracking for previous kernel_event marker values and previous pre-signal instrs at the kernel_event and kernel_xfer markers. Previously we did some part of this book-keeping at instrs after the kernel_xfer marker, which would be erroneously skipped when there are back-to-back signals without any intervening instrs.

Adds tracking for last seen instr at each signal depth. This is required to correctly find the pre-signal instr for the second signal in a back-to-back signal pair. Without this, the pre-signal instr for the second signal is erroneously assumed to be whatever the last instr is, which may be some instr in the first signal's body. This also obviates the need to track the boolean for saw_kernel_xfer_after_prev_instr_, which was to detect nested signals Without any intervening instr.

Adds a test case that verifies the scenario with back-to-back signals without any intervening instr.

Fixes: #5984

### DIFF
--- a/clients/drcachesim/tests/invariant_checker_test.cpp
+++ b/clients/drcachesim/tests/invariant_checker_test.cpp
@@ -89,8 +89,9 @@ run_checker(const std::vector<memref_t> &memrefs, bool expect_error,
                 return false;
             }
         } else if (!checker.errors.empty()) {
-            for (const auto &error : checker.errors) {
-                std::cerr << "Unexpected error: " << error << "\n";
+            for (int i = 0; i < checker.errors.size(); ++i) {
+                std::cerr << "Unexpected error: " << checker.errors[i]
+                          << " at ref: " << checker.error_refs[i] << "\n";
             }
             return false;
         }
@@ -328,6 +329,33 @@ check_kernel_xfer()
         if (!run_checker(memrefs, false))
             return false;
     }
+    // Multiple nested signals without any intervening instr between the nested
+    // signals.
+    {
+        std::vector<memref_t> memrefs = {
+            gen_instr(1, 1),
+            gen_marker(1, TRACE_MARKER_TYPE_KERNEL_EVENT, 2),
+            gen_instr(1, 101),
+            // First nested signal.
+            gen_marker(1, TRACE_MARKER_TYPE_KERNEL_EVENT, 102),
+            gen_instr(1, 201),
+            // XXX: This marker value is actually not guaranteed, yet the checker
+            // requires it and the view tool prints it.
+            gen_marker(1, TRACE_MARKER_TYPE_KERNEL_XFER, 202),
+            // No intervening instr here.
+            // Second nested signal.
+            gen_marker(1, TRACE_MARKER_TYPE_KERNEL_EVENT, 102),
+            gen_instr(1, 201),
+            // XXX: This marker value is actually not guaranteed, yet the checker
+            // requires it and the view tool prints it.
+            gen_marker(1, TRACE_MARKER_TYPE_KERNEL_XFER, 202),
+            gen_instr(1, 102),
+            gen_marker(1, TRACE_MARKER_TYPE_KERNEL_XFER, 103),
+            gen_instr(1, 2),
+        };
+        if (!run_checker(memrefs, false))
+            return false;
+    }
     // Fail to return to recorded interruption point.
     {
         std::vector<memref_t> memrefs = {
@@ -521,7 +549,7 @@ check_duplicate_syscall_with_same_pc()
             gen_instr_encoded(ADDR + 2, { 0x0f, 0x05 }), // 0x7fcf3b9dd9eb: 0f 05 syscall
 #    elif defined(ARM_64)
             gen_instr_encoded(ADDR, 0xd4000001,
-                              2), // 0x7fcf3b9dd9e9: 0xd4000001 svc #0x0
+                              2),          // 0x7fcf3b9dd9e9: 0xd4000001 svc #0x0
             gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
             gen_marker(1, TRACE_MARKER_TYPE_CPU_ID, 3),
             gen_instr_encoded(ADDR + 4, 0xd4000001,
--- a/clients/drcachesim/tools/invariant_checker.cpp
+++ b/clients/drcachesim/tools/invariant_checker.cpp
@@ -36,6 +36,7 @@
 #include <algorithm>
 #include <iostream>
 #include <string.h>
+#include <cassert>
 
 analysis_tool_t *
 invariant_checker_create(bool offline, unsigned int verbose)
@@ -439,25 +440,28 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
         // Ensure signal handlers return to the interruption point.
         if (shard->prev_xfer_marker_.marker.marker_type ==
             TRACE_MARKER_TYPE_KERNEL_XFER) {
+            // We popped some values in last_xfer_int_pc_ and
+            // last_pre_signal_instr_ on the last TRACE_MARKER_TYPE_KERNEL_XFER
+            // marker.
             report_if_false(
                 shard,
-                ((memref.instr.addr == shard->prev_xfer_int_pc_.top() ||
+                ((memref.instr.addr == shard->last_xfer_int_pc_ ||
                   // DR hands us a different address for sysenter than the
                   // resumption point.
-                  shard->pre_signal_instr_.top().instr.type ==
+                  shard->last_pre_signal_instr_.instr.type ==
                       TRACE_TYPE_INSTR_SYSENTER) &&
                  (
                      // Skip pre_signal_instr_ check if there was no such instr.
-                     shard->pre_signal_instr_.top().instr.addr == 0 ||
-                     memref.instr.addr == shard->pre_signal_instr_.top().instr.addr ||
+                     shard->last_pre_signal_instr_.instr.addr == 0 ||
+                     memref.instr.addr == shard->last_pre_signal_instr_.instr.addr ||
                      // Asynch will go to the subsequent instr.
                      memref.instr.addr ==
-                         shard->pre_signal_instr_.top().instr.addr +
-                             shard->pre_signal_instr_.top().instr.size ||
+                         shard->last_pre_signal_instr_.instr.addr +
+                             shard->last_pre_signal_instr_.instr.size ||
                      // Too hard to figure out branch targets.  We have the
-                     // prev_xfer_int_pc_ though.
-                     type_is_instr_branch(shard->pre_signal_instr_.top().instr.type) ||
-                     shard->pre_signal_instr_.top().instr.type ==
+                     // last_xfer_int_pc_ though.
+                     type_is_instr_branch(shard->last_pre_signal_instr_.instr.type) ||
+                     shard->last_pre_signal_instr_.instr.type ==
                          TRACE_TYPE_INSTR_SYSENTER)) ||
                     // Nested signal.  XXX: This only works for our annotated test
                     // signal_invariants where we know shard->app_handler_pc_.
@@ -469,17 +473,13 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                     TESTANY(OFFLINE_FILE_TYPE_FILTERED | OFFLINE_FILE_TYPE_IFILTERED,
                             shard->file_type_),
                 "Signal handler return point incorrect");
-            // We assume paired signal entry-exit (so no longjmp and no rseq
-            // inside signal handlers).
-            shard->prev_xfer_int_pc_.pop();
-            shard->pre_signal_instr_.pop();
         }
 #endif
         shard->prev_instr_ = memref;
+        shard->prev_instr_at_depth_[shard->cur_signal_depth_] = memref;
         if (shard->prev_instr_decoded_ != nullptr)
             instr_free(GLOBAL_DCONTEXT, shard->prev_instr_decoded_.get());
         shard->prev_instr_decoded_ = std::move(cur_instr_decoded);
-        shard->saw_kernel_xfer_after_prev_instr_ = false;
         // Clear prev_xfer_marker_ on an instr (not a memref which could come between an
         // instr and a kernel-mediated far-away instr) to ensure it's *immediately*
         // prior (i#3937).
@@ -519,27 +519,41 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                       << std::hex << memref.marker.marker_value << std::dec << "\n";
         }
 #ifdef UNIX
-        if (memref.marker.marker_type == TRACE_MARKER_TYPE_KERNEL_EVENT)
-            shard->prev_xfer_int_pc_.push(memref.marker.marker_value);
         report_if_false(shard, memref.marker.marker_value != 0,
                         "Kernel event marker value missing");
-        if (memref.marker.marker_type == TRACE_MARKER_TYPE_KERNEL_EVENT &&
-            // XXX: Handle the back-to-back signals case where the second
-            // signal arrives just after the return from the first without
-            // any intervening instrs. The return point of the second one
-            // would be the pc in the kernel xfer marker of the first.
-            shard->prev_xfer_marker_.marker.marker_type !=
-                TRACE_MARKER_TYPE_KERNEL_XFER) {
-            if (shard->saw_kernel_xfer_after_prev_instr_) {
-                // We have nested signals without an intervening app instr.
-                // Push an empty instr to mean that this shouldn't be used.
-                shard->pre_signal_instr_.push({});
+        if (memref.marker.marker_type == TRACE_MARKER_TYPE_KERNEL_XFER) {
+            shard->prev_instr_at_depth_[shard->cur_signal_depth_] = {};
+            --shard->cur_signal_depth_;
+            assert(shard->cur_signal_depth_ >= 0);
+            // We assume paired signal entry-exit (so no longjmp and no rseq
+            // inside signal handlers). Also, we must have seen a prior
+            // TRACE_MARKER_TYPE_KERNEL_EVENT. This is an invariant because traces
+            // do not start already in a signal handler.
+            if (shard->prev_xfer_int_pc_.empty() || shard->pre_signal_instr_.empty()) {
+                report_if_false(shard, false, "Found signal exit without signal entry");
+                // Try to continue.
+                shard->last_xfer_int_pc_ = 0;
+                shard->last_pre_signal_instr_ = {};
             } else {
-                shard->saw_kernel_xfer_after_prev_instr_ = true;
-                // If there was a kernel xfer marker at the very beginning
-                // of the trace, we may still push an empty instr here.
-                shard->pre_signal_instr_.push(shard->prev_instr_);
+                shard->last_xfer_int_pc_ = shard->prev_xfer_int_pc_.top();
+                shard->prev_xfer_int_pc_.pop();
+                shard->last_pre_signal_instr_ = shard->pre_signal_instr_.top();
+                shard->pre_signal_instr_.pop();
             }
+        }
+        if (memref.marker.marker_type == TRACE_MARKER_TYPE_KERNEL_EVENT) {
+            ++shard->cur_signal_depth_;
+            assert(shard->cur_signal_depth_ < MAX_SIGNAL_DEPTH);
+            shard->prev_instr_at_depth_[shard->cur_signal_depth_] = {};
+            shard->prev_xfer_int_pc_.push(memref.marker.marker_value);
+            // The last instr in the context immediately outer to this signal may
+            // be {} in some cases:
+            // - for nested signals without any intervening instr
+            // - if there's a signal at the very beginning of the trace
+            // In both these cases the empty instr implies that it should not
+            // be used for the pre_signal_instr check above.
+            shard->pre_signal_instr_.push(
+                shard->prev_instr_at_depth_[shard->cur_signal_depth_ - 1]);
         }
 #endif
         shard->prev_xfer_marker_ = memref;
--- a/clients/drcachesim/tools/invariant_checker.h
+++ b/clients/drcachesim/tools/invariant_checker.h
@@ -86,6 +86,11 @@ protected:
         memtrace_stream_t *stream = nullptr;
         memref_t prev_entry_ = {};
         memref_t prev_instr_ = {};
+
+#define MAX_SIGNAL_DEPTH 100
+        memref_t prev_instr_at_depth_[MAX_SIGNAL_DEPTH] = {};
+        int cur_signal_depth_ = 0;
+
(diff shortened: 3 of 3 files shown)


abhinav92003:
AArch64 failure is #5635.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -328,6 +331,73 @@ check_kernel_xfer()
...
+        };
+        if (!run_checker(memrefs, false))
+            return false;
+    }
+    // Trace starts in a signal.
+    {
+        std::vector<memref_t> memrefs = {
+            // Already inside the first signal.
+            gen_instr(1, 11),
+            // XXX: This marker value is actually not guaranteed, yet the checker
```
derekbruening, line 367: What does "not guaranteed" mean?  Why is this different from any other sigreturn/etc.?  This marker value is required to be correct and the checker should be checking it.

Ditto below.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -328,6 +331,73 @@ check_kernel_xfer()
...
+            // requires it and the view tool prints it.
+            gen_marker(1, TRACE_MARKER_TYPE_KERNEL_XFER, 12),
+            // No intervening instr here. Should skip pre-signal instr check on
+            // return; this is a special case as it would require *removing* the
+            // pc = 1 instr from pre_signal_instr_ as it was not in this newly
```
derekbruening, line 389: ?  There is no pc=1 instr here: you mean pc=11?

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -370,6 +440,8 @@ check_rseq()
             gen_marker(1, TRACE_MARKER_TYPE_RSEQ_ABORT, 2),
             gen_marker(1, TRACE_MARKER_TYPE_KERNEL_EVENT, 2),
             gen_marker(1, TRACE_MARKER_TYPE_KERNEL_EVENT, 4),
+            gen_instr(1, 10),
```
derekbruening, line 443: Does the checker fail without these additions?  What if the trace did actually end before these were added?

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -439,25 +454,36 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+                     memref.instr.addr == shard->last_xfer_int_pc_ ||
+                     // DR hands us a different address for sysenter than the
+                     // resumption point.
+                     shard->last_pre_signal_instr_.instr.type ==
+                         TRACE_TYPE_INSTR_SYSENTER) &&
                  (
-                     // Skip pre_signal_instr_ check if there was no such instr.
-                     shard->pre_signal_instr_.top().instr.addr == 0 ||
-                     memref.instr.addr == shard->pre_signal_instr_.top().instr.addr ||
+                     // Skip pre-signal instr check if there was no such instr.
```
derekbruening, line 472: Should we split these into two different error strings: the post==marker (above) with one error string and the pre==post (below) with a different one?  May make it easier to analyze/group reported invariants, and split the complex boolean expression here up too.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -439,25 +454,36 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
-                  // DR hands us a different address for sysenter than the
-                  // resumption point.
-                  shard->pre_signal_instr_.top().instr.type ==
-                      TRACE_TYPE_INSTR_SYSENTER) &&
+                ((
+                     // Skip this check if we did not see the corresponding
+                     // kernel_event marker in the trace because the trace
+                     // started mid-signal.
+                     shard->last_xfer_int_pc_ == 0 ||
+                     memref.instr.addr == shard->last_xfer_int_pc_ ||
```
derekbruening, line 466: With the large comments on the other conditions around this, I found it hard to identify this as its own thing rather than part of the condition above.  Maybe put this first, or add a comment saying this is the regular check for post==marker.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -439,25 +454,36 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+                     // handler, verified using last_xfer_int_pc_ above.
+                     shard->last_xfer_aborted_rseq_ ||
+                     memref.instr.addr == shard->last_pre_signal_instr_.instr.addr ||
                      // Asynch will go to the subsequent instr.
                      memref.instr.addr ==
-                         shard->pre_signal_instr_.top().instr.addr +
-                             shard->pre_signal_instr_.top().instr.size ||
+                         shard->last_pre_signal_instr_.instr.addr +
+                             shard->last_pre_signal_instr_.instr.size ||
                      // Too hard to figure out branch targets.  We have the
```
derekbruening, line 483: But we do have branch targets now: it's right there in a shard var, isn't it?

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -469,17 +495,18 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
-            // We assume paired signal entry-exit (so no longjmp and no rseq
-            // inside signal handlers).
-            shard->prev_xfer_int_pc_.pop();
-            shard->pre_signal_instr_.pop();
         }
+        // It is a little inefficient to replace the top instr on the stack everytime.
+        // But we cannot perform this book-keeping using prev_instr_ on a
+        // TRACE_MARKER_TYPE_KERNEL_EVENT marker. If there was no instruction between
+        // two nested signals, we do not want to record any pre-signal instr for the
+        // second signal.
```
derekbruening, line 503: "between two nested signals": there is no test of this added here so I wonder if you mean "between two consecutive signals at the same scope" (i.e., not nested wrt each other).

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -519,26 +550,52 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+                // forget the last instr we saw in what was incorrectly assumed previously
+                // to be the outer-most context.
+                replace_top<memref_t>(shard->pre_signal_instr_, {});
+            } else {
+                shard->last_xfer_int_pc_ = shard->prev_xfer_int_pc_.top();
+                shard->prev_xfer_int_pc_.pop();
+                shard->pre_signal_instr_.pop();
+                // The last instr in the context immediately outer to this signal may
+                // be {} in some cases:
+                // - for nested signals without any intervening instr
```
derekbruening, line 580: See comment in the test: there is no test of this.  Do you mean back-to-back signals at the same scope instead of nested signals?

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -469,17 +495,18 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
                 "Signal handler return point incorrect");
-            // We assume paired signal entry-exit (so no longjmp and no rseq
-            // inside signal handlers).
-            shard->prev_xfer_int_pc_.pop();
-            shard->pre_signal_instr_.pop();
         }
+        // It is a little inefficient to replace the top instr on the stack everytime.
+        // But we cannot perform this book-keeping using prev_instr_ on a
+        // TRACE_MARKER_TYPE_KERNEL_EVENT marker. If there was no instruction between
+        // two nested signals, we do not want to record any pre-signal instr for the
```
derekbruening, line 502: But we do want the pre-signal instr for an instr prior to two consecutive signals at the same scope.  What we want is a prev_instr stack so we get the prev_instr for our scope -- but isn't that what shard->pre_signal_instr is already?  I think I'm confused and not understanding this right: not understanding this replace_top.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -519,26 +550,52 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+                shard->last_xfer_int_pc_ = shard->prev_xfer_int_pc_.top();
+                shard->prev_xfer_int_pc_.pop();
+                shard->pre_signal_instr_.pop();
+                // The last instr in the context immediately outer to this signal may
+                // be {} in some cases:
+                // - for nested signals without any intervening instr
+                // - if there's a signal at the very beginning of the trace
+                // In both these cases the empty instr implies that it should not
+                // be used for the pre-signal instr check.
+                shard->last_pre_signal_instr_ = shard->pre_signal_instr_.top();
```
derekbruening, line 584: Why do some of these 3 stacks do "last=top; pop" while others do "pop; last=top"?  Shouldn't they all be the same wrt whether the top or second slot is what we want?

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -519,26 +550,52 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+                shard->last_xfer_aborted_rseq_ = shard->prev_xfer_aborted_rseq_.top();
+                shard->prev_xfer_aborted_rseq_.pop();
+            }
+        }
+        if (memref.marker.marker_type == TRACE_MARKER_TYPE_KERNEL_EVENT) {
+            // If preceded by an RSEQ abort marker, this is not really a signal.
+            if (shard->prev_entry_.marker.marker_type == TRACE_MARKER_TYPE_RSEQ_ABORT) {
```
derekbruening, line 591: Doesn't this need to first check that prev_entry_.marker.type is a marker?

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -553,6 +610,13 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
     }
 
 #ifdef UNIX
+    if (saw_rseq_abort) {
+        shard->saw_rseq_abort_ = true;
+    } else if (!(memref.marker.type == TRACE_TYPE_MARKER &&
+                 (memref.marker.marker_type == TRACE_MARKER_TYPE_TIMESTAMP ||
+                  memref.marker.marker_type == TRACE_MARKER_TYPE_CPU_ID))) {
```
derekbruening, line 617: What is the logic for clearing on anything except a timestamp/cpuid?

derekbruening (commented):
I think I'm confused about some aspects here.

clients/drcachesim/tests/invariant_checker_test.cpp
abhinav92003: This comment is same as many other similar comments in the test file. I think it points to the fact that a sigreturn is not necessary when coming back from a signal.

clients/drcachesim/tests/invariant_checker_test.cpp
abhinav92003: Ah yes. It should be pc = 11.

clients/drcachesim/tests/invariant_checker_test.cpp
abhinav92003: It does not fail.

As mentioned in the PR description, this is to verify that we don't raise a "signal handler return point incorrect" because of the apparent discontinuity between the post-signal instr and the instr just before the rseq-abort-causing signal.

See the paragraph with "Modified an existing invariant checker unit test to verify this scenario also." in the PR description.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -328,6 +331,73 @@ check_kernel_xfer()
         if (!run_checker(memrefs, false))
             return false;
     }
+    // Multiple nested signals without any intervening instr between the nested
+    // signals.
+    {
+        std::vector<memref_t> memrefs = {
+            gen_instr(1, 1),
+            gen_marker(1, TRACE_MARKER_TYPE_KERNEL_EVENT, 2),
+            gen_instr(1, 101),
```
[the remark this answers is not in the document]
abhinav92003: There's already a test for "Signal before any instr in the trace." and "Nested signals without any intervening instr". I'll add one with both in the same test.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: Do you mean I should change it in this PR itself? Since this PR is already a bit complex, maybe we can do it in a follow up PR.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -519,26 +550,52 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
-                // We have nested signals without an intervening app instr.
-                // Push an empty instr to mean that this shouldn't be used.
-                shard->pre_signal_instr_.push({});
+        if (memref.marker.marker_type == TRACE_MARKER_TYPE_KERNEL_XFER) {
+            // We assume paired signal entry-exit (so no longjmp and no rseq
+            // inside signal handlers).
+            if (shard->prev_xfer_int_pc_.empty() ||
+                shard->pre_signal_instr_.size() == 1 ||
+                shard->prev_xfer_aborted_rseq_.empty()) {
+                // If one is true, all should be so.
```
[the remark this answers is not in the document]
abhinav92003: If the trace starts with KERNEL_EVENT, all three of them will have something pushed. See L596.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: We have test of both cases: see "Nested signals without any intervening instr." and "Multiple nested signals without any intervening instr between the nested".

> I wonder if you mean "between two consecutive signals at the same scope" (i.e., not nested wrt each other).

That's correct. I'll rephrase it to be clearer.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: Here, I do mean nested signals. See "Nested signals without any intervening instr." in the test file.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: > But we do want the pre-signal instr for an instr prior to two consecutive signals at the same scope

That's right. We do do that here. See comment in the "Multiple nested signals without any intervening instr between the nested" test.

> but isn't that what shard->pre_signal_instr is already?

Yes.

> not understanding this replace_top

When we see a new instr, that instr becomes the last instr seen in the current scope. That's what the replacement means.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: last_pre_signal_instr_ needs to get the second-to-top (the instr in the scope that was interrupted by that signal).

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: To detect whether the current signal caused an rseq abort we need to find the following pattern:
```
rseq_abort
kernel_event
kernel_event
```

I found that sometimes the two kernel events are separated by timestamp+cpu.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: That's right. I'll add that.

clients/drcachesim/tests/invariant_checker_test.cpp
derekbruening: OK, so `s/This marker value/This marker/` and `s/guaranteed/guaranteed to be present/` -- though this is a particular sequence in a test so I would just remove this comment altogether.

abhinav92003: Removed.

clients/drcachesim/tests/invariant_checker_test.cpp
abhinav92003: Done.

clients/drcachesim/tests/invariant_checker_test.cpp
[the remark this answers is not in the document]
abhinav92003: Done.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -36,6 +36,7 @@
 #include <algorithm>
 #include <iostream>
 #include <string.h>
+#include <cassert>
```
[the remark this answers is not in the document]
abhinav92003: Done.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: I split the boolean expression to simplify it.

But I think the different error messages are being worked on already in Sahil's PR. So I'll leave it to that.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: Added comment.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -439,25 +454,36 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
-                     // Skip pre_signal_instr_ check if there was no such instr.
-                     shard->pre_signal_instr_.top().instr.addr == 0 ||
-                     memref.instr.addr == shard->pre_signal_instr_.top().instr.addr ||
+                     // Skip pre-signal instr check if there was no such instr.
+                     shard->last_pre_signal_instr_.instr.addr == 0 ||
+                     // Skip pre_signal_instr_ check for signals that caused an rseq
+                     // abort. In this case, control is transferred directly to the abort
+                     // handler, verified using last_xfer_int_pc_ above.
+                     shard->last_xfer_aborted_rseq_ ||
+                     memref.instr.addr == shard->last_pre_signal_instr_.instr.addr ||
```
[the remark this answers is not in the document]
abhinav92003: Added the comment.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: Actually this comment applies to both: nested signals and consecutive signals. I clarified in the comment.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: Done.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: Added a comment to clarify.

clients/drcachesim/tools/invariant_checker.cpp
derekbruening: I don't follow this: what's at the top then?

I'm thinking we should have a single stack with a compound data structure, rather than a bunch of separate stacks which we have to worry about being out of sync.

clients/drcachesim/tools/invariant_checker.cpp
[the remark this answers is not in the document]
derekbruening: See below: a single stack with a compound data structure sounds like the right solution.

clients/drcachesim/tools/invariant_checker.cpp
derekbruening: What we want is shard->cur_scope->prev_instr.  Instead of popping and pushing to store prev_instr, if we do the compound data structure below we can more simply store into the cur scope (== top of scope/context stack)'s field for prev_instr.

Should we go further and get rid of shard->prev_instr in general and always use shard->cur_scope->prev_instr everywhere?

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: > I don't follow this: what's at the top then?

The latest instruction in the current context (current signal).

> I'm thinking we should have a single stack with a compound data structure,

That does seem simpler but it may create some more complexities IMHO: we'd need to look at two entries of the stack for the operation here (since we want the second-to-top for pre_signal_instr_). Also updating the top instr (replace_top) will become a bit more expensive since it's a bigger entry.

> we have to worry about being out of sync.

We push/pop all at the same time always. 

If you feel strongly about this, I can try changing it. But I don't see a clear win in clarity.

derekbruening (commented):
Should have done the last few comments as a review instead of separately.

clients/drcachesim/tools/invariant_checker.cpp
derekbruening: > The latest instruction in the current context (current signal).

But isn't that the current record?  So this is just an ordering problem where it's already set the prev to equal the current?

derekbruening: > We push/pop all at the same time always.

This seems error-prone.

> If you feel strongly about this, I can try changing it. But I don't see a clear win in clarity.

Yes I do think a single stack is a better solution.  The code is simpler and clearer; the invariants are clearer (no more multiple comments and asserts about how all these separate stacks are supposed to be coordinated); changing an aspect of the top of the stack looks simpler (set a field, instead of pop;push).

abhinav92003: Thinking more on this:  I think this can be simplified further. I'll add something like `last_instr_in_cur_context` which can repeatedly be updated, and then keep all context for the current signal in the same entry on one stack.

abhinav92003: Done.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: I switched to using a separate var `last_instr_in_cur_context_` that keeps getting updated.

abhinav92003: > Should we go further and get rid of shard->prev_instr in general and always use shard->cur_scope->prev_instr everywhere?

That'd get a little complex. `shard->cur_scope->prev_instr ` (or shard->signal_stack_.top().pre_signal_instr in the code now) may not be set. It can be `{ }` (to denote absence of any pre-signal instr in that context). So we'll need to search the whole signal_stack_ from the top to find the first entry with a valid pre_signal_instr.

clients/drcachesim/tools/invariant_checker.cpp
derekbruening: Right, I think I was confused before, this makes sense.

derekbruening (approved):
Thanks for making the change to a single stack: IMHO it's nicer now.  LGTM.

clients/drcachesim/tools/invariant_checker.cpp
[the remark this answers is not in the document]
abhinav92003: Done.

abhinav92003:
> Thanks for making the change to a single stack: IMHO it's nicer now. LGTM.

Yes, thanks for the suggestion. Combined with the separate last_instr_in_cur_context, it does become much simpler

PR #5992, opened 2023-04-21
Add Native aarch64 Test Workflow

sm-powell:
Add workflow to test on physical aarch64 hardware.


spcaipers-arm (commented):
It looks good to me, but I think the team with internal knowledge should approve it.

PR #5993, opened 2023-04-21
Update offset field name in encode API doc

abhinav92003:
Updates the documentation for various encode APIs to correctly state that
the offset field of instr_t is used for instr_t target offset computation, instead
of the note field which was used prior to #5791.


derekbruening (approved):

abhinav92003:
run arm tests

PR #6014, opened 2023-04-28
 i#3544 riscv64: Implement some base instructions

shiptux:
- Added SYSNUM_REG for RISC-V64
- Implemented atomic_add
- Implemented dynamorio_sigreturn
- Implemented dynamorio_sys_exit
- Implemented call_switch_stack

Notice: Although some base instructions have been implemented for RISC-V64, there are still many unimplemented pieces in Dynamorio for this architecture, and it cannot be tested via QEMU then.

Issue: #3544


core/arch/asm_defines.asm
```diff
@@ -563,6 +563,7 @@ ASSUME fs:_DATA @N@\
 # define ARG6 REG_R15
 # define ARG7 REG_R16
 # define ARG8 REG_R17
+# define SYSNUM_REG REG_R17
```
abhinav92003, line 566: Based on https://man7.org/linux/man-pages/man2/syscall.2.html a7 holds the system call number.

core/arch/riscv64/riscv64.asm
```diff
@@ -59,7 +59,25 @@ GLOBAL_LABEL(cpuid_supported:)
...
 GLOBAL_LABEL(call_switch_stack:)
-/* FIXME i#3544: Not implemented */
+        addi    sp, sp, -24 /* Init the stack */
+        sd      ra, 16 (sp)
+        sd      s0, 8 (sp)
+        sd      s1, 0 (sp)
+        /* Check mutex_to_free. */
+        beqz     ARG4, call_dispatch_alt_stack_no_free
+        /* Release the mutex. */
+        addi     ARG4,x0,0
```
abhinav92003, line 69: nit: formatting: add space between operands.

core/arch/riscv64/riscv64.asm
```diff
@@ -59,7 +59,25 @@ GLOBAL_LABEL(cpuid_supported:)
...
+call_dispatch_alt_stack_no_free:
+        mv       s0, ARG5
+        addi     t0, sp, 0    /* Save the current stack to temporary register */
+        addi     sp, ARG2, 0 /* Move the stack */
```
abhinav92003, line 73: nit: formatting: align comment with comment above.

core/arch/asm_defines.asm
```diff
@@ -563,6 +563,7 @@ ASSUME fs:_DATA @N@\
 # define ARG6 REG_R15
 # define ARG7 REG_R16
 # define ARG8 REG_R17
```
abhinav92003, line 565: For L544-548: s/Calee/Callee/

core/arch/riscv64/riscv64.asm
abhinav92003, line 69: Comparing with the AArch64 version, we're storing zero in [ARG4] (`str      wzr, [ARG4]`) whereas here we seem to be storing zero in `ARG4`.

core/arch/riscv64/riscv64.asm
```diff
@@ -59,7 +59,25 @@ GLOBAL_LABEL(cpuid_supported:)
...
+        jr       ARG3
+        addi     sp, t0, 0
```
abhinav92003, line 75: Adding comments similar to the AArch64 version would be helpful

core/arch/riscv64/riscv64.asm
```diff
@@ -59,7 +59,25 @@ GLOBAL_LABEL(cpuid_supported:)
...
+        beqz     s0, GLOBAL_LABEL(unexpected_return)
+        ld       s1, 0 (sp)
+        ld       s2, 8 (sp)
```
abhinav92003, line 78: Shouldn't we be loading s0 here? Since that's what's we stored at `8 (sp)` above

abhinav92003 (changes requested):

abhinav92003:
Also, s/instrutions/instructions/ in the PR title.

core/arch/riscv64/riscv64.asm
```diff
@@ -59,7 +59,25 @@ GLOBAL_LABEL(cpuid_supported:)
  */
         DECLARE_FUNC(call_switch_stack)
 GLOBAL_LABEL(call_switch_stack:)
-/* FIXME i#3544: Not implemented */
+        addi    sp, sp, -24 /* Init the stack */
+        sd      ra, 16 (sp)
+        sd      s0, 8 (sp)
+        sd      s1, 0 (sp)
```
derekbruening, line 65: The alignment of the start of the operands on the 4 lines doesn't match our code style which wants them on column 17

core/arch/asm_defines.asm
shiptux: You are right. But a7 is the abi name, a7 equal x17. Based on https://github.com/DynamoRIO/dynamorio/blob/master/core/arch/asm_defines.asm#L399 and https://github.com/DynamoRIO/dynamorio/blob/master/core/arch/asm_defines.asm#L565. REG_R17 exactly is a7.

core/arch/riscv64/riscv64.asm
shiptux: done

core/arch/riscv64/riscv64.asm
shiptux: Done.

core/arch/asm_defines.asm
shiptux: done

core/arch/riscv64/riscv64.asm
shiptux: Done

core/arch/riscv64/riscv64.asm
shiptux: X0 register is a zero register in RISC-V. 
So add a imm zero and a zero register equals zero.

shiptux:
```
addi      ARG4, x0 ,0
                0  =  0 + 0
```

core/arch/riscv64/riscv64.asm
shiptux: This is a small mistake. Fixed.

core/arch/riscv64/riscv64.asm
shiptux: Thanks for your advice. I added more comment.

shiptux:
Is SYSNUM_REG only used in aarch64? But it seems that it is defined in all arches.
https://github.com/DynamoRIO/dynamorio/blob/master/core/arch/asm_defines.asm#L133
@derekbruening @abhinav92003

core/arch/riscv64/riscv64.asm
abhinav92003: I mean: do we need to store a zero in the address pointed by the ARG4 reg or in the ARG4 reg itself?

abhinav92003:
> Is SYSNUM_REG only used in aarch64? But it seems that it is defined in all arches. https://github.com/DynamoRIO/dynamorio/blob/master/core/arch/asm_defines.asm#L133 @derekbruening @abhinav92003

Does seem that some of that `#else` part is intended only for AArch64.

abhinav92003:
> Is SYSNUM_REG only used in aarch64? But it seems that it is defined in all arches. https://github.com/DynamoRIO/dynamorio/blob/master/core/arch/asm_defines.asm#L133 @derekbruening @abhinav92003

Is this blocking this PR? If not, we can fix it separately; can you raise a new issue for it?

core/arch/riscv64/riscv64.asm
shiptux: Ok. In the new version, I sd a zero in the address pointed by ARG4 (with no offset)

shiptux:
> > Is SYSNUM_REG only used in aarch64? But it seems that it is defined in all arches. https://github.com/DynamoRIO/dynamorio/blob/master/core/arch/asm_defines.asm#L133 @derekbruening @abhinav92003
> 
> Is this blocking this PR? If not, we can fix it separately; can you raise a new issue for it?

This makes dynamorio fail to compile on RISC-V. See the CI log in this PR.

abhinav92003:
> > > Is SYSNUM_REG only used in aarch64? But it seems that it is defined in all arches. https://github.com/DynamoRIO/dynamorio/blob/master/core/arch/asm_defines.asm#L133 @derekbruening @abhinav92003
> > 
> > 
> > Is this blocking this PR? If not, we can fix it separately; can you raise a new issue for it?
> 
> This makes dynamorio fail to compile on RISC-V. See the CI log in this PR.

Oh I didn't notice that before since it's a warning `asm_defines.asm:566: warning: "SYSNUM_REG" redefined` that doesn't show up as a CI failure. The build failure seems to be due to a recent commit on this branch: ` Error: illegal operands 'ld s2 0(sp)'`.

Regardless, I agree we should fix the warning. Like I said above, some of that `#else` part does seem targeted to only AArch64. Can you maybe add another conditional define there for now? And also move the risc-v definition to the same part of code? (We might want to refactor existing x86 code to also use a SYSNUM_REG constant to be consistent across all archs, but we can do it independently of this PR so as to not block it).

shiptux:
> > > > Is SYSNUM_REG only used in aarch64? But it seems that it is defined in all arches. https://github.com/DynamoRIO/dynamorio/blob/master/core/arch/asm_defines.asm#L133 @derekbruening @abhinav92003
> > > 
> > > 
> > > Is this blocking this PR? If not, we can fix it separately; can you raise a new issue for it?
> > 
> > 
> > This makes dynamorio fail to compile on RISC-V. See the CI log in this PR.
> 
> Oh I didn't notice that before since it's a warning `asm_defines.asm:566: warning: "SYSNUM_REG" redefined` that doesn't show up as a CI failure. The build failure seems to be due to a recent commit on this branch: ` Error: illegal operands 'ld s2 0(sp)'`.
> 
> Regardless, I agree we should fix the warning. Like I said above, some of that `#else` part does seem targeted to only AArch64. Can you maybe add another conditional define there for now? And also move the risc-v definition to the same part of code? (We might want to refactor existing x86 code to also use a SYSNUM_REG constant to be consistent across all archs, but we can do it independently of this PR so as to not block it).

Ok. I fixed the CI error. It is not blocked by SYSNUM_REG.
I think move the define in another PR will be better.

abhinav92003 (approved):

abhinav92003:
> Ok. I fixed the CI error. It is not blocked by SYSNUM_REG. I think move the define in another PR will be better.

Okay. I raised #6027 for the cleanup.

derekbruening:
Looks like @abhinav92003 already reviewed and approved: not sure why another review was requested, @shiptux ?

shiptux:
> Looks like @abhinav92003 already reviewed and approved: not sure why another review was requested, @shiptux ?

Sorry for requested that.
I saw there have two reviewer, I mistaken for should both reviewer approval.

Sorry to bother you.

shiptux:
@abhinav92003 @derekbruening It seems I don't have permission to merge this PR. (Need to merge by the project collaborators.) Could anyone help me?

abhinav92003:
> @abhinav92003 @derekbruening It seems I don't have permission to merge this PR. (Need to merge by the project collaborators.) Could anyone help me?

I sent you an invite for Committers access to the repo. Once you accept it, you should be able to submit.

PR #6016, opened 2023-05-01
i#5843 scheduler: Implement speculation with nops

derekbruening:
Implements initial speculation support, supplying nops.  Speculation is separated into its own class where we can fill in different strategies in the future.

The start_speculation() function takes a flag controlling whether the scheduler queues up the current record and re-returns it as the first record after speculation stops.  This is often what a simulator wants as it has to read the instruction record following a branch to determine whether it is on the wrong path, and it would like to resume with that already-read instruction after speculation.

Adds a unit test.

Issue: #5843

### DIFF
--- a/clients/drcachesim/scheduler/scheduler.h
+++ b/clients/drcachesim/scheduler/scheduler.h
@@ -51,6 +51,7 @@
 #include "memtrace_stream.h"
 #include "reader.h"
 #include "record_file_reader.h"
+#include "speculator.h"
 #include "utils.h"
 
 namespace dynamorio {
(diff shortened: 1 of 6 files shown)


clients/drcachesim/scheduler/scheduler.h
```diff
@@ -480,16 +481,28 @@ template <typename RecordType, typename ReaderType> class scheduler_tmpl_t {
...
-         * Stops speculative execution and resumes the regular stream of records
-         * from the point at which the prior start_speculation call was made.
-         * Returns STATUS_INVALID if there was no prior start_speculation() call.
+         * Stops speculative execution and resumes the regular stream of records from
+         * the point at which the most distant prior start_speculation() call without an
+         * intervening stop_speculation() call was made (either repeating the current
+         * record at that time, if "true" was passed for "queue_current_record" to
+         * start_speculation(), or continuing on the subsequent record if "false" was
+         * passed).  Returns #dynamorio::drmemtrace::scheduler_tmpl_t::STATUS_INVALID if
+         * there was no prior start_speculation() call.
```
abhinav92003, line 505: or if stop_speculation was already called.

clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -796,6 +796,126 @@ test_synthetic()
...
+        case 4:
+            assert(type_is_instr(memref.instr.type));
+            assert(memref.instr.addr == 2);
+            // We realize now that we mispredicted that the branch would be taken.
+            // We ask to queue this record for post-speculation.
+            stream->start_speculation(100, true);
+            break;
+        case 5:
+            // We should now see nops from the speculator.
+            assert(type_is_instr(memref.instr.type));
```
abhinav92003, line 868: This doesn't verify that it's a nop. Maybe we can have a hard-coded check here for the encoding.

clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -796,6 +796,126 @@ test_synthetic()
...
+        case 8:
+            assert(type_is_instr(memref.instr.type));
+            assert(memref.instr.addr == 3);
+            break;
+        case 9:
+            assert(type_is_instr(memref.instr.type));
+            assert(memref.instr.addr == 4);
+            // We realize now that we mispredicted that the branch would be taken.
+            // This time we do *not* ask to queue this record for post-speculation.
+            stream->start_speculation(200, false);
```
abhinav92003, line 897: Maybe also test multiple start_speculation calls to confirm that PC is changed after each?

abhinav92003 (approved):

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: I tried linking in the decoder to decode and run instr_is_nop() and that worked w/o too much effort so I'll go with that.

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: Done.

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: Maybe pulling in the decoder was a mistake: dealing with Windows toolchain issues due to that...

Issue #5641, opened 2022-09-08
HANG in linux.sigmask-noalarm test


derekbruening:
Happened once on 64-bit GA CI:
https://github.com/DynamoRIO/dynamorio/runs/8241241189?check_suite_focus=true
```
155/431 Test  #91: code_api|linux.sigmask-noalarm ...............................***Failed  Required regular expression not found. Regex=[^sending 10
in handler for signal 10
sending 28 with value
in handler for signal 28 from -1 value 0xdeadbeef
in handler for signal 12
init thread now inside handler: setting up itimer
done with itimer; exiting
all done
$
] 90.00 sec
sending 10
in handler for signal 10
sending 28 with value
in handler for signal 28 from -1 value 0xdeadbeef
in handler for signal 12
init thread now inside handler: setting up itimer
done with itimer; exiting
```

PR #6021, opened 2023-05-02
i#5858: Fix drmemtrace docs location in menu

### LINKED ISSUE #5858: Docs: tracing framework menu entry in wrong place
The online html docs somehow have the `Tracing and Analysis Framework` at the top level listed under License, rather than a sub-menu entry of `Available Tools`.

derekbruening:
The online html docs ended up with the Tracing and Analysis Framework at the top level listed under License, rather than a sub-menu entry of Available Tools, due to PR #5685 which moved drcachesim.dox.in into a docs/ subdir and now the cmake glob doesn't find it.

We fix that by extending the glob.

Tested with a local docs build.

Fixes #5858


johnfxgalea (approved):
LGTM but I think using GLOB_RECURSE would be a better solution in the long run.

derekbruening:
Ah didn't know there was a GLOB_RECURSE: switching to that.

derekbruening:
x86-32 failure is #5641 sigmask-noalarm timeout

Issue #6027, opened 2023-05-04
Cleanup: define asm SYSNUM_REG for all arch 


abhinav92003:
Reported by @shiptux in https://github.com/DynamoRIO/dynamorio/pull/6014#issuecomment-1533203462:

The asm SYSNUM_REG constant is defined too broadly at https://github.com/DynamoRIO/dynamorio/blob/6cb94a8ad63892bf8c3aca4280a5ef512c22a27a/core/arch/asm_defines.asm#L163. It should be defined only for AArch64.

Further, we should define SYSNUM_REG for all envs and refactor existing asm code to use it.

PR #6028, opened 2023-05-04
i#6027: define w8 only for AArch64

shiptux:
The asm SYSNUM_REG w8 is currently defined too broadly. 
This commit move it to only defined in AArch64

Issue: #6027


core/arch/asm_defines.asm
```diff
@@ -376,6 +374,7 @@ ASSUME fs:_DATA @N@\
 # define REG_R10 x10
 # define REG_R11 x11
 # define REG_R12 x12
+# define SYSNUM_REG w8
```
abhinav92003, line 377: This seems to override the definition for `defined(MACOS) && defined(AARCH64)` at L133. Perhaps move that here as well?

abhinav92003 (changes requested):

core/arch/asm_defines.asm
shiptux: I move it here, too.

abhinav92003 (approved):

shiptux:
```
This pull request can be automatically merged by project collaborators
Only those with [write access](https://docs.github.com/articles/what-are-the-different-access-permissions) to this repository can merge pull requests
```
@abhinav92003 It Seems I don't have permission to merge it. Could you please help me?

PR #6029, opened 2023-05-04
i#5843 scheduler: Add input locks

### LINKED ISSUE #5843: Create simulator interface which provides thread scheduling + speculative fetching

derekbruening:
Adds a lock for each input to enforce missing synchronization during scheduling decisions.

Fixes a bug with the existing scheduler lock.

Adds a multi-threaded test.

Tested a similar multi-threaded test under ThreadSanitizer which now reports no races (it did before these code changes).

Fixes #5843


clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -811,6 +812,53 @@ test_synthetic()
...
+            std::this_thread::yield();
+            continue;
+        }
+        assert(status == scheduler_t::STATUS_OK);
+    }
+}
+#endif
+
+static void
+test_synthetic_multi_threaded(const char *testdir)
```
abhinav92003, line 832: Does this test fail without the fixes?

PR #6030, opened 2023-05-04
Rename new aarch64 native Actions worfklow

derekbruening:
The new aarch64 Github Actions workflow's official name was "Explore-GitHub-Actions".  We rename it to "aarch64-native" here.

### DIFF
--- a/.github/workflows/ci-aarch64-native.yml
+++ b/.github/workflows/ci-aarch64-native.yml
@@ -46,7 +46,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  Explore-GitHub-Actions:
+  aarch64-native:
     runs-on: [self-hosted, linux, ARM64]
     steps:
       - name: Check out repository code


AssadHashmi:
I'm happy with this name change. Just two questions (possibly for @sm-powell as well):
- Why isn't there a `Required` label for the check?
- The `Explore-GitHub-Actions` check is still in the checklist. Does that need to be removed?

sm-powell:
Thanks for catching this naming, it is a leftover from the example workflow I used as a base.

AssadHashmi (approved):

derekbruening:
> I'm happy with this name change. Just two questions (possibly for @sm-powell as well):
> 
> * Why isn't there a `Required` label for the check?
> * The `Explore-GitHub-Actions` check is still in the checklist. Does that need to be removed?

We can't make a check required until it's run at least once so we have to commit this first and then update the required check list.

derekbruening:
> > I'm happy with this name change. Just two questions (possibly for @sm-powell as well):
> > 
> > * Why isn't there a `Required` label for the check?
> > * The `Explore-GitHub-Actions` check is still in the checklist. Does that need to be removed?
> 
> We can't make a check required until it's run at least once so we have to commit this first and then update the required check list.

(Another route is to do a manual run from this branch to have a finished run which would allow the update)

PR #6034, opened 2023-05-07
removed "docs/" prefix from drmemory.org/docs links

Pavel-Durov:
Removed "docs/" prefix as was suggested by [derekbruening](https://github.com/derekbruening)
Related  [PR](https://github.com/DynamoRIO/dynamorio.github.io/pull/37)


derekbruening (approved):
Thank you for the fix.

PR #6036, opened 2023-05-08
i#5843 scheduler: Add record + replay

derekbruening:
Adds recording of the dynamic schedule and replay of that recording.

Each thread segment is stored in an in-memory structure with the data written out at the end to a specified ostream.

Replay uses a new mapping mode MAP_AS_PREVIOUSLY.  It reads from an istream into memory, under the assumption that the schedule is small enough to fit in memory.  The file format is an archive, however, to future-proof it in case we need to read the cores incrementally in parallel.  The record format has different needs from the existing trace-recording schedule and so is a new structure.

Adds archive_istream.h and extends zipfile_istream.h to support reading arbitrary zip files and advancing to specific subfiles.

The schedule data includes timestamps for more accurate re-creation of the original inter-output-stream context switch timing, with an output stream waiting if its timestamp gets ahead of other outputs.

Simultaneous reading of one input by multiple outputs is not supported.  If one output stream gets ahead of another on the same input stream, it has to wait for the other to catch up (which it would do anyway based on the timestamp).

Waiting for another output currently has no kernel wait state support: outputs are forced to spin and poll.

Regions of interest are handled with special skip entries in the recorded schedule.  The synthetic records added during regular scheduling are duplicated on replay.

Adds multiple unit tests.

A proprietary test setup with a real application trace was run under ThreadSanitizer 500x to check for races.

Issue: #5843


clients/drcachesim/common/zipfile_istream.h
```diff
@@ -103,25 +104,42 @@ class zipfile_istreambuf_t : public std::basic_streambuf<char, std::char_traits<
         }
         return gptr() - eback();
     }
+    std::string
+    open_component(const std::string &name)
+    {
+        if (unzCloseCurrentFile(file_) != UNZ_OK)
```
abhinav92003, line 110: Assuming it doesn't crash or something if file_ is nullptr?

clients/drcachesim/common/zipfile_istream.h
```diff
@@ -103,25 +104,42 @@ class zipfile_istreambuf_t : public std::basic_streambuf<char, std::char_traits<
...
+            return "Failed to close current component";
```
abhinav92003, line 111: Wasn't there some error struct that gave details for why an unz operation failed? Should we print that?

clients/drcachesim/scheduler/scheduler.h
```diff
@@ -408,6 +423,20 @@ template <typename RecordType, typename ReaderType> class scheduler_tmpl_t {
          * more frequent diagnostics.
          */
         int verbosity = 0;
+        /**
+         * Output stream for recording the schedule for later replay.
+         * write_recorded_schedule() must be called when finished to write the
+         * in-memory data out to this stream.
+         */
+        archive_ostream_t *record_ostream = nullptr;
```
abhinav92003, line 431: schedule_record_ostream would be better; same below.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -507,7 +513,24 @@ scheduler_tmpl_t<RecordType, ReaderType>::init(
                                   // TODO i#5843: Add more flags for other options.
                                   : spec_type_t::LAST_FROM_TRACE,
                               verbosity_);
+        if (options_.record_ostream != nullptr) {
+            sched_type_t::stream_status_t status = record_schedule_segment(
+                i, schedule_record_t::VERSION, schedule_record_t::VERSION_CURRENT, i, 0);
```
abhinav92003, line 518: Here we're seeing `input` to `schedule_record_t::VERSION_CURRENT`?

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -561,6 +584,91 @@ scheduler_tmpl_t<RecordType, ReaderType>::init(
...
+        if (status != sched_type_t::STATUS_OK)
+            return STATUS_ERROR_FILE_WRITE_FAILED;
+        std::string name = recorded_schedule_component_name(i);
+        std::string err = options_.record_ostream->open_new_component(name);
+        if (!err.empty()) {
+            VPRINT(this, 1, "Failed to open component %s in record file: %s\n",
+                   name.c_str(), err.c_str());
+            return STATUS_ERROR_FILE_WRITE_FAILED;
+        }
+        if (!options_.record_ostream->write(
```
abhinav92003, line 617: No need to close?

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -561,6 +584,91 @@ scheduler_tmpl_t<RecordType, ReaderType>::init(
...
+}
+
+template <typename RecordType, typename ReaderType>
+typename scheduler_tmpl_t<RecordType, ReaderType>::scheduler_status_t
+scheduler_tmpl_t<RecordType, ReaderType>::write_recorded_schedule()
+{
+    if (options_.record_ostream == nullptr)
+        return STATUS_ERROR_INVALID_PARAMETER;
+    std::lock_guard<std::mutex> guard(sched_lock_);
+    for (int i = 0; i < static_cast<int>(outputs_.size()); i++) {
```
abhinav92003, line 605: nit: use ++i consistently throughout

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -561,6 +584,91 @@ scheduler_tmpl_t<RecordType, ReaderType>::init(
...
+    // non-archive format and store the output ordinal in the version record.)
+    for (int i = 0; i < static_cast<int>(outputs_.size()); i++) {
+        std::string err =
+            options_.replay_istream->open_component(recorded_schedule_component_name(i));
+        if (!err.empty()) {
+            error_string_ = "Failed to open replay_istream component " +
+                recorded_schedule_component_name(i) + ": " + err;
+            return STATUS_ERROR_INVALID_PARAMETER;
+        }
+        // XXX: This could be made more efficient if we stored the record count
```
abhinav92003, line 646: Well, std::vector does some smart sizing internally right?

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -759,14 +878,36 @@ scheduler_tmpl_t<RecordType, ReaderType>::advance_region_of_interest(
 
     VPRINT(this, 2, "skipping from %" PRId64 " to %" PRId64 " instrs for ROI\n",
            cur_instr, cur_range.start_instruction);
+    if (options_.record_ostream != nullptr) {
+        sched_type_t::stream_status_t status = close_schedule_segment(output, input);
+        if (status != sched_type_t::STATUS_OK)
+            return status;
+        status = record_schedule_segment(output, schedule_record_t::SKIP, input.index,
+                                         cur_instr, cur_range.start_instruction);
```
abhinav92003, line 886: So we're not recording that [cur_instr, cur_range.start_instruction) was skipped?

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -794,45 +935,224 @@ scheduler_tmpl_t<RecordType, ReaderType>::advance_region_of_interest(
...
+    return sched_type_t::STATUS_OK;
+}
+
+template <typename RecordType, typename ReaderType>
+typename scheduler_tmpl_t<RecordType, ReaderType>::stream_status_t
+scheduler_tmpl_t<RecordType, ReaderType>::close_schedule_segment(output_ordinal_t output,
+                                                                 input_info_t &input)
+{
+    assert(output >= 0 && output < static_cast<output_ordinal_t>(outputs_.size()));
+    if (outputs_[output].record.back().type == schedule_record_t::SKIP) {
```
abhinav92003, line 986: Assert !outputs_[output].record.empty()

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -794,45 +935,224 @@ scheduler_tmpl_t<RecordType, ReaderType>::advance_region_of_interest(
...
+                        // We'll come back here on the next next_record() call.
+                        set_cur_input(output, INVALID_INPUT_ORDINAL);
+                        outputs_[output].waiting = true;
+                        return sched_type_t::STATUS_WAIT;
+                    }
+                }
+                if (segment.type == schedule_record_t::SKIP) {
+                    std::lock_guard<std::mutex> lock(*inputs_[index].lock);
+                    if (segment.stop_instruction == 0) {
+                        // A stop of 0 means an artificial end to this input as we're past
```
abhinav92003, line 1125: Is it worth creating a new schedule_record_t value to denote need to add thread exit?

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -846,7 +1166,7 @@ scheduler_tmpl_t<RecordType, ReaderType>::pick_next_input(output_ordinal_t outpu
                 }
             } else if (options_.deps == DEPENDENCY_TIMESTAMPS) {
                 // TODO i#5843: This should require a lock for >1 outputs too.
-                uint64_t min_time = 0xffffffffffffffff;
+                uint64_t min_time = 0xffffffffffffffffU;
```
abhinav92003, line 1169: std::numeric_limits<uint64_t>::max()?

clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -989,6 +994,542 @@ test_speculation()
...
+                                                   /*verbosity=*/4);
+        zipfile_istream_t infile(record_fname);
+        sched_ops.replay_istream = &infile;
+
+        scheduler_t scheduler;
+        if (scheduler.init(sched_inputs, NUM_OUTPUTS, sched_ops) !=
+            scheduler_t::STATUS_SUCCESS)
+            assert(false);
+        // Walk the outputs in lockstep for crude but deterministic concurrency.
+        std::vector<scheduler_t::stream_t *> outputs(NUM_OUTPUTS, nullptr);
```
abhinav92003, line 1103: Can this part be moved to a common function?

clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -989,6 +994,542 @@ test_speculation()
...
+        for (int i = 0; i < NUM_OUTPUTS; i++) {
+            std::cerr << "cpu #" << i << " schedule: " << sched_as_string[i] << "\n";
+        }
+        assert(sched_as_string[0] == CORE0_SCHED_STRING);
+        assert(sched_as_string[1] == CORE1_SCHED_STRING);
+        if (scheduler.write_recorded_schedule() != scheduler_t::STATUS_SUCCESS)
+            assert(false);
+    }
+    // Now replay the schedule.
+    {
```
abhinav92003, line 1081: Perhaps replay should be repeated a bunch of times to make sure.

abhinav92003 (approved):

abhinav92003 (approved):

clients/drcachesim/CMakeLists.txt
```diff
@@ -299,6 +299,8 @@ install_client_nonDR_header(drmemtrace common/utils.h)
 install_client_nonDR_header(drmemtrace common/trace_entry.h)
```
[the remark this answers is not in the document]
derekbruening: Done and done by editing the description (which with the new options will become the commit message).

clients/drcachesim/common/zipfile_istream.h
derekbruening: Added a check for nullptr.

clients/drcachesim/common/zipfile_istream.h
```diff
@@ -103,25 +104,42 @@ class zipfile_istreambuf_t : public std::basic_streambuf<char, std::char_traits<
...
+        if (unzLocateFile(file_, name.c_str(), /*iCaseSensitivity=*/true) != UNZ_OK)
+            return "Failed to open zipfile component " + name;
```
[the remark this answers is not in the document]
derekbruening: Done.

clients/drcachesim/common/zipfile_istream.h
derekbruening: The return value has some different error codes defined.  I added that to the returned string here and in the two cases below.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Since input_ordinal_t and version are both ints we can't overload unfortunately.  Added a comment to record_schedule_segment's declaration.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: std::ostream does not have a close method.  raw2trace also takes in ostreams and has no way to close.  It will be closed generally by the caller who owns the lifetime when it goes out of scope there.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: It looks like this file is consistently using i++ so it is already consistent?

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: It presumably doubles each time so it won't be too bad, right, and this is just a one-time thing.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -561,6 +584,91 @@ scheduler_tmpl_t<RecordType, ReaderType>::init(
...
+                                             sizeof(record))) {
+            if (record.type == schedule_record_t::VERSION) {
+                if (record.key.version != schedule_record_t::VERSION_CURRENT)
+                    return STATUS_ERROR_INVALID_PARAMETER;
+            } else if (record.type == schedule_record_t::FOOTER)
+                break;
+            else
+                outputs_[i].record.push_back(record);
+        }
+        VPRINT(this, 1, "Read %zu recorded records for output #%d\n",
```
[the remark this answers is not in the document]
derekbruening: Done.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -743,6 +851,17 @@ scheduler_tmpl_t<RecordType, ReaderType>::advance_region_of_interest(
                 return sched_type_t::STATUS_EOF;
             else {
                 // We let the user know we're done.
+                if (options_.record_ostream != nullptr) {
+                    sched_type_t::stream_status_t status =
+                        close_schedule_segment(output, input);
+                    if (status != sched_type_t::STATUS_OK)
+                        return status;
+                    // Indicate we need a synthetic thread exit on replay.
```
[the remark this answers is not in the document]
derekbruening: Done.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: I don't understand: we are recording exactly that, that that range was skipped.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Done.

clients/drcachesim/scheduler/scheduler.cpp
abhinav92003: Some other files in the PR were using ++i, the tests I think. ++i is generally more efficient, so we can update all instances in this file, optionally.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Done.

clients/drcachesim/scheduler/scheduler.cpp
abhinav92003: I think I misread. Ignore.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Done.

clients/drcachesim/scheduler/scheduler.h
derekbruening: Hmm, I feel that "schedule" is kind of implied as this is a scheduler, but willing to rename it.  Done.

derekbruening: This makes the Doxygen references too long, violating the vera++ check.  Searching again for possible ways to avoid fully qualifying for Doxygen links.

derekbruening: Splitting across the comment line (and still indenting + *) seems to work, in a weird way though generating two links: but good enough.

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: Done.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: For an int it doesn't matter and i++ is the traditional C style -- but I do see the Google C++ style guide votes for ++i.  Sure, changed.

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: Done.

Issue #5634, opened 2022-08-31
drmemtrace 32-bit truncates timestamps and other 64-bit markers: should add 64-bit value support


derekbruening:
In 32-bit builds, drmemtrace does not support a 2-entry marker and limits the marker value to 32 bits.  This means that timestamps are truncated, and other markers cannot support 64-bit values.  We should add the 2-entry solution we use on 64-bit.

derekbruening:
This means we need an analysis-tool-visible extended-payload marker type, or maybe make the reader hide that from the user.

derekbruening:
Another option is to deprecate the 32-bit record format and only support
64-bit fields.  This would allow cross-bitwidth analysis: today a 64-bit
drmemtrace build can't read a 32-bit trace and vice versa.

Issue #14, opened 2014-11-27
gcc 4.3.0 bug: ASSERT on all x64 apps: "reachability contraints not satisfiable"


derekbruening:
_From [derek.br...@gmail.com](https://code.google.com/u/117968039472581148324/) on February 15, 2009 17:11:00_

ASSERT with gcc 4.3.0 on all x64 apps: "reachability contraints not
satisfiable"

SYSLOG_ERROR: Application hello64 (1982).  Internal Error Internal
DynamoRIO Error: heap.c:474 heap_allowable_region_start <=
must_reach_region_start && "PR 215395 reachability contraints not satisfiable"

This is due to gcc 4.3.0, even at -O0, considering "ptr - const < ptr" (in
REACHABLE_32BIT_END()) to always be true, which is an invalid optimization
as it ignores underflow.  I have a simple test app that shows the problem
very clearly, inlined below with objdump -dS.  I don't see any gcc options
to affect this optimization.

The disturbing aspect is I don't know what other bugs this gcc behavior is
going to cause: certainly throughout win32/os.c we watch for overflow via
"if (pb + mbi.RegionSize < pb)".

The toolchain compiler gcc 4.1.2 does not have the buggy behavior of course.

/work/dr/test/gcc-underflow-bug.c:
  #include <stdio.h>

  int main()
  {
      unsigned char _ptr1 = (unsigned char *) 0x0000000071542000;
      unsigned char *ptr3 = ptr1 - 0x80000000; /_ 0xfffffffff1542000 */
      if (ptr3 < ptr1)
          printf("ptr3 less than\n");
      if ((ptr1 - 0x80000000) < ptr1)
          printf("temp less than\n");
  }

  #if 0
  gcc 4.3.0 -O0:
  int main()
  {
    4004cc:       55                      push   &#37;rbp
    4004cd:       48 89 e5                mov    &#37;rsp,&#37;rbp
    4004d0:       48 83 ec 20             sub    $0x20,&#37;rsp
      unsigned char _ptr1 = (unsigned char *) 0x0000000071542000;
    4004d4:       48 c7 45 f0 00 20 54    movq   $0x71542000,-0x10(&#37;rbp)
    4004db:       71
      unsigned char *ptr3 = ptr1 - 0x80000000; /_ 0xfffffffff1542000 */
    4004dc:       48 8b 45 f0             mov    -0x10(&#37;rbp),&#37;rax
[...]

PR #6037, opened 2023-05-10
Add timestamp marker increase monotonically invariant check.

ivankyluk:
Please review the change to add an invariant check to verify timestamp markers are increasing monotonically.

Reference:

From https://github.com/DynamoRIO/dynamorio/issues/2039#issuecomment-1294014977 and https://github.com/DynamoRIO/dynamorio/pull/5704#discussion_r1007187555


clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -691,6 +691,33 @@ check_rseq_side_exit_discontinuity()
...
 
+bool
+check_timestamp_increase_monotonically(void)
+{
+    // Positive test: timestamps increase monotonically.
+    {
+        std::vector<memref_t> memrefs = {
+            gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
+            gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 10),
+            gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 10),
```
derekbruening, line 702: Hmm, I believe "monotonically" implies values do not stay the same.  Are you worried about two identical real timestamps in a row?  I don't think this will happen with our microsecond timestamps on the current platforms; it seems it would only happen on some platform without a reliable OS-provided time interface but there we would probably use the processor timestamp counter.

ivankyluk: Based on 

https://en.wikipedia.org/wiki/Monotonic_function: 
A function is called monotonically increasing (also increasing or non-decreasing)[3] if for all x and y such that x <=y one has f(x) <= f(y), so f preserves the order.

https://en.wiktionary.org/wiki/monotonic_increasing#:~:text=monotonic%20increasing%20(not%20comparable),contrast%20this%20with%20strictly%20increasing:
(mathematics, of a function) always increasing or remaining constant, and never decreasing; contrast this with strictly increasing

And Wolfram MathWorld has the definition you mention:
https://mathworld.wolfram.com/MonotoneIncreasing.html
Always increasing; never remaining constant or decreasing. Also called strictly increasing.

Anyway, let me update the code to ensure the timestamps are increasing, and do not stay the same.

ivankyluk: I've updated the code with a second commit, https://github.com/DynamoRIO/dynamorio/pull/6037/commits/1f90288d7775034fad0e96add549c2f4685f965b.

I forgot to put a comment here. :o(

derekbruening: > I forgot to put a comment here. :o(

I think you mean, you forgot to re-request review: there is a circular arrow in the upper right of this page for that.  I just clicked on it.

derekbruening (approved):

derekbruening:
Thank you for the PR.  I will merge it.

derekbruening:
Well we hit a violation causing test failure in x86-64 tool.drcacheoff.phys-SUDO:
https://github.com/DynamoRIO/dynamorio/actions/runs/5007379198/jobs/8973920770?pr=6037
```
344:              6           0:       25181 <marker: timestamp 13328827882613812>
344:              7           0:       25181 <marker: tid 25181 on core 1>
344:              8           0:       25181 <marker: physical address for following virtual: 0x24a098000>
344:              9           0:       25181 <marker: virtual address for prior physical: 0x401000>
344:             10           0:       25181 <marker: physical address for following virtual: 0x272d1e00e>
344:             11           0:       25181 <marker: virtual address for prior physical: 0x40200e>
344:             12           0:       25181 <marker: physical address for following virtual: 0x27767c000>
344:             13           0:       25181 <marker: virtual address for prior physical: 0x403000>
344: 
344:   Trace invariant failure in T25181 at ref # 14: Timestamp does not increase
```

Unfortunately it's not printing ref #14 for us there (not sure who does that printing) -- but looking at process_entry_for_physaddr() it seems to deliberately clone a timestamp, so we have 2 in a row with the same value.  Looks like disallowing identical timestamps was a mistake :(

derekbruening:
Also seeing Windows test failures: `Trace invariant failure in T6600 at ref # 63: Timestamp does not increase` -- could it really be from consecutive live timestamps?  Maybe in these VM's used for testing it's possible to get the same microsecond value.

ivankyluk:
Let me update the code to allow timestamps to stay the same.

derekbruening:
My original request was misguided it seems: sorry about that.  Maybe can just revert the 2nd commit.

derekbruening (approved):

derekbruening:
Wow the 32-bit Windows filter-simple test failed with:
```
 Trace invariant failure in T1040 at ref # 58: Timestamp is less than the previous one
```
So did tool.drcachesim.invariants:
```
Trace invariant failure in T2556 at ref # 299: Timestamp is less than the previous one
```

So it's not sthg weird with filtering: the invariants test runs the `client.winxfer` app under regular online tracing.  Hmm.  We get the timestamp from the OS.  Not sure how it can decrease...is it some artifact of -align_endpoints?

derekbruening:
> Wow the 32-bit Windows filter-simple test failed with:

@abhinav92003 reminded me that 32-bit truncates the timestamp counters (!!): which is #5634.  So for the short term until that is fixed (32-bit is not a high priority though...) I think we have to restrict this check to 64-bit.

ivankyluk:
I would like to clarify if "drmemtrace 32-bit truncates timestamps" means it's taking the least significant 32 bits?
Or the most significant 32 bits.

If it's taking the least significant 32 bits, the timestamp check can be modified to add a rollover check. For example,
the previous timestamp is 2^31 (or another number) more than the current one, assuming there should be no such gaps in between in non rollover case.

derekbruening:
Least significant:
```
$ git grep -A 1 i#5634
clients/drcachesim/tracer/raw2trace.cpp:                                // XXX i#5634: Truncated for 32-bit, as documented.
clients/drcachesim/tracer/raw2trace.cpp-                                (uintptr_t)timestamp);
```

ivankyluk:
I filed #6187 to track it using a different pull request.

PR #6038, opened 2023-05-10
i#5843 scheduler: Update checked-in trace

derekbruening:
Updates the checked-in thread sig trace, which is old and small, to a new trace in the zip format and with more threads and more code.  This is a better stress test of the scheduler.  This trace is already used by the scheduler tests.

Augments the scheduler_launcher to add a sanity test running it on the new trace added here.

Issue: #5843


derekbruening:
Looks like record_filter_unit_tests can't handle the new trace: I think b/c the record filter doesn't support .zip format.

abhinav92003 (approved):

derekbruening:
x86-32 error is api.startstop vdso sysenter: #4604

PR #6039, opened 2023-05-10
i#5843: Add scheduler CLI tool

derekbruening:
Adds a new larger manual test harness in the form of a standalone launcher for the scheduler which can be used on larger application threads for experimentation.

Issue: #5843

### DIFF
--- a/clients/drcachesim/CMakeLists.txt
+++ b/clients/drcachesim/CMakeLists.txt
@@ -342,6 +342,11 @@ add_executable(record_filter_launcher
 target_link_libraries(record_filter_launcher drmemtrace_analyzer drmemtrace_record_filter)
 use_DynamoRIO_extension(record_filter_launcher droption)
 
+add_executable(scheduler_launcher tests/scheduler_launcher.cpp)
+target_link_libraries(scheduler_launcher drmemtrace_analyzer drfrontendlib)
+use_DynamoRIO_extension(scheduler_launcher droption)
+add_dependencies(scheduler_launcher api_headers)
+
 # We have a companion test built using a separate --build-and-test CMake project in
 # tests/analyzer_separate.cpp to better test 3rd-party usage.
 set_property(GLOBAL PROPERTY DynamoRIO_drmemtrace_src_dir
@@ -373,6 +378,7 @@ target_link_libraries(drmemtrace_analyzer ${zlib_libs})
 if (NOT AARCH64 AND NOT APPLE)
   target_link_libraries(opcode_mix_launcher ${zlib_libs})
 endif ()
+target_link_libraries(scheduler_launcher ${zlib_libs})
 
 macro(add_drmemtrace name type)
   if (${type} STREQUAL "STATIC")
@@ -480,6 +486,7 @@ restore_nonclient_flags(prefetch_analyzer_launcher)
 if (NOT AARCH64 AND NOT APPLE)
   restore_nonclient_flags(opcode_mix_launcher)
 endif ()
+restore_nonclient_flags(scheduler_launcher)
 restore_nonclient_flags(drmemtrace_simulator)
 restore_nonclient_flags(drmemtrace_reuse_distance)
 restore_nonclient_flags(drmemtrace_histogram)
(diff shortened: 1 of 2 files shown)


clients/drcachesim/tests/scheduler_launcher.cpp
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    scheduler_t scheduler;
+    std::vector<scheduler_t::input_workload_t> sched_inputs;
+    sched_inputs.emplace_back(op_trace_dir.get_value());
+    scheduler_t::scheduler_options_t sched_ops(
+        scheduler_t::MAP_TO_ANY_OUTPUT, scheduler_t::DEPENDENCY_IGNORE,
+        scheduler_t::SCHEDULER_DEFAULTS, op_verbose.get_value());
+    sched_ops.quantum_duration = op_sched_quantum.get_value();
+#ifdef HAS_ZLIB
```
abhinav92003, line 154: s/HAS_ZLIB/HAS_ZIP/?

I think the latter implies the former but not vice versa.

clients/drcachesim/tests/scheduler_launcher.cpp
```diff
@@ -0,0 +1,201 @@
...
+    std::unique_ptr<zipfile_ostream_t> record_zip;
+    std::unique_ptr<zipfile_istream_t> replay_zip;
+    if (!op_record_file.get_value().empty()) {
+        record_zip.reset(new zipfile_ostream_t(op_record_file.get_value()));
+        sched_ops.schedule_record_ostream = record_zip.get();
+    }
+    if (!op_replay_file.get_value().empty()) {
```
abhinav92003, line 161: else if? Because I suppose both cannot be set at the same time?

clients/drcachesim/tests/scheduler_launcher.cpp
```diff
@@ -0,0 +1,201 @@
...
+        threads.emplace_back(std::thread(&simulate_core, i, scheduler.get_stream(i),
+                                         std::ref(scheduler), std::ref(schedules[i])));
+    }
+    for (std::thread &thread : threads)
+        thread.join();
+
+    for (int i = 0; i < op_num_cores.get_value(); ++i) {
+        std::cerr << "Core #" << i << ": ";
+        for (memref_tid_t tid : schedules[i])
+            std::cerr << tid << " ";
```
abhinav92003, line 187: Count of instrs executed by the tid on this core isn't useful?

clients/drcachesim/CMakeLists.txt
```diff
@@ -342,6 +342,11 @@ add_executable(record_filter_launcher
 target_link_libraries(record_filter_launcher drmemtrace_analyzer drmemtrace_record_filter)
 use_DynamoRIO_extension(record_filter_launcher droption)
 
+add_executable(scheduler_launcher tests/scheduler_launcher.cpp)
```
abhinav92003, line 345: Simple run as a test would be useful.

abhinav92003 (approved):

derekbruening:
Whoops: will apply the suggested changes to the related PR #6038

clients/drcachesim/tests/scheduler_launcher.cpp
derekbruening: Done.

clients/drcachesim/tests/scheduler_launcher.cpp
derekbruening: Done.

clients/drcachesim/tests/scheduler_launcher.cpp
derekbruening: I made this off by default when I was trying to make a test out of this and was having trouble with non-atomic long writes and with ordinals not matching; but since the scheduler_unit_tests now contains that test, yes we can make this print by default.  Done.

clients/drcachesim/CMakeLists.txt
derekbruening: Done.

Issue #5694, opened 2022-10-20
Add core-oriented drmemtrace iterator


derekbruening:
When drmemtrace tools want to analyze a trace using the recorded schedule they may want to look at hardware threads rather than software threads.  An iterator over each core would be useful for this purpose.  For interactions with fast seeking #5538 the plan is to generate a schedule summary file that contains `<tid, timestamp, cpuid, instr-count>` tuples to make it possible to compute instruction counts on a core across multiple software threads.

PR #6045, opened 2023-05-12
i#5843 scheduler: Add replay of the as-traced schedule

derekbruening:
Adds replay of the as-traced schedule using the cpu_schedule.zip file written by raw2trace.  That file is converted into the record+replay format for dynamic schedules and the existing replay code leveraged.

This implements as-traced cpu-oriented parallel iteration in the scheduler which is part of #5694.  Adding analysis tool support for this will be done separately.

Updates all replays (whether as-traced or as-previously) to consider DEPENDENCY_TIMESTAMPS to indicate whether to have one output wait if it gets ahead of another's timestamp.

Removes STATUS_IDLE since we always use STATUS_WAIT.  We're between releases so we can delete from the public enum for this still-under-development component without worrying about binary compatibility.

Adds a missing check for output count mismatches on as-previously replay.

Adds unit tests of synthetic and from-file replay.

Issue: #5843, #5694

### DIFF
--- a/clients/drcachesim/scheduler/scheduler.cpp
+++ b/clients/drcachesim/scheduler/scheduler.cpp
@@ -778,12 +778,15 @@ scheduler_tmpl_t<RecordType, ReaderType>::read_traced_schedule()
                     +toadd.key.input, +toadd.start_instruction,
                     +all_sched[i][j].stop_instruction, +toadd.timestamp);
                 start_consec = -1;
-                auto &added = outputs_[i].record.back();
-                VPRINT(this, 3,
-                       "segment #%zu: input=%d start=%" PRId64 " stop=%" PRId64
-                       " time=%" PRId64 "\n",
-                       outputs_[i].record.size() - 1, added.key.input,
-                       added.start_instruction, added.stop_instruction, added.timestamp);
+                VDO(this, 3, {
+                    auto &added = outputs_[i].record.back();
+                    VPRINT(this, 3,
+                           "segment #%zu: input=%d start=%" PRId64 " stop=%" PRId64
+                           " time=%" PRId64 "\n",
+                           outputs_[i].record.size() - 1, added.key.input,
+                           added.start_instruction, added.stop_instruction,
+                           added.timestamp);
+                });
             }
         }
         VPRINT(this, 1, "Collapsed duplicates for %zu as-traced records for output #%d\n",
--- a/clients/drcachesim/scheduler/scheduler.cpp
+++ b/clients/drcachesim/scheduler/scheduler.cpp
@@ -30,6 +30,9 @@
  * DAMAGE.
  */
 
+#include <map>
+#include <unordered_map>
+
 #include "scheduler.h"
 #include "file_reader.h"
 #ifdef HAS_ZLIB
@@ -526,8 +529,7 @@ scheduler_tmpl_t<RecordType, ReaderType>::init(
     }
     if (options_.mapping == MAP_AS_PREVIOUSLY) {
         if (options_.schedule_replay_istream == nullptr ||
-            options_.schedule_record_ostream != nullptr ||
-            options_.deps != DEPENDENCY_IGNORE)
+            options_.schedule_record_ostream != nullptr)
             return STATUS_ERROR_INVALID_PARAMETER;
         sched_type_t::scheduler_status_t status = read_recorded_schedule();
         if (status != sched_type_t::STATUS_SUCCESS)
@@ -545,14 +547,15 @@ scheduler_tmpl_t<RecordType, ReaderType>::init(
             VPRINT(this, 2, "Assigning input #%d to output #%zd\n", i, index);
         }
     } else if (options_.mapping == MAP_TO_RECORDED_OUTPUT) {
-        // TODO i#5843,i#5694: Implement cpu-oriented parallel iteration.
-        // Initially we only support analyzer_t's serial mode.
-        if (options_.deps != DEPENDENCY_TIMESTAMPS)
-            return STATUS_ERROR_NOT_IMPLEMENTED;
-        // TODO i#5843: Support more than one output.
-        if (output_count != 1)
-            return STATUS_ERROR_NOT_IMPLEMENTED;
-        if (inputs_.size() == 1) {
+        if (output_count > 1) {
+            if (options_.replay_as_traced_istream == nullptr)
+                return STATUS_ERROR_INVALID_PARAMETER;
+            sched_type_t::scheduler_status_t status = read_traced_schedule();
+            if (status != sched_type_t::STATUS_SUCCESS)
+                return STATUS_ERROR_INVALID_PARAMETER;
+            // Now leverage the regular replay code.
+            options_.mapping = MAP_AS_PREVIOUSLY;
+        } else if (inputs_.size() == 1) {
             set_cur_input(0, 0);
         } else {
             // The old file_reader_t interleaving would output the top headers for every
@@ -564,9 +567,6 @@ scheduler_tmpl_t<RecordType, ReaderType>::init(
                 return res;
         }
     } else {
-        // TODO i#5843: Honor deps when scheduling on synthetic cores.
-        if (options_.deps != DEPENDENCY_IGNORE)
-            return STATUS_ERROR_NOT_IMPLEMENTED;
         // TODO i#5843: Implement time-based quanta.
         if (options_.quantum_unit != QUANTUM_INSTRUCTIONS)
             return STATUS_ERROR_NOT_IMPLEMENTED;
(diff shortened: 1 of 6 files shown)


clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -669,16 +669,145 @@ scheduler_tmpl_t<RecordType, ReaderType>::read_recorded_schedule()
...
+    // designed for skip targets only), so we take two passes to get that information.
+    // We also need to translate the thread and cpu id values into 0-based ordinals.
+    std::unordered_map<memref_tid_t, input_ordinal_t> tid2input;
+    for (int i = 0; i < static_cast<input_ordinal_t>(inputs_.size()); ++i) {
+        tid2input[inputs_[i].tid] = i;
+    }
+    std::unordered_map<uint64_t, output_ordinal_t> cpu2output;
+    std::vector<std::map<uint64_t, uint64_t>> start2stop(inputs_.size());
+    // We number the outputs according to their order in the file.
+    // XXX: Should we support some direction from the user on this?
```
abhinav92003, line 711: You mean the core_id -> output_id mapping? Does that  matter? I assumed no as the ids are an internal detail. If it does, does it also matter which thread_id is mapped to which input_id?

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -669,16 +669,145 @@ scheduler_tmpl_t<RecordType, ReaderType>::read_recorded_schedule()
...
+                }
+            }
+            cur_cpu = entry.cpu;
+            cpu2output[entry.cpu] = cur_output;
+        }
+        input_ordinal_t input = tid2input[entry.thread];
+        // We'll fill in the stop ordinal in our second pass below.
+        uint64_t start = entry.instr_count;
+        uint64_t timestamp = entry.timestamp;
+        // Some entries have no instructions.
```
abhinav92003, line 735: Add why.

Side question: does it affect size of the schedule in any significant way?

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -669,16 +669,145 @@ scheduler_tmpl_t<RecordType, ReaderType>::read_recorded_schedule()
...
+        if (all_sched[cur_output].size() > 0 &&
```
abhinav92003, line 736: Use !_.empty()

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -669,16 +669,145 @@ scheduler_tmpl_t<RecordType, ReaderType>::read_recorded_schedule()
...
+            VPRINT(this, 3,
+                   "Output #%d: as-read segment #%zu has no instructions: skipping\n",
+                   cur_output, all_sched[cur_output].size() - 1);
+            continue;
+        }
+        all_sched[cur_output].emplace_back(schedule_record_t::DEFAULT, input, start, 0,
+                                           timestamp);
+        start2stop[input][entry.instr_count] = entry.instr_count;
+    }
+    for (int i = 0; i < static_cast<output_ordinal_t>(outputs_.size()); ++i) {
```
abhinav92003, line 748: Would it help memory usage efficiency if we collapsed each output as soon as it is created, instead of reading all outputs first? Or does that make code more complex?

abhinav92003, line 748: I find it useful to use more verbose iterator vars when there are multiple iterators or access to a > 1-D data structure. Optionally use output_idx here.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -669,16 +669,145 @@ scheduler_tmpl_t<RecordType, ReaderType>::read_recorded_schedule()
...
+            input == all_sched[cur_output].back().key.input &&
+            start == all_sched[cur_output].back().start_instruction) {
+            VPRINT(this, 3,
+                   "Output #%d: as-read segment #%zu has no instructions: skipping\n",
+                   cur_output, all_sched[cur_output].size() - 1);
+            continue;
+        }
+        all_sched[cur_output].emplace_back(schedule_record_t::DEFAULT, input, start, 0,
+                                           timestamp);
+        start2stop[input][entry.instr_count] = entry.instr_count;
```
abhinav92003, line 746: Perhaps a `set` is what we need?

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -669,16 +669,145 @@ scheduler_tmpl_t<RecordType, ReaderType>::read_recorded_schedule()
...
+        }
+        input_ordinal_t input = tid2input[entry.thread];
+        // We'll fill in the stop ordinal in our second pass below.
+        uint64_t start = entry.instr_count;
+        uint64_t timestamp = entry.timestamp;
+        // Some entries have no instructions.
+        if (all_sched[cur_output].size() > 0 &&
+            input == all_sched[cur_output].back().key.input &&
+            start == all_sched[cur_output].back().start_instruction) {
+            VPRINT(this, 3,
```
abhinav92003, line 739: Would the timestamp field be different for such cases?

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -669,16 +669,145 @@ scheduler_tmpl_t<RecordType, ReaderType>::read_recorded_schedule()
...
+                    start_consec >= 0 ? all_sched[i][start_consec] : all_sched[i][j];
+                outputs_[i].record.emplace_back(
+                    static_cast<typename schedule_record_t::record_type_t>(toadd.type),
+                    +toadd.key.input, +toadd.start_instruction,
+                    +all_sched[i][j].stop_instruction, +toadd.timestamp);
+                start_consec = -1;
+                VDO(this, 3, {
+                    auto &added = outputs_[i].record.back();
+                    VPRINT(this, 3,
+                           "segment #%zu: input=%d start=%" PRId64 " stop=%" PRId64
```
abhinav92003, line 784: Maybe add output-idx to log too?

Also, if one want sto see the final entries but not the as-read entries, use a higher verbosity level above or a lower here.

abhinav92003 (approved):

clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -1468,6 +1467,145 @@ test_replay_skip()
...
+        inputs[i].push_back(make_exit(tid));
+    }
+
+    // Synthesize a cpu-schedule file.
+    std::string cpu_fname = "tmp_test_cpu_as_traced.zip";
+    static const char *const CORE0_SCHED_STRING = "AAACCCAAACCCBBBDDD";
+    static const char *const CORE1_SCHED_STRING = "BBBDDDBBBDDDAAACCC";
+    {
+        std::vector<schedule_entry_t> sched0;
+        sched0.emplace_back(TID_BASE, 101, CPU0, 0);
```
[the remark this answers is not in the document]
derekbruening: The `0` is the start instr ordinal.  You have to look at the subsequent record's start to find the count.  The 0 equals 1 (our ordinals are 1-based) so the 0..4 is 3 instrs; 4..7 and 7..infinity (9 total) are also 3.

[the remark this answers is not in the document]
abhinav92003: Ah that's not the count of instructions executed. The "instr_count" identifier is confusing. It should be "instr_ordinal" or something to denote the point in the thread's instr stream..

clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -1468,6 +1467,145 @@ test_replay_skip()
...
+#if (defined(X86_64) || defined(ARM_64)) && defined(HAS_ZIP)
+    std::cerr << "\n----------------\nTesting replay as-traced from a file\n";
+    std::string path = std::string(testdir) + "/drmemtrace.threadsig.x64.tracedir";
+    std::string cpu_file =
+        std::string(testdir) + "/drmemtrace.threadsig.x64.tracedir/cpu_schedule.bin.zip";
+    // This checked-in trace has 8 threads on 9 cores, with one core starting
+    // out in a wait state.
+    static const char *const SCHED_STRING =
+        "Core #0: 3018553 \nCore #1: 3018556 \nCore #2: 3018557 \nCore #3: 3018558 "
+        "\nCore #4: 3018520 \nCore #5: 3018520 \nCore #6: 3018554 \nCore #7: 3018555 "
```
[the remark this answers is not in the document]
derekbruening: I was also surprised it didn't have more.  There is a tradeoff here, though, where we want our GA CI to not take forever to run and we don't want to check in large binary files into our repository so we mostly run small apps.  This is the "threadsig" app we use for drmemtrace_samples.  This already stressed several corner cases (core 5 having to wait for core 4 right up front; merging duplicate entries) so it is a pretty good test; plus, the unit test has migration to test that aspect.  I could try to trace the threadsig app a few more times and hope to get more migration but it may not happen on this machine (plus it is work to then update the other tests that use this checked-in trace).  We also have a bigger internal proprietary test which does have a ton of migration.  I'm leaning toward just adding a comment here acknowledge the few migrations but pointed at the synthetic test having more migration.

clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -1468,6 +1467,145 @@ test_replay_skip()
...
+        inputs[i].push_back(make_thread(tid));
+        inputs[i].push_back(make_pid(1));
+        for (int j = 0; j < NUM_INSTRS; j++)
+            inputs[i].push_back(make_instr(42 + j * 4));
+        inputs[i].push_back(make_exit(tid));
+    }
+
+    // Synthesize a cpu-schedule file.
+    std::string cpu_fname = "tmp_test_cpu_as_traced.zip";
+    static const char *const CORE0_SCHED_STRING = "AAACCCAAACCCBBBDDD";
```
abhinav92003, line 1494: Do we verify in any way if the BBB in core0 are the instrs 7-9 in B's instrs stream? Could they be instrs 4-6?

abhinav92003: Oh it's executes lockstep, so the fact that the index of Bs in core0 is > index of Bs in core1, that ensures it.

clients/drcachesim/tests/scheduler_unit_tests.cpp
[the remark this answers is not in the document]
abhinav92003: Ack. That's fine.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Simulation may want to preserve the NUMA relationships.  E.g., for a 2-socket chip, keep the simulated cores separated into their respective sockets to match how it was traced.  So the simulator will want to know which output streams are from which original cpuid.  We do have a proposal to add a cpuid query to the memtrace_stream_t interface for convenience but some of the output stream to core mapping is probably best done up front in the simulator before reading any actual stream records.

clients/drcachesim/tests/scheduler_unit_tests.cpp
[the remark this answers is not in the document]
derekbruening: Yes it confused me as well reading it last week even though I wrote it (months ago).  I just added a comment at least to that field.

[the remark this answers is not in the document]
derekbruening: OK I also renamed the field to `start_instruction`: just a couple of uses; not likely external users are looking at this.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: I don't think the input mappings matter; just the output.

derekbruening: I upgraded the XXX to a TODO: i#5843 and elaborated in the comment.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -669,16 +669,145 @@ scheduler_tmpl_t<RecordType, ReaderType>::read_recorded_schedule()
...
+    // We do not have a subfile listing feature in archive_istream_t, but we can
+    // read sequentially as each record has a cpu field.
+    // This schedule_entry_t format doesn't have the stop instruction ordinal (as it was
+    // designed for skip targets only), so we take two passes to get that information.
+    // We also need to translate the thread and cpu id values into 0-based ordinals.
+    std::unordered_map<memref_tid_t, input_ordinal_t> tid2input;
+    for (int i = 0; i < static_cast<input_ordinal_t>(inputs_.size()); ++i) {
+        tid2input[inputs_[i].tid] = i;
+    }
+    std::unordered_map<uint64_t, output_ordinal_t> cpu2output;
```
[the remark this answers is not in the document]
derekbruening: Not needed for this impl; removed.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Done.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: The cpu_schedule file records every tracer timestamp location even if consecutive ones are on the same cpuid (it is trying to also be a record of all timestamps).  The examples I saw were things like a signal arriving right up front at thread attach before any instrs were executed so we had 2 timestamps w/ no instrs in between.  Added this to the comment.

Re: size: it affects the size of the cpu_schedule; it does not affect the scheduler internal format.  It is rare.  The bigger source if size inflation in the cpu_schedule is just repeating consecutive timestamps w/ the same cpuid.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: It is actually impossible because each entry only stores its start and not its stop so we don't know if there was a migration in between two consecutive starts on one cpu.  The original use case for this cpu_schedule file was to make per-core skipping faster.  Maybe we could make a v2 that adds the stop; for now I would say let's live with these 2 passes and having to pull the whole every-timestamp entry into memory and only if we hit issues should we spend more time on it.  I augmented the comment toward the top of the file regarding this.

derekbruening: Done and s/j/shed_idx/

abhinav92003: I don't mean that we should remove the two passes. Currently, we first read all entries for *all* outputs. I was suggesting reading for one output and then collapsing it, then reading the entries for the next output.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Yes, we just need a sorted container where we can ++ to find the next element.  Changed.

clients/drcachesim/scheduler/scheduler.cpp
abhinav92003: I guess it won't make that big of a difference since the no-instruction entries are rare.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Each output is identified by the "Read %zu..." above.  Raised verbosity above.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: The no-instruction entries are already collapsed as they're read so they do not occupy memory.  The same-cpuid-consecutive entries are what are impossible to collapse until all other outputs are read.

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: Right.

PR #6049, opened 2023-05-16
Fixed bug in LRU and FIFO cache replacement policies

jperrett256:
Fixed bug in LRU and FIFO cache replacement policies; the counters were being cleared on invalidations when they weren't supposed to be. Clearing the counters was only the correct thing to do for the LFU implementation.

Also other minor fixes/changes, most notably: made use_physical option available from config file (previously only available via command line).


clients/drcachesim/reader/config_reader.cpp
```diff
@@ -145,6 +145,18 @@ config_reader_t::configure(std::istream *config_file, cache_simulator_knobs_t &k
             } else {
                 knobs.model_coherence = false;
             }
+        } else if (param == "use_physical") {
```
derekbruening, line 148: This looks unrelated: please separate into a new PR.

clients/drcachesim/simulator/cache_fifo.cpp
```diff
@@ -62,6 +63,12 @@ cache_fifo_t::init(int associativity, int block_size, int total_size,
     return true;
 }
 
+void
+cache_fifo_t::invalidate_update(caching_device_block_t *block)
+{
+    // Leave counters unchanged for FIFO.
```
derekbruening, line 69: Please augment the FIFO and LRU unit tests in cache_replacement_policy_unit_test.cpp to catch this bug without the fix and pass only with the fix for a regression test.

derekbruening (commented):
Thank you for submitting this PR.  Hoping you can add test cases to this PR: we have unit tests for FIFO and LRU but they must be incomplete if they did not catch this?

clients/drcachesim/reader/config_reader.cpp
jperrett256: ._.

clients/drcachesim/simulator/cache_fifo.cpp
jperrett256: The test now checks invalidations set the counters appropriately. I went and added tests for LFU as well. I also ran the test without the fixes and it breaks horribly (as expected).

clients/drcachesim/reader/config_reader.cpp
jperrett256: Done.

jperrett256:
Can you ignore the complaints from clang format here? The changes would make those comments much harder to read.

derekbruening:
> Can you ignore the complaints from clang format here? The changes would make those comments much harder to read.

`git grep "clang-format off"` to see how to do that.

@brettcoon will do the final review of the code.

brettcoon:
Regarding the clang warnings, another option is to put the column-alignment spaces inside the comment rather than outside, so clang will leave them alone.

A bigger question about this change is why does _any_ replacement need to modify block->counter_ for an invalidate?  If I just comment out that line in simulator/caching_device.h in an otherwise unmodified drcachesim, your (much appreciated) extended test passes.  It appears to me that no replacement actually needs invalidate_update() functionality.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
```diff
@@ -140,14 +149,28 @@ unit_test_cache_lru_four_way()
...
-    cache_lru_test.access_and_check_cache(addr_vec[ADDR_A], 1); // A b C D
-    cache_lru_test.access_and_check_cache(addr_vec[ADDR_E], 2); // A E c D
+    cache_lru_test.access_and_check(addr_vec[ADDR_A], 1);     // A x X X
+    cache_lru_test.access_and_check(addr_vec[ADDR_B], 2);     // A B x X
+    cache_lru_test.access_and_check(addr_vec[ADDR_C], 3);     // A B C x
+    cache_lru_test.access_and_check(addr_vec[ADDR_D], 0);     // a B C D
+    cache_lru_test.access_and_check(addr_vec[ADDR_A], 1);     // A b C D
+    cache_lru_test.access_and_check(addr_vec[ADDR_A], 1);     // A b C D
+    cache_lru_test.access_and_check(addr_vec[ADDR_A], 1);     // A b C D
+    cache_lru_test.access_and_check(addr_vec[ADDR_E], 2);     // A E c D
```
brettcoon, line 159: If you move the extra padding spaces inside the comment, e.g. from before to after the //, I think clang will allow them.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
brettcoon, line 1: Thanks for expanding this test!

clients/drcachesim/simulator/caching_device.cpp
```diff
@@ -252,6 +252,13 @@ caching_device_t::request(const memref_t &memref_in)
     }
 }
 
+void
+caching_device_t::invalidate_update(caching_device_block_t *block)
+{
+    // Clear the counter for LFU.
+    block->counter_ = 0;
```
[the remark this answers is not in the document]
jperrett256: You were right, I thought LFU relied on empty/invalid cache lines having their counters set to 0, but it doesn't. I've removed the method.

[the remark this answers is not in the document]
jperrett256: This means that all that was needed to fix this bug for FIFO/LRU was removing the line in `invalidate_caching_device_block` that clears the counter.

clients/drcachesim/tests/cache_replacement_policy_unit_test.cpp
jperrett256: Done. I don't think it's super pretty, but it's definitely readable. Thanks for the suggestion.

clients/drcachesim/simulator/cache_lru.cpp
```diff
@@ -62,6 +62,12 @@ cache_lru_t::init(int associativity, int block_size, int total_size,
     return true;
 }
 
+void
+cache_lru_t::invalidate_update(caching_device_block_t *block)
+{
+    // Leave counters unchanged for LRU.
```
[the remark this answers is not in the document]
jperrett256: I've removed the function now (based on your earlier comment), which makes it difficult to put a comment in.

I'm probably biased from having stared at this for a while, but I think it's fairly clear the LRU implementation very much relies on the counter values of invalid lines. Otherwise, why initialise the counters in each set to have unique values at the start (`cache_lru_t::init`), given they're all invalid lines?

clients/drcachesim/simulator/cache_fifo.cpp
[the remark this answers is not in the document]
jperrett256: I've removed the function now (based on your earlier comment), which makes it difficult to put a comment in.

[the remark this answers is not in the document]
jperrett256: I should say I agree the comment I had wasn't fantastic.

brettcoon (approved):
Looks good, thanks!

clients/drcachesim/simulator/cache_lru.cpp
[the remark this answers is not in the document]
brettcoon: I puzzled over that for a while, too.  I believe it's because the line replacement code expects the way counters to be unique (to specify a full ordering) at all times, even for invalid lines.  So the init code sets this property and everything else has to maintain it.
The FIFO code expects the counters to maintain the property of exactly 1 non-zero counter, and again breaks if the property is violated.

jperrett256:
Is there any update with this?

brettcoon:
My mistake, I thought you could submit, but it looks like I have to kick that off.  I started the merge-and-submit process.

PR #6051, opened 2023-05-17
drcachesim: added use_physical option to config file reader

jperrett256:
As the title says, this makes an existing option (use_physical) available in the cache simulator config file reader, so that it can be used when a config file is used to initialise the cache simulator instead of command line options.


jperrett256:
It's worth noting that cache simulator parameters that are passed on command line are ignored when using a config file. So this is not simply about making something easier, it makes it _possible_ to have complicated cache hierarchies (via config file) and use physical addresses when simulating.

clients/drcachesim/reader/config_reader.cpp
```diff
@@ -145,6 +145,18 @@ config_reader_t::configure(std::istream *config_file, cache_simulator_knobs_t &k
             } else {
                 knobs.model_coherence = false;
             }
+        } else if (param == "use_physical") {
```
derekbruening, line 148: Please add this to the documented list of parameters supported in config files (at `sec_drcachesim_config_file`).

derekbruening, line 148: It would also be nice to have a test: should just take 2 lines to get a simple sanity check in one of the existing tests: add the param to tests/single_core.conf and add a knobs check in tests/config_reader_test.cpp.

derekbruening (commented):
Thank you for the PR: just requesting single lines added to the docs and test to make it complete.

derekbruening:
> It's worth noting that cache simulator parameters that are passed on command line are ignored when using a config file. So this is not simply about making something easier, it makes it _possible_ to have complicated cache hierarchies (via config file) and use physical addresses when simulating.

Yes, we have hit this before ourselves and it is not ideal -- I don't think an issue on this exists; maybe you could file an issue requesting that command line parameters augment and override config file parameters?

clients/drcachesim/reader/config_reader.cpp
jperrett256: Thanks for pointing me to the right places, it makes this much easier. It looks like that test is never run. I can't find any mention of that cpp file in any of the CMakeLists.txt files.

jperrett256: I don't think I was. It would've failed if it had been running.

jperrett256:
Thanks, I'll make an issue in a couple weeks when I have more time.

clients/drcachesim/reader/config_reader.cpp
derekbruening: Hmm, I think you are right: commit fa4cc4a651d34e1afc8e29e8740a18563505c205 added the file but seems to have failed to either build it or make it a test!  Wow.   These are the real tests with "config" for x86-64:
```
  Test #285: code_api|tool.drcachesim.simple-config-file
  Test #288: code_api|tool.drcachesim.missfile-config-file
  Test #305: code_api|tool.drcachesim.threads-with-config-file
```
Do you feel up to adding it to drcachesim/CMakeLists.txt?  Model it on tool.drcachesim.unit_tests add_executable..add_test.

derekbruening (approved):
Thank you for enabling the test.

PR #6025, opened 2023-05-03
Fix init_scheduler call-site in analyzer_t

abhinav92003:
Adds missing only_thread arg value in call-site for init_scheduler. We
need to re-specify the default value for only_thread if we want to specify
verbosity.

PR #6052, opened 2023-05-17
drcachesim: minor fix in function call (wrong arguments passed)

jperrett256:
The wrong arguments are being passed in this function call.
This coincidentally doesn't cause issues in the default case, as the default values for both of the integer arguments are 0 IIRC.

There is a larger issue (and I'm not sure where to mention it): serial/legacy traces don't seem to be working (at least as they used to), since the addition of this new scheduler stuff. I'm working on an older fork, where trace file data is still considated directly by a method in `file_reader_t`.
I was able to get my stuff running locally by changing the default serial `scheduler_options_t` returned by `make_scheduler_serial_options()` to contain `MAP_TO_CONSISTENT_OUTPUT` and `DEPENDENCY_IGNORE` (instead of `MAP_TO_RECORDED_OUTPUT` and `DEPENDENCY_TIMESTAMPS`).
I'm assuming those default values are what they are for a reason, and I don't have time to produce a general fix for this at the moment. I thought I would mention in case the people who know more about this can find time to provide a fix.


derekbruening:
> There is a larger issue (and I'm not sure where to mention it): serial/legacy traces don't seem to be working (at least as they used to), since the addition of this new scheduler stuff. I'm working on an older fork, where trace file data is still considated directly by a method in file_reader_t.
> I was able to get my stuff running locally by changing the default serial scheduler_options_t returned by make_scheduler_serial_options() to contain MAP_TO_CONSISTENT_OUTPUT and DEPENDENCY_IGNORE (instead of MAP_TO_RECORDED_OUTPUT and DEPENDENCY_TIMESTAMPS).
> I'm assuming those default values are what they are for a reason, and I don't have time to produce a general fix for this at the moment. I thought I would mention in case the people who know more about this can find time to provide a fix.

I'm not understanding what the issue is: could you elaborate on exactly what is wrong?  What are the symptoms?

jperrett256:
> I'm not understanding what the issue is: could you elaborate on exactly what is wrong? What are the symptoms?

It used to be the case that you could run these very simple traces offline. These were gzipped traces, containing lots of `trace_entry_t`s. You needed to add a header, tid, pid, timestamp and cpuid markers, but then you could just insert the actual instruction and data accesses. If you had multiple files, it would merge them into a single stream of `trace_entry_t`s based on timestamp markers. All of this logic was handled in `file_reader_t` (dynamorio/clients/drcachesim/reader/file_reader.h), prior to commit `f59f1b3a2183ba837177d1cac62ce7b706ed555e`.

I am saying that if I now run `bin64/drrun -t drcachesim -infile <path to simple trace> -config_file <path to cache config file>`, it fails to read anything whatsoever. The call to `scheduler_tmpl_t<RecordType, ReaderType>::stream_t::next_record` in `analyzer_tmpl_t<RecordType, ReaderType>::process_serial` (dynamorio/clients/drcachesim/analyzer.cpp) does not return `sched_type_t::STATUS_OK`, and so it quits immediately. It then shows me the results of my cache simulation, for which the statistics are all zeros. (The L1 statistics are hidden, because the L1s are never accessed. I assume this change was to avoid clutter when you have many cores, some of which are unused.)

Running drcachesim with `-infile` causes it to set up the scheduler with the default "serial" scheduler options (calls `make_scheduler_serial_options`). I believe the problem is that the default "serial" scheduler options are not appropriate, at least when running with `-infile`. This is based on the fact that changing them, as described above, fixes my issue.

I don't actually know though, since I haven't looked into what exactly the scheduler is supposed to solve. It's clearly designed to be used for many use cases, and I have no idea what they are. So I will leave that with the people who a) know more and b) have time.

jperrett256:
Just to be clear, this particular change does not address what I've described. It's only relevant in the sense that it's on the execution path when `-infile` is used. It just fixes what appears to be a typo involving optional arguments resulting arguments being passed incorrectly.

derekbruening (approved):
The code looks good.  Xref PR #6025 on a similar issue.

derekbruening:
> > I'm not understanding what the issue is: could you elaborate on exactly what is wrong? What are the symptoms?
> 
> It used to be the case that you could run these very simple traces offline. These were gzipped traces, containing lots of `trace_entry_t`s. You needed to add a header, tid, pid, timestamp and cpuid markers, but then you could just insert the actual instruction and data accesses.

You mean you were synthesizing your own files, rather than getting them by tracing an application.

> If you had multiple files, it would merge them into a single stream of `trace_entry_t`s based on timestamp markers. All of this logic was handled in `file_reader_t` (dynamorio/clients/drcachesim/reader/file_reader.h), prior to commit `f59f1b3a2183ba837177d1cac62ce7b706ed555e`.

The same merging by timestamp is now in the scheduler.
 
> I am saying that if I now run `bin64/drrun -t drcachesim -infile <path to simple trace> -config_file <path to cache config file>`, it fails to read anything whatsoever.

`-infile` as a general parameter works fine today: we have regression tests that use it.  So I would assume the issue is due to the format of these files you are using.  Ideally a better error message would be propagated.  Running with `-verbose 4` may help. 

If this is due to a custom format in your files, I would suggest adding support to our repository and adding tests.  We can only reasonably maintain support for what is in the github tests here, and all our tests of serial mode whether with -infile or -indir or whatnot pass.

Maybe it would be best to file an issue and move discussion there.

clients/drcachesim/analyzer_multi.cpp
```diff
@@ -143,7 +143,8 @@ analyzer_multi_t::analyzer_multi_t()
         }
     } else {
         // Legacy file.
-        if (!init_scheduler(op_infile.get_value(), op_verbose.get_value()))
+        if (!init_scheduler(op_infile.get_value(),
+            INVALID_THREAD_ID /*all threads*/, op_verbose.get_value()))
```
derekbruening, line 147: As the tests show this fails clang-format: see https://github.com/DynamoRIO/dynamorio/actions/runs/5009310064/jobs/8978112282?pr=6052

jperrett256:
Thanks! I'll submit an issue / add tests in a couple weeks when I have more time.

For (my) future reference, it looks like `test_serial` (dynamorio/clients/drcachesim/tests/scheduler_unit_tests.cpp) might be useful to look at.

PR #6056, opened 2023-05-17
i#6003: Only accept FS/GS segment prefixes on X64.

### LINKED ISSUE #6003: Incorrect decoding of FS prefix
Drdecode decodes `26 26 26 26 26 26 26 64 26 26 26 26 13 04 0a` as `adc %es:(%rdx,%rcx)[4byte] %eax -> %eax`.

However, gdb thinks it's `fs` instead of `es`:
```
Thread 1 "fuzzer" received signal SIGSEGV, Segmentation fault.
(gdb) p /x $rip
$1 = 0x40000002
(gdb) x/i 0x40000002
=> 0x40000002:	es es es es es es es fs es es es adc %fs:(%rdx,%rcx,1),%eax
```

This is indirectly confirmed by the fact that `%es:(%rdx,%rcx)` address is valid and should not cause SIGSEGV, while fs is not valid and CPU generates an exception.

Also:
```
echo -en "\x26\x26\x26\x26\x26\x26\x26\x64\x26\x26\x26\x26\x13\x04\x0a" > /tmp/bin | objdump -mi386 -Mx86-64 -b binary -D /tmp/bin
es es es es es es es fs es es es adc %fs:(%rdx,%rcx,1),%eax
```
while llvm-mc agrees with drdecode:
```
echo -en "0x26 0x26 0x26 0x26 0x26 0x26 0x26 0x64 0x26 0x26 0x26 0x26 0x13 0x04 0x0a" | llvm-mc --disassemble -
	adcl	%es:(%rdx,%rcx), %eax
```

On commit e879d71f0be203a2297dc9578f4fd8e3a65cc36e.

khuey:
On 64 bit the legacy CS/SS/DS/ES prefixes are ignored. Account for that so instructions decode correctly.

Fixes #6003


khuey:
I'm not sure why code_api|tool.drcachesim.threads keeps failing, it passes here locally.

derekbruening:
We just turned on the new merge queue feature.  I'll add this one.  We'll see how it works this time (had issues last time).

derekbruening:
@khuey there may be a problem with this PR where `OP_jo` is supposed to (according to DR's own type `OPSZ_4_short2xi4`) honor a data size prefix if on AMD: so consider something like `66 0f 80 00 00 00 00` which DR is now (after this PR) always interpreting as a single `jo` whether on Intel or AMD. 

Other tools seem to honor the 66 even when run on Intel so they seem to default to AMD:
```
$ echo -en "0x66 0x0f 0x80 0x00 0x00 0x00 0x00" | llvm-mc-14 --disassemble -
jo 0
addb %al, (%rax)
$ echo -en "\x66\x0f\x80\x00\x00\x00\x00\x00" > /tmp/bin | objdump -mi386 -Mx86-64 -b binary -D /tmp/bin
   0: 66 0f 80 00 00        jo     0x5
   5: 00 00                add    %al,(%rax)
```
For objdump we can tell it to interpret as Intel:
```
$ echo -en "\x66\x0f\x80\x00\x00\x00\x00\x00" > /tmp/bin | objdump -mi386 -Mx86-64 -Mintel64 -b binary -D /tmp/bin
   0:   66 0f 80 00 00 00 00    data16 jo 0x7
```

I don't think we exposed a `drdisas` option for Intel vs AMD (we should add it): probably it tries to query the underlying processor.  We do have `proc_set_vendor` for programmatic setting of the target with `drdecode`.

khuey:
This PR shouldn't have affected the handling of non-segment prefixes.

Issue #6020, opened 2023-05-02
Enable analysis tools to print heartbeat stats


abhinav92003:
It'll be useful to allow drmemtrace analysis tools to print heartbeat stats, or even buffer heartbeat stats for later use.

abhinav92003:
Xref PR #6097 where we fixed excessive memory consumption by an std::set tracked by basic_counts in its interval_state_snapshot_t. This is a use case for merging shard-local interval snapshots as soon as they become available instead of waiting for the end. One complexity here is that if #workers < #shards, the whole-trace snapshots can be formed only when the last thread starts getting analyzed. But maybe we can change our analyzer-scheduler code to ensure equal progress on all shards.

PR #6057, opened 2023-05-18
i#6020: Fix TID tracking in basic_counts.

abhinav92003:
Fixes TID tracking in basic counts which is broken for data-only traces.
The breakage was caused by the recently submitted PR #6022.

Issue: #6020

### DIFF
--- a/clients/drcachesim/tests/filter-no-i.templatex
+++ b/clients/drcachesim/tests/filter-no-i.templatex
@@ -21,7 +21,7 @@ Total counts:
            0 total physical address unavailable markers
      .* total other markers
      .* total encodings
-Thread .* counts:
+Thread [1-9][0-9]* counts:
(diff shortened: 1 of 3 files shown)


clients/drcachesim/tests/filter-no-i.templatex
```diff
@@ -21,7 +21,7 @@ Total counts:
            0 total physical address unavailable markers
```
derekbruening, line 21: > Fixes TID tracking in basic counts which is broken for data-only traces.
> This was changed recently by PR https://github.com/DynamoRIO/dynamorio/pull/6022.

This description confused me into thinking this was broken prior to PR #6022 but looking at that PR it seems to directly cause the breakage.  Please reword to make it clearer that this was not a pre-existing problem.

derekbruening (approved):

PR #6058, opened 2023-05-18
i#6020: Fix incorrect args to std::move.

abhinav92003:
Fixes incorrect args to std::move. This was another bug introduced in PR #6022.

Issue: #6020

### DIFF
--- a/clients/drcachesim/analyzer.h
+++ b/clients/drcachesim/analyzer.h
@@ -125,7 +125,7 @@ protected:
         analyzer_tool_shard_data_t(analyzer_tool_shard_data_t &&src)
         {
             shard_data = src.shard_data;
-            interval_snapshot_data = std::move(interval_snapshot_data);
+            interval_snapshot_data = std::move(src.interval_snapshot_data);
         }
 
         void *shard_data;
@@ -144,7 +144,7 @@ protected:
         analyzer_shard_data_t(analyzer_shard_data_t &&src)
         {
             cur_interval_index = src.cur_interval_index;
-            tool_data = std::move(tool_data);
+            tool_data = std::move(src.tool_data);
         }
 
         uint64_t cur_interval_index;


abhinav92003:
This was pointed out by @ksco in #6022. Apologies for the breakage.

clients/drcachesim/analyzer.h
```diff
@@ -125,7 +125,7 @@ template <typename RecordType, typename ReaderType> class analyzer_tmpl_t {
         analyzer_tool_shard_data_t(analyzer_tool_shard_data_t &&src)
         {
             shard_data = src.shard_data;
-            interval_snapshot_data = std::move(interval_snapshot_data);
+            interval_snapshot_data = std::move(src.interval_snapshot_data);
```
derekbruening, line 128: What were the consequences?  Do we need to add a unit test case that would catch this?

derekbruening (approved):
What were the consequences?  Do we nee

clients/drcachesim/analyzer.h
abhinav92003: @ksco mentioned that the LicheePi4A (RISC-V SBC) compiler complained about this. I don't think this actually crashed anything.

abhinav92003: https://github.com/DynamoRIO/dynamorio/pull/6022#discussion_r1197255790

abhinav92003: Created #6060 to improve our compiler warnings.

derekbruening: But what are the actual consequences of having this bug?  What code calls this function?  What happens without doing the requested move?

derekbruening: If you're saying there are no consequences and all you're worried about is a compiler warning, then let's delete this function.

abhinav92003: We need the move constructor for the same reason we have it for analyzer_worker_data_t: for it to work with std::vector.

The bug fixed in this PR could've been problematic if the `analyzer_shard_data_t::tool_data`  vector did an internal re-allocation (in response to addition of new elements). But we allocate enough space in the vector for all tools using an upfront `resize(num_tools_)`, so it wouldn't cause any real issues I think.

The move constructor for analyzer_shard_data_t though could be deleted. The current code uses an unordered_map for storing analyzer_shard_data_t, not a vector.

I sent PR #6066 for this cleanup and some other misc stuff that should've been done before.

PR #6061, opened 2023-05-18
i#3544 RV64: Some fixes and improvements to call_switch_stack

ksco:
Some fixes and improvements to `call_switch_stack` for RV64, including:
1. Reformatted a bit
2. Used `mv` instead of `addi` for readability
3. Fixed errors of stack swapping

Issue: #3544

### DIFF
--- a/core/arch/riscv64/riscv64.asm
+++ b/core/arch/riscv64/riscv64.asm
@@ -60,31 +60,31 @@ GLOBAL_LABEL(cpuid_supported:)
         DECLARE_FUNC(call_switch_stack)
 GLOBAL_LABEL(call_switch_stack:)
         /* Init the stack. */
-        addi    sp, sp, -32
+        addi     sp, sp, -32
         /* Use two callee-saved regs to call func. */
-        sd      ra, 16(sp)
-        sd      s0, 8 (sp)
-        sd      s1, 0 (sp)
+        sd       ra, 16(sp)
+        sd       s0, 8 (sp)
+        sd       s1, 0 (sp)
         /* Check mutex_to_free. */
-        beqz    ARG4, call_dispatch_alt_stack_no_free
+        beqz     ARG4, call_dispatch_alt_stack_no_free
         /* Release the mutex. */
-        sd      x0, 0(ARG4)
+        sd       x0, 0(ARG4)
 call_dispatch_alt_stack_no_free:
         /* Copy ARG5 (return_on_return) to callee-saved reg. */
-        mv      s1, ARG5
+        mv       s1, ARG5
         /* Switch the stack. */
-        mv      s0, sp
-        mv      sp, ARG2
+        mv       s0, sp
+        mv       sp, ARG2
         /* Call func. */
-        jalr    ARG3
+        jalr     ARG3
         /* Switch stack back. */
-        mv      sp, s0
-        beqz    s1, GLOBAL_LABEL(unexpected_return)
+        mv       sp, s0
+        beqz     s1, GLOBAL_LABEL(unexpected_return)
         /* Restore the stack. */
-        ld      s1, 0 (sp)
-        ld      s0, 8 (sp)
-        ld      ra, 16(sp)
-        addi    sp, sp, 32
+        ld       s1, 0 (sp)
+        ld       s0, 8 (sp)
+        ld       ra, 16(sp)
+        addi     sp, sp, 32
         ret
         END_FUNC(call_switch_stack)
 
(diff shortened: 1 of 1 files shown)


ksco:
I would like to participate in the development of the RV64 port. I'm still getting familiar with the codebase and reviewing the code of the RV64 port that is already written.

core/arch/riscv64/riscv64.asm
```diff
@@ -60,33 +60,31 @@ GLOBAL_LABEL(cpuid_supported:)
         DECLARE_FUNC(call_switch_stack)
 GLOBAL_LABEL(call_switch_stack:)
         /* Init the stack. */
-        addi     sp, sp, -32
-        /* Use two callee-save regs to call func. */
-        sd       ra, 24 (sp)
-        sd       s0, 16 (sp)
-        sd       s1, 8 (sp)
-        sd       s2, 0 (sp)
+        addi    sp, sp, -32
```
shiptux, line 63: The alignment of the start of the operands is different to other code styles.

core/arch/riscv64/riscv64.asm
```diff
@@ -60,33 +60,31 @@ GLOBAL_LABEL(cpuid_supported:)
...
-        beqz     s2, GLOBAL_LABEL(unexpected_return)
+        mv      sp, s0
+        beqz    s1, GLOBAL_LABEL(unexpected_return)
         /* Restore the stack. */
-        ld       s2, 0 (sp)
-        ld       s1, 8 (sp)
-        ld       s0, 16 (sp)
-        ld       ra, 24 (sp)
-        addi     sp, sp, 32
+        ld      s1, 0 (sp)
```
shiptux, line 84: Ditto.

shiptux (changes requested):

core/arch/riscv64/riscv64.asm
ksco: It aligned to 8 spaces.

shiptux: I mean the operands, not the operators. For example:
```
        addi    sp, sp, -32
        ldp      x20, x30, [sp], #32
        ldp      x20, x30, [sp], #32
        addi    sp, sp, -32
```

ksco: Yes, the first operand is also aligned to 8 spaces. ARM and X64 ports are aligned to 9 spaces, which is strange. I don't think we have to follow that unless there is a reason. 8 spaces are more convenient for typing.

derekbruening: The style convention is documented at https://dynamorio.org/page_code_style.html#autotoc_md126:
```
16. In .asm files, place opcodes on column 8 and start operands on column 17.
```
Other ports are following that convention: even if you find some exception in some old code, please follow the documented style conventions to keep all new code consistent.

ksco: Oh OK, didn't noticed that.

ksco: Done. @derekbruening Could you review more?

core/arch/riscv64/riscv64.asm
```diff
@@ -61,31 +61,29 @@ GLOBAL_LABEL(cpuid_supported:)
 GLOBAL_LABEL(call_switch_stack:)
         /* Init the stack. */
         addi     sp, sp, -32
```
shiptux, line 63: Should be `addi  sp, sp, -24` here?
```
________ Stack Bottom
|_s_ra_| 
|__s0__| 
|__s1__| _____SP
```

ksco: No, `sp` needs to be aligned to 16bytes.

core/arch/riscv64/riscv64.asm
```diff
@@ -61,31 +61,29 @@ GLOBAL_LABEL(cpuid_supported:)
...
-        /* Use two callee-save regs to call func. */
-        sd       ra, 24 (sp)
-        sd       s0, 16 (sp)
-        sd       s1, 8 (sp)
-        sd       s2, 0 (sp)
+        /* Use two callee-saved regs to call func. */
+        sd       ra, 16(sp)
```
abhinav92003, line 65: nit: since ra is not callee-saved, it's a little bit clearer if `sd ra, 16(sp)` is moved before the comment above (which references callee-saved regs).

abhinav92003 (approved):

core/arch/riscv64/riscv64.asm
ksco: Yes, moved.

shiptux (approved):

PR #6062, opened 2023-05-18
i#3544 RV64: Implemented dr_call_on_clean_stack

ksco:
This patch implemented `dr_call_on_clean_stack` for RV64.

Issue: #3544

### DIFF
--- a/core/arch/riscv64/riscv64.asm
+++ b/core/arch/riscv64/riscv64.asm
@@ -129,8 +129,8 @@ GLOBAL_LABEL(dr_call_on_clean_stack:)
         mv       ARG4, ARG6          /* void *arg4 */
         mv       ARG5, ARG7          /* void *arg5 */
         mv       ARG6, ARG8          /* void *arg6 */
-        ld       ARG7, 0(s0)         /* void *arg7, retrived from the stack */
-        ld       ARG8, ARG_SZ(s0)    /* void *arg8, retrived from the stack */
+        ld       ARG7, 2*ARG_SZ(s0)  /* void *arg7, retrived from the stack */
+        ld       ARG8, 3*ARG_SZ(s0)  /* void *arg8, retrived from the stack */
         jalr     ra
         /* Swap stacks. */
         mv       sp, s0
--- a/core/arch/riscv64/riscv64.asm
+++ b/core/arch/riscv64/riscv64.asm
@@ -40,6 +40,11 @@ START_FILE
 # error Non-Linux is not supported
 #endif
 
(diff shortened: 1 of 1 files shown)


shiptux:
It's not recommended to use force-push. The commit message is cleaned up in the final squash-and-merge so there is no benefit and only negatives to force-pushing a PR.

ksco:
1) there is only 1 commit in this PR, 2) this PR is not yet reviewed, it will keep the commit history clean for review. Besides you can always use the "compare" button to know the changes.

derekbruening:
Adding to https://github.com/DynamoRIO/dynamorio/pull/6062#issuecomment-1552797079: as documented at https://dynamorio.org/page_code_reviews.html force pushes should not be used once a branch is shared as in a PR.

derekbruening:
> OK, then should I rebase or just merge when I need to sync changes from master into the PR branch?

Presumably you've run devsetup.sh and would use `git pullall` as documented: https://dynamorio.org/page_workflow.html#autotoc_md329 (does several things; master update is a merge)

core/arch/riscv64/riscv64.asm
```diff
@@ -110,7 +115,28 @@ call_dispatch_alt_stack_no_free:
...
+        ld       sp, dstack_OFFSET(ARG1)
+        /* Set up args. */
+        mv       ra,   ARG2 /* void *(*func)(arg1...arg8) */
+        mv       ARG1, ARG3          /* void *arg1 */
+        mv       ARG2, ARG4          /* void *arg2 */
+        mv       ARG3, ARG5          /* void *arg3 */
+        mv       ARG4, ARG6          /* void *arg4 */
+        mv       ARG5, ARG7          /* void *arg5 */
+        mv       ARG6, ARG8          /* void *arg6 */
+        ld       ARG7, 0(s0)         /* void *arg7, retrived from the stack */
```
abhinav92003, line 132: If I'm understand this correctly, arg7 and arg8 are expected to be passed on stack to dr_call_on_clean_stack. Here we're retrieving them from the stack top. But doesn't the stack top actually contain ra and s0 since we pushed them above?

abhinav92003 (changes requested):

core/arch/riscv64/riscv64.asm
ksco: Yes, I think you're right, we should retrieve them from `16(s0)` and `24(s0)`.

ksco:
Seems I cannot merge this myself, @abhinav92003 please merge this for me.

PR #6064, opened 2023-05-18
i#3544 RV64: Implemented more global assembly functions

ksco:
This patch implemented the following global assembly functions for RV64:
1. `dr_app_start`
2. `dr_app_take_over`
3. `dr_app_running_under_dynamorio`
4. `dynamorio_app_take_over`

Issue: #3544

### DIFF
--- a/core/arch/interp.c
+++ b/core/arch/interp.c
@@ -3147,6 +3147,7 @@ mangle_pre_client(dcontext_t *dcontext, build_bb_t *bb)
         instr_t *ret = instrlist_last(bb->ilist);
         instr_t *mov = instr_get_prev(ret);
         LOG(THREAD, LOG_INTERP, 3, "Found dr_app_running_under_dynamorio\n");
+        // clang-format off
         ASSERT(ret != NULL && instr_is_return(ret) && mov != NULL &&
             IF_X86(instr_get_opcode(mov) == OP_mov_imm &&)
             IF_ARM(instr_get_opcode(mov) == OP_mov &&
@@ -3156,6 +3157,7 @@ mangle_pre_client(dcontext_t *dcontext, build_bb_t *bb)
                 bb->start_pc == instr_get_raw_bits(mov) ||
                 /* the translation field might be NULL */
                 bb->start_pc == instr_get_translation(mov)));
+        // clang-format on
         /* i#1998: ensure the instr is Level 3+ */
         instr_decode(dcontext, mov);
         instr_set_src(mov, 0, OPND_CREATE_INT32(1));
--- a/core/arch/interp.c
+++ b/core/arch/interp.c
@@ -3151,11 +3151,11 @@ mangle_pre_client(dcontext_t *dcontext, build_bb_t *bb)
                IF_X86(instr_get_opcode(mov) == OP_mov_imm &&)
                    IF_ARM(instr_get_opcode(mov) == OP_mov &&
                           OPND_IS_IMMED_INT(instr_get_src(mov, 0)) &&)
-                       IF_AARCH64_ELSE(instr_get_opcode(mov) == OP_movz &&,
-                                       IF_RISCV64(instr_get_opcode(mov) == OP_addi &&))(
-                           bb->start_pc == instr_get_raw_bits(mov) ||
-                           /* the translation field might be NULL */
-                           bb->start_pc == instr_get_translation(mov)));
+                       IF_AARCH64(instr_get_opcode(mov) == OP_movz &&)
+                           IF_RISCV64(instr_get_opcode(mov) == OP_addi &&)(
+                               bb->start_pc == instr_get_raw_bits(mov) ||
+                               /* the translation field might be NULL */
+                               bb->start_pc == instr_get_translation(mov)));
         /* i#1998: ensure the instr is Level 3+ */
         instr_decode(dcontext, mov);
         instr_set_src(mov, 0, OPND_CREATE_INT32(1));
--- a/core/arch/interp.c
+++ b/core/arch/interp.c
@@ -3147,17 +3147,15 @@ mangle_pre_client(dcontext_t *dcontext, build_bb_t *bb)
         instr_t *ret = instrlist_last(bb->ilist);
         instr_t *mov = instr_get_prev(ret);
         LOG(THREAD, LOG_INTERP, 3, "Found dr_app_running_under_dynamorio\n");
-        // clang-format off
         ASSERT(ret != NULL && instr_is_return(ret) && mov != NULL &&
-            IF_X86(instr_get_opcode(mov) == OP_mov_imm &&)
-            IF_ARM(instr_get_opcode(mov) == OP_mov &&
-                OPND_IS_IMMED_INT(instr_get_src(mov, 0)) &&)
-            IF_AARCH64(instr_get_opcode(mov) == OP_movz &&)
-            IF_RISCV64(instr_get_opcode(mov) == OP_addi &&) (
-                bb->start_pc == instr_get_raw_bits(mov) ||
-                /* the translation field might be NULL */
-                bb->start_pc == instr_get_translation(mov)));
-        // clang-format on
+               IF_X86(instr_get_opcode(mov) == OP_mov_imm &&)
+                   IF_ARM(instr_get_opcode(mov) == OP_mov &&
+                          OPND_IS_IMMED_INT(instr_get_src(mov, 0)) &&)
+                       IF_AARCH64_ELSE(instr_get_opcode(mov) == OP_movz &&,
+                                       IF_RISCV64(instr_get_opcode(mov) == OP_addi &&))(
+                           bb->start_pc == instr_get_raw_bits(mov) ||
+                           /* the translation field might be NULL */
+                           bb->start_pc == instr_get_translation(mov)));
         /* i#1998: ensure the instr is Level 3+ */
         instr_decode(dcontext, mov);
         instr_set_src(mov, 0, OPND_CREATE_INT32(1));
--- a/core/arch/interp.c
+++ b/core/arch/interp.c
@@ -3148,13 +3148,14 @@ mangle_pre_client(dcontext_t *dcontext, build_bb_t *bb)
         instr_t *mov = instr_get_prev(ret);
         LOG(THREAD, LOG_INTERP, 3, "Found dr_app_running_under_dynamorio\n");
         ASSERT(ret != NULL && instr_is_return(ret) && mov != NULL &&
-               IF_X86(instr_get_opcode(mov) == OP_mov_imm &&)
-                   IF_ARM(instr_get_opcode(mov) == OP_mov &&
-                          OPND_IS_IMMED_INT(instr_get_src(mov, 0)) &&)
-                       IF_AARCH64(instr_get_opcode(mov) == OP_movz &&)(
-                           bb->start_pc == instr_get_raw_bits(mov) ||
-                           /* the translation field might be NULL */
-                           bb->start_pc == instr_get_translation(mov)));
+            IF_X86(instr_get_opcode(mov) == OP_mov_imm &&)
+            IF_ARM(instr_get_opcode(mov) == OP_mov &&
+                OPND_IS_IMMED_INT(instr_get_src(mov, 0)) &&)
+            IF_AARCH64(instr_get_opcode(mov) == OP_movz &&)
+            IF_RISCV64(instr_get_opcode(mov) == OP_addi &&) (
+                bb->start_pc == instr_get_raw_bits(mov) ||
+                /* the translation field might be NULL */
+                bb->start_pc == instr_get_translation(mov)));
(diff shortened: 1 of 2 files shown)


core/arch/riscv64/riscv64.asm
```diff
@@ -118,29 +128,140 @@ GLOBAL_LABEL(dr_call_on_clean_stack:)
...
+        /* Preverse stack space. */
+        addi     sp, sp, -PRIV_MCONTEXT_SIZE
+        /* Push x3 on to stack to use it as a scratch. */
+        sd       x3, 4*ARG_SZ(sp)
+        /* Compute original sp. */
+        addi     x3, sp, PRIV_MCONTEXT_SIZE+16
+        /* Save link register on to stack. */
+        sd       ra, 1*ARG_SZ(sp)
+        /* Save original sp on to stack. */
+        sd       x3, 2*ARG_SZ(sp)
```
abhinav92003, line 216: Move this up immediately after x3 is computed.

core/arch/riscv64/riscv64.asm
```diff
@@ -118,29 +128,140 @@ GLOBAL_LABEL(dr_call_on_clean_stack:)
...
+        /* Save ra as pc */
+        sd       ra, 32*ARG_SZ(sp)
+        CALLC1(save_priv_mcontext_helper, sp)
```
abhinav92003, line 219: We're passing sp here, and `save_priv_mcontext_helper` starts storing at `1*ARG_SZ(ARG1)`. Won't that overwrite what we wrote here at `1*ARG_SZ(sp)`?

core/arch/riscv64/riscv64.asm
```diff
@@ -118,29 +128,140 @@ GLOBAL_LABEL(dr_call_on_clean_stack:)
...
+        /* Preverse stack space. */
+        addi     sp, sp, -PRIV_MCONTEXT_SIZE
+        /* Push x3 on to stack to use it as a scratch. */
+        sd       x3, 4*ARG_SZ(sp)
+        /* Compute original sp. */
+        addi     x3, sp, PRIV_MCONTEXT_SIZE+16
+        /* Save link register on to stack. */
+        sd       ra, 1*ARG_SZ(sp)
+        /* Save original sp on to stack. */
+        sd       x3, 2*ARG_SZ(sp)
```
abhinav92003, line 256: Same here: move this up immediately after the computation of x3.

core/arch/riscv64/riscv64.asm
```diff
@@ -118,29 +128,140 @@ GLOBAL_LABEL(dr_call_on_clean_stack:)
...
+        /* Save ra as pc */
+        sd       ra, 32*ARG_SZ(sp)
+        CALLC1(save_priv_mcontext_helper, sp)
```
abhinav92003, line 259: Same concern here: are we clobbering regs already preserved (since `save_priv_mcontext_helper` will overwrite `1*ARG_SZ(ARG1)` which we wrote above at `sd ra, 1*ARG_SZ(sp)`)? Or am I missing something.

core/arch/interp.c
```diff
@@ -3147,14 +3147,17 @@ mangle_pre_client(dcontext_t *dcontext, build_bb_t *bb)
...
-                   IF_ARM(instr_get_opcode(mov) == OP_mov &&
-                          OPND_IS_IMMED_INT(instr_get_src(mov, 0)) &&)
-                       IF_AARCH64(instr_get_opcode(mov) == OP_movz &&)(
-                           bb->start_pc == instr_get_raw_bits(mov) ||
-                           /* the translation field might be NULL */
-                           bb->start_pc == instr_get_translation(mov)));
+            IF_X86(instr_get_opcode(mov) == OP_mov_imm &&)
+            IF_ARM(instr_get_opcode(mov) == OP_mov &&
+                OPND_IS_IMMED_INT(instr_get_src(mov, 0)) &&)
+            IF_AARCH64(instr_get_opcode(mov) == OP_movz &&)
```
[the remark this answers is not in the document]
ksco: Done in https://github.com/DynamoRIO/dynamorio/pull/6064/commits/78845cf30b59835b573fdde376ad184307299dca.

core/arch/interp.c
```diff
@@ -3147,14 +3147,17 @@ mangle_pre_client(dcontext_t *dcontext, build_bb_t *bb)
         instr_t *ret = instrlist_last(bb->ilist);
         instr_t *mov = instr_get_prev(ret);
         LOG(THREAD, LOG_INTERP, 3, "Found dr_app_running_under_dynamorio\n");
+        // clang-format off
```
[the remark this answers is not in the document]
ksco: Removed in https://github.com/DynamoRIO/dynamorio/pull/6064/commits/78845cf30b59835b573fdde376ad184307299dca.

core/arch/riscv64/riscv64.asm
ksco: Moved up in https://github.com/DynamoRIO/dynamorio/pull/6064/commits/af6a2333cd0d586a591b17d09ef382faf184c5a6.

core/arch/riscv64/riscv64.asm
ksco: Move up in https://github.com/DynamoRIO/dynamorio/pull/6064/commits/af6a2333cd0d586a591b17d09ef382faf184c5a6.

core/arch/riscv64/riscv64.asm
ksco: Yes, you're right, I fixed this in https://github.com/DynamoRIO/dynamorio/pull/6064/commits/ab6cdc5398253654fe9e68904694549b32e42635.

core/arch/riscv64/riscv64.asm
ksco: Fixed in https://github.com/DynamoRIO/dynamorio/pull/6064/commits/ab6cdc5398253654fe9e68904694549b32e42635.

core/arch/riscv64/riscv64.asm
```diff
@@ -128,73 +128,74 @@ GLOBAL_LABEL(dr_call_on_clean_stack:)
...
-        fsd      f30, 60*ARG_SZ(ARG1)
-        fsd      f31, 61*ARG_SZ(ARG1)
+        sd       x4,   4*ARG_SZ(ARG1)
+        sd       x5,   5*ARG_SZ(ARG1)
+        sd       x6,   6*ARG_SZ(ARG1)
+        sd       x7,   7*ARG_SZ(ARG1)
+        sd       x8,   8*ARG_SZ(ARG1)
+        sd       x9,   9*ARG_SZ(ARG1)
+        /* a0 (10*ARG_SZ) is already saved. */
+        sd       x10, 11*ARG_SZ(ARG1)
```
abhinav92003, line 143: The slot and register don't match. I think we want to tore x11 here?

core/arch/riscv64/riscv64.asm
```diff
@@ -128,73 +128,74 @@ GLOBAL_LABEL(dr_call_on_clean_stack:)
...
-        fsd      f29, 59*ARG_SZ(ARG1)
-        fsd      f30, 60*ARG_SZ(ARG1)
-        fsd      f31, 61*ARG_SZ(ARG1)
+        sd       x4,   4*ARG_SZ(ARG1)
+        sd       x5,   5*ARG_SZ(ARG1)
+        sd       x6,   6*ARG_SZ(ARG1)
+        sd       x7,   7*ARG_SZ(ARG1)
+        sd       x8,   8*ARG_SZ(ARG1)
+        sd       x9,   9*ARG_SZ(ARG1)
+        /* a0 (10*ARG_SZ) is already saved. */
```
abhinav92003, line 142: The comment above `save_priv_mcontext_helper:` says that x11 is already stored, but here we say a0 (x10) is, which is conflicting. I think x10 (arg1) is already stored, so the comment above needs to be updated?

core/arch/riscv64/riscv64.asm
ksco: Yes, sorry.

core/arch/riscv64/riscv64.asm
```diff
@@ -128,73 +128,74 @@ GLOBAL_LABEL(dr_call_on_clean_stack:)
...
-        fsd      f23, 53*ARG_SZ(ARG1)
-        fsd      f24, 54*ARG_SZ(ARG1)
-        fsd      f25, 55*ARG_SZ(ARG1)
-        fsd      f26, 56*ARG_SZ(ARG1)
-        fsd      f27, 57*ARG_SZ(ARG1)
-        fsd      f28, 58*ARG_SZ(ARG1)
-        fsd      f29, 59*ARG_SZ(ARG1)
-        fsd      f30, 60*ARG_SZ(ARG1)
-        fsd      f31, 61*ARG_SZ(ARG1)
+        sd       x4,   4*ARG_SZ(ARG1)
```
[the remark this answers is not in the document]
ksco: Yes, I think it's ok, and it keeps the size aligned to 16 bytes.

core/arch/riscv64/riscv64.asm
ksco: Sorry for my carelessness, fixed.

core/arch/interp.c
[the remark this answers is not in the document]
ksco: Yeah, I removed the `clang-format off/on` pair, it's now what clang-format thinks it should be. Maybe we keep it this way for now? I'll figure out something more nice later.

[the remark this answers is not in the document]
ksco: I realize I might misunderstand you, what do you mean by the existing one? Not sure if I follow.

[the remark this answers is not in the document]
ksco: Something like `IF_X86_ELSE(... &&, IF_ARM_ELSE(... &&, IF_AARCH64_ELSE(.. &&, IF_RISCV64(...))`?

core/arch/riscv64/riscv64.asm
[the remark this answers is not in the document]
abhinav92003: I don't understand. Storing a zero at 0*ARG_SZ(ARG1) won't change any alignment.

Am I correct to assume that the `0*ARG_SZ(ARG1)` slot corresponds to x0? Are we storing anything there currently?

core/arch/riscv64/riscv64.asm
```diff
@@ -40,6 +40,16 @@ START_FILE
 # error Non-Linux is not supported
 #endif
 
+/* sizeof(priv_mcontext_t) rounded up to a multiple of 16 */
+#define PRIV_MCONTEXT_SIZE 0x410
```
abhinav92003, line 44: Please add to the comment how this computes to be 0x410 (1040) bytes.
We have 32 int regs, 32 float regs, 1 fcsr, 1 pc, all 8 bytes (I think), that's 528 bytes?

Or do we reserve space for the MCXT_NUM_SIMD_SLOTS simd regs even though we don't save them below? If so, please add to comment.

abhinav92003 (changes requested):

core/arch/riscv64/riscv64.asm
ksco: Yes, we do have some reserved space for SIMD, and it won't compile if removed. I added that comment in [cd9d282](https://github.com/DynamoRIO/dynamorio/pull/6064/commits/cd9d28256b90f4b95f30fea7e1ba92d4d818df75).

core/arch/riscv64/riscv64.asm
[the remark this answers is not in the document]
ksco: So, as the comment said, the `sizeof(priv_mcontext_t)` needs to be rounded up to a multiple of 16. We have 32 gprs (including zero), 32 fprs, 1 pc, 1 fcsr, and some reserved space for SIMD, it's already aligned to 16 bytes if we keep the `zero` register, so there is no harm to keep it.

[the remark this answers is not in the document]
ksco: > Am I correct to assume that the 0*ARG_SZ(ARG1) slot corresponds to x0? 

Yes.

> Are we storing anything there currently?

No, it's unused.

core/arch/riscv64/riscv64.asm
ksco: Seems I computed this wrong somehow, 528 + 8*16 = 656, which is 0x290.

core/arch/riscv64/riscv64.asm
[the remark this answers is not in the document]
abhinav92003: > if we keep the zero register, so there is no harm to keep it.

I agree.

> > Are we storing anything there currently?

> No, it's unused.

I meant that we can actually store a zero in that slot. Optional.

core/arch/riscv64/riscv64.asm
abhinav92003: Okay, thanks for finding and correcting.

abhinav92003 (approved):

PR #6065, opened 2023-05-18
i#3544 riscv64: Implemented atomic_swap and Improved atomic_add

shiptux:

- Implemented atomic_swap
- Improved atomic_add

Issue: #3544


core/arch/riscv64/riscv64.asm
```diff
@@ -233,7 +233,10 @@ GLOBAL_LABEL(dr_longjmp:)
         /* int atomic_swap(int *adr, int val) */
         DECLARE_FUNC(atomic_swap)
 GLOBAL_LABEL(atomic_swap:)
-/* FIXME i#3544: Not implemented */
+1:      lr.d     ARG3, (ARG1)
+        sc.d     ARG4, ARG2, (ARG1)
+        bnez     ARG4, 1b
+        mv       ARG1, ARG3
```
ksco, line 239: Just use `amoswap.d`.

shiptux: Maybe amoswap.d has a better performance. But the instruction is too long.
It looks like this:
```
         amoswap.d     ARG1, ARG2, (ARG1)
1:      lr.d     ARG3, (ARG1)
        sc.d     ARG4, ARG2, (ARG1)
        bnez     ARG4, 1b
        mv       ARG1, ARG3
```

core/arch/riscv64/riscv64.asm
```diff
@@ -233,7 +233,10 @@ GLOBAL_LABEL(dr_longjmp:)
         /* int atomic_swap(int *adr, int val) */
```
abhinav92003, line 233: Re the PR title:
s/Implemented atmoic_swap/Implemented atomic_swap/

core/arch/riscv64/riscv64.asm
abhinav92003: > But the instruction is too long.

I don't understand the concern. With `amoswap` it takes just one instruction to accomplish the `atomic_swap`. 

Also, in your snippet I see `amoswap` and the lr-sc sequence both; however we just need one of them right?

abhinav92003 (changes requested):

core/arch/riscv64/riscv64.asm
shiptux: > Also, in your snippet I see amoswap and the lr-sc sequence both;
Only one is needed.

>  With amoswap it takes just one instruction to accomplish the atomic_swap.
These two options are both viable,  amoswap has better readability. Here, I would like to discuss whether the use of amoswap exceeds Dynamorio's coding style conventions(In .asm files, place opcodes on column 8 and start operands on column 17.). Maybe we should make an exception?

core/arch/riscv64/riscv64.asm
shiptux: Thanks.

core/arch/riscv64/riscv64.asm
abhinav92003: Yes, it should be fine.

core/arch/riscv64/riscv64.asm
```diff
@@ -165,10 +165,7 @@ GLOBAL_LABEL(cleanup_and_terminate:)
         /* void atomic_add(int *adr, int val) */
         DECLARE_FUNC(atomic_add)
 GLOBAL_LABEL(atomic_add:)
-1:      lr.d       a2, (a0)
-        add        a2, a2, a1
-        sc.d       a3, a2, (a0)
-        bnez       a3, 1b
+        amoswap.d      ARG1, ARG2, 0 (ARG1)
```
abhinav92003, line 168: Shouldn't this be amoadd.d?

core/arch/riscv64/riscv64.asm
```diff
@@ -233,7 +230,7 @@ GLOBAL_LABEL(dr_longjmp:)
         /* int atomic_swap(int *adr, int val) */
         DECLARE_FUNC(atomic_swap)
 GLOBAL_LABEL(atomic_swap:)
-/* FIXME i#3544: Not implemented */
+        amoswap.d      ARG1, ARG2, 0 (ARG1)
```
abhinav92003, line 233: We're using the same reg for the source addr and destination. Can that be problematic? Looking at the RISC-V manual on amo operations:

```
load a data value from the address in rs1, place the value into register rd,
apply a binary operator to the loaded value and the original value in rs2, then store the result back
to the address in rs1.
```
My understanding of the above: if rs1 is same as rd, then the loaded value will overwrite the address and the address will be lost, and the later store back into rs1 wouldn't work as expected.

It'll probably be best to try this sequence out in a small asm code.

abhinav92003 (changes requested):

core/arch/riscv64/riscv64.asm
ksco: No, it won't be a problem. Also, you can look at [how Spike implemented it](https://github.com/riscv-software-src/riscv-isa-sim/blob/master/riscv/insns/amoswap_d.h) to be sure, for `amoswap`, the address in RS1 is precomputed. The spec tends to not give any pseudo code for reference, so Spike is the golden reference.

abhinav92003 (approved):

core/arch/riscv64/riscv64.asm
shiptux: > We're using the same reg for the source addr and destination. Can that be problematic? Looking at the RISC-V manual on amo operations:
> 
> ```
> load a data value from the address in rs1, place the value into register rd,
> apply a binary operator to the loaded value and the original value in rs2, then store the result back
> to the address in rs1.
> ```
> 
> My understanding of the above: if rs1 is same as rd, then the loaded value will overwrite the address and the address will be lost, and the later store back into rs1 wouldn't work as expected.
> 
> It'll probably be best to try this sequence out in a small asm code.

Some test codes runs in qemu:
```
    #...
    la   a0, print_fmt
    la   t0, data
    li   t2, 12
    amoadd.w t2, t2, 0(t0)
    mv   a1, t2
    call printf
    #...
        la a0, print_fmt
    la t0, data
    addi t2, zero, 1
    amoswap.w t2, t2, 0(t0)
    lw   a1, 0(t0)
    call printf
    #...
```
When I test the code `amoadd.w t0, t2, (t0)`. And then I got a SIGSEGV in ` lw   a1, 0(t0)`. So I think if `amoadd.w t2, t2, (t0)` would be better? Do we really need the value of rd if we only care about the correct operation on *rs1? After all, rd is just a copy of rs1. Isn't that right?

abhinav92003: Thanks for checking.

PR #6067, opened 2023-05-18
i#6020: Add TID to basic counts interval snapshot

abhinav92003:
Adds the TID to the basic_counts_t::count_snapshot_t struct. This is to
allow the tool to figure out the TID the snapshot belongs to if needed.

Issue: #6020

### DIFF
--- a/clients/drcachesim/tools/basic_counts.cpp
+++ b/clients/drcachesim/tools/basic_counts.cpp
@@ -305,6 +305,7 @@ basic_counts_t::generate_shard_interval_snapshot(void *shard_data, uint64_t inte
     for (const auto &ctr : per_shard->counters) {
         snapshot->counters += ctr;
     }
+    snapshot->tid = per_shard->tid;
     return snapshot;
 }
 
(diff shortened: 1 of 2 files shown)


clients/drcachesim/tools/basic_counts.cpp
```diff
@@ -317,6 +318,7 @@ basic_counts_t::generate_interval_snapshot(uint64_t interval_id)
             snapshot->counters += ctr;
         }
     }
+    snapshot->tid = 0;
```
derekbruening, line 321: Should this be INVALID_THREAD_ID?

derekbruening (approved):

clients/drcachesim/tools/basic_counts.cpp
abhinav92003: I think INVALID_THREAD_ID means something that is not initialized but may be later. But here we just have a convention to set the tid to zero for serial. I'll add a comment instead.

PR #6068, opened 2023-05-18
i#6020 interval analysis: Provide convenient instr counts

abhinav92003:
Provides instr counts cumulative till an interval and the incremental delta vs
the previous interval, in the analysis_tool_t::interval_state_snapshot_t
struct. This makes it easier for tools to compute per-kilo instr metrics,
without having to duplicate instr counting for intervals.

Also, adds comparison of the interval framework provided value with the
basic counts value, under the verbose flag.

Issue: #6020

### DIFF
--- a/clients/drcachesim/tools/basic_counts.cpp
+++ b/clients/drcachesim/tools/basic_counts.cpp
@@ -350,6 +350,20 @@ basic_counts_t::print_interval_results(
         diff -= last;
         print_counters(diff, 0, " interval delta");
         last = snapshot->counters;
+        if (knob_verbose_ > 0) {
+            if (snapshot->instr_count_cumulative !=
+                static_cast<uint64_t>(snapshot->counters.instrs)) {
+                std::cerr << "Cumulative instr count value provided by framework ("
+                          << snapshot->instr_count_cumulative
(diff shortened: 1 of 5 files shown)


clients/drcachesim/tools/basic_counts.cpp
```diff
@@ -350,6 +350,20 @@ basic_counts_t::print_interval_results(
         diff -= last;
         print_counters(diff, 0, " interval delta");
         last = snapshot->counters;
+        if (knob_verbose_ > 0) {
+            if (snapshot->instr_count_cumulative !=
+                static_cast<uint64_t>(snapshot->counters.instrs)) {
+                std::cerr << "Cumulative instr count value provided by framework ("
+                          << snapshot->instr_count_cumulative
+                          << ") not equal to tool value (" << snapshot->counters.instrs
+                          << ")\n";
```
derekbruening, line 359: Will anyone notice this message if this returns true?  I would expect this to set error_string_ and return false.  Ditto below.

derekbruening (approved):

clients/drcachesim/tools/basic_counts.cpp
abhinav92003: Done.

PR #6069, opened 2023-05-18
Silence an unused variable warning that only appears on x86.

khuey:
This file doesn't include core/utils.h so the usual UNUSED macro isn't available.

This is a hack, maybe there's something better to do.


core/ir/instr_inline_api.h
```diff
@@ -156,7 +156,7 @@ opnd_is_predicate_reg(opnd_t op)
     return IF_AARCH64_ELSE(op.kind == REG_kind &&
                                op.value.reg_and_element_size.reg >= DR_REG_P0 &&
                                op.value.reg_and_element_size.reg <= DR_REG_P15,
-                           false);
+                           false && op.kind == REG_kind);
```
derekbruening, line 159: Add a comment explaining why this is here (future readers are going to puzzle over this otherwise).

derekbruening (approved):

core/ir/instr_inline_api.h
khuey: Done.

PR #6070, opened 2023-05-19
 i#3544 RV64: Implemented a naive version of memfuncs

ksco:
This patch implements `memcpy` and `memset` for RV64.

`memset` is used in `drrun` when parsing config files at startup. After this patch, it can successfully parse the config file, and then segfaults in `execv`(expected, as the RV64 port is far from complete).

Issue: #3544

### DIFF
--- a/core/arch/riscv64/memfuncs.asm
+++ b/core/arch/riscv64/memfuncs.asm
@@ -44,16 +44,30 @@ START_FILE
  */
         DECLARE_FUNC(memcpy)
 GLOBAL_LABEL(memcpy:)
-/* FIXME i#3544: Not implemented */
-        ret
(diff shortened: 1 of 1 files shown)


core/arch/riscv64/memfuncs.asm
```diff
@@ -44,16 +44,30 @@ START_FILE
  */
```
abhinav92003, line 44: nit re the title: drop the parenthesis, just "Implemented a naive version of memfuncs" is cleaner.

abhinav92003 (approved):

core/arch/riscv64/memfuncs.asm
ksco: Done.

PR #6071, opened 2023-05-19
i#3544 riscv64: Implemented dr_setjmp and dr_longjmp

shiptux:

- Implemented dr_setjmp
- Implemented dr_longjmp

Issue: #3544


core/arch/riscv64/riscv64.asm
```diff
@@ -208,16 +208,51 @@ ADDRTAKEN_LABEL(safe_read_asm_recover:)
  */
         DECLARE_EXPORTED_FUNC(dr_try_start)
 GLOBAL_LABEL(dr_try_start:)
-        addi      ARG1, ARG1, TRY_CXT_SETJMP_OFFS
+        addi     ARG1, ARG1, TRY_CXT_SETJMP_OFFS
         j        GLOBAL_REF(dr_setjmp)
         END_FUNC(dr_try_start)
 
-/*
+/* We save only callee-saved regites: SP, x8/fp, x9, x18-x27, f8-9, f18-27:
```
abhinav92003, line 215: Do you mean s/regites/regs/?

core/arch/riscv64/riscv64.asm
```diff
@@ -208,16 +208,51 @@ ADDRTAKEN_LABEL(safe_read_asm_recover:)
...
+        sd       x19, 8 (ARG1)
+        sd       x20, 16 (ARG1)
+        sd       x21, 24 (ARG1)
+        sd       x22, 32 (ARG1)
+        sd       x23, 40 (ARG1)
+        sd       x24, 48 (ARG1)
+        sd       x25, 56 (ARG1)
+        sd       x26, 64 (ARG1)
+        sd       x27, 72 (ARG1)
+        mv       ARG7, sp
```
abhinav92003, line 232: Curious: here we're using ARG7 as a temporary reg, right? Could we have used any of the many tx temporary regs provided by riscv?

core/arch/riscv64/riscv64.asm
```diff
@@ -208,16 +208,51 @@ ADDRTAKEN_LABEL(safe_read_asm_recover:)
...
+        sd       ARG7, 80 (ARG1)
+        sd       x8, 88 (ARG1)
```
abhinav92003, line 234: It'd be nicer if this list followed the same sequence as in mcxtx_api.h. That is, x2(sp), x8, x9 moved up top.

core/arch/riscv64/riscv64.asm
```diff
@@ -226,8 +261,36 @@ GLOBAL_LABEL(dr_setjmp:)
  */
         DECLARE_FUNC(dr_longjmp)
 GLOBAL_LABEL(dr_longjmp:)
-/* FIXME i#3544: Not implemented */
-        ret
+        sd       x18, 0 (ARG1)
```
abhinav92003, line 264: But longjmp should be loading the values from the given dr_jmp_buf into registers. See the equivalent version in core/arch/aarch64/aarch64.asm.

abhinav92003 (changes requested):

core/arch/riscv64/riscv64.asm
shiptux: Done.

core/arch/riscv64/riscv64.asm
shiptux: Done

core/arch/riscv64/riscv64.asm
shiptux: Done. Fixed https://github.com/DynamoRIO/dynamorio/pull/6071/commits/3859e2075255495463f4d09652e54771306f0a75

core/arch/riscv64/riscv64.asm
shiptux: Fix https://github.com/DynamoRIO/dynamorio/pull/6071/commits/3859e2075255495463f4d09652e54771306f0a75

core/arch/riscv64/riscv64.asm
shiptux: Fixed https://github.com/DynamoRIO/dynamorio/pull/6071/commits/3859e2075255495463f4d09652e54771306f0a75

core/arch/riscv64/riscv64.asm
shiptux: Fixed https://github.com/DynamoRIO/dynamorio/pull/6071/commits/3859e2075255495463f4d09652e54771306f0a75

core/arch/riscv64/riscv64.asm
```diff
@@ -226,8 +261,36 @@ GLOBAL_LABEL(dr_setjmp:)
...
+        fld      f21, 144 (ARG1)
+        fld      f22, 152 (ARG1)
+        fld      f23, 160 (ARG1)
+        fld      f24, 168 (ARG1)
+        fld      f25, 176 (ARG1)
+        fld      f26, 184 (ARG1)
+        fld      f27, 192 (ARG1)
+        li       t0, 0
+        snez     t0, ARG1
+        add      ARG1, ARG, t0
```
abhinav92003, line 292: Where is `ARG` defined? Was this supposed to be `ARG2`? If yes, then are we returning `val + 1` if `buf != null`? Why?

core/arch/riscv64/riscv64.asm
```diff
@@ -226,8 +261,36 @@ GLOBAL_LABEL(dr_setjmp:)
...
+        fld      f20, 136 (ARG1)
+        fld      f21, 144 (ARG1)
+        fld      f22, 152 (ARG1)
+        fld      f23, 160 (ARG1)
+        fld      f24, 168 (ARG1)
+        fld      f25, 176 (ARG1)
+        fld      f26, 184 (ARG1)
+        fld      f27, 192 (ARG1)
+        li       t0, 0
+        snez     t0, ARG1
```
abhinav92003, line 291: But if ARG1 is zero, then the above loads would fault. Not sure what this accomplishes.

abhinav92003 (changes requested):

core/arch/riscv64/riscv64.asm
shiptux: It should be ARG1. If ARG1 = 0, then t0 will get 0, and add ARG1, ARG1, t0 will also be 0, which means that ARG1 remains unchanged. If ARG1 is not 0, then t0 will get 1, and add ARG1, ARG1, t0 will increment ARG1 by 1.

core/arch/riscv64/riscv64.asm
shiptux: I have changed to a different way of implementation

core/arch/riscv64/riscv64.asm
```diff
@@ -226,8 +261,36 @@ GLOBAL_LABEL(dr_setjmp:)
...
+        fld      f19, 128 (ARG1)
+        fld      f20, 136 (ARG1)
+        fld      f21, 144 (ARG1)
+        fld      f22, 152 (ARG1)
+        fld      f23, 160 (ARG1)
+        fld      f24, 168 (ARG1)
+        fld      f25, 176 (ARG1)
+        fld      f26, 184 (ARG1)
+        fld      f27, 192 (ARG1)
+        beqz     ARG1, skip
```
abhinav92003, line 290: I still don't understand how this is useful. Is this to check whether ARG1 is a valid address (not null)? If yes, then shouldn't that check be at the top of this function, before we use ARG1 for the above loads? Otherwise please add a comment explaining this part.

core/arch/riscv64/riscv64.asm
```diff
@@ -226,8 +261,36 @@ GLOBAL_LABEL(dr_setjmp:)
...
+        addi     ARG1, ARG1, 1
+skip:
+        jalr     ra
```
abhinav92003, line 293: ARG2 is unused. Shouldn't that be put in the return value reg (x10/arg1)?

abhinav92003, line 293: This ra is the return address for this dr_longjmp function call. But a longjmp does not return to where it was called. It should return to where the corresponding setjmp was called. Shouldn't we be saving the return address in dr_setjmp and then return to that address here? See the x86 implementation in core/arch/x86/x86.asm.

abhinav92003 (changes requested):

core/arch/riscv64/riscv64.asm
abhinav92003: See reference for longjmp: https://linux.die.net/man/3/longjmp.

shiptux: Fixed https://github.com/DynamoRIO/dynamorio/pull/6071/commits/6779b4195a5960141158905e3623fd9b9d7868fe

shiptux: Fixed https://github.com/DynamoRIO/dynamorio/pull/6071/commits/6779b4195a5960141158905e3623fd9b9d7868fe

core/arch/riscv64/riscv64.asm
```diff
@@ -208,16 +208,52 @@ ADDRTAKEN_LABEL(safe_read_asm_recover:)
...
+ * a total of 26 reg_t (64-bit) slots. See definition of dr_jmp_buf_t.
+ *
  * int dr_setjmp(dr_jmp_buf_t *buf);
  */
         DECLARE_FUNC(dr_setjmp)
 GLOBAL_LABEL(dr_setjmp:)
-/* FIXME i#3544: Not implemented */
+        sd       ra, 0 (ARG1)
+        sd       t0, ARG_SZ (ARG1)
+        mv       t0, sp
```
abhinav92003, line 224: I think this `mv t0, sp` should be before the above `sd t0, ...`

abhinav92003 (approved):

core/arch/riscv64/riscv64.asm
shiptux: done

shiptux: Fixed https://github.com/DynamoRIO/dynamorio/pull/6071/commits/1451c1cbcbfbef45d68fa62cb0c32144f99f686c

PR #6077, opened 2023-05-22
i#5843 scheduler: Add nested speculation layers

derekbruening:
Changes speculation support to use distinct layers, allowing an inner speculation to roll back to the next-outermost interruption point.

Adds unit tests.

Issue: #5843

### DIFF
--- a/clients/drcachesim/scheduler/scheduler.cpp
+++ b/clients/drcachesim/scheduler/scheduler.cpp
@@ -1406,7 +1406,8 @@ scheduler_tmpl_t<RecordType, ReaderType>::next_record(output_ordinal_t output,
     }
     input = &inputs_[outputs_[output].cur_input];
     auto lock = std::unique_lock<std::mutex>(*input->lock);
-    if (outputs_[output].speculating) {
+    if (!outputs_[output].speculation_split.empty()) {
+        outputs_[output].prev_speculate_pc = outputs_[output].speculate_pc;
         error_string_ = outputs_[output].speculator.next_record(
             outputs_[output].speculate_pc, record);
         if (!error_string_.empty())
@@ -1547,11 +1548,21 @@ scheduler_tmpl_t<RecordType, ReaderType>::start_speculation(output_ordinal_t out
                                                             addr_t start_address,
                                                             bool queue_current_record)
 {
-    if (!outputs_[output].speculating && queue_current_record) {
-        inputs_[outputs_[output].cur_input].queue.push_back(outputs_[output].last_record);
+    auto &outinfo = outputs_[output];
+    if (outinfo.speculation_split.empty()) {
+        if (queue_current_record)
+            inputs_[outinfo.cur_input].queue.push_back(outinfo.last_record);
+        outinfo.speculation_split.push(0);
+    } else {
+        if (queue_current_record) {
+            // XXX: We'll re-call the speculator so we're assuming a repeatable response.
+            outinfo.speculation_split.push(outinfo.prev_speculate_pc);
+        } else
+            outinfo.speculation_split.push(outinfo.speculate_pc);
     }
-    outputs_[output].speculating = true;
-    outputs_[output].speculate_pc = start_address;
+    outinfo.speculate_pc = start_address;
+    VPRINT(this, 2, "start_speculation layer=%zu pc=0x%" PRIx64 "\n",
+           outinfo.speculation_split.size(), start_address);
     return sched_type_t::STATUS_OK;
 }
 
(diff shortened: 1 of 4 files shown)


clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -1547,21 +1548,38 @@ scheduler_tmpl_t<RecordType, ReaderType>::start_speculation(output_ordinal_t out
...
-    if (!outputs_[output].speculating && queue_current_record) {
-        inputs_[outputs_[output].cur_input].queue.push_back(outputs_[output].last_record);
+    auto &outinfo = outputs_[output];
+    if (outinfo.speculation_split.empty()) {
+        if (queue_current_record)
+            inputs_[outinfo.cur_input].queue.push_back(outinfo.last_record);
+        outinfo.speculation_split.push(0);
+    } else {
+        if (queue_current_record) {
+            // XXX: We'll re-call the speculator so we're assuming a repeatable response.
```
abhinav92003, line 1558: You mean we'll get the same response from the speculator when we re-start at prev_speculate_pc? I don't know how the speculator is implemented so not sure if it depends on any other state. But does it matter if the speculator returns a different pc after resumption at prev_speculate_pc?

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -1547,21 +1548,38 @@ scheduler_tmpl_t<RecordType, ReaderType>::start_speculation(output_ordinal_t out
...
+            outinfo.speculation_split.push(outinfo.prev_speculate_pc);
+        } else
+            outinfo.speculation_split.push(outinfo.speculate_pc);
     }
-    outputs_[output].speculating = true;
-    outputs_[output].speculate_pc = start_address;
+    outinfo.speculate_pc = start_address;
```
abhinav92003, line 1563: What should be prev_speculate_pc in this case?

clients/drcachesim/scheduler/scheduler.h
```diff
@@ -522,25 +523,23 @@ template <typename RecordType, typename ReaderType> class scheduler_tmpl_t {
...
-         * Stops speculative execution and resumes the regular stream of records from
-         * the point at which the most distant prior start_speculation() call without an
-         * intervening stop_speculation() call was made (either repeating the current
-         * record at that time, if "true" was passed for "queue_current_record" to
-         * start_speculation(), or continuing on the subsequent record if "false" was
-         * passed).  Returns #dynamorio::drmemtrace::scheduler_tmpl_t::STATUS_INVALID if
-         * there was no prior start_speculation() call or if stop_speculation() was
-         * already called since the last start.
+         * Stops speculative execution, resuming execution at the
+         * stream of records from the point at which the prior start_speculation()
```
abhinav92003, line 537: s/prior/prior matching/

clients/drcachesim/scheduler/scheduler.h
```diff
@@ -891,9 +890,10 @@ template <typename RecordType, typename ReaderType> class scheduler_tmpl_t {
         std::vector<input_ordinal_t> input_indices;
         int input_indices_index = 0;
         // Speculation support.
-        bool speculating = false;
+        std::stack<addr_t> speculation_split; // PC of resumption point.
```
abhinav92003, line 893: Why the "_split" suffix? I'd have expected speculation_stack.

clients/drcachesim/scheduler/scheduler.h
```diff
@@ -891,9 +890,10 @@ template <typename RecordType, typename ReaderType> class scheduler_tmpl_t {
...
         speculator_tmpl_t<RecordType> speculator;
         addr_t speculate_pc = 0;
+        addr_t prev_speculate_pc = 0;
```
abhinav92003, line 896: Add comment what this represents.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: If queuing of the already-seen instruction was requested at start_speculation, the speculator has to return that same instruction in all details, not just the next pc.  Maybe it should be saved -- the speculator could always remember every start-request instruction, or maybe the scheduler should add a saved record to this stack.

derekbruening: For now I will put a TODO both here and in the speculator.  This won't be an issue until we do more than return a nop.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -1547,21 +1548,38 @@ scheduler_tmpl_t<RecordType, ReaderType>::start_speculation(output_ordinal_t out
                                                             addr_t start_address,
                                                             bool queue_current_record)
 {
-    if (!outputs_[output].speculating && queue_current_record) {
-        inputs_[outputs_[output].cur_input].queue.push_back(outputs_[output].last_record);
+    auto &outinfo = outputs_[output];
+    if (outinfo.speculation_split.empty()) {
+        if (queue_current_record)
+            inputs_[outinfo.cur_input].queue.push_back(outinfo.last_record);
+        outinfo.speculation_split.push(0);
```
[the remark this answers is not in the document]
derekbruening: Done.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Done.

clients/drcachesim/scheduler/scheduler.h
derekbruening: Done.

clients/drcachesim/scheduler/scheduler.h
derekbruening: Because the stack is recording the split or divergence in path from the regular or prior-nested context: that's what is stored in the stack.  Maybe that's not clear.  I'm ok renaming to `speculation_stack`.  Done.

clients/drcachesim/scheduler/scheduler.h
derekbruening: Done.

PR #6079, opened 2023-05-23
Reuse Distance: Add data access histogram to collected stats.

brettcoon:
Adds a separate histogram for non-instruction access histograms. This primary histogram continues to include all accesses.  The final output reports instruction and data reference counts, and print_histogram() includes both total and data-only accesses.


clients/drcachesim/tools/reuse_distance.h
```diff
@@ -101,10 +104,22 @@ class reuse_distance_t : public analysis_tool_t {
...
+        // These are our reuse distance histograms: one for all accesses and one
+        // only for data references.  An instruction histogram can be computed by
+        // subtracting data references from the full histogram.  The final histogram
+        // statistics need a full histogram to sort, and in most traces the majority of
+        // accesss are instruction references, so the histograms are split this way to
+        // provide the full histogram we need with the smallest secondary histogram.
+        // Furthermore, during analysis each reference is added to only one histogram
+        // to minimize the performance impact of dual histogram collection, with the data
+        // references added into the primary histogram during final result aggregation.
+        // This means dist_map is effectively instruction-only until aggregation.
```
derekbruening, line 116: Hmm, this has potential to cause confusion: I guess we don't want to pay the space for two separate histograms, dist_map_instr and dist_map_combined?

clients/drcachesim/tools/reuse_distance.h
```diff
@@ -101,10 +104,22 @@ class reuse_distance_t : public analysis_tool_t {
...
+        distance_histogram_t dist_map;
+        distance_histogram_t dist_map_data;
+        bool dist_map_needs_merge = true;
```
derekbruening, line 119: Rename to `dist_map_is_instr_only`?

derekbruening (approved):

clients/drcachesim/tests/reuse_distance_test.cpp
```diff
@@ -358,36 +377,128 @@ print_histogram_mult_1p2_test()
...
+    knobs.verbose = 0;
+    reuse_distance_test_t reuse_distance(knobs);
+
+    // Create address generator with a predictable access pattern.
+    address_generator_t agen(TEST_ADDRESS, LINE_SIZE);
+
+    int instruction_hits = 0;
+    int data_hits = 0;
+    // Simple function to decide if a given access should be TRACE_TYPE_INSTR.
+    auto UseInstrType = [&](int distance) {
```
derekbruening, line 448: style: no CamelCase: `use_instr_type`

clients/drcachesim/tests/reuse_distance.templatex
```diff
@@ -1,8 +1,10 @@
 Hello, world!
 ---- <application exited with code 0> ----
 Reuse distance tool aggregated results:
-Total accesses: [0-9]+
-Unique accesses: [0-9]+
+Total accesses: [1-9][0-9]+
+Instruction accesses: [1-9][0-9]+
+Data accesses: [1-9][0-9]+
+Unique accesses: [1-9][0-9]+
```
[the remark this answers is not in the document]
brettcoon: That statistic (and most others as well) is only available for the combined access stream.  So while it might be interesting to separate this, that would require a much more intrusive and potentially performance impacting change.  The split done in this PR is only for the distance histogram, with no changes or duplication to the expensive cacheline tracking logic.

clients/drcachesim/tests/reuse_distance_test.cpp
```diff
@@ -326,24 +336,33 @@ print_histogram_mult_1p0_test()
...
+        // Put 1/3 of these in the data map as well.
+        if (i % 3 == 0) {
+            distance_map_data[i] += sorted.back().second;
+        }
     }
 
     std::stringstream output_string;
-    reuse_distance.print_histogram(output_string, count, sorted);
+    reuse_distance.print_histogram(output_string, count, sorted, distance_map_data);
+    if (knobs.verbose > 1) {
```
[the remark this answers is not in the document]
brettcoon: Done.

clients/drcachesim/tests/reuse_distance_test.cpp
```diff
@@ -358,36 +377,128 @@ print_histogram_mult_1p2_test()
...
+    }
+    if (TEST_VERBOSE(1)) {
+        reuse_distance.print_results();
+    }
+
+    assert(reuse_distance.get_shard_map().size() == 1);
+
+    auto *shard = reuse_distance.get_aggregated_results();
+
+    assert(shard->data_refs > data_hits);
```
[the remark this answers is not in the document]
brettcoon: Because it's comparing the number of data references to the number of **hits**, and by construction not every reference is a hit (to get a distance of X, the helper function inserts X unique references between two references to the same address).

clients/drcachesim/tools/reuse_distance.h
brettcoon: Yes, I debated this, and agree it's a bit confusing.  I started going down the path of having instruction-only and data-only histograms.  But during the final summary and aggregation stages there are multiple uses of the unified histogram, so it either needed to be created and kept around as a third histogram, or else the instruction-only histogram needed to be repurposed.  I guess my solution is the latter, except I named the dual-use histogram by its final form, rather then the as-collected form.  I'll keep it this way unless you feel strongly.

clients/drcachesim/tools/reuse_distance.h
brettcoon: Done.

clients/drcachesim/tests/reuse_distance_test.cpp
brettcoon: Doh!  Fixed.

PR #6082, opened 2023-05-25
i#6020 interval analysis: add TID and positive interval_id

abhinav92003:
Adds TID to the interval_state_snapshot_t base struct, so that tools do not need to track it themselves if they need it. The TID is stored as "shard_id" which in future may contain a different id if we define shards differently.

Also changes interval_id to always be a positive integer, which makes some things simpler downstream.

Issue: #6020

### DIFF
--- a/clients/drcachesim/tests/trace_interval_analysis_unit_tests.cpp
+++ b/clients/drcachesim/tests/trace_interval_analysis_unit_tests.cpp
@@ -442,7 +442,9 @@ public:
                 auto recorded_snapshot =
                     dynamic_cast<const recorded_snapshot_t *const>(snapshot);
                 if (recorded_snapshot->tool_shard_id != recorded_snapshot->shard_id) {
-                    std::cerr << "shard_id stored by tool and framework mismatch";
+                    std::cerr << "shard_id stored by tool ("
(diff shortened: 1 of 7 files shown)


abhinav92003:
AArch64 `samples_proj` failure is #6083.

clients/drcachesim/tests/trace_interval_analysis_unit_tests.cpp
```diff
@@ -333,6 +341,9 @@ class test_analysis_tool_t : public analysis_tool_t {
         // this contains a list of size equal to the count of shard interval snapshots
         // that were combined to create this snapshot.
         std::vector<interval_end_point_t> component_intervals;
+        // Stores the shard_id recorded by the test tool. Compared with the shard_id
+        // stored by the tool in the base struct.
```
derekbruening, line 345: the framework

derekbruening (approved):

clients/drcachesim/tests/trace_interval_analysis_unit_tests.cpp
abhinav92003: Ah. Done.

PR #6085, opened 2023-05-26
i#5843 scheduler: Add core binding support

derekbruening:
Adds support for inputs bound to a subset of the available cores. This is implemented by repeatedly popping the priority queue of ready inputs until one able to run on the core in question is found; the rest are re-added to the queue.

Adds a test.

Issue: #5843

### DIFF
--- a/clients/drcachesim/scheduler/scheduler.cpp
+++ b/clients/drcachesim/scheduler/scheduler.cpp
@@ -1237,14 +1237,15 @@ scheduler_tmpl_t<RecordType, ReaderType>::add_to_ready_queue(input_info_t *input
 
 template <typename RecordType, typename ReaderType>
 typename scheduler_tmpl_t<RecordType, ReaderType>::input_info_t *
-scheduler_tmpl_t<RecordType, ReaderType>::pop_from_ready_queue(output_ordinal_t output)
+scheduler_tmpl_t<RecordType, ReaderType>::pop_from_ready_queue(
+    output_ordinal_t for_output)
 {
     std::set<input_info_t *> skipped;
     input_info_t *res = nullptr;
     do {
         res = ready_priority_.top();
         ready_priority_.pop();
-        if (res->binding.empty() || res->binding.find(output) != res->binding.end())
+        if (res->binding.empty() || res->binding.find(for_output) != res->binding.end())
             break;
         // We keep searching for a suitable input.
         skipped.insert(res);
@@ -1269,6 +1270,9 @@ typename scheduler_tmpl_t<RecordType, ReaderType>::stream_status_t
 scheduler_tmpl_t<RecordType, ReaderType>::set_cur_input(output_ordinal_t output,
                                                         input_ordinal_t input)
 {
+    // XXX i#5843: Merge tracking of current inputs with ready_queue_ to better manage
+    // the possible 3 states of each input (a live cur_input for an output stream, in
+    // the ready_queue_, or at EOF).
     assert(output >= 0 && output < static_cast<output_ordinal_t>(outputs_.size()));
     // 'input' might be INVALID_INPUT_ORDINAL.
     assert(input < static_cast<input_ordinal_t>(inputs_.size()));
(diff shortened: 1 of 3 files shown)


clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -987,6 +987,74 @@ test_synthetic_with_priorities()
...
+        scheduler_t::STATUS_SUCCESS)
+        assert(false);
+    std::vector<std::string> sched_as_string =
+        run_lockstep_simulation(scheduler, NUM_OUTPUTS, TID_BASE);
+    for (int i = 0; i < NUM_OUTPUTS; i++) {
+        std::cerr << "cpu #" << i << " schedule: " << sched_as_string[i] << "\n";
+    }
+    // We have {A,B,C} on {2,4}, {D,E,F} on {0,1}, and {G,H,I} on {1,2,3}:
+    assert(sched_as_string[0] == "DDDFFFDDDFFFDDDFFF");
+    assert(sched_as_string[1] == "EEEHHHEEEIIIEEE");
```
abhinav92003, line 1052: The first input scheduled on core 1: I see it is E, why not G? G is eligible and also has a lower workload-local timestamp than E. Then on core 3, you'll have H.

abhinav92003 (approved):

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -1240,14 +1237,30 @@ scheduler_tmpl_t<RecordType, ReaderType>::add_to_ready_queue(input_info_t *input
 
 template <typename RecordType, typename ReaderType>
 typename scheduler_tmpl_t<RecordType, ReaderType>::input_info_t *
-scheduler_tmpl_t<RecordType, ReaderType>::pop_from_ready_queue()
+scheduler_tmpl_t<RecordType, ReaderType>::pop_from_ready_queue(output_ordinal_t output)
```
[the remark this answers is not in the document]
derekbruening: Done.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -1422,8 +1435,9 @@ scheduler_tmpl_t<RecordType, ReaderType>::pick_next_input(output_ordinal_t outpu
                     // shouldn't switch.  The queue preserves FIFO for same-priority
                     // cases so we will switch if someone of equal priority is waiting.
                     set_cur_input(output, INVALID_INPUT_ORDINAL);
```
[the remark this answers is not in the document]
derekbruening: A 3rd state for an input is to be at EOF.  I added a XXX comment at the top of set_cur_input to possibly merge its logic into some new tracking object that combines with the queue.

clients/drcachesim/scheduler/scheduler.h
```diff
@@ -1103,7 +1111,7 @@ template <typename RecordType, typename ReaderType> class scheduler_tmpl_t {
 
     // sched_lock_ must be held by the caller.
     input_info_t *
-    pop_from_ready_queue();
+    pop_from_ready_queue(output_ordinal_t output);
```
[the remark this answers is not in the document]
derekbruening: Done.

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: These actually all have the same workload-local timestamps (this doesn't use the weighted timestamps that the other test has) which makes it easier to reason about the results here.

abhinav92003: But if it did use weighted timestamps would we see a different schedule (G instead of E)?

derekbruening: The timestamps from the other test are inverted to G would have later timestamps than E being 1st vs 2nd in the triplet.  I tried it and actually it found a bug in pop_from_ready_queue() where if the q is empty it thinks there's nothing, skipping the last one: fixed now.  We then end up with:
```
cpu #0 schedule: FFFFFFFFFEEEEEEEEE
cpu #1 schedule: IIIIIIIIIDDDDDDDDD
cpu #2 schedule: CCCCCCCCCAAAAAAAAA
cpu #3 schedule: HHHHHHHHHGGGGGGGGG
cpu #4 schedule: BBBBBBBBB
```
It is less interesting as we get no interleaving.

derekbruening: I'm adding this as a separate test to include that final entry case.

Issue #6043, opened 2023-05-12
DR fails to build with libipt


abhinav92003:
Reported by a user at https://groups.google.com/g/dynamorio-users/c/rFUALcXdskE

DR fails to lihnk libipt-sb.so:

```
$ git clone https://github.com/DynamoRIO/dynamorio.git
$ cd dynamorio/
$ git submodule init
$ git submodule update

$ mkdir build/
$ cd build/
$ cmake ../dynamorio/
$ make -j

[ 26%] Linking C shared library ../../../../third_party/lib64/release/libipt-sb.so
/usr/bin/ld: ../../../../third_party/lib64/release/libpevent.a(pevent.c.o): relocation R_X86_64_PC32 against symbol `pev_time_to_tsc' can                                                                          not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Bad value
collect2: error: ld returned 1 exit status
clients/drcachesim/drpt2trace/libipt/CMakeFiles/ipt-sb.dir/build.make:174: recipe for target 'third_party/lib64/release/libipt-sb.so' fail                                                                         ed
make[2]: *** [third_party/lib64/release/libipt-sb.so] Error 1
CMakeFiles/Makefile2:2179: recipe for target 'clients/drcachesim/drpt2trace/libipt/CMakeFiles/ipt-sb.dir/all' failed
make[1]: *** [clients/drcachesim/drpt2trace/libipt/CMakeFiles/ipt-sb.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

I could not reproduce the issue on my machine.

prasun3:
It fails on these two Ubuntu 18 systems

```
VERSION="18.04.6 LTS (Bionic Beaver)"
cmake version 3.10.2
gcc (Ubuntu 9.4.0-1ubuntu1~18.04) 9.4.0
GNU ld (GNU Binutils for Ubuntu) 2.30
GNU C Library (Ubuntu GLIBC 2.27-3ubuntu1.6) stable release version 2.27.
---
VERSION="18.04.1 LTS (Bionic Beaver)"
cmake version 3.10.2
gcc (Ubuntu 10.1.0-2ubuntu1~18.04) 10.1.0
GNU ld (GNU Binutils for Ubuntu) 2.30
GNU C Library (Ubuntu GLIBC 2.27-3ubuntu1) stable release version 2.27.
```

It builds successfully on this Ubuntu 20 system
```
[...]

PR #6086, opened 2023-05-26
Document list of supported environments

abhinav92003:
Adds a link to the list of officially supported environments to the page on building from source. As requested on #6043.

### DIFF
--- a/api/docs/building.dox
+++ b/api/docs/building.dox
@@ -1,5 +1,5 @@
 /* ******************************************************************************
- * Copyright (c) 2010-2022 Google, Inc.  All rights reserved.
+ * Copyright (c) 2010-2023 Google, Inc.  All rights reserved.
  * ******************************************************************************/
 
 /*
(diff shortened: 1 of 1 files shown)


api/docs/building.dox
```diff
@@ -38,6 +38,15 @@
 
 If you don't need to modify DR and just need a recent build, you can download [weekly package builds](@ref page_weekly_builds).
 
+The official list of environments for which we have complete or in-progress support are the ones on our [Github
+CI workflows page](https://github.com/DynamoRIO/dynamorio/tree/master/.github/workflows), These
```
derekbruening, line 42: s/,/./

api/docs/building.dox
```diff
@@ -38,6 +38,15 @@
...
+workflows run our test suite before the submission of each master merge, and also a longer test suite after each
+master merge. This helps in ensuring that DynamoRIO continues to build and run properly on these environments.
+For other environments, contributions from users are greatly appreciated. For help with any issue building or
+running on any environment, feel free to start a thread on our
+[discussion list](http://groups.google.com/group/DynamoRIO-Users) or report a bug on our
```
derekbruening, line 47: I would remove the issue tracker link if it's "For help".

api/docs/building.dox
```diff
@@ -38,6 +38,15 @@
 
 If you don't need to modify DR and just need a recent build, you can download [weekly package builds](@ref page_weekly_builds).
 
+The official list of environments for which we have complete or in-progress support are the ones on our [Github
```
derekbruening, line 41: This needs to be reconciled with `sec_limit_platforms` which has lists of versions of each OS which are supposedly supported; the rendered contents are at https://dynamorio.org/release_notes.html#sec_limit_platforms.

A link to that section should be included here, and that section itself should likely be updated, and some wording explaining the level of support for CI workflows vs "should work" on the limit list.

PR #6087, opened 2023-05-27
i#5979 decode: Rename ud2a/ud2b to ud2/ud1 and decode ud1 operands

### LINKED ISSUE #5979: Incorrect decoding of UD2
This byte sequence is a single UD1 instruction:
```
echo -en "\x67\x0f\xb9\x40\x16" > /tmp/bin | objdump -mi386 -Mx86-64 -b binary -D /tmp/bin
   0:	67 0f b9 40 16       	ud1    0x16(%eax),%eax
```

```
echo -en "0x67 0x0f 0xb9 0x40 0x16" | llvm-mc --disassemble -
	ud1l	22(%eax), %eax
```

but drdecode decodes it as 3-byte instruction. This breaks decoding of instruction stream.
I see lots of these instructions in real binaries compiled with clang, it seems to use it as padding at the end of functions or something.

I am using commit 45c9973e363254d534b819d3cc508601f534bb71.

khuey:
The opcodes DR calls ud2a and ud2b are more commonly known (and documented by Intel) as ud2 and ud1, respectively. Rename them, preserving the old operand constants and instruction creation macros through preprocessor defines. Additionally, 
ud1 actually takes two operands. Fix the decode table for them.

Fixes #5979


suite/tests/api/ir_x86.c
```diff
@@ -2765,6 +2765,25 @@ test_extra_leading_prefixes(void *dc)
 #endif
 }
 
+static void
+test_ud1_operands(void *dc)
```
derekbruening, line 2769: It seems confusing to call it ud1 here yet ud2b everywhere else (including in the disasm strings inside this function): add a comment to the decode_table entry mentioning it is also called "ud1"?

Maybe also rename this function to use DR's name?

derekbruening (approved):

suite/tests/api/ir_x86.c
khuey: What are the backwards-compat limitations here on renaming ud2b to ud1 and ud2a to ud2? Do we need to preserve #defines for the old OP_xxx, instr_create_xxx, etc?

derekbruening: We should preserve the OP_ numbers (no adding/removing in the middle) and we should preserve the old OP_ud2b.  What I would do is rename the OP_ud2b in the enum (and table) to OP_ud1 and then add OP_ud2a as an alias define equal to OP_ud1.  I would add a release note to release.dox as a minor compat change that the instr_create and decoding of OP_ud2b is now different and would not preserve the exact semantics of the instr_create.

khuey: Alright, how about this?

api/docs/release.dox
```diff
@@ -179,6 +179,10 @@ changes:
    which is required for producing whole-trace results in the parallel mode of analyzer
    operation, and also release_interval_snapshot() which is used to release the
    analyzer framework's claim to the interval snapshot objects.
+ - ud2a and ud2b have been renamed to ud2 and ud1, respectively. The old constants
+   OP_ud2a and OP_ud2b, as well as the INSTR_CREATE_ud2a()/INSTR_CREATE_ud2b() macros,
```
derekbruening, line 183: Add `#` to get doxygen links to the OP_ names: `#OP_ud2a`, etc.  Maybe should mention `#OP_ud1` to get a link there too.

clients/drcachesim/tests/signal_invariants.c
```diff
@@ -194,7 +194,7 @@ GLOBAL_LABEL(FUNCNAME:)
         prefetcht2 [3]
         nop
         nop
-        ud2a
+        ud2
```
derekbruening, line 197: (There was some general community rename at some point as DR got its names ud2a + ud2b from somewhere, and even today assemblers are happy with `ud2a`: not sure assemblers take `ud2b`.)

core/ir/x86/opcode_api.h
```diff
@@ -1692,4 +1692,8 @@ enum {
 #define OP_icebp OP_int1
 #define OP_setalc OP_salc
 
+/* renamed opcodes */
```
derekbruening, line 1695: style nit: Prefer capitalized + punctuated for new comments esp whole-line comments.

suite/tests/api/ir_x86.c
derekbruening: Yes, looks good.

derekbruening (approved):

suite/tests/api/ir_x86.c
derekbruening: Don't forget to update the PR description (which becomes the commit msg) to mention the renames.

api/docs/release.dox
derekbruening: That generally means you have an error and either omitted a Doxygen comment for the target or didn't pass that file to Doxygen -- please double-check.

core/ir/x86/opcode_api.h
```diff
@@ -1692,4 +1692,8 @@ enum {
 #define OP_icebp OP_int1
 #define OP_setalc OP_salc
 
+/* Renamed opcodes. */
+#define OP_ud2a OP_ud2
```
derekbruening, line 1696: Indeed, you have no Doxygen comment here -- this is why Doxygen couldn't add the link.

PR #6088, opened 2023-05-27
i#3544 RV64: Added machine type support for ELF hdr verification

ksco:
This patch added ELF header checks for RISC-V.


Reese0710:
I don't know if you ran into this problem when I tried to compile the latest version of the dynamorio like this
image
then it's failed
image
I've tried the gcc 7.3.0 & 8.5.0
my OS is centos 8.0 x86, cpu: Intel(R) Xeon(R) Gold 6278 CPU @ 2.60GHz

ksco:
RISC-V is kinda new, upgrading the compiler to a newer version should fix this. I don't know if we're still supporting old compilers, can you open a new issue so we can discuss it there?

PR #6089, opened 2023-05-27
i#3544 RV64: Fixed atomic functions and macros

ksco:
This patch fixed some atomic functions and macros of the RV64 port.

### DIFF
--- a/core/arch/atomic_exports.h
+++ b/core/arch/atomic_exports.h
@@ -935,11 +935,11 @@ atomic_compare_exchange_int(volatile int *var, int compare, int exchange)
     ASSERT(ALIGNED(var, 4));
     /* This code is based on GCC's atomic_compare_exchange_strong().
      * The fence ensures no re-ordering beyond the start of the lr/sc loop.
-     * The lr/sc use acquire and release ordering to ensure that the full loop
-     * has sequentially consistent memory order semantics.
+     * Provides acquire/release semantics for LR/SC sequences by setting the
+     * aq and rl bits on lr and sc respectively.
      */
     __asm__ __volatile__("fence   iorw,ow\n\t"
-                         "1: lr.w.aqrl  t0, (%1)\n\t"
+                         "1: lr.w.aq    t0, (%1)\n\t"
                          "   bne        t0, %2, 2f\n\t"
                          "   sc.w.rl    %0, %3, (%1)\n\t"
                          "   bnez       %0, 1b\n\t"
@@ -960,11 +960,11 @@ atomic_compare_exchange_int64(volatile int64 *var, int64 compare, int64 exchange
     ASSERT(ALIGNED(var, 8));
     /* This code is based on GCC's atomic_compare_exchange_strong().
      * The fence ensures no re-ordering beyond the start of the lr/sc loop.
-     * The lr/sc use acquire ordering to ensure that the full loop has
-     * sequentially consistent memory order semantics.
+     * Provides acquire/release semantics for LR/SC sequences by setting the
+     * aq and rl bits on lr and sc respectively.
      */
     __asm__ __volatile__("fence   iorw,ow\n\t"
-                         "1: lr.d.aqrl  t0, (%1)\n\t"
+                         "1: lr.d.aq    t0, (%1)\n\t"
                          "   bne        t0, %2, 2f\n\t"
                          "   sc.d.rl    %0, %3, (%1)\n\t"
                          "   bnez       %0, 1b\n\t"
--- a/core/arch/atomic_exports.h
+++ b/core/arch/atomic_exports.h
@@ -730,14 +730,14 @@ atomic_dec_becomes_zero(volatile int *var)
 #        define SET_IF_NOT_ZERO(flag) SET_FLAG(ne, flag)
 #        define SET_IF_NOT_LESS(flag) SET_FLAG(ge, flag)
 #    elif defined(DR_HOST_RISCV64)
-#        define ATOMIC_1BYTE_READ(addr_src, addr_res)               \
-            do {                                                    \
-                /* Fence for acquire semantics. */                  \
-                __asm__ __volatile__("lbu %0, 0(%1)\n\t"            \
-                                     "fence iorw,iorw"              \
-                                     :                              \
-                                     : "r"(addr_src), "r"(addr_res) \
-                                     : "memory");                   \
+#        define ATOMIC_1BYTE_READ(addr_src, res)         \
+            do {                                         \
+                /* Fence for acquire semantics. */       \
+                __asm__ __volatile__("lbu %0, 0(%1)\n\t" \
+                                     "fence iorw,iorw"   \
+                                     : "=r"(res)         \
+                                     : "r"(addr_src)     \
+                                     : "memory");        \
             } while (0)
 #        define ATOMIC_1BYTE_WRITE(target, value, hot_patch)   \
             do {                                               \
@@ -745,9 +745,9 @@ atomic_dec_becomes_zero(volatile int *var)
                 ASSERT_CURIOSITY(!hot_patch);                  \
                 /* Fence for release semantics. */             \
                 __asm__ __volatile__("fence iorw,iorw\n\t"     \
-                                     "sb %0, 0(%1)\n\t"        \
+                                     "sb %1, 0(%0)\n\t"        \
                                      :                         \
-                                     : "r"(value), "r"(target) \
+                                     : "r"(target), "r"(value) \
                                      : "memory");              \
             } while (0)
 #        define ATOMIC_4BYTE_ALIGNED_WRITE(target, value, hot_patch) \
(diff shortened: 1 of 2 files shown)


core/arch/atomic_exports.h
```diff
@@ -936,11 +936,12 @@ atomic_compare_exchange_int(volatile int *var, int compare, int exchange)
...
-                         "   bne     t0, %2, 2f\n\t"
-                         "   sc.w.aq %0, %3, (%1)\n\t"
-                         "   bnez    %0, 1b\n\t"
-                         "2:"
+                         "1: lr.w.aqrl  t0, (%1)\n\t"
+                         "   bne        t0, %2, 2f\n\t"
+                         "   sc.w.rl    %0, %3, (%1)\n\t"
+                         "   bnez       %0, 1b\n\t"
+                         "2:\n\t"
+                         "   subw       %0, t0, %2\n\t"
```
abhinav92003, line 944: Why is this `subw` needed? Without it the logic seems to work fine:

* if `[var] != compare`, then `res` remains 1, so we return false below, as expected.
* if `[var] == compare`, then `res` will be zero when the `lr-sc` loop completes successfully, so we return true below as expected.

This `subw` seems to only redundantly set `res`.

Am I missing something?

core/arch/atomic_exports.h
```diff
@@ -936,11 +936,12 @@ atomic_compare_exchange_int(volatile int *var, int compare, int exchange)
      * sequentially consistent memory order semantics.
      */
     __asm__ __volatile__("fence   iorw,ow\n\t"
-                         "1: lr.w.aq t0, (%1)\n\t"
-                         "   bne     t0, %2, 2f\n\t"
-                         "   sc.w.aq %0, %3, (%1)\n\t"
-                         "   bnez    %0, 1b\n\t"
-                         "2:"
+                         "1: lr.w.aqrl  t0, (%1)\n\t"
```
abhinav92003, line 939: Maybe add a comment on these suffixes  to lr and sc. What do they mean and why are they needed?

abhinav92003 (changes requested):

core/arch/atomic_exports.h
```diff
@@ -730,24 +730,24 @@ atomic_dec_becomes_zero(volatile int *var)
 #        define SET_IF_NOT_ZERO(flag) SET_FLAG(ne, flag)
```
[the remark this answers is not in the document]
ksco: Done

core/arch/atomic_exports.h
```diff
@@ -730,24 +730,24 @@ atomic_dec_becomes_zero(volatile int *var)
...
 #    elif defined(DR_HOST_RISCV64)
-#        define ATOMIC_1BYTE_READ(addr_src, addr_res)               \
-            do {                                                    \
-                /* Fence for acquire semantics. */                  \
-                __asm__ __volatile__("lbu %0, 0(%1)\n\t"            \
-                                     "fence iorw,iorw"              \
-                                     :                              \
-                                     : "r"(addr_src), "r"(addr_res) \
-                                     : "memory");                   \
+#        define ATOMIC_1BYTE_READ(addr_src, res)         \
```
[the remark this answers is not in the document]
ksco: RISC-V's load-reserved (LR) and store-conditional (SC) instructions are tricky, they cannot be used with normal memory access instructions. E.g. it's invalid to do something like this:

```
LR.W ...
LW   ...
SC.W ...
```

IIUC, something like this is totally fine in ARM.

core/arch/atomic_exports.h
```diff
@@ -730,24 +730,24 @@ atomic_dec_becomes_zero(volatile int *var)
...
+            do {                                         \
+                /* Fence for acquire semantics. */       \
+                __asm__ __volatile__("lbu %0, 0(%1)\n\t" \
+                                     "fence iorw,iorw"   \
+                                     : "=r"(res)         \
```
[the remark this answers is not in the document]
ksco: In `__asm__` statement, the destination operand seems to have to be after the first colon.

core/arch/atomic_exports.h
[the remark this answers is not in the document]
ksco: From riscv-spec-20191213.pdf 8.3 Eventual Success of Store-Conditional Instructions

> An LR/SC sequence begins with an LR instruction and ends with an SC instruction. The dynamic code executed between the LR and SC instructions can only contain instructions from the base “I” instruction set, excluding loads, stores, backward jumps, taken backward branches, JALR, FENCE, FENCE.I, and SYSTEM instructions. If the “C” extension is supported, then compressed forms of the aforementioned “I” instructions are also permitted.

core/arch/atomic_exports.h
ksco: Indeed, the `subw` is redundant. I forgot we initialized res as 1 already, the `subw` is only needed when res is not initialized.

core/arch/atomic_exports.h
```diff
@@ -764,26 +764,26 @@ atomic_dec_becomes_zero(volatile int *var)
...
-                ASSERT(ALIGNED(addr_res, 4));                       \
-                /* Fence for acquire semantics. */                  \
-                __asm__ __volatile__("lw %0, 0(%1)\n\t"             \
-                                     "fence iorw,iorw"              \
-                                     :                              \
-                                     : "r"(addr_src), "r"(addr_res) \
-                                     : "memory");                   \
+#        define ATOMIC_4BYTE_ALIGNED_READ(addr_src, res)          \
+            do {                                                  \
+                /* Page 25 of Volume I: RISC-V Unprivileged ISA   \
```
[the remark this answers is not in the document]
ksco: The same kind of comment is already there.

core/arch/atomic_exports.h
ksco: It's basically what's described in the existing comment. But like what is said in spec[1], an `sc.w.aq` makes no sense. I changed the comment slightly in https://github.com/DynamoRIO/dynamorio/pull/6089/commits/3be7fd4d7b709206f05d900faa031773afc68cc5 to reflect this modification.

> Software should not set the rl bit on an LR instruction unless the aq bit is also set, nor should software set the aq bit on an SC instruction unless the rl bit is also set. LR.rl and SC.aq instructions are not guaranteed to provide any stronger ordering than those with both bits clear, but may result in lower performance.

core/arch/atomic_exports.h
ksco: Fixed in https://github.com/DynamoRIO/dynamorio/pull/6089/commits/3be7fd4d7b709206f05d900faa031773afc68cc5

abhinav92003: We should also fix it in atomic_compare_exchange_int64.

core/arch/atomic_exports.h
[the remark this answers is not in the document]
abhinav92003: The destination of these ATOMIC_xBYTE_READ macors would be a local memory location and the source would be a shared memory location. So only the read part needs to be "atomic" (that is, it should follow acquire-release semantics and (for multiple bytes) be read together). It doesn't make sense to write "atomically" to a local location.

So we can probably read into a temporary reg (like we're doing now), and then have a separate regular write to memory at addr_res.

@derekbruening to verify my understanding is correct.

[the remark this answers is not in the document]
derekbruening: Yes, writing the result for the caller is a separate, local step.  No need to add the complexity of changing the signature just for one arch.

"Atomic" for DR means untorn and acquire-release semantics.
Generally these macros should do the same thing the compiler would for `res = std::atomic<X>::load(std::memory_order_acquire)` (or `store` or `fetch_add` with `std::memory_order_release`).  
I would suggest directly observing what a compiler does for each of these to avoid making a mistake trying to hand-construct a sequence which can be error-prone.  Xref #4264 on using the compiler in the first place now that modern compilers have support -- but that would change the types and we'd need every compiler on board.

core/arch/atomic_exports.h
ksco: Fixed in https://github.com/DynamoRIO/dynamorio/pull/6089/commits/0cdef8be9747ba3903a0b771cab6d841964cd6e5.

abhinav92003:
Please request re-review from the UI when ready.

abhinav92003 (approved):

core/arch/atomic_exports.h
```diff
@@ -733,7 +733,8 @@ atomic_dec_becomes_zero(volatile int *var)
 #        define ATOMIC_1BYTE_READ(addr_src, addr_res)               \
```
derekbruening, line 733: Please update the PR description: it seems to talk about memory to memory atomic reads which is not done on any architecture in these atomic functions.

core/arch/atomic_exports.h
```diff
@@ -932,15 +935,15 @@ atomic_compare_exchange_int(volatile int *var, int compare, int exchange)
     ASSERT(ALIGNED(var, 4));
     /* This code is based on GCC's atomic_compare_exchange_strong().
      * The fence ensures no re-ordering beyond the start of the lr/sc loop.
-     * The lr/sc use acquire ordering to ensure that the full loop has
-     * sequentially consistent memory order semantics.
+     * The lr/sc use acquire and release ordering to ensure that the full loop
+     * has sequentially consistent memory order semantics.
```
derekbruening, line 939: We only need acquire/release: we don't need sequential consistency.

core/arch/atomic_exports.h
```diff
@@ -961,11 +964,11 @@ atomic_compare_exchange_int64(volatile int64 *var, int64 compare, int64 exchange
      * sequentially consistent memory order semantics.
```
derekbruening, line 964: We only need acquire/release: we don't need sequential consistency.

core/arch/atomic_exports.h
ksco: Done.

core/arch/atomic_exports.h
ksco: Fixed in https://github.com/DynamoRIO/dynamorio/pull/6089/commits/43eaccfa161c78355d61b04bbcd317e3466edebd.

core/arch/atomic_exports.h
ksco: Fixed in https://github.com/DynamoRIO/dynamorio/pull/6089/commits/43eaccfa161c78355d61b04bbcd317e3466edebd.

derekbruening (approved):
(I didn't go through the exact sequences)

Issue #2204, opened 2017-02-20
try "ctest --rerun-failed" for handling flaky tests


derekbruening:
For Travis #1962 and AppVeyor #2145 we may want to try adding --rerun-failed to ctest.

Will our output parsing handle it?

Is this new?  Which version of ctest added it?  Can we rely on it being there?  I don't remember seeing it when I first started using ctest.

derekbruening:
OK this is not what I thought it was.  Xref https://gitlab.kitware.com/cmake/cmake/issues/16314
What we really want is the "--retry-fail-until-pass N" being proposed which is not yet in the main branch.

derekbruening:
Xref #2984.

derekbruening:
We're using ctest_test().  Looks like in 3.17 it has support for retrying failed tests:

https://cmake.org/cmake/help/latest/command/ctest_test.html

`REPEAT UNTIL_PASS:3` would retry up to 3 times.

This may make greatly improve our problems with flaky tests on some platforms.

derekbruening:
I will try this out.  We're currently requiring cmake 3.7.  We should investigate the cmake versions on the GA CI VM's.  Looks like the Ubuntu20 image has cmake 3.26: https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md.  We don't use Ubuntu 18 anymore but it looks like it has 3.10 so requiring 3.17 could affect some users.

derekbruening:
The GA CI VS2019 image also has 3.26 https://github.com/actions/runner-images/blob/main/images/win/Windows2019-Readme.md

We could conditionally enable this feature if 3.17+ is detected, instead of mandating a new min version.

derekbruening:
Re-opening for two purposes:

1. We still have some tests that fail 3x in a row as noted here for scattergather #5329: https://github.com/DynamoRIO/dynamorio/issues/5873#issuecomment-1557561894.  We should track these and take some action on them (fix; add to auto-ignore list).
2. While it is nice to have the commits and PR pages look greener, we would like to notice when new tests start failing intermittently.  One idea is to have the post-merge Action have some scripting to identify failures that later passed and send an email (even if the final result is a green check).

derekbruening:
scattergather once again failed 3x in a row for x86-32 with the pipe assert #5329 for the merge to master of #6076.  So it is failing 3x in a row on a regular basis.

derekbruening:
[...]

PR #6091, opened 2023-05-27
i#3544 RV64: Implement TLS functions

ksco:
This patch adds the implementation of RV64-specific TLS functions.

Issue: #3544


core/unix/os.c
```diff
@@ -3182,7 +3178,7 @@ replace_thread_id(thread_id_t old, thread_id_t new)
     thread_id_t new_tid = new;
     ASSERT(is_thread_tls_initialized());
     DOCHECK(1, {
-        thread_id_t old_tid;
+        thread_id_t old_tid = 0;
```
abhinav92003, line 3181: Good to not leave these uninitialized, but just curious whether it caused any visible issue? Since these are anyway written in the READ macro.

core/unix/tls_linux_riscv64.c
```diff
@@ -48,29 +48,46 @@
...
+        return NULL;
+    return (byte **)(lib_tls_base + DR_TLS_BASE_OFFSET);
 }
 
 void
 tls_thread_init(os_local_state_t *os_tls, byte *segment)
 {
-    /* FIXME i#3544: Not implemented */
-    ASSERT_NOT_IMPLEMENTED(false);
+    ASSERT(read_thread_register(TLS_REG_LIB) != 0);
```
abhinav92003, line 60: s/0/NULL/

abhinav92003 (commented):
Would be good to get Derek's review too.

core/unix/os.c
ksco: I remember it causes compile errors if leaves uninitialized, I'll recheck later.

ksco: Only part of them are necessary, but I think we can initialize them anyway.

core/unix/tls_linux_riscv64.c
ksco: `read_thread_register` returns a `ptr_uint_t`, which is a `uint64`, so we can't use `NULL` here.

core/unix/os.c
```diff
@@ -2159,7 +2155,7 @@ get_local_state()
 void
 os_enter_dynamorio(void)
 {
-#    ifdef ARM
+#    if defined(ARM) || defined(RISCV64)
```
derekbruening, line 2158: Please s/ARM/AARCHXX/ as this looks like an oversight from the AArch64 port to stay ARM-only.

core/unix/tls_linux_riscv64.c
```diff
@@ -48,29 +48,60 @@
 byte **
 get_dr_tls_base_addr(void)
 {
```
derekbruening, line 50: This entire file is word-for-word identical to tls_linux_aarchxx.c.  Let's start a convention for sharing among AArchXX and RISCV: there may well be other things that are shared.  Using the term "RISC" comes to mind; hopefully not too close to "RISCV".  So let's rename the existing `tls_linux_aarchxx.c` to `tls_linux_risc.c` and use it on all 3 arches, is my proposal.

core/unix/os.c
ksco: Done

core/unix/tls_linux_riscv64.c
ksco: Great advice, done.

derekbruening (approved):

derekbruening:
Hmm, we had a merge queue failure again?  I looked at it and added it to https://github.com/DynamoRIO/dynamorio/issues/2204#issuecomment-1583885867.  Please add any observed merge queue failure to the same issue in the future if you encounter one so there is a list of still-flaky tests for someone to try to fix if someone has time in the future.

PR #6093, opened 2023-05-27
i#3544 RV64: Small improvments to some assembly functions

ksco:
This patch optimizes some basic assembly functions and adds some comments to improve readability.

Issue: #3544

### DIFF
--- a/core/arch/riscv64/riscv64.asm
+++ b/core/arch/riscv64/riscv64.asm
@@ -394,9 +394,11 @@ GLOBAL_LABEL(dr_setjmp:)
         fsd      fs10, 24*ARG_SZ (ARG1)
         fsd      fs11, 25*ARG_SZ (ARG1)
 # ifdef UNIX
-        mv       t0, ra
+        addi     sp, sp, -16
+        sd       ra, 0 (sp)
         jal      GLOBAL_REF(dr_setjmp_sigmask)
-        mv       ra, t0
+        ld       ra, 0 (sp)
+        add      sp, sp, 16
 # endif
         li       a0, 0
         ret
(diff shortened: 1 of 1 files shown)


core/arch/riscv64/riscv64.asm
```diff
@@ -368,8 +368,7 @@ GLOBAL_LABEL(dr_try_start:)
         DECLARE_FUNC(dr_setjmp)
```
abhinav92003, line 368: Per the commit messages section at https://dynamorio.org/page_code_reviews.html: we prefer to not keep the PR description empty. Please add a description of the changes made here.

core/arch/riscv64/riscv64.asm
```diff
@@ -395,11 +394,9 @@ GLOBAL_LABEL(dr_setjmp:)
         fsd      fs10, 24*ARG_SZ (ARG1)
         fsd      fs11, 25*ARG_SZ (ARG1)
 # ifdef UNIX
-        addi     sp, sp, -16
-        sd       ra, 0 (sp)
+        mv       t0, ra
```
abhinav92003, line 397: Based on the RISC-V calling convention, t0 is volatile across calls.

abhinav92003 (changes requested):

core/arch/riscv64/riscv64.asm
ksco: Done

core/arch/riscv64/riscv64.asm
ksco: Sorry, I reverted this change.

abhinav92003 (approved):

PR #6094, opened 2023-05-30
i#6024: Fix ThreadLocalStorage issue

### LINKED ISSUE #6024: Windows 11 and Windows 10 22H2 crash due to wrong handling of ThreadLocalStoragePointer
Issue is in **core/win32/loader.c**:

```
/* Isolate TEB->ThreadLocalStoragePointer. */
static void *pre_static_tls;
```

Then we follow to the first reference here in function **os_loader_init_prologue**

```
        if (should_swap_teb_static_tls()) {
            pre_static_tls = d_r_get_tls(STATIC_TLS_TIB_OFFSET);
            d_r_set_tls(STATIC_TLS_TIB_OFFSET, NULL);
            LOG(GLOBAL, LOG_LOADER, 2,
                "initial thread TEB->ThreadLocalStoragePointer=" PFX "\n",
                pre_static_tls);
        }
```

So far so good, **should_swap_teb_static_tls()** returns **false**, as **tls_next_idx == 0**. Then we load our client, and it will load **kernelbase.dll** which has TLS (Win11/Win10 22H2) in it which will cause **tls_next_idx** to be incremented in **privload_os_finalize**. Next check for  **should_swap_teb_static_tls** will return **true** and in **os_loader_thread_init_prologue** will lead to this code:

```
            if (should_swap_teb_static_tls()) {
                dcontext->app_static_tls = pre_static_tls;
                d_r_set_tls(STATIC_TLS_TIB_OFFSET, dcontext->app_static_tls);
            }
```

Which in the end will set **ThreadLocalStoragePointer** to **NULL** and will cause application crash.

Proposed solution is to always initialize **pre_static_tls** in **os_loader_init_prologue** and replace code with:

```
//        if (should_swap_teb_static_tls()) {
            //always save ThreadLocalStoragePointer
            pre_static_tls = d_r_get_tls(STATIC_TLS_TIB_OFFSET);
            d_r_set_tls(STATIC_TLS_TIB_OFFSET, NULL);
            LOG(GLOBAL, LOG_LOADER, 2,
                "initial thread TEB->ThreadLocalStoragePointer=" PFX "\n",
                pre_static_tls);
//        }
```

cr4zyserb:
Fixes the issue when ThreadLocalStorage is set to NULL, when client or dependency DLL is using TLS.  During loader initialization some actions were gated on TLS state in private libraries before those libraries are loaded, so those conditions were incorrect.

Fixes #6024


derekbruening:
Adding link to prior PR #6032 which had some code content discussion

core/win32/loader.c
```diff
@@ -253,13 +253,18 @@ os_loader_init_prologue(void)
...
+         * ThreadLocalStoragePointer to NULL. If we set it to NULL, it will lead to
+         * crash when application uses ThreadLocalStoragePointer (eg. __declspec(thread)
+         * is used), but client or kernelbase.dll (Win11/Win22H2) or client doesn't
+         * have TLS which will fail to restore TheadLocalStoragePointer later in
+         * calls to should_swap_teb_static_tls() as tlx_next_idx is 0
+         */
+        pre_static_tls = d_r_get_tls(STATIC_TLS_TIB_OFFSET);
+        LOG(GLOBAL, LOG_LOADER, 2,
+            "initial thread TEB->ThreadLocalStoragePointer=" PFX "\n",
+            pre_static_tls);
```
derekbruening, line 267: Is it possible to add a test that crashes without the fix and passes with the fix?  I guess our current test platform is https://github.com/actions/runner-images/blob/main/images/win/Windows2019-Readme.md which is Server 2019 which I think is Windows 10 1809 which is rather old: and this issue is only on 22H2.  What application and/or client is needed to trigger this?  #6024 doesn't say.  Would some of our existing tests fail if run on 22H2 or 11?

cr4zyserb: From what I have seen. This issue for example should happen if you use ``opcode_count.dll`` and try to load ``gdi32.dll`` from instrumented program, as ``opcode_count.dll`` has TLS, and ``gdi32.dll`` also uses TLS during ``DllMain``, or if you use ``__declspec(thread)`` in main program, while using ``opcode_count.dll`` client to instrument it. (at least in my builds ``opcode_count.dll`` comes with TLS which would increment ``tls_next_idx``)

derekbruening (approved):

core/win32/loader.c
derekbruening: But this won't trigger on our existing Github Actions test machines because they are too old, unfortunately.  So we'll put this in as-is since upgrading to newer Windows on those machines is a large project in its own right.

derekbruening:
Thank you for the PR!  I've added it to the merge queue.

Issue #1307, opened 2014-11-28
add extra decoding entries to get the "proper" opcode names for width or reg-reg variants?


derekbruening:
_From [bruen...@google.com](https://code.google.com/u/109494838902877177630/) on November 01, 2013 10:56:41_

There are a number of opcodes whose mnemonic name changes based on bitwidth or
on whether reg-reg or not: see the sequence in instr.h from OP_cbw through
OP_vmovlhps.  These are more substantially different than the condition code
variants: "jae" vs "jnb" doesn't matter much, while "pinsrd" vs "pinsrq" does
matter a litle as it tells you the size of the argment, which is hard to
determine with DR's default disassembly if it's a memory operand.  We could add
extra entries that split on modrm or on rex.w or on the data size prefix in
order to have the right name, but it's not clear it's worthwhile.

_Original issue: http://code.google.com/p/dynamorio/issues/detail?id=1307_

derekbruening:
There doesn't seem to be a separate issue on adding first-class support for disassembly in conventional assembler styles like Intel and AT&T for x86: xref #1683 for that on AArch32.  Generally we've thought about adding printf templates to each decode table entry -- but we would still need to split entries up as noted in this issue.

PR #6095, opened 2023-05-30
i#3544 RV64: Added an encoder and some fixes and improvments to the decoder

ksco:
This patch added the following changes:

1. Added an encoder
2. Fixed various disassemble issues
3. Optimized disassembled immediate format
4. Added unit tests for the encoder and decoder


core/ir/opnd_api.h
```diff
@@ -1820,6 +1820,12 @@ struct _opnd_t {
         /* Used for ARM: REG_kind, BASE_DISP_kind, and IMMED_INTEGER_kind */
         ushort /*dr_opnd_flags_t*/ flags;
     } aux;
+#    ifdef RISCV64
+    /* Used by disassemble, indicates whether an immediate value should be formatted
+     * in decimal or hexadecimal, valid iff opnd is an IMMED_INTEGER_kind.
+     */
+    bool decimal;
```
abhinav92003, line 1827: In what operands is this set to true vs false. Is there any semantic property we can attach to this? If yes, then this may be better off as a new dr_opnd_flags_t bit.

core/ir/riscv64/codec.c
```diff
@@ -1401,3 +1420,993 @@ decode_common(dcontext_t *dcontext, byte *pc, byte *orig_pc, instr_t *instr)
...
+/* Encode the immediate of the caddi16sp instruction:
+ * |15 13|   12   |11  7|6              2|1      0|
+ * | ... | imm[9] | ... | imm[4|6|8:7|5] | opcode |
+ *        ^------^       ^--------------^
+ * From:
+ * |31     9|8        4|3   0|
+ * | imm[9] | imm[8:4] |  0  |
+ */
+static bool
+encode_caddi16sp_imm_opnd(instr_t *instr, byte *pc, int idx, uint32_t *out)
```
abhinav92003, line 1998: I think it will be worth spending the effort to add the tests in this PR itself. E.g. see suite/tests/api/ir_aarch64.c

abhinav92003 (dismissed):

core/ir/riscv64/codec.c
ksco: Agree, will do that.

core/ir/opnd_api.h
ksco: I think it's only a format thing, i.e. make the disassembled string looks better.

abhinav92003: I mean -- for what operands did you decide to use the hex format vs the decimal format.

ksco: Basically, if an immediate is used as an arithmetic operand (like addi, slli), or as an offset (like load, store), it is better displayed as a decimal, as objdump does. It would be better to add a new dr_opnd_flags_t bit for this indeed.

ksco: But I didn't quite think of the name of this flag.

abhinav92003: I think it'd be reasonable to just add something like DR_OPND_IMM_PRINT_DECIMAL. @derekbruening any thoughts?

derekbruening: In general disassembly format has not historically had first-class support in DR's decoder.  For example, on x86 to support Intel or AT&T assembler format conventions we have code on top of the actual decode tables that kind of hackily and imperfectly applies those format conventions.  We've thought about giving disassembly in standard assembler styles first-class support: usually our idea has been to add a printf string template to each decode table entry (xref #1683 on this for AArch32 and #1307 on x86), but it's never been a high priority.  That said, if it can be done without any drawbacks to DR's normal operation (i.e., without increasing IR data type sizes or adding IR overhead), then I'm all for it.

A dr_opnd_flags_t bit seems like a reasonable solution; an alternative solution would be a printf string template per opcode encoding.  Are there other printing decisions that would need other IR changes, while a printf template would solve them all at once?  Or is this decimal-vs-hex the only foreseen issue?

I wonder if any other arches would want to use this dr_opnd_flags_t bit -- seems like for other arches it might be part of a runtime option for numeric print preference.

ksco: In addition to the display format of immediate data, known CSR registers may be better displayed by the name of the register. And fallback to hex decimals for unknown CSRs. Currently, all CSRs are displayed as hex.

ksco: Added `DR_OPND_IMM_PRINT_DECIMAL` enum in https://github.com/DynamoRIO/dynamorio/pull/6095/commits/2f2d61574fbd3569e463c11555389b17009d1642.

core/ir/riscv64/codec.c
ksco: Seems a lot of work is needed to turn on tests (i.e. make `make test` work), so I think it's better to submit in another PR.

abhinav92003: > Seems a lot of work is needed to turn on tests (i.e. make make test work)

Do you mean you're seeing test failures or issues in getting the tests set up?

Ideally tests should go together with code as that helps in detecting issues that may be missed in review by  humans. We seem to be adding a lot of encode/decode changes in this PR, for which tests will be even more useful. I think adding a working unit test module will improve dev velocity in the long term.

ksco: > Do you mean you're seeing test failures or issues in getting the tests set up?

Yes, when I run `make test`, most of the tests don't work. Some of them are caused by atomics that are fixed in https://github.com/DynamoRIO/dynamorio/pull/6089, some of them are not, but I haven't analyzed the cause. All tests are currently disabled in CI, so we don't see them failing.

I agree we should add unit tests for the encoder and decoder. But if I understand correctly, to do that, we need to get existing tests to work first. How about we keep this PR open, and I'll submit a new PR to make `make test` work first, after that merges, I can add the tests for this one. Sorry for the inconvenience.

derekbruening: There is no need to get all existing tests to work before adding some unit tests here.  There are multiple approaches.  Are they currently building now and BUILD_TESTS is on and enable_testing() is being called, and the only issue is that many tests fail when executed?  Then use a CTest label to select just a subset of tests to run by default and only label the new unit test(s).  We are using labels to limit the tests for other not-fully-supported-yet platforms like Mac: see the `OSX` label in suite/tests/CMakeLists.txt, or the other labels there like `RUNS_ON_QEMU`.

ksco: Thanks! Will try that.

ksco:
I wrote a simple test case in the above commits to check if things work ok, now I can use `ctest -L RISCV64` to run the `api.ir` test, but the test itself doesn't work (Segfaulted due to uninitialized `our_environ` variable used by `standalone_init`).

I noticed that the RISC-V version of `api.ir` binary doesn't contain the `_init` symbol, so some initialization routines are not being called before `main`.

```
$ # RISC-V
$ riscv64-linux-gnu-objdump -d ./suite/tests/bin/api.ir | grep "<_init>"
$
$ # x86_64
$ objdump -d ./suite/tests/bin/api.ir | grep "<_init>"
0000000000003000 <_init>:
$
```

I'm guessing it's a CMake mistake, but I'm not familiar with CMake, do you know what's possibly wrong? @abhinav92003 @derekbruening, thanks.

ksco:
Thank you! According to what you said, I noticed RISC-V gcc does not generate `_init/_fini` at all, I don't why, maybe they're too old and have been replaced by `init_array/fini_array`? Anyway, after the above commit, it seems working.

ksco:
I've added unit tests for the encoder and decoder and also enabled them in CI. Please review again, thanks!

core/ir/disassemble_shared.c
```diff
@@ -399,6 +399,11 @@ opnd_base_disp_disassemble(char *buf, size_t bufsz, size_t *sofar INOUT, opnd_t
                 print_to_buffer(buf, bufsz, sofar, "-");
             }
         }
+#if defined(RISCV64)
```
derekbruening, line 402: How about removing the ifdefs and adding DR_DISASM_RISCV instead.  We want to move away from ifdefs in core/ir so we can have one build of drdecode decode any arch (#1684).  This also matches ARM.

core/ir/opnd_api.h
```diff
@@ -1284,10 +1284,10 @@ enum {
     DR_REG_LAST_VALID_ENUM = DR_REG_FCSR, /**< Last valid register enum. */
     DR_REG_LAST_ENUM = DR_REG_FCSR,       /**< Last value of register enums. */
 
-    DR_REG_START_64 = DR_REG_X0,  /**< Start of 64-bit general register enum values. */
```
derekbruening, line 1287: Why remove "general"?  That is what distinguishes this from SIMD (x86 has 64-bit SIMD registers: MMX).

core/ir/opnd_api.h
```diff
@@ -2009,6 +2008,18 @@ DR_API
 opnd_t
 opnd_create_immed_int(ptr_int_t i, opnd_size_t data_size);
 
+#ifdef RISCV64
+DR_API
+/**
+ * Returns a signed immediate integer operand with value \p i and size
+ * \p data_size; \p data_size must be a OPSZ_ constant.
+ *
+ * The integer will be formatted as a decimal when disassemble.
```
derekbruening, line 2017: grammar: disassembled

core/ir/opnd_api.h
```diff
@@ -2009,6 +2008,18 @@ DR_API
...
+ */
+opnd_t
+opnd_create_immed_int_decimal(ptr_int_t i, opnd_size_t data_size);
```
derekbruening, line 2020: How about removing this function and instead using opnd_add_flags(onpd_create_immed_int(), DR_OPND_IMM_PRINT_DECIMAL).  That is what opnd_add_flags is for, to compose so we don't need to make special routines for combinations.

core/ir/opnd_api.h
```diff
@@ -2153,6 +2164,26 @@ opnd_t
...
+ * - base_reg + index_reg*scale + disp
+ *
+ * The operand has data size data_size (must be a OPSZ_ constant).
+ * Both \p base_reg and \p index_reg must be DR_REG_ constants.
+ * \p scale must be either 0, 1, 2, 4, or 8.
+ *
+ * The integer will be formatted as a decimal when disassemble.
+ */
+opnd_t
+opnd_create_base_disp_decimal(reg_id_t base_reg, reg_id_t index_reg, int scale, int disp,
```
derekbruening, line 2183: Let's remove this too and use opnd_add_flags with the existing creation routine.

core/unix/os.c
```diff
@@ -762,7 +762,7 @@ our_init(int argc, char **argv, char **envp)
     return 0;
 }
 
-#if defined(STATIC_LIBRARY) || defined(STANDALONE_UNIT_TEST)
+#if defined(STATIC_LIBRARY) || defined(STANDALONE_UNIT_TEST) || defined(RISCV64)
```
derekbruening, line 765: Please update the comment because as is it doesn't make sense why you'd add RISCV64 to the defines above.

suite/runsuite.cmake
```diff
@@ -555,9 +555,10 @@ endif ()
 if (ARCH_IS_X86 AND UNIX)
   # TODO i#3544: Run tests under QEMU
   set(orig_extra_ctest_args ${extra_ctest_args})
+  set(extra_ctest_args INCLUDE_LABEL RISCV64)
```
derekbruening, line 558: Add a comment that this is temporary with `TODO i#3544: Get all the tests working`.

suite/tests/CMakeLists.txt
```diff
@@ -5586,3 +5584,11 @@ if (LINUX AND X64 AND HAVE_RSEQ AND NOT RISCV64) # TODO i#3544: Port tests to RI
...
     PROPERTIES LABELS UBUNTU_22)
 endif ()
+
+# XXX i#3544: Enable passed tests on riscv64.
+if (RISCV64)
+  set_tests_properties(
+    code_api|api.ir
+    code_api|api.ir-static
+    PROPERTIES LABELS RISCV64)
+endif ()
```
derekbruening, line 5594: Missing newline at end of line 5594?

suite/tests/api/ir_riscv64.c
```diff
@@ -0,0 +1,1333 @@
...
+        ASSERT(instr_same(instr, decin));
+    }
+
+    instr_destroy(dc, instr);
+    instr_destroy(dc, decin);
+    return pc;
+}
+
+static void
+test_instr_encoding_branch(void *dc, uint opcode, instr_t *instr)
```
derekbruening, line 91: Why is this needed: it seems like test_instr_encoding() could be used for branches too.  Why do we have to remove some checks like instr_same for branches?  Add comments.

suite/tests/api/ir_riscv64.c
```diff
@@ -0,0 +1,1333 @@
...
+test_instr_encoding(void *dc, uint opcode, instr_t *instr)
+{
+    instr_t *decin;
+    byte *pc;
+
+    ASSERT(instr_get_opcode(instr) == opcode);
+    instr_disassemble(dc, instr, STDERR);
+    print("\n");
+    ASSERT(instr_is_encoding_possible(instr));
+    pc = instr_encode(dc, instr, buf);
```
derekbruening, line 75: Check that pc is not null

suite/tests/api/ir_riscv64.c
```diff
@@ -0,0 +1,1333 @@
...
+    decin = instr_create(dc);
+    decode(dc, buf, decin);
```
derekbruening, line 77: Check the return value

suite/tests/api/ir_riscv64.c
```diff
@@ -0,0 +1,1333 @@
...
+}
+
+static void
+test_instr_encoding_branch(void *dc, uint opcode, instr_t *instr)
+{
+    instr_t *decin;
+
+    ASSERT(instr_get_opcode(instr) == opcode);
+    ASSERT(instr_is_encoding_possible(instr));
+    instr_encode(dc, instr, buf);
```
derekbruening, line 97: Check the return value

suite/tests/api/ir_riscv64.c
```diff
@@ -0,0 +1,1333 @@
...
+    decin = instr_create(dc);
+    decode(dc, buf, decin);
```
derekbruening, line 99: Check the return value

derekbruening (commented):
Thank you for adding tests!  Just a few comments/suggestions.

core/ir/disassemble_shared.c
ksco: This is indeed a better way, done in https://github.com/DynamoRIO/dynamorio/pull/6095/commits/09399327a52f14082f4275c07f4d47f3a352f3eb.

core/ir/opnd_api.h
ksco: I thought it is used to distinguish integer regs from float regs. Sorry, reverted in https://github.com/DynamoRIO/dynamorio/pull/6095/commits/4dc5fb5ebcab5db0f70080a3ebd3f4a404457d1f.

core/ir/opnd_api.h
ksco: Fixed in https://github.com/DynamoRIO/dynamorio/pull/6095/commits/4dc5fb5ebcab5db0f70080a3ebd3f4a404457d1f.

core/ir/opnd_api.h
ksco: Yes, it's better, done in https://github.com/DynamoRIO/dynamorio/pull/6095/commits/f6aacbe145f8f671760d4f1639f4d0994937b3ea.

core/unix/os.c
ksco: Added comments for this change in https://github.com/DynamoRIO/dynamorio/pull/6095/commits/b127d66452df89e4ec84e81d7c04bbd09cf5937f.

suite/tests/api/ir_riscv64.c
ksco: Added comments in https://github.com/DynamoRIO/dynamorio/pull/6095/commits/9b7072f023b6b424c6c770cc8e461f74876d1df9.

suite/tests/api/ir_riscv64.c
ksco: Done in https://github.com/DynamoRIO/dynamorio/pull/6095/commits/9b7072f023b6b424c6c770cc8e461f74876d1df9.

core/ir/disassemble_shared.c
ksco: I also removed the register prefix `%` for RISC-V in above commit.

suite/tests/api/ir_riscv64.c
```diff
@@ -88,15 +90,25 @@ test_instr_encoding(void *dc, uint opcode, instr_t *instr)
 }
 
 static void
-test_instr_encoding_branch(void *dc, uint opcode, instr_t *instr)
+test_instr_encoding_jump_or_branch(void *dc, uint opcode, instr_t *instr)
 {
+    /* XXX i#3544: For jump and branch instructions, current disassembler will print
+     * the complete jump address, that is, a address relative to `buf`. But the
```
derekbruening, line 96: grammar nit: an address

derekbruening (approved):
Thank you for making the requested changes, it looks good.

derekbruening:
> @derekbruening Hi, I successfully ran a simple hello world on RISC-V, but the size of the patch is relatively large. Is it better to split it into small PRs and submit them separately?
> 
> This is probably not the right place to ask this kind of question, please let me know if there are other recommended ways.

Smaller is always better: easier and faster to review; easier to isolate and revert if something goes wrong.

PR #6097, opened 2023-05-30
i#6020 interval analysis: optimize basic_counts memory

abhinav92003:
Avoids tracking unique pc addresses in counters_t for interval state snapshots.

This takes up excessive memory on apps with a lot of threads, even if the
interval count is not that high. For apps with O(1K) threads and O(100) intervals,
the number of tracked "unique" pcs goes up by O(10^5), causing memory
usage to grow even beyond 300GB in some cases.

Issue: #6020

### DIFF
--- a/clients/drcachesim/tools/basic_counts.h
+++ b/clients/drcachesim/tools/basic_counts.h
@@ -97,8 +97,10 @@ public:
             icache_flushes += rhs.icache_flushes;
             dcache_flushes += rhs.dcache_flushes;
             encodings += rhs.encodings;
-            for (const uint64_t addr : rhs.unique_pc_addrs) {
-                unique_pc_addrs.insert(addr);
+            if (track_unique_pc_addrs) {
+                for (const uint64_t addr : rhs.unique_pc_addrs) {
+                    unique_pc_addrs.insert(addr);
+                }
             }
             return *this;
         }
@@ -168,6 +170,23 @@ public:
         // we use encoding_is_new as a proxy.
         int_least64_t encodings = 0;
         std::unordered_set<uint64_t> unique_pc_addrs;
+
+        // Stops tracking unique_pc_addrs. Tracking unique_pc_addrs can be very
+        // memory intensive. We skip it for interval state snapshots.
+        void
+        skip_tracking_unique_pc_addrs()
+        {
+            track_unique_pc_addrs = false;
+            unique_pc_addrs.clear();
+        }
+        bool
+        is_tracking_unique_pc_addrs() const
+        {
+            return track_unique_pc_addrs;
+        }
+
+    private:
+        bool track_unique_pc_addrs = true;
     };
     counters_t
     get_total_counts();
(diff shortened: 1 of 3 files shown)


clients/drcachesim/tools/basic_counts.h
```diff
@@ -168,6 +170,23 @@ class basic_counts_t : public analysis_tool_t {
         // we use encoding_is_new as a proxy.
         int_least64_t encodings = 0;
         std::unordered_set<uint64_t> unique_pc_addrs;
+
+        // Stops tracking unique_pc_addrs. Tracking unique_pc_addrs can be very
+        // memory intensive. We skip it for interval state snapshots.
```
derekbruening, line 175: Hmm, how many snapshots are created here?  The unique PC count is generally not all that high so I wouldn't expect this to matter unless we're making millions of snapshots and then I would say that's a usage error and we should generally stick to only about 1K snapshots (maybe even dynamically adjusting the intervals by merging them to keep a constant count, like Dr. Heapstat does with memory profile snapshots: maybe we should file a feature request on that so the user doesn't have to figure out the interval duration needed to get 1K snapshots).  Are there use cases that want more than a snapshot count in the 1K to 10K range?  Is it reasonable to need more than 1K-10K snapshots to find phase behavior that matters?

Do you have some example usage numbers you could list in the comment or the PR description?  I assume memory usage is tiny on any test in our suite or sthg like SPEC; this is only on huge proprietary apps?

derekbruening: Still trying to get a feel for this: what is the largest unique pc count you see in one thread's single interval?  Let's say it gets as high as 50K (does it?).  We then have 50K x 1K x 100 x 8-bytes-each = 40GB.  That is a bit on the high side.  But if it's more like 5K then it's 4GB which doesn't seem prohibitive when analyzing a huge trace.  I would suggest augmenting the comments added since although they say 10^5 really we were already doing per-thread so it's really just 100x bigger and if it wasn't super big to start with 100x may not matter, so I don't find the added comment accurate (not really 10^5) or persuasive without a baseline value to understand what 100x results in.

Also, having the unique pc count per interval seems kind of useful: just want to make sure we're making a good tradeoff.

abhinav92003: > Let's say it gets as high as 50K (does it?)

Out of the ~500 threads in this 10s trace, the top ~200 have more than 500k unique pcs; sum of unique pcs across all threads is 140M. (I got these stats from our existing per-thread results, so without my PR's changes).

abhinav92003: > so I don't find the added comment accurate (not really 10^5) 

You're right that relative to the existing per-thread, it increases only by O(#intervals). The `O(10^5)` factor I mentioned is relative to the unique pc count you'd expect in the whole trace (I assumed your earlier comment on unique pc count generally not being high was for the whole trace, which is also the only mode of operation here upstream)

derekbruening (approved):

PR #6108, opened 2023-06-03
drsyms: link against import targets

fmoessbauer:
We already create cmake import targets for the vendored libraries. By that, also use them in target_link_libraries instead of using the full path. In most cases this does not have any effect, but it helps when debugging linking issues. It also makes the logic easier to read.

### DIFF
--- a/ext/drsyms/CMakeLists.txt
+++ b/ext/drsyms/CMakeLists.txt
@@ -186,9 +186,9 @@ configure_extension(drsyms_static ON)
 configure_drsyms_target(drsyms_static)
 use_DynamoRIO_extension(drsyms_static drcontainers)
 
-target_link_libraries(drsyms "${dwarf_libpath}" "${elftc_libpath}")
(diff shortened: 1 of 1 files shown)


fmoessbauer:
IMHO the failed RISC-V test is unrelated to the change. I'll try it again.

abhinav92003 (approved):
The aarch64 and x86-64 `tool.drcacheoff.interval-count-output` failures are unrelated to this PR. I see them in a recent post-commit CI run (for a4c601b) too. Not sure why they didn't show up pre-commit. Looking into it right now. Should be a simple fix.

abhinav92003:
Created #6116 to fix `tool.drcacheoff.interval-count-output`.

abhinav92003:
> Created #6116 to fix `tool.drcacheoff.interval-count-output`.

#6116 is now submitted. If you update this branch and re-run the tests, the above failing test should pass.

PR #6110, opened 2023-06-05
i#5505 raw2trace: Upgrading raw2trace for processing kernel PT data

dolanzhao:
This update enhances raw2trace to decode kernel Processor Trace (PT) files and append the kernel trace to the output trace. 

- (1) Adds a new file type flag to indicate whether the output directory contains PT data. 
- (2) Upgrades raw2trace_directory to support managing kernel PT files.
- (3) Enables PT raw data conversion into trace_entry_t instances using drpt2ir and drir2trace in raw2trace.
- (4) Upgrades the basic_counts tool to support printing the number of userspace and kernel instructions.
- (5) Refactors pt2ir_t to use a new static member variable to store images shared between all pt2ir_t instances.
- (6) Fixes the memory leak issue on drpttracer.

Issue: #5505


suite/tests/CMakeLists.txt
```diff
@@ -4271,9 +4280,11 @@ if (BUILD_CLIENTS)
         set(cmd_pfx "sudo@")
         set(${testname_full}_precmd
           "foreach@${cmd_pfx}${CMAKE_COMMAND}@-E@remove_directory@${testname_full}.*.dir")
```
abhinav92003, line 4282: Another test we should add is to run basic counts on the user+kernel trace, to count the number of kernel instrs. It'd need some changes to basic counts too.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -91,9 +91,7 @@ namespace dynamorio {
 namespace drmemtrace {
 
 char logsubdir[MAXIMUM_PATH];
-#ifdef BUILD_PT_TRACER
```
abhinav92003, line 94: Why was this `#ifdef` removed?

clients/drcachesim/tracer/syscall_pt_trace.cpp
```diff
@@ -150,12 +152,11 @@ syscall_pt_trace_t::stop_syscall_pt_trace()
                                 pttracer_output_buffer_.data) != DRPTTRACER_SUCCESS) {
         return false;
     }
-
+    recorded_syscall_count_++;
     if (!trace_data_dump(pttracer_output_buffer_)) {
         return false;
     }
     cur_recording_sysnum_ = -1;
-    recorded_syscall_count_++;
```
abhinav92003, line 158: Better to start recorded_syscall_count_ at zero. We can then keep this increment and also increment before returning false.

abhinav92003 (changes requested):

suite/tests/CMakeLists.txt
dolanzhao: I will add a end marker first.

clients/drcachesim/common/trace_entry.h
```diff
@@ -914,7 +915,8 @@ typedef struct _syscall_pt_entry_t syscall_pt_entry_t;
 typedef enum {
     /* Index of a syscall PT entry of type SYSCALL_PT_ENTRY_TYPE_PID in the PDB header. */
     PDB_HEADER_PID_IDX = 0,
-    /* Index of a syscall PT entry of type SYSCALL_PT_ENTRY_TYPE_THREAD in the PDB header.
+    /* Index of a syscall PT entry of type SYSCALL_PT_ENTRY_TYPE_THREAD_ID in the PDB
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/drpt2trace/pt2ir.h
```diff
@@ -376,6 +386,12 @@ class pt2ir_t {
 
     /* The size of the PT data within the raw buffer. */
     uint64_t pt_raw_buffer_data_size_;
+
+    /* The shared image section cache.
+     * The pt2ir_t instance is designed to work with a single thread, while the image is
+     * shared among all threads.
+     */
+    static pt_iscache_autoclean_t share_iscache_;
```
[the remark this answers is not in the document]
dolanzhao: Yes. The libipt library incorporates pthread mutex to wrap all its data race operations, ensuring thread safety.

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -739,6 +745,12 @@ raw2trace_t::process_header(raw2trace_thread_data_t *tdata)
     VPRINT(2, "File %u is process %u\n", tdata->index, (uint)header.pid);
     thread_id_t tid = header.tid;
     tdata->tid = tid;
+    if (TESTANY(OFFLINE_FILE_TYPE_SYSCALL_PT, tdata->file_type)) {
+        auto it = kthread_files_map_.find(tid);
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -739,6 +745,12 @@ raw2trace_t::process_header(raw2trace_thread_data_t *tdata)
...
+        if (it != kthread_files_map_.end()) {
+            tdata->kthread_file = it->second;
+        }
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/tracer.cpp
dolanzhao: Revert.

clients/drcachesim/tracer/syscall_pt_trace.cpp
dolanzhao: Done.

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -784,6 +796,126 @@ raw2trace_t::process_header(raw2trace_thread_data_t *tdata)
...
+    if (tdata->pre_read_pt_entries[PDB_HEADER_DATA_BOUNDARY_IDX].pt_data_boundary.type !=
+        SYSCALL_PT_ENTRY_TYPE_PT_DATA_BOUNDARY) {
+        return "Invalid PT raw trace format";
+    }
+    if (tdata->pre_read_pt_entries[PDB_HEADER_SYSCALL_IDX_IDX].syscall_idx.type !=
+        SYSCALL_PT_ENTRY_TYPE_SYSCALL_IDX) {
+        return "Invalid PT raw trace format";
+    }
+    if (tdata->pre_read_pt_entries[PDB_HEADER_SYSCALL_IDX_IDX].syscall_idx.idx !=
+        syscall_idx) {
```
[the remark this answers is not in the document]
dolanzhao: Yes. Now it returns an error message.

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -784,6 +796,126 @@ raw2trace_t::process_header(raw2trace_thread_data_t *tdata)
...
 }
 
+std::string
+raw2trace_t::process_syscall_pt(raw2trace_thread_data_t *tdata, uint64_t syscall_idx)
+{
+#ifdef BUILD_PT_POST_PROCESSOR
+    if (tdata->kthread_file == nullptr)
+        return "";
+    if (!tdata->pt_metadata_processed) {
+        syscall_pt_entry_t header[PT_METADATA_PDB_HEADER_ENTRY_NUM];
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -2849,6 +3032,42 @@ trace_metadata_reader_t::check_entry_thread_start(const offline_entry_t *entry)
     return error;
 }
 
+#ifdef BUILD_PT_POST_PROCESSOR
+bool
+ktrace_metadata_reader_t::is_thread_shared_pt_metadata_header(
+    const syscall_pt_entry_t *entry)
+{
+    if (entry->pt_metadata_boundary.type != SYSCALL_PT_ENTRY_TYPE_PT_METADATA_BOUNDARY)
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -2849,6 +3032,42 @@ trace_metadata_reader_t::check_entry_thread_start(const offline_entry_t *entry)
...
+        return false;
+    return true;
+}
+
+bool
+ktrace_metadata_reader_t::is_syscall_pt_data_header(const syscall_pt_entry_t *entry)
+{
+    if (entry->pt_data_boundary.type != SYSCALL_PT_ENTRY_TYPE_PT_DATA_BOUNDARY)
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tracer/raw2trace_directory.cpp
```diff
@@ -418,6 +501,19 @@ raw2trace_directory_t::initialize(const std::string &indir, const std::string &o
     if (!err.empty())
         return err;
 
+    /* Open the kernel files. */
+    kcoredir_ = "";
+    kallsymsdir_ = "";
+    kernel_indir_ = indir_ + std::string(DIRSEP) + ".." + std::string(DIRSEP) +
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/drpt2trace/pt2ir.h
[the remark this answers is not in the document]
abhinav92003: Add a comment for that here.

clients/drcachesim/tools/basic_counts.cpp
```diff
@@ -106,6 +113,13 @@ basic_counts_t::parallel_shard_memref(void *shard_data, const memref_t &memref)
             ++counters->encodings;
     } else if (memref.instr.type == TRACE_TYPE_INSTR_NO_FETCH) {
         ++counters->instrs_nofetch;
+        if (TESTANY(OFFLINE_FILE_TYPE_SYSCALL_PT, per_shard->filetype_)) {
```
abhinav92003, line 116: Matching what we do currently, we should either add a new counter for these kernel_nofetch instrs, or just not count them. I think the latter option is okay because the kernel won't have these no-fetch instrs.

clients/drcachesim/tools/basic_counts.h
```diff
@@ -203,6 +223,8 @@ class basic_counts_t : public analysis_tool_t {
         std::string error;
         intptr_t last_window = -1;
         intptr_t filetype_ = -1;
+        /* Indicates whether is analyzing kernel trace. */
```
abhinav92003, line 226: "Indicates whether we're currently in the kernel region of the trace, which means we've seen a TRACE_MARKER_TYPE_SYSCALL_TRACE_START without its matching TRACE_MARKER_TYPE_SYSCALL_TRACE_STOP"

clients/drcachesim/tracer/raw2trace.cpp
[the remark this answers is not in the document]
abhinav92003: We don't seem to be throwing the error in the latest code. Is that not what we want?

clients/drcachesim/tools/basic_counts.cpp
```diff
@@ -215,6 +234,12 @@ basic_counts_t::print_counters(const counters_t &counters, int_least64_t num_thr
     }
     std::cerr << std::setw(12) << counters.instrs_nofetch << prefix
               << " non-fetched instructions\n";
+    if (counters.is_counting_kernel_instrs()) {
```
abhinav92003, line 237: Instead of tracking is_counting_kernel_instrs, can we just pass a bool to print_counters `for_kernel_trace` (set by the caller based on the file type, or maybe stored as a var in basic_counts_t) and use it here?

clients/drcachesim/tracer/raw2trace_directory.cpp
```diff
@@ -511,8 +511,11 @@ raw2trace_directory_t::initialize(const std::string &indir, const std::string &o
...
     kcoredir_ = "";
     kallsymsdir_ = "";
-    kernel_indir_ = indir_ + std::string(DIRSEP) + ".." + std::string(DIRSEP) +
+    kernel_indir_ = modfile_dir + std::string(DIRSEP) + ".." + std::string(DIRSEP) +
         DRMEMTRACE_KERNEL_PT_SUBDIR;
+    /* If the -enable_kernel_tracing option is not specified during tracing, the output
+     * directory will not include a kernel directory, and raw2trace will not process it.
+     */
     if (directory_iterator_t::is_directory(kernel_indir_)) {
         kcoredir_ = kernel_indir_ + std::string(DIRSEP) + DRMEMTRACE_KCORE_FILENAME;
```
abhinav92003, line 520: Does this also need to be conditional on BUILD_PT_POST_PROCESSOR?

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1363,17 +1365,18 @@ event_pre_syscall(void *drcontext, int sysnum)
 
 #ifdef BUILD_PT_TRACER
     if (op_offline.get_value() && op_enable_kernel_tracing.get_value()) {
-        if (!syscall_pt_trace_t::is_syscall_pt_trace_enabled(sysnum)) {
```
abhinav92003, line 1367: Why did we need to move this? If `!is_syscall_pt_trace_enabled`, then does the  `data->syscall_pt_trace.get_cur_recording_sysnum() != INVALID_SYSNUM` check even matter?

suite/tests/CMakeLists.txt
```diff
@@ -4266,11 +4266,12 @@ if (BUILD_CLIENTS)
   endif ()
 
   if (proc_supports_pt)
-    if (BUILD_PT_TRACER AND BUILD_PT_POST_PROCESSOR)
+    # TODO i#5505: Due to the excessive warning messages generated by the decoding of
+    # invalid instructions during the execution of drcachesim, the regex matching function
```
abhinav92003, line 4270: How about we print only the first ~10 instances of those error messages?

abhinav92003 (approved):
LGTM from my side. Please request a review from Derek after addressing the review comments.

clients/drcachesim/tracer/raw2trace.cpp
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/drpt2trace/pt2ir.h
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/tools/basic_counts.cpp
dolanzhao: Done.

clients/drcachesim/tools/basic_counts.h
dolanzhao: Done.

clients/drcachesim/tracer/tracer.cpp
dolanzhao: Yes. When encountering a new system call, verifying whether the tracing of the previous system call has been stopped is necessary. Regardless of whether or not the new system call can be traced, the checking of the previous system call should not be omitted. Skip the checking may result in the recording of more noise instructions.

clients/drcachesim/tools/basic_counts.cpp
dolanzhao: Done.

suite/tests/CMakeLists.txt
dolanzhao: Done.

clients/drcachesim/drpt2trace/ir2trace.cpp
```diff
@@ -74,7 +80,14 @@ ir2trace_t::convert(IN drir_t &drir, INOUT std::vector<trace_entry_t> &trace)
             }
         } else {
 #ifdef DEBUG
-            ERRMSG(WARNMSG_HEADER "Try to convert an invalid instruction.\n");
+            warning_msg_mutex_.lock();
```
abhinav92003, line 83: Hmm.. rethinking this suggestion. Having a lock-unlock at each instr is a bit inefficient. How about printing these errors only if verbosity > some_threshold? And then set a lower threshold in tests?

abhinav92003 (approved):

clients/drcachesim/drpt2trace/ir2trace.cpp
dolanzhao: Done.

clients/drcachesim/analyzer_multi.cpp
```diff
@@ -121,7 +121,8 @@ analyzer_multi_t::analyzer_multi_t()
                 dir.modfile_bytes_, dir.in_files_, dir.out_files_, dir.out_archives_,
                 dir.encoding_file_, dir.serial_schedule_file_, dir.cpu_schedule_file_,
                 nullptr, op_verbose.get_value(), op_jobs.get_value(),
-                op_alt_module_dir.get_value(), op_chunk_instr_count.get_value());
+                op_alt_module_dir.get_value(), op_chunk_instr_count.get_value(),
+                dir.in_kfiles_map_, dir.kcoredir_, dir.kallsymsdir_);
```
derekbruening, line 125: We should have done this earlier: this arg list is getting unwieldy being this long.  We should have an options struct instead.

clients/drcachesim/common/trace_entry.h
```diff
@@ -409,8 +409,12 @@ typedef enum {
     TRACE_MARKER_TYPE_PAGE_SIZE,
 
     /**
-     * This marker is emitted prior to each system call when -enable_kernel_tracing is
-     * specified. The marker value contains a unique system call identifier.
+     * This marker is emitted after each system call that records kernel PT, and its value
```
derekbruening, line 412: See the comment I added to event_post_syscall:

    /* TODO i#5505: We should move this _IDX marker to pre-syscall for two
     * reasons: 1) Some syscalls have no post event (e.g., exit or sigreturn);
     * 2) A blocking syscall may have a thread switch in between, separating the
     * syscall instruction from this marker by quite a distance when interleaved
     * with other threads.
     */

What was the original reason to place the marker in post-syscall?  Is there any advantage to doing so?  What about these significant disadvantages in the comment?

clients/drcachesim/common/trace_entry.h
```diff
@@ -472,6 +476,16 @@ typedef enum {
      */
     TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL,
 
+    /**
+     * Indicates a point in the trace where a syscall's trace start.
```
derekbruening, line 480: grammar: s/start/starts/

derekbruening, line 480: "syscall's kernel trace" would be more clear.  Ditto below.

clients/drcachesim/common/trace_entry.h
```diff
@@ -726,6 +740,7 @@ typedef enum {
      * 1 (success).
      */
     OFFLINE_FILE_TYPE_BLOCKING_SYSCALLS = 0x800,
+    OFFLINE_FILE_TYPE_SYSCALL_PT = 0x1000, /**< Syscalls' PT tracings are included. */
```
derekbruening, line 743: Explain what "PT" means.  Also, if we use equivalent technology on other platforms, we could use this same filetype (the arch should be enough to identify the technology), right?  So rename to OFFLINE_FILE_TYPE_KERNEL_SYSCALLS or something?

clients/drcachesim/common/trace_entry.h
```diff
@@ -982,23 +997,31 @@ typedef struct _syscall_pt_entry_t syscall_pt_entry_t;
         (SYSCALL_METADATA_ENTRY_NUM * sizeof(syscall_pt_entry_t))
 
 typedef enum {
-    /* Index of a syscall PT entry of type SYSCALL_PT_ENTRY_TYPE_PID in the PDB header. */
+    /**
```
derekbruening, line 1000: Are these really public enums?  Aren't these internal to the PT code?  Why do they have Doxygen comments?

clients/drcachesim/common/trace_entry.h
```diff
@@ -1039,4 +1062,22 @@ typedef enum {
...
 #define DRMEMTRACE_CPU_SCHEDULE_FILENAME "cpu_schedule.bin.zip"
 
+/**
+ * The name of the folder in -offline mode where the kernel's per thread PT data is
+ * stored. This data is captured during online tracing.
+ */
+#define DRMEMTRACE_KERNEL_PT_SUBDIR "kernel.raw"
+
+/**
+ * The name of the file in -offline mode where the kernel code segments is stored. This
```
derekbruening, line 1072: grammar: s/is/are/

clients/drcachesim/common/trace_entry.h
```diff
@@ -1039,4 +1062,22 @@ typedef enum {
  */
 #define DRMEMTRACE_CPU_SCHEDULE_FILENAME "cpu_schedule.bin.zip"
 
+/**
+ * The name of the folder in -offline mode where the kernel's per thread PT data is
+ * stored. This data is captured during online tracing.
```
derekbruening, line 1067: "online" is going to be confusing as "online" is used everywhere else here to mean the analysis tool is operating during tracing.  Best to delete "online" here and below.

derekbruening (commented):
Comments on the 1st 3 files.

clients/drcachesim/analyzer_multi.cpp
derekbruening: Probably such a change should be done separately and does involve changes to external subclasses.  Add a TODO and file an issue?

clients/drcachesim/common/trace_entry.h
```diff
@@ -472,6 +476,16 @@ typedef enum {
      */
     TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL,
 
+    /**
+     * Indicates a point in the trace where a syscall's trace start.
+     */
+    TRACE_MARKER_TYPE_SYSCALL_TRACE_START,
```
derekbruening, line 482: Add printing code for every new marker to tools/view.cpp.

derekbruening (commented):
Not sure it's necessary for me to re-review all the code already reviewed so will stop here.

clients/drcachesim/CMakeLists.txt
```diff
@@ -201,6 +201,11 @@ target_link_libraries(directory_iterator drfrontendlib)
 
 if (BUILD_PT_POST_PROCESSOR)
   add_subdirectory(drpt2trace)
+  # Add the directory containing libipt.so and libipt-sb.so to the linker search path.
+  set(libipt_install_dir "${PROJECT_BINARY_DIR}/third_party/${INSTALL_LIB}")
+  SET(CMAKE_EXE_LINKER_FLAGS
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/CMakeLists.txt
```diff
@@ -201,6 +201,11 @@ target_link_libraries(directory_iterator drfrontendlib)
...
+    "${CMAKE_EXE_LINKER_FLAGS} -Wl,-rpath='${libipt_install_dir}'"
```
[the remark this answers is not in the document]
dolanzhao: Done.

clients/drcachesim/common/trace_entry.h
dolanzhao: Done.

dolanzhao: Done.

clients/drcachesim/common/trace_entry.h
dolanzhao: Done.

clients/drcachesim/common/trace_entry.h
dolanzhao: These aren't public enums. I convert them into the internal comments format.

clients/drcachesim/common/trace_entry.h
dolanzhao: Done.

clients/drcachesim/common/trace_entry.h
dolanzhao: Done.

clients/drcachesim/common/trace_entry.h
dolanzhao: Done.

clients/drcachesim/common/trace_entry.h
dolanzhao: Done.

clients/drcachesim/analyzer_multi.cpp
dolanzhao: Create a new feature request [issue#6145](https://github.com/DynamoRIO/dynamorio/issues/6145).

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -388,6 +391,22 @@ struct trace_metadata_reader_t {
     check_entry_thread_start(const offline_entry_t *entry);
 };
 
+#ifdef BUILD_PT_POST_PROCESSOR
+/**
```
derekbruening, line 395: If a struct is Doxygen-documented it should have Doxygen comments on every field, not just the struct itself.  However, I don't think it needs any docs.  raw2trace used to be split into pieces and one was exported for use externally and that is why it has some Doxygen-documented interfaces: probably those should have been changed to regular comments when that split and that exposed interface was refactored and removed recently.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -388,6 +391,22 @@ struct trace_metadata_reader_t {
...
+ * Functions for decoding and verifying raw memtrace data headers.
+ */
+struct ktrace_metadata_reader_t {
+    static bool
+    is_pt_metadata_header(const syscall_pt_entry_t *entry);
```
derekbruening, line 400: I would expect these function operating on this syscall_pt_entry_t type to be part of the syscall_pt_entry_t struct, not some separate struct.

derekbruening, line 400: style: A struct isn't supposed to have functions; if it has more than constructors we generally make it a class.

dolanzhao: These functions are very simple and only be called in a single location. Therefore, it would be most efficient to inline all of these functions.

abhinav92003: IIRC you pointed out to me that we also have the similar trace_metadata_reader_t which is also a struct with functions.

abhinav92003: And that we wanted to do the same thing for kernel traces to allow for more complex logic in future.

derekbruening: Looking -- ok I see this trace_metadata_reader_t and _writer_t.  They don't seem appealing to me.  I don't think we should model new code on those.  Probably they were part of the raw2trace refactoring done by someone else who was trying to move pieces to external code for parallelism -- but we don't need that anymore.

abhinav92003 (approved):

suite/tests/CMakeLists.txt
```diff
@@ -4298,10 +4296,8 @@ if (BUILD_CLIENTS)
         set(cmd_pfx "sudo@")
         set(${testname_full}_precmd
           "foreach@${cmd_pfx}${CMAKE_COMMAND}@-E@remove_directory@${testname_full}.*.dir")
-        set (${testname_full}_postcmd
-          "${cmd_pfx}chmod@777@-R@${testname_full}.*.dir")
-        set(${testname_full}_postcmd2
-          "firstglob@${cmd_pfx}${drpt2trace_path}@-mode@ELF@-raw_pt_format@DRMEMTRACE@-raw_pt@${testname_full}.*.dir/kernel.raw/*.pt@-elf@${testname_full}.*.dir/kernel.raw/kcore")
+        set(${testname_full}_postcmd
+          "firstglob@${cmd_pfx}${drcachesim_path}@-simulator_type@basic_counts@-verbose@0@-indir@${testname_full}.*.dir${sim_atops}")
```
abhinav92003, line 4300: If zero verbosity is the default, do we really need to specify it again?

dolanzhao: Done.

PR #6111, opened 2023-06-05
Fix dr_insert_clean_call callee arg in tutorial

Pavel-Durov:
Fixes the callee arg in the dr_insert_clean_call call in the dynamic basic block size tutorial. The callee pointer should be cast to `void*` for it to compile successfully.

Related to: https://groups.google.com/g/dynamorio-users/c/BWVdySjLUn8


api/docs/tutorial.dox
```diff
@@ -300,7 +300,7 @@ event_basic_block(void *drcontext, void *tag, instrlist_t *bb,
     dr_mutex_unlock(as_built_lock);
```
abhinav92003, line 300: Thanks for the fix! This looks good to me.

Just a quick note about the PR description and title: please see https://dynamorio.org/page_code_reviews.html#sec_code_review_non_member.

Particularly, it'll be clearer if the PR title is something like: `Fix dr_insert_clean_call callee arg in tutorial`

And the PR description could be:

```
Fixes the callee arg in the dr_insert_clean_call call in the dynamic basic block size tutorial. The callee pointer should be cast to void* for it to compile successfully.
```

api/docs/tutorial.dox
```diff
@@ -300,7 +300,7 @@ event_basic_block(void *drcontext, void *tag, instrlist_t *bb,
...
 
 +     /* insert clean call */
-+     dr_insert_clean_call(drcontext, bb, instrlist_first(bb), clean_call, false, 1,
++     dr_insert_clean_call(drcontext, bb, instrlist_first(bb), (void *)clean_call, false, 1,
```
abhinav92003, line 303: Move `false, 1` to the next line to avoid too long line.

abhinav92003 (changes requested):

api/docs/tutorial.dox
Pavel-Durov: done

derekbruening:
Just saw the merge email on this: this was never merged?  There is now a weird OSX docs build error.  It seems to be on every PR e.g. #6990.  Filed #6992.

Issue #6112, opened 2023-06-05
runmulti.cmake fails to require a full test output match


derekbruening:
For PR #6096 I had a test passing when the output file was missing the first line.  The reason is that runmulti.cmake does not add `^$` around the regex (this is added for single-command tests for their `PASS_REGULAR_EXPRESSION`).  I'll add it and run all the test and we'll see which existing ones fail.

PR #6114, opened 2023-06-06
i#6107: Fix schedule file count error

### LINKED ISSUE #6107: drmemtrace schedule file errors hit by new drcacheoff.sysnums test
Yet another issue discovered by PR #6096: it adds a large-ish test under -test_mode and the invariant checker raises 2 errors.

First, we have for 32-bit x86:
```
316:   Trace invariant failure in T-1 at ref # 0: Serial schedule entry count does
316:   not match trace
```
It's b/c there's no zip lib support, and the raw2trace schedule recording
is dependent on that -- despite the serial schedule not needing it (though
I guess the original point of the serial schedule file was to help with
fast forward which only works with zip).  raw2trace ends up producing a
zero-sized serial file (and no cpu file).

Next, for aarch64:
```
237:   Trace invariant failure in T-1 at ref # 46937: Serial schedule entry does
237:   not match trace
```
I can reproduce this one too:
```
(gdb) p next
$1 = {thread = 4129782, timestamp = 13330207028197999, cpu = 48, start_instruction = 205}
(gdb) p serial[global.ref_count_]
$2 = {thread = 4129782, timestamp = 13330207028197999, cpu = 48, start_instruction = 10000205}
(gdb) p serial[global.ref_count_-1]
$3 = {thread = 4129782, timestamp = 13330207028197984, cpu = 48, start_instruction = 9999995}
(gdb) p global.ref_count_
$4 = 46976
(gdb) p sizeof(next)
$5 = 32
```
And indeed in the file it wraps around at 10M:
```
$ zcat suite/tests/tool.drcacheoff.sysnums.linux.signal_pre_syscall.4129782.9480.dir/trace/serial_schedule.bin.gz | od -t d8 -Ax -w32 | less
16efc0              4129782    13330207028197969                   48              9999785
16efe0              4129782    13330207028197984                   48              9999995
16f000              4129782    13330207028197999                   48                  205
16f020              4129782    13330207028198014                   48                  415
```
Investigating why that is (10M is a strange boundary).

derekbruening:
Fixes an error in raw2trace where the instruction count it writes in the schedule files was incorrectly modulo the chunk count.

Adds checks in the scheduler to raise an error on a decreasing count.

Adds a workaround where the scheduler detects a schedule file with the bug and corrects the counts.  This handles multiple inputs by building and sorting per-input sequences and then updating the schedule structures in a separate
step using a timestamp-keyed hashtable.

Adds a unit test for the workaround.  I also tested this on a real, large schedule file from the linux.signal-pre-syscall test which is part of PR #6096.

Fixes the scheduler to support replaying as-traced with a single output if that is what the user requests (I needed this for the just-mentioned manual test on the pre-syscall trace).

Disables writing any serial schedule file without zip support.  The raw2trace code was already conditional on zip support, so we were creating an empty file.  The serial file is meant only for fast skipping support, which requires zip format. 
This fixes an invariant check failure in the forthcoming drcacheoff.sysnums test in PR #6096 from an emtpy file.

Issue: #5843, #6107
Fixes #6107

### DIFF
--- a/clients/drcachesim/scheduler/scheduler.cpp
+++ b/clients/drcachesim/scheduler/scheduler.cpp
@@ -30,6 +30,7 @@
  * DAMAGE.
  */
 
+#include <algorithm>
 #include <set>
 #include <unordered_map>
 
@@ -793,8 +794,8 @@ scheduler_tmpl_t<RecordType, ReaderType>::read_traced_schedule()
     // We also want to collapse same-cpu consecutive records so we start with
     // a temporary local vector.
     std::vector<std::vector<schedule_record_t>> all_sched(outputs_.size());
-    uint64_t prev_start = 0; // Only used to detect a decrease.
-    uint64_t add_to_start = 0;
+    // Work around i#6107 by tracking counts sorted by timestamp for each input.
+    std::vector<std::vector<schedule_record_t>> input_sched(inputs_.size());
     while (options_.replay_as_traced_istream->read(reinterpret_cast<char *>(&entry),
                                                    sizeof(entry))) {
         if (entry.cpu != cur_cpu) {
@@ -806,29 +807,10 @@ scheduler_tmpl_t<RecordType, ReaderType>::read_traced_schedule()
                 }
             }
             cur_cpu = entry.cpu;
-            prev_start = 0;
         }
         input_ordinal_t input = tid2input[entry.thread];
         // We'll fill in the stop ordinal in our second pass below.
         uint64_t start = entry.start_instruction;
-        if (start < prev_start) {
-            // Work around i#6107 where the counts in the file are incorrectly modulo
-            // the chunk size.  We haven't read into the trace far enough to find the
-            // actual chunk size for for this workaround we only support what was the
-            // default in raw2trace up to this point, 10M.
-            static constexpr uint64_t DEFAULT_CHUNK_SIZE = 10 * 1000 * 1000;
-            // If within 50% of the end of the chunk we assume it's i#6107.
-            if (prev_start * 2 > DEFAULT_CHUNK_SIZE) {
-                add_to_start += DEFAULT_CHUNK_SIZE;
-                VPRINT(this, 2, "Working around i#6107 by adding %" PRId64 " instrs\n",
-                       add_to_start);
-            } else {
-                error_string_ = "Invalid decreasing start field in schedule file";
-                return STATUS_ERROR_INVALID_PARAMETER;
-            }
-        }
-        prev_start = start;
-        start += add_to_start;
         uint64_t timestamp = entry.timestamp;
         // Some entries have no instructions (there is an entry for each timestamp, and
         // a signal can come in after a prior timestamp with no intervening instrs).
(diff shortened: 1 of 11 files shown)


derekbruening:
The test failure looks like one of the #6112 test output expansions was too limited for interval-count-output.  Can tweak here or could be separate PR.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -561,14 +562,14 @@ scheduler_tmpl_t<RecordType, ReaderType>::set_initial_schedule(
             VPRINT(this, 2, "Assigning input #%d to output #%zd\n", i, index);
         }
     } else if (options_.mapping == MAP_TO_RECORDED_OUTPUT) {
-        if (outputs_.size() > 1) {
-            if (options_.replay_as_traced_istream == nullptr)
-                return STATUS_ERROR_INVALID_PARAMETER;
+        if (options_.replay_as_traced_istream != nullptr) {
             sched_type_t::scheduler_status_t status = read_traced_schedule();
```
abhinav92003, line 566: So now control may flow here for `output_.size() == 1` too? There's no schedule as-such for a single output stream right?

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -2562,13 +2562,16 @@ raw2trace_t::write(raw2trace_thread_data_t *tdata, const trace_entry_t *start,
                 if (it->size == TRACE_MARKER_TYPE_TIMESTAMP)
                     tdata->last_timestamp_ = it->addr;
                 else if (it->size == TRACE_MARKER_TYPE_CPU_ID) {
+                    DR_CHECK(tdata->chunk_count_ > 0, "chunk_count_ should be +1");
```
abhinav92003, line 2565: s/+1/+ve/?

clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -1884,6 +1884,90 @@ test_replay_as_traced()
...
+    if (scheduler.init(sched_inputs, NUM_OUTPUTS, sched_ops) !=
+        scheduler_t::STATUS_SUCCESS)
+        assert(false);
+    // Since it initialized we didn't get an invalid schedule order.
+    // Make sure the stream works too.
+    auto *stream = scheduler.get_stream(0);
+    memref_t memref;
+    for (scheduler_t::stream_status_t status = stream->next_record(memref);
+         status != scheduler_t::STATUS_EOF; status = stream->next_record(memref)) {
+        assert(status == scheduler_t::STATUS_OK);
```
abhinav92003, line 1966: So with the fix in this PR, a schedule with the bug would return a non-ok status if it were not for the backward-compatibility workaround?

abhinav92003 (approved):

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: A schedule with the bug fails in scheduler_t::init up front.

clients/drcachesim/tracer/raw2trace.cpp
derekbruening: Reworded to "chunk_count_ should have been incremented already"

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: I found it useful to replay the schedule for one output for some manual tests and there's no real reason to not allow it if the user wants to do it: but yes it should match the analyzer serial mode so there's no big benefit to reading the schedule file.

derekbruening: Added a comment.

PR #6115, opened 2023-06-06
raw2trace optimization: Use lz4 compression for final trace files to speedup conversation time

kuhanov:
Right now we use 3-phase scheme for trace processing: drmemtrace(collector) + drraw2trace(convertor) + analysis tools.
And trace converting is the most time consuming part. Ex. java -version
drmemtrace: real 0m36.873s user 0m43.101s sys 0m4.459s
drraw2trace: real 2m27.796s user 3m1.469s sys 0m1.248s
opcodes tool: real 0m38.956s user 0m48.660s sys 0m0.340s

We used specjvm2008 benchmarks for verification (crypto.aes, xml.transform, crypto.rsa)

crypto.aes
Original convertation time      293m
lz4 converstation time             137m

xml.transform
Original convertation time      293m
lz4 converstation time             137m

crypto.rsa
Original convertation time      293m
lz4 converstation time             137m

Adding bitset_hash_table_t gives 20% speed up if we stay on original approach.


clients/drcachesim/CMakeLists.txt
```diff
@@ -136,6 +136,7 @@ endif()
```
derekbruening, line 136: I found the PR description not very clear: please reword to make it clear that this adding lz4 as a supported file format for final trace files (it is already used for raw trace files).

Please change the title: "migration on lz4" is not grammatically correct at least for American English.  I would s/migration on lz4/use lz4 for final trace files/.

Also in the description I don't know what "drmemtrace.crypto.aes" and the other similar strings means: is that naming an application or benchmark or something with a "drmemtrace." prefix for some reason?

derekbruening, line 136: There are drmemtrace features that rely on the .zip archive format, such as fast skipping over large instruction counts, which is in turn a key part of the simulator (and regular analysis tool) interface for regions of interest or simpoints.  Such fast skipping is not possible with a single-encrypted file like this lz4 format so we are not excited about promoting it and are not likely to use it ourselves.  We may also continue adding features that require zip.

Have you tried Z_BEST_SPEED instead of Z_DEFAULT_COMPRESSION?

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -85,7 +86,10 @@
 #define WINDOW_SUBDIR_FORMAT "window.%04zd" /* ptr_int_t is the window number type. */
 #define WINDOW_SUBDIR_FIRST "window.0000"
 #define TRACE_SUBDIR "trace"
-#ifdef HAS_ZLIB
+
+#ifdef HAS_LZ4
+#    define TRACE_SUFFIX "trace.lz4"
```
derekbruening, line 91: This should not be the default as lz4 does not support fast skipping: see the other comment.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -936,7 +1042,7 @@ class raw2trace_t {
         uint last_cpu_ = 0;
         app_pc last_pc_if_syscall_ = 0;
 
-        std::set<app_pc> encoding_emitted;
+        bitset_hash_table_t<app_pc> encoding_emitted;
```
derekbruening, line 1045: This is not in the PR description: please add it there.

What % of speedup you see is due to this vs lz4?

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1122,7 +1228,8 @@ class raw2trace_t {
     // instruction.
     std::string
     write(raw2trace_thread_data_t *tdata, const trace_entry_t *start,
-          const trace_entry_t *end, std::vector<app_pc> decode_pcs = {});
+          const trace_entry_t *end, app_pc *decode_pcs = nullptr,
+          size_t decode_pcs_size = 0);
```
derekbruening, line 1232: These other changes are also not in the PR description: please add.

What % of speedup is from these?

clients/drcachesim/CMakeLists.txt
derekbruening, line 136: Please include in the PR description the size comparison for lz4 vs zip.  IIRC lz4 size was pretty good despite being faster, from evaluating it for the raw files, but let's have some actual numbers, as for some users size matters more than time for the raw2trace step.

clients/drcachesim/tracer/raw2trace.h
kuhanov: bitset_hash_table_t have us about 20% speed up

clients/drcachesim/CMakeLists.txt
derekbruening: If your use cases don't need fast skipping, lz4 may well work well.  A chunked format using lz4 is possible but not as easy as zip.  If you could make lz4 not be the default in this PR (we'd prefer the skip features to be there by default) and require an option to turn on that sounds good.

kuhanov: > Derek, we performed some measurements from conversion of trace got from Firefox browser startup. These are the results:
> 
> gz	zip DEFAULT	zip BEST_SPEED	lz4 fast	lz4 CLEVEL_MIN	gz+patch	lz4 fast +patch	lz4 CLEVEL_MIN +patch
> debug	13m24.289s (942.9 MB)	17m22.780s (945.5 MB)	16m25.397s (1.3 GB)	11m36.813s (1.4 GB)	11m48.851s (834.2 MB)	6m15.137s (942.9 MB)	4m22.005s (1.4 GB)	4m47.105s (834.2 MB)
> release	3m52.183s (942.9 MB)	5m10.639s (945.5 MB)	4m3.868s (1.3 GB)	2m5.202s (1.4 GB)	2m26.787s (834.2 MB)	3m10.944s (942.9 MB)	1m9.434s (1.4 GB)	1m32.259s (834.2 MB)

thx Yury for data

kuhanov:
Hi, @derekbruening. We fixed our patch to have green tests. Could you review patch now? Thx, Kirill

derekbruening (commented):
Made it part-way; will try to finish tomorrow.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -933,6 +1037,10 @@ class raw2trace_t {
 
         // Used to delay a thread-buffer-final branch to keep it next to its target.
         std::vector<trace_entry_t> delayed_branch;
+        // This is the first step of optimization of using delayed_branch vector
+        // Checking the bool value is cheeper than delayed_branch.empty()
+        // In general it's better to avoid this vector at all.
+        bool delayed_branch_empty_ { true };
```
derekbruening, line 1043: Best to match the existing code using ` = ` rather than list init.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -933,6 +1037,10 @@ class raw2trace_t {
 
         // Used to delay a thread-buffer-final branch to keep it next to its target.
         std::vector<trace_entry_t> delayed_branch;
+        // This is the first step of optimization of using delayed_branch vector
```
derekbruening, line 1040: style: Missing end-punctuation `.`

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -933,6 +1037,10 @@ class raw2trace_t {
...
+        // Checking the bool value is cheeper than delayed_branch.empty()
```
derekbruening, line 1041: spelling: cheaper
style: missing `.`

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -2775,7 +2777,7 @@ raw2trace_t::insert_post_chunk_encodings(raw2trace_thread_data_t *tdata,
 // extra processing here).
 std::string
 raw2trace_t::write(raw2trace_thread_data_t *tdata, const trace_entry_t *start,
-                   const trace_entry_t *end, std::vector<app_pc> decode_pcs)
+                   const trace_entry_t *end, app_pc *decode_pcs, size_t decode_pcs_size)
```
derekbruening, line 2780: If this is about the vector copy could we just pass the vector by reference?

clients/drcachesim/tracer/raw2trace.h
derekbruening: This was resolved but I don't see this in the updated PR description?  I'm seeing 4 different changes here:

1. lz4
2. bitset instead of set for encoding history
3. not passing vector by value
4. keeping bool for whether delayed branch container is empty

I think these should be explicitly listed in the PR description.  Otherwise it is confusing reading the code and puzzling over how these relate to lz4 or the bitset.  Maybe at least some of these should be in a separate PR; maybe the lz4 in one PR, and the other 3 in another.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -672,6 +679,103 @@ struct trace_header_t {
...
+    const size_t BASIC_BUCKET_COUNT = (1 << 15);
+    std::unordered_map<T, std::bitset<BLOCK_SIZE>> page_table;
+    typename std::unordered_map<T, std::bitset<BLOCK_SIZE>>::iterator last_block =
+        page_table.end();
+
+    inline std::pair<T, ushort>
+    convert(T pc)
+    {
+        return std::pair<T, ushort>(
+            reinterpret_cast<T>(reinterpret_cast<size_t>(pc) & (~(BLOCK_SIZE - 1))),
```
derekbruening, line 695: Please add comments to make it more clear what this function is doing and why it's doing it.  A more descriptive name for the function which indicates what it is converting to would help.  Creating a custom struct with named fields would also help instead of using pair which leaves us with "first" and "second" which do not help readability.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -672,6 +679,103 @@ struct trace_header_t {
             return msg;    \
     } while (0)
 
+template <typename T> class bitset_hash_table_t {
+private:
+    const static size_t BLOCK_SIZE_BIT = 13;
+    const static size_t BLOCK_SIZE = (1 << BLOCK_SIZE_BIT);
+    const size_t BASIC_BUCKET_COUNT = (1 << 15);
+    std::unordered_map<T, std::bitset<BLOCK_SIZE>> page_table;
```
derekbruening, line 687: style: class member variable names end in _

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -672,6 +679,103 @@ struct trace_header_t {
...
+    typename std::unordered_map<T, std::bitset<BLOCK_SIZE>>::iterator last_block =
```
derekbruening, line 688: style: class member variable names end in _

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -672,6 +679,103 @@ struct trace_header_t {
             return msg;    \
     } while (0)
 
+template <typename T> class bitset_hash_table_t {
+private:
+    const static size_t BLOCK_SIZE_BIT = 13;
```
derekbruening, line 684: Add a comment on how this value was picked.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -672,6 +679,103 @@ struct trace_header_t {
...
+    const static size_t BLOCK_SIZE = (1 << BLOCK_SIZE_BIT);
+    const size_t BASIC_BUCKET_COUNT = (1 << 15);
```
derekbruening, line 686: Add a comment on how this value was picked.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -672,6 +679,103 @@ struct trace_header_t {
...
+    typename std::unordered_map<T, std::bitset<BLOCK_SIZE>>::iterator last_block =
+        page_table.end();
+
+    inline std::pair<T, ushort>
+    convert(T pc)
+    {
+        return std::pair<T, ushort>(
+            reinterpret_cast<T>(reinterpret_cast<size_t>(pc) & (~(BLOCK_SIZE - 1))),
+            static_cast<ushort>(reinterpret_cast<size_t>(pc) & (BLOCK_SIZE - 1) &
+                                0xFFFF));
```
derekbruening, line 697: Looks like there are multiple assumptions here on BLOCK_SIZE and this final value fitting in a ushort which should be documented.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -672,6 +679,103 @@ struct trace_header_t {
...
+            if (last_block == page_table.end()) {
+                last_block = (page_table.emplace(std::make_pair(
+                                  block.first, std::move(std::bitset<BLOCK_SIZE>()))))
+                                 .first;
+                last_block->second[block.second] = 1;
+                return true;
+            }
+        }
+        if (last_block->second[block.second] == 1)
+            return false;
```
derekbruening, line 735: This looks backward: if the bit is 1 shouldn't it return true?

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -672,6 +679,103 @@ struct trace_header_t {
...
+        auto block = convert(pc);
+        if (block.first != last_block->first) {
+            last_block = page_table.find(block.first);
+            if (last_block == page_table.end()) {
+                last_block = (page_table.emplace(std::make_pair(
+                                  block.first, std::move(std::bitset<BLOCK_SIZE>()))))
+                                 .first;
+            }
+        }
+        return (last_block != page_table.end() && last_block->second[block.second]);
```
derekbruening, line 761: == 1

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -672,6 +679,103 @@ struct trace_header_t {
...
+    }
+
+    void
+    insert(T pc)
+    {
+        auto block = convert(pc);
+        if (last_block->first != block.first)
+            last_block = page_table.find(block.first);
+        auto val = block.second;
+        last_block->second[block.second] = 1;
```
derekbruening, line 717: This looks like undefined behavior: if last_block is the result of the find() on the unordered_map and it returns end, then it is unsafe to de-reference it.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -672,6 +679,103 @@ struct trace_header_t {
...
+        return true;
+    }
+
+    void
+    erase(T pc)
+    {
+        auto block = convert(pc);
+        if (last_block->first != block.first)
+            last_block = page_table.find(block.first);
+        last_block->second[block.second] = 0;
```
derekbruening, line 746: This again looks like undefined behavior: if last_block is the result of the find() on the unordered_map and it returns end, then it is unsafe to de-reference it.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -672,6 +679,103 @@ struct trace_header_t {
...
+        }
+        return (last_block != page_table.end() && last_block->second[block.second]);
+    }
+
+    void
+    clear()
+    {
+        page_table.clear();
+        page_table.reserve(BASIC_BUCKET_COUNT);
+        page_table.emplace(T(0), std::move(std::bitset<BLOCK_SIZE>()));
```
derekbruening, line 769: Not sure what the purpose of inserting an entry for 0 is?

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -672,6 +679,103 @@ struct trace_header_t {
...
+                                0xFFFF));
+    }
+
+public:
+    bitset_hash_table_t()
+    {
+        static_assert(std::is_pointer<T>::value || std::is_integral<T>::value,
+                      "Pointer or integral type required");
+        page_table.reserve(BASIC_BUCKET_COUNT);
+        page_table.emplace(T(0), std::move(std::bitset<BLOCK_SIZE>()));
```
derekbruening, line 706: Not sure what the purpose of inserting an entry for 0 is?

derekbruening (commented):
Sorry for the delay: finished the review.

clients/drcachesim/tracer/raw2trace.h
kuhanov: ok. 2-4 will be in separate PR

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -672,6 +679,103 @@ struct trace_header_t {
...
     } while (0)
 
+template <typename T> class bitset_hash_table_t {
+private:
+    const static size_t BLOCK_SIZE_BIT = 13;
+    const static size_t BLOCK_SIZE = (1 << BLOCK_SIZE_BIT);
+    const size_t BASIC_BUCKET_COUNT = (1 << 15);
+    std::unordered_map<T, std::bitset<BLOCK_SIZE>> page_table;
+    typename std::unordered_map<T, std::bitset<BLOCK_SIZE>>::iterator last_block =
+        page_table.end();
```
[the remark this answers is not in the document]
kuhanov: ok #6176

clients/drcachesim/tracer/raw2trace.h
kuhanov: done  #6176

clients/drcachesim/tracer/raw2trace.h
kuhanov: done #6176

clients/drcachesim/tracer/raw2trace.h
kuhanov: done #6176

clients/drcachesim/tracer/raw2trace.h
kuhanov: done #6176

clients/drcachesim/tracer/raw2trace.h
kuhanov: Done #6176

clients/drcachesim/tracer/raw2trace_launcher.cpp
```diff
@@ -80,6 +80,15 @@ static droption_t<int>
             "disables concurrency and uses  single thread to perform all operations.  A "
             "negative value sets the job count to the number of hardware threads.");
 
+static droption_t<std::string> op_trace_compress(
```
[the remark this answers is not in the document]
kuhanov: added op_trace_compress

clients/drcachesim/tracer/raw2trace.cpp
kuhanov: There are some cases when vector is null and we could not use reference in this case.
PR #6176

clients/drcachesim/tracer/raw2trace.h
kuhanov: True is return in case of new record only. If we have in table this pc we return false

clients/drcachesim/tracer/raw2trace.h
kuhanov: remove this function. it was not needed at all

clients/drcachesim/tracer/raw2trace.h
kuhanov: removed this function, it was not needed at all

clients/drcachesim/tracer/raw2trace.h
kuhanov: done #6176

clients/drcachesim/tracer/raw2trace.h
kuhanov: just initialize last_block_ iterator to have reference

clients/drcachesim/tracer/raw2trace.h
kuhanov: #6176

clients/drcachesim/tracer/raw2trace.h
kuhanov: #6176

clients/drcachesim/tracer/raw2trace.h
kuhanov: #6176

clients/drcachesim/tracer/raw2trace.h
kuhanov: #6176

clients/drcachesim/tracer/raw2trace.h
derekbruening: Also, shouldn't it be "find_or_insert" and not "find_and_insert" as it would not do both.

clients/drcachesim/common/options.h
```diff
@@ -58,6 +58,12 @@
 #define CACHE_TYPE_UNIFIED "unified"
 #define CACHE_PARENT_MEMORY "memory"
 
+#ifdef HAS_ZIP
+#    define DEFAULT_TRACE_COMPRESSION_TYPE "zip"
+#else
+#    define DEFAULT_TRACE_COMPRESSION_TYPE "gzip"
```
derekbruening, line 64: If zip is not available lz4 could be the default if HAS_LZ4.  But gzip requires HAS_ZLIB right?  And I guess if none of the HAS_ is there the default would have to be "none".

clients/drcachesim/common/options.cpp
```diff
@@ -374,6 +374,18 @@ droption_t<std::string> op_raw_compress(
     "for an SSD, zlib and gzip typically add overhead and would only be used if space is "
     "at a premium; snappy_nocrc and lz4 are nearly always performance wins.");
 
+droption_t<std::string> op_trace_compress(
+    DROPTION_SCOPE_FRONTEND, "compress", DEFAULT_TRACE_COMPRESSION_TYPE,
+    "Trace compression: \"gzip\",\"zlib\",\"lz4\",\"none\"",
```
derekbruening, line 379: s/zlib/zip/.  (Best to put zip first too to match the list on the line below and b/c it is the expected format for full feature support.)

clients/drcachesim/common/options.cpp
```diff
@@ -374,6 +374,18 @@ droption_t<std::string> op_raw_compress(
...
+    "Specifies the compression type to use for trace files: \"zip\", "
+    "\"gzip\", \"lz4\", or \"none\".  When it comes to storage types, the impact "
+    "on overhead varies: "
+    "for SSDs, zip and gzip often increase overhead and should only be chosen "
+    "if space is limited; lz4 compression generally improves performance "
+    "and is recommended in most cases."
```
derekbruening, line 385: nit: " in most cases where fast skipping is not needed".  (So then probabaly should move last sentence up before the "When it comes".)

derekbruening (approved):

PR #6118, opened 2023-06-07
i#6112 runmulti: Remove diagnostic print

derekbruening:
Removes a message line accidentally left in PR #6113.

Issue: #6112

### DIFF
--- a/suite/tests/runmulti.cmake
+++ b/suite/tests/runmulti.cmake
@@ -118,7 +118,6 @@ endmacro()
 process_cmdline(precmd ON ignore)
 
 process_cmdline(cmd OFF tomatch)
-message("output: |${tomatch}|")
 
 if (NOT "${postcmd}" STREQUAL "")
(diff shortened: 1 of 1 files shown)


suite/tests/runmulti.cmake
```diff
@@ -118,7 +118,6 @@ endmacro()
 process_cmdline(precmd ON ignore)
 
 process_cmdline(cmd OFF tomatch)
-message("output: |${tomatch}|")
```
abhinav92003, line 121: During the review of the previous PR, I thought this was intended.  But I see now that there's another message down below that prints it only for mismatch cases.

abhinav92003 (approved):

suite/tests/runmulti.cmake
derekbruening: Right, usually I mark temporary output with a comment so it's easy to find later but failed to do so for this case.

Issue #6120, opened 2023-06-07
Invariant checker should also report the last timestamp and # of instructions since the last timestamp


lihasgupta:
**Is your feature request related to a problem? Please describe.**
Currently, when invariant violations are reported, there is knowledge of the thread id and the type of invariant. It would be useful for many debugging purposes if information about the last known timestamp and the number of instructions since the last known timestamp were also reported.

**Describe the solution you'd like**
I would like this state to be stored in the `invariant_checker_t` data structure and then reported in the `report_if_false()` function call where we currently output invariant violation messages.

PR #6121, opened 2023-06-07
i#6120: Append last timestamp and instrs since last timestamp to invariant checker message

lihasgupta:
This PR appends the invariant_checker error messages to include the value of the last timestamp marker and the number of instructions since the last timestamp. This information is useful when debugging raw traces and trying to find the location of the invariant. 

Fixes: #6120

### DIFF
--- a/clients/drcachesim/tests/invariant_checker_test.cpp
+++ b/clients/drcachesim/tests/invariant_checker_test.cpp
@@ -963,10 +963,11 @@ check_schedule_file()
         std::ifstream serial_read(serial_fname, std::ifstream::binary);
         if (!serial_read)
             return false;
-        if (!run_checker(
-                memrefs, true,
-                { "Serial schedule entry count does not match trace", -1, 0, 0, 0 },
-                "Failed to catch incorrect serial schedule count", &serial_read))
+        if (!run_checker(memrefs, true,
+                         { "Serial schedule entry count does not match trace", /*tid=*/-1,
+                           /*ref_ordinal=*/0, /*last_timestamp=*/0,
+                           /*instrs_since_last_timestamp=*/0 },
+                         "Failed to catch incorrect serial schedule count", &serial_read))
             return false;
     }
     {


clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -79,7 +79,9 @@ invariant_checker_t::report_if_false(per_shard_t *shard, bool condition,
     if (!condition) {
         std::cerr << "Trace invariant failure in T" << shard->tid_ << " at ref # "
                   << shard->stream->get_record_ordinal() << ": " << invariant_name
-                  << "\n";
+                  << "\nlast_timestamp: " << shard->last_timestamp_
+                  << "  instr_count_since_last_timestamp: "
```
derekbruening, line 83: Better to work this into the existing message's location in a human-intended form, instead of printing variable names on a separate line.  Maybe "at ref # nnnn (mmm instrs since last timestamp kkkk): ...".

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -135,8 +137,10 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
     // per-shard counts for error reporting; XXX: we could add our own global
     // counts to compare to the serial stream).
     ++shard->ref_count_;
-    if (type_is_instr(memref.instr.type))
+    if (type_is_instr(memref.instr.type)) {
         ++shard->instr_count_;
+        ++shard->instr_count_since_last_timestamp_;
```
derekbruening, line 142: (Storing the count as of the last timestamp is more efficient, but an extra increment shouldn't matter so let's leave it.)

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -78,8 +78,10 @@ invariant_checker_t::report_if_false(per_shard_t *shard, bool condition,
 {
     if (!condition) {
         std::cerr << "Trace invariant failure in T" << shard->tid_ << " at ref # "
-                  << shard->stream->get_record_ordinal() << ": " << invariant_name
-                  << "\n";
+                  << shard->stream->get_record_ordinal() << " ("
+                  << shard->instr_count_since_last_timestamp_
+                  << " instrs since timestamp " << shard->last_timestamp_
+                  << "): " << invariant_name << "\n";
```
derekbruening, line 84: Let's add a simple test that these new fields are correct: we can easily have a checker unit test verify with a scenario or two of records with timestamps.  Then we'll catch any regression from future code changes right up front in our suite.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -58,9 +58,14 @@ class checker_no_abort_t : public invariant_checker_t {
     {
     }
 
-    std::vector<std::string> errors;
-    std::vector<memref_tid_t> error_tids;
-    std::vector<uint64_t> error_refs; // Memref count (ordinal) at error point.
+    struct error {
```
derekbruening, line 61: style violation: all types end in _t: e.g., `error_info_t`.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -58,9 +58,14 @@ class checker_no_abort_t : public invariant_checker_t {
...
+        std::string invariant_name;
+        memref_tid_t tid;
+        uint64_t refs; // Memref count (ordinal) at error point.
```
derekbruening, line 64: How about renaming to "ref_ordinal" and then no comment is needed.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -89,6 +96,7 @@ class checker_no_abort_t : public invariant_checker_t {
 bool
 run_checker(const std::vector<memref_t> &memrefs, bool expect_error,
             memref_tid_t error_tid = 0, uint64_t error_refs = 0,
+            uint64_t error_last_timestamp = 0, uint64_t error_instr_count = 0,
```
derekbruening, line 99: Let's use the struct here too instead of 5 separate parameters.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -101,16 +109,19 @@ run_checker(const std::vector<memref_t> &memrefs, bool expect_error,
...
         if (expect_error) {
-            if (checker.errors.size() != 1 || checker.errors[0] != expected_message ||
-                checker.error_tids[0] != error_tid ||
-                checker.error_refs[0] != error_refs) {
+            if (checker.errors.size() != 1 ||
+                checker.errors[0].invariant_name != expected_message ||
+                checker.errors[0].tid != error_tid ||
+                checker.errors[0].refs != error_refs ||
+                checker.errors[0].last_timestamp != error_last_timestamp ||
+                checker.errors[0].instrs_since_last_timestamp != error_instr_count) {
```
derekbruening, line 117: Then here the comparison of the two structs could be done by operator==.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -178,12 +192,12 @@ check_branch_target_after_branch()
         std::vector<memref_t> memrefs = {
             gen_instr(1, 1),
             gen_branch(1, 2),
-            gen_marker(2, TRACE_MARKER_TYPE_TIMESTAMP, 0),
+            gen_marker(2, TRACE_MARKER_TYPE_TIMESTAMP, 3),
             gen_instr(2, 1),
-            gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 0),
+            gen_marker(1, TRACE_MARKER_TYPE_TIMESTAMP, 3),
```
derekbruening, line 197: Use a named constant rather than repeating "3" 3 times.

Ditto below.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -893,9 +963,10 @@ check_schedule_file()
         std::ifstream serial_read(serial_fname, std::ifstream::binary);
         if (!serial_read)
             return false;
-        if (!run_checker(memrefs, true, -1, 0,
-                         "Serial schedule entry count does not match trace",
-                         "Failed to catch incorrect serial schedule count", &serial_read))
+        if (!run_checker(
+                memrefs, true,
+                { "Serial schedule entry count does not match trace", -1, 0, 0, 0 },
```
derekbruening, line 968: A `/*tid=*/` for the -1 may help readability.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -77,19 +91,20 @@ class checker_no_abort_t : public invariant_checker_t {
...
         if (!condition) {
             std::cerr << "Recording |" << invariant_name << "| in T" << shard->tid_
-                      << " @" << shard->ref_count_ << "\n";
-            errors.push_back(invariant_name);
-            error_tids.push_back(shard->tid_);
-            error_refs.push_back(shard->ref_count_);
+                      << " @ ref # " << shard->ref_count_ << " ("
+                      << shard->instr_count_since_last_timestamp_
+                      << " instrs since timestamp " << shard->last_timestamp_ << ")\n";
+            errors.push_back({ invariant_name, shard->tid_, shard->ref_count_,
```
derekbruening, line 97: optional: emplace_back with a constructor is more efficient

PR #6125, opened 2023-06-08
Remove checked-in ninja.exe

derekbruening:
A follow-up to PR #6123, we remove the checked-in ninja.exe as developers should install a recent ninja on their own.  Its presence also complicates license checkers as it has a different license even though it is not incorporated into our deliverables.


derekbruening:
(I thought I had included this in PR #6123 eventually; turns out it wasn't there)

abhinav92003 (approved):

PR #6132, opened 2023-06-09
i#5843 scheduler: Record futex parameters and maybe-blocking syscalls

derekbruening:
Adds markers that help with scheduling and interleaving our per-software-thread user-mode traces, to identify inter-thread dependencies and kernel wait points: "maybe-blocking" plus kernel locks on Linux (SYS_futex), as further described below:

Adds tracing of SYS_futex's parameter values on Linux.  This uses the function tracing markers, with an ID equal to a new system call base value plus the system call number so we can add other system calls in the future.  For the return value, dr_syscall_result_info_t.succeeded is used to avoid complexities on some platforms (although it has its own downsides).

Adds a new marker TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL and sets it in raw2trace for Linux using a preliminary list of numbers which might block.  Other operating systems are not yet supported.

Adds a new filetype OFFLINE_FILE_TYPE_BLOCKING_SYSCALLS which indicates whether the futex parameters and blocking syscall markers are present.

Adds checks to the invariant_checker tool.

Updates test outputs that count markers.

Adds a new burst_futex test.  The trace markers look like this:
```
        1269         890:       26493 ifetch       2 byte(s) @ 0x00007fdb221005a7 0f 05                syscall
        1270         890:       26493 <marker: system call 202>
        1271         890:       26493 <marker: maybe-blocking sytem call>
        1272         890:       26493 <marker: function==syscall #202>
        1273         890:       26493 <marker: function argument 0x55c677a7de38>
        1274         890:       26493 <marker: function argument 0x1>
        1275         890:       26493 <marker: function argument 0x1>
        1276         890:       26493 <marker: function argument 0x0>
        1277         890:       26493 <marker: function argument 0x0>
        1278         890:       26493 <marker: function argument 0x0>
        1279         890:       26493 <marker: timestamp 13330804804062625>
        1280         890:       26493 <marker: tid 26493 on core 1>
        1281         890:       26493 <marker: function==syscall #202>
        1282         890:       26493 <marker: function return value 0x1>
        1283         891:       26493 ifetch       6 byte(s) @ 0x00007fdb221005a9 48 3d 01 f0 ff ff    cmp    %rax, $0xfffff001
```

Issue: #5843


clients/drcachesim/common/trace_entry.h
```diff
@@ -272,9 +272,17 @@ typedef enum {
...
+     * The marker value contains the function id for functions traced by the user via
+     * the -record_function (and -record_heap_value if -record_heap is specified)
+     * option.  The mapping from this numeric ID to a library-qualified symbolic name
+     * is recorded during tracing in a file "funclist.log" whose format is described by
+     * the drmemtrace_get_funclist_path() function's documentation.
+     *
+     * This marker is also used to record parameter values for certain system calls such
+     * as for #OFFLINE_FILE_TYPE_KERNEL_SCHED.  These use large identifiers equal to
+     * #TRACE_FUNC_ID_SYSCALL_BASE plus the system call number (or its bottom 16 bits
+     * for 32-bit marker values).  These identifiers are not stored in the function
```
abhinav92003, line 284: Does this mean that for 32-bit marker values its simply bottom 16 bits without consideration to TRACE_FUNC_ID_SYSCALL_BASE? If so, let's move this part out of parenthesis. If not, maybe it is simpler to just say  TRACE_FUNC_ID_SYSCALL_BASE here and not the bits at all (since the user should rely on TRACE_FUNC_ID_SYSCALL_BASE only).

clients/drcachesim/common/trace_entry.h
```diff
@@ -272,9 +272,17 @@ typedef enum {
...
-     * -record_function (and -record_heap_value if -record_heap is specified)
-     * option.
+     * The marker value contains the function id for functions traced by the user via
+     * the -record_function (and -record_heap_value if -record_heap is specified)
+     * option.  The mapping from this numeric ID to a library-qualified symbolic name
+     * is recorded during tracing in a file "funclist.log" whose format is described by
+     * the drmemtrace_get_funclist_path() function's documentation.
+     *
+     * This marker is also used to record parameter values for certain system calls such
+     * as for #OFFLINE_FILE_TYPE_KERNEL_SCHED.  These use large identifiers equal to
```
abhinav92003, line 282: Do we need to provide more details about these system call identifiers? Normalized vs not normalized etc.  That they are present only for LINUX.

clients/drcachesim/common/trace_entry.h
```diff
@@ -300,6 +308,9 @@ typedef enum {
      * The marker value contains the return value of the just-entered function,
      * whose id is specified by the closest previous #TRACE_MARKER_TYPE_FUNC_ID
      * marker entry
+     *
+     * The marker value for system calls (see #TRACE_FUNC_ID_SYSCALL_BASE) is either 0
+     * (failure) or 1 (success).
```
abhinav92003, line 313: Add the caveats to this "success" code here (i#5843).

clients/drcachesim/common/trace_entry.h
```diff
@@ -442,13 +453,38 @@ typedef enum {
      */
     TRACE_MARKER_TYPE_SYSCALL,
 
+    /**
+     * This marker is emitted prior to a system call which is known to block under some
+     * circumstances.  Whether it blocks may depend on the values of parameters or the
```
abhinav92003, line 458: So absence of the marker means it surely doesn't block?

Edit:
From is_maybe_blocking_syscall, it looks like it may still block. Add a one-line caveat here?

clients/drcachesim/common/trace_entry.h
```diff
@@ -442,13 +453,38 @@ typedef enum {
...
 
+/** Constants related to function or system call parameter tracing. */
+enum {
+/**
+ * When system call parameter and return values are provided, they use the function
+ * tracing markers #TRACE_MARKER_TYPE_FUNC_ID, #TRACE_MARKER_TYPE_FUNC_ARG, and
+ * #TRACE_MARKER_TYPE_FUNC_RETVAL.  The identifier used for
+ * #TRACE_MARKER_TYPE_FUNC_ID is equal to this base value plus the 32-bit system
+ * call number for 64-bit marker values or this base value plus the lower 16 bits
+ * of the system call number for 32-bit marker values.
```
abhinav92003, line 479: "lower 16 bits of the system call number": are syscall numbers larger than that at all?

clients/drcachesim/common/trace_entry.h
```diff
@@ -663,8 +699,22 @@ typedef enum {
...
+     * markers #TRACE_MARKER_TYPE_FUNC_ID, #TRACE_MARKER_TYPE_FUNC_ARG, and
+     * #TRACE_MARKER_TYPE_FUNC_RETVAL with an identifier equal to
+     * #TRACE_FUNC_ID_SYSCALL_BASE plus the system call number (or its bottom 16 bits
+     * for 32-bit marker values).  These identifiers are not stored in the function
+     * list file (drmemtrace_get_funclist_path()).
+     *
+     * The #TRACE_MARKER_TYPE_FUNC_RETVAL for system calls is either 0 (failure) or
+     * 1 (success).
+     */
+    OFFLINE_FILE_TYPE_KERNEL_SCHED = 0x800,
```
abhinav92003, line 717: We may identify more data required for scheduling in future. So better to use a name that describes the actual data added to the trace, not what that data enables; something like `OFFLINE_FILE_TYPE_BLOCKING_SYSCALL_AND_FUTEX` or just `OFFLINE_FILE_TYPE_BLOCKING_SYSCALL`

clients/drcachesim/tests/burst_futex.cpp
```diff
@@ -0,0 +1,196 @@
...
+            ++arg_ord;
+        }
+        if (memref.marker.marker_type == TRACE_MARKER_TYPE_FUNC_RETVAL) {
+            // Should have succeeded.
+            assert(memref.marker.marker_value == 1);
+        }
+    }
+    assert(saw_maybe_blocking);
+    static constexpr int FUTEX_ARG_COUNT = 6;
+    assert(arg_ord == FUTEX_ARG_COUNT);
```
abhinav92003, line 191: (This assumes only one futex call in the trace. Okay I guess.)

abhinav92003, line 191: This would fail as intended if the trace didn't have any func marker for futex, but may be better to add an explicit assert for that anyway.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -322,6 +322,16 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                                 TRACE_MARKER_TYPE_SYSCALL,
                         "Syscall instruction not followed by syscall marker");
     }
+    if (memref.marker.type == TRACE_TYPE_MARKER &&
+        memref.marker.marker_type == TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL) {
+        shard->found_blocking_marker_ = true;
+        report_if_false(shard,
+                        shard->found_syscall_marker_ &&
```
abhinav92003, line 329: `shard->found_syscall_marker_ &&` can be omitted I think, since that would be set if the second condition is true.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -355,9 +365,17 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+    if (memref.marker.type == TRACE_TYPE_MARKER &&
+        memref.marker.marker_type == TRACE_MARKER_TYPE_FUNC_ID) {
+        shard->prev_func_id_ = memref.marker.marker_value;
+    }
     if (memref.marker.type == TRACE_TYPE_MARKER &&
         marker_type_is_function_marker(memref.marker.marker_type)) {
-        report_if_false(shard, type_is_instr_branch(shard->prev_instr_.instr.type),
+        report_if_false(shard,
+                        shard->prev_func_id_ >= TRACE_FUNC_ID_SYSCALL_BASE ||
+                            (memref.marker.marker_type == TRACE_MARKER_TYPE_FUNC_ID &&
```
abhinav92003, line 376: Why do we need this second part of the condition? Isn't the check on `prev_func_id_ >= TRACE_FUNC_ID_SYSCALL_BASE` enough?

If the second part is true, prev_func_id_ would be set to a value >= TRACE_FUNC_ID_SYSCALL_BASE by the if-block on L368, causing the first part to be true.

abhinav92003: Or: Maybe it is better to remove prev_func_id_ or make it a local var instead of a shard_data var.

abhinav92003 (approved):

clients/drcachesim/common/trace_entry.h
derekbruening: It wouldn't really make sense to ignore the base and have just the bottom 16 bits as it just said it's a large identifier.  But I reworded to "#TRACE_FUNC_ID_SYSCALL_BASE plus the system call number (for 32-bit marker values
     * just the bottom 16 bits of the system call number are added to the base)."

clients/drcachesim/common/trace_entry.h
derekbruening: Added a sentence pointing at  dr_register_pre_syscall_event()'s docs for normalized numbers for the source of the number used.

derekbruening: For only Linux: it points at the KERNEL_SCHED which lists only futex so I think that's enough.

clients/drcachesim/common/trace_entry.h
derekbruening: Added links to the DR docs about this.

clients/drcachesim/common/trace_entry.h
derekbruening: Yes, in multiple OS variants upper bits are set to indicate various things.  DR removes some Mac upper bits but there are others -- e.g. see some of the WOW64 numbers: https://github.com/DynamoRIO/drmemory/blob/master/drsyscall/drsyscall_numx.h#L2115

clients/drcachesim/common/trace_entry.h
derekbruening: Added caveats about not being guaranteed to be complete.

clients/drcachesim/common/trace_entry.h
derekbruening: Hmm, we could do KERNEL_SCHED_2.  I liked KERNEL_SCHED tying these together.  But maybe lumping futex into BLOCKING_SYSCALLS is ok.  All right, will go with BLOCKING_SYSCALLS.

clients/drcachesim/tests/burst_futex.cpp
derekbruening: If glibc startup adds a futex call we'll see the failure so it won't be silent.

derekbruening: Added a comment.

derekbruening: Done.

clients/drcachesim/tools/invariant_checker.cpp
derekbruening: Done.

clients/drcachesim/tools/invariant_checker.cpp
derekbruening: This was added in f3c36937c7c4f886540db068e5f752bc2de05cdf but it also removed the else on this if and I think that by itself may be enough.  I removed the 2nd condition.

We cannot remove prev_func_id_ nor make it a local var because there are *multiple* markers for the retaddr and args after the ID.

derekbruening:
Not understanding why 4 jobs were cancelled: no reason given.  No email like is sent from cancelling due to a subsequent test for the same PR.  ??  I guess just request a re-run and see if this keeps happening -- seems to have happened on PR #6137 too.

Issue #6131, opened 2023-06-09
HANG drcacheoff.gencode and drcacheoff.burst_replace on 32-bit Windows


derekbruening:
First reported at https://github.com/DynamoRIO/dynamorio/issues/2204#issuecomment-1583885867:
win32 tool.drcacheoff.gencode failed 3x in a row with a timeout each time: https://github.com/DynamoRIO/dynamorio/actions/runs/5217398902/jobs/9417192797

https://github.com/DynamoRIO/dynamorio/issues/2204#issuecomment-1583971307
win32 tool.drcacheoff.gencode and tool.drcacheoff.burst_replace failed due to timeout again: https://github.com/DynamoRIO/dynamorio/actions/runs/5217720624/jobs/9417832232#step:6:19819

derekbruening:
This seems to have first happened in the post-merge tests for PR #6096 and since then it has happened quite a bit.

ksco:
I've tried multi times, these timeouts seem pretty "stable", so basically prevent new PRs get merged..

derekbruening:
@ksco have you done any debugging so far, like tmate on the GA runner?

ksco:
No, I haven't done any debugging on this yet.

derekbruening:
One theory is that it's a pop-up from a VC++ assert failure -- I see we don't have the `_CrtSetReportMode` in these static-DR apps that we have in all our regular tests.  We should add that.  As for why it seems to only show up on merge-queue runs (PR #6130 passed) or master-merge runs: no idea; in the past a GA CI image upgrade might trigger such things.

derekbruening:
OK I will volunteer on this one -- maybe someone else can get the next one.

Note that this again succeeded in the regular PR run in my new PR #6132.  But it will probably fail in the merge queue.

I tried to run the gencode test in 32-bit in an older Windows machine and it indeed hits a VC++ assert with a pop-up:

```
---------------------------
Microsoft Visual C++ Runtime Library
---------------------------
Debug Error!

Program: ...ild_x86_dbg_tests\suite\tests\bin\tool.drcacheoff.gencode.exe

abort() has been called

(Press Retry to debug the application)

---------------------------
Abort   Retry   Ignore   
---------------------------
```

Console output is (buffered by runmulti so not surprising we don't see this in the runner output):
```
pre-DR init
pre-DR start
pre-DR detach
all done
raw2trace failed: Failed to process file for thread 6688: memref entry found outside of bb
[...]

PR #6133, opened 2023-06-09
i#6131 hang win32: Disable popups in drcachesim tests

derekbruening:
Adds a library test_helpers for drcachesim tests which provides main() and calls disable_popups() to avoid Windows assert and exception pop-up windows hanging the test suite.  Tests now define a test_main() which is called by the test_helpers main().  (The DR core test suite has its own popup disabling code but it is complex to link that into
the standalone drcachesim tests.)

For _tmain, we compile test_helpers.cpp separately to access its disable_popups() but with a define to disable main().

Issue: #6131

### DIFF
--- a/clients/drcachesim/CMakeLists.txt
+++ b/clients/drcachesim/CMakeLists.txt
@@ -319,6 +319,8 @@ install_client_nonDR_header(drmemtrace tracer/raw2trace.h)
 install_client_nonDR_header(drmemtrace scheduler/scheduler.h)
 install_client_nonDR_header(drmemtrace scheduler/speculator.h)
 
+add_library(test_helpers STATIC tests/test_helpers.cpp)
+
 # We show one example of how to create a standalone analyzer of trace
 # files that does not need to link with DR.
 add_executable(histogram_launcher
@@ -343,7 +345,7 @@ target_link_libraries(record_filter_launcher drmemtrace_analyzer drmemtrace_reco
 use_DynamoRIO_extension(record_filter_launcher droption)
 
 add_executable(scheduler_launcher tests/scheduler_launcher.cpp)
-target_link_libraries(scheduler_launcher drmemtrace_analyzer drfrontendlib)
+target_link_libraries(scheduler_launcher drmemtrace_analyzer drfrontendlib test_helpers)
 use_DynamoRIO_extension(scheduler_launcher droption)
 add_dependencies(scheduler_launcher api_headers)
 # Simple test to ensure this launcher keeps working.
@@ -515,6 +517,9 @@ restore_nonclient_flags(drmemtrace_invariant_checker)
 # get rid of add_win32_flags(target) since that just adds a couple of flags
 # back, and using the ORIG_ values is preferable.
 macro(add_win32_flags target)
+  if (DEBUG)
+    append_property_list(TARGET ${target} COMPILE_DEFINITIONS "DEBUG")
+  endif ()
   if (WIN32)
     if (DEBUG)
       get_property(cur TARGET ${target} PROPERTY COMPILE_FLAGS)
@@ -671,7 +676,7 @@ endif ()
 if (BUILD_TESTS)
   add_executable(tool.reuse_distance.unit_tests tests/reuse_distance_test.cpp)
   target_link_libraries(tool.reuse_distance.unit_tests drmemtrace_reuse_distance
-    drmemtrace_static)
+    drmemtrace_static test_helpers)
   add_win32_flags(tool.reuse_distance.unit_tests)
   add_test(NAME tool.reuse_distance.unit_tests
        COMMAND tool.reuse_distance.unit_tests)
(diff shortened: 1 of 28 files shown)


derekbruening:
We'll see what this one does in the merge queue.  All the burst_ tests passed in the in-PR test run here.  I didn't actually fix any assert here but hopefully it should now be visibile and it won't take an hour.

clients/drcachesim/tests/view_test.cpp
```diff
@@ -590,6 +591,8 @@ run_chunk_tests(void *drcontext)
 int
 main(int argc, const char *argv[])
 {
+    disable_popups();
```
abhinav92003, line 594: Would it be better that tests implement a `test_main()` that is invoked by `main()` defined inside test_helpers.h? This way tests won't forget to call `disable_popups()` or other test setup code we may have in future.

abhinav92003 (approved):

clients/drcachesim/CMakeLists.txt
```diff
@@ -822,30 +832,36 @@ if (BUILD_TESTS)
...
       add_executable(tool.drcacheoff.burst_noreach tests/burst_noreach.cpp)
       configure_DynamoRIO_static(tool.drcacheoff.burst_noreach)
       use_DynamoRIO_static_client(tool.drcacheoff.burst_noreach drmemtrace_static)
+      target_link_libraries(tool.drcacheoff.burst_noreach test_helpers)
       add_win32_flags(tool.drcacheoff.burst_noreach)
     endif ()
     if (LINUX) # Uses mremap.
-      add_executable(tool.drcacheoff.burst_maps tests/burst_maps.cpp)
+      add_executable(tool.drcacheoff.burst_maps tests/burst_maps.cpp
+        tests/test_helpers.cpp)
```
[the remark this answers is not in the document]
derekbruening: No, that's a mistake.  Removed.

clients/drcachesim/tests/test_helpers.cpp
```diff
@@ -0,0 +1,43 @@
+/* **********************************************************
+ * Copyright (c) 2017-2023 Google, Inc.  All rights reserved.
```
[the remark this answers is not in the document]
derekbruening: Done.

clients/drcachesim/tests/view_test.cpp
derekbruening: You mean inside test_helpers.cpp?  And then we don't even need a header file test_helpers.h at all?  OK.

derekbruening: A test could still skip setup if they defined their own main().  We could keep the header and have it #define main to something else -- but a test might not include the header.

clients/drcachesim/CMakeLists.txt
```diff
@@ -515,6 +517,9 @@ restore_nonclient_flags(drmemtrace_invariant_checker)
 # get rid of add_win32_flags(target) since that just adds a couple of flags
 # back, and using the ORIG_ values is preferable.
 macro(add_win32_flags target)
+  if (DEBUG)
+    append_property_list(TARGET ${target} COMPILE_DEFINITIONS "DEBUG")
```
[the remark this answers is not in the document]
derekbruening: Part of the flags being clobbered for DR clients and CMake having global or per-file properties.  restore_nonclient_flags() puts them back but there was some reason we couldn't use it for clients and these burst_ tests all have clients in them....this could all be revisited and tried to make cleaner since modern CMake might have better control but that is out of scope for this PR.

clients/drcachesim/tests/view_test.cpp
abhinav92003: > but a test might not include the header.

Right, but it's easier to instruct tests to use test_main, than a possibly complex sequence of setup. I guess for now it's fine because it's just one disable_popups call.

derekbruening: For suite/tests every test calls OS_INIT in main but I agree it is easier to forget that.  I will make the test_main change.

derekbruening: OK I found a flaw: schedule_launcher uses _tmain.

derekbruening: I'm using test_helpers.h to declare disable_popups() for use with _tmain and everything else uses test_main().

derekbruening: That doesn't work though with test_helpers as a library as it adds a main symbol...and to add a test_tmain requires frontendlib.

derekbruening:
Both the gencode and replace tests are *still* timing out after 10 minutes on the merge run!!  This change is still a good one so I think I should just direct-merge it.  Possible next steps:

+ Lower the timeout from 10 mins for the dr$sim tests
+ Can we detect a merge queue run and disable a test *only* in that run?  If so, disable these 2 tests there for 32-bit win.  If not, disable them for 32-bit win for all but master merges?
+ Try to use tmate with the merge queue: sounds tricky for sthg that only shows up on the queue or I guess also on master merges and not on PR runs
+ Try more local debugging on different machines

PR #6134, opened 2023-06-09
i#6020: Add raw2trace stat for trace duration

abhinav92003:
Adds raw2trace stats for earliest and latest trace timestamp. These are used to compute the trace's wall-clock duration which is printed at the end of raw2trace decomposition, along with other raw2trace stats.

It may be useful for the user to know the trace's wall-clock duration to compute the value for `-trace_interval` (which is the length of each trace interval for interval analysis) to get the required number of total intervals.

Issue: #6020

### DIFF
--- a/clients/drcachesim/tracer/raw2trace.cpp
+++ b/clients/drcachesim/tracer/raw2trace.cpp
@@ -820,10 +820,10 @@ raw2trace_t::process_next_thread_buffer(raw2trace_thread_data_t *tdata,
         if (entry.timestamp.type == OFFLINE_TYPE_TIMESTAMP) {
             VPRINT(2, "Thread %u timestamp 0x" ZHEX64_FORMAT_STRING "\n",
                    (uint)tdata->tid, (uint64)entry.timestamp.usec);
-            add_to_statistic(tdata, RAW2TRACE_STAT_EARLIEST_TRACE_TIMESTAMP,
-                             (uint64)entry.timestamp.usec);
-            add_to_statistic(tdata, RAW2TRACE_STAT_LATEST_TRACE_TIMESTAMP,
-                             (uint64)entry.timestamp.usec);
+            accumulate_to_statistic(tdata, RAW2TRACE_STAT_EARLIEST_TRACE_TIMESTAMP,
+                                    static_cast<uint64>(entry.timestamp.usec));
+            accumulate_to_statistic(tdata, RAW2TRACE_STAT_LATEST_TRACE_TIMESTAMP,
+                                    static_cast<uint64>(entry.timestamp.usec));
             byte *buf = buf_base +
                 trace_metadata_writer_t::write_timestamp(buf_base,
                                                          (uintptr_t)entry.timestamp.usec);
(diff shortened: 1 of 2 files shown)


clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -2979,14 +2993,21 @@ drmemtrace_get_timestamp_from_offline_trace(const void *trace, size_t trace_size
...
 {
     switch (stat) {
     case RAW2TRACE_STAT_COUNT_ELIDED: tdata->count_elided += value; break;
     case RAW2TRACE_STAT_DUPLICATE_SYSCALL: tdata->count_duplicate_syscall += value; break;
     case RAW2TRACE_STAT_FALSE_SYSCALL: tdata->count_false_syscall += value; break;
     case RAW2TRACE_STAT_RSEQ_ABORT: tdata->count_rseq_abort += value; break;
     case RAW2TRACE_STAT_RSEQ_SIDE_EXIT: tdata->count_rseq_side_exit += value; break;
+    case RAW2TRACE_STAT_EARLIEST_TRACE_TIMESTAMP:
+        tdata->earliest_trace_timestamp =
+            std::min(tdata->earliest_trace_timestamp, value);
```
derekbruening, line 3006: I found it confusing to see the call above which says "add": can we rename this function to "merge_statistic" or something along those lines -- I think that will make it easier to understand when seeing just the calls at line 823, 823.

derekbruening (approved):

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -820,6 +820,10 @@ raw2trace_t::process_next_thread_buffer(raw2trace_thread_data_t *tdata,
         if (entry.timestamp.type == OFFLINE_TYPE_TIMESTAMP) {
             VPRINT(2, "Thread %u timestamp 0x" ZHEX64_FORMAT_STRING "\n",
                    (uint)tdata->tid, (uint64)entry.timestamp.usec);
+            add_to_statistic(tdata, RAW2TRACE_STAT_EARLIEST_TRACE_TIMESTAMP,
+                             (uint64)entry.timestamp.usec);
```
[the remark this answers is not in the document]
abhinav92003: Done.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1246,7 +1253,7 @@ class raw2trace_t {
     // Increases the per-thread counter for the statistic identified by stat by value.
```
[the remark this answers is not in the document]
abhinav92003: Done.

clients/drcachesim/tracer/raw2trace.cpp
abhinav92003: Done. Used "accumulate" instead of "merge"

derekbruening: Hmm, "accumulate" sounds like "add" to me: keeping just the max or min doesn't fit with "accumulate" IMHO.

PR #6137, opened 2023-06-12
i#6131 hang win32: Add test timeouts to dr$sim and runcmp tests

derekbruening:
The dr$sim unit tests, tests using runcmp (such as nearly all
drcacheoff.* tests), and native tests had no drrun or runstats or
ctest timeout set, leaving them the too-long default 10 minutes,
which combined with the try-3-times results in 30 minutes for a
hung test.  We set the 90s timeout here for these tests.

Issue: #6131

### DIFF
--- a/clients/drcachesim/tests/burst_gencode.cpp
+++ b/clients/drcachesim/tests/burst_gencode.cpp
@@ -381,5 +381,7 @@ int
 test_main(int argc, const char *argv[])
 {
     std::string trace_dir = gather_trace();
+    while (argc < 10)
+        ;
     return look_for_gencode(trace_dir);
 }
--- a/clients/drcachesim/tests/burst_gencode.cpp
+++ b/clients/drcachesim/tests/burst_gencode.cpp
@@ -381,7 +381,5 @@ int
 test_main(int argc, const char *argv[])
 {
     std::string trace_dir = gather_trace();
-    while (argc < 10)
-        ;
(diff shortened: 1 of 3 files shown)


clients/drcachesim/tests/burst_gencode.cpp
```diff
@@ -381,5 +381,7 @@ int
 test_main(int argc, const char *argv[])
 {
     std::string trace_dir = gather_trace();
+    while (argc < 10)
```
abhinav92003, line 384: Seems like this was meant to be removed?

abhinav92003 (approved):

derekbruening:
Hit a couple of tests which seem to need more than the 90s in some cases on windows.  The max-global I wonder if it's something wrong in the test where it never reaches the global like in #5366 or #4711 but no time now to investigate.

Testing merge queue again, I want to see the replace and gencode tests fail with 90s timeouts and have the failure be ignored.  90s 3x is 4.5 mins which should be much faster than the previous 30 minutes each.

derekbruening:
Huh, this passed the merge queue tests.  Looking at the vs2019-32 result: it never ran the gencode or burst_replace tests.  Looking at what we disable from PR runs vs master merge: those are on the list.  So maybe the sudden appearance of failures was b/c the merge queue logic didn't say it was a pull request, and now they've fixed that?!?  And the gencode failure for 32-bit Windows was a real regression?

PR #6147, opened 2023-06-15
i#3544 fix: risc-v TCB offset from tp register is negative according to psABI

vflouris:
According to RISCV processor specific application binary interface, psABI : https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc "tp containing the address one past the end of the TCB." further documentation for that can be found in glibc sources sysdeps/riscv/nptl/tls.h:
/* Return the address of the dtv for the current thread.  */
  (((tcbhead_t *) (READ_THREAD_POINTER () - TLS_TCB_OFFSET))[-1].dtv)

### DIFF
--- a/core/unix/os_exports.h
+++ b/core/unix/os_exports.h
@@ -193,14 +193,15 @@ extern uint android_tls_base_offs;
 #endif
 
 #ifdef RISCV64
-/* FIXME i#3544: We might need to re-use ARM's approach and store DR TLS in
- * tcb_head_t::private field: typedef struct
+/* Re-using ARM's approach and store DR TLS in tcb_head_t::private,
+ * with the only difference being tp register points at the end of TCB
+ * typedef struct
  *   {
  *     dtv_t *dtv;
  *     void *private;
  *   } tcb_head_t;
  */
-#    define DR_TLS_BASE_OFFSET IF_X64_ELSE(-8, -4) /* skip dtv */
+#    define DR_TLS_BASE_OFFSET IF_X64_ELSE(-8, -4) /* tcb->private, skip dtv */
 #endif
 
 #ifdef LINUX
--- a/core/unix/os_exports.h
+++ b/core/unix/os_exports.h
@@ -200,7 +200,7 @@ extern uint android_tls_base_offs;
  *     void *private;
  *   } tcb_head_t;
  */
-#    define DR_TLS_BASE_OFFSET IF_X64_ELSE(8, 4) /* skip dtv */
+#    define DR_TLS_BASE_OFFSET IF_X64_ELSE(-8, -4) /* skip dtv */
 #endif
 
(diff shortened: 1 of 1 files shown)


ksco:
I think this fix is right. 

As the PR's desc said, RISC-V, same as ARM, uses Variant I (below) layout of the thread-local storage, but `tp` is set up to point to the end of the TCB instead of the beginning. So it's only similar to Variant I, not strictly it.

`tcbhead_t` is defined as:

```
typedef struct
{
  dtv_t *dtv;
  void *private;
} tcbhead_t;
```

So, to get the address of `private`, we need to subtract `8` for RISC-V and add `8` for Aarch64.

image

core/unix/os_exports.h
```diff
@@ -200,7 +200,7 @@ extern uint android_tls_base_offs;
  *     void *private;
  *   } tcb_head_t;
  */
```
ksco, line 202: The comments should also be updated.

vflouris: I thought about it, but it still, sort of, skips dtv by jumping only 4/8 bytes, instead of the beginning of the TCB (8/16). Perhaps something like /* point to tcb->private, skipping dtv */ would be more appropriate?

ksco: Yeah I think it’s fine. But what I mean is the comment above:

```
/* FIXME i#3544: We might need to re-use ARM's approach and store DR TLS in
 * tcb_head_t::private field: typedef struct
 *   {
 *     dtv_t *dtv;
  *     void *private;
  *   } tcb_head_t;
  */
```

vflouris: How about:
/* Re-using ARM's approach and store DR TLS in tcb_head_t::private, 
* with the only difference being tp register points at the end of TCB
* typedef struct
*   {
 *     dtv_t *dtv;
  *     void *private;
  *   } tcb_head_t;
  */

ksco: I'm fine with it.

ksco (approved):

core/unix/os_exports.h
```diff
@@ -193,14 +193,15 @@ extern uint android_tls_base_offs;
 #endif
 
 #ifdef RISCV64
-/* FIXME i#3544: We might need to re-use ARM's approach and store DR TLS in
- * tcb_head_t::private field: typedef struct
+/* Re-using ARM's approach and store DR TLS in tcb_head_t::private,
+ * with the only difference being tp register points at the end of TCB
```
derekbruening, line 197: nit: Add a period to end this sentence (and maybe also add a blank line after it -- right now it just runs right into the typedef which without any separation at all which is confusing)

derekbruening (approved):

PR #6163, opened 2023-06-21
i#5843 scheduler: Switch on a blocking syscall

derekbruening:
Adds an input switch on a blocking syscall, without considering the current input even if it is the highest priority.

Refactors pick_next_input() to use a helper function pick_next_input_as_previously() to make it easier to read the new wait state code.

Adds a unit test with blocking syscalls and priorities to ensure lower-priority inputs are scheduled when a higher-priority input goes into a wait state.

Regenerates the checked-in threadsig.x64 trace to contain syscall number and maybe-blocking markers and updates the associated tests' thread ids and expected output.

Issue: #5843


derekbruening:
Hit scattergather on x86-32; re-running.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -1555,11 +1560,14 @@ scheduler_tmpl_t<RecordType, ReaderType>::pick_next_input(output_ordinal_t outpu
...
-                    // shouldn't switch.  The queue preserves FIFO for same-priority
-                    // cases so we will switch if someone of equal priority is waiting.
-                    set_cur_input(output, INVALID_INPUT_ORDINAL);
+                    if (!in_wait_state) {
+                        // Give up the input before we go to the queue so we can add
+                        // ourselves to the queue.  If we're the highest priority we
+                        // shouldn't switch.  The queue preserves FIFO for same-priority
+                        // cases so we will switch if someone of equal priority is
+                        // waiting.
+                        set_cur_input(output, INVALID_INPUT_ORDINAL);
```
abhinav92003, line 1569: What if the only thread that's left is the thread that executed the blocking system call?

clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -1126,6 +1126,93 @@ test_synthetic_with_bindings_weighted()
...
+    if (scheduler.init(sched_inputs, NUM_OUTPUTS, sched_ops) !=
+        scheduler_t::STATUS_SUCCESS)
+        assert(false);
+    std::vector<std::string> sched_as_string =
+        run_lockstep_simulation(scheduler, NUM_OUTPUTS, TID_BASE);
+    for (int i = 0; i < NUM_OUTPUTS; i++) {
+        std::cerr << "cpu #" << i << " schedule: " << sched_as_string[i] << "\n";
+    }
+    // See the test_synthetic_with_priorities() test which has our base sequence.
+    // But now B hits a syscall every instr, and H every other instr, so neither
```
abhinav92003, line 1208: You mean E hits a syscall every other instr and H hits it every third instr?

clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -1126,6 +1126,93 @@ test_synthetic_with_bindings_weighted()
...
+    // reaches its 3-instr quantum.  Furthermore, B isn't finished by the time E and H
```
abhinav92003, line 1209: H does execute three instrs. Maybe it gets switched out not because of the quantum but because of the syscall though.

clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -1126,6 +1126,93 @@ test_synthetic_with_bindings_weighted()
...
+                std::unique_ptr<mock_reader_t>(new mock_reader_t()), tid);
+        }
+        sched_inputs.emplace_back(std::move(readers));
+        // Set some different priorities for the middle threads.
+        // The others retain the default 0 priority.
+        sched_inputs.back().thread_modifiers.emplace_back(
+            get_tid(workload_idx, /*input_idx=*/1), /*priority=*/1);
+    }
+    // We have one input with lower timestamps than everyone, to test that it never gets
+    // switched out once we get to it among the default-priority inputs.
```
abhinav92003, line 1177: Say that it's 'J'.

clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -1126,6 +1126,93 @@ test_synthetic_with_bindings_weighted()
...
+    for (int i = 0; i < NUM_OUTPUTS; i++) {
+        std::cerr << "cpu #" << i << " schedule: " << sched_as_string[i] << "\n";
+    }
+    // See the test_synthetic_with_priorities() test which has our base sequence.
+    // But now B hits a syscall every instr, and H every other instr, so neither
+    // reaches its 3-instr quantum.  Furthermore, B isn't finished by the time E and H
+    // are done and we see the lower-priority C and F getting scheduled while B is in
+    // a wait state for its blocking syscall.
+    assert(sched_as_string[0] == "BHHHBEEBHHHBCCCBIIICCCFFFIIIFFFAAAGGGDDDAAAGGG");
+    assert(sched_as_string[1] == "EEBEEHHHEEBEBFFFBJJJJJJJJJCCCIIIDDDAAAGGGDDD");
```
abhinav92003, line 1213: We should in general also have a check that the same workload isn't scheduled on both cores at the same time.

abhinav92003 (approved):

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: Yes, updated.

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: As you noted in the prior review comment I have a typo there: it should be E which is every other.  For H the quantum and syscall coincide.

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: Done.

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: You mean the same input: i.e., that the same letter isn't on both cores at once?  OK, added a loop over the strings with that sanity check.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: In that case it will go into the `if (ready_queue_empty())` condition right above this one.  I went ahead and added a test of just one thread with blocking syscalls just to confirm there's no hang or anything.

PR #6165, opened 2023-06-22
i#6104 multi-window: Refresh timestamps in a new window

### LINKED ISSUE #6104: Multi-window traces contain low activity at the beginning of a new window
Our internal tools discovered a case when creating multi-window traces, there are missing instructions in a new buffer in the new window:
```
     1	instrs      6952, segments      1, threads     1
     2	instrs         0, segments      0, threads     0
     3	instrs         0, segments      0, threads     0
     4	instrs         0, segments      0, threads     0
     5	instrs         0, segments      0, threads     0
     6	instrs         0, segments      0, threads     0
     7	instrs         0, segments      0, threads     0
     8	instrs         0, segments      0, threads     0
     9	instrs         0, segments      0, threads     0
    10	instrs         0, segments      0, threads     0
    11	instrs         0, segments      0, threads     0
    12	instrs         0, segments      0, threads     0
    13	instrs         0, segments      0, threads     0
    14	instrs         0, segments      0, threads     0
    15	instrs         0, segments      0, threads     0
    <...>
   290	instrs         0, segments      0, threads     0
   291	instrs         0, segments      0, threads     0
   292	instrs         0, segments      0, threads     0
   293	instrs         0, segments      0, threads     0
   294	instrs         0, segments      0, threads     0
   295	instrs   1766865, segments    240, threads     1
   296	instrs  11316277, segments   1537, threads     1
   297	instrs  11348190, segments   1541, threads     1
   298	instrs  11342445, segments   1540, threads     1
   299	instrs  11324532, segments   1538, threads     1
   <...>
   996	instrs  16444436, segments   2037, threads     1
   997	instrs  15637905, segments   1956, threads     1
   998	instrs  16135216, segments   2003, threads     1
   999	instrs  16327844, segments   2017, threads     1
  1000	instrs  14145335, segments   1825, threads     1
  ```
  This was caused by a timestamp from the end of the prior window was not updated and being used in the first buffer in the new window.

### LINKED ISSUE #6143: schedule file same-timestamp entries not in same order as invariant checker expects
Seeing errors in the invariant checker's schedule file checks:

```
file: 4003708, 13330748099001637, 6, 0 vs trace 4003713, 13330748099001637, 4, 0
Trace invariant failure in shard 0x356ebe2cbc00 T-1 at ref # 0: Serial schedule entry does not match trace

file: 3997606, 13330748099001637, 10, 0 vs trace 4003716, 13330748099001637, 7, 0
Trace invariant failure in shard 0x356ebe2cbc00 T-1 at ref # 1: Serial schedule entry does not match trace

file: 4003705, 13330748099001637, 1, 0 vs trace 4003739, 13330748099001637, 2, 0
Trace invariant failure in shard 0x356ebe2cbc00 T-1 at ref # 2: Serial schedule entry does not match trace
```

It seems to be from same-timestamp entries being in a different order.  Mostly it's at the start of the file, from align_endpoints?
```
000000              4003708    13330748099001637                    6                    0
000020              3997606    13330748099001637                   10                    0
000040              4003705    13330748099001637                    1                    0
000060              4003739    13330748099001637                    2                    0
000080              4003706    13330748099001637                    3                    0
0000a0              4003743    13330748099001637                    4                    0
0000c0              4003707    13330748099001637                    6                    0
0000e0              4003714    13330748099001637                    0                    0
000100              4003703    13330748099001637                    0                    0
000120              4003715    13330748099001637                   11                    0
000140              4003716    13330748099001637                    7                    0
000160              4003711    13330748099001637                    8                    0
000180              4003710    13330748099001637                    2                    0
0001a0              4003713    13330748099001637                    4                    0
0001c0              4003726    13330748099001637                    5                    0
0001e0              4003704    13330748099001637                    5                    0
000200              4003715    13330748099003914                   11                   75
```

But sometimes in the middle two threads will have the same timestamp:
```
000f20              4003716    13330748099172378                    7                    0
000f40              4003715    13330748099172387                   10                  266
000f60              4003716    13330748099172404                    7                  246
000f80              4003715    13330748099172404                   10                  272
000fa0              4003715    13330748099172421                   10                  272
000fc0              4003716    13330748099172428                    7                  246
```

### LINKED ISSUE #6245: Multi-window tracing fails to write the last window raw file
When trying to collect a multi-window trace to test PR #6165 , the raw trace for the last window fails to be to be written:
```
Hit tracing window #0 limit: disabling tracing.
Hit retrace threshold: enabling tracing for window #1.
Hit tracing window #1 limit: disabling tracing.
Hit retrace threshold: enabling tracing for window #2.
Hit tracing window #2 limit: disabling tracing.
Hit tracing window #10 limit: disabling tracing.
Hit retrace threshold: enabling tracing for window #11.
...
Hit retrace threshold: enabling tracing for window #23.
Hit tracing window #23 limit: disabling tracing.
Hit retrace threshold: enabling tracing for window #24.
Hit tracing window #24 limit: disabling tracing.
Fatal error: failed to create trace file ./drmemtrace.tool.drcacheoff.windows-timestamps.3949632.4419.dir/raw/window.0025/drmemtrace.tool.drcacheoff.windows-timestamps.3949632.6596.raw
```

**To Reproduce**
1. Checkout PR #6165 and build tests
2. Run the new `windows-timestamps` test

### PR DESCRIPTION
**Pull request by bete0**
Updates the timestamp in the first buffer of a new window in multi-window traces
by refreshing the timestamp when re-starting tracing. Previously, the timestamp in
the first buffer was coming from the end of a prior window causing low activity at the
beginning of a new window.

Adds a unit test that fails without the fix in this change but passes with the change.

Fixes an issue where multi-window tracing fails to write the last window raw file.

Fixes #6104 
Fixes #6245

### CODE REVIEW

**Comment by bete0** [comment]
Got the following error (https://github.com/DynamoRIO/dynamorio/actions/runs/5664479853/job/15347856021):

```
Mode                LastWriteTime         Length Name                                                                  
----                -------------         ------ ----                                                                  
d-----        7/26/2023   3:59 AM                install                                                               
Exception calling "DownloadFile" with "2" argument(s): "The remote server returned an error: (504) Gateway Timeout."
At D:\a\_temp\2ec[14](https://github.com/DynamoRIO/dynamorio/actions/runs/5664479853/job/15347856021#step:5:15)538-3f2a-4097-b196-8d018e891f69.ps1:4 char:1
+ (New-Object System.Net.WebClient).DownloadFile("https://sourceforge.n ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : WebException
 
Error: Process completed with exit code 1.
```

Rerunning test.

**Comment by bete0** [comment]
`code_api|api.rseq` test failed on ci-x86/x86-64-ubuntu22.

File: clients/drcachesim/tracer/instru.h [hunk 1]
```diff
@@ -292,6 +292,10 @@ class instru_t {
     get_cpu_id();
     static uint64
     get_timestamp();
+    static uint64
+    get_attached_timestamp();
```
derekbruening, line 296: Put up by the frozen timestamp routine, and make the variable a member variable like the frozen timestamp?  See the same suggestion in another comment.  It's not really related to get_timestamp().

clients/drcachesim/CMakeLists.txt
```diff
@@ -965,6 +965,33 @@ if (BUILD_TESTS)
...
+        drmemtrace_analyzer test_helpers)
+      if (WIN32)
+        # Just like for burst_replace, linking everything together takes effort.
+        target_link_libraries(tool.drcacheoff.window_test ${static_libc})
+      endif ()
+      set(tool.drcacheoff.window_test_uses_configure ON) # Avoid dup DEBUG.
+      add_win32_flags(tool.drcacheoff.window_test)
+      use_DynamoRIO_drmemtrace_tracer(tool.drcacheoff.window_test)
+      use_DynamoRIO_extension(tool.drcacheoff.window_test drcovlib_static)
+    endif (NOT RISCV64)
```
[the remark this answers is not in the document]
derekbruening: You wrote:

> Adds a unit test that fails without the fix in this change but passes with the change.

How did you run the test?  It is not enabled as a test and has no output to match.  You just ran the executable locally and manually looked at the output?  There is no way to run it with ctest or in the suite.

clients/drcachesim/tests/window_test.cpp
```diff
@@ -0,0 +1,211 @@
...
+#    include "dr_api.h"
+#    include "drmemtrace/drmemtrace.h"
+#    include "drcovlib.h"
+#    include "analysis_tool.h"
+#    include "scheduler.h"
+#    include "tracer/raw2trace.h"
+#    include "tracer/raw2trace_directory.h"
+#    include <assert.h>
+#    include <iostream>
+#    include <math.h>
```
derekbruening, line 45: Are all of these headers really needed?

clients/drcachesim/tracer/output.cpp
```diff
@@ -979,19 +979,22 @@ process_and_output_buffer(void *drcontext, bool skip_size_cap)
...
-        // tracing.  (Switching back to timestamps at buffer output is actually
-        // worse as we then have the identical frozen timestamp for all the flushes
-        // during detach, plus they are all on the same cpu too.)
-        uint64 min_timestamp = attached_timestamp.load(std::memory_order_acquire);
-        if (min_timestamp == 0) {
-            // This data is too early: we drop it.
-            NOTIFY(1, "Dropping too-early data for T%zd\n", dr_get_thread_id(drcontext));
-            BUF_PTR(data->seg_base) = data->buf_base + header_size;
-            return;
+    if (align_attach_detach_endpoints() || has_tracing_windows()) {
```
derekbruening, line 982: So now if windows are enabled and we attach to the process, we'll align the buffers at the start even if -align_endpoints is explicitly turned off by the user.  I don't think that's what we want.  Maybe it is best to have a separate timestamp for restarting tracing (see same comment elsewhere).

clients/drcachesim/tracer/instr_counter.cpp
```diff
@@ -121,6 +121,8 @@ hit_instr_count_threshold(app_pc next_pc)
         dr_mutex_unlock(mutex);
         return;
     }
+    auto timestamp = instru_t::get_timestamp();
+    instru_t::set_attached_timestamp(timestamp);
```
derekbruening, line 125: Hmm, this might be confusing: someone might expect "attached_timestamp" to be from the single attach, not to be reset on every window start.  I think it might be best to have two separate timestamps.

derekbruening (changes requested):
Biggest issue is that the new test is never added as a test so it is never run.

clients/drcachesim/CMakeLists.txt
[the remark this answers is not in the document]
bete0: Enabled the test and added a `.templatex` file. Currently just looking at the the log `window_test passed`. Not sure if we can rely on the retrace logs.

clients/drcachesim/tests/window_test.cpp
```diff
@@ -0,0 +1,211 @@
...
+ * ARE DISCLAIMED. IN NO EVENT SHALL GOOGLE, INC. OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
```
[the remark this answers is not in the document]
bete0: Done.

clients/drcachesim/tests/window_test.cpp
```diff
@@ -0,0 +1,211 @@
...
+    return fib_with_sleep(n - 1) + fib_with_sleep(n - 2);
+}
+
+static std::string
+post_process(const std::string &out_subdir)
+{
+    const char *raw_dir;
+    drmemtrace_status_t mem_res = drmemtrace_get_output_path(&raw_dir);
+    assert(mem_res == DRMEMTRACE_SUCCESS);
+    std::string outdir = std::string(raw_dir) + DIRSEP + "../trace";
```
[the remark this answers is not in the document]
bete0: Ack. Removed it altogether.

clients/drcachesim/CMakeLists.txt
```diff
@@ -965,6 +965,33 @@ if (BUILD_TESTS)
...
+      target_link_libraries(tool.drcacheoff.window_test drmemtrace_raw2trace
+        drmemtrace_analyzer test_helpers)
+      if (WIN32)
+        # Just like for burst_replace, linking everything together takes effort.
+        target_link_libraries(tool.drcacheoff.window_test ${static_libc})
+      endif ()
+      set(tool.drcacheoff.window_test_uses_configure ON) # Avoid dup DEBUG.
+      add_win32_flags(tool.drcacheoff.window_test)
+      use_DynamoRIO_drmemtrace_tracer(tool.drcacheoff.window_test)
+      use_DynamoRIO_extension(tool.drcacheoff.window_test drcovlib_static)
```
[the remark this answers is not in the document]
bete0: Done. Removed it and cleanup the file.

clients/drcachesim/CMakeLists.txt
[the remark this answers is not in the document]
bete0: Thanks for the catch! I was running the `./clients/bin64/tool.drcacheoff.window_test`
and looking at the output. Have enabled the test and run it with ctest.

clients/drcachesim/tests/window_test.cpp
```diff
@@ -0,0 +1,211 @@
...
+#    include <thread>
+#    include "dr_api.h"
+#    include "drmemtrace/drmemtrace.h"
+#    include "drcovlib.h"
+#    include "analysis_tool.h"
+#    include "scheduler.h"
+#    include "tracer/raw2trace.h"
+#    include "tracer/raw2trace_directory.h"
+#    include <assert.h>
+#    include <iostream>
```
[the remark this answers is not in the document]
bete0: Done.

clients/drcachesim/tests/window_test.cpp
bete0: Ack. Removed the unused ones.

clients/drcachesim/CMakeLists.txt
```diff
@@ -965,6 +965,33 @@ if (BUILD_TESTS)
       use_DynamoRIO_extension(tool.drcacheoff.burst_traceopts drcovlib_static)
```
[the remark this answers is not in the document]
bete0: Done. Thanks for the catch!

clients/drcachesim/CMakeLists.txt
```diff
@@ -965,6 +965,33 @@ if (BUILD_TESTS)
...
 
+    if (NOT RISCV64) # TODO i#3544: Port tests to RISC-V 64
+      set(source_abs "${CMAKE_CURRENT_SOURCE_DIR}/tests/window_test.cpp")
+      set(asm_deps
+        "${PROJECT_SOURCE_DIR}/core/arch/asm_defines.asm"
+        "${PROJECT_BINARY_DIR}/configure.h")
+      include_directories(${PROJECT_SOURCE_DIR}/core/arch) # asm_defines.asm
+      add_split_asm_target("${source_abs}" asm_source gen_asm_tgt
+        "_asm" "" "${asm_deps}")
+      add_executable(tool.drcacheoff.window_test ${source_abs}
```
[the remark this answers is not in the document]
bete0: Done.

clients/drcachesim/tests/window_test.cpp
```diff
@@ -0,0 +1,211 @@
...
+    dr_standalone_exit();
+    std::cerr << "window_test passed\n";
+    return 0;
+}
+
+} // namespace drmemtrace
+} // namespace dynamorio
+
+#else
+#    undef UNIX
```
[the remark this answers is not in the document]
bete0: Ack. Removed this section and cleaned up the the build setup.

clients/drcachesim/tests/window_test.cpp
```diff
@@ -0,0 +1,211 @@
...
+            memref_fib.marker.marker_type == TRACE_MARKER_TYPE_TIMESTAMP) {
+            if (first_timestamp != 0) {
+                if ((memref_fib.marker.marker_value - first_timestamp) >
+                    SECONDS_TO_SLEEP * 1000000) {
+                    std::cerr << "window_test FAILED\n";
+                    exit(1);
+                }
+                break;
+            }
+            first_timestamp = memref_fib.marker.marker_value;
```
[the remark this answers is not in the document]
bete0: Ack. Renamed it to `prior_timestamp`.

clients/drcachesim/tests/window_test.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+int
+fib_with_sleep(int n)
+{
+    std::cerr << "Sleeping for " << SECONDS_TO_SLEEP << " seconds\n";
+    std::this_thread::sleep_for(std::chrono::seconds(SECONDS_TO_SLEEP));
+    if (n <= 1)
+        return 1;
+    noret_func(n, n + 1);
+    noargs();
```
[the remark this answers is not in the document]
bete0: Ack. Removed them.

clients/drcachesim/tests/window_test.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+static std::string
+post_process(const std::string &out_subdir)
+{
+    const char *raw_dir;
+    drmemtrace_status_t mem_res = drmemtrace_get_output_path(&raw_dir);
+    assert(mem_res == DRMEMTRACE_SUCCESS);
+    std::string outdir = std::string(raw_dir) + DIRSEP + "../trace";
+    std::cerr << "rawdir: " << raw_dir << "\n";
+    std::cerr << "outdir: " << outdir << "\n";
```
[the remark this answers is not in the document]
bete0: Ack. Removed them.

clients/drcachesim/tracer/output.cpp
bete0: If I understand where the aligning logic is, we only only align (where it drops the too-early data) when `align_attach_detach_endpoints()` is true from the inner `if` on line 984. When `has_tracing_windows()` is true, we just refresh.

derekbruening: No, the `if` on line 984 doesn't do anything unless min_timestamp is 0.  The refresh on line 1001 will be called even if -align_endpoints is off.

bete0: I see. Are we ok with re-refreshing when we have both `-align_endpoings` and `has_tracing_windows()`? This is what I am thinking. Keep the original local for refreshing the header timestamp and add a new refresh with `retrace_attached_timestamp` when `has_tracing_windows()`:

```
    size_t stamp_offs =
        header_size > buf_hdr_slots_size ? header_size - buf_hdr_slots_size : 0;
    uint64 min_timestamp;
    if (align_attach_detach_endpoints()) {
        // This is the attach counterpart to instru_t::set_frozen_timestamp(): we
        // place timestamps at buffer creation, but that can be before we're fully
        // attached. We update any too-early timestamps to reflect when we actually
        // started tracing.  (Switching back to timestamps at buffer output is
        // actually worse as we then have the identical frozen timestamp for all the
        // flushes during detach, plus they are all on the same cpu too.)
        min_timestamp = attached_timestamp.load(std::memory_order_acquire);
        if (min_timestamp == 0) {
            // This data is too early: we drop it.
            NOTIFY(0, "Dropping too-early data for T%zd\n",
                   dr_get_thread_id(drcontext));
            BUF_PTR(data->seg_base) = data->buf_base + header_size;
            return;
        }
        instru->refresh_unit_header_timestamp(data->buf_base + stamp_offs, min_timestamp);
    }

    if (has_tracing_windows()) {
        min_timestamp = retrace_attached_timestamp.load(std::memory_order_acquire);
        instru->refresh_unit_header_timestamp(data->buf_base + stamp_offs, min_timestamp);
    }
```

derekbruening: Well, it only changes the timestamp if it's less than the passed-in min.  Maybe it should be renamed to s/refresh/clamp/ to make its conditional nature more clear.

clients/drcachesim/tracer/instr_counter.cpp
bete0: Ack. Added a new `retrace_attached_timestamp` and kept the original one as it was.

clients/drcachesim/tracer/instru.h
```diff
@@ -292,6 +292,10 @@ class instru_t {
...
+    static void
+    set_attached_timestamp(const uint64 timestamp);
```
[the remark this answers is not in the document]
bete0: Ack. Removed this helper altogether since we now have a separate timestamp for retrace.

clients/drcachesim/tests/window_test.cpp
```diff
@@ -0,0 +1,211 @@
...
+    assert(!dr_app_running_under_dynamorio());
+    return post_process(out_subdir);
+}
+
+int
+test_main(int argc, const char *argv[])
+{
+    std::string dir_fib = gather_trace("window.0001");
+    void *dr_context = dr_standalone_init();
+    scheduler_t scheduler_fib;
```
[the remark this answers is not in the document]
bete0: Done. Cleaned up these names.

clients/drcachesim/tests/window_test.cpp
```diff
@@ -0,0 +1,211 @@
...
+    std::vector<scheduler_t::input_workload_t> sched_opt_inputs;
```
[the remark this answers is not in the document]
bete0: Done. Cleaned up. Some artifact of reused code from other tests.

clients/drcachesim/tests/window_test.cpp
```diff
@@ -0,0 +1,211 @@
...
+    auto *stream_fib = scheduler_fib.get_stream(0);
+    uint64_t first_timestamp = 0;
+    while (true) {
+        memref_t memref_fib;
+        scheduler_t::stream_status_t status_fib = stream_fib->next_record(memref_fib);
+        assert(status_fib == scheduler_t::STATUS_OK);
+        if (memref_fib.marker.type == TRACE_TYPE_MARKER &&
+            memref_fib.marker.marker_type == TRACE_MARKER_TYPE_TIMESTAMP) {
+            if (first_timestamp != 0) {
+                if ((memref_fib.marker.marker_value - first_timestamp) >
```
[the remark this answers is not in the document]
bete0: Done. Added a comment.

clients/drcachesim/tests/window_test.cpp
```diff
@@ -0,0 +1,211 @@
...
+                    SECONDS_TO_SLEEP * 1000000) {
```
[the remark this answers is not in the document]
bete0: Done.

clients/drcachesim/tracer/instru.cpp
```diff
@@ -53,6 +53,8 @@
 namespace dynamorio {
 namespace drmemtrace {
 
+std::atomic<uint64> attached_timestamp;
```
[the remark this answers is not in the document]
bete0: Now that we have a separate timestamp for reattach, will keep this timestamp untouched.

clients/drcachesim/tests/window_test.cpp
```diff
@@ -0,0 +1,211 @@
...
+    return outdir;
+}
+
+static std::string
+gather_trace(const std::string &out_subdir)
+{
+    std::string dr_ops(
+        "-stderr_mask 0xc -client_lib ';;-offline -offline -trace_after_instrs 10000 "
+        "-trace_for_instrs 2000 -retrace_every_instrs 1000");
+    std::cerr << "dr_ops: " << dr_ops << "\n";
```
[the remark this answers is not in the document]
bete0: Done. Removed it.

clients/drcachesim/tracer/output.cpp
bete0: Done. Added `retrace_attached_timestamp` and `s/refresh/clamp/ `. Now `retrace_attached_timestamp` is updated when we start retracing and is used "clamp" the header timestamp when `has_tracing_windows()`

derekbruening: Since we don't do a detach and (re-)attach and DR was attached the entire time it may be best to not have "attached" in the name.  `retrace_start_timestamp` or something?

derekbruening:
Best to get the test passing before requesting review: I see it failing on 4 different test jobs.

derekbruening (changes requested):
Best to get the test passing before requesting review: I see it failing on 4 different test jobs.

bete0:
The new `windows-timestamps` test is failing on Windows:
```
63: 
263:   Hit retrace threshold: enabling tracing for window #137.
263: 
263:   Hit tracing window #137 limit: disabling tracing.
263: 
263:   4Hit retrace threshold: enabling tracing for window #138.
263: 
263:   Hit tracing window #138 limit: disabling tracing.
263: 
263:    seconds
263: 
263:   Hit retrace threshold: enabling tracing for window #139.
263: 
263:   Hit tracing window #139 limit: disabling tracing.
263: 
263:   Hit retrace threshold: enabling tracing for window #140.
263: 
263:   Hit tracing window #140 limit: disabling tracing.
263: 
263:   Fatal error: failed to create trace file
263:   .\drmemtrace.tool.drcacheoff.windows-timestamps.exe.05792.5808.dir\raw\window.0141\drmemtrace.tool.drcacheoff.windows-timestamps.exe.05628.2643.raw
263: 
```

Also the test app results in 100+ windowed traces in Windows vs <20 in Linux.

bete0:
ci-windows (VS2019-32), and both 64 and 32 vit ci-x86 tests failed.

````
348: Running cmd |/home/runner/work/dynamorio/dynamorio/build_debug-internal-32/clients/bin32/tool.drcacheoff.burst_threadfilter|
342: CMake Error at /home/runner/work/dynamorio/dynamorio/suite/tests/runmulti.cmake:111 (message):
342:   *** cmd failed (255): Hit delay threshold: enabling tracing.
342: 
342:   Hit tracing window #0 limit: disabling tracing.
342: 
342:   Hit retrace threshold: enabling tracing for window #1.
342: 
342:   Hit tracing window #1 limit: disabling tracing.
342: 
342:   Hit retrace threshold: enabling tracing for window #2.
342: 
342:   Hit tracing window #2 limit: disabling tracing.
342: 
342:   Hit retrace threshold: enabling tracing for window #3.
342: 
342:   Hit tracing window #3 limit: disabling tracing.
342: 
342:   Fatal error: failed to create trace file
342:   ./drmemtrace.tool.drcacheoff.windows-timestamps.23271.0799.dir/raw/window.0004/drmemtrace.tool.drcacheoff.windows-timestamps.23271.3689.raw
```
It is failing trying to create the raw files for presumably the final window (#4).

bete0:
The new `windows-timestamps` test is timing out on Windows (vs2019-64): https://github.com/DynamoRIO/dynamorio/actions/runs/5789897689/job/15691902277?pr=6165#step:6:19997

bete0:
The new windows-timestamps test is timing out on Windows (vs2019-64) after 240s (https://github.com/DynamoRIO/dynamorio/actions/runs/5922421765/job/16056357222?pr=6165#step:6:19479). It might be better to disable this test on Windows to unblock this PR.

derekbruening:
> The new windows-timestamps test is timing out on Windows (vs2019-64) after 240s (https://github.com/DynamoRIO/dynamorio/actions/runs/5922421765/job/16056357222?pr=6165#step:6:19479). It might be better to disable this test on Windows to unblock this PR.

I thought the last time Windows tests failed on this PR it turned out to be a fundamental bug not related to Windows.  Also, causing a hang with an application not doing corner-case things for core DR is likely a hang in drcachesim itself (right?) which is rather disturbing.

bete0:
> > > The new windows-timestamps test is timing out on Windows (vs2019-64) after 240s (https://github.com/DynamoRIO/dynamorio/actions/runs/5922421765/job/16056357222?pr=6165#step:6:19479). It might be better to disable this test on Windows to unblock this PR.
> > 
> > 
> > I thought the last time Windows tests failed on this PR it turned out to be a fundamental bug not related to Windows. Also, causing a hang with an application not doing corner-case things for core DR is likely a hang in drcachesim itself (right?) which is rather disturbing.
> 
> That last bug was addressed with patch [79596dc](https://github.com/DynamoRIO/dynamorio/commit/79596dc7e82defea8aff59e18a12d57bc5f01478) but the Windows timeout issue surfaced after that fix. Looking to see the cause of the hang.

It was failing on Windows because `dr_create_dir` wasn't able to create the "trace" dir relative to the "raw" dir (`std::string outdir = std::string(raw_dir) + DIRSEP + "../trace";`. Fixed that.

bete0:
> > It was failing on Windows because `dr_create_dir` wasn't able to create the "trace" dir relative to the "raw" dir (`std::string outdir = std::string(raw_dir) + DIRSEP + "../trace";`. Fixed that.
> 
> Why did that hang instead of giving an error message? Can we change that behavior to not hang?

When we fail to create the dir, we just assert (https://github.com/DynamoRIO/dynamorio/pull/6165/files#diff-42688c5bb96f9dc5b78d91a4de345326caee6c892355ee14e383cb1c9261395aR91) and I can see that assert immediately if running ctest locally as expected. But in the CI test suite, somehow it waits until the timeout is reached.

derekbruening:
> > > It was failing on Windows because `dr_create_dir` wasn't able to create the "trace" dir relative to the "raw" dir (`std::string outdir = std::string(raw_dir) + DIRSEP + "../trace";`. Fixed that.
> > 
> > 
> > Why did that hang instead of giving an error message? Can we change that behavior to not hang?
> 
> When we fail to create the dir, we just assert (https://github.com/DynamoRIO/dynamorio/pull/6165/files#diff-42688c5bb96f9dc5b78d91a4de345326caee6c892355ee14e383cb1c9261395aR91) and I can see that assert immediately if running ctest locally on Windows as expected. But in the CI test suite, somehow it waits until the timeout is reached and instead of the test being reported as Failed, it's reported as a timeout.

When you see the assert locally what does it look like?  Is it a popup window that doesn't go away and blocks the process until you click "OK"?  We try to disable all of those with disable_popups() which calls _CrtSetReportMode, etc.: are those calls missing for this test?

clients/drcachesim/CMakeLists.txt
```diff
@@ -965,6 +965,23 @@ if (BUILD_TESTS)
       use_DynamoRIO_extension(tool.drcacheoff.burst_traceopts drcovlib_static)
     endif (NOT RISCV64)
 
+    if (NOT RISCV64) # TODO i#3544: Port tests to RISC-V 64
+      set(source_abs "${CMAKE_CURRENT_SOURCE_DIR}/tests/window_test.cpp")
```
derekbruening, line 969: No need for a separate var for this as it's just used once.

clients/drcachesim/CMakeLists.txt
```diff
@@ -965,6 +965,23 @@ if (BUILD_TESTS)
...
+      target_link_libraries(tool.drcacheoff.windows-timestamps drmemtrace_raw2trace
+        drmemtrace_analyzer test_helpers)
+      if (WIN32)
+        # Just like for burst_replace, linking everything together takes effort.
+        target_link_libraries(tool.drcacheoff.windows-timestamps ${static_libc})
+      endif ()
+      set(tool.drcacheoff.windows-timestamps_uses_configure ON) # Avoid dup DEBUG.
+      add_win32_flags(tool.drcacheoff.windows-timestamps)
+      use_DynamoRIO_drmemtrace_tracer(tool.drcacheoff.windows-timestamps)
+      use_DynamoRIO_extension(tool.drcacheoff.windows-timestamps drcovlib_static)
```
derekbruening, line 982: Does it actually use drcovlib?  I don't see any use.

bete0:
> > > > It was failing on Windows because `dr_create_dir` wasn't able to create the "trace" dir relative to the "raw" dir (`std::string outdir = std::string(raw_dir) + DIRSEP + "../trace";`. Fixed that.
> > > 
> > > 
> > > Why did that hang instead of giving an error message? Can we change that behavior to not hang?
> > 
> > 
> > When we fail to create the dir, we just assert (https://github.com/DynamoRIO/dynamorio/pull/6165/files#diff-42688c5bb96f9dc5b78d91a4de345326caee6c892355ee14e383cb1c9261395aR91) and I can see that assert immediately if running ctest locally on Windows as expected. But in the CI test suite, somehow it waits until the timeout is reached and instead of the test being reported as Failed, it's reported as a timeout.
> 
> When you see the assert locally what does it look like? Is it a popup window that doesn't go away and blocks the process until you click "OK"? We try to disable all of those with disable_popups() which calls _CrtSetReportMode, etc.: are those calls missing for this test?

No, I don't see a popup window locally. I just see the error message as expected:
```
  Calculating fibonacci of: 3

  Hit tracing window #11 limit: disabling tracing.

  Calculating fibonacci of: 1

  Hit retrace threshold: enabling tracing for window #12.

  Calculating fibonacci of: 2

  Hit tracing window #12 limit: disabling tracing.

  Calculating fibonacci of: 0

  Hit retrace threshold: enabling tracing for window #13.

  Calculating fibonacci of: 1

  Failed to create output dirAssertion failed: false, file
  C:\cygwin64\dynamorio\clients\drcachesim\tests\window_test.cpp,
  line 91

  ***

Call Stack (most recent call first):
  C:/cygwin64/dynamorio/suite/tests/runmulti.cmake:120 (process_cmdline)




Total Test time (real) = 100.93 sec

The following tests FAILED:
        251 - code_api|tool.drcacheoff.windows-timestamps (Failed)
Errors while running CTest
```

We are using `tests/test_helpers` similar to other tests (`e.g. burst_traceopts`) so disable_popups is being used.

clients/drcachesim/tests/window_test.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+static std::string
+post_process()
+{
+    const char *raw_dir;
+    drmemtrace_status_t mem_res = drmemtrace_get_output_path(&raw_dir);
+    assert(mem_res == DRMEMTRACE_SUCCESS);
+    std::string outdir = std::string(raw_dir) + DIRSEP + "trace";
+    void *dr_context = dr_standalone_init();
+    {
```
derekbruening, line 87: Why is this new scope needed?  Not obvious.  Either add a comment or remove the scope.

clients/drcachesim/tests/window_test.cpp
```diff
@@ -0,0 +1,185 @@
...
+    dr_standalone_exit();
+    return outdir;
+}
+
+static std::string
+gather_trace()
+{
+    std::string dr_ops(
+        "-stderr_mask 0xc -client_lib ';;-offline -offline -trace_after_instrs 1000 "
+        "-trace_for_instrs 2500 -retrace_every_instrs 1000");
```
derekbruening, line 123: Comment on how these values were picked: related to the sleep time?

clients/drcachesim/tests/window_test.cpp
```diff
@@ -0,0 +1,185 @@
...
+{
+    std::string dir = gather_trace();
+    void *dr_context = dr_standalone_init();
+    scheduler_t scheduler;
+    std::vector<scheduler_t::input_workload_t> sched_inputs;
+    sched_inputs.emplace_back(dir);
+    if (scheduler.init(sched_inputs, 1, scheduler_t::make_scheduler_serial_options()) !=
+        scheduler_t::STATUS_SUCCESS) {
+        std::cerr << "Failed to initialize scheduler " << scheduler.get_error_string()
+                  << "\n";
```
derekbruening, line 147: This seems like a fatal error that should not continue.

clients/drcachesim/tracer/instr_counter.cpp
```diff
@@ -133,6 +134,8 @@ hit_instr_count_threshold(app_pc next_pc)
     else {
         NOTIFY(0, "Hit retrace threshold: enabling tracing for window #%zd.\n",
                tracing_window.load(std::memory_order_acquire));
+        auto timestamp = instru_t::get_timestamp();
```
derekbruening, line 137: No need for a local variable right?

suite/tests/api/static_maps_mixup_novars.templatex
```diff
@@ -1,5 +1,6 @@
 mix up maps
 remap base=0x[0-9a-f]+, offs=0, sz=[1-9][0-9]*
 pre-DR init
-/* Depending on the environment, sometimes we hit an assert, sometimes a segfault */
```
derekbruening, line 4: Why is this being removed? This does not seem related to this PR?

suite/tests/CMakeLists.txt
```diff
@@ -4134,6 +4134,11 @@ if (BUILD_CLIENTS)
       set(tool.drcacheoff.burst_traceopts_nodr ON)
       set(tool.drcacheoff.burst_traceopts_nopost ON)
       torunonly_drcacheoff(burst_traceopts tool.drcacheoff.burst_traceopts "" "" "")
+      set(tool.drcacheoff.windows-timestamps_nodr ON)
+      set(tool.drcacheoff.windows-timestamps_nopost ON)
+      torunonly_drcacheoff(windows-timestamps tool.drcacheoff.windows-timestamps "" "" "")
+      set(tool.drcacheoff.windows-timestamps-expectbase "offline-windows-timestamps")
+      set(tool.drcacheoff.windows-timestamps_timeout 600)
```
derekbruening, line 4141: This is quite long: 10 full minutes.  Does it need to be this high?  If we did end up with future failures causing timeouts we don't understand we'd rather have them as short as possible to avoid making the whole CI take multiple minutes longer.

clients/drcachesim/tracer/output.cpp
```diff
@@ -1340,11 +1347,19 @@ exit_thread_io(void *drcontext)
...
 
         ptr_int_t window = get_local_window(data);
-        process_and_output_buffer(drcontext,
-                                  /* If this thread already wrote some data, include
-                                   * its exit even if we're over a size limit.
-                                   */
-                                  data->bytes_written > 0);
+        if (has_tracing_windows()) {
+            if (is_in_tracing_mode(tracing_mode.load(std::memory_order_acquire)) &&
+                get_local_window(data) == window) {
```
derekbruening, line 1352: Isn't this always true?  We just set window on line 1349.

clients/drcachesim/tracer/output.cpp
```diff
@@ -1340,11 +1347,19 @@ exit_thread_io(void *drcontext)
...
+                process_and_output_buffer(
+                    drcontext,
+                    /* If this thread already wrote some data, include
+                     * its exit even if we're over a size limit.
+                     */
+                    data->bytes_written > 0);
+            }
+        } else {
+            process_and_output_buffer(drcontext, data->bytes_written > 0);
```
derekbruening, line 1361: Better to have just one call and a merged condition, to share the call and its 1355-1357 comment.

derekbruening (commented):
Mostly nits; only thing significant is some confusion on the condition in output.cpp and why the static_maps change is here.

derekbruening:
> > When you see the assert locally what does it look like? Is it a popup window that doesn't go away and blocks the process until you click "OK"? We try to disable all of those with disable_popups() which calls _CrtSetReportMode, etc.: are those calls missing for this test?
> 
> No, I don't see a popup window locally. I just see the error message as expected:

Printing to the console is a messy thing on Windows.  You were in the "cmd" console?  Or some cygwin or msys shell in its own window?  But regardless that would just affect whether any output is visible or is just thrown away; it shouldn't affect whether there's a blocking popup.  I guess this remains a mystery.

clients/drcachesim/CMakeLists.txt
bete0: Ack. Removed it.

clients/drcachesim/CMakeLists.txt
bete0: Done. Removed it.

clients/drcachesim/tests/window_test.cpp
bete0: It was an artifact from another test this one was based on. Removed it.

clients/drcachesim/tests/window_test.cpp
bete0: Done.

clients/drcachesim/tests/window_test.cpp
bete0: Done.

clients/drcachesim/tracer/instr_counter.cpp
bete0: Ack. Removed it.

clients/drcachesim/tracer/output.cpp
bete0: Right. I have removed it.

clients/drcachesim/tracer/output.cpp
bete0: Merged the condition. Made it clear that the extra check to `is_in_tracing_mode()` is used when `is_in_tracing_mode()`

suite/tests/CMakeLists.txt
bete0: Agreed. That was while debugging the hang issue. Moved it to 240s.

suite/tests/api/static_maps_mixup_novars.templatex
bete0: This seems to be no longer the case with this change and `api.static_maps_mixup_novars_FLAKY` fails with if this string is left.

suite/tests/CMakeLists.txt
derekbruening: Do we need 240?  Most are 90.

suite/tests/api/static_maps_mixup_novars.templatex
derekbruening: But the static_maps_mixup app doesn't use drcachesim/drmemtrace at all.

clients/drcachesim/tests/window_test.cpp
```diff
@@ -118,6 +115,8 @@ post_process()
 static std::string
 gather_trace()
 {
+    // Set -trace_for_instrs and -retrace_every_instrs in such away that the
```
derekbruening, line 118: nit: "a way"

clients/drcachesim/tracer/output.cpp
```diff
@@ -1347,18 +1347,14 @@ exit_thread_io(void *drcontext)
...
-                    drcontext,
-                    /* If this thread already wrote some data, include
-                     * its exit even if we're over a size limit.
-                     */
-                    data->bytes_written > 0);
-            }
-        } else {
-            process_and_output_buffer(drcontext, data->bytes_written > 0);
+        if (!has_tracing_windows() ||
+            (has_tracing_windows() &&
```
derekbruening, line 1351: This can be removed as it is redundant.

suite/tests/api/static_maps_mixup_novars.templatex
derekbruening: Plus, cpp will remove the comment, and line 5 is optional per its regex: so I don't see how removing this would make something pass that was failing.

clients/drcachesim/tests/window_test.cpp
bete0: Done.

clients/drcachesim/tracer/output.cpp
bete0: Done. simplified to `!has_tracing_windows() || is_in_tracing_mode` to make sure we are checking if is in tracing mode when we are window tracing.

suite/tests/CMakeLists.txt
bete0: How about 120? I saw it timing out at 90 locally.

suite/tests/api/static_maps_mixup_novars.templatex
bete0: Ack. Changed it back to original.

suite/tests/CMakeLists.txt
derekbruening: Oh ok, if you have evidence it can take a while then higher is reasonable.  That is a long test though if it's >90s on a lightly loaded local machine.

derekbruening (approved):

suite/tests/CMakeLists.txt
```diff
@@ -4134,6 +4134,11 @@ if (BUILD_CLIENTS)
       set(tool.drcacheoff.burst_traceopts_nodr ON)
       set(tool.drcacheoff.burst_traceopts_nopost ON)
       torunonly_drcacheoff(burst_traceopts tool.drcacheoff.burst_traceopts "" "" "")
+      set(tool.drcacheoff.windows-timestamps_nodr ON)
+      set(tool.drcacheoff.windows-timestamps_nopost ON)
+      torunonly_drcacheoff(windows-timestamps tool.drcacheoff.windows-timestamps "" "" "")
+      set(tool.drcacheoff.windows-timestamps-expectbase "offline-windows-timestamps")
+      set(tool.drcacheoff.windows-timestamps_timeout 120)
```
derekbruening, line 4141: I can't seem to find the prior comment thread on the timeout value.  Looks like it took 100s in the x64 run. 
 Why does the test take so long to run?  Can we shrink it by reducing the sleep times or shrinking the trace?  Trying to keep the CI closer to 15 mins and each mulit-minute test pushes it longer.

derekbruening: Your new test here is by far the longest test in the suite.  Here's x86-64:
```
$ grep 'Test #' x86-64/7_Run\ Suite.txt | sort -k 8 -n
...
2023-08-24T00:54:52.5859811Z 377/470 Test #378: code_api|tool.drcacheoff.burst_threads_counts ....................   Passed    7.02 sec
2023-08-24T00:54:52.5851752Z 376/470 Test #377: code_api|tool.drcacheoff.burst_threads ...........................   Passed    7.15 sec
2023-08-24T00:53:06.4469976Z 333/470 Test #329: code_api|tool.drcacheoff.multiproc ...............................   Passed    8.84 sec
2023-08-24T00:52:37.5857344Z 311/470 Test #310: code_api|tool.drcachesim.multiproc ...............................   Passed   10.29 sec
2023-08-24T00:54:04.7949137Z 366/470 Test #364: code_api|tool.drcacheoff.func_view_heap ..........................   Passed   11.14 sec
2023-08-24T00:54:52.5866791Z 378/470 Test #379: code_api|tool.drcacheoff.burst_malloc ............................   Passed   13.76 sec
2023-08-24T00:50:56.0188972Z 153/470 Test #138: code_api|client.alloc ............................................   Passed   14.82 sec
2023-08-24T00:55:25.4520321Z 381/470 Test #382: code_api|tool.drcacheoff.burst_threadL0filter ....................   Passed   15.01 sec
2023-08-24T00:55:25.4511924Z 380/470 Test #381: code_api|tool.drcacheoff.burst_threadfilter ......................   Passed   15.20 sec
2023-08-24T00:54:52.5874247Z 379/470 Test #380: code_api|tool.drcacheoff.burst_reattach ..........................   Passed   16.40 sec
2023-08-24T00:53:29.5240263Z 346/470 Test #337: code_api|tool.drcacheoff.sysnums .................................   Passed   23.07 sec
2023-08-24T00:52:50.0526383Z 320/470 Test #311: code_api|tool.drcachesim.miss_analyzer ...........................   Passed   25.35 sec
2023-08-24T01:08:42.9138230Z 132/132 Test #370: code_api|tool.drcacheoff.windows-timestamps ......................   Passed  100.06 sec
2023-08-24T00:55:48.3853602Z 406/470 Test #375: code_api|tool.drcacheoff.windows-timestamps ......................   Passed  100.35 sec
```

bete0: Done. Decreased the sleep time and input.

derekbruening:
I pulled your branch onto a Windows machine and reproduced a popup which shows that pops are not disabled and that it is the debug libc:
```
00 00ef4b68 773c4954 win32u!NtUserWaitMessage+0xc
01 00ef4b98 77421b3b user32!DialogBoxIndirectParamAorW+0x114
02 00ef4c64 774208d1 user32!SoftModalMessageBox+0x72b
03 00ef4dc0 77421397 user32!DrawStateA+0x1421
04 00ef4e4c 77421405 user32!MessageBoxTimeoutW+0x197
05 00ef4e6c 00568cf8 user32!MessageBoxW+0x45
06 00ef4e8c 00566162 tool_drcacheoff_windows_timestamps!__acrt_MessageBoxW+0x38 [minkernel\crts\ucrt\src\appcrt\internal\winapi_thunks.cpp @ 696] 
07 00ef4ea4 00566139 tool_drcacheoff_windows_timestamps!__crt_char_traits<wchar_t>::message_box<HWND__ *,wchar_t const * const &,wchar_t const * const &,unsigned int const &>+0x22 [minkernel\crts\ucrt\inc\corecrt_internal_traits.h @ 124] 
08 00ef4ed4 005661a6 tool_drcacheoff_windows_timestamps!common_show_message_box<wchar_t>+0x109 [minkernel\crts\ucrt\src\appcrt\misc\crtmbox.cpp @ 75] 
09 00ef4ee8 00542f38 tool_drcacheoff_windows_timestamps!__acrt_show_wide_message_box+0x16 [minkernel\crts\ucrt\src\appcrt\misc\crtmbox.cpp @ 93] 
0a 00ef7174 005430d2 tool_drcacheoff_windows_timestamps!common_message_window<wchar_t>+0x488 [minkernel\crts\ucrt\src\appcrt\misc\dbgrpt.cpp @ 409] 
0b 00ef7194 00567d14 tool_drcacheoff_windows_timestamps!__acrt_MessageWindowW+0x22 [minkernel\crts\ucrt\src\appcrt\misc\dbgrpt.cpp @ 464] 
0c 00eff24c 00543064 tool_drcacheoff_windows_timestamps!_VCrtDbgReportW+0x964 [minkernel\crts\ucrt\src\appcrt\misc\dbgrptt.cpp @ 673] 
0d 00eff278 0055ec2c tool_drcacheoff_windows_timestamps!_CrtDbgReportW+0x34 [minkernel\crts\ucrt\src\appcrt\misc\dbgrpt.cpp @ 278] 
0e 00eff29c 0055edd1 tool_drcacheoff_windows_timestamps!issue_debug_notification+0x1c [minkernel\crts\ucrt\src\appcrt\internal\report_runtime_error.cpp @ 25] 
0f 00eff2b4 00545f3a tool_drcacheoff_windows_timestamps!__acrt_report_runtime_error+0x11 [minkernel\crts\ucrt\src\appcrt\internal\report_runtime_error.cpp @ 154] 
10 00eff2c4 00541774 tool_drcacheoff_windows_timestamps!abort+0x1a [minkernel\crts\ucrt\src\appcrt\startup\abort.cpp @ 51] 
11 00eff2d8 00540a5a tool_drcacheoff_windows_timestamps!common_assert_to_stderr<wchar_t>+0xa4 [minkernel\crts\ucrt\src\appcrt\startup\assert.cpp @ 187] 
12 00eff2f0 00541cda tool_drcacheoff_windows_timestamps!common_assert<wchar_t>+0x2a [minkernel\crts\ucrt\src\appcrt\startup\assert.cpp @ 415] 
13 00eff308 005256d9 tool_drcacheoff_windows_timestamps!_wassert+0x1a [minkernel\crts\ucrt\src\appcrt\startup\assert.cpp @ 443] 
14 00eff75c 0052586a tool_drcacheoff_windows_timestamps!dynamorio::drmemtrace::post_process+0x349 [D:\derek\dr\git\src\clients\drcachesim\tests\window_test.cpp @ 109] 
15 00eff79c 005258e6 tool_drcacheoff_windows_timestamps!dynamorio::drmemtrace::gather_trace+0xfa [D:\derek\dr\git\src\clients\drcachesim\tests\window_test.cpp @ 132] 
16 00eff96c 008fa655 tool_drcacheoff_windows_timestamps!dynamorio::drmemtrace::test_main+0x36 [D:\derek\dr\git\src\clients\drcachesim\tests\window_test.cpp @ 138] 
17 00eff97c 0053a7d3 tool_drcacheoff_windows_timestamps!main+0x15

0:000> .frame 14
14 00eff75c 0052586a tool_drcacheoff_windows_timestamps!dynamorio::drmemtrace::post_process+0x349 [D:\derek\dr\git\src\clients\drcachesim\tests\window_test.cpp @ 109]
0:000> dv error
          error = "Failed to process file for thread 10352: memref entry found outside of bb"
```

disable_popups() needs DEBUG set but that is not set:
```
build clients\drcachesim\CMakeFiles\tool.drcacheoff.windows-timestamps.dir\tests\window_test.cpp.obj: CXX_COMPILER__tool.2edrcacheoff.2ewindows-timestamps D$:\derek\dr\git\src\clients\drcachesim\tests\window_test.cpp || cmake_object_order_depends_target_tool.drcacheoff.windows-timestamps
  DEFINES = -DDR_APP_EXPORTS -DDR_HOST_X86 -DWINDOWS -DX86_32 -D_CRT_SECURE_NO_WARNINGS -D_X86_
```

Looks like it's because you've disabled adding DEBUG:
```
      set(tool.drcacheoff.windows-timestamps_uses_configure ON) # Avoid dup DEBUG.
```
I don't see a configure.h being included in your test: so not sure why you have that line?
Do some other 

Getting DEBUG there by removing that line should get rid of the popups and avoid the timeouts.
There is still the assert itself for the raw2trace failure.

derekbruening:
```
282:   [drmemtrace]: Appending 3 instrs in bb 0A032AB0 in mod 5 +0x72ab0 =
282:   C:\WINDOWS\SYSTEM32\ntdll.dll
282:
282:   [drmemtrace]: Appending delayed branch tagalong entry type write (1) for
282:   thread 0
282:
282:   [drmemtrace]: Appending delayed branch tagalong entry type read (0) for
282:   thread 0
282:
282:   [drmemtrace]: Inserted indirect branch target 77802AB0
282:
282:   [drmemtrace]: Appending delayed branch type=15 pc=00508CE3 decode=09498CE3
282:   target=00000000 for thread 0
282:
282:   [drmemtrace]: Appending delayed branch tagalong entry type encoding (47)
282:   for thread 0
282:
282:   [drmemtrace]: Appending delayed branch tagalong entry type encoding (47)
282:   for thread 0
282:
282:   [drmemtrace]: Chunk instr count is now 0
282:
282:   [drmemtrace]: Appended encoding entry for 0A032AB0 sz=4 0x1a0008b8...
282:
282:   [drmemtrace]: Appended encoding entry for 0A032AB0 sz=1 0x00000000...
282:
282:   [drmemtrace]: Appended instr fetch for original 77802AB0
282:
282:   [drmemtrace]: Chunk instr count is now 0
282:
282:   [drmemtrace]: Appended encoding entry for 0A032AB5 sz=4 0x048a20ba...
282:
282:   [drmemtrace]: Appended encoding entry for 0A032AB5 sz=1 0x0000000a...
282:
282:   [drmemtrace]: Appended instr fetch for original 77802AB5
282:
282:   [drmemtrace]: Chunk instr count is now 0
282:
282:   [drmemtrace]: Omitting syscall instr without subsequent number marker.
282:
282:   [drmemtrace]: extra memref entry: 00BDF1AC
282:
282:   raw2trace failed: Failed to process file for thread 13192: memref entry
282:   found outside of bb
282:
282:   Assertion failed: false, file
282:   D:\derek\dr\git\src\clients\drcachesim\tests\window_test.cpp, line 109

0:000> U ntdll+72ab0
ntdll!NtWriteFile:
77802ab0 b808001a00      mov     eax,1A0008h
77802ab5 ba208a8177      mov     edx,offset ntdll!RtlInterlockedCompareExchange64+0x170 (77818a20)
77802aba ffd2            call    edx
77802abc c22400          ret     24h
77802abf 90              nop
```
Looks like that assert is from removing what DR treats as a syscall instr on WOW64 which is a call which has a store and the store is left dangling.
Filed #6280.

clients/drcachesim/CMakeLists.txt
```diff
@@ -966,6 +966,22 @@ if (BUILD_TESTS)
...
+        "${CMAKE_CURRENT_SOURCE_DIR}/tests/window_test.cpp")
+      configure_DynamoRIO_static(tool.drcacheoff.windows-timestamps)
+      use_DynamoRIO_static_client(tool.drcacheoff.windows-timestamps drmemtrace_static)
+      target_link_libraries(tool.drcacheoff.windows-timestamps drmemtrace_raw2trace
+        drmemtrace_analyzer test_helpers)
+      if (WIN32)
+        # Just like for burst_replace, linking everything together takes effort.
+        target_link_libraries(tool.drcacheoff.windows-timestamps ${static_libc})
+      endif ()
+      set(tool.drcacheoff.windows-timestamps_uses_configure ON) # Avoid dup DEBUG.
```
derekbruening, line 980: Please see the top-level comment about removing this (even if it doesn't affect the popups, it still should not be here and will affect any DEBUG use in the main test file).  Making a comment so it can be marked whether resolved.

bete0: Ack. Removed it.

PR #6169, opened 2023-06-22
i#4343: Add dynamorio::drmemtrace namespace to all drcachesim/ code

### LINKED ISSUE #4343: [cleanup] use namespaces in C++ code, in particular drcachesim

derekbruening:
Adds the dynamorio::drmemtrace namespace to all drcachesim/ code.  It was already present on newer code.

This breaks compatibility and external code using these files will need to be updated and recompiled.

Shifts suite/tests/rseq.c to be C++ code so it can continue to test drmemtrace_client_main() which is now inside the namespace.

Fixes #4343

### DIFF
--- a/clients/drcachesim/tools/invariant_checker.cpp
+++ b/clients/drcachesim/tools/invariant_checker.cpp
@@ -38,6 +38,9 @@
 #include <cassert>
 #include <string.h>
 
+namespace dynamorio {
+namespace drmemtrace {
+
 analysis_tool_t *
 invariant_checker_create(bool offline, unsigned int verbose)
 {
@@ -921,3 +924,6 @@ invariant_checker_t::check_for_pc_discontinuity(
 
     return error_msg;
 }
+
+} // namespace drmemtrace
+} // namespace dynamorio
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -166,24 +166,14 @@ changes:
    INSTR_CREATE_eor_sve_pred(), INSTR_CREATE_and_sve_pred() and
    INSTR_CREATE_bic_sve_pred() to use the new vector element registers and to
    correctly encode the predicate mode.
- - Added a new drmemtrace analyzer option \p -interval_microseconds and various new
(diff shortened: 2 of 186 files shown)


derekbruening:
Somehow making the rseq.c test into C++ triggered problems with raw2trace trying to place signal and rseq events in i-filtered traces.  Shouldn't it not try to do rseq rollbacks for i-filtered?

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -160,9 +163,12 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
              shard->memrefs_until_interrupt_ == 0)) {
             report_if_false(
                 shard,
-                (memref.marker.type == TRACE_TYPE_MARKER &&
-                 (memref.marker.marker_type == TRACE_MARKER_TYPE_KERNEL_EVENT ||
-                  memref.marker.marker_type == TRACE_MARKER_TYPE_RSEQ_ABORT)) ||
+                // I-filtered traces don't have all instrs so this doesn't apply.
+                TESTANY(OFFLINE_FILE_TYPE_FILTERED | OFFLINE_FILE_TYPE_IFILTERED,
```
abhinav92003, line 167: I was thinking maybe we should move refactoring-unrelated changes to a separate PR. But that's optional.

abhinav92003 (approved):

clients/drcachesim/tools/invariant_checker.cpp
derekbruening: I was also considering separating the two changes in commit [8d394c9](https://github.com/DynamoRIO/dynamorio/pull/6169/commits/8d394c9de4b159b789fc7f95cbfda719f45d7b24).  One reason not to separate is that there is no simple test that will reproduce: unless the change to C++ for rseq.c were also separated.  (I don't understand why this shows up with the C++ change.)

Issue #5844, opened 2023-01-31
Add scatter/gather support for AArch64's Scalable Vector Extension (SVE)


AssadHashmi:
We will shortly be upstreaming SVE scatter/gather instructions as part of AArch64 SVE codec and core/tools work (https://github.com/DynamoRIO/dynamorio/issues/5365, https://github.com/DynamoRIO/dynamorio/issues/3044). There are two aspects of this work:
- **IR and codec**: new AArch64 addressing modes, an extension of the existing `opnd_create_base_disp_aarch64()` or a new `opnd_create_base_disp_` function for SVE.
- **Core and tools**: emulation of SVE scatter/gather for observing memory, similar to AVX512, see https://dynamorio.org/page_scatter_gather_emulation.html

In both cases we would like to be consistent with the design of the AVX512 IR API and emulation implementation, as much as possible. We would appreciate any advice on the best approach and the avoidance of pitfalls which you may have have encountered during your work on AVX512. As well as pointers to code which is relevant to and/or will help us with implementing for SVE.

derekbruening:
AVX-2 and AVX-512 scatter/gather expansion was implemented by @abhinav92003 and there is a doc here: https://dynamorio.org/page_scatter_gather_emulation.html

Abhinav may have more to say as well.  From my side: it is maybe surprisingly complex for originating from a single instruction, as the example expansions in that doc show, and you can see in the code the state restore is equally complex with a giant state machine.

derekbruening:
I think this is a dup of #5036 which has a bunch of observations

Issue #5036, opened 2021-08-04
Expand A64 SVE memory access instructions


abhinav92003:
ARM has a rich set of SVE instructions[1]. For clients that need to instrument memrefs (e.g. drcachesim), we need to expand them to scalar loads and stores, like what we did for x86 scatter/gather in #2985. These instructions have many variants; I'm summarising my understanding below. [1] and the Scalable Vector Extension of the Arm manual[2] have a more detailed discussion.

There are multiple ways in which the memory address can be specified. These are all predicated loads/stores, meaning that an element may be active or inactive, based on special predicate registers. They have either the `LD1*` or `ST1*` prefix

- Scalar + immediate: For contiguous access. Memory address is generated by a 64-bit scalar base and immediate index.
- Scalar + scalar: For contiguous access. Memory address is generated by a 64-bit scalar base and scalar index which is added to the base address.
- Scalar + vector: For possible non-contiguous access, also known as gather load/scatter store. Memory addresses are generated by a 64-bit scalar base plus vector index.
- Vector + immediate:  For possible non-contiguous access, also known as gather load/scatter store. Memory addresses are generated by a vector base plus immediate index.

There are variants with different element sizes (unsigned double-word; signed and unsigned byte, halfword, word).

Faults for non-active elements are always suppressed. There are different load instruction variants based on how faults for active elements are treated: besides the usual, each of the above has a “first fault” (faults only for first active element) and “non fault” variants.

For “scalar plus scalar” and “scalar plus immediate” load instructions, there are variants that allow reading contiguous 2/3/4 elements, each to the same element number in 2/3/4 vector registers.  These have LDN* or STN* prefix, where N=2/3/4.

There are also some un-predicated instructions (LDR and STR) that use the "scalar + immediate" scheme to load/store vectors or predicate registers.

The x86 scatter/gather that we handled in #2985 is the "scalar + vector" variant with regular faulting behaviour. More work will be required to adapt `drx_expand_scatter_gather` to these other variants.

For the contiguous access variants, we could model them as a single memory address with a larger size. But this is not a correct model, because each element can be active/inactive based on the predicate register, so the memory addresses that end up being accessed can be non-contiguous. It'll be correct to model them as scatter/gather, using multiple element-sized accesses.

[1]: https://developer.arm.com/documentation/ddi0596/2021-03/SVE-Instructions

[2]: https://developer.arm.com/documentation/ddi0584/latest/

abhinav92003:
[...]

Issue #3837, opened 2019-09-16
Add support for drx_expand_scatter_gather() and instr_is_{scatter,gather}() on AArch32


hgreving2304:
The functions instr_is_scatter() and instr_is_gather() will be added along with a patch shortly. This issue covers support for ARM/AArch64.

derekbruening:
This issue includes arm support for drx_expand_scatter_gather(), right?  Maybe add that to the title.

derekbruening:
Xref #2985 which is the x86 issue.

derekbruening:
This only covers AArch32; #5036 covers AArch64.

derekbruening:
Turns out there are not any AArch32 scatter-gather instructions so we're closing this and leaving just #5036 as it has more discussion/content.

PR #6173, opened 2023-06-26
i#5844 Move x86 expand_scatter_gather to its own file

jackgallagher-arm:
The existing drx_expand_scatter_gather() function is > 2000 lines of code, including helper functions, and most of it is very x86 specific.

In preparation for writing the AArch64 version of this function, this commit moves the x86 implementation of drx_expand_scatter_gather() out of drx.c into its own architecture specific files, and creates dummy implementations for other architectures.

The behaviour of the function should be unchanged on all platforms.

As we develop the AArch64 version of drx_expand_scatter_gather(), any parts of the original x86 implementation which are usefull to both can be pulled out into helper functions which are shared between both architectures.

Issues: #5844, #5036, #3837


AssadHashmi:
@derekbruening @abhinav92003 I'm happy with this change from the Arm point-of-view. Will leave to you folks to do the final approval from the X86 point-of-view.

ext/drx/aarch64/scatter_gather.c
```diff
@@ -0,0 +1,53 @@
+/* **********************************************************
```
derekbruening, line 1: Unless we expect additional files added to these subdirectories, maybe it would be simpler to use suffixes rather than subdirs, like we do for things like `core/unix/loader_{android,linux,macos}.c` and `ext/drsyms/drsyms_{dwarf,elf,macho}.c` and other places: so we'd have `ext/drx/scatter_gather_{aarch64,arm,riscv64,x86}.c`?

ext/drx/aarch64/scatter_gather.c
```diff
@@ -0,0 +1,53 @@
...
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+/* DynamoRio eXtension utilities */
```
derekbruening, line 34: Update these comments to reflect what each particular file is for.

abhinav92003 (approved):

derekbruening:
This broke all my local builds:
```
.../src/ext/drx/scatter_gather_x86.c:53:18: fatal error: ../../../core/unix/include/syscall.h: No such file or directory
   53 | #        include "../../../core/unix/include/syscall.h"
```

The include path has too many `..` elements.  How did it pass all the automated tests?

derekbruening:
This also breaks our embedded build.

Still not understanding how it passes the GA tests: some particular build directory placement has a generated header build dir that's already added to the compiler include search path which just happens to reach core/ with 3 ..'s??

PR #6174, opened 2023-06-26
Fix out-of-bounds vector access in Scheduler

abhinav92003:
Avoids an out-of-bounds vector access by checking for init value. This occurs in the case when an output starts in the wait state.

Modifies an existing scheduler unit test to test this case.


derekbruening (approved):

abhinav92003:
Windows runs are failing for some reason in the merge queue.

abhinav92003:
```
Exception calling "DownloadFile" with "2" argument(s): "The remote server returned an error: (503) Server Unavailable."
```

I guess I'll just retry later.

PR #6175, opened 2023-06-26
Sort also by CPU for schedule file invariant check

abhinav92003:
In some internal tests we modify the TID of some entries to test the schedule file invariant check. This inadvertently inverts the order of some entries for this check if the new synthetic TID matches with an existing real TID. By sorting also by CPU id, we ensure consistent ordering.

### DIFF
--- a/clients/drcachesim/tools/invariant_checker.cpp
+++ b/clients/drcachesim/tools/invariant_checker.cpp
@@ -735,12 +735,15 @@ invariant_checker_t::check_schedule_data(per_shard_t *global)
             vec.insert(vec.end(), keyval.second.begin(), keyval.second.end());
         }
     }
-    std::sort(serial.begin(), serial.end(),
-              [](const schedule_entry_t &l, const schedule_entry_t &r) {
-                  if (l.timestamp == r.timestamp)
-                      return l.thread < r.thread;
-                  return l.timestamp < r.timestamp;
-              });
+    auto schedule_entry_comparator = [](const schedule_entry_t &l,
+                                        const schedule_entry_t &r) {
+        if (l.timestamp != r.timestamp)
+            return l.timestamp < r.timestamp;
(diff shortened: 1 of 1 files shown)


clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -735,12 +735,20 @@ invariant_checker_t::check_schedule_data(per_shard_t *global)
...
+                                        const schedule_entry_t &r) {
+        if (l.timestamp != r.timestamp)
+            return l.timestamp < r.timestamp;
+        if (l.cpu != r.cpu)
+            return l.cpu < r.cpu;
+        // We really need to sort by either (timestamp, cpu_id) or
+        // (timestamp, thread_id): a single thread cannot be on two CPUs at
+        // the same timestamp; also a single CPU cannot have two threads at the
+        // same timestamp. We still sort by (timestamp, cpu_id, thread_id)
+        // to prevent inadvertent issues with test data.
```
derekbruening, line 748: We should also fix the test to avoid the tid collision, right?

derekbruening (approved):

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: Well, there's nothing really wrong with the test. The test trace has two threads with consecutive TIDs, which is not wrong. And the test uses a simple strategy to introduce errors in the schedule file (incrementing the TID for some entries), which seems as good as any. The test also modifies every 10th entry, which again is a heuristic as good as any.

abhinav92003: With this fix, the test would work fine. Also, this fix does not introduce any significant overhead.

derekbruening: OK.

PR #6176, opened 2023-06-27
raw2trace optimization: bitset, not passing vector by value, check empty bool

kuhanov:
Split from #6115 

Adding bitset_hash_table_t gives 20% speed up if we stay on original approach.
Changes:
bitset instead of set for encoding history
not passing vector by value
keeping bool for whether delayed branch container is empty


derekbruening (approved):
(Still had a few nits from PR #6115 but small enough to not re-raise.)
(Typically we would raise an issue (e.g., issue on raw2trace performance) to cite in these PR's, but not required.)

PR #6183, opened 2023-07-04
i#5383 mac a64, part 5: Get hello,world to run

derekbruening:
Fixes global references in aarch64.asm which were loading the value
instead of the address (the Linux a64 assembler auto-magically gives
the address whether using ADD or LDR!).

Fixes and clarifies the Mac64 x86 and arm TLS slots vs offsets.

Uses gettimeofday() for query_time_seconds() instead of SYS_gettimeofday.

Removes an invalid hardcoded 0 sysnum for global_do_syscall.

Issue: #5383

### DIFF
--- a/api/samples/stl_test.cpp
+++ b/api/samples/stl_test.cpp
@@ -77,7 +77,9 @@ DR_EXPORT void
 dr_init(client_id_t client_id)
 {
     int i;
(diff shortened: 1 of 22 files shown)


derekbruening:
The diff vs base didn't go as planned here -- will try again w/ the precise base branch.

PR #6192, opened 2023-07-10
Fix EVEX vmovq encoding

### LINKED ISSUE #6190: Wrong EVEX prefix for AVX512 `vmovq`
Dynamorio encodes `vmovq  %xmm16[8byte] -> (%r13)[8byte]` into `62 c1 7d 18 d6 45 00`.
This is an illegal instruction and `62 c1 fd 08 d6 45 00` should be the correct encoding.

**To Reproduce**
1. Create and encode instruction `vmovq  %xmm16[8byte] -> (%r13)[8byte]`.
2. Examine the encoding.

**Expected behavior**
EVEX.W should be 1 and EVEX.b should be 0 according to the x86-64 manual.

**Versions**
cronbuild-9.93.19531

khuey:
The vmovq entries that sit in the prefix_extension table have incorrect encoding information. Specifically, the final byte 0x10, in EVEX mode, sets EVEX.b. 0x40 sets EVEX.W, which is what's needed here. This was likely simply copied over from the VEX entries without adjustment.

The vmovq entries in evex_Wb_extensions are already correct.

Fixes #6190


suite/tests/api/ir_x86.c
```diff
@@ -1544,6 +1544,23 @@ test_x64_vmovq(void *dc)
                               false /*no bytes*/, dbuf, BUFFER_SIZE_ELEMENTS(dbuf), &len);
     ASSERT(pc == &b2[7]);
     ASSERT(strcmp(dbuf, "vmovq  (%rdx,%rcx)[8byte] -> %xmm25\n") == 0);
+
+    const char expected1[] = { 0x62, 0xc1, 0xfe, 0x08, 0x7e, 0x45, 0x00 };
+    const char expected2[] = { 0x62, 0xc1, 0xfd, 0x08, 0xd6, 0x45, 0x00 };
```
derekbruening, line 1549: nit: the others up above in this function and throughout this file are all type `byte` (unsigned, while char is usually signed) -- best to be consistent.

derekbruening (approved):

khuey:
Tests seem too flaky to merge :/

derekbruening:
#2204 covers the 3x-retry feature and tests that still fail 3x in a row: please update it if a new failure is hit not already mentioned there.  I did this for the drcachesim failure which seems to be a stale pipe file which will prevent retries form helping any of those tests: https://github.com/DynamoRIO/dynamorio/issues/2204#issuecomment-1629916592

MikeWalrus:
Thanks!

PR #6194, opened 2023-07-11
Fix build with GCC 13.1 and Doxygen 1.9.7

MikeWalrus:
Adds missing include for `stdint.h` in one file that relies on `math.h`, which breaks with GCC 13.1.

Also, removes a redundant default-value assignment for DOT_TRANSPARENT, which is deprecated
and would abort in the newest Doxygen.

### DIFF
--- a/api/docs/API.doxy
+++ b/api/docs/API.doxy
@@ -239,7 +239,6 @@ DOT_IMAGE_FORMAT       = png
 DOT_PATH               =
 DOTFILE_DIRS           =
 MAX_DOT_GRAPH_DEPTH    = 1000
-DOT_TRANSPARENT        = NO
 DOT_MULTI_TARGETS      = NO
 GENERATE_LEGEND        = YES
 DOT_CLEANUP            = YES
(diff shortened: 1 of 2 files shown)


abhinav92003 (approved):

api/docs/API.doxy
```diff
@@ -239,7 +239,6 @@ DOT_IMAGE_FORMAT       = png
 DOT_PATH               =
 DOTFILE_DIRS           =
 MAX_DOT_GRAPH_DEPTH    = 1000
-DOT_TRANSPARENT        = NO
```
abhinav92003, line 242: I suppose we removed this because setting it to NO is a no-op since that is the default?

MikeWalrus: Yes, the default value for DOT_TRANSPARENT was NO. See https://github.com/doxygen/doxygen/blob/a44f6a7e5caf1fbf5fe271c99046a51b315d5f56/src/config.xml#L3957

abhinav92003: Ok.

abhinav92003:
Thanks for the fix. I'll merge this PR.

MikeWalrus:
Thanks.

Issue #6196, opened 2023-07-12
ASSERT UBSAN reports UB in opnd_type_ok


dvyukov:
Executing code that invokes instr_zeroes_ymmh() with -fsanitize=undefined, I am getting:
```
decoding c6 05 78 08 69 00 01 mov    $0x01 -> <rel> 0x000055ffb5d9a308[1byte] at 0x55ffb5709a89
core/ir/x86/encode.c:1293:21: runtime error: applying non-zero offset 17 to null pointer
    #0 0x7f4026f6f207 in opnd_type_ok core/ir/x86/encode.c:1293:21
    #1 0x7f4026f65ae3 in encoding_possible_pass core/ir/x86/encode.c:1534:5
    #2 0x7f4026f659a2 in encoding_possible core/ir/x86/encode.c:1607:10
    #3 0x7f4026f5084d in get_encoding_info core/ir/encode_shared.c:104:13
    #4 0x7f4026f41725 in instr_zeroes_ymmh core/ir/instr_shared.c:2046:32
```
It seems that di is initialized only with decode_info_init_for_instr(), which just zeros it, so `di->final_pc` is a NULL pointer.

@derekbruening said elsewhere: "Hmm, probably what should happen is this type of encoding template fit check should ignore reachability, and not execute that reachable expression at all".

On revision aca56afe82de1608110abd62b41dbc8d7787b235.

PR #6197, opened 2023-07-12
i#6196: fix UBSAN-reported issue in opnd_type_ok

dvyukov:
di->final_pc is NULL (not initialized) when called from encoding_possible(). So the reachability check does not really work and it's not necessary to determine possible encodings.
Adding something to a NULL pointer is also an UB as detected by UBSAN. So remove the reachability check in opnd_type_ok() for TYPE_O.

Fixes: #6196

### DIFF
--- a/core/ir/x86/encode.c
+++ b/core/ir/x86/encode.c
@@ -1285,15 +1285,6 @@ opnd_type_ok(decode_info_t *di /*prefixes field is IN/OUT; x86_mode is IN*/, opn
     }
     case TYPE_O:
         return ((opnd_is_abs_addr(opnd) ||
-#ifdef X64
-                 /* We'll take a relative address that rip-rel won't reach:
-                  * after all, OPND_CREATE_ABSMEM() makes a rip-rel.
-                  */
-                 (opnd_is_rel_addr(opnd) &&
-                  (!REL32_REACHABLE(di->final_pc + MAX_INSTR_LENGTH,
-                                    (byte *)opnd_get_addr(opnd)) ||
-                   !REL32_REACHABLE(di->final_pc + 4, (byte *)opnd_get_addr(opnd)))) ||
-#endif
                  (!X64_MODE(di) && opnd_is_mem_instr(opnd))) &&
                 size_ok(di, opnd_get_size(opnd), opsize, false /*!addr*/));
     case TYPE_X:
--- a/core/ir/x86/encode.c
+++ b/core/ir/x86/encode.c
@@ -1284,8 +1284,7 @@ opnd_type_ok(decode_info_t *di /*prefixes field is IN/OUT; x86_mode is IN*/, opn
         return (opnd_is_far_pc(opnd) || opnd_is_far_instr(opnd));
     }
     case TYPE_O:
-        return ((opnd_is_abs_addr(opnd) ||
-                 (!X64_MODE(di) && opnd_is_mem_instr(opnd))) &&
+        return ((opnd_is_abs_addr(opnd) || (!X64_MODE(di) && opnd_is_mem_instr(opnd))) &&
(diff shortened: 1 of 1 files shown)


dvyukov:
The windows failure looks unrelated:
```
debug-internal-64: **** 2 build errors ****
	D:\a\dynamorio\dynamorio\suite\tests\api\ir_x86.c(500) : fatal error C1002: compiler is out of heap space in pass 2
	D:\a\dynamorio\dynamorio\suite\tests\api\ir_x86.c(458) : fatal error C1002: compiler is out of heap space in pass 2
```

dvyukov:
Not sure about rseq test:
```
442: kernel_xfer_event: type 1
442: kernel_xfer_event: type 10

xxxxxxxxxx 30s elapsed xxxxxxxxxxx

xxxxxxxxxx 30s elapsed xxxxxxxxxxx
442: <Application /home/runner/work/dynamorio/dynamorio/build_debug-internal-64/suite/tests/bin/api.rseq (19837). Failed to synchronize with all threads when detaching.>
    Test #442: code_api|api.rseq .................***Failed  Required regular expression not found. Regex=[^kernel_xfer_event: type 10
```

Is there any history of flakiness for this test?

core/ir/x86/encode.c
```diff
@@ -1284,17 +1284,7 @@ opnd_type_ok(decode_info_t *di /*prefixes field is IN/OUT; x86_mode is IN*/, opn
         return (opnd_is_far_pc(opnd) || opnd_is_far_instr(opnd));
     }
     case TYPE_O:
-        return ((opnd_is_abs_addr(opnd) ||
-#ifdef X64
-                 /* We'll take a relative address that rip-rel won't reach:
-                  * after all, OPND_CREATE_ABSMEM() makes a rip-rel.
-                  */
-                 (opnd_is_rel_addr(opnd) &&
-                  (!REL32_REACHABLE(di->final_pc + MAX_INSTR_LENGTH,
```
derekbruening, line 1293: Hmm removing this completely was not what I had in mind initially.  Also I wanted to change at least one place where a test deliberately uses a PC of 0 as an additional fix for this issue.  Will try to get more feedback in next day.  Thank for taking a stab at this; it was on my list but never made it to the top.

derekbruening, line 1293: > di->final_pc is NULL (not initialized) when called from encoding_possible(). So the reachability check does not really work and it's not necessary to determine possible encodings.

But di->final_pc is set by instr_encode_arch() right before calling encoding_possible().  That is the typical call path.  The instance where final_pc is not set is when called by get_encoding_info() for e.g. instr_zeroes_ymmh().

derekbruening (changes requested):

core/ir/x86/encode.c
derekbruening: So this is used in walking the encoding template possibilities and I'm guessing TYPE_O is early in the sequence, as it is more specialized.  That's why it needs to exclude the reachable so that will bypass and pick a later template.

So the two calls outside a regular encode, get_encoding_info() and instr_is_encoding_possible(), which don't take in a final PC, may pick the wrong template for MOV.  Although it won't matter for instr_is_encoding_possible() who just wants a yes/no answer.  For the various callers of get_encoding_info(): instr_shrink_to_16_bits() is the only one which might have an issue: but it looks like it's ok.

I think what we want to do is document this and add comments in each caller stating that they're fine with this caveat.

derekbruening: I created #6202 for this.  Could I ask you to test it in your ubsan setup to confirm it eliminates the problem in front of you?

dvyukov: Thanks! Let's move the conversation to that PR.

PR #6200, opened 2023-07-12
drcachesim: add cache name and parameter accessors.

brettcoon:
Adds a name to caching_device, with accessors for the name and other cache parameters.  Added a "caching_device" pointer to caching_device_stats, to make it easier for client code to access the cache associated with a stats object.  Updated drcachesim output reports to include cache names and a short description of each cache's configuration.  Updated tests for the output format changes and new methods.

### DIFF
--- a/clients/drcachesim/simulator/cache_simulator.cpp
+++ b/clients/drcachesim/simulator/cache_simulator.cpp
@@ -84,14 +84,14 @@ cache_simulator_t::cache_simulator_t(const cache_simulator_knobs_t &knobs)
     // XXX i#1703: get defaults from hardware being run on.
 
     // This configuration allows for one shared LLC only.
-    cache_t *llc = create_cache(knobs_.replace_policy);
+    std::string cache_name = "LL";
+    cache_t *llc = create_cache(cache_name, knobs_.replace_policy);
     if (llc == NULL) {
         error_string_ = "create_cache failed for the LLC";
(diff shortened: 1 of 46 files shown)


clients/drcachesim/simulator/cache_simulator.cpp
```diff
@@ -578,26 +578,31 @@ cache_simulator_t::print_results()
         print_core(i);
         if (thread_ever_counts_[i] > 0) {
             if (l1_icaches_[i] != l1_dcaches_[i]) {
-                std::cerr << "  L1I stats:" << std::endl;
+                std::cerr << "  " << l1_icaches_[i]->get_name()
+                          << " stats: " << l1_icaches_[i]->get_description() << std::endl;
```
derekbruening, line 582: So it would be like:
```
L1D stats: size=n, assoc=n, block=n, FIFO, coherent
  Hits:
```

It may be better as:
```
L1D (size=n, assoc=n, block=n, FIFO, coherent) stats:
  Hits:
```

derekbruening (approved):

clients/drcachesim/simulator/cache_simulator.cpp
brettcoon: You're right, that is better.  Just wish it didn't affect so many templates...
Oh, well, right is right.

PR #6202, opened 2023-07-13
i#6196: Avoid adding to a null pointer

### LINKED ISSUE #6196: ASSERT UBSAN reports UB in opnd_type_ok

derekbruening:
Avoids non-regular-encode code paths which do not set a final PC and thus have a null PC value from manipulating that PC in opnd_type_ok().

Documents that get_encoding_info() may not find the same template as when called with the real final PC.  Documents in key callers that this is fine.

Adjusts the raw2trace unit test helper to use a non-null final PC.

Fixes #6196

### DIFF
--- a/clients/drcachesim/tracer/raw2trace.h
+++ b/clients/drcachesim/tracer/raw2trace.h
@@ -643,9 +643,11 @@ public:
     test_module_mapper_t(instrlist_t *instrs, void *drcontext)
         : module_mapper_t(nullptr)
     {
-        // We encode for 0-based addresses for simpler tests with low values.
-        byte *pc = instrlist_encode_to_copy(drcontext, instrs, decode_buf_, nullptr,
-                                            nullptr, true);
+        // We encode for 1-based addresses for simpler tests with low values while
+        // avoiding null pointer manipulation complaints (xref i#6196).
+        byte *pc = instrlist_encode_to_copy(
+            drcontext, instrs, decode_buf_,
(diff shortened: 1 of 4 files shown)


dvyukov (approved):
Yes, this helps with UBSAN in my setup. Thanks!

abhinav92003 (approved):

derekbruening:
x86-32 failure is once again scattergather #5329

PR #6208, opened 2023-07-17
i#3544 RV64: Added a new immediate format for AUIPC

ksco:
AUIPC is a PC-relative instruction, which needs to be mangled before going into fcache, so the src opnd needs to be a PC or instr. This patch added a new immediate type called `u_immpc` for AUIPC to achieve this.


suite/tests/api/ir_riscv64.c
```diff
@@ -109,7 +109,13 @@ test_instr_encoding_jump_or_branch(void *dc, uint opcode, instr_t *instr)
     decin = instr_create(dc);
     next_pc = decode(dc, buf, decin);
     ASSERT(next_pc != NULL);
-    ASSERT(instr_same(instr, decin));
+    /* AUIPC only encodes the high 20-bit immediate value, so it is expected that the
```
abhinav92003, line 112: If the encoding pc remains the same and the immediate was at-most 20 bit long, I don't see why we'll get different instrs on re-encoding. Do you have an example?

suite/tests/api/ir_riscv64.expect
```diff
@@ -221,7 +221,7 @@ c.xor  a5 -> fp
 c.sub  a5 -> fp
 test_integer_arith complete
 lui    0x2a -> a0
-auipc  0x2a -> a0
+jalr   a1 0x2a -> a0
```
abhinav92003, line 224: Looks like the instrs tested by test_instr_encoding_pc_related are not in the expected output at all?

abhinav92003 (changes requested):

suite/tests/api/ir_riscv64.c
ksco: When we decode the recoded instr, we passed the `buf` as the new pc, that's how it got different. Ideally, decoding of auipc should always be given a pc that is aligned to `(1<<12)`, but I think that's not doable, right?

core/ir/riscv64/codec.c
```diff
@@ -1777,6 +1796,29 @@ encode_u_imm_opnd(instr_t *instr, byte *pc, int idx, uint32_t *out)
...
+{
+    opnd_t opnd = instr_get_src(instr, idx);
+    uint32_t imm;
+    if (opnd.kind == PC_kind)
+        imm = opnd_get_pc(opnd) - pc;
+    else if (opnd.kind == INSTR_kind)
+        imm = (byte *)opnd_get_instr(opnd)->offset - (byte *)instr->offset;
+    else
+        return false;
+    *out |= SET_FIELD(imm >> 12, 31, 12);
```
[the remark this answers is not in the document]
ksco: > Seems to me that the set of acceptable offsets are only the ones with nothing in their lower 12 bits.

Yes, if the lower 12 bits are non-zero, they will be thrown away.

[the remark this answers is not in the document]
abhinav92003: I mean to ask: should we throw an error if the bottom 12 bits are non-zero. Is that better than encoding an instr which the user did not intend to.

suite/tests/api/ir_riscv64.expect
abhinav92003: I understand. My question was: should we also add expected outputs for the instrs tested by test_instr_encoding_pc_related?

suite/tests/api/ir_riscv64.c
abhinav92003: For OP_auipc, we could use instr_encode_to_copy.

core/ir/riscv64/codec.c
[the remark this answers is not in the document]
ksco: Yeah, I think doing an assertion here is better, and the corresponding test also needs a rewrite.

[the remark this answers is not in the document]
ksco: If we do an assertion here, `instr_is_encoding_possible()` will stop working for auipc. :(

```
bool
encoding_possible(decode_info_t *di, instr_t *in, const instr_info_t *ii)
{
    uint enc;

    byte tmp[RISCV64_INSTR_SIZE];
    enc = encode_common(&tmp[0], in, di);
    return enc != ENCFAIL;
}
```

suite/tests/api/ir_riscv64.c
ksco: Fixed in https://github.com/DynamoRIO/dynamorio/pull/6208/commits/9ab092dfefeb8675dd0323ef2d14d6cbad1d84cc

core/ir/riscv64/codec.c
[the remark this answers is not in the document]
abhinav92003: Shouldn't encoding_possible also simply fail for OP_auipc if the provided offset has a non-zero bit in its last 12 bits?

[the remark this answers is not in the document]
abhinav92003: Oh, the instr doesn't have the offset as an operand anymore, it has the target pc. The offset is computed as the target_pc - encode_pc later.

@derekbruening the RISC-V auipc instr is a bit complex to encode. It should fail when the `offset & 0xfff != 0`. So it depends on the pc where it is being encoded. But encoding_possible doesn't take in the encode pc.

I guess there are rip-relative instrs on x86 that depend on the encode pc, but encoding_possible doesn't need to know the encode pc to decide whether or not it is encode-able, unlike here.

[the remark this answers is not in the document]
derekbruening: I believe that other arches like x86 do *not* check for reachability in encoding_possible and instead fail later in the full encode; I think that was not considered ideal but that is the current state.  I could see it either way: living with that behavior here, or trying to clean it up and passing an encode pc: but it may be complex if not every caller has an encode pc (I'm pretty sure some callers do not...like instr_zeroes_zmmh()?)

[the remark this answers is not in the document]
abhinav92003: So we could set `decode_info_t.check_reachable` to false in get_encoding_info (like we do for AARCHXX), and use it to skip the correctness check for the provided target pc.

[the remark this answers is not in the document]
ksco: > So we could set decode_info_t.check_reachable to false.

Currently, in `encode_u_immpc_opnd()`, we don't have `decode_info_t* di` available. So it requires a lot of changes in the encoder to pass it down. Should I do this in this PR or submit a new one?

```
static bool
encode_u_immpc_opnd(instr_t *instr, byte *pc, int idx, uint32_t *out);
````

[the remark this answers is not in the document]
Comment (abhinav92003) [reply core/ir/riscv64/codec.c:1818]: Sure. Let's create a new issue with the required details so someone can pick it up in future, and add a TODO here. I assume encode_u_immpc_opnd will simply drop the last 12 bits for now?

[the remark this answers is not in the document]
Comment (ksco) [reply core/ir/riscv64/codec.c:1821]: Added a FIXME in https://github.com/DynamoRIO/dynamorio/pull/6208/commits/7161172061eb98915096a7ba53cc03bed5e0a901.

File: suite/tests/api/ir_riscv64.c [hunk 4]
```diff
@@ -1082,45 +1082,45 @@ test_jump_and_branch(void *dc)
                              opnd_create_immed_int(42, OPSZ_20b));
     pc = test_instr_encoding(dc, OP_lui, instr);
     instr = INSTR_CREATE_auipc(dc, opnd_create_reg(DR_REG_A0),
-                               opnd_create_immed_int(42, OPSZ_20b));
-    test_instr_encoding(dc, OP_auipc, instr);
+                               opnd_create_pc(pc + (3 << 12)));
+    test_instr_encoding_pc_relative(dc, OP_auipc, pc, instr);
```
abhinav92003, line 1086: I don't see why the pc argument is related to the instruction being tested. Is it just an arbitrary pc we're using for the test? If yes, maybe test_instr_encoding_pc_relative doesn't need to take it as an arg and can generate one on its own.

abhinav92003 (approved):

suite/tests/api/ir_riscv64.c
```diff
@@ -103,11 +103,11 @@ test_instr_encoding_jump_or_branch(void *dc, uint opcode, instr_t *instr)
     byte *pc, *next_pc;
 
     ASSERT(instr_get_opcode(instr) == opcode);
-    ASSERT(instr_is_encoding_possible(instr));
-    pc = instr_encode(dc, instr, buf);
+    ASSERT(opcode == OP_auipc || instr_is_encoding_possible(instr));
```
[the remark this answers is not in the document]
ksco: Yes, removed.

suite/tests/api/ir_riscv64.c
ksco: It's mainly for AUIPC, we have to pass in a pc whose difference from the current pc is aligned to `1 << 12`. But considering we're not asserting that, it should be ok to use an arbitrary value for now.

abhinav92003: An arbitrary value is fine. My suggestion was that test_instr_encoding_pc_relative shouldn't need to take it as an arg.

```
 If yes, maybe test_instr_encoding_pc_relative doesn't need to take it as an arg and can generate one on its own.
```

abhinav92003: In fact, since we're not not asserting the alignment to 1<<12 currently anymore, do we even need to use instr_encode_to_copy? Can we just use instr_encode?

ksco: Yes, I understand that. It’s indeed cleaner without the arg. The point is once we added the assertion on AUIPC, the arg will be needed anyway, so it is no harm if we keep it as is, right?

abhinav92003: IMHO it's better to keep code consistent at each point. Right now, there's no need for the extra arg, so it may be confusing to a reader if we keep it.

ksco: Sorry I missed out on this, but I just find out that the `instr_pc` arg is necessary for `ASSERT(instr_same())` at line 112 to work.

abhinav92003: Do you mean we need instr_pc to be set to some specific argument for that assert to work?

ksco: Yes: 

```
    instr = INSTR_CREATE_auipc(dc, opnd_create_reg(DR_REG_A0),
                               opnd_create_pc(pc + (3 << 12)));
    test_instr_encoding_pc_relative(dc, OP_auipc, pc, instr);
```

PCs passed into `INSTR_CREATE_auipc()` and `test_instr_encoding_pc_relative()` has to align to `1<<12` so later in `test_instr_encoding_pc_relative()`, we can get the same instr back after encode and decode.

abhinav92003: Okay so we need it for the auipc test, but not for the others right? Maybe have some default value on the arg and   don't pass it when not needed? Also when it is not passed we can use instr_encode instead of instr_encode_to_copy.

ksco: How about we use `test_instr_encoding_jal_or_branch()` for JAL and branch instructions, and use `test_instr_encoding_auipc()` for AUIPC. And eventually change `test_instr_encoding_jal_or_branch()` to `test_instr_encoding_jal()` once disassemble issue on branchh is fixed?

ksco: I push said changes in https://github.com/DynamoRIO/dynamorio/pull/6208/commits/7ddef9cc7a2ead0db25aa994f9de4449046554e5.

abhinav92003 (approved):

PR #6209, opened 2023-07-17
i#3544 RV64: Implemented machine_cache_sync

ksco:
We need to flush the icache on all harts, which is not feasible for FENCE.I, so we use `SYS_riscv_flush_icache` to let the kernel do this. Note this is also how gcc implements `__builtin___clear_cache`.

### DIFF
--- a/core/arch/riscv64/proc.c
+++ b/core/arch/riscv64/proc.c
@@ -140,8 +140,11 @@ proc_has_feature(feature_bit_t f)
 void
 machine_cache_sync(void *pc_start, void *pc_end, bool flush_icache)
 {
-    /* FIXME i#3544: Not implemented. FENCE.I ? Maybe a syscall to clean icache? */
-    ASSERT_NOT_IMPLEMENTED(false);
+    /*
+     * RISC-V doesn't have an instruction to flush parts of the instruction cache,
+     * so instead we just flush the whole thing.
+     */
+    __asm__ __volatile__("fence.i" : : : "memory");
 }
 
 DR_API
--- a/core/arch/riscv64/proc.c
+++ b/core/arch/riscv64/proc.c
@@ -32,6 +32,11 @@
 
 #include "../globals.h"
 #include "proc.h"
+#ifdef UNIX
+#    include "../../unix/include/syscall.h"
+#else
+#    error NYI
+#endif
 
 static int num_simd_saved;
 static int num_simd_registers;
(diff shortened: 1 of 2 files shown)


core/arch/riscv64/proc.c
```diff
@@ -140,8 +140,11 @@ proc_has_feature(feature_bit_t f)
 void
 machine_cache_sync(void *pc_start, void *pc_end, bool flush_icache)
 {
-    /* FIXME i#3544: Not implemented. FENCE.I ? Maybe a syscall to clean icache? */
-    ASSERT_NOT_IMPLEMENTED(false);
+    /*
+     * RISC-V doesn't have an instruction to flush parts of the instruction cache,
+     * so instead we just flush the whole thing.
```
derekbruening, line 145: We also need to execute FENCE.I on all other cores?

Xref #1911 on similar topics for AArchXX (and x86).  Is there a membarrier parameter that executes FENCE.I on the other cores?  For x86 where it's not clear that anything is really needed and what's needed is just a "serializing instruction", the next trip into kernel code on each other core is likely to complete that operation; similarly for AArch64 it is likely that other cores will execute the barriers needed eventually, and when modifying branches at least they guarantee that only the old or new code will be observed.

What is the behavior for RISC-V?  Will the kernel typically execute FENCE.I on a context switch??  If not, no other core will ever see the change if we don't take action?

ksco: > We also need to execute FENCE.I on all other cores?

Thanks, I never thought about the multicore situation before. A single FENCE.i is not viable as the kernel might schedule the process onto another core.

 I looked into this, seems the only way we can achieve this is to call the syscall specially made for RISC-V: `sys_riscv_flush_icache`. This is also how glibc implements `__builtin___clear_cache `.

ksco: Done.

derekbruening: > A single FENCE.i is not viable as the kernel might schedule the thread onto another core.

The code being written is (mostly) thread-shared, so it's not just about this thread being migrated: we have to ensure other threads on other cores see the code change too.  This system call should do what we want.

core/arch/riscv64/proc.c
```diff
@@ -140,8 +145,12 @@ proc_has_feature(feature_bit_t f)
 void
 machine_cache_sync(void *pc_start, void *pc_end, bool flush_icache)
 {
-    /* FIXME i#3544: Not implemented. FENCE.I ? Maybe a syscall to clean icache? */
-    ASSERT_NOT_IMPLEMENTED(false);
+    /* We need to flush the icache on all harts, which is not feasible for FENCE.I, so we
+     * use SYS_riscv_flush_icache to let the kernel do this.
+     * The flag here is set to SYS_RISCV_FLUSH_ICACHE_LOCAL, which is defined as 1 in the
```
derekbruening, line 150: Let's make a #define (or const int) for SYS_RISCV_FLUSH_ICACHE_LOCAL and use it in the syscall here.

derekbruening (approved):

core/arch/riscv64/proc.c
derekbruening: It could just be local to this file for now; if we need it somewhere else we can move it to a header at that time.

ksco: Done in https://github.com/DynamoRIO/dynamorio/pull/6209/commits/89b268e727be5e5d7d09167c2c3f0ea914ef5e31.

PR #6211, opened 2023-07-17
RFC: encode instr_info_t::code as uint

dvyukov:
RFC: encode instr_info_t::code as uint

Change instr_info_t::code from ptr_int_t to uint.
This strips ~128K from drdecode binary and ~50K
from runtime VM overhead:

```
    FILE SIZE        VM SIZE
 --------------  --------------
   +21% +5.60Ki   +22% +5.61Ki    .rodata
  +1.5%    +720  [ = ]       0    .symtab
  +1.8%    +431  [ = ]       0    .strtab
  +0.5%    +384  [ = ]       0    [ELF Section Headers]
  +0.5%    +264  [ = ]       0    .rela.text
  +0.1%    +117  +0.1%    +137    .text
   +31%     +17  [ = ]       0    .llvm_addrsig
  +0.2%     +15  [ = ]       0    [Unmapped]
  +0.0%      +4  [ = ]       0    .rodata.str1.1
 -19.3% -61.3Ki -19.3% -61.3Ki    .data.rel.ro
 -31.0% -73.9Ki  [ = ]       0    .rela.data.rel.ro
 -11.5%  -127Ki -10.6% -55.6Ki    TOTAL
```

The idea is to change the pointer to table number
+ index in the table encoded as a single uint.
This removes the need in runtime relocations for pointers
(each taking 24 bytes in the binary).
Along with #6206 this should also shrink sizeof(instr_info_t)
from 48 bytes to 32 bytes (not fully realized w/o #6206 due
to alignment).

If this change looks good in general,
I will work on productionizing it and update other arches.
Format of the uint encoding is discus-sable.
For example, if we collect all instr_info_t into a single
global array (and then make individual tables pointers into it),
then the encoding can be made simpler and faster,
but will require more significant changes.


dvyukov:
CC @melver

derekbruening:
I discussed this with @abhinav92003 and we agree the change looks reasonable: the code readability impact is acceptable.  I would keep it as the two-level lookup rather than trying to make a single array.

derekbruening:
Maybe see if @khuey has any further input

Issue #725, opened 2014-11-28
attach feature on Windows


derekbruening:
_From [bruen...@google.com](https://code.google.com/u/109494838902877177630/) on April 09, 2012 14:58:26_

xref issue #38 : attach injection on Linux
xref issue #722 : internally-triggered attach

the problem with attach on windows of course is the inability to
maintain control across a callback return when the callback
occurred prior to attaching.

idea: have thin_client presence up front that hooks all alertable
syscalls.  risk: there are many alertable syscalls and maybe we
don't really know all of them.  probably best to hook all the
file i/o calls even though their alertability depends on their
args.

in absence of control up front, we could go ahead and try on windows and lose control briefly, counting on native_exec_syscalls like we do when we lose control on AppInit (though here there are multiple threads), and
just document the limitations to any users

_Original issue: http://code.google.com/p/dynamorio/issues/detail?id=725_

derekbruening:
9c95ba5aadd1dcc42b1e43161ad0193ca561a071 added enumeration and takeover of unknown threads, invoked by default on regular startup

OrBenPorath:
Added in #5075.
Test still requires works, and to be removed from the flaky list.



Test occasionally failing in one of two ways:
1. [Nothing happens until the test times out](https://github.com/DynamoRIO/dynamorio/runs/3685679149?check_suite_focus=true).
Unclear how that could happen, as one of the first things the target does is print something.
Maybe it doesn't even start?

2. [Test output missing 'MessageBox closed' before 'done'](https://github.com/DynamoRIO/dynamorio/pull/5075/checks?check_run_id=3684533264):
Unclear how the target terminates, as there is no FATAL_ERROR reported, and all normal termination paths have another print before "done".

derekbruening:
I thought the Windows detach via nudge was exposed today but it's not:

The existing Windows detach nudge feature is internal to DR and does not go
through a client.  It looks like it is not exposed in drconfig and the old
tests for it are not all enabled; the old way to trigger it is in the
no-longer-supported "drcontrol" front-end, but it's just a different nudge
type so it should be feasible to tweak drconfig to enable it to try it out.
You can see the code paths in libutil/detach.c triggering the nudge to be
sent and handling it in core/nudge.c "TEST(NUDGE_GENERIC(detach),
nudge_action_mask)" calling detach_helper() which calls the
shared-with-all-detach-types detach_on_permanent_stack().

[...]

PR #6221, opened 2023-07-23
i#725: Add an automated test of detach on Windows

onroadmuwl:
This patch adds an automated test of detach on Windows platform. The test start a target program and attach to the process firstly,  then use the drconfig front-end to detach from the process.
As shown in the .template file, the detach mechanism lets the target process be detached before stopping it.

Issue: [#725](https://github.com/DynamoRIO/dynamorio/issues/725)


suite/tests/runall.cmake
```diff
@@ -212,6 +212,32 @@ elseif ("${nudge}" MATCHES "<attach>")
...
+   )
+  # combine out and err
+  set(nudge_err "${nudge_out}${nudge_err}")
+  if (nudge_result)
+    kill_background_process(ON)
+    message(FATAL_ERROR "*** ${nudge_cmd} failed (${nudge_result}): ${nudge_err}***\n")
+  endif (nudge_result)
+  do_sleep(5000)
+  execute_process(COMMAND "${toolbindir}/drconfig.exe" "-detach" ${pid}
+  RESULT_VARIABLE detach_result
```
derekbruening, line 234: style: indent 2 spaces from line 233; ditto below for the other continuation lines

suite/tests/runall.cmake
```diff
@@ -212,6 +212,32 @@ elseif ("${nudge}" MATCHES "<attach>")
...
     message(FATAL_ERROR "*** ${nudge_cmd} failed (${nudge_result}): ${nudge_err}***\n")
   endif (nudge_result)
+elseif ("${nudge}" MATCHES "<detach>")
+  set(nudge_cmd run_in_bg)
+  string(REGEX REPLACE "<detach>"
+    "${toolbindir}/drrun@-attach@${pid}@-takeover_sleep@-takeovers@1000"
+    nudge "${nudge}")
+  string(REGEX REPLACE "@" ";" nudge "${nudge}")
+  execute_process(COMMAND "${toolbindir}/${nudge_cmd}" ${nudge}
+   RESULT_VARIABLE nudge_result
```
derekbruening, line 222: style: indent 2 spaces from line 233; ditto below for the other continuation lines

suite/tests/client-interface/detach_test.dll.c
```diff
@@ -0,0 +1,109 @@
...
+#endif
+
+static thread_id_t injection_tid;
+static bool first_thread = true;
+
+static void
+dr_exit(void)
+{
+#ifdef WINDOWS
+    dr_fprintf(STDERR, "detach\n");
```
derekbruening, line 47: The exit event could be for a regular app exit.  We want to ensure a detach happened.  Register a handler via dr_register_pre_detach_event to test this.

suite/runsuite_wrapper.pl
```diff
@@ -266,6 +267,7 @@
                 'code_api|tool.drcacheoff.burst_static' => 1, # i#4486
                 'code_api|tool.drcacheoff.burst_replace' => 1, # i#4486
                 'code_api|client.attach_test' => 1, # i#725
+                'code_api|client.detach_test' => 1, # i#725
```
derekbruening, line 270: We don't want to list tests here: we do it as a last resort for flaky tests to keep things green; listings here are supposed to be temporary.  Let's not have a new test start out on this list as that means it's just ignored from the start.  It seems the new test times out 3x in a row on both 32-bit and 64-bit: that means it's not working as intended.  The runall.cmake script waits for certain output, so that output must not be printed.  See below: is it "done"?

(Looks like the attach_test is here: that was supposed to be temporary, but the author never had time to fix it up I assume.  Looking at attach_test it seems to just have flaky or the wrong ordering of lines: I put that into #725.)

suite/tests/client-interface/detach_test.dll.c
derekbruening, line 47: runall.cmake waits for "done\n" -- is this why it's hanging, because you're not printing that?

derekbruening (commented):
Thank you for the PR.  It looks like the test is not fully working yet: let's get it working before merging.

suite/tests/win32/detachee.c
```diff
@@ -0,0 +1,61 @@
...
+{
+    /* We put the pid into the title so that tools/closewnd can target it
+     * uniquely when run in a parallel test suite.
+     * runall.cmake assumes this precise title.
+     */
+    char title[64];
+    _snprintf_s(title, BUFFER_SIZE_BYTES(title), BUFFER_SIZE_ELEMENTS(title),
+                "Infloop pid=%d", GetProcessId(GetCurrentProcess()));
+    SetTimer(NULL, 0, 180 * 1000 /*3 mins*/, TimerProc);
+    print("starting detachee\n");
```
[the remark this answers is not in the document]
onroadmuwl: I've replaced detachee with attachee.

suite/tests/client-interface/detach_test.dll.c
onroadmuwl: It's been fixed.

suite/runsuite_wrapper.pl
onroadmuwl: It's been fixed.

suite/tests/client-interface/detach_test.dll.c
onroadmuwl: I've add "done" into the dr_exit function, but vs2019-32 also complains Timeout.

suite/tests/runall.cmake
```diff
@@ -212,6 +212,32 @@ elseif ("${nudge}" MATCHES "<attach>")
     kill_background_process(ON)
     message(FATAL_ERROR "*** ${nudge_cmd} failed (${nudge_result}): ${nudge_err}***\n")
   endif (nudge_result)
+elseif ("${nudge}" MATCHES "<detach>")
+  set(nudge_cmd run_in_bg)
```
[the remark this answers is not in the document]
derekbruening: Could you elaborate: do you mean that drcontrol returns immediately after creating the remote thread and doesn't wait for it to complete?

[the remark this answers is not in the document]
onroadmuwl: If the attach operation is not running in the background, the` drrun -attach` command will continue to run without giving the following command line a chance to continue. The result of `client.detach_test-out`  is as follows：
```
starting attachee
thank you for testing detach
attach
thread init
```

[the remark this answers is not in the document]
derekbruening: OK so you mean that `drrun -attach` on Windows waits for the target process to exit.  OK, so it does have to be in the background if we keep that behavior.  Maybe best to add a comment b/c that wasn't clear to me.

suite/tests/client-interface/detach_test.dll.c
```diff
@@ -0,0 +1,116 @@
...
+
+static thread_id_t injection_tid;
+static bool first_thread = true;
+
+static void
+dr_exit(void)
+{
+#ifdef WINDOWS
+    dr_fprintf(STDERR, "done\n");
+#endif
```
derekbruening, line 48: Don't we want to print "done" for UNIX too?

suite/tests/runall.cmake
```diff
@@ -212,6 +212,32 @@ elseif ("${nudge}" MATCHES "<attach>")
...
+    ERROR_VARIABLE nudge_err
+    OUTPUT_VARIABLE nudge_out
+    )
+  # combine out and err
+  set(nudge_err "${nudge_out}${nudge_err}")
+  if (nudge_result)
+    kill_background_process(ON)
+    message(FATAL_ERROR "*** ${nudge_cmd} failed (${nudge_result}): ${nudge_err}***\n")
+  endif (nudge_result)
+  do_sleep(5000)
```
derekbruening, line 232: It would be better to wait on some output from the app so we know it's fully attached and we'll be doing a detach at the right point.  You could imagine a slow overloaded test VM still attaching after this 50ms.  Using sleeps like this tends to result in flaky tests.  In fact doesn't the script below here look for the "attach" and "thread init" output?  We want to wait for that before we detach, right?  This may be the source of your test failures.

suite/tests/CMakeLists.txt
```diff
@@ -1386,6 +1386,14 @@ function(torun test key source native standalone_dr dr_ops exe_ops added_out pas
       string(REGEX REPLACE ";" "@" nudge_arg "${nudge_arg}")
       # clear client from dr command and run native
       string(REGEX REPLACE ";-quiet;(.*);--" ";-no_inject;--;" rundr "${rundr}")
+    elseif ("${runall}" MATCHES "<detach>")
+      # move params to nudge
```
derekbruening, line 1390: style nit: We prefer capitalized sentences with end punctuation for all new comments: so "Move params to nudge.".  Ditto below.

suite/tests/client-interface/detach_test.dll.c
onroadmuwl: Yeah, I have remove ifdef WINDOWS

suite/tests/CMakeLists.txt
onroadmuwl: I have updated it

suite/tests/runall.cmake
onroadmuwl: Thanks for the heads up, I have modified it following your suggestions. The workflow of the test  looks more reasonable.  But the vs2019-32 test always complain timeout. Is it possible to add it into ignore-failures list?

derekbruening: Are you able to reproduce the 32-bit failure locally?

onroadmuwl: After my last commit, vs2019-32 work well on my laptop.

derekbruening: So it works locally, but hangs on GA (Github Actions).  One option is to use tmate to debug it on GA: https://github.com/marketplace/actions/debugging-with-tmate. It seems worth some effort to try to figure out why it's failing on GA, because it could indicate a real bug where detach for 32-bit doesn't work on a lot of setups and so users who try to use it might all hit hangs too.  If after some effort we can't figure it out then we can cut our losses and at least have the 64-bit test.  For the tmate what I would do is (temporarily) delete all the `.github/workflows/*.yml` files except the windows one and then edit that for tmate (and maybe to remove all the other tests).  (This assumes a PR from a clone will use the clone's yml files...I think so, once permissions are given by us to run our tests.)

onroadmuwl: I have solved this problem by chance. I commit [some test codes](https://github.com/DynamoRIO/dynamorio/pull/6221/commits/5b4262dde3f9dcd2dea1b12634369de9e9b15a7a) to output the detach's result, it shows that the attach and detach operation can be completed successfully. I guess that the target process can't be exit smoothly, so I [kill the process manually](https://github.com/DynamoRIO/dynamorio/pull/6221/commits/127834b102eaeebca5f7f1ec57f0689a96198460) at the end of runall.cmake, and then all the tests can pass.

suite/tests/runall.cmake
```diff
@@ -306,11 +345,10 @@ endwhile()
...
 
+# Sometimes infloop keeps running: FIXME: figure out why.
 if (UNIX)
-  # sometimes infloop keeps running: FIXME: figure out why
   execute_process(COMMAND "${KILL}" -9 ${pid} ERROR_QUIET OUTPUT_QUIET)
   # we can't run pkill b/c there are other tests running infloop (i#1341)
 else ()
-  # We could run "${toolbindir}/DRkill.exe" -pid ${pid} but we shouldn't need to
-  # as the app itself has a timeout.
+  execute_process(COMMAND "${toolbindir}/DRkill.exe" -pid ${pid} ERROR_QUIET OUTPUT_QUIET)
```
derekbruening, line 353: Hmm the closewnd.exe used by kill_background_process() should find the window: I wonder why it doesn't.  Maybe not worth further effort at this point.

onroadmuwl: Thanks for your guidance, I'll be working on detach for Linux in next phases, and maybe it‘ll start by studying the nudge mechanism first.

PR #6223, opened 2023-07-24
i#3544 RV64: Implemented cleanup_and_terminate and related macros

ksco:
This patch implements `cleanup_and_terminate` and related macros

- Implement `cleanup_and_terminate`
- Correct the value of `dstack_OFFSET`
- Similar to AARCHXX, add a `reg_stolen` field in `spill_state`, to save/restore dc stolen reg's value.

Issue: #3544


ksco:
Related: https://github.com/DynamoRIO/dynamorio/pull/6064#discussion_r1199157883, we're now using the x0 slot in dc for storing the original tp.

core/arch/riscv64/riscv64.asm
```diff
@@ -300,30 +300,84 @@ GLOBAL_LABEL(dynamorio_app_take_over:)
...
+        j        cat_no_thread
+cat_thread_only:
+        CALLC0(GLOBAL_REF(dynamo_thread_exit))
+cat_no_thread:
+        /* switch to d_r_initstack for cleanup of dstack */
+        la       s7, GLOBAL_REF(initstack_mutex)
+cat_spin:
+        li       a0, 1
+        amoswap.w a0, a0, (s7)
+        bnez     a0, cat_spin
```
derekbruening, line 345: Is there no yield or pause instruction?  If so, please add; if not, please add a comment saying we're missing what we'd like to do and what we do on other arches.

core/arch/riscv64/riscv64.asm
```diff
@@ -300,30 +300,84 @@ GLOBAL_LABEL(dynamorio_app_take_over:)
...
+
+        /* we have the spin lock */
+
+        /* switch stack */
+        la       a0, GLOBAL_REF(d_r_initstack)
+        ld       a0, 0(a0)
```
derekbruening, line 351: Is the 0 offset necessary to list?

core/arch/riscv64/riscv64.asm
```diff
@@ -300,30 +300,84 @@ GLOBAL_LABEL(dynamorio_app_take_over:)
...
+        mv       sp, a0
```
derekbruening, line 352: I thought RISC-V didn't have special registers: I would expect we can load directly into `sp`, unlike AArch64, and not need this extra move?

core/arch/riscv64/riscv64.asm
```diff
@@ -300,30 +300,84 @@ GLOBAL_LABEL(dynamorio_app_take_over:)
...
+
+        /* free dstack and call the EXIT_DR_HOOK */
+        CALLC1(GLOBAL_REF(dynamo_thread_stack_free_and_exit), s5) /* pass dstack */
+
+        /* give up initstack_mutex */
+        la       a0, GLOBAL_REF(initstack_mutex)
+        li       a1, 0
+        sd       a1, 0(a0)
```
derekbruening, line 360: Like above we can remove the 0 offset right?

core/arch/arch_exports.h
```diff
@@ -157,17 +157,30 @@ typedef struct _spill_state_t {
     reg_t reg_stolen; /* slot for the stolen register */
 #elif defined(RISCV64)
     reg_t a0, a1, a2, a3;
+    /* Slot for the "stolen" register.
+     * Note that on RISC-V, tp is the thread pointer, and it is also a general-purpose
+     * register, so unlike AARCHXX, we don't actually need to steal any register to store
+     * DR's tls, we can just use tp directly.
```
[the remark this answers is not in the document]
ksco: Note that on RISC-V, tp is the thread pointer, and it is also a general-purpose register, so we steal tp to store DR's tls.

Is this any better?

core/arch/arch_exports.h
```diff
@@ -157,17 +157,30 @@ typedef struct _spill_state_t {
...
+     * But this also introduces another issue, we will lose the original value of
+     * tp. So in insert_load_dr_tls_base() we put the original value of tp into the x0
+     * slot of the dcontext, and in append_save_gpr() we restore the original value back
+     * to tp.
+     * We'll also need to mangle any instructions that use tp.
```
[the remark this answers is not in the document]
ksco: If we use tp as the stolen register, it will be the only one we have to save & restore, I suppose it simplifies things, right?

core/arch/arch_exports.h
```diff
@@ -1051,18 +1069,19 @@ fill_with_nops(dr_isa_mode_t isa_mode, byte *addr, size_t size);
...
 #elif defined(RISCV64)
-/* FIXME i#3544: This can be 2B in C. */
+/* Instructions from C extensions are 2 bytes in size and we will never use this macro on
+ * these instructions. */
 #    define RISCV64_INSTR_SIZE 4
-/* FIXME i#3544: Not implemented */
-#    define FRAGMENT_BASE_PREFIX_SIZE(flags) RISCV64_INSTR_SIZE
-/* FIXME i#3544: Not implemented */
+#    define RISCV64_INSTR_COMPRESSED_SIZE 2
+#    define FRAGMENT_BASE_PREFIX_SIZE(flags) RISCV64_INSTR_SIZE * 2
```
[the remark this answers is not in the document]
ksco: > is this meant to go with a different PR?

Yes, I'm trying to keep the PR size small, is this proper?

core/arch/arch_exports.h
```diff
@@ -1051,18 +1069,19 @@ fill_with_nops(dr_isa_mode_t isa_mode, byte *addr, size_t size);
...
 #    define DIRECT_EXIT_STUB_SIZE(flags) \
-        (10 * RISCV64_INSTR_SIZE) /* see insert_exit_stub_other_flags */
+        (16 * RISCV64_INSTR_SIZE) /* see insert_exit_stub_other_flags() */
```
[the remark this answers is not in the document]
ksco: Yes.

core/arch/riscv64/riscv64.asm
```diff
@@ -300,30 +300,84 @@ GLOBAL_LABEL(dynamorio_app_take_over:)
...
+        mv       s2, ARG3  /* sys_arg1 */
+        mv       s3, ARG4  /* sys_arg2 */
+        mv       s4, ARG5  /* exitproc */
+                          /* s5 reserved for dstack ptr */
+                          /* s6 reserved for syscall ptr */
+
+        li       a1, 1
+        /* inc exiting_thread_count to avoid being killed once off all_threads list */
+        la       a0, GLOBAL_REF(exiting_thread_count)
+        amoadd.w zero, a1, (a0)
```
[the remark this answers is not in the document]
ksco: Yes, according to the golden model implementation[1], it works with `zero` as the destination register.

1. https://github.com/riscv-software-src/riscv-isa-sim/blob/master/riscv/insns/amoadd_w.h

core/arch/riscv64/riscv64.asm
ksco: No, it's unnecessary, but it's unusual to write it without the 0.

core/arch/riscv64/riscv64.asm
ksco: Sorry for copy-pasting and didn't catch that, fixed.

core/arch/riscv64/riscv64.asm
ksco: We don't have a YIELD-like hint instruction like what's available on Aarch64, so we use a plain spin lock here.

Is this comment good?

core/arch/riscv64/riscv64.asm
ksco: Ditto.

core/arch/riscv64/riscv64.asm
```diff
@@ -300,30 +300,84 @@ GLOBAL_LABEL(dynamorio_app_take_over:)
...
- *                       (2 more args that are ignored: Mac-only))
+ *                       bool exitproc)                   // A4
  *
  * See decl in arch_exports.h for description.
  */
         DECLARE_FUNC(cleanup_and_terminate)
 GLOBAL_LABEL(cleanup_and_terminate:)
-/* FIXME i#3544: Not implemented */
-        jal       GLOBAL_REF(unexpected_return)
+        /* move argument registers to callee saved registers */
```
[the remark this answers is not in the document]
ksco: Done.

ksco (commented):
Thanks for the detailed review!

core/arch/arch_exports.h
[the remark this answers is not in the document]
derekbruening: While in the code cache I don't think using tp is any different from another register (unless tp is used less frequently than any other register by app code).  The place where it may save work is transitioning back to DR system code if the same value is used for DR's C code.  But code cache operation is what should be optimized for so if some other register is used a lot less that might outweigh the cxt sw savings: cxt switches are already really expensive.  All that said: it seems fine to go with tp for now.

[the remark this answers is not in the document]
derekbruening: I guess on a clean call too: that may be more frequent.

core/arch/arch_exports.h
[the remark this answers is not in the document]
derekbruening: I'm guessing you had a branch with multiple pieces and later split it up and it was easiest to split by file and just leave these couple of defines in this PR instead of pulling the changes out of this already-modified file.  Since the RISC-V code is all still a work in progress and we're filling in many things this seems ok to not have a perfect separation so let's leave it.

core/arch/riscv64/riscv64.asm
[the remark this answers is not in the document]
derekbruening: OK.

core/arch/riscv64/riscv64.asm
derekbruening: Sure.

core/arch/riscv64/riscv64.asm
derekbruening: OK, a RISC-V convention then: but note that the amoadd and amoswap have it without the 0: should you add a 0 there?

core/arch/arch_exports.h
```diff
@@ -157,14 +157,12 @@ typedef struct _spill_state_t {
...
+    /* Slot for the stolen register, which is tp.
      * Note that on RISC-V, tp is the thread pointer, and it is also a general-purpose
-     * register, so unlike AARCHXX, we don't actually need to steal any register to store
-     * DR's tls, we can just use tp directly.
-     * But this also introduces another issue, we will lose the original value of
-     * tp. So in insert_load_dr_tls_base() we put the original value of tp into the x0
-     * slot of the dcontext, and in append_save_gpr() we restore the original value back
-     * to tp.
+     * register, so we steal tp to store DR's tls.
+     * But this also introduces another issue, we will lose the original value of tp. So
```
derekbruening, line 163: This sentence can be removed; we're stealing the register which implies we're properly preserving the app's value.

core/arch/arch_exports.h
```diff
@@ -157,14 +157,12 @@ typedef struct _spill_state_t {
...
+     * in insert_load_dr_tls_base() we put the original value of tp into the x0 slot of
```
derekbruening, line 164: Into x0?  That seems to be a recipe for confusion, to put one register into the slot meant for a different register.  I would vote for sticking to the designated slots.  Also, these changes being referenced are not in this PR: xref the other conversations; here it might be best to remove this sentence from this PR so that when the future PR that changes insert_load_dr_tls_base() is reviewed this comment is there in the diff view to read and it is easy to know to change it if the behavior is changed because it's all part of the same PR.

derekbruening (approved):

core/arch/arch_exports.h
ksco: Agree, removed.

core/arch/arch_exports.h
ksco: Removed.

PR #6227, opened 2023-07-25
Swap the order of the two lines in attach test's .template file

onroadmuwl:
Fixes incorrect output ordering for the client.attach_test.
Removes the client.attach_test from the ignore-failure list in the wrapper script as it now passes for 32-bit and 64-bit Windows.

Issue: [#725](https://github.com/DynamoRIO/dynamorio/issues/725)


derekbruening (approved):
Nice, now it is passing.  Thank you for the fix.

suite/tests/client-interface/attach_test.template
```diff
@@ -4,8 +4,8 @@ starting attachee
 starting
 #endif
 thank you for testing attach
-thread init
 event_post_attach
+thread init
```
derekbruening, line 8: Oh ok you're saying the Linux ones fail with the new order.  You can use `#ifdef WINDOWS` (just like below on line 9) to make the two different orders.

derekbruening (changes requested):

suite/tests/client-interface/attach_test.template
onroadmuwl: fix it.

derekbruening (approved):

Issue #5365, opened 2022-02-17
Build and validate DynamoRIO on AArch64 SVE hardware


AssadHashmi:
We need to fix build and runtime issues now that SVE support is becoming available on AArch64 hardware.

This ticket should only track work on the current master which has limited SVE instruction support.

Other issues should track the addition of full SVE and later SVE2 instruction support, e.g. #3044.

AssadHashmi:
User issue raised when running on A64FX https://groups.google.com/g/dynamorio-users/c/_7H9NZXh3wc

derekbruening:
#5036 covers expanding scatter/gather instructions for easier instrumentation

PR #6230, opened 2023-07-26
i#3044 AArch64 SVE codec: change LDR/STR and PRF to use byte offsets

AssadHashmi:
For the current decode/encode functions of:
```
LDR <Zt>, [<Xn|SP>{, #<imm>, MUL VL}]
LDR <Pt>, [<Xn|SP>{, #<imm>, MUL VL}]
STR <Zt>, [<Xn|SP>{, #<imm>, MUL VL}]
STR <Pt>, [<Xn|SP>{, #<imm>, MUL VL}]
PRFB <prfop>, <Pg>, [<Xn|SP>{, #<imm>, MUL VL}]
PRFH <prfop>, <Pg>, [<Xn|SP>{, #<imm>, MUL VL}]
PRFW <prfop>, <Pg>, [<Xn|SP>{, #<imm>, MUL VL}]
PRFD <prfop>, <Pg>, [<Xn|SP>{, #<imm>, MUL VL}]
```
Vector indexing is used in the memory operand at the IR level. However
the IR must always refer to the address in terms of the base register
value plus a byte offset displacement. This patch changes the
decode/encode functions for these instructions to expect byte offsets
at the IR level, converting to vector length offsets within the codec.

Issues #3044, #5365

### DIFF
--- a/core/ir/aarch64/codec.c
+++ b/core/ir/aarch64/codec.c
@@ -5178,9 +5178,19 @@ static inline bool
 decode_opnd_svemem_gpr_simm6_vl(uint enc, int opcode, byte *pc, OUT opnd_t *opnd)
 {
     const int offset = extract_int(enc, 16, 6);
+    IF_RETURN_FALSE(offset < -32 || offset > 31)
     const reg_id_t rn = decode_reg(extract_uint(enc, 5, 5), true, true);
     const opnd_size_t mem_transfer = op_is_prefetch(opcode) ? OPSZ_0 : OPSZ_SVE_VL;
-    *opnd = opnd_create_base_disp(rn, DR_REG_NULL, 0, offset, mem_transfer);
+
+    /* As specified in the AArch64 SVE reference manual for contiguous prefetch
+     * instructions, the immediate index value is a vector index into memory, NOT
+     * an element or byte index. In DynamoRIO's IR, base-displacement operands
+     * should always refer to the address as a base register value + the linear
+     * memory displacement. So when creating the address operand here, it should be
+     * multiplied by the current vector register length in bytes.
+     */
+    int vl_bytes = dr_get_sve_vl() / 8;
+    *opnd = opnd_create_base_disp(rn, DR_REG_NULL, 0, offset * vl_bytes, mem_transfer);
 
     return true;
 }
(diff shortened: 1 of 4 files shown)


core/ir/aarch64/codec.c
```diff
@@ -5193,9 +5203,20 @@ encode_opnd_svemem_gpr_simm6_vl(uint enc, int opcode, byte *pc, opnd_t opnd,
...
         return false;
+    if (!reg_is_gpr(opnd_get_base(opnd)))
+        return false;
+
+    /* As described in decode_opnd_svemem_gpr_simm6_vl(), disp is a multiple of
+     * vector length at the IR level, transformed to a vector index in the
+     * encoding.
+     */
+    int vl_bytes = dr_get_sve_vl() / 8;
+    ASSERT((opnd_get_disp(opnd) % vl_bytes) == 0);
```
derekbruening, line 5214: Best to just return false and fail the encoding like all the other un-encodable cases checked in this routine, rather than assert, if it's not an even multiple.

core/ir/aarch64/codec.c
```diff
@@ -5321,16 +5355,29 @@ encode_opnd_svemem_gpr_simm9_vl(uint enc, int opcode, byte *pc, opnd_t opnd,
...
-    if (!encode_reg(&rn, &is_x, opnd_get_base(opnd), true) || !is_x)
-        return false;
+    /* As described in decode_opnd_svemem_gpr_simm9_vl(), disp is a multiple of
+     * vector or predicate length at the IR level, transformed to a vector or
+     * predicate index in the encoding.
+     */
+    int vl_bytes = dr_get_sve_vl() / 8;
+    int pl_bytes = vl_bytes / 8;
+    if (is_vector)
+        ASSERT((opnd_get_disp(opnd) % vl_bytes) == 0);
```
derekbruening, line 5372: Like above I would have both of these just return false instead of asserting.

suite/tests/api/ir_aarch64_sve.c
```diff
@@ -8303,23 +8303,23 @@ TEST_INSTR(compact_sve)
 TEST_INSTR(str)
 {
     /* Testing STR <Zt>, [<Xn|SP>{, #<simm>, MUL VL}] */
-    int simm_0[6] = { 0, 255, -256, 127, -128, -1 };
+    int simm_0[6] = { 0, 32, -32, 8160, -4096, -8192 };
```
derekbruening, line 8306: Before it seemed to be checking boundary conditions: but the new numbers don't look like the same conditions?  Or at least they aren't a simple fixed scale-up.  Ditto below.  How did you come up with the new numbers?

derekbruening (approved):

core/ir/aarch64/codec.c
AssadHashmi: Done.

core/ir/aarch64/codec.c
AssadHashmi: Done.

suite/tests/api/ir_aarch64_sve.c
AssadHashmi: For test vector length of 256 bits the `simm_0` array does include boundary values:
255 x 32 bytes -> 8160
-256 x 32 bytes -> -8192

Updated two test values for `STR <Pt>, [<Xn|SP>{, #<simm>, MUL VL}]` and `LDR <Zt>, [<Xn|SP>{, #<simm>, MUL VL}]` which weren't boundary values.

derekbruening: OK, I was just expecting the same scale factor in the diff: so if -256 became -32, -128 should become -16, but it became -4096 instead.

derekbruening (approved):

Issue #6164, opened 2023-06-22
Add a new file type for traces which contain a mix of filtered and unfitered records.


prasun3:
**Add a pointer to any prior users list discussion.**
Traces can have a mix of filtered and unfiltered records (see #5675 #5199). Currently, we assume OFFLINE_FILE_TYPE_{D,I}FILTERED after seeing TRACE_MARKER_TYPE_FILTER_ENDPOINT. A new file type suggested by @derekbruening in #5717 will make this more explicit.

**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is.  E.g., "I'm always frustrated when [...]"

**Describe the solution you'd like**
Describe the desired functionality and its intended usage to give some context for how it would be used.

Do you have any implementation in mind for this feature?

**Describe alternatives you've considered**
Is there currently a method for implementing this functionality using the existing API and this requested feature will add convenience?  Or is there no method for accomplishing the desired task using the current API?

**Additional context**

PR #6231, opened 2023-07-26
i#6164: Add new file type for bimodal filtered traces

abhinav92003:
Adds OFFLINE_FILE_TYPE_BIMODAL_FILTERED_WARMUP for traces that are filtered before the TRACE_MARKER_TYPE_FILTER_ENDPOINT marker and unfiltered after it.

Adds invariant_checker checks to make sure the new file type and the filter end-point marker are
always present together, or not at all.

Adds a test that the record filter adds the filter end-point marker also for threads that start after
the filtering stop_timestamp. Adds a shard_count member to basic_counts_t::counters_t to make
it easier to write the new test without hard-coding count of threads in the test.

Fixes: #6164

### DIFF
--- a/clients/drcachesim/common/trace_entry.h
+++ b/clients/drcachesim/common/trace_entry.h
@@ -808,6 +808,14 @@ typedef enum {
      * The included kernel traces are in the Intel® Processor Trace format.
      */
     OFFLINE_FILE_TYPE_KERNEL_SYSCALLS = 0x1000,
+    /**
+     * Partially filtered trace. The initial part upto the
+     * #TRACE_MARKER_TYPE_FILTER_ENDPOINT marker is filtered, and the later part is not.
+     * Look for other filtering-related file types (#OFFLINE_FILE_TYPE_IFILTERED and
+     * #OFFLINE_FILE_TYPE_DFILTERED) to determine how the initial part was filtered.
+     * The initial part can be used by a simulator for warmup.
+     */
+    OFFLINE_FILE_TYPE_BIMODAL_FILTERED_WARMUP = 0x2000,
 } offline_file_type_t;
 
 static inline const char *
--- a/clients/drcachesim/common/trace_entry.h
+++ b/clients/drcachesim/common/trace_entry.h
@@ -474,9 +474,11 @@ typedef enum {
     TRACE_MARKER_TYPE_RECORD_ORDINAL,
 
     /**
-     * Indicates a point in the trace where filtering ended.
-     * This is currently added by the record_filter tool to annotate when the
-     * warmup part of the trace ends.
+     * Indicates the point in the trace where filtering ended. It is accompanied by
+     * #dynamorio::drmemtrace::OFFLINE_FILE_TYPE_BIMODAL_FILTERED_WARMUP in the file
+     * type. This is added by the record_filter tool and also by the drmemtrace tracer
+     * (with -L0_filter_until_instrs) to annotate when the warmup part of the trace
+     * ended.
      */
     TRACE_MARKER_TYPE_FILTER_ENDPOINT,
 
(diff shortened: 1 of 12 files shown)


clients/drcachesim/common/trace_entry.h
```diff
@@ -808,6 +808,14 @@ typedef enum {
      * The included kernel traces are in the Intel® Processor Trace format.
      */
     OFFLINE_FILE_TYPE_KERNEL_SYSCALLS = 0x1000,
+    /**
+     * Partially filtered trace. The initial part upto the
+     * #TRACE_MARKER_TYPE_FILTER_ENDPOINT marker is filtered, and the later part is not.
```
derekbruening, line 813: Add a note in the _ENDPOINT marker docs that it always goes with this filetype.

derekbruening (approved):
Maybe should re-review after invariant addition if desired.

clients/drcachesim/common/trace_entry.h
```diff
@@ -808,6 +808,14 @@ typedef enum {
      * The included kernel traces are in the Intel® Processor Trace format.
      */
     OFFLINE_FILE_TYPE_KERNEL_SYSCALLS = 0x1000,
+    /**
+     * Partially filtered trace. The initial part upto the
```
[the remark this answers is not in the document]
abhinav92003: Done.

clients/drcachesim/tools/filter/cache_filter.cpp
```diff
@@ -1,5 +1,5 @@
 /* **********************************************************
- * Copyright (c) 2022 Google, Inc.  All rights reserved.
+ * Copyright (c) 2022-2023 Google, Inc.  All rights reserved.
```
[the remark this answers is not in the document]
abhinav92003: Oops. There were initially. Removed.

clients/drcachesim/common/trace_entry.h
```diff
@@ -808,6 +808,14 @@ typedef enum {
      * The included kernel traces are in the Intel® Processor Trace format.
      */
     OFFLINE_FILE_TYPE_KERNEL_SYSCALLS = 0x1000,
+    /**
```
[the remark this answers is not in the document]
abhinav92003: Done.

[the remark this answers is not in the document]
abhinav92003: Done.

clients/drcachesim/common/trace_entry.h
abhinav92003: Done.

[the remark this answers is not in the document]
abhinav92003: Done. Also added a test for the record_filter that if stop_timestamp is earlier than the first timestamp in the trace, we still emit the endpoint marker. Also the drcachesim op_L0_filter_until_instrs also makes sure to emit that marker even for new threads that start after filtering stopped: https://github.com/DynamoRIO/dynamorio/blob/62e9995ecf25df3e1ae34823f47a66b177f9c989/clients/drcachesim/tracer/output.cpp#L1273.

abhinav92003:
Made a few changes to other analysis tools (record_filter and basic_counts), so requesting review again.

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -412,9 +412,14 @@ test_null_filter()
         std::unique_ptr<dynamorio::drmemtrace::record_filter_t::record_filter_func_t>>
         filter_funcs;
     filter_funcs.push_back(std::move(null_filter));
+    // We use a very small stop_timestamp for the record filter. This is to verify that
+    // we emit the TRACE_MARKER_TYPE_FILTER_ENDPOINT marker for each thread even if
+    // starts after the given stop_timestamp. Since the stop_timestamp is so small, all
```
derekbruening, line 417: nit: s/starts/it starts/

clients/drcachesim/tools/basic_counts.h
```diff
@@ -182,6 +182,12 @@ class basic_counts_t : public analysis_tool_t {
         int_least64_t encodings = 0;
         std::unordered_set<uint64_t> unique_pc_addrs;
 
+        // Metadata for the counts. These are not used for the equality, increment,
+        // or decrement operation, and must be set explicitly.
+
+        // Count of shards that were combined to produce the above counts.
+        int_least64_t shard_count = 1;
```
derekbruening, line 189: This should be in the operator+=, operator-=, operator== too right?

derekbruening (approved):

clients/drcachesim/tools/basic_counts.h
abhinav92003: I documented above that this (and any future) metadata vars would not be included in those ops. It's not as simple as counts. E.g. we += counters_t not only across shards but across windows too.

clients/drcachesim/tests/record_filter_unit_tests.cpp
abhinav92003: Done.

Issue #6185, opened 2023-07-05
api.rseq failing with "Failed to synchronize with all threads when detaching"


derekbruening:
This happened once on x86-64-ubuntu22:
https://github.com/DynamoRIO/dynamorio/actions/runs/5467288258/jobs/9953386272
```
442: <Application /home/runner/work/dynamorio/dynamorio/build_debug-internal-64/suite/tests/bin/api.rseq (19827). Failed to synchronize with all threads when detaching.>
```

Xref #4923

derekbruening:
Same issue on burst_threads test https://github.com/DynamoRIO/dynamorio/actions/runs/16324303569/job/46110186612
```
373: CMake Error at /home/runner/work/dynamorio/dynamorio/suite/tests/process_cmdline.cmake:109 (message):
373:   *** cmd failed (40): pre-DR init
373: 
373:   pre-DR start
373: 
373:   pre-DR detach
373: 
373:   pre-DR init
373: 
373:   pre-DR start
373: 
373:   pre-DR detach
373: 
373:   pre-DR init
373: 
373:   pre-DR start
373: 
373:   pre-DR detach
373: 
373:   pre-DR init
373: 
373:   pre-DR start
373: 
373:   <Application
373:   /home/runner/work/dynamorio/dynamorio/build_debug-internal-32/clients/bin32/tool.drcacheoff.burst_threads
373:   (25889).  Failed to synchronize with all threads when detaching.>
373: 
373:   ***
373: 
373: Call Stack (most recent call first):
373:   /home/runner/work/dynamorio/dynamorio/suite/tests/runmulti.cmake:57 (process_cmdline)
373: 
373: 
354/424 Test #373: code_api|tool.drcacheoff.burst_threads ...........................***Failed   64.69 sec
```

derekbruening:
Merging this into #5908.

PR #6236, opened 2023-08-01
i#5843 scheduler: Fix record-replay ordinal mismatches

derekbruening:
Adds read-ahead to timestamps when replaying, to match record ordinals and not just instruction ordinals at start and switch points when recording.

Adds use of the just-read timestamp when skipping in the reader, to avoid losing the timestamp when the scheduler has read ahead.

Adds recording of the ordinals at switch points in the unit tests for comprehensive testing of switch points instead of just testing thread sequences.

Fixes several tests which were passing incorrect flags.

Issue: #5843

### DIFF
--- a/clients/drcachesim/reader/reader.cpp
+++ b/clients/drcachesim/reader/reader.cpp
@@ -392,7 +392,7 @@ reader_t::skip_instructions_with_timestamp(uint64_t stop_instruction_count)
     // instr, so we look for the 1st unskipped instr: but we do not want to
     // process it so we do not use the ++ operator function.
     uint64_t stop_count = stop_instruction_count + 1;
-    trace_entry_t timestamp;
+    trace_entry_t timestamp = {};
     // Use the most recent timestamp.
     if (last_timestamp_ != 0) {
         timestamp.type = TRACE_TYPE_MARKER;
--- a/clients/drcachesim/reader/reader.cpp
+++ b/clients/drcachesim/reader/reader.cpp
@@ -79,7 +79,7 @@ reader_t::operator++()
             // We've already presented the thread exit entry to the analyzer.
             continue;
         }
-        VPRINT(this, 4, "RECV: type=%s (%d), size=%d, addr=0x%zx\n",
+        VPRINT(this, 5, "RECV: type=%s (%d), size=%d, addr=0x%zx\n",
                trace_type_names[input_entry_->type], input_entry_->type,
                input_entry_->size, input_entry_->addr);
         if (process_input_entry())
@@ -392,7 +392,13 @@ reader_t::skip_instructions_with_timestamp(uint64_t stop_instruction_count)
     // instr, so we look for the 1st unskipped instr: but we do not want to
     // process it so we do not use the ++ operator function.
     uint64_t stop_count = stop_instruction_count + 1;
-    trace_entry_t timestamp = {};
+    trace_entry_t timestamp;
+    // Use the most recent timestamp.
+    if (last_timestamp_ != 0) {
+        timestamp.type = TRACE_TYPE_MARKER;
+        timestamp.size = TRACE_MARKER_TYPE_TIMESTAMP;
+        timestamp.addr = static_cast<addr_t>(last_timestamp_);
+    }
     trace_entry_t cpu = {};
     trace_entry_t next_instr = {};
     bool prev_was_record_ord = false;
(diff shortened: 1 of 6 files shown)


derekbruening:
ubuntu22 failure is api.rseq #6185

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -1049,6 +1055,8 @@ scheduler_tmpl_t<RecordType, ReaderType>::get_initial_timestamps()
                 RecordType record = **input.reader;
                 if (record_type_is_timestamp(record, input.next_timestamp))
                     break;
+                if (record_type_is_instr(record))
```
abhinav92003, line 1058: So if we find an instr before a timestamp we return STATUS_ERROR_INVALID_PARAMETER, instead of just reading the next timestamp? Explain why in comment.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -1715,6 +1723,8 @@ scheduler_tmpl_t<RecordType, ReaderType>::next_record(output_ordinal_t output,
                 record = **input->reader;
             }
         }
+        VPRINT(this, 5, "next_record_mid[%d]: from %d: ", output, input->index);
```
abhinav92003, line 1726: Not clear what next_record_mid means. Was it for your local debugging?

clients/drcachesim/reader/reader.cpp
```diff
@@ -79,7 +79,7 @@ reader_t::operator++()
             // We've already presented the thread exit entry to the analyzer.
             continue;
         }
-        VPRINT(this, 4, "RECV: type=%s (%d), size=%d, addr=0x%zx\n",
+        VPRINT(this, 5, "RECV: type=%s (%d), size=%d, addr=0x%zx\n",
```
abhinav92003, line 82: In the PR description:

>  when the scheduler has read ahed.

s/ahed/ahead/

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -1732,6 +1742,11 @@ scheduler_tmpl_t<RecordType, ReaderType>::next_record(output_ordinal_t output,
                 // The stop is exclusive.  0 does mean to do nothing (easiest
                 // to have an empty record to share the next-entry for a start skip
                 // or other cases).
+                VPRINT(this, 5,
+                       "next_record_cmp[%d]: for %d: at instr#=%" PRId64
```
abhinav92003, line 1746: Wouldn't be clear to a reader what next_record_cmp is.

clients/drcachesim/scheduler/scheduler.cpp
abhinav92003, line 1058: > Adds read-ahead to timestamps when replaying, to match ordinals when recording.

I was trying to figure out how issues in timestamp read-ahead can cause ordinal mismatches. Is L1058 related to it?

abhinav92003 (approved):

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: We assume there are no instructions in our queue in skip_instructions().  Plus this means the trace is malformed in some way, so the invalid param fits.  Added comments.

clients/drcachesim/reader/reader.cpp
derekbruening: Done.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Yes, was going to just delete it, but thought it may be useful next time.  I replaced _mid with a separate description "candidate record from %d".  This is the possible next record which is examined to see if we should switch, when it is queued.  Useful to see what it is for some debugging.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Deleted this one; probably not as useful.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: I wanted to be able to compare both instr and record ordinals at start and context switch points in the input streams, and the read-ahead during record but not during replay made the input stream reported ordinals not match in tests.  (Would be better to have the records queued in the scheduler reflected in the input stream ordinals but that is not simple.)

derekbruening: Updated the PR description on this.

Issue #5994, opened 2023-04-22
Add to opcode_mix analyzer: branch density; memory density; arithmetic intensity


derekbruening:
Some common metrics for workload characterization are based purely on opcodes and would fit in as additions to the existing opcode_mix analyzer:

+ branch density
+ memory density: we have loads+stores and total instrs from the basic_counts tool but this fits in with the other density metrics here
+ arithmetic intensity: FLOPS vs memops; integer too?  DR's API identifies floating point sub-types (wondering whether this is implemented for aarch64; it is there for x86): https://source.corp.google.com/piper///depot/google3/third_party/dynamorio/trunk/core/ir/instr_api.h;l=1816?q=instr_api.h

derekbruening:
We should augment opcode_mix to use the new opcode categories (instr_get_category()) to produce counts of each category (note that x86 categories are not 100% filled in for every opcode yet; that will happen over time).  The categories can then be used to implement the metrics in the original post.

derekbruening:
Adding quick link to the current category list: https://dynamorio.org/dr__ir__instr_8h.html#a6a293267f240f04a49586895b3b67893

PR #5835, opened 2023-01-24
i#5365: Add AArch64 SVE support to the core (part 1)

AssadHashmi:
This patch adds Arm AArch64 Scalable Vector Extension (SVE) support to
the core including related changes to the codec, IR and relevant clients. 

SVE and SVE2 are major extensions to Arm's 64 bit architecture.
Developers and users should reference the relevant documentation at
developer.arm.com, (currently
https://developer.arm.com/Architectures/Scalable%20Vector%20Extensions).

The architecture allows hardware implementations to support vector
lengths from 128 to 2048 bits. This patch supports up to 512 bits due
to DynamoRIO's stack size limitation. There is currently no stock SVE
hardware with vector lengths greater than 512 bits. The vector length
is determined by get_processor_specific_info() at runtime on startup
and is available by calling proc_get_vector_length(). For Z registers,
reg_get_size() will return the vector size implemented by the hardware
rather than OPSZ_SCALABLE.

There will be follow up patches for:
- SVE scatter/gather emulation
- Full SVE signal context support
- Complete SVE support in sample clients and drcachesim tracer.

Issues: #5365, #3044

PR #6237, opened 2023-08-02
Add categorization of AARch64 instruction mix

kuhanov:
Added categorization for AARch64 instruction mix.
Each opcode is set to categories from the following list:
 - Integer arithmetic operations
 - Floating-Point arithmetic operations
 - Loads 
 - Stores
 - Branches
 - Operations with vector registers (SIMD) 
 - Other types of instructions

Calculation of category is based on C4.1.67 of ARM V8 Architecture reference manual
     *  https://developer.arm.com/documentation/ddi0487/

Also, we could add our categorization for x86_64 for issue #5994. Unfortunantly I don't have access to mentioned file https://source.corp.google.com/piper///depot/google3/third_party/dynamorio/trunk/core/ir/instr_api.h;l=1816?q=instr_api.h to check it.

Thx, Kirill


derekbruening:
> Also, we could add our categorization for x86_64 for issue #5994. Unfortunantly I don't have access to mentioned file https://source.corp.google.com/piper///depot/google3/third_party/dynamorio/trunk/core/ir/instr_api.h;l=1816?q=instr_api.h to check it.

Bad link: updated it to https://github.com/DynamoRIO/dynamorio/blob/master/core/ir/instr_api.h#L1845, just points at instr_is_floating().  There is also instr_is_floating_ex() with dr_fp_type_t: https://dynamorio.org/dr__ir__instr_8h.html#a6a279ef5b19e7c4f82f8804d9490da54

There is also a proposal to provide categories of opcodes for https://dynamorio.org/google_workload_traces.html which may overlap with an API like this; not yet filed in the public tracker -- will do so.

derekbruening:
> Calculation of category is based on C4.1.67 of ARM V8 Architecture reference manual * https://developer.arm.com/documentation/ddi0487/

Is that an older version?  The latest PDF at that link has "C4.1.67 SVE2 Histogram Computation - Segment" so it must be a different section in your version: what is the title?

core/ir/category_api.h
```diff
@@ -0,0 +1,15 @@
...
+
+typedef enum _dr_instr_category_type_t {
+    DR_INSTR_CATEGORY_UNCATEGORIZED = 0x0, /* Uncategorised. */
+    DR_INSTR_CATEGORY_MATH_INT = 0x1,      /* Integer arithmetic operations */
+    DR_INSTR_CATEGORY_MATH_FLOAT = 0x2,    /* Floating-Point arithmetic operations */
+    DR_INSTR_CATEGORY_LOAD = 0x4,          /* Loads */
+    DR_INSTR_CATEGORY_STORE = 0x8,         /* Stores */
+    DR_INSTR_CATEGORY_BRANCH = 0x10,       /* Branches */
+    DR_INSTR_CATEGORY_SIMD = 0x20,         /* Operations with vector registers (SIMD) */
+    DR_INSTR_CATEGORY_OTHER = 0x40         /* Other types of instructions */
```
derekbruening, line 12: Xref #6238 where we also had a list of categories, for the Google workload traces where we can't give out opcodes (but we'd want a programmatic way to find those categories from a real opcode, so this PR would fit the bill, if on all arches).

Comparing the category lists: looks like yours is also composable which is good.  We wanted to distinguish single and double precision floating point, and to separate out FMA (fused multiply add).  Maybe those are the main differences between the two lists; the extra categories in #6238 are less important.

kuhanov:
> > Calculation of category is based on C4.1.67 of ARM V8 Architecture reference manual * https://developer.arm.com/documentation/ddi0487/
> 
> Is that an older version? The latest PDF at that link has "C4.1.67 SVE2 Histogram Computation - Segment" so it must be a different section in your version: what is the title?

Need just use full 4.1 section
C4.1 A64 instruction set encoding

core/ir/category_api.h
kuhanov: > Xref #6238 where we also had a list of categories, for the Google workload traces where we can't give out opcodes (but we'd want a programmatic way to find those categories from a real opcode, so this PR would fit the bill, if on all arches).
> 
> Comparing the category lists: looks like yours is also composable which is good. We wanted to distinguish single and double precision floating point, and to separate out FMA (fused multiply add). Maybe those are the main differences between the two lists; the extra categories in #6238 are less important.

Could not find something like this for x86_64. We create just flat list for opcodes about their categories.
We could start from this list and add new ones categories. For example, some instructions could be load ans SIMD.
So, float could be in DR_INSTR_CATEGORY_MATH_FLOAT |  DR_INSTR_CATEGORY_MATH_SINGLE_FLOAT or  DR_INSTR_CATEGORY_MATH_FLOAT |  DR_INSTR_CATEGORY_MATH_DOUBLE_FLOAT

kuhanov:
Hi, @derekbruening . Could you tell me if I need to make anything else? Could I merge this patch?
Thx, Kirill

core/ir/category_api.h
derekbruening: There should be combinations today: a scatter instruction should be STORE|SIMD and gather LOAD|SIMD.  If that's not simple through the encoding bit scheme, maybe these would have to be flags added to every opcode in the codec*.txt files.  That is likely how it would be done for x86 and AArch32 as well: flags on every encoding entry (or else it has a be a big external switch statement on opcodes like today's instr_is_floating_ex()).

core/ir/instr_api.h
```diff
@@ -289,6 +289,8 @@ struct _instr_t {
 
     uint opcode;
 
+    uint category;
```
derekbruening, line 292: The instr_t size is tightly controlled and tuned for in the heap allocation buckets.  Adding a field requires making sure the BLOCK_SIZES in core/heap are sane, updating the commented sizes, and as a binary compatibility breakage this requires a major version bump and a listing in the release notes.

One alternative is not store anything and encode on demand when the category is asked about, like in instr_zeroes_ymmh() and others where we get the encoding info.  That has some downsides though of not having a target PC for PC-relative encoding differences and possible overhead: xref https://github.com/dvyukov/dynamorio/pull/1.

Another alternative is to fit the bits in an existing field like prefixes which is where the predicate bits are: but while they may well fit on AArch64 I think x86 uses a lot of those bits for internal decoding even if none are exposed.

core/ir/category_api.h
```diff
@@ -0,0 +1,15 @@
+#ifndef _DR_IR_CATEGORIES_H_
+#define _DR_IR_CATEGORIES_H_ 1
+
+typedef enum _dr_instr_category_type_t {
```
derekbruening, line 4: Missing Doxygen docs.

core/ir/category_api.h
```diff
@@ -0,0 +1,15 @@
+#ifndef _DR_IR_CATEGORIES_H_
```
derekbruening, line 1: Missing copyright and license header.

core/ir/category_api.h
```diff
@@ -0,0 +1,15 @@
...
+#define _DR_IR_CATEGORIES_H_ 1
+
+typedef enum _dr_instr_category_type_t {
+    DR_INSTR_CATEGORY_UNCATEGORIZED = 0x0, /* Uncategorised. */
```
derekbruening, line 5: Missing Doxygen docs: here and below.

core/ir/instr_api.h
```diff
@@ -738,6 +745,13 @@ DR_API
 void
 instr_set_opcode(instr_t *instr, int opcode);
 
+DR_API
+/** Assumes \p category is an set of DR_INSTR_CATEGORY_ constants
```
derekbruening, line 749: style: text doesn't go on the `/**` line for a multi-line comment

core/ir/instr_api.h
```diff
@@ -722,6 +724,11 @@ DR_API
 int
 instr_get_opcode(instr_t *instr);
 
+DR_API
+/** Returns \p instr's set of categories (set of DR_INSTR_CATEGORY_ constants). */
```
derekbruening, line 728: Always make references to other types actual links in Doxygen comments: `#dr_instr_category_type_t`

core/ir/instr_api.h
derekbruening, line 749: grammar: a set

core/lib/globals_api.h
```diff
@@ -377,9 +377,9 @@ typedef struct {
  */
 typedef struct {
 #    ifdef X64
-    uint black_box_uint[28];
+    uint black_box_uint[30];
```
derekbruening, line 380: 2 more?  Because of padding?  Please try to move the field to eliminate the extra uint of space.

core/lib/globals_api.h
```diff
@@ -377,9 +377,9 @@ typedef struct {
...
 #    else
-    uint black_box_uint[18];
+    uint black_box_uint[19];
 #    endif
 } instr_t;
 #else
```
derekbruening, line 385: I don't see any tests: all new features need tests.

derekbruening, line 385: All new API routines get a listing in the changelog in release.dox.

core/ir/category_api.h
derekbruening, line 12: What about instr_is_floating_ex() and dr_fp_type t (which seems to be completely missing for AArch64 today)?  I assume that would be implemented using the same category system: so we would want these types to also split into those categories?

core/ir/instr_api.h
derekbruening, line 728: The docs need to explain that this is only implemented for AArch64.

derekbruening, line 728: I only see categories being set on decode.  What if I modify a decoded instruction: I don't see the categories being invalidated.

derekbruening, line 728: Continuing the discussion of this PR only setting categories on decode: what if I synthesize a new instruction and then ask about its categories?  The existing IR behavior would be to encode and figure things out from the encoding: that's how instr_length(), instr_zeroes_ymmh(), etc. all operate.  I would expect the same here.  At the very least this behavior of only being valid for unmodified-since-decode instructions would have to be clearly documented: but that route does not seem very desirable.

core/CMakeLists.txt
```diff
@@ -1209,6 +1209,7 @@ DR_export_header(${CMAKE_CURRENT_SOURCE_DIR}/ir/instrlist_api.h dr_ir_instrlist.
 DR_export_header(${CMAKE_CURRENT_SOURCE_DIR}/ir/dr_ir_opcodes.h dr_ir_opcodes.h)
 DR_export_header(${CMAKE_CURRENT_SOURCE_DIR}/ir/x86/opcode_api.h dr_ir_opcodes_x86.h)
 DR_export_header(${CMAKE_CURRENT_SOURCE_DIR}/ir/arm/opcode_api.h dr_ir_opcodes_arm.h)
+DR_export_header(${CMAKE_CURRENT_SOURCE_DIR}/ir/category_api.h dr_ir_categories.h)
```
derekbruening, line 1212: Looks like this exported file will not be included by dr_api.h, unlike all the other files: that doesn't seem like what we want to suddenly require explicit includes for just this one file.

core/ir/aarch64/codec.c
```diff
@@ -9561,6 +9561,102 @@ encode_opnds_tbz(byte *pc, instr_t *instr, uint enc, decode_info_t *di)
...
 }
 
+static inline uint
+decode_load_store_category(uint enc)
+{
+    uint category = DR_INSTR_CATEGORY_OTHER;
+    /* Calculation of category is based on C4.1 'A64 instruction set encoding'
+     * of ARM V8 Architecture reference manual
+     *  https://developer.arm.com/documentation/ddi0487/
+     *  The encoding is:
```
derekbruening, line 9571: Changes involving AArch64 encoding should be reviewed by @AssadHashmi

derekbruening:
> Hi, @derekbruening . Could you tell me if I need to make anything else? Could I merge this patch? Thx, Kirill

Sorry for the delay -- I went ahead and did a full review: mostly requesting documentation and tests.  It would also be best if @AssadHashmi takes a look for any AArch64 decoder changes.

We are definitely interested in this type of feature for other projects: #6238, as mentioned.  That is why we're trying to iterate on some of the category details to make sure the same interface will cover all uses.  Thanks for bearing with the review and discussion process.

core/ir/category_api.h
kuhanov: That was done already over bits
        category =
            (BITS(enc, 22, 22) == 0) ? DR_INSTR_CATEGORY_STORE : DR_INSTR_CATEGORY_LOAD;
        if ((op0 & 0xc) == 0 && BITS(enc, 26, 26) == 1)
            category |= DR_INSTR_CATEGORY_SIMD;

core/ir/instr_api.h
kuhanov: Added check that `instr_operands_valid(instr))` is true. redecode if operands are invalid

core/CMakeLists.txt
kuhanov: Maybe we could remove this header at all? add categories to instr_api.h like dr_fp_type_t structure

core/ir/category_api.h
kuhanov: We will add on the next step. Need to discuss and fix that patch )))
Also we will share x86 approach. Uunfortunantly it is not programming just something like
```
[core/ir/x86/decode_table.c](https://github.com/DynamoRIO/dynamorio/commit/40cd2fa7cc53e54667b34430e8d5274bf713cecd#diff-f9b347cfa07cc46b1ea538b6e01652a0bdac4ad6c868183d734e4d823fa76853)
@@ -10703,4 +10703,1440 @@ const instr_info_t extra_operands[] =
    {OP_CONTD, 0xc00f0171, "<enclv cont'd>", ecx, edx, edx, xx, xx, mrm, x, END_LIST},
};

const uint category_by_opcode[] = {
    /* OP_INVALID */           DR_INSTR_CATEGORY_OTHER,
    /* OP_UNDECODED */         DR_INSTR_CATEGORY_OTHER,
    /* OP_CONTD */             DR_INSTR_CATEGORY_OTHER,
    /* OP_LABEL */             DR_INSTR_CATEGORY_OTHER,
    /* OP_add */               DR_INSTR_CATEGORY_MATH_INT,
   ...
```

derekbruening: OK.  Sure, let's start with this, and if we want to add single vs double, FMA, atomics, memory barriers, etc. later they can be appended.

core/CMakeLists.txt
derekbruening: Yes that sounds fine to me.

kuhanov:
Hi, @derekbruening 
added the following:
redecode if instr_operands_valid(instr) is not valid; 
implement initial instr_is_floating_ex for AArch64
added test (is this the correct place where we added it?)
moved categories from separate header

Should we fix anything else?
Thx, Kirill

core/ir/instr_api.h
```diff
@@ -287,7 +287,10 @@ struct _instr_t {
     /* translation target for this instr */
     app_pc translation;
 
-    uint opcode;
+    uint opcode : 16;
+
+    /* category of this instr (e.g. branch, load/store, etc.) */
+    uint category : 16;
```
derekbruening, line 293: This does seem like a good solution for your initial category set: I think shrinking the opcode will not cause any problems.  However, after some discussions here about categories we'd like to add on top of your initial ones, we're thinking a cap of 16 categories is too limiting.  We're about to bump the major version from 9 to 10 anyway so we're thinking let's go ahead and have a separate 32-bit category field and break compatibility now while the version bump is happening.

I'm seeing 3 32-bit fields at the top: if you add `uint category` after the `length` field it will not affect the 64-bit size (b/c it must currently have a padding slot to align the pointer field `bytes`) which will stay 112. 

This is still a compatibility break for 32-bit -- for that, update the black box as you did before, and the heap.c comment (I assume order didn't change there: asserts will fire if so), and should really bump the between-releases version too to satisfy our version policy https://dynamorio.org/page_new_release.html via: replace "9.93" with "9.94" in these files:
```
$ git grep '9\.93'
.github/workflows/ci-docs.yml:          export VERSION_NUMBER=9.93.$((`git log -n 1 --format=%ct` / (60*60*24)))
.github/workflows/ci-package.yml:          export VERSION_NUMBER=9.93.$((`git log -n 1 --format=%ct` / (60*60*24)))
.github/workflows/ci-package.yml:          export VERSION_NUMBER=9.93.$((`git log -n 1 --format=%ct` / (60*60*24)))
.github/workflows/ci-package.yml:          export VERSION_NUMBER=9.93.$((`git log -n 1 --format=%ct` / (60*60*24)))
.github/workflows/ci-package.yml:          export VERSION_NUMBER=9.93.$((`git log -n 1 --format=%ct` / (60*60*24)))
.github/workflows/ci-package.yml:          export VERSION_NUMBER=9.93.$((`git log -n 1 --format=%ct` / (60*60*24)))
.github/workflows/ci-package.yml:          export VERSION_NUMBER="9.93.$((`git log -n 1 --format=%ct` / (60*60*24)))"
CMakeLists.txt:set(VERSION_NUMBER_DEFAULT "9.93.${VERSION_NUMBER_PATCHLEVEL}")
```
Plus change `OLDEST_COMPATIBLE_VERSION_DEFAULT` to "994" in the top-level CMakeLists.txt.
(Or if PR #5835 goes in first and already bumps these to 9.94: bump further to 9.95.)

(Note that even with this change as-is to make `opcode` a 16-bit field we have compatibility breakage as the upper bits of opcode will now be set, even if no size change.)

core/lib/dr_api.h
```diff
@@ -101,6 +101,8 @@ extern "C" {
 /* Annotation handler registration routines */
 #include "dr_annotation.h"
 
+/* Instruction category constants */
```
derekbruening, line 104: Looks like this should be removed.

core/ir/instr_shared.c
```diff
@@ -449,6 +449,20 @@ instr_get_opcode(instr_t *instr)
 /* in rest of file, directly de-reference for performance (PR 622253) */
 #define instr_get_opcode inlined_instr_get_opcode
 
+#define inlined_instr_get_category(instr)                                         \
+    (IF_DEBUG_(CLIENT_ASSERT(sizeof(*instr) == sizeof(instr_t), "invalid type"))( \
+        ((instr)->category == DR_INSTR_CATEGORY_UNCATEGORIZED ||                  \
+         !instr_operands_valid(instr))                                            \
+            ? (instr_decode_with_current_dcontext(instr), (instr)->category)      \
```
derekbruening, line 456: But as mentioned in the prior review round, what about a synthetic instr with no raw bits?  Does an encoding have to be done (cached in the raw bits) and then a decoding?  At the very least, document that this API is not supported for synthetic instructions and is only supported for decoded instructions (and someone can add synthetic support later: document the lack of support in the header, and here put a `XXX i#6238: ...` comment about it).

suite/tests/api/drdecode_aarch64.c
```diff
@@ -130,6 +130,69 @@ test_noalloc(void)
      */
 }
 
+static void
+test_categories(void)
+{
+    const uint raw[] = {
+        0x00000000, /* no category */
+        0x12020000, /* int math */
```
derekbruening, line 138: Add the disassembly string in the comment for each so a human can understand this.

core/ir/instr_api.h
```diff
@@ -1854,6 +1875,20 @@ DR_API
 bool
 instr_is_rep_string_op(instr_t *instr);
 
+/**
+ * Indicates which category the instruction corresponds to.
+ */
+typedef enum {
+    DR_INSTR_CATEGORY_UNCATEGORIZED = 0x0, /**< Uncategorised. */
```
derekbruening, line 1882: nit: Weird to use the American "z" in the value but British "s" in the comment.

suite/tests/api/drdecode_aarch64.c
```diff
@@ -130,6 +130,69 @@ test_noalloc(void)
...
+        0x1D2104C0, /* store */
+        0xd65f03c0, /* branch */
+        0xc1e47c00, /* sme */
+        0xc5d57c04, /* sve2 */
+        0xc700c000, /* other */
+        0x02000000  /* other */
+    };
+
+    size_t instr_count = sizeof(raw) / sizeof(uint);
+    const uint categories[] = { DR_INSTR_CATEGORY_UNCATEGORIZED,
```
derekbruening, line 162: Would people expect _UNCATEGORIZED to mean an error: e.g., a synthetic instr for which we have no info?  And this should return _OTHER?

suite/tests/api/drdecode_aarch64.c
```diff
@@ -130,6 +130,69 @@ test_noalloc(void)
...
+
+    for (int i = 0; i < instr_count; i++) {
+        instr_t instr;
+        instr_init(GD, &instr);
+        instr_set_raw_bits(&instr, pc, 4);
+        uint cat = instr_get_category(&instr);
+        ASSERT(cat == categories[i]);
+        pc += 4;
+    }
+}
```
derekbruening, line 194: Add a test of a synthetic instr in an obvious category (e.g., `INSTR_CREATE_ldr`) and confirm it returns _UNCATEGORIZED.

derekbruening (commented):
If you could actually put back the full 32-bit field, but move it to a better spot, that will help us add more categories later (see comment on that).  Mostly just have minor comments elsewhere; should be good to go after this round.  Thank you for working through the review rounds.

suite/tests/api/drdecode_aarch64.c
kuhanov: We suppose that "UNCATEGORIZED" means we don't try to decode and set categories at all. _OTHERS means that we decode instruction and it doesn't assigned to known categories and move to othres

core/ir/aarch64/codec.c
```diff
@@ -9561,6 +9561,106 @@ encode_opnds_tbz(byte *pc, instr_t *instr, uint enc, decode_info_t *di)
...
+     *               op1
+     */
+
+    uint op1 = BITS(enc, 28, 25);
+    if ((BITS(enc, 31, 31) == 1 && op1 == 0) || op1 == 0x2) /* SME || SVE */
+        category = DR_INSTR_CATEGORY_SIMD;
+    else if (BITS(enc, 31, 31) == 0 && op1 == 0) /* op1 is 0 and 31 bit is 0 */
+        category = DR_INSTR_CATEGORY_UNCATEGORIZED;
+    else {
+        /*                       op1 - xxxx
```
AssadHashmi, line 9627: Might be worth adding a `DR_INSTR_CATEGORY_SVE`  with `op1` = `0010`. The codec supports all SVE instructions.

AssadHashmi: Hi @kuhanov, are you handling `op1` = `0001` and `op1` = `x111` as being in the same category of SIMD vector instructions for the purposes of this PR?

Bear in mind that the Advanced SIMD (NEON) instructions are for _fixed width_ 128 bit vectors. The op1 = 0010 encoding class covers _scalable vector widths_ from 128 to 2048 bits depending on the hardware implementation. SVE is a separate encoding class from SIMD even though they both handle vectors. There are 337 SVE opcodes, not counting the latest SVE2 instructions.

derekbruening: I would agree with @kuhanov on the cross-arch general _SIMD category grouping Neon and SVE for this PR.  I think the intended use cases for these categories are for more general purposes not tied to any one architecture.  Distinguishing individual ISA-specific characteristics does have its own uses cases but I think that would be more along the lines of FEATURE_ bits and would be a separate thing from these categories.

derekbruening:

kuhanov:
> t review when ready)

Hi, @derekbruening 
We are ready.  Could you approve? os should we fix anything else?
Thx, Kirill

core/ir/instr_api.h
kuhanov: Done. Upgrade to 9.95 and add to changes list

derekbruening (approved):
Thanks, I will merge it.

PR #6241, opened 2023-08-03
i#5490 ind br tgt: Stop skip and switch at target marker

derekbruening:
Updates reader_t::skip_instructions() and the scheduler's context switches and replay to stop at an indirect branch target marker instead of at the subsequent instruction record itself.

Adds an indirect branch to allasm_x86 and uses it in a new skip_unit_tests test.

Adds a new raw2trace test that an indirect branch marker is after a chunk boundary.

Adds scheduler tests that a context switch during live and replay modes goes to the marker.

Updates the invariant checker to avoid failure on the offline.skip* tests where the filetype record is not seen: now the checker uses the stream to get the filetype, except for the checker unit tests (added a XXX comment on possibly giving them a valid stream).

Issue: #5490


clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -2387,6 +2386,144 @@ test_replay_as_traced_from_file(const char *testdir)
...
+            check_ref(refs, idx, TID_A, TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_FUNC_ID) &&
+            check_ref(refs, idx, TID_A, TRACE_TYPE_MARKER,
+                      TRACE_MARKER_TYPE_FUNC_RETVAL) &&
+            // Shouldn't switch until after all the syscall's markers.
+            check_ref(refs, idx, TID_B, TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_VERSION) &&
+            check_ref(refs, idx, TID_B, TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_TIMESTAMP) &&
+            check_ref(refs, idx, TID_B, TRACE_TYPE_INSTR) &&
+            check_ref(refs, idx, TID_B, TRACE_TYPE_INSTR) &&
+            check_ref(refs, idx, TID_B, TRACE_TYPE_THREAD_EXIT) &&
+            check_ref(refs, idx, TID_A, TRACE_TYPE_MARKER,
```
abhinav92003, line 2457: Add comment here too.

clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -2387,6 +2386,144 @@ test_replay_as_traced_from_file(const char *testdir)
...
+                      TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL) &&
+            check_ref(refs, idx, TID_A, TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_FUNC_ID) &&
+            check_ref(refs, idx, TID_A, TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_FUNC_ARG) &&
+            check_ref(refs, idx, TID_A, TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_TIMESTAMP) &&
+            check_ref(refs, idx, TID_A, TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_CPU_ID) &&
+            check_ref(refs, idx, TID_A, TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_FUNC_ID) &&
+            check_ref(refs, idx, TID_A, TRACE_TYPE_MARKER,
+                      TRACE_MARKER_TYPE_FUNC_RETVAL) &&
+            // Shouldn't switch until after all the syscall's markers.
```
abhinav92003, line 2451: This one isn't relevant to this test.

clients/drcachesim/tests/skip_unit_tests.cpp
```diff
@@ -154,12 +164,57 @@ test_skip_initial()
...
+    std::string res;
+    if (!run_view_tool(skip_instrs, res))
+        return false;
+    std::stringstream res_stream(res);
+    // What we expect:
+    //   Output format:
+    //   <--record#-> <--instr#->: <---tid---> <record details>
+    //   ------------------------------------------------------------
+    //            202         131:     1260485 <marker: timestamp 13335506424391481>
+    //            203         131:     1260485 <marker: tid 1260485 on core 1>
```
abhinav92003, line 186: So without this PR, we would get 205 onwards? And now this PR makes sure we get 204 also, and 202-203 are due to existing logic of duplicating timestamp-tid? Maybe add a comment about why this is the expected output.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -812,6 +814,12 @@ invariant_checker_t::check_schedule_data(per_shard_t *global)
 {
     if (serial_schedule_file_ == nullptr && cpu_schedule_file_ == nullptr)
         return;
+    // We can't verify for a partial trace (this happens when running the checker
+    // with -skip_instrs, often in -test_mode combined with another tool).
```
abhinav92003, line 818: This won't WAI if we skipped instrs in the middle of execution.

abhinav92003 (approved):

clients/drcachesim/reader/reader.cpp
```diff
@@ -420,10 +420,12 @@ reader_t::skip_instructions_with_timestamp(uint64_t stop_instruction_count)
         VPRINT(this, 4, "SKIP: type=%s (%d), size=%d, addr=0x%zx\n",
                trace_type_names[next->type], next->type, next->size, next->addr);
         // We need to pass up memrefs for the final skipped instr, but we don't
-        // want to process_input_entry() on the first unskipped instr so we can
-        // insert the timestamp+cpu first.
+        // want to process_input_entry() on the first unskipped instr (or its target
```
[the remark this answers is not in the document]
derekbruening: Done.

clients/drcachesim/reader/reader.cpp
```diff
@@ -420,10 +420,12 @@ reader_t::skip_instructions_with_timestamp(uint64_t stop_instruction_count)
...
+        // marker) so we can insert the timestamp+cpu first.
         if (cur_instr_count_ + 1 == stop_count &&
-            type_is_instr(static_cast<trace_type_t>(next->type))) {
+            (type_is_instr(static_cast<trace_type_t>(next->type)) ||
+             (next->type == TRACE_TYPE_MARKER &&
+              next->size == TRACE_MARKER_TYPE_BRANCH_TARGET))) {
```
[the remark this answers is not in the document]
derekbruening: Yes had to update the scheduler too and not just the skip logic here.

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: Done.

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: Removed.

clients/drcachesim/tools/invariant_checker.cpp
derekbruening: You're referring to the fact that `skipped_instrs_` currently doesn't detect a skip in the middle: that looks to be true.  It has a `XXX` comment about it.  Added another such comment here.  The good news is we'll find out if someone does skip in the middle b/c the schedule invariants will fire so it's not like this lack will be some hidden problem.

derekbruening:
win32 failure is client.attach_test which was removed from the ignore list by PR #6227 reordering the expected output.  But it seems non-deterministic in the order which I just noted at https://github.com/DynamoRIO/dynamorio/issues/725#issuecomment-1666027989

PR #6243, opened 2023-08-04
i#5843 scheduler: Add API to disable an output stream

derekbruening:
Adds a new set_active() API on the output stream.  This can be used to mark a virtual cpu as inactive and reschedule its current input onto other outputs.

Adds a test.

Issue: #5843


derekbruening:
x64 api.rseq is #6185 failed-to-synch assert

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -1590,6 +1597,8 @@ scheduler_tmpl_t<RecordType, ReaderType>::pick_next_input(output_ordinal_t outpu
                     return res;
             } else if (options_.mapping == MAP_TO_ANY_OUTPUT) {
                 if (ready_queue_empty()) {
+                    if (prev_index == -1)
```
abhinav92003, line 1600: s/-1/some symbolic constant/

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -1943,6 +1954,29 @@ scheduler_tmpl_t<RecordType, ReaderType>::stop_speculation(output_ordinal_t outp
...
+        return sched_type_t::STATUS_INVALID;
+    outputs_[output].active = active;
+    VPRINT(this, 2, "Output stream %d is now %s\n", output,
+           active ? "active" : "inactive");
+    std::lock_guard<std::mutex> guard(sched_lock_);
+    if (!active) {
+        // Make the now-inactive output's input available for other cores.
+        // This will reset its quantum too.
+        // We haven't read ahead so we're "pre_instruction" vs regular switches.
+        inputs_[outputs_[output].cur_input].switching_pre_instruction = true;
```
abhinav92003, line 1972: If the output was already inactive, would cur_input be some out-of-bounds index? Maybe return early if `active` is same as the current `outputs_[output].active`.

abhinav92003 (approved):

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Changed to INVALID_INPUT_ORDINAL.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Yes, good catch.  Added the suggestion, and added a test.

PR #6244, opened 2023-08-04
i#725 attach: Remove flakiness from attach test

### LINKED ISSUE #725: attach feature on Windows

derekbruening:
Removes the printing of "event_post_attach" in the client.attach test, as it had non-deterministic ordering versus the "thread init" printing (and runall.cmake requires "thread init").  Replaces with a boolean check and error print at exit to ensure we saw the attach event.

This completes the base attach feature.  Future problems should open new issues.

Fixes #725

### DIFF
--- a/suite/tests/client-interface/attach_test.dll.c
+++ b/suite/tests/client-interface/attach_test.dll.c
@@ -1,5 +1,5 @@
 /* **********************************************************
- * Copyright (c) 2021-2022 Google, Inc.  All rights reserved.
+ * Copyright (c) 2021-2023 Google, Inc.  All rights reserved.
  * Copyright (c) 2008-2010 VMware, Inc.  All rights reserved.
  * **********************************************************/
 
@@ -39,14 +39,17 @@
 
 static thread_id_t injection_tid;
 static bool first_thread = true;
+static bool saw_attach_event = false;
 
 static void
 dr_exit(void)
 {
+    if (!saw_attach_event)
+        dr_fprintf(STDERR, "Error: never saw attach event!\n");
 #ifdef WINDOWS
     dr_fprintf(STDERR, "done\n");
 #else
-    /* The app prints 'done' for us. */
+        /* The app prints 'done' for us. */
 #endif
 }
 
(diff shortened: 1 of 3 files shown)


suite/tests/client-interface/attach_test.dll.c
```diff
@@ -39,14 +39,17 @@
...
 static void
 dr_exit(void)
 {
+    if (!saw_attach_event)
+        dr_fprintf(STDERR, "Error: never saw attach event!\n");
 #ifdef WINDOWS
     dr_fprintf(STDERR, "done\n");
 #else
-    /* The app prints 'done' for us. */
+        /* The app prints 'done' for us. */
```
ivankyluk, line 52: nit: the indentation seems odd. Was it changed by clang-format?

ivankyluk (approved):

suite/tests/client-interface/attach_test.dll.c
derekbruening: Yes, it was clang-format.  The fact that the clang-format-diff check in our test suite here didn't complain means that it agrees: so maybe an old version of clang-format did it the old way and a newer version on my machine and the CI does it this new way?  The new way looks worse...but it doesn't seem worth fighting with clang-format with markers to disable it just for this comment.

PR #6246, opened 2023-08-04
i#5843 scheduler: Add query of input workload

derekbruening:
Adds a new stream API get_input_workload_ordinal() which allows the output stream to figure out which workload its current input stream came from.

Adds a couple of sanity tests.

Issue: #5843


derekbruening:
x64 failure is api.rseq assert on synch again!!!  #6185

ivankyluk (approved):

PR #6247, opened 2023-08-04
i#5490 ind br tgt: Only set flag when recording

derekbruening:
Follow-up to PR #6241: only set the switching_pre_instruction flag when recording.  Otherwise an assert added by the PR fires.

Adds a no-record test which reproduces the assert and passes with the fix.

Issue: #5490

### DIFF
--- a/clients/drcachesim/scheduler/scheduler.cpp
+++ b/clients/drcachesim/scheduler/scheduler.cpp
@@ -1806,7 +1806,7 @@ scheduler_tmpl_t<RecordType, ReaderType>::next_record(output_ordinal_t output,
                     need_new_input = true;
                     in_wait_state = true;
                     input->processing_blocking_syscall = false;
-                    if (pre_instruction)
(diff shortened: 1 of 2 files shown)


derekbruening:
This was not detected in the original PR as release build passed.  This time we have a dedicated no-record upstream test added here, and I am testing non-release for import as well.

derekbruening:
Requesting review before all CI tests to try to avoid leaving things broken over weekend.

derekbruening:
There is actually a further problem with other outputs waiting on an input to reach the count but it never does b/c it's 1 less.  Grr.  Wondering whether moving the marker to after the instruction would have been the simpler way to go....

abhinav92003:
> Hoping to avoid leaving things broken over the weekend.

Apologies. Looked at the PR a bit late.

> There is actually a further problem

But looks like the review is not needed right now anyway.

derekbruening:
Replaced with #6250.

PR #6250, opened 2023-08-08
i#5490 ib targets: Add memref ind branch target field

derekbruening:
Adds a new "indirect_branch_target" field to memref_t.instr. Converts the new TRACE_MARKER_TYPE_BRANCH_TARGET marker to an internal record used by reader_t to fill in the new field.

Updates the invariant checker and test.
Updates the legacy test's marker counts.
Updates the documentation.

Issue: #5490

### DIFF
--- a/clients/drcachesim/reader/reader.cpp
+++ b/clients/drcachesim/reader/reader.cpp
@@ -174,7 +174,12 @@ reader_t::process_input_entry()
             cur_ref_.instr.tid = cur_tid_;
             cur_ref_.instr.type = (trace_type_t)input_entry_->type;
             cur_ref_.instr.size = input_entry_->size;
-            cur_ref_.instr.indirect_branch_target = last_branch_target_;
+            if (type_is_instr_branch(cur_ref_.instr.type) &&
+                !type_is_instr_direct_branch(cur_ref_.instr.type)) {
+                cur_ref_.instr.indirect_branch_target = last_branch_target_;
+            } else {
+                cur_ref_.instr.indirect_branch_target = 0;
+            }
             cur_pc_ = input_entry_->addr;
             cur_ref_.instr.addr = cur_pc_;
             next_pc_ = cur_pc_ + cur_ref_.instr.size;
--- a/clients/drcachesim/reader/reader.cpp
+++ b/clients/drcachesim/reader/reader.cpp
@@ -174,6 +174,7 @@ reader_t::process_input_entry()
             cur_ref_.instr.tid = cur_tid_;
             cur_ref_.instr.type = (trace_type_t)input_entry_->type;
             cur_ref_.instr.size = input_entry_->size;
+            cur_ref_.instr.indirect_branch_target = last_branch_target_;
             cur_pc_ = input_entry_->addr;
             cur_ref_.instr.addr = cur_pc_;
             next_pc_ = cur_pc_ + cur_ref_.instr.size;
@@ -286,6 +287,9 @@ reader_t::process_input_entry()
             skip_chunk_header_.erase(cur_tid_);
         } else if (cur_ref_.marker.marker_type == TRACE_MARKER_TYPE_RECORD_ORDINAL) {
             // Not exposed to tools.
+        } else if (cur_ref_.marker.marker_type == TRACE_MARKER_TYPE_BRANCH_TARGET) {
+            // Not exposed to tools.
+            last_branch_target_ = cur_ref_.marker.marker_value;
         } else {
             have_memref = true;
         }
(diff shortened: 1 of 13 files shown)


derekbruening:
Updated drmemtrace_samples marker counts and disassembly output in https://github.com/DynamoRIO/drmemtrace_samples/commit/b7b48eb96cb9bd1c09053994f5fae2e995b15e69

clients/drcachesim/reader/reader.cpp
```diff
@@ -174,6 +174,7 @@ reader_t::process_input_entry()
             cur_ref_.instr.tid = cur_tid_;
             cur_ref_.instr.type = (trace_type_t)input_entry_->type;
             cur_ref_.instr.size = input_entry_->size;
+            cur_ref_.instr.indirect_branch_target = last_branch_target_;
```
abhinav92003, line 177: We always set indirect_branch_target? I don't see last_branch_target_ being set to zero beyond init. Are we just letting it be a non-zero val and expecting users to not use it if it's not one of the indirect branch types? If yes, it may cause some confusion.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -1034,44 +1034,44 @@ check_branch_decoration()
...
-                         { "Indirect branches must be preceded by their targets", TID,
+                         { "Indirect branches must contain targets", TID,
                            /*ref_ordinal=*/3, /*last_timestamp=*/0,
                            /*instrs_since_last_timestamp=*/2 },
-                         "Failed to catch missing indirect branch target marker"))
+                         "Failed to catch missing indirect branch target field"))
             return false;
     }
-    // Indirect branch target: marker value incorrect.
+    // Indirect branch target: incorrect target value.
```
abhinav92003, line 1061: Should we call this an incorrect target value or a PC disc?

clients/drcachesim/tools/view.cpp
```diff
@@ -534,6 +534,15 @@ view_t::parallel_shard_memref(void *shard_data, const memref_t &memref)
             disasm.insert(newline, " (taken)");
         else if (memref.instr.type == TRACE_TYPE_INSTR_UNTAKEN_JUMP)
             disasm.insert(newline, " (untaken)");
+        if (trace_version_ >= TRACE_ENTRY_VERSION_BRANCH_INFO &&
+            (memref.instr.type == TRACE_TYPE_INSTR_INDIRECT_JUMP ||
+             memref.instr.type == TRACE_TYPE_INSTR_INDIRECT_CALL ||
```
abhinav92003, line 539: Would be useful to add a utility func (if one doesn't exist already) to identify these three types.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -750,19 +745,12 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
             if (!type_is_instr_direct_branch(memref.instr.type)) {
-                shard->last_indirect_target_ = shard->last_branch_marker_value_;
                 report_if_false(shard,
-                                shard->last_indirect_target_ != 0 &&
-                                    shard->prev_entry_.marker.type == TRACE_TYPE_MARKER &&
-                                    shard->prev_entry_.marker.marker_type ==
-                                        TRACE_MARKER_TYPE_BRANCH_TARGET,
-                                "Indirect branches must be preceded by their targets");
+                                // We assume the app doesn't actually target PC=0.
+                                memref.instr.indirect_branch_target != 0,
```
abhinav92003, line 750: Also assert that it's zero in all other cases.

abhinav92003 (approved):

clients/drcachesim/reader/reader.cpp
derekbruening: Good point.  Added a check for an indirect branch type; if it's not, this field is set to 0.

clients/drcachesim/tests/invariant_checker_test.cpp
derekbruening: It seems a different thing from a discontinuity b/c the next PC in the trace seems like the "real" one.

clients/drcachesim/tools/view.cpp
derekbruening: Oh right I was using `type_is_instr_branch() && !type_is_instr_direct_branch()` elsewhere.  I put that in here too.

clients/drcachesim/tools/invariant_checker.cpp
derekbruening: Done.

PR #6251, opened 2023-08-08
Fix gcc 12.2 warning in a64-on-x64 build

derekbruening:
Removes an unused variable which turns into an error in the aarch64-on-x86_64 build with gcc 12.2.


derekbruening:
Turns out this only shows up with a mis-configured build.

PR #6253, opened 2023-08-08
AARCHXX: Remove redundant macro definition

ksco:
`core/arch/arch.h:143` is duplicated with `core/arch/arch.h:188`, the definition is identical, which is allowed by C. But it will become different when line 188 adds support for RISC-V, this PR pre-removes the duplicate definition to avoid this issue.

### DIFF
--- a/core/arch/arch.h
+++ b/core/arch/arch.h
@@ -140,7 +140,6 @@ mixed_mode_enabled(void)
 #    define R12_OFFSET ((MC_OFFS) + (offsetof(priv_mcontext_t, r12)))
 #    define R13_OFFSET ((MC_OFFS) + (offsetof(priv_mcontext_t, r13)))
 #    define R14_OFFSET ((MC_OFFS) + (offsetof(priv_mcontext_t, r14)))
-#    define XFLAGS_OFFSET ((MC_OFFS) + (offsetof(priv_mcontext_t, xflags)))
 #    define PC_OFFSET ((MC_OFFS) + (offsetof(priv_mcontext_t, pc)))
 #    define SCRATCH_REG0 DR_REG_R0
 #    define SCRATCH_REG1 DR_REG_R1


core/arch/arch.h
```diff
@@ -140,7 +140,6 @@ mixed_mode_enabled(void)
 #    define R12_OFFSET ((MC_OFFS) + (offsetof(priv_mcontext_t, r12)))
 #    define R13_OFFSET ((MC_OFFS) + (offsetof(priv_mcontext_t, r13)))
 #    define R14_OFFSET ((MC_OFFS) + (offsetof(priv_mcontext_t, r14)))
-#    define XFLAGS_OFFSET ((MC_OFFS) + (offsetof(priv_mcontext_t, xflags)))
```
abhinav92003, line 143: The PR title says this is redundant: can you point to the other definition of this offset, maybe add it to the PR description too?

ksco: It's here: https://github.com/DynamoRIO/dynamorio/pull/6253/files#diff-5930c3a3d189a4f789955c17c5131c7015843070785649c71b254c92f7ded6b2R187

How should I point to that line? A GitHub link like the above or some plain text line like `core/arch/arch.h:187`?

ksco: I think the later is better, and PR description updated.

ksco:
CI keeps failing for some reason:

```
debug-internal-32: **** 1 build errors ****
	Error copying file "/.../dynamorio/build_debug-internal-32/lib32/debug/../drconfiglib.dll" to "/.../dynamorio/build_debug-internal-32/suite/tests/bin/drconfiglib.dll".
```

derekbruening:
A search should find #5888.  Unfortunately the 3x-retry only applies to test failures not build failures.

ksco:
Ah, ok, sorry for not searching, I'll keep retrying then.

ksco:
Ah... it's hard to have them all passed...

Issue #6238, opened 2023-08-02
Add instruction opcode categories to Google workload traces


derekbruening:
This is a proposal to provide categories of opcodes for the public Google workload traces https://dynamorio.org/google_workload_traces.html where the actual opcodes cannot be released.

Instead of linear list of absolute categories we were considering a composable bit vector of attributions:
+ Integer ops
+ Single Fp ops
+ Double Fp ops
+ FMA ops (to calculate roofline)
+ Vector 
+ Logical
+ Branch
+ Special ops (memory barrier, syscall?)
+ Micro-coded / multi-ops / complex
+ Conditional?
+ Atomic?

Examples:
+ SIMD Floating Point (VADDPS) → Vector, FP
+ Gather → Vector, Integer?, Load, Multi-ops

Xref instr_is_floating_ex() in today's interface which has:
```C
typedef enum {
    DR_FP_STATE,   /**< Loads, stores, or queries general floating point state. */
    DR_FP_MOVE,    /**< Moves floating point values from one location to another. */
    DR_FP_CONVERT, /**< Converts to or from floating point values. */
    DR_FP_MATH,    /**< Performs arithmetic or conditional operations. */
} dr_fp_type_t;
```

derekbruening:
We'll need to be able to programmatically convert real opcodes to categories for the traces, so we may as well add a category feature to the regular IR.  PR #6237 is an independently generated step in that direction.

We should also try to incorporate the `dr_fptype_t` types into the categories and re-implement that using the new category implementation, which will likely use bits added to the decoding tables for x86 and aarch32 and maybe to the decoding text sources for aarch64 and riscv64 if the latter can't use generalized bit groupings for all our categories (those seem to be enough for the initial set in PR #6237).

derekbruening:
We now have a good base for the categories.  Some additions/improvements may include:

+ Fill in the rest of the x86 decode table, the ones marked UNCATEGORIZED
+ Try to categorize for decode_fast, or document it's not supported

PR #6254, opened 2023-08-08
i#3544 RV64: Assert AUIPC instruction imm alignment

ksco:
Adds an assert to `encode_u_immpc_opnd` to ensure that the lower 12 bits of imm are all 0s.

Also, in order to share `.expect` files between static link and dynamic link tests, we added the ability to ignore specific lines in runcmp.cmake and removed the dependence on the external `diff` command. Note that we fall back to `STREQUAL` when the expected file is too large. We can use ZIP_LISTS to provide better performance, but this feature requires CMake 3.17, but bumping to 3.17 requires some effort on the Windows side.

Related: #6208 
Issue: #3544


core/ir/aarch64/encode.c
```diff
@@ -219,7 +219,8 @@ encoding_possible(decode_info_t *di, instr_t *in, const instr_info_t *ii)
 void
 decode_info_init_for_instr(decode_info_t *di, instr_t *instr)
 {
-    ASSERT_NOT_IMPLEMENTED(false); /* FIXME i#1569 */
+    memset(di, 0, sizeof(*di));
+    di->check_reachable = false;
```
abhinav92003, line 223: Do we need this? Wouldn't false be the default after the memset above?

core/ir/aarch64/encode.c
```diff
@@ -219,7 +219,8 @@ encoding_possible(decode_info_t *di, instr_t *in, const instr_info_t *ii)
 void
 decode_info_init_for_instr(decode_info_t *di, instr_t *instr)
```
abhinav92003, line 220: The current logic doesn't need to use the passed-in instr at all. Can't we do a simple brace init instead of this memset? Or maybe do `di.check_reachable = false` also on RISC-V (like we do for AArch64), which would add less overhead?

core/ir/riscv64/codec.c
```diff
@@ -1815,9 +1816,7 @@ encode_u_immpc_opnd(instr_t *instr, byte *pc, int idx, uint32_t *out)
         imm = (byte *)opnd_get_instr(opnd)->offset - (byte *)instr->offset;
     else
         return false;
-    /* FIXME i#3544: Add an assertion here to ensure that the lower 12 bits of imm are all
-     * 0. Assert only if decode_info_t.check_reachable is true. We should mark it as false
-     * to skip the check in get_encoding_info(), as we did for AARCHXX. */
+    ASSERT(!di->check_reachable || imm >> 12 << 12 == imm);
```
abhinav92003, line 1819: Maybe add some pairs of parenthesis for readability.

core/ir/riscv64/encode.c
```diff
@@ -91,8 +91,8 @@ encoding_possible(decode_info_t *di, instr_t *in, const instr_info_t *ii)
 void
 decode_info_init_for_instr(decode_info_t *di, instr_t *instr)
 {
-    /* FIXME i#3544: Not implemented */
-    ASSERT_NOT_IMPLEMENTED(false);
+    memset(di, 0, sizeof(*di));
```
abhinav92003, line 94: Same comment as above.

abhinav92003 (changes requested):

core/ir/aarch64/encode.c
abhinav92003: My concern is that initializing using memset is slower than initializing just the one field that's required, which may become apparent on frequent calls.

abhinav92003: For decoder related implementations (where code can potentially be invoked for each instr), I feel it may be better to just set check_reachable instead of memset.

ksco: At least for now, `check_reachable` is the only field in `decode_info_t` for both Aarch64/RISC-V. How about just delete the `memset(di, 0, sizeof(*di));` line?

ksco: Deleted the `memset`s in both Aarch64 and RISC-V.

abhinav92003: Thanks.

core/ir/riscv64/codec.c
```diff
@@ -1815,9 +1816,7 @@ encode_u_immpc_opnd(instr_t *instr, byte *pc, int idx, uint32_t *out)
         imm = (byte *)opnd_get_instr(opnd)->offset - (byte *)instr->offset;
     else
         return false;
-    /* FIXME i#3544: Add an assertion here to ensure that the lower 12 bits of imm are all
-     * 0. Assert only if decode_info_t.check_reachable is true. We should mark it as false
-     * to skip the check in get_encoding_info(), as we did for AARCHXX. */
+    ASSERT(!di->check_reachable || ((imm >> 12) << 12) == imm);
```
abhinav92003, line 1819: We should also add a test to verify that we fail gracefully for cases where check_reachable is true and the lower bits are not zero. Perhaps it makes more sense to return false for graceful failures (instead of the assert).

abhinav92003 (changes requested):

core/ir/riscv64/codec.c
ksco: Done. Thanks for the advice, I should've added more faulty tests.

suite/tests/api/ir_riscv64.c
```diff
@@ -1104,6 +1114,10 @@ test_jump_and_branch(void *dc)
     instr = INSTR_CREATE_auipc(dc, opnd_create_reg(DR_REG_A0),
                                opnd_create_pc(pc + (3 << 12)));
     test_instr_encoding_auipc(dc, OP_auipc, pc, instr);
+
+    instr = INSTR_CREATE_auipc(dc, opnd_create_reg(DR_REG_A0),
+                               opnd_create_pc(pc + (3 << 12)));
+    test_instr_encoding_failure(dc, OP_auipc, pc + 4, instr);
```
abhinav92003, line 1120: Add a comment here that we expect failure since the `target_pc - instr_encode_pc` has non-zero lower 12 bits.

abhinav92003 (approved):

suite/tests/api/ir_riscv64.c
ksco: Done.

ksco: Unfortunately, `instr_encode_to_copy` is asserting on reachability, thus causing CI to fail. We could use `instr_encode_check_reachability` but that's an unexported procedure.

abhinav92003: Perhaps we can export instr_encode_check_reachability. @derekbruening that does seem useful for tests and maybe for clients too?

derekbruening: There have been a couple of recent instances where we wanted to turn off reachability while encoding: #6196 and code that wants to get an x86/aarch32 encoding template (via get_encoding_info()) without any final PC in mind.  We might do the same thing for instr categories for #6238.  Those are all inside DR but I guess the point is that yes it is useful to control reachability checks.  Wait, it's asserting?  For client or any other use you would think we'd want to either fail or ignore, not assert: asserting is really only for internal DR encoding of blocks indicating it messed up the type of branch and we want a clearer error than "failed to encode".  Probably we should change all calls from clients to not assert and instead fail; and possibly allow ignoring reachability too.

abhinav92003: > Wait, it's asserting? For client or any other use you would think we'd want to either fail or ignore, not assert: asserting is really only for internal DR encoding of blocks indicating it messed up the type of branch and we want a clearer error than "failed to encode".

Yes; our implementation of `instr_encode_arch` asserts for reachability when its `assert_reachable` arg is set to true, e.g. in `instr_encode_to_copy`.

> Probably we should change all calls from clients to not assert and instead fail; and possibly allow ignoring reachability too.

Here are the three instr encode APIs that use `instr_encode_arch`: https://github.com/DynamoRIO/dynamorio/blob/fff1147153dd388faa1b69893a4b0910e35d484d/core/ir/encode_shared.c#L116

Maybe we should export all three of the above APIs and let the client decide which one they want? Even though asserting is not ideal, that's the behaviour clients have been relying on till now.

ksco: Sorry I know too little about DynamoRIO so I can't get involved in the discussion...

ksco:
@abhinav92003 @derekbruening  Can I ask some questions here?

Anyway, here is a code snippet from the second bb when running a program linked with libc:

(this is from a native-run, e.g. `gdb echo`)
```
│    0x3ff7ff199a <_dl_start+56>     auipc   s3,0xfffef         # s3 = pc + (0xfffef<<12)
│    0x3ff7ff199e <_dl_start+60>     add     s3,s3,1638
```

After this, `s3` has the value `0x3ff7fe1000`, and `p/x *0x3ff7fe1000` gives `$1 = 0x464c457f`. So the value in `s3` is a valid memory address.

When I ran `echo` with `gdb bin64/drrun` then `r echo`:

```
|    0x3fb3cf40cc    lui     s3,0x40f8
│    0x3fb3cf40d0    addw    s3,s3,-827
│    0x3fb3cf40d4    sll     s3,s3,0xc
│    0x3fb3cf40d8    add     s3,s3,-1638
│    0x3fb3cf40dc    add     s3,s3,1638
```

`auipc` is mangled into 4 instructions to load an absolute value. So far so good, but after this, `s3` has the value `0x40f7cc5000`, and `p/x *0x40f7cc5000` gives `Cannot access memory at address 0x40f7cc5000`.

In gdb, I checked `insert_mov_immed_arch()`, and it's working as expected. I also checked `decode_u_immpc_opnd()` where `auipc` instr got its `PC_kind` operand, seems all good.

So my conclusion is in the native-run, `pc + (0xfffef<<12) + 1638` is a valid memory address, while in dr, it's not. Any ideas what's wrong?

ksco:
Ah! I know the reason, `auipc`'s imm should be a sign-extended 32 bits value, but in `decode_u_immpc_opnd()` I declared it as a uint. It's poorly tested. Just writing the question down already helped a lot...

derekbruening:
> So my conclusion is in the native-run, `pc + (0xfffef<<12) + 1638` is a valid memory address, while in dr, it's not. Any ideas what's wrong?

Any instruction that reads the current actual PC has to be mangled to use the original PC and not the code cache PC so it sounds like the mangling is not using the proper target value.

ksco:
> sounds like the mangling is not using the proper target value.

Thanks and yes, it's using the original PC as the target. I just find the cause, see https://github.com/DynamoRIO/dynamorio/pull/6254#issuecomment-1680931350. Sorry for the bother. Now move on to the next problem. I feel like it's not too far to run a simple program with libc. :)

suite/tests/api/ir_riscv64.c
abhinav92003: @ksco The suggestion is to set assert_reachable to false in the implementation of instr_encode_to_copy, and note this change in release.dox as a minor compatibility change. With this change, failure in encoding would not cause a crash and we can still write tests that detect reachability issues.

ksco: Done.

suite/tests/api/ir_riscv64.expect
```diff
@@ -221,8 +221,6 @@ c.xor  a5 -> fp
 c.sub  a5 -> fp
 test_integer_arith complete
 lui    0x2a -> a0
-<Internal Error: Failed to encode instruction: 'auipc  0x0000004000016234 -> a0'
```
abhinav92003: We should continue printing this error (and not crashing). Perhaps when `check_reachable` is set to true.

abhinav92003 (approved):

suite/tests/api/ir_riscv64.expect
ksco: This error message is printed by `SYSLOG_INTERNAL_ERROR`, so I think we could use `SYSLOG_INTERNAL_WARNING` for check_reachable printing (without crashing). But with this change:

```diff
    if (enc == ENCFAIL) {
+       if (check_reachable) {
+           char disas_instr[MAX_INSTR_DIS_SZ];
+           instr_disassemble_to_buffer(dcontext, instr, disas_instr, MAX_INSTR_DIS_SZ);
+           SYSLOG_INTERNAL_WARNING("Internal Error: Failed to encode instruction: '%s'\n", disas_instr);
+        }
```

It seems `SYSLOG_INTERNAL_WARNING` evaluated to empty in suite/tests/bin/api.ir-static, causing a difference with suite/tests/bin/api.ir. Should we just use `print_file(STDERR, ...)`?

abhinav92003: But if it's the debug build `SYSLOG_INTERNAL_ERROR` should be set appropriately. It was before when we were crashing right?

ksco: > It was before when we were crashing right?

I also thought so, but it's not: https://github.com/DynamoRIO/dynamorio/actions/runs/5848221837/job/15855283218#step:6:1932. The reason CI was failing is that -- in api.ir-static, `SYSLOG_INTERNAL_ERROR` evaluated to empty while in api.ir it's not. `SYSLOG_INTERNAL_ERROR` is (surprisingly) not causing a crash, so assert_reachable is not really asserting.

And the reason why `SYSLOG_INTERNAL_ERROR` is evaluated to empty in api.ir-static is that -- api.ir-static is built as a STANDALONE_DECODER.

abhinav92003: > I also thought so, but it's not

So then maybe we don't need to set `assert_reachable` to `false` in this PR? Let's revert that part.

>  so assert_reachable is not really asserting.

I see #5298 added the `assert_reachable` arg. The description says that it silences a warning, not an assert statement. I guess the name `assert_reachable` could be changed to `warn_reachable` or something, but let's not change that in this PR.

>  in api.ir-static, SYSLOG_INTERNAL_ERROR evaluated to empty while in api.ir it's not.

Perhaps the `ASSERT(pc == NULL)` in `test_instr_encoding_failure` is all we need then. For api.ir where the warning will be printed we can have a regex like `(<message>)?` which would work also for api.ir-static.

ksco: > we can have a regex like `(<message>)?`

For api.ir, we are using `runcmp.cmake` to compare which uses `diff`:

```
if (AARCH64 OR RISCV64)
  # The ir_aarch64.expect file is too large for CMake's regexes.
  set(api.ir-static_runcmp "${CMAKE_CURRENT_SOURCE_DIR}/runcmp.cmake")
  set(api.ir-static_runcmp_capture "stderr")
endif ()
```

Maybe set assert_reachable to false is the way to go?

abhinav92003: > Maybe set assert_reachable to false is the way to go?

This doesn't sound like a good idea now that we understand it's quite innocuous and suits non-test code well. The real problem here seems to be that api.ir and api.ir-static use the same `expect` file. I'm not suggesting we split that because they share most of their contents. I also see
```
    # The ir_aarch64.expect file is too large for CMake's regexes.
```
so we cannot just switch to using regexps.

Circling back to your earlier suggestion of using `print_file(STDERR, ...)`:

@derekbruening why do we not define syslog macros for the stand-alone case? Can we just define those macros? Or maybe replace this particular log with an explicit call to `d_r_notify`?

derekbruening: I would guess it's because for STANDALONE_DECODER there is no mechanism to turn off notifications because there is no place to set stderr_mask or msgbox_mask.  Remember that on Windows an error by default has a pop-up window which is really annoying if you can't turn it off (at the same time it kind of has to default to that for GUI apps run without a console where a stderr print will never be visible).  Maybe if you set msgbox_mask to 0 for STANDALONE_DECODER and have some way to set options -- which ties into us wanting a real initialization for all decoder use which could solve it by reading the DYNAMORIO_OPTIONS env var: xref #2499, #6002.

abhinav92003: @ksco From offline discussion with Derek: since runcmp.cmake is our own script we can perhaps add macros or something to it so that the additional logs are expected only in the non-static test. I haven't looked at the implementation myself so you might want to check if that's easily done in this PR or not.

derekbruening: Possibilities:
+ Add per-line (to avoid too-large issue) regex support in runcmp
+ Split the expect file and share the common with `#include` which either runcmp is augmented to look at or you run cpp
+ Add conditional marker capabilities: maybe leveraging how suite/tests/CMakeLists.txt does it for .template* files, or something new in runcmp

ksco: Sure, thank you! I’ll take a look.

ksco:
@derekbruening @abhinav92003 
I came up with something in the above commits to addressing the differences between `api.ir` and `api.ir-static`. Instead of using `STREQUAL` to perform a strict match, we now use `diff --ignore-matching-lines 'regex' a b` to do per-line matching.

`ignore-matching-lines` is optional -- when not set, it's essentially a strict match just like before; when set, lines match the regex will be filtered out.

I don't know if this is an acceptable way, please let me know what you think.

suite/tests/CMakeLists.txt
```diff
@@ -1923,10 +1924,16 @@ endif ()
...
+if (AARCH64)
   # The ir_aarch64.expect file is too large for CMake's regexes.
   set(api.ir-static_runcmp "${CMAKE_CURRENT_SOURCE_DIR}/runcmp.cmake")
   set(api.ir-static_runcmp_capture "stderr")
+elseif (RISCV64)
+  # The ir_riscv64.expect file is too large for CMake's regexes.
+  set(api.ir-static_runcmp "${CMAKE_CURRENT_SOURCE_DIR}/runcmp.cmake")
+  set(api.ir-static_runcmp_capture "stderr")
+  # Internal error lines are ignored as they are not printed in static builds.
+  set(api.ir-static_runcmp_ignore_matching_lines "^[><]")
```
derekbruening, line 1936: Note that there should be no need for `>`: that indicates an erroneously included newline which should be removed from the string (a newline is auto-added later).

suite/tests/api/ir_riscv64.expect
```diff
@@ -221,6 +221,8 @@ c.xor  a5 -> fp
 c.sub  a5 -> fp
 test_integer_arith complete
 lui    0x2a -> a0
+<Internal Error: Failed to encode instruction: 'auipc  0x0000004000016234 -> a0'
+>
```
derekbruening, line 225: See above: let's remove the erroneous newline before the `>`.

suite/tests/api/ir_riscv64.expect
```diff
@@ -221,6 +221,8 @@ c.xor  a5 -> fp
 c.sub  a5 -> fp
 test_integer_arith complete
 lui    0x2a -> a0
+<Internal Error: Failed to encode instruction: 'auipc  0x0000004000016234 -> a0'
```
derekbruening, line 224: If this is ignored, why do we need to list it here?

suite/tests/runcmp.cmake
```diff
@@ -51,38 +52,37 @@ if (cmd_result)
...
-if (NOT "${output}" STREQUAL "${str}")
-  # make it easier to debug
-  set(tmp "${cmp}-out")
-  file(WRITE "${tmp}" "${output}")
-  set(tmp2 "${cmp}-expect")
-  file(WRITE "${tmp2}" "${str}")
-
-  set(diffcmd "diff")
-  execute_process(COMMAND ${diffcmd} ${tmp} ${tmp2}
+set(diffcmd "diff")
```
derekbruening, line 72: This is problematic for Windows where there may not be a "diff" command; plus for UNIX it may not be on the path.  Previously this was just a hacky piece of code only ever run on failure.  But to run "diff" every time it should be done properly: use `find_program` to find it (to try to locate if not on the path); it still could cause problems for Windows developers.  Even if the GA CI VM's have a "diff" installed, is that standard on Windows?  I doubt it.  It might be best to either do a line-by-line compare and implement the ignore-line manually in CMake code (this fits better with the general philosophy: see below), or use the original code for Windows.

Note that generally the philosophy has been to rely on as few external applications as possible, primarily motived by Windows where often nothing but CMake is installed, but also Mac where utilities don't have GNU extensions.  Generally we have preferred doing everything in CMake, and not relying on little utility applications being installed.

ksco: Thanks for the detailed review, I get the idea. Let's reimplement it fully in CMake then. I'm not familiar with CMake, so it will take some time.

suite/tests/api/ir_riscv64.expect
ksco: It's ignored only in static builds (e.g. `api.ir-static`) not in dynamic builds (e.g. `api.ir`).

suite/tests/runcmp.cmake
derekbruening: There are examples of walking line by line in our other code: e.g.:
```
$ git grep 'foreach.*line' 
core/CMake_symbol_check.cmake:foreach(line ${lines})
```

ksco: > Let's reimplement it fully in CMake then.

Done.

suite/tests/runcmp.cmake
```diff
@@ -57,32 +57,60 @@ if ("${capture}" STREQUAL "stderr")
...
 
-# We do not support regex in expect file b/c ctest can't handle big regex:
-#   "RegularExpression::compile(): Expression too big."
+# Convert file contents into a CMake list, where each element in the list is one
+# line of the file, empty lines are ignored.
+string(REGEX REPLACE "^[\n]+" "" output "${output}")
+string(REGEX REPLACE "[\n]+$" "" output "${output}")
+string(REPLACE "\n\n" "\n" output "${output}")
+STRING(REGEX REPLACE ";" "\\\\;" output "${output}")
+STRING(REGEX REPLACE "\n" ";" output "${output}")
```
derekbruening, line 70: This looks much more complex than the approach I pointed you at in CMake_symbol_check.cmake:
```
string(REGEX MATCHALL "([^\n]+)\n" lines "${oneword}")
```
OK, most of the work here is about blank lines.  Do we really need to spend all this effort on leading and trailing and double newlines?  It seems ok to require the expect file to match the output file's newlines exactly: weren't we already doing that?

suite/tests/runcmp.cmake
```diff
@@ -57,32 +57,60 @@ if ("${capture}" STREQUAL "stderr")
...
 
-# Use diff -I to get the ability to skip some lines.
-
-set(diffcmd "diff")
-if (ignore_matching_lines)
-  execute_process(COMMAND ${diffcmd} -I ${ignore_matching_lines} ${tmp} ${cmp}
-    RESULT_VARIABLE dcmd_result
-    ERROR_VARIABLE dcmd_err
-    OUTPUT_VARIABLE dcmd_out)
+if(NOT "${ignore_matching_lines}" STREQUAL "")
```
derekbruening, line 78: style: `if (`

suite/tests/runcmp.cmake
```diff
@@ -57,32 +57,60 @@ if ("${capture}" STREQUAL "stderr")
...
+  # Filter out all the ignored lines.
+  set(filtered_expect)
+  foreach (item ${expect})
+      string(REGEX MATCH "${ignore_matching_lines}" match_result "${item}")
+      if (NOT match_result)
```
derekbruening, line 83: These 2 lines can be collapsed into `if (NOT "${item}" MATCHES "${ignore_matching_lines}")`

suite/tests/runcmp.cmake
```diff
@@ -57,32 +57,60 @@ if ("${capture}" STREQUAL "stderr")
...
+  endforeach ()
 else ()
-  execute_process(COMMAND ${diffcmd} ${tmp} ${cmp}
-    RESULT_VARIABLE dcmd_result
-    ERROR_VARIABLE dcmd_err
-    OUTPUT_VARIABLE dcmd_out)
+  set(filtered_expect ${expect})
+endif ()
+
+list (LENGTH filtered_expect filtered_expect_length)
```
derekbruening, line 91: style: `list(`.  Ditto below.  (Only space for control flow keywords.)

derekbruening (approved):

ksco:
@derekbruening I pushed something above, are there any more revisions?

suite/tests/runcmp.cmake
```diff
@@ -62,54 +62,52 @@ endif ()
...
-    if (NOT ${item1} STREQUAL ${item2})
-        set(lists_identical FALSE)
-        message(STATUS "The first difference:")
-        message(STATUS "< ${item1}")
-        message(STATUS "> ${item2}")
-        break ()
+  if (filtered_expect_length GREATER "1000")
+    # CMake list(GET ...) operation is slow, do STREQUAL for large files, give up
+    # the diff output.
+    if(NOT filtered_expect STREQUAL output)
```
derekbruening, line 93: style: `if (`

suite/tests/runcmp.cmake
```diff
@@ -62,54 +62,52 @@ endif ()
...
+      set(lists_identical FALSE)
     endif ()
-  endforeach ()
-endif ()
+  else ()
+    math(EXPR len "${filtered_expect_length} - 1")
+    foreach (index RANGE ${len})
```
derekbruening, line 98: I think it'd be ok to bump the min cmake version to 3.17 (ubuntu20 has 3.27 at https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md) and use ZIP_LISTS -- does that eliminate the slowness (your comment above)?

derekbruening (approved):
Sorry, I missed the re-review request, prob b/c I did say "Approve" on the last one, but good to look again after non-trivial changes.

suite/tests/runcmp.cmake
ksco: I bumped CMake version to 3.17, but there are some CMake errors in Windows, which I don't understand, could you take a look?

derekbruening: Never seen that -- did you try searching?  Searching shows solutions like https://github.com/RichieSams/FiberTaskingLib/issues/109.  But if it's too messy I would just bail and go back to the prior cmake with a XXX comment about switching to ZIP_LISTS when we've put in the effort elsewhere to get to 3.17.

ksco: Yes I searched but didn't find out why.

> I would just bail and go back to the prior cmake with a XXX comment about switching to ZIP_LISTS when we've put in the effort elsewhere to get to 3.17.

Okay.

derekbruening:
This PR seems to have removed the printing out of details on the diff when runcmp.cmake detects a failure: #6411

ksco:
> This PR seems to have removed the printing out of details on the diff when runcmp.cmake detects a failure: #6411

Looking at runcmp.cmake, seems we will only do `STREQUAL` for files with more than 1000 lines due to performance reasons, so there is no diff output at all for large files. We can `ZIP_LISTS` but this would require CMake 3.17. Even for files under 1000 lines, only the first line of the diff is printed, I guess b/c a full diff output is not easy to implement.

Quoting you from the above [comment](https://github.com/DynamoRIO/dynamorio/pull/6254#discussion_r1308988186): Note that generally the philosophy has been to rely on as few external applications as possible -- maybe it's time to bump CMake and then use `ZIP_LISTS`? or maybe we could detect if the current env is in GA and then use `diff` to print the error results like before?

derekbruening:
You wrote:
```
CMake list(GET ...) operation is slow
```

Just how slow is it to do the diff?  (Maybe you said so somewhere in the review comments above from a few months ago...if so please remind me.).

We are deciding between:
1. Put in the work to get CMake 3.17 to work on Windows
2. Live with a slow diff
3. Use an external diff program

I would vote for #3: how about we do a find_program and if we find an external diff, then for >1000-line output we run the external diff?

ksco:
> Just how slow is it to do the diff?

It's so slow that some test cases will time out.

> I would vote for #3: how about we do a find_program and if we find an external diff, then for >1000-line output we run the external diff?

Yes, this sounds practical to do, I'll submit a PR soon.

Issue #4103, opened 2020-02-14
Run clang-tidy to help improve code quality


derekbruening:
Just an idea to run clang-tidy on Travis like we do with clang-format.

derekbruening:
clang-tidy complains about 05ac1a102d4ba8881803794fe190e5a861426841 from PR #4121 for #4120:
```
function 'drreg_restore_all' has a definition with different parameter names
```
@johnfxgalea maybe sthg to clean up in a separate commit at some point.

johnfxgalea:
Yea np.

derekbruening:
I bet there are a bunch more of these: I only end up seeing ones in code we've changed when I import it into a clone.  I have not run clang-tidy on the whole code base.

derekbruening:
More errors:

clients/drcachesim/tools/invariant_checker.cpp: no header providing "instr_is_syscall" is directly included

clients/drcachesim/tracer/instru_online.cpp: no header providing "uint64" is directly included

clients/drcachesim/tracer/output.cpp: no header providing "TIDFMT" is directly included

PR #6257, opened 2023-08-08
i#4103 clang-tidy: Clean up headers to include what is used

derekbruening:
Updates many drcachesim files to include what is used.  This is to address clang-tidy warnings seen when we import the files into our environment.

Issue: #4103

### DIFF
--- a/clients/drcachesim/analyzer.cpp
+++ b/clients/drcachesim/analyzer.cpp
@@ -30,15 +30,34 @@
  * DAMAGE.
  */
 
-#include <inttypes.h>
-#include <iostream>
-#include <thread>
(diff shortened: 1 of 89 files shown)


derekbruening:
ub22 failure is api.rseq #6185

derekbruening:
I ran an "include what you use" tool on the imported version of these so it only changed the headers we use.  I then had to spend quite a bit of time cleaning it up (duplicate headers; a couple of ordering dependences which I commented).  It's not perfect as there are still some ../ paths in other files but this is a good improvement.

abhinav92003 (approved):

Issue #6258, opened 2023-08-09
Invariant checker fails with "Function marker retaddr should match prior call".


ivankyluk:
Invariant checker fails with "Function marker retaddr should match prior call" when there are tailcalls.

**To Reproduce**
Run invariant checker against program with tailcalls. For example:

static int cnt = 0;
static int dummy_cnt = 0;

int dummy() {
  dummy_cnt++;
}

int do_while() {
  do {
    cnt++;
    dummy();
  } while (cnt < 5);
  return cnt;
}

int
main(void)
{
    return do_while();
}

and compile with "gcc -fomit-frame-pointer".

**Expected behavior**
Programs with tailcall should not fail the invariant check.

**Versions**
 - What version of DynamoRIO are you using? latest
 - Does the latest build from https://github.com/DynamoRIO/dynamorio/releases solve the problem? yes
- What operating system version are you running on? ("Windows 10" is *not* sufficient: give the release number.) Linux
- Is your application 32-bit or 64-bit? 64-bit

PR #6261, opened 2023-08-10
i#6258: handle nested  and tail calls for function marker return address check.

ivankyluk:
i#6258: handle nested and tail calls for function marker return address check.

Uses a stack to handle return addresses for nested function calls. Also stores the return address when a signal handler (kernel transfer) is called.

Fixes: #6258


clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -587,6 +587,8 @@ check_function_markers()
     constexpr memref_tid_t TID = 1;
```
derekbruening, line 587: I would mention tailcalls in the PR description as well.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -672,6 +674,72 @@ check_function_markers()
...
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_ID, 2),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_RETADDR, CALL_PC + CALL_SZ),
+
+            gen_instr_type(TRACE_TYPE_INSTR_DIRECT_CALL, TID, SECOND_CALL_PC,
+                           SECOND_CALL_SZ),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_ID, 3),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_RETADDR,
+                       SECOND_CALL_PC + SECOND_CALL_SZ),
+            gen_instr_type(TRACE_TYPE_INSTR_RETURN, TID,
+                           /*pc=*/SECOND_CALL_PC + SECOND_CALL_SZ + 1, /*size=*/1),
```
derekbruening, line 703: nit: Seems odd to have the pc= here and lines 706 and 735 when it's not on any other arg using a named _PC value.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -672,6 +674,72 @@ check_function_markers()
...
+        std::vector<memref_t> memrefs = {
+            gen_instr(TID, 1),
+            gen_instr_type(TRACE_TYPE_INSTR_DIRECT_JUMP, TID, CALL_PC, CALL_SZ),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_ID, 2),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_RETADDR, /*pc=*/123456),
+        };
+        if (!run_checker(memrefs, false))
+            return false;
+    }
+    // Correctly handle nested function calls.
```
derekbruening, line 689: " including tailcalls."

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -672,6 +674,72 @@ check_function_markers()
...
+            gen_instr_type(TRACE_TYPE_INSTR_DIRECT_CALL, TID, SECOND_CALL_PC,
+                           SECOND_CALL_SZ),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_ID, 3),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_RETADDR,
+                       SECOND_CALL_PC + SECOND_CALL_SZ),
+
+            gen_instr_type(TRACE_TYPE_INSTR_RETURN, TID,
+                           SECOND_CALL_PC + SECOND_CALL_SZ + 1, 1),
+
+            gen_marker(TID, TRACE_MARKER_TYPE_SYSCALL, 15),
```
derekbruening, line 733: There's no KERNEL_XFER marker though so this looks either malformed if x64 Linux or with the wrong syscall number or it's some other arch where 15 is not a sigreturn.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -672,6 +674,72 @@ check_function_markers()
...
+
+            gen_instr_type(TRACE_TYPE_INSTR_DIRECT_JUMP, TID, /*pc=*/SECOND_CALL_PC + 2,
```
derekbruening, line 735: If a KERNEL_XFER is added above, then this is a PC discontinuity as it doesn't return to the interruption point.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -672,6 +674,72 @@ check_function_markers()
...
+
+            gen_instr_type(TRACE_TYPE_INSTR_DIRECT_CALL, TID, SECOND_CALL_PC,
+                           SECOND_CALL_SZ),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_ID, 3),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_RETADDR,
+                       SECOND_CALL_PC + SECOND_CALL_SZ),
+            gen_instr_type(TRACE_TYPE_INSTR_RETURN, TID,
+                           /*pc=*/SECOND_CALL_PC + SECOND_CALL_SZ + 1, /*size=*/1),
+
+            gen_instr_type(TRACE_TYPE_INSTR_DIRECT_JUMP, TID,
```
derekbruening, line 705: Seems worth adding a comment about this being a tailcall.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -653,6 +663,10 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                       << "marker type " << memref.marker.marker_type << " value 0x"
                       << std::hex << memref.marker.marker_value << std::dec << "\n";
         }
+        // Push the return address when kernel transfer happens.
+        if (memref.marker.marker_type == TRACE_MARKER_TYPE_KERNEL_EVENT) {
+            shard->retaddr_stack_.push(memref.marker.marker_value);
```
derekbruening, line 668: The interruption point is not a target of a return, though.  The comment needs to be updated to explain that this is to match the return used post-signal-handler to run the restorer code.  It should also explain that we thus assume all signal handlers return normally and don't longjmp (we would need to store the stack pointer in the stack and detect skipped frames to handle that).

As for the value pushed, probably better to push some sentinel or 0 or something, as the current value you are pushing here is just going to be confusing.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: Done.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: Done.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: I rewrote the test case, please take another look.

derekbruening: You're requesting the next round of review?  Normally you'd reply to all the comments and resolve all (potentially leaving some unresolved if they need more discussion) and then click the circular arrow in the upper right to re-request a review.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: I've removed the /*pc=*/ comments.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: Ack.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: Done.

clients/drcachesim/tools/invariant_checker.cpp
ivankyluk: I've updated the comment and the code accordingly.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: I rewrote the test to make it clear, ptal.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -672,6 +672,146 @@ check_function_markers()
...
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_ID, 2),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_RETADDR, /*pc=*/123456),
+        };
+        if (!run_checker(memrefs, false))
+            return false;
+    }
+    // Correctly handle nested function calls including tailcalls.
+    {
+        instr_t *label1 = INSTR_CREATE_label(GLOBAL_DCONTEXT);
+        instr_t *call1 = XINST_CREATE_call(GLOBAL_DCONTEXT, opnd_create_instr(label1));
```
derekbruening, line 690: No reason to have the labels here right?  These can directly target the real instrs.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -672,6 +672,146 @@ check_function_markers()
...
+        instrlist_append(ilist, label1);
+        instrlist_append(ilist, call2);
+        instrlist_append(ilist, label2);
+        instrlist_append(ilist, ret2);
+        instrlist_append(ilist, jmp);
+
+        static constexpr addr_t BASE_ADDR = 0x123450;
+        const int call_instr_length = instr_length(GLOBAL_DCONTEXT, call1);
+        std::vector<memref_with_IR_t> memref_setup = {
+            { gen_marker(TID, TRACE_MARKER_TYPE_FILETYPE, OFFLINE_FILE_TYPE_ENCODINGS),
```
derekbruening, line 709: Do we need the encodings?  This invariant check never looks at a decoded instruction, right?  We could simplify like the above cases and have no encodings?

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -672,6 +672,146 @@ check_function_markers()
...
+              nullptr },
+            { gen_instr_type(TRACE_TYPE_INSTR_DIRECT_CALL, TID), call1 },
+            { gen_marker(TID, TRACE_MARKER_TYPE_FUNC_ID, 1), nullptr },
+            { gen_marker(TID, TRACE_MARKER_TYPE_FUNC_RETADDR,
+                         BASE_ADDR + call_instr_length),
+              nullptr },
+            { gen_instr_type(TRACE_TYPE_INSTR_DIRECT_CALL, TID), call2 },
+            { gen_marker(TID, TRACE_MARKER_TYPE_FUNC_ID, 2), nullptr },
+            { gen_marker(TID, TRACE_MARKER_TYPE_FUNC_RETADDR,
+                         BASE_ADDR + call_instr_length * 2),
```
derekbruening, line 719: This assumes call1 and call2 are the same size; assert they're the same up above?  Or if no encodings you can specify the sizes and make them the same by definition.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -672,6 +672,146 @@ check_function_markers()
...
+              nullptr },
+            { gen_instr_type(TRACE_TYPE_INSTR_RETURN, TID), ret2 },
+            // A tail call to the first function.
+            { gen_instr_type(TRACE_TYPE_INSTR_DIRECT_JUMP, TID), jmp },
```
derekbruening, line 723: This is not a tail call.  It is internal control flow within the same function.  A tail call would be something like targeting label2==ret2 rather than label1 and thus having its return skip over this function and go to the caller.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -672,6 +672,146 @@ check_function_markers()
...
+        instr_t *call1 = XINST_CREATE_call(GLOBAL_DCONTEXT, opnd_create_instr(label1));
+
+        instr_t *label2 = INSTR_CREATE_label(GLOBAL_DCONTEXT);
+        instr_t *call2 = XINST_CREATE_call(GLOBAL_DCONTEXT, opnd_create_instr(label2));
+        instr_t *ret2 = XINST_CREATE_return(GLOBAL_DCONTEXT);
+
+        instr_t *jmp = XINST_CREATE_jump(GLOBAL_DCONTEXT, opnd_create_instr(label1));
+
+        instrlist_t *ilist = instrlist_create(GLOBAL_DCONTEXT);
+        instrlist_append(ilist, call1);
```
derekbruening, line 699: A call whose target is the next instruction that is used as a regular call (and not as a mechanism to get the return address) is very weird as it will have an infinite loop.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -672,6 +672,146 @@ check_function_markers()
...
+            { gen_marker(TID, TRACE_MARKER_TYPE_FILETYPE, OFFLINE_FILE_TYPE_ENCODINGS),
+              nullptr },
+            { gen_instr_type(TRACE_TYPE_INSTR_DIRECT_CALL, TID), call1 },
+            { gen_marker(TID, TRACE_MARKER_TYPE_FUNC_ID, 1), nullptr },
+            { gen_marker(TID, TRACE_MARKER_TYPE_FUNC_RETADDR,
+                         BASE_ADDR + call_instr_length),
+              nullptr },
+
+            { gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_EVENT,
+                         BASE_ADDR + call_instr_length),
```
derekbruening, line 785: This is confusing as the return address equals the call target: see above on why this is strange.  Probably better to have a more normal code sequence.  (I don't think this is testing some valuable corner case as with the addresses all the same it won't as easily find errors.)

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -672,6 +672,146 @@ check_function_markers()
...
+            return false;
+    }
+    // Correctly handle kernel transfer, sigreturn, nested function calls
+    // including tailcalls.
+    {
+        // TODO i#5949: For WOW64 instr_is_syscall() always returns false, so our
+        // checks do not currently work properly there.
+#if !defined(WINDOWS) || defined(X64)
+        // XXX: Just like raw2trace_unit_tests, we need to create a syscall instruction
+        // and it turns out there is no simple cross-platform way.
```
derekbruening, line 741: I would again question why we are bothering to add encodings here: they're just adding complexity.  Are there enough sequential instructions below that manually listing the PC values is more complex?

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -672,6 +672,146 @@ check_function_markers()
...
+              nullptr },
+            { gen_marker(TID, TRACE_MARKER_TYPE_FUNC_ID, 1), nullptr },
+
+            { gen_instr_type(TRACE_TYPE_INSTR_RETURN, TID), ret2 },
+
+            { gen_instr_type(TRACE_TYPE_INSTR_RETURN, TID), sig_ret },
+            { gen_instr(TID), sys },
+            { gen_marker(TID, TRACE_MARKER_TYPE_SYSCALL, 15), nullptr },
+            { gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_XFER,
+                         BASE_ADDR + call_instr_length),
```
derekbruening, line 800: This marker value looks incorrect: doesn't the checker complain about this?  It should be the syscall PC + its length.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -672,6 +672,146 @@ check_function_markers()
...
+              nullptr },
+
+            { gen_instr_type(TRACE_TYPE_INSTR_DIRECT_JUMP, TID), jmp },
```
derekbruening, line 803: This is also incorrect and the checker should complain: the signal interrupted `BASE_ADDR + call_instr_length` but we didn't go back there.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -669,6 +679,13 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                       << "marker type " << memref.marker.marker_type << " value 0x"
                       << std::hex << memref.marker.marker_value << std::dec << "\n";
         }
+        // The interruption point is not a target for a return. Zero is push as
```
derekbruening, line 682: grammar: s/push/pushed/

derekbruening, line 682: Not sure the first sentence is needed?  I think it adds confusion?

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: I rewrote the code to remove encoding, and the labels.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: I rewrote the test to remove encoding, and use a fixed size like tests above.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: I rewrote the test, please take another look.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: I've removed the encodings.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: Yes, no encodings are needed . I've removed them.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: I rewrote the code, please take another look.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: The checker didn't complain. I rewrote the code, and used new PCs, PTAL.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: Please take another look at the rewritten test.

clients/drcachesim/tools/invariant_checker.cpp
ivankyluk: fixed.

ivankyluk: I've removed the first sentence.

clients/drcachesim/tests/invariant_checker_test.cpp
derekbruening: Not seeing a comment for this line?

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -672,6 +674,88 @@ check_function_markers()
...
+            gen_instr_type(TRACE_TYPE_INSTR_DIRECT_CALL, TID, CALL_PC_2, CALL_SZ),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_ID, 3),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_RETADDR, CALL_PC_2 + CALL_SZ),
+            gen_instr_type(TRACE_TYPE_INSTR_RETURN, TID, CALL_PC_2 + CALL_SZ, RETURN_SZ),
+
+            gen_instr_type(TRACE_TYPE_INSTR_RETURN, TID, CALL_PC + CALL_SZ, RETURN_SZ),
+
+#    if defined(X86_64)
+            gen_instr_encoded(CALL_PC + CALL_SZ,
+                              { 0x0f, 0x05 }), // 0x7fcf3b9d: 0f 05 syscall
```
derekbruening, line 733: I didn't think anything related to these checks looked at the encoding -- plus, the filetype doesn't list encodings so nothing will look at the encoding.  This can just be a `gen_instr` right?  Ditto below.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -672,6 +674,88 @@ check_function_markers()
...
+                           CALL_PC_2 + CALL_SZ + RETURN_SZ, 5),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_ID, 2),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_RETADDR, CALL_PC + CALL_SZ),
+        };
+        if (!run_checker(memrefs, false))
+            return false;
+    }
+    // Correctly handle kernel transfer, sigreturn, nested function calls
+    // including tailcalls.
+#if defined(X86_64) || defined(X86_32) || defined(ARM_64)
```
derekbruening, line 712: The ifdefs can go away b/c the encodings went away right?

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -672,6 +674,88 @@ check_function_markers()
...
+        std::vector<memref_t> memrefs = {
+            gen_instr(TID, 1),
+            gen_instr_type(TRACE_TYPE_INSTR_DIRECT_CALL, TID, CALL_PC, CALL_SZ),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_ID, 2),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_RETADDR, CALL_PC + CALL_SZ),
+
+            gen_instr_type(TRACE_TYPE_INSTR_DIRECT_CALL, TID, CALL_PC_2, CALL_SZ),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_ID, 3),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_RETADDR, CALL_PC_2 + CALL_SZ),
+            gen_instr_type(TRACE_TYPE_INSTR_RETURN, TID, CALL_PC_2 + CALL_SZ, RETURN_SZ),
```
derekbruening, line 700: This sequence is impossible: it is "call; ret; jmp" with the ret targeting the jmp when the ret has to target itself.  I know this is about just the specific checks in question but it would be best to make a feasible PC sequence to avoid readers getting confused.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -672,6 +674,88 @@ check_function_markers()
...
+    // including tailcalls.
+#if defined(X86_64) || defined(X86_32) || defined(ARM_64)
+    {
+        std::vector<memref_t> memrefs = {
+            gen_instr(TID, 1),
+            gen_instr_type(TRACE_TYPE_INSTR_DIRECT_CALL, TID, CALL_PC, CALL_SZ),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_ID, 2),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_RETADDR, CALL_PC + CALL_SZ),
+
+            gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_EVENT, CALL_PC + CALL_SZ),
```
derekbruening, line 720: I believe in the prior review round I suggested avoiding this unusual call targeting the very next instruction, for 2 reasons: for a feasible PC sequence and to avoid confusion by not having the interruption PC equal the return address.  As a reader is it harder to understand an impossible arrangement.  Maybe you were using the encodings to try to enforce a realistic sequence?  But it was this same sequence, just with labels which are invisible.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -669,6 +679,12 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                       << "marker type " << memref.marker.marker_type << " value 0x"
                       << std::hex << memref.marker.marker_value << std::dec << "\n";
         }
+        // Zero is pushed as a sentinel. Ths push matches the return used by post
```
derekbruening, line 682: typo: "This"

derekbruening (commented):
Still having trouble reading the test cases because they seem to be impossible sequences of instructions.

clients/drcachesim/tools/invariant_checker.cpp
ivankyluk: fixed.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: I've updated the sequence to fix the addresses, and added more comments. Please take another look.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: I've updated the sequence a add comment to describe the jump  which jumps back to the beginning of the function, and have a function marker.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: I've updated the sequence to use a feasible sequence as the test, and added more comments. Please take another look.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: I've updated the code sequence to use a normal code sequence.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: I've updated the sequence accordingly.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: That's right, I've removed the ifdef.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -672,6 +672,96 @@ check_function_markers()
         if (!run_checker(memrefs, false))
             return false;
     }
+    // Correctly skip return address check when the return address is
+    // unavailable.
+    {
+        std::vector<memref_t> memrefs = {
+            gen_instr(TID, 1),
+            gen_instr_type(TRACE_TYPE_INSTR_DIRECT_JUMP, TID, CALL_PC, CALL_SZ),
```
derekbruening, line 680: nit: Kind of odd to have a jump at a pc named CALL_PC and to have its size be CALL_SZ.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -672,6 +672,96 @@ check_function_markers()
...
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_ID, 2),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_RETADDR, /*pc=*/123456),
+        };
+        if (!run_checker(memrefs, false))
+            return false;
+    }
+    // Correctly handle nested function calls including tailcalls.
+    {
+        constexpr addr_t BASE_ADDR = 100;
```
derekbruening, line 689: I think `s/ADDR/PC/` will make these clearer; here and the 2 below.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -672,6 +672,96 @@ check_function_markers()
...
+            gen_instr_type(TRACE_TYPE_INSTR_DIRECT_CALL, TID, BASE_ADDR, CALL_SZ),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_ID, 1),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_RETADDR, BASE_ADDR + CALL_SZ),
+
+            // Call function 2.
+            gen_instr_type(TRACE_TYPE_INSTR_DIRECT_CALL, TID, FUNC1_BASE, CALL_SZ),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_ID, 2),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_RETADDR, FUNC1_BASE + CALL_SZ),
+
+            gen_instr(TID, FUNC2_BASE, 8),
```
derekbruening, line 705: `/*size=*/8` or better a named constant since it's used in more than one place (line below).

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -672,6 +672,96 @@ check_function_markers()
...
+            // Return from function 2.
+            gen_instr_type(TRACE_TYPE_INSTR_RETURN, TID, FUNC2_BASE + 8, RETURN_SZ),
+
+            gen_instr(TID, FUNC1_BASE + CALL_SZ, 8),
+            // Jump back to the beginning of function 1.
+            gen_instr_type(TRACE_TYPE_INSTR_TAKEN_JUMP, TID, FUNC1_BASE + CALL_SZ + 8),
```
derekbruening, line 711: So this is tail recursion: put that in the comment.  I was going to say an unconditional jump is what would normally be used for a tailcall but I guess tail recursion might have a conditional as you have here.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -672,6 +672,96 @@ check_function_markers()
...
+            // Call function 2.
+            gen_instr_type(TRACE_TYPE_INSTR_DIRECT_CALL, TID, FUNC1_BASE, CALL_SZ),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_ID, 2),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_RETADDR, FUNC1_BASE + CALL_SZ),
+
+            gen_instr(TID, FUNC2_BASE, 8),
+            // Return from function 2.
+            gen_instr_type(TRACE_TYPE_INSTR_RETURN, TID, FUNC2_BASE + 8, RETURN_SZ),
+
+            gen_instr(TID, FUNC1_BASE + CALL_SZ, 8),
```
derekbruening, line 709: Again since 8 used twice I would make a named constant (could use same as the other one as NON_BRANCH_SZ or sthg).

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -672,6 +672,96 @@ check_function_markers()
...
+            // Jump back to the beginning of function 1.
+            gen_instr_type(TRACE_TYPE_INSTR_TAKEN_JUMP, TID, FUNC1_BASE + CALL_SZ + 8),
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_ID, 1),
+            // The return address should be the same as the return address of
+            // the original call to function 1.
+            gen_marker(TID, TRACE_MARKER_TYPE_FUNC_RETADDR, BASE_ADDR + CALL_SZ),
```
derekbruening, line 715: OK so this is:
```
BASE_PC:
  call FUNC1_PC
..
FUNC1_PC:
  call FUNC2_PC
  xx
  jz FUNC1_PC
...
FUNC2_PC:
  xx
  ret
```
OK.  Maybe put that sequence in a comment at the top to make it easier to understand?

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -672,6 +672,96 @@ check_function_markers()
...
+        };
+        if (!run_checker(memrefs, false))
+            return false;
+    }
+    // Correctly handle kernel transfer, sigreturn, nested function calls
+    // including tailcalls.
+    {
+        constexpr addr_t BASE_ADDR = 100;
```
derekbruening, line 723: I would again `s/ADDR/PC/` for all of these.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -672,6 +672,96 @@ check_function_markers()
...
+            gen_marker(TID, TRACE_MARKER_TYPE_TIMESTAMP, 16),
+            gen_marker(TID, TRACE_MARKER_TYPE_CPU_ID, 3),
+            // syscall xfer.
+            gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_XFER, SYSCALL_BASE + SYSCALL_SZ),
+            gen_marker(TID, TRACE_MARKER_TYPE_TIMESTAMP, 17),
+            gen_marker(TID, TRACE_MARKER_TYPE_CPU_ID, 3),
+        };
+        if (!run_checker(memrefs, false))
+            return false;
+    }
```
derekbruening, line 764: It may be worth testing a nested signal and a back-to-back signal as well.

derekbruening (approved):

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: I've added a test for nested signal, and another one for back to back signal.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: I've renamed name to JUMP_PC, and JUMP_SZ.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: I've added a constant for it.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: done.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: done.

Issue #6262, opened 2023-08-11
Provide a way to query cpu id for scheduler output streams


prasun3:
**Add a pointer to any prior users list discussion.**
https://groups.google.com/g/DynamoRIO-Users/c/mSbhO2PRz90

While running in MAP_TO_RECORDED_OUTPUT mode, we need a way to query which cpu is running on a given output stream.

**Is your feature request related to a problem? Please describe.**
During simulation, we may want to preserve CPU numbering to reflect the original topology.

**Describe the solution you'd like**
We could have an API e.g., `scheduler.get_output_cpuid(ordinal)`

Do you have any implementation in mind for this feature?
Yes

**Describe alternatives you've considered**
None

**Additional context**
None

PR #6263, opened 2023-08-11
i#6262: query cpu id for scheduler output streams

prasun3:
Added an API to query cpu id

Issue: #6262

### DIFF
--- a/clients/drcachesim/scheduler/scheduler.h
+++ b/clients/drcachesim/scheduler/scheduler.h
@@ -778,6 +778,18 @@ public:
         return &outputs_[ordinal].stream;
     }
 
+    /**
+     * Returns the cpu number for the 'ordinal'-th output stream. This is only valid for
+     * MAP_TO_RECORDED_OUTPUT and returns -1 for other mappings.
+     */
+    virtual int
+    get_output_cpuid(output_ordinal_t ordinal)
+    {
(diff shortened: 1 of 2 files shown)


clients/drcachesim/scheduler/scheduler.h
```diff
@@ -778,6 +778,18 @@ template <typename RecordType, typename ReaderType> class scheduler_tmpl_t {
         return &outputs_[ordinal].stream;
     }
 
+    /**
+     * Returns the cpu number for the 'ordinal'-th output stream. This is only valid for
+     * MAP_TO_RECORDED_OUTPUT and returns -1 for other mappings.
+     */
+    virtual int
+    get_output_cpuid(output_ordinal_t ordinal)
```
derekbruening, line 786: We figured this would be on the stream, not the scheduler.  We may also want this for other modes, including core-oriented analysis tool modes #5694, where it has to be on the stream as the scheduler is not exposed through the analyzer framework.

derekbruening: I ended up needing this for #5694 and I put it into PR #6279 -- does that meet your needs as well?

prasun3: Thanks! Looks like this should be good.

Issue #5259, opened 2021-12-20
32-bit support for scatter-gather variations that have qword index or value


abhinav92003:
`drx_expand_scatter_gather` in drx does not support expansion for scatter/gather instrs with qword index or value on 32-bit systems.

https://github.com/DynamoRIO/dynamorio/blob/686f1a118c045c74f6a8446aef5508ad9cddd6fc/ext/drx/drx.c#L2427

Issue #5005, opened 2021-07-12
Verify drx state restoration for scatter/gather expansion with other instrumentation


abhinav92003:
`drx_restore_state_scatter_gather` performs state restoration for blocks containing the scatter/gather expansion using various state machines. As noted by this comment, https://github.com/DynamoRIO/dynamorio/blob/6d84fea04a036038db5a3af2e979e77d2cd356c0/ext/drx/drx.c#L2540 there may be issues in this due to other instrumentation added by clients. In addition to the cases described there, there could be instrumentation added by clients that use `gpr_scratch_index` and pass this check: https://github.com/DynamoRIO/dynamorio/blob/6d84fea04a036038db5a3af2e979e77d2cd356c0/ext/drx/drx.c#L3298, e.g. loads/stores that use the mcontext base are not recognised by `instr_is_reg_spill_or_restore`, and also load of the mcontext base itself is not recognised by `instr_is_reg_spill_or_restore` after PR #4987.

This issue is to investigate those cases and make this logic more robust if needed. One of the ideas is what's mentioned in the comment there already: use emulation labels in passed down instruction list (#3801) to make it easier to recognise the scatter/gather expansion instrs.

PR #6267, opened 2023-08-16
i#5036 A64 scatter/gather, part 1: Expand scalar+vector loads

jackgallagher-arm:
Creates an AArch64 version of drx_expand_scatter_gather() and tests for it. So far only SVE scalar+vector loads are supported. Support and tests for more instructions will follow in future commits.

State restore is also not yet supported and will be implemented when i#5365 is complete.

Issue: #5036


ext/drx/scatter_gather_aarch64.c
```diff
@@ -38,6 +38,479 @@
...
+    if (!info->fragment_info.app_code_consistent) {
+        /* Can't verify application code.
+         * XXX i#2985: is it better to keep searching?
+         */
+        return true;
+    }
+    instr_init(drcontext, &inst);
+    byte *pc = decode(drcontext, dr_fragment_app_pc(info->fragment_info.tag), &inst);
+    if (pc != NULL) {
+        if (instr_is_gather(&inst) || instr_is_gather(&inst)) {
```
derekbruening, line 117: Two identical calls: presumably one is supposed to be scatter.

ext/drx/scatter_gather_aarch64.c
```diff
@@ -38,6 +38,479 @@
...
+            /* TODO i#5365, i#5844: Restore the scratch predicate register.
+             *                      We need to add support for handling SVE state during
+             *                      signals first.
```
derekbruening, line 120: This lack of state restore should be listed in the PR description.

ext/drx/scatter_gather_aarch64.c
```diff
@@ -38,6 +38,479 @@
...
+        /* Can't verify application code.
+         * XXX i#2985: is it better to keep searching?
+         */
+        return true;
+    }
+    instr_init(drcontext, &inst);
+    byte *pc = decode(drcontext, dr_fragment_app_pc(info->fragment_info.tag), &inst);
+    if (pc != NULL) {
+        if (instr_is_gather(&inst) || instr_is_gather(&inst)) {
+            /* TODO i#5365, i#5844: Restore the scratch predicate register.
```
derekbruening, line 118: i#5036

ext/drx/scatter_gather_aarch64.c
```diff
@@ -38,6 +38,479 @@
...
+             *                      We need to add support for handling SVE state during
+             *                      signals first.
+             */
+            DR_ASSERT_MSG(false, "NYI i5365 i5844");
```
derekbruening, line 122: i#5036

ext/drx/scatter_gather_aarch64.c
```diff
@@ -38,6 +38,479 @@
...
+    tls_idx = drmgr_register_tls_field();
+    if (tls_idx == -1)
+        return false;
+
+    if (!drmgr_register_thread_init_event(drx_thread_init) ||
+        !drmgr_register_thread_exit_event(drx_thread_exit))
+        return false;
+
+    return true;
+}
```
derekbruening, line 149: This entire function is identical to the x86 one: share it?

ext/drx/scatter_gather_aarch64.c
```diff
@@ -38,6 +38,479 @@
...
+    if (dr_atomic_load32(&drx_scatter_gather_expanded) == 0) {
+        /* Nothing to do if nobody had never called expand_scatter_gather() before. */
+        return true;
+    }
+    if (!info->fragment_info.app_code_consistent) {
+        /* Can't verify application code.
+         * XXX i#2985: is it better to keep searching?
+         */
+        return true;
+    }
```
derekbruening, line 113: This is all identical to the x86 one too: share this?

ext/drx/scatter_gather_aarch64.c
```diff
@@ -38,6 +38,479 @@
...
+        DRX_CASE(ld1rqd, 1, OPSZ_8, false, DRX_NORMAL_FAULTING);
+#undef DRX_CASE
+
+    default: DR_ASSERT_MSG(false, "Invalid scatter/gather instruction");
+    }
+}
+
+#define EMIT(op, ...)    \
+    instrlist_preinsert( \
+        bb, sg_instr, INSTR_XL8(INSTR_CREATE_##op(drcontext, __VA_ARGS__), orig_app_pc))
```
derekbruening, line 284: With these implicit variables it is probably best to move this define inside the function (or else make these actual parameters).

ext/drx/scatter_gather_aarch64.c
```diff
@@ -38,6 +38,479 @@
...
+                     reg_id_t scratch_gpr, reg_id_t min_register, reg_id_t max_register,
+                     size_t slot_offset, opnd_size_t reg_size)
+{
+    /* Search the instruction for an unused register we will use as a temp. */
+    reg_id_t reg;
+    for (reg = min_register; reg <= max_register; ++reg) {
+        if (!instr_uses_reg(where, reg))
+            break;
+    }
+
```
derekbruening, line 449: Assert you found an unused register

ext/drx/scatter_gather_aarch64.c
```diff
@@ -38,6 +38,479 @@
...
+    drmgr_insert_read_tls_field(drcontext, tls_idx, bb, where, scratch_gpr);
```
derekbruening, line 450: You could get directly-addressable spill space: this is an extra load.  But given the heavyweight unrolling here in general it is probably ok and some platforms do have limited direct slots.

ext/drx/scatter_gather_aarch64.c
```diff
@@ -38,6 +38,479 @@
...
+
+    /* ldr scratch_gpr, [scratch_gpr, #slot_offset] */
+    instrlist_meta_preinsert(
+        bb, where,
+        INSTR_CREATE_ldr(drcontext, opnd_create_reg(scratch_gpr),
+                         OPND_CREATE_MEMPTR(scratch_gpr, slot_offset)));
```
derekbruening, line 456: So there is no check that the same slot isn't used again before a restore is done?

ext/drx/scatter_gather_aarch64.c
```diff
@@ -48,8 +521,113 @@
...
+    }
+    DR_ASSERT(sg_instr != NULL);
+
+    scatter_gather_info_t sg_info;
+    bool res = false;
+    get_scatter_gather_info(sg_instr, &sg_info);
+
+    if (!(sg_info.is_load && reg_is_z(sg_info.index_reg) &&
+          sg_info.faulting_behavior == DRX_NORMAL_FAULTING))
+        return true;
```
derekbruening, line 542: I don't understand this: what condition is this that just returns true?  A comment would help.

ext/drx/scatter_gather_aarch64.c
```diff
@@ -48,8 +521,113 @@
...
     if (expanded != NULL)
         *expanded = false;
-    return drmgr_current_bb_phase(drcontext) == DRMGR_PHASE_APP2APP;
+
+    if (drmgr_current_bb_phase(drcontext) != DRMGR_PHASE_APP2APP)
+        return false;
+
+    instr_t *sg_instr = NULL;
+    if (!scatter_gather_split_bb(drcontext, bb, &sg_instr)) {
+        return true;
```
derekbruening, line 532: This seems backward: if it fails to split we return "successfully expanded"?  Or is the return value just confusing?  Consider renaming the function or reversing its return values so this reads more naturally.

ext/drx/scatter_gather_shared.h
```diff
@@ -0,0 +1,78 @@
...
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+/* DynamoRio eXtension utilities */
+
+#include "dr_api.h"
+
+/* Make each scatter or gather instruction be in their own basic block.
```
derekbruening, line 38: Explain what the return value means (see above where the call site was confusing).

suite/tests/client-interface/drx-scattergather-bbdup.dll.c
```diff
@@ -61,6 +61,7 @@ inscount(uint num_instrs)
     global_sg_count += num_instrs;
 }
 
+#if defined(X86)
```
derekbruening, line 64: What about AArch64 equivalent tests -- are these state restore tests at points in the expansion where key registers are clobbered?  Should there be a TODO i#5036 here?

suite/tests/client-interface/drx-scattergather.dll.c
```diff
@@ -58,6 +58,7 @@ inscount(uint num_instrs)
     global_sg_count += num_instrs;
 }
 
+#if defined(X86)
```
derekbruening, line 61: Same thing here.

suite/tests/client-interface/drx-scattergather-aarch64.cpp
```diff
@@ -0,0 +1,1170 @@
...
+constexpr predicate_reg_value128_t UNINITIALIZED_PREDICATE = 0xDEAD;
+
+enum class element_size_t {
+    BYTE = 1,
+    HALF = 2,
+    SINGLE = 4,
+    DOUBLE = 8,
+};
+
+/*! Exhaustive lists of valid 128-bit vl predicate register values for for
```
derekbruening, line 73: Should be a normal comment.  Ditto below (except perhaps the ones using `\param`). I would expect a Doxygen-documented file to have *everything* (all types and functions) with Doxygen comments; this scattering in a test file where Doxygen is never run on this is what seems unusual.  Just a comment for a local reader here seems like it should be a regular comment.  (If it were Doxygen it should follow the entire rest of the code base and use `/**` not `/*!`.)

derekbruening, line 73: "for for"

suite/tests/client-interface/drx-scattergather-aarch64.cpp
```diff
@@ -0,0 +1,1170 @@
...
+ * \param element_size The size of the elements in the mask.
+ * \return             \c true if \p element is active in the predicate \p mask.
+ */
+bool
+element_is_active(size_t element, predicate_reg_value128_t mask,
+                  element_size_t element_size)
+{
+    const auto element_size_bytes = static_cast<size_t>(element_size);
+
+    return (mask >> (element_size_bytes * element)) & 1 != 0;
```
derekbruening, line 100: style: normally we use TESTALL/TESTANY for bit checks

suite/tests/client-interface/drx-scattergather-aarch64.cpp
```diff
@@ -0,0 +1,1170 @@
...
+        0x1010, 0x1011, 0x1100, 0x1101, 0x1110, 0x1111 } },
+    { element_size_t::DOUBLE, { 0x0000, 0x0001, 0x0100, 0x0101 } },
+};
+
+enum test_result_t {
+    FAIL,
+    PASS,
+};
+
+/*!
```
derekbruening, line 88: OK, using `\param`, but match the `/**` style everywhere else in the codebase instead of `/*!`.  There is no pass that's running this through doxygen, so syntax errors or missing links or anything is not going to be caught.  Ditto below.

suite/tests/client-interface/drx-scattergather-aarch64.cpp
```diff
@@ -0,0 +1,1170 @@
...
+{
+    print("0x");
+    for (size_t i = 0; i < value.size; i++) {
+        print("%02x", value.data[i]);
+    }
+}
+
+/*!
+ * Print a predicate register value as a binary number. Each bit is printed with a space
+ * inbetween so that the bit will line up vertically with the corresponding byte of a
```
derekbruening, line 169: "in between"

suite/tests/client-interface/drx-scattergather-aarch64.cpp
```diff
@@ -0,0 +1,1170 @@
...
+ *     vec:  0x12345678
+ *     pred: 0b 0 1 0 1
+ */
+void
+print_predicate(const scalable_reg_value_t &value)
+{
+    print("0b");
+    for (size_t byte_i = 0; byte_i < value.size; byte_i++) {
+        for (unsigned bit = 0; bit < 8; bit++) {
+            if (((value.data[byte_i] >> bit) & 1) == 0)
```
derekbruening, line 180: style: normally we use TESTALL/TESTANY for bit checks

suite/tests/client-interface/drx-scattergather-aarch64.cpp
```diff
@@ -0,0 +1,1170 @@
...
+        const auto reg_offset = pl_bytes * reg_num;
+        for (size_t i = 0; i < pl_bytes / sizeof(value); i++) {
+            const auto slice_offset = reg_offset + (sizeof(value) * i);
+            memcpy(&p[slice_offset], &value, sizeof(value));
+        }
+    }
+};
+
+struct test_register_data_t {
+    sve_register_file_t before; //!< Values the registers will be set to before the test.
```
derekbruening, line 240: More Doxygen and with the inconsistent style...ditto below.

derekbruening (commented):
Probably @abhinav92003 should take a look as the x86 scatter/gather author.  I did not finish the long suite/tests/client-interface/drx-scattergather-aarch64.cpp file but sending out what I have; mostly minor style nits.

CMakeLists.txt
```diff
@@ -621,6 +621,12 @@ if (X86 AND UNIX)
   check_avx512_processor_and_compiler_support(proc_supports_avx512)
```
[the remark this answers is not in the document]
jackgallagher-arm: > there are scatter/gather instructions in AArch32 right?
No there aren't any AArch32 scatter/gather instructions. SVE is AArch64-only and Neon does not have any scatter/gather instructions.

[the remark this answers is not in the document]
derekbruening: Ah, good to know.  Let's close another one then -- all else being equal we'd keep the older one, but #5036 seems to have more discussion/info inside it so I'll close #3837.

ext/drx/scatter_gather_aarch64.c
```diff
@@ -38,6 +38,481 @@
...
+
+    /* pfalse   scratch_pred.b */
+    EMIT(pfalse_sve, opnd_create_reg_element_vector(scratch_pred, OPSZ_1));
+
+    instr_t *loop_label = INSTR_CREATE_label(drcontext);
+    instr_t *end_label = INSTR_CREATE_label(drcontext);
+
+    instrlist_meta_preinsert(bb, sg_instr, loop_label);
+
+    /* pnext scratch_pred.elsz, mask_reg, scratch_pred.elsz */
```
abhinav92003, line 376: What does `elsz` mean?

ext/drx/scatter_gather_aarch64.c
```diff
@@ -38,6 +38,481 @@
...
+            sg_info->scalar_value_size, sg_info->extend_amount);
+
+        if (sg_info->is_scalar_value_signed) {
+            const reg_id_t ld_dst =
+                reg_resize_to_opsz(scratch_gpr, sg_info->element_size);
+            switch (sg_info->scalar_value_size) {
+            case OPSZ_1: EMIT(ldrsb, opnd_create_reg(ld_dst), mem); break;
+            case OPSZ_2: EMIT(ldrsh, opnd_create_reg(ld_dst), mem); break;
+            case OPSZ_4: EMIT(ldrsw, opnd_create_reg(ld_dst), mem); break;
+            default: DR_ASSERT_MSG(false, "Invalid scatter_gather_info_t data");
```
abhinav92003, line 404: No 8-byte version for signed load?

ext/drx/scatter_gather_aarch64.c
```diff
@@ -38,6 +38,481 @@
...
+
+    instrlist_meta_preinsert(bb, sg_instr, end_label);
+
+#undef EMIT
+}
+
+/* Spill a scratch predicate or vector register.
+ * TODO i#3844: drreg does not support spilling predicate regs yet, so we do it
+ * ourselves.
+ * When that support is available, this funcion can be replaced with a drreg API call.
```
abhinav92003, line 438: s/funcion/function/

suite/tests/CMakeLists.txt
```diff
@@ -2819,31 +2833,43 @@ if (NOT RISCV64) # TODO i#3544: Port tests to RISC-V 64
   use_DynamoRIO_extension(client.tls.dll drmgr)
 endif (NOT RISCV64)
 
-if (X86)
+if (X86 OR AARCH64)
   if (NOT MACOS)
     # XXX i#2985: The test's asm doesn't build with MacOS's clang.
-    tobuild_ci(client.drx-scattergather client-interface/drx-scattergather.c "" "" "")
+    set(client.drx-scattergather_client_source "client-interface/drx-scattergather.dll.c")
```
abhinav92003, line 2839: Do we expect to be able to reuse a lot of code from this dll file?

suite/tests/client-interface/drx-scattergather-aarch64.cpp
```diff
@@ -0,0 +1,1158 @@
...
+        }
+    }
+
+    return data;
+}
+
+size_t
+get_vl_bytes()
+{
+    const int returned_value = prctl(PR_SVE_GET_VL);
```
abhinav92003, line 120: Can we set this to static so it doesn't need to do a prctl call everytime?

ext/drx/scatter_gather_shared.h
```diff
@@ -0,0 +1,78 @@
...
+bool
+scatter_gather_split_bb(void *drcontext, instrlist_t *bb, OUT instr_t **sg_instr);
+
+typedef struct _scatter_gather_info_t {
+#if defined(X86)
+    bool is_evex;
+#endif
+    bool is_load;
+#if defined(AARCH64)
+    opnd_size_t element_size;
```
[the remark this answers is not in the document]
jackgallagher-arm: `scalar_index_size` makes sense for the scalar+vector instructions but it doesn't really fit for the other addressing modes which don't have a vector index
(vector+immediate instructions e.g `LD1SB { <Zt>.<T> }, <Pg>/Z, [<Zn>.S{, #<imm>}]`, scalar+scalar predicated contiguous accesses e.g `LD1SB { <Zt>.<T> }, <Pg>/Z, [<Xn|SP>, <Xm>]`, scalar+immediate `LD1SB { <Zt>.<T> }, <Pg>/Z, [<Xn|SP>{, #<imm>, MUL VL}]`)

ext/drx/scatter_gather_aarch64.c
```diff
@@ -38,6 +38,481 @@
...
+            DR_ASSERT_MSG(false, "NYI i#5365 i#5036");
+        }
+    }
+    instr_free(drcontext, &inst);
+    return true;
+}
+
+bool
+drx_scatter_gather_init()
+{
```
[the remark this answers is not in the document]
jackgallagher-arm: `drx_scatter_gather_init()` isn't a public function. Its called in `drx_init()` in `ext/drx/drx.c` which already does guard against multiple calls.

ext/drx/scatter_gather_aarch64.c
```diff
@@ -38,6 +38,481 @@
...
+ * }
+
+ * For example
+ *     ld1d   (%x0,%z26.d,lsl #3)[32byte] %p1/z -> %z27.d
+ * expands to:
+ *
+ *       dup    $0x00 lsl $0x00 -> %z27.d       ; Clear dst register
+ *       pfalse  -> %p0.b
+ *   loop:
+ *       pnext  %p1 %p0.d -> %p0.d              ; p1 = next active element of p0
```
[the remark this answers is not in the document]
jackgallagher-arm: > Re the comment: do you mean "p0 = next active element of p1"?
Yes. Ta.

> Does pnext also zero already-processed active elements?
Yes. Essentially it creates a mask where the only bit set is the next active element. I will clarify the comment. https://developer.arm.com/documentation/ddi0596/2021-12/SVE-Instructions/PNEXT--Find-next-active-predicate-?lang=en

ext/drx/scatter_gather_aarch64.c
```diff
@@ -38,6 +38,481 @@
...
+ *         scalar_store(src[e], base, offsets[e], mod);
+ * }
+
+ * For example
+ *     ld1d   (%x0,%z26.d,lsl #3)[32byte] %p1/z -> %z27.d
+ * expands to:
+ *
+ *       dup    $0x00 lsl $0x00 -> %z27.d       ; Clear dst register
+ *       pfalse  -> %p0.b
+ *   loop:
```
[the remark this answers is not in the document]
jackgallagher-arm: The trouble with unrolling this loop is that we don't know until run time how many iterations there are. The x86 version loops over each element of the vector but here we only loop over the active elements, and we don't know how many active elements there are at translation time because we don't know the value of the guarding predicate.

If we have access to the values of the predicate registers in the state restore function then we should be able to determine which element access faulted by looking at the value of the scratch_pred register.

ext/drx/scatter_gather_aarch64.c
```diff
@@ -38,6 +38,481 @@
...
+                          reg_id_t scratch_pred, app_pc orig_app_pc)
+{
+#define EMIT(op, ...)    \
+    instrlist_preinsert( \
+        bb, sg_instr, INSTR_XL8(INSTR_CREATE_##op(drcontext, __VA_ARGS__), orig_app_pc))
+
+    DR_ASSERT_MSG(reg_is_z(sg_info->index_reg), "Index must be a Z register");
+
+    if (sg_info->is_load) {
+        /* First we deal with the inactive elements. Gather loads are always zeroing so we
```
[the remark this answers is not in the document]
jackgallagher-arm: Also from that page:

> Note: Not all instructions have predicate options. Also, not all predicate operations have both merging and zeroing options. You must refer to the [SVE Architecture Supplement](https://developer.arm.com/documentation/ddi0584/latest) for the specification details of each instruction.

All gather loads and predicated contiguous loads are zeroing only.

ext/drx/scatter_gather_aarch64.c
```diff
@@ -38,6 +38,481 @@
...
+             * value of the active elements so we can use them as offsets. We do this by
+             * cpying a 0 value into the dst register using the inverse of the mask_reg as
+             * the governing predicate.
+             */
+
+            /* ptrue    scratch_pred.b */
+            EMIT(ptrue_sve, opnd_create_reg_element_vector(scratch_pred, OPSZ_1),
+                 opnd_create_immed_pred_constr(DR_PRED_CONSTR_ALL));
+
+            /* not      scratch_pred.b, scratch_pred/z, mask_reg.b */
```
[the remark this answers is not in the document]
jackgallagher-arm: No, I did mean `not` and `scratch_pred/z`. `not` is an alias for `eor` in AArch64 assembly, but the DynamoRIO codec doesn't support all instruction aliases so we need to encode it using the underlying `eor` encoding.

https://developer.arm.com/documentation/ddi0602/2023-06/SVE-Instructions/NOT--predicate---Bitwise-invert-predicate--an-alias-of-EOR--predicates--?lang=en

>`NOT <Pd>.B, <Pg>/Z, <Pn>.B`
is equivalent to
`EOR <Pd>.B, <Pg>/Z, <Pn>.B, <Pg>.B`

I put `not` in the comment because I think it is clearer what the intention is but I can see that if you don't happen to know that `not` is an alias for `eor`, then it does look like a mistake.

I should have just added an `INSTR_CREATE_not_sve_pred_b` macro so we can just use the alias to encode the instruction.

ext/drx/scatter_gather_aarch64.c
jackgallagher-arm: Its short for element size. The abbreviation is used in the codec but its a bit cryptic if you aren't familiar. I'll change it to `element_size`.

ext/drx/scatter_gather_aarch64.c
jackgallagher-arm: The signed loads sign extend the loaded value to the destination register size (either a 32-bit W register or 64-bit X register). It only makes sense for loads that are smaller than the register size. There is no `ldrsd` instruction (nor a `ld1sd` vector equivalent) because there are no bits to extend when you are loading a 64-bit value to a 64-bit register.

suite/tests/CMakeLists.txt
jackgallagher-arm: I'm not sure I understand what you are asking. The `drx-scattergather.dll` test client is used for both the x86 and AArch64 versions of the test. There are a few x86-specific bits which are `#ifdef`ed out but the rest is shared.

ext/drx/scatter_gather_aarch64.c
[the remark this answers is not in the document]
abhinav92003: Ack.

ext/drx/scatter_gather_aarch64.c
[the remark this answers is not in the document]
abhinav92003: Ah thanks. Should've read all the way through :)

suite/tests/CMakeLists.txt
abhinav92003: I was trying to weigh whether it will be cleaner to create a new dll test client file for aarch64, but let's keep the ifdefs and the same file for now.

ext/drx/scatter_gather_aarch64.c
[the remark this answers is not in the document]
abhinav92003: > The x86 version loops over each element of the vector but here we only loop over the active elements,

Well, if needed, we could loop over each element and skip the inactive elements on our own.

> If we have access to the values of the predicate registers in the state restore function then we should be able to determine which element access faulted by looking at the value of the scratch_pred register.

Hmm.. this should work. I like the loop better if it works.

I'm trying to recall if there was any other motivation for unrolling the loop in x86.

@derekbruening Looking at my notes from a couple years ago, I see
```
TODO: separate issue about making it a loop.
```
I don't see a filed issue though.

Maybe we discussed changing it to a loop but decided it's low priority?

ext/drx/scatter_gather_aarch64.c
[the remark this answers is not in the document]
abhinav92003: > I should have just added an INSTR_CREATE_not_sve_pred_b macro so we can just use the alias to encode the instruction.

Sure we could just do that. Or a clarifying comment would work too.

abhinav92003 (approved):

ext/drx/scatter_gather_aarch64.c
jackgallagher-arm: It depends on the instruction. Instructions (like pnext) which treat the register value as a mask generally need to know the element size, but bitwise operations (like pfalse, and, orr, etc) just operate on the register value and are conventionally written with the element size `b` to indicate it operates on every bit in the register.

ext/drx/scatter_gather_aarch64.c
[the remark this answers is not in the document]
derekbruening: My notes only have links to the following issues:
- #2985: problems instrumenting AVX2 vpgatherdd
- #5005: Verify drx state restoration for scatter/gather expansion with other instrumentation
- #5036: Expand A64 SVE scatter/gather memory access instructions
- #5259: 32-bit support for scatter-gather variations that have qword index or value
I don't have any further info written down about a loop.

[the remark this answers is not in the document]
abhinav92003: Okay.

Since all iterations are in a loop, the instrumentation will be repeated too. So we need to mark the first output pc entry and subsequent ones differently (like we do for x86 repstr; see TRACE_TYPE_INSTR_MAYBE_FETCH). This was simpler on x86 because of the unrolling.

https://github.com/DynamoRIO/dynamorio/blob/1b697e5847b6f5b5c277a13aa1e5aeb0af44a97e/clients/drcachesim/tracer/instru.cpp#L79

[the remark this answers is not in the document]
abhinav92003: I assume drmemtrace specific changes will come later? Maybe add that to the PR description?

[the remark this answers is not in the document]
derekbruening: We wanted to get rid of the MAYBE_FETCH concept altogether though: #4915, #4948.  So I'm not sure we want to add another instance that uses it if we can avoid it?  To avoid it we'd have to unroll the loop?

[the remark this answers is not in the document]
abhinav92003: We could unroll the loop or use one of the options discussed in https://github.com/DynamoRIO/dynamorio/issues/4915#issuecomment-1064242403 for repstr.

The simplest option seems to be D, which suggests removing the non-first entries in raw2trace.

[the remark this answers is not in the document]
jackgallagher-arm: For now at least, I have unrolled the loop. If we have a different solution in the future we can revisit making it a loop if we want to.

ext/drx/scatter_gather_shared.c
```diff
@@ -53,18 +71,95 @@ scatter_gather_split_bb(void *drcontext, instrlist_t *bb, OUT instr_t **sg_instr
...
+    if (first_app_is_scatter_gather && (sg_instr != NULL)) {
         *sg_instr = first_app;
     }
 
-    return first_app != NULL &&
-        (instr_is_gather(first_app) || instr_is_scatter(first_app));
+    return first_app_is_scatter_gather;
+}
+
+/* These architecture specific functions and defined in scatter_gather_${ARCH_NAME}.c */
```
abhinav92003, line 91: s/and/are/

suite/tests/client-interface/drx-scattergather-bbdup.dll.c
```diff
@@ -260,6 +260,10 @@ event_bb_app2app(void *drcontext, void *tag, instrlist_t *bb, bool for_trace,
         } else if (instr_is_scatter(instr)) {
             scatter_gather_present = true;
 #if defined(X86)
+            /* TODO i#5036: Port this code to AArch64 to test state restoration of
+             * clobbered predicate registers (when we have added support for state
+             *              restoration).
```
abhinav92003, line 265: nit: remove extra horizontal white-space. Also below.

ext/drx/scatter_gather_aarch64.c
```diff
@@ -58,14 +58,27 @@
...
 #define SVE_PREDICATE_SPILL_SLOT_SIZE SVE_MAX_PREDICATE_LENGTH_BYTES
 
-static int tls_idx;
 typedef struct _per_thread_t {
     void *scratch_pred_spill_slot; /* Storage for spilled predicate register. */
 } per_thread_t;
 
-static int drx_scatter_gather_expanded = 0;
+/* Track the state of manual spill slots for SVE registers.
+ * This corresponds to the splill slot storage in per_thread_t.
```
abhinav92003, line 66: s/splill/spill/

ext/drx/scatter_gather_aarch64.c
```diff
@@ -288,29 +242,51 @@ get_scatter_gather_info(instr_t *instr, OUT scatter_gather_info_t *sg_info)
  * from the scalar register Xn, to an offset read from the corresponding element of the
  * vector index register Zm.
  *
- * if (is_load)
- *     clear_inactive_elements(dst);
- * for (e=first_active_element(); e < num_elements; e =  next_active_element()) {
+ * The emitted code roughly implements this algorithm:
```
abhinav92003, line 245: Also add to comment why we decided to unroll the loop.

ext/drx/scatter_gather_aarch64.c
```diff
@@ -368,65 +348,64 @@ expand_scalar_plus_vector(void *drcontext, instrlist_t *bb, instr_t *sg_instr,
...
+
+        /* lastb    scratch_gpr, scratch_pred, index_reg.element_size */
+        EMIT(lastb_sve_scalar, opnd_create_reg(scratch_gpr),
+             opnd_create_reg(scratch_pred),
+             opnd_create_reg_element_vector(sg_info->index_reg, sg_info->element_size));
+
+        if (sg_info->is_load) {
+            /* ldr[bh]  scratch_gpr, [base_reg, scratch_gpr, mod #amount] */
+            opnd_t mem = opnd_create_base_disp_shift_aarch64(
+                sg_info->base_reg, scratch_gpr, sg_info->extend, sg_info->scaled, 0, 0,
```
abhinav92003, line 375: nit: s/0, 0/`/*disp=*/0, /*flags=*/0`/

ext/drx/scatter_gather_aarch64.c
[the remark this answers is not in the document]
abhinav92003: Thanks for making the change. I rather like the loop version so if we figure out a solution for repstr, we can consider using the same for scatter-gather (x86 and AArch64). Could you please also leave a comment on #4948 about this design decision?

abhinav92003 (approved):

suite/tests/client-interface/drx-scattergather-bbdup.dll.c
derekbruening: Generally, please reply with a message before resolving so it's clear why it was resolved.

derekbruening (approved):

CMakeLists.txt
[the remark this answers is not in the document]
jackgallagher-arm: Done

suite/tests/client-interface/drx-scattergather-bbdup.dll.c
jackgallagher-arm: > Should there be a TODO i#5036 here?
I have added TODOs to this file and drx-scattergather.dll.c

> Generally, please reply with a message before resolving so it's clear why it was resolved.

Sorry. I will do that from now on.

ext/drx/scatter_gather_aarch64.c
jackgallagher-arm: This was checking whether the instruction is one of the ones that is currently supported and returning straight away if not. I have taken it out now because it will be caught further down the function anyway.

ext/drx/scatter_gather_aarch64.c
jackgallagher-arm: We aren't returning "successfully expanded". There is a separate `OUT bool *expanded` argument to indicate whether anything was expanded and when we return `true` here `expanded` has been set to `false`. I have added comments here and to `scatter_gather_split_bb()` that I hope clarify how this works

clients/drcachesim/tests/scattergather-aarch64.templatex
```diff
@@ -0,0 +1,86 @@
...
+ld1d 32bit unpacked scaled offset uxtw: PASS
+ld1d 32bit unpacked scaled offset sxtw: PASS
+ld1d 32bit unpacked unscaled offset uxtw: PASS
+ld1d 32bit unpacked unscaled offset sxtw: PASS
+ld1d 64bit scaled offset: PASS
+ld1d 64bit unscaled offset: PASS
+ld1d 64bit unscaled offset Zt==Zm: PASS
+#endif /* __ARM_FEATURE_SVE */
+---- <application exited with code 0> ----
+Basic counts tool results:
```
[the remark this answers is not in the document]
jackgallagher-arm: Changing that would effect the equivalent x86 test as well so I think its out of scope for this PR

ext/drx/scatter_gather_aarch64.c
jackgallagher-arm: I did it this way mainly because that's what the x86 implementation does so it might not be the best solution for AArch64. I'm going to leave it as it is for now but I'll come back to it and look at optimising it as I build out support for the rest of the scatter/gather instructions.

PR #6274, opened 2023-08-18
i#5843 scheduler: Reduce default instr quantum to 500K

derekbruening:
Updates the default instruction count scheduling quantum to be smaller and consistent: 500K instructions instead of the 1M it was in the launcher and 10M in the scheduler.

Issue: #5843

### DIFF
--- a/clients/drcachesim/scheduler/scheduler.h
+++ b/clients/drcachesim/scheduler/scheduler.h
@@ -448,7 +448,7 @@ public:
          * specified by
          * #dynamorio::drmemtrace::scheduler_tmpl_t::scheduler_options_t::quantum_unit.
          */
-        uint64_t quantum_duration = 10 * 1000 * 1000;
+        uint64_t quantum_duration = 500 * 1000;
         /**
          * If > 0, diagnostic messages are printed to stderr.  Higher values produce
          * more frequent diagnostics.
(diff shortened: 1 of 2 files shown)


derekbruening:
Actually I'm rethinking this one: let's discuss further offline.

derekbruening:
I can't figure out how to retract the review request -- so I marked as Draft.

clients/drcachesim/scheduler/scheduler.h
```diff
@@ -448,7 +448,7 @@ template <typename RecordType, typename ReaderType> class scheduler_tmpl_t {
          * specified by
          * #dynamorio::drmemtrace::scheduler_tmpl_t::scheduler_options_t::quantum_unit.
          */
-        uint64_t quantum_duration = 10 * 1000 * 1000;
+        uint64_t quantum_duration = 500 * 1000;
```
abhinav92003, line 451: So the other default is just for the launcher which would override this one, whereas this one is for any uses of scheduler_options_t?

abhinav92003, line 451: Perhaps a quick note on why we think 500K is more appropriate (since we decided to switch from 1M to this, I assume there was a non-trivial reason).

abhinav92003 (approved):

clients/drcachesim/scheduler/scheduler.h
```diff
@@ -448,7 +448,7 @@ template <typename RecordType, typename ReaderType> class scheduler_tmpl_t {
          * specified by
```
abhinav92003, line 448: > I can't figure out how to retract the review request -- so I marked as Draft.

Didn't see this before. Retracting the approval.

abhinav92003 (changes requested):

derekbruening:
I think with the blocking-syscall switch points the original 10M is reasonable; it could even be made a little higher (20M+ is more like it for 3GHz 10ms IPC a little below 1).  Abandoning this change.

Issue #1556, opened 2014-11-28
mark conditional branches and string loops as predicated


derekbruening:
_From [bruen...@google.com](https://code.google.com/u/109494838902877177630/) on October 30, 2014 18:52:04_

The new predication API (xref issue #269 , issue #1181 ) should probably consider general
conditional branches to also be predicated.  For x86, OP_jcc is simple to
translate to DR_PRED_ constants -- but we'd have to add new ones or ensure
DR_PRED_COMPLEX's assumptions hold for OP_jecxz.

What about OP_rep_movs for when ecx is 0?

_Original issue: http://code.google.com/p/dynamorio/issues/detail?id=1556_

derekbruening:
I think arm and aarch64 conditional branches are already marked as predicated?  Except for OP_cbz and OP_tbz -- similarly to OP_jecxz.

PR #6284, opened 2023-08-29
i#1556: Mark x86 rep instructions as predicated.

ivankyluk:
i#1556: Mark x86 rep instructions as predicated.

Issue: #1556


core/ir/x86/decode_table.c
```diff
@@ -9597,33 +9597,33 @@ const instr_info_t repne_extensions[][6] = {
...
   },
   { /* repne extension 1 */
     {OP_cmps,       0xa70000, "cmps",         axSI, axDI, Xv, Yv, axSI, xop_next, (fW6|fRD), tne[0][0]},
     {OP_CONTD,      0xa70000, "cmps",         xx, xx, axDI, xx, xx, no, (fW6|fRD), END_LIST},
-    {OP_rep_cmps,   0xf3a70000, "rep cmps",   axSI, axDI, Xv, Yv, axSI, xop_next, (fW6|fRD|fRZ), tne[0][2]},
-    {OP_CONTD,      0xf3a70000, "rep cmps",   axCX, xx, axDI, axCX, xx, no, (fW6|fRD), END_LIST},
-    {OP_repne_cmps, 0xf2a70000, "repne cmps", axSI, axDI, Xv, Yv, axSI, xop_next, (fW6|fRD|fRZ), tne[0][4]},
-    {OP_CONTD,      0xf2a70000, "repne cmps", axCX, xx, axDI, axCX, xx, no, (fW6|fRD), END_LIST},
+    {OP_rep_cmps,   0xf3a70000, "rep cmps",   axSI, axDI, Xv, Yv, axSI, xop_next|predcx, (fW6|fRD|fRZ), tne[0][2]},
+    {OP_CONTD,      0xf3a70000, "rep cmps",   axCX, xx, axDI, axCX, xx, predcx, (fW6|fRD), END_LIST},
```
ivankyluk, line 9608: Copy the comment from https://github.com/DynamoRIO/dynamorio/pull/6283

> Hmm this is missing fRZ -- ditto for the others. Does the decoder take the flags from the final entry? I think it does.

I might have missed it.

read_instruction in ir/x86/decode.c 

    } else if (info->type == REP_EXT) {
        /* discard old info, get new one */
        int code = (int)info->code;
        int idx = (di->rep_prefix ? 2 : 0);
        info = &rep_extensions[code][idx];
        if (di->rep_prefix)
            di->rep_prefix = false;

points to the first entry,  and decode_opcopde/decode_common sets eflags after getting the entry by calling read_instruction.

So, it appears the eflag from the first entry is used.

derekbruening: Still seems best to give them the same values (could even have decoder check in debug build).

ivankyluk: I've updated the table accordingly, and modified the test to cover eflags.

ivankyluk: PTAL.

derekbruening (approved):

PR #6285, opened 2023-08-30
drraw2trace. Change the type of error codes from string to int.

yuriershov:
Error codes type changed from string to int. Performance of drraw2trace is increased ~15%.

### DIFF
--- a/clients/drcachesim/tracer/raw2trace.cpp
+++ b/clients/drcachesim/tracer/raw2trace.cpp
@@ -855,8 +855,8 @@ raw2trace_t::process_syscall_pt(raw2trace_thread_data_t *tdata, uint64_t syscall
         DR_ASSERT(syscall_idx == 0);
         syscall_pt_entry_t header[PT_METADATA_PDB_HEADER_ENTRY_NUM];
         if (!tdata->kthread_file->read((char *)&header[0], PT_METADATA_PDB_HEADER_SIZE)) {
-            tdata->error = "Unable to read the PDB header of PT metadate form kernel thread log "
-                   "file";
+            tdata->error = "Unable to read the PDB header of PT metadate form kernel "
+                   "thread log file";
             return false;
         }
         if (header[PDB_HEADER_DATA_BOUNDARY_IDX].pt_metadata_boundary.type !=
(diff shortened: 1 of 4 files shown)


clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1188,16 +1268,16 @@ class raw2trace_t {
      * all buffers from any one traced thread must be processed by the same worker
      * thread, both for correct ordering and correct synchronization.
      */
-    std::string
+    raw2trace_error_t
```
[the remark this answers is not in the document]
derekbruening: What do you think of a solution where `bool` is the return value with `false` indicating an error and a separate `get_error_string()` which returns a descriptive string?  That then allows our subclass (and anyone else's) to have new error types without having to try to add them to the public base class.  It also eliminates the separate string array which has maintenance costs for keeping it in sync.  It should have the same performance of the enum solution.

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -2437,11 +2391,9 @@ raw2trace_t::adjust_and_emit_rseq_buffer(raw2trace_thread_data_t *tdata, addr_t
     tdata->rseq_buffering_enabled_ = false;
 
     log(4, "Writing out rseq buffer: %zd entries\n", tdata->rseq_buffer_.size());
-    auto success =
-        write(tdata, &tdata->rseq_buffer_[0], &tdata->rseq_buffer_.back() + 1,
-              tdata->rseq_decode_pcs_.data(), tdata->rseq_decode_pcs_.size());
-    if (!success)
-        return success;
+    if (! write(tdata, &tdata->rseq_buffer_[0], &tdata->rseq_buffer_.back() + 1,
```
derekbruening, line 2394: nit: Extra space after `!` (surprised clang-format doesn't complain)

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -566,7 +566,6 @@ raw2trace_t::process_offline_entry(raw2trace_thread_data_t *tdata,
                                    OUT bool *end_of_record, OUT bool *last_bb_handled,
                                    OUT bool *flush_decode_cache)
```
derekbruening, line 567: I think some tests are failing b/c of too-long >90 lines -- please fix to get the tests green.

derekbruening (approved):
LGTM with just a couple of nits.  Thank you for making the changes.

kuhanov:
Hi, @derekbruening. Could you add Yury (@yuriershov) to contributor list? He's going to share a few patches.
Thx, Kirill

kuhanov:
> The title and description are now inaccurate: they say "int" when it's actually "bool". Please remember to update when the code changes.

Oh, ok.

PR #6292, opened 2023-09-01
i#6288: Handle func markers delayed by signal arriving after a  branch

### LINKED ISSUE #6288: Invariant checker should handle signal arriving between the call/jmp and the function entry
Invariant checker doesn't handle the case when a signal arrives between call/jmp and the function entry when it checks for function marker. For example:

172899745: T408787 ifetch       2 byte(s) @ 0x00007fc608bc71c7 0f 05                syscall  -> %rcx %r11
172899746: T408787 <marker: timestamp 13310991841737435>
172899747: T408787 <marker: tid 408787 on core 23>
172899748: T408787 <marker: syscall xfer from 0x7fc608bc71c9>
172899749: T408787 <marker: timestamp 13310991841737533>
172899750: T408787 <marker: tid 408787 on core 23>
172899751: T408787 <marker: function #12>
172899752: T408787 <marker: function return value 0x7fc486120e70>

Function markers at ref 172899751 and 172899752 will be flagged as "Function marker should be after a branch" errors

ivankyluk:
When a signal arrives after a branch to a function, function markers are inserted after the corresponding syscall xfer marker.

There are a few corner cases which need to be handled:
1. Function markers in the beginning of a trace before any instructions.
2. syscall xfer markers without the corresponding kernel xfer markers. This case happens when a trace is started in the middle of a signal handler, or nested signals.

In order to avoid false positives, the check for function markers after a branch assumes markers are place correctly in these cases.

Fixes #6288


clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -686,6 +696,207 @@ check_function_markers()
...
+
+        std::vector<memref_t> memrefs = {
+            gen_instr(TID, 1),
+            gen_instr_type(TRACE_TYPE_INSTR_DIRECT_CALL, TID, CALL_PC, CALL_SZ),
+            gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_EVENT, CALL_PC + CALL_SZ),
+
+            gen_instr(TID, SIG_HANDLER_PC),
+            // Return from the signal handler.
+            gen_instr_type(TRACE_TYPE_INSTR_RETURN, TID, SIG_HANDLER_PC + 1, RETURN_SZ),
+            // sigreturn.
```
derekbruening, line 715: nit: Inconsistent capitalization; plus style guide requests capitalized comments: "Sigreturn." and below "Syscall xfer." and repeated again multiple times.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -686,6 +696,207 @@ check_function_markers()
...
+    {
+        constexpr addr_t SIG_HANDLER_PC = 400;
+        constexpr addr_t SYSCALL_PC = 500;
+        constexpr size_t SYSCALL_SZ = 2;
+        constexpr size_t RETURN_SZ = 3;
+
+        std::vector<memref_t> memrefs = {
+            gen_instr(TID, 1),
+            gen_instr_type(TRACE_TYPE_INSTR_DIRECT_CALL, TID, CALL_PC, CALL_SZ),
+            gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_EVENT, CALL_PC + CALL_SZ),
```
derekbruening, line 710: The signal arrived at the PC right after the call?  Meaning the call targeted the very next instr?  That makes it confusing: normally that is only done on x86 when trying to get the retaddr, not when actually making a call, as for an actual call you can't then have a normal return from the callee.  I would request changing this to a realistic callee PC.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -686,6 +696,207 @@ check_function_markers()
...
+        constexpr addr_t SIG1_PC = 200;
+        constexpr addr_t SIG2_PC = 300;
+        constexpr size_t INSTR_SZ = 1;
+        constexpr size_t RETURN_SZ = 3;
+        constexpr size_t SYSCALL_SZ = 2;
+
+        std::vector<memref_t> memrefs = {
+            gen_instr_type(TRACE_TYPE_INSTR_DIRECT_CALL, TID, BASE_PC, CALL_SZ),
+            // First signal.
+            gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_EVENT, BASE_PC + CALL_SZ),
```
derekbruening, line 836: Same comment: please make the call target==interruption PC something realistic.

Ditto for other cases below.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -686,6 +696,207 @@ check_function_markers()
...
+            // Second signal. No intervening instr here. Should skip pre-signal instr
```
derekbruening, line 837: It shouldn't skip it: it should use last_instr_in_cur_context_.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -686,6 +696,207 @@ check_function_markers()
...
+            // check on return.
+            gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_EVENT, SIG1_PC),
+            gen_instr(TID, SIG2_PC, INSTR_SZ),
+            gen_instr_type(TRACE_TYPE_INSTR_RETURN, TID, SIG2_PC + INSTR_SZ, RETURN_SZ),
+            // sigreturn of the second signal.
+            gen_instr(TID, SIG2_PC + INSTR_SZ + RETURN_SZ, SYSCALL_SZ),
+            gen_marker(TID, TRACE_MARKER_TYPE_SYSCALL, 15),
+            // XXX: This marker value is actually not guaranteed, yet the checker
+            // requires it and the view tool prints it.
```
derekbruening, line 846: What do you mean "not guaranteed"?

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -686,6 +696,207 @@ check_function_markers()
...
+            gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_XFER,
+                       SIG2_PC + INSTR_SZ + RETURN_SZ + SYSCALL_SZ),
+            gen_instr(TID, SIG1_PC, INSTR_SZ),
+
```
derekbruening, line 850: Seems like the blank line would be better served separating the inner and outer handlers?  Not sure why there would be a separate line in the middle of the outer handler as you have here.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -686,6 +696,207 @@ check_function_markers()
...
+            gen_instr(TID, SIG1_PC, INSTR_SZ),
+            gen_instr_type(TRACE_TYPE_INSTR_RETURN, TID, SIG1_PC + INSTR_SZ, RETURN_SZ),
+            // sigreturn.
+            gen_instr(TID, SIG1_PC + INSTR_SZ + RETURN_SZ, SYSCALL_SZ),
+            gen_marker(TID, TRACE_MARKER_TYPE_SYSCALL, 15),
+            gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_XFER,
+                       SIG1_PC + INSTR_SZ + RETURN_SZ + SYSCALL_SZ),
+
+            // Second signal with no intervening instr.
+            gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_EVENT, SIG1_PC + INSTR_SZ),
```
derekbruening, line 886: This is the wrong interruption PC: it should have the same interruption PC as the 1st signal.

This should fail the invariant checks -- or is that what the longstanding PR #5965 would check but it's not merged yet @lihasgupta ?

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -395,28 +395,39 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
         }
+
+        // Function markers may appear in the beginning of the trace before any
+        // instructions are recorded, i.e. shard->instr_count_ == 0. In other to avoid
+        // false positives in this case, we assume the markers are placed correctly.
+        report_if_false(
+            shard,
+            shard->prev_func_id_ >=
+                    static_cast<uintptr_t>(func_trace_t::TRACE_FUNC_ID_SYSCALL_BASE) ||
+                type_is_instr_branch(shard->prev_instr_.instr.type) ||
```
derekbruening, line 421: This should use last_instr_in_cur_context_ to grab the right pre-signal instr across nested signals -- xref #5965 which was going to change prev_instr to always be the right thing: @lihasgupta right?

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -688,6 +699,26 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
             if (shard->prev_entry_.marker.type != TRACE_TYPE_MARKER ||
                 shard->prev_entry_.marker.marker_type != TRACE_MARKER_TYPE_RSEQ_ABORT) {
                 shard->retaddr_stack_.push(0);
+                // When a signal arrives after a branch, function markers will appear in
+                // the trace when the corresponding syscall xfer happens. When a
+                // kernel xfer appears in the beginning of a trace before any
+                // instructions are recorded, we assume the signal arrives after a
+                // branch to avoid false positives.
+                shard->delayed_function_marker_stack_.push(
```
derekbruening, line 707: I don't think we need the concept of "delayed" at all: we can use last_instr_in_cur_context to always have the right pre-signal instr, and if that's clear then we'll know the trace started mid-signal.

(If we did want to add another stack that tracks signals: best to add a new field to signal_stack_ instead of making a separate stack, I would think.)

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -395,28 +395,39 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+
+        // Function markers may appear in the beginning of the trace before any
+        // instructions are recorded, i.e. shard->instr_count_ == 0. In other to avoid
+        // false positives in this case, we assume the markers are placed correctly.
+        report_if_false(
+            shard,
+            shard->prev_func_id_ >=
+                    static_cast<uintptr_t>(func_trace_t::TRACE_FUNC_ID_SYSCALL_BASE) ||
+                type_is_instr_branch(shard->prev_instr_.instr.type) ||
+                shard->instr_count_ == 0 || shard->delayed_function_marker_expected_,
```
derekbruening, line 422: I think both of these conditions can be replaced with checking whether last_instr_in_cur_context is empty or not.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: I've updated the callee PC here and below.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: I copied it from check_kernel_xfer(). I've removed the comment here, and in check_kernel_xfer().

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: I've fixed the interruption PC.

clients/drcachesim/tools/invariant_checker.cpp
ivankyluk: I've updated the code to use last_instr_in_cur_context.

clients/drcachesim/tools/invariant_checker.cpp
ivankyluk: I've changed the code to use last_instr_in_cur_context.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: I've removed the blank lines in the new tests.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: I've changed the code to use last_instr_in_cur_context_, and updated the comment here.

derekbruening (approved):

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -357,12 +357,8 @@ check_kernel_xfer()
     // Return to recorded interruption point.
```
derekbruening, line 357: The PR description is missing a `Fixes #6288` or `Issue: #6288` line (which one depends on whether it should close the issue).

derekbruening (approved):

PR #6296, opened 2023-09-06
AArch64: Remove unused function get_fcache_return_tls_offs

ksco:
The internal API `get_fcache_return_tls_offs()` was never used in Aarch64 and this patch removes it. Also changed the function signature in ARM to be static.

### DIFF
--- a/core/arch/arch.h
+++ b/core/arch/arch.h
@@ -1576,9 +1576,6 @@ get_direct_exit_target(dcontext_t *dcontext, uint flags);
 
 #ifdef AARCHXX
 size_t
-get_fcache_return_tls_offs(dcontext_t *dcontext, uint flags);
-
-size_t
 get_ibl_entry_tls_offs(dcontext_t *dcontext, cache_pc ibl_entry);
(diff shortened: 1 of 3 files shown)


core/arch/arch.h
```diff
@@ -1575,9 +1575,6 @@ cache_pc
 get_direct_exit_target(dcontext_t *dcontext, uint flags);
 
 #ifdef AARCHXX
-size_t
-get_fcache_return_tls_offs(dcontext_t *dcontext, uint flags);
-
```
abhinav92003, line 1580: Since this is ARM-local, it's probably fine to clean it from  AArch64. But I'm curious what motivated this change: did this show up as some sort of error in RISC-V?

abhinav92003 (approved):

core/arch/arch.h
ksco: With this change, I can just change the line above to `#if defined(AARCHXX) || defined(RISCV64)`, and no need to implement an unused function.

PR #6302, opened 2023-09-07
i#5843 scheduler: Add unread_last_record()

derekbruening:
Adds a new scheduler output stream interface unread_last_record(). This queues the just-read record for reading later.  This is meant to aid simulators reading one record ahead to the next instruction, but not yet processing it.  The new interface is limited in that it can only be called once at a time and while the output ordinals are fixed up other metrics like the output last timestamp and the input ordinals are not.

Adds a new record type TRACE_TYPE_INVALID for use as a sentinel.

Adds some sanity tests.

Issue: #5843

### DIFF
--- a/clients/drcachesim/scheduler/scheduler.h
+++ b/clients/drcachesim/scheduler/scheduler.h
@@ -549,6 +549,18 @@ public:
         next_record(RecordType &record, uint64_t cur_time);
 
         /**
+         * Queues the last-read record returned by next_record() such that it will be
+         * returned on the subsequent call to next_record() when this same input is
+         * active.  May not be called multiple times in a row without an intervening
+         * next_record() call.  This will cause ordinal queries on the current input to be
+         * off by one until the record is re-read.  Furthermore, the get_last_timestamp()
+         * query may still include this record, whether called on the input or output
+         * stream, immediately after this call.
+         */
+        virtual stream_status_t
+        unread_last_record();
+
+        /**
          * Begins a diversion from the regular inputs to a side stream of records
          * representing speculative execution starting at 'start_address'.
          *
@@ -556,7 +568,9 @@ public:
          * before knowing whether a simulator is on the wrong path or not, this routine
          * supports putting back the current record so that it will be re-provided as
          * the first record after (the outermost) stop_speculation(), if
-         * "queue_current_record" is true.
+         * "queue_current_record" is true.  An alternative way to accomplish this
+         * is to call unread_last_record().  The same caveats on the input stream
+         * ordinals and last timestamp apply to the queued record here.
          *
          * This call can be nested; each call needs to be paired with a corresponding
          * stop_speculation() call.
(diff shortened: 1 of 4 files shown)


derekbruening:
Hit a presumed-flaky failure in x86-64 off.invar test I don't think we've filed before but it seems familiar: filed as #6303; re-running.

clients/drcachesim/scheduler/scheduler.h
```diff
@@ -556,7 +568,9 @@ template <typename RecordType, typename ReaderType> class scheduler_tmpl_t {
          * before knowing whether a simulator is on the wrong path or not, this routine
          * supports putting back the current record so that it will be re-provided as
          * the first record after (the outermost) stop_speculation(), if
-         * "queue_current_record" is true.
+         * "queue_current_record" is true.  An alternative way to accomplish this
```
brettcoon, line 571: My reading of the comment suggests that start_speculation w/ queue_current_record=true supports arbitrary nesting, while unread_last_record() does not.  If so, that seems like a strong argument against using unread_last_record() as an alternative here.  Unless start_speculation() will _also_ nest multiple unread last records.
For that matter, it's a little unclear what the interaction is between unread_last_record() and start_speculation().  What happens if you call unread_last_record(), start_speculation(), and then next_record()?  Do you get the unread record or the side stream?

clients/drcachesim/scheduler/scheduler.h
```diff
@@ -1033,6 +1048,10 @@ template <typename RecordType, typename ReaderType> class scheduler_tmpl_t {
     next_record(output_ordinal_t output, RecordType &record, input_info_t *&input,
                 uint64_t cur_time = 0);
 
+    // Undoes the last read.  May only be called once between next_record() calls.
```
brettcoon, line 1051: As mentioned above, please clarify how this interacts with speculation, including speculation that uses the queue_next_record flag.

brettcoon (approved):
Mostly clarification requests.

clients/drcachesim/scheduler/scheduler.h
brettcoon: Probably worth adding a test for this interaction as well.

clients/drcachesim/scheduler/scheduler.h
```diff
@@ -548,6 +548,18 @@ template <typename RecordType, typename ReaderType> class scheduler_tmpl_t {
         virtual stream_status_t
         next_record(RecordType &record, uint64_t cur_time);
 
+        /**
+         * Queues the last-read record returned by next_record() such that it will be
+         * returned on the subsequent call to next_record() when this same input is
+         * active.  May not be called multiple times in a row without an intervening
+         * next_record() call.  This will cause ordinal queries on the current input to be
```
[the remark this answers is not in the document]
derekbruening: Done, and done.

clients/drcachesim/scheduler/scheduler.h
derekbruening: On further examination, you are right: the interactions do get messy.  I think the best thing is to disallowed mixing them: have unread_last_record() fail during speculation and have start_speculation(queue_current_record=true) fail if unread_last_record() was just called.  I documented this and added sanity checks for each failure case.

clients/drcachesim/scheduler/scheduler.h
derekbruening: Done (disallowed to mix), and done on adding a test of each failure case.

PR #6307, opened 2023-09-08
i#3544 RV64: Refine instr_branch_type

ksco:
Implements `instr_branch_type()`, which is called during the execution of a simple program. Since compressed instructions have all the implicit operands listed in instr_t, they don't need any special treatment.


abhinav92003 (approved):

derekbruening:
Don't forget an `Issue: #3544` line so these get linked to that issue.  I added it here and merged.

PR #6308, opened 2023-09-11
i#6238: Add categorization of x86 instruction mix and subcategories for FP category.

kuhanov:
This PR extends #6237 by adding categorization for x86 instruction mix. It introduces subcategories like MATH, CONVERT, and MOVE for both x86 and AArch64. For instance, arithmetic floating-point operations will have DR_INSTR_CATEGORY_FP | DR_INSTR_CATEGORY_MATH category.

Issue: #6238

Not all x86 opcodes are categorized nd such instructions are marked 'UNCATEGORIZED"


kuhanov:
code_api|tool.drcachesim.scattergather-x86 test is failed without these changes.
I've tried master and the last cronbuild-10.0.19608
318: ASSERT FAILURE: /home/huawei/DRIO_TESTING/11/master/clients/drcachesim/tracer/output.cpp:505: towrite <= ipc_pipe.get_atomic_write_size() && towrite > 0 ()

towrite size is 4128 and ipc_pipe.get_atomic_write_size() is 4096 on the fail check
Thx, Kirill

kuhanov:
code_api|tool.drcachesim.scattergather-x86 test is failed without these changes.
I've tried master and the last cronbuild-10.0.19608
318: ASSERT FAILURE: ~/master/clients/drcachesim/tracer/output.cpp:505: towrite <= ipc_pipe.get_atomic_write_size() && towrite > 0 ()

towrite size is 4128 and ipc_pipe.get_atomic_write_size() is 4096 on the fail check
Thx, Kirill

kuhanov:
Fail was started after #6290. Looks like you know about issue :)
"x86-64 failure is scattergather-x86 pipe assert #5329"

@derekbruening , could you review categorization for x86.
Thx, Kirill

> code_api|tool.drcachesim.scattergather-x86 test is failed without these changes. I've tried master and the last cronbuild-10.0.19608 318: ASSERT FAILURE: ~/master/clients/drcachesim/tracer/output.cpp:505: towrite <= ipc_pipe.get_atomic_write_size() && towrite > 0 ()
> 
> towrite size is 4128 and ipc_pipe.get_atomic_write_size() is 4096 on the fail check Thx, Kirill

Fail was started after #6290. Looks like you know about issue :)
"x86-64 failure is scattergather-x86 pipe assert #5329"

@derekbruening , could you review categorization for x86.
Thx, Kirill

derekbruening:
Was traveling.  Will take a look.

core/ir/instr_api.h
```diff
@@ -1884,13 +1884,17 @@ instr_is_rep_string_op(instr_t *instr);
...
-    DR_INSTR_CATEGORY_FP_MATH = 0x2,       /**< Floating-Point arithmetic operations. */
+    DR_INSTR_CATEGORY_INT = 0x1,           /**< Integer arithmetic operations. */
+    DR_INSTR_CATEGORY_FP = 0x2,            /**< Floating-Point operations. */
     DR_INSTR_CATEGORY_LOAD = 0x4,          /**< Loads. */
     DR_INSTR_CATEGORY_STORE = 0x8,         /**< Stores. */
     DR_INSTR_CATEGORY_BRANCH = 0x10,       /**< Branches. */
-    DR_INSTR_CATEGORY_SIMD = 0x20, /**< Operations with vector registers (SIMD). */
-    DR_INSTR_CATEGORY_OTHER = 0x40 /**< Other types of instructions. */
+    DR_INSTR_CATEGORY_SIMD = 0x20,     /**< Operations with vector registers (SIMD). */
+    DR_INSTR_CATEGORY_STATE = 0x40,    /**< Loads, stores. */
```
derekbruening, line 1907: The DR_FP_STATE is about the status flags and saving/restoring the entire FP state (e.g., XSAVE): it's not for regular loads and stores.

derekbruening (commented):
Sorry, I don't have much time this week: only made it through the first 4 files so far but had some comments so sending them out.

core/ir/x86/decode_private.h
```diff
@@ -561,6 +561,14 @@ extern const instr_info_t xop_extensions[];
 extern const instr_info_t evex_prefix_extensions[][2];
 extern const instr_info_t evex_Wb_extensions[][4];
 
+/* A table that associates opcodes with corresponding categories.
+ * When a new opcode is added, it will be automatically assigned the default
+ * category of "DR_INSTR_CATEGORY_UNCATEGORIZED".
+ * It is important to update the table by assigning appropriate categories
+ * to these new opcodes.
+ */
+extern const uint category_by_opcode[OP_AFTER_LAST];
```
derekbruening, line 570: The question is, should these instead be added as a new field in instr_info_t.  To decide: 
+ Do they vary by encoding template? If that happens a bunch it may be better in instr_info_t instead of lots of callouts to custom code to deduce.
+ Is it easier to maintain when the category is in the encoding template instead of in a separate array we need to update?  OTOH if many opcodes have multiple  encoding entries and they all have identical categories maybe it is easier with a separate out-of-line table.

core/ir/x86/decode_table.c
```diff
@@ -10703,4 +10703,1547 @@ const instr_info_t extra_operands[] =
...
 };
 
+const uint category_by_opcode[] = {
+    /* OP_INVALID */           DR_INSTR_CATEGORY_OTHER,
+    /* OP_UNDECODED */         DR_INSTR_CATEGORY_OTHER,
+    /* OP_CONTD */             DR_INSTR_CATEGORY_OTHER,
+    /* OP_LABEL */             DR_INSTR_CATEGORY_OTHER,
+
+    /* OP_add */               DR_INSTR_CATEGORY_INT,
+    /* OP_or */                DR_INSTR_CATEGORY_OTHER,
```
derekbruening, line 10713: A bunch of these are not filled in.  That's fine, but it should be clear that this is deliberate, and that this feature is incomplete: the PR description should state that a bunch of opcodes are not yet categorized; the doc comment for the API function should also say something until these are filled in; a TODO comment should go somewhere right here so a read doesn't get really confused thinking you deliberately thought that OR wasn't an integer math operation,

core/ir/aarch64/codec.c
```diff
@@ -9589,8 +9589,6 @@ decode_load_store_category(uint enc)
     } else if ((op0 & 0x3) == 0 || (op0 & 0x3) == 0x2) { /* xx00, xx10 */
```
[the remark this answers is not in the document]
kuhanov: thx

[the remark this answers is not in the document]
kuhanov: thx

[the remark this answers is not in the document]
kuhanov: thx.
Fixed.

core/ir/instr_api.h
```diff
@@ -1884,13 +1884,17 @@ instr_is_rep_string_op(instr_t *instr);
  */
 typedef enum {
     DR_INSTR_CATEGORY_UNCATEGORIZED = 0x0, /**< Uncategorized. */
-    DR_INSTR_CATEGORY_INT_MATH = 0x1,      /**< Integer arithmetic operations. */
-    DR_INSTR_CATEGORY_FP_MATH = 0x2,       /**< Floating-Point arithmetic operations. */
+    DR_INSTR_CATEGORY_INT = 0x1,           /**< Integer arithmetic operations. */
```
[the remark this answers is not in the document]
kuhanov: All ints are math operations. ok, I mark them as DR_INSTR_CATEGORY_INT | DR_INSTR_CATEGORY_MATH

core/ir/decode_api.h
```diff
@@ -143,6 +143,18 @@ DR_API
 const char *
 decode_opcode_name(int opcode);
 
+#ifdef X86
+DR_API
+/** Decodes the instruction category by opcode (OP_ constant). */
+bool
+decode_category(int opcode, instr_t *instr);
```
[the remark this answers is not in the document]
kuhanov: ok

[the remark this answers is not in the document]
kuhanov: The export has been removed. The primary objective was to utilize the decode_category function in the test in order to compare the sizes of the opcode and categories tables. I am wondering if there might be an alternative approach available to achieve this comparison.

core/ir/x86/decode.c
```diff
@@ -2541,6 +2549,7 @@ decode_common(dcontext_t *dcontext, byte *pc, byte *orig_pc, instr_t *instr)
                                         decode operands too */
                                _IF_DEBUG(!TEST(INSTR_IGNORE_INVALID, instr->flags)));
     instr_set_opcode(instr, info->type);
+    decode_category(info->type, instr);
```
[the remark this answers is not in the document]
kuhanov: It could not be fail. remove return value ay all

core/ir/aarch64/instr.c
```diff
@@ -312,17 +312,30 @@ instr_is_rep_string_op(instr_t *instr)
...
         if (type != NULL)
             *type = DR_FP_MATH;
         return true;
+    } else if (TEST(DR_INSTR_CATEGORY_CONVERT, cat)) {
+        if (type != NULL)
+            *type = DR_FP_CONVERT;
+        return true;
+    } else if (TEST(DR_INSTR_CATEGORY_LOAD, cat) || TEST(DR_INSTR_CATEGORY_STORE, cat)) {
+        if (type != NULL)
+            *type = DR_FP_STATE;
```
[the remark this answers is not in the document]
kuhanov: Let me move to TODO for STATES

core/ir/instr_api.h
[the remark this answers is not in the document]
kuhanov: I agree. This is minor changes

core/ir/decode_api.h
```diff
@@ -143,6 +143,18 @@ DR_API
 const char *
 decode_opcode_name(int opcode);
 
+#ifdef X86
+DR_API
+/** Decodes the instruction category by opcode (OP_ constant). */
```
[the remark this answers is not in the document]
kuhanov: remove from exports and don't return anything

core/ir/instr_api.h
kuhanov: mark definition as TODO right now. 
Need to split STATES to separate patch: set a correct definition; define the list of instructions and so on

core/ir/decode_api.h
```diff
@@ -143,6 +143,18 @@ DR_API
...
+bool
+decode_category(int opcode, instr_t *instr);
+#elif defined(AARCH64)
+DR_API
+/**  Decodes the instruction category by raw bits. */
+bool
+decode_category(uint enc, instr_t *instr);
```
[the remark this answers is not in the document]
kuhanov: Not clear here. We get category from raw bits. For me this is decoding. Am I wrong?

core/ir/x86/decode.c
```diff
@@ -2410,6 +2410,14 @@ decode_get_tuple_type_input_size(const instr_info_t *info, decode_info_t *di)
         di->input_size = OPSZ_NA;
 }
 
+bool
+decode_category(int opcode, instr_t *instr)
+{
+    int cat = category_by_opcode[opcode];
```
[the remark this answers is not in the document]
kuhanov: ok. Right now we could add OP_mov_ld in both (load and move)

core/ir/x86/decode_private.h
kuhanov: Sounds much better with field in  instr_info_t.
Added category member for x86 only. could be issue with others architecture here?
Thx, Kirill

core/ir/x86/decode_table.c
kuhanov: add TODO to decode_category body

kuhanov:
please, review categories in instr_info_t.
I didn't add doc comments yet.
Thx, Kirill

core/ir/x86/decode.c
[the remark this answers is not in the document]
derekbruening: Actually I believe a whole bunch of opcodes are in the same boat: all the TYPE_E operands can be a register or base-disp memory.  So that includes all the regular arithmetic (ADD, SUB, etc.).  Maybe mark them all as the INT or MATH type and then have the code always check for that type (or all types?) whether it reads or writes memory and if so it adds the LOAD or STORE category or whatever the name was for that?

derekbruening:
(Out of office and traveling next few days too)

core/ir/aarch64/instr.c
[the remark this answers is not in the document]
kuhanov: Hi @AssadHashmi 
Could you help me for this? Does AArch64 have anythin like 'XSAVE' instructions? or do it save processor state in the different way? like Interrupt handlers doc? 
https://developer.arm.com/documentation/dui0203/j/handling-processor-exceptions/armv6-and-earlier--armv7-a-and-armv7-r-profiles/interrupt-handlers
Thx, Kirill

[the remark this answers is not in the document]
AssadHashmi: > Hi @AssadHashmi
> Could you help me for this? Does AArch64 have anythin like 'XSAVE' instructions? or do it save processor state in the different way? like Interrupt handlers doc?
> https://developer.arm.com/documentation/dui0203/j/handling-processor-exceptions/armv6-and-earlier--armv7-a-and-armv7-r-profiles/interrupt-handlers
> Thx, Kirill

I don't recall an AArch64 instruction equivalent to X86-64's `XSAVE`. Processor state is saved/restored with loads and stores.

core/ir/decode_api.h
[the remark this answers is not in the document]
derekbruening: I mean the parameter name: `s/enc/encoding/`

derekbruening:
> @[kuhanov](https://github.com/kuhanov) kuhanov [force-pushed](https://github.com/DynamoRIO/dynamorio/compare/7ba6b51a8ab6fdb4cd92b4d0fafa50033c3c5d65..1029ba5a7356f40d557a142445abf0d904f25be3) the x86_instruction_categorization branch from [7ba6b51 ](https://github.com/DynamoRIO/dynamorio/commit/7ba6b51a8ab6fdb4cd92b4d0fafa50033c3c5d65)to [1029ba5 ](https://github.com/DynamoRIO/dynamorio/commit/1029ba5a7356f40d557a142445abf0d904f25be3)
[4 days ago](https://github.com/DynamoRIO/dynamorio/pull/6308#event-10380933830)

Please avoid force-pushing as it messes up all the comment histories and breaks the view of just recent changes so now the reviewer can't see what changed and has to go through everything all over again: it makes the reviewer's job harder (potentially much harder), with no benefit since the squash-and-merge cleans it all up at the end.  (This is documented in multiple places on https://dynamorio.org/page_code_reviews.html)

core/ir/aarch64/instr.c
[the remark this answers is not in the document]
kuhanov: Thx.
So STATE subcategory is not applicable for AArch64

kuhanov:
> Please avoid force-pushing as it messes up all the comment histories and breaks the view of just recent changes so now the reviewer can't see what changed and has to go through everything all over again: it makes the reviewer's job harder (potentially much harder), with no benefit since the squash-and-merge cleans it all up at the end. (This is documented in multiple places on https://dynamorio.org/page_code_reviews.html)

ok, sorry about that

> > @[kuhanov](https://github.com/kuhanov) kuhanov [force-pushed](https://github.com/DynamoRIO/dynamorio/compare/7ba6b51a8ab6fdb4cd92b4d0fafa50033c3c5d65..1029ba5a7356f40d557a142445abf0d904f25be3) the x86_instruction_categorization branch from [7ba6b51 ](https://github.com/DynamoRIO/dynamorio/commit/7ba6b51a8ab6fdb4cd92b4d0fafa50033c3c5d65)to [1029ba5 ](https://github.com/DynamoRIO/dynamorio/commit/1029ba5a7356f40d557a142445abf0d904f25be3)
> > [4 days ago](https://github.com/DynamoRIO/dynamorio/pull/6308#event-10380933830)
> 
> Please avoid force-pushing as it messes up all the comment histories and breaks the view of just recent changes so now the reviewer can't see what changed and has to go through everything all over again: it makes the reviewer's job harder (potentially much harder), with no benefit since the squash-and-merge cleans it all up at the end. (This is documented in multiple places on https://dynamorio.org/page_code_reviews.html)

ok, sorry about that.
Kirill

core/ir/decode_api.h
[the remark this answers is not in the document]
kuhanov: I see

core/ir/x86/decode.c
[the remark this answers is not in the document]
kuhanov: As I understand, currently it is in instr_is_mov.

core/ir/decode_api.h
```diff
@@ -1,5 +1,5 @@
 /* **********************************************************
- * Copyright (c) 2010-2021 Google, Inc.  All rights reserved.
+ * Copyright (c) 2010-2023 Google, Inc.  All rights reserved.
```
derekbruening, line 2: Revert since no other change in this file.

core/ir/instr_api.h
```diff
@@ -1898,13 +1898,17 @@ instr_is_rep_string_op(instr_t *instr);
...
+    DR_INSTR_CATEGORY_FP = 0x2,            /**< Floating-Point operations. */
     DR_INSTR_CATEGORY_LOAD = 0x4,          /**< Loads. */
     DR_INSTR_CATEGORY_STORE = 0x8,         /**< Stores. */
     DR_INSTR_CATEGORY_BRANCH = 0x10,       /**< Branches. */
-    DR_INSTR_CATEGORY_SIMD = 0x20, /**< Operations with vector registers (SIMD). */
-    DR_INSTR_CATEGORY_OTHER = 0x40 /**< Other types of instructions. */
+    DR_INSTR_CATEGORY_SIMD = 0x20,     /**< Operations with vector registers (SIMD). */
+    DR_INSTR_CATEGORY_STATE = 0x40,    /**< Loads, stores, or queries processor state. */
```
derekbruening, line 1907: To avoid confusion, how about "Saves, restores, or queries processor state."  Then nobody will think it's a regular load or store.  (Make the same change to the FP category description too, please.)

core/ir/x86/decode.c
```diff
@@ -2410,6 +2436,22 @@ decode_get_tuple_type_input_size(const instr_info_t *info, decode_info_t *di)
...
+ */
+static inline void
+decode_category(instr_t *instr)
+{
+    if (instr) {
+        if (op_instr[instr->opcode])
+            instr_set_category(instr, op_instr[instr->opcode]->category);
+        else
+            /* nonvalid opcode */
+            instr_set_category(instr, DR_INSTR_CATEGORY_UNCATEGORIZED);
```
derekbruening, line 2451: style violation: use {} around multi-line (regardless of statement count) body

core/ir/x86/decode_private.h
```diff
@@ -1,5 +1,5 @@
 /* **********************************************************
- * Copyright (c) 2011-2022 Google, Inc.  All rights reserved.
+ * Copyright (c) 2011-2023 Google, Inc.  All rights reserved.
```
derekbruening, line 2: Only change in the file so should be reverted

core/ir/x86/opcode_api.h
```diff
@@ -1,5 +1,5 @@
 /* **********************************************************
- * Copyright (c) 2011-2021 Google, Inc.  All rights reserved.
+ * Copyright (c) 2011-2023 Google, Inc.  All rights reserved.
```
derekbruening, line 2: Revert since no other change

suite/tests/api/drdecode_x86.c
```diff
@@ -155,6 +155,57 @@ test_noalloc(void)
      */
 }
 
+static void
+test_categories(void)
+{
+    const uint raw[] = {
+        0x10100ff3, /* fp move avx; OP_vmovss; movss  xmm2,DWORD PTR [eax] */
```
derekbruening, line 162: I would suggest synthesizing these from DR IR and encoding them and then decoding from those encoded raw bits.  Then no human needs to deal with raw bits; we don't need these comments since we can read the IR fairly well; no worries about >4-byte encodings; easier to tweak opcodes or operands; easier to add new entries; etc.  Maybe the same thing for the aarch64 test.

suite/tests/api/drdecode_x86.c
```diff
@@ -155,6 +155,57 @@ test_noalloc(void)
...
+        DR_INSTR_CATEGORY_OTHER,
+        DR_INSTR_CATEGORY_FP | DR_INSTR_CATEGORY_CONVERT | DR_INSTR_CATEGORY_SIMD,
+        DR_INSTR_CATEGORY_FP | DR_INSTR_CATEGORY_MATH | DR_INSTR_CATEGORY_SIMD,
+    };
+
+    size_t instr_count = sizeof(raw) / sizeof(uint);
+    int length;
+    byte *pc = (byte *)raw;
+    for (size_t i = 0; i < instr_count; i++) {
+        length = 4 - i % 2;
```
derekbruening, line 193: This needs a comment here and up in the array as this looks fragile and someone adding some new entry above is easily going to break this.  Or see the suggestion above of encoding from IR where we would then programmatically get the length and avoid all this.

core/ir/x86/decode.c
```diff
@@ -2410,6 +2436,22 @@ decode_get_tuple_type_input_size(const instr_info_t *info, decode_info_t *di)
         di->input_size = OPSZ_NA;
 }
 
+/* TODO i#6238: Not all opcodes have been reviewed.
+ * In case an opcode has not been reviewed,
+ * the default category assigned to it is DR_INSTR_CATEGORY_OTHER.
```
derekbruening, line 2441: I think it would be best to have it UNCATEGORIZED so it's clear it's not yet been examined.  OTHER implies someone looked at it and decided it didn't fit into any category.

core/ir/x86/decode.c
```diff
@@ -2410,6 +2436,22 @@ decode_get_tuple_type_input_size(const instr_info_t *info, decode_info_t *di)
...
+ */
+static inline void
+decode_category(instr_t *instr)
+{
+    if (instr) {
+        if (op_instr[instr->opcode])
+            instr_set_category(instr, op_instr[instr->opcode]->category);
```
derekbruening, line 2448: But what about the other comment thread: we're going to be missing a lot of _LOAD and _STORE and we need to add them here for instr_reads_memory() or instr_writes_memory().  That seems simple enough to just add here (a couple of lines) instead of adding a TODO for later?

core/ir/x86/decode.c
```diff
@@ -2410,6 +2436,22 @@ decode_get_tuple_type_input_size(const instr_info_t *info, decode_info_t *di)
...
 
+/* TODO i#6238: Not all opcodes have been reviewed.
+ * In case an opcode has not been reviewed,
+ * the default category assigned to it is DR_INSTR_CATEGORY_OTHER.
+ */
+static inline void
+decode_category(instr_t *instr)
+{
+    if (instr) {
+        if (op_instr[instr->opcode])
```
derekbruening, line 2447: style: explicit bool: `!= NULL`

core/ir/x86/decode.c
```diff
@@ -2410,6 +2436,22 @@ decode_get_tuple_type_input_size(const instr_info_t *info, decode_info_t *di)
...
 }
 
+/* TODO i#6238: Not all opcodes have been reviewed.
+ * In case an opcode has not been reviewed,
+ * the default category assigned to it is DR_INSTR_CATEGORY_OTHER.
+ */
+static inline void
+decode_category(instr_t *instr)
+{
+    if (instr) {
```
derekbruening, line 2446: style: explicit bool: `!= NULL`

core/ir/x86/decode.c
```diff
@@ -2410,6 +2436,22 @@ decode_get_tuple_type_input_size(const instr_info_t *info, decode_info_t *di)
...
+        if (op_instr[instr->opcode])
+            instr_set_category(instr, op_instr[instr->opcode]->category);
+        else
+            /* nonvalid opcode */
+            instr_set_category(instr, DR_INSTR_CATEGORY_UNCATEGORIZED);
+    }
+}
+
```
derekbruening, line 2454: Wow, the decode_table.c diff is too big and github won't even let me expand it or comment on it!  So adding comments here:

>+    {OP_add,  0x010000, DR_INSTR_CATEGORY_INT | DR_INSTR_CATEGORY_MATH, "add",  Ev, xx, Gv, Ev, xx, mrm, fW6, tfb[0x00]},
>+    {OP_add,  0x020000, DR_INSTR_CATEGORY_INT | DR_INSTR_CATEGORY_MATH, "add",  Gb, xx, Eb, Gb, xx, mrm, fW6, tfb[0x01]},

I was going to say that moving into the decode entries was good because here we have the same opcode being either a load or a store: but I guess for this case we plan to add _LOAD or _STORE in decode_category() so maybe these could have been in an opcode array.  But I bet we find other cases where it's better in the entries.

Given that we want to be able to view the table lines without them being incredibly wide, I think it's worth sacrificing a little verbosity by using abbreviations for the long enum names.  Maybe `s/DR_INSTR_CATEGORY_INT | DR_INSTR_CATEGORY_MATH/catInt|catMath/`?

derekbruening, line 2454: I'm seeing a bunch of entries that I think have the wrong category.  Here are some:

>+    {OP_push, 0x060000, DR_INSTR_CATEGORY_OTHER, "push", xsp, i_xSPo1, es, xsp, xx, i64, x, tfb[0x0e]},

This should be a _STORE.  (Some OP_push variants can also be a _LOAD but we let the C code add that I guess since I think it's another type E.)

>+    {OP_mov_imm, 0xb00000, DR_INSTR_CATEGORY_OTHER, "mov", al_x, xx, Ib, xx, xx, no, x, tfb[0xb1]},

Shouldn't this be _MOVE?

>+    {OP_adc,  0x100000, DR_INSTR_CATEGORY_OTHER, "adc",  Eb, xx, Gb, Eb, xx, mrm, (fW6|fRC), tex[1][2]},

This is an addition with carry so it should be _INT|_MATH.

If you don't want to go update all of these, which is a bunch of work, maybe for this initial PR just replace all the _OTHER with _UNCATEGORIZED?  Then a later PR can fix up the uncats and it's clear what's missing.

derekbruening, line 2454: Please update the PR description to talk about what's not yet implemented: not all x86 opcodes are categorized.

derekbruening (commented):
Looking pretty good, but maybe worth one more glance after this set of suggestions before merging.

core/ir/instr_api.h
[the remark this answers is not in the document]
derekbruening: Revisiting this: but we probably aren't going to mark a GPR move as _INT|_MOVE (I only see _FP|_MOVE in this PR)?  So if we are only going to ever use _INT when combined with _MATH: do we need _INT?  Should we go with the absence of _FP and absence of _SIMD implying integer/GPR and thus plain _MATH implies integer?

core/ir/x86/decode.c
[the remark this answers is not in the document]
kuhanov: @derekbruening could we add all instructions from instr_is_mov to INT | MOVE right now?

core/ir/x86/decode.c
kuhanov: agree

core/ir/x86/decode.c
kuhanov: ok. replaced to _UNCATEGORIZED

kuhanov: @derekbruening 
Do you suggest to replace enum to the following?
typedef enum {
    catUncategorized = 0x0, /**< Uncategorized. */
    catInt = 0x1,           /**< Integer arithmetic operations. */
    catFP = 0x2,            /**< Floating-Point operations. */
    catLoad = 0x4,          /**< Loads. */
    catStore = 0x8,         /**< Stores. */
    catBranch = 0x10,       /**< Branches. */
    catSIMD = 0x20,  /**< Operations with vector registers (SIMD). */
    catState = 0x40, /**< Saves, restores, or queries processor state. */
    catMove = 0x80,  /**< Moves value from one location to another. */
    catConvert = 0x100, /**< Converts to or from value. */
    catMath = 0x200, /**< Performs arithmetic or conditional operations. */
    catOther = 0x400 /**< Other types of instructions. */
} dr_instr_category_t;

core/ir/instr_api.h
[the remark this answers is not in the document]
kuhanov: Agree. We could skip INT at all and mark them just _MATH without FP flag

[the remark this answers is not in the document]
kuhanov: remove INT

core/ir/x86/decode.c
derekbruening: No, I would keep the enum how it is.  I would define local macros just in decode_table.c to shorten those long lines (like the existing short names for the various other fields).

core/ir/x86/decode.c
```diff
@@ -2438,17 +2438,24 @@ decode_get_tuple_type_input_size(const instr_info_t *info, decode_info_t *di)
...
-    if (instr) {
-        if (op_instr[instr->opcode])
-            instr_set_category(instr, op_instr[instr->opcode]->category);
-        else
+    if (instr != NULL) {
+        if (op_instr[instr->opcode] != NULL) {
+            uint default_category = op_instr[instr->opcode]->category;
+            if (instr_operands_valid(instr) && instr_reads_memory(instr))
+                instr_set_category(instr, default_category | DR_INSTR_CATEGORY_LOAD);
+            else if (instr_operands_valid(instr) && instr_writes_memory(instr))
```
derekbruening, line 2451: Some are both loads and stores, so no `else`.

api/docs/release.dox
```diff
@@ -127,6 +127,8 @@ clients.
 The changes between version \DR_VERSION and 10.0.0 include the following compatibility
 changes:
  - Marked x86 rep instructions as predicated.
+ - The #dr_instr_category_t enum underwent changes to support subcategories
+   such as STATE, MOVE, CONVERT, and MATH.
```
derekbruening, line 131: nit: Best to say that INT was removed and all the numbers changed.

derekbruening: Make that INT_MATH that was removed.

derekbruening (approved):
Thank you for contributing this feature and working through the review process.

suite/tests/api/drdecode_x86.c
```diff
@@ -155,6 +155,36 @@ test_noalloc(void)
      */
 }
 
+#define CHECK_CATEGORY(dcontext, instr, pc, category) \
+    instr_encode(dcontext, instr, pc);                \
```
derekbruening, line 159: It would be good to check the return values of this call and decode() below.

PR #6310, opened 2023-09-13
i#3544 RV64: Implement instr_* functions

ksco:
This patch implements the necessary instr_* functions that are called when running simple programs. Also, the opnd type of `auipc` has been changed again to `REL_ADDR_kind`. This is essential for `opc_is_not_a_real_memory_load()` and `instr_has_rel_addr_reference()` to work correctly.

Issue: #3544

### DIFF
--- a/core/ir/riscv64/codec.c
+++ b/core/ir/riscv64/codec.c
@@ -486,7 +486,7 @@ decode_u_immpc_opnd(dcontext_t *dc, uint32_t inst, int op_sz, byte *pc, byte *or
                     int idx, instr_t *out)
 {
     int32_t uimm = GET_FIELD(inst, 31, 12);
-    opnd_t opnd = opnd_create_pc(orig_pc + (uimm << 12));
+    opnd_t opnd = opnd_create_rel_addr(orig_pc + (uimm << 12), OPSZ_0);
     instr_set_src(out, idx, opnd);
     return true;
 }
@@ -1937,8 +1937,8 @@ encode_u_immpc_opnd(instr_t *instr, byte *pc, int idx, uint32_t *out, decode_inf
 {
     opnd_t opnd = instr_get_src(instr, idx);
     uint32_t imm;
-    if (opnd.kind == PC_kind)
-        imm = opnd_get_pc(opnd) - pc;
+    if (opnd.kind == REL_ADDR_kind)
+        imm = (app_pc)opnd_get_addr(opnd) - pc;
     else if (opnd.kind == INSTR_kind)
         imm = (byte *)opnd_get_instr(opnd)->offset - (byte *)instr->offset;
     else
--- a/core/ir/riscv64/codec.c
+++ b/core/ir/riscv64/codec.c
@@ -486,6 +486,8 @@ decode_u_immpc_opnd(dcontext_t *dc, uint32_t inst, int op_sz, byte *pc, byte *or
                     int idx, instr_t *out)
 {
     int32_t uimm = GET_FIELD(inst, 31, 12);
+    /* OPSZ_0 is used here to indicate that this is not a real memory access instruction.
+     */
     opnd_t opnd = opnd_create_rel_addr(orig_pc + (uimm << 12), OPSZ_0);
     instr_set_src(out, idx, opnd);
     return true;
--- a/core/ir/riscv64/instr.c
+++ b/core/ir/riscv64/instr.c
@@ -48,16 +48,58 @@ instr_get_isa_mode(instr_t *instr)
 int
 instr_length_arch(dcontext_t *dcontext, instr_t *instr)
 {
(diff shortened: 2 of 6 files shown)


core/ir/riscv64/codec.c
```diff
@@ -486,7 +486,7 @@ decode_u_immpc_opnd(dcontext_t *dc, uint32_t inst, int op_sz, byte *pc, byte *or
                     int idx, instr_t *out)
 {
     int32_t uimm = GET_FIELD(inst, 31, 12);
-    opnd_t opnd = opnd_create_pc(orig_pc + (uimm << 12));
+    opnd_t opnd = opnd_create_rel_addr(orig_pc + (uimm << 12), OPSZ_0);
```
abhinav92003, line 489: OPSZ_0 because we we're not really getting any data from the addr, but only a pc? Maybe add a comment.

core/ir/riscv64/instr.c
```diff
@@ -120,46 +162,40 @@ bool
...
     int opc = instr_get_opcode(instr);
-    if (instr_num_srcs(instr) < 1 || instr_num_dsts(instr) < 1)
-        return false;
-    opnd_t rd = instr_get_dst(instr, 0);
-    opnd_t rs = instr_get_src(instr, 0);
-    return (opc == OP_jalr && opnd_get_reg(rd) == DR_REG_X0 &&
-            opnd_is_memory_reference(rs) && opnd_get_base(rs) == DR_REG_X1 &&
-            opnd_get_disp(rs) == 0);
+    return ((opc == OP_c_jr || opc == OP_jalr) &&
+            opnd_get_reg(instr_get_dst(instr, 0)) == DR_REG_X0 &&
```
abhinav92003, line 196: I assume these opcodes are guaranteed to have 2 srcs and 1 dst?

core/ir/riscv64/codec.c
abhinav92003, line 489: Please update the comment by REL_ADDR_kind in core/ir/opnd_api.h.

core/ir/riscv64/instr.c
```diff
@@ -120,46 +162,40 @@ bool
 instr_is_call_arch(instr_t *instr)
 {
     int opc = instr_get_opcode(instr);
-    return (opc == OP_jal || opc == OP_jalr || opc == OP_c_j || opc == OP_c_jal ||
-            opc == OP_c_jr || opc == OP_c_jalr);
+    return ((opc == OP_jal || opc == OP_jalr || opc == OP_c_jal || opc == OP_c_jalr) &&
+            opnd_get_reg(instr_get_dst(instr, 0)) != DR_REG_ZERO);
```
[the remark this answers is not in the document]
ksco: The condition is `!= DR_REG_ZERO`, for `c.jal / c.jalr`, that's always true. But since we're including all the implicit operands, I think it's nicer to treat them just as the expanded one. i.e. I think it's better to write `(opc == OP_jal || opc == OP_c_jal) && dst0 != ZERO` than `(opc == OP_jal && dst0 != ZERO) || opc == OP_c_jal`.

core/ir/riscv64/instr.c
ksco: Yes.

core/ir/riscv64/codec.c
ksco: Added a comment

ksco: Updated.

ksco:
I have another question. 
Now that RISC-V port can run C programs (on my fork, still upstreaming), I made a simple benchmark for finding prime numbers.
The performance on RISC-V side is a bit poor: `time drrun ./prime` takes 8s, while `time ./prime` only takes 2s. On Aarch64, however, the time on same program, using `drrun` or not, is indistinguishable. I used `perf top` to take a look, and the main time was indeed spent on JIT code on both platforms, so I feel something is off. How should I debug this? Are there any suggestions?

abhinav92003 (approved):

abhinav92003:
> the main time was indeed spent on JIT code on both platforms

By JIT code you mean some code cache generation code in DR? Do you have a list of hot functions/callstacks?

abhinav92003:
> > the main time was indeed spent on JIT code on both platforms
> 
> By JIT code you mean some code cache generation code in DR? Do you have a list of hot functions/callstacks?

I suppose you mean most time was spent in JIT code itself (and not the JIT generator like I assumed previously)? You could use `dr_set_itimer` to set a callback where you get `dr_where_am_i_t` samples using  `dr_where_am_i` and `mcontext->pc`.

ksco:
Yes, the generated code itself, and thanks for the advise!

abhinav92003:
That's the `tool.drcachesim.scattergather-x86` failure which is #5329. Looks like it is happening much more frequently now.

abhinav92003:
@ksco Should be fixed now. Please retry.

PR #6313, opened 2023-09-18
i#5329: Fix tool.drcachesim.scattergather-x86 test failure

abhinav92003:
Avoids going over the atomic write size while waiting for enough entries to write to IPC pipe. If we find we went over, we just write till the last ok-to-split ref.

tool.drcachesim.scattergather-x86 worked fine 100x on local workstation with this fix.

This issue manifests on this test because some scatter-gather instruction has eight consecutive memrefs which puts us over the threshold suddenly without any opportunity to write the buffer before all the memrefs.

Fixes: #5329

### DIFF
--- a/clients/drcachesim/tracer/output.cpp
+++ b/clients/drcachesim/tracer/output.cpp
@@ -754,9 +754,9 @@ output_buffer(void *drcontext, per_thread_t *data, byte *buf_base, byte *buf_ptr
                     // Check if we went over the edge waiting for enough entries to
                     // write. If we did, we simply write till the last ok-to-split ref.
                     if (mem_ref - pipe_start > ipc_pipe.get_atomic_write_size()) {
-                        // If the following assert triggers, we found too many entries
-                        // without an ok-to-split point.
-                        DR_ASSERT(last_ok_to_split_ref != nullptr);
+                        DR_ASSERT_MSG(
+                            last_ok_to_split_ref != nullptr,
+                            "Found too many entries without an ok-to-split point");
                         pipe_start =
                             atomic_pipe_write(drcontext, pipe_start, last_ok_to_split_ref,
                                               get_local_window(data));
@@ -781,9 +781,8 @@ output_buffer(void *drcontext, per_thread_t *data, byte *buf_base, byte *buf_ptr
             DR_ASSERT(
                 is_ok_to_split_before(instru->get_entry_type(pipe_start + header_size),
                                       instru->get_entry_size(pipe_start + header_size)));
-            // If the following assert triggers, we found too many entries without an
-            // ok-to-split point.
-            DR_ASSERT(last_ok_to_split_ref != nullptr);
+            DR_ASSERT_MSG(last_ok_to_split_ref != nullptr,
+                          "Found too many entries without an ok-to-split point");
             pipe_start = atomic_pipe_write(drcontext, pipe_start, last_ok_to_split_ref,
                                            get_local_window(data));
         }
--- a/clients/drcachesim/tracer/output.cpp
+++ b/clients/drcachesim/tracer/output.cpp
@@ -731,6 +731,7 @@ output_buffer(void *drcontext, per_thread_t *data, byte *buf_base, byte *buf_ptr
     byte *pipe_end = pipe_start;
     if (!op_offline.get_value()) {
         byte *post_header = buf_base + header_size;
+        byte *last_ok_to_split_ref = nullptr;
         // Pipe split headers are just the tid.
         header_size = instru->sizeof_entry();
         for (byte *mem_ref = post_header; mem_ref < buf_ptr;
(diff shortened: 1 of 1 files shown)


clients/drcachesim/tracer/output.cpp
```diff
@@ -728,9 +728,9 @@ output_buffer(void *drcontext, per_thread_t *data, byte *buf_base, byte *buf_ptr
               size_t header_size)
```
derekbruening, line 728: There is a test failure: is it a known one?  Does it repeat?

clients/drcachesim/tracer/output.cpp
```diff
@@ -752,8 +751,23 @@ output_buffer(void *drcontext, per_thread_t *data, byte *buf_base, byte *buf_ptr
                     DR_ASSERT(is_ok_to_split_before(
                         instru->get_entry_type(pipe_start + header_size),
                         instru->get_entry_size(pipe_start + header_size)));
-                    pipe_start = atomic_pipe_write(drcontext, pipe_start, pipe_end,
-                                                   get_local_window(data));
+                    // Check if we went over the edge waiting for enough entries to
+                    // write. If we did, we simply write till the last ok-to-split ref.
+                    if (mem_ref - pipe_start > ipc_pipe.get_atomic_write_size()) {
+                        // If the following assert triggers, we found too many entries
+                        // without an ok-to-split point.
```
derekbruening, line 758: Instead of this comment this can be added to the assert with DR_ASSERT_MSG (or `&& "string"` if that weren't available): "found too many entries with no split point" or something.  Then it will be visible to the dev/user if this fires.

clients/drcachesim/tracer/output.cpp
```diff
@@ -767,7 +781,10 @@ output_buffer(void *drcontext, per_thread_t *data, byte *buf_base, byte *buf_ptr
             DR_ASSERT(
                 is_ok_to_split_before(instru->get_entry_type(pipe_start + header_size),
                                       instru->get_entry_size(pipe_start + header_size)));
-            pipe_start = atomic_pipe_write(drcontext, pipe_start, pipe_end,
+            // If the following assert triggers, we found too many entries without an
+            // ok-to-split point.
```
derekbruening, line 785: Same comment: move this to the assert as a string so it is instantly visible to the user/dev and isn't just in the code.

clients/drcachesim/tracer/output.cpp
```diff
@@ -752,8 +751,23 @@ output_buffer(void *drcontext, per_thread_t *data, byte *buf_base, byte *buf_ptr
                     DR_ASSERT(is_ok_to_split_before(
                         instru->get_entry_type(pipe_start + header_size),
                         instru->get_entry_size(pipe_start + header_size)));
-                    pipe_start = atomic_pipe_write(drcontext, pipe_start, pipe_end,
-                                                   get_local_window(data));
+                    // Check if we went over the edge waiting for enough entries to
+                    // write. If we did, we simply write till the last ok-to-split ref.
```
derekbruening, line 755: Is there a reason this kept happening only with scatter/gather?  Do we understand why?  If so, can we use that knowledge to make a targeted test that always reproduces the issue?

derekbruening (approved):

clients/drcachesim/tracer/output.cpp
abhinav92003: I think it's because there's an underlying assumption here that we'll re-do the check at L749 sooner than every `(MAX_NUM_DELAY_ENTRIES +1)` (==5) entries, but with scatter-gather I'm seeing 8 intervening memrefs. So we go over the threshold by the time we get the opportunity to check whether we should output data already.

clients/drcachesim/tracer/output.cpp
abhinav92003, line 728: @derekbruening I'm curious about MAX_NUM_DELAY_ENTRIES and MAX_NUM_DELAY_INSTRS.

```
#define MAX_NUM_DELAY_INSTRS 32
// Really sizeof(trace_entry_t.length)
#define MAX_NUM_DELAY_ENTRIES (MAX_NUM_DELAY_INSTRS / sizeof(addr_t))
```

But doesn't each entry correspond to one instruction in the online mode?

derekbruening: There are instruction bundles which combine multiple instructions.

abhinav92003: Right. But why do we divide by the sizeof?

clients/drcachesim/tracer/output.cpp
abhinav92003: The existing scatter-gather test seems targeted enough? It's hitting this case consistently.

clients/drcachesim/tracer/output.cpp
abhinav92003: It was `tool.drcachesim.opcode_mix`

```
Encoding size 2 != instr size 6
drcachesim: /home/runner/work/dynamorio/dynamorio/clients/drcachesim/reader/reader.cpp:203: virtual bool dynamorio::drmemtrace::reader_t::process_input_entry(): Assertion `false' failed.
```

I don't think we've seen this before.

abhinav92003: It didn't repeat though when I reran the workflow.

abhinav92003: Created #6314

clients/drcachesim/tracer/output.cpp
abhinav92003: Done.

clients/drcachesim/tracer/output.cpp
abhinav92003: Done

clients/drcachesim/tracer/output.cpp
abhinav92003: Will follow up offline about what I'm missing.

clients/drcachesim/tracer/output.cpp
abhinav92003: Added more details to description

abhinav92003:
x86-64 `api.rseq` failure is #6313

PR #6315, opened 2023-09-20
i#6303: Avoid duplicate encoding entry on chunk-split branch pair

### LINKED ISSUE #6303: Tests failing with "Encoding size 2 != instr size 1": drcacheoff.invariant_checker, drcachesim.opcode_mix
Happened on x86-64 here:
https://github.com/DynamoRIO/dynamorio/actions/runs/6113660426/job/16593592056?pr=6302
```
390: CMake Error at /home/runner/work/dynamorio/dynamorio/suite/tests/runmulti.cmake:111 (message):
390:   *** postcmd failed (Subprocess aborted): Encoding size 2 != instr size 1
390: 
390:   drcachesim:
390:   /home/runner/work/dynamorio/dynamorio/clients/drcachesim/reader/reader.cpp:203:
390:   virtual bool dynamorio::drmemtrace::reader_t::process_input_entry():
390:   Assertion `false' failed.
390: 
390:   ***
390: 
390: Call Stack (most recent call first):
390:   /home/runner/work/dynamorio/dynamorio/suite/tests/runmulti.cmake:123 (process_cmdline)
390: 
390: 
389/471 Test #390: code_api|tool.drcacheoff.invariant_checker .......................***Failed    2.08 sec
```

derekbruening:
Fixes a case where a branch followed by an indirect branch with a chunk split in the middle results in an extra encoding entry after the indirect branch target marker.  This extra entry is accumulated in the reader and results in a fatal error.

Adds a unit test.  Confirmed the fix on the
drcacheoff.invariant_checker test as well where it was failing reliably on one machine.

Fixes #6303

### DIFF
--- a/clients/drcachesim/tests/raw2trace_unit_tests.cpp
+++ b/clients/drcachesim/tests/raw2trace_unit_tests.cpp
@@ -887,7 +887,7 @@ test_chunk_encodings(void *drcontext)
         check_entry(entries, idx, TRACE_TYPE_INSTR, -1) &&
         // Block 4.
         check_entry(entries, idx, TRACE_TYPE_ENCODING, -1) &&
-        check_entry(entries, idx, TRACE_TYPE_INSTR, -1) &&
+        check_entry(entries, idx, TRACE_TYPE_INSTR, -1, offs_nop_start) &&
         check_entry(entries, idx, TRACE_TYPE_ENCODING, -1) &&
         check_entry(entries, idx, TRACE_TYPE_INSTR, -1) &&
         check_entry(entries, idx, TRACE_TYPE_ENCODING, -1) &&
--- a/clients/drcachesim/tests/raw2trace_unit_tests.cpp
+++ b/clients/drcachesim/tests/raw2trace_unit_tests.cpp
@@ -771,26 +771,43 @@ test_chunk_encodings(void *drcontext)
     instr_t *nop = XINST_CREATE_nop(drcontext);
     // Test i#5724 where a chunk boundary between consecutive branches results
     // in a missing encoding entry.
+    // Also test i#6303 where a delayed indirect branch with tagalong encoding has its
+    // encoding repeated, causing a reader assert as it accumulates the duplicate.
     instr_t *move1 =
         XINST_CREATE_move(drcontext, opnd_create_reg(REG1), opnd_create_reg(REG2));
     instr_t *move2 =
         XINST_CREATE_move(drcontext, opnd_create_reg(REG1), opnd_create_reg(REG2));
-    instr_t *jmp2 = XINST_CREATE_jump(drcontext, opnd_create_instr(move2));
-    instr_t *jmp1 = XINST_CREATE_jump(drcontext, opnd_create_instr(jmp2));
+    instr_t *jmp_move2 = XINST_CREATE_jump(drcontext, opnd_create_instr(move2));
+    instr_t *jmp_jmp = XINST_CREATE_jump(drcontext, opnd_create_instr(jmp_move2));
+    instr_t *nop_start = XINST_CREATE_nop(drcontext);
+    instr_t *jcc =
+        XINST_CREATE_jump_cond(drcontext, DR_PRED_EQ, opnd_create_instr(move1));
+    instr_t *ret = XINST_CREATE_return(drcontext);
     instrlist_append(ilist, nop);
     // Block 1.
     instrlist_append(ilist, move1);
-    instrlist_append(ilist, jmp1);
+    instrlist_append(ilist, jmp_jmp);
     // Block 2.
-    instrlist_append(ilist, jmp2);
+    instrlist_append(ilist, jmp_move2);
     // Block 3.
     instrlist_append(ilist, move2);
+    // Block 4.
+    instrlist_append(ilist, nop_start);
+    instrlist_append(ilist, XINST_CREATE_nop(drcontext));
+    instrlist_append(ilist, XINST_CREATE_nop(drcontext));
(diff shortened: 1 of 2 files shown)


clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -3100,8 +3100,13 @@ raw2trace_t::write(raw2trace_thread_data_t *tdata, const trace_entry_t *start,
             }
```
abhinav92003, line 3100: For L3082: It'd be useful to add a comment above that reminds the reader that prev_was_encoding is to avoid a duplicate encoding being emitted after an actual encoding entry, specifically for the case *after we open a new chunk* (because record_encoding_emitted will still return `true` since we just cleared that cache after opening a new chunk)

(or possibly a better worded summary)

clients/drcachesim/tests/raw2trace_unit_tests.cpp
```diff
@@ -771,26 +771,43 @@ test_chunk_encodings(void *drcontext)
...
     instr_t *move1 =
         XINST_CREATE_move(drcontext, opnd_create_reg(REG1), opnd_create_reg(REG2));
     instr_t *move2 =
         XINST_CREATE_move(drcontext, opnd_create_reg(REG1), opnd_create_reg(REG2));
-    instr_t *jmp2 = XINST_CREATE_jump(drcontext, opnd_create_instr(move2));
-    instr_t *jmp1 = XINST_CREATE_jump(drcontext, opnd_create_instr(jmp2));
+    instr_t *jmp_move2 = XINST_CREATE_jump(drcontext, opnd_create_instr(move2));
+    instr_t *jmp_jmp = XINST_CREATE_jump(drcontext, opnd_create_instr(jmp_move2));
+    instr_t *nop_start = XINST_CREATE_nop(drcontext);
+    instr_t *jcc =
```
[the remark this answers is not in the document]
derekbruening: It doesn't matter where the jcc goes as it's not taken, but to make it consistent with the others here I renamed it.

clients/drcachesim/tests/raw2trace_unit_tests.cpp
```diff
@@ -801,11 +818,16 @@ test_chunk_encodings(void *drcontext)
     raw.push_back(make_timestamp());
     raw.push_back(make_core());
     raw.push_back(make_block(offs_move1, 2));
-    raw.push_back(make_block(offs_jmp2, 1));
+    raw.push_back(make_block(offs_jmp_move2, 1));
     raw.push_back(make_block(offs_move2, 1));
     // Repeat the jmp,jmp to test re-emitting encodings in new chunks.
```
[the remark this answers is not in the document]
derekbruening: No it is correct: the move1 block has 2 instrs, move1 and then jmp_jmp which targets jmp_move2.

clients/drcachesim/tracer/raw2trace.cpp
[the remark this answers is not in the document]
derekbruening: Done.

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -3100,8 +3100,13 @@ raw2trace_t::write(raw2trace_thread_data_t *tdata, const trace_entry_t *start,
...
             if (it->type == TRACE_TYPE_ENCODING)
                 prev_was_encoding = true;
-            else
-                prev_was_encoding = false;
+            else {
+                // Do not clear across an indirect branch target, to avoid
+                // a duplicate encoding on the other side.
```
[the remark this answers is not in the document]
derekbruening: Yes; reworded.

clients/drcachesim/tracer/raw2trace.cpp
derekbruening: Hmm: your questions are making me realize that record_encoding_emitted is never called for these delayed encodings (it was called pre-chunk but was cleared) -- it seems like what we want is to call that, and get rid of prev_was_encoding.  I just added a 2nd ret to the unit test and indeed we (incorrectly) emit another encoding there!

abhinav92003: > your questions are making me realize that record_encoding_emitted is never called for these delayed encodings (it was called pre-chunk but was cleared) 

Wait why?

By delayed encoding, you mean an encoding that we deliberately repeat after opening a new chunk (L3091), right? But control would enter that if-block only after calling record_encoding_emitted.

derekbruening: That if-block is only for instructions and it short-circuits if prev_was_encoding.  But we don't have the hash key: so I think we have to wait for the instr and if prev_was_encoding add to the hash there.

abhinav92003: > That if-block is only for instructions

Right, but that's fine because we want to emit the delayed encoding only when we see an actual instruction for it.

> it short-circuits if prev_was_encoding

Right, but for delayed encodings, by definition, we don't have a new encoding just before that instruction (we had an encoding before a previous dynamic instance of that instruction).

derekbruening: > Right, but for delayed encodings, by definition, we don't have a new encoding just before that instruction (we had an encoding before a previous dynamic instance of that instruction).

This *is* the previous dynamic instance: the delayed branch.

I've put in the updated test and new code if you'd like to take a look.

derekbruening: Going to go ahead with this.

PR #6322, opened 2023-09-22
i#3544 RV64: Add preliminary stolen register support

ksco:
This patch adds preliminary support for stolen register. Similar to AArch64, when inside a basic block, we need to steal a register to hold os_local_state. All instructions in the basic block that use stolen register and/or tp need to be mangled.

Issue: #3544


core/arch/arch_exports.h
```diff
@@ -157,9 +157,8 @@ typedef struct _spill_state_t {
     reg_t reg_stolen; /* slot for the stolen register */
 #elif defined(RISCV64)
     reg_t a0, a1, a2, a3;
-    /* Slot for the stolen register, which is tp.
-     * Note that on RISC-V, tp is the thread pointer, and it is also a general-purpose
-     * register, so we steal tp to store DR's tls.
+    /* Slots for the stolen register. Similar to AArch64, we use reg_stolen slot to hold
+     * the guest's stolen reg value.
```
abhinav92003, line 161: Terminology nit: do you mean the original value? We usually use "app" for the original value.
s/guest's/app's/

core/arch/arch_exports.h
```diff
@@ -235,10 +234,12 @@ typedef struct _local_state_extended_t {
 #    define TLS_REG2_SLOT ((ushort)offsetof(spill_state_t, a2))
 #    define TLS_REG3_SLOT ((ushort)offsetof(spill_state_t, a3))
 #    define TLS_REG_STOLEN_SLOT ((ushort)offsetof(spill_state_t, reg_stolen))
+#    define TLS_REG_TP_SLOT os_get_app_tls_base_offset(TLS_REG_LIB)
 #    define SCRATCH_REG0 DR_REG_A0
 #    define SCRATCH_REG1 DR_REG_A1
 #    define SCRATCH_REG2 DR_REG_A2
 #    define SCRATCH_REG3 DR_REG_A3
+#    define SCRATCH_REG_LAST DR_REG_A3
```
abhinav92003, line 242: Seems unused

core/arch/arch_exports.h
```diff
@@ -235,10 +234,12 @@ typedef struct _local_state_extended_t {
 #    define TLS_REG2_SLOT ((ushort)offsetof(spill_state_t, a2))
 #    define TLS_REG3_SLOT ((ushort)offsetof(spill_state_t, a3))
 #    define TLS_REG_STOLEN_SLOT ((ushort)offsetof(spill_state_t, reg_stolen))
+#    define TLS_REG_TP_SLOT os_get_app_tls_base_offset(TLS_REG_LIB)
```
abhinav92003, line 237: Seems unused

core/arch/emit_utils_shared.c
```diff
@@ -2159,10 +2159,10 @@ emit_fcache_enter_common(dcontext_t *dcontext, generated_code_t *code, byte *pc,
                        SCRATCH_REG0 /*scratch*/, false /*to app*/);
 #endif
 
-#ifdef AARCH64
-    /* Put app's X0, X1 in TLS_REG0_SLOT, TLS_REG1_SLOT; this is required by
+    /* Put app's X0/A0, X1/A1 in TLS_REG0_SLOT, TLS_REG1_SLOT; this is required by
```
abhinav92003, line 2162: Would be clearer to say
```
Put app state into TLS_REG0_SLOT, TLS_REG1_SLOT (respectively X0, X1 for AArch64; A0, A1 for RISC-V)
```

core/arch/emit_utils_shared.c
```diff
@@ -4830,16 +4861,28 @@ emit_do_syscall_common(dcontext_t *dcontext, generated_code_t *code, byte *pc,
     APP(&ilist, instr_create_save_to_tls(dcontext, DR_REG_R0, TLS_REG0_SLOT));
     /* XXX: should have a proper patch list entry */
     *syscall_offs += THUMB_LONG_INSTR_SIZE;
-#elif defined(AARCH64)
-    /* For AArch64, we need to save both x0 and x1 into SLOT 0 and SLOT 1
+
+    /* For AArch64/RISCV64, we need to save both x0/a0 and x1/a1 into SLOT 0 and SLOT 1
```
abhinav92003, line 4865: Again: it'd be clearer to mention both archs separately.

core/arch/emit_utils_shared.c
```diff
@@ -4876,8 +4919,10 @@ emit_do_syscall_common(dcontext_t *dcontext, generated_code_t *code, byte *pc,
             instr_create_save_to_dcontext(dcontext, SCRATCH_REG0, SCRATCH_REG0_OFFS));
     }
 
+    /* Save SCRATCH_REG1 as this is used for the indirect branch in the exit stub. */
 #ifdef AARCH64
-    /* Save X1 as this is used for the indirect branch in the exit stub. */
+    APP(&ilist, instr_create_save_to_tls(dcontext, SCRATCH_REG1, TLS_REG1_SLOT));
```
abhinav92003, line 4924: Both statements are exactly the same. Can we put them in the same #ifdef.

abhinav92003 (approved):

core/arch/arch_exports.h
ksco: Removed, will add them back when actually needed.

core/arch/emit_utils_shared.c
ksco: Since there is an `#if defined(ARM)` above at the same level, I separated the comment into two, which seems cleaner.

PR #6325, opened 2023-09-22
i#5036 Fix memval_simple_scattergather failures

jackgallagher-arm:
While developing the the AArch64 version of drx_expand_scatter_gather I have seen the code_api|sample.memval_simple_scattergather fail intermittently with the not yet implemented assert from drx_scatter_gather_restore_state() and tracked down the problem to the trace buffer management mechanism. When the trace buffer is full the next store will be past the end of the buffer and trigger a fault. The fault is meant to be intercepted by the client and used to flush the buffer.

If the buffer happens to be full when tracing an expanded scatter/gather instruction then the scatter/gather state restore function will be called so it would previously just fail with the assert because state restoration is not implemented yet.

Until scatter/gather state restore on AArch64 is implemented properly we can work around the problem by marking the expansion loads and stores we emit with a note that we can check for in the state restore function. If the faulting access does not originate from an expansion instruction then we pass the fault on to the client to handle instead.

Issue: #5036


ext/drx/scatter_gather_aarch64.c
```diff
@@ -673,11 +709,50 @@ bool
...
+     * client to handle, otherwise we can get spurious failures with clients like
+     * memval_simple.
+     */
+    if (info->fragment_info.ilist != NULL) {
+        byte *pc = info->fragment_info.cache_start_pc;
+        for (instr_t *instr = instrlist_first(info->fragment_info.ilist); instr != NULL;
+             instr = instr_get_next(instr)) {
+            if (pc == info->raw_mcontext->pc && !instr_is_label(instr)) {
+                /* Found the faulting instruction */
+                if (!scatter_gather_is_expanded_ld_st(instr)) {
```
abhinav92003, line 736: So this needs the instrlist_t in state restore? I remember drreg has two state restore strategies, with and without ilist, because the ilist is not available always (in some cases we only have the encoded bytes to work with).

Let's add an assert for info->fragment_info.ilist != NULL.

ext/drx/scatter_gather_aarch64.c
```diff
@@ -369,57 +423,39 @@ expand_scalar_plus_vector(void *drcontext, instrlist_t *bb, instr_t *sg_instr,
...
-                case OPSZ_2: EMIT(ldrh, opnd_create_reg(scratch_gpr0_w), mem); break;
-                case OPSZ_4: EMIT(ldr, opnd_create_reg(scratch_gpr0_w), mem); break;
-                case OPSZ_8: EMIT(ldr, opnd_create_reg(scratch_gpr0), mem); break;
-                default: DR_ASSERT_MSG(false, "Invalid scatter_gather_info_t data");
-                }
-            }
+            const reg_id_t scalar_dst = scratch_gpr0;
+
+            /* ldr[bh]  scalar_dst, [mem] */
+            instrlist_preinsert(
```
abhinav92003, line 432: nit: optional: sg_info->is_load is reevaluated inside create_scalar_load_or_store. I wonder if it'll more any more readable/simpler if we split create_scalar_load_or_store into create_scalar_load and create_scalar_store.

ext/drx/scatter_gather_aarch64.c
abhinav92003, line 736: I wonder if we can use drmgr_is_emulation_start and drmgr_is_emulation_end instead for this. Hmm.. but that would not help us differentiate between drmemtrace-added instrumentation and the expansion itself I guess.

ext/drx/scatter_gather_shared.c
```diff
@@ -155,6 +164,8 @@ drx_scatter_gather_init()
         !drmgr_register_thread_exit_event(drx_scatter_gather_thread_exit))
         return false;
 
+    note_base = drmgr_reserve_note_range(SG_NOTE_COUNT);
```
derekbruening, line 167: Check for failure.

abhinav92003 (approved):

ext/drx/scatter_gather_aarch64.c
```diff
@@ -233,6 +233,60 @@ get_scatter_gather_info(instr_t *instr, OUT scatter_gather_info_t *sg_info)
     }
 }
 
+instr_t *
```
abhinav92003, line 236: Actually, one more comment: would it be possible to add a test too? Maybe a scatter-gather with a bad address?

ext/drx/scatter_gather_aarch64.c
```diff
@@ -673,11 +709,50 @@ bool
...
+     *
+     * Previously this function would always assert but this causes a problem with
+     * clients (such as memval_simple) that use drx_buf (or similar) which uses faulting
+     * stores to manage the trace buffer.
+     * Until we implement proper state restoration we need to filter out faults that
+     * don't come from scatter/gather expansion instructions and pass them on to the
+     * client to handle, otherwise we can get spurious failures with clients like
+     * memval_simple.
+     */
+    if (info->fragment_info.ilist != NULL) {
```
derekbruening, line 730: The ilist is not always available, so we may still fail.  Xref #3801 (also #4937).  For now how about just a comment pointing at #3801.

derekbruening (approved):

ext/drx/scatter_gather_aarch64.c
derekbruening: But that would fail -- do you mean a simple client that introduces its own fault inside a scatter-gather region?

abhinav92003: Even the app can recover though right? With a custom signal handler. suite/tests/client-interface/drx-scattergather-x86.c and suite/tests/client-interface/drreg-test.c do it.

derekbruening: I'm guessing you meant something else besides "Maybe a scatter-gather with a bad address": a fault in a scatter or gather load/store will hit the assert as support is not implemented.  I thought we would want a test of what this PR adds, which is not failing for a fault that is not from a scatter/gather address.

abhinav92003: Ah yes. We don't want to hit the assert at L756. So a faulting instr adding by the client in the scatter-gather region is what we want.

ext/drx/scatter_gather_aarch64.c
jackgallagher-arm: As you say, that wouldn't let us distinguish the expansion instructions from instructions added by later instrumentation stages.

ext/drx/scatter_gather_shared.c
jackgallagher-arm: Done.

ext/drx/scatter_gather_aarch64.c
jackgallagher-arm: I have added a comment.

ext/drbbdup/drbbdup.c
```diff
@@ -2425,6 +2427,8 @@ drbbdup_init(drbbdup_options_t *ops_in)
             return DRBBDUP_ERROR;
     }
 
+    note_base = drmgr_reserve_note_range(DRBBDUP_NOTE_COUNT);
```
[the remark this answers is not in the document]
jackgallagher-arm: Done.

ext/drx/scatter_gather_aarch64.c
jackgallagher-arm: I looked at this but I think its simpler as it is now.

ext/drx/scatter_gather_aarch64.c
jackgallagher-arm: I have added comments addressing what happens if `info->fragment_info.ilist == NULL` and it will hit the existing NYI assert.

suite/tests/client-interface/drx-scattergather-instrumentation-fault.dll.c
```diff
@@ -0,0 +1,279 @@
...
+}
+
+static dr_emit_flags_t
+event_bb_app2app(void *drcontext, void *tag, instrlist_t *bb, bool for_trace,
+                 bool translating)
+{
+    if (!drx_expand_scatter_gather(drcontext, bb, NULL)) {
+        DR_ASSERT(false);
+    }
+    drx_tail_pad_block(drcontext, bb);
```
abhinav92003, line 239: Why is this needed?

suite/tests/client-interface/drx-scattergather-instrumentation-fault.dll.c
```diff
@@ -0,0 +1,279 @@
...
+{
+    int i;
+    bool seen_memref = false;
+    instru_data_t *data = (instru_data_t *)user_data;
+
+    /* If the previous instruction was a scatter/gather expansion load/store,
+     * we insert another faulting store after it.
+     */
+    if (data->do_post_instruction_instrumentation) {
+        handle_post_instruction(drcontext, bb, where, data->app_addr);
```
abhinav92003, line 189: Why is this post instr instrumentation needed? Does it test some scenario that the other part of the test doesn't? If not we can simplify this client by removing the post instr instrumentation or else maybe add a comment saying why it is needed.

suite/tests/client-interface/drx-scattergather-instrumentation-fault.dll.c
```diff
@@ -0,0 +1,279 @@
...
+        }
+    }
+}
+
+static dr_emit_flags_t
+event_app_analysis(void *drcontext, void *tag, instrlist_t *bb, bool for_trace,
+                   bool translating, void **user_data)
+{
+    instru_data_t *data = (instru_data_t *)dr_thread_alloc(drcontext, sizeof(*data));
+    data->app_addr = 0;
```
abhinav92003, line 152: This app_addr is only used as a value for the faulting store? We could make it simpler by  using just any other value. Same for the other stuff we compute in the app2app phase (see other comments). Can we just remove the app2app phase completely?

suite/tests/client-interface/drx-scattergather-instrumentation-fault.dll.c
```diff
@@ -0,0 +1,279 @@
...
+    data->is_scatter_gather = false;
```
abhinav92003, line 153: Could we figure this out in instrumentation phase itself using drmg_orig_app_instr_for_fetch?

abhinav92003 (approved):

suite/tests/client-interface/drx-scattergather-instrumentation-fault.dll.c
jackgallagher-arm: Its not the stored value, its the translation address used with `INSTR_XL8`. I have changed it to use `instr_get_app_pc()` inline so we don't have to pass the value around in the user_data. I don't know why I didn't do it that way in the first place.

suite/tests/client-interface/drx-scattergather-instrumentation-fault.dll.c
jackgallagher-arm: I don't know. This client started out as a stripped down memval_simple and memval_simple has it. I have removed it.

suite/tests/client-interface/drx-scattergather-instrumentation-fault.dll.c
```diff
@@ -0,0 +1,279 @@
...
+    drmgr_exit();
+    drx_exit();
+}
+
+DR_EXPORT void
+dr_client_main(client_id_t id, int argc, const char *argv[])
+{
+    drreg_options_t ops = { sizeof(ops), 2 /*max slots needed*/, false };
+
+    dr_set_client_name("DynamoRIO Sample Client 'memval'", "http://dynamorio.org/issues");
```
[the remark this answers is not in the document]
jackgallagher-arm: Done.

suite/tests/client-interface/drx-scattergather-instrumentation-fault.dll.c
abhinav92003: Oh because the scalar loads/stores have their app-pc set to the original scatter-gather. +1 Thanks.

suite/tests/client-interface/drx-scattergather-instrumentation-fault.dll.c
abhinav92003: Thanks.

abhinav92003 (approved):

suite/tests/client-interface/drx-scattergather-instrumentation-fault.dll.c
derekbruening: IIRC this is required for the post-store-instruction instrumentation used by memval and without it it will fail (probably assert in debug build; not sure about release), if this test has the same post instrumentation.

abhinav92003: We removed the post-instrumentation as that did not add any additional coverage to the test.

PR #6331, opened 2023-09-26
i#6238: validate instructions: push,pop,cmp,pusha,popa,push_imm,ret,ret_far,loopne,loope,loop,movzx,cmp,rcr

kuhanov:
Add instructions to categories:
OP_push,OP_pop,OP_cmp,OP_pusha,OP_popa,OP_push_imm,OP_ret,OP_ret_far,OP_loopne,OP_loope,OP_loop,OP_movzx,OP_cmp,OP_rcr

Fix: Moved category decoding after operands decoding to check STORE/LOAD memory access 

Issue: https://github.com/DynamoRIO/dynamorio/issues/6238

### DIFF
--- a/core/ir/x86/decode.c
+++ b/core/ir/x86/decode.c
@@ -2449,11 +2449,17 @@ decode_category(instr_t *instr)
             if (instr_operands_valid(instr)) {
                 if (instr_reads_memory(instr)) {
                     category |= DR_INSTR_CATEGORY_LOAD;
-                    category &= ~DR_INSTR_CATEGORY_MOVE;
+                    if (TEST(DR_INSTR_CATEGORY_MOVE, category)) {
+                        category &= ~DR_INSTR_CATEGORY_MOVE;
+                        category &= ~DR_INSTR_CATEGORY_FP;
+                    }
                 }
                 if (instr_writes_memory(instr)) {
                     category |= DR_INSTR_CATEGORY_STORE;
-                    category &= ~DR_INSTR_CATEGORY_MOVE;
+                    if (TEST(DR_INSTR_CATEGORY_MOVE, category)) {
+                        category &= ~DR_INSTR_CATEGORY_MOVE;
+                        category &= ~DR_INSTR_CATEGORY_FP;
+                    }
                 }
             }
             instr_set_category(instr, category);
--- a/core/ir/x86/decode.c
+++ b/core/ir/x86/decode.c
@@ -2447,10 +2447,14 @@ decode_category(instr_t *instr)
         if (op_instr[instr->opcode] != NULL) {
             uint category = op_instr[instr->opcode]->category;
             if (instr_operands_valid(instr)) {
-                if (instr_reads_memory(instr))
+                if (instr_reads_memory(instr)) {
                     category |= DR_INSTR_CATEGORY_LOAD;
-                if (instr_writes_memory(instr))
+                    category &= ~DR_INSTR_CATEGORY_MOVE;
+                }
+                if (instr_writes_memory(instr)) {
                     category |= DR_INSTR_CATEGORY_STORE;
+                    category &= ~DR_INSTR_CATEGORY_MOVE;
+                }
             }
             instr_set_category(instr, category);
         } else {
--- a/core/ir/x86/decode.c
+++ b/core/ir/x86/decode.c
@@ -2591,7 +2591,6 @@ decode_common(dcontext_t *dcontext, byte *pc, byte *orig_pc, instr_t *instr)
                                         decode operands too */
                                _IF_DEBUG(!TEST(INSTR_IGNORE_INVALID, instr->flags)));
     instr_set_opcode(instr, info->type);
-    decode_category(instr);
     IF_X64(instr_set_x86_mode(instr, di.x86_mode));
     /* failure up to this point handled fine -- we set opcode to OP_INVALID */
     if (next_pc == NULL) {
@@ -2735,6 +2734,8 @@ decode_common(dcontext_t *dcontext, byte *pc, byte *orig_pc, instr_t *instr)
         /* We must do this AFTER setting raw bits to avoid being invalidated. */
         instr_set_rip_rel_pos(instr, (int)(di.disp_abs - di.start_pc));
     }
+
+    decode_category(instr);
 
     return next_pc;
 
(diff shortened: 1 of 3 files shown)


suite/tests/api/drdecode_x86.c
```diff
@@ -172,12 +172,12 @@ test_categories(void)
     /*  55 OP_mov_ld */
     instr = XINST_CREATE_load(GD, opnd_create_reg(DR_REG_XAX),
                               OPND_CREATE_MEMPTR(DR_REG_XAX, 42));
-    CHECK_CATEGORY(GD, instr, buf, DR_INSTR_CATEGORY_MOVE);
+    CHECK_CATEGORY(GD, instr, buf, (DR_INSTR_CATEGORY_MOVE | DR_INSTR_CATEGORY_LOAD));
```
derekbruening, line 175: Hmm, is it weird that MOV is MOVE|LOAD yet POP is just LOAD?  Both are moving data around.  Should we have the logic that adds the LOAD also remove any MOVE?  Or should we add MOVE to all the PUSH and POP opcodes?  How do these operate on AArch64?  I'm guessing that there, there is no LOAD that is also MOVE?  So probably we should follow that convention?  (Except REP MOVS or a PUSH or POP which is both LOAD and STORE as they are moving from one memory slot to another and not loading/storing to a register endpoint??  Gray areas...)

core/ir/x86/decode.c
```diff
@@ -2591,7 +2591,6 @@ decode_common(dcontext_t *dcontext, byte *pc, byte *orig_pc, instr_t *instr)
                                         decode operands too */
```
derekbruening, line 2591: > i#6238: validate instructions: OP_push,OP_pop,OP_cmp,OP_pusha,OP_popa,OP_push_imm,OP_ret,OP_ret_far,OP_loopne,OP_loope,OP_loop,OP_movzx,OP_cmp,OP_rcr 

Title is very long and likely to be truncated in various summary tools.  How about shrinking a bit: "i#6238: Fill in categories for x86 push,pop,ret,loop,movzx,cmp,rcr".  (Leaving the full list in the body.)

derekbruening (approved):

suite/tests/api/drdecode_x86.c
kuhanov: I think we need to remove move in this case. In my analysis I skip move in this case.
Thx, Kirill

derekbruening:
> remove FP | MOVE if we change the default category to LOAD and/or STORE

You removed FP for LOAD/STORE?  I think people would want to distinguish a load or store to an FP reg from a GPR.

kuhanov:
> > remove FP | MOVE if we change the default category to LOAD and/or STORE
> 
> You removed FP for LOAD/STORE? I think people would want to distinguish a load or store to an FP reg from a GPR.

Hm. Let me add to new types for FP to eclude subcategory assert in this case.
Is it ok?
```
diff --git a/core/ir/instr_api.h b/core/ir/instr_api.h
index 995f7267c..366314ae9 100644
--- a/core/ir/instr_api.h
+++ b/core/ir/instr_api.h
@@ -1918,6 +1918,8 @@ typedef enum {
     DR_FP_MOVE,    /**< Moves floating point values from one location to another. */
     DR_FP_CONVERT, /**< Converts to or from floating point values. */
     DR_FP_MATH,    /**< Performs arithmetic or conditional operations. */
+    DR_FP_LOAD,    /**< Loads. */
+    DR_FP_STORE,   /**< Stores. */
 } dr_fp_type_t;
 
 DR_API
diff --git a/core/ir/x86/decode.c b/core/ir/x86/decode.c
index 78e074203..ca05ad645 100644
--- a/core/ir/x86/decode.c
+++ b/core/ir/x86/decode.c
@@ -2449,17 +2449,11 @@ decode_category(instr_t *instr)
             if (instr_operands_valid(instr)) {
                 if (instr_reads_memory(instr)) {
                     category |= DR_INSTR_CATEGORY_LOAD;
-                    if (TEST(DR_INSTR_CATEGORY_MOVE, category)) {
-                        category &= ~DR_INSTR_CATEGORY_MOVE;
-                        category &= ~DR_INSTR_CATEGORY_FP;
-                    }
+                    category &= ~DR_INSTR_CATEGORY_MOVE;
                 }
                 if (instr_writes_memory(instr)) {
                     category |= DR_INSTR_CATEGORY_STORE;
-                    if (TEST(DR_INSTR_CATEGORY_MOVE, category)) {
-                        category &= ~DR_INSTR_CATEGORY_MOVE;
-                        category &= ~DR_INSTR_CATEGORY_FP;
-                    }
+                    category &= ~DR_INSTR_CATEGORY_MOVE;
                 }
             }
             instr_set_category(instr, category);
diff --git a/core/ir/x86/instr.c b/core/ir/x86/instr.c
index 618adba98..a08b51496 100644
--- a/core/ir/x86/instr.c
+++ b/core/ir/x86/instr.c
@@ -930,6 +930,14 @@ instr_is_floating_ex(instr_t *instr, dr_fp_type_t *type OUT)
         if (type != NULL)
             *type = DR_FP_MOVE;
         return true;
+    } else if (TEST(DR_INSTR_CATEGORY_STORE, cat)) {
+        if (type != NULL)
+            *type = DR_FP_STORE;
+        return true;
+    } else if (TEST(DR_INSTR_CATEGORY_LOAD, cat)) {
+        if (type != NULL)
+            *type = DR_FP_LOAD;
+        return true;
     } else {
         CLIENT_ASSERT(false, "instr_is_floating_ex: FP instruction without subcategory");
         return false;
```

derekbruening:
I don't think we should be making compound categories like FP_MOVE, FP_STORE, or FP_LOAD when we can compose the existing categories: what is the advantage of FP_STORE over FP|STORE?  I would think combining FP with the other categories is what we want.

kuhanov:
> I don't think we should be making compound categories

Let's remove dr_fp_type_t at all in this case and migrate to categories.
I could cteate patch tommorow.
Do you agree?
Kirill

derekbruening:
Oh I misread it: I thought that patch in your comment https://github.com/DynamoRIO/dynamorio/pull/6331#issuecomment-1737702631 was adding DR_INSTR_CATEGORY_FP_LOAD.  I don't think we need to update the older dr_fp_type_t.  Marking it as deprecated (there is a Doxygen tag for that) sounds fine.

PR #6339, opened 2023-09-29
i#5694 core-sharded: Add stream input query interfaces

derekbruening:
Adds several routines to the memtrace_stream_t interface for drmemtrace analysis tools in core-sharded mode:
+ get_workload_id()
+ get_input_id()
+ get_input_interface()

Adds a new analysis_unit_tests executable with some sanity tests. Splits out the mock_reader_t and helpers used by scheduler_unit_tests to share them with the new test.

Documents the additions.

Fixes a bug with core-sharded analysis tools where parallel_shard_exit was called for every thread, resulting use-after-frees when more than one thread was on a core and the tool deletes its shard data structure in the parallel_shard_exit routine (most of our tools do not, which is why this was not noticed before).

Issue: #5694

### DIFF
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -149,6 +149,9 @@ Further non-compatibility-affecting changes include:
  - Added type_is_read() API that returns true if a trace type reads from memory.
  - Added instr_num_memory_read_access() and instr_num_memory_write_access() that return
    the number of memory read and write accesses of an instruction respectively.
+ - Added several routines to the #dynamorio::drmemtrace::memtrace_stream_t interface
+   for drmemtrace analysis tools: get_output_cpuid(), get_workload_id(),
+   get_input_id(), get_input_interface().
 
 **************************************************
 <hr>
--- a/clients/drcachesim/analyzer.cpp
+++ b/clients/drcachesim/analyzer.cpp
@@ -471,6 +471,29 @@ analyzer_tmpl_t<RecordType, ReaderType>::process_tasks(analyzer_worker_data_t *w
 
     for (int i = 0; i < num_tools_; ++i)
         user_worker_data[i] = tools_[i]->parallel_worker_init(worker->index);
+
+    auto process_shard_exit = [=](int shard_index) {
+        VPRINT(this, 1, "Worker %d finished trace shard %s\n", worker->index,
+               worker->stream->get_stream_name().c_str());
+        if (interval_microseconds_ != 0 &&
+            !process_interval(
+                worker->shard_data[shard_index].cur_interval_index,
+                worker->shard_data[shard_index].cur_interval_init_instr_count, worker,
+                /*parallel=*/true, shard_index))
+            return;
+        for (int i = 0; i < num_tools_; ++i) {
+            if (!tools_[i]->parallel_shard_exit(
+                    worker->shard_data[shard_index].tool_data[i].shard_data)) {
+                worker->error = tools_[i]->parallel_shard_error(
+                    worker->shard_data[shard_index].tool_data[i].shard_data);
+                VPRINT(this, 1, "Worker %d hit shard exit error %s on trace shard %s\n",
+                       worker->index, worker->error.c_str(),
+                       worker->stream->get_stream_name().c_str());
+                return;
+            }
+        }
+    };
+
     RecordType record;
     // The current time is used for time quanta; for instr quanta, it's ignored and
     // we pass 0.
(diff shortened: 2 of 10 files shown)


derekbruening:
Failure is 32-bit scattergather: not the fixed pipe assert, but the type_is_instr assert: #3320

api/docs/release.dox
```diff
@@ -149,6 +149,9 @@ Further non-compatibility-affecting changes include:
  - Added type_is_read() API that returns true if a trace type reads from memory.
  - Added instr_num_memory_read_access() and instr_num_memory_write_access() that return
    the number of memory read and write accesses of an instruction respectively.
+ - Added several routines to the #dynamorio::drmemtrace::memtrace_stream_t interface
+   for drmemtrace analysis tools: get_output_cpuid(), get_workload_id(),
```
abhinav92003, line 153: How is get_output_cpuid related to the shard_id in parallel_shard_init?

clients/drcachesim/analyzer.cpp
```diff
@@ -471,6 +471,29 @@ analyzer_tmpl_t<RecordType, ReaderType>::process_tasks(analyzer_worker_data_t *w
 
     for (int i = 0; i < num_tools_; ++i)
         user_worker_data[i] = tools_[i]->parallel_worker_init(worker->index);
+
+    auto process_shard_exit = [=](int shard_index) {
```
abhinav92003, line 475: Instead of the by-value capture clause, maybe take specific data by ref? Feels safer.

Also, curious why we made this a lambda instead of a simple routine?

clients/drcachesim/common/memtrace_stream.h
```diff
@@ -167,6 +167,40 @@ class memtrace_stream_t {
     {
         return -1;
     }
+
+    /**
+     * Returns a unique identifier for the current workload.  This might be an
+     * ordinal from the list of active workloads, or some other identifier.
```
abhinav92003, line 173: Should we return a PID instead? More meaningful than an ordinal and still unique enough (not guaranteed unique but we can live with that IMHO).

abhinav92003: Edit:
oh maybe that's what your comment here is saying "or some other identifier". Worth adding "like the PID"?

clients/drcachesim/scheduler/scheduler.h
```diff
@@ -775,6 +775,35 @@ template <typename RecordType, typename ReaderType> class scheduler_tmpl_t {
...
         }
 
+        /**
+         * Returns the ordinal for the current
+         * #dynamorio::drmemtrace::scheduler_tmpl_t::input_workload_t.
+         */
+        int64_t
+        get_workload_id() const override
+        {
+            return static_cast<int64_t>(get_input_workload_ordinal());
```
abhinav92003, line 785: Maybe here is the right place to comment this: why not return the PID here and the TID below, which are more meaningful than ordinals?

clients/drcachesim/tests/analysis_unit_tests.cpp
```diff
@@ -0,0 +1,189 @@
...
+            return true;
+        }
+        bool
+        parallel_shard_memref(void *shard_data, const memref_t &memref) override
+        {
+            per_shard_t *shard = reinterpret_cast<per_shard_t *>(shard_data);
+            // These are our testing goals: these queries.
+            // We have one thread for each of our NUM_INPUTS workloads.
+            assert(shard->stream->get_output_cpuid() == shard->index);
+            assert(shard->stream->get_workload_id() == memref.instr.tid - BASE_TID);
```
abhinav92003, line 156: Workload id is same as the input id?

api/docs/release.dox
abhinav92003: Oh I see the test. They're the same, as expected.

clients/drcachesim/tests/analysis_unit_tests.cpp
abhinav92003: Oh that's because the way we've set up this test: each thread is its own workload.

Add a comment.

abhinav92003 (approved):

api/docs/release.dox
derekbruening: Right, this was added previously (but not to this release list) so you're not seeing its docs in this PR: the docs do say it's the output cpu ordinal for dynamic schedules.

clients/drcachesim/common/memtrace_stream.h
derekbruening: For internal use we are using a UUID here.  For general use, there is already a pid in the memref_t struct, so having it here doesn't seem to add much.  The pid can be duplicated so a guaranteed-unique value was the idea here: UUID internally; ordinal externally.

derekbruening: I added "This is guaranteed to
     * be unique among all inputs, unlike the process and thread identifiers in
     * #memref_t.".

clients/drcachesim/tests/analysis_unit_tests.cpp
derekbruening: Done.

clients/drcachesim/analyzer.cpp
derekbruening: Just felt like an inner function.  Moved it to a top-level helper.

derekbruening:
windows x64 failure is a drcacheoff.max-global timeout.  I am trying to figure out how it could be caused by the changes here which seem like they would not affect it at all.

derekbruening:
It passed on a re-run.  It seems it must just be close to the limit which is 4 minutes.  On my Windows machine it actually takes 6 minutes run w/o a timeout, and times out reliably under ctest.  I will file a separate issue on it.

derekbruening:
The max-global timeout I filed as #6342 and fixed in #6343.

PR #6343, opened 2023-10-03
i#6342: Speed up function tracing on Windows

### LINKED ISSUE #6342: tool.drcacheoff.max-global test is very slow on Windows, often timing out
https://github.com/DynamoRIO/dynamorio/actions/runs/6384343147/job/17328624474
```
254/302 Test #245: code_api|tool.drcacheoff.max-global ..............................***Timeout 241.17 sec
```

On my local Windows machine it takes > 6 minutes!
```
% /usr/bin/time bin64/drrun -t drcachesim -offline -max_trace_size 1M -trace_after_instrs 20K -max_global_trace_refs 10K -record_heap -- suite/tests/bin/client.annotation-concurrency.exe client.annotation-concurrency.appdll.dll A 4 64 3
<...>
<Stopping application D:\derek\dr\git\build_x64_dbg_tests\suite\tests\bin\client.annotation-concurrency.exe (10600)>
0.00user 0.03system 6:17.83elapsed 0%CPU (0avgtext+0avgdata 5232maxresident)k
```

It looks like all the time is in looking up symbols for `-record_heap`: without that parameter it takes <30s.  Xref Dr. Memory caching symbols due to very slow Windows lookups so this is a common problem with dbghelp.dll.

derekbruening:
Looking up symbols is very slow on Windows.  Here we avoid looking at dynamorio.dll and the client and extension libraries.

For the drcacheoff.max-global test, we also reduce the number of functions from -record_heap to just -record_function 'malloc|1'.

Combined, these shrink the test from taking over 6 minutes on my local machine to taking 24 seconds.  The test was timing out periodically on GA CI and this should fix that.

Fixes #6342

### DIFF
--- a/clients/drcachesim/tracer/func_trace.cpp
+++ b/clients/drcachesim/tracer/func_trace.cpp
@@ -237,7 +237,16 @@ instru_funcs_module_load(void *drcontext, const module_data_t *mod, bool loaded)
 {
     if (drcontext == NULL || mod == NULL)
         return;
-
+#ifndef DRMEMTRACE_STATIC
+    // Skip DR itself and the client and its libraries as symbol lookup is slow
+    // on Windows and the fewer libs we check the better (i#6342), unless we're
+    // statically linked (when the app itself might be excluded here).
+    if (dr_memory_is_dr_internal(mod->start) || dr_memory_is_in_client(mod->start)) {
+        NOTIFY(3, "Not looking for symbols in DR/client library %s\n",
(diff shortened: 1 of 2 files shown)


derekbruening:
The GA CI for this PR shows the improved test time, nowhere near its 240s limit for 64-bit Windows:
```
246/302 Test #245: code_api|tool.drcacheoff.max-global ..............................   Passed   16.38 sec
```

PR #6347, opened 2023-10-05
Add raw2trace statistic for instr count

abhinav92003:
Adds a new raw2trace statistic that counts the number of instructions written to the final trace.

This instr count is printed out along with other raw2trace stats at the end of raw2trace conversion. It is also available for sub-classes to obtain using the get_statistic API.

Adds tests for the new and existing raw2trace stats in raw2trace_unit_tests.

Verified that the log printed by raw2trace has the same instr count as what's printed by basic_counts, for both the default .zip and `-compress none`.

### DIFF
--- a/clients/drcachesim/tests/raw2trace_unit_tests.cpp
+++ b/clients/drcachesim/tests/raw2trace_unit_tests.cpp
@@ -183,11 +183,11 @@ make_pid()
 }
 
 offline_entry_t
-make_tid()
+make_tid(memref_tid_t tid = 1)
 {
     offline_entry_t entry;
     entry.tid.type = OFFLINE_TYPE_THREAD;
-    entry.tid.tid = 1;
+    entry.tid.tid = tid;
     return entry;
 }
 
@@ -294,10 +294,22 @@ check_entry(std::vector<trace_entry_t> &entries, int &idx, unsigned short expect
     return true;
 }
 
+void
+populate_all_stats(raw2trace_test_t &raw2trace, std::vector<uint64_t> *stats)
+{
+    if (stats == nullptr)
+        return;
+    for (int i = 0; i < RAW2TRACE_STAT_MAX; ++i) {
+        stats->push_back(raw2trace.get_statistic(
+            static_cast<dynamorio::drmemtrace::raw2trace_statistic_t>(i)));
+    }
+}
+
 // Takes ownership of ilist and destroys it.
 bool
 run_raw2trace(void *drcontext, const std::vector<offline_entry_t> raw, instrlist_t *ilist,
-              std::vector<trace_entry_t> &entries, int chunk_instr_count = 0,
+              std::vector<trace_entry_t> &entries, std::vector<uint64_t> *stats = nullptr,
+              int chunk_instr_count = 0,
               const std::vector<test_multi_module_mapper_t::bounds_t> &modules = {})
 {
     // We need an istream so we use istringstream.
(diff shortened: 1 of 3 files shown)


clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1255,6 +1257,7 @@ class raw2trace_t {
     uint64 count_rseq_side_exit_ = 0;
     uint64 earliest_trace_timestamp_ = (std::numeric_limits<uint64>::max)();
     uint64 latest_trace_timestamp_ = 0;
+    uint64 final_trace_instr_count_ = 0;
```
derekbruening, line 1260: How about a sanity unit test in raw2trace_unit_tests with a couple of threads.  Looking: surprisingly there is no test at all of any of these stats.  Would be good to add something to at least exercise this.

derekbruening (approved):

clients/drcachesim/tracer/raw2trace.h
abhinav92003: Done. Would be good to do a re-review for the new test code.

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -3156,12 +3156,21 @@ raw2trace_t::write(raw2trace_thread_data_t *tdata, const trace_entry_t *start,
...
         }
     }
-    if (end > start &&
-        !tdata->out_file->write(reinterpret_cast<const char *>(start),
-                                reinterpret_cast<const char *>(end) -
-                                    reinterpret_cast<const char *>(start))) {
-        tdata->error = "Failed to write to output file";
-        return false;
+    if (end > start) {
+        // We reach here also when we're not writing to a zip archive.
```
derekbruening, line 3160: What do you mean "also" -- do we come into here for a zip output too?  If so won't the instr count then be wrong?

derekbruening (approved):

clients/drcachesim/tracer/raw2trace.cpp
abhinav92003: I thought in some cases there may be some entries pending to be written to the archive. Otherwise I'd've expected this logic to be part of an `else` clause for `if (tdata->out_archive != nullptr)`, instead of a separate if condition. Is that not the case?

clients/drcachesim/tests/raw2trace_unit_tests.cpp
```diff
@@ -294,10 +294,22 @@ check_entry(std::vector<trace_entry_t> &entries, int &idx, unsigned short expect
     return true;
 }
 
+void
+populate_all_stats(raw2trace_test_t &raw2trace, std::vector<uint64_t> *stats)
+{
+    if (stats == nullptr)
+        return;
+    for (int i = 0; i < RAW2TRACE_STAT_MAX; ++i) {
```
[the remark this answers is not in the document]
abhinav92003: Enum values cannot be incremented or decremented apparently.

clients/drcachesim/tracer/raw2trace.cpp
abhinav92003: Oh wait. It is overcounting. I did not re-verify the zip instr count after I made this part of the change in 62c085ead6bb80c8c0e4b7b8a341aafc2318406e. Let me fix this.

abhinav92003: I was overcounting the last chunk's data. Fixed it.

abhinav92003: But control indeed reaches here even for zip output -- for the last chunk.

abhinav92003:
x86-64-ubuntu22 api.rseq failure is #6185.

PR #6348, opened 2023-10-05
i#6344: Add -record_syscall to drmemtrace

### LINKED ISSUE #6344: Add -record_syscall feature to record syscall parameters in drmemtrace
For #5843, PR #6132 added hardcoded recording of SYS_futex parameters.  This is a feature request to generalize that to support recording parameters for arbitrary syscalls, specified by a new option `-record_syscall` which would look like `-record_function` where the function name is the syscall number.  We don't want to record all syscall parameters with no way to disable as that could be undesirable in workloads where values other than neutral addresses may pose privacy concerns (though syscall parameters are generally not problematic as opposed to function parameter values).

derekbruening:
Adds a new option -record_sysall to drmemtrace which records the parameter and success values for the given system call numbers.  Just like with -record_function, the user must specify the parameter count.

SYS_futex is left as traced by default, but it can be disabled.

Adds documentation and a test.

Further manual testing:
```
  ---------------------------------------------------------------------------
  $ rm -rf drmemtrace.*.dir; bin64/drrun -t drcachesim -offline -record_syscall '1|-3&12|4&9|2' -record_syscall '12|2&158|4' -- suite/tests/bin/simple_app && bin64/drrun -t drcachesim -indir drmemtrace.*.dir -simulator_type view 2>&1 | egrep 'system call |function'
  Error: -record_syscall invalid parameter count -3
  ---------------------------------------------------------------------------

  ---------------------------------------------------------------------------
  $ rm -rf drmemtrace.*.dir; bin64/drrun -t drcachesim -offline -record_syscall '1|-3&12|4&9|2' -record_syscall '12|2&158|4' -- suite/tests/bin/simple_app && bin64/drrun -t drcachesim -indir drmemtrace.*.dir -simulator_type view 2>&1 | egrep 'system call |function'
  <Application simple_app (484125) DynamoRIO usage error : invalid system call parameter number>
  ---------------------------------------------------------------------------

  ---------------------------------------------------------------------------
  $ rm -rf drmemtrace.*.dir; bin64/drrun -t drcachesim -offline -record_syscall '1|3&12|4&9|2' -record_syscall '12|2&158|4' -- suite/tests/bin/simple_app && bin64/drrun -t drcachesim -indir drmemtrace.*.dir -simulator_type view 2>&1 | egrep 'system call |function'
  ...
  Hello, world!
  <Stopping application simple_app (484049)>
         32411       26808:      484049 <marker: system call 12>
         32412       26808:      484049 <marker: function==syscall #12>
         32413       26808:      484049 <marker: function argument 0x0>
         32414       26808:      484049 <marker: function argument 0x7ffc87c52d4c>
         32415       26808:      484049 <marker: function==syscall #12>
         32416       26808:      484049 <marker: function return value 0x1>
         50436       41126:      484049 <marker: system call 9>
         50437       41126:      484049 <marker: function==syscall #9>
         50438       41126:      484049 <marker: function argument 0x0>
         50439       41126:      484049 <marker: function argument 0x2000>
         50440       41126:      484049 <marker: function==syscall #9>
         50441       41126:      484049 <marker: function return value 0x1>
         50980       41485:      484049 <marker: system call 21>
         52193       42391:      484049 <marker: system call 257>
         52223       42409:      484049 <marker: system call 262>
         52253       42430:      484049 <marker: system call 9>
         52254       42430:      484049 <marker: function==syscall #9>
         52255       42430:      484049 <marker: function argument 0x0>
         52256       42430:      484049 <marker: function argument 0x1b5c7>
         52257       42430:      484049 <marker: function==syscall #9>
         52258       42430:      484049 <marker: function return value 0x1>
         52274       42439:      484049 <marker: system call 3>
         54224       44056:      484049 <marker: system call 257>
         54252       44074:      484049 <marker: system call 0>
         54329       44130:      484049 <marker: system call 17>
         54455       44208:      484049 <marker: system call 262>
         54988       44591:      484049 <marker: system call 17>
         55653       45105:      484049 <marker: system call 9>
         55654       45105:      484049 <marker: function==syscall #9>
         55655       45105:      484049 <marker: function argument 0x0>
         55656       45105:      484049 <marker: function argument 0x1e1f50>
         55657       45105:      484049 <marker: function==syscall #9>
         55658       45105:      484049 <marker: function return value 0x1>
         ...
  ---------------------------------------------------------------------------
```
Fixes #6344

### DIFF
--- a/clients/drcachesim/common/options.cpp
+++ b/clients/drcachesim/common/options.cpp
@@ -749,6 +749,24 @@ droption_t<bool> op_record_replace_retaddr(
     "replacement, which has lower overhead, but runs the risk of breaking an "
     "application that examines or changes its own return addresses in the recorded "
     "functions.");
+droption_t<std::string> op_record_syscall(
+    DROPTION_SCOPE_CLIENT, "record_syscall", DROPTION_FLAG_ACCUMULATE,
+    OP_RECORD_FUNC_ITEM_SEP, "", "Record parameters for the specified syscall number(s).",
+    "Record the parameters and success of the specified system call number(s)."
+    " The option value should fit this format:"
+    " sycsall_number|parameter_number"
+    " E.g., -record_syscall \"2|2\" will record SYS_open's 2 parameters and whether"
+    " successful (1 for success or 0 for failure, in a function return value record)"
+    " for x86 Linux.  SYS_futex is recorded by default on Linux and this option's value"
+    " adds to futex rather than replacing it (setting futex to 0 parameters disables)."
+    " The trace identifies which syscall owns each set of parameter and return value"
+    " records via a numeric ID equal to the syscall number + TRACE_FUNC_ID_SYSCALL_BASE."
+    " Recording multiple syscalls can be achieved by using the separator"
+    " \"" OP_RECORD_FUNC_ITEM_SEP
+    "\" (e.g., -record_syscall \"202|6" OP_RECORD_FUNC_ITEM_SEP "3|1\"), or"
+    " specifying multiple -record_syscall options."
+    " It is up to the user to ensure the values are correct; a too-large parameter"
+    " count may cause tracing to fail with an error mid-run.");
 droption_t<unsigned int> op_miss_count_threshold(
     DROPTION_SCOPE_FRONTEND, "miss_count_threshold", 50000,
     "For cache miss analysis: minimum LLC miss count for a load to be eligible for "
--- a/api/docs/release.dox
+++ b/api/docs/release.dox
@@ -157,6 +157,7 @@ Further non-compatibility-affecting changes include:
  - Added several routines to the #dynamorio::drmemtrace::memtrace_stream_t interface
    for drmemtrace analysis tools: get_output_cpuid(), get_workload_id(),
    get_input_id(), get_input_interface().
+ - Added -record_syscall to drmemtrace for recording syscall parameters.
 
 **************************************************
 <hr>
--- a/clients/drcachesim/common/options.h
+++ b/clients/drcachesim/common/options.h
@@ -57,6 +57,10 @@
 #define CACHE_TYPE_DATA "data"
 #define CACHE_TYPE_UNIFIED "unified"
 #define CACHE_PARENT_MEMORY "memory"
+// The expected pattern for a single_op_value is:
+//     function_name|function_id|arguments_num
+// where function_name can contain spaces (for instance, C++ namespace prefix)
+#define PATTERN_SEPARATOR "|"
 
 #ifdef HAS_ZIP
 #    define DEFAULT_TRACE_COMPRESSION_TYPE "zip"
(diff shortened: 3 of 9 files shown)


clients/drcachesim/common/options.cpp
```diff
@@ -749,6 +749,24 @@ droption_t<bool> op_record_replace_retaddr(
...
+droption_t<std::string> op_record_syscall(
+    DROPTION_SCOPE_CLIENT, "record_syscall", DROPTION_FLAG_ACCUMULATE,
+    OP_RECORD_FUNC_ITEM_SEP, "", "Record parameters for the specified syscall number(s).",
+    "Record the parameters and success of the specified system call number(s)."
+    " The option value should fit this format:"
+    " sycsall_number|parameter_number"
+    " E.g., -record_syscall \"2|2\" will record SYS_open's 2 parameters and whether"
+    " successful (1 for success or 0 for failure, in a function return value record)"
+    " for x86 Linux.  SYS_futex is recorded by default on Linux and this option's value"
+    " adds to futex rather than replacing it (setting futex to 0 parameters disables)."
```
abhinav92003, line 761: > setting futex to 0 parameters disables

I would expect setting the number of parameters to just reduce the num of parameters recorded but not disable it entirely. Since it's documented, it's probably okay. But do we want to support gathering only system call success and id vals (maybe the user doesn't want the syscall params for some reason?).

clients/drcachesim/common/options.cpp
```diff
@@ -749,6 +749,24 @@ droption_t<bool> op_record_replace_retaddr(
...
+    " The trace identifies which syscall owns each set of parameter and return value"
+    " records via a numeric ID equal to the syscall number + TRACE_FUNC_ID_SYSCALL_BASE."
+    " Recording multiple syscalls can be achieved by using the separator"
+    " \"" OP_RECORD_FUNC_ITEM_SEP
+    "\" (e.g., -record_syscall \"202|6" OP_RECORD_FUNC_ITEM_SEP "3|1\"), or"
+    " specifying multiple -record_syscall options."
+    " It is up to the user to ensure the values are correct; a too-large parameter"
```
abhinav92003, line 768: If the parameter count is too large, why would it cause a crash? I'd imagine we'd simply be writing more reg values to the trace.

abhinav92003 (approved):

clients/drcachesim/common/options.cpp
derekbruening: It is a little weird but I liked the idea of having futex be on by default without trying to A) get the SYS_ value for all OS's inside options.cpp for a default value and B) the user having to re-specify futex to avoid disabling when trying to just add a syscall.  Alternatives could be a separate -record_futex parameter or having a parameter count of -1 disabling?? I would lean toward as-is with 0 disabling; it doesn't seem likely you would just want success.

clients/drcachesim/common/options.cpp
derekbruening: DR fails dr_syscall_get_param() with `CLIENT_ASSERT(false, "invalid system call parameter number");`: https://github.com/DynamoRIO/dynamorio/blob/master/core/unix/os.c#L5403.  (On Windows where params are on the stack I think there is no issue; it's only Linux where there is a fixed maximum b/c they're in certain registers only and don't escape to the stack.)

Issue #6334, opened 2023-09-27
Invariant checker reports "Serial schedule entry count does not match trace" error with thread ID -1


ivankyluk:
Invariant checker reports "Serial schedule entry count does not match trace" error with thread ID -1.

PR #6349, opened 2023-10-05
i#6334: use the same comparator in raw2trace and invariant_checker

prasun3:
Use the same comparator for sorting scheduler entries to prevent checker failures due to schedule mismatches.
Add a test that sometimes shows this failure.

Issue: #6334

### DIFF
--- a/clients/drcachesim/tracer/raw2trace.cpp
+++ b/clients/drcachesim/tracer/raw2trace.cpp
@@ -1214,6 +1214,7 @@ raw2trace_t::do_conversion()
                 earliest_trace_timestamp_, thread_data_[i]->earliest_trace_timestamp);
             latest_trace_timestamp_ = std::max(latest_trace_timestamp_,
                                                thread_data_[i]->latest_trace_timestamp);
+            final_trace_instr_count_ += thread_data_[i]->final_trace_instr_count;
         }
     } else {
         // The files can be converted concurrently.
@@ -1238,6 +1239,7 @@ raw2trace_t::do_conversion()
                 std::min(earliest_trace_timestamp_, tdata->earliest_trace_timestamp);
             latest_trace_timestamp_ =
                 std::max(latest_trace_timestamp_, tdata->latest_trace_timestamp);
+            final_trace_instr_count_ += tdata->final_trace_instr_count;
         }
     }
     error = aggregate_and_write_schedule_files();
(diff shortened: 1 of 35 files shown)


clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -1276,10 +1276,21 @@ raw2trace_t::aggregate_and_write_schedule_files()
...
+                                        const schedule_entry_t &r) {
+        if (l.timestamp != r.timestamp)
+            return l.timestamp < r.timestamp;
+        if (l.cpu != r.cpu)
+            return l.cpu < r.cpu;
+        // We really need to sort by either (timestamp, cpu_id) or
+        // (timestamp, thread_id): a single thread cannot be on two CPUs at
+        // the same timestamp; also a single CPU cannot have two threads at the
+        // same timestamp. We still sort by (timestamp, cpu_id, thread_id)
+        // to prevent inadvertent issues with test data.
```
derekbruening, line 1289: Instead of duplicating this long comment verbatim how about having one of them point at the other ("See comment in ...").

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -1276,10 +1276,21 @@ raw2trace_t::aggregate_and_write_schedule_files()
             vec.insert(vec.end(), keyval.second.begin(), keyval.second.end());
         }
     }
-    std::sort(serial.begin(), serial.end(),
-              [](const schedule_entry_t &l, const schedule_entry_t &r) {
-                  return l.timestamp < r.timestamp;
-              });
+    auto schedule_entry_comparator = [](const schedule_entry_t &l,
```
derekbruening, line 1279: Add comments both here and in invariant check's comparison that if one is changed the other must be changed, to avoid future inconsistency (or, share the code by exporting from raw2trace, but not sure how messy that gets).

derekbruening (approved):
Thank you for tracking down the problem and contributing a fix.

clients/drcachesim/tracer/raw2trace.cpp
prasun3: Done.

clients/drcachesim/tracer/raw2trace.cpp
prasun3: Done.

PR #6352, opened 2023-10-06
Fix include guard in memcache.h

kiri11:
The current one doesn't work because defined name is different from the one that's checked with `#ifndef`.
This can lead to including this header more than once.

Also changed `#define` formatting to be consistent with other files.

### DIFF
--- a/core/unix/memcache.h
+++ b/core/unix/memcache.h
@@ -31,7 +31,7 @@
  */
 
 #ifndef _MEMCACHE_H_
-#    define _MEMCACHE_H_ 1
+#define _MEMCACHE_H_ 1
 
 void
 memcache_init(void);
@@ -62,10 +62,10 @@ memcache_remove(app_pc start, app_pc end);
 bool
 memcache_query_memory(const byte *pc, OUT dr_mem_info_t *out_info);
 
-#    if defined(DEBUG) && defined(INTERNAL)
+#if defined(DEBUG) && defined(INTERNAL)
 void
 memcache_print(file_t outf, const char *prefix);
-#    endif
+#endif
 
 void
 memcache_handle_mmap(dcontext_t *dcontext, app_pc base, size_t size, uint prot,
--- a/core/unix/memcache.h
+++ b/core/unix/memcache.h
@@ -31,7 +31,7 @@
  */
 
 #ifndef _MEMCACHE_H_
-#    define _MEMCACHE_ 1
+#    define _MEMCACHE_H_ 1
 
 void
 memcache_init(void);


core/unix/memcache.h
```diff
@@ -31,7 +31,7 @@
  */
 
 #ifndef _MEMCACHE_H_
-#    define _MEMCACHE_ 1
+#define _MEMCACHE_H_ 1
```
derekbruening, line 34: This un-indents the #if on line 65 as well: please fix to avoid a future change running afoul of clang-format (the clang-format-diff in the CI doesn't see far enough down to fail this PR).

kiri11: I did not 100% understand whether you want me to revert unindent or do it on line 65 as well. So I decided to return it how it was before.

kiri11: Hm, now clang-format fails, so it's worse?

derekbruening: What we want is for this file to be formatted with clang-format.  If I put in the fix `s/_MEMCACHE_/_MEMCACHE_H_/` in my editor, clang-format then recognizes the header guard and not only unindents line 34 but unindents lines 65 and 68 (which is what you would expect).  If we had clang-format-diff see far enough down in the file, your PR would have failed because you failed to unindent lines 65 and 68.  Failing to fix the file here is going to cause pain for someone down the line who edits the middle of the file and won't have their whole-file clang-format be able to match the local clang-format-diff run on the PR's.  I'm guessing you don't have clang-format set up on your editor: I would suggest doing so.

kiri11: Thank you for the detailed explanation! I am installing clang-format, will submit the change shortly. 
Please let me know whether I should squash/rebase changes into just two commits (fix itself + clang-format), or keep the history as-is.

kiri11: I decided to change the history to make it cleaner before the merge.

derekbruening: > Please let me know whether I should squash/rebase changes into just two commits (fix itself + clang-format), or keep the history as-is.
> I decided to change the history to make it cleaner before the merge.

Please do not change the history.  In general, do not force-push to a PR branch.  The final squash-and-merge collapses all the commits and uses the PR description as the commit message, so there is zero benefit to changing the history, and there are many downsides: it breaks the review history and review comment threads get untethered and now the reivewer can't click on "just show changes since last review": it makes everything painful.  (Requests to never force-push are in multiple places in our docs:https://dynamorio.org/page_code_reviews.html#sec_code_review_non_member, https://dynamorio.org/page_code_reviews.html#autotoc_md114)

kiri11: Sorry, will not do this again. This is my first contribution, and I missed some things. It's hard to read and remember all the docs at once. Now I know how to do this properly, thanks a lot for your patience and explanations! Hopefully this change isn't that complicated to follow the review history.

derekbruening: No problem, this PR is small enough it had no impact, and different projects do have different settings (not all squash on merging) so it can be confusing: just requesting to avoid force-pushes for the future.  Thanks again for sending us the fix!

Issue #6354, opened 2023-10-10
seeing more than one cpu id on one output stream with MAP_TO_RECORDED_OUTPUT


prasun3:
**Describe the bug**

See https://groups.google.com/g/DynamoRIO-Users/c/Dsf4As64r7o

In some of my tests, I am seeing more than one cpu id on one output stream with MAP_TO_RECORDED_OUTPUT.

The value returned by stream->get_output_cpuid matches the initial CPU_ID marker values, but at some point on some streams, we see other CPU_ID values as shown below.

Example:

```
Output #5 processing CPU #7                   (from get_stream_cpuid)
Output #5 saw CPU_ID marker: 7
Output #5 saw CPU_ID marker: 7
Output #5 saw CPU_ID marker: 7
...
Output #5 saw CPU_ID marker: 7
Output #5 saw CPU_ID marker: 7
Output #5 saw CPU_ID marker: 5
```

**To Reproduce**
1. Collect a trace with `drrun -t drcachesim -offline`
2. Use the scheduler infrastructure with the MAP_TO_RECORDED_OUTPUT option
3. We see multiple CPU IDs on the same output stream which is not expected

Please also answer these questions:
 - What happens when you run without any client?
   - Not applicable
 - What happens when you run with debug build ("-debug" flag to drrun/drconfig/drinject)?
   - No difference

**Expected behavior**
With the MAP_TO_RECORDED_OUTPUT option, each output stream maps to exactly one CPU

**Screenshots or Pasted Text**

**Versions**
 - What version of DynamoRIO are you using?
   - Commit 910ccb25a66175bcba3f349cb7a43c8da7d9f7c3
 - Does the latest build from https://github.com/DynamoRIO/dynamorio/releases solve the problem?
   - No
- What operating system version are you running on? ("Windows 10" is *not* sufficient: give the release number.)
   - Ubuntu 18
- Is your application 32-bit or 64-bit?
   - 64-bit

**Additional context**

derekbruening:
It looks like this is the culprit, this thread with zero instructions:
```
$ bin64/drrun -t drcachesim -simulator_type view -only_thread 85300 -indir ../src/clients/drcachesim/tests/drmemtrace.schedtest.x64
Output format:
<--record#-> <--instr#->: <---tid---> <record details>
------------------------------------------------------------
           1           0:       85300 <marker: version 6>
           2           0:       85300 <marker: filetype 0xe40>
[...]

PR #6355, opened 2023-10-10
i#6354 test: Verify one CPU ID per stream with MAP_TO_RECORDED

prasun3:
Added check to verify that output stream's CPU ID matches the CPU ID markers seen in the trace.

Issue: #6354


clients/drcachesim/CMakeLists.txt
```diff
@@ -391,6 +391,13 @@ if (X86 AND X64 AND ZIP_FOUND)
   add_test(NAME tool.scheduler_launcher
     COMMAND scheduler_launcher -trace_dir ${trace_dir})
   set_tests_properties(tool.scheduler_launcher PROPERTIES TIMEOUT ${test_seconds})
+
+  set(trace_dir_2
+    "${PROJECT_SOURCE_DIR}/clients/drcachesim/tests/drmemtrace.schedtest.x64")
```
derekbruening, line 396: Since the issue seems to be that there is an empty thread in here maybe we don't want to check in this test data but rather prevent future empty traces.  Maybe add an invariant check that a (non-filtered) trace has at least one instruction, too.

prasun3:
This issue is mostly resolved with #6356 but I still ran into a (very rare) failure. I have added a test trace where the threads have non-zero instructions, but we still see this failure.

derekbruening:
> This issue is mostly resolved with #6356 but I still ran into a (very rare) failure. I have added a test trace where the threads have non-zero instructions, but we still see this failure.

I put the analysis into the bug proper: https://github.com/DynamoRIO/dynamorio/issues/6354#issuecomment-1760255042

Issue #3346, opened 2019-01-15
CRASH and RANK ORDER VIOLATION with -exit_after_tracing


derekbruening:
When I was gathering the new test trace
clients/drcachesim/tests/drmemtrace.threadsig.x64.tracedir/drmemtrace.threadsig.10506.7343.trace.gz
I hit problems when I used -exit_after_tracing but only with certain values.
This only happens with a large -trace_after_instrs: so if we start tracing
near the end.

```
$ rm -rf drmemtrace.*.dir; gdb --args bin64/drrun -msgbox_mask 12 -t drcachesim -offline -trace_after_instrs 2180K -exit_after_tracing 120K -- ~/dr/test/threadsig 8 2000

add-symbol-file '/home/bruening/dr/git/build_x64_dbg_tests/bin64/../clients/lib64/debug/libdrmemtrace.so' 0x00007fa4afa97c10
add-symbol-file '/home/bruening/dr/git/build_x64_dbg_tests/ext/lib64/debug/libdrsyms.so' 0x00007fa4afcd0260
add-symbol-file '/home/bruening/dr/git/build_x64_dbg_tests/ext/lib64/debug/libdrwrap.so' 0x00007fa4aff18020
add-symbol-file '/home/bruening/dr/git/build_x64_dbg_tests/ext/lib64/debug/libdrmgr.so' 0x00007fa4b0126570
add-symbol-file '/home/bruening/dr/git/build_x64_dbg_tests/ext/lib64/debug/libdrutil.so' 0x00007fa4b03316a0
add-symbol-file '/home/bruening/dr/git/build_x64_dbg_tests/ext/lib64/debug/libdrcovlib.so' 0x00007fa4b0537dc0
add-symbol-file '/home/bruening/dr/git/build_x64_dbg_tests/ext/lib64/debug/libdrx.so' 0x00007fa4b07431f0
add-symbol-file '/home/bruening/dr/git/build_x64_dbg_tests/ext/lib64/debug/libdrreg.so' 0x00007fa4b094cb10
add-symbol-file '/lib/x86_64-linux-gnu/libgcc_s.so.1' 0x00007fa5329f4ac0
add-symbol-file '/lib64/ld-linux-x86-64.so.2' 0x00007fa532c0baa0
add-symbol-file '/lib/x86_64-linux-gnu/libc.so.6' 0x00007fa532e51910
[...]

PR #6356, opened 2023-10-10
i#6354: Do not output zero-instruction (unfiltered) threads

derekbruening:
Tightens the normal-mode output rule to discard a thread with no instructions (keeping the no-data check for i-filtered modes) and expands the empty-thread behavior to include regular exit runs and not just detach.

Adds the scheduler_launcher CPU check from PR #6355.

It is not easy to create a test with a thread that reliably executes zero instructions yet still outputs some data.  I added an invariant check for a zero-instruction thread and confirmed it fires on the trace in PR #6355:

```
$ bin64/drrun -t drcachesim -simulator_type invariant_checker -indir ~/tmp/drmemtrace.schedtest.x64
Trace invariant failure in T85300 at ref # 12 (0 instrs since timestamp 13341391214912820): An unfiltered thread should have at least 1 instruction
```

Issue: #6354

### DIFF
--- a/clients/drcachesim/tests/scheduler_launcher.cpp
+++ b/clients/drcachesim/tests/scheduler_launcher.cpp
@@ -231,10 +231,11 @@ simulate_core(int ordinal, scheduler_t::stream_t *stream, const scheduler_t &sch
                 thread_sequence += 'A' + static_cast<char>(input % 26);
                 cur_segment_instrs = 0;
             }
-        } else if (record.marker.type == dynamorio::drmemtrace::TRACE_TYPE_MARKER) {
+        }
+#ifdef HAS_ZIP
+        else if (record.marker.type == dynamorio::drmemtrace::TRACE_TYPE_MARKER) {
             if (record.marker.marker_type ==
                 dynamorio::drmemtrace::TRACE_MARKER_TYPE_CPU_ID) {
-#ifdef HAS_ZIP
                 if (!op_cpu_schedule_file.get_value().empty()) {
                     int cpu = (int)record.marker.marker_value;
                     int output_cpuid = stream->get_output_cpuid();
@@ -244,9 +245,9 @@ simulate_core(int ordinal, scheduler_t::stream_t *stream, const scheduler_t &sch
                                     cpu, ordinal, output_cpuid);
                     }
                 }
-#endif
             }
         }
+#endif
     }
 }
 
--- a/clients/drcachesim/tests/scheduler_launcher.cpp
+++ b/clients/drcachesim/tests/scheduler_launcher.cpp
@@ -231,6 +231,21 @@ simulate_core(int ordinal, scheduler_t::stream_t *stream, const scheduler_t &sch
                 thread_sequence += 'A' + static_cast<char>(input % 26);
                 cur_segment_instrs = 0;
             }
+        } else if (record.marker.type == dynamorio::drmemtrace::TRACE_TYPE_MARKER) {
+            if (record.marker.marker_type ==
+                dynamorio::drmemtrace::TRACE_MARKER_TYPE_CPU_ID) {
+#ifdef HAS_ZIP
+                if (!op_cpu_schedule_file.get_value().empty()) {
+                    int cpu = (int)record.marker.marker_value;
+                    int output_cpuid = stream->get_output_cpuid();
+                    if (cpu != output_cpuid) {
+                        FATAL_ERROR("CPU marker %d on core #%d differs from output "
+                                    "stream CPU ID %d\n",
+                                    cpu, ordinal, output_cpuid);
+                    }
+                }
+#endif
+            }
         }
     }
 }
--- a/clients/drcachesim/tracer/output.cpp
+++ b/clients/drcachesim/tracer/output.cpp
@@ -111,6 +111,22 @@ local_instr_count_threshold(uint64 trace_for_instrs)
     }
 }
 
+static bool
+buffer_contains_nontrivial_data(per_thread_t *data)
+{
+    if (op_L0I_filter.get_value()) {
+        return BUF_PTR(data->seg_base) - data->buf_base >
+            static_cast<ssize_t>(data->init_header_size + buf_hdr_slots_size);
+    }
+    byte *buf_ptr = BUF_PTR(data->seg_base);
+    for (byte *mem_ref = data->buf_base + buf_hdr_slots_size; mem_ref < buf_ptr;
+         mem_ref += instru->sizeof_entry()) {
+        if (instru->get_instr_count(mem_ref) > 0)
+            return true;
+    }
+    return false;
+}
+
 // Returns whether we've reached the end of this tracing window.
 static bool
 count_traced_instrs(void *drcontext, uintptr_t toadd, uint64 trace_for_instrs)
@@ -1377,9 +1393,7 @@ exit_thread_io(void *drcontext)
         // each thread.  However, we omit threads that did nothing the entire time
         // we were attached.
         (!has_tracing_windows() && align_attach_detach_endpoints() &&
-         (data->bytes_written > 0 ||
-          BUF_PTR(data->seg_base) - data->buf_base >
-              static_cast<ssize_t>(data->init_header_size + buf_hdr_slots_size)))) {
+         (data->bytes_written > 0 || buffer_contains_nontrivial_data(data)))) {
         BUF_PTR(data->seg_base) += instru->append_thread_exit(
             BUF_PTR(data->seg_base), dr_get_thread_id(drcontext));
         process_and_output_buffer(drcontext,
--- a/clients/drcachesim/tracer/output.cpp
+++ b/clients/drcachesim/tracer/output.cpp
@@ -1378,22 +1378,28 @@ exit_thread_io(void *drcontext)
     }
 #endif
 
-    // Append a thread exit marker and output remaining records for this thread
-    // if we are still in a tracing mode or the thread has data from a prior window
-    // that it never wrote out.
-    if ((has_tracing_windows() &&
-         // If non-split we always want to append a thread exit marker as there
-         // wouldn't be on otherwise (split has one at the end of each window file).
-         (!op_split_windows.get_value() ||
-          // If split, we only need to write if we have data from a prior window.
-          (get_local_window(data) < tracing_window.load(std::memory_order_acquire) &&
-           !is_new_window_buffer_empty(data)))) ||
-        // Throw out empty threads even in a tracing mode.
-        ((data->bytes_written > 0 || buffer_contains_nontrivial_data(data)) &&
-         (is_in_tracing_mode(tracing_mode.load(std::memory_order_acquire)) ||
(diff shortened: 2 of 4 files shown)


derekbruening:
Failure is coherence test invalid type 23 before a bundle #4167

prasun3:
I can reproduce this with the cmd line below. It happens when there is a long running test and we use `-exit_after_tracing`. The same issue occurs with `-max_trace_size` followed by manually killing the application. In both cases, I see an exception, so the exit syscall is probably not executed

```
build/bin64/drrun -t drcachesim -trace_after_instrs 9000M -exit_after_tracing 5M --offline -- ./threadsig 2 1000000000
```


```
<Stopping application ... threadsig (123613)>
<Application ... threadsig (123613).  DynamoRIO Cache Simulator Tracer internal crash at PC 0x00007efdcac8f603.  Please report this at http://dynamorio.org/issues.  Program aborted.
Received SIGSEGV at pc 0x00007efdcac8f603 in thread 123614
Base: 0x00007efdcb26b000
Registers:eax=0x0000000000000240 ebx=0x00007efb8716dd60 ecx=0x00007efd87081980 edx=0x7a2ab66beb93f336
        esi=0x0000000000000000 edi=0x00007efd872cf320 esp=0x00007efb872cf290 ebp=0x00007efdcb03c708
        r8 =0x0000000000000000 r9 =0x00007efd872cf320 r10=0x0000000000000003 r11=0x0000000000000246
        r12=0x00007efb8716db30 r13=0x00007efd872c90e0 r14=0x00007efb8716db40 r15=0x0000000000000072
        eflags=0x0000000000010246
version 10.0.19640, custom build
```

With `-max_trace_size` followed by `kill`

```
build/bin64/drrun -t drcachesim -trace_after_instrs 9000M -max_trace_size 5M --offline -- ./threadsig 2 1000000000
```
```
<Application ... threadsig (123671).  Application exception at PC 0x00007fd58a647d2b.
Signal 15 delivered to application as default action.
Callstack:
        0x00007fd58a647d2b   </lib/x86_64-linux-gnu/libpthread-2.27.so+0x8d2b>
        0x00007fd587eda0f0
        0xc48348c2440f48ff
>
<Stopping application ... threadsig (123671)>
```

prasun3:
> @prasun3 could you check whether with this branch in place you no longer get that no-instruction thread? I'm not sure how to create a test to recreate that: how to have a thread reliably execute the post-futex yet not execute any instructions.

The same behaviour is seen even with this branch.

derekbruening:
> <Application ... threadsig (123613).  DynamoRIO Cache Simulator Tracer internal crash at PC 0x00007efdcac8f603.  Please report this at http://dynamorio.org/issues.  Program aborted.

But the tool crashed!  All bets are off.  The trace might be corrupted or not finalized properly.  Probably raw2trace or invariant_checker will fail/complain if so and apparently they did not for you so maybe it's ok but I would be wary of it.  That crash should be fixed: was an issue filed on it?

derekbruening:
> The same behaviour is seen even with this branch.

Looks like -exit_after_tracing doesn't turn off tracing mode.  We always detach which turns it off...probably we should have all exit methods do so, or otherwise enable the drop-empty-thread code.  Will update the PR.

derekbruening:
> > The same behaviour is seen even with this branch.
> 
> Looks like -exit_after_tracing doesn't turn off tracing mode. We always detach which turns it off...probably we should have all exit methods do so, or otherwise enable the drop-empty-thread code. Will update the PR.

Updated.  The empty thread should no longer exist.

prasun3:
I'm not seeing zero-inst traces with this branch and scheduler_launcher also seems to pass on these threads.

Is detach only available with source instrumentation?

I'll file an issue about the crash with `-exit_after_tracing`.

prasun3:
I don't fully understand the impact of the changes in output.cpp - could maybe someone more familiar do the code review?

prasun3:
> I'll file an issue about the crash with `-exit_after_tracing`.

Looks like we discussed it here https://github.com/DynamoRIO/dynamorio/issues/3346#issuecomment-1558818317

derekbruening:
Still not sure how to make a reliable test.  I tried the -exit_after_tracing command and it did result in one empty thread but it had just one cpuid: it did not have a 2nd and so would not have triggered the cpuid mismatch.  We may have to just go with the manual testing from @prasun3.

clients/drcachesim/tests/scheduler_launcher.cpp
```diff
@@ -231,6 +231,21 @@ simulate_core(int ordinal, scheduler_t::stream_t *stream, const scheduler_t &sch
                 thread_sequence += 'A' + static_cast<char>(input % 26);
                 cur_segment_instrs = 0;
             }
+        } else if (record.marker.type == dynamorio::drmemtrace::TRACE_TYPE_MARKER) {
+            if (record.marker.marker_type ==
+                dynamorio::drmemtrace::TRACE_MARKER_TYPE_CPU_ID) {
+#ifdef HAS_ZIP
```
abhinav92003, line 237: nit: the whole else-if block could be conditionally compiled.

clients/drcachesim/tracer/output.cpp
```diff
@@ -1365,21 +1381,19 @@ exit_thread_io(void *drcontext)
...
            !is_new_window_buffer_empty(data)))) ||
-        // For attach we switch to BBDUP_MODE_NOP but still need to finalize
-        // each thread.  However, we omit threads that did nothing the entire time
-        // we were attached.
-        (!has_tracing_windows() && align_attach_detach_endpoints() &&
-         (data->bytes_written > 0 ||
-          BUF_PTR(data->seg_base) - data->buf_base >
-              static_cast<ssize_t>(data->init_header_size + buf_hdr_slots_size)))) {
+        // Throw out empty threads even in a tracing mode.
+        ((data->bytes_written > 0 || buffer_contains_nontrivial_data(data)) &&
```
abhinav92003, line 1392: Curious why we need both data->bytes_written > 0 and buffer_contains_nontrivial_data(data). The latter seems a stricter condition; does it happen that one is true without the other?

clients/drcachesim/tracer/output.cpp
```diff
@@ -1365,21 +1381,19 @@ exit_thread_io(void *drcontext)
     // Append a thread exit marker and output remaining records for this thread
     // if we are still in a tracing mode or the thread has data from a prior window
     // that it never wrote out.
-    if (is_in_tracing_mode(tracing_mode.load(std::memory_order_acquire)) ||
-        (has_tracing_windows() &&
+    if ((has_tracing_windows() &&
```
abhinav92003, line 1384: It'd be nice to split this huge condition into multiple booleans, each with its own relevant comments atop it.

abhinav92003 (approved):

clients/drcachesim/tests/scheduler_launcher.cpp
derekbruening: Done.

clients/drcachesim/tracer/output.cpp
derekbruening: They are quite different: bytes_written tells us whether prior buffer contents have already been written out (if so, we definitely want to close out this thread with an exit record), while the other check is about the just the current buffer.  We only want to ignore a thread that both has a trivially-empty current buffer and never wrote anything before.

Added a comment.

clients/drcachesim/tracer/output.cpp
derekbruening: Done.

prasun3:
> Still not sure how to make a reliable test. I tried the -exit_after_tracing command and it did result in one empty thread but it had just one cpuid: it did not have a 2nd and so would not have triggered the cpuid mismatch. We may have to just go with the manual testing from @prasun3.

This is the test that I ran

```
build/bin64/drrun -pidfile pid.txt -t drcachesim -trace_after_instrs 9000M -exit_after_tracing 5M --offline -- ./threadsig 2 1000000000 
build/clients/bin64/drraw2trace -indir drmemtrace.threadsig.$(cat pid.txt).*
build/clients/bin64/scheduler_launcher -trace_dir drmemtrace.threadsig.$(cat pid.txt).*/trace -cpu_schedule_file drmemtrace.threadsig.$(cat pid.txt).*/trace/cpu_schedule.bin.zip
```

prasun3:
@derekbruening I am seeing zero-insr threads with L0_filter. Is this because there are two INSTRUCTION_COUNT markers?

Cmd line
```
build/bin64/drrun -pidfile pid.txt -t drcachesim -trace_after_instrs 9000M \
    -L0_filter \
    -exit_after_tracing 100K --offline -- ./threadsig 2 1000000000
```

Raw file entries
```
EXTENDED: header
THREAD
PID
EXTENDED: marker: TRACE_MARKER_TYPE_CACHE_LINE_SIZE
EXTENDED: marker: TRACE_MARKER_TYPE_PAGE_SIZE
TIMESTAMP
EXTENDED: marker: TRACE_MARKER_TYPE_CPU_ID
EXTENDED: marker: TRACE_MARKER_TYPE_INSTRUCTION_COUNT
EXTENDED: marker: TRACE_MARKER_TYPE_INSTRUCTION_COUNT
EXTENDED: footer
```

Trace
```
Output format:
<--record#-> <--instr#->: <---tid---> <record details>
------------------------------------------------------------
           1           0:      210474 <marker: version 6>
           2           0:      210474 <marker: filetype 0xbc2>
           3           0:      210474 <marker: cache line size 64>
           4           0:      210474 <marker: chunk instruction count 10000000>
           5           0:      210474 <marker: page size 4096>
           6           0:      210474 <marker: timestamp 13342269712523323>
           7           0:      210474 <marker: tid 210474 on core 20696>
           8           0:      210474 <marker: instruction count 0>
           9           0:      210474 <marker: instruction count 0>
          10           0:      210474 <thread 210474 exited>
View tool results:
              0 : total instructions
```

derekbruening:
For an instruction filter we can't really have the same standard as there could be zero instruction records but still some data records.  I guess the check could look for data records too?  But how is an instr-filtered trace being fed to the scheduler: it must be a warmup with no post-warmup?  I had assumed the filtered part would not be asked to be scheduled.

prasun3:
I originally saw the scheduling issue with a bimodal trace, and since the issue was root-caused to empty traces, the example above reproduces empty traces with filtering enabled.

I thought running bimodal traces through the scheduler would be okay since they have cpu id markers, but sounds like that is not the case since it relies on instructions being present?

derekbruening:
> I thought running bimodal traces through the scheduler would be okay since they have cpu id markers, but sounds like that is not the case since it relies on instructions being present?

We didn't really think about this use case.  The basics would work but preempting by instruction count would not (so the simulator/tool would have to pass in a time value) and skipping for regions of interest or recorded schedules would not. File a separate issue on scheduler support for i-filtered traces?  Maybe the rep string proposed solution Proposal #E at https://github.com/DynamoRIO/dynamorio/issues/6354#issuecomment-1779737101 would also solve this.

prasun3:
I am interested in just generating per-core streams so it is okay for me even if those other features you listed don't work. But I can file an issue in case it is useful in the future or to others.

PR #6359, opened 2023-10-11
i#3544 RV64: Implement insert_exit_stub_other_flags

ksco:
The implementation of this function is roughly the same as AArch64, except for some additional handling of C extension instructions.

Issue: #3544

PR #6360, opened 2023-10-11
i#3544 RV64: Implement patch_stub and stub_is_patched

ksco:
This patch implements patch_stub() and stub_is_patched(). Similar to AArch64, our RISC-V port supports near-fragment linking and far-fragment linking. For the former, we use `J offset` (short for `JAL x0, offset`) to link the basic blocks. For the latter, we update target_pc_slot to reuse the original indirect jump.

Issue: #3544


ksco:
Note that in order to compile while keeping the patch size small, the `get_target_pc_slot()` function definition is duplicated with #6359, I'll remove one of them once the other gets merged.

abhinav92003 (approved):

PR #6362, opened 2023-10-12
i3346 test: Added a test that reproduces the crash

prasun3:
Issue: #3346


prasun3:
Reproduced the failure in x86-64 and x86-32

Issue #6364, opened 2023-10-12
Invariant checker reports Cpu/Serial schedule entry does not match trace.


ivankyluk:
The current implementation of comparator for schedule entries compares the timestamp, CPU, and tread ID, implicitly assuming entries with the different start_instruction will have different timestamp.

The resolution of the timestamp is us, if the start_instructions are close, they could have the same timestamp. And since the start_instruction is not used by the comparator, the sorting result is indeterministic. 

When the schedule entry is compared against the trace with the same timestamp, but different start_instruction, it causes the Cpu/Serial schedule entry does not match trace error.

The fix is to add start_instruction to the comparator.

PR #6366, opened 2023-10-12
i#6364: add start_instruction to schedule entry comparator.

ivankyluk:
We have captured traces from the same thread with the same CPU and timestamp, their start_instructions are tens of instructions apart.

The current comparators in raw2trace and the invariant checker do not take start_instruction into account. And it leads to indeterministic sorting results. It causes the invariant checker to report cpu/serial entry does not match trace error when raw2trace and the invariant checker sort the entries differently.

The change is to compare the start_instruction when two schedule entries have the same CPU, thread ID, and timestamp.
And add a unit test to cover the case where two schedule entries have the same CPU, thread ID, and timestamp, but different start_instruction.

Fixes: #6364

### DIFF
--- a/clients/drcachesim/tests/invariant_checker_test.cpp
+++ b/clients/drcachesim/tests/invariant_checker_test.cpp
@@ -1615,6 +1615,8 @@ check_schedule_file()
     sched.emplace_back(TID_BASE + 1, TIMESTAMP_BASE, CPU_BASE + 2, 0);
     sched.emplace_back(TID_BASE, TIMESTAMP_BASE + 1, CPU_BASE + 1, 2);
     sched.emplace_back(TID_BASE + 1, TIMESTAMP_BASE + 2, CPU_BASE, 1);
+    // Include records with the same thread ID, timestamp, and CPU, but
+    // different start_instruction is being used for comparison.
     sched.emplace_back(TID_BASE + 2, TIMESTAMP_BASE + 3, CPU_BASE + 2, 3);
     sched.emplace_back(TID_BASE + 2, TIMESTAMP_BASE + 3, CPU_BASE + 2, 4);
     {
@@ -1650,6 +1652,8 @@ check_schedule_file()
             gen_marker(TID_BASE + 2, TRACE_MARKER_TYPE_TIMESTAMP, TIMESTAMP_BASE + 3),
             gen_marker(TID_BASE + 2, TRACE_MARKER_TYPE_CPU_ID, CPU_BASE + 2),
             gen_instr(TID_BASE + 2, 4),
+            // Markers for the second schedule entry with the same thread ID,
+            // timestamp, and CPU as the previous one with a different start_instruction.
             gen_marker(TID_BASE + 2, TRACE_MARKER_TYPE_TIMESTAMP, TIMESTAMP_BASE + 3),
             gen_marker(TID_BASE + 2, TRACE_MARKER_TYPE_CPU_ID, CPU_BASE + 2),
         };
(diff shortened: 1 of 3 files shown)


clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -1616,6 +1616,7 @@ check_schedule_file()
     sched.emplace_back(TID_BASE, TIMESTAMP_BASE + 1, CPU_BASE + 1, 2);
     sched.emplace_back(TID_BASE + 1, TIMESTAMP_BASE + 2, CPU_BASE, 1);
     sched.emplace_back(TID_BASE + 2, TIMESTAMP_BASE + 3, CPU_BASE + 2, 3);
+    sched.emplace_back(TID_BASE + 2, TIMESTAMP_BASE + 3, CPU_BASE + 2, 4);
```
derekbruening, line 1619: Add to the PR description that a test is added which fails without the fix.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -1649,6 +1650,8 @@ check_schedule_file()
             gen_marker(TID_BASE + 2, TRACE_MARKER_TYPE_TIMESTAMP, TIMESTAMP_BASE + 3),
             gen_marker(TID_BASE + 2, TRACE_MARKER_TYPE_CPU_ID, CPU_BASE + 2),
             gen_instr(TID_BASE + 2, 4),
+            gen_marker(TID_BASE + 2, TRACE_MARKER_TYPE_TIMESTAMP, TIMESTAMP_BASE + 3),
```
derekbruening, line 1653: Add a comment explaining what this is testing.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -1023,8 +1025,12 @@ invariant_checker_t::check_schedule_data(per_shard_t *global)
     for (auto &keyval : cpu2sched_file) {
         std::sort(keyval.second.begin(), keyval.second.end(),
                   [](const schedule_entry_t &l, const schedule_entry_t &r) {
-                      if (l.timestamp == r.timestamp)
-                          return l.thread < r.thread;
+                      if (l.timestamp == r.timestamp) {
```
derekbruening, line 1028: Hmm so this one wasn't updated before.  Best to make it use the same code (the != sequence) as the other as otherwise it looks like it's trying to be different.  Use the same code: you can use the same lambda, right?  That seems worth the wasted check for cpu.

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -1310,7 +1315,9 @@ raw2trace_t::aggregate_and_write_schedule_files()
     for (auto &keyval : cpu2sched) {
         std::sort(keyval.second.begin(), keyval.second.end(),
                   [](const schedule_entry_t &l, const schedule_entry_t &r) {
-                      return l.timestamp < r.timestamp;
+                      if (l.timestamp != r.timestamp)
+                          return l.timestamp < r.timestamp;
+                      return l.start_instruction < r.start_instruction;
```
derekbruening, line 1320: I would again use the same lambda to make it clear it's the same comparison and avoid further errors where someone updates one and not the other (which apparently happened recently right?)

Isn't this wrong since it's not checking the thread?  Yet another reason to use the same lambda.

derekbruening (approved):

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: Done.

clients/drcachesim/tools/invariant_checker.cpp
ivankyluk: Done.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: Done.

clients/drcachesim/tracer/raw2trace.cpp
ivankyluk: Done.

PR #6370, opened 2023-10-16
i#5036 A64 scatter/gather: Fix drcachesim offline tracing

jackgallagher-arm:
A couple of places in the drcachesim tracer there was code that handles scatter/gather instructions that was guarded by #ifdef X86 which meant offline traces that include scatter/gather instructions were producing errors on AArch64.

For example running the aarch64 scattergather test app:
```
$bin64/drrun -t drcachesim -offline -- suite/tests/bin/client.drx-scattergather
$bin64/drrun -t drcachesim -indir drmemtrace.client.drx-scattergather.662655.1769.dir
```

produced the error:
```
ERROR: failed to initialize analyzer: raw2trace failed: Failed to process file for thread 662655: memref entry found outside of bb
```

This commit enables the code on AArch64 as well and fixes the error.

Issue: #5036

### DIFF
--- a/clients/drcachesim/tracer/instru_offline.cpp
+++ b/clients/drcachesim/tracer/instru_offline.cpp
@@ -955,8 +955,7 @@ offline_instru_t::identify_elidable_addresses(void *drcontext, instrlist_t *ilis
         // view by expanding the instr in raw2trace (e.g. using
         // drx_expand_scatter_gather) when building the ilist.
         if (drutil_instr_is_stringop_loop(instr)
-            // TODO i#5036: Scatter/gather support incomplete on AArch64.
-            IF_X86(|| instr_is_scatter(instr) || instr_is_gather(instr))) {
+            IF_X86_OR_AARCH64(|| instr_is_scatter(instr) || instr_is_gather(instr))) {
             return;
         }
         if (drmgr_is_emulation_start(instr) || drmgr_is_emulation_end(instr)) {
--- a/clients/drcachesim/tracer/instru_offline.cpp
+++ b/clients/drcachesim/tracer/instru_offline.cpp
@@ -955,7 +955,7 @@ offline_instru_t::identify_elidable_addresses(void *drcontext, instrlist_t *ilis
         // view by expanding the instr in raw2trace (e.g. using
         // drx_expand_scatter_gather) when building the ilist.
         if (drutil_instr_is_stringop_loop(instr)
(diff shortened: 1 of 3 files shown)


clients/drcachesim/tracer/instru_offline.cpp
```diff
@@ -955,8 +955,7 @@ offline_instru_t::identify_elidable_addresses(void *drcontext, instrlist_t *ilis
         // view by expanding the instr in raw2trace (e.g. using
         // drx_expand_scatter_gather) when building the ilist.
         if (drutil_instr_is_stringop_loop(instr)
-            // TODO i#5036: Scatter/gather support incomplete on AArch64.
-            IF_X86(|| instr_is_scatter(instr) || instr_is_gather(instr))) {
+                IF_X86_OR_AARCH64(|| instr_is_scatter(instr) || instr_is_gather(instr))) {
```
abhinav92003, line 958: All archs define instr_is_scatter and instr_is_gather, and return default false if not implemented. How about we just remove the conditional compilation here and below?

abhinav92003 (approved):

clients/drcachesim/tracer/instru_offline.cpp
```diff
@@ -955,8 +955,7 @@ offline_instru_t::identify_elidable_addresses(void *drcontext, instrlist_t *ilis
         // view by expanding the instr in raw2trace (e.g. using
```
abhinav92003, line 955: I assume this is in preparation for a scatter-gather drmemtrace basic counts test? We could add that in this PR too if possible.

clients/drcachesim/tracer/instru_offline.cpp
jackgallagher-arm: The riscv64 implementations have `ASSERT_NOT_IMPLEMENTED(false)` in so wouldn't taking out the conditional compilation stop drcachesim working on riscv64?

clients/drcachesim/tracer/instru_offline.cpp
abhinav92003: Okay. I'm fine with submitting just this one for now.

clients/drcachesim/tracer/instru_offline.cpp
abhinav92003: Hmm, right. Okay, let's keep it the way it is.

Issue #1672, opened 2015-04-09
port annotation infrastructure to ARM


derekbruening:
Split from #1551

derekbruening:
We also need to port to AArch64 which we could include here as the implementations may be quite similar.

derekbruening:
There is further information here along with a potential contributor: https://groups.google.com/forum/#!topic/dynamorio-users/0HZGzngZmZ0

derekbruening:
@byron-hawkins made a nice parameterizable multi-threaded test app for annotations `client.annotation-concurrency` and we ended up using it as a multi-threaded non-trivially-sized test app for many drcachesim tests. 
 However, it contains annotations, so the lack of annotation support on AArchXX now blocks running many drcachesim tests.  Ideally we would just implement annotation support on AArchXX, but we could potentially hack the test to make the annotation macros resolve to empty code to make progress.

PR #6371, opened 2023-10-16
i#5365: Build core unit tests with SVE enabled

philramsey-arm:
Build most core tests with SVE flags and high optimisation (-O3), if building on a AARCH64 SVE machine.

Tests which fail when built with -O3 are not included.

Add some error checking to a few tests to allow the -O3 build and update template (expected output) files as necessary.

Issue #6429 raised to cover making the removal of optimization flags more granular.

Issue: #5365


AssadHashmi:
@derekbruening I'm happy to approve this change. Is there anything you would like to call out?

CMakeLists.txt
```diff
@@ -1612,6 +1612,10 @@ if (BUILD_TESTS)
   include(CTest)
 endif (BUILD_TESTS)
 
+if (AARCH64)
+  option(BUILD_TESTS_SVE "Build tests with SVE enabled and high optimization (-O3)" OFF)
```
derekbruening, line 1616: The PR description doesn't say what the long-term plan is: please add something there.  Is the plan that some outer cmake test suite code is going to figure out whether to turn this on?  Is the plan that this will be on by default soon once some blocking issues are fixed?  Or will this always be some off-by-default option only used manually?

derekbruening, line 1616: I would expect the tests to be built at the same optimization/debug level as the rest of the project: should the tests be O3 for a debug build?  I would not expect that.

suite/tests/CMakeLists.txt
```diff
@@ -553,6 +553,24 @@ function(add_exe test source)
       "${CMAKE_GENERATOR}" MATCHES "Visual Studio")
     add_dependencies(${test} ${gen_asm_tgt})
   endif ()
+  if (BUILD_TESTS_SVE)
+    target_compile_options(${test} PRIVATE
+      "-march=armv8.4-a+crypto+rcpc+sha3+sm4+sve+rng+ssbs+nodotprod")
+
+    # Some tests give very obscure build errors when built with "-O3"
+    # FIXME i#5365: fix build issues due to optimisation, unless no
+    # optimisation is a requirement for a test case.
```
derekbruening, line 562: I am a little confused: aren't these tests already built -O3 in a release build (see above about expecting tests to match the rest of the project), which is a build setting in the GA CI suite?  So we should already be exercising -O3, so is it the SVE flags causing the problems you are seeing?

Despite thinking we already exercise -O3, it is not surprising to see some tests fail with high optimization as they often construct particular patterns using simple conditions and often have unused results or potentially-dead code.  This is esp true for tool tests like Dr. Memory (in a different repository) which wants weird and buggy code patterns and often has to fight the compiler.  You could image drcov or other tools having similar problems.  You could imagine drmemtrace tests having load/store references for testing without significant use and the compiler throws them away at high optimization.  But again I thought we already built these -O3.

CMakeLists.txt
AssadHashmi: This patch is preparation work to defend SVE support because we intend to start precommit testing on SVE hardware in the near future. We will enable the SVE tests in a GitHub Runner hosted by the SVE machine, separate from the existing AArch64 recommit testing.

AssadHashmi: In order to maximise SVE test coverage, we need to enable autovectorisation in the compiler. Without `-O3`, there will be very few SVE instruction sequences and the value of the functional tests greatly diminished.

derekbruening: Normally a compiler flag would be auto-enabled after running `CHECK_C_COMPILER_FLAG` so I'm not sure why this is a user-set option?  I would expect this to be auto-set based on the toolchain and platform (we already have `proc_supports_sve`)?

derekbruening: OK you mean this will be auto-set long-term, but is a user-set option for preliminary testing?  Please add a TODO comment by this option explaining that this is temporary and will be replaced by auto-setting.

suite/tests/CMakeLists.txt
```diff
@@ -553,6 +553,24 @@ function(add_exe test source)
       "${CMAKE_GENERATOR}" MATCHES "Visual Studio")
     add_dependencies(${test} ${gen_asm_tgt})
   endif ()
+  if (BUILD_TESTS_SVE)
+    target_compile_options(${test} PRIVATE
+      "-march=armv8.4-a+crypto+rcpc+sha3+sm4+sve+rng+ssbs+nodotprod")
```
derekbruening, line 558: Normally we would use `CHECK_C_COMPILER_FLAG` to confirm that a flag is actually supported.

suite/tests/CMakeLists.txt
derekbruening: OK I found on line 280 of this file:

```
  # some tests rely on specific "nop;nop" patterns that optimization ruins
  # we should probably move the -O from top level into core/CMakeLists.txt
  string(REGEX REPLACE "-O[0-9]? " " " CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")
```

Some client tests have the nops as sentinels.  If annotations were implemented on aarch64 (#1672) we could probably switch to those in all cases but we don't have that today.

So the tests, in in release build, are in fact built without high optimizations today.

Given the nop issues: probably for the short term only the tests you are targeting with auto-vectorization should have -O3 enabled; i.e., an allowlist, instead of the denylist here.  Have you run all the tests in -O3?  According to that comment we'd expect many client tests to fail.  Was the not the case?

derekbruening: Or, we can try to change that line 280 removal of -O* to only remove for a list of tests that have the nop;nop or similar patterns.

suite/tests/CMakeLists.txt
philramsey-arm: Have added TODO here to replace with CHECK_C_COMPILER_FLAG

suite/tests/CMakeLists.txt
philramsey-arm: With -O3 we get 36 test failures rather than the normal 6. But most of the tests still pass.

I think the plan is to investigate the new failures and if necessary build them without -O3

philramsey-arm: I like your idea of removing line 280 and removing -O3 in a more granular way. We can look into that.

philramsey-arm: I looked into removing the line
`string(REGEX REPLACE "-O[0-9]? " " " CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")`
and selectively adding the `-O0` flag to certain sources using `target_compile_options`

However this didn't work as the files were built with both `-O0` and `-O3` flags but the `-O3` came last so had precedence.

We could have an allow list for adding `-O3` but it would be much longer than the deny list.

derekbruening: Removing compile flags in CMake in a local manner has never been well-supported, which is why we have such hacky solutions like clearing CMAKE_C*_FLAGS completely in core DR so we can control it fully (which then leads to other problems like missing debug flags on "regular" targets).  I haven't kept up with new CMake features (I had never heard of `target_compile_options` until this PR): is it possible there is new support in CMake for removing options locally (per-directory or per-target) or for specifying that our target flags should be appended instead of prepended to the base flags?

philramsey-arm: @derekbruening 
My approach at the moment is to remove the line
`string(REGEX REPLACE "-O[0-9]? " " " CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")`
and then add code in `set_cflags` (for core tests) and `_DR_set_compile_flags` (for client tests) to remove "-O3" for tests that fail when built this way.

Although my change will end up being more verbose than the original change, it will avoid the blanket removal of "-Ox" for all tests.

Does this approach sound reasonable?

derekbruening: `_DR_set_compile_flags` in `make/DynamoRIOConfig.cmake.in` is used for real user clients, not just for tests, so that is not the right place to remove flags as users will not be happy with that.  That would presumably include us as that would remove optimizations from all of our provided tools including drmemtrace, drcachesim, etc. in our release packages.

philramsey-arm: @derekbruening I have reworked this PR now to detect if the build machine supports SVE rather than use a build option. I also have a whitelist of tests to build for SVE rather than a blacklist of tests not to build.

derekbruening: Please update the description which still talks about a build option.  If you click on the circular arrow you can then re-request a review.

suite/tests/CMakeLists.txt
```diff
@@ -474,6 +474,49 @@ function(append_link_flags target newflags)
...
+      security-common.retnonexisting security-common.TestAllocWE
+      security-common.TestMemProtChg_FLAKY client.drx-test
+     )
+
+function(add_sve_flags target)
+  if (proc_supports_sve)
+    if (target IN_LIST sve_tests)
+      target_compile_options(${target} PRIVATE
+        -march=armv8.4-a+crypto+rcpc+sha3+sm4+sve+rng+ssbs+nodotprod
+        -O3)
```
derekbruening, line 515: Please add a comment by line 282 where it removes -O* that we re-add it down here (trying the help future readers).

derekbruening, line 515: Comment that this re-instates the -O3 removed on line 282 (for release builds; debug maybe never had -O3)

derekbruening, line 515: I just noticed that this file has an `optimize` function which adds -O3.  It is called for a handful of targets, some of which are not on your list above.  Let's add a comment by that function pointing at getting rid of the line 282 change and pointing at this function and how we may want to combine them all.

suite/tests/CMakeLists.txt
```diff
@@ -474,6 +474,49 @@ function(append_link_flags target newflags)
...
+      linux.zero-length-mem-ranges linux.sigaction linux.syscall_pwait linux.sigaction_nosignals
+      linux.thread linux.threadexit linux.threadexit2 linux.signalfd linux.alarm
+      linux.signal_racesys linux.signal_pre_syscall linux.bad-signal-stack linux.sigsuspend
+      linux.sigmask linux.mangle_asynch linux.app_tls linux.readlink
+      linux.fib-conflict linux.fib-static linux.fib-pie linux.vfork pthreads.pthreads
+      pthreads.pthreads_exit pthreads.ptsig pthreads.pthreads_fork_FLAKY security-linux.trampoline
+      linux.infloop linux.rseq_disable security-common.codemod security-common.ret_noncall_trace
+      security-common.retnonexisting security-common.TestAllocWE
+      security-common.TestMemProtChg_FLAKY client.drx-test
+     )
```
derekbruening, line 508: Hmm, the earlier plan to remove the -O3 removal and have a blocklist of tests that do not want optimizations is nicer, as new tests won't be optimized unless someone comes and adds them here.  I know you spent some time on that and abandoned it, but let's at least put in some comments so it's clear what we consider finished/ideal and what we think still needs improvement.  Please add XXX or TODO comments about wanting to turn this from an allowlist to a blocklist (combined with eliminating the -O* removal).

suite/tests/CMakeLists.txt
```diff
@@ -474,6 +474,49 @@ function(append_link_flags target newflags)
...
+
+function(add_sve_flags target)
+  if (proc_supports_sve)
+    if (target IN_LIST sve_tests)
+      target_compile_options(${target} PRIVATE
```
derekbruening, line 513: Some targets are under various platform-specific options: I guess yours was pre-curated to be aarch64 Linux only?  I think it should be mentioned in a comment above the sve_tests list that it is only aarch64 Linux (b/c if someone adds some test not built for a64 linux we'll get errors).

suite/tests/CMakeLists.txt
```diff
@@ -474,6 +474,49 @@ function(append_link_flags target newflags)
     LINK_FLAGS "${cur_ldflags} ${newflags}")
 endfunction(append_link_flags)
 
+set(  sve_tests
```
derekbruening, line 477: nit: superfluous spaces

derekbruening (approved):
Mostly requesting adding a bunch of comments so future readers (probably ourselves after we forget the details) can more easily understand the combination of removing -O* and re-instating it in two different places (`optimize()` and now `add_sve_flags()`) that we have now.

suite/tests/CMakeLists.txt
derekbruening: And we reinstate it in `optimize()`.

suite/tests/CMakeLists.txt
```diff
@@ -512,6 +533,9 @@ function(add_sve_flags target)
     if (target IN_LIST sve_tests)
       target_compile_options(${target} PRIVATE
         -march=armv8.4-a+crypto+rcpc+sha3+sm4+sve+rng+ssbs+nodotprod
+        # reinstate the -03 flag which is removed with the
```
derekbruening, line 536: style nit: Prefer capitalized and punctuated sentences: s/r/R/; trailing `.`.

suite/tests/CMakeLists.txt
```diff
@@ -1746,7 +1770,11 @@ function(use_MD_not_MTd source_file)
   endif ()
 endfunction()
 
-# optimize the target
+# the -O3 flag is removed above with the
```
derekbruening, line 1773: style nit: `s/# t/# T/`

suite/tests/CMakeLists.txt
```diff
@@ -482,12 +482,12 @@ endfunction(append_link_flags)
...
 # support it.
 # This was investigated and proved to be non trivial as the client tests
-# use _DR_set_compile_flags() to set the compile flags whereas the core 
+# use _DR_set_compile_flags() to set the compile flags whereas the core
 # tests use set_cflags(). _DR_set_compile_flags() is a public function and we
 # didn't want to add a blocklist in there.
 # For now we create a list of tests that can be built with -O3.
 # This is for AARCH64 UNIX only.
-# TODO change this allowlist to a blocklist
+# TODO i#6429 Optimization flags are removed for all tests.
```
derekbruening, line 490: I would keep your original text: `TODO i#6429: Change this allowlist to a blocklist.`

derekbruening (approved):

PR #6372, opened 2023-10-16
i#5036 A64 scatter/gather, part 4: Expand scalar+scalar

jackgallagher-arm:
Adds support to drx_expand_scatter_gather() for SVE scalar+scalar contiguous load and store instructions, along with tests.

Issue: #5036

### DIFF
--- a/ext/drx/CMakeLists.txt
+++ b/ext/drx/CMakeLists.txt
@@ -62,6 +62,9 @@ configure_extension(drx OFF)
 use_DynamoRIO_extension(drx drcontainers)
 use_DynamoRIO_extension(drx drmgr)
 use_DynamoRIO_extension(drx drreg)
+if (AARCH64)
+    use_DynamoRIO_extension(drx drutil)
+endif ()
 
 # While we use drmgr in our implementation of drx, we only use pieces that
 # do not rely on drmgr's 4 phases or event orderings, and we do not include
@@ -84,6 +87,9 @@ configure_extension(drx_static ON)
 use_DynamoRIO_extension(drx_static drcontainers)
 use_DynamoRIO_extension(drx_static drmgr_static)
 use_DynamoRIO_extension(drx_static drreg_static)
+if (AARCH64)
+    use_DynamoRIO_extension(drx_static drutil_static)
+endif ()
 configure_drx_target(drx_static)
 
 install_ext_header(drx.h)
--- a/ext/drx/scatter_gather_aarch64.c
+++ b/ext/drx/scatter_gather_aarch64.c
@@ -39,6 +39,7 @@
 #include "drx.h"
 #include "drmgr.h"
 #include "drreg.h"
+#include "drutil.h"
 #include "../ext_utils.h"
 #include "scatter_gather_shared.h"
 
@@ -59,23 +60,32 @@
 #define SVE_PREDICATE_SPILL_SLOT_SIZE SVE_MAX_PREDICATE_LENGTH_BYTES
 
 typedef struct _per_thread_t {
-    void *scratch_pred_spill_slot; /* Storage for spilled predicate register. */
+    void *scratch_pred_spill_slot;           /* Storage for spilled predicate register. */
+    void *scratch_vector_spill_slot;         /* Storage for spilled vector register. */
+    void *scratch_vector_spill_slot_aligned; /* Aligned ptr inside
+                                                scratch_vector_spill_slot to save/restore
+                                                spilled Z vector register. */
 } per_thread_t;
 
 /* Track the state of manual spill slots for SVE registers.
  * This corresponds to the spill slot storage in per_thread_t.
  */
 typedef struct _spill_slot_state_t {
-    reg_id_t pred_slots[1];
+#define NUM_PRED_SLOTS 1
+    reg_id_t pred_slots[NUM_PRED_SLOTS];
+
+#define NUM_VECTOR_SLOTS 1
+    reg_id_t vector_slots[NUM_VECTOR_SLOTS];
 } spill_slot_state_t;
 
 void
 init_spill_slot_state(OUT spill_slot_state_t *spill_slot_state)
 {
-    const size_t num_pred_slots =
-        sizeof(spill_slot_state->pred_slots) / sizeof(spill_slot_state->pred_slots[0]);
-    for (size_t i = 0; i < num_pred_slots; i++)
+    for (size_t i = 0; i < NUM_PRED_SLOTS; i++)
         spill_slot_state->pred_slots[i] = DR_REG_NULL;
+
+    for (size_t i = 0; i < NUM_VECTOR_SLOTS; i++)
+        spill_slot_state->vector_slots[i] = DR_REG_NULL;
 }
 
 void
@@ -96,6 +106,15 @@ drx_scatter_gather_thread_init(void *drcontext)
     DR_ASSERT_MSG(ALIGNED(pt->scratch_pred_spill_slot, SVE_PREDICATE_ALIGNMENT_BYTES),
                   "scratch_pred_spill_slot is misaligned");
 
+    /*
+     * The scalable vector versions of LDR/STR require 16 byte alignment so we have to
+     * over-allocate and get an aligned pointer inside the allocated memory.
+     */
+    pt->scratch_vector_spill_slot =
+        dr_thread_alloc(drcontext, SVE_VECTOR_SPILL_SLOT_SIZE);
+    pt->scratch_vector_spill_slot_aligned =
+        (void *)ALIGN_FORWARD(pt->scratch_vector_spill_slot, SVE_VECTOR_ALIGNMENT_BYTES);
+
     drmgr_set_tls_field(drcontext, drx_scatter_gather_tls_idx, (void *)pt);
 }
 
@@ -105,6 +124,7 @@ drx_scatter_gather_thread_exit(void *drcontext)
     per_thread_t *pt =
         (per_thread_t *)drmgr_get_tls_field(drcontext, drx_scatter_gather_tls_idx);
     dr_thread_free(drcontext, pt->scratch_pred_spill_slot, SVE_PREDICATE_SPILL_SLOT_SIZE);
+    dr_thread_free(drcontext, pt->scratch_vector_spill_slot, SVE_VECTOR_SPILL_SLOT_SIZE);
     dr_thread_free(drcontext, pt, sizeof(*pt));
 }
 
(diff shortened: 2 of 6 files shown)


ext/drx/scatter_gather_shared.h
```diff
@@ -85,6 +85,7 @@ typedef struct _scatter_gather_info_t {
                      */
     bool scaled;
     bool is_scalar_value_signed;
+    bool is_replicating;
```
abhinav92003, line 88: Add comment about what some of these mean. To avoid the reader having to go to the manual.

Could you also add a link to the canonical reference to the manual anyway? Apologies if it's already there in the header and I missed it.

ext/drx/scatter_gather_aarch64.c
```diff
@@ -623,16 +772,31 @@ drx_expand_scatter_gather(void *drcontext, instrlist_t *bb, OUT bool *expanded)
...
+     * We return true with *expanded=false here to indicate that no error occurred but
+     * we didn't expand any instructions. This matches the behaviour of this function
+     * for architectures with no scatter/gather expansion support.
+     */
+    if (sg_info.faulting_behavior != DRX_NORMAL_FAULTING) {
+        /* TODO i#5036: Add support for first-fault and non-fault accesses. */
+        return true;
+    }
+    if (!reg_is_z(sg_info.base_reg) && sg_info.index_reg == DR_REG_NULL) {
+        /* TODO i#5036: Add support for scalar+immediate contiguous accesses. */
```
abhinav92003, line 785: scalar+imm is probably not very different from scalar+scalar right? Or is there much additional complexity? (Not suggesting that we add that support here; just curious).

ext/drx/scatter_gather_aarch64.c
```diff
@@ -641,11 +805,19 @@ drx_expand_scatter_gather(void *drcontext, instrlist_t *bb, OUT bool *expanded)
...
+                                              DR_REG_INVALID };
     drvector_t allowed;
     drreg_init_and_fill_vector(&allowed, true);
 
+    /* Figure out how many scratch registers we need */
+    uint num_scratch_gpr_needed = 1;
+    if (!sg_info.is_load)
+        num_scratch_gpr_needed++;
+    if (is_contiguous)
+        num_scratch_gpr_needed++;
```
abhinav92003, line 819: Would be nice to add a comment over all three of these to say exactly what app reg we need the scratch for.

E.g. here we can say that it's for the gpr that we use to store the base reg of the contiguous memory range.

suite/tests/client-interface/drx-scattergather-aarch64.cpp
```diff
@@ -1004,484 +1023,483 @@ test_ld1_scalar_plus_vector()
...
             std::array<uint64_t, 2> { 1, 22 },
-            input_data.base_addr_for_data_size(element_size_t::BYTE),
+            INPUT_DATA.base_addr_for_data_size(element_size_t::BYTE),
         },
         {
             "ld1b scalar+vector 32bit unpacked unscaled offset sxtw",
             TEST_FUNC("ld1b z3.d, p4/z, [%[base], z28.d, sxtw]"),
             { /*zt=*/3, /*pg=*/4, /*zm=*/28 },
-            std::array<uint64_t, 2> { 0xF2, 0x29 },
+            std::array<uint64_t, 2> { 0xF2, 0xF3 },
```
abhinav92003, line 1048: Curious why these changes in the test reference values were needed.

abhinav92003 (approved):

ext/drx/scatter_gather_aarch64.c
```diff
@@ -499,6 +521,105 @@ expand_scatter_gather(void *drcontext, instrlist_t *bb, instr_t *sg_instr,
...
+     *     ld1h z31.h, p0/z, [x2, z0.h, lsl #1]
+     * which is not a valid SVE instruction (scatter/gather instructions only support
+     * S and D element sizes).
+     * It doesn't matter that this theoretical instruction does not exist;
+     * expand_scatter_gather() is able to generate a sequence of valid instructions that
+     * carry out the described operation correctly anyway.
+     */
+
+    /* Expand the instruction as if it were a scalar+vector scatter/gather instruction */
+    expand_scatter_gather(drcontext, bb, sg_instr, &modified_sg_info, scratch_gpr1,
```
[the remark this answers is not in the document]
jackgallagher-arm: >  is there any environment where the contiguous operation is supported but the non-contiguous operations (and perhaps some operations added by expand_scatter_gather) are not?

No. All the instructions that `drx_expand_scatter_gather()` supports so far are part of the base SVE instruction set extension, and I have been careful to make sure that the instructions which `drx_expand_scatter_gather()` emits are base SVE instructions too so it should support any SVE hardware.

ext/drx/scatter_gather_aarch64.c
jackgallagher-arm: I have actually got that working too and there should be a patch for it coming soon. It's a simple change to `ext/drx/scatter_gather_aarch64.c` but it uncovered a bug in the AArch64 codec for the scalar+immediate instructions which needs to be fixed first (patch for that also coming soon).

suite/tests/client-interface/drx-scattergather-aarch64.cpp
jackgallagher-arm: Its because I changed how the input data is laid out. It used to just be a sequence of 40 values, and that was fine for testing the true scatter/gather instructions because each element can point anywhere within the input data range. For the contiguous instructions you can set the start of the contiguous range wherever you like but then you need to be able to load however many bytes of data after that base. For something like an ld4b which loads 4 vector registers that means if you have 512-bit vector registers you need `4 * (512/8) = 256` bytes so it couldn't fit in the old input data scheme.
I changed it to increase the size of the input data a lot and fill the data with a repeating 32 value pattern so we should now have enough input data to test contiguous instructions with a variety of offsets, but that change meant some of the original scatter/gather test needed updating because the input data layout was different.

ext/drx/scatter_gather_aarch64.c
```diff
@@ -59,23 +60,32 @@
 #define SVE_PREDICATE_SPILL_SLOT_SIZE SVE_MAX_PREDICATE_LENGTH_BYTES
 
 typedef struct _per_thread_t {
-    void *scratch_pred_spill_slot; /* Storage for spilled predicate register. */
+    void *scratch_pred_spill_slot;           /* Storage for spilled predicate register. */
+    void *scratch_vector_spill_slot;         /* Storage for spilled vector register. */
```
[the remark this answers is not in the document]
jackgallagher-arm: Done.

ext/drx/scatter_gather_aarch64.c
```diff
@@ -141,92 +161,94 @@ get_scatter_gather_info(instr_t *instr, OUT scatter_gather_info_t *sg_info)
     sg_info->scatter_gather_size = opnd_get_size(memopnd);
 
     switch (instr_get_opcode(instr)) {
-#define DRX_CASE(op, _reg_count, value_size, value_signed, _faulting_behavior) \
-    case OP_##op:                                                              \
-        sg_info->reg_count = _reg_count;                                       \
-        sg_info->scalar_value_size = value_size;                               \
-        sg_info->is_scalar_value_signed = value_signed;                        \
-        sg_info->faulting_behavior = _faulting_behavior;                       \
+#define DRX_CASE(op, _reg_count, value_size, value_signed, replicating, \
```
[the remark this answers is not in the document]
jackgallagher-arm: The `_` is just there to stop the argument name clashing with the struct member name. I have changed it to make the names consistent.

ext/drx/scatter_gather_shared.h
jackgallagher-arm: Done.

ext/drx/scatter_gather_aarch64.c
```diff
@@ -499,6 +521,105 @@ expand_scatter_gather(void *drcontext, instrlist_t *bb, instr_t *sg_instr,
...
+
+    /* Populate the new vector index register */
+
+    /* index    scratch_vec.element_size, #0, #1 */
+    instrlist_preinsert(
+        bb, sg_instr,
+        INSTR_XL8(INSTR_CREATE_index_sve(
+                      drcontext,
+                      opnd_create_reg_element_vector(scratch_vec, sg_info->element_size),
+                      opnd_create_immed_int(0, OPSZ_5b),
```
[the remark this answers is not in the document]
jackgallagher-arm: Done.

ext/drx/scatter_gather_aarch64.c
jackgallagher-arm: Done.

ext/drx/scatter_gather_aarch64.c
```diff
@@ -594,6 +730,19 @@ unreserve_pred_register(void *drcontext, instrlist_t *bb, instr_t *where,
                            opnd_size_from_bytes(proc_get_vector_length_bytes() / 8));
 }
 
+void
+unreserve_vector_register(void *drcontext, instrlist_t *bb, instr_t *where,
+                          reg_id_t scratch_gpr0, reg_id_t scratch_vec,
+                          spill_slot_state_t *slot_state)
+{
+    DR_ASSERT(slot_state->vector_slots[0] == scratch_vec);
+    slot_state->vector_slots[0] = DR_REG_NULL;
```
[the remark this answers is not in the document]
jackgallagher-arm: I have rewritten this code a bit and hopefully now its clearer how it is used.
`*scratch_vector_spill_slot_aligned` is now a buffer of size `NUM_VECTOR_SLOTS * vector_length_in_bytes` and `slot_state->vector_slots` is an array corresponding to the slots inside `scratch_vector_spill_slot_aligned`. It just happens that `NUM_VECTOR_SLOTS` is `1` so it only contains one slot at the moment. This is how I always envisioned it, there were just a few places before that assumed `NUM_VECTOR_SLOTS == 1` so it might not have been completely clear.

abhinav92003 (approved):

ext/drx/CMakeLists.txt
```diff
@@ -62,6 +62,9 @@ configure_extension(drx OFF)
 use_DynamoRIO_extension(drx drcontainers)
 use_DynamoRIO_extension(drx drmgr)
 use_DynamoRIO_extension(drx drreg)
+if (AARCH64)
+    use_DynamoRIO_extension(drx drutil)
```
derekbruening, line 66: Adding a drutil dependence to drx is problematic for the licenses: drx was created to be separate from drutil's LGPL license (which was there because it came from Dr. Memory which is LGPL).  So this dependence kind of defeats the purpose of that license split.  Most of our large tools end up using drutil anyway, but this may affect some external user.

Is there an alternative?

This should at least be well-documented: these places would have to be updated:
```
ext/drx/README.md:utilities for instrumentation and sports a BSD license, as opposed to the drutil Extension
ext/drx/drx.dox:instrumentation and sports a BSD license, as opposed to the \p drutil Extension
```

jackgallagher-arm: Thanks. I didn't realise. I will have a look for an alternative solution.

jackgallagher-arm: Fixed by PR: #6389

Issue #6373, opened 2023-10-16
Invariant checker does not handle back-to-back signals after an rseq abort correctly.


ivankyluk:
After an rseq abort, a kernel event marker is added. If a signal follows immediately, another kernel event maker will be added with a  different value.

If a back-to-back signal happens without any intervening instructions, the new kernel event marker should have the same value as the previous one. 

The current implementation of the invariant check fails to consider this case, and flag an error.

If the instruction before the rseq marker is a branch, the invariant checker reports "Branch does not go to the correct target @ kernel_event marker", otherwise it reports "Non-explicit control flow has no marker @ kernel_event marker".

The fix is to check if the kernel event marker has the same value as the previous one, and there are no intervening instructions between signals, do not flag an error.

PR #6374, opened 2023-10-16
i#6373: handle back-to-back signals after an rseq abort.

ivankyluk:
When a signal causes an rseq abort, there will be a rseq  abort marker, followed by a kernel event marker.

When another signal follows immediately, another kernel event maker will be added with a different value.

If a back-to-back signal happens without any intervening instructions, the new kernel event marker (3rd one) should have the same value as the second kernel event marker.

The current implementation of the invariant check fails to consider this case, and flag an error.

If the instruction before the rseq marker is a branch, the invariant checker reports "Branch does not go to the correct target @ kernel_event marker", otherwise it reports "Non-explicit control flow has no marker @ kernel_event marker".

The fix is to check if the kernel event marker has the same value as the previous one, and if there are no intervening instructions between signals, do not flag an error.

Fixes: #6373

### DIFF
--- a/clients/drcachesim/tests/invariant_checker_test.cpp
+++ b/clients/drcachesim/tests/invariant_checker_test.cpp
@@ -470,6 +470,55 @@ check_sane_control_flow()
             return false;
         }
     }
+    // Correct test: back-to-back signals after an RSEQ abort.
+    {
+        std::vector<memref_t> memrefs = {
+            gen_instr(TID, /*pc=*/101, /*size=*/1),
+            // The RSEQ_ABORT marker is always follwed by a KERNEL_EVENT marker.
+            gen_marker(TID, TRACE_MARKER_TYPE_RSEQ_ABORT, 102),
+            gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_EVENT, 102),
+            // We get a signal after the RSEQ abort.
+            gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_EVENT, 301),
+            gen_instr(TID, /*pc=*/201, /*size=*/1),
+            gen_marker(TID, TRACE_MARKER_TYPE_SYSCALL, 15),
+            gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_XFER, 202),
+            // The kernel event marker has the same value as the previous one.
+            gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_EVENT, 301),
+            gen_instr(TID, /*pc=*/201, /*size=*/1),
+            gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_XFER, 202),
+            gen_instr(TID, /*pc=*/301, /*size=*/1),
+        };
+        if (!run_checker(memrefs, false)) {
+            return false;
+        }
+    }
+    // Incorrect test: back-to-back signals with an intervening instruction after an RSEQ
+    // abort.
+    {
+        std::vector<memref_t> memrefs = {
+            gen_instr(TID, /*pc=*/101, /*size=*/1),
+            // The RSEQ_ABORT marker is always follwed by a KERNEL_EVENT marker.
+            gen_marker(TID, TRACE_MARKER_TYPE_RSEQ_ABORT, 102),
+            gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_EVENT, 102),
+            // We get a signal after the RSEQ abort.
+            gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_EVENT, 301),
+            gen_instr(TID, /*pc=*/201, /*size=*/1),
+            gen_marker(TID, TRACE_MARKER_TYPE_SYSCALL, 15),
+            gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_XFER, 202),
+            gen_instr(TID, /*pc=*/301, /*size=*/1),
+            gen_instr(TID, /*pc=*/302, /*size=*/1),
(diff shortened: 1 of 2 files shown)


clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -470,6 +470,55 @@ check_sane_control_flow()
...
         }
     }
+    // Correct test: back-to-back signals after an RSEQ abort.
+    {
+        std::vector<memref_t> memrefs = {
+            gen_instr(TID, /*pc=*/101, /*size=*/1),
+            // The RSEQ_ABORT marker is always follwed by a KERNEL_EVENT marker.
+            gen_marker(TID, TRACE_MARKER_TYPE_RSEQ_ABORT, 102),
+            gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_EVENT, 102),
+            // We get a signal after the RSEQ abort.
```
derekbruening, line 480: This is actually the signal that *caused* the rseq abort (otherwise the marker values would have to match) so I would reword this comment.

(102 is the rseq interruption point, which != handler only for signal interruption; 301 is rseq handler which is the signal resumption point)

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -470,6 +470,55 @@ check_sane_control_flow()
...
+    }
+    // Incorrect test: back-to-back signals with an intervening instruction after an RSEQ
+    // abort.
+    {
+        std::vector<memref_t> memrefs = {
+            gen_instr(TID, /*pc=*/101, /*size=*/1),
+            // The RSEQ_ABORT marker is always follwed by a KERNEL_EVENT marker.
+            gen_marker(TID, TRACE_MARKER_TYPE_RSEQ_ABORT, 102),
+            gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_EVENT, 102),
+            // We get a signal after the RSEQ abort.
```
derekbruening, line 503: Ditto.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -470,6 +470,55 @@ check_sane_control_flow()
...
+            gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_EVENT, 301),
+            gen_instr(TID, /*pc=*/201, /*size=*/1),
+            gen_marker(TID, TRACE_MARKER_TYPE_SYSCALL, 15),
+            gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_XFER, 202),
+            gen_instr(TID, /*pc=*/301, /*size=*/1),
+            gen_instr(TID, /*pc=*/302, /*size=*/1),
+            gen_marker(TID, TRACE_MARKER_TYPE_KERNEL_EVENT, 301),
```
derekbruening, line 510: OK so the error is here because 301 != 302 or 303 -- add comment to make it clear?

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -1130,6 +1130,13 @@ invariant_checker_t::check_for_pc_discontinuity(
          (shard->prev_xfer_marker_.marker.marker_type == TRACE_MARKER_TYPE_KERNEL_EVENT ||
           shard->prev_xfer_marker_.marker.marker_type == TRACE_MARKER_TYPE_KERNEL_XFER ||
           shard->prev_xfer_marker_.marker.marker_type == TRACE_MARKER_TYPE_RSEQ_ABORT)) ||
+#ifdef UNIX
+        // Same PC is allowed for back-to-back signals without any intervening
+        // instructions.
+        (cur_pc == shard->last_signal_context_.xfer_int_pc &&
+         prev_instr_trace_pc ==
+             shard->last_signal_context_.pre_signal_instr.memref.instr.addr) ||
```
derekbruening, line 1138: Should this include `at_kernel_event`?

derekbruening, line 1138: Trying to wrap my head around this: so we already allow the pre-signal-instr PC to equal the post-signal PC (for synchronous signal), up at line 1127?  Oh, but that's just on interruption: the check on sigreturn is done elsewhere at line 642.  So line 1127 is allowing the pre-signal-instr PC to equal the interruption PC.  But that's not enough here b/c the rseq abort causes the pre-signal-instr PC to not equal the interruption PC which is the rseq handler.

So a key thing here is that last_signal_context_ is *not* yet updated: it holds the prior signal, not the one we're entering.  Please add a comment about that.  (Otherwise these checks don't make sense.). With that fact: ok, yes, it does make sense.

derekbruening (approved):

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: Done.

clients/drcachesim/tools/invariant_checker.cpp
ivankyluk: Yes, I've added at_kernel_event.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: Done.

clients/drcachesim/tests/invariant_checker_test.cpp
ivankyluk: Done.

clients/drcachesim/tools/invariant_checker.cpp
ivankyluk: Done.

PR #6379, opened 2023-10-19
Document the endianness of drmemtrace wire formats

derekbruening:
Adds explicit documentation that the various drmemtrace packed on-disk formats are little-endian, except the raw offline trace format (for efficiency).  The memref_t format is left as flexible as the reader can convert from the little-endian trace_entry_t if desired.

### DIFF
--- a/clients/drcachesim/common/trace_entry.h
+++ b/clients/drcachesim/common/trace_entry.h
@@ -680,8 +680,10 @@ marker_type_is_function_marker(const trace_marker_type_t mark)
  * - a thread/process.
  * All fields are stored as little-endian.  The raw records from the tracer may
  * be big-endian (per the architecture trace type field), in which case raw2trace must
- * convert them to little-endian.  The #memref_t fields may also be big-endian to
- * simplify analyzers, with the #dynamorio::drmemtrace::reader_t class converting.
+ * convert them to little-endian.  The #memref_t fields may be presented as big-endian
+ * to simplify analyzers running on big-endian machines, in which case the conversion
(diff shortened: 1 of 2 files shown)


clients/drcachesim/common/trace_entry.h
```diff
@@ -677,7 +677,11 @@ marker_type_is_function_marker(const trace_marker_type_t mark)
  * - a bundle of instrs
  * - a flush request
  * - a prefetch request
- * - a thread/process
+ * - a thread/process.
+ * All fields are stored as little-endian.  The raw records from the tracer may
+ * be big-endian (per the architecture trace type field), in which case raw2trace must
+ * convert them to little-endian.  The #memref_t fields may also be big-endian to
+ * simplify analyzers, with the #dynamorio::drmemtrace::reader_t class converting.
```
brettcoon, line 684: I had to read this last sentence a few times to understand.  I think it might help to emphasize memref_t is not the trace format, and this "also be big-endian" clause refers to trace playback conversion (and is independent of whether or not the trace was captured on a big-endian machine).  How about something like "The #memref_t fields may be presented as big-endian to simplify analyzers running on big-endian machines, in which case the conversion from trace format to big-endian is performed by the #dynamorio::drmemtrace::reader_t class."

brettcoon (approved):
Looks good.

clients/drcachesim/common/trace_entry.h
derekbruening: Done.

PR #6381, opened 2023-10-19
i#5505 kernel tracing: Changes to enable third-party import

abhinav92003:
Makes various changes to the kernel PT tracing code to allow third-party code to import and integrate the kernel tracing functionality.

Adds structs for pt_data_buf_t and pt_metadata_buf_t that describe one unit of data read from the kthread file. Extracts pt_metadata_t from drpttracer.h to a new header (drpttracer_shared.h) to allow easy sharing with drmemtrace.

Adds drmemtrace APIs to get the kcore, kallsyms and kernel log file paths.

Separates out the logic to read kernel PT data from kthread file into a separate get_next_kernel_entry function which can be overridden by 3P code.

Adds raw2trace stats for count of system calls decoded using PT and the count of kernel instructions in the trace.

Updates the expected output for the tool.drcacheoff.kernel.simple test.

Makes other misc fixes, like using non-relative headers in include statements, and fixes a bug with uninitialised vars and wrong drcontext being used.

Removes pre_read_pt_entries from tdata which is used to buffer the PT data header until we read the PT data. This is
not really needed as we always read the data immediately after reading the header.

Modifies tool.drcacheoff.kernel.simple_SUDO to use "-raw_compress none" which enables check for unsafe static behaviour
(use of malloc etc) for the kernel tracing flow.

Verified locally that the kernel tests, which require root privileges and the PT feature on the processor, pass: client.drpttracer_SUDO-test and tool.drcacheoff.kernel.simple_SUDO.

Issue: #5505


prasun3:
Curious what 3P means?

clients/drcachesim/tracer/syscall_pt_trace.cpp
```diff
@@ -230,6 +233,8 @@ syscall_pt_trace_t::trace_data_dump(drpttracer_output_autoclean_t &output)
         return false;
     }
 
+    uint64_t total_size = PT_DATA_PDB_HEADER_SIZE + data->pt_size;
+    auto to_write = std::unique_ptr<uint8_t[]>(new uint8_t[total_size]);
```
derekbruening, line 237: It is not safe to call new for static app link setups.  Use placement new instead via DR's allocator.

There is an automated check that should have failed with this -- ah is it because GA never tests BUILD_PT_TRACER?  And your local runs only ran these couple of tests?

This is not a good situation: we need our tests to catch things like this.  I don't remember exactly which tests would catch this: please see  #DR_DISALLOW_UNSAFE_STATIC and dr_disallow_unsafe_static_behavior().  Please try to set it up so that GA will fail on this -- before submitting this PR ideally.

clients/drcachesim/tracer/syscall_pt_trace.cpp
```diff
@@ -260,19 +265,17 @@ syscall_pt_trace_t::trace_data_dump(drpttracer_output_autoclean_t &output)
...
-    }
+    pdb_header[PDB_HEADER_DATA_BOUNDARY_IDX].pt_data_boundary.data_size = data->pt_size +
+        PT_DATA_PDB_HEADER_SIZE +
+        pdb_header[PDB_HEADER_NUM_ARGS_IDX].syscall_args_num.args_num * sizeof(uint64_t);
 
-    /* Write the syscall's PT data to the output file */
-    if (write_file_func_(output_file_, data->pt_buffer, data->pt_size) == 0) {
+    /* We want to write the data in a single call to write_file_func_ for easier 3P
+     * integration.
+     * XXX: Could we avoid the second memcpy which may be potentially large?
```
derekbruening, line 274: Is this better in the end?  As you say you have an extra copy -- doesn't that outweigh the overhead of a 2nd write call?  Why is a 2nd write call so bad?  At first glance the original looks better to me.

clients/drcachesim/tracer/syscall_pt_trace.cpp
```diff
@@ -188,29 +193,27 @@ syscall_pt_trace_t::metadata_dump(pt_metadata_t metadata)
...
+    to_write.header[PDB_HEADER_DATA_BOUNDARY_IDX].pt_metadata_boundary.data_size =
         sizeof(pt_metadata_t);
-    pdb_header[PDB_HEADER_DATA_BOUNDARY_IDX].pt_metadata_boundary.type =
+    to_write.header[PDB_HEADER_DATA_BOUNDARY_IDX].pt_metadata_boundary.type =
         SYSCALL_PT_ENTRY_TYPE_PT_METADATA_BOUNDARY;
 
-    /* Write the buffer header to the output file */
-    if (write_file_func_(output_file_, pdb_header, PT_METADATA_PDB_HEADER_SIZE) == 0) {
+    /* We want to write the data in a single call to write_file_func_ for easier 3P
+     * integration.
```
derekbruening, line 209: Please elaborate on why this makes integration easier: the write call is already indirected, so why would one vs 2 calls affect integration?  It should only affect overhead -- and there it is not clear which is better if the combining uses a copy and there is buffering already on the indirected write mechanism.

clients/drcachesim/tracer/syscall_pt_trace.cpp
```diff
@@ -172,9 +177,9 @@ syscall_pt_trace_t::stop_syscall_pt_trace()
     cur_recording_sysnum_ = -1;
 
     /* Reset the pttracer handle for next syscall.
-     * TODO i#5505: To reduce the overhead caused by pttracer initialization, we need to
-     * share the same pttracer handle for all syscalls per thread. And also, we need to
-     * improve the libpt2ir to support streaming decoding.
+     * TODO i#5505: To reduce the overhead caused by pttracer initialization, we
+     * need to share the same pttracer handle for all syscalls per thread. And
+     * also, we need to improve the libpt2ir to support streaming decoding.
```
derekbruening, line 182: Not seeing any change here?  Accidental reformat?

clients/drcachesim/tracer/syscall_pt_trace.cpp
```diff
@@ -125,9 +130,9 @@ syscall_pt_trace_t::start_syscall_pt_trace(IN int sysnum)
         return false;
     }
 
-    /* All syscalls within a single thread share the same pttracer configuration, and
-     * thus, the same pt_metadata. Metadata is dumped at the beginning of the output
-     * file and occurs only once.
+    /* All syscalls within a single thread share the same pttracer configuration,
+     * and thus, the same pt_metadata. Metadata is dumped at the beginning of the
+     * output file and occurs only once.
```
derekbruening, line 135: Not seeing any change here? Accidental reformat?

clients/drcachesim/tracer/syscall_pt_trace.cpp
```diff
@@ -91,18 +94,20 @@ syscall_pt_trace_t::init(void *drcontext, char *pt_dir_name,
...
         "/" + std::to_string(dr_get_thread_id(drcontext_)) + "." + OUTFILE_SUFFIX_PT;
-    output_file_ = open_file_func_(output_file_name.c_str(), DR_FILE_WRITE_REQUIRE_NEW);
+    // TODO i#5505: Pass the window-id to support windowed traces.
+    output_file_ = open_file_ex_func_(output_file_name.c_str(), DR_FILE_WRITE_REQUIRE_NEW,
+                                      dr_get_thread_id(drcontext_), 0);
 
-    /* Create a buffer to store the data generated by drpttracer. For syscall traces, only
-     * the PT data is dumped, and the sideband data is not included.
+    /* Create a buffer to store the data generated by drpttracer. For syscall
+     * traces, only the PT data is dumped, and the sideband data is not included.
```
derekbruening, line 110: Not seeing any change here? Accidental reformat?

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -128,6 +128,8 @@ typedef enum {
     RAW2TRACE_STAT_EARLIEST_TRACE_TIMESTAMP,
     RAW2TRACE_STAT_LATEST_TRACE_TIMESTAMP,
     RAW2TRACE_STAT_FINAL_TRACE_INSTRUCTION_COUNT,
+    RAW2TRACE_STAT_KERNEL_INSTR_COUNT,
+    RAW2TRACE_STAT_SYSCALL_PT_TRACES_DECODED,
```
derekbruening, line 132: Again thinking ahead: for interfaces where it's hard to rename things, I would not put "PT" in here.  Use a generic time like "KERNEL".

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -845,42 +844,93 @@ raw2trace_t::process_header(raw2trace_thread_data_t *tdata)
...
+                                   std::unique_ptr<pt_metadata_buf_t> &pt_metadata,
+                                   uint64_t expected_syscall_idx)
+{
+    DR_ASSERT(tdata->kthread_file != nullptr);
+    if (tdata->kthread_file->eof())
+        return nullptr;
+    if (!tdata->pt_metadata_processed) {
+        log(2, "Reading PT metadata for tid " INT64_FORMAT_STRING "\n", tdata->tid);
+        pt_metadata = std::unique_ptr<pt_metadata_buf_t>(new pt_metadata_buf_t());
+        if (!tdata->kthread_file->read((char *)&pt_metadata->header[0],
```
derekbruening, line 859: Prefer C++-style casts.  Ditto for the 2 similar ones below.

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -894,55 +944,36 @@ raw2trace_t::process_syscall_pt(raw2trace_thread_data_t *tdata, uint64_t syscall
...
         VPRINT(1, "Finished decoding all PT data for thread %d\n", tdata->tid);
         return true;
     }
+    uint64_t syscall_args_num =
+        pt_data->header[PDB_HEADER_NUM_ARGS_IDX].syscall_args_num.args_num;
+    uint64_t pt_data_size =
+        pt_data->header[PDB_HEADER_DATA_BOUNDARY_IDX].pt_data_boundary.data_size -
+        PT_DATA_PDB_HEADER_SIZE - syscall_args_num * sizeof(uint64_t);
 
-    if (tdata->pre_read_pt_entries.empty()) {
```
derekbruening, line 902: Confused about where this line went in the refactoring?

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -894,55 +944,36 @@ raw2trace_t::process_syscall_pt(raw2trace_thread_data_t *tdata, uint64_t syscall
...
-        tdata->pre_read_pt_entries.insert(tdata->pre_read_pt_entries.end(), header,
-                                          header + PT_DATA_PDB_HEADER_ENTRY_NUM);
-    }
-
-    if (tdata->pre_read_pt_entries[PDB_HEADER_DATA_BOUNDARY_IDX].pt_data_boundary.type !=
-            SYSCALL_PT_ENTRY_TYPE_PT_DATA_BOUNDARY ||
-        tdata->pre_read_pt_entries[PDB_HEADER_SYSCALL_IDX_IDX].syscall_idx.type !=
-            SYSCALL_PT_ENTRY_TYPE_SYSCALL_IDX ||
-        tdata->pre_read_pt_entries[PDB_HEADER_SYSCALL_IDX_IDX].syscall_idx.idx !=
-            syscall_idx) {
```
derekbruening, line 922: Where did these checks go?

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -845,42 +844,93 @@ raw2trace_t::process_header(raw2trace_thread_data_t *tdata)
...
+    uint64_t pt_data_size =
+        pt_data->header[PDB_HEADER_DATA_BOUNDARY_IDX].pt_data_boundary.data_size -
+        PT_DATA_PDB_HEADER_SIZE - syscall_args_num * sizeof(uint64_t);
+
+    log(2,
+        "Reading PT data for syscall_idx " INT64_FORMAT_STRING
+        " size " INT64_FORMAT_STRING " tid " INT64_FORMAT_STRING
+        " pid " INT64_FORMAT_STRING " num " INT64_FORMAT_STRING "\n",
+        syscall_idx, pt_data_size, tid, pid, sysnum);
+    pt_data->data.reset(new uint8_t[pt_data_size]);
```
derekbruening, line 902: Not sure I completely follow this: you allocate memory on line 876 and read the header.  You pull out pid, tid, etc. in local vars lines 887-895 -- but then you throw away the header and never use those locals (except `pt_data_size`).  So the header is only needed for the size?  Should those other locals be inside a debug-only construct or sthg, or only extracted in the log()?

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1099,12 +1103,24 @@ class raw2trace_t {
 
 #ifdef BUILD_PT_POST_PROCESSOR
         std::istream *kthread_file;
-        std::vector<syscall_pt_entry_t> pre_read_pt_entries;
```
derekbruening, line 1102: What was this used for before such that is it not needed now?  May be worth a mention in the PR description.

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -894,55 +944,36 @@ raw2trace_t::process_syscall_pt(raw2trace_thread_data_t *tdata, uint64_t syscall
...
 
-    if (tdata->pre_read_pt_entries.empty() && tdata->kthread_file->eof()) {
+    if (pt_data == nullptr) {
+        if (!tdata->error.empty())
+            return false;
         VPRINT(1, "Finished decoding all PT data for thread %d\n", tdata->tid);
         return true;
     }
+    uint64_t syscall_args_num =
+        pt_data->header[PDB_HEADER_NUM_ARGS_IDX].syscall_args_num.args_num;
```
derekbruening, line 954: Another local that seems unused?

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -845,42 +844,93 @@ raw2trace_t::process_header(raw2trace_thread_data_t *tdata)
...
+    return pt_data;
+}
+
 bool
 raw2trace_t::process_syscall_pt(raw2trace_thread_data_t *tdata, uint64_t syscall_idx)
 {
-    DR_ASSERT(tdata->kthread_file != nullptr);
     DR_ASSERT(TESTANY(OFFLINE_FILE_TYPE_KERNEL_SYSCALLS, tdata->file_type));
-
+    std::unique_ptr<pt_metadata_buf_t> pt_metadata = nullptr;
```
derekbruening, line 915: The default construtor is sufficient?

derekbruening (commented):
Want to make sure I understand the lack of an automated test failure for the `new` call before LGTM and whether we can do something about it easily or whether that has to be separated out.

abhinav92003:
> Want to make sure I understand the lack of an automated test failure for the `new` call before LGTM and whether we can do something about it easily or whether that has to be separated out.

That's because PT tests require root privileges. The PT tests are run only when RUN_SUDO_TESTS=ON. From what I remember, we found that Github actions would block at the password prompt if we try to run it in the workflow.

derekbruening:
> > Want to make sure I understand the lack of an automated test failure for the `new` call before LGTM and whether we can do something about it easily or whether that has to be separated out.
> 
> That's because PT tests require root privileges. The PT tests are run only when RUN_SUDO_TESTS=ON. From what I remember, we found that Github actions would block at the password prompt if we try to run it in the workflow.

No, that is not correct.  The GA VMs have passwordless sudo and we are running `tool.drcachesim.phys-threads_SUDO` today.

abhinav92003:
> > > Want to make sure I understand the lack of an automated test failure for the `new` call before LGTM and whether we can do something about it easily or whether that has to be separated out.
> > 
> > 
> > That's because PT tests require root privileges. The PT tests are run only when RUN_SUDO_TESTS=ON. From what I remember, we found that Github actions would block at the password prompt if we try to run it in the workflow.
> 
> No, that is not correct. The GA VMs have passwordless sudo and we are running `tool.drcachesim.phys-threads_SUDO` today.

Not sure why I'm remembering incorrectly. I also found this comment that suggests that interactive password requests are problematic. When I run locally on my workstation I do get an interactive password request so I assumed that would happen also on the GA runner.
https://github.com/DynamoRIO/dynamorio/blob/3ac43757dd604c770cd54e2c90dce44fa3a77a90/suite/tests/CMakeLists.txt#L57

PT tests are also gated by proc_supports_pt, which isn't true on VMs.

clients/drcachesim/CMakeLists.txt
```diff
@@ -296,6 +296,10 @@ include_directories(${CMAKE_CURRENT_SOURCE_DIR}/tracer)
 include_directories(${CMAKE_CURRENT_SOURCE_DIR}/scheduler)
```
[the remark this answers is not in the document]
abhinav92003: > There are a bunch of PT repos on Github: are they able to automate their tests?

I found a few but they didn't have Github actions configured. Can you share links for the ones you found?

Are these actually collecting traces using PT (which requires root and the PT feature on the processor) or just doing decoding using libipt?

>  Or maybe they set up their own-hosted runner -- if so, we should look into that as it would be best to have automated testing.

If we don't find anything I'll create a new issue (separate from #5505) for this.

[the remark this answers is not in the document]
abhinav92003: Also: I think we considered having a checked-in PT trace so that we could at least have automated tests for the raw2trace part, but that would require checking-in the kcore too which can be ~100MB.

clients/drcachesim/common/trace_entry.h
```diff
@@ -1122,6 +1124,44 @@ typedef enum {
...
+     * DR extensions.
+     */
+    struct {
+        uint16_t cpu_family;
+        uint8_t cpu_model;
+        uint8_t cpu_stepping;
+        uint16_t time_shift;
+        uint32_t time_mult;
+        uint64_t time_zero;
+    } __attribute__((__packed__)) metadata;
```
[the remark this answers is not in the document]
abhinav92003: Done.

clients/drcachesim/common/trace_entry.h
```diff
@@ -1122,6 +1124,44 @@ typedef enum {
...
+
+/**
+ * This is the format in which syscall_pt_trace writes the PT metadata for each
+ * thread before writing any system call's PT data.
+ */
+struct _pt_metadata_buf_t {
+    syscall_pt_entry_t header[PT_METADATA_PDB_HEADER_ENTRY_NUM];
+
+    /**
+     * XXX: This is currently duplicated from pt_metadata_t defined in drpttracer.h.
```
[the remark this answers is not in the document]
abhinav92003: Done.

clients/drcachesim/tracer/syscall_pt_trace.cpp
```diff
@@ -282,8 +285,8 @@ syscall_pt_trace_t::trace_data_dump(drpttracer_output_autoclean_t &output)
 bool
 syscall_pt_trace_t::is_syscall_pt_trace_enabled(IN int sysnum)
 {
-    /* The following syscall's post syscall callback can't be triggered. So we don't
-     * support to recording the kernel PT of them.
+    /* The following syscall's post syscall callback can't be triggered. So we
```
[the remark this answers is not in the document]
abhinav92003: Reworded.

clients/drcachesim/tracer/syscall_pt_trace.cpp
abhinav92003: Not sure. I assumed it was clang-format's doing.

abhinav92003: Reverted

clients/drcachesim/tracer/syscall_pt_trace.cpp
abhinav92003: Not sure. I assumed it was clang-format's doing. Reverted.

clients/drcachesim/tracer/raw2trace.cpp
abhinav92003: Yes. Done.

clients/drcachesim/tracer/raw2trace.cpp
abhinav92003: It goes into the expression for pt_data_size below.

clients/drcachesim/tracer/raw2trace.cpp
abhinav92003: Done.

clients/drcachesim/tracer/raw2trace.cpp
abhinav92003: It was removed.

We read the header and the data together since they always occur together.  I don't see any path where we read just the header only. Further, since pre_read_pt_entries is used locally only, it doesn't need to be in tdata.

clients/drcachesim/tracer/raw2trace.cpp
abhinav92003: Boundary check continues to be here. idx check is just below.

clients/drcachesim/tracer/syscall_pt_trace.cpp
abhinav92003: Not sure. I assumed it was clang-format's doing. Reverted.

clients/drcachesim/tracer/raw2trace.cpp
abhinav92003: Even if we make a change to write the header and data in separate writes, they will still be read back-to-back in code here.

clients/drcachesim/tracer/tracer.cpp
```diff
@@ -1783,14 +1788,15 @@ init_thread_in_process(void *drcontext)
...
-            // finalized the PT output scheme we should pass those parameters.
-            [](const char *fname, uint mode_flags) {
-                return file_ops_func.open_process_file(fname, mode_flags);
+            [](const char *fname, uint mode_flags, thread_id_t thread_id,
+               int64 window_id) {
+                return file_ops_func.call_open_file(fname, mode_flags, thread_id,
+                                                    window_id);
             },
             file_ops_func.write_file, file_ops_func.close_file);
+        DR_ASSERT(ret);
```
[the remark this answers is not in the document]
abhinav92003: Failure to init is FATAL. Changed.

clients/drcachesim/tracer/raw2trace.h
abhinav92003: See other comment. Will add to PR description.

clients/drcachesim/CMakeLists.txt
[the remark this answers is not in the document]
derekbruening: I only did a shallow search so I don't think I saw anything you wouldn't see in a quick search.

For the manual tests you ran: if none of those have memtrace static-linked into the app, could we add such a test, and confirm that it fails on the malloc import issue?

clients/drcachesim/tracer/syscall_pt_trace.cpp
abhinav92003: We're not trying to avoid a 2nd write call for performance reasons. See other comment.

clients/drcachesim/tracer/syscall_pt_trace.cpp
abhinav92003: Based on the implementation of the user-provided write func, it may not be acceptable to separate the header and data away from each other.

clients/drcachesim/tracer/syscall_pt_trace.cpp
abhinav92003: That's probably what happened yeah.

clients/drcachesim/tracer/raw2trace.h
abhinav92003: For this stat, we do want to say syscall, if not syscall_pt. Renamed.

clients/drcachesim/tracer/raw2trace.cpp
abhinav92003: We do need syscall_num and syscall_args_num (when we want to reconstruct the memory address  trace), we need the syscall_idx to ensure we got the correct syscall next (I should change that from an assert to a release build error probably). The pid and tid can potentially be omitted.

How about if I do it in a follow up PR. I'm trying to make this a strictly refactor PR.

clients/drcachesim/tracer/drmemtrace.h
```diff
@@ -285,6 +285,33 @@ DR_EXPORT
...
+drmemtrace_get_kallsyms_path(OUT const char **path);
+
+DR_EXPORT
+/**
+ * Retrieves the full path to the output directory in -offline mode
+ * where kernel PT data is written when kernel tracing is enabled using
+ * -enable_kernel_tracing.
+ */
+drmemtrace_status_t
+drmemtrace_get_kernel_pt_output_path(OUT const char **path);
```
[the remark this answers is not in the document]
abhinav92003: For some of these interfaces like this API and the raw2trace stats and a few others, yes. But for PT specific structures, like pt_data_buf_t and pt_metadata_buf_t, I assume we want to keep them PT-specific right?

clients/drcachesim/common/trace_entry.h
```diff
@@ -1122,6 +1124,44 @@ typedef enum {
...
+     * Figure out the proper code sharing strategy between the drmemtrace client and
+     * DR extensions.
+     */
+    struct {
+        uint16_t cpu_family;
```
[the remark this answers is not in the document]
abhinav92003: I pointed to pt_metadata_t for docx since I didn't want to duplicate it here. Is this sufficient?

clients/drcachesim/common/trace_entry.h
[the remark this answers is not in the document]
abhinav92003: Done.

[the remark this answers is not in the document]
abhinav92003: Done.

clients/drcachesim/tests/offline-kernel-simple.templatex
```diff
@@ -13,7 +13,17 @@ Total counts:
  .* total dcache flushes
  .* total threads
  .* total scheduling markers
-.*
+ .* total transfer markers
```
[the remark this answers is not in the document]
abhinav92003: Done.

clients/drcachesim/tracer/drmemtrace.h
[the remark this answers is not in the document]
abhinav92003: Done.

clients/drcachesim/CMakeLists.txt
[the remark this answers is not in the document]
abhinav92003: I added -enable_kernel_tracing to the client options in burst_static but it didn't fail.

clients/drcachesim/tracer/syscall_pt_trace.cpp
abhinav92003: BUILD_PT_TRACER is always ON for LINUX AND X86 AND X64. If there's no PT support, it'll just fail if the user tries to trace system calls. So the tests should've failed and this is a regression?

abhinav92003: The check seems to be performed in non-static builds. I was able to get the existing kernel test to fail by adding `-raw_compress none` (thanks for the suggestion in our offline discussion). I modified the existing test to use that setting.

clients/drcachesim/tracer/raw2trace.cpp
abhinav92003: Added a TODO for now.

clients/drcachesim/tracer/syscall_pt_trace.cpp
abhinav92003: Added to comment.

clients/drcachesim/CMakeLists.txt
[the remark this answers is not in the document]
abhinav92003: As updated in other comment: I was able to get the existing kernel.simple test fail by adding `-raw_compress none`. So we're testing the use of unsafe functions now.

clients/drcachesim/tracer/raw2trace.cpp
abhinav92003: > I should change that from an assert to a release build error probably

Oh it's already that, not just a DR_ASSERT.

abhinav92003:
x86-64-ubuntu22 api.rseq failure is #6185.

clients/drcachesim/tracer/syscall_pt_trace.cpp
abhinav92003: Wait... I think it should be possible to still make the 3p implementation work with the multiple writes; though it gets a bit more complicated. Let me try and update this thread.

clients/drcachesim/tracer/syscall_pt_trace.cpp
abhinav92003: I refactored 3p code to work with separate writes for header and data.

clients/drcachesim/tracer/syscall_pt_trace.cpp
abhinav92003: I refactored 3p code to work with separate writes for header and data.

This allowed me to remove the heavy std::memcpy here. PTAL.

abhinav92003:
FYI this is ready for re-review.

clients/drcachesim/common/trace_entry.h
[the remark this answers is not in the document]
derekbruening: ?  If this is the same as another named struct: why isn't it just referencing that other struct instead of redefining it?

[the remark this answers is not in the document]
abhinav92003: See the XXX comment on the struct. I'm not sure what's the best sharing strategy between the drmemtrace client and DR extensions.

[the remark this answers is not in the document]
abhinav92003: Following up to our offline discussion: I extracted pt_metadata_t into drpttracer_shared.h and included that in trace_entry.h which solved it.

derekbruening (approved):

clients/drcachesim/common/trace_entry.h
```diff
@@ -1128,24 +1136,20 @@ typedef enum {
...
-    struct {
-        uint16_t cpu_family;
-        uint8_t cpu_model;
-        uint8_t cpu_stepping;
-        uint16_t time_shift;
-        uint32_t time_mult;
-        uint64_t time_zero;
-    } __attribute__((__packed__)) metadata;
-};
+    pt_metadata_t metadata;
```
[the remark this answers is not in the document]
abhinav92003: Done.

clients/drcachesim/common/trace_entry.h
```diff
@@ -967,6 +971,7 @@ struct schedule_entry_t {
 
 #if defined(BUILD_PT_TRACER) || defined(BUILD_PT_POST_PROCESSOR)
 
+/***** Trace entries related to the kernel trace -- start *****/
```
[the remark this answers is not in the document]
abhinav92003: Done.

abhinav92003:
The RISC-V workflow is failing with errors like

E: Unable to locate package libunwind8:riscv64

PR #6383, opened 2023-10-20
i#3544 RV64: Implement link_indirect_exit_arch

ksco:
The implementation of this function is roughly the same as AArch64, except for some additional handling of C extension instructions.

Issue: #3544


core/arch/riscv64/emit_utils.c
```diff
@@ -423,12 +425,65 @@ exit_cti_disp_pc(cache_pc branch_pc)
...
+                  RISCV64_INSTR_COMPRESSED_SIZE);
+
+    pc = get_stub_branch(pc) - 1;
+
+    byte *write_start = vmcode_get_writable_addr((byte *)pc);
+    instrlist_t ilist;
+    instrlist_init(&ilist);
+
+    APP(&ilist,
+        INSTR_CREATE_ld(dcontext, opnd_create_reg(DR_REG_A1),
```
abhinav92003, line 477: Because this is just one instruction, can we just hard-code its encoding here? (Similar to the AArch64 version). It would avoid the extra overhead which would be really good for the linking code.

core/arch/riscv64/emit_utils.c
```diff
@@ -423,12 +425,65 @@ exit_cti_disp_pc(cache_pc branch_pc)
...
+                        opnd_create_base_disp(
+                            dr_reg_stolen, DR_REG_NULL, 0,
+                            get_ibl_entry_tls_offs(dcontext, exit_target), OPSZ_8)));
+
+    instrlist_encode(dcontext, &ilist, write_start, false);
+    instrlist_clear(dcontext, &ilist);
+
+    if (hot_patch)
+        machine_cache_sync(pc, pc + 1, true);
```
abhinav92003, line 486: So we're changing just the one byte at pc? Is the ld instr just one byte?

abhinav92003 (approved):

core/arch/riscv64/emit_utils.c
ksco: `pc` is a `uint *`, so it's 4 bytes. But it doesn't make any difference anyway, on RISC-V, `machine_cache_sync` will flush the i$ on all harts regardless of the range. The argument is just to keep the API consistent with other arches. See #6209.

abhinav92003: > On RISC-V, machine_cache_sync will flush the i$ on all harts regardless of the range.

Add that to a comment here.

ksco: I'm not sure we should do it. We're using `riscv_flush_icache` syscall to implement `machine_cache_sync`, the syscall is taking the `pc_start` and `pc_end` as args, it's just [not used in the kernel currently](https://github.com/torvalds/linux/blob/ce55c22ec8b223a90ff3e084d842f73cfba35588/arch/riscv/kernel/sys_riscv.c#L68), but that might have changed one day.

abhinav92003: Well, in that case we should also pass a range that we really want. [pc, pc + length of instr]

ksco: Yes, we are. The `ld` instr length is 4 bytes, `pc` is a `uint *`, so `pc + 1` is correct.

abhinav92003: Got it, thanks.

PR #6385, opened 2023-10-22
i#3544 RV64: Implement various emit utility procedures

ksco:
Implement various procedures to patch stubs, save and restore registers, etc. At this point, we have implemented all the emit utils needed to run simple programs.

Issue: #3544


core/arch/riscv64/emit_utils.c
```diff
@@ -679,9 +802,50 @@ fill_with_nops(dr_isa_mode_t isa_mode, byte *addr, size_t size)
...
-    ASSERT_NOT_IMPLEMENTED(false);
+    ASSERT_NOT_IMPLEMENTED(!absolute &&
+                           !TEST(SELFPROT_DCONTEXT, dynamo_options.protect_mask));
+
+    instr_t *no_signals = INSTR_CREATE_label(dcontext);
+
+    /* Save callee-saved reg in case we return for a signal. */
+    APP(ilist,
+        XINST_CREATE_move(dcontext, opnd_create_reg(DR_REG_A1),
+                          opnd_create_reg(REG_DCXT)));
```
ksco, line 823: I'm not sure if we need to do this on RISC-V (and also on AArch64) as `REG_DCXT` (`a5` on RISC-V, `x5` on AArch64) is not a callee-saved register. I think we can get rid of this one and the corresponding one below, am I missing something?

core/arch/arch.h
```diff
@@ -188,7 +188,8 @@ mixed_mode_enabled(void)
 #    define SS_RETVAL_REG a0
 #endif /* X86/ARM/RISCV64 */
 #define XSP_OFFSET ((MC_OFFS) + (offsetof(priv_mcontext_t, xsp)))
-#define XFLAGS_OFFSET ((MC_OFFS) + (offsetof(priv_mcontext_t, xflags)))
+#define XFLAGS_OFFSET \
+    ((MC_OFFS) + (offsetof(priv_mcontext_t, IF_RISCV64_ELSE(fcsr, xflags))))
```
ivankyluk, line 192: It will be more consistent to have XFLAGS_OFFSET defined under #ifdef X86, AARCHXX,  and RISCV64 separately,  like SS_RETVAL_REG or  REGo_OFFSET above.

core/arch/arch_exports.h
```diff
@@ -234,6 +234,7 @@ typedef struct _local_state_extended_t {
 #    define TLS_REG2_SLOT ((ushort)offsetof(spill_state_t, a2))
 #    define TLS_REG3_SLOT ((ushort)offsetof(spill_state_t, a3))
 #    define TLS_REG_STOLEN_SLOT ((ushort)offsetof(spill_state_t, reg_stolen))
+#    define TLS_REG_TP_SLOT os_get_app_tls_base_offset(TLS_REG_LIB)
```
ivankyluk, line 237: This define is used only in core/arch/riscv64/emit_utils.c, right? 

Let's follow the examples in core/arch/aarchxx/mangle.c, and use os_get_app_tls_base_offset(TLS_REG_LIB) in emit_utils.c there instead.

Otherwise, we are adding an unnecessary dependency on unix/os_export.h

core/arch/riscv64/emit_utils.c
```diff
@@ -340,8 +340,35 @@ stub_is_patched(dcontext_t *dcontext, fragment_t *f, cache_pc stub_pc)
...
+     */
+    if (stub_is_patched_for_intermediate_fragment_link(dcontext, stub_pc)) {
+        /* Restore the sd a0, offs(reg_stolen), see insert_exit_stub_other_flags().
+         * Format of the sd instruction:
+         *  | imm[11:5] |  rs2  |  rs1  |011| imm[4:0] |0100011|
+         *  ^   31-25   ^ 24-20 ^ 19-15 ^   ^   11-7   ^
+         */
+        ASSERT(TLS_REG0_SLOT <= (2 << 11) - 1);
+        *(uint *)vmcode_get_writable_addr(stub_pc) =
+            (0x3023 | TLS_REG0_SLOT >> 5 << 25 | (DR_REG_A0 - DR_REG_ZERO) << 20 |
```
ivankyluk, line 357: please add a #define for 0x3023, and add comment to describe what it is.

core/arch/riscv64/emit_utils.c
```diff
@@ -340,8 +340,35 @@ stub_is_patched(dcontext_t *dcontext, fragment_t *f, cache_pc stub_pc)
...
+             (dr_reg_stolen - DR_REG_ZERO) << 15 | (TLS_REG0_SLOT & 0x1f) << 7);
+        if (hot_patch)
+            machine_cache_sync(stub_pc, stub_pc + 4, true);
```
ivankyluk, line 360: nit: /*flush_icache=*/true

core/arch/riscv64/emit_utils.c
```diff
@@ -340,8 +340,35 @@ stub_is_patched(dcontext_t *dcontext, fragment_t *f, cache_pc stub_pc)
...
+    } else if (stub_is_patched_for_far_fragment_link(dcontext, f, stub_pc)) {
+        /* Restore the data slot to fcache return address. */
+        /* RISCV64 uses shared gencode. So, fcache_return routine address should be
+         * same, no matter which thread creates/unpatches the stub.
```
ivankyluk, line 364: nit: should be the same

core/arch/riscv64/emit_utils.c
```diff
@@ -540,8 +574,28 @@ fragment_ibt_prefix_size(uint flags)
 void
 insert_fragment_prefix(dcontext_t *dcontext, fragment_t *f)
 {
-    /* FIXME i#3544: Not implemented */
-    ASSERT_NOT_IMPLEMENTED(false);
+    /* Always use prefix on RISCV64 as there is no load to PC. */
+    byte *write_start = vmcode_get_writable_addr(f->start_pc);
+    byte *pc = write_start;
+    ASSERT(f->prefix_size == 0);
```
ivankyluk, line 580: nit: move the ASSERT to the top of the function.

core/arch/riscv64/emit_utils.c
```diff
@@ -540,8 +574,28 @@ fragment_ibt_prefix_size(uint flags)
...
+            opnd_create_base_disp(dr_reg_stolen, DR_REG_NULL, 0, TLS_REG0_SLOT, OPSZ_8)));
+    APP(&ilist,
+        INSTR_CREATE_ld(
+            dcontext, opnd_create_reg(DR_REG_A1),
+            opnd_create_base_disp(dr_reg_stolen, DR_REG_NULL, 0, TLS_REG1_SLOT, OPSZ_8)));
+
+    pc = instrlist_encode(dcontext, &ilist, pc, false);
+    instrlist_clear(dcontext, &ilist);
+
+    f->prefix_size = (byte)(((cache_pc)pc) - write_start);
```
ivankyluk, line 597: Both pc, and write_start are byte*, what's the reason to cast pc to cache_pc?

core/arch/riscv64/emit_utils.c
```diff
@@ -558,32 +612,99 @@ append_call_exit_dr_hook(dcontext_t *dcontext, instrlist_t *ilist, bool absolute
 void
 append_restore_xflags(dcontext_t *dcontext, instrlist_t *ilist, bool absolute)
 {
-    /* FIXME i#3544: Not implemented */
-    ASSERT_NOT_IMPLEMENTED(false);
+    APP(ilist, RESTORE_FROM_DC(dcontext, DR_REG_A0, XFLAGS_OFFSET));
+    APP(ilist,
+        INSTR_CREATE_csrrw(dcontext, opnd_create_reg(DR_REG_X0),
+                           opnd_create_reg(DR_REG_A0),
+                           opnd_create_immed_int(/*fcsr*/ 0x003, OPSZ_12b)));
```
ivankyluk, line 619: nit: /*fcsr=*/

core/arch/riscv64/emit_utils.c
```diff
@@ -558,32 +612,99 @@ append_call_exit_dr_hook(dcontext_t *dcontext, instrlist_t *ilist, bool absolute
...
 }
 
 /* dcontext is in REG_DCXT; other registers can be used as scratch.
  */
 void
 append_restore_simd_reg(dcontext_t *dcontext, instrlist_t *ilist, bool absolute)
 {
-    /* FIXME i#3544: Not implemented */
-    ASSERT_NOT_IMPLEMENTED(false);
+    /* Nothing. */
```
ivankyluk, line 627: How about "no-op" instead of Nothing?

core/arch/riscv64/emit_utils.c
```diff
@@ -558,32 +612,99 @@ append_call_exit_dr_hook(dcontext_t *dcontext, instrlist_t *ilist, bool absolute
...
+        ASSERT_NOT_IMPLEMENTED(false);
+    }
+
+    /* a0 and a1 will always have been saved in TLS slots before executing
+     * the code generated here. See, for example:
+     * emit_do_syscall_common, emit_indirect_branch_lookup, handle_sigreturn,
+     * insert_exit_stub_other_flags, execute_handler_from_{cache,dispatch},
+     * transfer_from_sig_handler_to_fcache_return
+     */
+    for (int i = 1; i < 32; i++) {
```
ivankyluk, line 678: An alternative would be

for (int reg = DR_REG_X0 + 1; reg < DR_REG_X0 + 32; reg++) {

and use reg instead of DR_REG_X0 below.

core/arch/riscv64/emit_utils.c
```diff
@@ -592,24 +713,26 @@ append_save_gpr(dcontext_t *dcontext, instrlist_t *ilist, bool ibl_end, bool abs
 void
 append_save_simd_reg(dcontext_t *dcontext, instrlist_t *ilist, bool absolute)
 {
-    /* FIXME i#3544: Not implemented */
-    ASSERT_NOT_IMPLEMENTED(false);
+    /* Nothing. */
```
ivankyluk, line 716: no-op

core/arch/riscv64/emit_utils.c
```diff
@@ -592,24 +713,26 @@ append_save_gpr(dcontext_t *dcontext, instrlist_t *ilist, bool ibl_end, bool abs
...
 /* Scratch reg0 is holding exit stub. */
 void
 append_save_clear_xflags(dcontext_t *dcontext, instrlist_t *ilist, bool absolute)
 {
-    /* FIXME i#3544: Not implemented */
-    ASSERT_NOT_IMPLEMENTED(false);
+    APP(ilist,
+        INSTR_CREATE_csrrs(dcontext, opnd_create_reg(DR_REG_A1),
+                           opnd_create_reg(DR_REG_X0),
+                           opnd_create_immed_int(/*fcsr*/ 0x003, OPSZ_12b)));
```
ivankyluk, line 726: /*fcsr=*/

core/arch/riscv64/emit_utils.c
```diff
@@ -679,9 +802,50 @@ fill_with_nops(dr_isa_mode_t isa_mode, byte *addr, size_t size)
     return true;
 }
 
+/* Having only one thread register tp shared between app and DR, we steal a register for
+ * DR's TLS base in the code cache, and store DR's TLS base into an private lib's TLS slot
```
ivankyluk, line 806: an private -> a private

core/ir/riscv64/decode.c
```diff
@@ -121,33 +121,21 @@ decode_raw(dcontext_t *dcontext, byte *pc, instr_t *instr)
 bool
 decode_raw_is_jmp(dcontext_t *dcontext, byte *pc)
 {
-    /* FIXME i#3544: Not implemented */
-    ASSERT_NOT_IMPLEMENTED(false);
-    return false;
+    return ((*(uint *)pc & 0x7f) == 0x6f); /* JAL */
```
ivankyluk, line 124: Is there a #define or constant for JAL = 0x6f that can be used here?

core/ir/riscv64/decode.c
```diff
@@ -121,33 +121,21 @@ decode_raw(dcontext_t *dcontext, byte *pc, instr_t *instr)
...
-    return false;
+    /* Format of the J-type instruction:
+     * |   31    |30       21|   20    |19        12|11   7|6      0|
+     * | imm[20] | imm[10:1] | imm[11] | imm[19:12] |  rd  | opcode |
+     *  ^------------------------------------------^
+     */
+    uint enc = *(uint *)pc;
+    int32_t imm = (((enc >> 31) & 1) << 20) | (((enc >> 12) & 0xff) << 12) |
+        (((enc >> 20) & 1) << 11) | (((enc >> 21) & 0x3ff) << 1);
+    return pc + ((imm << 11) >> 11);
```
ivankyluk, line 138: The most significant 11 bits of imm should already be zero, right? So << 11, follow by >> 11 is not necessary?

core/arch/arch.h
ksco: Done.

core/arch/arch_exports.h
ksco: It will be used in several functions in riscv64/mangle.c for mangling the `tp` register.

core/arch/riscv64/emit_utils.c
ksco: The comment above should explain what 0x3023 is. If I were to add a #define, what should I call it?

core/arch/riscv64/emit_utils.c
ksco: Done.

core/arch/riscv64/emit_utils.c
ksco: Done.

core/arch/riscv64/emit_utils.c
ksco: Done.

core/arch/riscv64/emit_utils.c
ksco: I think it's copied from the other 3 arches... fixed.

core/ir/riscv64/decode.c
ksco: Currently, no. Should we #define one?

core/ir/riscv64/decode.c
ksco: Ah, imm here is a signed integer, so it's doing sign extension.

core/arch/riscv64/emit_utils.c
ksco: Done.

ksco (commented):
Thanks for the detailed review!

core/arch/arch_exports.h
ivankyluk: I think we should avoid adding the dependency on core/unix/os_exports.h.

core/ir/riscv64/decode.c
ivankyluk: Isn't it OP_jal, like the one in core/ir/riscv64/instr.c?

core/ir/riscv64/decode.c
ivankyluk: I see, thanks.

core/ir/riscv64/decode.c
ksco: well, OP_jal is an enum, it's used in the IR, and `x06f` here is a raw instruction pattern.

core/arch/arch_exports.h
ksco: > I think we should avoid adding the dependency on core/unix/os_exports.h.

Sure, fixed in https://github.com/DynamoRIO/dynamorio/pull/6385/commits/2f6542391479336ee09214a4befd85e5a419b25f. Also could you explain why? I don't quite follow, but I'd like to avoid this kind of issue in future patches.

core/arch/riscv64/emit_utils.c
ivankyluk: I raised the question to Derek, his answer is yes. 

Here's his answer:

yes on that return path since it is all code we control only care about not clobbering callee-saved (when app code gets involved we do not assume anything about ABI, plus we interrupt at block boundaries, so usually in DR gencode we do not care about ABI conventions, which is maybe why this one is confusing)

core/arch/arch_exports.h
ivankyluk: os_get_app_tls_base_offset is defined in core/unix/os_exports.h, and implemented in core/unix/os.c.

By adding os_get_app_tls_base_offset here, all files with RISCV64 defined which include arch_exports.h have to include os_exports.h, and link with the lib including os.c.

While it's the case today, i.e. RISCV64 is used only with UNIX, it'll be hard to decouple later if the situation changes in the future.

core/arch/riscv64/emit_utils.c
ivankyluk: I'd call it something like SD_INSTR_MASK. I'll leave it to you. It's fine to leave it as is.

ivankyluk (approved):

core/arch/arch_exports.h
ksco: Thanks!

PR #6386, opened 2023-10-22
i#3544 RV64: Implement insert_{push,pop}_all_registers procedures

ksco:
This patch implements insert_push_all_registers() and insert_pop_all_registers(), which are used when executing simple programs.

Issue: #3544


derekbruening:
@abhinav92003 maybe this was overlooked?  @ksco please ping us in such cases.

core/arch/riscv64/mangle.c
```diff
@@ -32,40 +32,170 @@
...
+    max_offs = get_clean_call_switch_stack_size();
+
+    PRE(ilist, instr,
+        INSTR_CREATE_addi(dcontext, opnd_create_reg(DR_REG_SP),
+                          opnd_create_reg(DR_REG_SP),
+                          opnd_add_flags(opnd_create_immed_int(-max_offs, OPSZ_12b),
+                                         DR_OPND_IMM_PRINT_DECIMAL)));
+
+    /* Push GPRs. */
+    for (int i = 1; i < DR_NUM_GPR_REGS; i += 1) {
```
derekbruening, line 80: nit: i++ is the pattern used in all existing C code; the += 1 jumps out and makes it seem like there is something unusual here.  Best to match conventional style.

Ditto below.

core/arch/riscv64/mangle.c
```diff
@@ -32,40 +32,170 @@
...
+        if (cci->reg_skip[i])
+            continue;
+
+        PRE(ilist, instr,
+            INSTR_CREATE_sd(dcontext,
+                            opnd_add_flags(opnd_create_base_disp(
+                                               DR_REG_SP, DR_REG_NULL, 0,
+                                               dstack_offs + i * sizeof(reg_t), OPSZ_8),
```
derekbruening, line 88: `sizeof(reg_t)` here but `XSP_SZ` on line 93: use the same one for both (I guess `XSP_SZ` is what the code on other arches is using)

core/arch/riscv64/mangle.c
```diff
@@ -32,40 +32,170 @@
...
+
+    dstack_offs += DR_NUM_GPR_REGS * XSP_SZ;
+
+    if (opnd_is_immed_int(push_pc)) {
+        PRE(ilist, instr,
+            XINST_CREATE_load_int(dcontext, opnd_create_reg(DR_REG_A0), push_pc));
+    } else {
+        ASSERT(opnd_is_reg(push_pc));
+        reg_id_t push_pc_reg = opnd_get_reg(push_pc);
+        /* push_pc opnd is already pushed on the stack. */
```
derekbruening, line 101: But the GPR itself still holds the value, right?  We don't need to load from the stack: we can directly use the reg, right?

core/arch/riscv64/mangle.c
```diff
@@ -32,40 +32,170 @@
...
+
+    dstack_offs += XSP_SZ;
+
+    /* Push FPRs. */
+    for (int i = 0; i < DR_NUM_FPR_REGS; i += 1) {
+        PRE(ilist, instr,
+            INSTR_CREATE_fsd(dcontext,
+                             opnd_add_flags(opnd_create_base_disp(
+                                                DR_REG_SP, DR_REG_NULL, 0,
+                                                dstack_offs + i * sizeof(reg_t), OPSZ_8),
```
derekbruening, line 119: As above: more consistent to use `s/sizeof(reg_t)/XSP_SZ/`

core/arch/riscv64/mangle.c
```diff
@@ -32,40 +32,170 @@
...
+                           opnd_create_immed_int(0x003 /* fcsr */, OPSZ_12b)));
+
+    PRE(ilist, instr,
+        INSTR_CREATE_sd(dcontext, OPND_CREATE_MEM64(DR_REG_SP, dstack_offs),
+                        opnd_create_reg(DR_REG_A0)));
+
+    dstack_offs += XSP_SZ;
+
+    /* Keep mcontext shape. */
+    dstack_offs += (proc_num_simd_registers() * sizeof(dr_simd_t));
```
derekbruening, line 140: Wait, is pushing the SIMD registers not yet implemented?  Please add a TODO comment for that so it's clear something is missing.

core/arch/riscv64/mangle.c
```diff
@@ -32,40 +32,170 @@
...
+
+    current_offs -= DR_NUM_FPR_REGS * XSP_SZ;
+
+    /* Pop FPRs. */
+    for (int i = 0; i < DR_NUM_FPR_REGS; i += 1) {
+        PRE(ilist, instr,
+            INSTR_CREATE_fld(dcontext, opnd_create_reg(DR_REG_F0 + i),
+                             opnd_add_flags(opnd_create_base_disp(
+                                                DR_REG_SP, DR_REG_NULL, 0,
+                                                current_offs + i * sizeof(reg_t), OPSZ_8),
```
derekbruening, line 178: Same as above; same on line 196 too.

core/arch/riscv64/mangle.c
```diff
@@ -32,40 +32,170 @@
...
+
+    PRE(ilist, instr,
+        INSTR_CREATE_ld(dcontext, opnd_create_reg(DR_REG_A0),
+                        OPND_CREATE_MEM64(DR_REG_SP, current_offs)));
+
+    /* csrw a0, fcsr */
+    PRE(ilist, instr,
+        INSTR_CREATE_csrrw(dcontext, opnd_create_reg(DR_REG_X0),
+                           opnd_create_reg(DR_REG_A0),
+                           opnd_create_immed_int(/*fcsr*/ 0x003, OPSZ_12b)));
```
derekbruening, line 168: 2nd time 0x003 is used: add a named constant for this file, even if temporary.  Put the same FIXME by its definition for longer-term solution.

derekbruening (approved):

ksco:
> @abhinav92003 maybe this was overlooked? @ksco please ping us in such cases.

Okay. I was prepared to wait a few more days. 😄

derekbruening:
> > @abhinav92003 maybe this was overlooked? @ksco please ping us in such cases.
> 
> Okay. I was prepared to wait a few more days. 😄

Thank you for your patience, though generally more than 2 days is worth a ping (among those directly working on the same feature, should be even faster, though that is not the case here).

core/arch/riscv64/mangle.c
ksco: Yes, I've changed the code from `load+store` to `store push_pc offs(sp)`.

core/arch/riscv64/mangle.c
ksco: Done, also changed  `emit_utils.c` which uses `0x003` several times too. (I think a better way is to represent CSRs as registers than number literals, also added this to the comment).

abhinav92003:
> @abhinav92003 maybe this was overlooked? @ksco please ping us in such cases.

Apologies -- slipped through the cracks.

PR #6388, opened 2023-10-22
i#3544 RV64: Implement mangle_rel_addr()

ksco:
JAL is the only instruction in RISC-V with PC-relative address reference, which needs to be mangled. Loading a PC-relative address into the thread pointer (tp) register should be rare and we left it unimplemented for now.


core/arch/riscv64/mangle.c
```diff
@@ -32,6 +32,12 @@
 
 #include "../globals.h"
 #include "arch.h"
+#include "instr_create_shared.h"
```
bete0, line 35: Do we need to comment the reason for including this header?

core/arch/riscv64/mangle.c
```diff
@@ -32,6 +32,12 @@
...
+#include "instrument.h" /* instrlist_meta_preinsert */
+/* Make code more readable by shortening long lines.
+ * We mark everything we add as non-app instr.
+ */
+#define PRE instrlist_meta_preinsert
```
bete0, line 40: I feel like seeing `PRE` down below isn't clear enough. Can you maybe make it more descriptive?

bete0 (approved):

bete0:
Looks like code_api|api.rseq test failed bu that might be https://github.com/DynamoRIO/dynamorio/issues/6185.

core/arch/riscv64/mangle.c
ksco: Ah, that header is actually not needed for this PR. It's included for the `INSTR_*` and `XINST_*` macros, but this PR does not need these macros. I remove this line in https://github.com/DynamoRIO/dynamorio/pull/6388/commits/0403d68fc29718a11a2332d5c163b937f97246e4.

core/arch/riscv64/mangle.c
ksco: You mean change `PRE` to `PREINSERT` or something? The `PRE` and `POST` naming convention is used everywhere, so I think it might be better to stay consistent.

ksco:
> Looks like code_api|api.rseq test failed bu that might be #6185.

Yes, it's an unstable test.

ksco:
@bete0 Can this be merged now? There is still one unresolved conversation so I'm not sure.

ksco:
Merging this as it’s already approved.

Issue #6394, opened 2023-10-25
Incorrect function return address for tailcall when -record_replace_retaddr flag is used.


ivankyluk:
When drcachesim is used with  -offline, -record_replace_retaddr, -record_function, it leaves sentinels in the trace when a tailcall to a function which is being recorded is encountered.

The issue can be reproduced with common.getretaddr under build_suite as follows:

./build/bin64/drrun -t drcachesim -offline -record_replace_retaddr -record_function 'tailcall_with_retaddr|1&foo|1' -- ../build_suite/build_debug-internal-64/suite/tests/bin/common.getretaddr

 <marker: function #0>
 <marker: function return address PC>
 <marker: function argument ARG1>

<marker: function #1>
<marker: function return address SENTINEL>  
<marker: function argument ARG2>
     
The return address of function #1 should be PC, the same as function #0. 
The current implementation leaves the SENTINEL in the trace instead.

PR #6395, opened 2023-10-25
i#6394: Replace sentinel in trace with real return address.

ivankyluk:
Update drwrap to check if the return address is a sentinel. If it is, look for the real return address at the outer level.

When  -record_replace_retaddr is used in X86 platforms, the return address in the stack is replaced by the sentinel.
For a tailcall, the current implementation uses the address in the stack, the sentinel, as the return address. The change is to check if the return address in the stack is the sentinel or not. If it is, replace it with the return address of the outer level.

The problem can be reproduced by 

./build/bin64/drrun -t drcachesim -offline -record_replace_retaddr -record_function 'tailcall_with_retaddr|1&foo|1' -- ../build_suite/build_debug-internal-64/suite/tests/bin/common.getretaddr

A test is added to cover -record_replace_retaddr with common.getretaddr which has a tailcall.

Fixes: #6394


ext/drwrap/drwrap.c
```diff
@@ -2020,8 +2020,21 @@ drwrap_in_callee(void *arg1, reg_t xsp _IF_NOT_X86(reg_t lr))
...
-                        IF_X86_ELSE(get_retaddr_from_stack(xsp), (app_pc)lr));
+    app_pc retaddr = IF_X86_ELSE(get_retaddr_from_stack(xsp), (app_pc)lr);
+#ifdef X86
+    if (TEST(DRWRAP_REPLACE_RETADDR, global_flags)) {
+        /* In case of a tailcall for X86, the return address has already been replaced by
+         * the sentinel in the stack, we need to retrieve the return address from the
+         * outer level.
+         */
+        if (retaddr == (app_pc)replace_retaddr_sentinel) {
+            retaddr = pt->retaddr[pt->wrap_level];
```
derekbruening, line 2031: wrap_level starts out -1.  ASSERT that it is >= 0.

ext/drwrap/drwrap.c
```diff
@@ -2782,6 +2795,14 @@ drwrap_get_retaddr_if_sentinel(void *drcontext, INOUT app_pc *possibly_sentinel)
     /* If we see the sentinel, we must be inside a wrapped function. */
     ASSERT(pt != NULL && pt->wrap_level >= 0, "Invalid drwrap state.");
     *possibly_sentinel = pt->retaddr[pt->wrap_level];
+    /* In case of a tailcall, the return address has already been replaced by
```
derekbruening, line 2798: Wait: I thought if we replaced in the tailcall entry, we did not need this?  This can be removed, right?  Replaced by an assert that the value is not the sentinel?

derekbruening (approved):

ext/drwrap/drwrap.c
ivankyluk: done.

ext/drwrap/drwrap.c
ivankyluk: done.

PR #6398, opened 2023-10-27
i#5036 A64 scatter/gather, part 5: Expand scalar+immediate

jackgallagher-arm:
Adds support to drx_expand_scatter_gather() for SVE scalar+immediate predicated contiguous load and store instructions, along with tests.

Issue: #5036


ext/drx/scatter_gather_aarch64.c
```diff
@@ -595,12 +595,36 @@ expand_contiguous(void *drcontext, instrlist_t *bb, instr_t *sg_instr,
...
+            /* The displacement is 0 so the original base register already contains the
+             * base of the contiguous memory region.
+             */
+            new_base = sg_info->base_reg;
+        } else if (sg_info->disp > 0) {
+            /* add      new_base, base_reg, #disp */
+            EMIT(add, opnd_create_reg(new_base), opnd_create_reg(sg_info->base_reg),
+                 OPND_CREATE_INT(sg_info->disp));
+        } else {
+            /* sub      new_base, base_reg, #-disp */
```
abhinav92003, line 615: It's subtracting the negative displacement; double negatives turning it into addition? Is there a correctness issue?

abhinav92003 (approved):

ext/drx/scatter_gather_aarch64.c
jackgallagher-arm: > It's subtracting the negative displacement; double negatives turning it into addition?

In a way.

The `disp` value is a signed displacement that is added to the base, however the AArch64 `add` and `sub` instructions use an unsigned immediate value so we can't use `add` to add a negative `disp`.

For example, if `disp` was `-4`, what we want to do is:
```
add   x0, x1, #-4 ; Invalid instruction
```
but we can't so we turn it into
```
sub   x0, x1, #4 ; Subtracting 4 is the same as adding -4
```

PR #6403, opened 2023-11-01
i#5036 A64 scatter/gather, part 6: Expand multi-register instrs

jackgallagher-arm:
Adds support to drx_expand_scatter_gather() for SVE multi-register scalar+scalar and scalar+immediate predicated contiguous load and store instructions, along with tests.

Issue: #5036


clients/drcachesim/tracer/raw2trace.h
```diff
@@ -961,8 +961,13 @@ class raw2trace_t {
...
      * The trace_entry_t buffer returned by get_write_buffer() is assumed to be at least
      * #WRITE_BUFFER_SIZE large.
+     *
+     * Some of the AArch64 SVE scatter/gather instructions have a lot of memref entries.
+     * For example ld4b loads 4 registers with byte sized elements, so that is
+     *     (vl_bits / 8) * 4
+     * entries. With a 512-bit vector length that is (512 / 4 * 8) = 256 memref entries.
      */
-    static const uint WRITE_BUFFER_SIZE = 64;
+    static const uint WRITE_BUFFER_SIZE = 260;
```
jackgallagher-arm, line 970: While testing drcachesim offline I found that we were seeing the `"Too many entries"` error from `raw2trace.cpp:1825` on some of the larger multi-vector register instructions. Increasing the `WRITE_BUFFER_SIZE` seems to fix the issue; is that the right solution?

derekbruening: IIRC the buffer only needs to handle records for a single instruction.  But a scatter/gather can have many records so maybe it is not surprising it hit the limit?  But do the SVE ones have more records than x86 ones which did not hit the limit?  @abhinav92003 did you hit this issue on x86?

Yes, if it's slightly larger, increasing makes sense.  I assume it's not 10x larger.

abhinav92003: > @abhinav92003 did you hit this issue on x86?

On x86 I did not hit this issue on raw2trace. I did hit a similar issue in the tracer though: #6313.

jackgallagher-arm: It sounds like increasing the buffer size in an appropriate solution.

jackgallagher-arm:
@derekbruening @abhinav92003, could you take a look at this PR when you have an opportunity. Thanks.

abhinav92003:
> @derekbruening @abhinav92003, could you take a look at this PR when you have an opportunity. Thanks.

Wow this one seems to have been pending from my side for 2 weeks. Apologies for the delay. I'll get this done by today  or latest tomorrow (since it's a bit on the long side).

ext/drx/scatter_gather_aarch64.c
```diff
@@ -419,46 +423,54 @@ expand_scatter_gather(void *drcontext, instrlist_t *bb, instr_t *sg_instr,
...
 
+    DR_ASSERT_MSG(
+        (scalar_src_or_dst != scalar_index_or_base) ||
+            (sg_info->is_load && sg_info->reg_count == 1),
+        "Single register loads are the only instructions where it is safe for the scratch"
+        " registers to alias");
+
     const uint no_of_elements =
         (opnd_size_in_bytes(sg_info->scatter_gather_size) / sg_info->reg_count) /
         opnd_size_in_bytes(sg_info->scalar_value_size);
```
abhinav92003, line 435: Add a loop-top comment that here we're trying to zero what we can in the dest reg.

ext/drx/scatter_gather_aarch64.c
```diff
@@ -419,46 +423,54 @@ expand_scatter_gather(void *drcontext, instrlist_t *bb, instr_t *sg_instr,
...
-                 opnd_create_reg_element_vector(sg_info->gather_dst_reg,
-                                                sg_info->element_size),
-                 opnd_create_predicate_reg(scratch_pred, true), OPND_CREATE_INT8(0),
-                 opnd_create_immed_uint(0, OPSZ_1b));
-        } else {
-            /* We don't care about any values in the dst register so zero the whole thing.
+    for (uint reg_index = 0; reg_index < sg_info->reg_count; reg_index++) {
+        const reg_id_t vector_dst =
+            (((sg_info->gather_dst_reg - DR_REG_Z0) + reg_index) % 32) + DR_REG_Z0;
+        if (sg_info->is_load) {
```
abhinav92003, line 439: Looks like we can avoid the loop completely for !is_load if we move the condition before the loop.

ext/drx/scatter_gather_aarch64.c
```diff
@@ -419,46 +423,54 @@ expand_scatter_gather(void *drcontext, instrlist_t *bb, instr_t *sg_instr,
...
+            /* First we deal with the inactive elements. Gather loads are always zeroing
+             * so we need to set all inactive elements to 0.
              */
-
-            /* dup      gather_dst_reg.element_size, #0, lsl #0 */
-            EMIT(dup_sve_shift,
-                 opnd_create_reg_element_vector(sg_info->gather_dst_reg,
-                                                sg_info->element_size),
-                 OPND_CREATE_INT8(0), opnd_create_immed_uint(0, OPSZ_1b));
+            if ((sg_info->base_reg == vector_dst) || (sg_info->index_reg == vector_dst)) {
```
abhinav92003, line 443: If it's not too much refactoring could we move the iteration logic into a separate routine?

ext/drx/scatter_gather_aarch64.c
```diff
@@ -481,47 +493,60 @@ expand_scatter_gather(void *drcontext, instrlist_t *bb, instr_t *sg_instr,
...
-                 opnd_create_reg_element_vector(sg_info->scatter_src_reg,
-                                                sg_info->element_size));
-
-            /* str[bh]  scalar_src, [mem] */
-            instrlist_preinsert(
-                bb, sg_instr,
-                INSTR_XL8(create_scalar_load_or_store(drcontext, sg_info,
-                                                      scalar_index_or_base, scalar_src),
-                          orig_app_pc));
+            if (sg_info->is_load) {
```
abhinav92003, line 513: If it's not too much refactoring, could we move the iteration logic from this point onwards into its own separate routine?

ext/drx/scatter_gather_aarch64.c
```diff
@@ -630,10 +653,10 @@ expand_contiguous(void *drcontext, instrlist_t *bb, instr_t *sg_instr,
      * time.
```
abhinav92003, line 653: Update this comment too.

ext/drx/scatter_gather_aarch64.c
```diff
@@ -630,10 +653,10 @@ expand_contiguous(void *drcontext, instrlist_t *bb, instr_t *sg_instr,
...
      */
 
-    /* index    scratch_vec.element_size, #0, #1 */
+    /* index    scratch_vec.element_size, #0, #reg_count */
     EMIT(index_sve, opnd_create_reg_element_vector(scratch_vec, sg_info->element_size),
```
abhinav92003, line 657: So for ld4b, the first vector gets the first byte of all elements, second vector gets the second byte of all elements and so on till the fourth vector (where each element is 4 byte).

Curious to know a use case where one'd want to load bytes like that. Do you have an example?

ext/drx/scatter_gather_aarch64.c
```diff
@@ -481,47 +493,60 @@ expand_scatter_gather(void *drcontext, instrlist_t *bb, instr_t *sg_instr,
...
+                    reg_is_z(sg_info->base_reg) ? sg_info->base_reg : sg_info->index_reg;
+
+                /* lastb    scalar_index_or_base, scratch_pred,
+                 * reg_to_extract.element_size */
+                EMIT(lastb_sve_scalar, opnd_create_reg(scalar_index_or_base),
+                     opnd_create_reg(scratch_pred),
+                     opnd_create_reg_element_vector(reg_to_extract,
+                                                    sg_info->element_size));
+            } else {
+                // Increment index
```
abhinav92003, line 508: Ah this is where the magic happens where the ith byte is loaded to the ith vector. The comment doesn't do it justice. Please elaborate.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -931,8 +931,13 @@ class raw2trace_t {
     /**
      * The trace_entry_t buffer returned by get_write_buffer() is assumed to be at least
      * #WRITE_BUFFER_SIZE large.
+     *
+     * Some of the AArch64 SVE scatter/gather instructions have a lot of memref entries.
+     * For example ld4b loads 4 registers with byte sized elements, so that is
+     *     (vl_bits / 8) * 4
+     * entries. With a 512-bit vector length that is (512 / 4 * 8) = 256 memref entries.
```
abhinav92003, line 938: Do we want this to be large enough to fit one instr with its memrefs and that's why we've set it to 256 + 1 rounded up to 260? Maybe add to comment that this should be worth the largest instr + its memrefs.

suite/tests/client-interface/drx-scattergather-aarch64.cpp
```diff
@@ -30,6 +30,7 @@
  * DAMAGE.
  */
 
+#include <algorithm>
```
abhinav92003, line 33: I feel it's not easy to review the test values very closely. I think one way to easily catch test value bugs would be to run the test natively also (in addition to under DR and drmemtrace). Native runs would reveal correctness issues with assertion values. Could we add a torunonly_native test?

We do native run tests also when we want to verify our assumptions of kernel behavior.

abhinav92003 (approved):

ext/drx/scatter_gather_aarch64.c
jackgallagher-arm: It can be used to load an array of struct-like data. The struct gets deinterleaved so each register gets the values of a particular member.

```
struct Rgba {
    uint8_t r;
    uint8_t g;
    uint8_t b;
    uint8_t a;
};

A rgba[num];
```

Load with:
```ld4b {z0, z1, z2, z3}, pg/z [rgba]```

Now `z0` has all `r` values, `z1` has all `g` values, `z2` has all `b` values, `z3` has all `a` values.

File: ext/drx/scatter_gather_aarch64.c [hunk 11]
```diff
@@ -481,47 +493,60 @@ expand_scatter_gather(void *drcontext, instrlist_t *bb, instr_t *sg_instr,
...
-            /* str[bh]  scalar_src, [mem] */
-            instrlist_preinsert(
-                bb, sg_instr,
-                INSTR_XL8(create_scalar_load_or_store(drcontext, sg_info,
-                                                      scalar_index_or_base, scalar_src),
-                          orig_app_pc));
+            if (sg_info->is_load) {
+                const reg_id_t scalar_dst = scalar_src_or_dst;
+                const reg_id_t vector_dst =
+                    (((sg_info->gather_dst_reg - DR_REG_Z0) + reg_index) % 32) +
```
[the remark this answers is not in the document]
jackgallagher-arm: Turns out there is already `DR_NUM_SIMD_VECTOR_REGS` so I can use that.

ext/drx/scatter_gather_aarch64.c
jackgallagher-arm: I just spotted that too. Done.

suite/tests/client-interface/drx-scattergather-aarch64.cpp
jackgallagher-arm: Done. I have been testing the test app by running it natively as I go using 128-bit, 256-bit, and 512-bit vector lengths (all the sizes supported by DynamoRIO) so I know the reference data is correct, but its still good to add a `torunonly_native(...)` test to prove it stays correct.

clients/drcachesim/tracer/raw2trace.h
jackgallagher-arm: Done.

ext/drx/scatter_gather_aarch64.c
jackgallagher-arm: Done.

ext/drx/scatter_gather_aarch64.c
jackgallagher-arm: Done.

ext/drx/scatter_gather_aarch64.c
jackgallagher-arm: I think this function is getting a little too complicated trying to handle all these different kinds of instruction in one place. I want to try a bigger refactor to try and make this code a bit clearer but I think its too big of a change to do as part of this PR so this refactor will come in a later PR.

ext/drx/scatter_gather_aarch64.c
jackgallagher-arm: See my response above.

ext/drx/scatter_gather_aarch64.c
abhinav92003: Sure, ok.

ext/drx/scatter_gather_aarch64.c
abhinav92003: Thanks.

jackgallagher-arm:
The consistent `ci-x86/x86-32` failure looks like it is due to #6417

abhinav92003:
> The consistent `ci-x86/x86-32` failure looks like it is due to #6417

I can merge this PR when the CI completes.

abhinav92003:
I'm seeing new issues on the burst_traceopts test:


```
2023-11-21T15:50:33.0935549Z 267:   D:\a\dynamorio\dynamorio\build_debug-internal-64\clients\bin64\tool.drcacheoff.burst_traceopts.exe
2023-11-21T15:50:33.0937319Z 267:   (712) DynamoRIO usage error : internal isa mode error>
2023-11-21T15:50:33.0938001Z 267: 
2023-11-21T15:50:33.0938785Z 267:   <Usage error: internal isa mode error
2023-11-21T15:50:33.0939603Z 267:   (D:\a\dynamorio\dynamorio\core\ir\decode_shared.c, line 281)
```

and
```
2023-11-21T15:50:33.1036394Z 267:   D:\a\dynamorio\dynamorio\build_debug-internal-64\clients\bin64\tool.drcacheoff.burst_traceopts.exe
2023-11-21T15:50:33.1037652Z 267:   (1056).  Internal Error: DynamoRIO debug check failure:
2023-11-21T15:50:33.1038527Z 267:   D:\a\dynamorio\dynamorio\core\win32\callback.c:3404 false
2023-11-21T15:50:33.1054071Z 267: 
2023-11-21T15:50:33.1058780Z 267:   (Error occurred @18 frags in tid 2076)
```

Doesn't seem to be the same as #6423.

On the same test in a re-run, I also saw #6423 and 

```
2023-11-21T15:09:42.4290110Z CMake Error at D:/a/dynamorio/dynamorio/suite/tests/runmulti.cmake:135 (message):
2023-11-21T15:09:42.4291535Z   output |Trace mismatch found: Trace types do not match: 10 vs 14
2023-11-21T15:09:42.4292293Z 
2023-11-21T15:09:42.4292547Z   all done
2023-11-21T15:09:42.4292903Z 
2023-11-21T15:09:42.4293274Z   | failed to match expected output |^all done
```

abhinav92003:
Okay vs2019-64 passed on a re-run. Submitting.

jackgallagher-arm:
Thanks @abhinav92003 .

PR #6404, opened 2023-11-01
i#5843 scheduler: Add direct thread switch marker

derekbruening:
Adds a new marker type TRACE_MARKER_TYPE_DIRECT_THREAD_SWITCH for use with custom kernel scheduling features where one thread directly switches to another on the same cpu.

Refactors raw2trace marker processing code to allow a subclass to insert the new marker.

Makes the raw2trace blocking syscall code virtual to allow a subclass to label custom syscalls as blocking.

Given that the changes are used in separate code it is not simple to make a test of the raw2trace refactoring + virtual. For the marker: tests that use the marker will be forthcoming in scheduler_unit_tests.

Issue: #5843

### DIFF
--- a/clients/drcachesim/common/trace_entry.h
+++ b/clients/drcachesim/common/trace_entry.h
@@ -560,6 +560,14 @@ typedef enum {
      */
     TRACE_MARKER_TYPE_SYSCALL_FAILED,
 
+    /**
+     * This marker is emitted prior to a system call that causes an immediate switch to
+     * another thread on the same core (with the current thread entering an unscheduled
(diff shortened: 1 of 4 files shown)


derekbruening:
ubuntu22 failure is api.rseq yet again #6185

derekbruening:
Given that the changes are used in separate code it is not simple to make a test of the raw2trace refactoring + virtual.  For the marker: tests that *use* the marker will be forthcoming in scheduler_unit_tests.

abhinav92003 (approved):

Issue #6406, opened 2023-11-03
Update raw2trace.cpp to Import syscall_linux_xxx header file based on the target architecture.


ivankyluk:
raw2trace_t::is_maybe_blocking_syscall() depends on the definitions of SYS_xxx to determine if the syscall may be blocking or not. 

core/unix/include/syscall.h imports syscall_linux_xxx header file based on the host architecture. For cross platform support, the syscall_linux_xxx header file should be imported based on the target architecture.

derekbruening:
This has to be carefully considered as there may well be syscalls executed in host!=target modes which of course have to match the host.

PR #6407, opened 2023-11-03
i#6406: Import syscall_linux header file based on the target architecture.

ivankyluk:
raw2trace_t::is_maybe_blocking_syscall() depends on the definitions of SYS_xxx to determine if the syscall may be blocking or not. 

core/unix/include/syscall.h imports syscall_linux_xxx header file based on the host architecture. For cross platform support, the syscall_linux_xxx header file should be imported based on the target architecture.

Fixes: i6406


clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -47,7 +47,17 @@
 #include "utils.h"
```
abhinav92003, line 47: Re the PR description: I suppose you mean s/i6404/#6406/

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -47,7 +47,17 @@
...
 #ifdef LINUX
 // XXX: We should have the core export this to an include dir.
-#    include "../../core/unix/include/syscall.h"
+#    ifdef X86
```
abhinav92003, line 50: Could create a header similar to syscall.h with this logic so it can be reused; maybe call it syscall_target.h.

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -47,7 +47,17 @@
...
+#        include "../../core/unix/include/syscall_linux_x86.h"
```
abhinav92003, line 51: Xref @derekbruening's comment on #6406: https://github.com/DynamoRIO/dynamorio/issues/6406#issuecomment-1793322402

(Moving the discussion here)

If host!=target, for raw2trace it seems correct to me that we would always want to use the target arch's system calls (like we use the target arch's decoder). In what case would we need the host's?

clients/drcachesim/tracer/raw2trace.cpp
ivankyluk: Yup. I've fixed the description.

clients/drcachesim/tracer/raw2trace.cpp
ivankyluk: I've updated the changes accordingly.

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -47,7 +47,7 @@
 #include "utils.h"
```
abhinav92003, line 47: Could we also add a test that ensures that is_maybe_blocking_syscall does the right thing, and make sure it runs in the a64-on-x86 workflow too?

abhinav92003 (approved):

clients/drcachesim/tracer/raw2trace.cpp
abhinav92003: That could be done in raw2trace_unit_tests.cpp I guess where you can add a public API to raw2trace_test_t  that calls the protected is_maybe_blocking_syscall.

clients/drcachesim/tracer/raw2trace.cpp
derekbruening: For raw2trace, yes, we always want the target.  But for core DR even when just used for decoding it might do printing to the console or other actions using direct syscalls (if the same executable is using full DR statically linked and so can't easily switch to drdecodelib) and there it wants host syscalls.  So the picture is complex for core DR.

PR #6410, opened 2023-11-06
Fix incorrect doxygen xref

derekbruening:
Fixes a doxygen ref to "dr_abort_ex" to the correct "dr_abort_with_code".


derekbruening:
We are seeing newly frequent flaky failures:

x64 failure is #2267 floatpc_xl8all which we think is a virtualization bug.

win64 has api.dis and decenc.drdecode_decenc_x86_64: but w/o details on the failure b/c of PR #6254.  Filed as #6411 but may be the same as #5725.

abhinav92003 (approved):

Issue #6411, opened 2023-11-06
api.dis and decenc.drdecode_decenc_x86_64 failures in output matching


derekbruening:
This seems to have started recently (though xref #5725):

x64 Linux:
https://github.com/DynamoRIO/dynamorio/actions/runs/6752187857/job/18358030169
```
420/473 Test #422: code_api|api.dis .................................................***Failed    0.10 sec
CMake Error at /home/runner/work/dynamorio/dynamorio/suite/tests/runcmp.cmake:114 (message):
  failed to match expected output
```

Same failure on win64:
https://github.com/DynamoRIO/dynamorio/actions/runs/6775123864/job/18413681364?pr=6410

Similarly obscure failure in the same win64 run:
```
297/303 Test #290: code_api|decenc.drdecode_decenc_x86_64 ...........................***Failed    2.10 sec
CMake Error at D:/a/dynamorio/dynamorio/suite/tests/runcmp.cmake:114 (message):
  failed to match expected output
```

Did we change runcmp.cmake to not print out the diff?
I think PR #6254 is the culprit: that is why we now just have this opaque
message without details.  So these may be the same as #5725.

derekbruening:
From https://github.com/DynamoRIO/dynamorio/pull/6428#issuecomment-1803128207 adding the diff: this does look like #5725 so let's collapse into that one.

leewvic:
Attaching a diff output for the decoder problem.  Current theory:  there are some difference how AMD vs. Intel machines interpret the opcodes.
[6_Run Suite.txt](https://github.com/DynamoRIO/dynamorio/files/13311768/6_Run.Suite.txt)

PR #6414, opened 2023-11-07
i#5843 scheduler: Add flexible_queue_t and use it in scheduler_t

derekbruening:
Adds a flexible priority queue class which tracks indices and so supports asking whether an entry is in the queue and removing an entry from anywhere in the queue.  Adds a simple unit test.

Changes the scheduler to use this new queue class, in anticipation of needing both new features to handle direct targeted thread switches.

Issue: #5843


derekbruening:
Something changed in the GA VM's: we are seeing frequent flakiness.  Esp floatpc_xl8all #2267, api.dis and drdecode_decenc_x86_64 #6411.

clients/drcachesim/scheduler/flexible_queue.h
```diff
@@ -0,0 +1,219 @@
...
+
+private:
+    void
+    print(const std::string &message)
+    {
+        if (verbose_ <= 0)
+            return;
+        std::cout << message << "\n";
+        for (index_t i = 0; i < entries_.size(); ++i) {
+            std::cout << i << ": " << entries_[i] << " @ " << entry2index_[entries_[i]]
```
abhinav92003, line 145: This assumes T overloads <<. If it doesn't it'll cause a compile issue. Add to docx.

clients/drcachesim/scheduler/flexible_queue.h
```diff
@@ -0,0 +1,219 @@
...
+        index_t parent = parent_node(node);
+        if (parent == INVALID_INDEX || !compare_(entries_[parent], entries_[node]))
+            return;
+        swap(node, parent);
+        percolate_up(parent);
+    }
+
+    std::vector<T> entries_;
+    comparator_t compare_;
+    std::unordered_map<T, index_t> entry2index_;
```
abhinav92003, line 212: Assumes T is trivially hashable. Ideally we should accept a hash function too in the template.

clients/drcachesim/tests/flexible_queue_tests.cpp
```diff
@@ -0,0 +1,83 @@
...
+namespace dynamorio {
+namespace drmemtrace {
+namespace {
+
+bool
+test_basics()
+{
+    // Create a max-heap queue.
+    flexible_queue_t<int, std::greater<int>> q(/*verbose=*/1);
+    assert(!q.find(4));
```
abhinav92003, line 50: Isn't this a min heap? Since it's returning the min?

clients/drcachesim/scheduler/flexible_queue.h
```diff
@@ -0,0 +1,219 @@
...
+            return;
+        swap(node, should_be_parent);
+        percolate_down(should_be_parent);
+    }
+
+    void
+    percolate_up(index_t node)
+    {
+        index_t parent = parent_node(node);
+        if (parent == INVALID_INDEX || !compare_(entries_[parent], entries_[node]))
```
abhinav92003, line 204: Shouldn't all `!compare_` actually be `compare_`? I think this has inverted the order. E.g. for the test, since we're returning the min, we should have passed std::less (and also call it a min-heap).

abhinav92003 (approved):

clients/drcachesim/scheduler/flexible_queue.h
derekbruening: I originally had it inverted and the test had the default std::less, but the scheduler_t's std::priority_queue used the opposite convention: so to match std::priority_queue I had to invert it here with `!`.  So I assume the STL convention is that compare(a,b) returning true means a is "worse".  The std::priority_queue [docs](https://en.cppreference.com/w/cpp/container/priority_queue) say "e.g. using [std::greater](http://en.cppreference.com/w/cpp/utility/functional/greater)<T> would cause the smallest element to appear as the [top()](https://en.cppreference.com/w/cpp/container/priority_queue/top)."

clients/drcachesim/tests/flexible_queue_tests.cpp
derekbruening: Yes, switched it but failed to update the comment when I realized everything's inverted (see other comment).  Updated.

clients/drcachesim/scheduler/flexible_queue.h
abhinav92003: > The std::priority_queue docs say "e.g. using std::greater would cause the smallest element to appear as the top()."

Ah I see. So their convention is the other way around. Well, fine.

clients/drcachesim/scheduler/flexible_queue.h
```diff
@@ -0,0 +1,219 @@
...
+    bool
+    push(T entry)
+    {
+        if (entry2index_.find(entry) != entry2index_.end())
+            return false; // Duplicates not allowed.
+        entries_.push_back(entry);
+        index_t node = entries_.size() - 1;
+        entry2index_[entry] = node;
+        percolate_up(node);
+        print("after push");
```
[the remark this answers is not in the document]
derekbruening: That was my original plan as it is a nop by default but it does have a function call cost.  Maybe it should be disabled more.  Still, it may be useful in the future.  I renamed it to `vprint` with a verbosity threshold so it's clearer what it is.

clients/drcachesim/scheduler/flexible_queue.h
derekbruening: I thought about throwing out all the printing code...but it may be useful in the future.  I added the << to the doc comment.

clients/drcachesim/scheduler/flexible_queue.h
derekbruening: Added a comment by the compare_ field and in the class doc.

clients/drcachesim/scheduler/flexible_queue.h
derekbruening: Done.

derekbruening:
A lot of flaky failures (grrr; may need a Fixit and try to get all contributors helping to deal with this):
+ ubuntu22 x32: drcacheoff.simple invalid cti: #6416 
+ ubuntu20 x32: drcachesim.opcode_mix encoding size != instr size #6303
+ win32: floatpc_xl8all #2267 
+ win64: api.dis +decenc #6411, plus burst_traceopts which I don't know that we've seen before: trying to repro that one

derekbruening:
I tried to repro burst_traceopts in my win64 setup: but it passes 20x in a row.

I put in tmate and then burst_traceopts passed!  Grr.  Presumably a flake.  Will file an issue on it.

derekbruening:
Tried tmate again without the failure gate and logged in and ran `ctest --repeat-until-fail 50 -R burst_traceopts` and it passed all 50 times.  Grrr.  Yet another failure we can't reproduce unless it lands on a certain VM in the GA cloud??

derekbruening:
Filed #6423 on the traceopts failure on win64.

derekbruening:
OK this time we have just one failure: win32 floatpc #2267.

PR #6415, opened 2023-11-07
i#5694 core-sharded: Add TRACE_MARKER_TYPE_CORE_WAIT

derekbruening:
Adds a new TRACE_MARKER_TYPE_CORE_WAIT marker which is a synthetic marker inserted in core-sharded drmemtrace analysis tool mode when the scheduler returns STATUS_WAIT.  This is meant for tools which analyze schedules themselves.

Adds a unit test.

Issue: #5694


derekbruening:
Crazy flakiness seems to have suddenly appeared.  Filed #6417 for the 81 (!) 32-bit ones; #6411 covers api.dis + decenc; #2267 the floatpc.

clients/drcachesim/analyzer.cpp
```diff
@@ -122,6 +122,15 @@ analyzer_t::record_is_timestamp(const memref_t &record)
         record.marker.marker_type == TRACE_MARKER_TYPE_TIMESTAMP;
 }
 
+template <>
+void
+analyzer_t::set_wait_record(memref_t &record)
```
abhinav92003, line 127: create_wait_marker that returns the memref_t would be more readable.

clients/drcachesim/analyzer.cpp
```diff
@@ -122,6 +122,15 @@ analyzer_t::record_is_timestamp(const memref_t &record)
...
+{
+    record.marker.type = TRACE_TYPE_MARKER;
+    record.marker.marker_type = TRACE_MARKER_TYPE_WAIT;
+    record.marker.marker_value = 0;
```
abhinav92003, line 131: Make a symbolic constant or add a comment about the zero.

clients/drcachesim/analyzer.cpp
```diff
@@ -160,6 +169,15 @@ record_analyzer_t::record_is_timestamp(const trace_entry_t &record)
     return record.type == TRACE_TYPE_MARKER && record.size == TRACE_MARKER_TYPE_TIMESTAMP;
 }
 
+template <>
+void
+record_analyzer_t::set_wait_record(trace_entry_t &record)
+{
+    record.type = TRACE_TYPE_MARKER;
+    record.size = TRACE_MARKER_TYPE_WAIT;
```
abhinav92003, line 177: Same here.

clients/drcachesim/analyzer.cpp
```diff
@@ -510,11 +528,10 @@ analyzer_tmpl_t<RecordType, ReaderType>::process_tasks(analyzer_worker_data_t *w
...
         if (status == sched_type_t::STATUS_WAIT) {
-            // TODO i#5694: We'd like the forthcoming schedule_stats tool to know about
-            // waits and idle periods (to record "-" in its string): should the analyzer
-            // insert a new marker type that doesn't count toward ordinals (or else it
-            // needs a scheduler API to inject it)?
-            continue;
+            // We let tools know about waits so they can analyze the schedule.
+            // We synthesize a record here.  If we wanted this to count toward output
+            // stream ordinals we would need to add a scheduler API to inject it.
+            set_wait_record(record);
```
abhinav92003, line 534: What does a wait record on its own signify? There are no units or other ordering to denote how long the wait is/was.

clients/drcachesim/common/trace_entry.h
```diff
@@ -568,6 +568,13 @@ typedef enum {
      */
     TRACE_MARKER_TYPE_DIRECT_THREAD_SWITCH,
 
+    /**
+     * This marker is used for core-sharded analyses to indicate that the current
+     * core is waiting on another core.  This is primarily only useful for analyses
+     * studying the scheduling of threads onto cores.
+     */
+    TRACE_MARKER_TYPE_WAIT,
```
abhinav92003, line 576: "wait" on its own is ambiguous. How about core_sharded_wait or core_wait.

clients/drcachesim/reader/reader.h
```diff
@@ -190,6 +190,9 @@ class reader_t : public std::iterator<std::input_iterator_tag, memref_t>,
     bool
     is_record_synthetic() const override
     {
+        if (cur_ref_.marker.type == TRACE_TYPE_MARKER &&
+            cur_ref_.marker.marker_type == TRACE_MARKER_TYPE_WAIT)
+            return true;
         return suppress_ref_count_ >= 0;
```
abhinav92003, line 196: I don't remember how this logically implies a synthetic record. Please add a comment.

clients/drcachesim/tools/view.cpp
```diff
@@ -424,9 +424,13 @@ view_t::parallel_shard_memref(void *shard_data, const memref_t &memref)
             std::cerr << "<marker: system call trace end>\n";
             break;
         case TRACE_MARKER_TYPE_BRANCH_TARGET:
+            // These are not expected to be visible but we handle nonetheless.
             std::cerr << "<marker: indirect branch target 0x" << std::hex
                       << memref.marker.marker_value << std::dec << ">\n";
             break;
+        case TRACE_MARKER_TYPE_WAIT:
+            std::cerr << "<marker: wait for another core>\n";
```
abhinav92003, line 432: We should add more docx on how the user should interpret this. So the current core was waiting, how long?

clients/drcachesim/tools/view.cpp
```diff
@@ -424,9 +424,13 @@ view_t::parallel_shard_memref(void *shard_data, const memref_t &memref)
             std::cerr << "<marker: system call trace end>\n";
             break;
         case TRACE_MARKER_TYPE_BRANCH_TARGET:
+            // These are not expected to be visible but we handle nonetheless.
```
abhinav92003, line 427: Since it's added to memref.instr.indirect_branch_target and not visible separately.

clients/drcachesim/tests/analysis_unit_tests.cpp
```diff
@@ -177,10 +183,184 @@ test_queries()
...
+        inputs[i].push_back(make_pid(1));
+        // The last input will be earlier than all others. It will execute
+        // 3 instrs on each core. This is to test the case when an output
+        // begins in the wait state.
+        for (int j = 0; j < (i == NUM_INPUTS - 1 ? 6 : NUM_INSTRS); j++)
+            inputs[i].push_back(make_instr(42 + j * 4));
+        inputs[i].push_back(make_exit(tid));
+    }
+
+    // Synthesize a cpu-schedule file with some waits in it, if run in lockstep.
```
abhinav92003, line 211: Add what the dots and dashes mean for a quick reference for the reader. Dashes are waits and dots are non-instrs?

clients/drcachesim/tests/analysis_unit_tests.cpp
```diff
@@ -177,10 +183,184 @@ test_queries()
...
+    tools.push_back(test_tool.get());
+    mock_analyzer_t analyzer(sched_inputs, &tools[0], (int)tools.size(),
+                             /*parallel=*/true, NUM_OUTPUTS, &sched_ops);
+    assert(!!analyzer);
+    bool res = analyzer.run();
+    assert(res);
+    for (const auto &sched : schedule_strings) {
+        std::cerr << "Schedule: " << sched << "\n";
+    }
+    assert(schedule_strings[1][0] == '-');
```
abhinav92003, line 355: This is the final check the test does? That there was one wait? Add a comment.

clients/drcachesim/tests/analysis_unit_tests.cpp
```diff
@@ -177,10 +183,184 @@ test_queries()
...
+        inputs[i].push_back(make_exit(tid));
+    }
+
+    // Synthesize a cpu-schedule file with some waits in it, if run in lockstep.
+    std::string cpu_fname = "tmp_test_wait_records.zip";
+    static const char *const CORE0_SCHED_STRING = "EEE-AAA-CCCAAACCCBBB.DDD.";
+    static const char *const CORE1_SCHED_STRING = "---EEE.BBBDDDBBBDDDAAA.CCC.";
+    {
+        std::vector<schedule_entry_t> sched0;
+        sched0.emplace_back(TID_BASE + 4, 10, CPU0, 0);
```
abhinav92003, line 217: I forget... how does the schedule entry say that cpu0 executed only 3 instrs from E? Based on the next higher start_instrs in any other schedule_entry across all cores, which would be the one on L225, but that has start_instr = 4, so shouldn't there be 4 instrs executed from this input at cpu0?

clients/drcachesim/tests/analysis_unit_tests.cpp
```diff
@@ -177,10 +183,184 @@ test_queries()
...
+        // 3 instrs on each core. This is to test the case when an output
+        // begins in the wait state.
+        for (int j = 0; j < (i == NUM_INPUTS - 1 ? 6 : NUM_INSTRS); j++)
+            inputs[i].push_back(make_instr(42 + j * 4));
+        inputs[i].push_back(make_exit(tid));
+    }
+
+    // Synthesize a cpu-schedule file with some waits in it, if run in lockstep.
+    std::string cpu_fname = "tmp_test_wait_records.zip";
+    static const char *const CORE0_SCHED_STRING = "EEE-AAA-CCCAAACCCBBB.DDD.";
```
abhinav92003, line 213: These aren't used. Are they only for reference?

abhinav92003 (approved):

clients/drcachesim/analyzer.cpp
derekbruening: Done.

clients/drcachesim/analyzer.cpp
derekbruening: Changed it to memset the other fields.

clients/drcachesim/analyzer.cpp
derekbruening: Done.

clients/drcachesim/common/trace_entry.h
derekbruening: CORE_WAIT.

clients/drcachesim/analyzer.cpp
derekbruening: One wait per query to the scheduler.  I updated the docs to say "  A new marker is emitted each
     * time the tool analysis framework requests a new record from the scheduler and
     * is given a wait status."

clients/drcachesim/reader/reader.h
derekbruening: Done.  These are synthesized by the framework and not part of the original inputs nor counting toward ordinals.  See the docs for is_record_synthetic().

clients/drcachesim/tools/view.cpp
derekbruening: "  A new marker is emitted each
     * time the tool analysis framework requests a new record from the scheduler and
     * is given a wait status.  There are no units of time here but each repetition
     * is roughly the time where a regular record could have been read and passed
     * along."

clients/drcachesim/tools/view.cpp
derekbruening: Yes, augmented comment.

clients/drcachesim/tests/analysis_unit_tests.cpp
derekbruening: Yes, moved the constant strings into the comment as what would happen with pure lockstep.

clients/drcachesim/tests/analysis_unit_tests.cpp
derekbruening: Added a comment about non-determinism (hard to make it purely lockstep from here) meaning we can't have a very strict check.

clients/drcachesim/tests/analysis_unit_tests.cpp
derekbruening: Instr counts are 1-based.  (Yet the 1st entry lists a 0...there was a complication with the entry listing 1 IIRC.). Added a comment.

clients/drcachesim/tests/analysis_unit_tests.cpp
derekbruening: Yes, for pure lockstep.  Moved into comment.

derekbruening:
A lot of flaky failures:
+ ubuntu22 x32: drcacheoff.simple invalid cti: #6416 
+ ubuntu20 x64: api.dis +decenc #6411, floatpc_xl8all #2267
+ ubuntu20 x32: drcachesim.opcode_mix encoding size != instr size #6303
+ win64: race copying drconfiglib.dll #5888

This failure needs looking into: drcachesim.core_sharded `core_sharded_test.cpp:256: void dynamorio::drmemtrace::test_real_files(const char*): Assertion `replay_out == record_out' failed.`

derekbruening:
The record-replay test of the threadsig.x86 trace fails 1 out of 3 times: there is an extra
thread, with apparently zero records of any counted type as all the other
numbers match.  An erroneously inserted thread exit by the scheduler?
Oh, maybe it is the wait markers, as basic_counts counts threads by looking
blindly at the memref_t.tid field not matching the prior one.

derekbruening:
Now hitting:
+ ubuntu20 x64: api.dis +decenc + client.attach_test #6411 == #5725, floatpc_xl8all #2267
+ ubuntu20 x32: set of failures in #6417

PR #6419, opened 2023-11-07
i#3544 RV64: Implement tp, stolen reg and lr/sc sequence mangling

ksco:
This patch implements several mangling procedures for tp, stolen reg, and lr/sc sequences.

1. For any instruction using tp or stolen register, we use a scratch register to hold the app value and mangle the instr to use that scratch register.
2. For the lr/sc sequence, a similar approach to AARCHXX's exclusive monitor is adopted: mangling LR to a normal load, and SC to a compare-and-swap. The function name (mangle_exclusive_monitor_op) is kept unchanged for simplicity.

Note there are several places that are left unimplemented, they are rarely used in practice, so it should be okay to implement it later.

Issue: #3544


core/arch/riscv64/mangle.c
```diff
@@ -385,13 +394,406 @@ mangle_rel_addr(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
...
+
+    if (uses_reg_stolen) {
+        scratch_reg2 = pick_scratch_reg(dcontext, instr, scratch_reg1, &slot2);
+        PRE(ilist, instr, instr_create_save_to_tls(dcontext, scratch_reg2, slot2));
+        PRE(ilist, instr,
+            instr_create_restore_from_tls(dcontext, scratch_reg2, TLS_REG_STOLEN_SLOT));
+    }
+
+    /* Keep the release semantics if needed. */
+    if (aqrl & LRSC_ORDERING_RL_MASK) {
```
derekbruening, line 627: style: Use TESTALL

core/arch/riscv64/mangle.c
```diff
@@ -385,13 +394,406 @@ mangle_rel_addr(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
...
+    if (opnd_get_reg(src0) == dr_reg_stolen) {
+        opnd_replace_reg(&src0, dr_reg_stolen, scratch_reg2);
+    }
+    instr_reset(dcontext, instr);
+    instr_set_opcode(instr, opcode);
+    instr_set_num_opnds(dcontext, instr, 1, 1);
+    instr_set_dst(instr, 0, dst);
+    instr_set_src(
+        instr, 0,
+        opnd_add_flags(opnd_create_base_disp(opnd_get_reg(src0), DR_REG_NULL, 0, 0, opsz),
```
derekbruening, line 654: You call opnd_create_base_disp above in this file but don't add this printing flag there.  Shouldn't this be auto-added for all base-disp for RISC-V?

core/arch/riscv64/mangle.c
```diff
@@ -385,13 +394,406 @@ mangle_rel_addr(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
...
+                       DR_OPND_IMM_PRINT_DECIMAL));
+    instr_set_translation(instr, instrlist_get_translation_target(ilist));
+
+    /* Keep the acquire semantics if needed. */
+    if (aqrl & LRSC_ORDERING_AQ_MASK) {
```
derekbruening, line 659: style: Use TESTALL

core/arch/riscv64/mangle.c
```diff
@@ -385,13 +394,406 @@ mangle_rel_addr(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
...
+    PRE(ilist, instr,
+        INSTR_CREATE_bne(dcontext, opnd_create_instr(fail), opnd_create_reg(scratch_reg1),
+                         opnd_create_reg(scratch_reg2)));
+
+    PRE(ilist, instr,
+        instr_create_restore_from_tls(dcontext, scratch_reg1, TLS_LRSC_VALUE_SLOT));
+    PRE(ilist, instr, loop);
+
+    /* Convert exclusive store to a compare-and-swap.
+     * Begin of the LR/SC sequence.
```
derekbruening, line 744: Please add a comment somewhere showing a full sample code expansion to help the reader understand this.  For arm we have ldstex.dox which discusses the design and gives examples.

core/arch/riscv64/mangle.c
```diff
@@ -385,13 +394,406 @@ mangle_rel_addr(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
...
+                              opnd_add_flags(opnd_create_immed_int(1, OPSZ_12b),
+                                             DR_OPND_IMM_PRINT_DECIMAL)));
+
+    PRE(ilist, next_instr, final);
+    PRE(ilist, next_instr,
+        XINST_CREATE_load_int(dcontext, opnd_create_reg(scratch_reg1),
+                              opnd_add_flags(opnd_create_immed_int(-1, OPSZ_12b),
+                                             DR_OPND_IMM_PRINT_DECIMAL)));
+
+    /* Invalidate reservation regardless of success or failure. */
```
derekbruening, line 777: This belongs before the load_int too?  Writing -1 is what invalidates it?  Why -1?  I would expect 0.  Does it fail if not aligned or sthg?  Please explain in a comment.

core/arch/riscv64/mangle.c
```diff
@@ -385,13 +394,406 @@ mangle_rel_addr(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
...
+}
+
+static void
+mangle_stolen_reg_and_tp_reg(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
                              instr_t *next_instr)
 {
-    /* FIXME i#3544: Not implemented */
-    ASSERT_NOT_IMPLEMENTED(false);
-    return false;
+    ushort slot;
```
derekbruening, line 421: Add ASSERT_NOT_TESTED(), which can be removed once we have an actual test.  That makes it clear this has not actually been testing.  (Or if you are able to run an end to end test: please include that test in this PR.)

core/arch/riscv64/mangle.c
```diff
@@ -385,13 +394,406 @@ mangle_rel_addr(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
...
+        PRE(ilist, next_instr,
+            instr_create_restore_from_tls(dcontext, scratch_reg, slot));
+    }
+}
+
+static void
+mangle_cbr_stolen_reg_and_tp_reg(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
+                                 instr_t *next_instr)
+{
+    instr_t *fall = INSTR_CREATE_label(dcontext);
```
derekbruening, line 518: Add ASSERT_NOT_TESTED(), which can be removed once we have an actual test.  That makes it clear this has not actually been testing.  (Or if you are able to run an end to end test: please include that test in this PR.)

core/arch/riscv64/mangle.c
```diff
@@ -385,13 +394,406 @@ mangle_rel_addr(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
...
+    /* Restore scratch registers. */
+    PRE(ilist, next_instr, instr_create_restore_from_tls(dcontext, scratch_reg1, slot1));
+    PRE(ilist, next_instr, instr_create_restore_from_tls(dcontext, scratch_reg2, slot2));
+    return next_instr;
+}
+
+instr_t *
+mangle_exclusive_monitor_op(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
+                            instr_t *next_instr)
+{
```
derekbruening, line 790: Add ASSERT_NOT_TESTED(), which can be removed once we have an actual test.  That makes it clear this has not actually been testing.  (Or if you are able to run an end to end test: please include that test in this PR.)

core/arch/riscv64/mangle.c
ksco: > This belongs before the load_int too?

Yes, moved it above the load_int.

> Writing -1 is what invalidates it? Why -1? I would expect 0. Does it fail if not aligned or sthg? Please explain in a comment.

Write -1 to the lrsc address slot since -1 is never a valid address, 0 should work too yes. LR/SC does require the address to be aligned, but that will be caught by the CAS sequence. Updated the comment.

ksco (commented):
Thanks for the review! I've updated the PR as suggested.

core/arch/riscv64/mangle.c
```diff
@@ -784,10 +817,89 @@ mangle_exclusive_store(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
...
+ *      bne         a1, zero, loop          # retry on failure, end of the sequence
+ *      j           final
+ * fail:
+ *      li          a1, 1                   # sets non-zero value to dst on failure
+ * final:
+ *      li          a0, -1
+ *      sd          a0, lrsc_addr_slot(t3)  # invalidate reservation
+ *      ld          a0, a0_slot(t3)         # restore scratch register 1
+ *      ld          a4, a4_slot(t3)         # restore scratch register 2
+ */
```
derekbruening, line 893: (Nice, I think these comments really help.). (Trying to use parens to mark as "no action requested"...wish github reviews had that as a real option.)

derekbruening (approved):

PR #6422, opened 2023-11-08
i#4393 AArch64 codec: Add half-precision to fpimm8_13.

egrimley-arm:
In codec.c, the `fpimm8_13` functions now look more similar to the `fpimm8_5` ones. Tests added to `dis-a64-v82.txt`.

### DIFF
--- a/core/ir/aarch64/codec.c
+++ b/core/ir/aarch64/codec.c
@@ -6793,35 +6793,36 @@ encode_opnd_svemem_vec_vec_idx(uint enc, int opcode, byte *pc, opnd_t opnd,
     return true;
 }
 
-/* fpimm13: floating-point immediate for scalar fmov */
+/* fpimm8_13: floating-point immediate for scalar fmov */
 
 static inline bool
 decode_opnd_fpimm8_13(uint enc, int opcode, byte *pc, OUT opnd_t *opnd)
 {
-    uint a = extract_uint(enc, 20, 1);
-    uint b = extract_uint(enc, 19, 1);
-    uint c = extract_uint(enc, 18, 1);
-    uint defgh = extract_uint(enc, 13, 5);
-
-    if (extract_uint(enc, 22, 1) == 0) { /* 32 bits */
-        return decode_fpimm8_single(a, b, c, defgh, opnd);
-    } else { /* 64 bits */
-        return decode_fpimm8_double(a, b, c, defgh, opnd);
+    const uint size = extract_uint(enc, 22, 2);
+    const uint a = extract_uint(enc, 20, 1);
+    const uint b = extract_uint(enc, 19, 1);
+    const uint c = extract_uint(enc, 18, 1);
+    const uint defgh = extract_uint(enc, 13, 5);
+    switch (size) {
+    case 0b00: return decode_fpimm8_single(a, b, c, defgh, opnd);
+    case 0b01: return decode_fpimm8_double(a, b, c, defgh, opnd);
+    case 0b11: return decode_fpimm8_half(a, b, c, defgh, opnd);
     }
+
+    return false;
 }
 
 static inline bool
 encode_opnd_fpimm8_13(uint enc, int opcode, byte *pc, opnd_t opnd, OUT uint *enc_out)
 {
-    if (opnd_is_immed_float(opnd)) {
-        ASSERT(extract_uint(enc, 22, 1) == 0); /* 32 bit floating point */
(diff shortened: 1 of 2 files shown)


core/ir/aarch64/codec.c
```diff
@@ -6793,35 +6793,36 @@ encode_opnd_svemem_vec_vec_idx(uint enc, int opcode, byte *pc, opnd_t opnd,
...
-        return decode_fpimm8_double(a, b, c, defgh, opnd);
+    const uint size = extract_uint(enc, 22, 2);
+    const uint a = extract_uint(enc, 20, 1);
+    const uint b = extract_uint(enc, 19, 1);
+    const uint c = extract_uint(enc, 18, 1);
+    const uint defgh = extract_uint(enc, 13, 5);
+    switch (size) {
+    case 0b00: return decode_fpimm8_single(a, b, c, defgh, opnd);
+    case 0b01: return decode_fpimm8_double(a, b, c, defgh, opnd);
+    case 0b11: return decode_fpimm8_half(a, b, c, defgh, opnd);
```
AssadHashmi, line 6809: A `default: ASSERT_NOT_REACHED();` would be handy here. Same for the `switch` in `encode_opnd_fpimm8_13()`. I notice `fpimm8_5` doesn't have it either.

AssadHashmi (changes requested):

core/ir/aarch64/codec.c
egrimley-arm: Generally the `decode_opnd_` and `encode_opnd_` functions are supposed to return `false` when things don't work out rather than "crash" because the caller might be just querying whether an instruction can be decoded/encoded rather than relying on it succeeding. In the particular case of `fpimm8_13` and with the `codec*.txt` files as they are at the moment I think `case 0b10` really is unreachable but I think that's kind of a coincidence and might change in future without it being a bug so an `ASSERT` would be unhelpful.

With `fpimm8_5`, `case 0b00` is reachable I think, if you gave the decoder something like `00000101000100001100000000000000`.

AssadHashmi: > Generally the decode_opnd_ and encode_opnd_ functions are supposed to return false when things don't work out rather than "crash" because the caller might be just querying whether an instruction can be decoded/encoded rather than relying on it succeeding.

Yes, that was the case a while back but when we hit issues implementing Neoverse support, the decoder generated by `codec.py` had to be more precise based on calling a unique decode function for a given bitmask defined in `codec*.txt`. If  the decode function returns `false` the decode fails. Using `fpimm8_5` decode as an example, the decode function is:
```
/* ['z_size_hsd_0'] <- ['fpimm8_5'] */                                    
static bool                                                               
decode_opndsgen_2539c000_00c01fff(uint enc, dcontext_t *dcontext, byte *pc, instr_t *instr, int opcode)
{                                                                         
    opnd_t dst0, src0;                                                    
    if (!decode_opnd_z_size_hsd_0(enc & 0x00c0001f, opcode, pc, &dst0) || 
        !decode_opnd_fpimm8_5(enc & 0x00c01fe0, opcode, pc, &src0))       
        return false;                                                     
    instr_set_opcode(instr, opcode);                                      
    instr_set_num_opnds(dcontext, instr, 1, 1);                           
    instr_set_dst(instr, 0, dst0);                                        
    instr_set_src(instr, 0, src0);                                        
    return true;                                                          
}
```
Which is called by the decode tree logic:
```
    } else {                              
        if ((enc & 0xff3fe000) == 0x2539c000)
            return decode_opndsgen_2539c000_00c01fff(enc, dc, pc, instr, OP_fdup);
        else if ((enc & 0xff20c000) == 0x0520c000)
```
Whatever bool `decode_opndsgen_2539c000_00c01fff()` returns is transmitted up the callstack, in this case from `decoder_sve()`->`decoder_v86()`...`decoder_v80()`->`decode_common()`.

Unless I've missed something?

You are correct in the case of the encoder. That hasn't changed and still works on the "trial-and-error" basis.

AssadHashmi: > A `default: ASSERT_NOT_REACHED();` would be handy here. Same for the `switch` in `encode_opnd_fpimm8_13()`. I notice `fpimm8_5` doesn't have it either.

Sorry, I meant to say:
A `default: ASSERT_NOT_REACHED();` would be handy here. I notice `decode_opnd_fpimm8_5` doesn't have it either.

AssadHashmi (approved):

Issue #6416, opened 2023-11-07
tool.drcacheoff.simple fails on some x86-32-ubuntu22 VMs


abhinav92003:
The tool.drcacheoff.simple test is flaky on the x86-32-ubuntu22 workflow. I was able to reproduce it in a Ubuntu-22-04 VM but not always.  I encountered this first on #6408, none of other recent PRs seem to have this failure but I was able to reproduce this on the master branch on a VM.

Details of the VM where I was able to reproduce it:
```
abhinav92003@instance-1:~/dr/build/dr1$ uname -a
Linux instance-1 6.2.0-1018-gcp #20~22.04.1-Ubuntu SMP Mon Oct 23 12:29:43 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
abhinav92003@instance-1:~/dr/build/dr1$ ldd --version
ldd (Ubuntu GLIBC 2.35-0ubuntu3.4) 2.35
```

Even on the master branch it seems to fail:

```
test 302
    Start 302: code_api|tool.drcacheoff.simple

302: Test command: /usr/bin/cmake "-D" "precmd=foreach@/usr/bin/cmake@-E@remove_directory@tool.drcacheoff.simple.*.dir" "-D" "cmd=/home/abhinav92003/dr/build/master_build/bin32/drrun@-s@90@-quiet@-debug@-killpg@-stderr_mask@0xC@-dumpcore_mask@0@-code_api@-t@drcachesim@-offline@-subdir_prefix@tool.drcacheoff.simple@--@/home/abhinav92003/dr/build/master_build/suite/tests/bin/simple_app" "-D" "postcmd=firstglob@/home/abhinav92003/dr/build/master_build/clients/bin32/drcachesim@-indir@tool.drcacheoff.simple.*.dir@-chunk_instr_count@10K" "-D" "postcmd2=" "-D" "postcmd3=" "-D" "failok=" "-D" "cmp=/home/abhinav92003/dr/build/master_build/suite/tests/offline-simple.expect" "-D" "code=" "-D" "capture=" "-D" "ignore_matching_lines=" "-P" "/home/abhinav92003/dr/src/dr1/suite/tests/runmulti.cmake"
302: Test timeout computed to be: 90
[...]

Issue #5412, opened 2022-03-10
tool.drcacheoff.burst_replaceall failure on Windows with 2nd thread in simulator results


derekbruening:
Looks like we expect one thread, but two are running.  Is it one of the Windows pool threads?  Why is it only showing up on this one run?

https://github.com/DynamoRIO/dynamorio/runs/5502017781?check_suite_focus=true
```
238: Running postcmd |D:/a/dynamorio/dynamorio/build_debug-internal-64/clients/bin64/drcachesim.exe;-indir;D:/a/dynamorio/dynamorio/build_debug-internal-64/suite/tests/drmemtrace.tool.drcacheoff.burst_replaceall.exe.02076.9096.dir|
238: CMake Error at D:/a/dynamorio/dynamorio/suite/tests/runmulti.cmake:130 (message):
238:   output |replace all file functions
238: 
238:   pre-DR init
238: 
238:   pre-DR start
238: 
238:   pre-DR detach
238: 
238:   processing 15 buffers
238: 
238:   creating module file
238:   .\drmemtrace.tool.drcacheoff.burst_replaceall.exe.02076.9096.dir\raw\modules.log
238: 
238: 
238:   all done
238: 
238:   Cache simulation results:
238: 
238:   Core #0 (1 thread(s))
238: 
238:     L1I stats:
238:       Hits:                          212,510
238:       Misses:                            251
238:       Compulsory misses:                 251
238:       Invalidations:                       0
238:       Miss rate:                        0.12%
238:     L1D stats:
238:       Hits:                           78,521
238:       Misses:                             62
238:       Compulsory misses:                 100
238:       Invalidations:                       0
238:       Prefetch hits:                      24
238:       Prefetch misses:                    38
238:       Miss rate:                        0.08%
238: 
238:   Core #1 (1 thread(s))
238: 
238:     L1I stats:
238:       Hits:                            2,577
238:       Misses:                            168
238:       Compulsory misses:                 168
238:       Invalidations:                       0
238:       Miss rate:                        6.12%
238:     L1D stats:
238:       Hits:                            1,094
238:       Misses:                             98
238:       Compulsory misses:                 172
238:       Invalidations:                       0
238:       Prefetch hits:                      24
238:       Prefetch misses:                    74
238:       Miss rate:                        8.22%
238: 
238:   Core #2 (0 thread(s))
238: 
238:   Core #3 (0 thread(s))
238: 
238:   LL stats:
238: 
238:       Hits:                                3
[...]

PR #6428, opened 2023-11-09
i#6411 runcmp: Use external diff on failure if possible

ksco:
Use `find_program` to detect presence of the external diff command and, if such a command is detected, print a full diff on failure to ease debugging.

Issue: #6411
Related PR: #6254


ksco:
Now we have the diff output:

https://github.com/DynamoRIO/dynamorio/actions/runs/6807093606/job/18509368067?pr=6428#step:7:27881

derekbruening:
> Now we have the diff output:
> 
> https://github.com/DynamoRIO/dynamorio/actions/runs/6807093606/job/18509368067?pr=6428#step:7:27881

Good, so it does look like #6411 is the same as #5725!  Though we still don't understand the cause of the failure...trying to get tmate on it.

suite/tests/runcmp.cmake
```diff
@@ -111,5 +111,19 @@ if (lists_identical)
...
+  find_program(DIFF_CMD diff)
+  if (DIFF_CMD)
+    # Detected diff command, print a full diff for better debugging.
+    list(JOIN output "" output)
+    list(JOIN filtered_expect "" filtered_expect)
+    set(tmp "${cmp}-out")
+    set(tmp2 "${cmp}-expect")
+    file(WRITE "${tmp}" "${output}")
+    file(WRITE "${tmp2}" "${filtered_expect}")
+    set(diffcmd "diff")
```
derekbruening, line 123: I think this line should be removed and the execute_process should use `${DIFF_CMD}`, which is the one we found: right?

ksco: Sorry, updated to use `DIFF_CMD`.

derekbruening (approved):

derekbruening:
Hitting the suddenly flaky tests: win64: decoder #6411, replaceall #5412; ub22 #6416

PR #6435, opened 2023-11-09
i#3544 RV64: Fix a copy-paste mistake in execute_handler_from_dispatch

ksco:
This patch fixes some typos that may introduced by copy-paste from `execute_handler_from_cache()`.

Issue: #3544


ksco:
Still trying to figure out how signal handling and fault translation work, but looking at the code above for other architectures I'm pretty sure this is a mistake.

derekbruening (approved):

Issue #6417, opened 2023-11-07
Sudden flaky failures of 81 tests on x86 32-bit GA CI


derekbruening:
A run on GA CI suddently saw 81 failures.  The virtualization changed, or something else was updated on the VM's?  These all seem to be new flaky failures with maybe one seen before.  Xref #5725.

https://github.com/DynamoRIO/dynamorio/actions/runs/6779945230/job/18427829003?pr=6415
```
debug-internal-32: 333 tests passed, **** 82 tests failed, of which 1 were flaky: ****
	code_api|common.floatpc_xl8all 
	code_api|linux.signal0000 
	code_api|linux.signal0001 
	code_api|linux.signal0010 
	code_api|linux.signal0011 
	code_api|linux.signal0100 
	code_api|linux.signal0101 
	code_api|linux.signal0110 
	code_api|linux.signal0111 
	code_api|linux.signalfd 
	code_api|common.decode-stress 
	code_api|linux.eintr-noinline 
	code_api|client.flush 
	code_api|client.strace 
	code_api|tool.drcachesim.phys-threads_SUDO 
	code_api|tool.drcov.eintr 
	code_api|linux.eintr 
	code_api|tool.drcachesim.delay-global 
	code_api|tool.drcachesim.threads 
	code_api|tool.drcachesim.threads-with-config-file 
	code_api|tool.drcachesim.coherence 
	code_api|api.static_sideline_FLAKY 
```

linux.signal*:
```
52: Generating SIGSEGV
52: <Application /home/runner/work/dynamorio/dynamorio/build_debug-internal-32/suite/tests/bin/linux.signal0000 (21169).  Internal Error: DynamoRIO debug check failure: /home/runner/work/dynamorio/dynamorio/core/unix/signal.c:6008 syscall_signal || safe_is_in_fcache(dcontext, pc, (byte *)sc->SC_XSP)
(Error occurred @2170 frags in tid 21169)
```

drcachesim.threads:
```
292:      Started iteration 1 of the computation...
292: <Application /home/runner/work/dynamorio/dynamorio/build_debug-internal-32/suite/tests/bin/client.annotation-concurrency (22375).  Internal Error: DynamoRIO debug check failure: /home/runner/work/dynamorio/dynamorio/core/vmareas.c:925 start < end
292: (Error occurred @3233 frags in tid 22375)
```
This may be #6152

client.flush:
```
182: count = 402
182: ERROR removing 0xf7b93549
[...]

PR #6440, opened 2023-11-09
Avoid failing in syscall_mix for traces without syscall markers

abhinav92003:
Removes the check for trace file type that fails for older traces that do not have the syscall marker. It is better to simply report zero counts than to fail with a fatal error.


derekbruening (approved):

abhinav92003:
x86-32-ubuntu22 failure is #6416

abhinav92003:
x86-64 api.rseq failure is #6185

abhinav92003:
x86-32 bit failures are #6417.

abhinav92003:
I'm going to go ahead and merge this with admin privileges, so that we can unblock some dependent work. This PR is a very simple change and the failures are known issues: #6417 and #6416.

Issue #6426, opened 2023-11-09
Add drmemtrace schedule analysis tool


derekbruening:
#5843 covered creating a drmemtrace scheduler supporting a variety of schedules.

#5694 hooked the scheduler up to the analysis tool framework.

This issue covers creating an analysis tool that computes statistics about a schedule and provides some rough visualization of the interleavings.  The sched_launcher tool was a step in that direction, but an analysis tool is a better solution as it more easily fits into and combines with other tools and existing framework adaptations.

Here is my prototype, where I cranked up the quantum so we have no preempts as a test to see 100% voluntary switches (this counts a thread exit as a voluntary switch):

```
$ bin64/drrun -t drcachesim -simulator_type schedule_stats -indir ../src/clients/drcachesim/tests/drmemtrace.threadsig.x64.tracedir/ -core_sharded -print_every 5000 -sched_quantum 1000000
Schedule stats tool results:
Total counts:
           4cores
           8 threads
      638938 instructions
           5 total context switches
   0.0000078 CSPKI (context switches per 1000 instructions)
           5 voluntary context switches
           0 direct context switches
      100.00% voluntary switches
        0.00% direct switches
         161 system calls
           2 maybe-blocking system calls
           0 direct switch requests
Core #0 counts:
           2 threads
      117015 instructions
           2 total context switches
   0.0000171 CSPKI (context switches per 1000 instructions)
           2 voluntary context switches
           0 direct context switches
      100.00% voluntary switches
        0.00% direct switches
         125 system calls
           2 maybe-blocking system calls
           0 direct switch requests
Core #1 counts:
           2 threads
      170399 instructions
           1 total context switches
   0.0000059 CSPKI (context switches per 1000 instructions)
           1 voluntary context switches
           0 direct context switches
      100.00% voluntary switches
        0.00% direct switches
          10 system calls
           0 maybe-blocking system calls
           0 direct switch requests
Core #2 counts:
           2 threads
      175759 instructions
           1 total context switches
   0.0000057 CSPKI (context switches per 1000 instructions)
           1 voluntary context switches
           0 direct context switches
      100.00% voluntary switches
        0.00% direct switches
          13 system calls
           0 maybe-blocking system calls
           0 direct switch requests
Core #3 counts:
           2 threads
      175765 instructions
           1 total context switches
   0.0000057 CSPKI (context switches per 1000 instructions)
           1 voluntary context switches
           0 direct context switches
      100.00% voluntary switches
[...]

PR #6442, opened 2023-11-10
i#6426: Add schedule_stats tool

derekbruening:
Adds a new analysis tool "schedule_stats" which computes characteristics of a core-sharded schedule.  This is meant to help in understanding and studying the effects of varying the scheduler parameters.

This initial version counts instructions, blocking and all syscalls, voluntary and direct switches, and direct requests, and computes derived statistics CSPKI, I/CS, %voluntary, and %direct switches. Failed direct requests can be inferred.

The tool also records a string of letters representing input threads to help visualize the thread interleaving on each core.  The number of instructions per leter is controlled by a new parameter -print_every. This string visualization has its limits as more than 26 inputs will wrap around, but this approach has been useful in our unit tests and other small runs.

Adds a regression test which runs the checked-in threadsig trace.

Issue: #6426


derekbruening:
win32 and x64 failures: floatpc https://github.com/DynamoRIO/dynamorio/issues/2267, fixed by waiting-for-review https://github.com/DynamoRIO/dynamorio/pull/6441
ub22 failure: api.rseq #6185

clients/drcachesim/tools/schedule_stats_create.h
```diff
@@ -0,0 +1,60 @@
...
+namespace drmemtrace {
+
+/**
+ * @file drmemtrace/schedule_stats_create.h
+ * @brief DrMemtrace schedule statistics analysis tool creation.
+ */
+
+/**
+ * Creates an analysis tool which counts the number and type of context switches
+ * in a core-sharded trace schedule.
```
abhinav92003, line 52: Add to comment that it would fail in any other mode.

clients/drcachesim/tools/schedule_stats.cpp
```diff
@@ -0,0 +1,295 @@
...
+            shard->cur_segment_instrs = 0;
+            shard->prev_was_wait = true;
+        } else {
+            ++shard->cur_segment_instrs;
+            if (shard->cur_segment_instrs == knob_print_every_) {
+                shard->thread_sequence += '-';
+            }
+        }
+        return true;
+    } else
```
abhinav92003, line 173: `else` is redundant.

clients/drcachesim/tools/schedule_stats.cpp
```diff
@@ -0,0 +1,295 @@
...
+        shard->prev_was_wait = false;
```
abhinav92003, line 174: Ideally you'd set this at the end, just in case some other logic below wants to know whether prev_was_wait.

clients/drcachesim/tools/schedule_stats.cpp
```diff
@@ -0,0 +1,295 @@
...
+    int64_t input = shard->stream->get_input_id();
+    if (input != shard->prev_input) {
+        // We convert to letters which only works well for <=26 inputs.
+        if (!shard->thread_sequence.empty()) {
+            ++shard->counters.total_switches;
+            if (shard->saw_maybe_blocking || shard->saw_exit)
+                ++shard->counters.voluntary_switches;
+            if (shard->direct_switch_target == memref.marker.tid)
+                ++shard->counters.direct_switches;
+            shard->thread_sequence += ',';
```
abhinav92003, line 184: Maybe create separate constants up top together for all these special chars.

abhinav92003, line 184: We add a comma at each switch? Why? isn't the character changing enough? Or do you want to differentiate between thread 0 and thread 26.

clients/drcachesim/tools/schedule_stats.cpp
```diff
@@ -0,0 +1,295 @@
...
+                 << " == thread " << memref.instr.tid << "\n";
+            std::cerr << line.str();
+        }
+        shard->prev_input = input;
+    }
+    if (type_is_instr(memref.instr.type)) {
+        ++shard->counters.instrs;
+        // Print a single letter for any partial sequence.
+        if (shard->cur_segment_instrs == 0)
+            shard->thread_sequence += 'A' + static_cast<char>(input % 26);
```
abhinav92003, line 211: At a thread switch L186 adds the char to the thread sequence and sets cur_segment_insrs to zero, and here we'd add the char again?

clients/drcachesim/common/options.h
```diff
@@ -53,10 +46,20 @@
 #define VIEW "view"
 #define FUNC_VIEW "func_view"
 #define INVARIANT_CHECKER "invariant_checker"
+#define SCHEDULE_STATS "schedule_stats"
+
+#define REPLACE_POLICY_NON_SPECIFIED ""
```
abhinav92003, line 51: Why was this moved below? If the goal was  to separate vs the tool name constants, wouldn't just an additional vertical whitespace suffice?

Also, add comment to mark the logical separation of the constants. "all tool name constants" etc.

clients/drcachesim/common/options.cpp
```diff
@@ -854,5 +855,10 @@ droption_t<std::string>
                          "Path with stored as-traced schedule for replay.");
 #endif
 
+// Schedule_stats options.
+droption_t<uint64_t> op_print_every(DROPTION_SCOPE_ALL, "print_every", 5000,
```
abhinav92003, line 859: op_schedule_stats_print_every would be better.

abhinav92003 (approved):

clients/drcachesim/tools/schedule_stats_create.h
derekbruening: Done.

clients/drcachesim/tools/schedule_stats.cpp
derekbruening: Removed.

clients/drcachesim/tools/schedule_stats.cpp
derekbruening: Moved to the end.

clients/drcachesim/tools/schedule_stats.cpp
derekbruening: It seemed easier to read at a glance how long the sequences were.  Compare:
```
Core #0 schedule: AA,CC,HH,LL,UU,WW,YY,AA,CC,EE,FF,GG,HH,II,LL,MMM,KK,EE,KK,CC,KK,CC,KK,CC,KK,CC,KK,CC,KK,CC,KK,CC,KK,CC,KK,CC,KK,CC,KK,CC,KK,CC,KK,CC,KK,CC,KK,CC,KK,CC,KK,CC,KK,CC
Core #0 schedule: AACCHHLLUUWWYYAACCEEFFGGHHIILLMMMKKEEKKCCKKCCKKCCKKCCKKCCKKCCKKCCKKCCKKCCKKCCKKCCKKCCKKCCKKCCKKCCKKCCKKCCKKCC
```
Or:
```
Core #0 schedule: FFFFFF,HHHHHHHHHHHHHHHHHHH,FFF
Core #1 schedule: DDDDDDDDDDDDDDDDDD,CCCCCCCCCCCCCCCCCCC
Core #0 schedule: FFFFFFHHHHHHHHHHHHHHHHHHHFFF
Core #1 schedule: DDDDDDDDDDDDDDDDDDCCCCCCCCCCCCCCCCCCC
```

I imagine how we visualize longer sequences will evolve over time though.  I think starting with the , is nice.  I'll add a comment about maybe removing to make it more compact, but that making it harder to distinguish a 26-rollover switch like you said.

derekbruening: Done.

clients/drcachesim/tools/schedule_stats.cpp
derekbruening: Ah good catch.  Somehow I thought it was only adding after N so you wouldn't get a letter for a shorter sequence...not sure if the code changed in between.  But now, yes, this is redundant.  Moved this append inside the `== knob_print_every_`.

clients/drcachesim/common/options.h
derekbruening: Two reasons: A) to group it with the other per-tool defines below (CACHE_TYPE_INSTRUCTION, etc.) and B) it seemed more logical to have the tool names first, and then have per-tool specific constants.  Added comments.

clients/drcachesim/common/options.cpp
derekbruening: Done.

derekbruening:
Progress: we're down to just one test failing here.  x86-32: it's the set of failures in #6417.

PR #6443, opened 2023-11-10
Add better zipfile error diagnostics

derekbruening:
Stores the file path in the zipfile_file_reader_t and prints it in every diagnostic message.  Obtains the chunk name for the truncation message and includes that.

Tested on a trace with a truncated file.

### DIFF
--- a/clients/drcachesim/reader/zipfile_file_reader.cpp
+++ b/clients/drcachesim/reader/zipfile_file_reader.cpp
@@ -66,7 +66,7 @@ file_reader_t<zipfile_reader_t>::open_single_file(const std::string &path)
     unzFile file = unzOpen(path.c_str());
     if (file == nullptr)
         return false;
-    input_file_ = zipfile_reader_t(file);
+    input_file_ = zipfile_reader_t(file, path);
     if (unzGoToFirstFile(file) != UNZ_OK || unzOpenCurrentFile(file) != UNZ_OK)
         return false;
     VPRINT(this, 1, "Opened input file %s\n", path.c_str());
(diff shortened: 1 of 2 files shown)


derekbruening:
win32 failure: floatpc #2267, fixed by waiting-for-review #6441

derekbruening:
x32 failure: #6417

clients/drcachesim/reader/zipfile_file_reader.cpp
```diff
@@ -87,28 +87,31 @@ file_reader_t<zipfile_reader_t>::read_next_entry()
...
+                       zipfile->name, zipfile->path.c_str());
             }
 #endif
             // read_next_entry() stored the last-read entry into entry_copy_.
             if ((entry_copy_.type != TRACE_TYPE_MARKER ||
                  entry_copy_.size != TRACE_MARKER_TYPE_CHUNK_FOOTER) &&
                 entry_copy_.type != TRACE_TYPE_FOOTER) {
-                VPRINT(this, 1, "Chunk is missing footer: truncation detected\n");
+                zipfile->name[0] = '\0'; /* Just in case. */
+                unzGetCurrentFileInfo64(zipfile->file, nullptr, zipfile->name,
```
abhinav92003, line 103: This is only to get the filename? Isn't that in input_file_->path now?

abhinav92003 (approved):

clients/drcachesim/reader/zipfile_file_reader.cpp
derekbruening: That new path field is indeed the filename, but here we want the component name inside the zipfile (for ours, "chunk.0006" e.g.).

PR #6446, opened 2023-11-10
i#6444 no exit: Dump rseq buffer on mid-region thread end

### LINKED ISSUE #6444: Trace ending mid-rseq is truncated by raw2trace
I encountered a trace with a truncated file: no thread exit at the end.
It ends with a write after PC 0x1141e187:

```
000000 | 000a 0008 000000001140fc66
00000c | 000a 0003 000000001140fc6e
000018 | 000a 0003 000000001140fc71
000024 | 000a 0003 000000001140fc74
000030 | 000a 0003 000000001140fc77
00003c | 000e 0005 000000001140fc7a
000048 | 0001 0008 00007f1e1bca6958
000054 | 000a 0007 000000001141e180
000060 | 000a 0009 000000001141e187
00006c | 0001 0008 00007f1e1bca8d68
```

Re-running raw2trace: that PC is a ways up in the log, followed by a long rseq sequence (it has a loop) which doesn't reach its end: it has a timestap and thread exit appearing as the thread exited mid-rseq-region it seems.  So apparently we do not output the rseq buffer in this case.

```
   8294 [drmemtrace]: Chunk instr count is now 1313701
   8295   0x000000001141e187  64 48 89 04 25 68 66 mov    %rax -> %fs:0xffff6668[8byte]
   8296                       ff ff
   8297 [drmemtrace]: Appended instr fetch for original 0x1141e187
   8298 [drmemtrace]: Appended memref type write (1) size 8 to 0x7f1e1bca8d68
   8299 [drmemtrace]: Chunk instr count is now 1313702
   8300 [drmemtrace]: --- Reached rseq entry (end=0x1141e1d8): buffering all output ---
   8301 [drmemtrace]: Appended marker type 24 value 0x1141e1d8
<...long sequence because of a loop in the rseq region...>
   9966   0x000000001141e1ce  49 39 c3             cmp    %r11 %rax
   9967 [drmemtrace]: Appended instr fetch for original 0x1141e1ce
   9968   0x000000001141e1d1  75 ed                jnz    $0x000000001141e1c0
   9969 [drmemtrace]: Remembering rseq branch 0x1141e1d1 -> 0x1141e1c0
   9970 [drmemtrace]: Appended instr fetch for original 0x1141e1d1
   9971 [drmemtrace]: Delaying 1 entries for decode=0x1141e1d1
   9972 [drmemtrace]: Remembered delayed branch decode=0x1141e1d1 target=0x1141e1c0
   9973 [drmemtrace]: Thread 775342 timestamp 0x002f684c2ceca958
   9974 [drmemtrace]: Appended marker type 3 value 0x1024
   9975 [drmemtrace]: Thread 775342 timestamp 0x002f684c2cecb08b
   9976 [drmemtrace]: Appended marker type 3 value 0x1029
   9977 [drmemtrace]: Discarded 1 entries for final branch without subsequent instr
   9978 [drmemtrace]: Thread 775342 exit
   9979 I1109 22:24:49.979226 2274534 decomposer.cc:574] raw2trace complete!
```

Two action items:

1) Output rseq buffer if we hit thread exit mid-rseq
2) Add invariant check that a thread exit footer exists for every thread

derekbruening:
If the thread trace ends mid-rseq-region, dump the buffer instead of losing not only the rseq sequence so far but also the thread exit record.

Adds a unit test.  Also tested on a real trace with this issue.

Fixes #6444

### DIFF
--- a/clients/drcachesim/tests/raw2trace_unit_tests.cpp
+++ b/clients/drcachesim/tests/raw2trace_unit_tests.cpp
@@ -1984,6 +1984,82 @@ test_rseq_side_exit_inverted_with_timestamp(void *drcontext)
         check_entry(entries, idx, TRACE_TYPE_FOOTER, -1));
 }
 
+/* Tests a trace ending mid-rseq (i#6444). */
+bool
+test_midrseq_end(void *drcontext)
+{
+    std::cerr << "\n===============\nTesting mid-rseq trace end\n";
+    instrlist_t *ilist = instrlist_create(drcontext);
+    instr_t *nop = XINST_CREATE_nop(drcontext);
+    instr_t *move1 =
+        XINST_CREATE_move(drcontext, opnd_create_reg(REG1), opnd_create_reg(REG2));
+    instr_t *move2 =
+        XINST_CREATE_move(drcontext, opnd_create_reg(REG1), opnd_create_reg(REG2));
+    instr_t *jcc =
+        XINST_CREATE_jump_cond(drcontext, DR_PRED_EQ, opnd_create_instr(move2));
+    instr_t *store =
+        XINST_CREATE_store(drcontext, OPND_CREATE_MEMPTR(REG2, 0), opnd_create_reg(REG1));
+    instr_t *move3 =
+        XINST_CREATE_move(drcontext, opnd_create_reg(REG1), opnd_create_reg(REG2));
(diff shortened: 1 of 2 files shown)


derekbruening:
ub22 failure is drcacheoff.simple #6416

derekbruening:
x32 failure set is #6417

PR #6454, opened 2023-11-14
i#5995: Add exclusive mode to drcachesim.

### LINKED ISSUE #5995: Add exclusive cache support
Today drcachesim supports fully-inclusive caches but does not have support for exclusive caches where the LLC is essentially a victim cache.  This is how many real-world last-level caches operate (e.g., Intel Skylake) and we would like to be able to model that.

brettcoon:
Allow caches to be configured with an exclusive policy to function as victim caches.  Caches with this policy are only populated by lines evicted from their children.

Also added accessors for snoop filter stats to cache_simulator_t.

Fixes #5995


clients/drcachesim/simulator/caching_device.cpp
```diff
@@ -209,52 +224,16 @@ caching_device_t::request(const memref_t &memref_in)
...
-            // If no parent we assume we get the data from main memory
-            if (parent_ != NULL)
+            // If no parent we assume we get the data from main memory.
+            if (parent_ != NULL) {
                 parent_->request(memref);
-            if (snoop_filter_ != NULL) {
-                // Update snoop filter, other private caches invalidated on write.
-                snoop_filter_->snoop(tag, id_, (memref.data.type == TRACE_TYPE_WRITE));
             }
-
```
derekbruening, line 219: (OK, all this moved to insert_tag(), with added logic for exclusive.)

clients/drcachesim/simulator/caching_device.cpp
```diff
@@ -359,28 +341,65 @@ caching_device_t::contains_tag(addr_t tag)
...
+//       * All caches *above* (parent) the snooped level are marked
+//         "non_coherent".
+//    * On HITS, writes get propagated up to the snoop filter for coherent
+//        unsnooped caches, via propagate_write().  This is to invalidate
+//        any copies of the line in other caches.
+//    * On MISSES:
+//      * the miss request gets propagated to parent as normal.
+//      * if cache is snooped, update snoop filter for the new line.
+//      * if the new line evicts an old line:
+//        * if cache in inclusive, invalidate evicted line from all children.
```
derekbruening, line 358: nit: s/in/is/

clients/drcachesim/simulator/caching_device.h
```diff
@@ -61,13 +61,17 @@ namespace drmemtrace {
 class snoop_filter_t;
 class prefetcher_t;
 
+// NINE = Non-Inclusive Non-Exclusive.
+enum class cache_inclusion_policy_t { NINE, INCLUSIVE, EXCLUSIVE };
```
derekbruening, line 65: Hmm, is "NINE" a widely recognized acronym?  Would it be better to be named "NEITHER"?  That seems more understandable at a glance without having to go hunt down this header comment.

clients/drcachesim/simulator/caching_device_stats.h
```diff
@@ -245,6 +247,9 @@ class caching_device_stats_t {
 
     int64_t num_inclusive_invalidates_;
     int64_t num_coherence_invalidates_;
+    // TODO: Hmm, maybe ought to combine inclusive and exclusive invalidation
+    // counters since they're exclusive.
```
derekbruening, line 251: Do you mean, because no one cache can have both?

clients/drcachesim/simulator/caching_device_stats.h
```diff
@@ -245,6 +247,9 @@ class caching_device_stats_t {
 
     int64_t num_inclusive_invalidates_;
     int64_t num_coherence_invalidates_;
+    // TODO: Hmm, maybe ought to combine inclusive and exclusive invalidation
```
derekbruening, line 250: Style: Always point at an issue with a TODO: `TODO i#5995:`.

clients/drcachesim/tests/drcachesim_unit_tests.cpp
```diff
@@ -313,6 +336,129 @@ LLC {
...
+
+    // L1s are 1-way, while L2 and LLC are both 4-way.
+    // If we cycle through 4 conflicting lines multiple times, the L2 will
+    // hold all four lines and never evict anything to LLC: we expect all
+    // misses in L1, 4 misses and many hits in L2, 4 misses and no hits
+    // in LLC.
+
+    // Test 4 conflicting lines.
+    const int NUM_LOOPS = 16;
+    const int L2_ASSOC = 4;
```
derekbruening, line 394: Unused.  Plus, has to match the config file: so if it were used, should have some check.

clients/drcachesim/tests/drcachesim_unit_tests.cpp
```diff
@@ -313,6 +336,129 @@ LLC {
...
+    const int LLC_ASSOC = 4;
```
derekbruening, line 395: Ditto.

clients/drcachesim/tests/drcachesim_unit_tests.cpp
```diff
@@ -313,6 +336,129 @@ LLC {
...
+    const int LLC_SIZE = 64 * 1024;
```
derekbruening, line 396: Add an assert that the parsed config has this same size.  Or just get rid of the constant and obtain the value from cache_sim.

clients/drcachesim/tests/drcachesim_unit_tests.cpp
```diff
@@ -313,6 +336,129 @@ LLC {
...
+    const int ADDR_STRIDE = LLC_SIZE; // Maximize conflicts.
+    const int CONFLICTING_ADDRESSES = 4;
+    for (int n = 0; n < NUM_LOOPS; n++) {
```
derekbruening, line 399: Style: Prefer `++n` in C++ code.

clients/drcachesim/tests/drcachesim_unit_tests.cpp
```diff
@@ -313,6 +336,129 @@ LLC {
...
+            if (!cache_sim.process_memref(make_memref(ADDR_STRIDE * i))) {
+                std::cerr << "drcachesim unit_test_child_hits failed: "
+                          << cache_sim.get_error_string() << "\n";
+                exit(1);
+            }
+        }
+    }
+
+    std::cerr << "Results from second exclusive-cache sequence.\n";
+    cache_sim.print_results();
```
derekbruening, line 445: Is this supposed to be only for verbose/diagnostic runs?  Or was it supposed to be printed after the first sequence too?

derekbruening (approved):

clients/drcachesim/reader/config_reader.cpp
```diff
@@ -273,15 +273,36 @@ config_reader_t::configure_cache(cache_params_t &cache)
...
+                }
                 cache.inclusive = true;
             } else {
                 cache.inclusive = false;
             }
+        } else if (param == "exclusive") {
+            // Is the cache exclusive of its children.
+            std::string bool_val;
+            if (!(*fin_ >> bool_val)) {
+                ERRMSG("Error reading cache inclusion policy from "
```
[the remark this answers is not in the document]
brettcoon: I think both inclusive and exclusive are considered "inclusion" policies, which is why I used to same message for both.  But I agree it's probably nicer to have more specific messages.

clients/drcachesim/tests/drcachesim_unit_tests.cpp
brettcoon: Either option is trickier than it sounds since there's no straightforward way to query the cache parameters from a cache_simulator_t object.  But that would improve the test, so I made a wrapper that exposes the cache_t objects so I can check the parameters.

clients/drcachesim/tests/drcachesim_unit_tests.cpp
brettcoon: Weird, I (almost) never use post-increment.  Not sure what got into me there.

clients/drcachesim/tests/drcachesim_unit_tests.cpp
brettcoon: Oops, yes, it was leftover from my development/debug.  Originally I printed both.  It's gone.

clients/drcachesim/simulator/caching_device.h
brettcoon: Honestly, I had the same question, but I found that label used on Wikipedia's entry for inclusion policy, as well as in unrelated online materials I found (such as https://youtu.be/zJzxHxThgPg?si=Z_vEWVl9zHOYstfI&t=188).  So it's an established label even if not universally known.  I don't like NEITHER since it implies only two other options, and there could be other variations added later, for example to model some real-world caches that are just _mostly_ exclusive.  The alternative I considered is NON_INC_NON_EXC, which is maybe more obvious to those unfamiliar with NINE, though I obviously favor keeping the "official" name for this policy.

clients/drcachesim/simulator/caching_device_stats.h
brettcoon: Right, that was just a note to self that I forgot to delete.

clients/drcachesim/simulator/caching_device.h
derekbruening: The problem is NINE looks like 9.  N_I_N_E would be better IMHO.  I would vote for NON_INC_NON_EXC.

PR #6455, opened 2023-11-14
i#3348 sym conflicts: add DR_PARAM_ to IN, OUT, INOUT

derekbruening:
To avoid symbol conflicts with IN, OUT, and INOUT, which are defined in our exported headers, we add DR_PARAM_ to the front.

The aarch64 codec.c was left as the shorter names since the aarch64_check_codec_order.py assumes single-line signatures.

Issue: #3348


derekbruening:
Failures were all known flakes; re-ran them.

The DO NOT CHECK IN disabling the clang-format-diff will be reverted right before commit: it fails b/c of clang-format-diff not matching full clang-format in a few places.  I've hit some of these before and reverted the format on those prior merges, but here we're just going to put in the full new format and avoid future problems.

ivankyluk (approved):

derekbruening:
Failures are ub22 off.simple #6416 and x32 slate of AMD failures #6417.  Failures from this PR would be build failures in any case, not test failures.

PR #6456, opened 2023-11-15
i#6444 mid-rseq-exit: Handle at-exit trace exit

derekbruening:
Augments the mid-rseq-trace-exit case to handle the rseq final instr being reached but not passed.  Adds a test case.

Issue: #6444


ivankyluk (approved):

derekbruening:
x64 failure is api.rseq failed to synch #6185 
win64 failure is replaceall #5412

PR #6457, opened 2023-11-15
i#2267 utils.cmake: Use <unknown> for cpu_vendor as fallback

ksco:
`vendor_id` is not always available, only do `REGEX REPLACE` it a match if found.

Issue: #2267

### DIFF
--- a/make/utils.cmake
+++ b/make/utils.cmake
@@ -360,9 +360,11 @@ function (get_processor_vendor out)
       "[HKEY_LOCAL_MACHINE\\Hardware\\Description\\System\\CentralProcessor\\0;VendorIdentifier]"
       NAME)
   elseif (EXISTS "/proc/cpuinfo")
+    set (regex ".*vendor_id[ \t]*:[ \t]+([a-zA-Z0-9_-]+).*")
     file(READ "/proc/cpuinfo" contents)
-    string(REGEX REPLACE ".*vendor_id[ \t]*:[ \t]+([a-zA-Z0-9_-]+).*" "\\1"
-      cpu_vendor "${contents}")
+    if (contents MATCHES ${regex})
+      string(REGEX REPLACE ${regex} "\\1" cpu_vendor "${contents}")
+    endif ()
   endif ()
   message(STATUS "Processor vendor is ${cpu_vendor}")
   set(${out} ${cpu_vendor} PARENT_SCOPE)


make/utils.cmake
```diff
@@ -360,9 +360,11 @@ function (get_processor_vendor out)
       "[HKEY_LOCAL_MACHINE\\Hardware\\Description\\System\\CentralProcessor\\0;VendorIdentifier]"
       NAME)
   elseif (EXISTS "/proc/cpuinfo")
+    set (regex ".*vendor_id[ \t]*:[ \t]+([a-zA-Z0-9_-]+).*")
     file(READ "/proc/cpuinfo" contents)
-    string(REGEX REPLACE ".*vendor_id[ \t]*:[ \t]+([a-zA-Z0-9_-]+).*" "\\1"
-      cpu_vendor "${contents}")
+    if (contents MATCHES ${regex})
+      string(REGEX REPLACE ${regex} "\\1" cpu_vendor "${contents}")
```
derekbruening, line 366: This looks correct but with CMake it is always confusing: does `${regex}` need quotes around it, e.g.?  Best to test it: we really want to test it on an AMD machine.  Keep retrying some of the GA jobs until you get `Processor vendor is AuthenticAMD`, and mention that in the PR description that you did test on an AMD machine?

derekbruening (approved):

make/utils.cmake
ksco: Sure, we can see that on aarch64-native runners, the output is: [-- Processor vendor is \<unknown\>](https://github.com/DynamoRIO/dynamorio/actions/runs/6874202478/job/18695449275?pr=6457#step:6:44); on x86_64 AMD runners, the output is: [-- Processor vendor is AuthenticAMD](https://github.com/DynamoRIO/dynamorio/actions/runs/6874202466/job/18695449975?pr=6457#step:7:46). PR description updated.

derekbruening: Perfect, thanks.

ksco:
x86-32-ubuntu22 #6416 and x86-32 #6417 failed 5 times in a row.

derekbruening:
> x86-32-ubuntu22 #6416 and x86-32 #6417 failed 5 times in a row.

Any help on #6417 is appreciated.  Meanwhile I will merge this.

Issue #6423, opened 2023-11-08
win64 burst_traceopts non-det failure


derekbruening:
This started failing, maybe as part of the recent spate of failures?
https://github.com/DynamoRIO/dynamorio/actions/runs/6801885635/job/18493723120?pr=6414
```
263/305 Test #266: code_api|tool.drcacheoff.burst_traceopts .........................***Failed    0.39 sec
Running cmd |D:/a/dynamorio/dynamorio/build_debug-internal-64/clients/bin64/tool.drcacheoff.burst_traceopts.exe|
CMake Error at D:/a/dynamorio/dynamorio/suite/tests/runmulti.cmake:111 (message):
  *** cmd failed (3): Assertion failed: status_opt == scheduler_t::STATUS_OK
  && status_noopt == scheduler_t::STATUS_OK, file
  D:\a\dynamorio\dynamorio\clients\drcachesim\tests\burst_traceopts.cpp, line
  265

  abort() has been called***
```

I can't repro on my win64 setup: it passes 20x in a row there.

When I put in tmate and ran just this test: it passed!
https://github.com/DynamoRIO/dynamorio/actions/runs/6802953296/job/18497172665?pr=6414

I repeated, with tmate enabled on success too, and again it passed:
https://github.com/DynamoRIO/dynamorio/actions/runs/6803115563/job/18497701618?pr=6414

I logged in and ran ctest --repeat-until-fail 50 -R burst_traceopts and it
passed all 50 times. Grrr. Yet another failure we can't reproduce unless it
lands on a certain VM in the GA cloud??

PR #6458, opened 2023-11-16
i#5843 scheduler: Only consider long-latency syscalls blocking

derekbruening:
Rather than context switching on every syscall labeled maybe-blocking, the scheduler uses the now-available syscall latency to decide whether the syscall should block and result in a context switch.

Adds two new command line options, -sched_syscall_switch_us (default 500us) and -sched_blocking_switch_us (default 100us), and corresponding scheduler_t inputs, to control the latency thresholds. To avoid relying too much on the maybe-blocking labels, we do consider a very-high-latency syscall not marked as maybe-blocking to block.

Adds a new schedule_stats unit test.

Tested in a large proprietary app where this reduces the context switch rate from ~100x too high down to ~10x too high.  The next step of adding i/o wait times should further improve the representativeness.

Issue: #5843

### DIFF
--- a/clients/drcachesim/CMakeLists.txt
+++ b/clients/drcachesim/CMakeLists.txt
@@ -748,6 +748,9 @@ add_subdirectory(tools/external)
 # We build larger executables here.  All tests are added in suite/tests/ except unit tests.
 # Be sure to give the targets qualified test names ("tool.drcache*...").
 
+# XXX: Try to add a macro add_drcachesim_test() to share common pieces
+# of these executables.
+
 if (BUILD_TESTS)
   add_executable(tool.reuse_distance.unit_tests tests/reuse_distance_test.cpp)
   target_link_libraries(tool.reuse_distance.unit_tests drmemtrace_reuse_distance
--- a/clients/drcachesim/CMakeLists.txt
+++ b/clients/drcachesim/CMakeLists.txt
@@ -907,6 +907,16 @@ if (BUILD_TESTS)
     set_tests_properties(tool.drcachesim.invariant_checker_test PROPERTIES
       TIMEOUT ${test_seconds})
 
+    add_executable(tool.drcachesim.schedule_stats_test tests/schedule_stats_test.cpp)
+    configure_DynamoRIO_standalone(tool.drcachesim.schedule_stats_test)
+    add_win32_flags(tool.drcachesim.schedule_stats_test)
+    target_link_libraries(tool.drcachesim.schedule_stats_test drmemtrace_schedule_stats
+        drmemtrace_static drmemtrace_analyzer test_helpers)
+    add_test(NAME tool.drcachesim.schedule_stats_test
+             COMMAND tool.drcachesim.schedule_stats_test)
+    set_tests_properties(tool.drcachesim.schedule_stats_test PROPERTIES
+      TIMEOUT ${test_seconds})
+
     add_executable(tool.drcacheoff.view_test tests/view_test.cpp reader/file_reader.cpp)
     configure_DynamoRIO_standalone(tool.drcacheoff.view_test)
     add_win32_flags(tool.drcacheoff.view_test)
--- a/clients/drcachesim/common/options.cpp
+++ b/clients/drcachesim/common/options.cpp
@@ -838,6 +838,18 @@ droption_t<bool> op_sched_order_time(DROPTION_SCOPE_ALL, "sched_order_time", tru
                                      "Applies to -core_sharded and -core_serial. "
                                      "Whether to honor recorded timestamps for ordering");
 
+droption_t<uint64_t> op_sched_syscall_switch_us(
+    DROPTION_SCOPE_ALL, "sched_syscall_switch_us", 500,
+    "Minimum latency to consider any syscall blocking.",
+    "Minimum latency in timestamp units (us) to consider any syscall blocking. "
+    "Applies to -core_sharded and -core_serial. ");
+
+droption_t<uint64_t> op_sched_blocking_switch_us(
+    DROPTION_SCOPE_ALL, "sched_blocking_switch_us", 100,
+    "Minimum latency to consider a maybe-blocking syscall blocking.",
+    "Minimum latency in timestamp units (us) to consider any syscall that is marked as "
+    "maybe-blocking to be blocking. Applies to -core_sharded and -core_serial. ");
+
 #ifdef HAS_ZIP
 droption_t<std::string> op_record_file(DROPTION_SCOPE_FRONTEND, "record_file", "",
                                        "Path for storing record of schedule",
--- a/clients/drcachesim/common/options.cpp
+++ b/clients/drcachesim/common/options.cpp
@@ -840,15 +840,16 @@ droption_t<bool> op_sched_order_time(DROPTION_SCOPE_ALL, "sched_order_time", tru
 
 droption_t<uint64_t> op_sched_syscall_switch_us(
     DROPTION_SCOPE_ALL, "sched_syscall_switch_us", 500,
-    "Minimum latency to consider any syscall blocking.",
-    "Minimum latency in timestamp units (us) to consider any syscall blocking. "
-    "Applies to -core_sharded and -core_serial. ");
+    "Minimum latency to consider any syscall as incurring a context switch.",
+    "Minimum latency in timestamp units (us) to consider any syscall as incurring "
+    "a context switch.  Applies to -core_sharded and -core_serial. ");
 
 droption_t<uint64_t> op_sched_blocking_switch_us(
     DROPTION_SCOPE_ALL, "sched_blocking_switch_us", 100,
-    "Minimum latency to consider a maybe-blocking syscall blocking.",
+    "Minimum latency to consider a maybe-blocking syscall as incurring a context switch.",
     "Minimum latency in timestamp units (us) to consider any syscall that is marked as "
-    "maybe-blocking to be blocking. Applies to -core_sharded and -core_serial. ");
+    "maybe-blocking to incur a context switch. Applies to -core_sharded and "
+    "-core_serial. ");
 
 #ifdef HAS_ZIP
 droption_t<std::string> op_record_file(DROPTION_SCOPE_FRONTEND, "record_file", "",
--- a/clients/drcachesim/scheduler/scheduler.cpp
+++ b/clients/drcachesim/scheduler/scheduler.cpp
@@ -1565,6 +1565,21 @@ scheduler_tmpl_t<RecordType, ReaderType>::pop_from_ready_queue(
 }
 
 template <typename RecordType, typename ReaderType>
+bool
+scheduler_tmpl_t<RecordType, ReaderType>::treat_syscall_as_blocking(input_info_t *input)
+{
+    uint64_t post_time = input->reader->get_last_timestamp();
+    assert(input->processing_syscall);
+    assert(input->pre_syscall_timestamp > 0);
+    assert(input->pre_syscall_timestamp < post_time);
+    uint64_t latency = post_time - input->pre_syscall_timestamp;
+    VPRINT(this, 3, "input %d syscall latency: %" PRIu64 "\n", input->index, latency);
+    return (input->processing_blocking_syscall &&
+            latency >= options_.blocking_switch_threshold) ||
+        latency >= options_.syscall_switch_threshold;
+}
+
+template <typename RecordType, typename ReaderType>
 typename scheduler_tmpl_t<RecordType, ReaderType>::stream_status_t
 scheduler_tmpl_t<RecordType, ReaderType>::set_cur_input(output_ordinal_t output,
                                                         input_ordinal_t input)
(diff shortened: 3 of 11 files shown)


derekbruening:
ub22 is off.simple #6416

derekbruening:
win64 is traceopts #6423 and replaceall #5412 known flakes

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -2006,7 +2027,7 @@ scheduler_tmpl_t<RecordType, ReaderType>::next_record(output_ordinal_t output,
         } else if (options_.mapping == MAP_TO_ANY_OUTPUT) {
             trace_marker_type_t marker_type;
             uintptr_t marker_value;
-            if (input->processing_blocking_syscall) {
+            if (input->processing_syscall || input->processing_blocking_syscall) {
```
abhinav92003, line 2030: Enough to check just `processing_syscall`? L2064 would set it for also maybe-blocking ones right?

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -2025,29 +2046,43 @@ scheduler_tmpl_t<RecordType, ReaderType>::next_record(output_ordinal_t output,
...
-            } else if (options_.quantum_unit == QUANTUM_INSTRUCTIONS &&
-                       record_type_is_instr(record)) {
+                // Generally we should already have the timestamp from a just-prior
+                // syscall marker, but we support tests and other synthetic sequences
+                // with just a maybe-blocking.
+                input->pre_syscall_timestamp = input->reader->get_last_timestamp();
+            }
+            if (options_.quantum_unit == QUANTUM_INSTRUCTIONS &&
+                record_type_is_instr(record)) {
                 ++input->instrs_in_quantum;
```
abhinav92003, line 2077: So before we were not incrementing instrs_in_quantum when processing_blocking_syscall and record_type_is_instr, but we should have? And now we are because we s/`else if`/`if`?

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -2025,29 +2046,43 @@ scheduler_tmpl_t<RecordType, ReaderType>::next_record(output_ordinal_t output,
...
                     // input has the oldest timestamp, prioritizing context switches
                     // over timestamp ordering.
                     need_new_input = true;
                 }
-            } else if (options_.quantum_unit == QUANTUM_TIME) {
-                // The above if-else cases are all either for non-instrs or
-                // QUANTUM_INSTRUCTIONS, except the blocking syscall next instr which is
-                // already switching: so an else{} works here.
+            }
+            if (options_.quantum_unit == QUANTUM_TIME) {
```
abhinav92003, line 2085: This can remain an `else-if` right? Because both this and L2075 can never be true together.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -2025,29 +2046,43 @@ scheduler_tmpl_t<RecordType, ReaderType>::next_record(output_ordinal_t output,
...
                 }
-            } else if (record_type_is_marker(record, marker_type, marker_value) &&
-                       marker_type == TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL) {
+            }
+            if (record_type_is_marker(record, marker_type, marker_value) &&
+                marker_type == TRACE_MARKER_TYPE_SYSCALL) {
+                input->processing_syscall = true;
+                input->pre_syscall_timestamp = input->reader->get_last_timestamp();
+            }
+            if (record_type_is_marker(record, marker_type, marker_value) &&
```
abhinav92003, line 2067: L2062 and this line can still remain `else if` can they not? Unless we expect a consecutive pair of TRACE_MARKER_TYPE_SYSCALL and TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL (the first one sets the processing_* bool to true which would take L2030 in the next iter and not this block).

Need more comments if there is some real case.

clients/drcachesim/tests/schedule_stats_test.cpp
```diff
@@ -0,0 +1,208 @@
...
+        void
+        set_input_id(int64_t input_id)
+        {
+            input_id_ = input_id;
+        }
+        int64_t
+        get_input_id() const override
+        {
+            return input_id_;
+        }
```
abhinav92003, line 81: We can potentially override get_input_interface here to return another default_memtrace_stream_t, which will obviate the need for the test-specific check `shard->stream->get_input_interface() != nullptr`.

clients/drcachesim/tools/schedule_stats.cpp
```diff
@@ -140,7 +140,7 @@ schedule_stats_t::parallel_shard_memref(void *shard_data, const memref_t &memref
     static constexpr char THREAD_SEPARATOR = ',';
     static constexpr char WAIT_SYMBOL = '-';
     per_shard_t *shard = reinterpret_cast<per_shard_t *>(shard_data);
-    if (knob_verbose_ >= 4) {
+    if (knob_verbose_ >= 4 && shard->stream->get_input_interface() != nullptr) {
```
abhinav92003, line 143: Add comment that this condition is for tests, or better yet, support this in the test (see other comment).

clients/drcachesim/tools/schedule_stats.cpp
```diff
@@ -223,8 +223,12 @@ schedule_stats_t::parallel_shard_memref(void *shard_data, const memref_t &memref
     if (memref.instr.tid != INVALID_THREAD_ID)
         shard->counters.threads.insert(memref.instr.tid);
     if (memref.marker.type == TRACE_TYPE_MARKER) {
-        if (memref.marker.marker_type == TRACE_MARKER_TYPE_SYSCALL)
+        if (memref.marker.marker_type == TRACE_MARKER_TYPE_SYSCALL) {
             ++shard->counters.syscalls;
+            // The scheduler will block on a long-enough-latency non-marked
+            // syscall, so we label all to get the right % voluntary.
```
abhinav92003, line 229: Document that maybe_blocking numbers will be an overestimation.

We mark all system calls as blocking, will that number of useful?

clients/drcachesim/tools/schedule_stats.cpp
```diff
@@ -223,8 +223,12 @@ schedule_stats_t::parallel_shard_memref(void *shard_data, const memref_t &memref
...
+            shard->saw_maybe_blocking = true;
+        }
         if (memref.marker.marker_type == TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL) {
```
abhinav92003, line 232: nit: Else if? Or a switch-case maybe?

clients/drcachesim/scheduler/scheduler.h
```diff
@@ -965,8 +979,10 @@ template <typename RecordType, typename ReaderType> class scheduler_tmpl_t {
         bool order_by_timestamp = false;
         // Global ready queue counter used to provide FIFO for same-priority inputs.
         uint64_t queue_counter = 0;
-        // Used to switch on the instruction *after* a blocking syscall.
+        // Used to switch on the instruction *after* a long-latency syscall.
+        bool processing_syscall = false;
         bool processing_blocking_syscall = false;
```
abhinav92003, line 984: s/processing_blocking_syscall/processing_maybe_blocking_syscall/

abhinav92003 (approved):

clients/drcachesim/CMakeLists.txt
```diff
@@ -907,6 +907,16 @@ if (BUILD_TESTS)
     set_tests_properties(tool.drcachesim.invariant_checker_test PROPERTIES
       TIMEOUT ${test_seconds})
 
+    add_executable(tool.drcachesim.schedule_stats_test tests/schedule_stats_test.cpp)
```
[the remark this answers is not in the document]
derekbruening: I started on this but so many of these vary in complex ways it seems like a bunch of work I don't want to do now, so I added a XXX comment.

clients/drcachesim/common/options.cpp
```diff
@@ -838,6 +838,18 @@ droption_t<bool> op_sched_order_time(DROPTION_SCOPE_ALL, "sched_order_time", tru
                                      "Applies to -core_sharded and -core_serial. "
                                      "Whether to honor recorded timestamps for ordering");
 
+droption_t<uint64_t> op_sched_syscall_switch_us(
+    DROPTION_SCOPE_ALL, "sched_syscall_switch_us", 500,
+    "Minimum latency to consider any syscall blocking.",
```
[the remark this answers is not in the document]
derekbruening: I do mean "trigger a context switch".  I was using "blocking" to mean the same thing but maybe they are not quite the same as blocking to wait for a result may not 100% of the time lead to a switch, though generally it does.

A switch in the middle of kernel code: it depends on how the kernel is configured, for Linux.  Older kernels have zero preemption in any kernel code.  Modern ones have certain voluntary preemption points.  Real-time adds preemption anywhere not in a critical section (or possibly there too if replaced w/ spinlocks).  But, we are likely to see a long-latency syscall and do a context switch under the assumption it's blocking and we failed to mark it maybe-blocking: we have no plans to model preemption in syscalls.

I reworded the option docs s/blocking/incurring a context switch/, and s/treat_syscall_as_blocking/syscall_incurs_switch/.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -1564,6 +1564,27 @@ scheduler_tmpl_t<RecordType, ReaderType>::pop_from_ready_queue(
...
+    if (input->reader->get_version() < TRACE_ENTRY_VERSION_FREQUENT_TIMESTAMPS) {
+        // This is a legacy trace that does not have timestamps bracketing syscalls.
+        // We switch on every maybe-blocking syscall in this case.
+        return input->processing_blocking_syscall;
+    }
+    assert(input->pre_syscall_timestamp > 0);
+    assert(input->pre_syscall_timestamp < post_time);
+    uint64_t latency = post_time - input->pre_syscall_timestamp;
+    VPRINT(this, 3, "input %d %ssyscall latency: %" PRIu64 "\n", input->index,
+           input->processing_blocking_syscall ? "blocking " : "", latency);
```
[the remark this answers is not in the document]
derekbruening: It's just whether it was marked maybe.  Added "maybe-".

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -2025,29 +2046,43 @@ scheduler_tmpl_t<RecordType, ReaderType>::next_record(output_ordinal_t output,
                         input->switch_to_input = it->second;
                     }
                 } else if (record_type_is_instr(record)) {
-                    // Assume it will block and we should switch to a different input.
-                    need_new_input = true;
-                    in_wait_state = true;
+                    if (treat_syscall_as_blocking(input)) {
```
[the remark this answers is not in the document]
derekbruening: No this is the instr *after* a syscall (we have to read all the markers).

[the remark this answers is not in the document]
derekbruening: This is documented in the comment above at line 2031.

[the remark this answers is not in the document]
derekbruening: See the XXX comment above about wanting the switch before the return markers even now, but it not being simple to do that as we need to post-syscall timestamp.  Yes we'll have to figure it out.  Added another comment about that.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Some synthetic sequences in tests only set the maybe-blocking, and it looked painful to change them.  Added a comment.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: It does appear that way.  It doesn't seem to have affected any tests, even the ones with tiny quanta.  The elses caused other problems too with the changes in this PR so they had to go.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Yes.  OK, put it back.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: I guess so.  OK put this else back.

clients/drcachesim/tests/schedule_stats_test.cpp
derekbruening: Done.

clients/drcachesim/tools/schedule_stats.cpp
derekbruening: Removed.

clients/drcachesim/tools/schedule_stats.cpp
derekbruening: The counter is separate and we are not over-counting.  This saw_ boolean is only used for voluntary.

derekbruening: Renamed to saw_syscall to avoid confusion.

clients/drcachesim/tools/schedule_stats.cpp
derekbruening: Done.

clients/drcachesim/scheduler/scheduler.h
derekbruening: Done.

derekbruening:
The 2 failures are the AMD #6416 and #6417

PR #6460, opened 2023-11-17
i#5843 scheduler: Fix assert to allow 0-latency syscalls

derekbruening:
Fixes a < assert from PR #6458 to be <=, to allow the pre-syscall timestamp to equal the post-syscall timestamp.

Adds a test that fails without the fix.

Issue: #5843


derekbruening:
ub22 failure is .simple test #6416 
x32 failure is the set of known failures #6417

PR #6461, opened 2023-11-17
i#3544 RV64: Refine fault translation and signal handling

ksco:
Refine fault translation and signal handling, part of the linux.signalXXXX tests works on real hardware; however, they do not work on QEMU, so we can't enable them on CI.

Issue: #3544

### DIFF
--- a/core/arch/riscv64/mangle.c
+++ b/core/arch/riscv64/mangle.c
@@ -674,7 +674,6 @@ instr_is_ldstex_mangling(dcontext_t *dcontext, instr_t *instr)
 
     ptr_int_t val;
     if (instr_get_opcode(instr) == OP_fence || instr_get_opcode(instr) == OP_bne ||
-        instr_get_opcode(instr) == OP_bne ||
         /* Check for sc.w/d+bne+jal pattern.  */
         (instr_get_opcode(instr) == OP_jal && instr_get_prev(instr) != NULL &&
          instr_get_opcode(instr_get_prev(instr)) == OP_bne &&
--- a/core/arch/riscv64/mangle.c
+++ b/core/arch/riscv64/mangle.c
@@ -648,6 +648,47 @@ mangle_special_registers(dcontext_t *dcontext, instrlist_t *ilist, instr_t *inst
     return next_instr;
 }
 
+/***************************************************************************
+ * LR/SC sequence mangling.
+ */
+
+bool
+instr_is_ldstex_mangling(dcontext_t *dcontext, instr_t *instr)
+{
+    /* This should be kept in sync with mangle_exclusive_monitor_op(). */
+    if (!instr_is_our_mangling(instr))
+        return false;
+
+    opnd_t memop = opnd_create_null();
+    if (instr_get_opcode(instr) == OP_sd)
+        memop = instr_get_src(instr, 0);
+    else if (instr_get_opcode(instr) == OP_ld)
+        memop = instr_get_dst(instr, 0);
+    if (opnd_is_base_disp(memop)) {
+        ASSERT(opnd_get_index(memop) == DR_REG_NULL && opnd_get_scale(memop) == 0);
+        uint offs = opnd_get_disp(memop);
+        if (opnd_get_base(memop) == dr_reg_stolen && offs >= TLS_LRSC_ADDR_SLOT &&
+            offs <= TLS_LRSC_SIZE_SLOT)
+            return true;
+    }
+
+    ptr_int_t val;
+    if (instr_get_opcode(instr) == OP_fence || instr_get_opcode(instr) == OP_bne ||
+        instr_get_opcode(instr) == OP_bne ||
+        /* Check for sc.w/d+bne+jal pattern.  */
+        (instr_get_opcode(instr) == OP_jal && instr_get_prev(instr) != NULL &&
+         instr_get_opcode(instr_get_prev(instr)) == OP_bne &&
+         instr_get_prev(instr_get_prev(instr)) != NULL &&
+         instr_is_exclusive_store(instr_get_prev(instr_get_prev(instr)))) ||
+        instr_is_exclusive_load(instr) || instr_is_exclusive_store(instr) ||
+        (instr_is_mov_constant(instr, &val) &&
+         (val == 1 /* cas fail */ || val == -1 /* reservation invalidation */ ||
+          val == 4 /* lr.w/sc.w size */ || val == 8 /* lr.d/sc.d size */)))
+        return true;
+
+    return false;
+}
+
 static instr_t *
 mangle_exclusive_load(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
                       instr_t *next_instr)
--- a/core/arch/riscv64/mangle.c
+++ b/core/arch/riscv64/mangle.c
@@ -272,8 +272,7 @@ patch_mov_immed_arch(dcontext_t *dcontext, ptr_int_t val, byte *pc, instr_t *fir
 bool
 instr_check_xsp_mangling(dcontext_t *dcontext, instr_t *inst, int *xsp_adjust)
 {
-    /* FIXME i#3544: Not implemented */
-    ASSERT_NOT_IMPLEMENTED(false);
+    /* Not apply to RISC-V. */
     return false;
 }
 
--- a/core/arch/riscv64/mangle.c
+++ b/core/arch/riscv64/mangle.c
@@ -681,6 +681,8 @@ instr_is_ldstex_mangling(dcontext_t *dcontext, instr_t *instr)
          instr_is_exclusive_store(instr_get_prev(instr_get_prev(instr)))) ||
         instr_is_exclusive_load(instr) || instr_is_exclusive_store(instr) ||
         (instr_is_mov_constant(instr, &val) &&
+         /* XXX: These are fragile, should we look backward a bit to check for more
+            specific patterns? */
          (val == 1 /* cas fail */ || val == -1 /* reservation invalidation */ ||
           val == 4 /* lr.w/sc.w size */ || val == 8 /* lr.d/sc.d size */)))
         return true;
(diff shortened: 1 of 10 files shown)


ksco:
QEMU does not work well with drrun on signal tests for some unknown reason. Running `qemu-riscv64 -L /usr/riscv64-linux-gnu bin64/drrun -debug -xarch_root /usr/riscv64-linux-gnu -- suite/tests/bin/linux.signal0000` results following error while it works fine on real hardware.

```
Sending SIGUSR2
Sending SIGUSR1
<Application /root/dynamorio/build/suite/tests/bin/linux.signal0000 (32907).  DynamoRIO internal crash at PC 0x0000000071235c78.  Please report this at http://dynamorio.org/issues/.  Program aborted.
Received SIGSEGV at pc 0x0000000071235c78 in thread 32907
Base: 0x0000000071000000
Registers:  pc=0x0000000071235c78 ra=0x000000007120e3f0 sp =0x0000004004ad8730 gp =0x0000004000809948
        tp=0x0000004004aab7a0 t0=0x00000000000878b0 t1 =0x0000004004acc378 t2 =0x0000000000000000
        s0=0x0000004004ad8760 s1=0x0000000000000001 a0 =0x0000004004acc038 a1 =0x0000004004ad9000
        a2=0x0000000000000110 a3=0x0000000000000000 a4 =0x0000004004ad8cc0 a5 =0x0000004004acc488
        a6=0x0000000000000000 a7=0x0000000000000081 s2 =0x0000000000000000 s3 =0x0000004000808de8
        s4=0x000000400080684a s5=0x00000040008000b8 s6 =0x0000004000808de8 s7 =0x0000004004a13d18
        s8=0x0000004004a14050 s9=0x0000000000000000 s10=0x0000000000000000 s11=0x0000000000000000
        t3=0x0000004004aa8000 t4=0x000000000000eec0 t5 =0x0000000000000004 t6 =0xffffffffffffffff

version 10.0.19678, custom build
```

ksco:
vs2019-64 looks like #6423, but there is one more test failed: [code_api|tool.drcacheoff.burst_replaceall](https://github.com/DynamoRIO/dynamorio/actions/runs/6907657696/job/18795281194?pr=6461#step:6:23264).
x86-32-ubuntu22 #6416
x86-32 #6417

core/arch/riscv64/mangle.c
```diff
@@ -648,6 +648,47 @@ mangle_special_registers(dcontext_t *dcontext, instrlist_t *ilist, instr_t *inst
...
+        ASSERT(opnd_get_index(memop) == DR_REG_NULL && opnd_get_scale(memop) == 0);
+        uint offs = opnd_get_disp(memop);
+        if (opnd_get_base(memop) == dr_reg_stolen && offs >= TLS_LRSC_ADDR_SLOT &&
+            offs <= TLS_LRSC_SIZE_SLOT)
+            return true;
+    }
+
+    ptr_int_t val;
+    if (instr_get_opcode(instr) == OP_fence || instr_get_opcode(instr) == OP_bne ||
+        instr_get_opcode(instr) == OP_bne ||
```
derekbruening, line 677: Duplicate `OP_bne`: do you mean `OP_bnez` here?

core/arch/riscv64/mangle.c
```diff
@@ -648,6 +648,47 @@ mangle_special_registers(dcontext_t *dcontext, instrlist_t *ilist, instr_t *inst
...
+        /* Check for sc.w/d+bne+jal pattern.  */
+        (instr_get_opcode(instr) == OP_jal && instr_get_prev(instr) != NULL &&
+         instr_get_opcode(instr_get_prev(instr)) == OP_bne &&
+         instr_get_prev(instr_get_prev(instr)) != NULL &&
+         instr_is_exclusive_store(instr_get_prev(instr_get_prev(instr)))) ||
+        instr_is_exclusive_load(instr) || instr_is_exclusive_store(instr) ||
+        (instr_is_mov_constant(instr, &val) &&
+         (val == 1 /* cas fail */ || val == -1 /* reservation invalidation */ ||
```
derekbruening, line 685: Hmm I wonder if this can be a false positive if other mangling has similar constants: should we look backward so the mov_constant is qualified a bit?  Or just put a XXX comment that this could possibly cause issues if two different mangling sequences look similar.  (Looking at arm: I guess it has the same issue.  Maybe there are no other mangling sequences that use constants like this?)

core/unix/signal.c
```diff
@@ -2921,6 +2920,13 @@ sigcontext_to_mcontext(priv_mcontext_t *mc, sig_full_cxt_t *sc_full,
  * thread_get_mcontext() sets the dcontext isa_mode from cpsr, and thread_set_mcontext()
  * sets it.
  */
+/* XXX: on RISC-V, the tp register is a general-purpose register, but as its name
+ * suggests, it is conventionally used as the thread register.
+ * We save DR's tls by keeping tp's value unchanged, regardless in DR or in application.
```
derekbruening, line 2925: Is tp the stolen register?  But arm is writing the stolen register in these functions, and I think handling it in separate code (like set_sigcxt_stolen_reg()).  Is all code using these mcontext<->sigcontext conversion routines updating tp afterward and that's why it doesn't matter whether it's touched here?

derekbruening (approved):

core/unix/signal.c
ksco: It's not the stolen register. Per calling convention, `tp` is the thread-pointer register used to point to the thread-local storage. We also have a stolen register (default to x28) same as ARM/AArch64. The strategy is to keep tp unchanged, i.e. mangle every instruction that uses tp to something like:

```assembly
sd scratch_reg, scratch_slot(x28) ; x28 is the stolen reg
ld scratch_reg, app_tls_slot(x28)

; mangled instruction here (tp to scratch_reg)

sd scratch_reg, app_tls_slot(x28)
ld scratch_reg, scratch_slot(x28)
```

So the only difference is on RISC-V, tp is a general register while on ARM/AArch64, it's a special register, which is naturally unchanged.

And here in the mcontext<->sigcontext conversion, if we do not exclude tp, DR(host)'s tp register will be corrupted.

But I'm not sure if this is the correct way to do it after reading your comment. Could you give some advice?

derekbruening: So essentially `tp` is a 2nd stolen register.  For any stolen register we need to translate it to the app value when the kernel presents the `tp` register to the app, such as when passed to a signal handler.  Is that done today?  Everywhere we do things like `set_sigcxt_stolen_reg()` we need to also ensure `tp` has the right value.

I think these mcontext<->sigcontext conversion routines assume the caller will deal with translating stolen registers: they just blindly copy fields from one to the other.  So I would expect them to include `tp`.  The caller has to translate it to/from app/DR values if necessary.

core/arch/riscv64/mangle.c
ksco: On the RISC-V side, there actually no false positive case I think. There are indeed mov_constants in other mangle routines, but these values are unique at least for now. I added a XXX comment.

core/unix/signal.c
ksco: On second thought, the difference between tp and stolen_reg is that we cannot restore tp's value as there is no backup for it. And since tp is always pointing to host tls, we can't translate it to the app value in the signal handler, or we will lose its original value (host tls) -- it will be mangled if it's used by any instruction in the app.

derekbruening: > we cannot restore tp's value as there is no backup for it. And since tp is always pointing to host tls, we can't translate it to the app value in the signal handler, or we will lose its original value (host tls) 

For transparency we need to translate it to the app value in the signal handler.  We have to add storage of DR's value somewhere in the OS fields (as is done on other arches) (probably it is already stored somewhere today I would think).

ksco: > (probably it is already stored somewhere today I would think)

Yes, we have `dr_tls_base` in the OS field, but that's *DR tls*, not host tls (I just found out we call host tls `lib_tls_base`, sorry for the confusion). IIUC other arches like AArch64 don't have the problem here, since its `lib_tls_base` is stored in a special register `TPIDR_EL0` and is not affected by signal handler sifgreturn.

On the RISC-V side, how do we resume `lib_tls_base` back to tp if the kernel replaced it with app tls? (since we need to translate it to the app value in the signal handler).

ksco: Looking again at the callers of `set_sigcxt_stolen_reg()`, do we need to translate `tp` to the `app_tls` at the signal handler and then restore it back to `lib_tls_base` before `sigreturn`? If so, is this the transparency for clients or something?

derekbruening: Transparency for the app, as it can examine the machine state passed to its signal handler.  See https://dynamorio.org/transparency.html#sec_trans_pretend and the transparency slides in the DR tutorial.

ksco: Thanks, I reverted this change and added `set_sigcxt_tp_reg()` in [a4610b7](https://github.com/DynamoRIO/dynamorio/pull/6461/commits/a4610b74b9884c5dcd80cba36a00c18fd72a094a). Also enabled building of `linux.eintr` test, since it's working on real hardware. I think it's better to have another round of review.

core/arch/riscv64/mangle.c
```diff
@@ -272,8 +272,7 @@ patch_mov_immed_arch(dcontext_t *dcontext, ptr_int_t val, byte *pc, instr_t *fir
 bool
 instr_check_xsp_mangling(dcontext_t *dcontext, instr_t *inst, int *xsp_adjust)
 {
-    /* FIXME i#3544: Not implemented */
-    ASSERT_NOT_IMPLEMENTED(false);
+    /* Not apply to RISC-V. */
```
derekbruening, line 275: nit: s/Not/Does not/

derekbruening (approved):

ksco:
@derekbruening Could you merge this? Thanks.

ksco:
Yeah, I understand that. Thank you for the effort!

PR #6476, opened 2023-11-28
i#3544 RV64: Fix CI rsync issue

derekbruening:
Adds a check for the existence of a package install dir when manually copying packages for the cross-compile setup as it seems to not be present in newer versions of the packages.

Issue: #3544


ksco (approved):
Checked locally, `../extract/lib/` becomes empty now.

derekbruening:
x32 is AMD #6417
win64 is replacall #5412

PR #6477, opened 2023-11-28
i#3544 RV64: Add support for XTheadCmo and XTheadSync extensions

ksco:
Some of these instructions are suddenly used in newer OS release of my SBC LicheePi 4A, so added them to make it happy.

Issue: #3544


core/ir/riscv64/isl/xtheadcmo.txt
```diff
@@ -0,0 +1,14 @@
+# XTheadCmo (Cache Management Operations)
+# Version 1.0
+# https://github.com/T-head-Semi/thead-extension-spec
+
+# Only list instructions that are valid in usermode.
```
derekbruening, line 5: Note that we would like to cover privileged as well, to enable various uses like the standalone drdecode library for decoding arbitrary bits, or e.g. with hardware kernel tracing (currently only on x86) we end up stitching kernel instructions into drmemtrace traces.  How about a TODO comment.

derekbruening (approved):

core/ir/riscv64/isl/xtheadcmo.txt
ksco: > with hardware kernel tracing

This is cool! Added TODOs.

PR #6479, opened 2023-11-28
i#5505 kernel tracing: Add syscall instr encodings

abhinav92003:
Adds encodings for kernel system call instructions to the trace in raw2trace. Kernel system call traces are decoded using libipt which also provides the instruction encodings. We add support to drir_t to write these encodings to a new buffer which is re-used for all dynamic instances of that instr even across multiple system call traces.

Fixes taken/not-taken detection for conditional branches in the syscall trace.

Adds support in the syscall_mix tool to report the counts of each system call's traces also. Adds sysnum to system call trace start and end markers to achieve this.

Ran all Intel-PT tests locally:

```
$ ctest -VV -R 'SUDO'
...
The following tests passed:
	code_api|client.drpttracer_SUDO-test
	code_api|tool.drcachesim.phys_SUDO # not really PT. Just included because of ctest -R.
	code_api|tool.drcachesim.phys-threads_SUDO # not really PT. Just included because of ctest -R.
	code_api|tool.drcacheoff.phys_SUDO # not really PT. Just included because of ctest -R.
	code_api|tool.drcacheoff.kernel.simple_SUDO
	code_api|tool.drcacheoff.kernel.opcode-mix_SUDO
	code_api|tool.drcacheoff.kernel.syscall-mix_SUDO

100% tests passed, 0 tests failed out of 7

```

Found some flakiness due to #6486 in local runs of the kernel sudo tests, which will be addressed separately.

Issue: #5505


clients/drcachesim/tests/offline-kernel-opcode-mix.templatex
```diff
@@ -0,0 +1,7 @@
+Hello, world!
+Opcode mix tool results:
+.*: total executed instructions
+.*:.*
+.*
```
abhinav92003, line 5: Before this PR, syscall was at the top (all kernel instrs happened to be counted as the last encoding). So this test was useful for me currently. But to make it more useful for future changes, maybe we should check the presence of some kernel-only opcode?

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -1032,25 +1036,89 @@ raw2trace_t::process_syscall_pt(raw2trace_thread_data_t *tdata, uint64_t syscall
...
+            // treat this like a cache and clear previously stored encodings
+            // because any re-use in the decode address will confuse our write()
+            // logic (remember that we key encodings using the decode pc).
+            tdata->syscall_instr_encodings_.emplace_back(
+                new uint8_t[SYSCALL_PT_ENCODING_BUF_SIZE]);
+            tdata->syscall_next_encoding_offset_ = 0;
+        }
+        app_pc encode_pc =
+            &tdata->syscall_instr_encodings_.back()[tdata->syscall_next_encoding_offset_];
+        byte *ret = instr_encode_to_copy(dcontext_, instr, encode_pc, orig_pc);
```
derekbruening, line 1104: We already have a copy of the original encoding bits, so I don't think we need to re-encode somewhere else.  See comment above where I'm confused as to why we need multiple layers of copies.

clients/drcachesim/common/options.cpp
```diff
@@ -117,6 +117,14 @@ droption_t<std::string> op_alt_module_dir(
     "analysis tools, or in the raw modules file for post-prcoessing of offline "
     "raw trace files.  This directory takes precedence over the recorded path.");
 
+droption_t<bool> op_ignore_decode_failure(
+    DROPTION_SCOPE_FRONTEND, "ignore_decode_failure", false,
+    "Whether instruction decode failures should be ignored",
+    "Specifies whether failures to decode the instruction encodings should be ignored by "
+    "the analysis tools. This is useful especially when decoding kernel traces collected "
+    "by Intel-PT where decode failures may not indicate a critical issue and can be "
+    "moved past.");
```
derekbruening, line 126: I looked into pt2ir.cpp and I see this:
```
            /* TODO i#2103: Currently, the PT raw data may contain 'STAC' and 'CLAC'
             * instructions that are not supported by Dynamorio.
             */
```
If those are the only 2 missing opcodes, let's just add them to the DR decoder and then not need any of these options.  I suspect it will not be much more work to add those opcodes than to add and plumb through all these options.

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -1016,12 +1016,16 @@ raw2trace_t::process_syscall_pt(raw2trace_thread_data_t *tdata, uint64_t syscall
...
+        return false;
+    }
     /* Convert the DR IR to trace entries. */
+    addr_t sysnum =
+        pt_data->header[dynamorio::drmemtrace::PDB_HEADER_SYSNUM_IDX].sysnum.sysnum;
     std::vector<trace_entry_t> entries;
     trace_entry_t start_entry = { .type = TRACE_TYPE_MARKER,
                                   .size = TRACE_MARKER_TYPE_SYSCALL_TRACE_START,
-                                  .addr = 0 };
+                                  .addr = sysnum };
```
derekbruening, line 1028: The docs for this marker need to be updated to say what the value holds.

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -1032,25 +1036,89 @@ raw2trace_t::process_syscall_pt(raw2trace_thread_data_t *tdata, uint64_t syscall
     }
     trace_entry_t end_entry = { .type = TRACE_TYPE_MARKER,
                                 .size = TRACE_MARKER_TYPE_SYSCALL_TRACE_END,
-                                .addr = 0 };
+                                .addr = sysnum };
```
derekbruening, line 1039: The docs for this marker need to be updated to say what the value holds.

clients/drcachesim/tracer/raw2trace.cpp
derekbruening, line 1028: The view tool should be updated too, to print the marker value (ditto for _END).

derekbruening (commented):
Not sure about this new option: seems we should just update the decoder.  Also confused about the need for 3rd copy of the encodings.

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -1032,25 +1036,89 @@ raw2trace_t::process_syscall_pt(raw2trace_thread_data_t *tdata, uint64_t syscall
...
+                                        instrlist_t *ilist)
+{
+    instr_t *instr = instrlist_first(ilist);
+    while (instr != NULL) {
+        app_pc orig_pc = instr_get_app_pc(instr);
+        int instr_len = instr_length(dcontext_, instr);
+        if (tdata->syscall_instr_encodings_.empty() ||
+            tdata->syscall_next_encoding_offset_ + instr_len >=
+                SYSCALL_PT_ENCODING_BUF_SIZE) {
+            // We allocate new memory to store kernel instruction encodings in
```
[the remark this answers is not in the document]
abhinav92003: >  I see pt2ir is making a copy of the encoding for every instr today

You mean the following? https://github.com/DynamoRIO/dynamorio/blob/cd5521ce48238af7384c79716891640161131c3d/clients/drcachesim/drpt2trace/pt2ir.cpp#L387

That's pt2ir decoding the insn raw bits returned by libipt to create the instr_t used here (we need to create instr_t because the memory address reconstruction logic in ir2trace needs the full instr_t). libipt internally sets those raw bits; insn is discarded after we create the instr_t.

I guess getting the bits from kcore is also a possibility, but I wasn't sure if libipt is doing any other magic in its implementation. But since we have the raw bits given by libipt anyway, I thought there's value in just using that instead of going back into the kcore ourselves.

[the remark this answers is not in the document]
abhinav92003: I think it should be possible to reuse the insn bits here instead of encoding the instr again. Would still be better than getting them from the kcore again ourselves.

clients/drcachesim/tracer/raw2trace.cpp
abhinav92003: Done.

clients/drcachesim/tracer/raw2trace.cpp
abhinav92003: Done.

clients/drcachesim/tracer/raw2trace.cpp
abhinav92003: Done.

clients/drcachesim/common/options.cpp
abhinav92003: Done. #6484

clients/drcachesim/tracer/raw2trace.cpp
abhinav92003: Done. Removed the encode, reusing the bits from PT directly.

clients/drcachesim/tracer/raw2trace.h
```diff
@@ -1182,6 +1187,14 @@ class raw2trace_t {
...
     bool
     process_syscall_pt(raw2trace_thread_data_t *tdata, uint64_t syscall_idx);
+
+    /**
+     * Records the encodings for all instructions in the given ilist. Note that
+     * kernel system call traces gathered using Intel-PT are decoded using libipt
+     * which also provides the instr encodings.
+     */
+    bool
+    track_syscall_pt_encodings(raw2trace_thread_data_t *tdata, instrlist_t *ilist);
```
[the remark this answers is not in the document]
abhinav92003: Changed it to record_encoding in drir.h

clients/drcachesim/tools/opcode_mix.cpp
```diff
@@ -276,7 +283,9 @@ opcode_mix_t::print_results()
     std::sort(sorted.begin(), sorted.end(), cmp_val);
     for (const auto &keyvals : sorted) {
         std::cerr << std::setw(15) << keyvals.second << " : " << std::setw(9)
-                  << decode_opcode_name(keyvals.first) << "\n";
+                  << (keyvals.first == kInvalidOpcode ? "<INVALID>"
+                                                      : decode_opcode_name(keyvals.first))
```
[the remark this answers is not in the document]
abhinav92003: This is reverted now.

clients/drcachesim/tools/opcode_mix.cpp
```diff
@@ -64,19 +64,22 @@ namespace dynamorio {
 namespace drmemtrace {
 
 const std::string opcode_mix_t::TOOL_NAME = "Opcode mix tool";
+constexpr int kInvalidOpcode = -1;
```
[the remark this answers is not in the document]
abhinav92003: This is reverted now.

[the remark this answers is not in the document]
abhinav92003: This is reverted now.

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -1032,25 +1036,89 @@ raw2trace_t::process_syscall_pt(raw2trace_thread_data_t *tdata, uint64_t syscall
...
+            // logic (remember that we key encodings using the decode pc).
+            tdata->syscall_instr_encodings_.emplace_back(
+                new uint8_t[SYSCALL_PT_ENCODING_BUF_SIZE]);
+            tdata->syscall_next_encoding_offset_ = 0;
+        }
+        app_pc encode_pc =
+            &tdata->syscall_instr_encodings_.back()[tdata->syscall_next_encoding_offset_];
+        byte *ret = instr_encode_to_copy(dcontext_, instr, encode_pc, orig_pc);
+        if (ret == NULL) {
+            tdata->error = "Failed to encode a system call kernel instruction.";
```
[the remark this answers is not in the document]
abhinav92003: Removed the encoding step.

clients/drcachesim/tests/offline-kernel-opcode-mix.templatex
abhinav92003: Used clac.

clients/drcachesim/drpt2trace/test_simple.templatex
```diff
@@ -0,0 +1,14 @@
...
+          .*
+ \+20   L3 .* mov    \$0x0000000e -> %edx
+ \+25   L3 .* syscall  -> %rcx %r11
+ \+27   L3 .* mov    \$0x0000003c -> %eax
+ \+32   L3 .* mov    \$0x00000000 -> %edi
+ \+37   L3 .* syscall  -> %rcx %r11
+END 0x0000000000000000
+.*
+Number of Instructions: 8
+Number of Trace Entries: 8
```
abhinav92003, line 14: Will fix the missing newline in the next push.

clients/drcachesim/drpt2trace/ir2trace.cpp
```diff
@@ -55,16 +55,19 @@ namespace drmemtrace {
...
+    if (drir == nullptr || drir->get_ilist() == NULL) {
         return IR2TRACE_CONV_ERROR_INVALID_PARAMETER;
     }
-    instr_t *instr = instrlist_first(drir.get_ilist());
+    instr_t *instr = instrlist_first(drir->get_ilist());
+    bool prev_was_repstr = false;
     while (instr != NULL) {
         trace_entry_t entry = {};
+        entry.size = instr_length(GLOBAL_DCONTEXT, instr);
+        entry.addr = (uintptr_t)instr_get_app_pc(instr);
```
derekbruening, line 70: Prefer C++-style casts

clients/drcachesim/drpt2trace/ir2trace.h
```diff
@@ -92,7 +92,7 @@ class ir2trace_t {
      * error code.
      */
     static ir2trace_convert_status_t
-    convert(DR_PARAM_IN drir_t &drir, DR_PARAM_INOUT std::vector<trace_entry_t> &trace,
+    convert(DR_PARAM_IN drir_t *drir, DR_PARAM_INOUT std::vector<trace_entry_t> &trace,
```
derekbruening, line 95: If this pointer is DR_PARAM_IN, make it const.  If it gets modified it should not be _IN.

clients/drcachesim/drpt2trace/pt2ir.cpp
```diff
@@ -379,27 +379,13 @@ pt2ir_t::convert(DR_PARAM_IN const uint8_t *pt_data, DR_PARAM_IN size_t pt_data_
...
-             * instructions that are not supported by Dynamorio.
-             */
-            if (!instr_valid) {
-                /* The decode() function will not correctly identify the raw bits for
-                 * invalid instruction. So we need to set the raw bits of instr manually.
-                 */
-                instr_free_raw_bits(drir.get_drcontext(), instr);
-                instr_set_raw_bits(instr, insn.raw, insn.size);
-                instr_allocate_raw_bits(drir.get_drcontext(), instr, insn.size);
+            if (decode_from_copy(drir->get_drcontext(), insn.raw, (app_pc)insn.ip,
```
derekbruening, line 386: Prefer C++-style casts

clients/drcachesim/drpt2trace/pt2ir.cpp
```diff
@@ -412,7 +398,7 @@ pt2ir_t::convert(DR_PARAM_IN const uint8_t *pt_data, DR_PARAM_IN size_t pt_data_
                 }
 #endif
             }
-            drir.append(instr);
+            drir->append(instr, (app_pc)insn.ip, insn.size, insn.raw);
```
derekbruening, line 401: Prefer C++-style casts

clients/drcachesim/drpt2trace/drir.h
```diff
@@ -72,23 +78,79 @@ class drir_t {
...
+        return decode_pc_[orig_pc].first;
+    }
+
 private:
     void *drcontext_;
     instrlist_t *ilist_;
+#define SYSCALL_PT_ENCODING_BUF_SIZE (1024 * 1024)
+    std::unordered_map<app_pc, std::pair<app_pc, int>> decode_pc_;
+    std::vector<std::unique_ptr<uint8_t[]>> instr_encodings_;
+    size_t next_encoding_offset_ = 0;
```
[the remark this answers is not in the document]
abhinav92003: Each instr_encodings_ vector element is a byte buffer. next_encoding_offset_ is the next available offset into the last byte buffer in instr_encodings_.

Added comments.

clients/drcachesim/drpt2trace/drir.h
```diff
@@ -72,23 +78,79 @@ class drir_t {
...
+            next_encoding_offset_ = 0;
+        }
+        auto it = decode_pc_.find(orig_pc);
+        // We record the encoding only if we don't already have the same encoding for
+        // the given orig_pc.
+        if (it == decode_pc_.end() ||
+            // We confirm that the instruction encoding has not changed. Just in case
+            // the kernel is doing JIT.
+            it->second.second != instr_len ||
+            memcmp(it->second.first, encoding, it->second.second) != 0) {
```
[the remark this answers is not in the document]
abhinav92003: Done.

clients/drcachesim/drpt2trace/ir2trace.cpp
abhinav92003: Done

clients/drcachesim/drpt2trace/pt2ir.cpp
abhinav92003: Done.

clients/drcachesim/drpt2trace/pt2ir.cpp
abhinav92003: Done

clients/drcachesim/drpt2trace/ir2trace.h
abhinav92003: I could've made it const but instrlist_first doesn't have a const param.

abhinav92003: It doesn't get modified so I'll let it be DR_PARAM_IN.

clients/drcachesim/drpt2trace/test_simple.templatex
abhinav92003: Done.

clients/drcachesim/drpt2trace/drir.h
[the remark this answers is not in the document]
abhinav92003: Done.

Issue #6471, opened 2023-11-22
Model i/o and idle time in drmemtrace scheduler


derekbruening:
Today the drmemtrace scheduler packs work into the cpus with no idle time.  It detects context switches on long-latency syscalls but does not yet insert i/o wait times.  The plan is to add STATS_IDLE and insert idle time and then add i/o time where an input is blocked for a period of time.

Xref the original scheduler issue #5843.

derekbruening:
Closing as complete.  Further improvements can be filed separately.

PR #6481, opened 2023-11-29
i#6471 sched idle: Add replay fallback and fix

derekbruening:
Adds a fallback to return EOF when all outputs are past the last record in replay mode, to handle a schedule where not every input reaches EOF or has regions of interest.  Adds a unit test for this case, which hangs without the fix.

Fixes a record bug which seems to cause the final segment entry for an input to have a stop instruction ordinal prior to the input's real endpoint.  This is what caused hangs in actual usage of PR #6472 and led to adding the fallback just described (which is useful on its own).  On selecting a new input which turned out to be at EOF, the original code called close_schedule_segment() on *prev_input* for some reason.  That replaces the default -1 sentinel (read to EOF) with the current ordinal.  The code then tries again for a new input, but if it ends up finding the prior one it would keep executing that without a new schedule segment entry.  If that's the last input on that core, close_schedule_segment() is not called (the default is relied upon). I observed missing sentinels for one input in some runs and this is my theory as to what happened; those cases are gone now with this fix. Unfortunately it is difficult to create a unit test for this, but testing on the original hangs (without the fallback) show it fixed.

Issue: #6471


clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -2328,6 +2328,116 @@ test_replay_timestamps()
...
+    write_test_schedule(std::string record_fname)
+    {
+        // We duplicate test_scheduler_t but we have one input ending early before
+        // eof.
+        scheduler_t scheduler;
+        std::vector<schedule_record_t> sched0;
+        sched0.emplace_back(scheduler_t::schedule_record_t::VERSION, 0, 0, 0, 0);
+        sched0.emplace_back(scheduler_t::schedule_record_t::DEFAULT, 0, 0, 4, 11);
+        // There is a huge time gap here.
+        sched0.emplace_back(scheduler_t::schedule_record_t::DEFAULT, 2, 7,
```
abhinav92003, line 2345: Re-iterate in comment here that "Max numeric value means continue until EOF."

clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -2328,6 +2328,116 @@ test_replay_timestamps()
...
+        sched1.emplace_back(scheduler_t::schedule_record_t::DEFAULT, 2, 4, 7, 60);
+        sched1.emplace_back(scheduler_t::schedule_record_t::DEFAULT, 3, 0, 4, 30);
+        sched1.emplace_back(scheduler_t::schedule_record_t::DEFAULT, 0, 4, 7, 40);
+        sched1.emplace_back(scheduler_t::schedule_record_t::DEFAULT, 1, 4, 7, 50);
+        sched1.emplace_back(scheduler_t::schedule_record_t::DEFAULT, 3, 4, 7, 70);
+        sched1.emplace_back(scheduler_t::schedule_record_t::DEFAULT, 0, 7,
+                            0xffffffffffffffffUL, 80);
+        sched1.emplace_back(scheduler_t::schedule_record_t::DEFAULT, 1, 7,
+                            0xffffffffffffffffUL, 90);
+        // Input 3 never reaches EOF (end is exclusive).
```
abhinav92003, line 2363: "never reaches eof"  because there's no schedule entry with max numeric value as end_instruction or the actual last instruction ordinal in input-3?

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -2276,17 +2279,26 @@ template <typename RecordType, typename ReaderType>
...
 {
     if (options_.mapping == MAP_TO_CONSISTENT_OUTPUT ||
-        live_input_count_.load(std::memory_order_acquire) == 0) {
+        live_input_count_.load(std::memory_order_acquire) == 0 ||
+        // While a full schedule recorded should have each input hit either its
+        // EOF or ROI end, we have a fallback to avoid hangs for possible recorded
+        // schedules that end an input early deliberately without an ROI.
+        (options_.mapping == MAP_AS_PREVIOUSLY &&
+         live_replay_output_count_.load(std::memory_order_acquire) == 0)) {
         return sched_type_t::STATUS_EOF;
```
abhinav92003, line 2302: Also assert for MAP_AS_PREVIOUSLY, we return STATUS_EOF iff at_eof is set.

abhinav92003 (approved):

clients/drcachesim/scheduler/scheduler.cpp
abhinav92003: Maybe worth asserting that at_eof is set if we're returning STATUS_EOF?

derekbruening: Done.

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: Done.

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: Yes, the furthest endpoint for intput#3 is 8, when its actual end is 9, so it never reaches 9.  Augmented the comment to explicitly say this.

derekbruening:
x32 failure is AMD #6417

PR #6482, opened 2023-11-29
Minor changes to permit building with older toolchains.

egrimley-arm:
Make some code conditional on EM_RISCV or PR_SVE_GET_VL being defined. This was needed on RHE7, which will be around for a few more years.

We also check for another older macro not being defined in case all the macros get replaced by an enum.


AssadHashmi:
The x86 failures look like https://github.com/DynamoRIO/dynamorio/issues/6417

derekbruening (approved):

prasun3:
On RHEL 7, `/usr/include/elf.h` does not define EM_RISCV, but now we have `third_party/elfutils/libelf/elf.h` that does define it. Could we use the latter header instead of the former?

derekbruening:
> On RHEL 7, `/usr/include/elf.h` does not define EM_RISCV, but now we have `third_party/elfutils/libelf/elf.h` that does define it. Could we use the latter header instead of the former?

Submodules used to be optional but we did make elfutils required on Linux so yes that seems reasonable.

prasun3:
> On RHEL 7, `/usr/include/elf.h` does not define EM_RISCV, but now we have `third_party/elfutils/libelf/elf.h` that does define it. Could we use the latter header instead of the former?

@egrimley-arm do you want to try this approach if you are still planning to work on this issue

egrimley-arm:
The problem with `EM_RISCV` was solved by 827791f26712302f0c6481520187fa00f954da25.

The problem with `PR_SVE_GET_VL` is presumably specific to AArch64. Perhaps nobody is using arm64 RHE7 any more.

PR #6489, opened 2023-12-01
i#3544 RV64: Adds immediate display format via IR

ksco:
Instructions created by `INSTR_CREATE_*` macros now automatically set the immediate display format. Also enables tool.drdisas test.

Issue #3544


suite/tests/api/ir_riscv64.expect
```diff
@@ -311,15 +312,15 @@ sd     a0 -> (a1)[8byte]
 sb     a0 -> (a1)[1byte]
 sh     a0 -> (a1)[2byte]
 addi   a1 0 -> a0
-jalr   a0 0x0 -> zero
```
abhinav92003, line 314: Isn't it nicer if they are hex?

abhinav92003 (approved):

suite/tests/api/ir_riscv64.expect
ksco: Ah, I simply followed the objdump style:

```
$ objdump -b binary -m riscv -D testfile

testfile:     file format binary


Disassembly of section .data:

0000000000000000 <.data>:
   0:	064500e7          	jalr	100(a0)
```

PR #6493, opened 2023-12-04
i#3544 RV64: Enable sample.bbbuf test

ksco:
This patch enables sample.bbbuf test with the following fixes:

- x0 on RISC-V is like XZR on AArch64, so do not consider it as a general register
- tp should be avoided in `drreg_reserve_reg_internal()`
- RISC-V does not have control flags, so a dummy hello test is used for `control_flags` instead

Also removed `linux.sigaction_nosignals` and `pthreads.pthreads` from `RUNS_ON_QEMU` since they are not working on real hardware.

#3544

### DIFF
--- a/suite/tests/CMakeLists.txt
+++ b/suite/tests/CMakeLists.txt
@@ -5911,9 +5911,7 @@ if (RISCV64)
     code_api|linux.infinite
     code_api|linux.longjmp
     code_api|linux.prctl
-    code_api|linux.sigaction_nosignals
     code_api|linux.signalfd
-    code_api|pthreads.pthreads
     code_api|pthreads.pthreads_exit
     code_api|sample.bbbuf
     code_api|security-linux.trampoline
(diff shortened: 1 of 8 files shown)


suite/tests/CMakeLists.txt
```diff
@@ -5910,10 +5911,9 @@ if (RISCV64)
     code_api|linux.infinite
     code_api|linux.longjmp
     code_api|linux.prctl
-    code_api|linux.sigaction_nosignals
```
abhinav92003, line 5913: They're not working on real hardware but they're working on QEMU? How? Would be good to keep the QEMU run if they're indeed working fine there.

abhinav92003 (approved):

suite/tests/CMakeLists.txt
ksco: > if they're indeed working fine there

It's strange, so I'm not sure if they're working as expected, but will add them back once figure out why.

ksco:
@abhinav92003 Can you merge this? Thanks.

abhinav92003:
x86-32 failures are #6417.

Issue #6081, opened 2023-05-24
ASSERT callback.c:1477 in tool.drcacheoff.gencode test on win32


derekbruening:
This happened 3x in a row (else it would have passed) -- yet I don't think we've ever seen this before.  This test passed in other runs of the same code so it is non-deterministic -- maybe in some environment it becomes repeatable.

https://github.com/DynamoRIO/dynamorio/actions/runs/5073007903/jobs/9111470544
```
275:  pre-DR start
275: 
275:   <Application
275:   D:\a\dynamorio\dynamorio\build_debug-internal-32\suite\tests\bin\tool.drcacheoff.gencode.exe
275:   (7024).  Internal Error: DynamoRIO debug check failure:
275:   D:\a\dynamorio\dynamorio\core\win32\callback.c:1477 false
275: 
275:   (Error occurred @77 frags in tid 6752)
```

derekbruening:
This is now at callback.c:1480 FTR.  Happened again in windows-timestamps at https://github.com/DynamoRIO/dynamorio/actions/runs/7123468151/job/19396119425?pr=6494

edeiana:
Hitting the same callback.c:1480 debug check failure (PR #6698).
(Note: it only happened once, it usually passes).

`2024-03-12T05:50:34.2666884Z 268/310 Test #271: code_api|tool.drcacheoff.windows-timestamps ......................***Failed    0.18 sec
2024-03-12T05:50:34.2668775Z Running cmd |D:/a/dynamorio/dynamorio/build_debug-internal-64/clients/bin64/tool.drcacheoff.windows-timestamps.exe|
2024-03-12T05:50:34.2670610Z CMake Error at D:/a/dynamorio/dynamorio/suite/tests/runmulti.cmake:114 (message):
2024-03-12T05:50:34.2671691Z   *** cmd failed (-1): <Application
2024-03-12T05:50:34.2672944Z   D:\a\dynamorio\dynamorio\build_debug-internal-64\clients\bin64\tool.drcacheoff.windows-timestamps.exe
2024-03-12T05:50:34.2674322Z   (4156).  Internal Error: DynamoRIO debug check failure:
2024-03-12T05:50:34.2675291Z   D:\a\dynamorio\dynamorio\core\win32\callback.c:1480 false
2024-03-12T05:50:34.2675977Z 
2024-03-12T05:50:34.2676322Z   (Error occurred @55 frags in tid 5312)
[...]

PR #6494, opened 2023-12-05
i#6471 sched idle: Add idle time from blocking syscalls

derekbruening:
Adds a longer waiting period on blocking syscalls using either provided output time or for instruction-based quanta a count of queue selections before a blocked input is actually selected.  Since the scheduler does not have timer interrupts or regular points of control and relies on its user calling it, idle inputs are kept on the ready queue and are checked for becoming unblocked when the ready queue is queried.

The wait duration is set based on the "wait time factor" which is the syscall latency divided by the context switch threshold multipled by a user-provided "block_time_scale" option which can be used to scale up or down the durations.

The wait duration is erased on a direct switch to an input.

Adds a new replay record type to represent idle time on replay.

Augments the unit tests to include blocking high-latency syscalls to test the new feature in various sub-tests.

Issue: #6471

### DIFF
--- a/clients/drcachesim/scheduler/scheduler.cpp
+++ b/clients/drcachesim/scheduler/scheduler.cpp
@@ -744,7 +744,12 @@ scheduler_tmpl_t<RecordType, ReaderType>::set_initial_schedule(
             }
             for (int i = 0; i < static_cast<output_ordinal_t>(outputs_.size()); ++i) {
                 if (i < static_cast<input_ordinal_t>(inputs_.size())) {
-                    input_info_t *queue_next = pop_from_ready_queue(i);
+                    input_info_t *queue_next;
+#ifdef DEBUG
+                    sched_type_t::stream_status_t status =
+#endif
+                        pop_from_ready_queue(i, queue_next);
+                    assert(status == STATUS_OK); // No blocked inputs yet.
                     if (queue_next == nullptr)
                         set_cur_input(i, INVALID_INPUT_ORDINAL);
                     else
@@ -860,10 +865,17 @@ scheduler_tmpl_t<RecordType, ReaderType>::read_recorded_schedule()
         return STATUS_ERROR_INVALID_PARAMETER;
     }
     for (int i = 0; i < static_cast<output_ordinal_t>(outputs_.size()); ++i) {
-        if (!outputs_[i].record.empty()) {
-            set_cur_input(i, outputs_[i].record[0].key.input);
-        } else
+        if (outputs_[i].record.empty()) {
             set_cur_input(i, INVALID_INPUT_ORDINAL);
+        } else if (outputs_[i].record[0].type == schedule_record_t::IDLE) {
+            set_cur_input(i, INVALID_INPUT_ORDINAL);
+            outputs_[i].waiting = true;
+            outputs_[i].wait_start_time = 0; // Updated on first next_record().
+            VPRINT(this, 3, "output %d starting out idle\n", i);
+        } else {
+            assert(outputs_[i].record[0].type == schedule_record_t::DEFAULT);
+            set_cur_input(i, outputs_[i].record[0].key.input);
+        }
     }
     return STATUS_SUCCESS;
 }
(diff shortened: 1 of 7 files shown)


clients/drcachesim/scheduler/scheduler.h
```diff
@@ -522,6 +522,17 @@ template <typename RecordType, typename ReaderType> class scheduler_tmpl_t {
          * blocking and trigger a context switch.
          */
         uint64_t blocking_switch_threshold = 100;
+        /**
+         * Controls the amount of time inputs are considered blocked at a syscall whose
+         * latency exceeds #syscall_switch_threshold or #blocking_switch_threshold.  A
+         * "block time factor" is computed from the syscall latency divided by either
+         * #syscall_switch_threshold or #blocking_switch_threshold.  This factor is
```
abhinav92003, line 529: These thresholds are applied in raw2trace whereas the block_time_scale is applied post-raw2trace. If the thresholds change after raw2trace (and therefore the maybe_blocking markers) would that be okay?

clients/drcachesim/scheduler/scheduler.h
```diff
@@ -522,6 +522,17 @@ template <typename RecordType, typename ReaderType> class scheduler_tmpl_t {
...
+         * multiplied by this field #block_time_scale.  For #QUANTUM_TIME, that amount of
+         * time, as reported by the time parameter to next_record(), must pass before the
+         * input is no longer considered blocked.  For instruction quanta, that count of
```
abhinav92003, line 532: Do you instead mean scheduler selections equal to the instruction_quanta times the factor must occur?

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -928,7 +940,7 @@ scheduler_tmpl_t<RecordType, ReaderType>::read_traced_schedule()
         // a signal can come in after a prior timestamp with no intervening instrs).
         if (!all_sched[cur_output].empty() &&
             input == all_sched[cur_output].back().key.input &&
-            start == all_sched[cur_output].back().start_instruction) {
+            start == all_sched[cur_output].back().value.start_instruction) {
```
abhinav92003, line 943: Do we need to confirm whether all_sched[cur_output].back() is not an IDLE entry? Same elsewhere.

abhinav92003 (approved):

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -744,7 +744,12 @@ scheduler_tmpl_t<RecordType, ReaderType>::set_initial_schedule(
             }
             for (int i = 0; i < static_cast<output_ordinal_t>(outputs_.size()); ++i) {
                 if (i < static_cast<input_ordinal_t>(inputs_.size())) {
-                    input_info_t *queue_next = pop_from_ready_queue(i);
+                    input_info_t *queue_next;
+#ifdef DEBUG
+                    sched_type_t::stream_status_t status =
+#endif
+                        pop_from_ready_queue(i, queue_next);
+                    assert(status == STATUS_OK); // No blocked inputs yet.
```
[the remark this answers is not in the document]
abhinav92003: But wait... asserts still work in non-debug builds, so how is this building successfully if there's no `status` in non-debug.

clients/drcachesim/scheduler/scheduler.h
derekbruening: Yes, "that count" is the final number referenced at line 530 "This factor is multiplied by this field block_time_scale".  Reworded to call it "a final value" and s/that count/that final value/ so it's clearer.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Here it cannot be, by construction: this is an as-traced schedule from which we construct all_sched from only DEFAULT types.  I added an assert here as a sanity check and another one in the check_and_fix_* routine.

derekbruening:
In further testing on large traces, I am actually having trouble getting enough idle time.  I can get a number of idle periods but they never add up to a large % of the instruction count.  (With large core count I can get plenty of idle time.).  I can't get arbitrary idle times by raising the scale: the scheduler gets slow before I reach idle % targets.  But, I think it is worth merging this as-is for now to get some base blocked-idle time in there for base testing, even if I figure out how to improve it later.

derekbruening:
x32 is amd #6417 
win64 is windows-timestamps callback.c assert #6081

clients/drcachesim/scheduler/scheduler.cpp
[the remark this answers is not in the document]
derekbruening: Grr, on Github there are no builds that have assert enabled but don't have DEBUG.  But in our internal builds, we do have such a mode (and my pre-merge testing, which was done here, only tests release build and so didn't see it).

PR #6497, opened 2023-12-07
i#6471 sched idle: Fix assert define

derekbruening:
Fixes an assert variable decl to use #ifndef NDEBUG instead of #ifdef DEBUG to support builds in between DEBUG and NDEBUG.  We don't have such builds on Github but some third-party builders use them.

Issue: #6471


derekbruening:
Putting this one in as TBR.

abhinav92003 (commented):
LGTM

PR #6498, opened 2023-12-07
i#3544 RV64: Enable sample.bbcount test

ksco:
This patch enables sample.bbbuf test with the following changes:

- Adds `IF_AARCHXX_OR_RISCV64*` macros so they can share some code
- Implements `drx_insert_counter_update()` without `DRX_COUNTER_REL_ACQ` support

Issue: #3544


abhinav92003:
x86-32 failures are #6417.

Issue #6452, opened 2023-11-13
client.attach_test failing sometimes: kill cannot find the target process


derekbruening:
Split from #5725.  This still sometimes happens, here on x64 Linux:

https://github.com/DynamoRIO/dynamorio/actions/runs/6855254564/job/18639921367?pr=6445
```
183/476 Test #183: code_api|client.attach_test ......................................***Failed  Required regular expression not found. Regex=[^starting
thank you for testing attach
thread init
done
$
]  0.25 sec
CMake Error at /home/runner/work/dynamorio/dynamorio/suite/tests/runall.cmake:124 (message):
  *** kill failed (1): kill: (24386): No such process

  ***
```

PR #6502, opened 2023-12-11
i#5843 scheduler: Accumulate quanta across runs

derekbruening:
Changes the quanta accounting to match the real kernel by accumulating it across executions if a prior execution was terminated early due to a voluntary context switch.

Adds new testing, and updates old tests with the behavior change. Scheduler unit test string changes were carefully vetted.  E.g., for test_synthetic_with_syscalls_multiple(): the output strings changed because H's quantum accumulates and it hits a preempt in the middle of its second HH sequence, which decrements B's quantum, causing B to become available sooner.

Issue: #5843

### DIFF
--- a/clients/drcachesim/scheduler/scheduler.cpp
+++ b/clients/drcachesim/scheduler/scheduler.cpp
@@ -1724,8 +1724,7 @@ scheduler_tmpl_t<RecordType, ReaderType>::set_cur_input(output_ordinal_t output,
     if (prev_input == input)
         return STATUS_OK;
     std::lock_guard<std::mutex> lock(*inputs_[input].lock);
-    inputs_[input].instrs_in_quantum = 0;
-    inputs_[input].start_time_in_quantum = outputs_[output].cur_time;
+    inputs_[input].prev_time_in_quantum = outputs_[output].cur_time;
     if (options_.schedule_record_ostream != nullptr) {
         uint64_t instr_ord = inputs_[input].reader->get_instruction_ordinal();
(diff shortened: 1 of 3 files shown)


clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -2335,6 +2340,15 @@ scheduler_tmpl_t<RecordType, ReaderType>::next_record(output_ordinal_t output,
                 lock.lock();
                 VPRINT(this, 5, "next_record_mid[%d]: switching from %d to %d\n", output,
                        prev_input, outputs_[output].cur_input);
+                if (!preempt) {
+                    if (options_.quantum_unit == QUANTUM_INSTRUCTIONS &&
+                        record_type_is_instr(record)) {
+                        --inputs_[prev_input].instrs_in_quantum;
+                    } else if (options_.quantum_unit == QUANTUM_TIME) {
+                        inputs_[prev_input].time_in_quantum -=
```
abhinav92003, line 2348: Looks like time_in_quantum is really the time_spent_in_quantum? And prev_time_in_quantum is the snapshot of the last timestamp? Maybe rename s/time_in_quantum/time_spent_in_quantum/ to make the distinction clearer?

abhinav92003 (approved):

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Done.

derekbruening:
x64 failure is client.attach_test not finding the process to kill: #6452
x32 is #6417

Issue #6495, opened 2023-12-06
Inject system call kernel trace templates in raw2trace


abhinav92003:
Today drmemtrace supports tracing kernel execution during system calls using `-enable_kernel_tracing`, but this feature is available only on x86 machines with the Intel-PT  feature. Another possible way to get user+kernel traces is to collect templates for system call kernel execution using QEMU/Gem5 etc and then inject them using raw2trace at the system call marker locations in the user-space trace.

Such traces would not have the actual kernel execution done during system calls, but they may still be useful to model certain aspects of the application's execution. E.g. a cache simulator may yield better results for user-space MPKI because the kernel interference is modelled to some extent. More experimentation should reveal how useful this method can be.

To start, we can collect and inject only one trace template per system call, but later we may collect multiple trace templates for each system call, differentiated perhaps by its arguments or the return value and inject the appropriate one in the trace during raw2trace. We may also adjust some aspects of the trace template while injecting it; e.g. use a different base address for the kernel code or data accesses.

PR #6503, opened 2023-12-12
i#6495 syscall templates: Fix format specifier in log

abhinav92003:
Fixes print format specifier in a raw2trace verbose level 2 log. This causes a segmentation fault with -verbose 2.

Issue: #6495


derekbruening (approved):

abhinav92003:
x86-32 failures are #6417.

PR #6504, opened 2023-12-12
i3544 RV64: Enable a few more sample tests

ksco:
This patch fixes a bug in `insert_pop_all_registers()` and thus enables the following sample tests:

code_api|sample.bbsize
code_api|sample.div
code_api|sample.empty
code_api|sample.inline
code_api|sample.inscount
code_api|sample.opcode_count
code_api|sample.signal
code_api|sample.stl_test
code_api|sample.syscall

Issue: #3544


abhinav92003 (approved):

abhinav92003:
We have made x86-32 a non-blocking check, so you should be able to merge this yourself after updating the branch.

PR #6505, opened 2023-12-13
i#6471 idle: Add better idle time modeling

derekbruening:
Changes instruction quantum idle handling to use wall-clock time, instead of a counter decremented in queue pops.  The counter was skewed unfairly by the rate of queue queries.  This results in all quanta using a unified time-based approach for blocked time.

Changes block_time_scale to remove the division by the latency threshold.  Now the scale is directly multiplied by the latency to result in the blocked time units.  The default scale is set to 1000 which matches the wall-clock time to process an instruction by the schedule_stats tool (about 1 instruction per microsecond) and which should be a rough match for many simulators passing one nanosecond cycle per instruction as the time unit.

Adds a new scheduler option block_time_max and drcachesim option -sched_block_max which caps the blocked time for a syscall (default 25 seconds) to avoid outliers being scaled to extreme amounts of time.

Adds a heartbeat of queue lookups, currently only in debug build, to help understand behavior over time.

Sets the input ordinal to invalid while idle.  This also causes schedule_stats to consider a transition from idle to a valid input to be a context switch, which matches how the Linux kernel counts switches.

Augments schedule_stats with a wall-clock measure of cpu and idle time for a much fairer %cpu metric than the previous one based purely on record counts, as schedule_stats processes an instruction much more quickly than an idle record.  Adds two %cpu metrics: one that does not include idle time past the final instruction (for skewed finishes across cores) and one that includes idle time until all cpus are done.

Increases the default -schedule_stats_print_every to 500K to avoid extremely long strings for larger workloads.

Updates the scheduler unit tests for the timing changes, changing the ones testing idle time to use a deterministic mock time quantum to avoid wall-clock time flakiness.

Tested on large traces on dozens of cores with known idle characteristics where by tweaking the parameters I was able to get reasonably representative idle times.

Issue: #6471

### DIFF
--- a/clients/drcachesim/common/options.cpp
+++ b/clients/drcachesim/common/options.cpp
@@ -851,11 +851,18 @@ droption_t<uint64_t> op_sched_blocking_switch_us(
     "maybe-blocking to incur a context switch. Applies to -core_sharded and "
     "-core_serial. ");
 
-droption_t<double>
-    op_sched_block_scale(DROPTION_SCOPE_ALL, "sched_block_scale", 1.,
-                         "Input block time scale factor",
-                         "A higher value here results in blocking syscalls "
-                         "keeping inputs unscheduled for longer.");
+droption_t<double> op_sched_block_scale(
+    DROPTION_SCOPE_ALL, "sched_block_scale", 1000., "Input block time scale factor",
+    "The scale applied to the microsecond latency of blocking system calls.  A higher "
+    "value here results in blocking syscalls keeping inputs unscheduled for longer.  "
+    "This should roughly equal the slowdown of instruction record processing versus the "
+    "original (untraced) application execution.");
+
+droption_t<uint64_t> op_sched_block_max_us(DROPTION_SCOPE_ALL, "sched_block_max_us",
+                                           25000000,
+                                           "Maximum blocked input time, in microseconds",
+                                           "The maximum blocked time, after scaling with "
+                                           "-sched_block_scale.");
 #ifdef HAS_ZIP
 droption_t<std::string> op_record_file(DROPTION_SCOPE_FRONTEND, "record_file", "",
                                        "Path for storing record of schedule",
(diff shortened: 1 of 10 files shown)


derekbruening:
The wall-clock time usage for instr quanta is causing non-determinism in the tests (despite switching most to the mock time, we need a few using instr).  Will have to relax their checks.

derekbruening:
x64 failure is client.attach_test #6452

clients/drcachesim/common/options.cpp
```diff
@@ -851,11 +851,18 @@ droption_t<uint64_t> op_sched_blocking_switch_us(
...
-                         "keeping inputs unscheduled for longer.");
+droption_t<double> op_sched_block_scale(
+    DROPTION_SCOPE_ALL, "sched_block_scale", 1000., "Input block time scale factor",
+    "The scale applied to the microsecond latency of blocking system calls.  A higher "
+    "value here results in blocking syscalls keeping inputs unscheduled for longer.  "
+    "This should roughly equal the slowdown of instruction record processing versus the "
+    "original (untraced) application execution.");
+
+droption_t<uint64_t> op_sched_block_max_us(DROPTION_SCOPE_ALL, "sched_block_max_us",
+                                           25000000,
```
abhinav92003, line 862: How was the default decided?

abhinav92003: I see there's a comment at clients/drcachesim/scheduler/scheduler.cpp:1681. Maybe repeat some of it here?

abhinav92003: Do we need this to be a flag? Or can it be a constant.

abhinav92003 (approved):

clients/drcachesim/common/options.cpp
derekbruening: Added the comment you pointed at, plus added a statement about needing to tweak this for different workloads and simulator speeds, and a statement about the default being picked while tuning a 1-minute-long schedule_stats run on a 112-core 500-thread large application to produce good cpu usage without unduly increasing tool runtime.

derekbruening:
x64 failure is client.attach_test #6452
win64 failure is burst_traceopts #6423
x32 failure is AMD #6417

PR #6512, opened 2023-12-17
i#5994 opcode_mix, part1: add categories to opcode_mix output

edeiana:
Groups instructions that belong to the same categories together, and prints (to std::cerr) a count of them at the end of opcode counts.

Issue: #5994


clients/drcachesim/tools/opcode_mix.cpp
```diff
@@ -229,6 +229,20 @@ opcode_mix_t::parallel_shard_memref(void *shard_data, const memref_t &memref)
         instr_free(dcontext_.dcontext, &instr);
     }
     ++shard->opcode_counts[opcode];
+
+    uint category;
+    instr_t instr;
+    instr_init(dcontext_.dcontext, &instr);
+    app_pc next_pc = decode_from_copy(dcontext_.dcontext, decode_pc, trace_pc, &instr);
```
derekbruening, line 236: Decoding is expensive and we took some effort above to cache the results, so let's leverage the same caching by using the same decode call as above and storing the categories in the (maybe renamed) opcode_cache.

core/ir/instr_api.h
```diff
@@ -738,6 +738,19 @@ DR_API
 uint
 instr_get_category(instr_t *instr);
 
+DR_API
+/**
+ * Fills \p category_names with a comma-separated string of categories.
```
derekbruening, line 743: Hmm, it seems more flexible to instead provide a string name for an individual category so the user can combine them in whatever way they want, rather than hardcoding commas or a certain ordering?

clients/drcachesim/tools/opcode_mix.h
```diff
@@ -103,6 +103,7 @@ class opcode_mix_t : public analysis_tool_t {
         worker_data_t *worker;
         int64_t instr_count;
         std::unordered_map<int, int64_t> opcode_counts;
+        std::unordered_map<uint, int64_t> category_counts;
         std::string error;
         app_pc last_trace_module_start;
         size_t last_trace_module_size;
```
derekbruening, line 109: Please add a test.  Maybe something like:
```
      torunonly_drcacheoff(opcode_categories allasm_x86_64
        "" "@-simulator_type@opcode_mix" "")
```
That tiny app has a small fixed instr sequence so you can hardcode the category output into a template file `clients/drcachesim/tests/offline-opcode_categories.templatex`.

derekbruening (commented):
Thank you for contributing!  I have some suggestions.

clients/drcachesim/tools/opcode_mix.cpp
edeiana: Done.

clients/drcachesim/tools/opcode_mix.h
edeiana: Done.

clients/drcachesim/tests/offline-opcode_categories.templatex
```diff
@@ -0,0 +1,32 @@
...
+             17 :   syscall
+             16 :       cmp
+             16 :       jnz
+             16 :       sub
+             16 :       lea
+              1 :       and
+
+              4 : sets of categories
+             51 :      move
+             33 :    branch
```
derekbruening, line 30: (Interesting to see category results: this is going to be a good tool addition.

Hmm, to get 33 it must count `syscall` in addition to `jnz`?  I guess a system call gateway is a kind of branch.  Would everyone call it a branch?  Sorry, I realize this discussion is not really in the scope of this PR: just interesting; not clearly a problem so I guess we leave it and see if someone complains.)

clients/drcachesim/tools/opcode_mix.h
```diff
@@ -83,7 +84,7 @@ class opcode_mix_t : public analysis_tool_t {
 
 protected:
     struct worker_data_t {
-        std::unordered_map<app_pc, int> opcode_cache;
+        std::unordered_map<app_pc, std::pair<int, uint>> opcode_category_cache;
```
derekbruening, line 87: Hmm I was going to say `s/uint/dr_instr_category_t/` but this combines them.  Add a comment on why it's a uint and not the enum type?

core/ir/instr_api.h
```diff
@@ -2539,17 +2548,17 @@ instr_is_reg_spill_or_restore(void *drcontext, instr_t *instr, bool *tls DR_PARA
...
-#    define EFLAGS_READ_AF 0x00000004  /**< Reads AF (Auxiliary Carry Flag). */
-#    define EFLAGS_READ_ZF 0x00000008  /**< Reads ZF (Zero Flag). */
-#    define EFLAGS_READ_SF 0x00000010  /**< Reads SF (Sign Flag). */
-#    define EFLAGS_READ_TF 0x00000020  /**< Reads TF (Trap Flag). */
-#    define EFLAGS_READ_IF 0x00000040  /**< Reads IF (Interrupt Enable Flag). */
-#    define EFLAGS_READ_DF 0x00000080  /**< Reads DF (Direction Flag). */
-#    define EFLAGS_READ_OF 0x00000100  /**< Reads OF (Overflow Flag). */
-#    define EFLAGS_READ_NT 0x00000200  /**< Reads NT (Nested Task). */
-#    define EFLAGS_READ_RF 0x00000400  /**< Reads RF (Resume Flag). */
+#    define EFLAGS_READ_CF 0x00000001 /**< Reads CF (Carry Flag). */
```
derekbruening, line 2551: (I guess your clang-format is different -- but if the GA CI passed someone else would hit this anyway as it must be an older version vs newer.  Small enough to not need separation/tweaking; seems ok to leave in this PR.)

clients/drcachesim/tools/opcode_mix.h
derekbruening, line 87: (Optional (wish Github had a way to mark things as optional): use a new struct with meaningful names rather than std::pair.  This is not in our Github style guide so optional.)

clients/drcachesim/tools/opcode_mix.cpp
```diff
@@ -255,6 +259,57 @@ cmp_val(const std::pair<int, int64_t> &l, const std::pair<int, int64_t> &r)
...
+{
+    std::string category_name;
+    if (category == DR_INSTR_CATEGORY_UNCATEGORIZED) {
+        category_name += instr_get_category_name(DR_INSTR_CATEGORY_UNCATEGORIZED);
+        return category_name;
+    }
+    if (TESTANY(DR_INSTR_CATEGORY_FP, category)) {
+        category_name += " ";
+        category_name += instr_get_category_name(DR_INSTR_CATEGORY_FP);
+    }
```
derekbruening, line 273: How about a loop that marches through all the bits, to generalize this code and avoid us having to edit it when a new category is added?   Maybe don't even need a max named bit: going to the 32nd bit seems ok to future-proof.

derekbruening (commented):
Looking good, but I had a few more suggestions.

core/ir/instr_shared.c
```diff
@@ -470,6 +470,25 @@ instr_get_category(instr_t *instr)
 /* in rest of file, directly de-reference for performance (PR 622253) */
 #define instr_get_category inlined_instr_get_category
 
+const char *
+instr_get_category_name(dr_instr_category_t category)
```
derekbruening, line 474: We should have at least a sanity test of a new API in the DR core's tests (clients/drcachesim is considered separate from the "core").  Please add some spot-checks of category strings, maybe in test_categories() in suite/tests/api/drdecode_x86.c.

derekbruening, line 474: Please add the new API to the changelist at https://github.com/DynamoRIO/dynamorio/blob/master/api/docs/release.dox#L195

derekbruening (commented):
Two more requests.  (Sorry, not trying to "move the goalposts": should have put these in the first round of reviews.)

clients/drcachesim/tests/offline-opcode_categories.templatex
edeiana: Right, https://github.com/DynamoRIO/dynamorio/blob/master/core/ir/x86/decode_table.c#L2467 categorized `syscall` as a branch.
Not really sure if that's a good semantics, but as pointed out, it's outside the scope of this PR.

clients/drcachesim/tools/opcode_mix.h
edeiana: Done, I changed it to a more meaningful structure as suggested.

core/ir/instr_api.h
edeiana: Yes, this is a clang-format issue. I'm using version 9.0.0, not sure if that's old or new in tis context. I can always revert back if you'd like me to!

clients/drcachesim/tools/opcode_mix.h
edeiana: Done.

clients/drcachesim/tools/opcode_mix.cpp
edeiana: Whoops, this makes so much more sense. Done.

core/ir/instr_shared.c
edeiana: Done. I added the check in the CATEGORY_CHECK macro, but if you want me to separate the two and add more "spot-checks" for more instructions I can do that.

edeiana: Done.

core/ir/instr_api.h
derekbruening: We have documented that we're using version 14: https://dynamorio.org/page_code_style.html#autotoc_md122
Given that yours is older, yes please revert.  (If yours was newer we might consider updating.)

clients/drcachesim/tools/opcode_mix.cpp
```diff
@@ -225,10 +227,12 @@ opcode_mix_t::parallel_shard_memref(void *shard_data, const memref_t &memref)
             return false;
         }
         opcode = instr_get_opcode(&instr);
-        shard->worker->opcode_cache[trace_pc] = opcode;
+        category = instr_get_category(&instr);
+        shard->worker->opcode_category_cache[trace_pc] = opcode_category_data_t(opcode, category);
```
derekbruening, line 231: This line looks too long: shouldn't clang-format have split it at 90?

core/ir/instr_shared.c
derekbruening: Maybe add one instance of a multi-category instruction?

derekbruening (approved):

clients/drcachesim/tools/opcode_mix.cpp
edeiana: Yes, made me realize I had broken my vim config and clang-format wasn't running after saving. Fixed, thanks!

core/ir/instr_api.h
edeiana: Sorry, missed that. Fixed.

core/ir/instr_shared.c
edeiana: Done. Added one for "uncategorized" as well, since it's kind of an edge case being all 0s.

suite/tests/api/drdecode_x86.c
```diff
@@ -34,15 +34,17 @@
...
 #include <stdlib.h>
 #include <string.h>
 
 #define GD GLOBAL_DCONTEXT
 
-#define ASSERT(x)                                                                    \
-    ((void)((!(x)) ? (printf("ASSERT FAILURE: %s:%d: %s\n", __FILE__, __LINE__, #x), \
-                      abort(), 0)                                                    \
+#define ASSERT(x)                                                                        \
+    ((void)((!(x)) ? (fprintf(stdout, "ASSERT FAILURE: %s:%d: %s\n", __FILE__, __LINE__, \
```
derekbruening, line 45: stderr seems better for errors.

suite/tests/api/drdecode_x86.c
```diff
@@ -244,7 +271,7 @@ main()
 
     test_store_source();
 
-    printf("done\n");
+    fprintf(stdout, "done\n");
```
derekbruening, line 274: Could have remained as it was?  (Aside for future PR's on other tests: Note that tests that interleave output with DR's messages to stderr would want stderr for everything, and they use "print()" from tools.h but this standalone app test does not.)

derekbruening (approved):

suite/tests/api/drdecode_x86.c
edeiana: I agree, I just wanted to preserve the same behavior to avoid "breaking" anything.
The reason I changed it to `fprintf` is that I included `suite/tests/tools.h` to get access to `TESTANY` in `CHECK_CATEGORY`, but `tools.h` has https://github.com/DynamoRIO/dynamorio/blob/e4da1a7711c1f1dc1bb5757445d3e1097837177d/suite/tests/tools.h#L360 which prevents using `printf`.
I'll change it to stderr (similar to other `ASSERT` I've seen: https://github.com/DynamoRIO/dynamorio/blob/e4da1a7711c1f1dc1bb5757445d3e1097837177d/suite/tests/api/ir_x86.c#L71)

suite/tests/api/drdecode_x86.c
edeiana: Same reason of https://github.com/DynamoRIO/dynamorio/pull/6512#discussion_r1463943359 .
Should I change it to `print` now that `tools.h` is included? (Or was I not supposed to include `tools.h` here?)

suite/tests/api/drdecode_x86.c
derekbruening: Oh if tools.h is successfully included then use print()

suite/tests/api/drdecode_x86.c
derekbruening: If tools.h is successfully included then use print().  There is some build config complexity with tools.h vs config.h vs dr_api.h but if it worked then we're good.

suite/tests/api/drdecode_x86.c
edeiana: Done. Using `print()` also in this `ASSERT`.

suite/tests/api/drdecode_x86.c
edeiana: Done. The test is built correctly, and the print statements work, at least when I run the `code_api|api.drdecode` test.

derekbruening (approved):

edeiana:
Test opcode_categories fixed.
It was in the wrong place in suite/tests/CMakeLists.txt.
However, a few (minor) things changed since last approval, so I'd like a second pair of eyes before merging into master.

clients/drcachesim/tools/opcode_mix.cpp
```diff
@@ -252,7 +257,35 @@ opcode_mix_t::process_memref(const memref_t &memref)
...
+std::string
+get_category_names(uint category)
+{
+    std::string category_name;
+    if (category == DR_INSTR_CATEGORY_UNCATEGORIZED) {
+        category_name += instr_get_category_name(DR_INSTR_CATEGORY_UNCATEGORIZED);
+        return category_name;
+    }
+
+    uint max_mask = 0x80000000;
```
derekbruening, line 272: This could be a constant (either `const` or `#define`, and name in all-caps).

clients/drcachesim/tools/opcode_mix.h
```diff
@@ -82,8 +83,29 @@ class opcode_mix_t : public analysis_tool_t {
     parallel_shard_error(void *shard_data) override;
 
 protected:
+    struct opcode_category_data_t {
+        opcode_category_data_t()
+            : opcode(0)
```
derekbruening, line 88: s/0/OP_INVALID/

clients/drcachesim/tools/opcode_mix.h
```diff
@@ -82,8 +83,29 @@ class opcode_mix_t : public analysis_tool_t {
...
+            , category(0)
```
derekbruening, line 89: s/0/DR_INSTR_CATEGORY_UNCATEGORIZED/

clients/drcachesim/tools/opcode_mix.h
```diff
@@ -82,8 +83,29 @@ class opcode_mix_t : public analysis_tool_t {
...
+         * multiple category bits can be set when an instruction belongs to more
+         * than one category.  We assume 32 bits (i.e., 32 categories) is enough
+         * to be future-proof.
+         */
+        uint category;
+    };
+
     struct worker_data_t {
-        std::unordered_map<app_pc, int> opcode_cache;
+        std::unordered_map<app_pc, opcode_category_data_t> opcode_category_cache;
```
derekbruening, line 108: Optional: "opcode_data_cache" or "opcode_info_cache" sounds more like it includes several pieces of info and not just a category.  On that same note: "opcode_category_data_t" sounds like just the category and maybe "opcode_data_t" or "opcode_info_t" is more descriptive?  It's subjective though: if you like it how it is better please keep it.

core/ir/instr_shared.c
```diff
@@ -470,6 +470,25 @@ instr_get_category(instr_t *instr)
...
+{
+    switch (category) {
+    case DR_INSTR_CATEGORY_UNCATEGORIZED: return "uncategorized";
```
derekbruening, line 477: Add a (non-Doxygen) comment by the dr_instr_category_t enum as a reminder to update this function when adding new categories.

suite/tests/api/drdecode_x86.c
```diff
@@ -155,12 +157,20 @@ test_noalloc(void)
...
 }
 
-#define CHECK_CATEGORY(dcontext, instr, pc, category)                          \
-    ASSERT(instr_encode(dcontext, instr, pc) - pc < BUFFER_SIZE_ELEMENTS(pc)); \
-    instr_reset(dcontext, instr);                                              \
-    instr_set_operands_valid(instr, true);                                     \
-    ASSERT(decode(dcontext, pc, instr) != NULL);                               \
-    ASSERT(instr_get_category(instr) == category);                             \
+#define CHECK_CATEGORY(dcontext, instr, pc, categories, category_names)           \
+    ASSERT(instr_encode(dcontext, instr, pc) - pc < BUFFER_SIZE_ELEMENTS(pc));    \
```
derekbruening, line 161: Putting key steps needed for subsequent code inside something like ASSERT which is sometimes debug-only and turns into a nop in release build can be dangerous as the subsequent code will then fail.  Here it looks like ASSERT is defined in release build too.  But imagine this code being copied/modeled in some other file where that's not the case.  So it may be worth making safer code by separating the instr_encode and decode calls from the asserts.  Or add a comment stating the assumption that ASSERT is always enabled.

suite/tests/api/drdecode_x86.c
```diff
@@ -172,17 +182,36 @@ test_categories(void)
     /*  55 OP_mov_ld */
     instr = XINST_CREATE_load(GD, opnd_create_reg(DR_REG_XAX),
                               OPND_CREATE_MEMPTR(DR_REG_XAX, 42));
-    CHECK_CATEGORY(GD, instr, buf, DR_INSTR_CATEGORY_LOAD);
+    dr_instr_category_t categories_load[] = { DR_INSTR_CATEGORY_LOAD };
```
derekbruening, line 185: Can be `const`.  Ditto below.

derekbruening (approved):

clients/drcachesim/tools/opcode_mix.cpp
edeiana: Done.

clients/drcachesim/tools/opcode_mix.h
edeiana: Done.

clients/drcachesim/tools/opcode_mix.h
edeiana: I like the "data" variant. Done.

core/ir/instr_shared.c
edeiana: Done.

suite/tests/api/drdecode_x86.c
edeiana: Done.

suite/tests/api/drdecode_x86.c
edeiana: Done.

PR #6515, opened 2023-12-18
Test exit status of child in linux.clone test

xdje42:
In the linux.clone testcase, if the child crashes or exits with a non-zero exit code,
the test would otherwise get marked as passing.


suite/tests/linux/clone.c
```diff
@@ -351,10 +351,13 @@ delete_thread(pid_t pid, void *stack)
 {
```
derekbruening, line 351: > iX: Test exit status of child to make sure child completed successfully #6515

style: While we put iX in the branch name, we do not put it in the issue title: if there's no issue number we just don't have a prefix.  If you look at the history of commit titles you'll see examples with no issue.  A recent one: a7e64bfd7de003d38234c0ed77a067a10ee3f65c

derekbruening, line 351: > If the child crashes or exits with a non-zero exit code the test would otherwise get marked as passing.

More context would help: in particular, which test is this.  Someone looking at the commit history will have no idea what "the test" is, or in the title what part of the code this is referring to.  I would put the test name "linux.clone" in both the title (maybe "Text exit status of child in linux.clone test" as the title) and the description ("In the linux.clone test, ...").

xdje42: Just my luck looking for examples and finding a counter-example. :-)
Text updated.
Recommend updating the docs.

derekbruening (approved):

PR #6516, opened 2023-12-19
Remove "pre" tags from bug template

derekbruening:
Removes the `<pre>` tags which made the backticks visible when rendered (as in the issue template view at
https://github.com/DynamoRIO/dynamorio/issues/templates/edit) in favor of making the raw source edit less confusing.


xdje42 (approved):

derekbruening:
Failures are known flakes: win64 burst_replaceall #5412; x86-32 AMD #6417.

PR #6521, opened 2023-12-20
i#3044 AArch64 SVE codec: Add SVE2 TBL variant

philramsey-arm:
This patch adds the appropriate macros, tests and codec entries to decode and encode the following instruction:
```TBL    <Zd>.<Ts>, { <Zn1>.<Ts>, <Zn2>.<Ts> }, <Zm>.<Ts>```

Issue: #3044


AssadHashmi:
The x86 failures look like https://github.com/DynamoRIO/dynamorio/issues/6417.

AssadHashmi (approved):

PR #6522, opened 2023-12-20
Identify each test variation in linux.clone test output

xdje42:
The output is basically the same for each test variation, making it hard to correlate each test's output with the test's code.

### DIFF
--- a/suite/tests/linux/clone.c
+++ b/suite/tests/linux/clone.c
@@ -104,8 +104,8 @@ static void *stack;
 void
 test_thread(bool share_sighand, bool clone_vm, bool use_clone3)
 {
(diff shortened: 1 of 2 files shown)


abhinav92003 (approved):

suite/tests/linux/clone.c
```diff
@@ -104,6 +104,8 @@ static void *stack;
 void
```
abhinav92003, line 104: Re the title starting with `iX`: re-iterating Derek's comment: https://github.com/DynamoRIO/dynamorio/pull/6515#discussion_r1431775992.

xdje42: Ack

xdje42:
win64 failure in burst_traceopts #6423

Issue #6514, opened 2023-12-18
APP CRASH: Missing handling of stack ptr == NULL in clone(CLONE_VFORK + CLONE_VM)


xdje42:
**Describe the bug**

The Go process spawn support invokes clone() with CLONE_VFORK + CLONE_VM + stack_ptr == NULL.
Glibc doesn't support this but the syscall itself does.
stack_ptr == NULL means set $sp in the child to be $sp in the parent. This is dangerous, but the use of CLONE_VFORK makes this workable.

DR doesn't handle this causing the app to get a segv when it references values off of $sp (== NULL in the child).

**To Reproduce**

<pre>
```
$ cat >test-exec.go <<EOF
package main
import (
        "fmt"
        "os/exec"
)
func main() {
        cmd := exec.Command("/bin/echo", "xyz")
        out, err := cmd.CombinedOutput()
        fmt.Printf("%s\n", out)
        println(err)
        println("Done")
}
EOF
$ go build test-exec.go
$ drrun -- ./test-exec
```
</pre>
-->
signal 11 received during fork
fatal error: signal received during fork

Please also answer these questions:
 - What happens when you run without any client?
    Program runs to completion
 - What happens when you run with debug build ("-debug" flag to drrun/drconfig/drinject)?
    Same segv

**Expected behavior**

Program runs to completion.

**Screenshots or Pasted Text**

None

**Versions**

git master @6d1912f6f6655b9185aa6fc73243819c8e905cac
on amd64 linux

**Additional context**

None

xdje42:
Hmm. The repro example got deleted (it was a here document).

```
package main
import (
	"fmt"
	"os/exec"
)
func main() {
	cmd := exec.Command("/bin/echo", "xyz")
	out, err := cmd.CombinedOutput()
	fmt.Printf("%s\n", out)
	println(err)
	println("Done")
}
```

```
$ go build test-exec.go
$ drrun -- ./test-exec
```

derekbruening:
It wasn't deleted: hidden by the markdown.  I clicked "edit" and fixed the markdown in the original comment.

PR #6525, opened 2023-12-21
i#6514 clone: Fix app segv when passing null stack ptr

xdje42:
Recognize the special case of passing NULL for the stack pointer to clone(), clone3(). Passing NULL for the stack pointer is just a shorthand way of saying that that the child gets the parent's stack pointer value. This is what Go does for its fork/exec support.

Fixes: #6514

### DIFF
--- a/suite/tests/linux/clone.c
+++ b/suite/tests/linux/clone.c
@@ -78,10 +78,13 @@ clone(int (*fn)(void *arg), void *child_stack, int flags, void *arg, ...);
 
 /* forward declarations */
 static int
+make_clone_syscall(ptr_uint_t flags, void *stack,
+                   pid_t *parent_tid, pid_t *child_tid, void *tls,
+                   void (*fcn)(void));
+static int
 make_clone3_syscall(void *clone_args, ulong clone_args_size, void (*fcn)(void));
 static pid_t
-create_thread(int (*fcn)(void *), void *arg, void **stack, bool share_sighand,
-              bool clone_vm);
+create_thread(void (*fcn)(void), void **stack, bool share_sighand, bool clone_vm);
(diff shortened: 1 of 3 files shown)


suite/tests/linux/clone.c
```diff
@@ -185,15 +229,127 @@ run_with_exit(void)
...
+        "testq %%rax, %%rax\n\t"
+        "jnz clone_parent\n\t"
+        "xorq %%rbp, %%rbp\n\t"
+        "call *%%r15\n\t"
+        "clone_parent:\n\t"
+        "mov %%eax, %[result]\n\t"
+        : [result] "=r"(result)
+        : [syscall_num] "i"(SYS_clone), [flags] "rm"(flags), [stack] "rm"(stack),
+          [parent_tid] "rm"(parent_tid), [child_tid] "rm"(child_tid), [tls] "rm"(tls),
+          [fcn] "rm"(fcn)
```
derekbruening, line 273: We have dynamorio_syscall() (DR's asm syscall wrapper) in the drlibc library explicitly so that tests can use it; you can see several uses in suite/tests/ for examples.  Could we move dynamorio_clone() into drlibc so it can be used here to avoid needing any of this assembly and giving immediate support for all supported architectures?

xdje42: Cool, I wasn't aware of dynamorio_clone. It will need to be augmented to handle passing NULL for the stack ptr, and we'll need a clone3 variant, for each architecture.

Ergo, immediate support for all supported architectures isn't possible, but it's still a reasonable way to go.

xdje42:
This PR replaced with PRs #6557, #6594, #6601.

Issue #6524, opened 2023-12-21
libgcc/udivmoddi4.c uses different types from GCC


xdje42:
**Describe the bug**

In GCC, SI type is 32 bits and DI type is 64 bits (regardless of architecture).

dynamorio/third_party/libgcc/udivmoddi4.c has these typedefs:

```
typedef short Wtype;
typedef int DWtype;
typedef unsigned int UWtype;
typedef unsigned long long UDWtype;
#if __BYTE_ORDER__ != __ORDER_LITTLE_ENDIAN__
struct DWstruct {Wtype high, low;};
#else
struct DWstruct {Wtype low, high;};
#endif
typedef union {
  struct DWstruct s;
  DWtype ll;
} DWunion;
```

This is causing a compilation error with the Debian 13.2.0-4 compiler on amd64 linux.

```
FAILED: core/CMakeFiles/dynamorio.dir/__/third_party/libgcc/udivmoddi4.c.o
...
n function ‘__udivmoddi4’, 
    inlined from ‘__moddi3’ at /home/dje/memtrace/upstream/i6514/dynamorio/third_party/libgcc/udivmoddi4.c:117:10: 
/home/dje/memtrace/upstream/i6514/dynamorio/third_party/libgcc/udivmoddi4.c:99:9: error: array subscript ‘UDWtype {aka long long unsigned int}[\
0]’ is partly outside array bounds of ‘DWtype[1]’ {aka ‘int[1]’} [-Werror=array-bounds=]
   99 |     *rp = r;
      |     ~~~~^~~
/home/dje/memtrace/upstream/i6514/dynamorio/third_party/libgcc/udivmoddi4.c: In function ‘__moddi3’:
/home/dje/memtrace/upstream/i6514/dynamorio/third_party/libgcc/udivmoddi4.c:109:10: note: object ‘w’ of size 4
  109 |   DWtype w;
      |          ^
cc1: all warnings being treated as errors
```

**To Reproduce**

On amd64 linux:
```
$ mkdir build
$ cd build
$ CXXFLAGS=-m32 CFLAGS=-m32 cmake -GNinja -DBUILD_TESTS=ON ../dynamorio
$ ninja
```

**Versions**

Git tree at commit d2f47f3d8ea579e88d6245b00c4671bbd6d9bca7

**Suggested Fix**

[...]

PR #2605, opened 2017-08-09
i#2601: fix thread exit race on attach

derekbruening:
To handle a thread exiting on attach, adds a timeout to wait_event() and
its corresponding implementations: os_wait_event() on Windows and
ksynch_wait() on UNIX.  Uses the timeout to check whether a thread exited,
and if so, to move on.

Augments the test from #2600 to test this race.

Fixes #2061

PR #6526, opened 2023-12-21
i#6524: Fix udivmoddi4.c 32-bit linux build fail and semantics

xdje42:
For the targets we care about LIBGCC2_UNITS_PER_WORD == 4 (SImode in GCC parlance).
Thus a double word is 8 bytes (DImode in GCC parlance).

Previously, __moddi3 was using a mix of word and double word widths,
causing a compilation failure on 32 bit linux with the Debian 13.2.0-4 compiler.
The semantics were also incorrect: inputs and outputs should be 64 bits for the
hosts we care about (x86, aarch, 32 and 64 bit, and riscv64).

Fixes: #6524


third_party/libgcc/udivmoddi4.c
```diff
@@ -23,11 +23,15 @@ a copy of the GCC Runtime Library Exception along with this program;
 see the files COPYING3 and COPYING.RUNTIME respectively.  If not, see
```
abhinav92003, line 23: Following the example set by PR #2605, I believe we also need to update third_party/libgcc/README.dynamorio.

third_party/libgcc/udivmoddi4.c
```diff
@@ -23,11 +23,15 @@ a copy of the GCC Runtime Library Exception along with this program;
...
 
-/* This is extracted from gcc's libgcc/libgcc2.c with these typedefs added: */
-typedef short Wtype;
-typedef int DWtype;
-typedef unsigned int UWtype;
-typedef unsigned long long UDWtype;
+#include <stdint.h>
+
+/* This is extracted from gcc's libgcc/libgcc2.c with these typedefs added.
+   Note that for the targets we care about LIBGCC2_UNITS_PER_WORD == 4.
```
abhinav92003, line 29: Would be nicer for clarity to mention the targets, so that we know if/when this needs to be updated.

abhinav92003 (approved):

third_party/libgcc/udivmoddi4.c
xdje42: Done

third_party/libgcc/udivmoddi4.c
xdje42: Done

PR #6527, opened 2023-12-21
Reformat udivmoddi4.c with clang-format

xdje42:


xdje42:
Or is there a way to disable the format check for files in third_party (that aren't submodules) ?

derekbruening:
If you look at the original commit f1713ec4a9a856d1038c6095da67a5bd95b6a1c7 it excludes third_party as stated in the description but I don't remember if an exception for clang-format-diff was added (not seeing it at first glance) or the clang-format was just not applied.  But generally we don't want to reformat third_party code so maybe add an exception now if there isn't one: maybe a flag to the git diff run to produce the contents in runsuite.cmake can remove third_party.

derekbruening (changes requested):
We generally don't want to reformat third_party.  We would either disable the clang-format check manually for the review and to confirm tests are green and then re-enable right before merge, or you could try to add a permanent third_party exclusion (maybe flag to the git diff in runsuite.cmake).

PR #6528, opened 2023-12-22
i#3544 RV64: Treat atomic instructions as load/store instructions

ksco:
Atomic instructions from the A extension like LR/SC and AMO* access memory so they should be treated as load/store. In addition, AMO* instructions perform read-modify-write operations, they should have both a source and a destination operands to comply with DR's IR convention. We make the destination an implicit operand as they are identical to the source.

Related PR: #6506
Issue: #3544


suite/tests/api/ir_riscv64.c
```diff
@@ -306,97 +306,120 @@ test_atomic(void *dc)
...
+        dc, opnd_create_reg(DR_REG_A0),
+        opnd_create_base_disp(DR_REG_A1, DR_REG_NULL, 0, 0, OPSZ_4),
+        opnd_create_reg(DR_REG_A2), opnd_create_immed_int(0b01, OPSZ_2b));
     test_instr_encoding(dc, OP_amoxor_w, instr);
-    instr = INSTR_CREATE_amoxor_d(dc, opnd_create_reg(DR_REG_A0),
-                                  opnd_create_reg(DR_REG_A1), opnd_create_reg(DR_REG_A2),
-                                  opnd_create_immed_int(0b01, OPSZ_2b));
+    instr = INSTR_CREATE_amoxor_d(
+        dc, opnd_create_reg(DR_REG_A0),
+        opnd_create_base_disp(DR_REG_A1, DR_REG_NULL, 0, 0, OPSZ_8),
```
abhinav92003, line 361: Are usages with non-zero index or displacement supported? Add tests if yes.

abhinav92003 (approved):

suite/tests/api/ir_riscv64.c
ksco: No, they're not supported, amo instructions only have a base register.

PR #6532, opened 2024-01-02
i#6530: Support alternate libc name in drsyms-test

### LINKED ISSUE #6530: client.dryms-test fails when libc path has no dash
The client.dryms-test fails when the libc path has no dash.  It looks for "/libc-" but in my case there's no dash as the path is /usr/lib/x86_64-linux-gnu/libc.so.6.  The failure is a missing line about glibc in the output, the middle line here from this output snippet of the template:

```
finished unmangling.                                                                                                                                                                
(found|couldn.t find) glibc malloc (and|or) __GI___libc_malloc.                                                                                                                     
enumerating with DRSYM_LEAVE_MANGLED                                                                                                                                                
```

derekbruening:
The drsyms-test was looking for "/libc-" to find the libc library, which failes to find "/usr/lib/x86_64-linux-gnu/libc.so.6".  We update the search here to allow a dot.

Tested locally where the test failed before.

Fixes #6530

### DIFF
--- a/suite/tests/client-interface/drsyms-test.dll.cpp
+++ b/suite/tests/client-interface/drsyms-test.dll.cpp
@@ -580,7 +580,7 @@ lookup_dll_syms(void *dc, const module_data_t *dll_data, bool loaded)
     dll_base = dll_data->start;
 
 #ifdef UNIX
-    if (strstr(dll_path, "/libc-")) {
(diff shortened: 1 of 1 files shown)


derekbruening:
Failures are AMD x32 #6417 and attach_test flake #6452

abhinav92003 (approved):

Issue #5926, opened 2023-03-23
tool.drcov.fib failing in local runs due to lack of DWARF5 support in libelftc used by drsyms


abhinav92003:
Xref the user report on dynamorio-users: https://groups.google.com/g/dynamorio-users/c/mDwTAPFek40, where following https://dynamorio.org/page_drcov.html for a simple app generates an empty file.

I was able to reproduce the user-reported issue on my workstation. @derekbruening found that `tool.drcov.fib` is also failing. It's passing in our CI so maybe this is the result of some recent lib update.

abhinav92003:
I tried updating the version in `rebuild_elftc.sh` to the latest but some changes in https://sourceforge.net/p/elftoolchain/code/3898 are breaking the build. Also, r3897 doesn't fix the test failure.

abhinav92003:
I was able to get around it by ignoring warnings (at least for now to test if the latest libelftc version fixes it), but I'm hitting other issues with building libelftc.

```
In file included from elf.c:31:
_libelf.h:36:10: fatal error: _elftc.h: No such file or directory
   36 | #include "_elftc.h"
      |          ^~~~~~~~~~
compilation terminated.
*** Error code 1

```

abhinav92003:
I tried to run rebuild_elftc.sh on a Ubuntu 20.04 VM (since that is what the build instructions at https://sourceforge.net/p/elftoolchain/code/HEAD/tree/trunk/INSTALL use), but the script still fails with the same issue. Next I thought maybe rebuild_elftc.sh itself needs to be updated, and tried building libelftc myself. I could do successfully that on the Ubuntu 20.04 VM (but not on my own workstation), and replaced the generated .a and header files and re-ran the test, but it still failed.

abhinav92003:
I was able to use rebuild_elftc.sh to get new 64-bit libraries and headers from the latest libelftc (32-bit still not building successfully). It needed some extra steps: either installing libelftc or at least copying the elfdefinitions.h header file to /use/include/sys. But I'm still getting the same error in the tool.drcov.fib test: `drsym_lookup_address` returning `DRSYM_ERROR_LINE_NOT_AVAILABLE`. Maybe the issue is in drsym; looking.

abhinav92003:
[...]

PR #6533, opened 2024-01-02
i#5926 libdw: Add new interface dr_running_under_dynamorio()

derekbruening:
Adds a new API routine dr_running_under_dynamorio() which is similar to the application interface dr_app_running_under_dynamorio() but is callable from a client.  The use case is code meant for us both in managed mode under DR and in standalone library mode.  Sometimes the behavior for the two differs significantly, such as whether certain library routines are safe when linked statically.

Adds sanity checks to some existing tests.

The driving use case is in drsyms's support for libdw.  The actual use of this will come in that libdw commit.

Issue: #5926

### DIFF
--- a/core/lib/dr_tools.h
+++ b/core/lib/dr_tools.h
@@ -1,5 +1,5 @@
 /* **********************************************************
- * Copyright (c) 2010-2023, Inc.  All rights reserved.
+ * Copyright (c) 2010-2024, Inc.  All rights reserved.
  * Copyright (c) 2002-2010 VMware, Inc.  All rights reserved.
  * **********************************************************/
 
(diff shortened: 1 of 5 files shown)


derekbruening:
Failures are amd 32-bit #6417 and ub22 api.rseq failed to synch #6185

core/lib/dr_tools.h
```diff
@@ -103,6 +103,15 @@ DR_API
 bool
 dr_using_all_private_caches(void);
 
+DR_API
+/**
+ * Returns false if DynamoRIO is being used as a regular standalone library.
+ * Returns true if DynamoRIO is controlling the application by running
+ * its code through a sofwtare code cache.
```
abhinav92003, line 110: s/sofwtare/software/

core/lib/dr_tools.h
```diff
@@ -103,6 +103,15 @@ DR_API
 bool
 dr_using_all_private_caches(void);
 
+DR_API
+/**
+ * Returns false if DynamoRIO is being used as a regular standalone library.
```
abhinav92003, line 108: Examples would be nice (drdecode etc).

abhinav92003 (approved):

core/lib/dr_tools.h
derekbruening: Done.

core/lib/dr_tools.h
derekbruening: Added a link to page_standalone in the docs (which talks about drdecode) and to dr_standalone_init().

derekbruening:
win64 failure is burst_traceopts flake #6423

Issue #6499, opened 2023-12-08
ASSERT: Invariant test failure: Too many reads (AArch64+sve)


joshua-warburton:
**Describe the bug**
When running various applications, including fleetbench, we run into the following assert:
```
Trace invariant failure in T1909 at ref # 19484 (957 instrs since timestamp 13345924302819833): Too many read records
```
This failing test compares a predicted number of reads with the actual number of reads recorded by dynamorio.

The above case is failing at this point, when the second read is considered:
```
       19482       15318:        1909 ifetch       4 byte(s) @ 0x0000000000405b3c a54a4681   ld1w   (%x20,%x10,lsl #2) %p1/z -> %z1.s
       19483       15318:        1909 read         0 byte(s) @ 0x0000000000954e80 by PC 0x0000000000405b3c
       19484       15318:        1909 read         0 byte(s) @ 0x0000000000954e84 by PC 0x0000000000405b3c
       19485       15318:        1909 read         0 byte(s) @ 0x0000000000954e88 by PC 0x0000000000405b3c
       19486       15318:        1909 read         0 byte(s) @ 0x0000000000954e8c by PC 0x0000000000405b3c
       19487       15318:        1909 read         0 byte(s) @ 0x0000000000954e90 by PC 0x0000000000405b3c
       19488       15318:        1909 read         0 byte(s) @ 0x0000000000954e94 by PC 0x0000000000405b3c
       19489       15318:        1909 read         0 byte(s) @ 0x0000000000954e98 by PC 0x0000000000405b3c
       19490       15318:        1909 read         0 byte(s) @ 0x0000000000954e9c by PC 0x0000000000405b3c
       19491       15319:        1909 ifetch       4 byte(s) @ 0x0000000000405b4 
```
As this is run on a machine with a 256 bit vector length, we see 8 reads as expected for this instruction.

The invariant checker predicts the reads using the following function
```
uint
instr_num_memory_read_access(instr_t *instr)
{
    int i;
    opnd_t curop;
    int count = 0;
    const int opc = instr_get_opcode(instr);
    if (opc_is_not_a_real_memory_load(opc))
        return 0;
    for (i = 0; i < instr_num_srcs(instr); i++) {
        curop = instr_get_src(instr, i);
        if (opnd_is_memory_reference(curop)) {
            ++count;
        }
    }
    return count;
}
```

This means that we can only ever expect one read per memory accessing operand (effectively one per instruction). 

[...]

Issue #6510, opened 2023-12-16
Mark x86 SIMD operations with masks as predicated


derekbruening:
Xref #6499 where a64 scatter/gather are going to marked as predicated.  This issue covers doing the same thing for x86.  This should include all SIMD operations with masks.  We may need a new predicate type?

derekbruening:
Xref AArch64 adding predicates PR #6535

PR #6535, opened 2024-01-02
i#6499: AArch64 codec: mark instructions with governing predicate as predicated

joshua-warburton:
This patch adds the DR_PRED_MASKED flag to instructions containing any governing predicate registers as well as tagging the governing register with DR_OPND_IS_MASKED.

To do this, it extends the codec functionality such that operands may have "flags" which allow us to add some special code generation when they are encountered. In this case we amend the instr decode/encode functions to both tag the governing predicate and the instr itself.

issue: #6499


core/ir/opnd_api.h
```diff
@@ -2569,22 +2574,28 @@ opnd_is_element_vector_reg(opnd_t opnd);
...
 INSTR_INLINE
-/** Returns true iff \p opnd is a predicate register. */
+/** Returns true iff \p opnd is an SVE predicate register. */
 bool
 opnd_is_predicate_reg(opnd_t opnd);
 
 DR_API
 INSTR_INLINE
-/** Returns true iff \p opnd is a merging predicate register. */
+/** Returns true iff \p opnd is a n SVEmerging predicate register. */
```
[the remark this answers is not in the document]
joshua-warburton: Done

core/ir/instr.h
```diff
@@ -516,6 +516,20 @@ instr_t *
 instr_set_translation_mangling_epilogue(dcontext_t *dcontext, instrlist_t *ilist,
                                         instr_t *instr);
 
+#ifdef AARCH64
```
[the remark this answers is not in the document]
joshua-warburton: Done, surprised clang format didn't pick this up

core/ir/aarch64/instr_create_api.h
```diff
@@ -8816,8 +8834,10 @@
  *
```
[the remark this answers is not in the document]
joshua-warburton: done

core/ir/instr_api.h
```diff
@@ -139,6 +139,9 @@ typedef enum _dr_pred_type_t {
     DR_PRED_AL, /**< ARM condition: 1110 Always (unconditional)                    */
 #    ifdef AARCH64
     DR_PRED_NV, /**< ARM condition: 1111 Never, meaning always                     */
+    DR_PRED_GOVERNING, /** Used for AArch64 SVE instructions with governing predicate
```
derekbruening, line 142: Since the x86 mask concept seems to be the same (right? a bit per load/store element to enable/disable it?), let's try to share the interface as much as possible to make it easier on clients.  So SVE calls it a "governing predicate register" while AVX-512 calls them "opmask registers".  Is there some generic term we could use for the predicate so we can have the same DR_PRED_ name for both architectures so clients don't have to have ifdefs for the same functional code?  "DR_PRED_MASKED" seems clearer to me but I'm not sure if that's just b/c of the x86 description: does "masked" makes sense on the aarch64 side?  Even if we decided "DR_PRED_GOVERNING" was the right cross-arch name we'd want to update the dox comment to make it clear this is a cross-arch predicate concept.

core/ir/aarch64/disassemble.c
```diff
@@ -55,6 +55,7 @@ static const char *const pred_names[] = {
     "le", /* DR_PRED_LE */
     "al", /* DR_PRED_AL */
     "nv", /* DR_PRED_NV */
+    "",   /* DR_PRED_GOVERNING */
```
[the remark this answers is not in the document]
AssadHashmi: I was thinking of an operand modifier string rather than for the opcode.
I guess that's more work than we need to do just now.

core/ir/instr_inline_api.h
```diff
@@ -175,6 +175,13 @@ opnd_is_predicate_zero(opnd_t op)
     return opnd_is_predicate_reg(op) && ((op.aux.flags & DR_OPND_IS_ZERO_PREDICATE) != 0);
 }
 
+INSTR_INLINE
+bool
+opnd_is_governing(opnd_t op)
+{
+    return opnd_is_predicate_reg(op) && ((op.aux.flags & DR_OPND_IS_GOVERNING) != 0);
```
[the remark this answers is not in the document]
joshua-warburton: Because these can be used for clients, I believe TEST is undefined. The other test macros in this file all appear to be behind a ifded DYNAMORIO_INTERNAL

core/ir/instr_api.h
AssadHashmi: > "DR_PRED_MASKED" seems clearer to me but I'm not sure if that's just b/c of the x86 description: does "masked" makes sense on the aarch64 side?

Short answer is that "masked" is too limited a name for AArch64 SVE predicate register behaviour.

Longer answer:
SVE has 3 flavours of governing predicate registers, two mask-like active/inactive and one control/selector (e.g. select which elements from source vector are acted upon by op):
```
<Pg>/Z as in e.g. BICS <Pd>.B, <Pg>/Z, <Pn>.B, <Pm>.B
<Pg>/M as in e.g. ABS <Zd>.<T>, <Pg>/M, <Zn>.<T>
<Pg> as in e.g. SPLICE <Zdn>.<T>, <Pg>, <Zdn>.<T>, <Zm>.<T>
```
Because of this, we went for `DR_PRED_GOVERNING` rather than `DR_PRED_MASKED`.

AssadHashmi: @derekbruening are you happy with using `DR_PRED_GOVERNING` and updates to Doxygen comments to clarify the cross-architectural difference between the predicates?

Alternatively, we could stick to `DR_PRED_MASKED` if it keeps the code simpler, and highlight the difference in meaning between x86 and aarch64 in Doxygen comments.

derekbruening: `DR_PRED_GOVERNING` can work: I don't think the term "governing" is used anywhere in the x86 ISA so it is just a generic term.  (Note that for me "masked" could cover the "control/selector for which elements are acted upon" -- "masked" also seems a general term, and "which are acted upon" sounds very similar to "active/inactive".)

AssadHashmi: We'll stick with `DR_PRED_MASKED` to keep naming consistent and simple.

AssadHashmi:
The x86 failures look like https://github.com/DynamoRIO/dynamorio/issues/6417.

AssadHashmi (approved):

core/ir/instr_api.h
joshua-warburton: Latest patch changes all the DR_PRED_GOVERNING to DR_PRED_MASKED. This is currently still only defined for AArch64 as I assume the work to add it to x86 falls under #6510

derekbruening:
This PR broke our build with a python error: #6580

PR #6539, opened 2024-01-03
i#6538: Add lz4 to drmemtrace_proj

### LINKED ISSUE #6538: drmemtrace_proj fails to link if lz4 is in drmemtrace
The drmemtrace_proj test fails with several link errors like this on my machine:
```
/usr/local/google/home/bruening/dr/git/src/clients/drcachesim/common/lz4_istream.h:81:(.text._ZN9dynamorio10drmemtrace16lz4_istreambuf_tD2Ev[_ZN9dynamorio10drmemtrace16lz4_istreambuf_tD5Ev]+0x7d): undefined reference to `LZ4F_freeDecompressionContext'
```
The cmake code looks like it just needs a check for lz4 and a link line added.

derekbruening:
Adds linking of lz4 into the drmemtrace_proj test executable, if present.

Tested on a platform where the drmemtrace_proj test was failing.

Fixes #6538

### DIFF
--- a/clients/drcachesim/tests/CMakeLists.txt
+++ b/clients/drcachesim/tests/CMakeLists.txt
@@ -1,5 +1,5 @@
 # **********************************************************
-# Copyright (c) 2015-2017 Google, Inc.    All rights reserved.
+# Copyright (c) 2015-2024 Google, Inc.    All rights reserved.
 # **********************************************************
 
 # Redistribution and use in source and binary forms, with or without
@@ -45,6 +45,8 @@ if (UNIX) # The shipped drmemtrace on Windows has no zlib support.
     message(WARNING "zlib not found: linking will fail if drmemtrace has compressed "
       "file support built-in")
   endif()
+  # Similarly, we need lz4 if drmemtrace has it.
+  find_library(liblz4 lz4)
 endif ()
 
 find_package(DynamoRIO)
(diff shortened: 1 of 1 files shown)


clients/drcachesim/tests/CMakeLists.txt
```diff
@@ -62,3 +64,6 @@ target_link_libraries(analyzer_separate drmemtrace_analyzer drmemtrace_histogram
 if (ZLIB_FOUND)
   target_link_libraries(analyzer_separate ${ZLIB_LIBRARIES})
 endif ()
+if (liblz4)
+  target_link_libraries(analyzer_separate lz4)
```
abhinav92003, line 68: I was initially confused about how analyzer_separate is related to drmemtrace_proj, then I saw the comment in clients/drcachesim/CMakelists.txt:

```
# We have a companion test built using a separate --build-and-test CMake project in
# tests/analyzer_separate.cpp to better test 3rd-party usage.
```
I suppose drmemtrace_proj is the test being referred to above right?

Maybe add more details to the PR description as to how they're connected.

abhinav92003 (approved):

derekbruening:
Wow it turns out a clean build fixed the issue: the DynamoRIOTarget64.cmake is not updated properly on config changes I guess.  I'll turn this PR into removing the zlib link, which shouldn't be needed.

clients/drcachesim/tests/CMakeLists.txt
derekbruening: Done, and updated title and changed the code completely as it turns out we need just the search and not the link.

abhinav92003 (approved):

derekbruening:
ub22 is api.rseq #6185

PR #6540, opened 2024-01-04
i#3544 RV64: Enable reuse_distance test

ksco:
tool.reuse_distance test is enabled by performing the following changes:

1. Set -max_bb_instrs to 15 by default on RISC-V
2. Use `c.ldsp/c.sdsp/c.fldsp/c.fsdsp` to reduce the size of context switches for clean calls
3. Swap sc.w/d rd and mem operand positions so that mem becomes the first operand to be consistent with AArch64
4. Implement `drutil_insert_get_mem_addr()`

Related PR: #6506
Issue: #3544


ksco:
@derekbruening Hi, do you know what caused this error: [<ERROR: using undefined symbol!>](https://github.com/DynamoRIO/dynamorio/actions/runs/7413743441/job/20173349282#step:6:2635) in CI? It's the newly enabled test `tool.reuse_distance`, this test passed on my local machine, both on QEMU and real hardware, so I'm not sure what caused the difference in the CI.

derekbruening:
> @derekbruening Hi, do you know what caused this error: [<ERROR: using undefined symbol!>](https://github.com/DynamoRIO/dynamorio/actions/runs/7413743441/job/20173349282#step:6:2635) in CI? It's the newly enabled test `tool.reuse_distance`, this test passed on my local machine, both on QEMU and real hardware, so I'm not sure what caused the difference in the CI.

This means the private loader failed to find the source for an import in a client or its dep libs and the import was actually invoked.  I thought there was supposed to be an earlier `WARNING! symbol lookup error: <lib> undefined symbol <name` which gave the actual name of the symbol: is it because these tests are passing `-stderr_mask 0xc` which hides warnings?  See `git grep -4 UNDEFINED_SYMBOL`.  Try using tmate to log in to the GA CI runner VM to run the test command manually, or temporarily tweak to `-stderr_mask 0xf` to see the symbol name.

clients/drcachesim/CMakeLists.txt
```diff
@@ -512,6 +512,9 @@ macro(add_drmemtrace name type)
   if (liblz4)
     target_link_libraries(${name} lz4)
   endif ()
+  if (RISCV64)
+    target_link_libraries(${name} atomic)
+  endif ()
```
ksco, line 517: @derekbruening I did some debugging as suggested and found that the missing symbol is `__atomic_is_lock_free`. This change makes the `reuse_distance` pass, but I'm not sure if it's a proper way to do it.

core/ir/riscv64/codec.py
```diff
@@ -872,6 +872,9 @@ def __fixup_uncompressed_inst(self, inst: Instruction):
             elif rs3 == 0x3: # SC.W/D instructions
                 dbg(f'fixup: {inst.name} {[f.name for f in inst.flds]}')
                 inst.flds[2] = Field.V_S_RS1_DISP
+                # Swap the rd and mem operand positions so that mem becomes the
+                # first operand to be consistent with AArch64.
+                inst.flds[2], inst.flds[3] = inst.flds[3], inst.flds[2]
```
derekbruening, line 877: Are there other load/store instructions that need this?

Should this be done in the table/database source, instead of here?  Or that order is the bit position order?

ext/drutil/drutil.c
```diff
@@ -511,8 +505,10 @@ drutil_insert_get_mem_addr_arm(void *drcontext, instrlist_t *bb, instr_t *where,
                                           opnd_create_reg(base), opnd_create_reg(index),
                                           OPND_CREATE_INT(extend),
                                           OPND_CREATE_INT(amount));
-#    endif /* ARM/AARCH64 */
             PRE(bb, where, instr);
+#    else  /* RISCV64 */
+            ASSERT(false, "unreachable");
```
derekbruening, line 510: Maybe add in the string sthg about there not being any base reg + index reg addressing mode in RISC-V?

derekbruening (approved):

core/ir/riscv64/codec.py
ksco: > Are there other load/store instructions that need this?

This is the only instruction that has both a register destination and a memory destination. Well, it's also true for "read-modify-write" atomic instructions like `amoswap`, but their memory destination is implicit, so it's not the same case.

> Should this be done in the table/database source, instead of here? Or that order is the bit position order?

Yes, the operand order in the txt files is the bit position order, but it's not used by our codegen, so I think both ways are acceptable, I'll keep it this way.

clients/drcachesim/CMakeLists.txt
ksco: Yeah, it is weird. After searching, I found [this page](https://blog.jiejiss.com/A-RISC-V-gcc-pitfall-revealed-by-a-glibc-update/) states that: *gcc does not support inlining subword atomics, and this is the reason why -latomic must be presented when invoking gcc.* Maybe that's the reason for this? not sure.

Issue #6541, opened 2024-01-05
ASSERT synch.c:261 on client.destructor test on ubuntu22 32-bit


derekbruening:
I tried to reproduce the synch.c:261 assert on the drcachesim tests on GA
CI ubuntu22 32-bit seen in PR #6534 on a local machine and failed.  But on
running the whole suite there I did hit the assert in the client.destructor
test.

Running under gdb, first we have SIGSEGV in privlib constructor:
```
Program received signal SIGSEGV, Segmentation fault.
0xf724fbed in ?? ()
(gdb) x/3i 0xf724fbed 
   0xf724fbed:	mov    %gs:0x4fe,%al
   0xf724fbf3:	movzbl %al,%eax
(gdb) bt
#0  0xf724fbed in ?? ()
#1  0xf72d649d in write ()
#2  0xf733dd65 in _IO_file_write ()
#3  0xf733d5a6 in ?? ()
#4  0xf733df15 in _IO_file_xsputn ()
#5  0xf723f904 in fwrite ()
#6  0xf78c5b49 in __gnu_cxx::stdio_sync_filebuf<char, std::char_traits<char> >::xsputn(char const*, int) ()
#7  0xf78f2df2 in std::basic_ostream<char, std::char_traits<char> >& std::__ostream_insert<char, std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&, char const*, int) ()
#8  0xf78f3192 in std::basic_ostream<char, std::char_traits<char> >& std::operator<< <std::char_traits<char> >(std::basic_ostream<char, std::char_traits<char> >&, char const*) ()
#9  0xf7fba274 in hasdtr::hasdtr (this=0xf7fbd00f) at /home/derek/dr/git/src/suite/tests/client-interface/destructor.dll.cpp:46
#10 0xf7fba1ff in __static_initialization_and_destruction_0 (__initialize_p=1, __priority=65535) at /home/derek/dr/git/src/suite/tests/client-interface/destructor.dll.cpp:54
#11 0xf7fba247 in _GLOBAL__sub_I_destructor.dll.cpp(void) () at /home/derek/dr/git/src/suite/tests/client-interface/destructor.dll.cpp:60
#12 0xf7c8370a in privload_call_lib_func (func=0xf7fba228 <_GLOBAL__sub_I_destructor.dll.cpp(void)>) at /home/derek/dr/git/src/core/unix/loader.c:1079
[...]

PR #6542, opened 2024-01-05
i#6541 privlib: Add try/except around private library init/fini

derekbruening:
Wraps private library init/fini calls in a try/except on Linux (as is already done on Windows) and reports a crash as a SYSLOG_ERROR pointing at the library's name.  This is much better than a weird assert or mysterious secondary crash later.

Adds a check for shared_data==NULL in check_thread_in_last_vmwarea() to avoid a crash seen there.

Issue: #6541


derekbruening:
x64 failure is attach-test #6452

abhinav92003 (approved):

Issue #6218, opened 2023-07-18
Should upper bit zeroing of overlaid registers be modeled in DR's IR?


derekbruening:
We were revisiting this decision to not model the upper bit zeroing in the IR as part of a users list discussion: https://groups.google.com/g/dynamorio-users/c/jzTHyiY4ZsA/m/l6VPPVVsAQAJ

The original issue was #1382 about sub-SIMD register sizes; seemed worth filing a fresh issue for this top bit zeroing revisti.

We may resolve that discussion by speeding up instr_zeroes_ymmh(); adding top bit behavior to the IR may remain an open topic.  This includes not just upper SIMD bit zeroing but also GPR zeroing (write to EAX and it zeroes the top bits of RAX).

Our thoughts from a brief discussion:

+ Change register operand or add 2nd dest operand? ⇒ No
+ Add flag to operand or to instr_t? ⇒ instr_t b/c operational; do not expect 2 dests w/ diff behavior
  - Set by decode; user sets for synthetic & encoder picks encoding
+ Only for SIMD or set flag for x86 GPRs?  Make zeroing the default and then no flag?  What about single-vector-element ops which preserve: all have to set flag (today using size to indicate sub-reg?)?
+ Aside: vzerupper should be in instr_zeroes_ymmh() too? => #6217
  - Should vzeroall + vzeroupper have explicit operands?

Are there cases of the same opcode+opnds which can be encoded legacy or VEX?
If so, which one is picked?
Will it maintain re-encoding fidelity (xref #3339)?
=>
I believe all the 2-variant cases where the same operation can be encoded as legacy or vex actually have a different opcode for vex with a "v" prefix.  E.g., OP_padd vs OP_vpadd.  They do the same operation but have a distinct opcode.

PR #6544, opened 2024-01-08
i#5365 AArch64: Fix 0 size read/write records in drmemtrace

jackgallagher-arm:
When debugging i#6499 we noticed that drcachesim was producing 0 byte read/write records for some SVE load/store instructions:

```
 ifetch       4 byte(s) @ 0x0000000000405b3c a54a4681   ld1w   (%x20,%x10,lsl #2) %p1/z -> %z1.s
 read         0 byte(s) @ 0x0000000000954e80 by PC 0x0000000000405b3c
 read         0 byte(s) @ 0x0000000000954e84 by PC 0x0000000000405b3c
 read         0 byte(s) @ 0x0000000000954e88 by PC 0x0000000000405b3c
 read         0 byte(s) @ 0x0000000000954e8c by PC 0x0000000000405b3c
 read         0 byte(s) @ 0x0000000000954e90 by PC 0x0000000000405b3c
 read         0 byte(s) @ 0x0000000000954e94 by PC 0x0000000000405b3c
 read         0 byte(s) @ 0x0000000000954e98 by PC 0x0000000000405b3c
 read         0 byte(s) @ 0x0000000000954e9c by PC 0x0000000000405b3c
 ifetch       4 byte(s) @ 0x0000000000405b4
```

This turned out to be due to drdecode being linked into drcachesim twice: once into the drcachesim executable, once into libdynamorio. drdecode uses a global variable to store the SVE vector length to use when decoding so we end up with two copies of that variable and only one was being initialized.
To fix this properly we would need to refactor the libraries so that there is only one copy of the sve_veclen global variable, or change the way that the decoder gets the vector length so its no longer stored in a global variable. In the mean time we have a workaround which makes sure both copies of the variable get initialized and drcachesim produces correct results.

With that workaround in place however, the results were still wrong. For expanded scatter/gather instructions when you are using an offline trace, raw2trace doesn't have access to the load/store instructions from the expansion, only the original app scatter/gather instruction. It has to create the read/write records using only information from the original scatter/gather instruction and it uses the size of the memory operand to determine the size of each read/write. This works for x86 because the x86 IR uses the per-element data size as for the memory operand of scatter/gather instructions. This doesn't work for AArch64 because the AArch64 codec uses the maximum data transferred (per-element data size * number of elements) like other SIMD load/store instructions.

We plan to make the AArch64 IR consistent with x86 by changing it to use the same convention as x86 for scatter/gather instructions but in the mean time we can work around the inconsistency by fixing the size in raw2trace based on the instruction's opcode.

Issues: #6499, #5365, #5036


jackgallagher-arm:
Hi @derekbruening @abhinav92003, the problems I found debugging the 0-byte read/write issue we spotted in #6499 raise a few questions that I would value your input on:

1. Can you confirm how the memory operand size works for x86 scatter/gather instructions? AArch64 currently uses the maximum number of bytes transferred by the instruction (the amount of data that would be read/written if all elements are active, i.e per-element-size * number-of-elements) but raw2trace appears to expect it to be expecting it to be the per-element size. We want to make sure that the AArch64 IR is consistent with the x86 IR so I want to make sure I understand how it works in x86.
2. The drcachesim process ends up with two copies of `drdecode.a`: one linked in to `drcachesim`, one linked in to `libdynamorio.so`. Is this intended?
3. This patch sets the vector length used when decoding instructions in an offline raw trace to the VL for the current process. Most of the time if you are processing the trace on the same machine the trace was produced on this will be correct, but if you are using a different machine with a different VL (or you are using the same machine but have configured a different VL), it won't be correct. raw2trace really needs to be configuring the decoder to use the VL of the traced process, not the current process. Is there a way we can record the VL in the raw trace so that we can read it back later in raw2trace and call `dr_set_sve_vector_length()` to configure it correctly?

abhinav92003:
> We want to make sure that the AArch64 IR is consistent with the x86 IR so I want to make sure I understand how it works in x86.

On x86, the IR has a fixed operand size for each scatter-gather opcode, which is equal to the per-element size. E.g. [2], [3] where for `vpgatherdd` the operand size is set to 4 bytes.

> Is there a way we can record the VL in the raw trace so that we can read it back later in raw2trace and call dr_set_sve_vector_length() to configure it correctly?

We can perhaps add a new marker[4] for vector size in the thread raw trace header [5], similar to page and cache sizes.

Question:
Can an application use multiple vector lengths throughout its execution by setting ZCR_Elx.LEN[1] to different values? Or is the vector length guaranteed to be fixed throughout for an app? 


[1]: https://developer.arm.com/documentation/102476/0100/SVE-architecture-fundamentals/Configurable-vector-length
[2]: https://github.com/DynamoRIO/dynamorio/blob/960d1d7f45b4772cb94f748d0218a9091cbee3bd/core/ir/x86/decode_table.c#L7829
[3]: https://github.com/DynamoRIO/dynamorio/blob/960d1d7f45b4772cb94f748d0218a9091cbee3bd/core/ir/x86/decode_table.c#L1865
[4]: https://github.com/DynamoRIO/dynamorio/blob/960d1d7f45b4772cb94f748d0218a9091cbee3bd/clients/drcachesim/common/trace_entry.h#L620
[5]: https://github.com/DynamoRIO/dynamorio/blob/960d1d7f45b4772cb94f748d0218a9091cbee3bd/clients/drcachesim/tracer/instru_offline.cpp#L385

derekbruening:
> 2. The drcachesim process ends up with two copies of drdecode.a: one linked in to drcachesim, one linked in to libdynamorio.so. Is this intended?

No.  We have put effort into supporting the same libdynamorio used for both standalone decoding and managed mode (at different time of course) in the same process, so the separate copy should not be needed and it is surprising to see it.  Is it easy to remove it?

jackgallagher-arm:
> On x86, the IR has a fixed operand size for each scatter-gather opcode, which is equal to the per-element size. E.g. [2](https://github.com/DynamoRIO/dynamorio/blob/960d1d7f45b4772cb94f748d0218a9091cbee3bd/core/ir/x86/decode_table.c#L7829), [3](https://github.com/DynamoRIO/dynamorio/blob/960d1d7f45b4772cb94f748d0218a9091cbee3bd/core/ir/x86/decode_table.c#L1865) where for `vpgatherdd` the operand size is set to 4 bytes.

Thanks. We will adjust the AArch64 IR so SVE scatter/gather instructions use the same convention.

> We can perhaps add a new marker[4](https://github.com/DynamoRIO/dynamorio/blob/960d1d7f45b4772cb94f748d0218a9091cbee3bd/clients/drcachesim/common/trace_entry.h#L620) for vector size in the thread raw trace header [5](https://github.com/DynamoRIO/dynamorio/blob/960d1d7f45b4772cb94f748d0218a9091cbee3bd/clients/drcachesim/tracer/instru_offline.cpp#L385), similar to page and cache sizes.

Thanks. I'll have a look in to it.

> Question: Can an application use multiple vector lengths throughout its execution by setting ZCR_Elx.LEN[1](https://developer.arm.com/documentation/102476/0100/SVE-architecture-fundamentals/Configurable-vector-length) to different values? Or is the vector length guaranteed to be fixed throughout for an app?

Technically at a hardware level, yes you can change the vector length by writing to `ZCR_Elx` and Linux provides a system call that allows you to change the vector length for the current thread, but as I understand it doing so during a running application is considered undefined behaviour and DynamoRIO doesn't support it. DynamoRIO samples the VL once during initialization and assumes that it stays the same for the duration of the application.

jackgallagher-arm:
> > 2. The drcachesim process ends up with two copies of drdecode.a: one linked in to drcachesim, one linked in to libdynamorio.so. Is this intended?
> 
> No. We have put effort into supporting the same libdynamorio used for both standalone decoding and managed mode (at different time of course) in the same process, so the separate copy should not be needed and it is surprising to see it.

Right now there are two functions used by the codec which are not exported by `libdynamorio`:
`dr_set_sve_vector_length(int vl)` and `dr_get_sve_vector_length()`. This means when using `libdynamorio` you can't call `dr_set_sve_vector_length(vl)` to override the vector length that the codec uses. If we want 'libdynamorio' and 'drdecode' to be equivalent we will need to export those functions or rethink how it gets configured.

> Is it easy to remove it?

I was able to make it to link correctly with one copy of `drdecode.a` by manually modifying the linker command, but I wasn't able to get CMake to generate that link command. I think we will need to look at the graph of dependency libraries to get this to build robustly with one copy of 'drdecode.a'

abhinav92003:
> Technically at a hardware level, yes you can change the vector length by writing to `ZCR_Elx` and Linux provides a system call that allows you to change the vector length for the current thread, but as I understand it doing so during a running application is considered undefined behaviour and DynamoRIO doesn't support it.

Do you mean the hardware considers it undefined behaviour or DynamoRIO does? I assume you mean the latter. Well, we do have pre and [post](https://github.com/DynamoRIO/dynamorio/blob/960d1d7f45b4772cb94f748d0218a9091cbee3bd/core/unix/os.c#L8747) system call control points in DynamoRIO where we can potentially add some special handling -- maybe invoke some hook in drcachesim which adds a marker to the raw trace denoting a change in vector size; in raw2trace we can then make the required adjustments when these markers are found.

derekbruening:
> Right now there are two functions used by the codec which are not exported by libdynamorio:
dr_set_sve_vector_length(int vl) and dr_get_sve_vector_length(). This means when using libdynamorio you can't call dr_set_sve_vector_length(vl) to override the vector length that the codec uses. If we want 'libdynamorio' and 'drdecode' to be equivalent we will need to export those functions or rethink how it gets configured.

All standalone decode/encode functions should be available in libdynamorio.  Looking at dr_set_sve_vector_length() and dr_get_sve_vector_length() in encode_api.h: they are missing `DR_API` specifiers and this is why they are not exported when in a shared library.

derekbruening:
> they are missing DR_API specifiers and this is why they are not exported when in a shared library

It may be worth a quick audit of any other new interfaces added recently to ensure they all have `DR_API`.

jackgallagher-arm:
> Do you mean the hardware considers it undefined behaviour or DynamoRIO does? I assume you mean the latter.

I mean that language runtimes treat the vector length as a runtime constant. For example LLVM's [vector type IR](https://llvm.org/docs/LangRef.html#vector-type) uses a `vscale` constant to represent scalable vectors:

>  For scalable vectors, the total number of elements is a constant multiple (called vscale) of the specified number of elements; vscale is a positive integer that is unknown at compile time and the same hardware-dependent constant for all scalable vectors at run time. The size of a specific scalable vector type is thus constant within IR, even if the exact size in bytes cannot be determined until run time.

So although an application author could call `prctl(...)` and change the vector length, the compiler isn't expecting it to change and you might end up in a bad state if you do so. For example, if you had spilled a vector register to memory before changing the vector length, and try to restore it after.

> Well, we do have pre and post system call control points in DynamoRIO where we can potentially add some special handling -- maybe invoke some hook in drcachesim which adds a marker to the raw trace denoting a change in vector size; in raw2trace we can then make the required adjustments when these markers are found.

Thanks. The current plan is to keep things as they are and assume that the vector length doesn't change but its worth knowing there is a possible way forward if that plan changes.

derekbruening:
Generally, we want to be able to run any program that the hardware supports, not just those that follow software conventions in compilers or ABIs, since there always seem to be real programs that violate those conventions, and there are cases where DR is used to run deliberately violating programs (analyzing malware, etc.).  Only when such support is intractable do we reluctantly relax that and assume conventions: such as some rseq corner cases.  So if it is practical to support it changing we would want to do that.  We should add a handler for the prctl call that at least detects a change and provides a warning or error with a TODO to actually handle it.

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -3738,6 +3846,17 @@ raw2trace_t::raw2trace_t(
     decode_cache_.reserve(cache_count);
     for (int i = 0; i < cache_count; ++i)
         decode_cache_.emplace_back(cache_count);
+
+#if defined(AARCH64)
+    // TODO i#5365: The decoder uses a global sve_veclen variable to store the vector
+    // length value it uses when decoding. drdecodelib ends up being linked into
+    // drcachesim twice: once into the drcachesim executable, and one into libdynamorio.
```
derekbruening, line 3853: We wanted to avoid needing full standalone DR in raw2trace but were unable to do so.  Given that we have standalone DR, we should remove the drdecode that's linked in.  That should avoid all the duplication.  (Xref #1684 on having libdynamorio link the actual drdecode which I think is not simple; but even w/ that we'd need to change these executable to use static libdynamorio to de-dup with both via that route.). (Xref the missing DR_API which should be fixed as well if these routines are not available in libdynamorio.)

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -2882,6 +2882,82 @@ raw2trace_t::set_instr_summary_flags(raw2trace_thread_data_t *tdata, uint64 modi
     return true;
 }
 
+#if defined(AARCH64)
+/* TODO i#5365: append_bb_entries() takes the size of the scatter/gather memory operand
+ * to be the per-element value and uses that to create the read/write memref entries.
+ * The AArch64 IR currently uses the maximum amount of data transferred by the instruction
+ * (number of elements * per element size) instead so until we change the codec to use the
```
derekbruening, line 2889: Hmm I think x86 also has the max size -- right @abhinav92003?  I see `OP_vpscatterdd, 0x6638a008, catSIMD, "vpscatterdd", MVd, KEw, KEw, Ve, xx, mrm|evex|reqp|ttnone|nok0, x, END_LIST},` and `#define Mvd TYPE_M, OPSZ_16_vex32`.  So what happens today for x86 for the per record size?

abhinav92003: Looking at the trace for clients/drcachesim/tests/allasm_scattergather_x86.asm, each mem accesses has the per-element size:

```
          56          45:     1393700 ifetch      10 byte(s) @ 0x0000000000401124 c4 22 11 90 24 9d 39 vpgatherdd 0x00402039(,%xmm11,4), %xmm13, %xmm12
          56          45:     1393700                                             20 40 00
          57          45:     1393700 read         4 byte(s) @ 0x000000000040203d by PC 0x0000000000401124
          58          45:     1393700 read         4 byte(s) @ 0x0000000000402045 by PC 0x0000000000401124
          59          45:     1393700 read         4 byte(s) @ 0x0000000000402041 by PC 0x0000000000401124
```

As linked in my other comment, I saw a different definition for `MVd`, which made sense so I didn't look further.

```
#define MVd TYPE_VSIB, OPSZ_4
```

but looks like there are two definitions. Not sure how that works.

derekbruening: I am looking into why we don't get a warning on the re-definition, and whether this is making some non-scatter instructions incorrectly have OPSZ_4 as their size (we only want the element size for scatter/gather or rep-string -- regular SIMD should have the total size).

derekbruening: Ah, mystery solved: it's `MVd` vs `Mvd`: the case of the `v` is different.  (My emacs does a case-insensitive interactive search.)

jackgallagher-arm: > we only want the element size for scatter/gather or rep-string -- regular SIMD should have the total size

What about predicated contiguous load/store instructions? I'm not sure there is a direct equivalent in x86. Because they have a predicate mask like the scatter/gather instructions we treat them like scatter/gather instructions in a lot of places in DynamoRIO so I would suggest we keep them consistent with scatter/gather instructions and use the per-element size.

derekbruening: There are mask registers in AVX-512 which I think apply to many operations and not just scatter/gather so I think the same concept does exist there.

Hmm, it's seeming like we want dr_opnd_query_flags_t for the size now.  For predicated contiguous, a taint-tracking tool (such as Dr. Memory) might want the max size for loads when checking taint bits (on a fastpath anyway) but would have to loop over the per-element for stores when setting taint bits.  Hmm.  I guess the slowpath would loop too.  Maybe file an issue on this predicated size problem covering all platforms, and go ahead w/ your current plan for now?

This is similar to the half-register and holes-in-register complex SIMD interleaving operations: xref #1382, #6218.

Xref #5638.

jackgallagher-arm:
> Generally, we want to be able to run any program that the hardware supports, not just those that follow software conventions in compilers or ABIs, since there always seem to be real programs that violate those conventions, and there are cases where DR is used to run deliberately violating programs (analyzing malware, etc.). Only when such support is intractable do we reluctantly relax that and assume conventions: such as some rseq corner cases. So if it is practical to support it changing we would want to do that. We should add a handler for the prctl call that at least detects a change and provides a warning or error with a TODO to actually handle it.

That makes sense. We will create a PR to detect vector length changes and issue a warning. Thanks.

clients/drcachesim/tracer/raw2trace.cpp
jackgallagher-arm: Thanks. Created #6561

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -2882,6 +2882,82 @@ raw2trace_t::set_instr_summary_flags(raw2trace_thread_data_t *tdata, uint64 modi
     return true;
```
derekbruening, line 2882: Seems like #5036 should be referenced in the description?

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -2882,6 +2882,82 @@ raw2trace_t::set_instr_summary_flags(raw2trace_thread_data_t *tdata, uint64 modi
...
 }
 
+#if defined(AARCH64)
+/* TODO i#5365: append_bb_entries() takes the size of the scatter/gather memory operand
+ * to be the per-element value and uses that to create the read/write memref entries.
+ * The AArch64 IR currently uses the maximum amount of data transferred by the instruction
+ * (number of elements * per element size) instead so until we change the codec to use the
+ * per-element size we need to use this function to set the per-element size based on the
+ * instruction opcode.
+ * When we have made the codec/IR changes this function can be removed.
```
derekbruening, line 2892: An alternative solution is to store the encoding during tracing like we do for vdso or JIT code (see offline_instru_t::record_instr_encodings()), but not the encoding of the emulated instr info's pointer to the original scatter/gather: rather, the unrolled single load/store.  This would greatly simplify raw2trace and questions about what the IR should store.  OTOH it adds some tracing overhead: which probably outweighs gains in reducing code complexity.  So I'm not voting that we do it.  But how about mentioning it here as a possible alternative.

derekbruening (approved):

clients/drcachesim/tracer/raw2trace.cpp
jackgallagher-arm: Done.

clients/drcachesim/tracer/raw2trace.cpp
jackgallagher-arm: Done.

jackgallagher-arm:
Proper fix for the memory operand sizes: #6574

derekbruening:
This PR has broken our internal build: we get undefined references to `dr_set_sve_vector_length`, due to what was noted above.  Filed #6575 as this is blocking us.  I will add the missing DR_API to expedite the fix.

derekbruening:
We have hit two more problems around vector lengths: a buffer overflow #6584, and incorrectly using the underlying machine's length when processing a trace gathered on a different architecture, which was added by this PR: #6585.

PR #6545, opened 2024-01-09
Use 64-bit in Linux examples in docs

derekbruening:
Updates the example command lines in the Linux Deployment docs to use 64-bit which is far more common these days.

### DIFF
--- a/api/docs/deployment.dox
+++ b/api/docs/deployment.dox
@@ -1,5 +1,5 @@
 /* ******************************************************************************
- * Copyright (c) 2010-2022 Google, Inc.  All rights reserved.
+ * Copyright (c) 2010-2024 Google, Inc.  All rights reserved.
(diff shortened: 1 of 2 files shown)


derekbruening:
(Will confirm all failures are flakes pre-merge but almost certainly they will be for this docs tweak)

api/docs/deployment.dox
```diff
@@ -1,5 +1,5 @@
 /* ******************************************************************************
```
abhinav92003, line 1: In the section on "Cross-Compiling for ARM on Linux" in building.dox, the cmake command could also be updated to use the arm64 config.

abhinav92003 (approved):

api/docs/deployment.dox
derekbruening: Added a new section above it for cross-compiling for 64-bit arm.

derekbruening:
x64 is client.attach #6452; win64 is replaceall #5412

Issue #6486, opened 2023-12-01
libipt decode fails due to missing mapped addresses


abhinav92003:
I'm seeing errors like the following happen sporadically during decoding of Intel-PT kernel traces:

```
408:   *** postcmd failed (1): drpt2ir: [28430, IP:ffffffffc11dd000] get next
408:   instruction error: no memory mapped at this address
```

When it happens, it's always the same address that's unmapped. When I hard-coded that address  (and another that was revealed when this error was fixed) to be copied in kcore_copy, the error went away. I couldn't find details of the address in /proc/modules or /proc/kallsyms but it is a part of the kcore code section (we copy only the memory that corresponds to the live kernel modules from /proc/modules). It's presumably some kernel code that's executing during system calls.

The errors don't happen every time but they seem to have started happening frequently enough. Not sure if it was some change in my machine's kernel that caused the unmapped instrs to be executed.

dolanzhao:
I have successfully tested the main branch on my local system and can confirm that it is functioning as expected.

derekbruening:
I hit this missing address error in every run in these 3 tests on my machine:
```
        408 - code_api|tool.drcacheoff.kernel.simple_SUDO (Failed)
        409 - code_api|tool.drcacheoff.kernel.opcode-mix_SUDO (Failed)
        410 - code_api|tool.drcacheoff.kernel.syscall-mix_SUDO (Failed)
```
It is a Debian-ish 6.5.6 kernel.  Should be same as @abhinav92003 's machine.

abhinav92003:
FTR the following is the local workaround I'm using. It adds two addresses that showed up in error messages. They may be different on different machines of course.


```
$ git diff
diff --git a/clients/drcachesim/tracer/kcore_copy.cpp b/clients/drcachesim/tracer/kcore_copy.cpp
index 962ab73f7..1e2424636 100644
--- a/clients/drcachesim/tracer/kcore_copy.cpp
+++ b/clients/drcachesim/tracer/kcore_copy.cpp
@@ -370,6 +370,21 @@ kcore_copy_t::read_modules()
             last_module = module;
         }
     }
+    proc_module_t *module = (proc_module_t *)dr_global_alloc(sizeof(proc_module_t));
+    module->start = 0xffffffffc039b000;
+    module->end = 0xffffffffc039b000 + 0x10000;
+    module->next = nullptr;
+    kcore_code_segments_num_++;
+    last_module->next = module;
+    last_module = module;
+
[...]

PR #6552, opened 2024-01-10
i#6486 Update libipt to Stable v2.1

dolanzhao:
This update upgrades libipt to its latest stable version (v2.1). The new version of libipt introduces support for newer CPU models and addresses several decoding-related bugs. 

Issue: #6486


abhinav92003 (approved):

abhinav92003:
Looks like the clang-format workflow failures are from libipt. Can we somehow disable clang for that dir?

abhinav92003:
> Looks like the clang-format workflow failures are from libipt. Can we somehow disable clang for that dir?

Saw later that the failures are also there in other workflows. So it's a real build issue with the new libipt? Maybe some changes are needed to our build?

Issue #1551, opened 2014-11-28
port to ARM master issue


derekbruening:
_From [bruen...@google.com](https://code.google.com/u/109494838902877177630/) on October 08, 2014 00:19:33_

xref issue #1550 xref issue #269 This is a container case for references in the code for tasks we're delaying while we're porting.

_Original issue: http://code.google.com/p/dynamorio/issues/detail?id=1551_

derekbruening:
After a massive amount of work and hundreds of commits, AArch32 support is pretty much complete.  We can run large applications now (including Chrome with simple workloads, though some bugs are seen there).  There are still some known missing features (IT splits, some cases of fault handling) but we'll treat those as corner cases and file separate issues on them from here on out.

algr:
The ARM port is looking great.  It is already very functional, and it seems right that this issue is now closed.  The source still has several comments "FIXME i#1551: NYI on ARM".  Perhaps these could be made specific to open issues for individual features?

zhaoqin:
```
> git grep "FIXME i#1551: NYI on ARM" | wc -l
74
```
We put FIXME comment back then and thought those are much less common cases. Some are small fixes and some need more efforts. ASSERT will be triggered if any is hit, so we can file issue or fix it then.

derekbruening:
> The source still has several comments "FIXME i#1551: NYI on ARM". Perhaps these could be made
> specific to open issues for individual features?

Yes, these should be cleaned up.  Some should be removed as they are in code we do not plan to ever port.  Likely some should be updated to already-filed issues.  There may be some that need new issues: someone will need to go through them.  Any volunteers?

PR #6554, opened 2024-01-11
i#3544, i#1551, i#1569 signal: Remove a non-existing field in sigframe_rt_t

ksco:
The `retcode` field in `sigframe_rt_t` is only available in x86 and AArch32, this fixes a buffer overflow in `memcpy_rt_frame()` on RISC-V.

Issue: #3544 #1551 #1569

### DIFF
--- a/core/unix/signal.c
+++ b/core/unix/signal.c
@@ -3644,7 +3644,6 @@ convert_frame_to_nonrt(dcontext_t *dcontext, int sig, sigframe_rt_t *f_old,
     f_new->sig_noclobber = f_new->sig;
 #    elif defined(ARM)
     memcpy(&f_new->uc, &f_old->uc, sizeof(f_new->uc));
-    memcpy(f_new->retcode, f_old->retcode, sizeof(f_new->retcode));
     /* now fill in our extra field */
     f_new->sig_noclobber = f_old->info.si_signo;
 #    endif /* X86 */
--- a/core/unix/signal.c
+++ b/core/unix/signal.c
@@ -3644,6 +3644,7 @@ convert_frame_to_nonrt(dcontext_t *dcontext, int sig, sigframe_rt_t *f_old,
     f_new->sig_noclobber = f_new->sig;
 #    elif defined(ARM)
     memcpy(&f_new->uc, &f_old->uc, sizeof(f_new->uc));
+    memcpy(f_new->retcode, f_old->retcode, sizeof(f_new->retcode));
     /* now fill in our extra field */
     f_new->sig_noclobber = f_old->info.si_signo;
 #    endif /* X86 */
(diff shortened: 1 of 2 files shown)


core/unix/signal_private.h
```diff
@@ -280,11 +284,9 @@ typedef struct rt_sigframe {
 #    elif defined(AARCHXX)
     kernel_siginfo_t info;
     kernel_ucontext_t uc;
-    char retcode[RETCODE_SIZE];
```
derekbruening, line 283: I believe AArch32 has retcode in the frame set up by the kernel.  I think it is not just x86: also AArch32.

core/unix/signal.c
```diff
@@ -3644,7 +3644,6 @@ convert_frame_to_nonrt(dcontext_t *dcontext, int sig, sigframe_rt_t *f_old,
     f_new->sig_noclobber = f_new->sig;
 #    elif defined(ARM)
     memcpy(&f_new->uc, &f_old->uc, sizeof(f_new->uc));
-    memcpy(f_new->retcode, f_old->retcode, sizeof(f_new->retcode));
```
derekbruening, line 3647: I think AArch32 does have retcode: I see it in https://elixir.bootlin.com/linux/latest/source/arch/arm/kernel/signal.c#L402

core/unix/signal_private.h
```diff
@@ -214,7 +214,11 @@ typedef _STRUCT_UCONTEXT /* == __darwin_ucontext */ kernel_ucontext_t;
  * (these are from /usr/src/linux/arch/i386/kernel/signal.c for kernel 2.4.17)
  */
 
-#    define RETCODE_SIZE 8
+#    if defined(X86)
+#        define RETCODE_SIZE 8
+#    elif defined(ARM)
+#        define RETCODE_SIZE 16
```
derekbruening, line 220: Hmm, could you explain where this came from?

core/unix/signal.c
ksco: I'm very sorry for misreading this, yes, AArch32 has a retcode.

core/unix/signal_private.h
ksco: It's from https://elixir.bootlin.com/linux/latest/source/arch/arm/kernel/signal.h#L5

core/unix/signal_private.h
ksco: Yes, you're right, there is a `retcode` in AArch32, but not in AArch64.

core/unix/signal_private.h
derekbruening: I see, thanks.

core/unix/signal_private.h
```diff
@@ -214,7 +214,11 @@ typedef _STRUCT_UCONTEXT /* == __darwin_ucontext */ kernel_ucontext_t;
  * (these are from /usr/src/linux/arch/i386/kernel/signal.c for kernel 2.4.17)
```
derekbruening, line 214: Please update the PR/merge description to include arm.

derekbruening (approved):

PR #6557, opened 2024-01-11
i#6514 null SP to clone(), part 1: Move dynamorio_clone to drlibc

xdje42:
This patch is just moving code, no functional changes. dynamorio_clone is moved to drlibc so that it can also be used by tests.

Issue: #6514


derekbruening:
Usually we wait for green tests before requesting review.  (The last few months with the AMD cloud changes x32 has always been red; plus a couple of flaky tests have gotten more repeatable: usually we explain each failure w/ a pointer to the issue, so "wait for green tests or green + explanation of red".)

derekbruening (commented):
I should have reset the review assignment with my comment about the tests not passing.  Please re-request review when the tests are green or red-with-explanation.

xdje42:
Required tests pass now, except for amd64 because of the client.detach_test (known to have problems).

core/drlibc/drlibc_aarch64.asm
```diff
@@ -125,4 +129,32 @@ GLOBAL_LABEL(dynamorio_mach_syscall:)
...
+GLOBAL_LABEL(dynamorio_clone:)
+        stp      ARG6, x0, [ARG2, #-16]! /* func is now on TOS of newsp */
+        /* All args are already in syscall registers. */
+        mov      SYSNUM_REG, #SYS_clone
+        svc      #0
+        cbnz     x0, dynamorio_clone_parent
+        ldp      x0, x1, [sp], #16
+        blr      x0
+        /* aarch64 doesn't provide unexpected_return (ref i#4304). Instead of
+         * providing our own copy, just inline its definition here.
```
derekbruening, line 146: We could have aarch64 include it in drlibc but named `drlibc_unexpected_return` and then reference it here and it would reach, right?  That seems nicer than duplicating it in the caller.

derekbruening (approved):

core/drlibc/drlibc_aarch64.asm
```diff
@@ -125,4 +129,32 @@ GLOBAL_LABEL(dynamorio_mach_syscall:)
...
+        DECLARE_FUNC(dynamorio_clone)
+GLOBAL_LABEL(dynamorio_clone:)
+        stp      ARG6, x0, [ARG2, #-16]! /* func is now on TOS of newsp */
+        /* All args are already in syscall registers. */
+        mov      SYSNUM_REG, #SYS_clone
+        svc      #0
+        cbnz     x0, dynamorio_clone_parent
+        ldp      x0, x1, [sp], #16
+        blr      x0
+        /* aarch64 doesn't provide unexpected_return (ref i#4304). Instead of
```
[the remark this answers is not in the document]
xdje42: Wording changed to "no functional changes" which is what I meant to say.

core/drlibc/drlibc_aarch64.asm
xdje42: Code updated to call unexpected_return

PR #6560, opened 2024-01-12
i#5926: Update elfutils submodule to https protocol

ksco:
Use HTTPs version of the elfutils submodule URL as it's easier for users behind an HTTP proxy.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -6,4 +6,4 @@
     url = https://github.com/madler/zlib.git
 [submodule "third_party/elfutils"]
     path = third_party/elfutils
-    url = git://sourceware.org/git/elfutils.git
+    url = https://sourceware.org/git/elfutils.git


fmoessbauer:
Thanks, I stumbled upon this as well.  I'm just wondering why the github UI link is still broken.

ksco:
> Thanks, I stumbled upon this as well. I'm just wondering why the github UI link is still broken.

What do you mean by "github UI link"?

fmoessbauer:
grafik

This link still points to https://github.com/DynamoRIO/dynamorio/blob/master/third_party/elfutils , but this might also be a github bug. Don't know.

ksco:
> This link still points to https://github.com/DynamoRIO/dynamorio/blob/master/third_party/elfutils , but this might also be a github bug. Don't know.

Ah, I believe this is how GitHub displays submodules of external sites, no need to worry.

derekbruening (approved):

Issue #6558, opened 2024-01-12
TEST HANG(10.0.19732 client.attach_blocking): Test hands if yama.ptrace_scope != 0


xdje42:
**Describe the bug**

The client.attach_blocking test hangs if yama.ptrace_scope != 0.

**To Reproduce**

```
$ ctest -R client.attach_blocking
<hang>
Ctrl-Z
$ ps
    PID TTY          TIME CMD
3715888 pts/12   00:00:00 bash
3748099 pts/12   00:00:00 ctest
3748101 pts/12   00:00:00 cmake
3748103 pts/12   00:00:00 linux.infloop
3748479 pts/12   00:00:00 sleep
3748480 pts/12   00:00:00 ps
$ kill %1
$ sudo sysctl -w kernel.yama.ptrace_scope=0
$ ctest -R client.attach_blocking
<no hang>
```

**Versions**
 - What version of DynamoRIO are you using? HEAD as of 1/11/2024 (commit d28ac5bb1bed9283d0cfeea4c904644878ca3f1c)
 - Does the latest build from https://github.com/DynamoRIO/dynamorio/releases solve the problem? No
- What operating system version are you running on? ("Windows 10" is *not* sufficient: give the release number.) Linux, amd64 and aarch64
- Is your application 32-bit or 64-bit? 64-bit

**Additional context**

Related issue: #37
Related issue: #38

derekbruening:
The action item would be for the test config to look at the procfs setting and gate it -- like we do for sudo tests but automated or sthg.  Or better for drrun (or drinjectlib) to try to check privs -- but that may not be easy to check all cases of missing privs.

derekbruening:
Another action item is to shrink the timeout which is excessive.

derekbruening:
I am fixing the timeout issue in #6563

PR #6563, opened 2024-01-12
i#6558, i#6127: Set CTest timeouts for all tests

### LINKED ISSUE #6127: Speed up test suite to <20 minutes from >30 minutes
Xref some discussion at https://github.com/DynamoRIO/dynamorio/pull/6096#issuecomment-1569110298 and https://github.com/DynamoRIO/dynamorio/pull/6096#issuecomment-1578824111 at adding parallelism to the ci-aarch64-native runner to avoid a backlog from simultaneous PR's waiting on each other.

The latency of the ci-aarch64-native job is also something to look at: when it was Jenkins it was pretty fast, under 10 minutes IIRC.  Now you can see it is consistently over 33 minutes and is by far the slowest of our jobs:

https://github.com/DynamoRIO/dynamorio/actions/workflows/ci-aarch64-native.yml

The general goal is to be under 20 minutes.  The Windows and Linux x86 jobs have gotten slower over time and are both over 20 minutes now.  This issue covers trying to speed all of these up.  The ci-aarch64-native seems the highest priority at the longest and because it used to be faster making it seems like there may be some low-hanging fruit there for improvements.

derekbruening:
Before, we relied on drrun -s for all test suite timeouts except for runcmp tests where we set a CTest timeout.  This resulted in the default 10 minute CTest timeout for all tests, which was the only timeout for runall tests and caused long suite times on the AArch64 machine which accidentally had no ptrace privileges (#5740, #6558,

Here, we set the CTest time for runall in addition to runcmp, and for all other tests with no timeout specified (which are presumably relying on drrun -s) we set a timeout of the drrun timeout plus 30 seconds.

Tested on the attach test:

Before:
```
123: Test timeout computed to be: 1500
```
Now:
```
$ echo 1 | sudo tee /proc/sys/kernel/yama/ptrace_scope; /usr/bin/time ctest -V -R client.attach_test; echo 0 | sudo tee /proc/sys/kernel/yama/ptrace_scope
...
    Start 121: code_api|client.attach_test
...
121: Test timeout computed to be: 90
1/1 Test #121: code_api|client.attach_test ......***Timeout  90.11 sec
...
The following tests FAILED:
	121 - code_api|client.attach_test (Timeout)
...
Command exited with non-zero status 8
1.13user 0.80system 1:30.14elapsed 2%CPU (0avgtext+0avgdata 13196maxresident)k
```

The property being set on more tests was confirmed on debug x86-64:
Before:
```
$ grep -c TIMEOUT suite/tests/CTestTestfile.cmake
94
```
After:
```
$ grep -c TIMEOUT suite/tests/CTestTestfile.cmake
463
```
There seem to still be a few missing the property: the ones that don't go through suite/.  There were other efforts to avoid hangs on those such as PR #6137.

Fixes #6127
Issue: #6127, #6558, #5740

### DIFF
--- a/suite/tests/CMakeLists.txt
+++ b/suite/tests/CMakeLists.txt
@@ -1,5 +1,5 @@
 # **********************************************************
-# Copyright (c) 2010-2023 Google, Inc.    All rights reserved.
+# Copyright (c) 2010-2024 Google, Inc.    All rights reserved.
 # Copyright (c) 2009-2010 VMware, Inc.    All rights reserved.
 # Copyright (c) 2016-2023 ARM Limited.    All rights reserved.
 # **********************************************************
@@ -1631,9 +1631,14 @@ function(torun test key source native standalone_dr dr_ops exe_ops added_out pas
     # Though we mostly use drrun's -s we set this too in case the requested
     # value is higher than ctest's default.
     set_tests_properties(${test} PROPERTIES TIMEOUT ${${key}_timeout})
-  elseif (is_runcmp)
-    # Runcmp generally doesn't use drrun or runstats so we need a ctest timeout.
+  elseif (is_runcmp OR is_runall)
+    # Runcmp/runall generally don't use drrun or runstats so we need a ctest timeout.
     set_tests_properties(${test} PROPERTIES TIMEOUT ${TEST_SECONDS})
+  else ()
+    # Even though we expect drrun -s to enforce a timeout, set one in ctest just
+    # in case, but give time for drrun first.
+    math(EXPR timeout "${TEST_SECONDS}+30")
+    set_tests_properties(${test} PROPERTIES TIMEOUT ${timeout})
   endif ()
 
   # Though we use drrun and runstats -s timeout parameters, we have


derekbruening:
amd32-bit failure is #6417 (getting to the point where we should either remove it from the set of jobs or have a Fixit to get it green...)

xdje42:
I don't yet have an opinion on this patch, but wouldn't a better fix for the yama ptrace_scope tests (eg, #6558) be to have them fail fast instead of timing out?

derekbruening:
> I don't yet have an opinion on this patch, but wouldn't a better fix for the yama ptrace_scope tests (eg, #6558) be to have them fail fast instead of timing out?

This is not trying to fix the attach/detach tests: it is ensuring the timeout value of 90s we thought we had set for all tests is actually set for all tests instead of having some tests fall back to the general CTest default of 10 minutes which we never want.

derekbruening:
(Generally it's better to use the review comments to have an actual threaded conversation rather than trying to do so in these top-level comments -- please use the review interface.)

xdje42:
JTBC, Is the rule that review comments shall be used for all comments?
Or put another way, when *would* one use the top level comments?

derekbruening:
> JTBC, Is the rule that review comments shall be used for all comments? Or put another way, when _would_ one use the top level comments?

IMHO, in a PR they are only useful for noting flaky failures.  There is no threading or marking of "resolved" to gate submission so they are not useful for much else.

xdje42 (approved):

Issue #4422, opened 2020-08-27
rebulid-elftc.sh launch through CMake


ofry:
**Add a pointer to any prior users list discussion.**
Please use the users list https://groups.google.com/forum/#!forum/DynamoRIO-Users for questions or to propose initial feature ideas if you're not sure there's an existing solution.  The users list will reach a wider audience of people who may find the information beneficial.  If you have already asked on the users list and the consensus was to file a new issue, please include the URL to the discussion thread here.

**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is.  E.g., "I'm always frustrated when [...]"

It will be easier to debug this script, if it could be launched via CMake on MinGW-w64. Related to issue https://github.com/DynamoRIO/dynamorio/issues/638

**Describe the solution you'd like**
Describe the desired functionality and its intended usage to give some context for how it would be used.

Do you have any implementation in mind for this feature?

**Describe alternatives you've considered**
Is there currently a method for implementing this functionality using the existing API and this requested feature will add convenience?  Or is there no method for accomplishing the desired task using the current API?

**Additional context**

derekbruening:
It is not clear what you mean (and leaving all the template instructions so the one (!) sentence added is buried doesn't help): are you proposing adding the source code of libelftc into the DR repository and building it as part of every DR build?  If so, be explicit: "launch through CMake" does not make that clear.  If libelftc were built as part of the DR build, we would probably not use a shell script and use first-class CMake code.  But we made a decision back when we add libelftc to *not* include the sources and include the binaries instead.  Please elaborate on the logic behind this proposal.  ("debug this script" is not clear: do you mean debugging the rebuild-elft.sh (spelled wrong in the title) script itself?  It's a pretty short sequence of commands, with all the arguments right there: that seems a lot easier to understand than CMake code where the final set of flags is not as visible in one place.)

ofry:
I mean "if binaries are absent, launch this *.sh script directly from CMake".

derekbruening:
> I mean "if binaries are absent, launch this *.sh script directly from CMake instead of abort compilation".

But the binaries are checked in to the repository, so they will never be absent.

ofry:
For mingw it's absent for now!

derekbruening:
But the drsyms CMakeLists.txt for mingw will either be WIN32 or UNIX and so it will pick a checked-in binary and try to use it, right?

[...]

PR #6564, opened 2024-01-12
i#5926 elfutils: Remove elftc lib{dwarf,elf}.a from Linux targets

derekbruening:
While drsyms on Linux still uses libelftc.a for demangling when template arg expansion is needed, as elfutils does not supply that, the checked-in elftc libdwarf.a and libelf.a binaries for Linux can be removed. (If we added official *BSD or other UNIX variant support we would just make those use elfutils and would not want these static libs there either.)

Issue: #5926, #4422


derekbruening:
(32-bit amd will fail with #6417...attempts to slowly improve that are in progress in the background)

fmoessbauer (approved):

Issue #6536, opened 2024-01-02
client.detach_test failing: kill cannot find target process


derekbruening:
The Linux client.detach_test was just added in PR #6513 and it is failing (non-deterministically) on x86-64.

https://github.com/DynamoRIO/dynamorio/actions/runs/7381905578/job/20080993968?pr=6531
```
      Test #185: code_api|client.detach_test ......................................***Failed  Required regular expression not found. Regex=[^starting
thank you for testing detach
thread init
detach
done
$
]  0.26 sec
CMake Error at /home/runner/work/dynamorio/dynamorio/suite/tests/runall.cmake:130 (message):
  *** kill failed (1): kill: (24600): No such process
  ***
Call Stack (most recent call first):
  /home/runner/work/dynamorio/dynamorio/suite/tests/runall.cmake:317 (kill_background_process)
```

derekbruening:
Xref client.attach_test failing with a similar error; the frequency of its failures may have increased with PR #6513?  Maybe not enough data to say for that but it failed 2 out of 4 times in recent PR's.

derekbruening:
@onroadmuwl hoping you can help eliminate the flakiness in this new test that you added

onroadmuwl:
> @onroadmuwl hoping you can help eliminate the flakiness in this new test that you added

I'm sorry to hear that. I'm not able to reproduce this problem in my local testing. And I haven't seen similar problems when I submit code for previous PR. There may be a conflict with the newly submitted code. I'm trying to debug the test suite step by step in a new PR [#6537](https://github.com/DynamoRIO/dynamorio/pull/6537 ). If I delete the test code for Linux's detachment, the attach test works well. So I think there may be some problems in the detach test instead of real detach code's problem.

derekbruening:
The attach test was failing every once in a while with the same output: #6452.  So it may be the same underlying pre-existing issue that is just more frequent in the detach test.

From the output it looks like the attach never happened b/c it never prints "thank you for testing..."?

derekbruening:
@onroadmuwl did you try running many times inside the ctest framework to see if it reproduces non-deterministically on your local machine?
E.g.:
```
$ ctest --repeat-until-fail 500 -R client.detach
Test project /usr/local/google/home/bruening/dr/git/build_x64_dbg_tests
    Start 185: code_api|client.detach_test
    Test #185: code_api|client.detach_test ......   Passed    0.35 sec
    Start 185: code_api|client.detach_test
[...]

PR #6566, opened 2024-01-17
i#4304 unexpected_return unreachable on aarch64: Use long branch

### LINKED ISSUE #4304: AArch64 static DR link error: relocation out of range for unexpected_return

xdje42:
Fix i#4304 in a different way by always using a branch/call instruction with enough bits of offset to reach the unexpected_return function. Conditional branches don't always have enough offsets to branch to any function within a binary: they're generally only intended for branching within a single function. On aarch64 conditional branch instructions have 19 bits of offset and unconditional branch and call instructions have 26 bits, and is used for every other function call so we should be good for awhile.

Fixes #4304

### DIFF
--- a/core/arch/aarch64/aarch64.asm
+++ b/core/arch/aarch64/aarch64.asm
@@ -104,19 +104,6 @@ DECL_EXTERN(initstack_mutex)
 DECL_EXTERN(icache_op_struct)
 DECL_EXTERN(linkstub_selfmod)
 
-/* For debugging: report an error if the function called by call_switch_stack()
- * unexpectedly returns.  Also used elsewhere.
- */
-        DECLARE_FUNC(unexpected_return)
-GLOBAL_LABEL(unexpected_return:)
-        CALLC3(GLOBAL_REF(d_r_internal_error), HEX(0), HEX(0), HEX(0))
-        /* d_r_internal_error normally never returns */
-        /* Infinite loop is intentional.  Can we do better in release build?
-         * XXX: why not a debug instr?
-         */
-        JUMP  GLOBAL_REF(unexpected_return)
-        END_FUNC(unexpected_return)
-
 /* bool mrs_id_reg_supported(void)
  * Checks for kernel support of the MRS instr when reading system registers
  * above exception level EL0, by attempting to read Instruction Set Attribute
@@ -158,7 +145,9 @@ call_dispatch_alt_stack_no_free:
         /* Switch stack back. */
         mov      sp, x19
         /* Test return_on_return. */
-        cbz      w20, GLOBAL_REF(unexpected_return)
+        cbnz     w20, call_dispatch_alt_stack_ok_return
+        bl       GLOBAL_REF(unexpected_return)
+call_dispatch_alt_stack_ok_return:
         /* Restore and return. */
         ldr      x19, [sp, #16]
         ldp      x20, x30, [sp], #32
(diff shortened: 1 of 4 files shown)


xdje42:
The failing tests are not related to the patch.

core/arch/aarch64/aarch64.asm
```diff
@@ -104,19 +104,6 @@ DECL_EXTERN(initstack_mutex)
 DECL_EXTERN(icache_op_struct)
```
derekbruening, line 104: > The failing tests are not related to the patch.

Please cite the issues covering the failed tests (naming the tests) to prove they are known flakes, ensure all failes we've seen are filed, and give a count of PR's that have hit each flake (each cross-ref is auto-added to the issue).  (Too often new contributors assume something is not related to their own patch just because they think they only touched limited files but don't realize ripple effects.)

derekbruening, line 104: > On aarch64 conditional branch instructions have 19 bits of offset and unconditional branch and call instructions have 26 bits, and is used for every other function call so we should be good for awhile.

grammar: s/is used/are used/; s/awhile/a while/

core/drlibc/drlibc_xarch.asm
```diff
@@ -37,15 +37,15 @@
...
- * have a relocation reachability error if we include it here (i#4304), so
- * we limit its use in drlibc to x86 or arm.
- */
-#ifndef AARCH64
 DECL_EXTERN(d_r_internal_error)
 
 /* For debugging: report an error if the function called by call_switch_stack()
  * unexpectedly returns.  Also used elsewhere.
+ * i#4304: When calling this function do not use short branch instructions,
+ * eg, conditional branch instructions on aarch64. They don't have enough
```
derekbruening, line 45: nit: s/eg,/e.g.,/

core/drlibc/drlibc_xarch.asm
```diff
@@ -37,15 +37,15 @@
...
+ * bits of offset to reach this. Instead, use unconditional branch or
+ * call instructions (eg, b, bl on aarch64) that have a sufficient number of
```
derekbruening, line 47: nit: s/eg,/e.g.,/

derekbruening, line 47: Better to not recommend a call instruction on x86 as it requires the stack and this routine is called in fragile places.  Consider call_switch_stack: if something goes wrong the stack may well be bogus and we'll crash if we execute a CALL.

derekbruening (approved):

core/arch/aarch64/aarch64.asm
xdje42: Done

core/drlibc/drlibc_xarch.asm
xdje42: Done

core/drlibc/drlibc_xarch.asm
xdje42: Done

xdje42: Done

core/arch/aarch64/aarch64.asm
xdje42: The x64 failure is #6536 (detach_test).
Previous windows failures have been marked as expected to fail (unit_tests, client.loader).

PR #6569, opened 2024-01-19
i#6423,i#5412: Ignore Windows flaky tests

derekbruening:
Adds two tracing tests to the 64-bit Windows ignore list:

i#6423 tool.drcacheoff.burst_traceopts: though a disturbing assert, many attempts to reproduce including under tmate have failed.

i#5412 tool.drcacheoff.burst_replaceall: this one appears to be non-deterministic output ordering so a true flake.

Issue: #5412, #6423


derekbruening:
Wow, new failures have suddenly appeared, from an OS update?  unit_tests and client.loader on win32; unit_tests on win64.  I filed #6570.

abhinav92003 (approved):

Issue #6561, opened 2024-01-12
What should the memory operand size be for SVE predicated contiguous loads and stores


jackgallagher-arm:
Unpredicated loads and stores use the full size of access (the size of the full vector register) but scatter/gather instructions use the per-element transfer size.

We treat the predicated contiguous load/store instructions the same way as scatter/gather instructions in `drx_expand_scatter_gather()` and they are handled in a similar way in `drcachesim` so it is more convenient if they follow the same convention as the scatter/gather instructions, but other tools might have different needs.

@derekbruening commented:
> Hmm, it's seeming like we want dr_opnd_query_flags_t for the size now.  For predicated contiguous, a taint-tracking tool (such as Dr. Memory) might want the max size for loads when checking taint bits (on a fastpath anyway) but would have to loop over the per-element for stores when setting taint bits.  Hmm.  I guess the slowpath would loop too.  Maybe file an issue on this predicated size problem covering all platforms, and go ahead w/ your current plan for now?

> This is similar to the half-register and holes-in-register complex SIMD interleaving operations: xref #1382, #6218.

> Xref #5638.

_Originally posted by @derekbruening in https://github.com/DynamoRIO/dynamorio/pull/6544#discussion_r1449439366_

PR #6574, opened 2024-01-23
i#5365 AArch64: Change scatter/gather instructions to per-element size

jackgallagher-arm:
Make the AArch64 IR consistent with x86 which already uses the per-element
transfer size for the scatter/gather memory operand size.

This changes the AArch64 codec for the scatter/gather and predicated contiguous
load/store instructions to use the per-element access size for the memory operand
instead of the maximum total transfer size that it used previously, and updates
the tests accordingly.
    
Issues: #5365, #5036, #6561


AssadHashmi:
x86-32 regression related to https://github.com/DynamoRIO/dynamorio/issues/6417.

derekbruening (approved):
I did not check the codec details, just glanced over the other changes: LGTM

Issue #6575, opened 2024-01-23
dr_set_sve_vector_length is not exported, causing link errors when used from libdynamorio


derekbruening:
Quoting from https://github.com/DynamoRIO/dynamorio/pull/6544#issuecomment-1883354811

> All standalone decode/encode functions should be available in libdynamorio. Looking at dr_set_sve_vector_length() and dr_get_sve_vector_length() in encode_api.h: they are missing DR_API specifiers and this is why they are not exported when in a shared library.

This has now broken our internal build with PR #6544 adding a call to dr_set_sve_vector_length to raw2trace:
```
ld: error: undefined reference...: dr_set_sve_vector_length        
```

derekbruening:
There also seems to be no test of either API routine?

derekbruening:
Adding a call to one of these from ir_aarch64.c immediately reproduces the problem:
```
suite/tests/api/ir_aarch64.c:6931: undefined reference to `dr_get_sve_vector_length'
```

derekbruening:
Also, the set routine returns null, but says it will not work if on an SVE system: how does the user know A) whether on an SVE system and B) whether the set routine succeeded?  Should it return a bool instead?

derekbruening:
I'm adding this test, but it seems it may fail on actual SVE hardware.  I'll leave updating it as future work as I'm not sure how to detect the SVE hardware.  I think the set routine should return a bool.

derekbruening:
Re-assigning to get the test to work on SVE hardware.  My suggestion is to have the set routine return a bool as well.

PR #6576, opened 2024-01-23
i#6575: dr_{get,set}_sve_vector_length missing export specifiers

derekbruening:
dr_set_sve_vector_length() and dr_get_sve_vector_length() in encode_api.h were missing DR_API specifiers and thus were not exported from libdynamorio, causing link errors.

Adds a test to ir_aarch64.c which fails without the fix.  However, the test may fail on actual SVE hardware: updating it is future work, maybe by changing dr_set_sve_vector_length() to return a bool.

Also tested on the same internal build where the failure was observed.

Issue: #6575


derekbruening:
x64 failure is client.detach #6536

ivankyluk (approved):

Issue #6573, opened 2024-01-22
Remove the de-duplication logic of record filter


huyangh-430:
Current de-dup logic remove the consecutive timestamp and CPU_ID. It causes several new failures listed below. This problem is not triggered in a legacy trace.

Failed checks:

EXPECT_EQ(result_proto.record_filter_results().input_entry_count(), \
              result_proto.record_filter_results().output_entry_count());

EXPECT_THAT(new_counts, EqualsProto(original_counts)); \
modified: scheduling_marker_count: 22412 -> 19426

PR #6577, opened 2024-01-23
i#6573: Remove the de-duplication logic of record filter 

huyangh-430:
Remove the de-duplication logic for consecutive timestamp and cpu markers. Fix the problem that output entry count is not equal to the input entry count after run a null record filter test. 

We added many more timestamps to drmemtrace traces beyond the original ones which were only at the start of a trace buffer. Additionally, zip-format traces have duplicated timestamps at the start of each chunk. This means that we sometimes have deliberate back-to-back timestamp;cpu pairs in existing traces, and we have updated trace consumers to be aware of this. This has caused us to revisit the removal of back-to-back pairs in the record filter and we have decided to just eliminate that feature: we're ok if a filter removing intervening records results in consecutive timestamp;cpuid pairs that were not consecutive before.

Fixes: #6573


clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -226,21 +226,6 @@ record_filter_t::parallel_shard_memref(void *shard_data, const trace_entry_t &in
...
-            // optimization. We should ensure that we output it at least once. We
-            // skip handling this corner case for now.
-            per_shard->last_delayed_unit_header.clear();
-            ANNOTATE_FALLTHROUGH;
-        case TRACE_MARKER_TYPE_WINDOW_ID:
-        case TRACE_MARKER_TYPE_CPU_ID:
-            // Optimize space by outputting the unit header only if we are outputting
-            // something from that unit.
-            if (output)
-                per_shard->last_delayed_unit_header.push_back(entry);
```
derekbruening, line 242: I think all uses of `last_delayed_unit_header` can be removed as it only exists to avoid duplicate unit headers.  Let's bring @abhinav92003 in on this as he wrote this code.

abhinav92003: Right. It can be removed now.

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -226,10 +226,10 @@ test_cache_and_type_filter()
         // is output from this unit.
```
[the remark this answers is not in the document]
huyangh-430: Add the information you put in issue #6573.

clients/drcachesim/tools/filter/record_filter.cpp
derekbruening: I don't see the other references being removed?

abhinav92003: @huyangh-430 Please remove the other refs to last_delayed_unit_header; e.g., see L264 below which will always be false now.

huyangh-430: Removed code about last_delayed_unit_header, L264-268 in record_filter.cpp

derekbruening: Please remove all instances: search for it.  It's still in the header.

huyangh-430: Removed another instance from record_filter.h.

derekbruening (approved):
Note that it is best to give relevant, descriptive commit messages for the update commits: all your commits have the same description as the top level.  In the future describe what is actually in the separate commit.

PR #6578, opened 2024-01-23
Export drlibc interface functions

derekbruening:
Exports dynamorio_syscall, dr_fpu_exception_init, get_mmx_val, and unexpected_return in drlibc.  This solves linker issues when the drlibc and drstring libraries are linked in other build systems. Unfortunately it is not simple to make a test here; manually tested in a separate build system


derekbruening:
It turns out we don't actually want this change.

PR #6579, opened 2024-01-24
i#6536: eliminate the instability of detachment on x86-64 platform

onroadmuwl:
Use _sigreturn_ to set nudged thread‘s context on x86-64 platform instead of jumping to the modified PC address directly.
Fixes: https://github.com/DynamoRIO/dynamorio/issues/6536

### DIFF
--- a/core/unix/signal.c
+++ b/core/unix/signal.c
@@ -3169,7 +3169,7 @@ void
 thread_set_self_context(void *cxt, bool is_detach_external)
 {
 #ifdef X86
-    if (!INTERNAL_OPTION(use_sigreturn_setcontext) || is_detach_external) {
+    if (!INTERNAL_OPTION(use_sigreturn_setcontext)) {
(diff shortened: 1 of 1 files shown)


derekbruening:
I re-ran the x86-64 job and the detach test still fails:
```
185/486 Test #185: code_api|client.detach_test ......................................***Failed  Required regular expression not found. Regex=[^starting
thank you for testing detach
thread init
detach
done
$
]  0.27 sec
CMake Error at /home/runner/work/dynamorio/dynamorio/suite/tests/runall.cmake:130 (message):
  *** kill failed (1): kill: (27149): No such process
  ***
```

I also took this patch and ran it locally and it fails there too in the same way.

abhinav92003:
I see more changes were made since Derek's last comment. When this PR is ready for a re-review, please use the Github UI to signal that (see list of reviewers on top-right side of the page).

PR #6583, opened 2024-01-25
Record how to mark a test as flaky

xdje42:
Adds instructions for how to mark flaky tests.
Tested by viewing $build/api/docs/embed/html/page_test_suite.html.


api/docs/test_suite.dox
```diff
@@ -315,6 +315,15 @@ The comments at the top of runsuite_ssh.cmake describe additional options.
```
derekbruening, line 315: > iX: Record how to mark a test as flaky

Repeating the same comment from your first PR: we do not put "IX: " in titles; there is no prefix if there is no issue.

api/docs/test_suite.dox
```diff
@@ -315,6 +315,15 @@ The comments at the top of runsuite_ssh.cmake describe additional options.
 
 Unfortunately our test suite is not as clean as it could be.  Some tests can be flaky and while they pass on the machines of the existing developers and on our automated test machines, they may fail occasionally on a new machine.  Please search the issue tracker before filing a new issue to see if a test failure has already been seen once before.  We welcome contributions to fix flaky tests.
 
+Flaky tests are marked in one of two ways:
+
+ - Append "_FLAKY" to the test's name
+    - See suite/tests/CMakeLists.txt for examples
```
derekbruening, line 321: Let's make these links to the online file viewer, here and below.

api/docs/test_suite.dox
```diff
@@ -315,6 +315,15 @@ The comments at the top of runsuite_ssh.cmake describe additional options.
 
 Unfortunately our test suite is not as clean as it could be.  Some tests can be flaky and while they pass on the machines of the existing developers and on our automated test machines, they may fail occasionally on a new machine.  Please search the issue tracker before filing a new issue to see if a test failure has already been seen once before.  We welcome contributions to fix flaky tests.
 
+Flaky tests are marked in one of two ways:
```
derekbruening, line 318: If we're updating docs about flaky tests, let's talk about the SKIP_FLAKY_TESTS option here.

derekbruening, line 318: Let's update the runsuite.cmake comments (line 141) about _FLAKY as well as they are stale.

api/docs/test_suite.dox
```diff
@@ -315,6 +315,15 @@ The comments at the top of runsuite_ssh.cmake describe additional options.
...
+
+ - Append "_FLAKY" to the test's name
+    - See suite/tests/CMakeLists.txt for examples
+ - Mention the test in runsuite_wrapper.pl
```
derekbruening, line 322: Re-iterate that an issue should always be filed covering fixing the test, and its issue number should be referenced when added here.

api/docs/test_suite.dox
xdje42: I propose doing this in a separate patch as it's orthogonal to marking tests.

xdje42: What's the right wording to use here?

The code refers to #2974 but that seems to be just a general bug for recording the addition of aarch64 support.

Here's the best I can come up with:
```
Skip tests marked _FLAKY. This is one of the ways that flaky tests are marked as such
(the other way, for higher priority flakes to fix, is to mention them in runsuite_wrapper.pl).
We need CDash to ignore flakes and avoid going red and sending emails,
and we rely on our CI for a history of _FLAKY results.
```

api/docs/test_suite.dox
xdje42: Done

api/docs/test_suite.dox
xdje42: Done

api/docs/test_suite.dox
xdje42: Done, with separate PR #6586 to codify the rule.

api/docs/test_suite.dox
derekbruening: We can remove the CDash sentence as we no longer use CDash.

derekbruening (approved):

api/docs/test_suite.dox
xdje42: Done

PR #6586, opened 2024-01-25
Clarify commit summary text for non-issue related commits

xdje42:
Some non-issue related commits prefix the summary with "iX:". Thus if one is looking at the output of "git log" to find the required format one will come to the wrong conclusion. Make it clear in the docs what the rule is.

Also reformat the text to not be one long line.
One could do this in a separate patch, but this is just documentation.

### DIFF
--- a/api/docs/code_reviews.dox
+++ b/api/docs/code_reviews.dox
@@ -79,7 +79,16 @@ can select a particular reviewer if desired.
 
 \section sec_commit_messages Commit Messages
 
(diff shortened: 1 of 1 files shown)


api/docs/code_reviews.dox
```diff
@@ -79,7 +79,16 @@ can select a particular reviewer if desired.
```
derekbruening, line 79: > Some non-issue related commits prefix the summary with "iX:"

I was wondering where you had seen this: in fact there are 2 (and no more: exactly 2 out of 6K commits) that accidentally did this as new contributors.

I would re-word this line to say "There are 2 cases of commits that incorrectly prefix..." as otherwise this sounds like there are many or that this used to be the preferred policy or something.

I guess if you explicitly search for "iX:" and stop at the first match and never look at any other examples you might mistakenly conclude this is used extensively?  If you just look at the last 100 commits you see many examples with no prefix and zero examples of "iX:".

derekbruening (approved):

api/docs/code_reviews.dox
xdje42: Done.

PR #6588, opened 2024-01-25
i#6452,i#6536: Disable flaky attach_test, disable_test for unix x86-64

xdje42:
Adds two tests to the x86-64 unix ignore list:

i#6452 code_api|client.attach_test
i#6536 code_api|client.detach_test

Issue: #6452, #6536

### DIFF
--- a/suite/runsuite_wrapper.pl
+++ b/suite/runsuite_wrapper.pl
@@ -389,6 +389,8 @@ for (my $i = 0; $i <= $#lines; ++$i) {
                 );
             %ignore_failures_64 = (
                 'code_api|tool.drcacheoff.burst_threadfilter' => 1, # i#2941
+                'code_api|client.attach_test' => 1, # i#6452
+                'code_api|client.detach_test' => 1, # i#6536
                 # These are from the long suite.
                 'code_api,opt_memory|common.loglevel' => 1, # i#1807
                 'code_api,opt_speed|common.decode-stress' => 1, # i#1807


suite/runsuite_wrapper.pl
```diff
@@ -389,6 +389,8 @@
                 );
```
abhinav92003, line 389: Would be good to also mention x86-64 in the PR description and title.

abhinav92003 (approved):

suite/runsuite_wrapper.pl
xdje42: These tests have also been seen to fail on aarch64 though.
Aren't references to the issues enough?

abhinav92003: Well, since this PR is disabling them for a specific suite, it makes sense to mention the suite in the description.

derekbruening: Could you point to an aarch64 failure?  Prior to the ptrace fix they failed due to permissions; after that fix they have never failed AFAIK.  I logged into that test machine and ran each test 200x in a row w/ no failures manually as I noted in the issue/PR.

derekbruening: > Adds two tests to the unix ignore list:

I'm agreeing with @abhinav92003  -- there is no ignore list for all of "unix": this PR is adding these only to the x86_64 Linux list.

xdje42: Done

PR #6594, opened 2024-01-26
i#6514 null SP to clone(), part 2: Have linux clone test use dynamori…

xdje42:
…o_clone

Now that dynamorio_clone() is in drlibc we can use it in the linux clone.c test.

Issue #6514

### DIFF
--- a/suite/tests/linux/clone.c
+++ b/suite/tests/linux/clone.c
@@ -64,10 +64,6 @@ typedef unsigned long ulong;
  */
 #define CLONE3_SYSCALL_NUM 435
 
-/* i#762: Hard to get clone() from sched.h, so copy prototype. */
-extern int
-clone(int (*fn)(void *arg), void *child_stack, int flags, void *arg, ...);
-
 #define THREAD_STACK_SIZE (32 * 1024)
 
 #ifdef X64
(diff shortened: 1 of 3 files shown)


suite/tests/linux/clone.c
```diff
@@ -113,10 +108,10 @@ test_thread(bool share_sighand, bool clone_vm, bool use_clone3)
         /* If SYS_clone3 is not defined, we simply use SYS_clone instead, so that
          * the expected output is the same in both cases.
          */
-        child = create_thread(run, NULL, &stack, share_sighand, clone_vm);
+        child = create_thread(run_with_exit, &stack, share_sighand, clone_vm);
```
derekbruening, line 111: The raw clone syscall takes no function; the glibc wrapper takes a function and appends an exit syscall to it.  Looks like dynamorio_clone is in between, taking a function but requiring the user to have it end with exit?  Should we update the comments for dynamorio_clone and create_thread here to make this clear?

derekbruening (approved):

suite/tests/linux/clone.c
xdje42: Add a comment to dynamorio_clone. I think create_thread can live without one.

Issue #6580, opened 2024-01-24
Build broken by python error in aarch64/codec.py: No attribute 'split' on Tuple


derekbruening:
Our internal build, once past the breakage from #6575, is still broken when trying to pull in the latest Github sources, on this error:
```
File "core/ir/aarch64/codec.py", line 73, in opnd_stem: No attribute 'split' on Tuple[List[str], List[str]] [attribute-error]
  In Union[Any, Tuple[List[str], List[str]], str]
Called from (traceback):
  line 782, in current file
  line 758, in main
  line 393, in generate_encoder

For more details, see [http://go/pytype-errors#attribute-error](https://www.google.com/url?q=http://go/pytype-errors%23attribute-error&sa=D)
```

joshua-warburton:
I've had a quick look into this. I would expect this to fail if presented with a tuple, but it seems that the line it's failing on expects a string and would fail if it got a tuple even before the change. I've followed the code through and don't see anything in the patch itself that would cause this to fail as by this point we've re-written the opndsets with the function opndset_naming to remove the lists.

I can try to reproduce this if you have the python version you're using and which ref you're trying to build

joshua-warburton:
Detail:
The build is failing because we attempt to run `.split()` on a tuple, `pattern.opndset` via the `opnd_stem()` function. However, if we remove that function we're attempting to encode a tuple as a string, which will fail still with:
```
>>> t = "%s"
>>> a = ("a", "b", "c")
>>> t %a
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: not all arguments converted during string formatting
```

Tracing back, `pattern.opndset` is initially populated by either the highlighted arguments below
```
patterns.append(Pattern(pattern, opcode_bits, opnd_bits, high_soft_bits, opcode, ***(dsts, srcs)***, enum, feat))
```
```
patterns.append(Pattern(pattern, opcode_bits, opnd_bits, high_soft_bits, opcode, ***opndset***, enum, feat))
```

However, this is later flattened in the function `opndset_naming` on line 742. It takes both the string and non-string variants and rewrites these as something like "gen_04205000_001f07ff" and guarantees it will be a string from there forward.

Clearly, something is getting past this check but I'm not sure how and the patch mentioned above hasn't changed any of this.

derekbruening:
[...]

PR #6595, opened 2024-01-26
i#6580: AArch64 codec.py: Remove variable reuse in function naming flow

joshua-warburton:
This patch refactors how the script gets hold of the name of generated functions. Previously there was a pass in opndset_naming() that overwrote the value of the opndset attribute. This was causing issues with type checking because this changed the type of an opndset from a tuple or string to always a string which was flagged as an error when .split() was used on it.

As part of this change the fall-through code became completely dead code and has been removed. It was obsoleted originally by the addition of + masks for the operands.

It was also revealed by this change that several encodings in the codecs were no longer used and represented less correct versions of other encodings. They have been removed.

issue: #6580


AssadHashmi (approved):

derekbruening (commented):
(Going to defer to Assad's review as he is more familiar with this code.)

Issue #6596, opened 2024-01-26
linux.clone* tests fail when run in a container under Ubuntu 22.04


philramsey-arm:
The linux.clone* tests expect clone3 to be available if SYS_clone3 is defined.

However in some scenarios SYS_clone3 is defined but clone3 returns ENOSYS e.g. when running in a container under Ubuntu 22.04
(see https://github.com/moby/moby/pull/42681).

The linux.clone* tests should not fail if clone3 returns ENOSYS.

**To Reproduce**
# On Ubuntu 22.04:
docker run --rm -it --name dynamorio ubuntu/base:22.04 bash

git clone https://github.com/DynamoRIO/dynamorio.git && cd dynamorio
git submodule init
git submodule update

# build tests
cd ..
mkdir dynamorio_build
cd dynamorio_build
cmake -DDEBUG=ON -DBUILD_TESTS=ON ../dynamorio
make -j

# run tests
ctest --output-on-failure --timeout 120

PR #6597, opened 2024-01-26
i#6596: linux.clone* tests fail when run in a container on Ubuntu 22.04

philramsey-arm:
Set a flag if clone3() sets errno to ENOSYS, so clone() is used instead.

issue: #6596


derekbruening:
Re-routing to @xdje42 who has been touching the clone code.

suite/tests/linux/clone.c
```diff
@@ -106,24 +109,39 @@ test_thread(bool share_sighand, bool clone_vm, bool use_clone3)
...
 main()
 {
+    /* Try using clone3 when it is possibly not defined. */
+    int ret_failure_clone3 = make_clone3_syscall(NULL, 0, NULL);
+    assert(ret_failure_clone3 == -1);
+#ifdef SYS_clone3
+    /* In some scenarios SYS_clone3 is defined but clone3 returns ENOSYS
+    * e.g. when running in a container under Ubuntu 22.04
+    * see https://github.com/moby/moby/pull/42681
+    */
```
xdje42, line 136: Indention of '*' is off by one here.

Also, add reference to issue# here.

suite/tests/linux/clone.c
```diff
@@ -106,24 +109,39 @@ test_thread(bool share_sighand, bool clone_vm, bool use_clone3)
...
+    if (errno != ENOSYS)
+        clone3_available = true;
```
xdje42, line 138: IIUC, we're only setting clone_available to true in the case where SYS_clone3 is defined.
The code can handle clone3 being supported even if SYS_clone3 is not defined, do we want to handle that here?

xdje42 (changes requested):
Approved with requested changes.

suite/tests/linux/clone.c
philramsey-arm: Have removed the `#ifdef SYS_clone3` so that `clone3` can be tested when SYS_clone3 is not defined.

suite/tests/linux/clone.c
```diff
@@ -101,24 +103,36 @@ test_thread(bool share_sighand, bool clone_vm, bool use_clone3)
...
-#else
-        /* If SYS_clone3 is not defined, we simply use SYS_clone instead, so that
-         * the expected output is the same in both cases.
-         */
-        child = create_thread(run_with_exit, &stack, share_sighand, clone_vm);
-#endif
-    } else
-        child = create_thread(run_with_exit, &stack, share_sighand, clone_vm);
+
+    pid_t (*create_thread_func)(void (*fcn)(void), void **stack, bool share_sighand,
```
xdje42, line 107: I'd preserve the comment that we use clone instead of clone3 when clone3 is requested and not available so that the expected output is the same in both cases. Maybe something like:

    /* Use create_thread when clone3 is asked for and not available so that
     * the expected output is the same in all cases.
     */

xdje42 (approved):
Approved with requested change.

suite/tests/linux/clone.c
```diff
@@ -298,7 +297,6 @@ make_clone3_syscall(void *clone_args, ulong clone_args_size, void (*fcn)(void))
     return result;
 }
 
-#ifdef SYS_clone3
```
xdje42, line 301: Hmmm. The removal of the #ifdef breaks the android build. Replace with #ifndef ANDROID?
[here and at the call site]
I don't know enough about android support to be confident in what the right fix is though.

xdje42 (changes requested):

suite/tests/linux/clone.c
philramsey-arm: Done

xdje42 (approved):

AssadHashmi (approved):

PR #6598, opened 2024-01-26
I6593 zipfile reading support

huyangh-430:
The record filter does not support filtering .zip traces today. Since we have moved to .zip as the default format, we should add that support.

### DIFF
--- a/clients/drcachesim/reader/zipfile_file_reader.cpp
+++ b/clients/drcachesim/reader/zipfile_file_reader.cpp
@@ -226,6 +226,21 @@ file_reader_t<zipfile_reader_t>::skip_instructions(uint64_t instruction_count)
  */
 
 template <>
+record_file_reader_t<zipfile_reader_t>::record_file_reader_t()
+{
+    input_file_->file = nullptr;
+}
+
+template <>
+record_file_reader_t<zipfile_reader_t>::~record_file_reader_t<zipfile_reader_t>()
+{
+    if (input_file_->file != nullptr) {
+        unzClose(input_file_->file);
+        input_file_->file = nullptr;
+    }
+}
+
+template <>
 bool
 record_file_reader_t<zipfile_reader_t>::open_single_file(const std::string &path)
 {
--- a/clients/drcachesim/reader/zipfile_file_reader.cpp
+++ b/clients/drcachesim/reader/zipfile_file_reader.cpp
@@ -40,6 +40,27 @@ namespace drmemtrace {
 // sources.  The docs are the header files:
 // https://github.com/madler/zlib/blob/master/contrib/minizip/unzip.h
 
+/**************************************************************************
+ * Common logic used in the zipfile_reader_t specializations for file_reader_t
+ * and record_file_reader_t.
+ */
+
+bool
+open_single_file_common(const std::string &path, zipfile_reader_t &zread)
+{
+    unzFile file = unzOpen(path.c_str());
+    if (file == nullptr)
+        return false;
+    zread = zipfile_reader_t(file, path);
+    if (unzGoToFirstFile(file) != UNZ_OK || unzOpenCurrentFile(file) != UNZ_OK)
+        return false;
+    return true;
+}
+
+/**************************************************
+ * zipfile_reader_t specializations for file_reader_t.
+ */
+
 /* clang-format off */ /* (make vera++ newline-after-type check happy) */
 template <>
 /* clang-format on */
@@ -63,12 +84,9 @@ template <>
 bool
 file_reader_t<zipfile_reader_t>::open_single_file(const std::string &path)
 {
-    unzFile file = unzOpen(path.c_str());
-    if (file == nullptr)
-        return false;
-    input_file_ = zipfile_reader_t(file, path);
-    if (unzGoToFirstFile(file) != UNZ_OK || unzOpenCurrentFile(file) != UNZ_OK)
-        return false;
+    zipfile_reader_t zread;
+    if (!open_single_file_common(path,zread)) return false;
+    input_file_ = std::move(zread);
     VPRINT(this, 1, "Opened input file %s\n", path.c_str());
     return true;
 }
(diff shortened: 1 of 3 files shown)


derekbruening:
Not understanding why caching_device_stats.h is hitting the VS problems with `max` we've seen elsewhere when this PR doesn't touch it or AFAICT any headers it includes...

derekbruening:
The win64 failure is my new test which uses the record_filter_launcher.  The record_filter_launcher fails up front on Windows: doesn't parse its args properly.  I assume it was never run there?  There was no automated test before.  I tracked it down to missing a _UNICODE define.  Will just fix it as part of this PR.

derekbruening:
Now the failure is an invariant on the filtered trace:
```
273:   *** postcmd failed (3): Trace invariant failure in T3104 at ref # 99016 (0
273:   instrs since timestamp 13351308211805717): An unfiltered thread should have
273:   at least 1 instruction
```

It looks like a general bug though that could happen on any trace: or at least my theory is it's from a non-fetched instr right before thread exit, as this check demands a regular instr.  This test does an early exit on Windows to speed it up, which makes that more likely.  I'll see if that is the fix here; then will make a separate issue and PR for that fix if so.

derekbruening:
Decided to separate the launcher fix: filed #6620 with PR #6621; will merge that before this.

derekbruening:
Also made a separate PR #6623 for the invariant fix.

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -344,13 +356,15 @@ test_cache_and_type_filter()
 
         // Process each trace entry.
         for (int i = 0; i < static_cast<int>(entries.size()); ++i) {
+            stream->set_last_timestamp(entries[i].entry.addr);
```
abhinav92003, line 359: This probably needs to be removed.

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -396,7 +410,141 @@ test_cache_and_type_filter()
...
+            fprintf(stderr, "Too few entries in filtered output. Expected: ");
+            print_entry(entries[i].entry);
+            fprintf(stderr, "\n");
+            return false;
+        }
+        if (entries[i].entry.type == TRACE_TYPE_MARKER &&
+            entries[i].entry.size == TRACE_MARKER_TYPE_RECORD_ORDINAL) {
+            if (filtered[j].type != entries[i].entry.type ||
+                filtered[j].size != entries[i].entry.size ||
+                filtered[j].addr != entries[i].entry.addr - REMOVED_VISIBLE_COUNT) {
```
abhinav92003, line 512: The other test adds entries for both the input and output stream in the std::vector<struct test_case>, and uses a bool to determine whether or not to use the entry as input to the tool. IMHO it makes it easier to read the expected output (replacement entries are right next to the input entries), and also avoids all this special casing. Could optionally use that strategy in this test too.

clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -105,35 +109,62 @@ record_filter_t::parallel_shard_supported()
...
                                             memtrace_stream_t *shard_stream)
 {
     auto per_shard = new per_shard_t;
-    per_shard->writer = get_writer(per_shard, shard_stream);
+    std::string error = get_writer(per_shard, shard_stream);
+    if (!error.empty()) {
+        per_shard->error = "Failure in opening writer: " + error;
+        success_ = false;
+    }
+    if (!per_shard->writer && !per_shard->archive_writer) {
```
abhinav92003, line 160: We probably want to check only per_shard->writer here.

clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -160,7 +191,8 @@ record_filter_t::parallel_shard_exit(void *shard_data)
     // Destroy the writer since we do not need it anymore. This also makes sure
     // that data is written out to the file; curiously, a simple flush doesn't
     // do it.
-    per_shard->writer.reset(nullptr);
+    per_shard->file_writer.reset(nullptr);
+    per_shard->archive_writer.reset(nullptr);
```
abhinav92003, line 195: Also good to set per_shard->writer to nullptr.

clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -105,35 +109,62 @@ record_filter_t::parallel_shard_supported()
...
     auto per_shard = new per_shard_t;
-    per_shard->writer = get_writer(per_shard, shard_stream);
+    std::string error = get_writer(per_shard, shard_stream);
+    if (!error.empty()) {
+        per_shard->error = "Failure in opening writer: " + error;
+        success_ = false;
+    }
+    if (!per_shard->writer && !per_shard->archive_writer) {
+        per_shard->error = "Could not open a writer for " + per_shard->output_path;
+        success_ = false;
```
abhinav92003, line 162: This is fatal right? Return early?

abhinav92003 (approved):

clients/drcachesim/reader/zipfile_file_reader.cpp
```diff
@@ -40,6 +40,96 @@ namespace drmemtrace {
...
 
+/**************************************************************************
+ * Common logic used in the zipfile_reader_t specializations for file_reader_t
+ * and record_file_reader_t.
+ */
+
+namespace {
+
+#ifdef DEBUG
+#    define ZPRINT(verbosity, level, ...)     \
```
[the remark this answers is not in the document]
derekbruening: Added clarifying comments:
```
// We use the VPRINT from reader.h for member function code.
// For common routines we need a separate variant taking verbosity in directly.
```

clients/drcachesim/tests/record_filter_unit_tests.cpp
derekbruening: Hmm, weird.  Removed.

clients/drcachesim/tools/filter/record_filter.cpp
derekbruening: Done.

clients/drcachesim/tools/filter/record_filter.cpp
derekbruening: Done.

clients/drcachesim/tests/record_filter_unit_tests.cpp
derekbruening: When I first looked at it, it looked like I couldn't re-use much.  But on a second look: I was able to make a helper and share all the same checking code, changing the new test to use the combined form.  Done.

Issue #100, opened 2014-11-27
docs build fails using cygwin mogrify with certain drive letters


derekbruening:
_From [derek.br...@gmail.com](https://code.google.com/u/117968039472581148324/) on March 30, 2009 22:09:35_

cygwin mogrify mis-parses mixed paths when the drive is a, b,
c, or g (haven't tracked down why those particular letters):

  &#37; mogrify -resize 33&#37; -format png c:/client.tiff[0]
  mogrify: Must specify image size `/client.tiff'.

I didn't notice this in any initial testing when building the docs b/c my
windows test machine uses d: and e:.

Haven't found a workaround other than providing a cygwin path.

_Original issue: http://code.google.com/p/dynamorio/issues/detail?id=100_

Issue #35, opened 2014-11-27
build: gcc 4.1.2 warnings with new -O


derekbruening:
_From [derek.br...@gmail.com](https://code.google.com/u/117968039472581148324/) on February 24, 2009 09:43:39_

after r20 swapped debug build to -O, now with gcc 4.1.2 we have warnings on
all strings that are longer than the 50-char STAT_NAME_MAX_LEN:
In file included from dynamo.c:320: 
./lib/statsx.h: In function 'statistics_init': 
./lib/statsx.h:180: warning: value computed is not used 
./lib/statsx.h:181: warning: value computed is not used 
...

_Original issue: http://code.google.com/p/dynamorio/issues/detail?id=35_

Issue #34, opened 2014-11-27
REG_* enum name conflict with sys/ucontext.h


derekbruening:
_From [derek.br...@gmail.com](https://code.google.com/u/117968039472581148324/) on February 24, 2009 09:39:38_

this was PR 371339

if we build a client that includes ucontext.h, usually via signal.h, with
g++ we hit a name conflict:

> g++ -m32 -g3 -DLINUX -DSHOW_RESULTS -DX86_32 -fPIC -shared -nostartfiles
-I/work/opensource/dr/exports/include signal.c -o signal.so
In file included from /usr/include/signal.h:353,
                 from signal.c:40:
/usr/include/sys/ucontext.h:168: error: conflicting declaration ‘REG_EDI’
/work/opensource/dr/exports/include/dr_ir_opnd.h:73: error: ‘REG_EDI’ has a
previous declaration as ‘<anonymous enum> REG_EDI’
/usr/include/sys/ucontext.h:170: error: conflicting declaration ‘REG_ESI’
/work/opensource/dr/exports/include/dr_ir_opnd.h:72: error: ‘REG_ESI’ has a
previous declaration as ‘<anonymous enum> REG_ESI’
/usr/include/sys/ucontext.h:172: error: conflicting declaration ‘REG_EBP’
/work/opensource/dr/exports/include/dr_ir_opnd.h:71: error: ‘REG_EBP’ has a
previous declaration as ‘<anonymous enum> REG_EBP’
/usr/include/sys/ucontext.h:174: error: conflicting declaration ‘REG_ESP’
/work/opensource/dr/exports/include/dr_ir_opnd.h:70: error: ‘REG_ESP’ has a
previous declaration as ‘<anonymous enum> REG_ESP’
/usr/include/sys/ucontext.h:176: error: conflicting declaration ‘REG_EBX’
/work/opensource/dr/exports/include/dr_ir_opnd.h:69: error: ‘REG_EBX’ has a
previous declaration as ‘<anonymous enum> REG_EBX’
/usr/include/sys/ucontext.h:178: error: conflicting declaration ‘REG_EDX’
/work/opensource/dr/exports/include/dr_ir_opnd.h:68: error: ‘REG_EDX’ has a
previous declaration as ‘<anonymous enum> REG_EDX’
/usr/include/sys/ucontext.h:180: error: conflicting declaration ‘REG_ECX’
[...]

PR #6601, opened 2024-01-27
i#6514 null SP to clone(), part 3: Handle null SP passed to clone

### LINKED ISSUE #6514: APP CRASH: Missing handling of stack ptr == NULL in clone(CLONE_VFORK + CLONE_VM)

xdje42:
The kernel supports passing NULL for SP to clone: It means to use the SP of the parent. Supporting this requires augmenting dynamorio_clone to handle SP==NULL, and augmenting create_clone_record to handle SP==NULL.

Only tested X86 in this pass. Other arches deferred.

With 32-bit testing temporarily downsized, this was manually tested on i386:
$ ctest -R clone
Test project /home/dje/upstream/i6514/build32
    Start  34: code_api|linux.clone
1/3 Test  #34: code_api|linux.clone .............   Passed    1.00 sec
    Start  35: code_api|linux.clone-pie
2/3 Test  #35: code_api|linux.clone-pie .........   Passed    0.97 sec
    Start 100: code_api|linux.clone-reset
3/3 Test #100: code_api|linux.clone-reset .......   Passed    0.97 sec
100% tests passed, 0 tests failed out of 3
Total Test time (real) =   3.03 sec

Issue #6514


core/drlibc/drlibc_x86.asm
```diff
@@ -762,46 +762,72 @@ GLOBAL_LABEL(load_dynamo_failure:)
...
-        mov      REG_XSI, DWORD [7*ARG_SZ + REG_XSP] /* ARG4 + 3 pushes */
-        mov      REG_XDI, DWORD [8*ARG_SZ + REG_XSP] /* ARG5 + 3 pushes */
+        /* Save space for func for newsp == NULL case. */
+        sub      REG_XSP, ARG_SZ
+        /* i#6514: Save func on the child's stack. Remember that if newsp is
+         * NULL then the child's stack is our stack.
+         */
+        and      REG_XCX, HEX(fffffff0) /* Ensure child xsp is aligned. */
+        jnz      stack_not_null
+        lea      REG_XCX, [ARG_SZ + REG_XSP] /* Add ARG_SZ back for the sub. */
```
derekbruening, line 800: Line 793 says you did the SUB there for this ecx==0 case, and here it says you're undoing a SUB which at first seems to be the line 794 SUB and so it all seems like nops and is confusing.  Really this comment refers to the line 802 SUB, right?  Please add "the sub *below*" or something.

core/drlibc/drlibc_x86.asm
```diff
@@ -762,46 +762,72 @@ GLOBAL_LABEL(load_dynamo_failure:)
...
-         * when using clone_record_t we have 4 slots we can clobber.
+        /* Save func for use post-syscall on the newsp.
+         * This is tricky because we have to handle the case of newsp == NULL.
          */
 # ifdef X64
-        sub      ARG2, ARG_SZ
-        mov      [ARG2], ARG6 /* func is now on TOS of newsp */
-        /* all args are already in syscall registers */
+        /* The syscall preserves all registers except rax, rcx, r11. */
+        sub      REG_XSP, FRAME_ALIGNMENT
```
derekbruening, line 775: Better to align the stack instead of mis-align it: see below where the mis-alignment here breaks the ABI for newsp==NULL.  For newsp!=NULL it is also safer to align instead, for possible future changes to this code.

core/drlibc/drlibc_x86.asm
```diff
@@ -762,46 +762,72 @@ GLOBAL_LABEL(load_dynamo_failure:)
...
+        /* Preserve callee-saved regs. */
         push     REG_XBX
         push     REG_XSI
         push     REG_XDI
-        /* now can't use ARG* since xsp modified by pushes */
-        mov      REG_XBX, DWORD [4*ARG_SZ + REG_XSP] /* ARG1 + 3 pushes */
-        mov      REG_XSI, DWORD [7*ARG_SZ + REG_XSP] /* ARG4 + 3 pushes */
-        mov      REG_XDI, DWORD [8*ARG_SZ + REG_XSP] /* ARG5 + 3 pushes */
+        /* Save space for func for newsp == NULL case. */
+        sub      REG_XSP, ARG_SZ
```
derekbruening, line 794: Similar to 64-bit case: now the stack is mis-aligned.  Add this to the comment so it's clear we're aware of it and future changes to the code don't assume esp is aligned mid-function (which is a reasonable assumption, normally).

core/drlibc/drlibc_x86.asm
```diff
@@ -762,46 +762,72 @@ GLOBAL_LABEL(load_dynamo_failure:)
...
         int      HEX(80)
 # endif
         cmp      REG_XAX, 0
         jne      dynamorio_clone_parent
+# ifdef X64
+        call     r15
+# else
         pop      REG_XCX
+        /* Ensure child xsp is aligned again, for newsp == NULL case. */
+        and      REG_XSP, HEX(fffffff0)
```
derekbruening, line 819: But for newsp==NULL we know it's aligned, if esp it were aligned on entry, so why do we need this?

core/drlibc/drlibc_x86.asm
```diff
@@ -762,46 +762,72 @@ GLOBAL_LABEL(load_dynamo_failure:)
...
         /* shouldn't return */
         jmp      GLOBAL_REF(unexpected_return)
 dynamorio_clone_parent:
-# ifndef X64
+# ifdef X64
+        mov      r15, [REG_XSP]
+        add      REG_XSP, FRAME_ALIGNMENT
+# else
         /* restore callee-saved regs */
+        add      REG_XSP, ARG_SZ
```
derekbruening, line 830: I ripped out most 32-bit tests in PR #6590.  Please either add the UBUNTU_22 label to the clone tests so they're run, or describe manual 32-bit testing done in the PR description.

core/drlibc/drlibc.h
```diff
@@ -72,7 +72,11 @@ dynamorio_mach_syscall(uint sysnum, uint num_args, ...);
 #    else
 ptr_int_t
 dynamorio_syscall(uint sysnum, uint num_args, ...);
-/* N.B. func must not return. */
+/* Wrapper for clone().
+ * N.B. func must not return.
+ * On x86 (32 and 64-bit) this supports passing NULL for newsp which is just
+ * shorthand for the child using the same value as the parent.
```
derekbruening, line 78: Please add TODO or XXX #6514 comments to the other dynamorio_clone routines on other arches.

core/unix/signal.c
```diff
@@ -817,12 +817,18 @@ create_clone_record(dcontext_t *dcontext, reg_t *app_thread_xsp)
              * cl_args->stack. But we expect the highest (non-inclusive)
              * in the clone record's app_thread_xsp.
              */
-            record->app_thread_xsp = dr_clone_args->stack + dr_clone_args->stack_size;
+            if (dr_clone_args->stack == 0)
+                record->app_thread_xsp = get_mcontext(dcontext)->xsp;
```
derekbruening, line 821: It looks like all arches are supported here: the PR description about only x86 applies only to the *test* of this, right?  So it should be updated.

suite/tests/linux/clone.c
```diff
@@ -116,6 +119,40 @@ test_thread(bool share_sighand, bool clone_vm, bool use_clone3)
...
+void
+test_with_null_stack_pointer(bool clone_vm, bool use_clone3)
+{
+    print("%s(clone_vm %d, use_clone3 %d)\n", __FUNCTION__, clone_vm, use_clone3);
+    int flags = clone_vm ? (CLONE_VFORK | CLONE_VM) : 0;
+    int ret;
+    /* If we don't have SYS_clone3, keep expected output the same and just use SYS_clone.
+     */
+    bool really_use_clone3 = use_clone3;
+#    ifndef SYS_clone3
```
derekbruening, line 134: For core DR the presence of a SYS_ constant does not mean the syscall is supported by the underlying kernel: we have to support building on one machine and running on another.  For tests that's not nec the case, but here don't we get our SYS constants from the DR headers, which will define them no matter what the underlying kernel supports?  I believe this has to do a check for ENOSYS.  I think that's what #6596 is about, right?

suite/tests/linux/clone.c
```diff
@@ -116,6 +119,40 @@ test_thread(bool share_sighand, bool clone_vm, bool use_clone3)
...
+    really_use_clone3 = false;
+#    endif
+    if (really_use_clone3) {
+        struct clone_args cl_args = { 0 };
+        cl_args.flags = flags;
+        cl_args.exit_signal = SIGCHLD;
+        ret = make_clone3_syscall(&cl_args, sizeof(cl_args), run_with_exit);
+    } else {
+        flags = flags | SIGCHLD;
+        ret = make_clone_syscall(flags, /*stack*/ NULL, /*parent_tid*/ NULL,
```
derekbruening, line 144: style: These all need equal signs: s/stack/stack=/, etc.

suite/tests/linux/clone.c
```diff
@@ -221,7 +277,7 @@ create_thread(void (*fcn)(void), void **stack, bool share_sighand, bool clone_vm
     return newpid;
 }
 
-/* glibc does not provide a wrapper for clone3 yet. This makes it difficult
+/* drlibc does not provide a wrapper for clone3 yet. This makes it difficult
```
derekbruening, line 280: Probably this should say "Neither glibc nor drlibc provides..."

core/drlibc/drlibc_x86.asm
```diff
@@ -762,46 +762,72 @@ GLOBAL_LABEL(load_dynamo_failure:)
...
-        /* now can't use ARG* since xsp modified by pushes */
-        mov      REG_XBX, DWORD [4*ARG_SZ + REG_XSP] /* ARG1 + 3 pushes */
-        mov      REG_XSI, DWORD [7*ARG_SZ + REG_XSP] /* ARG4 + 3 pushes */
-        mov      REG_XDI, DWORD [8*ARG_SZ + REG_XSP] /* ARG5 + 3 pushes */
+        /* Save space for func for newsp == NULL case. */
+        sub      REG_XSP, ARG_SZ
+        /* i#6514: Save func on the child's stack. Remember that if newsp is
+         * NULL then the child's stack is our stack.
+         */
+        and      REG_XCX, HEX(fffffff0) /* Ensure child xsp is aligned. */
```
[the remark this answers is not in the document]
xdje42: glibc does this. One can make a case either way, but since it keeps things simple and glibc does this I like it.

suite/tests/linux/clone.c
xdje42: While you are correct that compile time presence of SYS_constant does not guarantee runtime presence of the syscall, the current patch is not trying to also solve this problem and instead sticks with what is already there.
Compare this code with the current definition of test_thread().

core/drlibc/drlibc_x86.asm
```diff
@@ -762,46 +762,72 @@ GLOBAL_LABEL(load_dynamo_failure:)
...
+        mov      REG_XSI, DWORD [8*ARG_SZ + REG_XSP] /* ARG4 + 4 pushes */
+        mov      REG_XDI, DWORD [9*ARG_SZ + REG_XSP] /* ARG5 + 4 pushes */
         mov      REG_XAX, SYS_clone
         /* PR 254280: we assume int$80 is ok even for LOL64 */
         int      HEX(80)
 # endif
         cmp      REG_XAX, 0
         jne      dynamorio_clone_parent
+# ifdef X64
+        call     r15
```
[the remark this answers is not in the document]
xdje42: Done. For some silly reason was thinking aarch64 call semantics.

core/drlibc/drlibc_x86.asm
```diff
@@ -762,46 +762,72 @@ GLOBAL_LABEL(load_dynamo_failure:)
...
+         * NULL then the child's stack is our stack.
+         */
+        and      REG_XCX, HEX(fffffff0) /* Ensure child xsp is aligned. */
+        jnz      stack_not_null
+        lea      REG_XCX, [ARG_SZ + REG_XSP] /* Add ARG_SZ back for the sub. */
+stack_not_null:
+        sub      REG_XCX, ARG_SZ
+        mov      [REG_XCX], REG_XAX /* func is now on TOS of newsp */
+        /* Now can't use ARG* since xsp modified by pushes. */
+        mov      REG_XBX, DWORD [5*ARG_SZ + REG_XSP] /* ARG1 + 4 pushes */
```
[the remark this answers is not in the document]
xdje42: Back to 3 pushes.

core/drlibc/drlibc_x86.asm
xdje42: Simplified by just always pushing func onto the parent's stack, even if newsp != NULL.

core/drlibc/drlibc.h
xdje42: Done

suite/tests/linux/clone.c
xdje42: Done

suite/tests/linux/clone.c
xdje42: Done

core/unix/signal.c
xdje42: Text updated.

core/drlibc/drlibc_x86.asm
```diff
@@ -762,46 +762,72 @@ GLOBAL_LABEL(load_dynamo_failure:)
...
+        sub      REG_XSP, ARG_SZ
+        /* i#6514: Save func on the child's stack. Remember that if newsp is
+         * NULL then the child's stack is our stack.
+         */
+        and      REG_XCX, HEX(fffffff0) /* Ensure child xsp is aligned. */
+        jnz      stack_not_null
+        lea      REG_XCX, [ARG_SZ + REG_XSP] /* Add ARG_SZ back for the sub. */
+stack_not_null:
+        sub      REG_XCX, ARG_SZ
+        mov      [REG_XCX], REG_XAX /* func is now on TOS of newsp */
```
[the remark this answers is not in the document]
xdje42: Done

core/drlibc/drlibc_x86.asm
xdje42: Done

core/drlibc/drlibc_x86.asm
xdje42: Simplified

core/drlibc/drlibc_x86.asm
xdje42: Updated

core/drlibc/drlibc_x86.asm
xdje42: Simplified

suite/tests/linux/clone.c
derekbruening: The problem is, what is already there is breaking ongoing testing efforts and is being actively changed concurrently with this PR to fix that other testing: #6596.  So this new code should match the post-#6595-fixed code to avoid regressing and breaking the container testing yet again.

derekbruening:
(Reminder to click re-request review when ready.)

suite/tests/linux/clone.c
xdje42: I'm aware of #6596. One of these will go in first and the other can be updated then.
It's trivial to update either when that happens.

derekbruening: > I'm aware of https://github.com/DynamoRIO/dynamorio/issues/6596.

Please make this clear by having a TODO comment so the reader knows this ifdef is not meant to be permanent.

> One of these will go in first and the other can be updated then.
> It's trivial to update either when that happens.

I don't know that it's that simple: my fear is that there is no automated container test in our GA CI tests, so if this PR goes in first and the author of the other just clicks on "merge" and the automated merge works, that author may have no idea that this PR has broken the container test.  It is not unusual to not re-do all manual testing after every merge from master.  So it seems there is a decent chance this PR would break what the other PR is trying to fix.  I would suggest putting a comment on the other PR so they are aware of this PR potentially breaking them.

xdje42: Fixes for #6596 have gone in and this patch has been updated with those changes.

core/drlibc/drlibc_x86.asm
```diff
@@ -762,51 +762,76 @@ GLOBAL_LABEL(load_dynamo_failure:)
...
+         * This is tricky because we have to handle the case of newsp == NULL.
          */
 # ifdef X64
-        sub      ARG2, ARG_SZ
-        mov      [ARG2], ARG6 /* func is now on TOS of newsp */
-        /* all args are already in syscall registers */
+        /* The syscall preserves all registers except rax, rcx, r11. */
+        push     r15
+        mov      r15, ARG6 /* Func is now in r15. */
+        sub      REG_XSP, 2*ARG_SZ /* Keep xsp aligned, for newsp==NULL case. */
```
derekbruening, line 777: But it's already aligned after the push on line 775.  Seems like this line can be removed.

core/drlibc/drlibc_x86.asm
```diff
@@ -762,51 +762,76 @@ GLOBAL_LABEL(load_dynamo_failure:)
...
+        and      REG_XSI, -FRAME_ALIGNMENT /* For glibc compatibility, align newsp. */
```
derekbruening, line 778: ARG2 may be more clear than REG_XSI.

core/drlibc/drlibc_x86.asm
```diff
@@ -762,51 +762,76 @@ GLOBAL_LABEL(load_dynamo_failure:)
...
         call     REG_XCX
-        /* shouldn't return */
+# endif
+        /* Shouldn't return. */
         jmp      GLOBAL_REF(unexpected_return)
 dynamorio_clone_parent:
-# ifndef X64
-        /* restore callee-saved regs */
+# ifdef X64
+        add      REG_XSP, 2*ARG_SZ
```
derekbruening, line 825: See above: this would be removed as well.

suite/tests/linux/clone.c
```diff
@@ -154,6 +186,14 @@ main()
      */
     test_thread(true /*share_sighand*/, true /*clone_vm*/, false /*use_clone3*/);
     test_thread(true /*share_sighand*/, true /*clone_vm*/, true /*use_clone3*/);
+
+#if defined(X86)
+    /* Test passing NULL for the stack pointer (xref i#6514). */
+    test_with_null_stack_pointer(false /*clone_vm*/, false /*use_clone3*/);
```
derekbruening, line 192: style: with `=` and before the value: so `/*clone_vm=*/false`.  Ditto for 2 instances on each of these 4 lines.

derekbruening (approved):

core/drlibc/drlibc_x86.asm
xdje42: Done

core/drlibc/drlibc_x86.asm
xdje42: Done

suite/tests/linux/clone.c
xdje42: Done

PR #6603, opened 2024-01-31
i#6585: Add drcachesim vector length trace marker

jackgallagher-arm:
Adds a new trace marker
```TRACE_MARKER_TYPE_DYNAMIC_VECTOR_LENGTH```
to drcachesim that indicates the current vector length for architectures which have a dynamic vector length that can't be statically determined from the instruction.

The marker is emitted as part of the thread header when running on AArch64 with SVE support, but in the future could also be used to track changes in the vector length after prctl(PR_SVE_SET_VL, ..) system calls (i#6625).

Some SVE load and store instructions such as
```LDR <Zt>, [<Xn|SP>{, #<imm>, MUL VL}]```
or
```ST1D { <Zt>.D }, <Pg>, [<Xn|SP>{, #<imm>, MUL VL}]```

scale the immediate offset based on the hardware vector length so knowing the correct vector length for the traced application is important to properly decode and analyse these instructions.

Fixes: #6585
Issues: #6625

### DIFF
--- a/clients/drcachesim/common/trace_entry.h
+++ b/clients/drcachesim/common/trace_entry.h
@@ -617,7 +617,7 @@ typedef enum {
      * dynamic vector length, for example the AArch64 SVE vector length. This does not
      * apply to fixed length vector architectures.
      */
-    TRACE_MARKER_TYPE_DYNAMIC_VECTOR_LENGTH,
+    TRACE_MARKER_TYPE_VECTOR_LENGTH,
 
     // ...
     // These values are reserved for future built-in marker types.
--- a/clients/drcachesim/common/trace_entry.h
+++ b/clients/drcachesim/common/trace_entry.h
@@ -613,9 +613,9 @@ typedef enum {
     TRACE_MARKER_TYPE_CONTEXT_SWITCH_END,
 
     /**
-     * This marker's value is the current vector length in bytes for architectures with a
-     * dynamic vector length, for example the AArch64 SVE vector length. This does not
-     * apply to fixed length vector architectures.
+     * This marker's value is the current thread's vector length in bytes, for
+     * architectures with a dynamic vector length. For example the AArch64 SVE vector
+     * length. This does not apply to fixed length vector architectures.
      */
     TRACE_MARKER_TYPE_VECTOR_LENGTH,
 
--- a/clients/drcachesim/common/trace_entry.h
+++ b/clients/drcachesim/common/trace_entry.h
@@ -612,6 +612,13 @@ typedef enum {
      */
     TRACE_MARKER_TYPE_CONTEXT_SWITCH_END,
 
+    /**
+     * The marker is used to indicate the current vector length in bytes for
+     * architectures with a dynamic vector length, for example the AArch64 SVE vector
+     * length. This does not apply to fixed length vector architectures.
+     */
+    TRACE_MARKER_TYPE_DYNAMIC_VECTOR_LENGTH,
+
     // ...
     // These values are reserved for future built-in marker types.
     // ...
--- a/clients/drcachesim/common/trace_entry.h
+++ b/clients/drcachesim/common/trace_entry.h
@@ -613,9 +613,9 @@ typedef enum {
     TRACE_MARKER_TYPE_CONTEXT_SWITCH_END,
 
     /**
-     * The marker is used to indicate the current vector length in bytes for
-     * architectures with a dynamic vector length, for example the AArch64 SVE vector
-     * length. This does not apply to fixed length vector architectures.
+     * This marker's value is the current vector length in bytes for architectures with a
+     * dynamic vector length, for example the AArch64 SVE vector length. This does not
+     * apply to fixed length vector architectures.
      */
     TRACE_MARKER_TYPE_DYNAMIC_VECTOR_LENGTH,
 
--- a/clients/drcachesim/common/trace_entry.h
+++ b/clients/drcachesim/common/trace_entry.h
@@ -614,8 +614,16 @@ typedef enum {
 
     /**
      * This marker's value is the current thread's vector length in bytes, for
-     * architectures with a dynamic vector length. For example the AArch64 SVE vector
-     * length. This does not apply to fixed length vector architectures.
+     * architectures with a dynamic vector length. It is currently only used on AArch64.
+     *
+     * On AArch64 the marker's value contains the SVE vector length. The marker is
+     * emitted with the thread header to establish the initial vector length for that
+     * thread. In the future it will also be emitted later in the trace if the app
+     * changes the vector length at runtime (TODO i#6625). In all cases the vector
+     * length value is specific to the current thread.
+     * The vector length affects how some SVE instructions are decoded so any tools which
+     * decode instructions should clear any cached data and set the vector length used by
+     * the decoder using dr_set_sve_vector_length().
      */
     TRACE_MARKER_TYPE_VECTOR_LENGTH,
 
--- a/clients/drcachesim/tests/offline-view.templatex
+++ b/clients/drcachesim/tests/offline-view.templatex
@@ -5,16 +5,14 @@ Output format:
            1           0: +[0-9]+ <marker: version [0-9]>
            2           0: +[0-9]+ <marker: filetype 0x[0-9a-f]*>
            3           0: +[0-9]+ <marker: cache line size [0-9]*>
+           4           0: +[0-9]+ <marker: chunk instruction count [0-9]*>
+           5           0: +[0-9]+ <marker: page size [0-9]*>
 #ifdef __ARM_FEATURE_SVE
-           4           0: +[0-9]+ <marker: vector length [0-9]* bytes>
-           5           0: +[0-9]+ <marker: chunk instruction count [0-9]*>
-           6           0: +[0-9]+ <marker: page size [0-9]*>
+           6           0: +[0-9]+ <marker: vector length [0-9]* bytes>
            7           0: +[0-9]+ <marker: timestamp [0-9]*>
            8           0: +[0-9]+ <marker: tid [0-9]* on core [0-9]*>
            9           1: +[0-9]+ ifetch      .*
 #else
-           4           0: +[0-9]+ <marker: chunk instruction count [0-9]*>
-           5           0: +[0-9]+ <marker: page size [0-9]*>
            6           0: +[0-9]+ <marker: timestamp [0-9]*>
            7           0: +[0-9]+ <marker: tid [0-9]* on core [0-9]*>
            8           1: +[0-9]+ ifetch      .*
--- a/clients/drcachesim/tests/offline-view.templatex
+++ b/clients/drcachesim/tests/offline-view.templatex
@@ -5,11 +5,20 @@ Output format:
            1           0: +[0-9]+ <marker: version [0-9]>
            2           0: +[0-9]+ <marker: filetype 0x[0-9a-f]*>
            3           0: +[0-9]+ <marker: cache line size [0-9]*>
+#ifdef __ARM_FEATURE_SVE
+           4           0: +[0-9]+ <marker: vector length [0-9]* bytes>
+           5           0: +[0-9]+ <marker: chunk instruction count [0-9]*>
+           6           0: +[0-9]+ <marker: page size [0-9]*>
+           7           0: +[0-9]+ <marker: timestamp [0-9]*>
+           8           0: +[0-9]+ <marker: tid [0-9]* on core [0-9]*>
+           9           1: +[0-9]+ ifetch      .*
+#else
            4           0: +[0-9]+ <marker: chunk instruction count [0-9]*>
            5           0: +[0-9]+ <marker: page size [0-9]*>
            6           0: +[0-9]+ <marker: timestamp [0-9]*>
            7           0: +[0-9]+ <marker: tid [0-9]* on core [0-9]*>
            8           1: +[0-9]+ ifetch      .*
+#endif
 .*
 View tool results:
     *[0-9]* : total instructions
--- a/clients/drcachesim/tools/invariant_checker.cpp
+++ b/clients/drcachesim/tools/invariant_checker.cpp
@@ -614,8 +614,12 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
         }
         if (!TESTANY(OFFLINE_FILE_TYPE_FILTERED | OFFLINE_FILE_TYPE_IFILTERED,
                      shard->file_type_)) {
-            report_if_false(shard, type_is_instr(shard->prev_instr_.memref.instr.type),
-                            "An unfiltered thread should have at least 1 instruction");
+            report_if_false(
+                shard,
+                type_is_instr(shard->prev_instr_.memref.instr.type) ||
+                    shard->prev_instr_.memref.instr.type == TRACE_TYPE_PREFETCH_INSTR ||
+                    shard->prev_instr_.memref.instr.type == TRACE_TYPE_INSTR_NO_FETCH,
+                "An unfiltered thread should have at least 1 instruction");
         }
     }
     if (shard->prev_entry_.marker.type == TRACE_TYPE_MARKER &&
--- a/clients/drcachesim/tools/invariant_checker.cpp
+++ b/clients/drcachesim/tools/invariant_checker.cpp
@@ -372,9 +372,9 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
         static const int MAX_VL_BYTES = 256; // SVE's maximum vector length is 2048-bit
         // Vector length must be a multiple of 16 bytes between 16 and 256.
         report_if_false(shard,
-                        (memref.marker.marker_value > 0) &&
-                            (memref.marker.marker_value <= MAX_VL_BYTES) &&
-                            (memref.marker.marker_value % 16 == 0),
+                        memref.marker.marker_value > 0 &&
+                            memref.marker.marker_value <= MAX_VL_BYTES &&
+                            memref.marker.marker_value % 16 == 0,
                         "Vector length marker has invalid size");
 
         const int new_vl_bits = memref.marker.marker_value * 8;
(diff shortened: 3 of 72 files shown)

### CODE REVIEW

File: clients/drcachesim/common/trace_entry.h [hunk 1]
```diff
@@ -612,6 +612,13 @@ typedef enum {
      */
     TRACE_MARKER_TYPE_CONTEXT_SWITCH_END,
 
+    /**
+     * The marker is used to indicate the current vector length in bytes for
```
abhinav92003, line 616: When we start adding this marker on `prctl(PR_SVE_SET_VL, ..)` calls, we might need to somehow add the marker in all thread's buffers the moment the change occurred. (I assume the prctl call changes the vector size for all threads.) Add a TODO?

clients/drcachesim/tests/offline-view.templatex
```diff
@@ -5,11 +5,20 @@ Output format:
            1           0: +[0-9]+ <marker: version [0-9]>
            2           0: +[0-9]+ <marker: filetype 0x[0-9a-f]*>
            3           0: +[0-9]+ <marker: cache line size [0-9]*>
+#ifdef __ARM_FEATURE_SVE
+           4           0: +[0-9]+ <marker: vector length [0-9]* bytes>
```
abhinav92003, line 9: Trying to understand if it is better to add this in the header or just as one of the trace entries. Because the markers we add on prctl calls will of course not be in the header. @derekbruening

clients/drcachesim/tracer/instru_online.cpp
```diff
@@ -178,6 +178,12 @@ online_instru_t::append_thread_header(byte *buf_ptr, thread_id_t tid,
     new_buf += append_marker(new_buf, TRACE_MARKER_TYPE_FILETYPE, file_type);
     new_buf += append_marker(new_buf, TRACE_MARKER_TYPE_CACHE_LINE_SIZE,
                              proc_get_cache_line_size());
+#if defined(AARCH64)
+    if (proc_has_feature(FEATURE_SVE)) {
+        new_buf += append_marker(new_buf, TRACE_MARKER_TYPE_DYNAMIC_VECTOR_LENGTH,
```
abhinav92003, line 183: Would it be simpler to just add the marker on other archs too? Helps keep the traces consistent IMHO. @derekbruening

abhinav92003 (approved):

clients/drcachesim/common/trace_entry.h
jackgallagher-arm: > (I assume the prctl call changes the vector size for all threads.)

`prctl(PR_SVE_SET_VL, ..)` only changes the vector length for the current thread so that shouldn't necessary.
https://man7.org/linux/man-pages/man2/prctl.2.html (search for PR_SVE_SET_VL)

clients/drcachesim/common/trace_entry.h
```diff
@@ -612,6 +612,13 @@ typedef enum {
      */
```
derekbruening, line 612: >  but in the future could also be used to track changes in the vector length after prctl(PR_SVE_SET_VL, ..) system calls.

Please add a reference to the issue covering that future work.

clients/drcachesim/common/trace_entry.h
derekbruening, line 616: nit: All other marker docs explicitly say "the marker value contains"; that would make it crystal clear that the length is in fact in the value (and not indicated in some other way...I agree I don't know how but good to match the others and be very clear).

clients/drcachesim/tests/offline-view.templatex
derekbruening: You mean, the raw2trace code below should not look for this in a header but in its general record processing?  Then when we add prctl support we don't have to change raw2trace?

suite/tests/CMakeLists.txt
```diff
@@ -4499,6 +4503,22 @@ if (BUILD_CLIENTS)
             "allasm-scattergather-basic-counts-${ARCH_NAME}")
     endif ()
 
+    if (UNIX AND (AARCH64 AND proc_supports_sve))
```
derekbruening, line 4506: nit: Redundant parens

clients/drcachesim/tracer/instru_online.cpp
derekbruening: Just a TRACE_MARKER_TYPE_VECTOR_LENGTH, and extend proc_get_vector_length_bytes() to other arches so on x86 it would tell you whether AVX-512 is there or not?  I could see that.  It would eliminate ifdefs in the test output above.  All the marker processing code is already not inside ifdefs which is good.  I don't feel strongly about it though: could live with it being aarch64 only.

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -932,6 +932,23 @@ raw2trace_t::read_header(raw2trace_thread_data_t *tdata,
         header->cache_line_size = proc_get_cache_line_size();
         unread_last_entry(tdata);
     }
+
+    in_entry = get_next_entry(tdata);
```
derekbruening, line 936: See comment above: to handle prctl this will have to be in the general mid-trace processing, so should we start out with it there instead of having to look for this marker in two different places?

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -1002,6 +1019,13 @@ raw2trace_t::process_header(raw2trace_thread_data_t *tdata)
     create_essential_header_entries(buf, version, tdata->file_type, tid, pid);
     buf += trace_metadata_writer_t::write_marker(buf, TRACE_MARKER_TYPE_CACHE_LINE_SIZE,
                                                  header.cache_line_size);
+    if (header.vector_length_bytes > 0) {
+#ifdef AARCH64
+        dr_set_sve_vector_length(header.vector_length_bytes * 8);
```
derekbruening, line 1024: Please update the opcode_mix and invariant_checker tools to set this as well so their decoding will be accurate.

derekbruening, line 1024: We'll have to udpate internal tools as well; along with other users who have their own tools.  Please add this to the release notes so it's visible.

Probably it should be added to the docs too: to sec_drcachesim_format where it talks about the encoding_is_new field, etc.?

clients/drcachesim/tests/offline-view.templatex
abhinav92003: Yes.

clients/drcachesim/common/trace_entry.h
jackgallagher-arm: Done.

clients/drcachesim/common/trace_entry.h
jackgallagher-arm: Done.

suite/tests/CMakeLists.txt
jackgallagher-arm: Done.

clients/drcachesim/tracer/raw2trace.cpp
jackgallagher-arm: Done.

clients/drcachesim/tracer/raw2trace.cpp
jackgallagher-arm: Done.

jackgallagher-arm: Done.

clients/drcachesim/tracer/instru_online.cpp
jackgallagher-arm: > could live with it being aarch64 only
It is only used on AArch64 at the moment but RISC-V also has a vector length that can be changed at runtime so it might be useful for that too in the future.

The first version of this patch actually did call the maker `TRACE_MARKER_TYPE_VECTOR_LENGTH` and tried to emit it on all architectures but I changed it to the way it is now when I started testing it on x86 and realised `proc_get_vector_length_bytes()` is aarch64-only. I didn't want to have to define what `proc_get_vector_length_bytes()` means for every architecture. It's easier to start with a narrow, well defined scope and broaden it later than it is to start with a broad scope and narrow/change it later if your broad definition turns out to be wrong.

clients/drcachesim/tests/offline-view.templatex
jackgallagher-arm: I changed it to handle the marker in the general record processing.

clients/drcachesim/common/trace_entry.h
```diff
@@ -613,9 +613,9 @@ typedef enum {
     TRACE_MARKER_TYPE_CONTEXT_SWITCH_END,
 
     /**
-     * The marker is used to indicate the current vector length in bytes for
-     * architectures with a dynamic vector length, for example the AArch64 SVE vector
-     * length. This does not apply to fixed length vector architectures.
+     * This marker's value is the current vector length in bytes for architectures with a
+     * dynamic vector length, for example the AArch64 SVE vector length. This does not
```
abhinav92003, line 617: Also mention that the vector length has thread-local scope.

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -883,6 +883,21 @@ raw2trace_t::process_marker_additionally(raw2trace_thread_data_t *tdata,
         log(2, "Maybe-blocking syscall %zu\n", marker_val);
         buf += trace_metadata_writer_t::write_marker(
             buf, TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL, 0);
+    } else if (marker_type == TRACE_MARKER_TYPE_DYNAMIC_VECTOR_LENGTH) {
+#ifdef AARCH64
+        log(4, "Setting SVE vector length to %zu bytes\n", marker_val);
```
abhinav92003, line 888: Also mention the thread id, because the indicated vector length is thread-local IIUC.

clients/drcachesim/tracer/instru_online.cpp
```diff
@@ -178,13 +178,13 @@ online_instru_t::append_thread_header(byte *buf_ptr, thread_id_t tid,
     new_buf += append_marker(new_buf, TRACE_MARKER_TYPE_FILETYPE, file_type);
     new_buf += append_marker(new_buf, TRACE_MARKER_TYPE_CACHE_LINE_SIZE,
                              proc_get_cache_line_size());
+    new_buf += append_marker(new_buf, TRACE_MARKER_TYPE_PAGE_SIZE, dr_page_size());
 #if defined(AARCH64)
```
abhinav92003, line 182: Since this is still inside a function named append_thread_header, maybe add a comment highlighting that there may be other instances too in the trace besides the one in the header, and that this marker should preferably be the last thing emitted by the header (assuming that's what we prefer).

clients/drcachesim/tracer/instru_offline.cpp
```diff
@@ -396,13 +396,13 @@ offline_instru_t::append_thread_header(byte *buf_ptr, thread_id_t tid,
     new_buf += append_pid(new_buf, dr_get_process_id());
     new_buf += append_marker(new_buf, TRACE_MARKER_TYPE_CACHE_LINE_SIZE,
                              proc_get_cache_line_size());
+    new_buf += append_marker(new_buf, TRACE_MARKER_TYPE_PAGE_SIZE, dr_page_size());
 #if defined(AARCH64)
```
abhinav92003, line 400: Same comment as in instru_online.cpp

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -366,6 +366,28 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                             memref.marker.marker_value == shard->stream->get_page_size(),
                         "Stream interface page size != trace marker");
     }
+    if (memref.marker.type == TRACE_TYPE_MARKER &&
+        memref.marker.marker_type == TRACE_MARKER_TYPE_DYNAMIC_VECTOR_LENGTH) {
+#ifdef AARCH64
+        static const int MAX_VL_BYTES = 256; // SVE's maximum vector length is 2048-bit
+        // Vector length must be a multiple of 16 bytes between 16 and 256.
+        report_if_false(shard,
+                        (memref.marker.marker_value > 0) &&
```
abhinav92003, line 375: Drop redundant parenthesis (multiple instances).

clients/drcachesim/tracer/raw2trace.cpp
abhinav92003: Should add this note to the marker docs too.

abhinav92003 (approved):

clients/drcachesim/tracer/instru_online.cpp
derekbruening: The thing that is hard to change later is the name of the marker, so if we think we do want to generalize proc_get_vector_length_bytes() it seems best to remove the _DYNAMIC_ from the name even if the implementation starts out aarch64-only (that iniitial limitation would be noted in the marker docs).

jackgallagher-arm: I have changed it to `TRACE_MARKER_TYPE_VECTOR_LENGTH`

clients/drcachesim/tracer/raw2trace.cpp
jackgallagher-arm: Done.

clients/drcachesim/common/trace_entry.h
jackgallagher-arm: Done.

clients/drcachesim/tracer/raw2trace.cpp
jackgallagher-arm: Done.

clients/drcachesim/tracer/instru_online.cpp
jackgallagher-arm: I have added a comment that there might be later instances of the marker.

clients/drcachesim/tools/invariant_checker.cpp
jackgallagher-arm: Done.

clients/drcachesim/tracer/instru_offline.cpp
jackgallagher-arm: Done.

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -3783,17 +3801,6 @@ raw2trace_t::raw2trace_t(
...
-
-#if defined(AARCH64)
-    // TODO i#6556, i#1684: The decoder uses a global sve_veclen variable to store the
-    // vector length value it uses when decoding. drdecodelib ends up being linked into
-    // drcachesim twice: once into the drcachesim executable, and one into libdynamorio.
-    // When we call dr_standalone_init() above it will initialize the version of
-    // sve_veclen in libdynamorio, but not the one in drcachesim.
-    // Unfortunately it is the version of sve_veclen in drcachesim that gets used when
-    // decoding in raw2trace so we need to explicitly initialize its sve_veclen here.
-    dr_set_sve_vector_length(proc_get_vector_length_bytes() * 8);
```
derekbruening, line 3795: If someone processes an older trace without the new marker, no one will set the length and it will be the default which seems to be 0?  If 0, what does the decoder do: does it error out?  Should there be a better default?

derekbruening, line 3795: nit: too late to fix I know: the description has the wrong name `TRACE_MARKER_TYPE_DYNAMIC_VECTOR_LENGTH`

derekbruening: Outside of drmemtrace, this seems to make using DR's standalone decoder painful for SVE: if the default is 0 would someone decoding SVE have the decoder fall over?  Anyone using the decoder now has to call the set vector length function?  Should that be documented?

jackgallagher-arm: > If someone processes an older trace without the new marker, no one will set the length and it will be the default which seems to be 0? If 0, what does the decoder do: does it error out? Should there be a better default?

Currently, if the vector length is set to 0 the decoder will not produce an error. It will just use 0 in its calculations which will result in some SVE load/store instructions being decoded with a 0 byte offset when they shouldn't be. drdisas has a `-vl` command line option to set the vector length. We could add an equivalent option to drcachesim to allow users to specify a vector length for traces which don't contain a vector length marker.

> Outside of drmemtrace, this seems to make using DR's standalone decoder painful for SVE: if the default is 0 would someone decoding SVE have the decoder fall over? Anyone using the decoder now has to call the set vector length function? Should that be documented?

In order to handle vector length changes properly in DynamoRIO (#6625) I think we are going to have to change the way vector length is stored and managed, so that might mean changes to `dr_set_sve_vector_length()`. We can make sure as part of that issue that how it works, and any requirements on standalone decoder users are properly documented.

derekbruening: Probably the default should be 128 or something valid so that reasonable results are obtained without explicit calls to set vector lengths: I suspect 99% of users are not going to make that call on the first pass and a reasonable default should be provided.

abhinav92003: > so that reasonable results are obtained without explicit calls to set vector lengths:

Explicit failures may be better than hidden trace/decoding correctness issues.

> a reasonable default should be provided.

Instead of a fixed number, could we initialize it using proc_get_vector_length_bytes in the decoder? E.g., https://github.com/DynamoRIO/dynamorio/blob/62d441e8bd0c8aad489e8a02dcb421db30d72da4/core/ir/opnd_shared.c#L2762

If it cannot be obtained from the environment, it may be better to fail, so that the user can set it explicitly. (except in cases where we know a certain number to be a very good default)

derekbruening: Having a decoder that fails when you try to use if you don't make a call to set the vector length does not sound user-friendly to me: most will not know they're supposed to call that; many will not care and just want to see the opcode in offline decoding.  Some kind of general initialization routine seems better; default also seem good to me.

Let's file a new issue on this.  It relates to several other issues on trying to make drdecode work without explicit initialization, which maybe we have to abandon for other reasons.  I will file one and put in the xrefs.

derekbruening: Filed #6646

PR #6609, opened 2024-01-31
i#6596 linux.clone* tests on Ubuntu: Handle struct clone_args not defined

### LINKED ISSUE #6596: linux.clone* tests fail when run in a container under Ubuntu 22.04

xdje42:
Fix build errors resulting from #6597 by using DR's copy of struct clone_args

This has happened on Android.

Fixes #6596

### DIFF
--- a/suite/tests/linux/clone.c
+++ b/suite/tests/linux/clone.c
@@ -109,7 +109,11 @@ test_thread(bool share_sighand, bool clone_vm, bool use_clone3)
      */
     pid_t (*create_thread_func)(void (*fcn)(void), void **stack, bool share_sighand,
                                 bool clone_vm) =
-        (use_clone3 && clone3_available) ? create_thread_clone3 : create_thread;
+        create_thread;
+#ifdef SYS_clone3
+    if (use_clone3 && clone3_available)
+        create_thread_func = create_thread_clone3;
+#endif
 
     child = create_thread_func(run_with_exit, &stack, share_sighand, clone_vm);
 
@@ -124,19 +128,22 @@ main()
     int ret_failure_clone3 = make_clone3_syscall(NULL, 0, NULL);
     assert(ret_failure_clone3 == -1);
 
-#ifndef ANDROID /* clone3 not supported on Android */
-    /* i#6596 In some scenarios SYS_clone3 is defined but clone3 returns ENOSYS
-     * e.g. when running in a container under Ubuntu 22.04
-     * see https://github.com/moby/moby/pull/42681
-     */
-    if (errno != ENOSYS)
-        clone3_available = true;
-#endif
-
     /* On some environments, we see that the kernel supports clone3 even though
      * SYS_clone3 is not defined by glibc.
      */
     assert(errno == ENOSYS || errno == EINVAL);
+
+    /* i#6596 In some scenarios SYS_clone3 is defined but clone3 returns ENOSYS
+     * e.g. when running in a container under Ubuntu 22.04
+     * see https://github.com/moby/moby/pull/42681
+     * The presence of SYS_clone3 is also a gating factor on whether glibc
+     * defines struct clone_args. So only try clone3 if the kernel supports it
+     * and the definition of struct clone_args is available.
+     */
+#ifdef SYS_clone3
+    if (errno != ENOSYS)
+        clone3_available = true;
+#endif
 
     /* First test a thread that does not share signal handlers
      * (xref i#2089).
@@ -302,15 +309,8 @@ make_clone3_syscall(void *clone_args, ulong clone_args_size, void (*fcn)(void))
     return result;
 }
 
-#ifdef ANDROID
-static pid_t
-create_thread_clone3(void (*fcn)(void), void **stack, bool share_sighand, bool clone_vm)
-{
-    /* Should never get here */
-    assert(0);
-    return -1;
-}
-#else
+#ifdef SYS_clone3
+
 static pid_t
 create_thread_clone3(void (*fcn)(void), void **stack, bool share_sighand, bool clone_vm)
 {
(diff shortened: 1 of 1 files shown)


suite/tests/linux/clone.c
```diff
@@ -124,20 +129,23 @@ main()
     int ret_failure_clone3 = make_clone3_syscall(NULL, 0, NULL);
     assert(ret_failure_clone3 == -1);
 
-#ifndef ANDROID /* clone3 not supported on Android */
+    /* On some environments, we see that the kernel supports clone3 even though
+     * SYS_clone3 is not defined by glibc.
```
derekbruening, line 133: This doesn't all seem consistent: it makes a clone3 syscall outside of any ifdefs, and has a comment saying clone3 may be available even if SYS_clone3 is not defined, yet it puts the enabling and use of it back under an ifdef for SYS_clone3?  It doesn't seem worth trying the syscall, or implying here we should go ahead and use clone3, if below we go back to the ifdef for what matters, setting clone3_available.  This code sequence lines 128 through 147 would be simpler and more understandable if it were all under ifdef SYS_clone3 if we're never going to use clone3 if that's not defined.

philramsey-arm:
Have confirmed that the linux.clone* tests still pass in a Ubuntu 22:04 container.

suite/tests/linux/clone.c
xdje42: One needs to go back to how things were, pre- #6597.
That code was calling clone3 outside of ifdefs, presumably to make sure the right errno was returned?
Removing that is outside the scope of #6596 (which #6597 was an attempt to fix).

derekbruening: I don't follow the argument: this PR is taking a sequence of code that does work to compute `clone3_available` and adds an ifdef only around the final step instead of around all the work, when all the work is *only* for computing that final step `clone3_available`: so it is this PR which is introducing the awkward code.  Not seeing why the history matters.

abhinav92003: > Question: Why does it make a clone3 syscall ifndef SYS_clone3 ?
> That code was calling clone3 outside of ifdefs, presumably to make sure the right errno was returned?

Maybe your question is why do we need to verify clone3 behaviour if SYS_clone3 is not defined?

Note that clone3 is one of the syscalls that DR intercepts. So if SYS_clone3 is not defined but the app still tries to make the syscall, we want to make sure that DR returns the correct error code.

xdje42: Rewrote patch to use DR's copy of struct clone_args from core/unix/include/clone3.h.

suite/tests/linux/clone.c
```diff
@@ -48,6 +48,7 @@
 #include <errno.h>
```
abhinav92003, line 48: Q: Is the test failing by not building successfully, or by running incorrectly? I assume it's the latter because the issue description says "The linux.clone* tests should not fail if clone3 returns ENOSYS." But I would have expected a missing struct definition to cause a build failure.

I'm probably missing some context. Could you clarify in the PR description?

xdje42: The test is failing due to build errors.

suite/tests/linux/clone.c
```diff
@@ -120,23 +121,23 @@ test_thread(bool share_sighand, bool clone_vm, bool use_clone3)
...
     int ret_failure_clone3 = make_clone3_syscall(NULL, 0, NULL);
     assert(ret_failure_clone3 == -1);
 
-#ifndef ANDROID /* clone3 not supported on Android */
-    /* i#6596 In some scenarios SYS_clone3 is defined but clone3 returns ENOSYS
-     * e.g. when running in a container under Ubuntu 22.04
+    /* In some environments, we see that the kernel supports clone3 even though
+     * SYS_clone3 is not defined by glibc. So we don't predicate our efforts on
+     * whether SYS_clone3 is defined.
+    /* Plus in some scenarios SYS_clone3 is defined but clone3 returns ENOSYS.
```
abhinav92003, line 134: Since this "Plus..." is a continuation of the above statement, better to not keep it in a separate comment, otherwise it loses context.

suite/tests/linux/clone.c
```diff
@@ -48,6 +48,7 @@
 #include <errno.h>
 
 #include "tools.h" /* for nolibc_* wrappers. */
```
abhinav92003, line 50: Re the PR description: The issue is likely not specific to Ubuntu. It probably happens where the glibc version is older (so the struct is not defined)?

Either way, instead of trying to fit that into the concise PR title, maybe include the affected platforms in the PR description.

abhinav92003 (approved):

suite/tests/linux/clone.c
xdje42: Done, modulo internal builds.

suite/tests/linux/clone.c
xdje42: Done

PR #6610, opened 2024-01-31
Add instructions on how to watch for test failures to triage docs

xdje42:
And remove reference to Jenkins since, while used, is done as part of a Github Action, and thus there's no separate place to look for failures.

### DIFF
--- a/api/docs/triager.dox
+++ b/api/docs/triager.dox
@@ -36,9 +36,11 @@
 
 We have a rotating Triager who is responsible for monitoring our continuous testing infrastructure and for triaging incoming requests from users.  Specific duties include:
 
-- Ensure that our Continuous Integration testing through Github Actions and Jenkins are operating smoothly.
+- Ensure that our Continuous Integration testing through Github Actions are operating smoothly.
+  - Do this by going to the Github Actions page: https://github.com/DynamoRIO/dynamorio/actions.
   - If flaky tests are failing too often, assign someone to fix them ASAP, or mark them to be ignored in `runsuite_wrapper.pl`.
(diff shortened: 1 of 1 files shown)


api/docs/triager.dox
```diff
@@ -36,9 +36,11 @@
```
derekbruening, line 36: > And remove reference to Jenkins since, while used, is done as part of a Github Action, and thus there's no separate place to look for failures.

nit: Generally the body should stand on its own, including repeating what is in the title if necessary: i.e., best to not start it with "And".  The title is not always right there depending on what history tool is used.

api/docs/triager.dox
```diff
@@ -36,9 +36,11 @@
 
 We have a rotating Triager who is responsible for monitoring our continuous testing infrastructure and for triaging incoming requests from users.  Specific duties include:
 
-- Ensure that our Continuous Integration testing through Github Actions and Jenkins are operating smoothly.
+- Ensure that our Continuous Integration testing through Github Actions are operating smoothly.
+  - Do this by going to the Github Actions page: https://github.com/DynamoRIO/dynamorio/actions.
   - If flaky tests are failing too often, assign someone to fix them ASAP, or mark them to be ignored in `runsuite_wrapper.pl`.
 - Watch merges to master for failures on the longer test suite.
+  - Do this by watching for merges on the dynamorio-devs@ list.
```
derekbruening, line 43: No need to look at the devs list.  These are on the same Actions page linked above, to find details -- but the summary page is the green check or red X by each commit at https://github.com/DynamoRIO/dynamorio/commits/master/ .  That's what the goal is: to have that page contain nothing but green checks.

PR #6614, opened 2024-02-01
iX: RHEL-7 build fix due to EM_RISCV.

prasun3:
The elf.h header file on RHEL-7 does not define EM_RISCV which is being used in the source files. This results in a build failure. This change fixes the issue.

### DIFF
--- a/core/unix/module_elf.h
+++ b/core/unix/module_elf.h
@@ -42,6 +42,11 @@
 #    define DT_RELR 36
 #endif
 
+/* XXX: Workaround for EM_RISCV not being defined in elf.h on RHEL-7. */
+#ifndef EM_RISCV
+#    define EM_RISCV 243
+#endif
+
 /* XXX i#1345: support mixed-mode 32-bit and 64-bit in one process.
  * There is no official support for that on Linux or Mac and for now we do
  * not support it either, especially not mixing libraries.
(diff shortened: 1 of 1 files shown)


core/unix/module_elf.h
```diff
@@ -42,6 +42,11 @@
 #    define DT_RELR 36
 #endif
 
+/* XXX: Workaround for EM_RISCV not being defined in elf.h on RHEL-7. */
```
derekbruening, line 45: XXX means there is still work to do (though less work or less urgently than TODO) so it should just be removed here.

core/unix/module_elf.h
```diff
@@ -42,6 +42,11 @@
 #    define DT_RELR 36
```
derekbruening, line 42: > iX: RHEL-7 build fix due to EM_RISCV.

style: We don't put `iX: ` in the title: just "RHEL-7 build fix due to EM_RISCV".

derekbruening (approved):

core/unix/module_elf.h
prasun3: Fixed.

core/unix/module_elf.h
prasun3: Fixed.

PR #6616, opened 2024-02-01
i#6615: linux.bad-signal-stack test fails on Ubuntu 22.04

### LINKED ISSUE #6615: linux.bad-signal-stack test fails when run in a container under Ubuntu 22.04
`bad-signal-stack.c` includes `tools.h` which defines `_GNU_SOURCE`.
Having `_GNU_SOURCE` defined when `signal.h` is included causes
the test to fail when run in a Ubuntu 22.04 container.

This can be fixed by including `tools.h` after the other header
files, as other source files like signal_pre_syscall.c do.

**To Reproduce**
# On Ubuntu 22.04:
docker run --rm -it --name dynamorio ubuntu/base:22.04 bash

git clone https://github.com/DynamoRIO/dynamorio.git && cd dynamorio
git submodule init
git submodule update

# build tests
cd ..
mkdir dynamorio_build
cd dynamorio_build
cmake -DDEBUG=ON -DBUILD_TESTS=ON ../dynamorio
make -j

# run tests
ctest --output-on-failure --timeout 120

philramsey-arm:
Update bad-signal-stack.c to include tools.h after the other header files so that _GNU_SOURCE is not defined when signal.h is included.

Fixes #6615

### DIFF
--- a/suite/tests/linux/bad-signal-stack.c
+++ b/suite/tests/linux/bad-signal-stack.c
@@ -30,13 +30,15 @@
  * DAMAGE.
  */
 
-#include "tools.h"
 #include <unistd.h>
 #include <signal.h>
 #include <ucontext.h>
 #include <errno.h>
 #include <stdlib.h>
 #include <assert.h>
+/* i#6615 include tools.h after signal.h to avoid issues on
+ * ubuntu 22:04 caused by _GNU_SOURCE being defined */
+#include "tools.h"
 
 #define ALT_STACK_SIZE (SIGSTKSZ * 4)
 
--- a/suite/tests/linux/bad-signal-stack.c
+++ b/suite/tests/linux/bad-signal-stack.c
@@ -32,7 +32,6 @@
 
 #include <unistd.h>
 #include <signal.h>
-#include <ucontext.h>
 #include <errno.h>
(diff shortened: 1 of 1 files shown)


suite/tests/linux/bad-signal-stack.c
```diff
@@ -30,13 +30,15 @@
...
-#include "tools.h"
 #include <unistd.h>
 #include <signal.h>
 #include <ucontext.h>
 #include <errno.h>
 #include <stdlib.h>
 #include <assert.h>
+/* i#6615 include tools.h after signal.h to avoid issues on
+ * ubuntu 22:04 caused by _GNU_SOURCE being defined */
+#include "tools.h"
```
derekbruening, line 41: Looks like this breaks the Mac build.

derekbruening, line 41: If we want to support building inside a container, could that be added to the automated tests?  Otherwise it is likely to be broken unknowingly in the future.

philramsey-arm: I imagine going forward we will use containers more and more, especially for CI. We already do for our internal CI.

I'll raise this internally.

Issue #6620, opened 2024-02-02
record_filter_launcher fails up front with usage message on Windows regardless of args


derekbruening:
On Windows, the record_filter_launcher fails up front with its Usage message regardless of what's passed in.  The two required args are always empty.  I tracked the problem down to a missing _UNICODE define.

PR #6621, opened 2024-02-02
i#6620: Fix record_filter_launcher on Windows

derekbruening:
Fixes an up-front failure with the record_filter_launcher tool on Windows by defining _UNICODE.

Tested both manually and with the new automated test in PR #6598.

Fixes: #6620

### DIFF
--- a/clients/drcachesim/tools/record_filter_launcher.cpp
+++ b/clients/drcachesim/tools/record_filter_launcher.cpp
@@ -34,8 +34,8 @@
 
 #ifdef WINDOWS
 #    define NOMINMAX // Avoid windows.h messing up std::max.
-#    define UNICODE
-#    define _UNICODE
+#    define UNICODE  // For Windows headers.
+#    define _UNICODE // For C headers.
 #endif
 
 #include "analyzer.h"
--- a/clients/drcachesim/tools/record_filter_launcher.cpp
+++ b/clients/drcachesim/tools/record_filter_launcher.cpp
@@ -32,6 +32,12 @@
 
 /* Standalone record filter tool launcher for file traces. */
 
+#ifdef WINDOWS
+#    define NOMINMAX // Avoid windows.h messing up std::max.
+#    define UNICODE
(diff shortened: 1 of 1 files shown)


clients/drcachesim/tools/record_filter_launcher.cpp
```diff
@@ -32,6 +32,12 @@
 
 /* Standalone record filter tool launcher for file traces. */
 
+#ifdef WINDOWS
+#    define NOMINMAX // Avoid windows.h messing up std::max.
+#    define UNICODE
```
abhinav92003, line 37: So we need both UNICODE defs? Maybe add a comment describing why they're required (e.g. I'm not familiar with some Windows-specific things like this)

clients/drcachesim/tools/record_filter_launcher.cpp
```diff
@@ -125,6 +133,10 @@ _tmain(int argc, const TCHAR *targv[])
 {
     disable_popups();
 
+#if defined(WINDOWS) && !defined(_UNICODE)
```
abhinav92003, line 136: Do we need this? Seems redundant to have a check like this when the fix is just another #define.

abhinav92003 (approved):

clients/drcachesim/tools/record_filter_launcher.cpp
derekbruening: Yes, IIRC one is Windows headers and one is C headers.  Right, confirmed: https://stackoverflow.com/questions/7953025/why-both-unicode-and-unicode

Added a comment.

clients/drcachesim/tools/record_filter_launcher.cpp
derekbruening: I want to avoid someone else having to spend time tracking down a weird error like I had to do, in the case that maybe these defines get shuffled down below some earlier #include of the headers that need the defines.  So IMHO it is worth having this check down here.

PR #6623, opened 2024-02-02
i#6622: Fix invariant on missing instructions

### LINKED ISSUE #6622: Invariant failure: "An unfiltered thread should have at least 1 instruction"
This showed up in my new test in PR #6598 but it seems a general issue that could show up anywhere with a detach or other non-syscall exit:

https://github.com/DynamoRIO/dynamorio/actions/runs/7749549688/job/21134306402?pr=6598
```
273:   *** postcmd failed (3): Trace invariant failure in T3104 at ref # 99016 (0
273:   instrs since timestamp 13351308211805717): An unfiltered thread should have
273:   at least 1 instruction
```

It looks like a general bug: the check doesn't consider unfetched or prefetches, which are copied into prev_instr_.
The exit-early case in this test probably made it more likely: need a no-fetch instr right before thread exit.

derekbruening:
Fixes an error in the "An unfiltered thread should have at least 1 instruction" drmemtrace invariant where it does not consider a no-fetch or prefetch instruction.

Adds a unit test which fails without the fix.

Fixes #6622

### DIFF
--- a/clients/drcachesim/tests/invariant_checker_test.cpp
+++ b/clients/drcachesim/tests/invariant_checker_test.cpp
@@ -3169,6 +3169,50 @@ check_kernel_syscall_trace(void)
 #endif
 }
 
+bool
+check_has_instructions(void)
+{
+    std::cerr << "Testing at-least-1-instruction\n";
+    // Correct: 1 regular instruction.
+    {
+        std::vector<memref_t> memrefs = {
+            gen_marker(TID_A, TRACE_MARKER_TYPE_CACHE_LINE_SIZE, 64),
+            gen_marker(TID_A, TRACE_MARKER_TYPE_PAGE_SIZE, 4096),
+            gen_instr(TID_A),
+            gen_exit(TID_A),
+        };
+        if (!run_checker(memrefs, false))
+            return false;
+    }
+    // Correct: 1 unfetched instruction.
+    {
+        std::vector<memref_t> memrefs = {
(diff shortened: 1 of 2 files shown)


clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -3169,6 +3169,50 @@ check_kernel_syscall_trace(void)
...
+        std::vector<memref_t> memrefs = {
+            gen_marker(TID_A, TRACE_MARKER_TYPE_CACHE_LINE_SIZE, 64),
+            gen_marker(TID_A, TRACE_MARKER_TYPE_PAGE_SIZE, 4096),
+            gen_instr(TID_A),
+            gen_exit(TID_A),
+        };
+        if (!run_checker(memrefs, false))
+            return false;
+    }
+    // Correct: 1 unfetched instruction.
```
abhinav92003, line 3187: (Optional) Some coverage analyses would like to see a test for TRACE_TYPE_PREFETCH_INSTR too.

abhinav92003 (approved):

clients/drcachesim/tests/invariant_checker_test.cpp
derekbruening: Hmm somehow this didn't show up in the page and it said no conversations and it only appeared after merging...strange, but not just a client cache issue b/c the server shouldn't let it merge w/ unresolved conversations.  I think for this we will live with just testing the other type.

PR #6629, opened 2024-02-06
Reformat suite/tests/linux/clone.c

xdje42:
Reformat with updated clang-format. There are no functional changes.

### DIFF
--- a/suite/tests/linux/clone.c
+++ b/suite/tests/linux/clone.c
@@ -47,7 +47,7 @@
 #include <string.h>
 #include <errno.h>
 
-#include "tools.h" /* for nolibc_* wrappers. */
+#include "tools.h"                          /* for nolibc_* wrappers. */
 #include "../../core/unix/include/clone3.h" /* for clone3_syscall_args_t */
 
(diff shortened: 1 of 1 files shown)


suite/tests/linux/clone.c
```diff
@@ -47,7 +47,7 @@
 #include <string.h>
 #include <errno.h>
 
-#include "tools.h" /* for nolibc_* wrappers. */
+#include "tools.h"                          /* for nolibc_* wrappers. */
```
derekbruening, line 50: I found the PR description misleading: there are no specific style rules being fixed here.  Instead it looks like clang-format itself has changed its internal rules.  The current file was formatted with whatever clang-format version was used on the whole codebase a few years ago (maybe it was version 6?).  We didn't change any style rules since then nor any clang-format config; just bumped the clang-format version a few times, now at 14.  Please update the PR description to make it clear there is no style rule being violated anywhere; these are internal-to-clang-format tweaks from upgrading our version of clang-format.

derekbruening (approved):

PR #6631, opened 2024-02-06
Add more detailed docs on using tmate

derekbruening:
Expands the docs on using tmate with detailed steps including removing other jobs and using a label to zero in on just one test.


derekbruening:
win64 hit some transient network error during setup; re-running.

edeiana (changes requested):
We should mention that this line of action seems to work only as a Project Member and not from PR coming from forks of dynamorio.

edeiana (approved):

PR #6632, opened 2024-02-06
i#2626: AArch64 codec: Add BTI hinted instruction

joshua-warburton:
This patch adds the tests and code to recognise the BTI instruction when it is encoded as a HINT

As this is the first v8.5 instruction, it also adds the appropriate infrastructure for generating that
codec.

issues: #2626

### DIFF
--- a/core/arch/aarch64/proc.c
+++ b/core/arch/aarch64/proc.c
@@ -223,6 +223,7 @@ proc_init_arch(void)
         LOG(GLOBAL, LOG_TOP, 1, "Processor features:\n ID_AA64PFR1_EL1 = 0x%016lx\n",
             cpu_info.features.flags_aa64pfr1);
         LOG_FEATURE(FEATURE_MTE);
+        LOG_FEATURE(FEATURE_BTI);
     });
 #    endif
 #endif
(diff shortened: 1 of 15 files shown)


AssadHashmi (approved):

xdje42:
Was this missed? I don't know enough to say, but there's no commenting in the file to suggest otherwise.

```
diff --git a/make/CMake_aarch64_gen_codec.cmake b/make/CMake_aarch64_gen_codec.cmake
index 06785ff9e..01c78bba4 100644
--- a/make/CMake_aarch64_gen_codec.cmake
+++ b/make/CMake_aarch64_gen_codec.cmake
@@ -44,6 +44,8 @@ set(AARCH64_CODEC_GEN_SRCS
   ${PROJECT_BINARY_DIR}/decode_gen_v81.h
   ${PROJECT_BINARY_DIR}/decode_gen_v82.h
   ${PROJECT_BINARY_DIR}/decode_gen_v83.h
+  ${PROJECT_BINARY_DIR}/decode_gen_v84.h
+  ${PROJECT_BINARY_DIR}/decode_gen_v85.h
   ${PROJECT_BINARY_DIR}/decode_gen_v86.h
   ${PROJECT_BINARY_DIR}/decode_gen_sve.h
   ${PROJECT_BINARY_DIR}/decode_gen_sve2.h
@@ -51,6 +53,8 @@ set(AARCH64_CODEC_GEN_SRCS
   ${PROJECT_BINARY_DIR}/encode_gen_v81.h
   ${PROJECT_BINARY_DIR}/encode_gen_v82.h
   ${PROJECT_BINARY_DIR}/encode_gen_v83.h
+  ${PROJECT_BINARY_DIR}/encode_gen_v84.h
+  ${PROJECT_BINARY_DIR}/encode_gen_v85.h
   ${PROJECT_BINARY_DIR}/encode_gen_v86.h
   ${PROJECT_BINARY_DIR}/encode_gen_sve.h
   ${PROJECT_BINARY_DIR}/encode_gen_sve2.h
```

I'm not sure what this does actually, since apparently nothing broke.

joshua-warburton:
I believe it was a bug, and would have caused that file not to be generated except that codec.py has its own list of files to generate so if any of those files change it regenerates them all. 

If this wasn't the case, and only encode/decode_gen_v84 had changed then the dependent files wouldn't have been updated in a non-fresh build. It's a pretty obscure circumstance so even if we didn't always write that file I doubt we'd frequently hit it.

Issue #6627, opened 2024-02-05
COMPILATION ERROR: Inlining clone.c:make_clone3_syscall -> multiply defined symbols


xdje42:
**Describe the bug**

This is a general bug, there are several places where labels are used in embedded asms.
This bug concerns the immediate issue of compilation failures when clone.c:make_clone3_syscall is inlined,
causing the label "parent" to be multiply defined.

**To Reproduce**

```
$ clang --version
Debian clang version 16.0.6 (16)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
$ clang -I ../dynamorio/core/drlibc -I ../dynamorio/core/lib -I ../dynamorio/suite/tests -I . -I include \
  -O1 -c ../dynamorio/suite/tests/linux/clone.c -o clone.o
```

-->

```
In file included from ../dynamorio/suite/tests/linux/clone.c:50:
In file included from ../dynamorio/suite/tests/tools.h:46:
../dynamorio/core/drlibc/drlibc.h:147:43: warning: declaration of
      'struct stat64' will not be visible outside of this function
      [-Wvisibility]
dr_stat_syscall(const char *fname, struct stat64 *st);
                                          ^
../dynamorio/suite/tests/linux/clone.c:253:32: error: symbol 'parent' is
      already defined
                 "call *%%rdx\n\t"
                               ^
<inline asm>:9:2: note: instantiated into assembly here
        parent:
        ^
../dynamorio/suite/tests/linux/clone.c:253:32: error: symbol 'parent' is
      already defined
                 "call *%%rdx\n\t"
                               ^
<inline asm>:9:2: note: instantiated into assembly here
        parent:
        ^
../dynamorio/suite/tests/linux/clone.c:253:32: error: symbol 'parent' is
      already defined
                 "call *%%rdx\n\t"
                               ^
<inline asm>:9:2: note: instantiated into assembly here
        parent:
        ^
../dynamorio/suite/tests/linux/clone.c:253:32: error: symbol 'parent' is
      already defined
                 "call *%%rdx\n\t"
                               ^
<inline asm>:9:2: note: instantiated into assembly here
        parent:
        ^
1 warning and 4 errors generated.
```

**Versions**
 - What version of DynamoRIO are you using? HEAD @ 3da572fb4fdb7021d58c88c421ba5993b9c9c0b9
- What operating system version are you running on? 64 bit linux

xdje42:
[...]

PR #6626, opened 2024-02-05
i#6611: Fix pthread lock hang

derekbruening:
Fixes a hang on glibc 2.37 by initializing the tid pthread TLS field. Its offset is located by decoding an exported function known to reference it in a new routine privload_set_pthread_tls_fields().

Only x86 is supported with this fix as no aarch64 machine with the required glibc is available for developing and testing the decode fix.

Tested on a glibc 2.37 machine where without this fix the client.drcallstack test hangs in release build.

Also tested on a Dr. Memory 2.6.19737 build pointing at a release build DR with this fix and confirmed it fixes the hang there.

Fixes #6611

PR #6633, opened 2024-02-06
Revert "i#6627: Fix multiply defined label in linux/clone.c (#6628)"

derekbruening:
This reverts commit 984e33548cde2b9869d16a60b15383c494470386 as it breaks our internal build.

Issue: #6627


xdje42:
Technically speaking, we don't know if it breaks the build, thus the description is inaccurate.
The patch is just being conservative.

derekbruening:
> Technically speaking, we don't know if it breaks the build, thus the description is inaccurate. The patch is just being conservative.

We do know as it requires an extra step to remove the patch file and the patch fails to apply (by inspection, plus the patch failed when I tested my PR #6626 so we know it fails).

derekbruening:
Reasonable to have TBR here IMHO to get the repo back to a pull-able state quickly.

PR #6634, opened 2024-02-06
Revert "Reformat suite/tests/linux/clone.c (#6629)"

derekbruening:
This reverts commit ef572627cab469b830d08d7be22f1de9d581e8da as it breaks out internal build's workaround for i#6627.

PR #6637, opened 2024-02-07
Reformat suite/tests/linux/clone.c

xdje42:
Reformat with updated clang-format. There are no functional changes.

This reapplies commit ef572627cab469b830d08d7be22f1de9d581e8da (#6629)
that was reverted in commit c244df14f1695b4e171b594cf58a3f47dff91f8a (#6634).

### DIFF
--- a/suite/tests/linux/clone.c
+++ b/suite/tests/linux/clone.c
@@ -47,7 +47,7 @@
 #include <string.h>
 #include <errno.h>
 
-#include "tools.h" /* for nolibc_* wrappers. */
+#include "tools.h"                          /* for nolibc_* wrappers. */
 #include "../../core/unix/include/clone3.h" /* for clone3_syscall_args_t */
 
 #ifdef ANDROID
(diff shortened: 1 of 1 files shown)


xdje42:
Dunno what the protocol is for specifying in the description the patch is the re-application of a reverted patch.
Please advise.

suite/tests/linux/clone.c
```diff
@@ -47,7 +47,7 @@
 #include <string.h>
```
derekbruening, line 47: > Dunno what the protocol is for specifying in the description the patch is the re-application of a reverted patch.
> Please advise.

There's nothing written down but it seems best to put a reference to the reverted and to the reverting PR numbers in the commit message so they'll all be linked.

derekbruening, line 47: (Discussing offline the plan for handling complications with importing clone.c changes internally.)

xdje42: Done.

PR #6639, opened 2024-02-07
i#6636: Add randomized schedule feature

### LINKED ISSUE #6636: Add random next input to schedule for sensitivity studies
For studying scheduler sensitivity studies, we can turn off timestamp ordering, but there's still a FIFO in the scheduler. The idea here is to add a flag to do a random order instead of FIFO to help with these studies.

derekbruening:
Adds a new scheduler option randomize_next_input and corresponding launcher option -sched_randomize.  When enabled, priorities and timestamps and FIFO ordering are ignored and instead a random next element from the ready queue is selected.  This will be useful for schedule sensitivity studies.

Adds a unit test.

Tested manually end-to-end as well:
```
  $ for ((i=0; i<10; ++i)); do bin64/drrun -t drcachesim -simulator_type schedule_stats -core_sharded -indir ../src/clients/drcachesim/tests/drmemtrace.threadsig.x64.tracedir/ -cores 2 -sched_randomize 2>&1 | grep schedule; done
  Core #0 schedule: HEFBF_
  Core #1 schedule: GCDFA
  Core #0 schedule: GDBH
  Core #1 schedule: CFAFEF__
  Core #0 schedule: GDBF__
  Core #1 schedule: EFCFAH
  Core #0 schedule: BHFDF__
  Core #1 schedule: EGFAC
  Core #0 schedule: HAFEF__
  Core #1 schedule: GDCFB
  Core #0 schedule: ABFGF__
  Core #1 schedule: CEDFH
  Core #0 schedule: HDBF_F_F__
  Core #1 schedule: ECGA
  Core #0 schedule: HDEFA
  Core #1 schedule: CBFGF__
  Core #0 schedule: FHGFCF_
  Core #1 schedule: DEAB
  Core #0 schedule: EFABF_
  Core #1 schedule: GCDFH
```

Fixes #6636

### DIFF
--- a/clients/drcachesim/common/options.cpp
+++ b/clients/drcachesim/common/options.cpp
@@ -899,7 +899,7 @@ droption_t<std::string> op_sched_switch_file(
 
 droption_t<bool> op_sched_randomize(
     DROPTION_SCOPE_FRONTEND, "sched_randomize", false,
-    "Pick next inputs randomnly on context switches",
+    "Pick next inputs randomly on context switches",
     "Applies to -core_sharded and -core_serial.  Disables the normal methods of "
     "choosing the next input based on priority, timestamps (if -sched_order_time is "
     "set), and FIFO order and instead selects the next input randomly. "
--- a/clients/drcachesim/common/options.cpp
+++ b/clients/drcachesim/common/options.cpp
@@ -1,5 +1,5 @@
 /* **********************************************************
- * Copyright (c) 2015-2023 Google, Inc.  All rights reserved.
+ * Copyright (c) 2015-2024 Google, Inc.  All rights reserved.
  * **********************************************************/
 
 /*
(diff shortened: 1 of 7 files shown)


derekbruening:
Hmm, every schedule is the same on Windows.  The randomizer is seeded from the clock.  Is the clock virtualization that bad on these runner VM's?

derekbruening:
> Hmm, every schedule is the same on Windows. The randomizer is seeded from the clock. Is the clock virtualization that bad on these runner VM's?

It looks like GetSystemTime just has poor granularity: it returns the same time throughout the whole test.  QueryPerformanceCounter is much better and makes the test pass.  But, there are several other uses of GetSystemTime so maybe a separate PR should update them all and then this one can merge it in.  So I think this can be reviewed as-is; won't be merged until that Windows change is made separately and this becomes green.

clients/drcachesim/common/options.cpp
```diff
@@ -897,6 +897,14 @@ droption_t<std::string> op_sched_switch_file(
     "and the sequence proper bracketed by TRACE_MARKER_TYPE_CONTEXT_SWITCH_START and "
     "TRACE_MARKER_TYPE_CONTEXT_SWITCH_END markers.");
 
+droption_t<bool> op_sched_randomize(
+    DROPTION_SCOPE_FRONTEND, "sched_randomize", false,
+    "Pick next inputs randomnly on context switches",
```
abhinav92003, line 902: s/randomnly/randomly/

clients/drcachesim/scheduler/flexible_queue.h
```diff
@@ -96,6 +97,14 @@ class flexible_queue_t {
         return entries_[0]; // Undefined if empty.
     }
 
+    T
+    get_random_entry() // Not const as it change rand_gen's state.
+    {
+        // minstd_rand returns uint_fast32_t.  We do not support get_random_entry()
+        // for queues with >2^32 entries.
+        return entries_[rand_gen_() % size()]; // Undefined if empty.
```
abhinav92003, line 105: I assume we don't want to assert(size() > 0) to keep this path fast?

abhinav92003 (approved):

clients/drcachesim/common/options.cpp
derekbruening: Done.

clients/drcachesim/scheduler/flexible_queue.h
derekbruening: I think that was part of it for top() above too, and that the STL versions do the same thing.  But assert() will go away in release build so I think we can add it.  I added it here and in top() above.

Issue #6648, opened 2024-02-13
Create drmemtrace trimming tool to remove the start and/or end of a trace


derekbruening:
We would like to create a tool to remove the start and/or end of a drmemtrace to trim away undesirable startup or shutdown sequences.

The plan is to do this with the record_filter which first requires #6593.  The base support for removing non-instructions is in place there but we'll need re-chunking.

Once we have that, we just need a simple record filter. My prototype takes in `-keep_{start,end}_timestamp`.

PR #6650, opened 2024-02-13
i#6648 trim tool: Update allasm_x86_64 test trace

derekbruening:
Updates the checked-in allasm_x86_64 trace to have the latest timestamps and markers around system calls.

Updates the invariant checker to get the filetype from the stream on a skip.

Updates the skip test output for the extra records in the new trace.

Issue: #6648


clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -589,7 +589,8 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                         "Missing page size marker");
         report_if_false(
             shard,
-            shard->found_syscall_marker_ ==
+            shard->skipped_instrs_ ||
```
abhinav92003, line 592: Is this relaxation (and the one below) needed in case we skip too many instrs and don't see any syscall number marker? Did that actually happen in a test?

Intuitively it feels like these markers should be frequent enough and in most cases the skip count wouldn't be large enough as to not see any of these markers.

abhinav92003 (approved):

clients/drcachesim/tools/invariant_checker.cpp
derekbruening: Yes, both happened in these skip tests.  Although looking more closely: It looks like it's all about the filetype marker not being visible.  I changed the invariant checker to get the filetype from the stream on a skip, and removed these relaxations.

I'm still a little confused about the comments in the tests cmake for skip vs skip2: it says one will see real headers and the other won't; but neither does.  They skip 62 and 63.  Did we used to have a test trace with a chunk size of 63 or sthg?

derekbruening: Going to live with that mystery.

PR #6653, opened 2024-02-14
Fix TESTANY parameter error in type_filter_t

derekbruening:
Fixes an error where type_filter_t inverted the flag and value parameters to TESTANY.  This was found by inspection and was not observed to have caused any problems (TESTALL would have cared about the order but for TESTANY the order ends up not mattering).

### DIFF
--- a/clients/drcachesim/tools/filter/type_filter.h
+++ b/clients/drcachesim/tools/filter/type_filter.h
@@ -1,5 +1,5 @@
 /* **********************************************************
- * Copyright (c) 2022-2023 Google, Inc.  All rights reserved.
+ * Copyright (c) 2022-2024 Google, Inc.  All rights reserved.
  * **********************************************************/
 
 /*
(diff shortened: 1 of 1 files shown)


clients/drcachesim/tools/filter/type_filter.h
```diff
@@ -1,5 +1,5 @@
 /* **********************************************************
```
abhinav92003, line 1: > (because the flag constant being checked is a single bit).

I think with TESTANY the order doesn't matter, no matter the bit count. For TESTALL, incorrect param order would've been a correctness bug.

abhinav92003 (approved):

clients/drcachesim/tools/filter/type_filter.h
derekbruening: Agreed.  Updated the description.

PR #6657, opened 2024-02-16
Re-enable trailing space check

derekbruening:
The trailing space runsuite check was disabled for the elfutils .patch file commits but was accidentally not re-enabled.

### DIFF
--- a/suite/runsuite.cmake
+++ b/suite/runsuite.cmake
@@ -318,7 +318,7 @@ string(REGEX MATCH "[^\n] \n" match "${diff_contents}")
 if (NOT "${match}" STREQUAL "")
   # Get more context
   string(REGEX MATCH "\n[^\n]+ \n" match "${diff_contents}")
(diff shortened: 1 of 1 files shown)


abhinav92003 (approved):

suite/runsuite.cmake
```diff
@@ -318,7 +318,7 @@ string(REGEX MATCH "[^\n] \n" match "${diff_contents}")
 if (NOT "${match}" STREQUAL "")
   # Get more context
   string(REGEX MATCH "\n[^\n]+ \n" match "${diff_contents}")
-#NOCOMMIT  message(FATAL_ERROR "ERROR: diff contains trailing spaces: ${match}")
```
abhinav92003, line 321: Isn't this one of the keywords that actually prevents submission to master?

derekbruening: Today it only looks for NOCHECKIN IIRC.  But it's tricky here b/c this has to be a word that does not hit any checks, in order to run the tests.  Maybe we should try to create some other way to skip certain up-front checks, if there were some way without touching the code...not sure that exists.

Issue #6662, opened 2024-02-20
Add new record filter tool for public release of traces


edeiana:
We want to create a new tool to filter traces of Google workloads for public release.
The new public Google workload traces will contain more information compared to the previous version, while still preserving confidentiality of Google's IP.

AssadHashmi:
We'd like to raise a point of discussion about the use of `DR_REG_V` as a prefix for virtual registers.
Most Arm programmers' documentation uses `V` for vector registers, e.g.`ADD Vd.4S, Vn.4S, Vm.4S`.

The DynamoRIO user documentation uses R for reasons of convention and generality, e.g.`INSTR_CREATE_add_vector(dc, Rd, Rm, Rn, width)`, see:
https://dynamorio.org/dr__ir__macros__aarch64_8h.html#ad6fa6d2ab7764783481efd209cf11b76

A new or relatively inexperienced user may probably and intuitively but wrongly try something like this first:
```
INSTR_CREATE_add_vector(dc, opnd_create_reg(DR_REG_V0), opnd_create_reg(DR_REG_V1), opnd_create_reg(DR_REG_V2), OPND_CREATE_SINGLE())
```

Rather than the correct:
```
INSTR_CREATE_add_vector(dc, opnd_create_reg(DR_REG_Q0), opnd_create_reg(DR_REG_Q1), opnd_create_reg(DR_REG_Q2), OPND_CREATE_SINGLE())
```
or e.g.
```
INSTR_CREATE_add_vector(dc, opnd_create_reg(DR_REG_D20), opnd_create_reg(DR_REG_D10), opnd_create_reg(DR_REG_D14), OPND_CREATE_HALF())
```

This possible stumbling block is probably less of a problem for scalable vectors, e.g. `INSTR_CREATE_add_sve(dc, Zd, Zn, Zm)`, because the documentation implies `DR_REG_Z`, see
https://dynamorio.org/dr__ir__macros__aarch64_8h.html#a14d0f0b7fad176b301539c3e9254771b

What do you think? Are we worrying unnecessarily?
Is the level of DR knowledge required to use instruction macros in e.g. clients, enough for developers to know which register names/IDs are correct?

derekbruening:
`s/DR_REG_V0/DR_REG_VIRT0/`?

AssadHashmi:
> `s/DR_REG_V0/DR_REG_VIRT0/`?

That's fine.

edeiana:
[...]

PR #6663, opened 2024-02-21
i#6662 public traces, part 2: encoding_filter

edeiana:
New record_filter_t::record_filter_func_t filter, which we call "encodings2regdeps", that modifies the encoding of
trace_entry_t records from a real ISA to the synthetic regdeps ISA.

"encodings2regdeps" can add or remove trace_entry_t records that contain encodings depending on the regdeps 
encoding size. Note that "encodings2regdeps" only changes the encoding of instructions, but it does not adjust 
their length (or changes the instruction PC), hence the output trace will have encoding sizes that do not match the 
instruction length. For this reason we disable the encoding size vs instruction length check in reader_t when the
trace has DR_ISA_REGDEPS encodings.

This filter is part of the "record_filter" tool and can be invoked with:
```
drrun -t drcachesim -simulator_type record_filter -filter_encodings2regdeps -indir path/to/input/trace -outdir path/to/output/trace
```

Issue: #6662


clients/drcachesim/tools/filter/encoding_filter.h
```diff
@@ -0,0 +1,173 @@
...
+    parallel_shard_init(memtrace_stream_t *shard_stream,
+                        bool partial_trace_filter) override
+    {
+        dcontext_.dcontext = dr_standalone_init();
+        return nullptr;
+    }
+
+    bool
+    parallel_shard_filter(trace_entry_t &entry, void *shard_data,
+                          std::vector<trace_entry_t> &last_encoding) override
```
edeiana, line 69: `last_encoding` is all we need from the record_filter, but perhaps we want a better interface for record_filter to "communicate" with its filters?
Or do we want to postpone this decision when we have a need for record_filter to pass more information to its filters?

clients/drcachesim/tools/filter/encoding_filter.h
```diff
@@ -0,0 +1,173 @@
...
+    {
+        /* TODO i#6662: modify trace_entry_t header entry to regdeps ISA, instead of the
+         * real ISA of the incoming trace.
+         */
+
+        /* We have encoding to convert.
+         * Normally the sequence of trace_entry_t(s) looks like:
+         * [encoding,]+ instr_with_PC, [read | write]*
```
derekbruening, line 77: TRACE_MARKER_TYPE_BRANCH_TARGET can be in between encodings and the instr.

clients/drcachesim/tools/filter/encoding_filter.h
```diff
@@ -0,0 +1,173 @@
...
+                error_string_ =
+                    "Failed to decode instruction " + to_hex_string(entry.addr);
+                return false;
+            }
+
+            /* Convert the real ISA instr_t into a regdeps ISA instr_t.
+             */
+            instr_t instr_regdeps;
+            instr_init(dcontext_.dcontext, &instr_regdeps);
+            instr_convert_to_isa_regdeps(dcontext_.dcontext, &instr, &instr_regdeps);
```
derekbruening, line 112: Does this not return a success value to check here?

clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -143,20 +144,36 @@ record_filter_tool_create(const std::string &output_dir, uint64_t stop_timestamp
...
 
-    return new dynamorio::drmemtrace::record_filter_t(output_dir, std::move(filter_funcs),
-                                                      stop_timestamp, verbose);
+    /* If we are changing the encoding of trace_entry_t, we will be generating
+     * discrepancies between encoding size and instruction length.  So, we need to tell
+     * reader_t, which here comes in the form of memref_counter_t, to ignore such
+     * discrepancies.  Note that simulators that deal with these filtered traces will
+     * also have to handle the fact that encoding_size != instruction_length.
+     */
+    bool ignore_encoding_size_vs_instr_length_check = encoding_filter_enabled;
```
edeiana, line 161: Alternative to remove this extra argument to `record_filter_t` constructor: loop through the `filter_funcs` in the `record_filter_t` constructor and set `ignore_encoding_size_vs_instr_length_check` there if we have a `record_filter_func_t` that is `encoding_filter_t`.

clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -143,20 +144,36 @@ record_filter_tool_create(const std::string &output_dir, uint64_t stop_timestamp
...
+    }
+
     // TODO i#5675: Add other filters.
 
-    return new dynamorio::drmemtrace::record_filter_t(output_dir, std::move(filter_funcs),
-                                                      stop_timestamp, verbose);
+    /* If we are changing the encoding of trace_entry_t, we will be generating
+     * discrepancies between encoding size and instruction length.  So, we need to tell
+     * reader_t, which here comes in the form of memref_counter_t, to ignore such
+     * discrepancies.  Note that simulators that deal with these filtered traces will
```
derekbruening, line 158: What about reader_t used by memref_t analysis tools on the final trace: will we get errors on any analysis tool use?

clients/drcachesim/tools/filter/encoding_filter.h
```diff
@@ -0,0 +1,173 @@
...
+
+            /* Compute number of trace_entry_t to contain regdeps ISA encoding.
+             * Each trace_entry_t record can contain 8 byte encoding.
+             */
+            uint trace_entry_encoding_size = (uint)sizeof(entry.addr); /* == 8 */
+            uint regdeps_encoding_size = (uint)(next_pc_regdeps - encoding_regdeps);
+            uint num_regdeps_encoding_entries =
+                ALIGN_FORWARD(regdeps_encoding_size, trace_entry_encoding_size) /
+                trace_entry_encoding_size;
+            last_encoding.resize(num_regdeps_encoding_entries);
```
derekbruening, line 129: record_filter.cpp:643 says "we don't support partial encoding removal": this PR needs to either add that support or update that comment as this resize could result in partial encoding removal if it shrinks the encoding record count, right?

clients/drcachesim/tools/filter/encoding_filter.h
```diff
@@ -0,0 +1,173 @@
...
+ * redefine DR_ISA_REGDEPS alignment requirement here.
+ */
+#define REGDEPS_ALIGN_BYTES 4
+
+#define REGDEPS_MAX_ENCODING_LENGTH 16
+
+namespace dynamorio {
+namespace drmemtrace {
+
+class encoding_filter_t : public record_filter_t::record_filter_func_t {
```
derekbruening, line 53: "encoding_filter_t" sounds like a general encoding modification filter: but instead this is hardcoded to convert to regdeps.  I think it should be renamed to make that clear: regdep_encoding_filter_t or sthg like that.

derekbruening: Or encodings2regdeps (see option name suggestion).

clients/drcachesim/tools/filter/record_filter.cpp
derekbruening: See my other comment: this has to be set for every analyzer on the final trace, right?  Just setting it here is not enough.  I think this flag should be removed from here and we need to instead add a filetype and have the reader relax the check on its own based on that, and then have the reader derivative used here somehow set the filetype instead of having a separate mechanism.

clients/drcachesim/tools/filter/encoding_filter.h
edeiana: Right, we rely on record_filter to write encodings, so even if we store the encodings in encodings2regdeps filter we'd need a way to pass them back to record_filter anyway (right now the only info the filters provide to record_filter is a bool: true keep the record, false drop the record).
When I talked to @abhinav92003 he actually suggested me to pass the `last_encoding`, so keeping it like this for now, at least until we need to pass more info in future filters.

clients/drcachesim/tools/filter/record_filter.cpp
edeiana: Yes, now that we change the file_type in the trace this issue is solved.

clients/drcachesim/tools/filter/encoding_filter.h
edeiana: Added to the comment, but it shouldn't change the logic in the code.

clients/drcachesim/tools/filter/encoding_filter.h
edeiana: It does not, but the encoding to DR_ISA_REGDEPS (which uses the output of `instr_convert_to_isa_regdeps()`) does, so checking that.

clients/drcachesim/tools/filter/record_filter.cpp
edeiana: Solved with changing the trace file_type.

clients/drcachesim/tools/filter/encoding_filter.h
edeiana: From my understanding that refers to cases where a filter removes the encodings of an instruction which will then be encountered later, and when we look for that instruction encoding we won't be able to find them.
In the encodings2regdeps filter we always have the encoding of instructions, even if they can be larger or smaller than the original.
However, I need to investigate this issue more. It's not entirely clear to me what that code is trying to achieve.

clients/drcachesim/tools/filter/encoding_filter.h
edeiana: Going with encodings2regdeps. Done.

clients/drcachesim/CMakeLists.txt
```diff
@@ -392,9 +393,14 @@ add_dependencies(prefetch_analyzer_launcher api_headers)
 add_executable(record_filter_launcher
   tools/record_filter_launcher.cpp
   tests/test_helpers.cpp)
-target_link_libraries(record_filter_launcher drmemtrace_analyzer drmemtrace_record_filter)
+target_link_libraries(record_filter_launcher drmemtrace_analyzer drmemtrace_record_filter
+drmemtrace_raw2trace drcovlib_static drfrontendlib ${static_libc} ${zlib_libs})
```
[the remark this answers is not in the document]
edeiana: Done.

clients/drcachesim/reader/reader.cpp
```diff
@@ -210,7 +210,8 @@ reader_t::process_input_entry()
                 ++cur_instr_count_;
             // Look for encoding bits that belong to this instr.
             if (last_encoding_.size > 0) {
-                if (last_encoding_.size != cur_ref_.instr.size) {
+                if (!ignore_encoding_size_vs_instr_length_check_ &&
```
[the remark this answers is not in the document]
edeiana: Yes. Done.

clients/drcachesim/common/options.cpp
```diff
@@ -972,6 +972,11 @@ droption_t<std::string>
                            "Comma-separated integers for marker types to remove. "
                            "See trace_marker_type_t for the list of marker types.");
 
+droption_t<bool> op_encoding_filter_enabled(
+    DROPTION_SCOPE_FRONTEND, "encoding_filter_enabled", false,
```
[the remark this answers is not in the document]
edeiana: Going with encodings2regdeps. Done.

clients/drcachesim/tools/filter/record_filter.h
```diff
@@ -53,6 +53,11 @@
 namespace dynamorio {
 namespace drmemtrace {
 
+/* Utility function.
+ */
+bool
+is_any_instr_type(trace_type_t type);
```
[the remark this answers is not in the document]
edeiana: Done.

clients/drcachesim/tools/filter/record_filter.cpp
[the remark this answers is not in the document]
edeiana: I am able to run opcode_mix on a single-threaded trace.
But it's failing on a core-sharded one.
Needs more investigation.

clients/drcachesim/tools/opcode_mix.cpp
```diff
@@ -232,12 +235,19 @@ opcode_mix_t::parallel_shard_memref(void *shard_data, const memref_t &memref)
         instr_init(dcontext_.dcontext, &instr);
         app_pc next_pc =
             decode_from_copy(dcontext_.dcontext, decode_pc, trace_pc, &instr);
-        if (next_pc == NULL || !instr_valid(&instr)) {
```
edeiana, line 235: Not sure if we should "fix" `instr_valid()` to work with DR_ISA_REGDEPS instructions (more overhead? unexpected behavior?) or if we should "fix" the analyzers instead like I did here.

clients/drcachesim/tools/opcode_mix.cpp
```diff
@@ -316,16 +326,27 @@ opcode_mix_t::print_results()
...
         }
     }
     std::cerr << TOOL_NAME << " results:\n";
     std::cerr << std::setw(15) << total.instr_count << " : total executed instructions\n";
     std::vector<std::pair<int, int64_t>> sorted(total.opcode_counts.begin(),
                                                 total.opcode_counts.end());
     std::sort(sorted.begin(), sorted.end(), cmp_val);
-    for (const auto &keyvals : sorted) {
-        std::cerr << std::setw(15) << keyvals.second << " : " << std::setw(9)
-                  << decode_opcode_name(keyvals.first) << "\n";
```
edeiana, line 328: Similarly to https://github.com/DynamoRIO/dynamorio/pull/6663/files#r1584655333 here.
Should we add a name to OP_INVALID (is the failure on decode_opcode_name(OP_INVALID) intended?) or work around it as I did here?

suite/tests/CMakeLists.txt
```diff
@@ -4674,6 +4674,17 @@ if (BUILD_CLIENTS)
...
 
+    if (X86 AND X64 AND ZLIB_FOUND)
+        set(testname "tool.record_filter_encodings2regdeps")
+        torun_record_filter("${testname}" ${ci_shared_app}
+          "record_filter_encodings2regdeps"
+          # We assume the app name starts with "s" here to avoid colliding with
+          # our output dir, while still letting the single precmd remove both.
+          "${drcachesim_path}@-simulator_type@record_filter@-encodings2regdeps@-indir@${testname}.s*.dir/trace@-core_sharded@-cores@4@-outdir@${testname}.filtered.dir"
+          # We run the opcode_mix analyzer to test econdings2regdeps filtered traces.
+          "opcode_mix")
```
edeiana, line 4685: Is this the right way to run "opcode_mix" on a trace filtered by record_filter (with encodings2regdeps)?

clients/drcachesim/tools/filter/encoding_filter.h
derekbruening: Yes that does sound right: "partial encoding removal" means removing some encodings from part of the trace but leaving others in other parts, which gets hard to support with the chunking.

clients/drcachesim/tools/opcode_mix.cpp
derekbruening: I think instr_valid() should return true for a decoded regdep instr.  OP_INVALID seems incorrect for any decoded instr.  So the regdep decoder should use OP_UNDECODED -- or will that trigger up-decode requests?  Maybe we can add logic to those requests.

clients/drcachesim/tools/filter/encoding_filter.h
derekbruening: I think it may be worth making a struct of info with last_encoding inside it so it can be changed later without breaking the interface.

abhinav92003: +1 on using a struct.

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -170,6 +172,17 @@ struct test_case_t {
...
 
+struct test_entries_t {
+    /* Represents entry in the input trace.
+     */
+    trace_entry_t entry_test;
+    /* Represents corresponding expected entry in the output trace.
+     * XXX i#6662: consider making this into a std::vector<trace_entry_t> to test cases
+     * where DR_ISA_REGDEPS encoding is longer or shorter than entry_test.
+     */
+    trace_entry_t entry_expected;
```
derekbruening, line 183: But the tests are all using test_case_t.  Best to use the same instead of making a similar new scheme.  I don't see anything here that doesn't fit test_case_t.

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -287,6 +300,123 @@ process_entries_and_check_result(test_record_filter_t *record_filter,
     return true;
 }
 
+bool
+process_entries_and_check_result_against_ground_truth(
```
derekbruening, line 304: Please re-use the existing process routine which all the existing tests use.  I don't think there's anything special about this filter that isn't already handled?  If so it should be possible to add it to the existing routine?  That seems much better than this duplication.

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -1020,10 +1151,10 @@ test_null_filter()
...
     std::vector<std::unique_ptr<record_filter_func_t>> filter_funcs;
     filter_funcs.push_back(std::move(null_filter));
-    // We use a very small stop_timestamp for the record filter. This is to verify that
-    // we emit the TRACE_MARKER_TYPE_FILTER_ENDPOINT marker for each thread even if it
-    // starts after the given stop_timestamp. Since the stop_timestamp is so small, all
-    // other entries are expected to stay.
+    // We use a very small stop_timestamp for the record filter. This is to verify
+    // that we emit the TRACE_MARKER_TYPE_FILTER_ENDPOINT marker for each thread even
+    // if it starts after the given stop_timestamp. Since the stop_timestamp is so
+    // small, all other entries are expected to stay.
```
derekbruening, line 1157: This looks like a mistaken change that should be reverted (probably a messed-up brace with a clang-format that was undone?)

clients/drcachesim/tools/filter/encodings2regdeps.h
```diff
@@ -0,0 +1,192 @@
...
+                instr_encode(dcontext_.dcontext, &instr_regdeps, encoding_regdeps);
+            instr_free(dcontext_.dcontext, &instr_regdeps);
+            if (next_pc_regdeps == NULL) {
+                error_string_ =
+                    "Failed to encode regdeps instruction " + to_hex_string(entry.addr);
+                return false;
+            }
+
+            /* Compute number of trace_entry_t to contain regdeps ISA encoding.
+             * Each trace_entry_t record can contain 8 byte encoding.
```
derekbruening, line 141: pointer-sized, not 8: 4 for 32-bit.

clients/drcachesim/tools/filter/encodings2regdeps.h
```diff
@@ -0,0 +1,192 @@
...
+             */
+            uint trace_entry_encoding_size = (uint)sizeof(entry.addr); /* == 8 */
```
derekbruening, line 143: Ditto.

clients/drcachesim/tools/filter/encodings2regdeps.h
```diff
@@ -0,0 +1,192 @@
...
+                trace_entry_encoding_size;
+            last_encoding.resize(num_regdeps_encoding_entries);
+
+            /* Copy regdeps ISA encoding, splitting it among the last_encoding
+             * trace_entry_t records.
+             */
+            uint regdeps_encoding_offset = 0;
+            for (trace_entry_t &encoding_entry : last_encoding) {
+                encoding_entry.type = TRACE_TYPE_ENCODING;
+                uint size = regdeps_encoding_size < trace_entry_encoding_size
```
derekbruening, line 156: Use std::min

clients/drcachesim/tools/filter/encodings2regdeps.h
```diff
@@ -0,0 +1,192 @@
...
+                    ? regdeps_encoding_size
+                    : trace_entry_encoding_size;
+                encoding_entry.size = (unsigned short)size;
```
derekbruening, line 159: Prefer C++-style casts

clients/drcachesim/tools/filter/encodings2regdeps.h
```diff
@@ -0,0 +1,192 @@
...
+                error_string_ =
+                    "Failed to encode regdeps instruction " + to_hex_string(entry.addr);
+                return false;
+            }
+
+            /* Compute number of trace_entry_t to contain regdeps ISA encoding.
+             * Each trace_entry_t record can contain 8 byte encoding.
+             */
+            uint trace_entry_encoding_size = (uint)sizeof(entry.addr); /* == 8 */
+            uint regdeps_encoding_size = (uint)(next_pc_regdeps - encoding_regdeps);
```
derekbruening, line 144: Prefer C++-style casts

clients/drcachesim/tools/filter/encodings2regdeps.h
derekbruening, line 143: Prefer C++-style casts

clients/drcachesim/tools/filter/encodings2regdeps.h
```diff
@@ -0,0 +1,192 @@
...
+            for (trace_entry_t &encoding_entry : last_encoding) {
+                encoding_entry.type = TRACE_TYPE_ENCODING;
+                uint size = regdeps_encoding_size < trace_entry_encoding_size
+                    ? regdeps_encoding_size
+                    : trace_entry_encoding_size;
+                encoding_entry.size = (unsigned short)size;
+                memset(encoding_entry.encoding, 0, trace_entry_encoding_size);
+                memcpy(encoding_entry.encoding,
+                       encoding_regdeps + regdeps_encoding_offset, encoding_entry.size);
+                regdeps_encoding_size -= trace_entry_encoding_size;
```
derekbruening, line 163: encoding_entry.size should be subtracted

clients/drcachesim/tools/filter/encodings2regdeps.h
```diff
@@ -0,0 +1,192 @@
...
+
+#include "record_filter.h"
+#include "trace_entry.h"
+#include "utils.h"
+
+#include <cstring>
+#include <vector>
+
+/* We are not exporting the defines in core/ir/isa_regdeps/encoding_common.h, so we
+ * redefine DR_ISA_REGDEPS alignment requirement here.
```
derekbruening, line 44: Hmm wondering whether we should export the alignment and max encoding?

clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -143,6 +137,20 @@ record_filter_tool_create(const std::string &output_dir, uint64_t stop_timestamp
...
                                                          trim_after_timestamp)));
     }
+    if (encodings2regdeps) {
+        /* If we are changing the encoding of trace_entry_t with encodings2regdeps, we
+         * will be generating discrepancies between encoding size and instruction length.
+         * So, we need to tell reader_t, which here comes in the form of memref_counter_t,
+         * to ignore such discrepancies. We do so by adding OFFLINE_FILE_TYPE_ARCH_REGDEPS
+         * to the file type of the filtered trace. Note that simulators that deal with
+         * these filtered traces will also have to handle the fact that
+         * encoding_size != instruction_length.
```
derekbruening, line 147: This comment belongs in the filter, not here, right?

clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -158,6 +166,15 @@ record_filter_t::record_filter_t(
     , stop_timestamp_(stop_timestamp)
     , verbosity_(verbose)
 {
+    /* Check if encodings2regdeps filter is present.
```
derekbruening, line 169: See below: I think this should be removed in favor of a general filetype interface.

clients/drcachesim/tools/filter/record_filter.h
```diff
@@ -248,11 +249,13 @@ class record_filter_t : public record_analysis_tool_t {
...
-        if (stop_timestamp_ != 0) {
+        if (stop_timestamp_ != 0)
             filetype |= OFFLINE_FILE_TYPE_BIMODAL_FILTERED_WARMUP;
-        }
-        if (shard_type_ == SHARD_BY_CORE) {
+        if (shard_type_ == SHARD_BY_CORE)
             filetype |= OFFLINE_FILE_TYPE_CORE_SHARDED;
+        if (encodings2regdeps_) {
+            filetype &= ~OFFLINE_FILE_TYPE_ARCH_ALL;
+            filetype |= OFFLINE_FILE_TYPE_ARCH_REGDEPS;
```
derekbruening, line 258: This is in 2 different places now.  Let's make a filetype interface so any filter modifying the filetype will work, and remove this encodings2regdeps_.  Does that require a new API routine asking for the filetype from the filter?  That seems cleaner than this hardcoding for this one filter.

clients/drcachesim/tools/opcode_mix.cpp
derekbruening: We should make it work.  See the OP_UNDECODED suggestion.  Use OP_nop??  No; I would still think making OP_UNDECODED work seems best.

suite/tests/CMakeLists.txt
derekbruening: Looks like it.  I don't think it needs the quotes.

clients/drcachesim/tools/filter/encoding_filter.h
edeiana: Done. Called: record_filter_info_t. As of now it only has a pointer to the `last_encoding` vector.

clients/drcachesim/tools/filter/encoding_filter.h
edeiana: Ok, in that case we don't need to add that support yet.
I clarified this in the two comments mentioning lack of support for partial encodings removal.

clients/drcachesim/tools/filter/record_filter.cpp
edeiana: Done.

clients/drcachesim/tools/filter/record_filter.h
edeiana: Done. Added `add_to_filetype()` routine in `record_filter_func_t`. By default it doesn't change anything. The `encodings2regdeps_t` filter overrides it removing arch info and adding "arch regdeps" to the file_type.

clients/drcachesim/tools/filter/record_filter.cpp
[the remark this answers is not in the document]
edeiana: This has been fixed.

clients/drcachesim/tools/opcode_mix.cpp
edeiana: OP_UNDECODED seems to work. Done.

clients/drcachesim/tools/opcode_mix.cpp
edeiana: OP_UNDECODED seems to work. Done.
The only change I had to make to the opcode_mix analyzer was changing the ISA to DR_ISA_REGDEPS id the filetype of the trace is regdeps.

suite/tests/CMakeLists.txt
edeiana: Yes, it works fine now.

clients/drcachesim/CMakeLists.txt
```diff
@@ -392,9 +393,14 @@ add_dependencies(prefetch_analyzer_launcher api_headers)
 add_executable(record_filter_launcher
   tools/record_filter_launcher.cpp
   tests/test_helpers.cpp)
-target_link_libraries(record_filter_launcher drmemtrace_analyzer drmemtrace_record_filter)
+target_link_libraries(record_filter_launcher drmemtrace_analyzer drmemtrace_record_filter
+  drmemtrace_raw2trace drcovlib_static drfrontendlib ${static_libc} ${zlib_libs})
```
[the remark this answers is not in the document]
edeiana: They are not, I forgot to remove them, thanks for pointing them out!
Removed now.

clients/drcachesim/common/trace_entry.h
```diff
@@ -665,6 +665,17 @@ type_is_instr(const trace_type_t type)
         type == TRACE_TYPE_INSTR_UNTAKEN_JUMP;
 }
 
+/**
+ * Returns whether the type represents an instruction.
```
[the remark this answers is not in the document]
edeiana: Done.

clients/drcachesim/common/options.cpp
```diff
@@ -972,6 +972,12 @@ droption_t<std::string>
                            "Comma-separated integers for marker types to remove. "
                            "See trace_marker_type_t for the list of marker types.");
 
+droption_t<bool> op_encodings2regdeps(
+    DROPTION_SCOPE_FRONTEND, "encodings2regdeps", false,
+    "Enable converting the encoding of instructions to synthetic ISA DR_ISA_REGDEPS.",
+    "When present, converts the encoding of instructions from a real ISA to the "
```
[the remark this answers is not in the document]
edeiana: Added filter_ prefix and xref in a comment. Done.

clients/drcachesim/common/trace_entry.h
```diff
@@ -968,19 +979,27 @@ typedef enum {
      * Each trace shard represents one core and contains interleaved software threads.
      */
     OFFLINE_FILE_TYPE_CORE_SHARDED = 0x10000,
+    /**
+     * Trace filtered by the record_filter tool using encodings2regdeps.
+     */
+    OFFLINE_FILE_TYPE_ARCH_REGDEPS = 0x20000,
```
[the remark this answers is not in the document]
edeiana: Done.

clients/drcachesim/reader/reader.cpp
```diff
@@ -210,7 +210,8 @@ reader_t::process_input_entry()
                 ++cur_instr_count_;
             // Look for encoding bits that belong to this instr.
             if (last_encoding_.size > 0) {
-                if (last_encoding_.size != cur_ref_.instr.size) {
+                if (!TESTANY(OFFLINE_FILE_TYPE_ARCH_REGDEPS, filetype_) &&
```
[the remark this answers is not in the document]
edeiana: Done.

clients/drcachesim/tools/filter/encodings2regdeps.h
edeiana: Ah, right, thanks. Fixed.

clients/drcachesim/tools/filter/encodings2regdeps.h
edeiana: Fixed.

clients/drcachesim/tools/filter/encodings2regdeps.h
edeiana: Right :) thanks! Done.

clients/drcachesim/tools/filter/encodings2regdeps.h
edeiana: Right, thanks. Fixed.

clients/drcachesim/tools/filter/record_filter.cpp
edeiana: Yes, makes more sense to have it in encodings2regdeps.h. Moved there. Thanks!

clients/drcachesim/tools/filter/encodings2regdeps.h
edeiana: We could, but I wonder what would be a good place? encode_api.h? (Although it doesn't have any other "exported defines". I think we usually redefine these constants where we need them, I remember seeing an x86 max_encoding_length of 17 in several places. But it would be nice to have single point of control.)

clients/drcachesim/tools/filter/encodings2regdeps.h
edeiana: Done.

clients/drcachesim/tools/filter/encodings2regdeps.h
edeiana: Done.

clients/drcachesim/tools/filter/encodings2regdeps.h
edeiana: Done.

clients/drcachesim/tests/record_filter_unit_tests.cpp
edeiana: Ah yes, clang-format changed that. Reverted back, thanks!

clients/drcachesim/tools/filter/record_filter.h
```diff
@@ -96,7 +96,8 @@ class record_filter_t : public record_analysis_tool_t {
          * An error is indicated by setting error_string_ to a non-empty value.
          */
         virtual bool
-        parallel_shard_filter(trace_entry_t &entry, void *shard_data) = 0;
+        parallel_shard_filter(trace_entry_t &entry, void *shard_data,
+                              std::vector<trace_entry_t> &last_encoding) = 0;
```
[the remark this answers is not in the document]
edeiana: Done.
For the version: I updated it in CMakeLists.txt, .github/workflows/ci-docs.yml, and .github/workflows/ci-package.yml. There is no other place I missed right?

clients/drcachesim/tests/record_filter_unit_tests.cpp
edeiana: I misunderstood how test_case_t works. Fixed using existing infrastructure.

clients/drcachesim/tests/record_filter_unit_tests.cpp
edeiana: Done. https://github.com/DynamoRIO/dynamorio/pull/6663#discussion_r1586255377

api/docs/release.dox
```diff
@@ -150,6 +150,10 @@ changes:
  - Changed the type of the AArch64 #dr_mcontext_t members svep and ffr to #dr_svep_t.
    This breaks binary compatibility with clients that were built against versions of
    DynamoRIO before this change.
+ - Changed dynamorio::drmemtrace::record_filter_t::record_filter_func_t
+   parallel_shard_filter() interface.  Added a new parameter of type
+   dynamorio::drmemtrace::record_filter_t::record_filter_info_t that allows record_filter
```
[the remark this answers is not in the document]
derekbruening: Looks like the record filter headers are not exported: not put into an export dir and thus never seen by doxygen.  @abhinav92003 was that a deliberate decision?  The record filter interfaces have Doxygen comments, implying some kind of export since we generally don't Doxygen-comment internal interfaces and don't run Doyxgen for anything except the public html.

clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -658,7 +650,10 @@ record_filter_t::process_delayed_encodings(per_shard_t *per_shard, trace_entry_t
     } else if (TESTANY(OFFLINE_FILE_TYPE_ENCODINGS, per_shard->filetype)) {
         // Output if we have encodings that haven't yet been output, and
         // there is no filter removing all encodings (we don't support
-        // partial encoding removal).
+        // partial encoding removal). Note that filters that modify encodings (even if
+        // they add or remove trace_entry_t records) do not incurr in this problem and we
```
derekbruening, line 654: s/incurr in/incur/

core/ir/x86/decode_table.c
```diff
@@ -75,7 +75,7 @@
 const instr_info_t * const op_instr[] =
 {
     /* OP_INVALID */   NULL,
-    /* OP_UNDECODED */ NULL,
+    /* OP_UNDECODED */ (instr_info_t[]){{1, 0xffffff, 0, "<undecoded>", 0, 0, 0, 0, 0, 0, 0, 0}},
```
derekbruening, line 78: The 1 should be OP_UNDECODED right?

derekbruening, line 78: I wonder if some code somewhere takes op_instr[op] as non-NULL to mean something.

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -287,6 +289,61 @@ process_entries_and_check_result(test_record_filter_t *record_filter,
...
+        { { TRACE_TYPE_PID, 0, { 0x5 } }, true, { true } },
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_CACHE_LINE_SIZE, { 0x6 } },
+          true,
+          { true } },
+        // Unit header.
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_TIMESTAMP, { 0x7 } }, true, { true } },
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_CPU_ID, { 0x8 } }, true, { true } },
+        // Encoding, modified by the record_filter encodings2regdeps.
+        { { TRACE_TYPE_ENCODING, 4, { 0xe78948 } }, true, { false } },
+        { { TRACE_TYPE_ENCODING, 8, { 0x0006090600010011 } }, false, { true } },
```
derekbruening, line 322: A test with duplicated encodings across a chunk boundary would be good to ensure the chunk code uses the modified and not original encoding.

clients/drcachesim/tools/filter/record_filter.h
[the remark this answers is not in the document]
derekbruening: Look at the previous PR to update them to see all the files it touched.  But wait for @abhinav92003 's reply to https://github.com/DynamoRIO/dynamorio/pull/6663#discussion_r1586429106 b/c if it wasn't really considered a public interface maybe we don't need a version bump?!?

api/docs/release.dox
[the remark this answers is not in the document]
abhinav92003: Hmm. I don't remember noting the extra steps for Doxygen in other targets. Maybe I missed it.

clients/drcachesim/tools/filter/record_filter.h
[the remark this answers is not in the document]
abhinav92003: record_filter_t should be a public interface.

api/docs/release.dox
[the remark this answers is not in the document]
abhinav92003: Doesn't the add_exported_library do all that's needed? https://github.com/DynamoRIO/dynamorio/blob/173fc643fdd604ac06d3688a87677d2211a0f74a/clients/drcachesim/CMakeLists.txt#L194

Oh you mean record_filter.h is missing from there.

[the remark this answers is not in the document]
derekbruening: Doxygen looks at `<builddir>clients/include/drmemtrace/` for the headers.

core/ir/x86/decode_table.c
edeiana: Yes, but OP_UNDECODED it's defined below and I didn't want to overturn this file too much just for that.
I can move `op_instr[]` down below though.

edeiana: I wonder the same.
Perhaps we should just add an `if (opcode == OP_UNDECODED) return "<undecoded>"` in `decode_opcode_name()` before we query `op_instr[]` and get NULL?

clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -605,7 +589,10 @@ record_filter_t::process_chunk_encodings(per_shard_t *per_shard, trace_entry_t &
         // XXX: What if there is a filter removing all encodings but only
         // to the stop point, so a partial remove that does not change
         // the filetype?  For now we do not support that, and we re-add
-        // encodings at chunk boundaries regardless.
+        // encodings at chunk boundaries regardless. Note that filters that modify
+        // encodings (even if they add or remove trace_entry_t records) do not incurr in
```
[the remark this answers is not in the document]
edeiana: Fixed, thanks.

clients/drcachesim/tools/filter/record_filter.cpp
edeiana: Fixed, thanks.

clients/drcachesim/common/options.cpp
```diff
@@ -972,11 +972,15 @@ droption_t<std::string>
...
+ * record_filter partition. For now we add the filter_ prefix to options that should be
+ * used in conjunction with record_filter.
+ */
 droption_t<bool> op_encodings2regdeps(
-    DROPTION_SCOPE_FRONTEND, "encodings2regdeps", false,
+    DROPTION_SCOPE_FRONTEND, "filter_encodings2regdeps", false,
     "Enable converting the encoding of instructions to synthetic ISA DR_ISA_REGDEPS.",
-    "When present, converts the encoding of instructions from a real ISA to the "
-    "DR_ISA_REGDEPS synthetic ISA.");
+    "This option is intended to be used with record_filter. When present, it converts "
```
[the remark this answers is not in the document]
edeiana: Done.

clients/drcachesim/common/trace_entry.h
```diff
@@ -666,8 +666,9 @@ type_is_instr(const trace_type_t type)
 }
 
 /**
- * Returns whether the type represents an instruction.
- * Includes TRACE_TYPE_INSTR_NO_FETCH and TRACE_TYPE_INSTR_MAYBE_FETCH.
+ * Returns whether \p type represents any type of instruction record whether an
+ * instruction fetch or operation hint. This is a superset of type_is_instr() and includes
+ * TRACE_TYPE_INSTR_NO_FETCH.
```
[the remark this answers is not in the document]
edeiana: Done.

clients/drcachesim/common/trace_entry.h
```diff
@@ -980,9 +978,15 @@ typedef enum {
      */
     OFFLINE_FILE_TYPE_CORE_SHARDED = 0x10000,
     /**
-     * Trace filtered by the record_filter tool using encodings2regdeps.
+     * Trace filtered by the record_filter tool using -filter_encodings2regdeps.
```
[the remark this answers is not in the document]
edeiana: Done.

clients/drcachesim/reader/reader.cpp
```diff
@@ -210,6 +210,10 @@ reader_t::process_input_entry()
                 ++cur_instr_count_;
             // Look for encoding bits that belong to this instr.
             if (last_encoding_.size > 0) {
+                /* OFFLINE_FILE_TYPE_ARCH_REGDEPS traces have encodings with
+                 * size != ifetch. It's a design choice, not an error, hence
+                 * we avoid this sanity check.
+                 */
                 if (!TESTANY(OFFLINE_FILE_TYPE_ARCH_REGDEPS, filetype_) &&
                     (last_encoding_.size != cur_ref_.instr.size)) {
```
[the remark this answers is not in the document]
edeiana: Done.

clients/drcachesim/tools/filter/record_filter.h
```diff
@@ -61,6 +61,13 @@ namespace drmemtrace {
  */
 class record_filter_t : public record_analysis_tool_t {
 public:
+    /**
+     * Interface for the record_filter to share data with its filters.
+     */
+    struct record_filter_info_t {
+        std::vector<trace_entry_t> *last_encoding;
```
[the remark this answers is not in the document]
edeiana: Done.

clients/drcachesim/tools/filter/record_filter.h
```diff
@@ -115,6 +124,17 @@ class record_filter_t : public record_analysis_tool_t {
             return error_string_;
         }
 
+        /**
+         * If a filter modifies the file type of a trace, its changes should be made here,
+         * so they are visible to the record_filter even if the #trace_entry_t containing
+         * the file type marker is not modified.
+         */
+        virtual uint64_t
+        add_to_filetype(uint64_t filetype)
```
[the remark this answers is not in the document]
edeiana: Done.

core/ir/x86/decode_table.c
derekbruening: The enum value is defined in opnd_api.h which must be included before this right?

api/docs/release.dox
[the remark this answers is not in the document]
edeiana: I think `install_client_nonDR_header` puts files in `<builddir>/clients/include/drmemtrace/`, so I added there.
It works.
Done.

core/ir/x86/decode_table.c
derekbruening: How about putting that special case only if op_instr[] is NULL: then the branch added to every call is at least checking for NULL which seems worthwhile.

clients/drcachesim/tests/record_filter_unit_tests.cpp
derekbruening: Include repeat instances of the same instr too to check that the filter messing with last_encoding isn't breaking basic things like when encodings are output.

core/ir/x86/decode_table.c
edeiana: Yes, it's in opcode_api.h, I'm not sure why I was getting an error from clangd when using the enum name.
s/1/OP_UNDECODED
Done.

clients/drcachesim/tools/filter/record_filter.h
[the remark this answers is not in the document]
edeiana: Ok, record_filter_func_t and record_filter_t are visible by doxygen as well now.
So, it seems we need to bump the version up.

core/ir/x86/decode_table.c
edeiana: Yeah, much better that way.
Done.

edeiana: No more relevant. https://github.com/DynamoRIO/dynamorio/pull/6663#discussion_r1586981783

clients/drcachesim/tests/record_filter_unit_tests.cpp
edeiana: Done.

api/docs/release.dox
```diff
@@ -226,6 +230,10 @@ Further non-compatibility-affecting changes include:
    purpose of preserving register dependencies.
  - Added instr_convert_to_isa_regdeps() API that converts an #instr_t from a real ISA
    (e.g., #DR_ISA_AMD64) to the #DR_ISA_REGDEPS synthetic ISA.
+ - Added encodings2regdeps_t filter to #dynamorio::drmemtrace::record_filter_t to generate
+   #DR_ISA_REGDEPS traces.
+ - Added #dynamorio::drmemtrace::OFFLINE_FILE_TYPE_ARCH_REGDEPS file type for
+   #DR_ISA_REGDEPS traces.
```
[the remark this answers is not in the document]
derekbruening: Yes.

clients/drcachesim/CMakeLists.txt
```diff
@@ -192,12 +192,14 @@ add_exported_library(drmemtrace_simulator STATIC
...
 add_exported_library(drmemtrace_record_filter STATIC
+  tools/filter/record_filter.h
   tools/filter/record_filter.cpp
   tools/filter/cache_filter.h
   tools/filter/cache_filter.cpp
   tools/filter/type_filter.h
+  tools/filter/encodings2regdeps.h
   tools/filter/null_filter.h)
-target_link_libraries(drmemtrace_record_filter drmemtrace_simulator)
+target_link_libraries(drmemtrace_record_filter drmemtrace_simulator dynamorio)
```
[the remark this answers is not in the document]
derekbruening: configure_DynamoRIO_standalone() is what opcode_mix, view, etc. are using.

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -321,6 +321,20 @@ test_encodings2regdeps_filter()
         { { TRACE_TYPE_ENCODING, 4, { 0xe78948 } }, true, { false } },
         { { TRACE_TYPE_ENCODING, 8, { 0x0006090600010011 } }, false, { true } },
         { { TRACE_TYPE_INSTR, 3, { 0x7f6fdd3ec360 } }, true, { true } },
+        // Same instr same chunk.
+        { { TRACE_TYPE_INSTR, 3, { 0x7f6fdd3ec360 } }, true, { true } },
+        // Unit header.
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_TIMESTAMP, { 0xc } }, true, { true } },
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_CPU_ID, { 0xd } }, true, { true } },
+        // Duplicated encoding across chunk boundary.
```
derekbruening, line 329: There's no chunk boundary here: I see only a timestamp+cpuid.  You need a chunk instr count, chunk footer, and record ordinal markers.  See test_chunk_update().

derekbruening: A chunk here is a compress storage (zipfile) subcomponent allowing fast seeking.

derekbruening: Chunk boundaries must not only be at precise instruction counts but include record count hints, so the record_filter framework has to update them for pretty much any filter doing anything.

edeiana: So the bug had nothing to do with record_filter and its encodings2regdeps filter, it was in the test case.
Specifically, I wasn't setting `write_archive=true` in the record_filter constructor, so that's why it wasn't generating new chunks.
The test case has been fixed and now it passes (I'm confident the expected output is correct, and I talked to @abhinav92003  about it as well, but an additional pair of eyes to check that doesn't hurt).

clients/drcachesim/common/trace_entry.h
```diff
@@ -476,8 +476,11 @@ typedef enum {
     TRACE_MARKER_TYPE_CHUNK_INSTR_COUNT,
 
     /**
-     * Marks the end of a chunk.  The final chunk does not have such a marker
-     * but instead relies on the #TRACE_TYPE_FOOTER entry.
+     * Marks the end of a chunk.  The value of this marker is the chunk ordinal, which
+     * is an increasing counter set by #dynamorio::drmemtrace::record_filter_t starting
```
derekbruening, line 480: No, for 99% of traces record_filter is *not* involved at all (raw2trace inserts this during normal post-processing).  Please remove this part.

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -314,27 +317,46 @@ test_encodings2regdeps_filter()
...
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_CHUNK_FOOTER, { 1 } }, true, { true } },
+
+        // Chunk 3.
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_RECORD_ORDINAL, { 0xe } },
+          true,
+          { true } },
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_TIMESTAMP, { 0x7 } }, true, { true } },
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_CPU_ID, { 0x8 } }, true, { true } },
+        // Same instr in different chunk. We expect record_filter to add
+        // ENCODING_REGDEPS_ISA here (not ENCODING_REAL_ISA).
```
derekbruening, line 356: ?  I don't understand this: where's the real encoding in the input?  Looks like you have an ill-formed input trace here with a missing encoding.  If that was the real intent, that has to be clearly commented: but I don't think that's the best way to go about testing.  To check changing chunks, add a trim filter or instruction removal filter or something to have the output be different from the input.

derekbruening (approved):

clients/drcachesim/common/trace_entry.h
edeiana: Removed.

clients/drcachesim/tests/record_filter_unit_tests.cpp
edeiana: Not my intent, that is indeed an ill-formed trace. I didn't realize `test_chunk_update()` was using a trim filter for that.
Fixed, and added a couple more test cases where the number of trace_entry_t records for encodings differ between real ISA and regdeps ISA.

Issue #6643, opened 2024-02-09
Support drmemtrace trace analyzer interval analysis by instr count


abhinav92003:
Today the drmemtrace trace analyzer supports trace interval analysis where intervals are defined in microseconds of trace time. For some use cases, it'll be useful to also support defining intervals by instr count.

The only limitation we're aware of for #instr interval is that merging them across shards is hard. It would need some sort of pre-processing step to figure out where the whole-trace interval boundaries lie for each shard. For now, we plan to support per-shard #instr intervals only.

PR #6664, opened 2024-02-21
i#6643: Add finalize_interval_snapshots API to analysis_tool_t

abhinav92003:
Adds a new finalize_interval_snapshots API to analysis_tool_t. This is invoked with the list of shard-local interval state snapshots for each shard separately in the parallel mode, or the whole-trace ones in serial mode. This allows the tool the opportunity to make any required holistic adjustments to the snapshots since now all of the snapshots can be observed together; e.g., computing diffs with the prior snapshot. This is invoked before the shard-local snapshots are possibly combined to create whole-trace snapshots, and before the snapshots are passed to print_interval_results.

Adds unit tests for the new API to the existing tool.drcacheoff.trace_interval_analysis_unit_tests tests.

Refactors some existing code to accumulate the interval snapshots in an std::vector instead of an std::queue. This adds some more complexity to the merge_shard_interval_results implementation, but is better because now we have more usages where an std::vector is needed (and we want to avoid a back-and-forth conversion between a queue and a vector).

Augments various documentation to provide more details about intended usages of the interval APIs. Notably: documented the new finalize_interval_snapshots API, and that modifications made after the combine_interval_snapshots API has been invoked do not have any effect).

Issue: #6643, #6020


clients/drcachesim/analysis_tool.h
```diff
@@ -277,6 +283,37 @@ template <typename RecordType> class analysis_tool_tmpl_t {
...
+     * If some snapshots are removed, release_interval_snapshot() will not be invoked
+     * for them and the tool is responsible to de-allocate the resources. Adding new
+     * snapshots to the list is undefined behavior; tools should operate only on the
+     * provided snapshots which were generated in prior generate_*interval_snapshot
+     * calls.
+     *
+     * Tools should not modify any data in the base \p interval_state_snapshot_t
+     * struct.
+     *
+     * In the parallel mode, this is invoked with the list of shard-local snapshots
```
brettcoon, line 302: I think it's worth clarifying that in parallel mode this method will be invoked once per shard.  Maybe just change the wording a bit to "...this is invoked on each list of shard-local snapshots...".

clients/drcachesim/analysis_tool.h
```diff
@@ -455,12 +500,21 @@ template <typename RecordType> class analysis_tool_tmpl_t {
      * \p interval_id is a positive ordinal of the trace interval that just ended.
      * Trace intervals have a length equal to the \p -interval_microseconds specified
      * to the framework. Trace intervals are measured using the value of the
-     * #TRACE_MARKER_TYPE_TIMESTAMP markers. The provided \p interval_id
-     * values will be monotonically increasing but may not be continuous,
-     * i.e. the tool may not see some \p interval_id if the trace shard did not
-     * have any activity in that interval.
+     * #TRACE_MARKER_TYPE_TIMESTAMP markers.  Alternatively, trace intervals may be
+     * defined in terms of the \p -interval_instr_count count of shard-local instrs
+     * specified to the framework.
```
brettcoon, line 505: Much of this comment appears identical to the comment text above, so my same comments apply about describing both interval modes together.  Also, duplicate descriptions like this are problematic, since it's so easy to update one and forget the other, and then end up with them out of sync.  Maybe one of the two could be condensed to minimal descriptions plus a pointer to the other more detailed comment.

brettcoon (commented):
I'm still going through this, but have a few documentation suggestions.  I'll continue reviewing tomorrow.

clients/drcachesim/analysis_tool.h
```diff
@@ -246,21 +248,25 @@ template <typename RecordType> class analysis_tool_tmpl_t {
...
      * \p interval_id is a positive ordinal of the trace interval that just ended.
      * Trace intervals have a length equal to the \p -interval_microseconds specified
      * to the framework. Trace intervals are measured using the value of the
-     * #TRACE_MARKER_TYPE_TIMESTAMP markers. The provided \p interval_id
-     * values will be monotonically increasing but may not be continuous,
-     * i.e. the tool may not see some \p interval_id if the trace did not have
-     * any activity in that interval.
+     * #TRACE_MARKER_TYPE_TIMESTAMP markers. Alternatively, trace intervals may be
+     * defined in terms of the \p -interval_instr_count count of shard-local instrs
+     * specified to the framework.
```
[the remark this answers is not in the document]
abhinav92003: Done.

clients/drcachesim/analysis_tool.h
```diff
@@ -246,21 +248,25 @@ template <typename RecordType> class analysis_tool_tmpl_t {
...
      *
-     * The returned \p interval_state_snapshot_t* will be passed to the
-     * combine_interval_snapshots() API which is invoked by the framework to merge
-     * multiple \p interval_state_snapshot_t from different shards in the parallel
-     * mode of the analyzer.
+     * The provided \p interval_id values will be monotonically increasing but may
+     * not be continuous, i.e. the tool may not see some \p interval_id if the trace
+     * did not have any activity in that interval.
```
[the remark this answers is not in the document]
abhinav92003: It doesn't apply to instr count intervals. Clarified that the non-continuous behavior is only for timestamp intervals.

clients/drcachesim/analysis_tool.h
abhinav92003: Done.

clients/drcachesim/analysis_tool.h
abhinav92003: Done.

abhinav92003:
Xref #6665 that marks the data members of interval_state_snapshot_t as private. This would obviate the need  to document warnings to prevent the user tools from modifying them.

clients/drcachesim/analyzer.cpp
```diff
@@ -1046,6 +1028,31 @@ analyzer_tmpl_t<RecordType, ReaderType>::print_stats()
     return true;
 }
 
+template <typename RecordType, typename ReaderType>
+bool
+analyzer_tmpl_t<RecordType, ReaderType>::finalize_interval_snapshots(
+    analyzer_worker_data_t *worker, bool parallel, int shard_idx)
+{
+    assert(parallel || shard_idx == 0); // Default to zero for the serial mode.
```
brettcoon, line 1036: I don't really understand this comment.  How about something like "// Only parallel mode supports a non-zero shard_idx."

clients/drcachesim/analyzer.h
```diff
@@ -262,6 +262,13 @@ template <typename RecordType, typename ReaderType> class analyzer_tmpl_t {
     RecordType
     create_idle_marker();
 
+    // Invoked after all interval state snapshots have been generated for the given
```
brettcoon, line 265: I think it's worth also mentioning here that it is invoked *before* snapshots are combined or printed.

brettcoon (approved):

clients/drcachesim/analyzer.cpp
abhinav92003: Done.

clients/drcachesim/analyzer.h
abhinav92003: Done.

Issue #6593, opened 2024-01-26
Add zipfile reading support to record filter tool


huyangh-430:
The record filter analysis tool does not support filtering .zip traces today. Since we have moved to .zip as the default format, we should add that support.

derekbruening:
Xref #6573

derekbruening:
Re-opening for instruction removal support which requires re-chunking.

PR #6669, opened 2024-02-21
i#6648 trim tool: Move file removal to the end

derekbruening:
Since file removal of now-empty shards can involve global operations that could race among shards, we move shard removal to the aggregation stage print_results() in the record filter.

Issue: #6648, #6593


derekbruening:
win64 failure is the build race #5888.  Will re-run.

abhinav92003 (approved):

PR #6673, opened 2024-02-23
i#6675: Collapse consecutive idle replay entries

derekbruening:
Saves substantial file space in the drmemtrace scheduler's record-replay file when cores are idling by combining consecutive idle entries.

Adds checks to all unit tests which create replay files; this requires access to the non-public record format, done via a helper class.

Also tested on several real applications: for one the record file drops from 34MB to 4K, matching the size of the as-traced schedule files.  Even tiny applications like threadsig show clear reductions.  Consecutive idle entries were ballooning these files.  Large applications had multi-GB files; with this fix they are orders of magnitude smaller.

Issue: #6471, #6675
Fixes: #6675

### DIFF
--- a/clients/drcachesim/scheduler/scheduler.cpp
+++ b/clients/drcachesim/scheduler/scheduler.cpp
@@ -106,6 +106,23 @@ typedef dynamorio::drmemtrace::record_file_reader_t<std::ifstream>
     default_record_file_reader_t;
 #endif
 
+std::string
+replay_file_checker_t::check(archive_istream_t *infile)
+{
+    // Ensure we don't have repeated idle records, which balloon the file size.
+    scheduler_t::schedule_record_t record;
+    bool prev_was_idle = false;
+    while (infile->read(reinterpret_cast<char *>(&record), sizeof(record))) {
+        if (record.type == scheduler_t::schedule_record_t::IDLE) {
+            if (prev_was_idle)
+                return "Error: consecutive idle records";
+            prev_was_idle = true;
+        } else
+            prev_was_idle = false;
+    }
+    return "";
+}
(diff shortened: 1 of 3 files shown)


clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -1628,6 +1628,12 @@ scheduler_tmpl_t<RecordType, ReaderType>::record_schedule_segment(
     // We always use the current wall-clock time, as the time stored in the prior
     // next_record() call can be out of order across outputs and lead to deadlocks.
     uint64_t timestamp = get_time_micros();
+    if (type == schedule_record_t::IDLE &&
```
abhinav92003, line 1631: Add a check to some existing test at scheduler_unit_tests.cpp?

abhinav92003 (approved):

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: It will not be trivial as while some tests do construct synthetic schedule files none of them peer inside written schedule files, and that is the only way to tell as otherwise it is a performance change.  I will spend a little time to see how easy it is.

derekbruening: I had to add a friend class to access the non-public schedule record type (those synthesized test ones are as-traced, not replay) but I did manage to add checks which fail without this fix.

derekbruening: I think a separate issue on this may be worthwhile in case others have hit this; I know someone internally did see very large files.  Will file.

Issue #6599, opened 2024-01-26
Invariant failure "Syscall marker not preceded by timestamp + cpuid" on win32


derekbruening:
The win32 long (post-merge) test is now failing every time.  I grabbed 3 cases and they all have the same failure:

```
$ tail -18 logs*/vs2019-32/6_Run\ Suite.txt | egrep -v 'ignore|Process'
==> logs/vs2019-32/6_Run Suite.txt <==
2024-01-26T04:44:07.8320478Z ====> FAILURE in debug-internal-32 <====
2024-01-26T04:44:07.8321455Z debug-internal-32: 556 tests passed, **** 15 tests failed, but ignoring 12 for i#2145: ****
2024-01-26T04:44:07.8324403Z 	code_api|tool.drcacheoff.windows-invar 
2024-01-26T04:44:07.8328496Z 	code_api|tool.drcacheoff.invariant_checker 
2024-01-26T04:44:07.8329258Z 	code_api|tool.drcacheoff.getretaddr_record_replace_retaddr 

==> logs2/vs2019-32/6_Run Suite.txt <==
2024-01-26T18:56:04.5217253Z ====> FAILURE in debug-internal-32 <====
2024-01-26T18:56:04.5218497Z debug-internal-32: 556 tests passed, **** 15 tests failed, but ignoring 12 for i#2145: ****
2024-01-26T18:56:04.5221822Z 	code_api|tool.drcacheoff.windows-invar 
2024-01-26T18:56:04.5226724Z 	code_api|tool.drcacheoff.invariant_checker 
2024-01-26T18:56:04.5227588Z 	code_api|tool.drcacheoff.getretaddr_record_replace_retaddr 

==> logs3/vs2019-32/6_Run Suite.txt <==
2024-01-26T06:24:10.3661901Z ====> FAILURE in debug-internal-32 <====
2024-01-26T06:24:10.3662896Z debug-internal-32: 556 tests passed, **** 15 tests failed, but ignoring 12 for i#2145: ****
2024-01-26T06:24:10.3666076Z 	code_api|tool.drcacheoff.windows-invar 
2024-01-26T06:24:10.3670281Z 	code_api|tool.drcacheoff.invariant_checker 
2024-01-26T06:24:10.3671114Z 	code_api|tool.drcacheoff.getretaddr_record_replace_retaddr 
```

All 3 tests are hitting this invariant failure:
```
[...]

Issue #6607, opened 2024-01-31
TEST REPORTING BUG: Wrong issue number can be reported for failing tests that are ignored


xdje42:
**Describe the bug**

Test runs can report the wrong issue number for ignored tests.

**To Reproduce**

See for example,
https://github.com/DynamoRIO/dynamorio/actions/runs/7731301531/job/21078628227#step:7:84492

```
====> FAILURE in debug-internal-64 <====
debug-internal-64: 902 tests passed, **** 6 tests failed, of which 2 were flaky, but ignoring 3 for i#2941: ****
	(ignore: i#2941) 	code_api|client.attach_test 
	code_api|api.static_maps_mixup_novars_FLAKY 
	(ignore: i#2941) 	code_api,opt_memory|common.loglevel 
	(ignore: i#2941) 	code_api,opt_speed|common.decode-stress 
	code_api|tool.drcachesim.invariants 
	code_api|api.static_sideline_FLAKY 
WARNING: maximum warning/error limit hit for release-external-64!
  Manually verify whether it succeeded.
```

Note i#2941 is being reported as the issue for the attach_test failure.
We'll need to augment ignored test marking to let one specify the issue so that it makes it through to the test output.

PR #6674, opened 2024-02-23
i#6599: Mark tool.record_filter test as flaky on win32

xdje42:
Mark the tool.record_filter test as flaky on win32

Issue: #6599

### DIFF
--- a/suite/runsuite_wrapper.pl
+++ b/suite/runsuite_wrapper.pl
@@ -249,6 +249,7 @@ for (my $i = 0; $i <= $#lines; ++$i) {
                 'code_api|tool.drcacheoff.windows-invar' => 1, # i#6599
                 'code_api|tool.drcacheoff.invariant_checker' => 1, # i#6599
                 'code_api|tool.drcacheoff.getretaddr_record_replace_retaddr' => 1, # i#6599
+                'code_api|tool.record_filter' => 1, # i#6599
                 );
 
             %ignore_failures_64 = (
--- a/suite/runsuite.cmake
+++ b/suite/runsuite.cmake
@@ -86,16 +86,11 @@ if (UNIX AND NOT APPLE AND NOT ANDROID AND NOT cross_riscv64_linux_only)
     RESULT_VARIABLE ldd_result ERROR_VARIABLE ldd_err OUTPUT_VARIABLE ldd_out)
   if (ldd_result OR ldd_err)
     # Failed; just move on.
-  elseif (ldd_out MATCHES "GLIBC 2.3[5-9]")
+  elseif (ldd_out MATCHES "xyzGLIBC 2.3[5-9]")
     # XXX i#5437, i#5431: While we work through Ubuntu22 issues we run
     # just a few tests.
     set(extra_ctest_args INCLUDE_LABEL UBUNTU_22)
     set(arg_debug_only ON)
-  elseif (arg_32_only AND NOT cross_aarchxx_linux_only AND NOT cross_android_only)
-    # TODO i#6417: The switch to AMD VM's for GA CI has broken many of our tests.
-    # This includes timeouts which increases suite length.
-    # Until we get ths x86-32 job back green, we drop back to a small set of tests.
-    set(extra_ctest_args INCLUDE_LABEL UBUNTU_22)
   endif ()
 endif ()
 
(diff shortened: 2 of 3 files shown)


abhinav92003 (approved):

suite/runsuite_wrapper.pl
```diff
@@ -249,6 +249,7 @@
                 'code_api|tool.drcacheoff.windows-invar' => 1, # i#6599
                 'code_api|tool.drcacheoff.invariant_checker' => 1, # i#6599
                 'code_api|tool.drcacheoff.getretaddr_record_replace_retaddr' => 1, # i#6599
+                'code_api|tool.record_filter' => 1, # i#6599
```
derekbruening, line 252: There's no record of this test ever failing that I can easily see: I don't see it in #6599, nor are there recent xref links from other PR's or issues.  Please update #6599 pointing to where you are seeing this test fail.

derekbruening: Adding that info the PR description would also help: is this only the post-merge, e.g., is key info.

abhinav92003: > Adding that info the PR description would also help: is this only the post-merge, e.g., is key info.

Yes I found logs (see my other comment) in the post-merge long suite.

abhinav92003: Since tool.record_filter is post-merge only, no PR would encounter the failure and have a reason to xref i#6599. This PR that adds it to the ignore list would be the first one to do so. Adding more details to the PR description (as suggested by both review comments) would help establish that link for future readers.

derekbruening: nit: I would expect the issue to be updated before sending this PR.

suite/runsuite_wrapper.pl
```diff
@@ -249,6 +249,7 @@
                 'code_api|tool.drcacheoff.windows-invar' => 1, # i#6599
                 'code_api|tool.drcacheoff.invariant_checker' => 1, # i#6599
```
derekbruening, line 250: (This comment belongs on line 242)

We should consider putting these "These are from the long suite" ones all under `$is_long`.  I don't think we had that variable available when these were first added, but we have it now.

derekbruening: OTOH it could be that these aren't run in the shortened 32-bit windows suite so they might fail in a short suite if present.

abhinav92003: I had a similar thought during my review: should we have a separate ignore list for long suite tests. But would that be useful? Would we want to ignore a test in one suite but not the other? Initially I thought no, but maybe you think we should?

abhinav92003: Comment race:

> OTOH it could be that these aren't run in the shortened 32-bit windows suite so they might fail in a short suite if present.

Right. So maybe it's better to not gate it with `$is_long`.

derekbruening: We do have some under $is_long today.  I think if we have the same test in both and it only fails in the long (can happen; remember the merge queue seems to be a little different env and things fail there but not in regular pre-merge) we'd want it in the separate long list.

xdje42: Unless a test is run differently !is_long vs is_long (is this one?) then I think it's a bug in the test harness and/or test if it fails in one and not the other. And if the test is run differently then it's really a different test - giving it the same name !long vs long will be a source of confusion.
Our test harness is already complex enough, and in need of some TLC (eg, #6607).
I would advocate for not trying to have a separate section for is_long.
Eg, moving a test from one to the other means having to update the flaky list - a test is either flaky or it isn't.
And if developers hack the test scripts to do experiments, remembering to update the flaky list is yet another bit of administrivia to have to deal with.

suite/runsuite_wrapper.pl
xdje42: The issue is automagically updated already. I wasn't expecting to repeat the info.
[Certainly updating the PR description is a good idea.]

xdje42: PR description updated

suite/runsuite_wrapper.pl
derekbruening: Like I said, we have seen tests fail one Github Actions setup consistently and pass in another, and it's the same test.  So this can happen with different environments, when it's the exact same test on our side.

suite/runsuite_wrapper.pl
derekbruening: The main request here was to update the issue #6599.

xdje42: Done

PR #6549, opened 2024-01-10
i#5365 Update ci-aarch64-native workflow with new runner

joshua-warburton:
Update ci-aarch64xx workflow with new runner on SVE hardware.

Adding a new self-hosted-runner for this pipeline to enhance the test
coverage. A list of unit tests which fail intermittently are added to
the ignore list ignore_failures_sve in runsuite_wrapper.pl.

Issue: 5365

PR #6679, opened 2024-02-26
i#6678: Rename native ci-aarchxx workflow to ci-aarch64-native

### LINKED ISSUE #6678: INFRASTRUCTURE: Two workflows named ci-aarchxx
There are currently two workflows named "ci-aarchxx".
One is for native aarch64 testing and one does cross-testing of arm and aarch64.
This can be seen by looking at the Workflows column on the left side of the screen here:
https://github.com/DynamoRIO/dynamorio/actions

One of the workflows used to be named ci-aarch64-native and was renamed to ci-aarchxx in #6549 .

It would be clearer to have the two workflows have different names.

xdje42:
Rename the native ci-aarchxx workflow back to its original name ci-aarch64-native. There are currently two workflows named ci-aarchxx which is confusing. The workflow was renamed to ci-aarchxx in #6549.

Fixes #6678

### DIFF
--- a/.github/workflows/ci-aarch64-native.yml
+++ b/.github/workflows/ci-aarch64-native.yml
@@ -31,7 +31,7 @@
 
 # Github Actions workflow for aarch64 Continuous Integration testing.
 
-name: ci-aarchxx
+name: ci-aarch64-native
 on:
   # Run on pushes to master and on pull request changes, including from a
   # forked repo with no "push" trigger, while avoiding duplicate triggers.


.github/workflows/ci-aarch64-native.yml
```diff
@@ -31,7 +31,7 @@
 
 # Github Actions workflow for aarch64 Continuous Integration testing.
 
-name: ci-aarchxx
+name: ci-aarch64-native
```
derekbruening, line 34: Note that this will break the required test list which is by name so that will need an admin update (normally after the 1st one is run; maybe the old one is still there though).

xdje42: That's what I figured. I have email outstanding to ARM folks asking exactly that (as in "what did they do when they changed the name").

derekbruening: I can update the required list if they're ok with the change, but please send this to @joshua-warburton for review: maybe there was a reason for the name change we are not aware of.

joshua-warburton: It was renamed to be more consistent with the existing tests which are generally ci-[archname]/agent

derekbruening: I could see it either way on the name: I don't have strong feelings.  The one thing I would like to improve if it's easy is to speed up the ubuntu-20-arm64-sve which is taking 20 minutes and is the slowest of our tests.

xdje42: Can we change name of the other ci-aarchxx workflow then?

Issue #6635, opened 2024-02-06
Add core-sharded support to record_filter


derekbruening:
We'd like to be able to write out new core-sharded drmemtrace files for the result of software thread interleaving on virtual cores from the trace scheduler.  The plan is to run the record_filter tool on the analysis framework (built for record operation) on top of the scheduler with the desired schedule config, and add support to generate new zipfile chunks to the record_filter tool (which we want anyway to be able to remove instructions from thread-sharded files: part of #6593).

derekbruening:
For analyzing core-sharded-on-disk files: my approach is to add a new filetype indicating core-sharded, and have the scheduler read every input to find its filetype record so the analyzer can check the filetype at init time.

That all works fine for typical runs: but it breaks many specific modes, from the readahead:

Breaks online: scheduler init just blocks!
=> new option read_inputs_in_init, turned off for ipc readers.

Breaks inv checks b/c input record ord is too far => using output ord until 1st
instr.

Breaks unit tests which have no filetype records => stopping readahead at
pagesize marker.

Breaks replay-as-traced b/c input record ord is too far => disabling read-ahead
for as-traced.

Breaks record filter legacy null check b/c legacy trace has no filetype and it
reads the timestamp which means the stream's last-timestamp is now >
stop-timestamp and so the resulting output trace has the filter-end marker 1st
thing!
```
$ zcat ../src/clients/drcachesim/tests/drmemtrace.legacy-for-record-filter.x64.tracedir/drmemtrace.threadsig.10506.7343.trace.gz | od -A x -t x2 -w12 | awk '{printf "%s | %s %s %s%s%s%s\n", $1, $2, $3, $7, $6, $5, $4}' | head
000000 | 0019 0000 0000000000000001
00000c | 0016 0004 000000000000290a
000018 | 0018 0004 000000000000290a
000024 | 001c 0002 002ede0c00c77676
000030 | 001c 0003 0000000000000001
```
=>
```
$ zcat record_filter_tests_tmp_output/null_filter/drmemtrace.threadsig.10511.8961.trace.gz | od -A x -t x2 -w12 | awk '{printf "%s | %s %s %s%s%s%s\n", $1, $2, $3, $7, $6, $5, $4}' | head
000000 | 001c 0017 0000000000000000
00000c | 0019 0000 0000000000000001
000018 | 0016 0004 000000000000290f
```

derekbruening:
Closing this and filing any further features or problems as separate issues.

PR #6680, opened 2024-02-27
i#6685: Add core-sharded-on-disk analysis support

### LINKED ISSUE #6685: Add first-class support for analyzing core-sharded-on-disk traces
Split from #6635 where we'll be producing core-sharded trace files from thread-sharded originals and storing them on disk (as opposed to just analyzing them in a core-sharded schedule #5694).

Adding a new type of trace file adds complexity, and we had originally hoped to always re-schedule the original traces: but there are multiple use cases where having the core schedule permanently on disk is useful.  This includes feeding to simulators that are not using our dynamic scheduler.

The plan is to tap into the existing dynamic core-sharded analysis from #5694.  We would add a new filetype marking the trace files as core-sharded.  The scheduler will have to read ahead to get the filetype: which may cause headaches with input ordinals.

We'll want to update the invariant checker so we can use it on such traces: #6684.

----------------------

Pasting from https://github.com/DynamoRIO/dynamorio/issues/6635#issuecomment-1967420976

For analyzing core-sharded-on-disk files: my approach is to add a new filetype indicating core-sharded, and have the scheduler read every input to find its filetype record so the analyzer can check the filetype at init time.

That all works fine for typical runs: but it breaks many specific modes, from the readahead:

Breaks online: scheduler init just blocks!
=> new option read_inputs_in_init, turned off for ipc readers.

Breaks inv checks b/c input record ord is too far => using output ord until 1st
instr.

Breaks unit tests which have no filetype records => stopping readahead at
pagesize marker.

Breaks replay-as-traced b/c input record ord is too far => disabling read-ahead
for as-traced.

Breaks record filter legacy null check b/c legacy trace has no filetype and it
reads the timestamp which means the stream's last-timestamp is now >
stop-timestamp and so the resulting output trace has the filter-end marker 1st
thing!
```
$ zcat ../src/clients/drcachesim/tests/drmemtrace.legacy-for-record-filter.x64.tracedir/drmemtrace.threadsig.10506.7343.trace.gz | od -A x -t x2 -w12 | awk '{printf "%s | %s %s %s%s%s%s\n", $1, $2, $3, $7, $6, $5, $4}' | head
000000 | 0019 0000 0000000000000001
00000c | 0016 0004 000000000000290a
000018 | 0018 0004 000000000000290a
000024 | 001c 0002 002ede0c00c77676
000030 | 001c 0003 0000000000000001
```
=>
```
$ zcat record_filter_tests_tmp_output/null_filter/drmemtrace.threadsig.10511.8961.trace.gz | od -A x -t x2 -w12 | awk '{printf "%s | %s %s %s%s%s%s\n", $1, $2, $3, $7, $6, $5, $4}' | head
000000 | 001c 0017 0000000000000000
00000c | 0019 0000 0000000000000001
000018 | 0016 0004 000000000000290f
```

derekbruening:
Add core-sharded-on-disk analysis support:

Adds a new filetype OFFLINE_FILE_TYPE_CORE_SHARDED set by record_filter when core-sharded.

Has the scheduler read ahead to the filetype in every input, and arrange to make it available at init time to output streams.

The analyzer looks for the filetype and sets the shard type as core-sharded if set.  This is different from a dynamic-scheduling core-sharded as we do not need a 1:1 shard:workers mapping.

Allows header records in reader_t::process_input_entry(), which are now expected with core-sharded-on-disk traces.

Adds core-sharded reader support for seeing only portions of some threads, where we have to relax some tid and pid checks.

Checks in 6 core-sharded-on-disk x64 threadsig traces for use in a test of basic_counts analyzing them.  That analysis fails without all of these framework changes.

The changes to the record filter to support generating those traces in the first place will come in separately.

Issue: #6685, #6635
Fixes #6685


derekbruening:
This has been a real nightmare to get all the modes and tests working: there is some fragility in the infrastructure for sure.  Finally have everything passing except the invariant checker pthreads test on the aarch64-sve runner: but I cannot reproduce that failure anywhere else including aarch64 machines I have access to; the failure itself is one I saw a number of commits back on this same test locally and fixed so it is weird it shows up still on this one machine.  I don't know that tmate works on self-hosted runners?

clients/drcachesim/scheduler/scheduler.h
```diff
@@ -1485,6 +1504,21 @@ template <typename RecordType, typename ReaderType> class scheduler_tmpl_t {
     memtrace_stream_t *
     get_input_stream(output_ordinal_t output);
 
+    // Returns the possibly adjusted (for queued records) record ordinal for the current
```
abhinav92003, line 1507: "possibly adjusted" is arguably an internal detail. It only means that the scheduler may have read ahead internally and if so would need to adjust the counts returned by the stream? Maybe it's not something the API user needs to be aware of.

clients/drcachesim/scheduler/scheduler.h
```diff
@@ -916,7 +931,9 @@ template <typename RecordType, typename ReaderType> class scheduler_tmpl_t {
         /**
          * Returns a unique identifier for the current output stream.  For
          * #MAP_TO_RECORDED_OUTPUT, the identifier is the as-traced cpuid mapped to this
-         * output.  For dynamic schedules, the identifier is the output stream ordinal.
+         * output.  For dynamic schedules, the identifier is the output stream ordinal,
+         * except for #OFFLINE_FILE_TYPE_CORE_SHARDED inputs where the identifier
```
abhinav92003, line 935: For OFFLINE_FILE_TYPE_CORE_SHARDED we have input_stream_ordinal == output_stream_ordinal right? If so, do we need this clarification?

clients/drcachesim/scheduler/scheduler.h
```diff
@@ -623,6 +625,15 @@ template <typename RecordType, typename ReaderType> class scheduler_tmpl_t {
          * sensitivity studies.
          */
         bool randomize_next_input = false;
+        /**
+         * If true, the scheduler will read from each input to determine its filetype
+         * during initialization.  If false, the filetype will not be available prior
+         * to explicit record rerieval by the user, but this may be required for
+         * inputs whose sources are not yet set up at scheduler init time.
```
abhinav92003, line 632: I was wondering in what case this must be set to false. Can you give an example of "inputs whose sources are not yet set up at scheduler init time"?

clients/drcachesim/scheduler/scheduler.h
```diff
@@ -1485,6 +1504,21 @@ template <typename RecordType, typename ReaderType> class scheduler_tmpl_t {
...
+
+    // Returns the possibly adjusted (for queued records) first timestamp for the current
+    // input stream interface for the 'output_ordinal'-th output stream.
+    uint64_t
+    get_input_first_timestamp(output_ordinal_t output);
+
+    // Returns the possibly adjusted (for queued records) last timestamp for the current
+    // input stream interface for the 'output_ordinal'-th output stream.
+    uint64_t
+    get_input_last_timestamp(output_ordinal_t output);
```
abhinav92003, line 1520: Could we add some checks in  scheduler_unit_tests.cpp for these new apis?

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -1254,28 +1274,68 @@ scheduler_tmpl_t<RecordType, ReaderType>::read_switch_sequences()
...
+    assert(options_.read_inputs_in_init);
     // Read ahead in each input until we find a timestamp record.
     // Queue up any skipped records to ensure we present them to the
     // output stream(s).
     for (size_t i = 0; i < inputs_.size(); ++i) {
         input_info_t &input = inputs_[i];
-        if (input.next_timestamp <= 0) {
+        bool found_filetype = false;
+        bool found_timestamp = !gather_timestamps || input.next_timestamp > 0;
+        if (!found_filetype || !found_timestamp) {
```
abhinav92003, line 1292: Some narration would be nice.

// Inputs may have queued records which should be checked first.

// Now read past the queued records until we find the timestamp and filetype.

Better: input_info_t should have an API to get the next record which automatically returns the queued entry if there is one. Would avoid the duplication of this logic.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -1254,28 +1274,68 @@ scheduler_tmpl_t<RecordType, ReaderType>::read_switch_sequences()
...
+                    VPRINT(this, 2, "Input %zu filetype %zu\n", i, marker_value);
+                }
                 if (record_type_is_timestamp(record, input.next_timestamp))
+                    found_timestamp = true;
+                if (found_filetype && found_timestamp)
                     break;
+                // Don't go too far if only looking for filetype, to avoid reaching
+                // the first instruction, which messes up some unit tests that don't
+                // bother to include a filetype.  Just exit with a 0 filetype.
+                if (!found_filetype &&
```
abhinav92003, line 1330: Wouldn't it be better to modify those unit tests to include a filetype (if possible)? These special-case checks make it harder to read sometimes.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -692,6 +692,34 @@ typename scheduler_tmpl_t<RecordType, ReaderType>::scheduler_status_t
...
+        }
+    }
+    bool gather_filetype = options_.read_inputs_in_init;
+    // Avoid reading ahead for replay as it makes the input ords not match in tests.
+    if (options_.mapping == MAP_TO_RECORDED_OUTPUT &&
+        options_.replay_as_traced_istream != nullptr)
+        gather_filetype = false;
+    if (gather_filetype || gather_timestamps) {
+        sched_type_t::scheduler_status_t res =
+            get_initial_input_content(gather_timestamps);
```
abhinav92003, line 714: Add comment: why not just do this everytime to make it simpler? What goes wrong if we do that?

abhinav92003, line 714: Also add to comment: filetype occurs before timestamp.

abhinav92003 (approved):

suite/tests/CMakeLists.txt
```diff
@@ -3872,6 +3872,13 @@ if (BUILD_CLIENTS)
           "-indir ${thread_trace_dir} -simulator_type reuse_time:reuse_distance:histogram:opcode_mix:syscall_mix -core_sharded"
           "")
         set(tool.core_sharded_rawtemp ON) # no preprocessor
+
+        # Test analysis of core-sharded-on-disk traces.
```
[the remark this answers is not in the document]
derekbruening: That is a possibility but we may want to keep examples of this type of trace around, to point external users and say simulator owners at for test traces to ensure their tools support these.  Though for that purpose maybe we should move them to the drmemtrace_samples repository.  It is also useful to have the thread-sharded threadsig ones we have checked in now: some tests prefer a fixed checked-in trace -- so I was leaning toward leaving them.

clients/drcachesim/tools/basic_counts.cpp
```diff
@@ -228,8 +228,6 @@ basic_counts_t::parallel_shard_memref(void *shard_data, const memref_t &memref)
             default: ++counters->other_markers; break;
             }
         }
-    } else if (memref.data.type == TRACE_TYPE_THREAD_EXIT) {
```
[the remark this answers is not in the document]
derekbruening: Wait a minute, I did restore this: this removal was from my analysis runs before I put in this framework support.  I think something got messed up when I re-ordered all the changes.  Grrr, I thought I had been careful there.  Restoring now.  It already has an exception for core-sharded in its assert.

clients/drcachesim/scheduler/scheduler.h
derekbruening: It's probably hard to see in the diff view, but these are actually not public, but protected: so they are more internal helpers.  Even so maybe your comment still applies as the stream class that calls these doesn't necessarily need to know what these are doing inside.  Removed.

clients/drcachesim/scheduler/scheduler.h
derekbruening: Actually if #workers < #shards they are not the same as multiple inputs will run on a single output.

clients/drcachesim/scheduler/scheduler.h
derekbruening: Online analysis sets up IPC readers which open pipes and those pipes block.  The launcher wants to initialize the analyzer+scheduler and then go run the app which will unblock the pipes with data so we want init routines to not block.  Added the example to this comment.

clients/drcachesim/scheduler/scheduler.h
derekbruening: These are not public so the unit tests can't directly call these (without some new subclass).  The tests of the public stream APIs do exercise these as they directly call them for analyzer parallel and serial modes.

derekbruening: These were the main things I had to work on to get all the tests to pass: multiple tests fail if something is off here, from unit tests to larger tests.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Added comments.

Started trying to write a helper function: but it's actually not simple as we don't want to consume the records in the queue, just observe them, while if we iterate over the real reader we have to push onto the queue.  Hmm.  I think at this point I will add a XXX comment about trying to come up with an improvement to the current code.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: It turns out legacy traces don't have filetypes either, including the test ones checked in, and hitting the instruction can hit ordinal problems (maybe taking over the ordinals completely would help that...there is a comment on that added in this PR).  Updated the comment.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Added comments on blocking inputs and ordinals.

derekbruening: Done.

clients/drcachesim/scheduler/scheduler.h
```diff
@@ -623,6 +625,17 @@ template <typename RecordType, typename ReaderType> class scheduler_tmpl_t {
          * sensitivity studies.
          */
         bool randomize_next_input = false;
+        /**
+         * If true, the scheduler will read from each input to determine its filetype
+         * during initialization.  If false, the filetype will not be available prior
+         * to explicit record rerieval by the user, but this may be required for
```
abhinav92003, line 631: s/rerieval/retrieval/

derekbruening: Done.

PR #6682, opened 2024-02-27
Remove unused z16 codec functions

derekbruening:
The compiler on my local machine complains about z16 aarch64 codec routines being unused, which is a fatal error with our build settings.  The compiler is:
```
Apple clang version 14.0.0 (clang-1400.0.29.202)
Target: arm64-apple-darwin23.3.0
```


AssadHashmi:
Can you share the type and version of compiler which emits the unused error?

derekbruening:
> Can you share the type and version of compiler which emits the unused error?

Sorry, should have included this up front:
```
Apple clang version 14.0.0 (clang-1400.0.29.202)
Target: arm64-apple-darwin23.3.0
```

I added this to the commit description too.

Issue #6002, opened 2023-04-25
data races reported when using DR for decoding without explicit dr_standalone_init


derekbruening:
As part of #2499 we try to support using drdecodelib, or libdynamorio.so's decode routines, without an explicit initialization call such as dr_standalone_init.  However, the lazy init's gating checks and initialization ends up flagged by race detectors such as ThreadSanitizer.  Some of these complaints, such as about the `standalone_library` variable itself, are less concerning on x86 where acquire-release semantics are the default; but on arm they do point out real potential issues.

Here are the complaints:

In the triggers for auto-calling standalone_init:
+ standalone_library (should be changed to use explicit acquire-release)
+ heapmgt->global_heap_writable (should be replaced by dynamo_heap_initialized using acquire-release)
+ heapmgt (ditto)

Things initialized:
+ initexit_isa_mode (for this one the initializer is now static so we can remove the code writing it in d_r_decode_init)
+ cpu_info: proc_init() vs proc_get_vendor() during decoding

For moving to acquire-release: we have ATOMIC_1BYTE_WRITE and atomic_read_bool.  The write on x86 is overkill for release, using xchg: but that ends up being a benefit as it satisfies ThreadSanitizer.

Xref #2502 on other lockless issues on arm.

Xref #1409 on refactoring DR code: we can probably remove some of the initialization for standalone mode.  It would be nice to remove DR heap initialization: but we can't easily invoke malloc for non-STATIC_LIBRARY.

derekbruening:
These errors are reported on the invariant_checker drmemtrace tool.  Having it explicitly call dr_standalone_init() eliminates the errors so we could put that in as a workaround.

Issue #6690, opened 2024-03-05
Add better support for decoding different ISAs in different standalone tool threads


derekbruening:
As documented at https://dynamorio.org/API_BT.html#sec_isa, each instr_t has its own ISA mode for
encoding.  However, for decoding, the dcontext's ISA mode is used, and instandalone mode we have
just the global dcontext which uses a global ISA mode var.  Thus, there is not good support for
a standalone tool that wanted to decode multiple ISA modes in separate tool threads.

One possibility would be to honor the ISA mode set in the instr_t passed for decoding, for decoding
routines that take an instr (so this would not help something like decode_sizeof).

Another possibility would be to use toolchain TLS support to create per-thread dcontext_t instances
in standalone mode, which would give us lock-free local heap for non-STATIC_LIBRARY for thread-private
objects.

derekbruening:
Looks like we have a pre-existing bug on the multi-dcontext solution: #1595.  It points at the AArch32 global `decode_state_t` and `encode_state_t` used for `GLOBAL_DCONTEXT`: so for AArch32, encoding is problematic as well as decoding.

derekbruening:
Xref #6002

Issue #1595, opened 2015-01-30
add multi-dcontext support to drdecodelib


derekbruening:
Currently drdecodelib only supports using GLOBAL_DCONTEXT for memory allocation and setting the ISA mode.  Now we're adding IT block info as well.  We should re-consider threading support with this separate library to make it easier to avoid races when changing the ISA modes.

derekbruening:
For a concrete example of existing code that tries to use a local dcontext for the ISA mode but ends up being racy with drdecodelib as there is only the global dcontext, we have this pattern which is in instr_get_eflags():

```
        dcontext_t *dcontext = get_thread_private_dcontext();
        dr_isa_mode_t old_mode;
        ...
        dr_set_isa_mode(dcontext, instr_get_isa_mode(instr), &old_mode);
        decode_eflags_usage(dcontext, instr_get_raw_bits(instr), &instr->eflags,
                            DR_QUERY_INCLUDE_ALL);
        dr_set_isa_mode(dcontext, old_mode, NULL);
```

derekbruening:
Given that AArch32 seems to be problematic for basic decoding and encoding with its global state for GLOBAL_DCONTEXT, maybe we should just fix this and fix #6002 at the same time by requiring an explicit init call in every thread (having it create a new dcontext) and dropping the implicit init support? There may be complexities with a new dcontext in standalone mode: will some components try to do managed-mode-only things with it at init time? Another challenge of not having clear separation between managed mode and regular library use.

PR #6692, opened 2024-03-06
i#6690 dcontext_t standalone: update doc

edeiana:
Documents the fact that standalone dcontext_t is not fully thread-safe, hence parallel analysis tools should be aware that when different threads set DR ISA mode at the same time a race condition happens.

Issue #6690

### DIFF
--- a/core/lib/dr_tools.h
+++ b/core/lib/dr_tools.h
@@ -48,6 +48,12 @@ DR_API
  * \warning This context cannot be used as the drcontext for a thread
  * running under DR control!  It is only for standalone programs that
  * wish to use DR as a library of disassembly, etc. routines.
+ * \warning This context is not fully thread-safe as it stores
+ * the current ISA mode DR uses in a global variable.
+ * Hence, having different threads (e.g., in a parallel analysis tool)
+ * setting different ISA modes at the same time results in a race condition.
+ * Currently we require parallel analysis tools to use their own lock/unlock
+ * mechanism to avoid such race condition when setting the ISA mode.
  * \return NULL on failure, such as running on an unsupported operating
  * system version.
  */
--- a/core/lib/dr_tools.h
+++ b/core/lib/dr_tools.h
@@ -48,12 +48,16 @@ DR_API
  * \warning This context cannot be used as the drcontext for a thread
  * running under DR control!  It is only for standalone programs that
  * wish to use DR as a library of disassembly, etc. routines.
- * \warning This context is not fully thread-safe as it stores
- * the current ISA mode DR uses in a global variable.
- * Hence, having different threads (e.g., in a parallel analysis tool)
- * setting different ISA modes at the same time results in a race condition.
- * Currently we require parallel analysis tools to use their own lock/unlock
- * mechanism to avoid such race condition when setting the ISA mode.
+ * \warning This context is not fully thread-safe as it stores some state
+ * (such as dr_isa_mode_t and other fields related to AArch32 encoding
+ * and decoding) that is global and may be prone to data races.
+ * For example, having different threads use dr_set_isa_mode() to set
+ * different ISA modes at the same time can result in a data race.
+ * Furthermore, encoding and decoding of AArch32 instructions in parallel
+ * may also result in a data race.
+ * Code that uses stanalone DR context across multiple threads should
+ * implement its own lock/unlock mechanism to avoid such data races
+ * when using dr_set_isa_mode() or encoding/decoding AArch32 instructions.
  * \return NULL on failure, such as running on an unsupported operating
  * system version.
  */
@@ -72,6 +76,9 @@ dr_standalone_exit(void);
 /**
  * Use this dcontext for use with the standalone static decoder library.
  * Pass it whenever a decoding-related API routine asks for a context.
+ * Note that this GLOBAL_DCONTEXT is returned by dr_standalone_init(),
+ * beware of its limitations (especially about thread-safety) described
+ * there.
  */
 #    define GLOBAL_DCONTEXT ((void *)-1)
 #endif
--- a/core/lib/dr_tools.h
+++ b/core/lib/dr_tools.h
@@ -49,17 +49,19 @@ DR_API
  * running under DR control!  It is only for standalone programs that
  * wish to use DR as a library of disassembly, etc. routines.
  * \warning This context is not fully thread-safe as it stores some state
- * (such as dr_isa_mode_t and other fields related to AArch32 encoding
+ * (such as #dr_isa_mode_t and other fields related to AArch32 encoding
  * and decoding) that is global and may be prone to data races.
  * For example, having different threads use dr_set_isa_mode() to set
  * different ISA modes at the same time can result in a data race.
  * Furthermore, encoding and decoding of AArch32 instructions in parallel
  * may also result in a data race.
- * Code that uses stanalone DR context across multiple threads should
+ * Code that uses a standalone DR context across multiple threads should
  * implement its own lock/unlock mechanism to avoid such data races
  * when using dr_set_isa_mode() or encoding/decoding AArch32 instructions.
  * \return NULL on failure, such as running on an unsupported operating
  * system version.
+ */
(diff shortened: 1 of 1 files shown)


core/lib/dr_tools.h
```diff
@@ -48,6 +48,12 @@ DR_API
  * \warning This context cannot be used as the drcontext for a thread
  * running under DR control!  It is only for standalone programs that
  * wish to use DR as a library of disassembly, etc. routines.
+ * \warning This context is not fully thread-safe as it stores
+ * the current ISA mode DR uses in a global variable.
```
abhinav92003, line 52: There are other variables in dcontext_t like encode_state, decode_state to which the same concern applies; and there may be more in future. Better to say that "... as it stores some state (like isa_mode) that may be prone to data races".

core/lib/dr_tools.h
```diff
@@ -48,6 +48,12 @@ DR_API
...
+ * Hence, having different threads (e.g., in a parallel analysis tool)
```
abhinav92003, line 53: s/Hence/E.g./

core/lib/dr_tools.h
```diff
@@ -48,6 +48,12 @@ DR_API
...
+ * setting different ISA modes at the same time results in a race condition.
+ * Currently we require parallel analysis tools to use their own lock/unlock
+ * mechanism to avoid such race condition when setting the ISA mode.
```
abhinav92003, line 56: We should also clarify that the GLOBAL_DCONTEXT uses DR global heap which has its own lock, so the user doesn't need to use locks around decoder calls that may allocate/free memory.

But we start getting into various implementation details:
* that dr_standalone_init returns GLOBAL_DCONTEXT (which also isn't true for -DSTANDALONE_UNIT_TEST)
* which decoder APIs have heap alloc/free calls, vs accesses to other race-prone data like isa_mode. How would a user decide which decoder API calls need explicit locks?

abhinav92003 (changes requested):

core/lib/dr_tools.h
abhinav92003: @derekbruening Thoughts on how we can include these clarifications in the API documentation?

core/lib/dr_tools.h
```diff
@@ -48,6 +48,12 @@ DR_API
...
  * \return NULL on failure, such as running on an unsupported operating
  * system version.
  */
```
derekbruening, line 59: Let's add a note to GLOBAL_DCONTEXT below (line 75) pointing to the warning here.

core/lib/dr_tools.h
derekbruening, line 52: Use an explicit link so it's clear what we're talking about: `dr_set_isa_mode()`, not `instr_set_isa_mode()`.

derekbruening: Ah, yes, and the comments in the code point at #1595 filed long ago on this dcontext issue.  So for AArch32, decoding and encoding are problematic, even if not switching ISA modes.

core/lib/dr_tools.h
derekbruening, line 53: "parallel analysis tool" is a specific reference to drmemtrace: I would remove it as this is more general than that.

core/lib/dr_tools.h
```diff
@@ -48,6 +48,12 @@ DR_API
  * \warning This context cannot be used as the drcontext for a thread
  * running under DR control!  It is only for standalone programs that
  * wish to use DR as a library of disassembly, etc. routines.
+ * \warning This context is not fully thread-safe as it stores
+ * the current ISA mode DR uses in a global variable.
+ * Hence, having different threads (e.g., in a parallel analysis tool)
+ * setting different ISA modes at the same time results in a race condition.
```
derekbruening, line 54: Looks like it's also AArch32 decoding and encoding.

core/lib/dr_tools.h
derekbruening: I don't think we need to worry about our internal unit tests in these docs.

Given that AArch32 seems to be problematic for basic decoding and encoding maybe we should just fix this and fix #6002 at the same time by requiring an explicit init call in every thread (having it create a new dcontext) and dropping the implicit init support?  There may be complexities with a new dcontext in standalone mode: will some components try to do managed-mode-only things with it at init time?  Another challenge of not having clear separation between managed mode and regular library use.

core/lib/dr_tools.h
edeiana: Done. Also mentioned AArch32 parallel encoding/decoding issues.

core/lib/dr_tools.h
edeiana: Done.

core/lib/dr_tools.h
edeiana: Updated the doc about AArch32 encoding/decoding for now.

core/lib/dr_tools.h
edeiana: Done.

core/lib/dr_tools.h
edeiana: Done.

core/lib/dr_tools.h
edeiana: Done.

core/lib/dr_tools.h
edeiana: Added that as well.

core/lib/dr_tools.h
```diff
@@ -48,6 +48,16 @@ DR_API
  * \warning This context cannot be used as the drcontext for a thread
  * running under DR control!  It is only for standalone programs that
  * wish to use DR as a library of disassembly, etc. routines.
+ * \warning This context is not fully thread-safe as it stores some state
+ * (such as dr_isa_mode_t and other fields related to AArch32 encoding
```
derekbruening, line 52: Prefix with `#` to make a link: `#dr_isa_mode_t`

core/lib/dr_tools.h
```diff
@@ -48,6 +48,16 @@ DR_API
...
+ * and decoding) that is global and may be prone to data races.
+ * For example, having different threads use dr_set_isa_mode() to set
+ * different ISA modes at the same time can result in a data race.
+ * Furthermore, encoding and decoding of AArch32 instructions in parallel
+ * may also result in a data race.
+ * Code that uses stanalone DR context across multiple threads should
```
derekbruening, line 58: spelling and missing article: `s/stanalone/a standalone/`

core/lib/dr_tools.h
```diff
@@ -48,6 +48,16 @@ DR_API
...
+ * implement its own lock/unlock mechanism to avoid such data races
+ * when using dr_set_isa_mode() or encoding/decoding AArch32 instructions.
```
derekbruening, line 60: Outside of the Doxygen comment (so it won't appear in the html docs) add a (regular) comment `TODO i#6690: Add better multi-thread standalone decoding support` or sthg like that.

core/lib/dr_tools.h
```diff
@@ -66,6 +76,9 @@ dr_standalone_exit(void);
 /**
  * Use this dcontext for use with the standalone static decoder library.
  * Pass it whenever a decoding-related API routine asks for a context.
+ * Note that this GLOBAL_DCONTEXT is returned by dr_standalone_init(),
```
derekbruening, line 79: nit: s/,/;/

core/lib/dr_tools.h
edeiana: Done.

core/lib/dr_tools.h
edeiana: Done.

core/lib/dr_tools.h
edeiana: Done.

core/lib/dr_tools.h
abhinav92003: You can check if it renders as expected in the generated html file.

edeiana: Done. It's not in the html.

PR #6697, opened 2024-03-08
i#6678 Rename aarchxx workflows, part 2: Rename native ci-aarchxx file

### LINKED ISSUE #6678: INFRASTRUCTURE: Two workflows named ci-aarchxx

xdje42:
There are currently two workflows name "ci-aarchxx" which is confusing. The ci-aarch64-native workflow was renamed to ci-aarchxx in #6549 with the goal of being consistent with other architecture workflows. However, this introduced the ambiguous "ci-aarchxx" naming.

Fix this by renaming ci-aarchxx.yml -> ci-aarchxx-cross.yml, and renaming ci-aarch64-native.yml -> ci-aarchxx.yml.
Furthermore, rename the aarchxx "cross" workflow to ci-aarchxx-cross.

This accomplishes three things:

1. Naming is consistent with the other workflow files: native workflows are named "ci-$ARCH" and live in file "ci-$ARCH.yml".
2. All workflows have unique names.
3. The yml file and workflow have the same name.

In part one (#6696) we renamed the cross ci-aarchxx to ci-aarchxx-cross, in the workflow name and yml file name.
In this part two we leave the workflow name alone, but still rename the file from ci-aarch64-native.yml to ci-aarchxx.yml. The renaming is split into two to better preserve the history: Git can track file renamings, but not when the original file is also replaced with something different.

Fixes #6678


joshua-warburton (approved):

derekbruening (commented):
> … file

Please shorten the title so it fits: if it wraps here it will likely wrap in other history view tools.

xdje42:
There was a merge snafu when updating this branch after #6696 went in. Not sure what the problem was, but the result was something I wasn't comfortable with committing. Instead, I made a copy of the change in #6705 and committed that instead.
This branch is abandoned.

PR #6699, opened 2024-03-08
i#6698 instr_t ISA mode: move from flags to its own field

### LINKED ISSUE #6698: Move ISA mode from instr_t flags to its own dr_isa_mode_t field in instr_t
This issue aims to bring us one step closer to full dynamic ISA selection in DynamoRio (xref #1684 #1595 #6690).
As of now the "flags" field in instr_t holds ISA mode information for instruction encoding.
We want to move such information into a new field of instr_t "dr_isa_mode_t isa_mode;" and update the setter and getter functions for instruction ISA mode accordingly.
This change increases the size of instr_t, but it also frees some bits in "flags" that can now be used for something else (note: all 4 bytes of "flags" were used, so we could use the space).

edeiana:
Moves instruction-related ISA mode from instr_t flags field to a new dr_isa_mode_t field in instr_t and updates setter and getter functions for instr_t ISA mode.

Fixes #6698

### DIFF
--- a/core/ir/aarch64/instr.c
+++ b/core/ir/aarch64/instr.c
@@ -35,18 +35,23 @@
 #include "instr.h"
 #include "decode.h"
 
+#include "encode_api.h"
 #include "opcode_names.h"
 
 bool
 instr_set_isa_mode(instr_t *instr, dr_isa_mode_t mode)
 {
-    return (mode == DR_ISA_ARM_A64);
+    if (mode != DR_ISA_ARM_A64) {
+        return false;
+    }
+    instr->isa_mode = DR_ISA_ARM_A64;
+    return true;
 }
 
 dr_isa_mode_t
 instr_get_isa_mode(instr_t *instr)
 {
-    return DR_ISA_ARM_A64;
+    return instr->isa_mode;
 }
 
 int
--- a/core/ir/arm/instr.c
+++ b/core/ir/arm/instr.c
@@ -39,19 +39,17 @@
 bool
 instr_set_isa_mode(instr_t *instr, dr_isa_mode_t mode)
 {
-    if (mode == DR_ISA_ARM_THUMB)
-        instr->flags |= INSTR_THUMB_MODE;
-    else if (mode == DR_ISA_ARM_A32)
-        instr->flags &= ~INSTR_THUMB_MODE;
-    else
+    if ((mode != DR_ISA_ARM_THUMB) && (mode != DR_ISA_ARM_A32)) {
         return false;
+    }
+    instr->isa_mode = mode;
     return true;
 }
 
 dr_isa_mode_t
 instr_get_isa_mode(instr_t *instr)
 {
-    return TEST(INSTR_THUMB_MODE, instr->flags) ? DR_ISA_ARM_THUMB : DR_ISA_ARM_A32;
+    return instr->isa_mode;
 }
 
 int
--- a/core/ir/instr.h
+++ b/core/ir/instr.h
@@ -202,19 +202,8 @@ enum {
     INSTR_DO_NOT_EMIT = 0x10000000,
     /* PR 251479: re-relativization support: is instr->rip_rel_pos valid? */
     INSTR_RIP_REL_VALID = 0x20000000,
-#ifdef X86
-    /* PR 278329: each instr stores its own mode */
-    INSTR_X86_MODE = 0x40000000,
-#elif defined(ARM)
-    /* We assume we don't need to distinguish A64 from A32 as you cannot swap
-     * between them in user mode.  Thus we only need one flag.
-     * XXX: we might want more power for drdecode, though the global isa_mode
-     * should be sufficient there.
-     */
-    INSTR_THUMB_MODE = 0x40000000,
-#endif
     /* PR 267260: distinguish our own mangling from client-added instrs */
-    INSTR_OUR_MANGLING = 0x80000000,
+    INSTR_OUR_MANGLING = 0x40000000,
 };
 
 #define DR_TUPLE_TYPE_BITS 4
--- a/core/ir/instr_api.h
+++ b/core/ir/instr_api.h
@@ -274,6 +274,11 @@ struct _instr_t {
     /* flags contains the constants defined above */
     uint flags;
 
+    /* Instruction ISA mode used to distinguish between 32/64-bit
+     * architectures for encoding purposes.
+     */
+    dr_isa_mode_t isa_mode;
+
     /* hints for encoding this instr in a specific way, holds dr_encoding_hint_type_t */
     uint encoding_hints;
 
@@ -339,7 +344,7 @@ struct _instr_t {
 
     /* Used to hold the relative offset within an instruction list when encoding. */
     size_t offset;
-};     /* instr_t */
+}; /* instr_t */
 #endif /* DR_FAST_IR */
 
 /**
--- a/core/ir/riscv64/instr.c
+++ b/core/ir/riscv64/instr.c
@@ -36,13 +36,17 @@
 bool
 instr_set_isa_mode(instr_t *instr, dr_isa_mode_t mode)
 {
-    return (mode == DR_ISA_RV64IMAFDC);
+    if (mode != DR_ISA_RV64IMAFDC) {
+        return false;
+    }
+    instr->isa_mode = DR_ISA_RV64IMAFDC;
+    return true;
 }
 
 dr_isa_mode_t
 instr_get_isa_mode(instr_t *instr)
 {
-    return DR_ISA_RV64IMAFDC;
+    return instr->isa_mode;
 }
 
 int
--- a/core/ir/x86/instr.c
+++ b/core/ir/x86/instr.c
@@ -40,7 +40,30 @@
 #include "instr.h"
 #include "decode.h"
 #include "decode_private.h"
+#include "encode_api.h"
 #include "instr_create_shared.h"
+
+bool
+instr_set_isa_mode(instr_t *instr, dr_isa_mode_t mode)
+{
+#ifdef X64
+    if ((mode != DR_ISA_IA32) && (mode != DR_ISA_AMD64)) {
+        return false;
+    }
+#else
+    if (mode != DR_ISA_IA32) {
+        return false;
+    }
+#endif
+    instr->isa_mode = mode;
+    return true;
+}
+
+dr_isa_mode_t
+instr_get_isa_mode(instr_t *instr)
+{
+    return instr->isa_mode;
+}
 
 #ifdef X64
 /*
@@ -50,10 +73,11 @@
 void
 instr_set_x86_mode(instr_t *instr, bool x86)
 {
-    if (x86)
-        instr->flags |= INSTR_X86_MODE;
-    else
-        instr->flags &= ~INSTR_X86_MODE;
+    if (x86) {
+        instr->isa_mode = DR_ISA_IA32;
+    } else {
+        instr->isa_mode = DR_ISA_AMD64;
+    }
 }
 
 /*
@@ -63,36 +87,9 @@ instr_set_x86_mode(instr_t *instr, bool x86)
 bool
 instr_get_x86_mode(instr_t *instr)
 {
-    return TEST(INSTR_X86_MODE, instr->flags);
+    return (instr->isa_mode == DR_ISA_IA32);
 }
 #endif
-
-bool
-instr_set_isa_mode(instr_t *instr, dr_isa_mode_t mode)
-{
-#ifdef X64
-    if (mode == DR_ISA_IA32)
-        instr_set_x86_mode(instr, true);
-    else if (mode == DR_ISA_AMD64)
-        instr_set_x86_mode(instr, false);
-    else
-        return false;
-#else
-    if (mode != DR_ISA_IA32)
-        return false;
-#endif
-    return true;
-}
-
-dr_isa_mode_t
-instr_get_isa_mode(instr_t *instr)
-{
-#ifdef X64
-    return TEST(INSTR_X86_MODE, instr->flags) ? DR_ISA_IA32 : DR_ISA_AMD64;
-#else
-    return DR_ISA_IA32;
-#endif
-}
 
 int
 instr_length_arch(dcontext_t *dcontext, instr_t *instr)


core/ir/x86/instr.c
```diff
@@ -40,8 +40,31 @@
 #include "instr.h"
 #include "decode.h"
 #include "decode_private.h"
+#include "encode_api.h"
```
edeiana, line 43: ?? I did not add this. clangd issue?
Will investigate it's necessary.

core/ir/aarch64/instr.c
```diff
@@ -35,18 +35,23 @@
 #include "instr.h"
 #include "decode.h"
 
+#include "encode_api.h"
```
edeiana, line 38: ?? I did not add this. clangd issue?
Will investigate if it's necessary.

core/ir/instr.h
```diff
@@ -202,19 +202,8 @@ enum {
...
-    /* We assume we don't need to distinguish A64 from A32 as you cannot swap
-     * between them in user mode.  Thus we only need one flag.
-     * XXX: we might want more power for drdecode, though the global isa_mode
-     * should be sufficient there.
-     */
-    INSTR_THUMB_MODE = 0x40000000,
-#endif
     /* PR 267260: distinguish our own mangling from client-added instrs */
-    INSTR_OUR_MANGLING = 0x80000000,
+    INSTR_OUR_MANGLING = 0x40000000,
```
edeiana, line 206: Moved this one bit down.
Not really necessary, but otherwise it looks weird.
Also, it should not break things.

core/ir/instr_api.h
```diff
@@ -339,7 +344,7 @@ struct _instr_t {
 
     /* Used to hold the relative offset within an instruction list when encoding. */
     size_t offset;
-};     /* instr_t */
+}; /* instr_t */
```
edeiana, line 347: clang-format change.
Need to investigate if our internal version is >= than 14.

core/ir/x86/instr.c
```diff
@@ -50,10 +73,11 @@
 void
 instr_set_x86_mode(instr_t *instr, bool x86)
```
edeiana, line 74: Perhaps we want to change all uses of instr_set_x86_mode() to use instr_set_isa_mode() instead?

core/ir/x86/instr.c
```diff
@@ -63,37 +87,10 @@ instr_set_x86_mode(instr_t *instr, bool x86)
 bool
 instr_get_x86_mode(instr_t *instr)
```
edeiana, line 88: See comment on instr_set_x86_mode().

edeiana:
Lots of tests are currently failing.
Probably due to the change in instr_t size.

core/ir/instr.h
derekbruening: We have to be very careful with changing constants or enums: we have to try to support binary compatibility between a client built against a different version of the headers than DR itself, and other two-piece different-version cases like persistent caches.  For this particular case: I can't think of a scenario that would break so it's probably ok.

core/ir/aarch64/instr.c
```diff
@@ -35,18 +35,23 @@
 #include "instr.h"
```
derekbruening, line 35: > Lots of tests are currently failing. Probably due to the change in instr_t size.

There is a "black box" definition with the size.  See e.g. #6237 which is the last change to the struct.  If there's no padding slot we can hide in and the size has changed for 64-bit, we should bump the minor version to 90 as well (see https://dynamorio.org/page_new_release.html#autotoc_md230).

derekbruening: There is a byte slot next to the 3 rep_rel_pos,num_{dsts,srcs}: but that would limit us to 8 bits of ISA flags (currently using 6 with synth addition?).  This would not change either 32-bit or 64-bit sizes if I'm looking at it right.  Hmm, worth considering?

core/ir/arm/instr.c
```diff
@@ -39,19 +39,17 @@
...
+    }
+    instr->isa_mode = mode;
     return true;
 }
 
 dr_isa_mode_t
 instr_get_isa_mode(instr_t *instr)
 {
-    return TEST(INSTR_THUMB_MODE, instr->flags) ? DR_ISA_ARM_THUMB : DR_ISA_ARM_A32;
+    return instr->isa_mode;
```
derekbruening, line 52: Let's move this function to instr_shared.c if all arches are now the same.

core/ir/x86/instr.c
edeiana: It's best practice to add encode_api.h (which defined dr_isa_mode_t) rather than relying on other headers.

core/ir/instr_api.h
edeiana: Manually fixed.

core/ir/aarch64/instr.c
edeiana: I went with your suggestion (thank you!).
Also, dr_isa_mode_t doesn't need to be a "flag", so with 1 byte we still have plenty of ISA modes.

core/ir/arm/instr.c
edeiana: Unfortunately, some instr_get_isa_mode() were just returning a specific ISA mode (without retrieving it from the instr_t flags).
Meaning that they were able to return a valid ISA mode, even though it was not previously set in the flags of instr_t.
There were several (>300) tests relying on the fact that the ISA mode could still be retrieved without being set first, and I wasn't sure if that was intentional, so I modified the instr_get_isa_mode() to reflect the previous behavior.
This also means that they are not all the same anymore.

edeiana:
Thank you for the early review Derek!
I went with you suggestion making isa_mode a 1 byte field in instr_t.
I also had to change instr_get_isa_mode() (see https://github.com/DynamoRIO/dynamorio/pull/6699#discussion_r1518834572) so they don't look the same anymore.

core/ir/arm/instr.c
derekbruening: I think we just need instr_create() to set the mode for A64 and RISCV?

core/ir/x86/instr.c
derekbruening: Hmm I see this in the docs:
```
api/docs/release.dox: - Deprecated instr_set_x86_mode(), replacing it with instr_set_isa_mode().
```

Yet we still have internal uses?  Did we only deprecate external calls?

core/ir/x86/instr.c
```diff
@@ -63,32 +65,31 @@ instr_set_x86_mode(instr_t *instr, bool x86)
...
 bool
 instr_set_isa_mode(instr_t *instr, dr_isa_mode_t mode)
 {
 #ifdef X64
-    if (mode == DR_ISA_IA32)
-        instr_set_x86_mode(instr, true);
-    else if (mode == DR_ISA_AMD64)
-        instr_set_x86_mode(instr, false);
-    else
+    if ((mode != DR_ISA_IA32) && (mode != DR_ISA_AMD64)) {
```
derekbruening, line 76: nit: Superfluous parens

derekbruening, line 76: Same comments about including TODO or XXX comments.

core/ir/x86/instr.c
```diff
@@ -63,32 +65,31 @@ instr_set_x86_mode(instr_t *instr, bool x86)
...
 }
 
 dr_isa_mode_t
 instr_get_isa_mode(instr_t *instr)
 {
 #ifdef X64
-    return TEST(INSTR_X86_MODE, instr->flags) ? DR_ISA_IA32 : DR_ISA_AMD64;
+    return (dr_isa_mode_t)instr->isa_mode;
 #else
     return DR_ISA_IA32;
```
derekbruening, line 94: Same comments as for the other arches.

core/ir/x86/instr.c
edeiana: Oh, I'd suppose so.
Internally they are still in a few places (e.g., https://github.com/DynamoRIO/dynamorio/blob/96c132d84290d6164990a41581f228d405008275/core/arch/emit_utils_shared.c#L2338 and https://github.com/DynamoRIO/dynamorio/blob/96c132d84290d6164990a41581f228d405008275/core/arch/mangle_shared.c#L1895).
If we decide to remove them, should that work go in a separate PR?

core/ir/arm/instr.c
edeiana: You're right, that fixed it.
Now that all instr_get_isa_mode() look the same I moved them to core/ir/instr_shared.c

core/ir/aarch64/instr.c
```diff
@@ -34,13 +34,17 @@
...
-
+#include "encode_api.h"
 #include "opcode_names.h"
 
 bool
 instr_set_isa_mode(instr_t *instr, dr_isa_mode_t mode)
 {
-    return (mode == DR_ISA_ARM_A64);
+    if (mode != DR_ISA_ARM_A64) {
+        return false;
```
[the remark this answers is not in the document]
edeiana: Done.

core/ir/aarch64/instr.c
```diff
@@ -34,13 +34,17 @@
...
+    }
+    instr->isa_mode = DR_ISA_ARM_A64;
+    return true;
 }
 
 dr_isa_mode_t
```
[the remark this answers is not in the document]
edeiana: That's right.
instr_get_isa_mode() now actually returns the instr_t isa_mode for all archs.

core/ir/riscv64/instr.c
```diff
@@ -32,11 +32,16 @@
...
 #include "../globals.h"
 #include "instr.h"
+#include "encode_api.h"
 
 bool
 instr_set_isa_mode(instr_t *instr, dr_isa_mode_t mode)
 {
-    return (mode == DR_ISA_RV64IMAFDC);
+    if (mode != DR_ISA_RV64IMAFDC) {
+        return false;
```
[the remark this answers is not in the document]
edeiana: Done.

core/ir/riscv64/instr.c
```diff
@@ -32,11 +32,16 @@
...
+    }
+    instr->isa_mode = DR_ISA_RV64IMAFDC;
+    return true;
 }
 
 dr_isa_mode_t
```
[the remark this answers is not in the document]
edeiana: Done.

core/ir/x86/instr.c
```diff
@@ -63,32 +65,31 @@ instr_set_x86_mode(instr_t *instr, bool x86)
 bool
 instr_get_x86_mode(instr_t *instr)
 {
-    return TEST(INSTR_X86_MODE, instr->flags);
+    return (instr->isa_mode == DR_ISA_IA32);
```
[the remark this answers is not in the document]
edeiana: Removed them.

core/ir/x86/instr.c
edeiana: Removed.

core/ir/instr_api.h
```diff
@@ -310,6 +310,11 @@ struct _instr_t {
     byte num_dsts;
     byte num_srcs;
 
+    /* Instruction ISA mode used to distinguish between 32/64-bit
+     * architectures for encoding purposes.
+     */
```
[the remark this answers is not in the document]
edeiana: Done.

core/ir/x86/instr.c
```diff
@@ -63,32 +65,31 @@ instr_set_x86_mode(instr_t *instr, bool x86)
...
-        instr_set_x86_mode(instr, true);
-    else if (mode == DR_ISA_AMD64)
-        instr_set_x86_mode(instr, false);
-    else
+    if ((mode != DR_ISA_IA32) && (mode != DR_ISA_AMD64)) {
         return false;
+    }
 #else
-    if (mode != DR_ISA_IA32)
+    if (mode != DR_ISA_IA32) {
```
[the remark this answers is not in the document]
edeiana: Removed brackets from if-then-else single statements.
(I'm used to put them all the time, mostly because if I add instructions later on and I forget to add brackets it always takes me a while to figure out why my code isn't behaving as it should :) but I'll follow the coding style).

core/ir/x86/instr.c
edeiana: Done.

core/ir/x86/instr.c
edeiana: Done.

core/ir/aarch64/instr.c
derekbruening: Ah, right, it's a regular enum, so 1 byte should be plenty as you said.

core/ir/arm/instr.c
```diff
@@ -32,28 +32,25 @@
 
 #include "../globals.h"
 #include "instr.h"
+#include "encode_api.h"
 #include "decode.h"
 
 /* FIXME i#1551: add A64 and Thumb support throughout */
-
+/* XXX: currently only A64 and Thumb is supported for instruction encoding.
```
derekbruening, line 39: s/A64/A32/

Ditto on line below.

core/ir/arm/instr.c
```diff
@@ -32,28 +32,25 @@
 
 #include "../globals.h"
 #include "instr.h"
+#include "encode_api.h"
 #include "decode.h"
 
 /* FIXME i#1551: add A64 and Thumb support throughout */
```
derekbruening, line 38: This comment can be removed: it is stale.

core/ir/instr_api.h
```diff
@@ -310,6 +310,13 @@ struct _instr_t {
     byte num_dsts;
     byte num_srcs;
 
+    /** This field assumes values of type #dr_isa_mode_t.
```
derekbruening, line 313: style: As documented, doxygen comments must either be one line or else the `/**` is on its own line.

Hmm although none of the fields here are doxygen-documented: deliberately b/c we don't want them part of the public interface, b/c we want to change them (like we're doing here).  So maybe just fold this into the other regular comment.

derekbruening, line 313: nit: "assumes" can be confused with assumptions: better "holds" or "stores".

core/ir/instr_shared.c
```diff
@@ -93,6 +93,10 @@ instr_create(void *drcontext)
     instr_set_isa_mode(instr, X64_CACHE_MODE_DC(dcontext) ? DR_ISA_AMD64 : DR_ISA_IA32);
 #elif defined(ARM)
     instr_set_isa_mode(instr, dr_get_isa_mode(dcontext));
+#elif defined(AARCH64)
+    instr_set_isa_mode(instr, DR_ISA_ARM_A64);
+#elif defined(RISCV64)
+    instr_set_isa_mode(instr, DR_ISA_RV64IMAFDC);
```
derekbruening, line 99: Let's have `#else #error` to be sure this gets updated on future arches.

Although: the ARM version of getting it from the dcontext seems like it should work for all arches: so `s/if defined(ARM)/se/` instead is better, keeping the default arch ISA in just one place.

core/ir/instr_shared.c
```diff
@@ -4196,4 +4206,4 @@ move_mm_avx512_reg_opcode(bool aligned64)
 }
 
 #endif /* !STANDALONE_DECODER */
-/****************************************************************************/
+       /****************************************************************************/
```
derekbruening, line 4209: Shouldn't be indented.

core/ir/x86/instr.c
derekbruening: Yes, but seems lower priority.

core/ir/aarch64/instr.c
```diff
@@ -34,19 +34,21 @@
 #include "../globals.h"
 #include "instr.h"
 #include "decode.h"
-
+#include "encode_api.h"
 #include "opcode_names.h"
 
+/* XXX: currently only A64 is supported for instruction encoding.
```
derekbruening, line 40: style: Include an issue # with XXX.

derekbruening: Ditto in all the similar lines below.

edeiana: Done. Added #6690.

core/ir/arm/instr.c
edeiana: Done.

core/ir/arm/instr.c
edeiana: Done.

core/ir/instr_api.h
edeiana: Transformed into a regular comment.
I'll remember the syntax for future doxygen comments, thanks!

edeiana: Oh right!
Done.
Using "holds".

core/ir/instr_shared.c
edeiana: clang-format is driving me crazy :) fixed.

core/ir/instr_shared.c
edeiana: Ok, ARM and all future arches set instr_t isa_mode from dcontext now.
I noticed we don't check the return value of instr_set_isa_mode().
Should we? Or is there a reason we don't?

derekbruening: I meant that a64 and riscv would not have separate code either: only x86 (maybe remove x86 too if you convince yourself it's the same also: but it looks complicated).

derekbruening: Checking return value sounds good.

edeiana: Ah, I had it right before.
Putting all arches except x86 in the #else then.
I'll have a look if x86 can be included as well, and add checks for instr_set_isa_mode() returned value.

edeiana: Had to keep the x86-related #ifdef.
Added a check for the return value of instr_set_isa_mode() using CLIENT_ASSERT(), following how other functions do it (out of curiosity: why is it "CLIENT" ASSERT? and not just "ASSERT"?).

derekbruening (approved):

edeiana:
code_api|tool.drcacheoff.windows-timestamps failed once (#6081) now passing.

PR #6701, opened 2024-03-08
i#6635 core filter, part 5: schedule_stats

derekbruening:
Changes schedule_stats to detect switches via <workload,tid> instead of the input ordinal, to support core-sharded-on-disk traces.

Updates schedule_stats_test which no longer needs its own mock stream if we add get_input_interface to default_memtrace_stream_t and call set_tid.

Adds a schedule_stats end-to-end test on the checked-in core-sharded-on-disk trace to ensure switches are detected.

Issue: #6635


abhinav92003:
This branch needs to be updated post #6700 but I guess it doesn't hinder review that much so I'll continue.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -1515,7 +1565,8 @@ scheduler_tmpl_t<RecordType, ReaderType>::get_tid(output_ordinal_t output)
     int index = outputs_[output].cur_input;
     if (index < 0)
         return -1;
-    if (inputs_[index].is_combined_stream())
+    if (inputs_[index].is_combined_stream() ||
+        TESTANY(OFFLINE_FILE_TYPE_CORE_SHARDED, inputs_[index].reader->get_filetype()))
```
abhinav92003, line 1569: Should is_combined_stream be modified to return true for core-sharded-on-disk?

abhinav92003 (approved):

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: No, I looked into this before and the other uses of is_combined_stream do not apply to core-sharded.

abhinav92003: It should probably be documented then. A core-sharded stream does seem "combined" in a way so some users may expect it to return true.

derekbruening: (It's not a public interface so it wouldn't be users looking at it, but developers.)  Added clarifying documentation.

PR #6707, opened 2024-03-14
i#6635 core filter, part 7: Improve init-idle subclass support

derekbruening:
Improves record_filter_t subclass support for initially-idle cores by refactoring get_output_basename() out of initialize_shard_output(), allowing a subclass to share the complex initial setup while still using its own output scheme.  Also moves the setup variable to protected for access to output_ext_ in subclasses.

Removes code that was refactored into initialize_shard_output() but accidentally left in place in PR #6704.

Tested internally.

Issue: #6635

### DIFF
--- a/clients/drcachesim/tools/filter/record_filter.cpp
+++ b/clients/drcachesim/tools/filter/record_filter.cpp
@@ -238,40 +238,6 @@ record_filter_t::get_writer(per_shard_t *per_shard, memtrace_stream_t *shard_str
 {
     if (per_shard->output_path.empty())
         return "Error: output_path is empty";
-    if (shard_type_ == SHARD_BY_CORE) {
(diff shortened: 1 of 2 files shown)


clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -228,40 +238,6 @@ record_filter_t::get_writer(per_shard_t *per_shard, memtrace_stream_t *shard_str
 {
     if (per_shard->output_path.empty())
         return "Error: output_path is empty";
-    if (shard_type_ == SHARD_BY_CORE) {
```
abhinav92003, line 231: (I did not notice this was not removed when I checked later commits in #6704)

abhinav92003 (approved):

PR #6708, opened 2024-03-14
Clarify docs on IR memory operands

derekbruening:
Adds text clarifying that memory operands do not have their constituent registers listed as separate operands.

This will show up in the online docs here:
https://dynamorio.org/API_BT.html#sec_IR

### DIFF
--- a/api/docs/bt.dox
+++ b/api/docs/bt.dox
@@ -1,5 +1,5 @@
 /* **********************************************************
- * Copyright (c) 2011-2023 Google, Inc.  All rights reserved.
+ * Copyright (c) 2011-2024 Google, Inc.  All rights reserved.
  * Copyright (c) 2007-2009 VMware, Inc.  All rights reserved.
  * **********************************************************/
 
@@ -89,6 +89,19 @@ a primary contributor to DynamoRIO's efficiency.
 The instruction representation includes all of the operands, whether
 implicit or explicit, and the condition code effects of each instruction.
 This allows for analysis of liveness of registers and condition codes.
+The operands are split into sources and destinations.
+
+A memory reference is treated as one operand even when it uses
+registers to compute its address: those constituent registers are not
+listed as their own separate source operands (unless they are read for
+other reasons such as updating the index register).  This means that a
+store to memory will have that store as a destination operand without
+listing the store's addressing mode registers as source operands in
+their own right.  Tools interested in all registers inside such
+operands can use interfaces like instr_reads_from_reg() to query (it
+will consider registers used to compute a store address as being read)
+or opnd_get_reg_used() to generically walk the registers inside an
+operand.
 
 DynamoRIO's IR is mostly opaque to clients.  Key data structures have their
 sizes exposed to allow for stack allocation, but their fields are opaque.  In
--- a/api/docs/bt.dox
+++ b/api/docs/bt.dox
@@ -64,7 +64,8 @@ following sections:
 \section sec_IR Instruction Representation
 
 The primary data structures involved in instruction manipulation are
-the #instr_t, which represents a single instruction, and the \c
+the #opnd_t, which represents one operand; the #instr_t, which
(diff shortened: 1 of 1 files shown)


api/docs/bt.dox
```diff
@@ -89,6 +89,19 @@ a primary contributor to DynamoRIO's efficiency.
...
+The operands are split into sources and destinations.
+
+A memory reference is treated as one operand even when it uses
+registers to compute its address: those constituent registers are not
+listed as their own separate source operands (unless they are read for
+other reasons such as updating the index register).  This means that a
+store to memory will have that store as a destination operand without
+listing the store's addressing mode registers as source operands in
+their own right.  Tools interested in all registers inside such
+operands can use interfaces like instr_reads_from_reg() to query (it
```
edeiana, line 101: When I looked at `instr_reads_from_reg()` it didn't seem like a good way to "query all the registers inside an operand".
My understanding is that `instr_reads_from_reg(instr_t *instr, reg_id_t reg, ...)` will simply tell you if `reg` is used inside any operand of `instr`, but to "query all the registers" as suggested, one would have to iterate over all registers and call `instr_reads_from_reg()` for every one of them.
If we want to suggest a way to query registers from operands, I think it makes more sense to mention: `int opnd_get_num_regs_used(opnd_t opnd)` and `reg_id_t opnd_get_reg_used(opnd_t opnd, int index)`.
Although now we'd have to at least explain what opnd_t is (which is fine btw, we could just mention it in the previous paragraph, something like: "The instruction representation includes all of the operands (which opnd_t represents), whether ... ").

All of this, assuming my understanding of `instr_reads_from_reg()` is correct.

edeiana (approved):

edeiana (approved):

api/docs/bt.dox
```diff
@@ -89,6 +89,19 @@ a primary contributor to DynamoRIO's efficiency.
...
+will consider registers used to compute a store address as being read)
+or opnd_get_reg_used() to generically walk the registers inside an
```
edeiana, line 103: We might want to mention `opnd_get_num_regs_used()` as well, since `opnd_get_reg_used()` also needs an index.
Although, "Used in conjunction with opnd_num_regs_used()" is already mentioned in doxygen for `opnd_get_reg_used()`.

edeiana (approved):

api/docs/bt.dox
derekbruening: Your earlier comments seem to have disappeared: you had one about include at least a ref to #opnd_t which is indeed missing.  I added it to the 1st paragraph which introduces #instr_t.

I reworded to move the instr_reads_from_reg() to be mentioned last to avoid your point about it not being the same as querying/iterating all the registers.

I added opnd_get_num_reg_used() as requested.

edeiana: Oh, yes I removed the whole discussion about `opnd_get_num_reg_used()` and `opnd_get_reg_used()` because I realized you talked about it right after `instr_reads_from_reg()`.
The reordering (talking about those two #opnd_t related functions before `instr_reads_from_reg()`) flows better with the whole explanation of "querying registers from an operand".
Mentioning #opnd_t I didn't think it was critical, but now that it's in, it reads better.

api/docs/bt.dox
```diff
@@ -98,10 +99,11 @@ other reasons such as updating the index register).  This means that a
 store to memory will have that store as a destination operand without
 listing the store's addressing mode registers as source operands in
 their own right.  Tools interested in all registers inside such
-operands can use interfaces like instr_reads_from_reg() to query (it
-will consider registers used to compute a store address as being read)
-or opnd_get_reg_used() to generically walk the registers inside an
-operand.
+operands can use opnd_get_num_regs_used() and opnd_get_reg_used() to
```
edeiana, line 102: LGTM.

edeiana (approved):

Issue #2328, opened 2017-04-06
ASSERT signal.c:3806 info->interrupted == NULL || info->interrupted == f


derekbruening:
Running an app that links DR statically and does an internal attach hits this assert maybe 1 in 10,000 runs, after attach is finished.

```
<Attached to 97 threads in application >
<Application .  Internal Error: DynamoRIO debug check failure: core/unix/signal.c:3806 info->interrupted == NULL || info->interrupted == f
(Error occurred @11002 frags)
internal version, custom build
```

derekbruening:
My theory is that this is coming from the "dropping 3rd pending alarm signal" path which does not clear info->interrupted.  If so, it's easy to fix, and it's innocuous: release build simply replaces info->interrupted.  However, I don't see that "dropping 3rd" message in some saved output from prior instances.

derekbruening:
Looking closer: the "dropping 3rd" drops an older one and keeps 
the new one, so it's not the culprit.

Another way it could happen: if a signal arrives in the IBL we guess at the
target fragment and we could get it wrong.  It seems unlikely though.

In an instance I got in the debugger, the interrupted PC is at the fcache entry.
so it does look like the gencode path.
It could be a signal in fcache_return, and we used asynch_target but had
just exited that fragment?

If that's the case, do we leave it unlinked forever?  Does dispatch notice and relink?
Seems like we should relink where this assert is as a fallback.

derekbruening:
I ruled out fcache_return and we still had some incorrect guesses: but there are other ibl spots where the same behavior could happen.  I added a flag indicating the guess and relaxed the assert when set.

I also put in a relink whenever replacing the interrupted fragment.

derekbruening:
I'm going to relax the assert for these guessing situations, and put in the relinking change, and a clone_record_t stack alignment fix that blocked the other changes by crashing AArch64, but the underying issue with having to guess is that we can have unbounded delay delivering signals.  I'm going to make that under #2042 for clarity.

Also xref #2019

derekbruening:
It turns out all the cases here are clean_call_{save,restore}.  I will add handling and re-instate the assert to help find any other missing cases.

ksco:
@derekbruening Hi, I know this is a very old issue, but I hit a similar one on RISC-V ~1/5 times when I run `linux.signal0001` but can't figure out what it's trying to check.

[...]

PR #6711, opened 2024-03-19
i#3544 RV64: Fix patch_stub for unaligned stub_pc

ksco:
To simplify codegen, we require the exit stub pc to be aligned to 4 bytes. If it is not aligned, we place a 2-byte `c.nop` at the beginning of the exit stub. Therefore, all functions that modify the exit stub should skip the possible `c.nop`.

The wrongly implemented `patch_stub()` makes `unlink_branch()` not affect the exit stub, which makes the async signal handling error-prone.

Before the fix, it hits the following assertion ~1/5 times when running `linux.signal0001` on a RISC-V machine, which indicates that the second signal comes from a different fragment after the unlinking, and the real reason is that the unlinking is not successful.

```
ASSERT signal.c:4954 info->interrupted == NULL || info->interrupted == f
```

It would be great if we could enable the signal tests in CI, but these tests do not work in QEMU for unknown reasons.

Issue: #3544
Related: #2328

### DIFF
--- a/core/arch/riscv64/emit_utils.c
+++ b/core/arch/riscv64/emit_utils.c
@@ -292,6 +292,9 @@ void
 patch_stub(fragment_t *f, cache_pc stub_pc, cache_pc target_pc, cache_pc target_prefix_pc,
            bool hot_patch)
 {
+    /* If stub_pc is not aligned to 4 bytes, the first instruction will be c.nop, see
+     * insert_exit_stub_other_flags(). */
+    stub_pc = ALIGNED(stub_pc, 4) ? stub_pc : stub_pc + 2;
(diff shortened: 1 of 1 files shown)


core/arch/riscv64/emit_utils.c
```diff
@@ -292,6 +292,9 @@ void
 patch_stub(fragment_t *f, cache_pc stub_pc, cache_pc target_pc, cache_pc target_prefix_pc,
```
abhinav92003, line 292: I assume this patch fixes some observed crash? Please add that detail to the PR description.

abhinav92003 (approved):

core/arch/riscv64/emit_utils.c
ksco: Added the details.

PR #6718, opened 2024-03-23
i#6495: Handle invariant errors in x86 QEMU syscall templates

abhinav92003:
Handles various invariant errors seen in system call trace templates collected on x86 QEMU.

Modifies syscall trace template file format to use the TRACE_MARKER_TYPE_SYSCALL_TRACE_START and TRACE_MARKER_TYPE_SYSCALL_TRACE_END markers to show start and end respectively of each syscall trace template, instead of separating them using a TRACE_MARKER_TYPE_SYSCALL marker. This makes it easier to write invariant checks that also work for the syscall trace template file (in addition to an actual trace file injected with trace templates).

Handles cases where there are a different number of read/write records than expected by the decoder; after iret, variants of xrstor, variants of xsaves, and prefetch instrs.

Relaxes the PC discontinuity check after hlt, and within two instrs of sti (which enables interrupts, so there may be an interrupt shortly after, as seen in some QEMU syscall trace templates).

Makes other misc changes to make sure the syscall trace template file passes the invariant checker: add thread exit (since we already have a thread start), relaxation of various invariant checks.

Adds and implements the instr_is_xrstor API that identifies variants of the xrstor opcode, and adds supervisor versions of xsave to instr_is_xsave.

Adds unit tests for these new scenarios. Added a TODO to handle other arch equivalent versions of these scenarios.

Adds a new flag `-abort_on_invariant_error` which is true by default, to allow the user to instruct the invariant checker to continue past invariant errors (using `-no_abort_on_invariant_error`). This is helpful since there are still a few instances of some invariant errors in the syscall trace template that are harder to generalize and fix/ignore.

Issue: #6495


clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -578,21 +587,25 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
-                             shard->stream->get_page_size() > 0),
-                        "Missing page size marker");
+        if (!TESTANY(OFFLINE_FILE_TYPE_KERNEL_SYSCALL_TRACE_TEMPLATES,
+                     shard->file_type_)) {
+            report_if_false(
+                shard,
+                !TESTANY(OFFLINE_FILE_TYPE_FILTERED | OFFLINE_FILE_TYPE_IFILTERED,
+                         shard->file_type_) ||
+                    shard->found_instr_count_marker_,
+                "Missing instr count markers");
```
derekbruening, line 597: Wait, why does this need to be disabled?  Are syscall templates marked filtered??  But they're not filtered.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -578,21 +587,25 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+            report_if_false(shard,
+                            shard->found_cache_line_size_marker_ ||
+                                (shard->skipped_instrs_ && !is_a_unit_test(shard) &&
+                                 shard->stream->get_cache_line_size() > 0),
+                            "Missing cache line marker");
```
derekbruening, line 602: Hmm, why doesn't this fire on the context switch templates too?  The unit test doesn't have them.  Oh actually the real thing does:

```
$ bin64/drrun -t drcachesim -simulator_type view -infile ../src/clients/drcachesim/tests/mock_switch_sequences.x64.zip
Output format:
<--record#-> <--instr#->: <---tid---> <record details>
------------------------------------------------------------
           1           0:     3198712 <marker: version 6>
           2           0:     3198712 <marker: filetype 0xe40>
           3           0:     3198712 <marker: cache line size 64>
           4           0:     3198712 <marker: chunk instruction count 10000000>
           5           0:     3198712 <marker: page size 4096>
...
```

Maybe the syscall ones should follow suit?  At least add a comment about doing that to avoid needing an exception here, since the switch templates don't need this exception.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -578,21 +587,25 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+            report_if_false(shard,
+                            shard->found_page_size_marker_ ||
+                                (shard->skipped_instrs_ && !is_a_unit_test(shard) &&
+                                 shard->stream->get_page_size() > 0),
+                            "Missing page size marker");
```
derekbruening, line 607: Ditto here

derekbruening, line 607: Hmm I tried running the checker and it gives an interesting error:
```
$ bin64/drrun -t drcachesim -simulator_type invariant_checker -infile ../src/clients/drcachesim/tests/mock_switch_sequences.x64.zip
Trace invariant failure in T3198712 at ref # 8 (0 instrs since timestamp 13347146067972416): Stream is_record_kernel() inaccurate
Aborted
```
Looks like that check is wrong: it needs to also handle context switches.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -705,8 +731,15 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
                 // Verify the number of read/write records matches the last
                 // operand. Skip D-filtered traces which don't have every load or
                 // store records.
-                report_if_false(shard, shard->expected_read_records_ == 0,
+                report_if_false(shard,
+                                shard->expected_read_records_ == 0 ||
+                                    // Some prefetches did not have any corresponding
+                                    // memory access in system call trace templates
+                                    // collected on QEMU.
```
derekbruening, line 738: Please elaborate: do we consider that a bug?  QEMU treats prefetch loads as not normal loads and doesn't put them into traces?  Put an issue # where this is discussed?  I see no mention in #6495.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -1029,15 +1062,37 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
                 // Skip D-filtered traces which don't have every load or store records.
-                report_if_false(shard, shard->expected_read_records_ > 0,
-                                "Too many read records");
+                report_if_false(
+                    shard,
+                    shard->expected_read_records_ >
+                        0 IF_X86(
+                            // iret pops bunch of data from the stack at interrupt
+                            // returns. Read count cannot be determined using the
+                            // decoder. System call trace templates collected on
```
derekbruening, line 1071: This doesn't sound right: the read count is statically determinable as otherwise DR's handling of iret would fail trying to find the retaddr and other args.  The decoder has a special category for it `TYPE_INDIR_VAR_REG_SIZEx3x5` so the size should be there.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -1029,15 +1062,37 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+                            // QEMU would show all reads.
+                            || (shard->between_kernel_syscall_trace_markers_ &&
+                                shard->prev_instr_.decoding.opcode == OP_iret)
+                            // Xrstor does multiple reads. Read count cannot be
+                            // determined using the decoder. System call trace
+                            // templates collected on QEMU would show all reads.
+                            || (shard->between_kernel_syscall_trace_markers_ &&
+                                shard->prev_instr_.decoding.is_xrstor)),
```
derekbruening, line 1079: If we have the opcode stored do we need these separate is_ bools?  Seems like you'd have one or the other, not both: redundant.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -1029,15 +1062,37 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
+                    shard->expected_read_records_ >
+                        0 IF_X86(
+                            // iret pops bunch of data from the stack at interrupt
+                            // returns. Read count cannot be determined using the
+                            // decoder. System call trace templates collected on
+                            // QEMU would show all reads.
+                            || (shard->between_kernel_syscall_trace_markers_ &&
+                                shard->prev_instr_.decoding.opcode == OP_iret)
+                            // Xrstor does multiple reads. Read count cannot be
+                            // determined using the decoder. System call trace
```
derekbruening, line 1076: You need a cpuid-queried value to determine the size IIRC.  It ends up being kind of like the SVE vector length: should the trace store the data in the headers?  At least add a comment on how we'd do it.  Simulators may well want that info.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -1029,15 +1062,37 @@ invariant_checker_t::parallel_shard_memref(void *shard_data, const memref_t &mem
...
             } else {
                 // Skip D-filtered traces which don't have every load or store records.
-                report_if_false(shard, shard->expected_write_records_ > 0,
-                                "Too many write records");
+                report_if_false(
+                    shard,
+                    shard->expected_write_records_ >
+                        0 IF_X86(
+                            // Xsave does multiple writes. Write count cannot be
+                            // determined using the decoder. System call trace
```
derekbruening, line 1091: Same as xrstor: add comments about how we could get the size.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -1290,7 +1347,14 @@ invariant_checker_t::check_for_pc_discontinuity(
              shard->last_signal_context_.pre_signal_instr.memref.instr.addr) ||
 #endif
         // We expect a gap on a window transition.
-        shard->window_transition_ || prev_instr.instr.type == TRACE_TYPE_INSTR_SYSENTER;
+        shard->window_transition_ ||
+        prev_instr.instr.type ==
+            TRACE_TYPE_INSTR_SYSENTER IF_X86(
+                ||
+                (shard->between_kernel_syscall_trace_markers_ &&
+                 (shard->prev_instr_.decoding.opcode == OP_hlt ||
```
derekbruening, line 1355: Explain why HLT is an exception.

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -1290,7 +1347,14 @@ invariant_checker_t::check_for_pc_discontinuity(
...
+                  // After interrupts are enabled, we may be interrupted.
+                  (shard->instrs_since_sti > 0 && shard->instrs_since_sti <= 2))));
```
derekbruening, line 1357: How did you compute 2?

derekbruening, line 1357: This could be ==0 couldn't it?  It executes the STI and before it records the next instr an interrupt arrives.

clients/drcachesim/tools/invariant_checker.h
```diff
@@ -143,6 +147,12 @@ class invariant_checker_t : public analysis_tool_t {
             uint num_memory_read_access = 0;
             uint num_memory_write_access = 0;
             addr_t branch_target = 0;
+            bool is_prefetch = false;
+            int opcode = 0;
```
derekbruening, line 151: Hmm, similar comment to before: having the opcode and then also is_syscall, is_xsave, etc. seems redundant.

clients/drcachesim/tracer/raw2trace.cpp
```diff
@@ -1479,10 +1479,11 @@ raw2trace_t::read_syscall_template_file()
         syscall_trace_template_encodings_.entry_memref_count(&entry);
         if (entry.type == TRACE_TYPE_MARKER) {
             switch (entry.size) {
-            case TRACE_MARKER_TYPE_SYSCALL:
```
derekbruening, line 1482: Will this break pre-existing template insertion?  Are there existing uses?  Does the version need bumping?

core/ir/instr_api.h
```diff
@@ -1549,10 +1549,17 @@ instr_zeroes_zmmh(instr_t *instr);
...
  */
 bool
 instr_is_xsave(instr_t *instr);
+
+DR_API
+/** Returns true if \p instr's opcode is #OP_xrstor32, #OP_xrstor64, #OP_xrstors32,
+ * or #OP_xrstors64.
+ */
+bool
+instr_is_xrstor(instr_t *instr);
```
derekbruening, line 1562: Generally we list all new API functions in the changelist.

core/ir/instr_api.h
```diff
@@ -1549,10 +1549,17 @@ instr_zeroes_zmmh(instr_t *instr);
...
 DR_API
 /** Returns true if \p instr's opcode is #OP_xsave32, #OP_xsaveopt32, #OP_xsave64,
- * #OP_xsaveopt64, #OP_xsavec32 or #OP_xsavec64.
+ * #OP_xsaveopt64, #OP_xsavec32, #OP_xsavec64, #OP_xsaves32, or #OP_xsaves64.
  */
 bool
 instr_is_xsave(instr_t *instr);
+
+DR_API
+/** Returns true if \p instr's opcode is #OP_xrstor32, #OP_xrstor64, #OP_xrstors32,
```
derekbruening, line 1558: style violation: https://dynamorio.org/page_code_style.html#autotoc_md126 "For comments starting with /∗∗, leave the rest of the first line empty, unless the entire comment is a single line. "

Please fix the instr_is_xsave one too.

derekbruening (approved):

clients/drcachesim/tools/invariant_checker.cpp
```diff
@@ -126,7 +128,8 @@ invariant_checker_t::report_if_false(per_shard_t *shard, bool condition,
                   << shard->instr_count_since_last_timestamp_
                   << " instrs since timestamp " << shard->last_timestamp_
                   << "): " << invariant_name << "\n";
-        abort();
+        if (abort_on_invariant_error_)
+            abort();
```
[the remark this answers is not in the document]
abhinav92003: > A user would expect the tool to still fail with an error at the end, instead of reporting success. 

Would you expect that to be the case even with -no_abort_on_invariant_error? Not failing in that mode seemed more natural to me since the invariant checker is only counting the errors, which still succeeds even if the trace is bad.

> That user might be us: we run invariant_checker in various tests. 

Tests continue to use the default mode that aborts on any error.

clients/drcachesim/common/options.cpp
```diff
@@ -976,5 +976,11 @@ droption_t<uint64_t> op_trim_after_timestamp(
     "Removes all records from the first TRACE_MARKER_TYPE_TIMESTAMP marker with "
     "timestamp larger than the specified value.");
 
+droption_t<bool> op_abort_on_invariant_error(
+    DROPTION_SCOPE_ALL, "abort_on_invariant_error", true,
+    "Abort invariant checker when a trace invariant error is found.",
+    "When set to true, the trace invariant checker analysis tool aborts when a trace "
+    "invariant error is found. Otherwise it prints the error and continues.");
```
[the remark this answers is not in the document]
abhinav92003: Done.

clients/drcachesim/tests/burst_syscall_inject.cpp
```diff
@@ -167,7 +171,14 @@ write_system_call_template(void *dr_context)
...
         writer,
         make_memref(READ_MEMADDR_GETTID, TRACE_TYPE_READ, opnd_size_in_bytes(OPSZ_PTR)));
+    write_trace_entry(writer,
+                      make_marker(TRACE_MARKER_TYPE_SYSCALL_TRACE_END, SYS_gettid));
 
+    // Write footer.
+    dynamorio::drmemtrace::trace_entry_t thread_exit = {
+        dynamorio::drmemtrace::TRACE_TYPE_THREAD_EXIT, 0, { /*tid=*/1 }
+    };
+    write_trace_entry(writer, thread_exit);
```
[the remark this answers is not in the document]
abhinav92003: This was discussed in a previous PR (https://github.com/DynamoRIO/dynamorio/pull/6496/files/fec4728deac3d1ba50f6c4ef2c46d25db94d260a#r1418372222 and https://github.com/DynamoRIO/dynamorio/pull/6496#discussion_r1419848311): we decided to defer zip file support because the record_reader_t support wasn't available then, and for initial experiments non-zipfile would be fine.

I'd prefer to take on the move to zipfile separately from this PR.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -3165,6 +3165,160 @@ check_kernel_syscall_trace(void)
...
+                { gen_instr(TID_A), xsaves },
+                // Multiple writes.
+                { gen_data(TID_A, false, 42, 8), nullptr },
+                { gen_data(TID_A, false, 42, 8), nullptr },
+                { gen_data(TID_A, false, 42, 8), nullptr },
+                { gen_data(TID_A, false, 42, 8), nullptr },
+                { gen_instr(TID_A), hlt },
+                // Missing nop3.
+                { gen_instr(TID_A), prefetch },
+                // Missing reads.
```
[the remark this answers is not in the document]
abhinav92003: Done.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: For some we have routines like instr_is_xrstor, instr_is_xsave. I wanted to re-use those where possible.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: ==0 points to the STI itself. We still want to report PC disc on the STI itself.

clients/drcachesim/tools/invariant_checker.h
abhinav92003: See other comment. It seems better to re-use the instr_is_x routines where available.

core/ir/instr_api.h
abhinav92003: Fixed. Also for instr_is_xsave above.

core/ir/instr_api.h
abhinav92003: Done.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: Empirically. It seemed to me that after enabling interrupts using STI, perhaps a queued interrupt is then delivered within 2 instrs.

clients/drcachesim/tracer/raw2trace.cpp
abhinav92003: It would. I wasn't as concerned about it though, because this feature is still experimental. I noticed op_syscall_template_file doesn't mention it as such though. So I added a note there about the format being prone to change.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: Well, the thing is that the cpuid value may be different for the system where the template is generated and where the trace being injected into was collected.

clients/drcachesim/tools/invariant_checker.cpp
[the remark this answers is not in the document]
derekbruening: Seems like it could go either way: on the one hand our internal version continues, and if it fails I think it might break the infrastructure running it which will assume some unrecoverable error?  OTOH upstream has no such setup and existing users may be relying on the exit code in the abort case.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: Done

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: Added comment.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: Done.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: For the following iret that is marked as having a 40 byte read, I'm seeing variable count of reads split across 8/4 byte reads, totalling to 80 or 48 bytes.

```
iret   %rsp (%rsp)[40byte] -> %rsp
```

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: I figured it is simply a case of QEMU not honoring prefetch instrs and just not issuing a real memory access.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: I modified syscall templates to have page size and cache line size.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: Done.

abhinav92003: Yeah there's a comment by that check that notes this. I started adding it to this PR but realized it's better off in a separate one.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: Added to comment.

clients/drcachesim/tools/invariant_checker.cpp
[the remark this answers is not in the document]
abhinav92003: > nd if it fails I think it might break the infrastructure running it which will assume some unrecoverable error

Right.

> OTOH upstream has no such setup and existing users may be relying on the exit code in the abort case.

I don't see the concern though. In the case where the user relies on the exit code, one error is as good as many, so they wouldn't use the -no_abort_on_invariant_error option (note that it is by default set to true) which would simply fail on the first error and avoid wasted work.

clients/drcachesim/tools/invariant_checker.cpp
abhinav92003: > as otherwise DR's handling of iret would fail

Perhaps an iret in user-space is different than one in the kernel? See https://www.felixcloutier.com/x86/iret:iretd:iretq which notes multiple different iret behavior in protected mode.

Note that the relaxation I introduced in the invariant check is only for the kernel part of the trace.

derekbruening: So a different-privilege-level return does more work: so the IR needs an update then instead of saying it's always 40 bytes.  File an issue and cite it here?

clients/drcachesim/tools/invariant_checker.cpp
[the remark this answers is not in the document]
abhinav92003: I think we're mixing up requirements for a user running the invariant checker (who may want to see all errors) and a machine doing the same (as part of a test e.g., use the returned exit code to fail).

Additionally, I don't know if the framework should fail despite the fact that the tool successfully did what it was supposed to do (find invariant errors). I would reserve that behavior for a true error, as in "I couldn't do what I was supposed to because of some fatal error".

To unblock this PR, I can change the exit code to return failure when any invariant error is found. I'll also add a comment detailing this discussion.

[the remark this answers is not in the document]
abhinav92003: The ideal end state to me is  flipping the abort_on_invariant_error flag to be false by default, and turn it on only for tests.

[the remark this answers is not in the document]
derekbruening: Hmm so we have 3 use cases?  Interactive user, test suite, and multi-tool framework?  How about a new flag fail_on_invariant_error which fails the tool at the end if it found them, and we turn that off for the framework?  I see the ambiguitiy: how to distinguish a fail b/c it found errors from an internal error type of fail.  I could see just going with your position of not failing and the user has to look for printed messages.  Sorry, don't mean to hold things up, was focusing on interactive user only.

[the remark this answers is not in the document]
abhinav92003: Okay so let me keep it as-is for now (though I can change it abort/return fail exit code at the end if you insist). I added a comment noting this.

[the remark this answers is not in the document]
abhinav92003: Done.

clients/drcachesim/tests/burst_syscall_inject.cpp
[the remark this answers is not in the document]
derekbruening: OK.

clients/drcachesim/tools/invariant_checker.cpp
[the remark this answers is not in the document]
derekbruening: OK.

clients/drcachesim/tools/invariant_checker.cpp
derekbruening: OK.

derekbruening (approved):

PR #6720, opened 2024-03-25
i#6417 AMD 32-bit: create a denylist for CI x86-32 tests.

ivankyluk:
Create a denylist for CI x86-32 tests, and move failing tests to it.

The change re-enables passing tests removed by  #6590.

Issue: #6417


suite/runsuite.cmake
```diff
@@ -98,7 +98,7 @@ if (UNIX AND NOT APPLE AND NOT ANDROID AND NOT cross_riscv64_linux_only)
     # TODO i#6417: The switch to AMD VM's for GA CI has broken many of our tests.
```
derekbruening, line 98: > The change re-enable passing tests removed by https://github.com/DynamoRIO/dynamorio/pull/6590.

grammar: s/enable/enables/

suite/runsuite_wrapper.pl
```diff
@@ -403,6 +403,7 @@
                 'code_api|client.drwrap-test-detach' => 1, # i#4593
                 'code_api|linux.thread-reset' => 1, # i#4604
                 'code_api|linux.clone-reset' => 1, # i#4604
+                'code_api|tool.drcachesim.phys-threads_SUDO' => 1, # i#6417
```
derekbruening, line 406: Why is this one separate and not on the new AMD denylist?  Is this failing on Intel too?  But you have the #6417 issue number there.

derekbruening (approved):

suite/runsuite_wrapper.pl
ivankyluk: I had the test in the denylist, and I got the following error:

CMake Error at suite/tests/CMakeLists.txt:6314 (set_tests_properties):
  set_tests_properties Can not find test to add properties to:
  code_api|tool.drcachesim.phys-threads_SUDO

I dumped TESTS (also ctest -N), and the test was not in the list of tests, so it couldn't be part of the list. Hence I added it to the ignore list. 

It wasn't failing on Intel before. Is there a way to narrow the ignore test to AMD only?

derekbruening: > I dumped TESTS (also ctest -N), and the test was not in the list of tests, so it couldn't be part of the list.

You mean that you looked at the list of tests on your local build, not on the GA CI machine (via tmate), right?  Probably you local build does not have RUN_SUDO_TESTS set (maybe also proc_supports_pt is not set).  You want to add the property under a conditional for RUN_SUDO_TESTS and proc_supports_pt.  Even better: is there a CMake function to find whether a test exists?  That would be the least fragile: make a list of tests and loop over them and check that function.

derekbruening: If we can generalize then we can get rid of the define added for the xarch test (might be able to get rid of that anyway: if the xarch build-and-test doesn't enable all the tests today how is it doing that?  we should be able to use the same mechanism to make the label conditional)

ivankyluk: Yeah, I missed the RUN_SUDO_TESTS  tests. The test is defined with -DRUN_SUDO_TESTS .

I've added a macro provided by Brett to skip unknown test.

PR #6722, opened 2024-03-26
i#6721: Read ahead in as-traced drmemtrace mode

### LINKED ISSUE #6721: record_filter produces bogus headers when run as-traced core-sharded
#6635 added core-sharded support to the record filter. It seems to mess up when running as-traced: the headers presumably of the start-idle cores have 0 for the version and filetype (the c-s-o-d flag is added to 0). Here we see one correct and one incorrect:
```
$ for i in *trace.zip; do echo $i; unzip -p $i chunk.0000 | hexdump -v -e '"%010_ax | " 6/2 "%04x " "\n"' | awk '{printf "%s %s %s %s %s%s%s%s\n", $1, $2, $3, $4, $8, $7, $6, $5}' | head -3; done
core.99.trace.zip
0000000000 | 0019 0000 0000000000000006
000000000c | 001c 000c 0000000000000006
0000000018 | 001c 0009 0000000000010e40
core.9.trace.zip
0000000000 | 0019 0000 0000000000000000
000000000c | 001c 000c 0000000000000000
0000000018 | 001c 0009 0000000000010000
```

derekbruening:
Removes the exception for avoiding reading ahead to get the version and filetype for as-traced drmemtrace mode, as we need those values for the record_filter.

Adds explicit error checking and error return up front to the record_filter for missing version info.

Adjusts the scheduler unit test's record ordinal to account for the new read-ahead.

Adds a test with a checked-in trace of hello,world (truncated to make it smaller) that ran on 4 different cores.  The test only reproduces the problem if the read-ahead in open_reader() to find the tid is removed, which I did temporarily; I kept the test as it is also a good additional general test of other corner cases.  To create a test that always reproduces in our suite we would need to combine a synthetic reader with use of the scheduler in as-traced mode, which none of our unit tests are set up for.

The fix was further tested on large inputs in an internal setup which always passes a synthetic reader, where the problem reproduces every time without the fix.

Fixes #6721


clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -3423,7 +3423,10 @@ test_replay_as_traced_from_file(const char *testdir)
     static const char *const SCHED_STRING =
         "Core #0: 1257598 \nCore #1: 1257603 \nCore #2: 1257601 \n"
         "Core #3: 1257599 => 1257604 @ <366987,87875,13331862029895453> "
-        "(<366986,87875,13331862029895453> => <1,0,0>) \n"
+        // The ordinal is really 1 ("<1,0,0>") but with the scheduler's readahead
+        // it becomes 2; easier to just check for that as trying to avoid readahead
+        // causes other problems (i#xxxx).
```
abhinav92003, line 3428: Did you want to fill/file the issue number?

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -809,10 +809,6 @@ scheduler_tmpl_t<RecordType, ReaderType>::set_initial_schedule(
     // The filetype, if present, is before the first timestamp.  If we only need the
     // filetype we avoid going as far as the timestamp.
     bool gather_filetype = options_.read_inputs_in_init;
-    // Avoid reading ahead for replay as it makes the input ords not match in tests.
```
abhinav92003, line 812: Trying to remember why we had this in the first place. The change in clients/drcachesim/tests/scheduler_unit_tests.cpp seems simple enough, but did we not want to make it before because it's not really correct (should be 1 and not 2)?

clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -3423,7 +3423,10 @@ test_replay_as_traced_from_file(const char *testdir)
     static const char *const SCHED_STRING =
         "Core #0: 1257598 \nCore #1: 1257603 \nCore #2: 1257601 \n"
         "Core #3: 1257599 => 1257604 @ <366987,87875,13331862029895453> "
-        "(<366986,87875,13331862029895453> => <1,0,0>) \n"
+        // The ordinal is really 1 ("<1,0,0>") but with the scheduler's readahead
+        // it becomes 2; easier to just check for that as trying to avoid readahead
```
abhinav92003, line 3427: This is only for the additional cores right?

abhinav92003 (approved):

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: There it is!  I knew I had left something unspecified but couldn't find it after I filed #6721.  Updated.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Yes that single unit test seems to be the sole reason.  I think at the time I didn't realize there weren't more tests beyond that and thought more test tweaks would be needed.  Also note that the ordinal is already ahead for files passed in because open_reader() reads ahead already so this is not really a change for normal upstream uses.

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: The "other problems"?  Yes, those are with start-idle cores.  Updated the comment.

Issue #6681, opened 2024-02-27
doxygen 1.9.8 build failure on page_home\copydetails


derekbruening:
doxygen 1.9.8 gives this error building the DR docs:

```
CMake Error at api/docs/CMake_rundoxygen.cmake:154 (message):
  *** doxygen failed: ***

  api/docs/home.dox:38: warning: @copybrief or
  @copydoc target 'page_home\copydetails' not found

  api/docs/home.dox:38: warning: @copybrief or
  @copydoc target 'page_home\copydetails' not found
```

derekbruening:
I have a further error on my Mac, though it's the same 1.9.8 doxygen:

```
CMake Error at api/docs/CMake_rundoxygen.cmake:338 (message):
  Cannot find "page_user_docs" menu file
```

Indeed, there aren't any page_*.js files generated!  Yet on another Linux machine with 1.9.8 those are there.

derekbruening:
git blame points to this as the purpose of that part of the CMake code massaging the menu to have the order we want on dynamorio.org: "Moves DynamoRIO Extensions from API Reference into User Docs by moving from navtreedata.js to page_user_docs.js."

derekbruening:
I believe all pieces here are now fixed.  The docs now build on my Mac and Linux machines.

PR #6724, opened 2024-03-26
i#6681: Fix Doxygen file generation for GitHub pages.

brettcoon:
Doxygen 1.9.8 changed the way javascript menu files are generated, by inlining much of the menu hierarchy that was previously written to a separate file per page.

Updates cmake rules to handle the new javascript.  Also removes extra curly braces from a doxygen page that were inadvertently triggering GitHub's Liquid template expansion and removing the intended text.

This fix was verified by building the embedded doxygen files, copying them to a
local dynamorio.github.io workspace, running a Jekyll server to host the pages,
and comparing the local site vs. the online dynamorio.org pages.

Issue: #6681

### DIFF
--- a/api/docs/CMake_rundoxygen.cmake
+++ b/api/docs/CMake_rundoxygen.cmake
@@ -256,6 +256,11 @@ if (embeddable)
   endforeach ()
 
   # We leverage the javascript menu from the non-embedded version.
+  # Menu hierarchy changes:
+  #   (1) Remove the outer "DynamoRIO" layer.
+  #   (2) Remove home page at the end.
+  #   (3) Insert an "API Reference" layer around API details.
+  #   (4) Move the "Extension API" menu inside "Build Your Own Tool".
   file(GLOB all_js ${CMAKE_CURRENT_BINARY_DIR}/../html/*.js)
   set(found_user_docs OFF)
   foreach (js ${all_js})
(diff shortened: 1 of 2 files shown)


api/docs/CMake_rundoxygen.cmake
```diff
@@ -256,6 +256,11 @@ if (embeddable)
   endforeach ()
```
derekbruening, line 256: > previously written to page_user_docs.js,

nit: It was written to a file per page: not just to this one particular page of ours.

api/docs/CMake_rundoxygen.cmake
```diff
@@ -286,7 +291,31 @@ if (embeddable)
...
+          message(FATAL_ERROR "Cannot move menu entry for \"Extension API\"")
+        endif ()
+        # Now insert the Extension API menu right above Disassembly Library.
+        if (NOT string MATCHES "Disassembly Library")
+          message(FATAL_ERROR "Cannot find menu entry for page \"Disassembly Library\"")
+        endif ()
+        string(REGEX REPLACE "(\n[^\n]+\"Disassembly Library)" "${ext_entry}\\1"
+          string "${string}")
+        # Set the user_docs flag since we found its contents here.
+        set(found_user_docs ON)
```
derekbruening, line 314: Add a sanity check !found_user_docs down on line 349 where it finds page_user_docs.js?

api/docs/CMake_rundoxygen.cmake
derekbruening, line 256: Describe briefly how this was tested?  I.e., that you actually did do a local jekyll check and aren't just guessing that the result looks right.

derekbruening (approved):
Thank you for the fix!

api/docs/CMake_rundoxygen.cmake
brettcoon: Done.

api/docs/CMake_rundoxygen.cmake
brettcoon: Updated the PR description.

Issue #6726, opened 2024-03-26
Provide control over which replayed core is which output drmemtrace stream


derekbruening:
In the drmemtrace scheduler, some sort of full control over which replayed core should be which output stream would be good, as the comment in read_traced_schedule() says:

    // We initially number the outputs according to their order in the file.
    // XXX i#5843: Should we support some direction from the user on this?  Simulation
    // may want to preserve the NUMA relationships and may need to set up its simulated
    // cores at init time, so it would prefer to partition by output stream identifier.
    // Maybe we could at least add the proposed memtrace_stream_t query for cpuid and
    // let it be called even before reading any records at all?

As a first step, we should at least sort the cpuids, as raw2trace inserts them in whatever hashtable order they were in, which is essentially random.  E.g., on 2 sockets, the order for one trace has the 2 sockets completely intermixed: 4185,70,4207,4126,4129,69,... (the top bit 4096 marks the 2nd socket).  This means the replay will not match the original order.

PR #6729, opened 2024-03-26
i#6726 replay cpuid: Sort as-traced outputs by cpuid

derekbruening:
Currently, in as-traced mode the output streams are assigned to the cores in the as-traced schedule file in file order.  But that order is essentially random, which scrambles key arrangements like which core is on which socket.  Here we sort by the recorded cpuid to recreate the same cpuid order as before.

Adds a unit test.  Also tested on larger cases with > 100 cores.

Issue: #6726


clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -1135,16 +1137,42 @@ scheduler_tmpl_t<RecordType, ReaderType>::read_traced_schedule()
...
+    // Probably raw2trace should do this in the first place, but we have many
+    // schedule files already out there so we still need a sort here.
+    std::sort(disk2index.begin(), disk2index.end(),
+              [disk2cpuid](const output_ordinal_t &l, const output_ordinal_t &r) {
+                  return disk2cpuid[l] < disk2cpuid[r];
+              });
+    // disk2index now holds the sorted index, but we need another array to
+    // store that in the disk order. E.g., if the original file held 6,2,3,7,
+    // disk2index would then hold 1,2,0,3, but we want 2,0,1,3.
+    std::vector<output_ordinal_t> disk2output(disk2index.size());
```
abhinav92003, line 1150: disk2index has size == count of entries in the schedule file. But the values in disk2index would only be as large as the count of cpus right? So we need disk2output to be only as large as the count of cpus (looking at L1152)?

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -1094,6 +1095,8 @@ scheduler_tmpl_t<RecordType, ReaderType>::read_traced_schedule()
     std::vector<std::vector<schedule_output_tracker_t>> all_sched(outputs_.size());
     // Work around i#6107 by tracking counts sorted by timestamp for each input.
     std::vector<std::vector<schedule_input_tracker_t>> input_sched(inputs_.size());
+    std::vector<output_ordinal_t> disk2index;
+    std::vector<uint64_t> disk2cpuid;
     while (options_.replay_as_traced_istream->read(reinterpret_cast<char *>(&entry),
                                                    sizeof(entry))) {
         if (entry.cpu != cur_cpu) {
```
abhinav92003, line 1102: Remind the reader that this will happen only at zip file boundaries. #cpu -1 times

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -1135,16 +1137,42 @@ scheduler_tmpl_t<RecordType, ReaderType>::read_traced_schedule()
...
+    // Sort by cpuid to get a more natural ordering.
+    // Probably raw2trace should do this in the first place, but we have many
+    // schedule files already out there so we still need a sort here.
+    std::sort(disk2index.begin(), disk2index.end(),
+              [disk2cpuid](const output_ordinal_t &l, const output_ordinal_t &r) {
+                  return disk2cpuid[l] < disk2cpuid[r];
+              });
+    // disk2index now holds the sorted index, but we need another array to
+    // store that in the disk order. E.g., if the original file held 6,2,3,7,
+    // disk2index would then hold 1,2,0,3, but we want 2,0,1,3.
```
abhinav92003, line 1149: This example should be improved. Each cpu can of course have multiple schedule entries. So disk2index may have repeated elements.

abhinav92003, line 1149: Still trying to wrap my head around this. Maybe I need to look more closely.

We have disk2output disk2index and disk2cpuid. Do all of them have the "ordinal of schedule entry on disk" as the index? If not, the identifiers could be made more descriptive to differentiate between what the index is for the respective array.

abhinav92003 (changes requested):

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -1135,16 +1137,42 @@ scheduler_tmpl_t<RecordType, ReaderType>::read_traced_schedule()
...
     if (res != sched_type_t::STATUS_SUCCESS)
         return res;
-    for (int output_idx = 0; output_idx < static_cast<output_ordinal_t>(outputs_.size());
-         ++output_idx) {
+    // Sort by cpuid to get a more natural ordering.
+    // Probably raw2trace should do this in the first place, but we have many
+    // schedule files already out there so we still need a sort here.
+    std::sort(disk2index.begin(), disk2index.end(),
+              [disk2cpuid](const output_ordinal_t &l, const output_ordinal_t &r) {
+                  return disk2cpuid[l] < disk2cpuid[r];
```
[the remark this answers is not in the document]
derekbruening: The vector only contains one entry per unique cpuid in the file: it does not contain one entry per file entry.  So its maximum is in the hundreds in a real file.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: No, disk2index has one entry per unique cpuid.

abhinav92003: Yeah noticed that later. Looks like I wasn't paying attention that L1111 and L1112 are actually inside that if-block.

clients/drcachesim/scheduler/scheduler.cpp
abhinav92003: It took me longer than I'd like to admit to completely follow this. Some suggestions.

Add narration:

disk2output[i] stores the output_idx for the cpu at ith ordinal in the cpu_schedule_file.
disk2cpuid[i] stores the actual cpu_id for the cpu at ith ordinal in the cpu_schedule_file
disk2index[i] initially stores i, but eventually (after the sorting step) stores the ordinal in the cpu_schedule_file that has the ith largest cpuid.

Maybe rename disk2index. Eventually (after the sorting step) the key isn't the ordinal in the cpu_schedule_file anymore.

Maybe combine disk2cpuid and disk2index into one struct.

Clarify what "disk" means. It means "ordinal in the cpu schedule file on disk". Maybe more verbose naming would be better. Maybe sched_idx is better.

clients/drcachesim/scheduler/scheduler.cpp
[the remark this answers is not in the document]
abhinav92003: Yeah got that.

clients/drcachesim/scheduler/scheduler.cpp
abhinav92003: Understood that it cannot. The example is fine.

abhinav92003 (approved):

clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -3410,6 +3410,93 @@ test_replay_as_traced_dup_start()
...
+        inputs[input_idx].push_back(make_timestamp(TIMESTAMP_BASE));
+        inputs[input_idx].push_back(
+            make_marker(TRACE_MARKER_TYPE_CPU_ID, CPUIDS[input_idx]));
+        for (int instr_idx = 0; instr_idx < NUM_INSTRS; ++instr_idx) {
+            inputs[input_idx].push_back(make_instr(PC_BASE + instr_idx));
+        }
+        inputs[input_idx].push_back(make_exit(tid));
+    }
+
+    // Synthesize a cpu-schedule file.
```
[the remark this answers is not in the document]
derekbruening: Done, pointed at the CPUIDS array above and put a comment there too.

clients/drcachesim/scheduler/scheduler.cpp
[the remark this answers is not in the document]
derekbruening: Reworded as "which we want to turn into".

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Done.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: It also took me a little while to figure out how to do this as well so agreed it may need more explaining.  If there weren't cross-indices pointing into all_sched I would have just sorted that (added a comment about that).

I changed s/disk/disk_ord/ and added a comment that it is the on-disk order.

Added descriptions of the vectors as suggested and elaborated the original comment a bit.

PR #6730, opened 2024-03-26
i#6727: Ignore wait records in record_filter

### LINKED ISSUE #6727: record_filter should not write out wait records during replay
#6635 added core-sharded support to the record filter. However, if run in replay mode, the TRACE_TYPE_WAIT records will be written out to the new files, which we do not want: those are artificial and are meant purely to keep the timing in the current run.

derekbruening:
Updates record_filter to ignore TRACE_MARKER_TYPE_CORE_WAIT records. These are artificial timing records: we do not want to output them, nor consider them real input records.

Adds a unit test.

Fixes #6727

### DIFF
--- a/clients/drcachesim/tests/record_filter_unit_tests.cpp
+++ b/clients/drcachesim/tests/record_filter_unit_tests.cpp
@@ -1052,6 +1052,48 @@ test_null_filter()
     return true;
 }
 
+static bool
+test_wait_filter()
+{
+    // Test that wait records (artificial timing during replay) are not preserved.
+    constexpr addr_t TID = 5;
+    constexpr addr_t PC_A = 0x1234;
+    constexpr addr_t ENCODING_A = 0x4321;
+    std::vector<test_case_t> entries = {
+        { { TRACE_TYPE_HEADER, 0, { 0x1 } }, true, { true } },
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_VERSION, { 0x2 } }, true, { true } },
+        { { TRACE_TYPE_MARKER,
+            TRACE_MARKER_TYPE_FILETYPE,
+            { OFFLINE_FILE_TYPE_ENCODINGS } },
+          true,
+          { true } },
+        { { TRACE_TYPE_THREAD, 0, { TID } }, true, { true } },
+        { { TRACE_TYPE_PID, 0, { 0x5 } }, true, { true } },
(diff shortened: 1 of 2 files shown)


clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -533,6 +533,12 @@ record_filter_t::process_markers(per_shard_t *per_shard, trace_entry_t &entry,
                        "supported";
             }
             break;
+        case TRACE_MARKER_TYPE_CORE_WAIT:
+            // These are artificial timing records: do not output them, nor consider
+            // them real input records.
+            output = false;
+            --per_shard->input_entry_count;
```
abhinav92003, line 540: Could we generalize this using is_record_synthetic?

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -1052,6 +1052,48 @@ test_null_filter()
...
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_CORE_IDLE, { 0 } }, true, { true } },
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_CORE_WAIT, { 0 } }, true, { false } },
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_CORE_IDLE, { 0 } }, true, { true } },
+        { { TRACE_TYPE_THREAD_EXIT, 0, { TID } }, true, { true } },
+        { { TRACE_TYPE_FOOTER, 0, { 0xa2 } }, true, { true } },
+    };
+    std::vector<std::unique_ptr<record_filter_func_t>> filters;
+    auto record_filter = std::unique_ptr<test_record_filter_t>(
+        new test_record_filter_t(std::move(filters), 0, /*write_archive=*/true));
+    if (!process_entries_and_check_result(record_filter.get(), entries, 0))
```
abhinav92003, line 1091: I guess we don't have an easy way to verify input_entry_count.

abhinav92003 (approved):

brettcoon (approved):
LGTM.

clients/drcachesim/tools/filter/record_filter.cpp
derekbruening: No since that includes IDLE and we do want to include IDLE for core-sharded to capture empty periods.

PR #6735, opened 2024-03-27
i#6734: Set filetype in record_filter start-idle shards

### LINKED ISSUE #6734: record_filter core-sharded start-idle cores have missing encodings
The drmemtrace record_filter fails to include encodings in core-sharded mode for outputs that start out idle.  This is detectable by running opcode_mix instead of schedule_stats in the tool.record_filter_bycore_uni test (unfortunately the invariant_checker does not yet support core-sharded: #6684).

The error turns out to be that the per_shard_t.filetype field is only set if an actual filetype record is seen: and the synthetic one written at output does not count.  Thus the encoding-handling code thinks the file doesn't have encodings at all.

derekbruening:
Sets the filetype for record_filter shards from the filetype_ var to ensure start-idle shards have the correct type.

Switches to using opcode_mix in the tool.record_filter_bycore_uni test, which fails with encoding errors without this fix.

Fixes #6734

### DIFF
--- a/clients/drcachesim/tools/filter/record_filter.cpp
+++ b/clients/drcachesim/tools/filter/record_filter.cpp
@@ -213,7 +213,7 @@ record_filter_t::initialize_shard_output(per_shard_t *per_shard,
                    shard_stream->get_shard_index(), output_ext_.c_str(), version_,
                    filetype_);
             per_shard->output_path += output_ext_;
-            per_shard->filetype = filetype_;
+            per_shard->filetype = static_cast<addr_t>(filetype_);
             lock.unlock();
         } else if (!input_name.empty()) {
             size_t last_dot = input_name.rfind('.');
@@ -233,7 +233,7 @@ record_filter_t::initialize_shard_output(per_shard_t *per_shard,
                    shard_stream->get_shard_index(), output_ext_.c_str(), version_,
                    filetype_);
             per_shard->output_path += output_ext_;
-            per_shard->filetype = filetype_;
+            per_shard->filetype = static_cast<addr_t>(filetype_);
             lock.unlock();
             input_info_cond_var_.notify_all();
         } else {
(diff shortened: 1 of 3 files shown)


derekbruening:
a64 failure looks like #3320 the type_is_instr assert: re-running now.

clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -213,6 +213,7 @@ record_filter_t::initialize_shard_output(per_shard_t *per_shard,
                    shard_stream->get_shard_index(), output_ext_.c_str(), version_,
                    filetype_);
             per_shard->output_path += output_ext_;
+            per_shard->filetype = static_cast<addr_t>(filetype_);
```
abhinav92003, line 216: Was this supposed to be part of #6722? Maybe make it part of a test?

derekbruening: It should have been part of #6704 which added these global values for the synthesized headers but didn't set them in the shards.  The test modified in this PR to use opcode_mix fails without this fix.  Do you mean, try to get a unit test to hit this instead of an end-to-end test?

derekbruening: Or do you mean, add an assert that the filetype is set at exit time or sthg?

abhinav92003: A record_filter_unit_test that verifies that filetype is set properly.

derekbruening: > A record_filter_unit_test that verifies that filetype is set properly.

That is not simple: the unit test uses mocks which do not reproduce any of the desired behavior here.  The behavior we want to test requires the filetype being available at init time in some streams (those with non-empty names) but not in others.  It also requires multiple worker threads to avoid hanging on the wait.  None of that infrastructure is in the unit test file now.  Probably the easiest thing would be to create a full analyzer and scheduler with worker threads: but that is much of the way to the end-to-end test.  WDYT?

abhinav92003: Sure since we have the end-to-end test, this is fine.

PR #6738, opened 2024-03-27
i#3544 RV64: Port stolen-reg test to RISC-V

ksco:
Partially enables stolen-reg test for RISC-V, with the synchall translation path unsupported. The following modifications/fixes were made:

1. Changed the default stolen-reg from `t3` to `s11` to be a callee-saved register;
2. Ported stolen-reg.c and stolen-reg.dll.c to support RISC-V;
3. Fixed `instr_is_mov_constant()` and `instr_is_nop()` for the pattern matching in `stolen-reg.dll.c` to work;
4. Fixed `READ_TLS_SLOT()` macro register allocation issue;
5. Added more necessary loads/restores for tp register alongside stolen register;
6. Removed duplicated tests in RUNS_ON_QEMU label;

With this patch, the stolen-reg test is now working on real hardware (without the synchall translation path), but like many other tests related to signals, it does not work on QEMU.

### DIFF
--- a/core/arch/arch.c
+++ b/core/arch/arch.c
@@ -3868,6 +3868,11 @@ set_stolen_reg_val(priv_mcontext_t *mc, reg_t newval)
 }
 
 #    ifdef RISCV64
+reg_t
+get_tp_reg_val(priv_mcontext_t *mc)
+{
+    return *(reg_t *)(((byte *)mc) + opnd_get_reg_dcontext_offs(DR_REG_TP));
+}
 void
 set_tp_reg_val(priv_mcontext_t *mc, reg_t newval)
 {
--- a/suite/tests/CMakeLists.txt
+++ b/suite/tests/CMakeLists.txt
@@ -3200,10 +3200,12 @@ if (X86) # FIXME i#1551, i#1569: port to ARM and AArch64
   set(DynamoRIO_SET_PREFERRED_BASE OFF)
 endif (X86)
 
-if (AARCHXX)
+if (AARCHXX OR RISCV64)
   tobuild_ci(client.stolen-reg client-interface/stolen-reg.c "" "" "")
   link_with_pthread(client.stolen-reg)
+endif ()
 
+if (AARCHXX)
   tobuild_ci(client.ldstex client-interface/ldstex.c "" "" "")
   link_with_pthread(client.ldstex)
 endif ()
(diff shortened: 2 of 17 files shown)


suite/tests/CMakeLists.txt
```diff
@@ -4272,7 +4272,7 @@ if (BUILD_CLIENTS)
     if (DR_HOST_X86 AND DR_HOST_X64 AND LINUX)
       torunonly_drcacheoff(opcode_categories allasm_x86_64 ""
         "@-simulator_type@opcode_mix" "")
-
+
```
ksco, line 4275: @derekbruening There are some trailing spaces that cannot be removed,  I think the trailing space checker reenabled recently should ignore diff lines starting with a `-`.

derekbruening: That makes sense: file an issue or send a PR?  (I remember the trailing space checker also complaining when checking in actual .diff files (e.g. patches for elfutils) and having to disable it during PR testing.)

ksco: Okay, I will send a PR.

suite/tests/client-interface/stolen-reg.c
```diff
@@ -223,17 +262,21 @@ main(int argc, char **argv)
 
     join_thread(thread);
 
+#ifndef RISCV64
```
derekbruening, line 265: Please make it clear this is only temporarily disabled: add something like `TODO i#3544: Add synchall support to RISC-V`

suite/tests/client-interface/stolen-reg.dll.c
```diff
@@ -94,11 +99,11 @@ read_and_restore_stolen_reg_value()
     dr_get_mcontext(drcontext, &mc);
 
     /* The key part of the test: that the modified value shows up here. */
-    DR_ASSERT(mc.IF_ARM_ELSE(r10, r28) == test_value);
+    DR_ASSERT(mc.IF_ARM_ELSE(r10, IF_AARCH64_ELSE(r28, s11)) == test_value);
```
derekbruening, line 102: Have a single define at the top of the file to avoid repeating this multi-condition sequence?

derekbruening (approved):

core/arch/arch.c
```diff
@@ -3868,6 +3868,11 @@ set_stolen_reg_val(priv_mcontext_t *mc, reg_t newval)
 }
 
 #    ifdef RISCV64
+reg_t
+get_tp_reg_val(priv_mcontext_t *mc)
```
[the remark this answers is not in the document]
ksco: Ack, I'll keep this one as is to be consistent with `get/set_stolen_reg_val()`.

Issue #6508, opened 2023-12-14
Inject kernel context switch sequences in the drmemtrace scheduler


derekbruening:
Just like we're adding canned kernel syscall sequences in #6495, we'd like to insert some roughly-representative kernel context switch sequence as well.  Like the syscalls this will probably be fed by a file, or files if we want to support different types of switches, which the scheduler will insert at switch points.  These injected records presumably will count toward output ordinals, but not toward preempt quanta, and won't have any corresponding input.

PR #6743, opened 2024-03-28
i#6508 context switches: Adjust and add invariant checks

abhinav92003:
Adjusts and adds some invariant checks for kernel context switch traces.
Specifically: adjusts the PC discontinuity check around the start and
end of the context switch trace; adds invariant checks to disallow nested
context switch traces and trace-end without trace-start.

Adds some unit tests for kernel context switch traces. Updates the
mock_switch_sequences.x64.zip file to add a thread exit before the
footer for both the context switch sequences in it. Adds a test that runs
the invariant checker on this zip file, which now passes.

Also fixes a bug in invariant error counting which should happen with
-no_abort_on_invariant_error.

Issue: #6508


clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -2941,6 +2941,69 @@ check_exit_found(void)
     return true;
```
abhinav92003, line 2941: @derekbruening When I run the invariant checker on the checked-in context switch trace file, I still see one error:

```

$ bin64/drrun -t drcachesim -simulator_type invariant_checker -infile ../../src/iX-context-switch-inv/clients/drcachesim/tests/mock_switch_sequences.x64.zip -no_abort_on_invariant_error
Trace invariant failure in T3198712 at ref # 22 (2 instrs since timestamp 13347146067972416): Thread is missing exit
Found 1 invariant errors

```

Do we want to add the thread-exit to the context switch trace, or relax the check?

abhinav92003: How do you generate mock_switch_sequences.x64.zip? I couldn't find a tool, just the test_kernel_switch_sequences test but it doesn't write it out.

clients/drcachesim/tests/invariant_checker_test.cpp
```diff
@@ -2941,6 +2941,69 @@ check_exit_found(void)
...
+            gen_marker(TID_A, TRACE_MARKER_TYPE_CACHE_LINE_SIZE, 64),
+            gen_marker(TID_A, TRACE_MARKER_TYPE_PAGE_SIZE, 4096),
+            gen_instr(TID_A, /*pc=*/1),
+            gen_marker(TID_A, TRACE_MARKER_TYPE_CONTEXT_SWITCH_START, 0),
+            gen_instr(TID_A, /*pc=*/10),
+            gen_instr(TID_A, /*pc=*/12),
+            gen_marker(TID_A, TRACE_MARKER_TYPE_CONTEXT_SWITCH_END, 0),
+            gen_instr(TID_A, /*pc=*/2),
+            gen_exit(TID_A),
+        };
```
derekbruening, line 2994: Let's add a test that looks like the checked-in switches with multiple back-to-back switch sequences with headers + footers for each but nothing else.

derekbruening (approved):

clients/drcachesim/tests/invariant_checker_test.cpp
derekbruening: I think I used a binary editor on a tiny asm trace.

abhinav92003: Done.

abhinav92003: Modified the existing mock_switch_sequences to add a thread exit before the footer, on both context switch sequences in the zip.

clients/drcachesim/tests/invariant_checker_test.cpp
abhinav92003: I added a test on the checked-in mock_switch_sequences. Seemed better that way.

abhinav92003: Or did you specifically want a unit test that looks like it? (for documenting the file format etc).

abhinav92003: Resolving and submitting. We can follow up with another PR if needed.

Issue #6745, opened 2024-03-29
drmemtrace -trace_after_instrs introduces gap between first and second timestamps


abhinav92003:
When drmemtrace tracing is done with -trace_after_instrs enabled, the timestamp in the trace buffer at thread-init is set to the then timestamp, but trace entries start pouring in only after the specified count of instrs have passed. This introduces a gap in the first and second timestamp in the trace, which is directly proportional to the requested -trace_after_instrs value.

For a `simple_app` trace collected using:
```
$ ./bin64/drrun -t drcachesim -offline -trace_after_instrs 100000 -- suite/tests/bin/simple_app
```
grepping for timestamps in the view tool output shows the diff.
```
$ grep timestamp /tmp/view.txt 
           6           0:     1922821 <marker: timestamp 13356213906110890>
       12485        9005:     1922821 <marker: timestamp 13356213907576928> <--- much after the first timestamp
       12487        9005:     1922821 <marker: timestamp 13356213907577279>
       15494       11278:     1922821 <marker: timestamp 13356213907766531>
       15497       11278:     1922821 <marker: timestamp 13356213907766846>
       21119       15375:     1922821 <marker: timestamp 13356213907800740>

```

derekbruening:
Sounds like the timestamp solutions involving freezing or using restart timestamps for windows #6104 just need to include this delayed-trace case.  Also xref the original frozen timestamps in #2039 and #5021.

PR #6746, opened 2024-03-29
i#6745: Fix timestamp gap in delayed tracing

abhinav92003:
Fixes a timestamp gap between the first and second timestamps in the trace caused when -trace_after_instrs is used.

Adds a unit test that reuses the existing window_test.cpp to reproduce the bug and fails without the fix.

Fixes: #6745


abhinav92003:
Using manual runs verified that the diff between the first and second timestamps is much reduced:

```
$ grep timestamp /tmp/view_before.txt | head -n 5
           6           0:     1966582 <marker: timestamp 13356217102311828>
       12485        9005:     1966582 <marker: timestamp 13356217103778957>
       12487        9005:     1966582 <marker: timestamp 13356217103779295>
       15494       11278:     1966582 <marker: timestamp 13356217103968370>
       15497       11278:     1966582 <marker: timestamp 13356217103968684>
$ grep timestamp /tmp/view_after.txt | head -n 5
           6           0:     1964040 <marker: timestamp 13356217068353707>
       12485        9005:     1964040 <marker: timestamp 13356217068363931>
       12487        9005:     1964040 <marker: timestamp 13356217068364279>
       15494       11278:     1964040 <marker: timestamp 13356217068536871>
       15497       11278:     1964040 <marker: timestamp 13356217068537158>

```

abhinav92003:
@derekbruening I know we solved some timestamp gap related issues before. Do we have a unit test for it that I can add to? If not, do we want one now?

derekbruening:
> @derekbruening I know we solved some timestamp gap related issues before. Do we have a unit test for it that I can add to? If not, do we want one now?

Look at the PR's for the issues I linked in #6745: I see @bete0 added a unit test in PR #6165 or at least the description claims so

clients/drcachesim/tracer/output.cpp
```diff
@@ -1059,7 +1059,7 @@ process_and_output_buffer(void *drcontext, bool skip_size_cap)
         instru->clamp_unit_header_timestamp(data->buf_base + stamp_offs, min_timestamp);
     }
 
-    if (has_tracing_windows()) {
+    if (has_tracing_windows() || op_trace_after_instrs.get_value() > 0) {
```
derekbruening, line 1062: Please add a test case, maybe augmenting the window_test added in PR #6165

abhinav92003: Done.

abhinav92003:
I found two failures for the new test: on win64 and x86-32. Still debugging.

abhinav92003:
x86-32 tool.drcachesim.opcode_mix failure is #6303

abhinav92003:
@derekbruening This is ready for review. PTAL.

clients/drcachesim/tests/window_test.cpp
```diff
@@ -159,7 +158,7 @@ test_main(int argc, const char *argv[])
         assert(status == scheduler_t::STATUS_OK);
         if (memref.marker.type == TRACE_TYPE_MARKER &&
             memref.marker.marker_type == TRACE_MARKER_TYPE_TIMESTAMP) {
-            // Looks at the gap between the current and the previous timestamps
+            // Looks at the gap between the first and second timestamps
```
derekbruening, line 161: Hmm, should it check all of them (i.e., remove the break)?

derekbruening (approved):

clients/drcachesim/tests/window_test.cpp
abhinav92003: That equates to checking that there's no sleep between any two trace timestamps, which is not true.

OTOH there _could_ be a sleep between the first two trace timestamps too. I feel this test is inherently brittle and may be working on because the -trace_after_instrs/-retrace_every_instrs/-trace_for_instrs settings are amenable.

abhinav92003: Opened #6847. Going ahead with this PR and will tackle these existing issues in #6847.

PR #6748, opened 2024-03-31
Fix trailing space checker to remove existing trailing spaces

ksco:
We rely on git diff to do trailing space checks, so to remove the trailing spaces in existing lines that may have been introduced by mistake, deletion lines should be excluded during the check.

Tested with following snippets:

```cmake
cmake_minimum_required(VERSION 3.0)
project(StringRegexMatchExample)

function(match_string_with_regex input_string regex_pattern)
    string(REGEX MATCH "${regex_pattern}" match_result "${input_string}")
    if(match_result)
        message("Match")
    else()
        message("Not match")
    endif()
endfunction()

set(regex_pattern "\n[+ ][^\n]* \n")

set(input_string "diff\n hello  \n")
match_string_with_regex("${input_string}" "${regex_pattern}")

set(input_string "diff\n+ \n")
match_string_with_regex("${input_string}" "${regex_pattern}")

set(input_string "diff\n  \n")
match_string_with_regex("${input_string}" "${regex_pattern}")

set(input_string "diff\n \n")
match_string_with_regex("${input_string}" "${regex_pattern}")

set(input_string "diff\n-hello  \n")
match_string_with_regex("${input_string}" "${regex_pattern}")

set(input_string "diff\n-  \n")
match_string_with_regex("${input_string}" "${regex_pattern}")

set(input_string "diff\n hello\n")
match_string_with_regex("${input_string}" "${regex_pattern}")

set(input_string "diff\n+hello\n-  \n")
match_string_with_regex("${input_string}" "${regex_pattern}")

set(input_string "diff\n hello\n-  \n")
match_string_with_regex("${input_string}" "${regex_pattern}")

set(input_string "diff\nhello\n")
match_string_with_regex("${input_string}" "${regex_pattern}")

set(input_string "diff\n\n")
match_string_with_regex("${input_string}" "${regex_pattern}")
```

Outputs:

```
Match
Match
Match
Not match
Not match
Not match
Not match
Not match
Not match
Not match
Not match
```

### DIFF
--- a/suite/runsuite.cmake
+++ b/suite/runsuite.cmake
@@ -316,9 +316,12 @@ endif ()
 # to get the diff and check it.  The vera++ rules do check C/C++ code.
 
 # Check for trailing space.  This is a diff with an initial column for +-,
-# so a blank line will have one space: thus we rule that out.
+# so we need to exclude the following cases:
+# 1. existing blank line will now have one space;
+# 2. lines starting with -.
+# We do this by matching lines that start with + or a space and end with a space.
 # The clang-format check will now find these in C files, but not non-C files.
-string(REGEX MATCH "[^\n] \n" match "${diff_contents}")
+string(REGEX MATCH "^[+ ].* \n" match "${diff_contents}")
 if (NOT "${match}" STREQUAL "")
   # Get more context
   string(REGEX MATCH "\n[^\n]+ \n" match "${diff_contents}")
--- a/suite/runsuite.cmake
+++ b/suite/runsuite.cmake
@@ -321,10 +321,8 @@ endif ()
 # 2. lines starting with -.
 # We do this by matching lines that start with + or a space and end with a space.
 # The clang-format check will now find these in C files, but not non-C files.
-string(REGEX MATCH "^[+ ].* \n" match "${diff_contents}")
+string(REGEX MATCH "\n[+ ][^\n]* \n" match "${diff_contents}")
 if (NOT "${match}" STREQUAL "")
-  # Get more context
-  string(REGEX MATCH "\n[^\n]+ \n" match "${diff_contents}")
   message(FATAL_ERROR "ERROR: diff contains trailing spaces: ${match}")
 endif ()
 
--- a/suite/runsuite.cmake
+++ b/suite/runsuite.cmake
@@ -321,6 +321,7 @@ endif ()
 # 2. lines starting with -.
 # We do this by matching lines that start with + or a space and end with a space.
 # The clang-format check will now find these in C files, but not non-C files.
+# The first line is always the diff header and can be safely skipped.
 string(REGEX MATCH "\n[+ ][^\n]* \n" match "${diff_contents}")
 if (NOT "${match}" STREQUAL "")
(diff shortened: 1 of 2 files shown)


suite/runsuite.cmake
```diff
@@ -316,9 +316,12 @@ endif ()
...
 
 # Check for trailing space.  This is a diff with an initial column for +-,
-# so a blank line will have one space: thus we rule that out.
+# so we need to exclude the following cases:
+# 1. existing blank line will now have one space;
+# 2. lines starting with -.
+# We do this by matching lines that start with + or a space and end with a space.
 # The clang-format check will now find these in C files, but not non-C files.
-string(REGEX MATCH "[^\n] \n" match "${diff_contents}")
+string(REGEX MATCH "^[+ ].* \n" match "${diff_contents}")
```
derekbruening, line 324: It would be good to describe how this was tested in the PR description: even including a couple small examples, to give confidence it works (e.g., unless I'm missing something with CMake's ^ handling, I assume this would fail on many small tests: maybe all tests since there's usually a header line in the diff).

derekbruening (changes requested):

suite/runsuite.cmake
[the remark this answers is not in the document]
ksco: Hmm, this is standard regex, right? I think the `[^\n]` part matches a character that is *not* a newline, so a line with exactly one space is excluded.

ksco: I did test the regex in a line-based regex evaluator and thought the CMake `string(REGEX MATCH` is also line-based. I'll do more tests in CMake to make sure this is working, sorry.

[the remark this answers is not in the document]
derekbruening: Yes right the original looks for a non-newline.

derekbruening: Unless the CMake regex has changed, I remember the ^ being fairly useless and having to check for \n all over the place in the regexes in our code.  https://groups.google.com/g/dynamorio-users/c/Yl5g5jM-NY8 says start of string, not start of line.

ksco: Thanks, yes, `^` indeed matches the start of the string, but what surprises me the most is `.` also matches `\n`. :O 
 Anyway, I submitted a fix and updated the PR description with test cases, should be good now. I should be more careful with CMake stuff.

suite/runsuite.cmake
```diff
@@ -316,12 +316,13 @@ endif ()
...
 
 # Check for trailing space.  This is a diff with an initial column for +-,
-# so a blank line will have one space: thus we rule that out.
+# so we need to exclude the following cases:
+# 1. existing blank line will now have one space;
+# 2. lines starting with -.
+# We do this by matching lines that start with + or a space and end with a space.
 # The clang-format check will now find these in C files, but not non-C files.
-string(REGEX MATCH "[^\n] \n" match "${diff_contents}")
+string(REGEX MATCH "\n[+ ][^\n]* \n" match "${diff_contents}")
```
derekbruening, line 324: This won't match the very first line but presumably there's always a header before the lines with actual content?  Maybe worth a comment.

derekbruening (approved):
Thank you for improving this check!

Issue #6712, opened 2024-03-19
Add drmemtrace replay-as-traced support to record_scheduler_t


derekbruening:
Trying to use record_filter running as-traced on a smallish app based on the threadsig sample fails:
```
Encoding size 9 != instr size 3 for PC 0x7ffa29f26ba9
```

```
[scheduler] next_record[6]: from 2 @1710797586236606: type=10 size=2 addr=0x7ffa2b0582fc
[scheduler] next_record[6]: from 2 @1710797586236631: type=47 size=6 addr=0x12c840f
[scheduler] next_record[6]: waiting because timestamp 13355270234506704 is ahead of output 0
[scheduler] next_record[6]: replay segment in=2 (@909) type=0 start=876 end=909
next_record[7]: from 2 @1710797586238158: next_record[3]: from 4 @1710797586238082: [scheduler] [record_filter] next_record[6]: advancing to input 3 instr #0
next_record[6]: replay segment in=3 (@0) type=0 start=0 end=-1
next_record[6]: from 3 @1710797586237492: type=25 size=0 addr=0x6
next_record[0]: from 1 @1710797586238563: next_record[6]: from 3 @1710797586238516: type=28 size=2 addr=0x2f728c2a350003
next_record[6]: from 3 @1710797586238712: type=28 size=9 addr=0xe40
[scheduler] next_record[6]: from 3 @1710797586238981: type=22 size=4 addr=0x41194
[scheduler] next_record[6]: from 3 @1710797586239016: type=24 size=4 addr=0x4115c
[scheduler] next_record[6]: from 3 @1710797586239212: type=28 size=10 addr=0x40
[scheduler] [scheduler] next_record[6]: from 3 @1710797586239336: type=28 size=20 addr=0x989680
[record_filter] [scheduler] next_record[6]: from 3 @1710797586239495: type=28 size=18 addr=0x1000
[scheduler] next_record[6]: from 3 @1710797586239827: [record_filter] type=28 size=2 addr=0x2f728c2a3505d0
[scheduler] next_record[3]: from 4 @1710797586239853: next_record[6]: from 3 @1710797586239920: type=28 size=3 addr=0x7
[scheduler] next_record[6]: from 3 @1710797586240088: type=47 size=3 addr=0xc08548
[scheduler] next_record[6]: from 3 @1710797586240206: type=10 size=3 addr=0x7ffa29f26ba9
```

It's the segment endpoint: have to stop before the encoding.
[...]

Issue #6733, opened 2024-03-26
Thread is missing exit invariant failure on drcachesim.invariants test


derekbruening:
The x86_64 tool.drcachesim.invariants has failed on the master merge a
number of times in the tool.drcachesim.invariants test:

Some older ones:
https://github.com/DynamoRIO/dynamorio/actions/runs/7731301531/job/21078628227
https://github.com/DynamoRIO/dynamorio/actions/runs/7732983950/job/21084129568

A few more recent ones too.
https://github.com/DynamoRIO/dynamorio/actions/runs/8442624559/job/23124442264

```
Trace invariant failure in T27434 at ref # 373883 (0 instrs since timestamp 13351202897274636): Thread is missing exit
```

Locally I can reproduce:
```
$ ctest --repeat-until-fail 500 -R sim.invariants
```
Failed after 29 successes.

edeiana:
Also failed on AARCH64 ("aarch64-precommit (ubuntu-20-arm64-sve)") one time (so far) in #6691 (https://productionresultssa14.blob.core.windows.net/actions-results/7de917d3-3760-4c4e-8868-17a821dcbe85/workflow-job-run-aec6087b-1676-5d99-51e5-fc3628aaa01c/logs/job/job-logs.txt?rsct=text%2Fplain&se=2024-03-27T03%3A15%3A40Z&sig=4uo4KtcH6yeXG%2BApsmdkKeOQhQJNsAmoHkwgo4riuNE%3D&sp=r&spr=https&sr=b&st=2024-03-27T03%3A05%3A35Z&sv=2021-12-02).

edeiana:
Sometimes it also times out (tested on x86_64):
```
$ ctest --repeat-until-fail 500 -R sim.invariants
Test project /OMITTED/dynamorio/build/dynamorio_inv_bug/build64
    Start 330: code_api|tool.drcachesim.invariants
    Test #330: code_api|tool.drcachesim.invariants ...***Timeout 180.02 sec

0% tests passed, 1 tests failed out of 1

Total Test time (real) = 180.10 sec

The following tests FAILED:
	330 - code_api|tool.drcachesim.invariants (Timeout)
Errors while running CTest
Output from these tests are in: /OMITTED/dynamorio/build/dynamorio_inv_bug/build64/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
```

derekbruening:
[...]

PR #6749, opened 2024-04-02
i#6712 record bounds: Add record filter sanity checks

derekbruening:
Adds two sanity checks developed to identify what was at first believed to be a new bug but turned out to be #6712:

+ Ensure that last_encoding is empty on an input switch, to avoid recording the wrong encoding in pc2encoding.

+ Ensure the encoding size matches the instr size when inserting chunk-new encodings.

+ Augment the reader check for mismatching encoding vs instr sizes, and abort in release build too.

Although #6712 is fixed, these checks are still useful to prevent regressions.  Plus, the reader_t check detects other bugs as well and we know at least one other is out there in #6303.  Making the reader check abort in release build should help avoid wasted work as already happened here where we didn't notice the printed warnings in release build.

Tested by running record_filter on large proprietary inputs: no checks failed.

Issue: #6712


clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -603,8 +603,16 @@ record_filter_t::process_chunk_encodings(per_shard_t *per_shard, trace_entry_t &
             VPRINT(this, 3,
                    "output new-chunk encoding chunk=%" PRIu64 " ref=%" PRIu64 "\n",
                    per_shard->chunk_ordinal, per_shard->cur_refs);
-            if (!write_trace_entries(per_shard,
-                                     per_shard->per_input->pc2encoding[entry.addr])) {
+            // Sanity check that the encoding size is correct.
+            const auto &enc = per_shard->per_input->pc2encoding[entry.addr];
+            size_t enc_sz = 0;
+            for (const auto &record : enc)
+                enc_sz += record.size;
```
abhinav92003, line 610: We could figure this out  without looping: `(enc.size() - 1) * enc[0].size + enc[enc.size()-1].size`

abhinav92003 (approved):

clients/drcachesim/reader/reader.cpp
```diff
@@ -204,8 +204,12 @@ reader_t::process_input_entry()
...
                 if (last_encoding_.size != cur_ref_.instr.size) {
-                    ERRMSG("Encoding size %zu != instr size %zu for PC 0x%zx\n",
-                           last_encoding_.size, cur_ref_.instr.size, cur_ref_.instr.addr);
+                    ERRMSG(
+                        "Encoding size %zu != instr size %zu for PC 0x%zx at ord %" PRIu64
+                        " instr %" PRIu64 " last_timestamp=0x%" PRIx64 "\n",
+                        last_encoding_.size, cur_ref_.instr.size, cur_ref_.instr.addr,
+                        get_record_ordinal(), get_instruction_ordinal(),
+                        get_last_timestamp());
                     assert(false);
```
abhinav92003, line 213: I suppose this assert doesn't fail in release build?

clients/drcachesim/reader/reader.cpp
```diff
@@ -204,8 +204,12 @@ reader_t::process_input_entry()
             // Look for encoding bits that belong to this instr.
```
abhinav92003, line 204: Is this same as the cause of i#6303? So we don't have a fix yet?

clients/drcachesim/reader/reader.cpp
derekbruening: No, and in fact that caused headaches where we didn't notice a problem until much later.  Maybe this should be made a release-build fatal error too.

abhinav92003: This PR only adds checks and failure messages. So the underlying issue that may cause the assert to be hit is still there?

clients/drcachesim/reader/reader.cpp
abhinav92003: > You're asking whether the record-filter-caused encoding problems like #6712 are related to #6303? 

This PR also adds an ERRMSG in reader code, so wasn't sure if there's any connection to #6303. Thanks for clarifying.

clients/drcachesim/tools/filter/record_filter.cpp
derekbruening: OK, though I'm not sure it's more readable?  The loop has just 1 or 2 iters nearly all the time.

abhinav92003: Fine either way.

clients/drcachesim/reader/reader.cpp
derekbruening: The record_scheduler_t spitting of encodings from instrs is the underlying issue for what I was using these extra checks to find: I only later found that out.  It was fixed in #6712.  But these checks still seem good to have, hence this PR, both in case of regressions but also this reader failure one must have multiple possible causes b/c #6303 is still open.  Updated the PR description.

derekbruening: It actually is not easy to fail here in release build as there is no error propagation path.  We could directly call abort() like the assert does in debug build.  Maybe that is the best approach, better than setting at_eof or sthg artificial and easy to mis-interpret.  I think having an abort in release will help us; it would have saved me hours or possibly days of work.  Done by adding `assert_release_too` which aborts in release too, for this and the encoding error right below.

clients/drcachesim/reader/reader.cpp
derekbruening: The reader ERRMSG was one way we detected a problem.  It is now a release-build abort so it should help us find such problems sooner in the future instead of days later in this case where we wasted time thinking things were fine.

clients/drcachesim/tools/filter/record_filter.cpp
derekbruening: Added a comment.

derekbruening:
a64 sve failure is dr$sim.invariants missing exit: #6733

PR #6752, opened 2024-04-03
i#6495 syscall templates: Avoid buffer overflow at instr_no_fetch

abhinav92003:
Writes out the buffered entries at each instr_no_fetch instance. This is to avoid overflowing the raw2trace write buffer, since some syscall trace templates have repstr with a very high iteration count.

Adds separate tracking in the basic counts tool for user and kernel nofetch instrs.

Adds a unit test to the existing burst_syscall_inject test that verifies injection of a syscall template with a high repstr iteration count. Also removes some log spam in the burst_syscall_inject test.

Issue: #6495


clients/drcachesim/tests/burst_syscall_inject.cpp
```diff
@@ -102,24 +103,13 @@ write_instr_entry(void *dr_context, std::unique_ptr<std::ostream> &writer, instr
...
-    write_trace_entry(
-        writer,
-        make_instr(reinterpret_cast<addr_t>(instr_app_pc), TRACE_TYPE_INSTR, len));
+    write_trace_entry(writer,
+                      make_instr(reinterpret_cast<addr_t>(instr_app_pc), type, len));
 }
 
-static std::string
-write_system_call_template(void *dr_context)
+void
```
derekbruening, line 110: This lost its static decl.  I guess in C++ if in an anon namespace it's not needed: but I don't see an anon namespace in this file.

The write_footer_entries below could also be static.

derekbruening (approved):

clients/drcachesim/tests/burst_syscall_inject.cpp
abhinav92003: Fixed. Also for various others in this file.

PR #6753, opened 2024-04-03
i#5036 A64 scatter/gather, part 9: State restore

jackgallagher-arm:
Adds support for restoring register app state when a fault is triggered by one of the instructions in a scatter/gather expansion sequence emitted by drx_expand_scatter_gather().

Also extends the existing scatter/gather expansion tests to check the value of the FFR register, and run the test instructions with different elements rigged to fault so we can test the state restoration behaviour.

Support for the first-faulting and non-faulting variants (ldff1*, ldnf1*) will be added in future commits.

Issue: #5036


jackgallagher-arm:
@derekbruening I have modified the scatter/gather test app to trigger scatter/gather instructions to fault so we can test the state restore behaviour. If the instruction faults the signal handler in the test app increments the PC to skip the faulting scatter/gather instruction, but now it fails with this error message from the invariant checker:

```
Trace invariant failure in T3924776 at ref # 2650074 (1 instrs since timestamp 13356630469438203): Signal handler return point incorrect
```

If I remove the check from `clients/drcachesim/tools/invariant_checker.cpp`, the test works correctly and passes. Is there a way I can disable that check for this test?

derekbruening:
> @derekbruening I have modified the scatter/gather test app to trigger scatter/gather instructions to fault so we can test the state restore behaviour. If the instruction faults the signal handler in the test app increments the PC to skip the faulting scatter/gather instruction, but now it fails with this error message from the invariant checker:
> 
> ```
> Trace invariant failure in T3924776 at ref # 2650074 (1 instrs since timestamp 13356630469438203): Signal handler return point incorrect
> ```
> 
> If I remove the check from `clients/drcachesim/tools/invariant_checker.cpp`, the test works correctly and passes. Is there a way I can disable that check for this test?

So the invariant is deliberately violated as it skips the faulting instruction rather than re-executing it.  There is a `test_name` parameter to the checker which can be used for things like this: see `knob_test_name_` in the code.  I think it's passed on the cmdline with `-test_mode_name`: you can see it being set in other existing tests.

AssadHashmi (approved):

PR #6725, opened 2024-03-26
i#5365 AArch64 SVE core, part 2: add signals support

AssadHashmi:
This patch adds SVE support for signals in the core. It is the follow on patch from the SVE core work part 1, in PR #5835 (f646a632d) and includes vector address computation for SVE scatter/gather, enabling first-fault load handling.

Issue: #5365, #5036

Co-authored-by: Jack Gallagher <jack.gallagher@arm.com>

PR #6754, opened 2024-04-03
i#6685 core-shard: Add idles and waits to basic_counts

derekbruening:
Adds counts of idle and wait records to basic_counts, which are useful when examining core-sharded-on-disk traces.

Rewords "scheduling markers" to "timestamp + cpuid markers" to clarify. Rewords "transfer markers" to "kernel transfer markers".

Updates the tests; the core-serial tests non-zero output here, and the core-sharded test also manually tested non-zero output as it required removing these counts for reliable comparison of record vs replay.

Issue: #6685


derekbruening:
Failure is record_filter_bycore_multi due to an inaccurate pi estimate print by the app on a64 sve: #6755

derekbruening:
(waiting on an issue with PR #6725 affecting our internal tests...will merge once that is resolved)

Issue #131, opened 2014-11-27
!HAVE_TLS causes debug builds to fail


derekbruening:
_From [bharath....@gmail.com](https://code.google.com/u/100705261247615317067/) on April 22, 2009 12:38:23_

What steps will reproduce the problem? Run any app that forks with a debug build of DR compiled with !HAVE_TLS. What is the expected output? What do you see instead? Instead of regular fork behavior the following tend to happen.
1. ASSERT(dcontext == get_thread_private_dcontext()) triggers.  If ignored,
2. KSTAT_DC(thread_measured) in kstats_thread_exit() fails assert about
inconsistent TOS for kstats stack.

Both of them are caused by the fact that get_thread_private_dcontext()
returns NULL in the child process when it enters dispatch() post_syscall
(i.e., post fork).  This is because the tid used to lookup tls_table is the
child's but the value in the table is still the parent's.  This is fixed in
replace_thread_tid() called in post_system_call() for fork.  So beginning
in dispatch post syscall till replace_thread_tid()
get_thread_private_dcontext() returns NULL.  This affects the assert
mentioned above and kstats.  In addition it affects a log of logging
(opening log files, writing to them, dumping core, etc.).

We have 2 choices, fix all the failure points to not depend upon dcontext,
i.e., TLS or fix !HAVE_TLS to get tid fixed in tls_table either in
post_syscall stub or on entry in dispatch().

Till then the work around is to ignore this assert and turn of kstats.

_Original issue: http://code.google.com/p/dynamorio/issues/detail?id=131_

PR #6756, opened 2024-04-04
i#5036 A64 scatter/gather, part 10: Non-fault loads

jackgallagher-arm:
Adds support for non-fault loads to drx_expand_scatter_gather(). Non-fault loads (ldnf1*) work similarly to scalar+immediate predicated contiguous ld1* loads, but with different behaviour if an element access faults. This commit implements this behaviour and extends the scatter/gather tests to include ldnf1* instructions.

Issue: #5036


jackgallagher-arm:
`client.detach_test` is timing out repeatedly on the `vs2019-32` build. I can't see anything in this patch which would affect that test so I suspect the failure is spurious.

```
201/262 Test #131: code_api|client.detach_test ..................................***Timeout  90.02 sec
```

derekbruening:
> `client.detach_test` is timing out repeatedly on the `vs2019-32` build. I can't see anything in this patch which would affect that test so I suspect the failure is spurious.
> 
> ```
> 201/262 Test #131: code_api|client.detach_test ..................................***Timeout  90.02 sec
> ```

As with any test failure, please file a bug and cite it here or cite an existing bug if one is already filed.

Issue #6755, opened 2024-04-04
tool.record_filter_bycore_multi failured due to inaccurate pi estimation


derekbruening:
The tool.record_filter_bycore_multi test failed on a64 sve at https://github.com/DynamoRIO/dynamorio/actions/runs/8546804195/job/23417839931?pr=6754

The output mismatch looks like it's b/c it estimated pi at 3.94... instead of 3.14...

```
2024-04-03T23:32:41.7151484Z 309: Running postcmd4 |/opt/actions-runner/_work/dynamorio/dynamorio/build/build_debug-internal-64/clients/bin64/drcachesim;-indir;tool.record_filter_bycore_multi.filtered.dir;-simulator_type;schedule_stats|
2024-04-03T23:32:41.7153291Z 309: CMake Error at /opt/actions-runner/_work/dynamorio/dynamorio/suite/tests/runmulti.cmake:138 (message):
2024-04-03T23:32:41.7154282Z 309:   output |Estimation of pi is 3.943208188861861
2024-04-03T23:32:41.7154698Z 309: 
2024-04-03T23:32:41.7154952Z 309:   Trace invariant checks passed
2024-04-03T23:32:41.7155312Z 309: 
2024-04-03T23:32:41.7155609Z 309:   Output 2098798 entries from 2098250 entries.
2024-04-03T23:32:41.7156026Z 309: 
2024-04-03T23:32:41.7156271Z 309:   Schedule stats tool results:
2024-04-03T23:32:41.7156619Z 309: 
2024-04-03T23:32:41.7156846Z 309:   Total counts:
2024-04-03T23:32:41.7157122Z 309: 
2024-04-03T23:32:41.7157346Z 309:              3 cores
2024-04-03T23:32:41.7157645Z 309:              3 threads
2024-04-03T23:32:41.7157968Z 309:         215938 instructions
2024-04-03T23:32:41.7158346Z 309:              9 total context switches
2024-04-03T23:32:41.7158880Z 309:      0.0416786 CSPKI (context switches per 1000 instructions)
2024-04-03T23:32:41.7159455Z 309:          23993 instructions per context switch
2024-04-03T23:32:41.7160100Z 309:              9 voluntary context switches
2024-04-03T23:32:41.7160551Z 309:              0 direct context switches
2024-04-03T23:32:41.7160965Z 309:         100.00% voluntary switches
2024-04-03T23:32:41.7161356Z 309:           0.00% direct switches
[...]

PR #6757, opened 2024-04-05
i#6758: Fix AArch64 FP state at fcache events and attach

### LINKED ISSUE #6758: linux.signalNNN1 test failing due to non-determinism in floating point operations
Splitting out from #6755 since we're not sure how much they overlap.

In downstream systems we're seeing that the linux.signalNNN1 tests are failing due to non-determinism in the floating point result of some trig math ops. This for some reason does not affect the test run on Github Actions.

These failures started since https://github.com/DynamoRIO/dynamorio/pull/6725.

abhinav92003:
Fixes the slot used to save and restore FP regs at fcache enter and return events. PR #6725 adjusted the slots used during signal handling in core/unix/signal_linux_aarch64.c but did not adjust the same in fcache enter/return and attach events. Prior to that PR, the FP regs were simply stored in a contiguous manner in signal handling code and fcache enter/return routines (instead of in their respective dr_simd_t struct member), which was a bit confusing.

The mismatch between slot usage in signal handling and fcache enter/return code caused failures in the signalNNN1 tests on some systems. Verified that those tests pass with this fix.

Also fixes the same issue in save_priv_mcontext_helper which is used in the dr_app_start API. Unit tests for this scenario will be added as part of #6759.

Issue: #5036, #6755, #5365
Fixes #6758


core/arch/aarch64/aarch64.asm
```diff
@@ -226,16 +226,42 @@ save_priv_mcontext_helper:
...
-        st1      {v8.2d-v11.2d}, [x4], #64
-        st1      {v12.2d-v15.2d}, [x4], #64
-        st1      {v16.2d-v19.2d}, [x4], #64
-        st1      {v20.2d-v23.2d}, [x4], #64
-        st1      {v24.2d-v27.2d}, [x4], #64
-        st1      {v28.2d-v31.2d}, [x4], #64
-        /* TODO i#5365: Save Z/P regs as well? Will require runtime check of
-         * ID_AA64PFR0_EL1 for FEAT_SVE.
+
+        /* Registers Q0-Q31 map directly to registers V0-V31. */
```
abhinav92003, line 230: @derekbruening Do you know if any existing test provides coverage for this change? Probably not because it didn't fail on/after #6725 (but then so didn't signalNNN1 on Github Actions). Maybe we need to port detach_state to AArch64 and add more state verifications to it?

abhinav92003:
This PR is probably not ready completely yet, but would be good to get some quick feedback since this is blocking downstream pulls.

core/arch/aarch64/emit_utils.c
```diff
@@ -778,20 +781,24 @@ append_save_simd_reg(dcontext_t *dcontext, instrlist_t *ilist, bool absolute)
...
-            INSTR_CREATE_stp(
+            INSTR_CREATE_str(
                 dcontext,
-                opnd_create_base_disp(DR_REG_X1, DR_REG_NULL, 0, i * 16, OPSZ_32),
-                opnd_create_reg(DR_REG_Q0 + i), opnd_create_reg(DR_REG_Q0 + i + 1)));
+                opnd_create_base_disp(DR_REG_X1, DR_REG_NULL, 0, i * 64, OPSZ_16),
+                opnd_create_reg(DR_REG_Q0 + i)));
     }
     if (proc_has_feature(FEATURE_SVE)) {
+        /* TODO i#5036: Ensure that the following uses the correct offset for the
```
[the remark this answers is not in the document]
derekbruening: DR supports a max SVE vec len of 512-bit which is 64 bytes (see comment above dr_simd_t def).

core/arch/aarch64/aarch64.asm
```diff
@@ -1,5 +1,5 @@
 /* **********************************************************
```
derekbruening, line 1: I would file a separate dedicated bug on the signal test failures we are seeing: I think that will be easier to understand and track this fix and have something to point at for what is blocking us.

derekbruening, line 1: > to that PR the FP regs were simply stored in a continuous manner in signal handling code, and fcache enter/exit routines 

s/continuous/contiguous/
s/code, and/code and/

core/arch/aarch64/aarch64.asm
derekbruening: client.cleancall has some mcontext field checks: but it looks like it was never ported to aarchxxx.  Yes, getting more testing parity on aarchxx would be good.

core/arch/aarch64/emit_utils.c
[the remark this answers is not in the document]
abhinav92003: Ah thanks for clarifying. I made the changes to the sve code too. PTAL.

core/arch/aarch64/aarch64.asm
abhinav92003: Filed #6758.

abhinav92003: Done.

core/arch/aarch64/aarch64.asm
abhinav92003: But let's not block this PR on that I assume?

derekbruening: Agreed.

core/arch/aarch64/aarch64.asm
```diff
@@ -226,16 +226,42 @@ save_priv_mcontext_helper:
...
-        st1      {v12.2d-v15.2d}, [x4], #64
-        st1      {v16.2d-v19.2d}, [x4], #64
-        st1      {v20.2d-v23.2d}, [x4], #64
-        st1      {v24.2d-v27.2d}, [x4], #64
-        st1      {v28.2d-v31.2d}, [x4], #64
-        /* TODO i#5365: Save Z/P regs as well? Will require runtime check of
-         * ID_AA64PFR0_EL1 for FEAT_SVE.
+
+        /* Registers Q0-Q31 map directly to registers V0-V31. */
+        str      q0, [x4], #64
```
derekbruening, line 231: In addition to assuming the vec len is the full 64 bytes, was the ST1 also storing just one D (half-size element) instead of the whole V==Q?

core/arch/aarch64/emit_utils.c
```diff
@@ -605,8 +604,7 @@ append_restore_simd_reg(dcontext_t *dcontext, instrlist_t *ilist, bool absolute)
                 INSTR_CREATE_ldr(
                     dcontext, opnd_create_reg(DR_REG_P0 + i),
                     opnd_create_base_disp(
-                        DR_REG_X1, DR_REG_NULL, 0,
-                        i * (proc_get_vector_length_bytes() / 8),
+                        DR_REG_X1, DR_REG_NULL, 0, i * sizeof(dr_simd_t),
```
derekbruening, line 607: Wait, are the pred regs stored in a dr_simd_t?  Indeed they are in mcxt_api.h: but aren't they 1/8 the size?  Isn't that a waste of space?  @AssadHashmi could you help us understand that?

derekbruening (approved):
The change makes sense to me but probably best to have @AssadHashmi take a look as I am not an expert in a64 SIMD: as you can tell I'm confused by the svep field being a full dr_simd_t and other things.

core/arch/aarch64/emit_utils.c
jackgallagher-arm, line 607: Lines `634` and `848` also load `P15` as part of saving/restoring the `FFR` register so those need to be updated too.

`/15 * (proc_get_vector_length_bytes() / 8)/15 * sizeof(dr_simd_t)/`

I think this is what is causing the `P15` corruption errors in the scatter/gather test.

jackgallagher-arm (changes requested):

core/arch/aarch64/emit_utils.c
abhinav92003: Done.

abhinav92003: PTAL

jackgallagher-arm (approved):

core/arch/aarch64/aarch64.asm
abhinav92003: As I understand the previous version did vx.2d which means 2 64-bit floating point = 128 bits. And it stored 4 such together (==64 bytes) in one st1 op. @jackgallagher-arm to confirm.

core/arch/aarch64/emit_utils.c
abhinav92003: @jackgallagher-arm any thoughts on this?

core/arch/aarch64/aarch64.asm
jackgallagher-arm: > As I understand the previous version did vx.2d which means 2 64-bit floating point = 128 bits. And it stored 4 such together (==64 bytes) in one st1 op. @jackgallagher-arm to confirm

That is correct.

core/arch/aarch64/emit_utils.c
AssadHashmi: Yes, predicates and FFR are 1/8th of the size. The idea was to get SVE working first then make such refinements, and this was a quick way of getting things up and running. Should we raise an issue to fix this or do you want to do it as part of this patch?

abhinav92003: Separate issue would be better, since this PR is needed to unblock our downstream merge.

core/arch/aarch64/aarch64.asm
AssadHashmi: I've created https://github.com/DynamoRIO/dynamorio/issues/6759 to improve coverage. If you have any other suggestions of x86 tests not ported to AArch64 which will catch bugs like this, please add them to that issue.

abhinav92003:
code_api|tool.drcov.eintr failed on ubuntu-20-arm64-sve. I see it did so also on #6756 though so very likely unrelated to this PR itself.

x86-32 tool.drcachesim.opcode_mix failure is #6303.

core/arch/aarch64/emit_utils.c
AssadHashmi: Created https://github.com/DynamoRIO/dynamorio/issues/6760.

AssadHashmi (approved):

abhinav92003:
#6761 is fixed now. Rerunning tests.

core/arch/aarch64/emit_utils.c
abhinav92003: Thanks!

core/arch/aarch64/aarch64.asm
abhinav92003: Thanks for creating #6759!

core/arch/aarch64/emit_utils.c
derekbruening: Note that such a change in the mcontext layout would break compatibility and require a major version change.

core/arch/aarch64/aarch64.asm
derekbruening: OK, so there are 2 variants of `ST1`: single-lane and multi-lane, all under the same `ST1` opcode.  But why call it `ST1` for multi-lane?  I thought the `1` meant "one lane".

core/arch/aarch64/emit_utils.c
derekbruening: Oh you mean PR #6725 changes the size of dr_simd_t!  That didn't sink in when I reviewed it.  Yes, it breaks compatibility and it should have bumped the minor version to 90 as part of the same PR per https://dynamorio.org/page_new_release.html#autotoc_md230 and clearly documented the breakage in the release notes.  @AssadHashmi please submit a follow-up PR for that.

derekbruening: Wait, PR #6725 doesn't change the size: the new union field member is the same 64 bytes.

@abhinav92003 please clarify what you mean about breaking compatibility

abhinav92003: PR #6725 doesn't break compatibility. But it sets the data differently in the existing field. I assume that also needs to be backward compatible (not just the struct definition).

abhinav92003: Changing how we set data is set in the existing same-size field also would require clients to be updated. They would need to read the struct accordingly.

derekbruening: Before, the SIMD data was written in a way that didn't even line up with the dr_simd_t structs, right?  Someone trying to manually pull the data out would be very confused.  In fact: was there a users list question about that with someone not able to get the data?  This seems like a bug fix.  Maybe someone figured out the way it was written before and wrote client code assuming that but it seems like it's not the same kind of change where we broke a documented interface.

abhinav92003: Right. That's what I meant.

> Though it really fixes it by spilling to the slots one would expect

Issue #6761, opened 2024-04-05
tool.drcov.eintr failing on ubuntu-20-arm64-sve


abhinav92003:
The tool.drcov.eintr test seems to have started failing on  ubuntu-20-arm64-sve. This was observed on #6756, #6757, and #6746.

The test logs mention an output mismatch, but I don't see it.

```
2024-04-04T21:14:23.5101279Z 206: CMake Error at /opt/actions-runner/_work/dynamorio/dynamorio/clients/drcov/runtest.cmake:95 (message):
2024-04-04T21:14:23.5102190Z 206:   tool output
2024-04-04T21:14:23.5102819Z 206:   SF:/opt/actions-runner/_work/dynamorio/dynamorio/suite/tests/linux/eintr.c
2024-04-04T21:14:23.5103404Z 206: 
2024-04-04T21:14:23.5103627Z 206:   DA:52,1
2024-04-04T21:14:23.5103973Z 206: 
2024-04-04T21:14:23.5104272Z 206:   DA:53,1
2024-04-04T21:14:23.5112911Z 206: 
2024-04-04T21:14:23.5113309Z 206:   DA:56,1
2024-04-04T21:14:23.5113565Z 206: 
2024-04-04T21:14:23.5113778Z 206:   DA:57,1
2024-04-04T21:14:23.5114021Z 206: 
2024-04-04T21:14:23.5114233Z 206:   DA:58,1
2024-04-04T21:14:23.5114469Z 206: 
2024-04-04T21:14:23.5114684Z 206:   DA:59,1
2024-04-04T21:14:23.5114917Z 206: 
2024-04-04T21:14:23.5115152Z 206:   DA:60,1
2024-04-04T21:14:23.5115389Z 206: 
2024-04-04T21:14:23.5115602Z 206:   DA:64,1
2024-04-04T21:14:23.5115840Z 206: 
2024-04-04T21:14:23.5116046Z 206:   DA:65,1
2024-04-04T21:14:23.5116283Z 206: 
2024-04-04T21:14:23.5116502Z 206:   DA:67,1
2024-04-04T21:14:23.5116738Z 206: 
2024-04-04T21:14:23.5116956Z 206:   DA:69,1
2024-04-04T21:14:23.5117190Z 206: 
2024-04-04T21:14:23.5117402Z 206:   DA:70,1
2024-04-04T21:14:23.5117641Z 206: 
2024-04-04T21:14:23.5117852Z 206:   DA:71,1
[...]

PR #6762, opened 2024-04-05
i#6761: Fix tool.drcov.eintr expected output

abhinav92003:
Allows either a covered or not-covered verdict for lines that may or may
not be executed depending on where signals are delivered in the test app.

It is better to adapt the expected output template than to modify the test
code because there's nothing really wrong with the test code (no race
since it already uses locks where required).

Fixes: #6761


suite/tests/linux/tool.drcov.eintr.templatex
```diff
@@ -40,7 +40,7 @@ DA:121,1
 DA:122,1
 DA:124,1
 DA:125,1
-DA:126,1
+DA:126,[01]
```
jackgallagher-arm, line 43: Line 105 in `eintr.c` has a very similar loop so it might suffer from a similar race condition. Should we update the expectation for that line too to avoid a repeat of this issue with that line?

jackgallagher-arm (approved):

suite/tests/linux/tool.drcov.eintr.templatex
abhinav92003: Definitely, thanks for highlighting that.

derekbruening (commented):
LGTM

PR #6765, opened 2024-04-09
i#3544 RV64: Port sample.memtrace_simple to RISC-V

ksco:
Implements `decode_opcode_name()`, which is needed by the newly enabled test. Also limits `-max_bb_instrs` further to 8 on RISC-V for its heavyweight instrumentation.

This is yet another test that does not work on QEMU, thus we cannot enable it in CI.

Issue: #3544


ksco:
Running `ctest -j4` on a real machine gives:

```
49% tests passed, 110 tests failed out of 216
```

That is almost half of the tests we enabled for compiling, still a long way to go.

suite/tests/CMakeLists.txt
```diff
@@ -3518,8 +3518,13 @@ if (BUILD_SAMPLES)
     foreach (sample ${sample_list})
       # FIXME i#4392: Remove sanity checks and replace with functioning tests like
       # that done for the opcode_count and memtrace_simple samples.
-      torunonly_ci(sample.${sample} common.${control_flags}
-        ${sample} common/${control_flags}.c "" "" "")
+      if (RISCV64 AND sample STREQUAL "memtrace_simple")
+        torunonly_ci(sample.${sample} common.${control_flags}
+          ${sample} common/${control_flags}.c "" "-max_bb_instrs 8" "")
```
derekbruening, line 3523: (I wonder if we should lower the default in the core -- or mention this in the documentation that for RISC-V for heavyweight tools this may need to be as low as 8.)

derekbruening (approved):

suite/tests/CMakeLists.txt
ksco: The problem is that for heavier usage, this number can be even lower. The extreme scenario we set 15 for in the core is that each instruction has at most 1 clean call and nothing else, but memtrace_simple exceeds this assumption.

We haven't done any RISC-V-related documentation at present. Maybe add a TODO at -max_bb_instrs option in the core first?

ksco: Added the TODO.

ksco: Could you point me to where the "clean call helper callout functions" are implemented for other architectures? I'm not sure what it is.

ksco: I see, thank you! This would indeed alleviate the overhead of clean call context switch, I will implement this soon.

PR #6770, opened 2024-04-11
i#5843 scheduler: Add option to disable direct switches

derekbruening:
Adds a new scheduler option field honor_direct_switches and a corresponding command-line parameter -sched_disable_direct_switches to allow a way to disable direct thread switches, primarily for scheduling experimentation.

Adds a unit test.

Issue #5843

### DIFF
--- a/clients/drcachesim/common/options.cpp
+++ b/clients/drcachesim/common/options.cpp
@@ -920,6 +920,14 @@ droption_t<bool> op_sched_randomize(
     "set), and FIFO order and instead selects the next input randomly. "
     "This is intended for experimental use in sensitivity studies.");
 
+droption_t<bool> op_sched_disable_direct_switches(
+    DROPTION_SCOPE_FRONTEND, "sched_disable_direct_switches", false,
+    "Ignore direct thread switch requests",
+    "Applies to -core_sharded and -core_serial.  Disables switching to the recorded "
+    "targets of TRACE_MARKER_TYPE_DIRECT_THREAD_SWITCH system call metadata markers "
+    "and causes the associated system call to be treated like any other call with a "
+    "switch being determined by latency and the next input in the queue.");
+
 // Schedule_stats options.
 droption_t<uint64_t>
     op_schedule_stats_print_every(DROPTION_SCOPE_ALL, "schedule_stats_print_every",
--- a/clients/drcachesim/tests/scheduler_unit_tests.cpp
+++ b/clients/drcachesim/tests/scheduler_unit_tests.cpp
@@ -3794,42 +3794,83 @@ test_direct_switch()
         make_marker(TRACE_MARKER_TYPE_DIRECT_THREAD_SWITCH, TID_BASE + 3),
         make_exit(TID_C),
     };
-    std::vector<scheduler_t::input_reader_t> readers;
-    readers.emplace_back(std::unique_ptr<mock_reader_t>(new mock_reader_t(refs_A)),
-                         std::unique_ptr<mock_reader_t>(new mock_reader_t()), TID_A);
-    readers.emplace_back(std::unique_ptr<mock_reader_t>(new mock_reader_t(refs_B)),
-                         std::unique_ptr<mock_reader_t>(new mock_reader_t()), TID_B);
-    readers.emplace_back(std::unique_ptr<mock_reader_t>(new mock_reader_t(refs_C)),
-                         std::unique_ptr<mock_reader_t>(new mock_reader_t()), TID_C);
-    // The string constructor writes "." for markers.
-    // We expect A's first switch to be to C even though B has an earlier timestamp.
-    // We expect C's direct switch to A to proceed immediately even though A still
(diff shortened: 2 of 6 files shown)


derekbruening:
x32 failure is dr$sim.opcode_mix encoding size #6303

clients/drcachesim/common/options.cpp
```diff
@@ -920,6 +920,14 @@ droption_t<bool> op_sched_randomize(
     "set), and FIFO order and instead selects the next input randomly. "
     "This is intended for experimental use in sensitivity studies.");
 
+droption_t<bool> op_sched_disable_direct_switches(
```
abhinav92003, line 923: I suppose it was on purpose that this is a "disable" flag? Generally "enable" flags are easier for humans to read and parse. Optional: would it help if it were sched_enable_direct_switches, with default set to true?

clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -3794,42 +3794,83 @@ test_direct_switch()
...
+                             std::unique_ptr<mock_reader_t>(new mock_reader_t()), TID_B);
+        readers.emplace_back(std::unique_ptr<mock_reader_t>(new mock_reader_t(refs_C)),
+                             std::unique_ptr<mock_reader_t>(new mock_reader_t()), TID_C);
+        // The string constructor writes "." for markers.
+        // We expect A's first switch to be to C even though B has an earlier timestamp.
+        // We expect C's direct switch to A to proceed immediately even though A still
+        // has significant blocked time left.  But then after B is scheduled and finishes,
+        // we still have to wait for C's block time so we see idle underscores:
+        static const char *const CORE0_SCHED_STRING =
+            "...AA.........CC......A....BBBB.______________C...";
```
abhinav92003, line 3812: Are TRACE_MARKER_TYPE_DIRECT_THREAD_SWITCH markers hidden from the stream when -sched_disable_direct_switches? Probably not; do they need to be?

clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -3794,42 +3794,83 @@ test_direct_switch()
...
+                             std::unique_ptr<mock_reader_t>(new mock_reader_t()), TID_A);
+        readers.emplace_back(std::unique_ptr<mock_reader_t>(new mock_reader_t(refs_B)),
+                             std::unique_ptr<mock_reader_t>(new mock_reader_t()), TID_B);
+        readers.emplace_back(std::unique_ptr<mock_reader_t>(new mock_reader_t(refs_C)),
+                             std::unique_ptr<mock_reader_t>(new mock_reader_t()), TID_C);
+        // The string constructor writes "." for markers.
+        // We expect A's first switch to be to B with an earlier timestamp.
+        // We expect C's direct switch to A to not happen until A's blocked time ends.
+        static const char *const CORE0_SCHED_STRING =
+            "...AA.........BBBB....CC......___________________C...___A.";
```
abhinav92003, line 3849: Why is there so much idle period (the `_`s)? `A` is presumably ready at that point and can be executed right (I'm assuming so because in the previous test, the 3rd `A` is seen much earlier than the end)? Or do the timestamp quantum based context switches don't allow that?

abhinav92003 (approved):

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: No, and this does cause minor issues like schedule_stats mis-interpreting a switch that happens to occur after a direct marker from being caused by the direct request even though direct are disabled.  But I think we just live with that.

clients/drcachesim/common/options.cpp
derekbruening: OTOH it is natural for humans to think that specifying an option changes the default: so if it's named --enable it sounds like it's not enabled unless the flag is specified.  The other factor is when options like this are stored in structures like protocol buffers which sometimes do not have an easy non-zero default, where having 0 be the default is just easier to avoid having differing defaults in command-line options versus unit test tools using structures.

derekbruening: Continuing on the protocol buffer discussion: having that variation in defaults has caused headaches in the past.

abhinav92003: SGTM

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: A has a long blocked time which is cut short by the direct switch in the first scenario (the comment up there does point that out).

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: Added a note to the CLI and scheduler option docs saying that the markers are left in the trace.

Issue #6751, opened 2024-04-03
Pretty printer for #trace_entry_t-format traces


edeiana:
Traces on disk are a sequence of 12 bytes entries of type #trace_entry_t.
Currently we visualize such traces in binary form as described here: https://dynamorio.org/page_debug_memtrace.html#autotoc_md136.

We want to implement a tool similar to `view` (in `clients/drcachesim/tools/view.cpp`) to "pretty print" on-disk-format-traces.
The reason we cannot use the `view` tool is that it operates on a higher level representation of traces (#memref_t described here: https://dynamorio.org/sec_drcachesim_format.html) and not #trace_entry_t.

Since tools can operate on both #memref_t (https://dynamorio.org/classdynamorio_1_1drmemtrace_1_1analysis__tool__tmpl__t.html) and #trace_entry_t (https://dynamorio.org/classdynamorio_1_1drmemtrace_1_1analysis__tool__tmpl__t.html) representation of traces, we believe a visualization tool for #trace_entry_t traces will be helpful for developers working on tools that manipulate #trace_entry_t traces.

derekbruening:
Note that trace_entry_t was originally not meant to be a public interface at all.  Even now with the record_filter_t tool operating on it, nearly all uses should be at the memref_t level.  We want to discourage new tools like converters from our format to other formats from using trace_entry_t as they will end up duplicating caching of encodings + v2p + other non-repeated items or will make mistakes; that is one reason we never wanted it to be well-documented/publicized/supported externally.

derekbruening:
Pasting from https://github.com/DynamoRIO/dynamorio/pull/6771#discussion_r1571715972

trace_entry_t is not meant to be a public interface with guaranteed stability. The public interface is the memref_t record and reader library and the libraries on top of those like the analysis tools. trace_entry_t is supposed to be treated more like a black box by users. Its individual fields are not doxygen-commented.

Hence, providing a pretty printer seems to send the wrong message, encouraging users to peer inside the black box. We would prefer things like converters to other simulator formats to use memref_t + reader_t whose abstraction layer hides a number of messy details.

The record_filter tool can be used while keeping trace_entry_t a black box, as basically all of its filters operate above the memref_t abstraction level: remove marker types; trim to timestamp values; apply cache filters. So it is a little different.

derekbruening:
Further observations:

[...]

PR #6771, opened 2024-04-11
i#6751 trace_entry_t: pretty printer

edeiana:
Skeleton for record_view, a new tool to visualize trace_entry_t records.
Currently it prints the type of all trace_entry_t in a trace.

Issue: #6751

### DIFF
--- a/clients/drcachesim/CMakeLists.txt
+++ b/clients/drcachesim/CMakeLists.txt
@@ -165,6 +165,7 @@ add_exported_library(drmemtrace_opcode_mix STATIC tools/opcode_mix.cpp)
 add_exported_library(drmemtrace_syscall_mix STATIC tools/syscall_mix.cpp)
 add_exported_library(drmemtrace_view STATIC tools/view.cpp)
 add_exported_library(drmemtrace_func_view STATIC tools/func_view.cpp)
+add_exported_library(drmemtrace_record_view STATIC tools/record_view.cpp)
 add_exported_library(drmemtrace_invariant_checker STATIC tools/invariant_checker.cpp)
 add_exported_library(drmemtrace_schedule_stats STATIC tools/schedule_stats.cpp)
 
@@ -172,6 +173,7 @@ target_link_libraries(drmemtrace_invariant_checker drdecode)
 
 configure_DynamoRIO_standalone(drmemtrace_opcode_mix)
 configure_DynamoRIO_standalone(drmemtrace_view)
+configure_DynamoRIO_standalone(drmemtrace_record_view)
 configure_DynamoRIO_standalone(drmemtrace_invariant_checker)
 
 # We combine the cache and TLB simulators as they share code already.
@@ -276,8 +278,8 @@ configure_DynamoRIO_standalone(drcachesim)
 target_link_libraries(drcachesim drmemtrace_simulator drmemtrace_reuse_distance
   drmemtrace_histogram drmemtrace_reuse_time drmemtrace_basic_counts
   drmemtrace_opcode_mix drmemtrace_syscall_mix drmemtrace_view drmemtrace_func_view
-  drmemtrace_raw2trace directory_iterator drmemtrace_invariant_checker
-  drmemtrace_schedule_stats drmemtrace_record_filter)
+  drmemtrace_record_view drmemtrace_raw2trace directory_iterator
+  drmemtrace_invariant_checker drmemtrace_schedule_stats drmemtrace_record_filter)
 if (UNIX)
     target_link_libraries(drcachesim dl)
 endif ()
@@ -357,6 +359,7 @@ install_client_nonDR_header(drmemtrace simulator/cache_simulator_create.h)
 install_client_nonDR_header(drmemtrace simulator/tlb_simulator_create.h)
 install_client_nonDR_header(drmemtrace tools/view_create.h)
 install_client_nonDR_header(drmemtrace tools/func_view_create.h)
+install_client_nonDR_header(drmemtrace tools/record_view_create.h)
 # TODO i#6412: Create a separate directory for non-tracer headers so that
 # we can more cleanly separate tracer and raw2trace code.
 install_client_nonDR_header(drmemtrace tracer/raw2trace.h)
@@ -578,6 +581,7 @@ restore_nonclient_flags(drmemtrace_opcode_mix)
 restore_nonclient_flags(drmemtrace_syscall_mix)
 restore_nonclient_flags(drmemtrace_view)
 restore_nonclient_flags(drmemtrace_func_view)
+restore_nonclient_flags(drmemtrace_record_view)
 restore_nonclient_flags(drmemtrace_record_filter)
 restore_nonclient_flags(drmemtrace_analyzer)
 restore_nonclient_flags(drmemtrace_invariant_checker)
@@ -644,6 +648,7 @@ add_win32_flags(drmemtrace_opcode_mix)
 add_win32_flags(drmemtrace_syscall_mix)
 add_win32_flags(drmemtrace_view)
 add_win32_flags(drmemtrace_func_view)
+add_win32_flags(drmemtrace_record_view)
 add_win32_flags(drmemtrace_record_filter)
 add_win32_flags(drmemtrace_analyzer)
 add_win32_flags(drmemtrace_invariant_checker)
@@ -821,8 +826,9 @@ if (BUILD_TESTS)
     drmemtrace_raw2trace drmemtrace_simulator drmemtrace_reuse_distance
     drmemtrace_histogram drmemtrace_reuse_time drmemtrace_basic_counts
     drmemtrace_opcode_mix drmemtrace_syscall_mix drmemtrace_view drmemtrace_func_view
-    drmemtrace_raw2trace directory_iterator drmemtrace_invariant_checker
-    drmemtrace_schedule_stats drmemtrace_analyzer drmemtrace_record_filter)
+    drmemtrace_record_view drmemtrace_raw2trace directory_iterator
+    drmemtrace_invariant_checker drmemtrace_schedule_stats drmemtrace_analyzer
+    drmemtrace_record_filter)
   if (UNIX)
     target_link_libraries(tool.drcachesim.core_sharded dl)
   endif ()
--- a/clients/drcachesim/common/options.h
+++ b/clients/drcachesim/common/options.h
@@ -49,6 +49,7 @@
 #define SYSCALL_MIX "syscall_mix"
 #define VIEW "view"
 #define FUNC_VIEW "func_view"
+#define RECORD_VIEW "record_view"
 #define INVARIANT_CHECKER "invariant_checker"
 #define SCHEDULE_STATS "schedule_stats"
 #define RECORD_FILTER "record_filter"
--- a/clients/drcachesim/tools/record_view.cpp
+++ b/clients/drcachesim/tools/record_view.cpp
@@ -1,0 +1,147 @@
+/* **********************************************************
+ * Copyright (c) 2024 Google, Inc.  All rights reserved.
+ * **********************************************************/
+
+/*
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice,
+ *   this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of Google, Inc. nor the names of its contributors may be
+ *   used to endorse or promote products derived from this software without
+ *   specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL VMWARE, INC. OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+#include "record_view.h"
+
+#include <inttypes.h>
+#include <stdint.h>
+
+#include <cstdio>
+#include <fstream>
+#include <iostream>
+#include <memory>
+#include <mutex>
+#include <string>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+#ifdef HAS_ZLIB
+#    include "common/gzip_ostream.h"
+#endif
+#ifdef HAS_ZIP
+#    include "common/zipfile_ostream.h"
+#endif
+#include "memref.h"
+#include "memtrace_stream.h"
+#include "raw2trace_shared.h"
+#include "trace_entry.h"
+#include "utils.h"
+
+#undef VPRINT
+#ifdef DEBUG
+#    define VPRINT(reader, level, ...)                            \
+        do {                                                      \
+            if ((reader)->verbosity_ >= (level)) {                \
+                fprintf(stderr, "%s ", (reader)->output_prefix_); \
+                fprintf(stderr, __VA_ARGS__);                     \
+            }                                                     \
+        } while (0)
+// clang-format off
+#    define UNUSED(x) /* nothing */
+// clang-format on
+#else
+#    define VPRINT(reader, level, ...) /* nothing */
+#    define UNUSED(x) ((void)(x))
+#endif
+
+namespace dynamorio {
+namespace drmemtrace {
+
+record_analysis_tool_t *
+record_view_tool_create(void)
+{
+    return new dynamorio::drmemtrace::record_view_t();
+}
+
+record_view_t::record_view_t()
+{
+}
+
+record_view_t::~record_view_t()
+{
+}
+
+bool
+record_view_t::parallel_shard_supported()
+{
+    return false;
+}
+
+std::string
+record_view_t::initialize_shard_type(shard_type_t shard_type)
+{
+    return "";
+}
+
+void *
+record_view_t::parallel_shard_init_stream(int shard_index, void *worker_data,
+                                          memtrace_stream_t *shard_stream)
+{
+    return shard_stream;
+}
+
+bool
+record_view_t::parallel_shard_exit(void *shard_data)
+{
+    return true;
+}
+
+std::string
+record_view_t::parallel_shard_error(void *shard_data)
+{
+    return "";
+}
+
+bool
+record_view_t::parallel_shard_memref(void *shard_data, const trace_entry_t &entry)
+{
+    const char *trace_type_name = trace_type_names[entry.type];
+    std::cerr << std::string(trace_type_name) << "\n";
+
+    return true;
+}
+
+bool
+record_view_t::process_memref(const trace_entry_t &entry)
+{
+    return parallel_shard_memref(NULL, entry);
+}
+
+bool
+record_view_t::print_results()
+{
+    return true;
+}
+
+} // namespace drmemtrace
+} // namespace dynamorio
--- a/clients/drcachesim/tools/record_view.cpp
+++ b/clients/drcachesim/tools/record_view.cpp
@@ -78,17 +78,28 @@ namespace dynamorio {
 namespace drmemtrace {
 
 record_analysis_tool_t *
-record_view_tool_create(void)
+record_view_tool_create(uint64_t sim_refs)
 {
-    return new dynamorio::drmemtrace::record_view_t();
+    return new dynamorio::drmemtrace::record_view_t(sim_refs);
 }
 
-record_view_t::record_view_t()
+record_view_t::record_view_t(uint64_t sim_refs)
+    : sim_refs_(sim_refs)
 {
 }
 
 record_view_t::~record_view_t()
 {
+}
+
+bool
+record_view_t::should_skip(void)
+{
+    if (sim_refs_ > 0) {
+        --sim_refs_;
+        return false;
+    }
+    return true;
 }
 
 bool
@@ -125,6 +136,9 @@ record_view_t::parallel_shard_error(void *shard_data)
 bool
 record_view_t::parallel_shard_memref(void *shard_data, const trace_entry_t &entry)
 {
+    if (should_skip())
+        return true;
+
     const char *trace_type_name = trace_type_names[entry.type];
     std::cerr << std::string(trace_type_name) << "\n";
 
--- a/clients/drcachesim/tools/record_view.h
+++ b/clients/drcachesim/tools/record_view.h
@@ -1,0 +1,95 @@
+/* **********************************************************
+ * Copyright (c) 2024 Google, Inc.  All rights reserved.
+ * **********************************************************/
+
+/*
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * * Redistributions of source code must retain the above copyright notice,
+ *   this list of conditions and the following disclaimer.
+ *
+ * * Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * * Neither the name of Google, Inc. nor the names of its contributors may be
+ *   used to endorse or promote products derived from this software without
+ *   specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL VMWARE, INC. OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+#ifndef _RECORD_VIEW_H_
+#define _RECORD_VIEW_H_ 1
+
+#include <stdint.h>
+
+#include <condition_variable>
+#include <memory>
+#include <mutex>
+#include <ostream>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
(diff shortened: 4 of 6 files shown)


clients/drcachesim/common/trace_entry.cpp
```diff
@@ -88,5 +88,55 @@ const char *const trace_type_names[] = {
...
+    "<unknown>",
+    "<unknown>",
+    "trace_entry_version_no_kernel_pc",
+    "trace_entry_version_kernel_pc",
+    "trace_entry_version_encodings",
+    "trace_entry_version_branch_info",
+    "trace_entry_version_frequent_timestamps",
+};
+
+const char *const trace_marker_names[] = {
```
abhinav92003, line 101: Would we want the strings at clients/drcachesim/tools/view.cpp to be replaced by these? Maybe. Add a TODO at view.cpp.

clients/drcachesim/tools/record_view.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+    trace_type_t trace_type = (trace_type_t)entry.type;
+    if (trace_type == TRACE_TYPE_INVALID) {
+        std::cerr << "ERROR: trace_entry_t invalid.\n";
+        return false;
+    }
+
+    std::string trace_type_name = std::string(trace_type_names[trace_type]);
+
+    /* Large if-else for all TRACE_TYPE_.  Prints one line per trace_entry_t.
```
abhinav92003, line 119: Replacing this with a switch-case may be more readable.

clients/drcachesim/tools/record_view.cpp
```diff
@@ -0,0 +1,215 @@
...
+        std::cerr << trace_type_name << "\n";
+    } else if ((trace_type == TRACE_TYPE_THREAD) ||
+               (trace_type == TRACE_TYPE_THREAD_EXIT)) {
+        uint tid = (uint)entry.addr;
+        std::cerr << trace_type_name << ", tid: " << tid << "\n";
+    } else if (trace_type == TRACE_TYPE_PID) {
+        uint pid = (uint)entry.addr;
+        std::cerr << trace_type_name << ", pid: " << pid << "\n";
+    } else if (trace_type == TRACE_TYPE_MARKER) {
+        /* XXX i#6751: we have a lot of different types of markers; we should use some
```
abhinav92003, line 138: clients/drcachesim/tools/view.cpp already has an exhaustive switch-case. It can be shared with this tool.

clients/drcachesim/tools/record_view.cpp
```diff
@@ -0,0 +1,215 @@
...
+                  << ", num_instructions_in_bundle: " << num_instructions_in_bundle
+                  << ", instrs_length:";
+        /* Print length of each instr in the bundle.
+         */
+        for (int i = 0; i < num_instructions_in_bundle; ++i) {
+            unsigned char instr_length = entry.length[i];
+            std::cerr << " " << instr_length;
+        }
+        std::cerr << "\n";
+    } else if (type_is_instr(trace_type)) {
```
abhinav92003, line 185: clients/drcachesim/tools/view.cpp disambiguates various kinds of instrs. We should share that logic with this tool.

clients/drcachesim/tools/record_view.cpp
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    std::string trace_type_name = std::string(trace_type_names[trace_type]);
+
+    /* Large if-else for all TRACE_TYPE_.  Prints one line per trace_entry_t.
+     * In some cases we use some helper functions (e.g., type_is_instr()), which group
+     * similar TRACE_TYPE_ together, otherwise we compare trace_type against one or two
+     * specific TRACE_TYPE_ directly.
+     */
+    if (trace_type == TRACE_TYPE_HEADER) {
```
abhinav92003, line 124: We should try to share more code with clients/drcachesim/tools/view.cpp where possible; e.g. see other comments.

clients/drcachesim/tools/record_view_create.h
```diff
@@ -0,0 +1,58 @@
...
+namespace drmemtrace {
+
+/**
+ * @file drmemtrace/record_view_create.h
+ * @brief DrMemtrace record view trace analysis tool creation.
+ */
+
+/**
+ * Creates an analysis tool which prints out the #trace_entry_t records in a trace file
+ * to std::err.
```
abhinav92003, line 50: s/std::err/std::cerr/

clients/drcachesim/CMakeLists.txt
```diff
@@ -165,13 +165,15 @@ add_exported_library(drmemtrace_opcode_mix STATIC tools/opcode_mix.cpp)
 add_exported_library(drmemtrace_syscall_mix STATIC tools/syscall_mix.cpp)
 add_exported_library(drmemtrace_view STATIC tools/view.cpp)
```
abhinav92003, line 166: Please add a unit test similar to clients/drcachesim/tests/view_test.cpp.

Also some test that verifies the output for some checked in raw trace (e.g. clients/drcachesim/tests/drmemtrace.threadsig.aarch64.raw/raw) like clients/drcachesim/tests/offline-view.templatex.

clients/drcachesim/common/trace_entry.cpp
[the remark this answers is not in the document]
abhinav92003: To a comment I mean.

clients/drcachesim/common/options.h
```diff
@@ -49,6 +49,7 @@
 #define SYSCALL_MIX "syscall_mix"
 #define VIEW "view"
 #define FUNC_VIEW "func_view"
+#define RECORD_VIEW "record_view"
```
derekbruening, line 52: I see the docs listing the supported tools for -simulator_type were not updated (nor the .dox files listing example usage of all the tools) as it normally would be for a new tool: which may well be the right decision as I do not think this tool should be publicized or documented as we really do not want external users operating on trace_entry_t.  But please make that clear with a comment here as otherwise it looks like a mistake.

clients/drcachesim/tools/record_view.h
```diff
@@ -0,0 +1,92 @@
...
+#include <string>
+
+#include "analysis_tool.h"
+#include "memtrace_stream.h"
+#include "trace_entry.h"
+
+namespace dynamorio {
+namespace drmemtrace {
+
+/**
```
derekbruening, line 47: I have mixed feelings about adding this tool: part of me does not want to add it as we do not want to encourage users to create tools that operate on the trace_entry_t format.  We consider that format subject to change and to not be a public interface.  It has once-only data that requires caching that easily leads to errors if users operate on it.  We have already seen external converters such as a champsim converter prototype incorrectly use this format.

What I'm getting as is should this have Doxygen comments for public docs?

clients/drcachesim/tools/record_view.cpp
```diff
@@ -0,0 +1,215 @@
...
+        } else { // For all remaining markers we print their value in hex.
+            std::cerr << trace_type_name << ", trace_marker_type: " << trace_marker_name
+                      << ", trace_marker_value: 0x" << std::hex << trace_marker_value
+                      << std::dec << "\n";
+        }
+    } else if (trace_type == TRACE_TYPE_ENCODING) {
+        unsigned short num_encoding_bytes = entry.size;
+        std::cerr << trace_type_name << ", num_encoding_bytes: " << num_encoding_bytes
+                  << ", encoding_bytes: 0x" << std::hex;
+        /* Print encoding byte by byte (little-endian).
```
derekbruening, line 166: This is not how encodings should be printed for AArch{32,64}.

derekbruening (commented):
I have mixed feelings about adding this tool: part of me does not want to add it as we do not want to encourage users to create tools that operate on the trace_entry_t format. We consider that format subject to change and to not be a public interface. It has once-only data that requires caching that easily leads to errors if users operate on it. We have already seen external converters such as a champsim converter prototype incorrectly use this format.

clients/drcachesim/CMakeLists.txt
```diff
@@ -165,13 +165,15 @@ add_exported_library(drmemtrace_opcode_mix STATIC tools/opcode_mix.cpp)
 add_exported_library(drmemtrace_syscall_mix STATIC tools/syscall_mix.cpp)
```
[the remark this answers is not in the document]
edeiana: Yes, I mixed errs() from LLVM and std:cerr.
Fixed, thanks!

clients/drcachesim/tools/record_view.cpp
```diff
@@ -0,0 +1,215 @@
...
+typedef unsigned int uint;
+
+record_analysis_tool_t *
+record_view_tool_create(uint64_t sim_refs)
+{
+    return new dynamorio::drmemtrace::record_view_t(sim_refs);
+}
+
+record_view_t::record_view_t(uint64_t sim_refs)
+    : sim_refs_(sim_refs)
```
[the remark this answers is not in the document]
edeiana: Done.

clients/drcachesim/common/trace_entry.cpp
derekbruening: That would change all the output recorded in the docs here and in the sample traces repo and in internal docs so would not be taken lightly.

derekbruening: The reverse might be an easier thing to do.  See all the comments about sharing code: eliminate this array and use the existing public pretty-printer's marker printing code.

clients/drcachesim/tools/record_view.cpp
edeiana: Mhm, it's actually just renaming trace_type_names[], which is weird, because every other tool uses trace_type_names[], but view.cpp does not.
Should we switch this around and have view.cpp use trace_type_names[]?
Or change trace_type_names[] with what view.cpp prints?
Or does this break too much documentation and/or other assumptions I'm not aware of?
(If we decide to go this way, the aforementioned changes should go in a separate PR IMHO).

clients/drcachesim/tools/record_view.cpp
edeiana: Ok.
Although not all of it can be shared.
view.cpp does some delayed printing and extra processing, which would defeat the purpose of "record_view": we want one line per trace_entry_t to show the exact sequence of on-disk records, we don't want to do the extra processing of view.cpp, it should be a 1 to 1 "mirror" of the trace file, just human readable.

clients/drcachesim/tools/record_view.cpp
edeiana: That would make it even larger though, and we'd be basically replicating utility functions like: type_is_intr() or type_has_address().

clients/drcachesim/tools/record_view_create.h
edeiana: Done.

clients/drcachesim/tools/record_view.h
edeiana: Oh, I thought tools that operate on trace_entry_t were public for some reason.
My vote is for giving "record_view" the same visibility as "record_filter" (which is also a tool that operates on trace_entry_t).
"record_filter" has doxygen comments for public docs, so I followed the same for "record_view".

If tools that operate on trace_entry_t are not supposed to be public, then we should remove the public documentation for all of them and transform doxygen comments in regular comments.
(Separate PR for "record_filter"?)

clients/drcachesim/common/trace_entry.cpp
```diff
@@ -88,5 +88,55 @@ const char *const trace_type_names[] = {
     "untaken_jump",
 };
 
+const char *const trace_version_names[] = {
```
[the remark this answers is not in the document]
edeiana: As in, for example: "trace_entry_version_no_kernel_pc" -> "trace_entry_version_no_kernel_pc_2" ? (2 is the enum value)
Although I'm not quite understanding why the enum value matters in the string representation of TRACE_ENTRY_VERSION_ enum. They have different names after all.
Unless you're referring to trace_version_names[0] (and [1]), which are <unknown>? If so, you mean I should just differentiate those 2?

clients/drcachesim/common/trace_entry.cpp
edeiana: https://github.com/DynamoRIO/dynamorio/pull/6771#discussion_r1571292467

clients/drcachesim/tools/record_view.h
derekbruening: No, trace_entry_t is not meant to be a public interface with guaranteed stability.  The public interface is the memref_t record and reader library and the libraries on top of those like the analysis tools.  trace_entry_t is supposed to be treated more like a black box by users.  Its individual fields are not doxygen-commented.

Hence, providing a pretty printer seems to send the wrong message, encouraging users to peer inside the black box.  We would prefer things like converters to other simulator formats to use memref_t + reader_t whose abstraction layer hides a number of messy details.

The record_filter tool can be used while keeping trace_entry_t a black box, as basically all of its filters operate above the memref_t abstraction level: remove marker types; trim to timestamp values; apply cache filters.  So it is different from a trace_entry_t pretty printer.

clients/drcachesim/common/trace_entry.cpp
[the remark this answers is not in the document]
abhinav92003: > Although I'm not quite understanding why the enum value matters in the string representation of TRACE_ENTRY_VERSION_ enum

I'm considering a use case where the user wants to know whether their trace has a version more recent than some other version (some of our tools want a minimum trace version). Having the enum value would be easier, but I guess the user can always look up the enum value themselves using just the enum string. So it's fine either way I guess.

clients/drcachesim/tools/record_view.cpp
abhinav92003: > That would make it even larger though

Do you mean there'd be more lines of code? I'd still suggest that a switch fits the bill better. It makes it clear that we're only doing equality checks. Also: https://stackoverflow.com/questions/427760/when-to-use-if-else-if-else-over-switch-statements-and-vice-versa.

 > we'd be basically replicating utility functions like: type_is_intr() or type_has_address().

Can you clarify this concern? I don't understand.

clients/drcachesim/tools/record_view.h
abhinav92003: What are the use cases of the record_view tool? The one I had in mind was debugging invariant errors in the raw trace. But that is likely more a trace-dev use case, rather than for trace users.

derekbruening: > The one I had in mind was debugging invariant errors in the raw trace.

You mean the underlying trace_entry_t file on disk and not the offline "raw" un-pre-processed files, right?

derekbruening: Along that line of thought: I would see a trace_entry_t pretty-printer along the same lines as a raw pretty-printer: used only by trace-internal devs and not by any trace gatherers/users.

abhinav92003: > You mean the underlying trace_entry_t file on disk and not the offline "raw" un-pre-processed files, right?

Well both can be useful for debugging invariant errors. I agree that both are internal use cases. But does that mean we should've add the record_view at all? Or maybe just make it a separate executable that's visibly marked internal (instead of supporting `-simulator_type record_view`).

derekbruening: Making it clear it's internal seems best; maybe even in the name: undocumented publicly, but can trigger via `-simulator_type internal_record_view`.  And it prints a header every time saying "The internal record format is not a public interface and is subject to change without notice; meant for debugging only; please do not rely on the details here" or somesuch.

clients/drcachesim/common/trace_entry.cpp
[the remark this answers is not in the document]
xdje42: Can we print the enum value along side the string?

clients/drcachesim/tools/record_view.h
xdje42: Adding "internal" to the name SGTM.

clients/drcachesim/common/trace_entry.cpp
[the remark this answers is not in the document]
edeiana: Got it, I didn't understand that the enum value had a meaning (larger == more recent trace version?).
I like @xdje42 suggestion. Now printing it when we print the trace_entry_t record (but not including it in `trace_version_names[]`).

clients/drcachesim/common/trace_entry.cpp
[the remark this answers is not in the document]
edeiana: Done.

edeiana: > The reverse might be an easier thing to do. See all the comments about sharing code: eliminate this array and use the existing public pretty-printer's marker printing code.

I modified this array to be coherent with view.cpp (which, now, it also uses).
I think we want a simple string representation of a marker type.
What view.cpp does is merging together the string representation of a marker type + a marker value.

Now, `trace_marker_names[]` contains the string representation of marker types, and it's used in the new `trace_marker_type_value_as_string()`, which returns a string representation of marker type + value (and it's in common between view.cpp and record_view.cpp).

clients/drcachesim/common/trace_entry.cpp
```diff
@@ -88,5 +88,55 @@ const char *const trace_type_names[] = {
...
+    "trace_marker_type_syscall_trace_end",
+    "trace_marker_type_branch_target",
+    "trace_marker_type_syscall_failed",
+    "trace_marker_type_direct_thread_switch",
+    "trace_marker_type_core_wait",
+    "trace_marker_type_core_idle",
+    "trace_marker_type_context_switch_start",
+    "trace_marker_type_context_switch_end",
+    "trace_marker_type_vector_length",
+};
```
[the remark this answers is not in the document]
edeiana: Done.

clients/drcachesim/common/trace_entry.cpp
```diff
@@ -88,5 +88,55 @@ const char *const trace_type_names[] = {
...
+    "trace_marker_type_cache_line_size",
+    "trace_marker_type_instruction_count",
+    "trace_marker_type_version",
+    "trace_marker_type_rseq_abort",
+    "trace_marker_type_window_id",
+    "trace_marker_type_physical_address",
+    "trace_marker_type_physical_address_not_available",
+    "trace_marker_type_virtual_address",
+    "trace_marker_type_page_size",
+    "trace_marker_type_syscall_idx",
```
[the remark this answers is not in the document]
edeiana: "trace_marker_type_" has been removed.
These strings changed to align them with view.cpp.
Considering this resolved.

clients/drcachesim/tools/record_view.cpp
edeiana: Yes, it would be larger (not my main concern though).
With "replicating utility functions like type_is_instr()" I mean that if we move to a switch we can't use type_is_instr() or type_has_address() anymore, and a case like:
```
...
} else if (type_is_instr(trace_type)) {
...
```
Becomes:
```
switch (...) {
...
case TRACE_TYPE_INSTR:
case TRACE_TYPE_DIRECT_JUMP:
...
case <all remaining TRACE_TYPE_ covered by type_is_instr()>:
...
}
```
Listing all the TRACE_TYPE_ cases in the switch for which type_is_instr() returns `true` is basically replicating what type_is_instr() does, which also means that if in the future we add a new TRACE_TYPE_ that is supposed to represent an instr, then we'd have (at least) two places to update.

I do agree that a switch would look more clear, but I think the cons outweight the pros in this case.
Unless I'm missing something?

clients/drcachesim/tools/record_view.cpp
edeiana: Now sharing (between view and record_view) the printing of all TRACE_MARKER_TYPE_ cases that do not require extra processing.

clients/drcachesim/tools/record_view.cpp
edeiana: For sharing of TRACE_TYPE_ pretty print between view.cpp and trace_view.cpp we need to come to a decision for what is the right way to print TRACE_TYPE_ enum values.
Right now there are two string representations: one hardcoded in view.cpp and one in `trace_type_names[]` (used by other tools).
https://github.com/DynamoRIO/dynamorio/pull/6771#discussion_r1571284574

clients/drcachesim/common/options.h
edeiana: Not adding it to the .dox was an oversight (but, I guess, a welcome one).
Since we want to limit the visibility of this tool, I agree on not adding it.

clients/drcachesim/tools/record_view.h
edeiana: Ok, I think the final decision here is to add the "warning header" (done) and renaming record_view to internal_record_view (done).

clients/drcachesim/common/options.h
derekbruening: As noted in the other thread, please rename to `internal_record_review`.

clients/drcachesim/tools/record_view.h
```diff
@@ -0,0 +1,96 @@
+/* **********************************************************
```
derekbruening, line 1: Maybe these files should be renamed to internal_record_review* as well so someone looking at the file structure doesn't see "record_view.cpp" and assume "that's what I should be using".

clients/drcachesim/common/options.h
edeiana: Yes, done.

clients/drcachesim/tools/record_view.h
edeiana: Done.

clients/drcachesim/CMakeLists.txt
edeiana: Yes, that's in the TODO (see PR description).
However, before working on that, I'd like to come to a consensus on the pretty printed output of internal_record_view (<--- new tool name), as that is what I will test against.

Here's what it prints now (a short "representative" snippet, note that I OMITTED a few things because I wasn't sure I could share them here... the actual output has those values):
```
The internal_record_view tool is a pretty printer for trace_entry_t records.
The trace_entry_t internal record format is not a public interface and is subject to change without notice.
This tool is meant for debugging only, please do not rely on the details here.
<header, trace_version: 6 == frequent_timestamps>
<marker: version 6 frequent_timestamps>
<marker: filetype 0xe40 x86_64>
<thread, tid: 259013>
<pid, pid: 259013>
<marker: cache line size OMITTED>
<marker: chunk instruction count 10000000>
<marker: page size OMITTED>
<marker: timestamp 13352585958813850>
<marker: cpu id 3>
<encoding, num_encoding_bytes: 8, encoding_bytes: OMITTED>
<instr, length: 8, pc: 0x7f6fdd3ec360>
<encoding, num_encoding_bytes: 8, encoding_bytes: OMITTED>
<direct_call, length: 8, pc: 0x7f6fdd3ec368>
<write, memref_size: 8, memref_addr: 0x7fffa4ef43c8>
<encoding, num_encoding_bytes: 8, encoding_bytes: OMITTED>
<instr, length: 8, pc: 0x7f6fdd3ec370>
<write, memref_size: 8, memref_addr: 0x7fffa4ef43c0>
...
```

derekbruening: I don't think it's worth spending any more time on this (esp for things like dividers and ordinals that the main view tool already has): this PR is just adding an internal tool that honestly will likely be rarely used, as evidenced by it not being needed in the past. Even now there likely is no use case where you wouldn't also go run the final view tool afterward, since that's the real view seen by tools.  E.g., for your own ISA transform and filtering, you can't just go by the output of this record view tool: you have to make sure it works at the memref_t layer, so you have to use the real view tool plus invariant checks anyway.  This internal tool is only a limited-use debugging step pretty much always followed up by using the "real" pretty printer.

derekbruening: Thinking more about when I've looked at the on-disk view with `od`: the most recent was debugging record_filter zip chunking support.  There I had to look just before and after the 10M-instr chunk boundaries.  Running `od` on one particular zipfile subfile at a time made this very easy.  Looking at the output above with no ordinals, and having to pass in the whole trace file/dir at once: how would I focus on chunk boundaries? I might end up going back to `od`!  I guess `-skip_instructions` is technically implemented for record_reader_t?  But it's slow: the fast chunk skips are not implemented.  Plus if the record_filter being debugged got the chunk boundary wrong I would have to go search for the chunk ord marker; if that's missing or wrong: would have to fall back to `od`.

Not sure there's an action item here: just an observation that this new tool is not enough to debug all on-disk issues as it abstracts away the file splits and I'm not sure it would have been useful the last time I did that.  Adding ordinals would help -- but I don't know it's worth the effort.  I would probably go run the regular view tool skipped to the chunk transition and take the closest timestamps and search for that in this tool's output or sthg.  But I would again want to be at the real file layer instead of the trace_entry_t stream layer's abstracting away of the file divisions.

clients/drcachesim/common/trace_entry.h
```diff
@@ -983,6 +987,120 @@ trace_arch_string(offline_file_type_t type)
...
+    case TRACE_MARKER_TYPE_SYSCALL_TRACE_START:
+    case TRACE_MARKER_TYPE_SYSCALL_TRACE_END:
+        ss << "<" << marker_name << " number " << marker_value << ">\n";
+        break;
+    case TRACE_MARKER_TYPE_CONTEXT_SWITCH_START:
+    case TRACE_MARKER_TYPE_CONTEXT_SWITCH_END:
+        ss << "<" << marker_name << " type " << marker_value << ">\n";
+        break;
+    /* We don't have a way to know the trace version here.  This might be an offset,
+     * but we don't make any distinction.
```
derekbruening, line 1041: Please elaborate as a new reader will have no idea what you mean: I only know b/c I remember a legacy version having offset instead of abs addr.

clients/drcachesim/launcher.cpp
```diff
@@ -323,7 +323,8 @@ _tmain(int argc, const TCHAR *targv[])
             FATAL_ERROR("invalid -outdir %s", op_outdir.get_value().c_str());
         }
     } else {
-        if (op_simulator_type.get_value() == RECORD_FILTER) {
+        if (op_simulator_type.get_value() == RECORD_FILTER ||
+            op_simulator_type.get_value() == INTERNAL_RECORD_VIEW) {
```
derekbruening, line 327: Do we need the multi-simulator support (colon-separated for memref_t tools) now that we have more than one tool?  Add a XXX comment acknowledging we're aware multi-tool is not currently supported for record tools.

clients/drcachesim/tools/internal_record_view.cpp
```diff
@@ -0,0 +1,217 @@
...
+        std::cerr << "<" << trace_type_name << ", tid: " << tid << ">\n";
+    } else if (trace_type == TRACE_TYPE_PID) {
+        uint pid = (uint)entry.addr;
+        std::cerr << "<" << trace_type_name << ", pid: " << pid << ">\n";
+    } else if (trace_type == TRACE_TYPE_ENCODING) {
+        unsigned short num_encoding_bytes = entry.size;
+        std::cerr << "<" << trace_type_name
+                  << ", num_encoding_bytes: " << num_encoding_bytes
+                  << ", encoding_bytes: 0x" << std::hex;
+        /* Print encoding byte by byte (little-endian).
```
derekbruening, line 161: Byte-by-byte is only desirable for x86; we'd want words for AArch64, words or half-words (presumably we have LSB 1 for Thumb) for AArch32, and words for RISCV.

clients/drcachesim/tools/view.cpp
```diff
@@ -314,142 +322,35 @@ view_t::parallel_shard_memref(void *shard_data, const memref_t &memref)
...
-            break;
-        case TRACE_MARKER_TYPE_CONTEXT_SWITCH_START:
-            std::cerr << "<marker: trace start for context switch type "
-                      << memref.marker.marker_value << ">\n";
-            break;
-        case TRACE_MARKER_TYPE_CONTEXT_SWITCH_END:
-            std::cerr << "<marker: trace end for context switch type "
+        case TRACE_MARKER_TYPE_SPLIT_VALUE:
+        case TRACE_MARKER_TYPE_SYSCALL_IDX:
+            std::cerr << "<marker: type " << memref.marker.marker_type << "; value "
```
derekbruening, line 344: Seems like these 2 cases can be omitted as this is the default behavior in the marker printing helper called below.

clients/drcachesim/tools/view.cpp
```diff
@@ -314,142 +322,35 @@ view_t::parallel_shard_memref(void *shard_data, const memref_t &memref)
...
-                      << memref.marker.marker_value << ">\n";
+            /* The previous cases have state and require extra processing.  Here we print
+             * all the remaining cases that do not need that extra processing and
+             * bookkeeping.
+             */
+            std::cerr << trace_marker_type_value_as_string(memref.marker.marker_type,
+                                                           memref.marker.marker_value);
             break;
         }
         return true;
```
derekbruening, line 356: Please add a unit test.  (This may be a dup of another comment thread.)

clients/drcachesim/tools/view.cpp
edeiana: We actually catch those two cases in `trace_marker_type_value_as_string()` and print a more informative output that looks different than the one here: <marker: type NUM; value NUM>.
So, to preserve the previous view.cpp behavior, we catch these 2 cases early and print them the same way view.cpp used to print them.

We can remove these two cases if:
1) we decide we don't care about printing a more informative output (hence we remove them from here and `trace_marker_type_value_as_string()`, letting the default in `trace_marker_type_value_as_string()` take care of it. (<--- not a fan of this, I think we should print the most "informative" version of marker type and value)
2) we decide to update view.cpp's printing for those two cases and align it to `trace_marker_type_value_as_string()`. (<--- might break doc and examples?)

derekbruening: SPLIT_VALUE will never be seen by view.cpp as it is never presented as a memref_t, so there is no reason to have an explicit case entry for it.

Anything not explicitly in a case statement in view.cpp can be considered to not matter: if someone wanted to see that marker we would add a case statement.  The `default` there is a fallback and we do not need to preserve it: we don't want to preserve it: we want to print the string name, not the marker enum number.

edeiana: Got it.
Then, since those two cases were both catched in what used to be view.cpp's default case, we can remove them and let `trace_marker_type_value_as_string()` print them (even though view.cpp will never see SPLIT_VALUE).
Going with 2).

Issue #6760, opened 2024-04-05
AArch64: Fix P register save/restore on 128-bit vector length systems


AssadHashmi:
Currently predicate and FFR state is stored in `dr_simd_t`.
This  wastes space as predicates and FFR are an 1/8th of the size of the implementation's scalable vector register length.

Issue raised during review of https://github.com/DynamoRIO/dynamorio/pull/6757.

SVE master issue https://github.com/DynamoRIO/dynamorio/issues/5365.

jackgallagher-arm:
It turns out that #6757 broke slot save/restore on systems with 128-bit vector length because `sizeof(dr_simd_t)*8` is out of range for `Pn` register `ldr`/`str` instructions (see #6774 for details) so this is a bug fix in addition to saving space.

Issue #6764, opened 2024-04-08
`client.detach_test` is timing out on the `vs2019-32` CI build


jackgallagher-arm:
The `windows-ci|vs2019-32` job for #6756 failed repeatedly with `client.detach_test` timing out.
```
201/262 Test #131: code_api|client.detach_test ..................................***Timeout  90.02 sec
```            

I also spotted it timing out in the same way for #6762, but in that case it passed on a re-run.
https://github.com/DynamoRIO/dynamorio/actions/runs/8571639502/attempts/1

PR #6774, opened 2024-04-12
i#6760 AArch64: Use smaller data types for SVE P and FFR registers

jackgallagher-arm:
PR #6757 fixed the way we read/write SVE register slots but unfortunately it is now broken on systems with 128-bit vector length.

Both SVE vector and predicate registers use dr_simd_t slots which is a 64-byte type meant to store up to 512-bit vector registers. SVE predicate registers are always 1/8 the size of the vector register so for 512-bit vector length systems we only really need 64 / 8 = 8 bytes to store predicate registers.

The ldr/str instructions we use to read and write the predicate register slots have a base+offset memory operand where the offset is a value in the range [-256, 255] scaled based by predicate register length. We read and write the registers by setting the base address to the address of the first slot, and setting the offset to n * sizeof(dr_simd_t) for each register Pn.
For systems with 128-bit vector length, this means the predicate registers are 16 / 8 = 2 bytes so the maximum offset we can reach is 2 * 255 = 510 bytes. This means on 128-bit VL systems we can only reach the first 8 predicate registers (8 * sizeof(dr_simd_t) = 512).

By changing the predicate register and FFR slots to use a new type dr_svep_t which is 1/8 the size of dr_simd_t we can fix this bug and save space.

dr_svep_t is currently 8 bytes to correspond to 64 byte vectors, but even if we extend DynamoRIO to support the maximum SVE vector length of 2048-bits (256 bytes) dr_svep_t will only need to be increased to 256 / 8 = 32 bytes so the maximum offset (15 * 32 = 480 bytes) will always be in range.

As this changes the size of the predicate register and FFR slots, this changes the size of the dr_mcontext_t structure and breaks backwards compatibility with earlier versions of DynamoRIO so the version number is increased to 10.90.

Issues: #6760, #5365
Fixes: #6760


jackgallagher-arm:
`vs2019-32` failure looks like #6764

core/arch/arch.c
```diff
@@ -3825,21 +3825,48 @@ dump_mcontext(priv_mcontext_t *context, file_t f, bool dump_xml)
     }
 #elif defined(AARCHXX)
     {
-        int i, j;
 #    ifdef AARCH64
-        int words = proc_has_feature(FEATURE_SVE) ? 16 : 4;
+        const uint elements = proc_has_feature(FEATURE_SVE)
```
derekbruening, line 3829: s/elements/uint_count/ or s/elements/u32_count/

core/arch/arch.c
```diff
@@ -3825,21 +3825,48 @@ dump_mcontext(priv_mcontext_t *context, file_t f, bool dump_xml)
...
+            for (uint j = 0; j < elements; j++) {
                 print_file(f, "%08x ", context->simd[i].u32[j]);
             }
             print_file(f, dump_xml ? "\"\n" : "\n");
         }
-        /* TODO i#5365: SVE predicate registers and FFR dump. */
+#    ifdef AARCH64
+        if (proc_has_feature(FEATURE_SVE)) {
+            /* Dump SVE P registers */
+            const uint pred_u16_elements =
```
derekbruening, line 3853: I think s/_elements/_count/ is clearer (else sounds like an array of of values or sthg to me)

core/arch/arm/arm.asm
```diff
@@ -60,17 +60,21 @@ DECL_EXTERN(initstack_mutex)
...
 # define MCXT_NUM_SIMD_SLOTS 16
 # define SIMD_REG_SIZE       16
+# define MCXT_NUM_PRED_SLOTS  0
+# define PRED_REG_SIZE        0
 # define NUM_GPR_SLOTS       17 /* incl flags */
 # define GPR_REG_SIZE         4
 #endif
 #define PRE_SIMD_PADDING     0
-#define PRIV_MCXT_SIMD_SIZE (PRE_SIMD_PADDING + MCXT_NUM_SIMD_SLOTS*SIMD_REG_SIZE)
+#define PRIV_MCXT_SIMD_SIZE (PRE_SIMD_PADDING + MCXT_NUM_SIMD_SLOTS*SIMD_REG_SIZE + MCXT_NUM_PRED_SLOTS*PRED_REG_SIZE)
```
derekbruening, line 77: style: Line too long

core/unix/signal_linux_aarch64.c
```diff
@@ -257,12 +257,12 @@ sigcontext_to_mcontext_simd(priv_mcontext_t *mc, sig_full_cxt_t *sc_full)
                         memcpy(&mc->simd[i].q, &fpc->vregs[i], sizeof(mc->simd->q));
                     }
                     for (i = 0; i < MCXT_NUM_SVEP_SLOTS; i++) {
-                        memcpy(&mc->svep[i].u32,
+                        memcpy(&mc->svep[i].u16,
                                (byte *)sve + SVE_SIG_PREG_OFFSET(quads_per_vector, i),
-                               sve->vl);
+                               sve->vl / 8);
```
derekbruening, line 262: So this used to copy 8x more data than it needed?

derekbruening (approved):
The most important thing IMHO is having a test that reproduces the 1st-8-pred-reg bug and shows that this PR fixes it (and will be a regression test for future changes).

suite/tests/client-interface/cleancall-opt-shared.h
```diff
@@ -353,30 +353,55 @@ dump_diff_mcontexts(void)
...
+            const size_t num_elements = veclen_bytes / sizeof(before_reg.u64[0]);
+            dr_fprintf(STDERR, "%s before: ", reg_name);
+            for (size_t element = 0; element < num_elements; element++) {
+                dr_fprintf(STDERR, PFMT, before_reg.u64[element]);
+            }
+            dr_fprintf(STDERR, " after: ");
+            for (size_t element = 0; element < num_elements; element++) {
+                dr_fprintf(STDERR, PFMT, after_reg.u64[element]);
+            }
+        }
```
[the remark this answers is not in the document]
jackgallagher-arm: > I assume the SVE GA CI machine is 128-bit and reproduces the bug?
The SVE CI machine has a 256-bit vector length so it doesn't reproduce the bug (otherwise PR #6757 would not have passed). We have an internal CI jobs which run the DynamoRIO test suite on a 128-bit vector length system and that caught the problem.

I don't think we need to add any new tests for this because existing tests caught the problem, but it might be a good idea to extend the CI tests to run with 128-bit vector length to improve coverage. I will discuss it with @AssadHashmi

core/arch/aarch64/emit_utils.c
```diff
@@ -791,41 +786,39 @@ append_save_simd_reg(dcontext_t *dcontext, instrlist_t *ilist, bool absolute)
...
+                                 opnd_create_reg(DR_REG_P0 + i)));
         }
         /* There is no store instruction for the first-fault register (FFR). Use
          * a temporary predicate register to store:
          * rdffr p15.b
          * add x2, x(dcxt), #(offset ffr)
-         * str p15, [x2, #(ffr)]
-         * ldr p15, [x1, #(15 mul vl)]
+         * str p15, [x2, #0]
+         * ldr p15, [x1, #(15 * sizeof(dr_svep_t))]
```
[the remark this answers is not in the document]
jackgallagher-arm: Done.

core/arch/arch.c
jackgallagher-arm: Done.

core/arch/arch.c
jackgallagher-arm: Done.

core/unix/signal_linux_aarch64.c
jackgallagher-arm: Yes it did.

core/unix/signal_linux_aarch64.c
```diff
@@ -257,12 +257,12 @@ sigcontext_to_mcontext_simd(priv_mcontext_t *mc, sig_full_cxt_t *sc_full)
...
                     }
                     memcpy(&mc->ffr, (byte *)sve + SVE_SIG_FFR_OFFSET(quads_per_vector),
-                           sve->vl);
+                           sve->vl / 8);
```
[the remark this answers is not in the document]
jackgallagher-arm: The kernel provides some macros to calculate the register sizes used in the signal structure. I'll switch to using those.

PR #6776, opened 2024-04-16
i#5036 A64 scatter/gather, part 11: First-fault loads

jackgallagher-arm:
Adds support for first-fault loads to drx_expand_scatter_gather(). First-fault loads (ldff1*) work similarly to
scalar+vector/vector+immediate gather instructions or scalar+scalar predicated contiguous ld1* loads, but with different behaviour if an element access faults. This commit implements this behaviour and extends the scatter/gather tests to include ldff1* instructions.

Issue: #5036


jackgallagher-arm:
The `vs2019-32` failure looks like #6764

AssadHashmi (approved):

PR #6780, opened 2024-04-18
i#6764: Mark detach_test flaky on win32

derekbruening:
The client.detach_test has been timing out on win32 quite frequently. We put it on the ignore list to get the test suite green.

Issue: #6764


derekbruening:
x32 failure is opcode_mix size (maybe that needs to be put on the ignore list too; someone did try to figure it out recently but failed): #6303

jackgallagher-arm (approved):

PR #6783, opened 2024-04-18
i#6662 regdeps ISA: virtual registers

edeiana:
Containing-register IDs can be >=256, hence their value does not fit
in the allotted 8 bits per register operand of regdeps encoding.
This was causing a memory corruption in instr_convert_to_isa_regdeps()
where src_reg_used and dst_reg_used have only 256 elements and are
laid out next to each other in memory. Writing to index >=256 into one was
overwriting the other. Fix: remap containing-register IDs to virtual-register
IDs starting from 0 for all architectures. We still have only up to 198 unique
containing registers (max number of containing registers for AARCH64),
so remapping allows to fit them in 8 bits.

In the re-mapping (from DR_REG_ to DR_REG_V) we exclude DR_REG_INVALID
and DR_REG_NULL to avoid issues with opnd_t operations for registers.

We introduce a private routine dr_reg_to_virtual() to do the mapping from real ISA
to virtual register. We use it in instr_convert_to_isa_regdeps() to avoid the issue
mentioned above.

We modified the get_register_name() public API to use the global dcontext and its
ISA mode to determine whether to return a real register name or a virtual one.
The signature of the API remained the same, but we document the use of the global
dcontext in doxygen.

We also re-introduce setting the size for register operands in
instr_convert_to_isa_reg_deps() and decode_isa_regdeps() as
instr_t.operation_size because not all DR_REG_V have a predefined size
based on their enum value (e.g., reserved DR_REG_XMM enum values).

We added tests to check that DR_REG_ with IDs >=256 don't cause problems.

Issue: #6662


suite/tests/api/ir_regdeps.c
```diff
@@ -328,6 +440,22 @@ test_instr_create_encode_decode_synthetic_riscv64(void *dc)
 }
 #endif
 
+void
+test_virtual_register(void)
+{
+    const char rv0[] = "rv0";
+    ASSERT(strncmp(get_virtual_register_name((reg_id_t)0), rv0, sizeof(rv0)) == 0);
```
derekbruening, line 447: Shouldn't these use the enum values?

suite/tests/api/ir_regdeps.c
```diff
@@ -121,6 +126,100 @@ test_instr_encode_decode_synthetic(void *dc, instr_t *instr)
...
+    // 62 22 7e 4f 72 b4 f5 00 00 00 10  vcvtneps2bf16 0x10000000\(%rbp,%r14,8\),%ymm30\{%k7\}
+    // 62 42 7e 58 72 31     vcvtneps2bf16 \(%r9\)\{1to16\},%ymm30
+    // 62 62 7e 48 72 71 7f  vcvtneps2bf16 0x1fc0\(%rcx\),%ymm30
+    // 62 62 7e df 72 b2 00 e0 ff ff   vcvtneps2bf16 -0x2000\(%rdx\)\{1to16\},%ymm30\{%k7\}\{z\}
+    // 62 02 16 40 52 f4     vdpbf16ps %zmm28,%zmm29,%zmm30
+    // 62 22 16 47 52 b4 f5 00 00 00 10  vdpbf16ps 0x10000000\(%rbp,%r14,8\),%zmm29,%zmm30\{%k7\}
+    // 62 42 16 50 52 31     vdpbf16ps \(%r9\)\{1to16\},%zmm29,%zmm30
+    // 62 62 16 40 52 71 7f  vdpbf16ps 0x1fc0\(%rcx\),%zmm29,%zmm30
+    // 62 62 16 d7 52 b2 00 e0 ff ff   vdpbf16ps -0x2000\(%rdx\)\{1to16\},%zmm29,%zmm30\{%k7\}\{z\}
+    byte out00[] = { 0x62, 0x02, 0x17, 0x40, 0x72, 0xf4,}; //
```
derekbruening, line 151: These are inputs, not outputs, so the name "out" is confusing.

derekbruening, line 151: Why a trailing //?  Best to remove.

derekbruening, line 151: Why do we need encodings at all?  Let's just synthesize some instructions at the IR level.  (It wouldn't really matter if they were illegal but may as well base them on valid instrs.)

suite/tests/api/ir_regdeps.c
```diff
@@ -121,6 +126,100 @@ test_instr_encode_decode_synthetic(void *dc, instr_t *instr)
...
+    byte out06[] = { 0x62, 0x22, 0x7e, 0x4f, 0x72, 0xb4, 0xf5, 0x00, 0x00, 0x00, 0x10,}; //
+    byte out07[] = { 0x62, 0x42, 0x7e, 0x58, 0x72, 0x31,}; //
+    byte out08[] = { 0x62, 0x62, 0x7e, 0x48, 0x72, 0x71, 0x7f,}; //
+    byte out09[] = { 0x62, 0x62, 0x7e, 0xdf, 0x72, 0xb2, 0x00, 0xe0, 0xff, 0xff,}; //
+    byte out10[] = { 0x62, 0x02, 0x16, 0x40, 0x52, 0xf4,}; //
+    byte out11[] = { 0x62, 0x22, 0x16, 0x47, 0x52, 0xb4, 0xf5, 0x00, 0x00, 0x00, 0x10,}; //
+    byte out12[] = { 0x62, 0x42, 0x16, 0x50, 0x52, 0x31,}; //
+    byte out13[] = { 0x62, 0x62, 0x16, 0x40, 0x52, 0x71, 0x7f,}; //
+    byte out14[] = { 0x62, 0x62, 0x16, 0xd7, 0x52, 0xb2, 0x00, 0xe0, 0xff, 0xff,}; //
+    opnd_t test0[][4] = {
```
derekbruening, line 166: Not following why there are 13 cases above and then a set of 4 here and 3 below.  See comment above: I would just throw all these encodings out and synthesize some IR with K and ZMM.

core/ir/aarch64/encode.c
```diff
@@ -198,6 +199,80 @@ const reg_id_t dr_reg_fixer[] = { REG_NULL,
 };
 /* clang-format on */
 
+/* Maps containing-registers to their DR_ISA_REGDEPS virtual-register.
+ * Same size as dr_reg_fixer[], keep them synched.
+ * Note: DR_REG_INVALID (== DR_REG_V1 == 1) has been excluded to avoid failures on opnd_t
```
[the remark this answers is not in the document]
edeiana: Unfortunately no, although for most arches DR_REG_INVALID is 1, for x86 DR_REG_INVALID is 187, so starting at 188 wouldn't leave us enough space.

core/ir/aarch64/encode.c
```diff
@@ -198,6 +199,80 @@ const reg_id_t dr_reg_fixer[] = { REG_NULL,
...
+ * operations.
+ */
+const reg_id_t dr_reg_virtual[] = {
+    DR_REG_V0,   DR_REG_V0,   DR_REG_V2,   DR_REG_V3,   DR_REG_V4,   DR_REG_V5,
```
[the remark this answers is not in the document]
edeiana: dr_reg_virtual[] mirrors dr_reg_fixer[] but with virtual registers, it's not a sequence, there are duplicates (like for dr_reg_fixer[]).
Note that in this specific case DR_REG_V1 would not appear anywhere because 1 is the DR_REG_INVALID for AARCH64.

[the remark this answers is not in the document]
derekbruening: I find this very confusing to have V0 be NULL: I don't think that has any use case?  See my suggestion below to have V0 start at 2 and to skip INVALID for all arches to end up with an understandable cross-arch enum set.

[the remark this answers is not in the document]
edeiana: One could say that DR_REG_V0 is the virtual DR_REG_NULL.
For regdeps ISA we don't really need the concept of "absence of a register", which, I think, is what DR_REG_NULL represents.
It's also weird IMHO to have a map from real registers to virtual registers that maps back to a "real" DR_REG_ register rather than a DR_REG_V register.
Note that DR_REG_V0 here would only appear where dr_reg_fixer[] has a DR_REG_NULL, it wouldn't be used for its other "valid" registers.
That being said, we could have a properly named "DR_REG_VNULL" value, rather than a generic "DR_REG_V0", but do we need a "null register" in the realm of regdeps ISA?

core/ir/aarch64/encode.c
```diff
@@ -198,6 +199,80 @@ const reg_id_t dr_reg_fixer[] = { REG_NULL,
...
+    DR_REG_V6,   DR_REG_V7,   DR_REG_V8,   DR_REG_V9,   DR_REG_V10,  DR_REG_V11,
+    DR_REG_V12,  DR_REG_V13,  DR_REG_V14,  DR_REG_V15,  DR_REG_V16,  DR_REG_V17,
+    DR_REG_V18,  DR_REG_V19,  DR_REG_V20,  DR_REG_V21,  DR_REG_V22,  DR_REG_V23,
+    DR_REG_V24,  DR_REG_V25,  DR_REG_V26,  DR_REG_V27,  DR_REG_V28,  DR_REG_V29,
+    DR_REG_V30,  DR_REG_V31,  DR_REG_V32,  DR_REG_V33,  DR_REG_V34,  DR_REG_V2,
```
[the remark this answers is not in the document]
edeiana: I'm not sure I follow.
This is not a sequence, it's a mapping, even if we have DR_REG_V0 starting after DR_REG_NULL and DR_REG_INVALID, we'd have a jump there.
https://github.com/DynamoRIO/dynamorio/pull/6783#discussion_r1571489157

core/ir/instr_shared.c
```diff
@@ -3018,9 +3018,10 @@ instr_convert_to_isa_regdeps(void *drcontext, instr_t *instr_real_isa,
                 /* Map sub-registers to their containing register.
                  */
                 reg_id_t reg_canonical = reg_to_pointer_sized(reg);
-                if (!src_reg_used[reg_canonical]) {
+                reg_id_t reg_virtual = reg_to_virtual(reg_canonical);
```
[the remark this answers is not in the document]
edeiana: True the dr_reg_virtual[] mapping returns the "canonical virtual register", so it's doing both: 1) getting the canonical (i.e., containing register) and 2) mapping it to a virtual register.
I was initially debating if dr_reg_virtual[] should have had an "invalid virtual register" for all indexes that are not containing registers, but I guess I'll just document these two actions and save the extra call to reg_to_pointer_sized().

core/ir/x86/encode.c
```diff
@@ -236,6 +237,70 @@ const reg_id_t dr_reg_fixer[] = {
     DR_REG_BND2,    DR_REG_BND3,
 };
 
+/* Maps containing-registers to their DR_ISA_REGDEPS virtual-register.
+ * Same size as dr_reg_fixer[], keep them synched.
+ * Note: DR_REG_INVALID (== DR_REG_V187 == 187) has been excluded to avoid failures on
```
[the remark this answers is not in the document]
edeiana: We can, but what's the advantage?
If you're thinking about the jumps you mentioned here: https://github.com/DynamoRIO/dynamorio/pull/6783#discussion_r1571407194 this would not solve the "problem", the jumps happen because of the mapping to containing registers (same as they happen in dr_reg_fixer[]).

core/ir/aarch64/encode.c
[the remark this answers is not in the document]
edeiana: The reason to exclude DR_REG_INVALID (which would have DR_REG_V1 as its corresponding virtual register) is to avoid having a real ISA register map to DR_REG_V1.
This would raise an "Invalid register" error in, for example, instr_same() when we compare the operands of 2 regdeps instructions.

core/ir/x86/encode.c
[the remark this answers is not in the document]
edeiana: Mhm, I guess the advantage is that all DR_REG_V would be valid (we'd still have a "jumps" in dr_reg_virtual[] though, that really depends on the mapping to containing registers).

core/ir/aarch64/encode.c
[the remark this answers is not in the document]
edeiana: > I find this very confusing to have V0 be NULL: I don't think that has any use case? See my suggestion below to have V0 start at 2 and to skip INVALID for all arches to end up with an understandable cross-arch enum set.

Ok, will change to that, so all DR_REG_V are valid.

core/ir/aarch64/encode.c
```diff
@@ -198,6 +199,80 @@ const reg_id_t dr_reg_fixer[] = { REG_NULL,
 };
 /* clang-format on */
 
+/* Maps containing-registers to their DR_ISA_REGDEPS virtual-register.
+ * Same size as dr_reg_fixer[], keep them synched.
+ * Note: DR_REG_INVALID (== DR_REG_V1 == 1) has been excluded to avoid failures on opnd_t
+ * operations.
+ */
+const reg_id_t dr_reg_virtual[] = {
```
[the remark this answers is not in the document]
edeiana: Done.

core/ir/isa_regdeps/encoding_common.c
```diff
@@ -0,0 +1,67 @@
...
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+#include "encoding_common.h"
+
+/* Keep reg_virtual_names[] and DR_REG_V* enum in encoding_common.h synched.
+ */
+const char *const reg_virtual_names[] = {
```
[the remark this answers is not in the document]
edeiana: Done.

core/ir/isa_regdeps/encoding_common.h
```diff
@@ -127,4 +127,267 @@
  */
 #define REGDEPS_MAX_NUM_REGS 256
 
+/* We need a seprate enum from the DR_REG_ in opnd_api.h so we can start counting virtual
+ * registers from 0.  Otherwise, the DR_REG_ enum values in opnd_api.h won't fit in the
+ * 1 byte operand size of DR_ISA_REGDEPS encoding.
+ */
+enum {
```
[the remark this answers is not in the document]
edeiana: Yes, let's make them public. Done.

core/ir/opnd_api.h
```diff
@@ -3008,6 +3008,14 @@ DR_API
 const char *
 get_register_name(reg_id_t reg);
 
+DR_API
+/**
+ * Assumes that \p reg is a DR_REG_V 8-bit register constant.
+ * Returns the string name for virtual \p reg.
+ */
+const char *
+get_virtual_register_name(reg_id_t reg);
```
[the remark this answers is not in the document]
edeiana: Done.

core/ir/x86/encode.c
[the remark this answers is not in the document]
edeiana: Done.

suite/tests/api/ir_regdeps.c
edeiana: Now that DR_REG_V are public, yes. Done.

core/ir/instr_shared.c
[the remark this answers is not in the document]
edeiana: Removed. Done.

core/ir/instr_shared.c
```diff
@@ -3101,6 +3105,7 @@ instr_convert_to_isa_regdeps(void *drcontext, instr_t *instr_real_isa,
         for (uint reg = 0; reg < REGDEPS_MAX_NUM_REGS; ++reg) {
             if (dst_reg_used[reg]) {
                 opnd_t dst_opnd = opnd_create_reg((reg_id_t)reg);
+                opnd_set_size(&dst_opnd, max_opnd_size);
```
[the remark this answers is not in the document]
edeiana: This is more to make instr_same() (which checks the register size) work.
Now that we are mapping to virtual registers, we might end up having a DR_REG_V whose value is, for example, one of the DR_REG_XMM reserved ones, so calling reg_get_size() on it would raise the ASSERT: "reg_get_size: invalid register", since it wouldn't match any of the if-else cases.
However, it never really made sense to query the size of a virtual register, even before.
Anyway, added comment.
Done.

suite/tests/api/ir_regdeps.c
edeiana: Initially I couldn't find an INSTR_CREATE to reproduce the problem, but actually one of those there does, so I removed that whole function with a single INSTR_CREATE that still triggers that bug.
Done.

suite/tests/api/ir_regdeps.c
edeiana: Yes, fixed. https://github.com/DynamoRIO/dynamorio/pull/6783#discussion_r1571820861

suite/tests/api/ir_regdeps.c
edeiana: Not there anymore, fixed. https://github.com/DynamoRIO/dynamorio/pull/6783#discussion_r1571820861

edeiana: Not there anymore, fixed. https://github.com/DynamoRIO/dynamorio/pull/6783#discussion_r1571820861

core/ir/aarch64/encode.c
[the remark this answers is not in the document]
edeiana: Done.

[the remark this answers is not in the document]
edeiana: No more relevant.

core/ir/aarch64/encode.c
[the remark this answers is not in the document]
derekbruening: I think fundamentally these arrays are hard to read by a reviewer or a future code maintainer, as evidenced by all my review comments and perhaps confusion.  The existing dr_reg_fixer are much easier to read as they have clear patterns: NULL, NULL, 0-32, etc.  I think these arrays need to have comments added identifying the original DR_REG_ value being mapped to each slot (or for each set of similar slots like GPRs etc.) as otherwise these will be too hard to read and maintain.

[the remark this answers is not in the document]
derekbruening: Losing track of which were resolved: going to unresolve this one.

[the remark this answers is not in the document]
edeiana: Ok, I added the original DR_REG_ being mapped as a comment, and followed the same "define pattern" of dr_reg_fixer[] in aarch64.
I can also add "define patterns" to the other arches (note that dr_reg_fixer[] does not though, so is gonna be a bit "asymmetric") if they are still unreadable.

core/ir/aarch64/encode.c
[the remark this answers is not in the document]
derekbruening: I think any DR_REG_ enum should leave the sentinel DR_REG_NULL and DR_REG_INVALID intact and not map something else to those values, as too much code acts on those (just as you already found for INVALID).

api/docs/release.dox
```diff
@@ -226,7 +226,10 @@ Further non-compatibility-affecting changes include:
    purpose of preserving register dependencies.
  - Added instr_convert_to_isa_regdeps() API that converts an #instr_t from a real ISA
    (e.g., #DR_ISA_AMD64) to the #DR_ISA_REGDEPS synthetic ISA.
-
+ - Added dr_reg_to_virtual() which takes a regular DR_REG_ register and returns a DR_REG_V
+   virtual register for #DR_ISA_REGDEPS.
+ - Added dr_get_virtual_register_name() which takes a virtual DR_REG_V register (used by
```
derekbruening, line 231: Long-term if we get multi-arch-in-same-build support we would need the same register printing routine (I guess today's get_register_name()) to work for any ISA.  Should we plan ahead for that and not have an ISA-specific routine like the one added here but instead some mechanism to get which ISA is being asked about.  Probably we won't create _ex versions of every single reg_ routine so presumably this will use the dcontext ISA mode.  So we wouldn't be adding a new API routine.

core/ir/aarch64/encode.c
```diff
@@ -196,6 +196,177 @@ const reg_id_t dr_reg_fixer[] = { REG_NULL,
...
+    DR_REG_V47, DR_REG_V48, DR_REG_V49, DR_REG_V50, DR_REG_V51, DR_REG_V52,             \
+    DR_REG_V53, DR_REG_V54, DR_REG_V55, DR_REG_V56, DR_REG_V57, DR_REG_V58,             \
+    DR_REG_V59, DR_REG_V60, DR_REG_V61, DR_REG_V62, DR_REG_V63, DR_REG_V64,
+
+    VIRTUAL_ZREGS /* from DR_REG_Z0 to DR_REG_Z31 */
+    VIRTUAL_ZREGS /* from DR_REG_Q0 to DR_REG_Q31 */
+    VIRTUAL_ZREGS /* from DR_REG_D0 to DR_REG_D31 */
+    VIRTUAL_ZREGS /* from DR_REG_S0 to DR_REG_S31 */
+    VIRTUAL_ZREGS /* from DR_REG_H0 to DR_REG_H31 */
+    VIRTUAL_ZREGS /* from DR_REG_B0 to DR_REG_B31 */
```
derekbruening, line 232: (Much much easier to read and understand.)

core/ir/arm/encode.c
```diff
@@ -461,6 +461,238 @@ const reg_id_t dr_reg_fixer[] = {
...
+    DR_REG_V22, /* DR_REG_S24 */
+    DR_REG_V22, /* DR_REG_S25 */
+    DR_REG_V22, /* DR_REG_S26 */
+    DR_REG_V22, /* DR_REG_S27 */
+    DR_REG_V23, /* DR_REG_S28 */
+    DR_REG_V23, /* DR_REG_S29 */
+    DR_REG_V23, /* DR_REG_S30 */
+    DR_REG_V23, /* DR_REG_S31 */
+
+    /* AArch64-only.
```
derekbruening, line 599: Wait why would aarch64-only things be in the arm/ directory?

core/ir/instr_shared.c
```diff
@@ -3101,6 +3102,16 @@ instr_convert_to_isa_regdeps(void *drcontext, instr_t *instr_real_isa,
         for (uint reg = 0; reg < REGDEPS_MAX_NUM_REGS; ++reg) {
             if (dst_reg_used[reg]) {
                 opnd_t dst_opnd = opnd_create_reg((reg_id_t)reg);
+                /* Virtual registers may have ID values not supported by reg_get_size().
+                 * We set the opnd_t.size field to be the operation_size max_opnd_size so
+                 * that reg_get_size() can return it without triggering a CLIENT_ASSERT
```
derekbruening, line 3107: Reading these 1st 2 sentences I am thinking "just go fix reg_get_size: why try to work around it here?  why are you treating reg_get_size like some read-only unchangeable thing?".  The key is the 3rd sentence that the sizes are not fixed: put that up front in the 1st sentence to make it clear we can't easily have reg_get_size support virtual registers.

core/ir/isa_regdeps/decode.c
```diff
@@ -98,6 +98,16 @@ decode_isa_regdeps(dcontext_t *dcontext, byte *encoded_instr, instr_t *instr)
     for (uint i = 0; i < num_dsts; ++i) {
         reg_id_t dst = (reg_id_t)encoded_instr[i + REGDEPS_OPND_INDEX];
         opnd_t dst_opnd = opnd_create_reg((reg_id_t)dst);
+        /* Virtual registers may have ID values not supported by reg_get_size().
```
derekbruening, line 101: Ditto.

core/ir/opnd_api.h
```diff
@@ -1307,6 +1307,268 @@ enum {
 #endif
 };
 
+/* We need a seprate DR_REG_V enum from DR_REG_ so we can start counting virtual registers
+ * from lower values.  Otherwise, the DR_REG_ enum values won't fit in the 1 byte operand
+ * size of #DR_ISA_REGDEPS encoding.  Note that DR_REG_V skip the value of DR_REG_INVALID
```
derekbruening, line 1312: s/skip the value of/skips both values of/

core/ir/opnd_api.h
```diff
@@ -1307,6 +1307,268 @@ enum {
...
+    DR_REG_V176,
+    DR_REG_V177,
+    DR_REG_V178,
+    DR_REG_V179,
+    DR_REG_V180,
+    DR_REG_V181,
+    DR_REG_V182,
+    DR_REG_V183,
+    DR_REG_V184,
+    DR_REG_V185 = 188, /* Skip x86 DR_REG_INVALID == 187. */
```
derekbruening, line 1502: Let's add an assert at the top of arch_init() that we got these values right: I guess a loop over the enum values ensuring none equals _NULL or _INVALID, probably under DOCHECK level 1 or 2.

derekbruening, line 1502: Clearer to be `DR_REG_V185 = DR_REG_INVALID + 1`?

core/ir/x86/encode.c
```diff
@@ -236,6 +236,349 @@ const reg_id_t dr_reg_fixer[] = {
...
+    DR_REG_V6,      /* DR_REG_RSI */
+    DR_REG_V7,      /* DR_REG_RDI */
+    DR_REG_V8,      /* DR_REG_R8 */
+    DR_REG_V9,      /* DR_REG_R9 */
+    DR_REG_V10,     /* DR_REG_R10 */
+    DR_REG_V11,     /* DR_REG_R11 */
+    DR_REG_V12,     /* DR_REG_R12 */
+    DR_REG_V13,     /* DR_REG_R13 */
+    DR_REG_V14,     /* DR_REG_R14 */
+    DR_REG_V15,     /* DR_REG_R15 */
```
derekbruening, line 260: (Would be nice to map just from the pointer-sized reg_fixer ones but it's not clear how to actually set that up...)

api/docs/release.dox
edeiana: I like the idea of having a single API to get a register's name (regardless of which real/synthetic ISA).
I guess we can use this as an opportunity to use the recent naming convention and have a `dr_get_register_name()` public API, which would deprecate `get_register_name()`.

Do we plan to use any information from `dcontext_t` other than its `isa_mode`?
If not, I think it would be better to have the new `dr_get_register_name()` just take a `dr_isa_mode_t` (and the, already present, `reg_id_t`).
This way a user can decide from where the isa_mode comes from (maybe there are use cases where having it coming from `instr_t` makes more sense?), instead of being "forced" to have a specific ISA set in `dcontext`.

core/ir/aarch64/encode.c
edeiana: Great! :)

core/ir/arm/encode.c
edeiana: To be honest, I think I just turned off my brain for a second here and just blindly followed what `dr_reg_fixer[]` does in core/ir/arm/encode.c (https://github.com/DynamoRIO/dynamorio/blob/master/core/ir/arm/encode.c#L287).
I am not even really sure why the motivation behind having those extra register is "but simpler code to not ifdef it" (old comment from a time when arm and aarch64 where together, similar to x86 and x64, perhaps?), as there's nothing to ifdef here; the `dr_reg_fixer[]` (and `d_r_reg_id_to_virtual[]`) indexes for aarch64 registers should just be mapped to DR_REG_INVALID, if I understand this correctly.

Either way, I think both `dr_reg_fixer[]` and `d_r_reg_id_to_virtual[]` should look "the same" (as we are asking future maintainers to keep them "in synch"), so we either keep them as they are or change both? Advice?

api/docs/release.dox
derekbruening: If this were the only such API routine I would agree -- but as my comment says we should consider all the other existing register and opnd_t and instr_t routines.  I don't know that we'd make new versions (that's what I meant by `_ex` versions) of every single one to pass in an ISA parameter.  We might have to go with a global (or per-dcontext in managed mode) state instead.

Plus, there is other state like the ISA: the SVE vector length, e.g.  If we took in a parameter it shouldn't just be an ISA but some kind of "ISA state" structure we can append to.

core/ir/instr_shared.c
edeiana: Right, done.

core/ir/isa_regdeps/decode.c
edeiana: Done.

core/ir/opnd_api.h
edeiana: Done.

core/ir/opnd_api.h
edeiana: Sounds good.
I was wondering if we had a place where we do some "sanity checks", good to know it's `d_r_arch_init()`.

edeiana: But then for non-x86 `DR_REG_V185` would be `==2`, which is `==DR_REG_V0`.
We could `ifdef` it, but honestly I find it a bit confusing (also, considering that we want to move away from having to build DR for a specific arch, we'd be adding to the "problem").
A similar argument could be made for `DR_REG_V0 = DR_REG_NULL + 2`. It really needs to be +2 for non-x86, for x86 could be +1. This could also be ifdef.

It looks clearer to me without ifdefs (it's a nit though, so I can ifdef it if you think it's better).

core/ir/x86/encode.c
edeiana: I was thinking about this as well (https://github.com/DynamoRIO/dynamorio/pull/6783#discussion_r1571517251).
We could have the indexes of non-containing registers map to `DR_REG_INVALID`.
But then a user would always need to call `dr_reg_to_pointer_size()` first (otherwise they'd get a `DR_REG_INVALID`). So we might just as well save us the headache of documenting that fact and allow "users" to just pass any register and get the "correct" (w.r.t. DR_ISA_REGDEPS) mapping to virtual registers?

api/docs/release.dox
edeiana: Oh, ok, we don't want a new API.
I'll look into "per-dcontext managed mode".
I think if we want to look into having a new "ISA state" structure (rather than just an enum), we should do it in a separate PR and decide there what we want to include.

core/ir/arm/encode.c
derekbruening: I guess it is just to avoid ifdefs.  Just leave as is I suppose.

core/ir/x86/encode.c
derekbruening: If we have the full array anyway, yes.  Was hoping we could map from just the pointer-sized but we need an associative container for that: not worth it.

edeiana: Ok, leaving as is.

core/ir/opnd_api.h
edeiana: > I guess a loop over the enum values ensuring none equals _NULL or _INVALID, probably under DOCHECK level 1 or 2.

Actually, we have a gap in the enum values (because of x86 DR_REG_INVALID == 187), so we can't really loop through them. I "unrolled" the loop with all DR_REG_V values, but I don't know if it's acceptable.
Perhaps we want to move that "monstrosity" into the tests in suite/tests/api/isa_regdeps.c?
Alternatively we could just check DR_REG_V0 and DR_REG_185 to be != than _NULL or _INVALID?

api/docs/release.dox
```diff
@@ -227,7 +227,6 @@ Further non-compatibility-affecting changes include:
  - Added instr_convert_to_isa_regdeps() API that converts an #instr_t from a real ISA
    (e.g., #DR_ISA_AMD64) to the #DR_ISA_REGDEPS synthetic ISA.
```
edeiana, line 229: Do we need to document the fact the ger_register_name() now uses the global context to determine what name to return? (virtual or real ISA register)

core/arch/arch.c
```diff
@@ -676,10 +677,280 @@ far_ibl_set_targets(ibl_code_t src_ibl[], ibl_code_t tgt_ibl[])
...
+#define ASSERT_NOT_NULL_OR_INVALID(x)                                            \
+    do {                                                                         \
+        reg_id_t reg_virtual = x;                                                \
+        ASSERT((reg_virtual != DR_REG_NULL) && (reg_virtual != DR_REG_INVALID)); \
+    } while (0);
+
+/* We can't loop through DR_REG_V enums because we have a gap for value 187
+ * (== DR_REG_INVALID for x86), so we unroll the loop and compare every enum manually.
+ */
+#define CHECK_VIRTUAL_REGISTER_IDS          \
```
edeiana, line 693: Not sure we want to keep this very long sequence of check here. Perhaps we want to move it in suite/tests/api/ir_regdeps.c?

core/ir/opnd_shared.c
```diff
@@ -2407,6 +2408,10 @@ opnd_compute_address(opnd_t opnd, dr_mcontext_t *mc)
 const char *
 get_register_name(reg_id_t reg)
 {
+    dcontext_t *dcontext = get_thread_private_dcontext();
```
edeiana, line 2411: Is this the right way of getting the global dcontext?
(It behaves as expected in the tests, I'm just not sure I'm taking every corner case into consideration)

core/ir/opnd_api.h
edeiana: https://github.com/DynamoRIO/dynamorio/pull/6783#discussion_r1575498611

core/ir/opnd_shared.c
derekbruening: + get_t_p_d can return NULL; dr_get_isa_mode handles that; call get_t* inside that call to avoid someone using the dcontext accidentally?  Or add a comment that it can be NULL.

+ Wondering about the overhead of calling get_t* in every little IR function: some of which are inlined as perf matters on all the getter and setter routines.  But maybe only a small % of these routines will need dc?  Is it all the register conversion ones, but none of the instr or opnd getter or setters?  I would add a comment about this to remind us when we expand for #1684.

+ Re: the prior conversation on an explicit parameter to be consistent w/ decode() and other key routines taking in dcontext: goes back to how many we'd have to change.  There is some precedence for IR routines to call get_t_p_d: instr_get_eflags, opnd_create_far_abs_addr, others.  Maybe include a mention of this in the aforementioned comment discussing tradeoffs/options.

+ Any issues w/ the calling thread being different from the target thread/dcontext: maybe.  Xref #6690.

core/ir/opnd_api.h
```diff
@@ -3266,18 +3266,13 @@ DR_API
 /**
  * Assumes that \p reg is a DR_REG_ 32-bit register constant.
  * Returns the string name for \p reg.
+ * \note It uses the global dcontext_t to determine the ISA mode.  If the ISA mode is a
+ * synthetic one (e.g., #DR_ISA_REGDEPS), it returns the name of a DR_REG_V virtual
```
derekbruening, line 3270: It may be worth replacing all the `DR_REG_V` with `#DR_REG_V0, etc.` or something so we have a real link: here and in all Doxygen comments added by this PR.

core/ir/isa_regdeps/decode.c
```diff
@@ -98,14 +98,20 @@ decode_isa_regdeps(dcontext_t *dcontext, byte *encoded_instr, instr_t *instr)
...
-         * instructions may have different sizes.  However, querying the size of
-         * a virtual register is not supported on purpose, so the user should not
-         * expect opnd_t.size to have any meaningful value.  We do the same for
-         * both src and dst register operands of DR_ISA_REGDEPS instructions.
+        /* Virtual registers don't have a fixed size like real ISA registers do.
+         * So, it can happen that the same virtual register in two different
+         * instructions may have different sizes.
+         *
+         * Even though querying the size of a virtual register is not supported on
+         * purpose (a user should query the instr_t.operation_size); we set the
```
derekbruening, line 106: nit: s/;/,/

core/ir/instr_shared.c
```diff
@@ -3102,14 +3103,20 @@ instr_convert_to_isa_regdeps(void *drcontext, instr_t *instr_real_isa,
...
-                 * instructions may have different sizes.  However, querying the size of
-                 * a virtual register is not supported on purpose, so the user should not
-                 * expect opnd_t.size to have any meaningful value.  We do the same for
-                 * both src and dst register operands of DR_ISA_REGDEPS instructions.
+                /* Virtual registers don't have a fixed size like real ISA registers do.
+                 * So, it can happen that the same virtual register in two different
+                 * instructions may have different sizes.
+                 *
+                 * Even though querying the size of a virtual register is not supported on
+                 * purpose (a user should query the instr_t.operation_size); we set the
```
derekbruening, line 3111: nit: s/;/,/

api/docs/release.dox
derekbruening: I think the local note is enough, because the default behavior won't change as the dcontext ISA should be set to the expected default.

core/arch/arch.c
derekbruening: Yes move to test.  Possible alternative is to just check the d_r_reg_virtual_names array for NULL and INVALID...it assumes that array is in synch w/ the enums.

derekbruening (approved):

core/arch/arch.c
edeiana: Moved to test. Done.

core/ir/isa_regdeps/decode.c
edeiana: Done.

core/ir/instr_shared.c
edeiana: Done.

core/ir/opnd_api.h
edeiana: Done.
Turns out that that was the only doxygen comment (we removed the other public API).

core/ir/opnd_shared.c
edeiana: Thank you for the thorough explanation!
Added comments.
And put get_thread_private_dcontext() as argument to get_isa_mode().
Done.

api/docs/release.dox
edeiana: Ok!

PR #6791, opened 2024-04-25
i#6426 sched stats: Print list of threads

derekbruening:
Adds the list of threads per cpu to the schedule_stats output. Updates the schedule_stats tests to confirm some tids are printed.

Issue: #6426

### DIFF
--- a/clients/drcachesim/tools/schedule_stats.cpp
+++ b/clients/drcachesim/tools/schedule_stats.cpp
@@ -330,7 +330,13 @@ schedule_stats_t::print_percentage(double numerator, double denominator,
 void
 schedule_stats_t::print_counters(const counters_t &counters)
 {
-    std::cerr << std::setw(12) << counters.threads.size() << " threads\n";
+    std::cerr << std::setw(12) << counters.threads.size() << " threads";
+    if (!counters.threads.empty()) {
+        std::cerr << ":";
+        for (auto thread : counters.threads)
+            std::cerr << " " << thread;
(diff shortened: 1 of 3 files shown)


derekbruening:
x32 failure is yet again opcode_mix #6303 size assert

clients/drcachesim/tools/schedule_stats.cpp
```diff
@@ -330,7 +330,13 @@ schedule_stats_t::print_percentage(double numerator, double denominator,
 void
 schedule_stats_t::print_counters(const counters_t &counters)
 {
-    std::cerr << std::setw(12) << counters.threads.size() << " threads\n";
+    std::cerr << std::setw(12) << counters.threads.size() << " threads";
+    if (!counters.threads.empty()) {
+        std::cerr << ":";
+        for (auto thread : counters.threads)
+            std::cerr << " " << thread;
```
abhinav92003, line 337: Would comma-separated be nicer?

abhinav92003 (approved):

clients/drcachesim/tools/schedule_stats.cpp
derekbruening: Hmm, maybe.  I was dealing with thread lists from another source which were just space-separated.  Comma might be easier to auto-parse.  Commas when iterating a set do require a slightly awkward loop, checking end inside the loop.  I guess we can live with that.  OK, done.

Issue #6728, opened 2024-03-26
schedule_stats overcounts idle time


derekbruening:
I ran schedule_stats on a record_filter --sched_randomize_inputs written trace and the output doesn't look right:
```
Schedule stats tool results:
Total counts:
         112 cores
        6410 threads
 68851855065 instructions
      409948 total context switches
   0.0059541 CSPKI (context switches per 1000 instructions)
      167953 instructions per context switch
      320900 voluntary context switches
       49747 direct context switches
       78.28% voluntary switches
       12.13% direct switches
     2606536 system calls
     1416152 maybe-blocking system calls
      110869 direct switch requests
270593082177 waits
     5367001 idles
       99.99% cpu busy by record count
   129531748 cpu microseconds
  7107011849 idle microseconds
  6754507124 idle microseconds at last instr
        1.79% cpu busy by time
        1.88% cpu busy by time, ignoring idle past last instr
```
The idle time looks inflated, given the low # of idle records compared to instruction records. Looking at the source code: the code looks like it won't count the time for the final idle record in a sequence. It also looks like it will count all the instrs since the last cxt sw leading up to the idle time?!

PR #6792, opened 2024-04-29
i#6728 sched time stats: Relax assert for zero time

derekbruening:
Relaxes an assert added in PR #6784 to allow zero time to have elapsed when updating the schedule_stats time buckets.

Issue: #6728


abhinav92003 (approved):

derekbruening:
x32 failure is opcode_mix #6303

Issue #6793, opened 2024-04-30
[drmemtrace analyzer] Incorrect end timestamp and instr counts in shard-final interval


abhinav92003:
The shard-final interval seems to be created with incorrect end instr count and end timestamp. This is likely because the analyzer framework worker processes all assigned shards before invoking the shard exit event which then creates the final interval for the shard.

abhinav92003:
This seems to affect traces that are missing a thread exit event (https://github.com/DynamoRIO/dynamorio/issues/6444, which was fixed in Nov 2023).

Due to the missing thread exit event, the fallback parallel_shard_exit at the very end (after all shards assigned to the worker are processed) is invoked: https://github.com/DynamoRIO/dynamorio/blob/6fb84c6ef4b31bd40120db9107be455545d0204e/clients/drcachesim/analyzer.cpp#L696

For other traces, the existing implementation already invokes parallel_shard_exit at the right time (immediately when the thread exit is seen, or at the end of core-sharded trace).

PR #6794, opened 2024-05-01
i#6793: Skip incorrect final interval during trace analysis

abhinav92003:
Skips generating the final interval snapshot for thread-shards that are missing the thread exit entry, during trace interval analysis in the drmemtrace analyzer framework. Such final intervals currently have incorrect data (the cumulative instr count, instr count delta, and end timestamp). Instead of generating incorrect data we skip such a final interval and print a non-fatal error message.

This issue is seen on interval analysis of only the traces affected by i#6444. That issue has already been fixed many months ago, and does not affect newer traces. This fix is to allow trace interval analysis of those older traces.

Due to i#6444, some threads were missing thread exit entries because tracing ended in an rseq region. For such traces, the i#6444 fix already added an analyzer workaround that ensures that process_shard_exit is called despite the missing thread exit event. Here we modify that workaround to skip processing the final interval when the workaround had to be invoked; this is required because the worker's trace stream would have moved on to other shards (if there were others assigned to the worker) and therefore cannot anymore provide the state (instr count, timestamp when the shard ended) required for generation of the final interval snapshot for that shard.

Note that for the shard that was the last shard in the worker's stream, we can still continue to generate the final interval snapshot, because the worker's trace stream is still good to generate state for that shard.

Adds modified version of existing interval analysis unit tests to verify the workaround.

Also fixes a log which may not print the correct stream name if the error was in a different shard than the current one (because of the same reason we're seeing the primary issue).

Issue: #6444
Fixes: #6793

### DIFF
--- a/clients/drcachesim/analyzer.cpp
+++ b/clients/drcachesim/analyzer.cpp
@@ -578,9 +578,9 @@ analyzer_tmpl_t<RecordType, ReaderType>::process_shard_exit(
     worker->shard_data[shard_index].exited = true;
     if (interval_microseconds_ != 0 || interval_instr_count_ != 0) {
         if (!do_process_final_interval) {
-            ERRMSG(
-                "i#6793: Skipping process_interval for final interval of shard id %d\n",
-                shard_index);
+            ERRMSG("i#6793: Skipping process_interval for final interval of shard index "
+                   "%d\n",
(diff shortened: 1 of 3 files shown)


brettcoon (approved):
Thanks for resolving this!

PR #6795, opened 2024-05-01
i#5036 Update scatter/gather docs with AArch64 details

jackgallagher-arm:
Adds details of the AArch64 scatter/gather expansion to the scatter/gather expansion developer documentation.

Issue: #5036

### DIFF
--- a/api/docs/scatter_gather_emulation.dox
+++ b/api/docs/scatter_gather_emulation.dox
@@ -182,8 +182,8 @@ if it is 16 bytes, and cannot be encoded for vector lengths > 32 bytes.
 ### Non-faulting loads
 
 Non-faulting loads (`ldnf*`) do not fault if an element read faults. Instead a special
-predicate register FFR is updated. If this happends the FFR element corresponding to the
-element that faulted, and all elements higher that that are set to 0. Elements lower than
+predicate register FFR is updated. If this happens the FFR element corresponding to the
+element that faulted, and all elements higher than that are set to 0. Elements lower than
 the faulting element are unchanged.
 Non-faulting loads support scalar+immediate addressing.
 
@@ -526,7 +526,7 @@ ldff1sb (%x1,%z2.d)[1byte] %p3/z -> %z28.d
    ldr    (%x0)[8byte] -> %x0
    ldr    +0x10(%x0)[8byte] -> %x0
    ldr    (%x0)[8byte] -> %x0
-   str    %p0 -> (%x0)[4byte]            // Spill a predicate register to use as the loop varaible mask
+   str    %p0 -> (%x0)[4byte]            // Spill a predicate register to use as the loop variable mask
    <label note=0x0000000000000001>
    dup    $0x00 lsl $0x00 -> %z28.d      // Clear destination register
    pfalse  -> %p0.b                      // Initialize loop variable to 0
--- a/api/docs/scatter_gather_emulation.dox
+++ b/api/docs/scatter_gather_emulation.dox
@@ -1,5 +1,6 @@
 /* ******************************************************************************
  * Copyright (c) 2010-2022 Google, Inc.  All rights reserved.
+ * Copyright (c) 2024      Arm Limited   All rights reserved.
  * ******************************************************************************/
 
 /*
@@ -32,11 +33,13 @@
 
 /**
  ****************************************************************************
-\page page_scatter_gather_emulation Emulating x86 Scatter and Gather Instructions
+\page page_scatter_gather_emulation Emulating Scatter and Gather Instructions
 
 \tableofcontents
 
 # Background
+
+## x86
 
 The x86 gather and scatter instructions were introduced in the AVX2 and AVX512
 instruction set extensions. They allow loading or storing a subset of elements in a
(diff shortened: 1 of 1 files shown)


api/docs/scatter_gather_emulation.dox
```diff
@@ -71,6 +74,126 @@ register `k1`. Elements may be scattered in any order. When an element is stored
 its mask is cleared. If some store faults, all elements to its right (closer to
 LSB) will be complete.
 
+## AArch64
+
+The AArch64 SVE (Scalable Vector Extension) introduced the first AArch64 scatter and
```
abhinav92003, line 79: A link to the manual would be nice here.

api/docs/scatter_gather_emulation.dox
```diff
@@ -71,6 +74,126 @@ register `k1`. Elements may be scattered in any order. When an element is stored
...
+in the vector index register `z0`. Elements that are not active in the mask contained by
+the governing predicate register `p0` are not loaded and the corresponding element in the
+destination register `z1` is set to `0`.
+
+```
+st1h   %z17.d %p5 -> (%x19,%z20.d)[2byte]
+```
+
+Above is a scalar+vector scatter store that writes the lowest 16-bits of the 64-bit
+elements of the `z17` vector register. Addresses are calculated by adding the base
```
abhinav92003, line 113: The above ld1b computed address using the 32-bit elements in z0, however here st1h uses 64-bit elements in z20. How is that determined?

api/docs/scatter_gather_emulation.dox
```diff
@@ -71,6 +74,126 @@ register `k1`. Elements may be scattered in any order. When an element is stored
...
+address from the scalar `x19` register to the corresponding 64-bit element in the vector
+index register `z20`. Elements that are not active in the mask contained by the governing
+predicate register `p5` are not stored.
+
+### Vector+immediate
+
+```
+ld1sw  +0x18(%z8.d)[4byte] %p2/z -> %z6.d
```
abhinav92003, line 121: What does the `/z` mean? (Does it mean zeroing? But that's the only behavior so why that need be specified?)

abhinav92003, line 121: Curious: if I replaced .d with .h, would that SIGILL? (since the instr would be to load 32-bit values in 16-bit vector elements)

api/docs/scatter_gather_emulation.dox
```diff
@@ -71,6 +74,126 @@ register `k1`. Elements may be scattered in any order. When an element is stored
...
+Above is a vector+immediate gather load that reads 32-bit values which are sign-extended
+and written to 64-bit elements of the `z6` vector register. Addresses are calculated by
+adding the immedate value 0x18 to the corresponding 64-bit element in the vector base
+register `z8`. Elements that are not active in the mask contained by governing predicate
+register `p2` are zeroed.
+
+### Vector+scalar
+
+```
+stnt1d %z27.d %p7 -> (%z25.d,%x29)[8byte]
```
abhinav92003, line 132: Can the vector reg suffix be different than the opcode? E.g., can we use .h here? which would mean storing 16-bit vector elements to 64-bit memory elements?

api/docs/scatter_gather_emulation.dox
```diff
@@ -71,6 +74,126 @@ register `k1`. Elements may be scattered in any order. When an element is stored
...
+
+### Scalar+immediate
+
+```
+st1w   %z10.s %p3 -> -0x60(%x11)[4byte]
+```
+The above instruction is a scalar+immediate predicated contiguous store that writes the
+32-bit elements of the `z10` vector register. The address for the first element is
+calculated by adding the immediate value `-0x60` to the value of the scalar base register
+`x11`. Addresses for subsequent elements are calculated by adding the size of the stored
```
abhinav92003, line 163: Xref the other comment about "Can the vector reg suffix be different than the opcode?"

"size of the stored value": if the destination elements are wider than the vector element, does this mean the address is incremented by the size of the stored value **at destination**?

api/docs/scatter_gather_emulation.dox
```diff
@@ -71,6 +74,126 @@ register `k1`. Elements may be scattered in any order. When an element is stored
...
+of the first element is calculated by adding the value of the scalar index register `x18`
+to the value of the scalar base register `x17`. Addresses for subsequent elements are
+calculated by adding the size of the loaded value (1-byte in this example) to the address
+of the previous element. Elements that are not active in the mask contained by governing
+predicate register `p5` are zeroed.
+
+### Scalar+immediate
+
+```
+st1w   %z10.s %p3 -> -0x60(%x11)[4byte]
```
abhinav92003, line 158: Could you remind me what ".s" stands for? single-precision floats?

api/docs/scatter_gather_emulation.dox
```diff
@@ -120,16 +239,23 @@ num_accesses = vector_size / element_size
...
   extract mask for the ith access from mask reg or mask vector
   if mask is set, then
-    extract ith element of index vector
-    compute address = base + ith index element
+    if index is vector, then
+        extract ith element of index vector
+        compute address = base + ith index element
+    else // base is vector
+        extract ith element of base vector
+        compute address = ith base element + index
```
abhinav92003, line 247: The index is multiplied by vector size too?

api/docs/scatter_gather_emulation.dox
```diff
@@ -120,16 +239,23 @@ num_accesses = vector_size / element_size
...
     if instr_is_gather, then
       load data from address into a scalar reg
       insert scalar data into destination vector
     else // instr_is_scatter
       extract scalar data from source vector to scalar reg
       store data from scalar reg to address
     done
-    clear ith mask in mask reg or mask vector
+    if x86, then
+        clear ith mask in mask reg or mask vector
```
abhinav92003, line 257: Doesn't aarch64 do the same to the predicate reg? If some operation faults, what would the predicate reg show?

abhinav92003 (commented):
Taking this opportunity to learn a bit more about SVE. Apologies if the questions are too naive. If some clarification would be useful to general audience, please add it to the dox.

api/docs/scatter_gather_emulation.dox
jackgallagher-arm: No. The governing predicate register is never modified by any of the scatter/gather instructions, whether there is a fault or not.
For first-fault and non-fault loads there is a special first-fault predicate register (FFR) which is modified (I have added details about that to this doc)

api/docs/scatter_gather_emulation.dox
jackgallagher-arm: `.b` byte
`.h` halfword (16-bit)
`.s` singleword (32-bit)
`.d` doubleword (64-bit)
`.q` quadword (128-bit) (though none of the scatter/gather instructions here use quadword)

api/docs/scatter_gather_emulation.dox
jackgallagher-arm: That instruction doesn't exist so you wouldn't be able to encode it. Only combinations that make sense are valid encodings.

jackgallagher-arm: `/z` does mean zeroing. Other SVE instructions support `/m` merging behaviour as well. It just happens that all current SVE loads only support zeroing.

api/docs/scatter_gather_emulation.dox
jackgallagher-arm: The element size is encoded in the instruction. In the disassembly the element size is indicated by the `.s` or `.d` after the register name (`s` for 32-bit singleword, `d` for 64-bit doubleword).

api/docs/scatter_gather_emulation.dox
jackgallagher-arm: Yes it can, but not in the way you suggest. Which combinations are valid depends on the instruction but in general the element size must always be >= to the opcode value size.
So you can't have a store that stores 16-bit vector elements to 64-bit memory locations.
`stnt1d %z27.h %p7 -> (%z25.d,%x29)[8byte] // invalid`
But you can have a store that stores the low 16-bits of 64-bit elements to 16-bit memory locations:
`stnt1h %z27.d %p7 -> (%z25.d,%x29)[8byte] // valid`


For example scalar+scalar `st1h`:
`ST1H { <Zt>.<T> }, <Pg>, [<Xn|SP>, <Xm>, LSL #1]`

allows `.h`, `.s`, or `.d` element sizes. For the `.s` and `.d` variants it stores the lowest halfword (16-bits) of each element.

api/docs/scatter_gather_emulation.dox
```diff
@@ -71,6 +74,126 @@ register `k1`. Elements may be scattered in any order. When an element is stored
...
+```
+st1w z10.s, p3, [x11, #-6, MUL VL]
+```
+if it is 16 bytes, and cannot be encoded for vector lengths > 32 bytes.
+
+### Non-faulting loads
+
+Non-faulting loads (`ldnf*`) do not fault if an element read faults. Instead a special
+predicate register FFR is updated. If this happends the FFR element corresponding to the
+element that faulted, and all elements higher that that are set to 0. Elements lower than
```
derekbruening, line 186: s/that that/than that/

api/docs/scatter_gather_emulation.dox
derekbruening: I believe on x86 a scatter/gather can be interrupted in the middle and resumed and that's what the mask reg update is for: so the resumption point is right there in the arch state.  Kind of like x86's `rep movs` and other string loops where the remaining trip count is in a register making it easy to restart from the middle.

On a64 can a scatter/gather not be interrupted in the middle and then resumed?  Is that why the faulting behavior is the way it is, zeroing out the future operations on a fault?

api/docs/scatter_gather_emulation.dox
```diff
@@ -376,9 +502,192 @@ vpscatterdd {%k1} %xmm10 -> 0x00402039(,%xmm11,4)[4byte] %k1
...
+   ldr    +0x10(%x0)[8byte] -> %x0
+   ldr    +0x20(%x0)[8byte] -> %x0
+   str    %z28 -> (%x0)[32byte]          // Save the value of the destination register in case we
+                                         // need to restore its value on a fault.
+   ldr    +0x38(%x28)[8byte] -> %x0
+   ldr    +0x0f50(%x0)[8byte] -> %x0
+   ldr    (%x0)[8byte] -> %x0
+   ldr    +0x10(%x0)[8byte] -> %x0
+   ldr    (%x0)[8byte] -> %x0
+   str    %p0 -> (%x0)[4byte]            // Spill a predicate register to use as the loop varaible mask
```
derekbruening, line 529: s/varaible/variable/

derekbruening (approved):

api/docs/scatter_gather_emulation.dox
jackgallagher-arm: The stored value size is never wider than the vector element size.

api/docs/scatter_gather_emulation.dox
jackgallagher-arm: The index isn't scaled by the vector length. Some instructions are unscaled, some scale the index by the element size.

api/docs/scatter_gather_emulation.dox
jackgallagher-arm: Done.

api/docs/scatter_gather_emulation.dox
jackgallagher-arm: Done.

api/docs/scatter_gather_emulation.dox
jackgallagher-arm: Done.

api/docs/scatter_gather_emulation.dox
abhinav92003: My question is actually about the size of the addresses. The above ld1b entry says:

```
Addresses are calculated by
adding the base address from the scalar `x0` register to the corresponding 32-bit element
in the vector index register `z0`.
```

whereas this st1h says:

````
Addresses are calculated by adding the base
address from the scalar `x19` register to the corresponding 64-bit element in the vector
index register `z20`
```

What determined the 32- vs 64-bit?

abhinav92003: Oh I guess that's the `.s` and `.d` respectively in %z0.s and %z20.d

abhinav92003: Thanks

api/docs/scatter_gather_emulation.dox
abhinav92003: > So you can't have a store that stores 16-bit vector elements to 64-bit memory locations.

Okay. I may have guessed that the remaining part of the 64-bit memory location is zero-extended or something. Thanks for clarifying.

abhinav92003 (approved):

PR #6799, opened 2024-05-07
i#6662 public traces, part 3: regdeps disasm

edeiana:
Adds disassembling for DR_ISA_REGDEPS instructions.
Specifically, when we disassemble DR_ISA_REGDEPS instructions, we print the 
instruction encoding (divided in 4 byte words, can span one or two lines, similar to x86),
we substitute the opcode with categories, we print the operations size (e.g., `[4byte]`),
and then the source and destination virtual register names (e.g., `%rv3`).

Disassembled instructions look as follows:
`00000812 06260606 load [8byte]       %rv4 -> %rv4 %rv36`

In general, they follow this pattern:
[encoding (in 4 byte words)]  [categories]  [operation_size]  [src_regs -> dst_regs]

Issue: #6662

### DIFF
--- a/clients/drcachesim/tests/view_test.cpp
+++ b/clients/drcachesim/tests/view_test.cpp
@@ -587,10 +587,12 @@ run_chunk_tests(void *drcontext)
 }
 
 /* Test view tool on a OFFLINE_FILE_TYPE_ARCH_REGDEPS trace.
+ * The trace hardcoded in entries is X64, so we only test on X64 architectures here.
  */
 bool
 run_regdeps_test(void *drcontext)
 {
+#ifdef X64
     const memref_tid_t t1 = 3;
     std::vector<memref_tid_t> tids = { t1 };
     constexpr addr_t PC = 0x00007f6fdd3ec360;
@@ -621,8 +623,8 @@ run_regdeps_test(void *drcontext)
            6           0:           3 <marker: tid 3 on core 2>
            7           1:           3 ifetch       )DELIM") +
         std::string(R"DELIM(3 byte(s) @ 0x00007f6fdd3ec360 )DELIM") +
-        std::string(R"DELIM(11 00 01 00 06 09 06 move   %rv4[8byte] -> %rv7[8byte]
-           7           1:           3                                             00
+        std::string(
+            R"DELIM(11 00 01 00 06 09 06 00 move   %rv4[8byte] -> %rv7[8byte]
 )DELIM");
     instrlist_t *ilist_unused = nullptr;
     view_nomod_test_t view(drcontext, *ilist_unused, 0, 0);
@@ -631,6 +633,7 @@ run_regdeps_test(void *drcontext)
         std::cerr << "Output mismatch: got |" << res << "| expected |" << expect << "|\n";
         return false;
     }
+#endif
     return true;
 }
 
--- a/clients/drcachesim/tests/view_test.cpp
+++ b/clients/drcachesim/tests/view_test.cpp
@@ -596,7 +596,11 @@ run_regdeps_test(void *drcontext)
     const memref_tid_t t1 = 3;
     std::vector<memref_tid_t> tids = { t1 };
     constexpr addr_t PC = 0x00007f6fdd3ec360;
+    constexpr addr_t PC_vpmovmskb = 0x00007f6fdc76cb35;
+    constexpr addr_t PC_vmovdqu = 0x00007f6fdc76cb2d;
     constexpr addr_t ENCODING_REGDEPS_ISA = 0x0006090600010011;
+    constexpr addr_t ENCODING_REGDEPS_ISA_vpmovmskb = 0x004e032100004011;
+    constexpr addr_t ENCODING_REGDEPS_ISA_vmovdqu = 0x0009492100004811;
     std::vector<std::vector<trace_entry_t>> entries = { {
         { TRACE_TYPE_HEADER, 0, { 0x1 } },
         { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_VERSION, { 3 } },
@@ -607,25 +611,49 @@ run_regdeps_test(void *drcontext)
         { TRACE_TYPE_THREAD, 0, { t1 } },
         { TRACE_TYPE_PID, 0, { t1 } },
         { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_CACHE_LINE_SIZE, { 64 } },
-        { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_CHUNK_INSTR_COUNT, { 2 } },
+        { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_CHUNK_INSTR_COUNT, { 3 } },
         { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_TIMESTAMP, { 1002 } },
         { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_CPU_ID, { 2 } },
         { TRACE_TYPE_ENCODING, 8, { ENCODING_REGDEPS_ISA } },
         { TRACE_TYPE_INSTR, 3, { PC } },
+        { TRACE_TYPE_ENCODING, 16, { ENCODING_REGDEPS_ISA_vpmovmskb } },
+        { TRACE_TYPE_INSTR, 4, { PC_vpmovmskb } },
+        { TRACE_TYPE_ENCODING, 16, { ENCODING_REGDEPS_ISA_vmovdqu } },
+        { TRACE_TYPE_INSTR, 4, { PC_vmovdqu } },
         { TRACE_TYPE_FOOTER, 0, { 0 } },
     } };
     std::string expect =
         std::string(R"DELIM(           1           0:           3 <marker: version 3>
            2           0:           3 <marker: filetype 0x20e00>
            3           0:           3 <marker: cache line size 64>
-           4           0:           3 <marker: chunk instruction count 2>
+           4           0:           3 <marker: chunk instruction count 3>
            5           0:           3 <marker: timestamp 1002>
            6           0:           3 <marker: tid 3 on core 2>
            7           1:           3 ifetch       )DELIM") +
         std::string(R"DELIM(3 byte(s) @ 0x00007f6fdd3ec360 )DELIM") +
+        std::string(R"DELIM(00010011 move [8byte]       %rv4 -> %rv7
+)DELIM") +
+        std::string(R"DELIM(           7           1:           3)DELIM") +
         std::string(
-            R"DELIM(11 00 01 00 06 09 06 00 move   %rv4[8byte] -> %rv7[8byte]
+            R"DELIM(                                             00060906
+)DELIM") +
+        std::string(R"DELIM(           8           2:           3 ifetch       )DELIM") +
+        std::string(R"DELIM(4 byte(s) @ 0x00007f6fdc76cb35 )DELIM") +
+        std::string(R"DELIM(00004011 simd [32byte]       %rv76 -> %rv1
+)DELIM") +
+        std::string(R"DELIM(           8           2:           3)DELIM") +
+        std::string(
+            R"DELIM(                                             004e0321
+)DELIM") +
+        std::string(R"DELIM(           9           3:           3 ifetch       )DELIM") +
+        std::string(R"DELIM(4 byte(s) @ 0x00007f6fdc76cb2d )DELIM") +
+        std::string(R"DELIM(00004811 load simd [32byte]       %rv7 -> %rv71
+)DELIM") +
+        std::string(R"DELIM(           9           3:           3)DELIM") +
+        std::string(
+            R"DELIM(                                             00094921
 )DELIM");
+
     instrlist_t *ilist_unused = nullptr;
     view_nomod_test_t view(drcontext, *ilist_unused, 0, 0);
     std::string res = run_serial_test_helper(view, entries, tids);
--- a/clients/drcachesim/tests/view_test.cpp
+++ b/clients/drcachesim/tests/view_test.cpp
@@ -586,11 +586,57 @@ run_chunk_tests(void *drcontext)
     return run_single_thread_chunk_test(drcontext) && run_serial_chunk_test(drcontext);
 }
 
+/* Test view tool on a OFFLINE_FILE_TYPE_ARCH_REGDEPS trace.
+ */
+bool
+run_regdeps_test(void *drcontext)
+{
+    const memref_tid_t t1 = 3;
+    std::vector<memref_tid_t> tids = { t1 };
+    constexpr addr_t PC = 0x00007f6fdd3ec360;
+    constexpr addr_t ENCODING_REGDEPS_ISA = 0x0006090600010011;
+    std::vector<std::vector<trace_entry_t>> entries = { {
+        { TRACE_TYPE_HEADER, 0, { 0x1 } },
+        { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_VERSION, { 3 } },
+        { TRACE_TYPE_MARKER,
+          TRACE_MARKER_TYPE_FILETYPE,
+          { OFFLINE_FILE_TYPE_ARCH_REGDEPS | OFFLINE_FILE_TYPE_ENCODINGS |
+            OFFLINE_FILE_TYPE_SYSCALL_NUMBERS | OFFLINE_FILE_TYPE_BLOCKING_SYSCALLS } },
+        { TRACE_TYPE_THREAD, 0, { t1 } },
+        { TRACE_TYPE_PID, 0, { t1 } },
+        { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_CACHE_LINE_SIZE, { 64 } },
+        { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_CHUNK_INSTR_COUNT, { 2 } },
+        { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_TIMESTAMP, { 1002 } },
+        { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_CPU_ID, { 2 } },
+        { TRACE_TYPE_ENCODING, 8, { ENCODING_REGDEPS_ISA } },
+        { TRACE_TYPE_INSTR, 3, { PC } },
+        { TRACE_TYPE_FOOTER, 0, { 0 } },
+    } };
+    const char *expect = R"DELIM(           1           0:           3 <marker: version 3>
+           2           0:           3 <marker: filetype 0x20e00>
+           3           0:           3 <marker: cache line size 64>
+           4           0:           3 <marker: chunk instruction count 2>
+           5           0:           3 <marker: timestamp 1002>
+           6           0:           3 <marker: tid 3 on core 2>
+           7           1:           3 ifetch       3 byte(s) @ 0x00007f6fdd3ec360 11 00 01 00 06 09 06 move   %rv4[8byte] -> %rv7[8byte]
+           7           1:           3                                             00
+)DELIM";
+    instrlist_t *ilist_unused = nullptr;
+    view_nomod_test_t view(drcontext, *ilist_unused, 0, 0);
+    std::string res = run_serial_test_helper(view, entries, tids);
+    if (res != expect) {
+        std::cerr << "Output mismatch: got |" << res << "| expected |" << expect << "|\n";
+        return false;
+    }
+    return true;
+}
+
 int
 test_main(int argc, const char *argv[])
 {
     void *drcontext = dr_standalone_init();
-    if (run_limit_tests(drcontext) && run_chunk_tests(drcontext)) {
+    if (run_limit_tests(drcontext) && run_chunk_tests(drcontext) &&
+        run_regdeps_test(drcontext)) {
         std::cerr << "view_test passed\n";
         return 0;
     }
--- a/clients/drcachesim/tests/view_test.cpp
+++ b/clients/drcachesim/tests/view_test.cpp
@@ -586,91 +586,11 @@ run_chunk_tests(void *drcontext)
     return run_single_thread_chunk_test(drcontext) && run_serial_chunk_test(drcontext);
 }
 
-/* Test view tool on a OFFLINE_FILE_TYPE_ARCH_REGDEPS trace.
- * The trace hardcoded in entries is X64, so we only test on X64 architectures here.
- */
-bool
-run_regdeps_test(void *drcontext)
-{
-#ifdef X64
-    const memref_tid_t t1 = 3;
-    std::vector<memref_tid_t> tids = { t1 };
-    constexpr addr_t PC = 0x00007f6fdd3ec360;
-    constexpr addr_t PC_vpmovmskb = 0x00007f6fdc76cb35;
-    constexpr addr_t PC_vmovdqu = 0x00007f6fdc76cb2d;
-    constexpr addr_t ENCODING_REGDEPS_ISA = 0x0006090600010011;
-    constexpr addr_t ENCODING_REGDEPS_ISA_vpmovmskb = 0x004e032100004011;
-    constexpr addr_t ENCODING_REGDEPS_ISA_vmovdqu = 0x0009492100004811;
-    std::vector<std::vector<trace_entry_t>> entries = { {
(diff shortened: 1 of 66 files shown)


core/ir/disassemble_shared.c
```diff
@@ -1164,6 +1224,48 @@ sign_extend_immed(instr_t *instr, int srcnum, opnd_t *src)
     }
 }
 
+#define MAX_LENGTH_ALL_CATEGORIES_AS_STRING 57
```
edeiana, line 1227: Perhaps this and the following two functions should be moved to a new file: core/ir/isa_regdeps/disassemble.c.

edeiana:
There are 2 tests consistently failing in ci-x86 / x86-64: code_api|tool.drcacheoff.skip and code_api|tool.drcacheoff.skip2.
The cause is that the header of the trace is being removed and that is causing the dcontext ISA mode to be set to DR_ISA_REGDEPS (on a, supposedly, x86 trace).
I need to investigate why it's happening.

core/ir/disassemble_shared.c
```diff
@@ -749,9 +751,15 @@ internal_opnd_disassemble(char *buf, size_t bufsz, size_t *sofar DR_PARAM_INOUT,
...
         case PC_kind:
         case FAR_PC_kind: break;
-        case REG_kind:
-            if (!opnd_is_reg_partial(opnd))
+        case REG_kind: {
+            /* We always want to print the operation size for DR_ISA_REGDEPS
+             * instructions.
+             */
+            bool is_global_isa_mode_synthetic =
+                dr_get_isa_mode(get_thread_private_dcontext()) == DR_ISA_REGDEPS;
```
derekbruening, line 759: Hmm, this overhead may well show up and slow down disassembly.  Don't we already have a dcontext passed in?  Use that instead of calling get_thread_*.  Even then, can we avoid this extra work on every single register by setting use_size_sfx differently in some way?  Hmm.

So for the operation size we have just one size for the largest source size, right?  Should we print that next to the categories instead of on every register?  It might be misleading on every register since that's not what the size means.

core/ir/disassemble_shared.c
```diff
@@ -809,6 +817,48 @@ print_extra_bytes_to_file(file_t outfile, byte *pc, byte *next_pc, int extra_sz,
...
+        extra_sz = 0;
+    for (i = 0; i < sz; i++)
+        print_to_buffer(buf, bufsz, sofar, " %02x", *(pc + i));
+    for (i = sz; i < REGDEPS_BYTES_PER_LINE; i++)
+        print_to_buffer(buf, bufsz, sofar, "   ");
+    print_to_buffer(buf, bufsz, sofar, " ");
+    return extra_sz;
+}
+
+void
```
derekbruening, line 845: static

core/ir/disassemble_shared.c
```diff
@@ -809,6 +817,48 @@ print_extra_bytes_to_file(file_t outfile, byte *pc, byte *next_pc, int extra_sz,
     os_write(outfile, buf, sofar);
 }
 
+/* DR_ISA_REGDEPS instruction encodings can be at most 16 bytes,
+ * hence we can have at most 2 lines.
+ */
+#define REGDEPS_BYTES_PER_LINE 8
+
+int
```
derekbruening, line 825: static (don't make global functions without a good reason)

core/ir/disassemble_shared.c
```diff
@@ -809,6 +817,48 @@ print_extra_bytes_to_file(file_t outfile, byte *pc, byte *next_pc, int extra_sz,
...
+{
+    int sz = (int)(next_pc - pc);
+    int i, extra_sz;
+    if (sz > REGDEPS_BYTES_PER_LINE) {
+        extra_sz = sz - REGDEPS_BYTES_PER_LINE;
+        sz = REGDEPS_BYTES_PER_LINE;
+    } else
+        extra_sz = 0;
+    for (i = 0; i < sz; i++)
+        print_to_buffer(buf, bufsz, sofar, " %02x", *(pc + i));
```
derekbruening, line 838: Do we want 1-byte separate bytes?  Every instruction has a 4-byte header: seems more natural to print a 4-byte word for that, right?

core/ir/disassemble_shared.c
```diff
@@ -809,6 +817,48 @@ print_extra_bytes_to_file(file_t outfile, byte *pc, byte *next_pc, int extra_sz,
     os_write(outfile, buf, sofar);
 }
 
+/* DR_ISA_REGDEPS instruction encodings can be at most 16 bytes,
+ * hence we can have at most 2 lines.
+ */
+#define REGDEPS_BYTES_PER_LINE 8
+
+int
+print_regdeps_encoding_bytes_to_buffer(char *buf, size_t bufsz,
```
derekbruening, line 826: Should these live in ir/isa_regdeps/ instead of here

core/ir/disassemble_shared.c
derekbruening: +1, see comment above too.

core/ir/disassemble_shared.c
```diff
@@ -1164,6 +1224,48 @@ sign_extend_immed(instr_t *instr, int srcnum, opnd_t *src)
...
+
+    /* We consider 0x80000000 enough to be future proof when adding new categories.
+     */
+    const uint max_mask = 0x80000000;
+    uint start_index = 0;
+    for (uint mask = 0x1; mask <= max_mask; mask <<= 1) {
+        if (TESTANY(mask, category)) {
+            const char *category_name = instr_get_category_name(mask);
+            size_t char_to_copy = strlen(category_name);
+            strncpy(category_names + start_index, category_name, char_to_copy);
```
derekbruening, line 1251: This should be able to use print_to_buffer() which handles all the updating of appending to a buffer.

core/ir/disassemble_shared.c
```diff
@@ -1164,6 +1224,48 @@ sign_extend_immed(instr_t *instr, int srcnum, opnd_t *src)
...
 
+#define MAX_LENGTH_ALL_CATEGORIES_AS_STRING 57
+
+/* Returns a space-separated string representation of the list of categories an
+ * instruction belongs to.
+ */
+void
+get_category_names(uint category, char category_names[], size_t category_names_size)
+{
+    memset(category_names, '\0', category_names_size);
```
derekbruening, line 1235: This seems inefficient: we just need a 0 at the end of the copy.  Normally we use standard helpers like print_to_buffer() or snprintf() + NULL_TERMINATE_BUFFER.

core/ir/disassemble_shared.c
```diff
@@ -1164,6 +1224,48 @@ sign_extend_immed(instr_t *instr, int srcnum, opnd_t *src)
...
+    if (category == DR_INSTR_CATEGORY_UNCATEGORIZED) {
+        const char *category_name = instr_get_category_name(category);
+        size_t char_to_copy = strlen(category_name);
+        strncpy(category_names, category_name, char_to_copy);
```
derekbruening, line 1239: This is not checking category_names_size; nor is it following https://dynamorio.org/page_code_content.html#autotoc_md109 which says "Always null-terminate buffers after calling `strncpy` or `snprintf`: for statically-sized buffers, use  NULL_TERMINATE_BUFFER()."

core/ir/disassemble_shared.c
derekbruening, line 759: I thought the mode of a decoded instruction was stored in the instr_t and not in the dcontext.  Maybe the instr_t needs to be passed in here.

clients/drcachesim/tools/view.cpp
```diff
@@ -50,6 +50,7 @@
 
 #include "analysis_tool.h"
 #include "dr_api.h"
+#include "dr_ir_encode.h"
```
derekbruening, line 53: Mixed feelings about clang-tidy's checks for headers which seem to go too far and do not allow how the DR headers were designed with dr_api.h including the others (this is also how some system headers work, along with other API headers most notably windows.h).  IMHO this include should not be necessary as it goes against the recommended usage in the DR docs.  We may be able to find a way to configure clang-tidy.

clients/drcachesim/tests/view_test.cpp
```diff
@@ -586,11 +586,63 @@ run_chunk_tests(void *drcontext)
...
+        std::string(R"DELIM(           1           0:           3 <marker: version 3>
+           2           0:           3 <marker: filetype 0x20e00>
+           3           0:           3 <marker: cache line size 64>
+           4           0:           3 <marker: chunk instruction count 2>
+           5           0:           3 <marker: timestamp 1002>
+           6           0:           3 <marker: tid 3 on core 2>
+           7           1:           3 ifetch       )DELIM") +
+        std::string(R"DELIM(3 byte(s) @ 0x00007f6fdd3ec360 )DELIM") +
+        std::string(
+            R"DELIM(11 00 01 00 06 09 06 00 move   %rv4[8byte] -> %rv7[8byte]
```
derekbruening, line 627: Could we have more examples: what do loads/stores with memops look like; something with differing real sizes like SIMD with k mask and GPR in there too, which should really highlight whether sizes on registers are going to work (I suspect we can easily see a GPR with an impossible SIMD-sized size -- hence the suggestion to move these sizes to a single size after the categories or sthg like that.)

derekbruening, line 627: Probably there should be core IR tests independent of this drmemtrace view tool (could be a separate PR for the core IR, split from the drmemtrace view tool) in suite/test/api/.

core/ir/disassemble_shared.c
edeiana: Done.

core/ir/disassemble_shared.c
edeiana: This check has been removed, I set `use_size_sfx = 0` for regdeps instructions and print only one size (the operation) after the categories if it's not 0 (e.g., `branch` does not have an operation size).

core/ir/disassemble_shared.c
edeiana: Moved to core/ir/isa_regdeps/disassemble.c.

core/ir/disassemble_shared.c
edeiana: Yes, 4 byte words makes more sense. Done.

core/ir/disassemble_shared.c
edeiana: Moved to core/ir/isa_regdeps/disassemble.c.

core/ir/disassemble_shared.c
edeiana: Using `print_to_buffer()`. Done.

core/ir/disassemble_shared.c
edeiana: Using `print_to_buffer()`, which takes care of that. Done.

core/ir/disassemble_shared.c
edeiana: Using `print_to_buffer()`, which takes care of that. Done.

core/ir/disassemble_shared.c
edeiana: Removed the need for this check altogether.

clients/drcachesim/tools/view.cpp
edeiana: Yes, clang-tidy is a bit strict. Removed.

clients/drcachesim/tests/view_test.cpp
edeiana: Added load with from vector register %ymm to a GPR and a SIMD instruction with k mask and GPR.
Operation size is not a single value after the category names, instead of being printed for all register operands.

edeiana: Ok, will add tests in suite/test/api/dis* in a separate PR.

derekbruening: What I meant was one PR for core disassembly + tests, and a separate PR for the logically separate drmemtrace view tool then using the new disasm.  Changes to the core should have their own tests.  The drmemtrace tool is a logically separate piece.

clients/drcachesim/tests/view_test.cpp
```diff
@@ -607,25 +611,49 @@ run_regdeps_test(void *drcontext)
...
+           4           0:           3 <marker: chunk instruction count 3>
            5           0:           3 <marker: timestamp 1002>
            6           0:           3 <marker: tid 3 on core 2>
            7           1:           3 ifetch       )DELIM") +
         std::string(R"DELIM(3 byte(s) @ 0x00007f6fdd3ec360 )DELIM") +
+        std::string(R"DELIM(00010011 move [8byte]       %rv4 -> %rv7
+)DELIM") +
+        std::string(R"DELIM(           7           1:           3)DELIM") +
+        std::string(
+            R"DELIM(                                             00060906
```
derekbruening, line 638: I'm finding this quite difficult to read when it split over all these lines with the DELIM in there.  It is hard to figure out what this "006*" is supposed to be, sitting here by itself with tons of spaces all around.  I think it would be best to have one matching line per output line even if it's long.

clients/drcachesim/tools/view.cpp
```diff
@@ -573,14 +572,16 @@ view_t::parallel_shard_memref(void *shard_data, const memref_t &memref)
          * disassemble DR_ISA_REGDEPS instructions.
          */
         dr_isa_mode_t old_isa_mode = dr_get_isa_mode(dcontext_.dcontext);
-        if (TESTANY(OFFLINE_FILE_TYPE_ARCH_REGDEPS, filetype_)) {
+        bool is_regdeps_trace =
+            filetype_ != -1 && TESTANY(OFFLINE_FILE_TYPE_ARCH_REGDEPS, filetype_);
```
derekbruening, line 576: Change the init value to 0?  Not ideal to have to do this extra check.

clients/drcachesim/tools/view.cpp
```diff
@@ -604,13 +605,16 @@ view_t::parallel_shard_memref(void *shard_data, const memref_t &memref)
         disasm.insert(newline, str.str());
     }
     // Put our prefix on raw byte spillover, and skip the other columns.
-    newline = disasm.find('\n');
-    if (newline != std::string::npos && newline < disasm.size() - 1) {
+    size_t pos = 0;
+    newline = disasm.find('\n', pos);
+    while (newline != std::string::npos && newline < disasm.size() - 1) {
```
derekbruening, line 610: This change goes away if it all fits on one or 2 lines?

core/ir/isa_regdeps/disassemble.h
```diff
@@ -0,0 +1,49 @@
...
+ * DAMAGE.
+ */
+
+#ifndef _REGDEPS_DISASSEMBLE_H_
+#define _REGDEPS_DISASSEMBLE_H_ 1
+
+#include "../globals.h"
+
+int
+print_regdeps_encoding_bytes_to_buffer(char *buf, size_t bufsz,
```
derekbruening, line 39: In C with no namespaces, esp where we statically link to the app in some use cases and are prevented from hiding our global symbols, it's good to have a common prefix: so how about starting these global names with the new convention d_r_ or maybe d_r_regdeps_ or disassemble_ or d_r_disassemble_.

core/ir/isa_regdeps/disassemble.c
```diff
@@ -0,0 +1,77 @@
...
+#include "../globals.h"
+#include "disassemble.h"
+
+/* DR_ISA_REGDEPS instruction encodings can be at most 16 bytes and are 4 byte aligned,
+ * hence we can have at most 4 lines of 4 bytes each.
+ */
+#define REGDEPS_BYTES_PER_LINE 4
+
+int
+print_regdeps_encoding_bytes_to_buffer(char *buf, size_t bufsz,
```
derekbruening, line 42: This function does not do what its name implies: it doesn't print all the bytes!  It just print 4 bytes!  Maybe best to rename ("d_r_regdeps_print_encoding_first_line"??), and probably add a comment explaining what it's doing and what it's returning, as that is not obvious.

Update: interesting, I see this parallels print_bytes_to_buffer: that also seems confusing.

core/ir/isa_regdeps/disassemble.c
```diff
@@ -0,0 +1,77 @@
...
+ * LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+#include "../globals.h"
+#include "disassemble.h"
+
+/* DR_ISA_REGDEPS instruction encodings can be at most 16 bytes and are 4 byte aligned,
+ * hence we can have at most 4 lines of 4 bytes each.
```
derekbruening, line 37: Why are we limited to 4 bytes per line?  It doesn't seem great to have a 16-byte encoding with the first line with 4 encoding bytes and the actual disasm and then 3 following lines each with just 4 bytes of more encoding.  Seems better to either put the whole 16 bytes one one line or if that is too much empty space and 90% of the instructions are at <=8 bytes then put 8 bytes on each of 2 lines?

clients/drcachesim/tests/view_test.cpp
edeiana: Oh, sorry about that, I missed the keyword: "split" in your first comment.
I'll split this PR in "disassembly + tests in suite/tests/api" and "view tool + current view_test.cpp on a regdeps trace".

clients/drcachesim/tests/view_test.cpp
edeiana: Initially I did that, but DR wouldn't build because of the 90 char per line limit.
I thought it was a clang-format issue, so I tried:
```
/* clang-format off */
{ code with lines over 90 chars}
/* clang-format on */
```

But it didn't work.
I'll investigate what I need to turn off to have one matching line per output.

clients/drcachesim/tools/view.cpp
edeiana: Yes, it would.

core/ir/isa_regdeps/disassemble.c
edeiana: I thought 4 bytes per line was more similar, in terms of layout, to our description and examples of the encoding in
https://github.com/DynamoRIO/dynamorio/blob/a3411cb0eff4c6b6c1bff2dee37906baf8e7b457/core/ir/isa_regdeps/encoding_common.h#L84
since everything is in 4 byte chunks.

But we are not limited by it, I'm fine with having 8 bytes per line (maybe divided into two 4 byte words?) so we can stay within 2 lines instead of 4.

core/ir/isa_regdeps/disassemble.h
edeiana: Ah, my bad, I forgot about the new convention we follow.
I'll go with: "d_r_disassemble_regdeps_".

clients/drcachesim/tools/view.cpp
edeiana: But 0 is `OFFLINE_FILE_TYPE_DEFAULT`.
I'll check if it's safe to do so, I haven't considered the side-effects of that.

core/ir/isa_regdeps/disassemble.c
edeiana: Right, I followed `print_bytes_to_buffer()`.
Perhaps I should rename both.

clients/drcachesim/tests/view_test.cpp
derekbruening: That `clang-format off` is how to disable clang-format.  There are also vera++ checks but I forget what they cover these days (most added before we had clang-format in place): I do see make/style_checks/scripts/rules/line_length_content.tcl and line_length_content is enabled.  Add this file to the exclusion list in make/style_checks/scripts/rules/line_length_content.tcl?

edeiana: Ah, yes, that should do the trick, thank you Derek!

clients/drcachesim/tests/view_test.cpp
edeiana: Done. This PR is only for core disassembly of regdeps instructions + tests.
The PR for the view tool + test is ready, once this one goes into master.

clients/drcachesim/tests/view_test.cpp
edeiana: Not relevant to this PR anymore.

clients/drcachesim/tools/view.cpp
edeiana: Not relevant to this PR anymore. Fixed in view tool PR.

core/ir/isa_regdeps/disassemble.h
edeiana: Done.
I went with you suggestion in the end: `d_r_regdeps_print_encoding_first_line()` (mine was getting too long).

core/ir/isa_regdeps/disassemble.c
edeiana: Renaming done, also for the `print_[extra_]bytes_to_buffer()` routines.
Now: `d_r_print_encoding_first_line_to_buffer()` and `d_r_print_encoding_second_line_to_buffer()`.

core/ir/isa_regdeps/disassemble.c
edeiana: Done. Went with 2 lines max, 8 bytes each, split in two 4 byte words (if the second/fourth word is present).

suite/tests/api/ir_regdeps.c
```diff
@@ -209,13 +250,16 @@ test_instr_create_encode_decode_synthetic_x86_64(void *dc)
...
     decode(dc, buf, instr);
-    test_instr_encode_decode_synthetic(dc, instr);
+    const char *expected_disasm_str_vdpbf16ps_mask =
+        " 00000031 65646640 uncategorized [64byte]       %rv98 %rv99 %rv102 -> %rv100\n "
+        "00000068";
+    test_instr_encode_decode_synthetic(dc, instr, expected_disasm_str_vdpbf16ps_mask);
 }
 #endif
 
 #ifdef ARM
```
edeiana, line 260: Both ARM and RISCV64 tests build but are not being run (looking at Github tests: "ci-aarchxx-cross / arm-cross-compile" and "ci-riscv64 / riscv64-cross-compile" respectively).
Are my define wrong? (But I believe those are what's used in other tests as well)
Or am I doing something wrong in CMakeLists.txt that I'm not seeing?

https://github.com/DynamoRIO/dynamorio/blob/a559edc0b4c6a2ac4c2c94702a6a7bf87e408533/suite/tests/CMakeLists.txt#L2040

core/ir/isa_regdeps/disassemble.c
derekbruening: SGTM

core/ir/isa_regdeps/disassemble.c
derekbruening: SGTM

suite/tests/api/ir_regdeps.c
derekbruening: The cross runs only execute a short list of tests known to succeed under QEMU.  Look for the RUNS_ON_QEMU test lists at the bottom of the file.

suite/tests/api/ir_regdeps.c
```diff
@@ -136,62 +160,79 @@ test_instr_create_encode_decode_synthetic_x86_64(void *dc)
     instr_encode(dc, instr, buf);
     instr_reset(dc, instr);
     decode(dc, buf, instr);
-    test_instr_encode_decode_synthetic(dc, instr);
+    const char *expected_disasm_str_push =
+        " 00001021 26060606 store [8byte]       %rv4 %rv36 -> %rv4\n";
```
derekbruening, line 164: (Style looks good to me.)

derekbruening (approved):

suite/tests/api/ir_regdeps.c
edeiana: Ah, that's what I was missing, thanks.

clients/drcachesim/tools/view.cpp
edeiana: Not relevant to this PR anymore.

suite/tests/api/ir_regdeps.c
edeiana: I enabled the `api.ir_regdeps` test for APPLE (still runs on x86_64, so no changes needed) and RISCV64 (categories there are all "uncategorized", but better than nothing).
Couldn't enable it for ARM, as I get: "/lib/ld-linux-armhf.so.3: No such file or directory", so that's disabled (added a FIXME comment about it, if we want to enable it in the future).

PR #6801, opened 2024-05-09
i#6426 sched stats: Add histogram of instrs per switch

derekbruening:
Adds a very simple histogram with fixed bin sizes recording the distribution of instructions per context switch in the schedule_stats tool.

Updates two tests to confirm the histogram is produced.

This feature is structured to allow replacing this simple histogram with a more sophisticated version in subclasses.

Issue: #6426

### DIFF
--- a/clients/drcachesim/tools/schedule_stats.cpp
+++ b/clients/drcachesim/tools/schedule_stats.cpp
@@ -174,13 +174,46 @@ schedule_stats_t::update_state_time(per_shard_t *shard, state_t state)
     return true;
 }
 
+void
+schedule_stats_t::record_context_switch(per_shard_t *shard, int64_t tid, int64_t input_id,
+                                        int64_t letter_ord)
+{
+    // We convert to letters which only works well for <=26 inputs.
+    if (!shard->thread_sequence.empty()) {
+        ++shard->counters.total_switches;
+        if (shard->saw_syscall || shard->saw_exit)
+            ++shard->counters.voluntary_switches;
+        if (shard->direct_switch_target == tid)
+            ++shard->counters.direct_switches;
+        uint64_t instr_delta = shard->counters.instrs - shard->switch_start_instrs;
+        shard->counters.instrs_per_switch->add(instr_delta);
+        shard->switch_start_instrs = shard->counters.instrs;
+    }
+    shard->thread_sequence +=
+        THREAD_LETTER_INITIAL_START + static_cast<char>(letter_ord % 26);
+    shard->cur_segment_instrs = 0;
+    if (knob_verbose_ >= 2) {
+        std::ostringstream line;
+        line << "Core #" << std::setw(2) << shard->core << " @" << std::setw(9)
+             << shard->stream->get_record_ordinal() << " refs, " << std::setw(9)
+             << shard->stream->get_instruction_ordinal() << " instrs: input "
+             << std::setw(4) << input_id << " @" << std::setw(9)
+             << shard->stream->get_input_interface()->get_record_ordinal() << " refs, "
+             << std::setw(9)
+             << shard->stream->get_input_interface()->get_instruction_ordinal()
+             << " instrs, time "
+             << std::setw(16)
+             // TODO i#5843: For time quanta, provide some way to get the
+             // latest time and print that here instead of the timestamp?
+             << shard->stream->get_input_interface()->get_last_timestamp()
+             << " == thread " << tid << "\n";
+        std::cerr << line.str();
+    }
+}
+
 bool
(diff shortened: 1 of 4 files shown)


clients/drcachesim/tools/schedule_stats.cpp
```diff
@@ -174,13 +174,46 @@ schedule_stats_t::update_state_time(per_shard_t *shard, state_t state)
...
+            ++shard->counters.voluntary_switches;
+        if (shard->direct_switch_target == tid)
+            ++shard->counters.direct_switches;
+        uint64_t instr_delta = shard->counters.instrs - shard->switch_start_instrs;
+        shard->counters.instrs_per_switch->add(instr_delta);
+        shard->switch_start_instrs = shard->counters.instrs;
+    }
+    shard->thread_sequence +=
+        THREAD_LETTER_INITIAL_START + static_cast<char>(letter_ord % 26);
+    shard->cur_segment_instrs = 0;
```
abhinav92003, line 194: Please add a comment by the declaration of cur_segment_instrs that defines what a "segment" is for this tool. I see it is reset on a new idle or wait, and when we reach knob_print_every_, and when we see a context switch.

abhinav92003 (approved):

clients/drcachesim/tools/schedule_stats.h
```diff
@@ -73,9 +75,51 @@ class schedule_stats_t : public analysis_tool_t {
     std::string
     parallel_shard_error(void *shard_data) override;
 
+    // Simple binning histogram for instrs-per-switch distribution.
+    class histogram_t {
```
abhinav92003, line 79: > This feature is structured to allow replacing this simple histogram with a more sophisticated version in subclasses.

From other proposed internal changes, I see that the plan is to create a derived class that overrides the various histogram_t APIs. But that may silently break things if histogram_t implements some new public API which is not overridden by the derived class.

It would be better to have a virtual base-class that can be implemented by both histogram_t and other more-sophisticated implementations. This way, if there's some new public API, that will show up as a clear build failure up-front.

abhinav92003 (changes requested):

clients/drcachesim/tools/schedule_stats.h
abhinav92003: > virtual base-class

I mean an abstract base-class with pure virtual functions.

clients/drcachesim/tools/schedule_stats.h
```diff
@@ -112,6 +157,7 @@ class schedule_stats_t : public analysis_tool_t {
         uint64_t cpu_microseconds = 0;
         uint64_t wait_microseconds = 0;
         std::unordered_set<memref_tid_t> threads;
+        std::unique_ptr<histogram_t> instrs_per_switch;
```
abhinav92003, line 160: If a user wants a more sophisticated histogram_t, I guess they'd just over-write this field with their other implementation? Cleanup should just work since this is a unique_ptr.

I suppose a nicer way to allow substituting a more sophisticated histogram_t would be to have a virtual create_histogram function that can be overridden by classes derived from counters_t. But that does create more complexity of creating a sub-class for counters_t.

Fine either way.

clients/drcachesim/tools/schedule_stats.cpp
derekbruening: Done.

clients/drcachesim/tools/schedule_stats.h
derekbruening: Done.

clients/drcachesim/tools/schedule_stats.h
derekbruening: I was thinking there must be a better way but with the counters_t extra layer it seems like any approach has a little complexity.  Going to go with what's there now.

abhinav92003 (approved):

clients/drcachesim/tools/schedule_stats.h
```diff
@@ -73,9 +75,63 @@ class schedule_stats_t : public analysis_tool_t {
...
+            // For now with relatively known data ranges we just stick
+            // with unchanging bin sizes.
+            uint64_t bin = value - (value % kInitialBinSize);
+            ++bin2count_[bin];
+        }
+
+        void
+        merge(const histogram_interface_t *rhs) override
+        {
+            const histogram_t *rhs_hist = reinterpret_cast<const histogram_t *>(rhs);
```
abhinav92003, line 108: Isn't a dynamic_cast more suitable?

derekbruening: Agreed, done.

derekbruening:
x32 failure is type_is_instr #3320

PR #6803, opened 2024-05-09
i#3544 RV64: Implement out-of-line clean calls

ksco:
Implements out-of-line clean calls on RISCV64, the implementation is similar to AArch64. Also enables out-of-line by default for all clean calls. With this patch, the previous trick of reducing `-max_bb_instrs` is no longer needed.

Issue: #3544


derekbruening (approved):
Good to see the max_bb_instrs doesn't have to be so low anymore!

PR #6805, opened 2024-05-09
i#6662 virtual regs: bug fix

edeiana:
Fixes a size mismatch between dr_reg_fixer[] and d_r_reg_id_to_virtual[] maps in aarch64.
Adds a check in encode_debug_checks() (in core/ir/${ARCH}/encode.c) for all architecures.

Issue: #6662, #3544, #1569

### DIFF
--- a/core/ir/aarch64/encode.c
+++ b/core/ir/aarch64/encode.c
@@ -205,7 +205,7 @@ const reg_id_t d_r_reg_id_to_virtual[] = {
     DR_REG_NULL, /* DR_REG_NULL */
     DR_REG_NULL, /* DR_REG_NULL */
 
-#define VIRTUAL_XREGS
+#define VIRTUAL_XREGS                                                                \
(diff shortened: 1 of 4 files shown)


core/ir/aarch64/encode.c
```diff
@@ -205,7 +205,7 @@ const reg_id_t d_r_reg_id_to_virtual[] = {
     DR_REG_NULL, /* DR_REG_NULL */
     DR_REG_NULL, /* DR_REG_NULL */
 
-#define VIRTUAL_XREGS
+#define VIRTUAL_XREGS                                                                \
```
edeiana, line 208: I can't believe I spent a full day chasing weird bugs for two '/'.

derekbruening: This happens to all of us at times :)

derekbruening (approved):

core/ir/aarch64/encode.c
edeiana: :)

PR #6809, opened 2024-05-13
i#6662 encodings2regdeps: data race bug fix

edeiana:
Fixes a data race due to multiple dr_standalone_init() done in parallel (per shard) by encodings2regdeps_filter_t.
dcontext is now initialized one time by record_filter_t and passed to its filters through the record_filter_info_t interface.

Avoids a data race in opcode_mix where all threads set the dcontext isa_mode to DR_ISA_REGDEPS for regdeps input traces.

Issue #6662 #6812

### DIFF
--- a/clients/drcachesim/tests/record_filter-offline.templatex
+++ b/clients/drcachesim/tests/record_filter-offline.templatex
@@ -5,7 +5,9 @@ Exiting process after .* references.
 Hello, world!
 #endif
 Trace invariant checks passed
+#ifndef WINDOWS
 <unable to determine lib path for cross-arch execve>
+#endif
 Output .* entries from .* entries.
 Done!
 Trace invariant checks passed
--- a/clients/drcachesim/tests/record_filter-offline.templatex
+++ b/clients/drcachesim/tests/record_filter-offline.templatex
@@ -5,6 +5,7 @@ Exiting process after .* references.
 Hello, world!
 #endif
 Trace invariant checks passed
+<unable to determine lib path for cross-arch execve>
 Output .* entries from .* entries.
 Done!
 Trace invariant checks passed
--- a/clients/drcachesim/tools/filter/encodings2regdeps_filter.h
+++ b/clients/drcachesim/tools/filter/encodings2regdeps_filter.h
@@ -67,7 +67,6 @@ public:
     parallel_shard_init(memtrace_stream_t *shard_stream,
                         bool partial_trace_filter) override
     {
-        dcontext_.dcontext = dr_standalone_init();
         return nullptr;
     }
 
@@ -77,6 +76,7 @@ public:
         record_filter_t::record_filter_info_t &record_filter_info) override
     {
         std::vector<trace_entry_t> *last_encoding = record_filter_info.last_encoding;
+        void *dcontext = record_filter_info.dcontext;
 
         /* Modify file_type to regdeps ISA, removing the real ISA of the input trace.
          */
@@ -117,10 +117,10 @@ public:
             /* Genenerate the real ISA instr_t by decoding the encoding bytes.
              */
             instr_t instr;
-            instr_init(dcontext_.dcontext, &instr);
-            app_pc next_pc = decode_from_copy(dcontext_.dcontext, encoding, pc, &instr);
+            instr_init(dcontext, &instr);
+            app_pc next_pc = decode_from_copy(dcontext, encoding, pc, &instr);
             if (next_pc == NULL || !instr_valid(&instr)) {
-                instr_free(dcontext_.dcontext, &instr);
+                instr_free(dcontext, &instr);
                 error_string_ =
                     "Failed to decode instruction " + to_hex_string(entry.addr);
                 return false;
@@ -129,9 +129,9 @@ public:
             /* Convert the real ISA instr_t into a regdeps ISA instr_t.
              */
             instr_t instr_regdeps;
-            instr_init(dcontext_.dcontext, &instr_regdeps);
-            instr_convert_to_isa_regdeps(dcontext_.dcontext, &instr, &instr_regdeps);
-            instr_free(dcontext_.dcontext, &instr);
+            instr_init(dcontext, &instr_regdeps);
+            instr_convert_to_isa_regdeps(dcontext, &instr, &instr_regdeps);
+            instr_free(dcontext, &instr);
 
             /* Obtain regdeps ISA instr_t encoding bytes.
              */
@@ -139,8 +139,8 @@ public:
                 encoding_regdeps[REGDEPS_MAX_ENCODING_LENGTH];
             memset(encoding_regdeps, 0, sizeof(encoding_regdeps));
             app_pc next_pc_regdeps =
-                instr_encode(dcontext_.dcontext, &instr_regdeps, encoding_regdeps);
-            instr_free(dcontext_.dcontext, &instr_regdeps);
+                instr_encode(dcontext, &instr_regdeps, encoding_regdeps);
+            instr_free(dcontext, &instr_regdeps);
             if (next_pc_regdeps == NULL) {
                 error_string_ =
                     "Failed to encode regdeps instruction " + to_hex_string(entry.addr);
@@ -190,19 +190,6 @@ public:
         filetype |= OFFLINE_FILE_TYPE_ARCH_REGDEPS;
         return filetype;
     }
-
-private:
-    struct dcontext_cleanup_last_t {
-    public:
-        ~dcontext_cleanup_last_t()
-        {
-            if (dcontext != nullptr)
-                dr_standalone_exit();
-        }
-        void *dcontext = nullptr;
-    };
-
-    dcontext_cleanup_last_t dcontext_;
 };
 
 } // namespace drmemtrace
--- a/clients/drcachesim/tools/filter/record_filter.cpp
+++ b/clients/drcachesim/tools/filter/record_filter.cpp
@@ -347,6 +347,13 @@ record_filter_t::open_new_chunk(per_shard_t *shard)
     return "";
 }
 
+std::string
+record_filter_t::initialize_stream(memtrace_stream_t *serial_stream)
+{
+    dcontext_.dcontext = dr_standalone_init();
+    return "";
+}
+
 void *
 record_filter_t::parallel_shard_init_stream(int shard_index, void *worker_data,
                                             memtrace_stream_t *shard_stream)
@@ -387,6 +394,7 @@ record_filter_t::parallel_shard_init_stream(int shard_index, void *worker_data,
         }
     }
     per_shard->record_filter_info.last_encoding = &per_shard->last_encoding;
+    per_shard->record_filter_info.dcontext = dcontext_.dcontext;
     std::lock_guard<std::mutex> guard(shard_map_mutex_);
     shard_map_[shard_index] = per_shard;
     return reinterpret_cast<void *>(per_shard);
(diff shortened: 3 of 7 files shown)


clients/drcachesim/tools/opcode_mix.cpp
```diff
@@ -170,10 +170,17 @@ opcode_mix_t::parallel_shard_memref(void *shard_data, const memref_t &memref)
...
+         * to the correct synthetic ISA (i.e., DR_ISA_REGDEPS). Because isa_mode in
+         * dcontext is a global resource, we guard its access to avoid data races
+         * (even though this is a benign data race, as all threads are writing the same
+         * isa_mode value).
          */
-        if (TESTANY(OFFLINE_FILE_TYPE_ARCH_REGDEPS, memref.marker.marker_value))
-            dr_set_isa_mode(dcontext_.dcontext, DR_ISA_REGDEPS, nullptr);
+        if (TESTANY(OFFLINE_FILE_TYPE_ARCH_REGDEPS, memref.marker.marker_value)) {
+            std::lock_guard<std::mutex> guard(dcontext_mutex_);
+            dr_isa_mode_t isa_mode = dr_get_isa_mode(dcontext_.dcontext);
```
edeiana, line 180: Is it always guaranteed that a shard has a TRACE_MARKER_TYPE_FILETYPE marker before all memref instructions?
Otherwise this won't work and I'll have to guard the whole `decode_from_copy()` routine below (which is where dcontext isa_mode is read).

derekbruening: If there is a filetype, it is in the header before any instructions (though maybe that has no invariant check and maybe is not written down...it would be ok to fail with a fatal error there though).  You can use the stream's get_filetype() which caches the filetype once seen if that helps.

clients/drcachesim/tools/filter/encodings2regdeps_filter.h
```diff
@@ -67,7 +67,6 @@ class encodings2regdeps_filter_t : public record_filter_t::record_filter_func_t
     parallel_shard_init(memtrace_stream_t *shard_stream,
                         bool partial_trace_filter) override
     {
-        dcontext_.dcontext = dr_standalone_init();
```
abhinav92003, line 70: I guess the multiple dr_standalone_init calls was not an issue right? Just the data race on dcontext_.dcontext?

If we implement the multi-dcontext solution (https://github.com/DynamoRIO/dynamorio/issues/1595#issuecomment-1981758848), we would then still want a dr_standalone_init in parallel_shard_init right?

So does that mean for now we only need to ensure we write the result of dr_standalone_init to a shard-local var?

edeiana: > I guess the multiple dr_standalone_init calls was not an issue right? Just the data race on dcontext_.dcontext?

Unfortunately no. Multiple `dr_standalone_init()` in parallel can still be a problem because of the global `bool standalone_library`:
https://github.com/DynamoRIO/dynamorio/blob/336e2a2d9f1546b7b098507b2d2094684df471f9/core/dynamo.c#L145

In this specific case, `dr_standalone_init()` writes it here
https://github.com/DynamoRIO/dynamorio/blob/336e2a2d9f1546b7b098507b2d2094684df471f9/core/dynamo.c#L903

But another record_filter thread could be already in `parallel_shard_memref()` and read it here (`record_filter_t::parallel_shard_memref() -> encodings2regdeps_filter_t:: parallel_shard_filter() -> instr_init() -> dr_get_isa_mode() -> get_thread_private_dcontext()`):
https://github.com/DynamoRIO/dynamorio/blob/336e2a2d9f1546b7b098507b2d2094684df471f9/core/unix/os.c#L3086

With a single dcontext we solve the data race (at least for encodings2regdeps_filter_t, which is the only user of dcontext in record_filter for now). After the initialization I believe we only read dcontext (e.g., we read its isa_mode when decoding for sure, but we don't change it after initialization). However, I remember that for AARCH32 there might be issues when encoding/decoding instructions (#6692). I guess they just didn't show up yet.

> If we implement the multi-dcontext solution (https://github.com/DynamoRIO/dynamorio/issues/1595#issuecomment-1981758848), we would then still want a dr_standalone_init in parallel_shard_init right?

In that case yes, we could have `dr_standalone_init()` in `parallel_shard_init()` again (`standalone_library` is also mentioned here #6002 although it's not the only "problem").

> So does that mean for now we only need to ensure we write the result of dr_standalone_init to a shard-local var?

For the operations performed by `encodings2regdeps_filter_t`, yes, that is sufficient. But if in the future we have a new filter that uses dcontext in a different way, I suppose it might not (?).

clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -387,6 +394,7 @@ record_filter_t::parallel_shard_init_stream(int shard_index, void *worker_data,
         }
     }
     per_shard->record_filter_info.last_encoding = &per_shard->last_encoding;
+    per_shard->record_filter_info.dcontext = dcontext_.dcontext;
```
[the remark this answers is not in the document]
edeiana: Good point, adding comment.

clients/drcachesim/tests/record_filter-offline.templatex
```diff
@@ -5,6 +5,9 @@ Exiting process after .* references.
 Hello, world!
```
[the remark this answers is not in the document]
edeiana: Correct, it was highlighted by thread-sanitizer, and yes this PR fixes it.

clients/drcachesim/tools/opcode_mix.cpp
edeiana: Ok, failing if we encounter instructions before we have a filetype for now.
Using stream's `get_filetype()` is the right way to do it, but it requires an update of opcode_mix to use `parallel_shard_init_stream()` rather than the now deprecated `parallel_shard_init()`, which does not have a stream.
I think it's outside the scope of this PR (which aims to just fix a data race). I'll add a comment and a new issue about updating opcode_mix.

edeiana: xref #6812.

clients/drcachesim/tools/filter/record_filter.h
```diff
@@ -70,6 +70,16 @@ class record_filter_t : public record_analysis_tool_t {
          * #trace_entry_t, hence the vector.
          */
         std::vector<trace_entry_t> *last_encoding;
+
+        /**
+         * Gives filters access to dcontext_t.
+         */
+        /* xref i#6690 i#1595: dcontext_t is not entirely thread-safe. AArch32 encoding
```
abhinav92003, line 77: Make this part of the doc above. That is: drop the `/` on this line and above.

clients/drcachesim/tools/opcode_mix.cpp
```diff
@@ -170,10 +170,17 @@ opcode_mix_t::parallel_shard_memref(void *shard_data, const memref_t &memref)
             return false;
         }
         /* If we are dealing with a regdeps trace, we need to set the dcontext ISA mode
-         * to the correct synthetic ISA (i.e., DR_ISA_REGDEPS).
+         * to the correct synthetic ISA (i.e., DR_ISA_REGDEPS). Because isa_mode in
+         * dcontext is a global resource, we guard its access to avoid data races
```
abhinav92003, line 174: nit: move the "Because isa_mode ..." comment to above std::lock_guard.

clients/drcachesim/tools/opcode_mix.cpp
```diff
@@ -190,6 +197,21 @@ opcode_mix_t::parallel_shard_memref(void *shard_data, const memref_t &memref)
...
+    /* At this point we start processing memref instructions, so we're past the header of
+     * the trace, which contains the trace filetype. If we didn't encounter a
+     * TRACE_MARKER_TYPE_FILETYPE we return an error and stop processing the trace.
+     * XXX i#6812: we could allow traces that have some shards with no filetype, as long
+     * as there is at least one shard with it, by caching the filetype from shards that
+     * have it and using that one. We can do this using memtrace_stream_t::get_filetype(),
+     * which requires updating opcode_mix to use parallel_shard_init_stream() rather than
+     * the current (and deprecated) parallel_shard_init().
+     */
+    if (shard->filetype == OFFLINE_FILE_TYPE_DEFAULT) {
```
abhinav92003, line 210: Maybe add a comment by the declaration of OFFLINE_FILE_TYPE_DEFAULT that says that we always expect the filetype to contain one of the other enums too (at least the arch e.g.), and so OFFLINE_FILE_TYPE_DEFAULT should never be present alone, and can be used an an error check.

clients/drcachesim/tools/opcode_mix.h
```diff
@@ -199,6 +199,9 @@ class opcode_mix_t : public analysis_tool_t {
     // For serial operation.
     worker_data_t serial_worker_;
     shard_data_t serial_shard_;
+    /* To guard the setting of isa_mode in dcontext.
```
abhinav92003, line 202: Better to use C++ style comments here. https://dynamorio.org/page_code_style.html#autotoc_md126 says "For C++ code, // comments are allowed." but doesn't disallow C-style comments. I wonder if it meant to.

clients/drcachesim/tools/opcode_mix.cpp
```diff
@@ -190,6 +197,21 @@ opcode_mix_t::parallel_shard_memref(void *shard_data, const memref_t &memref)
         memref.data.type != TRACE_TYPE_INSTR_NO_FETCH) {
         return true;
     }
+
+    /* At this point we start processing memref instructions, so we're past the header of
+     * the trace, which contains the trace filetype. If we didn't encounter a
+     * TRACE_MARKER_TYPE_FILETYPE we return an error and stop processing the trace.
+     * XXX i#6812: we could allow traces that have some shards with no filetype, as long
```
abhinav92003, line 204: I agree with updating opcode_mix to use the stream APIs (i#6812). But I don't know whether we want to tolerate "traces that have some shards with no filetype"

clients/drcachesim/tools/filter/encodings2regdeps_filter.h
abhinav92003: > In that case yes, we could have dr_standalone_init() in parallel_shard_init() again 

LGTM on the current PR. If filters want to do their own dr_standalone_init after the multi-dcontext solution is ready, they can always igonre the dcontext provided by the record_filter.

abhinav92003 (approved):

clients/drcachesim/tools/opcode_mix.h
derekbruening: C-style are still useful for inline such as parameters `, /*param_name=*/ true,`.  But for whole-line `//` should be preferred for C++.

clients/drcachesim/tools/opcode_mix.cpp
edeiana: Clarified that we should spend sometime deciding if we want to allow it first. For now we don't.

clients/drcachesim/tools/opcode_mix.h
edeiana: Got it, changed it to `//`.

clients/drcachesim/tools/opcode_mix.cpp
edeiana: Done.

clients/drcachesim/tools/filter/encodings2regdeps_filter.h
edeiana: SGTM.

clients/drcachesim/tools/filter/record_filter.h
edeiana: Done.

PR #6810, opened 2024-05-13
i#3544 RV64 vector part1: Added vector extension encoder/decoder support

ksco:
This is the first part of adding RISC-V vector (RVV) extension support to the core.

RVV is a vector architecture similar to SVE, and its vector length (VLEN) can vary from 64 up to 65536. For more information about RVV, please refer to https://github.com/riscv/riscv-isa-manual.

The code itself is compiled and preliminarily tested on real hardware with RVV (VLEN=256) support by running https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/main/examples/rvv_strlen.c.

isl/v.txt is transformed from https://github.com/riscv/riscv-opcodes/blob/master/rv_v by an off-tree one-time python script.

Follow-up patches will address the following parts with RVV support:
- encoder/decoder unit tests;
- code cache context switch;
- clean-call;
- signal context;
- scatter/gather emulation;
- sample clients and dr$sim;
- and maybe more.

Issue: #3544

### DIFF
--- a/core/ir/riscv64/codec.c
+++ b/core/ir/riscv64/codec.c
@@ -1198,6 +1198,96 @@ decode_icrs1___opnd(dcontext_t *dc, uint32_t inst, int op_sz, byte *pc, byte *or
     return true;
 }
 
+/* Decode the zimm immediate field in vsetivli instruction (V extension):
+ * |31 30|29    20|19  15|14   12|11       7|6      0|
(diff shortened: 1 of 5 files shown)


derekbruening (approved):
I did not check against the specification itself; I assume some kind of testing vs a reference decoder will help find errors.  Mostly I just spot-checked for self-consistency.

Issue #6814, opened 2024-05-16
Stack overflow crash on app signal delivery to gone-native threads


abhinav92003:
DR may attempt to execute the native app signal handler for a thread in the middle of detach at the following points:
1. it is waiting for the detacher thread to wake it up and continue sig_detach (https://github.com/DynamoRIO/dynamorio/blob/a559edc0b4c6a2ac4c2c94702a6a7bf87e408533/core/unix/signal.c#L8598)
2. it has completed sig_detach and reinstated the app sigblocked mask, but DR's signal handler is still installed

In both these cases execute_native_handler puts the native signal frame on top of DR's own signal frame, and notes the potential of a stack overflow 
https://github.com/DynamoRIO/dynamorio/blob/a559edc0b4c6a2ac4c2c94702a6a7bf87e408533/core/unix/signal.c#L3524

In case (1), the stack overflow may happen on DR's own signal stack, whereas in (2) it may happen on the app stack (signal stack or regular). We're seeing this happen on some large apps that deal with frequent profiling signals that use libunwind to get the thread's stack trace (libunwind is known to have high stack usage).

The solution:

(1) do not unblock signals when detaching
(2) when the native signal handler has to be executed using the same stack as the one DR's signal handler is currently on, we can just reuse the same frame and pass control on to the app's signal handler

abhinav92003:
Xref #6830 that fixes another related scenario.

PR #6815, opened 2024-05-16
i#6814: Fix stack overflow on signal delivery to mid-detach thread

abhinav92003:
Fixes two stack overflow scenarios that occur when DR delivers an app signal to
the native signal handler for a thread that is mid-detach.

First case: when a thread is handling the suspend signal and is waiting for the
detacher thread to wake it up and tell it to continue detaching. Currently, DR
unblocks signals before starting the wait. If the signal is delivered at this
point, currently execute_native_handler() incorrectly delivers the signal
to the native handler on DR's own signal stack. To fix this: we now do not
unblock signals during this wait as it complicates native signal delivery,
particularly for the second case described below.

Additionally, for a detaching thread, we now do not explicitly
restore the app's sigblocked mask; DR already restores the mask on the signal
frame, which would be restored automatically when the thread returns from the DR
detach signal handler. This avoids another case where the app may be on DR's
signal stack when the native signal is delivered.

Second case: when the thread has been woken up by the detacher thread, executed
sig_detach, and reinstated the app signal stack (if available). If the signal is
delivered at this point, execute_native_handler() adds a new signal frame on top
of DR's own signal frame on the app stack and invokes the native signal handler. This
ends up taking too much stack space which causes a stack overflow, as observed on
apps with frequent profiling signals that use the stack-intensive libunwind to get a
stack trace for all threads. To fix this: we reuse the same signal frame for
delivering the signal to the native signal handler, when the app doesn't need a
non-RT frame.

The new code is exercised by the existing detach_signal test. Also modified the test
to have some threads that have a very small sigstack, which helps reproduce the
crash originally seen on a real app. (There was already a note in detach_signal test
about using a large sigstack to avoid this stack overflow.)

Tested on an internal app where failures reduce from ~136/4000 to ~1/4000.

Issue: #6814


core/unix/signal.c
```diff
@@ -6693,6 +6700,141 @@ execute_native_handler(dcontext_t *dcontext, int sig, sigframe_rt_t *our_frame)
...
+{
+    // For a signal that arrived when the thread is native and the process is in
+    // the middle of detaching, we should try to reuse the current DR signal
+    // frame to deliver it if possible. This is to avoid using up too much app
+    // stack, which has been seen to cause a stack overflow in the app (i#6814).
+    // If we cannot optimize, we fall back to the regular
+    // execute_native_handler() on returning.
+    //
+    // Summary of relevant events when DR starts detaching:
+    // D: Detacher thread sends suspend signal to all other threads
```
derekbruening, line 6717: nit: Missing trailing `.`

core/unix/signal.c
```diff
@@ -6693,6 +6700,141 @@ execute_native_handler(dcontext_t *dcontext, int sig, sigframe_rt_t *our_frame)
...
+    //    the suspend signal. So at this point Other threads cannot receive
+    //    app signals. Other threads signal to the Detacher thread that they're
+    //    suspended and awaiting wakeup call to continue detaching (sig_detach).
+    // D: Detacher thread sets doing_detach to true. A side effect of this is
+    //    that get_thread_private_dcontext() will return NULL when any Other
+    //    thread tries to get its private dcontext.
+    // D: Detacher thread wakes up each Other thread, telling it to detach.
+    // O: Each Other thread wakes up, does sig_detach which reinstates the app
+    //    signal stack (if available) and the app's blocked signal set, lets the
+    //    Detacher thread know that its done detaching and resumes native
```
derekbruening, line 6729: s/its/it's/

core/unix/signal.c
```diff
@@ -6693,6 +6700,141 @@ execute_native_handler(dcontext_t *dcontext, int sig, sigframe_rt_t *our_frame)
...
+    //    execution. Note that at this point, DR's sigact config is still
+    //    installed (including main_signal_handler). When the Other thread
+    //    return from the suspend signal, signals get unblocked automatically
```
derekbruening, line 6732: s/return/returns/

core/unix/signal.c
```diff
@@ -6693,6 +6700,141 @@ execute_native_handler(dcontext_t *dcontext, int sig, sigframe_rt_t *our_frame)
...
+    //    signal handling configuration blocks all signals except SIGSEGV and
+    //    the suspend signal. So at this point Other threads cannot receive
+    //    app signals. Other threads signal to the Detacher thread that they're
+    //    suspended and awaiting wakeup call to continue detaching (sig_detach).
+    // D: Detacher thread sets doing_detach to true. A side effect of this is
+    //    that get_thread_private_dcontext() will return NULL when any Other
+    //    thread tries to get its private dcontext.
+    // D: Detacher thread wakes up each Other thread, telling it to detach.
+    // O: Each Other thread wakes up, does sig_detach which reinstates the app
+    //    signal stack (if available) and the app's blocked signal set, lets the
```
derekbruening, line 6728: I thought you removed the sigprocmask call

core/unix/signal.c
```diff
@@ -6693,6 +6700,141 @@ execute_native_handler(dcontext_t *dcontext, int sig, sigframe_rt_t *our_frame)
...
+    // Therefore, if we arrive here and find doing_detach to be true and
+    // dcontext to be NULL, we can be certain that the current thread has
+    // completed sig_detach. This means the interrupted context stack (sc->XSP)
+    // must be the app's own regular stack.
+    //
+    // Also, if the thread has a signal stack at all for any signal, we will
+    // be on it (because DR sets SA_ONSTACK for all signals, and that config is
+    // still in effect), regardless of whether the signal stack was registered
+    // for sig. If the signal stack was actually not registered for sig, then we
+    // return true (indicating that DR should continue trying to deliver the
```
derekbruening, line 6748: This contradicts above where it says false means to continue trying.

core/unix/signal.c
```diff
@@ -6693,6 +6700,141 @@ execute_native_handler(dcontext_t *dcontext, int sig, sigframe_rt_t *our_frame)
...
+    IF_DEBUG(int rc =)
+    sigaltstack_syscall(NULL, &app_sigstack);
+    ASSERT(rc == 0);
+    if (APP_HAS_SIGSTACK_EX(app_sigstack)) {
+        has_sigstack = true;
+        // If the app has a sigstack, we must be on it.
+        ASSERT(app_sigstack.ss_sp < (void *)xsp &&
+               (void *)xsp < app_sigstack.ss_sp + app_sigstack.ss_size);
+        // In release build, we bail on this optimization and try to deliver the
+        // signal.
```
derekbruening, line 6765: ?  I don't understand why release vs debug build matters; plus I don't see any release-vs-debug difference in the code.

core/unix/signal.c
```diff
@@ -6693,6 +6700,141 @@ execute_native_handler(dcontext_t *dcontext, int sig, sigframe_rt_t *our_frame)
...
+        return false;
```
derekbruening, line 6766: Why do we give up if the app has an altstack?

core/unix/signal.c
```diff
@@ -6693,6 +6700,141 @@ execute_native_handler(dcontext_t *dcontext, int sig, sigframe_rt_t *our_frame)
...
+    }
+#endif
+
+    if (sigact_struct.handler == (handler_t)SIG_IGN ||
+        sigact_struct.handler == (handler_t)SIG_DFL) {
+        // We handle logs, messages and asserts related to these cases in
+        // execute_native_handler() later.
+        return false;
+    }
+
```
derekbruening, line 6790: Where is the check for the app wanting an RT frame, which we discussed offline?  We can't use the current RT frame if the app expects a non-RT frame.

core/unix/signal.c
```diff
@@ -6693,6 +6700,141 @@ execute_native_handler(dcontext_t *dcontext, int sig, sigframe_rt_t *our_frame)
...
+    sigprocmask_syscall(SIG_SETMASK, &blocked, NULL, sizeof(blocked));
+    void (*asm_jmp_tgt)() = SIGACT_PRIMARY_HANDLER(&sigact_struct);
+    kernel_siginfo_t *siginfo_var = &((sigframe_rt_t *)xsp)->info;
+    kernel_ucontext_t *ucontext_var = &((sigframe_rt_t *)xsp)->uc;
+    asm volatile("mov %[jmp_tgt],  %%rcx\n\t"
+                 "mov     %[sig],  %%rdi\n\t"
+                 "mov %[siginfo],  %%rsi\n\t"
+                 "mov %[ucontext], %%rdx\n\t"
+                 "mov %[target_sp], %%rsp\n\t"
+                 "jmp *%%rcx\n\t"
```
derekbruening, line 6816: Do we need to create this entire alternate method of invoking the app's handler, which is x86-only?  I thought we would use the existing code in execute_native_handler() and just eliminate the copy of a new frame if the conditions of the stack and RT are met.  Isn't that simpler, more cross-arch, and avoids duplicate handling?

derekbruening (changes requested):

core/unix/signal.c
abhinav92003: The app's blocked signal set is reinstated to the signal frame; the following sigreturn would actually restore it. I'll clarify.

core/unix/signal.c
abhinav92003: I initially had the return wrapped inside an if-block with the assert's condition, but it got messed up in refactoring. I'll fix.

core/unix/signal.c
abhinav92003: execute_native_handler() redirects control to the native signal handler by setting up xip to the handler's address and setting up other regs so that the sigreturn from DR's main_signal_handler would cause the app's handler to be executed.

I tried something similar but concluded that we cannot let a sigreturn redirect control for us (since I saw crashes), because we want the app signal handler to operate on same the signal frame and need it to stay around.

derekbruening: If there were something fundamental, that should be explained in comments so it's clear why this is using its own separate seemingly-duplicated attempt to invoke the native handler.  But I would expect using the sigreturn to work, since I don't think the kernel does much there but restore the machine state: so if the restored stack pointer points to the right place it should re-use the frame.

core/unix/os.c
```diff
@@ -2952,10 +2952,16 @@ os_thread_under_dynamo(dcontext_t *dcontext)
...
 void
-os_thread_not_under_dynamo(dcontext_t *dcontext)
+os_thread_not_under_dynamo(dcontext_t *dcontext, bool restore_sigblocked)
 {
     stop_itimer(dcontext);
-    signal_swap_mask(dcontext, true /*to app*/);
+    // The caller may not want to restore the app's sigblocked mask. E.g., when
+    // a thread is in DR's signal handler to handle the detach signal, it can
+    // restore the mask simply by setting it on the signal frame, which is
+    // better.
```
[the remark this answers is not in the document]
abhinav92003: Done

core/unix/signal.c
```diff
@@ -167,17 +167,20 @@ sig_is_alarm_signal(int sig)
  * We do not support the "legacy stack switching" that uses the restorer field
  * as seen in kernel sources.
  */
-#define APP_HAS_SIGSTACK(info) \
-    ((info)->app_sigstack.ss_sp != NULL && (info)->app_sigstack.ss_flags != SS_DISABLE)
+#define APP_HAS_SIGSTACK_EX(app_sigstack) \
+    (app_sigstack.ss_sp != NULL && app_sigstack.ss_flags != SS_DISABLE)
```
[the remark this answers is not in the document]
abhinav92003: Done

core/unix/signal.c
abhinav92003: Done

core/unix/signal.c
abhinav92003: Done

core/unix/signal.c
abhinav92003: Done

core/unix/signal.c
abhinav92003: Oops. Fixed it.

core/unix/signal.c
abhinav92003: Fixed.

core/unix/signal.c
abhinav92003: We want to give up if there's an altstack but we're not on it. Fixed it.

core/unix/signal.c
abhinav92003: I'm not very familiar with the RT distinction. From get_app_frame_size in signal.c I understand IS_RT_FOR_APP is what I need?

core/unix/signal.c
abhinav92003: To do it the sigreturn way, we'd need to clobber xdi, xsi, and xdx in the sigcontext, and yet the same sigcontext will be used for the native signal delivery; this is a transparency issue and may cause problems for the app if the native signal handler wants to examine those values https://github.com/DynamoRIO/dynamorio/blob/6aa787349c98cca0befd5a831b0a2de542c7a9d7/core/unix/signal.c#L6653

core/dynamo.c
```diff
@@ -2882,7 +2882,7 @@ dynamo_thread_not_under_dynamo(dcontext_t *dcontext)
     LOG(THREAD, LOG_ASYNCH, 2, "thread %d not under DR control\n",
```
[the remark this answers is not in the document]
abhinav92003: Done.

core/unix/signal.c
derekbruening: Yes.

core/unix/signal.c
derekbruening: Ah, ok, so need to have the frame's state be the interrupted state yet also set the parameters to the app's handler using the same state.  We could target a trampoline that sets the registers?  I just see a lot of logic in the execute_native_handler() and want to avoid building up a whole parallel routine where it's easy to miss a check: if someone adds a check to execute_native_handler() they'd have to remember to go update this other parallel routine.  That's why it seems worth trying to route everything through the same code.  If a trampline is too hard to set up: could we still share the other code and just differ in the control xfer?

core/unix/signal.c
abhinav92003: That's done already then.

core/unix/signal.c
```diff
@@ -152,12 +152,11 @@ sig_is_alarm_signal(int sig)
...
-#    define IS_RT_FOR_APP(info, sig)                        \
-        IF_X64_ELSE(true,                                   \
-                    ((info)->sighand->action[(sig)] == NULL \
-                         ? true                             \
-                         : (TEST(SA_SIGINFO, (info)->sighand->action[(sig)]->flags))))
+#    define IS_RT_FOR_APP_EX(sigact) \
+        IF_X64_ELSE(true, ((sigact) == NULL ? true : (TEST(SA_SIGINFO, (sigact)->flags))))
+#    define IS_RT_FOR_APP(info, sig) IS_RT_FOR_APP_EX((info)->sighand->action[(sig)])
 #elif defined(MACOS)
+#    define IS_RT_FOR_APP_EX(sigaction) (true)
```
derekbruening, line 159: nit: Varying parameter name: `sigact` for consistency

core/unix/signal.c
```diff
@@ -8607,8 +8778,10 @@ handle_suspend_signal(dcontext_t *dcontext, kernel_siginfo_t *siginfo,
     }
     LOG(THREAD, LOG_ASYNCH, 2, "handle_suspend_signal: awake now\n");
 
-    /* re-block so our exit from main_signal_handler is not interrupted */
-    sigprocmask_syscall(SIG_SETMASK, &prevmask, NULL, sizeof(prevmask));
+    if (unblocked_sigs) {
+        /* re-block so our exit from main_signal_handler is not interrupted */
+        sigprocmask_syscall(SIG_SETMASK, &prevmask, NULL, sizeof(prevmask));
+    }
```
derekbruening, line 8784: I created api.detach_signal to test the native signal delivery path but I see no mention of it (or any automated test) in the PR description.  Please evaluate whether api.detach_signal exercises the various paths of this new code and tweak it to reproduce in at least some runs if not.

core/unix/signal.c
```diff
@@ -5784,7 +5790,7 @@ main_signal_handler_C(byte *xsp)
         /* For can_always_delay[sig] we could just return and drop it, but we
```
derekbruening, line 5790: Please update the comment at line 3529 which currently says:

```
        /* Handle DR's frame already being on the app stack.  For native delivery we
         * could try to re-use this frame, but that doesn't work with plain vs rt.
         * Instead we move below and live with the downsides of a potential stack
         * overflow.
         */
```

core/unix/signal.c
```diff
@@ -6693,6 +6699,154 @@ execute_native_handler(dcontext_t *dcontext, int sig, sigframe_rt_t *our_frame)
...
+    // return false (indicating that DR should continue trying to deliver the
+    // signal).
+    if (dcontext != NULL || !doing_detach)
+        return false;
+
+    bool has_sigstack = false;
+#ifdef HAVE_SIGALTSTACK
+    stack_t app_sigstack;
+    IF_DEBUG(int rc =)
+    sigaltstack_syscall(NULL, &app_sigstack);
```
derekbruening, line 6757: I still think there is too much separation from and in some sense duplication here with execute_native_handler.  execute_native_handler has a lot of checks for different cases; wouldn't the code be clearer if execute_native_handler was the one who invoked this (modified) routine from inside one of those conditionals?  As-is we have a huge comment explaining general detach issues and a bunch of completely separate checks which make it seem like execute_native_handler has to be treated as a black box and we have to explain and extract some of its checks into here, instead of integrating with that existing code and the existing checks.

I haven't gotten into all the check details but I think it might be nicer to:
+ Have the detach explanation be at top level with a banner comment grouping these related functions; or maybe inside execute_native_handler with the general detach up top and specific to this case where that decision is made to avoid sigreturn.
+ Remove the duplicate code here like getting the alt stack and instead use the code in execute_native_handler and have it call out to this routine to do the non-sigreturn transfer.  I'm guessing most of the conditions here are already checked in execute_native_handler and I'm requesting to not have a completely separate set of the same checks.  I just see a bunch of checks and cases inside execute_native_handler and it seems too easy to have these parallel sets of checks diverge and we add an important addition to one but forget the other.

core/unix/signal.c
abhinav92003: Done

core/unix/signal.c
abhinav92003: Don't need the macro anymore.

core/unix/signal.c
abhinav92003: Done

core/unix/signal.c
```diff
@@ -6523,9 +6521,15 @@ execute_handler_from_dispatch(dcontext_t *dcontext, int sig)
     return true;
 }
 
-/* Sends a signal to a currently-native thread.  dcontext can be NULL. */
+/* Sends a signal to a currently-native thread.
+ * dcontext can be NULL.
+ * cur_xsp is the stack pointer seen by the DR signal handler. May be NULL for
```
derekbruening, line 6526: "seen" is confusing: it sounds like the one in the frame, the interrupted sp, as that's what the handler "sees".  Please reword s/seen/initially executed on/ or something similar.

core/unix/signal.c
```diff
@@ -6631,11 +6633,120 @@ execute_native_handler(dcontext_t *dcontext, int sig, sigframe_rt_t *our_frame)
...
+     * must be the app's own stack.
+     *
+     * Also, if the thread has a signal stack at all for any signal, we will
+     * be on it (because DR sets SA_ONSTACK for all signals, and that config is
+     * still in effect), regardless of whether the signal stack was registered
+     * for sig.
+     */
+    bool reuse_cur_frame = false;
+    if (dcontext == NULL && doing_detach) {
+        ASSERT(cur_xsp != NULL);
```
derekbruening, line 6678: Above it says it's ok to pass NULL: maybe reword to more strongly say it's not ok for this case.

core/unix/signal.c
```diff
@@ -6631,11 +6633,120 @@ execute_native_handler(dcontext_t *dcontext, int sig, sigframe_rt_t *our_frame)
...
+            "jmp *%%rcx\n\t"
+            :
+            : [jmp_tgt] "m"(asm_jmp_tgt), [sig] "m"(sig), [siginfo] "m"(siginfo_var),
+              [ucontext] "m"(ucontext_var), [target_sp] "m"(cur_xsp)
+            /* Cannot add "rsp" as listing it in clobber list is deprecated. */
+            : "rdi", "rsi", "rdx", "rcx");
+        ASSERT_NOT_REACHED();
+    }
+#else
+    /* TODO i#6814: Add support for other archs. */
```
derekbruening, line 6743: I believe call_switch_stack() would work here if the handler took just one arg, and give us cross-arch implementations: or maybe it pushes a retaddr after switching. Hmm. Too bad using sigreturn requires a trampoline on the other side with some data passed through.

Will aarch64 be added ASAP?

core/unix/signal.c
derekbruening: Isn't that just because the test has large stack sizes?  It doesn't need any signal code path changes: just needs a tiny stack to repro that part I would assume.  It is probably worth spending a little time making the stack tiny to reproduce this; if that works maybe leave it tiny or parameterize and run some tiny and some regular.

derekbruening (approved):

core/unix/signal.c
abhinav92003: > Yes: call_switch_stack uses blx/blr/CALLC1 which uses call.

Though only the call is a concern (since blx/blr don't push to stack).

abhinav92003: Added the AArch64 version. Was easier to test it when I modified detach_signal to reproduce the issue.

core/unix/signal.c
abhinav92003: Done

core/unix/signal.c
abhinav92003: Lowering the altstack size in detach_signal reproduces this. Thanks for the suggestion.

>  if that works maybe leave it tiny or parameterize and run some tiny and some regular.

Half of the threads have the earlier altstack size and others have the smaller one.

abhinav92003:
detach_signal seems to be failing on Github Actions, but on my local runs (on both x86 and AArch64) it works fine.

derekbruening (approved):

abhinav92003:
> detach_signal seems to be failing on Github Actions, but on my local runs (on both x86 and AArch64) it works fine.

The sigmask to be restored before passing control to the app handler needed to include also the app's native sigmask from the frame. Added that, which fixed the failure.

abhinav92003:
The opcode_mix test is still flaky on x86-32 and needed some reruns (though x86-32 is not blocking submission)

PR #6816, opened 2024-05-17
i#6662 public traces, part 4: view tool

edeiana:
Modifies the view tool to handle OFFLINE_FILE_TYPE_ARCH_REGDEPS traces,
leveraging the disassembly of DR_ISA_REGDEPS instructions.
When visualizing DR_ISA_REGDEPS instructions, the view tool still prints the
instruction length and PC, which for OFFLINE_FILE_TYPE_ARCH_REGDEPS
traces are the same as those in the original trace.
Then, after the PC, the instruction encoding, categories, operation size, and 
registers are printed following the disassembly format of DR_ISA_REGDEPS
instructions (xref: #6799).

DR_ISA_REGDEPS instructions printed by the view tool look as follows:
```
[...] ifetch      10 byte(s) @ 0x00007f86ef03d107 00001931 04020204 load store [4byte]       %rv0 %rv2 %rv36 -> %rv0
[...]                                             00000026
```

We also fix a formatting bug in DR_ISA_REGDEPS instruction disassembly,
where we were missing a new line when the instruction encoding spills into
a second line.

Issue: #6662

### DIFF
--- a/clients/drcachesim/tests/view_test.cpp
+++ b/clients/drcachesim/tests/view_test.cpp
@@ -586,11 +586,91 @@ run_chunk_tests(void *drcontext)
     return run_single_thread_chunk_test(drcontext) && run_serial_chunk_test(drcontext);
 }
 
+/* Test view tool on a OFFLINE_FILE_TYPE_ARCH_REGDEPS trace.
+ * The trace hardcoded in entries is X64, so we only test on X64 architectures here.
+ */
+bool
+run_regdeps_test(void *drcontext)
+{
+#ifdef X64
+    const memref_tid_t t1 = 3;
+    std::vector<memref_tid_t> tids = { t1 };
+    constexpr addr_t PC = 0x00007f6fdd3ec360;
+    constexpr addr_t PC_vpmovmskb = 0x00007f6fdc76cb35;
+    constexpr addr_t PC_vmovdqu = 0x00007f6fdc76cb2d;
+    constexpr addr_t ENCODING_REGDEPS_ISA = 0x0006090600010011;
+    constexpr addr_t ENCODING_REGDEPS_ISA_vpmovmskb = 0x004e032100004011;
+    constexpr addr_t ENCODING_REGDEPS_ISA_vmovdqu = 0x0009492100004811;
+    std::vector<std::vector<trace_entry_t>> entries = { {
+        { TRACE_TYPE_HEADER, 0, { 0x1 } },
+        { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_VERSION, { 3 } },
+        { TRACE_TYPE_MARKER,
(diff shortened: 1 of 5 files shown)


core/ir/isa_regdeps/disassemble.c
```diff
@@ -109,4 +109,8 @@ d_r_regdeps_print_encoding_second_line(char *buf, size_t bufsz,
             buf, bufsz, sofar, " %08x",
             *((uint *)(pc + REGDEPS_BYTES_PER_LINE + REGDEPS_BYTES_PER_WORD)));
     }
+
+    /* Add a new line at the end.
+     */
+    print_to_buffer(buf, bufsz, sofar, "\n");
```
edeiana, line 115: I didn't realize I was missing a new line until I "view tool" printed a whole trace. Fixed.

clients/drcachesim/tests/view_test.cpp
```diff
@@ -586,11 +586,74 @@ run_chunk_tests(void *drcontext)
     return run_single_thread_chunk_test(drcontext) && run_serial_chunk_test(drcontext);
```
derekbruening, line 586: > Specifically, when we disassemble DR_ISA_REGDEPS instructions, we print the
instruction encoding (can span multiple lines), we substitute the opcode with
the categories, we print the operations size (e.g., [4byte]), and then the
virtual register names (e.g., %rv3).

This sounds like this PR is implementing all of that, but it's not.  Please reword to make it clear that's coming from DR's disassembler and is not added by this PR.

clients/drcachesim/tests/view_test.cpp
```diff
@@ -586,11 +586,74 @@ run_chunk_tests(void *drcontext)
...
+bool
+run_regdeps_test(void *drcontext)
+{
+#ifdef X64
+    const memref_tid_t t1 = 3;
+    std::vector<memref_tid_t> tids = { t1 };
+    constexpr addr_t PC = 0x00007f6fdd3ec360;
+    constexpr addr_t PC_vpmovmskb = 0x00007f6fdc76cb35;
+    constexpr addr_t PC_vmovdqu = 0x00007f6fdc76cb2d;
+    constexpr addr_t ENCODING_REGDEPS_ISA = 0x0006090600010011;
```
derekbruening, line 601: This should have a suffix to indicate it's a particular instruction, like the ones below.  Otherwise it looks like it's referring to some generic ISA value.

derekbruening (approved):

clients/drcachesim/tests/view_test.cpp
edeiana: Done. Also added a two line encoding instruction to the test.

clients/drcachesim/tests/view_test.cpp
edeiana: Done.

PR #6819, opened 2024-05-20
i#2626 AArch64 codec: Add FEAT_MTE2 instructions 

jackgallagher-arm:
This patch adds the appropriate macros, tests and codec entries to
decode and encode the following instructions:
```
LDGM    <Xt>, [<Xn|SP>]
STGM    <Xt>, [<Xn|SP>]
STZGM   <Xt>, [<Xn|SP>]
```

As part of adding support for these instructions I spotted that the
existing FEAT_MTE instructions were in the wrong architecture version
files, and in fact all the instructions in the v8.6 files didn't
belong there, so I moved them all to the correct files based on the
architecture version of the features:

BF16 -> v8.2
I8MM -> v8.2
MTE  -> v8.5
MTE2 -> v8.5

Issue: #2626


suite/tests/client-interface/features.dll.c
```diff
@@ -117,7 +117,7 @@ static const struct feature_strings features[] = {
     { "sha512", FEATURE_SHA512 }, { "atomics", FEATURE_LSE }, { "bf16", FEATURE_BF16 },
     { "jscvt", FEATURE_JSCVT },   { "lrcpc", FEATURE_LRCPC }, { "sm3", FEATURE_SM3 },
     { "sm4", FEATURE_SM4 },       { "i8mm", FEATURE_I8MM },   { "rng", FEATURE_RNG },
-    { "fphp", FEATURE_FP16 }
+    { "fphp", FEATURE_FP16 },     { "mte", FEATURE_MTE },
```
AssadHashmi, line 120: Does `mte` in `/proc/cpuinfo`'s `Features` string also cover MTE2?

AssadHashmi:
The commit message should mention moving tests and encodings definitions out from the v8.6 files.

suite/tests/client-interface/features.dll.c
jackgallagher-arm: This should have been `FEATURE_MTE2` actually. The "mte" feature string indicates `FEAT_MTE2` support. Linux doesn't appear to report `FEAT_MTE` separately.

AssadHashmi (approved):

PR #6820, opened 2024-05-21
i#6662 public traces, part 5: func_id_filter_t

edeiana:
Adds a new filter: `func_id_filter_t` to record_filter, which filters TRACE_MARKER_TYPE_FUNC_ markers based on the function ID.

The filter is enabled by `-filter_keep_func_ids` followed by a list of integers that represent the function IDs bound to TRACE_MARKER_TYPE_FUNC_ markers to keep in the trace.
Specifically, whenever we encounter a TRACE_MARKER_TYPE_FUNC_ID marker whose marker value is in the list we set a per-shard flag to indicate that all TRACE_MARKER_TYPE_FUNC_[ID | ARG | RETVAL | RETADDR] markers related to that function ID need to be preserved. We remove the TRACE_MARKER_TYPE_FUNC_ markers related to functions whose ID is not in the list.

This filter can be invoked with:
```
drrun -t drmemtrace -tool record_filter -filter_keep_func_ids 1,2,3,4 -indir path/to/input/trace -outdir path/to/output/trace
```
To preserve TRACE_MARKER_TYPE_FUNC_ markers related to functions with ID: 1, 2, 3, 4, and remove the TRACE_MARKER_TYPE_FUNC_ markers for all other ID values.

We use this filter to preserve markers related to SYS_futex functions in the public release of traces.

Issue #6662


clients/drcachesim/tools/filter/encodings2regdeps_filter.h
```diff
@@ -67,27 +72,78 @@ class encodings2regdeps_filter_t : public record_filter_t::record_filter_func_t
...
+            /* Modify file_type to regdeps ISA. Removes the real ISA of the input
+             * trace and adds OFFLINE_FILE_TYPE_ARCH_REGDEPS.
+             */
+            case TRACE_MARKER_TYPE_FILETYPE: {
                 uint64_t marker_value = static_cast<uint64_t>(entry.addr);
                 marker_value = update_filetype(marker_value);
                 entry.addr = static_cast<addr_t>(marker_value);
+                return true;
+            }
+            /* Output TRACE_MARKER_TYPE_FUNC_[ID | ARG | RETVAL] for SYS_futex syscalls.
```
derekbruening, line 108: This doesn't seem to belong in a filter that transforms the encodings.  I would expect it to be in the type filter or possibly a new filter.

clients/drcachesim/tools/filter/encodings2regdeps_filter.h
```diff
@@ -67,27 +72,78 @@ class encodings2regdeps_filter_t : public record_filter_t::record_filter_func_t
...
+            case TRACE_MARKER_TYPE_FUNC_ID: {
+#ifdef LINUX
+                uintptr_t marker_value = static_cast<uintptr_t>(entry.addr);
+                uintptr_t syscall_base =
+                    static_cast<uintptr_t>(func_trace_t::TRACE_FUNC_ID_SYSCALL_BASE);
+                uintptr_t syscall_num = marker_value - syscall_base;
+                /* Function is a SYS_futex syscall, so we output the
+                 * TRACE_MARKER_TYPE_FUNC_ID marker.
+                 */
+                if (syscall_num >= 0 && syscall_num == SYS_futex) {
```
derekbruening, line 119: Post-processing is not always done on the same system as tracing so comparing to the post-process machine's SYS_futex is not reliable.

Xref #6821 though it would need augmentation to help here.

clients/drcachesim/tools/filter/encodings2regdeps_filter.h
edeiana: You're right, the thing is that this is such a specific type of filtering where the only use-case (at least for now) is the release of public traces.
The type_filter would need major changes to do this type of filtering, as we're not simply listing the TRACE_MARKER_TYPE_ to drop, we're dropping TRACE_MARKER_TYPE_FUNC_ID depending on their value, and on top of that we have "state" to drop other, related markers, such as TRACE_MARKER_TYPE_FUNC_ARG, depending on the value of TRACE_MARKER_TYPE_FUNC_ID. It's probably best not to overcomplicate type_filter.
I think having a new "function_marker_filter" that focuses on TRACE_MARKER_TYPE_FUNC_ only and generalizes this a bit (listing the #function to keep, rather than hardcoding SYS_futex) is the way to go.

derekbruening: SGTM.  If we had #6821 the user could get the syscall nums that were traced; could cmp to the syscall lists for the supported arches and figure out which is futex -- with corner cases of colliding numbers...maybe need an offline discussion on how to solve that.

edeiana: If we leave the "burden" of specifying directly the syscall number for the TRACE_MARKER_TYPE_FUNC_ID value to the user we don't even need #6821 for now (it can be inferred by OFFLINE_FILE_TYPE_ARCH_ of the trace right?).
If we want to be more user-friendly (having something like `--function_marker_filter_func_ids_to_keep=SYS_futex,...` rather than `--function_marker_filter_func_ids_to_keep=202,...` for an OFFLINE_FILE_TYPE_ARCH_X86_64 trace), then we probably need to "synthesize a string name for the syscall" as described in #6821.

clients/drcachesim/tools/filter/encodings2regdeps_filter.h
edeiana: We now take function IDs directly, so we rely on the user to have the correct SYS_futex function ID.

api/docs/release.dox
```diff
@@ -241,6 +241,9 @@ Further non-compatibility-affecting changes include:
    trace analysis tool framework.
  - Added "-t drmemtrace" as the preferred launcher for the drmemtrace/drcachesim
    trace analysis tool framework.
+ - Added func_marker_filter_t filter to #dynamorio::drmemtrace::record_filter_t to filter
+   #dynamorio::drmemtrace::TRACE_MARKER_TYPE_FUNC_ID (and _ARG, _RETVAL, _RETADDR) entries
+   based on the marker value (i.e., the function ID).
```
derekbruening, line 246: Mentioning all the marker types and then saying it filters based on marker value seems confusing to me: the values of the other markers are different.  I would say it filters function markers based on the "function id".

api/docs/release.dox
```diff
@@ -241,6 +241,9 @@ Further non-compatibility-affecting changes include:
    trace analysis tool framework.
```
derekbruening, line 241: > The filter is enabled by -filter_func_markers_by_value followed by a list of integers that represent the function IDs of TRACE_MARKER_TYPE_FUNC_ markers to keep in the trace.

To keep?  That's the opposite of the other -filter options where you specify what to *remove*.

clients/drcachesim/common/options.cpp
```diff
@@ -982,6 +982,13 @@ droption_t<bool> op_encodings2regdeps(
     "This option is for -simulator_type " RECORD_FILTER ". When present, it converts "
     "the encoding of instructions from a real ISA to the DR_ISA_REGDEPS synthetic ISA.");
 
+droption_t<std::string>
+    op_filter_func_markers_by_value(DROPTION_SCOPE_FRONTEND,
```
derekbruening, line 986: See earlier comment: "by_value" sounds like the value of markers such as _ARG or _RETVAL.  I think this should be `-filter_func_markers_by_id` or even better just `-filter_func_ids`.

clients/drcachesim/common/options.cpp
```diff
@@ -982,6 +982,13 @@ droption_t<bool> op_encodings2regdeps(
...
+                                    "filter_func_markers_by_value", "",
+                                    "Comma-separated integers of function IDs to keep.",
```
derekbruening, line 988: Please add what you have above: "This option is for -simulator_type " RECORD_FILTER "".
Actually `s/simulator_type/tool/`; replace above too.

clients/drcachesim/CMakeLists.txt
```diff
@@ -198,6 +198,7 @@ add_exported_library(drmemtrace_record_filter STATIC
   tools/filter/cache_filter.cpp
   tools/filter/type_filter.h
   tools/filter/encodings2regdeps_filter.h
+  tools/filter/func_marker_filter.h
```
derekbruening, line 201: Rename to func_id_filter?

Could survey others to vote on it.

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -427,6 +430,110 @@ test_encodings2regdeps_filter()
     return true;
 }
 
+/* Test SYS_futex preservation.
+ */
+#define SYS_futex_test 202
```
derekbruening, line 435: These could be constexpr ints inside the function.  That avoids affecting subsequent tests down below and keeps it local.

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -427,6 +430,110 @@ test_encodings2regdeps_filter()
...
+#define SYS_fsync_test 74
+static bool
+test_func_marker_filter()
+{
+    constexpr addr_t syscall_base =
```
derekbruening, line 440: The DR style has all caps for constants; below too.

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -427,6 +430,110 @@ test_encodings2regdeps_filter()
...
+        { { TRACE_TYPE_THREAD, 0, { 0x4 } }, true, { true } },
+        { { TRACE_TYPE_PID, 0, { 0x5 } }, true, { true } },
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_CACHE_LINE_SIZE, { 0x6 } },
+          true,
+          { true } },
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_CHUNK_INSTR_COUNT, { 0x3 } },
+          true,
+          { true } },
+
+        /* Chunk 1.
```
derekbruening, line 466: I don't see separate chunks so this comment can be removed.

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -427,6 +430,110 @@ test_encodings2regdeps_filter()
...
+            { OFFLINE_FILE_TYPE_ARCH_X86_64 | OFFLINE_FILE_TYPE_ENCODINGS |
+              OFFLINE_FILE_TYPE_SYSCALL_NUMBERS | OFFLINE_FILE_TYPE_BLOCKING_SYSCALLS } },
+          true,
+          { true } },
+        { { TRACE_TYPE_THREAD, 0, { 0x4 } }, true, { true } },
+        { { TRACE_TYPE_PID, 0, { 0x5 } }, true, { true } },
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_CACHE_LINE_SIZE, { 0x6 } },
+          true,
+          { true } },
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_CHUNK_INSTR_COUNT, { 0x3 } },
```
derekbruening, line 462: This can be removed to simplify the test.

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -427,6 +430,110 @@ test_encodings2regdeps_filter()
...
+          true,
+          { true } },
+
+        /* Chunk 1.
+         */
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_TIMESTAMP, { 0x7 } }, true, { true } },
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_CPU_ID, { 0x8 } }, true, { true } },
+        /* We need at least one instruction with encodings.
```
derekbruening, line 470: We do?

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -427,6 +430,110 @@ test_encodings2regdeps_filter()
...
+         */
+        { { TRACE_TYPE_ENCODING, 3, { ENCODING } }, true, { true } },
+        { { TRACE_TYPE_INSTR, 3, { PC } }, true, { true } },
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_FUNC_ID, { syscall_num_futex } },
+          true,
+          { true } },
+        /* We don't care about the arg values, we just care that they are preserved.
```
derekbruening, line 477: Non-zero values seem to give more confidence on preservation (instead of creating new empty ones or sthg).

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -427,6 +430,110 @@ test_encodings2regdeps_filter()
...
+         * Note: SYS_futex has 6 args.
+         */
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_FUNC_ARG, { 0x0 } }, true, { true } },
```
derekbruening, line 480: Malformed: missing _RETADDR.

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -427,6 +430,110 @@ test_encodings2regdeps_filter()
...
+          { true } },
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_FUNC_RETVAL, { 0x0 } }, true, { true } },
+        /* Test that func_marker_filter_t doesn't output any
+         * TRACE_MARKER_TYPE_FUNC_ for functions that are not SYS_futex.
+         * We use SYS_fsync in this test.
+         */
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_FUNC_ID, { syscall_num_fsync } },
+          true,
+          { false } },
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_FUNC_ARG, { 0x0 } }, true, { false } },
```
derekbruening, line 497: Missing RETADDR.

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -427,6 +430,110 @@ test_encodings2regdeps_filter()
...
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_FUNC_ID, { syscall_num_fsync } },
+          true,
+          { false } },
+        { { TRACE_TYPE_MARKER, TRACE_MARKER_TYPE_FUNC_RETVAL, { 0x0 } },
```
derekbruening, line 501: It may be worth also testing:

+ Nested functions
+ Nested functions where the inner is kept but not the outer; vice versa?

Because you could imagine a state machine not handling nesting well.

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -427,6 +430,110 @@ test_encodings2regdeps_filter()
...
+          true,
+          { false } },
+
+        /* Trace shard footer.
```
derekbruening, line 505: (optional nit: Comment doesn't add much as the type is already FOOTER)

edeiana: Removed.

clients/drcachesim/tests/record_filter_unit_tests.cpp
edeiana: https://github.com/DynamoRIO/dynamorio/pull/6820#discussion_r1626540016

clients/drcachesim/tests/record_filter_unit_tests.cpp
edeiana: Right, using non-zero values.

clients/drcachesim/tests/record_filter_unit_tests.cpp
edeiana: Done.

clients/drcachesim/tests/record_filter_unit_tests.cpp
edeiana: Done.

clients/drcachesim/tests/record_filter_unit_tests.cpp
edeiana: Fixed.

clients/drcachesim/tests/record_filter_unit_tests.cpp
edeiana: Done.

clients/drcachesim/CMakeLists.txt
edeiana: Renamed to "func_id_filter.h" (and "func_id_filter_t" for the class name).

clients/drcachesim/common/options.cpp
edeiana: Done.

clients/drcachesim/common/options.cpp
edeiana: Went with "-filter_keep_func_ids" as discussed.

api/docs/release.dox
edeiana: Fixed.

clients/drcachesim/tests/record_filter_unit_tests.cpp
edeiana: If we only have function-related markers and no instruction, nothing gets output.
I did not investigate in detail why record_filter behaves that way, is it not supposed to happen? I can look into it.

api/docs/release.dox
edeiana: Right, but for our use case we have a lot more function IDs to remove (much more than TRACE_MARKER_TYPE_ enum values, which type_filter operates on), so it's more convenient to list the ones we want to keep.
We could add a second option to turn on func_id_filter_t: `-filter_func_ids` (as opposed to the current `-filter_keep_func_ids`) which would remove the markers related to the listed function IDs (these two options would be mutually exclusive).
But perhaps it makes sense to do so when we have a need for it?

api/docs/release.dox
```diff
@@ -241,6 +241,8 @@ Further non-compatibility-affecting changes include:
...
```
derekbruening, line 241: > Note that we remove the TRACE_MARKER_TYPE_FUNC_ markers related to functions whose ID is not in the list.

This is very important and should not be under "Note": `s/Note that we/We/`.

derekbruening, line 241: > To preserve TRACE_MARKER_TYPE_FUNC_ markers related to functions with ID: 1, 2, 3, and 4.

Add "and remove the FUNC_ markers for all other ID values."

api/docs/release.dox
```diff
@@ -241,6 +241,8 @@ Further non-compatibility-affecting changes include:
...
  - Added "-t drmemtrace" as the preferred launcher for the drmemtrace/drcachesim
    trace analysis tool framework.
+ - Added func_marker_filter_t filter to #dynamorio::drmemtrace::record_filter_t to filter
```
derekbruening, line 244: Use # for a link for the func_marker_filter_t type.  Maybe should mention the CLI option name also (or instead).

clients/drcachesim/common/options.cpp
```diff
@@ -979,9 +979,20 @@ droption_t<std::string>
...
+/* XXX i#6369: we should partition our options by tool. This one should belong to the
+ * record_filter partition. For now we add the filter_ prefix to options that should be
+ * used in conjunction with record_filter.
+ */
+droption_t<std::string>
+    op_filter_func_ids(DROPTION_SCOPE_FRONTEND, "filter_keep_func_ids", "",
+                       "Comma-separated integers of function IDs to keep.",
+                       "This option is for -tool " RECORD_FILTER ". It preserves "
+                       "TRACE_MARKER_TYPE_FUNC_[ID | ARG | RETVAL | RETADDR] markers "
+                       "for the listed function IDs.");
```
derekbruening, line 994: " and removes those belonging to unlisted function IDs."

clients/drcachesim/tests/record_filter_unit_tests.cpp
derekbruening: I think there are multiple cases like this due to the delayed state processing inherent in stream-oriented state machines.  Just augment the comment to make it clear *why* (meaning *who* needs this): as-is it sounds like a weird requirement of this func id filter itself.

clients/drcachesim/tests/record_filter_unit_tests.cpp
```diff
@@ -427,6 +431,174 @@ test_encodings2regdeps_filter()
...
+    };
+
+    /* Construct func_id_filter_t.
+     */
+    std::vector<uint64_t> func_ids_to_keep = { static_cast<uint64_t>(SYSCALL_FUTEX_ID),
+                                               static_cast<uint64_t>(FUNC_ID_TO_KEEP) };
+    std::vector<std::unique_ptr<record_filter_func_t>> filters;
+    auto func_id_filter = std::unique_ptr<record_filter_func_t>(
+        new dynamorio::drmemtrace::func_id_filter_t(func_ids_to_keep));
+    if (func_id_filter->get_error_string() != "") {
```
derekbruening, line 581: Prefer ! .empty()

clients/drcachesim/tools/filter/func_id_filter.h
```diff
@@ -0,0 +1,141 @@
...
+/* This filter takes a list of function IDs for which it preserves the related
+ * TRACE_MARKER_TYPE_FUNC_[ID | ARG | RETVAL | RETADDR] markers. It removes all the other
+ * TRACE_MARKER_TYPE_FUNC_ markers related to function IDs that are not in the list.
+ */
+class func_id_filter_t : public record_filter_t::record_filter_func_t {
+public:
+    func_id_filter_t(std::vector<uint64_t> keep_func_ids_list)
+    {
+        for (auto keep_func_id : keep_func_ids_list) {
+            keep_func_ids_set_.insert(keep_func_id);
```
derekbruening, line 56: Simpler to avoid the whole loop and use: keep_func_ids_set_.insert(keep_func_ids_list.cbegin(), keep_func_ids_list.cend())

clients/drcachesim/tools/filter/record_filter_create.h
```diff
@@ -64,14 +64,18 @@ namespace drmemtrace {
  *   whose value is greater than this parameter.
  * @param[in] encodings2regdeps  If true, converts instruction encodings from the real ISA
  *   of the input trace to the #DR_ISA_REGDEPS synthetic ISA.
+ * @param[in] keep_func_ids  A comma-separated list of integers representing the
+ *   function IDs related to #TRACE_MARKER_TYPE_FUNC_ID (and _ARG, _RETVAL, _RETADDR)
+ *   markers to preserve in the trace.
```
derekbruening, line 69: "while removing all other function markers"

clients/drcachesim/tools/record_filter_launcher.cpp
```diff
@@ -128,8 +128,19 @@ static droption_t<uint64_t> op_trim_after_timestamp(
 droption_t<bool> op_encodings2regdeps(
     DROPTION_SCOPE_FRONTEND, "filter_encodings2regdeps", false,
     "Enable converting the encoding of instructions to synthetic ISA DR_ISA_REGDEPS.",
-    "This option is for -simulator_type record_filter. When present, it converts "
+    "This option is for -tool record_filter. When present, it converts "
     "the encoding of instructions from a real ISA to the DR_ISA_REGDEPS synthetic ISA.");
+
+/* XXX i#6369: we should partition our options by tool. This one should belong to the
```
derekbruening, line 134: This comment doesn't really make sense here as this is a single tool: should be removed.

clients/drcachesim/tools/filter/func_id_filter.h
```diff
@@ -0,0 +1,141 @@
...
+    parallel_shard_filter(
+        trace_entry_t &entry, void *shard_data,
+        record_filter_t::record_filter_info_t &record_filter_info) override
+    {
+        /* Get per_shard private data.
+         */
+        per_shard_t *per_shard = reinterpret_cast<per_shard_t *>(shard_data);
+
+        trace_type_t entry_type = static_cast<trace_type_t>(entry.type);
+        if (entry_type == TRACE_TYPE_MARKER) {
```
derekbruening, line 79: Invert this to reduce indentation: return if it's not a marker, with the if body now at top level.

derekbruening (approved):

clients/drcachesim/tools/filter/func_id_filter.h
edeiana: Done.

clients/drcachesim/tools/record_filter_launcher.cpp
edeiana: Got it, removed, also from the other file.

clients/drcachesim/tools/filter/record_filter_create.h
edeiana: Done.

clients/drcachesim/tools/filter/func_id_filter.h
edeiana: Oh, right. Done.

clients/drcachesim/tests/record_filter_unit_tests.cpp
edeiana: Done.

clients/drcachesim/common/options.cpp
edeiana: Done.

api/docs/release.dox
edeiana: Done, also mentioned the CLI option "-filter_keep_func_ids".

api/docs/release.dox
edeiana: Done.

edeiana: Done.

clients/drcachesim/tests/record_filter_unit_tests.cpp
edeiana: Got it, clarified that it's "record_filter" that needs it to output the trace.

derekbruening:
No the other file options.cpp is for all tools and should keep the comment.

clients/drcachesim/tools/record_filter_launcher.cpp
edeiana: My bad, reintroduced it in options.cpp.

Issue #6822, opened 2024-05-21
Add unscheduled-thread modeling to the drmemtrace scheduler


derekbruening:
The drmemtrace scheduler currently has support for TRACE_MARKER_TYPE_DIRECT_SWITCH but it does not fully model "unscheduled" threads which will not run until explicitly woken up by another thread.  This issue covers adding this support through new markers for waiting, resuming, and timeouts.

PR #6823, opened 2024-05-21
i#6822 unscheduled: Virtualize scheduler marker processing

derekbruening:
Refactors the scheduler marker processing into two virtual functions subclasses can override: the initialization time header marker reading now has process_next_initial_record() while regular processing has process_marker().

No change in functionality.  Future work will add "unscheduled thread" support with new markers, building on this refactoring.

Issue: #6822

### DIFF
--- a/clients/drcachesim/scheduler/scheduler.cpp
+++ b/clients/drcachesim/scheduler/scheduler.cpp
@@ -1499,7 +1499,8 @@ scheduler_tmpl_t<RecordType, ReaderType>::get_initial_input_content(
         input_info_t &input = inputs_[i];
         bool found_filetype = false;
         bool found_timestamp = !gather_timestamps || input.next_timestamp > 0;
-        if (!found_filetype || !found_timestamp) {
+        if (process_next_initial_record(input, create_invalid_record(), found_filetype,
+                                        found_timestamp)) {
             // First, check any queued records in the input.
             // XXX: Can we create a helper to iterate the queue and then the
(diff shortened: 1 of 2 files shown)


clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -1507,7 +1515,8 @@ scheduler_tmpl_t<RecordType, ReaderType>::get_initial_input_content(
                 }
                 if (record_type_is_timestamp(record, input.next_timestamp))
                     found_timestamp = true;
-                if (found_filetype && found_timestamp)
+                if (!process_next_initial_record(input, record, found_filetype,
```
edeiana, line 1518: nit: `if (!found_filetype || !found_timestamp) {` at lines 1502 and 1525 check the same condition as the "default" (non-overidden) `process_next_initial_record()`.
I assume this was intentional (perhaps we don't want subclasses to change those two checks), but I thought I'd point it out, in case we actually want them to be the same.

edeiana (approved):

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Good point: it does seem it would be best to put the call to the new function in both places.  It is unlikely there is anything in the queue so in practice that part generally won't matter; the up-front check could matter if timestamp deps are turned off.  We have to pass an invalid record at those points but that is fine (added note about that to the decl).  Done.

PR #6825, opened 2024-05-23
Add encode/decode for new TSXLDTRK instructions.

khuey:
The new TSXLDTRK feature in Sapphire Rapids adds two new instructions, XSUSLDTRK and XRESLDTRK. This commit adds support for them to the decoding and encoding machinery and adds tests.

### DIFF
--- a/core/ir/x86/decode_table.c
+++ b/core/ir/x86/decode_table.c
@@ -7079,8 +7079,8 @@ const instr_info_t rm_extensions[][8] = {
     {OP_enclu,  0xd70f0172, catUncategorized, "enclu", eax, ebx, eax, ebx, ecx, mrm|xop, x, exop[0xff]},
   },
   { /* rm extension 5 */
-    {OP_xsusldtrk, 0xf201e808, catUncategorized, "xsusldtrk", xx, xx, xx, xx, xx, reqp, x, END_LIST},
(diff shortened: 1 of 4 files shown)


core/ir/x86/decode_table.c
```diff
@@ -7075,8 +7079,8 @@ const instr_info_t rm_extensions[][8] = {
     {OP_enclu,  0xd70f0172, catUncategorized, "enclu", eax, ebx, eax, ebx, ecx, mrm|xop, x, exop[0xff]},
   },
   { /* rm extension 5 */
-    {INVALID,   0x0f0131, catUncategorized, "(bad)", xx, xx, xx, xx, xx, no, x, NA},
-    {INVALID,   0x0f0131, catUncategorized, "(bad)", xx, xx, xx, xx, xx, no, x, NA},
+    {OP_xsusldtrk, 0xf201e808, catUncategorized, "xsusldtrk", xx, xx, xx, xx, xx, reqp, x, END_LIST},
```
derekbruening, line 7082: Let's put in a category for any new instruction.  I guess these would be catOther but that seems better than uncategorized.

derekbruening (approved):

core/ir/x86/decode_table.c
khuey: Done.

PR #6826, opened 2024-05-24
i#6822 unscheduled: Add unscheduled-input drmemtrace support

derekbruening:
Augments the drmemtrace scheduler with a new notion of an "unscheduled" input which, if it has no timeout, is not runnable indefinitely until another input explicitly wakes it up.

Adds 3 new markers:
+ TRACE_MARKER_TYPE_SYSCALL_UNSCHEDULE, which makes the caller "unscheduled".
+ TRACE_MARKER_TYPE_SYSCALL_SCHEDULE, which makes a target no longer "unscheduled".
+ TRACE_MARKER_TYPE_SYSCALL_ARG_TIMEOUT which adds a timeout parameter to TRACE_MARKER_TYPE_SYSCALL_UNSCHEDULE and TRACE_MARKER_TYPE_DIRECT_THREAD_SWITCH

Adds handling for the new marker types.

Changes TRACE_MARKER_TYPE_DIRECT_THREAD_SWITCH to make the source input unscheduled, unless a timeout is passed which is honored (previously it was ignored and the syscall latency was used as the block time if it was over the thresholds).

Adds a fallback in case a state is reached with no schedulable inputs yet some number of unscheduled inputs who would otherwise hang forever.

Adds unit tests.

Also tested on a large trace with many real-world cases of these direct switches and unschedulable threads.

Issue: #6822

### DIFF
--- a/clients/drcachesim/common/trace_entry.h
+++ b/clients/drcachesim/common/trace_entry.h
@@ -567,10 +567,16 @@ typedef enum {
      * #TRACE_MARKER_TYPE_SYSCALL and #TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL markers)
      * that causes an immediate switch to another thread on the same core (with the
      * current thread entering an unscheduled state), bypassing the kernel scheduler's
-     * normal dynamic switch code based on run queues.  The marker value holds the
-     * thread id of the target thread.  This should generally always be after a
-     * #TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL marker as such a switch always
-     * has a chance of blocking if the target needs to be migrated.
+     * normal dynamic switch code based on run queues.  The marker value holds the thread
+     * id of the target thread.  The current thread will remain unschedulable
+     * indefinitely unless another thread resumes it with either
+     * #TRACE_MARKER_TYPE_DIRECT_THREAD_SWITCH or #TRACE_MARKER_TYPE_SYSCALL_SCHEDULE;
+     * or, if a #TRACE_MARKER_TYPE_SYSCALL_ARG_TIMEOUT marker is present, the thread will
+     * become schedulable when that timeout expires.  This marker should generally always
+     * be after a #TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL marker as such a switch
+     * always has a chance of blocking the source thread.  See also
+     * #TRACE_MARKER_TYPE_SYSCALL_ARG_TIMEOUT, #TRACE_MARKER_TYPE_SYSCALL_UNSCHEDULE, and
+     * #TRACE_MARKER_TYPE_SYSCALL_SCHEDULE.
      */
     TRACE_MARKER_TYPE_DIRECT_THREAD_SWITCH,
 
@@ -626,6 +632,39 @@ typedef enum {
      * the decoder using dr_set_sve_vector_length().
      */
     TRACE_MARKER_TYPE_VECTOR_LENGTH,
+
+    /**
+     * This marker is emitted prior to a system call (but after the system call's
+     * #TRACE_MARKER_TYPE_SYSCALL and #TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL markers)
+     * that causes the current thread to become unschedulable (removed from all queues of
+     * runnable threads).  The thread will remain unschedulable indefinitely unless
+     * another thread resumes it with either #TRACE_MARKER_TYPE_DIRECT_THREAD_SWITCH or
+     * #TRACE_MARKER_TYPE_SYSCALL_SCHEDULE; or, if a
+     * #TRACE_MARKER_TYPE_SYSCALL_ARG_TIMEOUT marker is present, the thread will become
+     * schedulable when that timeout expires.  This marker should generally always be
+     * after a #TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL marker as becoming unschedulable
+     * is a form of blocking and results in a context switch.
+     */
+    TRACE_MARKER_TYPE_SYSCALL_UNSCHEDULE,
+
+    /**
+     * This marker is emitted prior to a system call (but after the system call's
+     * #TRACE_MARKER_TYPE_SYSCALL marker) that causes a target thread identified in the
+     * marker value to become schedulable again if it were currently unschedulable.  See
+     * also #TRACE_MARKER_TYPE_SYSCALL_UNSCHEDULE and
+     * #TRACE_MARKER_TYPE_DIRECT_THREAD_SWITCH.
+     */
+    TRACE_MARKER_TYPE_SYSCALL_SCHEDULE,
+
+    /**
+     * This marker is emitted prior to a system call (but after the system call's
+     * #TRACE_MARKER_TYPE_SYSCALL and #TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL markers)
+     * which also has a #TRACE_MARKER_TYPE_DIRECT_THREAD_SWITCH or
+     * #TRACE_MARKER_TYPE_SYSCALL_UNSCHEDULE marker.  This indicates a timeout after which
+     * the thread will become schedulable again.  The marker value holds the timeout
+     * duration in microseconds.
+     */
+    TRACE_MARKER_TYPE_SYSCALL_ARG_TIMEOUT,
 
     // ...
     // These values are reserved for future built-in marker types.
--- a/clients/drcachesim/common/trace_entry.h
+++ b/clients/drcachesim/common/trace_entry.h
@@ -572,11 +572,16 @@ typedef enum {
      * indefinitely unless another thread resumes it with either
      * #TRACE_MARKER_TYPE_DIRECT_THREAD_SWITCH or #TRACE_MARKER_TYPE_SYSCALL_SCHEDULE;
      * or, if a #TRACE_MARKER_TYPE_SYSCALL_ARG_TIMEOUT marker is present, the thread will
-     * become schedulable when that timeout expires.  This marker should generally always
-     * be after a #TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL marker as such a switch
-     * always has a chance of blocking the source thread.  See also
+     * become schedulable when that timeout expires.  This marker provides a mechanism to
+     * model these semantics while abstracting away whether the underlying system call is
+     * a custom kernel extension or a variant of "futex" or other selective wait-notify
+     * scheme.  This marker should generally always be after a
+     * #TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL marker as such a switch always has a
+     * chance of blocking the source thread.  See also
      * #TRACE_MARKER_TYPE_SYSCALL_ARG_TIMEOUT, #TRACE_MARKER_TYPE_SYSCALL_UNSCHEDULE, and
-     * #TRACE_MARKER_TYPE_SYSCALL_SCHEDULE.
+     * #TRACE_MARKER_TYPE_SYSCALL_SCHEDULE.  The scheduler only models this behavior when
+     * #dynamorio::drmemtrace::scheduler_tmpl_t::scheduler_options_t.honor_direct_switches
+     * is true.
      */
     TRACE_MARKER_TYPE_DIRECT_THREAD_SWITCH,
 
@@ -641,18 +646,30 @@ typedef enum {
      * another thread resumes it with either #TRACE_MARKER_TYPE_DIRECT_THREAD_SWITCH or
      * #TRACE_MARKER_TYPE_SYSCALL_SCHEDULE; or, if a
      * #TRACE_MARKER_TYPE_SYSCALL_ARG_TIMEOUT marker is present, the thread will become
-     * schedulable when that timeout expires.  This marker should generally always be
-     * after a #TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL marker as becoming unschedulable
-     * is a form of blocking and results in a context switch.
+     * schedulable when that timeout expires.  This marker provides a mechanism to model
+     * these semantics while abstracting away whether the underlying system call is a
+     * custom kernel extension or a variant of "futex" or other selective wait-notify
+     * scheme.  This marker should generally always be after a
+     * #TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL marker as becoming unschedulable is a
+     * form of blocking and results in a context switch.  The scheduler only models this
+     * behavior when
+     * #dynamorio::drmemtrace::scheduler_tmpl_t::scheduler_options_t.honor_direct_switches
+     * is true.
      */
     TRACE_MARKER_TYPE_SYSCALL_UNSCHEDULE,
 
     /**
      * This marker is emitted prior to a system call (but after the system call's
      * #TRACE_MARKER_TYPE_SYSCALL marker) that causes a target thread identified in the
-     * marker value to become schedulable again if it were currently unschedulable.  See
-     * also #TRACE_MARKER_TYPE_SYSCALL_UNSCHEDULE and
-     * #TRACE_MARKER_TYPE_DIRECT_THREAD_SWITCH.
+     * marker value to become schedulable again if it were currently unschedulable or if
+     * it is not currently unschedulable to *not* become unschedulable on its next action
+     * that would otherwise do so.  See also #TRACE_MARKER_TYPE_SYSCALL_UNSCHEDULE and
+     * #TRACE_MARKER_TYPE_DIRECT_THREAD_SWITCH.  This marker provides a mechanism to
+     * model these semantics while abstracting away whether the underlying system call is
+     * a custom kernel extension or a variant of "futex" or other selective wait-notify
+     * scheme.  The scheduler only models this behavior when
+     * #dynamorio::drmemtrace::scheduler_tmpl_t::scheduler_options_t.honor_direct_switches
+     * is true.
      */
     TRACE_MARKER_TYPE_SYSCALL_SCHEDULE,
 
@@ -660,9 +677,9 @@ typedef enum {
      * This marker is emitted prior to a system call (but after the system call's
      * #TRACE_MARKER_TYPE_SYSCALL and #TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL markers)
      * which also has a #TRACE_MARKER_TYPE_DIRECT_THREAD_SWITCH or
-     * #TRACE_MARKER_TYPE_SYSCALL_UNSCHEDULE marker.  This indicates a timeout after which
-     * the thread will become schedulable again.  The marker value holds the timeout
-     * duration in microseconds.
+     * #TRACE_MARKER_TYPE_SYSCALL_UNSCHEDULE marker.  This indicates a timeout provided
+     * by the application after which the thread will become schedulable again.  The
+     * marker value holds the timeout duration in microseconds.
      */
     TRACE_MARKER_TYPE_SYSCALL_ARG_TIMEOUT,
 
--- a/clients/drcachesim/scheduler/scheduler.cpp
+++ b/clients/drcachesim/scheduler/scheduler.cpp
@@ -931,15 +931,21 @@ scheduler_tmpl_t<RecordType, ReaderType>::set_initial_schedule(
                     set_cur_input(i, INVALID_INPUT_ORDINAL);
             }
         } else {
-            // Just take the 1st N inputs (even if all from the same workload).
+            // Just take the 1st N scheduled inputs (even if all from the same workload).
+            input_ordinal_t input = 0;
             for (int i = 0; i < static_cast<output_ordinal_t>(outputs_.size()); ++i) {
-                if (i < static_cast<input_ordinal_t>(inputs_.size()))
-                    set_cur_input(i, i);
+                while (input < static_cast<input_ordinal_t>(inputs_.size()) &&
+                       inputs_[input].unscheduled) {
+                    add_to_ready_queue(&inputs_[input]);
+                    ++input;
+                }
+                if (input < static_cast<input_ordinal_t>(inputs_.size()))
+                    set_cur_input(i, input);
                 else
                     set_cur_input(i, INVALID_INPUT_ORDINAL);
+                ++input;
             }
-            for (int i = static_cast<output_ordinal_t>(outputs_.size());
-                 i < static_cast<input_ordinal_t>(inputs_.size()); ++i) {
+            for (int i = input; i < static_cast<input_ordinal_t>(inputs_.size()); ++i) {
                 add_to_ready_queue(&inputs_[i]);
             }
         }
@@ -1479,6 +1485,12 @@ bool
 scheduler_tmpl_t<RecordType, ReaderType>::process_next_initial_record(
     input_info_t &input, RecordType record, bool found_filetype, bool found_timestamp)
 {
+    // TODO i#6822: Always look ahead until the first instruction, looking
+    // for threads that start out with an exit from an UNSCHEDULE or DIRECT
+    // syscall so we can have them start out unscheduled: though we can't
+    // easily know whether there was a timeout unless we read way ahead past
+    // signal handlers until the syscall exits to look for -ETIMEDOUT.
+    // Should we have raw2trace look for that?
     return !(found_filetype && found_timestamp);
 }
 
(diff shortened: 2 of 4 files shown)


abhinav92003:
I started the review but I feel I need to spend more time with this PR. This review will be a bit delayed. Sorry about that.

clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -3846,7 +3854,389 @@ test_direct_switch()
...
+        make_marker(TRACE_MARKER_TYPE_CPU_ID, 0),
+        // C makes a long-latency blocking syscall, testing whether
+        // A is still unscheduled.
+        // We also test _SCHEDULE avoiding a future unschedule when C
+        // unblocks.
+        make_marker(TRACE_MARKER_TYPE_SYSCALL, 999),
+        make_marker(TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL, 0),
+        make_timestamp(7002),
+        make_instr(/*pc=*/501),
+        // C asks to go unscheduled with no timeout, but a prior _SCHEDULE
```
abhinav92003, line 3989: It is interesting that the "schedule" marker may affect a future syscall. What real world scenario does this behavior model? Please also update the marker doc which currently says "become schedulable again if it were **currently** unschedulable"

clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -3846,7 +3854,389 @@ test_direct_switch()
...
+        make_marker(TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL, 0),
+        // Test going unscheduled with no timeout.
+        make_marker(TRACE_MARKER_TYPE_SYSCALL_UNSCHEDULE, 0),
+        make_timestamp(4202),
+        // B makes us scheduled again.
+        make_instr(/*pc=*/102),
+        make_marker(TRACE_MARKER_TYPE_CPU_ID, 0),
+        make_marker(TRACE_MARKER_TYPE_SYSCALL, 999),
+        make_marker(TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL, 0),
+        // Switch to a missing thread to leave us and B unscheduled,
```
abhinav92003, line 4119: How does switching to a missing thread leave B unscheduled? I expected to see a direct switch or a "schedule" marker to B below, but there is none. IOW: how does B rely on A to allow it to continue?

clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -3846,7 +3854,389 @@ test_direct_switch()
...
+        // to test hang workarounds.
+        make_marker(TRACE_MARKER_TYPE_DIRECT_THREAD_SWITCH, TID_BASE + 4),
+        make_timestamp(4402),
+        // We won't get here until the hang workaround.
```
abhinav92003, line 4123: s/hang/unknown-thread-switch-hang/ or something, to qualify the generic "hang".

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -3111,6 +3289,47 @@ scheduler_tmpl_t<RecordType, ReaderType>::eof_or_idle(output_ordinal_t output)
         assert(options_.mapping != MAP_AS_PREVIOUSLY || outputs_[output].at_eof);
         return sched_type_t::STATUS_EOF;
     } else {
+        if (options_.mapping == MAP_TO_ANY_OUTPUT) {
```
abhinav92003, line 3292: Would this class become any simpler if it were templatized based on options_.mapping? (not for this PR)

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -3111,6 +3289,47 @@ scheduler_tmpl_t<RecordType, ReaderType>::eof_or_idle(output_ordinal_t output)
...
+            auto scoped_lock = hold_sched_lock
+                ? std::unique_lock<std::mutex>()
+                : std::unique_lock<std::mutex>(sched_lock_);
+            VPRINT(this, 4,
+                   "eof_or_idle output=%d live=%d unsched=%zu runq=%zu blocked=%d\n",
+                   output, live_input_count_.load(std::memory_order_acquire),
+                   unscheduled_priority_.size(), ready_priority_.size(), num_blocked_);
+            if (ready_priority_.empty() && !unscheduled_priority_.empty()) {
+                if (outputs_[output].wait_start_time == 0) {
+                    outputs_[output].wait_start_time = get_output_time(output);
```
abhinav92003, line 3305: Add comments. I suppose this is the case where the output is indefinitely unscheduled? Curious why we're tracking wait_start_time at all? If there were a timeout, it would be added to ready_priority_ right?

Oh, is the else-block the hang workaround? Add to comment that that's why.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -3111,6 +3289,47 @@ scheduler_tmpl_t<RecordType, ReaderType>::eof_or_idle(output_ordinal_t output)
...
+                } else {
+                    uint64_t now = get_output_time(output);
+                    if (now - outputs_[output].wait_start_time >
+                        options_.block_time_max) {
+                        // XXX i#6822: We may want some other options here for
```
abhinav92003, line 3310: Add comment by eof_or_idle that it may actually make changes to state (I incorrectly assumed previously that it's just a convenience read-only method).

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -3111,6 +3289,47 @@ scheduler_tmpl_t<RecordType, ReaderType>::eof_or_idle(output_ordinal_t output)
...
+                        // what to do.  We have FIFO order in unscheduled_priority_
```
abhinav92003, line 3311: When would releasing multiple inputs lead to a different output? We're releasing the highest priority one, so that will be the one to run regardless of how many we release right?

clients/drcachesim/common/trace_entry.h
```diff
@@ -627,6 +633,39 @@ typedef enum {
...
+     * This marker is emitted prior to a system call (but after the system call's
+     * #TRACE_MARKER_TYPE_SYSCALL and #TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL markers)
+     * that causes the current thread to become unschedulable (removed from all queues of
+     * runnable threads).  The thread will remain unschedulable indefinitely unless
+     * another thread resumes it with either #TRACE_MARKER_TYPE_DIRECT_THREAD_SWITCH or
+     * #TRACE_MARKER_TYPE_SYSCALL_SCHEDULE; or, if a
+     * #TRACE_MARKER_TYPE_SYSCALL_ARG_TIMEOUT marker is present, the thread will become
+     * schedulable when that timeout expires.  This marker should generally always be
+     * after a #TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL marker as becoming unschedulable
+     * is a form of blocking and results in a context switch.
```
abhinav92003, line 646: Add that this behavior is modelled only when honor_direct_switches.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -2669,6 +2742,96 @@ scheduler_tmpl_t<RecordType, ReaderType>::process_marker(input_info_t &input,
...
+        break;
+    }
+    case TRACE_MARKER_TYPE_SYSCALL_ARG_TIMEOUT:
+        // This is cleared at the post-syscall instr.
+        input.syscall_timeout_arg = static_cast<uint64_t>(marker_value);
+        break;
+    case TRACE_MARKER_TYPE_SYSCALL_UNSCHEDULE:
+        if (!options_.honor_direct_switches)
+            break;
+        if (input.skip_next_unscheduled) {
```
abhinav92003, line 2772: Explain this further. It's interesting that a prior "schedule" marker may have such an effect on a future "unschedule", especially because the "schedule" marker docs say "if it were __currently__ unschedulable".

What real world scenario is this trying to model? What if the "do-not-unschedule" input was blocked on a I/O syscall?

Also, (not saying that we should) what's the rationale behind skip_next_unscheduled being a bool and not an int (maybe multiple other threads want to "schedule" the thread), therefore the thread should get that many get-out-of-unschedulable-state tokens.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -2669,6 +2742,96 @@ scheduler_tmpl_t<RecordType, ReaderType>::process_marker(input_info_t &input,
...
+                acquire_scoped_sched_lock_if_necessary(need_sched_lock);
+            input_info_t *target = &inputs_[target_idx];
+            std::lock_guard<std::mutex> lock(*target->lock);
+            if (target->unscheduled) {
+                target->unscheduled = false;
+                if (unscheduled_priority_.find(target)) {
+                    add_to_ready_queue(target);
+                    unscheduled_priority_.erase(target);
+                } else if (ready_priority_.find(target)) {
+                    if (target->blocked_time > 0) {
```
abhinav92003, line 2820: What if the thread is doing some I/O and actually must remain blocked? (In real world, such a blocked thread cannot be unblocked simply because some other thread requests a "schedule").

clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -3846,7 +3854,389 @@ test_direct_switch()
...
+        scheduler_t::scheduler_options_t sched_ops(scheduler_t::MAP_TO_ANY_OUTPUT,
+                                                   scheduler_t::DEPENDENCY_TIMESTAMPS,
+                                                   scheduler_t::SCHEDULER_DEFAULTS,
+                                                   /*verbosity=*/3);
+        sched_ops.quantum_duration = QUANTUM_DURATION;
+        // We use our mock's time==instruction count for a deterministic result.
+        sched_ops.quantum_unit = scheduler_t::QUANTUM_TIME;
+        sched_ops.blocking_switch_threshold = BLOCK_LATENCY;
+        sched_ops.block_time_scale = BLOCK_SCALE;
+        sched_ops.block_time_max = BLOCK_TIME_MAX;
```
abhinav92003, line 4215: Add to block_time_max docs that it is also important for avoiding hangs due to "unschedule".

abhinav92003 (approved):

clients/drcachesim/common/trace_entry.h
```diff
@@ -627,6 +633,39 @@ typedef enum {
...
+     */
+    TRACE_MARKER_TYPE_SYSCALL_UNSCHEDULE,
```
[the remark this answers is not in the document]
derekbruening: I don't agree: it should be thought of as part of the gathered trace as much as any other syscall annotation, as it indicates the original app's behavior there.  It's not based on a "heuristic": it's from the precise system call type.

clients/drcachesim/common/trace_entry.h
```diff
@@ -627,6 +633,39 @@ typedef enum {
...
+     * #TRACE_MARKER_TYPE_DIRECT_THREAD_SWITCH.
+     */
+    TRACE_MARKER_TYPE_SYSCALL_SCHEDULE,
+
+    /**
+     * This marker is emitted prior to a system call (but after the system call's
+     * #TRACE_MARKER_TYPE_SYSCALL and #TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL markers)
+     * which also has a #TRACE_MARKER_TYPE_DIRECT_THREAD_SWITCH or
+     * #TRACE_MARKER_TYPE_SYSCALL_UNSCHEDULE marker.  This indicates a timeout after which
+     * the thread will become schedulable again.  The marker value holds the timeout
```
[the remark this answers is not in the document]
derekbruening: None of these are from heuristics.  They are direct system call types and system call parameters.  They are markers because these system calls are not standard in the public Linux kernel.  The timeouts cannot be varied: these are real values used by the app.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -931,15 +931,21 @@ scheduler_tmpl_t<RecordType, ReaderType>::set_initial_schedule(
                     set_cur_input(i, INVALID_INPUT_ORDINAL);
             }
         } else {
-            // Just take the 1st N inputs (even if all from the same workload).
+            // Just take the 1st N scheduled inputs (even if all from the same workload).
```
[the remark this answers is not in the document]
derekbruening: Yes.  s/N/#outputs/.

clients/drcachesim/common/trace_entry.h
```diff
@@ -567,10 +567,16 @@ typedef enum {
      * #TRACE_MARKER_TYPE_SYSCALL and #TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL markers)
```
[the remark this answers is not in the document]
derekbruening: Reworded to "Adds a fallback in case a state is reached with no schedulable inputs yet some number of unscheduled inputs who would otherwise hang forever."

clients/drcachesim/common/trace_entry.h
```diff
@@ -627,6 +633,39 @@ typedef enum {
      */
     TRACE_MARKER_TYPE_VECTOR_LENGTH,
 
+    /**
+     * This marker is emitted prior to a system call (but after the system call's
```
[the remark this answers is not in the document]
derekbruening: It's actually after the system call gateway instruction fetch, like all the other syscall markers.

clients/drcachesim/common/trace_entry.h
```diff
@@ -627,6 +633,39 @@ typedef enum {
...
+
+    /**
+     * This marker is emitted prior to a system call (but after the system call's
+     * #TRACE_MARKER_TYPE_SYSCALL and #TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL markers)
+     * which also has a #TRACE_MARKER_TYPE_DIRECT_THREAD_SWITCH or
+     * #TRACE_MARKER_TYPE_SYSCALL_UNSCHEDULE marker.  This indicates a timeout after which
+     * the thread will become schedulable again.  The marker value holds the timeout
+     * duration in microseconds.
+     */
+    TRACE_MARKER_TYPE_SYSCALL_ARG_TIMEOUT,
```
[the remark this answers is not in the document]
derekbruening: It *is* an argument to the syscall.  It could be used for other syscalls too in the future so I don't think we need to limit it to directly-scheduling-related syscalls.

clients/drcachesim/common/trace_entry.h
```diff
@@ -627,6 +633,39 @@ typedef enum {
...
+    TRACE_MARKER_TYPE_SYSCALL_UNSCHEDULE,
+
+    /**
+     * This marker is emitted prior to a system call (but after the system call's
+     * #TRACE_MARKER_TYPE_SYSCALL marker) that causes a target thread identified in the
+     * marker value to become schedulable again if it were currently unschedulable.  See
+     * also #TRACE_MARKER_TYPE_SYSCALL_UNSCHEDULE and
+     * #TRACE_MARKER_TYPE_DIRECT_THREAD_SWITCH.
+     */
+    TRACE_MARKER_TYPE_SYSCALL_SCHEDULE,
```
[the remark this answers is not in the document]
derekbruening: The challenge is that this is modeling what is not a standard public Linux syscall in the mainline kernel.  Yet we want it in place for public uses of our public workload traces.

clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -3786,11 +3791,15 @@ test_direct_switch()
         make_marker(TRACE_MARKER_TYPE_CPU_ID, 0),
         make_marker(TRACE_MARKER_TYPE_SYSCALL, 999),
         make_marker(TRACE_MARKER_TYPE_MAYBE_BLOCKING_SYSCALL, 0),
+        // This test focuses on direct only with nothing "unscheduled".
```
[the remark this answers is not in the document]
derekbruening: The semantics are such that if no timeout is provided the source becomes unscheduled.  I added "thus, we always provide a timeout to avoid going unscheduled."

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -931,15 +931,21 @@ scheduler_tmpl_t<RecordType, ReaderType>::set_initial_schedule(
...
+            input_ordinal_t input = 0;
             for (int i = 0; i < static_cast<output_ordinal_t>(outputs_.size()); ++i) {
-                if (i < static_cast<input_ordinal_t>(inputs_.size()))
-                    set_cur_input(i, i);
+                while (input < static_cast<input_ordinal_t>(inputs_.size()) &&
```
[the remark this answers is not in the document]
derekbruening: Done.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -1479,6 +1485,12 @@ bool
 scheduler_tmpl_t<RecordType, ReaderType>::process_next_initial_record(
     input_info_t &input, RecordType record, bool found_filetype, bool found_timestamp)
 {
+    // TODO i#6822: Always look ahead until the first instruction, looking
+    // for threads that start out with an exit from an UNSCHEDULE or DIRECT
+    // syscall so we can have them start out unscheduled: though we can't
+    // easily know whether there was a timeout unless we read way ahead past
+    // signal handlers until the syscall exits to look for -ETIMEDOUT.
```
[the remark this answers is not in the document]
derekbruening: If a blocking syscall is interrupted early before its waited-for-event is finished, it still blocked, and our latency-based scheme would capture the amount it actually blocked.  So we don't need to do anything different there.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -2441,16 +2486,26 @@ scheduler_tmpl_t<RecordType, ReaderType>::pick_next_input_as_previously(
     return sched_type_t::STATUS_OK;
 }
 
+template <typename RecordType, typename ReaderType>
+std::unique_lock<std::mutex>
+scheduler_tmpl_t<RecordType, ReaderType>::acquire_scoped_sched_lock_if_necessary(
+    bool &need_lock)
+{
+    need_lock =
+        options_.mapping == MAP_TO_ANY_OUTPUT || options_.mapping == MAP_AS_PREVIOUSLY;
```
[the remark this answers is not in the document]
derekbruening: Trying to find where it is repeated: this function is already a common routine I created to avoid repeating.  I do see one other case, passing hold_sched_lock to eof_or_idle on line 2368.  Is that one case what you mean?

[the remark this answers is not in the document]
derekbruening: I made need_sched_lock() for these 2 uses.

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -2515,29 +2571,46 @@ scheduler_tmpl_t<RecordType, ReaderType>::pick_next_input(output_ordinal_t outpu
                                    output, target->index);
                             --num_blocked_;
                             target->blocked_time = 0;
+                            target->unscheduled = false;
                         }
+                    } else if (unscheduled_priority_.find(target)) {
+                        target->unscheduled = false;
+                        unscheduled_priority_.erase(target);
```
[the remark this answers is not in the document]
derekbruening: No need to do anything.  The q counter is only used when an input is in a q which this is not; if it's added to a q it will get a new counter at that point.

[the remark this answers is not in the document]
derekbruening: Added a note to input_info_t.queue_counter_.

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: B separately went unscheduled.  Reworded to 
"        // Switch to a missing thread to leave us unscheduled; B also went
        // unscheduled, leaving nothing scheduled, to test hang workarounds.
"

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: Put in "no-scheduled-input hang"

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: I previously had related thoughts on subclassing to avoid the conditionals.  Filed #6831 and put XXX comments in several places in this file.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: The "Workaround to avoid hangs" comment a few lines above is already saying this.  I augmented it to try to make it clearer.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Done.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: If we release just one at a time and wait block_time_max in between each (which was my first implementation) we can end up taking a very long time to schedule them all.  (Yes, the first one will run first; it's about how long in between them.). Reworded to try to make it clearer.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: The real-world kernel allows only a single skip. I believe the real-world feature's rationale is to avoid the need for synchronization in some not-uncommon pattern of calls but I'm not sure I know what that pattern is.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Added "We assume blocked_time is from _ARG_TIMEOUT and is not from regularly-blocking i/o.  We assume i/o getting into the mix is rare enough or does not matter enough to try to have separate timeouts."

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: Done.

clients/drcachesim/common/trace_entry.h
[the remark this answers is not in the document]
abhinav92003: > It's not based on a "heuristic": it's from the precise system call type.

I think I incorrectly and silently assumed that we'll be using it for some long latency maybe_blocking system calls.

clients/drcachesim/common/trace_entry.h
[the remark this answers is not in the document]
abhinav92003: > The timeouts cannot be varied: these are real values used by the app.

I see. I didn't know they are all app provided timeouts. Thanks.

clients/drcachesim/common/trace_entry.h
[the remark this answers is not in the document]
abhinav92003: If we want to model notify condition variables, would we use the same markers?

clients/drcachesim/common/trace_entry.h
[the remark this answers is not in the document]
abhinav92003: Got it from your clarification in the other comment. Thanks.

clients/drcachesim/scheduler/scheduler.cpp
abhinav92003: > The real-world kernel allows only a single skip.

Multiple threads could want some blocked thread to be scheduled, perhaps at different times in the app's execution, but at the same time when the trace is simulated/scheduled.

clients/drcachesim/common/trace_entry.h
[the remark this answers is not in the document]
derekbruening: Added "provided by the application" to the ARG_TIMEOUT docs.

clients/drcachesim/common/trace_entry.h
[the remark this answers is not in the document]
derekbruening: Kernel-mediate wake-notify: yes, possibly.  I added this to the 3 key markers here: "This marker provides a mechanism to
     * model these semantics while abstracting away whether the underlying system call is
     * a custom kernel extension or a variant of "futex" or other selective wait-notify
     * scheme."

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: Added the ref to custom kernel or futex behavior as mentioned in the other thread.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: > especially because the "schedule" marker docs say "if it were currently unschedulable".

This was updated, as noted in the other thread.

> Multiple threads could want some blocked thread to be scheduled, perhaps at different times in the app's execution, but at the same time when the trace is simulated/scheduled.

Maybe.  There are many theoretical cases.  This re-scheduling is always going to be best-effort.  Trying to have a queue of wakers seems just as likely to cause non-representative behavior than to fix it though so it does not seem worth going in that direction.  Added a comment.

abhinav92003: If the "schedule" events get lost due to getting accumulated, then the target thready may experience a hang if some of its "unschedulable" events don't get matched up with a "schedule".

Issue #6828, opened 2024-05-25
sigmask not restored for native signal delivery with copied frame


abhinav92003:
In some cases, execute_native_handler creates a copy of the signal frame to deliver the signal to the native signal handler.

https://github.com/DynamoRIO/dynamorio/blob/427e33e31a5fa1fc611b3dbc1e5240a5bf3ccf9d/core/unix/signal.c#L6771

Looks like in this case, the app's blocked sigmask is not restored properly. This is currently not revealed by the detach_signal test because it does a siglongjmp instead of a sigreturn. Changing it to use a sigreturn for a non-sigstack signal (for which DR has to create a frame copy), the bug is revealed.

PR #6829, opened 2024-05-25
i#6828: Incorrect sigmask on frame-copy at detach-time native signal

abhinav92003:
Reproduces the bug that causes an incorrect sigmask on the frame-copy created for native delivery of a detach time signal.

This is a just a draft meant to demonstrate the issue; not meant for submission. Will add a fix soon.

Issue: #6828


abhinav92003:
#6830 fixes this.

PR #6834, opened 2024-05-31
Support subclassing drmemtrace syscall_mix data

derekbruening:
Adds a virtual destructor to the drmemtrace tool
syscall_mix_t::shard_data_t, to support subclassing that struct
for extended usage such as tracking callstacks for each syscall.

### DIFF
--- a/clients/drcachesim/tools/syscall_mix.h
+++ b/clients/drcachesim/tools/syscall_mix.h
@@ -70,6 +70,8 @@ public:
 
 protected:
     struct shard_data_t {
+        // Provide a virtual destructor to allow subclassing.
+        virtual ~shard_data_t() = default;
         std::unordered_map<int, int64_t> syscall_counts;
         std::unordered_map<int, int64_t> syscall_trace_counts;
         std::string error;


derekbruening:
a64 failure is the invalid trace type online tracing error #3320

clients/drcachesim/tools/syscall_mix.h
```diff
@@ -70,6 +70,8 @@ class syscall_mix_t : public analysis_tool_t {
 
 protected:
     struct shard_data_t {
+        // Provide a virtual destructor to allow subclassing.
+        virtual ~shard_data_t() = default;
```
edeiana, line 74: nit: perhaps we could add a sentence to the PR description giving a hint on why we have a need for subclassing of `shard_data_t` now.

edeiana (approved):

clients/drcachesim/tools/syscall_mix.h
derekbruening: Added a mention of adding callstacks to each syscall, which is my planned usage (there is no memtrace callstack library upstream right now so this would be a separate extension via the subclass at least in the short term).

PR #6835, opened 2024-06-03
i#2626: Add AARCH64 Attribute registers to the DR_REG enum

philramsey-arm:
Registers like ID_AA64ISAR0_EL1 are readable at EL0 on Linux, but are not included in the DR_REG enum as this only includes EL0 registers.

This means that MRS instructions that read these registers will not be disassembled in the normal way in the DR log.

Issue: #2626

### DIFF
--- a/core/ir/aarch64/codec.c
+++ b/core/ir/aarch64/codec.c
@@ -283,6 +283,18 @@ decode_sysreg(uint imm15)
 {
     reg_t sysreg;
     switch (imm15) {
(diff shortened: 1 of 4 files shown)


derekbruening (approved):
Maybe ask Assad or someone to check the encodings?  I did not check those.

philramsey-arm:
> Maybe ask Assad or someone to check the encodings? I did not check those.

@derekbruening Thanks. I have double checked the encodings (and the patch has been internally reviewed.)

PR #6836, opened 2024-06-03
i#3544 RV64: Implements proc_has_feature() for vector extension

ksco:
Detects RISC-V vector extension support using SIGILL in signal_arch_init(). Manually tested on real hardware.

Issue: #3544

### DIFF
--- a/core/unix/signal_linux_riscv64.c
+++ b/core/unix/signal_linux_riscv64.c
@@ -144,9 +144,8 @@ sigill_detected(void *func)
 
     kernel_sigaction_t act = { 0 };
     kernel_sigaction_t old_act;
-    act.flags = SA_ONSTACK | SA_RESTART | SA_SIGINFO;
-    act.handler = catch_sigill;
 
+    set_handler_sigact(&act, SIGILL, (handler_t)catch_sigill);
     sigaction_syscall(SIGILL, &act, &old_act);
 
     if (dr_setjmp(&jmpbuf) == 0) {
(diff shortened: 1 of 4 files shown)


core/unix/signal_linux_riscv64.c
```diff
@@ -121,8 +126,56 @@ signal_frame_extra_size(bool include_alignment)
...
+}
+
 void
 signal_arch_init(void)
 {
-    /* Nothing. */
+    /* Detects RISC-V extensions support using SIGILL.
+     * We could also use the riscv_hwprobe syscall (since kernel 6.4) or /proc/cpuinfo to
+     * detect extension support, but as of year 2024, using SIGILL is still the most
+     * reliable way for varies devices and kernel versions.
```
derekbruening, line 174: s/varies/various/

core/unix/signal_linux_riscv64.c
```diff
@@ -121,8 +126,56 @@ signal_frame_extra_size(bool include_alignment)
...
+
+static int
+sigill_detected(void *func)
+{
+    sigill_caught = 0;
+
+    kernel_sigaction_t act = { 0 };
+    kernel_sigaction_t old_act;
+    act.flags = SA_ONSTACK | SA_RESTART | SA_SIGINFO;
+    act.handler = catch_sigill;
```
derekbruening, line 148: Seems safest to use set_handler_sigact(), like the x86 SIGILL at init does, rather than replicating flags here and now having more places to update/maintain.

derekbruening (approved):

PR #6843, opened 2024-06-12
i#2626 AArch64: Add missing INSTR_CREATE_orr macros

jackgallagher-arm:
While working on another patch I noticed we were missing macros to create the GPR ORR instructions so this patch adds them along with IR unit tests.

Issue: #2626

### DIFF
--- a/core/ir/aarch64/instr_create_api.h
+++ b/core/ir/aarch64/instr_create_api.h
@@ -111,8 +111,9 @@
 #define OPND_CREATE_ZR(reg) \
     opnd_create_reg(opnd_get_size(reg) == OPSZ_4 ? DR_REG_WZR : DR_REG_XZR)
 
-/** Create an operand specifying LSL (Logical Shift Left), the default shift type when
- *  there is no shift.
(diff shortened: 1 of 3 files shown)


core/ir/aarch64/instr_create_api.h
```diff
@@ -111,9 +111,20 @@
 #define OPND_CREATE_ZR(reg) \
     opnd_create_reg(opnd_get_size(reg) == OPSZ_4 ? DR_REG_WZR : DR_REG_XZR)
 
-/** Create an operand specifying LSL, the default shift type when there is no shift. */
+/** Create an operand specifying LSL (Logical Shift Left), the default shift type when
```
derekbruening, line 114: style: `/**` on its own line (unless whole comment is one line)

derekbruening (approved):

core/ir/aarch64/instr_create_api.h
jackgallagher-arm: Done.

PR #6845, opened 2024-06-14
i#6844: Add times_of_interest to drmemtrace scheduler

### LINKED ISSUE #6844: Add timestamps-of-interest drmemtrace scheduler feature
The drmemtrace scheduler has a regions-of-interest feature to specify subsets of a trace via instruction ordinal ranges.  However, for a many-threaded trace, if these subsets should cut across all threads that were active at the same time during traces, computing the instruction ordinals for each thread requires a separate pass over every thread's trace.  A simpler interface would be to provide a single timestamp range for the whole workload and have the scheduler use the recorded schedule file to map that to per-thread instruction ordinals.

derekbruening:
Adds a new feature times_of_interest to the drmemtrace scheduler. This allows specifying regions of interest via timestamps, which cut across all inputs in a workload, providing for a consistent starting point across all threads.

The feature requires a cpu schedule file which is used to build a mapping from timestamps to instruction ordinals.  The mapping is not perfect due to collapsed consecutive entries and coarse-grained timestamps so interpolation is used in between known points.

Adds a unit test.

Adds a new analyzer command-line option -skip_to_timestamp which sets a single time-of-interest in the scheduler.  Adds a test using the checked-in threadsig x86_64 trace.

Fixes #6844

### DIFF
--- a/clients/drcachesim/analyzer_multi.cpp
+++ b/clients/drcachesim/analyzer_multi.cpp
@@ -352,6 +352,12 @@ analyzer_multi_tmpl_t<RecordType, ReaderType>::analyzer_multi_tmpl_t()
     this->worker_count_ = op_jobs.get_value();
     this->skip_instrs_ = op_skip_instrs.get_value();
     this->skip_to_timestamp_ = op_skip_to_timestamp.get_value();
+    if (this->skip_instrs_ > 0 && this->skip_to_timestamp_ > 0) {
+        this->error_string_ = "Usage error: only one of -skip_instrs and "
+                              "-skip_to_timestamp can be used at a time";
+        this->success_ = false;
+        return;
+    }
     this->interval_microseconds_ = op_interval_microseconds.get_value();
     this->interval_instr_count_ = op_interval_instr_count.get_value();
     // Initial measurements show it's sometimes faster to keep the parallel model
@@ -531,6 +537,9 @@ analyzer_multi_tmpl_t<RecordType, ReaderType>::init_dynamic_schedule()
     } else if (!op_cpu_schedule_file.get_value().empty()) {
         cpu_schedule_zip_.reset(new zipfile_istream_t(op_cpu_schedule_file.get_value()));
         sched_ops.replay_as_traced_istream = cpu_schedule_zip_.get();
+        // -cpu_schedule_file is used for two different things: actually replaying,
+        // and just input for -skip_to_timestamp.  Only if -skip_to_timestamp is 0
+        // do we actually replay.
         if (op_skip_to_timestamp.get_value() == 0) {
             sched_ops.mapping = sched_type_t::MAP_TO_RECORDED_OUTPUT;
             sched_ops.deps = sched_type_t::DEPENDENCY_TIMESTAMPS;
--- a/clients/drcachesim/analyzer_multi.cpp
+++ b/clients/drcachesim/analyzer_multi.cpp
@@ -351,6 +351,7 @@ analyzer_multi_tmpl_t<RecordType, ReaderType>::analyzer_multi_tmpl_t()
 {
     this->worker_count_ = op_jobs.get_value();
     this->skip_instrs_ = op_skip_instrs.get_value();
+    this->skip_to_timestamp_ = op_skip_to_timestamp.get_value();
     this->interval_microseconds_ = op_interval_microseconds.get_value();
     this->interval_instr_count_ = op_interval_instr_count.get_value();
     // Initial measurements show it's sometimes faster to keep the parallel model
@@ -437,6 +438,14 @@ analyzer_multi_tmpl_t<RecordType, ReaderType>::analyzer_multi_tmpl_t()
             this->parallel_ = false;
         }
         sched_ops = init_dynamic_schedule();
+    } else if (op_skip_to_timestamp.get_value() > 0) {
+#ifdef HAS_ZIP
+        if (!op_cpu_schedule_file.get_value().empty()) {
+            cpu_schedule_zip_.reset(
+                new zipfile_istream_t(op_cpu_schedule_file.get_value()));
+            sched_ops.replay_as_traced_istream = cpu_schedule_zip_.get();
+        }
+#endif
     }
 
     if (!op_indir.get_value().empty()) {
@@ -521,9 +530,11 @@ analyzer_multi_tmpl_t<RecordType, ReaderType>::init_dynamic_schedule()
         sched_ops.deps = sched_type_t::DEPENDENCY_TIMESTAMPS;
     } else if (!op_cpu_schedule_file.get_value().empty()) {
         cpu_schedule_zip_.reset(new zipfile_istream_t(op_cpu_schedule_file.get_value()));
-        sched_ops.mapping = sched_type_t::MAP_TO_RECORDED_OUTPUT;
-        sched_ops.deps = sched_type_t::DEPENDENCY_TIMESTAMPS;
         sched_ops.replay_as_traced_istream = cpu_schedule_zip_.get();
+        if (op_skip_to_timestamp.get_value() == 0) {
+            sched_ops.mapping = sched_type_t::MAP_TO_RECORDED_OUTPUT;
+            sched_ops.deps = sched_type_t::DEPENDENCY_TIMESTAMPS;
+        }
     }
 #endif
     sched_ops.kernel_switch_trace_path = op_sched_switch_file.get_value();
--- a/clients/drcachesim/common/options.cpp
+++ b/clients/drcachesim/common/options.cpp
@@ -549,7 +549,8 @@ droption_t<bytesize_t> op_skip_instrs(
     "Specifies the number of instructions to skip in the beginning of the trace "
     "analysis.  For serial iteration, this number is "
     "computed just once across the interleaving sequence of all threads; for parallel "
-    "iteration, each thread skips this many insructions.  When built with zipfile "
+    "iteration, each thread skips this many insructions (see -skip_to_timestamp for "
+    "an alternative which does align all threads).  When built with zipfile "
     "support, this skipping is optimized and large instruction counts can be quickly "
     "skipped; this is not the case for -skip_refs.");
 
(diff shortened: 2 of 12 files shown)


derekbruening:
a64-sve failure is the known invariants test "missing exit" no one has figured out: #6733

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -1255,6 +1217,205 @@ scheduler_tmpl_t<RecordType, ReaderType>::read_traced_schedule()
...
+                       "tid %" PRIu64 " overlaps with times_of_interest [%" PRIu64
+                       ", %" PRIu64 ") @ [%" PRIu64 ", %" PRIu64 ")\n",
+                       tid_it.first, times.start_timestamp, times.stop_timestamp,
+                       instr_start, instr_end);
+            }
+        }
+        if (instr_ranges.empty()) {
+            // Exclude this thread completely.  We've already created its
+            // inputs_ entry with cross-indices stored in other structures
+            // so instead of trying to erase it we mark it invalid.
```
derekbruening, line 1296: This ends up complicating interfaces where the analyzer asks about the shard count or asks about the nth input.  Yet erasing has its own issues noted here.  (We assume the analyzer/user can't pass in N inputs plus times_of_interest and expect N shards/final inputs...right?). Suggestions?

clients/drcachesim/analyzer_multi.cpp
```diff
@@ -521,9 +530,11 @@ analyzer_multi_tmpl_t<RecordType, ReaderType>::init_dynamic_schedule()
         sched_ops.deps = sched_type_t::DEPENDENCY_TIMESTAMPS;
     } else if (!op_cpu_schedule_file.get_value().empty()) {
         cpu_schedule_zip_.reset(new zipfile_istream_t(op_cpu_schedule_file.get_value()));
-        sched_ops.mapping = sched_type_t::MAP_TO_RECORDED_OUTPUT;
-        sched_ops.deps = sched_type_t::DEPENDENCY_TIMESTAMPS;
         sched_ops.replay_as_traced_istream = cpu_schedule_zip_.get();
+        if (op_skip_to_timestamp.get_value() == 0) {
```
abhinav92003, line 534: Skipping to a certain timestamp seems to be a different thing than the mapping and deps (wouldn't they behave the same after we've done the skip-to-timestamp?). Explain in comment why we need this condition.

clients/drcachesim/common/options.cpp
```diff
@@ -549,7 +549,8 @@ droption_t<bytesize_t> op_skip_instrs(
     "Specifies the number of instructions to skip in the beginning of the trace "
     "analysis.  For serial iteration, this number is "
     "computed just once across the interleaving sequence of all threads; for parallel "
-    "iteration, each thread skips this many insructions.  When built with zipfile "
+    "iteration, each thread skips this many insructions (see -skip_to_timestamp for "
```
abhinav92003, line 552: s/insructions/instructions/

clients/drcachesim/common/options.cpp
```diff
@@ -561,6 +562,18 @@ droption_t<bytesize_t>
                  "of being simulated.  This skipping may be slow for large skip values; "
                  "consider -skip_instrs for a faster method of skipping.");
 
+droption_t<uint64_t> op_skip_to_timestamp(
+    DROPTION_SCOPE_FRONTEND, "skip_to_timestamp", 0, "Timestamp to start at",
+    "Specifies a timestamp to start at, skipping over prior records in the trace. "
+    "This is cross-cutting across all threads.  If the target timestamp is not "
+    "present as a timestamp marker, interpolation is used to approximate the "
+    "target location in each thread.  If -skip_instrs is also specified, this option "
```
abhinav92003, line 570: It would be nicer to crash if both -skip_isntrs and -skip_to_timestamp are specified. That's harder to miss than this comment.

clients/drcachesim/common/options.cpp
```diff
@@ -903,7 +916,9 @@ droption_t<std::string>
     op_cpu_schedule_file(DROPTION_SCOPE_FRONTEND, "cpu_schedule_file", "",
                          "Path with stored as-traced schedule for replay",
                          "Applies to -core_sharded and -core_serial. "
-                         "Path with stored as-traced schedule for replay.");
+                         "Path with stored as-traced schedule for replay.  If specified "
+                         "with a non-zero -skip_to_timestamp, replay is not triggered "
+                         "an instead the file is used for the skip request.");
```
abhinav92003, line 921: s/an/and/

clients/drcachesim/common/options.cpp
```diff
@@ -903,7 +916,9 @@ droption_t<std::string>
     op_cpu_schedule_file(DROPTION_SCOPE_FRONTEND, "cpu_schedule_file", "",
                          "Path with stored as-traced schedule for replay",
                          "Applies to -core_sharded and -core_serial. "
-                         "Path with stored as-traced schedule for replay.");
+                         "Path with stored as-traced schedule for replay.  If specified "
+                         "with a non-zero -skip_to_timestamp, replay is not triggered "
```
abhinav92003, line 920: > replay is not triggered

But after we skip to the given timestamp, we do resume replaying from that point right?

clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -698,6 +699,16 @@ scheduler_tmpl_t<RecordType, ReaderType>::init(
                 tid2input_[workload_tid_t(workload_idx, it.first)] = it.second;
             }
         }
+        if (!workload.times_of_interest.empty()) {
+            for (const auto &modifiers : workload.thread_modifiers) {
+                if (!modifiers.regions_of_interest.empty())
```
abhinav92003, line 704: Meaning one cannot specify regions and times of interest both?

clients/drcachesim/scheduler/scheduler.cpp
abhinav92003: Don't we have the same issue if the user specifies instr count based ROIs that end up excluding some threads? Do we do anything to handle that scenario?

clients/drcachesim/tests/scheduler_unit_tests.cpp
```diff
@@ -3555,6 +3555,161 @@ test_replay_as_traced_from_file(const char *testdir)
...
+        std::vector<scheduler_t::input_workload_t> sched_inputs;
+        std::vector<scheduler_t::input_reader_t> readers;
+        for (int i = 0; i < NUM_INPUTS; i++) {
+            memref_tid_t tid = TID_BASE + i;
+            readers.emplace_back(
+                std::unique_ptr<mock_reader_t>(new mock_reader_t(inputs[i])),
+                std::unique_ptr<mock_reader_t>(new mock_reader_t()), tid);
+        }
+        sched_inputs.emplace_back(std::move(readers));
+        sched_inputs.back().times_of_interest = { { 25, 30 }, { 32, 33 } };
```
abhinav92003, line 3635: Comment that the gap is expected between instr ordinals.

abhinav92003 (approved):

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: You mean like a really high -skip_instrs?  I think that marks inputs as eof if they don't go that high.  I wonder if it hits the same internal issue I'm hitting where there are no per-shard results because a thread was never seen, causing a crash when generating per-shard results.  But if I remove the whole "valid" and detection of no overlap and try to go the same route: I don't know the instr count so I'd have to set a numeric-limit skip or sthg and have the input go through the initial scheduling on some output but then immediately hitting eof. 
 Feels inefficient.  But if that ends up working maybe that is the simplest.

derekbruening: If that also hits that per-shard-gen crash: I'll file that separately as it's orthogonal.

clients/drcachesim/analyzer_multi.cpp
derekbruening: Added:

        // -cpu_schedule_file is used for two different things: actually replaying,
        // and just input for -skip_to_timestamp.  Only if -skip_to_timestamp is 0
        // do we actually replay.

clients/drcachesim/common/options.cpp
derekbruening: Done.

clients/drcachesim/common/options.cpp
derekbruening: Done.

clients/drcachesim/common/options.cpp
derekbruening: No.  Maybe a whole separate option should be used as this kind of overloads this option; but skipping or any change on replay is not supported.  Reworded to "there is no replay" to make this clearer.

clients/drcachesim/tests/scheduler_unit_tests.cpp
derekbruening: Added:

        // Pick times that have adjacent corresponding instructions: 30 and 32
        // have a time gap but no instruction gap.

clients/drcachesim/common/options.cpp
derekbruening: Done.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Right.  Added a comment.

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Confirmed both things: A) Ripping out this "valid" concept and just specifying a max-limit start works well: that input is scheduled but it hits eof before the user sees it; and B) A high -skip_instrs would hit the same thing: but there's a usage error that prevents it.  A solution that checks never-seen inputs is pretty simple and I'll put that in the 3rd-party code.

derekbruening: Done (removed the input_info_t.valid field in favor of an ROI with a max-limit start).

PR #6848, opened 2024-06-17
i#3544 RV64 vector part2: Add basic vector support to the core

ksco:
This is a follow-up patch of adding RISC-V vector (RVV) extension support to the core, part1 in PR #6810 (f1ce1bca).

This patch:
1. fixes several issues in the codec introduced in part1, codec unit tests will be submitted separately in follow-up PRs;
2. rename and reuse SVE vector length getter/setter functions to be more concise on APIs for vector extensions;
3. adds RISC-V vector support to drdisas;
4. support code cache and clean call context switch;

For now, we support RISC-V vector lengths up to 256 bits, longer vector lengths will exceed the limit of DynamoRIO stack size and 12-bit signed immediate range.

Issue: #3544

### DIFF
--- a/core/arch/riscv64/emit_utils.c
+++ b/core/arch/riscv64/emit_utils.c
@@ -689,8 +689,8 @@ append_restore_simd_reg(dcontext_t *dcontext, instrlist_t *ilist, bool absolute)
          *
          *           ma            ta         sew=8       lmul=8 */
         vtypei = (0b1 << 7) | (0b1 << 6) | (0b000 << 3) | 0b011;
-        memopnd = opnd_create_dcontext_field_via_reg_sz(dcontext, DR_REG_A1, 0,
-                                                        reg_get_size(DR_REG_VR0));
+        memopnd = opnd_create_dcontext_field_via_reg_sz(
+            dcontext, DR_REG_A1, 0, reg_get_size_lmul(DR_REG_VR0, RV64_LMUL_8));
         APP(ilist,
             INSTR_CREATE_addi(dcontext, opnd_create_reg(DR_REG_A1),
                               opnd_create_reg(REG_DCXT),
@@ -826,8 +826,8 @@ append_save_simd_reg(dcontext_t *dcontext, instrlist_t *ilist, bool absolute)
          *
          *           ma            ta         sew=8       lmul=8 */
         vtypei = (0b1 << 7) | (0b1 << 6) | (0b000 << 3) | 0b011;
-        memopnd = opnd_create_dcontext_field_via_reg_sz(dcontext, DR_REG_A1, 0,
-                                                        reg_get_size(DR_REG_VR0));
+        memopnd = opnd_create_dcontext_field_via_reg_sz(
+            dcontext, DR_REG_A1, 0, reg_get_size_lmul(DR_REG_VR0, RV64_LMUL_8));
         APP(ilist,
             INSTR_CREATE_addi(dcontext, opnd_create_reg(DR_REG_A1),
                               opnd_create_reg(REG_DCXT),
--- a/core/arch/riscv64/emit_utils.c
+++ b/core/arch/riscv64/emit_utils.c
@@ -695,15 +695,20 @@ append_restore_simd_reg(dcontext_t *dcontext, instrlist_t *ilist, bool absolute)
             INSTR_CREATE_addi(dcontext, opnd_create_reg(DR_REG_A1),
                               opnd_create_reg(REG_DCXT),
                               opnd_create_immed_int(VREG_OFFSET(DR_REG_VR0), OPSZ_12b)));
+        /* For the following vector instructions, set the element width to 8b, and use 8
+         * registers as a group (lmul=8).
+         */
         APP(ilist,
             INSTR_CREATE_vsetvli(dcontext, opnd_create_reg(DR_REG_A0),
                                  opnd_create_reg(DR_REG_ZERO),
                                  opnd_create_immed_uint(vtypei, OPSZ_11b)));
+        /* Uses lmul=8 to copy 8 registers at a time. */
         for (int reg = DR_REG_VR0; reg <= DR_REG_VR31; reg += 8) {
             APP(ilist,
                 INSTR_CREATE_vle8_v(dcontext, opnd_create_reg(reg), memopnd,
                                     opnd_create_immed_int(1, OPSZ_1b) /* mask disabled */,
                                     opnd_create_immed_int(0, OPSZ_3b) /* nfields = 1 */));
+            /* If it's the last vector register group, no need to increase the offset. */
             if (reg != DR_REG_VR24) {
                 APP(ilist,
                     INSTR_CREATE_addi(
@@ -827,15 +832,20 @@ append_save_simd_reg(dcontext_t *dcontext, instrlist_t *ilist, bool absolute)
             INSTR_CREATE_addi(dcontext, opnd_create_reg(DR_REG_A1),
                               opnd_create_reg(REG_DCXT),
                               opnd_create_immed_int(VREG_OFFSET(DR_REG_VR0), OPSZ_12b)));
+        /* For the following vector instructions, set the element width to 8b, and use 8
+         * registers as a group (lmul=8).
+         */
         APP(ilist,
             INSTR_CREATE_vsetvli(dcontext, opnd_create_reg(DR_REG_A2),
                                  opnd_create_reg(DR_REG_ZERO),
                                  opnd_create_immed_uint(vtypei, OPSZ_11b)));
+        /* Uses lmul=8 to copy 8 registers at a time. */
         for (int reg = DR_REG_VR0; reg <= DR_REG_VR31; reg += 8) {
             APP(ilist,
                 INSTR_CREATE_vse8_v(dcontext, memopnd, opnd_create_reg(reg),
                                     opnd_create_immed_int(1, OPSZ_1b) /* mask disabled */,
                                     opnd_create_immed_int(0, OPSZ_3b) /* nfields = 1 */));
+            /* If it's the last vector register group, no need to increase the offset. */
             if (reg != DR_REG_VR24) {
                 APP(ilist,
                     INSTR_CREATE_addi(
--- a/core/arch/riscv64/emit_utils.c
+++ b/core/arch/riscv64/emit_utils.c
@@ -52,6 +52,8 @@
 /* TODO i#3544: Think of a better way to represent CSR in the IR, maybe as registers? */
 /* Number of the fcsr register. */
 #define FCSR 0x003
+#define VSTART 0x008
+#define VCSR 0x00F
 
 /* Instruction fixed bits constants. */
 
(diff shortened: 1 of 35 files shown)


derekbruening:
(May not get to this for another day)

core/arch/riscv64/emit_utils.c
```diff
@@ -656,15 +670,48 @@ void
...
+                                                        reg_get_size(DR_REG_VR0));
+        APP(ilist,
+            INSTR_CREATE_addi(dcontext, opnd_create_reg(DR_REG_A1),
+                              opnd_create_reg(REG_DCXT),
+                              opnd_create_immed_int(VREG_OFFSET(DR_REG_VR0), OPSZ_12b)));
+        APP(ilist,
+            INSTR_CREATE_vsetvli(dcontext, opnd_create_reg(DR_REG_A0),
+                                 opnd_create_reg(DR_REG_ZERO),
+                                 opnd_create_immed_uint(vtypei, OPSZ_11b)));
+        for (int reg = DR_REG_VR0; reg <= DR_REG_VR31; reg += 8) {
```
derekbruening, line 702: +=8 doesn't sound right: don't we need to loop over all vector regs?

core/arch/riscv64/emit_utils.c
```diff
@@ -656,15 +670,48 @@ void
...
+            APP(ilist,
+                INSTR_CREATE_vle8_v(dcontext, opnd_create_reg(reg), memopnd,
+                                    opnd_create_immed_int(1, OPSZ_1b) /* mask disabled */,
+                                    opnd_create_immed_int(0, OPSZ_3b) /* nfields = 1 */));
+            if (reg != DR_REG_VR24) {
```
derekbruening, line 707: What is special about 24?  Add a comment.

core/arch/riscv64/emit_utils.c
```diff
@@ -656,15 +670,48 @@ void
...
+                APP(ilist,
+                    INSTR_CREATE_addi(
+                        dcontext, opnd_create_reg(DR_REG_A1), opnd_create_reg(DR_REG_A1),
+                        opnd_create_immed_int(8 * sizeof(dr_simd_t), OPSZ_12b)));
```
derekbruening, line 711: The 8x again: is this really loading 8 vector regs at a time?  Please add a comment.  If that's the case: why is the size of `memopnd` only the size of VR0?

core/arch/riscv64/emit_utils.c
```diff
@@ -748,22 +795,55 @@ append_save_gpr(dcontext_t *dcontext, instrlist_t *ilist, bool ibl_end, bool abs
...
+                                                        reg_get_size(DR_REG_VR0));
+        APP(ilist,
+            INSTR_CREATE_addi(dcontext, opnd_create_reg(DR_REG_A1),
+                              opnd_create_reg(REG_DCXT),
+                              opnd_create_immed_int(VREG_OFFSET(DR_REG_VR0), OPSZ_12b)));
+        APP(ilist,
+            INSTR_CREATE_vsetvli(dcontext, opnd_create_reg(DR_REG_A2),
+                                 opnd_create_reg(DR_REG_ZERO),
+                                 opnd_create_immed_uint(vtypei, OPSZ_11b)));
+        for (int reg = DR_REG_VR0; reg <= DR_REG_VR31; reg += 8) {
```
derekbruening, line 834: The same 3 comments in _restore_ apply here.

derekbruening (commented):
Main concern is the 8-vector-reg stride on all the loops: is that right?

derekbruening:
May want to have the Arm folks have a look mostly as a heads up on the vector length API name changes: I'll add @AssadHashmi

core/arch/riscv64/emit_utils.c
ksco: The `vsetvli` sets the configuration of the VPU for all the following vector instructions before the next `vset*vl*` to element width (sew): 8b; vector register group multiplier (lmul): 8, effective vector length (vl): maximum.

lmul=8 means every vector instruction is going to take 8 vector registers as a whole and operate on that. So we can use 4 vector instructions here to copy all 32 vector registers. Linux kernel uses the same approach, refer to https://elixir.bootlin.com/linux/latest/source/arch/riscv/include/asm/vector.h#L103.

Added a comment in the code to explain it.

core/arch/riscv64/emit_utils.c
ksco: Added a comment.

core/arch/riscv64/emit_utils.c
ksco: Added comments same as above.

About the size of `memopnd`, this is a mistake, thank you for pointing it out! Currently, we're using `reg_get_size(DR_REG_VR0)` to get the vector register size, this is not enough for RISC-V Vector extension. So I'm thinking something like this specific to RISC-V:

```
opnd_size_t
reg_get_size_lmul(reg_id_t reg, lmul_t lmul);
```

What do you think?

core/arch/riscv64/emit_utils.c
ksco: Ditto

core/arch/riscv64/mangle.c
```diff
@@ -160,8 +164,67 @@ insert_push_all_registers(dcontext_t *dcontext, clean_call_info_t *cci,
...
+                                                        reg_get_size(DR_REG_VR0));
+        PRE(ilist, instr,
+            INSTR_CREATE_addi(dcontext, opnd_create_reg(DR_REG_A0),
+                              opnd_create_reg(DR_REG_SP),
+                              opnd_create_immed_int(dstack_offs, OPSZ_12b)));
+        PRE(ilist, instr,
+            INSTR_CREATE_vsetvli(dcontext, opnd_create_reg(DR_REG_A1),
+                                 opnd_create_reg(DR_REG_ZERO),
+                                 opnd_create_immed_uint(vtypei, OPSZ_11b)));
+        for (int reg = DR_REG_VR0; reg <= DR_REG_VR31; reg += 8) {
```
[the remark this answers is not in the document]
ksco: Ditto.

core/arch/riscv64/mangle.c
```diff
@@ -184,24 +250,85 @@ insert_pop_all_registers(dcontext_t *dcontext, clean_call_info_t *cci, instrlist
...
+                                                        reg_get_size(DR_REG_VR0));
+        PRE(ilist, instr,
+            INSTR_CREATE_addi(dcontext, opnd_create_reg(DR_REG_A0),
+                              opnd_create_reg(DR_REG_SP),
+                              opnd_create_immed_int(current_offs, OPSZ_12b)));
+        PRE(ilist, instr,
+            INSTR_CREATE_vsetvli(dcontext, opnd_create_reg(DR_REG_A1),
+                                 opnd_create_reg(DR_REG_ZERO),
+                                 opnd_create_immed_uint(vtypei, OPSZ_11b)));
+        for (int reg = DR_REG_VR0; reg <= DR_REG_VR31; reg += 8) {
```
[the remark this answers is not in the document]
ksco: Ditto.

core/ir/decode_shared.c
```diff
@@ -174,29 +174,37 @@ const char *const size_names[] = {
...
-            sve_veclen = vl;
+#if defined(AARCH64)
+    for (int i = 0; i < sizeof(sve_vector_lengths) / sizeof(sve_vector_lengths[0]); i++) {
+        if (vl == sve_vector_lengths[i]) {
+            vector_length = vl;
             return true;
         }
     }
+#elif defined(RISCV64)
+    if (vl >= 64 && vl <= 65536 && (vl & (vl - 1)) == 0) {
```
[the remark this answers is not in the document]
ksco: Done.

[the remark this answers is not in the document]
ksco: Done.

core/arch/riscv64/emit_utils.c
derekbruening: OK, that seems reasonable.

core/ir/opnd_shared.c
```diff
@@ -2795,6 +2797,30 @@ reg_get_size(reg_id_t reg)
...
+#ifdef RISCV64
+/* Returns the OPSZ_ constant corresponding to the vector register size and lmul.
+ * Page 12 of RISC-V "V" Vector Extension Version 1.0.
+ */
+opnd_size_t
+reg_get_size_lmul(reg_id_t reg, lmul_t lmul)
+{
+    if (reg >= DR_REG_VR0 && reg <= DR_REG_VR31) {
+        int power = ((int)lmul << 29) << 29;
+        opnd_size_t opsz = opnd_size_from_bytes(dr_get_vector_length() >> (3 - power));
```
derekbruening, line 2809: So the 3 is to convert bits to bytes (worth a comment).  We then need to adjust for the LMUL: for our 8-at-a-time, we want to multiply by 3, so I assume you wanted `power` to be 3.  What about the 1/2, 1/4, 1/8?  Don't they need to shift right even more?  Is this `power` supposed to end up negative?

core/lib/globals_api.h
```diff
@@ -813,10 +813,16 @@ typedef union _dr_simd_t {
 #    define MCXT_NUM_OPMASK_SLOTS 8
 
 #elif defined(RISCV64)
-
+/**
+ * 256-bit RISC-V Vector extension registers.
+ * Vector register length can be from 64 to 65536 bits in the power of 2.
+ * Currently we supports implementations of up to 256 bits due to limit of DR's
```
derekbruening, line 819: grammar: s/supports/support/

derekbruening (changes requested):
Unless I'm missing something, the reg_get_size_lmul looks incorrect.

core/ir/opnd_api.h
```diff
@@ -1631,6 +1633,22 @@ typedef ushort reg_id_t; /**< The type of a DR_REG_ enum value. */
  */
 typedef byte opnd_size_t; /**< The type of an OPSZ_ enum value. */
 
+#ifdef RISCV64
+/**
+ * The LMUL type for RISCV64 vector extension.
+ * Page 12 of RISC-V "V" Vector Extension Version 1.0.
+ */
+typedef enum {
+    RV64_LMUL_1_8 = 0b101, /**< RISC-V vector extension LMUL 1/8. */
```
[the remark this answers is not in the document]
ksco: Yes, it's in sync with the specification for the convenience of `reg_get_size_lmul()` to compute the reg size. Refined the comment.

core/ir/opnd_shared.c
```diff
@@ -2795,6 +2797,30 @@ reg_get_size(reg_id_t reg)
...
 
+#ifdef RISCV64
+/* Returns the OPSZ_ constant corresponding to the vector register size and lmul.
+ * Page 12 of RISC-V "V" Vector Extension Version 1.0.
+ */
+opnd_size_t
+reg_get_size_lmul(reg_id_t reg, lmul_t lmul)
+{
+    if (reg >= DR_REG_VR0 && reg <= DR_REG_VR31) {
+        int power = ((int)lmul << 29) << 29;
```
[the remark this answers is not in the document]
ksco: I added the unit tests.

core/ir/opnd_shared.c
```diff
@@ -2805,7 +2805,18 @@ opnd_size_t
...
+         * 1/4: -2
+         * 1/2: -1
+         *   0:  0
+         *   1:  1
+         *   2:  2
+         *   3:  3
+         * so (vlen >> (3 - lmul)) converts the hardware vector length in bits to the
+         * effective vector length in bytes.
+         */
+        int power = ((int)lmul << 29) >> 29;
```
derekbruening, line 2819: OK so this is shifting left by 29 in order to set the top bit; it then shifts back down and seems to assume that this right-shift of a negative number will sign-extend.  But that is implementation-defined so it's not guaranteed it will do what you want.  Probably it will in most implementations (seems to in GA CI as the unit test passes).

Could we do something safer: have the enum for RV64_LMUL_1_8 be -3 instead of 5?  Then we don't have to shift it at all.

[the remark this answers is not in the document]
ksco: Done.

ksco: Done.

core/ir/encode_api.h
```diff
@@ -150,21 +150,23 @@ dr_get_isa_mode(void *drcontext);
...
+ *   128 256 384 512 640 768 896 1024 1152 1280 1408 1536 1664 1792 1920 2048
+ * - RISC-V Vector Extension's vector length in bit is from 64 to 65536 in the
+ *   power of 2.
  * Returns whether successful.
  * TODO i#3044: This function will only allow setting vector length if not
- * running on SVE.
+ * running on SVE or RVV.
  */
 bool
-dr_set_sve_vector_length(int vl);
```
derekbruening, line 160: An API change needs to be documented: please add this as a compatibility-breaking change to the changelist at https://github.com/DynamoRIO/dynamorio/blob/master/api/docs/release.dox#L127

(API additions are also listed in the separate list below that one in that same file -- but since RISCV is not yet officially released no need to add new RISCV-only interfaces there yet.)

derekbruening, line 160: A compatibility change should bump the version per https://dynamorio.org/page_new_release.html#autotoc_md232.  It should be bumped from 10.92 to 10.93.  See b4909144c3 the last PR to bump it (to .92) to see the places to change: I believe it's 3 files, 2 .yml files and the top-level CMakeLists.txt.

derekbruening (approved):

PR #6851, opened 2024-06-21
i#6822 unscheduled: Add start-unscheduled support

derekbruening:
Adds support for threads starting out in an "unscheduled" state.  This is accomplished by always reading ahead in each input and looking for a TRACE_MARKER_TYPE_SYSCALL_UNSCHEDULE marker *before* the first instruction.  Normally such a marker indicates the invocation of a system call and is after the system call instruction; for start-unscheduled threads it is present at the system call exit at the start of the trace.

Changes the scheduler's virtual method process_next_initial_record() to make the booleans on finding certain markers input-and-output parameters and moves filetype marker handling and timestamp recording into the function.  This also fixes a problem where an input's initial next_timestamp was replaced with the 2nd timestamp if a subclass read ahead.

The extra readahead causes complexities elsewhere which are addressed:
+ The reader caches the last cpuid to use for synthetic recores on skipping.
+ Generalizes the existing scheduler handling of readahead (the "recorded_in_schedule" field in input_info_t) to store a count of pre-read instructions, which will generally be either 0 or 1.  Adds a new internal interface get_instr_ordinal() to get the input reader's instruction ordinal minus the pre-read count.

Changes raw2trace's virtual function process_marker_additionally() to process_marker() and moves all marker processing (including timestamps, which are not markers in the raw format) there, to better support subclasses inserting start-unscheduled markers and deciding whether to insert new markers either before or after pre-existing markers.

Adds a scheduler test for the new feature.

Issue: #6822

### DIFF
--- a/clients/drcachesim/scheduler/scheduler.cpp
+++ b/clients/drcachesim/scheduler/scheduler.cpp
@@ -266,6 +266,13 @@ scheduler_tmpl_t<memref_t, reader_t>::record_type_is_marker(memref_t record,
 
 template <>
 bool
+scheduler_tmpl_t<memref_t, reader_t>::record_type_is_non_marker_header(memref_t record)
+{
+    return false;
+}
+
+template <>
+bool
 scheduler_tmpl_t<memref_t, reader_t>::record_type_is_timestamp(memref_t record,
                                                                uintptr_t &value)
 {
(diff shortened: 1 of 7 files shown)


clients/drcachesim/scheduler/scheduler.cpp
```diff
@@ -264,6 +264,13 @@ scheduler_tmpl_t<memref_t, reader_t>::record_type_is_marker(memref_t record,
     return true;
 }
 
+template <>
+bool
+scheduler_tmpl_t<memref_t, reader_t>::record_type_is_non_marker_header(memref_t record)
+{
+    return false;
```
abhinav92003, line 271: Remind the reader that non-marker headers are always hidden in the memref_t stream.

abhinav92003 (approved):

clients/drcachesim/scheduler/scheduler.cpp
derekbruening: Done.

derekbruening:
x32 failure is opcode_mix encoding error #6303

derekbruening:
Will wait to merge until corresponding 3rd party subclass changes are fully ready.

Issue #5623, opened 2022-08-21
Add support for pointer-authenticated indirect branches in ARMv8.3


derekbruening:
PR #5497 adds AArch64 IR support for BLRAAZ, BRAA, RETAA (and some other non-branches) because apparently they are in use on Mac on M1.  This issue covers adding the rest of that set of ISA enhancements, separated from the other general AArch64 filed issues because it may be implemented separately to support M1.

jackgallagher-arm:
@derekbruening I have implemented all of the v8.3 pointer authentication instructions and addressed all of the i#5623 TODOs in the code base. Can we close this issue now, or is there anything else to do?

derekbruening:
There is a significant unresolved issue:

Currently DR turns an authenticated indirect branch into an un-authenticated branch by running XPACI to clear the top bits (necessary for a canonical address for lookup in DR's hashtables) and then doing a regular un-authenticated indirect branch in the IBL (indirect branch lookup routine).  It should be possible for DR to preserve the top pointer bits in some way (maybe it does the authentication in the fragment before the XPACI) and use an AUTI* instruction to validate: right?  DR would need to ensure an authentication failure turned into the same type of fault as would happen without DR.

Issue #6276, opened 2023-08-24
AArch64 Regression: Failure to encode XPACI after PR #5835


derekbruening:
We have a regression: hello,world (simple_app) used to run to completion on Mac M1 AArch64 but now it fails.  I ran a git bisect and it found f646a632d5fe80c34210a3ee890502919f3443a9 == PR #5835 as the culprit.

Before PR #5835:
```
$ bin64/drrun -- suite/tests/bin/simple_app
...
<other logs including "Undefined HINT instruction found: encoding 0xd503245f (CRm:op2 0x22)">
Hello, world!
<Stopping application /Users/bruening/dr/git/build_m1_dbg/suite/tests/bin/simple_app (24493)>
```

After PR #5835:
```
$ bin64/drrun -- suite/tests/bin/simple_app
...
<Internal Error: Failed to encode instruction: 'xpaci  %x2 -> %x2'
>
<Application /Users/bruening/dr/git/build_m1_dbg/suite/tests/bin/simple_app (22028).  Internal Error: DynamoRIO debug check failure: /Users/bruening/dr/git/src/core/emit.c:363 pc != NULL
```

That XPACI is coming from DR itself I believe:
```
$ git grep -i xpaci
core/arch/aarchxx/mangle.c:        PRE(ilist, instr, INSTR_CREATE_xpaci(dcontext, opnd_create_reg(IBL_TARGET_REG)));
core/arch/aarchxx/mangle.c:        PRE(ilist, instr, INSTR_CREATE_xpaci(dcontext, opnd_create_reg(IBL_TARGET_REG)));
```

derekbruening:
From @AssadHashmi : since the culprit PR did not change codec.txt, maybe the problem comes from the feature change and now the encoder thinks that XPACI is not available.

An update to the api.ir tests comment https://github.com/DynamoRIO/dynamorio/issues/6276#issuecomment-1731634120: we do already have tests of XPACI, and presumably they never failed.

jackgallagher-arm:
The problem here is that feature detection is not implemented for AArch64 on macOS. When we try to emit the `XPACI` instruction the encoder thinks that `PAUTH` is not supported by the processor so it fails to encode it.

The proper fix is to implement AArch64 feature detection on macOS, but in the mean time as a quick fix we could just hard-code enable PAUTH on macOS by adding
```
#if defined(MACOS)
proc_set_feature(FEATURE_PAUTH, true);
#endif
```
[...]

PR #6856, opened 2024-06-25
i#5623 AArch64: Improve PAUTH transparency

jackgallagher-arm:
Previously direct/indirect branches with pointer authentication would be mangled to strip the pointer authentication code from the address using an xpaci instruction.

This means that code that might fail when running natively (because the pointer authentication doesn't pass) could succeed under DynamoRIO because the pointer was not being authenticated.

This commit changes the mangling code to use auti* instructions to authenticate the pointer instead, and adds tests to check that it behaves correctly for all the branch and authenticate instructions.

Issue: #5623
Fixes: #5623


suite/tests/linux/mangle_pauth.c
```diff
@@ -0,0 +1,384 @@
...
+#    define NON_FAULT_TEST(instr_name)            \
+        LOG("Non-fault test: " #instr_name "\n"); \
+        test_##instr_name(false)
+
+    FOR_EACH_TEST(NON_FAULT_TEST);
+
+    intercept_signal(SIGSEGV, (handler_3_t)&handle_signal, false);
+    intercept_signal(SIGILL, (handler_3_t)&handle_signal, false);
+
+#    define FAULT_TEST(instr_name)                                                     \
```
derekbruening, line 164: Ditto: do-while

suite/tests/linux/mangle_pauth.c
```diff
@@ -0,0 +1,384 @@
...
+#define MODIFIER_REG x4
+
+#define FUNCNAME test_retaa
+        DECLARE_FUNC_SEH(FUNCNAME)
+GLOBAL_LABEL(FUNCNAME:)
+        SAVE_BRANCH_TARGET_ADDR(x30, TMP1_REG)
+        SAVE_BRANCH_INSTR_ADDR(LOCAL_LABEL(branch_instr), TMP1_REG, TMP2_REG)
+        paciasp
+        /* target_addr = signed_addr ^ (trigger_fault << 53); */
+        eor     x30, x30, TRIGGER_FAULT_ARG_REG, lsl 53
```
derekbruening, line 225: (To make sure I understand: we enter this function with a regular branch-and-link.  We save the (plain) return address and the PC of our RETAA into global vars for the signal handler to access.  We then encode the return address, flip bit 53 if we're asked to generate a fault, and ask to decode and return.)

derekbruening, line 225: Why 53?  Just arbitrary I assume?  It's used in many places.  Let's have a named constant.

suite/tests/linux/mangle_pauth.c
```diff
@@ -0,0 +1,384 @@
...
+            SIGLONGJMP(mark, TEST_PC_MISMATCH);
+
+    } else if (signal == SIGILL) {
+        LOG("Handled SIGILL:\n");
+        /* CPU has FEAT_FPACCOMBINE so the branch instruction generated an authentication
+         * failure exception and the fault PC should match the branch instruction address.
+         */
+        const uintptr_t fault_pc = ucxt->uc_mcontext.pc;
+        LOG("    fault_pc =             " PFX "\n", fault_pc);
+        LOG("    branch_target_addr =   " PFX "\n", branch_target_addr);
```
derekbruening, line 128: Better to print branch_instr_addr.

core/arch/aarchxx/mangle.c
```diff
@@ -1346,6 +1351,201 @@ mangle_direct_call(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
...
+        default:
+            ASSERT_NOT_REACHED();
+            key_a = true; /* Quell compiler maybe-uninitialized warning. */
+        }
+
+        PRE(ilist, instr,
+            INSTR_CREATE_cmp(dcontext, opnd_create_reg(scratch),
+                             opnd_create_reg(IBL_TARGET_REG)));
+
+        PRE(ilist, instr, INSTR_CREATE_xpaci(dcontext, opnd_create_reg(IBL_TARGET_REG)));
```
derekbruening, line 1523: Is #6276 fixed too?

core/arch/aarchxx/mangle.c
```diff
@@ -1346,6 +1351,201 @@ mangle_direct_call(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
...
+        default: ASSERT_NOT_REACHED();
+        }
+    } else {
+        /* If FEAT_FPAC is implemented, auti* instructions will generate a fault when
+         * authentication fails, but combined auth+branch instructions won't.
+         * We need to perform the authentication ourselves by stripping and re-signing
+         * the target pointer and comparing that to the original value.
+         * If they match the authentication passes. If they don't match the
+         * authentication has failed and we need to insert a 2-bit error code into the
+         * the address in IBL_TARGET_REG to leave it with a non-canonical value that
```
derekbruening, line 1439: s/the// (double "the")

core/arch/aarchxx/mangle.c
```diff
@@ -1346,6 +1351,201 @@ mangle_direct_call(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
...
+         * should fault when the app tries to branch to it.
+         *
+         *     mov      scratch, IBL_TARGET_REG
+         *     xpaci    scratch                  // Remove PAC
+         *     paci*    scratch                  // Re-sign pointer
+         *     cmp      scratch, IBL_TARGET_REG  // Check addresses match
```
derekbruening, line 1445: Doesn't this clobber the app's condition codes, so we have to save and restore them?  Can we instead do a sub and cbz?

core/arch/aarchxx/mangle.c
```diff
@@ -1346,6 +1351,201 @@ mangle_direct_call(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
...
+                             opnd_create_reg(IBL_TARGET_REG)));
+
+        PRE(ilist, instr, INSTR_CREATE_xpaci(dcontext, opnd_create_reg(IBL_TARGET_REG)));
+
+        instr_t *end_label = INSTR_CREATE_label(dcontext);
+        PRE(ilist, instr,
+            INSTR_PRED(INSTR_CREATE_bcond(dcontext, opnd_create_instr(end_label)),
+                       DR_PRED_EQ));
+
+        /* Insert the error code to make the address fault */
```
derekbruening, line 1530: nit: Missing trailing '.'

core/arch/aarchxx/mangle.c
```diff
@@ -1346,6 +1351,201 @@ mangle_direct_call(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
...
+        const uint64 error_code = key_a ? 1 : 2;
```
derekbruening, line 1531: Where does this come from?  Is this defined in the spec for how the non-canonical upper bits are constructed?  Or something just for us?  Add a comment.

core/translate.c
```diff
@@ -1104,7 +1104,14 @@ recreate_app_state_from_ilist(dcontext_t *tdcontext, instrlist_t *ilist, byte *s
                         LOG(THREAD_GET, LOG_INTERP, 2,
                             "recreate_app -- found valid state pc " PFX "\n", answer);
                     } else
-#endif /* X86 */
+#elif defined(AARCH64)
+                    int op = instr_get_opcode(inst);
```
derekbruening, line 1108: This should be up in the cases in `translate_walk_track_post_instr` to avoid `unsupported_mangle` being set in the first place.  Probably the other instrs generated here like xpaci should be in there too so we can confidently generate full state in the middle for asynch translation (synchall for detach, e.g.: don't make it retry b/c it landed on xpaci).

api/samples/opcode_count.cpp
```diff
@@ -67,21 +67,33 @@ static droption_t<int>
...
+                                      "Print results to STDOUT.");
+#else
+static droption_t<bool> show_results_(DROPTION_SCOPE_CLIENT, "show_results", false,
+                                      "Print results to STDOUT",
+                                      "Print results to STDOUT.");
+#endif
+
 static uintptr_t global_opcode_count = 0;
 static uintptr_t global_total_count = 0;
+static bool show_results = false;
```
derekbruening, line 82: This seems superfluous with the option.

api/samples/opcode_count.cpp
```diff
@@ -67,21 +67,33 @@ static droption_t<int>
            "The opcode to consider when counting the number of times "
            "the instruction is executed. Default opcode is set to add.");
 
+#ifdef SHOW_RESULTS
+static droption_t<bool> show_results_(DROPTION_SCOPE_CLIENT, "show_results", true,
```
derekbruening, line 71: Hmm, I guess this does still satisfy the expectations of SHOW_RESULTS for exported samples vs the internal tests.  Seems better to have the ifdef only around the true/false default value to avoid duplicating the rest.

suite/tests/common/pauth.c
```diff
@@ -0,0 +1,58 @@
...
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
+ * DAMAGE.
+ */
+
+#include <stdio.h>
+
+void
+leaf()
+{
+    printf("leaf\n");
```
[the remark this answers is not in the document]
philramsey-arm: File removed.

suite/tests/common/pauth.c
```diff
@@ -0,0 +1,58 @@
...
+    printf("branch\n");
+    leaf();
+}
+
+int
+main()
+{
+    printf("Hello world!\n");
+
+    branch();
```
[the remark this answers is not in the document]
philramsey-arm: File removed.

core/translate.c
jackgallagher-arm: Done.

core/arch/aarchxx/mangle.c
jackgallagher-arm: Done.

core/arch/aarchxx/mangle.c
jackgallagher-arm: Done.

core/arch/aarchxx/mangle.c
jackgallagher-arm: Done.

core/arch/aarchxx/mangle.c
jackgallagher-arm: Done.

suite/tests/linux/mangle_pauth.c
jackgallagher-arm: Done.

suite/tests/linux/mangle_pauth.c
jackgallagher-arm: I have moved the address corrupting code into a macro and added a comment to explain why we use bit 53.

jackgallagher-arm: Yes that's right.

suite/tests/linux/mangle_pauth.c
jackgallagher-arm: Done.

suite/tests/linux/mangle_pauth.c
```diff
@@ -0,0 +1,384 @@
...
+DECL_EXTERN(branch_target_addr)
+DECL_EXTERN(dummy_func)
+
+#define PASTE(a, b) a##b
+
+/* Use inside a function to declare or reference a function-local label.
+ * Expects FUNCNAME to be defined.
+ */
+#define _LOCAL_LABEL(label, unique_id) PASTE(unique_id, label)
+#define LOCAL_LABEL(label) _LOCAL_LABEL(label, FUNCNAME)
```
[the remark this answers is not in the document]
jackgallagher-arm: Done.

suite/tests/linux/mangle_pauth.c
```diff
@@ -0,0 +1,384 @@
...
+            M(brabz);        \
+            M(braa);         \
+            M(brab);         \
+            M(blraaz);       \
+            M(blrabz);       \
+            M(blraa);        \
+            M(blrab);        \
+        } while (0)
+
+#    define NON_FAULT_TEST(instr_name)            \
```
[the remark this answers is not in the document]
jackgallagher-arm: Done.

make/utils.cmake
```diff
@@ -355,31 +355,51 @@ function (check_sve_processor_and_compiler_support out vl_out)
...
-  set(sve2_prog "int main() {
-                    asm(\"histcnt z0.d, p0/z, z0.d, z0.d\");
-                    return 0;
-                 }")
-  set(CMAKE_REQUIRED_FLAGS ${CFLAGS_SVE2})
+  set(CMAKE_REQUIRED_FLAGS ${c_flags})
   if (CMAKE_CROSSCOMPILING)
     # If we are cross-compiling check_c_source_runs() can't run the executable on the
-    # host to find out whether the target processor supports SVE2, so we assume it
+    # host to find out whether the target processor supports the featyre, so we assume it
```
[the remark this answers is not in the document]
jackgallagher-arm: Done.

suite/tests/linux/mangle_pauth.c
```diff
@@ -0,0 +1,384 @@
...
+    TEST_PASS = 1,
+    TEST_PC_MISMATCH = 2,
+};
+
+static void
+handle_signal(int signal, siginfo_t *siginfo, ucontext_t *ucxt)
+{
+    if (signal == SIGSEGV) {
+        LOG("Handled SIGSEGV:\n");
+        /* CPU does not have FEAT_FPACCOMBINE so it branched to a non-canonical address.
```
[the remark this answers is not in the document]
jackgallagher-arm: Yes. The test expects `FEAT_PAUTH` to be implemented as a baseline but there are two other architecture features that can be implemented in addition to `FEAT_PAUTH` that change the way certain instructions behave when authentication fails.
If `FEAT_FPACOMBINE` is implemented we expect to see a `SIGILL` when authentication fails on Linux, if it isn't we expect `SIGSEGV`.

My comment in mangle.c explains this in detail:
```
    /* There are three scenarios here depending on which architecture features are
     * implemented.
     *
     * FEAT_PAUTH:
     *     None of the pauth instructions fault if authenication fails.
     *     Instead AUTI* inserts an error code in to the pointer to leave it with a
     *     non-canonical value so a subsequent branch to the address will fault.
     *     On Linux SIGSEGV is raised with the non-canonical target address as the PC.
     *
     * FEAT_FPAC: (implies FEAT_PAUTH is also implemented)
     *     AUTI* instructions generate a fault when authentication fails.
     *     On Linux SIGILL will be raised with the faulting AUTI* instruction as the PC.
     *     Combined authenticate and branch instructions (RETA*, BLRA*, BRA*) behaviour
     *     is the same as FEAT_PAUTH.
     *
     * FEAT_FPACCOMBINE: (implies FEAT_PAUTH and FEAT_FPAC are also implemented)
     *     AUTI* instructions behave the same as with FEAT_FPAC.
     *     Combined authenicate and branch instructions behave the same way as AUTI*
     *     when authentication fails.
     *     On Linux SIGILL will be raised with the faulting instruction as the PC.
     *
     * For the FEAT_PAUTH and FEAT_PAUTHCOMBINE cases AUTI* and the combined auth+branch
     * instructions we are mangling behave the same way so we can just insert an AUTI*
     * to authenticate the pointer in IBL_TARGET_REG.
     * For the FEAT_FPAC case AUTI* and the combined auth+branch instructions behave
     * differently so we need to perform the authentication ourselves (see below).
     */
```

core/arch/aarchxx/mangle.c
jackgallagher-arm: Unfortunately not. The same problem preventing DR from emitting `xpaci` on macOS will prevent it from emitting `authi*` which we now emit instead. I have added a comment on that issue about what is needed to fix it.

suite/tests/linux/mangle_pauth.c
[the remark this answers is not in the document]
derekbruening: I didn't see that big mangle.c comment until much later in the review after posting my initial reaction.  The mangle.c comment does a good job of explaining; it was clear after that.  Maybe could point at it here, or should be ok as-is: resolving.

core/arch/aarchxx/mangle.c
derekbruening: Thanks.

derekbruening (approved):

PR #6857, opened 2024-06-26
i#5843 scheduler: Raise default drmemtrace sched quantum to 6M

derekbruening:
Fixes an inconsistency in the CLI drmemtrace scheduler quantum and the internal API by making them both the same at 6 million.  We pick 6 million to match 2 instructions per nanosecond with a 3ms quantum. The scheduler_launcher default is also made to match.

Issue: #5843


derekbruening:
a64 failure is sim.invariants missing exit #6733

ivankyluk (approved):

PR #6862, opened 2024-07-01
i#4474: Port client.strace test to AARCH64

philramsey-arm:
Port client.strace test to AARCH64.

Issue: #4474


philramsey-arm:
Will re-work this as VS 2019 does not support the _Atomic keyword.

PR #6870, opened 2024-07-09
i#6662 func_id_filter: large function IDs

edeiana:
When filtering function related markers (i.e., TRACE_MARKER_TYPE_FUNC_) by function ID we need to provide the original marker value (function ID) in the trace to func_id_filter.
When filtering out system calls, the ID can be large due to TRACE_FUNC_ID_SYSCALL_BASE (which is 0x100000000ULL == 4294967296 for x64) that is added to the system call number, and can cause an std::out_of_range error.
We now use std::stoull to handle large function IDs.
We add a test which runs func_id_filter (part of record_filter) with a large function ID and then checks that there is no function marker left in the filtered trace with basic_counts.

Issue #6662

### DIFF
--- a/clients/drcachesim/tools/filter/record_filter.cpp
+++ b/clients/drcachesim/tools/filter/record_filter.cpp
@@ -38,6 +38,7 @@
 #include <cstdio>
 #include <fstream>
 #include <iostream>
+#include <limits>
 #include <memory>
 #include <mutex>
 #include <string>
@@ -95,7 +96,12 @@ parse_string(const std::string &s, char sep = ',')
     std::vector<T> vec;
     do {
         pos = s.find(sep, at);
-        vec.push_back(static_cast<T>(std::stoll(s.substr(at, pos))));
+        unsigned long long parsed_number = std::stoull(s.substr(at, pos));
+        // Check that T can hold the parsed value. Otherwise, skip it.
+        if (parsed_number >= std::numeric_limits<T>::min() &&
+            parsed_number <= std::numeric_limits<T>::max()) {
+            vec.push_back(static_cast<T>(parsed_number));
+        }
         at = pos + 1;
     } while (pos != std::string::npos);
     return vec;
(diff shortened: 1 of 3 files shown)


clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -95,7 +95,7 @@ parse_string(const std::string &s, char sep = ',')
     std::vector<T> vec;
     do {
         pos = s.find(sep, at);
-        vec.push_back(static_cast<T>(std::stoi(s.substr(at, pos))));
+        vec.push_back(static_cast<T>(std::stoll(s.substr(at, pos))));
```
abhinav92003, line 98: This makes sense only when T is a 64-bit int. Maybe add a check that T can hold the parsed value.

clients/drcachesim/tools/filter/record_filter.cpp
```diff
@@ -95,7 +95,7 @@ parse_string(const std::string &s, char sep = ',')
     std::vector<T> vec;
```
abhinav92003, line 95: Do we have any record filter unit test that works with syscalls? If not, let's add one.

abhinav92003 (approved):

clients/drcachesim/tools/filter/record_filter.cpp
edeiana: Done.
If T cannot hold the value, we skip it.
Probably not the best way to handle it, but I guess it's slightly better than before (when we would simply truncate the value and add it to the vector anyway).
Happy to go with a better option if you have one in mind.

clients/drcachesim/tools/filter/record_filter.cpp
edeiana: Done.
I added `code_api|tool.record_filter_keep_func_ids`, which runs `func_id_filter` (part of `record_filter`) with a large function ID and then checks that there is no function marker left in the filtered trace with `basic_counts`.

PR #6881, opened 2024-07-15
Clarify commit template for single vs multi-part commits

derekbruening:
Removes "<issue-brief>" from the commit template title line as it is only for multi-part commits.  Adds a template line explaining what to do for multi-part commits.

Updates the unedited-template check.


xdje42 (approved):
LGTM

PR #6882, opened 2024-07-15
Update squash-and-merge docs

derekbruening:
Updates the docs on merging a pull request to the new scheme of using the pull request title + description.  The docs were still describing the old setup where devs had to manually edit a concatenated squash-and-merge message.

### DIFF
--- a/api/docs/code_reviews.dox
+++ b/api/docs/code_reviews.dox
@@ -90,11 +90,12 @@ can also auto-close a Dr. Memory issue with "Fixes: DynamoRIO/drmemory#NNN".
 You can see other options in the [GitHub
 documentation](https://help.github.com/articles/closing-issues-via-commit-messages/).
 
-The string "Fixes: #NNNNN" in a Pull Request description will also close that
-issue, even if the final commit message does not contain that string, so be sure to
-edit the PR description if it's decided that the issue should stay open.
-
 If a commit does not fully fix an issue, include a reference to the issue with a line of the form "Issue: #NNNNN".  Use this to add links to related issues as well, in a format that makes it easier for tools to locate within messages.
+
+Use line breaks to keep commit message lines below 80 characters.
+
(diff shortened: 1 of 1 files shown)


api/docs/code_reviews.dox
```diff
@@ -233,58 +234,19 @@ to make changes to your code just wastes resources.  Furthermore, in some
...
+review comments have been marked resolved, and all of the continuous integration
+tests pass can the change be merged into the master branch.  All commits that are
+part of the pull request branch are squashed together into a single commit on the
+master branch.  The commit message title for this single commit is the pull request
+title, with "(#NNNN)" auto-appended where NNNN is the pull request number.  The
+commit message body is the initial comment at the top level of the pull request: the
+pull request description.  This can be edited after the pull request is created and
+those edits will be used in the squash-and-merge commit message.  The reviewer may
+request changes to the message, which should be performed by editing it through the
+web interface.
```
edeiana, line 249: Since we're at it, do we want to add guidelines on: 1) #chars for line-break on the commit message? 2) use of markdown? (it displays nicely on the web interface, but it looks weird looking at plain text merge-commit messages)

edeiana, line 249: About the "Deleting the Branch" section below, I think Github automatically deletes the feature branch once the PR is merged to master. Should we remove this section?

edeiana (approved):

api/docs/code_reviews.dox
derekbruening: Deleted and put a sentence into the Merging section so devs will know the branch will be deleted.

derekbruening: I added up above in the message content section:

> Use line breaks to keep commit message lines below 80 characters.
>
> Markdown can make descriptions easier to read in the web view, and is allowed.  It
> can make things harder to read in text views through so keep that in mind.

derekbruening:
x32 failure is opcode_mix encoding #6303

PR #6887, opened 2024-07-17
i#2417 AArch64: Fix ldstex mangling

jackgallagher-arm:
The code to mangle excusive load and store instructions emits code that has UNPREDICTABLE behaviour on some AArch64 implementations.

For a sequence like this (from the client.ldstex test):

 ldaxp  (%x0)[8byte] -> %w1 %w2
 add    %w1 $0x0001 lsl $0x00 -> %w4
 add    %w2 $0x0001 lsl $0x00 -> %w5
 stlxp  %w4 %w5 -> (%x0)[8byte] %w3

The mangling would emit compare-and-swap code that does a 1x8-byte load instead of a 2x4-byte load:

 ldaxr  (%x0)[8byte] -> %x3
 ...
 stlxp  %w4 %w5 -> (%x0)[8byte] %w3

The store stays as a 2x4-byte access. This ldex/stex pair has the same address and access size, but access a different number of registers. This works on some AArch64 hardware but the behaviour is UNPREDICTABLE so this code can't be relied upon to work on all implementations:

    "An implementation of the LoadExcl and StoreExcl instructions can
    require that, in any thread of execution, the StoreExcl instruction
    accesses the same number of registers as the preceding LoadExcl
    instruction executed in that thread. If the StoreExcl instruction
    accesses a different number of registers than the preceding
    LoadExcl instruction in the same thread of execution, behavior is
    CONSTRAINED UNPREDICTABLE. As a result, software can rely on an
    LoadExcl/StoreExcl pair to eventually succeed only if they access
    the same number of registers."

    - Arm ARM DDI 0487K.a B2.17.5

On implementations that do not support ldstex pairs with different numbers of registers, the stlxp would always fail and the app would go in to an infinite loop.

I have changed the mangling code to modify the 2x4-byte stxp in this case and turn it in to a 1x8-byte stxr to match the ldxr so we should now always emit architecturally correct code.

Issue: #2417


core/arch/aarchxx/mangle.c
```diff
@@ -3149,9 +3149,7 @@ static instr_t *
 create_ldax_from_stex(dcontext_t *dcontext, instr_t *strex,
```
derekbruening, line 3149: > The code to mangle excusive load and store instructions emits code that has UNPREDICTALE behaviour on some AArch64 

spelling: UNPREDICTABLE

suite/tests/client-interface/ldstex.c
```diff
@@ -459,44 +459,28 @@ GLOBAL_LABEL(FUNCNAME:)
         add      w2, w2, #0x1
         stxp     w3, w1, w2, [x0]
         cbnz     w3, 2b
-        /* Test pair4-single8. */
```
derekbruening, line 462: Even if some behavior is officially unpredictable, we would prefer to handle it if it's easy to do so, since we may well see it in the field. It is not unheard of to encounter apps that execute officially-unpredictable instructions, especially low-level hacky code doing obfuscation or malware.

What will happen with this PR if the pair-singleton-mismatch test cases here are run under DR?  The checks for matching sizes will fail and we'll have an infinite loop?  To add back handling without introducing mismatches again for matching app code: for split-block we'd need some complex handling?  What about same-block cases: is that easy?

OTOH, we have already decided to not support several cases of unpredictable behavior for ldstex: mismatched sizes and addresses.  Since we've already done that for ldstex maybe this is ok.  Could you edit api/docs/ldstex.dox as part of this PR to augment the section that begins "Our scheme above checks for strict equality" to also state that we require pair parity or matching pairness or however it is best explained (probably needs more words than that).

core/arch/aarchxx/mangle.c
```diff
@@ -3191,16 +3189,8 @@ create_ldax_from_stex(dcontext_t *dcontext, instr_t *strex,
...
-        if (merge_pair) {
-            ASSERT(opsz == OPSZ_4);
-            *dest_reg = reg_resize_to_opsz(*dest_reg, OPSZ_8);
-            return INSTR_CREATE_ldaxr(dcontext, opnd_create_reg(*dest_reg), memop);
-        } else {
-            return INSTR_CREATE_ldaxp(
-                dcontext, regop, opnd_create_reg(reg_resize_to_opsz(dest_reg2, opsz)),
-                memop);
-        }
+        return INSTR_CREATE_ldaxp(
```
derekbruening, line 3192: Include a comment somewhere stating that we're aware we won't handle pair-singleton mismatches: make it clear that's an explicit design choice.

derekbruening (approved):

core/arch/aarchxx/mangle.c
jackgallagher-arm: Done.

suite/tests/client-interface/ldstex.c
jackgallagher-arm: > Even if some behavior is officially unpredictable, we would prefer to handle it if it's easy to do so

It turned out to be fairly easy to get it to handle the mismatched-number-of-registers case so I have rewritten the patch to fix the bug but keep support that case.

core/arch/aarchxx/mangle.c
jackgallagher-arm: I have changed to patch so that we do support these mismatches and added comments explaining how and why it is done that way.

core/arch/aarchxx/mangle.c
```diff
@@ -3863,7 +3864,64 @@ mangle_exclusive_store(dcontext_t *dcontext, instrlist_t *ilist, instr_t *instr,
...
+     * hardware, however it does work reliably on some AArch64 implementations so it is
+     * possible we will find it in app code and we have decided that DynamoRIO should be
+     * able to handle it.
+     */
+    instr_t *orig_instr_clone = instr;
+    if (!is_pair && instr_num_srcs(instr) > 1) {
+        /* We are mangling a 2x4-byte stxp and we need to replace it with a 1x8-byte
+         * stxr.
+         *
+         * First we need to the combine the values of the 2 32-bit source registers into
```
derekbruening, line 3888: s/the combine/combine/

suite/tests/client-interface/ldstex.c
derekbruening: OK, SGTM.

derekbruening (approved):

core/arch/aarchxx/mangle.c
jackgallagher-arm: Done.

